PR #1127, opened 2026-06-19
MvP-2011 grpc close

ctnguyen:
This PR resolve [#2011](https://github.com/bitcoin-sv/teranode/issues/2011)


ordishs (approved):
Approving — this is solid, careful work with a coherent ownership model ("daemon constructs and owns, services borrow and must not close") and real verification behind it. Verified locally against the PR branch:

```
go build ./...                                              → exit 0
go vet (servicemanager, netsync, propagation, validator,
        blockchain, daemon, stores/blockchain)              → exit 0
go test servicemanager, netsync                             → ok
go test -race -run TestAnnounceTx ./services/legacy/netsync → ok
go test propagation, validator, blockchain                  → ok
```

### Strengths
- Idempotency handled properly — `connCloseOnce`/`connCloseErr` in the blockchain client correctly cover both close paths (synchronous `Close()` and the subscription ctx-done goroutine).
- `ownsConn` on the propagation client correctly avoids closing a caller-injected conn while still draining the client-owned batcher.
- Bounded drains (5s) sit under the 30s stop window; `DrainBatcher` leaves a wedged flush to finish rather than blocking shutdown.
- Real tests, no mocking.

### Comments (non-blocking, but please address #3 or accept it explicitly)

**1. `ServiceManager.Wait()` now returns non-nil on shutdown.** Previously `context.Canceled` returned `nil` and `Stop()` failures were only `Warn`-logged; now they propagate (aggregated) at `Error`. This can change the daemon **process exit code** on shutdowns that were previously "clean." Please call this out in the PR description so CI/supervisors/k8s keying on exit 0 are checked.

**2. `StopTimeout` default 5s → 30s.** Worst-case per-service shutdown latency grows. Deployment `terminationGracePeriodSeconds` (k8s) / stop grace (compose) must be ≥ the configured value or the orchestrator SIGKILLs mid-flush, defeating the purpose.

**3. `propagation.Stop()`: `ps.udpWg.Wait()` is unbounded.** It runs before the producer stop and never selects on `ctx.Done()`. A single hung `ProcessTransaction` blocks `Stop()` *past* `StopTimeout` — the timeout cancels ctx but doesn't interrupt `Wait()`. Low likelihood, but it silently breaks the bounded-stop guarantee the rest of this PR is built around. Suggest a ctx-bounded wait (`Wait()` in a goroutine + `select` on `ctx.Done()`).

**4. Scattered `interface{ Close() error }` assertions** (`cmd/checkblock`, `getfsmstate`, `setfsmstate`, `monitor`, `diagnose`, `blockvalidation/Server`). If a client's `Close` signature ever drifts (e.g. to `Close(ctx)`), the assertion silently fails → silent leak, no compile error. Referencing a single shared named interface (you already have `clientCloser` in the daemon) would turn drift into a build failure.

**5. `SQL.Close` only partially idempotent** — `backgroundDone` is guarded but the cache `Stop()`s aren't. No live double-close today (the blockchain `Server.Stop()` does not close the state store — only `closeStores` does), but the new interface doc says "called once"; worth confirming the cache `Stop()`s tolerate a repeat call.

**6. Two sources of truth for the 30s default** — `settings.go` hardcodes `"30s"` to dodge the import cycle with `servicemanager.DefaultStopTimeout`. Acceptable, but a drift hazard.

**7. Minor:** `daemon/test_daemon.go` constructs a `propagation.Client` that's never closed in teardown (pre-existing test-infra leak).

oskarszoon (changes requested):
Solid shutdown hardening — the panic/double-close/use-after-free vectors a refactor like this invites are each properly fenced (`sync.Once` conn close, RWMutex tx-announce guard with race tests, `SafeSend`+recover on Kafka, correct LIFO close ordering). Builds clean, new tests pass under `-race`. Requesting changes on one P1 only; the P2s below are non-blocking.

## P1 (blocking)

**`services/propagation/Server.go:604` — `udpWg.Wait()` ignores the stop context.**

`Stop(ctx)` blocks unconditionally on `ps.udpWg.Wait()` (the doc comment even marks `ctx` as "unused"). `udpWg` tracks in-flight UDP `ProcessTransaction` workers, which publish to the validator Kafka producer on the long-lived listener ctx, not the stop ctx. If a publish blocks (unreachable broker / full async buffer), that worker's `Done()` never fires and `Stop()` hangs forever. The service manager calls `Stop()` synchronously in its reverse-order loop (`service_manager.go:298`), so this stalls the entire shutdown: every service ordered before Propagation never gets stopped, and the configured `StopTimeout` is silently not enforced for this service.

Fix by racing the wait against `ctx.Done()` — same pattern as `DrainBatcher` already in this PR:

```go
waitDone := make(chan struct{})
go func() { ps.udpWg.Wait(); close(waitDone) }()
select {
case <-waitDone:
case <-ctx.Done():
    ps.logger.Errorf("[Propagation] timed out waiting for UDP workers: %v", ctx.Err())
}
```

## P2 (non-blocking)

- **`util/servicemanager/service_manager.go:291-306` — no global shutdown deadline.** Per-service Stop timeouts are serial; this PR raises the cap 5s→30s, so worst case with ~14 services is ~7 min. It's a per-service cap (only hit if every service hangs) and the serial loop is pre-existing, so not blocking — but a single parent deadline over the per-service contexts (e.g. `settings.ServiceManagerGlobalShutdownTimeout`) would bound it.

- **`daemon/daemon_stores.go:297` — local-validator producers never flushed.** When `UseLocalValidator=true`, `GetValidatorClient` returns the `*Validator` (which owns 3 started Kafka producers) without `d.retainClient(...)` and never caches it, so `closeClients()` can't reach it — the exact coordinated-shutdown path this PR builds. `*Validator` has no `Close()`. Add `retainClient` + a `Close()` on `*Validator` that `Stop()`s the three producers.

- **`daemon/daemon.go:449-460` — `closeClients` drains batchers serially, outside the budget.** Runs after `ServiceManager.Wait()` (so outside `ServiceManagerStopTimeout`), serially calling `Close()` on each retained client; blockassembly (×~4 instances) + validator + propagation each run a 5s `DrainBatcher`. On a degraded broker this adds ~5s×N to exit, unobservable from the stop-timeout setting. Bounded per-client so it can't wedge, but worth fanning out concurrently or wrapping in one aggregate deadline; at minimum log start/elapsed.

- **`services/validator/Server.go:412-455` and `services/blockvalidation/BlockValidation.go:2470-2477` — new drain/producer-stop paths untested.** The Stop tests construct the servers with nil literals, so the new Kafka producer `Stop()` calls and the `*Validator` batcher drain-before-Stop ordering invariant are never exercised. Production code is correct and nil-guarded — these are coverage gaps, but the drain-ordering invariant has zero regression protection. Add subcases with non-nil mock producers / a `*Validator` carrying a `txmetaKafkaBatcher`.

A few doc/comment nits (e.g. the netsync `Stop()` comment at `manager.go:2396` crediting `handlerDone` instead of `sendDuringShutdown`'s recover for INV-producer send safety; the blockchain `Client.Close()` "synchronously" wording at `Client.go:268`) — minor, happy to drop inline if useful.

freemans13 (approved):
Approving. This is careful, well-documented work that fixes real gRPC connection/goroutine leaks and a genuine dropped-error bug in `ServiceManager.Wait`.

**Strengths**
- Idempotency handled correctly: `blockchain.Client.closeConn` (sync.Once), `KafkaAsyncProducer.Stop` (shuttingDown/closed guards).
- The one genuinely panic-prone path — go-batcher v2.0.4 `Put`-after-`Close` — gets the strongest treatment (RWMutex quiesce-then-drain + a concurrency test). Kafka producer-stop changes are safe by construction since `Publish`/`TryPublish` are guarded with `util.SafeSend`.
- `ownsConn` correctly avoids closing caller-injected conns; bounded `DrainBatcher` keeps a wedged flush from stalling shutdown; propagation batcher value→pointer fixes a latent mutex-copy.
- Verified zero external callers of the renamed `Stop()`→`Close()` methods, so the rename breaks nothing.

**Follow-ups (non-blocking)**
1. Confirm the UTXO-store ↔ blockchain-client teardown ordering: the blockchain client is injected into the UTXO store as a BlobDeletionScheduler, yet `closeClients` runs before `closeStores` — verify no store-Close path calls back into the now-closed client.
2. The `interface{ Close() error }` type-assertion pattern silently re-leaks if a client's Close signature ever drifts (no compile error). Consider a follow-up that declares `Close() error` on the client interfaces.
3. Add a small unit test for `retainClient`/`closeClients` (close exactly once, including on a Close that errors).

Please ensure CI is green on the full verification suite (`go test -race ./...`, vet, golangci-lint, staticcheck) — especially the SQL store `Close(ctx)` signature change across the ~30 touched test files.

ctnguyen:
I have addressed the blocking issue raised by @oskarszoon and some issues raised by @ordishs

oskarszoon (approved):
Re-reviewed at `5ef558301`. The P1 is resolved and verified, and the fix went further than asked: the bounded-producer-stop pattern is now factored into a shared, tested `util/kafka/StopProducerCtx` helper and applied across every async-producer Stop site (validator, blockchain, p2p, subtreevalidation, validator.Close, netsync), each passing the per-service stop budget. The two substantive P2s (local-validator producers not flushed; missing drain-ordering / Close coverage) are fixed with real `-race` tests. Built + vetted + ran all touched Stop/Close tests under `-race` locally — green.

Approving. None of the below blocks merge — follow-up only.

**Deferred P2s (non-blocking, your call):**
- `util/servicemanager/service_manager.go:294` — per-service stop timeouts are serial with no global cap; worst case ~N×StopTimeout. A single parent deadline over the per-service contexts would bound it.
- `daemon/daemon.go:449-460` — `closeClients` drains batcher-bearing clients serially, outside the stop budget; a degraded broker adds ~5s×N to exit. Fan out or wrap in one aggregate deadline; at minimum log start/elapsed.

**Nits:**
- `services/legacy/netsync/manager.go:2396` — the `Stop()` comment still says "handlerDone above guarantees no more sends to legacyKafkaInvCh", but `QueueInv` sends from per-peer read-loop goroutines that `handlerDone` doesn't join; the actual send-on-closed guard is `sendDuringShutdown`'s recover. Code is panic-safe — just fix the comment so nobody deletes the recover thinking `handlerDone` covers it.
- `services/p2p/Server.go:1670` — switching producer stops to `StopProducerCtx` means producer-stop errors are now log-only and no longer aggregated into `Stop()`'s returned error (consumer-close errors still are). Fine if intentional; flagging the behavior delta.
- `services/propagation/Server.go` — kept its inline bounded producer-stop instead of adopting the new `StopProducerCtx` helper. Functionally identical; consolidating would be tidier.

PR #928, opened 2026-05-21
fix(utxo): drain in-flight batched writes on shutdown

freemans13:
## The bug

Restarting teranode during legacy catchup can leave the UTXO store inconsistent. After redeploying a testnet node mid-catchup we observed:

\`\`\`
ERROR | netsync/handle_block.go:1018 | legacy| [extendTransactions] called for block ... height 54494 DONE in 817.351µs with error: PROCESSING (4): failed to decorate previous outputs for tx a1f6a4ff...c948afcca -> PROCESSING (4): failed to decorate previous outputs: 1 missing
INFO  | legacy/netsync/manager.go:886 | legacy| Resetting header sync state at height 54493
\`\`\`

Looping forever: peer sends block → tx references a parent UTXO that's not in our store → fail → header-resync → retry → same error. The only recovery is a DB wipe.

## The cause

All \`utxo.Store\` backends (aerospike, sql, and the queue/postgres store on a sibling PR) front their writes with \`go-batcher\` in \`background=true\` mode. Every \`Create\`/\`Spend\`/\`Get\`/\`Unlock\` call enqueues an item and returns immediately; the actual DB or Aerospike write happens in a worker goroutine.

When the process receives SIGTERM:
1. \`util/servicemanager/service_manager.go:66\` cancels the global context.
2. Services unwind via context cancellation and \`sm.Wait()\` returns.
3. \`daemon/daemon.go:376\` takes the \`case err = <-waitErr\` branch.
4. The function returns; the process exits.

**Nothing in this path closes the utxo store**, so the batcher workers are killed mid-flight. Items still in the input channel — and any batch dispatched but not yet committed by Aerospike/Postgres — are lost. The caller had already received a successful \`Create\`/\`Spend\` return (background mode acks on enqueue), the parent block had been committed to the blockchain store via gRPC, and on restart the chain advances past blocks whose UTXOs never reached durable storage. Subsequent blocks that spend those outputs fail with \`failed to decorate previous outputs: N missing\`.

Note that the existing \`closeStores\` in \`daemon.go\` only ran on the \`<-d.doneCh\` shutdown branch (explicit \`daemon.Stop()\`) and only closed the tx/subtree/temp blob stores. The OS-signal path never called it at all.

## The fix

- Add \`Close(ctx context.Context) error\` to \`utxo.Store\` (matching blob store naming).
- Implement \`Close\` on every backend:
[...]

PR #1129, opened 2026-06-19
test(utxo/aerospike): cover setLockedBatch child-record path

### LINKED ISSUE #1033: setLockedBatch: same-pool child-record recursion can exhaust the lockedBatcher worker pool
## Summary

`setLockedBatch` (`stores/utxo/aerospike/locked.go`) runs on a `lockedBatcher` worker and, when a transaction has child (pagination) records, re-submits those child records to the **same** `lockedBatcher` and then blocks on `g.Wait()`. The `lockedBatcher` runs a bounded worker pool (`SetMaxConcurrent(batcherMaxConcurrent)`, `aerospike.go`). Under enough concurrent multi-record locks, every worker can end up parked on a parent batch waiting for child records that themselves need a worker — a pool-exhaustion deadlock.

## Current state

PR #1025 added `waitForLockedResult`, which bounds the child wait to `batcherWait` (≈ batch policy `TotalTimeout`, ~5m). That converts a **permanent** hang into a bounded `ServiceUnavailable` — a strict improvement, but it does not eliminate the recursion. Under a burst of large-tx locks the operation can still fail with timeouts instead of completing.

(Raised in PR #1025 review by @ordishs and the automated review — recorded there as a non-blocking follow-up.)

## Trigger

Only **paginated** transactions (output count `> utxoBatchSize`, i.e. `ChildCount > 0`) recurse; single-record txs never do. `SetLocked` itself is on several hot-ish paths (block assembly `BlockAssembler.go:2474,2908`, block validation, validator, conflicting-tx processing), so the failure window is: a burst of large-tx locks concurrent enough to saturate the pool.

## Proposed fix

Stop recursing onto the same pool. After the master `setLocked` UDF returns `ChildCount`, build the child `setLocked` UDF records inline and issue a **single direct `s.batchOperate`** for them (bypassing the batcher), parse the results, aggregate, and signal the parent's `errCh`. This removes the `PutCtx`-to-same-pool + errgroup entirely while keeping the children batched in one call. It is unit-testable via the `batchOperateFn` seam, and `TestStore_AerospikeSplitTx` already exercises the child path.

Alternative: dispatch child records onto a separate batcher/pool.

## Scope

Contained to `setLockedBatch` (~25-30 lines). Sensitive (locking semantics) — warrants its own focused review + a dedicated unit test.

icellan:
## Summary

Issue #1033 reported a pool-exhaustion deadlock in `setLockedBatch`: it used to re-queue a multi-record tx's child/pagination records back into the **same** `lockedBatcher` worker pool and block on them, so a burst of large-tx locks could park every worker on children that themselves needed a worker.

**That fix already landed in #928**, which collects all child records and issues a single direct `s.batchOperate` (no batcher re-entry). What the issue also asked for — and what was still missing — is a dedicated **unit** test of the child-record path. The only existing coverage (`TestClose_DrainsQueuedSetLockedMultiRecord`) is an Aerospike integration test, skipped under `testing.Short()` / `make test`.

## What this adds

`stores/utxo/aerospike/set_locked_child_test.go` — hermetic unit tests driving the child path through the `batchOperateFn` seam (no live Aerospike, runs in short mode):

- **MultiRecordSuccess** — master `childCount=N` drives a second `batchOperate` carrying exactly N child records keyed for indexes `1..N`; item completes `nil`.
- **MixedChildCounts** — single-record items complete in the master pass; only paginated items contribute children; each item signalled exactly once.
- **ChildBatchError** — a failed child `batchOperate` hits only child-bearing items; single-record items are unaffected.
- **ChildRecordFailure** — a per-record child error fails only its owning item; siblings unaffected.

`newTestStoreForGet` leaves `lockedBatcher` nil, so every `require.NotPanics` is also a guard against the recursion being reintroduced.

## Notes

- No production code changes — the #928 fix is correct as-is; this is regression/characterization coverage.
- Verified not vacuously green via a throwaway mutation (force `extraRecords=0`), which makes the child-path assertions fail; reverted.
- `go test -race -short -run TestSetLockedBatch ./stores/utxo/aerospike/` passes; `go vet` and `golangci-lint` clean for the new file.

Closes #1033.

### DIFF
--- a/stores/utxo/aerospike/set_locked_child_test.go
+++ b/stores/utxo/aerospike/set_locked_child_test.go
@@ -42,11 +42,17 @@ func writeLuaErr(rec aerospike.BatchRecordIfc, msg string) {
 }
 
 func newLockedItem(b byte) *batchLocked {
-	return &batchLocked{ctx: context.Background(), txHash: chainhash.Hash{b}, setValue: true, errCh: make(chan error, 1)}
+	// Production sizes errCh at cap 1; here we use cap 2 so that a regression
+	// which double-signals an item (production sends via the non-blocking
+	// trySignal) is not silently dropped on the second send — it lands in the
+	// buffer where recvOnce's "exactly once" check can observe it.
+	return &batchLocked{ctx: context.Background(), txHash: chainhash.Hash{b}, setValue: true, errCh: make(chan error, 2)}
 }
 
-// recvOnce returns the single value delivered on a buffered-1 completion channel
-// and asserts it is signalled exactly once.
+// recvOnce returns the value delivered on a completion channel and asserts the
+// item was signalled exactly once. The channel is sized cap 2 (see
+// newLockedItem) so a spurious second trySignal is buffered rather than dropped,
+// making double-signalling observable here.
 func recvOnce(t *testing.T, ch chan error) error {
 	t.Helper()
 
--- a/stores/utxo/aerospike/set_locked_child_test.go
+++ b/stores/utxo/aerospike/set_locked_child_test.go
@@ -1,0 +1,233 @@
+package aerospike
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/bsv-blockchain/aerospike-client-go/v8"
+	"github.com/bsv-blockchain/aerospike-client-go/v8/types"
+	"github.com/bsv-blockchain/go-bt/v2/chainhash"
+	"github.com/bsv-blockchain/teranode/util/uaerospike"
+	"github.com/stretchr/testify/require"
+)
+
+// These tests cover setLockedBatch's child/extra-record path — the inline
+// "build children, issue one direct batchOperate, aggregate per item" logic
+// (locked.go) that replaced the old same-pool lockedBatcher recursion (issue
+// #1033, fixed by #928). The integration regression test
+// (TestClose_DrainsQueuedSetLockedMultiRecord) needs a live Aerospike and is
+// skipped in short mode; these drive the same branch hermetically through the
+// batchOperateFn seam.
+//
+// newTestStoreForGet leaves lockedBatcher nil, so any re-entry into the batcher
+// would panic — every require.NotPanics here is also a guard against the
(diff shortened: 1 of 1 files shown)


icellan:
Thanks @Copilot — good catch, and correct. With the test channels at cap 1, the production `trySignal` (non-blocking) would drop a spurious second signal, so `recvOnce`'s "exactly once" check could never actually fire — the assertion was effectively dead.

Fixed in 5c07b65b5: test `errCh` is now cap 2 so a second signal lands in the buffer where `recvOnce` observes it, and the helper's doc comment is updated to match. Verified with a throwaway double-signal mutation in `setLockedBatch`, which now fails as intended (`errCh signalled more than once`). Production keeps cap 1 — only the test needed widening to expose the regression.

gokutheengineer (approved):

oskarszoon (approved):
Solid hermetic addition. The 4 tests genuinely exercise the `setLockedBatch` child-record path through the `batchOperateFn` seam with real structural assertions — verified mock fidelity (writes into `rec.BatchRec().Record`, the field the real client sets), child-record ordering, digest-based `Key.Equals` checks, and the deliberate cap=2 `errCh` double-signal guard. All pass clean under `-race`. Fills a real gap — the path was previously integration-test-only, skipped under `-short`.

Two follow-up coverage gaps (P2, not blocking), both in the child pass and both reachable via the existing seam:

- **nil-child-response (`locked.go:245-246`)** — the child pass silently `continue`s when the record/Bins/`LuaSuccess` is nil, completing the item as **success**, while the master pass (`locked.go:163-168`) treats the same nil-response as an **error**. That asymmetry is itself an untested latent gap, not just missing coverage. `writeLuaOK`/`writeLuaErr` both populate the Record, so no test hits it.
- **per-record `.Err` (`locked.go:239-241`)** — uncovered. `TestSetLockedBatch_ChildBatchError` covers the whole-child-`batchOperate`-fails case; the per-record `.Err` field is a distinct mechanism and isn't exercised.

Issue #1001, opened 2026-06-01
aerospike-client-go/v8 nodeStats.updateOrInsert consumes 44-60% of legacy CPU during mainnet IBD


oskarszoon:
## Summary

The `bsv-blockchain/aerospike-client-go/v8@v8.7.1-bsv3` fork's per-record stats tracking (`nodeStats.updateOrInsert`) consumes 44–60% of the legacy service's CPU during mainnet IBD. Every record in every batch response triggers an atomic-map update; map contention scales with batch size and concurrency, dominating the hot path that should be doing useful work.

## Observed

30-second CPU profiles, mainnet IBD on 2026-06-01:

**eu-2 legacy** — sync rate ~6.6s/block, mixed block sizes:

```
44.64s total samples
- 43.88% github.com/bsv-blockchain/aerospike-client-go/v8.batchCommandOperate.Execute
  - 43.37% baseMultiCommand.parseResult
    - 42.47% batchCommandOperate.nsIter
      - 42.32% nodeStats.updateOrInsert
        - 39.20% nodeStats.updateOrInsert-range1
          - 20.99% atomic/map.(*Map[int,uint64]).Set
```

**eu-3 legacy** — sync rate ~3.3s/block, smaller average blocks:

```
39.98s total samples
- 59.60% batchCommandOperate.Execute
  - 58.75% parseResult
    - 57.80% nsIter
      - 57.60% nodeStats.updateOrInsert
        - 54.65% nodeStats.updateOrInsert-range1
          - 30.47% atomic/map.Set
```

Both hosts running `teranode v0.15.2-beta-4` (commit `e125d1ef8`), aerospike client `v8.7.1-bsv3`.

## What the hot path is doing

Per the call graph: for every record returned in a batch response, the parse loop calls `nsIter`, which calls `nodeStats.updateOrInsert(...)`, which calls into the internal atomic-map `Set`. With 30+ concurrent batch operations (per-service batchers, partition workers, parallel txMap processing) and thousands of records per batch, the atomic map becomes the contention bottleneck — `atomic/map.Set` alone is 20–30% of legacy's CPU.

## Why this hurts more on BSV/teranode than typical aerospike workloads

[...]

PR #1132, opened 2026-06-22
build(deps): bump aerospike-client-go to v8.7.1-bsv4

icellan:
## What

Bumps `github.com/bsv-blockchain/aerospike-client-go/v8` from `v8.7.1-bsv3` to `v8.7.1-bsv4`.

`bsv4` is a superset of `bsv3` — it carries the existing packer-pool work **and** the fix for the per-record batch result-code stats hot path (fork PR bsv-blockchain/aerospike-client-go#4):

- Each batch record's result code is now attributed to that record's **own** namespace. Previously every batch command recorded once per record while iterating the *whole batch's* namespace set — **O(N²) per batch**, which profiled at **44–60% of legacy CPU during mainnet IBD** (`nodeStats.updateOrInsert`, with `atomic/map.Set` alone at 20–30%).
- The counter increment moved from a racy `Get`-then-`Set` to a single-lock atomic update (the old path lost ~99% of increments under concurrent batchers).

## Why this is inert right now

Aerospike client metrics remain **disabled** in `util/aerospike.go` (`client.EnableMetrics` stays commented out). So this PR makes the corrected library available but does **not** exercise the fixed path. There is no behaviour change for teranode today.

Re-enabling `EnableMetrics` + re-profiling a busy legacy to confirm the `<10%` target is a deliberate, separate follow-up. That is why this references — but does not close — the issue.

## Verification

- `go build ./...`: clean
- `go.sum` updated with the new module hashes

Refs #1001


oskarszoon (approved):
The bump pulls in exactly PR #4 (batch result-code stats O(N²) fix — same 10 files, reviewed and approved this round, no P0/P1) plus a no-op `bsv/v8` merge; the dependency's own go.mod hash is unchanged (no transitive deps), go.sum is correct, and PR #4's changes are internal metrics only (no public API touched, no teranode build/behavior risk). Good to merge.

One thing to track, not blocking: PR #4's bounds-guard inconsistency (`batch_command_exists.go` guards `keys[batchIndex]`, the other 8 batch commands don't) now ships downstream — harmless unless the aerospike server returns an out-of-range `batchIndex`, but worth tightening in the fork.

ordishs (approved):
Approving — clean, minimal dependency bump.

Verified locally:
- `EnableMetrics` is still commented out (`util/aerospike.go:504`), so the fixed batch result-code path is dormant — no runtime behavior change today, consistent with the "Refs #1001" (not Closes) framing.
- `bsv3` is referenced nowhere in `*.go` or `go.mod` outside `go.sum`.
- go.mod hash is unchanged between bsv3 and bsv4 — consistent with a code-only fork change (no new transitive deps).
- `go mod tidy` resolves `v8.7.1-bsv4` cleanly.

One suggestion before merge: run `go mod tidy`. The PR added the bsv4 hashes but left the now-orphaned `v8.7.1-bsv3` entries in `go.sum`:

```
$ go mod tidy && git diff --stat go.sum
 go.sum | 2 --   # removes the two stale v8.7.1-bsv3 lines
```

Harmless to builds, but worth dropping the dead entries.

PR #1133, opened 2026-06-22
fix(blockvalidation): off-load setMined retry backoff to a cancellable goroutine (#4655)

gokutheengineer:
## Summary

  The single `setMinedChan` worker performed a blocking `time.Sleep(backoff)` and then
  re-sent the block onto `setMinedChan` from within the **only consumer** of that channel.
  This had two problems:

  - **Throughput stall** — during the backoff (exponential, up to 16s per attempt) the
  worker stopped draining `setMinedChan`, so incoming `BlockSubtreesSet` mined-status
  notifications backed up in the buffer and mined-status propagation stalled.
  - **Deadlock risk** — once the buffer (cap 1000) saturated, `u.setMinedChan <- blockHash`
  blocked forever, because the only goroutine that could free a slot was the one blocked
  trying to send.

  ## Fix

  Extracted the wait + re-send into a new `scheduleSetMinedRetry` helper that runs them on a
  dedicated goroutine:

  - The worker returns immediately and keeps draining the channel during backoff.
  - Both the timer wait and the send are cancellable via `ctx`, so the retry can't outlive
  shutdown.
  - The goroutine is registered with the existing `backgroundTasks` WaitGroup, so `Wait()`
  blocks until any pending retry drains — no goroutine leak.

  This is lighter on the hot path, not heavier.

  ## Changes

  - `services/blockvalidation/BlockValidation.go` — replaced the inline `time.Sleep` +
  self-send with `scheduleSetMinedRetry`.
  - `services/blockvalidation/BlockValidation_test.go` — added `TestScheduleSetMinedRetry`
  covering: re-queue after backoff without blocking the caller, `ctx` cancellation dropping
  the retry, and no deadlock when the buffer is full.

### DIFF
--- a/services/blockvalidation/BlockValidation.go
+++ b/services/blockvalidation/BlockValidation.go
@@ -216,6 +216,17 @@ type BlockValidation struct {
 	// (uint64 stored as any). Reset on success / MinedSet shortcut.
 	setMinedRetries sync.Map
 
+	// setMinedRetryPending marks blocks whose setTxMined retry is currently waiting
+	// out its backoff in a scheduleSetMinedRetry goroutine. Keyed by chainhash.Hash,
+	// value is struct{}. While an entry is present the setMinedChan worker skips
+	// re-processing that hash, so re-triggers arriving during the backoff window
+	// (e.g. a child re-triggering its not-yet-mined parent) do not race a second
+	// setTxMinedStatus and inflate setMinedRetries toward a premature drop. Set when
+	// a retry is scheduled, cleared when it fires (just before the re-enqueue) or on
+	// shutdown. Distinct from blockHashesCurrentlyValidated so the parent-finalization
+	// wait in checkParentProcessingComplete is unaffected by the backoff window.
+	setMinedRetryPending sync.Map
+
 	// revalidateBlockChan receives blocks that need revalidation
 	revalidateBlockChan chan revalidateBlockData
 
@@ -569,7 +580,19 @@ func (u *BlockValidation) start(ctx context.Context) error {
 	// start a worker to process the setMinedChan
 	u.logger.Infof("[BlockValidation:start] starting setMined goroutine")
 
+	// Register the worker in backgroundTasks (synchronously, before the goroutine
+	// launches) so the counter is >= 1 for the entire lifetime of the worker.
+	// scheduleSetMinedRetry does its own Add(1) from inside this worker; without a
+	// live registration the counter would transition 0 -> 1 there, which can race a
+	// concurrent Wait() on a zero counter during shutdown - illegal WaitGroup usage
+	// that can panic. Keeping the worker registered means every such Add(1) happens
+	// while the counter is already positive, and Wait() only returns once this worker
+	// (and any pending retries it scheduled) have drained. The worker's ctx is always
+	// cancelled before Server.Stop() -> Wait() runs, so this never deadlocks Wait().
+	u.backgroundTasks.Add(1)
+
 	go func() {
+		defer u.backgroundTasks.Done()
 		defer u.logger.Infof("[BlockValidation:start] setMinedChan worker stopped")
 
 		for {
@@ -604,6 +627,15 @@ func (u *BlockValidation) start(ctx context.Context) error {
 			if blockHeaderMeta.MinedSet {
 				u.logger.Infof("[BlockValidation:start][%s] block already has mined_set true, skipping setTxMined", blockHash.String())
 				u.setMinedRetries.Delete(*blockHash)
+				continue
+			}
+
+			// A retry for this block is already waiting out its backoff; the scheduled
+			// goroutine will re-enqueue it when it fires. Skip re-processing so re-triggers
+			// during the backoff window don't race a second setTxMinedStatus and inflate
+			// the retry counter (tripping the manual_intervention_required drop early).
+			if _, retryPending := u.setMinedRetryPending.Load(*blockHash); retryPending {
+				u.logger.Debugf("[BlockValidation:start][%s] setTxMined retry already pending, skipping", blockHash.String())
 				continue
 			}
 
@@ -659,21 +691,12 @@ func (u *BlockValidation) start(ctx context.Context) error {
 
 					backoff := setMinedRetryBackoff(int(attempt))
 					u.logger.Warnf("[BlockValidation:start][%s] setTxMined retry %d/%d in %s", blockHash.String(), attempt, setMinedMaxRetries, backoff)
-					time.Sleep(backoff)
 
-					select {
-					case <-ctx.Done():
-						return
-					default:
-					}
-
-					// put the block back in the setMinedChan for retry — non-blocking,
-					// because this runs on the worker goroutine that is the sole drainer
-					// of setMinedChan; a blocking self-send into a full channel wedges it.
-					// Re-queue order is best-effort (an overflow entry is only delivered
-					// after the channel drains); correctness relies on the worker's
-					// idempotent MinedSet/tryClaim guards, not on FIFO delivery.
-					u.enqueueSetMined(blockHash)
+					// Off-load the backoff wait + re-enqueue to a dedicated goroutine so this
+					// worker keeps draining setMinedChan (and its overflow) during the backoff
+					// instead of sleeping inline. The re-enqueue still goes through the
+					// non-blocking, overflow-backed enqueueSetMined.
+					u.scheduleSetMinedRetry(ctx, blockHash, backoff)
 					return
 				}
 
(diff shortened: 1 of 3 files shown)


ctnguyen (changes requested):
This PR addresses a real liveness problem: sleeping inside the only `setMinedChan` consumer makes the queue stop draining and can deadlock when the buffer fills. Moving the backoff wait and retry send into a context-aware goroutine is the right shape for keeping mined-status propagation flowing, and the new tests cover the intended caller behavior without widening the production diff much.

## Blocking issues

### ChiR1 - Register retry tasks under a live worker
**Problem:** `scheduleSetMinedRetry` calls `u.backgroundTasks.Add(1)` from the `setMinedChan` worker, but that worker is not itself registered in `backgroundTasks`. On shutdown, `Server.Stop()` calls `u.blockValidation.Wait()` after the service context is canceled, and `Wait()` can observe a zero counter while the still-running worker is between its `ctx.Done()` check and `scheduleSetMinedRetry`; a positive `Add` racing with `Wait` on a zero counter is illegal `sync.WaitGroup` usage and can panic.

**Why it matters:** A rare setTxMined failure during shutdown can crash the process instead of exiting cleanly, and the new graceful-shutdown tracking becomes unreliable.

**Fix:** Register the worker at the existing launch point in `start()`, with `Add(1)` executed synchronously before the goroutine starts, so it happens before any later `Stop()` -> `Wait()` call. If this helper may later be called outside that worker, add a separate lifecycle gate instead of relying on `ctx.Done()` checks alone.

```diff
@@ -548,8 +548,10 @@ func (u *BlockValidation) start(ctx context.Context) error {
 u.logger.Infof("[BlockValidation:start] starting setMined goroutine")
 
+u.backgroundTasks.Add(1)
 go func() {
+	defer u.backgroundTasks.Done()
 	defer u.logger.Infof("[BlockValidation:start] setMinedChan worker stopped")
 
 	for {
```

**Criticality:** 88%

## Non-blocking issues

### ChiR2 - Cover the production worker retry path
**Problem:** The new tests call `scheduleSetMinedRetry` directly, but the changed production path is the `start()` worker branch where `setTxMinedStatus` fails and the worker now schedules the retry instead of sleeping and self-sending. That integration site remains uncovered, which also leaves the new-code coverage gate below its required threshold.

**Why it matters:** Helper-level tests can pass while the worker wiring regresses, for example by failing to re-queue after `setTxMinedStatus` errors or by stalling the drain loop again.

**Fix:** Add a worker-level regression test that sends one hash that forces `setTxMinedStatus` to fail, immediately sends a second hash, and asserts the second hash is processed before the first retry backoff elapses while the first hash is re-queued later. Have the test fakes signal `nextHashProcessed` and `retryHashRequeued` channels at those observation points, and keep the backoff short via a test hook or injectable scheduler if needed.

```go
seen := func(ch <-chan struct{}) bool {
	select {
	case <-ch:
		return true
	default:
		return false
	}
}

blockValidation.setMinedChan <- retryHash
blockValidation.setMinedChan <- nextHash

	return seen(nextHashProcessed)
}, 500*time.Millisecond, 10*time.Millisecond)

	return seen(retryHashRequeued)
}, 2*time.Second, 10*time.Millisecond)
```

**Criticality:** 45%

### ChiR3 - Make the full-buffer retry test prove the send path
**Problem:** The `"cannot deadlock when the channel buffer is full"` subtest schedules a 1 ms retry and then cancels immediately after the scheduling assertion. Cancellation can win before the timer fires, so the test can pass by exercising only the earlier "cancel before timer" path rather than the full-channel send path it intends to protect.

**Why it matters:** The regression test may miss a future change that waits for the timer and then performs a non-cancellable send to a full `setMinedChan`.

**Fix:** Make the test wait with the channel full before cancellation, and avoid the strict scheduler-duration assertion. For example:

```go
u.scheduleSetMinedRetry(ctx, &h, 0)

done := make(chan struct{})
go func() {
	u.backgroundTasks.Wait()
	close(done)
}()

select {
case <-done:
	t.Fatal("retry exited before context cancellation while setMinedChan was full")
case <-time.After(25 * time.Millisecond):
}

cancel()
	select {
	case <-done:
		return true
	default:
		return false
	}
}, 2*time.Second, 10*time.Millisecond)
```

**Criticality:** 35%

## Recap

| ID | Description | Criticality |
|----|-------------|-------------|
| ChiR1 | Register retry tasks under a live worker to avoid WaitGroup misuse during shutdown | 88% |
| ChiR2 | Cover the production worker retry path so the changed call site and coverage gate are protected | 45% |
| ChiR3 | Make the full-buffer retry test deterministically cover the cancellable send path | 35% |

ctnguyen (approved):
The current diff addresses the root cause cleanly: the setMined retry wait and re-send now run in a cancellable background goroutine, so the sole worker continues draining and the old self-send deadlock is removed. Prior ChiR1 is resolved: `start()` registers the setMined worker with `backgroundTasks.Add(1)` before launching it and defers `Done()`, so retry-task `Add(1)` calls scheduled from that worker no longer form a 0-to-1 race against `Wait()`. Prior ChiR2 and ChiR3 are resolved: the production worker path is covered with a fault-injecting decorator that proves the next hash is processed during the first retry backoff and the failed hash is re-queued, and the full-buffer test now reaches the blocked-send path before cancellation instead of passing through the cancel-before-timer path. I do not see a new issue in the current diff: retry accounting and claim release appear unchanged around the scheduled retry, shutdown cancellation intentionally drops pending retries, and the added tests give strong, deterministic coverage with 100% new-code coverage.

icellan:
## Review — off-loaded setMined retry

Two issues worth addressing before merge. The off-loading itself is a good call (worker keeps draining during backoff), but tracking the worker in the WaitGroup and releasing the dedup claim during backoff each change behaviour in a way that bites.

### P1 — `Wait()` can hang indefinitely, ignoring the shutdown deadline

`BlockValidation.Wait()` is a bare `backgroundTasks.Wait()` (`BlockValidation.go:2495`), called from `Server.Stop()` at `Server.go:1064`. Now that the setMinedChan worker is registered in `backgroundTasks` (`Add(1)`/`Done` at `BlockValidation.go:581`/... ), `Wait()` blocks on it — and the worker only exits on the **Init** ctx (`start`'s `ctx.Done()`), not on the bounded `stopCtx` that `ServiceManager` passes into `Stop()`.

- Daemon path is fine: the errgroup cancels the shared ctx before `Stop()` runs.
- Any shutdown path that tears the server down **without first cancelling the Init ctx** — tests, alternate/embedded shutdown paths — now hangs forever, and the 5s `stopCtx` is silently discarded.

Before this PR the worker was untracked, so `Wait()` returned immediately regardless. Suggest making `Wait()` respect a context (select on `stopCtx.Done()`), or guarantee the worker ctx is always cancelled before `Wait()` is reached and document that invariant.

### P2 — releasing the dedup claim during backoff inflates the retry counter → premature `manual_intervention_required`

In the old code the `time.Sleep(backoff)` ran **inline inside the anon func**, so `blockHashesCurrentlyValidated` was held for the whole backoff and concurrent re-triggers for the same block were skipped. Now `scheduleSetMinedRetry` returns immediately, the anon func returns, and the `defer` deletes the claim at once (`BlockValidation.go:641`) — the claim is free for the entire backoff window.

This is reachable on the normal path, not just under a store outage: `ValidateBlock` re-triggers the parent on every child whose parent isn't mined yet — `u.setMinedChan <- block.Header.HashPrevBlock` (`BlockValidation.go:1435`). So while a parent's `setTxMinedStatus` is in backoff:

1. a child arrives, re-triggers the parent hash;
2. `MinedSet` still false → `tryClaimBlockForSetMined` now **succeeds** (claim released), so it runs `setTxMinedStatus` again;
3. it fails again and increments `setMinedRetries` (`:668-670`) — on top of the already-pending scheduled retry.

The counter advances faster than the 10 intended backoff steps, so the block hits `setMinedMaxRetries` and is dropped as `manual_intervention_required` (`:675`) prematurely. Not permanent data loss — the counter is deleted on drop (`:677`) and the next trigger restarts it — but it defeats the bounded-retry accounting this PR is adding: spurious operator alarms, an inflated drop metric, and duplicate `setTxMinedStatus` work.

Fix: keep the block claimed (or mark a "retry pending" state) across the backoff so re-triggers during the window are skipped as before, and only release/re-process when the scheduled retry actually re-runs.

### Nits (non-blocking)
- `revalidateBlockChan` re-queue (`BlockValidation.go:733`) still uses the exact untracked, non-ctx-guarded `go func(){ ch <- }()` pattern this PR replaces for setMinedChan — worth generalizing into a shared cancellable-requeue helper rather than fixing one channel.
- `scheduleSetMinedRetry` spawns one backoff-lived goroutine per failing block; a burst of distinct failing blocks multiplies live goroutines vs a single `time.AfterFunc`/delay queue. Fine at current scale.
- The `require.Less(time.Since(start), 20ms)` timing assertion in the test is a likely flake under CI load / `-race`.

icellan:
## Re-review — setMined retry rework

Checked both prior findings against the current code (not the commit message). **Both are genuinely fixed:**

- **Wait() no longer hangs** — `Wait(ctx)` now selects on `ctx.Done()` (`BlockValidation.go:2629`), so the stop deadline is honoured instead of blocking forever. 👍
- **Retry-counter inflation is gone** — `setMinedRetryPending` is set before the backoff (`BlockValidation.go:921`) and the worker skips a hash that's already retry-pending (`:637`), restoring the dedup the old inline `time.Sleep` gave implicitly. 👍

Two things the rework introduced, worth a look before merge:

### P2 — child parent-wait patience dropped in the block-found path

Releasing `blockHashesCurrentlyValidated` immediately (defer at `:651`) while the parent's retry waits in `setMinedRetryPending` means `checkParentProcessingComplete` (`Server.go:1530`) no longer covers the backoff window — it only polls `blockHashesCurrentlyValidated`, not the pending map. So a child now falls straight through to the **bounded** `waitForPreviousBlocksToBeProcessed` (`:1561`), whose budget is ~10s by default (`IsParentMinedRetryMaxRetry=45`, 20ms base, 4× — ~20s with the second round + re-trigger).

A parent whose cumulative setMined backoff outlasts that (worst case ~111s: 10 retries capped at 16s) makes the child hit "given up waiting on previous blocks to be ready" (`:1567`), which `processBlockFound` returns as a `ServiceError` (`Server.go:1499`). Not consensus — the block isn't marked invalid, it's re-processed later — but it's block-found churn/re-fetch that the old unbounded `checkParentProcessingComplete` wait absorbed. Consider having `checkParentProcessingComplete` also treat `setMinedRetryPending` as "parent still finalizing", or otherwise keep the child blocked across the parent's retry window.

### P2/P3 — WaitGroup Add races Wait during startup

`u.backgroundTasks.Add(1)` for the worker (`:592`) runs *inside* `go bv.start(ctx)` (`:507`), and only after the startup-recovery `g.Wait()` (`:551`) — a window that can be long with a large mined-not-set backlog. A `Server.Stop()→Wait(ctx)` in that window sees a zero counter, returns without waiting for the worker, and (Add starting from zero concurrent with Wait) risks a `WaitGroup misuse` panic. The comment at `:583-591` ("synchronously, before the goroutine launches") is misleading: it's synchronous w.r.t. the inner worker goroutine at `:594`, not w.r.t. `start()` itself, which is async. Either do the `Add(1)` in `NewBlockValidation` before `go bv.start`, or document the race as tolerated because the worker's ctx cancellation is the real stop signal.

### Nits (non-blocking)
- `Wait(ctx)` leaks the inner `backgroundTasks.Wait()` goroutine + `done` channel if ctx fires before drain (`:2632`), on shutdown paths that never cancel the Init ctx; repeated Stop calls stack them.
- Same paths: `Wait(ctx)` now blocks for the full stop budget (was prompt before the worker was tracked). The comment at `:2621-2628` documents this as the deliberate tradeoff — fine, just noting it's a new shutdown-latency cost.

icellan (approved):
Re-reviewed after `cc22f1e3c`. Both prior P2s verified fixed against the actual code (not the commit message), package builds, and the setMined worker/retry tests — including the new `TestIsParentFinalizing` — pass under `-race`.

- **Parent-wait** — `checkParentProcessingComplete` now waits on `isParentFinalizing`, which consults both `blockHashesCurrentlyValidated` and `setMinedRetryPending`, so a child keeps waiting across the parent's whole retry window. Failure→backoff transition has no "neither-map" gap (pending is stored before the claim is released), and the now-longer wait loop honours `ctx.Done()`.
- **WaitGroup startup race** — `Add(1)` moved into `NewBlockValidation` before `go start()`. Add/Done balance holds on every path: the `g.Wait()` early-return calls `Done()`, otherwise the worker's `defer Done()` covers it.

The two remaining shutdown-path costs (inner `Wait` goroutine leak, full stop-budget latency when the Init ctx isn't cancelled first) are now documented as deliberate tradeoffs with sound reasoning. Good to merge.

PR #944, opened 2026-05-26
fix(asset/centrifuge): speak bidirectional Centrifuge protocol (#938)

icellan:
## Summary

- Fixes #938: the Asset Server's `/connection/websocket` advertised `Unidirectional()=true` on a WebSocket transport, so centrifuge emitted the connect reply as a Push envelope without the command id. `centrifuge-go` / `centrifuge-js` clients couldn't complete the handshake and timed out after ~5s. This has blocked the node-validation acceptance suite (CLIENT-1, CLIENT-3) since v0.15.0-beta-2.
- The issue's suggested one-liner (just flip the boolean) would actually have **panicked** at runtime — `c.Connect()` always routes through `unidirectionalConnect(req, nil)`, and the bidirectional branch in `connectCmd` dereferences `cmd.Id` with no nil check. The handler had to be rewritten to the upstream `HandleReadFrame(c, reader)` pattern.
- Found and fixed a second latent bug while writing the test: `r.Context()` is cancelled the instant `ServeHTTP` returns (immediately after the goroutine launches). Centrifuge's `HandleCommand` checks `<-c.ctx.Done()` on every frame and aborted silently. The legacy `c.Connect()` path didn't go through `HandleCommand`, which masked it. Switched to `context.WithoutCancel(r.Context())` so the goroutine keeps request-scoped credentials (set by `authMiddleware`) but isn't killed by handler return.

## Changes

- `services/asset/centrifuge_impl/websocket.go`: `Unidirectional()` → `false`; replaced manual JSON-decode + `c.Connect()` + drain with `centrifuge.HandleReadFrame(c, reader)` loop over `conn.NextReader()`; deleted dead custom `ConnectRequest`/`SubscribeRequest` types and the no-op subs translation block; detached request-context cancellation.
- `services/asset/centrifuge_impl/centrifuge_test.go`: flipped the `Unidirectional()` assertion; added `TestWebsocketHandler_BidirectionalConnect` that exercises the real handshake end-to-end against a `centrifuge.Node` (asserts the reply carries `id=1`, a non-empty `connect.client`, and the `OnConnecting` subscription).
- `ui/dashboard/src/internal/stores/p2pStore.ts`: send `{id:1,connect:{}}` on open instead of `{}`; parse `push.pub.data` instead of unwrapped `pub.data`; ignore empty centrifuge ping replies; drop dead unwrapped-`json.type` fallback.
- `docs/topics/services/assetServer.md`: documented the bidirectional command-reply and push wire-format envelopes.

[...]

PR #1134, opened 2026-06-23
fix(asset/centrifuge): bidirectional protocol — negotiate protobuf + dashboard pong

icellan:
## Problem

Asset-service WebSocket clients are disconnected with `bad request`, and the dashboard's live updates churn:

```
[Centrifuge] disconnect after handling command: reason:bad request
```

## Root cause

The bidirectional Centrifuge handler (added in #944) has two protocol defects:

1. **Protocol forced to JSON.** `websocket.go` hardcoded `ProtocolTypeJSON` and the upgrader did not advertise the `centrifuge-protobuf` subprotocol. A client negotiating protobuf was silently served JSON; its binary frames decoded to a zero-value `Command`, which centrifuge treats as an unsolicited pong (`isPong` + `lastPing <= 0`) and rejects with `DisconnectBadRequest` — exactly the logged line, with an empty `command`.

2. **Dashboard never pongs.** The transport sets `PingPongConfig` (25s ping / ~8s `PongTimeout`) on a bidirectional connection, but the dashboard's raw-WebSocket store ignored the server's empty-object ping. With no pong, centrifuge disconnects with `DisconnectNoPong` every ~33s, dropping live updates.

## Fix

- **Server** (`services/asset/centrifuge_impl/websocket.go`): advertise `centrifuge-protobuf` and derive `protoType` from the negotiated subprotocol (matching upstream centrifuge). No-subprotocol clients (the dashboard) still use JSON, so this is a no-op for them.
- **Dashboard** (`ui/dashboard/src/internal/stores/p2pStore.ts`): reply to each server ping with an empty command (pong). Only ever sent in response to a ping, so it can never be an "unnecessary pong".

## Tests

- Added `TestWebsocketHandler_SubprotocolNegotiation`: a protobuf-requesting client negotiates `centrifuge-protobuf`; a no-subprotocol client stays JSON. Verified red→green (without the upgrader advertisement the protobuf case fails).
- `go vet` + existing centrifuge tests pass.

## Notes

- Branched off `main` (the defects are in `main`'s #944 handler, unrelated to any feature branch).
- Dashboard lint/typecheck not run locally (fresh worktree, no `node_modules`); the change is a single call on an `any`-typed socket — CI covers the TS toolchain.

### DIFF
--- a/services/asset/centrifuge_impl/centrifuge_test.go
+++ b/services/asset/centrifuge_impl/centrifuge_test.go
@@ -402,6 +402,46 @@ func TestNewWebsocketHandler(t *testing.T) {
 	})
 }
 
+// TestWebsocketHandler_SubprotocolNegotiation verifies the handler advertises
+// the centrifuge-protobuf subprotocol so a protobuf client negotiates it (rather
+// than being silently served JSON and disconnected with "bad request"), while a
+// client that requests no subprotocol still gets JSON (empty negotiated value).
+func TestWebsocketHandler_SubprotocolNegotiation(t *testing.T) {
+	node, err := centrifuge.New(centrifuge.Config{})
+	require.NoError(t, err)
+
+	node.OnConnecting(func(_ context.Context, _ centrifuge.ConnectEvent) (centrifuge.ConnectReply, error) {
+		return centrifuge.ConnectReply{}, nil
+	})
+
+	require.NoError(t, node.Run())
+	defer func() { _ = node.Shutdown(context.Background()) }()
+
+	// Permissive origin check: this test exercises subprotocol negotiation, not
+	// the same-host origin guard (which would otherwise reject the test dialer).
+	srv := httptest.NewServer(NewWebsocketHandler(node, WebsocketConfig{
+		CheckOrigin: func(*http.Request) bool { return true },
+	}))
+	defer srv.Close()
+
+	wsURL := "ws" + strings.TrimPrefix(srv.URL, "http")
+
+	t.Run("protobuf requested is negotiated", func(t *testing.T) {
+		d := websocket.Dialer{Subprotocols: []string{"centrifuge-protobuf"}}
+		conn, _, derr := d.Dial(wsURL, nil)
+		require.NoError(t, derr)
+		defer conn.Close()
+		require.Equal(t, "centrifuge-protobuf", conn.Subprotocol())
+	})
+
+	t.Run("no subprotocol stays JSON", func(t *testing.T) {
+		conn, _, derr := websocket.DefaultDialer.Dial(wsURL, nil)
+		require.NoError(t, derr)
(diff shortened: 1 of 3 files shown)


ordishs (approved):
Clean, targeted fix for the two bidirectional Centrifuge protocol defects.

Verified:
- The only in-repo client (services/asset/centrifuge_impl/client/client.go) uses NewJsonClient — requests no subprotocol, stays JSON, unaffected by the protobuf advertisement. No regression.
- protoType derivation is symmetric with writeData's Protocol() check (Text vs Binary), so reads and writes stay in the same codec.
- The dashboard's empty-object branch only fires on server pings (real replies carry connect/push), so the {} pong can never be unsolicited (lastPing > 0).

Minor, non-blocking follow-ups:
- The new test asserts subprotocol negotiation at the HTTP-upgrade layer only; it doesn't round-trip an actual protobuf command, so the protoType path isn't exercised end-to-end. Root cause is covered, so this is a coverage gap, not a defect.
- ui/dashboard/src/routes/wstest/+page.svelte (out of scope) has the same missing-pong issue and will still churn — worth a follow-up if anyone relies on /wstest.

LGTM.

oskarszoon (approved):
Correct, minimal fix for both bidirectional-protocol defects in the #944 handler.

- Server: `protoType` derived from negotiated subprotocol matches upstream centrifuge. Dashboard (no subprotocol) stays JSON — no behavior change.
- Dashboard: empty-object pong is only sent in reply to a server ping (connect reply has `.connect`, pubs have `.push`), so it can't be an unsolicited pong; no echo loop.

Verified locally: `go build` + `go vet` clean, new TestWebsocketHandler_SubprotocolNegotiation passes.

freemans13 (approved):
Approving — clean, well-scoped fix for the two bidirectional-protocol defects.

Verified the core mechanics against the pinned library source:
- `isPong` is `cmd.Id == 0 && cmd.Send == nil` (client.go:1100); the dashboard's `'{}'` pong decodes to exactly that and, with the server now issuing pings (`lastPing > 0`), is accepted rather than rejected as an unnecessary pong (client.go:1176).
- The `conn.Subprotocol() == "centrifuge-protobuf"` derivation matches upstream (handler_websocket.go:173-174), and `writeData` already keys message type off `Protocol()`.
- The dashboard sends no subprotocol, so the server change is a genuine no-op for it.

One minor (non-blocking) note: the upgrader advertises only `centrifuge-protobuf`, whereas upstream advertises both `centrifuge-json` and `centrifuge-protobuf`, so the "Matches the upstream centrifuge WebsocketHandler" comment in websocket.go is slightly inaccurate. No real client breaks from the omission (dashboard and centrifuge-go JSON mode send no subprotocol), but adding `centrifuge-json` would restore exact parity. Safe to merge as-is.

PR #1137, opened 2026-06-24
test(blockassembly): BA-STARTUP-008

sugh01:


ordishs (approved):
Reviewed. Clean test-only PR adding BA-STARTUP coverage (CheckInputValidation read-only scan, unmined-loading guards, validateUnminedTxInputs conflict + cascade).

Verified locally:
- All four test functions pass (`ok ... 0.755s`).
- Mock argument matchers and call sequences match the real validateUnminedTxInputs / CheckInputValidation / gRPC handlers.
- Read-only guarantee is genuinely asserted (no SetConflicting/Remove expectations + dryRun=true).
- Case A verifies the BFS cascade via AssertNumberOfCalls("Remove", 3).
- gci import ordering clean.

Minor (non-blocking): the unminedDropHashes stashing in markAsConflicting is left unverified since both BA-STARTUP-007 cases construct BlockAssembler with a nil map — worth a follow-up case if full coverage is desired.

gokutheengineer (approved):

oskarszoon (approved):
Good coverage of the BA-STARTUP gaps — the dryRun no-mutation guarantee (`AssertNotCalled` on `SetConflicting`/`Remove`) and the Case A/B `validateUnminedTxInputs` branches are the valuable bits, and they pass against the real impl rather than faking behavior. Approving.

One nit, non-blocking: repo convention (`.claude/rules/testing.md`) is `require`, not `assert`. The count/bool checks use `assert.Equal`/`assert.False` (lines 104, 141, 190, 232, 318, 364). Mostly terminal so behavior is identical, but in Case A the `assert.False` (318) is followed by `AssertExpectations`/`AssertNumberOfCalls` — on failure the test keeps running with noisier output. Worth switching to `require` for consistency.

PR #1138, opened 2026-06-24
fix(blockvalidation): re-open store reader on mmap fallback in readSubtree to avoid corrupt subtree

gokutheengineer:
## Summary

In `readSubtree` (`services/blockvalidation/quick_validate.go`), the mmap-deserialization fallback reset the buffered reader onto an **already-consumed** `io.ReadCloser` and retried. The underlying store e retry read from

## Summary

In `readSubtree` (`services/blockvalidation/quick_validate.go`),
the mmap-deserialffered readeronto an **already-consumed** `io.ReadCloser` and retried. The underlying store reader is not seekable, so the retry read from mid/end of the stream and produced a silently **corrupt subtree** (or a misleading deserialization error), which was then used for merkle/coinbase validation.

Fixes #4656.

## Root cause

```go


On fallback, close the consumed reader's role and open a **fresh** reader from the store (the same pattern already used a few lines below for the subtree-data reader), then deserialize on the heap from the start:

```go
if err != nil {
    u.logger.Warnf(...)
    fallbackReader, ferr := u.subtreeStore.GetIoReader(ctx, subtreeHash[:], localFileType)
    if ferr != nil {                                              return sutFoundError(...)}
    }
    defer fallbackReader.Close()                              bufferedReade
    subtree, err = subtreepkg.NewSubtreeFromReader(bufferedReader)
}
```

A separate `fallbackReader` variable is used (rather than reassigning `subteader's existing
`defer Close()` s handle — nodouble-close. The extra `GetIoReader` happens only on the rare fallback path, so there is no hot-path cost.

## Testing

- Added `TestReadSubtree_MmapFallbackReReadsFromStart` (`services/blockvalidation/quick_validate_mmap_test.go`): stores a valid subtree + subtree data in the memory blob store, points `mmapDir` at a non-existent directory to force the mmap allocator to fail, and asserts the fallback still produces a correct, **heap-backed** subtree (matching root hash, length, coinbase nilled). Before the fix this failed with `unable to read root hash: EOF`.
- `go build ./services/blockvalidation/` and `go vet ./services/blockvalidation/` pass.

> Note: the unit-locally in thisenvironment due to the known GoBDK cgo link issue; the test runs in CI. The fix mechanism was additionally validated end-to-end against the real `go-subtree` package in an isolated cgo-free
reproduction (mmaed path errors
`EOF`; fresh-readbtree).

## Risk

- Behavior changes only on the mmap-failure fallback path; the non-mmap and mmap-success paths are untouched.
- Data-integrity improvement: removes a path that could feed a corrupt `*Subtree` into merkle/coinbase validation.


services/blockvalidation/quick_validate.go
```diff
@@ -782,9 +782,19 @@ func (u *BlockValidation) readSubtree(ctx context.Context, block *model.Block, s
...
+			// reader onto it would read from mid-stream and produce a corrupt subtree.
+			// Open a fresh reader from the store so the heap path reads from the start.
 			u.logger.Warnf("[getBlockTransactions][%s] mmap deserialization failed for subtree %s, falling back to heap: %v", block.Hash().String(), subtreeHash.String(), err)
-			bufferedReader.Reset(subtreeReader)
+
+			fallbackReader, ferr := u.subtreeStore.GetIoReader(ctx, subtreeHash[:], localFileType)
+			if ferr != nil {
+				return subtreeResult{err: errors.NewNotFoundError("[getBlockTransactions][%s] failed to re-open subtree %s for heap fallback", block.Hash().String(), subtreeHash.String(), ferr)}
+			}
+			defer fallbackReader.Close()
```
freemans13, line 804: Minor (non-blocking): on this fallback path the original `subtreeReader` opened at line 766 stays open alongside `fallbackReader` until the function returns (its `defer Close()` fires at return, not at block exit). It's harmless — the rare failure path, and both readers close cleanly — but if you wanted to be tidy you could `subtreeReader.Close()` before opening the fresh reader, since it's no longer used once the mmap attempt has consumed it.

services/blockvalidation/quick_validate_mmap_test.go
```diff
@@ -0,0 +1,97 @@
...
+	subtreeDataBytes, err := subtreeData.Serialize()
+	require.NoError(t, err)
+	require.NoError(t, subtreeStore.Set(ctx, subtree.RootHash()[:], fileformat.FileTypeSubtreeData, subtreeDataBytes))
+
+	bv := NewBlockValidation(ctx, ulogger.TestLogger{}, tSettings, blockchainClient, subtreeStore, txStore, utxoStore, nil, subtreeValidationClient)
+
+	// Force the mmap path: a non-empty mmapDir enables it, and pointing it at a
+	// non-existent directory makes the mmap allocator's os.CreateTemp fail, which
+	// triggers the heap fallback after the underlying stream was already consumed.
+	bv.mmapDir = t.TempDir() + "/does-not-exist"
```
freemans13: Nit (cosmetic): `t.TempDir() + "/does-not-exist"` hard-codes the path separator. `filepath.Join(t.TempDir(), "does-not-exist")` reads a touch cleaner and is platform-agnostic. Functionally fine as-is.

icellan (approved):

freemans13 (approved):
Approve — correct, minimal, well-tested fix.

Verified the bug and fix against the pinned `go-subtree@v1.4.2` source rather than taking the description on trust:

- `deserializeFromReaderMmap` reads the 56-byte header **and** wraps the reader in its own 32KB `bufio` (reading far ahead) *before* the mmap allocation that fails on a bad `mmapDir`. So the underlying non-seekable store reader really is drained on the fallback path, and the old `bufferedReader.Reset(subtreeReader)` retry genuinely read from mid-stream → the `unable to read root hash: EOF` / corrupt subtree. Re-opening a fresh reader is the right remedy.
- No double-close: fallback closes `subtreeReader`, nils it (nil-guarded defer skips the second close), and `fallbackReader` has its own defer.
- `bufferedReader.Reset(fallbackReader)` discards the stale bytes the mmap attempt pulled through the pooled reader — reads restart from byte 0.
- Scope is complete: the other mmap site (`subtreeFromBytesWithMmap`) re-reads the original `[]byte`, so it's unaffected — no sibling instance of this bug left unfixed.
- Regression test forces the fallback reliably (mmapDir → non-existent subdir of t.TempDir()) and asserts a heap-backed subtree with the correct root/length.

Two optional nits, neither blocking:
1. The re-open failure returns `NewNotFoundError`; a file that opened microseconds earlier failing to re-open reads more like a transient `StorageError`. Consistent with the existing first-open handler though, so a judgment call.
2. The two adjacent comments both explain that the mmap attempt consumed `subtreeReader` — one is enough.

PR #1167, opened 2026-06-25
fix(blockassembly): process all hashes in removeTxsFromSubtrees

### LINKED ISSUE #1156: [TN-017][high] blockassembly: removeTxsFromSubtrees early-return leaves invalid coinbase-spend children in assembly
**Severity:** high · **Confidence:** high · **Pass:** UTXO integrity / consensus (self-produced block)
**Component:** `services/blockassembly/subtreeprocessor/SubtreeProcessor.go`

### Summary
`removeTxsFromSubtrees(hashes)` iterates a slice, but when a hash is found in the **current** subtree it `return`s out of the whole function — skipping every remaining hash AND the trailing `reChainSubtrees(0)` compaction. It is called from `removeCoinbaseUtxos`, which has already deleted those txs' parent (coinbase) UTXOs, so left-behind children remain in assembly while spending non-existent UTXOs.

### Evidence
- `SubtreeProcessor.go:2699-2750`: loop `for _, hash := range hashes`; `:2723-2726` `if foundSubtreeIndex == -1 { return …RemoveNodeAtIndex(foundIndex) }`; chained-subtree branch `:2735-2739` and final `reChainSubtrees(0)` `:2745` bypassed.
- `removeCoinbaseUtxos` `:4018-4052`: `GetAndLockChildren` → per-child `utxoStore.Delete` (`:4036`) + `removeMap.Put` (`:4042`) → then `removeTxsFromSubtrees` (`:4049`).
- Reorg/reset callers: `:1352`, `:1408`, `:3860`.

### Why it matters
On a reorg/reset orphaning a block whose matured coinbase outputs were spent, with ≥2 such child-spends where the first is in the current subtree, the remaining children stay in the mining candidate with deleted input UTXOs → this node builds/mines a **consensus-invalid block**. The skipped `reChainSubtrees(0)` also leaves a hole whenever a chained removal precedes a current-subtree removal. The one-line nature (`return` vs `continue`) indicates an unintentional bug.

### Suggested fix
Change the `return` at `:2726` to `continue`; ensure `reChainSubtrees(0)` always runs after the loop.

### Suggested tests
Assembly with N coinbase-spend children where children[0] is in the current subtree and children[1..] span a chained subtree; `removeTxsFromSubtrees(all)` → assert all removed from both and the chain recompacted. (Existing tests `SubtreeProcessor_test.go:3170-3337` miss this ordering case.)

---

ordishs:
## Summary

Fixes #1156.

`removeTxsFromSubtrees(hashes)` iterated the slice but, when a hash was found in the **current** subtree, it `return`ed out of the whole function — skipping every remaining hash *and* the trailing `reChainSubtrees(0)` compaction.

It is called from `removeCoinbaseUtxos`, which has already deleted those txs' parent (coinbase) UTXOs. So on a reorg/reset that orphans a block whose matured coinbase outputs were spent, with ≥2 such child-spends where the first sits in the current subtree, the remaining children stayed in the mining candidate while spending non-existent UTXOs → the node could build/mine a **consensus-invalid block**. The skipped `reChainSubtrees(0)` also left a hole whenever a chained removal preceded a current-subtree removal.

## Fix

In the current-subtree branch, remove the node (with an error check) and `continue` the loop instead of `return`ing. The loop then falls through to the existing `reChainSubtrees(0)`, which now always runs and recompacts every subtree.

## Tests

Added `TestRemoveTxsFromSubtreesOrdering` with both hash orderings the existing tests missed:
- current-subtree hash before chained-subtree hash (the primary bug: later hash never removed)
- chained-subtree hash before current-subtree hash (the skipped-recompaction hole)

Each asserts all targeted hashes are gone from every subtree and `currentTxMap`, untouched hashes survive exactly once, and every chained subtree is `IsComplete()` (no hole).

Verified red against the old code, green after the fix.

```
go test -race -tags testtxmetacache -run TestRemoveTxsFromSubtreesOrdering ./services/blockassembly/subtreeprocessor/   # ok
go test -race -tags testtxmetacache ./services/blockassembly/subtreeprocessor/                                          # ok  157.702s
go vet -tags testtxmetacache ./services/blockassembly/subtreeprocessor/                                                 # exit 0
```

### DIFF
--- a/services/blockassembly/subtreeprocessor/SubtreeProcessor.go
+++ b/services/blockassembly/subtreeprocessor/SubtreeProcessor.go
@@ -2721,9 +2721,14 @@ func (stp *SubtreeProcessor) removeTxsFromSubtrees(ctx context.Context, hashes [
 
 			// we found the transaction in a subtree
 			if foundSubtreeIndex == -1 {
-				// it was found in the current tree, remove it from there
-				// further processing is not needed, as the subtrees in the chainedSubtrees are older than the current subtree
-				return stp.currentSubtree.Load().RemoveNodeAtIndex(foundIndex)
+				// it was found in the current tree, remove it from there and continue
+				// with the remaining hashes. The trailing reChainSubtrees(0) compacts
+				// every subtree, including the current one, after the loop completes.
+				if err := stp.currentSubtree.Load().RemoveNodeAtIndex(foundIndex); err != nil {
+					return errors.NewProcessingError("[SubtreeProcessor][removeTxsFromSubtrees][%s] error removing node from current subtree", hash.String(), err)
+				}
+
+				continue
 			}
 
 			// it was found in a chained subtree, remove it from there and chain the subtrees again from the point it was removed
--- a/services/blockassembly/subtreeprocessor/SubtreeProcessor.go
+++ b/services/blockassembly/subtreeprocessor/SubtreeProcessor.go
@@ -2649,9 +2649,19 @@ func (stp *SubtreeProcessor) removeTxFromSubtrees(ctx context.Context, hash chai
 
 		// we found the transaction in a subtree
 		if foundSubtreeIndex == -1 {
-			// it was found in the current tree, remove it from there
-			// further processing is not needed, as the subtrees in the chainedSubtrees are older than the current subtree
-			return stp.currentSubtree.Load().RemoveNodeAtIndex(foundIndex)
+			// it was found in the current tree. Duplicate before mutating (mirroring the
+			// chained-subtree branch below) so any precomputed mining-data snapshot holding
+			// the original stays safe for concurrent reads. Further processing is not needed,
+			// as the subtrees in chainedSubtrees are older than the current subtree.
+			currentSubtree := stp.currentSubtree.Load().Duplicate()
+
+			if err := currentSubtree.RemoveNodeAtIndex(foundIndex); err != nil {
+				return errors.NewProcessingError("[SubtreeProcessor][removeTxFromSubtrees][%s] error removing node from current subtree", hash.String(), err)
+			}
+
+			stp.currentSubtree.Store(currentSubtree)
+
+			return nil
 		}
 
 		// it was found in a chained subtree, remove it from there and chain the subtrees again from the point it was removed
@@ -2696,6 +2706,8 @@ func (stp *SubtreeProcessor) removeTxsFromSubtrees(ctx context.Context, hashes [
 
 	defer deferFn()
 
+	removedAny := false
+
 	for _, hash := range hashes {
 		// find the transaction in the current and all chained subtrees
 		foundIndex := stp.currentSubtree.Load().NodeIndex(hash)
@@ -2713,6 +2725,8 @@ func (stp *SubtreeProcessor) removeTxsFromSubtrees(ctx context.Context, hashes [
 		}
 
 		if foundIndex >= 0 {
+			removedAny = true
+
 			// Save to deleted backup map before removing (for Server fallback during async storage)
 			if txInpoints, found := stp.currentTxMap.Get(hash); found {
 				stp.deletedTxs.Set(hash, *txInpoints)
@@ -2752,6 +2766,12 @@ func (stp *SubtreeProcessor) removeTxsFromSubtrees(ctx context.Context, hashes [
 				return errors.NewProcessingError("[SubtreeProcessor][removeTxsFromSubtrees][%s] error removing node from subtree", hash.String(), err)
 			}
 		}
+	}
+
+	// Nothing matched, so there is no hole to fill: skip the O(N) rebuild (which would
+	// also needlessly replace every subtree pointer and invalidate mining snapshots).
+	if !removedAny {
+		return nil
 	}
 
 	// all the chained subtrees should be complete, as we now have a hole in the one we just removed from
@@ -2805,8 +2825,13 @@ func (stp *SubtreeProcessor) reChainSubtrees(fromIndex int) error {
 	}
 	stp.currentSubtree.Store(newSubtree)
 
-	if len(originalSubtrees) == 0 {
-		// we must add the coinbase tx if we have no original subtrees
+	if fromIndex == 0 {
+		// Rebuilding from index 0 means the block's first subtree (which carries the
+		// coinbase placeholder at Nodes[0]) is part of the rebuild. The re-add loop below
+		// skips placeholder nodes, so re-assert the placeholder on the fresh first subtree;
+		// otherwise the first real transaction would slide into Nodes[0] and corrupt the
+		// coinbase commitment. For fromIndex > 0 the preserved chainedSubtrees[:fromIndex]
+		// still hold the placeholder, so the rebuilt tail must not add another one.
 		if err = stp.currentSubtree.Load().AddCoinbaseNode(); err != nil {
 			return errors.NewProcessingError("error adding coinbase node to new current subtree", err)
 		}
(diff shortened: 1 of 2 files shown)


icellan (approved):

oskarszoon (changes requested):
Re-reviewed at bd9e45b (unchanged since the last review). The "process all hashes, don't early-return" intent and the `Duplicate()`-before-mutate fix (commit 2) are correct, but a P0 still blocks merge.

**P0 — `reChainSubtrees(0)` drops the coinbase placeholder, and the new `continue` now routes the current-subtree removal into it** (`SubtreeProcessor.go:2740` → `:2759`). Replacing the early `return` with `continue` means every current-subtree removal now falls through to the unconditional `reChainSubtrees(0)`, which drops the coinbase placeholder from `subtree[0].Nodes[0]`:
- it appends the current subtree to `originalSubtrees` (`:2778`), so the `len(originalSubtrees)==0 → AddCoinbaseNode` branch (`:2808-2813`) is dead;
- the re-add loop `continue`s past any `CoinbasePlaceholder` node (`:2825-2827`);
- it stores a fresh **non-nil** empty current subtree (`:2806`), so `addNode`'s nil-guard re-establish (`:2095`) never fires.

The first real tx lands at `Nodes[0]`. This is not a locally-rejected block — it's worse: on submit `ReplaceRootNode` (go-subtree `subtree.go:279`) blindly overwrites `Nodes[0]`, destroying the regular tx that slid in, and the validator's `CheckMerkleRoot` (`Block.go:1413`) applies the **same** substitution, so peers **accept** the block with a silently-dropped transaction (self-consistent but wrong commitment), plus coinbase value drift from the `txCount--` accounting. Reachable on a reorg/reset orphaning a block whose spent coinbase has ≥1 child-spend in the current subtree (the #1156 scenario), via `removeCoinbaseUtxos → removeTxsFromSubtrees` from both `Reset` (`:1352`) and `moveBackBlock` (`:3874`). `TestRemoveTxsFromSubtreesOrdering` is green only because it never asserts `subtree[0].Nodes[0] == CoinbasePlaceholder`.

Fix direction: have `reChainSubtrees` re-assert the coinbase placeholder on the first subtree of the block **unconditionally** (guard on "first subtree of the block", not `len(originalSubtrees)==0`); add a test asserting `subtree[0].Nodes[0] == CoinbasePlaceholder` after removal; and harden `createIncompleteSubtreeCopy`'s `Nodes[1:]` copy so it can't silently drop a node when `Nodes[0]` isn't the placeholder.

**P1 — use-after-close of captured subtree refs in `moveBackBlock`.** `:3870-3871` capture `lastIncompleteSubtree`/`chainedSubtrees` before `removeCoinbaseUtxos` (`:3874`); `reChainSubtrees(0)` then `Close()`s those subtrees, and `moveBackBlockBulkBuild` reads their `.Nodes` at `:3987-3993`. Safe when the removal is in the current subtree (the `Duplicate()` means the live subtree isn't the one closed), but a reorg where all coinbase-child-spends are in **chained** subtrees closes the live current subtree and reads it after unmap → read-after-unmap / SIGSEGV for mmap-backed subtrees. Rebuild from live state or confirm the captured-ref lifetime.

**P1 — singular `removeTxFromSubtrees` still mutates the current subtree in place** (`:2654`, `return stp.currentSubtree.Load().RemoveNodeAtIndex(foundIndex)`). Commit 2's own rationale is that mutating the live current subtree is unsafe for a mining snapshot holding the pointer; the `Duplicate()` fix was applied to the plural function's branches but not here. Same bug class — apply the same fix.

Nits: the unconditional `reChainSubtrees(0)` on zero removals (`:2759`) runs an O(N) rebuild + coinbase drop even when no hash matched (the singular path gates on `foundIndex >= 0`); and the `NodeIndex` lazy-build (`subtree.go:561-579`) reads the map lock-free after a write-locked build (pre-existing, not introduced here).

ordishs:
@oskarszoon — addressed in 4240bb584. Each fix was driven by a regression test (red→green); full `go test -race` on `services/blockassembly/subtreeprocessor` is green and `go vet ./services/blockassembly/...` is clean.

**P0 — coinbase placeholder dropped by `reChainSubtrees(0)`.** Fixed in `reChainSubtrees`: re-assert the placeholder whenever the rebuild starts at index 0 (the block's first subtree), replacing the dead `len(originalSubtrees)==0` guard (the current subtree is always appended, so that branch never fired). For `fromIndex > 0` the preserved `chainedSubtrees[:fromIndex]` still carry the placeholder, so the rebuilt tail must not add another.

One correction on attribution, since it determined where the fix belongs: I verified empirically that the placeholder drop is **pre-existing in `reChainSubtrees`**, not introduced by the `continue`. A probe removing a hash from a *chained* subtree — which already routed through `reChainSubtrees(0)` before this PR — also drops the placeholder. The `continue` only widens the set of triggering scenarios. So the fix sits in `reChainSubtrees` and covers every caller (including the singular path's `reChainSubtrees(foundSubtreeIndex)` when `foundSubtreeIndex == 0`), rather than a narrower patch in `removeTxsFromSubtrees` that would have left the chained-removal path broken. Test asserts `subtree[0].Nodes[0] == CoinbasePlaceholderHashValue` after removal across the ordering cases. I did not independently reproduce the downstream accept-vs-reject claim (`ReplaceRootNode`/`CheckMerkleRoot`); the drop itself is confirmed and invalid regardless.

**P1a — stale/closed subtree refs in `moveBackBlock`.** Fixed by capturing `lastIncompleteSubtree`/`chainedSubtrees` **after** `removeCoinbaseUtxos` (live, post-removal state) instead of before. New test `TestMoveBackBlockDoesNotResurrectRemovedCoinbaseSpends` drives the full path with coinbase child-spends in a chained subtree and a non-empty current subtree; pre-fix it failed by **duplicating** the current-subtree txs (the heap-observable form of the same stale-capture bug — `expected 1, actual 2`), and it also covers the resurrection case. The mmap read-after-unmap you described is the same root cause on mmap-backed subtrees.

**P1b — singular `removeTxFromSubtrees` mutated the current subtree in place.** Applied the same `Duplicate()`-before-mutate + `Store` as the plural function; `TestRemoveTxFromSubtreesCurrentSnapshotSafety` asserts the pre-removal snapshot pointer is left untouched.

**Nit — unconditional `reChainSubtrees(0)` on zero matches.** Gated on a `removedAny` flag, matching the singular path's `foundIndex >= 0` gate; `TestRemoveTxsFromSubtreesNoMatchSkipsRechain` asserts no rebuild (no pointer churn) when nothing matched.

I did not touch `createIncompleteSubtreeCopy`'s `Nodes[1:]` copy — the P1a capture-after-removal change removes the path that fed it stale data. Happy to add defence-in-depth there if you'd prefer. The `NodeIndex` lazy-build lock observation I left as pre-existing/out-of-scope.

oskarszoon (approved):
Re-reviewed at 4240bb5. All three blockers from the previous review are fixed and verified against the new code:

- **P0 (coinbase placeholder)** — `reChainSubtrees` now re-asserts the placeholder whenever the rebuild starts at index 0, before the re-add loop, so `Nodes[0]` is the placeholder when the first real tx lands. Confirmed exactly one placeholder across the assembly and no double-add (the re-add path rejects the placeholder hash, and `AddCoinbaseNode` can't fire twice on a fresh subtree). Merkle/coinbase commitment is correct again.
- **P1 (moveBackBlock use-after-close)** — capturing the previous subtree state after `removeCoinbaseUtxos` folds the live rebuilt subtrees, eliminating the read-after-unmap and the resurrected child-spends.
- **P1 (singular `removeTxFromSubtrees`)** — now `Duplicate()`s before removal, matching the plural path.

The `removedAny` gate also resolves the unconditional-rebuild P2, and the new tests assert the coinbase-placeholder invariant. Suite green under `-race`.

Approving. Two unrelated, pre-existing items worth a separate follow-up (not blockers for this PR): the `createIncompleteSubtreeCopy` call at `:919` lacks the `chainedSubtreeCount == 0` guard the other two call sites have (it can run the `Nodes[1:]` copy on a non-first subtree), and the go-subtree `NodeIndex` lazy-build is a lock-free read of a write-lock-built map.

PR #1169, opened 2026-06-25
fix(txmetacache): never cache or trust partial inpoints-less tx meta

### LINKED ISSUE #1168: Subtree validation permanently wedges a block when txmeta cache holds a partial (inpoints-less) entry
## Summary

A node can **permanently wedge on block validation** when the txmeta cache holds a *partial* entry for a transaction — one cached with empty `TxInpoints` (no parent tx hashes). Subtree validation trusts the cache hit, never falls back to the store (which has the correct data), and `Meta.Serialize()` rejects the inpoints-less node. The block fails validation, retries forever, and the node stops advancing.

Observed during scaling tests: one node stuck at height N while peers advanced, retrying the same block ~once/second with an identical error. Not infra-related (no storage/OSD involvement) — a pure cache-correctness bug.

## Symptom

```
failed block validation BlockFound [Block <hash> (height: 304, txCount: 60817408, ...)]
 -> [CheckBlockSubtreesRequest] Failed to validate subtree <subtree-hash>
 -> [ValidateSubtreeInternal][<subtree-hash>] failed to serialize subtree meta
 -> cannot serialize, parent tx hashes are not set for node 12842: <txid>:
    cannot serialize, subtree is not set
```

The affected tx’s record in the UTXO store **does** contain its `inputs` (parents derivable) — so the data is present; only the *cached* metadata is incomplete. Deterministic and node-local (only the node whose cache holds the partial entry is affected), which is why it hit a single tx on a single node.

## Root cause

`TxMetaCache.BatchDecorate` caches **every** decorated unmined/non-conflicting result via `meta.Data.MetaBytes()`, **regardless of which fields the caller requested**. When a caller requests a reduced field set that excludes `TxInpoints`, the store does not populate it (`addAbstractedBins` only computes `TxInpoints` when explicitly requested), so `MetaBytes()` serializes an **empty** `TxInpoints` and that partial meta is written to the cache as if it were complete.

The poisoning caller lives in the subtree-validation process itself, so it pollutes the very cache it later reads:

- `prefetchParentBaseFields = {fields.BlockIDs, fields.BlockHeights}` — `services/subtreevalidation/check_block_subtrees.go:1263` (no `TxInpoints`/`Inputs`).
- `prefetchLevelParents` bulk-reads block parents only for mined-status via `u.utxoStore.BatchDecorate(gCtx, items, prefetchFields...)` — `services/subtreevalidation/check_block_subtrees.go:1335`.
- `u.utxoStore` is the `TxMetaCache`; its wrapper caches each result — `stores/txmetacache/txmetacache.go:509-559` (gated only on `BlockIDs`/`Conflicting`, never on whether `TxInpoints` was requested).
- `meta.Data.MetaBytes()` always serializes `TxInpoints` — `stores/utxo/meta/data.go:309,339` — so the round-trip faithfully stores the *empty* value.

Consumer side then trusts it:

- `processTxMetaUsingCache` marks any cache hit `isSet: true` and copies `txInpoints` verbatim with no completeness check, and a hit short-circuits the store fallback — `services/subtreevalidation/processTxMetaUsingCache.go:115-128`.
- `ValidateSubtreeInternal` sets the empty inpoints into the subtree Meta (`services/subtreevalidation/SubtreeValidation.go:826`), then `Meta.Serialize()` (go-subtree `subtree_meta.go`) rejects any non-coinbase node with nil `ParentTxHashes`.

The store *can* reconstruct `TxInpoints` from the `inputs` bin (`stores/utxo/aerospike/get.go:770` → `processInputsToTxInpoints` → `subtree.NewTxInpointsFromInputs`), so falling through to the store would resolve it — but the cache hit prevents that.

### Why it’s rare / timing-dependent

The entry must be (1) written partial via the prefetch path, (2) not overwritten by a full-field decorate, and (3) still resident (not evicted) when a subtree-serialize reads it. That intersection is uncommon, which is why it surfaces as a single tx on a single node rather than broadly.

## Proposed fix (two parts)

1. **Producer (root):** `TxMetaCache.BatchDecorate` must not cache a partial meta. Only populate the cache when the request includes the fields the cache contract promises — specifically `fields.TxInpoints` (or `Inputs`); otherwise skip caching. A cache entry must always be a complete, serialization-valid meta. (`prefetchParentBaseFields` requesting a reduced set is legitimate for its mined-status purpose — the bug is the wrapper caching that partial result.)
2. **Consumer (defensive):** `processTxMetaUsingCache` should treat a cache hit with empty `ParentTxHashes` on a non-coinbase tx as a **miss**, falling through to the store so any partial entry that slips through self-heals instead of wedging a block.

## Acceptance criteria

- A reduced-field `BatchDecorate` (e.g. `{BlockIDs, BlockHeights}`) does **not** create a cache entry that later yields empty `TxInpoints` for a non-coinbase tx.
- A subtree containing a tx whose only cache entry was written via the reduced-field path still serializes successfully (consumer falls through to the store).
- Regression test reproducing the wedge: cache a non-coinbase tx with empty `ParentTxHashes`, validate a subtree containing it, assert it resolves from the store and `Meta.Serialize()` succeeds.

## Impact

A wedged node cannot advance past the affected block and does not self-recover; it requires intervention (cache flush / service restart to repopulate from the store). High severity for follower availability under load.

icellan:
## Problem

A node can **permanently wedge on block validation** when the txmeta cache holds a *partial* entry for a non-coinbase transaction — one cached with empty `TxInpoints` (no parent tx hashes). Subtree validation trusts the cache hit, never falls back to the store (which has the correct data), and subtree-meta serialization rejects the inpoints-less node. The block fails validation, retries forever (~once/second with an identical error), and the node stops advancing. Observed during scaling tests; node-local and deterministic, requiring a cache flush / restart to recover.

## Root cause

`TxMetaCache.BatchDecorate` cached **every** decorated unmined/non-conflicting result, gated only on `BlockIDs`/`Conflicting` — never on whether `TxInpoints` was actually populated. A reduced-field parent prefetch (`{BlockIDs, BlockHeights}` in `check_block_subtrees.go`) produces a meta with empty `TxInpoints`; `MetaBytes()` happily serializes the empty value and the wrapper wrote that partial meta into the very cache subtree validation later reads. The consumer (`processTxMetaUsingCache`) then trusts any hit verbatim and short-circuits the store fallback.

## Fix (two parts)

1. **Producer (root):** `BatchDecorate` skips caching a non-coinbase meta with empty `ParentTxHashes`. Every non-coinbase tx has at least one parent, so an empty value here means the field was not populated — i.e. an incomplete entry that must not enter the cache. (Gating on data-completeness rather than the requested field set is more robust: only an explicit `fields.TxInpoints` request populates `Data.TxInpoints`, and this catches *any* partial-producing path.)
2. **Consumer (defensive):** `processTxMetaUsingCache` treats a cache hit with empty `ParentTxHashes` on a non-coinbase tx as a **miss**, falling through to the store — which reconstructs inpoints from the tx inputs — so any partial entry that slips through self-heals instead of wedging the block.

## Tests

- New regression tests (TDD, confirmed failing before the fix):
  - `stores/txmetacache`: a partial (inpoints-less, non-coinbase) meta is **not** cached; a complete meta **is** cached.
  - `services/subtreevalidation`: a partial cache entry is treated as a **miss**; a coinbase entry with no parents is still a **hit**.
- Updated the `populateCache` test helper to seed complete entries (it previously seeded empty-`ParentTxHashes` entries, which encoded the buggy behavior).
- `go test` (both packages), `go test -race` (new tests), and `go vet` all pass.

## Risk

`BatchDecorate` no longer caches reduced-field results lacking inpoints. The only such caller (parent prefetch) uses the returned map directly, not the cache, so its path is unaffected — it just no longer poisons the cache. `TxMetaCache.BatchDecorate` always reads through to the underlying store, so a consumer "miss" reliably re-fetches full inpoints.

Closes #1168

### DIFF
--- a/services/subtreevalidation/processTxMetaUsingCache.go
+++ b/services/subtreevalidation/processTxMetaUsingCache.go
@@ -113,7 +113,12 @@ func (p *TxMetaProcessor) processTxMetaUsingCache(i int) error {
 		}
 
 		txMeta, found := p.cache.GetMetaCached(p.ctx, txHash)
-		if found {
+		// Defensively reject a partial cache entry: a non-coinbase tx with no
+		// parent tx hashes is incomplete (TxInpoints was never populated) and
+		// would later fail subtree-meta serialization and wedge the block. Treat
+		// it as a miss so the store fallback — which can reconstruct inpoints from
+		// the tx inputs — resolves it and self-heals the cache.
+		if found && (txMeta.IsCoinbase || len(txMeta.TxInpoints.ParentTxHashes) > 0) {
 			p.txMetaSlice[i+j] = metaSliceItem{
 				fee:         txMeta.Fee,
 				sizeInBytes: txMeta.SizeInBytes,
--- a/services/subtreevalidation/processTxMetaUsingCache_test.go
+++ b/services/subtreevalidation/processTxMetaUsingCache_test.go
@@ -5,6 +5,7 @@ import (
 	"net/url"
 	"testing"
 
+	"github.com/bsv-blockchain/go-bt/v2"
 	"github.com/bsv-blockchain/go-bt/v2/chainhash"
 	"github.com/bsv-blockchain/go-subtree"
 	"github.com/bsv-blockchain/teranode/errors"
@@ -59,10 +60,19 @@ func generateTestHashes(count int) []chainhash.Hash {
 
 func populateCache(t testing.TB, cache *txmetacache.TxMetaCache, hashes []chainhash.Hash) {
 	t.Helper()
+	// Seed complete entries: a non-coinbase tx always has at least one parent, so
+	// a real cache entry carries a non-empty ParentTxHashes. (An empty one is a
+	// partial entry that the cache read-path now correctly treats as a miss.)
+	parent := chainhash.HashH([]byte("test-parent"))
+	in := &bt.Input{PreviousTxOutIndex: 0}
+	require.NoError(t, in.PreviousTxIDAdd(&parent))
+	ti, err := subtree.NewTxInpointsFromInputs([]*bt.Input{in})
+	require.NoError(t, err)
+
 	testMeta := &meta.Data{
 		Fee:         100,
 		SizeInBytes: 250,
-		TxInpoints:  subtree.TxInpoints{ParentTxHashes: []chainhash.Hash{}},
+		TxInpoints:  ti,
 		BlockIDs:    []uint32{},
 	}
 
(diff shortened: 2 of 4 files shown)


ordishs (approved):
LGTM. Fixes the deterministic block-validation wedge at the right altitude — layered producer (skip caching partial inpoints-less non-coinbase meta) + consumer (treat such a hit as a miss, self-heal via store) defense, backed by TDD regression tests.

Verified:
- Invariant holds: every non-coinbase tx has ≥1 parent hash (NewTxInpointsFromInputs emits one deduped entry per input; zero-input non-coinbase txs are rejected at consensus). No complete entry can be wrongly rejected.
- No nil-panic: GetMetaCached returns non-nil whenever found==true, and the && short-circuits safely.
- processTxMetaUsingCache is the sole cache reader and now has the defensive check; no other read path trusts partial entries.
- The reduced caching for fields-without-TxInpoints callers is the fix, not a regression — those were the poison entries, and the only reader requires inpoints.

Minor non-blocking note: the !IsCoinbase / IsCoinbase term in both checks is effectively dead today (IsCoinbase is field-gated and not requested on these paths); the real load-bearing guards are the BlockIDs gate (producer) and the placeholder filter (consumer). Correct as-is.

freemans13 (approved):
Reviewed: two-layer fix (producer skips caching partial inpoints-less non-coinbase meta in BatchDecorate; consumer treats such cache hits as misses and self-heals via the store fallback).

Verified:
- The only reduced-field BatchDecorate caller is the parent prefetch (check_block_subtrees.go:1295), which uses the decorated map directly and never relies on cache population.
- Self-heal confirmed: missed txs route to processTxMetaUsingStore, which re-decorates with TxMetaFieldsForDecorate (includes fields.TxInpoints), overwriting the poison entry.
- Consumer condition is nil-safe (short-circuits on !found); coinbase carve-out is correct.
- go vet clean; TDD tests cover positive/negative/coinbase cases; populateCache helper update is the necessary compensation.

Non-blocking note: the sibling SetCache in TxMetaCache.GetMeta (txmetacache.go:472) is still gated only on BlockIDs/Conflicting, not inpoints completeness. Not triggerable today since GetMeta returns full inpoints, but worth considering if you want the invariant enforced uniformly.

LGTM.

oskarszoon (approved):
Approve. Correct two-sided fix for the inpoints-less cache poisoning: producer (`BatchDecorate`) stops caching a non-coinbase meta with empty `ParentTxHashes`, consumer (`processTxMetaUsingCache`) treats such an entry as a miss so the store fallback self-heals it.

Verified:
- The producer `continue` only skips the cache write — caller decoration is already done by `utxoStore.BatchDecorate` at `txmetacache.go:510`, so no item is left undecorated.
- The consumer partial→miss falls through to the same `missed` path as a genuine miss (slot left unset), so no transaction is dropped.
- The load-bearing invariant holds: a valid non-coinbase tx always has ≥1 deduped parent hash (BDK rejects 0-input non-coinbase; `go-subtree` derives `ParentTxHashes` from inputs and dedups), so empty on a non-coinbase always means "not populated", never a legitimately empty set.
- Strictly conservative: a flagged-partial entry only ever routes to the authoritative store — no consensus-acceptance bypass. This is a liveness/wedge fix, not a soundness change.
- Guarding `len == 0` (vs the serializer's strict `nil` at `go-subtree subtree_meta.go:174`) is the right, stricter choice — it also rejects the empty-non-nil shape.

Non-blocking:
- The "self-heals the cache" behavior is correct by construction (the store fallback decorates with `TxInpoints`, so the re-cache passes the producer guard) but is untested — worth a partial→store→re-cache→hit round-trip test.
- Nits: add an N-inputs-single-parent case to pin dedup→still-cached; the partial-miss test only seeds the struct path (byte-mode covered indirectly); no producer-side coinbase-with-no-parents cache test.

PR #1096, opened 2026-06-15
chore(settings): docker.m Aerospike overload-safe batcher defaults

oskarszoon:
## Problem

The `docker.m` single-node context runs against **one** Aerospike node with a 128-connection pool (`utxostore.docker.m`: `ConnectionQueueSize=128`, `LimitConnectionsToQueueSize=true`, `MinConnectionsPerNode=16` → ~112 usable), but the batcher-concurrency defaults are sized for clustered deployments.

During legacy quick-validation IBD, the spend phase fires **three** batcher types concurrently (spend + increment + setDAH), each capped at the global `utxostore_batcherMaxConcurrent=64`. Worst case **3×64 = 192 concurrent `BatchOperate` calls against a ~112-usable pool** → connection exhaustion (`MAX_ERROR_RATE`, -15) and device write-queue overflow (`DEVICE_OVERLOAD`, 18).

Observed on a mainnet quickstart host: repeated stalls through the ~820k blocks (many >20k-output transactions). The overload itself was transient, but each one previously triggered a multi-hour `errors.Is` grind on the giant aggregated spend-error chain.

## Change

`docker.m`-scoped overrides only — clustered defaults untouched:

| Setting | docker.m | was (global) | why |
|---|---|---|---|
| `utxostore_batcherMaxConcurrent` | 24 | 64 | 3×24=72 < 112 pool, with headroom for get/store + 16 warm conns |
| `utxostore_incrementBatcherSize` | 1024 | 256 | bigger batches = fewer round-trips on >20k-output pagination records (first ops to fail) |
| `utxostore_setDAHBatcherSize` | 1024 | 256 | same |
| `legacy_storeBatcherConcurrency` | 4 | 32 | createUtxos feed 1024×32=32768 → 4096, matched to the batcher cap |
| `aerospike_batchPolicy` TotalTimeout | 10s | 5m | a single `BatchOperate` must finish well inside the server-side overload-retry budget for that retry to control cadence |

## How it fits with the in-flight overload work

This is the **preventive** half of a pair:

- **This PR** keeps offered concurrency under the connection pool / device capacity, so overload is *rare* rather than per-block.
- **#1085** (server-side overload retry, 30s budget) absorbs the *residual* transient spikes that still slip through — and its retry only works as designed if the inner client policy is shorter than its budget, which the `aerospike_batchPolicy.docker.m` timeout here provides.
[...]

PR #1025, opened 2026-06-02
fix(utxo/aerospike): stop caller-side goroutine/connection leak in batchers

oskarszoon:
## Problem

A production node (`bsva-ovh-teranode-eu-2`) accumulated **13,693 goroutines** — 8,664 parked in `(*Store).get`'s `select` and 4,096 in validator errgroups — that **persisted after the underlying Aerospike wedge cleared** and `client_connections` had returned to 0. A leak that survives the wedge is not the wedge; it's a caller-side bug holding the resources.

## Root cause (teranode-side, not the v8 client)

1. **Orphaned completion channels on panic.** go-batcher recovers panics raised inside the batch dispatch fn (`dispatchAndRecord` wraps `b.fn(batch)` in `defer recover()`). A panic part-way through a `sendXxxBatch` — e.g. an unchecked type assertion in `getTxFromBins` on a malformed/missing bin — left every not-yet-signalled per-item completion channel orphaned. The worker survived; the submitters parked **forever**, because the contexts threaded down from legacy sync / validation carry **no deadline**, so the `get` select's `ctx.Done()` arm never fires.
2. **Retry amplification.** `sendGetBatch` wrapped the already-retrying v8 batch call (MaxRetries within `TotalTimeout=5m`) in *another* 3× retry loop, stacking the worst-case stall to ~3× TotalTimeout (~15m) and growing the submitter backlog faster than it drained under sustained server slowness.

The v8 client itself bounds the batch read at `TotalTimeout` (5m) and releases the connection — verified against `v8.7.1-bsv3`. No v8 change is required to stop this leak. (One latent v8 robustness bug — `connection.go` resets the socket deadline on every partial read, so `SocketTimeout` is per-syscall and a slow-drip server evades it; only bites callers with `TotalTimeout=0`, which teranode is not — flagged separately, out of scope here.)

## Changes

- **Panic safety net** — `signalBatchPanic` + `trySignal` (new `batch_completion.go`); a recover-defer on all 7 dispatch fns (`sendGetBatch`, `sendOutpointBatch`, `sendStoreBatch`, `sendSpendBatchLua`, `sendIncrementBatch`, `sendSetDAHBatch`, `setLockedBatch`) re-signals every per-item channel on panic. Non-blocking, so it never double-delivers or wedges the worker.
[...]

PR #1172, opened 2026-06-26
fix(utxo/aerospike): make batcher leak guard budget-aware so overload retries aren't aborted

oskarszoon:
## Problem

Legacy block sync aborts under sustained Aerospike overload:

```
ERROR | netsync/handle_block.go:1366 | [extendTransactions] ... DONE in 40.22s with error:
PROCESSING (4): failed to batch-decorate previous outputs ->
SERVICE_UNAVAILABLE (50): aerospike outpoint batch did not complete within 40s
```

The submitter-side leak guard `batcherWait` (`stores/utxo/aerospike/batch_completion.go`, added in #1025) was derived from `batchPolicy.TotalTimeout + 30s` alone and had **no knowledge of the uaerospike overload-retry wrapper** (#1085 / #1136). That wrapper runs an initial `BatchOperate` and *only then* starts an `OverloadRetryMaxElapsed` budget clock (`retryBatchOnOverload` sets its deadline **after** the first attempt), re-issuing overloaded records until the budget is spent.

So the legitimate submit-to-completion wall time is roughly `initial attempt + overload budget`, and the guard could fire **while the wrapper was still legitimately retrying** — aborting the caller. This was acute for `docker.m` (40s guard vs 2m budget), but the 5m global default was affected too:

| context | old guard (`TT + 30s`) | legit max (`TT + 2m budget`) | premature abort? |
|---|---|---|---|
| global default (5m) | 5m30s | up to **7m** | yes |
| docker.m (10s) | **40s** | up to 2m10s | yes — the stall in the log |

## Fix

**Make the leak guard budget-aware** — the real, general fix. `batcherWaitFor` now derives the guard from the **sum** `TotalTimeout + OverloadRetryMaxElapsed + 30s`, not the larger of the two: the two windows are *sequential* (the budget clock starts after the initial attempt), so they add. This:

- fixes **every** context, not just `docker.m`;
- can't go stale when a context lowers `TotalTimeout` — the guard tracks the budget instead of a hand-tuned magic number;
- adds nothing when the overload retry layer is disabled (`OverloadRetryMaxElapsed <= 0`), preserving prior behaviour.

It is documented as **best-effort, not a strict bound**: the initial connection-permit wait (`~TotalTimeout/10`, `acquirePermit`) and the untimed `reacquirePermit` between retries still fall outside the sum; the +30s grace only partially absorbs them.

**Revert the `docker.m` `TotalTimeout` 90s → 10s.** The first commit raised it to 90s purely to make the *old* guard outlast the 2m budget. With the guard now budget-aware that bump is unnecessary, and keeping it had a cost (per review): the connection-permit acquisition wait scales with `TotalTimeout × 0.1`, so 90s widened it ~1s → ~9s. Reverting to 10s keeps that wait small and preserves the original small-inner-timeout / wrapper-controls-cadence design from #1096. Net `settings.conf` change is comment-only; the value matches `main`.

**Regression test.** `batcherWaitFor` is extracted as a pure helper so the invariant can be unit-tested without a live Aerospike client. `batcher_wait_test.go` pins:
- the **sum** formula (exact-equality, so a regression back to `max()` fails);
- the invariant `guard > overload budget` for the `docker.m` and global-default contexts;
- fallback (`TotalTimeout <= 0`) and retry-disabled (`budget <= 0`) behaviour.

## Review feedback addressed

- **ChiR1 (blocking)** — guard not proven to outlast the retry path; `max()` insufficient → switched to the sum `TotalTimeout + budget + 30s` and reworded the comment to drop the "covers" claim (now "best-effort").
- **ChiR2 (non-blocking)** — 90s widened the permit-acquire wait to ~9s → reverted to 10s, which moots it; documented the `~TotalTimeout/10` coupling in the comment.

## Scope / non-goals

The general default-hardening is the **code guard** — it applies to the 5m global default too. The `docker.m` policy *values* (10s / 5s / 2 retries) remain a deliberate **single-node** tuning and are intentionally not pushed onto the global default, which is sized for multi-node cluster batch latencies; revisiting the global policy values is a separate change with a larger blast radius.

## Verification

- `go build ./stores/utxo/aerospike/... ./settings/... ./util/...` — OK
- `go vet ./stores/utxo/aerospike/` — clean; `gofmt` — clean
- `go test ./stores/utxo/aerospike/ -run 'BatcherWait|Leak' -race` — PASS


icellan (approved):

ctnguyen (changes requested):
This is a tight, well-scoped config-only change aimed at a real defect: the `docker.m` leak guard (`batcherWait = TotalTimeout + 30s`, `batch_completion.go:95`) was firing at 40s and aborting legacy block sync while the overload-retry wrapper still had budget to keep going. The stale-coupling diagnosis is sound — `aerospike_overload_retry_max_elapsed` does default to `2m` with no `docker.m` override, and the old comment's "30s default" rationale is genuinely obsolete. Raising the guard from 40s to 120s is a clear improvement over the status quo, and the rewritten comment is far more informative than what it replaces.

## Blocking issues

### ChiR1 — The guard is not proven to outlast the retry path; `batcherWait` and the overload budget start at different points

- **Observed:** The PR frames the fix as `batcherWait (120s) ≥ overloadRetryMaxElapsed (2m)`, treating the two as a like-for-like comparison. They are not measured from the same instant:
  - `batcherWait` is armed at **submit time** — e.g. in `PreviousOutputsDecorate` the timer (`get.go:1226`) starts right after the items are queued, before the batcher worker has even dispatched the first `BatchOperate`.
  - `retryBatchOnOverload` (`util/uaerospike/overload_retry.go`) runs the initial `do(records)` **first** and only then sets `deadline := time.Now().Add(maxElapsed)`. So the 2m budget clock starts *after* the initial batch attempt returns an overload result.

  The wall-clock the legitimate retry path can consume, measured from submit, is therefore `initial permit acquisition + initial BatchOperate latency + 2m budget` (plus any blocking permit re-acquisition between retries — `reacquirePermit`, `client.go:593`, blocks with **no timeout**). Under connection pressure the initial permit wait alone can be up to ~9s (see ChiR2). A 120s guard that starts earlier than, and equals, a 2m budget that starts later does **not** provably outlast it — the guard can fire while the wrapper is still legitimately retrying, which is the exact failure mode this PR exists to eliminate.
- **Why it matters:** The PR's own lesson is that a timeout comment asserting a coverage invariant ("the inner call must finish well inside that budget") went stale and misled. Shipping a new comment that asserts `batcherWait ≥ budget → "covers it"` re-introduces the same trap, because the invariant is not actually established. In the tail (overloaded initial attempt + permit contention) legacy sync can still abort with `aerospike outpoint batch did not complete within 2m0s` — better than 40s, but not the guaranteed fix the description claims.
- **Fix:** Give the guard real margin over the *whole* submit-to-completion path — not just the larger of the two budgets — and stop the comment asserting coverage it cannot guarantee. The retry path's wall time from submit is roughly `initial permit acquisition (≤ ~9s, see ChiR2) + initial BatchOperate (≤ TotalTimeout) + overload budget (2m) + blocking permit reacquisitions`. The budget-aware guard the PR floats as follow-up, `max(TotalTimeout, OverloadRetryMaxElapsed) + 30s`, is an improvement — it tracks the budget instead of going stale — but it does **not** prove coverage: it adds only 30s over the larger *single* budget, while the initial permit wait, the initial attempt, and the untimed `reacquirePermit` all fall outside that budget. A formula closer to the stated invariant sums the pre-budget cost in:

  ```go
  // Cover the initial attempt AND the full overload budget, with grace for the
  // pre-budget permit wait. Still best-effort: reacquirePermit is untimed.
  base := util.GetAerospikeBatchPolicy(tSettings).TotalTimeout
  budget := tSettings.Aerospike.OverloadRetryMaxElapsed
  return base + budget + 30*time.Second
  ```

  Even this is not a strict guarantee while `reacquirePermit` (`client.go:593`) can block without bound; a true fix would also bound that wait or accept the guard as explicitly best-effort. At minimum, reword the `settings.conf` comment so it states the guard *improves on* the old 40s rather than "covers" the 2m budget.

## Non-blocking issues

### ChiR2 — Raising `TotalTimeout` 10s→90s also widens the semaphore-acquisition wait ~1s→~9s

- **Observed:** `acquirePermit` (`client.go:561`) derives its wait as `TotalTimeout * 0.1` (floor 100ms), independent of `SocketTimeout`/`MaxRetries`. So this change also raises the per-call connection-permit acquisition timeout from ~1s (at 10s) to ~9s (at 90s).
- **Why it matters:** This is exactly the kind of latency the `settings.conf` comment claims is "bound small" by `SocketTimeout=5s / MaxRetries=2` — but the up-to-9s permit wait precedes the attempt and is not bounded by either. It directly feeds the start-point skew in ChiR1 (it widens the window between when `batcherWait` starts and when the overload deadline starts) and, under connection pressure, lets a single decorate call sit ~9s before its first server round-trip.
- **Fix:** Acknowledge this in the comment as a real consequence of the bump (the "bound a single non-overload attempt small" claim is now incomplete), and factor the ~9s permit wait into whatever margin is chosen for ChiR1.

oskarszoon:
@ctnguyen addressed both findings in 21fc6b959 — the diagnosis was right, and it reframed the fix away from the docker.m magic number toward the code guard.

**ChiR1 (start-point skew / `max()` insufficient).** Confirmed: `retryBatchOnOverload` runs the initial `do(records)` and only then sets `deadline = now + maxElapsed` (`overload_retry.go:182`→`:204`), so the budget is sequential to the initial attempt, not overlapping. The guard now sums the two windows rather than taking the larger:

```
batcherWaitFor = TotalTimeout + OverloadRetryMaxElapsed + 30s
```

This covers initial-attempt (≤ TotalTimeout) + full budget, and — your wider point — it fixes every context, not just docker.m. The 5m global default had the same latent bug (old guard 5m30s vs up to 7m legitimate). I dropped the "covers the 2m budget" wording: the comment now states the guard is **best-effort**, since the initial permit wait (~TotalTimeout/10) and the untimed `reacquirePermit` between retries still fall outside the sum and the +30s only partially absorbs them.

**ChiR2 (90s widens permit-acquire wait ~1s→~9s).** With the guard now budget-aware, the 90s bump was only ever there to outlast the old guard, so it's no longer needed — reverted docker.m `TotalTimeout` back to 10s. Permit-acquire wait stays ~1s and the small-inner-timeout cadence design from #1096 is preserved. Net `settings.conf` change is now comment-only; the value matches `main`. The `~TotalTimeout/10` coupling is called out in the comment.

Also extracted `batcherWaitFor` as a pure helper and added `batcher_wait_test.go`, which pins the sum (exact-equality, so a regression back to `max()` fails) and the invariant `guard > overload budget` for the docker.m and global-default contexts.

Re-requesting review.

ordishs (approved):
Approving. Verified the mechanism against the code: \`retryBatchOnOverload\` sets the budget \`deadline\` *after* the initial \`do(records)\` (overload_retry.go:182,204), so the two windows are genuinely sequential and summing \`TotalTimeout + OverloadRetryMaxElapsed + 30s\` is correct (not \`max()\`). \`enabled()\` matches the "disabled when \`OverloadRetryMaxElapsed <= 0\`" comment, the only caller (aerospike.go:250) is behaviorally unchanged, and the new tests pass under \`-race\`. Tight, defensive change.

One non-blocking observation: the "legitimate wall time ≈ initial-attempt + budget" model in the \`batcherWaitTimeout\` doc (batch_completion.go:97-100) omits the *final* retry attempt. The loop checks \`remaining <= 0\` before sleeping (overload_retry.go:213), so the last \`do(failed)\` is dispatched while budget remains and can then run up to another full \`TotalTimeout\` past the deadline. The true legitimate ceiling is therefore ~\`TotalTimeout + budget + TotalTimeout\`, but the guard only adds one \`TotalTimeout\` — e.g. a 7m30s guard vs a ~12m theoretical ceiling on the 5m global default.

The trigger is narrow (under sustained overload the final attempt is rejected fast; it only bites if overload clears right at budget exhaustion and the final BatchOperate then runs slow-but-successful), and this is strictly better than the old guard, so not a blocker. Worth either a one-line comment noting the final-attempt term alongside the existing permit-wait caveat, or adding a second \`+TotalTimeout\` to the global path if you want the guard to cover it.

ctnguyen (approved):
### Well done, approved

✅ Reverted the docker.m TotalTimeout bump back to 10s — the permit-acquire-wait widening (~1s→~9s) is gone, and the settings.conf change is now comment-only.

✅ Made the guard budget-aware: batcherWaitFor = TotalTimeout + OverloadRetryMaxElapsed + 30s — summed, not max'd, which is correct given the windows are sequential (retryBatchOnOverload arms its budget clock only after the initial do(records), confirmed at overload_retry.go:182→204). Extracted as a pure helper, disabled-overload path preserved, and the comment now honestly describes the guard as best-effort rather than a guarantee.

✅ Added a test pinning the sum with exact-equality, so a regression back to max() fails.

PR #1173, opened 2026-06-26
test(blockassembly): add BA-SUBTREE contract conformance tests

sugh01:


ordishs (approved):
Reviewed the test-only changes (6 new files, 847 additions). Verified each test's assertions against the production code and ran the full new test set under `go test -race` — both packages pass clean. The concurrency-safety reasoning in the test comments holds up: the candidate data races on `chainedSubtrees`/`blockIntervals`/`subtreeNodeCounts` are refuted by valid happens-before chains and an idle dispatcher in these load-free isolated tests. Solid, carefully-constructed contract coverage. LGTM.

liam (approved):

oskarszoon (approved):
Tests look good and pin the real contract — verified `-race` passes on both `blockassembly` and `subtreeprocessor`. No blockers. A few optional tightenings:

- **`coinbase_placeholder_position_test.go` (control sub-test "correctly placed placeholder passes"):** the assertion is inside `if err != nil`, but a fresh processor returns `nil`, so the branch never runs — the sub-test passes vacuously. Use `require.NoError(t, err)` to actually guard the happy path.
- **`subtree_announcement_test.go` `TestSubtreeAnnouncementContent`:** asserting `captured.Hash == subtree.RootHash()` is tautological — `sendSubtreeNotification` just copies its argument. The doc comment claims it verifies a "content-derived" hash but doesn't exercise the store→sender pipeline. Either route through the full path or soften the comment.
- **`periodic_announcement_test.go`:** the `Length()` assertion can flake — the 75ms ticker shares a select with the dequeue path, so it can fire before all 3 batches are dequeued. Wait for `currentSubtree.Length() == announcedLength` before asserting.
- **`subtree_dah_test.go`:** the meta-DAH check silently assumes `StoreTxInpointsForSubtreeMeta=true`; add a precondition assert so a default flip fails loudly rather than with a misleading "must have been stored".
- Nits: typo `waitForPeriodicit` → `waitForPeriodicity`; and `dynamic_sizing_decrease_test.go` passes `blockchain.Mock{}` where `nil` (used in the periodic test for the same isolated case) fits the no-mock convention.

PR #1175, opened 2026-06-29
fix(utxo/aerospike): enforce first-seen in expression spend path (#1153)

### LINKED ISSUE #1153: [TN-014][high] utxo/aerospike: expression-spend path does not enforce first-seen double-spend (opt-in, default off)
**Severity:** high · **Confidence:** high (static; runtime proof needs live Aerospike) · **Pass:** UTXO integrity
**Component:** `stores/utxo/aerospike` (`spend_expressions.go`)

### Summary
With `aerospike_enable_spend_filter_expressions=true` (default **false**, requires `utxoBatchSize==1`), the spend filter never checks whether the target UTXO is already spent, and `ListSetOp` overwrites the element unconditionally. A second (double-spend) spender in a separate batch passes the filter and overwrites the original spender — violating first-seen at the store level. The default Lua and SQL paths are first-seen-safe.

### Evidence
- Flag default: `settings/aerospike_settings.go:22`, `settings/settings.go:174` (false). Gate: `stores/utxo/aerospike/spend.go:493-494`.
- `spend_expressions.go:110-205` (`buildSpendFilterExpression`): checks utxos-exists, `creating!=true`, `conflicting!=true`, `locked!=true`, maturity, no `UtxoSpendableIn`, list-size>offset — but **no unspent-element check**. Comment `:193-197` ("ListSetOp will fail if the UTXO doesn't match") is incorrect; `ListSetOp` (`:231`) only fails on out-of-bounds index.
- Only guard is in-batch `FilterConflictingDuplicateSpendClaims`; existing test covers in-batch only.
- Secondary: idempotent re-spend in a separate batch double-increments `spentUtxos` → DAH `spentUtxos==recordUtxos` never holds → record never pruned (leak).

### Why it matters
A spend path that fails first-seen is a direct double-spend / fund-safety hazard if an operator enables the flag for performance (its description invites this with only a generic "test thoroughly").

### Suggested fix
For `utxoBatchSize==1`, gate the write on an expression asserting the element is exactly 32 bytes (unspent) before `ListSetOp`; route filter-outs through the Lua UDF (which classifies SPENT). If not expressible, don't offer the expression path for spends. Until fixed, keep the flag off and document as unsafe.

### Suggested tests
Add a **cross-batch** sequential double-spend test (`SpendBatcherSize=1`, both `enableExpressions:false/true`): Create(parent); Spend(A); Spend(B, different spender, same outpoint) → assert `ErrSpent`, stored spender still A, `spentUtxos==1`. Runtime proof needs a TestContainers Aerospike (see open question OQ-14).

---

ordishs:
## Summary

Fixes #1153. With `aerospike_enable_spend_filter_expressions=true` (default **false**, requires `utxoBatchSize==1`), the spend filter never checked whether the target UTXO was already spent, and `ListSetOp` overwrote the element unconditionally. A second (double-spend) spender in a **separate batch** passed the filter and silently replaced the first-seen spender — breaking first-seen at the store level. The in-batch `FilterConflictingDuplicateSpendClaims` guard did not cover this cross-batch case.

A secondary defect: idempotent re-spends in a separate batch double-incremented `spentUtxos`, so the DAH prune condition (`spentUtxos==recordUtxos`) never held and the record leaked.

## Fix

`buildSpendFilterExpression` now asserts the element at `offset` is still the unspent **32-byte** UTXO hash before `ListSetOp`:

```go
ExpEq(
    ExpListGetByIndex(ListReturnTypeValue, ExpTypeBLOB, ExpIntVal(offset), utxosBin),
    ExpBlobVal(utxoHash),
)
```

An unspent element is the bare 32-byte hash; once spent it is 68 bytes (hash + 36-byte spendingData), so the equality fails the moment the UTXO is spent. The record is `FILTERED_OUT` and retried through the existing Lua-UDF path, which returns the correctly classified `ErrSpent` for a real double-spend, or an idempotent success **without re-incrementing** `spentUtxos` for a re-spend by the same spender (fixing the leak). Comparing against the expected hash also guards against a wrong-slot `ListSetOp` on a hash mismatch.

The prior comment claimed byte-by-byte comparison was impossible in expressions — that is incorrect; `ExpEq` over blob expressions is supported by the client.

## Tests

New `spend_expressions_doublespend_test.go`, table-driven over both the Lua and expression paths (`UtxoBatchSize=1`, `SpendBatcherSize=1`):

- **`TestStore_SpendExpressions_CrossBatchDoubleSpend`** — `Create`; `Spend(A)`; `Spend(B)` (different spender, same outpoint, separate batch) → asserts `ErrSpent`, stored spender still A, `spentUtxos==1`.
- **`TestStore_SpendExpressions_IdempotentRespendNoDoubleIncrement`** — re-spend by the same spender in a separate batch → idempotent success, `spentUtxos` stays 1.

TDD-red verified: with the fix reverted, the `expressions` subtest fails ("An error is expected but got nil") — the double-spend silently succeeds.

## Verification

- Full package suite (live Aerospike via TestContainers): `go test -tags testtxmetacache ./stores/utxo/aerospike/...` → **ok** (main package + pruner).
- `go vet ./stores/utxo/aerospike/` clean; `go build`/`go test ./settings/` ok.

## Docs

Updated the `EnableSpendFilterExpressions` setting longdesc to document first-seen safety and the `utxoBatchSize==1` requirement, replacing the prior "behavior differs / test thoroughly" caveat.


icellan (commented):
No blocking issues — this correctly closes a real store-level first-seen / double-spend hole on the expression spend path. Pre-fix, `ListSetOp` overwrote an already-spent element unconditionally, so a different spender arriving in a separate batch silently replaced the first-seen spender and `spentUtxos` was double-incremented (defeating the `spentUtxos==recordUtxos` prune). The new `ExpEq(ExpListGetByIndex(BLOB, offset), utxoHash)` clause is the right fix — unspent elements are the bare 32-byte hash, spent ones are 68 bytes, so the match flips false the moment the UTXO is spent and the record is re-routed to the authoritative Lua UDF. Happy and double-spend paths are well covered.

A few optional, non-blocking nits:

- **`stores/utxo/aerospike/spend_expressions.go:328`** — `SpendingData.Bytes()` is computed here to feed `buildSpendFilterExpression`, but that function's `spendingData` parameter (`:115`) is unused; the same value is recomputed at `:344` for `buildSpendOperations` (which does use it). Dropping the unused param + this argument saves an allocating `Bytes()` call per spend on the hot path.
- **`spend_expressions_doublespend_test.go:68` / `:138`** — the `{name: "lua", enableExpressions: false}` sub-case runs with the flag off, so `useExpressionSpend()` is false and it never exercises the new `ExpEq` clause or the `FILTERED_OUT → Lua` routing; it's a parity control against reference Lua behaviour. A one-line comment would keep it from reading as 2× coverage of the fix.
- **`spend_expressions_doublespend_test.go:115`** — `require.Equal(t, firstSpends[0].UTXOHash[:], elem[0:32])` is a tautology here: `spendTx` and `spendTx2` spend the same outpoint and therefore carry the same UTXO hash, so this holds regardless of which spender won. The first-seen guarantee rests entirely on the `elem[32:68]` spender-bytes check at `:116`.
- **`spend_expressions_doublespend_test.go:175`** — the idempotent re-spend test asserts only `spentUtxos == 1`; consider also asserting `elem[32:68]` still equals the first-seen spender, so a regression that preserves the counter but corrupts the stored spending data is caught.
- **`settings/aerospike_settings.go:22`** — the longdesc still recommends enabling "for reduced spend latency", but filtered-out records (already-spent / idempotent re-spend / `UtxoSpendableIn`) now pay the expression round-trip plus a follow-up Lua batch. Worth caveating that the latency benefit only holds when filter-outs are rare.

icellan (approved):

oskarszoon (approved):
Reviewed with bitcoin-expert + security-auditor + golang-pro. Fix is correct and minimal.

Verified against current code (not the PR description):
- Full-blob `ExpEq` is length-sensitive — a 68-byte spent element (hash + 36-byte spendingData) can never match the bare 32-byte hash, so an already-spent UTXO is always FILTERED_OUT. The assumption the fix rests on holds.
- No TOCTOU: the filter rides the same BatchWritePolicy as the write, evaluated atomically server-side.
- FILTERED_OUT routing is exclusive (only that code + KEY_NOT_FOUND get special branches); real Aerospike errors still surface as StorageError. Lua returns ErrSpent for a real double-spend, idempotent success without re-incrementing spentUtxos for a same-spender re-spend (leak fixed).
- Gated on utxoBatchSize==1, so offset is always 0 — no multi-UTXO slip-through.
- Tests exercise the genuine cross-batch path (separate Spend() calls bypass the in-batch FilterConflictingDuplicateSpendClaims guard) and run lua+expression parity.

No P0/P1. Optional follow-ups, none blocking:
- Dead `len(filterConditions)==1` branch (spend_expressions.go:224) — pre-existing, min is now 6 conditions.
- The new tests are integration-only (live Aerospike) — not run by `make test`; ensure the aerospike CI lane covers them so the regression has teeth.
- Stale "Returns nil" doc comment on buildSpendFilterExpression.

PR #1177, opened 2026-06-29
MvP-4491 Enforce BIP68 on non-coinbase txs in NewBlockFromMsgBlock

ctnguyen:
This resolve [[fix(model): enforce BIP68 on non-coinbase txs in NewBlockFromMsgBlock](https://github.com/bitcoin-sv/teranode/issues/4491)](https://github.com/bitcoin-sv/teranode/issues/4491)

### DIFF
--- a/daemon/test_daemon.go
+++ b/daemon/test_daemon.go
@@ -1712,7 +1712,7 @@ func createAndSaveSubtrees(ctx context.Context, subtreeStore blob.Store, txs []*
 		}
 	}
 
-	if err = storeSubtreeFiles(ctx, subtreeStore, subtree, subtreeData, subtreeMeta); err != nil {
+	if err = storeSubtreeFiles(ctx, subtreeStore, subtree, subtreeData, subtreeMeta, 100); err != nil {
 		return nil, err
 	}
 
@@ -1720,7 +1720,7 @@ func createAndSaveSubtrees(ctx context.Context, subtreeStore blob.Store, txs []*
 }
 
 // storeSubtreeFiles serializes and stores the subtree, subtree data, and subtree meta in the provided subtree store.
-func storeSubtreeFiles(ctx context.Context, subtreeStore blob.Store, subtree *subtreepkg.Subtree, subtreeData *subtreepkg.Data, subtreeMeta *subtreepkg.Meta) error {
+func storeSubtreeFiles(ctx context.Context, subtreeStore blob.Store, subtree *subtreepkg.Subtree, subtreeData *subtreepkg.Data, subtreeMeta *subtreepkg.Meta, deleteAtHeight uint32) error {
 	subtreeBytes, err := subtree.Serialize()
 	if err != nil {
 		return err
@@ -1731,7 +1731,7 @@ func storeSubtreeFiles(ctx context.Context, subtreeStore blob.Store, subtree *su
 		subtree.RootHash()[:],
 		fileformat.FileTypeSubtreeToCheck, // this needs to be FileTypeSubtreeToCheck for tx processing to occur
 		subtreeBytes,
-		options.WithDeleteAt(100),
+		options.WithDeleteAt(deleteAtHeight),
 		options.WithAllowOverwrite(true),
 	)
 	if err != nil {
@@ -1748,7 +1748,7 @@ func storeSubtreeFiles(ctx context.Context, subtreeStore blob.Store, subtree *su
 		subtreeData.RootHash()[:],
 		fileformat.FileTypeSubtreeData,
 		subtreeDataBytes,
-		options.WithDeleteAt(100),
+		options.WithDeleteAt(deleteAtHeight),
 		options.WithAllowOverwrite(true),
 	)
 	if err != nil {
@@ -1765,7 +1765,7 @@ func storeSubtreeFiles(ctx context.Context, subtreeStore blob.Store, subtree *su
 		subtree.RootHash()[:],
 		fileformat.FileTypeSubtreeMeta,
 		subtreeMetaBytes,
-		options.WithDeleteAt(100),
+		options.WithDeleteAt(deleteAtHeight),
 		options.WithAllowOverwrite(true),
 	)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


oskarszoon (approved):
Approving. The consensus core is correct and the fix does what it claims — packing the non-coinbase txs into a single subtree is what finally lets `block.Valid()` / the subtree-validation pass see them, which is the whole reason the BIP68 reject tests could be un-skipped.

Checked the things that could actually break and they hold up:
- The single-subtree path takes the `len(hashes)==1` branch in `CheckMerkleRoot` → `RootHashWithReplaceRootNode`, which rebuilds with the canonical duplicate-when-odd rule, so the root matches the canonical flat merkle root for **any** tx count, not just the power-of-two case the test covers. The padding-sensitive top-tree logic is never reached here.
- `fee=0` is the safe direction: `checkBlockRewardAndFees` sums `SubtreeSlices.Fees`, so the only failure mode is a false-*reject* if a coinbase claims fees — never a false-accept. And it's not a regression (pre-PR `SubtreeSlices` was nil → also 0).
- Only production caller is the genesis block (`stores/blockchain/sql/sql.go:1208`), which is coinbase-only, so the new loop and the not-size-limit-aware single-subtree shortcut never run on a real block. The new doc-comment warning against misuse is the right guard.

Non-blocking cleanups (P2), happy to merge without them but worth a follow-up:

1. **Merkle test only covers N=2 (power of two).** `TestNewBlockFromMsgBlock_MultiTransactionSubtree` uses coinbase+1 tx — a full 2-leaf tree with no odd-duplication. The loop runs for all non-coinbase txs; the regression case worth pinning is N=3 (coinbase+2), which forces the duplicate-when-odd path through `RootHashWithReplaceRootNode`. Add a 3-tx sub-case asserting `CheckMerkleRoot` passes against an independently-computed root. The algorithm is correct — this is just coverage.

2. **`model/Block.go:221`** — `return nil, sizeErr` is the only error path in the new loop that doesn't wrap; loses the tx index and the processing error code. Practically unreachable (`tx.Size()` ≥ 0), so minor, but:
   ```go
   return nil, errors.NewProcessingError("failed to compute size for transaction %d", i, sizeErr)
   ```

3. **`model/Block.go:209-224`** — each non-coinbase tx is serialized into a buffer then re-parsed into a `bt.Tx` that's immediately discarded; only the hash and size are used. Both are available off the wire tx directly (`msgBlock.Transactions[i].TxHash()` + `SerializeSize()`), dropping two allocs per iteration and the dead `bt.Tx` construction.

4. **BIP68 reject tests** depend on the height window `CSVHeight ≤ height < GenesisActivationHeight` (post-Genesis drops relative-locktime enforcement). Holds in the current regtest config, but a one-line comment noting the assumption would stop it silently turning into a no-op if regtest Genesis activation ever moves. The new `ErrorContains("sequence lock ... not satisfied")` assertions are a good guard against false-passing on an unrelated error.

Reviewed, not run — confirm the un-skipped e2e tests + the new unit test are green in CI before merge.

model/Block.go
```diff
@@ -182,15 +188,59 @@ func NewBlockFromMsgBlock(msgBlock *wire.MsgBlock, optionalSettings *settings.Se
...
+		// The transaction hash is the canonical double-SHA256 of the wire
+		// serialization, identical to the value the subtree/merkle path expects.
+		hash := msgBlock.Transactions[i].TxHash()
+
+		txSize, sizeErr := safeconversion.IntToUint64(msgBlock.Transactions[i].SerializeSize())
+		if sizeErr != nil {
+			return nil, errors.NewProcessingError("failed to compute size for transaction %d", i, sizeErr)
+		}
+
+		if addErr := subtree.AddNode(hash, 0, txSize); addErr != nil {
```
ordishs, line 218: Optional hardening (non-blocking): recording every non-coinbase node with fee `0` is fine for today's callers, but it's a latent reward-check trap. `Valid()` -> `checkBlockRewardAndFees` sums `SubtreeSlices[i].Fees` directly (no recomputation), so a future caller that builds a multi-tx block here whose coinbase legitimately claims `subsidy + fees` and then validates it in-memory (no subtree-store round trip) would be rejected with a spurious "coinbase output greater than fees + subsidy". It's currently only guarded by the doc comment above. Consider a hard guard instead — e.g. skip the reward check when fees are known-unpopulated, or reject in-memory reward checks for `txCount > 1`.

daemon/test_daemon.go
```diff
@@ -1775,6 +1775,63 @@ func storeSubtreeFiles(ctx context.Context, subtreeStore blob.Store, subtree *su
...
+// It is the test-side complement to model.NewBlockFromMsgBlock: that constructor
+// records the subtree root on the block, but the transaction bodies (carried only
+// in the original wire block) are not part of the serialised block and must be
+// supplied to the store separately for validation to reach them.
+//
+// deleteAtHeight controls blob retention and must be above the height of the
+// block being validated so the subtree is not pruned before validation runs.
+// The returned root hash matches block.Subtrees[0] for a block built from the
+// same coinbase-plus-transactions set.
+func (td *TestDaemon) StoreSubtreeForBlock(t *testing.T, txs []*bt.Tx, deleteAtHeight uint32) *chainhash.Hash {
```
ordishs, line 1794: Test cleanup (non-blocking): this reimplements two existing helpers — the build loop duplicates `createAndSaveSubtrees`, and the store block is a parameterized copy of `storeSubtreeFiles`. It can't reuse `storeSubtreeFiles` as-is only because that helper hardcodes `options.WithDeleteAt(100)`. Adding a `deleteAtHeight uint32` param to `storeSubtreeFiles` and having both callers use it would collapse all three into one persist path and drop the hardcoded `100`.

ordishs (approved):
Approving — the change is consensus-correct for its actual callers. The sole production caller (genesis block in `stores/blockchain/sql/sql.go:1208`) is coinbase-only, so the new `txCount > 1` branch never fires there and genesis behavior is byte-for-byte unchanged. `TxHash()` is the right hash for the subtree/merkle path (no witness data in BSV), `AddTx(tx, i+1)` indexing is correct, and the un-skipped BIP68 reject tests genuinely exercise enforcement: regtest `GenesisActivationHeight=10000` keeps block height 112 inside the `CSVHeight <= height < Genesis` window, and the subsidy-only coinbase makes the fee=0 reward sum pass.

Two non-blocking comments below — one optional hardening note, one test-code cleanup. Neither blocks merge.

freemans13:
## Code review (automated, medium effort)

Reviewed the diff across correctness, reuse/simplification/efficiency, altitude, and conventions angles, then verified candidates. **No confirmed correctness bug** — the scarier candidates were checked and refuted: the subtree the constructor builds and the one `StoreSubtreeForBlock` persists are sized identically (`len(txs)+1` == coinbase + non-coinbase) and reconcile to the same root; `ProcessBlock` with an empty `baseURL` takes the synchronous non-optimistic path so the rejection surfaces; and the single-subtree merkle reconciliation handles non-power-of-two tx counts correctly (canonical duplicate-last padding). The new tests asserting the *specific* rejection string is a nice touch that guards against passing on an unrelated error.

Three findings worth your assessment — none is a blocker:

### 1. `daemon/test_daemon.go:1794` — `StoreSubtreeForBlock` largely duplicates existing helpers (reuse)
The new helper is ~90% a copy of `createAndSaveSubtrees` (line 1677) followed by `storeSubtreeFiles` (line 1723): same `NewIncompleteTreeByLeafCount(len(txs)+1)` → `AddCoinbaseNode` → per-tx `AddNode`/`AddTx`/`SetTxInpointsFromTx` loop, and the same three `FileTypeSubtreeToCheck`/`Data`/`Meta` stores. The only thing preventing reuse is that `storeSubtreeFiles` hardcodes `options.WithDeleteAt(100)`, which would prune too early for these tests. Threading a `deleteAtHeight` parameter through `storeSubtreeFiles`/`createAndSaveSubtrees` would let the new helper delegate instead of fork. Note the two copies already diverge in the `AddNode` fee/size args (`(uint64(i), uint64(i))` vs `(0, tx.Size())`) — a maintainability trap once they drift further.

### 2. `model/Block.go:208` — wasteful `wire.MsgTx` → `bt.Tx` round-trip in the non-coinbase loop (efficiency)
Each non-coinbase tx is serialized into a `bytes.Buffer` and re-parsed via `bt.NewTxFromBytes`, purely to read `tx.TxIDChainHash()` and `tx.Size()`. `wire.MsgTx` already exposes `TxHash()` (the canonical txid the subtree needs) and `SerializeSize()` with no allocation — the test file uses `wire.MsgTx.TxHash()` itself at `Block_test.go:1581-1582`. Unlike the coinbase (which must be retained as a `bt.Tx`), the non-coinbase `bt.Tx` is discarded after the hash/size are read, so the full serialize+parse cycle per tx is avoidable.

### 3. `model/Block.go:199` — fee=0 leaves `block.Valid()`'s reward check unsound for fee-bearing blocks (documented limitation, no code guard)
`checkBlockRewardAndFees` (line 695) sums `SubtreeSlices[i].Fees` directly without recomputing, so a block built by this constructor whose coinbase claims fees (`coinbaseOutput > subsidy`) would be wrongly rejected when validated in-process via `block.Valid()`. Your comment documents this clearly and all current callers are subsidy-only offline/genesis blocks, so it's not a live bug — flagging only that the precondition is enforced by prose rather than code. Worth a guard (or a doc note at the call sites) before this constructor gains a new caller.

Findings 1 and 2 are safe to act on directly; 3 is a tradeoff you've already partly addressed in the comment.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #1178, opened 2026-06-30
feat(blockvalidation,legacy): below-checkpoint outpoint-only IBD fast path (SQL, p2p + legacy, default-off)

freemans13:
## Summary

A below-checkpoint **outpoint-only** IBD fast path for **both** sync paths (p2p catchup AND legacy netsync), behind a **default-OFF** setting, on **SQL** UTXO stores. Below a hardcoded checkpoint the chain prefix is already certified valid by the pinned block hash, yet Teranode still does a large amount of per-block work that exists only to *decide* validity — and IBD is bound on it (the node sits near 0% CPU; ~75% of per-block cost is reading historical parent-output data to build extended txs, fees, and a per-input integrity checksum). This change skips that work below the checkpoint.

Gated by `OutpointOnlyBelowCheckpoint` (default false) AND each path's existing quick-validation gate AND `block.Height <= HighestCheckpointHeight(ChainCfgParams.Checkpoints)` (the **hardcoded** checkpoints — never the operator catchup override) AND a SQL UTXO store, the fast path:

1. **Skips decorate** — no parent-output reads for spent inputs.
2. **Spends by outpoint** — the UTXO-hash integrity checksum is gated off (it is only a checksum, never an addressing key or consensus commitment; spends already address by `(txid, vout)`).
3. **Writes `fee = 0`** into subtrees — fees have no consensus consumer below checkpoint.
4. **Minimal create** — zeroes the per-input parent script/satoshis columns, while always persisting the input **outpoint** (pruner needs it) and **all outputs** in full.

**Both sync paths are delivered:**
- **p2p catchup** (`quick_validate.go`): the gate drives the four behaviours directly.
- **legacy netsync** (`handle_block.go`): skips the dominant `BatchPreviousOutputsDecorate`, zeroes subtree fees, minimal-creates, and threads a new validator option `WithOutpointOnlySpend` into `Validate` (alongside the existing `WithSkipScriptValidation`) so the validator skips its residual parent height-read and spends by outpoint. (In-process validator; same precondition the legacy quick path already relies on.)

## Why it's safe

A hardcoded checkpoint commits transitively to every ancestor, so below-checkpoint work is state *reconstruction*, not re-validation. Anything that only decides validity (script checks — already skipped; fee/no-inflation; BIP68; the integrity checksum) is redundant; anything affecting the reconstructed UTXO-set contents is preserved. Invariants held & tested:

- **Tamper-detection is fully retained below the checkpoint.** This fast path does *not* skip the checks that prove the data is authentic — only the checks that re-*decide* already-certified validity. Every below-checkpoint block still has its header hash / proof-of-work verified and its prev-hash linkage walked back to the pinned checkpoint hash (the block-chain hash checks), and its transactions are still bound to that header by full **merkle-root verification** of the subtrees (`CheckMerkleRoot`). So any block or transaction that had been altered in transit or on disk fails to reproduce the committed merkle root / checkpoint-linked hash and is rejected exactly as before — the checkpoint tells us *which* history is canonical, and these hash/merkle checks prove the bytes we reconstruct from *are* that history.
- **Outputs are byte-identical** to a normal sync (the per-output `utxo_hash` is computed from the tx's own outputs, not parent data) — so above-checkpoint spends of below-checkpoint outputs validate normally.
- **Double-spend detection, coinbase maturity, and the pruner are untouched** — only the utxo-hash comparison is gated.
- A height-guard on `checkBlockRewardAndFees` makes the `fee=0` subtrees safe against the `reconsiderblock`/revalidation paths that can re-run `Block.Valid` on a below-checkpoint block. The fee-write height (both paths) and this read-skip height are computed from the **same** hardcoded-checkpoint set.
- The new validator option enforces its precondition (`OutpointOnlySpend` requires `SkipScriptValidation`) at runtime, and — as defence-in-depth parity with the block-validation `I4` guard — the validator independently rejects `OutpointOnlySpend` above the highest hardcoded checkpoint.
- The below-checkpoint boundary is a **single source of truth** (`model.HighestCheckpointHeight`); the fee=0 write side, the `checkBlockRewardAndFees` read-skip side, and the validator guard all derive from it, so they cannot drift (invariant I3, guard-tested).

### Flag-off behaviour — two intentional, checkpoint-safe changes (not fully byte-identical)

The store-level create/spend paths are byte-identical when the options are unset. Two below-checkpoint behaviours, however, change **independently of `OutpointOnlyBelowCheckpoint`** even when it is off. Both are safe on any node (a below-checkpoint block cannot be accepted unless it chains to the pinned checkpoint hash, so the skipped work is redundant), but they are called out here rather than hidden under a blanket "byte-identical" claim:

1. **`checkBlockRewardAndFees` skips the coinbase no-inflation check at or below the highest hardcoded checkpoint, ungated by the setting.** This is required: a flag-on node writes subtree fees=0, and the read-skip must still hold after the operator turns the flag off (so a fee=0 block written earlier still revalidates). `model` cannot see the per-validation flag (import cycle), and gating the read side would itself be the bug.
2. **`SkipScriptValidation` now crosses gRPC where it previously did not.** Before this PR it was silently dropped to a *remote* validator, so below-checkpoint legacy blocks were fully script-validated on distributed deployments. They now correctly skip script validation on a remote validator too, independent of the fast-path flag. Ensure the validator and its callers (subtreevalidation/netsync) are upgraded together; an older validator ignores the field and full-validates (slower, never corrupts).

## Scope / non-goals

- **PostgreSQL/SQLite only.** The fast path is fenced to SQL UTXO stores via a shared `settings.IsSQLUtxoStore` allowlist; an Aerospike (or unconfigured-default) node keeps it OFF, so enabling the flag there cannot halt IBD. Aerospike support (create + Lua UDF + version-gated rollout) is **deferred to a follow-up (Stage B)**.
- **Works with both in-process and remote (gRPC) validators.** The validator options `SkipScriptValidation` and `OutpointOnlySpend` are carried over the wire via new `optional bool` proto fields (and the HTTP large-tx fallback), so the legacy path is not restricted to `UseLocalValidator=true`. Request-only fields with safe deployment skew: an older validator ignores them and performs full validation (slower, never corrupts). This also fixes the pre-existing gap where `WithSkipScriptValidation` was silently dropped over gRPC.
- No above-checkpoint / steady-state behavioural change. No consensus change.

## Testing

- Store-layer unit tests (real sqlite store): outpoint-only spend, flag-off still enforces the hash, minimal create across **all three** input-write paths, fee=0 without calling `GetFees`.
- The decisive **UTXO-set equivalence** test (path-agnostic): the same create+spend sequence (incl. a spent output and a sequence-locked tx) run flag-ON vs flag-OFF produces **byte-identical output rows**; fails on any output-hash divergence.
- Gate tests for both paths: hardcoded-checkpoint bound, operator-override fence, SQL-store guard, per-block metric, and a legacy `legacyOutpointOnly` truth table.
- Validator option: outpoint-only spend on un-extended inputs + the precondition guard.
- Legacy wiring: paired ON/OFF tests asserting decorate is skipped and subtree fees are 0.
- `Block.Valid` fee-skip below checkpoint + write/read checkpoint-height agreement.
- A few multi-block / RPC-level scenarios are `t.Skip`-deferred to e2e/smoketest — to be written before the flag is enabled on any shared node.

`go build ./...` and `go vet` clean; all touched packages pass under `-race`.

## Rollout

Default-off setting; enable on a PostgreSQL test node, run from-genesis IBD to the checkpoint, confirm the UTXO set matches a flag-off control, then measure the throughput gain on the legacy path. Reversible (flip off + restart; existing records remain valid — outputs were always fully persisted).


ctnguyen (changes requested):
The core insight — that below a hardcoded checkpoint the chain prefix is already certified, so per-block work that only *decides* validity is redundant — is correct, and the implementation preserves exactly the state that matters: outputs are persisted in full, outpoints are always retained for the pruner, and double-spend detection is untouched (spends still address rows by `(txid, vout)`, with `utxo_hash` used only for the now-gated integrity comparison — verified in `sql.go` `trySendSpendBatch*`). The test suite is genuinely strong: the UTXO-set byte-equivalence test (`TestUTXOSetEquivalence_OutputsByteIdentical`) is the right decisive check, the ON/OFF polarity tests are load-bearing rather than tautological, and both SQL input-write paths are exercised. Deployment skew for the new proto fields is handled correctly (request-only, additive, fail-safe in both directions), and the gate helpers fail closed on every missing conjunct.

## Prior review follow-up

- Previously-raised BIP68 phase-2 defect (relative-locktime txs wrongly rejected because `utxoHeights` was left nil on the outpoint-only path): **now addressed** — the phase-2 gate in `validateTransaction` leads with `validationOptions.OutpointOnlySpend ||`, and `TestValidate_OutpointOnlySpend_BIP68HeightSkipped` is a proper regression guard using a non-disabled height-based sequence lock.
- Previously-raised ungated re-extend in `validateTransaction` (the un-gated `!tx.IsExtended()` re-extend re-issued the per-parent `PreviousOutputsDecorate` reads the fast path exists to eliminate): **now addressed** — the re-extend is gated on `!validationOptions.OutpointOnlySpend`, with `TestValidate_OutpointOnlySpend_NoParentRead` asserting zero decorate calls.
- Previously-raised no-inflation-skip concern in `model/Block.go`: **still open** — carried forward as ChiR1 below.

## Blocking issues

### ChiR1 — No-inflation check is skipped below the checkpoint unconditionally, not gated on the feature flag
**Problem:** In `model/Block.go` `checkBlockRewardAndFees` (~lines 639–658), the new early-return fires for every block at or below the highest hardcoded checkpoint, independent of `OutpointOnlyBelowCheckpoint`. The flag is not even reachable at this layer. So `Block.Valid` becomes a no-op for the fee/no-inflation check below the checkpoint on paths that run regardless of the fast path — full-validation catchup (`CatchupAllowQuickValidation=false`) and `reconsiderblock`/revalidation — even on a node that never enabled the feature and synced with real subtree fees.
**Why it matters:** This directly contradicts the PR's central safety claim ("flag-off is byte-identical to current behaviour everywhere"), and it widens reduced validation to the *default* configuration in consensus-critical code. It is safe under the checkpoint commitment (a below-checkpoint main-chain block that inflated could not match the pinned hash), but that is a conscious reduction of defence-in-depth that reviewers/operators relying on the description will not expect.
**Fix:** There is a legitimate reason it must be flag-independent (fast-path-synced blocks persist `fee=0` subtrees, so a later flip to flag-off must still skip on `reconsiderblock`) — but that only applies to nodes that used the fast path. Either (a) gate the skip so it does not alter default-config behaviour, or (b) keep it flag-independent but obtain explicit consensus-domain sign-off and correct the PR description to state that flag-off is *not* byte-identical for below-checkpoint `Block.Valid`. State the rationale in the code comment either way.

## Non-blocking issues

### ChiR2 — Read-side checkpoint-height loop duplicates `HighestCheckpointHeight` and is under-tested
**Problem:** The inline loop in `checkBlockRewardAndFees` re-implements `blockchain.HighestCheckpointHeight` (confirmed byte-equivalent today) to avoid an import cycle. `TestCheckpointHeight_WriteReadAgree` only exercises `MainNetParams`, so a future edge-case change to the canonical function (e.g. negative-height handling) would silently drift on the read side.
**Why it matters:** A write/read divergence would desynchronise the fee-write boundary from the fee-skip boundary — exactly the invariant (I3) the PR relies on for revalidation safety.
**Fix:** Extract the loop into a tiny helper in a package `model` may import (or a shared `chaincfg` util) so there is one implementation, or make the agreement test table-driven over multiple param sets (incl. empty and negative-height checkpoints).

### ChiR3 — Decisive multi-block / cross-checkpoint / reconsiderblock coverage is deferred to e2e
**Problem:** The cross-checkpoint boundary spend (T-B1), fork rejection (T-B3), `reconsiderblock` RPC path (T-B4), and first-above-checkpoint validation (T-I2) are all `t.Skip`-placeholders deferred to e2e/smoketest. The store-level equivalence test is strong, but no integration test yet exercises the two sync pipelines end-to-end with the flag on.
**Why it matters:** The highest-risk interactions (a below-checkpoint output later spent above the checkpoint; revalidation of a `fee=0` block) are precisely the ones without automated coverage.
**Fix:** No change required for a default-off merge, but the deferred e2e tests must land and pass before the flag is enabled on any shared node — as the PR's own rollout section states. Track them explicitly so they are not lost.

### ChiR4 — `TestQuickValidateOutpointOnly_GateBounds` relies on the ambient settings context to satisfy the SQL-store guard
**Problem:** `GateBounds` builds `BlockValidation{settings: test.CreateBaseTestSettings(t)}` and asserts `quickValidateOutpointOnly` returns true at heights 150/200, but that now also requires `settings.IsSQLUtxoStore`. It never sets a store URL, so it depends on `SETTINGS_CONTEXT=test` resolving `utxostore` to `sqlite:///utxostore` (settings.conf:1258). Under the repo's canonical test invocation (`make test`, PR CI) that holds and the test passes; but under a base/unset context (`utxostore = null:///`) or an aerospike context, `IsSQLUtxoStore` returns false and the `require.True` assertions fail.
**Why it matters:** Every sibling test that expects the gate to engage (`TestQuickValidate_OutpointOnly_NoDecorate_ZeroFees`, `TestOutpointOnly_MetricIncrementsBelowOnly`, `TestOperatorOverrideFence`) explicitly establishes a SQL store via `setSQLUtxoStoreURL`; `GateBounds` is the lone exception, making it fragile to context and inconsistent with the suite.
**Fix:** Set an explicit SQL URL in `GateBounds` (e.g. `u.settings.UtxoStore.UtxoStore, _ = url.Parse("sqlitememory://test")`), or fold its cases into the `TestQuickValidateOutpointOnly_SQLStoreGuard` table, so the SQL precondition is stated in-test rather than inherited from the environment.

### ChiR5 — `TxMetaDataFromTxNoFee` duplicates `TxMetaDataFromTx`
**Problem:** `util/tx_meta.go` adds `TxMetaDataFromTxNoFee`, a near-verbatim copy of `TxMetaDataFromTx` differing only in `Fee: 0` and skipping `GetFees`.
**Why it matters:** Two copies can drift if a future field is added to `meta.Data` construction in only one.
**Fix:** Have `TxMetaDataFromTx` compute the fee and delegate to a shared builder (or add a `fee uint64` parameter to an internal helper) so the struct is constructed in one place.

## Recap

| ChiR1 | Unconditional no-inflation skip below checkpoint | YES | 70% |
| ChiR2 | Duplicated checkpoint-height loop, under-tested | | 30% |
| ChiR3 | Multi-block/reconsiderblock coverage deferred to e2e | | 25% |
| ChiR4 | GateBounds test depends on ambient SQL-store context | | 20% |
| ChiR5 | `TxMetaDataFromTxNoFee` duplicates `TxMetaDataFromTx` | | 10% |

ordishs (changes requested):
Overall this is well-built and unusually well-tested — the UTXO-set output-equivalence, coinbase-maturity, and double-spend guarantees all verify as genuinely safe, and the gate/wire round-trips are thoroughly covered. Requesting changes on a single blocking item; two trivial follow-ups noted below.

## Blocking

### 1. `model/Block.go:706` — the no-inflation check is disabled for **all** blocks ≤ highest checkpoint, ungated by the feature flag

`checkBlockRewardAndFees` now returns `nil` for `b.Height <= HighestCheckpointHeight(...)` on every node, even with `OutpointOnlyBelowCheckpoint` off, and it is reached from the full `Block.Valid()` path — not only the fast path. So this ships a consensus-relevant defence-in-depth removal to nodes that never opted into the feature.

The safety argument ("a below-checkpoint fork must chain to the pinned hash") only holds once the highest checkpoint block is actually part of the validated chain. With a checkpoint configured **above the live tip** (a forward/optimistic checkpoint on a custom/STN/new network, or before headers connect to the checkpoint), `height <= HighestCheckpointHeight` is still true while there is no transitive pinning — an inflated coinbase at that height would be accepted and written to the real UTXO set.

Please resolve one of:
- keep the skip gated behind the feature flag (and/or the same conditions that guarantee checkpoint pinning), **or**
- assert the highest hardcoded checkpoint is at/below an already-validated, hash-pinned height rather than relying on the operational assumption in the comment, **or**
- explicit maintainer sign-off that ungating this is intended, with the reasoning that no path reaches `Block.Valid()` on a below-checkpoint block before its ancestry to the pinned hash is confirmed (reorg candidate, detached block, forward checkpoint).

(Severity note: I confirmed the skip is ungated and on the full-validation path, but did not fully prove an exploitable path on mainnet — hence "justify or gate," not "exploitable today.")

## Please also fix — both trivial

### 3. `services/validator/Validator.go:1488` — add a store-type fail-closed guard to the `OutpointOnlySpend` path
The validator gates `OutpointOnlySpend` on checkpoint height + `SkipScriptValidation`, but never on the UTXO store type. Aerospike `Spend`/`Create` ignore `SkipUTXOHashCheck`/`SkipExtendedInputs` and call `GetSpends`/`TxMetaDataFromTx`, which hard-error on the un-decorated inputs and stall IBD. All in-tree callers gate on `settings.IsSQLUtxoStore`, so this is latent — but the fence lives only at the caller. A one-line fail-closed check at the validator (reject `OutpointOnlySpend` on a non-SQL store) closes it.

### 5. `model/Block_test.go:1938` — `TestBlock_CheckBlockRewardAndFees` "valid block reward and fees" is now vacuous
The subtest builds a height-1 block and calls `checkBlockRewardAndFees(&chaincfg.MainNetParams)`; height 1 is below mainnet's highest checkpoint, so the new early-return short-circuits before the reward arithmetic and the test no longer exercises what it was written for. The sibling `TestBlock_CheckRewardAndFees_WithHeight` was already updated to height `10_000_000` for exactly this reason — apply the same fix here.

## Non-blocking (follow-ups are fine)
- **#2** `services/validator/Client.go:292` — `SkipScriptValidation` now crosses gRPC (ungated), so remote-validator + legacy-netsync deployments stop full-validating below-checkpoint scripts on upgrade. Consensus-safe and already noted in the description; just ensure validator/netsync upgrade together.
- **#4** `services/validator/Validator.go:1475` — `OutpointOnlySpend` enforces `SkipScriptValidation` but not `SkipPolicyChecks`; a caller pairing outpoint-only with policy checks on would hit `bad-txns-inputs-too-large` on the nil script. Latent (all internal callers set `SkipPolicyChecks`).
- **#6** `services/legacy/netsync/handle_block.go:527` — `legacyOutpointOnly` re-checks `chainParams==nil` and re-computes `height <= HighestCheckpointHeight` after `quickValidationAllowed` already proved both; can `return true` after the guard.
- **#7 / #9** Asset API surfaces fast-path-degraded data for historical below-checkpoint blocks when the feature is on: `GetSubtree`/`GetBlockSubtrees` return `fee=0`, and `repository.go:312 GetTransaction` serves `ExtendedBytes()` with zero satoshis / empty prev-scripts (`Store.Get` reads the zeroed input columns back with no output-row fallback; the `TxStore` fallback only fires on error). Same root cause; resolve together or document as unsupported under the flag. (The higher-severity "wrong coinbase fee in a mined block" variant was checked and refuted — block assembly only sums live subtrees it built above the checkpoint.)
- **#8** `quick_validate.go:1199` I4 guard is documented-dead; `isSQLUtxoStore` is a pass-through inconsistent with the legacy path; `quickValidateOutpointOnly` is recomputed ~7×/block. Minor.

Cleared as **not bugs**: output byte-identity / per-output `utxo_hash`, coinbase maturity, double-spend detection, the single-sourced write/read checkpoint boundary, `baseOpts[:len:len]` aliasing guard, proto oneof nil-safety, and CLAUDE.md conventions.

freemans13:
Thanks @ordishs — addressed in `3089d0b4b`.

**Blocking**

**#&NoBreak;1 `checkBlockRewardAndFees` ungated skip** — now **gated on `OutpointOnlyBelowCheckpoint`**. A node that never enabled the fast path keeps the full no-inflation check on the `Block.Valid()` path; the skip only applies where fee=0 subtrees can actually exist. This is exactly your first suggested resolution and it closes the forward-checkpoint case you described: with a checkpoint configured above the validated tip, `height <= HighestCheckpointHeight` is true but a default (flag-off) node still runs the inflation check, so an inflated coinbase at that height is rejected. `Block.Valid()` already has `*settings.Settings`, so the flag threads through cleanly (`settings.BlockValidation.OutpointOnlyBelowCheckpoint`).

The one residual trade-off — a node that ran the fast path (flag on), then turns the flag **off** and reconsiders a below-checkpoint block, would re-run the check against its fee=0 subtrees and could reject it — is documented at the call site. Keeping the check active for the default population is the safer default; happy to revisit if you'd prefer a different balance.

**Please also fix**

**#&NoBreak;3 validator store-type fence** — added a one-line fail-closed guard in `validateInternal`: `OutpointOnlySpend` on a non-SQL store is now rejected before any store access, using the same `settings.IsSQLUtxoStore` allowlist the callers use. New test `TestValidate_OutpointOnlySpend_RejectedOnNonSQLStore`.

**#&NoBreak;5 vacuous test** — `TestBlock_CheckBlockRewardAndFees` "valid block reward and fees" now calls with the flag **off**, so the reward arithmetic genuinely runs at height 1 (coinbase == 50 BTC subsidy → valid). I also added a flag-off assertion to `TestCheckBlockRewardAndFees_SkipsBelowHardcodedCheckpoint` so the gate itself is covered (below checkpoint: flag-on skips, flag-off enforces).

**Non-blocking follow-ups** — leaving these for a follow-up as you suggested:
- **#&NoBreak;2** gRPC `SkipScriptValidation` propagation — documented in the PR description ("Flag-off behaviour" section); validator + netsync must upgrade together.
- **#&NoBreak;4** `OutpointOnlySpend` doesn't require `SkipPolicyChecks` — latent (all internal callers set it); can add a paired guard if you'd like.
- **#&NoBreak;6** `legacyOutpointOnly` redundant re-check after `quickValidationAllowed` — trivial simplification, will fold in.
- **#&NoBreak;7/#&NoBreak;9** Asset API surfacing fee=0 / zeroed-input data for historical below-checkpoint blocks under the flag — will resolve or document as unsupported-under-flag together.
- **#&NoBreak;8** I4 dead guard / `isSQLUtxoStore` pass-through / per-block recompute — minor; I4 comment already reworded to state it's a redundant fail-closed re-assertion.

Re-requesting review.

oskarszoon (approved):
Re-reviewed at 3089d0b4 (consensus-critical panel + adversarial verification, reconciled against ctnguyen's and ordishs' earlier change-requests). The three blocking/must-fix items are resolved — verified against current code, not just the commit message:

- **No-inflation skip now gated on the flag (ChiR1 / #1).** `checkBlockRewardAndFees` takes `skipBelowCheckpoint` and only skips when `OutpointOnlyBelowCheckpoint` is on, so a default (flag-off) node keeps the full check below checkpoint — closing the forward-checkpoint / pre-connection window where `height <= HighestCheckpointHeight` is true without transitive pinning. Regression test asserts both directions (flag-off + inflated coinbase below checkpoint → rejected). The flag-on→off→reconsider trade-off is documented in-code.
- **Validator store-type fail-closed guard (#3).** `OutpointOnlySpend && !IsSQLUtxoStore` now rejects before any store access, with a test. Aerospike is fenced at both the caller and the validator — a misconfigured caller can no longer reach the un-decorated-input hard-error / IBD stall on a non-SQL store.
- **Non-vacuous reward test (#5).** The height-1 subtest now runs flag-off and exercises the arithmetic.

Approving. Non-blocking follow-ups (none gate this default-off merge):

- **Release gate:** the deferred cross-checkpoint e2e (T-B1/T-B3/T-B4/T-I2, still `t.Skip`) must land and pass before the flag is enabled on any shared node, per the rollout section. Track them explicitly.
- `SkipScriptValidation` crossing gRPC is now the only skip-flag without an above-checkpoint height guard (OutpointOnlySpend has both a height and a store guard) — document it as intentionally trust-delegated or add a symmetric assertion.
- Asset API serves fast-path-degraded data under the flag (`repository.go:312` `GetTransaction` → `ExtendedBytes()` with zeroed input columns; `GetSubtree`/`GetBlockSubtrees` fee=0) — resolve together or document as unsupported under the flag.
- Minor: `TxMetaDataFromTxNoFee` duplicates `TxMetaDataFromTx` (DRY); `OutpointOnlySpend` doesn't also assert `SkipPolicyChecks` (latent); dead nil-guard + redundant recompute at `handle_block.go:527`.

freemans13:
Follow-up on the #3 store fence (commit `7b611d6ec`): replaced the `settings.IsSQLUtxoStore` URL-scheme sniff with a first-class capability on the `utxo.Store` interface —

```go
SupportsOutpointOnlySpend() bool
```

The store now declares whether it honours the fast-path ignore-flags, instead of three call sites guessing from the URL scheme. SQL returns `true`; Aerospike/NullStore return `false` (Stage B just flips Aerospike's method); the logger and txmetacache decorators delegate to the wrapped store — so a cache/logger wrapping SQL now correctly reports `true`, which the scheme sniff couldn't see. `quickValidateOutpointOnly`, `legacyOutpointOnly` and the validator guard all call `store.SupportsOutpointOnlySpend()`; `settings.IsSQLUtxoStore` is deleted.

Full blockvalidation suite + model/validator/netsync/stores all green; `go vet`/gofmt clean.

freemans13:
Thanks @ctnguyen — this review predates the last few pushes, so most items are already resolved. Mapping each:

**ChiR1 (blocking) — no-inflation skip ungated → FIXED** (`3089d0b4b`). `checkBlockRewardAndFees` now takes a `skipBelowCheckpoint bool`, wired from `settings.BlockValidation.OutpointOnlyBelowCheckpoint` at the `Block.Valid()` call site. A node that never enabled the fast path keeps the full no-inflation check below the checkpoint (default-config behaviour unchanged). The flag-independent-revalidation nuance you flagged is preserved as the documented trade-off in the code comment: a node that *did* run the fast path (flag on) then flips it off and reconsiders a below-checkpoint block would re-run the check against its `fee=0` subtrees — a rare deliberate operator sequence. The PR description's "Flag-off behaviour" section already states flag-off is *not* fully byte-identical.

**ChiR2 (non-blocking) — duplicated checkpoint loop → FIXED** (`36c022c64`). Extracted `model.HighestCheckpointHeight` as the single source; `blockchain.HighestCheckpointHeight` now delegates to it, and `checkBlockRewardAndFees` calls it directly (no more inline copy). `TestCheckpointHeight_WriteReadAgree` is now table-driven over mainnet, testnet, empty, negative-only, unordered, single and trailing-lower sets; a new model-package test pins `checkBlockRewardAndFees`'s boundary to the shared function.

**ChiR4 (non-blocking) — GateBounds ambient-context dependency → FIXED** (`7b611d6ec`). The whole `IsSQLUtxoStore` scheme-sniff is gone: the gate now asks `store.SupportsOutpointOnlySpend()` (a capability on the `utxo.Store` interface). `GateBounds` injects a mock with the capability set explicitly, so it no longer inherits anything from `SETTINGS_CONTEXT`.

**ChiR5 (non-blocking) — `TxMetaDataFromTxNoFee` duplication → FIXED** (`a0610dd3a`). Both entry points now delegate to a single unexported `txMetaDataFromTx(tx, fee)` builder (real fee vs `fee=0`); one construction site, no drift.

**ChiR3 (non-blocking) — multi-block/reconsiderblock e2e coverage.** The consensus core of T-B1 (a coinbase minimally-created below the checkpoint still enforces maturity when spent outpoint-only) is now covered executably at the store layer (`TestMinimalCreate_CoinbaseMaturity_OutpointOnlySpend`), and a real Postgres `createCTE` test was added. The full multi-block/reconsiderblock pipeline tests remain `t.Skip` e2e placeholders as you note — tracked to land and pass before the flag is enabled on any shared node, per the rollout section.

Re-requesting review.

model/Block.go
```diff
@@ -684,11 +684,34 @@ func (b *Block) releaseTxMap() {
...
+	// configures a checkpoint ABOVE the current validated tip (forward/optimistic
+	// checkpoint) has height <= HighestCheckpointHeight true while pinning does not yet
+	// hold; there, gating keeps the inflation check active. HighestCheckpointHeight is the
+	// single source of truth shared with the write side (invariant I3); see checkpoint.go.
+	//
+	// Trade-off: a node that ran the fast path (flag on), then turns the flag OFF and
+	// reconsiders a below-checkpoint block, would re-run this check against fee=0 subtrees
+	// and could reject it. That is a rare deliberate operator sequence; keeping the check
+	// active for the default (flag-off) population is the safer default.
+	if skipBelowCheckpoint && b.Height <= HighestCheckpointHeight(params.Checkpoints) {
```
liam: **Two issues with this skip predicate — it looks wrong in opposite directions.**

**(1) It contradicts the PR's stated design and causes false `BLOCK_INVALID` on flag-off revalidation.** The PR description says this skip must be *"ungated by the setting ... the read-skip must still hold after the operator turns the flag off (so a fee=0 block written earlier still revalidates)."* But `skipBelowCheckpoint` is `settings.BlockValidation.OutpointOnlyBelowCheckpoint`, so the skip *is* gated on the live setting. A node that synced below-checkpoint blocks with the flag ON (subtrees persisted with `Fees=0`), then restores the default (flag OFF) and runs `reconsiderblock`/`RevalidateBlock` on such a block, no longer takes this branch: it computes `coinbaseOutput = subsidy + realFees > subtreeFees(0) + subsidy` and rejects the genuinely-valid, checkpoint-pinned block as `BLOCK_INVALID` — exactly the scenario the design says must not happen. The in-code comment acknowledges this as a "trade-off," so the comment and the PR body disagree; one of them needs to change.

**(2) It omits the store-support gate that every write path requires.** Both write paths (`quickValidateOutpointOnly`, `legacyOutpointOnly`) also require `utxoStore.SupportsOutpointOnlySpend()`; this read-skip checks only flag + height. On a store that returns `false` (Aerospike / unconfigured default) with the flag ON, the write side stays full (real fees written, fast path never runs) yet `checkBlockRewardAndFees` still returns `nil` for every block at height <= highest checkpoint — silently disabling the coinbase no-inflation check on a node where the flag is documented as a complete no-op, with no operator-visible signal. Since `model` can't see the store, the fix likely belongs at the call site in `Block.Valid` (`model/Block.go:583`), passing a predicate that also reflects store support.

Net: the predicate is neither "always skip below checkpoint" (what the design needs to survive a flag-off) nor "skip only when the write side actually ran fee=0" (store-aware). Worth reconsidering before this ships.

services/blockvalidation/quick_validate.go
```diff
@@ -1166,6 +1191,16 @@ func (u *BlockValidation) createAndSpendUTXOsForBatch(ctx context.Context, block
...
 	}
 
+	// Invariant I4: outpoint-only mode must never be applied above the highest hardcoded
+	// checkpoint. This re-asserts the height bound that quickValidateOutpointOnly already
+	// applied — it is redundant today (not an independent signal), kept as a cheap
+	// fail-closed check so a future refactor that loosens the gate cannot silently spend
+	// by outpoint above the checkpoint at the point of no return (create+spend).
+	outpointOnly := u.quickValidateOutpointOnly(block)
+	if outpointOnly && block.Height > blockchain.HighestCheckpointHeight(u.settings.ChainCfgParams.Checkpoints) {
+		return errors.NewProcessingError("[createAndSpendUTXOsForBatch] invariant I4 violated: outpoint-only active above checkpoint at height %d", block.Height)
```
liam: **This I4 guard is unreachable dead code, not the defense-in-depth the comment claims.** `outpointOnly` is `u.quickValidateOutpointOnly(block)`, whose final line returns `block.Height <= HighestCheckpointHeight(...)`. So `outpointOnly && block.Height > HighestCheckpointHeight(...)` is a logical contradiction — it can never be true and this error can never fire. A future refactor that "loosens the gate" would loosen `quickValidateOutpointOnly` itself, so `outpointOnly` would still track the height bound and the guard would still never trigger. It provides zero protection and adds a second O(checkpoints) scan on top of the one inside `quickValidateOutpointOnly`. To be genuinely load-bearing, the height bound here must be derived independently of `quickValidateOutpointOnly`'s result (e.g. gate on the setting + store only, then assert the height separately).

liam (commented):
Review focused on the below-checkpoint fast-path gating. The retained-safety claims hold up (spends key by outpoint not by hash, outputs stay byte-identical, coinbase maturity height is still persisted, merkle/PoW/prev-hash linkage still verified, and all `utxo.Store` implementers add `SupportsOutpointOnlySpend()` so there's no compile break). The concerns cluster on the read-side fee-check skip in `checkBlockRewardAndFees`, whose gating predicate looks wrong in two opposite directions. Inline comments on the two most severe spots below; two lower-severity items follow here.

**3. `quick_validate.go:1201` — the I4 guard is unreachable dead code.** `outpointOnly` is `quickValidateOutpointOnly(block)`, whose final line already requires `block.Height <= HighestCheckpointHeight(...)`, so `outpointOnly && block.Height > HighestCheckpointHeight(...)` is a contradiction that can never fire. A future refactor that "loosens the gate" would loosen `quickValidateOutpointOnly` itself, so `outpointOnly` would still track the bound and the guard would still never trigger — it provides zero protection while adding a second O(checkpoints) scan. To be genuinely fail-closed it must derive the height bound independently of `quickValidateOutpointOnly`'s result.

**4. Per-block mode recomputed at ~7 (blockvalidation) + 4 (legacy) call sites.** `quickValidateOutpointOnly(block)` is re-evaluated at ~lines 141, 208, 302, 884, 1167, 1199, 1518 (and `legacyOutpointOnly` at four sites in `netsync/handle_block.go`), each doing a settings check, a `SupportsOutpointOnlySpend()` interface dispatch, and an O(checkpoints) slice scan. The value is a pure function of `(settings, store, block.Height)` and is invariant within a block. Beyond the redundant scans across a from-genesis IBD, this scatter is the altitude risk behind findings 1-2: the mode is a per-block property re-gated at every seam, so a future edit that misses one site can make phases disagree (e.g. `fee=0` written while decorate still runs). Capturing it once per block validation and threading it would remove both the waste and the drift surface.

ordishs (approved):
Re-reviewed at `a0610dd3a`. All three requested items are resolved and verified — approving.

## Blocking item resolved

**#1 (`model/Block.go`) — no-inflation skip is now flag-gated.** `checkBlockRewardAndFees` takes `skipBelowCheckpoint` and `Block.Valid()` passes `settings.BlockValidation.OutpointOnlyBelowCheckpoint`; the skip is now `if skipBelowCheckpoint && b.Height <= HighestCheckpointHeight(...)`. A node that never enabled the fast path keeps the full no-inflation check on the `Block.Valid()` path, and the forward/optimistic-checkpoint case (checkpoint above the validated tip) no longer silently disables it. The comment documents the one remaining trade-off (flag-on → write fee=0 → flag-off → reconsider could reject), which is a rare deliberate operator sequence — acceptable default. Directly tested with polarity: `TestCheckBlockRewardAndFees_SkipsBelowHardcodedCheckpoint` now asserts flag-off rejects the inflated below-checkpoint coinbase.

## Follow-ups resolved

**#3 (store guard) — fixed via a `SupportsOutpointOnlySpend()` store capability, which is cleaner than the fail-closed check I suggested.** The capability lives on the `Store` interface: SQL returns true, Aerospike/nullstore return false, decorators (`txmetacache`, `logger`) delegate, and the mock is configurable. Both gates (`quickValidateOutpointOnly`, `legacyOutpointOnly`) and the validator now consult it, and the validator fails closed (`OutpointOnlySpend && !SupportsOutpointOnlySpend()` → processing error). This also retires the `isSQLUtxoStore` pass-through and the `settings.IsSQLUtxoStore` scheme-sniffing (my #8), so the SQL-only fence is single-sourced on the store itself.

**#5 (vacuous test) — fixed.** `TestBlock_CheckBlockRewardAndFees` now calls with `false` and genuinely exercises the reward arithmetic (height-1 coinbase claims exactly the subsidy), and the skip test gained a flag-off polarity assertion.

Bonus: `util/tx_meta.go` consolidated onto a single `txMetaDataFromTx` builder, so the fee / no-fee entry points can't drift.

## Verification (fresh, at `a0610dd3a`)
- `go build ./model/... ./services/blockvalidation/... ./services/validator/... ./services/legacy/netsync/... ./stores/utxo/... ./util/...` — clean.
- `model` reward/fee tests — ok
- `validator` outpoint-only + store-guard + wire round-trip tests — ok
- `legacy/netsync` gate truth-table + fee/decorate tests — ok
- `blockvalidation` gate/override/metric tests — ok
- `stores/utxo/sql` minimal-create / outpoint-only / UTXO-set-equivalence tests — ok
- `util` — ok

## Non-blocking (as before — fine to defer)
- **#2** `SkipScriptValidation` crossing gRPC (ungated, consensus-safe below checkpoint) — coordinate validator/netsync upgrade; noted in the description.
- **#4** `OutpointOnlySpend` doesn't also require `SkipPolicyChecks` — latent (all internal callers set it).
- **#6** `legacyOutpointOnly` still re-checks `chainParams==nil` / re-derives the height bound after `quickValidationAllowed` proved both — minor.
- **#7/#9** Asset API surfaces `fee=0` and an `ExtendedBytes()` with zeroed input parent data for historical below-checkpoint txs when the feature is enabled — narrow, feature-on-only; worth a doc note or an output-row re-derivation in `Get`/`GetTransaction` eventually.

Approving; the remaining items are non-blocking polish.

freemans13:
Thanks @liam — this was the sharpest catch on the PR. You were right that the flag-gated predicate was incoherent in both directions. Fixed in `f2c436e4e` + `515b084fa`.

**`checkBlockRewardAndFees` (inline #1/#2) — reverted to an unconditional below-checkpoint skip** (dropped the flag parameter entirely). Your two failure cases both go away: (1) flag-off `reconsiderblock` of a fast-path `fee=0` block no longer recomputes `coinbaseOutput > 0+subsidy` and no longer false-rejects it; (2) there's no store/flag coupling to silently disable the check on a non-SQL node. The read side genuinely can't distinguish a fast-path `fee=0` block from a default one, so the skip can't depend on live config — unconditional is the only coherent option.

On the safety justification, I didn't want to hand-wave it, so I traced every `Block.Valid()` entry point. **Your instinct was right and my first comment overclaimed** — the literal "checkpoint enforcement rejects non-connected below-checkpoint blocks before full validation runs" is *false*. `Block.Valid` (with the skip) does run on not-yet-checkpoint-connected below-checkpoint blocks via: native catchup's standard path (`catchup.go:1094` when the fetched range has no checkpoint or `forkDepth>0`), initial sync before the checkpoint is reached, forward/optimistic checkpoints, and `reconsiderblock`/`RevalidateBlock` on a stored fork block (`BlockValidation.go:2136`, ancestry check is only "parent not marked invalid").

The property that *does* hold is narrower: such a block **cannot be adopted as the main chain past the checkpoint** — native catchup (`catchup.go:774`) and legacy headers-first (`netsync/manager.go:1779`) refuse a header chain whose in-range checkpoint hash mismatches the pinned hash, and `guardRunBelowHighestCheckpoint` (`blockchain/Server.go:2915`) blocks the FSM from RUNNING until the tip reaches the highest checkpoint, forcing a reorg onto the checkpoint-connected, inflation-free chain. So the skip is a deliberate defence-in-depth reduction on transient/fork paths, justified by the adoption gate. The code comment now says exactly that rather than the incorrect "before validation" claim.

This is the same defence-in-depth reduction @ordishs flagged (his "justify or gate"); I've now justified it precisely on the chain-adoption argument — flagging for a consensus-domain sign-off on that basis.

**#3 (dead I4 guard) — fixed** (`f2c436e4e`). You're right it was a tautology (`outpointOnly` already implies `height <= checkpoint`). It now derives the bound independently from the enabling inputs — `settings.OutpointOnlyBelowCheckpoint && utxoStore.SupportsOutpointOnlySpend() && height > HighestCheckpointHeight` — so it actually fires if a future edit drops the height clause from `quickValidateOutpointOnly`.

**#4 (per-block mode recomputed ~7+4×) — agree, tracking as a follow-up.** It's a pure function of `(settings, store, height)` and invariant within a block; capturing it once and threading it removes the recompute and the drift surface. It touches a lot of pipeline signatures in consensus-adjacent code, so I'd rather do it as its own reviewed change than fold it into this one — will open a follow-up. (The highest-drift seam, the model check, is now flag/store-independent, which removes the worst of the risk.)

Also, thanks for the doc nit: `CatchupAllowQuickValidation`'s longdesc says default `false` but `settings.go` sets `true` — noted, will correct.

Re-requesting review.

ctnguyen (commented):
## Items still needing work

### ChiR1 (blocking) — no-inflation check skipped unconditionally below the checkpoint
`checkBlockRewardAndFees` returns early for every block `<= HighestCheckpointHeight`, gated by nothing (signature unchanged, reached from the full `Block.Valid()` path — so it fires on full-validation catchup and `reconsiderblock` even on a node that never enabled the feature).

```go
if b.Height <= HighestCheckpointHeight(params.Checkpoints) {
    return nil
}
```

**Why it still needs work.** The revert to unconditional is defensible — pure flag-gating false-rejected `fee=0` fast-path blocks on `reconsiderblock`, so I accept the skip must not depend on the live flag. But the safety is currently *commentary, not code*: the forward/optimistic-checkpoint window (a checkpoint configured above the validated tip → `height <= HighestCheckpointHeight` true with no transitive pinning yet → an inflated coinbase could enter the real UTXO set) is closed only by an in-comment operational argument. A consensus-relevant check being disabled in the default config should rest on an enforced invariant or a recorded decision, not a comment. Resolve one of:

- **Option A (enforce).** Thread a caller-supplied "ancestry confirmed to the pinned checkpoint hash" predicate from the block-validation / netsync layer into `Block.Valid()` → `checkBlockRewardAndFees`; skip only when it holds. `model` cannot determine pinning itself, so the guarantee must come from the caller (not the highest-checkpoint-≤-validated-height test, which is too strong for initial catchup).
- **Option B (sign off).** Record explicit consensus-domain sign-off that no path reaches `Block.Valid()` on a below-checkpoint block before its ancestry to the pinned hash is confirmed (reorg candidate, detached block, forward checkpoint).

## Recap

| ID | Description | Required | Status |
|----|-------------|----------|--------|
| ChiR1 | No-inflation skip ungated below checkpoint (default config) | YES | TODO |
| ChiR2 | Duplicated checkpoint-height loop, under-tested | | FIXED |
| ChiR3 | Cross-checkpoint / reconsiderblock e2e coverage | | DEFERRED |
| ChiR4 | `GateBounds` test relied on ambient SQL context | | FIXED |
| ChiR5 | `TxMetaDataFromTxNoFee` duplicated `TxMetaDataFromTx` | | FIXED |

freemans13:
@liam — did #4 in this PR after all (`c8a0f8c5d`), rather than deferring: the fast-path mode is now computed **once per block** and threaded, so it's no longer re-derived at each seam.

- **blockvalidation**: computed in `quickValidateBlock`/`quickValidateBlockAsync`, carried on the `SubtreeProcessingBatch` struct (already threaded through the pipeline) + passed to the two standalone helpers. `quickValidateOutpointOnly(block)` is now evaluated exactly once per block per entry point.
- **legacy**: computed in `prepareSubtrees` and `ValidateTransactionsLegacyMode`, threaded into `extendTransactions`/`createSubtrees` and `createUtxos`/`PreValidateTransactions`. `legacyOutpointOnly(height)` evaluated once per orchestrator.

Consumers read the threaded value, so a new phase that forgets to thread it is a compile error rather than a silent inconsistency — which was the drift hazard you flagged. No behaviour change. Full blockvalidation + legacy/netsync suites pass.

Also fixed the doc nit you spotted separately (CatchupAllowQuickValidation longdesc said default `false`; it's `true`) — will fold in with the next push.

ctnguyen (approved):
No blocking defects were found this pass. The consensus-critical below-checkpoint no-inflation skip previously tracked here (ChiR1) has been **downgraded to non-blocking** — see the Status of previous points table above. Its code is in the correct, reviewer-endorsed form (flag-gating was proven wrong and must not be re-requested); what remains is a conscious accept-or-fix decision that rests with the author and the consensus-domain maintainers, not a defect this review gates merge on. It stays the **highest-attention item** in this review, and downgrading it does **not** mean the concern is resolved — see ChiR1's note for the two ways to settle it (Option A: enforce the ancestry predicate in code; Option B: record a consensus-domain sign-off).

## Status of previous points

| ID | Previous point | Status | Notes |
|----|----------------|--------|-------|
| ChiR1 | No-inflation check skipped below checkpoint unconditionally | NOT FIXED (non-blocking) | **Downgraded to non-blocking — this review no longer gates merge on it; whether to act is the author's / consensus-domain maintainers' call.** Not a code defect: the author first flag-gated the skip to satisfy the earlier reviews, then correctly reverted to the UNCONDITIONAL form after a later review proved flag-gating false-rejects a genuinely-valid fee=0 block on `reconsiderblock`/`RevalidateBlock` (the read side cannot distinguish a fast-path fee=0 block from a default one, so it cannot depend on live config). Flag-gating is off the table; unconditional is the correct read-side form, and the added in-code comment tracing every `Block.Valid()` entry point is good. The residual is a conscious trade-off, not a fix: the skip drops the coinbase no-inflation arithmetic below the checkpoint on every node (incl. full-validation catchup and `reconsiderblock`), safe only by the chain-adoption argument, with the narrow forward/optimistic-checkpoint window closed by comment rather than code. Downgrading does **not** assert the concern is resolved — it defers the decision: **fix it** via Option A (thread a caller-supplied ancestry-confirmed-to-pinned-hash predicate into `checkBlockRewardAndFees`, closing the window in code) — **or accept it** via Option B (a recorded consensus-domain maintainer sign-off covering the forward-checkpoint and detached-`reconsiderblock` cases) — or knowingly leave as-is. Recommended: record the sign-off so the decision is explicit rather than implicit. |
| ChiR2 | Read-side checkpoint loop duplicated `HighestCheckpointHeight` | FIXED | New `model.HighestCheckpointHeight` is the single source; `blockchain.HighestCheckpointHeight` delegates; table-driven `TestCheckpointHeight_WriteReadAgree`. |
| ChiR3 | Multi-block / cross-checkpoint / reconsiderblock e2e coverage | NOT FIXED | Deferred: still `t.Skip` (`TestCheckpointBoundary_B1/B3/B4/I2_Deferred`). Acceptable for a default-off merge, but these must land and pass before the flag is enabled on any shared node — track as a release gate. |
| ChiR4 | GateBounds test relied on ambient settings context | FIXED | Store capability now set in-test via `MockUtxostore{SupportsOutpointOnlySpendResult: true}`. |
| ChiR5 | `TxMetaDataFromTxNoFee` duplicated `TxMetaDataFromTx` | FIXED | Both delegate to one unexported `txMetaDataFromTx(tx, fee)` builder. |

icellan:
**Review — below-checkpoint outpoint-only IBD fast path** (reviewed at high + xhigh effort)

Solid fast path, but a few gaps to close before merge — two are actively triggerable on the validator endpoint.

**P1 — fix / decide before merge**

- **`services/validator/TxValidator.go:331` (guard `Validator.go:705`) — nil `PreviousTxScript` deref rejects valid txs.** The `OutpointOnlySpend` guard requires `SkipScriptValidation` but not `SkipPolicyChecks`. On the outpoint-only path the tx is un-extended (`PreviousTxScript==nil`), and `checkInputs` runs before the `SkipScriptValidation` short-circuit. On a node with `maxcoinsviewcachesize>0` receiving a below-checkpoint tx with `outpoint_only_spend=true`, `skip_script_validation=true`, `skip_policy_checks=false` (default), `checkInputs` derefs nil → rejects a valid tx with `bad-txns-inputs-too-large`. → require `SkipPolicyChecks` in the guard, or make `checkInputs` tolerate the un-extended path.
- **`services/validator/Validator.go:718` — validator honors `OutpointOnlySpend` without a SQL-store precondition.** The SQL-only guard lives only in the callers, not in the validator that consumes the flag. An Aerospike-backed validator skips parent reads → un-extended tx → `Spend` builds via `UTXOHashFromInput` → errors on nil locking script → valid tx rejected. → add `IsSQLUtxoStore` to the validator's own guard.
- **`model/Block.go:706` — coinbase inflation check skipped for all below-checkpoint blocks, ungated by the flag.** Returns `nil` for every block at/below the highest checkpoint regardless of `OutpointOnlyBelowCheckpoint`, contradicting the "default-off = current behaviour" description. Safe only if checkpoint-ancestry is enforced on every path into `Block.Valid`. → confirm that invariant and fix the wording, or gate the skip on the flag.
- **`services/validator/Validator.go:1771` — BIP68 relative-locktime check skipped on the outpoint-only path.** Same "rests on checkpoint ancestry" concern as the coinbase skip; BIP68 is active pre-Genesis. → confirm certified ancestry on every path reaching this skip.
- **`services/validator/Validator.go:832` — `saveAsConflicting` recreate drops `WithSkipExtendedInputs`.** If a caller ever combines `WithOutpointOnlySpend(true)` + `CreateConflicting(true)`, `GetFees` runs on un-decorated inputs → hard error stalls catchup. Latent today, unguarded. → forward `WithSkipExtendedInputs` or assert the combo is rejected.
- **`stores/utxo/sql/sql.go:2252` & `:2750` — UTXO-hash integrity check disabled under `skipUTXOHashCheck`.** Spends resolve by outpoint alone; also drops the corruption/bit-rot checksum against the stored row. Intended trade below a trusted checkpoint — please confirm and document as the explicit security boundary.

**P2 — plausible / lower blast radius**

- **`services/blockvalidation/BlockValidation.go:303`** — missing `highest == 0` guard (present in legacy `quickValidationAllowed`) → fast path engages on genesis on checkpoint-less networks (regtest). → add the guard.
- **`stores/utxo/sql/sql.go:889`** — minimal create persists `previous_tx_satoshis=0` / `previous_tx_script=NULL`; reading the extended tx back reconstructs a malformed tx (zero fee / false script result). Real only if below-checkpoint extended txs are read back — worth confirming they aren't.
- **`services/validator/Validator.go:1488`** — validator's checkpoint boundary (`v.settings.ChainCfgParams`) can diverge from netsync's (`sm.chainParams`) under fleet/config skew → processing error wedges legacy sync for a height range. Fail-closed. → align the source.

model/Block.go
freemans13: Both points addressed.

**Point 2 (store-support gate).** `checkBlockRewardAndFees` now takes `storeSupportsOutpointOnly`, computed at the `Block.Valid` call site from `txMetaStore.SupportsOutpointOnlySpend()`. The skip is retained only on stores that can actually produce fee=0 blocks; Aerospike / the unconfigured default keep full no-inflation enforcement below the checkpoint exactly as before this feature — no more silent disable where the fast path is a no-op. Commit 847d7a2.

**Point 1 (must not depend on live config).** Unchanged and intentional: the skip is *not* gated on `OutpointOnlyBelowCheckpoint`, so a fee=0 block written by the fast path still revalidates after the operator turns the flag off. Your original suggestion was to swap the flag gate for a call-site predicate rather than drop the gate entirely — that swap is what commit 847d7a2 does.

Going further than the store gate, the skip is now also gated on a caller-supplied confirmed-checkpoint-ancestor predicate (`BlockValidation.checkpointConfirmedAncestor`, threaded via `Block.SetCheckpointConfirmedAncestor`) — Option A for @ctnguyen's ChiR1, commit 0ed3eaa.

Net skip condition: `height <= highest hardcoded checkpoint` AND store supports the fast path AND the block is a confirmed ancestor of the pinned checkpoint. Fail-safe: above the checkpoint, on any lookup error, or when a caller doesn't set the predicate, the check runs.

model/Block.go
```diff
@@ -639,6 +639,26 @@ func (b *Block) checkBlockRewardAndFees(params *chaincfg.Params) error {
...
+	var highestCheckpoint uint32
+	for _, cp := range params.Checkpoints {
+		if cp.Height < 0 {
+			continue
+		}
+		if h := uint32(cp.Height); h > highestCheckpoint {
+			highestCheckpoint = h
+		}
+	}
+	if b.Height <= highestCheckpoint {
```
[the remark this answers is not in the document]
freemans13: Resolved across commits 847d7a2 + 0ed3eaa. The no-inflation skip is no longer height-only. It now fires only when **all three** conditions hold:

1. `height <= HighestCheckpointHeight` (highest hardcoded checkpoint — single source of truth, invariant I3),
2. the store can produce fee=0 fast-path blocks (`storeSupportsOutpointOnly`, from `txMetaStore.SupportsOutpointOnlySpend()` at the `Block.Valid` call site) — so Aerospike / the default keep full enforcement, and
3. the block is a confirmed checkpoint ancestor (`BlockValidation.checkpointConfirmedAncestor`, threaded via `Block.SetCheckpointConfirmedAncestor`): the main chain has reached the checkpoint height with the pinned hash AND the block is the main-chain block at its own height.

It stays ungated by `OutpointOnlyBelowCheckpoint` (a flag-gated skip false-rejects a genuinely-valid fee=0 block on `reconsiderblock`/`RevalidateBlock` after the flag is turned off; the read side can't distinguish a fast-path fee=0 block from a default one).

Condition 3 closes the two windows a height-only skip left open: the forward/optimistic-checkpoint window and a detached fork block reconsidered via `reconsiderblock` — both fail the predicate, so the check runs (correct, since non-fast-path blocks carry real fees). Fail-safe: above the checkpoint it short-circuits before any query; any lookup error/ambiguity → false → check runs.

Tests: `TestBlockValidation_checkpointConfirmedAncestor` (confirmed ancestor, forward window, detached fork, above-checkpoint no-query, empty checkpoints, fail-safe-on-error) and the three-condition gate in `TestCheckBlockRewardAndFees_*`.

freemans13:
@ctnguyen re **ChiR1** — addressed via **Option A** (code enforcement), commits 847d7a2 + 0ed3eaa.

The skip now requires a caller-supplied confirmed-checkpoint-ancestor predicate (`BlockValidation.checkpointConfirmedAncestor`, threaded to `model` via `Block.SetCheckpointConfirmedAncestor`) *in addition to* store fast-path support and the height bound. Both windows you flagged are now closed in code, not by comment:

- **Forward/optimistic-checkpoint window** — the predicate requires the main chain to have already reached the checkpoint height with the pinned hash; until then it is false, so the check runs.
- **Detached `reconsiderblock`** — the predicate requires the block to be the main-chain block at its own height; a fork block fails it, so the check runs.

In both cases running the check is correct because non-fast-path blocks carry real fees. The predicate is fail-safe: above the checkpoint it short-circuits before any blockchain query, and any lookup error/ambiguity (or a caller that never sets it) yields false → full enforcement.

Net skip condition: `height <= highest hardcoded checkpoint` AND store supports the fast path AND confirmed checkpoint ancestor.

Tests: `TestBlockValidation_checkpointConfirmedAncestor` + the three-condition gate in `TestCheckBlockRewardAndFees_*`. Also added `model.HighestCheckpointHash` alongside `HighestCheckpointHeight` so the pinned-hash and boundary-height derivations share one checkpoint list.

One correction from an earlier round of my own reasoning: I had briefly argued Option A couldn't work because it would break forward sync. That was wrong — the quick fast path never calls `Block.Valid`/`checkBlockRewardAndFees`, so this check only runs on full-validation and `reconsiderblock`, where the ancestry predicate is both available and correct.

**ChiR3** (deferred multi-block / cross-checkpoint / reconsiderblock e2e) remains open and tracked as a release gate before the flag is enabled on any shared node.

ctnguyen:
Review at commit [0ed3eaa9](https://github.com/bsv-blockchain/teranode/commit/0ed3eaa9c3638b21afd70232a6a22e7e55255069)

Since the last pass the author implemented **ChiR1 Option A**: the below-checkpoint coinbase no-inflation skip is no longer ungated/comment-justified — it is now gated in code on a caller-supplied ancestry predicate (`checkpointConfirmedAncestor`) threaded through `Block.SetCheckpointConfirmedAncestor`, plus store fast-path support. That closes the forward-checkpoint and detached-fork windows the earlier reviews were concerned about. The mechanism is sound; one implementation detail in the new predicate is worth hardening before the flag is enabled anywhere.

## What's done well

This is a strong iteration on already well-built work. ChiR1 was addressed the harder-but-better way (Option A: prove ancestry, rather than accept-by-sign-off), and the new `checkpointConfirmedAncestor` is genuinely fail-safe by construction — it early-returns `false` above the checkpoint (no DB cost in steady state), in the forward-checkpoint window, and on any lookup error, so a mis-wire degrades to full enforcement rather than to a skipped check. The predicate is threaded into **all four** `Block.Valid()` call sites (the three in `BlockValidation.go` plus `Server.go` `ValidateBlock`), and block assembly passes a nil store so the skip can never fire there. ChiR2/ChiR4/ChiR5 are all cleanly resolved (single-sourced `model.HighestCheckpointHeight`, in-test store capability, one shared `txMetaDataFromTx` builder), and the new `checkpointConfirmedAncestor` unit test covers the confirmed-ancestor, forward-window, detached-fork, above-checkpoint, no-checkpoints, and lookup-error branches.

## Status of previous points

ChiR1 is now **FIXED** via Option A — but the specific lookup it relies on has a distinct, narrower weakness carried forward as the new **ChiR6** below (not a reopening of ChiR1). ChiR3 remains a release gate: the cross-checkpoint / reconsiderblock e2e tests are still `t.Skip` placeholders (`TestCheckpointBoundary_B1/B3/B4/I2_Deferred`) — acceptable for a default-off merge, but they must land and pass before the flag is enabled on any shared node.

| ChiR1 | No-inflation skip ungated below checkpoint | YES | FIXED |
| ChiR2 | Duplicated checkpoint-height loop | | FIXED |
| ChiR3 | Cross-checkpoint / reconsiderblock e2e coverage | | NOT FIXED |
| ChiR4 | GateBounds relied on ambient SQL context | | FIXED |
| ChiR5 | `TxMetaDataFromTxNoFee` duplication | | FIXED |

## Non-blocking issues

### ChiR6 — `checkpointConfirmedAncestor` resolves the main-chain block at a height via a fork-returning API, then trusts index `[0]`
**Problem:** In `services/blockvalidation/BlockValidation.go`, `checkpointConfirmedAncestor` proves both "the chain reached the pinned checkpoint" and "block `b` is the main-chain block at its height" using `GetBlockHeadersFromHeight(ctx, h, 1)` and taking element `[0]`:

```go
bHeaders, _, err := u.blockchainClient.GetBlockHeadersFromHeight(ctx, b.Height, 1)
if err != nil || len(bHeaders) == 0 {
    return false
}
return bHeaders[0].Hash().IsEqual(b.Header.Hash())
```

But `GetBlockHeadersFromHeight` is explicitly documented (and implemented — `stores/blockchain/sql/GetBlockHeadersFromHeight.go`, `WHERE height >= $1 AND height < $2 ORDER BY height DESC`) to return **all** headers at that height *including competing forks*, with no "main-chain first" tiebreak among same-height rows. So `[0]` is whichever fork the DB happens to return first, not necessarily the main-chain block. Two failure modes result, both requiring a stored fork at a below-checkpoint height:
- **False rejection:** a genuine main-chain `fee=0` block reconsidered via `reconsiderblock`/`RevalidateBlock`, if a fork exists at its height and sorts ahead, yields `false` → the skip is missed → `checkBlockRewardAndFees` recomputes `coinbaseOutput > subtreeFees(0)+subsidy` and rejects a valid, checkpoint-pinned block as `BLOCK_INVALID`.
- **Skip on a non-ancestor:** a below-checkpoint fork block that sorts to `[0]` yields `true` → the no-inflation skip fires on a block that is *not* an ancestor of the pinned checkpoint — the exact "detached fork reconsidered" case the function's own comment claims to exclude.

**Why it matters:** This is newly-added consensus-adjacent code (post the last approval, so unreviewed by the earlier panel), and it gates whether a consensus check runs. Real-world probability is low — below-checkpoint forks are normally rejected at header validation and never stored, and the feature is default-off — but the predicate's correctness currently rests on that implicit "one block per below-checkpoint height" assumption rather than on the API contract. The unit test masks the gap by mocking `GetBlockHeadersFromHeight` to return a single header, which does not reflect the real multi-fork return.

**Fix:** Prove main-chain ancestry with an API that follows the chain, not one that returns forks. `GetHashOfAncestorBlock` from the pinned checkpoint hash is the most direct — it walks the main chain back from the checkpoint and is robust to forks, and it also subsumes the "chain reached the checkpoint" step (it fails if the pinned block isn't on-chain yet):

```go
// b is a confirmed ancestor iff walking back from the pinned checkpoint hash
// reaches b's hash at b.Height.
anc, err := u.blockchainClient.GetHashOfAncestorBlock(ctx, pinned, int(highest-b.Height))
if err != nil || anc == nil {
    return false
}
return anc.IsEqual(b.Header.Hash())
```

(`GetBlockByHeight`, which returns the main-chain block at a height, is an equivalent alternative.) Also update `TestBlockValidation_checkpointConfirmedAncestor` to model a multi-header return at the same height (b plus a fork) so the ordering assumption is actually exercised.

### ChiR7 — Feature-on nodes serve fast-path-degraded historical data (out of this diff)
**Problem:** When `OutpointOnlyBelowCheckpoint` is enabled, below-checkpoint subtrees carry `fee=0` and minimal-created inputs zero the parent script/satoshis columns. Asset-API read paths that surface this data for historical below-checkpoint transactions (`GetSubtree`/`GetBlockSubtrees` returning `fee=0`; `GetTransaction` serving `ExtendedBytes()` with zeroed input satoshis/empty prev-scripts) therefore return degraded values under the flag. The asset service is not modified by this PR, so this is a consequence of enabling the feature, not a defect in the diff.
**Why it matters:** Consumers of the asset API (explorers, downstream tooling) would see zeroed fees/prev-outs for old blocks on a flag-on node. It is consensus-safe and feature-on-only, but a surprise for operators who enable the flag on a serving node.
**Fix:** Either re-derive the input parent data from the persisted output rows in the read path, or explicitly document these endpoints as unsupported/degraded for below-checkpoint data while the flag is on. Track it alongside the ChiR3 e2e work as part of the pre-enable rollout gate.

## Recap

| ChiR6 | Ancestry predicate trusts fork-returning API's `[0]` | | 45% |
| ChiR7 | Asset API serves degraded below-checkpoint data under flag | | 15% |

PR #1181, opened 2026-06-30
docs: correct maxReorgDepth, bloom-filter, and optimistic-mining claims

### LINKED ISSUE #1166: [DOCS] Correct architecture docs: maxReorgDepth, bloom-filter, optimistic-mining default
**Severity:** low · **Confidence:** high · **Pass:** documentation
**Component:** `docs/` + `.claude/skills/debug-teranode` architecture reference

### Summary
Three architecture-doc claims were verified against code to be inaccurate during the system review.

### Items
1. **`blockchain_maxReorgDepth` (default 6) does not exist.** Deep-reorg protection is `CoinbaseMaturity` fork-depth + `SecretMiningThreshold` in the catchup path only; tip selection is uncapped most-chainwork. (See TN-007.)
2. **Already-mined / re-presentation detection uses exact block-ID set membership + an RPC fallback (`CheckBlockIsAncestorOfBlock`), not bloom filters** as the docs state. No bloom-filter false-positive/negative risk in this path.
3. **`OptimisticMining` defaults to `true`** (`settings/settings.go:343`, `settings/blockvalidation_settings.go:27`) — add-before-validate is ON by default; docs describe it as experimental/disabled.

### Suggested fix
Update `docs/topics/**` (and the debug-teranode skill architecture reference) to reflect the verified behaviour.

### Suggested tests
N/A (documentation). Optionally add a doc-lint check tying the documented defaults to the `settings/` struct tags.

---

ordishs:
Closes #1166

Corrects three architecture-doc claims that were verified against the code to be inaccurate.

### 1. `blockchain_maxReorgDepth` (default 6) does not exist
Removed the non-existent `blockchain_maxReorgDepth` setting — and the neighbouring fabricated `blockchain_chainTipCacheSize` / `blockchain_forkPointCacheSize` in the same config block. Documented the real behaviour: tip selection is uncapped most-cumulative-chainwork; deep-reorg / secret-mining protection is enforced in the Block Validation catchup path via `blockvalidation_secret_mining_threshold` (default = coinbase maturity − 1, i.e. 99) plus coinbase-maturity fork depth (`services/blockvalidation/catchup.go`).

### 2. Already-mined / re-presentation detection does not use bloom filters
The mechanism is **exact block-ID set membership** (current-chain header IDs from `GetBlockHeaderIDs`) with a `CheckBlockIsInCurrentChain` RPC fallback (`model/Block.go` `validOrderAndBlessed` / `checkParentExistsOnChain`, `services/blockvalidation/BlockValidation.go` `checkOldBlockIDs`). Because membership is exact there is no false-positive/negative risk.
- Removed the fabricated `recentBlocksBloomFilters` / `bloomFilterRetentionSize` / `blockBloomFiltersBeingCreated` / `bloomFilterStats` struct fields (and synced the struct to the real one).
- Removed the `bloomStats *model.BloomStats` parameter from the documented `ValidateBlock` signature.
- Rewrote the "Bloom Filter Implementation" sections and removed the orphaned `bloom_filter_validation` diagram.

### 3. `OptimisticMining` defaults to `true`
`blockvalidation_optimistic_mining` defaults to `true` (`settings/settings.go:343`, `settings/blockvalidation_settings.go:27`) — add-before-validate is ON by default. Corrected the "experimental / disabled by default" wording and resolved the related flagged ambiguity in the debug-teranode skill reference.

### Files changed
- `docs/topics/services/blockchain.md`
- `docs/topics/services/blockValidation.md`
- `docs/references/services/blockvalidation_reference.md`
- `.claude/skills/debug-teranode/references/architecture.md`
- removed `docs/topics/services/img/plantuml/blockvalidation/bloom_filter_validation.{puml,svg}`

### Verification
Documentation-only change (no Go code touched). Each cited setting key, struct field, and function name was confirmed to exist in the code; grep confirms no false claims remain. No local docs-site build available (`mkdocs`/`markdownlint` not installed), so the rendered site was not built.


sugh01 (approved):

ctnguyen (changes requested):
### In-memory chain-check description contradicts the store implementation

**Observed.** The new "Chain-Membership Check" section in `docs/topics/services/blockValidation.md` (§2.2.6) introduces this bullet:

> - **In-memory chain-check route**: when `blockchain_use_in_memory_chain_check` is enabled, every decision is deferred to `CheckBlockIsInCurrentChain`, which the store answers from its in-memory off-chain set without a recursive SQL query.

The first half (Block Validation deferring every decision to `CheckBlockIsInCurrentChain`) is correct, but the clause *"which the store answers from its in-memory off-chain set without a recursive SQL query"* is contradicted by the store implementation in `stores/blockchain/sql/CheckBlockIsInCurrentChain.go`:

- In the in-memory branch the store applies `maxBlockID` as an **in-memory upper-bound reject** for allocated-but-uncommitted / too-new IDs (`if id <= maxID`, dropping anything above the highest committed ID); the surviving committed candidates are then confirmed via SQL — `return s.checkBlockIsInCurrentChainSQL(ctx, candidates)`. The off-chain set is *not* consulted as a negative short-circuit here; the code comment is explicit: *"We deliberately do NOT use the in-memory off-chain set as a negative short-circuit."*
- `checkBlockIsInCurrentChainSQL` issues an `on_main_chain = true` SQL query and, on the about-to-reject path (and while the on-chain set is rebuilding, or when `maxBlockID` is uninitialised), **falls through to the recursive `parent_id` CTE** (`checkBlockIsInCurrentChainCTE`).

So the decision is confirmed in SQL against the `on_main_chain` flag — not answered "from the off-chain set" — and a recursive SQL CTE *can* run on this path.

**Why it matters.** This PR's entire purpose is to make the docs match the code; an introduced claim that contradicts the authoritative implementation re-creates exactly the class of inaccuracy the PR removes elsewhere. It would mislead an operator into believing the flag eliminates SQL/CTE work on the chain-membership path, which is false and security-relevant (the CTE on the reject path is precisely what prevents a transient `on_main_chain` flag from causing a false-negative, permanent block invalidation).

**Fix.** Narrow the bullet to what the code does, e.g.:

> - **In-memory chain-check route**: when `blockchain_use_in_memory_chain_check` is enabled, Block Validation holds no local prefetched ID set and defers every distinct parent-block-ID set to the authoritative `CheckBlockIsInCurrentChain`. The store applies `maxBlockID` as an in-memory upper-bound reject for uncommitted / too-new IDs, then confirms the remaining committed candidates in SQL — an `on_main_chain` flag query, with the recursive `parent_id` CTE as the fallback on the about-to-reject path and while the on-chain set is rebuilding.

ctnguyen (approved):

PR #1183, opened 2026-06-30
style(diag): use [[ ]] instead of [ ] in teranode-diag shell scripts

liam:
## Summary

Clears all **144** SonarCloud `shelldre:S7688` findings - *"Use '[[' instead of '[' for conditional tests. The '[[' construct is safer and more feature-rich."* They are confined to the two diagnostic helper scripts:

- `.claude/skills/debug-teranode-k8s/scripts/teranode-diag.sh` - 88 conversions
- `.claude/skills/debug-teranode-docker/scripts/teranode-diag-docker.sh` - 28 conversions

Both scripts are `#!/usr/bin/env bash`, so the bash-keyword `[[ ]]` is safe to use.

## What changed

A mechanical bracket-token swap: every single-bracket test command `[ ... ]` becomes `[[ ... ]]`. Operands, quoting, and operators are untouched (116 lines, 1:1).

Deliberately left alone:
- glob brackets (`[0-9]`)
- array subscripts (`${a[i]}`)
- pre-existing `[[ ]]` / `=~` matches
- string literals such as `"=[$BLASTER_CTRS]"`

## Why it is behaviour-preserving

Every flagged test uses quoted operands and standard operators (`-z -n -f -d -s -e -eq -ne -lt -gt`, and `=` with a quoted RHS). That avoids the only semantic differences between `[ ]` and `[[ ]]`:
- no unquoted RHS on `=`/`!=` (which would become glob matching under `[[ ]]`)
- no `-a`/`-o` (unsupported in `[[ ]]`)
- no escaped `\<`/`\>`

## Testing

- `bash -n` passes on both files (syntax valid, brackets balanced).
- Post-conversion scan: zero single-bracket test tokens remain.
- The full 116-line diff was reviewed line by line; every change is a real test command.

(No `shellcheck` was available in the environment; `bash -n` was the syntax gate.)

## Scope note

These are diagnostic helper scripts under `.claude/skills/`, committed in the repo and therefore Sonar-scanned. This is dev tooling, not production node code.

### DIFF
--- a/.claude/skills/debug-teranode-docker/scripts/teranode-diag-docker.sh
+++ b/.claude/skills/debug-teranode-docker/scripts/teranode-diag-docker.sh
@@ -58,11 +58,11 @@ subheader() { echo ""; echo "── $1 ──"; }
 
 # ── Discovery ────────────────────────────────────────────────────────────────
 # Compose project: the project label most common among containers running a teranode image.
-if [ -z "$COMPOSE_PROJECT" ]; then
+if [[ -z "$COMPOSE_PROJECT" ]]; then
     COMPOSE_PROJECT=$(docker ps --format '{{.Label "com.docker.compose.project"}}\t{{.Image}}' 2>/dev/null \
         | grep -iE 'teranode' | awk -F'\t' '{print $1}' | grep -v '^$' | sort | uniq -c | sort -rn | awk 'NR==1{print $2}')
 fi
-if [ -z "$COMPOSE_PROJECT" ]; then
+if [[ -z "$COMPOSE_PROJECT" ]]; then
     echo "No teranode compose project detected. Is the stack running? (docker compose ls)" >&2
     docker compose ls 2>/dev/null
     exit 1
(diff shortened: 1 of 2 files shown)


teracoder (approved):

ordishs (approved):
Clean mechanical `[ ]` → `[[ ]]` conversion. Verified the conversion is behavior-preserving for all the `-z`/`-n`/`-f`/`-d`/`-s`/`-e`/`-eq` tests and quoted-RHS `=` comparisons, and the PR correctly avoids the unquoted-RHS glob and `-a`/`-o` traps. Neither script runs under `set -e`, so no errexit surprises.

One low-severity nuance worth noting (not blocking): the three arithmetic comparisons paired with a trailing `2>/dev/null` in teranode-diag.sh (`CACHE_INT -lt 50`, `NR_T -gt 1000`, `IPC_INT -lt 60`) behave differently on a bare-alphabetic operand — `[ ]` errored and skipped, `[[ ]]` coerces the string to 0 in arithmetic context and can emit a spurious hint. Inputs are numeric in practice, so this is fine to accept as-is.

PR #1184, opened 2026-06-30
fix(scripts): shell hygiene — errors to stderr + [[ ]] conditionals

liam:
## Summary

Two related SonarCloud shell-lint cleanups across a set of dev/ops scripts. Two commits:

### 1. Redirect error messages to stderr (`shelldre:S7677`, 9 findings)
Error messages were printed to **stdout**, which pollutes output when a script's stdout is piped or captured. Append `>&2` to each flagged error echo in: `generate-blocks.sh`, `aerospike-asmt-wrapper.sh` (`log_error()`), `clean_datadir.sh` (x2), `kafka.sh`, `monitoring.sh`, `nginx-cache.sh` (x2), `run-teranode-with-logrotate.sh`. Only Sonar-flagged error lines are redirected; trailing usage/instruction text Sonar treats as informational is left on stdout.

### 2. Use `[[ ]]` instead of `[ ]` (`shelldre:S7688`, remaining 18 findings)
The bulk of S7688 (144, in the two diag scripts) was fixed in #1183. This clears the remaining 18 in `dev-stack-macos.sh` (11), `clean_datadir.sh` (6), `nginx-cache.sh` (1). All three are bash. Mechanical bracket-token swap only - operands and quoting untouched; quoted literals like `"[]"`, array lengths like `${#pids[@]}`, globs, and pre-existing `[[ ]]` are preserved. The single unquoted RHS (`clean_datadir.sh`: `= false`) has no glob metacharacters, so `[[ ]]` matches it identically.

## Why both rules in one PR

`clean_datadir.sh` and `nginx-cache.sh` are touched by both fixes. Keeping them in one PR avoids merge conflicts between two branches editing the same files on adjacent lines.

## Testing

`bash -n` passes on every changed file. Post-conversion scan: zero single-bracket test tokens remain in the three S7688 files. Diffs reviewed line by line.

### DIFF
--- a/compose/scripts/generate-blocks.sh
+++ b/compose/scripts/generate-blocks.sh
@@ -53,7 +53,7 @@ while [ "$#" -gt 0 ]; do
 done
 
 if [ -z "$HOST" ]; then
-  echo "Error: --host is a required argument."
+  echo "Error: --host is a required argument." >&2
   echo "Usage: $0 --host <host> [--port <port>] [--stopHeight <height>] [--generate] [--initialBlocks <count>]"
   exit 1
 fi
--- a/deploy/docker/base/aerospike-asmt-wrapper.sh
+++ b/deploy/docker/base/aerospike-asmt-wrapper.sh
@@ -25,7 +25,7 @@ log_warn() {
 }
 
 log_error() {
-    echo "[$(date '+%Y-%m-%d %H:%M:%S')] ❌ ERROR: $*"
+    echo "[$(date '+%Y-%m-%d %H:%M:%S')] ❌ ERROR: $*" >&2
 }
 
 # Install ASMT tool if not present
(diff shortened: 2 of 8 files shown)


ordishs (approved):
Clean, well-scoped change. Each `>&2` is appended to a plain echo/log_error line with no logic touched, and the deliberate split (leaving usage/instruction text on stdout) is documented and defensible. Resolves the 9 S7677 findings without side effects. LGTM.

freemans13 (approved):
Clean, mechanical change — 9 lines, 1:1 with the SonarCloud `shelldre:S7677` findings. Verified the only reusable helper touched (`log_error()` in aerospike-asmt-wrapper.sh) is never used in a command-substitution capture, so moving it to stderr has no behavioral side effect. Remaining changes are inline error echoes followed by `exit 1` or by usage/instruction text intentionally left on stdout per the documented scope. LGTM.

PR #1186, opened 2026-06-30
fix(utxo): stamp setMined delete-at-height from the mined block height

icellan:
## Problem

On the scaling environment, transactions were disappearing from the UTXO store **before** the configured retention window (`utxostore_blockHeightRetention`, default 288 blocks) had elapsed.

## Root cause

The pruner is correct — it deletes records whose `deleteAtHeight (DAH) <= currentHeight`. The bug is in how the DAH is **stamped** during `setMined`.

Both the Aerospike and SQL UTXO stores computed the DAH during `setMined` from the store's **cached chain tip** (`s.blockHeight + 1`) instead of the height of the block the transaction is actually mined into (`minedBlockInfo.BlockHeight`):

- aerospike: `set_mined.go`, `set_mined_expressions.go`, `teranode.lua` (`setMined` UDF)
- sql: `setMinedMultiChunk` in `sql.go`

`s.blockHeight` is refreshed asynchronously by a blockchain-notification subscription (`stores/utxo/factory/utxo.go`), so during catchup/sync and under high block throughput it lags behind the block being validated. When a transaction that was spent while still unmined is later marked mined, `setMined` stamps the DAH. Using the lagging cached height stamped it far too low — once the lag reached the retention window the DAH was already `<= currentHeight`, so the pruner deleted the record almost immediately.

The spend path was already correct (it uses the per-call block height). Block validation already passes the correct height in `minedBlockInfo.BlockHeight` (it is even used to populate the `blockHeights` bin/column), so the DAH is now computed from it. The `unset` and `unmined_since` paths intentionally keep using the current height, which is correct there.

## Fix

Compute the setMined DAH from `minedBlockInfo.BlockHeight + retention` in both backends:

- **aerospike** — master DAH (Lua UDF + filter-expression path) and child/pagination-record DAH
- **sql** — `setMinedMultiChunk`

## Tests

New regression tests for both backends (`set_mined_dah_test.go`) that reproduce the catchup condition (cached tip far below the mined block height) and assert `DAH == minedBlockInfo.BlockHeight + retention`:

- SQLite (in-memory)
- Aerospike (testcontainers) — covers both the Lua and filter-expression paths, and both the master and a pagination record

Updated one existing Aerospike test (`aerospike_spend_all_and_expire`) that had encoded the buggy value.

Verified locally: `go build ./...`, `go vet`, SQL tests (incl. `-race`), Aerospike integration tests (`TestAerospike`, setMined/DAH suites).

## Notes

- The Teraslab store is out of scope (it owns the DAH lifecycle server-side); the scaling environment uses Aerospike.
- The Aerospike Lua UDF changed, so it is re-registered from `teranode.lua` on startup.

### DIFF
--- a/stores/utxo/aerospike/aerospike.go
+++ b/stores/utxo/aerospike/aerospike.go
@@ -518,10 +518,24 @@ func (s *Store) GetBlockHeight() uint32 {
 	return s.blockHeight.Load()
 }
 
+// effectiveBlockHeight resolves the block height to use for DAH computation,
+// falling back to the store's current cached block height when the caller did not
+// supply one (blockHeight == 0). Paths that know the operation's block height (the
+// mined block height, or the spend's block height) pass it; paths that don't get
+// the current block height, preserving the historical behaviour.
+func (s *Store) effectiveBlockHeight(blockHeight uint32) uint32 {
+	if blockHeight == 0 {
+		return s.blockHeight.Load()
+	}
+
+	return blockHeight
+}
+
 // deleteAtHeightFor returns the delete-at-height (DAH) to stamp on a record that
 // becomes prunable at blockHeight — blockHeight + the configured retention window —
 // and whether retention is enabled. When retention is 0 ("don't use automatic
-// retention") the second return is false and no DAH should be set.
+// retention") the second return is false and no DAH should be set. A blockHeight of
+// 0 falls back to the current block height (see effectiveBlockHeight).
 //
 // This is the single source of truth for the Go DAH formula and the retention
 // guard, shared by the create, spend and setMined paths so they cannot drift. The
@@ -534,7 +548,7 @@ func (s *Store) deleteAtHeightFor(blockHeight uint32) (uint32, bool) {
 		return 0, false
 	}
 
-	return blockHeight + retention, true
+	return s.effectiveBlockHeight(blockHeight) + retention, true
 }
 
 func (s *Store) SetMedianBlockTime(medianTime uint32) error {
--- a/stores/utxo/aerospike/aerospike.go
+++ b/stores/utxo/aerospike/aerospike.go
@@ -518,6 +518,25 @@ func (s *Store) GetBlockHeight() uint32 {
 	return s.blockHeight.Load()
 }
 
+// deleteAtHeightFor returns the delete-at-height (DAH) to stamp on a record that
+// becomes prunable at blockHeight — blockHeight + the configured retention window —
+// and whether retention is enabled. When retention is 0 ("don't use automatic
+// retention") the second return is false and no DAH should be set.
+//
+// This is the single source of truth for the Go DAH formula and the retention
+// guard, shared by the create, spend and setMined paths so they cannot drift. The
+// Aerospike filter-expression path (buildDeleteAtHeightExpression) and the Lua
+// setDeleteAtHeight UDF compute the same blockHeight + retention and MUST be kept
+// in sync with this function.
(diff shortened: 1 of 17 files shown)


oskarszoon (approved):
Approve. The fix correctly closes the premature-pruning data-loss bug — stamping the delete-at-height from `minedBlockInfo.BlockHeight + retention` instead of the lagging cached tip, consistently across all four paths (Aerospike master/child/Lua + SQL). Verified: `minedBlockInfo.BlockHeight` is non-zero at every caller, the Go→Lua arg wiring is right (`blockHeight` for DAH, `currentBlockHeight` for `unmined_since`), master and child DAHs match, the dropped `+1` was pure lag-compensation, and the Lua UDF re-registers on existing clusters via the `teranode_v60→v61` name bump. Reorg-safe via the existing monotonic bump-forward guard.

The SQL spend path (`sql.go:2298`) looks like it still uses the lagging tip, but it can't actually defeat the fix: while the tx is unmined the spend writes no DAH (its `dah_upd` CTE is gated on `unmined_since IS NULL`), so setMined stamps the correct value; and once mined, the bump-forward guard (`delete_at_height IS NULL OR < newDAH`) refuses to lower the already-correct DAH. Only a narrow, pre-existing write-skew race remains (spend's snapshot sees NULL before setMined commits) — not introduced here and not the primary scenario.

Non-blocking follow-ups:
- **Dead code carrying the old bug**: `handleSetMinedSignal` (`set_mined.go:541-570`) has zero callers but still contains `dahHeight := thisBlockHeight + retention` — delete it so it can't be wired back up with the bug.
- **Dead parameter**: after the fix, `thisBlockHeight` in the expressions path (`set_mined_expressions.go:254,341,361`) is passed and declared but never read (the path uses the server-written `state.NewDeleteAtHeight`). Remove it + the `s.blockHeight.Load()+1` line in `SetMinedMultiWithExpressions`.
- **Consistency**: `setDAH` (`sql.go:2874`) and `setUnlockedBulk` (`sql.go:4467`) still anchor on the lagging tip. They're guarded by the same bump-forward logic so not a live data-loss path, but worth anchoring on a per-call height too; and trim the now-stale "+1 because blockHeight lags" comments.
- **Deploy note**: `registerLuaIfNecessary` is skip-if-exists, so the fix reaching deployed clusters is entirely load-bearing on the `v61` name bump — worth calling out that a binary swap without a process restart would leave the old UDF in place.

ordishs (approved):
Approving — the fix is correct and well-tested.

**Verified:**
- All four DAH code paths (Aerospike Lua master UDF, filter-expression path, Go child/pagination records, and SQL `setMinedMultiChunk`) now compute the same value: `minedBlockInfo.BlockHeight + retention`, with no off-by-one or double-counted retention.
- Every production caller passes a real, non-zero mined height. In particular `block.Height` in the `setMined` path is guarded at `services/blockvalidation/Server.go:1299` (`if height <= 0 { return ... }`) and set from `previousBlockMeta.Height + 1`, so the "BlockHeight == 0 → immediate prune" case is not reachable from the real mined path.

**One non-blocking nit:**

`stores/utxo/aerospike/set_mined.go:554` — `handleSetMinedSignal` still computes `dahHeight := thisBlockHeight + retention`, the exact pre-fix formula this PR corrects everywhere else. It currently has **no callers** repo-wide, so it is not a live regression, but it's a latent landmine: if a future change ever routes the signal-handling path back through this helper, it would silently reinstate the early-pruning bug. Suggest either deleting the dead function or updating it to `minedBlockInfo.BlockHeight + retention` for consistency.

Issue #1189, opened 2026-06-30
legacy sync throughput: block ingestion is serialized (read-loop blocks on processing); download not overlapped


oskarszoon:
## Problem

Legacy catchup throughput is **I/O-wait-bound, not CPU-bound** — the legacy service
sits at ~2-3 of 48 cores while syncing. Two issues gate throughput:

**C1 — serialized block ingestion (download never overlaps processing).**
`OnBlock` (`peer_server.go:910`) runs on the **peer read-loop goroutine**, calls
`QueueBlock`, then blocks on `err = <-sp.blockProcessed` (`peer_server.go:982`) until
the block is fully processed. So the read loop cannot download block N+1 until N
finishes:
```
read(N) → process(N)  [read-loop blocked]  → download(N+1) ~4s → process(N+1) → …
```
The depth-100 `blockQueue` (`manager.go:2103`) therefore never fills — only one block
is ever in flight (`blockBacklog`≈0). The next-block download (~4s) is serialized
after each block's processing instead of overlapping it.

**C2 — per-block processing is Aerospike-round-trip bound.** Processing wall time is
dominated by sequential-ish UTXO batch round-trips; the legacy CPU is idle throughout.

## Evidence

Per-block timing from `HandleBlockDirect … DONE in` cadence:

| node | proc/block | wall cadence | inter-block gap | blk/hr |
|---|---|---|---|---|
| eu-2 (small-block era) | 6.9s | 11.0s | **4.1s (37% of wall)** | 326 |
| eu-3 (big-block era, bounded) | 23.4s | 28.1s | 4.6s (17%) | 128 |

CPU profile (eu-2): top frame `legacy/peer.(*Peer).readMessageStreaming` 19.5% cum
(block download), Aerospike reads only 5.6%; legacy at ~2/48 cores. So both the gap
(download) and the processing are I/O-wait, not CPU.

## Fix direction

- **C1 — bounded async prefetch.** Let the read loop fetch a small number of blocks
  ahead into `blockQueue` while processing, so download overlaps processing and the
  ~4s gap is hidden. **Cap by memory** — the synchronous wait is intentional
  backpressure (`blockBacklog` comment, `manager.go:423`) against buffering huge
  blocks (a 2.6M-tx block is large); bound the prefetch depth, don't remove the
  backpressure.
- **C2 — reduce per-block round-trips / per-block fixed cost.** Note the tension with
  the overload issue: simply raising processing concurrency is what overloads
[...]

PR #1190, opened 2026-06-30
perf(legacy): overlap block download with processing via bounded async prefetch

oskarszoon:
## Problem (C1)

Legacy catchup was I/O-wait-bound, not CPU-bound. `OnBlock` runs on the peer read-loop goroutine, called `QueueBlock`, then blocked on `err = <-sp.blockProcessed` until the block was fully processed. So the read loop could not download block N+1 until N finished:

```
read(N) → process(N)  [read-loop blocked]  → download(N+1) ~4s → process(N+1) → …
```

Only one block was ever in flight; the next-block download (~4s in the small-block era) was serialized after each block's processing instead of overlapping it. The CPU sat idle (~2-3 of 48 cores) the whole time.

## Fix — bounded async prefetch

`OnBlock` no longer blocks on per-block completion. It admits the block against a **global byte-weighted budget** and returns, so the read-loop downloads the next block while the current one validates in the background. Block processing stays single-threaded FIFO, so **Aerospike concurrency is unchanged** (per the issue's explicit constraint — this is not "more parallelism").

Key properties, each chosen to hold an invariant:

- **Memory cap (`legacy_blockPrefetchBufferBytes`, default 256 MiB).** A global `semaphore.Weighted` bounds the total serialized bytes of received-but-not-yet-processed blocks across *all* peers and streams. Admission happens *before* `QueueBlock`, so a block can't sit in the deep `msgChan`/`blockQueue` pinning its decode arena uncounted. A block larger than the whole budget is admitted alone (weight clamped), preserving the original one-block-at-a-time backpressure for huge blocks. `0` disables prefetch entirely → original synchronous path (kill switch).
- **No cross-peer wedge.** Each queued block gets its own buffered(1) reply channel, so the single shared block-processing goroutine never blocks on `reply <- err` even when a peer disconnects mid-flight. (A shared reply channel + drainer here deadlocks *all* peers — that approach was rejected.)
- **Malicious-peer backpressure preserved.** Unrequested blocks are rejected before admission (mirrors the existing `handleBlockMsg` check + regtest exception), so a misbehaving peer cannot flood the shared budget or starve the real sync peer.
- **Stall detection still works.** Under prefetch a non-empty backlog is the steady state, so the stall detector now skips on read-loop admission-backpressure (`blockPrefetchWaiters`) instead of on `blockBacklog > 0`; a slow/stalled sync peer is still rotated.
- **Ordering preserved.** Admission is on the single per-peer read-loop and the netsync block queue stays strictly FIFO.

## Verification

- `go build ./...` — clean
- `go vet`, `staticcheck`, `golangci-lint` (changed packages) — 0 issues
- `go test -race` on `services/legacy/netsync`, `services/legacy`, `settings` — pass
- New unit tests cover: byte-budget acquire/clamp/oversized/blocking + waiter accounting + ctx-cancel, the unrequested-block gate, the stall-detector backpressure logic, and the shared disconnect-on-failure decision.

Empirical blk/hr improvement needs a sync run against a real peer; the change targets the diagnosed read-loop gap directly.

## Scope

Addresses **C1** only. **C2** (reducing per-block Aerospike round-trips) is a separate, harder change and is left as follow-up.

Addresses #1189


services/legacy/peer_server.go
```diff
@@ -960,36 +960,128 @@ func (sp *serverPeer) OnBlock(_ *peer.Peer, msg *wire.MsgBlock, buf []byte) {
...
+		}
+
+		// Bounded async prefetch. Admit the block against the global byte budget
+		// FIRST — this blocks the read-loop only when in-flight blocks already
+		// fill the budget, which is the backpressure that bounds the memory
+		// pinned by buffered blocks (a malicious peer cannot outrun it, and an
+		// oversized block is admitted alone). Admitting before QueueBlock is
+		// essential: otherwise the block would sit in the deep msgChan/blockQueue
+		// pinning its decode arena without being counted against the budget.
+		weight, err := sm.AcquireBlockPrefetch(sp.ctx, int64(msg.SerializeSize()))
```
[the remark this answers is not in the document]
oskarszoon: Done in 5e298678b — weight by len(buf) on the single-stream path, falling back to SerializeSize() only when buf is nil (the BlockPriority streaming path), so we no longer walk the tx set on the read-loop just for a byte count.

services/legacy/peer_server.go
```diff
@@ -960,36 +964,128 @@ func (sp *serverPeer) OnBlock(_ *peer.Peer, msg *wire.MsgBlock, buf []byte) {
...
```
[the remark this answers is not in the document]
oskarszoon: Done in 5e298678b — AcquireBlockPrefetch now also aborts the budget wait on the peer quit channel, not just ctx cancellation, so a budget-parked read-loop unblocks on peer teardown (sp.ctx being the long-lived Init context). Mirrors awaitBlockResult; the linking goroutine only exists while blocked.

oskarszoon:
Tested with a Teratestnet benchmark, sync from 0 to block 20k purely using legacy service went down from 10816 seconds to 8759 seconds. Significant performance improvement

ordishs (changes requested):
## Review

Nicely built change — the byte-budget mechanics are careful: budget release is exactly-once across every teardown path, oversized blocks clamp to the full budget so `Acquire` can't deadlock, per-peer FIFO ordering is preserved by the single `blockQueue` goroutine, and the `AcquireBlockPrefetch` quit→cancel linking goroutine doesn't leak. The kill switch (`legacy_blockPrefetchBufferBytes=0`) is a good safety valve.

The concerns below are liveness/heuristic regressions that surface under real IBD load — the exact regime this PR targets. None are consensus-correctness or memory-safety bugs. One is a blocker; the rest are discussion.

### 🔴 Blocker — stall-detector suppression narrowed, can rotate a healthy sync peer

`handleCheckSyncPeer` (`services/legacy/netsync/manager.go:895`) changed its skip condition from "any backlog" (`blockBacklog > 0`) to "a read-loop is budget-blocked" (`localReadBackpressured()` → `blockPrefetchWaiters > 0`).

The gap: a handful of legitimately-requested blocks are queued, the **budget is not full** (so no waiter is registered), and one block is slow to validate. Throughput and last-block-time go stale, `blockPrefetchWaiters == 0`, so the stall check is **not** suppressed and rotates a perfectly healthy sync peer. Rotation then drops sync progress and the replacement peer hits the same local backlog.

`TestHandleCheckSyncPeer_PrefetchBackpressure` already demonstrates a stalled peer being rotated with `blockBacklog=3, waiters=0` — the same path misfires for a healthy peer whose blocks are simply slow to process locally.

One question settles it definitively from the code: **can a block be backlogged / mid-validation with `blockPrefetchWaiters == 0` and stale throughput?** If yes, the detector will rotate a healthy peer — please confirm or guard (e.g. keep `blockBacklog` as a secondary suppression signal, or account the `blockQueue <-` wait into the waiter counter).

### 🟡 Please answer — `maxBlockQueue` (100) is now the tightest bound below a ~4096-block admission window

`services/legacy/netsync/manager.go:2208`. Default budget 256 MiB ÷ `minInFlightBlockWeight` 64 KiB ⇒ up to ~4096 blocks admissible, but `maxBlockQueue` is still 100 (sized for the old one-block-per-peer world). When `handleBlockMsg` can't keep up, the ~101st block blocks the single `blockHandler` select loop on `blockQueue <-`; since that loop is the sole consumer of `msgChan`, `donePeerMsg` (disconnect), `getSyncPeerMsg` (rotation), `invMsg`, `headersMsg` and tx dispatch all stall for **every** peer, not just the busy one — cross-peer head-of-line blocking that was unreachable when only one block/peer was in flight. It also doesn't increment `blockPrefetchWaiters`, so the suppression above doesn't cover it.

Not necessarily a hard blocker (it's backpressure, not deadlock, and self-resolves as the queue drains), but worth a decision: scale `maxBlockQueue` with `blockPrefetchBudgetBytes/minInFlightBlockWeight`, or fold this wait into the waiter accounting? Fine as a follow-up if you'd rather scope it out — just call it.

### ⚪ Non-blocking notes

- **`awaitBlockResult` swallows the validation-failure disconnect on teardown** (`services/legacy/peer_server.go:1080`). On `<-sp.quit` / `<-sp.ctx.Done()` it returns without reading `done`, so a block that fails validation exactly as the peer tears down never reaches `DisconnectWithWarning`/misbehavior accounting. The old synchronous path disconnected before returning. Low severity (the connection is already going away), but it's a behavior delta.
- **Synchronous fallback lacks a teardown escape** (`services/legacy/peer_server.go:990`, pre-existing). The bare `err = <-sp.blockProcessed` has no `select` on `sp.quit`/`sp.ctx.Done()`; if the block-queue drain loses the shutdown race it blocks forever. Not introduced here, but the async path was given exactly this guard — worth applying for symmetry.

services/legacy/netsync/manager.go
```diff
@@ -861,15 +892,21 @@ func (sm *SyncManager) handleCheckSyncPeer() {
...
+	// failing them afterwards. The deferred updateNetwork still runs, keeping
+	// throughput samples fresh for the next tick.
+	//
+	// With prefetch enabled this is when read-loops are blocked acquiring budget;
+	// with prefetch disabled it is the original condition that any block is
+	// queued or mid-validation (OnBlock then blocks on blockProcessed). Crucially,
+	// under prefetch a non-empty backlog is the normal steady state and must NOT
+	// suppress stall detection: when budget is available we keep reading, so a
+	// slow or stalled sync peer is still detected and rotated.
+	if sm.localReadBackpressured() {
```
freemans13, line 961: **Stall detector can now wrongly rotate a healthy sync peer when a block is slow to *validate* (not slow to download).**

The old guard skipped stall checks whenever `blockBacklog > 0` (any block queued or mid-validation). `localReadBackpressured()` only trips on `blockPrefetchWaiters > 0`, i.e. when a read-loop is actually *blocked in `Acquire`* because the budget is full. But `updateLastBlockTime` fires on block *processing completion* (`handle_block.go:65`), not on receipt.

**Failure scenario:** peer delivers block N quickly; it's admitted via `TryAcquire` (budget far from full → `waiters == 0`) and N+1 is already received and queued. Block N takes >3 min to validate (large tx set / slow Aerospike). `blockBacklog` is 2 but `waiters == 0`, so this check runs. `lastBlockSince > maxLastBlockTime (3m)`, and since both blocks are already received no bytes flow, so `hasHealthyDownloadThroughput` doesn't rescue it → the healthy peer is rotated. The old `blockBacklog > 0` guard suppressed exactly this. Net effect: sync-peer churn precisely when local validation is slow — the worst time to drop it.

Consider keeping a `blockBacklog > 0` component in `localReadBackpressured()` (or gating the last-block-time violation on it), so slow *validation* still suppresses rotation while a genuinely idle read-loop does not.

services/legacy/peer_server.go
```diff
@@ -960,36 +964,139 @@ func (sp *serverPeer) OnBlock(_ *peer.Peer, msg *wire.MsgBlock, buf []byte) {
...
+		// if this peer has since disconnected and no goroutine is waiting. That
+		// is what stops a disconnected peer from wedging the single shared
+		// block-processing goroutine — and through it every other peer.
+		done := make(chan error, 1)
+		sm.QueueBlock(block, sp.Peer, done)
+
+		// Return immediately so the read-loop downloads the next block while this
+		// one is validated; the result (budget release + disconnect-on-failure)
+		// is handled off the read-loop.
+		go sp.awaitBlockResult(done, weight, blockHash)
```
freemans13, line 1112: **The "at most one bad block in flight" protection is relaxed to `budget / minInFlightBlockWeight` (≈4096 at defaults).**

The removed comment justified blocking the read-loop as limiting "how many bad blocks a malicious peer can queue before being disconnected." The `BlockRequested` gate preserves this only for *unrequested* blocks. For *requested* blocks that turn out invalid, `OnBlock` admits/queues each and returns; the first failure only disconnects from `awaitBlockResult` after that block reaches the front of the FIFO queue.

**Failure scenario:** a malicious sync peer (for which we've requested many blocks, e.g. headers-first) streams back-to-back invalid requested blocks. All of them (up to ~budget/64KiB) are admitted, queued, and fully validated one-by-one before the first failure disconnects the peer — wasted validation work, vs exactly 1 in the old path. Memory-safe and bounded, but a documented protection is materially weakened. Consider skipping/dropping still-queued blocks for a peer already marked for disconnect.

services/legacy/peer_server.go
```diff
@@ -960,36 +964,139 @@ func (sp *serverPeer) OnBlock(_ *peer.Peer, msg *wire.MsgBlock, buf []byte) {
...
+// arena) so the block's memory can be released while processing proceeds.
+func (sp *serverPeer) awaitBlockResult(done chan error, weight int64, blockHash *chainhash.Hash) {
+	// Release the reserved budget exactly once on every exit path.
+	defer sp.server.syncManager.ReleaseBlockPrefetch(weight)
+
+	var err error
+
+	select {
+	case err = <-done:
+	case <-sp.quit:
```
freemans13, line 1239: **Shutdown can strand `awaitBlockResult` goroutines and their reserved budget (teardown-only, low severity).**

The block-queue handler in `manager.go` (~2143-2160) exits on `<-sm.quit` without draining `blockQueue`, so queued messages' `reply` channels are never signaled. Those blocks' `awaitBlockResult` goroutines then rely on `sp.quit` / `sp.ctx.Done()` — and as the comment here notes, `sp.ctx` is *not* cancelled by `Stop()`.

**Failure scenario:** `SyncManager.Stop()` runs on a path where peer disconnection lags (peers not yet torn down). In-flight prefetched blocks' goroutines block until `sp.quit` eventually closes, holding their semaphore weight meanwhile. Benign at true process exit, but a real leak window on partial/graceful-drain shutdown ordering. Draining `blockQueue` with an error reply on `sm.quit` would close it cleanly.

services/legacy/netsync/manager_prefetch_test.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	t.Run("keeps sync peer while a read-loop is blocked on prefetch budget", func(t *testing.T) {
+		sp := &peerpkg.Peer{}
+		sm := newSyncManager(sp, newStalledState())
+
+		sm.blockPrefetchWaiters.Add(1) // read-loop backpressured by our own processing
+
+		require.NotPanics(t, func() { sm.handleCheckSyncPeer() })
+		assert.Equal(t, sp, sm.loadSyncPeer())
```
freemans13: **Uses `assert` where the repo convention mandates `require`.**

`.claude/rules/testing.md`: *"Use `require` from testify (not `assert`)."* This line and line 317 (`assert.Panics`) use `assert.`; the file already imports `require`. `assert` continues after a failed check, which diverges from the project's stated test-failure semantics. Swap to `require.Equal` / `require.Panics`.

freemans13 (commented):
Automated review of the bounded async block-prefetch change. The core mechanism (per-block reply channels, byte-weighted admission, requested-block gate, teardown release) checks out; ordering/loss/double-release all look correct. Four findings below, most severe first. A TOCTOU concern on `requestedBlocks` was investigated and refuted — `expiringmap.Stop()` only halts eviction, and the per-peer map has a 60-min TTL, so a queued block stays 'requested'.

services/legacy/netsync/manager.go
oskarszoon: Fixed in b6388b4cb. `localReadBackpressured` again suppresses on `blockBacklog > 0` (not only the budget waiter), so a block slow to *validate* — backlog>0, budget not full, waiters==0 — no longer lets a stale last-block-time rotate a healthy peer. A genuinely stalled peer still drains the backlog and is rotated. `TestHandleCheckSyncPeer_PrefetchBackpressure` now asserts a draining backlog keeps the peer, plus an idle-stall case that still rotates.

services/legacy/netsync/manager_prefetch_test.go
oskarszoon: Fixed in b6388b4cb — swapped to `require.Equal`/`require.Panics` and removed the now-unused `assert` import.

services/legacy/peer_server.go
oskarszoon: Fixed in 67a4c979c — the block-queue handler now drains remaining queued blocks with an error reply on `sm.quit`, so their `awaitBlockResult` goroutines exit promptly and release budget on shutdown. Note `awaitBlockResult` already selects on `sp.quit` and `sp.ctx.Done()` (the ServiceManager errgroup ctx, cancelled on shutdown), so this was a bounded window rather than a permanent leak — the drain just makes cleanup deterministic.

services/legacy/peer_server.go
oskarszoon: The exposure is much smaller than `budget/64KiB`: we only getdata up to `dynamicMax` (≤20, scaled by blockSizeTracker) blocks from the sync peer at a time, and `requestedBlocks` holds ~that many, so at most ~`dynamicMax` *distinct* requested-but-invalid blocks are validated before the first failure disconnects — not ~4096. The byte budget + 64 KiB weight floor bound memory and goroutine count regardless. I prototyped an explicit `!peer.Connected()` skip in handleBlockMsg but it drops blocks from bare peers (broke `TestHandleBlockMsg_OrphanDuringCatchup`) and risks false-dropping valid blocks, so I left the bounded, memory-safe behavior documented rather than add a fragile guard. Can add a disconnect-scoped skip if you would prefer it.

oskarszoon:
Thanks for the detailed review — all points addressed.

- 🔴 **Stall detector (blocker)** — fixed in `b6388b4cb`. `localReadBackpressured` again suppresses on any `blockBacklog > 0`, not only a budget waiter. To your definitive question: yes, a block can be mid-validation with `blockPrefetchWaiters == 0` and stale throughput — that is exactly the case now re-suppressed, so slow *validation* no longer rotates a healthy peer. A genuinely stalled peer still drains the (bounded) backlog and is then rotated — delayed, not prevented. `TestHandleCheckSyncPeer_PrefetchBackpressure` now asserts a draining backlog keeps the peer, with an added idle-stall case that still rotates.
- 🟡 **maxBlockQueue vs ~4096 admission window** — fixed in `b6388b4cb`. Chose to scale the queue: `maxBlockQueue = budget/minInFlightBlockWeight` when prefetch is enabled, so the admission ceiling can never exceed the queue depth and HOL-block the sole `msgChan` consumer. The queue holds only `*blockQueueMsg` pointers; the referenced blocks stay bounded in bytes by the budget, so the deeper queue doesn't raise the memory ceiling.
- ⚪ **Sync-fallback teardown escape** — fixed in `b6388b4cb`. The `<-sp.blockProcessed` wait now `select`s on `sp.quit`/`sp.ctx.Done()`, matching the async path.
- ⚪ **awaitBlockResult swallows validation-failure disconnect on teardown** — left as-is (behavior delta acknowledged). It only skips the disconnect when `sp.quit`/`sp.ctx` has already fired, i.e. the peer/process is already tearing down, so the connection is going away regardless; reading `done` there would risk blocking on a reply the drained queue never sends. Separately, `67a4c979c` now drains the block queue with an error reply on `sm.quit`, so on shutdown the reply generally does arrive and the goroutines release budget promptly.

The one open thread (requested-invalid-block flood) has a reply: exposure is bounded by the getdata window (`dynamicMax` ≤ 20), not the ~4096 byte ceiling — left the bounded/memory-safe behavior documented rather than add a guard that false-drops valid blocks. Happy to add a disconnect-scoped skip if you'd prefer.

oskarszoon:
Mechanism validation from a live Teratestnet run of this build (complements the wall-clock number above):

With DEBUG on the legacy service, `fetchHeaderBlocks` shows the getdata pipeline held **full** throughout sync:

```
[fetchHeaderBlocks] Header list: 87651 blocks, in-flight: 19/20, avg size: 190 bytes, requesting: 1 more
```

- **in-flight `19/20` on 1194 of 1195 sampled lines** (the single `0/20` is the first tick) — i.e. blocks download ahead while others validate, vs the pre-fix "only one block ever in flight, backlog ≈ 0".
- Dynamic window sat at `20` for these ~190-byte blocks (scales down for large blocks).
- Zero `Got unrequested block` disconnects, zero `block processing failed` — the admission gate and disconnect paths didn't misfire on healthy sync.

At ~190-byte blocks the wall-clock win is modest (download is near-instant regardless); the throughput gain shows once blocks are large enough for download time to matter, which the 0→20k benchmark above (10816s → 8759s) already reflects.

icellan:
Reviewed the async-prefetch change. No correctness or consensus bug — the design is sound in the happy path. But dropping the synchronous "one block in flight, fully processed before the next is read" invariant weakens three things the old path enforced for free: the malicious-peer DoS bound, the byte-budget ceiling, and stall-detection accuracy. The items below are worth addressing before merge.

**P1**

- **`services/legacy/peer_server.go:1052`** — When a block fails validation and `awaitBlockResult` disconnects the peer, blocks already read/queued after it (N+1, N+2 …) are still fully validated instead of being skipped. The old synchronous path blocked the read-loop on `<-sp.blockProcessed`, so the disconnect fired before N+1 was ever read. Net effect under the new default: wasted validation work + repeated "block processing failed" log spam on every bad-block disconnect.

- **`services/legacy/peer_server.go:1074`** — On an individual peer disconnect (SyncManager still running), `awaitBlockResult` releases the block's full prefetch weight via `<-sp.quit`, but the decoded block is still sitting in the 100-deep `blockQueue` and still gets validated. The byte-budget then under-counts live buffered-block memory and over-admits — weakening the exact memory bound this feature exists to enforce, under churny disconnects.

- **`services/legacy/netsync/manager.go:2423`** — Stall-detector suppression narrowed from `blockBacklog > 0` to `blockPrefetchWaiters > 0`. A slow local validation that doesn't fill the budget no longer suppresses rotation: `lastBlockTime` goes stale, throughput reads as unhealthy, and with no waiter registered we rotate a perfectly healthy sync peer purely because our own processing was slow.

- **`services/legacy/peer_server.go:1035`** — The `peer.go` `PeerProcessingTimeout` (~3m) was never made backpressure-aware. In the async path `OnBlock` now blocks inside `AcquireBlockPrefetch` when the byte budget is full (a normal steady state when validation lags), but the processing timer keeps running across the blocking acquire and eventually fires `DisconnectWithInfo("Timeout processing message 'block'")` — falsely rotating a healthy peer. This defeats the `blockPrefetchWaiters` fix that was added to the netsync detector.

- **`services/legacy/netsync/manager.go:2423`** — `localReadBackpressured()` keys off the **global** `blockPrefetchWaiters`. A fast non-sync/stream peer that keeps the shared budget saturated holds it nonzero, so `handleCheckSyncPeer` never counts violations against a genuinely dead sync peer → sync can stall indefinitely. This masks a stalled sync peer far more persistently than the old per-node `blockBacklog` guard.

A few lower-severity items (regtest now takes the async path and loses the synchronous-ordering guarantee its harness depends on; duplicate requested-hash can be admitted twice against the budget; byte-budget vs 100-deep queue mismatch can pause the stall ticker; blocked-read-loop teardown deferred to the 3m timer; `awaitBlockResult` goroutine + budget leak on graceful `Stop()`) and two nits (dead `len(buf)` fast-path + now-misleading comment at `peer_server.go:1021`; new test uses `assert` instead of `require`) I'll leave out of here for brevity — happy to expand on any of them.

oskarszoon:
Your review snapshot predates b6388b4cb and 67a4c979c — two of the five P1s and most of the brevity list were already fixed at HEAD. Point by point:

- **Stall suppression narrowed to waiters** — fixed in b6388b4cb; `localReadBackpressured` suppresses on `blockBacklog > 0` again, waiters is an additional signal.
- **Global waiters masking a dead sync peer** — same commit; and the pre-PR guard was already the global `sm.blockBacklog`, so the final scope matches the original suppression.
- **Bad block → queued blocks still validated** — real but bounded by the getdata window (`dynamicMax` ≤ 20), not `budget/64KiB` ≈ 4096; see the thread with freemans13 for why an explicit skip-on-disconnect guard was rejected (false-drops valid blocks from bare peers).
- **Budget released on peer disconnect while the block is still queued** — agreed, fixed in 7c84da426: `awaitBlockResult` now holds the budget until the queued block actually completes processing, so the semaphore tracks live block memory, not peer lifetime.
- **`PeerProcessingTimeout` across a blocking acquire** — agreed, fixed in 90f19981c: the per-message watchdog is no longer armed for block messages when prefetch is enabled; block-stall detection under prefetch is owned by the netsync stall detector + idle timer + `MaxBlockDownloadTime`. Kill switch restores the old watchdog behaviour.
- **Regtest async path** — fixed in 7c84da426: regression net keeps the synchronous ingestion path, preserving the ordering the acceptance tooling depends on.

ordishs (approved):
## Re-review — all blocking feedback addressed ✅

Went through the fix commits (`b6388b4c`, `67a4c979`, `3bafb01a`, `90f19981`, `7c84da42`, `5150522c`) against the earlier findings. Everything is resolved, and cleanly.

**#1 — stall-detector suppression (was the blocker).** `localReadBackpressured()` now treats `blockBacklog > 0` as the primary suppression signal again, with the waiter count as an explicit secondary. The comment nails the exact case I was worried about ("a single block merely slow to validate while the budget was not full") and correctly argues it can't mask a real stall indefinitely — a genuinely stalled peer stops feeding the queue, the backlog drains, and the check resumes. Covered by `TestHandleCheckSyncPeer_*` / `TestLocalReadBackpressured`.

**#2 — `maxBlockQueue` sizing.** Now scales to `blockPrefetchBudgetBytes / minInFlightBlockWeight`, capped at `maxBlockQueueSlots` (65536). Since the queue capacity is ≥ the admission ceiling, the feeder can't wedge on `blockQueue <-` before the budget backpressures — so backpressure lands on `AcquireBlockPrefetch` (waiter-counted) rather than silently stalling cross-peer bookkeeping. Good.

**Shutdown races** (the concurrency angle raised): the block-queue enqueue is now `sm.quit`-guarded so a full queue can't hang `Stop()`, and shutdown drains queued blocks with an error reply so `awaitBlockResult` goroutines release budget promptly instead of waiting on `sp.quit`/`sp.ctx`.

**#3 — `awaitBlockResult` on teardown.** Improved beyond what I flagged: it now *holds* the budget until the block actually leaves the pipeline (fixing the over-admit-under-churn window), and logs a late validation failure for parity while correctly not disconnecting an already-gone peer.

**#4 — synchronous fallback.** Now `select`s on `sp.quit`/`sp.ctx.Done()` alongside `<-sp.blockProcessed`, so no shutdown goroutine leak.

**Bonus — processing watchdog.** Nice catch adding `shouldArmProcessingTimer`: a budget-parked `OnBlock` would otherwise trip `PeerProcessingTimeout`. The predicate (`budget>0 && net != RegTestNet`, disarmed only for `CmdBlock`) correctly mirrors the branch `OnBlock` takes via `UsePrefetchIngestion`, so regtest keeps both the synchronous path and the armed watchdog. Fully covered by `TestShouldArmProcessingTimer`.

Verified locally on `887a5972`: `go build ./services/legacy/... ./settings/...` clean; `go test -race` on the prefetch, `OnBlock`/`awaitBlockResult`, and watchdog tests all pass.

### Non-blocking nit
`UsePrefetchIngestion` compares chain params by pointer (`!= &chaincfg.RegressionNetParams`) while `shouldArmProcessingTimer` compares by value (`net != wire.RegTestNet`). They agree as long as `chainParams` is always the package global (which it is today), so this is only a latent smell — worth a one-line note if you ever pass a params copy. Not worth holding the PR.

Approving. Nicely handled — thanks for the thorough follow-through.

ctnguyen (approved):
This is a carefully engineered change that targets a real, well-diagnosed bottleneck (the read-loop serializing next-block download behind current-block processing) without touching the invariant that matters most — block processing stays single-threaded FIFO, so Aerospike concurrency is unchanged. The design holds its safety properties deliberately: a global byte-weighted semaphore bounds buffered-block memory across all peers, admission happens *before* `QueueBlock` so nothing sits uncounted in the deep queue, oversized blocks are admitted alone to preserve one-at-a-time backpressure, and per-block buffered(1) reply channels prevent a disconnected peer from wedging the shared processing goroutine. The unrequested-block gate preserves the old malicious-peer backpressure *for unrequested blocks* (requested-but-invalid blocks are a bounded, documented relaxation — see ChiR1). The kill-switch (`legacy_blockPrefetchBufferBytes=0`) and the regtest exception are thoughtful. Test coverage is genuinely strong — acquire/clamp/oversized/blocking/waiter-accounting/ctx-cancel/quit-abort, the unrequested gate, the stall-detector backpressure logic, the shutdown teardown of `awaitBlockResult`, and the processing-timer gate are all exercised. The extensive inline commentary makes the concurrency reasoning auditable, and a live Teratestnet run (0→20k blocks: 10816s → 8759s) confirms the intended overlap.

## Prior review follow-up

Substantive items raised in earlier passes and their status against the current code:

- Stale `OnBlock` godoc ("blocks until fully processed"): addressed — the godoc documents both ingestion modes.
- Missing settings-table row in `docs/references/settings/services/legacy_settings.md`: addressed.
- `SerializeSize()` walk vs `len(buf)`: applied in form, but the `len(buf)` branch is dead on the only live call path, so the walk still runs — carried forward as **ChiR2**.
- Teardown asymmetry in `AcquireBlockPrefetch` (only unblocked via ctx): addressed — it also aborts on the peer `quit` channel, mirroring `awaitBlockResult`.
- Stall detector wrongly rotating a healthy peer when a block is slow to *validate*: addressed — `localReadBackpressured()` again suppresses on `blockBacklog > 0`, not only on a budget waiter (`TestHandleCheckSyncPeer_PrefetchBackpressure` covers draining-backlog vs idle-stall).
- Shutdown stranding `awaitBlockResult` goroutines and their reserved budget: addressed — the block-queue handler drains remaining queued blocks with an error reply on `sm.quit`; it was a bounded window, not a permanent leak.
- `assert` used where the repo convention mandates `require`: addressed — swapped to `require.*` and the unused `assert` import removed.
- `PeerProcessingTimeout` firing while a read-loop is blocked in `AcquireBlockPrefetch`: addressed — the per-message watchdog is not armed for block messages when prefetch is enabled; block-stall detection is then owned by the netsync stall detector + idle timer + `MaxBlockDownloadTime` (kill switch restores the old watchdog).
- Regtest accidentally taking the async path: addressed — regression net keeps synchronous ingestion, preserving the submit-then-query ordering the acceptance tooling depends on.
- Budget released on peer disconnect while the block was still queued: addressed — `awaitBlockResult` holds the reserved budget until the queued block actually completes, so the semaphore tracks live block memory, not peer lifetime.
- `maxBlockQueue` smaller than the admission ceiling (cross-peer head-of-line risk on the sole `msgChan` consumer): addressed — the queue is scaled to `budget/minInFlightBlockWeight` (clamped to `maxBlockQueueSlots`); it holds only pointers, so memory stays bounded by the byte budget.
- Synchronous-fallback teardown escape: addressed — the `<-sp.blockProcessed` wait now also selects on `sp.quit`/`sp.ctx.Done()`, matching the async path.
- `awaitBlockResult` skips the validation-failure disconnect after `sp.quit`/`sp.ctx` has fired: **accepted tradeoff** — the disconnect is skipped only once the peer/process is already tearing down, so the connection is already going away; the late failure is logged for parity. Reasonable to keep.
- "One bad block before disconnect" relaxed for *requested-but-invalid* blocks: **still open / accepted** — carried forward as **ChiR1**.

## Non-blocking issues

### ChiR1 — "one bad block before disconnect" is relaxed for requested-but-invalid blocks
`services/legacy/peer_server.go` (OnBlock async path) → `handleBlockMsg` / `awaitBlockResult`

- **Observed:** The old synchronous path blocked the read-loop on each block, so a peer streaming invalid blocks was disconnected after exactly one. The `BlockRequested` gate restores this only for *unrequested* blocks. For *requested* blocks that turn out invalid, `OnBlock` admits and queues each and returns, so up to the outstanding request window can be admitted and sitting in the FIFO queue by the time processing begins. Blocks are processed in order and the first invalid one triggers the disconnect from its `awaitBlockResult`; the still-queued blocks behind it are then validated anyway — the block-queue keeps draining, and depending on teardown ordering some are processed after the disconnect fires — so several requested-invalid blocks get validated rather than one.
- **Why it matters:** It is wasted validation work against a misbehaving peer and a materially weaker version of a documented protection. It is *bounded and memory-safe*: the exposure is the getdata window (`dynamicMax`, ≤20 outstanding requested blocks), not `budget/64KiB` (~4096), and the byte budget + weight floor cap memory/goroutines regardless. A related, smaller relaxation: `BlockRequested` is a membership check that does not consume the request, so duplicate copies of the same requested hash also pass pre-admission until the first copy reaches `handleBlockMsg` and deletes the request — later duplicates are then rejected there before full validation, so they cost a queue slot and budget but not a full validation. Both are tradeoffs, not vulnerabilities — hence non-blocking.
- **Fix (optional):** If tightening is wanted, drop or skip still-queued blocks for a peer already marked for disconnect (a disconnect-scoped skip in `handleBlockMsg`, avoiding the bare-peer false-drop that the `!peer.Connected()` prototype hit). Otherwise, keeping the current bounded behavior with the existing code comment is acceptable.

### ChiR2 — `len(buf)` weight fast path is dead; `SerializeSize()` always runs on the read-loop
`services/legacy/peer_server.go` (OnBlock async path, ~line 948 of the diff)

- **Observed:** The weight is computed as `size := int64(len(buf)); if size == 0 { size = int64(msg.SerializeSize()) }`, with the comment "On the single-stream path buf is the raw serialized payload … the BlockPriority streaming path passes buf == nil." In the current codebase there is exactly one `OnBlock` call site — `services/legacy/peer/peer.go:2006` — and it passes `nil` (`p.cfg.Listeners.OnBlock(p, msg, nil)`), because the read loop uses `readMessageStreaming`, which does not return the payload bytes. The non-streaming `readMessage` (which does return bytes) is not used for dispatch anywhere. So `len(buf)` is always `0` and the code always falls through to `msg.SerializeSize()`.
- **Why it matters:** The `SerializeSize()` walk over the whole tx set — the exact cost the prior review's optimization was meant to avoid — runs on the peer read-loop for every block, and the "single-stream path" the comment describes doesn't exist in the tree. Behaviourally correct (the size is right), but the branch is dead and the comment misleads a future reader.
- **Fix:** Either drop the dead branch and the "single-stream path" comment and call `msg.SerializeSize()` directly (documenting that the streaming read loop never has the raw bytes in hand), or, if the perf intent matters, thread the payload length out of `readMessageStreaming` to the `OnBlock` listener so `len(buf)` is real. The cost is small relative to a multi-second download, hence non-blocking.

```go
// Streaming read loop never hands OnBlock the raw payload, so weigh by
// serialized size directly.
size := int64(msg.SerializeSize())
```

### ChiR3 — budget-holding read-loop can still block on `QueueBlock` during shutdown
`services/legacy/peer_server.go` (async path, `sm.QueueBlock(block, sp.Peer, done)`) and `services/legacy/netsync/manager.go:2257`

- **Observed:** In the async path the budget is acquired, then `QueueBlock` sends on `sm.msgChan` (`sm.msgChan <- &blockMsg{...}`), then `awaitBlockResult` is started. That `msgChan` send is not guarded by `sp.quit`/`sp.ctx` (unlike the `blockQueue` enqueue and the acquire itself, which the PR carefully guards). If the sync manager is shutting down and `msgChan` is full while `blockHandler` has stopped draining, this send blocks and the reserved budget is held until process exit.
- **Why it matters:** It leaks a read-loop goroutine and pins budget on shutdown. This is pre-existing behaviour (the synchronous path had the same unguarded `QueueBlock`), it is shutdown-only, and the process is terminating anyway — so purely for symmetry with the shutdown care taken elsewhere in this PR, not a real-operation risk.
- **Fix:** Optionally make the `blockMsg` enqueue in `QueueBlock` shutdown-aware (select on `sm.quit`/ctx as the `blockQueue` feeder already does), so a parked read-loop can exit cleanly and release budget on teardown.

## Recap

| ChiR1 | One-bad-block protection relaxed for requested-invalid blocks | | 40% |
| ChiR2 | Dead `len(buf)` branch; `SerializeSize` always walks | | 30% |
| ChiR3 | `QueueBlock` send unguarded on shutdown (holds budget) | | 12% |

icellan (changes requested):
Re-reviewed after the latest push. Most of the first round is genuinely addressed — budget is now held until the block leaves the pipeline, the stall-detector guard is restored, regtest is forced synchronous, shutdown drains the queue with guarded enqueue, the queue is sized to the admission ceiling, and the test uses `require`. Thanks for working through those.

Requesting changes on one blocker plus two anti-abuse regressions the async path introduced.

## Blocker

**`services/legacy/peer/peer.go:1956` — a genuine block-*processing* stall now silently wedges legacy sync with no timeout, disconnect, or sync-peer rotation.**

Disarming the per-message watchdog for block messages under prefetch fixed the *false* disconnect from the last round, but it removed the only timeout that covered the processing phase — and the other "fallbacks" the comment credits don't cover a processing stall:

- `blockBacklog.Add(-1)` (`manager.go:2193`) runs only *after* `handleBlockMsg` returns, so on a hang the backlog stays ≥1 and `localReadBackpressured()` suppresses `handleCheckSyncPeer` **forever**;
- the idle timer is stopped for the duration of `OnBlock`;
- `MaxBlockDownloadTime` is cleared on block receipt, so it bounds download, not processing.

On a realistic failure — Aerospike overload, a store/validator deadlock, a liveness-kill mid-`setTxMined` — the node just stops syncing and needs a manual restart. Pre-PR the 3-minute `PeerProcessingTimeout` fired and rotated the peer.

The fix that keeps the last round's improvement (no false rotation on slow-but-progressing validation) while restoring the liveness signal is to make the suppression **progress-aware** — suppress only while the backlog is actually *draining*:

```go
// new field, alongside blockBacklog:
lastBacklogProgress atomic.Int64 // unixnano; stamped when a block completes,
                                  // and on the 0->1 transition (window start)

// on the 0->1 enqueue transition and on every Add(-1):
sm.lastBacklogProgress.Store(time.Now().UnixNano())

func (sm *SyncManager) localReadBackpressured() bool {
    if sm.blockBacklog.Load() > 0 {
        // Suppress only while the backlog is still making progress. If no block
        // has completed for longer than the processing-stall timeout, this is a
        // genuine hang, not slow-but-progressing validation: stop suppressing so
        // handleCheckSyncPeer logs and rotates — restoring the pre-prefetch
        // liveness signal without the false rotation that motivated disarming
        // the watchdog.
        return time.Since(time.Unix(0, sm.lastBacklogProgress.Load())) < blockProcessingStallTimeout
    }
    return sm.blockPrefetchBudget != nil && sm.blockPrefetchWaiters.Load() > 0
}
```

`blockProcessingStallTimeout` can reuse `PeerProcessingTimeout` (~3m). This is the surgical version; the deeper fix (a validation-phase context timeout in `handleBlockMsg` so a hung Aerospike call is actually aborted rather than just surfaced) is a bigger change and out of scope for this PR — but at minimum the stall needs to become *visible* and rotate again, which the above restores.

## Also needs changing

- **`services/legacy/peer_server.go:1026`** — the async unrequested-block pre-check calls `DisconnectWithWarning` on the **stream sub-peer** and returns before `QueueBlock`, so `handleDonePeerMsg` only does `assoc.RemoveStream(...)` and leaves the primary/association intact. A BlockPriority peer flooding unrequested blocks over its DATA stream loses only the stream, re-opens it, and repeats. The sync path reaches `handleBlockMsg`, which reassigns `peer=primary` and evicts the whole association. Resolve to the primary before disconnecting here too.

- **`services/legacy/peer_server.go:1066`** — the async handoff drops the one-bad-block-in-flight cap: a peer streaming requested-but-invalid blocks can get up to a full budget (~256 MiB) worth admitted and validated FIFO before the head block's `awaitBlockResult` disconnects it. Worth bounding in-flight *unvalidated* blocks per peer, or disconnecting on the first validation failure before draining the rest.

## Worth fixing while here

- **`services/legacy/peer_server.go:1034`** — `buf` is always nil (sole caller `peer.go:2030` passes nil), so the `len(buf)` shortcut is dead and `msg.SerializeSize()` (an O(all-tx) walk) always runs on the read-loop before admission — worst for the largest blocks, and it eats the very download/processing overlap this feature exists to create. Either wire the real serialized length through, or compute weight off the header/tx-count instead of walking the whole block on the hot path.

Lower-severity items left in my notes (a >4 GiB-budget cross-peer head-of-line stall, a shutdown result-pairing race on the sync path, decoded-vs-serialized memory undercount at the 256 MiB default, a test asserting on an incidental panic, a dead regtest branch in `BlockRequested`, and the `shouldArmProcessingTimer`/`UsePrefetchIngestion` predicate duplicated across the package boundary) — happy to expand on any.

freemans13 (commented):
Reviewed the async prefetch change end-to-end (concurrency, teardown, budget accounting). It's carefully done — I verified the load-bearing invariants and they hold:

- Budget released exactly once on every exit path (`defer ReleaseBlockPrefetch`; `QueueBlock` shutdown-race still replies on the buffered(1) `done` so `awaitBlockResult` drains + releases).
- Oversized block admitted alone (weight clamped) — can't deadlock.
- Semaphore acquire vs ctx-cancel doesn't double-count; linking goroutine (`quit`→`cancel`) has no leak.
- `blockQueue` depth sized to `budget/minInFlightBlockWeight` + `sm.quit`-guarded feeder ⇒ no cross-peer head-of-line deadlock.
- Concurrent disconnects from multiple failed in-flight blocks are safe (`DisconnectWithLogFunc` is idempotent via `atomic.AddInt32 != 1`).
- FIFO ordering preserved; build + vet clean; the new tests pass under `-race`.

Two minor, non-blocking observations:

**1. Duplicated regtest+budget predicate (maintenance risk).** `shouldArmProcessingTimer` in `services/legacy/peer/peer.go` re-implements `netsync.UsePrefetchIngestion`'s `budget>0 && !regtest` rule (documented import-cycle workaround). It's currently correct — I confirmed `chaincfg.RegressionNetParams.Net == wire.RegTestNet` — but the two will silently drift if a third condition is ever added to `UsePrefetchIngestion`, and the failure mode (watchdog firing during legitimate `AcquireBlockPrefetch` backpressure, or not firing when it should) is subtle. Consider a small shared predicate, or a test asserting the two agree across the net/budget matrix.

**2. No end-to-end test of the async happy path.** The helpers are well covered, but nothing drives the full `OnBlock` async path through a real `SyncManager` to prove in-order `QueueBlock` + budget release on *successful* validation. Acknowledged already (empirical blk/hr needs a real peer sync) — noting as a known gap, not a blocker.

Also: the branch is behind `main` (`main` isn't an ancestor of the PR head) — worth a rebase before merge.

oskarszoon:
@icellan All four addressed in 347998f2a (branch also merged with main).

**Blocker — a processing stall rotates again.** Took your surgical version: `lastBacklogProgress` is stamped on the 0→1 enqueue transition and on every completion decrement (including the shutdown drain), and `localReadBackpressured()` suppresses only while the stamp is younger than the stall timeout (wired to `settings.Legacy.PeerProcessingTimeout`, default 3m, with a const fallback for settings-less test managers). Enqueues into a non-empty backlog deliberately don't stamp — only completions advance processing, so a peer can't keep the suppression alive by feeding more blocks into a hung pipeline. Also deliberately no fall-through to the waiter check when the backlog is stale: a hung pipeline with a full budget accumulates waiters, and rotation is wanted then. Covered by `TestLocalReadBackpressured` (fresh vs backdated stamp, stale+waiter still rotates) and a new stalled-backlog subtest in `TestHandleCheckSyncPeer_PrefetchBackpressure`.

**Unrequested-block pre-check (`peer_server.go`)** now resolves to the association primary and disconnects it, mirroring `handleBlockMsg`, so `handleDonePeerMsg` evicts the whole association instead of just `RemoveStream`. It additionally disconnects the stream sub-peer itself: on primary disconnect `handleDonePeerMsg` removes the association bookkeeping but never closes the sub-peers' TCP connections, so the flooding DATA stream would keep reading otherwise. Factored as `misbehaviorDisconnectTarget`, unit-tested with a synthetic association (lone peer → self, stream → primary, primary → self).

**Requested-but-invalid tail**: `handleBlockMsg` now skips any queued block whose source peer is already disconnected — `awaitBlockResult` disconnects on the first validation failure, and the skip deterministically stops validating the rest of that peer's admitted tail. The skip returns a ServiceError (benign to `shouldDisconnectOnBlockErr`, so no second disconnect, just budget release + log) and is gated on `UsePrefetchIngestion()` so the regtest/synchronous path is untouched. One detail: it checks the queued peer (the stream sub-peer for BlockPriority), not the resolved primary — that is the peer `awaitBlockResult` actually disconnects, so checking the primary would miss exactly the streaming case. `TestHandleBlockMsg_SkipsDisconnectedPeer`.

**SerializeSize on the read-loop**: confirmed the `len(buf)` branch was dead — the streaming read path never buffers the payload. The streaming reader's byte count is now plumbed through: the `OnBlock` listener gains `payloadSize int64` (message bytes minus the 24-byte wire header; exact for extmsg blocks too, verified against go-wire), and `blockAdmissionWeight` prefers it, keeping `len(buf)`/`SerializeSize()` as fallbacks. No tx walk on the read-loop for real blocks anymore.

The lower-severity list stays as-is per your note — say if you want any of them pulled into this PR.

oskarszoon:
@freemans13 Both picked up in 347998f2a:

- **Predicate duplication**: extracted `peer.UseBlockPrefetchIngestion(budgetBytes, net)` as the single source of truth; `shouldArmProcessingTimer` and `netsync.UsePrefetchIngestion` both delegate to it (the netsync side now compares `chainParams.Net` by value instead of the params pointer — also ordishs's nit). `TestShouldArmProcessingTimer` and `TestUsePrefetchIngestion` are now full budget × net matrices asserting both call sites track the shared predicate, so a third condition can't drift silently.
- **E2E async happy path**: looked at it and skipped deliberately. The budget lifecycle lives entirely in `peer_server.go` (`OnBlock` acquires, `awaitBlockResult` releases) — the `QueueBlock`→`handleBlockMsg` path a netsync-level test can drive never touches the budget, so such a test would pass trivially without exercising real budget handling. A genuine version needs a live `serverPeer`, which only the skipped `TestBlockchainSync` harness provides. FIFO ordering is structural (single feeder, single consumer) and release-on-completion is covered by the acquire/release tests, so leaving this as the acknowledged gap.
- Branch is merged with main (887a59725, ad5a45510).

icellan (changes requested):
Re-reviewed `347998f2a`. This addresses everything from the last round, and a couple of the fixes are better than what I asked for — the progress-aware stall suppression (`lastBacklogProgress` / `blockProcessingStallTimeout`) correctly refuses to refresh the liveness signal on enqueue-into-a-non-empty-backlog, and the wire-measured `payloadSize` keeps the `SerializeSize` walk off the read-loop. Nice.

One blocker before this can go in, plus some backpressure/liveness residuals.

## Blocker

**`services/legacy/peer_server.go:1188` (and the synchronous path at `:1008`) — a block *validation failure* never rotates a BlockPriority sync peer.**

The unrequested-block gate now correctly evicts via `misbehaviorDisconnectTarget(sp.Peer)` plus the stream disconnect — but the two *validation-failure* disconnects still call `sp.DisconnectWithWarning(...)` on `sp.Peer` directly. With the shipped defaults (`legacy_allowBlockPriority=true` + prefetch enabled), blocks arrive on the DATA1 stream sub-peer, so `sp.Peer` is the sub-peer: `handleDonePeerMsg` only runs `assoc.RemoveStream` and netsync's `DonePeer` early-returns (the sub-peer isn't in `peerStates`). The **primary stays the sync peer and is never rotated**, `openRequiredStreams` re-opens DATA1, and the node keeps syncing from a peer feeding invalid blocks — legacy IBD stalls. The comment's "disconnecting to trigger sync peer rotation" doesn't happen for exactly the default config.

Since the gate already got this right, the clean fix is to factor its logic into one helper and use it at all three misbehaviour-disconnect sites:

```go
// disconnectMisbehaving evicts a peer's whole association for misbehaviour:
// the association primary (drives associationMgr.Remove in handleDonePeerMsg),
// plus the stream sub-peer's live TCP when distinct.
func disconnectMisbehaving(sp *serverPeer, reason string) {
    target := misbehaviorDisconnectTarget(sp.Peer)
    target.DisconnectWithWarning(reason)
    if target != sp.Peer {
        sp.DisconnectWithWarning(reason + " (stream)")
    }
}
```

Then replace the bare `sp.DisconnectWithWarning(...)` at `:1008` and `:1188` with it (and reuse it at the `:1042` gate).

## Backpressure / liveness residuals (please assess)

1. **`services/legacy/peer/peer.go:1978`** — disarming the per-message watchdog for blocks removed the only per-peer timeout, but the compensating netsync stall detector only inspects the *sync* peer. `OnBlock`'s pre-admission work (the ban-check `server.query` round-trip and `GetBlockHeader` gRPC) runs before any backlog/waiter accounting, so if it hangs on a peer that isn't the current sync peer, nothing reclaims that peer/slot until daemon shutdown. Pre-PR the 3-minute watchdog covered it.

2. **`services/legacy/peer_server.go:1025`** — the unrequested-block gate admits duplicates of the same requested hash: `BlockRequested` only reads `state.requestedBlocks`, and the entry isn't deleted until `handleBlockMsg` (`manager.go:1436`), with no in-flight-by-hash dedup. A sync peer replaying the same near-budget-sized requested block once or twice fills the whole 256 MiB budget and parks every legacy peer's read-loop in `AcquireBlockPrefetch` — defeating the "a malicious peer cannot outrun the budget" property this gate was added to guarantee.

3. **`services/legacy/netsync/manager.go:2599`** — `legacy_blockPrefetchBufferBytes=0` is documented as a revert to "proven synchronous behaviour", but `localReadBackpressured` now timeout-gates suppression, so on the kill-switch path a single block that legitimately validates for longer than `PeerProcessingTimeout` can rotate a healthy sync peer — churn the old synchronous code never produced. Either keep the kill-switch path on the old unconditional suppression or adjust the doc.

Lower-severity: a shutdown-window `sp.blockProcessed` drain race on the synchronous path (`peer_server.go:1000`, confined to daemon exit), and a 32-bit-only budget truncation in `settings.go` (`int64(getInt(...))`, no `getInt64` helper).

## Nits

- `services/legacy/netsync/manager.go:2420` — `BlockPrefetchEnabled()` is now dead exported API (only its own test calls it; `UsePrefetchIngestion()` is the real gate). Worth removing so a future caller can't pick it and silently skip the regtest exclusion.
- `services/legacy/netsync/manager.go:2453` — `BlockRequested` is the fourth copy of the stream-peer to association-primary resolution walk (`handleBlockMsg`, `handleHeadersMsg`, `handleInvMsg` have the other three); a shared helper would keep them from drifting.

oskarszoon:
@icellan Blocker + residuals #2 and #3 fixed in 9cb7762b3 (branch also merged with latest main). #1, the lower-severity items, and the two nits are tracked for a follow-up — flagged below.

**Blocker — validation failure now evicts the association.** Confirmed exactly as you described: with the shipped BlockPriority + prefetch default, `sp.Peer` at both failure sites is the DATA1 sub-peer, so `sp.DisconnectWithWarning` left the primary as sync peer. Factored your `disconnectMisbehaving(sp, reason)` (primary via `misbehaviorDisconnectTarget` + the stream sub-peer when distinct, idempotent) and routed all three misbehaviour sites through it — the synchronous path (:1010), the async `awaitBlockResult` (:1212), and the unrequested-block gate (which previously inlined this). `TestDisconnectMisbehaving` covers stream → both, lone/primary → self.

**#3 — kill-switch keeps unconditional suppression.** `localReadBackpressured` now short-circuits to `true` on `blockPrefetchBudget == nil` before the progress-timeout branch. Rationale in the godoc: with prefetch off the per-message watchdog is still armed for blocks and owns processing-stall liveness, so the timeout-gate would only introduce rotation churn the synchronous path never had. The progress-aware timeout stays scoped to the prefetch path where the watchdog is disarmed. `TestLocalReadBackpressured` asserts budget-nil + stale backlog → still suppressed.

**#2 — in-flight-by-hash dedup.** Added a SyncManager-level `inFlightBlocks` set paired 1:1 with the byte budget. `AcquireBlockPrefetch` reserves the hash *before* the (possibly blocking) budget Acquire — so duplicates are bounded even while a copy is parked waiting for budget — and returns `ErrDuplicateBlockInFlight` if the hash is already in flight; `OnBlock` drops the duplicate without disconnecting or queueing (the surviving copy drives the validation outcome). Hash and budget share one lifetime: inserted at reserve, removed on Acquire-failure, and deleted in the single `ReleaseBlockPrefetch` defer in `awaitBlockResult` that already covers every teardown path — so the two halves can't drift and no path leaks a slot. Tests: dedup rejects the second copy, a duplicate consumes no budget (a different hash still admits), and the synchronous/budget-nil path skips dedup entirely.

**Deferred to a follow-up issue** (happy to fold any back in if you'd rather block on them):
- #1 — `OnBlock` pre-admission hang (ban-check round-trip + `GetBlockHeader`) on a non-sync peer with the watchdog disarmed. Needs a scoped timeout / coarse re-arm; narrower than the above (requires an actual hang in normally-fast calls).
- Lower-severity: synchronous-path `sp.blockProcessed` shutdown drain race; 32-bit `int64(getInt(...))` budget truncation.
- Nits: dead exported `BlockPrefetchEnabled()`; 4th copy of the stream→primary resolution walk.

`go build ./...`, `go vet`, `staticcheck`, and `go test -race` on legacy/netsync/peer/peercli all clean.

icellan:
Re-reviewed `9cb7762b3`. The two headline blockers from the previous rounds are genuinely gone: the in-flight dedup is sound (I went looking for a stale-timestamp and an insert-before-acquire race and neither holds), the kill-switch suppression is now unconditional when the budget is nil, and the bad-block eviction reaches all three OnBlock disconnect sites via `disconnectMisbehaving`. Not blocking the merge on the below — one real-but-narrow leak and a set of robustness residuals worth a look.

## The one correctness item

**`services/legacy/peer_server.go:1113` — `disconnectMisbehaving` leaks the DATA1 sub-peer when a bad block arrives on the primary.**

The helper closes the extra stream sub-peer only in the `target != sp.Peer` branch — i.e. only when the bad block arrived *on* the sub-peer. When it arrives on the primary while a DATA1 is open, `target == sp.Peer`, that branch is skipped, and only the primary is disconnected. `handleDonePeerMsg`'s primary branch (`peer_server.go:2205`) then runs `associationMgr.Remove` but never closes the sub-peers' connections, so the DATA1 `serverPeer`'s TCP and `inHandler` read-loop stay live and keep reading off the wire from a peer we already evicted — a leaked goroutine/connection plus continued ingestion. The function's own doc comment promises to prevent exactly this ("must be disconnected here too or its DATA stream keeps reading off the wire"), but only does it in the stream-arrival branch. Trigger is narrow (a misbehaving peer sending on the GENERAL stream rather than the default DATA1 block path), hence not a merge blocker, but the connection genuinely leaks.

Rather than teach the helper to enumerate the association's other streams, I'd centralize the teardown at the eviction point so no disconnect site can miss it — close the association's stream connections in `handleDonePeerMsg`'s primary-disconnect branch (it already tracks them via `RemoveStream`/`StreamType`):

```go
} else {
    // Primary disconnected: evict the association AND close its live stream
    // sub-peer connections. associationMgr.Remove only drops bookkeeping; the
    // sub-peers' TCP + read-loops would otherwise leak and keep reading from an
    // evicted peer.
    for _, stream := range assoc.Streams() { stream.Disconnect() }
    s.associationMgr.Remove(assoc.RawID())
}
```

Then `disconnectMisbehaving` collapses to "disconnect the primary", every association-eviction path is complete, and the downstream `manager.go:1426` disconnect (which can't call `disconnectMisbehaving` from the netsync package anyway) is covered for free.

## Robustness residuals (worth assessing, not blocking)

1. **`services/legacy/peer/peer.go:1978`** — with the per-message watchdog disarmed for blocks, a DATA1 read-loop parked in `AcquireBlockPrefetch` on a processing hang is never unparked: the idle timer is stopped, `MaxBlockDownloadTime` only bounds download, and the netsync stall detector rotates the *primary* (closes the primary's quit), not the parked sub-peer's quit. The progress-aware suppression restored sync-peer rotation but doesn't relieve a stuck sub-peer read-loop, so ingestion can silently stop until daemon shutdown.

2. **`services/legacy/peer_server.go:1052`** — the 256 MiB budget is one global semaphore shared across all peers/streams, so one peer's slow-to-validate blocks can exhaust it and park every other peer's read-loop (including one relaying a fresh tip). The pre-PR per-peer isolation prevented that. A per-peer sub-cap, or reserved headroom for non-sync peers, would bound it.

3. **`services/legacy/peer_server.go:1541`** — a block parked waiting for budget is fully decoded and held live but not yet charged against the budget, so peak resident memory is `budget + N*max_block`, above the "hard ceiling" the docs promise.

4. **`services/legacy/peer_server.go:997`** — shutdown-window: the sync-path `<-sp.ctx.Done()` early-return can leave `sp.blockProcessed` undrained, and a second undrained reply blocks the single shared `blockHandler` goroutine's `msg.reply <- err`, halting block processing for all peers. Confined to daemon shutdown.

5. **`services/legacy/peer_server.go:1491`** — `legacy_blockPrefetchBufferBytes=0` now disconnects via `disconnectMisbehaving` (whole-association eviction) on a block error instead of the pre-PR bare single-peer disconnect, so the documented "revert to proven synchronous behaviour" isn't byte-faithful on the disconnect action.

6. **`services/legacy/netsync/manager.go:1399`** — when a healthy sync peer drops mid-sync, the `!bmsg.peer.Connected()` skip discards the entire FIFO tail of its already-downloaded valid blocks; they're re-fetched from the replacement peer. Sync recovers, so it's re-download waste rather than a gap — but worth confirming the tradeoff is intended.

## Nits

- `services/legacy/netsync/manager.go:2445` — `BlockPrefetchEnabled()` is dead exported API (only its own test calls it; `UsePrefetchIngestion()` is the real gate). Remove/unexport so nobody picks the predicate that skips the regtest exclusion.
- `services/legacy/netsync/manager.go:2478` — the stream-peer to association-primary walk now lives in 4+ places; a single `PrimaryPeerOrSelf` helper on `*peer.Peer` would keep them from drifting.

oskarszoon:
@icellan Decided to fold the deferred tail in rather than punt to a follow-up — all of it is in 3a05b071f now, so the whole 07-10 list is addressed in this PR.

- **Residual # 1 — pre-admission hang.** `OnBlock`'s ban-check round-trip and `GetBlockHeader` are now bounded by a `PeerProcessingTimeout`-derived context (`checkBannedBounded` + `preAdmitCtx`). The timeout covers only pre-admission — `AcquireBlockPrefetch`/`QueueBlock` stay on `sp.ctx`/`sp.quit` — and those calls are sub-ms in a healthy node with budget backpressure strictly after them, so it catches a genuine hang on any peer (sync or not) without false-firing on healthy backpressure. On timeout the block is dropped, not disconnected (the hang isn't the peer's fault). Guarded with a 3m fallback if the setting is ever zeroed. `TestCheckBannedBounded` drives both timeout selects + the pass-through.
- **Lower-severity — sync-path reply channel.** The shared buffered(1) `sp.blockProcessed` is drained before the synchronous `QueueBlock`, so a value left by a prior block's shutdown-drain can't be read as this block's result. Benign today (daemon-exit only) — kept as robustness, sync path otherwise untouched.
- **Lower-severity — 32-bit truncation.** Added a `getInt64` settings helper; `legacy_blockPrefetchBufferBytes` no longer goes through `int64(getInt(...))`, so a >2 GiB budget can't truncate on a 32-bit build.
- **Nit — dead API.** Removed the now-unused exported `BlockPrefetchEnabled()` (only its own test called it; `UsePrefetchIngestion` is the gate) so nothing can pick it and skip the regtest exclusion.
- **Nit — resolution walk.** The stream→primary `peerStates` walk that was inlined in `handleBlockMsg`/`handleHeadersMsg`/`handleInvMsg`/`BlockRequested` is now one helper, `peerStateResolvingPrimary` — behaviour-preserving, each site keeps its own logging/reassignment.

`go build ./...`, `go vet`, `staticcheck`, and `go test -race` on legacy / netsync / peer / peercli / settings all clean. Branch is on latest main.

That should be the full list from this round and the last — nothing knowingly left open on the prefetch change now.

services/legacy/peer_server.go
```diff
@@ -955,41 +978,298 @@ func (sp *serverPeer) OnBlock(_ *peer.Peer, msg *wire.MsgBlock, buf []byte) {
...
+
+// checkBannedBounded runs the ban-check round-trip (server.query send + reply)
+// under ctx so a wedged query goroutine or a never-serviced channel cannot park
+// the read-loop. It returns (banned, ok); ok=false means the round-trip was
+// cancelled/timed out before a reply arrived and the caller must drop the block
+// without proceeding. Extracted from OnBlock so the bounded pre-admission path is
+// unit-testable (a query channel nobody services + a cancelled ctx).
+func (sp *serverPeer) checkBannedBounded(ctx context.Context, host string) (banned bool, ok bool) {
+	// Use a channel to get the response from the query.
+	respChan := make(chan bool)
```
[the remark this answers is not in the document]
oskarszoon: Fixed in 3a8010c01 — buffered respChan to (1) so the reply send is non-blocking even if the receive was abandoned on the deadline. Window is tiny, but the receiver became abandonable with this change, so worth closing at zero cost.

icellan:
Re-reviewed `d97bff7cf`. The residuals from last round are all addressed: the pre-admission phase is now bounded (`preAdmitCtx` + a buffered, ctx-guarded ban-check that can't wedge `handleQuery`), the budget config no longer truncates on 32-bit (`getInt64`), and the stream→primary walk is consolidated into `peerStateResolvingPrimary`. This is close. Not blocking — but the pre-admission timeout, good as the bound is, introduced two new timeout-branch regressions worth tightening, plus a couple of latent footguns.

## Timeout-branch regressions (from the new pre-admission bound)

1. **`services/legacy/peer_server.go:948`** — when the ban-check times out (`checkBannedBounded` returns `ok=false` after `preAdmitTimeout`, e.g. a wedged/backed-up `handleQuery`), OnBlock drops the requested block and keeps the peer connected, but neither re-queues it nor clears its `requestedBlocks` entry — so it isn't re-requested and that block stalls IBD. Pre-prefetch, the armed watchdog disconnected the peer and rotation re-fetched it from a healthy one. The bound is right; the timeout branch should trigger rotation / re-request (e.g. `disconnectMisbehaving`, or drop the `requestedBlocks` entry) rather than silently drop.

2. **`services/legacy/peer_server.go:985`** — `GetBlockHeader` now runs under `preAdmitCtx`, so a timed-out lookup errors and sets `blockIsKnownValid=false`, causing an already-known-valid block to be admitted against the budget and fully re-processed instead of skipped. Under blockchain-service load with a low `legacy_peerProcessingTimeout` that's redundant work + budget consumption the old untimed `sp.ctx` lookup avoided. Consider keeping the known-valid *skip* decision on an untimed lookup (or treating a header-lookup timeout as "unknown, retry later" rather than "not valid, reprocess now").

## Confirmed, but low-severity

3. **`services/legacy/netsync/manager.go:1398`** — the `!bmsg.peer.Connected()` tail-stop is best-effort, not a barrier: `blockHandler` hands bad block N to the async `awaitBlockResult` and then dequeues N+1 and checks `Connected()` before `disconnectMisbehaving` has flipped the atomic, so N+1 (and anything dequeued in that window) is still fully validated. A peer feeding expensive invalid blocks can burn CPU on more than one per disconnect — weaker than the old synchronous "one bad block in flight" guarantee. Inherent-ish to the async design and the peer's being dropped anyway, so noting rather than pushing hard.

## Latent (production-safe today)

4. **`services/legacy/netsync/manager.go:2453`** — `UsePrefetchIngestion()` evaluates `sm.chainParams.Net` as a call argument, so the pointer is dereferenced before `UseBlockPrefetchIngestion`'s `budgetBytes>0` short-circuit can guard it — and `handleBlockMsg` now calls it unconditionally for every block. A nil-`chainParams` SyncManager panics the block-handler goroutine. `New()` always sets it (and the diff had to add `chainParams` to several tests), so it's latent — a `sm.chainParams != nil` guard or passing `*chaincfg.Params` would close it.

5. **`services/legacy/netsync/manager.go:2467`** — two regtest-detection patterns for the same concept: `BlockRequested` uses pointer equality (`sm.chainParams == &chaincfg.RegressionNetParams`) while its sibling `UsePrefetchIngestion` uses value semantics (`.Net != wire.RegTestNet`). They disagree when `chainParams` is a copy of `RegressionNetParams` (as some tests construct). Harmless today (OnBlock doesn't call `BlockRequested` on the synchronous path), but route both through `.Net` so they can't drift.

## Still standing from the last review

The `disconnectMisbehaving` sub-peer leak I raised last round (`peer_server.go`, the `target == sp.Peer` path leaving an open DATA1 connection uncaught when a bad block arrives on the primary) is untouched here — still worth the `handleDonePeerMsg` centralization I sketched, which also covers the downstream `manager.go` unrequested-block disconnect. And the pre-admission bound explicitly doesn't extend to `AcquireBlockPrefetch`, so a DATA1 read-loop parked on a full budget during a genuine processing hang is still never unparked (rotation closes the primary's quit, not the sub-peer's). Neither is a merge blocker in my view, but both remain.

icellan (approved):
Approving to clear my earlier requested-changes — the two blockers from the prior rounds (the silent sync-wedge on a processing hang, and the sync-peer never rotating on a bad block) are resolved and verified. The items in my latest comment (the ban-check-timeout drop-without-re-request, the known-valid re-processing on header-lookup timeout, and the `disconnectMisbehaving` sub-peer leak on primary-arrival) are worth a follow-up but none is a default-config correctness blocker. Good work iterating on this.

oskarszoon:
Thanks — agree none of these should block, and I'd rather not keep growing this PR after four rounds, so I've tracked all seven as follow-ups grouped by root cause:

- #1277 (bug) — pre-admission timeout branches strand / reprocess blocks (your #1 + #2). Unified fix: treat a `preAdmitCtx` timeout as drop + re-request, distinct from a genuine not-found.
- #1278 — multistream sub-peer teardown on primary disconnect (your #6 + #7). Same root cause: primary disconnect doesn't close the association's stream connections / unpark their read-loops; fix by centralizing in `handleDonePeerMsg`.
- #1279 — chainParams nil-guard on `UsePrefetchIngestion` + unify regtest detection through `.Net` (your #4 + #5).
- #1280 — async bad-block tail-stop is best-effort not a barrier (your #3), noted with options.

All linked back to #1189. Calling out that #1 and #2 are regressions the pre-admission bound in this PR introduced — happy to pull #1277 back into this PR before merge if you'd rather not ship those in the timeout branch; otherwise it's queued.

PR #1099, opened 2026-06-16
perf(tracing): cut per-span allocation on the hot path (gocore.Stat, tags, StartTime)

oskarszoon:
## What

`UTracer.Start` did real work on every span **even when tracing is disabled** — the default in production and the documented "zero overhead" state. Two commits cut the per-span cost on the hot path:

**1. Skip gocore.Stat creation and tag allocation when tracing disabled**
- `NewStatFromContext` created a `gocore.Stat` (each with its own `RWMutex` and child `sync.Map`, taking the parent's child-map lock) on every span — unconditionally.
- the tag options allocated a `[]tracingTag` slice that is only read when building OTel span attributes (already skipped when disabled).

Both are now gated on `tracingEnabled`; `WithTag` is a no-op when disabled.

**2. Make StartTime context injection opt-in**
- `context.WithValue(ctx, StartTime, start)` boxed a `time.Time` and allocated a valueCtx on every span. Only one call site reads it back (`blockvalidation.catchupGetBlockHeaders`).
- new `WithStartTime()` option; injection happens only when set. The one consumer opts in (it already had a `time.Now()` fallback, so behavior is unchanged). Every other span skips the boxing.

## Where this came from

CPU profiling a mainnet legacy sync (operator single-node, Aerospike). `UTracer.Start` was ~10% cumulative CPU with tracing already off. The cost was the unconditional `gocore.Stat` machinery, tag allocation, and StartTime boxing — not the OTel SDK (which was correctly gated). After deploying commit 1 as a beta and re-profiling, the `gocore.Stat`/tag overhead was confirmed gone and the residual `UTracer.Start` cost was dominated by the StartTime `context.WithValue`, which commit 2 removes from the hot path.

## Behavior changes

- The `gocore.Stat` `/stats` perf tree is **no longer populated when tracing is disabled**. This affects `WithParentStat` consumers (blockvalidation / utxopersister / validator). Enable tracing (`tracing_enabled=true`) to restore it. Aligns with the `tracing_enabled` setting's documented contract ("all tracing operations become no-ops … for zero performance overhead").
- `tracing.StartTime` is now only present in the context when the span was started with `WithStartTime()`. The sole reader (`catchupGetBlockHeaders`) opts in and is unaffected.

## Benchmark

`BenchmarkStart_Disabled` (disabled path, 2 tags/span):

|                 | ns/op | B/op | allocs/op |
|-----------------|-------|------|-----------|
| baseline        | 437.4 | 760  | 10        |
[...]

PR #1191, opened 2026-06-30
perf(tracing): eliminate disabled-path per-span allocations in UTracer.Start

### LINKED ISSUE #1188: legacy/UTXO hot path: disabled-tracing wrapper + per-op batcher timer dominate CPU/GC during big-block catchup
## Problem

During big-block legacy catchup, ~30% of the legacy service's CPU is runtime
sync/scheduler/GC overhead rather than useful work. Two avoidable contributors, both
fixable in code with no consensus impact:

1. **Tracing wrapper, even when tracing is OFF.** `tracing.(*UTracer).Start` =
   **7.7% of CPU and 8.2% of allocations (~178 GB over the profile window)** — with
   tracing **disabled** (`tracing_enabled=false`, no exporter goroutines). `Start`
   allocates `TraceOptions` and runs every option closure (incl. `WithLogMessage`'s
   `args ...interface{}` boxing) *before* the disabled-check short-circuits
   (`util/tracing/tracing.go:393+`). The expensive `gocore.Stat` path is already
   gated, but the option/alloc work is not.

2. **Per-op `time.NewTimer` churn.** Every batcher submit-wait allocates a fresh
   `time.NewTimer(s.batcherWait)` (`create.go:235`, `get.go:391` & `:1226`,
   `spend.go:943`) = **~5% of allocations (~108 GB)**. `batcherWait` is ~2m40s
   (post-#1172) so the timer effectively never fires — pure churn, one per UTXO op
   across millions per block.

## Evidence

eu-3, 30s CPU profile during a 2.64M-tx block (341% CPU):
- `runtime.mallocgc` 15.2% cum + GC mark workers ~4.5% → GC ~15-20% of CPU.
- `UTracer.Start` 7.7% cum.
- heap alloc: `UTracer.Start` 178 GB, `time.NewTimer` 108 GB (source confirmed =
  `(*Store).Create`/`get`/`Spend` submit-wait).
- tracing confirmed off: 0 otel/exporter goroutines in the goroutine dump.

## Scope

This is a CPU/GC win in the **big-block / high-throughput** regime. Steady small-block
catchup is I/O-wait-bound (see the throughput issue), so this will not speed that up —
it reduces CPU/GC headroom pressure during heavy blocks.

## Fix direction

- `UTracer.Start`: short-circuit at the top when tracing is disabled — skip the
  `TraceOptions` alloc and option-closure evaluation (or make option application
  lazy); avoid `WithLogMessage` arg-boxing on the disabled path.
- Pool/reuse the `batcherWait` timer (`sync.Pool` + `Reset`, or a single shared
  deadline). Must preserve the #1172 leak-guard semantics (the timer is the guard
  that releases a caller when a batch wedges).

## Acceptance

With tracing disabled, `UTracer.Start` ≈ 0 on the hot path; `time.NewTimer` no longer
in the top allocations; measurable GC-CPU drop during big-block catchup.

oskarszoon:
## What

`UTracer.Start` did real per-span work even when tracing is disabled — the production default. #1099 gated the `gocore.Stat`, tag and `StartTime` work, but two per-span allocations remained on the hot path: the `&TraceOptions{}` itself and the returned end-function closure. With tracing disabled, at millions of spans per block during big-block legacy catchup, `UTracer.Start` was still ~7.7% CPU and ~178GB of allocations over a 30s profile window (see #1188).

- `TraceOptions` now comes from a `sync.Pool` and is `reset()` for reuse, instead of a fresh heap allocation per span.
- Spans with no finalisation work (tracing disabled + no `gocore.Stat`, prometheus metric, log message or context timeout — e.g. `aerospike:Create`, which passes no options) get a shared, allocation-free `noopEndFn` and recycle the options immediately.
- The end function finalises exactly once, guarded by a per-span `atomic.Bool` captured by the closure (declared after the fast-path return, so the 0-alloc path is untouched). It is **not** a field on the pooled object — a pooled flag is cleared by `reset()` on reissue, which would let a stale end call corrupt a recycled span.

Disabled-path prometheus metrics, logging and context-timeout cancellation are unchanged — they still fire on the slow path exactly as before.

## Benchmarks

Tracing disabled, `#1099` baseline → after:

| case | before | after |
|---|---|---|
| `Start`, no options (`aerospike:Create`) | 272 B/op, 2 allocs/op | **0 B/op, 0 allocs/op** |
| `Start`, 2 tags | 272 B/op, 2 allocs/op | **0 B/op, 0 allocs/op** |
| `Start`, histogram (validator per-tx) | 272 B/op, 2 allocs/op | 116 B/op, 2 allocs/op (baseline, no regression) |

The headline win is the disabled option-less / tag-only span (the millions-per-block UTXO / `aerospike:Create` path) → **0 allocs**. Metric-bearing spans stay at the pre-PR baseline of 2 allocs (the pool removes one alloc; the per-span once-guard adds it back) — no regression.

## Tests

- `go test -race ./util/tracing` — green. New/strengthened: `TestStart_PooledOptionsNoStateBleed`, `TestStart_DisabledTimeoutStillCancels`, `TestStart_DisabledDoubleEndIsSafe` (asserts no double-count), `TestStart_ConcurrentEndFinalisesOnce`, `TestStart_StaleEndAfterRecycleDoesNotCorrupt` (regression for the recycle/ABA hole).
- `go build ./...`, `go vet`, `golangci-lint`, `staticcheck` — clean.

## Scope note

This PR originally also pooled the aerospike batcher per-op submit-wait timers (#1188's second half). That is now **superseded by #1232** (group-completion protocol + `PutBatch`), which removed the per-op `time.NewTimer` churn entirely — a better fix. After merging latest `main`, the timer-pool half was dropped; this PR is now tracing-only.

Closes #1188


ordishs (changes requested):
Solid perf work — both optimizations are correct in the common case. The fast-path condition in `Start` faithfully mirrors everything the slow-path `endFn` does (`recordMetrics`, `logEndMessage`, `stat.AddTime`, `cancelFunc`), `reset()` covers all struct fields, and the timer recycling is safe under the Go 1.23+ semantics (go.mod is `go 1.26.0`). One thing worth resolving before merge, plus one nit.

### 1. The `released` once-guard is defeated by pool recycling (`util/tracing/tracing.go`)

The comment on `released` and `TestStart_ConcurrentEndFinalisesOnce` claim the atomic makes the end function safe to call twice / from multiple goroutines and "prevents a double-Put from aliasing the recycled object across two unrelated spans." That specific guarantee has a hole, because `reset()` stores `released=false` on every reissue:

1. Span S1's `endFn` (call #1) wins the CAS, finalises, `Put(O)`.
2. An unrelated `Start` does `Get() -> O`, `reset()` sets `O.released=false` and configures `O.Histogram` for S2.
3. S1's `endFn` fires a **second** time. Its CAS now sees `false` -> succeeds -> it reads `O.Histogram` (S2's metric), observes it against S1's `start`, then `Put(O)` again. O is now in the pool *and* live in S2, so a later span S3 aliases the same object as S2.

The added concurrent-end test doesn't exercise this: all 16 goroutines race on the *same* object before any recycle, so the CAS naturally serialises them. The gap only appears when a second end call lands *after* the object has been reissued — a real window under a hot pool.

Key question that decides the fix: **is the returned end function ever invoked more than once, or from more than one goroutine, by any real caller?**
- If no (always a single `defer endFn()`): this is a documentation over-claim — please soften the comment and the concurrency test to "must be called exactly once," and it's good to go.
- If yes (multi-goroutine end is genuinely relied on): the CAS guard needs to not live on the recycled object (e.g. a per-span sentinel captured by the closure), otherwise recycling corrupts unrelated spans.

### 2. Nit — `acquireBatchTimer` not deferred in `SetDAHForChildRecords` (`stores/utxo/aerospike/spend.go`)

The pool's own doc says *"Every call MUST be paired with releaseBatchTimer (typically `defer releaseBatchTimer(timer)`)"*, and the other six callsites all use `defer`. This one releases inline after the `select`, so a panic in the timeout branch (e.g. `txID.String()` on a nil `txID`) skips the release. Consequence is benign — the timer just isn't returned to the pool, GC reclaims it — and the pre-PR code had the same panic behavior, so not a blocker. A `defer` here would match the contract and the other callsites.

icellan (approved):
LGTM. Reviewed the tracing `TraceOptions` pooling and the aerospike batcher-timer pooling across all 7 callsites — no blocking issues found.

Key safety points check out:
- Timer recycling relies on Go 1.23+ Stop/Reset semantics (no stale tick); go.mod is 1.26.0, so that holds.
- The end-fn once-guard is a per-span captured `atomic.Bool` declared after the 0-alloc fast-path return, not a field on the pooled object — closes the recycle-aliasing/double-Put window.
- `reset()` clears all 10 `TraceOptions` fields; no state bleed across reuse.
- `SetDAHForChildRecords`'s per-iteration closure correctly scopes the `defer releaseBatchTimer` and preserves the leak-guard/timeout behavior on the panic path.

Non-blocking: the leak-guard goroutine/errCh-send-after-timeout behavior is unchanged by this PR (pre-existing, out of scope).

oskarszoon:
Fixed in 693b65ff4.

You're right — the guard lived on the pooled `TraceOptions` and `reset()` re-opened it on reissue, so a stale end after recycle could re-run finalisation against the next span and double-`Put`. Moving the CAS to the top (2976f6d1a) wasn't enough because the flag itself was on the recycled object.

The guard is now a per-span `atomic.Bool` captured by the end closure, declared *after* the no-op fast-path return — so the disabled hot path stays 0 allocs, and recycling the pooled object can't re-open it. Added `TestStart_StaleEndAfterRecycleDoesNotCorrupt` reproducing your exact scenario (S1 ends → object reissued to S2 → stale S1 end must not touch S2).

On your key question: no caller invokes the end function more than once — every one of the ~293 call sites uses a single `defer end()`, none stash it or fan it into goroutines. So it was a latent hazard, not a live bug, but the guard is correct regardless now.

Histogram/metric-bearing spans are back to the pre-PR baseline of 2 allocs (the transient 1-alloc I'd reported was only reachable via the buggy pooled-object guard); the headline fast-path win (option-less/tag-only disabled spans → 0 allocs) stands. PR description updated with the corrected numbers.

Nit #2: `setDAHForChildRecords` now `defer`s `releaseBatchTimer` via a per-iteration closure, so the timer returns to the pool even if the timeout branch panics — matches the other callsites.

Also merged latest main.

util/tracing/tracing.go
```diff
@@ -505,7 +544,38 @@ func (u *UTracer) Start(ctx context.Context, spanName string, opts ...Options) (
...
+	var ended atomic.Bool
+
 	endFn := func(optionalError ...error) {
+		// Finalise exactly once: only the call that wins the compare-and-swap runs
+		// finalisation and returns options to the pool. A later or concurrent call
+		// returns before touching options, which by then may have been recycled into
+		// an unrelated span. Winning the CAS is what licenses every read below.
+		if !ended.CompareAndSwap(false, true) {
+			return
+		}
```
freemans13, line 577: **ABA hole in the "finalise exactly once" guard (latent — no live trigger today).**

`released` is a field of the pooled `TraceOptions`, and `reset()` sets it back to `false` (line 109) when the object is re-issued. So the once-guarantee is scoped to the *object*, not to the specific `Start()` call / `endFn` closure. Sequence (reproducible single-threaded):

1. `_, _, endA := Start(ctx, "A", WithCounter(cA))`; `endA()` → CAS succeeds, finalises A, `Put(options)`.
2. `_, _, endB := Start(ctx, "B", WithCounter(cB))` → `Get()` returns the same object, `reset()` sets `released=false` and installs `cB`.
3. A stale/duplicate `endA()` (leaked/retried closure, or a duplicate `defer`+explicit call) → `CompareAndSwap(false, true)` now **succeeds** against B's live options: it runs `recordMetrics`/`logEndMessage` against `cB`/B's logger, then **double-`Put`s** the object while B still holds it — a later `Start()` for C can then alias B.

This contradicts the field comment above ("it prevents a double-Put from aliasing the recycled object … across two unrelated spans") and the "safe even if invoked from more than one goroutine" claim. No current caller double-invokes an end function, so there's no live trigger — but the pooling makes a future double-call far more damaging than the pre-PR behaviour (silent metric/log corruption + pool-exclusivity violation vs. a harmless extra call), and `TestStart_DisabledDoubleEndIsSafe` / `TestStart_ConcurrentEndFinalisesOnce` only cover an isolated span, never the `Start()`→`Get()` recycle interleaving.

Fix options: make the guard external to the pooled object — a closure-local `*atomic.Bool` allocated per `Start()`, or a generation counter compared inside the CAS — or soften the comment to state the guarantee holds only absent double-invocation.

oskarszoon: Fixed in 693b65ff4 — went with your first option.

The guard is now a per-span `atomic.Bool` captured by the `endFn` closure, not a field on the pooled `TraceOptions`. `reset()` no longer touches it (the `released` field is gone entirely), so recycling the object can't re-open a prior span's guard. It's declared *after* the no-op fast-path return, so the 0-alloc disabled path is untouched; on the slow path it rides in the end closure.

Added `TestStart_StaleEndAfterRecycleDoesNotCorrupt`, which reproduces exactly the `Start()`→`Get()` recycle interleaving you described (S1 ends → object reissued to S2 → stale S1 `end()` must not observe S2's counter or double-`Put`).

Confirmed no caller double-invokes an end function (single `defer end()` across all ~293 sites), so it was latent — but the guard is correct regardless now, and the field comment was rewritten to match.

ctnguyen (approved):
## Approved

Two well-scoped allocation reductions on the big-block catchup hot path — pooling `TraceOptions` (plus a shared no-op end function) in `UTracer.Start`, and pooling the batcher submit-wait `*time.Timer` across the seven UTXO callsites. The changes are behaviour-preserving and I found no blocking issues.

freemans13 (commented):
Reviewed the tracing changes end-to-end (full `Start` path, not just the diff) and ran the CI checks locally. The code is clean and mergeable as-is — one thing worth reconciling before merge.

## 🔴 Description describes work that isn't in the diff

The mergeable diff is **tracing-only**:

| file | +/− |
|---|---|
| `util/tracing/tracing.go` | +74 / −2 |
| `util/tracing/tracing_test.go` | +219 / −0 |

Section 2 of the description ("Batcher submit-wait `time.NewTimer` churn") and the aerospike batcher-timer work are **not present in the PR head**. Commit `c4f8e53e0` (adding `stores/utxo/aerospike/timer_pool.go`, `acquireBatchTimer`/`releaseBatchTimer`, plus edits to `create.go`/`get.go`/`locked.go`/`spend.go`) exists in the branch's object history but was dropped during a rebase — `timer_pool.go` and `acquireBatchTimer` are absent from the current tree.

As a result the description's aerospike benchmark row, the `setDAHForChildRecords` review-fix bullet, the `TestWaitForLockedResult_BoundedWaitWhenBatcherWedged` test bullet, and one Acceptance bullet all reference code that won't merge here.

**Suggestion:** either restore the aerospike commit if it was meant to ship in this PR, or trim section 2 / the batcher bullets from the description so it reflects the tracing-only change. Worth noting the "Closes #1188" line currently claims both halves of the issue are done.

## ✅ Tracing code — no correctness issues

- **Pool lifecycle** (`Get`→`reset`→`Put`) is safe: `options` is only referenced by the end closure and is `Put` as the last statement, after all reads — no use-after-recycle.
- **The `ended atomic.Bool` once-guard is the right design.** It's a per-span closure local rather than a field on the pooled object, so a stale/duplicate `endFn` call landing after the object was recycled into another span loses its own CAS and never touches the recycled object. `TestStart_StaleEndAfterRecycleDoesNotCorrupt` covers exactly that case.
- **The no-op fast path is provably equivalent to running the slow-path `endFn`.** Checked each finalisation step against its guard: `logEndMessage` early-returns when `Logger==nil || len(LogMessages)==0` (exactly the fast-path condition), `recordMetrics` no-ops when both metrics are nil, and `stat`/`cancelFunc`/`span.End()` are all excluded by the condition. Nothing is silently skipped.
- **`reset()` is correct** — nils `SpanStartOptions` (caller-owned, must not alias) while reusing the `Tags`/`LogMessages` backing arrays.

## 🟡 Minor (note only, non-actionable)

On the slow path, `endFn` returns the object to the pool without clearing the reference-holding fields (`Histogram`, `Counter`, `Logger`, `ParentStat`) — those are cleared on the next `Get→reset`. So an idle pooled object pins those referents. In practice they're long-lived singletons (prometheus metrics, loggers), so it's not a real leak — just flagging.

## CI checks (run locally against the PR head)

- `go build ./util/tracing/` — ✅
- `go vet ./util/tracing/` — ✅ clean
- `go test -race ./util/tracing/` — ✅ 14 tests pass under `-race`
- `golangci-lint run util/tracing/...` — ✅ no issues
- Benchmarks — ✅ claims confirmed: `Start_Disabled_NoOptions` = **0 B/op, 0 allocs/op**; `Start_Disabled_Histogram` = 116 B/op, 2 allocs/op (pre-PR baseline, no regression)

oskarszoon:
@freemans13 description's already tracing-only — you caught the v1 body; I rewrote it last push. Section 2, the aerospike benchmark row, and the setDAH / `WaitForLockedResult` bullets are gone, replaced by a scope note: the batcher-timer half is superseded by #1232 (group-completion). `Closes #1188` is intended — Part 2 landed via #1232, Part 1 lands here.

Thanks for the end-to-end tracing pass. The 🟡 (idle pooled object pinning the Histogram/Counter/Logger singletons until next reset) — leaving as-is: they're long-lived, and clearing on Put adds work to the hot path.

freemans13 (commented):
Re-reviewed after the update. The update was a rebase onto latest `main` plus a description rewrite — the tracing code itself is byte-identical to what I reviewed before, and it re-verifies clean.

## Finding 1 (description ↔ diff mismatch) — resolved ✅

The description is now correctly scoped to the tracing-only change, and the new "Scope note" explains *why* the aerospike batcher-timer half is gone rather than leaving it looking accidentally dropped. I confirmed the reasoning holds:

- **#1232 is merged** (2026-07-06, "group-completion protocol for batched UTXO ops + PutBatch/greedy adoption").
- On current `main` the per-op `time.NewTimer` churn in `stores/utxo/aerospike` is gone — the only `time.NewTimer` left there is the long-lived idle timer in `unmined_iterator.go`, which was always meant to stay.

So `Closes #1188` is legitimate: #1232 took the timer half, this PR takes the tracing half.

## Re-verification on the rebased head

- `go build ./util/tracing/` — ✅
- `go vet ./util/tracing/` — ✅ clean
- `go test -race ./util/tracing/` — ✅ 36 tests pass under `-race` (up from 14 — the rebase pulled in more tracing-package tests from `main`; all green)
- `golangci-lint run util/tracing/...` — ✅ 0 issues

No new code concerns. The minor note from my first pass still stands (pooled objects pin `Histogram`/`Counter`/`Logger`/`ParentStat` referents until the next `reset`) and remains non-actionable — those are long-lived singletons. LGTM.

freemans13 (approved):
Approving. Tracing-only change verified clean: build, `go vet`, `golangci-lint` all clean and `go test -race ./util/tracing` green (36 tests). The pool lifecycle and the per-span `atomic.Bool` once-guard are correct, and the fast-path skip is behaviourally equivalent to the slow path. Description now accurately scoped to the tracing half; the batcher-timer half is covered by the merged #1232, so `Closes #1188` is legitimate.

ordishs (approved):
Verified locally on the PR head: `go test -race ./util/tracing` green, benchmarks reproduce the claimed numbers (0 allocs/op for the disabled no-option/tag paths, 116 B/op / 2 allocs/op for histogram), `go vet`/`staticcheck` clean.

The pooling is sound — the once-guard living in the closure rather than on the pooled object correctly closes the recycle/stale-end hazard, and `TestStart_StaleEndAfterRecycleDoesNotCorrupt` pins it. Fast-path condition covers everything the end function would do on the disabled path.

One non-blocking nit: `reset()` truncates `Tags`/`LogMessages` with `[:0]`, leaving old elements (notably `logMessage.args`) reachable in the backing array until reuse or pool eviction. A `clear()` before truncating would release those references; fine as a follow-up.

PR #1193, opened 2026-07-01
fix(utxo/sql): don't clear parent locked flag on unflagged Unspend

### LINKED ISSUE #1154: [TN-015][medium] utxo/sql: Unspend unconditionally clears parent locked (2PC divergence from Aerospike)
**Severity:** medium · **Confidence:** medium · **Pass:** UTXO integrity
**Component:** `stores/utxo/sql/sql.go`

### Summary
`Unspend` runs `UPDATE transactions SET locked = $2` for every spend, with `locked` defaulting to **false** when no `flagAsLocked` is passed. The Aerospike `unspend` UDF never touches `locked`. So on SQL a spend-rollback can clear an independently-set two-phase-commit `locked` flag on a parent, making its outputs spendable before 2PC completes.

### Evidence
- `stores/utxo/sql/sql.go:2983-2993` — `locked := false; if len(flagAsLocked)>0 {…}`; `q2 = UPDATE transactions SET locked = $2 WHERE id = $1`, per-spend in the loop.
- Callers passing no flag: `services/validator/Validator.go:1461`; `stores/utxo/sql/sql.go:1951`.
- Aerospike `unspend` UDF (`teranode.lua`) clears spending data + DAH only; never `locked`.

### Why it matters
Breaks the "output not spendable until safe" invariant on SQL deployments during an overlapping rollback, and diverges from Aerospike behaviour (consensus-relevant across a mixed fleet).

### Suggested fix
Only update `locked` when `flagAsLocked` is explicitly supplied (or scope the clear to the spend the caller owns). Match Aerospike: don't touch `locked` on unspend unless asked.

### Suggested tests
SQL: `Create(parent, WithLocked(true))`; spend a child; `Unspend(child-spends)` (no flag) → assert parent `locked` still true.

---

ordishs:
Fixes #1154

## Problem
`stores/utxo/sql/sql.go` `Unspend` ran `UPDATE transactions SET locked = $2` for **every** spend, with `locked` defaulting to `false` when no `flagAsLocked` was supplied. A plain spend-rollback therefore cleared a parent's independently-set two-phase-commit `locked` flag, making its outputs spendable before 2PC completed.

This diverged from Aerospike: the `unspend` UDF (`teranode.lua:484-555`) never touches `locked` (it is cleared on `setMined`), and the Aerospike Go wrapper ignores `flagAsLocked` entirely. The dangerous no-flag callers are `services/validator/Validator.go:1461` (`reverseSpends`), the batched-spend rollback (`sql.go:1951`), and `cmd/rewindblockchain`.

## Fix
Apply the `locked` update **only when `flagAsLocked` is explicitly supplied**. When no flag is passed, `locked` is left untouched, matching Aerospike. `setDAH` housekeeping still runs unconditionally, and the deliberate lock/unlock semantics of `ProcessConflicting` (`true`) and `ReverseProcessConflicting` (`false`) are preserved.

## Tests
Added `TestUnspendLockedFlag` (`stores/utxo/sql/sql_test.go`):
- `Create(parent, WithLocked(true))` → spend a child with `IgnoreLocked` → `Unspend(no flag)` asserts parent `Locked` is still `true`.
- Also asserts `Unspend(true)` keeps it locked and `Unspend(false)` clears it (retained semantics).

Verified the test fails against the pre-fix code and passes with the fix.

```
go test -race -tags "testtxmetacache" ./stores/utxo/sql/... ./stores/utxo/
ok  	github.com/bsv-blockchain/teranode/stores/utxo/sql	21.689s
ok  	github.com/bsv-blockchain/teranode/stores/utxo/sql/pruner	1.572s
ok  	github.com/bsv-blockchain/teranode/stores/utxo	2.912s
```
`go vet ./stores/utxo/sql/...` clean.

### DIFF
--- a/stores/utxo/logger/logger.go
+++ b/stores/utxo/logger/logger.go
@@ -201,7 +201,7 @@ func (s *Store) Spend(ctx context.Context, tx *bt.Tx, blockHeight uint32, ignore
 }
 
 func (s *Store) Unspend(ctx context.Context, spends []*utxo.Spend, flagAsLocked ...bool) error {
-	err := s.store.Unspend(ctx, spends, false)
+	err := s.store.Unspend(ctx, spends, flagAsLocked...)
 	spendDetails := make([]string, len(spends))
 
 	for i, spend := range spends {
--- a/stores/utxo/logger/logger_test.go
+++ b/stores/utxo/logger/logger_test.go
@@ -528,23 +528,50 @@ func TestSpend(t *testing.T) {
 	mockStore.AssertExpectations(t)
 }
 
+// TestUnspend proves the decorator forwards flagAsLocked unchanged (#1154).
+// Dropping it and hardcoding false defeats the SQL fix on the ?logging=true path
+// (a no-flag rollback would still write locked=false) and inverts
+// ProcessConflicting's Unspend(..., true) lock-set into a lock-clear.
 func TestUnspend(t *testing.T) {
-	logger := ulogger.TestLogger{}
-	mockStore := &MockStore{}
-	store := New(context.Background(), logger, mockStore).(*Store)
-
 	ctx := context.Background()
 	spends := []*utxo.Spend{createTestSpend(1), createTestSpend(2)}
 	expectedErr := errors.NewError("unspend error")
 
-	// Note: The implementation always passes false as the third argument (ignoring variadic parameter)
-	// The mock receives it as []bool{false} due to variadic expansion
-	mockStore.On("Unspend", ctx, spends, []bool{false}).Return(expectedErr)
+	t.Run("no flag is forwarded as no flag", func(t *testing.T) {
+		mockStore := &MockStore{}
+		store := New(ctx, ulogger.TestLogger{}, mockStore).(*Store)
 
-	err := store.Unspend(ctx, spends, true) // flagAsLocked is ignored in implementation
+		mockStore.On("Unspend", ctx, spends, []bool(nil)).Return(expectedErr)
 
-	assert.Equal(t, expectedErr, err)
-	mockStore.AssertExpectations(t)
+		err := store.Unspend(ctx, spends)
+
+		assert.Equal(t, expectedErr, err)
+		mockStore.AssertExpectations(t)
+	})
+
+	t.Run("explicit true is forwarded unchanged", func(t *testing.T) {
+		mockStore := &MockStore{}
+		store := New(ctx, ulogger.TestLogger{}, mockStore).(*Store)
+
+		mockStore.On("Unspend", ctx, spends, []bool{true}).Return(expectedErr)
+
+		err := store.Unspend(ctx, spends, true)
+
+		assert.Equal(t, expectedErr, err)
+		mockStore.AssertExpectations(t)
+	})
+
+	t.Run("explicit false is forwarded unchanged", func(t *testing.T) {
+		mockStore := &MockStore{}
+		store := New(ctx, ulogger.TestLogger{}, mockStore).(*Store)
+
+		mockStore.On("Unspend", ctx, spends, []bool{false}).Return(expectedErr)
+
+		err := store.Unspend(ctx, spends, false)
+
+		assert.Equal(t, expectedErr, err)
+		mockStore.AssertExpectations(t)
+	})
 }
 
 func TestDelete(t *testing.T) {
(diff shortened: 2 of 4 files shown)


freemans13:
### Review finding: the logging decorator silently defeats this fix

The core diff looks correct and minimal — nice catch on the Aerospike parity. One follow-up that's worth folding into this PR:

**`stores/utxo/logger/logger.go:204` discards `flagAsLocked` and hardcodes `false`:**

```go
func (s *Store) Unspend(ctx context.Context, spends []*utxo.Spend, flagAsLocked ...bool) error {
    err := s.store.Unspend(ctx, spends, false)   // ← variadic dropped, literal false
```

Before this PR that was harmless (no-flag already defaulted to `false`). After this PR the no-flag and explicit-`false` cases diverge, and the decorator collapses them: it always passes a literal `false`, so the wrapped SQL store sees `flagLockedProvided == true` and runs `q2` with `locked = false`.

When the UTXO store URL is configured `?logging=true` (wired at `stores/utxo/factory/utxo.go:116-118`), this means:

- **The #1154 fix is defeated on the logging path** — a no-flag `reverseSpends`/batched rollback still writes `locked = false` on the parent.
- **`ProcessConflicting` is inverted** — `Unspend(ctx, affectedParentSpends, true)` (`process_conflicting.go:235`) gets flattened to `false`, turning the intended lock-*set* into a lock-*clear*.

Since it's gated on `?logging=true` (diagnostic path), it's lower severity, but it's a one-line fix that makes the fix complete and also correctly restores the `true` intent:

```go
err := s.store.Unspend(ctx, spends, flagAsLocked...)
```

(`stores/txmetacache/txmetacache.go:878` already forwards the flag this way.)

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

icellan (approved):

ordishs:
Good catch — confirmed and fixed in 24fc38c59.

`stores/utxo/logger/logger.go:204` now forwards the variadic unchanged:

```go
err := s.store.Unspend(ctx, spends, flagAsLocked...)
```

I also found `TestUnspend` (`logger_test.go`) was pinning the old behavior — it called `Unspend(..., true)` but asserted the mock received `[]bool{false}`, with a comment that the flag "is ignored in implementation." Rewrote it into three sub-cases asserting no-flag / `true` / `false` are each forwarded verbatim; it now fails against the pre-fix decorator and passes with it.

```
go test -race -tags "testtxmetacache" ./stores/utxo/logger/
ok  	github.com/bsv-blockchain/teranode/stores/utxo/logger	1.556s
```
`go vet` clean.

freemans13:
### Re-review ✅

Thanks for the quick turnaround. `fix(utxo/logger): forward flagAsLocked in Unspend decorator` resolves the finding:

- `logger.go` now forwards `flagAsLocked...` unchanged — the `?logging=true` path no longer defeats the #1154 fix, and `ProcessConflicting`'s `Unspend(..., true)` is no longer inverted to a lock-clear.
- The new `TestUnspend` subtests correctly distinguish **no-flag → `[]bool(nil)`** from **explicit `false` → `[]bool{false}`** (and `true` → `[]bool{true}`). The no-flag case is the meaningful guard — it fails against the old hardcoded-`false` decorator, so it's a genuine regression test.

The `sql.go` core fix is unchanged and correct. No new issues. LGTM.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

freemans13 (approved):
Approved. The SQL `Unspend` fix correctly gates the `locked` update behind an explicitly-supplied `flagAsLocked`, matching Aerospike semantics and resolving #1154, and the logger decorator now forwards the flag unchanged. Both changes are covered by targeted regression tests. LGTM 🚀

ctnguyen (approved):
## Approved

This is a tight, well-reasoned fix for a genuine correctness bug: the SQL `Unspend` unconditionally ran `UPDATE transactions SET locked = false` on every spend-rollback, silently clearing a parent's independently-set two-phase-commit `locked` flag and diverging from the Aerospike store, which never mutates `locked` on unspend. Gating the `q2` write behind `flagLockedProvided` (an explicitly-supplied `flagAsLocked`) restores parity while preserving the deliberate lock/unlock semantics of `ProcessConflicting` (`true`) and `ReverseProcessConflicting` (`false`), and `setDAH` housekeeping is correctly left unconditional. The regression test `TestUnspendLockedFlag` exercises no-flag / `true` / `false` and exercises both the `q1` and `qFindTxID` fallback paths; the follow-up logger-decorator fix (forwarding the variadic instead of hardcoding `false`) closes the `?logging=true` gap that would otherwise have defeated the fix and inverted `ProcessConflicting`'s intent. The Aerospike-parity premise checks out against `un_spend.go` (the flag is accepted but never forwarded to the Lua script).

PR #1194, opened 2026-07-01
refactor(blockchain/sql): rename GenerationalCache.Begin to NewOp

liam:
## What

Renames `GenerationalCache.Begin` → `NewOp` in `stores/blockchain/sql/` (method, all ~30 call sites, tests, and doc comments). Pure mechanical rename via `gopls`, no behaviour change.

## Why

`GenerationalCache` is a response-result cache, not a `database/sql` transaction. Its `Begin()` returns a `*CacheOperation` that exposes only `Get()`/`Set()` and has **no `Rollback()`**.

SonarQube's `godre:S8168` rule pattern-matches the method name `Begin` as a transaction and raises **23 false-positive reliability-HIGH issues** ("add `defer cacheOp.Rollback()`"). The suggested fix would not compile - there is nothing to roll back and no resource to release.

Renaming the constructor to `NewOp` removes the false positives at the source, avoiding per-issue "won't fix" triage or a quality-profile exclusion that would also hide any genuine future issue.

## Verification

- `go build ./stores/blockchain/sql/` - ok
- `go vet ./stores/blockchain/sql/` - clean
- `go test -tags testtxmetacache ./stores/blockchain/sql/` - ok

### DIFF
--- a/stores/blockchain/sql/GetBestBlockHeader.go
+++ b/stores/blockchain/sql/GetBestBlockHeader.go
@@ -93,7 +93,7 @@ func (s *SQL) GetBestBlockHeader(ctx context.Context) (*model.BlockHeader, *mode
 
 	// Begin cache-safe query - captures generation to prevent stale writes
 	cacheID := chainhash.HashH([]byte("GetBestBlockHeader"))
-	cacheOp := s.responseCache.Begin(cacheID)
+	cacheOp := s.responseCache.NewOp(cacheID)
 
 	cached := cacheOp.Get()
 	if cached != nil {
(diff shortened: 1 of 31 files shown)


oskarszoon (approved):

ordishs (approved):
Pure mechanical rename `GenerationalCache.Begin` → `NewOp` across 31 files (+55/-55). Verified: no `.Begin(` call sites remain, type is package-private with no external callers, no `NewOp` symbol collision, and `go build` + `go vet` pass clean on the package. Sound motivation (removes 23 SonarQube transaction false positives at the source) and semantically neutral. LGTM.

PR #1197, opened 2026-07-01
fix(blockvalidation): gate secret-mining verdict on chainwork, not fork depth

### LINKED ISSUE #1160: [P2P-002][medium] blockvalidation: valid deep reorg flagged 'secret mining' and peer quarantined
**Severity:** medium · **Confidence:** medium · **Pass:** P2P / consensus
**Component:** `services/blockvalidation/catchup.go` (`checkSecretMiningFromCommonAncestor`)

### Summary
`checkSecretMiningFromCommonAncestor` treats "common ancestor more than `SecretMiningThreshold` (~99) blocks back" as an attack and pins the peer's reputation to 5.0 (excluded from selection) — but a legitimate deeper reorg carrying more validated work is indistinguishable to this depth-only check, so every honest peer serving a better deep chain is marked malicious and the node refuses to follow the heavier chain.

### Evidence
- `catchup.go:1237-1275` computes `blocksBehind = currentHeight - commonAncestor.Height`, and if > `SecretMiningThreshold` calls `reportCatchupMalicious(…, "secret_mining")` and aborts.
- Default `SecretMiningThreshold = CoinbaseMaturity-1` (`settings/settings.go:354`).
- `RecordCatchupMalicious` pins reputation to 5.0 (`peer_registry.go:331-336,553-556`).
- Mitigation: soft-quarantine only (no IP ban); `ReconsiderBadPeers` recovers after a cooldown (`peer_registry.go:909-943`).

### Why it matters
The realistic downside (refusing a valid heavier chain because the fork is deep, and quarantining honest peers) is not surfaced as a deliberate policy decision.

### Suggested fix
Gate the penalty on **work, not depth** — only flag malicious if the offered chain does not exceed local validated work. Separate "refuse to auto-apply a deep reorg" (policy) from "this peer is malicious" (reputation).

### Suggested tests
Peer forking `SecretMiningThreshold+k` back with higher validated chainwork → assert not pinned to reputation 5 (or the policy/malicious distinction is observable). An actual withheld shorter-work deep chain → assert it IS flagged.

---

ordishs:
## Summary

Fixes #1160.

`checkSecretMiningFromCommonAncestor` (`services/blockvalidation/catchup.go`) treated a common ancestor more than `SecretMiningThreshold` (~99) blocks back as an attack, pinning the peer's reputation to 5.0 (excluded from selection) and aborting catchup. A **legitimate deeper reorg carrying more validated work is indistinguishable to a depth-only check**, so every honest peer serving a better deep chain was flagged malicious and the node refused to follow the heavier chain.

## What changed

- The secret-mining verdict is now **gated on cumulative proof-of-work, not fork depth**. When the depth threshold is exceeded, the offered chain's work (common-ancestor chainwork + Σ per-header work of the peer's offered headers) is compared against the local validated best-chain work:
  - **Offered work > local work** → legitimate heavier chain: follow it, return `nil`, and **do not** touch the peer's reputation.
  - **Offered work ≤ local work** → deep fork with no extra work (withheld shorter chain): flag malicious and abort, as before.
- If the local best block header can't be read, the catchup aborts **without** penalising the peer — separating the "refuse to auto-apply this reorg" policy decision from the "this peer is malicious" reputation decision. Deep-reorg depth policy remains independently enforced by `validateForkDepth` against coinbase maturity.
- New helper `offeredChainHasMoreWork` computes the comparison using the existing `work.CalcBlockWork` and the big-endian `ChainWork` convention already used in `services/blockchain/Difficulty.go`.

## Tests

`TestCheckSecretMiningWorkGating` (new) asserts the reputation distinction via a recording P2P client:
- heavier deep reorg → no error, `RecordCatchupMalicious` **not** called;
- deep fork with no extra work → error `secretly mined chain`, `RecordCatchupMalicious` **called** for the peer;
- best-block-header lookup failure → aborts without flagging malicious.

The existing `TestCatchup` "Too Far Behind" case was updated to set up the work comparison (local chain heavier) so it continues to exercise the malicious path.

### Verification
- `go build ./services/blockvalidation/` — clean
- `go vet ./services/blockvalidation/` — pass
- `go test -race -tags testtxmetacache ./services/blockvalidation/` — `ok` (full package)
- `golangci-lint run --new-from-rev=upstream/main ./services/blockvalidation/` — 0 issues

### DIFF
--- a/services/blockvalidation/catchup.go
+++ b/services/blockvalidation/catchup.go
@@ -3,6 +3,8 @@ package blockvalidation
 
 import (
 	"context"
+	"encoding/binary"
+	"math/big"
 	"net/url"
 	"strings"
 	"sync/atomic"
@@ -15,6 +17,7 @@ import (
 	"github.com/bsv-blockchain/teranode/pkg/fileformat"
 	"github.com/bsv-blockchain/teranode/services/blockchain"
 	"github.com/bsv-blockchain/teranode/services/blockchain/blockchain_api"
+	"github.com/bsv-blockchain/teranode/services/blockchain/work"
 	"github.com/bsv-blockchain/teranode/services/blockvalidation/catchup"
 	"github.com/bsv-blockchain/teranode/util/blockassemblyutil"
 	"github.com/bsv-blockchain/teranode/util/tracing"
@@ -612,7 +615,18 @@ func (u *Server) validateForkDepth(catchupCtx *CatchupContext) error {
 func (u *Server) checkSecretMining(ctx context.Context, catchupCtx *CatchupContext) error {
 	u.logger.Debugf("[catchup][%s] Step 4: Checking for secret mining", catchupCtx.blockUpTo.Hash().String())
 
-	return u.checkSecretMiningFromCommonAncestor(ctx, catchupCtx.blockUpTo, catchupCtx.peerID, catchupCtx.baseURL, catchupCtx.commonAncestorHash, catchupCtx.commonAncestorMeta)
+	// The headers the peer offers beyond the common ancestor form the candidate chain
+	// whose cumulative work we weigh against our local chain. Not yet filtered at this
+	// point, so slice directly after the common ancestor index.
+	var offeredHeaders []*model.BlockHeader
+	if catchupCtx.headersFetchResult != nil {
+		peerHeaders := catchupCtx.headersFetchResult.Headers
+		if catchupCtx.commonAncestorIndex >= 0 && catchupCtx.commonAncestorIndex+1 < len(peerHeaders) {
+			offeredHeaders = peerHeaders[catchupCtx.commonAncestorIndex+1:]
+		}
+	}
+
+	return u.checkSecretMiningFromCommonAncestor(ctx, catchupCtx.blockUpTo, catchupCtx.peerID, catchupCtx.baseURL, catchupCtx.commonAncestorHash, catchupCtx.commonAncestorMeta, offeredHeaders)
 }
 
 // filterHeaders filters headers to only those after the common ancestor that we don't have.
@@ -1226,7 +1240,14 @@ func getLowestCheckpointHeight(checkpoints []chaincfg.Checkpoint) uint32 {
 }
 
 // checkSecretMiningFromCommonAncestor detects if a peer withheld blocks (secret mining).
-// Checks if common ancestor is too far behind, indicating potential attack.
+//
+// A common ancestor more than SecretMiningThreshold blocks back is a necessary but not
+// sufficient signal: a legitimate reorg carrying more accumulated proof-of-work can fork
+// just as deeply. The malicious verdict is therefore gated on WORK, not depth — a deep
+// fork is only treated as secret mining when the chain the peer offers fails to exceed
+// our local validated chainwork. This keeps the reputation penalty ("peer is malicious")
+// separate from the policy decision of whether to auto-apply a deep reorg, which is
+// enforced independently by validateForkDepth against coinbase maturity.
 //
 // Parameters:
 //   - ctx: Context for cancellation
@@ -1234,10 +1255,11 @@ func getLowestCheckpointHeight(checkpoints []chaincfg.Checkpoint) uint32 {
 //   - baseURL: Peer URL for metrics
 //   - commonAncestorHash: Hash of the common ancestor
 //   - commonAncestorMeta: Metadata of the common ancestor
+//   - offeredHeaders: Peer's headers after the common ancestor (the candidate chain)
 //
 // Returns:
-//   - error: If secret mining is detected
-func (u *Server) checkSecretMiningFromCommonAncestor(ctx context.Context, blockUpTo *model.Block, peerID, baseURL string, commonAncestorHash *chainhash.Hash, commonAncestorMeta *model.BlockHeaderMeta) error {
+//   - error: If secret mining is detected, or the deep fork cannot be safely followed
+func (u *Server) checkSecretMiningFromCommonAncestor(ctx context.Context, blockUpTo *model.Block, peerID, baseURL string, commonAncestorHash *chainhash.Hash, commonAncestorMeta *model.BlockHeaderMeta, offeredHeaders []*model.BlockHeader) error {
 	// Check whether the common ancestor is more than X blocks behind our current chain.
 	// This indicates potential secret mining.
 	currentHeight := u.utxoStore.GetBlockHeight()
@@ -1255,7 +1277,23 @@ func (u *Server) checkSecretMiningFromCommonAncestor(ctx context.Context, blockU
 		return nil
 	}
 
-	// The chain is potentially a secretly mined chain
+	// The fork is deep. Weigh the offered chain's work against our own before penalising.
+	heavier, err := u.offeredChainHasMoreWork(ctx, commonAncestorMeta, offeredHeaders)
+	if err != nil {
+		// We cannot prove the offered chain is heavier, so we must not follow it here; but
+		// we also cannot prove the peer is malicious, so we do NOT touch its reputation.
+		// Abort this catchup and let peer selection try again.
+		u.logger.Warnf("[catchup][%s] deep fork from peer %s (ancestor height %d, %d blocks behind) but chainwork comparison failed: %v - aborting without flagging malicious", blockUpTo.Hash().String(), baseURL, commonAncestorMeta.Height, blocksBehind, err)
+		return errors.NewProcessingError("[catchup][%s] unable to compare chainwork for deep fork from common ancestor at height %d", blockUpTo.Hash().String(), commonAncestorMeta.Height, err)
+	}
+
+	if heavier {
+		// A legitimate heavier chain that happens to fork deep. Follow it; do not penalise the peer.
+		u.logger.Infof("[catchup][%s] deep reorg (%d blocks) from peer %s carries more validated work than local chain - following heavier chain, not flagging secret mining", blockUpTo.Hash().String(), blocksBehind, baseURL)
+		return nil
+	}
+
+	// A deep fork that offers no additional work: potential secret mining (withheld shorter chain).
 	u.logger.Errorf("[catchup][%s] is potentially a secretly mined chain from common ancestor %s at height %d, ignoring", blockUpTo.Hash().String(), commonAncestorHash.String(), commonAncestorMeta.Height)
 
 	// Record error metric for secret mining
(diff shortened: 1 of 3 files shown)


services/blockvalidation/catchup.go
```diff
@@ -1274,6 +1312,44 @@ func (u *Server) checkSecretMiningFromCommonAncestor(ctx context.Context, blockU
...
+// Parameters:
+//   - ctx: Context for the best-block-header lookup
+//   - commonAncestorMeta: Metadata of the common ancestor (supplies its cumulative work)
+//   - offeredHeaders: Peer's headers after the common ancestor
+//
+// Returns:
+//   - bool: True if the offered chain has strictly more cumulative work than the local chain
+//   - error: If the local best block header cannot be retrieved
+func (u *Server) offeredChainHasMoreWork(ctx context.Context, commonAncestorMeta *model.BlockHeaderMeta, offeredHeaders []*model.BlockHeader) (bool, error) {
+	_, bestMeta, err := u.blockchainClient.GetBestBlockHeader(ctx)
```
freemans13: **Minor (consistency):** `blocksBehind` is computed from `u.utxoStore.GetBlockHeight()` (~line 1262), but `localWork` here comes from `blockchainClient.GetBestBlockHeader`. These are two different views of the chain tip that can momentarily disagree during catchup. Not a bug — the work comparison is the authoritative gate — but deriving `blocksBehind` from `bestMeta.Height` too would make the depth check and the work check read the same tip.

services/blockvalidation/catchup.go
```diff
@@ -609,7 +612,18 @@ func (u *Server) validateForkDepth(catchupCtx *CatchupContext) error {
...
 
-	return u.checkSecretMiningFromCommonAncestor(ctx, catchupCtx.blockUpTo, catchupCtx.peerID, catchupCtx.baseURL, catchupCtx.commonAncestorHash, catchupCtx.commonAncestorMeta)
+	// The headers the peer offers beyond the common ancestor form the candidate chain
+	// whose cumulative work we weigh against our local chain. Not yet filtered at this
+	// point, so slice directly after the common ancestor index.
+	var offeredHeaders []*model.BlockHeader
+	if catchupCtx.headersFetchResult != nil {
+		peerHeaders := catchupCtx.headersFetchResult.Headers
+		if catchupCtx.commonAncestorIndex >= 0 && catchupCtx.commonAncestorIndex+1 < len(peerHeaders) {
+			offeredHeaders = peerHeaders[catchupCtx.commonAncestorIndex+1:]
```
freemans13, line 625: **Minor (edge case / philosophy):** when `offeredHeaders` ends up empty here (nil `headersFetchResult`, or `commonAncestorIndex+1 >= len`), `offeredChainHasMoreWork` sums only the ancestor's work → always `<= local` → the peer gets flagged **malicious** and pinned to reputation 5.0. That contradicts the principle this PR applies to the `GetBestBlockHeader` failure path (uncertainty → abort without penalising). It's practically unreachable for a genuine deep fork (the peer must have offered headers to reach this check), so it matches prior behaviour — but treating "couldn't determine the offered chain" like the header-lookup-failure path (abort, don't flag) would be more consistent.

freemans13 (commented):
Reviewed the work-gating change — it does exactly what #1160 asked, the chainwork math is correct (big-endian `SetBytes`, `Bits` decoded consistently with `work.CalculateWork`), and the fake-chainwork angle is closed because offered headers pass `validateBatchHeaders` PoW validation during fetch before this check runs. Nice separation of the reputation penalty from the deep-reorg policy decision.

Two optional, non-blocking observations below.

ordishs:
Thanks @freemans13 — both minor points addressed in 05324ae7b:

1. **Same-tip consistency:** `blocksBehind` now derives from the blockchain best block header (`GetBestBlockHeader`) rather than `utxoStore.GetBlockHeight()`, so the depth trigger and the work gate read the same tip. The best header is fetched once and its cumulative work reused (dropping the duplicate lookup); a read failure aborts the catchup without touching reputation.

2. **Empty offered headers:** a deep fork with no headers to weigh is now treated like the header-lookup-failure path — abort without flagging malicious — instead of scoring the ancestor-only work as a shorter chain. Added a dedicated test case.

Re: the bot's scope note (coinbase-maturity path flags depths > CoinbaseMaturity before this gate) — agreed that's the intended scope of #1160 and out of scope here; the work-gate covers the `SecretMiningThreshold < blocksBehind ≤ CoinbaseMaturity` band. Happy to file a follow-up if we want the coinbase-maturity abort to also stop pinning honest peers.

Full `blockvalidation` suite green under `-race`, vet clean, no new lint.

freemans13:
Re-reviewed after the update (head `e98e1765c`). Both earlier observations are cleanly resolved:

1. **Tip-source consistency** — `checkSecretMiningFromCommonAncestor` now reads the best-header tip once via `GetBestBlockHeader` and uses `bestMeta.Height` for the depth trigger and `bestMeta.ChainWork` for the work gate, so both reason about the same tip. Refactoring `offeredChainHasMoreWork` into a pure function taking `localChainWork []byte` also removes the double read — nice.

2. **Unweighable deep fork** — the empty-`offeredHeaders` case now aborts without penalising the peer (`"deep fork with no offered headers to compare chainwork"`), matching the same uncertainty-≠-malice principle already applied to the `GetBestBlockHeader` failure path.

Good catch on the follow-up test fix, too: giving the local chain height-proportional chainwork in `TestCatchup_NoMinedSetChurnOnRejectedFork` is necessary now that a zero-chainwork mock tip would make any offered fork look heavier and silently bypass the check.

Verified locally on `e98e1765c`: `go build` + `go vet ./services/blockvalidation/` clean, `go test -race -tags testtxmetacache -run 'TestCheckSecretMiningWorkGating|TestCatchup' ./services/blockvalidation/` → 163 passed. LGTM.

ctnguyen (approved):
Review at commit [e98e1765c25a](https://github.com/bsv-blockchain/teranode/commit/e98e1765c25a)

This PR reworks `checkSecretMiningFromCommonAncestor` so a deep fork is only branded "secret mining" when the peer's offered chain fails to exceed the local validated chainwork, decoupling the peer-reputation penalty from the deep-reorg policy decision. The chainwork math and the uncertainty-vs-malice separation are sound. The most significant finding (ChiR1) is a real regression — the work sum trusts an unlinked, un-deduplicated header set, so a peer can cheaply inflate it — but its realized harm is bounded and its reach is narrow, so it is recommended hardening rather than a merge blocker. There are no blocking issues.

## What's done well
The chainwork math is sound and carefully aligned with existing conventions: the big-endian `SetBytes` reconstruction matches `calculateAndPrepareChainWork` storage, and the per-header `binary.LittleEndian.Uint32(header.Bits.CloneBytes())` + `work.CalcBlockWork` path is byte-for-byte identical to `work.CalculateWork`. Per-header PoW validation at fetch (`validateBatchHeaders`) does prevent a peer from *forging* high-work headers out of nothing — a fabricated header must carry real PoW to be counted. (What it does **not** prevent is re-counting a genuine header's work via duplication or non-contiguous padding, because linkage is only checked downstream of this gate — see ChiR1.) The uncertainty-vs-malice separation is a genuine improvement — both the best-header read failure and the empty-offered-headers case abort the catchup without pinning the peer's reputation, which is the right instinct. The new `TestCheckSecretMiningWorkGating` covers all three verdicts crisply via a recording P2P client, and the earlier-round reviewer feedback (same-tip consistency, empty-offered-headers path) was addressed cleanly.

## Blocking issues

None. The most significant finding (ChiR1) is a genuine regression with a cheap fix, but its realized harm is bounded (no consensus break, no data loss) and it is exploitable only in a narrow fork-depth band, so it is recommended hardening rather than a merge blocker (see the reachability note in ChiR1).

## Non-blocking issues

### ChiR1 — Offered-work sum trusts an unlinked, un-deduplicated header set → cheap bypass of the malicious flag (high-severity hardening)
**Problem:** `offeredChainHasMoreWork` (catchup.go, new helper) sums `work.CalcBlockWork` over *every* entry in `offeredHeaders` (the `peerHeaders[commonAncestorIndex+1:]` slice produced in `checkSecretMining`). Those headers were only validated **per header** — PoW, merkle root, timestamp — at fetch time (`validateBatchHeaders` catchup.go:1296, `ParseBlockHeaders` helpers.go:52). Nothing on the path from fetch to this gate checks that they form a contiguous chain or are distinct: `ValidateChainConnection` (catchup/common_ancestor.go:76) is unused in production, and linkage is first enforced at **Step 7** `BuildFromHeaders` ("chain broken", header_chain_cache.go:120) and **Step 8** `verifyChainContinuity` (catchup.go:251) — both *after* **Step 5** `checkSecretMining` (catchup.go:191). A malicious secret-miner controls its own HTTP header response and can pad `offeredHeaders` with repeated copies of a single valid-PoW header; each duplicate re-counts that header's work with zero extra hashing, so `offeredChainHasMoreWork` returns true, the check returns `nil`, and the peer escapes `RecordCatchupMalicious`.
**Why it matters:** This diff *introduces* a trivial evasion of the very reputation heuristic it reworks — the pre-PR depth-only trigger had no such bypass. The realized harm is confined to reputation-flag evasion plus mild catchup churn: the non-contiguous/duplicate chain is still rejected downstream at Step 7/8, so there is no consensus break, no data loss, and the bogus chain is never applied. Reach is also narrow — as ChiR2 details, this gate runs only after `validateForkDepth` (catchup.go:584) rejects `forkDepth > CoinbaseMaturity (100)`, and the deep branch needs `blocksBehind > SecretMiningThreshold (99)`, so the offered-work sum is consulted only at fork depth ~100, i.e. the attacker needs a genuine common ancestor almost exactly coinbase-maturity deep. Bounded impact plus narrow reach is why this is recommended hardening rather than a merge blocker; it is nonetheless the top finding because the PR's sole purpose is this reputation gate and the fix is cheap.
**Fix:** Before trusting the summed work, require the offered headers to form a contiguous chain rooted at the common ancestor — `offeredHeaders[0].HashPrevBlock == commonAncestorHash` and each `offeredHeaders[i].HashPrevBlock == offeredHeaders[i-1].Hash()`. The linkage logic already exists in `ValidateChainConnection`, but note it is a method on `*CommonAncestorFinder` (catchup/common_ancestor.go:76), not on `*Server`, so `checkSecretMiningFromCommonAncestor` must either invoke it through a `CommonAncestorFinder` instance or replicate the check inline; alternatively, dedup the offered headers by hash before summing. Treat a non-contiguous/duplicate-padded set as uncertainty (abort without flagging), consistent with the empty-headers and header-read-failure paths — not as heavier work. Add a test with a duplicate-padded offered set asserting it does **not** bypass the malicious flag.

### ChiR2 — Work gate is reachable only in a one-block depth band; #1160's symptom persists for deeper forks
**Problem:** `checkSecretMining` (Step 5) runs only *after* `validateForkDepth` (Step 4), which rejects any fork where `forkDepth > CoinbaseMaturity` and *itself* flags the peer malicious via `recordMaliciousAttempt(..., "coinbase_maturity_violation")` (catchup.go:584-595). With the shipped defaults — `SecretMiningThreshold = CoinbaseMaturity - 1 = 99` (settings/blockvalidation_settings.go:38, settings/settings.go:359) and `CoinbaseMaturity = 100` — the new work gate is only consulted when `99 < blocksBehind ≤ 100`. A legitimate heavier reorg deeper than coinbase maturity — precisely the #1160 case of an honest peer serving a better deep chain — is still penalised and aborted by `validateForkDepth` before this gate is ever reached. (The band is slightly wider than one block in practice because `blocksBehind` here now reads the blockchain-store tip while `forkDepth` reads the UTXO-store tip, but the limitation stands.)
**Why it matters:** The PR's stated goal (stop flagging honest peers on deep heavier chains) is only partially met; the most common deep-reorg case still hits the coinbase-maturity path.
**Fix:** No code change required in this PR — the maintainer has confirmed this is the intended scope. Track the follow-up so `validateForkDepth`'s coinbase-maturity abort also stops pinning honest-but-heavier peers (or apply the same work-gate before recording maliciousness there), and note the residual limitation in the issue.

### ChiR3 — Boundary cases of the `offeredHeaders` slicing lack direct coverage
**Problem:** The happy-path slicing added in `checkSecretMining` — `peerHeaders[commonAncestorIndex+1:]` guarded by `commonAncestorIndex >= 0 && commonAncestorIndex+1 < len(peerHeaders)` (catchup.go, Step 5) — *is* exercised end-to-end: `TestCatchup_NoMinedSetChurnOnRejectedFork/SecretMiningDepth` (catchup_fork_handling_test.go:703) drives the full `server.catchup(...)` pipeline through `checkSecretMining` with a non-empty offered slice, and that test's setup was updated by this diff. What is not directly covered are the boundary cases: `commonAncestorIndex == -1`, a nil `headersFetchResult`, and an ancestor at the last index (all three → empty slice → the "no offered headers" abort). The guard is correct as written and verified — each of these yields nil→empty→abort — so this is a coverage gap, not a latent misroute.
**Why it matters:** Nothing wrong today, but these guard boundaries are easy to break in a future refactor and no test pins the empty-slice abort behaviour.
**Fix (defensive, optional):** Add a small unit test on `checkSecretMining` (or a table over `commonAncestorIndex` / nil-`headersFetchResult` values) asserting the derived `offeredHeaders` slice and that the last-index / negative-index cases reach the empty-headers abort. Extracting the slice into a pure helper would make this trivial to assert.

## Recap

| ChiR1 | Offered-work sum trusts unlinked/duplicated headers → cheap malicious-flag bypass (narrow reach, no consensus harm) | | 55% |
| ChiR2 | Work gate only reached in narrow depth band | | 35% |
| ChiR3 | Boundary cases of offeredHeaders slicing untested | | 10% |

galt-tr (approved):

PR #1198, opened 2026-07-01
fix(blockassembly): stable-sort unmined reload with parents-first tiebreak

### LINKED ISSUE #1157: [TN-018][medium] blockassembly: unmined-tx reload sorts by ms CreatedAt with unstable sort (drops same-ms parent/child)
**Severity:** medium · **Confidence:** high · **Pass:** concurrency/ordering / tests
**Component:** `services/blockassembly/BlockAssembler.go`

### Summary
On startup, Block Assembly reloads unmined txs ordered by `CreatedAt` (`UnixMilli`) using `sort.Slice` (unstable, no tiebreak). A same-millisecond parent/child can sort child-first; on the default in-memory path `validateParentChain` then **drops the child and cascades to descendants** rather than reordering — silently shedding part of the mempool from the mining candidate until a later restart/resubmit.

### Evidence
- `BlockAssembler.go:2686-2689` (in-memory) and `:3174-3176` (disk-sort): `sort.Slice(...CreatedAt < ...CreatedAt)`.
- `CreatedAt` = `UnixMilli()` (`stores/utxo/sql/unmined_iterator.go:242`, `stores/utxo/aerospike/create.go:907`).
- Default flags `OnRestartValidateParentChain=true`, `OnRestartRemoveInvalidParentChainTxs=true` (`settings.go:293/295`); drop+cascade `BlockAssembler.go:2222`, `:2232-2247`.

### Why it matters
Post-restart fee/mining-candidate completeness on a busy node: a burst of same-ms parent/child txs can be partially discarded. Not consensus-unsafe (any in-block order is valid post-Genesis; txs remain in the UTXO store), but a real timing-dependent throughput regression.

### Suggested fix
Use `sort.SliceStable` with a parents-first tiebreak for equal `CreatedAt` (or a DAG-rank pre-pass); ensure the iterator yields parents before children on ties.

### Suggested tests
Unmined set with parent and child sharing `CreatedAt`; run reload across many shuffles → assert child retained and ordered after parent.

---

ordishs:
Closes #1157

## Problem

On startup, Block Assembly reloads unmined transactions ordered by `CreatedAt` using an unstable `sort.Slice` with no tiebreak. `CreatedAt` is stamped per transaction with **no enforced gap** and **coarse resolution**:

- **Aerospike** (`create.go:907`): `time.Now().UnixMilli()` per `Create` — millisecond resolution.
- **SQLite** (`sql.go`): `inserted_at` defaults to `CURRENT_TIMESTAMP`, emitted as second-precision `"YYYY-MM-DD HH:MM:SS"` — **every transaction created in the same second shares a `CreatedAt`**.
- **Postgres**: `TIMESTAMPTZ` truncated by `UnixMilli()` — same-ms collisions under load.

The tx dependency only guarantees `parent.CreatedAt ≤ child.CreatedAt` in wall-clock; it does **not** guarantee a gap. So a same-timestamp parent/child can sort child-first. `validateParentChain` (`BlockAssembler.go:2222`) then treats parent-after-child as invalid ordering and, with the default flags `OnRestartValidateParentChain=true` / `OnRestartRemoveInvalidParentChainTxs=true`, **drops the child and cascades to its descendants** (`:2232-2247`) — silently shedding part of the mining candidate until a later restart or resubmit.

Not consensus-unsafe (any in-block order is valid post-Genesis; txs remain in the UTXO store), but a real timing-dependent mining-candidate completeness regression.

## Fix

- **In-memory path** (`BlockAssembler.go:2686`): replace the unstable sort with `stableSortUnminedByCreatedAt` — a stable sort by `CreatedAt` followed by a per-run topological reorder that places in-set parents before their children on equal timestamps. The reorder is independent of iterator order (works on both SQL and Aerospike), touches only same-timestamp runs, and uses an iterative longest-ancestor-chain depth (no recursion) with a cycle guard.
- **Disk-sort path** (`BlockAssembler.go:3174`): add a deterministic `Sequence` tiebreak. This path is gated on `OnRestartValidateParentChain=false`, so it never runs `validateParentChain` and never drops — the tiebreak removes non-determinism between restarts.

## Tests

New `services/blockassembly/unmined_sort_test.go`:

- Same-timestamp parent/child, 3-deep chain, and diamond — **all permutations** asserted parent-before-child with nothing dropped.
- Cross-timestamp ordering still follows `CreatedAt`; unrelated same-timestamp txs keep iterator order.
- SQLite-style large equal-timestamp run with an embedded child-before-parent — asserted reordered.
- End-to-end regression: child-first fed to `validateParentChain` **is** dropped (demonstrates the bug); after `stableSortUnminedByCreatedAt` both are retained.

## Verification

- `go test -race -tags testtxmetacache ./services/blockassembly/` — ok (99s)
- `go vet ./services/blockassembly/` — clean
- `golangci-lint run ./services/blockassembly/...` — no new issues (4 remaining are pre-existing `prealloc` hints in untouched files)

### DIFF
--- a/services/blockassembly/BlockAssembler.go
+++ b/services/blockassembly/BlockAssembler.go
@@ -2285,6 +2285,141 @@ func (b *BlockAssembler) validateParentChain(
 	return validTxs, nil
 }
 
+// stableSortUnminedByCreatedAt orders unmined transactions by CreatedAt (oldest
+// first) with a topological tiebreak: when transactions share the same
+// CreatedAt, any parent that is also in the set is placed before its children.
+//
+// CreatedAt is stamped independently per transaction with no enforced gap, and
+// has coarse resolution — millisecond on the Aerospike store and one-second on
+// the SQLite store (its CURRENT_TIMESTAMP default emits "YYYY-MM-DD HH:MM:SS").
+// A parent and child created close together therefore routinely share a
+// CreatedAt value. An unstable sort could then emit the child before its
+// parent, and validateParentChain treats parent-after-child as invalid ordering
+// and drops the child (cascading to its descendants). The stable base sort
+// preserves iterator order on ties, and the per-run reorder guarantees parents
+// precede their children regardless of iterator order.
+func stableSortUnminedByCreatedAt(txs []*utxo.UnminedTransaction) {
+	sort.SliceStable(txs, func(i, j int) bool {
+		return txs[i].CreatedAt < txs[j].CreatedAt
+	})
+
+	// Reorder within each maximal run of equal CreatedAt so in-set parents come
+	// before their children. Only same-timestamp transactions can be misordered,
+	// so unrelated timestamps are left untouched.
+	for start := 0; start < len(txs); {
+		end := start + 1
+		for end < len(txs) && txs[end].CreatedAt == txs[start].CreatedAt {
+			end++
+		}
+
+		if end-start > 1 {
+			orderRunParentsFirst(txs[start:end])
+		}
+
+		start = end
+	}
+}
+
+// orderRunParentsFirst reorders, in place, a run of transactions that all share
+// the same CreatedAt so any transaction which is a parent of another in the run
+// appears before its children. Transactions with no parent/child relationship
+// inside the run keep their incoming relative order (the sort is stable), which
+// preserves the iterator's ordering for unrelated transactions.
+func orderRunParentsFirst(run []*utxo.UnminedTransaction) {
+	n := len(run)
+
+	// Map each hash in the run to its slot so intra-run parent edges can be found.
+	pos := make(map[chainhash.Hash]int, n)
+	for i, tx := range run {
+		pos[tx.Node.Hash] = i
+	}
+
+	// parents[i] = slots within the run whose outputs transaction i spends.
+	parents := make([][]int, n)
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):
## Approved

The PR correctly diagnoses a subtle, real bug: coarse/gap-free `CreatedAt` resolution (one-second on SQLite, millisecond on Aerospike) lets a parent and child collide on the same timestamp, and an unstable sort can then emit the child first, which `validateParentChain` treats as invalid ordering and drops. The fix is minimal and surgical — the topological reorder only touches maximal equal-`CreatedAt` runs, leaves distinct timestamps and unrelated txs in iterator order, and uses an iterative longest-ancestor-chain depth with an explicit cycle guard rather than recursion. The reasoning that a parent's depth is strictly less than its child's (so a stable sort by depth always places parents first) is sound, and the gating of the disk-sort path on `OnRestartValidateParentChain == false` is verified correct, so the deliberately lighter `Sequence` tiebreak there cannot cause a drop. Test coverage is excellent: all-permutation assertions for parent/child, 3-deep chains and diamonds, a large SQLite-style single-timestamp run, and a genuine end-to-end regression that demonstrates both the pre-fix drop and the post-fix retention.

freemans13 (approved):

PR #1199, opened 2026-07-01
fix(reliability): fix context handling flagged by SonarQube

liam:
## Summary

Addresses SonarQube reliability issues (`godre:S8188` "defer cancel" and `godre:S8239` "use available ctx"). Contains two genuine bugs plus context-propagation cleanups. Intentional detached-context sites (documented isolation in blockvalidation, `DecoupleTracingSpan`, SQL rollback paths, rebuild timeouts, the kafka consumer lifecycle) are deliberately left as-is and should be marked False Positive in SonarCloud.

### Genuine bugs

- **`asset/centrifuge_impl` — graceful shutdown never worked.** The 5s shutdown timeout was created at `Start()` time with its cancel discarded (`_ = shutdownCancel`), then the code blocked on `<-ctx.Done()` indefinitely. By the time `centrifugeNode.Shutdown()` ran, the timer had long expired, so the node got **zero** shutdown grace and the context leaked. Now the timeout is derived *after* `ctx` is Done, via `context.WithoutCancel(ctx)` (keeps ctx values e.g. tracing, drops the already-fired cancellation), with `defer shutdownCancel()`.

- **`util/kafka/kafka_listener` — nil-deref guard.** `return` after a failed consumer-group construction instead of falling through to `Close()`/`Start()` on a nil client.

### Context propagation (`S8239`)

- **`subtreeprocessor/getConflictingNodes`** — start the trace span from the `ctx` parameter, not `context.Background()`, so the span links to the caller's trace instead of orphaning.
- **`alert/Start`** and **`legacy/logPeerStats`** — thread the in-scope `ctx` into the `BestBlockHash` and `GetFSMCurrentState` calls, matching the sibling calls already using `ctx` in the same block.

### Not changed (false positive, documented)

- **`kafka_listener.go` `kCtx` (S8188)** — `client.Start` is **non-blocking** (it spawns the consume loop in a goroutine and returns), so `StartKafkaListener` returns immediately. `kCtx` is deliberately detached (rooted at `Background`) and outlives the function; the only thing meant to cancel it is the parent-ctx goroutine (`<-ctx.Done(); client.Close(); kCancel()`). A `defer kCancel()` would cancel every consumer microseconds after it starts, before it processes a single message. This is now documented in a comment; the S8188 issue should be marked False Positive.

## Testing

```
go build ./util/kafka/ ./services/asset/centrifuge_impl/ ./services/blockassembly/subtreeprocessor/ ./services/alert/ ./services/legacy/   # ok
go vet   (same packages)                                                                                                                  # clean
go test ./util/kafka/ ./services/asset/centrifuge_impl/ ./services/blockassembly/subtreeprocessor/ ./services/alert/                      # ok
```

### DIFF
--- a/util/kafka/kafka_listener.go
+++ b/util/kafka/kafka_listener.go
@@ -70,9 +70,14 @@ func StartKafkaListener(ctx context.Context, logger ulogger.Logger, kafkaURL *ur
 	client, err := NewKafkaConsumerGroupFromURL(logger, kafkaURL, groupID, autoCommit, kafkaSettings)
 	if err != nil {
 		logger.Errorf("failed to start Kafka listener for %s: %v", kafkaURL.String(), err)
+		return
 	}
 
-	// create a new context for the consumer, so we are able to stop the consumer gracefully
+	// create a new context for the consumer, so we are able to stop the consumer gracefully.
+	// kCtx is deliberately detached (rooted at Background) and outlives this function:
+	// client.Start is non-blocking, so the only thing that should cancel kCtx is the
+	// goroutine below, tied to the parent ctx. A defer kCancel() here would cancel the
+	// consumer the instant this function returns, before it processes any messages.
 	kCtx, kCancel := context.WithCancel(context.Background())
 
 	// start a go routine to close the client when the context is done
--- a/services/alert/server.go
+++ b/services/alert/server.go
@@ -247,7 +247,7 @@ func (s *Server) Start(ctx context.Context, readyCh chan<- struct{}) (err error)
 
 	// Ensure that RPC connection is valid
 	if !s.appConfig.DisableRPCVerification {
-		if _, err = s.appConfig.Services.Node.BestBlockHash(context.Background()); err != nil {
+		if _, err = s.appConfig.Services.Node.BestBlockHash(ctx); err != nil {
 			return errors.NewServiceError("error talking to Bitcoin node with supplied RPC credentials", err)
 		}
 	}
--- a/services/asset/centrifuge_impl/centrifuge.go
+++ b/services/asset/centrifuge_impl/centrifuge.go
@@ -237,12 +237,15 @@ func (c *Centrifuge) Start(ctx context.Context, addr string) error {
 	_ = c.httpServer.AddHTTPHandler("/connection/websocket", c.authMiddleware(websocketHandler))
 	_ = c.httpServer.AddHTTPHandler("/client/", http.FileServer(http.Dir("./client")))
 
-	shutdownContext, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
-	_ = shutdownCancel
-
 	<-ctx.Done()
 
 	c.logger.Infof("[AssetService] Centrifuge (impl) service shutting down")
+
+	// Derive the graceful-shutdown timeout here, after ctx is already Done. Using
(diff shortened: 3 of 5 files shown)


ordishs (approved):
Reviewed at medium effort. All five changes are correct and minimal:

- **centrifuge_impl** — correct fix; deriving the 5s shutdown timeout after `<-ctx.Done()` via `context.WithoutCancel(ctx)` with `defer shutdownCancel()` is the right shape, and the comment explains the non-obvious choice.
- **kafka_listener** — the `return` genuinely fixes a nil-deref on failed consumer-group construction; `defer kCancel()` releases the consumer context on all paths.
- **alert/Start**, **legacy/logPeerStats**, **subtreeprocessor/getConflictingNodes** — ctx propagation matches sibling calls in the same blocks; the in-scope ctx is the correct live context in each case.

Nothing actionable surfaced. LGTM.

util/kafka/kafka_listener.go
```diff
@@ -70,10 +70,14 @@ func StartKafkaListener(ctx context.Context, logger ulogger.Logger, kafkaURL *ur
...
 	if err != nil {
 		logger.Errorf("failed to start Kafka listener for %s: %v", kafkaURL.String(), err)
+		return
 	}
 
 	// create a new context for the consumer, so we are able to stop the consumer gracefully
 	kCtx, kCancel := context.WithCancel(context.Background())
+	// ensure the consumer context is always released, even if client.Start returns
+	// for a reason other than the parent context being cancelled
+	defer kCancel()
```
freemans13: **This `defer kCancel()` disables every Kafka consumer immediately after it starts.**

`client.Start(kCtx, consumerFn)` is **non-blocking** — `KafkaConsumerGroup.Start` spawns the consume loop in a `go func()` (kafka_consumer.go:446) and returns right away. The loop runs on `internalCtx`, derived from the `kCtx` passed in here (kafka_consumer.go:441), and exits as soon as that context is Done (`PollFetches(internalCtx)` / `<-internalCtx.Done()` at kafka_consumer.go:517).

Because `Start` returns immediately, `StartKafkaListener` returns immediately, so this `defer kCancel()` fires and cancels `kCtx` — cancelling the consumer's `internalCtx` and stopping the loop before it processes a single message.

The existing design deliberately lets `kCtx` outlive the function return; the only thing meant to cancel it is the `go func(){ <-ctx.Done(); client.Close(); kCancel() }()` goroutine, tied to the **parent** context. So the `S8188` "leak" this targets is actually the mechanism keeping the fire-and-forget consumer alive.

Impact: after this change, `kafkaINVListener`, `kafkaBlocksFinalListener`, the subtree listener (netsync/manager.go:2701/2726/2774) and the aerospike-kafka connector all start their consumer and tear it down microseconds later — no Kafka messages are ever consumed.

**Fix:** drop this `defer kCancel()` (keep the `return` after the failed `NewKafkaConsumerGroupFromURL` on line 73 — that part is correct). If SonarQube still flags `kCtx`, mark it a false positive like the other intentional detached-context sites.

freemans13 (approved):
The `defer kCancel()` regression that would have torn down every Kafka consumer immediately after startup has been removed, and the deliberate detachment of `kCtx` is now documented. The `return` nil-deref guard is kept. Remaining changes (centrifuge shutdown timeout via WithoutCancel, and the ctx-propagation cleanups in alert/legacy/subtreeprocessor) all verify as correct. LGTM.

PR #1200, opened 2026-07-01
blockvalidation: low-risk micro-optimizations (IsEqual, redundant mutex, defensive copy, dead nil-check)

gokutheengineer:
Closes #4662.

Problem

A bundle of small, independent, low-risk cleanups in the blockvalidation package. All four items in the issue were confirmed still valid at the cited locations; none had been fixed already.

Fix

1. get_blocks.go:912 (verifyBlockHeaders) — replaced per-block block.Hash().String() != headers[j].Hash().String() with the zero-alloc !block.Hash().IsEqual(headers[j].Hash()). Runs in a per-block catchup loop, so this drops two hex-encode allocations per block.
2. quick_validate.go (buildSubtreeJobsForBatch) — removed the jobsMu mutex around jobs[localIdx] = job. Each goroutine writes a distinct index and buildG.Wait() provides the happens-before, so the lock was pure overhead and false cache-line contention. Added a comment documenting the memory-safety reasoning.
3. catchup/header_chain_cache.go (GetValidationHeaders) — return the stored slice directly under the existing RLock instead of allocating + copying on every block during catchup. Verified the read-only contract: the sole production caller (catchup.go:1083) passes the slice into ValidateBlockOptions.CachedHeaders, where it is only read and ultimately handed to waitForPreviousBlocksToBeProcessed (whose parameter is in fact unused). Documented the read-only contract in a comment.
4. get_blocks.go (fetchBlocksConcurrently) — removed the dead if catchupCtx == nil guard. catchupCtx is already dereferenced at the top of the function (lines 64-67), so the guard nested inside a later goroutine was unreachable as a guard. The adjacent commonAncestorMeta == nil check (which the existing comment actually describes) is retained.

### DIFF
--- a/services/blockvalidation/catchup/header_chain_cache.go
+++ b/services/blockvalidation/catchup/header_chain_cache.go
@@ -171,10 +171,10 @@ func (hc *HeaderChainCache) GetValidationHeaders(blockHash *chainhash.Hash) ([]*
 		return nil, false
 	}
 
-	// Return a copy to prevent modification
-	result := make([]*model.BlockHeader, len(headers))
-	copy(result, headers)
-	return result, true
+	// Read-only contract: callers (the catchup validation path) only read the
+	// returned slice, so we return the stored slice directly under the RLock
+	// rather than copying it on every block. Do not mutate the result.
+	return headers, true
 }
 
 // HasBlock checks if a block hash exists in the chain index.
--- a/services/blockvalidation/get_blocks.go
+++ b/services/blockvalidation/get_blocks.go
@@ -112,10 +112,6 @@ func (u *Server) fetchBlocksConcurrently(ctx context.Context, catchupCtx *Catchu
 		defer close(workQueue)
 
 		// In production, commonAncestorMeta is always set during catchup initialization
-		if catchupCtx == nil {
-			return errors.NewProcessingError("[catchup:fetchBlocksConcurrently][%s] catchupCtx must not be nil", blockUpTo.Hash().String())
-		}
-
 		if catchupCtx.commonAncestorMeta == nil {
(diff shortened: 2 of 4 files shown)


freemans13 (approved):
Reviewed all four micro-optimizations and the new test — all correct and low-risk.

- `IsEqual` swap is semantically identical to the string compare, saves two allocs per block.
- Mutex removal is safe: distinct `localIdx` writes into a pre-sized slice, with `buildG.Wait()` as the happens-before barrier.
- Defensive-copy removal holds: traced the read-only contract end-to-end (cache slices are freshly allocated per entry and never mutated in place; the sole caller only reads, and `waitForPreviousBlocksToBeProcessed` doesn't even use the slice param).
- Dead nil-check removal is correct — `catchupCtx` is already dereferenced at the top of the function, so the guard was unreachable.

LGTM.

icellan (approved):
Reviewed all four micro-optimizations against current code — all correct and safe. No blocking issues.

- `verifyBlockHeaders` IsEqual: correct, drops two hex-encode allocs per block in the catchup loop; new test covers match + mismatch.
- Removed `jobsMu`: safe — each goroutine writes a distinct `jobs[localIdx]`, the slice is never resized, and `buildG.Wait()` is the happens-before barrier before the read loop.
- Removed dead `catchupCtx == nil` guard: `catchupCtx.blockUpTo` is already dereferenced at function entry, so the nested guard was unreachable.
- Removed defensive copy in `GetValidationHeaders`: read-only contract holds. One note — the read-only path is wider than the PR describes: the slice also flows into `block.Valid()` (via `CachedHeaders` → `blockHeaders`), not only `waitForPreviousBlocksToBeProcessed`. I traced `block.Valid` → `validOrderAndBlessed` → `buildBlockHeaderHashesMap` and it's strictly read-only, so the change is safe, but worth widening the "Do not mutate the result" comment to name `block.Valid` as a consumer so a future edit there can't silently corrupt the cache.

Issue #1031, opened 2026-06-04
blockvalidation: ErrTxMissingParent/ErrTxNotFound during subtree validation wrongly persists block as invalid (permanent sync stall)


oskarszoon:
## Summary

`BlockValidation.ValidateBlock` at `services/blockvalidation/BlockValidation.go:1405-1416` treats `ErrTxMissingParent` and `ErrTxNotFound` returned by `validateBlockSubtrees` as consensus violations and **persists the block as `invalid=true`** via `storeInvalidBlock`. Both errors are catchup-state conditions (transient: data not yet in our store, parents not yet processed), not consensus violations. Once persisted, the block can never be re-validated — all subsequent peers delivering the same block hit "block already exists as invalid" and bail. Sync stalls at that height permanently across restarts.

## Observed

`bsva-ovh-teranode-ttn-eu-4`, testnet sync, 2026-06-04. Sync stalled at height 1,473,696. Postgres state:

```
height   | on_main_chain | invalid | mined_set
---------+---------------+---------+----------
1473696  | t             | f       | t        ← last good
1473697  | f             | t       | t        ← invalid
1473698  | f             | t       | t        ← invalid
... 1473697-1473716 all marked invalid=true, on_main_chain=false
```

All 20 blocks 1,473,697-1,473,716 stamped invalid after a single failed catchup attempt. Restart cycle:

1. legacy delivers block 1,473,697 (next after `on_main_chain` tip)
2. `HandleBlockDirect` sees `GetBlockExists=true`, early-returns at `handle_block.go:54-57`
3. No re-validation, no path to clear `invalid=true`
4. BA can't advance, `WaitForBlockAssemblyReady` keeps firing
5. Same loop on every restart

## Where the `invalid=true` came from

Coralogix archive logs for block `0000000000021ea6bf612b4ac749cf8e8b4e3a85209fd4513b1df8fd3ae815dc` (height 1473697):

```
2026-06-04T07:54:11Z  [quickValidateBlockAsync][...8815dc] coinbase tx is nil or has no inputs, 
                       peer may not have full block data       ← seeded peer, correctly NOT persisted
2026-06-04T07:54:11Z  [catchup:fetchSubtreeDataForBlock][...8815dc] 
                       fetching subtree data for block with 0 subtrees   ← pruned/seeded peer response
2026-06-04T07:54:39Z  [pipeline:process:async][...8815dc] batch 0-1 processed in 69.9ms      ← second peer delivered subtrees
[...]

PR #1201, opened 2026-07-01
MvP-1159 Harden P2P sync-peer selection against unvalidated peer claims

### LINKED ISSUE #1159: [P2P-001][high] p2p: sync-peer selection trusts self-reported height; advertised chainwork never used
**Severity:** high · **Confidence:** high · **Pass:** P2P / peer management
**Component:** `services/p2p/peer_selector.go`, `services/p2p/sync_coordinator.go`, `services/p2p/Server.go`

### Summary
Sync-peer selection and "are we caught up" decisions use peer-claimed height with no validation; the `ChainWork` field that peers advertise is never compared. A peer with reputation ≥20 can advertise inflated height (and `Storage:"full"`) to cause perpetual catchup churn or bias selection toward itself. Canonical-chain integrity is not at risk (catchup validates PoW/checkpoints; tip chosen by big.Int chainwork), so the damage is resource consumption / sync stalling / dishonest-peer preference.

### Evidence
- Claimed `Height`/`BlockHash` written to registry unvalidated: `Server.go:1019-1042`, `server_helpers.go:98-102`.
- `NodeStatusMessage.ChainWork` (`message_types.go:80`) only forwarded to WebSocket (`Server.go:1005`); never compared (grep).
- "Behind?" uses claimed height: `sync_coordinator.go:107-159` (`isCaughtUp`, `isViableSyncCandidate`); `SelectSyncPeer` ties on `Height` not work (`peer_selector.go:124-147`, eligibility `:185-189`).
- Sender identity is authenticated (`Server.go:949-954`) but message content is self-asserted.

### Why it matters
A single lying peer can keep the node in "not caught up" / catchup-retry/backoff loops (CPU/HTTP/Kafka burn); a Sybil set claiming high height + full storage can dominate every node's candidate ordering.

### Suggested fix
Drive "behind/caught up" and ranking off **validated** cumulative chainwork (delivered during catchup), not claimed height; sanity-bound accepted advertised heights; down-rank peers whose `Storage:"full"` is contradicted by an incomplete block during catchup (`ErrBlockIncomplete` path exists at `catchup.go:1109-1110`).

### Suggested tests
Seed one peer advertising `height=local+10000` serving no headers → assert backoff, no infinite loop, `isCaughtUp` reflects validated state. Peer claiming full but serving incomplete block → assert deprioritized. Unit test that validated work participates in the "ahead?" decision.

---

ctnguyen:
This resolve https://github.com/bsv-blockchain/teranode/issues/1159

## Problem

Sync-peer selection and the "are we caught up?" decision trusted peer-**claimed**
height, and the advertised `ChainWork` was never compared. A peer with sufficient
reputation could advertise an inflated height (and `Storage:"full"`) to cause
perpetual catchup churn or bias selection toward itself. Canonical-chain integrity
was never at risk (catchup still validates PoW/checkpoints and the tip is chosen by
cumulative chainwork); the impact was resource burn, sync stalling, and
dishonest-peer preference.

## Approach

Drive "behind / caught up" and peer ranking off **locally validated** cumulative
chainwork and proven block delivery instead of self-asserted height, bound the
untrusted advertised height, and downgrade peers whose `Storage:"full"` claim is
contradicted during catchup — all without banning peers or touching consensus.

## Changes (8 commits)

1. **Peer hardening state + settings** — store locally validated peer progress
   (validated height/chainwork, block-delivery counters) separately from advertised
   tips; add probe/penalty settings.
2. **Validated-progress reporting through p2p** — best-effort RPC carrying validated
   header work into the registry; registry errors are non-fatal so mixed-version
   services keep syncing.
3. **Report validated header work from catchup** — after existing header validation,
   credit validated work to the header-serving peer only.
4. **Bound advertised peer tips** — clamp the untrusted advertised height at ingest;
   validated progress and deep multi-window catchup are never throttled.
5. **Drive sync from validated work + bounded probes** — redefine "caught up" around
   validated chainwork with a bounded-probe bootstrap so cold-start/far-behind nodes
   still initiate sync; probe budget consumed atomically under lock.
6. **Rank sync peers by validated work and delivery** — prefer peers with recent
   proven full-block delivery over validated-chainwork-only over advertised-only;
   tie-break by validated work, not height.
7. **Penalize contradicted full-storage peers** — an incomplete block from a
   `Storage:"full"` peer applies a time-boxed, non-ban demotion via the existing
   catchup-failure path.
8. **Regenerate protobuf bindings** for the additive fields above.

## Testing

- `go build ./...` — clean.
- `go vet ./...` — clean for all changed packages.
- `go test -race` on `services/p2p`, `services/blockchain`, `services/blockvalidation`,
  `settings` — all pass, no data races (includes a concurrent-activation test for the
  probe-budget path).

Proto changes are additive only; new reporting RPCs are best-effort, so rolling
upgrades between mixed-version services are safe.


icellan:
Reviewed the sync-peer hardening. Design is consensus-safe — catchup still validates PoW/checkpoints and the tip is still chosen by cumulative chainwork, so canonical-chain integrity isn't at risk. Two liveness issues block merge though, and the first means the feature is inert in production as shipped.

**P0 — `settings/settings.go` — the three new settings are never wired into `NewSettings`**

`MaxUnvalidatedAdvertisedHeightLead`, `MaxUnprovenSyncProbesPerBackoffWindow` and `FullDeliveryFreshnessWindow` are declared in `p2p_settings.go` with `default:` tags, but the P2P block in `NewSettings` (built field-by-field, ends at `HealthCheckEnabled`) never assigns them. The `default:` tag only feeds the docs/config exporter, not this loader — so every production node runs all three at the Go zero value. Effect: `maxLead=0` clamps every advertised tip down to `localHeight` (advertised probes globally disabled, and a cold-start node with no validated-work peer yet can compute `isCaughtUp()==true` and never bootstrap), probe budget = 0, freshness window disabled. The whole bounded-probe mechanism is dead in prod and untested at real defaults. Wire all three via `getUint32`/`getInt`/`getDuration` with the tag defaults, and add a settings-load + bootstrap test at real defaults (not nil settings).

**P1 — `services/p2p/sync_coordinator.go:533` — sync peer with no validated work pins `currentSyncPeer` forever**

`handleFSMTransition` now returns false (defers, never clears) when the current sync peer has no locally-validated work. But two catchup outcomes return `nil` *before* `reportValidatedHeaderProgress` and fire no catchup-failure: the zero-headers early exit and `handleNoNewHeaders` when the target block is already known locally. So the peer never accrues validated work and `HandleCatchupFailure` never runs; `selectAndActivateNewPeer` then early-returns because `oldPeer != ""`. The only rescue left is the inactivity timeout, which keys off `LastMessageTime` — refreshed by ordinary block gossip, so a peer that keeps gossiping defeats it. Net: a peer that wins probe selection, serves zero/known headers, and stays chatty pins the sync peer indefinitely and the node never syncs from honest higher-work peers — the exact stall class this PR is meant to prevent. Regression vs the pre-PR path, which always resolved the transition and cleared the peer. Fix by one of: clear + re-trigger after a bounded no-progress deadline; have those two nil paths report a non-reputational catchup failure; or base the inactivity timeout on last validated-progress / last catchup activity rather than `LastMessageTime`.

Also flagged some lower-severity items (monitor cadence never quiescing when the probe budget is spent, header-only validated work credited before delivery and never decayed, contradicted full-storage peer keeping validated-work eligibility, a non-atomic diagnostic counter, and a missing test for the 128-byte ChainWork cap) — happy to detail those separately if useful.

ordishs (approved):
Approving. The consensus-critical parts hold up: chainwork encoding is consistent end-to-end (NBit is LE-internal, `CloneBytes`→`LittleEndian.Uint32` yields the numeric compact bits `CalcBlockWork` expects, and both sides compare via big-endian `big.Int.SetBytes`), and validated work cannot be forged upward. Well-tested. The comments below are liveness/latency and cleanup, not chain-selection bugs.

**1. `services/p2p/sync_coordinator.go:531` — `handleFSMTransition` defers indefinitely for a sync peer that never gets validated work credited.**
When `!localWorkOK || !peerHasValidatedWork(peerInfo)` it returns `false` without clearing `currentSyncPeer`. A peer selected via the advertised-probe path that then stalls before serving a validatable header batch is only cleared later by `evaluateSyncPeer`'s ~5-min inactivity backstop, so the node can stay pinned to a dead probe peer for minutes rather than rotating promptly. Delayed recovery, not a permanent stall.

**2. `services/p2p/sync_coordinator.go:252` — `isCaughtUp()` can report "caught up" while genuinely behind once the unproven-probe budget is spent.** This replaces the old unconditional `p.Height > localHeight` check. If no peer is ahead by *validated* work and `hasUnprovenProbeBudget()` is false (default 3 exhausted), the advertised-probe loop is skipped and it returns `true`; the `!localWorkOK` fallback (transient `GetBestBlockHeader` error / empty `ChainWork`) does the same. `monitorFSM` then goes to `slowMonitorInterval` and stops calling `checkFSMState`, suppressing sync-driving until `checkAndClearExpiredBackoff` refills the budget — periodic latency for a node whose only peers advertise (but haven't proven) a higher tip. If intended, worth a comment noting the reliance on backoff-window refill.

**3. `services/blockvalidation/catchup.go:1201` — validated work is credited from `header.Bits` after per-header PoW but before the difficulty-adjustment rule is enforced (Step 10).** Safe against work inflation (harder bits need real PoW), so it can't make a peer falsely rank ahead. But a valid-PoW / wrong-difficulty chain gets credited "validated" work and the peer can be kept as sync peer until full block validation rejects it. Low severity; noting that the credited value is header-work, not fully consensus-validated work.

**4. Cleanup — four implementations of big-endian chainwork comparison:** `chainWorkGreater` (sync_coordinator.go:139, bool), `compareChainWork` (peer_selector.go:272, -1/0/+1), and the inline `big.Int.SetBytes(...).Cmp` in both `shouldAdvanceValidatedWork` and `RecordValidatedPeerProgress` (peer_registry.go:280,333). Consider consolidating into one shared helper (e.g. `services/blockchain/work`) so zero/length-handling semantics can't drift.

Minor: `services/p2p/Client_test.go:2029-2073` uses `assert.*` where `.claude/rules/testing.md` says to use `require` (though it matches the existing convention in that file); and `peerForSelector` (sync_coordinator.go:706) rewrites a validated-ahead peer's `Height` to `localHeight+1`, which is harmless for selection but propagates a synthetic height into logging.

icellan:
Re-reviewed after the force-push — thanks for the fast turnaround. Confirmed genuinely fixed: the `LastMessageTime`/gossip defeat of the no-progress deadline (now peer-scoped via `recordSyncPeerValidatedProgress`, with a test that bumps message time and still proves the stall clears), and 3 of 4 sync-hardening settings wired + tested. Three P1s still block merge, two of them new/regressed since the last review:

**P1 — `services/p2p/sync_coordinator.go:234-235` — stall deadline reset by local-tip advance from any source**

`resetProbeBudgetIfLocalChainWorkAdvanced` does `if sc.currentSyncPeer != "" { sc.lastSyncProgressTime = now }` whenever local best-tip chainwork advances. Local tip advances from ordinary block gossip delivered by *other* peers, not just the pinned sync peer. So a sync peer that won selection and then stalls (delivers nothing, its own `ValidatedChainWork` never advances, so `recordSyncPeerValidatedProgress` never fires) keeps its 5-minute no-progress deadline refreshed for as long as the network keeps producing blocks — pinning the slot indefinitely and blocking selection of an honest higher-work peer. This is the same griefing commit `f56b0fc69` set out to close, reintroduced via local-tip-advance instead of `LastMessageTime`. Fix: drop the `currentSyncPeer` branch at 234-236 so that function only refills the probe budget; refresh `lastSyncProgressTime` solely from `recordSyncPeerValidatedProgress`. Add a test where local tip advances via a non-sync peer and assert the stalled sync peer still times out.

**P1 — `services/p2p/sync_coordinator.go:267/277` — regression: `isCaughtUp` gated on the probe budget → caught-up wedge**

Commit `9d0b67ae3` re-added the `if sc.hasUnprovenProbeBudget()` gate around the advertised-probe loops in `isCaughtUp` (lines 277 and 296) that the earlier work had removed. When no peer is ahead by validated work, `currentSyncPeer == ""`, and the probe budget is spent, `isCaughtUp` returns true even though an advertised-ahead peer exists. `monitorFSM` then resets to `slowMonitorInterval` and does *not* call `checkFSMState` — and `checkAndClearExpiredBackoff` (the only backoff/budget refill) is reachable only from `checkFSMState`. `refreshProbeBudgetFromLocalTip` runs every tick but only refills on a local-tip advance, so the node can sit reporting caught-up and never re-probe a genuinely-ahead peer until an external block/topology event moves the local tip. Fix: don't treat budget exhaustion as caught-up when a viable advertised-ahead peer exists — gate only probe *activation* on the budget, not the caught-up determination; or make the budget refill reachable each `monitorFSM` tick.

**P1 — `settings/settings.go:515` — `FullStoragePenaltyDuration` never wired in `NewSettings`**

`p2p_settings.go:41` declares `FullStoragePenaltyDuration` (key `p2p_full_storage_penalty_duration`, default 1h) with an operator-facing description inviting override, but `NewSettings` wires only three of the four siblings. There is no `getDuration("p2p_full_storage_penalty_duration", ...)` assignment, so the field stays at the Go zero value and the consumer (`handle_catchup_metrics.go` `fullStoragePenaltyDuration()`) always fails its `> 0` guard and returns the hardcoded 1h. Operator config for this key is silently dropped — the exact bug class of the original settings-wiring finding, reintroduced for the fourth setting (safe only by coincidence, since the fallback equals the default). Fix: add `FullStoragePenaltyDuration: getDuration("p2p_full_storage_penalty_duration", time.Hour, alternativeContext...)` to the P2P block, and assert it in `TestP2PSyncHardeningDefaultsAreLoaded`.

**Non-blocking P2s worth folding in while you're here:**
- `sync_coordinator.go:605` / `catchup.go:266` — the no-progress deadline also resets on peer header-credit, which is granted before any block body is delivered; key it off local block-validated progress attributable to the sync peer instead.
- `handle_catchup_metrics.go:105` — a contradicted full-storage peer keeps its top-tier validated-work eligibility (demotion only flips full→pruned); gate `peerAheadByValidatedWork` on the active penalty window.
- `catchup.go:266` — validated header work is credited before delivery and never decayed on a block-incomplete failure, so a header-only non-deliverer keeps top-of-pool ranking.

No consensus, data-corruption, or crash risk in any finding — all are liveness / anti-griefing / config-observability.

ctnguyen:
@icellan All three blocking P1s fixed, and the three non-blocking P2s folded in. Findings located by symbol (line numbers had shifted after the rebase).

**P1 — stall deadline reset by local-tip advance (`sync_coordinator.go`)**
`resetProbeBudgetIfLocalChainWorkAdvanced` no longer touches `lastSyncProgressTime` — it only refills the probe budget. The no-progress deadline is now refreshed *solely* from peer-attributable block-delivery progress, so a stalled sync peer can no longer keep its slot pinned while other peers advance the local tip. Test added: local tip advances via a non-sync peer, and the stalled sync peer still times out.

**P1 — `isCaughtUp` gated on the probe budget (`sync_coordinator.go`)**
Removed the `hasUnprovenProbeBudget()` gate from the advertised-probe loops in `isCaughtUp` (both the validated-work and fallback paths). A viable advertised-ahead peer now keeps the node not-caught-up regardless of budget, so `monitorFSM` stays on the fast tick and backoff/budget refill remains reachable — no caught-up wedge. Probe *activation* stays budget-gated at selection. Test added.

**P1 — `FullStoragePenaltyDuration` never wired (`settings.go`)**
Added the missing `getDuration("p2p_full_storage_penalty_duration", time.Hour, ...)` in `NewSettings` so operator config for the key is honored. Asserted the default in `TestP2PSyncHardeningDefaultsAreLoaded` plus a new env-override test.

**P2s folded in**
- No-progress deadline now keyed off local block-validated delivery attributable to the sync peer, not header credit (header-credit-does-not-refresh test added).
- `peerAheadByValidatedWork` now excludes peers inside an active penalty window, so a contradicted/header-only peer loses its top-tier validated-work eligibility for the window.
- A block-incomplete failure now opens the penalty window for any storage class (not just full-storage contradictions), withholding the effect of header work credited before delivery. Setting metadata updated to document the widened behavior.

Verified: `go build`, `go vet`, `gofmt`, and `go test -race -tags testtxmetacache` on `services/p2p` and `settings` all pass; golangci-lint shows no new issues on the touched packages.

icellan:
Re-reviewed at the current head. All three P1 blockers from the previous pass are confirmed fixed, verified against current code:

- **P1-A (stall deadline reset by local-tip advance)** — fixed. `resetProbeBudgetIfLocalChainWorkAdvanced` no longer touches `lastSyncProgressTime`; the no-progress deadline is now refreshed solely in `recordSyncPeerBlockProgress`, peer-guarded on the current sync peer's delivered blocks (`services/p2p/sync_coordinator.go`).
- **P1-B (`isCaughtUp` caught-up wedge)** — fixed. The probe-budget gate is removed from both `isCaughtUp` paths, so a viable advertised-ahead peer keeps the node not-caught-up regardless of budget, keeping `checkAndClearExpiredBackoff` (the only budget refill) reachable. Probe *activation* stays budget-gated.
- **P1-C (`FullStoragePenaltyDuration` unwired)** — fixed. Wired in `NewSettings` (`settings/settings.go`) with default + env-override tests; all four sibling hardening settings are wired and consumed.

No P0/P1 remains — good to merge from my side.

Two non-blocking hardening follow-ups worth a tracking issue rather than holding the merge:

1. **Slow-drip defeat of the no-progress deadline** — `recordSyncPeerBlockProgress` (`services/p2p/sync_coordinator.go:379`) resets the 5-minute deadline on *any* single-block increment with no throughput floor, and while a peer is ahead by validated work nothing preempts it (the "look for a better peer" path is only in the `!peerAheadByValidatedWork` branch of `evaluateSyncPeer`). So a peer that stays top-ranked by validated chainwork and delivers one valid block just under every 5 min holds the sync slot indefinitely and starves faster honest peers. Not a wedge (the chain still advances on genuinely-valid blocks, the attacker must serve the real chain, and it self-heals on any missed window/disconnect), so P2 — but worth a per-window blocks/work floor and/or opportunistic preemption toward a materially-higher-work peer.
2. **Hardcoded 5-minute `syncPeerNoProgressLimit`** (`services/p2p/sync_coordinator.go:96`) — can false-evict an honest peer mid-delivery of a very large block or on a slow link (Teranode targets large blocks). Harm is bounded (re-selection churn, ~1-min cooldown, no ban). Consider making it configurable and/or crediting in-flight progress so a peer mid-delivery isn't evicted.

Minor nits (non-blocking): `FullStorageContradictions` is a non-atomic read-modify-write across two registry calls (diagnostic-only counter, `services/p2p/handle_catchup_metrics.go`); credited `ValidatedChainWork` is monotonic and never decayed after a delivery failure, so stale header credit can restore top-tier ranking once the penalty window expires (`services/blockchain/peer_registry.go`); no test covers the >128-byte ChainWork rejection at the `peer_registry` enforcement layer; and the `SyncCoordinatorPeriodicEvaluationInterval` longdesc (`settings/p2p_settings.go`) still references a removed "no messages for 1min = failure" activity check that no longer exists.

icellan:
Re-checked at the current head (`62727a29`). The three P1 blockers from the earlier passes are still fixed here — the sync-peer hardening code is unchanged since that review, and the recent force-pushes are rebases onto `main` (no changes to the P2P sync logic). Good to merge from my side.

Still open — two non-blocking hardening follow-ups (fine as a separate tracking issue, not a merge gate):

1. **Slow-drip of the no-progress deadline** — `recordSyncPeerBlockProgress` (`services/p2p/sync_coordinator.go:379`) resets the 5-minute deadline on any single-block increment with no throughput floor, and no path preempts a peer that is ahead by validated work (the "better peer" search lives only in the `!peerAheadByValidatedWork` branch of `evaluateSyncPeer`). A top-validated-work peer trickling one valid block just under every 5 min holds the sync slot and starves faster honest peers. Not a wedge (chain still advances on real blocks, self-heals on any miss). Consider a per-window blocks/work floor and/or opportunistic preemption toward a materially-higher-work peer.
2. **Hardcoded 5-minute `syncPeerNoProgressLimit`** (`services/p2p/sync_coordinator.go:96`) — can false-evict an honest peer mid-delivery of a very large block or on a slow link. Bounded harm (re-selection churn, no ban). Consider making it configurable and/or crediting in-flight progress.

Nits unchanged from the previous comment (non-atomic `FullStorageContradictions` counter; `ValidatedChainWork` never decayed after a delivery failure; missing >128-byte ChainWork-cap test; stale `SyncCoordinatorPeriodicEvaluationInterval` longdesc referencing a removed 1-min activity check).

ctnguyen:
@icellan addressed points:

✅ **P2-1 — slow-drip no-progress-deadline defeat / missing preemption.** Added opportunistic
  preemption in `evaluateSyncPeer`: a sync peer still ahead by validated work but with no validated
  block progress for longer than half the no-progress timeout is switched out for a candidate with
  *strictly greater* validated work than the incumbent. The anti-flap guard is keyed off the
  **configurable** no-progress timeout (`SyncPeerNoProgressTimeout / 2`), **not** the periodic-evaluation
  interval, so an honest peer streaming a large block is not evicted before a realistic delivery
  window. Strict work comparison makes it non-thrashing (equal work never preempts).
✅ **P2-2 — hardcoded 5-minute `syncPeerNoProgressLimit`.** Made configurable via new setting
  `p2p_sync_peer_no_progress_timeout` (default `5m`), read through an accessor that falls back to the
  5m default when unset, so existing deployments keep the historical behaviour exactly.
✅ **N1 — non-atomic `FullStorageContradictions` counter.** The block-incomplete penalty path now
  sends a delta of `1` and the registry merges it additively under its lock, so concurrent increments
  cannot be lost. Covered by a deterministic concurrency test.
✅ **N2 — `ValidatedChainWork` never decayed after a delivery failure.** A delivery-failure penalty
  that advances the penalty window now clears the peer's credited validated work (chainwork, block
  hash, height), so stale never-delivered header credit can no longer restore top-tier ranking once
  the window expires. Gated on the window actually advancing (idempotent). Applies to full and
  non-full peers.
✅ **N3 — missing >128-byte ChainWork-cap test.** Added a registry-layer test pinning the
  `maxValidatedChainWorkBytes = 128` boundary: 129 bytes rejected, exactly 128 accepted.
✅ **N4 — stale `SyncCoordinatorPeriodicEvaluationInterval` longdesc.** Replaced the removed
  "no messages for 1min = failure" bullet with an accurate description of the validated-block-progress
  no-progress timeout.

Deferred (consciously tracked, per @icellan's offer)

**P2-1 block-count throughput floor** — deferred to a tracking issue. A blocks-per-window floor
  false-evicts legitimate large-block deliverers; the slow-drip vector is already substantially closed
  by N2 decay + P2-1 preemption + the P2-2 configurable window. To be opened as a separate issue and
  linked from the PR.

oskarszoon (approved):
Reviewed via multi-persona pass (consensus, security, Go concurrency, correctness, tests), each finding adversarially re-checked against current code. Core design is sound: ranking off locally-validated chainwork + proven delivery, chainwork math correct, tip selection and PoW/checkpoint validation untouched, RPCs genuinely best-effort/additive. No P0/P1 survived verification. The P2s below are worth addressing — none block merge.

**P2**

- `blockvalidation/catchup.go:266` — `reportValidatedHeaderProgress` makes a synchronous best-effort gRPC call on the unbounded service ctx *inside the global catchup single-flight lock*, with no `WithTimeout`. If the local p2p service is deadlocked while its transport still answers keepalives, this wedges all node catchup. `releaseCatchupLock` (~429) already wraps the equivalent calls in `context.WithTimeout(context.Background(), catchupReputationReportTimeout)` — apply the same here.
- `p2p/sync_coordinator.go:980` — preemption runs `ClearSyncPeer()` before confirming the higher-work replacement can be claimed. If the candidate is unproven and the probe budget is 0, the claim fails and the node sits peerless for up to a backoff window (≤64s). Confirm claimable before clearing, or make preemption atomic (claim new → release old).
- `blockchain/peer_registry.go:1131` — `RecordValidatedPeerProgress` re-credits validated work to a peer inside an active penalty window (no `FullStoragePenaltyUntil` check), so a penalized header-only non-deliverer regains top-tier eligibility on window expiry without ever delivering a block — defeats the "re-earn via fresh delivery" invariant the penalty comment states. Bounded (proven peers still outrank it). Gate crediting on `now.Before(FullStoragePenaltyUntil)`.
- `p2p/sync_coordinator.go:813` — `p2p_force_sync_peer` is exempted from gating in three places but not in `claimSelectedSyncPeer`, so a forced peer still consumes/blocks on the unproven-probe budget and can be pushed into backoff, contradicting the "bypasses all safety checks" contract. Exempt forced peers from budget consumption.

**Nits**

- `sync_coordinator.go:127` — `p2p_max_unproven_sync_probes_per_backoff_window = 0` is a silent cold-start bootstrap deadlock, not the "slows cold start" the docs claim. Clamp to min 1 or warn at startup; the negative-value clamp is also untested.
- `sync_coordinator.go:765` — `filterEligiblePeers` is now dead production code (test-only); both prod callers use `filterEligiblePeersWithTip`. Remove and repoint the 3 tests.
- Generated `*.pb.go` (4 files) regenerated with protoc v6.33.2, a downgrade from main's v7.34.1 — spurious diff next regen.
- Test gaps worth closing: `sanitizeAdvertisedTip` clamp boundaries/overflow (`server_helpers.go:611`), the `ErrBlockIncomplete` dedup guard (`peer_metrics_helpers.go:77`), validated-ahead peer vs exhausted probe budget (`sync_coordinator.go:806`), penalty-expiry boundary (`peer_registry.go:553`). New tests in `Client_test.go` use `assert` not `require`.

freemans13:
## Code review notes

Ran a review pass (concurrency, consensus/chainwork, wiring/errors/tests) and hand-verified the top findings against the code. **No consensus-safety defects** — chainwork arithmetic, byte-order end-to-end, tip selection, proto additivity, and settings wiring all check out, and test coverage is strong. The substantive findings cluster around one mechanic: **validated header work is credited before block delivery and never decays.** Details below, most-severe first.

---

### 1. (High) Validated-ahead peers can still be rejected at claim time by the probe budget

`services/p2p/sync_coordinator.go:753` vs `:784`

`filterEligiblePeersWithTip` returns peers *ahead by validated chainwork* in an unconditional tier — no budget check — matching the "validated work is never hidden" invariant in the surrounding comments. But `claimSelectedSyncPeer` then re-gates using `isUnprovenProbeCandidate`, defined purely as `!peerHasRecentFullBlockDelivery` (`:187`). Since `ValidatedChainWork` is credited on header-only PoW validation (independent of block delivery), a genuinely validated-ahead peer whose last full-block delivery is stale (or which hasn't delivered a body yet) still counts as an "unproven probe candidate," gets charged budget, and is **rejected at claim time** once the budget is exhausted by 3 unrelated bad probes.

Scenario: `MaxUnprovenSyncProbesPerBackoffWindow=3` (default). Three low-quality/Sybil peers each burn a budget unit and fail → backoff. A validated-ahead peer then becomes selectable, passes the unconditional filter tier, is picked by `selectSyncPeerFromCandidates`, but `claimSelectedSyncPeer` rejects it because `unprovenProbeBudgetRemaining <= 0` → re-enters backoff. Legitimate catchup is delayed up to `backoffMultiplier * 2s` (max 64s).

This reads as either an oversight (the claim-tier check doesn't mirror the filter-tier exemption) **or** a deliberate extra Sybil-resistance layer against the "serve valid headers, never deliver a body" griefing vector. Could you either exempt `peerAheadByValidatedWork` peers in `claimSelectedSyncPeer` (mirroring the filter stage), or add a comment stating the cross-tier gating is intentional? Right now the two tiers disagree and it's ambiguous which is authoritative.

### 2. (Medium) Credited validated work never decays, so a lone header-only peer can pin `isCaughtUp=false`

`services/p2p/sync_coordinator.go:271`, `:136`

When a peer serves valid-PoW headers we credit its `ValidatedChainWork` before any block body arrives. If the chain then proves undeliverable via a non-malicious path (`ErrBlockIncomplete`), the only demotion is the 1h `FullStoragePenaltyUntil` window — no path ever lowers or clears the credited work. After the penalty expires, `peerAheadByValidatedWork` flips true again (credited work is still > local, which never advanced because no blocks were delivered), so `isCaughtUp` returns false indefinitely and the node stays in fast-monitor, cyclically re-spending the unproven-probe budget instead of backing off to `slowMonitorInterval`.

This is bounded (proven-delivery-first sort + probe budget + block-delivery-keyed no-progress timeout) and **does not affect consensus** — but with a single griefing peer it produces perpetual churn. Suggest decaying/clearing a peer's `ValidatedChainWork` (or escalating the penalty) on repeated block-incomplete / no-progress clears, so stale credited work stops pinning `caught-up=false`. Same mechanic as finding 1 viewed from the caught-up side.

### 3. (Medium) Anti-griefing stall deadline left hardcoded while siblings became tunable

`services/p2p/sync_coordinator.go:96` — `syncPeerNoProgressLimit = 5 * time.Minute`

This is the no-progress deadline that clears a sync peer serving headers but no bodies — the exact griefing vector this PR hardens against. The PR made four related knobs tunable (`MaxUnvalidatedAdvertisedHeightLead`, `MaxUnprovenSyncProbesPerBackoffWindow`, `FullStoragePenaltyDuration`, `FullDeliveryFreshnessWindow`) but left this one a compile-time const. Operators on slow links / archival cold-starts can't lengthen it and defenders can't shorten it without a rebuild. Suggest promoting to a `P2PSettings` field (e.g. `p2p_sync_peer_no_progress_limit`, default 5m) with the same fallback pattern as `fullStoragePenaltyDuration()`. (Minor: it does sit alongside `fastMonitorInterval`/`slowMonitorInterval` which are also consts, so leaving it is defensible — flagging for consistency with the PR's own philosophy.)

### 4. (Low) `128`-byte chainwork bound duplicated across packages

`services/p2p/handle_catchup_metrics.go:16` (`maxReportedChainWorkBytes = 128`) duplicates `services/blockchain/peer_registry.go:26` (`maxValidatedChainWorkBytes = 128`). Two independent consts guarding the same conceptual limit (outer RPC check + inner store re-check); if one changes and the other doesn't they silently diverge. Suggest exporting one as the single source of truth, or cross-linking them with a comment.

### 5. (Low) Comment invariant in `chainWorkGreater` is false on first call

`services/p2p/sync_coordinator.go:155` — the comment states "Callers only pass real, positive chainwork (both slices non-empty)", but `resetProbeBudgetIfLocalChainWorkAdvanced` calls it with `sc.lastLocalChainWork`, which is empty on the first invocation. Behaviour is still correct (empty compares as zero), but the stated invariant is wrong — suggest softening the comment.

### 6. (Low) `FullStorageContradictions` incremented via cross-RPC read-modify-write

`services/p2p/handle_catchup_metrics.go:122` reads the counter via `GetPeer` then writes `+1` via a separate `RegisterPeer` RPC whose merge overwrites rather than accumulates. Two concurrent block-incomplete failures for the same peer can collapse to one increment. Benign today — the counter is only serialized/displayed and never gates a decision. Worth moving the increment inside the registry under its lock if it's ever promoted to a threshold; otherwise a "best-effort/advisory" comment would help.

### 7. (Low) `FullStoragePenalty` on `ErrBlockIncomplete` could transiently demote an honest peer

`services/p2p/handle_catchup_metrics.go:95` applies the penalty for any `ErrBlockIncomplete` regardless of storage class, and treats a `Storage:"full"` claim as contradicted. If `ErrBlockIncomplete` can fire on a benign local race (body requested before it's locally available) rather than genuine peer inability, an honest full peer eats a 1h demotion. Self-healing on next delivery, so minor — just worth confirming `ErrBlockIncomplete` only signals genuine peer-side incompleteness.

### 8. (Low) `filterEligiblePeers` is now test-only

`services/p2p/sync_coordinator.go:736` — production callers all use `filterEligiblePeersWithTip` directly; this thin wrapper is only reached from tests. Consider moving it to the test file or routing a production caller through it.

---

**Verified clean:** chainwork math (`big.Int` unsigned big-endian, no uint64 truncation, per-block `CalcBlockWork` matches consensus); byte-order consistent across DB tip / local tip / peer-reported work; work credited only post-PoW and only to the serving peer; strictly-monotonic absolute writes (no double-credit/replay inflation); consensus tip selection untouched; height clamp applies to advertised metadata only (deep multi-window catchup unthrottled); probe-budget check-and-decrement under a single lock with `currentSyncPeer` as a mutex slot (no TOCTOU/double-spend); no blocking RPC/Kafka under `sc.mu`; proto changes additive-only (rolling-upgrade safe); settings wiring complete with matching defaults; best-effort RPC errors logged at `Warnf` not swallowed.

icellan (changes requested):
Re-reviewed the six follow-up commits. Five are solid — configurable no-progress timeout, validated-work decay (clean nil-wipe under the registry lock, idempotent against duplicate penalties, no arithmetic underflow), atomic full-storage-contradiction counter (single-lock `+=`, race-tested with 200 goroutines), the registry-layer chainwork-cap test, and the longdesc correction. One blocks merge:

**P1 — `services/p2p/sync_coordinator.go:980` — opportunistic preemption re-pins the stalled proven incumbent and defeats the no-progress eviction**

The preemption path does `ClearSyncPeer()` + `TriggerSync()` rather than activating the candidate it just chose. `TriggerSync` reads `oldPeer := sc.currentSyncPeerLocked()` *after* the clear (line 451), so re-selection runs with `oldPeer=""`/`PreviousPeer=""` — the just-evicted incumbent is neither excluded (`filterEligiblePeersWithTip`) nor anti-affinity-bumped (`peer_selector.go` `selectedIndex` path). The selector's proven-first sort (`peer_selector.go:137`, `iProven` ranked above validated chainwork) then re-picks the proven-but-stalled incumbent over the unproven higher-validated-work candidate, and `claimSelectedSyncPeer` resets `lastSyncProgressTime = now`.

Because the preemption guard is half the no-progress timeout (`preemptionProgressGuard() = timeout/2`, ~2.5m at the 5m default), every periodic evaluation resets the progress clock before the 5m hard eviction (`syncPeerNoProgressTimedOut`) can ever fire. Net effect: a *proven* peer that drips/stalls holds the sync slot up to the 24h full-delivery freshness window instead of the intended 5m bound, and a faster unproven peer is never tried. So the preemption is a no-op for its own stated target and regresses the no-progress eviction this PR relies on.

The four shipped preemption tests all use an *unproven* incumbent (`BlocksReceived=0`, no `LastBlockTime`), so they exercise the path where preemption happens to work and never hit the proven-incumbent slow-drip profile the commit targets.

Fix: record `RecordSyncAttempt(currentPeer)` before `ClearSyncPeer` (mirroring `clearNoProgressSyncPeer`) so the immediate re-selection excludes the incumbent during its cooldown, or claim the already-chosen candidate atomically instead of clear-then-reselect; and do not reset the incumbent's `syncStartTime`/`lastSyncProgressTime` when it is re-pinned. Please add a regression test with a proven incumbent (`BlocksReceived>0`, recent `LastBlockTime`) plus an unproven higher-work candidate.

**Non-blocking follow-ups (fine as separate items):**
- `services/p2p/handle_catchup_metrics.go:119` — the new work-decay + 1h full-storage penalty also fires on the *transient/local* `ErrBlockIncomplete` case (missing-parent, `isPeerError=false`, per issue #1031), demoting an honest and possibly sole-source ahead peer. Distinguish the transient-local source from the genuinely peer-attributable one at origin (they currently share the `reportIncompleteBlock` flag) so only the latter is penalized.
- `services/p2p/sync_coordinator.go:966` — the preemption guard is a hardcoded half of the now-configurable timeout. An operator who raises `p2p_sync_peer_no_progress_timeout` to tolerate large-block delivery on a slow link still exposes an unproven peer to preemption at half that value. Document the halving in the setting's longdesc, or expose the fraction as its own setting.
- nits: `RecordValidatedPeerProgress`'s own `>128`-byte branch (`peer_registry.go:1119`) is still untested for the over-cap case; the `sync_coordinator.go:957` comment / commit title says "materially higher" but the comparison is a strict zero-margin `>`; and the sole-top-work slow-dripper residual (no higher-work rival, evicted only via the hard timeout) is undocumented.

ctnguyen:
Thanks @oskarszoon @freemans13 @icellan, addressed points :

#### @oskarszoon's points addressed (approving review, 2026-07-08)

✅ **OPEN-1** — preemption now activates the chosen candidate **atomically** (never leaves the node peerless) instead of clear-then-reselect. *(commit 00e00ca06; also @icellan)*
✅ **OPEN-2** — `reportValidatedHeaderProgress` now runs under a bounded `context.WithTimeout(...)` (matching `releaseCatchupLock`); no unbounded service ctx under the single-flight lock. *(commit b97740c44)*
✅ **OPEN-3** — `RecordValidatedPeerProgress` refuses to credit validated work while `now.Before(FullStoragePenaltyUntil)`; complementary to the delivery-failure decay wipe. *(commit 2e1dbecae)*
✅ **OPEN-4** — forced sync peers are exempt from the claim-time unproven-probe budget in `claimSelectedSyncPeer`; tested through `TriggerSync`. *(commit 00e00ca06)*
✅ **NIT-1** — startup warning when the unproven-probe budget is `0` and no forced peer is set (preserving "0 = disable"); clamp tested negative/zero/positive; longdesc documents the cold-start caveat. *(commits 00e00ca06 + 546ebab90)*
✅ **NIT-2** — dead test-only `filterEligiblePeers` removed, three test call sites repointed. *(commit 00e00ca06; also @freemans13)*
✅ **NIT-4** — added `sanitizeAdvertisedTip` clamp/overflow tests; new `Client_test.go` assertions converted to `require`. *(commit 00e00ca06)*

#### @freemans13's points addressed (review notes, 2026-07-08)

✅ **OPEN-5** (finding 1) — validated-ahead peers are exempt from the claim-time probe budget (claim tier now mirrors the filter tier); tested through `TriggerSync`. *(commit 00e00ca06)*
✅ **OPEN-6** (finding 7) — transient-local (issue-1031) incompletes are marked at origin and no longer penalized; only genuinely peer-attributable incompletes open a penalty window. *(commit b97740c44; also @icellan)*
✅ **NIT-8** (finding 4) — the 128-byte cap is now a single exported source of truth (`blockchain.MaxValidatedChainWorkBytes`), aliased on the p2p side. *(commit 2e1dbecae)*
✅ **NIT-9** (finding 5) — the `chainWorkGreater` comment is softened to acknowledge empty/nil operands (first call) are treated as zero. *(commit 00e00ca06)*
✅ **NIT-2** (finding 8) — dead test-only `filterEligiblePeers` removed. *(commit 00e00ca06; also @oskarszoon)*
✅ **REPLY-1** (finding 6, atomic `FullStorageContradictions`) — already fixed by an earlier PR commit; reply-only, text in §5.4. *(no code change)*
✅ **REPLY-2** (finding 2, validated work never decays) — already fixed by an earlier PR commit; reply-only, text in §5.4. *(no code change)*
✅ **REPLY-3** (finding 3, hardcoded no-progress limit) — already fixed by an earlier PR commit; reply-only, text in §5.4. *(no code change)*

#### @icellan's points addressed (change request, 2026-07-10)

✅ **OPEN-1** (P1 blocker) — atomic preemption + benched incumbent put in cooldown via `RecordSyncAttempt` + incumbent progress clock not reset. Regression test with a **proven** incumbent + unproven higher-work candidate (fails against the old path). *(commit 00e00ca06; also @oskarszoon)*
✅ **OPEN-6** (non-blocking) — transient-local `ERR_BLOCK_INCOMPLETE` (issue 1031) distinguished at origin so only peer-attributable incompletes are penalized. *(commit b97740c44; also @freemans13)*
✅ **OPEN-7** (non-blocking) — the preemption-guard halving is documented on the `SyncPeerNoProgressTimeout` longdesc. *(commit 546ebab90)*
✅ **NIT-5** — the over-cap (>128-byte) rejection branch of `RecordValidatedPeerProgress` is now tested (129 rejected, 128 accepted). *(commit 2e1dbecae)*
✅ **NIT-6** — the "materially higher" wording reconciled with the strict zero-margin `>` comparison ("strictly-higher-work"). *(commit 00e00ca06)*
✅ **NIT-7** — the sole-top-work slow-dripper residual (evicted only by the hard timeout) is documented in-code. *(commit 00e00ca06)*

#### Automated inline review thread (2026-07-08)

✅ **OPEN-8** — the stale `AllowPrunedNodeFallback` and `ForceSyncPeer` longdescs (and the `usage:` tag) now describe the validated-work ranking; the "youngest / lowest-height / minimize UTXO pruning risk" wording is gone. *(commit 546ebab90)*

**Deferred (not done) — @oskarszoon nit:**

[ ] **NIT-3** — regenerate the 4 `*.pb.go` with the pinned protoc (v34.1). The local toolchain is v33.2 (the downgraded version), so regenerating here would not fix the spurious diff. Do this on a host with the pinned toolchain, or let it regenerate at merge.

## Recap resolved points

| ID | Point | Reviewer | Blocking | Code Change | Criticality |
|----|-------|----------|----------|-------------|-------------|
| OPEN-1 | Preemption re-pins stalled proven incumbent | Siggi, Oli | YES | YES | 100% |
| OPEN-2 | Validated-header report has no timeout inside catchup lock | Oli | NO | YES | 72% |
| OPEN-3 | Validated work re-credited during penalty window | Oli | NO | YES | 68% |
| OPEN-5 | Validated-ahead peer rejected by probe budget at claim | Stu | NO | YES | 62% |
| OPEN-4 | Forced peer not exempt from probe budget at claim | Oli | NO | YES | 58% |
| OPEN-6 | Transient-local incomplete block penalizes honest peer | Siggi, Stu | NO | YES | 55% |
| OPEN-8 | Stale pruned-fallback / force-peer longdescs | AI | NO | YES | 46% |
| OPEN-7 | Preemption-guard halving undocumented | Siggi | NO | YES | 40% |
| NIT-1 | Probe budget `0` = silent cold-start deadlock | Oli | nit | YES | 38% |
| NIT-2 | Dead `filterEligiblePeers` (test-only) | Oli, Stu | nit | YES | 30% |
| NIT-5 | Over-cap branch untested in registry crediting | Siggi | nit | YES | 26% |
| NIT-4 | Assorted test gaps; `assert`→`require` | Oli | nit | YES | 24% |
| NIT-6 | "Materially higher" comment vs strict `>` | Siggi | nit | YES | 18% |
| NIT-8 | 128-byte cap constant duplicated | Stu | nit | YES | 15% |
| NIT-9 | `chainWorkGreater` comment false on first call | Stu | nit | YES | 13% |
| NIT-7 | Sole-top-work slow-dripper residual undocumented | Siggi | nit | YES | 12% |
| NIT-3 | `*.pb.go` regenerated with downgraded protoc | Oli | nit | YES | 10% |
| REPLY-1 | Atomic `FullStorageContradictions` | Stu | NO | NO | 5% |
| REPLY-2 | Validated work never decays | Stu | NO | NO | 5% |
| REPLY-3 | Hardcoded no-progress limit | Stu | NO | NO | 5% |

**Reply only, no code:** REPLY-1, REPLY-2, REPLY-3 (already fixed by follow-up commits).
**Code change:** everything else (OPEN-1…8, NIT-1…9); post an accompanying thread reply when
pushing each fix.

icellan (approved):
Re-reviewed after the preemption fix — the P1 that my earlier request-changes blocked on is resolved, so clearing the block.

`preemptSyncPeer` (`services/p2p/sync_coordinator.go:1020`) now does an atomic slot swap under `sc.mu` directly to the already-chosen candidate — no more `ClearSyncPeer()` + `TriggerSync()` re-selection — guarded by `if sc.currentSyncPeer != incumbent { return false }` so it never strands the node peerless, resets the no-progress clock for the new candidate, and benches the incumbent via `RecordSyncAttempt` (cooldown, not a ban). The candidate is chosen with the incumbent excluded and gated on strictly-higher validated work, so it's validated-work-ahead and budget-exempt. The exact prior failure — a proven, stalled incumbent being re-pinned over an unproven higher-work candidate while the 5-min eviction clock kept resetting — no longer reproduces. The three other follow-ups are also good: transient (local, issue-#1031) `ErrBlockIncomplete` is now marked at origin and matched before the generic case so it no longer penalizes/decays an honest sole-source peer; validated-work crediting is penalty-window-gated in `RecordValidatedPeerProgress`; and the guard-fraction + doc nits are corrected. The new catchup lock-release refactor releases on all paths with no double-release/deadlock, and the transient marker survives because catchup validates in-process.

No blockers remain. A few optional, non-blocking nits for a follow-up if you want them — none need to hold this PR:

- `sync_coordinator.go:915` vs `:986` — preemption compares a stale incumbent snapshot against a freshly-refetched candidate; re-fetch the incumbent (or fetch both together) right before the comparison. Low-ms window, self-correcting.
- `sync_coordinator.go:1029` — the clock-reset residual is broader than the documented sole-dripper: a run of distinct peers each credited with successively higher validated (real-PoW) header work but withholding bodies is preempted one-to-the-next, each swap resetting the clock and deferring the hard timeout. Bounded (per-peer monotonic work isn't cheap, and a peer-attributable incomplete block wipes the offender's work), but worth either a stall baseline that survives preemption or a comment covering the multi-peer case.
- `peer_registry.go:322` — the `Register` merge path advances validated work without the penalty-window gate that `RecordValidatedPeerProgress` has; not reachable by any current production caller, so defensive-hardening only.
- `services/blockvalidation/Server.go:1394` — the gRPC `ValidateBlock` comment says "transient" but constructs the peer-attributable `NewBlockIncompleteError`; cosmetic, align the constructor or the comment.

Nice work turning this around across the revisions.

ctnguyen:
Thanks @icellan for the last review. I'll merge the PR as it is, since that was a long enough PR ( nearly 30 commits ) and all important points has been addressed.

PR #1203, opened 2026-07-01
test: consolidate duplicate mocks onto canonical mocks

icellan:
## What

Several unit tests hand-rolled their own copies of mocks that already exist canonically in the interface's home package (e.g. `stores/utxo/mock.go` → `utxo.MockUnminedTxIterator`). This consolidates the clearly-misplaced duplicates and removes dead mock declarations. Test-only changes.

## Changes

**Replace byte-for-byte duplicate iterator mocks with the canonical `utxo.MockUnminedTxIterator`:**
- `stores/utxo/pruner_coverage_test.go` (`mockIterator`)
- `stores/utxo/logger/logger_test.go` (`MockIterator`)
- `services/blockassembly/pruner_test.go` (`MockUnminedTxIterator`)

**Delete unused/dead mock declarations (never instantiated):**
- `services/blockvalidation/Server_test.go` (`mockBlockValidationInterface`)
- `services/blockassembly/server_test.go` (`MockBlockchainClientForCoverage`)
- `services/blockassembly/BlockAssembler_test.go` (`MockCleanupService`)
- `services/validator/http_server_test.go` (`MockValidator`)
- `pkg/k8sresolver/k8s_test.go` (`MockKubernetesInterface`)

Net: 8 files, −182 lines, no production code touched.

## Scope note

This is the safe, high-confidence subset of a broader mock-placement audit. Other candidates were deliberately **left as-is** after per-site verification — they turned out to be legitimate purpose-built test doubles (stateful fakes, log-capturing loggers, panic-guard stores) or would require breaking many `.On()` call sites, not genuine misplacements.

## Verification

- `go vet` green on all affected packages
- Affected package tests pass (`stores/utxo`, `stores/utxo/logger`, `services/blockassembly`, `pkg/k8sresolver`, `services/validator`)

### DIFF
--- a/pkg/k8sresolver/k8s_test.go
+++ b/pkg/k8sresolver/k8s_test.go
@@ -36,12 +36,6 @@ func (m *MockLogger) Warnf(format string, args ...interface{}) {
 	m.Called(format, args)
 }
 
-// MockKubernetesInterface for testing serviceClient without real K8s cluster
-type MockKubernetesInterface struct {
-	mock.Mock
-	kubernetes.Interface
-}
-
 // Helper to create test serviceClient
 func createTestServiceClient(k8s kubernetes.Interface, namespace string) *serviceClient {
 	logger := &MockLogger{}
--- a/services/blockassembly/BlockAssembler_test.go
+++ b/services/blockassembly/BlockAssembler_test.go
@@ -2002,24 +2002,6 @@ func TestBlockAssembly_CoinbaseCalculationFix(t *testing.T) {
 	})
 }
 
-// MockCleanupService is a mock implementation of the cleanup service interface
-type MockCleanupService struct {
-	mock.Mock
-}
-
-func (m *MockCleanupService) Start(ctx context.Context) {
-	m.Called(ctx)
-}
-
-func (m *MockCleanupService) Prune(ctx context.Context, height uint32) (int64, error) {
-	args := m.Called(ctx, height)
-	return args.Get(0).(int64), args.Error(1)
-}
-
-func (m *MockCleanupService) SetPersistedHeightGetter(getter func() uint32) {
-	m.Called(getter)
-}
-
 // containsHash is a helper to check if a slice of hashes contains a specific hash
 func containsHash(list []chainhash.Hash, target chainhash.Hash) bool {
 	for _, h := range list {
--- a/services/blockassembly/pruner_test.go
+++ b/services/blockassembly/pruner_test.go
@@ -33,7 +33,7 @@ func TestCleanupDuringStartup(t *testing.T) {
 		var iteratorCalled bool
 
 		// Then iterator should be called
-		mockIterator := new(MockUnminedTxIterator)
+		mockIterator := new(utxo.MockUnminedTxIterator)
 		mockStore.On("GetUnminedTxIterator").
 			Return(mockIterator, nil).
 			Run(func(args mock.Arguments) {
@@ -75,30 +75,6 @@ func TestCleanupDuringStartup(t *testing.T) {
 	})
 }
 
-// Mock types for testing
-
-type MockUnminedTxIterator struct {
-	mock.Mock
-}
-
-func (m *MockUnminedTxIterator) Next(ctx context.Context) ([]*utxo.UnminedTransaction, error) {
-	args := m.Called(ctx)
-	if args.Get(0) == nil {
-		return nil, args.Error(1)
-	}
-
-	return args.Get(0).([]*utxo.UnminedTransaction), args.Error(1)
-}
-
-func (m *MockUnminedTxIterator) Err() error {
-	args := m.Called()
-	return args.Error(0)
-}
-
-func (m *MockUnminedTxIterator) Close() error {
-	args := m.Called()
-	return args.Error(0)
-}
 
 // TestLoadUnminedTransactionsExcludesConflicting tests that loadUnminedTransactions excludes conflicting transactions
 func TestLoadUnminedTransactionsExcludesConflicting(t *testing.T) {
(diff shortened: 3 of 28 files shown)


ctnguyen (approved):
This is exactly the kind of low-risk hygiene change that pays down maintenance debt without incurring behavioural risk. The three iterator-mock replacements are safe: the canonical `utxo.MockUnminedTxIterator` (defined in the non-test `stores/utxo/mock.go`, so it is importable from consuming packages) is byte-for-byte identical to the removed local copies — the `Next`/`Err`/`Close` signatures and bodies match, so no test semantics change. The five deleted mock types (`MockKubernetesInterface`, `MockCleanupService`, `MockBlockchainClientForCoverage`, `mockBlockValidationInterface`, and the external-test `MockValidator`) each have zero remaining instantiations, and the collateral import removals (`go-bt`, `go-subtree` in `http_server_test.go`) have no other uses in their file. Worth calling out that the reviewer correctly navigated a subtle trap: the deleted `MockValidator` lives in `package validator_test`, so it does not collide with the untouched production `validator.MockValidator` in `Interface.go` (line 138's `var _ Interface = &MockValidator{}` assertion still resolves to the internal type). The scope note is honest about what was deliberately left alone, and the description matches the diff precisely.

ctnguyen (approved):
## Approved

The consolidation is executed carefully and end-to-end: every one of the 15 files that referenced `validator.MockValidatorClient` is updated, the three duplicated `UnminedTxIterator` mocks collapse onto the canonical `utxo.MockUnminedTxIterator`, and the dead mock declarations are genuinely unused (their sibling imports remain heavily used, so nothing is orphaned). The merge of `MockValidatorClient` and the old no-op `MockValidator` into one type is thoughtfully done — the new precedence chain (`ValidateFunc` → error queue → `UtxoStore` → tx-derived default) is well documented, and adding the `UtxoStore != nil` guard makes the zero-value mock strictly safer than before (it no longer nil-panics). The zero-value behaviour matches the old no-op mock, so the many untouched call sites that use `validator.MockValidator{}` keep working. There were no substantive inline review threads or line comments to carry forward.

## Non-blocking issues

### ChiR1 — PR description understates scope and mislabels the changed files
**Problem:** The body claims "Net: 8 files, −182 lines, no production code touched" and "Test-only changes," but the actual diff is 28 files, +129/−473, and it renames the exported `MockValidatorClient` → `MockValidator` (merging in and deleting the old no-op `MockValidator`). Two of the edited files — `services/validator/Interface.go` and `services/validator/Mock.go` — are non-`_test.go` files compiled into the package's normal build (as is `services/blockvalidation/catchup_test_suite.go`), so "no production code touched" is not accurate, and this is an exported-symbol rename, not just internal test wiring.
**Why it matters:** Reviewers and anyone reading the history will size their review and their testing to the stated 8-file/test-only scope and miss the extra packages (`propagation`, `legacy/netsync`, `blockvalidation`, `subtreevalidation`, and the `test/longtest` variant) whose `validator.MockValidator{}` semantics changed through the merge.
**Fix:** Update the description to describe the `MockValidatorClient`→`MockValidator` consolidation, list the real file/line counts, and note that non-test mock files (`Interface.go`, `Mock.go`) are affected. No code change required.

### ChiR2 — Stated verification omits the packages whose `MockValidator` semantics changed
**Problem:** The Verification section lists only `stores/utxo`, `stores/utxo/logger`, `services/blockassembly`, `pkg/k8sresolver`, and `services/validator`. The merge changes the concrete type behind `validator.MockValidator` for `services/propagation`, `services/legacy/netsync`, `services/blockvalidation`, `services/subtreevalidation`, and the `test/longtest/services/subtreevalidation` package — none of which are in that list.
**Why it matters:** These packages construct `&validator.MockValidator{}` (zero value) and embed the type (e.g. `countingValidator` in `services/legacy/netsync/handle_block_test.go:697`); the merged type now carries a `sync.Mutex`, and its `Validate` path changed. I verified by inspection that zero-value behaviour is equivalent, that the embedders only use pointers (so `go vet` copylocks is clean), and that the deletions orphan no imports — but that verification should be recorded, not assumed.
**Fix:** Run and note `go build ./...`, `go vet ./...`, and `go test` for the four affected service packages, and either include the `test/longtest` subtreevalidation package in that run or explicitly state it is intentionally out of scope. The captured artifacts do not show which commands CI actually ran, so confirm rather than presume coverage.

### ChiR3 — Merged mock discards the caller `context` in `Validate`
**Problem:** In `services/validator/Mock.go`, `MockValidator.Validate` ignores its received context and delegates with a fresh `context.Background()`, and `ValidateWithOptions` likewise passes `context.Background()` into `UtxoStore.Create`. The behaviour is carried over verbatim from the old `MockValidatorClient`, but the newly added `ValidateFunc` path is now reachable through `Validate` and inherits the same dropped context.
**Why it matters:** Callers such as `propagation` and `legacy/netsync` invoke `validator.Validate(ctx, …)`; a `ValidateFunc` (or the `UtxoStore.Create` path) set up to observe cancellation, a deadline, or trace values will never see them, which can mask cancellation-related test expectations.
**Fix:** Thread the received context through instead of substituting `context.Background()`:
```go
func (m *MockValidator) Validate(ctx context.Context, tx *bt.Tx, blockHeight uint32, opts ...Option) (*meta.Data, error) {
	return m.ValidateWithOptions(ctx, tx, blockHeight, ProcessOptions(opts...))
}
// …and in ValidateWithOptions, prefer the received ctx for the store call:
	return m.UtxoStore.Create(ctx, tx, 0)
```

## Recap

| ChiR1 | Description understates scope / mislabels files | | 35% |
| ChiR2 | Verification omits affected packages | | 28% |
| ChiR3 | Merged mock drops caller context | | 22% |

freemans13 (approved):

PR #1204, opened 2026-07-02
refactor(blockvalidation): remove dead subtreeDeDuplicator (#4661)

gokutheengineer:
DeDuplicate was never called in non-test code. Block-level dedup is handled by blocksCurrentlyValidating and subtree-level dedup lives in the subtreevalidation service, so the DeDuplicator was redundant dead code carrying two latent bugs (unbounded cache growth via an unread deleteAt, and a cond.Wait loop that ignored ctx cancellation).

Remove deduplicator.go and its test, the subtreeDeDuplicator field and its construction, the test struct-literal references, and the stale docs describing behavior that never ran.


ctnguyen (approved):
This is a clean, well-scoped dead-code removal. I verified the central claim independently: there are no call sites of `.DeDuplicate(` anywhere in production code, and no references to `DeDuplicator`, `NewDeDuplicator`, or the `subtreeDeDuplicator` field survive outside the four files this PR touches — the type was only ever wired up in the constructor and referenced from test struct literals, never invoked on a real validation path. The removal is complete and self-contained: `deduplicator.go` and its test go together, the struct field and its `NewDeDuplicator(...)` construction are dropped from `BlockValidation.go`, all three test struct-literal references are removed, and the two documentation sections describing the never-executed behavior are deleted so nothing is left describing phantom functionality. The description is also honest about why the code was worth removing rather than keeping — the unread `deleteAt` meant the cache would have grown without bound, and the `cond.Wait()` loop ignored context cancellation — both of which I confirmed against the deleted source, so removing this reduces latent risk rather than merely tidying up.

freemans13:
Nice clean removal — verified no dangling references to `DeDuplicator`/`subtreeDeDuplicator` remain in the Go tree.

One stale-docs miss though: `docs/topics/services/blockValidation.md:1111-1112` still lists the deleted files in its directory-structure map:

```
├── deduplicator.go                         - Subtree deduplication to prevent redundant processing.
├── deduplicator_test.go                    - Deduplication tests.
```

The PR removed the prose section (2.2.11) but left this file-tree listing, so it now points at files that no longer exist — which contradicts the PR's stated goal of removing the stale docs. Could you drop those two lines?

icellan:
**blockValidation.md:1111** — the package source-tree listing still names `deduplicator.go` and `deduplicator_test.go`, but this PR deletes them. Drop those two lines from the directory map so the docs don't reference files that no longer exist.

freemans13 (approved):
Approving. Verified this is a clean, fully-justified dead-code removal:

- `DeDuplicate` is never called anywhere in non-test code (grep repo-wide → zero call sites); only the constructor was wired into the struct.
- Block-level dedup is genuinely handled by `blocksCurrentlyValidating` (`SetIfNotExists` at BlockValidation.go:1142), and the still-used `invalidSubtreeDeDuplicateMap` in subtreevalidation is correctly left untouched.
- Both cited latent bugs are real: `deleteAt` was written but never read (no eviction path → unbounded growth), and the `cond.Wait()` loop had no ctx-aware exit.
- Docs cleaned consistently — no dangling "dedup" references remain.
- `go build` and `go vet ./services/blockvalidation/` both pass (vet compiles the test package too, so the removed struct-literal references are all gone cleanly). No dangling references to `DeDuplicator`/`NewDeDuplicator`/`.DeDuplicate`.

Nice cleanup.

PR #1205, opened 2026-07-02
fix(blockvalidation): build orphan fork ID from full block hash

gokutheengineer:
**Closes #4663.**

### Problem
`DetermineForkID` derived the orphan fork ID from only the last 8 hex chars of the block hash (`fork_manager.go:559-561`) — 32 bits of entropy. Two distinct orphan blocks sharing those 8 chars collide onto one `ForkBranch`: the `!exists` branch is skipped, so the second orphan reuses the first fork's branch and is tracked under the wrong fork's lifecycle. A ~1-in-4-billion accidental collision, and trivially forced on a public network.

### Fix
Use the full block hash (256 bits) for the orphan fork ID, so distinct orphan blocks can never alias onto the same fork.

### DIFF
--- a/services/blockvalidation/fork_manager.go
+++ b/services/blockvalidation/fork_manager.go
@@ -557,8 +557,10 @@ func (fm *ForkManager) DetermineForkID(ctx context.Context, block *model.Block, 
 	}
 
 	if err != nil {
-		blockHashStr := block.Hash().String()
-		forkID := "orphan-" + blockHashStr[len(blockHashStr)-8:]
+		// Use the full block hash (256 bits) so distinct orphan blocks never collide
+		// onto the same ForkBranch. Truncating to the last 8 hex chars (32 bits) allowed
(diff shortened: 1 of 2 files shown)


icellan (approved):

ctnguyen (approved):
This is a precise, well-diagnosed fix. The root cause is called out exactly: deriving the orphan fork ID from only the last 8 hex chars (32 bits) let two distinct orphan blocks alias onto one `ForkBranch`, and because the block/parent mapping is written only inside the `if _, exists := fm.forks[forkID]; !exists` guard, the second colliding orphan silently skipped registration and was tracked under the wrong fork's lifecycle. Switching to the full 256-bit hash removes the aliasing entirely, and the change is a single line with no ripple effects — no code parses the `orphan-` prefix and the Prometheus label is the constant `"orphan"`, so there is no metrics-cardinality impact. The two added tests are accurate against the real `MockStore`/`NewLocalClient` APIs and cover both the full-hash property and the distinct-forks behaviour.

PR #1206, opened 2026-07-02
fix blockvalidation WaitForBlock fork-blocked spin

gokutheengineer:
Closes #4657.

**Problem**

WaitForBlock is meant to park on a condition variable when no block is processable, "to avoid busy loop burning CPU". But the double-check guard at services/blockvalidation/block_priority.go:270 only tested len(pq.items) > 0. GetAllBlocked is by definition returned on a non-empty queue, so the guard was always true, the early continue always fired, and cond.Wait() was never reached. A worker spun at 100% CPU re-running Get() → CanProcessBlock() (fork-manager lookups) for the entire time every queued block was fork-blocked, multiplied across workers. The intended back-pressure never engaged.

Confirmed still present at the cited location. Single-item issue, no items skipped.

**Fix**

Gate the early-continue on status == GetEmpty && len(pq.items) > 0. Only GetEmpty can be resolved by items appearing; GetAllBlocked now falls through to cond.Wait() and is woken by the existing Signal() calls on Add (block_priority.go:154), FinishProcessingBlock (fork_manager.go:383) and Release (fork_manager.go:91). Strictly less work than the spin, matching the proposed fix.

I verified the wake-up path before adopting it: a worker only parks on GetAllBlocked while other workers hold the processing slots that block it, and those workers Signal() on finish, so parking cannot stall the pipeline. There is a narrow theoretical lost-wakeup window (the fork-manager signals under fm.mu, not pq.mu), but the finishing worker re-enters Get() itself and keeps draining the queue, so a missed wake to a spare parked worker only leaves it idle until the next Add/finish - never a deadlock.


ctnguyen (approved):
The root-cause analysis is exactly right and the fix is minimal: the old double-check guard `len(pq.items) > 0` was unconditionally true for `GetAllBlocked` (an all-blocked queue is non-empty by definition), so the worker never reached `cond.Wait()` and burned a core re-running `Get()` → `CanProcessBlock()`. Gating the early-continue on `status == GetEmpty && len(pq.items) > 0` correctly lets `GetAllBlocked` fall through to park while keeping the classic condition-variable pattern intact for the `GetEmpty` case. The PR body is honest about the residual lost-wakeup window and correctly reasons it is non-fatal — I traced the worker loop (`Server.go:1624`→`1638`) and confirmed the finishing worker always re-enters `Get()` after `FinishProcessingBlock`, so a missed wake to a spare parked worker only leaves it briefly idle, never deadlocked. The added test exercises the parking behaviour with a thread-safe mock and a bounded call-count assertion, and uses `require` per project convention.

freemans13:
Nice fix — the root cause and wake-up safety analysis both hold up on review.

One minor nit worth correcting before merge:

In the new comment at `services/blockvalidation/block_priority.go:273`:

> ```
> // Add/removeItem/fork-resolution Signal wake us when a block becomes processable
> ```

`removeItem` (block_priority.go:116) doesn't actually call `Signal()` — it only mutates the slice/index and updates metrics. The real signalers are the three you correctly list in the PR description: `Add` (block_priority.go:154), `FinishProcessingBlock` (fork_manager.go:383), and `Release` (fork_manager.go:91).

Could you update the comment to reference those instead of `removeItem`? It'll keep the wake-up path accurate for the next reader auditing correctness here.

freemans13 (approved):
Reviewed and verified. Correct fix for a real bug: the double-check guard at `block_priority.go` gated only on `len(pq.items) > 0`, which is always true for `GetAllBlocked` (that status is returned on a non-empty queue), so `cond.Wait()` was never reached and workers spun re-running `Get()`→`CanProcessBlock()` at 100% CPU. Gating on `status == GetEmpty && len(pq.items) > 0` lets `GetAllBlocked` fall through to park; `GetEmpty` behaviour is unchanged.

Verified the two correctness risks:
- `cond.Wait()` is called with `pq.mu` held and the outer loop re-checks the predicate via `Get()` on wake — spurious wakeups handled correctly.
- The lost-wakeup window (fork-manager signals under `fm.mu`, not `pq.mu`) is benign: `blockProcessingWorker` loops straight back into `WaitForBlock`→`Get()` after `FinishProcessingBlock`, so the worker that frees a slot always re-drains the queue. A missed wake only leaves a spare worker idle until the next Add/finish — no stall or deadlock. Honestly disclosed in the PR body.

Checks: `go test -race` passes; the new test held over `-count=10 -race` (no flakiness); `go vet` clean.

Minor, non-blocking: in the new test, if the worker hasn't reached `cond.Wait()` when `pq.Signal()` fires, the signal is lost and `wg.Wait()` would hang until timeout rather than fail cleanly. The 150ms sleep makes this very unlikely and it held up over 10 race runs. The `≤ 10` assertion direction is well-chosen (a slow CI box makes fewer calls, never a false failure). Not worth changing.

LGTM.

PR #1207, opened 2026-07-02
fix(blockvalidation): drop throwaway BlockPriorityQueue in updateQueueSizeMetrics

gokutheengineer:
Closes #4664

updateQueueSizeMetrics heap-allocated a throwaway BlockPriorityQueue struct on every queue mutation (Add, removeItem, RequeueForRetry) solely to call getQueueStats, which reads only the items slice and no field of the receiver. Make getQueueStats a plain package-level function and call it directly, removing the allocation from the block-scheduling hot path.

### DIFF
--- a/services/blockvalidation/block_priority.go
+++ b/services/blockvalidation/block_priority.go
@@ -343,7 +343,7 @@ func (pq *BlockPriorityQueue) GetAlternativeSource(blockHash *chainhash.Hash) (p
 }
 
 // getQueueStats returns statistics about the queue without locking
-func (pq *BlockPriorityQueue) getQueueStats(items []*PrioritizedBlock) (chainExtending, nearFork, deepFork int) {
+func getQueueStats(items []*PrioritizedBlock) (chainExtending, nearFork, deepFork int) {
 	for _, item := range items {
 		switch item.priority {
 		case PriorityChainExtending:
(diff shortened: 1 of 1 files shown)


ctnguyen (approved):
This is a tight, well-scoped change that does exactly what its description claims and nothing more. The author correctly observed that `getQueueStats` reads only its `items` argument and touches no field of its receiver, so promoting it to a package-level function is safe and lets `updateQueueSizeMetrics` skip the per-mutation heap allocation of a throwaway `&BlockPriorityQueue{items: items}`. Both call sites — the exported `GetQueueStats` wrapper (which still holds the lock) and `updateQueueSizeMetrics` — are updated consistently, and no other caller of the old method remains. The optimization sits on a genuine hot path (invoked from `Add`, `removeItem`, and `RequeueForRetry` on every queue mutation) while preserving identical observable behaviour and metric output.

freemans13 (approved):
Reviewed the diff: behavior-preserving refactor that drops a throwaway BlockPriorityQueue allocation from the metrics hot path. getQueueStats only ever read the passed items slice, so converting it to a package-level function is safe. All three updateQueueSizeMetrics callers pass pq.items under lock. Build, vet, and 30 priority-queue tests pass. LGTM.

PR #1212, opened 2026-07-02
test(main): accept non-semver v-prefixed tags in version format check

icellan:
Extracted from #828 (feat/teranode-native-ops); test-only.

The version-format check rejected `v`-prefixed non-semver tags (e.g. build tags like `v-scaling-test-YYYYMMDD`). Accept them.

### Verified
- `go build ./...` clean against `main`
- `go test ./ -run TestPopulateVersionInfo` passes

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -280,8 +280,13 @@ func TestPopulateVersionInfoComponents(t *testing.T) {
 			assert.True(t, len(parts) >= 3, "Timestamp-based version should have format v0.0.0-TIMESTAMP-COMMIT")
 			assert.Equal(t, "v0.0.0", parts[0], "Timestamp-based version should start with v0.0.0")
 		} else {
-			// Tag-based format (can have any format, but typically vX.Y.Z with optional suffix)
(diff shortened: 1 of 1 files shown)


ctnguyen (approved):
A small, correct test-only fix. The version-format assertion is loosened from a semver-only pattern (`^v\d+\.\d+\.\d+`) to `^v\S+$`, matching what `populateVersionInfo` actually produces. Ready to merge.

oskarszoon (approved):
Approve. The new `^v\S+$` matches what `populateVersionInfo` actually does — `main.go:34` sets `version` to the literal git tag whenever it starts with `v` (`strings.HasPrefix(gitTag, "v")`), with no semver requirement — so the old `^v\d+\.\d+\.\d+` was asserting a shape the code never guarantees. It's also a real fix, not cosmetic: with non-semver deploy tags like `v-scaling-test-YYYYMMDD` now in use, a build at such a tag produces a version that passes the production v-prefix gate but would fail the old semver assertion, reddening CI. The relaxed regex still enforces the genuine invariant (v-prefixed, single token).

PR #1215, opened 2026-07-02
Add TTL to blockvalidation processBlockNotify cache

gokutheengineer:
**Closes #4665.**

### Problem
`processBlockNotify` was constructed with a bare `ttlcache.New[chainhash.Hash, bool]()`, so `ttlcache.DefaultTTL` used at every `Set` resolved to `0` (no expiry). The reaper goroutine (`go u.processBlockNotify.Start()`) had no per-entry TTL to act on, and the inline "auto-cleanup via TTL" comment was inaccurate. Entries were removed only by explicit `Delete` calls, so any missed `Delete` on an error/early-return branch leaked the entry permanently. The asymmetry with the sibling `catchupAlternatives` cache (10m TTL, same in-flight block) confirmed the omission was unintentional.

### Fix
Give `processBlockNotify` a 10-minute TTL, matching `catchupAlternatives`. Explicit `Delete` calls still remove entries promptly on the normal paths; the TTL is a safety net so a missed `Delete` cannot leak an entry forever. Added a comment explaining the intent.

### Tests
Added `TestServer_processBlockNotifyHasTTL` in `services/blockvalidation/Server_test.go`: builds the server via `New(...)`, sets an entry with `ttlcache.DefaultTTL`, and asserts the item's `ExpiresAt()` is non-zero and in the future — guarding against a regression back to a TTL-less cache.


freemans13 (approved):
Reviewed. Focused, well-reasoned change: gives `processBlockNotify` a 10-minute TTL as a safety net against missed `Delete` calls leaking cache entries, mirroring the sibling `catchupAlternatives` cache. Verified the regression-guard test compiles and passes. LGTM.

Minor (non-blocking): `ttlcache` touches-on-hit by default, so the `Get` dedup gate at Server.go:2010 extends the TTL on every duplicate `blockFound`. A missed-`Delete` entry that keeps being re-announced within each 10m window won't expire — the same failure mode `blockCatchupAttempts` guards against with `WithDisableTouchOnHit`. Bounded in practice; consider `WithDisableTouchOnHit` if a hard 'cannot leak forever' guarantee is desired.

ctnguyen (approved):
## Approved

This is a tight, well-scoped fix that correctly diagnoses a real defect: `processBlockNotify` was built with a bare `ttlcache.New[...]()`, so every `Set(..., ttlcache.DefaultTTL)` resolved to `0` (no expiry) and the reaper goroutine (`go u.processBlockNotify.Start()`) had nothing to collect — leaving any missed `Delete` on an error/early-return branch to leak the entry permanently. Giving it a 10-minute TTL to match the sibling `catchupAlternatives` cache (same in-flight block) is the right call, the inline comment now accurately documents the "safety net" intent, and the added `TestServer_processBlockNotifyHasTTL` is a genuine regression guard that pins `DefaultTTL` to a real future expiry. The diff is minimal and touches nothing beyond the fix.

icellan (approved):

PR #1216, opened 2026-07-02
chore(sonar): exclude test files from cognitive complexity rule

liam:
## Summary

Excludes test files (`**/*_test.go`) from SonarQube's Cognitive Complexity rule (`go:S3776`).

## Why

`go:S3776` currently fires on **99 test files — ~29% of the open S3776 issues** — including the single worst offender in the whole project, `services/legacy/txscript/sign_test.go` (complexity **435**). Table-driven test setup naturally inflates cognitive complexity (big literal tables, nested case builders) without signalling any real maintainability debt, and it can't be cleared by tuning the threshold because it's a *file class*, not a complexity level.

With the threshold recently raised to 25, S3776 sits at 345 open issues:

| Category | Count |
|---|---|
| Production | 220 |
| Test files | 99 |
| cmd/compose tooling | 26 |

This change drops the list to **246** (production + tooling) and removes the permanent 435 outlier from the top of the list, so the remaining issues reflect actual production code.

## Change

A `sonar.issue.ignore.multicriteria` entry scoping `go:S3776` to exclude `**/*_test.go`. The rule stays fully active for all non-test code. No source or behaviour changes.

## Verification

Config-only change. Takes effect on the next SonarCloud analysis; the 99 test-file S3776 issues will auto-resolve.

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -19,3 +19,11 @@ sonar.cpd.exclusions=poc/**,ui/dashboard/**,cmd/**
 sonar.tests=.
 sonar.test.inclusions=**/*_test.go
 sonar.python.version=3.12
+
+# Don't raise Cognitive Complexity (go:S3776) on test files. Table-driven test
+# setup inflates the metric without signalling real maintainability debt, and it
+# dominates the issue list (e.g. sign_test.go). Keep the rule active for
+# production code only.
+sonar.issue.ignore.multicriteria=e1
+sonar.issue.ignore.multicriteria.e1.ruleKey=go:S3776
+sonar.issue.ignore.multicriteria.e1.resourceKey=**/*_test.go


oskarszoon (approved):
Approve. Config-only, correct mechanism. Confirmed against the full file at HEAD that there's no pre-existing `sonar.issue.ignore.multicriteria` key, so the new `e1` entry doesn't silently clobber an earlier one (properties are last-key-wins). Syntax is the standard three-line multicriteria idiom, `go:S3776` is the right Go cognitive-complexity rule key, and `**/*_test.go` scopes it to the test file class while leaving the rule fully active on production + tooling. A per-rule `resourceKey` ignore is the correct fix here — it's a file class, not a threshold level, so it can't be cleared by tuning. One thing to be aware of (not a blocker): this also stops flagging any genuinely over-complex test *helper*, which is the intended tradeoff.

ordishs (approved):
Config-only change. Verified: `sonar.issue.ignore.multicriteria` is defined only here (no clobbering), the multicriteria/ruleKey/resourceKey format is correct, `go:S3776` scoped to `**/*_test.go` leaves production and cmd/ code fully covered. Clean and well-scoped.

PR #1218, opened 2026-07-02
Ba spec reset reorg

sugh01:


ctnguyen (changes requested):
Review at commit [2a1c22507411](https://github.com/bsv-blockchain/teranode/commit/2a1c22507411)

This PR adds two adversarial reorg test files — `reorg_contract_test.go` (orchestration-layer probes against `BlockAssembler.handleReorg`/`processNewBlockAnnouncement`) and `subtreeprocessor/reorg_equivalence_test.go` (subtree-processor equivalence, conflict, duplicate, and rollback probes). The test design is strong, but two subtests are committed as hard `require` assertions that fail against the current implementation, which turns `go test ./...` red for the whole branch — so the PR cannot merge as-is.

## What's done well
The fixtures are careful and genuinely adversarial: `newReorgMockStp` stubs *both* the in-place and reset paths so a wrong branch fails cleanly with an assertion rather than an unexpected-call panic, the in-memory-vs-from-store equivalence harness is a thoughtful way to pin BA-REORG-010, and the current-subtree-invalidation and duplicate-move-forward tests cover real gaps in the existing suite. The comments explain the intent and the spec references precisely, and the settle-then-assert pattern (`require.Eventually` on `GetCurrentLength`) is the right way to tame the async processor. These tests also surface two legitimate spec/impl divergences that are worth fixing.

## Blocking issues

### ChiR1 — `reorg_contract_test.go` deep-reorg boundary subtest hard-fails CI
**Problem:** The subtest "depth == maturity must reorg in place (EXPECTED FAIL: impl uses >=)" (`reorg_contract_test.go:137-142`; `t.Run` at :137, `require.NoError(t, err, ...)` at :139, `AssertNotCalled(..., "Reset", ...)` at :142) runs `runReorgDecision(t, 1001, 1)`. At the reviewed commit the gate at `BlockAssembler.go:1656` is `(len(moveBackBlocksWithMeta) >= CoinbaseMaturity || len(moveForwardBlocksWithMeta) >= CoinbaseMaturity) && currentHeight > 1000`, which for `moveBack=1, CoinbaseMaturity=1, currentHeight=1001` evaluates `1 >= 1 && 1001 > 1000` → true. `handleReorg` therefore takes the reset path and returns a non-nil `BlockAssemblyResetError` (`BlockAssembler.go:1717`), and `Reset` is invoked — so the assertion aborts at `:139` and `:142` would also fail. (Line numbers are as they appear at the linked commit `2a1c225`, matching the in-code comment on `reorg_contract_test.go:138` and the review bot; `main` has since drifted the gate down to `~:1643` and the reset return to `~:1655`, but this test-only PR does not touch `BlockAssembler.go`, so the reviewed-commit numbers are the ones a reader following the commit link will land on.)
**Why it matters:** A deliberately-failing assertion with no accompanying production change breaks `go test ./...` for everyone on the branch, violating the project's green-CI rule.
**Fix:** Pick one and apply it consistently with ChiR2: (a) include the one-line production change `>=` → `>` at `BlockAssembler.go:1656` **only if** the reset boundary is genuinely meant to move — this shifts a runtime behaviour of reorg handling (when a full block-assembly reset fires on a maturity-depth reorg) and would need the existing positive "large reorg triggers reset" test re-tuned; resolve the spec/impl divergence on its own merits (fix the spec if `>=` is correct, fix the code if `>` is), never flip it merely to make this probe pass; (b) `t.Skip("BA-REORG-006: impl uses >= at BlockAssembler.go:1656; see <tracking issue>")`; or (c) convert to a non-failing documented-gap `t.Logf`. Do not commit it as a hard failure. Track the gap per ChiR3.

### ChiR2 — `TestReorg_RollbackOnPartialFailure` hard-fails CI
**Problem:** `reorg_equivalence_test.go:702-705` asserts `require.NoError(t, stp.CheckSubtreeProcessor())` (at :702) and `require.False(t, txMap.Exists(txBack))` (at :704). In `reorgBlocks`, `disableCurrentTxMapPool = true` is set for the whole call (`SubtreeProcessor.go:3194`), and the else-branch snapshot `originalCurrentTxMap := stp.currentTxMap` (`SubtreeProcessor.go:~3292`) captures the *same* map value that the move-back mutates in place (it re-adds `txBack`). The deferred rollback reassigns that same pointer, so `txBack` is still present after rollback, and `CheckSubtreeProcessor` then sees `currentTxMap` (txA, txB, txBack) larger than the restored subtrees (txA, txB) — a size mismatch (`currentTxMap size 3 does not match txCount 2`). The `CheckSubtreeProcessor` assertion fails first at `:702`; the subsequent `txMap.Exists(txBack)` assertion at `:704` would also fail but is never reached, because `require` aborts the test at the first failure.
**Why it matters:** Same as ChiR1 — a committed hard-failing test reddens CI for the branch; the gap it documents is real but is masked in production by the mandatory `reset()` fallback, so there is no production fix in this PR to make the assertion pass.
**Fix:** Same three options as ChiR1, applied consistently: skip with a tracking issue, revert these two lines to the non-failing documented-gap `t.Logf`, or land the production fix that makes rollback deep-restore `currentTxMap` content (e.g. snapshot/clone the map before move-back rather than aliasing the live pointer). Keep the earlier assertions in the test (tip unchanged, txA/txB survive) as they pass today. Track the gap per ChiR3.

## Non-blocking issues

### ChiR3 — (meta / cross-cutting) Track both spec/impl gaps as issues and apply one consistent treatment across the two files
**Problem:** This is not a third independent code defect — it is the cross-file consistency and tracking obligation that spans ChiR1 and ChiR2, consolidated here so neither blocking item has to restate it. ChiR1 and ChiR2 each offer skip / log / fix; whatever is chosen, the two spec-probe tests must be treated identically. The head commit ("make TestReorg_RollbackOnPartialFailure failing") pushed them *apart*, turning the equivalence probe into a hard `require` while the contract probe was already one, so both now redden CI. If the remedy is to skip or downgrade, the underlying divergences (BA-REORG-006 uses `>=` where the spec says "exceeds"/`>`; BA-REORG-007's rollback restores the best-block reference but not `currentTxMap` content) must not evaporate into an un-referenced log line.
**Why it matters:** These divergences are the actual value of the PR. Keeping the tracking-and-consistency angle as one explicit meta-item is clearer than duplicating it inside both blocking findings.
**Fix:** File one tracking issue per gap and reference it in the `t.Skip`/`t.Logf` message on **both** files, so the treatments match and the findings survive as tracked work rather than a silently softened comment.

## Recap
| ChiR1 | Deep-reorg boundary subtest hard-fails CI | YES | 82% |
| ChiR2 | Rollback test hard-fails CI | YES | 80% |
| ChiR3 | Track both gaps as issues; treat both files consistently | | 38% |

PR #1220, opened 2026-07-02
feat(utxopersister): one-shot --end-height UTXO-set build

ordishs:
## Summary

Adds a one-shot mode to `teranode-cli utxopersister` that consolidates the existing per-block `utxo-additions` / `utxo-deletions` delta files into a single cumulative `utxo-set` (plus its `utxo-headers`) **up to an explicit block height**, then exits — instead of running the continuous tip−100 service loop.

```
# genesis -> 300
teranode-cli utxopersister --end-height 300

# build on top of an existing set at 229 -> 300
teranode-cli utxopersister --start-height 229 --end-height 300

# build and advance the service's resume marker
teranode-cli utxopersister --end-height 300 --update-last-processed
```

Without `--end-height`, the command behaves exactly as before (starts the long-running service).

## Behaviour

- **Reuses the existing consolidation primitives** (`NewConsolidator` → `ConsolidateBlockRange` → `CreateUTXOSet` → `WriteHeadersToStore`); `BuildUTXOSetToHeight` is a parameterised sibling of `processNextBlock` with a caller-supplied range, no confirmation window, and no state mutation beyond an opt-in `lastProcessed.dat` write.
- **`--start-height M`** (default 0 = genesis) builds on top of the existing `utxo-set` at M; the base set is validated (via the existing stored-hash guard) and never deleted.
- **Require complete deltas:** before writing anything, every block in `[start+1 .. end]` must have *both* its `utxo-additions` and `utxo-deletions` files; any gap (tail or middle) fails cleanly with no partial output.
- **`--update-last-processed`** (default off) advances the service resume marker to the built height.
- **Requires direct mode** (`UTXOPersisterDirect=true` + configured `blockchain_store`), since height→hash resolution and the `utxo-headers` write need the blockchain store; guarded with a clear error otherwise.
- **Re-run is fail-loud:** re-building an existing height errors ("already exists") and writes nothing — delete the target set first to rebuild.

## Incidental fix

While adding direct-mode test coverage, this surfaced a pre-existing crash in the shared `WriteHeadersToStore`: the `utxo-headers` `.sha256` checksum sidecar was written without `WithSkipHeader`, so `fileformat.NewHeader` panicked on the unregistered `"utxo-headers.sha256"` type. This panics on both the file and memory blob stores and is hit on any direct-mode header write. It went unnoticed because the daemon runs utxopersister in client mode (never reaching that branch). Fixed by skipping the magic header on the checksum sidecar (a checksum file should not carry one) and covered by a new direct-mode test.

## Testing

- `go test -race -tags testtxmetacache ./services/utxopersister/... ./cmd/utxopersister/... ./cmd/teranodecli/...` — all pass.
- `go vet` clean; `golangci-lint run` — 0 issues across the touched packages.
- New tests: genesis happy path, opt-in marker, argument guards, chain-too-short, missing additions/deletions (asserts no write on failure), missing base set, incremental-equals-single-pass (0→M→N == 0→N with base preserved), and direct-mode headers write (file store + sqlitememory chain).


ctnguyen (changes requested):
This is a clean, well-scoped feature that reuses the existing consolidation primitives (`NewConsolidator` → `ConsolidateBlockRange` → `CreateUTXOSet` → `WriteHeadersToStore`) rather than duplicating them, so `BuildUTXOSetToHeight` reads as a natural, parameterised sibling of `processNextBlock`. The default (no `--end-height`) path is preserved exactly, the CLI guards are explicit, and error returns replace the previous config `panic` paths. The pre-flight delta check is a genuinely nice touch: it prevents a missing-delta gap (tail or middle) from ever writing a target set. Test coverage is thorough — genesis happy path, incremental-equals-single-pass with base preservation, missing-delta no-write assertions, chain-too-short, and a real file-store + sqlitememory direct-mode path.

## Prior review follow-up

- Previously-raised `.sha256` panic (inline comment on `services/utxopersister/UTXOSet.go`): **still open** — carried forward as ChiR1. The `WithSkipHeader` fix closes the file backend only.
- Previously-raised minor note on the `go.mod`/`go.sum` bumps and the new `tstn` settings context: **still open** — carried forward as ChiR4.

## Blocking issues

### ChiR1 — `.sha256` sidecar still panics on S3 and memory blockstores
**File:** `services/utxopersister/UTXOSet.go:938` (fix), `stores/blob/s3/s3.go:314` & `:256`, `stores/blob/memory/memory.go:217`
**Criticality: 60%**

- **Observed:** The fix adds `options.WithSkipHeader(true)` to the `.sha256` sidecar write, but only the **file** backend honors `SkipHeader` (`stores/blob/file/file.go:769`). Both S3 (`Set` at `s3.go:314`, `SetFromReader` at `:256`) and memory (`memory.go:217`) call `fileformat.NewHeader(fileType)` unconditionally. `fileType` here is `FileTypeUtxoHeaders+".sha256"` = `"utxo-headers.sha256"`, which is **not** a registered type (`pkg/fileformat/header.go:67`), so `NewHeader` panics with `unknown file type "utxo-headers.sha256"`.
- **Why it matters:** `RunUtxoPersisterToHeight` builds the blockstore from the configured `blockstore` URL with no `file://` restriction. A direct-mode `--end-height` build against an **S3 blockstore** — a realistic production configuration — will crash (a panic, not a returned error, so it is not caught by the `Warnf` guard) after the utxo-set has already been written but before headers complete. The PR description's claim that the panic is fixed for the *memory* store is also inaccurate; the new direct-mode test uses a `file://` store precisely because memory ignores the flag, leaving the S3/memory paths unexercised.
- **Fix:** Use a backend-agnostic approach instead of a per-backend option only one backend implements. Register the checksum pseudo-type in `fileTypeToMagic`, or make the S3/memory `Set` honor `SkipHeader`, or write the checksum via a code path that never calls `NewHeader`. For example, mirror the file backend:

```go
// memory.go / s3.go Set:
if !merged.SkipHeader {
    header := fileformat.NewHeader(fileType)
    data = append(data, header.Bytes()...)
}
```

### ChiR2 — Failed `utxo-headers` write is swallowed, leaving an unrepairable partial artifact
**File:** `services/utxopersister/Server.go:328-342`
**Criticality: 55%**

- **Observed:** When `s.blockchainStore != nil` (always true in direct mode), a `WriteHeadersToStore` error is logged via `s.logger.Warnf(...)`; control then falls through to the `updateLastProcessed` write (`writeLastHeight(ctx, endHeight)`) and `BuildUTXOSetToHeight` returns `nil` (success).
- **Why it matters:** The command's stated job is to build the utxo-set **plus its utxo-headers**. On a headers-write failure it reports success while producing an incomplete artifact — and with `--update-last-processed` it also advances the resume marker to a height whose headers are missing. Worse, the PR advertises fail-loud re-run behaviour: re-running errors because the target utxo-set already exists, so the missing headers **cannot be repaired** without manually deleting the set first. A scripted caller gets no error to act on.
- **Fix:** Return the `WriteHeadersToStore` error rather than warning (and do not update `lastProcessed` when it fails). If a non-fatal headers write is genuinely intended to match `processNextBlock`, at least skip the `lastProcessed` advance and surface a non-zero exit status so the incomplete build is visible.

## Non-blocking issues

### ChiR3 — Set and headers are resolved from two independent chain views
**File:** `services/utxopersister/Server.go:246`, `:308-329`; `services/utxopersister/consolidator.go:230`; `services/utxopersister/UTXOSet.go:539`
**Criticality: 45%**

- **Observed:** `BuildUTXOSetToHeight` resolves `hashByHeight`/`targetHash` and runs the pre-flight from one `GetBlockHeadersByHeight(from, endHeight)` call. It then calls `ConsolidateBlockRange`, which issues its **own** `GetBlockHeadersByHeight` and sets `c.lastBlockHash` from that second fetch; `CreateUTXOSet` writes the set keyed under `c.lastBlockHash` (`UTXOSet.go:539`), while `WriteHeadersToStore`, logging, and `lastProcessed` all use the earlier `targetHash`.
- **Why it matters:** If the active chain changes between the two reads (or the header source returns a different view), the set is written under one tip while headers/marker point at another — a silent set/headers mismatch. The window is narrow for an offline one-shot build against a settled store, hence non-blocking, but the guard is cheap.
- **Fix:** Reuse a single resolved header range for both the pre-flight and consolidation, or assert before writing headers/updating the marker that `c.lastBlockHash` equals `hashByHeight[endHeight]` (and that the first consolidated block links to `seedHash`).

### ChiR4 — Unrelated changes bundled into the feature
**File:** `settings.conf` (new `tstn` context lines), `go.mod` / `go.sum`
**Criticality: 20%**

- **Observed:** The PR adds a new `tstn` settings context (`blockchain_store.tstn`, `network.tstn`, `startPruner.tstn`, etc.) and bumps `go-chaincfg`, `gocore`, and `go-wire`. None of these are exercised by the one-shot build.
- **Why it matters:** Mixing environment/config and dependency bumps into a feature PR (the Small Diff Rule in `AGENTS.md`) makes the change harder to review and to revert cleanly, and caused the automated pass to flag them as likely rebase artifacts.
- **Fix:** Split the `tstn` context and dependency bumps into a separate chore PR, or note in the description why they are required here (e.g. if `ChainCfgParams.GenesisHash` needs the `go-chaincfg` bump).

### ChiR5 — Silent `uint` → `uint32` truncation of height flags
**File:** `cmd/teranodecli/teranodecli/cli.go:204-223` (the `case "utxopersister"` flag defs and `Execute` closure)
**Criticality: 12%**

- **Observed:** `--start-height` / `--end-height` are `flag.Uint` (platform-width `uint`); the CLI validates `*startHeight >= *endHeight` on the full-width values, then passes `uint32(*startHeight)`, `uint32(*endHeight)` to the builder.
- **Why it matters:** On 64-bit platforms a value above `2^32` truncates silently, so validation could pass on the wide values while the truncated pair is inconsistent. Realistic block heights make this practically unreachable, hence low priority.
- **Fix:** Validate against `math.MaxUint32` (or parse directly into `uint32`) before the range check, returning a clear error on overflow.

## Recap

| ChiR1 | `.sha256` panic on S3/memory blockstores | YES | 60% |
| ChiR2 | Failed headers write swallowed; unrepairable partial | YES | 55% |
| ChiR3 | Set vs headers from two chain views | | 45% |
| ChiR4 | Unrelated `tstn`/dep-bump changes bundled in | | 20% |
| ChiR5 | Silent uint→uint32 height truncation | | 12% |

teracoder (commented):
DevOps/correctness pass on the seed build+load path. Five points, ranked. #1 and #2 are the important ones: both are "important failure silently reported as success," which for a seed tool is the worst failure mode. #3–#5 are smaller robustness/cleanup items.

services/utxopersister/Server.go
```diff
@@ -571,6 +571,146 @@ func (s *Server) processNextBlock(ctx context.Context) (time.Duration, error) {
...
+	}
+
+	targetHash := hashByHeight[endHeight]
+
+	// Write the utxo-headers file when a full blockchain store is available
+	// (always the case in direct mode; skipped for client-only setups, matching
+	// processNextBlock's handling).
+	if s.blockchainStore != nil {
+		if err := WriteHeadersToStore(ctx, s.logger, s.settings, s.blockStore, s.blockchainStore, targetHash, endHeight); err != nil {
+			s.logger.Warnf("[UTXOPersister] error writing headers file for height %d: %v", endHeight, err)
```
teracoder: **One-shot ships an incomplete seed on header-write failure.** This warn-only handling is copied from `processNextBlock`, but that runs in a loop and retries on the next tick — the one-shot has no retry. If `WriteHeadersToStore` fails here, the `.utxo-set` is written, `.utxo-headers` is missing, and `BuildUTXOSetToHeight` still `return nil` → the CLI exits 0 and the operator believes the seed is complete. `seeder` then can't import headers from a file that was never produced.

Suggest: in the one-shot path treat a header-write failure as fatal (`return err`) rather than `Warnf`.

cmd/seeder/seeder.go
```diff
@@ -182,46 +216,43 @@ func Seeder(logger ulogger.Logger, appSettings *settings.Settings, inputDir stri
...
+
+	// Persist the BlockAssembler checkpoint only after the UTXO set has been
+	// imported successfully. The checkpoint reflects the UTXO store's
+	// completeness (block assembly is the sole creator of coinbase UTXOs), so it
+	// must never be advanced when UTXOs were skipped or failed. When UTXOs were
+	// skipped because lastProcessed.dat already existed, utxoTip is nil and any
+	// existing checkpoint is left untouched.
+	if !skipUTXOs && utxoErr == nil && utxoTip != nil {
+		if err := writeBlockAssemblerState(ctx, logger, blockchainStore, utxoTip); err != nil {
+			logger.Errorf("Failed to set BlockAssembler state: %v", err)
```
teracoder: **Checkpoint failure is reported as success.** This is only `logger.Errorf`'d and `Seeder` sets no non-zero exit, so the seeder can import headers + UTXOs, fail to write the BlockAssembler checkpoint, and still "complete." The node then starts assembly from genesis — the auto-checkpoint that is the headline of this PR silently no-ops. Same class as the `Server.go` header case.

Suggest: make this fatal, or at minimum propagate a non-zero exit so the failure is visible to whoever ran the seed.

cmd/utxopersister/utxo_persister.go
```diff
@@ -162,3 +164,50 @@ func RunUtxoPersister(logger ulogger.Logger, settings *settings.Settings) {
...
+	}
+
+	blockStore, err := blob.NewStore(logger, blockStoreURL, options.WithHashPrefix(hashPrefix))
+	if err != nil {
+		return errors.NewStorageError("[UTXOPersister] failed to create blockStore", err)
+	}
+
+	blockchainStore, err := blockchainstore.NewStore(logger, settings.BlockChain.StoreURL, settings)
+	if err != nil {
+		return errors.NewStorageError("[UTXOPersister] failed to create blockchainStore", err)
```
teracoder, line 204: **Opened stores are never closed.** `blockStore` and `blockchainStore` are created here but there's no `defer ...Close(ctx)`. It's a one-shot so the leak is bounded by process exit, but there's no clean-shutdown / flush guarantee on the blockchain store before the tool returns success.

Suggest: `defer blockchainStore.Close(ctx)` (and `blockStore` if it owns resources).

cmd/seeder/seeder.go
```diff
@@ -566,6 +614,58 @@ func processUTXO(ctx context.Context, store utxo.Store, utxoWrapper *utxopersist
...
+// pre-flight check and the final state write.
+func newBlockchainStore(logger ulogger.Logger, appSettings *settings.Settings) (blockchain.Store, error) {
+	blockchainStoreURL := appSettings.BlockChain.StoreURL
+	if blockchainStoreURL == nil {
+		return nil, errors.NewConfigurationError("blockchain store URL not found in config")
+	}
+
+	q := blockchainStoreURL.Query()
+	q.Set("seeder", "true")
+	blockchainStoreURL.RawQuery = q.Encode()
```
teracoder, line 780: **Mutates shared settings in place.** `blockchainStoreURL` is the pointer held in `appSettings.BlockChain.StoreURL`; setting `RawQuery` here permanently stamps `seeder=true` onto the shared `*url.URL`, so any later reader of the settings sees it. Carried over from the old `processHeaders`, but now it lives in a reused helper.

Suggest: copy the URL before mutating (e.g. `u := *blockchainStoreURL; ...` and build the query on the copy).

cmd/seeder/seeder.go
```diff
@@ -566,6 +614,58 @@ func processUTXO(ctx context.Context, store utxo.Store, utxoWrapper *utxopersist
...
+	return blockchain.NewStore(logger, blockchainStoreURL, appSettings)
+}
+
+// blockAssemblerStateExists reports whether a BlockAssembler checkpoint is
+// already persisted in the blockchain store. A missing checkpoint is the normal
+// state of an unseeded node and is not treated as an error.
+func blockAssemblerStateExists(ctx context.Context, store blockchain.Store) (bool, error) {
+	data, err := store.GetState(ctx, blockassembly.StateKey)
+	if err != nil {
+		if errors.Is(err, errors.ErrNotFound) || strings.Contains(err.Error(), sql.ErrNoRows.Error()) {
```
teracoder, line 791: **Brittle error classification.** Matching `strings.Contains(err.Error(), sql.ErrNoRows.Error())` to mean "no state" is fragile: a backend that wraps not-found differently makes a fresh store look like a real error, or a real error look like "no state" (false negative on the overwrite guard).

Suggest: `errors.Is(err, sql.ErrNoRows)` instead of the string match.

ctnguyen (commented):
The blockassembly state codec is a genuinely good extraction: `EncodeState`/`DecodeState` now live in one place (`services/blockassembly/state.go`), `GetState`/`SetState`, `rewindblockchain`, and the asset heights endpoint all route through it, and `SetState` loses its old `append(state[:4], …)` aliasing quirk in the process — the "format never drifts" comment is now actually enforced. The new seeder checkpoint is carefully sequenced (written only after a successful UTXO import, never when UTXOs were skipped/failed, and it refuses to write a checkpoint for a tip whose header is absent), and the "refuse to seed over a store already owned by an assembler" guard is a sensible safety addition. Test coverage for the new codec and seeder-state paths is solid.

## Status of previous review points

| ID | Prior finding | Required | Status |
|----|---------------|----------|--------|
| ChiR1 | `.sha256` sidecar panics on S3/memory blockstores | YES | **Not fixed** — code unchanged; S3/memory still `NewHeader` unconditionally |
| ChiR2 | Failed `utxo-headers` write swallowed; success returned | YES | **Not fixed** — still `Warnf` then advances marker and returns `nil` |
| ChiR3 | Set vs headers resolved from two chain views | — | **Not fixed** — no single-range reuse or tip assertion added |
| ChiR4 | Unrelated changes bundled into the feature | — | **Not fixed (wider)** — seeder/refactor/`tstn` seeds/dep bumps added |
| ChiR5 | Silent `uint`→`uint32` height truncation | — | **Not fixed** — still `uint32(*startHeight)` after full-width validation |

Notes:

- **ChiR1** — `services/utxopersister/UTXOSet.go:938` still passes only `options.WithSkipHeader(true)`. Only the file backend honours that flag (`stores/blob/file/file.go:769`); `stores/blob/s3/s3.go:256,314` and `stores/blob/memory/memory.go:217` still call `fileformat.NewHeader(fileType)` on the unregistered `"utxo-headers.sha256"` type, which panics (`pkg/fileformat/header.go:122`). A direct-mode build against an S3 blockstore still crashes after the set is written. **Both blocking items remain the top priority for this PR.**
- **ChiR2** — `services/utxopersister/Server.go` `BuildUTXOSetToHeight` is unchanged: on a `WriteHeadersToStore` failure it logs `Warnf`, then runs `writeLastHeight(endHeight)` and returns `nil`.
- **ChiR4** — the shared codec refactor is now _partly_ justified (the new seeder checkpoint legitimately needs a shared encode/decode), so that slice is defensible. The `go-chaincfg`/`gocore`/`go-wire` bumps, the `tstn` settings context, and the committed binary seed files remain unrelated to the one-shot build and still belong in a separate PR.

services/utxopersister/Server.go
```diff
@@ -571,6 +571,161 @@ func (s *Server) processNextBlock(ctx context.Context) (time.Duration, error) {
...
+		return errors.NewProcessingError("[UTXOPersister] error creating utxo-set for height %d", endHeight, err)
+	}
+
+	targetHash := hashByHeight[endHeight]
+
+	// Write the utxo-headers file when a full blockchain store is available
+	// (always the case in direct mode; skipped for client-only setups, matching
+	// processNextBlock's handling).
+	if s.blockchainStore != nil {
+		if err := WriteHeadersToStore(ctx, s.logger, s.settings, s.blockStore, s.blockchainStore, targetHash, endHeight); err != nil {
```
teracoder, line 716: **Still open (#1).** The header write is still warn-only, so a one-shot build can produce a `.utxo-set` with no `.utxo-headers` and still `return nil` → CLI exits 0. The `// matching processNextBlock's handling` rationale doesn't carry over: the service loop retries next tick, the one-shot has no retry. The failure just resurfaces later and more confusingly — at seed-load time, when `seeder`'s (now-fatal) header pass can't find the file the build never wrote.

Since #2 made the seeder fail loud, this is the one remaining silent-incomplete-output path. Suggest `return err` here in the one-shot.

cmd/utxopersister/utxo_persister.go
```diff
@@ -162,3 +164,50 @@ func RunUtxoPersister(logger ulogger.Logger, settings *settings.Settings) {
...
+
+		hashPrefix = hp
+	}
+
+	blockStore, err := blob.NewStore(logger, blockStoreURL, options.WithHashPrefix(hashPrefix))
+	if err != nil {
+		return errors.NewStorageError("[UTXOPersister] failed to create blockStore", err)
+	}
+
+	blockchainStore, err := blockchainstore.NewStore(logger, settings.BlockChain.StoreURL, settings)
```
teracoder, line 202: **Still open (#3).** `blockStore` and `blockchainStore` are created here with no `defer ...Close(ctx)`. Bounded by process exit since it's a one-shot, but there's no clean-shutdown/flush guarantee before the tool returns success. Suggest `defer blockchainStore.Close(ctx)` (and `blockStore` if it owns resources).

cmd/seeder/seeder.go
```diff
@@ -566,6 +629,58 @@ func processUTXO(ctx context.Context, store utxo.Store, utxoWrapper *utxopersist
...
+// pre-flight check and the final state write.
+func newBlockchainStore(logger ulogger.Logger, appSettings *settings.Settings) (blockchain.Store, error) {
+	blockchainStoreURL := appSettings.BlockChain.StoreURL
+	if blockchainStoreURL == nil {
+		return nil, errors.NewConfigurationError("blockchain store URL not found in config")
+	}
+
+	q := blockchainStoreURL.Query()
+	q.Set("seeder", "true")
+	blockchainStoreURL.RawQuery = q.Encode()
```
teracoder, line 780: **Still open (#4).** `blockchainStoreURL` is the pointer held in `appSettings.BlockChain.StoreURL`; setting `RawQuery` here permanently stamps `seeder=true` onto the shared `*url.URL`, visible to any later reader of the settings. Copy the URL before mutating (e.g. `u := *blockchainStoreURL` and build the query on the copy).

cmd/seeder/seeder.go
```diff
@@ -566,6 +629,58 @@ func processUTXO(ctx context.Context, store utxo.Store, utxoWrapper *utxopersist
...
+	return blockchain.NewStore(logger, blockchainStoreURL, appSettings)
+}
+
+// blockAssemblerStateExists reports whether a BlockAssembler checkpoint is
+// already persisted in the blockchain store. A missing checkpoint is the normal
+// state of an unseeded node and is not treated as an error.
+func blockAssemblerStateExists(ctx context.Context, store blockchain.Store) (bool, error) {
+	data, err := store.GetState(ctx, blockassembly.StateKey)
+	if err != nil {
+		if errors.Is(err, errors.ErrNotFound) || strings.Contains(err.Error(), sql.ErrNoRows.Error()) {
```
teracoder, line 791: **Still open (#5).** `strings.Contains(err.Error(), sql.ErrNoRows.Error())` to classify "no state" is brittle — a backend that wraps not-found differently makes a fresh store look like a real error, or a real error look like "no state" (false negative on the overwrite guard). Prefer `errors.Is(err, sql.ErrNoRows)`.

teracoder (commented):
Re-review after update. Resolved since last pass:

- **#2 (checkpoint/seed failure reported as success)** — fixed well. `Seeder` now returns an error, propagates header/UTXO failures, and treats a checkpoint-write failure as fatal (good explanatory comment on why). Wired to a non-zero exit via `cli.go` `return seeder.Seeder(...)`.
- **Height flags** — the `uint`→`uint32` truncation is gone; `uint32Flag` rejects out-of-range values. 

Still open: #1, #3, #4, #5 (re-flagged inline below).

ctnguyen (approved):

teracoder (approved):

PR #1221, opened 2026-07-03
fix(blockvalidation): reuse precomputed fullSubtreeExists on write path

gokutheengineer:
**Closes #4666.**

## Problem

`readSubtree` already computes `fullSubtreeExists` (returned in `subtreeResult`) precisely to avoid a later store round-trip, and it is threaded through the batch as `batch.fullSubtreeExists`. The async build path (`buildSubtreeAndQueueWrite`) already consumes it, but the synchronous write path `writeSubtreeFilesFromTxs` ignored it and re-issued `subtreeStore.Exists(...)` for every subtree at write time.

- Precomputed: `services/blockvalidation/quick_validate.go:825-839`
- Redundant re-check: `services/blockvalidation/quick_validate.go:847`

At thousands of subtrees per block during fast sync, this doubled the existence lookups in the write phase.

## Fix

- Added `fullSubtreeExists bool` to `writeSubtreeFilesFromTxs` and removed its internal `subtreeStore.Exists` call.
- `writeSubtreeFilesForBatch` now passes `batch.fullSubtreeExists[localIdx]`, the value already computed during prefetch.

This makes the synchronous write path consistent with the async path, which used the precomputed value all along.

### DIFF
--- a/services/blockvalidation/quick_validate.go
+++ b/services/blockvalidation/quick_validate.go
@@ -846,12 +846,9 @@ func (u *BlockValidation) readSubtree(ctx context.Context, block *model.Block, s
 // Takes transactions directly (without coinbase nil entry).
 // Note: Subtree meta files (.subtreemeta) are intentionally skipped during quick validation
 // for performance. They will be generated on-demand if needed later.
-func (u *BlockValidation) writeSubtreeFilesFromTxs(ctx context.Context, block *model.Block, subtreeIdx int, subtree *subtreepkg.Subtree, txs []*bt.Tx, subtreeHash chainhash.Hash) error {
-	fullSubtreeExists, err := u.subtreeStore.Exists(ctx, subtreeHash[:], fileformat.FileTypeSubtree)
-	if err != nil {
-		return errors.NewProcessingError("[writeSubtreeFilesFromTxs][%s] failed to check existence of full subtree %s", block.Hash().String(), subtreeHash.String(), err)
-	}
-
+func (u *BlockValidation) writeSubtreeFilesFromTxs(ctx context.Context, block *model.Block, subtreeIdx int, subtree *subtreepkg.Subtree, txs []*bt.Tx, subtreeHash chainhash.Hash, fullSubtreeExists bool) error {
+	// fullSubtreeExists was already computed during prefetch (readSubtree); reuse it
+	// instead of issuing another subtreeStore.Exists round-trip here.
 	if !fullSubtreeExists {
 		fullSubtree, err := subtreepkg.NewIncompleteTreeByLeafCount(subtree.Size())
 		if err != nil {
@@ -1265,9 +1262,10 @@ func (u *BlockValidation) writeSubtreeFilesForBatch(ctx context.Context, block *
 		txRange := batch.txRanges[localIdx]
 		subtreeTxs := batch.batchTxs[txRange[0]:txRange[1]]
 		subtreeHash := batch.subtreeHashes[localIdx]
+		fullSubtreeExists := batch.fullSubtreeExists[localIdx]
 
 		writeG.Go(func() error {
-			return u.writeSubtreeFilesFromTxs(writeCtx, block, globalIdx, subtree, subtreeTxs, subtreeHash)
+			return u.writeSubtreeFilesFromTxs(writeCtx, block, globalIdx, subtree, subtreeTxs, subtreeHash, fullSubtreeExists)
 		})
 	}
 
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):
## Approved

The optimization is exactly right and the diff is tight: the flag is genuinely precomputed (`readSubtree` → `result.fullSubtreeExists` → `batch.fullSubtreeExists[i]`), the write batch already carried it, and both `localIdx` and prefetch use the same batch-local index so the slots line up. The stated goal — making the sync path consistent with the async `buildSubtreeAndQueueWrite`, which trusted the flag all along — is achieved cleanly, and the signature change has only one caller, which is updated in the same commit. The new test is a nice touch: the `countingSubtreeStore` wrapper asserting zero `Exists` calls directly proves the round-trip was removed rather than just relocated, and all three existence branches are exercised.

## Non-blocking issues

### ChiR1 — Trusting the prefetch flag drops the self-healing rebuild the old sync path had

**Observed:** The replaced code re-checked existence at write time, so if the full `.subtree` file was gone by then, `fullSubtreeExists` came back `false` and the code rebuilt it from the in-memory txs. The new code trusts the prefetch-time `true` and takes the load branch (`quick_validate.go:901-905`); if the file has since vanished, `Get` fails and the whole batch write returns a `NotFoundError` instead of regenerating the file it already has all the inputs for.

**Why it matters:** This is a genuine robustness regression versus the code being replaced within a narrow TOCTOU window (e.g. DAH expiry / external deletion between prefetch and write). In practice the window is small — subtree files are written with `dah = block.Height + subtreeBlockHeightRetention`, well beyond the current block's processing — and, critically, the async path (`buildSubtreeAndQueueWrite`) *already* behaves this way, so the PR makes the two paths consistent rather than introducing new divergence. Worth being aware of, not a defect.

**Suggested action:** No code change required. If you want to preserve the old self-healing behaviour later, the load branch could fall back to a rebuild *only* when the file is genuinely missing — narrow the fallback to `errors.Is(err, errors.ErrNotFound)` and let every other error (context cancellation, transient store failure, permission, corruption) propagate unchanged, so real store problems are not masked:

```go
} else {
    fullSubtreeBytes, err := u.subtreeStore.Get(ctx, subtreeHash[:], fileformat.FileTypeSubtree)
    if errors.Is(err, errors.ErrNotFound) {
        // file expired/removed since prefetch — regenerate rather than fail the batch
        return u.writeSubtreeFilesFromTxs(ctx, block, subtreeIdx, subtree, txs, subtreeHash, false)
    }
    if err != nil {
        return errors.NewNotFoundError("[writeSubtreeFilesFromTxs][%s] failed to get full subtree %s", block.Hash().String(), subtreeHash.String(), err)
    }
    ...
}
```

### ChiR2 — Test covers the leaf function but not the batch index-alignment that the PR actually changed

**Observed:** `quick_validate_write_test.go` calls `writeSubtreeFilesFromTxs` directly with a hand-passed flag. The line the PR changed in the plumbing — `fullSubtreeExists := batch.fullSubtreeExists[localIdx]` in `writeSubtreeFilesForBatch` (`quick_validate.go:1265`) — is not exercised by any test.

**Why it matters:** The most plausible way this optimization could regress is a slot-misalignment between the prefetch write (`batch.fullSubtreeExists[i]`) and the read at the write site (`[localIdx]`). That correctness property is currently only verified by inspection. A test at the `writeSubtreeFilesForBatch` level (or through prefetch) would lock the wiring in against future refactors.

**Suggested action:** Add one case that runs a small batch through `writeSubtreeFilesForBatch` with a mix of existing/non-existing subtrees and asserts each subtree took the correct branch, so the index mapping is regression-protected.

### ChiR3 — The write path no longer has independent error handling for the existence check

**Observed:** The removed code returned a `ProcessingError` when `Exists` failed. The prefetch source of the flag (`readSubtree:832-833`) swallows the error (`fullSubtreeExists, _ = u.subtreeStore.Exists(...)`), so a transient store error at prefetch silently yields `false`, and the write path will now rebuild-and-overwrite (`WithAllowOverwrite(true)`) instead of surfacing the error.

**Why it matters:** The outcome is benign — the rebuilt file is byte-equivalent (same txs/fees), so this is at worst a little wasted work, and arguably more resilient than the old hard failure. But it does mean a class of store errors is now handled only by the swallowing `_` at prefetch. Purely an awareness note on the pre-existing swallowed error, which this change now makes load-bearing for the write path.

**Suggested action:** Optional — consider logging (not failing) when the prefetch `Exists` errors, so a persistent store problem is at least observable rather than fully silent.

## Recap

| ChiR1 | Trusting flag drops self-healing rebuild (TOCTOU); consistent w/ async path |  | 35% |
| ChiR2 | Test skips batch index-alignment plumbing |  | 20% |
| ChiR3 | No independent error handling; prefetch swallows Exists error |  | 10% |

icellan (approved):

PR #1222, opened 2026-07-03
blockvalidation: avoid full queue copy in BlockPriorityQueue.Get steady-state path

gokutheengineer:
**Closes #4667.**

### Problem
`BlockPriorityQueue.Get` copied the entire `items` slice on every call before releasing the lock (`block_priority.go:181-183`), an O(N) allocation + copy paid unconditionally by every worker on every scheduling iteration via `WaitForBlock`. In steady state the front item (the chain-extending block) is immediately processable, so the full copy was wasted.

### Fix
Added a fast path: after sorting, probe only the front (highest-priority) block via `CanProcessBlock` without copying, and claim it with a new `takeIfPresent` helper that re-verifies presence under the lock (concurrency-safe against another worker taking it). The existing full-copy scan is kept intact as a slow-path fallback, entered only when the front block is blocked or already taken. This preserves all prior behavior (empty, all-blocked skip-count batching, context cancellation, panic recovery). The only cost is one extra `CanProcessBlock` on the front block in the fallback case; that call is a cheap in-memory check (atomic load + map lookups in `ForkManager`), so it is negligible.

### Tests
Added to `block_priority_get_test.go`:
- `TestBlockPriorityQueue_Get_FastPathDeepQueue` — front processable behind 99 blocked followers: returns the front, removes only it, probes exactly one block.
- `TestBlockPriorityQueue_Get_SlowPathFallbackDeepQueue` — front blocked, only the deepest follower processable: fallback scan still returns it.

### DIFF
--- a/services/blockvalidation/block_priority.go
+++ b/services/blockvalidation/block_priority.go
@@ -159,6 +159,28 @@ func (pq *BlockPriorityQueue) Add(blockFound processBlockFound, priority BlockPr
 	updateQueueSizeMetrics(pq.items)
 }
 
+// takeIfPresent removes and returns the block matching hash if it is still queued,
+// reporting whether it was found. Used by Get's fast path to claim the front block
+// without copying the whole queue.
+func (pq *BlockPriorityQueue) takeIfPresent(hash *chainhash.Hash) (processBlockFound, bool) {
+	pq.mu.Lock()
+	defer pq.mu.Unlock()
+
+	for i, currentItem := range pq.items {
+		if currentItem.blockFound.hash.IsEqual(hash) {
+			if blockQueueWaitTime != nil {
+				blockQueueWaitTime.Observe(time.Since(currentItem.timestamp).Seconds())
+			}
+
+			pq.removeItem(i)
+
+			return currentItem.blockFound, true
+		}
+	}
+
+	return processBlockFound{}, false
+}
+
 func (pq *BlockPriorityQueue) Get(ctx context.Context, bp BlockProcessor) (block processBlockFound, status GetStatus) {
 	defer func() {
 		if r := recover(); r != nil {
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):
## Approved

The change targets a genuinely hot path and the design is careful: the fast path holds the lock only long enough to sort and read `items[0]`, then does the `CanProcessBlock` round-trip lock-free, and claims the block via `takeIfPresent`, which re-verifies presence *under the lock* by hash — so it is safe against another worker taking the same block concurrently. The slow-path fallback keeps the scan's mechanics intact (skip-count batching, context cancellation, the not-found recursion, panic recovery) — apart from the sorted-snapshot regression in ChiR1 — and both the returned value (`currentItem.blockFound`) and the nil-guarded `blockQueueWaitTime` observation match the existing path exactly. The two added tests assert the two meaningful outcomes — fast path probes exactly one block; slow path still reaches the deepest processable follower — and correctly use `require`. The extra `CanProcessBlock` on the front block in the fallback case is a fair, acknowledged trade for removing the copy, and it is verifiably a read-only in-memory check (`ForkManager.CanProcessBlock` is an atomic load plus `RLock`ed map lookups, no mutation), so double-probing is side-effect-free.

## Non-blocking issues

### ChiR1 — Slow path copies `pq.items` without re-checking `needsSort`, dropping the always-sorted guarantee

**Observed:** In the fast path the queue is sorted (line ~178), then unlocked. The slow path re-acquires the lock and copies `pq.items` directly, without re-checking `pq.needsSort` / re-sorting. In the replaced code, `sortItems()` ran immediately before the copy under the *same* lock hold, so the copied slice was always sorted. There is now a window between the fast-path `Unlock()` and the slow-path re-lock in which another goroutine's `Add()` (or `RequeueForRetry()`) can set `needsSort = true` and append an item. The slow path then scans an unsorted copy and returns the *first* processable block it finds — which may not be the highest-priority one (e.g. a freshly added chain-extending block appended at the tail would be probed last, behind lower-priority forks probed first).

**Why it matters:** It is a behavioral change from the prior always-sorted-before-copy invariant. Impact is low and self-correcting (the next `Get` re-sorts, and no block is lost or double-processed), but under sustained fork activity it can transiently process a lower-priority block ahead of a chain-extending one — the exact ordering this queue exists to enforce.

**Fix:** Re-check `needsSort` after re-acquiring the lock in the slow path, before copying:

```go
pq.mu.Lock()

if len(pq.items) == 0 {
    pq.mu.Unlock()
    return processBlockFound{}, GetEmpty
}

if pq.needsSort {
    pq.sortItems()
}

itemsCopy := make([]*PrioritizedBlock, len(pq.items))
copy(itemsCopy, pq.items)
pq.mu.Unlock()
```

### ChiR2 — New tests do not exercise `takeIfPresent`'s concurrency guard or the fast-path miss branch

**Observed:** The sole justification for `takeIfPresent` (re-verifying presence under the lock) is concurrency safety, yet both new tests are single-threaded. Neither covers the branch where the front is processable in the fast path but `takeIfPresent` returns `false` because another worker already claimed it, forcing the fallback.

**Why it matters:** The most subtle part of the change — the lock-free probe followed by a re-checked claim, and the fallback when the claim loses the race — is untested, so a future regression there would pass CI. The `MockBlockProcessor` already carries a `blockDelay` field that makes an interleaving test feasible.

**Fix:** Add a test that starts two `Get` calls concurrently on a queue whose front is processable and asserts exactly one returns `GetOK` for that block and the other falls through cleanly (either a different block or `GetAllBlocked`/`GetEmpty`), with the front removed exactly once. Optionally assert `GetCallCount()` in the slow-path test to pin the "front re-probed once in fallback" cost.

## Recap

| ChiR1 | Slow path copies unsorted `items` (lost always-sorted guarantee) | | 45% |
| ChiR2 | No concurrency / fast-path-miss test coverage | | 20% |

icellan (approved):

PR #1219, opened 2026-07-02
fix(sonar): stop mistyped code-smells inflating the reliability rating

liam:
## Summary

SonarCloud's **reliability rating is C** even though there are **zero real reliability issues** open. The rating is driven by `type=BUG` issues, and there are **327 open BUG-typed MAJOR issues that are actually maintainability concerns mistyped as bugs**. This PR retypes/fixes all of them so the rating returns to **A**.

Root cause breakdown:

| Count | Source | Nature |
|---|---|---|
| 320 | `scripts/check_filenames.sh` | filename-convention violations hard-coded as `type:BUG, severity:MAJOR` |
| 7 | `external_govet:inline` | deprecated-stdlib hints imported as BUG |

Neither is a reliability defect. Because Standard-mode reliability rating = `≥1 MAJOR bug → C`, **both** groups have to be cleared to move the grade — fixing one alone leaves the other holding it at C.

## Changes

**1. Filename check → CODE_SMELL (clears 320)**
`scripts/check_filenames.sh` emitted `"type":"BUG","severity":"MAJOR"` for a snake_case naming check. Retyped to `CODE_SMELL / MINOR` so the issues land under maintainability where they belong. Verified the script still emits valid JSON (369 issues, all now `CODE_SMELL/MINOR`).

**2. go vet modernizations → fix the code (clears 7)**
Behaviour-preserving stdlib updates:
- `reflect.PtrTo` → `reflect.PointerTo`
- `reflect.Ptr` → `reflect.Pointer`
- `ioutil.ReadFile` → `os.ReadFile` (import `io/ioutil` → `os`)

Also updates 3 further `reflect.Ptr` uses in the same files (`Settings.go:74`, `cmdparse.go:243/310`) that weren't yet flagged, so they don't resurface. These are pure aliases — no behaviour change.

## Verification

```
go build ./cmd/settings/ ./services/rpc/bsvjson/         # ok
go vet ./cmd/settings/ ./services/rpc/bsvjson/ ./services/blockchain/ ./test/consensus/   # clean
bash scripts/check_filenames.sh                          # valid JSON, all CODE_SMELL/MINOR
```

Once merged and re-scanned, no `type=BUG` issues remain → reliability rating returns to A.

PR #1223, opened 2026-07-03
fix(rpc,settings): complete reflect.Ptr → reflect.Pointer modernization

liam:
## Summary

Follow-up to #1219. That PR only updated the `reflect.Ptr` uses SonarCloud had surfaced in its govet sample; a repo-wide sweep found **13 more** deprecated `reflect.Ptr` uses across `services/rpc/bsvjson/` and `settings/export.go`.

Includes the **3 MAJOR `external_govet:inline` bugs** (`help.go:104/152/318`) that are currently driving the **overall reliability rating to C**, plus the remaining uses that would surface on the next scan (`help.go:391/570`, `register.go:100/121/177/215`, `settings/export.go:124/165/204`, `cmdparse_test.go:182`).

`reflect.Ptr` is a deprecated alias for `reflect.Pointer` — pure rename, no behaviour change.

## Verification

```
go build ./services/rpc/bsvjson/ ./settings/   # ok
go vet   ./services/rpc/bsvjson/ ./settings/    # clean
go test  ./services/rpc/bsvjson/ ./settings/    # ok
```

Once merged and re-scanned, the 3 MAJOR reliability bugs clear → overall reliability rating C → B. (The remaining new-code B, from mistyped filename code-smells, is being addressed separately.)

### DIFF
--- a/services/rpc/bsvjson/cmdparse_test.go
+++ b/services/rpc/bsvjson/cmdparse_test.go
@@ -179,7 +179,7 @@ func TestAssignField(t *testing.T) {
 
 		// Inidirect through to the base types to ensure their values
 		// are the same.
-		for dst.Kind() == reflect.Ptr {
+		for dst.Kind() == reflect.Pointer {
 			dst = dst.Elem()
 		}
 
--- a/services/rpc/bsvjson/help.go
+++ b/services/rpc/bsvjson/help.go
@@ -101,7 +101,7 @@ func resultStructHelp(xT descLookupFunc, rt reflect.Type, indentLevel int) []str
 
 		// Deference pointer if needed.
 		rtfType := rtf.Type
-		if rtfType.Kind() == reflect.Ptr {
+		if rtfType.Kind() == reflect.Pointer {
 			rtfType = rtf.Type.Elem()
 		}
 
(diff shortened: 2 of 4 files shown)


teracoder (approved):

ctnguyen (approved):
This is exactly the kind of small, self-contained follow-up that is easy to review and safe to merge, and I verified its core claims against the checked-out `main` tree rather than taking them on trust. It finishes the job #1219 started: the 13 changed sites are the *only* remaining `reflect.Ptr` uses in the codebase (all within these four files, at the exact lines the description lists), so the "complete" claim genuinely holds and nothing is left behind. `reflect.Ptr` is a deprecated standard-library alias for `reflect.Pointer` (`Ptr = Pointer`, identical `reflect.Kind` value since Go 1.18), and the module targets Go 1.26, so every hunk is a behaviour-preserving rename on a fully supported API with no logic change. The diff is minimal and touches nothing beyond the rename, and it actually improves consistency — sibling files in the same package (`cmdparse.go`, `cmdinfo.go`) and `settings/redact.go` already use `reflect.Pointer`, so this brings the four laggard files in line with the established convention rather than introducing a new style. No blocking or non-blocking issues; the PR is clean and ready to merge as-is.

PR #1224, opened 2026-07-03
chore(sonar): raise coverage above 80% — exclude tooling + test blob_deletion.go

liam:
## Summary

Raises overall coverage from **79.3% → ~80.1%** via two complementary changes (both needed to cross 80%):

### 1. Exclude non-product tooling from coverage
Adds three entries to `sonar.coverage.exclusions`, consistent with the existing exclusions for entry points (`cmd/**`) and diagnostics (`tracing/**`):

| Added | What | Uncovered lines |
|---|---|---|
| `compose/cmd/**` | CLI entry points — `cmd/**` only matches top-level `cmd/`, so these slipped through | 84 |
| `internal/profiling/**` | profiling/diagnostics tooling — same class as `tracing/**` | 209 |
| `**/benchmark.go` | benchmark scaffolding (0% covered by design) | 227 |

### 2. Test `stores/blockchain/sql/blob_deletion.go`
That file was **191 lines at 0% coverage** — the largest *testable* gap. Adds table-driven tests over the `sqlitememory` store (repo convention: real store, no mocks) covering all 9 methods: schedule/upsert, get-pending, cancel (found + not-found), remove, increment-retry (below/at max), list (height/store/pagination filters), batch-complete (remove + retry + max-retry removal), and acquire-batch. Result: **0% → ~80%** on that file.

## Why both

Deliberately did **not** exclude real-but-undertested product code (`p2p/Server.go`, `catchup_performance.go`, etc.) — that would game the metric. The exclusions alone reach only ~79.9%; the blob_deletion test carries it over 80%.

## Verification

```
go test -race -count=1 ./stores/blockchain/sql/   # ok (52s)
go tool cover: blob_deletion.go 0% -> ~80%
```

Projected overall coverage after re-scan: **~80.1%**. (Overall coverage is not a gate condition — `new_coverage` is, at 86% — so this is badge polish done the honest way.)

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -14,7 +14,7 @@ sonar.go.golangci-lint.reportPaths=golangci-lint-report.xml
 sonar.externalIssuesReportPaths=sonar_filename_report.json
 
 sonar.exclusions=**test**,**/docs/**,**vendor**,ui/**,ui_v2/**,**/*.pb.go,**/mock.go,**/Mock.go,**/interface.go,**/Interface.go
-sonar.coverage.exclusions=poc/**,ui/dashboard/**,k8sresolver/**,cmd/**,tracing/**,compose/chainintegrity/**,scripts/validate-markdown.py,services/legacy/**,model/TestHelper.go,util/kafka/kafka_producer_async.go
+sonar.coverage.exclusions=poc/**,ui/dashboard/**,k8sresolver/**,cmd/**,compose/cmd/**,tracing/**,internal/profiling/**,compose/chainintegrity/**,**/benchmark.go,scripts/validate-markdown.py,services/legacy/**,model/TestHelper.go,util/kafka/kafka_producer_async.go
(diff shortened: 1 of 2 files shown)


ordishs (approved):
Reviewed the diff (Sonar coverage-exclusion config + new blob_deletion_test.go).

Verified:
- Ran the new tests: `go test -count=1 -run 'BlobDeletion|BlobDeletions' ./stores/blockchain/sql/` → ok 0.666s
- Traced every assertion against blob_deletion.go — retry-count arithmetic, upsert reset-to-0, height ordering, and pagination total semantics all match the source.
- Audited the newly-excluded paths: internal/profiling/** and compose/cmd/** are non-product tooling, and the only file matching **/benchmark.go (subtreeprocessor/benchmark.go) is reachable only from a teranodecli benchmark subcommand and a test — genuine scaffolding, not runtime code. The exclusions don't game the metric.

Tests are meaningful and pass; exclusions are honest. LGTM.

sugh01 (approved):

PR #1225, opened 2026-07-03
test(settings): guard profiler contention-profiling loader wiring

icellan:
## Summary

- \`profiler_blockProfileRate\` / \`profiler_mutexProfileFraction\` were already fully wired on \`main\` (struct tags in \`settings/interface.go\`, \`NewSettings()\` loader entries in \`settings/settings.go\`, and the \`runtime.SetBlockProfileRate\`/\`SetMutexProfileFraction\` gating in \`daemon.startProfilerAndMetrics\`) but had no dedicated test proving the loader actually reads them.
- Both default to \`0\`, which is also the Go zero value — so a missing \`getInt()\` entry in \`settings.go\` would silently leave contention profiling permanently disabled in every deployment without any test catching it.
- Adds an override-based loader test (set a non-zero value via \`gocore.Config().Set\`, call \`NewSettings()\`, assert the field changed), following the same pattern already used for this class of bug in \`TestBatcherTickInterval_LoaderReadsAllKeys\`.

No production code changes — this is test-only, closing an observability gap identified while investigating a throughput bottleneck on the dev-scale-1 cluster (mutex/block profiling wasn't enabled, so contention had to be inferred indirectly from CPU/goroutine profiles).

## Test plan

- [x] \`go build ./settings/...\`
- [x] \`go vet ./settings/...\`
- [x] \`go test -race ./settings/...\`
- [x] \`golangci-lint run ./settings/...\` — 0 issues
- [x] \`staticcheck ./settings/...\` — clean

### DIFF
--- a/settings/profiler_settings_test.go
+++ b/settings/profiler_settings_test.go
@@ -1,0 +1,54 @@
+package settings
+
+import (
+	"testing"
+
+	"github.com/ordishs/gocore"
+	"github.com/stretchr/testify/require"
+)
+
+// TestProfilerContentionSettings_LoaderReadsBothKeys guards
+// BlockProfileRate/MutexProfileFraction against the field-exists-but-loader-
+// never-reads-it bug: each field has a `key:` tag, but if NewSettings() does
+// not call getInt for it the value stays at Go zero regardless of what's
+// configured, silently leaving contention profiling permanently disabled.
+//
+// Default 0 == Go zero, so a default-value assertion alone would pass
+// spuriously even with a missing loader entry. The honest test is: set a
+// non-zero override, call NewSettings(), assert the field changed.
+func TestProfilerContentionSettings_LoaderReadsBothKeys(t *testing.T) {
+	def := NewSettings()
+	require.Equal(t, 0, def.BlockProfileRate,
+		"default BlockProfileRate must be 0 (disabled); block profiling must not run by default in prod")
+	require.Equal(t, 0, def.MutexProfileFraction,
+		"default MutexProfileFraction must be 0 (disabled); mutex profiling must not run by default in prod")
+
+	type kv struct {
+		key      string
+		override string
+		check    func(t *testing.T, s *Settings)
+	}
+
+	cases := []kv{
+		{
+			key:      "profiler_blockProfileRate",
+			override: "1000000",
+			check:    func(t *testing.T, s *Settings) { require.Equal(t, 1000000, s.BlockProfileRate) },
+		},
+		{
+			key:      "profiler_mutexProfileFraction",
+			override: "100",
+			check:    func(t *testing.T, s *Settings) { require.Equal(t, 100, s.MutexProfileFraction) },
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.key, func(t *testing.T) {
+			gocore.Config().Set(tc.key, tc.override)
+			t.Cleanup(func() { gocore.Config().Set(tc.key, "") })
+
+			s := NewSettings()
+			tc.check(t, s)
+		})
+	}
+}


ctnguyen (approved):
Review at commit [9bc808f0d84d](https://github.com/bsv-blockchain/teranode/commit/9bc808f0d84d)

This test-only PR adds `TestProfilerContentionSettings_LoaderReadsBothKeys`, which sets non-zero overrides for `profiler_blockProfileRate` / `profiler_mutexProfileFraction`, calls `NewSettings()`, and asserts the values propagate — guarding against the "key tag exists but the loader never reads it" regression, which would silently disable contention profiling in every deployment. The change is sound and safe to merge; one non-blocking hermeticity improvement is worth making.

## Non-blocking issues

### ChiR1 — Default-value assertions and cleanup are not hermetic against ambient config
- **Problem:** Lines 27-30 assert `def.BlockProfileRate == 0` / `def.MutexProfileFraction == 0` against a `NewSettings()` built from whatever `gocore` config is ambient. If a developer or a CI runner has `profiler_blockProfileRate` / `profiler_mutexProfileFraction` set via env/context (exactly the scenario this feature exists for), these assertions fail even though the loader wiring under test is perfectly correct. Separately, the subtest cleanup `gocore.Config().Set(tc.key, "")` (line 54) resets the key to empty rather than to whatever value was present before the test ran, leaving process-global config altered for any later test in the same binary.
- **Why it matters:** A test that can fail based on the runner's environment produces false negatives, undermining the very regression signal it is meant to provide. In practice the flake risk is theoretical rather than demonstrated: the sibling `TestBatcherTickInterval_LoaderReadsAllKeys` uses the identical non-hermetic pattern — plain default-value assertions against ambient config plus `gocore.Config().Set(key, "")` cleanup — and is stable in CI, so this is established, already-accepted repo convention. The cleanup blast radius is likewise small — `getInt` treats `""` identically to an unset key (both fall back to the default `0`; verified in gocore's `getNumber`), and no other code in the `settings` package reads these keys — so restoring the prior value is test hygiene (don't clobber ambient config), not a fix for a concrete downstream break. Both facts reinforce the non-blocking call.
- **Fix:** This PR faithfully mirrors the sibling's non-hermetic pattern, so the gap is pre-existing, not a divergence from convention. The genuine improvement — which the sibling also lacks — is snapshot/clear/restore: snapshot the existing value for each key, clear it before the default assertions, and restore the snapshot in cleanup. For example:
  ```go
  for _, key := range []string{"profiler_blockProfileRate", "profiler_mutexProfileFraction"} {
      prev, _ := gocore.Config().Get(key) // Get returns (string, bool)
      gocore.Config().Set(key, "")
      t.Cleanup(func() { gocore.Config().Set(key, prev) })
  }
  def := NewSettings() // now the 0-default assertions are hermetic
  ```
  Apply the same snapshot/restore in each subtest instead of unconditionally setting `""`. (The per-iteration `key`/`prev` capture in the `t.Cleanup` closure is correct under Go 1.22+ loop semantics, which this repo — `go 1.26` — uses.) For style consistency, ideally apply the same treatment to the sibling `TestBatcherTickInterval_LoaderReadsAllKeys`, or explicitly defer it so the two tests don't diverge.

## Recap

| ChiR1 | Non-hermetic default assertions / cleanup vs ambient config | | 30% |

freemans13 (approved):
Reviewed locally — clean, well-scoped test-only change.

Verified:
- Wiring the test guards is real: `key:` tags in `settings/interface.go:46-47`, `getInt()` loader entries in `settings/settings.go:63-64`, and runtime gating in `daemon/daemon_services.go:168-174`.
- Test is genuinely meaningful — non-zero overrides (1000000 / 100) defeat the default-0-equals-Go-zero false-pass trap (explicitly documented in the test).
- Follows the established package pattern exactly (`batcher_tickinterval_test.go`, `asset_settings_test.go`).
- Passes with `-race`.

LGTM 👍

PR #1226, opened 2026-07-03
feat(model,legacy): skip validOrderAndBlessed + parent-mined wait below hardcoded checkpoint

freemans13:
## Summary

**#1178 is now merged to `main`**, so this PR's diff is clean — just the four files below. (It was previously stacked on #1178 and reworked when #1178 was hardened before merge; this branch has been merged up to the current `main` and brought in line with the merged safety model.)

Below the highest **hardcoded** checkpoint, on the (default-off) `OutpointOnlyBelowCheckpoint` fast path with a store that supports the outpoint-only spend path:

1. **`model/Block.go`**: `Valid()` skips `validOrderAndBlessed` (measured 72ms of a ~353ms serial legacy-IBD critical path on mainnet). The skip is governed by the **same safety contract as the sibling fee skip** (`checkBlockRewardAndFees`), so both engage on exactly the same blocks: the operator has opted in (`OutpointOnlyBelowCheckpoint`), the store supports the fast path (`utxo.Store.SupportsOutpointOnlySpend()`), the block is a **confirmed checkpoint ancestor** (`checkpointConfirmedAncestor`, set only by blockvalidation via `SetCheckpointConfirmedAncestor`), a hardcoded checkpoint exists, and `0 < height ≤ checkpoint`. The `checkpointConfirmedAncestor` gate closes the forward/optimistic-checkpoint window and the detached-fork-reconsider window that a height-only gate would leave open. The integrity floor still runs unconditionally — PoW, timestamp/MTP, coinbase checks, `CheckMerkleRoot`, and `checkDuplicateTransactions` (CVE-2012-2459) — binding the local bytes to the checkpoint-certified chain. Precedent: the native catchup path (`quickValidateBlock`) never runs `Valid()` below the checkpoint at all.
2. **`services/legacy/netsync/handle_block.go`**: skips `waitForPreviousBlockMined` — its documented purpose is BIP68 parent-height lookup (already skipped below checkpoint); the single serial `blockHandler` plus synchronous `AddBlock(MinedSet=true)` already guarantee parent commitment. Measured always-true, 0.35ms/block over 388k calls. Gated via `legacyOutpointOnly` (flag + `SupportsOutpointOnlySpend()` + quick-validation-allowed + height ≤ checkpoint).

Every conjunct must hold; otherwise behaviour is byte-identical.

**Scope note:** the model-level gate applies to *all* `Valid()` callers when the conjuncts hold (not only legacy netsync). The safety argument is path-independent — any block reaching `Valid()` as a confirmed checkpoint ancestor carries the checkpoint-certified header chain, and the merkle root binds the body to it — and the flag is operator opt-in.

## Tests

- `TestBlock_SkipOrderAndBlessedBelowCheckpoint_Predicate` — full truth table incl. the store-support and confirmed-ancestor gates + nil settings/store
- `TestBlock_Valid_SkipsValidOrderAndBlessedBelowCheckpoint` — `Valid()` succeeds with a panic-on-any-use txMetaStore, proving the stage never runs
- `TestBlock_MerkleFloor_RejectsTamperingWhileSkipIsActive` — real stocked subtree store: a tampered merkle root is still **rejected** while the skip is active, and the correct-root case passes with the panic store
- `TestSyncManager_needsParentMinedWait` — truth table on upstream's store-capability harness
- `model`, `netsync`, `settings` packages green incl. `-race`; full `go build ./...` and `go vet` clean

https://claude.ai/code/session_01X38rKLsgpo5qn4uhrevGcD


ctnguyen (approved):
Review at commit [850bed6b33b3](https://github.com/bsv-blockchain/teranode/commit/850bed6b33b3)

This PR extends the operator-opt-in `OutpointOnlyBelowCheckpoint` fast path: below the highest *hardcoded* checkpoint (SQL UTXO store only) it skips `validOrderAndBlessed` in `model.Block.Valid`, skips the legacy netsync parent-mined wait, adds an `OutpointOnlySpend` validator mode (no parent reads, no BIP68, UTXO-hash check off), and teaches the SQL store a minimal (inputs-un-decorated) create. The design is fail-safe (every conjunct must hold or behaviour is byte-identical), the height bound is a single source of truth, and I found no blocking defect.

## Non-blocking issues

### ChiR1 — `checkBlockRewardAndFees` no-inflation skip is ungated on the feature flag
- **Problem:** The new early return `if b.Height <= HighestCheckpointHeight(params.Checkpoints) { return nil }` (~model/Block.go:745; diff line 99) is deliberately *not* gated on `OutpointOnlyBelowCheckpoint` — the code comment explicitly says "NOT gated on OutpointOnlyBelowCheckpoint". Every node that upgrades — including operators who never enable the fast path, i.e. the default configuration — stops enforcing the coinbase reward / no-inflation arithmetic for all blocks at or below the highest hardcoded checkpoint (a very large historical range on mainnet), on any `Valid()` call (e.g. `reconsiderblock`/revalidate).
- **Why it matters:** This is the one finding that is actually live in the default configuration on a consensus-critical function, not just on the opt-in path — hence it ranks at the top. The reasoning (the checkpoint-pinned hash certifies the prefix; forks below the checkpoint are rejected at the checkpoint layer; mirrors the existing quick-validate path) is sound, so this is safe — but it is a behaviour change every upgrader inherits silently and should be surfaced deliberately.
- **Fix:** Keep the behaviour if intentional (it is defensible), but call it out explicitly in the release notes/PR description as an unconditional change, and consider whether it should instead be gated so the default configuration retains full reward validation below the checkpoint.

### ChiR2 — `SkipScriptValidation` is newly wire-transmittable but, unlike `OutpointOnlySpend`, has no server-side height guard
- **Problem:** This PR adds two new proto request fields — `skip_script_validation` (field 13) and `outpoint_only_spend` (field 14). Before this change `SkipScriptValidation` was an in-process `Option` only: the client never put it on the wire and the server dropped it (the PR's own wire test comment calls the prior behaviour "silently-dropped"). Now `buildValidateTxRequest` (Client.go ~289) sends it, and both `optionsFromValidateRequest` (Server.go ~508) and the HTTP handler (`skipScriptValidation` query param, Server.go ~842) parse it. The validator guards `OutpointOnlySpend` two ways (must be paired with `SkipScriptValidation`; rejected above the highest checkpoint at ~Validator.go:715), but `SkipScriptValidation` *on its own* has **no** height/checkpoint/provenance guard. By the same "buggy or hostile gRPC caller" model used to justify the `OutpointOnlySpend` guard, any wire-reachable caller can now set `skipScriptValidation=true` alone at *any* height and, controlling `WithAddTXToBlockAssembly`, push a transaction with invalid signatures through spend+create into block assembly, bypassing BDK script/signature validation.
- **Why it matters:** This is a new externally-reachable attack surface that did not exist before the field was wired. In-tree there is no above-checkpoint caller of `WithSkipScriptValidation` (the only user is the netsync below-checkpoint fast path, always paired with the height-guarded `OutpointOnlySpend`), so it is not a live default-path behaviour change today — but exploitability is deployment-dependent on whether the validator `Validate` gRPC/HTTP endpoint is reachable by untrusted callers. It ranks below ChiR1 (which is live in the default consensus path) but above the reachable-only-in-future ChiR3.
- **Fix:** Give `SkipScriptValidation` the same defence-in-depth as its sibling — reject/ignore it from untrusted wire callers (e.g. gate it on the below-checkpoint height bound, or require a trusted-provenance signal), *or* explicitly document that the validator `Validate` API is trusted-internal-only and must not be exposed. Also confirm no above-checkpoint/steady-state caller sets `WithSkipScriptValidation`; if one is ever added, it becomes a live script-validation-skip on the default path like ChiR1.

### ChiR3 — `Valid()` skip leans on `CheckMerkleRoot`, which is not unconditional
- **Problem:** `skipOrderAndBlessedBelowCheckpoint`'s docstring states "PoW, CheckMerkleRoot and checkDuplicateTransactions (CVE-2012-2459) still run unconditionally". PoW and dedup are genuinely unconditional (step 11 runs against the in-memory `b.SubtreeSlices` with no guard, and its comment already annotates the dedup as an intentional unconditional guard held for future nil-subtreeStore callers), but `CheckMerkleRoot` only runs inside the `if subtreeStore != nil && len(b.Subtrees) > 0` block. The load-bearing point stands: with a `nil` subtreeStore and the skip active there is **no merkle binding of the body to the checkpoint-certified header** — the block passes on PoW + coinbase + in-memory dedup alone, without the merkle root that the whole safety argument relies on. The PR's own `TestBlock_Valid_SkipsValidOrderAndBlessedBelowCheckpoint` exercises exactly this (nil subtreeStore, skip active, `valid=true`), so the coupling to `CheckMerkleRoot` having run is implicit, not enforced.
- **Why it matters:** All current production callers pass a non-nil `u.subtreeStore`, so this is not reachable today — but the PR's own scope note says the model-level gate "applies to *all* `Valid()` callers," which makes a future nil-subtreeStore caller a silent consensus hole in a security-critical function.
- **Fix:** Make `subtreeStore != nil && len(b.Subtrees) > 0` a *precondition* for honouring the skip, so `CheckMerkleRoot` is guaranteed to have run before `validOrderAndBlessed` is skipped; when the precondition does not hold, do not take the skip. Note that merely falling through to `validOrderAndBlessed` does **not** restore the body↔header binding — `validOrderAndBlessed` checks order/blessing against the subtree store/slices (also nil in that case), not against `Header.HashMerkleRoot`; only `CheckMerkleRoot` establishes that binding. Correct the docstring's "unconditionally" wording to state the `subtreeStore`/`len(Subtrees)` precondition for `CheckMerkleRoot` rather than implying it always runs.

### ChiR4 — Height-0 handling differs between the model gate and the legacy/quick gates
- **Problem:** `skipOrderAndBlessedBelowCheckpoint` requires `b.Height > 0`, so it returns false at height 0, whereas `legacyOutpointOnly` (and `quickValidateOutpointOnly`) return true at height 0 (their truth tables assert `height 0 → true`). The two gates apply the same fast-path concept with a different lower bound.
- **Why it matters:** Harmless in practice (the genesis block carries only a coinbase, so `validOrderAndBlessed` is trivial there and no spends occur), but the divergence undermines the "single mental model / one boundary" story the PR otherwise maintains and could confuse a future maintainer.
- **Fix:** Align the lower bound across the three gates (either all exclude height 0 or all include it), with a one-line comment on why genesis is treated as it is.

## Recap

| ChiR1 | Reward/no-inflation skip ungated on feature flag (live in default config) | | 45% |
| ChiR2 | Newly wire-exposed `SkipScriptValidation` has no height guard (asymmetry vs `OutpointOnlySpend`) | | 40% |
| ChiR3 | Valid() skip relies on conditional CheckMerkleRoot | | 30% |
| ChiR4 | Height-0 gate inconsistency across paths | | 12% |

freemans13:
Thanks @ctnguyen — worked through all four ChiR points against the **current** diff. Since your review (at `850bed6b3`) #1178 landed on `main` separately, so this branch's diff shrank to four files; two of your findings have moved out of scope as a result, and the other two are fixed in `2715f0659`.

### ChiR1 — `checkBlockRewardAndFees` fee-skip ungated on the flag → **out of scope now**
That early return lives in the merged #1178 (`model/Block.go` on `main`); this branch's net diff only *references* `checkBlockRewardAndFees` in a comment and does not modify the function. Your recommendation to surface it explicitly as an unconditional, every-upgrader change is well taken — it belongs with #1178, and I'll flag it there rather than here.

### ChiR2 — newly wire-transmittable `SkipScriptValidation` has no height guard → **out of scope here, real gap worth tracking**
The proto fields (`skip_script_validation` = 13, `outpoint_only_spend` = 14) and the `Client`/`Server`/`Validator` wiring are all in #1178 on `main` — none are in this branch's four-file diff. I agree with the substance: unlike its height-guarded sibling `OutpointOnlySpend`, a wire-reachable `skipScriptValidation=true` has no height/provenance guard, so this is a genuine defence-in-depth gap. I'll track it separately against `main` (either a guard symmetrical to `OutpointOnlySpend`, or explicitly documenting the `Validate` gRPC/HTTP endpoint as trusted-internal-only) rather than folding it into this PR.

### ChiR3 — skip leaned on the non-unconditional `CheckMerkleRoot` → **fixed in `2715f0659`**
`Block.Valid` now tracks `merkleRootChecked` (set only where `CheckMerkleRoot` succeeds, inside the `subtreeStore != nil && len(b.Subtrees) > 0` block) and **requires it as a precondition of taking the skip**. A nil-subtreeStore / no-subtree caller can therefore never skip `validOrderAndBlessed` without the body↔header binding — the skip is refused and the full check runs instead. Also corrected the docstring wording that called `CheckMerkleRoot` "unconditional".

New regression guard `TestBlock_Valid_SkipRefusedWhenMerkleRootNotChecked`: with the merkle block not run (`merkleRootChecked == false`) but every policy conjunct satisfied, it proves `validOrderAndBlessed` genuinely runs (observed via `oldBlockIDsMap` being populated). Confirmed RED without the precondition. The "skip taken once the merkle root *has* been checked" path stays covered by `TestBlock_MerkleFloor_RejectsTamperingWhileSkipIsActive` (Case B).

### ChiR4 — height-0 gate divergence → **clarified in `2715f0659`**
Kept the model gate's `b.Height > 0`: genesis carries only a coinbase, so `validOrderAndBlessed` is trivial there and cheap to run in full, and excluding it matches the sibling fee skip's own `if b.Height == 0 { return nil }`. Added a comment explaining the rationale and noting the legacy/quick gates' height-0 inclusion is immaterial (genesis has no spends to order), per your "one-line comment on why genesis is treated as it is".

PR #1230, opened 2026-07-03
chore(sonar): knock out low-risk major code smells

liam:
## Summary

Low-risk maintainability quick wins across several rules. No product-logic behaviour changes.

| Rule | What | Count |
|---|---|---|
| exclusion | `.claude/**` excluded from analysis — Claude Code skill/diagnostic scripts, not product code (same rationale as `ui/**`, `docs/**`) | ~36 shell smells |
| `go:S1871` | merge identical switch cases — `help.go` → `case reflect.Struct, reflect.Map`; `settings_handler.go` drops the redundant `"category"` case (`default` already handles it identically) | 2 |
| `go:S108` | annotate the intentional channel-drain loops (`for range ch {}`) | 4 |
| `go:S1134` | reword a comment so the literal `FIXME` token (describing Bitcoin ABC history, not our TODO) no longer trips the rule | 1 |
| `shelldre:S7679` | assign `$2`/`$3` output-file args to named locals in the bestblock/lastblocks helpers and `chaos_heal` | 10 |

Also removed a stray bare-URL line in `bestblock-teranet.sh` that would error as a command on every call (copy-paste artifact).

## Deliberately skipped (not low-risk)

- **`multinode.sh` arg-parser `case "$1"` loops** (`S7679`, ~6) — idiomatic argument parsing; the refactor adds noise and risks `shift` bugs for a MINOR smell.
- **`script_utils_test.go` 107-branch opcode→name switch** (`S1479`) — a genuine refactor (map dispatch), not a quick win.

## Verification

```
bash -n on all edited shell scripts        # ok
go build / go vet (edited packages)        # ok
go test ./services/rpc/bsvjson/ ./services/asset/httpimpl/   # ok
gofmt -l (edited Go files)                 # clean
```

## Related (not in this PR)

Raising the `go:S107` "too many parameters" threshold 7→10 is a separate quality-profile change (clears ~37 of 45), pending maintainer sign-off since it edits the shared org profile.

### DIFF
--- a/compose/multinode.sh
+++ b/compose/multinode.sh
@@ -537,12 +537,13 @@ chaos_isolate() {
 }
 
 chaos_heal() {
-  if [[ -n "${1:-}" ]]; then
+  local node_num="${1:-}"
+  if [[ -n "$node_num" ]]; then
     local ctr
-    ctr=$(container_name "$1")
-    echo "restoring teranode$1 peer traffic..."
+    ctr=$(container_name "$node_num")
+    echo "restoring teranode$node_num peer traffic..."
     netns_iptables "$ctr" -F 2>/dev/null || true
-    echo "teranode$1 healed"
+    echo "teranode$node_num healed"
     return
   fi
   echo "restoring all nodes..."
--- a/scripts/bestblock-dev.sh
+++ b/scripts/bestblock-dev.sh
@@ -2,20 +2,23 @@
 
 function get_block_header() {
   local node=$1
+  local output_file=$2
   local url="http://$node/api/v1/bestblockheader/json"
-  curl -s "$url" | jq -r '. | "\(.height): \(.hash)"' > "$2" # Write output to temp file
+  curl -s "$url" | jq -r '. | "\(.height): \(.hash)"' > "$output_file" # Write output to temp file
 }
 
 function get_fsm_state() {
   local node=$1
+  local output_file=$2
   local url="http://$node/api/v1/fsm/state"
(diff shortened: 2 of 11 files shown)


freemans13 (approved):
Pre-PR review clean. Verified: bash -n on all 4 shell scripts OK; gofmt clean and go vet no-issues on edited Go packages; arg-position renames match all call sites; the removed bare-URL line in bestblock-teranet.sh was a stray copy-paste that executed as a command each call; both switch-case merges are behavior-preserving. All changes align with the commit scope. LGTM.

ordishs (approved):
Approving — clean cosmetic Sonar sweep, no product-logic changes. Verified the two hunks that carry any behavior:

- **`settings_handler.go`** — the removed `case "category"` block was logic-identical to `default` (compare category, tie-break by key), so falling through to `default` sorts identically. Behavior-preserving. ✅
- **Shell arg renames** (`bestblock-dev.sh`, `bestblock-teranet.sh`, `lastblocks.sh`, `multinode.sh`) — all call sites still pass positional args, and the named locals are assigned before any command that could clobber `$1/$2/$3`. No behavior change. ✅
- Removing the stray bare-URL line in `bestblock-teranet.sh` is a genuine fix (it executed as a command on every call).

Non-blocking (pre-existing, out of scope): `bestblock-dev.sh` uses unquoted `$(cat $tmp0)` / `echo $output` in several places — worth a future quoting pass, but untouched by this PR.

LGTM.

PR #1233, opened 2026-07-06
refactor: reduce cognitive complexity in two functions (S3776)

liam:
## Summary

Two behaviour-preserving extractions to bring functions under the cognitive-complexity threshold (25). No logic changes.

| Function | Before | After |
|---|---|---|
| `rpc.handleGetBlockHeader` | 29 | **3** |
| `rpc.getBlockHeaderVerboseResult` (new) | — | **14** |
| `profiling.parseSmaps` | 27 | **16** |
| `profiling.parseRegionStat` (new) | — | **1** |

(measured with `gocognit`)

## Changes

- **`handleGetBlockHeader`**: invert to a guard clause for the non-verbose path (`if !*c.Verbose { return hex }`) and extract the verbose-reply construction into a `getBlockHeaderVerboseResult` method. The main handler is now a trivial dispatch.
- **`parseSmaps`**: extract the 10-branch smaps statistic parser into `parseRegionStat`, rewritten as a `switch`. Field assignments and the running virtual/rss/pss totals are unchanged.

The large line count in `handlers.go` is the verbose block moving into the new method (indentation), not logic churn.

## Verification

```
gofmt -l           # clean
go build/vet ./services/rpc/ ./internal/profiling/   # ok
go test ./services/rpc/                              # ok
gocognit: all four functions < 25
```

### DIFF
--- a/internal/profiling/memory_analyzer.go
+++ b/internal/profiling/memory_analyzer.go
@@ -120,42 +120,8 @@ func parseSmaps(breakdown *MemoryBreakdown) error {
 			continue
 		}
 
-		// Parse memory statistics for current region
-		var value int64
-		if strings.HasPrefix(line, "Size:") {
-			_, _ = fmt.Sscanf(line, "Size: %d kB", &value)
-			currentRegion.Size = value * 1024
-			breakdown.TotalVirtual += value * 1024
-		} else if strings.HasPrefix(line, "Rss:") {
-			_, _ = fmt.Sscanf(line, "Rss: %d kB", &value)
-			currentRegion.Rss = value * 1024
-			breakdown.TotalRss += value * 1024
-		} else if strings.HasPrefix(line, "Pss:") {
-			_, _ = fmt.Sscanf(line, "Pss: %d kB", &value)
-			currentRegion.Pss = value * 1024
-			breakdown.TotalPss += value * 1024
-		} else if strings.HasPrefix(line, "Shared_Clean:") {
-			_, _ = fmt.Sscanf(line, "Shared_Clean: %d kB", &value)
-			currentRegion.SharedClean = value * 1024
-		} else if strings.HasPrefix(line, "Shared_Dirty:") {
-			_, _ = fmt.Sscanf(line, "Shared_Dirty: %d kB", &value)
-			currentRegion.SharedDirty = value * 1024
-		} else if strings.HasPrefix(line, "Private_Clean:") {
-			_, _ = fmt.Sscanf(line, "Private_Clean: %d kB", &value)
-			currentRegion.PrivateClean = value * 1024
-		} else if strings.HasPrefix(line, "Private_Dirty:") {
-			_, _ = fmt.Sscanf(line, "Private_Dirty: %d kB", &value)
-			currentRegion.PrivateDirty = value * 1024
-		} else if strings.HasPrefix(line, "Referenced:") {
-			_, _ = fmt.Sscanf(line, "Referenced: %d kB", &value)
-			currentRegion.Referenced = value * 1024
-		} else if strings.HasPrefix(line, "Anonymous:") {
-			_, _ = fmt.Sscanf(line, "Anonymous: %d kB", &value)
-			currentRegion.Anonymous = value * 1024
-		} else if strings.HasPrefix(line, "Swap:") {
-			_, _ = fmt.Sscanf(line, "Swap: %d kB", &value)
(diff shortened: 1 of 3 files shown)


ordishs (approved):
Compared the old and new code paths statement-by-statement — the extraction is faithful, no behaviour drift. Approving, with a few comments (only the first is worth fixing before merge):

1. **`internal/profiling/memory_analyzer_test.go`** — the new tests use testify `assert` for nearly every check, but our convention is `require` (`.claude/rules/testing.md`, `docs/references/codingConventions.md:326`). If an early check fails (e.g. `region.Size`), `assert` cascades a wall of follow-on failures instead of stopping at the root cause. Please switch to `require.Equal` and drop the `assert` import.

2. **`internal/profiling/memory_analyzer.go:138`** — `parseRegionStat` keeps 10 copy-paste cases that each repeat the stat name twice (the `HasPrefix` key and the `Sscanf` format). Since the block was rewritten anyway, a table-driven form — or `strings.Cut` on `:` + `strconv.ParseInt` — would remove the duplication and the per-line reflective `Sscanf` cost. A future prefix/format mismatch compiles fine and silently parses zero. Optional here, fine as follow-up.

3. **`services/rpc/handlers.go:366`** — the method ends with `if !isOnMainChain { headerReply.Confirmations = -1; return headerReply, nil }` immediately followed by the identical `return headerReply, nil`. Both branches return the same value, so this can collapse to a plain assignment and a single return.

4. **`services/rpc/handlers.go:274`** — `getBlockHeaderVerboseResult` returns `interface{}`, which was only ever an artifact of the handler-table signature. Returning the concrete `*bsvjson.GetBlockHeaderVerboseResult` keeps the contract compile-checked; `handleGetBlockHeader` can still return it as `interface{}`.

Follow-up material (pre-existing, moved verbatim, not blockers):

5. `parseRegionStat` duplicates the smaps parsing chain in `cmd/memanalyzer/main.go:111-131` — two parsers to keep in sync.
6. The verbose builder carries four verbatim twins of logic in `blockToJSON` (median-time fallback, genesis prev-hash blanking, next-hash lookup, confirmations formula), and the twins already diverge subtly (`:409` propagates non-`ErrBlockNotFound` errors while `:322` swallows all). Worth consolidating in a separate PR. If that line gets touched, `blockToJSON:460` already uses the overflow-safe `1 + int64(...) - int64(...)` confirmations form — matching it would be a free consistency fix.

liam:
Thanks for the careful statement-by-statement pass. Addressed in `0f130eff7`:

1. **`require` vs `assert`** ✅ Fixed — switched all checks to `require.Equal` and dropped the `assert` import. (Good catch, and I should've caught my own convention slip here.)

3. **Duplicated `return headerReply, nil`** ✅ Fixed — collapsed to `if !isOnMainChain { headerReply.Confirmations = -1 }` + a single return.

4. **`interface{}` return type** ✅ Fixed — `getBlockHeaderVerboseResult` now returns the concrete `*bsvjson.GetBlockHeaderVerboseResult`; `handleGetBlockHeader` still returns it as `interface{}` via passthrough. Verified the error-path typed-nil is a non-issue: the dispatcher (`Server.go:1015`/`1043`) returns `result` alongside a non-nil `err` and never marshals it on error, and a typed-nil pointer marshals to `null` identically anyway.

2. **Table-driven `parseRegionStat`** — agree it's the better shape; taking it as a follow-up as you suggested. One wrinkle to flag for whoever picks it up: `Size`/`Rss`/`Pss` also feed the `breakdown.Total*` accumulators while the other seven don't, so a naive `prefix → *field` table doesn't capture it uniformly — the follow-up should either special-case those three or carry an optional accumulator per row. `strings.Cut(line, ":")` + `strconv.ParseInt` to kill the reflective `Sscanf` is a clean win regardless.

**Follow-up material** (noted, not in this PR):
- `parseRegionStat` ↔ `cmd/memanalyzer/main.go:111-131` duplication — worth unifying when the table-driven version lands.
- The four verbatim twins in `blockToJSON` (median-time fallback, genesis prev-hash, next-hash, confirmations) and the noted divergence (`:409` propagates non-`ErrBlockNotFound` vs `:322` swallowing all), plus matching the overflow-safe confirmations form at `:460`. Agree this deserves its own consolidation PR rather than being smuggled into this one.

`gocognit` still 3 / 14 for the two functions; `go build`/`vet`/`test` green on both packages.

freemans13 (approved):

PR #1235, opened 2026-07-06
MvP-1140 Add Go-side satoshi-range/no-inflation backstop on SkipScriptValidati…

### LINKED ISSUE #1140: [TN-001][medium] validator: no Go-side money/inflation backstop; MaxSatoshis is dead code
**Severity:** medium · **Confidence:** high · **Pass:** consensus
**Component:** `services/validator` (`TxValidator.go`, `Validator.go`)

### Summary
The "no inflation" / satoshi-range invariants (sum(outputs) ≤ sum(inputs); 0 ≤ value ≤ 21M BSV) are enforced only inside GoBDK. The named constant `MaxSatoshis` is dead code. On the `SkipScriptValidation` path BDK is skipped entirely, so there is no value/inflation check at all there.

### Evidence
- `services/validator/Validator.go:48,52` — `MaxSatoshis = 21_000_000_00_000_000`; referenced only at its own definition (repo-wide grep), never in a comparison.
- `services/validator/TxValidator.go:151-167` — when `SkipScriptValidation` set, returns after only an unconfirmed-parent guard (and only if `SkipPolicyChecks`); BDK never called.
- `TxValidator.go:174` — normal path delegates everything to `bdk.ValidateTransaction`.
- Mitigation: comment at `TxValidator.go:147-150` requires the caller to use this only for checkpoint/PoW-trusted sync.

### Why it matters
No Go-side defense-in-depth for the most important consensus invariant. A future caller setting `SkipScriptValidation` on an untrusted path would accept coins-from-nothing with zero backstop. The dead `MaxSatoshis` misleads readers into thinking Teranode enforces this itself.

### Suggested fix
Add an overflow-safe Go-side check (`sum(outputs) ≤ sum(inputs)`, each output and total within `[0, MaxSatoshis]`) that runs even when `SkipScriptValidation` is true; wire in `MaxSatoshis` or delete it.

### Suggested tests
Validate a tx with `SkipScriptValidation=true,SkipPolicyChecks=true` whose outputs exceed inputs, and one >21M BSV (incl. via overflow) → assert rejection.

---

ctnguyen:
This resolve https://github.com/bsv-blockchain/teranode/issues/1140

## Problem

The "no inflation" and satoshi-range invariants — `sum(outputs) ≤ sum(inputs)` and every output/total within `[0, MaxSatoshis]` — were enforced **only inside GoBDK**. Two consequences:

- On the `SkipScriptValidation` path, `TxValidator.ValidateTransaction` returns after only an unconfirmed-parent guard and **never calls BDK**, so there was *no* value or inflation check at all on that path. A future caller setting `SkipScriptValidation` on an untrusted path would accept coins-from-nothing with zero Go-side backstop.
- The `MaxSatoshis` constant was dead code (referenced only at its own definition), which misleads readers into thinking Teranode enforces the supply cap itself.

Today the only production setter of the flag is checkpoint/PoW-trusted catchup sync, so this is a defense-in-depth gap rather than a live exploit — but the gap is real and the misleading constant should be resolved.

## Solution

Add an overflow-safe, Go-side value/inflation backstop that runs on the `SkipScriptValidation` path (where BDK is bypassed), designed to be a strict **subset of BDK's checks** so it can never reject a transaction that BDK/legacy bitcoin-sv would accept:

- `checkOutputValues` — per output: `> math.MaxInt64` → `bad-txns-vout-negative` (matching BDK reading values into a signed `Amount`), then `> MaxSatoshis` → `bad-txns-vout-toolarge`; running total `> MaxSatoshis` → `bad-txns-txouttotal-toolarge`.
- `checkInputValuesAndInflation` — per input/total range → `bad-txns-inputvalues-outofrange`; `sum(inputs) < sum(outputs)` → `bad-txns-in-belowout`.
- **Reject strings, boundary conditions, and check ordering mirror BDK exactly.** The skip-path ordering is era-aware to match BDK's control flow for a transaction that is both money-invalid and carries an unconfirmed-parent sentinel: outputs first; pre-Genesis sentinel → inflation; post-Genesis inflation → sentinel.
- `MaxSatoshis` is now **wired in** as the money bound (rather than deleted).
- The path continues to rely on the documented extended-transaction precondition and **fails closed** — no fail-open branch for value-less transactions.

## Tests

11 new unit tests in `TxValidator_test.go`, including:

- Inflation, output-negative (high-bit), output-over-max, output-total-over-max, and input-over-max rejections, each asserting the exact reject string.
- Zero-fee / zero-value outputs accepted; non-extended input fails closed.
- Policy-mode (`SkipPolicyChecks=false`) rejection, proving the backstop is not consensus-branch-only.
- Mixed-defect ordering tests on both sides of the Genesis boundary.
- A parity table that runs each mutated transaction through **both** the skip path and the real GoBDK path and asserts both return the same marker — a genuine drift guard.

## Verification

`go build ./...`, `go vet`, and `golangci-lint run` on the validator package are clean; all validator unit tests pass (the one skipped failure is a pre-existing Aerospike/Docker integration test unrelated to this change).

### DIFF
--- a/services/validator/TxValidator.go
+++ b/services/validator/TxValidator.go
@@ -157,7 +157,8 @@ func (tv *TxValidator) ValidateTransaction(tx *bt.Tx, blockHeight uint32, utxoHe
 		// ERA-DEPENDENT order that mirrors BDK: pre-Genesis the sentinel first,
 		// post-Genesis the input/inflation check first. See the era switch below
 		// for why (bdk/core/txvalidator.cpp).
-		if err := tv.checkOutputValues(tx); err != nil {
+		totalOut, err := tv.checkOutputValues(tx)
+		if err != nil {
 			return err
 		}
 
@@ -190,11 +191,11 @@ func (tv *TxValidator) ValidateTransaction(tx *bt.Tx, blockHeight uint32, utxoHe
 			if err := sentinelGuard(); err != nil {
 				return err
 			}
-			if err := tv.checkInputValuesAndInflation(tx); err != nil {
+			if err := tv.checkInputValuesAndInflation(tx, totalOut); err != nil {
 				return err
 			}
 		} else {
-			if err := tv.checkInputValuesAndInflation(tx); err != nil {
+			if err := tv.checkInputValuesAndInflation(tx, totalOut); err != nil {
 				return err
 			}
 			if err := sentinelGuard(); err != nil {
@@ -220,31 +221,32 @@ func (tv *TxValidator) ValidateTransaction(tx *bt.Tx, blockHeight uint32, utxoHe
 // check. The running total is bounded after each addition, which also makes the
 // accumulation overflow-safe: every addend is within [0, MaxSatoshis], so the
 // total cannot wrap uint64 before it is rejected.
-func (tv *TxValidator) checkOutputValues(tx *bt.Tx) error {
+func (tv *TxValidator) checkOutputValues(tx *bt.Tx) (uint64, error) {
 	totalOut := uint64(0)
 
 	for _, out := range tx.Outputs {
 		if out.Satoshis > math.MaxInt64 {
-			return errors.NewTxInvalidError("bad-txns-vout-negative")
+			return 0, errors.NewTxInvalidError("bad-txns-vout-negative")
 		}
 
 		if out.Satoshis > MaxSatoshis {
-			return errors.NewTxInvalidError("bad-txns-vout-toolarge")
+			return 0, errors.NewTxInvalidError("bad-txns-vout-toolarge")
 		}
 
 		totalOut += out.Satoshis
 		if totalOut > MaxSatoshis {
-			return errors.NewTxInvalidError("bad-txns-txouttotal-toolarge")
+			return 0, errors.NewTxInvalidError("bad-txns-txouttotal-toolarge")
 		}
 	}
 
-	return nil
+	return totalOut, nil
 }
 
 // checkInputValuesAndInflation enforces the input money-range and no-inflation
 // invariants that BDK applies in implCheckInputValues: each input value and the
 // running total must be within [0, MaxSatoshis], and the sum of inputs must be at
-// least the sum of outputs (equality is allowed, i.e. a zero-fee transaction).
+// least totalOut (the output total already computed and bounded by
+// checkOutputValues; equality is allowed, i.e. a zero-fee transaction).
 // A high-bit or over-range input value both map to bad-txns-inputvalues-outofrange,
 // matching how the money-range predicate collapses the two cases.
 //
@@ -252,7 +254,7 @@ func (tv *TxValidator) checkOutputValues(tx *bt.Tx) error {
 // this holds for every caller that reaches ValidateTransaction. If a value-less
 // transaction were passed, the sum of inputs would be zero and the check fails
 // closed (bad-txns-in-belowout) rather than admitting coins from nothing.
-func (tv *TxValidator) checkInputValuesAndInflation(tx *bt.Tx) error {
+func (tv *TxValidator) checkInputValuesAndInflation(tx *bt.Tx, totalOut uint64) error {
 	totalIn := uint64(0)
 
 	for _, in := range tx.Inputs {
@@ -264,11 +266,6 @@ func (tv *TxValidator) checkInputValuesAndInflation(tx *bt.Tx) error {
 		if totalIn > MaxSatoshis {
 			return errors.NewTxInvalidError("bad-txns-inputvalues-outofrange")
 		}
-	}
-
-	totalOut := uint64(0)
-	for _, out := range tx.Outputs {
-		totalOut += out.Satoshis
 	}
 
 	if totalIn < totalOut {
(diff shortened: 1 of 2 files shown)


ordishs (approved):
Approving — the backstop is consensus-sound. The checks are a faithful, reject-string-exact strict subset of bitcoin-sv, the strict-subset property holds (nothing BDK would accept is rejected here), and the extended-tx precondition is enforced structurally: on the only production path that sets `SkipScriptValidation` (`netsync/handle_block.go` → `PreValidateTransactions`), `extendTransactions` runs unconditionally and fails closed, so a non-extended tx can never reach `checkInputValuesAndInflation` and be wrongly rejected as `bad-txns-in-belowout` during checkpoint sync. Test coverage — including the BDK parity table — is solid.

One cheap follow-up I'd like to see in this PR:

**Make the inflation check overflow-safe by construction, not by call order.**

`checkInputValuesAndInflation` re-sums `tx.Outputs` (TxValidator.go:267-270) with no per-output or running-total bound, unlike bitcoin-sv's `GetValueOut()`. It's safe today *only* because the sole caller runs `checkOutputValues` immediately before, which caps each output and the total to `MaxSatoshis`. If this method is ever reused independently, a tx with outputs near `math.MaxUint64` would wrap `totalOut` to a small value, making `totalIn < totalOut` false and silently admitting inflation.

Suggested fix: have `checkOutputValues` return `(uint64, error)` and pass the already-computed `totalOut` into the inflation check. That converts "correct by call-order convention" into "correct by construction" and also removes the redundant second walk over `tx.Outputs`.

Non-blocking nit: the era-aware sentinel ordering (blockHeight < GenesisActivationHeight branch) only selects which error string a doubly-invalid tx returns — both branches reject. Fine to keep for BDK string parity, but worth a comment noting the branch buys error-marker parity only, not any accept/reject difference.

icellan (approved):
LGTM. Clean, well-tested, correctly-scoped defense-in-depth. The backstop is a genuine strict subset of BDK: coinbase is excluded upstream at `TxValidator.go:139` before the branch, the accumulation is overflow-safe (`MaxSatoshis=2.1e15 ≪ MaxInt64`), and I confirmed the skip-path can't false-reject valid catchup data — stored txs are written via `ExtendedBytes()`, which serializes 8 bytes of `PreviousTxSatoshis` per input unconditionally, so the `extendPerTxFallback` recovery copy always carries the real value; the only `totalIn<totalOut` trigger for a valid tx is all-zero inputs, whose outputs also sum to zero (accepted). Perf on the hot catchup path is a non-issue: the `sentinelGuard` closure inlines (no per-tx heap alloc), so it's just cheap integer loops.

One nit worth fixing before merge, plus optional cleanups:

- **Fix the header comment (`TxValidator.go:152-157`).** It describes a single fixed order (outputs → sentinel → inputs/inflation), but post-Genesis the code runs inputs/inflation *before* the sentinel — the actual era-dependent behavior is only explained 25 lines down. As written it's misleading and could invite an incorrect reorder later.
- **(optional, perf) Outputs are summed twice (`TxValidator.go:268`).** `checkOutputValues` computes a running `totalOut` and discards it; `checkInputValuesAndInflation` re-sums the same outputs. Have `checkOutputValues` return the total and reuse it — free, and it's exactly the hot path this PR targets.
- **(optional) Era-dependent ordering (`:187`)** only changes which reject string a doubly-invalid tx returns, never accept/reject. Fine to keep for BDK parity — just don't grow the complexity, and fix the comment above.

Not blocking. Nice work on the drift-guard parity test.

ctnguyen:
Thanks @icellan for the latest reviews. New commits :

✅ **P1/P3** — header comment rewritten to state era-dependent order (`docs` commit `550e5ee`).
✅ **P2** — output total computed once and threaded through; duplicate sum removed, semantics/reject-strings unchanged (`perf` commit `459178338`).
✅ **Verified** — `go vet`, `make lint` (0 issues), and the `SkipScriptValidation_*` + `TestSkipPath_MoneyChecks_MatchBDKPath` tests all pass (`chore` verify commit `66c28fa`).
✅ **Self-Reviewed** — independent review pass on the diff: no defects; no behaviour change.

oskarszoon (approved):
Reviewed as consensus-critical. Traced the obvious chain-split risk — coinbase reaching the new inflation check — and it can't happen: coinbase is rejected at `TxValidator.go:139` before the skip block and is excluded from the validated txMap upstream. uint64 accumulation is overflow-safe (each addend bounded before the add, total re-checked per iteration). Verified the era-dependent sentinel/inflation ordering against `gobdk v1.2.5` `txvalidator.cpp` — pre-Genesis sentinel-first, post-Genesis inflation-first matches exactly. The backstop is a strict subset in the accept direction, so it can't reject anything BDK/legacy would accept. LGTM.

Two nits, neither blocking:

- "strict subset of BDK's checks" reads as a full structural replica, but it only covers the money-range/inflation sub-checks — not `bad-txns-vin-empty` / `bad-txns-vout-empty`. A zero-output burn tx passes the backstop (it can't reach the path — a block containing one fails the pinned checkpoint-hash match). Worth tightening the wording.
- No accept-side boundary parity case for `input == MaxSatoshis` (only the output boundary is pinned). Guards a hypothetical future `>`→`>=` edit on the input check; not practically testable via fixture mutation since mutating amounts breaks `aTx`'s sighash. Skip unless cheap.

PR #1236, opened 2026-07-06
refactor(asset/httpimpl): extract handler bodies to reduce cognitive complexity (S3776)

liam:
## Summary

The four asset `Get*` handlers each wrap their entire body in a returned `func(c echo.Context) error` closure. Because everything sits inside that lambda, every branch carries a nesting penalty that pushed cognitive complexity over the threshold (25). This moves each closure body into a private method and has the closure just delegate.

| Handler | Extracted method | complexity after |
|---|---|---|
| `GetBlockByHash` | `getBlockByHash` | 17 |
| `GetTransaction` | `getTransaction` | 19 |
| `GetBlockSubtrees` | `getBlockSubtrees` | 19 |
| `GetUTXOsByTxID` | `getUTXOsByTxID` | 19 |

(the outer handlers become trivial; measured with `gocognit`)

## No behaviour or performance change

Pure structural extraction — a **whitespace-ignoring diff (`git diff -w`) shows only the added delegating call + method signature per file; the moved bodies are byte-identical**. Control flow, error paths, tracing spans, the `errgroup` fan-out in `GetUTXOsByTxID` (captured vars and all) are unchanged. The only runtime cost is one extra function call per request on these HTTP read-path handlers — negligible, and off the tx/block pipeline. Public handler signatures and routes are untouched.

## Verification

```
gofmt -l (4 files)                       # clean
go vet  ./services/asset/httpimpl/        # clean
go test ./services/asset/httpimpl/        # ok
gocognit: all four extracted methods < 25; outer handlers trivial
```

### DIFF
--- a/services/asset/httpimpl/GetBlock.go
+++ b/services/asset/httpimpl/GetBlock.go
@@ -288,76 +288,81 @@ func (h *HTTP) GetBlockByHeight(mode ReadMode) func(c echo.Context) error {
 //	GET /block/hash/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f/hex
 func (h *HTTP) GetBlockByHash(mode ReadMode) func(c echo.Context) error {
 	return func(c echo.Context) error {
-		hashStr := c.Param("hash")
+		return h.getBlockByHash(c, mode)
+	}
+}
 
-		ctx, _, deferFn := tracing.Tracer("asset").Start(c.Request().Context(), "GetBlockByHash_http",
-			tracing.WithParentStat(AssetStat),
-			tracing.WithDebugLogMessage(h.logger, "[Asset_http] GetBlockByHash in %s for %s: %s", mode, c.RealIP(), hashStr),
-		)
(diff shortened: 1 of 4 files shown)


freemans13 (approved):
Reviewed as a pure structural refactor. Verified mechanically:

- `git diff -w` confirms each file adds only a delegating closure + new method signature; the moved bodies are byte-identical.
- Closures captured only `c` and `mode` (a value type), both now passed explicitly — no captured local state, so no semantic change. The `errgroup` fan-out in `GetUTXOsByTxID` captures only method-local vars — safe.
- `go vet ./services/asset/httpimpl/` and `gofmt -l` on all four files: clean.
- Exported handlers keep their full godoc; routes and public signatures unchanged, so no callers are affected.

Complexity reduced with zero behavioural change. LGTM.

ordishs (approved):
Pure structural refactor — verified the four extracted handler bodies are byte-identical (whitespace-ignoring diff shows only the delegating call + method signature per file). Control flow, error paths, tracing spans, and the errgroup fan-out in getUTXOsByTxID are unchanged. No name collisions, package builds clean. LGTM.

Issue #1237, opened 2026-07-07
counter-conflicting: point-lookup deleted_children markers instead of full-set read


oskarszoon:
## Summary

The counter-conflicting `deletedChildren` marker read materializes a parent's **entire** marker set into a map on every consult, but both consumers only test membership of a single spender. Convert to a targeted `(parent_hash, child_hash)` point lookup.

Surfaced by self-review of #1217; non-blocking (bounded, transient, cold path) — tracked here as a follow-up per that PR's disposition.

## Where

- SQL read: `stores/utxo/sql/sql.go` (~1674) — `SELECT child_hash FROM deleted_children WHERE parent_hash = $1` (no `child_hash` filter), loaded into `meta.Data.DeletedChildren`.
- Aerospike read: `stores/utxo/aerospike/get.go` `mergePageDeletedChildren` (~1382) — unions every page record's `deletedChildren` map for the parent.
- Consumers (both test exactly one spender):
  - `stores/utxo/process_conflicting.go:~1351` `discriminateGhostSpender` — `deletedChildren[*spender]`
  - `stores/utxo/process_conflicting.go:~1050` `collectDanglingWinnerInputSpends` ghost path — `parentMeta.DeletedChildren[spender]`

## Why it's only P2

Growth is **bounded**, not unbounded: the `deleted_children` PK `(parent_hash, child_hash)` dedups, so the set size is capped by the parent's mined-then-pruned child count. It is reaped when the parent record itself is pruned (the walk then fails closed on the missing parent anyway) — the "not cleaned while the parent survives" lifecycle is by design. The only real cost is a transient tens-to-hundreds-of-MB map built and freed on one rare `Get`, for a long-lived high-fanout parent (exchange/faucet splitter kept alive by unspent outputs) touched by a counter-conflicting walk. No correctness, safety, or consensus property is affected.

## Proposed fix

Add a targeted marker-existence check to the UTXO Store rather than fetching the whole set:

- New method, e.g. `IsChildDeleted(ctx, parent, child *chainhash.Hash) (bool, error)` (native types per the Interface.go convention).
  - SQL: `SELECT 1 FROM deleted_children WHERE parent_hash = $1 AND child_hash = $2`.
  - Aerospike: check the single key in the parent's (page-aggregated) `deletedChildren` bin without building the full Go map.
[...]

Issue #1214, opened 2026-07-02
utxo: tx record deleted while parent SpendingDatas still reference it — audit all delete paths for dangling spender refs


oskarszoon:
> **Status (2026-07-08) — PRIMARY / sole track.** The read-side recovery approach (#1213; PRs #1217, #1239) has been **closed**: the affected mainnet instance was reset, so surviving-a-ghost is no longer needed and we're not carrying that fix through more review. This issue — **preventing the dangling ref from forming at the write side** — is the durable fix and the only track going forward. To be worked in a dedicated session.

## Summary

A tx record was deleted from the UTXO store while a mined parent's `SpendingDatas` still referenced it as the spender, leaving a dangling ref. Downstream, every reader that walks that ref (`GetCounterConflictingTxHashes`, `GetConflictingChildren`, `ProcessConflicting`) hits `TX_NOT_FOUND`. On one mainnet node ("eu-1", v0.15.2) this wedged catchup permanently — see #1213 for the wedge itself. This issue tracks the write side: **what deleted the record without cleaning up the parents, and how do we make every delete path safe.**

## Evidence

- Parent `dd856380fa467f6c0b4d3543ec758cf2bcb14924d44d0bd61b109cca11d8a5ea:4` (mined, record present): `SpendingDatas[4] → fbdebded4b373806c6b6883b5b16c91d7d6e6ea77aa56e077f74b165b4a0dabb`
- `fbdebded…`: record absent from the store (`TX_NOT_FOUND`), while the network chain mines its rival `1f8e4dcc…` at height 956015 — i.e. `fbdebded…` is a never-mined first-seen loser whose record got reaped, but its parent spends were never unspent and (if it was marked conflicting) parent `conflictingChildren` refs were not removed.
- Invariant that should hold: if tx T's record is deleted, no surviving record may still reference T (parent `SpendingDatas`, parent `conflictingChildren`).

## Suspects (to investigate)

1. **`SetConflicting` stamps DAH on the losing tx** (`stores/utxo/aerospike/conflicting.go` — lua `setConflicting` is called with current height + `GetUtxoStoreBlockHeightRetention()`), then the pruner reaps the record at DAH. If conflict resolution (`ProcessConflicting` step 2 Unspend) never ran for that loser — crash window (#861, WAL added in #1126), reorg partial (#845), or the rival's block never processed — the parent spend stays pointing at the reaped record.
[...]

PR #1239, opened 2026-07-07
fix(utxo): tolerate dangling spender refs in counter-conflicting walk

### LINKED ISSUE #1213: subtreevalidation: dangling parent SpendingData → TX_NOT_FOUND in counter-conflicting walk wedges catchup forever
## Summary

A dangling parent `SpendingData` reference (parent output marked spent by a tx whose record no longer exists in the UTXO store) makes the counter-conflicting walk fail with `TX_NOT_FOUND`. Subtree validation treats that as a hard `ProcessingError`, the block fails, catchup retries the same block forever. No exit condition, no metric — the node wedges permanently.

Hit on a mainnet node ("eu-1", v0.15.2, EKS, Aerospike pruned mode), stuck at height 956014 with target 956074.

## Error chain

```
block processing failed: SUBTREE_ERROR (29): failed to check subtree
-> PROCESSING (4): [CheckSubtree] Failed to validate legacy subtree 694fa1dd35acb415711927339769e959d260e9848556029d97ce954bdcee3953
-> CONTEXT_CANCELED (7): [processTxMetaUsingStore]
-> PROCESSING (4): [processTxMetaUsingStore][1f8e4dccade679021a6bd00ef2b301523243d0f8d90cb9fe496a43498bbcf7fa] failed to check counter conflicting tx on current chain
-> PROCESSING (4): [checkCounterConflictingOnCurrentChain][1f8e4dcc…] failed to get counter conflicting tx hashes
-> TX_NOT_FOUND (30): fbdebded4b373806c6b6883b5b16c91d7d6e6ea77aa56e077f74b165b4a0dabb not found
```

Same ghost hash (`fbdebded…`) also wedges a second tx in the same subtree (`6b02e27b…`) via the p2p catchup path (`CheckBlockSubtreesRequest`).

## Root cause

Store state: mined parent `dd856380…:4` has `SpendingDatas[4] → fbdebded…`, but `fbdebded…`'s record was deleted (see companion write-side issue). Incoming block 956015 contains `1f8e4dcc…`, which also spends `dd856380…:4`:

1. `processTxMetaUsingStore.go:174` — tx meta has `Conflicting=true` → `checkCounterConflictingOnCurrentChain`
2. `SubtreeValidation.go:470` → `utxo.GetCounterConflictingTxHashes`
3. `process_conflicting.go:1122` — walks parent's `SpendingDatas`, calls `GetConflictingChildren(fbdebded…)` → `s.Get` → aerospike `ErrKeyNotFound` → `TX_NOT_FOUND` (`aerospike/get.go:628`)
4. Error propagates unwrapped (`process_conflicting.go:1074/1093/1124`); `checkCounterConflictingOnCurrentChain` converts any error into a hard failure → subtree fails → block fails → retry loop, forever.

## Why this is unrecoverable in the field today

- The retry has no bound and no escalation (same family as #980).
- `cmd/rewindblockchain` cannot repair it: Phase 1 iterates only unmined + conflicting records; the dangling ref lives on a **mined** parent, and the ghost's own record is already gone, so neither iterator ever visits it. Verified live: a full rewind run (9,299 unmined + 36 conflicting purged, 49 parent refs cleaned, Phase 4 verify green) did not clear the error — the incoming block re-creates the conflicting tx and re-derives the same walk on the next cycle.

## Proposed fix (read side)

Absent ⇒ no BlockIDs ⇒ definitionally not mined on our chain ⇒ the check should pass, not error.

- `GetCounterConflictingTxHashes` (`process_conflicting.go:1067`): tolerate NOT_FOUND on the tx itself (return empty set), on a parent lookup (skip parent), and on the spender's `GetConflictingChildren` (exclude ghost from counter set). NOT_FOUND = `errors.Is(err, errors.ErrTxNotFound) || errors.Is(err, errors.ErrNotFound)` (aerospike vs SQL, same pairing as `cmd/rewindblockchain/rewindblockchain/tx_delete.go:17`).
- `GetConflictingChildren` BFS (`process_conflicting.go:991`): tolerate NOT_FOUND per node (result already nil-skipped).
- `checkCounterConflictingOnCurrentChain` (`SubtreeValidation.go:467`): tolerate NOT_FOUND on per-counter `GetMeta`, nil-guard the final BlockIDs loop.
- Warn log + prometheus counter on every tolerated ghost so store inconsistency is visible instead of silent.

Self-heal: the same walk feeds `ProcessConflicting`; when the block containing the rival is accepted, winner promotion re-spends the parent inputs and overwrites the dangling `SpendingData` — the store repairs itself on the forward path. Ghost losers are skipped (nothing to demote).

Consensus safety argument: a tx mined on our chain keeps its record until DAH (mined height + retention). A counter-tx with an absent record is either a never-mined purged loser (check must pass) or would imply the network chain double-spends an output spent ≥ retention blocks deep — a consensus-invalid block unreachable on the honest chain. Frozen-tx sentinels are distinct values and keep their current handling.

## Related

- #861 / #1126 (crash-safety WAL), #980 (unbounded retry, no escalation), #704, #765, #845
- Write-side cause (what deleted the record without unspending its parents) tracked separately — see companion issue.
- Follow-up candidates (separate): bounded retry/escalation for persistent block validation failure during catchup; `rewindblockchain` limitation note for this failure class.

oskarszoon:
Fixes #1213. Supersedes #1217 (same fix, rebuilt on a clean branch to drop the long review-iteration history). Write-side cause tracked in #1214.

## Problem

A dangling spender reference — a mined parent output whose `SpendingDatas` still records a tx whose own record was later deleted — makes `GetCounterConflictingTxHashes` fail with `TX_NOT_FOUND`. Subtree validation treats any error there as fatal, so legacy catchup retries the same block forever with no exit condition. Hit in production on a mainnet node (v0.15.2, Aerospike pruned mode): wedged at height 956014, target 956074, `blocks_validated: 0` across every attempt. A `rewindblockchain` run did not clear it — its Phase 1 iterates only unmined/conflicting records and never visits the mined parent holding the stale slot.

## Fix

Tolerate the dangling reference in the counter-conflicting read path instead of treating it as fatal, with a marker to keep the tolerance consensus-safe.

- **Tolerate a missing spender, fail closed on a missing parent.** A missing spender still leaves the surviving parent identifying the current spender, so the walk excludes it and continues. A missing *parent* erases the whole spend graph for that input, so the walk propagates the error (fail closed) rather than dereference a nil slot.
- **The `deletedChildren` marker discriminates a deliberately-reaped mined counter from a benign dangling ref.** The mined-on-chain gate compares counters against a `retention*2` (576-block) window, but a mined-spent counter is pruned at `retention` (288) — so in the (288, 576] band a genuinely-mined counter is already a ghost plain tolerance would wave through. The pruner records a `deletedChildren` marker when it reaps a child, and the marker carries a precise meaning: **marker present ⇒ the reaped spender was mined on our longest chain** ⇒ fail closed (`TxInvalidError`); marker absent ⇒ tolerate. The marker is written **only for mined-on-chain reaped children** on both backends (aerospike: `unminedSince` clear + non-empty block heights; SQL: `unmined_since IS NULL` and a `block_ids` row — the same mined-on-longest-chain test the store's own delete-at-height gate uses), so a conflicting **never-mined** loser reaped at its DAH is intentionally left unmarked and its ghost is tolerated. That mined-only gating is what makes the marked-ghost `TxInvalidError` sound: it can only reject a block spending a counter that really was mined here.
- **Marker reliability, both backends.** Aerospike writes it unconditionally for mined children (no cuckoo-filter pre-filter skip) and page-keyed (bounded per page record); the counter-conflicting walk reads it page-aggregating (unions the master and every page record's map). SQL freezes the deletable set in a temp table so no mined row is deleted unmarked under READ COMMITTED. A corrupt/torn marker fails the walk **closed** on both backends. A blob-missing external mined tx that cannot be marked is deferred, then GC'd marker-less only once past `DAH + max(retention, CoinbaseMaturity)` — outside every validatable block's window, so it can never be an in-window unmarked ghost. The SQL prune pass is bounded and drained across passes so a large first pass cannot time the marker INSERT out into a wedge.
- **TOCTOU fail-closed.** `checkCounterConflictingOnCurrentChain` fails closed on all `GetMeta` errors: the walk saw the counter alive moments earlier, so a NOT_FOUND at re-read is a racing delete, and skipping it would skip the mined-on-our-chain rejection itself. On retry the walk re-runs and applies the marker discriminator.
- **Frozen sentinels** stay visible in the counter set and BFS results (so the subtree-path placeholder guard rejects a frozen counter) but are never existence-checked or descended into, and `MarkConflictingRecursively` filters them before `SetConflicting`.

### Residual (documented, accepted)

A missing marker for a mined counter is historical/transient only: pre-fix aerospike deletes the cuckoo pre-filter suppressed, and SQL stores upgraded mid-life whose `deleted_children` table starts empty (one 576-block window after upgrade, then fresh pruning covers the live band). All are backstopped by the primary Spend double-spend defense — the surviving parent still records the spender in its `(parent,vout)` slot, so a double-spender is rejected as `ErrSpent` independent of the pruned counter's record and of the marker — and require an attacker fork inside the 576-block window. Closing the upgraded-store corner has no clean form (an epoch/low-water-mark fail-closed would re-wedge the legitimate never-added-ghost case this PR fixes). The window constants are deliberately untouched.

## Deliberately out of scope

- **ProcessConflicting self-heal.** An earlier iteration also repaired a dangling winner-input slot inside `ProcessConflicting` (block-assembly conflict resolution). That path never wedged in the incident (it is not on the subtree-validation/catchup flow), a ghost there fails step-3 `Spend` closed with `ErrSpent` and rolls back cleanly, and that behaviour matches pre-PR baseline — so it is left out to keep this change focused on the observed wedge. Not a regression.
- **The write-side cause** (what deletes a tx record while a parent still references it) is #1214.
- **Point-lookup of `deleted_children`** instead of the full-set read is #1237 (bounded, cold-path).

## Tests

Read-side tolerance (missing spender, fail-closed-on-missing-parent, BFS ghost eviction), marker discriminator (marked→`TxInvalidError`, unmarked→tolerated, `(nil,nil)` and corrupt-marker shapes), frozen-sentinel guards, GetMeta TOCTOU fail-closed, pruner mined-only marker gate (mined→marker, never-mined→no marker) on both backends, page-aggregating read, bounded prune + GC-on-horizon, and the SQL/aerospike marker round-trips (unit + Aerospike container).

- `go test -race ./stores/utxo/ ./stores/utxo/aerospike/pruner/` — ok
- `go test ./stores/utxo/sql/ ./stores/utxo/sql/pruner/ ./services/blockassembly/` — ok
- `go test -tags testtxmetacache ./services/subtreevalidation/` — ok
- `go build ./...`, `go vet`, `golangci-lint run --new-from-rev` — clean

## Config note

`pruner_utxoPrunedSetMaxEntries` is removed. It sized the in-memory pruned-tx set whose cuckoo false-positives could suppress a `deletedChildren` marker; the pruner no longer uses that set, so the marker is written reliably. A stale value left in config is silently ignored — remove it.

## Operational note

Recovery for a wedged node: cherry-pick, deploy, `setfsmstate RUNNING` — the node blesses past the ghost and proceeds. No reseed, no store surgery.


ctnguyen (changes requested):
Review at commit [1d3bafc](https://github.com/bsv-blockchain/teranode/commit/1d3bafcaaeae413977c57dd0523523518779610f)

This is a hard, consensus-critical problem attacked with real rigour, and the marker design is correct in the common case (a mempool-only counter → both teranode and legacy accept). The fail-closed-on-missing-parent / tolerate-missing-spender asymmetry is well-argued, the marker is written before the child delete on both backends (aerospike ordered `flushCleanupBatches`; SQL single-txn temp-table freeze) so a mined counter is not reaped unmarked on the primary paths, and corrupt/torn markers fail the read *closed* everywhere. The SQL `SetConflicting` read-before-write-txn split is a legitimate deadlock fix, the bounded-and-drained prune passes address a real timeout-wedge, and dropping the cuckoo pruned-set is a clean simplification that also improves marker reliability. The frozen-sentinel handling — which I initially suspected — turns out to be consensus-*correct*: the all-0xFF sentinel doubles as teranode's frozen-TXO marker (get.go records it as the spender of a `Status_FROZEN` output), and legacy likewise rejects spends of frozen outputs (`CFrozenTXOCheck`, `bitcoin-sv src/frozentxo.h`), so the hard-error on a frozen child is right. Test coverage is broad and the inline documentation is exceptional — it made this review far easier.

## Blocking issues

### ChiR1 — SQL `deletedChildren` marker is not a longest-chain test → teranode rejects a block legacy accepts (consensus split)
**Problem:** The design rests on *"marker present ⟹ the reaped spender was mined on our longest chain,"* and the comments assert both backends apply *"the same mined-on-longest-chain test."* That is false for SQL. Aerospike is correctly gated (mine UDF clears `unminedSince` only under `onLongestChain`, [teranode.lua:639](stores/utxo/aerospike/teranode.lua#L639)). But the SQL marker predicate is `tx.unmined_since IS NULL AND EXISTS(block_ids row)`, and SQL `Create` sets `unmined_since = NULL` for *any* tx carrying a `MinedBlockInfo`, ignoring `OnLongestChain` ([sql.go:466-468](stores/utxo/sql/sql.go#L466-L468)), inserting a `block_ids` row regardless ([sql.go:792](stores/utxo/sql/sql.go#L792)) — and both production callers leave `OnLongestChain` at its `false` default ([quick_validate.go:1261](services/blockvalidation/quick_validate.go#L1261), [handle_block.go:1018](services/legacy/netsync/handle_block.go#L1018)). A tx mined only into a **side-fork** block therefore satisfies the SQL predicate; nothing restores `unmined_since` for a fork block that never became the tip (only a `moveBack` reorg does). If such a tx is a conflicting loser, `SetConflicting` stamps a DAH with no mined gate ([sql.go:4317](stores/utxo/sql/sql.go#L4317)), the pruner reaps it and **writes a marker**, and the walk's `discriminateGhostSpender` returns `TxInvalidError` → the block is rejected ([check_block_subtrees.go:1196](services/subtreevalidation/check_block_subtrees.go#L1196)).
**Why it matters:** Legacy never applies a side-fork block's spends to the active UTXO view (`UpdateCoins`/`SpendCoin` run only on active-chain connect, `bitcoin-sv src/validation.cpp:2480-2495, 3440`), so that output stays unspent and legacy **accepts** the block (`HaveInputs` true, coins.cpp:16-20). teranode rejecting it is a node forking off the honest chain — reachable by a natural orphan/fork race on the SQL backend, no attacker required, and with no backstop (this is fail-*closed*).
**Fix:** Gate the SQL marker on genuine longest-chain membership, matching aerospike's `onLongestChain` semantics — e.g. require the `block_ids` row to reference a block on the current main chain rather than merely to exist:

```sql
-- insertMarkersQuery: replace  EXISTS (SELECT 1 FROM block_ids b WHERE b.transaction_id = tx.id)
-- with a main-chain-qualified membership (join block_ids to the active-chain block set),
-- so a fork-only-mined tx is left UNMARKED exactly as aerospike leaves it.
```
Also correct the "both backends use the same longest-chain test" comments, which are currently inaccurate.

### ChiR2 — Marker-less GC of a mined counter → teranode accepts a block legacy rejects (consensus split); the stated safety proof is unsound
**Problem:** The marker-less GC of a blob-gone external tx (`errExternalInputsUnrecoverable` branch, [pruner_service.go](stores/utxo/aerospike/pruner/pruner_service.go)) is justified by *"block validation caps fork depth at CoinbaseMaturity, so a GC'd record is outside every validatable block's window."* But `validateForkDepth` runs **only** on the catchup path ([catchup.go:189](services/blockvalidation/catchup.go#L189), reached only when a block's parent is missing, [Server.go:1461](services/blockvalidation/Server.go#L1461)). A deep fork delivered block-by-block (parent present) on the normal `ValidateBlockWithOptions` path or the legacy netsync path has no fork-depth cap, so a block can be subtree-validated far enough below the pruner tip that its `retention×2` window reaches a marker-less GC'd counter. That counter is then an unmarked ghost → the walk tolerates it → `checkCounterConflictingOnCurrentChain` returns nil → the tx is blessed as conflicting and the block is **accepted**.
**Why it matters:** If that counter was genuinely mined+spent on the *active* chain, legacy rejects the block wholesale (`bad-txns-inputs-missingorspent`, `bitcoin-sv src/validation.cpp:3328-3332`, whole-block `DoS(100)` → `InvalidBlockFound`). teranode accepting it is the worst split direction — accepting an invalid block. Note the `ErrSpent` backstop does **not** save this: `blessMissingTransaction` downgrades `ErrSpent`→conflicting and the block still proceeds ([SubtreeValidation.go blessMissingTransaction](services/subtreevalidation/SubtreeValidation.go)); only a `TxInvalidError` from the counter-conflicting check rejects a block. Reachability is narrow/adversarial (a fork block validated deeper than ~`retention` below tip on a non-catchup path, an *external* counter whose blob is fully GC'd, that counter mined+spent on the active chain) — but the severity when reached is a hard split, and the PR's own soundness proof is demonstrably false on the paths that reach the walk.
**Fix:** Either enforce the fork-depth cap on the normal/legacy validation path (so the horizon proof actually holds), or demonstrate that side-fork blocks below the cap are never subtree-validated/blessed, and re-ground the horizon rationale accordingly (it cannot rest on a cap the reaching paths bypass).

## Non-blocking issues

### ChiR3 — Missing-parent fail-closed relocates the wedge; write-side cause (#&NoBreak;1214) unfixed
**Problem:** The walk now fails closed on a missing *parent* record, returning a `ProcessingError` (not `ErrTxInvalid`), which lands in the indefinite-retry class ([check_block_subtrees.go:1196](services/subtreevalidation/check_block_subtrees.go#L1196)). The same write-side corruption that produces a dangling *spender* (#&NoBreak;1214, out of scope) can produce a dangling *parent*, re-creating the #&NoBreak;1213 wedge in a new spot.
**Why it matters:** A liveness/parity gap (legacy always renders a binary verdict), not a verdict split — but the PR does not fully close the wedge *class* it targets while #&NoBreak;1214 is unfixed.
**Fix:** No code change required for this PR; state the tolerate-spender / fail-on-parent asymmetry as a known residual and consider a bounded-retry-then-decision for the parent case.

### ChiR4 — Third NOT_FOUND-tolerance site in the walk bypasses the `deletedChildren` discriminator
**Problem:** In `GetCounterConflictingTxHashes`, the `GetConflictingChildren` error path tolerates NOT_FOUND unconditionally (bump counter, `delete`, `continue`) *without* calling `discriminateGhostSpender`, unlike the two sibling sites. Unreachable for both shipped backends today (the shared `GetConflictingChildren` no longer propagates a bare NOT_FOUND), but the comment frames it as a fallback for other `Store` implementations — for which it is a latent fail-open.
**Why it matters:** Inconsistent with the fail-closed philosophy of the rest of the function; a decorator/future backend that does propagate NOT_FOUND would silently tolerate a marked (mined) ghost.
**Fix:** Route this site through `discriminateGhostSpender(txHash, spendingTxID, parentInfo.deletedChildren)` too, so all three ghost paths apply the identical rule.

### ChiR5 — `errExternalInputsUnrecoverable` matched by `==` identity is fragile
**Problem:** `processRecordChunk` compares `err == errExternalInputsUnrecoverable` ([pruner_service.go](stores/utxo/aerospike/pruner/pruner_service.go)). Correct today (returned unwrapped), but every other error in that function is wrapped with `errors.NewProcessingError(...)`; any future change that wraps this one would silently break the identity match and route blob-missing externals into the hard-failure path instead of defer-and-GC — quietly re-wedging the pruner.
**Why it matters:** Latent fragility in a codebase that otherwise standardises on `errors.Is`/error-code matching.
**Fix:** Match via `errors.Is` against a stable sentinel error rather than raw `==`.

### ChiR6 — Ghost eviction in `GetConflictingChildren` can re-fetch and double-count the same missing node
**Problem:** `delete(visited, currentLevel[i])` un-visits a ghost mid-BFS ([process_conflicting.go](stores/utxo/process_conflicting.go)). If another node references the same ghost, it is re-added, re-fetched (NOT_FOUND again), and the `dangling_refs{site="bfs"}` counter is incremented again. Bounded by fan-in (not an infinite loop) and the final set is still correct, but it is redundant work and a metric-accuracy defect, with mild RPC-amplification potential on an adversarially-shaped conflict graph.
**Why it matters:** Efficiency and metric fidelity, not correctness.
**Fix:** Track evicted ghosts in a separate "known-missing" set consulted by the visited-check, so each missing node is fetched and counted once.

### ChiR7 — Documentation drift and a dead redundant check
**Problem:** A few comments no longer match the code: (1) the "both backends use the same mined-on-longest-chain test" claim (`meta/data.go`, pruner, `discriminateGhostSpender`) is inaccurate per ChiR1; (2) [meta/data.go:114](stores/utxo/meta/data.go#L114) still describes the blob-missing GC horizon as `DAH + 2*retention`, whereas the code computes `DAH + retention + max(retention, CoinbaseMaturity)`; (3) `errors.IsNotFound`'s comment says the SQL stores surface `ErrNotFound` for a missing tx, but both backends' `Get` actually return `ErrTxNotFound` (checking both is harmlessly inclusive). Separately, the frozen-descendant check in `checkCounterConflictingOnCurrentChain` ([SubtreeValidation.go:502-514](services/subtreevalidation/SubtreeValidation.go#L502-L514)) is now dead code — `GetCounterConflictingTxHashes` already hard-errors on the identical condition first (pre-existing, not introduced here).
**Why it matters:** Misleading rationale in consensus-critical code costs the next maintainer real time; the dead branch obscures intent.
**Fix:** Correct the three comments to match behaviour, and remove or document the unreachable frozen-descendant check.

## Recap

| ChiR1 | SQL marker not longest-chain gated → reject block legacy accepts | YES | 90% |
| ChiR2 | GC marker-less → accept block legacy rejects; proof unsound | YES | 72% |
| ChiR3 | Missing-parent fail-closed relocates the wedge |  | 40% |
| ChiR4 | Third NOT_FOUND site bypasses marker discriminator |  | 30% |
| ChiR5 | Sentinel error matched by `==` identity, fragile |  | 20% |
| ChiR6 | Ghost eviction re-fetch / metric double-count |  | 15% |
| ChiR7 | Documentation drift + dead redundant check |  | 10% |

oskarszoon:
Closing. The affected mainnet instance was reset, so the read-side recovery fix isn't needed — and I'd rather not carry it through further review rounds. The durable fix is **prevention**: stop the dangling spender ref from forming at the write side, tracked in #1214. Closing the recovery symptom #1213 alongside this.

PR #1240, opened 2026-07-07
MvP-1141 Enforce coinbase scriptSig length in Block.Valid

### LINKED ISSUE #1141: [TN-002][low] model/blockvalidation: coinbase scriptSig length check not centralized in Block.Valid
**Severity:** low · **Confidence:** high · **Pass:** consensus
**Component:** `model/Block.go`, `services/blockvalidation/BlockValidation.go`

### Summary
The consensus rule that a coinbase scriptSig is 2..`MaxCoinbaseScriptSigSize` bytes is enforced only in `ValidateBlockWithOptions`, not inside `model/Block.Valid()` where the other structural coinbase checks live. The main peer-block path **is** covered (verified), so this is a robustness/centralization gap, not an active bypass.

### Evidence
- `model/Block.go:474-503` — `Valid()` checks coinbase presence + `IsCoinbase()` + BIP34 height; no scriptSig length check.
- `services/blockvalidation/BlockValidation.go:1325` — only `< 2 || > MaxCoinbaseScriptSigSize` check; enclosing func `ValidateBlockWithOptions` (`:1238`), runs before the `block.Valid` call at `:1642`.
- Other `block.Valid` callers without an independent length check: `:1544`, `:2028` (`reValidateBlock`), `services/blockassembly/Server.go:1580` (self-built), `services/blockvalidation/Server.go:1321`.

### Why it matters
A future caller validating an external block via `Block.Valid` without the external guard would accept an out-of-range coinbase scriptSig. Latent today.

### Suggested fix
Move the length check into `model/Block.Valid` next to the existing coinbase/BIP34 checks.

### Suggested tests
`model/Block_test.go`: coinbase scriptSig of 1 byte and one > Max → assert `Valid` rejects directly.

---

ctnguyen:
Closes https://github.com/bsv-blockchain/teranode/issues/1141

## Summary

The consensus rule that a coinbase transaction's scriptSig must be between 2 and
`MaxCoinbaseScriptSigSize` bytes (inclusive) was enforced only in the block-validation
and block-assembly services, not in `model/Block.Valid()` where the other structural
coinbase checks live. This PR centralizes the check into `Block.Valid`, closing a latent
gap on the code paths that reach `Block.Valid` without an independent length guard.

## The gap

`Block.Valid()` already checks coinbase presence, `IsCoinbase()`, and the BIP-34 height,
but not the scriptSig length. The length check existed in two other places:

- `services/blockvalidation/BlockValidation.go` — `ValidateBlockWithOptions`
- `services/blockassembly/Server.go` — `submitMiningSolution`

Both run their check *before* calling `Block.Valid`, so the main peer-block and
self-built-block paths were covered. But two callers reached `Block.Valid` without any
independent length guard:

- `BlockValidation.reValidateBlock`
- the gRPC `blockvalidation.Server.ValidateBlock` entry point

A future or existing caller validating an external block through `Block.Valid` without the
service-layer guard would accept an out-of-range coinbase scriptSig. Latent today, but a
consensus-relevant robustness hole.

## The change

Add the coinbase scriptSig length check inside `Block.Valid`, immediately after the
`IsCoinbase()` check and before the BIP-34 height block:

- **Inclusive bounds**, matching bitcoin-sv `CheckCoinbase` (`bad-cb-length`): reject iff
  `size < 2 || size > MaxCoinbaseScriptSigSize`. `size == 2` and `size == Max` both pass.
- The bound is read from **per-network chain config**
  (`settings.ChainCfgParams.MaxCoinbaseScriptSigSize`), never hard-coded.
- **Nil-safe**: a nil unlocking script is treated as length 0 and rejected (an empty
  scriptSig fails the lower bound). `Inputs[0]` is safe to index because `IsCoinbase()`
  above guarantees exactly one input.
- Error class is `BlockInvalidError` (hard-invalid), matching the reference node's
  `REJECT_INVALID` / DoS-100 semantics for this rule.

### Why the existing checks are kept (defense-in-depth, not removed)

The service-layer checks are intentionally retained rather than removed, because they
carry behavior `Block.Valid` neither has nor should reproduce:

- `ValidateBlockWithOptions` distinguishes a *missing/empty* coinbase
  (`BlockIncompleteError`, so the block can be re-fetched from another peer) from a
  *bad-length* coinbase (`BlockInvalidError` + `storeInvalidBlock`).
- `submitMiningSolution` returns a `ProcessingError` for the node's own mining-solution
  submission — a different error class appropriate to that RPC.

Removing them would change error-class and side-effect behavior on consensus-relevant
paths. Once the check is centralized in `Block.Valid`, those checks become genuine
defense-in-depth (they short-circuit earlier with richer semantics) with **zero
behavioral change** to their paths, while the new check closes the gap on the
revalidation and gRPC `ValidateBlock` paths.

## Consensus parity

This matches the reference bitcoin-sv node's `CheckCoinbase` rule
(`MAX_COINBASE_SCRIPTSIG_SIZE = 100`, `bad-cb-length`, DoS 100). Teranode's configured
`MaxCoinbaseScriptSigSize` is 100 on every network, so the enforced bound is identical.
No coinbase the reference node would accept is now rejected (no off-by-one; existing
in-range fixtures continue to pass).

## Tests

Added `TestBlock_Valid_CoinbaseScriptSigLength` in `model/Block_test.go`, covering:

- scriptSig length 1 (too short) → rejected
- scriptSig length Max+1 (too long) → rejected
- scriptSig length exactly 2 (lower boundary) → passes
- scriptSig length exactly Max (upper boundary) → passes
- nil unlocking script → rejected

The passing boundary cases use a minimal coinbase-only block setup so unrelated validation
(header difficulty, BIP-34, subtree, reward) does not interfere.

## Files changed

- `model/Block.go` — the centralized length check
- `model/Block_test.go` — the new test cases

No changes to `services/blockvalidation/*` or `services/blockassembly/*`.

## Verification

- `go build ./model/...` — pass
- `go vet ./model/...` — pass
- `go test ./model/...` — pass (full package)
- `go test -race ./model/` — pass
- `golangci-lint run` (changed files vs `main`) — 0 issues

### DIFF
--- a/model/Block.go
+++ b/model/Block.go
@@ -530,6 +530,19 @@ func (b *Block) Valid(ctx context.Context, logger ulogger.Logger, subtreeStore S
 		return false, errors.NewBlockInvalidError("[BLOCK][%s] block coinbase tx is not a valid coinbase tx", b.String())
 	}
 
+	// 4b. Check that the coinbase scriptSig (unlocking script) length is within consensus bounds.
+	// Parity with bitcoin-sv CheckCoinbase (bad-cb-length): inclusive 2 <= size <= MaxCoinbaseScriptSigSize.
+	// IsCoinbase() above guarantees exactly one input, so Inputs[0] is safe to index; a nil
+	// UnlockingScript is treated as length 0 and fails the lower bound, matching an empty scriptSig.
+	scriptSigLen := 0
+	if us := b.CoinbaseTx.Inputs[0].UnlockingScript; us != nil {
+		scriptSigLen = len(*us)
+	}
+
+	if scriptSigLen < 2 || scriptSigLen > int(settings.ChainCfgParams.MaxCoinbaseScriptSigSize) {
+		return false, errors.NewBlockInvalidError("[BLOCK][%s] bad coinbase length", b.String())
+	}
+
 	// We can only calculate the height from coinbase transactions in block versions 2 and higher
 
 	// https://en.bitcoin.it/wiki/BIP_0034
--- a/model/Block_test.go
+++ b/model/Block_test.go
@@ -438,6 +438,88 @@ func TestBlock_Valid_ComprehensiveCoverage(t *testing.T) {
 	})
 }
 
+// TestBlock_Valid_CoinbaseScriptSigLength pins the coinbase scriptSig length check in Block.Valid.
+// Parity with bitcoin-sv CheckCoinbase (bad-cb-length): valid iff 2 <= size <= MaxCoinbaseScriptSigSize,
+// inclusive. See GitHub issue #1141.
+func TestBlock_Valid_CoinbaseScriptSigLength(t *testing.T) {
+	tSettings := test.CreateBaseTestSettings(t)
+	maxLen := int(tSettings.ChainCfgParams.MaxCoinbaseScriptSigSize)
+
+	// buildBlock returns a coinbase-only, otherwise-valid block whose coinbase scriptSig is
+	// either a controlled byte length (setNil == false) or a nil UnlockingScript (setNil == true).
+	// Height 0 skips BIP-34 and the reward/fee checks; nil subtreeStore/txMetaStore skip the
+	// subtree and order/blessed checks; empty currentChain skips the median-time-past check.
+	buildBlock := func(t *testing.T, n int, setNil bool) *Block {
+		t.Helper()
+
+		blockHeaderBytes, err := hex.DecodeString(block1Header)
+		require.NoError(t, err)
+
+		blockHeader, err := NewBlockHeaderFromBytes(blockHeaderBytes)
+		require.NoError(t, err)
+
+		coinbase, err := bt.NewTxFromString(CoinbaseHex)
+		require.NoError(t, err)
+
+		if setNil {
+			coinbase.Inputs[0].UnlockingScript = nil
+		} else {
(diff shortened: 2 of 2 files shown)


icellan (approved):

freemans13 (approved):
Reviewed the diff (13 lines in `model/Block.go`, 82 lines of tests). Clean, well-scoped, consensus-safe.

Verified:
- `Inputs[0]` is safe — `IsCoinbase()` in go-bt v2.6.7 requires `len(Inputs)==1`.
- Bound `< 2 || > int(MaxCoinbaseScriptSigSize)` is byte-identical to the existing checks in `BlockValidation.go:1429` and `blockassembly/Server.go:1464`; error class `BlockInvalidError` and message `bad coinbase length` match.
- `MaxCoinbaseScriptSigSize` is `uint32`, default 100 on every network, so `int()` is overflow-safe.
- `bad-cb-length` is an always-on `CheckBlock` rule in bitcoin-sv, so no historical block on any network is newly rejected — safe on the revalidate and gRPC `ValidateBlock` paths this closes the gap on.
- New nil-safe handling is an improvement over the retained service-layer checks (which call `.Bytes()` on a possibly-nil script).

Locally: `go vet ./model/` clean, `go test ./model/` 449 pass (incl. the new 6 subtests), `gofmt` clean.

Nice defense-in-depth fix — LGTM.

PR #1241, opened 2026-07-07
fix(ci): harden gitleaks workflow (template injection + HTTPS enforcement)

liam:
## Summary

Two security hardenings to the secret-scan workflow added in #1238:

1. **`githubactions:S6506` (HTTPS enforcement)** — the `curl -sSfL` install used `-L` (follow redirects) without pinning the protocol, so a redirect could downgrade to HTTP. Added `--proto '=https' --proto-redir '=https' --tlsv1.2` so both the initial request and any redirect must stay HTTPS. *This is the finding currently failing the SonarCloud gate.*

2. **Template injection** — the scan step interpolated `${{ github.event.pull_request.*.sha }}` directly into a `run:` shell. Moved to the `env:` context and referenced as shell vars (`${BASE_SHA}..${HEAD_SHA}`), so GitHub never substitutes context text into the script.

## Verification

- YAML valid.
- No `${{ github.event.* }}` interpolation inside any `run:` block (only in `env:`).
- Download restricted to HTTPS on request + redirect.
- Scan semantics unchanged (same `base..head` range, pinned gitleaks 8.18.4).

Once merged and re-scanned, the security rating returns to A and the gate goes green.

### DIFF
--- a/.github/workflows/gitleaks.yaml
+++ b/.github/workflows/gitleaks.yaml
@@ -28,7 +28,8 @@ jobs:
         run: |
           set -euo pipefail
           mkdir -p "${RUNNER_TEMP}/bin"
-          curl -sSfL "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_arm64.tar.gz" \
+          curl -sSfL --proto '=https' --proto-redir '=https' --tlsv1.2 \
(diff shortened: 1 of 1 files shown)


ctnguyen (approved):

ordishs (approved):
Correct and minimal. curl protocol pinning (`--proto '=https' --proto-redir '=https' --tlsv1.2`) resolves S6506 without breaking the GitHub release redirect, and moving the PR SHAs into `env:` follows the canonical Actions hardening pattern. Scan semantics unchanged.

Non-blocking follow-up: consider checksum-pinning the gitleaks tarball (download to file + `sha256sum -c`) since the version pin alone doesn't verify integrity.

PR #1242, opened 2026-07-08
fix(rpc): bound legacy ban RPC calls with a timeout (#591)

liam:
## Problem

The ban RPC handlers (`isbanned` / `listbanned` / `clearbanned` / `setban`) call the legacy peer client whenever it is non-nil. That client is **lazily dialed** and is non-nil whenever `legacy_grpcAddress` is configured, so on a scaling/testnet node without a running legacy service every ban RPC issued a gRPC call that stalled - `handleIsBanned` and `handleClearBanned` didn't even wrap it in a timeout, so they blocked on the parent context (#591).

## Approach (revised)

An earlier version of this PR gated the legacy calls on `isSubscriberActive(SubscriberLegacy)`, mirroring `handleGetInfo`. @oskarszoon correctly pointed out that `"legacy/manager"` has had **no `Subscribe()` producer since #735**, so that gate is *always false* - it disabled legacy banning entirely rather than skipping it only when inactive. That approach was dropped.

Instead: keep the `legacyP2PClient != nil` check and wrap each legacy ban call (`isbanned`, `clearbanned`, both `setban` arms) in a `ClientCallTimeout`-bounded context (`listbanned` already had one). An absent-but-configured legacy service can no longer stall the RPC, and legacy banning still works when legacy is present.

## Also in this PR

- **`setban` reports failure when no ban leg applied it.** Previously `success` only tracked the p2p leg and was only checked inside the legacy block, so a failed p2p ban with no active legacy silently returned success. `success` is now hoisted and checked after the switch.
- **Unban wording.** The failure message branches on the command, so a failed `setban remove` returns "Failed to remove ban", not "Failed to apply ban".

## Contract change (note)

`setban` is now strict: a call where no applicable leg applied the ban returns an error where it previously could return silent success (e.g. p2p-only failure, or no clients configured). No in-repo caller depends on the old no-op; this is the correct behaviour for an administrative anti-abuse control.

## Tests

- `TestHandleSetBan_SuccessTracking` - legacy-only success carries the result (load-bearing for the `success = true` on the legacy branches), and the no-clients path returns the command-appropriate failure, for both `add` and `remove`.
- `TestBanHandlers_LegacyCallBounded` - a legacy mock that blocks until its context is cancelled; the handler returns promptly with the per-call timeout and hangs without it.
- `go test ./services/rpc/` passes; `go vet` clean.

Closes #591

### DIFF
--- a/services/rpc/handlers.go
+++ b/services/rpc/handlers.go
@@ -1604,15 +1604,13 @@ func calculateMedianTime(ctx context.Context, blockchainClient blockchain.Client
 	return medianTimestampUint32, nil
 }
 
-// isSubscriberActive checks whether a blockchain subscriber whose source
-// contains substr is currently registered. This lets RPC handlers skip
-// expensive calls to services that are not running.
 // isSubscriberActive checks whether source is present in the blockchain
 // subscriber list. This lets RPC handlers skip expensive calls to services
 // that are not running.
 func isSubscriberActive(ctx context.Context, s *RPCServer, source string) bool {
 	subs, err := s.blockchainClient.GetSubscribers(ctx)
 	if err != nil {
+		s.logger.Warnf("[isSubscriberActive] GetSubscribers failed, treating %s as inactive: %v", source, err)
 		return false
 	}
 	for _, src := range subs {
--- a/services/rpc/handlers.go
+++ b/services/rpc/handlers.go
@@ -1604,13 +1604,15 @@ func calculateMedianTime(ctx context.Context, blockchainClient blockchain.Client
 	return medianTimestampUint32, nil
 }
 
+// isSubscriberActive checks whether a blockchain subscriber whose source
+// contains substr is currently registered. This lets RPC handlers skip
+// expensive calls to services that are not running.
 // isSubscriberActive checks whether source is present in the blockchain
 // subscriber list. This lets RPC handlers skip expensive calls to services
 // that are not running.
 func isSubscriberActive(ctx context.Context, s *RPCServer, source string) bool {
 	subs, err := s.blockchainClient.GetSubscribers(ctx)
 	if err != nil {
-		s.logger.Warnf("[isSubscriberActive] GetSubscribers failed, treating %s as inactive: %v", source, err)
 		return false
 	}
 	for _, src := range subs {
@@ -2116,8 +2118,16 @@ func handleIsBanned(ctx context.Context, s *RPCServer, cmd interface{}, _ <-chan
 	// Legacy service only handles IPs
 	var peerBanned bool
 
-	if isIP && s.legacyP2PClient != nil && isSubscriberActive(ctx, s, blockchain.SubscriberLegacy) {
-		isBannedLegacy, err := s.legacyP2PClient.IsBanned(ctx, &peer_api.IsBannedRequest{IpOrSubnet: c.IPOrSubnet})
+	if isIP && s.legacyP2PClient != nil {
+		// Bound the call: the legacy peer client is lazily dialed and non-nil
+		// whenever legacy_grpcAddress is configured, so on a node without a
+		// running legacy service this would otherwise block on the parent
+		// context until the RPC's own deadline (see #591).
+		legacyCtx, cancel := context.WithTimeout(ctx, s.settings.RPC.ClientCallTimeout)
+		isBannedLegacy, err := s.legacyP2PClient.IsBanned(legacyCtx, &peer_api.IsBannedRequest{IpOrSubnet: c.IPOrSubnet})
+
+		cancel()
+
 		if err != nil {
 			s.logger.Warnf("Failed to check if banned in legacy peer service: %v", err)
 		} else {
@@ -2199,7 +2209,7 @@ func handleListBanned(ctx context.Context, s *RPCServer, cmd interface{}, _ <-ch
 	}
 
 	// check if legacy peer service is available
-	if s.legacyP2PClient != nil && isSubscriberActive(ctx, s, blockchain.SubscriberLegacy) {
+	if s.legacyP2PClient != nil {
 		// Create a timeout context for the legacy peer client call
 		legacyCtx, cancel := context.WithTimeout(ctx, clientCallTimeout)
 		defer cancel()
@@ -2276,8 +2286,14 @@ func handleClearBanned(ctx context.Context, s *RPCServer, cmd interface{}, _ <-c
 		}
 	}
 	// check if legacy peer service is available
-	if s.legacyP2PClient != nil && isSubscriberActive(ctx, s, blockchain.SubscriberLegacy) {
-		_, err := s.legacyP2PClient.ClearBanned(ctx, &emptypb.Empty{})
+	if s.legacyP2PClient != nil {
+		// Bound the call so an absent-but-configured legacy service can't stall
+		// the RPC on the parent context (#591).
+		legacyCtx, cancel := context.WithTimeout(ctx, s.settings.RPC.ClientCallTimeout)
+		_, err := s.legacyP2PClient.ClearBanned(legacyCtx, &emptypb.Empty{})
+
+		cancel()
+
 		if err != nil {
 			s.logger.Warnf("Failed to clear banned list in legacy peer service: %v", err)
 		}
@@ -2342,11 +2358,9 @@ func handleSetBan(ctx context.Context, s *RPCServer, cmd interface{}, _ <-chan s
 	}
 
 	// success tracks whether at least one applicable ban leg actually applied
-	// the ban. It is checked after the switch so the outcome no longer depends
-	// on which legs ran: previously a failed p2p ban was only surfaced as an
-	// error inside the legacy block, so when the legacy leg was skipped (client
-	// absent, or - after the subscriber gate - configured but inactive) a failed
-	// ban silently returned success.
+	// the ban; checked after the switch so the outcome no longer depends on
+	// which legs ran (previously a failed p2p ban was only surfaced when the
+	// legacy leg also ran and failed).
 	var success bool
 
 	// Handle the command
@@ -2383,30 +2397,22 @@ func handleSetBan(ctx context.Context, s *RPCServer, cmd interface{}, _ <-chan s
 		}
 
 		// and ban legacy peers
-		if s.legacyP2PClient != nil && isSubscriberActive(ctx, s, blockchain.SubscriberLegacy) {
-			until := expirationTimeInt64
+		if s.legacyP2PClient != nil {
+			// Bound the call so an absent-but-configured legacy service can't
+			// stall the RPC on the parent context (#591).
+			legacyCtx, cancel := context.WithTimeout(ctx, s.settings.RPC.ClientCallTimeout)
 
-			resp, err := s.legacyP2PClient.BanPeer(ctx, &peer_api.BanPeerRequest{
+			resp, err := s.legacyP2PClient.BanPeer(legacyCtx, &peer_api.BanPeerRequest{
 				Addr:  c.IPOrSubnet,
-				Until: until,
+				Until: expirationTimeInt64,
 			})
+
+			cancel()
 
 			if err != nil {
 				s.logger.Warnf("Error while trying to ban legacy peer: %v", err)
-
-				if !success {
-					return nil, &bsvjson.RPCError{
-						Code:    bsvjson.ErrRPCInvalidParameter,
-						Message: "Failed to add ban",
-					}
-				}
 			} else if resp == nil || !resp.Ok {
-				if !success {
-					return nil, &bsvjson.RPCError{
-						Code:    bsvjson.ErrRPCInvalidParameter,
-						Message: "Failed to ban peer",
-					}
-				}
+				s.logger.Warnf("Legacy peer service did not apply ban for %s", c.IPOrSubnet)
 			} else {
 				success = true
 				s.logger.Debugf("Added ban for %s until %v", c.IPOrSubnet, expirationTime)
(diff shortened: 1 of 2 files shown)


ordishs (approved):
Correct, minimal fix. Reusing the `isSubscriberActive` gate that `handleGetInfo` already uses is the right call — no new mechanism, no interface changes, and short-circuit ordering on the nil check is right. Verified locally on the branch: `go test ./services/rpc/` passes and `go vet` is clean.

A few non-blocking suggestions:

1. **Silent skip on `GetSubscribers` error.** `isSubscriberActive` swallows the error and returns `false`. For `getinfo` that's harmless, but for `setban add` it means a transient blockchain-client failure while legacy *is* running applies the ban to p2p and silently skips legacy — previously a failed legacy `BanPeer` at least logged a warning. A debug/info log inside `isSubscriberActive` on error would give operators a trail.
2. **Negative test coverage.** The "legacy not subscribed → client not called" regression subtest only covers `handleIsBanned`. A cheap table-driven negative test across `listbanned`, `clearbanned`, and both `setban` arms would guard against future drift, since all five sites share the same gate.
3. **Stale doc comment (pre-existing, adjacent).** `isSubscriberActive` carries a duplicated doc comment and the first copy describes "contains substr" semantics that don't match the exact-match implementation. Since this PR is all about that helper, deleting the stale block would be a fair drive-by.
4. **Helper placement nit.** `newLegacySubscriberMock` is inserted between the `TestHandleIsBannedComprehensive` doc comment and the function it documents, orphaning the comment — move the helper above it.

Note for a follow-up rather than this PR: the `isSubscriberActive` call (and the legacy calls in `handleIsBanned`/`handleClearBanned`, as the description notes) still run on the raw parent context with no timeout.

ctnguyen (changes requested):
Review at commit [59ac626](https://github.com/bsv-blockchain/teranode/commit/59ac626)

This is a focused, minimal fix that solves a real operational problem (ban RPCs stalling on an absent legacy service) by reusing the exact `isSubscriberActive` gate `handleGetInfo` already relies on — no new mechanism, no interface or signature churn. The short-circuit ordering is right (`legacyP2PClient != nil` before the `GetSubscribers` call, and `isIP &&` before both in `handleIsBanned`), so no wasted gRPC calls. The test wiring is thoughtful: existing comprehensive cases were updated to keep exercising the legacy path via a small `newLegacySubscriberMock` helper, and a genuine negative regression test was added.

## Blocking issues

### ChiR1 — `handleSetBan` now reports success when the only attempted ban leg fails
**Problem:** In `handleSetBan` (both the `add` and `remove` arms, [handlers.go:2347-2452](services/rpc/handlers.go#L2347-L2452)), the `success` flag set by the p2p leg is only ever inspected *inside* the `if s.legacyP2PClient != nil && isSubscriberActive(...)` block. The function otherwise falls through to `return nil, nil`. Before this PR, a node with `legacy_grpcAddress` configured but the service down would enter the legacy block, get an error, and — with `!success` — return `"Failed to add ban"`. After this PR, that same node (legacy configured but not an active subscriber — the exact scaling/testnet deployment this PR targets) skips the legacy block entirely, so a failed p2p `BanPeer` now returns `nil, nil`: the RPC reports success while no ban was applied.
**Why it matters:** `setban` is an administrative anti-abuse/anti-DoS control. Silently reporting success when the ban did not take effect leaves an operator believing a malicious peer is banned when it is not, with no error surfaced. The regression lands precisely in the deployment scenario the PR is meant to improve.
**Fix:** Check `success` after the `switch` so the outcome no longer depends on which legs were attempted, e.g.:
```go
	default:
		return nil, &bsvjson.RPCError{Code: bsvjson.ErrRPCInvalidParameter, Message: "Invalid command. Must be 'add' or 'remove'."}
	}

	if !success {
		return nil, &bsvjson.RPCError{Code: bsvjson.ErrRPCInvalidParameter, Message: "Failed to apply ban"}
	}
	return nil, nil
```
(The underlying "`success` never checked after the switch" bug is pre-existing for `legacyP2PClient == nil` nodes; this PR widens it to configured-but-inactive-legacy nodes, so it's worth closing here or in a tracked follow-up rather than shipping the regression silently.)

## Non-blocking issues

### ChiR2 — `isSubscriberActive` swallows `GetSubscribers` errors with no log
**Problem:** [handlers.go:1613-1624](services/rpc/handlers.go#L1613-L1624) returns `false` on a `GetSubscribers` error. Every other client-call failure in these handlers emits a `Warnf`; this one is silent.
**Why it matters:** A transient blockchain-client error makes ban RPCs skip the legacy leg with no trace — an operator cannot distinguish "legacy genuinely inactive" from "the subscriber lookup failed."
**Fix:** Log before returning:
```go
	subs, err := s.blockchainClient.GetSubscribers(ctx)
	if err != nil {
		s.logger.Warnf("[isSubscriberActive] GetSubscribers failed, treating %s as inactive: %v", source, err)
		return false
	}
```

### ChiR3 — Negative regression test covers only `handleIsBanned`
**Problem:** The new "legacy skipped when legacy not an active subscriber" subtest exists only for `handleIsBanned`. `handleListBanned`, `handleClearBanned`, and both `handleSetBan` arms got the positive wiring but no test proving the legacy call is skipped when legacy is inactive.
**Why it matters:** All five sites share an identical gate; a future copy-paste/typo drift in any of the other four would go undetected.
**Fix:** Add the same "legacy not called" assertion to the other comprehensive tests (a small table-driven helper over the five handlers would keep it compact).

### ChiR4 — Stale, duplicated doc comment on `isSubscriberActive`
**Problem:** [handlers.go:1607-1612](services/rpc/handlers.go#L1607-L1612) carries two doc-comment blocks; the first describes "source **contains substr**" semantics, but the implementation does exact equality (`src == source`).
**Why it matters:** The stale copy is misleading, and this PR is centred on that helper — a fair drive-by cleanup.
**Fix:** Delete the first three-line block, keeping only the accurate one.

### ChiR5 — Test helper orphans the `TestHandleIsBannedComprehensive` doc comment
**Problem:** `newLegacySubscriberMock` was inserted between the `// TestHandleIsBannedComprehensive tests the handleIsBanned handler` comment and the function it documents, with no blank line — so Go attaches that comment to the helper and the test function is left undocumented.
**Why it matters:** Cosmetic, but it misattributes documentation.
**Fix:** Move `newLegacySubscriberMock` (with its own comment) above the `// TestHandleIsBannedComprehensive...` line.

## Recap

| ChiR1 | setban reports success on failed ban | YES | 55% |
| ChiR2 | GetSubscribers error silently unlogged |  | 30% |
| ChiR3 | Negative test only for handleIsBanned |  | 22% |
| ChiR4 | Stale/duplicated doc comment |  | 12% |
| ChiR5 | Helper orphans test doc comment |  | 7% |

liam:
Thanks both. Addressed all five in two commits (`a33a817` for the non-blocking items, `de0cfe1` for ChiR1):

**ChiR1 (blocking) — setban reports success on a failed ban.** Fixed. `success` is now hoisted to the handler scope and set true on any successful leg (p2p or legacy, add and remove); after the switch, if no applicable leg applied the ban it returns `Failed to apply ban`. This also closes the pre-existing hole for `legacyP2PClient == nil` nodes. Note this makes setban strict where it was previously lenient, so I updated the `add ban p2p returns false` test to expect the error (it matched its own "Returns false when ban fails" comment) and added a regression test for the configured-but-inactive-legacy case. Confirmed both failure tests fail without the post-switch check.

**ChiR2 — silent GetSubscribers error.** `isSubscriberActive` now logs `Warnf("[isSubscriberActive] GetSubscribers failed, treating %s as inactive: %v", ...)` before returning false.

**ChiR3 — negative coverage.** Added `TestBanHandlers_LegacySkippedWhenInactive` covering listbanned, clearbanned, and both setban arms (isbanned was already covered), each asserting the legacy client is not consulted when legacy is not an active subscriber.

**ChiR4 — stale doc comment.** Deleted the first "contains substr" block; kept the accurate exact-match description.

**ChiR5 — orphaned test doc comment.** Moved `newLegacySubscriberMock` above the `TestHandleIsBannedComprehensive` doc comment.

`go test ./services/rpc/` passes and `go vet` is clean. Left as follow-ups per the notes: the raw-context timeout on `isSubscriberActive`/`handleIsBanned`/`handleClearBanned`, and the same `!= nil`-only gate in `handleGetPeerInfo`.

ctnguyen (approved):
Review at commit [de0cfe1](https://github.com/bsv-blockchain/teranode/commit/de0cfe1)

## Status of previous points

| ChiR1 | setban false success | YES | FIXED |
| ChiR2 | GetSubscribers error unlogged | | FIXED |
| ChiR3 | skip test only on isbanned | | FIXED |
| ChiR4 | stale duplicated doc comment | | FIXED |
| ChiR5 | test helper orphans doc comment | | FIXED |

oskarszoon (changes requested):
The `setban` success-tracking refactor is a real fix and correctly done — I traced the full p2p×legacy matrix for both `add` and `remove` and found no path where a genuine success reports failure or a genuine failure reports success. The problem is the headline gate.

## P1 — `isSubscriberActive(SubscriberLegacy)` is always false, so this disables legacy peer banning instead of gating it

`blockchain.SubscriberLegacy = "legacy/manager"` has no live producer. #735 (`6ea612007`) removed the only `Subscribe(ctx, SubscriberLegacy)` call from `services/legacy/netsync/manager.go`, and nothing else registers that source. So `GetSubscribers()` can never contain `"legacy/manager"`, and all five gates evaluate false on every node — including nodes actively running legacy with `legacyP2PClient != nil`:

- `handlers.go:2119` (isbanned), `:2202` (listbanned), `:2279` (clearbanned), `:2386` / `:2426` (setban add/remove)

Effect: `isbanned` / `listbanned` / `clearbanned` / `setban add|remove` silently ignore legacy (SV Node) peers. That's the inverse of "skip when inactive" — it's "skip always," silently no-opping a security control (peer banning). The `handleGetInfo` precedent this mirrors has been broken the same way since #735 (cosmetic there; not on the ban surface).

Use a real liveness signal instead of the blockchain subscriber list:

1. restore the legacy manager's `Subscribe(SubscriberLegacy)` registration so the gate means something again; or
2. gate on the legacy service's own health/FSM, or a short-timeout probe of the peer service; or
3. for #591 specifically, keep the `!= nil` check and just wrap the two unbounded calls (`handleIsBanned`, `handleClearBanned`) in a timeout context so a missing legacy service can't stall the RPC — no semantic change to the legacy-present case.

## P1 — the fix's own new lines aren't covered

Every test that reaches the switch wires a non-nil `p2pClient` that succeeds whenever the call is expected to succeed. So the `success = true` added on the legacy-only success branches (`handlers.go:2411`, `:2447`) — the code this fix relies on for the legacy leg to independently produce success — is never load-bearing: delete both lines and no test fails. And the tests only pass at all because `newLegacySubscriberMock` fabricates a `["legacy/manager"]` subscriber list that cannot occur in production.

Add: (a) `p2pClient` nil + legacy active-success → no error (legacy leg alone carries success); (b) both legs absent/skipped → "Failed to apply ban" (only p2p-fails + legacy-skipped is tested today), for both `add` and `remove`.

## P2 — wrong wording on the unban path

`handlers.go:2461-2465` returns `"Failed to apply ban"` for `setban remove` too. An operator unbanning an IP sees "Failed to apply ban," which reads as if a ban was being added. Branch the message on `c.Command`, or use neutral wording ("Failed to apply setban command").

## Note (not blocking)

`p2pClient == nil` + legacy-skipped now returns an error where it previously returned silent `nil`. Defensible (an admin control shouldn't report false success) and no in-repo caller depends on the old no-op, but it's an RPC-contract change — worth a line in the PR description.

oskarszoon:
Re: tests — please don't close the coverage gap with another `GetSubscribers`-returns-`"legacy/manager"` mock; that's the exact blind spot that let this land green. `newLegacySubscriberMock` fabricates a subscriber state that can't occur in production (nothing has registered `SubscriberLegacy` since #735 removed the `Subscribe(ctx, SubscriberLegacy)` call in `services/legacy/netsync/manager.go`), so those assertions pass while the real path is dead.

What's needed is a test that exercises the real liveness path:

1. a regression pin that would fail today — with the legacy service actually running, assert `GetSubscribers()` contains `SubscriberLegacy` (red right now, which is the point); and/or
2. an integration/smoke test — legacy enabled, `setban add <ip>` then `isbanned <ip>` returns true, round-tripping through the real `peer_api` server.

Whichever liveness signal replaces the current gate, the test should assert it's true under a genuinely-running-legacy config, not a mocked one.

liam:
Great catch on the dead gate, @oskarszoon - you're right, `"legacy/manager"` has no `Subscribe()` producer since #735, so `isSubscriberActive(SubscriberLegacy)` is always false and my gate disabled legacy banning entirely. I went with your option 3.

**P1 (dead gate → disables legacy banning):** Dropped the `isSubscriberActive` gate on all five sites (back to `legacyP2PClient != nil`) and instead bounded each legacy ban call with a `ClientCallTimeout` context (`isbanned`, `clearbanned`, both `setban` arms; `listbanned` already had one). This solves the original #591 stall without disabling banning. `handleGetInfo`'s identical dead gate is left alone - it's pre-existing and cosmetic there; happy to file a separate issue.

**P1 (uncovered new lines):** Added `TestHandleSetBan_SuccessTracking` - the "p2p absent, legacy applies ban → success" cases make the `success = true` on the legacy branches load-bearing (verified they fail if those lines are removed), plus the both-absent → failure cases, for `add` and `remove`.

**P2 (unban wording):** The post-switch message now branches on the command - `setban remove` returns "Failed to remove ban".

**Note (contract change):** Documented in the updated PR description - `setban` is now strict where it was lenient.

Also added `TestBanHandlers_LegacyCallBounded` (legacy mock blocks on ctx; handler hangs without the per-call timeout) to guard the #591 fix directly.

oskarszoon (approved):
Re-reviewed at `39ab91f99`. The pass-1 P1 is fixed the right way: the dead `isSubscriberActive(SubscriberLegacy)` gate is gone from all five ban legs, replaced by the `legacyP2PClient != nil` check plus a `context.WithTimeout(ctx, ClientCallTimeout)` bound on each call. Legacy peer banning works again on legacy-enabled nodes, and an absent-but-configured legacy service can no longer stall the RPC on the parent context (#591). `ClientCallTimeout` defaults to 5s, so there's no instant-fail regression. The "Failed to apply ban" wording is branched to "Failed to remove ban" on the unban path, and the tests dropped the fabricated `["legacy/manager"]` subscriber mock in favour of a real timeout-bound test (`TestBanHandlers_LegacyCallBounded`). Traced the setban success matrix again — still no success/failure inversion on any path. Clearing my change request; approving.

Two trivial doc nits, non-blocking:
- `handlers.go:1607-1612` — `isSubscriberActive` now has two stacked, contradictory doc comments, and the new one says "whose source contains substr" while the body still does exact equality. Drop the duplicate and the "contains" wording.
- The commit removed `isSubscriberActive`'s warn log on a `GetSubscribers` error (now silently returns false). Only affects `handleGetInfo` now (its sole remaining caller), so cosmetic — but that also means the dead-gate story now lives entirely in `getinfo`; worth a follow-up to either retire `isSubscriberActive` or fix getinfo's legacy-peer display, separate from this PR.

PR #1243, opened 2026-07-08
fix(validator): correct off-by-one in parent-tx input bounds guard

### LINKED ISSUE #1144: [TN-005][informational] validator: off-by-one in dead bounds guard (idx > len should be >=)
**Severity:** informational · **Confidence:** high · **Pass:** consensus
**Component:** `services/validator/Validator.go`

### Summary
A bounds guard protecting a `tx.Inputs[idx]` access uses `idx > len(tx.Inputs)` instead of `>=`. Currently unreachable (`idx` comes from `range tx.Inputs`), so dormant.

### Evidence
- `services/validator/Validator.go:1076` — `if idx > len(tx.Inputs)`; index source `range tx.Inputs` at `:980-987`.

### Why it matters
If the index source ever changes, `idx == len` slips past the guard into an out-of-bounds panic.

### Suggested fix
Change to `idx >= len(tx.Inputs)`.

### Suggested tests
Unit-test `getUtxoBlockHeightAndExtendForParentTx` with `idxs=[len(tx.Inputs)]` → clean error, not panic.

---

liam:
## Problem

In `getUtxoBlockHeightAndExtendForParentTx` (`services/validator/Validator.go`), the extend path guards with:

```go
if idx > len(tx.Inputs) {
    return errors.NewProcessingError(... "out of bounds" ...)
}
// next statement:
... tx.Inputs[idx] ...
```

The very next statement dereferences `tx.Inputs[idx]`, so an index equal to `len(tx.Inputs)` slips past the guard and causes an out-of-bounds panic. The comparison should be `>=`.

Reported as [TN-005] in #1144.

## Fix

Change the comparison to `idx >= len(tx.Inputs)`.

The branch is **unreachable today** — every `idx` originates from `range tx.Inputs`, which only yields indices in `[0, len)` — so this is defensive hardening with no behavior change for any current caller. It removes a latent panic if the index source ever changes.

## Test

`Test_getUtxoBlockHeightAndExtendForParentTx_InputIndexOutOfBounds` drives the boundary directly (`idx == len(tx.Inputs)`), supplying the parent via the prefetch map so no store is needed. Verified: it panics (`index out of range [1] with length 1`) without the fix and returns a clean out-of-bounds error with it.

`go vet` clean. (golangci-lint can't run in this env — its Go 1.25 build rejects the 1.26 target.)

Closes #1144

### DIFF
--- a/services/validator/Validator.go
+++ b/services/validator/Validator.go
@@ -1172,7 +1172,7 @@ func (v *Validator) getUtxoBlockHeightAndExtendForParentTx(gCtx context.Context,
 	if extend {
 		// extend the transaction inputs with the parent tx outputs
 		for _, idx := range idxs {
-			if idx > len(tx.Inputs) {
+			if idx >= len(tx.Inputs) {
 				return errors.NewProcessingError("[Validate][%s] input index %d out of bounds for transaction with %d inputs",
 					tx.TxIDChainHash().String(), idx, len(tx.Inputs))
 			}
--- a/services/validator/Validator.go
+++ b/services/validator/Validator.go
@@ -1186,14 +1186,22 @@ func (v *Validator) getUtxoBlockHeightAndExtendForParentTx(gCtx context.Context,
 		// extend the transaction inputs with the parent tx outputs (idx bounds
 		// already validated at the top of the function)
 		for _, idx := range idxs {
-			if txMeta.Tx == nil || txMeta.Tx.Outputs == nil || txMeta.Tx.Outputs[tx.Inputs[idx].PreviousTxOutIndex] == nil {
-				return errors.NewProcessingError("[Validate][%s] parent transaction %s does not have outputs for input index %d",
-					tx.TxIDChainHash().String(), parentTxHash.String(), idx)
+			// PreviousTxOutIndex comes from the (untrusted) child transaction, so
+			// bound it against the parent's output count before indexing.
+			// Otherwise a tx referencing a real parent but a non-existent vout
+			// (e.g. vout 99 on a 2-output parent) panics here with index out of
+			// range and crashes the validator. Mirrors the guard in
+			// stores/utxo/aerospike/get.go.
+			vout := tx.Inputs[idx].PreviousTxOutIndex
+			if txMeta.Tx == nil || txMeta.Tx.Outputs == nil ||
+				int(vout) >= len(txMeta.Tx.Outputs) || txMeta.Tx.Outputs[vout] == nil {
+				return errors.NewProcessingError("[Validate][%s] parent transaction %s has no output for index %d",
+					tx.TxIDChainHash().String(), parentTxHash.String(), vout)
 			}
 
 			// extend the input with the parent tx outputs
-			tx.Inputs[idx].PreviousTxSatoshis = txMeta.Tx.Outputs[tx.Inputs[idx].PreviousTxOutIndex].Satoshis
-			tx.Inputs[idx].PreviousTxScript = txMeta.Tx.Outputs[tx.Inputs[idx].PreviousTxOutIndex].LockingScript
+			tx.Inputs[idx].PreviousTxSatoshis = txMeta.Tx.Outputs[vout].Satoshis
+			tx.Inputs[idx].PreviousTxScript = txMeta.Tx.Outputs[vout].LockingScript
 		}
 	}
 
--- a/services/validator/Validator.go
+++ b/services/validator/Validator.go
@@ -1130,6 +1130,19 @@ func (v *Validator) getUtxoBlockHeightsAndExtendTx(ctx context.Context, tx *bt.T
 //     height in policy mode.
 func (v *Validator) getUtxoBlockHeightAndExtendForParentTx(gCtx context.Context, parentTxHash chainhash.Hash, idxs []int,
 	utxoHeights []uint32, tx *bt.Tx, extend bool, prefetched map[chainhash.Hash]*meta.Data) error {
+	// Validate every target index up front, before any utxoHeights[idx] (the
+	// height loops below) or tx.Inputs[idx] (the extend loop) dereference. idxs
+	// are positions in tx.Inputs, and the caller sizes utxoHeights to
+	// len(tx.Inputs), so an out-of-range index would otherwise panic in the
+	// height loop before reaching the extend path. Unreachable today (idxs
+	// derive from range tx.Inputs), so this is purely defensive hardening.
+	for _, idx := range idxs {
+		if idx < 0 || idx >= len(tx.Inputs) || idx >= len(utxoHeights) {
+			return errors.NewProcessingError("[Validate][%s] input index %d out of bounds (%d inputs, %d height slots)",
+				tx.TxIDChainHash().String(), idx, len(tx.Inputs), len(utxoHeights))
+		}
+	}
+
 	f := []fields.FieldName{fields.BlockIDs, fields.BlockHeights}
 
 	if extend {
@@ -1170,13 +1183,9 @@ func (v *Validator) getUtxoBlockHeightAndExtendForParentTx(gCtx context.Context,
 	}
 
 	if extend {
-		// extend the transaction inputs with the parent tx outputs
+		// extend the transaction inputs with the parent tx outputs (idx bounds
+		// already validated at the top of the function)
 		for _, idx := range idxs {
-			if idx >= len(tx.Inputs) {
-				return errors.NewProcessingError("[Validate][%s] input index %d out of bounds for transaction with %d inputs",
-					tx.TxIDChainHash().String(), idx, len(tx.Inputs))
-			}
-
 			if txMeta.Tx == nil || txMeta.Tx.Outputs == nil || txMeta.Tx.Outputs[tx.Inputs[idx].PreviousTxOutIndex] == nil {
 				return errors.NewProcessingError("[Validate][%s] parent transaction %s does not have outputs for input index %d",
 					tx.TxIDChainHash().String(), parentTxHash.String(), idx)
--- a/services/validator/validator_prefetched_parents_test.go
+++ b/services/validator/validator_prefetched_parents_test.go
@@ -125,3 +125,29 @@ func Test_getUtxoBlockHeightAndExtendForParentTx_InputIndexOutOfBounds(t *testin
 	require.Error(t, err)
 	require.Contains(t, err.Error(), "out of bounds")
 }
+
+// Test_getUtxoBlockHeightAndExtendForParentTx_VoutOutOfRange guards the extend
+// path against an out-of-range PreviousTxOutIndex. The vout comes from the
+// (untrusted) child transaction; a raw tx that references a real parent but a
+// vout beyond that parent's output count must be rejected with a clean error
+// rather than panicking on txMeta.Tx.Outputs[vout] and crashing the validator.
+func Test_getUtxoBlockHeightAndExtendForParentTx_VoutOutOfRange(t *testing.T) {
+	ctx := context.Background()
+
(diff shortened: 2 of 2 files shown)


ordishs (approved):
Verified locally: the new test panics (`index out of range [1] with length 1`) on main and passes with the fix; full validator package green with `-race`, `go vet` clean. The sole caller derives indices from `range tx.Inputs`, so this strictly widens an unreachable error path — no behaviour change.

Two non-blocking follow-ups for the same function (fine as separate work):
- The height-population loops above the guard write `utxoHeights[idx]` unguarded, so an out-of-bounds index through the real call path would panic there before reaching this check. Validating `idxs` once at the top of the function would cover both loops.
- Line 1180 indexes `txMeta.Tx.Outputs[tx.Inputs[idx].PreviousTxOutIndex]` without checking `PreviousTxOutIndex < len(Outputs)` — same latent-panic class as this fix.

ctnguyen (changes requested):
Review at commit [ff0bd00](https://github.com/bsv-blockchain/teranode/commit/ff0bd00)

The fix itself is correct: `>=` is the right guard for the 0-indexed `tx.Inputs[idx]` dereference that follows, and the PR is commendably honest that the branch is unreachable today (every `idx` originates from `range tx.Inputs`). The regression test drives the exact boundary and the description clearly explains the intent. The instinct — hardening this extension path against index panics — is exactly right; it just needs to point at the reachable hazard.

## Blocking issues

### ChiR1 — Reachable, unauthenticated remote-DoS panic on out-of-range `PreviousTxOutIndex` (same function, 6 lines below the fix)
**Problem:** At [Validator.go:1180](services/validator/Validator.go#L1180) the extend loop indexes the parent's output slice with an attacker-controlled child-input field:
```go
if txMeta.Tx == nil || txMeta.Tx.Outputs == nil || txMeta.Tx.Outputs[tx.Inputs[idx].PreviousTxOutIndex] == nil {
```
The nil checks guard a nil `Tx`/`Outputs`, but nothing bounds-checks `PreviousTxOutIndex` against `len(txMeta.Tx.Outputs)`. A raw (non-extended) transaction that references a **real, existing parent txid** but a vout beyond that parent's output count (e.g. vout 99 on a 2-output parent) reaches this line and panics with `index out of range`. This was confirmed empirically (a repro sized `utxoHeights` to the production shape and observed `runtime error: index out of range [99] with length 2`), and the path is reachable from both the gRPC `ValidateTransaction` handler and the Kafka submission path — the extend step at line 769 runs *before* `spendUtxos` (the outpoint-existence check) at line 818, so nothing rejects the bogus vout first. The validator's gRPC server is started with `authOptions=nil` (no recovery interceptor), grpc-go v1.81.1 has no built-in handler-panic recovery, and the Kafka consumer path has no `recover()` either — so the panic crashes the process.
**Why it matters:** A single unauthenticated, malformed transaction can crash the validator service — a remote DoS, and substantially more severe than the off-by-one this PR fixes. Since this PR's stated purpose is to harden exactly this function against index panics, shipping it while leaving the reachable panic in place delivers a false sense that the function is now safe.
**Fix:** Bounds-check the vout before indexing, mirroring the guard the codebase already applies elsewhere ([stores/utxo/aerospike/get.go:1477](stores/utxo/aerospike/get.go#L1477)):
```go
vout := tx.Inputs[idx].PreviousTxOutIndex
if txMeta.Tx == nil || txMeta.Tx.Outputs == nil ||
	int(vout) >= len(txMeta.Tx.Outputs) || txMeta.Tx.Outputs[vout] == nil {
	return errors.NewProcessingError("[Validate][%s] parent transaction %s has no output for index %d",
		tx.TxIDChainHash().String(), parentTxHash.String(), vout)
}
```
(This is a pre-existing defect, not introduced by the diff — but it belongs in this hardening change, or in an urgently tracked follow-up, rather than being left behind a fix to dead code.)

## Non-blocking issues

### ChiR2 — The guard hardens an unreachable branch; the delivered protection and its test overstate what was fixed
**Problem:** The guarded read `tx.Inputs[idx]` at line 1175 is preceded by unguarded writes to `utxoHeights[idx]` at [Validator.go:1164](services/validator/Validator.go#L1164) and [1168](services/validator/Validator.go#L1168). The sole production caller sizes `utxoHeights := make([]uint32, len(tx.Inputs))` ([line 1077](services/validator/Validator.go#L1077)), so any `idx >= len(tx.Inputs)` is *also* out of range for `utxoHeights` and would panic at 1164/1168 — before ever reaching this guard. The new test only reaches the guard because it artificially sizes `utxoHeights := make([]uint32, 2)` against a 1-input tx, a shape no real caller produces. So the description's claim ("removes a latent panic if the index source ever changes") isn't delivered by this guard alone.
**Why it matters:** The change is correct but cosmetic for any realistic caller; the test creates false confidence that a real-world panic path was closed.
**Fix:** Either soften the description to "cosmetic fix to a currently-dead branch," or make the hardening real by validating indices before the `utxoHeights` writes and testing with a realistic caller shape:
```go
for _, idx := range idxs {
	if idx < 0 || idx >= len(tx.Inputs) || idx >= len(utxoHeights) {
		return errors.NewProcessingError("[Validate][%s] input index %d out of bounds (%d inputs)",
			tx.TxIDChainHash().String(), idx, len(tx.Inputs))
	}
}
// ... then the existing utxoHeights writes and extend loop
```

## Recap

| ChiR1 | Reachable remote-DoS panic on bad PreviousTxOutIndex | YES | 85% |
| ChiR2 | Guard/test harden dead code, overstate protection |  | 35% |

liam:
Good catch, you're right that the guard was only partial. `utxoHeights` is sized `make([]uint32, len(tx.Inputs))` by the caller and written at the height loops (`utxoHeights[idx]`) before the extend path, so an out-of-range `idx` would panic there first, and my test only dodged that by oversizing `utxoHeights`.

Fixed in the follow-up commit: the bounds check is now hoisted to the top of `getUtxoBlockHeightAndExtendForParentTx` and validates every `idx` against both `len(tx.Inputs)` and `len(utxoHeights)` before any dereference; the redundant in-loop check is dropped. The test now sizes `utxoHeights` as the real caller does (`make([]uint32, len(tx.Inputs))`), and I confirmed it panics in the height loop (`index out of range [1] with length 1`) without the hoisted guard and returns a clean out-of-bounds error with it. Still unreachable today, so purely defensive as before.

ctnguyen (commented):
Review at commit [9e4d3e7](https://github.com/bsv-blockchain/teranode/commit/9e4d3e7)

Since the last pass the author reworked the fix in response to ChiR2: the bounds check is now hoisted to the top of `getUtxoBlockHeightAndExtendForParentTx` and validates each `idx` against both `len(tx.Inputs)` and `len(utxoHeights)` before any dereference, and the regression test now sizes `utxoHeights` as the real caller does. That rework is a clean improvement, but the blocking issue ChiR1 remains unaddressed, so this PR is **not yet mergeable**.

## Outstanding blocking issue (from the previous pass)

ChiR1 — the reachable, unauthenticated remote-DoS panic on an out-of-range `PreviousTxOutIndex` — is **still not fixed** at this head. Line 1189 still indexes `txMeta.Tx.Outputs[tx.Inputs[idx].PreviousTxOutIndex]` with only nil-checks on `Tx`/`Outputs` and no bounds check of the attacker-controlled `PreviousTxOutIndex` against `len(txMeta.Tx.Outputs)`. A raw transaction referencing a real parent txid but a vout beyond that parent's output count still reaches this line and panics `index out of range`. The path is reachable (the extend step runs before `spendUtxos`, so the bogus vout is not rejected earlier) from both the gRPC and Kafka submission paths, with no recovery interceptor, so the panic crashes the process. This point is **not renumbered** — it stays ChiR1 and appears in the status table below with Status NOT FIXED. It must be resolved before merge; the fix is the vout bounds-check that mirrors the existing guard at `stores/utxo/aerospike/get.go:1477` (see the prior point's snippet).

## Status of previous points

| ChiR1 | Bounds-check PreviousTxOutIndex vout | YES | NOT FIXED |
| ChiR2 | Guard hardened dead branch; test overstated |  | FIXED |

liam:
ChiR1 fixed in `286ca44` — thanks, this was a real reachable panic and I'd missed it in the earlier pass.

`getUtxoBlockHeightAndExtendForParentTx` now bounds-checks the vout before indexing, mirroring the guard in `stores/utxo/aerospike/get.go`:

```go
vout := tx.Inputs[idx].PreviousTxOutIndex
if txMeta.Tx == nil || txMeta.Tx.Outputs == nil ||
    int(vout) >= len(txMeta.Tx.Outputs) || txMeta.Tx.Outputs[vout] == nil {
    return errors.NewProcessingError("[Validate][%s] parent transaction %s has no output for index %d",
        tx.TxIDChainHash().String(), parentTxHash.String(), vout)
}
```

Added `Test_getUtxoBlockHeightAndExtendForParentTx_VoutOutOfRange` (raw tx → real parent → vout 99 on a 2-output parent). Verified it panics `index out of range [99] with length 2` without the guard and returns a clean error with it.

Unrelated note: `TestExtendedTxa1f6a4ff...` currently fails on `upstream/main` as well (pre-existing, "parent tx hashes and indexes length mismatch"), so it's not from this change.

freemans13 (approved):
Reviewed the diff (both guards live in `getUtxoBlockHeightAndExtendForParentTx`). The off-by-one fix is correct: hoisting the check above the height-write loops and switching `idx > len` to `idx >= len(tx.Inputs)` closes both the extend-path panic and the earlier `utxoHeights[idx]` one. The added vout bounds-check mirrors the aerospike guard (`stores/utxo/aerospike/get.go:1481`) exactly and fixes a genuine second panic on an out-of-range `PreviousTxOutIndex`. Verified locally: `go vet ./services/validator/` clean, and both new tests plus the existing `getUtxoBlockHeightsAndExtendTx` suite pass. The only production caller sources `idxs` from `range tx.Inputs`, so the "unreachable today, defensive hardening" framing holds.

One suggestion (non-blocking): the PR body only describes the idx off-by-one fix — consider adding a line about the `PreviousTxOutIndex` bounds-check and its `VoutOutOfRange` test so the second hardening isn't a surprise.

LGTM.

ctnguyen (approved):
Review at commit [286ca44](https://github.com/bsv-blockchain/teranode/commit/286ca44)

This revision closes the loop cleanly. The reachable, unauthenticated DoS panic flagged in the prior pass (ChiR1) is now fixed: `PreviousTxOutIndex` is bound against `len(txMeta.Tx.Outputs)` before indexing, short-circuiting bounds-first then nil-check, and mirroring the existing guard in `stores/utxo/aerospike/get.go`. The up-front index guard (ChiR2) validates every `idx` against both `len(tx.Inputs)` and `len(utxoHeights)` before any dereference, so the height-write loops are covered too, not just the extend loop. Both fixes ship with targeted regression tests that drive the exact boundaries, and the comments accurately explain the untrusted-input threat model. Good, honest hardening.

The extend path is now panic-safe against both attacker-controlled surfaces: an out-of-range `idx` (caught up front) and an out-of-range `vout` (caught in the loop). I verified the concurrency angle — each input index appears under exactly one parent in `parentTxHashes`, so the errgroup goroutines write disjoint `utxoHeights[idx]` and `tx.Inputs[idx]` slots with no shared-element race; the added guard reads only lengths and is race-free. The vout guard's ordering (`Outputs == nil` → `int(vout) >= len` → `Outputs[vout] == nil`) correctly avoids indexing an empty or nil slice. Behaviour for valid transactions is unchanged; the new condition only rejects inputs that previously panicked.

## Status of previous points

| ChiR1 | Bounds-check PreviousTxOutIndex vout | YES | FIXED |
| ChiR2 | Guard hardened dead branch; test overstated |  | FIXED |

PR #1245, opened 2026-07-08
fix(expiringmap): make TTL clean() eviction send non-blocking

### LINKED ISSUE #1152: [TN-013][low] util/expiringmap: blocking send to evictionCh under write lock (latent)
**Severity:** low · **Confidence:** high · **Pass:** concurrency
**Component:** `util/expiringmap/expiringmap.go`

### Summary
`clean()` (background ticker) and the evict-oldest path do a blocking send to `evictionCh` while holding `m.mu`. If `evictionCh` is ever wired with a slow consumer, the ticker blocks under the lock and every `Get`/`Set`/`Delete`/`Len` deadlocks. Latent: no prod code wires `WithEvictionChannel` (only `WithEvictionFunction` is used).

### Evidence
- `util/expiringmap/expiringmap.go:261` blocking send in `clean()` under `m.mu.Lock()` (`:238`); `:150` similar in evict-oldest. `:64` `WithEvictionChannel` (no prod caller).

### Why it matters
A latent deadlock that surfaces the moment the eviction-channel API is adopted.

### Suggested fix
Make both sends non-blocking (`select { case ch<-…: default: }`) or send after releasing the lock.

### Suggested tests
Wire a 0-capacity `evictionCh` with no reader, trigger `clean()` → assert `Get`/`Set` still return within a timeout.

---

liam:
## Problem

`ExpiringMap.clean()` (the TTL ticker path) sent expired items to the eviction channel with a **blocking** send while holding `m.mu`:

```go
if m.evictionCh != nil && len(expiredItems) > 0 {
    m.evictionCh <- expiredItems   // blocking, under m.mu
}
```

If a configured eviction channel were ever full or unconsumed, the ticker goroutine would block on this send while holding the write lock, deadlocking every `Get`/`Set`/`Delete`/`Len`. The sibling cap-eviction path (`evictOldestLocked`) already uses a non-blocking send; only `clean()` was left blocking.

Reported as [TN-013] in #1152.

## Fix

Use a non-blocking `select { case m.evictionCh <- expiredItems: default: }`, matching the cap-eviction path, so eviction never blocks under `m.mu`. The `WithMaxSize` doc note (which previously stated the clean() send "remains blocking") is updated to match.

This is **latent today** — no production code wires `WithEvictionChannel` — so there is no behavior change for current callers; it removes the deadlock hazard for any future consumer.

## Test

`TestExpiringMap_CleanEvictionSendDoesNotBlockUnderLock` configures an unbuffered, unread eviction channel, inserts an already-expired item, and asserts `clean()` returns within a timeout (and still evicts the item). Verified: it hangs → fails on the 2s timeout without the fix, passes with it. Full package passes under `-race`.

Closes #1152

### DIFF
--- a/util/expiringmap/expiringmap.go
+++ b/util/expiringmap/expiringmap.go
@@ -84,9 +84,9 @@ func (m *ExpiringMap[K, V]) WithEvictionFunction(f func(K, V) bool) *ExpiringMap
 // cap is always honored. A value of 0 (the default) disables the cap and
 // preserves the original unbounded behaviour.
 //
-// Note: cap-eviction's eviction-channel send is non-blocking (the notification
-// is dropped if the channel is full), since cap-eviction runs synchronously
-// inside Set. The TTL clean() path's eviction-channel send remains blocking.
+// Note: the eviction-channel send is non-blocking on both the cap-eviction
+// path (inside Set) and the TTL clean() path (the ticker) — the notification
+// is dropped if the channel is full — so eviction never blocks under m.mu.
 func (m *ExpiringMap[K, V]) WithMaxSize(n int) *ExpiringMap[K, V] {
 	m.mu.Lock()
 	defer m.mu.Unlock()
@@ -258,6 +258,13 @@ func (m *ExpiringMap[K, V]) clean() {
 	}
 
 	if m.evictionCh != nil && len(expiredItems) > 0 {
-		m.evictionCh <- expiredItems
+		// Non-blocking send: clean() holds m.mu, so a blocking send to a full
+		// (or unconsumed) eviction channel would stall the ticker under the
+		// write lock and deadlock all Get/Set/Delete/Len. Drop the notification
+		// if the consumer can't keep up, matching the cap-eviction path.
+		select {
+		case m.evictionCh <- expiredItems:
+		default:
+		}
 	}
 }
--- a/util/expiringmap/expiringmap_test.go
+++ b/util/expiringmap/expiringmap_test.go
@@ -158,3 +158,42 @@ func TestExpiringMap_NoMaxSize_DefaultUnbounded(t *testing.T) {
 
 	require.Equal(t, 100, m.Len())
 }
+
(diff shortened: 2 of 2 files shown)


ordishs (approved):
Correct, minimal fix. Verified locally:

- `go test -race -count=1 ./util/expiringmap/` passes on the branch.
- Temporarily reverting just the `select`/`default` back to a blocking send makes `TestExpiringMap_CleanEvictionSendDoesNotBlockUnderLock` fail on the 2s timeout as described — the test genuinely guards the regression.
- Confirmed `WithEvictionChannel` has no production callers, so the drop-on-full semantics change affects no current behavior.

The two eviction paths now have consistent drop-on-full semantics, and the `WithMaxSize` doc note was updated to match — good attention to detail.

One nit, non-blocking: the new test omits `defer m.Stop()`, which every other test in the file uses. `New(time.Hour)` does start the ticker goroutine (it just never fires within the test), so adding the `Stop()` would avoid the leaked goroutine and match the file's convention.

Design note for the record: `clean()` drops the whole expired batch when the consumer is slow, whereas cap-eviction drops at most one item per `Set`. Since the channel is unused in production and `evictionFn` exists for guaranteed per-item hooks, drop-on-full is the right call here — just worth remembering if a future consumer relies on the channel for mandatory resource cleanup.

ctnguyen (approved):
This is a clean, minimal fix for a real latent deadlock — a blocking channel send under `m.mu` in the ticker path could have stalled every `Get`/`Set`/`Delete`/`Len`, and the `select { … default: }` rewrite closes that hazard while bringing `clean()` into line with the cap-eviction path's existing drop-on-full semantics. The `WithMaxSize` doc note is updated accurately to reflect that both paths are now non-blocking, and the regression test genuinely exercises the failure (hangs without the fix, passes with it, and confirms the item is still evicted). The PR is also refreshingly honest about scope: it frames the bug as latent with no production caller of `WithEvictionChannel`, so there's no behavior change for anyone today.

PR #1246, opened 2026-07-08
fix(model): overflow-safe coinbase reward/fee check

### LINKED ISSUE #1148: [TN-009][low] model: coinbase reward/fee check lacks overflow/MaxMoney guard
**Component:** `model/Block.go` (`checkBlockRewardAndFees`), `util/block_subsidy.go`

### Summary
`checkBlockRewardAndFees` sums coinbase outputs without an overflow check and compares `coinbaseOutputSatoshis > subtreeFees + coinbaseReward` without bounding either side to `MaxMoney`. SV Node enforces an explicit `MoneyRange` here. Not currently reachable (would need ~184bn BTC-equiv to wrap uint64).

### Evidence
- `model/Block.go:642-657` — output sum loop (`:642-645`), comparison `:656`; no MaxMoney cap.
- `util/block_subsidy.go` — no MaxMoney cap.

### Why it matters
If the sum ever wrapped uint64, an over-issuing coinbase could pass. Defensive belt-and-suspenders.

### Suggested fix
Overflow-safe accumulation + explicit `MaxMoney` range check on the coinbase total and `subtreeFees+coinbaseReward`.

### Suggested tests
Block whose coinbase outputs wrap uint64 (and/or `subtreeFees` near MaxUint64) → assert rejection.

---

liam:
## Problem

`Block.checkBlockRewardAndFees` accumulated the coinbase output total and the subtree fees with plain unchecked `uint64` additions, and compared against an unchecked `subtreeFees + coinbaseReward`:

```go
coinbaseOutputSatoshis += tx.Satoshis   // unchecked
subtreeFees += subtree.Fees             // unchecked
if coinbaseOutputSatoshis > subtreeFees+coinbaseReward { ... }   // RHS unchecked
```

A `uint64` wraparound in any of these sums could make the no-inflation comparison pass on a block that is actually inflating supply. SV Node enforces `MoneyRange` on these values; Teranode had no overflow/range guard here.

Reported as [TN-009] in #1148.

## Fix

Detect overflow on each accumulation (`sum < prev`) and on the `fees + subsidy` sum, returning `BlockInvalidError`.

This is **purely additive**: for every real block the sums are far below `2^64` satoshis, so `allowedReward == subtreeFees + coinbaseReward` and the comparison is identical to before. Only impossible, overflowing values are newly rejected. As the issue notes, wrapping `uint64` would require ~184 quintillion satoshis, so this is defense-in-depth, not a reachable exploit today.

## Test

New subtest `coinbase output satoshi sum overflow is rejected` sets two coinbase outputs whose satoshi sum wraps `uint64` and asserts the block is rejected. Verified: without the guard the wrapped total (0) slips past the comparison and the block is wrongly accepted (test fails); with the guard it's rejected. All existing `checkBlockRewardAndFees` / checkpoint-skip tests still pass; `go vet` clean.

Closes #1148

### DIFF
--- a/model/Block.go
+++ b/model/Block.go
@@ -765,22 +765,45 @@ func (b *Block) checkBlockRewardAndFees(params *chaincfg.Params, storeSupportsOu
 		return nil
 	}
 
+	// Accumulate with overflow detection. A uint64 wraparound in any of these
+	// sums could make the no-inflation comparison below pass on a block that is
+	// actually inflating supply, so a sum that overflows is itself invalid.
+	// These overflows cannot occur in a valid block (total value is far below
+	// 2^64 satoshis), so this only rejects impossible values and does not change
+	// the result for any real block.
 	coinbaseOutputSatoshis := uint64(0)
+
 	for _, tx := range b.CoinbaseTx.Outputs {
-		coinbaseOutputSatoshis += tx.Satoshis
+		sum := coinbaseOutputSatoshis + tx.Satoshis
+		if sum < coinbaseOutputSatoshis {
+			return errors.NewBlockInvalidError("[checkBlockRewardAndFees][%s] coinbase output satoshis overflow uint64", b.String())
+		}
+
+		coinbaseOutputSatoshis = sum
 	}
 
 	subtreeFees := uint64(0)
 
 	for i := 0; i < len(b.SubtreeSlices); i++ {
 		subtree := b.SubtreeSlices[i]
-		subtreeFees += subtree.Fees
+
+		sum := subtreeFees + subtree.Fees
+		if sum < subtreeFees {
+			return errors.NewBlockInvalidError("[checkBlockRewardAndFees][%s] subtree fees overflow uint64", b.String())
+		}
+
+		subtreeFees = sum
 	}
 
 	coinbaseReward := util.GetBlockSubsidyForHeight(b.Height, params)
 
-	if coinbaseOutputSatoshis > subtreeFees+coinbaseReward {
-		return errors.NewBlockInvalidError("[checkBlockRewardAndFees][%s] coinbase output (%d) is greater than the fees + block subsidy (%d)", b.String(), coinbaseOutputSatoshis, subtreeFees+coinbaseReward)
+	allowedReward := subtreeFees + coinbaseReward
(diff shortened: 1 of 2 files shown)


ordishs (approved):
Overflow guards are correct (sum < prev on all three accumulation points) and behaviour-preserving for any real block — the comparison is unchanged whenever the sums don't wrap. Guard sits after the checkpoint fast-path skip, so that logic is untouched. New subtest genuinely exercises the check (height 1, no checkpoints) and pins the exact wrapped-to-zero case that slipped past the old comparison. Ran the reward/fee and checkpoint tests with -race plus go vet on model/ — all green.

Non-blocking suggestion: add subtests for the other two guards (subtree fees overflow, fees + subsidy overflow) so all three branches are pinned.

ctnguyen (approved):
Solid defense-in-depth. The `sum := a + b; if sum < a` idiom is the canonical unsigned-overflow test and is applied correctly at all three points, with all operands confirmed `uint64`. The change is genuinely additive: for any non-overflowing block `allowedReward == subtreeFees + coinbaseReward` bit-for-bit, so the comparison is identical to before and no valid block is newly rejected. It's correctly placed after the checkpoint fast-path skip, so that path is untouched, and it actually *narrows* the accept-set toward SV Node (which rejects wrapping money values via `MoneyRange`) rather than introducing any divergence. The new subtest genuinely reaches the guard (height 1, no checkpoints) and pins the exact wrapped-to-zero case that slipped past the old comparison.

## Non-blocking issues

### ChiR1 — Two of the three overflow branches are untested
**Problem:** The new subtest exercises only the coinbase-output-overflow branch. The `subtree fees overflow uint64` and `fees + block subsidy overflow uint64` branches — though they use the same idiom — have no coverage.
**Why it matters:** All three are consensus-relevant guards; pinning each independently protects against a future edit silently dropping or breaking one of them.
**Fix:** Add two small subtests mirroring the existing one — e.g. two `SubtreeSlices` with `Fees = ^uint64(0)` and `1` for the subtree-fees branch, and a block whose `subtreeFees` is near `^uint64(0)` at a height with a non-zero subsidy for the `fees + subsidy` branch — each asserting rejection with `"overflow"`.

PR #1094, opened 2026-06-15
 blockvalidation: collapse redundant RPCs in findCommonAncestor

gokutheengineer:
## Summary

  `findCommonAncestor` issued **2 sequential blockchain RPCs per existing header** during the
  common-ancestor walk (`GetBlockExists` → `GetBlockHeader`), plus a **3rd redundant 
  `GetBlockHeader`** after the loop to fetch metadata it had already seen. This collapses all of
  them down to a **single `GetBlockHeader` per header**.

  The insight: `GetBlockHeader` already conveys both existence (it returns `ErrBlockNotFound`
  when the block is absent) and height in one call. `GetBlockExists` and `GetBlockHeader` query
  the same `blocks` table by hash, so they're semantically equivalent for an existence check.

  ## Why it matters

  The loop walks peer headers oldest→newest and stops at the first missing block. When a node is
  mostly in sync (the common ancestor sits near the end of a large header batch), nearly every
  header in the batch incurred 2 RPCs instead of 1. With `maxBlockHeadersPerRequest` in the
  thousands, that's thousands of redundant round-trips per catchup iteration. Per-iteration RPC
  count drops from ~`2N+1` to `N`.

  ## Changes

  - **`services/blockvalidation/catchup.go`** — `findCommonAncestor` now issues one
  `GetBlockHeader` per header, treating `ErrBlockNotFound` as the "stop the search" signal and
  any other error as a genuine failure. The common ancestor's metadata is captured during the
  walk, eliminating the post-loop re-fetch.
  - **`services/blockvalidation/catchup_test.go`** — updated the two direct 
  `TestFindCommonAncestor_*` tests to drive the new single-RPC path via `GetBlockHeader` mocks 
  (including a not-found return at the divergence point) instead of `GetBlockExists`.

  ## Correctness notes

  - Behavior is unchanged: the same headers are walked, the same height-vs-UTXO comparison is 
  applied, and the same common ancestor is selected.
  - The not-found error code survives the gRPC round-trip — `WrapGRPC`/`UnwrapGRPC` preserve 
  `ERR_BLOCK_NOT_FOUND`, so `errors.Is(err, errors.ErrBlockNotFound)` works in production, not 
  just against the in-process mock.
  - Treating not-found as "stop" is marginally more robust than before: the old code would error 
  out if a block was pruned between the `GetBlockExists` and `GetBlockHeader` calls; the new code
  degrades gracefully.
[...]

PR #1247, opened 2026-07-08
test(blockvalidation): fix broken main — catch-all GetBlockHeader mock in catchup fork scenario

oskarszoon:
## What

Fixes `TestCatchup_NoMinedSetChurnOnRejectedFork/SecretMiningDepth`, which has been failing on `main` since #1094 merged (breaks the `Teranode main test build push` Go test job).

## Root cause — semantic merge conflict

Two blockvalidation PRs collided:

- #1197 (`gate secret-mining verdict on chainwork`) added this test plus a guard that bails with `"no offered headers to compare chainwork"` when the peer offers no headers past the common ancestor. Its `buildForkScenario` mocks `GetBlockHeader` to return a valid header (nil error) for **any** hash — correct against the *old* `findCommonAncestor`, which detected block absence through a separate `GetBlockExists` call (true only for the ancestor).
- #1094 (`collapse redundant RPCs in findCommonAncestor`) refactored the walk to drop `GetBlockExists` and detect absence purely from `GetBlockHeader` returning `ErrBlockNotFound`. It branched **before** #1197, so it never saw that test; its own test updates use `NewBlockNotFoundError` catch-alls.

The merge was textually clean (disjoint hunks) but semantically broken. With the refactored walk, the test's catch-all `GetBlockHeader` never returns not-found, so every fork header looks present → the common ancestor lands on the **last** peer header → `offeredHeaders` is empty → the code trips the `"no offered headers"` guard instead of reaching the `"secretly mined"` verdict the test asserts.

**Production code was already correct** — a real blockchain client returns `ErrBlockNotFound` for the peer's fork blocks. Only the test mock modeled reality wrong.

## Fix

Test-only. The catch-all `GetBlockHeader` mock in `buildForkScenario` now returns `NewBlockNotFoundError` for non-ancestor hashes, matching:
- the real blockchain client,
- the `GetBlockExists`=false mock already in the same scenario,
- the sibling `catchup_test.go` mocks updated in #1094.

Walk now stops at the first fork header → `offeredHeaders` holds the 5 fork headers → local chain outweighs the fork → `"secretly mined"` verdict. Both subtests pass.

## Verification

```
go test -tags testtxmetacache -run TestCatchup_NoMinedSetChurnOnRejectedFork ./services/blockvalidation/   # ok
go test -tags testtxmetacache ./services/blockvalidation/                                                  # ok, 278s
```

### DIFF
--- a/services/blockvalidation/catchup_fork_handling_test.go
+++ b/services/blockvalidation/catchup_fork_handling_test.go
@@ -690,8 +690,12 @@ func TestCatchup_NoMinedSetChurnOnRejectedFork(t *testing.T) {
 		// Metadata for the common ancestor.
 		mockBlockchainClient.On("GetBlockHeader", mock.Anything, ancestorHeader.Hash()).
 			Return(ancestorHeader, &model.BlockHeaderMeta{Height: ancestorHeight, ID: 1, ChainWork: ancestorChainWork.Bytes()}, nil).Maybe()
+		// Any other header (the peer's fork blocks) is absent from our chain, so
+		// GetBlockHeader reports not found and the common-ancestor walk stops at the
+		// divergence point, leaving the fork blocks as offered headers.
 		mockBlockchainClient.On("GetBlockHeader", mock.Anything, mock.Anything).
-			Return(&model.BlockHeader{}, &model.BlockHeaderMeta{Height: ancestorHeight, ChainWork: ancestorChainWork.Bytes()}, nil).Maybe()
+			Return((*model.BlockHeader)(nil), (*model.BlockHeaderMeta)(nil),
+				errors.NewBlockNotFoundError("block not found")).Maybe()
 
 		// Header-fetch plumbing.
 		mockBlockchainClient.On("GetBestBlockHeader", mock.Anything).


freemans13 (approved):

ctnguyen (approved):
The root-cause analysis is accurate and well-reasoned: this is a genuine semantic merge conflict between #1094 (absence now detected via `GetBlockHeader` returning `ErrBlockNotFound`, `catchup.go:509-533`) and #1197's test, which was written against the older `GetBlockExists`-based walk. The fix is minimal and strictly test-only (+5/-1), and it makes the mock behave like the real blockchain client — returning not-found for non-ancestor hashes so the common-ancestor walk stops at the divergence point and leaves the fork blocks as offered headers. It matches the sibling convention exactly (`catchup_test.go:2574-2579` uses the same typed-nil + `NewBlockNotFoundError("block not found")` pattern and near-identical comment), the `errors` import and constructor are in place, and the explanatory comment clearly documents why the mock returns not-found. Both subtests trace through correctly against the refactored production code. Nothing to block or improve — a clean fix.

PR #1248, opened 2026-07-08
ci(sonar): post Quality Gate result as PR commit status

oskarszoon:
## What

`sonar-pr-analyze.yaml` runs via `workflow_run`, so its Quality Gate result never
became a PR check and couldn't be marked required. This adds a `github-script` step
that posts the gate result as a commit status on the **tested commit** under the
context **`SonarQube Quality Gate`**, which can then be required in branch protection.

## Why this route (not the SonarCloud App)

`workflow_run` is kept deliberately: fork PRs can't read `SONAR_TOKEN`, so the scan
must run in base-repo context. A commit status stays entirely inside that fork-safe
flow — no App install, no analysis-method change, coverage/lint imports untouched.

## Status resolves on every path that runs

Everything — checkout, coverage, scan, and the posted status — describes the same
commit (`github.event.workflow_run.head_sha`), so the check vouches only for the tree
that was actually scanned.

| Upstream tests | Analysis | Status posted to tested commit |
|---|---|---|
| success | gate PASSED / WARN | `success` |
| success | gate FAILED | `failure` |
| success | scan errored (gate skipped) | `error` |
| failed / skipped / timed_out | steps guarded off | `failure` — "Upstream tests … — Sonar not run" |
| cancelled (superseded by a newer push) | job excluded at `if:` | none — the successor run reports on the new head |
| tests never triggered, or a manual cancel on the current head | this workflow never runs | none — uncoverable via `workflow_run` |

Implementation:
- **Posts to `workflow_run.head_sha`** (the tested commit), not the live PR head — and
  the `Checkout tested commit` step is pinned to the same SHA, so scan + coverage +
  status all agree. Posting to the live head would stamp commit A's result onto B.
- **Job `if:` excludes `cancelled`** upstream runs: `teranode_pr_tests` uses
  `cancel-in-progress`, so a push to an active PR cancels the superseded commit's run;
  its successor reports on the new head, so acting here would only post a spurious red.
- **Status step `if: ${{ !cancelled() }}`** — a self-cancelled sonar run never posts a
  transient error. It reads the gate output + upstream conclusion, and guards `if (!sha)`
  so it fails loud instead of posting a false green. The `WARN` gate outcome is treated
  as `success` ("passed with warnings").
- Analysis steps are guarded to `conclusion == 'success'` (they need artifacts a failed
  run never produces); GitHub implicitly ANDs `success()` onto a custom step `if:`, so a
  mid-job scan failure also skips the gate and is caught as `error`.
- Permissions: `statuses: write` (+ `contents`/`pull-requests: read`) only.

## Enforcing later

1. Mark **`SonarQube Quality Gate`** a required status check.
2. Require the tests check alongside it — the uncoverable case is the tests workflow
   never triggering (or a manual cancel on the current head), which also never triggers
   this one, so a required sonar status alone could hang.

### DIFF
--- a/.github/workflows/sonar-pr-analyze.yaml
+++ b/.github/workflows/sonar-pr-analyze.yaml
@@ -23,9 +23,8 @@ concurrency:
 
 permissions:
   contents: read
-  pull-requests: write
+  pull-requests: read
   statuses: write
-  id-token: write
 
 env:
   GO_VERSION: "1.26.0"
@@ -92,7 +91,6 @@ jobs:
             core.setOutput('number', prNumber);
             core.setOutput('base', pr.data.base.ref);
             core.setOutput('head', pr.data.head.ref);
-            core.setOutput('head_sha', pr.data.head.sha);
 
       - name: Checkout base branch
         if: github.event.workflow_run.conclusion == 'success'
@@ -173,9 +171,10 @@ jobs:
       # commit A's result onto B — a stale green on unscanned code, or a spurious
       # red — which is exactly what a required check must never do.
       #
-      # workflow_run runs in base-repo context, so this also works for fork PRs
-      # (their SONAR_TOKEN-less builds never reach here). Mark it a required check
-      # in branch protection under the context "SonarQube Quality Gate".
+      # workflow_run runs in base-repo context where SONAR_TOKEN is available, so
+      # fork PRs are scanned here too — the fork's own token-less test run never
+      # runs this job. Mark it a required check in branch protection under the
+      # context "SonarQube Quality Gate".
       #
       # if: !cancelled() — a self-cancelled sonar run (its own concurrency group)
       # must never post a transient error. On success/failure it resolves every
--- a/.github/workflows/sonar-pr-analyze.yaml
+++ b/.github/workflows/sonar-pr-analyze.yaml
@@ -36,14 +36,20 @@ jobs:
     # arm64-safe: only downloads coverage artifacts + runs the sonar scanner CLI
     # (arm64-available); no Go build, no containers.
     runs-on: teranode-runner-4-core-arm
-    # Run for every PR-triggered test outcome (not only success) so the sonar
-    # commit status ALWAYS resolves — a fully skipped job would leave a required
-    # PR check stuck "pending". Only pull_request runs are handled (push-to-main
-    # test runs are filtered out). The analysis steps below are individually
-    # guarded to conclusion == 'success' because they need artifacts that failed
-    # runs never produce; on a non-success run we skip straight to posting a
-    # failure status.
-    if: github.event.workflow_run.event == 'pull_request'
+    # Run for PR-triggered test outcomes other than 'cancelled' so the sonar
+    # commit status resolves instead of leaving a required check stuck "pending".
+    # 'cancelled' is excluded on purpose: teranode_pr_tests uses
+    # cancel-in-progress, so every push to a PR with in-flight tests emits a
+    # cancelled completion for the *superseded* commit. Its successor run reports
+    # on the new head, so acting here would only stamp a spurious failure.
+    # Only pull_request runs are handled (push-to-main test runs are filtered
+    # out). The analysis steps below are individually guarded to
+    # conclusion == 'success' because they need artifacts that failed runs never
+    # produce; on a non-success (but not cancelled) run we skip straight to
+    # posting a failure status.
+    if: >-
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion != 'cancelled'
 
     steps:
       - name: Get PR number and metadata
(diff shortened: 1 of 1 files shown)


ordishs (changes requested):
This is a well-designed solution to an awkward problem — the fork-safety reasoning for staying on `workflow_run` is exactly right, the resolution matrix in the description shows real care, and details like the implicit `success()` AND on custom step `if:` conditions and the fail-loud `!sha` guard are the kind of thing that usually gets missed. The architecture here doesn't need to change at all.

I'm requesting changes for one issue that I think has to land before this can be made a required check. It's a small fix — everything lives in the status script step.

## Status is posted to the current PR head, not the tested commit

The script posts to `steps.pr.outputs.head_sha`, which comes from `pulls.get` at the time *this* workflow runs — not `github.event.workflow_run.head_sha`, the commit that was actually tested and scanned. When those differ, the result for commit A gets stamped onto commit B. Two consequences:

**Spurious red on every push (the common path, not an edge case).** `teranode_pr_tests.yaml` line 8 has `cancel-in-progress: true`, so every push to a PR with in-flight tests emits a `cancelled` completion. With the job `if:` now broadened, that triggers this workflow, which posts `failure — "Upstream tests cancelled — Sonar not run"` onto the **new** head commit — whose tests are running normally. Given our test durations, that's a misleading red X sitting on most active PRs for an hour at a time. Your resolution matrix covers the conclusions correctly — the row that bites is that `cancelled` fires constantly here, and the SHA being posted to moves out from under the run.

**Stale green on unscanned code.** The inverse: gate passes on A, developer has pushed B meanwhile, `success` lands on B — a commit whose code was never scanned. For a check that's about to be made *required*, this is the one invariant it exists to provide.

Both fall out of the same fix:

1. Post the status to `context.payload.workflow_run.head_sha` (available directly in the script's payload, so the SHA no longer depends on the `pr` step either — shrinks the `!sha` failure surface).
2. Skip posting entirely when the conclusion is `cancelled` — a cancelled run is by definition superseded, and its successor will report on the new head.

## Smaller items (fine to fold into the same change or follow up)

- **`WARN` gate outcome is misreported.** `quality-gate-status` can be `PASSED`, `WARN`, or `FAILED`; the script currently lumps `WARN` into `error` / "did not complete (analysis failed)". Worth an explicit branch — `success` with a "passed with warnings" description is the usual policy, but either way the description should be accurate.
- **`always()` → `!cancelled()`** on the status step: if the sonar job itself gets cancelled by its own concurrency group, `always()` can still fire and post a transient `error`. A cancelled run should never report.
- **Hardcoded project key** (`bsv-blockchain_teranode`) in the `target_url` duplicates what's in the sonar config — fine as-is, just worth a comment noting the drift risk.

Permissions (`statuses: write` only), pinned action SHAs, and passing values via `env:` rather than interpolation all look right. Happy to re-review as soon as the SHA fix is in — the rest of this is ready.

oskarszoon:
Good catch — the SHA was wrong, and it's the one thing that had to be right before this can be required. Fixed in `6fe31228a`:

- Status now posts to `github.event.workflow_run.head_sha` (the tested commit), not the live PR head from `pulls.get`. Closes both the stale-green-on-B and the spurious-red cases.
- Went one step past the request and pinned the **checkout** to the same tested commit, so the scanned tree, the downloaded coverage, and the posted status all describe one commit — not just the status. It falls back to fetching the exact SHA and bails (`::error::`) if it's unreachable, i.e. a force-push/rebase since the run.
- Job `if:` now excludes `cancelled` upstream runs, so cancel-in-progress churn stops posting spurious failures onto the superseded commit's successor — that run reports on the new head itself.
- `always()` → `!cancelled()` on the status step, so a self-cancelled sonar run never posts a transient error.
- `WARN` gate outcome handled explicitly (`success`, "passed with warnings") instead of being lumped into error.
- Added a comment noting the `projectKey` duplication with `sonar-project.properties`.

Ready for re-review.

liam (approved):

ordishs (approved):
Every point addressed, and the fix goes further than asked: checking out the *tested* commit means the scanned tree, the downloaded coverage, and the commit status now all describe the same SHA — the analysis is coherent end-to-end, not just correctly labelled. Excluding `cancelled` at the job level is also cleaner than skipping the post step; superseded runs now do nothing at all.

Re-verified on the new diff:

- Status and checkout both keyed to `workflow_run.head_sha` — the stale-green / spurious-red window is closed.
- `WARN` handled explicitly.
- `!cancelled()` on the status step, and the status is now independent of the metadata step for its SHA — a PR-lookup failure still posts a correct state (only the `target_url` PR param goes empty, which is trivial).
- The new checkout shell: env vars quoted, values from GitHub's own API, fail-closed with a clear error when a force-push makes the tested SHA unreachable. No injection surface.
- Permission tightening checked: `pull-requests: read` covers the two `pulls.*` calls, and recent production runs confirm artifact downloads work without `actions: read`, so no regression there.

One thing to watch on the first real run: the `id-token: write` removal. Nothing in the workflow visibly uses OIDC, and if something did depend on it silently, the scan errors and the status posts `error` — fail-closed and obvious — so it's not worth holding the PR for.

Nice work — approving. Once this has a green run or two, it should be safe to flip "SonarQube Quality Gate" to required (alongside the tests check, per the description).

Issue #936, opened 2026-05-24
legacy: createUtxos calls SetMinedMulti with unbounded slice — stalls aerospike on fat blocks (regression from #854)


oskarszoon:
## Summary

`legacy/netsync/handle_block.go:createUtxos` calls `SetMinedMulti` with one unbounded slice containing every pre-existing tx in the block. On fat blocks (mainnet 755,880 = 2.87M txs, almost all pre-existing via propagation) this becomes a monolithic 2.87M-record aerospike `BatchOperate` that exhausts the client connection pool, hits `MAX_RETRIES_EXCEEDED` with `NETWORK_ERROR` / `connection reset by peer`, and stalls sync in an infinite retry loop.

Regression introduced by #854 (`fix(legacy): merge blockID into pre-existing tx in createUtxos`, merged 2026-05-13).

## Reproduction

Mainnet sync past block 755,880 (2.3 GB, 2,867,288 txs) on a single-node aerospike `.docker.m` deployment.

Observed on `bsva-ovh-teranode-eu-3` running `v0.15.2-beta-1` (commit `f44080f06`, includes #929 arena fix).

Log signature, repeating every ~30s:

```
ERROR | netsync/handle_block.go:698 | legacy| [HandleBlockDirect][000000000000000002c365bec2f13cb3ba6334ebee5a0325201464e67b7fcecc 755880] 2867288 txs, peer ... DONE in 34.5s with error: PROCESSING (4): failed to merge blockID into 2867287 pre-existing txs -> STORAGE_ERROR (69): aerospike BatchOperate error -> UNKNOWN (0): ResultCode: MAX_RETRIES_EXCEEDED, Iteration: 5, InDoubt: true, Node: A1 172.18.0.5:3000: command execution timed out on client: Exceeded number of retries.
  ResultCode: NETWORK_ERROR, ...
  write tcp 172.18.0.10:53374->172.18.0.5:3000: write: connection reset by peer
```

Block download + decode succeeds (34s). The `SetMinedMulti` merge step is where it dies. After failure, sync resets to 755,879, downloads block again, repeats.

## Root cause

`services/legacy/netsync/handle_block.go:643-700` (post-#854):

```go
var (
    existingTxsMu    sync.Mutex
    existingTxHashes []*chainhash.Hash
)

// create all the utxos first
for _, txHash := range txMap.Keys() {           // iterates every tx in the block
    g.Go(func() error {
[...]

PR #1250, opened 2026-07-08
fix(blockvalidation): port legacy hardening to quick validation (chunked SetMinedMulti #936, transient spend retry)

freemans13:
## Summary

Two hardening fixes that exist in the legacy netsync path but were missing from the quick-validation path (`quick_validate.go`), closing verified drift between the two below-checkpoint block-processing implementations:

1. **Chunked `SetMinedMulti` (#936 port).** Legacy chunks the pre-existing-tx blockID merge across a bounded worker pool because a single call with every existing tx overran the Aerospike client connection pool on fat blocks (mainnet 755880 = 2.87M txs). Quick validation still made one unchunked call — a latent fat-batch hazard on dirty-restart retries during catchup (on retry, every tx in the batch already exists). Extracted to a shared `utxo.SetMinedMultiChunked` helper used by both paths. Legacy behaviour is byte-identical (same chunk math, floors, worker fan-out; the one observable delta is the debug-log prefix, now `[SetMinedMultiChunked]`).
   - Reviewer note: the two callers deliberately keep different `MinedBlockInfo` values — legacy sets `OnLongestChain: true, SubtreeIdx: 0`, quick validation passes zero values. That divergence pre-exists this PR; the helper passes `info` through verbatim rather than silently unifying. Flagging for confirmation that quick validation's `OnLongestChain: false` is intended.
2. **Transient spend retry.** Quick validation's spend phase hard-failed on ANY spend error. Legacy (`PreValidateTransactions`) retries transient store errors (`errors.IsRetryableError`, maxRetries=10, 2s backoff) with no-progress give-up. Ported the same retry semantics into a new `spendBatchWithRetry` so both paths converge identically after dirty restarts and under transient store overload (e.g. Aerospike DEVICE_OVERLOAD).
   - **Deliberately NOT ported: legacy's `ErrTxConflicting` tolerance.** Consensus review during development established that legacy's tolerance is only safe because legacy runs the validator with `WithCreateConflicting`, so block assembly's `ProcessConflicting` reconciles the loser at block acceptance. The quick path never writes conflicting subtree nodes and has no such resolver — tolerating there would leave an output's spend permanently attributed to a non-canonical tx. Quick validation therefore fails closed on conflicts (and dirty-restart replay needs no tolerance: re-spending with the same spender is the store's idempotent success path). The comments in the code document this asymmetry.

No new flags, no consensus-rule changes — error-handling parity where safe, fail-closed where the legacy mechanism doesn't exist.

## Tests

- `TestSetMinedMultiChunked` — chunk-size ceiling, exactly-once hash coverage, floors, error propagation, caller-info passthrough (spy store proves the helper touches nothing but SetMinedMulti)
- `TestSpendBatchWithRetry` — clean path / transient-converges / conflicting-fails-hard-never-retried / non-retryable-hard-fail / no-progress-give-up
- Full `stores/utxo`, `services/legacy/netsync`, `services/blockvalidation`, `catchup` packages green incl. `-race`; `go vet` clean; lint findings limited to pre-existing untouched files

### DIFF
--- a/services/blockvalidation/BlockValidation.go
+++ b/services/blockvalidation/BlockValidation.go
@@ -241,6 +241,10 @@ type BlockValidation struct {
 
 	// mmapDir, when non-empty, enables mmap-backed subtree loading.
 	mmapDir string
+
+	// spendRetryBackoff overrides the pause between quick-validate spend retry
+	// attempts; zero means spendRetryBackoffDefault. Settable in tests.
+	spendRetryBackoff time.Duration
 }
 
 // subtreeFromBytesWithMmap creates a subtree from bytes, using mmap if dir is non-empty.
--- a/services/blockvalidation/quick_validate.go
+++ b/services/blockvalidation/quick_validate.go
@@ -1284,9 +1284,13 @@ func (u *BlockValidation) createAndSpendUTXOsForBatch(ctx context.Context, block
 	}
 
 	// Phase 1.5: Update mined info for transactions that already existed
-	// This handles the case where a previous attempt created UTXOs with a different block ID
+	// This handles the case where a previous attempt created UTXOs with a different
+	// block ID. Chunked via the shared helper (issue 936): on a fat-batch retry every tx
+	// in the batch already exists, and a single unchunked SetMinedMulti call would
+	// overrun the aerospike client connection pool.
 	if len(existingTxHashes) > 0 {
-		if _, err := u.utxoStore.SetMinedMulti(ctx, existingTxHashes, minedBlockInfo); err != nil {
+		if err := utxo.SetMinedMultiChunked(ctx, u.logger, u.utxoStore, existingTxHashes, minedBlockInfo,
+			u.settings.UtxoStore.MaxMinedBatchSize, u.settings.UtxoStore.MaxMinedRoutines); err != nil {
 			return errors.NewProcessingError("[createAndSpendUTXOsForBatch][%s] failed to update mined info for %d existing txs", block.Hash().String(), len(existingTxHashes), err)
 		}
 	}
--- a/services/blockvalidation/quick_validate.go
+++ b/services/blockvalidation/quick_validate.go
@@ -1295,21 +1295,99 @@ func (u *BlockValidation) createAndSpendUTXOsForBatch(ctx context.Context, block
 		}
 	}
 
-	// Phase 2: Spend all transactions in parallel
-	spendG, spendCtx := errgroup.WithContext(ctx)
-	util.SafeSetLimit(u.logger, spendG, u.settings.UtxoStore.SpendBatcherSize*u.settings.UtxoStore.SpendBatcherConcurrency*2)
+	// Phase 2: Spend all transactions, with the same hardening the legacy path has
+	// (services/legacy/netsync PreValidateTransactions): tolerate ErrTxConflicting
+	// (the block is checkpoint-certified, so its txs take precedence and conflicts
+	// are resolved at block acceptance), retry transient store errors with backoff
+	// and progress detection, hard-fail everything else.
+	return u.spendBatchWithRetry(ctx, block, batch.batchTxs, outpointOnly)
+}
 
-	for _, tx := range batch.batchTxs {
-		tx := tx
-		spendG.Go(func() error {
-			if _, err := u.utxoStore.Spend(spendCtx, tx, block.Height, utxo.IgnoreFlags{IgnoreLocked: true, SkipUTXOHashCheck: outpointOnly}); err != nil {
-				return errors.NewProcessingError("[createAndSpendUTXOsForBatch][%s] failed to spend tx %s", block.Hash().String(), tx.TxIDChainHash().String(), err)
-			}
-			return nil
-		})
+// spendRetryBackoffDefault is the pause between spend retry attempts. Matches the
+// legacy path's retryBackoff (services/legacy/netsync PreValidateTransactions).
+const spendRetryBackoffDefault = 2 * time.Second
+
+// spendBatchWithRetry spends txs in parallel with bounded retries. Per attempt:
+// ErrTxConflicting is tolerated (block txs take precedence below checkpoint-certified
+// blocks; resolution happens at block acceptance), retryable errors (transient store
+// overload) queue the tx for the next attempt, anything else fails hard immediately.
+// Gives up early when an attempt makes no progress. Mirrors the legacy path's
+// PreValidateTransactions retry semantics (maxRetries=10, 2s backoff) so both
+// below-checkpoint implementations converge identically after dirty restarts.
+func (u *BlockValidation) spendBatchWithRetry(ctx context.Context, block *model.Block, txs []*bt.Tx, outpointOnly bool) error {
+	const maxRetries = 10
+
+	backoff := u.spendRetryBackoff
+	if backoff <= 0 {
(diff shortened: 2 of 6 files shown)


ctnguyen (approved):
The extraction is faithful: `SetMinedMultiChunked` reproduces the legacy inline chunk math, worker fan-out, floors, and error wrap verbatim, and both call sites are updated correctly with the only observable delta being the debug-log prefix. `spendBatchWithRetry` is a byte-faithful reproduction of the legacy `PreValidateTransactions` retry loop (maxRetries=10, 2s backoff, no-progress give-up, retryable-vs-hard-fail classification via `IsRetryableError`), with a clean test-only backoff seam. The single most important design decision — failing closed on conflicts rather than copying legacy's tolerance — is correct and clearly reasoned: the quick path has no `ProcessConflicting` resolver, and same-spender dirty-restart replay is the store's idempotent success path, so no legitimate conflict tolerance is lost. Tests cover the meaningful branches (chunk ceiling / exactly-once / floors / error propagation; clean / transient-converges / conflict-hard-fail / non-retryable / no-progress), and the PR description honestly flags the pre-existing `MinedBlockInfo` divergence rather than silently unifying it.

## Non-blocking issues

### ChiR1 — Existing-tx merge passes SubtreeIdx=0 / OnLongestChain=false in quick validation

**Problem:** The quick path builds `minedBlockInfo` with only `BlockID`/`BlockHeight` set (`SubtreeIdx=0`, `OnLongestChain=false`) and passes it into `SetMinedMultiChunked` for pre-existing txs, whereas legacy passes the real values (`OnLongestChain=true`, and the fresh `Create` above writes the true `SubtreeIdx=sIdx`). On Aerospike, `SetMinedMulti` *appends* `SubtreeIdx` to the `subtreeIdxs` list (set_mined_expressions.go:285), so a dirty-restart retry adds a spurious `0` entry rather than overwriting; `OnLongestChain=false` also skips the delete-at-height retention branch that legacy's `true` takes (set_mined_expressions.go:293). This value divergence is **pre-existing** — the old unchunked `SetMinedMulti` call passed the identical `minedBlockInfo` — and is not introduced by this PR; the helper passes `info` through verbatim by design.

**Why it matters:** If `OnLongestChain` or the `subtreeIdxs` list is consumed for anything consensus-relevant on these below-checkpoint merged txs (e.g. the next block's `validOrderAndBlessed` / `getParentTxMetaBlockIDs` path, or DAH-based pruning), the two below-checkpoint implementations would disagree. The blockID merge itself — the thing that check actually needs — is performed identically, so this is most likely inert, but it is worth confirming rather than assuming.

**Fix:** No code change required for this PR. As a follow-up, confirm `SubtreeIdx`/`OnLongestChain` are inert for existing-tx merges on the quick path; if `OnLongestChain` matters downstream, align quick validation with legacy in a separate change:

```go
minedBlockInfo := utxo.MinedBlockInfo{
    BlockID:        block.ID,
    BlockHeight:    block.Height,
    OnLongestChain: true, // match legacy if downstream relies on it
}
```

### ChiR2 — Hard-fail no longer promptly aborts sibling spends in the same attempt

**Problem:** In `spendBatchWithRetry`, every goroutine returns `nil` and records outcomes under a mutex; a non-retryable error only sets `hardFail`. Because no goroutine ever returns an error, the errgroup's derived `spendCtx` is never cancelled, so on a hard-fail all other in-flight/queued `Spend` calls in that attempt run to completion before `spendG.Wait()` returns and `hardFail` is checked. The previous quick-path code returned the error directly from the goroutine, which cancelled `spendCtx` and aborted the remaining spends promptly.

**Why it matters:** On a fat block (the 2.87M-tx case this PR targets), a single non-retryable spend error now lets the entire remaining batch complete against the store before the block fails — more wasted work on the failure path than before. It is not a correctness bug (the block fails regardless), and it intentionally mirrors legacy `PreValidateTransactions`, so it is not blocking.

**Fix:** Optional. If faster abort is desired, have the goroutine return the non-retryable error (or cancel a local context) to short-circuit the errgroup while still recording retryable txs under the mutex. Given the deliberate legacy-parity goal, leaving it as-is with a one-line comment noting the trade-off is also acceptable.

### ChiR3 — Retries-exhausted give-up message omits the underlying store cause

**Problem:** The post-loop return `"[spendBatchWithRetry][%s] %d of %d spends still failing after %d retries"` does not include `lastErr`, so when the loop exhausts all 10 retries while still making incremental progress each attempt, the actual store error driving the failures is not surfaced. (The no-progress give-up branch does include `lastErr`; only the retries-exhausted branch omits it.)

**Why it matters:** Diagnostic only. An operator debugging a block that failed after ~20s of retries under sustained store overload would not see the root storage error in this message. Legacy `PreValidateTransactions` (handle_block.go:1300) has the identical omission, so this is a pre-existing observability nit carried over, not a regression.

**Fix:** Append `lastErr` to the final return:

```go
return errors.NewProcessingError("[spendBatchWithRetry][%s] %d of %d spends still failing after %d retries",
    block.Hash().String(), len(pending), total, maxRetries, lastErr)
```

## Recap

| ChiR1 | Existing-tx merge SubtreeIdx/OnLongestChain divergence | | 22% |
| ChiR2 | Hard-fail no longer aborts sibling spends | | 16% |
| ChiR3 | Give-up message drops underlying cause | | 10% |

ordishs (approved):
Approving. Verified the helper extraction against the legacy source line-by-line (chunk math, floors, worker fan-out, error wrap are byte-identical; only the debug-log prefix changed), and the retry loop is a faithful port of `PreValidateTransactions` (same attempt structure, no-progress give-up, final error). The fail-closed decision on `ErrTxConflicting` is the right consensus-safety call given the quick path has no `ProcessConflicting` reconciliation. Both new test suites pass with `-race` locally.

A few non-blocking comments:

1. **Tautological assertion** in `quick_validate_spend_retry_test.go` (conflicting subtest): `errors.Is(err, errors.ErrTxConflicting) || errors.Is(err, errors.ErrProcessing)` — the hard-fail path always wraps in `NewProcessingError`, so the second disjunct is always true and the assertion can never fail. The `spendCalls == 3` check is what actually protects the never-retried behaviour. Suggest tightening to `require.ErrorIs(t, err, errors.ErrTxConflicting)` if the wrap chain preserves it, or dropping the `ErrProcessing` disjunct.

2. **Double error wrapping** in `createAndSpendUTXOsForBatch` phase 1.5: the helper already wraps store failures in "failed to merge blockID into %d pre-existing txs" and the caller wraps that again. Cosmetic only — the legacy caller now returns the helper error bare.

3. **Hard fail no longer short-circuits the attempt**: the old quick path cancelled in-flight spends via the errgroup on first error; the new loop lets the whole attempt's spends complete before returning the hard failure. On a multi-million-tx block with an early non-retryable failure that's wasted work — but it exactly matches legacy behaviour, so it's consistent with the parity goal. Flagging so it's a conscious change.

4. **`time.Sleep(backoff)` doesn't respect ctx cancellation** during the 2s pause — detected only on the next attempt when spends fail with (non-retryable) context errors, so it still terminates promptly. Legacy has the same shape; fine to leave or fix in both later.

5. **`OnLongestChain: false` divergence** — thanks for flagging rather than silently unifying. This deserves an explicit answer from whoever owns the quick path: if quick-validated blocks are by definition on the checkpointed canonical chain, `OnLongestChain: false` may itself be a latent bug. Worth a follow-up issue if confirmation doesn't land here.

PR #1251, opened 2026-07-08
refactor: collapse below-checkpoint gate predicates into model + unify subtree DAH retention (stacked on #1226)

freemans13:
## Summary

**Stacked on #1226** — review only the two commits after its head (`1391e76c5`, `38c7c4968`); the diff shrinks to those once #1226 merges.

Drift-repair for the below-checkpoint fast path. Four non-identical gate formulations (netsync `legacyOutpointOnly`/`quickValidationAllowed`, blockvalidation `quickValidateOutpointOnly`/`quickValidateSkipsUtxoLock`, model `skipOrderAndBlessedBelowCheckpoint`) collapse onto two shared pure functions in `model` (bottom of the import graph): `BelowCheckpoint` (the boundary: `highest > 0 && height > 0 && height <= highest`) and `OutpointOnlyEligible` (flag + store capability + nil-params guard + boundary). Chain params are passed explicitly so each caller keeps its provenance (netsync: `sm.chainParams`; others: `settings.ChainCfgParams`). The model gate keeps its extra `checkpointConfirmedAncestor` conjunct layered on top; the `merkleRootChecked` call-site precondition in `Block.Valid` is untouched. catchup's `useQuickValidation` is deliberately out of scope (contextual checkpoint-hash verification, not a formulation of this gate), as are the difficulty-skip and fee-skip gates (different semantics).

**Declared fail-closed tightenings** — the strictest existing conjuncts now apply everywhere; verified in review that **no gate opens wider for any input**:
1. **Height 0 no longer fast-paths** in the netsync/blockvalidation gates (model already excluded it). Unreachable in real flows: `HandleBlockDirect` derives height as parent+1 (min 1), and catchup never dispatches genesis. Two test truth-table rows flipped accordingly, with justification comments.
2. The blockvalidation gates gain **nil-`ChainCfgParams` / no-checkpoints guards** (previously a latent nil dereference, and height 0 could pass when no checkpoints were defined).
3. netsync subtree-file DAH switches from raw `GlobalBlockHeightRetention` to **`GetSubtreeValidationBlockHeightRetention()`** — the source `quick_validate.go` and `get_blocks.go` already use (all three `writeSubtree` sites; the storage-mode window in `peer_server.go` correctly stays on the global value). Identical at the default adjustment of 0; a configured `SubtreeValidation.BlockHeightRetentionAdjustment` now applies to subtree files consistently on every path.
4. **Nil UTXO store → false instead of panic** in the shared gate (fail-safe; reachable only in minimal test constructions).

## Tests

- `TestBelowCheckpoint` (8 cases) + `TestOutpointOnlyEligible` (10 cases incl. nil settings/store/params)
- Existing gate truth tables updated only where behaviour deliberately changed (the two height-0 rows, RED-confirmed before the change)
- Full `model`, `netsync`, `blockvalidation`, `catchup` packages green incl. `-race` on netsync; vet clean; lint findings limited to pre-existing untouched files


ctnguyen (approved):
This is careful, well-reasoned work. The two shared predicates sit correctly at the bottom of the import graph, each collapsed call site keeps its own chain-params provenance, and the fail-closed tightenings (height-0 exclusion, nil-`ChainCfgParams`/no-checkpoints guards, nil-store → `false` instead of panic) only ever narrow the gates. The safety argument for skipping `validOrderAndBlessed` matches the code: the integrity floor (PoW at step 1, `checkDuplicateTransactions` at step 11) is genuinely unconditional, and the `merkleRootChecked` precondition ties the skip to a header binding that a nil-subtree caller can never satisfy. Test coverage is excellent — full truth tables for both predicates, the merkle-root-not-checked regression guard, the tampered-merkle floor test, and the flag-off revalidation asymmetry test all exercise real `Block.Valid` behavior rather than mocks.

## Non-blocking issues

### ChiR1 — Consensus-adjacent behavior changes presented as a pure refactor

**Problem:** Despite the "refactor / collapse predicates" framing and the claim that "no gate opens wider for any input," the diff introduces two genuinely new runtime behaviors, neither of which exists on the base:

1. `model/Block.go` step 12 changes from an *unconditional* `validOrderAndBlessed` to a conditional skip gated on the newly-added `skipOrderAndBlessedBelowCheckpoint` (plus the `merkleRootChecked` precondition). On the below-checkpoint outpoint-only fast path this skips `validOrderAndBlessed` entirely — and therefore also the downstream `checkOldBlockIDs`, which only ever inspects the `oldBlockIDsMap` that `validOrderAndBlessed` populates.
2. `services/legacy/netsync/handle_block.go` replaces the unconditional `blockHeight > 1` parent-mined wait with `needsParentMinedWait`, removing `waitForPreviousBlockMined` on the fast path.

The description also lists `model.skipOrderAndBlessedBelowCheckpoint` as one of four *pre-existing* "gate formulations" being collapsed, but it does not exist on the base — this PR adds it.

**Why it matters:** These are functional changes on a consensus-adjacent path, not a behavior-preserving consolidation. I independently confirmed both are sound: the skip rests on the `merkleRootChecked` header binding, the unconditional PoW/`checkDuplicateTransactions` floor, and `checkpointConfirmedAncestor` (set only by blockvalidation, default `false`); below a confirmed checkpoint the referenced-old-block set is already certified by the pinned chain, so skipping `checkOldBlockIDs` is safe. The parent-mined wait is redundant because block dispatch through `blockQueue` → `handleBlockMsg` → `HandleBlockDirect` is serial, the fast path commits UTXOs with mined info at creation (`WithMinedSet(true)`, `setTxMined` skipped), and BIP68 is not evaluated below the checkpoint. But a reviewer trusting the "no behavior change" framing may under-scrutinize the two newly-enabled skips.

**Fix:** No code change required. Re-title/annotate so the new below-checkpoint `validOrderAndBlessed` skip (and its `checkOldBlockIDs` consequence) and the `needsParentMinedWait` removal are signed off as functional changes, and correct the description's claim that `model.skipOrderAndBlessedBelowCheckpoint` pre-existed. For example, drop it from the "four formulations collapsed" list and call it out under a dedicated "new behavior" heading alongside the netsync wait removal.

## Recap

| ChiR1 | New skips framed as a refactor | | 22% |

ordishs (approved):
Reviewed the two stacked commits (1391e76c5, 38c7c4968). Verified independently that no gate opens wider for any input — the only behavioural deltas are the four declared fail-closed tightenings (height 0 excluded, nil chain-params and nil-store guards, no-checkpoints at height 0). Gate truth tables in model, netsync and blockvalidation all pass locally; build and vet clean. DAH retention switch is byte-identical at the default adjustment.

Non-blocking follow-ups:
- `checkpointConfirmedAncestor` (BlockValidation.go:2172) still re-derives the boundary comparison; switching to `!model.BelowCheckpoint(...)` would honour the new single-definition invariant and skip two gRPC lookups for height-0 blocks.
- The out-of-scope difficulty-skip gates (BlockValidation.go:1614 and reValidateBlock) still dereference `ChainCfgParams` without a nil guard — same latent panic this PR fixed in the sibling gates; worth a follow-up issue.
- The subtree DAH change means a non-zero `subtreevalidation_blockHeightRetentionAdjustment` now applies to netsync-written subtree files too — deserves a release-note line for operators who tuned that setting.

PR #1252, opened 2026-07-08
feat: flag-gated unified below-checkpoint route — legacy blocks through quick validation (stacked on #1251)

freemans13:
## Summary

**Stacked on #1251** (→ #1226 → merged #1178). Review only the commits after #1251's head; the diff shrinks as the stack merges. **Default off** — with the flag off, behaviour is byte-identical.

First structural step of the legacy-IBD unification: behind `blockvalidation_legacy_unified_below_checkpoint`, legacy netsync blocks **at or below the highest hardcoded checkpoint** are processed by the **same `quickValidateBlock` machinery native catchup already uses**, instead of legacy's own inline UTXO pipeline. This removes a duplicated below-checkpoint implementation (the drift source behind two prior consensus bugs).

**Two seams:**
1. **Server-side route** (`Server.go` `processBlockFound`, after block-assembly-ready, before `ValidateBlockOptions`): when the flag is on, `baseURL=="legacy"`, and the shared `model.OutpointOnlyEligible` gate holds (hardcoded checkpoints only — never the operator override), call `quickValidateBlock` instead of `ValidateBlockWithOptions`.
2. **Netsync adapter** (`prepareSubtrees`): under the flag, skip blockID assignment (server assigns/reuses idempotently; blockID 0 on the wire = "assign server-side") and skip the UTXO create+spend (`ValidateTransactionsLegacyMode`) — the server does them. Still partitions the wire block, builds the txMap, extends in-block parents, and writes the `.subtree`/`.subtreedata`/`.subtreemeta` files.

**Integrity floor preserved (consensus-critical).** `block.Valid()` no longer runs on this route, and — unlike catchup, whose subtree lists arrive bound to checkpoint-verified headers — legacy builds its own subtrees from the wire block. So `HandleBlockDirect` explicitly verifies, before create+spend, on every multi-tx unified block:
- **PoW** (`HasMetTargetDifficulty`, pre-existing),
- **`CheckMerkleRoot`** (binds the netsync-built subtrees to the PoW-verified, checkpoint-pinned header),
- **`CheckSubtreeSlicesForDuplicateTxs`** — a new shared `model` helper restoring the **CVE-2012-2459** duplicate-transaction floor. This is load-bearing, not defence-in-depth: a test (`TestCVE20122459_MerklePassesDedupCatches`) proves the duplicate-last-when-odd mutation leaves the merkle root **unchanged** (so `CheckMerkleRoot` passes) and the dedup check is the only thing that rejects it.

Header-chain checkpoint verification (headers-first) + parent-exists induction certify the block is on the checkpoint chain before any UTXO commit.

## Consensus review

A full bitcoin-expert consensus review was run against the whole branch (spec §4). It returned one MUST-FIX — the dropped CVE-2012-2459 dedup floor — now fixed and proven above. All other items (single-tx-block merkle handling, validator fail-closed/I4 guard for legacy-routed blocks, CatchupContext-free eligibility, and the declared effect deltas) were ruled SAFE, each matching existing native-catchup behaviour.

**Declared effect deltas vs the inline legacy path** (all existing catchup-path behaviour, ruled operational/non-consensus): no `lastValidatedBlocks` cache entry, no `txStore` coinbase blob, no `computeAndSetCoinbaseBUMP`, `AddBlock` carries `WithSubtreesSet(true)`, no `ReValidateBlock`/`storeInvalidBlock` retry (a gRPC error drops the block; legacy re-delivers).

## Gates before enabling on mainnet (flag is default-off, so not PR blockers)

1. **PR #1250 (quick-validate hardening: chunked `SetMinedMulti`, transient-retry) must land first** — the unified route inherits catchup's create+spend, whose retry path is unhardened until #1250.
2. **Resolve the `.subtreemeta` / extended-subtreeData downstream-consumer question** (spec open item) before the Step-4 mainnet soak.
3. Separately: the consensus review noted native catchup *also* lacks the CVE-2012-2459 dedup below checkpoint (pre-existing, not introduced here) — worth a follow-up.

## Tests

- `TestServer_legacyUnifiedRoute` (8-row routing truth table)
- `TestSyncManager_legacyUnified` + `TestSyncManager_prepareSubtrees_UnifiedSkipsUTXOOps` (adapter: blockID 0, zero UTXO ops, files still written)
- `TestUnifiedRouteParity` (inline vs unified → identical UTXO-store state: block-group + spender-identity cross-run assertions)
- `CheckSubtreeSlicesForDuplicateTxs` unit tests + `TestCVE20122459_MerklePassesDedupCatches` (the load-bearing proof)
- Full `model`, `netsync`, `blockvalidation`, `catchup` green incl. `-race`; vet clean; lint findings pre-existing files only


ctnguyen (approved):
The consensus-safety reasoning is unusually rigorous and, importantly, matches the actual code. The CVE-2012-2459 dedup floor (`CheckSubtreeSlicesForDuplicateTxs`) runs on netsync while `SubtreeSlices` is still populated and before `ProcessBlock`, closing the exact gap `CheckMerkleRoot` cannot catch (proven by `TestCVE20122459_MerklePassesDedupCatches`), and the `merkleRootChecked` precondition on the `Block.Valid` step-12 skip is a correct fail-closed hardening for nil-subtreeStore callers. I independently confirmed the load-bearing assumptions against the tree: the gRPC `ProcessBlock` handler calls `processBlockFound` → `quickValidateBlock` synchronously, and legacy block dispatch is genuinely serial (a single `go sm.blockHandler()` goroutine), so deferring UTXO create+spend server-side and skipping `needsParentMinedWait` on the fast path is safe — the parent's UTXOs are committed before the child begins. Gating is centralized into shared, fail-safe helpers (`BelowCheckpoint`/`OutpointOnlyEligible`) with nil-param handling and exhaustive truth-table tests, and the cross-run UTXO parity harness is honest about its fidelity gaps.

## Non-blocking issues

### ChiR1 — Unified route relies on two services agreeing on config, with no cross-service assertion

**Problem:** The routing decision is made independently by two microservices keyed off the same-named settings and store capability. Netsync's `SyncManager.legacyUnified` (`services/legacy/netsync/handle_block.go`) decides to skip UTXO create+spend and send `blockID 0`; blockvalidation's `Server.legacyUnifiedRoute` (`services/blockvalidation/Server.go:1158`) independently decides to route to `quickValidateBlock`. Both derive from `blockvalidation_legacy_unified_below_checkpoint`, `blockvalidation_outpoint_only_below_checkpoint`, `ChainCfgParams`, and `utxoStore.SupportsOutpointOnlySpend()` via `model.OutpointOnlyEligible`. If the two deployments diverge (different flag value, or UTXO stores whose `SupportsOutpointOnlySpend` differs), the seams split: e.g. netsync runs the full inline create+spend and assigns a non-zero blockID, while the server *also* routes to `quickValidateBlock` and repeats create+spend — a double UTXO commit — or, in the other direction, netsync defers everything (blockID 0, outpoint-only-shaped subtree files) while the server takes the full `ValidateBlockWithOptions` path against those minimal files.

**Why it matters:** In a normal single-network deployment both services resolve the same settings context and share one UTXO store, so they agree, and the divergent failure mode is a rejected/errored block (fail-safe; legacy re-delivers) rather than silent UTXO corruption. But this is a consensus-adjacent path where the two seams must be enabled together, and nothing in the code or docs makes the coupling explicit or observable.

**Fix:** Document in the setting `longdesc` (and an operator note) that `blockvalidation_legacy_unified_below_checkpoint` and `blockvalidation_outpoint_only_below_checkpoint` must be set identically on the legacy/netsync and blockvalidation deployments, and add a one-line startup log on each service reporting resolved unified-route eligibility so a mismatch is observable, e.g.:

```go
u.logger.Infof("[blockvalidation] unified-below-checkpoint route eligible=%v (legacyUnified=%v outpointOnly=%v storeSupports=%v)",
    eligible, u.settings.BlockValidation.LegacyUnifiedBelowCheckpoint,
    u.settings.BlockValidation.OutpointOnlyBelowCheckpoint, u.utxoStore.SupportsOutpointOnlySpend())
```

### ChiR2 — Unconditional subtree-file DAH change weakens the "byte-identical when off" guarantee

**Problem:** `writeSubtree`'s three DAH computations change from `bi.height + sm.settings.GlobalBlockHeightRetention` to `bi.height + sm.settings.GetSubtreeValidationBlockHeightRetention()` (`services/legacy/netsync/handle_block.go`, ~lines 574/617/676 in base). This is in the *unconditional* `writeSubtree` path — it applies to every legacy block regardless of the unified/outpoint-only flags, including the default flag-off legacy sync path.

**Why it matters:** `GetSubtreeValidationBlockHeightRetention()` equals `GlobalBlockHeightRetention` only when `subtreevalidation_blockHeightRetentionAdjustment == 0` (the default), so for default deployments this is a no-op and the PR's "byte-identical when off" claim holds. But an operator running a non-zero (especially negative) adjustment now gets a different subtree-file Delete-At-Height on the legacy path than before — a real, if minor, behavior change on the default path that the PR description's "byte-identical off" claim does not call out.

**Fix:** Keep the consistency fix (aligning legacy subtree files to the subtree-validation retention is defensible and matches `quick_validate.go`/`get_blocks.go`), but note in the PR description that flag-off behavior changes for operators with a non-zero `subtreevalidation_blockHeightRetentionAdjustment`. If strict off-path invariance is required instead, gate the retention change behind the unified/outpoint-only path like the other adapter changes.

## Recap

| ChiR1 | Cross-service unified-route config coupling not asserted/documented | | 30% |
| ChiR2 | Unconditional subtree DAH change breaks "byte-identical off" | | 16% |

ordishs (approved):
Approving — the flag-off default is genuinely safe, the routing gates are fail-safe conjunctions with exhaustive truth-table coverage, and the CVE-2012-2459 work is unusually well-evidenced (`TestCVE20122459_MerklePassesDedupCatches` proving merkle-passes/dedup-catches on the same mutated block is exactly the right test). I verified the load-bearing claims against the branch: the `ProcessBlock` → `processBlockFound` → `quickValidateBlock` chain is fully synchronous (so the parent-mined-wait skip's serial-dispatch argument holds), `processBlockSubtreesSequential` reads the subtree files netsync already wrote (the `baseURL="legacy"` sentinel is never dereferenced), and the server-side blockID reuse in `quick_validate.go` replaces the skipped `reuseBlockIDFromUTXO`. All new tests pass locally at the PR head.

Comments, roughly in priority order:

1. **Coinbase-only blocks on the unified route are never bound to the header.** `prepareSubtrees` returns nil slices for `txCount <= 1`, so the `preparedSubtreeSlices != nil` floor in `HandleBlockDirect` is skipped, and `quickValidateBlock` takes the no-subtrees branch with no merkle check either — nothing verifies `coinbaseTx.TxID == header.HashMerkleRoot`, so an eclipse-capable legacy peer could substitute a coinbase body under an authentic checkpoint-chain header. The gap is arguably pre-existing (flag-off `Block.Valid` also skips `CheckMerkleRoot` when there are no subtrees), but the fix here is one line: `CheckMerkleRoot`'s zero-subtree `default:` branch already compares the coinbase TXID against `HashMerkleRoot`, so calling it unconditionally on the unified route closes the gap for free. Worth doing in this PR (or the immediate follow-up), with the flag-off/full-validation equivalent filed alongside the already-noted native-catchup dedup follow-up.

2. **Two changes are not gated by the new flag**, which slightly contradicts the "flag off = byte-identical" framing:
   - `needsParentMinedWait` skips the wait under `legacyOutpointOnly` alone (unified flag off), retroactively changing the semantics of the flag shipped in the earlier stacked PR. The redundancy argument is sound, but if intentional it should be stated in the PR body; otherwise gate it on `legacyUnified`.
   - `writeSubtree` now uses `GetSubtreeValidationBlockHeightRetention()` unconditionally. Default adjustment 0 keeps defaults identical, but operators with `subtreevalidation_blockHeightRetentionAdjustment != 0` get different legacy-path subtree DAHs with all flags off. Right consistency fix (both writers of the same files must agree) — just declare the delta.

3. **`TestHandleBlockDirect_UnifiedRoute_DeduplicatesViaPreparedSlices` doesn't test what its name claims.** It never invokes `HandleBlockDirect`; it calls the model helper directly, duplicating `TestUnifiedRouteDedup_DuplicateTxRejected`, and the `_ = context.Background()` to keep an unused import is the tell. Rename/remove it, or replace with a real integration test — the actual wiring assurance today is code inspection plus the Step-4 soak, and the suite shouldn't suggest otherwise.

4. **Dedup map memory on large below-checkpoint blocks.** `CheckSubtreeSlicesForDuplicateTxs` builds one flat `map[chainhash.Hash]struct{}` over the whole block; BSV mainnet below a recent checkpoint includes multi-GB blocks with millions of txs, so the map can transiently reach hundreds of MB (this is why `Block.Valid`'s equivalent went pooled/disk-backed). Netsync already holds `txMap` at the same cardinality so it's not a new order of magnitude, but measure it during the soak — and consider whether `txMap` (which already contains every non-coinbase hash) can serve the dedup instead of a second map.

Minor:
- `sm.legacyUnified(bi.height)` is evaluated twice in `prepareSubtrees`; compute once alongside `outpointOnly`.
- The parity test's Run B hand-rolls `Create`/`Spend` rather than driving `createAndSpendUTXOsForBatch` — the in-file scope-reduction note is admirably explicit, but the PR body's "identical UTXO-store state" slightly oversells it. Also `makeSpendValidator` hardcodes height 500 for all three blocks.
- A sentence on `legacyUnifiedRoute` noting that the route's integrity checks (PoW/merkle/dedup) live client-side in `HandleBlockDirect` would help future auditors of the intra-cluster trust boundary.
- Several comments narrate the review process ("RED before the gate exists") rather than stating constraints; the `merkleRootChecked` and CVE rationale comments are the ones worth keeping long-term.

None of these block merge given the flag is default-off; items 1 and 3 would ideally land before the flag is ever enabled, and items 2 and 4 need only PR-description/soak-plan acknowledgment.

PR #1253, opened 2026-07-08
refactor(blockvalidation): extract shared commitBlock (Step 8 groundwork, stacked on #1252)

freemans13:
## Summary

**Stacked on #1252.** First increment of the Step-8 parallel window: a pure, behaviour-identical extraction of the commit tail shared by `quickValidateBlock` and `quickValidateBlockAsync` into one `commitBlock` method.

- De-duplicates the identical tail: `AddBlock` (`WithSubtreesSet`+`WithMinedSet`+`WithID`) → `unlockSubtreeTransactionsIfNeeded` → `updateSubtreesDAH` → `SetBlockExists`.
- Establishes the per-block commit unit the window's height-ordered committer will call (Increment 2).
- No concurrency, no create/spend change. The only behavioural delta is a stray `[ValidateBlock]` log prefix on one sync error-log line, corrected to `[quickValidateBlock]`.

## Tests

- `TestCommitBlock_AddsBlockAndSetsExists` — drives `commitBlock` on a real in-memory `blockchain.LocalClient` (sqlitememory), asserts the block is committed and mined via genuine `GetBlockExists`/`GetBlockIsMined` queries (not mocks).
- Full `blockvalidation` + `catchup` packages green incl. `-race` — extraction is behaviour-identical.

## Context

Groundwork for the below-checkpoint parallel window (Step 8): a memory-budgeted K-block window (mass-create ∥ → mass-spend ∥ → ordered-commit) to fill the ~80%-idle box during IBD. `commitBlock` is the per-block unit the ordered committer invokes in height order.


ctnguyen (approved):
Review at commit [518253e1234c](https://github.com/bsv-blockchain/teranode/commit/518253e1234c)

The `commitBlock` extraction is genuinely behaviour-identical: both call sites previously held byte-for-byte duplicated tails and now share one method, with a `caller` label threaded through so each log line keeps its original prefix — and it incidentally corrects a stray `[ValidateBlock]` prefix in `quickValidateBlock`. It is backed by a real test that drives a live in-memory `blockchain.LocalClient` (sqlitememory) and asserts via genuine `GetBlockExists`/`GetBlockIsMined` rather than mocks. The stacked below-checkpoint work is defensively engineered: every gate (`OutpointOnlyEligible`, `BelowCheckpoint`, `skipOrderAndBlessedBelowCheckpoint`, `legacyUnifiedRoute`) is fail-closed on nil settings/params/store, the CVE-2012-2459 dedup floor and PoW remain unconditional, and the `merkleRootChecked` precondition correctly blocks the order/blessed skip on a caller that never bound the body to the header. Test coverage is exceptional — including a discriminator test proving `CheckMerkleRoot` alone cannot catch the duplicate-last-when-odd mutation while `CheckSubtreeSlicesForDuplicateTxs` does.

## Non-blocking issues

### ChiR1 — `needsParentMinedWait` removal rests on cross-file invariants

**Problem:** `needsParentMinedWait` (services/legacy/netsync/handle_block.go) drops the parent-mined wait for the entire outpoint-only path — `height > 1 && !legacyOutpointOnly(height)` — i.e. on both the inline and unified below-checkpoint routes whenever the flag is on, not just the new unified route. Its safety rests on three invariants documented only in a comment: (1) BIP68 is skipped below the checkpoint; (2) block dispatch is serial; (3) the parent's UTXO create+spend commits before the child begins. I confirmed all three hold today — manager.go starts exactly one `blockHandler` goroutine (`go sm.blockHandler()` at manager.go:2372) draining a single `blockQueue` into `handleBlockMsg`/`HandleBlockDirect` in order, and dispatch is synchronous — so there is no live defect.

**Why it matters:** The removed wait is load-bearing for UTXO correctness, yet the only thing keeping a child below-checkpoint block from starting before its parent's UTXOs are committed is prose spanning three files. If block dispatch is ever parallelized, the guard silently disappears.

**Fix:** Add a regression test pinning the serial-dispatch invariant (e.g. assert exactly one `blockHandler` is started, or a parent-before-child ordering test), and cross-reference `needsParentMinedWait` from `blockHandler` so a future refactor that parallelizes dispatch is forced to revisit the wait.

### ChiR2 — Subtree-file retention change is unconditional, not flag-gated

**Problem:** `writeSubtree` (services/legacy/netsync/handle_block.go, three sites) changes the subtree-file DAH from `bi.height + sm.settings.GlobalBlockHeightRetention` to `bi.height + sm.settings.GetSubtreeValidationBlockHeightRetention()`. Unlike the rest of this PR, this is applied on every legacy netsync block regardless of the fast-path flags. `GetSubtreeValidationBlockHeightRetention()` returns `GlobalBlockHeightRetention + SubtreeValidation.BlockHeightRetentionAdjustment`, so with the default adjustment (0) it is a no-op, but a deployment that has tuned `SubtreeValidation.BlockHeightRetentionAdjustment` to a non-zero value will see the legacy path's subtree-file retention shift — even with every new flag left off.

**Why it matters:** The default-off-flag safety argument that covers the rest of the diff does not cover this line. It is a real behaviour change to the legacy path bundled into a "refactor" PR. That said, it is a defensible consistency fix — it aligns legacy-written subtree files with `get_blocks.go` and `subtreevalidation` (one retention source) — so the intent looks correct.

**Fix:** Keep the change but call it out explicitly in the PR description as an unconditional legacy-path retention alignment (not gated by the new flags), so operators who tune the subtree-validation adjustment are aware:

```go
// one retention source for subtree files on every path (applies unconditionally)
dah := bi.height + sm.settings.GetSubtreeValidationBlockHeightRetention()
```

### ChiR3 — Stray planning comment in `TestCommitBlock_AddsBlockAndSetsExists`

**Problem:** services/blockvalidation/commit_block_test.go carries a leftover planning comment at the top of the test: "Build u (\*BlockValidation) + a valid below-checkpoint block with block.ID set and its subtree(s) in the subtree store ... (Implementer: reuse that harness.)". This contradicts the actual harness — `newCommitBlockHarness` deliberately builds a block with no subtrees (`NewBlock(..., nil, ...)`), and the trailing "(Implementer: ...)" reads as a stray authoring instruction.

**Why it matters:** Cosmetic only, no runtime effect, but it misdescribes the test setup and was already flagged inline. Misleading comments cost future readers time.

**Fix:** Delete the three-line comment block; the `newCommitBlockHarness` doc comment already describes the setup accurately.

### ChiR4 — Refactor-titled PR carries the full consensus-sensitive stack

**Problem:** The title scopes this to "extract shared commitBlock", but the diff is +2185/-129 across 19 files and includes the below-checkpoint `validOrderAndBlessed` skip (model/Block.go), the shared `BelowCheckpoint`/`OutpointOnlyEligible` gates (model/checkpoint.go), the unified legacy route (Server.go, handle_block.go), `CheckSubtreeSlicesForDuplicateTxs`, `needsParentMinedWait`, and a new setting. Only the `commitBlock` extraction (quick_validate.go + commit_block_test.go) is this PR's genuine increment.

**Why it matters:** Not a code defect, but it makes the true delta hard to isolate and risks the consensus-sensitive stacked changes being approved implicitly under a refactor-only title.

**Fix:** When merging, confirm the incremental delta is limited to the `commitBlock` extraction and that the model/netsync consensus changes are tracked and signed off in their parent PRs (#1178/#1252), not rubber-stamped here.

## Recap

| ChiR1 | Parent-mined-wait skip relies on cross-file invariants | | 32% |
| ChiR2 | Unconditional subtree-retention change on legacy path | | 22% |
| ChiR3 | Stray planning comment in commit test | | 15% |
| ChiR4 | Consensus stack under a refactor title | | 10% |

ordishs (approved):
Approving — the `commitBlock` increment (top two commits) is ready as-is. I verified the extraction is behaviour-identical against both removed tails (only deltas: the `caller` label threading and the corrected `[ValidateBlock]` → `[quickValidateBlock]` log prefix, as described), and ran the new model, blockvalidation and netsync tests locally — all green with `-count=1`.

`TestCommitBlock_AddsBlockAndSetsExists` is a good template: real sqlitememory store behind a `LocalClient`, asserted via genuine `GetBlockExists`/`GetBlockIsMined` rather than mocks.

Because this is stacked on #1252 with base `main`, the diff carries the whole unified-route stack. Four observations — the first two belong to the #1252 content and are worth addressing there (or in a follow-up) before the unified flag is ever enabled in anger:

1. **Coinbase-only blocks bypass the unified route's integrity floor.** `prepareSubtrees` returns nil slices for `txCount <= 1`, so the merkle + dedup guard in `HandleBlockDirect` never runs, and server-side `quickValidateBlock` takes the `len(block.Subtrees) == 0` branch straight to `commitBlock` — nothing on the route binds the coinbase to the header. A peer serving IBD could pair a genuine checkpoint-era header (PoW and linkage pass) of an empty block with a substituted coinbase tx and it would be committed. Early mainnet blocks — exactly the below-checkpoint population — are overwhelmingly coinbase-only. The fix is nearly free: `CheckMerkleRoot`'s zero-subtree `default` branch already compares the coinbase txid against `Header.HashMerkleRoot`, so calling it unconditionally on the unified route (not only when `preparedSubtreeSlices != nil`) closes the gap. `Block.Valid` has the same blind spot for subtree-less blocks (pre-existing), but the route's comment claims a tampered wire block "can never reach the UTXO store" — worth making that true.

2. **The `"legacy"` sentinel now gates skipping validation entirely.** `Server.ProcessBlock` defaults an empty `baseURL` to `"legacy"`. Previously that only toggled optimistic mining and catchup allowances; with the unified route it now means "skip full validation below the checkpoint", and `quickValidateBlock` checks neither PoW nor linkage — it trusts the caller. Today only netsync passes `"legacy"` in production, so this isn't currently exploitable, but any future internal caller omitting `baseURL` silently inherits quick-validation semantics. Consider an explicit request flag for the unified route, or a cheap `HasMetTargetDifficulty` backstop at the top of `quickValidateBlock`.

3. **`writeSubtree` retention change is not flag-gated.** Switching `GlobalBlockHeightRetention` → `GetSubtreeValidationBlockHeightRetention()` applies to the default legacy path too. A no-op at the default adjustment of 0, but deployments with `SubtreeValidation.BlockHeightRetentionAdjustment` set will see legacy subtree-file lifetimes change with flags off — worth a line in the PR description since the stack otherwise promises flag-off is byte-identical.

4. **`TestHandleBlockDirect_UnifiedRoute_DeduplicatesViaPreparedSlices` doesn't test the wiring.** It calls `model.CheckSubtreeSlicesForDuplicateTxs` directly, so it would still pass if the guard were deleted from `handle_block.go`, and its comment overclaims ("would have FAILED before the fix"). The model-level tests already cover the function — suggest dropping it or replacing it with a real `HandleBlockDirect` integration test when the harness allows.

What holds up particularly well: the unified branch in `processBlockFound` is correctly ordered after the malicious-peer / block-exists / parent-exists / block-assembly checks; the `needsParentMinedWait` skip's redundancy argument checks out (`sm.ProcessBlock` is a synchronous gRPC round-trip on a serial dispatch loop, and quick-validated blocks are created `MinedSet(true)`); the gate consolidation onto `BelowCheckpoint`/`OutpointOnlyEligible` is strictly more conservative (nil params, no checkpoints, height 0 all fail closed); and `TestCVE20122459_MerklePassesDedupCatches` proving — on the same mutated block — that `CheckMerkleRoot` passes while the dedup floor rejects is exactly the right shape of test. The netsync guard running *before* `ProcessBlock` also matters, since server-side `validateSubtrees` only verifies the merkle root after `createAndSpendUTXOsForBatch` has written UTXOs.

PR #1255, opened 2026-07-09
fix(p2p): keep banlist subscribers across Clear

gokutheengineer:
**Closes #4680.**

### Problem
`BanList.Clear()` recreated the `subscribers` map alongside `bannedPeers`, so every channel previously returned by `Subscribe()` was silently detached. The p2p server subscribes once at startup and consumes ban events in `listenForBanEvents`; since `Clear()` is reachable from the `ClearBanned` gRPC, a single call permanently disabled event-driven ban handling for the rest of the process lifetime.

### Fix
`Clear()` now resets only `bannedPeers` and deletes the DB rows, never `subscribers`. Subscriber lifecycle remains managed exclusively by `Unsubscribe()`.

### Tests
Added `TestBanList_ClearKeepsSubscribers` in `internal/banlist/ban_list_test.go`: a subscriber receives an event, `Clear()` runs, and a subsequent `Add()` must still deliver an event to the same channel (fails by timeout on the old code).

### DIFF
--- a/internal/banlist/ban_list.go
+++ b/internal/banlist/ban_list.go
@@ -299,7 +299,6 @@ func (b *BanList) Unsubscribe(ch chan BanEvent) {
 func (b *BanList) Clear() {
 	b.mu.Lock()
 	b.bannedPeers = make(map[string]BanInfo)
-	b.subscribers = make(map[chan BanEvent]struct{})
 	b.mu.Unlock()
 
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):

icellan (approved):
Correct fix. The old reset orphaned every subscriber channel, so a single ClearBanned stopped the p2p server reacting to all future ban events (subscribers only Subscribe once at GetBanList). Test covers it and passes under -race.

Two non-blocking nits, both pre-existing:

- `ban_list.go:334` — `notifySubscribersAsync` does a blocking `SafeSend` with no timeout. A stalled consumer (buffer cap 100) leaks the Add/Remove goroutine forever. Keeping subscribers alive across Clear means a stuck one now persists instead of getting accidentally detached, so worth the timeout variant.
- `ban_list.go:299` — `Clear()` emits no event, so surviving subscribers keep a stale view of the list. Harmless today since the p2p handler only acts on "add", but a latent inconsistency this change exposes.

PR #1256, opened 2026-07-09
p2p: enforce bans at every layer available (IP extraction, ban-event handling, disconnect RPCs, early gossip checks)

gokutheengineer:
**Closes #4679.**

### Problem
All five reported items were confirmed still present:

1. No network-layer disconnect exists - go-p2p-message-bus v0.1.17 `Client` exposes no disconnect/gater/host access, and `onPeerBanned`/`disconnectBannedPeerByID` only touched the registry.
2. Ban events were dropped - `handleBanEvent` ignored any event without a PeerID, and `disconnectPreExistingBannedPeers` replayed startup bans as IP-only events, so they were all ignored.
3. Score-based bans stored raw multiaddr strings (`/ip4/1.2.3.4/tcp/...`) in the IP ban list, where they failed `parseAddress` and never matched a lookup.
4. `ConnectPeer`/`DisconnectPeer` RPCs fell through to `UnimplementedPeerServiceServer`.
5. The banned-peer check ran after peer re-registration and WebSocket forwarding in the gossip handlers, and was entirely missing from `handleNodeStatusTopic`.

### Fix
- `onPeerBanned` now reduces each multiaddr to its literal IP (`extractIPFromMultiaddr`, deduped; DNS/relay addrs skipped) before calling `banList.Add`, so entries actually match `IsBanned` lookups. Removed the dead `myBanEventHandler`, which duplicated the same multiaddr bug.
- `handleBanEvent` acts on IP-only events (the only kind `BanList.Add` emits): it scans connected peers and disconnects any whose address IP/subnet is on the ban list (`disconnectPeersOnBanList`). Startup replay uses the same matching instead of fabricating events - better than the proposed "populate PeerID", because stored bans have no PeerID to populate.
- Implemented `ConnectPeer` (refuses banned literal-IP addresses, then dials via the client) and `DisconnectPeer` (shares `disconnectBannedPeerByID` with the ban path).
- Ban check now runs first in all four gossip handlers - immediately after the spoofing check, before WebSocket forwarding, peer-map storage, registration, and metrics - and was added to `handleNodeStatusTopic`, which had none. Own messages are exempted via `isOwnMessage`, so self-published traffic pays no extra registry round trip.
- Network-layer severance: the library cannot provide it - verified nothing up to v0.1.21 exposes the host, a gater, or a disconnect. Added a `networkDisconnector` capability interface that every disconnect path type-asserts against, so upgrading to a library version implementing `DisconnectPeer(ctx, peer.ID)` enables real connection closing with no further Teranode changes. Until then, a 1-minute sweep in `listenForBanEvents` re-disconnects banned peers that reconnect; the sweep only acts (and logs) when the peer re-entered the registry or the client can actually sever, so it stays quiet instead of logging fake disconnects. The library-side change (host.Network().ClosePeer + ConnectionGater seeded from the ban list) remains a follow-up in go-p2p-message-bus.


ctnguyen (approved):
Review at commit [4e2a63cddcb1](https://github.com/bsv-blockchain/teranode/commit/4e2a63cddcb1)

The core defects are fixed cleanly. `extractIPFromMultiaddr` reduces multiaddrs to a bare IP with dedup and DNS/relay skipping, which is exactly what `banList.Add`/`IsBanned` require (the previous code stored raw `/ip4/.../p2p/...` strings that failed `parseAddress` and never matched). The ban check is genuinely relocated ahead of byte accounting and forwarding in every handler, own messages are exempted to avoid a needless registry round-trip, and the `networkDisconnector` interface is a forward-compatible seam that lights up real libp2p severance once the library exposes it — with the nil-interface type assertion handled safely. Test coverage is thorough and faithful to the real code paths (IP-extraction table, dedup, startup sweep, RPC success/error, early-drop-before-registration), and removing the dead `myBanEventHandler` eliminates a duplicate of the same multiaddr bug.

## Non-blocking issues

### ChiR1 — Operator IP bans of an already-connected peer are not enforced at the gossip layer, and the sweep churns

**Problem:** The four gossip early-checks key off `shouldSkipBannedPeer` (`services/p2p/server_helpers.go:916`), which only consults the registry's score-based `IsPeerBanned`. An operator ban via `BanPeer` (`services/p2p/Server.go:1782`) writes an IP to `banList` but adds no registry entry. Because the library exposes no connection severance in v0.1.17, `disconnectPeersOnBanList` → `disconnectBannedPeerByID` → `removePeer` only drops the peer from the registry; the libp2p connection stays open. The very next gossip message from that peer is not skipped (registry ban is false), so it is fully processed and the peer is re-registered via `updateBytesReceived`/`UpdatePeerMetrics`. The 1-minute sweep then re-matches the re-registered peer and calls `disconnectBannedPeerByID` again, logging `Disconnecting banned peer` at Info on every pass — churn rather than enforcement. Score-based bans are unaffected (dropped before registration).

**Why it matters:** The PR title advertises enforcement "at every layer available", but for an already-connected, purely IP-banned peer the gossip path is neither dropped at the handler nor durably disconnected, and it emits a recurring Info log. This is still an improvement over the prior behaviour (IP-only events were dropped entirely), so it is a residual gap, not a regression.

**Fix:** In the gossip handlers, additionally resolve the connected peer's IP (map `fromID` → addr via `P2PClient.GetPeers`, then `banList.IsBanned`) so IP-banned peers are dropped even without a registry entry; and/or rate-limit / down-level the per-sweep Info log so a peer that cannot yet be severed does not log every minute. At minimum, align the PR description with the score-ban-only reality of the gossip path until the go-p2p-message-bus severance follow-up lands.

### ChiR2 — `ConnectPeer` ban check is bypassed for DNS/relay multiaddrs

**Problem:** `ConnectPeer` guards with `if ip := extractIPFromMultiaddr(req.PeerAddress); ip != "" && s.banList.IsBanned(ip)`. `extractIPFromMultiaddr` returns `""` for `/dns4/`, `/dns6/`, and relay-circuit addresses, so supplying a `/dns4/host/tcp/.../p2p/...` multiaddr skips the guard entirely and the dial proceeds even if the resolved host is banned.

**Why it matters:** Defense-in-depth gap only — `ConnectPeer` is an operator/RPC surface, not attacker-reachable, and no network-layer gate exists today regardless. Low impact.

**Fix:** For DNS multiaddrs, resolve to IP(s) (or reject) before dialing, or document that `ConnectPeer` only refuses literal-IP banned addresses:

```go
if ip := extractIPFromMultiaddr(req.PeerAddress); ip == "" {
    // DNS/relay addr — resolve or reject before dialing
}
```

### ChiR3 — Reference docs contradict the now-implemented ConnectPeer/DisconnectPeer handlers

**Problem:** `docs/references/services/p2p_reference.md:286` and `:292` both state "the server-side gRPC handler is not yet implemented" for `ConnectPeer` and `DisconnectPeer`. This PR implements both in `services/p2p/Server.go`, so those notes are now stale. The doc file is outside the diff.

**Why it matters:** Documentation-only; no runtime effect, but it misleads operators about which RPCs are available.

**Fix:** Remove the two "not yet implemented" notes in `docs/references/services/p2p_reference.md`.

## Recap

| ChiR1 | IP-ban gossip gap + sweep churn | | 34% |
| ChiR2 | ConnectPeer DNS ban bypass | | 14% |
| ChiR3 | Stale ConnectPeer/DisconnectPeer docs | | 10% |

icellan (commented):
Solid refactor overall — normalising ban entries to bare IPs, moving the gossip ban-drop ahead of registration/WebSocket forwarding, and consolidating the disconnect logic are all correct and well-tested. One real gap against the stated goal, plus two lower-severity issues.

**P1 — IP/subnet bans aren't enforced against gossip** (`services/p2p/server_helpers.go:922`, `:1156`)
The gossip drop path (`shouldSkipBannedPeer`) only checks the score-based registry ban (`peerRegistry.IsPeerBanned` → `reg.IsBannedPeer`); it never consults `s.banList`. An operator `BanPeer` RPC / subnet ban only populates `banList`. Since `go-p2p-message-bus` doesn't implement `networkDisconnector` (`canSever == false` in production), `disconnectPeersOnBanList` can't actually sever the connection, so the peer stays connected and its next gossip message re-registers it via `addPeer` — and `shouldSkipBannedPeer` returns false for it. Net effect: IP/subnet bans are a no-op for gossip traffic in production. Suggest consulting `banList.IsBanned(<peer IP>)` in the drop path, not just the registry.

**P2 — `ConnectPeer` ban check bypassed for DNS multiaddrs** (`services/p2p/Server.go:1821`)
`extractIPFromMultiaddr` returns `""` for `/dns4`, `/dns6`, `/dnsaddr` addresses (only P_IP4/P_IP6 are extracted). The guard `ip != "" && s.banList.IsBanned(ip)` therefore short-circuits and the dial proceeds. A banned host reachable via a DNS multiaddr connects anyway. Refuse (or resolve-then-check) when the IP can't be extracted from a non-loopback DNS address.

**P3 — sweep churns the sync coordinator for still-connected IP-banned peers** (`services/p2p/server_helpers.go:1156`, sweep at `:552`)
With `canSever == false`, each 1-minute sweep re-finds an IP-banned peer that has re-registered and calls `disconnectBannedPeerByID` → `removePeer` → `syncCoordinator.HandlePeerDisconnected(pid)` + a registry `RemovePeer`, without actually disconnecting it. If that peer is the selected sync peer, `HandlePeerDisconnected` fires against a live, syncing peer once per minute — needless churn for a ban that isn't taking effect anyway. Same root cause as P1: without severance the "disconnect" is cosmetic.

P1 and P3 share the root cause (no libp2p severance with the current library), so wiring `banList` into the gossip drop path largely addresses both.

Nit (left out of this summary): `handleRejectedTxTopic` evaluates `isOwnMessage` twice back-to-back — the own-message check could come first since that handler has no notification send.

icellan (approved):
Re-reviewed against the latest push (`d7947d959`). All three points from my earlier review are addressed, plus the redundant-`isOwnMessage` nit. Build and `go vet` clean, ban tests green.

- **P1 — IP bans on gossip:** `shouldSkipBannedPeer` now also consults `isPeerIPBanned`, which matches the peer's connected multiaddrs against the ban list (cached briefly via `ipBanCache`). The `TestHandleBlockTopic_IPBannedPeerDropped` / `TestShouldSkipBannedPeer_IPBanWithoutRegistryBan` tests confirm an operator IP ban both drops the gossip and stops re-registration — so it's effective even without libp2p severance. Resolved.
- **P2 — `ConnectPeer` DNS bypass:** `checkMultiaddrBanned` now resolves `/dns`, `/dns4`, `/dns6`, `/dnsaddr` and checks every resolved IP, and fails closed on a parse/resolution error. Covered by the banned-DNS and unresolvable-DNS tests. Resolved.
- **P3 — sweep churn:** with P1 in place the banned peer no longer re-registers, so the periodic sweep stops re-removing it after the first pass. Resolved as a consequence.

`isPeerIPBanned`, `checkMultiaddrBanned`, and the new `ipBanCache` eviction in `cleanupPeerMaps` all look correct — cache is bounded and access is safe. The only residual is a ~5s `ipBanCache` staleness window (a peer banned via `BanPeer` could have gossip accepted for up to 5s while a "not banned" entry is cached); it's bounded, symmetric with `reputationCache`, and the ban-event path disconnects independently, so not a blocker.

LGTM.

PR #1257, opened 2026-07-09
p2p: require admin API key for all state-mutating p2p admin RPCs

gokutheengineer:
**Closes #4678.**

### Problem
Only `BanPeer` and `UnbanPeer` were in the p2p gRPC auth allowlist (`services/p2p/Server.go`), so `ClearBanned`, `AddBanScore`, `ResetReputation`, `ConnectPeer`, and `DisconnectPeer` were reachable without the admin API key - enough to wipe the ban list, reset reputations, or ban arbitrary peers from any client that can reach the port. Additionally, when `grpc_admin_api_key` was unset, a random key was silently generated, making protected methods uncallable with no operator-visible signal. Both cited code locations were confirmed still present.

### Fix
- Extracted the protected set into `adminProtectedMethods()`, now covering all seven state-mutating admin RPCs: `BanPeer`, `UnbanPeer`, `ClearBanned`, `AddBanScore`, `ResetReputation`, `ConnectPeer`, `DisconnectPeer`. Legitimate callers (RPC service, asset HTTP, test daemon) go through the p2p `Client`, which already attaches `GRPCAdminAPIKey` from settings, so no caller changes are needed.
- Internal data-plane RPCs (`RecordCatchup*`, `UpdateCatchup*`, `ReportValidSubtree/Block`, `RecordBytesDownloaded`) deliberately stay unauthenticated: block/subtree validation call them continuously and protecting them would break reputation tracking wherever the admin key is empty. Instead of inverting the interceptor model at runtime, a classification test forces every PeerService RPC to be explicitly listed as protected or public, so new mutating RPCs are caught at test time rather than shipping open by omission.
- The coverage test also asserts `PeerService_ServiceDesc.Streams` is empty: the auth interceptor is unary-only, so a future streaming RPC would bypass authentication entirely; the test forces stream auth to be wired first.
- When `grpc_admin_api_key` is unset, the server still fails closed with a random key but now logs a warning that admin RPCs are unreachable until a key is configured (the key itself is never logged).

### Tests
- `TestAdminProtectedMethodsCoverAllRPCs` - enumerates `PeerService_ServiceDesc` and requires every RPC to be classified exactly once (protected or public), every listed entry to be a real RPC, and no streaming RPCs to exist while auth is unary-only.
- `TestAuthInterceptorProtectsAdminMethods` - runs `util.CreateAuthInterceptor` with the p2p protected set: each admin RPC is rejected with `Unauthenticated` when the key is missing or wrong and succeeds with the correct key; public RPCs pass through with no key.

Commands run (affected package only):
```
```
All green, including the pre-existing p2p suite.

### DIFF
--- a/services/p2p/Server.go
+++ b/services/p2p/Server.go
@@ -668,18 +668,14 @@ func (s *Server) Start(ctx context.Context, readyCh chan<- struct{}) error {
 		if err != nil {
 			return errors.NewServiceError("error generating random API key", err)
 		}
-	}
 
-	// Define protected methods - use the full gRPC method path
-	protectedMethods := map[string]bool{
-		"/p2p_api.PeerService/BanPeer":   true,
-		"/p2p_api.PeerService/UnbanPeer": true,
+		s.logger.Warnf("[P2P] grpc_admin_api_key is not set; a random key was generated so admin RPCs (ban, unban, clear bans, ban score, reputation reset, connect/disconnect peer) are unreachable until a key is configured")
 	}
 
 	// Create auth options
 	authOptions := &util.AuthOptions{
 		APIKey:           apiKey,
-		ProtectedMethods: protectedMethods,
+		ProtectedMethods: adminProtectedMethods(),
 	}
 
 	// this will block
@@ -845,6 +841,25 @@ func (s *Server) rejectedTxHandler(ctx context.Context) func(msg *kafka.KafkaMes
 func (s *Server) disconnectPreExistingBannedPeers(ctx context.Context) {
 	for _, banned := range s.banList.ListBanned() {
 		s.handleBanEvent(ctx, BanEvent{Action: banActionAdd, IP: banned})
+	}
+}
+
+// adminProtectedMethods returns the full gRPC method paths of every
+// state-mutating admin RPC on the PeerService; the auth interceptor requires
+// the admin API key for these. Read-only queries and internal data-plane
+// reporting RPCs (catchup metrics, valid block/subtree reports, bytes
+// downloaded) stay unauthenticated because other services call them without
+// admin credentials. Any new mutating admin RPC must be added here; the
+// classification is enforced by TestAdminProtectedMethodsCoverAllRPCs.
+func adminProtectedMethods() map[string]bool {
+	return map[string]bool{
+		"/p2p_api.PeerService/BanPeer":         true,
+		"/p2p_api.PeerService/UnbanPeer":       true,
+		"/p2p_api.PeerService/ClearBanned":     true,
+		"/p2p_api.PeerService/AddBanScore":     true,
+		"/p2p_api.PeerService/ResetReputation": true,
+		"/p2p_api.PeerService/ConnectPeer":     true,
+		"/p2p_api.PeerService/DisconnectPeer":  true,
 	}
 }
 
--- a/services/p2p/server_auth_test.go
+++ b/services/p2p/server_auth_test.go
@@ -1,0 +1,127 @@
+package p2p
+
+import (
+	"context"
+	"testing"
+
+	"github.com/bsv-blockchain/teranode/services/p2p/p2p_api"
+	"github.com/bsv-blockchain/teranode/util"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+// publicPeerServiceMethods are the PeerService RPCs deliberately reachable
+// without the admin API key: read-only queries and internal data-plane
+// reporting calls made by other services (block/subtree validation) that do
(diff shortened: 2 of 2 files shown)


ctnguyen (approved):
Review at commit [209da724fe99](https://github.com/bsv-blockchain/teranode/commit/209da724fe99)

This is a clean, security-positive fix. The classification of all 24 PeerService RPCs is complete and matches the generated `PeerService_ServiceDesc` (7 protected, 17 public), and `TestAdminProtectedMethodsCoverAllRPCs` enforces that mapping so a future mutating RPC cannot ship open by omission. The assertion that `Streams` is empty is a genuinely valuable guard, since the auth interceptor is unary-only and a streaming RPC would otherwise silently bypass authentication. Fail-closed behavior is preserved (a random key is still generated when the setting is empty, so the interceptor is always installed), the new warning is single-line and never logs the key, and every legitimate caller (RPC service, asset HTTP, test daemon) reaches these RPCs through the p2p `Client`, which attaches `GRPCAdminAPIKey`, so no caller changes are needed. Test coverage exercises no-metadata, wrong-key, and correct-key paths for each protected method plus pass-through for public ones.

## Non-blocking issues

### ChiR1 — Reputation-affecting data-plane RPCs stay unauthenticated

**Problem:** `RecordCatchupMalicious` (and the other `Record*`/`Update*` data-plane RPCs) remain in `publicPeerServiceMethods`, yet `RecordCatchupMalicious` calls `peerRegistry.UpdatePeerMetrics(..., malicious=true, ...)` at `services/p2p/handle_catchup_metrics.go:76`, which degrades a peer's reputation and can lead to banning. Any unauthenticated client that can reach the p2p gRPC port can therefore still influence peer reputation and effectively ban arbitrary peers — the same class of harm the PR sets out to close for `ClearBanned`/`AddBanScore`/etc.

**Why it matters:** The PR's stated goal is to stop reachable-without-key RPCs from being used to "wipe the ban list, reset reputations, or ban arbitrary peers." Leaving `RecordCatchupMalicious` public keeps a peer-banning vector open, so the attack surface is narrowed but not eliminated. This is explicitly acknowledged in the PR body as a deliberate tradeoff (block/subtree validation callers carry no admin credentials), so it is not a regression and not blocking — but it is real residual risk that should be tracked rather than silently accepted.

**Fix:** Keep the current scope (protecting these would break internal validation callers that carry no admin key), but open a follow-up to give internal data-plane callers a separate credential/interceptor path — or rely on network-level isolation of the p2p gRPC port — so reputation-mutating RPCs are not reachable by arbitrary clients. Optionally leave a short comment or issue reference next to the `RecordCatchup*` entries in `publicPeerServiceMethods` pointing at that follow-up, e.g.:

```go
// Internal data-plane reporting (unauthenticated by design; see follow-up
// to isolate reputation-mutating calls such as RecordCatchupMalicious).
"/p2p_api.PeerService/RecordCatchupMalicious": true,
```

## Recap

| ChiR1 | Reputation-mutating data-plane RPCs stay unauthenticated | | 30% |

freemans13 (approved):
Reviewed and approving. I traced the change end to end and it holds up: all 24 `PeerService` RPCs are now classified, the 7 state-mutating admin ones are protected, and the classification test forces any future RPC to be explicitly sorted into protected-or-public rather than shipping open by omission. The assertion that `PeerService_ServiceDesc.Streams` is empty is a nice touch — it stops a streaming RPC from silently bypassing the unary-only interceptor. `go vet` is clean and both new tests pass locally.

The thing I most wanted to rule out was breaking an internal caller of the newly-protected RPCs, and it's clear: the hot ban-score path (`server_helpers.go:730`, `Server.go:1994`) calls `s.AddBanScore` in-process, so the interceptor never runs; every genuine gRPC caller (asset HTTP `ResetReputation`, RPC `ClearBanned`, test-daemon `ConnectPeer`/`DisconnectPeer`) goes through the p2p `Client`, which attaches `GRPCAdminAPIKey` on dial; and the committed default `grpc_admin_api_key = testkey` keeps server and clients in sync for tests and dev.

Two notes for the record, neither a blocker:

1. **Behavioural change worth a changelog line.** When an operator leaves `grpc_admin_api_key` empty, the server fails closed with a random key, so `ClearBanned`, `AddBanScore`, `ResetReputation`, `ConnectPeer`, and `DisconnectPeer` — which previously worked without a key — become unreachable until a key is configured. This is the correct hardening and the new warning log surfaces it well (and never logs the key); it just changes observable behaviour for keyless deployments, so it's worth flagging in release notes.

2. **Residual risk, pre-existing.** The public data-plane RPCs that influence reputation and report valid blocks/subtrees (`UpdateCatchupReputation`, `RecordCatchupMalicious`, `ReportValidSubtree`, `ReportValidBlock`, …) stay unauthenticated on the same port, so anyone who can reach it can nudge reputation or report fake-valid work. You reasoned about this explicitly and the tradeoff is sound (validation calls them continuously without admin credentials); noting it only as context, since it predates this branch.

PR #1258, opened 2026-07-10
fix(util/aerospike): set MaxBufferSize once in init to fix data race

liam:
## What

`aerospike.MaxBufferSize` is a package-global in the aerospike client library that its batch/read commands consult concurrently. `GetAerospikeClient` wrote it (to a constant, 512MB) on **every** call. When a new node's daemon startup calls `GetAerospikeClient`, that write races with in-flight batch reads on already-running nodes' clients.

The race detector catches this intermittently in the multi-node `chainintegrity` E2E test:

```
WARNING: DATA RACE
Write at util/aerospike.go:98  ->  aerospike.MaxBufferSize = 1024 * 1024 * 512
   (GetAerospikeClient, during a new node's daemon startup)
Previous read by aerospike-client-go batchCommandOperate.parseRecord
   (an in-flight batch read on an already-running node)
```

## Fix

The value never changes, so hoist the assignment into package `init()`. It runs exactly once, before `main` and before any client or batch read can exist, giving a clean happens-before edge with the library's reads. Behaviorally equivalent (previously the value was set on the first `GetAerospikeClient` call, before any batch read against that freshly-created client) minus the repeated racy write.

## Test

`TestAerospikeBufferSize` updated to assert the new invariant: `init` sets `MaxBufferSize` to 512MB, and a failed `GetAerospikeClient` leaves it untouched. Verified load-bearing (removing the `init` write drops `MaxBufferSize` to the library default of 120MB and fails the test). `go test -race ./util/` passes.

### DIFF
--- a/util/aerospike.go
+++ b/util/aerospike.go
@@ -67,6 +67,14 @@ var aerospikeHistogramLastBuckets = *safemap.New[string, []uint64]()
 
 func init() {
 	aerospikeConnections = make(map[string]*uaerospike.Client)
+
+	// Raise the aerospike client's read buffer to 512MB for large records.
+	// aerospike.MaxBufferSize is a package-global in the client library that its
+	// batch/read commands consult concurrently. Writing it here in init (once,
+	// before main and before any client or batch read exists) establishes a
+	// happens-before edge with those reads; writing it on every GetAerospikeClient
+	// call instead raced with in-flight batch reads on already-running clients.
+	aerospike.MaxBufferSize = 1024 * 1024 * 512 // 512MB
 }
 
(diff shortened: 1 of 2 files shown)


icellan (approved):

sgextcsi (approved):

ordishs (approved):
Minimal, correctly scoped fix for a real data race. Moving the write-once `MaxBufferSize` assignment into `init()` gives a clean happens-before edge with the client library's concurrent reads, and it's now the only write site in the repo. Test upgraded from a tautological check to a load-bearing invariant (init sets 512MB; failed connect leaves it untouched), and it passes under `-race`.

PR #1262, opened 2026-07-10
chore(grafana): add teranode pruner dashboard

teracoder:
Adds the Teranode **pruner** Grafana dashboard, sourced from the `teranode-argocd-deployments` GitOps repo (source of truth for these dashboards). New file under `deploy/docker/base/grafana_dashboards/teranode/`, auto-provisioned via `foldersFromFilesStructure`.


oskarszoon (approved):
Reviewed as part of the dashboard batch. Checked structurally rather than line-by-line: JSON parses, uid is unique across the batch, Prometheus datasource is templated (so it imports cleanly), the PromQL references real teranode/infra series, and there are no embedded secrets. Deploy-only config, nothing risky. Approving.

PR #1266, opened 2026-07-10
chore(grafana): add teranode logs dashboard

teracoder:
Adds the Teranode **logs** Grafana dashboard, sourced from the `teranode-argocd-deployments` GitOps repo (source of truth). New file under `deploy/docker/base/grafana_dashboards/teranode/`, auto-provisioned via `foldersFromFilesStructure`.

### DIFF
--- a/deploy/docker/base/grafana_dashboards/teranode/teranode-logs-dashboard.json
+++ b/deploy/docker/base/grafana_dashboards/teranode/teranode-logs-dashboard.json
@@ -1,0 +1,498 @@
+{
+  "annotations": {
+    "list": [
+      {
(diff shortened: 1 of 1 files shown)


oskarszoon (approved):
Reviewed as part of the dashboard batch. JSON parses, LogQL references real series, no secrets. Approving.

One non-blocking nit: the Loki datasource is referenced by a literal `uid: "loki"` rather than a `${datasource}` template variable (the Prometheus dashboards in this batch are all templated). That breaks import on any stack where the Loki datasource uid isn't literally `loki`. Worth converting to a datasource variable when convenient — not blocking.

PR #1269, opened 2026-07-10
fix(pruner): route combined-cleanup parent updates through the native-op builder

icellan:
Follow-up to #828. The native-ops migration left the pruner still emitting a per-block Lua UDF burst on native-op-enabled deployments.

## Problem

Observed on the scaling clusters: with native ops fully enabled (probe passing on all services), the spend/setMined hot path is native (`batch_sub_write`), but `batch_sub_udf_complete` still climbs in a **per-block burst up to ~3.5M/s**. Traced to the pruner's `addDeletedChildren` parent-update.

The pruner has two cleanup paths:

- `executeBatchParentUpdates` (defensive mode on, two-call) — correctly prefers the injected `buildAddDeletedChildrenRecord` native-op builder.
- `executeBatchCleanupCombined` (defensive mode off, single round-trip) — chose UDF-vs-native purely on `luaPackage != ""`, **ignoring the native builder entirely**.

`pruner_provider.go` always injects the native builder *and* sets `LuaPackage` as the fallback, so the combined path always emitted a raw `NewBatchUDF`. Deployments run `pruner_utxoDefensiveEnabled=false` (the default) → combined path → the exact `batch_sub_udf` stream #828 set out to eliminate.

## Fix

- Extract record-building into `buildCombinedCleanupRecords`, which prefers the native builder, falls back to `NewBatchUDF`, then to the plain MapPutItems `BatchWrite` — mirroring `executeBatchParentUpdates`.
- Result parsing keys on `parentUsesModTeranode` (native + UDF both return a SUCCESS map and handle missing parents server-side) instead of `useUDF`, and normalises the native dispatcher's `map[string]interface{}` response via the existing `normaliseAddDeletedChildrenResponse`.

## Tests

- **Unit** (`combined_cleanup_native_op_test.go`): record routing for all three paths — native builder present, UDF fallback, plain MapPutItems.
- **Integration** (`native_op_add_deleted_children_integration_test.go`, build tag `aerospike`): against a real **Teraspike** server (BSV fork, wire op 200), asserts the store's native probe enables native ops, the record is a `BatchWrite` (not `NewBatchUDF`), the parent's `deletedChildren` map is mutated server-side, and the server's `batch_sub_udf_complete` counter does **not** move. Guarded by the build tag so it stays out of the default `go test`; skips cleanly on a stock Aerospike image. Verified locally against `aerospike-server:8.1.2.0-rc-9`.
- `go vet` (tagged + untagged) / `golangci-lint` clean on the changed files.

### DIFF
--- a/services/blockvalidation/catchup_fork_handling_test.go
+++ b/services/blockvalidation/catchup_fork_handling_test.go
@@ -687,11 +687,16 @@ func TestCatchup_NoMinedSetChurnOnRejectedFork(t *testing.T) {
 		mockBlockchainClient.On("GetBlockExists", mock.Anything, mock.Anything).
 			Return(false, nil).Maybe()
 
-		// Metadata for the common ancestor.
+		// findCommonAncestor now conveys existence through GetBlockHeader: the common
+		// ancestor resolves to its header+meta, while every fork block (and the target)
+		// must report ErrBlockNotFound so the walk stops at the ancestor and the fork
+		// headers are treated as the peer's offered chain. A permissive catch-all that
+		// returned metadata for the fork blocks would make the whole peer chain look
+		// already-present, leaving zero offered headers for the secret-mining check.
 		mockBlockchainClient.On("GetBlockHeader", mock.Anything, ancestorHeader.Hash()).
 			Return(ancestorHeader, &model.BlockHeaderMeta{Height: ancestorHeight, ID: 1, ChainWork: ancestorChainWork.Bytes()}, nil).Maybe()
 		mockBlockchainClient.On("GetBlockHeader", mock.Anything, mock.Anything).
-			Return(&model.BlockHeader{}, &model.BlockHeaderMeta{Height: ancestorHeight, ChainWork: ancestorChainWork.Bytes()}, nil).Maybe()
+			Return((*model.BlockHeader)(nil), (*model.BlockHeaderMeta)(nil), errors.NewBlockNotFoundError("block not in our chain")).Maybe()
 
 		// Header-fetch plumbing.
 		mockBlockchainClient.On("GetBestBlockHeader", mock.Anything).
--- a/stores/utxo/aerospike/native_op_add_deleted_children_integration_test.go
+++ b/stores/utxo/aerospike/native_op_add_deleted_children_integration_test.go
@@ -1,0 +1,143 @@
+//go:build aerospike
+
+package aerospike
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"os"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/bsv-blockchain/aerospike-client-go/v8"
+	"github.com/bsv-blockchain/go-bt/v2/chainhash"
+	"github.com/bsv-blockchain/teranode/stores/blob/memory"
+	"github.com/bsv-blockchain/teranode/stores/utxo/fields"
+	"github.com/bsv-blockchain/teranode/ulogger"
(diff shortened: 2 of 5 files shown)


teracoder (approved):

freemans13 (approved):
Reviewed locally against `feat/teranode-native-ops`. This is a tight, correct fix: the combined single-round-trip cleanup path was keying UDF-vs-native purely on `luaPackage != ""`, so it always emitted a raw `NewBatchUDF` and ignored the injected native-op builder — exactly the `batch_sub_udf` burst #828 aimed to eliminate. The extracted `buildCombinedCleanupRecords` now mirrors the already-correct `executeBatchParentUpdates` (native builder → UDF fallback → plain MapPutItems), and result parsing correctly switches from `useUDF` to `parentUsesModTeranode`.

Verified locally:
- `go vet ./stores/utxo/aerospike/pruner/` clean
- New `TestBuildCombinedCleanupRecords_*` (3) pass — cover native-present, UDF-fallback, and plain-MapPutItems routing
- gofmt clean; no stale `useUDF` references left in the package
- `normaliseAddDeletedChildrenResponse` already has coverage in `parent_update_response_test.go`, so the native response-shape handling isn't untested

Non-blocking notes:
1. As you flag, the native wire-op 200 path isn't exercised by the stock test container — worth confirming `rate(aerospike_namespace_batch_sub_udf_complete)` drops to zero on a native-op cluster after redeploy.
2. Trivial: the `firstParentErr` rationale comment (~line 1685) still says "UDF SUCCESS-map"; the in-loop comments were correctly generalised to "native + UDF". Purely cosmetic.

LGTM.

oskarszoon (approved):
Reviewed with golang-pro + a second-eyes pass. The fix is correct: it faithfully mirrors the proven two-call `executeBatchParentUpdates` path — identical native-builder invocation (same shared `*BatchUDFPolicy`) and identical result-parsing (`normaliseAddDeletedChildrenResponse` on `Bins["SUCCESS"]`), and `parentUsesModTeranode` is threaded as a single value shared by build and parse so the two can't drift. Line-by-line the extraction is behaviour-preserving beyond the intended edits, and the new native-routing predicate now matches the two-call gate, removing a real inconsistency. The important part: the old code was self-consistently broken (raw UDF on both build and parse), so this had to fix both sides in lockstep — and it does. No marker-loss/miscount; the mined-only gate is upstream, so this only changes transport. Approving.

Non-blocking:
- **The combined result-parsing loop has no test.** The new tests cover record *building* for all three routes, but nothing pushes a native (`map[string]interface{}`) or UDF SUCCESS-map response through the `parentUsesModTeranode`-keyed parse loop — the existing integration tests build `Service` without the native/lua builders, so they only hit the plain-BatchWrite/`KEY_NOT_FOUND` branch. `normaliseAddDeletedChildrenResponse` itself is unit-tested for both map shapes, so the untested surface is the parse wiring. A table test injecting a fake native SUCCESS response closes it cheaply. Also: all three new tests pass `keys=nil`, so the parent/child split (`parentEnd`) is never exercised — add a case with both.
- Still worth the cluster confirmation you flagged (`rate(aerospike_namespace_batch_sub_udf_complete)` → 0 after redeploy), since the native path (wire-op 200) can't be exercised by the test container.

Follow-up (pre-existing, not this PR): `normaliseAddDeletedChildrenResponse` is a narrower duplicate of the canonical `ParseLuaMapResponse` (no reflect fallback), and the parse `switch` has no `default` — a shape-parse miss or unexpected status is silently counted as a parent-update success. Worth converging on the shared helper.

PR #1271, opened 2026-07-13
Remove dead code from p2p service

gokutheengineer:
**Closes #52.**

### Problem
Three unexported symbols in `services/p2p/` were defined but never wired in:
- `shouldSkipDuringSync` (server_helpers.go) - the during-sync announcement gate was never called, so the filtering it implies never actually ran.
- `contains` (server_helpers.go) - a multiaddr/peer-ID membership helper with no callers (the remaining `contains` in the package is an unrelated `clientChannelMap` method).
- `myBanEventHandler` (Server.go) - a dead type duplicating multiaddr-to-`banList.Add` logic.

### Fix
Deleted all three, plus the `go-multiaddr` import that only `contains` used. Chose deletion over wiring `shouldSkipDuringSync` into the gossip handlers: enabling a new during-sync filter is a behavior change that should be designed and tested in its own PR, not bundled into dead-code cleanup. Helpers the deleted code called (`getSyncPeer`, `isBlockchainSyncingOrCatchingUp`, `removePeer`) have other callers and are untouched.

Pre-existing and out of scope: `maxP2PMessageSize` (Server.go) is also unused but predates this change and isn't part of the issue.

### Tests
No new tests - pure dead-code removal with no behavior to exercise. Verified with:
```
go build ./services/p2p/ && go vet ./services/p2p/
```
All green (`ok github.com/bsv-blockchain/teranode/services/p2p 14.075s`).


freemans13:
One optional adjacent cleanup, if you feel like sweeping it in: `MockBanEventHandler` (`services/p2p/test_helpers.go:23-28`) looks orphaned now. It mocks the `BanEventHandler` interface, but that interface is gone — only the "Replaces the old BanEventHandler indirection" comment at `Server.go:1869` still references the concept — and the mock itself is never instantiated anywhere (`MockBanEventHandler{` matches nothing in the tree). It's the last remnant of the same ban-handler indirection you're removing here.

Totally your call whether it belongs in this PR: it sits in an untouched test file and just outside the three symbols named in the issue, so leaving it for a follow-up is equally defensible. Just flagging it so it doesn't get lost.

The rest of the change verifies clean on my end — all three removals are provably dead (no callers anywhere), the deleted `myBanEventHandler.OnPeerBanned` is fully superseded by the live `s.onPeerBanned` method that the `TestServerHelpers_OnPeerBanned_*` tests target, and `go build` / `go vet` / `go test -race -tags testtxmetacache ./services/p2p/...` (302 passed) are all green.

icellan (approved):

ctnguyen (approved):

PR #1273, opened 2026-07-14
fix(blockvalidation): validate catchup header nBits against DAA target (#1147)

### LINKED ISSUE #1147: [TN-008][low] blockvalidation: catchup header batch validates PoW vs self-declared target, not DAA-required nBits
**Component:** `services/blockvalidation/catchup.go`, `services/blockvalidation/catchup/header_validation.go`

### Summary
During catchup header-fetch, each header's PoW is checked only against the target encoded in its own `nBits` (`HasMetTargetDifficulty`); the DAA-required `nBits` for the height is not recomputed/compared at this stage. A peer could supply self-consistent low-difficulty headers.

### Evidence
- `catchup.go:1293-1335` (`validateBatchHeaders`) → `catchup/header_validation.go:15-27` (`ValidateHeaderProofOfWork` calls only `header.HasMetTargetDifficulty()`).
- Downstream mitigation: full-block validation compares `nBits` to `GetNextWorkRequired` (`BlockValidation.go:1426-1440`); low-difficulty chain accrues low chainwork so can't win tip selection.

### Why it matters
Early header acceptance and secret-mining/fork-depth decisions run on unverified-difficulty headers before full validation, widening the resource-exhaustion / decision-skew surface during catchup.

### Suggested fix
In `validateBatchHeaders`, validate each header's `nBits` against the expected DAA target once heights are known.

### Suggested tests
Feed `validateBatchHeaders` headers with artificially easy `nBits` whose hashes meet that easy target → assert rejection.

---

ordishs:
Closes #1147

## Problem

During catchup header-fetch, each header's proof-of-work was checked only against the target encoded in its own `nBits` (`HasMetTargetDifficulty`). The DAA-required `nBits` for the height was never recomputed at the header stage, so a peer could supply a self-consistent low-difficulty chain. That chain was accepted early (widening the resource-exhaustion / decision-skew surface during catchup) and only rejected later, during full-block validation.

## Fix

Recompute the DAA-required difficulty from the fetched header chain and reject mismatches **before** fetching full blocks.

- **`blockchain.ComputeTarget`** — extracted the pure DAA target arithmetic out of the unexported `Difficulty.computeTarget` into an exported free function; the method now delegates to it. This lets the catchup path reuse the *exact* consensus math rather than a parallel reimplementation. Behaviour-preserving.
- **`validateHeaderChainDifficulty`** (new, `catchup_difficulty.go`) — reconstructs cumulative chainwork forward from the common ancestor's stored chainwork (same pattern as the existing `offeredChainHasMoreWork`), selects the median-of-three "suitable" blocks exactly as `GetSuitableBlock` does, and compares each header's `nBits` to `ComputeTarget`. Honours the regtest (no-adjustment), testnet min-difficulty, and early-height rules identically to `CalcNextWorkRequired`. A mismatch flags the peer as malicious.
- **Wiring** — runs as a new step in `catchup()` after checkpoint verification and before `fetchAndValidateBlocks`.

### Design notes
Two refinements over the issue's literal suggestion, both because the DAA needs the fork point's cumulative chainwork and height:
- Placed at the orchestration level (after `findCommonAncestor` establishes the authoritative anchor) rather than inside per-batch `validateBatchHeaders`, which runs before the common ancestor is known.
- Covers every header whose full 144-block window is in memory — the deep, peer-controlled region this finding is about. Headers within a window-depth of the anchor remain covered by the downstream full-block DAA check (`GetNextWorkRequired`), which reads the store.

## Tests

New `catchup_difficulty_test.go`:
- easy-`nBits` header deep in the chain → rejected as malicious (the finding's suggested test)
- valid steady-state chain accepted
- regtest (no-adjustment) → any difficulty accepted
- chain shorter than the window → nothing checked at this stage
- testnet min-difficulty rule honoured (accepted with pow-limit target, rejected without)
- cross-package guard that `ComputeTarget` matches the store-backed result

## Verification

- `go test ./services/blockchain/` → ok
- `go test ./services/blockvalidation/` → ok (full suite, incl. new tests and the wired step)
- `go vet` → clean
- `golangci-lint` → no new issues in changed files
- `go build ./...` → clean

### DIFF
--- a/services/blockchain/Difficulty.go
+++ b/services/blockchain/Difficulty.go
@@ -167,10 +167,24 @@ func (d *Difficulty) CalcNextWorkRequired(ctx context.Context, blockHeader *mode
 //   - suitableFirstBlock: First block in the difficulty adjustment window
 //   - suitableLastBlock: Last block in the difficulty adjustment window
 func (d *Difficulty) computeTarget(suitableFirstBlock *model.SuitableBlock, suitableLastBlock *model.SuitableBlock) (*model.NBit, error) {
+	return ComputeTarget(d.settings, suitableFirstBlock, suitableLastBlock)
+}
+
+// ComputeTarget calculates the DAA target difficulty from the first and last suitable
+// blocks of the difficulty adjustment window. It is the pure arithmetic core shared by
+// the store-backed Difficulty.CalcNextWorkRequired path and by callers that reconstruct
+// the window from headers not yet persisted (e.g. block validation catchup). It reads
+// only the ChainWork and Time of the two suitable blocks plus the chain parameters, so
+// it makes no store access and holds no state.
+//
+// Parameters:
+//   - tSettings: chain settings (TargetTimePerBlock, PowLimit, NoDifficultyAdjustment)
+//   - suitableFirstBlock: first (older) suitable block in the window
+//   - suitableLastBlock: last (newer) suitable block in the window
+func ComputeTarget(tSettings *settings.Settings, suitableFirstBlock *model.SuitableBlock, suitableLastBlock *model.SuitableBlock) (*model.NBit, error) {
 	lastSuitableBits, _ := model.NewNBitFromSlice(suitableLastBlock.NBits)
 	// If regtest we don't adjust the difficulty
-	if d.settings.ChainCfgParams.NoDifficultyAdjustment {
-		d.logger.Debugf("no difficulty adjustment - returning %v", lastSuitableBits)
+	if tSettings.ChainCfgParams.NoDifficultyAdjustment {
 		return lastSuitableBits, nil
 	}
 
@@ -179,29 +193,22 @@ func (d *Difficulty) computeTarget(suitableFirstBlock *model.SuitableBlock, suit
 	lastChainwork := new(big.Int).SetBytes(suitableLastBlock.ChainWork)
 
 	work := new(big.Int).Sub(lastChainwork, firstChainwork)
-	d.logger.Debugf("work: %s", work.String())
 
 	// In order to avoid difficulty cliffs, we bound the amplitude of the
 	// adjustment we are going to do.
-	d.logger.Debugf("suitableLastBlock.Height: %d, suitableFirstBlock.Height: %d", suitableLastBlock.Height, suitableFirstBlock.Height)
-	d.logger.Debugf("suitableLastBlock.Time: %d, suitableFirstBlock.Time: %d", suitableLastBlock.Time, suitableFirstBlock.Time)
-
 	duration := int64(suitableLastBlock.Time - suitableFirstBlock.Time)
-	if duration > 288*int64(d.settings.ChainCfgParams.TargetTimePerBlock.Seconds()) {
-		d.logger.Debugf("duration %d is greater than 288 * target time per block %.0f - setting to 288 * target time per block", duration, d.settings.ChainCfgParams.TargetTimePerBlock.Seconds())
-		duration = 288 * int64(d.settings.ChainCfgParams.TargetTimePerBlock.Seconds())
-	} else if duration < 72*int64(d.settings.ChainCfgParams.TargetTimePerBlock.Seconds()) {
-		d.logger.Debugf("duration %d is less than 72 * target time per block %.0f - setting to 72 * target time per block", duration, d.settings.ChainCfgParams.TargetTimePerBlock.Seconds())
-		duration = 72 * int64(d.settings.ChainCfgParams.TargetTimePerBlock.Seconds())
+	if duration > 288*int64(tSettings.ChainCfgParams.TargetTimePerBlock.Seconds()) {
+		duration = 288 * int64(tSettings.ChainCfgParams.TargetTimePerBlock.Seconds())
+	} else if duration < 72*int64(tSettings.ChainCfgParams.TargetTimePerBlock.Seconds()) {
+		duration = 72 * int64(tSettings.ChainCfgParams.TargetTimePerBlock.Seconds())
 	}
 
 	// Calculate the projected work by multiplying the current work by the target time per block (in seconds).
-	projectedWork := new(big.Int).Mul(work, big.NewInt(int64(d.settings.ChainCfgParams.TargetTimePerBlock.Seconds())))
+	projectedWork := new(big.Int).Mul(work, big.NewInt(int64(tSettings.ChainCfgParams.TargetTimePerBlock.Seconds())))
 
 	// Divide the projected work by the actual time duration between the blocks to get the adjusted work.
 	// check if duration is zero
 	if duration == 0 {
-		d.logger.Debugf("duration is zero - returning %v", lastSuitableBits)
 		return lastSuitableBits, nil
 	}
 
@@ -222,7 +229,6 @@ func (d *Difficulty) computeTarget(suitableFirstBlock *model.SuitableBlock, suit
 	// Calculate the new target by dividing the result by the adjusted work. This gives the new difficulty target.
 	// check if pw is zero
 	if pw.Sign() == 0 {
-		d.logger.Debugf("pw is zero, - returning %v", lastSuitableBits)
 		return lastSuitableBits, nil
 	}
 
@@ -233,9 +239,8 @@ func (d *Difficulty) computeTarget(suitableFirstBlock *model.SuitableBlock, suit
 	// This was incorrectly squaring the target, making it much larger (easier difficulty)
 
 	// clip again if above minimum target (too easy)
-	if newTarget.Cmp(d.settings.ChainCfgParams.PowLimit) > 0 {
-		d.logger.Debugf("new target would be above pow limit, set to pow limit")
-		newTarget.Set(d.settings.ChainCfgParams.PowLimit)
+	if newTarget.Cmp(tSettings.ChainCfgParams.PowLimit) > 0 {
+		newTarget.Set(tSettings.ChainCfgParams.PowLimit)
 	}
 
 	// Convert back to compact format
--- a/services/blockvalidation/catchup.go
+++ b/services/blockvalidation/catchup.go
@@ -265,6 +265,14 @@ func (u *Server) catchup(ctx context.Context, blockUpTo *model.Block, peerID, ba
 
 	u.reportValidatedHeaderProgress(catchupCtx)
 
+	// Step 9.5: Validate header-chain difficulty (DAA) before fetching full blocks.
+	// Header-fetch validation only checks each header's PoW against its own nBits; this
+	// recomputes the DAA-required nBits from the header chain and rejects a peer that
+	// supplied a self-consistent low-difficulty chain.
+	if err = u.validateCatchupHeaderDifficulty(ctx, catchupCtx); err != nil {
+		return err
+	}
+
 	// Step 10: Fetch and validate blocks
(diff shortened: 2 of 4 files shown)


ctnguyen (changes requested):
Review at commit [745cae5](https://github.com/bsv-blockchain/teranode/commit/745cae5)

The design choices are sound and clearly justified. Extracting `ComputeTarget` as a pure free function so the catchup path reuses the *exact* consensus arithmetic (rather than a parallel reimplementation) is the right call, and the extraction is genuinely behaviour-preserving — the only removals are debug logs. The chainwork reconstruction correctly mirrors `work.CalculateWork`/`getCumulativeChainWork` (same `CalcBlockWork` increment, same big-endian `SetBytes` treatment as the store), the window/ancestor offset math (`firstIdx = parentIdx - window`, guard `firstIdx >= 2`) lines up precisely with `CalcNextWorkRequired`'s `GetHashOfAncestorBlock(parent, 144)` + median lookback, the testnet-min-difficulty and early-height branches track the store path, and the placement after `findCommonAncestor` is well-reasoned. Tests cover the core accept/reject/skip cases and cross-check the extraction against the store-backed result.

## Blocking issues

### ChiR1 — Median-of-three input ordering diverges from the store, breaking DAA parity on timestamp ties
`median3` builds the candidate slice newest-first:
```go
s := []*model.SuitableBlock{suitable(idx), suitable(idx - 1), suitable(idx - 2)} // [block, parent, grandparent]
```
but the store's `GetSuitableBlock` feeds `getMedianBlock` candidates ordered `depth DESC`, i.e. `[grandparent, parent, block]` — the reverse. `util.SortForDifficultyAdjustment` is an unstable sorting network keyed only on `GetTime()`, so when two of the three candidates share a timestamp the selected median (`s[1]`) depends on input position and the two orderings pick **different** blocks. Concretely, for times `grandparent=100, parent=200, block=100` the store selects `block` while this code selects `grandparent` — same time but different `ChainWork`, so the window boundary's cumulative work differs and `ComputeTarget` yields a different expected `nBits`.

Why it matters: this is the one place the PR deviates from the canonical consensus math it claims to mirror. Equal timestamps between blocks two apart occur on real chains; when such a pair lands on a difficulty-window boundary during catchup, the check computes the wrong expected target and flags an **honest** peer as malicious — and since the recomputation is deterministic, every honest peer serving that region is rejected, stalling catchup (a liveness bug). (The reverse — accepting a bad chain — is caught downstream by the store-backed full-block DAA check, so the harm is the false-positive.)

Fix: build the slice in the same order the store uses, oldest chain-position first, so the tie-break matches:
```go
s := []*model.SuitableBlock{suitable(idx - 2), suitable(idx - 1), suitable(idx)}
```
Criticality: 60%

## Non-blocking issues

### ChiR2 — No test covers the equal-timestamp median tie
The tie path is exactly the scenario ChiR1 gets wrong, yet every test uses strictly increasing or constant-spacing timestamps, so the ordering bug passes CI. Add a case where two window-boundary candidates two apart share a timestamp (with the middle one higher), asserting the result matches a store-backed `GetSuitableBlock`/`CalcNextWorkRequired` computation. This both proves the ChiR1 fix and locks the ordering against future regressions.
Criticality: 40%

### ChiR3 — No-adjustment networks are checked more laxly than the store path
For `NoDifficultyAdjustment` chains, `validateHeaderChainDifficulty` returns `nil` immediately and accepts *any* `nBits`. The canonical path is not that permissive: `CalcNextWorkRequired` returns `&blockHeader.Bits` — the parent header's bits — under no-adjustment (`Difficulty.go:65-67`, via `GetNextWorkRequired`), so a header whose bits differ from its parent's *is* rejected by full-block validation. The header-stage check thus deviates from the parity the PR claims. This is not a security hole — no-adjustment is regtest, and mismatched bits are still caught downstream by the store-backed DAA check — but it contradicts the stated goal and the regtest test's implicit expectation. Either match the parent's bits here (skip only when `headers[i].Bits == headers[i-1].Bits`, with headers[0] compared against a parent bits value if available), or add a one-line comment stating that catchup intentionally stays laxer than full-block validation on no-adjustment networks.
Criticality: 30%

### ChiR4 — `ComputeTarget` recomputes `2^256` on every call inside an O(n) loop
`ComputeTarget` does `new(big.Int).Exp(big.NewInt(2), big.NewInt(256), nil)` per invocation, and the loop calls it (via the default branch) up to once per header across the whole in-memory header set. The constant `2^256` could be computed once and reused. Impact is modest relative to block download, but it's pure repeated work on a per-header hot path; worth hoisting the constant (in `ComputeTarget` or as a package-level value) if touched.
Criticality: 25%

### ChiR5 — Context cancellation is checked once, not inside the validation loop
`validateCatchupHeaderDifficulty` checks `ctx.Done()` before delegating, but `validateHeaderChainDifficulty` runs its full O(n) loop with no further cancellation check. For a large fetched header set the step can't be interrupted mid-way. Consider a periodic `ctx` check inside the loop (or pass `ctx` down) for responsiveness during shutdown/reorg.
Criticality: 15%

## Recap

| ChiR1 | Median tie-break order diverges from store | YES | 60% |
| ChiR2 | Missing equal-timestamp tie test | | 40% |
| ChiR3 | No-adjustment check laxer than store path | | 30% |
| ChiR4 | `2^256` recomputed per header | | 25% |
| ChiR5 | No ctx cancellation inside loop | | 15% |

ctnguyen (approved):
Review at commit [7b08b2f](https://github.com/ordishs/teranode/commit/7b08b2f52e0210458cb5e19b5d5b9bb11237a505)

Nice work — the one blocking issue from the last pass (ChiR1) is now fixed correctly. `median3` builds its candidate slice oldest-first (`suitable(idx-2), suitable(idx-1), suitable(idx)`) so the unstable sort's tie-break lines up exactly with the store's depth-DESC ordering, and the comment explains why. The fix is clean and tightly scoped — it touches only the median selection and its test file, with no collateral changes and no new defects, so this unblocks merge. The remaining items below are all non-blocking; the main thing worth flagging is that ChiR1 currently ships without a test that actually exercises the tie path it fixes.

## Status of previous points

ChiR1 is genuinely fixed and verified. For ChiR2, note the caveat: the added test `TestValidateHeaderChainDifficulty_MedianOrderingMatches` is effectively a no-op — its own docstring admits it does not construct an equal-timestamp tie, and its inputs are identical to `TestValidateHeaderChainDifficulty_ValidConstantChain` (mainnet, `180a097a`, 300 headers, constant 600s spacing). With constant spacing there are no timestamp ties, so it would pass even under the old buggy newest-first ordering. It therefore provides zero coverage of the tie path and cannot catch a regression of the ChiR1 fix — the fix currently ships without an effective regression test. A case with two window-boundary candidates two apart sharing a timestamp (middle one higher), asserted against a store-backed `GetSuitableBlock`/`CalcNextWorkRequired` result, would actually lock the ordering.

| ChiR1 | Median-of-three ordering diverged from store | YES | FIXED |
| ChiR2 | No equal-timestamp median tie test | | PARTIAL FIXED |

freemans13:
## Review notes

I ran a pre-PR-style review focused on the consensus-critical part of this change — the in-memory DAA recomputation. Because the existing full-block check still catches a bad chain downstream, the failure mode I weighted most heavily wasn't *letting a bad chain through*, it was the opposite: **falsely flagging an honest peer as malicious**. So most of the effort went into confirming the in-memory path is a faithful mirror of the store-backed `CalcNextWorkRequired`.

### What I verified holds up

I traced each component of `validateHeaderChainDifficulty` against the store path and they line up exactly:

- **Per-block work + chainwork accumulation** — identical `CalcBlockWork` formula, and the anchor's chainwork baseline cancels in `ComputeTarget` (it only uses the *difference*), so the check is robust even if the anchor chainwork were empty/wrong.
- **Window anchor** — `firstIdx = parentIdx - 144` matches `GetHashOfAncestorBlock(prev, 144)` depth semantics; no off-by-one.
- **Height mapping** — `parentHeight` maps to the *previous* block's height, which is exactly what the store feeds `CalcNextWorkRequired` via `meta.Height`, so the early-height and testnet min-difficulty rules trigger on the same boundary in both paths.
- **Median-of-three tie-break** (the second commit) — the store's CTE returns candidates oldest-first (`ORDER BY depth DESC`) and the fix now builds them oldest-first too, so `SortForDifficultyAdjustment`'s tie behaviour is identical. This was a genuine latent false-positive bug and the fix is correct.
- **The `cumWork` loop** allocates a fresh `big.Int` per iteration, avoiding the aliasing trap where every slice entry points at the same accumulator.

Locally: `go vet` clean, the 7 new tests pass, `golangci-lint` clean on the changed files. I did **not** run the heavier CI (integration/smoke/sequential) or a live testnet catchup, so the testnet min-difficulty behaviour is verified by parity-of-logic with the store, not by observation on a running node.

### Findings — all non-blocking; 0 correctness bugs

**1. The tie-break regression test doesn't exercise the scenario it protects** (`catchup_difficulty_test.go`, `TestValidateHeaderChainDifficulty_MedianOrderingMatches`)
This is the one I'd genuinely push on before merge. The test for the commit-2 ordering fix doesn't actually construct equal-timestamp blocks at a window boundary — its own docstring admits this — it just re-runs the constant-chain case. So it would not fail if the ordering regressed. Suggest a case with a suitable-block triple carrying equal timestamps but differing chainwork (or a direct unit test on `median3`) that asserts the honest chain is still accepted.

**2. `anchor == nil` guard silently accepts everything** (`catchup_difficulty.go`)
Returning `nil` on a nil anchor quietly disables the security check if the invariant ever breaks. It's safe today — `verifyCheckpointsInHeaderChain` (step 9) dereferences `commonAncestorMeta` unconditionally before step 9.5, so it's guaranteed non-nil here. Fine as defensive code; optionally log at debug if it's ever hit so a future invariant break is visible rather than silent.

**3. Minor redundant big-int work in the hot loop** (`Difficulty.go`, `ComputeTarget`)
`2^256` is recomputed via `big.Int.Exp` on every call, and `median3`/`suitable` allocate a fresh slice + SuitableBlocks (~6 per header). Not a regression — the store path pays the same per-call cost — but since the catchup batch path is a tight loop over potentially many headers, the `2^256` constant could be hoisted. Low priority.

**4. `powLimitNBit` duplicates `NewDifficulty`'s pow-limit derivation** (`catchup_difficulty.go`)
Same little-endian → `NewNBitFromSlice` derivation as `blockchain.NewDifficulty`. Minor; a shared helper would keep the two in lockstep if `PowLimitBits` handling ever changes.

**5. Extraction dropped the `Debugf` diagnostics from `computeTarget`** (`Difficulty.go`)
Behaviour-preserving, but the work-value/heights/times/clamp debug lines are gone. Acceptable trade-off for a pure stateless function on a hot path — flagging only so it's a conscious choice.

Overall: the consensus math is faithfully replicated, and the one real false-positive risk (median ordering) was already caught and fixed. Solid, well-scoped hardening. The only thing I'd want addressed before merge is finding 1 — the fix is currently guarded by a test that can't fail.

freemans13 (approved):
Approving. Reviewed the consensus-critical DAA recomputation in depth and confirmed the in-memory path faithfully mirrors the store-backed CalcNextWorkRequired (chainwork accumulation, window anchor, height mapping, and the median tie-break fixed in the second commit) — no correctness bugs, and the false-positive risk is well controlled. Findings in my earlier comment are all non-blocking; the only one worth addressing before/after merge is the tie-break regression test, which doesn't yet exercise the equal-timestamp scenario it protects.

PR #1275, opened 2026-07-14
fix(validator): bound parent vout before extend deref (backport #1243 to release/v0.15)

oskarszoon:
v0.15.4 (`ea6d8ce76`) backported only the unreachable `idx >= len` half of #1243 and dropped the half that stops the crash: bounding the untrusted `PreviousTxOutIndex` before indexing `txMeta.Tx.Outputs` in the validator's transaction-validation path. A child spending a non-existent vout of a real parent (e.g. vout 2 on a 2-output parent) still panics the validator with `index out of range` on v0.15.4 — an operator hit exactly this in prod (`getUtxoBlockHeightAndExtendForParentTx`). It is reachable via tx/subtree validation of peer-supplied transactions, so it's a remotely-triggerable crash/DoS.

## Scope
This fixes the validator-helper path only — the exact path the reported crash occurred on. It is a faithful, complete backport of #1243 (which likewise only touched this helper).

It does **not** cover the block-validation quick-validate path (`services/blockvalidation/quick_validate.go`), which dereferences the same untrusted vout unguarded and is a separate, pre-existing gap present on `main` too. Raised by @ctnguyen in review (ChiR1) and tracked as a main-first follow-up — see #1283.

## Fix
Adds the vout bound to the extend path in `getUtxoBlockHeightAndExtendForParentTx`. Because Go `||` short-circuits, `int(vout) >= len(txMeta.Tx.Outputs)` runs before the deref, so an out-of-range vout returns a clean processing error instead of panicking. The guard is identical to the one already on `main` from #1243.

## Test
Standalone store-based regression test. The upstream PR's test (`validator_prefetched_parents_test.go`) depends on prefetched-parents (#1118), which is not on this release train, so the test is rewritten against the UTXO-store path. It reproduces the panic on current `release/v0.15` and passes with the guard.

## Verification
- Test fails (panics `index out of range [99] with length 2`) on unpatched `release/v0.15`, passes after the guard.
- Full validator package green under `-race`; `go vet`, `gofmt`, `golangci-lint` clean.

## Ship path
Cut `v0.15.5-beta-1` from this, have the affected operator confirm no panic, then promote to `v0.15.5`. Block-validation path (#1283) handled separately, main-first.

Refs #1243

### DIFF
--- a/services/validator/Validator.go
+++ b/services/validator/Validator.go
@@ -1025,14 +1025,22 @@ func (v *Validator) getUtxoBlockHeightAndExtendForParentTx(gCtx context.Context,
 					tx.TxIDChainHash().String(), idx, len(tx.Inputs))
 			}
 
-			if txMeta.Tx == nil || txMeta.Tx.Outputs == nil || txMeta.Tx.Outputs[tx.Inputs[idx].PreviousTxOutIndex] == nil {
-				return errors.NewProcessingError("[Validate][%s] parent transaction %s does not have outputs for input index %d",
-					tx.TxIDChainHash().String(), parentTxHash.String(), idx)
+			// PreviousTxOutIndex comes from the (untrusted) child transaction, so
+			// bound it against the parent's output count BEFORE indexing. Otherwise a
+			// tx referencing a real parent but a non-existent vout (e.g. vout 2 on a
+			// 2-output parent) panics here with index out of range and crashes the
+			// validator. This is the reachable half of #1243 that the v0.15.4 backport
+			// (ea6d8ce76) omitted.
+			vout := tx.Inputs[idx].PreviousTxOutIndex
+			if txMeta.Tx == nil || txMeta.Tx.Outputs == nil ||
+				int(vout) >= len(txMeta.Tx.Outputs) || txMeta.Tx.Outputs[vout] == nil {
+				return errors.NewProcessingError("[Validate][%s] parent transaction %s has no output for index %d",
+					tx.TxIDChainHash().String(), parentTxHash.String(), vout)
 			}
 
 			// extend the input with the parent tx outputs
-			tx.Inputs[idx].PreviousTxSatoshis = txMeta.Tx.Outputs[tx.Inputs[idx].PreviousTxOutIndex].Satoshis
-			tx.Inputs[idx].PreviousTxScript = txMeta.Tx.Outputs[tx.Inputs[idx].PreviousTxOutIndex].LockingScript
+			tx.Inputs[idx].PreviousTxSatoshis = txMeta.Tx.Outputs[vout].Satoshis
+			tx.Inputs[idx].PreviousTxScript = txMeta.Tx.Outputs[vout].LockingScript
 		}
 	}
 
--- a/services/validator/validator_vout_bounds_test.go
+++ b/services/validator/validator_vout_bounds_test.go
@@ -1,0 +1,52 @@
+package validator
+
+import (
+	"context"
+	"testing"
+
+	"github.com/bsv-blockchain/go-bt/v2"
+	"github.com/bsv-blockchain/go-bt/v2/chainhash"
+	"github.com/bsv-blockchain/teranode/settings"
+	utxostore "github.com/bsv-blockchain/teranode/stores/utxo"
+	"github.com/bsv-blockchain/teranode/stores/utxo/meta"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+// Test_getUtxoBlockHeightAndExtendForParentTx_VoutOutOfRange is a regression test
+// for the validator crash reported on v0.15.2/v0.15.4: a child tx whose input
+// references a non-existent output index (vout) of a real parent panicked at
+// txMeta.Tx.Outputs[PreviousTxOutIndex] with "index out of range" and killed the
+// validator process. The extend path must reject an out-of-range vout with a
+// clean processing error, not panic.
+//
+// The v0.15.4 backport (ea6d8ce76) ported only the unreachable idx>=len half of
+// #1243 and omitted this reachable bound, so the deref stayed unguarded.
+func Test_getUtxoBlockHeightAndExtendForParentTx_VoutOutOfRange(t *testing.T) {
+	ctx := context.Background()
+
+	// Child tx: single validly-indexed input (idx 0) whose PreviousTxOutIndex
+	// points past the parent's outputs.
+	childTx := &bt.Tx{Inputs: []*bt.Input{{PreviousTxOutIndex: 99}}}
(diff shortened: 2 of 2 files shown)


icellan (approved):

freemans13 (approved):

ctnguyen (changes requested):
Review at commit [a865aaf](https://github.com/bsv-blockchain/teranode/commit/a865aaf)

The change to the validator helper is minimal and correct. Because Go's `||` short-circuits, `int(vout) >= len(txMeta.Tx.Outputs)` is evaluated before `txMeta.Tx.Outputs[vout]`, so an out-of-range vout on a real parent returns a clean processing error instead of panicking — and the guard is byte-identical to the one already merged to `main` from #1243, which is exactly what you want in a backport. The regression test is meaningful rather than ceremonial: it drives the `extend=true` store path with a child input pointing at vout 99 on a 2-output parent, reproducing the reported prod crash and asserting the new error string. The PR description is exemplary — it pins the omitted half to the specific v0.15.4 SHA, explains why the upstream test was rewritten (prefetched-parents #1118 isn't on this train), and lays out a sensible beta-then-promote ship path.

## Blocking issues

### ChiR1 — Same-block-parent extend path in block validation still panics on the same untrusted vout

The PR body states the crash is "reachable via subtree/block validation of transactions from peers," but the fix only closes the `Validator.getUtxoBlockHeightAndExtendForParentTx` store path. On `release/v0.15`, `services/blockvalidation/quick_validate.go` dereferences the same untrusted vout without a bound, in two places:

```go
// quick_validate.go:1104-1105 and :1442-1443
tx.Inputs[j].PreviousTxSatoshis = parentTx.Outputs[input.PreviousTxOutIndex].Satoshis
tx.Inputs[j].PreviousTxScript   = parentTx.Outputs[input.PreviousTxOutIndex].LockingScript
```

Here `parentTx` comes from the same-block map (`extendedTxsFromPrevBatches[*parentHash]`, reached from the batch-extend callers at :525 and :656) and `input.PreviousTxOutIndex` is attacker-controlled. A malicious block/subtree that places a real parent in an earlier batch and a child later with `PreviousTxOutIndex >= len(parentTx.Outputs)` panics with `index out of range` here — bypassing `BatchPreviousOutputsDecorate` and crashing before validation can return a clean error. That is the same remotely-triggerable DoS this PR sets out to eliminate, so the fix is only partially effective on this release train.

Why it matters: the stated threat is peer-supplied txs via subtree/block validation; leaving this path open means an operator on the resulting `v0.15.5` can still be crashed by exactly the input class the release notes will claim is fixed.

Fix: add the identical bound before both derefs, plus a regression test on the same-block-parent path:

```go
vout := input.PreviousTxOutIndex
if parentTx.Outputs == nil || int(vout) >= len(parentTx.Outputs) || parentTx.Outputs[vout] == nil {
    // return a clean processing error for this tx/block instead of indexing
}
```

It's defensible to land this as an immediate follow-up PR rather than expand this one — but the incompleteness should be resolved (or explicitly tracked and disclosed in the release notes) before `v0.15.5` is promoted, so it is raised as blocking. Note this gap also exists on `main` and is pre-existing, not introduced here.

## Non-blocking issues

### ChiR2 — `int(vout)` cast is only guaranteed safe on 64-bit targets

`vout` is `uint32`. On a 64-bit build `int(vout)` is always the exact non-negative value, so the bound holds. On a 32-bit build `int` is 32-bit: a `vout` above `math.MaxInt32` wraps to negative, `int(vout) >= len(...)` is then false, and the subsequent `txMeta.Tx.Outputs[vout]` (indexed with the raw `uint32`) panics again — the very crash this PR removes.

Why it matters: it's a latent re-introduction of the same DoS on 32-bit. In practice Teranode is a 64-bit node so this is theoretical, and the same pattern already lives on `main`, so it is not backport-specific — but if you ever want it airtight, compare in a width that can't wrap:

```go
if txMeta.Tx == nil || txMeta.Tx.Outputs == nil ||
    uint64(vout) >= uint64(len(txMeta.Tx.Outputs)) || txMeta.Tx.Outputs[vout] == nil {
```

Since the goal is to mirror `main`, the cleaner path is to fix it upstream first and re-backport, rather than diverge here.

### ChiR3 — Test covers only the over-range case

The regression test exercises `vout = 99` (well past the end) but not the boundary `vout == len(Outputs)` (off-by-one, the case most likely to regress if someone later rewrites the comparison), nor the `Tx == nil` / `Outputs == nil` / nil-element branches that share this guard. Adding the `vout == 2` boundary sub-case would lock in the `>=` (vs `>`) semantics cheaply. Not required for the fix, but worth a follow-up given this is the exact hardening path.

## Recap

| ChiR1 | Same-block parent deref in block validation still panics | YES | 80% |
| ChiR2 | `int(vout)` cast unsafe on 32-bit | | 20% |
| ChiR3 | Test omits boundary/nil cases | | 12% |

oskarszoon:
@ctnguyen ChiR1 confirmed — verified the derefs at `quick_validate.go:1104-1105` and `:1442-1443` are unguarded, and that it's reachable: in all three processing variants the extend runs during batch processing and `validateSubtrees`→`CheckMerkleRoot` runs after (:451 / :590 / :728), so a peer serving a below-checkpoint block with a tampered/corrupt tx panics at the deref before the merkle check rejects it. The below-checkpoint gating doesn't save it.

Handling it as a separate main-first follow-up rather than expanding this PR: #1283. Reasons — the reported prod crash was the validator-helper path (`getUtxoBlockHeightAndExtendForParentTx`), which this PR closes and which is a faithful, complete backport of #1243; and the quick-validate gap is pre-existing on `main` too, so guarding it only on `release/v0.15` here would re-introduce a main/release divergence (the exact failure mode that produced the broken v0.15.4 backport). Fixing main first and backporting keeps parity.

ChiR2 (uint64 width-safety) folded into that follow-up — agree it's cleaner to fix the comparison on main and backport than to diverge.

ChiR3 (boundary/nil sub-cases): this PR's test drives the over-range case that reproduces the prod panic; the boundary and nil branches go in the follow-up alongside the quick-validate test.

Tightened the PR description to scope the claim to the validator path. Re-review when you get a chance.

ctnguyen (approved):
No code changed since the prior pass — head is still `a865aaf`. The maintainer instead (a) tightened the PR description to disclose the uncovered `quick_validate.go` path, and (b) opened follow-up #1283 to fix ChiR1/ChiR2/ChiR3 main-first then backport, avoiding the main/release divergence that produced the broken v0.15.4 backport. Deferring the two non-blocking items to #1283 is a defensible plan. ChiR1 is judged PARTIALLY FIXED because its own acceptance criterion allowed "resolved OR explicitly tracked and disclosed": the concern is now acknowledged, tracked (#1283), and disclosed in the PR body — but the actual DoS still exists on this release train until #1283 lands, and the disclosure currently lives in the PR body, not yet in released notes.

PR #1276, opened 2026-07-14
MvP-1150 Run ExpiringConcurrentCache fetchFunc without holding the write lock

### LINKED ISSUE #1150: [TN-011][medium] util: ExpiringConcurrentCache.GetOrSet runs fetchFunc under the write lock
**Severity:** medium · **Confidence:** high · **Pass:** concurrency/performance
**Component:** `util/expiring_concurrent_cache.go`; sole caller `stores/utxo/aerospike` (`externalTxCache`)

### Summary
The single-flight cache invokes the caller-supplied `fetchFunc()` while holding the global write lock. A single in-flight miss blocks all other `GetOrSet` callers (even for different keys) for the whole fetch, and any `fetchFunc` that re-enters the cache deadlocks. (The separately-claimed `wgw.result` data race was investigated and dismissed — the WaitGroup synchronises it.)

### Evidence
- `util/expiring_concurrent_cache.go:94` `Unlock()` then `:97` `Lock()`, `:99-104` deferred unlock, `:107` `fetchFunc()` under lock.
- Sole prod caller: `stores/utxo/aerospike/get.go:1447,1536` — fetch reads an external (large) tx (I/O); cache at `aerospike/aerospike.go:227` (10s TTL).

### Why it matters
On large-tx-heavy workloads, external-tx lookups serialise through one mutex held across I/O, collapsing concurrency to one in-flight fetch globally and adding tail latency; a re-entrant `fetchFunc` would hard-deadlock.

### Suggested fix
Drop `c.mu` before `fetchFunc()`; re-acquire only to publish the result. The per-key `wg` map already serialises duplicate fetches.

### Suggested tests
`-race`: two goroutines `GetOrSet` different keys with blocking fetches → assert concurrent. Re-entrancy: a `fetchFunc` calling `GetOrSet` on another key returns instead of deadlocking.

---

ctnguyen:
## Problem

`ExpiringConcurrentCache.GetOrSet` invoked the caller-supplied `fetchFunc()` while
holding the global write lock `c.mu`. A single in-flight miss blocked every other
`GetOrSet` call — even for different keys — for the whole duration of the fetch, and
any `fetchFunc` that re-entered the cache hard-deadlocked.

The sole production caller (`stores/utxo/aerospike` `externalTxCache`) does external-store
I/O inside `fetchFunc` to read large transactions, so on large-tx-heavy workloads external-tx
lookups serialised through one mutex held across I/O, collapsing concurrency to one in-flight
fetch globally and adding tail latency.

## Change

- Run `fetchFunc()` **without** holding `c.mu`; re-acquire the lock only to publish the
  result and clean up the per-key waitgroup entry. The per-key `wg` map already serialises
  duplicate fetches for the same key, so single-flight behaviour is unchanged.
- Publishing (cache set + result), `wg.Done()`, and the map delete now all happen under
  `c.mu` in that order, gated by an explicit success flag so a panicking `fetchFunc`
  publishes nothing and re-propagates the panic.
- TTL, `allowCache` handling, error propagation, and waiter fall-through are preserved.

## Tests

Two `-race` regression tests added, both with bounded timeouts so a regression fails fast:

- concurrent fetches for **different keys** run in flight simultaneously (times out on the
  old lock-under-fetch code);
- a `fetchFunc` that **re-enters** `GetOrSet` for another key completes instead of
  deadlocking.

## Verification

- `go test -race ./util/` — pass
- `go test -race -tags test_util ./util/` — pass
- `go vet ./util/` and `go vet ./stores/utxo/aerospike/` — clean
- `golangci-lint run` on the changed files — 0 issues

## Scope

No public API/signature change, no data-structure redesign, no caller changes; the fix
makes re-entrancy safe regardless.

---

Resolves https://github.com/bsv-blockchain/teranode/issues/1150

### DIFF
--- a/util/expiring_concurrent_cache.go
+++ b/util/expiring_concurrent_cache.go
@@ -11,7 +11,14 @@ import (
 type expiringConcurrentCacheWait[V any] struct {
 	wg     *sync.WaitGroup
 	result *V
+	err    error // leader's fetch error, shared verbatim with waiters on this flight
 }
+
+// testHookWaiterAboutToWait, when non-nil, is invoked by a waiting goroutine immediately after it
+// has captured the in-flight entry and released c.mu, just before it blocks on the flight's
+// WaitGroup. It is nil in production (a single nil check on the already-contended waiter path, no
+// behavior change) and is set only by tests needing deterministic leader/waiter ordering.
+var testHookWaiterAboutToWait func()
 
 type ExpiringConcurrentCache[K comparable, V any] struct {
 	mu        sync.RWMutex
@@ -69,6 +76,11 @@ func (c *ExpiringConcurrentCache[K, V]) GetOrSet(key K, fetchFunc func() (V, boo
 	// If not, check if there is an ongoing request
 	if wgw, found = c.wg[key]; found {
 		c.mu.Unlock()
+
+		if testHookWaiterAboutToWait != nil {
+			testHookWaiterAboutToWait()
+		}
+
 		wgw.wg.Wait() // Wait for the other goroutine to finish
 
 		if val, found = c.cache.Get(key); found {
@@ -78,6 +90,11 @@ func (c *ExpiringConcurrentCache[K, V]) GetOrSet(key K, fetchFunc func() (V, boo
 		// check the result in the wait group
 		if wgw.result != nil {
 			return *wgw.result, nil
+		}
+
+		// share the leader's real error with waiters, if the fetch failed
+		if wgw.err != nil {
+			return c.ZeroValue, wgw.err
 		}
 
 		return c.ZeroValue, errors.NewProcessingError("cache: failed to get value after waiting")
@@ -99,13 +116,18 @@ func (c *ExpiringConcurrentCache[K, V]) GetOrSet(key K, fetchFunc func() (V, boo
 	defer func() {
 		c.mu.Lock()
 
-		if fetchOK && err == nil {
-			// Cache the result
-			if allowCache {
-				c.cache.Set(key, val)
-			}
+		if fetchOK {
+			if err == nil {
+				// Cache the successful result
+				if allowCache {
+					c.cache.Set(key, val)
+				}
 
-			c.wg[key].result = &val
+				c.wg[key].result = &val
+			} else {
+				// Share the leader's real error with any waiters on this flight
+				c.wg[key].err = err
+			}
 		}
 
 		wg.Done()         // after publish so waiters observe result via Done/Wait
--- a/util/expiring_concurrent_cache.go
+++ b/util/expiring_concurrent_cache.go
@@ -90,31 +90,37 @@ func (c *ExpiringConcurrentCache[K, V]) GetOrSet(key K, fetchFunc func() (V, boo
 		wg: wg,
 	}
 
-	// Release the global lock, for others to wait on the wait group
+	// Release the global lock, for others to wait on the wait group.
 	c.mu.Unlock()
 
-	// Perform the fetch, with a lock on the cache
-	c.mu.Lock()
+	// Publish the result and clean up under the lock, but run fetchFunc() WITHOUT the lock.
+	fetchOK := false
 
 	defer func() {
-		wg.Done()         // Mark the wait group as done
-		delete(c.wg, key) // Remove it from the map
+		c.mu.Lock()
+
+		if fetchOK && err == nil {
+			// Cache the result
+			if allowCache {
+				c.cache.Set(key, val)
+			}
+
+			c.wg[key].result = &val
+		}
+
+		wg.Done()         // after publish so waiters observe result via Done/Wait
+		delete(c.wg, key) // remove in-flight entry, still under c.mu
 
 		c.mu.Unlock()
 	}()
 
-	// Perform the fetch
+	// Perform the fetch WITHOUT holding c.mu.
 	val, allowCache, err = fetchFunc()
+	fetchOK = true // reached only if fetchFunc returned (did not panic)
+
 	if err != nil {
 		return c.ZeroValue, err
 	}
-
-	// Cache the result
-	if allowCache {
-		c.cache.Set(key, val)
-	}
-
-	c.wg[key].result = &val
 
 	return val, nil
 }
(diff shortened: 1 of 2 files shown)


icellan (commented):
The lock-narrowing itself looks correct — running `fetchFunc()` outside `c.mu` fixes the different-key re-entrancy deadlock, and the publish/cleanup is properly re-guarded under the lock with `fetchOK` covering the panic case. Two correctness gaps remain, both pre-existing but now in scope since this PR adds a re-entrancy test:

**`util/expiring_concurrent_cache.go:72` — same-key re-entrant fetch still self-deadlocks.** Different-key re-entrancy is fixed, but a `fetchFunc` that calls `GetOrSet` for the *same* key hangs forever: the outer entry is still in `c.wg[key]`, the inner call blocks on `wgw.wg.Wait()`, and that WaitGroup is only `Done()` by the outer defer — which can't run because the outer fetch is blocked on the inner call. Permanent goroutine leak. The new re-entrancy test only covers different keys, so this path is untested. Add a same-key test and define the intended behavior (detect + error, or document as unsupported).

**`util/expiring_concurrent_cache.go:83` — waiters get an opaque error on leader failure and don't re-fetch.** When the leader's `fetchFunc` errors, publish is gated on `fetchOK && err == nil`, so nothing is cached/published. Waiting goroutines miss the cache, see `result == nil`, and return the generic `"cache: failed to get value after waiting"` instead of the leader's real error — with no retry. A transient failure becomes opaque for every concurrent waiter. Propagate the leader's error to waiters, or let one waiter re-fetch.

ordishs (approved):
Correct, minimal fix for a genuine global-serialization/deadlock bug: fetchFunc now runs without holding c.mu, with publish, wg.Done(), and the in-flight map delete done atomically under the lock and gated by an explicit success flag so a panicking fetchFunc publishes nothing. Per-key single-flight is unchanged. Both regression tests are well-designed (barrier-based concurrency proof, bounded timeouts) and pass locally with -race.

Two optional polish points, non-blocking:
- In the defer, capture the wait entry pointer at creation instead of re-looking up c.wg[key] — makes the safety self-evident rather than resting on the only-owner-deletes invariant.
- Same-key re-entrancy still self-deadlocks (inherent to single-flight); a one-line caveat on the GetOrSet doc comment would avoid the implied blanket re-entrancy guarantee.

ctnguyen:
Thanks @icellan for the feedback — both points are addressed.

### Concern 1 — same-key re-entrant fetch (`expiring_concurrent_cache.go`)
Rather than detect re-entry at runtime (goroutine-ID inspection is brittle and, worse, has a real
blind spot: a `fetchFunc` that re-enters the same key *from a new goroutine it then waits on* has a
different goroutine id and would slip past any such guard), the contract is now **documented as
unsupported** on `GetOrSet`:

> `fetchFunc` must not call `GetOrSet` for the SAME key it is fetching — neither directly nor from a
> goroutine it then waits on … Re-entrancy for DIFFERENT keys is fully supported.

Same-key re-entry is genuinely undecidable to detect reliably, and no production caller does it
(`externalTxCache` in both `GetOutpointsFromExternalStore` and `GetTxFromExternalStore` fetches
distinct keys). The requested **same-key test** is added as `TestGetOrSetReentrantSameKeyDeadlocks`:
it exercises the unsupported deadlock in a **re-exec child process** (marker emitted at the actual
re-entry point; parent asserts the child does not exit within the bound, then kills and reaps) so no
permanently-blocked goroutine leaks into the test binary.

### Concern 2 — waiters no longer get an opaque error (`expiring_concurrent_cache.go`)
The leader's real error is now shared with waiters. A new `err` field on the per-key wait struct is
set under `c.mu` in the leader's `defer` when the fetch fails (before `wg.Done()`); waiters read it
after `wg.Wait()` and return it **verbatim** instead of the generic
`"cache: failed to get value after waiting"`. The `Done()`→`Wait()` happens-before edge makes the
field visible without extra locking (same guarantee the existing `result` read relies on). Errors are
never cached, so a later call for the same key re-fetches — a transient failure self-heals. This
mirrors `golang.org/x/sync/singleflight` (shared error, no per-waiter re-fetch / thundering herd).

icellan (approved):

icellan (commented):
Re-reviewed at 18cce6dc1. Both P1s resolved and verified against the code (ran the util cache tests with `-race`, all green):

- **Same-key re-entrant deadlock** — documented as unsupported (`expiring_concurrent_cache.go:48-52`) and pinned by `TestGetOrSetReentrantSameKeyDeadlocks`, which proves the deadlock in a bounded re-exec child and reaps it, so nothing leaks in the parent. Test gap closed.
- **Waiters got an opaque error** — leader's real error is now stored on the flight (`:135`) and returned verbatim to waiters (`:102-104`). `TestGetOrSetWaitersReceiveLeaderError` asserts pointer-identity propagation, zero re-fetches, and absence of the generic message, using a deterministic hook rather than sleeps. `TestGetOrSetErrorNotCachedRetriesNextCall` confirms failures aren't cached and the next call retries.

Two nits, neither blocking:

- `expiring_concurrent_cache_unit_test.go:637,653,690,732,751,767,779` — goroutine startup/drain is still bounded by hard `time.After(2 * time.Second)` `t.Fatal`s. The ordering-sensitive logic uses the deterministic hook now, but these wall-clock caps can still flake on a loaded CI runner. Consider 5–10s or adaptive.
- `expiring_concurrent_cache_unit_test.go:659,660,695` — trailing value checks use `assert.Equal` where AGENTS.md wants `require`. The critical assertions already use `require`; only these slipped.

ctnguyen:
Follow-up nits

@icellan (test-only, `util/expiring_concurrent_cache_unit_test.go`):
✅ `assert.Equal` → `require.Equal` on the 3 trailing value checks (659/660/695).
✅ Raised the 7 happy-path `time.After(2s)` guards to `10s` to reduce CI flake; deadlock-test 30s/3s/1s left untouched.

@ordishs (`util/expiring_concurrent_cache.go`):
✅ Capture the wait-entry pointer at creation and use it in the defer instead of re-looking up `c.wg[key]` (behavior-identical).
✅ Same-key re-entrancy doc caveat 

**Overlap:** @ordishs's doc-caveat point and @icellan's earlier same-key deadlock P1 are the same concern (documenting that `fetchFunc` must not re-enter for the same key).

PR #1281, opened 2026-07-14
fix(legacy): rotate sync peer on OnBlock pre-admission timeout (#1277)

### LINKED ISSUE #1277: legacy prefetch: OnBlock pre-admission timeout strands / reprocesses blocks (follow-up from #1190)
Follow-up from #1190 (bounded async block prefetch for legacy catchup). Non-blocking per review; filed to merge #1190 as-is.

The pre-admission liveness bound added in #1190 (`preAdmitCtx` + `checkBannedBounded`) correctly stops a wedged `handleQuery` or hung `GetBlockHeader` from parking the read-loop with the per-message watchdog disarmed. But the two timeout branches it introduced regress behaviour vs the pre-prefetch path (where the armed watchdog disconnected + rotated on such a stall):

### 1. Ban-check timeout silently strands a requested block
`services/legacy/peer_server.go` — when `checkBannedBounded` returns `ok=false` (deadline hit before a reply, e.g. a backed-up `handleQuery`), `OnBlock` drops the block and keeps the peer connected, but never clears the block's `requestedBlocks` entry. Since it stays "requested", the getdata logic won't re-request it → single-block IBD stall even after `handleQuery` recovers.

### 2. Header-lookup timeout reprocesses an already-known-valid block
`services/legacy/peer_server.go` — `GetBlockHeader` now runs under `preAdmitCtx`, so a timed-out lookup errors → `blockIsKnownValid=false` → a block we already have is admitted against the budget and fully re-validated instead of skipped. Under blockchain-service load with a low `legacy_peerProcessingTimeout` that is redundant CPU + budget the old untimed `sp.ctx` lookup avoided.

### Suggested fix (unified)
Treat a pre-admission **timeout** distinctly from a genuine result: distinguish `preAdmitCtx.Err() != nil` from a real not-found. On timeout, **drop + re-request** (clear the `requestedBlocks` entry via a small netsync helper, or rotate) rather than "proceed and reprocess"; on a genuine not-found, process as today. This fixes both branches with one change and avoids both the stall and the reprocess.

Parent: #1189

oskarszoon:
Fixes #1277. Follow-up from #1190; parent #1189.

## Problem
PR #1190's pre-admission liveness bound (`preAdmitCtx`) let `OnBlock` treat a context expiry as if it were a real result, regressing two cases versus the pre-prefetch armed-watchdog path:

1. **Ban-check timeout stranded a requested block.** `checkBannedBounded` `ok=false` dropped the block but never cleared its `requestedBlocks` entry. In headers-first mode `fetchHeaderBlocks` has already advanced `startHeader` past the hash, so nothing re-requests it → single-block IBD stall.
2. **Header-lookup timeout reprocessed a known-valid block.** A `GetBlockHeader` that erred under a fired `preAdmitCtx` forced `blockIsKnownValid=false`, admitting a block we may already have against the prefetch budget for a full re-validation.

Both share one root: a `preAdmitCtx` timeout is indistinguishable from a genuine result (not-banned / not-found), and each branch picks the wrong degraded behaviour.

## Fix
On a `preAdmitCtx` **deadline** (`context.DeadlineExceeded`, not a parent cancel) in either pre-admission call, rotate the sync peer via `disconnectMisbehaving`. The primary disconnect drives `handleDonePeerMsg → updateSyncPeer → startSync`, which clears `requestedBlocks` and re-drives the fetch from a fresh locator — restoring the pre-prefetch watchdog behaviour and recovering the stranded block.

A bare `requestedBlocks` delete is **not** sufficient: the headers-first cursor (`startHeader`) has already advanced past the hash, so deleting the entry makes the block eligible but nothing issues the request. Rotation is what actually re-drives.

On **parent cancel** (daemon shutdown / peer teardown) the block is dropped without a disconnect, so shutdown does not storm every peer currently inside `OnBlock`.

Extracts `preAdmitTimedOut` for the deadline-vs-cancel decision (mirroring the `checkBannedBounded` / `shouldDisconnectOnBlockErr` pattern) so the policy is a single, unit-tested point.

## Tests
- New `TestPreAdmitTimedOut` (table-driven): deadline → rotate; parent cancel / direct cancel / live ctx → no rotate. The parent-cancel case is load-bearing — it proves shutdown does not disconnect-storm.
- All existing legacy pre-admission tests unchanged and green.

## Verification
```
go vet ./services/legacy/...
go test -race ./services/legacy/            # ok
go test -race ./services/legacy/netsync/    # ok (rotation path)
```

### DIFF
--- a/services/legacy/peer_server.go
+++ b/services/legacy/peer_server.go
@@ -943,11 +943,25 @@ func (sp *serverPeer) OnBlock(_ *peer.Peer, msg *wire.MsgBlock, buf []byte, payl
 
 	if err == nil {
 		// Ban-check round-trip bounded by preAdmitCtx. ok=false means the query
-		// send or its reply was cancelled/timed out before an answer arrived, so
-		// drop the block rather than park the read-loop on a wedged query goroutine.
+		// send or its reply was cancelled/timed out before an answer arrived.
 		isBanned, ok := sp.checkBannedBounded(preAdmitCtx, host)
 		if !ok {
-			sp.server.logger.Warnf("dropping block %s: ban-check timed out/cancelled for peer %s", msg.BlockHash(), sp)
+			// A pre-admission DEADLINE means the query path is wedged: the block
+			// was solicited (netsync marked it requested and fetchHeaderBlocks
+			// already advanced startHeader past it), so silently dropping it
+			// strands the hash — nothing re-requests it and IBD stalls on this
+			// single block. Rotate the sync peer instead: the primary disconnect
+			// drives handleDonePeerMsg → updateSyncPeer → startSync, which clears
+			// requestedBlocks and re-drives the fetch from a fresh locator
+			// (restoring the pre-prefetch watchdog behaviour). On parent cancel
+			// (daemon shutdown / peer teardown) just drop the block.
+			if preAdmitTimedOut(preAdmitCtx) {
+				disconnectMisbehaving(sp, fmt.Sprintf("pre-admission ban-check timed out for block %s, disconnecting to trigger sync peer rotation", msg.BlockHash()))
+				return
+			}
+
+			sp.server.logger.Warnf("dropping block %s: ban-check cancelled by teardown for peer %s", msg.BlockHash(), sp)
+
 			return
 		}
 
@@ -979,10 +993,23 @@ func (sp *serverPeer) OnBlock(_ *peer.Peer, msg *wire.MsgBlock, buf []byte, payl
 	sp.AddKnownInventory(iv)
 
 	// single round-trip: GetBlockHeader tells us both existence and validity.
-	// preAdmitCtx bounds this lookup (see the pre-admission timeout above); a
-	// timed-out lookup returns an error → blockIsKnownValid=false, so the block
-	// proceeds as "not known valid" exactly as an errored lookup does today.
+	// preAdmitCtx bounds this lookup (see the pre-admission timeout above).
 	_, meta, err := sp.server.blockchainClient.GetBlockHeader(preAdmitCtx, blockHash)
+	if err != nil && preAdmitCtx.Err() != nil {
+		// The lookup was cut short by the pre-admission context, so the error
+		// says nothing about whether we already have the block. Proceeding as
+		// "not known valid" would charge a possibly known-valid block against
+		// the prefetch budget and fully re-validate it. On a genuine deadline,
+		// treat the wedged lookup like the wedged ban-check above: rotate the
+		// sync peer so netsync re-drives the fetch. On parent cancel (daemon
+		// shutdown / peer teardown) just drop the block.
+		if preAdmitTimedOut(preAdmitCtx) {
+			disconnectMisbehaving(sp, fmt.Sprintf("pre-admission header lookup timed out for block %s, disconnecting to trigger sync peer rotation", blockHash))
+		}
+
+		return
+	}
+
 	blockIsKnownValid := err == nil && !meta.Invalid
 
 	if !blockIsKnownValid {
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):
Review at commit [b7ab1bd](https://github.com/bsv-blockchain/teranode/commit/b7ab1bd)

A tight, correct regression fix: it restores the pre-prefetch armed-watchdog recovery for the two pre-admission cases #1190 flattened, and does so without widening the blast radius.

## Non-blocking issues

### ChiR1 — Rotation wiring in `OnBlock` is only tested at the predicate
**Observed:** `TestPreAdmitTimedOut` covers the deadline/parent-cancel/direct-cancel/live matrix of the helper, but nothing exercises `OnBlock` actually invoking `disconnectMisbehaving` on the ban-check `ok=false` and header-lookup-error branches (vs. dropping). The `return` placement and the `err != nil && preAdmitCtx.Err() != nil` gate are asserted only by reading, not by a test.
**Why it matters:** The predicate is the easy half; the branch selection and early-returns are where a future edit could silently regress (e.g. dropping the `return` after a drop-only path, or inverting the gate).
**Fix:** This is consistent with the repo's existing approach (helpers are unit-tested, `OnBlock` is not), so it's a nice-to-have. If a seam is cheap, a table test driving `OnBlock` with a never-serviced query channel / stubbed `GetBlockHeader` under a fired vs. cancelled `preAdmitCtx`, asserting disconnect-vs-drop, would lock the wiring down.
Criticality: 12%

## Recap
| ChiR1 | `OnBlock` wiring untested | | 12% |

icellan:
**OnBlock pre-admission deadline → disconnect**

- `peer_server.go:1006`/`:958` — the deadline can't tell a slow peer from a wedged *local* dependency. `GetBlockHeader` and the ban-check both hit shared local infra (the header RPC; the single handleQuery goroutine). If that wedges, every peer in OnBlock deadlines and disconnects at once — a fleet-wide sync-peer rotation/reconnect storm that can't clear a local stall (the replacement peer wedges on the same call). Worth bounding (e.g. only rotate the current sync peer, or a circuit-breaker on repeated pre-admission deadlines).
- `peer_server.go:998` — guard keys on `preAdmitCtx.Err() != nil`, not on the error wrapping the ctx error; a real `GetBlockHeader` error coinciding with the deadline gets misclassified as a wedge and disconnects the peer.

freemans13 (approved):
Approving. The deadline-vs-cancel split is the right fix for #1277: a genuine `preAdmitCtx` deadline (wedged local round-trip that stranded a solicited block) rotates the sync peer to re-drive the fetch, while a parent cancel drops the block without a shutdown disconnect-storm. The `preAdmitTimedOut` extraction keeps that policy at a single, unit-tested point, and the parent-cancel test case is the load-bearing one.

Verified locally: `go vet ./services/legacy/` clean, `TestPreAdmitTimedOut` + `TestCheckBannedBounded` green. The header-path gate (`err != nil && preAdmitCtx.Err() != nil`) is correct — not-found is the common case so `err != nil` is near-universal during IBD, and the early return is prevented only by `preAdmitCtx.Err()` being nil on the healthy path. Budget is acquired strictly after both pre-admission calls, so the early returns leak nothing.

A few non-blocking notes, none worth holding the merge:
- The header-lookup teardown-cancel drop returns silently, whereas the ban-check cancel path logs "cancelled by teardown". A matching log line would make a shutdown-time drop traceable.
- On a persistent local wedge (blockchain service/DB hung, not the peer's fault) this disconnects sync peers one after another with no backoff. That's by design — it restores the pre-#1190 armed-watchdog behaviour — and the 3-min default deadline makes it rare, but the failure mode under a real local hang is peer-churn rather than backoff.
- The helper is unit-tested in isolation; the wiring (OnBlock actually calling disconnectMisbehaving on a deadline) isn't covered by a test, which is reasonable given OnBlock's dependencies.

PR #1282, opened 2026-07-14
start up spec tests

sugh01:


ctnguyen (approved):
Review at commit [88fee8b](https://github.com/bsv-blockchain/teranode/commit/88fee8b4cf89ecaacc0381b1e5cf454a22a30f22)

This is a high-quality, purely additive test PR that materially raises confidence in the block-assembly startup/recovery path. Each test maps to a named BA-STARTUP acceptance criterion, drives a distinct failure mode (iterator construction, mid-iteration, missing best-chain header IDs, post-load reconciliation, corrupt checkpoint), and asserts the *refusal* semantics that the pre-existing happy-path suite only proved as "it ran." I verified every asserted error string against the production source and confirmed the mock return shapes cannot panic; the tests compile and will pass. The thoughtful touches — returning a non-nil iterator alongside the error to dodge the mock's unchecked type assertion, omitting the subtree processor so an accidental reach panics loudly, and clearing the genesis anchor before exercising `initState` — show real care about false positives.

## Non-blocking issues

### ChiR1 — Adversarial tests mask a production iterator leak that fires on every clean boot
**Observed:** `TestStartup_RefusesOnMidIterationFailure` and the reconciliation test register `iter.On("Close").Return(nil).Maybe()`, so the iterator's cleanup is neither required nor forbidden. In fact `loadUnminedTransactions` (BlockAssembler.go:2544–2929) never calls `it.Close()` on *any* path — including the normal success path — so the iterator obtained at line 2621 is leaked on every successful startup. Both sibling loaders `defer it.Close()` (BlockAssembler.go:2460, 3067), making this an inconsistency, not an intentional omission.
**Why it matters:** This is a real production resource leak on the normal hot path (a store cursor held until process exit), and it is exactly the class of defect an "adversarial" recovery suite is positioned to catch — yet the permissive `.Maybe()` lets it through.
**Fix:** Make the expectation strict on the mid-iteration path so the gap surfaces:
```go
iter.On("Close").Return(nil).Once()
// ... after loadUnminedTransactions returns:
iter.AssertExpectations(t) // fails today → forces the missing defer it.Close()
```
Then add `defer it.Close()` after line 2621 in production (separate change), matching the two sibling loaders, and the test locks it in.
**Criticality: 42%**

### ChiR2 — Tests never drive `Start()`/FSM, so the "MUST refuse to start" contract is asserted only by a (misattributed) comment
**Observed:** All four `loadUnminedTransactions`/`initState` tests invoke the helpers directly and never call `Start()` or exercise the FSM, yet the docstrings assert the stronger BA-STARTUP-004 contract ("the service MUST refuse to start … MUST NOT enter Running"). The actual error propagation that enforces this lives at `BlockAssembler.go:932–934` (`errors.NewStorageError("[BlockAssembler] failed to load un-mined transactions: %v", err)`), inside `BlockAssembler.Start`. The `TestStartup_RefusesWhenIteratorConstructionFails` docstring misattributes that string to *"Server.go"*.
**Why it matters:** A regression that swallowed the load/`initState` error inside `Start` (e.g. log-and-continue at line 932) would leave all four tests green while violating the contract they claim to protect — the "refuse to start" guarantee is currently only documented, not exercised. The wrong `Server.go` reference also sends a future reader to the wrong file.
**Fix:** Correct the docstring reference to `BlockAssembler.go:932` (`BlockAssembler.Start`), and add at least one end-to-end assertion that drives `Start()` and asserts it returns the wrapped `failed to load un-mined transactions` / `failed to initialize state` error, so the propagation itself is guarded rather than assumed.
**Criticality: 24%**

### ChiR3 — Corrupt-checkpoint test does not actually guard the "distinct arm" it claims
**Observed:** The docstring for `TestStartup_RefusesOnCorruptCheckpointAtLiveHeight` states it exercises "the DecodeState-error arm of initState — distinct from the ErrNoRows arm." But in `initState` (BlockAssembler.go:1223–1266) both a decode error and `ErrNoRows` set `stateFound = false` and converge on the *same* refuse branch; they differ only in an un-asserted log line (`Errorf` vs `Warnf`). The test asserts only `ErrorContains(err, "refusing to start")`, which is identical for both arms.
**Why it matters:** A future change that made the decode-error arm behave differently from the `ErrNoRows` arm would not be caught, despite the comment implying it is protected — the claim overstates the guarantee.
**Fix:** Either soften the docstring to "exercises the decode-error path to the same refuse outcome," or add an assertion that actually distinguishes the arm (e.g. capture the log via `ulogger` and assert the error-level "error getting state" line fired, not the warn-level "no state found").
**Criticality: 18%**

### ChiR4 — Mock expectations are never verified as consumed
**Observed:** No test calls `store.AssertExpectations(t)` / `bc.AssertExpectations(t)`. Configured `.On(...)` stubs that are never hit pass silently.
**Why it matters:** A refactor that reordered startup so, say, `GetBlockHeaderIDs` was no longer called would leave the stub unused and the test still green, weakening the regression value.
**Fix:** Add `t.Cleanup(func() { store.AssertExpectations(t); bc.AssertExpectations(t) })` in `newRecoveryTestAssembler`, or call `AssertExpectations` at the end of each test. (Use `.Maybe()` deliberately only where a call is genuinely optional.)
**Criticality: 8%**

## Recap

| ChiR1 | Iterator leak masked by `.Maybe()`; fires every boot | | 42% |
| ChiR2 | Tests never drive `Start()`/FSM; ref misattributed | | 24% |
| ChiR3 | Corrupt-checkpoint test doesn't guard claimed arm | | 18% |
| ChiR4 | Mock expectations not asserted | | 8% |

liam (approved):

PR #1284, opened 2026-07-14
p2p: track catchup-specific counters in the peer registry

gokutheengineer:
### Problem
`GetPeersForCatchup` populated `CatchupAttempts`/`CatchupSuccesses`/`CatchupFailures` from the peer's generic interaction counters (`InteractionSuccesses`/`InteractionFailures`), which mix in blocks, subtrees, and transactions. `RecordCatchupAttempt` only recorded sync backoff state (`RecordSyncAttempt`), so catchup attempts that produced neither success nor failure were invisible, and the reported "catchup" metrics were really "all interactions". This degraded catchup peer ranking and diagnostics.

### Fix
Added dedicated catchup counters to the centralized registry, following the existing dedicated-RPC pattern (`RecordSyncAttempt`, `RecordCatchupError`):

- `PeerInfo` gains `CatchupAttempts`, `CatchupSuccesses`, `CatchupFailures`.
- New registry methods: `RecordCatchupAttempt` (increments the attempt counter and keeps the existing `LastSyncAttempt`/`SyncAttemptCount` backoff tracking, so prior behavior is preserved), `RecordCatchupSuccess` (increments the counter and records a generic successful interaction with response time, reusing `recordReceivedSuccessLocked`), `RecordCatchupFailure` (increments the counter and records a generic failed interaction). Reputation computation is unchanged.
- Wired through the proto (`RecordCatchupAttempt`/`RecordCatchupSuccess`/`RecordCatchupFailure` RPCs, `PeerRegistryInfo` fields 37-39), the `Blockchain` gRPC handlers, `PeerRegistryClientI` and both implementations.
- The p2p `RecordCatchupAttempt`/`RecordCatchupSuccess`/`RecordCatchupFailure` handlers now call the catchup-specific methods, and `GetPeersForCatchup` reports the catchup-specific counters.
- `ResetReputation` also clears the new counters; JSON persistence picks them up automatically.

Note: pb files regenerated with the repo's canonical toolchain (protoc v33.2, `protoc-gen-go` 1.36.11, `protoc-gen-go-grpc` 1.6.1), so the generated diff contains only the new fields/RPCs - no version-stamp churn.

### Review-round fixes
- **Rolling upgrade**: the registry client's `RecordCatchupAttempt/Success/Failure` fall back to the legacy `RecordSyncAttempt`/`UpdatePeerMetrics` RPCs on `codes.Unimplemented`, so backoff and reputation keep working against an old blockchain service.
- **One success per attempt**: the header-fetch stage no longer reports a catchup success; `doCatchup` reports the single outcome per attempt.
- **Headers-stage reputation credit (`ReportValidBlockHeaders`)**: removing the header-stage catchup-success report initially also removed its generic interaction credit, which broke catchup recovery — with only failures recorded, a fresh peer's reputation dropped to ~5 after one failed attempt (`0*0.6 + 50*0.4 - 15`), `isPeerBad` then excluded the only peer with the data, and the `chainintegrity`/`TestCatchup` e2e tests timed out (the 429 rate-limit storms in those tests fail the block-fetch stage repeatedly, and previously each header-stage success kept the ratio at ~50%). The new `ReportValidBlockHeaders` p2p RPC (mirroring `ReportValidBlock`/`ReportValidSubtree`) restores exactly that credit: generic interaction success + response time, no catchup counter — so `CatchupSuccesses ≤ CatchupAttempts` still holds while a headers-serving peer stays healthy through failed catchup cycles. Rolling upgrade is covered here too: on `codes.Unimplemented` from an old p2p service the client falls back to the legacy `RecordCatchupSuccess` RPC, which on such servers is implemented as exactly this generic credit.
- **Success-rate diagnostics** compute over resolved outcomes (`successes/(successes+failures)`), not raw attempts, since attempts now include in-flight/unresolved catchups.
- **Dashboard**: `p2p_api.PeerRegistryInfo` carries the catchup counters (fields 27-29) and `get_peers.go` serves them under the `catchup_*` JSON names instead of the generic interaction counters. Catchup-scoped timestamps are not tracked; the `catchup_last_*` timestamps remain generic-interaction ones (commented at the mapping site).
- **One failure per attempt**: header-stage failures were reported twice — once inside `catchupGetBlockHeaders` (closest to the failure, and the only report for HTTP-status errors, whose `ServiceError` classification the top-level dispatch deliberately does not report) and again by the top-level catchup handler when the error propagated up. The header stage now wraps its returned errors with an already-reported marker (`markCatchupFailureReported`: a native `ProcessingError` wrap carrying an error-data key, since the errors package flattens foreign wrapper types and would break code matching), and `reportCatchupFailureForError` skips marked errors — so one failed attempt records exactly one `CatchupFailure` and one reputation penalty. The pre-existing double reputation penalty this fixes was surfaced by the new dedicated counter.
- **Persistence**: `persistedRegistryVersion` bumped to 2 so a downgraded binary refuses to re-save (and silently zero) counters it does not understand.

### Tests
- `TestCentralizedPeerRegistry_RecordCatchupCounters`: counters increment correctly, block/subtree activity does not bleed into them, attempts feed sync backoff, success/failure still feed generic interaction metrics, unknown peers are a no-op.
- `TestCentralizedPeerRegistry_ResetReputation`: extended to verify the new counters are cleared.
- `TestGRPC_RecordCatchupCounters`: the three new gRPC handlers against a real registry.
- `TestGRPC_PeerInfoProtoRoundTrip`: extended with the new fields.
- `TestGetPeersForCatchup_UsesCatchupSpecificCounters` (regression for this issue): attempts without an outcome are counted, and non-catchup interactions no longer inflate the reported catchup metrics.
- Existing `TestRecordCatchup*` p2p handler tests extended to assert the new counters.
- `TestPeerRegistryClient_CatchupRPCs_FallBackOnUnimplemented`: rolling-upgrade fallback against a real gRPC server implementing only the legacy RPCs.
- `TestReportValidBlockHeaders_CreditsReputationNotCatchupCounters` and `TestReportValidBlockHeaders_KeepsFailingPeerHealthy`: the headers-stage credit does not touch catchup counters, and a peer serving headers through repeated catchup failures stays healthy (the e2e regression mechanism).
- `TestReportCatchupFailureForError_SkipsAlreadyReported` and `TestMarkCatchupFailureReported_TransparentToErrorCodes`: the already-reported marker suppresses the top-level duplicate, survives teranode error wrapping, and does not break `ErrServiceError`/malicious error-code dispatch.
- `TestSimpleClientReportValidBlockHeaders`: p2p client happy path plus the `Unimplemented` → `RecordCatchupSuccess` rolling-upgrade fallback.

```
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/blockchain/...   # ok
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/p2p/...          # ok
```

### DIFF
--- a/services/asset/httpimpl/get_peers.go
+++ b/services/asset/httpimpl/get_peers.go
@@ -99,10 +99,11 @@ func (h *HTTP) GetPeers(c echo.Context) error {
 			LastBlockTime:   peer.LastBlockTime.Unix(),
 			LastMessageTime: peer.LastMessageTime.Unix(),
 
-			// Interaction/catchup metrics (using the original field names for backward compatibility)
-			CatchupAttempts:        peer.InteractionAttempts,
-			CatchupSuccesses:       peer.InteractionSuccesses,
-			CatchupFailures:        peer.InteractionFailures,
+			// Catchup-specific counters (timestamps below remain the generic
+			// interaction ones; catchup-scoped timestamps are not tracked).
+			CatchupAttempts:        peer.CatchupAttempts,
+			CatchupSuccesses:       peer.CatchupSuccesses,
+			CatchupFailures:        peer.CatchupFailures,
 			CatchupLastAttempt:     peer.LastInteractionAttempt.Unix(),
 			CatchupLastSuccess:     peer.LastInteractionSuccess.Unix(),
 			CatchupLastFailure:     peer.LastInteractionFailure.Unix(),
--- a/services/blockchain/blockchain_api/blockchain_api.pb.go
+++ b/services/blockchain/blockchain_api/blockchain_api.pb.go
@@ -3915,8 +3915,13 @@ type PeerRegistryInfo struct {
 	LastValidatedAt           *timestamppb.Timestamp `protobuf:"bytes,40,opt,name=last_validated_at,json=lastValidatedAt,proto3" json:"last_validated_at,omitempty"`
 	FullStorageContradictions int64                  `protobuf:"varint,41,opt,name=full_storage_contradictions,json=fullStorageContradictions,proto3" json:"full_storage_contradictions,omitempty"` // Count of observed contradictions to full-storage claims
 	FullStoragePenaltyUntil   *timestamppb.Timestamp `protobuf:"bytes,42,opt,name=full_storage_penalty_until,json=fullStoragePenaltyUntil,proto3" json:"full_storage_penalty_until,omitempty"`
-	unknownFields             protoimpl.UnknownFields
-	sizeCache                 protoimpl.SizeCache
+	// Catchup-specific counters, maintained via RecordCatchupAttempt/Success/Failure.
+	// Distinct from the generic interaction counters, which mix in blocks/subtrees/txs.
+	CatchupAttempts  int64 `protobuf:"varint,43,opt,name=catchup_attempts,json=catchupAttempts,proto3" json:"catchup_attempts,omitempty"`
+	CatchupSuccesses int64 `protobuf:"varint,44,opt,name=catchup_successes,json=catchupSuccesses,proto3" json:"catchup_successes,omitempty"`
+	CatchupFailures  int64 `protobuf:"varint,45,opt,name=catchup_failures,json=catchupFailures,proto3" json:"catchup_failures,omitempty"`
+	unknownFields    protoimpl.UnknownFields
+	sizeCache        protoimpl.SizeCache
 }
 
 func (x *PeerRegistryInfo) Reset() {
@@ -4241,6 +4246,27 @@ func (x *PeerRegistryInfo) GetFullStoragePenaltyUntil() *timestamppb.Timestamp {
 		return x.FullStoragePenaltyUntil
 	}
 	return nil
+}
+
+func (x *PeerRegistryInfo) GetCatchupAttempts() int64 {
+	if x != nil {
+		return x.CatchupAttempts
+	}
+	return 0
+}
+
+func (x *PeerRegistryInfo) GetCatchupSuccesses() int64 {
+	if x != nil {
+		return x.CatchupSuccesses
+	}
+	return 0
+}
+
+func (x *PeerRegistryInfo) GetCatchupFailures() int64 {
+	if x != nil {
+		return x.CatchupFailures
+	}
+	return 0
 }
 
 // RegisterPeerRequest registers or updates a peer in the centralized registry.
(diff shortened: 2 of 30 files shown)


ctnguyen (approved):
Review at commit [b0f9ae6](https://github.com/bsv-blockchain/teranode/commit/b0f9ae644b01b3edd78a72ea13eb94d70edcbab9)

This is a clean, well-scoped fix for a genuine correctness problem: the reported "catchup" metrics were really "all interactions". The new dedicated counters follow the existing dedicated-RPC pattern faithfully, and the wiring is complete and consistent end-to-end — proto, generated stubs, gRPC handlers, both `PeerRegistryClientI` implementations, the p2p handlers, and JSON persistence via `peerInfoToProto`/`protoToPeerInfo`. Importantly, the new registry methods preserve prior behaviour exactly: `RecordCatchupSuccess`/`RecordCatchupFailure` reproduce the old `UpdateMetrics` interaction/response-time/reputation side-effects (adding only the catchup counter), and `RecordCatchupAttempt` keeps the `LastSyncAttempt`/`SyncAttemptCount` backoff bump — so reputation is unchanged as claimed. Test coverage is thorough and targeted: bleed-through is explicitly asserted against, the outcome-less attempt case (the actual bug) is regression-tested, and the proto round-trip and reset paths are extended for the new fields.

## Non-blocking issues

### ChiR1 — The same conflated-metrics bug still surfaces on the `/peers` dashboard

The PR fixes the conflation for peer *selection* (`GetPeersForCatchup`), but the operator-facing metrics path is untouched. `services/asset/httpimpl/get_peers.go:103-105` still populates the `catchup_*` JSON keys from `peer.InteractionAttempts`/`InteractionSuccesses`/`InteractionFailures`, and the `p2p/Client.go` `PeerRegistryInfo`→`PeerInfo` conversion (the `PeerRegistryInfo` case, ~lines 767-769) maps only the `Interaction*` fields — it was not extended to carry the three new `CatchupAttempts`/`CatchupSuccesses`/`CatchupFailures` proto fields this PR added (the p2p-side `PeerInfo` struct has no such fields to receive them). So the exact "catchup metrics are really all interactions" problem the PR set out to fix still shows on the `/peers` endpoint.

- **Why it matters:** operators diagnosing catchup behaviour via `/peers` see the same misleading numbers the PR corrects for internal ranking; the fix is only half-visible.
- **Fix (follow-up, arguably out of scope):** add `CatchupAttempts`/`CatchupSuccesses`/`CatchupFailures` to the p2p `PeerInfo` struct, carry them through the `p2p/Client.go` proto conversion, and source the `catchup_*` keys in `get_peers.go` from those fields.

### ChiR2 — Generated pb files stamped with a divergent protoc version

The regenerated `blockchain_api.pb.go` / `blockchain_api_grpc.pb.go` change the `protoc` stamp from `v6.33.2` to `v6.32.0`, while the rest of the repo's generated files are dominantly on `v7.34.1`. This is functionally harmless (the `protoc-gen-go`/`protoc-gen-go-grpc` versions match), but it adds noise to the diff and hurts reproducibility of `make gen` for the next contributor.

- **Why it matters:** committed generated code should be reproducible from the project-standard toolchain; version-stamp drift causes spurious diffs and confusion.
- **Fix:** regenerate with the project-standard protoc (the `v7.34.1`/`v34.1` toolchain) so the stamp matches the surrounding generated files, then re-commit.

### ChiR3 — `ResetReputation` resets the new catchup counters but not the sibling cumulative counters

`resetOne` now zeroes `CatchupAttempts`/`CatchupSuccesses`/`CatchupFailures`, but leaves the other cumulative reception counters untouched — `BlocksReceived`, `SubtreesReceived`, `TransactionsReceived`, and the existing catchup-family `CatchupBlocks`. Since catchup success/failure feed `InteractionSuccesses`/`InteractionFailures` (which the reset does clear), zeroing the catchup counters is defensible; but the treatment across the counter family is now uneven.

- **Why it matters:** it's an easy source of future confusion — after a reset a peer shows `CatchupSuccesses=0` yet may still show a nonzero `CatchupBlocks`/`BlocksReceived`, and it's not obvious from the code which counters are lifetime totals vs reputation-lifecycle state.
- **Fix:** decide the intent explicitly — either leave a short comment in `resetOne` noting that catchup counters are reset alongside the interaction counters they feed while `*Received`/`CatchupBlocks` are lifetime diagnostics, or reset `CatchupBlocks` too for family consistency.

## Recap

| ChiR1 | Conflated catchup metrics still shown on `/peers` dashboard | | 22% |
| ChiR2 | Divergent protoc version stamp in regenerated pb files | | 20% |
| ChiR3 | Uneven counter reset in `ResetReputation` | | 12% |

freemans13:
Reviewed this and it looks solid — the logic is correct and behavior-preserving. I checked the two things most likely to break:

- **Reputation is preserved exactly.** `RecordCatchupSuccess` routes through `recordReceivedSuccessLocked`, which does the identical bookkeeping the old `UpdatePeerMetrics(success=true, …)` path did; `RecordCatchupFailure` sets the same fields as the old `UpdatePeerMetrics(failure=true)` path. Attempts deliberately don't touch `InteractionAttempts` (only the catchup counter + sync-backoff fields), matching the old `RecordSyncAttempt` behavior, so nothing regresses.
- **`RecordSyncAttempt` isn't orphaned** — `sync_coordinator.go` still uses it in three places, and since the new `RecordCatchupAttempt` bumps `SyncAttemptCount` the same way the handler's old `RecordSyncAttempt` call did, there's no new double-counting.

New fields propagate correctly through `clonePeerInfo` (full struct copy) and the proto round-trip. `go vet` clean on both packages; the affected tests pass. Test coverage is genuinely thorough.

One cosmetic thing worth fixing before merge:

**Generated protoc stamp downgraded.** The regeneration moved the stamp in `services/blockchain/blockchain_api/*.pb.go` from `v6.33.2` down to `v6.32.0`, while the sibling `services/p2p/p2p_api/p2p_api.pb.go` still stamps `v6.33` — so the repo now has inconsistent protoc stamps across generated files. It's stamps-only (no wire/descriptor impact), but it churns for the next person who regenerates and could trip a "generated files up to date" CI gate. Suggest regenerating with the repo's canonical protoc (v6.33.2) so the stamp doesn't move. Minor: the PR description says "local protoc v32.0 vs v34.1 used previously," which doesn't match the actual stamp change (v6.33.2 → v6.32.0).

icellan:
Split of catchup counters is the right direction but it's half-shipped — five things to fix before merge:

- **Rolling upgrade breaks all catchup metric recording.** RecordCatchupAttempt/Success/Failure are new RPCs; with an old blockchain service they return Unimplemented and the error is only warn-logged (`services/blockvalidation/peer_metrics_helpers.go`). The old path used RecordSyncAttempt/UpdatePeerMetrics which old servers implement, so reputation penalties and sync backoff silently stop during rollout and blockvalidation keeps re-selecting a broken peer. Needs a fallback to the old RPCs or an explicit compat call-out.
- **Successes can exceed attempts.** blockvalidation reports success twice per catchup (`catchup_get_block_headers.go:454` and `catchup.go:277`) against one attempt (`catchup.go:143`) — one clean catchup logs "2/1 = 200%", and a headers-ok/download-failed catchup shows a 100% success rate.
- **Attempts != successes + failures.** The old derived counter held that invariant; raw attempts now include never-resolved outcomes (ErrBlockIncomplete is deliberately not reported as failure, in-flight/abandoned attempts). A usable peer can read 0% success in selection diagnostics.
- **Dashboard contradicts the new counters.** `services/asset/httpimpl/get_peers.go:103-105` still serves generic interaction counters under `catchup_attempts`/`catchup_successes`/`catchup_failures` JSON names; `p2p_api.PeerRegistryInfo` lacks the new fields so the two surfaces disagree for the same peer.
- **Persistence version not bumped.** New Catchup* fields in the JSON-persisted PeerInfo without bumping `persistedRegistryVersion` (`services/blockchain/peer_registry.go:103`) — downgrade-then-upgrade silently zeroes the counters, the exact failure mode the version guard documents.

icellan (approved):

PR #1285, opened 2026-07-14
fix(p2p): count only connected peers in node_status

gokutheengineer:
### Problem
`ConnectedPeersCount` in the node_status message was set to `len(allPeers)` from `peerRegistry.ListPeers(...)`, which returns every peer in the registry - including gossiped and long-disconnected peers. The dashboard and any consumer of `connected_peers_count` saw an inflated value that misrepresented node connectivity.

### Fix
`getNodeStatusMessage` now counts only peers with `IsConnected == true`, the same filter `GetPeers` already applies to the same `ListPeers` result. The list is filtered client-side because `PeerRegistryClientI.ListPeers` has no connected-only parameter; adding one would touch the blockchain gRPC API for no benefit.

### Tests
Added `TestGetNodeStatusMessage_CountsOnlyConnectedPeers` in `services/p2p/server_helpers_test.go`: two peers registered via `addConnectedPeer` plus one gossiped-only peer via `addPeer`, asserting the message reports 2, not 3. Uses the existing `CentralizedPeerRegistry`-backed local client (no mocks).

Ran (affected package only):
```
```
Result: `ok github.com/bsv-blockchain/teranode/services/p2p 13.300s`

### DIFF
--- a/services/p2p/Server.go
+++ b/services/p2p/Server.go
@@ -1294,14 +1294,19 @@ func (s *Server) getNodeStatusMessage(ctx context.Context) *notificationMsg {
 		s.logger.Debugf("[getNodeStatusMessage] Policy settings not available, using default MinMiningTxFee: %f", defaultFee)
 	}
 
-	// Get connected peers count from the registry
+	// Get connected peers count from the registry. The registry also holds
+	// gossiped/disconnected peers, so count only directly connected ones.
 	connectedPeersCount := 0
 	if s.peerRegistry != nil {
 		allPeers, err := s.peerRegistry.ListPeers(ctx, nil, 0, 0, false, false)
 		if err != nil {
 			s.logger.Warnf("[getNodeStatusMessage] ListPeers failed: %v", err)
 		} else {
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):
Review at commit [5324c74](https://github.com/bsv-blockchain/teranode/commit/5324c7467a0b74464eb7c97ceaa6e5ecc2b06895)

This is a tight, correctly scoped bug fix. The new filter mirrors exactly the `if !p.IsConnected { continue }` logic that `GetPeers` already applies to the same `ListPeers(ctx, nil, 0, 0, false, false)` result, so `connected_peers_count` is now consistent with the peer list the dashboard renders. The decision to filter client-side rather than extend the `PeerRegistryClientI.ListPeers` gRPC signature is well justified and correctly documented. The regression test exercises the real `CentralizedPeerRegistry`-backed local client (no mocks, per project convention) and pins the precise off-by-gossip behaviour the fix targets.

icellan (approved):

PR #1286, opened 2026-07-14
fix(legacy): tear down association sub-peer streams on primary disconnect (#1278)

### LINKED ISSUE #1278: legacy multistream: sub-peer stream not torn down on primary disconnect (follow-up from #1190)
Follow-up from #1190 (bounded async block prefetch for legacy catchup). Non-blocking per review.

Two related gaps share one root cause: **a multistream association's sub-peer teardown is not driven by the primary's disconnect.**

### 6. `disconnectMisbehaving` leaks the DATA1 stream when a bad block arrives on the primary
`services/legacy/peer_server.go` — #1190 fixed the stream→primary direction (bad block on a DATA stream evicts the whole association). The mirror is still open: when a bad block arrives on the **primary** (`target == sp.Peer`), only the primary is disconnected; `handleDonePeerMsg` tears down association bookkeeping on primary disconnect but never closes the sub-peers' live TCP connections, so the DATA1 stream stays open.

### 7. Budget-parked sub-peer read-loop never unparked on a genuine processing hang
`services/legacy/netsync/manager.go` — the pre-admission bound deliberately does not extend to `AcquireBlockPrefetch`. A DATA1 read-loop parked on a full budget during a real processing hang is never released: rotation closes the **primary's** quit channel, not the sub-peer's.

### Suggested fix
Centralize association-stream teardown in `handleDonePeerMsg`: on primary disconnect, close every association stream's quit channel / connection. That covers both #6 (stream connection closed) and #7 (parked sub-peer read-loop unparked), and also the downstream unrequested-block disconnect in `manager.go`. Arguably a pre-existing multistream-teardown weakness that the async prefetch change exposes more of.

Parent: #1189

oskarszoon:
Fixes #1278. Follow-up from #1190; parent #1189.

## Problem
A multistream association's sub-peer teardown was not driven by the primary's disconnect. The server-side `handleDonePeerMsg` removed only association *bookkeeping* (`associationMgr.Remove`) on primary disconnect and never closed the sub-peers' live TCP connections:

1. **DATA1 stream leak on bad-block-at-primary (#&NoBreak;6).** When a bad block arrives on the primary, `disconnectMisbehaving` disconnects only the primary. The DATA1 sub-peer then keeps reading blocks off the wire for an association that no longer exists.
2. **Budget-parked sub-peer never unparked (#&NoBreak;7).** `OnBlock`'s `AcquireBlockPrefetch` waits on the sub-peer's own `serverPeer.quit`, which only closes when the sub-peer's connection drops. Sync-peer rotation closes the *primary's* quit, not the sub-peer's — so a DATA1 read-loop parked on a full budget during a real processing hang stays parked forever.

One root cause: primary disconnect doesn't cascade to sub-peer connection teardown.

## Fix
Centralize sub-peer teardown in `handleDonePeerMsg`'s primary branch via a new `tearDownAssociationStreams(assoc, except)` helper that `DisconnectWithInfo`s every stream peer except the primary (already disconnecting).

The teardown chain closes both gaps at one site:
`subPeer.DisconnectWithInfo` → `close(peer.Peer.quit)` → sub-peer's `peerDoneHandler` `WaitForDisconnect` returns → `close(serverPeer.quit)` → `AcquireBlockPrefetch`'s quit-linked cancel fires → read-loop unparks. The same call closes the TCP connection (the whole of #&NoBreak;6).

- `DisconnectWithInfo` (not Warning): the sub-peer did nothing wrong. Idempotent atomic guard makes the overlap with `disconnectMisbehaving`'s explicit stream disconnect (kept as the immediate fast path) harmless.
- No re-entrant deadlock: the knock-on `donePeers` send runs on the sub-peer's *own* goroutine, and `donePeers` is buffered (`cfg.MaxPeers`). The sub-peer later flows through the STREAM branch (`RemoveStream`) — an idempotent map delete on an already-removed association.

## Tests
- New `TestTearDownAssociationStreams`: builds primary + DATA1 + DATA2, asserts the primary is left connected and both sub-peers are disconnected; plus nil-assoc no-op and idempotency (a pre-disconnected sub-peer) subtests.
- Existing `TestDisconnectMisbehaving` / `TestMisbehaviorDisconnectTarget` unchanged and green.

## Verification
```
go vet ./services/legacy/...
go test -race ./services/legacy/            # ok
go test -race ./services/legacy/netsync/    # ok
```

### DIFF
--- a/services/legacy/peer_server.go
+++ b/services/legacy/peer_server.go
@@ -1192,6 +1192,35 @@ func shouldDisconnectOnBlockErr(err error) bool {
 		!errors.Is(err, errors.ErrStorageError)
 }
 
+// tearDownAssociationStreams closes the live connection of every stream
+// sub-peer in assoc except the given peer. It is the primary-disconnect
+// counterpart to disconnectMisbehaving: when an association's PRIMARY goes away
+// (misbehaviour eviction, sync-peer rotation, remote hang-up), handleDonePeerMsg
+// removes only the association BOOKKEEPING, which left the DATA1 sub-peer's TCP
+// connection open reading blocks off the wire for a dead association, and left a
+// budget-parked sub-peer read-loop parked forever — AcquireBlockPrefetch waits
+// on the sub-peer's own serverPeer.quit, which only closes when the sub-peer's
+// own connection drops. Disconnecting the sub-peer closes its peer.Peer.quit,
+// unblocking its peerDoneHandler, which closes its serverPeer.quit and unparks
+// the waiter. StreamPeers() includes the primary (registered as the GENERAL
+// stream at construction), so callers pass it as except — it is already
+// disconnecting. DisconnectWithInfo is idempotent (atomic guard), so overlap
+// with disconnectMisbehaving's explicit stream disconnect, or a sub-peer already
+// mid-teardown, is safe.
+func tearDownAssociationStreams(assoc *peer.Association, except *peer.Peer) {
+	if assoc == nil {
+		return
+	}
+
+	for _, p := range assoc.StreamPeers() {
+		if p == nil || p == except {
+			continue
+		}
+
+		p.DisconnectWithInfo("association primary disconnected")
+	}
+}
+
 // disconnectMisbehaving evicts a peer's whole association for misbehaviour. It
 // disconnects the association PRIMARY — whose disconnect drives
 // handleDonePeerMsg's associationMgr.Remove(...), rotating the sync peer — and
@@ -1203,11 +1232,13 @@ func shouldDisconnectOnBlockErr(err error) bool {
 // handleDonePeerMsg only does assoc.RemoveStream(...), so openRequiredStreams
 // re-opens DATA1 and the node keeps syncing from a peer serving invalid blocks
 // (IBD stalls). Resolving to the primary is what actually rotates the sync peer.
-// Conversely, handleDonePeerMsg tears down the association bookkeeping on primary
-// disconnect but never closes the sub-peers' connections, so the sub-peer must be
-// disconnected here too or its DATA stream keeps reading off the wire.
-// DisconnectWithWarning is idempotent (atomic guard), so the non-stream case
-// (target == sp.Peer) disconnects exactly once.
+// handleDonePeerMsg now also tears down the association's sub-peer connections on
+// primary disconnect (via tearDownAssociationStreams), so eviction of the whole
+// association is covered even when the bad block arrives on the primary; the
+// explicit stream disconnect here is kept as the fast path (it fires immediately,
+// not one peerHandler round-trip later). DisconnectWithWarning is idempotent
+// (atomic guard), so the non-stream case (target == sp.Peer) disconnects exactly
+// once and the overlap with the centralized teardown is harmless.
 func disconnectMisbehaving(sp *serverPeer, reason string) {
 	target := misbehaviorDisconnectTarget(sp.Peer)
 	target.DisconnectWithWarning(reason)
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):
Review at commit [1d1b6d082d77](https://github.com/bsv-blockchain/teranode/commit/1d1b6d082d77)

A correct, well-scoped fix that closes both leak paths at the single right place. No blocking issues found.

## Non-blocking issues

### ChiR1 — Test asserts the first link of the chain, not the two behaviours the PR fixes
**Problem:** `TestTearDownAssociationStreams` uses `peer.NewInboundPeer(...)` peers that are never connected and have no `peerDoneHandler`/read-loop. `disconnected()` only observes that `peer.Peer.quit` closed (`WaitForDisconnect` returns). It therefore proves the helper skips `except` and closes each sub-peer's `quit`, but it does *not* exercise the actual value of the PR: gap 6 (the TCP connection is closed — `conn.Close()` is skipped because `connected == 0` in the test) or gap 7 (`peer.Peer.quit` → `peerDoneHandler` → `close(serverPeer.quit)` → `AcquireBlockPrefetch` unpark).
**Why it matters:** The fix's correctness rests on the full cascade through existing mechanisms; a regression that broke, say, the `serverPeer.quit` close in `peerDoneHandler` would leave this test green while gap 7 silently reopened.
**Fix:** Not required for merge given the chain is composed of already-tested pieces, but consider an integration-style assertion that a sub-peer parked in `AcquireBlockPrefetch` unparks when the primary's done message is processed by `handleDonePeerMsg` (or at least a comment noting the unit test only covers the first link).
**Criticality: 30%**

### ChiR2 — `disconnected()` helper leaks a goroutine on the negative assertion
**Problem:** For each `require.False(t, disconnected(primary), ...)` call, the helper spawns `go func() { p.WaitForDisconnect(); close(done) }()`. When the peer is never disconnected (the primary), `WaitForDisconnect` blocks on `<-p.quit` forever, so that goroutine never exits and leaks past the test (two leaks across the two subtests that check the primary).
**Why it matters:** Leaked goroutines blocked on a channel can trip goroutine-leak detectors and add noise under `-race`/`-count`; harmless to the assertion itself but avoidable.
**Fix:** Close the peer at end of test, or restructure so the negative case doesn't strand a waiter — e.g. register `t.Cleanup(func(){ primary.DisconnectWithWarning("test cleanup") })`, or poll `primary.Connected()`/a disconnect flag instead of spawning a blocking `WaitForDisconnect` goroutine for the negative check.
**Criticality: 15%**

## Recap
| ChiR1 | Test covers first link, not gaps 6/7 end-to-end | | 30% |
| ChiR2 | Test helper leaks goroutine on primary check | | 15% |

ordishs (approved):
Approving. Verified the key claims against the code and ran the tests locally:

- `DisconnectWithInfo` idempotency (atomic guard in `DisconnectWithLogFunc`), so overlap with `disconnectMisbehaving`'s fast-path stream disconnect and pre-disconnected sub-peers is safe.
- No re-entrant deadlock: teardown only closes quit channels synchronously on the peerHandler goroutine; the knock-on `donePeers` send runs on each sub-peer's own `peerDoneHandler` goroutine into a `cfg.MaxPeers`-buffered channel.
- Sub-peer re-entry through `handleDonePeerMsg`'s stream branch is an idempotent `RemoveStream` on the already-detached association, returning before the peer-list bookkeeping.
- `except` filter is necessary and correct — `NewAssociation` registers the primary as the GENERAL stream, so `StreamPeers()` includes it.
- Teardown-before-`Remove` ordering means stale connections are dropping before the association ID becomes reusable.

`go vet ./services/legacy/...` clean; `go test -race -run 'TestTearDownAssociationStreams|TestDisconnectMisbehaving|TestMisbehaviorDisconnectTarget' ./services/legacy/` all pass.

One residual (pre-existing) gap, filed as #1293: `openRequiredStreams` holds a direct association reference across a blocking dial/streamack handshake (up to ~30s), so a DATA1 stream completing while the primary is tearing down escapes the `StreamPeers()` sweep and attaches a live connection to the removed association. Strictly narrower than before this PR, so not blocking here.

PR #1287, opened 2026-07-14
fix(model): explicit nil-settings guards in the coinbase candidate builders

freemans13:
## What happened

The `NilSettings` edge-case test asserts that `CreateCoinbaseTxCandidate` panics when handed nil settings — and until now that contract was met *implicitly*: the first field dereference tripped Go's nil-pointer conversion into a recoverable panic. That conversion only works while the accessed field's offset stays inside the runtime's ~4KB nil guard page. The `Settings` struct has grown to the point where field offsets are crossing that line: on a feature branch where the struct is slightly larger, the very same dereference became an **unrecoverable hardware fault** (`unexpected fault address 0x1020` / `fatal error: fault`) that killed the entire test process instead of unwinding into `assert.Panics`.

## The change

Both coinbase candidate builders (`CreateCoinbaseTxCandidate` and `CreateCoinbaseTxCandidateForAddress`) now check for nil settings explicitly and panic with a named message, so the documented contract holds no matter how large `Settings` grows. There is no caller-visible behaviour change for non-nil settings.

Worth knowing beyond this PR: every new settings field pushes more offsets toward that 4KB cliff, so any other code path that relies on an implicit nil-settings panic has the same latent hazard.

## Testing

- `go test ./model/ -run TestMiningCandidate` — all pass, including the `NilSettings` panic assertions.
- The failure mode itself was reproduced (and is fixed) on the larger-struct feature branch where it was discovered.

### DIFF
--- a/model/mining_candidate.go
+++ b/model/mining_candidate.go
@@ -14,6 +14,18 @@ import (
 // p2pk is an optional parameter to specify if the coinbase output should be a pay to public key
 // instead of a pay to public key hash
 func (mc *MiningCandidate) CreateCoinbaseTxCandidate(tSettings *settings.Settings, p2pk ...bool) (*bt.Tx, error) {
+	// Explicit nil-guard preserving the documented (and tested) panic contract.
+	// Relying on the implicit nil-dereference stopped being safe once the
+	// Settings struct grew toward 4KB: Go only converts a nil dereference into a
+	// recoverable panic while the accessed field's offset stays inside the
+	// runtime's nil guard page (~4KB). Beyond it the access becomes an
+	// UNRECOVERABLE fault ("unexpected fault address") that kills the process
+	// instead of unwinding to the caller's recover — observed live once the
+	// struct crossed the line on a feature branch.
+	if tSettings == nil {
+		panic("CreateCoinbaseTxCandidate: nil settings")
+	}
+
 	// Create a new coinbase transaction
 	arbitraryText := tSettings.Coinbase.ArbitraryText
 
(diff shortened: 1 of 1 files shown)


ctnguyen (approved):
Review at commit [0cc521cb87af](https://github.com/bsv-blockchain/teranode/commit/0cc521cb87af)

A small, correct, well-reasoned defensive fix. Safe to merge; the notes below are follow-ups, none blocking.

## Non-blocking issues

### ChiR1 — New guard on `CreateCoinbaseTxCandidateForAddress` is untested
**Problem:** The PR adds a nil-settings guard to *both* builders, but only `CreateCoinbaseTxCandidate` has a `NilSettings` panic assertion (`mining_candidate_test.go:303`). `TestMiningCandidate_CreateCoinbaseTxCandidateForAddress` covers `NilAddress`/`EmptyAddress`/`ZeroCoinbaseValue` but never passes `nil` settings, so the second guard — the exact code this PR is motivated by — has no regression test.
**Why it matters:** The whole point of the change is that the implicit panic silently degraded into a process-killing fault as `Settings` grew. Without a test on the `ForAddress` path, the same regression can recur there undetected on the next struct growth.
**Fix:** Add a sibling subtest, e.g.
```go
t.Run("NilSettings", func(t *testing.T) {
    addr := "1DkmRkb5iQFkDu4NBysog5bugnsyx7kwtn"
    require.Panics(t, func() {
        _, _ = mc.CreateCoinbaseTxCandidateForAddress(nil, &addr)
    })
})
```
(Per project convention, prefer `require` over `assert` in the new test.)

### ChiR2 — nil settings panics while nil address returns an error — inconsistent contract
**Problem:** In `CreateCoinbaseTxCandidateForAddress`, a nil `address` returns `errors.NewConfigurationError(...)`, but a nil `tSettings` now panics — two invalid-argument conditions in the same `(*bt.Tx, error)` function reported through two different mechanisms.
**Why it matters:** Panic-on-nil-arg in a function that otherwise returns errors is surprising for callers and slightly un-idiomatic for Go.
**Fix:** Keep the panic in this PR (it's a deliberately preserved, tested contract). Follow-up recommendation: file an issue to converge these builders on returning a `ConfigurationError` for nil settings — consistent with the nil-address path — and update the tests to assert the error instead of a panic.

## Recap
| ChiR1 | ForAddress guard has no test | | 35% |
| ChiR2 | Panic vs error inconsistency | | 20% |

ordishs (approved):
Verified the premise independently: on main, sizeof(settings.Settings) is 5408 bytes and Coinbase.ArbitraryText sits at offset 4008 — 88 bytes below the 4096-byte nil guard page, so the implicit panic was about to become an unrecoverable fault. Guards are placed before any dereference, no behaviour change for non-nil settings, and the fix transitively covers the nil-settings panic assertion in services/blockassembly/mining/mine_test.go. All TestMiningCandidate tests pass locally.

PR #1288, opened 2026-07-14
fix(legacy/netsync): CVE-2012-2459 dedup floor on the inline below-checkpoint route

freemans13:
## What happened

The explicit duplicate-transaction scan (`CheckSubtreeSlicesForDuplicateTxs`) only ran inside `HandleBlockDirect`'s `preparedSubtreeSlices != nil` block — and `prepareSubtrees` returns non-nil slices **only on the unified route**. On the default inline route the returned slices are nil, so the CVE check never ran on that path at all.

## Why it matters

A CVE-2012-2459 mutation duplicates trailing transactions in a pattern that **preserves the merkle root** (the duplicate-last-node-when-odd rule), so the mutated body hashes to the same trusted block hash and reuses the honest block's proof-of-work — no mining required to trigger it. Below the checkpoint only the header *chain* is trusted; block *bodies* still come from untrusted peers, so an unchecked inline route is a peer-triggerable state-corruption / sync-DoS vector.

`createTxMap` can't catch it either: `txMap.Set` silently overwrites the duplicated txid, while `createSubtrees` replays the block-order `txOrder` (which keeps both copies) and adds the duplicate into the slices twice.

## The change

One unconditional call in `prepareSubtrees`, right after `createSubtrees` and before any UTXO create/spend or subtree write, so **every** route (inline, unified, non-quick) gets the floor from a single place. The existing gated call in `HandleBlockDirect` stays as defensive belt-and-braces on the unified route's returned slices, and its comment now says so. The `CheckSubtreeSlicesForDuplicateTxs` doc comment is enriched to state plainly why it's a mandatory consensus floor rather than an optimization.

## Testing

- New `TestLegacyInline_DuplicateTxid_Rejected`: a `[coinbase, txX, txX]` block on the inline route (unified flag off) is rejected with `BlockInvalidError` before any UTXO commit; without the fix it slips past the guard and reaches `ValidateTransactionsLegacyMode`.
- Existing `unified_dedup` tests stay green (the unified route's guard is unchanged).
- Full `netsync` and `model` suites pass.

## Notes

Independent of the window/flow-control work — this is the security floor on its own, deliberately sliced out first.

### DIFF
--- a/model/check_duplicate_txs.go
+++ b/model/check_duplicate_txs.go
@@ -6,21 +6,39 @@ import (
 	"github.com/bsv-blockchain/teranode/errors"
 )
 
-// CheckSubtreeSlicesForDuplicateTxs is the consensus integrity floor for the
-// CVE-2012-2459 duplicate-transaction check on any code path that holds the
-// block's subtree slices in memory but does not go through Block.Valid (which
-// runs the full pooled/disk-backed checkDuplicateTransactions).
+// CheckSubtreeSlicesForDuplicateTxs is a MANDATORY consensus security floor —
+// not an optimization. It enforces the CVE-2012-2459 duplicate-transaction
+// check on any code path that holds the block's subtree slices in memory but
+// does not go through Block.Valid (which runs the full pooled/disk-backed
+// checkDuplicateTransactions). Every such path MUST call it before creating or
+// spending UTXOs.
+//
+// WHY IT IS REQUIRED (do not remove or gate it on "the block is trusted"):
+// the merkle root CANNOT detect a CVE-2012-2459 duplicate. The merkle tree's
+// duplicate-last-node-when-odd rule means a block whose trailing transactions
+// are duplicated in a specific pattern produces the SAME merkle root — and
+// therefore the same block header and the same block hash — as the honest
+// block. So CheckMerkleRoot passing does NOT imply the transaction list is
+// duplicate-free; this scan is the only thing that catches it.
+//
+// WHY IT STILL APPLIES BELOW THE CHECKPOINT: "below checkpoint" trusts the
+// chain of block HASHES (the header chain is checkpoint-anchored), NOT the
+// block BODIES, which are still downloaded from untrusted peers. A CVE-2012-2459
+// mutation produces a body that hashes to the same, trusted block hash while
+// carrying a repeated transaction — so a below-checkpoint block being "safe"
+// (its hash is on the trusted chain) does not make a peer-supplied body safe.
+// Skipping this on a below-checkpoint fast path is a cheap, peer-triggerable
+// sync-DoS / state-corruption vector, because the mutated body reuses the
+// honest block's proof-of-work (no mining required).
 //
 // It scans every node hash across all subtree slices into a plain Go map,
 // skipping the coinbase placeholder at position [0][0], and returns a
-// BlockInvalidError on the first duplicate. The function is intentionally
-// lightweight — O(N) in the total transaction count with one map allocation —
-// and is suitable for the below-checkpoint unified route where blocks are
-// small enough that the slice-based in-memory scan is the right trade-off.
-//
-// Callers on the full validation path (Block.Valid) continue to use the
-// existing pooled/disk-backed checkDuplicateTransactions; this helper is an
-// independent, non-invasive floor for slice-only paths.
+// BlockInvalidError on the first duplicate. It is intentionally lightweight —
+// O(N) in the total transaction count with one map allocation — so it is cheap
+// enough to run unconditionally on every slice-only path. Callers on the full
+// validation path (Block.Valid) instead use the pooled/disk-backed
+// checkDuplicateTransactions; this helper is the independent floor for the
+// paths that skip it.
 func CheckSubtreeSlicesForDuplicateTxs(slices []*subtreepkg.Subtree) error {
 	// Estimate total node count for the initial map capacity to avoid rehashing.
 	totalNodes := 0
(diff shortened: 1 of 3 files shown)


ctnguyen (approved):
Review at commit [6b5fa4b8e53a](https://github.com/bsv-blockchain/teranode/commit/6b5fa4b8e53a)

Solid, correctly-scoped security fix: it closes a real, peer-triggerable gap by moving the duplicate-tx floor to the one place every route passes through, and backs it with a regression test. No blocking issues found.

## Non-blocking issues

### ChiR1 — Regression test and comments mislabel `[coinbase, txX, txX]` as a merkle-preserving CVE vector
**Problem:** `makeDuplicateTxidBlock` and its comments claim `[coinbase, txX, txX]` "is the CVE-2012-2459 mutation shape ... the merkle root of such a block is preserved by the duplicate-last-when-odd rule, so the merkle check alone would pass." That is false for this shape. Its honest counterpart `[coinbase, txX]` (2 leaves, even) roots to `h(cb‖txX)`; the mutated `[coinbase, txX, txX]` (3 leaves, odd) roots to `h(h(cb‖txX)‖h(txX‖txX))` — a *different* root, so merkle validation would in fact reject it. The genuinely merkle-preserving CVE shape duplicates the trailing tx of an *odd*-length honest list, e.g. honest `[coinbase, txA, txB]` vs mutated `[coinbase, txA, txB, txB]`. Compounding this, the test drives `prepareSubtrees` directly with a zero merkle root in the header and never runs `CheckMerkleRoot`, so it never demonstrates the "merkle passes but dedup catches" property it asserts.
**Why it matters:** The dedup scan is shape-agnostic (it rejects any repeated hash), so the guard is genuinely exercised and the shipped fix is sound — but the test's stated rationale misrepresents the exact threat it is meant to pin down. A maintainer trusting the comment could conclude the merkle-bypass case is covered end-to-end when the test proves nothing about merkle behaviour.
**Fix:** Either drop the merkle-preservation wording and describe the block plainly as "contains a duplicated txid," or — better for a security regression — use a true CVE vector (`[coinbase, txA, txB, txB]`), set the header merkle root to the honest `[coinbase, txA, txB]` root, and assert `CheckMerkleRoot` *passes* while the dedup floor still rejects, so the test demonstrates that merkle validation alone is insufficient.
**Criticality: 45%**

### ChiR2 — Enriched doc comment mandates a floor the quick-validate flow doesn't run
**Problem:** The new doc comment on `CheckSubtreeSlicesForDuplicateTxs` asserts "Every such path MUST call it before creating or spending UTXOs." But the server-side quick-validate flow in `services/blockvalidation/quick_validate.go` creates and spends UTXOs in `createAndSpendUTXOsForBatch` (line 704) and only checks the merkle root at the end via `validateSubtrees → CheckMerkleRoot` (line 747) — it never calls the floor. On the legacy unified route this is compensated by the client-side call in `HandleBlockDirect`, and native catchup relies on the separate "subtree list bound to checkpoint-verified headers" trust argument — but neither is what the universal wording promises.
**Why it matters:** A universal "MUST" invites a future reader to assume the invariant is enforced everywhere, when in fact it is enforced by external callers on some paths and by a different trust model on others. If the catchup trust assumption ever weakens, the gap is silent.
**Fix:** Document the exemption/trust model in the comment (checkpoint-bound subtree hashes for catchup; client-side pre-gate for the unified route) and narrow the "every path" wording. Note that simply calling the floor inside `validateSubtrees` would *not* help: `validateSubtrees` runs at `quick_validate.go:731`, after `createAndSpendUTXOsForBatch` has already committed UTXOs per batch, and the async pipeline populates `block.SubtreeSlices` batch-by-batch interleaved with those writes — there is no cheap whole-block pre-UTXO scan point. Actually enforcing the floor there would require per-batch dedup (a larger change); documenting the exemption is the proportionate fix.
**Criticality: 35%**

### ChiR3 — Unified route now runs the dedup scan twice on the largest blocks
**Problem:** On the unified route `prepareSubtrees` scans `slices` (new call), then returns those same slice objects as `preparedSubtreeSlices`, and `HandleBlockDirect` scans them again (line 232) with no mutation in between. That is a second full O(N) map allocation + scan of identical in-memory data, precisely on the below-checkpoint fast path built for the biggest historical blocks (millions of txs).
**Why it matters:** The second scan adds cost without adding safety — same slices, same process, no intervening write — on exactly the path where N is largest. It is small relative to extend/create/spend, but it is pure waste.
**Fix:** Drop the `CheckSubtreeSlicesForDuplicateTxs` call in `HandleBlockDirect` now that `prepareSubtrees` runs it unconditionally, keeping only `CheckMerkleRoot` there. If the belt-and-braces call is retained deliberately, the comment already says so — but weigh that against the doubled cost on fat blocks.
**Criticality: 28%**

### ChiR4 — Regression test asserts an error but not "before any UTXO commit"
**Problem:** `TestLegacyInline_DuplicateTxid_Rejected` asserts a `BlockInvalidError` is returned, which pins the *fact* of rejection and (via the error type) that it fires before `ValidateTransactionsLegacyMode`. It does not assert the UTXO store is untouched, which is the property the PR body actually advertises ("rejected ... before any UTXO commit").
**Why it matters:** The placement guarantee is the security-relevant part; a test that only checks the error type would still pass if a future refactor moved the scan after a partial `createUtxos`.
**Fix:** After the `require.Error`, assert the store holds no entry for the duplicated txid, e.g. `_, err := store.Get(ctx, tx.TxIDChainHash()); require.Error(t, err)` (or a count/`ErrTxNotFound` check), so the ordering invariant is what the test locks down.
**Criticality: 15%**

## Recap
| ChiR1 | Test mislabels vector as merkle-preserving | | 45% |
| ChiR2 | Doc "MUST" not enforced on quick-validate flow | | 35% |
| ChiR3 | Double dedup scan on unified route | | 28% |
| ChiR4 | Test misses no-UTXO-committed assertion | | 15% |

ordishs (approved):
Verified the gap is real: with the prepareSubtrees check removed, the new inline-route test passes a [coinbase, txX, txX] block through the whole path with no error — createTxMap collapses the duplicate via txMap.Set while createSubtrees replays txOrder with both copies. The fix sits at the right choke point (after createSubtrees, before any UTXO create/spend or subtree write) and covers inline, unified, and non-quick routes from one place. The coinbase-placeholder skip is safe for the threat model since a merkle-preserving duplication can only affect a trailing suffix of leaves, never leaf 0. Full netsync and model suites plus go vet green on this checkout.

PR #1289, opened 2026-07-14
feat(legacy/netsync): default-off fail-closed lever for the below-checkpoint inline path

freemans13:
## What this adds

`blockvalidation_legacy_below_checkpoint_fail_closed` (default **false**). When ON, the non-unified legacy below-checkpoint inline pipeline drops `validator.WithCreateConflicting` on the spend, so a genuine conflict **hard-fails** the block (a non-retryable `ErrSpent`/`ErrTxConflicting` reaching `hardFail`) instead of writing a conflicting subtree node that block assembly must later reconcile.

Below a hard-coded checkpoint a genuine conflict is impossible in a valid chain, so hard-failing is the correct, corruption-surfacing outcome — it makes the inline path behave like the quick-validation path.

## Details

- `legacyFailClosed(height)` gates on **all** of: the flag, `legacyOutpointOnly`, and NOT `legacyUnified` (the unified route is already fail-closed end-to-end, so this inline lever only applies when the unified route is off). Threaded into `PreValidateTransactions` as an explicit bool.
- The `ErrTxConflicting` swallow in `PreValidateTransactions` is now gated on `!failClosed` — both to preserve today's behaviour when the flag is off, and so a future refactor that re-introduces a conflict path cannot silently swallow a genuine below-checkpoint conflict.
- Startup WARN when the flag is set without `blockvalidation_outpoint_only_below_checkpoint` (the lever is a no-op without it).

**With the flag off the inline pipeline is byte-identical to today.**

## Testing

- New `fail_closed_test.go`: flag on ⇒ conflict hard-fails; flag off ⇒ unchanged swallow-and-reconcile.
- Existing netsync suite green.

## Stacking

Stacked on #1288 (the every-route CVE-2012-2459 dedup floor) — both touch the inline `prepareSubtrees` / `PreValidateTransactions` region, so please merge #1288 first. Review the [compare against that branch](https://github.com/bsv-blockchain/teranode/compare/stu/cve-dedup-every-route...stu/legacy-fail-closed-lever) to see only this PR's delta.

Note: the `QuickValidated` marker chain (stamp → carry over gRPC → block-assembly gate) that pairs with this lever on the source branch is deliberately held back as its own later slice — it's a producer→wire→consumer feature that deserves separate review.


ctnguyen (approved):
Review at commit [24e6554a4d6f](https://github.com/bsv-blockchain/teranode/commit/24e6554a4d6f)

A tightly-scoped, default-off lever with a clean gate, honest documentation, and good flag-off parity. It is safe to land as-is; the substantive concerns are about the ON behaviour and test fidelity, not the mechanics of the change.

## Non-blocking issues

### ChiR1 — Fail-closed premise vs. the "stale spending data is expected during catchup" contract
**Problem:** The PR's premise is "below a checkpoint a genuine conflict is impossible in a valid chain." That holds for *chain-level* double-spends, but the swallow it replaces was not written for those. The comment above the swallow says so: *"ErrTxConflicting is expected during legacy catchup when the UTXO store has stale spending data. The block is confirmed, so its transactions take precedence"* (`handle_block.go:1283`). The genuine open question is whether a below-checkpoint outpoint can already carry a **different-spender** record in the store — e.g. left by a rolled-back re-org or a prior non-canonical spend. `store.Spend` returns `ErrSpent` only when the recorded spending data differs from this tx's (`sql.go:2246`; a same-spender re-spend is idempotent success, `sql.go:2249`, and dirty-restart replay is explicitly covered, `quick_validate.go:1292`). So crash-replay is *not* at risk — but a genuine different-spender stale record is exactly what the old swallow tolerated and what fail-closed now turns into a non-retryable, uncommittable block.
**Why it matters:** If such stale different-spender data is reachable below a checkpoint, enabling the flag converts a recoverable, reconcilable condition into a hard block rejection that can stall sync — the opposite of the write-and-reconcile path's resilience. It is default-off and the longdesc already says "leave false until a controlled soak," so not a merge blocker, but it is the headline caveat and the reason the flag must stay experimental.
**Fix:** Before enabling anywhere, establish (and ideally assert in code or a test — see ChiR2) that a below-checkpoint `ErrSpent`/`ErrTxConflicting` on this path can *only* be a genuine chain-level double-spend and never pre-existing different-spender stale data. If that cannot be guaranteed, admit one bounded reload-and-retry before hard-failing to absorb the stale-data case. At minimum, reconcile the new premise with the contradicting comment at `handle_block.go:1283` so a future reader is not misled.
Criticality: 50%

### ChiR2 — Tests never exercise the real validator, and the missing case is the stale different-spender one
**Problem:** `TestLegacyFailClosed_GenuineConflict_HardFails` uses `createConflictingSpyValidator`, which returns `errors.ErrTxConflicting` verbatim, so it proves the *netsync gating* (swallow skipped ⇒ non-retryable ⇒ hardFail) but not that dropping `WithCreateConflicting` through the real `Validator.Validate` yields a non-retryable error — production returns a `ProcessingError`/`ErrUtxoError` wrapping the joined spends (`Validator.go:915`), a different shape than the stub's raw error, and no test drives that branch. `TestLegacyFailClosed_SameBlockParentChain_NoConflictingNodes` calls `PreValidateTransactions` with an always-succeeding spy and never runs `createUtxos`, so it is structurally incapable of surfacing a spurious failure — it only confirms which option was passed. Note that `makeSpendValidator` is *also* a `MockValidator` (it calls `store.Spend` directly, `unified_parity_test.go:156`), so a test built on it still bypasses `saveAsConflicting`/`ProcessingError`; it proves store-spend + netsync gating, not the validator's conflict branch.
**Why it matters:** The PR's central guarantee rests on the *validator's* behaviour, which none of these tests touch; a regression in `Validator.go`'s conflict branch would leave them green. And the one case the old swallow actually handled — pre-existing stale spend data — is untested.
**Fix:** Add the test that directly exercises the ChiR1 concern: seed a real `sqlitememory` store with a prior *different-spender* spend of an outpoint, then process the checkpointed canonical block's spending tx with the flag ON, and assert the intended outcome (a non-retryable `ProcessingError`, not a `BlockInvalidError` — that classification belongs to the dedup path). Scope it honestly: with `makeSpendValidator` this proves the store-spend + netsync hard-fail; to prove the validator's `saveAsConflicting`/`ProcessingError` wrapping as well, stand up an actual validator instance rather than a mock. Keep the spy tests for the gating assertions.
Criticality: 40%

## Recap
| ChiR1 | Premise ignores stale different-spender data | | 50% |
| ChiR2 | Tests never exercise the real validator | | 40% |

ordishs (approved):
Approving. Traced the load-bearing claim through the real validator/store path (the new tests only exercise a spy validator) and it holds:

- With `CreateConflicting` dropped, a spend clash in `Validator.go` skips the `saveAsConflicting` branch and falls through to a `ProcessingError` wrapping the store's `UtxoError` → `UtxoSpentError` (`ERR_UTXO_SPENT`). None of those codes are in `IsRetryableError`'s retryable set, so a genuine conflict lands in `hardFail` on the **first attempt** — no wasted 10×2s retry loop.
- Crash-replay safety: a re-spend by the *same* spender is idempotent in the SQL store (the concurrent-idempotent re-check only flags a *different* spender), so a mid-block crash and replay won't spuriously hard-fail under the new flag.
- Flag-off byte-identity confirmed: `WithCreateConflicting(true)` still appended, `ErrTxConflicting` swallow unchanged. Gating the swallow on `!failClosed` rather than relying on unreachability is good defensive design.

Verified locally: `go test -race -count=1 ./services/legacy/netsync/ ./model/` — both green.

Minor, non-blocking suggestions:

1. The startup WARN covers `fail_closed` without `outpoint_only`, but not `fail_closed` alongside `legacy_unified` — that combination also makes the lever a no-op (bypassed by design). Same "operator A/B-testing gets nothing silently" argument applies; extending the WARN would be cheap.
2. `PreValidateTransactions` now ends in two adjacent bools (`…, outpointOnly, failClosed`) — call sites like `(…, true, false)` are easy to transpose. Worth a small mode struct if a third lever arrives.
3. `require.Contains(err.Error(), "non-retryable")` couples the test to a log string; the `errors.IsRetryableError` assertion next to it is the meaningful one.
4. Test-coverage gap (accepted risk): the conflict ⇒ non-retryable mapping through the real validator+store is established by code reading, not by a test. An integration test in the style of `inline_dedup_test.go` (real sqlitememory store, genuine double-spend, flag on ⇒ error) would pin the contract — if a store ever wraps a spend clash in a retryable `StorageError`, the block would still fail, but only after the full retry loop, and nothing would catch the regression.

To be merged after the stacked base per the PR description.

PR #1294, opened 2026-07-15
fix(legacy): nil-guard UsePrefetchIngestion + BlockRequested regtest-by-value (#1279)

### LINKED ISSUE #1279: legacy netsync: chainParams nil-guard + unify regtest detection (follow-up from #1190)
Follow-up from #1190 (bounded async block prefetch for legacy catchup). Latent / production-safe today; cheap hardening.

### 4. `UsePrefetchIngestion()` dereferences `chainParams` before the budget short-circuit
`services/legacy/netsync/manager.go` — `UsePrefetchIngestion()` passes `sm.chainParams.Net` as a call argument to `UseBlockPrefetchIngestion`, so the pointer is dereferenced before that function's `budgetBytes > 0` short-circuit can guard it. `handleBlockMsg` now calls this for **every** block, so a nil-`chainParams` SyncManager panics the block-handler goroutine. `New()` always sets `chainParams` (and #1190 had to add it to several test fixtures), so it is latent — but a `sm.chainParams != nil` guard (or passing `*chaincfg.Params` and checking inside) closes it.

### 5. Divergent regtest-detection patterns
`services/legacy/netsync/manager.go` — `BlockRequested` uses pointer equality (`sm.chainParams == &chaincfg.RegressionNetParams`) while its sibling `UsePrefetchIngestion` uses value semantics (`.Net != wire.RegTestNet`). They disagree when `chainParams` is a copy of `RegressionNetParams` (as some tests construct). Harmless today (OnBlock doesn't call `BlockRequested` on the synchronous path), but route both through `.Net` so they can't drift.

Parent: #1189

oskarszoon:
Fixes #1279. Follow-up from #1190; parent #1189. Latent/production-safe today; cheap hardening.

## 1. nil-guard `UsePrefetchIngestion` (#4)
It passed `sm.chainParams.Net` as a call argument to `peerpkg.UseBlockPrefetchIngestion`, so the pointer was dereferenced *before* that function's `budgetBytes > 0` short-circuit could guard it. `handleBlockMsg` now calls `UsePrefetchIngestion` for every block, so a nil-`chainParams` SyncManager would panic the block-handler goroutine. Latent (`New()` always sets `chainParams`) but guarded now: `chainParams == nil → false` (fail closed to the synchronous path). The shared `peerpkg.UseBlockPrefetchIngestion` predicate is still called, so it stays in lockstep with the read-loop's `shouldArmProcessingTimer`.

## 2. `BlockRequested` regtest-by-value (#5)
`BlockRequested` used pointer equality (`chainParams == &chaincfg.RegressionNetParams`); its prefetch-path sibling `UsePrefetchIngestion` uses value semantics (`.Net != wire.RegTestNet`). They disagree when `chainParams` is a *copy* of `RegressionNetParams`. Added a nil-safe value-based `isRegtest()` helper (`.Net == wire.RegTestNet`) and routed `BlockRequested` through it so the two prefetch-path siblings cannot drift.

## Scope note — why only `BlockRequested`
The issue frames the divergence generally, and an earlier draft converted **all four** regtest checks (adding `startSync`'s headers-first gate, `isSyncCandidate`, `handleBlockMsg`'s unrequested-block disconnect). That broke `legacy-sync`/`smoketest` in CI, and the reason is instructive:

- Those three sites run on the **synchronous (non-prefetch)** path that regtest always takes (`UsePrefetchIngestion` is false on regtest, so `BlockRequested` isn't even called there).
- The **E2E harness builds `chainParams` as a copy** of `RegressionNetParams` (`util/test/helpers.go`), so pointer-equality evaluates to *not-regtest* in E2E.
- Converting them to value semantics flipped real behavior: `isSyncCandidate` began applying the regtest localhost restriction (rejecting the sync peer → "best height 0"), and `startSync` dropped headers-first.

So those three keep pointer equality — it's load-bearing under the copied-params E2E path. Only `BlockRequested` (the actual prefetch-path sibling named in the issue, and inert on the regtest synchronous path) is unified. A comment on `isRegtest()` records this so the trap isn't re-sprung.

## Tests
- New `TestIsRegtest` (nil / global pointer / value copy / mainnet).
- New `TestUsePrefetchIngestion` subtest: nil `chainParams` with a configured budget → no panic, returns false.
- New `TestBlockRequested` subtest: a copied `RegressionNetParams` admits any block (pins the #5 fix).
- Existing `TestUsePrefetchIngestion` / `TestBlockRequested` matrices unchanged and green.

## Verification
```
go vet ./services/legacy/netsync/...
go test -race ./services/legacy/netsync/   # ok
go test -race ./services/legacy/           # ok
```

### DIFF
--- a/services/legacy/netsync/manager.go
+++ b/services/legacy/netsync/manager.go
@@ -839,6 +839,25 @@ func (sm *SyncManager) IsHeadersFirstMode() bool {
 	return sm.headersFirstMode.Load()
 }
 
+// isRegtest reports whether the active chain params are regression net by
+// network magic rather than pointer identity with chaincfg.RegressionNetParams,
+// so a copied Params value (as some tests construct) is still recognized, and a
+// nil chainParams is safely not-regtest. It exists to give BlockRequested the
+// SAME value semantics as peerpkg.UseBlockPrefetchIngestion (.Net != RegTestNet)
+// so those two prefetch-path siblings cannot drift on a copied-params manager.
+//
+// It deliberately does NOT replace the pointer-equality regtest checks
+// elsewhere in this file (startSync's headers-first gate, isSyncCandidate,
+// handleBlockMsg's unrequested-block disconnect). Those run on the synchronous
+// (non-prefetch) path that regtest always takes, and the E2E harness builds
+// chainParams as a *copy* of RegressionNetParams — so switching them to value
+// semantics flips real behavior (e.g. isSyncCandidate would apply the regtest
+// localhost restriction, and startSync would drop headers-first) and breaks
+// legacy-sync/smoketest. Pointer equality there is load-bearing; leave it.
+func (sm *SyncManager) isRegtest() bool {
+	return sm.chainParams != nil && sm.chainParams.Net == wire.RegTestNet
+}
+
 // isSyncCandidate returns whether or not the peer is a candidate to consider
 // syncing from.
 func (sm *SyncManager) isSyncCandidate(peer *peerpkg.Peer) bool {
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):
Review at commit [52e02e2](https://github.com/oskarszoon/teranode/commit/52e02e23edebccf0a51e310c020eaff1d3801013)

Nice, tight hardening PR. Both fixes are correct, the diff is small and easy to reason about, and the reasoning is unusually well documented — the `isRegtest` godoc that spells out *why the other three pointer-equality checks must stay* is exactly the kind of note that stops someone re-springing the trap in six months. The nil-guard on `UsePrefetchIngestion` fails closed to the synchronous path, which is the right conservative default (the deref happened as a call argument, ahead of `UseBlockPrefetchIngestion`'s budget short-circuit), and routing `BlockRequested` through the value-based check keeps it in lockstep with its prefetch-path sibling. The tests earn their place too: the `TestIsRegtest` matrix, the nil-params fail-closed subtest, and the copied-regtest-params subtest each pin exactly the behavior that changed. Verified the untouched pointer-equality sites (`startSync`, `isSyncCandidate`, `handleBlockMsg`) against the base and the scope call is sound. Nothing blocking, nothing to nit — good to merge.

ordishs (approved):

Issue #1272, opened 2026-07-13
Propagation strips actionable rejection reasons (UTXO_NON_FINAL, TX_LOCK_TIME, …) from client-facing errors


galt-tr:
## Summary

Propagation's client-facing error surface strips the actual rejection reason from validator failures. Submitters see only:

```
PROCESSING (4): [ProcessTransaction][<txid>] failed to validate transaction
```

while the actionable inner error — which teranode itself produces and logs — is discarded before it reaches the client. Broadcast services in front of teranode (arcade/ARC-style) therefore cannot classify rejections, cannot tell a wallet what to fix, and cannot distinguish a *transient* condition (tx not final yet) from a *permanent* one (invalid script, bad fee).

## Where the detail is lost

- `services/propagation/Server.go` `processTransactionInternal` wraps any `validator.Validate` error as `errors.NewProcessingError("[ProcessTransaction][%s] failed to validate transaction", …, err)` — the inner error is attached to the chain.
- HTTP `/tx` and `/txs` responses render via `errors.UserMessage(err)` (`errors/errors.go`), which formats only the **top-level** code + message (`%s (%d): %s`), discarding the wrapped chain.
- The gRPC surface uses `WrapGRPCPublic` / `WrapPublic`, whose `PublicError` likewise keeps only the outermost code + message.
- The full chain survives only in server logs and on the rejected-tx Kafka topics — places a submitter cannot reach.

## Motivating incident (mainnet, 2026-07-11)

A consensus-valid transaction (`04c1bebd30f7242eb2616f7d8607fa8654744ebfa4f59ba0a2c1c96d4877fe0e`, later mined in block 957,459) was submitted while its timestamp nLockTime was still ahead of the chain's median-time-past. Teranode correctly rejected it as non-final, and internally had a perfect description of why:

```
UTXO_NON_FINAL (71): [Validate][04c1bebd…fe0e] transaction is not final
  -> TX_LOCK_TIME (35): lock time (1783806110) as timestamp is not less than median block time (1783805456)
```

[...]

PR #1295, opened 2026-07-16
MvP-4688 Surface actionable rejection reasons on the public tx error surfaces

ctnguyen:
Resolves https://github.com/bitcoin-sv/teranode/issues/4688
(public report: https://github.com/bsv-blockchain/teranode/issues/1272)

## Problem

When a submitted transaction is rejected, all four public surfaces (HTTP `/tx`, HTTP
`/txs`, gRPC single, gRPC batch) return only the generic outermost error —
`PROCESSING (4): ... failed to validate transaction` — while the actionable verdict
Teranode already computed (e.g. `UTXO_NON_FINAL (71) -> TX_LOCK_TIME (35): lock time ...
median block time ...`) is stripped by the public error boundary. Clients cannot tell a
transient rejection (not final yet) from a permanent one (invalid script, bad fee). Two
status-class bugs compound it: `/txs` always returns HTTP 500, and gRPC maps every tx
rejection to `codes.Internal` — both mis-signal "server fault" for client-side rejections.

## Fix

- `errors`: an allowlist of client-safe verdict codes (non-final/lock-time/policy/invalid
  plus the conflict family) and a `DeepestPublicCause` helper; `UserMessage` and
  `PublicError` now surface the innermost allowlisted cause's `code`+`message`. Everything
  off the allowlist keeps the current generic collapse; file/line, `data`, and the wrapped
  chain are never emitted.
- `errors`: `ErrorCodeToGRPCCode` maps tx-content rejections to `codes.InvalidArgument`
  and chain-state conflicts (double-spend/conflicting/spent/locked) to
  `codes.FailedPrecondition`; everything else still `codes.Internal`.
- `propagation`: `/txs` derives its aggregate status from the per-tx errors (first
  client-error status wins; genuine system/storage errors still dominate to 500; OK-class
  outcomes such as already-known txs no longer force 500).
- Tests: end-to-end coverage of all four public surfaces, asserting the allowlisted reason
  is carried verbatim and no internal metadata leaks; unit coverage for the allowlist
  walk, the collapse of non-allowlisted chains, and the gRPC code mapping.
- Docs: a short validator note recording that Teranode is mempool-less and terminally
  rejects non-final transactions by design (unlike svnode's `CTimeLockedMempool`
  retention); clients resubmit once the tip's median-time-past passes the locktime.

## Non-goals

No consensus/validation change (the finality rule is at parity with svnode), no non-final
retention/mempool, no bdk changes. Structured per-tx response fields and bdk policy-code
names in messages are follow-ups.

### DIFF
--- a/docs/topics/services/validator.md
+++ b/docs/topics/services/validator.md
@@ -246,6 +246,8 @@ The above represents an implementation of the core Teranode validation rules:
 
         - Note: This means that Teranode will deem non-final transactions invalid and REJECT these transactions. It is up to the user to create proper non-final transactions to ensure that Teranode is aware of them. For clarity, if a transaction has a locktime in the future, the Tx Validator will reject it.
 
+        - Design note (mempool-less rejection): Teranode has no mempool — neither a main mempool nor a non-final holding pool — so a non-final transaction is **terminally rejected** rather than retained. This differs from svnode, whose `CTimeLockedMempool` holds a non-final transaction and re-enqueues it once it becomes final. In Teranode this is expected behaviour, not a defect: the client is responsible for resubmitting once the chain tip's median-time-past (BIP113) passes the transaction's locktime. The rejection now carries the actionable reason (`UTXO_NON_FINAL` → `TX_LOCK_TIME`, including the locktime and the compared median block time) through to the public HTTP and gRPC surfaces, so the submitter can see why the transaction was rejected and when it is worth resubmitting.
+
     - No output must be Pay-to-Script-Hash (P2SH)
 
     - A new transaction must not have any output which includes P2SH, as creation of new P2SH transactions is not allowed.
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -700,6 +700,14 @@ func WrapGRPC(err error) error {
 // code+message may be surfaced through the public error boundary. Each names a
 // verdict about the submitted transaction, carries no node-internal state, and
 // is actionable by the submitter; everything else collapses to the outermost code.
+//
+// This allowlist is intentionally narrower than the set of codes an HTTP handler
+// may classify as client errors: some codes earn a specific 4xx status yet are
+// deliberately absent here and from ErrorCodeToGRPCCode. An HTTP status only sorts
+// the request into a coarse client-error category and exposes neither the detailed
+// ERR code nor its message; surfacing a message is held to the stricter bar of a
+// client-safe, actionable verdict. Widening this set therefore needs the same
+// safety review — it is not kept in sync with the HTTP classifier by design.
 var publicCauseCodes = map[ERR]struct{}{
 	ERR_UTXO_NON_FINAL:          {},
 	ERR_TX_LOCK_TIME:            {},
@@ -722,6 +730,13 @@ func isPublicCause(code ERR) bool {
 // is present. The walk is bounded like (*Error).Is to stay safe on pathological
 // chains. Only code+message are meaningful on the returned error; callers must
 // never surface its file/line/function, data, or the rest of the chain.
+//
+// This helper backs the shared public error boundary — UserMessage, PublicError,
+// and thus WrapGRPCPublic — consumed by all external surfaces (propagation /tx and
+// /txs, the asset HTTP error boundary, p2p), so preferring the deepest allowlisted
+// cause is a global boundary behaviour, not scoped to a single service. It is safe
+// by construction: only codes on publicCauseCodes are ever surfaced, and any other
+// cause collapses to the outermost code.
 func DeepestPublicCause(err error) *Error {
 	if err == nil {
 		return nil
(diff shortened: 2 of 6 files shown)


oskarszoon (approved):
Reviewed across security (info-leak/allowlist), Go-correctness, and blast-radius lenses, each finding verified against the code rather than the PR narrative. The mechanism is sound: `DeepestPublicCause` picks the innermost allowlisted verdict and `PublicError`/`UserMessage` emit only its code+message — I confirmed they build a fresh `Error{code,message}` with no wrapped chain, no `data`, and no file/line, so internal detail can't cross the boundary and off-allowlist errors still collapse to the generic code. The `/txs` aggregate status (system-fault→500 dominates, else first client-error wins, OK-class skipped) is implemented correctly, the gRPC re-mapping is more correct than the old blanket `Internal`, and the docs note matches behavior. No info-leak, no consensus change, no functional regression. Approving.

The stale-test CI failure I'd flagged is already fixed in `df84fff3` — the batch validation case now expects the derived 400, and the added assertion that the aggregated body carries `ERR_TX_INVALID (35): ...` is a nice touch (it pins the actual point of the PR: the reason survives the boundary instead of collapsing). `test` job is green; the remaining SonarQube "fail" is an analysis-didn't-complete infra issue, not a code finding.

Two nits, non-blocking:

- **gRPC mapping is global, comment reads as scoped.** `ErrorCodeToGRPCCode` is consumed by the generic `WrapGRPC` (outermost-code based), so the 8 tx-verdict mappings flip the gRPC status for every `WrapGRPC` caller — including the validator's own `ValidateTransaction`/`Batch` endpoints — not just the 4 named public surfaces. Verified benign (no caller branches on `codes.Internal`, retry gate keys only on Unavailable/DeadlineExceeded, app control-flow keys on the reconstructed ERR code) and semantically more correct. Worth: clarify the comment at `errors.go:954-955` (it implies public-cause scope; the mapping is global), and consider one unit assertion on the validator's own gRPC status so the widened surface is pinned.
- **Latent status/message divergence.** The message is chosen by `DeepestPublicCause` (innermost) and the HTTP status by `httpStatusForTxError` (`errors.Is` priority walk, 409-conflict before 400-invalid). No live divergence — the validator never nests a conflict-family code around an invalid-family one, so they always agree today — but nothing enforces the invariant. A defensive test asserting family agreement would close the footgun. gRPC is immune (both derived from the same `PublicError`).

ordishs:
Heads-up: this branch breaks an existing test that wasn't updated with the new `/txs` status aggregation.

`TestHandleMultipleTx/Transaction_validation_error` (`services/propagation/http_handlers_test.go`, ~line 420) injects `NewTxInvalidError` into a `/txs` batch and still expects `http.StatusInternalServerError`, but the new per-tx aggregation now (correctly) classifies it as a client error:

```
--- FAIL: TestHandleMultipleTx (0.00s)
    --- FAIL: TestHandleMultipleTx/Transaction_validation_error (0.00s)
        http_handlers_test.go:460:
            Error:      Not equal:
                        expected: 500
                        actual  : 400
FAIL	github.com/bsv-blockchain/teranode/services/propagation
```

`Test_handleMultipleTx_ErrorOrderPreserved` in `Server_test.go` was updated for this behaviour change, but this one was missed. Since 400 is the intended new behaviour, the fix is to update the expectation to `http.StatusBadRequest` (and the test name/comment if desired).

ctnguyen:
> Heads-up: this branch breaks an existing test that wasn't updated with the new `/txs` status aggregation.
> 
> `TestHandleMultipleTx/Transaction_validation_error` (`services/propagation/http_handlers_test.go`, ~line 420) injects `NewTxInvalidError` into a `/txs` batch and still expects `http.StatusInternalServerError`, but the new per-tx aggregation now (correctly) classifies it as a client error:
> 
> ```
> --- FAIL: TestHandleMultipleTx (0.00s)
>     --- FAIL: TestHandleMultipleTx/Transaction_validation_error (0.00s)
>         http_handlers_test.go:460:
>             Error:      Not equal:
>                         expected: 500
>                         actual  : 400
> FAIL	github.com/bsv-blockchain/teranode/services/propagation
> ```
> 
> `Test_handleMultipleTx_ErrorOrderPreserved` in `Server_test.go` was updated for this behaviour change, but this one was missed. Since 400 is the intended new behaviour, the fix is to update the expectation to `http.StatusBadRequest` (and the test name/comment if desired).

Thanks @ordishs, I've  pushed the commit to fix the broken test. The fail was at my first push, it's all green now

ordishs (approved):
The broken test is fixed in df84fff30 and verified locally — `TestHandleMultipleTx`, the full `services/propagation` and `errors` packages all pass, and CI is green across the board.

The change itself is solid: the `DeepestPublicCause` walk is bounded and handles foreign wrappers correctly, the `/txs` aggregation precedence (server fault dominates, first client error wins, success-class outcomes skipped) is right, the gRPC code-class mapping doesn't disturb internal clients (they recover application codes via `UnwrapGRPC`, and the retry helper only retries Unavailable/DeadlineExceeded), and the tests assert no file/line/data leakage on every surface.

Two non-blocking follow-up candidates: (1) `httpStatusForTxError` classifies a few codes as client errors (`ErrFrozen`, `ErrTxMissingParent`, `ErrTxCoinbaseImmature`, `ErrUtxoInvalidSize`) that aren't on the `publicCauseCodes` allowlist or in the gRPC mapping, so those surfaces are inconsistent for them; a comment on the allowlist noting whether that's deliberate would prevent drift. (2) `UserMessage` is also used by the asset service error boundary, so the deepest-cause preference applies there too — likely an improvement, just wider than the four surfaces named in the description.

ctnguyen:
Thanks @ordishs and @oskarszoon for the review, I fixed some non blocking issues

| Point | Reviewer | Verdict | Rationale (one line) |
|-------|----------|---------|----------------------|
| O1(a) clarify `ErrorCodeToGRPCCode` is global | oskarszoon | IMPLEMENT NOW | Comment-only; current wording reads public-scoped though `WrapGRPC` uses it globally. |
| O1(b) pin generic `WrapGRPC` status for verdict codes | oskarszoon | IMPLEMENT NOW | Additive test of the generic `WrapGRPC` path (single-tx `ValidateTransaction`); not batch, which uses `errors.Wrap` + nil top-level status. |
| O2 defensive HTTP/gRPC family-agreement test | oskarszoon | DEFER | A per-code table doesn't exercise `DeepestPublicCause`/nested chains; the real risk (mixed-family nesting) doesn't occur live, so no grounded expectation to assert on an approved PR. |
| R1 comment on allowlist exclusion being deliberate | ordishs | IMPLEMENT NOW | Comment-only; documents the intentional HTTP-4xx-but-not-allowlisted asymmetry to prevent drift. |
| R2 scope comment on shared deepest-cause boundary | ordishs | IMPLEMENT NOW | Comment-only on `DeepestPublicCause` (bot anchor errors.go:725); records the global boundary reach. |

PR #1296, opened 2026-07-16
docs(legacy): correct async bad-block tail-stop comment to best-effort (#1280)

### LINKED ISSUE #1280: legacy prefetch: async bad-block tail-stop is best-effort, not a barrier (follow-up from #1190)
Follow-up from #1190 (bounded async block prefetch for legacy catchup). Low severity; inherent-ish to the async design. Noted by review rather than pushed.

### 3. `!bmsg.peer.Connected()` tail-stop is best-effort, not a barrier
`services/legacy/netsync/manager.go` — `blockHandler` hands bad block N to the async `awaitBlockResult`, then dequeues N+1 and checks `Connected()` *before* `disconnectMisbehaving` has flipped the peer's atomic. So N+1 (and anything dequeued in that window) is still fully validated. A peer feeding expensive invalid blocks can burn CPU on more than one block per disconnect — weaker than the pre-prefetch "one bad block in flight" guarantee.

Inherent to decoupling download from processing, and the peer is being dropped anyway, so this is a bounded cost. Options if we want to tighten: a true barrier (mark the peer/association un-processable synchronously at disconnect and skip all its queued blocks), or accept + document the bound.

Parent: #1189

oskarszoon:
Fixes #1280. Follow-up from #1190; parent #1189. Low severity; comment-only, no behavior change.

## What
The `handleBlockMsg` `!bmsg.peer.Connected()` tail-stop guard's comment claimed it *"deterministically stops the tail, bounding wasted validation to the block already in flight."* That overstates the guarantee.

`awaitBlockResult` disconnects the source peer in its own goroutine, and only after it receives block N's failure reply. So in the window between N failing and the connected flag flipping, the single FIFO consumer (`blockHandler`'s inner goroutine) can dequeue and fully validate N+1, N+2, … . The guard is a **best-effort** tail-stop, not a barrier.

## Resolution
Per the issue's own options (tighten with a true barrier, or accept + document), this takes **accept + document**: the cost is bounded and self-limiting — the peer is being dropped regardless, the async-disconnect window is short, and total in-flight bytes are already capped by the prefetch budget. Wasted validation is O(blocks dequeued in the window), not unbounded.

The comment now describes this honestly and records that a true synchronous barrier (mark the peer un-processable on failure, in the single FIFO consumer, and skip its whole queued tail) was considered and deliberately not taken — not worth the hot-path state and teardown lifecycle for a bounded, low-severity cost inherent to decoupling download from processing.

## Verification
Comment-only change; no logic touched.
```
go vet ./services/legacy/netsync/...   # clean
go build ./services/legacy/netsync/... # ok
```

### DIFF
--- a/services/legacy/netsync/manager.go
+++ b/services/legacy/netsync/manager.go
@@ -1436,14 +1436,17 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockQueueMsg) error {
 	// from processing.
 	//
 	// We test bmsg.peer — the exact peer OnBlock queued and awaitBlockResult
-	// disconnects (sp.Peer) — NOT the resolved primary: for a BlockPriority stream
-	// the failure disconnects only the stream sub-peer while its association
-	// primary stays connected, so the resolved `peer` would miss precisely the
-	// streaming case this guards. The ServiceError is benign to
-	// shouldDisconnectOnBlockErr, so it only makes awaitBlockResult release budget
-	// and log — no second disconnect. Gated on UsePrefetchIngestion so the
-	// regtest/synchronous path, where block-acceptance tooling feeds blocks in
-	// ways this must not disturb, is completely untouched.
+	// tears down (sp.Peer) — NOT the resolved primary. On a bad block
+	// awaitBlockResult calls disconnectMisbehaving, which drops the WHOLE
+	// association (the primary first, then the stream sub-peer), so either flag
+	// would flip for the misbehaviour case; but bmsg.peer is the peer that queued
+	// this tail, and it also drops on sub-peer-scoped teardowns (TCP loss,
+	// RemoveStream) that a primary check would not reflect — so it is the tighter
+	// guard. The ServiceError is benign to shouldDisconnectOnBlockErr, so it only
+	// makes awaitBlockResult release budget and log — no second disconnect. Gated
+	// on UsePrefetchIngestion so the regtest/synchronous path, where
+	// block-acceptance tooling feeds blocks in ways this must not disturb, is
+	// completely untouched.
 	if sm.UsePrefetchIngestion() && !bmsg.peer.Connected() {
 		sm.logger.Debugf("[handleBlockMsg][%s] skipping block from disconnected peer %s", bmsg.blockHash, bmsg.peer)
 		return errors.NewServiceError("[handleBlockMsg] skipping block %s from disconnected peer %s", bmsg.blockHash, bmsg.peer)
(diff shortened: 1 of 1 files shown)


ctnguyen (changes requested):
Review at commit [c803b7d](https://github.com/oskarszoon/teranode/commit/c803b7dbd5d77cb51a8f4b07156234b91f0ec6c4)

## Blocking issue

### ChiR1 — Retained "primary stays connected" rationale is factually wrong
**Problem:** The comment (reflowed but carried over verbatim) still justifies testing `bmsg.peer` instead of the resolved primary with: "for a BlockPriority stream the failure disconnects only the stream sub-peer while its association primary stays connected, so the resolved `peer` would miss precisely the streaming case this guards" (manager.go:1438–1442). That is not what the code does. On failure, `awaitBlockResult` calls `disconnectMisbehaving(sp, ...)` (peer_server.go:1345), which resolves the sub-peer to its association primary via `misbehaviorDisconnectTarget` and disconnects the **primary first**, then the sub-peer (peer_server.go:1242–1249). #1190 introduced `disconnectMisbehaving` specifically *because* "a bare sp disconnect misses the primary" — it replaced the sub-peer-only disconnect that this comment still describes. So in the misbehaviour path both flags flip, and testing the resolved primary would *not* miss the streaming case.

**Why it matters:** The sole purpose of this PR is to make this comment describe reality; it fixes the deterministic→best-effort overstatement but re-commits a second inaccuracy in the same hunk, and that sentence now directly contradicts `disconnectMisbehaving`'s own doc comment in `peer_server.go`. Shipping a comment-only "accuracy" fix that still asserts the primary survives a bad-block disconnect — the opposite of what the code guarantees — defeats the change's purpose and leaves a self-contradiction for the next reader to trip over. It also undersells `bmsg.peer`'s real merit: it's the peer that actually queued the tail, so it also catches sub-peer-scoped teardowns (TCP drop, `RemoveStream`) that a primary check wouldn't — the disconnect symmetry, not asymmetry, is beside the point.

**Fix:** Drop the "primary stays connected" claim and state the actual reason. For example:

```go
// We test bmsg.peer — the exact peer OnBlock queued and awaitBlockResult
// tears down (sp.Peer) — NOT the resolved primary. On a bad block
// awaitBlockResult calls disconnectMisbehaving, which drops the whole
// association (primary + stream sub-peer), so either flag would flip here;
// but bmsg.peer is the peer that queued this tail and it also drops on
// sub-peer-scoped teardowns (TCP loss, RemoveStream) that the primary
// wouldn't reflect, so it is the tighter guard.
```

oskarszoon:
Fixed in 4d1092417.

You're right — `awaitBlockResult` calls `disconnectMisbehaving`, which drops the whole association (primary first, then the sub-peer), so the "primary stays connected" rationale was wrong and contradicted `disconnectMisbehaving`'s own doc. Rewrote it to state the actual reason for testing `bmsg.peer`: either flag flips for the misbehaviour case, but `bmsg.peer` is the peer that queued this tail and also drops on sub-peer-scoped teardowns (TCP loss, `RemoveStream`) that a primary check wouldn't reflect — the tighter guard.

ctnguyen (approved):

icellan (approved):

PR #1297, opened 2026-07-16
p2p: enforce DataHub URL blacklist on block and node_status announcements

gokutheengineer:
**Closes bitcoin-sv/teranode#4684.**

### Problem
`handleSubtreeTopic` rejects announcements whose DataHub URL matches the configured blacklist, but `handleBlockTopic` only runs the SSRF check - a blacklisted host is blocked for subtrees yet accepted for blocks, and the resulting catchup fetches blocks from it. `handleNodeStatusTopic` has the same gap: it stores the announced `BaseURL` in the peer registry as the peer's DataHub URL (used by sync/catchup) without a blacklist check.

Further holes surfaced during review:
- Host comparison did not canonicalize: rooted FQDNs (`http://evil.example./`, one or more trailing dots) bypassed the blacklist, and the SSRF localhost guard was case-sensitive and dot-sensitive (`http://LOCALHOST/`, `http://localhost./`, `http://127.0.0.1.`).
- Scheme-less blacklist entries (`evil.example`) parsed as a URL path, extracted no host, and silently never matched a real full-URL announcement - the entry was dead with no warning.
- Gossip-time checks cannot evict a URL stored before its host was blacklisted: the registry ignores empty-URL updates and an active peer never expires, so an operator blacklisting a live peer's host had no effect on catchup.

### Fix
- Factored the URL trust checks into one helper, `checkDataHubURL` (SSRF validation, which counts as a protocol violation on failure, followed by the operator blacklist, which drops the message without penalising the peer), applied in `handleBlockTopic` and `handleSubtreeTopic`. Blacklisted announcements are dropped before WebSocket forwarding, peer registration, and Kafka publish.
- `handleNodeStatusTopic` runs the same two checks inline: node_status is telemetry, so a blacklisted `BaseURL` is stripped from the message (not stored, not forwarded) while the rest of the peer's status still reaches the registry and dashboard.
- `extractHost` and `validateDataHubURL` canonicalize the hostname (lowercase, trim all trailing dots) before comparing, closing the rooted-FQDN and case bypasses.
- Scheme-less blacklist entries are retried in protocol-relative form (`blacklistEntryHost`), so `evil.example` and `evil.example:9000` now match full-URL announcements; entries with no parseable host at all trigger a startup warning in `NewServer` instead of failing silently (exact-match fallback retained).
- The blacklist is also enforced at the point of use, covering URLs stored before the blacklist entry existed: `GetPeersForCatchup` skips peers whose stored DataHubURL is blacklisted, and `PeerSelector.isEligible` refuses them for sync-peer selection. `isBlacklistedBaseURL` was made package-level (`isBaseURLBlacklisted`) so the selector shares the exact same matching logic.
- The caught-up determination agrees with selection about the blacklist: `SyncCoordinator.isViableSyncCandidate` (a method layering the blacklist over the unconditional viability filters) is now used by `isCaughtUp`, `filterEligiblePeersWithTip`, `checkAllPeersAttempted`, and advertised-probe eligibility. Without this, a blacklisted-but-ahead peer kept the node reporting not-caught-up forever while `SelectSyncPeer` could never pick it.

### Tests
Added to `services/p2p`:
- `TestIsBaseURLBlacklisted` - direct matcher table: scheme-less entries (with and without port) against full-URL announcements, single and double trailing dots, host case, unparseable-entry exact-match fallback.
- `TestHandleBlockTopic_RejectsBlacklistedDataHubURL` - blacklisted host (host-level match with different port/path, and the trailing-dot form) produces no notification, no registry entry, no Kafka publish, and no ban score.
- `TestHandleSubtreeTopic_RejectsBlacklistedDataHubURL` - regression guard for the pre-existing subtree behaviour (exact and trailing-dot forms) through the new helper.
- `TestHandleNodeStatusTopic_StripsBlacklistedBaseURL` - blacklisted BaseURL (trailing-dot form) is not stored and not forwarded, while the telemetry itself still reaches the registry and WebSocket clients.
- `TestBlacklistedDataHubURL_StoredBeforeBlacklist_NotUsedForCatchup` - a URL registered before the blacklist entry survives the node_status strip (documented limitation) but `GetPeersForCatchup` no longer surfaces the peer.
- `TestPeerSelector_SelectSyncPeer_SkipsBlacklistedDataHubURL` - a blacklisted peer loses selection to a viable peer and is not selected even as the only candidate.
- `TestSyncCoordinator_IsCaughtUp_BlacklistedPeerIsCaughtUp` - an ahead peer keeps the node not-caught-up while its host is allowed (control), and stops doing so once the host is blacklisted, keeping the caught-up view consistent with selection.
- `validateDataHubURL` table cases for trailing-dot, double-trailing-dot, and uppercase localhost/loopback forms.

Ran:
```
```
All green.

### DIFF
--- a/services/p2p/Server.go
+++ b/services/p2p/Server.go
@@ -990,13 +990,11 @@ func (s *Server) handleNodeStatusTopic(_ context.Context, m []byte, peerID strin
 		return
 	}
 
-	// Validate BaseURL to prevent SSRF attacks
-	if nodeStatusMessage.BaseURL != "" {
-		if err := s.validateDataHubURL(nodeStatusMessage.BaseURL); err != nil {
-			s.logger.Errorf("[handleNodeStatusTopic] invalid BaseURL from peer %s: %v", peerID, err)
-			s.applyBanScore(peerID, ReasonProtocolViolation)
-			return
-		}
+	// Validate BaseURL (SSRF + operator blacklist) — it is stored in the peer
+	// registry as the peer's DataHub URL, so it must pass the same trust checks
+	// as block/subtree announcements.
+	if nodeStatusMessage.BaseURL != "" && !s.checkDataHubURL(nodeStatusMessage.BaseURL, peerID, "handleNodeStatusTopic") {
+		return
 	}
 
 	if !isSelf && nodeStatusMessage.BestHeight > 0 && nodeStatusMessage.PeerID != "" {
--- a/services/p2p/Server.go
+++ b/services/p2p/Server.go
@@ -998,9 +998,12 @@ func (s *Server) handleNodeStatusTopic(_ context.Context, m []byte, peerID strin
 			return
 		}
 
-		// A blacklisted BaseURL must never reach the peer registry (catchup
-		// would fetch from it), but node_status is telemetry: keep the message
-		// with the URL removed instead of hiding the peer from monitoring.
+		// A blacklisted BaseURL must not reach the peer registry, but
+		// node_status is telemetry: keep the message with the URL removed
+		// instead of hiding the peer from monitoring. This only stops fresh
+		// registrations; a URL stored before its host was blacklisted stays in
+		// the registry and is filtered at the point of use instead
+		// (GetPeersForCatchup and PeerSelector.isEligible).
 		if s.isBlacklistedBaseURL(nodeStatusMessage.BaseURL) {
 			s.logger.Warnf("[handleNodeStatusTopic] removed blacklisted BaseURL %s from node_status of peer %s", nodeStatusMessage.BaseURL, peerID)
 			nodeStatusMessage.BaseURL = ""
--- a/services/p2p/Server.go
+++ b/services/p2p/Server.go
@@ -990,11 +990,21 @@ func (s *Server) handleNodeStatusTopic(_ context.Context, m []byte, peerID strin
 		return
 	}
 
-	// Validate BaseURL (SSRF + operator blacklist) — it is stored in the peer
-	// registry as the peer's DataHub URL, so it must pass the same trust checks
-	// as block/subtree announcements.
-	if nodeStatusMessage.BaseURL != "" && !s.checkDataHubURL(nodeStatusMessage.BaseURL, peerID, "handleNodeStatusTopic") {
-		return
+	if nodeStatusMessage.BaseURL != "" {
+		// Validate BaseURL to prevent SSRF attacks
+		if err := s.validateDataHubURL(nodeStatusMessage.BaseURL); err != nil {
+			s.logger.Errorf("[handleNodeStatusTopic] invalid BaseURL from peer %s: %v", peerID, err)
+			s.applyBanScore(peerID, ReasonProtocolViolation)
+			return
+		}
+
+		// A blacklisted BaseURL must never reach the peer registry (catchup
+		// would fetch from it), but node_status is telemetry: keep the message
+		// with the URL removed instead of hiding the peer from monitoring.
+		if s.isBlacklistedBaseURL(nodeStatusMessage.BaseURL) {
+			s.logger.Warnf("[handleNodeStatusTopic] removed blacklisted BaseURL %s from node_status of peer %s", nodeStatusMessage.BaseURL, peerID)
+			nodeStatusMessage.BaseURL = ""
+		}
 	}
 
 	if !isSelf && nodeStatusMessage.BestHeight > 0 && nodeStatusMessage.PeerID != "" {
--- a/services/p2p/Server.go
+++ b/services/p2p/Server.go
@@ -240,6 +240,15 @@ func NewServer(
 		return nil, errors.NewConfigurationError("listen_mode must be one of '%s', '%s', or '%s' (got '%s')", settings.ListenModeFull, settings.ListenModeListenOnly, settings.ListenModeSilent, listenMode)
 	}
 
+	// Surface blacklist entries with no parseable host: they can only ever
+	// match an announcement byte-for-byte, so the operator almost certainly
+	// misconfigured them. Warn loudly instead of leaving the entry silently inert.
+	for blocked := range tSettings.SubtreeValidation.BlacklistedBaseURLs {
+		if blacklistEntryHost(blocked) == "" {
+			logger.Warnf("[P2P] blacklisted base URL %q has no parseable host and will only match announcements exactly equal to it", blocked)
+		}
+	}
+
 	banlist, banChan, err := GetBanList(ctx, logger, tSettings)
 	if err != nil {
 		return nil, errors.NewServiceError("error getting banlist", err)
--- a/services/p2p/handle_catchup_metrics.go
+++ b/services/p2p/handle_catchup_metrics.go
@@ -215,6 +215,14 @@ func (s *Server) GetPeersForCatchup(ctx context.Context, _ *p2p_api.GetPeersForC
 			continue
 		}
 
+		// Enforce the operator blacklist at the point of use: a URL stored in
+		// the registry before its host was blacklisted must not be handed to
+		// catchup (gossip-time checks cannot evict already-stored URLs).
+		if s.isBlacklistedBaseURL(p.DataHubURL) {
+			s.logger.Debugf("[GetPeersForCatchup] skipping peer %s with blacklisted DataHubURL %s", p.ID, p.DataHubURL)
+			continue
+		}
+
 		totalAttempts := p.InteractionSuccesses + p.InteractionFailures
 
 		blockHashStr := ""
--- a/services/p2p/peer_selector.go
+++ b/services/p2p/peer_selector.go
@@ -196,6 +196,14 @@ func (ps *PeerSelector) isEligible(p *blockchain.PeerInfo, criteria SelectionCri
 		return false
 	}
 
+	// Enforce the operator blacklist at the point of use: a URL stored in the
+	// registry before its host was blacklisted must never be selected for
+	// sync/catchup (gossip-time checks cannot evict already-stored URLs).
+	if ps.settings != nil && isBaseURLBlacklisted(p.DataHubURL, ps.settings.SubtreeValidation.BlacklistedBaseURLs) {
+		ps.logger.Debugf("[PeerSelector] Peer %s has blacklisted DataHub URL %s", p.ID, p.DataHubURL)
+		return false
+	}
+
 	// Check reputation threshold - peers with very low reputation should not be selected
 	if p.ReputationScore < 20.0 {
 		ps.logger.Debugf("[PeerSelector] Peer %s has very low reputation %.2f (below threshold 20.0)", p.ID, p.ReputationScore)
--- a/services/p2p/peer_selector.go
+++ b/services/p2p/peer_selector.go
@@ -46,7 +46,11 @@ func NewPeerSelector(logger ulogger.Logger, settings *settings.Settings) *PeerSe
 // This is a pure function - no side effects, no network calls.
 // Peer IDs are canonical libp2p ID strings.
 func (ps *PeerSelector) SelectSyncPeer(peers []*blockchain.PeerInfo, criteria SelectionCriteria) string {
-	// Handle forced peer - always select it if it exists, regardless of eligibility
+	// Handle forced peer - always select it if it exists, regardless of
+	// eligibility. This deliberately skips the blacklist check in isEligible
+	// too, but a blacklisted DataHub URL is still refused downstream:
+	// sendSyncTriggerToKafka drops the trigger rather than handing the URL to
+	// block validation, so forcing a peer cannot override the blacklist.
 	if criteria.ForcedPeerID != "" {
 		for _, p := range peers {
 			if p.ID == criteria.ForcedPeerID {
--- a/services/p2p/peer_selector_test.go
+++ b/services/p2p/peer_selector_test.go
@@ -280,3 +280,28 @@ func TestPeerSelector_SelectSyncPeer_UnprovenProbeBudgetBoundsHeaderOnlyPeers(t 
 	boundedBudget.UnprovenProbeBudgetRemaining = 3
 	require.NotEmpty(t, ps.SelectSyncPeer(peers, boundedBudget))
 }
+
+// TestPeerSelector_SelectSyncPeer_SkipsBlacklistedDataHubURL: the operator
+// blacklist is enforced at sync-peer selection so a DataHub URL stored in the
+// registry before its host was blacklisted can never be chosen for catchup.
+func TestPeerSelector_SelectSyncPeer_SkipsBlacklistedDataHubURL(t *testing.T) {
+	ps := NewPeerSelector(ulogger.TestLogger{}, &settings.Settings{
+		P2P: settings.P2PSettings{
+			AllowPrunedNodeFallback:     true,
+			FullDeliveryFreshnessWindow: 24 * time.Hour,
+		},
+		SubtreeValidation: settings.SubtreeValidationSettings{
+			BlacklistedBaseURLs: map[string]struct{}{"http://evil.example": {}},
+		},
+	})
+
+	good := newPeer("good", 100, "full", 60, 0)
+	bad := newPeer("bad", 200, "full", 90, 0)
+	bad.DataHubURL = "http://evil.example:8080/api" // same host as blacklist entry
+
+	got := ps.SelectSyncPeer([]*blockchain.PeerInfo{bad, good}, advertisedProbeCriteria(50))
+	require.Equal(t, "good", got, "peer with blacklisted DataHub URL must not win selection")
+
+	got = ps.SelectSyncPeer([]*blockchain.PeerInfo{bad}, advertisedProbeCriteria(50))
+	require.Empty(t, got, "a blacklisted peer must not be selected even as the only candidate")
+}
--- a/services/p2p/server_helpers.go
+++ b/services/p2p/server_helpers.go
@@ -361,9 +361,10 @@ func (s *Server) validateDataHubURL(urlStr string) error {
 		return errors.NewInvalidArgumentError("DataHubURL has invalid scheme: %s (only http/https allowed)", parsed.Scheme)
 	}
 
-	// Strip the trailing dot of a rooted FQDN so "localhost." or "127.0.0.1."
-	// cannot slip past the checks below.
-	hostname := strings.TrimSuffix(parsed.Hostname(), ".")
+	// Canonicalize before checking: strip the trailing dot of a rooted FQDN and
+	// lowercase, so "localhost.", "LOCALHOST" or "127.0.0.1." cannot slip past
+	// the checks below (DNS resolution is case-insensitive).
+	hostname := strings.ToLower(strings.TrimSuffix(parsed.Hostname(), "."))
 	if hostname == "" {
 		return errors.NewInvalidArgumentError("DataHubURL has no hostname")
 	}
--- a/services/p2p/server_helpers.go
+++ b/services/p2p/server_helpers.go
@@ -309,7 +309,9 @@ func (s *Server) extractHost(urlStr string) string {
 		return ""
 	}
 
-	host := parsedURL.Hostname()
+	// Strip the trailing dot of a rooted FQDN so "evil.example." matches a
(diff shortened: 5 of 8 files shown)


ctnguyen (changes requested):
Review at commit [c36d759](https://github.com/gokutheengineer/teranode/commit/c36d759dbc2188ed0a305bc394d07261323193c2)

This closes a real and clearly-articulated gap: on main, only `handleSubtreeTopic` ran the operator blacklist, so a blacklisted DataHub host was correctly refused for subtrees yet still accepted for block and node_status announcements — and for blocks that meant the blacklisted host went on to feed catchup, while for node_status the blacklisted `BaseURL` was persisted into the peer registry that sync/catchup reads from. Factoring the two trust checks into `checkDataHubURL` is the right call: it removes the divergence between the three handlers by construction rather than by three parallel edits, and the semantics are preserved exactly — SSRF failure still counts as a protocol violation (peer penalised via `addProtocolViolation`), while a blacklist match still drops the message without a ban score, since that is an operator policy choice and not peer misbehaviour. The tests are well-targeted: the block test verifies the full drop (no notification, no registry entry, no Kafka publish) *and* asserts the peer is not banned, and it uses `http://evil.example:8080/path` against a `http://evil.example` blacklist entry, so it exercises host-level matching via `extractHost` rather than a trivial string equality. Placement is correct too — in all three handlers the check runs before the notification channel send, registry write, and Kafka publish. Two items below; the first is blocking.

## Blocking issues

### ChiR1 — hostname canonicalization bypasses the blacklist (and the localhost SSRF guard)
**Problem:** The host comparison never canonicalizes a DNS trailing dot. `extractHost` (`server_helpers.go:315-327`) only calls `Hostname()` and `ToLower` — it does not strip a trailing `.`. So `url.Parse("http://evil.example./").Hostname()` is `evil.example.`, which does not equal the blacklist entry's extracted host `evil.example`. Both `input` and `blocked` go through `extractHost` in `isBlacklistedBaseURL`, so a blacklist entry of `http://evil.example` fails to match an announced `http://evil.example./` — the message is accepted. The same normalization gap exists in the SSRF localhost check: `validateDataHubURL` (`server_helpers.go:385`) tests `isLocalhostHostname("localhost.")`, which matches neither `== "localhost"` nor `HasSuffix(".localhost")`, so `http://localhost./` slips past the loopback guard.

**Why it matters:** The trailing-dot form is a fully-resolvable, equivalent hostname (a rooted FQDN), so this is a trivial one-character evasion of the exact control this PR exists to enforce. This PR newly routes `handleBlockTopic` and `handleNodeStatusTopic` through the blacklist, which makes the blacklist-bypass *newly reachable* on the block and node_status paths — an operator who blacklists `http://evil.example` still gets catchup pointed at it, and the peer's `BaseURL` still lands in the registry, via `http://evil.example./`. Being precise about scope: the localhost/SSRF half is *pre-existing* (`validateDataHubURL` is unchanged here), but it shares the same root cause and is cheap to fix in the same spot.

**Fix:** Strip a single trailing dot during host normalization, in both places:

```go
// extractHost, before returning:
host := parsedURL.Hostname()
if host == "" {
    return ""
}
host = strings.TrimSuffix(host, ".")
return strings.ToLower(host)

// validateDataHubURL, before the localhost check:
hostname := parsed.Hostname()
hostname = strings.TrimSuffix(hostname, ".")
```

Add a case to the block/subtree/node_status blacklist tests that announces the trailing-dot form against a non-dotted blacklist entry, and a `validateDataHubURL` case for `http://localhost./`.

## Non-blocking issues

### ChiR2 — node_status blacklist drop also suppresses monitoring telemetry
**Problem:** In `handleNodeStatusTopic`, a blacklisted `BaseURL` now causes an early `return` before the WebSocket notification is sent (`Server.go:993`, ahead of the `notificationCh` send at `Server.go:1037`) and before the sender is added to the registry (`Server.go:1093`). node_status is primarily a telemetry message consumed by the dashboard, so alongside the intended exclusion from catchup, the dashboard also stops receiving status for that peer.

**Why it matters:** The security objective — never store or catchup from a blacklisted URL — is met purely by not persisting the `BaseURL` into the registry (the `addPeer` at `Server.go:1081`). Dropping the WebSocket forward as well is an observability tradeoff rather than a correctness requirement: it is stricter than the objective needs, and it costs an operator the dashboard signal for a peer they have deliberately blacklisted. This is a defensible judgment call — it mirrors the existing SSRF early-return, which also drops the whole message — so it is called out as a tradeoff to confirm, not a defect.

**Fix:** Optional. If preserving observability is preferred, follow the pattern the unhealthy-peer path already uses at `Server.go:1027-1031` — treat the blacklist match like an unhealthy peer (skip peer-data storage but still forward to WebSocket) instead of returning outright. If the current full-drop is intentional, no change needed.

## Recap

| ChiR1 | hostname canonicalization bypass | YES | 70% |
| ChiR2 | node_status drop hides telemetry | | 25% |

ctnguyen (changes requested):
Review at commit [d290536](https://github.com/gokutheengineer/teranode/commit/d290536a92adddfc62d45eeedb83be2234a22490)

Both points from the previous round are handled, and the shared-helper factoring (`checkDataHubURL`) is a clean way to remove the divergence between the handlers while preserving the semantics (SSRF failure bans, blacklist match drops silently). ChiR1 is closed in both `extractHost` and `validateDataHubURL` with matching tests, and ChiR2's strip-and-forward keeps the dashboard telemetry flowing. The re-review did turn up one problem, though: the node_status strip doesn't actually deliver the "never in the registry" guarantee for a peer that is already registered — written up as blocking below.

## Status of previous points

| ChiR1 | Hostname canonicalization bypass | YES | FIXED |
| ChiR2 | node_status drop hides telemetry | | FIXED |

## Blocking issues

There are still one blocking issue that need to be fixed before merge

### ChiR3 — node_status strip is a no-op for an already-registered peer, so a mid-operation blacklist can't stop catchup
**Problem:** The blacklist is enforced only at gossip *ingestion*, never at the point of *use* (catchup), and for an already-registered peer the node_status strip cannot remove a URL that is already stored. `handleNodeStatusTopic` sets a blacklisted `BaseURL` to `""` and passes it to `addPeer(..., "")`, but `CentralizedPeerRegistry.Register` only overwrites an existing peer's URL when the new value is non-empty, while still refreshing `LastSeen` unconditionally (`services/blockchain/peer_registry.go:308,366`):

```go
if info.DataHubURL != "" {
    existing.DataHubURL = info.DataHubURL
}
...
existing.LastSeen = now   // refreshed even when DataHubURL was empty
```

So the strip cannot evict a URL that predates the blacklist entry. Sequence: an active peer advertises `http://evil.example` while it is not blacklisted → stored; the operator then blacklists it; the peer keeps announcing → node_status strips to `""` (a no-op for the registry) and block/subtree announcements are now dropped entirely, so nothing refreshes *or clears* the field. Catchup reads `DataHubURL` straight from the registry with no blacklist check of its own (`grep isBlacklistedBaseURL` shows the only call sites are the three gossip handlers).

**Why it matters:** This defeats the guarantee the change asserts in its own comment — "a blacklisted BaseURL must never reach the peer registry (catchup would fetch from it)" — for the primary operational case: an operator blacklists a host in response to observed abuse, i.e. a host a live peer has already registered. TTL/eviction does not rescue it: `Cleanup` keys on `peerActivity`, the newer of `LastSeen`/`LastMessageTime` (`peer_registry.go:1284,1336-1347`), and both are refreshed on every node_status (`Register` bumps `LastSeen`; `handleNodeStatusTopic` calls `updatePeerLastMessageTime`), so an active peer is never expired and the stale blacklisted URL persists indefinitely while catchup keeps using it. (My earlier note that eviction "eventually clears it" was wrong for an active peer.)

**Fix:** Enforce the blacklist at the point of use — check `isBlacklistedBaseURL` against the stored `DataHubURL` before catchup fetches from a peer — so the entry takes effect regardless of when the URL was ingested. If ingestion-time eviction is preferred instead, the strip must actively clear the stored field (Register's empty-skip means `""` won't do it), e.g. a small `UpdateDataHubURL(peerID, "")` that unconditionally sets it:

```go
if s.isBlacklistedBaseURL(nodeStatusMessage.BaseURL) {
    s.logger.Warnf("[handleNodeStatusTopic] removed blacklisted BaseURL %s from node_status of peer %s", nodeStatusMessage.BaseURL, peerID)
    nodeStatusMessage.BaseURL = ""
    s.clearPeerDataHubURL(peerID) // actively evict any URL stored before the blacklist entry
}
```

Add a regression test where a peer is first registered with a URL, that URL is then added to the blacklist, and a subsequent node_status (and a catchup peer selection) must not surface the blacklisted URL — the current test only exercises a fresh registry, which is exactly the case the strip does handle.

icellan (commented):
**DataHub blacklist matcher has two bypasses that make this PR's broadening ineffective for common inputs.**

**P1 — fix before merge**
- `server_helpers.go:286-300` — scheme-less blacklist entries silently never match. An entry like `evil.example` (no scheme) parses as a path, so `extractHost` returns `""` and the code falls back to exact string compare, which a real full-URL announcement never satisfies. The host is effectively un-blacklisted; catchup still fetches from it. Normalize entries at config load (prepend a scheme) or match on host-suffix, and warn on un-parseable entries so silent-inert config is visible.
- `server_helpers.go:314` and `:367` — `TrimSuffix(host, ".")` strips only one trailing dot. `evil.example..` → `evil.example.` ≠ `evil.example`, bypassing the blacklist; `127.0.0.1..` likewise skips the SSRF loopback guard. Use `TrimRight(host, ".")`.

**P2 — decide intentionally**
- `Server.go:993-1008` — a DataHubURL registered before its host was blacklisted is never evicted: the empty-string strip is a no-op in `Register` (only overwrites when non-empty), block/subtree paths `return` before `addPeer`, and peer-selection does no blacklist check. Only triggers if the blacklist changes mid-run or the registry outlives a restart — confirm that precondition, then either clear on strip or blacklist-check at selection.
- `server_helpers.go:70` — the block handler now drops the WebSocket/Kafka notification on a blacklist match, unlike node_status which keeps telemetry. Fine if intended; flagging so it's a deliberate choice.

Nits (severity Errorf→Warnf at `:402`, duplicated trust sequence, duplicated host normalization, per-announcement blacklist re-parse) left out of this comment.

ctnguyen (commented):
Rereview at commit [e382372](https://github.com/gokutheengineer/teranode/commit/e382372c6308dc6e78bfadbcd2251749fa496923)

Nice follow-through on this round — all three of my previous points are genuinely handled, not papered over. The trailing-dot canonicalization landed in both `extractHost` and `validateDataHubURL` with matching table cases, the node_status strip-and-forward keeps the dashboard telemetry flowing, and ChiR3 was answered the robust way: point-of-use enforcement in `GetPeersForCatchup` and `PeerSelector.isEligible`, backed by a regression test that actually reproduces the stored-before-blacklist sequence. Promoting the matcher to a package-level `isBaseURLBlacklisted` so the selector and the gossip handlers share one implementation is the right instinct — it removes the drift that started this whole PR. The tests are well-aimed too (host-level match with different port/path, trailing-dot form, and the "not banned" assertion that distinguishes operator policy from protocol violation). A few normalization edge cases in the matcher still let a host slip the blacklist; one of them is a live silent gap and is blocking, the rest are smaller. Two of these were also raised independently by icellan after this head commit.

## Status of previous points

ChiR3 is fixed via point-of-use filtering rather than active eviction; the stored registry entry is still never physically cleared, which is a smaller residual tracked below as ChiR6.

| ChiR1 | Hostname canonicalization bypass | YES | FIXED |
| ChiR2 | node_status drop hides telemetry | | FIXED |
| ChiR3 | Mid-op blacklist can't stop catchup | YES | FIXED |

## Blocking issues

### ChiR4 — scheme-less blacklist entries silently never match

*Same as [@icellan's point P1 "scheme-less blacklist entries" (22-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1297#pullrequestreview-4754949281).*

**Problem:** In `isBaseURLBlacklisted` (`server_helpers.go`), each blacklist entry is run through `extractHost`. A scheme-less entry such as `evil.example` parses as a URL *path*, so `url.Parse("evil.example").Hostname()` is `""`, `blockedHost` is `""`, and the code falls back to exact string comparison `baseURL == blocked`. Real announcements always carry a full URL (`http://evil.example:8080/...`), so that equality never holds — the entry is dead and the host is effectively un-blacklisted. Nothing warns the operator, and there is no test covering a scheme-less entry (every test uses `http://evil.example`). This PR widens the blast radius: the same silently-inert entry now fails to protect the block, node_status, and catchup paths, not just subtrees.

**Why it matters:** This is the exact control the PR exists to enforce, failing silently for a plausible operator input. An operator who blacklists `evil.example` in response to observed abuse gets no protection at all — catchup keeps fetching from it — while believing the host is blocked. Silent failure of a security control is worse than a loud one.

**Fix:** Normalize entries before matching in the package-level `isBaseURLBlacklisted` (no logger here, so keep this change logging-free):

```go
blockedHost := extractHost(blocked)
if blockedHost == "" {
    // Retry scheme-less entries ("evil.example") as a bare host.
    blockedHost = extractHost("//" + blocked)
}
if blockedHost == "" {
    // still un-parseable: keep the exact-match fallback
    if baseURL == blocked {
        return true
    }
    continue
}
```

Separately, warn on truly un-parseable entries at config load (where a logger is available) so a dead entry is visible rather than silently inert. Add a table case with a scheme-less entry (`evil.example`) against a full-URL announcement.

## Non-blocking issues

### ChiR5 — normalization strips only a single trailing dot

*Same as [@icellan's point P1 "TrimSuffix strips only one trailing dot" (22-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1297#pullrequestreview-4754949281).*

**Problem:** Both `extractHost` and `validateDataHubURL` use `strings.TrimSuffix(host, ".")`, which removes exactly one trailing dot. `evil.example..` normalizes to `evil.example.` (≠ `evil.example`), and `127.0.0.1..` similarly skips the loopback comparison. This is the same canonicalization family as ChiR1, just one dot deeper. In practice a doubled trailing dot is an empty DNS label and generally won't resolve, so live exploitability is weak — but the fix is trivial and strictly safer, and it hardens the loopback guard as a side effect.

**Why it matters:** It is a completeness gap in the normalization that ChiR1 just tightened; leaving it half-done invites a future "why does `..` behave differently" surprise, and the loopback half is a defense-in-depth check where being conservative is cheap.

**Fix:** Trim all trailing dots in both spots:

```go
host := strings.TrimRight(parsedURL.Hostname(), ".")
```

Add a `..` case alongside the existing trailing-dot tests.

### ChiR6 — stored registry entry is never evicted; enforcement relies on every consumer re-checking

*Same as [@icellan's point P2 "never evicted" (22-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1297#pullrequestreview-4754949281).*

**Problem:** ChiR3 is resolved by filtering at `GetPeersForCatchup` and `PeerSelector.isEligible`, but the blacklisted `DataHubURL` is never actually removed from the registry — the node_status strip passes `""` and `Register` ignores empty updates, so the stale URL lingers (the PR's own test asserts this). Correctness now depends on *every* current and future reader of `PeerInfo.DataHubURL` replicating the blacklist check. The two known catchup entry points are covered, so this is not a live bypass today, but a new consumer added later would silently reintroduce ChiR3.

**Why it matters:** Point-of-use filtering is the correct primary control, but a lingering blacklisted value in shared state is a latent trap: the safety property is not enforced by construction, only by convention at each call site. Actively clearing the field on strip gives belt-and-suspenders and removes the stale data.

**Fix:** In addition to the point-of-use checks, evict the stored value when node_status strips a blacklisted URL — e.g. a small `UpdateDataHubURL(peerID, "")` that unconditionally sets the field (plain `""` through `Register` won't, per its empty-skip). Consider a comment at the `DataHubURL` field pointing readers at the blacklist requirement.

### ChiR7 — blacklist drop logged at Errorf (noise + inconsistent with node_status)

**Problem:** `checkDataHubURL` logs a blacklist match with `s.logger.Errorf("[%s] blocked notification from blacklisted DataHubURL ...")`. A blacklist hit is an expected operator-policy outcome, not an error, and node_status already logs the same event at `Warnf`. A blacklisted peer that keeps gossiping will flood the error log and can trip error-rate alerting, potentially masking real errors.

**Why it matters:** Log level for a frequent, expected, security-relevant event has real operational impact (alert noise), and the inconsistency with the node_status path is avoidable.

**Fix:** Use `Warnf` (or `Debugf`) for the blacklist drop to match `handleNodeStatusTopic`:

```go
s.logger.Warnf("[%s] blocked notification from blacklisted DataHubURL %s (peer %s)", handlerName, dataHubURL, fromID)
```

## Recap

| ChiR4 | Scheme-less entries never match | YES | 65% |
| ChiR5 | Only one trailing dot stripped | | 45% |
| ChiR6 | Stale entry, per-consumer reliance | | 30% |
| ChiR7 | Errorf for expected drop | | 15% |

ctnguyen (approved):
Review at commit [265f05a](https://github.com/gokutheengineer/teranode/commit/265f05ac33c05572c937a68ac932fb9690419e9a)

This round lands the substance: every blocking point from the earlier rounds is genuinely fixed at this head, not papered over. `extractHost` and `validateDataHubURL` now trim *all* trailing dots and lowercase before comparing, scheme-less entries match via `blacklistEntryHost` with a loud startup warning for the truly unparseable ones, and the point-of-use filtering in `GetPeersForCatchup` and `PeerSelector.isEligible` is backed by regression tests that actually reproduce the stored-before-blacklist sequence. The `checkDataHubURL` factoring keeps the SSRF-vs-blacklist semantics distinct (SSRF failure bans, blacklist match drops silently) and the test table is thorough. What's left is smaller: two of the earlier non-blocking notes (ChiR6, ChiR7) are still open, and while re-checking ChiR6 I found the "never used for catchup" guarantee has a couple of concrete holes on paths that read `DataHubURL` straight from the registry. Nothing blocking this time.

## Status of previous points

ChiR4 and ChiR5 (raised at `e382372`) are both fixed at this head: `blacklistEntryHost` retries scheme-less entries in protocol-relative form, `NewServer` warns on entries with no parseable host, and both `extractHost`/`validateDataHubURL` use `TrimRight(host, ".")`. ChiR6 remains open — point-of-use filtering is present but the stored blacklisted URL is still never physically cleared, and I found two catchup-capable readers that bypass the two guarded entry points (tracked below as ChiR8), which is exactly the "every consumer must re-check" trap ChiR6 warned about. ChiR7 is unchanged: the blacklist drop in `checkDataHubURL` still logs at `Errorf`.

| ChiR1 | Hostname canonicalization bypass | YES | FIXED |
| ChiR2 | node_status drop hides telemetry | | FIXED |
| ChiR3 | Mid-op blacklist can't stop catchup | YES | FIXED |
| ChiR4 | Scheme-less entries never match | YES | FIXED |
| ChiR5 | Only one trailing dot stripped | | FIXED |
| ChiR6 | Stale entry, per-consumer reliance | | NOT FIXED |
| ChiR7 | Errorf for expected drop | | NOT FIXED |

icellan (commented):
Re-reviewed at `ccc67f35a`. Both P1 blacklist bypasses are fixed (scheme-less `//host` retry, `TrimRight` for multi-dot), and enforcing the blacklist at point-of-use in `isEligible`/`sendSyncTriggerToKafka` is a cleaner fix for the stale-URL concern than evicting the registry. One new issue from that enforcement:

**`sync_coordinator.go:315` — `isCaughtUp` and peer selection disagree on the blacklist.** The check was added to `isEligible` but not to `isViableSyncCandidate` (`:131`), which `isCaughtUp` uses at `:315`. If a behind node's only ahead peer has a blacklisted DataHub host, `isCaughtUp()` returns false (peer is viable + ahead by validated work) while `SelectSyncPeer` filters it out and returns `""` — the node reports not-caught-up forever and busy-loops on failed selection instead of parking. Contradictory operator config, but worth deciding: either teach `isViableSyncCandidate`/`isCaughtUp` about the blacklist so the caught-up view matches selection, or confirm the FSM tolerates an indefinite not-caught-up state without blocking block assembly.

gokutheengineer:
@icellan  Went with option 1, fixed.  Added a coordinator-level isViableSyncCandidate that applies the blacklist, used by all decision paths, plus a test for the caught-up flip.

icellan (approved):

PR #1298, opened 2026-07-16
p2p: fail startup on HTTP bind/TLS config errors and surface serve failures via Health

gokutheengineer:
### Problem
- `StartHTTP` returned `nil` before `Serve`/`ServeTLS` ran; serve errors were only logged inside a detached goroutine.
- Missing `ServerCertFile`/`ServerKeyFile` in HTTPS mode just logged and returned from the goroutine, leaving the node running with no HTTP surface.
- `Start` launched `StartHTTP` in a fire-and-forget goroutine and closed `readyCh` regardless, and `Health(checkLiveness=true)` returned 200 unconditionally, so orchestration kept routing to a node whose HTTP/websocket endpoint was down.
- The shared `err` variable was written by the serve goroutine after `StartHTTP` had returned (dead write / latent race).

### Fix
- Validate cert/key synchronously in `StartHTTP` and return a `ConfigurationError`, so HTTPS misconfiguration fails startup instead of being swallowed.
- Call `StartHTTP` synchronously from `Start` (it only binds; serving stays in its own goroutine), so listener bind and TLS config errors propagate out of `Start` before readiness is signalled.
- The serve goroutine records any unexpected `Serve`/`ServeTLS` exit in a new `httpServeErr atomic.Pointer[error]` field (goroutine-local `serveErr` fixes the dead write). `Health` liveness returns 503 when set, and readiness gains an `HTTPServer` check.
- A recorded-error + health-check approach was chosen over a short select window after startup: a fixed window is racy and only covers startup, while the health check catches serve-goroutine death at any point in the process lifetime.

### Tests
Added `services/p2p/server_http_test.go`:
- `TestStartHTTPMissingCertFailsStartup` / `TestStartHTTPMissingKeyFailsStartup` - HTTPS mode without cert/key returns a `ConfigurationError`.
- `TestStartHTTPServeErrorSurfacesInHealth` - unreadable cert files make `ServeTLS` fail after startup; liveness flips to 503 and readiness reports the `HTTPServer` check as failed.
- `TestStartHTTPServesAndShutsDownCleanly` - plain HTTP serves `/health`, and a context-cancel shutdown (`ErrServerClosed`) is not reported as a failure.

Ran (affected package only):
```
```
All green except `TestAdminProtectedMethodsCoverAllRPCs`, which also fails on clean `upstream/main` (pre-existing, unrelated: `ReportValidatedChainProgress` unclassified in admin auth lists).

### Design decision: restart on HTTP death
A dead HTTP serve goroutine flips liveness to 503, not just readiness, so the service manager restarts the whole P2P process even though libp2p and gRPC may be healthy. This is deliberate: once `Serve`/`ServeTLS` returns, the goroutine is gone and the listener may be closed, so the HTTP/websocket surface cannot recover in-process; a restart is the only way to bring it back. Surfacing it via readiness only would stop routing but leave the node running permanently without its HTTP endpoint, which is worse. The rationale is documented in a comment in the `Health` liveness branch.


ctnguyen (approved):
Review at commit [d65822f](https://github.com/gokutheengineer/teranode/commit/d65822f7fc36b67aa9aa578c1c624c3e7cc064f3)

This is a clean, well-reasoned fix that closes a real gap: previously `Start` fired `StartHTTP` into a goroutine and swallowed its return, so a failed listener bind or a missing cert left the node "ready" with no HTTP surface, and the serve goroutine wrote the outer `err` variable after `StartHTTP` had already returned nil — a dead write to a returned-from variable, i.e. a race-prone pattern. Calling `StartHTTP` synchronously so listener bind errors and missing cert/key configuration propagate out of startup, moving to a goroutine-local `serveErr` recorded in `httpServeErr atomic.Pointer[error]`, switching to `errors.Is(..., http.ErrServerClosed)`, and surfacing serve-goroutine death through both liveness and readiness are all the right calls. Validating TLS config *before* `GetListener` also correctly avoids leaking a bound listener on a config failure, and the design note explaining why a recorded-error + health-check beats a fixed post-startup select window is sound — it catches serve death at any point, not just at boot. The tests are focused and cover the meaningful cases (missing cert/key, serve failure surfacing in health, and a clean serve/shutdown that must *not* be recorded as failure). The never-reset error pointer is fine here: `Serve`/`ServeTLS` only return once the HTTP server is genuinely down, so there is no state in which the recorded error is stale-but-serving. One gap around TLS validation completeness is worth a look before merge is not required — details below.

## Non-blocking issues

### ChiR1 — Present-but-invalid cert/key fails via crashloop instead of at startup

**Problem:** `StartHTTP` now validates only that `ServerCertFile`/`ServerKeyFile` are *non-empty* before binding:

```go
if s.settings.SecurityLevelHTTP != 0 {
    certFile = s.settings.ServerCertFile
    if certFile == "" {
        return errors.NewConfigurationError("[StartHTTP] server_certFile is required for HTTPS")
    }
    keyFile = s.settings.ServerKeyFile
    if keyFile == "" {
        return errors.NewConfigurationError("[StartHTTP] server_keyFile is required for HTTPS")
    }
}
```

A path that is set but unreadable, mismatched, or not a valid key pair passes this check. `StartHTTP` returns `nil`, `Start` completes, `readyCh` closes, and only then does `ServeTLS` fail in the goroutine, storing into `httpServeErr`. Liveness then returns 503, which signals "restart needed" — so a persistently bad cert produces a restart/crashloop rather than a clean startup failure.

**Why it matters:** This leaves an asymmetry against the PR's own stated goal ("fail startup on … TLS config errors"): an *empty* cert path fails startup with an immediate, legible `ConfigurationError`, but a *bad* cert path degrades to a delayed 503 and a crashloop. The crashloop is fail-loud and defensible, but it is a worse operator experience than an immediate error message, and the failure surfaces far from its cause. This is scoped as an improvement, not a defect — the health path does catch it — hence non-blocking.

**Fix:** Validate the key pair is actually loadable in `StartHTTP`, alongside the non-empty check, so a bad cert fails startup with a clear error for the same reasons an empty one does:

```go
if _, err := tls.LoadX509KeyPair(certFile, keyFile); err != nil {
    return errors.NewConfigurationError("[StartHTTP] failed to load TLS key pair", err)
}
```

That closes the gap between "missing" and "present-but-invalid" and keeps both on the fail-fast startup path.

freemans13:
One thing worth a conscious confirm: with this change, a dead HTTP serve goroutine now flips **liveness** to 503, not just readiness. Since the service manager treats a 503 liveness as "restart needed", an HTTP-surface failure will bounce the whole P2P process — killing the healthy libp2p peer connections and the gRPC server along with it.

I think that's defensible, and it reads as deliberate: once `Serve`/`ServeTLS` returns, the goroutine is gone and the listener may be closed, so the HTTP surface can't recover in-process — a restart is the only way to bring it back. So restarting is arguably the right call.

The alternative would be to surface it via readiness only (stop routing to the node) and leave liveness at 200, on the grounds that P2P's core job is libp2p + gRPC and the HTTP surface is secondary. But that would leave the node running permanently without its HTTP/websocket endpoint, which is probably worse.

No change requested — just flagging so the restart-on-HTTP-death behaviour is a chosen outcome rather than an incidental one. Everything else in the PR looks clean to me (the `serveErr` goroutine-local fix nicely closes the old dead-write race).

freemans13 (approved):

PR #1299, opened 2026-07-16
fix(p2p): classify ReportValidatedChainProgress in RPC auth coverage test

gokutheengineer:
### Problem
`TestAdminProtectedMethodsCoverAllRPCs` (added in #1257) fails on main: the `ReportValidatedChainProgress` RPC landed alongside that PR and is not classified as either admin-protected or public, so the classification check trips exactly as designed.

### Fix
Classify `ReportValidatedChainProgress` as an internal data-plane reporting RPC in `publicPeerServiceMethods`. It is called continuously by block validation (`services/blockvalidation/peer_metrics_helpers.go`) through the p2p client to report locally validated header-chain progress - the same family as `ReportValidBlock`/`ReportValidSubtree`, not an admin operation, so it must stay reachable without the admin API key.

### Tests
```
```
Full p2p package green, including the previously failing `TestAdminProtectedMethodsCoverAllRPCs`.

### DIFF
--- a/services/p2p/server_auth_test.go
+++ b/services/p2p/server_auth_test.go
@@ -29,15 +29,16 @@ var publicPeerServiceMethods = map[string]bool{
 	"/p2p_api.PeerService/GetPeerRegistry":    true,
 	"/p2p_api.PeerService/GetPeer":            true,
 	// Internal data-plane reporting
-	"/p2p_api.PeerService/RecordCatchupAttempt":    true,
-	"/p2p_api.PeerService/RecordCatchupSuccess":    true,
-	"/p2p_api.PeerService/RecordCatchupFailure":    true,
-	"/p2p_api.PeerService/RecordCatchupMalicious":  true,
-	"/p2p_api.PeerService/UpdateCatchupReputation": true,
-	"/p2p_api.PeerService/UpdateCatchupError":      true,
-	"/p2p_api.PeerService/ReportValidSubtree":      true,
-	"/p2p_api.PeerService/ReportValidBlock":        true,
-	"/p2p_api.PeerService/RecordBytesDownloaded":   true,
+	"/p2p_api.PeerService/RecordCatchupAttempt":         true,
+	"/p2p_api.PeerService/RecordCatchupSuccess":         true,
+	"/p2p_api.PeerService/RecordCatchupFailure":         true,
+	"/p2p_api.PeerService/RecordCatchupMalicious":       true,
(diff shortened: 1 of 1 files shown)


oskarszoon (approved):
Correct fix, approving. Verified the classification rather than taking it on faith: runtime p2p auth is a denylist (`Server.go:686` wires `ProtectedMethods: adminProtectedMethods()`), so only the seven state-mutating admin RPCs require the key and `ReportValidatedChainProgress` is already public in production — the test just tripped the unclassified-RPC guard. Public is also the forced choice: the internal caller in block validation (`p2p/Client.go:594`) carries no admin key, so admin-gating it would self-DoS the progress reporting. The handler is safe for an unauthenticated caller — validates peer/height/hash, caps `ChainWork` at `blockchain.MaxValidatedChainWorkBytes`, non-fatal on registry failure, no ban/connect/consensus effect — and it's the same family as `ReportValidBlock`/`ReportValidSubtree`. No streaming RPC added, so the stream-auth guard still holds. `go test -race ./services/p2p/` for the two auth tests passes locally.

ctnguyen (approved):
Review at commit [d8cf840](https://github.com/gokutheengineer/teranode/commit/d8cf8406b4f5a47d7b69bf4338a0605124cfb193)

This is the right fix, and a well-scoped one. `TestAdminProtectedMethodsCoverAllRPCs` is a deliberate fail-closed gate — a new PeerService RPC that lands unclassified trips it by design — and `ReportValidatedChainProgress` (added around #1257) was exactly that omission. The judgment call this PR makes is the important part, and it gets it right: classifying the RPC as PUBLIC rather than admin-protected. It belongs to the same inter-service data-plane family as `ReportValidBlock`/`ReportValidSubtree`/`RecordCatchup*`, is called advisory-only by block validation (`services/blockvalidation/peer_metrics_helpers.go:128`) as an internal inter-service data-plane operation that is not (and must not be required to be) admin-authenticated, so admin-protecting it would break that inter-service path outright. The handler is also defensively coded for an unauthenticated surface — peer-ID decode, non-zero height, block-hash parse, non-empty chainwork, a `maxReportedChainWorkBytes` cap, and non-fatal registry-failure handling. The remaining churn is pure gofmt map-value re-alignment. Worth noting for context, not as a finding here: the whole `Report*`/`Record*` family is unauthenticated yet mutates peer-registry state used for sync-peer selection — a pre-existing trust-model property tied to earlier hardening (#1201), neither introduced nor widened by this test-only change. Nothing blocks; nothing to nit.

PR #1301, opened 2026-07-17
MvP-4690 Enforce mandatory block-version floors (BIP34/66/65)

ctnguyen:
Resolve https://github.com/bitcoin-sv/teranode/issues/4690

## What

Teranode never rejected a block for carrying an outdated header version, and it gated its BIP34
coinbase-height check on `Header.Version > 1`, so a version-1 block skipped coinbase-height
verification entirely. This change brings Teranode into consensus parity with the reference
implementation.

- **New `model.CheckBlockVersion(version, height, params)`** — rejects `version < 2/3/4` at or after
  the network's `BIP0034Height` / `BIP0066Height` / `BIP0065Height`, mirroring bitcoin-sv
  `ContextualCheckBlockHeader` (`validation.cpp:5918-5924`). The version is compared as a **signed
  `int32`**, exactly as svnode compares `CBlockHeader::nVersion`, so a high-bit version (e.g.
  `0xffffffff` → `-1`) is treated as below the floor. Genesis (height 0) is exempt, matching svnode
  which never runs this check on the genesis block. The reject carries the `bad-version(0x%08x)`
  token for string parity.
- **Enforced on every active-chain acceptance path** — called from `Block.Valid` (before any
  body/coinbase checks, matching svnode's error priority) and as the first check in both
  quick-validation catchup functions (`quickValidateBlock` / `quickValidateBlockAsync`). The
  quick-path redundancy is deliberate defence-in-depth over the checkpoint guarantee.
  `Block.Valid` is the authoritative enforcement point; the block-validation service's outer coinbase
  prechecks are intentionally left ahead of the version floor. A complete below-floor block is
  rejected by `Block.Valid` with the `bad-version` token; a below-floor block that also fails those
  outer coinbase prechecks is rejected earlier with the documented non-parity reason
  (incomplete/bad-coinbase) instead. Either way the block is rejected — a deliberate error-priority
  non-parity only.
- **Retired `LastV1Block` (227835)** and moved the BIP34 coinbase-height gate onto
  `ChainCfgParams.BIP0034Height`, height-driven and version-independent (a `b.Height > 0` guard
  preserves the genesis exemption on networks where `BIP0034Height == 0`). This closes the v1
  coinbase-height bypass and the 227835-vs-227931 boundary discrepancy.
- **Empty mining candidate** now uses the safe `0x20000000` version instead of inheriting the tip's
  version, so Teranode never mines a below-floor block at an activation boundary.
- **SQL blockchain store** — `validateCoinbaseHeight` now derives its activation height from
  `chainParams.BIP0034Height` and no longer skips v1 blocks via a version sub-condition, removing a
  second copy of the stale hardcoded height and version bypass.

## Why

Consensus parity with bitcoin-sv `ContextualCheckBlockHeader` (`validation.cpp:5918-5924`) and
`ContextualCheckBlock` (`validation.cpp:6039`). At a height where the rest of the network rejects an
outdated-version block, Teranode previously accepted it (and additionally skipped BIP34
coinbase-height verification for v1 blocks) — a chain-split vector in the dangerous direction.

**Chain-split analysis:** this change is net split-*reducing*. Every block on the live
mainnet/testnet chains already satisfies the floor (they are, by definition, the chains svnode
accepted), so no historic block can be rejected during IBD/catchup. The activation heights are
sourced from `go-chaincfg`, which matches svnode's mainnet/testnet/regtest values exactly, and the
rule uses the same operator (`>=`), the same height variable (the block's own height), the same
signed comparison, and the same genesis exemption as svnode — so it rejects exactly what svnode
rejects and introduces no new reverse-split on mainnet/testnet/regtest.

## Deliberately out of scope

- **Main mining candidate** (`BlockAssembler.go:1533`): already `0x20000000`; safe.
- **Catchup header validators** (`catchup/header_validation.go`): stateless header helpers with no
  height/params in scope; the floor is contextual and every active-chain acceptance path already
  enforces it.
- **RPC reject mapping** (`services/rpc/Server.go:454-455`): commented-out and unused; left as-is.

## Note for maintainers — STN BIP65/66

`go-chaincfg` sets STN `BIP0065Height`/`BIP0066Height` to 1251/1351, but svnode `CStnParams`
assigns only `BIP34Height` and leaves BIP65/66 unset (`chainparams.cpp:1212`), so exact STN BIP65/66
parity cannot be confirmed from svnode source. This change uses go-chaincfg's values (Teranode's own
source of truth). STN is a non-value test network; please confirm the intended STN BIP65/66 behaviour.

### DIFF
--- a/model/Block.go
+++ b/model/Block.go
@@ -46,8 +46,37 @@ var bufioReaderPool = sync.Pool{
 
 const GenesisBlockID = 0
 
-// LastV1Block https://github.com/bitcoin/bips/blob/master/bip-0034.mediawiki
-const LastV1Block = 227_835
+// heightAtOrAfterActivation reports whether block height h is at or after an activation height
+// taken from chaincfg. chaincfg activation heights are int32 and non-negative for BIP34/65/66; a
+// negative value (should not occur) is treated as "never active" so it can never wrongly reject.
+func heightAtOrAfterActivation(h uint32, activation int32) bool {
+	if activation < 0 {
+		return false
+	}
+
+	return h >= uint32(activation)
+}
+
+// CheckBlockVersion enforces the BIP34/66/65 mandatory version floors, mirroring bitcoin-sv
+// ContextualCheckBlockHeader (validation.cpp:5918-5924). version is compared as a SIGNED int32
+// exactly as svnode compares CBlockHeader::nVersion, so a high-bit version (e.g. 0xffffffff -> -1)
+// is treated as below the floor. Genesis (height 0) is exempt, matching svnode which never runs
+// this check on the genesis block. Returns a BlockInvalidError carrying the bad-version token.
+func CheckBlockVersion(version uint32, height uint32, params *chaincfg.Params) error {
+	if height == 0 {
+		return nil
+	}
+
+	v := int32(version)
+	if (v < 2 && heightAtOrAfterActivation(height, params.BIP0034Height)) ||
+		(v < 3 && heightAtOrAfterActivation(height, params.BIP0066Height)) ||
+		(v < 4 && heightAtOrAfterActivation(height, params.BIP0065Height)) {
+		return errors.NewBlockInvalidError(
+			"bad-version(0x%08x) rejected nVersion=0x%08x block", version, version)
+	}
+
+	return nil
+}
 
 var (
 	emptyTX = &bt.Tx{}
@@ -539,6 +568,13 @@ func (b *Block) Valid(ctx context.Context, logger ulogger.Logger, subtreeStore S
 		}
 	}
 
+	// 3b. Reject outdated block versions once the matching upgrade has activated (BIP34/66/65).
+	// Parity with bitcoin-sv ContextualCheckBlockHeader; must run before body/coinbase checks to
+	// match svnode's bad-version error priority.
+	if err := CheckBlockVersion(b.Header.Version, b.Height, settings.ChainCfgParams); err != nil {
+		return false, errors.NewBlockInvalidError("[BLOCK][%s] outdated block version", b.String(), err)
+	}
+
 	// 4. Check that the coinbase transaction is valid (reward checked later).
 	if b.CoinbaseTx == nil {
 		return false, errors.NewBlockInvalidError("[BLOCK][%s] block has no coinbase tx", b.String())
@@ -571,8 +607,14 @@ func (b *Block) Valid(ctx context.Context, logger ulogger.Logger, subtreeStore S
 
 	// TODO - do this another way, if necessary
 
-	// 5. Check that the coinbase transaction includes the correct block height.
-	if b.Header.Version > 1 && b.Height > LastV1Block {
+	// 5. Check that the coinbase transaction includes the correct block height (BIP34).
+	// Parity with bitcoin-sv ContextualCheckBlock: enforced for every block at/after BIP34Height.
+	// Version < 2 blocks are already rejected above (bad-version), so no version sub-condition here.
+	// The explicit b.Height > 0 guard is MANDATORY: on teratestnet/tstn BIP0034Height == 0, so
+	// heightAtOrAfterActivation(0, 0) is true and a height-0 (genesis-like) block driven through
+	// Valid() would otherwise attempt ExtractCoinbaseHeight — contradicting the genesis exemption
+	// that CheckBlockVersion already applies. svnode never runs this check on genesis.
+	if b.Height > 0 && heightAtOrAfterActivation(b.Height, settings.ChainCfgParams.BIP0034Height) {
 		height, err := b.ExtractCoinbaseHeight()
 		if err != nil {
 			return false, errors.NewBlockInvalidError("[BLOCK][%s] error extracting coinbase height", b.String(), err)
--- a/model/Block.go
+++ b/model/Block.go
@@ -597,15 +597,11 @@ func (b *Block) Valid(ctx context.Context, logger ulogger.Logger, subtreeStore S
 		return false, errors.NewBlockInvalidError("[BLOCK][%s] bad coinbase length", b.String())
 	}
 
-	// We can only calculate the height from coinbase transactions in block versions 2 and higher
-
 	// https://en.bitcoin.it/wiki/BIP_0034
-	// BIP-34 was created to force miners to add the block height to the coinbase tx.
-	// This BIP came into effect at block 227,835, which is after the first halving
-	// at block 210,000.  Therefore, until this happened, we do not know the actual
-	// height of the block we are checking for.
-
-	// TODO - do this another way, if necessary
+	// BIP-34 forces miners to encode the block height in the coinbase tx. It activates per network
+	// at ChainCfgParams.BIP0034Height; before that height the coinbase need not encode the height, so
+	// the check below is skipped. Enforcement is height-driven and version-independent — blocks below
+	// the mandatory version floor are already rejected above by CheckBlockVersion.
 
 	// 5. Check that the coinbase transaction includes the correct block height (BIP34).
 	// Parity with bitcoin-sv ContextualCheckBlock: enforced for every block at/after BIP34Height.
--- a/model/Block_test.go
+++ b/model/Block_test.go
@@ -372,7 +372,8 @@ func TestBlock_Valid_ComprehensiveCoverage(t *testing.T) {
 		coinbase, err := bt.NewTxFromString(CoinbaseHex)
 		require.NoError(t, err)
 
-		// Set a height at which the coinbase-height validation path is exercised
+		// With regtest params (CreateBaseTestSettings) BIP34 activates at 1e8, so height 300000 is
+		// below it and the coinbase-height check is not reached.
 		const testHeight = 300000
 		block, err := NewBlock(blockHeader, coinbase, []*chainhash.Hash{}, 1, 123, testHeight, 0)
 		require.NoError(t, err)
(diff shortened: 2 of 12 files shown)


ordishs (changes requested):
Solid, carefully-scoped consensus fix. I verified the parity claims against svnode source (`validation.cpp` / `chainparams.cpp` / `block.h`): the floor condition, the `bad-version(0x%08x)` / `rejected nVersion=0x%08x block` strings, the signed-int32 comparison, the mainnet activation heights (227931/363725/388381), the height-driven BIP34 coinbase rule with no version sub-condition, and the genesis exemption all match. The integration points also check out: quick-validation is gated on `block.Height <= highestCheckpointHeight` so the height is settled before `CheckBlockVersion` runs, and the SQL store's genesis branch short-circuits before `validateCoinbaseHeight`. Test coverage is strong and the new tests pass locally.

Requesting one change before merge:

**`validateCoinbaseHeight` now attempts coinbase-height extraction on every pre-BIP34 block** (`stores/blockchain/sql/StoreBlock.go`). The old code only extracted for `Version > 1` blocks; the new code runs `ExtractCoinbaseHeight` for every non-genesis block and emits a `Warnf` on each failure. During a mainnet IBD that's up to ~227k early v1 blocks whose coinbases mostly don't parse as a height push — a warning per block is significant log spam and wasted work, and the result is discarded when `!postBIP34` anyway. Early-return instead:

```go
if !postBIP34 {
    return nil
}
```

This also matches svnode more closely, which doesn't inspect the coinbase at all below `BIP34Height`.

Minor / follow-ups (non-blocking):

1. **Byte-exact BIP34 prefix (pre-existing, suggest a follow-up issue).** svnode requires the coinbase scriptSig to begin with the byte-exact minimal `CScript() << nHeight` serialization; `ExtractCoinbaseHeight` extracts a number and compares it, so a non-minimally-encoded height (e.g. `02 64 00` for height 100, or an OP_PUSHDATA1 push) passes Teranode but is rejected by svnode — a residual accept-side split vector on the exact rule this PR hardens. Predates this PR, but worth tracking since the parity analysis otherwise reads as complete.
2. **Magic number duplication.** `0x20000000` is now hardcoded in both the main and empty candidate paths in `BlockAssembler.go`; a shared named constant would keep them from drifting.
3. **Error-type shift on quick-validate paths (just noting, looks deliberate).** The version floor now runs before the missing-coinbase check, so a below-floor block from a seeded peer lacking coinbase data returns `BlockInvalid` instead of `BlockIncomplete`, which catchup handles differently. Correct — header version plus height alone prove invalidity — but it is a behavioral change on that corner.

Also seconding the PR's own maintainer note: the STN BIP65/66 heights in go-chaincfg (1251/1351) can't be confirmed from svnode's `CStnParams`, so whoever owns STN should confirm the intended behaviour before this rides a release.

oskarszoon (approved):
Reviewed as a consensus change across bitcoin-parity, security/split-vector, Go and second-eyes lenses, verifying each claim against svnode source and go-chaincfg rather than the PR narrative. The core is sound and I read it as net split-reducing: the signed-int32 comparison, the `>=` operator, the mainnet/testnet/regtest activation heights (227931/363725/388381, 21111/330776/581885, 100000000/1251/1351) all match svnode exactly, the genesis exemption holds on both paths including the mandatory `b.Height>0` guard for the `BIP0034Height==0` nets, `CheckBlockVersion` is a faithful mirror of `ContextualCheckBlockHeader`, the `LastV1Block` retirement closes the v1 coinbase-height bypass and the 227835-vs-227931 boundary, and the empty mining candidate no longer inherits the tip version. No P0/P1 defect. Two gates before merge, plus one confirm-before-release item.

**Gate 1 — CI is red on formatting.** The 3-line comment before `Version: 0x20000000` in `generateEmptyBlockCandidate` split the struct-literal alignment group, so goimports/gci wants `Id`/`PreviousHash`/`CoinbaseValue` re-padded (`BlockAssembler.go:1602`). The PR lint job doesn't suppress the exit code, so it's a hard red gate. One command: `gofmt -w` / `golangci-lint fmt`.

**Gate 2 — seconding ordishs' `validateCoinbaseHeight` change.** Confirmed: dropping the old `Version>1` short-circuit means `ExtractCoinbaseHeight` now runs for every non-genesis block, discarded below BIP34. His `if !postBIP34 { return nil }` early-return is correct, aligns with the already-gated `Block.Valid` path, and matches svnode. One refinement to the rationale though: the "~227k warning lines" is smaller than described — early mainnet coinbases start with the nBits `0x04` push, so `ExtractCoinbaseHeight` *succeeds* and silently discards the result with no warning; only the minority whose first byte >8 logs. So the real cost is a cheap per-block extraction during from-genesis IBD, not a warning storm. Still worth the cleanup.

**Confirm before any STN release — BIP65/66 parity.** go-chaincfg gives STN `BIP0065Height=1351`/`BIP0066Height=1251` (copied from regtest), but svnode `CStnParams` never assigns BIP65/66 — they're indeterminate, and svnode's floor reads them — so there's no defined svnode STN rule to match. No live impact (every real STN block is `0x20000000`, above the floor; window is empty) and mainnet/testnet/regtest are unaffected, but the intended STN rule should be pinned in go-chaincfg before STN version enforcement is relied on. Chi's own note and ordishs both called this.

Non-blocking follow-up on ordishs' byte-exact-prefix point (his #1): it's real and worth a tracked issue — `ExtractCoinbaseHeight` numeric-decodes, svnode requires the byte-exact minimal `CScript()<<nHeight`, so a non-minimal correct-value push is accept-side divergent. Two corrections though: it's pre-existing (`util/coinbase.go` isn't touched here), and the gate move actually *narrows* the divergence rather than widening it — the old mainnet-only `227835` constant ran no coinbase check on testnet 21111-227835 at all, and mainnet's gate now lands on svnode's exact `BIP34Height`. (Also the `OP_PUSHDATA1` example isn't a vector — Teranode reads `0x4c=76 > 8` and rejects it too.) Fold the byte-exact tightening into a separate issue.

Nits: `0x20000000` is now hardcoded in both mining-candidate paths (a shared constant would stop them drifting); the `bad-version` string drops svnode's comma + `(code N)` suffix (cosmetic, nothing parses it); the rewritten "version 2 height validation" subtest comment names a path the version floor now intercepts (no assertions there, so harmless).

The logic is right — green the CI, take ordishs' early-return, and file the byte-exact follow-up.

ctnguyen:
Thanks you for the careful review. Addressed points :

---

@ordishs's points

✅ **Blocking — coinbase-height extraction below BIP34:**
  `validateCoinbaseHeight` now early-returns before touching the coinbase when the block is below
  the network's BIP34 activation height, so pre-activation blocks are no longer parsed during IBD.
  This matches svnode, which only inspects the coinbase at/after `BIP34Height`
  (`ContextualCheckBlock`, `validation.cpp:6039`).

✅ **Non-blocking — magic number `0x20000000`:** It is now a single named
  constant, `miningCandidateVersion`, used by both the normal and empty-block mining-candidate
  paths.

✅ **Non-blocking — error-type shift on the quick-validation path:** deliberate and correct. The
  version floor runs before the coinbase-presence check to mirror svnode's header-first error
  priority (`ContextualCheckBlockHeader` runs ahead of the body/coinbase checks), so a below-floor
  block from a seeded peer resolves as *invalid* rather than *incomplete* — header version plus the
  settled height are sufficient proof, and the block should not be retried as merely incomplete.
  This shift is confined to the quick-validation path; the full `ValidateBlock` path is unchanged.

✅ **Non-blocking — byte-exact minimal-push prefix:** agreed this is a real, **pre-existing**
  accept-side gap that is independent of this PR (`util/coinbase.go` is not modified here). Tracking
  it as a separate follow-up rather than widening this PR's scope

---

@oskarszoon's points

✅ **CI formatting (Gate 1):** the empty-candidate comment was splitting the
  struct-literal alignment group; it now sits above the literal so gofmt/gci see one contiguous
  group.

✅ **Pre-BIP34 early-return (Gate 2):** Your read is right: the pre-fix cost was a cheap per-block
  extraction during a from-genesis IBD that mostly succeeded silently — warnings came only from a
  minority of coinbases (leading push byte > 8, or empty/too-short scripts), not a per-block
  warning storm. The comment no longer implies "~227k warnings".

✅ **Shared constant for `0x20000000`:** done — `miningCandidateVersion`.

✅ **`bad-version` string dropping svnode's comma + `(code N)`:** left as-is, intentionally. svnode's
  full log form (`"bad-version(0x%08x), rejected nVersion=0x%08x block (code N)"`) gets its comma
  and trailing `(code N)` from `FormatStateMessage` (`validation.cpp:642-647`); Teranode has no
  equivalent numeric `REJECT_*` code taxonomy, so a synthetic `(code N)` would be misleading rather
  than matching. The machine-relevant substrings (`bad-version(0x%08x)` and
  `rejected nVersion=0x%08x block`) already match byte-for-byte, and nothing parses the string.

✅ **Stale "version 2 height validation" subtest comment:** With the regtest
  params that `CreateBaseTestSettings` produces (BIP34 at 1e8), the v2 block at height 300000 is
  below BIP34, so the coinbase-height check is never reached; the comments now describe the subtest
  accurately as a `Block.Valid` code-path smoke case that clears the version floor.

ordishs (approved):
Re-reviewed the follow-up commits — everything requested has been addressed. Approving.

- **Pre-BIP34 extraction skip** (`0fc3f0c`): `validateCoinbaseHeight` now early-returns before touching the coinbase when `!postBIP34`, and the now-dead pre-BIP34 warning branch and redundant sub-conditions are gone. Matches svnode `ContextualCheckBlock`, which never inspects the coinbase below activation.
- **Genesis guard mirrored** (`3de05aa`): the added `currentHeight > 0` guard keeps genesis exempt on `BIP0034Height == 0` networks even though the caller's genesis branch already bypasses this function — good defence-in-depth, consistent with `Block.Valid`, and covered by the new `TestValidateCoinbaseHeight_GenesisExemptOnZeroBIP34`.
- **Named constant** (`a23db47`): `miningCandidateVersion` now carries the rationale once and is used by both candidate paths.
- **Stale test comment** (`991819a`): the v2 subtest comments now correctly state that height 300000 is below the regtest BIP34 sentinel, so the coinbase-height path isn't reached there.

Verified locally at the current head: all block-version tests pass across `model`, `stores/blockchain/sql`, `services/blockvalidation`, and `services/blockassembly`, including the new genesis-exemption test.

One item from the previous review remains as a suggested follow-up issue (not blocking): svnode's BIP34 check is a byte-exact prefix match against `CScript() << nHeight`, while `ExtractCoinbaseHeight` compares extracted values, so non-minimal height encodings pass here but are rejected by svnode. Pre-existing behaviour, worth tracking separately.

PR #1302, opened 2026-07-17
fix(blockvalidation): bounds-check same-block-parent vout in quick validation

### LINKED ISSUE #1283: Unguarded parent vout deref in blockvalidation quick_validate.go (index-out-of-range DoS)
## Summary

`services/blockvalidation/quick_validate.go` dereferences an attacker-controlled `PreviousTxOutIndex` into a parent transaction's outputs without a bounds check, in two places. A peer serving a below-checkpoint block whose child tx references a non-existent vout of a same-block parent panics the node with `index out of range` — a remotely-triggerable crash/DoS during catchup/IBD.

Pre-existing on both `main` and `release/v0.15`. Distinct from the validator-helper path fixed in #1243 / #1275 — that fix does **not** cover this path.

## The unguarded derefs

```go
// quick_validate.go:1104-1105 (processSubtreeBatch, via extendedTxsFromPrevBatches)
tx.Inputs[j].PreviousTxSatoshis = parentTx.Outputs[input.PreviousTxOutIndex].Satoshis
tx.Inputs[j].PreviousTxScript   = parentTx.Outputs[input.PreviousTxOutIndex].LockingScript

// quick_validate.go:1442-1443 (extendBatch, via extendedTxs)
tx.Inputs[j].PreviousTxSatoshis = parentTx.Outputs[input.PreviousTxOutIndex].Satoshis
tx.Inputs[j].PreviousTxScript   = parentTx.Outputs[input.PreviousTxOutIndex].LockingScript
```

`parentTx` comes from the same-block map; `input.PreviousTxOutIndex` comes from the (untrusted) child transaction.

## Reachability

quick-validate is the below-checkpoint fast path ("safe because checkpoints guarantee these blocks are valid"). That gating does **not** protect this deref: in all three processing variants the extend runs during batch processing and the merkle-root check (`validateSubtrees` → `block.CheckMerkleRoot`) runs *after*:

- `processBlockSubtreesSequential`: extend at :1104 → `validateSubtrees` at :451
- `processBlockSubtreesPipeline`: extend at :1442 → `validateSubtrees` at :590
- `processBlockSubtreesPipelineAsync`: extend at :1442 → `validateSubtrees` at :728

So a tampered/corrupt tx panics at the extend deref **before** the merkle check can reject it. `quickValidateBlock` processes peer catchup data (`peerID`, `baseURL`), so a malicious or corrupt peer can crash a syncing node.

## Fix

Bound the vout before indexing at both sites, mirroring the guard already on `main` from #1243 and now on `release/v0.15` via #1275:

```go
vout := input.PreviousTxOutIndex
if parentTx.Outputs == nil || int(vout) >= len(parentTx.Outputs) || parentTx.Outputs[vout] == nil {
    // return a clean processing error for this tx/block instead of indexing
}
```

Plus a regression test on the same-block-parent path (child input with vout past the parent's output count).

Related hardening (ChiR2 from #1275 review): the `int(vout)` cast is only width-safe on 64-bit. If airtight cross-width behaviour is wanted, compare as `uint64(vout) >= uint64(len(...))`. Since the goal is main/release parity, fix the comparison on `main` first and backport, rather than diverging.

## Sequencing

Fix on `main` first, then backport to `release/v0.15`, to keep the two trees in parity (avoid a release-ahead-of-main divergence).

Refs #1243 #1275

liam:
## Problem

Closes #1283.

The two same-block-parent fast paths in quick validation index `parentTx.Outputs[input.PreviousTxOutIndex]` with no bounds check, and `PreviousTxOutIndex` (vout) comes straight from an untrusted child transaction:

- `services/blockvalidation/quick_validate.go` `processSubtreeBatch` (via `extendedTxsFromPrevBatches`)
- `services/blockvalidation/quick_validate.go` `extendBatch` (via `extendedTxs`)

A peer-served below-checkpoint block whose child input references a non-existent vout of a same-block parent triggers `index out of range` and panics the node during catchup/IBD, before the post-processing merkle-root check can reject the block.

## Change

Guard both sites with a width-safe bounds check (`parentTx.Outputs == nil || int(vout) >= len(parentTx.Outputs)`) that returns a `ProcessingError`, converting the panic into a clean per-block validation failure. A valid checkpointed block never has an out-of-range vout, so no legitimate path changes behaviour. This mirrors the guard already in `services/validator/Validator.go` `extendTransaction` (from #1243 / #1275), which does not cover these two catchup-path sites. The `processSubtreeBatch` guard cancels the reader errgroup before returning so its context is not leaked.

## Testing

- New regression test `TestExtendBatch_SameBlockParentVoutBounds` exercises a non-extended child referencing an out-of-range vout of a same-block parent; it panics on the pre-fix code and returns a clean error with the fix.
- `go vet` clean; the broader quick-validate / batch test suite passes.

### DIFF
--- a/services/blockvalidation/quick_validate.go
+++ b/services/blockvalidation/quick_validate.go
@@ -1049,6 +1049,37 @@ func (b *SubtreeProcessingBatch) Close() {
 	}
 }
 
+// extendTxFromSameBlockParents extends tx's inputs whose parents are present in
+// parents (same-block parents already decoded in this batch), returning whether
+// any input still needs an external (store) lookup.
+//
+// PreviousTxOutIndex comes from the untrusted child tx, so the parent's output
+// count is bounds-checked before indexing: an out-of-range reference returns an
+// error instead of panicking the node with index-out-of-range (issue 1283). This
+// mirrors the guard in services/validator/Validator.go extendTransaction.
+func extendTxFromSameBlockParents(tx *bt.Tx, parents map[chainhash.Hash]*bt.Tx) (needsExternalLookup bool, err error) {
+	for j, input := range tx.Inputs {
+		parentHash := input.PreviousTxIDChainHash()
+
+		parentTx, ok := parents[*parentHash]
+		if !ok {
+			needsExternalLookup = true
+			continue
+		}
+
+		vout := input.PreviousTxOutIndex
+		if parentTx.Outputs == nil || int(vout) >= len(parentTx.Outputs) {
+			return false, errors.NewProcessingError("tx %s input %d references non-existent output %d of same-block parent %s",
+				tx.TxIDChainHash().String(), j, vout, parentHash.String())
+		}
+
+		tx.Inputs[j].PreviousTxSatoshis = parentTx.Outputs[vout].Satoshis
+		tx.Inputs[j].PreviousTxScript = parentTx.Outputs[vout].LockingScript
+	}
+
+	return needsExternalLookup, nil
+}
+
 // processSubtreeBatch reads and extends a batch of subtrees.
 // This is the shared first phase of both quick and normal validation.
 //
@@ -1146,27 +1177,12 @@ func (u *BlockValidation) processSubtreeBatch(
 
 			// Try to extend from same-block parents first
 			if !tx.IsExtended() {
-				needsExternalLookup := false
-				for j, input := range tx.Inputs {
-					parentHash := input.PreviousTxIDChainHash()
-					if parentTx, ok := extendedTxsFromPrevBatches[*parentHash]; ok {
-						// PreviousTxOutIndex comes from the (untrusted) child tx; bound it
-						// against the same-block parent's output count before indexing, else a
-						// crafted block panics the node with index-out-of-range (issue 1283).
-						// Mirrors the guard in services/validator/Validator.go extendTransaction.
-						vout := input.PreviousTxOutIndex
-						if parentTx.Outputs == nil || int(vout) >= len(parentTx.Outputs) {
-							cancelReaders()
-							return nil, errors.NewProcessingError("[processSubtreeBatch][%s] tx %s input %d references non-existent output %d of same-block parent %s",
-								block.Hash().String(), tx.TxIDChainHash().String(), j, vout, parentHash.String())
-						}
-
-						tx.Inputs[j].PreviousTxSatoshis = parentTx.Outputs[vout].Satoshis
-						tx.Inputs[j].PreviousTxScript = parentTx.Outputs[vout].LockingScript
-					} else {
-						needsExternalLookup = true
-					}
+				needsExternalLookup, extendErr := extendTxFromSameBlockParents(tx, extendedTxsFromPrevBatches)
+				if extendErr != nil {
+					cancelReaders()
+					return nil, errors.NewProcessingError("[processSubtreeBatch][%s] same-block parent extension failed", block.Hash().String(), extendErr)
 				}
+
 				if needsExternalLookup {
 					txsNeedingExtension = append(txsNeedingExtension, tx)
 				}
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):
Review at commit [0c744a9](https://github.com/liam/teranode/commit/0c744a9bea1aaecbbd89d179f798987a0b84b706)

Nice, focused hardening. Turning an untrusted `PreviousTxOutIndex` into a clean per-block `ProcessingError` instead of an `index out of range` panic during catchup/IBD is exactly the right call, and closing both fast paths (`processSubtreeBatch` and `extendBatch`) rather than just one is what makes this complete. Pulling the duplicated loop into `extendTxFromSameBlockParents` keeps the two sites in lock-step, the width-safe check (`int(vout) >= len(parentTx.Outputs)`) is correct, and cancelling the reader errgroup before returning on the `processSubtreeBatch` path avoids leaking its context. The error wrapping is right too — the trailing `extendErr` is captured as the wrapped error and its "non-existent output" message survives in the chain, which is what the tests assert. Test coverage is genuinely thorough: a focused unit test for the helper plus two end-to-end regressions that drive both the sequential and batch paths through the new branch. Just two small non-blocking notes below.

## Non-blocking issues

### ChiR1 - Doc comment points at the wrong mirror function
**Problem:** The helper doc comment says the bounds check "mirrors the guard in services/validator/Validator.go extendTransaction". `extendTransaction` ([Validator.go:1603](services/validator/Validator.go#L1603)) has no bounds check — it just delegates to `PreviousOutputsDecorate` and sets `extended`. The actual matching guard lives in `getUtxoBlockHeightAndExtendForParentTx` ([Validator.go:1188-1205](services/validator/Validator.go#L1188-L1205)), which does the same `int(vout) >= len(Outputs)` check.
**Why it matters:** A future reader who follows the reference to verify parity lands on a function that doesn't contain the guard, which undermines the point of the cross-reference.
**Fix:** Repoint the comment, e.g. `This mirrors the guard in services/validator/Validator.go getUtxoBlockHeightAndExtendForParentTx.`

### ChiR2 - Guard omits the per-output nil check the referenced guard has
**Problem:** The referenced guard in `getUtxoBlockHeightAndExtendForParentTx` also rejects a nil output entry (`... || txMeta.Tx.Outputs[vout] == nil`), whereas the new helper only checks `Outputs == nil || int(vout) >= len(parentTx.Outputs)` before dereferencing `parentTx.Outputs[vout].Satoshis`.
**Why it matters:** This is unreachable today — same-block parents are parsed from subtree data, so individual `Outputs[i]` are never nil — so it is purely a parity/defensiveness gap, not a live bug. But if the helper is ever reused with a differently-sourced parent map it would diverge from the guard it claims to mirror.
**Fix:** For exact parity, extend the condition:
```go
if parentTx.Outputs == nil || int(vout) >= len(parentTx.Outputs) || parentTx.Outputs[vout] == nil {
```

## Recap

| ChiR1 | Doc references wrong function | | 25% |
| ChiR2 | Missing per-output nil check | | 10% |

ordishs (approved):
Correct, minimal fix for the peer-triggerable panic in the two same-block-parent fast paths, and the shared helper removes real duplication while preserving the original extend semantics. Verified locally: the new regression tests pass under -race and vet is clean, with both call sites covered end-to-end (sequential and pipeline paths).

Two optional polish items, non-blocking:
- Mirror the validator guard fully by also checking `parentTx.Outputs[vout] == nil` before dereferencing.
- `uint64(vout) >= uint64(len(parentTx.Outputs))` would be unconditionally width-safe (the `int` cast is fine on our 64-bit targets, but the uint64 form costs nothing).

One pre-existing observation for a follow-up issue, not this PR: the error paths in processSubtreeBatch/extendBatch never Close() already-collected mmap-backed subtrees in the batch, so repeated rejected blocks leak mmaps until finalizers run.

PR #1303, opened 2026-07-17
fix(uaerospike): make overload-retry backoff cancellable on Close

### LINKED ISSUE #1270: uaerospike: make overload-retry loop cancellable on Close to bound shutdown stall
Follow-up from PR #1231 (drain-guarded `uaerospike.Client.Close`), non-blocking review point raised by @ordishs.

## Problem

`retryBatchOnOverload` / `retryOnOverload` (`util/uaerospike/overload_retry.go`) back off with a wall-clock `time.Sleep` loop bounded by `maxElapsed` (default 2m, enabled by default) and no cancellation. Because `BatchOperate` holds the drain `RLock` for its whole duration — including this retry loop — `Store.Close` → `util.CloseAerospikeClient` → `uaerospike.Client.Close` (which takes the exclusive `drainMu.Lock`) can block for up to the full retry budget (~2 min) if a batch op is stuck in overload backoff during shutdown.

On the normal shutdown path this is benign (batchers are drained before `CloseAerospikeClient`, so the drain is uncontended). It only bites if `Close` is reached with an in-flight overloaded batch op.

## Why it's not a trivial fix

The cancel signal must be raised **before** `Close` acquires the exclusive lock — otherwise `Close` is already blocked waiting on the very op it wants to interrupt. So this needs:

- a `closed chan struct{}` (guarded by `sync.Once`) on `Client`, closed by `Close` *before* taking `drainMu.Lock`;
- a `select { case <-time.After(wait): case <-c.closed: return ... }` replacing the bare `time.Sleep` in both retry loops;
- careful zero-value handling — tests construct `&Client{}` with a nil `closed` channel, and `close(nil)` panics.

## Acceptance

- `Close` interrupts an in-flight overload-retry backoff promptly instead of waiting out `maxElapsed`.
- Zero-value `&Client{}` (used in `client_drain_test.go`) stays safe.
- Regression test: a `BatchOperate` parked in overload backoff is unblocked by a concurrent `Close` within a small bound.

liam:
## Problem

Closes #1270.

The overload-retry loop (`retryOnOverload` / `retryBatchOnOverload`) backs off with a bare, uncancellable `time.Sleep(wait)` (`util/uaerospike/overload_retry.go`), and `BatchOperate` holds the drain read-lock for the whole loop (via `beginOp`/`endOp`). `Close()` takes the exclusive drain lock, so a `BatchOperate` parked in overload backoff makes `Close` block for up to the full retry budget (`maxElapsed`, ~2 min by default) before it can shut the client down. A plain context does not help, because the signal must fire *before* `Close` blocks on the lock.

## Change

- Add a `closed` channel (guarded by `sync.Once`) to `Client`. `Close()` signals it **before** acquiring the drain `Lock`.
- Both retry loops replace `time.Sleep(wait)` with a `select` over `time.After(wait)` and `<-c.closed`; on close they stop retrying and return the last overload error.
- The connection-permit `releasePermit`/`reacquirePermit` stay paired across the wait so the semaphore is never imbalanced on the abort path.
- A zero-value `Client` (never `Close()`d, e.g. some unit tests) leaves the channel nil, so the `select` falls through to the full wait, preserving prior behaviour.

## Testing

- New `TestRetryOnOverload_AbortsOnClose` (single-record and batch variants): a loop parked in backoff with a 2-minute budget returns in well under 5s once `Close` fires. Verified load-bearing (both subtests time out on the pre-fix `time.Sleep`) and race-clean.
- Full `util/uaerospike` package passes under `-race`; `go vet` clean.

### DIFF
--- a/util/uaerospike/client.go
+++ b/util/uaerospike/client.go
@@ -218,6 +218,14 @@ type Client struct {
 	// isn't warranted here; revisit if a scan is ever added to a shutdown-critical
 	// path.
 	drainMu sync.RWMutex
+
+	// closed is signalled by Close to abort in-flight overload-retry backoff so a
+	// shutdown is not stalled for up to the retry budget (maxElapsed) while a
+	// BatchOperate is parked in backoff holding the drain RLock. closeOnce guards
+	// the single close. A zero-value Client leaves closed nil, which simply means
+	// the backoff runs to completion (such clients are never Close()d).
+	closed    chan struct{}
+	closeOnce sync.Once
 }
 
 // beginOp registers an in-flight BatchOperate, blocking a concurrent Close from
@@ -252,6 +260,14 @@ func (c *Client) endOp() { c.drainMu.RUnlock() }
 //     the one being closed. If that path ever becomes reachable under load, close
 //     outside the mutex (snapshot+delete under lock, Close after unlock).
 func (c *Client) Close() {
+	// Signal any in-flight overload-retry backoff to abort BEFORE blocking on the
+	// drain lock. A BatchOperate parked in backoff holds the drain RLock, so
+	// without this signal Close would wait out the full retry budget (maxElapsed,
+	// ~2 min by default) before acquiring the exclusive Lock.
+	if c.closed != nil {
+		c.closeOnce.Do(func() { close(c.closed) })
+	}
+
 	c.drainMu.Lock()
 	defer c.drainMu.Unlock()
 
@@ -281,6 +297,7 @@ func NewClient(hostname string, port int, opts ...ClientOption) (*Client, error)
 		stats:         NewClientStats(),
 		overloadRetry: cfg.overloadRetry,
 		logger:        cfg.logger,
+		closed:        make(chan struct{}),
 	}, nil
 }
 
@@ -353,6 +370,7 @@ func NewClientWithPolicyAndHostOpts(policy *aerospike.ClientPolicy, hosts []*aer
 		stats:         NewClientStats(),
 		overloadRetry: cfg.overloadRetry,
 		logger:        cfg.logger,
+		closed:        make(chan struct{}),
 	}, nil
 }
 
--- a/util/uaerospike/overload_retry.go
+++ b/util/uaerospike/overload_retry.go
@@ -121,6 +121,19 @@ func isOverloadError(err aerospike.Error) bool {
 // each backoff sleep and re-acquired before the retry: the server isn't
 // seeing the op while we sleep, so holding the slot would only starve other
 // callers (see the loop body and reacquirePermit).
+// sleepOrAbort waits for d, or returns true early if the client is being closed.
+// A true return tells the overload-retry loop to stop retrying so shutdown is not
+// stalled for the remaining retry budget. A nil closed channel (zero-value Client
+// that is never Close()d) makes the select fall through to the full wait.
+func (c *Client) sleepOrAbort(d time.Duration) bool {
+	select {
+	case <-time.After(d):
+		return false
+	case <-c.closed:
+		return true
+	}
+}
+
 func (c *Client) retryOnOverload(do func() aerospike.Error) aerospike.Error {
 	err := do()
 	if err == nil || !isOverloadError(err) || !c.overloadRetry.enabled() {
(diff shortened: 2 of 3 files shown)


ctnguyen (approved):
Review at commit [8491410](https://github.com/liam/teranode/commit/84914102f4cbcdcbd5f4a96f0a029515e70942c0)

Well-targeted fix for a real shutdown stall. Signalling `closed` *before* taking the exclusive drain `Lock` in `Close()` is the key insight — a plain context wouldn't help because the abort has to fire before `Close` parks on the lock — and the ordering here gets that right. The permit accounting stays balanced on the new abort path (`releasePermit` → `sleepOrAbort` → `reacquirePermit` remain paired before the early return), the nil-channel fallthrough keeps zero-value clients behaving exactly as before, and `closeOnce` plus the `c.closed != nil` guard make the close safe and idempotent. No data race: `closed` is set once at construction and closing concurrently with a `select` receive is fine. The tests are load-bearing — they genuinely time out against the pre-fix `time.Sleep` — and cover both the single-record and batch loops. Two non-blocking notes below, both about the surrounding docs/logging rather than the fix itself.

## Non-blocking issues

### ChiR1 - New function detaches the `retryOnOverload` doc comment
**Problem:** `sleepOrAbort` (and its own doc comment) is inserted between the `retryOnOverload` doc comment and the `retryOnOverload` function, with no blank line separating the two comment blocks. Godoc treats consecutive comment lines with no gap as one comment, so it now attaches the whole block — the `retryOnOverload` prose plus the `sleepOrAbort` prose — to `sleepOrAbort`, and leaves `retryOnOverload` ([overload_retry.go:124](util/uaerospike/overload_retry.go#L124)) with no doc comment at all.
**Why it matters:** `go doc sleepOrAbort` now shows a paragraph describing a different function, and the carefully written `retryOnOverload` explanation is lost from its own godoc entry.
**Fix:** Move `sleepOrAbort` (with its comment) above the `retryOnOverload` doc block, or below the `retryOnOverload` function body, so each comment stays attached to the function it documents.

### ChiR2 - Give-up log is misleading on the Close abort path
**Problem:** On the abort path both loops call `logOverloadGiveUp`, which logs at `Errorf` level: `"aerospike still overloaded after %v (attempt %d), giving up"` with `%v` = `c.overloadRetry.maxElapsed` (e.g. `2m0s`) — see [overload_retry.go:350](util/uaerospike/overload_retry.go#L350). During a `Close`-triggered abort only milliseconds have actually elapsed.
**Why it matters:** A normal shutdown that happens to interrupt a backoff emits a spurious ERROR reading as if the full ~2-minute budget was spent, which is noisy and can mislead operators triaging shutdown logs.
**Fix:** Distinguish the abort case from the budget-exhausted case — e.g. log the abort at a lower level with different wording rather than reusing `logOverloadGiveUp`:
```go
if aborted {
    c.logger.Debugf("aerospike overload retry aborted on close (attempt %d): %v", attempt, err)
    return err
}
```

## Recap

| ChiR1 | Doc comment detached by insertion | | 30% |
| ChiR2 | Misleading give-up log on close | | 25% |

ordishs (approved):
Approving — the fix is well-designed: the signal-before-lock ordering in `Close()` is the only ordering that works here, the permit release/reacquire pairing stays balanced across the abort path, and the nil-channel fallback correctly preserves prior behaviour for zero-value clients. Regression tests are load-bearing (2-minute budget vs 10-second timeout) and pass under `-race`.

One change requested before merge (non-blocking):

**Doc comment misattached to `sleepOrAbort`** — in `util/uaerospike/overload_retry.go` the new `sleepOrAbort` function and its comment were inserted between `retryOnOverload`'s existing doc comment and the `retryOnOverload` function, so the two comment blocks are now one contiguous block attached to `sleepOrAbort`. Godoc shows `retryOnOverload`'s documentation on the wrong function and leaves `retryOnOverload` undocumented. Please move `sleepOrAbort` (with its own comment) above the `retryOnOverload` comment block or below the function.

Optional nit: the abort path calls `logOverloadGiveUp`, which logs "still overloaded after <maxElapsed> ... giving up" — misleading during a post-mortem since the loop aborted for shutdown, not budget exhaustion. A distinct "aborting overload retry: client closing" message would be clearer.

PR #1304, opened 2026-07-17
fix(blockvalidation): guard revalidateBlockChan sends against a stopped worker

### LINKED ISSUE #1151: [TN-012][medium] blockvalidation: unguarded channel sends can block worker (self-send) and stall shutdown
**Severity:** medium · **Confidence:** high · **Pass:** concurrency/liveness
**Component:** `services/blockvalidation/BlockValidation.go`

### Summary
Several internal buffered-channel sends are bare (no `select`/`ctx.Done()`). Notably the `setMinedChan` retry re-enqueue happens **inside the same worker goroutine that drains the channel** — if the 1000-buffer fills during a sustained `SetMinedMulti` failure, the worker blocks on its own send and stops draining (self-deadlock). The same pattern stalls graceful shutdown.

### Evidence
- Channel+worker: `BlockValidation.go:347` (`make(chan …, 1000)`), worker `:548-561`.
- Unguarded sends: `:461`, `:470`, `:647` (inside the worker's retry path), `:753`, `:764`, `:1395`.
- `revalidateBlockChan` (cap 2): `:692-695` spawns `go func(){ ch <- blockData }()` (leaks at shutdown); bare send `:1944`.

### Why it matters
A downstream stall (UTXO store slow) can fill `setMinedChan` and wedge its only consumer — marking-mined halts until restart. Unguarded sends also hang shutdown and leak goroutines.

### Suggested fix
Wrap every internal send in `select { case ch<-x: case <-ctx.Done(): return }`. For the worker self-retry, use a separate retry queue/timer instead of blocking back onto the same channel.

### Suggested tests
Fill `setMinedChan` with a stubbed always-failing setter → assert no deadlock and prompt ctx cancellation. Shutdown with a full `revalidateBlockChan` → assert no goroutine leak (goleak).

---

liam:
## Problem

Closes #1151 (remaining half).

`revalidateBlockChan` (cap 2) has a single consumer started in `start(ctx)`. Two producers sent to it unguarded, so once that consumer exits at shutdown they could block forever on a full channel:
- the retry re-enqueue goroutine in the worker loop (`BlockValidation.go` ~`:763`) leaked a goroutine, and
- `ReValidateBlock` (`~:2151`) blocked its caller.

The issue's **primary** finding — the `setMinedChan` self-send deadlock — was already fixed (non-blocking `enqueueSetMined` + `setMinedOverflow` set). This PR closes the remaining `revalidateBlockChan` half.

## Change

- Add a `revalidateWorkerStopped` channel, closed via `defer` when the sole consumer exits.
- The retry re-enqueue now `select`s on `ctx.Done()` (the worker's own context), so it can't leak at shutdown.
- `ReValidateBlock` prefers a non-blocking send and only falls back to the worker-stopped guard when the channel is genuinely full — so a running node enqueues exactly as before, and only a stopped-worker + full-channel path drops (with a warning) instead of blocking the caller.

No change to validation behaviour; this only affects shutdown/blocking semantics.

## Testing

- New `TestReValidateBlock_DoesNotBlockAfterWorkerStopped`: fills the channel, closes the worker signal, and asserts `ReValidateBlock` returns promptly. Verified load-bearing (blocks/times out on the pre-fix unguarded send).
- The non-blocking-preferred send is required to keep `TestBlockValidation_OptimisticFloaterRetriedDuringCatchup` (issue #1031 regression guard) green — it reads `revalidateBlockChan` directly with the worker disabled, so the enqueue must still succeed when the channel has space.
- `go vet` and `gci` clean; revalidation/floater suite passes.

### DIFF
--- a/services/blockvalidation/BlockValidation.go
+++ b/services/blockvalidation/BlockValidation.go
@@ -230,6 +230,11 @@ type BlockValidation struct {
 	// revalidateBlockChan receives blocks that need revalidation
 	revalidateBlockChan chan revalidateBlockData
 
+	// revalidateWorkerStopped is closed when the single revalidateBlockChan
+	// consumer (started in start) exits, so producers can select on it and avoid
+	// blocking forever on a full channel after the worker is gone at shutdown.
+	revalidateWorkerStopped chan struct{}
+
 	// stats tracks operational metrics for monitoring
 	stats *gocore.Stat
 
@@ -377,6 +382,7 @@ func NewBlockValidation(ctx context.Context, logger ulogger.Logger, tSettings *s
 		setMinedOverflow:              make(map[chainhash.Hash]struct{}),
 		setMinedOverflowSignal:        make(chan struct{}, 1),
 		revalidateBlockChan:           make(chan revalidateBlockData, 2),
+		revalidateWorkerStopped:       make(chan struct{}),
 		stats:                         gocore.NewStat("blockvalidation"),
 		mmapDir:                       tSettings.BlockValidation.SubtreeMmapDir,
 	}
@@ -726,6 +732,9 @@ func (u *BlockValidation) start(ctx context.Context) error {
 	u.logger.Infof("[BlockValidation:start] starting reValidation goroutine")
 
 	go func() {
+		// Signal producers (site 1's re-enqueue and ReValidateBlock) that this sole
+		// consumer is gone, so they don't block forever on a full channel at shutdown.
+		defer close(u.revalidateWorkerStopped)
 		defer u.logger.Infof("[BlockValidation:start] revalidateBlockChan worker stopped")
 
 		for {
@@ -755,7 +764,13 @@ func (u *BlockValidation) start(ctx context.Context) error {
 					if blockData.retries < 3 {
 						blockData.retries++
 						go func() {
-							u.revalidateBlockChan <- blockData
+							// Guard the re-enqueue: on shutdown the consumer (this
+							// worker) exits, so an unguarded send would leak this
+							// goroutine forever on the full channel.
+							select {
+							case u.revalidateBlockChan <- blockData:
+							case <-ctx.Done():
+							}
 						}()
 					} else {
 						u.logger.Errorf("[BlockValidation:start][%s] failed block revalidation, retries exhausted: %s", blockData.block.String(), err)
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):
Review at commit [726d5a1](https://github.com/liam/teranode/commit/726d5a14f59e0ffab928ca93cb002c83bcb32569)

Clean, minimal fix that closes the remaining half of the shutdown-deadlock issue. Both producers to `revalidateBlockChan` are now guarded: the retry re-enqueue goroutine selects on the worker's own `ctx.Done()`, and `ReValidateBlock` prefers a non-blocking send and only guards the blocking fallback with `revalidateWorkerStopped`. I verified those are the only two send sites, and that the worker's sole exit path really is `ctx.Done()` ([BlockValidation.go:733,747](services/blockvalidation/BlockValidation.go#L733)), so using `ctx.Done()` for the re-enqueue guard is equivalent to the worker-stopped signal for that path. The prefer-non-blocking-then-guard ordering nicely preserves the existing backpressure behaviour for a running node while eliminating the shutdown block — and it's what keeps the `#1031` floater-retry regression test green, which the PR description calls out. The regression test is load-bearing (fills the channel, closes the signal, asserts prompt return). One narrow non-blocking observation below.

## Non-blocking issue

### ChiR1 - Guard assumes the worker goroutine was actually launched
**Problem:** `revalidateWorkerStopped` is only ever closed by the `defer` inside the revalidation worker goroutine. `start()` has an early-return path — if `g.Wait()` fails at [BlockValidation.go:565-573](services/blockvalidation/BlockValidation.go#L565-L573) it returns before reaching the goroutine launch — on which the worker is never started and the channel is never closed. A later `ReValidateBlock` whose channel is already full would then block on the second `select` forever, since `revalidateWorkerStopped` never fires.
**Why it matters:** This is not a regression (the pre-fix unguarded send also blocked in that state) and it is practically unreachable (the service has failed to start, so `ReValidateBlock` is unlikely to be driven), but it means the guard's shutdown-safety guarantee doesn't cover the failed-start case. Worth being aware of, and it also leaves a slight asymmetry: site 1 guards on `ctx.Done()` while site 2 guards on `revalidateWorkerStopped`.
**Fix:** If you want the guard to hold on that path too, close `revalidateWorkerStopped` on the failed-start early return (alongside the existing `backgroundTasks.Done()`), or have `ReValidateBlock` also select on `ctx.Done()` so it can never block once the service context is cancelled:
```go
if err := g.Wait(); err != nil {
    u.backgroundTasks.Done()
    close(u.revalidateWorkerStopped) // keep the send-guard valid on the no-worker path
    return errors.NewServiceError(...)
}
```

ordishs (approved):
Fix is correct, minimal, and the new regression test is load-bearing — verified it (and the #1031 floater guard) passes with `-race` on the PR head. The two-step send in `ReValidateBlock` is the right shape: the plain `select {send, stopped}` alone would non-deterministically drop when the worker is stopped but buffer space exists, so the non-blocking-preferred fast path is justified, not just test plumbing.

A few non-blocking suggestions:

1. **`start()` early-return path never closes `revalidateWorkerStopped`.** If `g.Wait()` fails during old-block mined/subtree recovery, `start()` returns before launching the revalidation worker, so the guard channel never closes — `ReValidateBlock` on a full channel then blocks forever, the exact symptom this PR fixes. Not a regression (pre-fix code blocked there too), but worth closing the channel on that early return alongside the existing compensating `backgroundTasks.Done()`.
2. **Comment nit:** the comment above the worker's `defer close(...)` says "site 1's re-enqueue" — that label doesn't exist in the code, so a future reader can't resolve it. "the worker's retry re-enqueue below" would be self-contained.
3. **Drop counter:** the sibling setMined fix added `prometheusBlockValidationSetMinedDrops` when it started dropping work; this drop path only logs a `Warnf`. A matching counter would keep the two halves of #1151 symmetric and make an unexpected drop visible.

PR #1305, opened 2026-07-17
docs(blockassembly): document post-Genesis any-order safety at the tx queue

### LINKED ISSUE #1158: [TN-019][low] blockassembly: parent-before-child not guaranteed at queue producer (safe post-Genesis)
**Severity:** low · **Confidence:** high · **Pass:** consensus/concurrency
**Component:** `services/blockassembly/subtreeprocessor/queue.go`, `services/validator/Validator.go`

### Summary
Parent and child txs validated by concurrent `Store()` calls race on the MPSC queue's `tail.Swap`, so a child's batch can be enqueued (and placed in an earlier subtree slot) before its parent's; the dequeue loop doesn't hold a child back. Per BSV Genesis (TTOR removed at height 620,538; any in-block order valid if the parent is present and no double-spend), this yields a valid block — not a consensus defect on the current chain.

### Evidence
- `queue.go:59-77` MPSC `enqueueBatch` via `tail.Swap`; consumer `SubtreeProcessor.go:1021-1082` appends in dequeue order, no parent reorder.
- Producer: `Validator.go:845` (parent UTXO create) → `:887` (`sendToBlockAssembler`) — separate steps, so concurrent producers interleave.
- Mitigation `DoubleSpendWindow>0` holds batches back (`queue.go:96`) but defaults to 0.

### Why it matters
Only matters for blocks below the Genesis height or if combined with a path that drops the parent. Benign on the live chain, but the code relies implicitly on the Genesis rule without documenting it.

### Suggested fix
None required for consensus. Add a comment at the dequeue loop documenting the post-Genesis any-order assumption; add a parent-presence hold if strict ordering is ever needed.

### Suggested tests
Concurrency test enqueueing parent+child from two goroutines → assert the produced block is accepted by the validator.

---

liam:
## Summary

Closes #1158.

The lock-free ingest queue (`services/blockassembly/subtreeprocessor/queue.go`) is multi-producer: batch order is decided by whichever producer wins the `tail.Swap`, so parent-before-child ordering is not guaranteed at enqueue, and the single consumer appends nodes in dequeue order without topological reordering.

Per the triage on #1158 this is **benign on the live chain**: post-Genesis any within-block transaction order is valid as long as every parent is present and there is no double-spend, both of which the validator already enforces before a tx reaches this queue. A parent-presence hold would add hot-path cost with no benefit on a post-Genesis chain.

This PR just documents that invariant at `enqueueBatch` so the missing ordering guarantee reads as intentional rather than a latent bug. **No code change.**

### DIFF
--- a/services/blockassembly/subtreeprocessor/queue.go
+++ b/services/blockassembly/subtreeprocessor/queue.go
@@ -53,6 +53,15 @@ func (q *LockFreeQueue) length() int64 {
 // It uses atomic operations to ensure thread safety during concurrent enqueue operations.
 // The entire batch receives a single timestamp when enqueued.
 //
+// Ordering note (issue 1158): this is a multi-producer enqueue — batch order is
+// decided by whichever producer wins the tail.Swap below, so parent-before-child
+// ordering is NOT guaranteed at the producer, and the single consumer appends nodes
+// in dequeue order without topological reordering. This is safe on the live chain:
+// post-Genesis any within-block transaction order is valid as long as every parent is
+// present and there is no double-spend, both of which the validator already enforces
+// before a tx reaches this queue. A parent-presence hold would be a hot-path cost with
+// no benefit on a post-Genesis chain.
+//
 // Parameters:
 //   - nodes: The transaction nodes to add
 //   - txInpoints: Parent transaction references for each node


ctnguyen (approved):

liam:
Closing this — the premise is wrong.

This doc note asserted that the MPSC tx queue's lack of parent-before-child ordering is "safe post-Genesis" because BSV removed the topological transaction ordering (TTOR) requirement. That is incorrect. Verified against the SV reference node (`bitcoin-sv/bitcoin-sv@master`): `ConnectBlock` still enforces parent-before-child ordering as a consensus rule. `BlockValidateTxns` (src/validation.cpp) processes each block tx check-inputs-then-add-outputs in block order — `shard.HaveInputs(tx)` (:3328-3332) then `UpdateCoins` adds that tx's outputs (:3438-3445) — with no pre-pass that adds block-created outputs to the coins view. A same-block child at an earlier index than its parent is rejected with `bad-txns-inputs-missingorspent`. The only Genesis conditional in that loop (:3352) merely skips sigop counting; it does not relax ordering. `TxnGrouper` preserves block order within dependency groups (txn_grouper.cpp:28).

Teranode matches this: `model/Block.go` `checkParentTransactions` (:1352-1354) rejects a block with `BlockInvalidError` when a parent's index exceeds its child's, via `validOrderAndBlessed()`.

So documenting "any-order is safe" would enshrine a false invariant. The underlying concern in #1158 (can the queue producer race actually emit a child-before-parent ordering?) is being re-triaged on the issue — if reachable, it is a real invalid-block bug, not a benign one, and a doc comment is the wrong fix.

PR #1307, opened 2026-07-20
fix(p2p): wire invalid-blocks Kafka consumer exactly once

gokutheengineer:
**Closes bitcoin-sv/teranode#4687.**

### Problem
The invalid-blocks Kafka consumer was started twice. `Start()` started the injected consumer with `invalidBlockHandler`, whose real work (`ReportInvalidBlock`) was commented out - it only unmarshalled and returned nil. Then `startInvalidBlockConsumer` built a second consumer, overwrote `s.invalidBlocksKafkaConsumerClient`, and started it with the real `processInvalidBlockMessage`. The first consumer was leaked (never `Close()`d, unreachable after the overwrite) and ran a fully redundant consumer group (`p2p.<clientName>` vs `<InvalidBlocks>-consumer`), each independently consuming every message.

### Fix
Single consumer, owned by the injected client (the same pattern the invalid-subtrees consumer uses):
- `Start()` wires the injected consumer via a small `startInvalidBlocksConsumer` helper that starts it with the real `s.processInvalidBlockMessage`. The field is never reassigned, so `Stop()` closes the consumer that is actually running.
- Deleted the old `startInvalidBlockConsumer` (~100 lines duplicating consumer construction that `daemon_kafka.go` already handles via `InvalidBlocksConfig`) and the no-op `invalidBlockHandler` with its commented-out `ReportInvalidBlock` block.
- Handler cleanups: log prefixes aligned with the function name, `s.gCtx` instead of `context.Background()` so an in-flight `AddBanScore` is cancelled at shutdown, and the `ReasonInvalidBlock` constant instead of a string literal.

Error handling: the consumer starts with `kafka.WithRetryAndMoveOn(2, 2, time.Second)` - three attempts total with 2s/4s backoff between them - not `WithLogErrorAndMoveOn`. Move-on would commit the offset on error, silently dropping the ban on a transient registry failure while `processInvalidBlockMessage` deliberately keeps its `blockPeerMap` entry for retry. After the retries are exhausted the message is skipped so a poison message cannot stall the partition. Malformed messages return nil from the handler (log and skip) since retrying an unmarshal failure can never succeed. (The old default-path "retry via redelivery" was itself leaky under autocommit: a later message's `MarkCommitRecords` advances the committed offset past the failed record, so redelivery only happened if a restart/rebalance occurred first.)

Retry wrapper rework in `util/kafka` (review follow-ups): the copy-pasted retry loops in `wrapConsumerFn` are now a single `retryWithBackoff` helper (plus `messageKey` for the log lines), with these fixes:
- The backoff selects on `ctx.Done()` and aborts immediately on shutdown, returning the handler error so the in-flight message is not committed and is redelivered after restart. The wrapper is bound to the consumer's `internalCtx`, so both caller cancellation and `Close()` abort it. Cancellation is signalled explicitly (a `cancelled` return), not inferred from `ctx.Err()`, so exhaustion is never misclassified.
- No more pointless backoff sleep after the final attempt.
- `maxRetries` now means retries after the first attempt, and `maxRetries=0` performs exactly one attempt. This is what the executable spec always said: `test/longtest/util/kafka/kafka_test.go` asserts "1 + 3 retries" = 4 invocations for `WithRetryAndMoveOn(3, ...)` and "no retry, handler once + stopFn" for `WithRetryAndStop(0, ...)` - both impossible under the old loop, which ran `maxRetries` total attempts and never invoked the handler at all for 0. Those longtests need Docker and could not be run locally; they run in the dispatched long-tests workflow.
- The in-memory (`memory://`) consumer path now reuses `wrapConsumerFn` instead of its own divergent re-implementation in `inMemoryConsumerHandler.ConsumeClaim` (which had an uninterruptible `time.Sleep` backoff and a `withRetryAndStop` branch that never retried). Since dev/test run with `KAFKA_SCHEMA=memory`, this gives dev/test daemons the same cancellable retry semantics as production.

Unset config: the deleted fallback that built a consumer from `Kafka.Hosts` + `Kafka.InvalidBlocks` is intentionally not restored - no sibling consumer has one and the default config sets `kafka_invalidBlocksConfig`. Instead the unconfigured case logs an **error** (invalid-block banning is a DoS-prevention control per the setting's docstring). That log previously could not fire at all: the daemon passed a typed-nil `*kafka.KafkaConsumerGroup` into the interface parameter, so `!= nil` checks passed and `Start()`'s nil-receiver guard silently no-opped. The typed-nil guard is a shared `nonNilConsumerGroup` helper in `daemon_kafka.go`, applied to both optional consumers (invalid-blocks and invalid-subtrees). The rejected-tx consumer is mandatory (its getter errors when unset), so it needs no guard. Docs updated (`p2p_reference.md`, `kafka_settings.md` - the latter still described the deleted fallback and the old `{topic}-consumer` group).

### Tests
`services/p2p/invalid_blocks_consumer_test.go` (handler and wiring at 100% statement coverage):
- `TestStartInvalidBlocksConsumer_BansPeer`: calls the same `startInvalidBlocksConsumer` that `Start()` uses against the in-memory broker (which now exercises the real `wrapConsumerFn` wiring), produces a `KafkaInvalidBlockTopicMessage`, and asserts the sending peer receives a ban score and the `blockPeerMap` entry is removed.
- `TestStartInvalidBlocksConsumer_NilConsumerIsNoOp`, `TestProcessInvalidBlockMessage_UnknownBlockIsNotAnError`, `TestProcessInvalidBlockMessage_MalformedMessageIsSkipped`, `TestProcessInvalidBlockMessage_AddBanScoreErrorIsReturned`.

`util/kafka/kafka_consumer_test.go` (`messageKey`/`retryWithBackoff` at 100%):
- `TestWrapConsumerFnRetryBackoffIsCancellable`: cancellation mid-backoff aborts immediately, surfaces the handler error, no further attempts.
- `TestWrapConsumerFnRetryAndMoveOnExhaustsThenSkips` / `TestWrapConsumerFnRetryAndStopCallsStopFn`: exhaustion after 1+maxRetries attempts skips / fires stopFn.
- `TestWrapConsumerFnRetrySucceedsAfterFailure`, `TestWrapConsumerFnLogErrorAndMoveOnSwallowsError`, `TestWrapConsumerFnZeroMaxRetriesStillAttemptsOnce`.

`daemon/daemon_kafka_test.go`: `TestNonNilConsumerGroup` (typed nil maps to true nil).

```
ok  github.com/bsv-blockchain/teranode/services/p2p  14.097s
SETTINGS_CONTEXT=test go test -race -count=1 -skip 'TestFlushBytesRegression.*' ./util/kafka/...
ok  github.com/bsv-blockchain/teranode/util/kafka  17.294s
ok  github.com/bsv-blockchain/teranode/util/kafka/in_memory_kafka  2.568s
```
Skipped `TestFlushBytesRegression` tests require Docker (unavailable locally, untouched code paths). `./daemon` builds and vets clean; its test binary does not link locally (known GoBDK cgo issue), so the daemon test runs in CI.

Known pre-existing issues surfaced during review, left out of scope: `InMemoryConsumerGroup.Close()` waits on a WaitGroup nothing `Add`s to and leaks its `Consume` goroutine (test-only harness); the daemon's p2p wiring (`startP2PService`) has no direct test.

### DIFF
--- a/daemon/daemon_kafka.go
+++ b/daemon/daemon_kafka.go
@@ -186,6 +186,16 @@ func getKafkaTxmetaConsumerGroup(logger ulogger.Logger, settings *settings.Setti
 	return getKafkaConsumerGroup(logger, kafkaTxmetaConfig, consumerGroupID, true, &settings.Kafka)
 }
 
+// nonNilConsumerGroup returns the consumer group as an interface, mapping a
+// typed nil to a true nil so downstream nil checks behave correctly.
+func nonNilConsumerGroup(group *kafka.KafkaConsumerGroup) kafka.KafkaConsumerGroupI {
+	if group == nil {
+		return nil
+	}
+
+	return group
+}
+
 func getKafkaInvalidBlocksConsumerGroup(logger ulogger.Logger, settings *settings.Settings, consumerGroupID string) (*kafka.KafkaConsumerGroup, error) {
 	kafkaInvalidBlocksConfig := settings.Kafka.InvalidBlocksConfig
 	if kafkaInvalidBlocksConfig == nil {
--- a/daemon/daemon_kafka_test.go
+++ b/daemon/daemon_kafka_test.go
@@ -1,0 +1,18 @@
+package daemon
+
+import (
+	"testing"
+
+	"github.com/bsv-blockchain/teranode/util/kafka"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNonNilConsumerGroup(t *testing.T) {
+	// A typed nil must map to a true nil interface so downstream nil checks
+	// (e.g. p2p's startInvalidBlocksConsumer) behave correctly.
+	require.Nil(t, nonNilConsumerGroup(nil))
+
+	group := &kafka.KafkaConsumerGroup{}
+	require.NotNil(t, nonNilConsumerGroup(group))
+	require.Same(t, group, nonNilConsumerGroup(group))
+}
--- a/daemon/daemon_services.go
+++ b/daemon/daemon_services.go
@@ -315,22 +315,21 @@ func (d *Daemon) startP2PService(ctx context.Context, appSettings *settings.Sett
 		return err
 	}
 
+	// These consumers are optional (nil when unconfigured): map typed nils to
+	// true nils so the server's nil checks behave correctly.
 	invalidBlocksKafkaConsumerGroup, err := getKafkaInvalidBlocksConsumerGroup(createLogger("kpib"), appSettings, serviceNameP2P+"."+appSettings.ClientName)
 	if err != nil {
 		return err
 	}
 
-	// Assign only a non-nil consumer group to the interface: a typed nil would
-	// make the server's nil checks pass and silently disable consumption.
-	var invalidBlocksKafkaConsumerClient kafka.KafkaConsumerGroupI
-	if invalidBlocksKafkaConsumerGroup != nil {
-		invalidBlocksKafkaConsumerClient = invalidBlocksKafkaConsumerGroup
-	}
+	invalidBlocksKafkaConsumerClient := nonNilConsumerGroup(invalidBlocksKafkaConsumerGroup)
 
-	invalidSubtreeKafkaConsumerClient, err := getKafkaInvalidSubtreeConsumerGroup(createLogger("kpis"), appSettings, serviceNameP2P+"."+appSettings.ClientName)
+	invalidSubtreeKafkaConsumerGroup, err := getKafkaInvalidSubtreeConsumerGroup(createLogger("kpis"), appSettings, serviceNameP2P+"."+appSettings.ClientName)
 	if err != nil {
 		return err
 	}
+
+	invalidSubtreeKafkaConsumerClient := nonNilConsumerGroup(invalidSubtreeKafkaConsumerGroup)
 
 	// Create Kafka producers for subtrees and blocks
 	var subtreeKafkaProducerClient *kafka.KafkaAsyncProducer
--- a/daemon/daemon_services.go
+++ b/daemon/daemon_services.go
@@ -315,9 +315,16 @@ func (d *Daemon) startP2PService(ctx context.Context, appSettings *settings.Sett
 		return err
 	}
 
-	invalidBlocksKafkaConsumerClient, err := getKafkaInvalidBlocksConsumerGroup(createLogger("kpib"), appSettings, serviceNameP2P+"."+appSettings.ClientName)
+	invalidBlocksKafkaConsumerGroup, err := getKafkaInvalidBlocksConsumerGroup(createLogger("kpib"), appSettings, serviceNameP2P+"."+appSettings.ClientName)
 	if err != nil {
 		return err
+	}
+
+	// Assign only a non-nil consumer group to the interface: a typed nil would
+	// make the server's nil checks pass and silently disable consumption.
+	var invalidBlocksKafkaConsumerClient kafka.KafkaConsumerGroupI
+	if invalidBlocksKafkaConsumerGroup != nil {
+		invalidBlocksKafkaConsumerClient = invalidBlocksKafkaConsumerGroup
 	}
 
 	invalidSubtreeKafkaConsumerClient, err := getKafkaInvalidSubtreeConsumerGroup(createLogger("kpis"), appSettings, serviceNameP2P+"."+appSettings.ClientName)
--- a/docs/references/services/p2p_reference.md
+++ b/docs/references/services/p2p_reference.md
@@ -35,7 +35,6 @@ type Server struct {
     blockTopicName                    string
     subtreeTopicName                  string
     rejectedTxTopicName               string
-    invalidBlocksTopicName            string             // Kafka topic for invalid blocks
     invalidSubtreeTopicName           string             // Kafka topic for invalid subtrees
     nodeStatusTopicName               string             // pubsub topic for node status messages
     topicPrefix                       string             // Chain identifier prefix for topic validation
--- a/docs/references/services/p2p_reference.md
+++ b/docs/references/services/p2p_reference.md
@@ -480,7 +480,7 @@ Records bytes downloaded from a peer via HTTP (typically from their DataHub).
 - `handleSubtreeTopic`: Handles incoming subtree messages and processes subtree data.
 - `handleRejectedTxTopic`: Handles rejected transaction notifications from peers.
 - `handleNodeStatusTopic`: Handles incoming node status update messages.
-- `invalidBlockHandler`: Processes notifications about invalid blocks from Kafka.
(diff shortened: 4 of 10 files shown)


ctnguyen (approved):
Review at commit [96577f6](https://github.com/gokutheengineer/teranode/commit/96577f63b78a7ec6c08c0087e40a6715dbeda77b)

Nice catch on this one — the double-started invalid-blocks consumer was a genuine bug (a leaked consumer group plus every message processed twice), and consolidating onto the single injected client is exactly the right move. Deleting ~100 lines of hand-rolled consumer construction that `daemon_kafka.go` already provides, aligning with the invalid-subtrees pattern, and fixing up the stale log prefixes and docs all make this a clean, well-scoped diff. The test file is thorough: it exercises the real `Start()` wiring against the in-memory broker and covers the nil-consumer, unknown-block, and unmarshal-error branches. A couple of things around the error path are worth a second look before merge.

## Non-blocking issues

### ChiR1 - WithLogErrorAndMoveOn drops ban on transient error, contradicting the test
**Problem:** `startInvalidBlocksConsumer` starts the consumer with `kafka.WithLogErrorAndMoveOn()`, which wraps `processInvalidBlockMessage` so any returned error is logged and turned into `nil`. Because this consumer is `AutoCommitEnabled` (`getKafkaInvalidBlocksConsumerGroup` → `getKafkaConsumerGroup(..., true, ...)`), the record's offset is then marked committed regardless of whether `AddBanScore` actually succeeded — so a transient peer-registry failure means the ban is dropped and the message is never redelivered. The old dedicated consumer passed no error option, and the real consume loop leaves failed records uncommitted for redelivery, so this is a behavior change from "eventually ban" to "drop". `TestProcessInvalidBlockMessage_AddBanScoreErrorIsReturned` asserts the `blockPeerMap` entry is kept "so a retry can still ban the peer" — but it calls `processInvalidBlockMessage` directly, bypassing the `WithLogErrorAndMoveOn` wrapper, so that retry invariant is false for the wired path this same PR ships.
**Why it matters:** The banning path is anti-DoS defense-in-depth; silently dropping a ban attempt during a registry hiccup weakens it for that window, and a green test documents a retry guarantee the running code does not provide, giving future maintainers false confidence.
**Fix:** Pick one direction and align the test. If log-and-drop is intended (consistent with the sibling `invalidSubtreeHandler`/`rejectedTxHandler`, which swallow their own errors), have `processInvalidBlockMessage` swallow the `AddBanScore` error itself and rewrite the misleading test/comment. If retry-on-transient-failure is intended, use bounded retries and cover it end-to-end:
```go
s.invalidBlocksKafkaConsumerClient.Start(ctx, s.processInvalidBlockMessage,
    kafka.WithRetryAndMoveOn(maxRetries, backoffMult, time.Second))
```
and extend the wiring test to push a message through `Start()` with a failing registry, asserting the retry/redelivery behavior rather than the raw function's return value.

### ChiR2 - The "exactly once" property itself isn't tested
**Problem:** The PR title is "wire ... exactly once", but the new tests validate handler wiring and payload processing, not the property that regressed: the consumer being started twice. `Stop()` closing the single client is now correct by construction, yet nothing guards against the double-start reappearing.
**Why it matters:** Without a regression test for the exact defect fixed, a future refactor could reintroduce a second `Start` and the suite would stay green.
**Fix:** Add a small test that calls `s.startInvalidBlocksConsumer(ctx)` and asserts only one consumer subscribes (e.g. via the in-memory broker's `HasConsumer`/group membership), or that a second call is a safe no-op — or a short comment explaining why the single `Start()` call site makes this out of scope.

## Recap

| ChiR1 | Error swallowed, ban dropped | | 55% |
| ChiR2 | No exactly-once regression test | | 20% |

icellan (changes requested):
Two blocking issues.

**P0 — `WithLogErrorAndMoveOn` commits the offset on error, killing the retry this code depends on** (`services/p2p/server_helpers.go:624`)

The old consumer started with no error option, so a handler error took the default path (`kafka_consumer.go:575-582`): offset left uncommitted, Kafka redelivers on restart/rebalance, retry happens. The new `kafka.WithLogErrorAndMoveOn()` wrapper (`kafka_consumer.go:761-773`) logs the error and returns nil, so `MarkCommitRecords` commits the offset.

`processInvalidBlockMessage` deliberately keeps the `blockPeerMap` entry on error (the `Delete` at line 663 is only reached on success) so a redelivery can retry the ban. With move-on there is no redelivery: a transient `AddBanScore` failure means the entry TTL-expires via `cleanupPeerMaps` and the peer is never banned. `TestProcessInvalidBlockMessage_AddBanScoreErrorIsReturned` ("entry must be kept so a retry can still ban the peer") asserts a retry path that no longer exists in production.

Pick one: drop the option (restore default no-commit-on-error), use `WithRetryAndMoveOn`/`WithRetryAndStop`, or delete the entry on error and drop the retry contract — but don't keep the entry *and* commit the offset.

**P1 — settings fallback removed: consumer silently disabled when `kafka_invalidBlocksConfig` is unset** (`services/p2p/server_helpers.go:618-620`, `daemon/daemon_kafka.go:189-196`)

`getKafkaInvalidBlocksConsumerGroup` returns `nil, nil` when `InvalidBlocksConfig` is nil, and `startInvalidBlocksConsumer` now just returns on a nil consumer. The deleted `startInvalidBlockConsumer` had an else-branch building a consumer from `Kafka.Hosts` + `Kafka.InvalidBlocks` in that case, and errored `"no valid Kafka hosts found"` if hosts were missing rather than silently doing nothing. A node with `kafka_hosts` + `kafka_invalidBlocks` but no dedicated `kafka_invalidBlocksConfig` URL now consumes nothing — the ban path is off with no log or error. If dropping the fallback is intentional, at least warn on the nil consumer instead of returning silently.

Nit: startup log at line 623 names the `Kafka.InvalidBlocks` topic while the consumer subscribes to the topic in the `InvalidBlocksConfig` URL; if they differ the log misleads.

icellan (approved):
Re-reviewed at 1ad8bbd4b. All findings from the earlier review are addressed:

- AddBanScore now runs on `s.gCtx`, so it's cancelled on shutdown (and consistent with the existing call at Server.go:1945).
- The wasted trailing backoff is gone — `retryWithBackoff` breaks before the final sleep.
- The in-memory consumer shares `wrapConsumerFn`, so retry/skip/stop semantics (and cancellable backoff) match production; the duplicate loop and the handler's logger/options/topic fields are removed.
- The offset-commit-on-retry P0 and the silent-disable P1 were fixed in the prior push (WithRetryAndMoveOn + nonNilConsumerGroup + the nil-consumer error log).

The sustained-registry-outage ban loss is a deliberate, documented move-on tradeoff, fine.

One cosmetic leftover, non-blocking: the startup log at server_helpers.go:624 names `Kafka.InvalidBlocks` while the consumer actually subscribes to the topic in the `InvalidBlocksConfig` URL — log-only, only visible when those two settings disagree. Fix whenever.

Approving.

PR #1308, opened 2026-07-20
fix(p2p): make syncConnectionTimes cleanup reachable and bound the map

gokutheengineer:
**Closes #4686.**

### Problem
In `getNodeStatusMessage` (`services/p2p/Server.go`), the `else` branch that clears `syncConnectionTimes` was nested under a duplicated `if syncPeer != ""` condition, making it unreachable. Entries stored for each sync peer were never removed, so the map grew unbounded over the node's lifetime - one permanent entry per distinct sync peer.

### Fix
- Removed the duplicated inner condition so the `else` now clears the map whenever there is no sync peer.
- Additionally prunes entries for previous sync peers when a sync peer is set. This covers direct peer-to-peer rotation (A to B without an intermediate "no sync peer" state), which the unnesting alone would still leak on. The map now holds at most one entry.

### Tests
Added `TestGetNodeStatusMessage_SyncConnectionTimesBounded` in `services/p2p/server_helpers_test.go`, verifying the connection time is recorded for the current sync peer, stale entries are pruned on rotation, and the map is cleared when no sync peer is selected.

```
ok  github.com/bsv-blockchain/teranode/services/p2p
```

### DIFF
--- a/services/p2p/Server.go
+++ b/services/p2p/Server.go
@@ -1240,40 +1240,46 @@ func (s *Server) getNodeStatusMessage(ctx context.Context) *notificationMsg {
 	// Get current sync peer
 	syncPeer := s.getSyncPeer()
 	if syncPeer != "" {
-		if syncPeer != "" {
-			syncPeerID = syncPeer.String()
+		syncPeerID = syncPeer.String()
 
-			// Track when we first connected to this sync peer
-			if existingTime, ok := s.syncConnectionTimes.Load(syncPeerID); ok {
-				syncConnectedAt = existingTime.(int64)
-			} else {
-				// First time connecting to this sync peer
-				syncConnectedAt = time.Now().Unix()
-				s.syncConnectionTimes.Store(syncPeerID, syncConnectedAt)
-				s.logger.Debugf("[handleNodeStatusNotification] Recording sync connection time for peer %s: %d", syncPeerID, syncConnectedAt)
-			}
-
-			// Get sync peer's height and block hash
-			for _, peerInfo := range s.P2PClient.GetPeers() {
-				if peerInfo.ID == syncPeer.String() {
-					// Get the peer's best block hash from registry
-					if pInfo, exists := s.getPeer(syncPeer); exists {
-						if pInfo.BlockHash != nil {
-							syncPeerBlockHash = pInfo.BlockHash.String()
-						}
-
-						syncPeerHeight = pInfo.Height
-					}
-					break
-				}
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):
Review at commit [e53dd9e](https://github.com/gokutheengineer/teranode/commit/e53dd9e9e725316fbec6a3d554d39034f24650de)

This is a tidy fix for a real leak — the duplicated inner `if syncPeer != ""` had orphaned the `else`, so the "no sync peer, clear the map" branch was dead and `syncConnectionTimes` grew one permanent entry per distinct sync peer forever. Re-pairing the `else` with the outer `if` and additionally pruning stale peers on rotation genuinely bounds the map, and switching `interface{}` to `any` is a nice touch. The regression test covers the three cases that matter (record on selection, prune on rotation, clear on no-peer) and reads clearly. One minor concurrency note below; nothing blocking.

## Non-blocking issues

### ChiR1 - Prune runs on every call and races concurrent callers
**Problem:** `getNodeStatusMessage` is invoked from more than one goroutine — the 10s ticker in `publishNodeStatus` (`go s.publishNodeStatus(ctx)`), `handleBlockNotification`, and `sendInitialNodeStatuses` on each new websocket client. The new `Range`+`Delete` prune now runs on every call where a sync peer is set. If two calls observe different current sync peers during a rotation window, each one's prune can delete the entry the other just stored (G1 stores A, G2 stores B, G1 deletes non-A, G2 deletes non-B), transiently leaving the map empty and causing `syncConnectedAt` to be re-recorded as "now" on the next call without an actual reconnect.
**Why it matters:** This does not reintroduce the unbounded growth (the map only ever shrinks under the race, never grows), so the fix's core guarantee holds — but the "connected since" timestamp on `node_status` can spuriously reset during sync-peer flapping, a minor metrics-accuracy blip. The same is true for A→B→A rotation, where A's original connect time is discarded; that appears to be the intended memory-vs-continuity tradeoff and is worth confirming for any downstream consumer of `SyncConnectedAt`.
**Fix:** If timestamp precision under concurrent rotation matters, read the sync peer once and guard the touch of `syncConnectionTimes` with a small mutex for the duration of this block, e.g.:
```go
s.syncConnMu.Lock()
// Load/Store syncConnectedAt, then Range+Delete stale keys
s.syncConnMu.Unlock()
```
Otherwise a one-line comment noting the prune is best-effort under concurrent callers is enough.

icellan (approved):

PR #1309, opened 2026-07-21
fix(p2p): serialize compound sync decisions in SyncCoordinator

gokutheengineer:
### Problem
The sync coordinator's compound clear -> select -> activate sequences were not serialized as a unit. `sc.mu` guarded individual fields, and `claimSelectedSyncPeer` made the claim itself atomic, but the surrounding sequences still interleaved across goroutines: `HandleCatchupFailure`, `UpdateBanStatus`, `evaluateSyncPeer`, and the FSM-transition handler each read the current peer, performed registry/network work, then unconditionally cleared and re-triggered. A concurrent trigger could activate a new peer in between, and the stale clear would evict it - recording duplicate sync attempts and publishing conflicting sync triggers to Kafka. The disconnect handler's check-and-clear could likewise fire between a claim and its Kafka send.

### Fix
Added a dedicated `decisionMu` mutex that serializes every compound sync decision. Entry points (`TriggerSync`, `ClearSyncPeer`, `HandleCatchupFailure`, `HandlePeerDisconnected`, `UpdateBanStatus`, `checkFSMState`, `evaluateSyncPeer`) acquire it; internal paths that already hold it call a new `triggerSyncLocked` and `clearSyncPeerIfCurrent` directly. Lock ordering is strict: `decisionMu` before `sc.mu`, never the reverse, and `sc.mu` remains the field-level lock. The mutex was chosen over a single-consumer channel as the minimal change preserving the existing synchronous call structure. A side benefit: the previously-unlocked `backoffMultiplier` read in `considerReputationRecovery` is now serialized because all its writers run under `decisionMu`.

Residual (pre-existing, out of scope): `HandleCatchupFailure(reason)` does not identify the failed peer, so a late failure report can penalize a peer that legitimately replaced the failed one; fixing this requires widening the Server-to-coordinator interface.

### Tests
- `TestSyncCoordinator_HandleCatchupFailure_WaitsForInFlightDecision` - holds a decision in flight via a blocking `GetBestBlockHeader` and asserts `HandleCatchupFailure` cannot clear the sync peer mid-decision. Fails on the pre-fix code, passes with the fix.
- `TestSyncCoordinator_ConcurrentSyncDecisions_NoDeadlock` - hammers all decision entry points from 6 goroutines under `-race` to guard the lock ordering against deadlock regressions.

Commands run (affected package only):
```
```
All green, including the pre-existing suite.

### DIFF
--- a/services/p2p/sync_coordinator.go
+++ b/services/p2p/sync_coordinator.go
@@ -32,6 +32,14 @@ type SyncCoordinator struct {
 	// Per-RPC contexts are derived from this when needed.
 	ctx context.Context
 
+	// decisionMu serialises compound sync decisions (clear -> select -> activate)
+	// so only one decision is in flight at a time. Without it, concurrent triggers
+	// from the FSM monitor, periodic evaluation, catchup-failure and disconnect
+	// handlers can each select a peer and emit conflicting Kafka sync triggers, or
+	// clear a peer another path just activated. decisionMu is always acquired
+	// before mu and never while holding mu; mu remains the field-level lock.
+	decisionMu sync.Mutex
+
 	// Current sync state. currentSyncPeer holds the canonical libp2p ID string.
 	mu                         sync.RWMutex
 	currentSyncPeer            string
@@ -381,8 +389,13 @@ func (sc *SyncCoordinator) GetCurrentSyncPeer() string {
 	return sc.currentSyncPeer
 }
 
-// ClearSyncPeer clears the current sync peer
+// ClearSyncPeer clears the current sync peer. It serialises behind any
+// in-flight sync decision so an external clear cannot interleave with a
+// select-and-activate sequence; decision paths that already hold decisionMu
+// call clearSyncPeerIfCurrent directly.
 func (sc *SyncCoordinator) ClearSyncPeer() {
+	sc.decisionMu.Lock()
+	defer sc.decisionMu.Unlock()
 	sc.clearSyncPeerIfCurrent("")
 }
 
@@ -441,6 +454,7 @@ func (sc *SyncCoordinator) syncPeerNoProgressTimedOut(now time.Time) (string, ti
 	return currentPeer, progressAge, progressAge > sc.syncPeerNoProgressLimit()
 }
 
+// clearNoProgressSyncPeer requires decisionMu to be held by the caller.
 func (sc *SyncCoordinator) clearNoProgressSyncPeer(peerID string, progressAge time.Duration) bool {
 	sc.logger.Warnf("[SyncCoordinator] Sync peer %s made no validated progress for %v", peerID, progressAge.Round(time.Second))
 	if err := sc.registry.RecordSyncAttempt(sc.ctx, peerID); err != nil {
@@ -449,12 +463,20 @@ func (sc *SyncCoordinator) clearNoProgressSyncPeer(peerID string, progressAge ti
 	if !sc.clearSyncPeerIfCurrent(peerID) {
 		return false
 	}
-	_ = sc.TriggerSync()
+	_ = sc.triggerSyncLocked()
 	return true
 }
 
 // TriggerSync triggers a new sync operation
 func (sc *SyncCoordinator) TriggerSync() error {
+	sc.decisionMu.Lock()
+	defer sc.decisionMu.Unlock()
+	return sc.triggerSyncLocked()
+}
+
+// triggerSyncLocked runs the select-and-activate decision. It requires
+// decisionMu to be held by the caller.
+func (sc *SyncCoordinator) triggerSyncLocked() error {
 	sc.logger.Debugf("[SyncCoordinator] Sync triggered")
 
 	localHeight := sc.getLocalHeightSafe()
@@ -473,13 +495,14 @@ func (sc *SyncCoordinator) HandlePeerDisconnected(peerID peer.ID) {
 		sc.logger.Warnf("[SyncCoordinator] RemovePeer %s failed: %v", idStr, err)
 	}
 
-	sc.mu.RLock()
-	isSyncPeer := sc.currentSyncPeer == idStr
-	sc.mu.RUnlock()
+	// Check-and-clear under decisionMu so the clear cannot interleave with an
+	// in-flight activation of a different peer.
+	sc.decisionMu.Lock()
+	isSyncPeer := sc.clearSyncPeerIfCurrent(idStr)
+	sc.decisionMu.Unlock()
 
 	if isSyncPeer {
 		sc.logger.Infof("[SyncCoordinator] Sync peer %s disconnected", idStr)
-		sc.ClearSyncPeer()
 
 		// Trigger selection of new sync peer
 		go func() {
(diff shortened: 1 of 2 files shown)


ctnguyen (approved):
Review at commit [4422c49](https://github.com/gokutheengineer/teranode/commit/4422c49f4fa5a8a4d9e468c5e96eb0dcd65f87bf)

Nice, tightly-scoped fix. The root cause is stated precisely — field-level locking left the compound *clear → select → activate* sequences free to interleave — and `decisionMu` closes that without reshaping the call graph. The lock discipline holds up under scrutiny: ordering is strictly `decisionMu` before `sc.mu` and never the reverse, and every internal path that already holds `decisionMu` routes through `triggerSyncLocked`/`clearSyncPeerIfCurrent` rather than the re-entrant public wrappers, so there's no self-deadlock. Converting the disconnect/ban handlers to a check-*and*-clear is a real correctness gain over the old unconditional clear. The serialization test nails the core invariant; the `-race` stress test is a useful start but doesn't yet reach the hardest paths (see ChiR2). Two non-blocking notes below.

## Non-blocking issues

### ChiR1 — decisionMu held across blocking RPCs

**Problem:** Several entry points now hold `decisionMu` across multiple sequential blocking calls, not just the field mutation. `checkFSMState` holds it across `refreshProbeBudgetFromLocalTip` (`GetBestBlockHeader`), `GetFSMCurrentState`, `handleFSMTransition` (`getLocalTipWorkSafe` + `GetPeer` + possibly `triggerSyncLocked` → `listAllPeers`), `considerReputationRecovery` (`ReconsiderBadPeers`), and `handleRunningState` (`selectAndActivateNewPeer`). `evaluateSyncPeer` similarly holds it across `GetPeer`, `getLocalTipWorkSafe`, `selectNewSyncPeer` (another `ListPeers` + `GetBestBlockHeader`), and `sendSyncMessage`. Because all decision entry points contend on the same mutex, a slow RPC chain in the 30s periodic evaluation (or the FSM check) serializes the latency-sensitive reactions — `HandlePeerDisconnected`'s check-and-clear and `UpdateBanStatus` — behind it. A corollary: if the FSM monitor is parked in `checkFSMState` waiting on `decisionMu` held by an evaluation blocked on a hung RPC, `Stop()`'s `wg.Wait()` is delayed too (the RPCs on `sc.ctx` aren't cancelled by `Stop`).

**Why it matters:** This is inherent to the "serialize the whole decision as a unit" design and is a conscious tradeoff, so it's not a correctness defect — under healthy in-process/low-latency registry access the hold is negligible, and the Kafka publish is async. But once the registry is a remote gRPC service under load, decision throughput and disconnect/ban responsiveness are gated by the slowest RPC round-trip, and that coupling is easy to forget later.

**Fix:** Keep the genuinely-atomic part (select → claim → send-to-Kafka) inside the lock, and make the in-lock RPCs run on a bounded/cancellable context so `decisionMu` can't be pinned indefinitely (protecting both disconnect/ban handling and shutdown). Be careful hoisting the advisory work out, though: `considerReputationRecovery` both reads `backoffMultiplier` (unsynchronized — it relies on `decisionMu` today) and calls `resetBackoff`, and both must stay in the serialized decision path, so at most the `ReconsiderBadPeers` RPC call itself could be split out — and even that needs its cooldown computed under the lock first. At minimum, extend the `decisionMu` doc comment to record that it is intentionally held across blocking RPCs so the invariant is explicit for the next editor:

```go
// decisionMu ... mu remains the field-level lock.
// NOTE: decisionMu is deliberately held across blocking registry/blockchain
// RPCs for the duration of a decision; those calls must stay bounded so the
// mutex is never pinned (disconnect/ban handling and Stop() contend on it).
```

### ChiR2 — deadlock stress test misses two decision paths

**Problem:** `TestSyncCoordinator_ConcurrentSyncDecisions_NoDeadlock` claims to hammer "every decision entry point", but it doesn't. Its worker loop calls only `TriggerSync`, `evaluateSyncPeer`, `HandleCatchupFailure`, `UpdateBanStatus`, and `ClearSyncPeer` — it never calls `HandlePeerDisconnected` (which acquires `decisionMu` for its check-and-clear and then spawns a goroutine that re-enters via `TriggerSync`) or `checkFSMState`/`handleFSMTransition` (the deepest nested-lock path: `handleFSMTransition` → `clearNoProgressSyncPeer` → `triggerSyncLocked`, plus `considerReputationRecovery` and `handleRunningState`). Worse, the one ban path it does call is inert: `pid` is registered with `ReputationScore: 50` and no `IsBanned`, so `UpdateBanStatus` returns at `if !banned { return }` *before* ever taking `decisionMu`. So the two structurally-hardest paths to get the lock ordering wrong are exactly the ones the guard doesn't cover.

**Why it matters:** A lock-ordering regression introduced in `checkFSMState` or `HandlePeerDisconnected` — the paths most likely to grow future nested-lock logic — would sail past this test, which undercuts its stated purpose as a deadlock backstop. It's not a defect in the shipped fix (I traced those paths by hand and they're sound), only a gap in the safety net.

**Fix:** Either fold the two paths in *properly*, or narrow the test comment to the set actually exercised. Two traps to avoid when adding them: `HandlePeerDisconnected(mustNewPeerID(t))` with a random ID only hits the no-op branch — `clearSyncPeerIfCurrent` returns false when the ID isn't the current sync peer — so to run the clear plus the delayed-`TriggerSync` goroutine you must disconnect the *current* sync peer. And `UpdateBanStatus` reaches `decisionMu` only when the peer is genuinely banned, so ban it via the registry (`AddBanScore` past the threshold) rather than registering it unbanned. Roughly:

```go
// Ban pid so UpdateBanStatus(pid) actually reaches decisionMu.
reg.AddBanScore(pid.String(), "stress", banThresholdPoints)
// Let checkFSMState(ctx) run.
client.On("GetFSMCurrentState", mock.Anything).
    Return(ptrFSMState(blockchain_api.FSMStateType_RUNNING), nil)

// ... inside the worker loop, drive the *current* sync peer through disconnect:
sc.mu.Lock(); sc.currentSyncPeer = pid.String(); sc.mu.Unlock()
sc.HandlePeerDisconnected(pid) // now runs the clear + delayed TriggerSync goroutine
sc.checkFSMState(context.Background())
```

## Recap

| ChiR1 | decisionMu held across blocking RPCs | | 35% |
| ChiR2 | stress test misses two decision paths | | 30% |

freemans13 (approved):
Reviewed the full diff, audited the lock ordering by hand, and ran the affected tests under `-race`.

**The fix is sound.** Wrapping each compound decision in `decisionMu` while leaving `mu` as the field-level lock closes the real hole: the "read current peer → registry/network work → clear → reselect" sequences can no longer interleave, so a stale clear can't evict a freshly-activated peer and two conflicting Kafka triggers can't race out. The `*Locked` split is applied consistently.

Lock ordering holds throughout — no path takes `mu` then `decisionMu`, and no function holding `decisionMu` re-enters it, so neither a lock-order deadlock nor an `RWMutex` re-entry deadlock is reachable. The `select → claim → send-to-Kafka` path is now only ever reached under `decisionMu`.

The two new tests earn their place: the blocking-`GetBestBlockHeader` test proves `HandleCatchupFailure` waits for an in-flight decision instead of clearing mid-sequence (fails on the pre-fix code), and the 6-goroutine stress test is a good deadlock tripwire.

Verification: all 71 `TestSyncCoordinator` tests pass under `-race -tags testtxmetacache`; `go vet` clean.

One non-blocking note: `decisionMu` is now held across the blockchain gRPC calls, so a hung RPC would stall all sync decisions (including reacting to disconnects/bans). That's the intended design and the "waits for in-flight decision" test asserts it — just worth ensuring those calls stay context-bounded. The `HandleCatchupFailure` peer-identity residual you called out is correctly out of scope.

LGTM.

PR #1310, opened 2026-07-22
fix(p2p): bound sync-coordinator RPCs and wire Stop into server shutdown

gokutheengineer:
**Closes #4718.**

### Problem
Every sync-coordinator RPC passed the coordinator-lifetime `sc.ctx` (or the monitor-loop context) directly with no per-call timeout, so a slow or hung blockchain service - which hosts both the peer registry and the blockchain client endpoints - could wedge `monitorFSM`/`periodicEvaluation` indefinitely. Separately, `SyncCoordinator.Stop()` was never invoked by `Server.Stop()`, leaving the shutdown machinery dead code and coordinator goroutines undrained at shutdown. Both items confirmed still valid on main.

### Fix
- Added `boundedRPCContext()`, which derives a `context.WithTimeout` (5s default, `rpcTimeout` field) from a coordinator-owned cancelable context. All registry calls go through thin `registry*` wrappers built on it, so the timeout policy lives in one place.
- Beyond the issue's proposed scope, the blockchain-client calls made from the same loops (`GetBestBlockHeader` in `getLocalTipWorkSafe`, `GetFSMCurrentState` in `checkFSMState`) are bounded the same way - they hit the same service as the registry, so leaving them unbounded would keep the wedge alive.
- Per review (ChiR2): the local-height callback was the one remaining unbounded path - `Server.getLocalHeight` ran `GetBestBlockHeader(s.gCtx)` with no deadline, reached on every monitor tick and as the fallback when the bounded tip fetch fails. The callback now takes a `context.Context`: the coordinator feeds it `boundedRPCContext()` (timeout + abort on Stop), `Server.getLocalHeight(ctx)` additionally self-bounds with `defaultRPCTimeout`, and the two gossip-handler call sites pass their handler context instead of discarding it.
- Per review (ChiR1): `Server.Stop(ctx)` calls `s.syncCoordinator.Stop(...)` first, before tearing down the Kafka producers the coordinator publishes to. `Stop(ctx)` is `sync.Once`-guarded, cancels the coordinator context (so in-flight RPCs abort immediately), and bounds its `wg.Wait` with the caller's context: a goroutine parked in the non-context-aware `blocksKafkaProducerClient.Publish` cannot be released by cancellation, so on ctx expiry Stop logs and returns, leaking that goroutine exactly as pre-wiring shutdown did instead of hanging the node. Idempotent and safe before `Start`.
- Per follow-up review: the coordinator drain now gets a 10s sub-budget (`syncCoordinatorStopTimeout`) of the per-service stop context rather than the whole 30s budget, so a wedged coordinator goroutine cannot hand the DC11 producer flushes later in `Server.Stop` an already-expired context. The wg watcher goroutine is spawned once inside `stopOnce`, so repeated `Stop` calls share it instead of leaking one per call.
- Deliberate, now documented on `Stop`: registry updates issued by late libp2p callbacks (`UpdatePeerInfo`, `HandlePeerDisconnected`, `UpdateBanStatus`) in the window between Stop and `P2PClient.Close` fail against the cancelled coordinator context and degrade to a logged warning. They are best-effort peer bookkeeping with no value on a node that is shutting down, and waiting on them (they are not wg-tracked) would block the drain this method exists to bound.
- Verified no further unbounded call is reachable from the monitor loops: the peer selector's HTTP health check already carries a hard 2s client timeout (`util/health/http_health.go`).
- Rebased on the `decisionMu` serialization now on main; `checkFSMState` keeps holding `decisionMu` for the whole serialised decision, and `Stop` never touches `decisionMu`.
- Known pre-existing follow-up (unchanged here): making `blocksKafkaProducerClient.Publish` context-aware would let shutdown drain that path instead of leaking it.

### Tests
- `TestSyncCoordinator_RegistryCallsAreTimeBounded` - a blocking registry stub proves the per-RPC context carries a deadline and `listAllPeers` returns instead of hanging.
- `TestSyncCoordinator_LocalHeightCallbackIsTimeBounded` (ChiR2 regression) - parks the local-height callback against a hung RPC and asserts the monitor path (`isCaughtUp`) still returns, with a deadline on the callback context.
- `TestSyncCoordinator_StopUnblocksInFlightRPC` - parks a monitor goroutine inside a blocking `GetBestBlockHeader` (RPC timeout set to 1 minute) and asserts `Stop` aborts the call and drains the goroutines.
- `TestSyncCoordinator_StopHonorsContextDeadlineWhenGoroutineStuck` (ChiR1 regression) - simulates a goroutine stuck in a non-context-aware blocking call and asserts `Stop(ctx)` returns on its deadline, that a repeated timed-out `Stop` shares the single wg watcher, and that a later `Stop` drains fully.
- `TestSyncCoordinator_StopDrainsGoroutinesAndIsIdempotent`, `TestSyncCoordinator_StopBeforeStartReturns`.

Run (affected package only):
`SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/p2p/...` - all pass, including the concurrency-stress tests that came in with the decisionMu change.


ctnguyen (changes requested):
Review at commit [065dfc7](https://github.com/gokutheengineer/teranode/commit/065dfc7693b7662826aa3d940a4d66a0e26f1570)

Focused, well-tested fix: bounding every coordinator RPC through a single `boundedRPCContext` and finally wiring `Stop()` into `Server.Stop` are both the right moves, and the tests park a goroutine inside a blocking RPC to prove that cancellation — not the timeout — does the unblocking. One issue with the shutdown path below.

### Blocking issues

### ChiR1 — `Server.Stop` can hang past its deadline because `Stop()` drains unconditionally

**Problem:** This PR wires `s.syncCoordinator.Stop()` into `Server.Stop(ctx)` (`Server.go:1703`), and `Stop()` blocks on `wg.Wait()` unconditionally with no context (`sync_coordinator.go:464`). `ctxCancel()` only unblocks the gRPC paths; a monitor goroutine parked in `blocksKafkaProducerClient.Publish` (`sync_coordinator.go:1332`, a blocking channel send the PR itself notes is not context-aware) is not released. That path is reachable from both monitor goroutines — `monitorFSM → checkFSMState → handleRunningState → selectAndActivateNewPeer → sendSyncMessage → sendSyncTriggerToKafka`, and `periodicEvaluation → evaluateSyncPeer → sendSyncMessage`. So `Stop()`, and therefore `Server.Stop`, can block indefinitely and ignores the shutdown context it is handed.

**Why it matters:** Before this PR `Server.Stop` never waited on these goroutines, so a stuck `Publish` merely leaked a goroutine at exit. This PR turns that into a shutdown hang: if the blocks Kafka channel is full and not draining (broker unavailable/wedged at shutdown), graceful node shutdown blocks until a hard kill. That directly contradicts the PR's central claim of a bounded, deterministic drain, and `Server.Stop(ctx)` is documented as honoring its context for time-bound shutdown.

**Fix:** Make `Stop` context-aware so `Server.Stop`'s deadline actually bounds the drain, and add a regression test covering the blocked-`Publish` / `Stop(ctx)` deadline path. For example:

```go
func (sc *SyncCoordinator) Stop(ctx context.Context) {
    sc.stopOnce.Do(func() {
        close(sc.stopCh)
        sc.ctxCancel()
    })
    done := make(chan struct{})
    go func() { sc.wg.Wait(); close(done) }()
    select {
    case <-done:
    case <-ctx.Done():
        sc.logger.Warnf("[SyncCoordinator] Stop timed out before goroutines drained: %v", ctx.Err())
    }
}
```

and at the call site: `s.syncCoordinator.Stop(ctx)`. Softening the doc comment alone is not enough — the hang path needs to be bounded.

ctnguyen (changes requested):
Review at commit [49342c2](https://github.com/gokutheengineer/teranode/commit/49342c2d196184d23944a2e354af9fc6094eb87d)

The shutdown fix from last round is genuinely solid — `Stop` is now context-aware, cancels the coordinator context to abort in-flight RPCs, bounds `wg.Wait` against the shutdown `ctx`, and the service manager hands each service a real deadline, so ChiR1 is properly resolved with a dedicated regression test. One gap remains, though: not every blockchain RPC actually goes through `boundedRPCContext`. The local-height callback reaches an unbounded `GetBestBlockHeader` on the server-lifetime context, and it sits on the monitor hot path — so the wedge this PR set out to close is still reachable. Details below.

## Status of previous points

ChiR1 was raised at 065dfc7; verified against the current head 49342c2. `Stop(ctx)` matches the proposed fix (`sync_coordinator.go:470`), the call site passes the shutdown context (`Server.go:1704`), and `TestSyncCoordinator_StopHonorsContextDeadlineWhenGoroutineStuck` covers the blocked-`Publish` deadline path.

| ChiR1 | `Server.Stop` hangs past deadline; `Stop` drains unconditionally | YES | FIXED |

## New blocking issue

### ChiR2 — Monitor loops still reach an unbounded blockchain RPC via the local-height callback

**Problem:** `getLocalHeightSafe` invokes the `getLocalHeight` callback (`sync_coordinator.go:844`), which is wired to `Server.getLocalHeight` (`Server.go:442`). That method calls `s.blockchainClient.GetBestBlockHeader(s.gCtx)` (`server_helpers.go:574`) on the server-lifetime context `s.gCtx` (`Server.go:403`) — no per-call timeout, and not derived from `sc.ctx`, so `boundedRPCContext` does not cover it and `Stop`'s `ctxCancel()` cannot abort a call parked there. It is reached on every monitor tick (`isCaughtUp` → `getLocalHeightSafe`, `sync_coordinator.go:393`) and from `triggerSyncLocked`/`handleRunningState` (`sync_coordinator.go:586`, `830`). Worse, it is the fallback inside the *bounded* `getLocalTipWorkSafe` (`sync_coordinator.go:333`, `340`, `343`): when the bounded `GetBestBlockHeader` times out against a hung service, the error path calls `getLocalHeightSafe()`, which then fires a second, unbounded `GetBestBlockHeader(s.gCtx)` — defeating the very timeout this PR added.

**Why it matters:** This is exactly the wedge the PR set out to eliminate (issue #4718): a slow or hung blockchain service can still block `monitorFSM`/`periodicEvaluation` indefinitely, and a monitor goroutine parked here is not released by `Stop`'s cancellation. The PR deliberately widened scope to bound `GetBestBlockHeader`/`GetFSMCurrentState` "because they hit the same service" — this path hits the same service from the same loops but was missed, so the stated guarantee does not fully hold.

**Fix:** Bound the callback's RPC. Minimal change in `server_helpers.go`:

```go
func (s *Server) getLocalHeight() uint32 {
	if s.blockchainClient == nil {
		return 0
	}
	ctx, cancel := context.WithTimeout(s.gCtx, 5*time.Second)
	defer cancel()
	_, bhMeta, err := s.blockchainClient.GetBestBlockHeader(ctx)
	if err != nil || bhMeta == nil {
		return 0
	}
	return bhMeta.Height
}
```

To also get cancellation on `Stop`, route the height fetch through the coordinator's bounded context — e.g. give the callback a `context.Context` parameter fed from `boundedRPCContext`. Add a regression test that parks the callback's RPC against a hung client and asserts the monitor path still returns.

ctnguyen (approved):
Review at commit [414ef15](https://github.com/gokutheengineer/teranode/commit/414ef155345a6a59f857eb2747a899c84072294b)

This lands cleanly, and both of the previous points are properly addressed. `Stop(ctx)` is now `sync.Once`-guarded, cancels the coordinator context so in-flight RPCs abort immediately, and races `wg.Wait` against the shutdown deadline instead of blocking on it — exactly the shape ChiR1 asked for — and the local-height callback now takes a context fed from `boundedRPCContext`, closing the last unbounded path from ChiR2. Routing every registry/blockchain call through the thin `registry*` wrappers keeps the timeout policy in one place, and the accompanying regression tests (stuck-goroutine deadline, in-flight cancel, bounded callback) genuinely exercise the paths that used to hang. The remaining notes below are all non-blocking: they concern the fail-open behaviour that the new timeouts make reachable faster, and the way a single wedged component interacts with the shared shutdown budget. None of them block merge.

## Status of previous points

| ChiR1 | `Stop` drains unconditionally, ignores ctx | YES | FIXED |
| ChiR2 | Local-height callback unbounded RPC | YES | FIXED |

## Non-blocking issues

### ChiR3 - Fail-open height 0 rewrites gossiped peer tips in the shared registry
**Problem:** `Server.getLocalHeight` returns `0` on both error *and* timeout (`server_helpers.go:581-582`). Both gossip handlers feed that straight into `sanitizeAdvertisedTip(..., s.getLocalHeight(ctx))` (`server_helpers.go:87`, `Server.go:1030`), which computes `maxAcceptedHeight = localHeight + MaxUnvalidatedAdvertisedHeightLead` and, when the peer's advertised height exceeds it, **caps** the height and returns `ok = true` (`server_helpers.go:600-608`). The capped value is then persisted into the centralized registry via `addPeer` → `registerPeer` → `RegisterPeer` (`server_helpers.go:122`, `Server.go:1107`, `server_helpers.go:631`). With the default `p2p_max_unvalidated_advertised_height_lead` of `10000` (`settings/p2p_settings.go:39`), every peer gossiping while the blockchain service is slow gets its tip rewritten to `10000` with its real block hash. Note the fail-open-to-0 itself pre-dates this PR (a hard RPC error already produced it); what the PR adds is the *timeout* path — where the call previously blocked, it now returns `0` after 5s and reaches the cap.
**Why it matters:** The registry drives peer selection, so during a blockchain-service blip the coordinator sees peers as only slightly ahead (10000) rather than at their true tip, underestimating how far behind it is. It self-heals on the next successful gossip once the service recovers, so this is degraded data rather than a permanent wedge — but it is wrong data written to a shared store on the peer-selection path, and the new timeout makes it reachable without a hard error.
**Fix:** Have the height fetch report unknown-ness rather than collapsing it to `0`, and skip the cap / skip the registry update when the local height is unknown:
```go
func (s *Server) getLocalHeight(ctx context.Context) (uint32, bool) {
    // ... return (height, true) on success, (0, false) on error/timeout
}
// caller: if !ok { /* leave the advertised tip unsanitized-capped, or skip addPeer */ }
```

### ChiR4 - Coordinator drain can consume the whole shared `Server.Stop` deadline
**Problem:** `Server.Stop` threads one `ctx` through every teardown step, and `s.syncCoordinator.Stop(ctx)` runs **first** (`Server.go:1744`). In the documented wedged-`Publish` case, `Stop` blocks until that ctx expires (`sync_coordinator.go:488-492`). Everything after it then runs with an already-expired context: both `kafka.StopProducerCtx(ctx, ...)` calls (`Server.go:1771-1772`) hit their `ctx.Done()` branch immediately and skip the flush wait (defeating the DC11 intent noted just above them — see `util/kafka/kafka_producer_stop.go:37-44`), and `s.e.Shutdown(ctx)` (`Server.go:1775`) returns `ctx.Err()`, which `collect` records so `Server.Stop` reports a shutdown failure. In normal shutdown `sc.ctx` is already cancelled by the errgroup, so the drain is instant and this never triggers; it only bites in the pathological wedged-broker case — which is exactly the case the deadline exists for.
**Why it matters:** One stuck component starves unrelated teardown of its share of the shutdown budget, turning a single leaked coordinator goroutine into skipped producer flushes and a spurious `Server.Stop` error.
**Fix:** Give the coordinator drain its own sub-budget so the later steps keep a share, e.g.
```go
if s.syncCoordinator != nil {
    coordCtx, cancel := context.WithTimeout(ctx, coordDrainBudget) // a fraction of ctx
    s.syncCoordinator.Stop(coordCtx)
    cancel()
}
```

### ChiR5 - Per-RPC bound does not bound a monitor iteration; the tip fetch double-charges
**Problem:** `getLocalTipWorkSafe` (`sync_coordinator.go:337`) falls back to `getLocalHeightSafe()` on error / nil-meta / empty chainwork (`:346`, `:349`), and that fallback issues a *second* bounded `GetBestBlockHeader` against the same unhealthy service (`getLocalHeightSafe` → callback → `Server.getLocalHeight`, `sync_coordinator.go:852`). So a single tip fetch costs up to ~10s when the service is degraded. One `monitorFSM` tick chains `refreshProbeBudgetFromLocalTip` (`:721`) and `isCaughtUp` (a `getLocalHeightSafe` plus a `getLocalTipWorkSafe`, `:399-400`) — both unlocked — and then `checkFSMState` (`:722`), which takes `decisionMu` (`:741`) and holds it across its *own* refresh/FSM/activation work: a second `refreshProbeBudgetFromLocalTip`, `GetFSMCurrentState`, and `handleRunningState` → `selectAndActivateNewPeer`, each of which issues the same degraded RPCs. So while a full tick serializes many timeouts against the 2s `fastMonitorInterval`, the part under the lock is `checkFSMState`'s span — and for its duration `HandlePeerDisconnected` (`:610`), `HandleCatchupFailure` (`:632`), and `ClearSyncPeer` (`:507`) all stall behind it. This is still a strict improvement over the old hang-forever behaviour.
**Why it matters:** The RPCs are individually bounded but a decision is not, so a degraded blockchain service can hold `decisionMu` for tens of seconds and delay the gossip-driven peer-management paths, not just the monitor loop.
**Fix:** Skip the `getLocalHeightSafe` fallback when the tip fetch failed with a deadline/cancellation error (a second call to the same service will just time out again), and/or share one deadline across a whole decision rather than deriving a fresh `rpcTimeout` per RPC.

### ChiR6 - `defaultRPCTimeout` is hard-coded
**Problem:** `defaultRPCTimeout = 5 * time.Second` (`sync_coordinator.go:128`) is the only source for both the coordinator's `rpcTimeout` field (`:40`, `:101`, overridden only from tests) and `Server.getLocalHeight` (`server_helpers.go:577`). This PR introduces a new operator-visible failure mode — RPCs give up after 5s — but exposes no knob to tune the bound (the `rpcTimeout` field is set from tests only, never from config). Several other constants in this file are also hard-coded (`fastMonitorInterval`, `slowMonitorInterval`, etc.), but those are internal cadence choices; a deadline against an external service is the kind of value operators of a slow deployment realistically need to raise.
**Why it matters:** An operator on a deployment with a legitimately slower blockchain service cannot raise the bound without a rebuild, and 5s is aggressive relative to some gRPC round-trips under load.
**Fix:** Source it from a `settings.P2P` key (e.g. `p2p_sync_coordinator_rpc_timeout`) with a 5s default, keeping the current behaviour unchanged out of the box.

## Recap

| ChiR3 | Fail-open height pollutes registry |  | 45% |
| ChiR4 | Drain eats shared stop deadline |  | 30% |
| ChiR5 | Tip fetch double-charges under load |  | 25% |
| ChiR6 | RPC timeout not configurable |  | 10% |

icellan:
Solid shutdown hardening. Two things worth a look before merge:

- `Server.go:1704` — `syncCoordinator.Stop(ctx)` shares the single 30s per-service stop ctx with the later Kafka producer flushes. A coordinator goroutine stuck in a non-cancellable `Publish` can burn the whole budget, so the DC11 async-producer `Stop()` calls below get an already-expired ctx and skip their final flush. Consider a sub-budget for the coordinator, or ordering the producer flush first.
- `sync_coordinator.go:253` — `UpdatePeerInfo`/`HandlePeerDisconnected`/`UpdateBanStatus` derive their RPC ctx from `sc.ctx`, which is cancelled before `P2PClient.Close`. Late libp2p callbacks in that window drop their registry update (Canceled → `Warnf` → continue), and they're not tracked by `sc.wg` so nothing waits. Fine if intentional, but worth confirming the drop is acceptable during drain.

Nits (`sync_coordinator.go:476` watcher-goroutine leak on repeated `Stop`, `:746` inline `cancel()` instead of `defer`) are minor.

icellan (approved):
Re-reviewed after the amend. Both earlier P1s addressed:

- The shared 30s stop budget is now a dedicated 10s coordinator sub-budget (`syncCoordinatorStopTimeout`, `Server.go:1733`), so the Kafka producer flushes below no longer share it.
- Inline `cancel()` in `checkFSMState` now uses the `defer` wrapper, and the repeated-Stop watcher multiplication is gone (spawn moved inside `stopOnce`).

One thing left, non-blocking: `boundedRPCContext()` still derives from `sc.ctx`, which `Stop` cancels before `P2PClient.Close`. A libp2p `HandlePeerDisconnected`/`UpdatePeerInfo`/`UpdateBanStatus` callback landing in that window drops its registry update (Canceled → `Warnf` → continue), and those handlers aren't in `sc.wg`. Fine if that's the intended drain behaviour — just confirm it's acceptable.

Minor for later: `checkFSMState` holds `decisionMu` across the sequential bounded RPCs, so a fully-hung blockchain service stalls the decision path for the sum of the 5s timeouts, not 5s; and `refreshProbeBudgetFromLocalTip` + `checkFSMState` do two `GetBestBlockHeader` round-trips per non-caught-up tick (pre-existing).

Approving.

PR #1321, opened 2026-07-22
test(blockassembly): spec conformance tests for BA-SOLUTION and BA-CA…

sugh01:
…NDIDATE

Add negative-path conformance tests for the Mining Solution Contract (BA-SOLUTION-NNN) and Mining Candidate Contract (BA-CANDIDATE-NNN) in services/blockassembly/spec.md.

Ten tests pin conformant rejection paths: NotFound for unknown/expired or malformed candidate IDs (BA-SOLUTION-004, BA-CANDIDATE-011), the exact linear-extension stale message (BA-SOLUTION-005), coinbase request-validation (exactly one input, scriptSig length, parseability), whole-JobStore clear on successful submission (BA-CANDIDATE-005), invalid-block reset with unchanged chain (BA-SOLUTION-008), and stale-candidate survival across tip change (BA-CANDIDATE-006).

Two tests assert spec-correct behaviour and currently FAIL, exposing defects:

- BA-SOLUTION-003/007: a coinbase TX-store persistence failure is only logged; the block is still announced to the blockchain and the call returns Ok. No InvalidateBlock exists on the post-submission failure path, so a block can reach the chain with its coinbase never persisted.

- BA-STATE-001: GetMiningCandidate gates only on the blockchain service FSM, never on the block-assembly unminedTransactionsLoading flag. During startup recovery it serves a candidate built from a partially-recovered pipeline where the state matrix requires "reject (not ready)".

### DIFF
--- a/services/blockassembly/submit_mining_solution_spec_test.go
+++ b/services/blockassembly/submit_mining_solution_spec_test.go
@@ -1,0 +1,588 @@
+package blockassembly
+
+// Negative-path conformance tests for the Mining Solution Contract
+// (BA-SOLUTION-NNN) defined in services/blockassembly/spec.md.
+//
+// These tests deliberately exercise the *rejection* and *failure* paths of
+// SubmitMiningSolution / submitMiningSolution, driving the service against the
+// letter of the spec to surface divergences. Where a test pins behaviour that
+// currently diverges from the spec, the divergence is called out inline with a
+// `SPEC DIVERGENCE` comment so it is discoverable.
+//
+// Contract references:
+//   BA-SOLUTION-003  ordered persist(coinbase)->submit(block), all-or-invalidate
+//   BA-SOLUTION-004  unknown/expired candidate -> NotFound
+//   BA-SOLUTION-005  linear-extension stale -> exact stale message
+//   BA-SOLUTION-007  post-submission step failure -> invalidate block
+//   BA-SOLUTION-008  invalid block -> job removed + full Reset, chain unchanged
+//   BA-CANDIDATE-005 successful submission clears ALL candidates
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/bsv-blockchain/go-bt/v2"
+	"github.com/bsv-blockchain/go-bt/v2/bscript"
+	"github.com/bsv-blockchain/go-bt/v2/chainhash"
+	subtreepkg "github.com/bsv-blockchain/go-subtree"
+	"github.com/bsv-blockchain/teranode/errors"
+	"github.com/bsv-blockchain/teranode/model"
+	"github.com/bsv-blockchain/teranode/pkg/fileformat"
+	"github.com/bsv-blockchain/teranode/services/blockassembly/blockassembly_api"
+	"github.com/bsv-blockchain/teranode/services/blockassembly/mining"
+	"github.com/bsv-blockchain/teranode/services/blockassembly/subtreeprocessor"
+	"github.com/bsv-blockchain/teranode/stores/blob"
+	"github.com/bsv-blockchain/teranode/stores/blob/memory"
+	"github.com/bsv-blockchain/teranode/stores/blob/options"
+	"github.com/bsv-blockchain/teranode/util/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+// ---------------------------------------------------------------------------
+// helpers
+// ---------------------------------------------------------------------------
+
+// craftJob registers a synthetic job with the given previous-block hash into the
+// jobStore and returns its 32-byte candidate ID. It bypasses GetMiningCandidate
+// so the negative paths downstream of the jobStore lookup can be driven
+// deterministically without a completed subtree pipeline.
+func craftJob(t *testing.T, s *BlockAssembly, prevHash chainhash.Hash) []byte {
+	t.Helper()
+
+	id := chainhash.HashH(fmt.Appendf(nil, "crafted-job-%s", t.Name()))
+
+	nbits := model.NBit{0xff, 0xff, 0x7f, 0x20} // regtest max target
+	s.jobStore.Set(id, &subtreeprocessor.Job{
+		ID:       &id,
+		Subtrees: nil, // empty: coinbase-only candidate, enough for the rejection paths
+		MiningCandidate: &model.MiningCandidate{
+			Id:           id[:],
+			PreviousHash: prevHash[:],
+			Version:      1,
+			NBits:        nbits[:],
+			Height:       124,
+		},
+	}, jobTTL)
+
+	return id[:]
+}
+
+// coinbaseWithInputs builds a parseable transaction with the requested number of
+// inputs, each carrying an unlocking script of scriptSigLen bytes. It is used to
+// exercise the coinbase request-validation branches of submitMiningSolution.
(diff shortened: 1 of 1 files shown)


ctnguyen (changes requested):
Review at commit [0daf656](https://github.com/sugh01/teranode/commit/0daf65625d3961aa23ecf84360127beab2bf1081)

Nice, focused negative-path suite. The thing I like most is that it drives the *real* code paths: real in-memory stores (`sqlitememory` blockchain/UTXO, `memory` blob), a real started assembler, and Kafka-free wiring — so it respects the project convention of not mocking the blockchain client or the pipeline. The two-tier helper design is sound too: `craftJob` injects a synthetic job to hit the fast rejection branches deterministically, while `mineRealCandidate` covers the end-to-end submission path. The inline `SPEC DIVERGENCE` comments anchored to contract IDs are genuinely useful, and pinning the exact stale message (BA-SOLUTION-005) is the right kind of assertion. The rejection-path coverage (unknown/malformed candidate, coinbase input/scriptSig/parse validation, whole-JobStore clear on success) is thorough and reads cleanly. My concerns are a build-level one about how the intentionally-failing tests land, plus two spots where an assertion claims more coverage than it actually delivers.

## Blocking issues

### ChiR1 — Two conformance tests are committed in a permanently-failing state with no skip/xfail guard
**Problem:** `TestBASolution003_CoinbaseStoreFailure_AllOrInvalidate` and `TestBAState001_GetMiningCandidateDuringLoading_MustReject` assert spec-correct behaviour the implementation does not yet provide, so they fail against `0daf656`. `submitMiningSolution` only logs a coinbase `txStore.Set` failure and proceeds to `AddBlock`/returns `Ok` (`Server.go:1596-1601`), so `require.Error` fails; `GetMiningCandidate` gates solely on the FSM and never reads `unminedTransactionsLoading` (`Server.go:1286-1293`), so its `require.Error` fails too. Neither test is `t.Skip`-ped or behind a build tag.
**Why it matters:** `go test ./services/blockassembly/...` will be red on `main` from the moment this merges. That breaks CI for the whole package, and a permanently-failing suite trains everyone to ignore red — which then masks real regressions. It also isn't a documented conformance gap: neither divergence appears in `spec.md`'s Conformance Gaps section (which lists only Idle state, config validation, anti-creep), so there's no tracked "expected fail" for these.
**Fix:** Keep the assertions (they encode the correct target behaviour) but skip until the implementation is fixed, referencing the contract, so intent is preserved and CI stays green. For example:
```go
func TestBASolution003_CoinbaseStoreFailure_AllOrInvalidate(t *testing.T) {
	t.Skip("KNOWN FAIL — BA-SOLUTION-003/007: coinbase persist failure is logged, not invalidated; " +
		"un-skip when Server.go submitMiningSolution aborts/invalidates on txStore.Set error")
	// ...
}
```
Alternatively gate both behind a `//go:build conformance_pending` tag run in a separate non-gating lane. Either way, also add these two divergences to the spec's Conformance Gaps list so they are tracked rather than living only in a test comment.

### ChiR2 — `TestBASolution008` does not actually verify the Reset it claims to pin
**Problem:** The test name and comment claim it pins AC-BA-SOLUTION-008.1's requirement that an invalid block "MUST drive the assembler through Reset back to Running," but the only reset-related assertion is:
```go
	return s.blockAssembler.GetCurrentRunningState() == StateRunning
}, 5*time.Second, 50*time.Millisecond, "assembler must return to Running after invalid-block reset")
```
`newRunningServer` already drove the assembler to `StateRunning`, and `StateRunning` is the steady state, so this predicate is true regardless of whether `Reset` ran. A hypothetical implementation that rejected the invalid block *without* calling `Reset` (`Server.go:1584`) would leave the state at `StateRunning` and still pass. The height-unchanged and `NotFound` checks do cover "chain unchanged" and "candidate deleted," but the Reset itself is unverified.
**Why it matters:** For a test-only PR the assertions *are* the deliverable. A vacuous check that advertises coverage of a MUST requirement is worse than no check — it signals the reset path is protected when a regression there would sail through green.
**Fix:** Assert something only a full Reset mutates. For example, seed extra transactions into the pipeline before the invalid submission and assert the post-reset pipeline state reflects a rebuild (e.g. capture `s.blockAssembler.TxCount()` / subtree-processor state before and assert the reset-driven change afterward), rather than polling the steady-state string. If no clean reset-specific observable is reachable from the public surface, drop "Resets" from the name/comment so the test stops claiming coverage it does not provide.

## Non-blocking issues

### ChiR3 — `TestBASolution003` asserts the error but not the all-or-invalidate *effect*
**Problem:** With the forced `txStore.Set` failure, the test asserts only `require.Error` and `resp.Ok == false`. BA-SOLUTION-003/007 require that the block does not remain on the chain (abort before submit, or invalidate after). Nothing here checks chain state, so a future implementation that returns an error while still leaving the block accepted would pass this test yet violate the contract.
**Why it matters:** Once the implementation is fixed to fail the submission, the test should lock in the *observable* guarantee (no block on the chain), not just that some error came back — otherwise the fix could regress silently into "errors loudly but still commits the block."
**Fix:** Add a chain-unchanged / no-submission assertion around the failure, mirroring the height check in the 008 test — e.g. capture the best-block header (or height) before the submission and assert it is unchanged afterward, being mindful that tip updates propagate asynchronously (poll or query the blockchain client's best block header directly).

### ChiR4 — `TestBACandidate006` largely duplicates `TestBASolution005`
**Problem:** `TestBACandidate006_ChainTipChangeDoesNotWipeCandidates` crafts a job whose parent matches the tip's parent and asserts the stale error (not `NotFound`) — the same arrange/act/assert as `TestBASolution005_LinearExtensionStale`. The comment itself concedes the behavioural consequence "is covered by `TestBASolution005_LinearExtensionStale`."
**Why it matters:** Two tests exercising the same branch cost maintenance without adding signal, and the 006 name promises something stronger (that tip-change notifications don't wipe the JobStore) than a single stale lookup actually verifies.
**Fix:** Either fold 006 into 005 with a comment noting both contract IDs, or make 006 genuinely test its stated claim — e.g. issue a real candidate, drive a chain-tip advance through the notification path, then assert the candidate is still *recognized* (stale error) rather than `NotFound`. As written it only re-checks the lazy stale comparison.

### ChiR5 — Malformed-ID tests assert only "some error", not the error kind
**Problem:** `TestBASolution004_MalformedCandidateID` and the malformed subtest of `TestBACandidate011` assert just `require.Error` / `require.Nil(resp)`. The current failure comes from `chainhash.NewHash` returning a raw error (`Server.go:1429-1432`), which is neither typed as invalid-argument nor `NotFound`.
**Why it matters:** A bare `require.Error` passes for *any* error, so it does not pin the request-validation contract ("Candidate ID is 32 bytes"); a change that failed with an unintended error kind would still look green.
**Fix:** Pin the error kind, e.g. `require.True(t, errors.Is(err, errors.ErrInvalidArgument))` (or whatever classification is intended), matching the specificity already used in the `NotFound` tests.

## Recap

|---|---|---|---|
| ChiR1 | Failing tests, no skip/xfail | YES | 85% |
| ChiR2 | 008 reset assertion vacuous | YES | 78% |
| ChiR3 | 003 no chain-unchanged check | | 45% |
| ChiR4 | 006 duplicates 005 | | 35% |
| ChiR5 | Weak malformed-ID assertions | | 30% |

PR #1322, opened 2026-07-22
MvP-4691 blockassembly: reject P2SH coinbase outputs on mining submission

ctnguyen:
Resolves https://github.com/bitcoin-sv/teranode/issues/4691

## Why

The issue claimed Teranode could mine a block with a P2SH coinbase output that SV
nodes would reject, forking itself off the network. Verification against bitcoin-sv,
BDK, and Teranode sources refuted that: in bitcoin-sv the `bad-txns-vout-p2sh` rule
(`CheckRegularTransaction`) applies to non-coinbase transactions only, and
`CheckCoinbase` — the consensus path for the coinbase — never inspects output scripts.
A P2SH-coinbase block is therefore consensus-valid network-wide, and Teranode's
acceptance behaviour already matches svnode exactly (verdict: no consensus bug).

What svnode does have is a **local mining guard**: its mining RPCs
(`generateblocks` / `submitblock` / `submitminingsolution`, via `HasP2SHOutput` in
`rpc/mining.cpp`) refuse to *originate* such a block, protecting the miner from paying
the reward into an output that post-Genesis degrades to an anyone-can-spend hash
puzzle. Teranode's `SubmitMiningSolution` lacked that guard — a miner- or pool-supplied
coinbase template paying to P2SH was assembled and announced unchecked. This PR closes
that gap.

## What

- `SubmitMiningSolution` now rejects a coinbase paying to a P2SH-shaped output
  (exact svnode `IsP2SH` shape: 23 bytes, `OP_HASH160 <20-byte push> OP_EQUAL`) with
  `bad-txns-vout-p2sh`. Both coinbase sources (pool-supplied template and internally
  recreated candidate) converge through one shared check
  (`validateCoinbaseForSubmission`).
- Deliberately **not** added to `Block.Valid` or any block-validation path: svnode's
  `CheckBlock`/`CheckCoinbase` accepts a P2SH coinbase from peers, so rejecting it
  during validation would itself create a consensus divergence — the opposite of the
  issue's intent.

## Tests

- Guard: exact-shape and near-miss cases, end-to-end rejection of a pool-supplied
  P2SH coinbase, non-P2SH control, nil-safety, shared-check seam
  (`services/blockassembly/submit_mining_solution_test.go`).
- Accept-side regression pin: `TestBlockValid_AcceptsP2SHCoinbaseOutput`
  (`model/Block_test.go`) validates a P2SH-coinbase block at a post-Genesis height and
  asserts `Block.Valid` accepts it, so a future attempt to move this guard into the
  validation path fails loudly.
- `go build`, `go vet`, `go test -race` on the touched packages, and lint: all green.

### DIFF
--- a/model/Block_test.go
+++ b/model/Block_test.go
@@ -1383,6 +1383,113 @@ func TestBlock_ValidWithOneTransaction(t *testing.T) {
 	require.False(t, hasTransactionsReferencingOldBlocks)
 }
 
+// TestBlockValid_AcceptsP2SHCoinbaseOutput pins accept-side parity with bitcoin-sv:
+// CheckBlock/CheckCoinbase never runs the P2SH output scan on the coinbase (the
+// bad-txns-vout-p2sh rule in CheckRegularTransaction applies to non-coinbase
+// transactions only), so a peer block whose coinbase pays to a P2SH-shaped output is
+// consensus-valid and Block.Valid must accept it. Rejecting it here would make the
+// node reject blocks the rest of the network accepts. The mining-side guard against
+// *originating* such a coinbase lives in blockassembly SubmitMiningSolution and must
+// never be mirrored into this validation path.
+func TestBlockValid_AcceptsP2SHCoinbaseOutput(t *testing.T) {
+	validateWithCoinbase := func(t *testing.T, coinbase *bt.Tx) (bool, error) {
+		t.Helper()
+
+		tSettings := test.CreateBaseTestSettings(t)
+
+		// The pin must hold at a POST-Genesis height: a (wrong) Genesis-gated
+		// P2SH rejection added to Block.Valid would not fire at height 0.
+		// Regression params: GenesisActivationHeight=10000, BIP0034Height is
+		// unreachable (BIP34 coinbase-height check stays off), BIP65/66 floors
+		// require header version 4 at this height.
+		height := tSettings.ChainCfgParams.GenesisActivationHeight + 1
+
+		// The coinbase may claim at most fees+subsidy at this height (all outputs count).
+		for _, out := range coinbase.Outputs {
+			out.Satoshis = 0
+		}
+		coinbase.Outputs[0].Satoshis = util.GetBlockSubsidyForHeight(height, tSettings.ChainCfgParams)
+
+		// Easy-difficulty header bound to this coinbase (merkle root of a
+		// coinbase-only block is the coinbase txid).
+		bits, err := NewNBitFromString("207fffff")
+		require.NoError(t, err)
+
+		blockHeader := &BlockHeader{
+			Version:        4,
+			HashPrevBlock:  &chainhash.Hash{},
+			HashMerkleRoot: coinbase.TxIDChainHash(),
+			Timestamp:      uint32(time.Now().Unix()), // nolint:gosec
+			Bits:           *bits,
+			Nonce:          0,
+		}
+
+		// Grind the nonce until the easy target is met; HasMetTargetDifficulty
+		// returns an error (not just false) for a miss, so only `ok` matters here.
+		for {
+			ok, _, _ := blockHeader.HasMetTargetDifficulty()
+			if ok {
+				break
+			}
+
+			require.Less(t, blockHeader.Nonce, uint32(1_000_000), "could not grind a nonce meeting the easy target")
+			blockHeader.Nonce++
+		}
+
+		b, err := NewBlock(blockHeader, coinbase, []*chainhash.Hash{}, 1, 123, height, 0)
+		require.NoError(t, err)
+
+		subtreeStore, err := null.New(ulogger.TestLogger{})
+		require.NoError(t, err)
+
+		utxoStoreURL, err := url.Parse("sqlitememory:///test")
+		require.NoError(t, err)
+
+		utxoStore, err := sql.New(context.Background(), ulogger.TestLogger{}, tSettings, utxoStoreURL)
+		require.NoError(t, err)
+
+		currentChain := make([]*BlockHeader, 11)
+		currentChainIDs := make([]uint32, 11)
+
+		for i := 0; i < 11; i++ {
+			currentChain[i] = &BlockHeader{
+				HashPrevBlock:  &chainhash.Hash{},
+				HashMerkleRoot: &chainhash.Hash{},
+				Timestamp:      blockHeader.Timestamp - 100 - uint32(i), // nolint:gosec
+			}
+			currentChainIDs[i] = uint32(i) // nolint:gosec
+		}
+
+		oldBlockIDs := txmap.NewSyncedMap[chainhash.Hash, []uint32]()
+
+		return b.Valid(context.Background(), ulogger.TestLogger{}, subtreeStore, utxoStore, oldBlockIDs, currentChain, currentChainIDs, tSettings, nil)
+	}
+
+	t.Run("control: unmodified coinbase is accepted", func(t *testing.T) {
+		coinbase, err := bt.NewTxFromString(CoinbaseHex)
+		require.NoError(t, err)
+
+		v, err := validateWithCoinbase(t, coinbase)
+		require.NoError(t, err)
+		require.True(t, v)
+	})
+
+	t.Run("P2SH coinbase output is accepted", func(t *testing.T) {
+		coinbase, err := bt.NewTxFromString(CoinbaseHex)
+		require.NoError(t, err)
+
+		// Redirect the payout to the exact P2SH shape svnode's IsP2SH tests:
+		// OP_HASH160 <20-byte push> OP_EQUAL.
+		p2shScript, err := bscript.NewFromHexString("a914000000000000000000000000000000000000000087")
+		require.NoError(t, err)
+		coinbase.Outputs[0].LockingScript = p2shScript
+
+		v, err := validateWithCoinbase(t, coinbase)
+		require.NoError(t, err)
+		require.True(t, v)
+	})
+}
+
 func TestGetAndValidateSubtrees(t *testing.T) {
 	tSettings := test.CreateBaseTestSettings(t)
 	blockHeaderBytes, _ := hex.DecodeString(block1Header)
--- a/model/Block_test.go
+++ b/model/Block_test.go
@@ -1386,7 +1386,10 @@ func TestBlock_ValidWithOneTransaction(t *testing.T) {
 // TestBlockValid_AcceptsP2SHCoinbaseOutput pins accept-side parity with bitcoin-sv:
 // CheckBlock/CheckCoinbase never runs the P2SH output scan on the coinbase (the
 // bad-txns-vout-p2sh rule in CheckRegularTransaction applies to non-coinbase
-// transactions only), so a peer block whose coinbase pays to a P2SH-shaped output is
+// transactions only — svnode validation.cpp:611-623, DoS 100, on both the mempool and
+// block-consensus paths; the mining RPCs reuse the same bad-txns-vout-p2sh string via
+// HasP2SHOutput, but that is a separate mining-side emitter, not the consensus rule), so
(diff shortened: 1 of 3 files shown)


ordishs (approved):
Clean, well-reasoned, well-tested. The framing is the strongest part: correctly identified as a mining-origination guard (not a consensus rule), deliberately kept out of `Block.Valid`, with a regression pin (`TestBlockValid_AcceptsP2SHCoinbaseOutput`) that fails loudly if anyone later moves it into validation — exactly right, since mirroring it into validation would fork the node off the network.

- Shape check matches svnode's `IsPayToScriptHash()` byte-for-byte (23 bytes, 0xa9 / 0x14 / 0x87); len==23 guard precedes the s[22] access, no OOB.
- All coinbase outputs scanned, not just output 0.
- Both branches (pool-supplied and recreated candidate) converge on the single `validateCoinbaseForSubmission` call, with a comment and a direct seam test pinning that convergence.
- Nil-safety handled at both levels as defense-in-depth.

Minor (non-blocking): the pool branch checks input count both at the call site (guarding the Inputs[0] deref) and again inside the shared function, producing two slightly different error strings for the same condition. And the guard is unconditional rather than Genesis-gated — safe for an origination guard, just marginally stricter than a strictly post-Genesis reading, with no consensus consequence.

Low risk, additive, narrowly scoped. Approving.

icellan (approved):

icellan:
**P0 — coinbase P2SH guard also blocks `generatetoaddress` (`services/blockassembly/Server.go:1535`)**

The guard is placed at the shared `submitMiningSolution` convergence point, so it fires on the `generate`/`generatetoaddress` path too, not just pool submissions.

`generatetoaddress` with a P2SH (`3...`) address now hard-fails:
`handleGenerateToAddress → GenerateBlocks → generateBlock → mining.Mine(address) → CreateCoinbaseTxCandidateForAddress → CreateCoinbase → AddressToScript` emits the 23-byte `OP_HASH160 <20> OP_EQUAL` script, then `validateCoinbaseForSubmission` rejects it with `bad-txns-vout-p2sh`. That block succeeded before this PR.

bitcoin-sv only guards the pool candidate/solution RPCs here (`HasP2SHOutput`); its `generatetoaddress` still allows a P2SH coinbase. This PR is stricter than upstream on that path.

Scope the check to the pool-submission branch rather than the shared convergence, and add a test covering the generate path (current tests only cover the pool branch + the helper).

---

Pre-existing, out of scope for this PR but worth a follow-up: `Server.go:1493` dereferences `bestBlockHeader` from `CurrentBlock()` without checking the discarded second return — `CurrentBlock()` can return `(nil, 0)` before the assembler's best block is loaded (early startup / post-reset), which would panic. This PR routes coinbase validation through the same function but didn't introduce the deref.

ctnguyen:
Thanks @icellan for the review, you're exactly right about the teranode call graph: the guard lives at the
shared `validateCoinbaseForSubmission` convergence, so `generatetoaddress` with a P2SH
address does now hit it via `GenerateBlocks → generateBlock → submitMiningSolution`.

On the parity point I think the upstream comparison lands the other way, though. In
bitcoin-sv, `generatetoaddress` (`src/rpc/mining.cpp:227`) just delegates to
`generateBlocks` (`:267`), and `generateBlocks` carries the guard itself:

```cpp
// src/rpc/mining.cpp:199-201
if (shared_pblock->vtx[0]->HasP2SHOutput()) {
    throw JSONRPCError(RPC_TRANSACTION_REJECTED, "bad-txns-vout-p2sh");
}
```

That runs before `ProcessNewBlock`, so upstream `generatetoaddress` with a `3…` address
hard-fails with `bad-txns-vout-p2sh` too. The `HasP2SHOutput` check on the
pool/submission RPCs (via `processBlock`, `:841`) is in addition to this one, not a
replacement. So keeping our guard at the shared convergence is exact parity across all
origination paths; scoping it to the pool branch would actually leave our generate path
looser than bitcoin-sv. On that basis I've kept the placement as-is.

Your test point was well taken regardless — the existing tests only exercised the pool
branch and the helper. Fixed in the two commits now on the PR:

**Commit "blockassembly: cover the generate path in the coinbase P2SH guard tests"**
- `TestGenerateBlock_RejectsP2SHCoinbaseOutput`: drives `generateBlock` with a real
  P2SH address (`2N…`), exercising the full
  `Mine → CreateCoinbaseTxCandidateForAddress → AddressToScript →
  submitMiningSolution → validateCoinbaseForSubmission` chain, and asserts the
  `bad-txns-vout-p2sh` rejection end-to-end.
- `TestGenerateBlock_AllowsP2PKHCoinbaseOutput`: the non-P2SH companion — the guard
  must not fire on a standard address.
- The guard's comment now cites bitcoin-sv's `generateBlocks` guard on the generate
  path (`rpc/mining.cpp:199`) alongside the general `rpc/mining.cpp` mining-RPC
  reference, so the shared-convergence placement is self-documenting as parity, not
  extra strictness.

**Commit "blockassembly: guard against nil CurrentBlock() in mining submission"**
Good catch on the deref. `CurrentBlock()` returns
`(nil, 0)` before the best block is loaded, and it turned out the generate flow had
the same unchecked pattern in two more spots beyond the one you cited. All three are
fixed since this PR routes generation through `submitMiningSolution`:
- `submitMiningSolution` (the site you flagged): nil header now returns a clean
  `no current best block yet` processing error instead of panicking.
- `generateBlock` pre-submit capture: same clean-error treatment.
- `waitForBestBlockHeaderUpdate` post-submit poll: a nil tick is treated as "not
  loaded yet" — the loop skips it and keeps polling until the update or its timeout,
  no error.
- Covered by `TestSubmitMiningSolution_NoCurrentBlock` (constructs an assembler with
  no best block, asserts the clean error at the new guard) and
  `TestWaitForBestBlockHeaderUpdate_ToleratesNilCurrentBlock` (nil ticks, returns on
  timeout, no panic).

A `grep` sweep found three more `CurrentBlock()` derefs elsewhere in the file — the
post-assembly debug log in `submitMiningSolution`, and the `GetBlockAssemblyState` and
`GetCurrentDifficulty` handlers — which sit outside the reviewed generate/submission
flow; left out of this PR deliberately and noted as follow-up scope.

oskarszoon (approved):
Reviewed with bitcoin-expert, golang-pro, and code-reviewer. The core P2SH-coinbase guard is correct: shape check is byte-for-byte svnode `IsPayToScriptHash`, it scans all outputs (matching `HasP2SHOutput`), both coinbase branches converge through `validateCoinbaseForSubmission` before assembly, and keeping it origination-only (not in `Block.Valid`) is the right call — mirroring it into validation would self-fork off the network. The accept-side pin locks that in. Approving.

Non-blocking, all test-quality / comment accuracy:

- `submit_mining_solution_test.go:493` `TestWaitForBestBlockHeaderUpdate_ToleratesNilCurrentBlock` can't catch its target regression. `waitForBestBlockHeaderUpdate` returns nil on both exit paths, and `(*BlockHeader)(nil).Hash()` doesn't actually panic (`Bytes()` nil-guards the receiver) — so the test stays green whether the guard is present or dropped. Assert elapsed `>= ~150ms`, or assert the warn-log fired.
- `submit_mining_solution_test.go:401-416` and `:445-457` — the two "not rejected" tests are near-vacuous: `if err != nil { require.NotContains(...) }` stays green even if the guard is deleted. The property is genuinely covered by the unit tests; these read stronger than they are.
- `Server.go:2308` `generateBlock` nil-guard is the one of the three that's untested (`Start()` sets `bestBlock` synchronously, so the nil branch is dead from the suite's view).

nits:
- `Server.go:1662-1663` reuses `bestBlockHeader` with an unguarded `.Timestamp` deref in a Debugf arg (evaluated regardless of log level). Safe today only because `bestBlock` is monotonic once set + the `1498` guard already passed — not a live bug, but the one `CurrentBlock()` site the hardening skipped. A guard or an invariant comment would close it.
- `Server.go:2306` and `:2350` guard comments overclaim "panic" — `Hash()` is nil-safe, so those two guards prevent a bogus-hash logic error, not a crash. `1497`'s comment is accurate (`.HashPrevBlock` genuinely panics). Fix the reasons, keep the code.
- `model/Block_test.go:1302` attributes `bad-txns-vout-p2sh` to `CheckRegularTransaction`; that string is the mining-RPC guard, not a regular-tx consensus rule.

ctnguyen:
Thanks @oskarszoon for the careful pass — these are good catches on the test strength. Addressed in
two commits on the PR ("blockassembly: strengthen the coinbase-guard negative tests" and
"blockassembly: correct guard comments and document nil-header invariants"):

  ✅ nil-tolerance test couldn't fail. You're right: both loop exits return
  `nil` and `Hash()` is nil-safe (`Bytes()` guards the receiver), so it stayed green
  either way. Added an elapsed-time floor — with the guard the loop only returns at the
  ~200ms timeout, without it the first tick returns in ~10ms, so the assertion now
  actually pins the behaviour (verified both directions by temporarily removing the
  guard: the test fails without it).
  ✅ "not rejected" tests near-vacuous. Agreed they don't catch deletion — they
  only catch an over-broad guard. Asserted the fixture is a true negative
  (`coinbaseHasP2SHOutput == false`) so the input side isn't vacuous, and reworded the
  comments to state exactly what they do and don't prove; the deletion direction stays
  covered by the helper/validator unit tests.
  ✅ `generateBlock` guard unexercised. Correct — every generate-path test
  starts the assembler, which loads the best block synchronously, and reaching the nil
  branch would need a candidate to mine without a best block, which can't happen. Kept
  it as defence-in-depth with a comment saying so.
  ✅ unguarded `.Timestamp` reuse. Good eye — documented the invariant on the
  post-assembly deref (the earlier guard plus no production path clearing `bestBlock`
  back to nil once loaded), rather than adding a redundant check.
  ✅ guard comments overclaiming "panic". Fixed: the `generateBlock` (and, for
  clarity, the polling-loop) comment now says those guards prevent a bogus-hash
  comparison, not a crash, since `Hash()` is nil-safe. Left the `submitMiningSolution`
  comment as-is because `.HashPrevBlock` really does panic on a nil header.
  ✅ `bad-txns-vout-p2sh` attribution. Added a clarifying citation to the test
  comment, but kept the attribution, which I believe is right as written: it is a
  regular-transaction consensus rule — svnode raises it in `CheckRegularTransaction`
  (`src/validation.cpp:611-623`, DoS 100), reached from both the mempool path
  (`TxnValidation`) and the block path (`CheckBlock`). The mining RPCs (`generateBlocks`
  at `src/rpc/mining.cpp:199`, `processBlock` at `:841`) reuse the *same string* via
  `HasP2SHOutput`, which is probably where the overlap comes from — but the consensus
  rule itself lives in `CheckRegularTransaction`. Happy to adjust if you're reading a
  different code path.

oskarszoon (approved):
Re-reviewed at 4c61e0e4c — the two follow-up commits address everything from the last pass, verified against the code rather than the commit messages. Approve stands.

- The nil-tolerance test now catches its target regression: the added 150ms elapsed floor sits between the ~10ms early-exit and the 200ms deadline. Confirmed by mutation — dropping the `continue` guard fails the test at 0.02s. Good.
- The two "not rejected" tests now pin the input as genuinely non-P2SH (`require.False(coinbaseHasP2SHOutput(...))`) and the comments state plainly what they prove vs. don't, pointing at the unit tests that cover guard-deletion. No longer reads stronger than it is.
- Both guard comments corrected (`Hash()` is nil-safe → bogus-hash logic bug, not a panic), and the `Server.go:1662` `.Timestamp` deref now documents the monotonic-bestBlock invariant.
- The `Block_test.go` P2SH comment now cites `validation.cpp:611-623` (DoS 100, Genesis-gated, non-coinbase) vs. the `rpc/mining.cpp` `HasP2SHOutput` mining-side emitter — checked both against svnode; accurate, and it correctly separates the consensus rule from the mining-RPC guard.

CI green. No new findings.

PR #1324, opened 2026-07-22
fix(p2p): read backoffMultiplier under lock in considerReputationRecovery

gokutheengineer:
**Closes #4720.**

### Problem
`SyncCoordinator.considerReputationRecovery` read `sc.backoffMultiplier` without holding `sc.mu`, while `checkAndClearExpiredBackoff`, `resetBackoff`, and `enterBackoffMode` write it under the lock from other goroutines. This is a data race (fails `-race`) and can miscompute the reputation-recovery cooldown from a torn/stale read.

### Fix
Snapshot `backoffMultiplier` into a local variable under `sc.mu.RLock()` at the start of `considerReputationRecovery` and compute the cooldown from the snapshot. Kept the mutex approach (rather than converting to an atomic) since all other accesses already synchronize via `sc.mu`.

### Tests
Added `TestSyncCoordinator_ConsiderReputationRecovery_ConcurrentWithBackoffWriters`, which runs the backoff writers (`enterBackoffMode` / `checkAndClearExpiredBackoff` / `resetBackoff`) concurrently with `considerReputationRecovery`. Verified it fails with the exact race report on the unfixed code and passes with the fix.

Ran (affected package only):
```
```
All green.

### DIFF
--- a/services/p2p/sync_coordinator.go
+++ b/services/p2p/sync_coordinator.go
@@ -1186,11 +1186,15 @@ func (sc *SyncCoordinator) checkAllPeersAttempted() {
 
 // considerReputationRecovery checks if any bad peers should have their reputation reset
 func (sc *SyncCoordinator) considerReputationRecovery() {
+	sc.mu.RLock()
+	backoffMultiplier := sc.backoffMultiplier
+	sc.mu.RUnlock()
+
 	// Calculate cooldown based on how many times we've been in backoff
 	baseCooldown := 5 * time.Minute
-	if sc.backoffMultiplier > 1 {
+	if backoffMultiplier > 1 {
 		// Exponentially increase cooldown if we've been in backoff multiple times
-		cooldownMultiplier := sc.backoffMultiplier / 2
+		cooldownMultiplier := backoffMultiplier / 2
 		if cooldownMultiplier < 1 {
 			cooldownMultiplier = 1
 		}
--- a/services/p2p/sync_coordinator_test.go
+++ b/services/p2p/sync_coordinator_test.go
@@ -245,6 +245,37 @@ func TestSyncCoordinator_ConsiderReputationRecovery_NoCandidatesIsNoOp(t *testin
 	require.GreaterOrEqual(t, got.ReputationScore, 50.0, "healthy peer reputation untouched")
 }
 
+// Exercises concurrent backoff writers against considerReputationRecovery's read of
+// backoffMultiplier; fails under -race if the read is not synchronized.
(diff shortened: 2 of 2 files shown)


ctnguyen (approved):
Review at commit [72e0d47](https://github.com/bsv-blockchain/teranode/commit/72e0d4744ad7f8251733c8a978acc1b5dd254ecd)

Nice, tidy fix — this is exactly the right shape for a data-race patch: minimal, well-scoped, and honest about the tradeoff. Snapshotting `backoffMultiplier` under `sc.mu.RLock()` and computing the cooldown off the local copy is the correct call, and keeping the mutex (rather than reaching for an atomic) is the right instinct given every sibling access already goes through `sc.mu`; converting one field to atomic would have left the struct with two inconsistent synchronization stories. I checked the completeness angle you'd want checked: `backoffMultiplier`, `allPeersAttempted`, `lastAllPeersAttemptTime`, and `maxBackoffMultiplier` are now touched only in `checkAndClearExpiredBackoff`, `resetBackoff`, `enterBackoffMode` (all under `Lock`), the constructor before the value escapes, and the new `RLock` snapshot — so no stragglers remain on this path. The RLock scope is a read-and-release, `considerReputationRecovery`'s sole caller (`checkFSMState`) holds no lock when it calls in, and the later `resetBackoff()` re-acquires cleanly, so there's no deadlock or lock-ordering hazard. The regression test drives the real backoff writers against the reader hard enough to surface the race under `-race`, which is the idiomatic way to lock this in. Complete, correct, and I don't have anything blocking or non-blocking to raise.

freemans13 (approved):
Reviewed the diff, read the full surrounding code, and re-ran the race test locally — LGTM.

The fix is correct and complete: I checked every access to `backoffMultiplier` in the file, and every other read/write (lines 1138–1187) already holds `sc.mu`. The read in `considerReputationRecovery` was the only one left outside the lock, so this single `RLock` snapshot closes the last gap.

Design choices are sound — `RLock` is right for a pure read on the RWMutex, and keeping the mutex rather than converting to an atomic avoids guarding the field two different ways. The residual time-of-check/time-of-use window is harmless since the cooldown is best-effort; the torn read was the actual bug.

The test earns its place: the writer goroutine forces backoff expiry and re-enters backoff so all three write paths fire, and it uses the real registry + `context.Background()` so there's no nil-panic risk. Fair to note it only proves anything under `-race`, but that's the standard shape for a race regression test.

Confirmed locally:
```
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 \
  -run 'TestSyncCoordinator_ConsiderReputationRecovery' ./services/p2p/
→ 2 passed
```

Issue #1323, opened 2026-07-22
cmd/teranode-cli: guarded "unspend" command to clear dangling spender refs (operational unwedge)


oskarszoon:
# cmd/teranode-cli: guarded `unspend` command to clear dangling spender refs (operational unwedge)

## Goal

An operational escape hatch to unwedge a node stuck on a block whose counter-conflicting walk dereferences a **dangling spender ref** — a parent output marked spent by a tx whose own UTXO record is absent. It is deliberately small, non-consensus, and reviewable quickly — faster to land than the read-side guard (#1320) or the write-side prevention (#1214), both of which are still in flight. This tool does not replace either: it repairs an already-corrupt store on a live incident.

No existing tool can do this. `cmd/rewindblockchain` iterates only unmined/conflicting records and never visits the mined parent holding the stale slot (verified in #1213). `RemoveFromConflictingChildren` currently has zero runtime callers.

## Motivation

`mainnet-eu-1` (v0.15.5) is wedged at block 958729. Store state confirms the dangling ref: the winner `54f3dae6…`'s three inputs are each marked SPENT by ghost `d37d9851…`, whose record is `TX_NOT_FOUND`:

| Parent output | status | spendingData.txId | ghost record |
|---|---|---|---|
| `07146677…:1` | SPENT | `d37d9851…` vin6 | absent |
| `fa6eb99a…:1` | SPENT | `d37d9851…` vin5 | absent |
| `d988c204…:1` | SPENT | `d37d9851…` vin4 | absent |

Clearing these slots lets the node re-bless the winner cleanly on the next attempt: winner spends the now-unspent outputs → no conflict → walk not tripped → subtree validates → block accepted → self-heal (winner promotion rewrites the slots).

## Command

```
teranode-cli unspend <parentTxid:vout> [<parentTxid:vout> ...]   # dry-run by default
teranode-cli unspend --commit <parentTxid:vout> ...              # perform the writes
```

Operator supplies the exact dangling parent outputs (the ghost's own tx bytes are gone, so the tool cannot discover them from the ghost — it works from the parent side). Dry-run is the default and prints exactly what would change; `--commit` performs writes and is gated behind the existing dangerous-action confirmation prompt.

## Safety invariants (non-negotiable)

Per named output, the command **refuses to act** unless ALL hold:

1. **The output is currently SPENT.** An unspent output is a no-op (idempotent).
[...]

PR #1325, opened 2026-07-22
fix(utxo): tolerate confirmed ghost spenders in counter-conflicting walk and heal their dangling spends

### LINKED ISSUE #1320: subtreevalidation: tolerate a dangling spender ref in the counter-conflicting walk (minimal read-side unwedge)
# subtreevalidation: tolerate a dangling spender ref in the counter-conflicting walk (minimal read-side unwedge)

## Goal

Unwedge nodes stuck on a block whose conflicting tx references a spender whose UTXO record no longer exists (or never existed). Deliberately minimal: the tolerance + observability only. The consensus-band marker hardening that sank #1217/#1239 is explicitly out of scope here and tracked separately.

This is the read-side counterpart to #1214 (write-side, prevents the dangling ref forming). #1213 was the original read-side issue; it was closed when its only known instance was reset and the fix PRs (#1217 → #1239) were not carried through. New in-the-wild occurrences (see #1214, block 958729) show the dangling ref recurs and re-wedges nodes on v0.15.5, so the read-side guard is needed independently of the write-side fix — and for a node that never held the loser, the write-side fix cannot help it.

## Failure

`GetCounterConflictingTxHashes` walks a mined parent's `SpendingDatas`, and for each recorded spender calls `GetConflictingChildren(spendingTxID)`. When that spender's own record is absent, the store `Get` returns `TX_NOT_FOUND`, which propagates unwrapped (`process_conflicting.go:1122-1125`). `checkCounterConflictingOnCurrentChain` (`SubtreeValidation.go:471-473`) converts any error into a fatal `ProcessingError`, the subtree fails, the block fails, and both the legacy `HandleBlockDirect` and the catchup `CheckBlockSubtreesRequest` paths retry the same block forever. No exit condition, no metric.

```
[CheckSubtree] Failed to validate legacy subtree <st>
-> [processTxMetaUsingStore][<winner>] failed to check counter conflicting tx on current chain
-> [checkCounterConflictingOnCurrentChain][<winner>] failed to get counter conflicting tx hashes
-> TX_NOT_FOUND (30): <ghost> not found
```

## In scope (the minimal change)

- `stores/utxo/process_conflicting.go` `GetCounterConflictingTxHashes` (~L1118-1134): when `GetConflictingChildren(spendingTxID)` returns NOT_FOUND, treat that spender as an absent ghost — exclude it from the counter-conflicting set and continue the walk. Do not add the absent spender at L1120; skip its child expansion. NOT_FOUND = `errors.Is(err, errors.ErrTxNotFound) || errors.Is(err, errors.ErrNotFound)` (aerospike vs SQL; same pairing as `cmd/rewindblockchain/rewindblockchain/tx_delete.go`).
- Fail closed, unchanged, on:
  - the tx-under-check itself missing (`s.Get(txHash)`, L1072-1075),
  - a missing **parent** record (L1091-1094) — an absent parent erases the whole spend graph for that input; propagate the error rather than dereference a nil slot,
  - all `GetMeta` / `GetConflictingChildren` errors in `checkCounterConflictingOnCurrentChain` (L488, L504) — the walk saw the counter alive moments earlier, so a NOT_FOUND at re-read is a racing delete (TOCTOU); on retry the walk re-runs.
- Observability: WARN log + a prometheus counter incremented once per tolerated ghost (parent txid:vout + ghost txid), so store inconsistency is visible instead of silent.
- Tests: unit test in `stores/utxo/process_conflicting_test.go` for the tolerated-spender case (extends the existing `TestGetCounterConflictingTxHashes_*` set), and a subtree-validation test asserting a block referencing a ghost spender validates instead of erroring.

Single chokepoint: both wedge paths route through `GetCounterConflictingTxHashes`, so the one store-layer change covers legacy and catchup.

## Explicitly out of scope (what sank #1217 / #1239 — track separately)

- The `deletedChildren` marker discriminator to distinguish a deliberately-reaped **mined** counter from a benign never-mined ghost in the `(retention, retention*2]` = `(288, 576]` band.
- Pruner marker write-reliability rework on either backend (SQL deletable-set temp table + single-transaction marker/delete; aerospike unconditional page-keyed marker writes, cuckoo-filter removal, ordered marker-before-delete atomicity).
- Frozen-sentinel handling rework and `MarkConflictingRecursively` filtering changes.
- Bounded retry / escalation for persistent block-validation failure during catchup (#980).

## Consensus safety / accepted residual

The check exists to reject blessing a conflicting tx whose counter-party is already mined on **our** chain. An absent counter has no BlockIDs, so it is definitionally not mined on our chain and the check should pass. This is exactly correct for a never-mined purged loser (every field occurrence so far).

Accepted residual: in the `(288, 576]` band a counter that genuinely was mined on our chain and then pruned at its retention horizon is also absent, and plain tolerance waves it through. Reaching that state requires an incoming block that spends an output already spent >retention (288) blocks deep on our chain — a double-spend below finality, only reachable via a >288-deep reorg, which is rejected by finality/reorg handling, not this check. #1214 additionally removes the mechanism that creates the dangling ref. The marker discriminator that closes this band deterministically is deferred to the follow-up above; it is defense-in-depth, not the unwedge.

## Acceptance criteria

- [ ] A block whose conflicting tx references a spender with no store record validates (does not return `ProcessingError`) on both the legacy `HandleBlockDirect` and catchup `CheckBlockSubtreesRequest` paths.
- [ ] Missing parent and missing tx-under-check still fail closed; `checkCounterConflictingOnCurrentChain` `GetMeta`/children still fail closed.
- [ ] Every tolerated ghost emits a WARN + increments a prometheus counter.
- [ ] Unit + subtree-validation tests cover tolerated-spender, missing-parent (fail-closed), and mined-counter-present (still rejects) cases.
- [ ] Diff is confined to `process_conflicting.go` + tests (+ the metric registration); no pruner or marker changes.
- [ ] Backport target: patch on the `v0.15.x` line so wedged instances unwedge on redeploy.

## Related

- #1213 (original read-side issue, closed), #1217 → #1239 (read-side PRs, not carried — scope grew into the marker hardening now deferred above)
- #1214 (write-side cause; prevents the dangling ref forming)
- #980 (unbounded retry / no escalation — the reason a single bad block wedges forever)

icellan:
Closes #1320. v0.15 line only — main gets a different fix. Write-side cause tracked in #1214.

## Problem

A multi-input tx makes its valid spends, errors on another input, and its record is never created — the parent outputs are left recording a spender that does not exist (a ghost). A block containing a conflicting tx over such a slot wedges the node twice:

1. **Subtree validation**: `GetCounterConflictingTxHashes` fails with `TX_NOT_FOUND`, `checkCounterConflictingOnCurrentChain` turns it into a fatal `ProcessingError`, and both the legacy `HandleBlockDirect` and catchup `CheckBlockSubtreesRequest` paths retry the same block forever.
2. **Block assembly**: fixing (1) alone just moves the wedge. `moveForwardBlock` feeds the blessed winner to `ProcessConflicting`, which runs the same walk — and even with a tolerant walk, step-3 `Spend(winner)` fails `SPENT`: the Lua spend rejects a mismatched occupant, and no loser-derived unspend can clear a ghost's spend (the unspend ownership check only clears a spend the caller owns, and a ghost has no record to derive it from).

## Fix

- The walk probes a spender whose conflict BFS returns NOT_FOUND and tolerates only a **confirmed-absent** spender: excluded from the counter set, reported as a ghost slot (parent txid, vout, utxohash, recorded spending data).
- The probe is not optional. `GetConflictingChildren` is a BFS — NOT_FOUND can mean a missing *descendant* of a live spender, possibly mined on our chain within the mined-check horizon (retention×2 = 576 blocks vs 288-block record retention). Dropping the spender in that case would bless a same-chain double-spend, no reorg needed. The issue's original prescription had this hole; with the probe, a live spender with a missing descendant stays fail-closed.
- `ProcessConflicting` appends the ghost slots to the step-2 unspend, passing the recorded ghost spending data as the expected value — this passes the ownership check on both backends, the winner spends the slot in step 3, and the dangling reference is healed permanently. The heal is monotonic under rollback: a cleared slot stays cleared, and the retry converges (the walk then sees an unspent slot).
- Fail closed, unchanged: the tx-under-check missing, a missing parent record, an inconclusive probe, and all `GetMeta`/children errors in `checkCounterConflictingOnCurrentChain`.
- Observability: every tolerated ghost increments `teranode_utxo_counter_conflicting_ghost_spends` (the free functions have no logger, so a counter is the surface — same conclusion #1239 reached).

`Store.GetCounterConflicting` keeps its signature; `ProcessConflicting` calls the ghost-aware walk directly because it needs the slots.

## Tests

- `stores/utxo/aerospike/counter_conflicting_ghost_test.go` — end-to-end against real Aerospike: ghost planted through the store API (multi-input spends applied, record never created), walk tolerates, `ProcessConflicting` heals the contested slot through the real Lua ownership check, winner spends, uncontested ghost slot untouched, parent unlocked.
- `stores/utxo/sql/counter_conflicting_ghost_test.go` — walk tolerance on sqlitememory (resolution can't run there: `SetConflicting` write-txn + pool reads self-deadlock on SQLite, see `setconflicting_cascade_test.go`).
- `services/subtreevalidation/counter_conflicting_ghost_test.go` — ghost validates; a live mined counter still rejects.
- Unit tests for the fail-closed partition (live spender + missing descendant, missing parent, probe error) and the resolution orchestration. Existing `ProcessConflicting` mock tests rewritten to stub the walk realistically — the old `GetCounterConflicting` stubs wrongly excluded the winner from the loser set.

`go test -race` green on `stores/utxo`, `stores/utxo/sql`, `services/subtreevalidation`, `services/blockassembly/subtreeprocessor`, `stores/txmetacache`. Pre-existing failures in `stores/utxo/sql` (Postgres container tests, `TestMinedThenSpendAllPrunes_SQLite`) fail identically on the clean release/v0.15 baseline.

## Accepted residuals

- A counter genuinely mined on our chain and DAH-pruned (≥~288 deep, inside the 576-block horizon) is indistinguishable from a ghost once its record is gone; tolerance waves it through. Exploiting it costs a miner a full block for a tx other implementations reject anyway. The marker discriminator that closes this band stays deferred, per the issue.
- Frozen-child checks are skipped for a genuine root ghost's subtree (a freeze sitting strictly below a pruned spender). Narrow; the probe cannot close it.
- The Lua conflicting-branch DAH (`teranode.lua:985-995`) reaps flagged records ignoring `blockIDs` — undermines "absent ⟹ not mined" independently of this fix. Separate issue to follow.

### DIFF
--- a/services/blockassembly/mark_as_conflicting_test.go
+++ b/services/blockassembly/mark_as_conflicting_test.go
@@ -10,6 +10,8 @@ import (
 	"github.com/bsv-blockchain/teranode/services/blockassembly/subtreeprocessor"
 	"github.com/bsv-blockchain/teranode/settings"
 	utxoStore "github.com/bsv-blockchain/teranode/stores/utxo"
+	"github.com/bsv-blockchain/teranode/stores/utxo/fields"
+	"github.com/bsv-blockchain/teranode/stores/utxo/meta"
 	"github.com/bsv-blockchain/teranode/ulogger"
 	"github.com/stretchr/testify/mock"
 )
@@ -43,6 +45,13 @@ func TestMarkAsConflicting_EvictsCascadedDescendants(t *testing.T) {
 	// Level 3: grandchild has no children — BFS terminates
 	mockStore.On("SetConflicting", mock.Anything, []chainhash.Hash{grandchildHash}, true).
 		Return([]*utxoStore.Spend{{TxID: &grandchildHash, Vout: 0}}, []chainhash.Hash{}, nil)
+
+	// MarkConflictingRecursively probes each discovered child before recursing
+	// (the ghost filter); both descendants are live records, so the probe finds them.
+	mockStore.On("Get", mock.Anything, &childHash, []fields.FieldName{fields.Conflicting}).
+		Return(&meta.Data{}, nil)
+	mockStore.On("Get", mock.Anything, &grandchildHash, []fields.FieldName{fields.Conflicting}).
+		Return(&meta.Data{}, nil)
 
 	mockStp := &subtreeprocessor.MockSubtreeProcessor{}
 	mockStp.On("Remove", mock.Anything, parentHash).Return(nil).Once()
--- a/services/blockassembly/subtreeprocessor/ConflictingTransactions_test.go
+++ b/services/blockassembly/subtreeprocessor/ConflictingTransactions_test.go
@@ -94,7 +94,7 @@ func TestProcessConflictingTransactions(t *testing.T) {
 	_ = conflictingTxInput.PreviousTxIDAdd(&conflictingParentHash)
 	conflictingTxBody.Inputs = append(conflictingTxBody.Inputs, conflictingTxInput)
 
-	mockUtxoStore.On("Get", mock.Anything, mock.Anything, []fields.FieldName{fields.Utxos}).
+	mockUtxoStore.On("Get", mock.Anything, mock.Anything, []fields.FieldName{fields.Utxos, fields.DeletedChildren}).
 		Return(&meta.Data{SpendingDatas: []*spend.SpendingData{nil}}, nil)
 	mockUtxoStore.On("Get", mock.Anything, mock.Anything, mock.Anything).Return(&meta.Data{Tx: conflictingTxBody, Conflicting: true}, nil)
 	mockUtxoStore.On("SetConflicting", mock.Anything, mock.Anything, mock.Anything).Return([]*utxo.Spend{}, []chainhash.Hash{}, nil)
--- a/services/blockassembly/subtreeprocessor/ConflictingTransactions_test.go
+++ b/services/blockassembly/subtreeprocessor/ConflictingTransactions_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/bsv-blockchain/go-bt/v2"
 	"github.com/bsv-blockchain/go-bt/v2/chainhash"
 	subtreepkg "github.com/bsv-blockchain/go-subtree"
 	txmap "github.com/bsv-blockchain/go-tx-map"
@@ -16,6 +17,7 @@ import (
 	"github.com/bsv-blockchain/teranode/stores/utxo"
 	"github.com/bsv-blockchain/teranode/stores/utxo/fields"
 	"github.com/bsv-blockchain/teranode/stores/utxo/meta"
+	"github.com/bsv-blockchain/teranode/stores/utxo/spend"
 	"github.com/bsv-blockchain/teranode/stores/utxo/sql"
 	"github.com/bsv-blockchain/teranode/ulogger"
 	"github.com/bsv-blockchain/teranode/util/test"
@@ -82,8 +84,19 @@ func TestProcessConflictingTransactions(t *testing.T) {
 	_ = losingTxMap.Put(conflictingTx2, 1)
 
 	mockUtxoStore.On("ProcessConflicting", mock.Anything, conflictingNodes).Return(losingTxMap, nil)
-	mockUtxoStore.On("Get", mock.Anything, mock.Anything, mock.Anything).Return(&meta.Data{Conflicting: true}, nil)
-	mockUtxoStore.On("GetCounterConflicting", mock.Anything, mock.Anything).Return([]chainhash.Hash{conflictingTx1, conflictingTx2}, nil)
+
+	// The ghost-aware counter-conflicting walk reads the tx body and its parents
+	// directly: give both conflicting txs the same single-input body whose parent
+	// slot is unspent, so each walk resolves to just the tx itself.
+	conflictingTxBody := bt.NewTx()
+	conflictingTxInput := &bt.Input{PreviousTxOutIndex: 0}
+	conflictingParentHash := chainhash.HashH([]byte("conflicting-parent"))
+	_ = conflictingTxInput.PreviousTxIDAdd(&conflictingParentHash)
+	conflictingTxBody.Inputs = append(conflictingTxBody.Inputs, conflictingTxInput)
+
+	mockUtxoStore.On("Get", mock.Anything, mock.Anything, []fields.FieldName{fields.Utxos}).
+		Return(&meta.Data{SpendingDatas: []*spend.SpendingData{nil}}, nil)
+	mockUtxoStore.On("Get", mock.Anything, mock.Anything, mock.Anything).Return(&meta.Data{Tx: conflictingTxBody, Conflicting: true}, nil)
 	mockUtxoStore.On("SetConflicting", mock.Anything, mock.Anything, mock.Anything).Return([]*utxo.Spend{}, []chainhash.Hash{}, nil)
 	mockUtxoStore.On("Unspend", mock.Anything, mock.Anything, mock.Anything).Return(nil)
 	mockUtxoStore.On("Spend", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return([]*utxo.Spend{}, nil)
--- a/services/subtreevalidation/counter_conflicting_ghost_test.go
+++ b/services/subtreevalidation/counter_conflicting_ghost_test.go
@@ -51,7 +51,7 @@ func TestCheckCounterConflictingOnCurrentChain_ToleratesGhostSpender(t *testing.
 
 	mockStore.On("Get", mock.Anything, hashMatcher(winningTxHash), []fields.FieldName{fields.Tx}).
 		Return(&meta.Data{Tx: winningTx}, nil)
-	mockStore.On("Get", mock.Anything, hashMatcher(parentTxHash), []fields.FieldName{fields.Utxos}).
+	mockStore.On("Get", mock.Anything, hashMatcher(parentTxHash), []fields.FieldName{fields.Utxos, fields.DeletedChildren}).
 		Return(&meta.Data{SpendingDatas: []*spend.SpendingData{spend.NewSpendingData(&ghostTxHash, 0)}}, nil)
 	mockStore.On("GetConflictingChildren", mock.Anything, ghostTxHash).
 		Return(([]chainhash.Hash)(nil), errors.NewTxNotFoundError("%v not found", ghostTxHash))
@@ -63,7 +63,7 @@ func TestCheckCounterConflictingOnCurrentChain_ToleratesGhostSpender(t *testing.
 	// the surviving counter set is just the tx itself
 	mockStore.On("GetMeta", mock.Anything, hashMatcher(winningTxHash), mock.Anything).
 		Return(&meta.Data{}, nil)
-	mockStore.On("Get", mock.Anything, hashMatcher(winningTxHash), []fields.FieldName{fields.Utxos, fields.ConflictingChildren}).
+	mockStore.On("Get", mock.Anything, hashMatcher(winningTxHash), []fields.FieldName{fields.Utxos, fields.ConflictingChildren, fields.DeletedChildren}).
 		Return(&meta.Data{}, nil)
 
 	u := &Server{utxoStore: mockStore}
@@ -88,7 +88,7 @@ func TestCheckCounterConflictingOnCurrentChain_StillRejectsMinedCounter(t *testi
 
 	mockStore.On("Get", mock.Anything, hashMatcher(conflictingTxHash), []fields.FieldName{fields.Tx}).
 		Return(&meta.Data{Tx: conflictingTx}, nil)
-	mockStore.On("Get", mock.Anything, hashMatcher(parentTxHash), []fields.FieldName{fields.Utxos}).
+	mockStore.On("Get", mock.Anything, hashMatcher(parentTxHash), []fields.FieldName{fields.Utxos, fields.DeletedChildren}).
 		Return(&meta.Data{SpendingDatas: []*spend.SpendingData{spend.NewSpendingData(&counterTxHash, 0)}}, nil)
 	mockStore.On("GetConflictingChildren", mock.Anything, counterTxHash).
 		Return([]chainhash.Hash{}, nil)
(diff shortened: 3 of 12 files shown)


ctnguyen (approved):
Review at commit [6e7f228](https://github.com/icellan/teranode/commit/6e7f2289e4be98f21049864b71f56024fb2899b2)

This is a careful, well-scoped fix, and it clearly benefits from the two prior attempts it distills. The part I was most worried about — the fail-closed partition — holds up: the probe correctly separates a genuinely-absent root ghost from a live spender whose BFS merely tripped on a missing descendant, which closes the exact hole in #1320's original "just skip on NOT_FOUND" prescription (a live, possibly same-chain-mined spender would otherwise have been dropped). The heal in `ProcessConflicting` verifies cleanly on the Aerospike/Lua path — passing the recorded ghost `SpendingData` back as the expected value lines up precisely with the Lua `unspend` ownership check (`bytes_equal(existingSpendingData, expectedSpendingData)` at `teranode.lua:520-521`), and the Aerospike integration test exercises that against the real UDF rather than a mock; the SQL side of that same claim is still unproven, which is the coverage gap I flag in ChiR3. Rollback monotonicity is honest: a cleared ghost slot stays cleared and the retry converges on an unspent slot. The residuals (the (288,576] pruned-mined band, the frozen-child-below-root-ghost gap) are scoped truthfully and match the issue's accepted tradeoffs. The items below are all non-blocking — one is a deviation from #1320's own acceptance criteria, the other two are coverage gaps.

## Non-blocking issues

### ChiR1 — Tolerated ghosts are counted but not logged, losing the identifiers #1320 asked for

**Problem:** #1320's acceptance criteria state "Every tolerated ghost emits a WARN + increments a prometheus counter" (identifying the parent txid:vout and the ghost txid). The implementation increments `prometheusUtxoCounterConflictingGhostSpends` (`process_conflicting.go:1106`) but emits no log. The walk's identifiers are available at the tolerate site, but `GetCounterConflictingTxHashes` throws them away:

```go
func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhash.Hash) ([]chainhash.Hash, error) {
	counterConflicting, _, err := getCounterConflictingTxHashesAndGhostSpends(ctx, s, txHash)
	return counterConflicting, err   // ghost slots dropped here
}
```

So the subtree-validation caller `checkCounterConflictingOnCurrentChain` — which *does* hold a logger (`u.logger`) — never sees which slot was tolerated.

**Why it matters:** A bare counter tells an operator *how many* ghosts were waved through, but not *which* `(parent,vout)` or ghost txid. In every mainnet incident in the lineage (#1213, #1214, #1320) the investigation turned on exactly those identifiers — to cross-check the write-side audit and to feed the `teranode-cli unspend` tool (#1323) the right slots. Tolerating a ghost is precisely the "store is inconsistent" signal the issue wanted made visible, and the counter alone can't drive a follow-up.

**Fix:** Surface the tolerated-ghost identifiers to a caller that has a logger and emit the one-line WARN. Minimal path: have the subtree layer log them (the ghost `Spend`s already carry `TxID`/`Vout` and the recorded `SpendingData.TxID`), e.g. thread the slots up rather than discarding them:

```go
// checkCounterConflictingOnCurrentChain / ProcessConflicting site
for _, g := range ghostSpends {
	u.logger.Warnf("[counterConflicting] tolerating confirmed ghost spender: parent %s:%d ghost %s",
		g.TxID, g.Vout, g.SpendingData.TxID)
}
```

### ChiR2 — Neither #1320 block-validation entry point is exercised; the guard is only unit-tested in isolation

**Problem:** `services/subtreevalidation/counter_conflicting_ghost_test.go` calls the guard function directly on a hand-assembled server with a mock store:

```go
u := &Server{utxoStore: mockStore}
err := u.checkCounterConflictingOnCurrentChain(ctx, winningTxHash, map[uint32]bool{7: true})
```

It never drives either path #1320's acceptance criteria name — "validates ... on both the legacy `HandleBlockDirect` and catchup `CheckBlockSubtreesRequest` paths." So nothing proves that a block referencing a ghost slot actually *reaches* the guard and clears through `validateSubtree` / `processTxMetaUsingStore` / `blessMissingTransaction`, which are the retry loops that wedged in the field.

**Why it matters:** This is the whole point of the fix — the wedge was never in the guard in isolation, it was in the block-validation entry points spinning on the guard's error forever. A regression that changed how those callers classify or route the guard's result (e.g. reintroducing the fatal `ProcessingError` wrap, or short-circuiting before the guard) would sail past this unit test while re-wedging production. The shared chokepoint keeps this below blocking — both wedge paths funnel through `GetCounterConflictingTxHashes`, which *is* covered — but for a consensus-sensitive unwedge the retrying entry points deserve at least one real pass.

**Fix:** Add a subtree-validation test that plants a ghost-referencing tx in a real store (sqlitememory, per project convention) and exercises it through one of the actual entry points, asserting the block validates instead of erroring:

```go
// drive HandleBlockDirect (or CheckBlockSubtreesRequest) over a subtree whose
// conflicting tx references a confirmed ghost slot; assert no ProcessingError
require.NoError(t, server.HandleBlockDirect(ctx, blockWithGhostReferencingTx))
```

### ChiR3 — The SQL backend's ghost heal is never exercised end-to-end

**Problem:** The resolution path (step-2 `Unspend` clearing the ghost slot via the ownership check, then the winner's step-3 spend) is verified end-to-end only against Aerospike (`aerospike/counter_conflicting_ghost_test.go`). The SQL test deliberately stops at walk tolerance and documents why:

```go
// stores/utxo/sql/counter_conflicting_ghost_test.go
// The full resolution flow (ProcessConflicting healing the contested slot) is
// covered against real Aerospike ... it cannot run here:
// SetConflicting opens a write transaction and then reads via the pool, which
// self-deadlocks on SQLite ...
```

The PR body asserts the recorded-spending-data trick "passes the ownership check on both backends," but on SQL that claim rests on reasoning alone — the SQL `Unspend` ownership semantics have no test asserting the ghost slot actually clears.

**Why it matters:** This is a consensus-relevant path, and SQL/sqlitememory is a supported store. If the SQL `Unspend` treats the expected `SpendingData` differently from the Lua check (rejects a mismatch, or ignores it and clears unconditionally in a way that diverges), the heal could silently fail or over-clear on a SQL-backed node with nothing to catch it. The whole-`ProcessConflicting` deadlock is a test-harness limitation, not a reason the ownership check itself can't be asserted.

**Fix:** Decouple the ownership-check assertion from the deadlocking orchestration — drive `store.Unspend` directly with the recorded ghost spending data and assert the slot clears (and that a *wrong* expected value is refused):

```go
ghostSlot := &utxo.Spend{TxID: parentTx.TxIDChainHash(), Vout: 0, UTXOHash: utxoHash0,
	SpendingData: spend.NewSpendingData(ghostTx.TxIDChainHash(), 0)}
require.NoError(t, store.Unspend(ctx, []*utxo.Spend{ghostSlot}, true))
resp, err := store.GetSpend(ctx, ghostSlot)
require.NoError(t, err)
require.Nil(t, resp.SpendingData, "ghost slot must clear once the recorded spend is presented as expected")
```

## Recap

| ChiR1 | Ghost tolerance not logged | | 40% |
| ChiR2 | Entry-path coverage missing | | 38% |
| ChiR3 | SQL heal path untested | | 35% |

ordishs (approved):
Approved.

Solid, well-tested fix for the two-stage ghost-spender wedge (#1320). The fail-closed partitioning is genuinely careful — probing a NOT_FOUND spender before tolerating it correctly closes the same-chain double-spend hole in the issue's original prescription, and the reaped-vs-ghost discrimination via `deletedChildren` is applied symmetrically across both the depth-1 counter walk and the depth-N BFS. Test coverage is excellent: real-Aerospike end-to-end heal, depth-2 descendant, reaped-mined fail-closed, plus the full fail-closed unit partition.

## Note: the SQL reaped-mined gate gap is moot in practice

My one earlier reservation was that `deletedChildren` is Aerospike-only, so on the SQL backend the reaped-mined fail-closed gate never fires and a settled/reaped mined spender could be mistaken for a ghost. Per maintainer confirmation, **SQL utxo stores are not a production backend** — production runs exclusively on Aerospike, where the marker is maintained and the guard is effective. The concern is therefore not a blocker. Recommend a one-line note in the "Accepted residuals" section stating the reaped-mined guard is Aerospike-only and SQL is not a supported conflict-resolution backend, so the assumption is recorded for future readers.

## Minor (non-blocking) follow-ups
- `reapedByParent` uses first-writer-wins when a child is reachable from two enqueuing parents; deterministic within a run, but a short comment ("any listing parent should win") would harden it if semantics ever tighten.
- `newGhostSlotSpend` returns an error tagged `[GetCounterConflictingTxHashes]` — retag with the actual function name for cleaner logs.
- Frozen-child check skipped for a root ghost's subtree — already acknowledged as a narrow residual.

I did not independently run the suite; relying on the reported green `go test -race` on the affected packages.

oskarszoon (approved):
Re-reviewed at 38089f5c — approving. The double-spend I flagged is fixed on the aerospike (production) path, verified against the code across passes: the `deletedChildren` discriminator fails closed when a probe-absent spender is a reaped-mined child rather than a true ghost, the pruner records every reaped child, and that map is never evicted, so it holds. That logic is unchanged in the latest push.

The latest push also caught a second issue in the same family that's worth calling out: the frozen / coinbase-placeholder sentinel is a record-less pseudo-hash, so a `Get` on it returns NOT_FOUND and the new ghost tolerance would have swallowed it — defeating frozen-tx rejection during conflict resolution. The fix keeps the sentinel in the result set so the frozen-child checks still fire but never recurses into it, and it's covered by new tests. That same change also removed the unmocked `Get` that was panicking the suite, so CI is green now.

Scope: this is an aerospike-only prod hotfix, which is the right call — the discriminator is aerospike-only by design and production runs on aerospike. Worth a one-line note in the PR (or a tracked follow-up) that the SQL path is intentionally out of scope, since `DeletedChildren` isn't populated there and the reaped-mined case is untested/undefined on SQL — just so it's a conscious decision rather than a silent gap.

One minor thing, non-blocking: worth confirming the pruner writes the parent's `deletedChildren` before (or atomically with) the child delete, so a counter-walk running concurrently with pruning can't briefly observe "child gone, parent not yet listed" and mis-tolerate. Narrow transient race, not the original permanent window.

Core fix is correct on the path that matters and CI is green. Approving.

oskarszoon:
Tested on **v0.15.6-beta-1**: my mainnet instance that was wedged on dangling conflicts recovered — the stuck slots cleared and it's back in sync with the chain tip.

PR #1327, opened 2026-07-22
fix(blockvalidation): run contextual header checks synchronously before optimistic AddBlock

### LINKED ISSUE #1149: [TN-010][low] blockvalidation: optimistic-mining background Valid() uses pre-AddBlock MTP snapshot; 2h check at background time
**Severity:** low · **Confidence:** medium · **Pass:** consensus/concurrency
**Component:** `services/blockvalidation/BlockValidation.go`, `model/Block.go`

### Summary
With optimistic mining (default ON), a block is added/mined before `block.Valid()` runs in a background goroutine that reuses the pre-`AddBlock` header snapshot for MTP and runs the `now+2h` future check at background-execution time. A too-far-future / MTP-violating block is transiently accepted then reverted. The revert path itself was reviewed and looks complete.

### Evidence
- `BlockValidation.go:1526-1544` — background goroutine reuses `blockHeaders` captured at `:1335`/`:1370` (before `AddBlock` at `:1510`); only `blockHeaderIDs` re-fetched.
- `model/Block.go:424-432` — `now+2h` check uses `time.Now()` inside `Valid()`.

### Why it matters
A block a non-optimistic validator would reject up front can be transiently accepted/mined then reverted — wasted hashpower / transient bad tip, not permanent corruption.

### Suggested fix
Run cheap header-only checks (2h-future, MTP, PoW, nBits) synchronously before `AddBlock`; leave expensive subtree/tx checks for the background goroutine.

### Suggested tests
Submit a block with `Timestamp = now+3h` under optimistic mining → assert not left on the canonical chain; 2h check evaluated at receipt time.

---

liam:
## Summary

Closes #1149.

Under optimistic mining (default on) a block is added and mined before `block.Valid()` runs in a background goroutine. That goroutine evaluated the 2-hours-in-the-future timestamp check (and the median-time-past and block-version-floor checks) against a `time.Now()` that had drifted past block-receipt time, so a too-far-future or MTP-violating block could be transiently accepted and mined, then reverted, instead of being rejected up front like a non-optimistic validator would.

## Fix

- **`model/Block.go`**: factor the header checks whose outcome depends on wall-clock time or the current chain - the 2-hours-in-the-future bound, median-time-past, and BIP34/66/65 version floor - out of `block.Valid()` into a new `Block.CheckHeaderContextual` method. `Valid()` still calls it, so it stays a single source of truth and the non-optimistic path is unchanged.
- **`services/blockvalidation/BlockValidation.go`**: call `CheckHeaderContextual` synchronously in the optimistic path **before** `AddBlock`, storing the block invalid and returning on failure.
- Proof-of-work and nBits are already checked synchronously by the service before the optimistic add, so they stay put and are not repeated.

This is the subset of `Valid()`'s header checks that must be evaluated at receipt time; the expensive subtree/tx checks remain in the background goroutine.

## Tests

- `model.TestBlock_CheckHeaderContextual` - unit-covers the extracted checks: a `now+3h` timestamp is rejected at call time, a just-under-2h timestamp is accepted, and an MTP violation is rejected when generate is not supported.
- `blockvalidation.TestBlockValidation_OptimisticMining_RejectsFutureTimestampSynchronously` - drives a `now+3h` block through the optimistic path and asserts it is rejected synchronously (`ErrBlockInvalid`, "two hours in the future") and that the optimistic background goroutine never starts.

The service test is load-bearing: without the fix `ValidateBlock` returns nil and the block is added then reverted in the background.

## Verification

- `go test ./model/ ./services/blockvalidation/` - both suites pass (blockvalidation full run ~274s)
- New tests pass under `-race`
- `go vet` clean; `gci` / `gofmt` clean

### DIFF
--- a/model/Block.go
+++ b/model/Block.go
@@ -500,32 +500,30 @@ type SubtreeStore interface {
 	GetIoReader(ctx context.Context, key []byte, fileType fileformat.FileType, opts ...options.FileOption) (io.ReadCloser, error)
 }
 
-func (b *Block) Valid(ctx context.Context, logger ulogger.Logger, subtreeStore SubtreeStore, txMetaStore utxo.Store, oldBlockIDsMap *txmap.SyncedMap[chainhash.Hash, []uint32],
-	currentChain []*BlockHeader, currentBlockHeaderIDs []uint32, settings *settings.Settings, metaRegenerator SubtreeMetaRegeneratorI) (bool, error) {
-	ctx, _, deferFn := tracing.Tracer("block").Start(ctx, "Valid",
-		tracing.WithHistogram(prometheusBlockValid),
-		tracing.WithLogMessage(logger, "[Block:Valid] called for %s", b.Header.String()),
-	)
-	defer deferFn()
-
-	// 1. Check that the block header hash is less than the target difficulty.
-	headerValid, _, err := b.Header.HasMetTargetDifficulty()
-	if err != nil {
-		return false, errors.NewProcessingError("[BLOCK][%s] error checking target difficulty", b.String(), err)
-	}
-
-	if !headerValid {
-		return false, errors.NewBlockInvalidError("[BLOCK][%s] block header hash is not less than the target difficulty", b.String())
-	}
-
+// CheckHeaderContextual runs the header checks whose outcome depends on wall-clock time
+// (time.Now()) or on the current chain: the 2-hours-in-the-future timestamp bound, the
+// median-time-past rule, and the mandatory block-version floor (BIP34/66/65). It mirrors the
+// timestamp/version portion of bitcoin-sv's ContextualCheckBlockHeader (proof-of-work and nBits
+// are checked separately by the block-validation service, so they are not repeated here).
+//
+// It is the subset of Valid()'s header checks that must be evaluated at block-receipt time. Under
+// optimistic mining Valid() runs in a background goroutine after the block has already been added,
+// so evaluating the 2-hours-in-the-future check there uses a drifted time.Now() and can transiently
+// accept a too-far-future block; the block-validation service calls this synchronously before the
+// optimistic AddBlock to close that gap (issue #1149). Valid() also calls it so the check remains a
+// single source of truth.
+//
+// currentChain is the run of previous headers ending at the block's parent; when empty the
+// median-time-past rule is skipped (same as Valid()). Returns nil when the header passes.
+func (b *Block) CheckHeaderContextual(currentChain []*BlockHeader, settings *settings.Settings, logger ulogger.Logger) error {
 	// 2. Check that the block timestamp is not more than two hours in the future.
 	twoHoursToTheFutureTimestampUint32, err := safeconversion.Int64ToUint32(time.Now().Add(2 * time.Hour).Unix())
 	if err != nil {
-		return false, errors.NewProcessingError("[BLOCK][%s] failed to convert two hours to the future timestamp to uint32", b.String(), err)
+		return errors.NewProcessingError("[BLOCK][%s] failed to convert two hours to the future timestamp to uint32", b.String(), err)
 	}
 
 	if b.Header.Timestamp > twoHoursToTheFutureTimestampUint32 {
-		return false, errors.NewBlockInvalidError("[BLOCK][%s] block timestamp is more than two hours in the future", b.String())
+		return errors.NewBlockInvalidError("[BLOCK][%s] block timestamp is more than two hours in the future", b.String())
 	}
 
 	// 3. Check that the median time past of the block is after the median time past of the last 11 blocks.
@@ -549,19 +547,19 @@ func (b *Block) Valid(ctx context.Context, logger ulogger.Logger, subtreeStore S
 		// calculate the median timestamp
 		medianTimestamp, err := CalculateMedianTimestamp(prevTimeStamps)
 		if err != nil {
-			return false, err
+			return err
 		}
 
 		b.medianTimestamp, err = safeconversion.Int64ToUint32(medianTimestamp.Unix())
 		if err != nil {
-			return false, err
+			return err
 		}
 
 		// validate that the block's timestamp is after the median timestamp
 		if b.Header.Timestamp <= b.medianTimestamp {
 			// if we're not on a chain that allows blocks to be generated quickly then return an error
 			if !settings.ChainCfgParams.GenerateSupported {
-				return false, errors.NewBlockInvalidError("block timestamp %d is not after median time past of last %d blocks %d", b.Header.Timestamp, pruneLength, medianTimestamp.Unix())
+				return errors.NewBlockInvalidError("block timestamp %d is not after median time past of last %d blocks %d", b.Header.Timestamp, pruneLength, medianTimestamp.Unix())
 			}
 			// otherwise just warn
 			logger.Warnf("block timestamp %d is not after median time past of last %d blocks %d", b.Header.Timestamp, pruneLength, medianTimestamp.Unix())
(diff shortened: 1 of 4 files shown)


ctnguyen (approved):
Review at commit [8c2e5b5](https://github.com/liam/teranode/commit/8c2e5b54fb3570011bb7d91425ba00e78aba3591)

This is a tidy, well-scoped fix that does exactly what #1149 asked for. Pulling the wall-clock/chain-dependent header checks into `CheckHeaderContextual` and running them synchronously before the optimistic `AddBlock` closes the transient-accept window cleanly, and keeping `Valid()` as a caller means there's still a single source of truth — the non-optimistic path is genuinely untouched. The reasoning about the MTP snapshot holds up: because a block's median-time-past is a function of its fixed ancestors, evaluating it against the pre-`AddBlock` snapshot at receipt time is the correct point, so both halves of the reported problem (the drifted 2h check and the MTP-snapshot accept/revert) are addressed rather than just the timestamp one. The doc comments carry the "why", and the two new tests are targeted — the service test in particular is load-bearing and proves the rejection happens before the goroutine starts. Only a couple of small test-coverage notes below; nothing blocks merge.

## Non-blocking issues

### ChiR1 - No synchronous-MTP service coverage
**Problem:** The linked issue names two accept/revert paths — the drifted 2h-future check *and* the MTP-snapshot violation — but the service-level regression test (`TestBlockValidation_OptimisticMining_RejectsFutureTimestampSynchronously`) only drives the future-timestamp case, and it stubs `GetBlockHeaders` as empty (`[]*model.BlockHeader{}`), which is exactly the input for which `CheckHeaderContextual` *skips* MTP. So nothing at the service level proves an MTP violation is rejected synchronously before `AddBlock`; that behaviour rests on the unit test alone, which doesn't exercise the optimistic path.
**Why it matters:** The MTP half of #1149 is the part where a stale snapshot most plausibly reintroduces the bug (e.g. a future refactor passing the wrong header slice into the synchronous call). Without a service test using a non-empty parent snapshot, a regression on the optimistic-path MTP wiring would pass CI.
**Fix:** Add a sibling service test that supplies a non-empty parent header snapshot whose median makes the block's timestamp an MTP violation, on params with `GenerateSupported == false`, and assert synchronous rejection before the optimistic add:
```go
mainParams := chaincfg.MainNetParams
tSettings.ChainCfgParams = &mainParams
now := uint32(time.Now().Unix())
parents := make([]*model.BlockHeader, 11)
for i := range parents { parents[i] = &model.BlockHeader{Timestamp: now} }
mockBlockchain.On("GetBlockHeaders", mock.Anything, mock.Anything, mock.Anything).
    Return(parents, make([]*model.BlockHeaderMeta, 11), nil)
// block timestamp == median (well inside the 2h bound) -> MTP violation, not future-timestamp
...
err = bv.ValidateBlock(ctx, block, "test", false)
require.True(t, errors.Is(err, errors.ErrBlockInvalid))
require.Contains(t, err.Error(), "median time past")
mockBlockchain.AssertNotCalled(t, "GetBlockHeaderIDs", mock.Anything, mock.Anything, mock.Anything)
```

### ChiR2 - Version-floor branch untested
**Problem:** `CheckHeaderContextual` now owns the BIP34/66/65 version-floor check (`CheckBlockVersion`), but the new `TestBlock_CheckHeaderContextual` never exercises it — `buildBlock` intentionally pins `Version: 1` at height 1 so the version check never fires, and the MTP subtest also stays below `BIP34Height`. So the one branch newly routed through this method has no direct coverage here.
**Why it matters:** The method is the synchronous gate for the optimistic path; a future refactor that broke the version branch (e.g. dropped the error wrap or the ordering that gives svnode-compatible error priority) would pass this suite silently.
**Fix:** Add a subtest that drives the version floor, e.g. mainnet params at a post-activation height with a too-low version:
```go
t.Run("outdated block version is rejected", func(t *testing.T) {
    tSettings := test.CreateBaseTestSettings(t)
    mainParams := chaincfg.MainNetParams
    tSettings.ChainCfgParams = &mainParams
    block := buildBlock(t, uint32(time.Now().Unix())) // version 1
    block.Height = 700000 // past BIP34/66/65 activation
    err := block.CheckHeaderContextual([]*BlockHeader{}, tSettings, logger)
    require.Error(t, err)
    require.Contains(t, err.Error(), "block version")
    require.True(t, errors.Is(err, errors.ErrBlockInvalid))
})
```

### ChiR3 - Synchronous-rejection assertion relies on a proxy
**Problem:** `TestBlockValidation_OptimisticMining_RejectsFutureTimestampSynchronously` proves "not optimistically added" via `AssertNotCalled(GetBlockHeaderIDs)`. That works, but it's indirect: `AddBlock` is mocked with a single matcher that matches both the optimistic accept path and the `storeInvalidBlock` (WithInvalid) path, so the `AddBlock` mock itself can't distinguish "added to chain" from "stored invalid". The issue's suggested assertion was more direct — that the block is not left on the canonical chain.
**Why it matters:** If a regression optimistically added the block but still returned an error, the goroutine-start proxy is a weaker signal than checking `AddBlock` directly; and simply asserting the invalid store fired would not prove a canonical add did *not* also happen first.
**Fix:** Split the `AddBlock` expectation by option so the test enforces "invalid store happened, canonical add did not". Match the invalid store explicitly and forbid any non-invalid add — for example register a matcher that inspects the variadic options for `WithInvalid(true)`, `AssertCalled` on it, and `AssertNotCalled` (or assert total `AddBlock` count `== 1`) for the canonical variant:
```go
isInvalid := func(opts ...blockchain.AddBlockOption) bool { /* true iff opts carry WithInvalid(true) */ }
mockBlockchain.On("AddBlock", mock.Anything, block, mock.Anything,
    mock.MatchedBy(func(o []blockchain.AddBlockOption) bool { return isInvalid(o...) })).Return(nil)
// after ValidateBlock:
mockBlockchain.AssertCalled(t, "AddBlock", mock.Anything, block, mock.Anything,
    mock.MatchedBy(func(o []blockchain.AddBlockOption) bool { return isInvalid(o...) }))
mockBlockchain.AssertNotCalled(t, "AddBlock", mock.Anything, block, mock.Anything,
    mock.MatchedBy(func(o []blockchain.AddBlockOption) bool { return !isInvalid(o...) }))
```
keeping the existing `AssertNotCalled(GetBlockHeaderIDs)` as a belt-and-suspenders check.

## Recap

| ChiR1 | No synchronous-MTP service test |  | 25% |
| ChiR2 | Version-floor branch untested |  | 20% |
| ChiR3 | Rejection asserted via proxy |  | 12% |

ordishs (approved):
Approve with comments.

The fix is correct, minimal, and well-tested. Extracting the wall-clock/chain-dependent header checks into `CheckHeaderContextual` and running them synchronously before the optimistic `AddBlock` closes the transient-acceptance gap cleanly, and keeping `Valid()` delegating to it preserves a single source of truth. The moved logic is byte-for-byte identical, so the non-optimistic path is genuinely unchanged.

**Non-blocking suggestions:**

1. **Performance / fail-fast (minor):** The synchronous `CheckHeaderContextual` call sits inside the `if useOptimisticMining` block, which is *after* `validateBlockSubtrees`. So a far-future block still pays for full subtree/tx validation before this near-free check rejects it. Its inputs (`blockHeaders`) are already available earlier, and PoW/nBits gate the path so the DoS exposure is bounded — but running it right after the PoW/nBits block (before `validateBlockSubtrees`) would fail-fast on a doomed block and better match the "reject up front" intent. Not a regression (matches the non-optimistic ordering).

2. **Nit:** `CheckHeaderContextual(currentChain, settings, logger)` puts `logger` last while `Valid()` takes it early — slight convention jar.

3. **Nit:** `storeInvalidBlock` records `err.Error()` (the specific inner reason) while the returned error wraps with a generic message — intentional, just flagging.

**Verified fine:** MTP logic moved verbatim using the same `blockHeaders` snapshot the background `Valid()` consumes (no new correctness surface); `b.medianTimestamp` mutation is race-free since the goroutine starts only after the sync call returns; transient/processing errors correctly avoid `storeInvalidBlock`; `IsRevalidation` respected.

The `GetBlockHeaderIDs` AssertNotCalled proxy in the service test is a nice way to prove the optimistic background goroutine never starts.

PR #1244, opened 2026-07-08
fix(util): avoid input aliasing in BuildMerkleRootFromCoinbase

liam:
## Problem

`BuildMerkleRootFromCoinbase` built each step's concatenation with:

```go
concat := append(acc, branch...)
```

On the **first** iteration `acc` still aliases the caller's `coinbaseHash` backing array. If that slice has spare capacity (`cap > len`), `append` writes `branch` into the backing array **in place** instead of allocating - silently corrupting the caller's buffer (and, in consensus-critical merkle code, risking a wrong root for whoever reuses that buffer).

Reported as [TN-004] in #1143.

## Fix

Build the concatenation in a freshly allocated buffer:

```go
concat := make([]byte, 0, len(acc)+len(branch))
concat = append(concat, acc...)
concat = append(concat, branch...)
```

This is **latent, not active**: every current caller passes a length-32, capacity-32 slice (from `chainhash.Hash[:]`), so `append` always reallocates and no input is mutated today. The change just removes the hazard for any future caller passing a spare-capacity slice. No behavior change for existing callers (the existing block-797330 and coinbase-only root tests still pass).

## Test

`TestBuildMerkleRootFromCoinbaseDoesNotMutateInput` passes the coinbase hash as a spare-capacity slice and asserts (a) the merkle root is still correct and (b) the caller's backing array - including its spare capacity - is untouched. Verified: it fails on the mutation assertion without the fix and passes with it.

Closes #1143

PR #1328, opened 2026-07-23
fix(alert): bounds-check parent tx outputs before indexing in node handlers

liam:
## Problem

The alert-service `Node` handlers indexed `parentTxMeta.Tx.Outputs[...]` with a vout taken from the incoming alert message, without bounds-checking it against the parent transaction's output count, and dereferenced `parentTxMeta.Tx` without a nil check:

- **`AddToConsensusBlacklist`** (`services/alert/node.go`) - indexed `Outputs[fund.TxOut.Vout]`, where `Vout` comes from the fund in the alert message.
- **`AddToConfiscationTransactionWhitelist`** - indexed `Outputs[txIn.PreviousTxOutIndex]` at three sites, where `PreviousTxOutIndex` comes from a confiscation transaction parsed from attacker-supplied hex.

A vout beyond the parent's outputs, or a parent whose `Tx` field is unpopulated, panics with `index out of range` / nil-pointer, taking down the alert-service goroutine. Same unguarded-index class as the fixes for #1283 / #1244.

Reachability is gated behind alert-signature verification (these methods dispatch only after the alert library verifies the message against configured alert keys), so this is a robustness / defense-in-depth fix rather than an anonymous DoS - but a validly-signed alert referencing a vout the parent doesn't have (operator error, or a pruned/mismatched parent) still crashes the service today.

## Fix

Guard both sites: on a nil `Tx` or an out-of-range index, append to the existing `NotProcessed` response with a `"parent tx output N not found"` reason instead of indexing. One guard in the confiscation handler covers all three uses. The consensus-blacklist site now also indexes with the already-validated `vout` (uint32) for consistency.

## Tests

Two regression tests (`TestNode_AddToConsensusBlacklist_OutOfRangeVout`, `TestNode_AddToConfiscationTransactionWhitelist_OutOfRangeInputIndex`) drive an out-of-range vout (99 against a 10-output parent) through the real `sqlitememory` store and assert a graceful `NotProcessed` response. Both panic with `index out of range [99] with length 10` when the guard is removed, and pass with it.

## Verification

- `go test -tags testtxmetacache ./services/alert/` - pass
- `go test -race` on the affected tests - pass
- `go vet ./services/alert/` - clean
- `gofmt` - clean

### DIFF
--- a/services/alert/node.go
+++ b/services/alert/node.go
@@ -305,8 +305,14 @@ func (n *Node) AddToConsensusBlacklist(ctx context.Context, funds []models.Fund)
 			continue
 		}
 
+		// guard against a nil parent tx or an out-of-range vout before indexing the outputs
+		if parentTxMeta.Tx == nil || int(vout) >= len(parentTxMeta.Tx.Outputs) {
+			response.NotProcessed = append(response.NotProcessed, n.getAddToConsensusBlacklistResponse(fund, errors.NewError("parent tx output %d not found", vout))...)
+			continue
+		}
+
 		// calculate the utxo hash from the output script
-		utxoHash, err := util.UTXOHashFromOutput(parentTxMeta.Tx.TxIDChainHash(), parentTxMeta.Tx.Outputs[fund.TxOut.Vout], vout)
+		utxoHash, err := util.UTXOHashFromOutput(parentTxMeta.Tx.TxIDChainHash(), parentTxMeta.Tx.Outputs[vout], vout)
 		if err != nil {
 			response.NotProcessed = append(response.NotProcessed, n.getAddToConsensusBlacklistResponse(fund, err)...)
 			continue
@@ -405,6 +411,12 @@ func (n *Node) AddToConfiscationTransactionWhitelist(ctx context.Context, txs []
 			parentTxMeta, err := n.utxoStore.Get(ctx, txIn.PreviousTxIDChainHash(), fields.Tx)
 			if err != nil {
 				response.NotProcessed = append(response.NotProcessed, n.getAddToConfiscationTransactionWhitelistResponse(tx.TxIDChainHash().String(), err)...)
+				continue
+			}
+
+			// guard against a nil parent tx or an out-of-range input index before indexing the outputs
+			if parentTxMeta.Tx == nil || int(txIn.PreviousTxOutIndex) >= len(parentTxMeta.Tx.Outputs) {
+				response.NotProcessed = append(response.NotProcessed, n.getAddToConfiscationTransactionWhitelistResponse(tx.TxIDChainHash().String(), errors.NewError("parent tx output %d not found", txIn.PreviousTxOutIndex))...)
 				continue
 			}
 
--- a/services/alert/node.go
+++ b/services/alert/node.go
@@ -305,8 +305,11 @@ func (n *Node) AddToConsensusBlacklist(ctx context.Context, funds []models.Fund)
 			continue
 		}
 
-		// guard against a nil parent tx or an out-of-range vout before indexing the outputs
-		if parentTxMeta.Tx == nil || int(vout) >= len(parentTxMeta.Tx.Outputs) {
+		// guard against a nil parent tx, an out-of-range vout, or a nil output element before
+		// indexing the outputs (external outputs-only parents can have nil holes at in-range indices)
+		if parentTxMeta.Tx == nil ||
+			uint64(vout) >= uint64(len(parentTxMeta.Tx.Outputs)) ||
+			parentTxMeta.Tx.Outputs[vout] == nil {
 			response.NotProcessed = append(response.NotProcessed, n.getAddToConsensusBlacklistResponse(fund, errors.NewError("parent tx output %d not found", vout))...)
 			continue
 		}
@@ -414,8 +417,11 @@ func (n *Node) AddToConfiscationTransactionWhitelist(ctx context.Context, txs []
 				continue
 			}
 
-			// guard against a nil parent tx or an out-of-range input index before indexing the outputs
-			if parentTxMeta.Tx == nil || int(txIn.PreviousTxOutIndex) >= len(parentTxMeta.Tx.Outputs) {
+			// guard against a nil parent tx, an out-of-range input index, or a nil output element before
+			// indexing the outputs (external outputs-only parents can have nil holes at in-range indices)
+			if parentTxMeta.Tx == nil ||
+				uint64(txIn.PreviousTxOutIndex) >= uint64(len(parentTxMeta.Tx.Outputs)) ||
+				parentTxMeta.Tx.Outputs[txIn.PreviousTxOutIndex] == nil {
 				response.NotProcessed = append(response.NotProcessed, n.getAddToConfiscationTransactionWhitelistResponse(tx.TxIDChainHash().String(), errors.NewError("parent tx output %d not found", txIn.PreviousTxOutIndex))...)
 				continue
 			}
--- a/services/alert/node_test.go
+++ b/services/alert/node_test.go
@@ -287,3 +287,102 @@ func TestNode_getAddToConsensusBlacklistResponse(t *testing.T) {
 		require.Nil(t, utxoSpend.SpendingData)
 	})
 }
+
+// TestNode_AddToConsensusBlacklist_OutOfRangeVout ensures a fund referencing a
+// vout beyond the parent tx's output count is rejected gracefully rather than
+// panicking with an index-out-of-range on parentTxMeta.Tx.Outputs.
+func TestNode_AddToConsensusBlacklist_OutOfRangeVout(t *testing.T) {
+	ctx := context.Background()
+
+	tSettings := test.CreateBaseTestSettings(t)
+	logger := ulogger.NewErrorTestLogger(t)
+
+	utxoStoreURL, err := url.Parse("sqlitememory:///test")
+	require.NoError(t, err)
+
+	utxoStore, err := sql.New(ctx, logger, tSettings, utxoStoreURL)
+	require.NoError(t, err)
+
+	_ = utxoStore.SetBlockHeight(101)
+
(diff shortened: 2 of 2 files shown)


ordishs (approved):
## Approve

Clean, minimal defensive fix. Verified locally.

**Correctness**
- Guard placement is right: in `AddToConsensusBlacklist` the guard sits after the `safeconversion.IntToUint32` of `vout` and uses the validated value; in `AddToConfiscationTransactionWhitelist` the single guard at the top of the input loop correctly dominates all three downstream `Outputs[...]` uses (satoshis check, old-UTXO hash, amended-output satoshis).
- Nil check short-circuits before the `len()` deref, so the guard can't panic itself.
- The `Outputs[fund.TxOut.Vout]` → `Outputs[vout]` change is a real consistency improvement — index and the third arg to `UTXOHashFromOutput` now use the same validated `uint32`.
- Confirmed the `tx` fixture has 10 outputs, so vout 99 is a genuine out-of-range case. Ran both new tests under `-race`: pass.

**Conventions** — `errors.NewError`, `require`, real `sqlitememory` store, `testtxmetacache` tag — all in line with the project guardrails.

**Minor (non-blocking)**
- `int(vout) >= len(...)`: on a 32-bit build a `uint32` above 2^31 would wrap negative and bypass the guard. Teranode is 64-bit-only in practice, but `uint64(vout) >= uint64(len(...))` would be unconditionally correct at no cost. Same for `txIn.PreviousTxOutIndex`.
- Per-input `continue` yields one `NotProcessed` entry per bad input rather than per tx — consistent with the existing error paths in that loop, just noting it.

Risk is low: additive guards on error paths, no success-path behavior change. Tests demonstrably panic without the guard and pass with it.

ctnguyen (changes requested):
Review at commit [dc1e6f5](https://github.com/liam/teranode/commit/dc1e6f5dd1741e2b96735c63f1b4988739add298)

Good instinct and mostly the right fix — the out-of-range index and nil-`Tx` derefs were real crash paths, one guard sensibly covers all three confiscation sites, and the regression tests are the honest kind (they genuinely panic without the guard). It also quietly cleans up the `fund.TxOut.Vout`/`vout` indexing mismatch. The catch: the same guard leaves a sibling nil-pointer case open — an in-range output that is itself `nil` — which is reachable through the Aerospike external path and still panics. Details below.

## Blocking issues

### ChiR1 — In-range but nil output element still panics

**Problem:** The guards check only `int(idx) >= len(parentTxMeta.Tx.Outputs)`, i.e. that the index is within the slice. But `Tx.Outputs` can legitimately contain `nil` element pointers at in-range positions. When a parent is externally stored and only its outputs file is available (the `FileTypeOutputs` fallback), `getExternalTransaction` reconstructs `tx.Outputs = make([]*bt.Output, len(PadUTXOsWithNil(...)))` and populates only the still-present UTXOs (`stores/utxo/aerospike/get.go:1656-1667`); every spent/unstored index in between stays `nil`. `PadUTXOsWithNil` sizes the slice to `maxIndex+1` (`services/utxopersister/UTXOSet.go:816`) and fills only present indices (`services/utxopersister/UTXOSet.go:819-820`), so those nil holes sit at valid in-range offsets. An alert referencing such a vout passes the new guard and then dereferences a nil `*bt.Output`:

- Confiscation: `services/alert/node.go:412` reads `parentTxMeta.Tx.Outputs[txIn.PreviousTxOutIndex].Satoshis` → nil deref.
- Consensus: `util.UTXOHashFromOutput(..., parentTxMeta.Tx.Outputs[vout], vout)` → `UTXOHash(hash, vOut, output.LockingScript, output.Satoshis)` dereferences `output.LockingScript` at `util/utxo_hash.go:53`. The `lockingScript == nil` check at line 19 runs *after* `output` is already dereferenced, so it does not protect against a nil `output`.

This is the same crash class the PR sets out to close, in the exact lines it touches — a length check alone doesn't make the index safe.

**Why it matters:** The stated goal is that a malformed/mismatched-but-signed alert degrades to a `NotProcessed` entry instead of taking down the alert-service goroutine. As written it still panics on a reachable production input (Aerospike parent, tx file pruned, alert points at an already-spent sibling output), so the hardening is incomplete.

**Fix:** also reject a nil element, and bind the checked output once so both handlers use it:

```go
if parentTxMeta.Tx == nil ||
    uint64(vout) >= uint64(len(parentTxMeta.Tx.Outputs)) ||
    parentTxMeta.Tx.Outputs[vout] == nil {
    response.NotProcessed = append(response.NotProcessed, n.getAddToConsensusBlacklistResponse(fund, errors.NewError("parent tx output %d not found", vout))...)
    continue
}
out := parentTxMeta.Tx.Outputs[vout] // use out for the hash / satoshis
```

Mirror the same `... || parentTxMeta.Tx.Outputs[txIn.PreviousTxOutIndex] == nil` clause in the confiscation guard so the `.Satoshis` read and the two `UTXOHashFromOutput` calls all consume a checked, non-nil output. Add a regression test that drives this branch (see ChiR3 for why sqlitememory can't).

## Non-blocking issues

### ChiR2 — `int()` index compare can truncate on a 32-bit build (confiscation handler)

*Same as [@ordishs's point 1 (23-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1328#pullrequestreview-4763679814).*

**Problem:** The confiscation guard uses `int(txIn.PreviousTxOutIndex) >= len(...)`. `PreviousTxOutIndex` is a `uint32` parsed straight from attacker-supplied hex, so it spans the full 0..2³²-1 range; on a 32-bit build `int` is 32-bit and any value ≥ 2³¹ wraps negative, slips past `>=`, and re-opens the out-of-range panic. Note this does *not* apply to the consensus-blacklist path: `fund.TxOut.Vout` is an `int` field, so on a 32-bit build it can never exceed `MaxInt32`, and `safeconversion.IntToUint32` preserves that — `int(vout)` can't wrap there. Teranode is 64-bit-only in practice, so this is latent rather than live.

**Why it matters:** A guard whose whole job is to be unconditionally safe shouldn't have a build-width-dependent hole, and the correct form is free.

**Fix:** compare in a width that can't truncate — `uint64(txIn.PreviousTxOutIndex) >= uint64(len(parentTxMeta.Tx.Outputs))` (and likewise `uint64(vout) >= ...` in the consensus handler for consistency, as shown in ChiR1).

### ChiR3 — New guard branches aren't reachable through the test store

**Problem:** The two regression tests drive the out-of-range branch through the real `sqlitememory` store, which is good — but that store's `Get` builds a fully-populated, non-nil `Tx` for existing rows (`stores/utxo/sql/sql.go:1753-1754`) and returns an error for missing rows (`stores/utxo/sql/sql.go:1558`). So neither the `parentTxMeta.Tx == nil` branch nor the in-range-nil-element case (ChiR1) can ever be produced with it; those branches are asserted only by construction, not by test.

**Why it matters:** The nil-`Tx` and (once added) nil-element guards are the defensive core of this PR, yet nothing would go red if a later refactor dropped them.

**Fix:** cover them with a store that can actually yield the shape — a small fake/stub returning `meta.Data` with a nil `Tx` and with interior-nil `Outputs`, or the Aerospike external-outputs reconstruction path — and assert the same graceful `NotProcessed` reason.

## Recap

| ChiR1 | In-range nil output panics | YES | 55% |
| ChiR2 | 32-bit index truncation (confiscation) | | 15% |
| ChiR3 | Guard branches untested via store | | 10% |

liam:
Thanks @ctnguyen - ChiR1 was a real reachable panic, verified it before fixing. Pushed 43ddfd85f addressing all three:

**ChiR1 (blocking) - in-range nil output element.** Confirmed: the outputs-only external-fetch path (`aerospike/get.go:1656-1667`) builds `Outputs` with `make(...)` and fills only present indices, leaving in-range `nil` holes; and `UTXOHashFromOutput` (`util/utxo_hash.go:52-53`) reads `output.LockingScript`/`output.Satoshis` before the `lockingScript == nil` check at line 19, so a nil `output` panics. Both guards now also reject a nil element:

```go
if parentTxMeta.Tx == nil ||
    uint64(vout) >= uint64(len(parentTxMeta.Tx.Outputs)) ||
    parentTxMeta.Tx.Outputs[vout] == nil {
```

so the `.Satoshis` read and both `UTXOHashFromOutput` calls consume a checked, non-nil output.

**ChiR2 - index truncation.** Switched both compares to `uint64(...)` so a `uint32` above 2^31 can't wrap negative on a 32-bit build. Applied to the consensus handler too for consistency, as suggested.

**ChiR3 - untested guard branches.** Added `TestNode_AddToConsensusBlacklist_NilParentOutput` and `TestNode_AddToConfiscationTransactionWhitelist_NilParentOutput`, each with `nil Tx` and `in-range nil output element` subtests driven through a `MockUtxostore` (the real `sqlitememory` store always returns a fully-populated non-nil `Tx`, so it can't produce these shapes). Verified load-bearing: the in-range-nil subtests panic with `nil pointer dereference` when the nil-element clause is removed, and pass with it.

Verification: `go vet`, `go test`, `go test -race` on the alert package, and `gofmt` all clean.

ctnguyen (approved):
Review at commit [43ddfd8](https://github.com/liam/teranode/commit/43ddfd85f697ce718fd9805ef1a3a635f2141220)

This addresses everything cleanly. The in-range nil-element case I flagged as blocking is now closed in both guards — the `Outputs[idx] == nil` clause sits after the length check, so it short-circuits safely and the `.Satoshis` read plus both `UTXOHashFromOutput` calls now consume a checked, non-nil output. The width fix and the new coverage landed too, and the added tests are the honest kind: I confirmed the in-range-nil subtests are load-bearing (they panic without the clause), and that `bt.Tx.Clone()` allocates a fresh `Outputs` slice, so the `holeTx.Outputs[1] = nil` trick doesn't leak into the shared package-level fixture. Nothing new to raise.

## Status of previous points

| ChiR1 | In-range nil output panics | YES | FIXED |
| ChiR2 | 32-bit index truncation (confiscation) | | FIXED |
| ChiR3 | Guard branches untested via store | | FIXED |

PR #1330, opened 2026-07-24
fix(blockchain): refuse to invalidate the genesis block

liam:
## Problem

`SQL.InvalidateBlock` (`stores/blockchain/sql/InvalidateBlock.go`) runs a recursive descendant CTE rooted at the target hash:

```sql
WITH RECURSIVE children AS (
    SELECT ... WHERE hash = $1
    UNION
    SELECT ... INNER JOIN children c ON c.hash = b.previous_hash )
UPDATE blocks SET invalid = true, mined_set = false, on_main_chain = false
WHERE id IN (SELECT id FROM children)
```

Nothing at any layer (`services/rpc/handlers.go` handler → `services/blockchain/Server.go` gRPC → SQL store) checks the target against the genesis hash. Genesis is a real row, so an admin `invalidateblock` RPC on the genesis hash walks every descendant and marks the **entire chain** invalid + off-main-chain in a single statement.

Recovery is worse than bitcoind: `RevalidateBlock` updates only the single named block (`UPDATE ... WHERE hash = $1`, no recursion), so `reconsiderblock` cannot undo a chain-wide invalidation - you would have to reconsider every block individually.

`invalidateblock` is an authenticated operator RPC, so this is an operator-footgun / safety-rail gap rather than an anonymous vector - but the blast radius is maximal and unrecoverable via RPC, which is exactly why bitcoind and the Java reference node both forbid marking genesis.

Found via the Java-POC to Teranode comparative gap analysis (bitcoin-sv/teranode#4693), verified against current `main`.

## Fix

Add a genesis guard at the top of `SQL.InvalidateBlock`, before any DB work:

```go
if s.chainParams != nil && blockHash.IsEqual(s.chainParams.GenesisHash) {
    return nil, errors.NewProcessingError("cannot invalidate the genesis block")
}
```

and mirror it in the blockchain gRPC server (`services/blockchain/Server.go`) for defense in depth. Matches bitcoind's `MarkGenesisException` behaviour.

## Tests

- `TestSQLInvalidateBlock_GenesisGuard` - stores genesis → block1 → block2 → block3 (the regtest fixtures form exactly this chain), asserts genesis invalidation is rejected and every block stays valid. Load-bearing: with the store guard removed the test fails because the whole chain is invalidated.
- `TestServerInvalidateBlock_GenesisGuard` - covers the gRPC server path; verified the server-level guard fires independently (still rejects with the store guard disabled).

## Verification

- `go vet ./stores/blockchain/sql/ ./services/blockchain/` - clean
- `go test` both packages - pass
- `go test -race` on the new tests - pass
- `gofmt` - clean

### DIFF
--- a/services/blockchain/Server.go
+++ b/services/blockchain/Server.go
@@ -2301,6 +2301,13 @@ func (b *Blockchain) InvalidateBlock(ctx context.Context, request *blockchain_ap
 		return nil, errors.WrapGRPC(errors.NewBlockInvalidError("[Blockchain][InvalidateBlock] request's hash is not valid", err))
 	}
 
+	// Refuse to invalidate the genesis block. The store's recursive CTE would flip the
+	// whole chain invalid in one statement and it could not be undone via reconsiderblock.
+	// Guarded in the store too (defense in depth); reject early here with a clear error.
+	if b.settings != nil && b.settings.ChainCfgParams != nil && blockHash.IsEqual(b.settings.ChainCfgParams.GenesisHash) {
+		return nil, errors.WrapGRPC(errors.NewBlockInvalidError("[Blockchain][InvalidateBlock] cannot invalidate the genesis block"))
+	}
+
 	// invalidate block will also invalidate all child blocks
 	invalidatedHashes, err := b.store.InvalidateBlock(ctx, blockHash)
 	if err != nil {
--- a/services/blockchain/mtp_cache_integration_test.go
+++ b/services/blockchain/mtp_cache_integration_test.go
@@ -163,3 +163,16 @@ func TestMTPCacheIntegration_InvalidateBlockTruncatesCache(t *testing.T) {
 	_, ok = ctx.server.mtpCache.get(3)
 	require.False(t, ok, "height 3 (above invalidated) must be truncated from cache")
 }
+
+// TestServerInvalidateBlock_GenesisGuard verifies the gRPC server rejects an
+// attempt to invalidate the genesis block (defense-in-depth mirror of the guard
+// in the SQL store), so a stray admin invalidateblock call cannot wipe the chain.
+func TestServerInvalidateBlock_GenesisGuard(t *testing.T) {
+	ctx := setup(t)
+
+	_, err := ctx.server.InvalidateBlock(context.Background(), &blockchain_api.InvalidateBlockRequest{
+		BlockHash: ctx.server.settings.ChainCfgParams.GenesisHash.CloneBytes(),
+	})
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "cannot invalidate the genesis block")
+}
--- a/stores/blockchain/sql/InvalidateBlock.go
+++ b/stores/blockchain/sql/InvalidateBlock.go
@@ -62,6 +62,15 @@ import (
 func (s *SQL) InvalidateBlock(ctx context.Context, blockHash *chainhash.Hash) (invalidatedHashes []chainhash.Hash, err error) {
 	s.logger.Debugf("InvalidateBlock %s", blockHash.String())
 
+	// Refuse to invalidate the genesis block. Genesis is the root of the recursive
+	// descendant CTE below, so invalidating it would flip the entire chain to
+	// invalid/off-main-chain in a single statement, and RevalidateBlock (single block,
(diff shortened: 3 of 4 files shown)


ordishs (approved):
Approve. Correct, well-scoped fix for a real operator footgun.

- Store guard is placed before any DB work and before taking `slowPathMu`.
- Both guards are nil-safe (`chainParams`/`ChainCfgParams` checks; `IsEqual` tolerates nil).
- Defense-in-depth is load-bearing, not decorative: `LocalClient.InvalidateBlock` calls the store directly, bypassing the Server.go guard — so the store guard is the only protection on that path.
- No legitimate internal caller invalidates genesis, so the new reject path can't break a real flow.
- Both new tests pass locally, and the SQL test is load-bearing (fails if the store guard is removed).

Nits (non-blocking): the final loop uses `assert.False` where AGENTS.md prefers `require`; and the two layers return different error types (ProcessingError vs BlockInvalidError) — cosmetic since both share the matched substring.

ctnguyen (approved):
Review at commit [24f2f63](https://github.com/liam/teranode/commit/24f2f63a982a10d52945e54e92cf2d01676b669b)

Nice, focused safety fix — the problem is real and well-diagnosed, and the guard is placed in the right two spots. Rejecting in the store covers the `LocalClient` path that calls `store.InvalidateBlock` directly and bypasses the gRPC server, while the server guard covers the gRPC/RPC/HTTP paths, so the two layers cover genuinely different entry points rather than being redundant. `IsEqual` is nil-safe and the store test is load-bearing (block1's parent is the regtest genesis). Small diff, matches bitcoind's intent. Two non-blocking notes below.

## Non-blocking issues

### ChiR1 — Server-guard test doesn't isolate the server guard
**Problem:** `TestServerInvalidateBlock_GenesisGuard` builds its server via `setup(t)` (`services/blockchain/server_test.go:149`), which wires it to a real `sql` store that already carries its own genesis guard returning the identical `"cannot invalidate the genesis block"` message. If the server-level guard were deleted, the call would fall through to `b.store.InvalidateBlock` and produce the same error, so the test would still pass. It therefore can't tell the two guards apart and doesn't prove the server guard fires. (The PR text says the server guard was "verified to fire independently with the store guard disabled," but that verification isn't in the committed test.)
**Why it matters:** This test is the regression guard for the server-level defense-in-depth layer; as written it would not catch that layer's removal — the store guard would silently keep the test green.
**Fix:** Make reaching the store observable. A thin call-recording wrapper over the real sqlitememory store (a spy, not a behavioural mock, so it stays within the "use the real store" convention) proves the server returns first:
```go
type invalidateSpyStore struct {
    blockchain.Store
    called bool
}
func (s *invalidateSpyStore) InvalidateBlock(ctx context.Context, h *chainhash.Hash) ([]chainhash.Hash, error) {
    s.called = true
    return s.Store.InvalidateBlock(ctx, h)
}
// build the server around the spy, then:
require.Error(t, err)
require.Contains(t, err.Error(), "cannot invalidate the genesis block")
require.False(t, spy.called, "server guard must reject before reaching the store")
```

### ChiR2 — Genesis rejection surfaces two different application error codes
**Problem:** The store guard returns `errors.NewProcessingError` (code `ERR_PROCESSING`) while the server guard returns `errors.NewBlockInvalidError` (code `ERR_BLOCK_INVALID`). Both map to `codes.Internal` via the `default` case in `ErrorCodeToGRPCCode`, so the gRPC *status* is identical — but the application error code is carried across the gRPC boundary in the `TError` details and reconstructed on the client, so a caller that branches on `err.Code()` sees a different code depending on which guard fired (the server path vs the `LocalClient`/direct-store path). `BlockInvalidError` is also a slightly off label here: genesis is not an invalid block, the operation is being refused.
**Why it matters:** The same logical refusal produces two different reconstructed application codes across paths, and the label misrepresents the condition. Low impact, but trivially avoidable.
**Fix:** Use one error type in both guards; a processing/refusal error reads better than "block invalid":
```go
// services/blockchain/Server.go
return nil, errors.WrapGRPC(errors.NewProcessingError("[Blockchain][InvalidateBlock] cannot invalidate the genesis block"))
```

## Recap

| ChiR1 | Server test not isolated | | 35% |
| ChiR2 | Inconsistent error type | | 22% |

PR #1331, opened 2026-07-24
chore: define constants for duplicated string literals (S1192)

liam:
## What

Resolves the **104 open `go:S1192`** findings ("Define a constant instead of duplicating this literal N times"), flagged **HIGH** severity in SonarQube, across **45 files**.

Each duplicated string literal is now an unexported package-level constant, with every occurrence in the same file replaced by that constant.

## Why

`go:S1192` is the top HIGH-severity rule by count on the project. The fixes are individually trivial and collectively clear the whole rule.

## Approach

Purely mechanical and behaviour-preserving:
- Constant values are byte-identical to the literals they replace, so log/error messages, `Printf`/`Errorf` format strings, CLI flag names, RPC help text and UI labels are unchanged at runtime.
- Constants are scoped to the same package/file as the literal; names are distinct within each package.
- Overlapping-prefix literals (e.g. `"parameter #%d '%s' "` vs `"parameter #%d '%s' must "` in `bsvjson/cmdparse.go`, and `"invalid peer ID"` vs `"invalid peer ID: %v"` in `handle_catchup_metrics.go`) were handled by matching complete quoted tokens only, so no longer literal was corrupted.
- No logic changes, no `_test.go` files, and no generated code touched.

## Files

45 files across `cmd/`, `compose/`, `daemon/`, `services/` (rpc, legacy, blockassembly, blockchain, blockvalidation, p2p, asset, subtreevalidation, utxopersister), `stores/` (blockchain/sql, txmetacache, utxo/{aerospike,sql,factory}) and `util/`.

## Verification

- `go build` across all source trees (untagged) and `go build -tags aerospike ./stores/utxo/aerospike/...` - clean
- `go vet` on all modified packages - clean
- `gofmt -l` on all 45 files - clean
- Test files in all modified packages compile; ran representative package suites (`stores/txmetacache`, `util/merkleproof`, `services/rpc/bsvjson`, `services/blockchain`) - pass
- Spot-checked several literals to confirm each now appears exactly once (the constant definition)

### DIFF
--- a/cmd/compare-benchmarks/main.go
+++ b/cmd/compare-benchmarks/main.go
@@ -17,6 +17,11 @@ import (
 	"github.com/bsv-blockchain/teranode/errors"
 )
 
+const (
+	secOpUnit   = "sec/op"
+	vsBaseLabel = "vs base"
+)
+
 type comparison struct {
 	Name          string
 	BaselineValue string
@@ -120,21 +125,21 @@ func parseBenchstat(output string, threshold, alpha float64) []comparison {
 	noChangeRe := regexp.MustCompile(`~\s+\(p=(\d+\.?\d*)\s+n=\d+\)`)
 
 	scanner := bufio.NewScanner(strings.NewReader(output))
-	currentMetric := "sec/op"
+	currentMetric := secOpUnit
 
 	for scanner.Scan() {
 		line := scanner.Text()
 
 		// Detect metric headers (sec/op, B/op, allocs/op)
-		if strings.Contains(line, "sec/op") && strings.Contains(line, "vs base") {
-			currentMetric = "sec/op"
+		if strings.Contains(line, secOpUnit) && strings.Contains(line, vsBaseLabel) {
+			currentMetric = secOpUnit
 			continue
 		}
-		if strings.Contains(line, "B/op") && strings.Contains(line, "vs base") {
+		if strings.Contains(line, "B/op") && strings.Contains(line, vsBaseLabel) {
 			currentMetric = "B/op"
 			continue
 		}
-		if strings.Contains(line, "allocs/op") && strings.Contains(line, "vs base") {
+		if strings.Contains(line, "allocs/op") && strings.Contains(line, vsBaseLabel) {
 			currentMetric = "allocs/op"
 			continue
 		}
@@ -205,7 +210,7 @@ func parseBenchstat(output string, threshold, alpha float64) []comparison {
 			comp.PValue, _ = strconv.ParseFloat(m[2], 64)
 			comp.HasPValue = true
 
-			if currentMetric == "sec/op" {
+			if currentMetric == secOpUnit {
 				comp.Degraded = comp.PValue < alpha && comp.PercentChange > threshold
 				comp.Improved = comp.PValue < alpha && comp.PercentChange < -threshold
 			}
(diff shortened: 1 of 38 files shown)


ordishs (commented):
One small fix before merge — otherwise this is a clean, byte-identical S1192 sweep. 👍

**`stores/blockchain/sql/sql.go` (lines 50-55): the new constant landed inside an existing doc comment.**

```go
// chainWalkCacheTTL is the time-to-live for chain walk cache entries (GetBlockHeaderIDs,
// GetBlockHeaders). Set to 10 minutes because block validation in production can take
// several minutes, and the cached results (parent_id walks) are immutable.
const errCouldNotCreateBlocksTable = "could not create blocks table"

const chainWalkCacheTTL = 10 * time.Minute
```

The comment now documents `errCouldNotCreateBlocksTable`, leaving `chainWalkCacheTTL` undocumented. Please move `errCouldNotCreateBlocksTable` above the comment block (or below the `chainWalkCacheTTL` declaration) so the doc comment reattaches to its constant. This is the only spot in the PR where a new const was inserted into existing comment context — everywhere else the placement is clean.

ctnguyen (approved):
Review at commit [bff19d5](https://github.com/liam/teranode/commit/bff19d5d77090205c2d8990e669f270c2452f613)

This is a clean, careful mechanical refactor and the "byte-identical, behaviour-preserving" claim holds up. I checked the parts most likely to break — the concatenated format strings in `bsvjson/cmdparse.go`, the overlapping-prefix literals (`"parameter #%d '%s' "` vs `"...must "` vs `"...must be type "`, and `"invalid peer ID"` vs `"invalid peer ID: %v"`), the `unable to parse `/`data push of %d bytes encoded ` prefixes — and in every case the constant value, the verb count, and the argument order match the original exactly. Extracted literals genuinely appear once now, constants stay unexported and colocated with their use sites (package-local, not exported), and nothing leaks across packages. Nice work keeping a 45-file sweep this disciplined.

ordishs (approved):
Approving — the doc-comment fix in `08ed7cbcb` resolves the only outstanding item.

Confirmed on the current head:
- `chainWalkCacheTTL` doc comment is reattached to its constant; `errCouldNotCreateBlocksTable` moved below it, value byte-identical.
- `go vet ./stores/blockchain/sql/` → exit 0.

The full sweep is byte-identical across all 45 files (format verbs, bracket prefixes, trailing spaces, newlines all preserved), both overlapping-prefix files (`cmdparse.go`, `handle_catchup_metrics.go`) correctly use distinct constants, and no runtime string changed. Clean S1192 cleanup — good to merge. Optionally squash the fixup commit into the sweep commit on merge.

liam:
### Quality gate context (mechanical refactor)

This is a **pure, behaviour-preserving `go:S1192` sweep** - every change replaces a duplicated string literal with a package-level constant whose value is **byte-identical** to the literal. No logic, no test files, no generated code.

**Duplication condition — addressed.** The earlier `new_duplicated_lines_density` (4.1%) was not duplication introduced by this PR: the flagged lines were literal-substitutions that happened to fall *inside pre-existing duplicated blocks* (e.g. the many near-identical methods in `stores/txmetacache/improved_cache.go`), which SonarQube then re-attributes to "new code." I've dropped the sweep from the 7 files that contributed those lines (`improved_cache.go`, `grpc_helper.go`, `BlockAssembler.go`, `daemon_stores.go`, `blob_deletion.go`, `UTXOSet.go`, `SubtreeProcessor.go`), so this condition should now pass. The PR now covers 38 files.

**Coverage condition — requesting a maintainer override.** `new_coverage` (~40%) cannot be reasonably met by a mechanical constant-extraction sweep: the changed lines are literal substitutions inside existing (largely untested) error/log paths across cmd tools and services, and there is no proportionate way to add meaningful unit tests for a rename. The changes are byte-identical and cannot alter runtime behaviour, so coverage of these specific lines carries no correctness signal here.

Could a maintainer waive/override the coverage condition for this refactor-only PR? Happy to split it further or adjust if preferred.

PR #1335, opened 2026-07-24
fix(errors): strip orphaned %w in errors.New*, sweep call sites, add CI guard (#1332)

### LINKED ISSUE #1332: blockchain/sql: coinbase-height store error renders "%!w(MISSING)" (literal %w in errors.New* message)
**Component:** `stores/blockchain/sql/StoreBlock.go:922` · `errors/errors.go` (`New`)

### Summary

Error messages built with a literal trailing `: %w` in the teranode `errors.New*` constructors render as `%!w(MISSING)` (or a raw literal `%w`), because the custom errors package strips the trailing `error` param *before* calling `fmt.Errorf`, leaving the `%w` verb unsatisfied.

```
STORAGE_ERROR (69): failed to extract coinbase height for block 00000000b0d8e3f1cc18460d4ea585b013134e305c56dab9987e503c7809d14b (height 1): %!w(MISSING) -> BLOCK_COINBASE_MISSING_HEIGHT (15): the coinbase signature script must start with the serialized block height
```

The `-> BLOCK_COINBASE_MISSING_HEIGHT ...` portion is the correct wrap chain (the errors package appends the wrapped error via ` -> ` in `Error()`). The `%!w(MISSING)` immediately before it is the bug.

### Root cause

`errors.New(code, message, params...)` (`errors/errors.go:354-372`):
1. If the last param is an `error`, it is extracted as the wrapped error and **removed from `params`**.
2. Remaining params are then applied with `fmt.Errorf(message, params...)`.

So a call like:
```go
errors.NewStorageError("failed to extract coinbase height for block %s (height %d): %w", hash, currentHeight, err)
```
becomes `fmt.Errorf("...%s...%d...%w", hash, currentHeight)` — the `%w` has no argument → `%!w(MISSING)`. When `%w` is the *only* verb, the format step is skipped entirely and the literal `%w` survives in the message.

### Fix

Drop the literal `: %w` from `StoreBlock.go:922`. The wrapped error is already chained and printed via ` -> `.

### Broader pattern (separate cleanup)

`: %w` in a `New*Error(...)` message is a codebase-wide footgun with this errors package. Other call sites share it, e.g.:
- `stores/blockchain/sql/StoreBlock.go:840,845`
- `stores/utxo/aerospike/spend.go:828`
- `stores/utxo/aerospike/spend_expressions.go:366,446`

Worth a lint rule (forbid `%w` inside `errors.New*` format strings) or a sweep, tracked separately from the immediate teratestnet log-noise fix above.

oskarszoon:
## What

Fixes #1332.

`errors.New` (`errors/errors.go`) extracts a trailing `error`/`*Error` argument as the wrapped error **before** calling `fmt.Errorf(message, params...)`. Any `%w` verb left in the format string is therefore orphaned:

- other verbs present → `fmt.Errorf` runs one arg short → renders `%!w(MISSING)`
- `%w` is the only verb → `params` is empty, formatting is skipped → the **literal** `%w` survives

The wrapped error is already rendered by `Error()` via ` -> ` chaining, so `%w` is redundant and always wrong in this API. Observed on teratestnet:

```
STORAGE_ERROR (69): failed to extract coinbase height for block ...: %!w(MISSING) -> BLOCK_COINBASE_MISSING_HEIGHT (15): ...
```

## Changes

- **Defensive fix (root cause):** `errors.New` now strips an orphaned `%w` once the trailing error has been extracted. The scan is escape-aware, so an intentional `%%w` (literal `%w`) is preserved; a dangling separator (`: `, ` - `, `, `) left before the verb is tidied. This fixes every current and future render at the source.
- **Call-site sweep:** dropped the redundant `: %w` from the `errors.New*` call sites across the tree (mechanical; the trailing error arg stays and is still chained).
- **CI guard:** `scripts/check_error_wrap_verbs.sh` fails the build if `%w` reappears inside an `errors.New*` format string (the `errors` package is exempt — its tests build such strings deliberately). Wired into the `golangci-lint` job (PR) and the Sonarqube job (main). `forbidigo`/`govet` can't express this rule; see the script header.
- **Tests:** `errors_test.go` covers the StoreBlock shape, `%w`-only, escaped `%%w`, mid-string, separator variants, multiple `%w`, and the no-`%w` hot-path (byte-identical output).

## Verification

- `go build ./...`, `go vet` on changed packages — clean
- `go test ./errors/... ./util/ ./util/usql/... ./services/blockvalidation/... ./services/p2p/...` — green
- `./scripts/check_error_wrap_verbs.sh` — passes (0 remaining)

### DIFF
--- a/.github/workflows/teranode_main_tests.yaml
+++ b/.github/workflows/teranode_main_tests.yaml
@@ -161,6 +161,9 @@ jobs:
         run: |
           ./scripts/check_filenames.sh
 
+      - name: Forbid %w in errors.New* format strings
+        run: ./scripts/check_error_wrap_verbs.sh
+
       - name: Set up Go
         uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5
         with:
--- a/.github/workflows/teranode_pr_tests.yaml
+++ b/.github/workflows/teranode_pr_tests.yaml
@@ -35,6 +35,9 @@ jobs:
     steps:
       - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
 
+      - name: Forbid %w in errors.New* format strings
+        run: ./scripts/check_error_wrap_verbs.sh
+
       - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5
         with:
           go-version: ${{ env.GO_VERSION }}
--- a/cmd/resetblockassembly/resetblockassembly.go
+++ b/cmd/resetblockassembly/resetblockassembly.go
@@ -27,21 +27,21 @@ func ResetBlockAssembly(logger ulogger.Logger, settings *settings.Settings, full
 	// Initialize the block assembly service
 	ba, err := blockassembly.NewClient(ctx, logger, settings)
 	if err != nil {
-		return errors.NewConfigurationError("failed to create block assembly client: %w", err)
+		return errors.NewConfigurationError("failed to create block assembly client", err)
 	}
 	defer func() { _ = ba.Close() }()
 
 	if validateInputs {
 		if err = ba.ResetBlockAssemblyValidateInputs(ctx); err != nil {
-			return errors.NewProcessingError("failed to reset block assembly with input validation: %w", err)
+			return errors.NewProcessingError("failed to reset block assembly with input validation", err)
 		}
 	} else if fullReset {
 		if err = ba.ResetBlockAssemblyFully(ctx); err != nil {
-			return errors.NewProcessingError("failed to reset block assembly: %w", err)
+			return errors.NewProcessingError("failed to reset block assembly", err)
 		}
 	} else {
 		if err = ba.ResetBlockAssembly(ctx); err != nil {
-			return errors.NewProcessingError("failed to reset block assembly: %w", err)
+			return errors.NewProcessingError("failed to reset block assembly", err)
 		}
 	}
 
--- a/cmd/rewindblockchain/rewindblockchain/phase1_unmined.go
+++ b/cmd/rewindblockchain/rewindblockchain/phase1_unmined.go
@@ -45,7 +45,7 @@ func (e *env) runCleanupIterator(ctx context.Context, pf *preflightResult, mode 
 		return errors.NewProcessingError("unknown iterator mode %d", mode)
 	}
 	if err != nil {
-		return errors.NewStorageError("failed to open iterator: %w", err)
+		return errors.NewStorageError("failed to open iterator", err)
 	}
 	if it == nil {
 		return nil
@@ -60,7 +60,7 @@ func (e *env) runCleanupIterator(ctx context.Context, pf *preflightResult, mode 
 	for {
 		batch, nextErr := it.Next(ctx)
 		if nextErr != nil {
-			return errors.NewStorageError("iterator Next: %w", nextErr)
+			return errors.NewStorageError("iterator Next", nextErr)
 		}
 		if len(batch) == 0 {
 			return nil
--- a/cmd/rewindblockchain/rewindblockchain/phase2_blocks.go
+++ b/cmd/rewindblockchain/rewindblockchain/phase2_blocks.go
@@ -16,7 +16,7 @@ import (
 func (e *env) phase2Blocks(ctx context.Context, pf *preflightResult) error {
 	for _, b := range pf.deleteList {
 		if err := e.rewindOneBlock(ctx, pf, b); err != nil {
-			return errors.NewProcessingError("rewinding block %s (height %d): %w", b.hash.String(), b.height, err)
+			return errors.NewProcessingError("rewinding block %s (height %d)", b.hash.String(), b.height, err)
 		}
 	}
 	return nil
@@ -25,12 +25,12 @@ func (e *env) phase2Blocks(ctx context.Context, pf *preflightResult) error {
 func (e *env) rewindOneBlock(ctx context.Context, pf *preflightResult, bl blockToDelete) error {
 	block, _, err := e.blockchainStore.GetBlock(ctx, bl.hash)
 	if err != nil {
-		return errors.NewStorageError("GetBlock: %w", err)
+		return errors.NewStorageError("GetBlock", err)
 	}
 
 	subtrees, err := block.GetSubtrees(ctx, e.logger, e.subtreeStore, e.concurrency)
 	if err != nil {
-		return errors.NewStorageError("GetSubtrees: %w", err)
+		return errors.NewStorageError("GetSubtrees", err)
 	}
 
 	collector := newRemovalCollector()
@@ -75,7 +75,7 @@ func (e *env) rewindOneBlock(ctx context.Context, pf *preflightResult, bl blockT
 	}
 
 	if err = e.blockchainStore.DeleteBlock(ctx, bl.hash); err != nil {
-		return errors.NewStorageError("DeleteBlock: %w", err)
+		return errors.NewStorageError("DeleteBlock", err)
 	}
 	e.stats.BlocksDeleted++
 
@@ -91,7 +91,7 @@ func (e *env) deleteCoinbase(ctx context.Context, pf *preflightResult, coinbaseH
 		if isNotFound(err) {
 			return nil
 		}
-		return errors.NewStorageError("Get coinbase %s: %w", coinbaseHash.String(), err)
+		return errors.NewStorageError("Get coinbase %s", coinbaseHash.String(), err)
 	}
 
 	var deletedHits, survivors []uint32
@@ -106,7 +106,7 @@ func (e *env) deleteCoinbase(ctx context.Context, pf *preflightResult, coinbaseH
 		if len(deletedHits) > 0 {
 			removals := []utxo.BlockIDsRemoval{{TxHash: coinbaseHash, BlockIDs: deletedHits}}
 			if err = e.utxoStore.RemoveBlockIDs(ctx, removals); err != nil {
-				return errors.NewStorageError("RemoveBlockIDs coinbase: %w", err)
+				return errors.NewStorageError("RemoveBlockIDs coinbase", err)
 			}
 			e.stats.TxsBlockIDsTrimmed++
 		}
@@ -115,7 +115,7 @@ func (e *env) deleteCoinbase(ctx context.Context, pf *preflightResult, coinbaseH
 
 	if err = e.utxoStore.Delete(ctx, coinbaseHash); err != nil {
 		if !isNotFound(err) {
-			return errors.NewStorageError("Delete coinbase: %w", err)
+			return errors.NewStorageError("Delete coinbase", err)
 		}
 	}
 	e.stats.TxsDeleted++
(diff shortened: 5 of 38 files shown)


ordishs (approved):
Approve — solid fix with a good belt-and-braces design (runtime strip + call-site sweep + CI guard + tests). The defensive `stripOrphanedWrapVerbs` is correct for this package (`%w` is never meaningful here), escape-aware for `%%w`, and gated behind a `strings.Contains` fast-path so the hot path stays cheap. Traced all the stripping branches and separator edge cases against the tests — they hold.

One requested follow-up before/after merge:

### Sweep is incomplete, and the CI guard can't see the misses

Two `errors.New*` call sites in a file the sweep *did* touch still contain `: %w`, and the new guard reports clean:

```
cmd/rewindblockchain/rewindblockchain/preflight.go:120
  errors.NewStorageError(`failed to read state["BlockAssembler"] (pass --target-height to override): %w`, err)
cmd/rewindblockchain/rewindblockchain/preflight.go:125
  errors.NewProcessingError(`failed to decode state["BlockAssembler"] (pass --target-height to override): %w`, err)
```

(Line 113 in the same function was swept; 120/125 were not.)

**Root cause:** the guard regex `errors\.New[A-Za-z]*\([^)]*%w` uses `[^)]*`, which stops at the first `)`. The `)` in `(pass --target-height to override)` truncates the match before it reaches `%w`. The sweep was almost certainly driven by the same grep, which is why it missed exactly the lines the guard also can't see — so this is a real false-negative, not a theoretical one.

**Runtime impact: none** — the defensive strip handles these correctly at render time, so there's no user-facing `%!w`. This actually validates the defensive layer as the right call. But it does undercut two stated goals: the sweep isn't complete "across the tree," and the guard won't catch the pattern reappearing.

Suggested:
- Drop `: %w` from `preflight.go:120` and `:125` for consistency.
- Harden the guard against a `)` inside the format string (e.g. a second pass with `.*%w` instead of `[^)]*%w`, or find-`%w`-then-filter). At minimum, document this blind spot in the script header alongside the two already noted (continuation lines, pre-formatted `%%w`).
- Consider adding a test case with a `)` before the `%w` — the exact shape that slipped through here.

### Minor
- Guard `--exclude-dir=errors` excludes any dir named `errors` at any depth (fine today, mild footgun).
- Guard is a line-based text grep, so a comment containing `errors.NewX(...%w` would also fail the build.

Nice touches worth calling out: the `services/p2p/Server.go` change that drops the `fmt.Sprintf(..."%%w", ...)` workaround, and the aerospike `spend.go` relabel that also fixes a pre-existing unlabeled `%d`.

ctnguyen (changes requested):
Review at commit [3af8792](https://github.com/oskarszoon/teranode/commit/3af879227878ebeff7396f4755b98b7fcf1fd96d)

This is a well-constructed fix. The three-layer design — runtime strip, call-site sweep, CI guard — is the right shape for a footgun like this, and the defensive `stripOrphanedWrapVerbs` is the load-bearing piece done correctly: it's escape-aware so `%%w` survives, gated behind a `strings.Contains` fast-path so the no-`%w` path stays cheap, and it only runs once a trailing error has actually been extracted. Importantly, wrapping in this package never depended on `fmt`'s `%w` (the cause is stored in the `wrappedErr` field and rendered via ` -> ` in `Error()`), so stripping the verb changes rendering only and leaves `errors.Is`/`errors.As` untouched — I traced that through `New` and it holds. The doc comment on `New`, the `services/p2p/Server.go` cleanup that drops the old `fmt.Sprintf(..."%%w")` workaround, and the `util/grpc.go` `[%w]`→plain fix are all nice touches. The test matrix is genuine and the separator edge cases line up with the implementation. One real gap remains below.

## Blocking issues

### ChiR1 — Sweep misses two sites and the CI guard can't see them
*Same as [@ordishs's point "Sweep is incomplete, and the CI guard can't see the misses" (24-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1335#pullrequestreview-4773104822).*

**Problem:** Two `errors.New*` call sites still carry `: %w` at HEAD, in a file the sweep otherwise touched (line 113 was swept):

```
cmd/rewindblockchain/rewindblockchain/preflight.go:120
  errors.NewStorageError(`failed to read state["BlockAssembler"] (pass --target-height to override): %w`, err)
cmd/rewindblockchain/rewindblockchain/preflight.go:125
  errors.NewProcessingError(`failed to decode state["BlockAssembler"] (pass --target-height to override): %w`, err)
```

The new guard reports `OK` on these. Root cause is the regex `errors\.New[A-Za-z]*\([^)]*%w`: `[^)]*` stops at the first `)`, and the `)` in `(pass --target-height to override)` truncates the match before it reaches `%w`. The sweep was evidently driven by the same grep, so it missed exactly the lines the guard also can't see — a real false negative, not a theoretical one. I ran the guard's own regex across the tree at HEAD and it returns zero matches, while a `)`-tolerant scan finds precisely these two lines (and no others — the rest of the sweep is complete).

**Why it matters:** Runtime impact is nil — the defensive strip renders both cleanly (e.g. `...override): %w` → `...override)`), which actually validates the belt-and-braces design. But it undercuts two of the PR's stated deliverables: the sweep is not complete "across the tree," and the guard — the mechanism meant to stop the pattern reappearing — will silently pass any future `errors.New*` whose format string contains a `)` before the `%w`. A guard that green-lights the exact pattern it ships to forbid isn't yet doing its job.

**Fix:** Drop `: %w` from both lines, and harden the regex so a `)` inside the format string no longer truncates the match:

```bash
# was: 'errors\.New[A-Za-z]*\([^)]*%w'
matches=$(grep -rnE 'errors\.New[A-Za-z]*\(.*%w' \
  --include='*.go' --exclude-dir=vendor --exclude-dir=errors . || true)
```

Also add a regression test case with a `)` before the `%w` (the exact shape that slipped through) to `TestStripOrphanedWrapVerbs`, e.g. `"read state (pass flag to override): %w"` → `"read state (pass flag to override)"`. The remaining line-based blind spots (a `%w` on a continuation line, a pre-formatted `%%w`) are already covered at runtime and noted in the script header — worth leaving a one-line note there that the `)` case is now handled by `.*` rather than `[^)]*`.

## Non-blocking issues

### ChiR2 — `--exclude-dir=errors` is broader than intended
*Same as [@ordishs's Minor point 1 (24-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1335#pullrequestreview-4773104822).*

**Problem:** `--exclude-dir=errors` excludes *any* directory named `errors` at any depth, not just the top-level `errors/` package. It also silently exempts non-test `.go` files inside that package, so a stray `errors.New*("...%w", err)` added to the package's own production code would go unchecked, not just the deliberate `%w` strings in `errors_test.go`.

**Why it matters:** Harmless today, but it's a quiet way for the guard's coverage to erode as the tree grows. The exemption is only meant to cover the errors package's own tests.

**Fix:** Anchor the exclusion to the intended path rather than a bare name, e.g. `--exclude-dir` won't take a path, so filter explicitly — restrict the exemption to test files with a post-filter (`| grep -v '^errors/.*_test\.go:'`) or scope the exclusion to the repo-root `./errors` only. A short comment on why the exemption exists would also help.

### ChiR3 — Guard is a line grep, so comments can fail the build
*Same as [@ordishs's Minor point 2 (24-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1335#pullrequestreview-4773104822).*

**Problem:** The guard matches raw text per line, so a comment or doc string mentioning the forbidden shape — e.g. `// don't write errors.NewStorageError("...: %w", err)` — would fail the build even though it compiles to nothing. Hardening the regex to `.*%w` (ChiR1) widens this slightly, since more of the line is now in scope.

**Why it matters:** A false positive on a comment is annoying rather than dangerous, but it makes the guard feel unreliable and invites `# nolint`-style workarounds that weaken it.

**Fix:** Low priority given the payoff, but if it ever bites, strip trailing line comments before matching (e.g. a `sed 's://.*$::'` pre-pass) or accept the limitation and document it in the script header alongside the continuation-line and `%%w` notes already there.

## Recap

| ChiR1 | Incomplete sweep, guard blind spot | YES | 55% |
| ChiR2 | `--exclude-dir=errors` too broad | | 20% |
| ChiR3 | Comment false positive | | 15% |

oskarszoon:
### On the Sonar "coverage on new code" gate (57.9% < 80%)

This is a measurement artifact of the sweep, not a real coverage regression.

- The actual behavioral change is the defensive `%w` strip in `errors.New`, and that's fully covered by the new tests in `errors_test.go`.
- Everything else in the diff is mechanical: it only removes a redundant `: %w` from existing `errors.New*` format-string literals. No logic changed, no branches added.
- Sonar counts each of those edited lines as "new code to cover." Most are error-return branches in production files (`Server.go`, `StoreBlock.go`, `catchup*.go`, `spend*.go`, `util/grpc.go`, …) that were **already uncovered before this PR** — the string edit just re-marks them as "new," so they now count against new-code coverage.

The overall project coverage condition passed (`coverage: OK`, ≥80%) — nothing across the codebase actually lost coverage. Only the new-code slice dipped, and only because pre-existing uncovered error branches got re-touched.

Keeping the sweep + CI guard intentionally (belt-and-braces with the runtime strip). Requesting a gate override for this PR on that basis.

oskarszoon:
Addressed in 4a11bb3e6.

**ChiR1** — confirmed the false negative. `preflight.go:120,125` used backtick raw strings with `(pass --target-height to override)` before the `%w`; the `)` truncated both the sweep grep and the guard's `[^)]*`. A `.*%w` scan across the tree finds exactly those two and nothing else, so the rest of the sweep is complete. Dropped `: %w` from both, changed the guard regex to `.*%w`, and added a `)`-before-`%w` case to both `TestStripOrphanedWrapVerbs` and `TestNewStripsOrphanedWrapVerbs`.

**ChiR2** — replaced `--exclude-dir=errors` with a post-filter scoped to `errors/*_test.go`. The errors package's own callers use the unprefixed `New(...)` form so they never match the pattern anyway; the exemption now only covers the deliberate `%w` strings in the package's tests, not its production code, and is commented.

**ChiR3** — left as a line grep and documented the blind spot in the header alongside the continuation-line and `%%w` notes. A `//`-stripping pre-pass would introduce a worse false *negative* on error strings containing `//` (e.g. a URL), so it's a deliberate trade.

Guard now catches the `)` shape (verified) and is clean at HEAD.

freemans13 (approved):
Reviewed the full diff and ran the checks locally — approving.

**What I verified**
- `./scripts/check_error_wrap_verbs.sh` passes (0 remaining); a tree-wide sweep for leftover `%w` in non-test format strings also comes back empty, including continuation lines, so the documented multi-line blind spot is purely theoretical right now.
- `go test ./errors/...` — 448 pass; `go vet` on the errors package plus the swept packages is clean.
- Script has the executable bit (100755) and both CI steps are placed after `actions/checkout`, so they actually run.
- Spot-checked the call-site sweep (catchup, StoreBlock, advisory_lock): the trailing error argument is always kept, only the redundant `: %w` is dropped — no wrapped errors lost.

The fix is at the right layer: the runtime `stripOrphanedWrapVerbs` guard means nothing user-facing can ever render `%!w(MISSING)` or a literal `%w` again, the sweep cleans up today's call sites, and the CI script stops it regressing. Escape-aware handling of `%%w`, the fast-path early return, and thorough tests (StoreBlock shape, `%w`-only, escaped, mid-string, separator variants, `)` truncation) all look good.

One optional footnote, not blocking: both the strip and the CI regex key on the literal bytes `%w`, so a flagged verb like `%+w` or indexed `%[1]w` would slip past both. Obscure enough that nobody writes it in practice — mentioning only for completeness.

ctnguyen (approved):
Review at commit [4a11bb3](https://github.com/oskarszoon/teranode/commit/4a11bb3e6e57b4881f667110ac7d78b4e96fc4d8) all three points from the previous round are genuinely addressed

## Status of previous points

| ChiR1 | Incomplete sweep, guard blind spot | YES | FIXED |
| ChiR2 | `--exclude-dir=errors` too broad | | FIXED |
| ChiR3 | Comment false positive | | FIXED |

Issue #1187, opened 2026-06-30
legacy sync: unbounded UTXO batcher fan-out overloads single-node Aerospike → cluster-wide stall


oskarszoon:
## Problem

On a single-node mainnet operator stack (`docker.m`), legacy block sync below
checkpoint repeatedly wedges Aerospike when it hits a large block. Processing a
big block (e.g. mainnet height 874485, ~2.64M txs) drives the UTXO store batchers
at up to ~4096-way concurrency against the one Aerospike node. The node saturates,
batch ops time out, the block fails with `STORAGE_ERROR`, is re-queued, and the
full multi-million-record operation retries — re-loading Aerospike each time.
Sustained, Aerospike goes unresponsive to all clients including `asadm`, and only a
full restart recovers it (slow — rebuilds the in-RAM index from disk). Seen multiple
times during mainnet catchup.

Representative error (repeats on the same block):
```
PROCESSING (4): failed to batch-decorate previous outputs ->
STORAGE_ERROR (69): error in aerospike send outpoint batch records ->
UNKNOWN (0): ResultCode: MAX_RETRIES_EXCEEDED ... command execution timed out on client
```

## Evidence

- **Infinite re-try on one block:** legacy re-attempted height 874485 (2.64M txs)
  **145+ times in 18h, zero forward progress**; each attempt ~5 min in
  `extendTransactions … DONE … with error`.
- **Fan-out (`.docker.m`):** store errgroup = `legacy_storeBatcherSize 1024 ×
  legacy_storeBatcherConcurrency 4 = 4096` (`handle_block.go:948-974`, `Create`);
  spend errgroup `1024 × 4` (`handle_block.go:1324-1325`); decorate errgroup
  `utxostore_outpointBatcherSize = 4096` (`get.go:1264`). Each batcher then
  dispatches `utxostore_batcherMaxConcurrent.docker.m = 24` concurrent `BatchOperate`
  to the single node (`aerospike.go:387`), across 4 batcher types.
- **Not fd/memory exhaustion:** at load, Aerospike `client_connections` ~190-250 of
  `proto-fd-max=30000`, asd open fds ~1650 of `nofile=32768`, `system_free_mem_pct`
  77, `stop_writes=false`. It's **request-thread saturation** (`service-threads=240`,
  `batch-index-threads=48` on a 48-core host) under the looping batch load — `asadm`
  times out because the info path queues behind saturated batch threads.
[...]

PR #1336, opened 2026-07-24
fix(legacy): stop NOT_FOUND ERROR-spam cascade from a failed parent block (#1333)

### LINKED ISSUE #1333: legacy/netsync: failed parent triggers ERROR-spam cascade of "previous block NOT_FOUND" for all descendants
**Component:** `services/legacy/netsync` (`handle_block.go:73`, `HandleBlockDirect`)

### Summary

When a block fails to store/validate, legacy keeps processing its descendant blocks anyway. Each descendant then fails its parent lookup and logs an `ERROR`, so a single bad/unstored block produces an unbounded cascade of misleading `previous block ... NOT_FOUND` errors that bury the real root cause.

Root cause was block 1 failing to store (coinbase-height parser — see the BIP34 issue). Downstream, legacy logged a wall of:

```
ERROR | netsync/handle_block.go:73 | legacy | [HandleBlockDirect][0000000084e4742a...] failed to get block header for previous block 000000005f02edae...: BLOCK_NOT_FOUND (10): error in GetBlockHeader -> NOT_FOUND (3): not found
ERROR | netsync/handle_block.go:73 | legacy | [HandleBlockDirect][00000000206c5603...] failed to get block header for previous block 0000000084e4742a...: BLOCK_NOT_FOUND (10): ...
ERROR | netsync/handle_block.go:73 | legacy | [HandleBlockDirect][00000000b1fe69f8...] failed to get block header for previous block 00000000206c5603...: BLOCK_NOT_FOUND (10): ...
... (continuous)
```

These errors "don't make sense" on their own — the parent is missing only because an ancestor was rejected upstream. They add no signal and hide the one error that matters.

### Expected

Once a block is known-bad or its parent is known-missing/invalid, don't keep attempting its descendants as if they were independently processable. Options:
- short-circuit: skip processing blocks whose parent is a known-missing/rejected block, or
- downgrade the expected orphan/parent-missing cascade to `DEBUG`/`WARN` and log the root rejection once at `ERROR`.

### Impact

Log pollution + operator confusion during any from-genesis IBD that hits an early rejection. No functional data loss, but it makes triage materially harder.

oskarszoon:
## What

Fixes #1333.

When a block fails to store/validate, legacy `netsync` kept processing its already-queued descendants. Each descendant failed its parent lookup at `handle_block.go` and logged an `ERROR`, so a single bad/unstored block (root cause on teratestnet: block 1 failing the BIP34 coinbase-height parser, #1332) produced an unbounded cascade of misleading `previous block NOT_FOUND` errors that buried the one error that mattered.

## Changes

- **Short-circuit the cascade.** `handleBlockMsg` now tracks recently-failed block hashes in a TTL-bounded, size-capped `expiringmap` (10 min TTL, 1024 cap, deleted on a successful reprocess). A block whose parent is a known-failed hash is skipped **before the block-lookup RPCs** (no `GetBlockExists`/`GetBlockHeader`) and before the headers-first block-size sampling, marked failed itself so the whole descendant chain is suppressed transitively, and its delivering peer's stall timer is refreshed (the fault is a rejected ancestor, not the peer). The skip still fires the existing getblocks recovery (`GetBestBlockHeader` + `GetBlockLocator`), so sync resumes once the root block is resolved.
- **Log level.** `HandleBlockDirect` logs a missing parent (`ErrBlockNotFound`) at `DEBUG` — it's an expected orphan / out-of-order / rejected-ancestor condition the caller recovers from. Genuine lookup failures (e.g. storage errors) still log at `ERROR`. Net effect: the root rejection is logged once at `ERROR`; descendants produce no `ERROR` spam.
- **Refactor.** Extracted the getblocks recovery into `requestMissingBlocks`, shared by the new skip path and the existing orphan path.

The map is independent of the #1187 `blockFailureBackoff` knobs so cascade suppression works even when that backoff is disabled; both maps are nil-guarded (struct-literal test managers) and stopped in `Stop()`.

## Verification

- `go test ./services/legacy/netsync/` — green (incl. the #1187 backoff suite, unchanged)
- `go vet`, `gofmt`, `golangci-lint run ./services/legacy/netsync/` — clean
- New tests: cascade short-circuit (no RPC, no `ERROR`, transitive mark, getblocks fired), record-on-failure, clear-on-success, normal-orphan-not-marked.

### DIFF
--- a/services/legacy/netsync/handle_block.go
+++ b/services/legacy/netsync/handle_block.go
@@ -70,7 +70,18 @@ func (sm *SyncManager) HandleBlockDirect(ctx context.Context, peer *peer.Peer, b
 	// Lookup previous block height from blockchain
 	_, previousBlockHeaderMeta, err = sm.blockchainClient.GetBlockHeader(ctx, &block.MsgBlock().Header.PrevBlock)
 	if err != nil {
-		sm.logger.Errorf("[HandleBlockDirect][%s] failed to get block header for previous block %s: %s", blockHash.String(), block.MsgBlock().Header.PrevBlock, err)
+		// A missing parent is an expected, recoverable condition: a normal orphan /
+		// out-of-order tip announce, or a descendant of a block that was rejected
+		// upstream. The caller (handleBlockMsg) answers with a getblocks and, for a
+		// known-failed ancestor, short-circuits the descendant cascade (#1333) — so
+		// logging it at ERROR here only produces misleading "previous block
+		// NOT_FOUND" spam. Genuine lookup failures (e.g. storage errors) still ERROR.
+		if errors.Is(err, errors.ErrBlockNotFound) {
+			sm.logger.Debugf("[HandleBlockDirect][%s] previous block %s not found (orphan/out-of-order; caller will request missing blocks): %v", blockHash.String(), block.MsgBlock().Header.PrevBlock, err)
+		} else {
+			sm.logger.Errorf("[HandleBlockDirect][%s] failed to get block header for previous block %s: %s", blockHash.String(), block.MsgBlock().Header.PrevBlock, err)
+		}
+
 		return errors.NewProcessingError("failed to get block header for previous block %s", block.MsgBlock().Header.PrevBlock, err)
 	}
 
--- a/services/legacy/netsync/manager.go
+++ b/services/legacy/netsync/manager.go
@@ -1662,21 +1662,6 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockQueueMsg) error {
 		}
 	}
 
-	// Track block size for dynamic in-flight adjustment during headers-first mode.
-	// This allows us to start aggressive (20 blocks) and automatically reduce
-	// to 1 block when encountering large (>2GB) blocks on mainnet.
-	if sm.headersFirstMode.Load() && bmsg.block != nil {
-		blockSize := int64(bmsg.block.SerializeSize())
-		sm.blockSizeTracker.addBlockSize(blockSize)
-
-		dynamicMax := sm.blockSizeTracker.calculateMaxInFlightBlocks()
-		avgSize := sm.blockSizeTracker.getAverageSize()
-		sm.logger.Debugf("[handleBlockMsg][%s] Block size: %d bytes, avg: %d bytes, dynamic max in-flight: %d",
-			bmsg.blockHash, blockSize, avgSize, dynamicMax)
-	}
-
-	sm.logger.Debugf("[handleBlockMsg][%s] calling HandleBlockDirect", bmsg.blockHash)
-
 	// Hand sole ownership of the decoded block to HandleBlockDirect. The
 	// blockHandler goroutine keeps *bmsg alive until the reply is sent, so
 	// leaving the field set would pin the multi-GB wire block (and its decode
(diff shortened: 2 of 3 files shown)


ordishs (approved):
Approve. Solid, well-scoped fix — the cascade suppression is correctly bounded (10-min TTL + 1024 cap), self-healing (delete-on-success + transitive marking + TTL), nil-guarded, and stopped in `Stop()`. No backlog leak: `blockBacklog` is decremented unconditionally by the `blockHandler` caller, and the skip's `return nil` matches the orphan path's success signal. Net RPC reduction vs. the old cascade (drops both lookup RPCs and the ERROR per descendant). Verified the four new tests pass locally (`go test -tags testtxmetacache ... ./services/legacy/netsync/` → ok).

A few non-blocking comments:

- **Block-size sampling placement (low).** The short-circuit sits *after* the headers-first block-size sampling, whereas the #1187 backoff skip was deliberately placed *before* it (comment at the sampling block) so skipped blocks don't bias `calculateMaxInFlightBlocks()`. A short-circuited descendant still feeds its size into `blockSizeTracker`. Small impact (real sizes, not garbage), but inconsistent with the stated #1187 rationale — consider moving the skip above the sampling block (read `prevBlockHash` before `bmsg.block = nil`).

- **PR-body wording.** "skipped before any RPC" is slightly overstated: the skip still issues `GetBestBlockHeader` + `GetBlockLocator` per descendant via `requestMissingBlocks`. The test is honest (only asserts the two block-lookup RPCs aren't called). Suggest "before the block-lookup RPCs."

- **Convention (informational).** Tests use `blockchain2.Mock` while AGENTS.md prefers the `sqlitememory` store — but this follows the established pattern in this file (`newBackoffTestManager`, #1187 suite), so consistent with surrounding code.

oskarszoon:
Addressed in 0742cb0fa.

**Block-size sampling placement** — good catch, moved the #1333 short-circuit above the headers-first sampling block (hoisting the `msgBlock`/`prevBlockHash` extraction with it and sampling from `msgBlock`), matching the #1187 backoff-skip placement. A short-circuited, re-delivered descendant no longer feeds `blockSizeTracker`.

**PR-body wording** — fixed: "before the block-lookup RPCs"; the skip still issues `GetBestBlockHeader` + `GetBlockLocator` via `requestMissingBlocks`.

**Test convention** — left as `blockchain2.Mock` to match the existing `newBackoffTestManager` / #1187 suite in this file.

freemans13 (approved):
Reviewed the full modified files, traced the logic through `handleBlockMsg`, checked the blast radius of the log-level change, and ran the tests (`go vet` clean; full `services/legacy/netsync` package green; the 12 `TestHandleBlockMsg_*` tests pass under `-race`).

The key correctness property holds: normal orphans (parent still in flight) are never marked failed, so legitimate blocks aren't suppressed — and `TestHandleBlockMsg_OrphanNotMarkedFailed` proves it. The log-level downgrade is well-contained (`HandleBlockDirect` is only called from `handleBlockMsg` in production). The new map is thread-safe (RWMutex-guarded `expiringmap`, nil-guarded, stopped in `Stop()`, built after the last fallible step in `New()`). FIFO-by-insertion eviction means the immediate parent is always among the newest entries and never evicted before its child checks it, so no suppression leak even on deep cascades.

Two cosmetic notes, non-blocking:
- "The root rejection is logged once at ERROR" is slightly optimistic for a permanently-bad root: the getblocks recovery keeps re-fetching it, so it re-fails and re-logs its genuine ERROR each cycle. The descendant NOT_FOUND spam (the actual target) is gone, which is what matters.
- The skip still calls `requestMissingBlocks` (GetBestBlockHeader + GetBlockLocator) per descendant, so it's cheaper than before, not RPC-free. Matches the pre-existing orphan path, so no regression.

LGTM.

PR #1337, opened 2026-07-24
fix(asset): back off centrifuge p2p dial, stop ERROR-spam when no peer present (#1334)

### LINKED ISSUE #1334: asset/centrifuge: ERROR-spam dialing unresolvable "peer" p2p host when no p2p server present
**Component:** `services/asset/centrifuge_impl/centrifuge.go:265-305` (`startP2PListener`)

### Summary

On a stack with no reachable p2p server (e.g. the `teranode-quickstart` / legacy-only teratestnet setup, context `docker.m`), asset's centrifuge p2p listener retries a dial roughly once per second and logs each failure at `ERROR`, spamming the logs indefinitely.

```
ERROR | centrifuge.go:305 | asset | [Centrifuge] error dialing p2p server: dial tcp: lookup peer on 127.0.0.11:53: server misbehaving
ERROR | centrifuge.go:305 | asset | [Centrifuge] error dialing p2p server: dial tcp: lookup peer on 127.0.0.11:53: server misbehaving
... (~1/sec, continuous)
```

The dial target is built from `c.settings.P2P.HTTPAddress` (`centrifuge.go:266`) as `ws://<host>/p2p-ws`. The host resolves to the literal name `peer`, which doesn't exist in this stack — there is no p2p/peer container; sync is via legacy. So `P2P.HTTPAddress` is a placeholder/default that was never meant to be dialed here.

### Expected

- If `P2P.HTTPAddress` is unset/unresolvable, don't spin a retry loop that logs at `ERROR` every second — either skip the listener (as `startP2PListener` already does for an empty address, `centrifuge.go:267`) or back off and log at `WARN`/`DEBUG`.
- Fix the quickstart/`docker.m` default so `P2P.HTTPAddress` isn't set to an unreachable `peer` host when no p2p service is present.

### Impact

Benign for sync — cosmetic log noise only. Low priority, but constant and it clutters the asset service logs.

oskarszoon:
## What

Fixes #1334.

On a stack with no reachable p2p server (e.g. the legacy-only teratestnet / quickstart compose stacks, context `docker.m`, where the `peer` service is commented out), `centrifuge.connect()` retried the dial ~once per second and logged every failure at `ERROR`, spamming the asset logs indefinitely:

```
ERROR | centrifuge.go:305 | asset | [Centrifuge] error dialing p2p server: dial tcp: lookup peer on 127.0.0.11:53: server misbehaving
... (~1/sec, continuous)
```

## Changes

- **Backoff + log-downgrade in `connect()`.** Failed dials now back off exponentially (1s → 60s cap, via the existing `util/retry.CappedExponentialBackoff`). The first failure of a streak logs once at `WARN` — with a config hint when the host can't be resolved (`*net.DNSError`) — and subsequent failures log at `DEBUG`. Both the failure count and backoff reset on a successful connect, so a peer that appears late still reconnects within ≤60s. DNS failures do **not** stop the loop (they're routinely transient in compose/k8s). The inter-attempt sleep is context-aware, so shutdown stays prompt during a long backoff.
- **Graceful skip on empty address.** `startP2PListener` now logs a `WARN` and returns `nil` when `p2p_httpAddress` is empty, instead of failing the whole asset service (the global default is `localhost:9906`, so empty is explicit operator intent).

No compose/config changes: with the backoff in place, an occasional `WARN`/`DEBUG` on a peerless stack is fine, so there's no need to flip `asset_centrifuge_disable` in the shipped compose files.

Dial and sleep are injected via unexported struct seams (defaulted in `New`) so the backoff timing and dial outcomes are testable without real DNS or sleeps.

## Verification

- `go test ./services/asset/centrifuge_impl/` — green
- `go vet`, `gofmt`, `golangci-lint run` — clean
- New tests: backoff growth/cap + log levels, DNS-hint WARN, reset-after-reconnect, prompt shutdown during backoff, empty-address skip.

### DIFF
--- a/services/asset/centrifuge_impl/centrifuge.go
+++ b/services/asset/centrifuge_impl/centrifuge.go
@@ -4,6 +4,7 @@ package centrifuge_impl
 import (
 	"context"
 	"encoding/json"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -17,6 +18,7 @@ import (
 	"github.com/bsv-blockchain/teranode/services/blockchain"
 	"github.com/bsv-blockchain/teranode/settings"
 	"github.com/bsv-blockchain/teranode/ulogger"
+	"github.com/bsv-blockchain/teranode/util/retry"
 	"github.com/centrifugal/centrifuge"
 	"github.com/google/uuid"
 	"github.com/gorilla/websocket"
@@ -27,6 +29,28 @@ const (
 	AccessControlAllowHeaders     = "Access-Control-Allow-Headers"
 	AccessControlAllowCredentials = "Access-Control-Allow-Credentials"
 )
+
+const (
+	// p2p dial retry backoff (#1334): on a stack with no reachable p2p server the
+	// dial fails repeatedly; back off exponentially and log at DEBUG after the
+	// first failure instead of an ERROR every second.
+	p2pDialInitialBackoff = 1 * time.Second
+	p2pDialBackoffFactor  = 2.0
+	p2pDialMaxBackoff     = 60 * time.Second
+)
+
+// sleepWithContext sleeps for d, returning early if ctx is cancelled. It is the
+// default for Centrifuge.sleepWithContext and keeps shutdown prompt during a long
+// dial backoff.
+func sleepWithContext(ctx context.Context, d time.Duration) {
+	timer := time.NewTimer(d)
+	defer timer.Stop()
+
+	select {
+	case <-ctx.Done():
+	case <-timer.C:
+	}
+}
 
 // Centrifuge manages real-time blockchain data broadcasting to WebSocket clients.
 // It provides pub/sub for blockchain events (new blocks, node status updates) using
@@ -45,6 +69,11 @@ type Centrifuge struct {
 	cachedCurrentNodeStatus *notificationMsg // Cached current node status for new clients
 	currentNodePeerID       string           // Track the current node's peer ID
 	statusMutex             sync.RWMutex     // Protects cachedCurrentNodeStatus and currentNodePeerID
+
+	// Test seams for the p2p dial loop (#1334), defaulted in New. Overridden in
+	// tests to drive dial outcomes and backoff timing without real DNS or sleeps.
+	dialWebsocket    func(urlStr string) (*websocket.Conn, *http.Response, error)
+	sleepWithContext func(ctx context.Context, d time.Duration)
 }
 
 // notificationMsg represents a notification message relayed to WebSocket clients.
@@ -112,6 +141,10 @@ func New(logger ulogger.Logger, tSettings *settings.Settings, repo *repository.R
 		repository: repo,
 		baseURL:    assetHTTPAddress,
 		httpServer: httpServer,
+		dialWebsocket: func(urlStr string) (*websocket.Conn, *http.Response, error) {
+			return websocket.DefaultDialer.Dial(urlStr, nil)
+		},
+		sleepWithContext: sleepWithContext,
 	}
 
 	// Only set blockchainClient if repo is not nil
@@ -265,7 +298,12 @@ func (c *Centrifuge) Start(ctx context.Context, addr string) error {
 func (c *Centrifuge) startP2PListener(ctx context.Context) error {
 	p2pServerAddress := c.settings.P2P.HTTPAddress
 	if p2pServerAddress == "" {
-		return errors.NewConfigurationError("p2p_httpAddress not found in config")
+		// No p2p server configured — skip the listener rather than failing the
+		// whole asset service. The global default is localhost:9906, so an empty
+		// value is explicit operator intent (e.g. a legacy-only stack). Centrifuge
+		// only sources messages from the p2p websocket, so it is simply inert here.
+		c.logger.Warnf("[Centrifuge] p2p_httpAddress not configured - skipping p2p listener; live block/node status updates will be unavailable")
+		return nil
 	}
 
 	u := url.URL{Scheme: "ws", Host: p2pServerAddress, Path: "/p2p-ws"}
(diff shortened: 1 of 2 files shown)


services/asset/centrifuge_impl/centrifuge_test.go
```diff
@@ -1163,6 +1166,132 @@ func TestCentrifuge_connect(t *testing.T) {
...
+			return nil, nil, &net.OpError{Op: "dial", Net: "tcp", Err: &net.DNSError{Err: "server misbehaving", Name: "peer"}}
+		}
+
+		ctx, cancel := context.WithCancel(context.Background())
+		c.sleepWithContext = func(_ context.Context, _ time.Duration) { cancel() }
+
+		c.connect(ctx, dialURL, &atomic.Pointer[websocket.Conn]{}, &atomic.Bool{})
+
+		ml.AssertNumberOfCalls(t, "Errorf", 0)
+		ml.AssertNumberOfCalls(t, "Warnf", 1)
```
freemans13: This test is named for the DNS-hint branch, but it only asserts `Warnf` was called once — it never checks *which* WARN ran. `MockLogger` records call counts only (it discards the format string and args), so the generic-WARN branch and the DNS-hint branch are indistinguishable here. If `errors.As(err, &dnsErr)` ever stopped matching — a future `errors.As` refactor, or production errors wrapped differently than this hand-built `net.OpError` — the test would still pass while the config hint silently disappeared.

Suggestion: capture the formatted message (extend `MockLogger` to store messages, or use a logger that does) and assert the WARN contains the hint substring (e.g. `asset_centrifuge_disable`). The same applies to the "one WARN then DEBUG" case — it counts levels but never confirms the DEBUG lines carry the attempt/backoff detail they're meant to.

services/asset/centrifuge_impl/centrifuge.go
```diff
@@ -291,28 +329,77 @@ func (c *Centrifuge) startP2PListener(ctx context.Context) error {
 //   - client: Atomic pointer to the WebSocket connection
 //   - clientConnected: Atomic boolean indicating connection status
 func (c *Centrifuge) connect(ctx context.Context, u url.URL, client *atomic.Pointer[websocket.Conn], clientConnected *atomic.Bool) {
+	dial := c.dialWebsocket
```
freemans13, line 334: Minor: these nil-guards for `dialWebsocket`/`sleepWithContext` can't fire in practice — `New` always sets both, the fields are unexported, and `New` is the only constructor (the tests go through it too), so a nil seam is unreachable. Harmless, but it's defensive code that never executes. Either drop the guards and rely on `New`, or keep them with a one-line comment noting they guard zero-value construction. Not worth blocking on.

services/asset/centrifuge_impl/centrifuge.go
```diff
@@ -291,28 +329,77 @@ func (c *Centrifuge) startP2PListener(ctx context.Context) error {
...
+					switch {
+					case failures > 1:
+						// Steady-state noise stays at DEBUG.
+						c.logger.Debugf("[Centrifuge] error dialing p2p server (attempt %d, next retry in %s): %v", failures, backoff, err)
+					default:
+						// First failure of a streak: one WARN, with a config hint when
+						// the host cannot be resolved (the common "no p2p service here" case).
+						var dnsErr *net.DNSError
+						if errors.As(err, &dnsErr) {
+							c.logger.Warnf("[Centrifuge] cannot resolve p2p server host %q: %v - retrying with backoff (further attempts logged at DEBUG); if this deployment has no p2p service, set asset_centrifuge_disable=true or leave p2p_httpAddress empty", u.Host, err)
```
freemans13, line 383: Optional / subjective: by design, after the single WARN per streak everything drops to DEBUG. At the default INFO level a permanently peerless node then goes completely silent about the still-failing dial after one line — an operator scanning logs later sees no evidence it's still retrying. That's a reasonable trade-off (it's the whole point of the fix), but a very-low-frequency reminder — e.g. a re-WARN every N minutes, or once when the backoff first hits the 60s cap — would preserve visibility without reintroducing the spam. Fine to leave as-is.

freemans13 (commented):
Nice, well-scoped fix — the exponential backoff, the reset-on-reconnect, and treating an empty `p2p_httpAddress` as deliberate operator intent all read correctly, and I couldn't find a correctness bug in the production code. Three advisory notes below (one real test-coverage gap plus two minor/optional points); none are blockers.

oskarszoon:
Addressed in 6e43bb227 — all three were fair.

**Test-coverage gap (the real one)** — agreed, the DNS-hint test only counted WARNs and couldn't tell the hint branch from the generic one. Added a small `capturingLogger` (embeds `MockLogger`) that records formatted messages, and the test now asserts the WARN names the unresolvable host and contains `asset_centrifuge_disable` — so a future `errors.As` regression that drops the hint fails the test. The backoff test likewise now asserts the DEBUG lines carry the attempt/next-retry detail.

**Silent-after-WARN** — took the suggestion in its minimal form: one WARN when the backoff first saturates at the 60s cap ("still cannot reach ... retrying every 1m0s"), then DEBUG. A permanently peerless node now leaves two breadcrumbs per streak, no periodic spam. Resets on reconnect.

**Dead nil-guards** — kept them but added a comment noting they only cover zero-value struct-literal construction (New is the sole constructor and always sets both).

freemans13 (approved):
All three review points addressed cleanly — thanks.

1. **Test gap** — the new `capturingLogger` records the formatted message per level, and the DNS test now asserts the WARN actually contains `cannot resolve p2p server host` and `asset_centrifuge_disable`, so it pins the DNS-hint branch rather than just counting that some WARN fired. The backoff test also confirms the DEBUG lines carry the attempt/backoff detail. Wired correctly: `New` stores the logger verbatim and `connect` logs through `c.logger`, so nothing bypasses the capture.
2. **Nil-guards** — kept with a comment noting they only cover zero-value struct-literal construction, not a live path. Good.
3. **Observability** — the `cappedWarned` breadcrumb emits one extra WARN when the backoff first saturates at 60s and resets on a successful connect, so a permanently-peerless node leaves exactly two WARNs total (first failure + cap reached) and then stays quiet. Traced it against the updated test: wait sequence `1,2,4,8,16,32,60,60`, cap WARN on the 6th failure — matches the two-WARN assertion.

Approving. Note I verified by reading and tracing rather than a local build, so I'm relying on CI for the actual test run.

ordishs (approved):
Solid, well-scoped fix. Verified locally at head `ffdfacd7d`: `go test ./services/asset/centrifuge_impl/` → `ok` (3.7s), `go vet` → clean.

### What it does
- **`connect()`**: failed p2p dials back off exponentially (1s→60s cap via `retry.CappedExponentialBackoff`), log one `WARN` per failure streak (with a DNS config-hint) then `DEBUG`, instead of an `ERROR` every second. Count + backoff reset on success; inter-attempt sleep is context-aware.
- **`startP2PListener()`**: empty `p2p_httpAddress` logs a `WARN` and returns `nil` (service keeps running) instead of failing the asset service.
- Dial + sleep injected via unexported seams, defaulted in `New`.

### Correctness — sound
- Backoff progression `[1,2,4,8,16,32,60,60]` matches `CappedExponentialBackoff`.
- `failures`/`backoff` are goroutine-local; `client`/`clientConnected` stay atomic — no new races.
- Real dial error (`*net.OpError` wrapping `*net.DNSError` from Docker's embedded DNS) is matched by teranode's `errors.As` — traced through `isGRPCWrappedError`/`*Error` down to stdlib `errors.As`, and `net.OpError.Unwrap()` exposes the `*net.DNSError`.
- Context-aware sleep interrupts a long backoff on shutdown (test asserts <500ms).

### Comments (non-blocking)
1. **(test coverage) The DNS-hint test doesn't actually verify the hint branch.** `TestCentrifuge_connectBackoff/"unresolvable host logs a WARN with a config hint"` asserts only `Warnf` count == 1 / `Errorf` == 0. Both branches (DNS-hint vs generic) emit exactly one `Warnf`, and `mocklogger` records only the method name — so if `errors.As` silently failed to match `*net.DNSError`, the test would still pass green. Since the `errors.As`-through-teranode-errors path is the fragile part, consider asserting on message content (e.g. that it contains `asset_centrifuge_disable`).
2. **(dead code) Redundant nil-guards in `connect()`.** Lines 332–342 re-default `dial`/`sleep` when nil, but `New` always sets both and every test builds via `New`; the fallback dial closure also duplicates the one in `New`. Harmless, but unreachable — drop it or note why it's kept.
3. **(behavior change, flagging) Empty `p2p_httpAddress` is now non-fatal.** Deliberate and well-justified (default is `localhost:9906`, so empty = explicit intent). Residual risk: an operator who *accidentally* blanks the address now gets only a `WARN` + silently missing live updates rather than a hard failure. Message is clear, so acceptable.
4. **(nit)** DNS config-hint is only evaluated on the first failure of a streak; if attempt 1 is a transient non-DNS error and later ones are DNS, the hint never surfaces. Low impact on the target scenario.

Approving — the only substantive item is #1 (the DNS-hint test names a branch it doesn't verify). #2–#4 are optional polish.

icellan (approved):
Reviewed the diff. No P0/P1 issues. Backoff/reset semantics are consistent with the added tests, and the `asset_centrifuge_disable=true` remedy in the DNS-failure WARN (centrifuge.go:377) is a real setting. LGTM.

PR #1338, opened 2026-07-24
fix(model): decode BIP34 coinbase height with SV Node CScriptNum parity (#1142)

### LINKED ISSUE #1142: [TN-003][low] util/coinbase: BIP34 height parse ignores push-opcode/minimal encoding (SV Node divergence)
**Component:** `util/coinbase.go` (`extractCoinbaseHeightAndText`), used by `model/Block.ExtractCoinbaseHeight`

### Summary
`extractCoinbaseHeightAndText` reads `serializedLen := int(sigScript[0])` and treats the next bytes as a little-endian height, rather than decoding it as a script number (CScriptNum) like SV Node. Standard minimal push (`0x03`+3 bytes) is correct; `OP_PUSHDATA1`-prefixed or non-minimal encodings are mis-parsed.

### Evidence
- `util/coinbase.go:57-85`. Empirically: `03 bf ea 07`→518847 ✓; `04 bf ea 07 00` (non-minimal)→518847 (SV Node CScriptNum rejects non-minimal); `4c 03 …` (OP_PUSHDATA1)→mis-parsed (treats 0x4c=76 as length).
- `model/Block.go:494-503` uses `ExtractCoinbaseHeight()` as the authoritative BIP34 check.

### Why it matters
A legal `OP_PUSHDATA1` / non-minimal coinbase height would parse differently in Teranode vs SV Node — a potential BIP34 consensus divergence. Latent (all real miners use minimal pushes).

### Suggested fix
Decode the height with the script-number/opcode reader to match SV Node CScriptNum semantics.

### Suggested tests
Coinbase with OP_PUSHDATA1-encoded and non-minimal height → assert accept/reject parity with SV Node.

---

oskarszoon:
Closes #1142.

## Problem

`util/coinbase.go` `extractCoinbaseHeightAndText` read the first coinbase scriptSig byte as a raw length prefix (`serializedLen := int(sigScript[0])`), not as a Bitcoin script push opcode. Consequences:

- **`OP_1..OP_16` (heights 1–16) errored** — `0x51` was read as "length 81", overran the script, and returned `BLOCK_COINBASE_MISSING_HEIGHT`. This hard-blocks teratestnet from-genesis IBD: block 1's coinbase starts with `OP_1`, so the node stalls at stored height 0 and every later block fails with `previous block NOT_FOUND`. Surfaced by #1301, which correctly switched BIP34 enforcement to per-network `BIP0034Height` (0 on teratestnet/tstn), so extraction now runs from block 1.
- **Non-minimal and `OP_PUSHDATA`-prefixed encodings that SV Node rejects were silently accepted** — a latent BIP34 consensus divergence.

## Fix

Decode the first push with real push-opcode semantics (`OP_0`, `OP_1NEGATE`, `OP_1..OP_16`, direct pushes, `OP_PUSHDATA1/2/4`) as a CScriptNum, reject negative / out-of-range values, then require the encoding to be the canonical minimal push by re-encoding the decoded value and comparing the consumed prefix byte-for-byte.

This is the exact equivalent of bitcoin-sv `ContextualCheckBlock`, which builds `expect = CScript() << nHeight` and requires the scriptSig to start with those bytes. Result: bit-for-bit accept/reject parity with SV Node. Teranode's parse-then-compare architecture is preserved — the parity lands entirely in the extractor, so `Block.Valid` and `validateCoinbaseHeight` (gating + `height != b.Height`) are unchanged.

### Coupled builder fix

`model/GetCoinbaseParts.go` `makeCoinbase1` emitted a **fixed 3-byte** height push (`0x03` + 3 LE bytes). That is non-minimal below height 32768 and truncates at/above 16777216, so the strict extractor would reject Teranode's own mined blocks on BIP34-at-0 networks. It now encodes the height canonically via the shared `EncodeCoinbaseHeightPush`, matching SV Node and the extractor.

## Encoding reference

| Height | Canonical push | Form |
|--------|----------------|------|
| 0 | `00` | OP_0 |
| 1–16 | `51`..`60` | OP_1..OP_16 |
| 17 | `01 11` | 1-byte ScriptNum |
| 128 | `02 80 00` | sign-bit pad |
| 518847 | `03 bf ea 07` | 3-byte ScriptNum |

Rejected (non-canonical): `02 01 00` (should be OP_1), `04 bf ea 07 00`, `4c 03 bf ea 07` (OP_PUSHDATA1), `4f` (OP_1NEGATE), pushes > 8 bytes.

## Tests

- `util`: full accept/reject tables at script and tx level; teratestnet block-1 regression (`51 00 0f 22…` → height 1, `/Galts-Gulch/`); cross-check round-trip against an independent reference encoder for 14 heights (0/1/16/17/…/MaxUint32).
- `model`: producer→consumer round-trip through `GetCoinbaseParts`/`BuildCoinbase` across the encoding boundaries.

## Verification

`go build ./...`, `go vet`, `gofmt`, `gci`, `golangci-lint` clean. Green: `util`, `model`, `stores/blockchain/sql`, `services/blockvalidation`, `services/validator`, `services/subtreevalidation`, `services/asset/repository`, `services/blockassembly` (+ mining), `util/merkleproof`, `util/bump`.

Live teratestnet spot-checked (heights 1 / 1000 / 26800) — all canonical, so strict validation accepts every existing block and unblocks IBD with no re-sync risk for that network.

## Deploy note

Only throwaway internal dev clusters previously mined with the old non-minimal `makeCoinbase1` output at low heights can't re-IBD from genesis under strict parity and need a reset (SV Node would reject those chains too). teratestnet and mainnet are unaffected.

### DIFF
--- a/model/Block_test.go
+++ b/model/Block_test.go
@@ -2099,8 +2099,16 @@ func TestGenesisBytesFromModelBlock(t *testing.T) {
 
 func TestBlock_ExtractCoinbaseHeight(t *testing.T) {
 	t.Run("valid coinbase with height", func(t *testing.T) {
-		// Use the existing coinbase transaction from the test constants
-		coinbaseTx, err := bt.NewTxFromString(CoinbaseHex)
+		// Build a canonical coinbase via the production path. GetCoinbaseParts/makeCoinbase1 encode
+		// the height minimally (matching SV Node), so it round-trips through the strict extractor.
+		// The shared CoinbaseHex constant uses a legacy non-minimal 3-byte height push and is
+		// deliberately not used here (extracting it would now, correctly, fail SV Node parity).
+		const wantHeight = uint32(1019)
+
+		c1, c2, err := GetCoinbaseParts(wantHeight, 5000000000, "/m2-us/", []string{"1DkmRkb5iQFkDu4NBysog5bugnsyx7kwtn"})
+		require.NoError(t, err)
+
+		coinbaseTx, err := bt.NewTxFromBytes(BuildCoinbase(c1, c2, "0000000000000000", "00000000"))
 		require.NoError(t, err)
 
 		blockHeaderBytes, _ := hex.DecodeString(block1Header)
@@ -2112,7 +2120,7 @@ func TestBlock_ExtractCoinbaseHeight(t *testing.T) {
 
 		height, err := b.ExtractCoinbaseHeight()
 		require.NoError(t, err)
-		assert.Equal(t, uint32(1019), height) // Height extracted from coinbase scriptSig
+		assert.Equal(t, wantHeight, height) // Height extracted from coinbase scriptSig
 	})
 
 	t.Run("no coinbase transaction", func(t *testing.T) {
--- a/model/GetCoinbaseParts.go
+++ b/model/GetCoinbaseParts.go
@@ -182,7 +182,7 @@ func makeCoinbase1(height uint32, coinbaseText string) []byte {
 	// BIP34 block height, encoded canonically (CScript() << nHeight) so it is minimally encoded and
 	// accepted by SV Node and by Teranode's own coinbase-height parser. A fixed 3-byte push is
 	// non-minimal below height 32768 and truncates above 16777215.
-	arbitraryData = append(arbitraryData, util.EncodeCoinbaseHeightPush(int64(height))...)
+	arbitraryData = append(arbitraryData, util.EncodeCoinbaseHeightPush(height)...)
 	arbitraryData = append(arbitraryData, []byte(coinbaseText)...)
 
 	// Arbitrary data should leave enough space for the extra nonce
--- a/model/GetCoinbaseParts.go
+++ b/model/GetCoinbaseParts.go
@@ -44,6 +44,7 @@ import (
 	safeconversion "github.com/bsv-blockchain/go-safe-conversion"
 	base58 "github.com/bsv-blockchain/go-sdk/compat/base58" //nolint:depguard
 	"github.com/bsv-blockchain/teranode/errors"
+	"github.com/bsv-blockchain/teranode/util"
 )
 
 // BuildCoinbase recombines the different parts of the coinbase transaction.
@@ -177,12 +178,11 @@ func makeCoinbaseOutputTransactions(coinbaseValue uint64, walletAddresses []stri
 func makeCoinbase1(height uint32, coinbaseText string) []byte {
 	spaceForExtraNonce := 12
 
-	blockHeightBytes := make([]byte, 4)
-	binary.LittleEndian.PutUint32(blockHeightBytes, height) // Block height
-
 	arbitraryData := []byte{}
-	arbitraryData = append(arbitraryData, 0x03)
-	arbitraryData = append(arbitraryData, blockHeightBytes[:3]...)
+	// BIP34 block height, encoded canonically (CScript() << nHeight) so it is minimally encoded and
+	// accepted by SV Node and by Teranode's own coinbase-height parser. A fixed 3-byte push is
+	// non-minimal below height 32768 and truncates above 16777215.
+	arbitraryData = append(arbitraryData, util.EncodeCoinbaseHeightPush(int64(height))...)
 	arbitraryData = append(arbitraryData, []byte(coinbaseText)...)
 
 	// Arbitrary data should leave enough space for the extra nonce
--- a/model/GetCoinbaseParts_test.go
+++ b/model/GetCoinbaseParts_test.go
@@ -3,9 +3,12 @@ package model
 import (
 	"encoding/binary"
 	"encoding/hex"
+	"fmt"
 	"testing"
 
+	"github.com/bsv-blockchain/go-bt/v2"
 	"github.com/bsv-blockchain/teranode/errors"
+	"github.com/bsv-blockchain/teranode/util"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -439,6 +442,35 @@ func TestMakeCoinbase1_EdgeCases(t *testing.T) {
 		// The difference should be less than the full long text length
 		assert.Less(t, resultLongDiff, len(longText))
 	})
+}
+
+// TestMakeCoinbase1HeightRoundTrip verifies that the height Teranode encodes into its own coinbase
+// (via makeCoinbase1 → GetCoinbaseParts → BuildCoinbase) is decoded back to the same value by the
+// coinbase-height extractor, across the small-int, single-byte, multi-byte and sign-pad encoding
+// boundaries. This is the producer/consumer contract the strict (SV Node parity) extractor depends
+// on: a non-minimal producer would make Teranode reject its own mined blocks.
+func TestMakeCoinbase1HeightRoundTrip(t *testing.T) {
+	const walletAddress = "1DkmRkb5iQFkDu4NBysog5bugnsyx7kwtn"
+
+	heights := []uint32{1, 2, 16, 17, 100, 127, 128, 255, 256, 1000, 12345, 32767, 32768, 518847, 4294967295}
+
+	for _, h := range heights {
+		t.Run(fmt.Sprintf("height_%d", h), func(t *testing.T) {
+			c1, c2, err := GetCoinbaseParts(h, 5000000000, "/teranode/", []string{walletAddress})
+			require.NoError(t, err)
+
+			// 12 bytes of extra nonce, matching the space makeCoinbase1 reserves in the scriptSig length.
+			coinbase := BuildCoinbase(c1, c2, "0000000000000000", "00000000")
+
+			tx, err := bt.NewTxFromBytes(coinbase)
+			require.NoError(t, err)
+			require.True(t, tx.IsCoinbase())
+
+			height, err := util.ExtractCoinbaseHeight(tx)
+			require.NoError(t, err)
+			require.Equal(t, h, height)
+		})
+	}
 }
 
 func TestMakeCoinbase2_Comprehensive(t *testing.T) {
--- a/services/validator/Validator_test.go
+++ b/services/validator/Validator_test.go
@@ -132,8 +132,11 @@ func TestValidate_CoinbaseTransaction(t *testing.T) {
 		panic(err)
 	}
 
-	height, err := util.ExtractCoinbaseHeight(coinbase)
-	require.NoError(t, err)
+	// model.CoinbaseHex encodes height 1019 with a legacy non-minimal 3-byte push, which the strict
+	// (SV Node parity) extractor now rejects. This test only needs a block-height context to exercise
+	// that Validate refuses a coinbase transaction, so use the height literally rather than extracting
+	// it. Height-extraction parity is covered in util/coinbase_test.go and model/*_test.go.
+	const height = uint32(1019)
 
 	_, err = v.Validate(context.Background(), coinbase, height, WithSkipPolicyChecks(true))
 	require.Error(t, err)
--- a/util/coinbase.go
+++ b/util/coinbase.go
@@ -84,8 +84,8 @@ func extractCoinbaseHeightAndText(sigScript bscript.Script, raw bool) (uint32, s
 	}
 
 	// Enforce canonical minimal encoding (SV Node prefix-equality parity): the consumed prefix must
-	// equal CScript() << value byte-for-byte.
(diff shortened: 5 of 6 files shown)


ordishs (approved):
Correct fix for a real IBD-blocking bug (heights 1–16 misread as length prefixes) plus a genuine latent consensus divergence (non-minimal / OP_PUSHDATA-prefixed pushes silently accepted).

Verified the parity argument against bitcoin-sv `ContextualCheckBlock`: decode-then-re-encode-and-compare is equivalent to the `expect = CScript() << nHeight` prefix-equality check, since canonical encoding is injective and the caller's `height != b.Height` comparison closes the loop. `EncodeCoinbaseHeightPush` faithfully mirrors `push_int64`.

Edge cases traced and confirmed safe: overflow (maxHeightBytes checked as uint64 before any int() conversion), no slice panics (every push path validates length before slicing), decodeScriptNum magnitude fits int64, range gate catches OP_1NEGATE/sign-bit values before the minimal check. Both consensus call sites gate on `Height > 0 && >= BIP0034Height`, so the stricter extractor only runs where SV Node runs it — no mainnet re-sync risk.

Test coverage is strong, including an independently-written reference encoder that cross-checks the production encoder rather than testing it against itself.

Ran locally: `go test -tags testtxmetacache` for the coinbase tests in ./util and ./model — both pass.

Minor, non-blocking observations:
- `compose/chainintegrity/chain_integrity.go` calls `ExtractCoinbaseHeight` ungated by BIP34; for a pre-BIP34 coinbase whose first byte is a valid push opcode it now returns a plausible-but-wrong height and logs a mismatch rather than an extraction failure. Diagnostic/log-only.
- `ExtractCoinbaseMiner` splits the height off differently for such legacy coinbases (display-only).

freemans13 (approved):
Reviewed for consensus correctness — approving. This is a faithful, well-reasoned fix.

## Consensus parity — verified

The decode-then-re-encode-and-compare approach is bit-for-bit equivalent to SV Node's `ContextualCheckBlock` (`expect = CScript() << nHeight`, prefix-equality). I worked through it against the SV Node reference across every boundary height (0, 1–16, 17, 128, 32767/32768, 16777215/16, MaxUint32, including the sign-pad byte) and every adversarial input (OP_1NEGATE, OP_0, small-int direct pushes, OP_PUSHDATA1/2/4 wrappers, non-minimal multi-byte, oversized >8 bytes, negative/sign-bit, >MaxUint32, empty script) — accept/reject matches SV Node in every case. `decodeScriptNum` is a true inverse of `encodeScriptNumMinimal` on canonical inputs, so canonical pushes round-trip and everything non-canonical is rejected. No panic/overflow in the 8-byte decode path and no index-out-of-range on truncated OP_PUSHDATA headers.

## Testing (local)

- `util` + `model` unit tests: pass (1215), covering the changed extractor, the coupled `makeCoinbase1` builder, and the producer→consumer round-trip.
- `go vet`, `gofmt`, `golangci-lint` on the changed files: clean.
- `services/validator` and `stores/blockchain/sql`: only environmental failures locally (Aerospike-via-Docker not available; sql suite exceeded the 10m default timeout) — no test-level failures attributable to this change.

## Advisory (non-blocking, no changes requested)

1. **Reject reason differs from SV Node.** Oversized/short/empty pushes surface as `ErrBlockCoinbaseMissingHeight` rather than SV's `bad-cb-height`. The accept/reject *decision* is identical; only the logged reason differs (deliberate, so the miner-display path can suppress missing-height). Worth knowing when diffing reject logs across nodes.
2. **Parity is a two-part contract.** This file decodes + canonical-checks; the height *equality* check and the BIP34 activation gate live in `Block.Valid` (`model/Block.go:641`). A future caller of `ExtractCoinbaseHeight` that bypasses that gate would diverge from SV for pre-BIP34 blocks — a one-line doc note on the exported function would guard against that.
3. **`model.CoinbaseHex` is now a latent test trap.** The shared constant is a non-minimal legacy push that the strict extractor correctly rejects. This PR updated the two tests that assert on extraction; the other tests using it don't drive it through the strict path, so they still pass — but any future test that feeds `CoinbaseHex` to `ExtractCoinbaseHeight` will fail confusingly. A comment on the constant (or a canonical replacement) would help.
4. **Minor dead code.** The negative-sign branches in `encodeScriptNumMinimal` are unreachable on the height path (heights are always ≥0). Correct, just dead.

None of these block merge.

ctnguyen (approved):
Review at commit [636bc6f](https://github.com/oskarszoon/teranode/commit/636bc6f8d5e0e37c1e7970b2e6ed7171244d5118)

This is a well-scoped, correct fix for a real IBD-blocking bug: reading the first coinbase scriptSig byte as a raw length prefix mis-handled `OP_1..OP_16`, which is exactly the canonical BIP34 push for heights 1–16, so a from-genesis teratestnet sync (BIP34Height=0) stalled at stored height 0 the moment #1301 started running extraction from block 1. The rewrite decodes with real push-opcode semantics and then enforces canonicality by re-encoding and comparing byte-for-byte — which is the right way to mirror SV Node here. I checked the parity claim against svnode source rather than the PR narrative: `EncodeCoinbaseHeightPush`/`encodeScriptNumMinimal` reproduce `CScript::push_int64` (src/script/script.cpp:224-242) and the `bsv::serialize` CScriptNum sign-magnitude/sign-pad rules (src/script/int_serialization.h:24-59) for non-negative heights, and the decode-then-re-encode-equality construction is byte-for-byte equivalent to `ContextualCheckBlock`'s `expect = CScript() << nHeight` prefix check (src/validation.cpp:6038-6054), including its rejection of non-minimal, `OP_PUSHDATA`-prefixed, and negative encodings. Note one framing caveat: svnode computes the contextual `nHeight` as an `int32_t`, so the top of the tested range (`4294967295`) is not a height svnode can ever contextually produce — that case exercises Teranode's own internal `uint32` height boundary and the generic `push_int64` encoding, not a literal svnode contextual-height value; it's still a good boundary to test, just not "SV Node parity" in the strict sense. I also confirmed the parse-then-compare architecture is preserved: both consensus call sites still perform the equality compare (`height != b.Height` at model/Block.go:647, `blockHeight != currentHeight` at stores/blockchain/sql/StoreBlock.go:925) under an unchanged BIP34Height gate with the genesis guard, so a canonical-but-wrong height cannot slip through; the coupled `makeCoinbase1` change is self-consistent with the 12-byte extranonce reservation and the 2–100-byte scriptSig bounds; and the current `decodeCoinbaseHeightPush` bounds-checks before every slice, so no malformed push panics today. Two non-blocking observations below.

## Non-blocking issues

### ChiR1 - Exported encoder negative/MinInt64 panic

**Exported `EncodeCoinbaseHeightPush` is documented as a `push_int64` mirror but isn't one for negatives — and panics on `math.MinInt64`.**

**Problem:** `EncodeCoinbaseHeightPush(height int64)` and its helper `encodeScriptNumMinimal(n int64)` are exported/reachable with an arbitrary `int64`, and the doc comment advertises parity with "bitcoin-sv's `CScript() << nHeight` (push_int64)". Two things diverge from that contract on negative input:

- `push_int64` maps `-1` to `OP_1NEGATE` (`0x4f`). `EncodeCoinbaseHeightPush(-1)` instead falls to the default branch and emits a direct ScriptNum push `01 81`. So it is *not* a `push_int64` mirror for negatives.
- `encodeScriptNumMinimal(math.MinInt64)` panics. `n = -n` overflows and stays `math.MinInt64` (negative), so the `for n > 0` loop never executes, `result` is left empty, and the trailing `switch` dereferences `result[len(result)-1]` on a zero-length slice — an index-out-of-range panic.

Today's consensus callers are safe: the extractor only re-encodes a `value` already guarded to `[0, MaxUint32]`, and `makeCoinbase1` passes a `uint32`. So this is not a current block-acceptance defect and stays non-blocking. But it is more than a doc nit — the function is exported from `util`, and any future non-consensus caller (miner tooling, tests, a display path) that passes a negative or `MinInt64` `int64` gets a wrong encoding or a crash.

**Why it matters:** A consensus-adjacent encoder that carries an incorrect, untested negative path is a latent trap: it invites a future caller to assume full `push_int64` semantics (they don't hold), and the `MinInt64` case turns a bad argument into a panic rather than a clean error. It also leaves the three `negative`-only branches in `encodeScriptNumMinimal` permanently unexercised.

**Fix:** Make the non-negative contract explicit in the type, which also deletes the dead sign branches and the panic path:

```go
// EncodeCoinbaseHeightPush returns the canonical BIP34 coinbase-height push for
// height, matching bitcoin-sv's CScript() << nHeight for non-negative heights.
func EncodeCoinbaseHeightPush(height uint32) []byte { ... }

// encodeScriptNumMinimal encodes a non-negative n as a minimal little-endian
// CScriptNum (the byte form pushed by bitcoin-sv's CScriptNum::serialize).
func encodeScriptNumMinimal(n uint32) []byte {
    var result []byte
    for n > 0 {
        result = append(result, byte(n&0xff))
        n >>= 8
    }
    if len(result) > 0 && result[len(result)-1]&0x80 != 0 {
        result = append(result, 0x00) // sign-bit pad
    }
    return result
}
```

(The extractor already holds a validated `value` in `[0, MaxUint32]`, so `EncodeCoinbaseHeightPush(uint32(value))` is a safe call site.) If you prefer to keep the `int64` signature, guard `n < 0` at the top of both functions and return an error/`nil` instead of falling through. Either way, if the negative path is meant to stay, add the `-1 → OP_1NEGATE` case and a negative round-trip test so the branches are actually covered.

### ChiR2 - Missing malformed-push-boundary tests

**No test locks down the malformed / truncated push-boundary inputs the parser is claimed to survive.**

**Problem:** `decodeCoinbaseHeightPush` makes an implicit no-panic/no-overflow guarantee for a consensus parser, and the code does bounds-check correctly today — `len(s) < 2/3/5` before reading `s[1]`, `s[1:3]`, `s[1:5]`, and `len(s) < end` before `s[headerLen:end]`. But the accept/reject tables don't exercise those guards, so a future refactor could silently reintroduce an out-of-range slice. Missing cases:

- `OP_DATA<n>` (e.g. `03 bf ea`) declaring more bytes than remain.
- `OP_PUSHDATA1/2/4` with a truncated header (`4c`, `4d 03`, `4e 03 00 00`) — i.e. the header length bytes themselves are missing.
- `OP_PUSHDATA1/2/4` with a declared length `<= maxHeightBytes` but a truncated payload (e.g. `4c 03 bf ea` — declares 3 data bytes, only 2 follow).

The existing "too short" tests cover the empty script and `"a90102"` (non-push opcode), but not these push-length-vs-remaining-bytes boundaries.

**Why it matters:** The scriptSig is attacker-controlled data on the block-validation path. The no-panic property is currently real, but it's asserted only by inspection; a boundary test suite is what keeps it real across future edits. This is a coverage gap, not a live bug.

**Fix:** Add reject cases to `TestExtractCoinbaseHeightAndTextScripts` (which already asserts `ErrorIs(err, errors.ErrBlockCoinbaseMissingHeight)`), e.g.:

```go
{name: "OP_DATA3 with only 2 trailing bytes", script: "03bfea", expectError: true},
{name: "OP_PUSHDATA1 with no length byte", script: "4c", expectError: true},
{name: "OP_PUSHDATA2 with truncated length", script: "4d03", expectError: true},
{name: "OP_PUSHDATA4 with truncated length header", script: "4e030000", expectError: true},
{name: "OP_PUSHDATA4 declares 3 bytes, payload truncated", script: "4e03000000", expectError: true},
{name: "OP_PUSHDATA1 declares 3 bytes, 2 present", script: "4c03bfea", expectError: true},
```

## Recap

| ChiR1 | Exported encoder negative/MinInt64 panic | | 40% |
| ChiR2 | Missing malformed-push-boundary tests | | 22% |

oskarszoon:
Both non-blocking items addressed in f0b7f2f84.

- **ChiR1**: `EncodeCoinbaseHeightPush` and `encodeScriptNumMinimal` now take `uint32`. That makes the non-negative contract explicit in the type and removes both the incorrect negative path (`-1` → `01 81` instead of `OP_1NEGATE`) and the `math.MinInt64` panic — the dead sign branches are gone. Call sites: the extractor passes `uint32(value)` (already guarded to `[0, MaxUint32]`), `makeCoinbase1` passes the `uint32` height directly.
- **ChiR2**: added truncated-push / `OP_PUSHDATA1/2/4` boundary reject cases (`03bfea`, `4c`, `4c03bfea`, `4d03`, `4e030000`, `4e03000000`) to `TestExtractCoinbaseHeightAndTextScripts`, so the parser's bounds checks on attacker-controlled scriptSig data are covered by tests rather than inspection.

`util` + `model` green (`-count=1`), build/vet/gofmt/gci/golangci-lint clean.

PR #1342, opened 2026-07-25
fix(rewindblockchain): never move BlockPersisterHeight forward on rewind (#1340)

### LINKED ISSUE #1340: rewindblockchain: resetBlockPersisterHeight moves the persister forward when it is behind the rewind target, silently skipping persistence
## What

`cmd/rewindblockchain` rewrites `state["BlockPersisterHeight"]` to the rewind target unconditionally. When the block persister is *behind* the target — normal on a recently-resynced node — this moves it **forward**, and every block between its real position and the target is never persisted. Silent: no warning, no error, and the summary log reports it as a normal rewrite.

`cmd/rewindblockchain/rewindblockchain/phase3_finalize.go:57`:

```go
func (e *env) resetBlockPersisterHeight(ctx context.Context, pf *preflightResult) error {
	existing, err := e.blockchainStore.GetState(ctx, blockPersisterHeightKey)
	...
	payload := make([]byte, 4)
	binary.LittleEndian.PutUint32(payload, pf.target)   // ← no comparison against existing
	if err = e.blockchainStore.SetState(ctx, blockPersisterHeightKey, payload); err != nil {
```

`existing` is read and checked for presence/emptiness, then never compared. The doc comment says "writes the persisted height **down** to the target", which is the intent; the code doesn't enforce it.

## Concrete case

testnet-eu-1, wedged at 1749337 by the pre-#1116 OP_RETURN UTXO divergence, needs a 3-block rewind to 1749334 so the affected parent tx gets re-created under the fixed predicate. `asset_service_heights`:

```
block_assembly_height:   1749337
block_persister_height:    422492
```

Running the tool as-is would set `BlockPersisterHeight = 1749334`, skipping persistence of ~1.33M blocks (422493–1749334).

The nodes most likely to need a rewind are the ones that recently resynced — exactly the ones whose persister lags furthest. Bug is present in `release/v0.15` and `main`.

## Fix

```go
if existingHeight := binary.LittleEndian.Uint32(existing); existingHeight <= pf.target {
	e.logger.Infof("state[%s]=%d already at/below target %d; leaving alone",
		blockPersisterHeightKey, existingHeight, pf.target)
	return nil
}
```

Guard `len(existing) >= 4` before the decode (the existing `len(existing) == 0` check isn't enough). Worth a `--verify` (phase 4) assertion that the persister height never increased across the run.

## Test

Rewind with `state["BlockPersisterHeight"]` below the target → assert the key is untouched. Above the target → assert it is rewritten down to the target.

oskarszoon:
Fixes #1340

## Problem

`resetBlockPersisterHeight` read `state["BlockPersisterHeight"]`, checked only that it was present and non-empty, then wrote `LE(target)` unconditionally. On a node whose block persister sits *below* the rewind target — normal after a resync — that moves the recorded height **forward**. No warning, no error; the summary log reports a normal rewrite. The function's own doc comment said it writes the height "down to the target"; the code never enforced it.

testnet-eu-1 shape from the issue: `block_assembly_height=1749337`, `block_persister_height=422492`. A 3-block rewind to 1749334 would have written `BlockPersisterHeight = 1749334`.

## Severity: data-loss-capable, not transient misreporting

An earlier revision of this description called the key "derived data" that block persister republishes on startup, so nothing is permanently lost. That is true of the *stored key* and false of its consumers. The chain, all verified in-tree:

1. The pruner reads the key **once, in `Init()`** (`services/pruner/server.go:250`) into `lastPersistedHeight`.
2. Its cached copy is afterwards updated **only** from `BlockPersisted` notification metadata (`:163-176`).
3. Block persister's corrective republish on startup is a **bare `SetState` with no notification** (`services/blockpersister/Server.go:329`) — unlike the runtime path (`:401-417`), which sends the notification first.

So a pruner that has latched an inflated value never learns better, and the key being repaired in the database does not repair the process acting on it. From there: `worker.go:200` → `blob_deletion_worker.go:75-99` computes `safeHeight = blockHeight - BlobDeletionSafetyWindow` and `blobStore.Del`s at `:232` — files off disk, no soft-delete. `worker.go:249` DAH-prunes UTXO records to the same inflated height, and `:263` advances `lastProcessedHeight` past it, so every subsequent real `BlockPersisted` is dropped by the `<= lastProcessedHeight` guard and pruning stalls for the whole gap.

The safety gates do not catch it, because the buggy tool wrote exactly `LE(target)` — the largest value that still resolves in `GetBlockHeadersByHeight(target, target)`, and the same value written to `state[BlockAssembler]`, so `checkBlockAssemblySafeForPruner` passes too. A value above the tip would have been rejected. Stock config is the exposed path: `pruner_block_trigger` defaults to `OnBlockPersisted`, `pruner_skipBlobDeletion` to `false`, `pruner_force_ignore_block_persister_height` to `false`.

Whether any given node actually lost blobs depends on rows existing in the scheduled-deletion table under the inflated height at that moment — code path verified, no live database inspected.

## Change

- New `readBlockPersisterHeight` helper: one decode/tolerance policy for all call sites. Absent / empty / shorter than 4 bytes → "unknown"; real storage errors propagate.
- `resetBlockPersisterHeight` writes only when the stored height is strictly above the target. At/below target, or undecodable, it logs and leaves the value alone.
- Preflight captures the pre-run height and **cross-checks it against ground truth** — the lowest block with `persisted_at IS NULL`, minus one, which is what block persister derives on startup (genesis skipped; it is never marked persisted, and block persister skips it the same way). A key sitting above that real position is the fingerprint of a pre-fix run, and this tool cannot repair it, so it says so and names the manual repair: delete the `state` row and let block persister republish on next start. The query is backed by `idx_not_persisted_height`, so it stays cheap. Preflight also warns when the persister lags the target by more than coinbase maturity, spelling out that the rewind will not move where it resumes.
- A genuine read failure on the key now aborts before Phase 0 instead of surfacing in Phase 3 with phases 0-2 already applied.
- `--verify` (Phase 4) fails the run if the height increased, **or** if a key that was unreadable going in has a value coming out — this tool never creates the key. Under `--force-not-idle` both are downgraded to warnings naming the cause, since a running block persister legitimately raises the key and a hard failure there would condemn a run that actually succeeded after all mutation is applied.

Kept as adjust-down rather than delete-the-row (the policy the sibling `deleteUTXOPersisterLastProcessed` uses for the utxo persister's position): the row is the operator's only record of where the persister was on a node they are already debugging, and discarding it to save a warning is a bad trade. The cross-check covers the case delete would have repaired.

No new flag, nothing touched outside `cmd/rewindblockchain/rewindblockchain/`.

## Consumers of the key

Four read sites, all gating on `len(...) >= 4`: `services/pruner/server.go:250`, `services/p2p/Server.go:1409`, `services/legacy/peer_server.go:3549`, and `services/asset/httpimpl/get_service_heights.go:31` — the last exposes it as `block_persister_height` over HTTP, i.e. an inflated key lies on the very endpoint an operator would hit to check whether the persister is behind.

Three writers, and only one could produce a value not derived from ground truth: `blockpersister/Server.go:329` (startup, from `persisted_at`), `:417` (runtime, from `block.Height`), and this tool. Fixing it here closes the class in-tree.

## Tests

`rewind_test.go`: below / equal / above target, 2-byte payload, empty payload, absent key — each asserting the `SetState` **call count**, not just the stored bytes, because at the equality boundary skipping the write and rewriting `LE(target)` leave identical bytes (the value-only assertion survived deleting the guard; the counter kills it). Ground-truth derivation, inflated-key warning, large-lag warning, and the preflight decision log. Phase 4: increase fails, unchanged passes, unreadable-before-set-after fails, `--force-not-idle` downgrades, lost-value reported as lost rather than as a live `0`. Plus a full `Rewind` run with `--verify` against a lagging persister height. The existing integration test still covers the write-down path.

`go test -race ./cmd/rewindblockchain/...` 33 pass. `make test` 11982 tests, 0 failures. `go vet` clean, `make lint` 0 issues.

## Backport

`release/v0.15` **and** `release/v0.16` both carry the identical pre-fix body. Both still use `%w` in the `errors.New*` format strings that `main` stripped in #1335, so the cherry-pick will conflict on exactly those lines in `resetBlockPersisterHeight` — resolve toward the release branch's `%w` style.

Follow-up, not in this PR: `BlockPersisterHeight` has no shared accessor, unlike `blockassembly.StateKey`/`EncodeState` — the key literal and the 4-byte LE encode/decode are hand-rolled at six sites.

### DIFF
--- a/cmd/rewindblockchain/rewindblockchain/phase3_finalize.go
+++ b/cmd/rewindblockchain/rewindblockchain/phase3_finalize.go
@@ -78,6 +78,38 @@ func (e *env) readBlockPersisterHeight(ctx context.Context) (uint32, bool, error
 	return binary.LittleEndian.Uint32(existing), true, nil
 }
 
+// blockPersisterGroundTruth derives the block persister's real position from the
+// blocks table — the lowest block with persisted_at IS NULL, minus one — which
+// is the same derivation block persister performs on startup
+// (services/blockpersister/Server.go:322-334). The state key is only a cache of
+// this; the table is the truth.
+//
+// known is false when there is no pending block above genesis to derive from,
+// or when the query fails. Diagnostics only, so a failure is logged and never
+// fatal. The query is backed by idx_not_persisted_height
+// (stores/blockchain/sql/sql.go:786), so it stays cheap on a full chain.
+//
+// Genesis is skipped: it carries persisted_at IS NULL for the life of the
+// database, so it would otherwise mask every real pending block. Block
+// persister skips it the same way, via its blocks[0].Height > 0 guard — hence
+// asking for two rows, since an unpersisted genesis always sorts first.
+func (e *env) blockPersisterGroundTruth(ctx context.Context) (uint32, bool) {
+	blocks, err := e.blockchainStore.GetBlocksNotPersisted(ctx, 2)
+	if err != nil {
+		e.logger.Debugf("could not derive the real block persister position: %v", err)
+		return 0, false
+	}
+
+	for _, b := range blocks {
+		if b.Height == 0 {
+			continue
+		}
+		return b.Height - 1, true
+	}
+
+	return 0, false
+}
+
 // resetBlockPersisterHeight moves the persisted height down to the target. It
 // never moves it forward: when the block persister is behind the target —
 // normal on a recently-resynced node — the value is left alone.
--- a/cmd/rewindblockchain/rewindblockchain/phase3_finalize.go
+++ b/cmd/rewindblockchain/rewindblockchain/phase3_finalize.go
@@ -52,20 +52,52 @@ func (e *env) resetBlockAssemblerState(ctx context.Context, pf *preflightResult)
 	return nil
 }
 
-// resetBlockPersisterHeight writes the persisted height down to the target.
-// The underlying value is a LE uint32 — see services/blockpersister/Server.go.
-func (e *env) resetBlockPersisterHeight(ctx context.Context, pf *preflightResult) error {
+// readBlockPersisterHeight reads and decodes state["BlockPersisterHeight"].
+// known is false when the key is absent, empty, or too short to decode — in
+// those cases the caller cannot assume anything about the persister's real
+// position. Genuine storage failures are returned as errors.
+func (e *env) readBlockPersisterHeight(ctx context.Context) (uint32, bool, error) {
 	existing, err := e.blockchainStore.GetState(ctx, blockPersisterHeightKey)
 	if err != nil {
 		// SQL blockchain store returns sql.ErrNoRows directly for missing keys;
 		// future implementations may wrap into errors.ErrNotFound.
 		if errors.Is(err, sql.ErrNoRows) || errors.Is(err, errors.ErrNotFound) {
 			e.logger.Debugf("state[%s] not present: %v", blockPersisterHeightKey, err)
-			return nil
+			return 0, false, nil
 		}
-		return errors.NewStorageError("failed to read state[%s]", blockPersisterHeightKey, err)
+		return 0, false, errors.NewStorageError("failed to read state[%s]", blockPersisterHeightKey, err)
 	}
+
 	if len(existing) == 0 {
+		return 0, false, nil
+	}
+
+	if len(existing) < 4 {
+		e.logger.Warnf("state[%s] is %d bytes, too short to decode as LE uint32; leaving it untouched", blockPersisterHeightKey, len(existing))
+		return 0, false, nil
+	}
+
+	return binary.LittleEndian.Uint32(existing), true, nil
+}
+
+// resetBlockPersisterHeight moves the persisted height down to the target. It
+// never moves it forward: when the block persister is behind the target —
+// normal on a recently-resynced node — the value is left alone. Raising it
+// would tell the pruner's safe-prune calculation and the p2p storage-mode check
+// that blocks were persisted when they were not. The underlying value is a LE
+// uint32 — see services/blockpersister/Server.go.
+func (e *env) resetBlockPersisterHeight(ctx context.Context, pf *preflightResult) error {
+	existingHeight, known, err := e.readBlockPersisterHeight(ctx)
+	if err != nil {
+		return err
+	}
+
+	if !known {
+		return nil
+	}
+
+	if existingHeight <= pf.target {
+		e.logger.Infof("state[%s]=%d already at/below target %d; leaving it untouched", blockPersisterHeightKey, existingHeight, pf.target)
 		return nil
 	}
 
@@ -76,7 +108,7 @@ func (e *env) resetBlockPersisterHeight(ctx context.Context, pf *preflightResult
 		return errors.NewStorageError("failed to rewrite state[%s]", blockPersisterHeightKey, err)
 	}
 
-	e.logger.Infof("state[%s] rewritten to %d", blockPersisterHeightKey, pf.target)
+	e.logger.Infof("state[%s] rewritten from %d down to %d", blockPersisterHeightKey, existingHeight, pf.target)
 	return nil
 }
 
--- a/cmd/rewindblockchain/rewindblockchain/phase3_finalize.go
+++ b/cmd/rewindblockchain/rewindblockchain/phase3_finalize.go
@@ -68,10 +68,8 @@ func (e *env) readBlockPersisterHeight(ctx context.Context) (uint32, bool, error
 		return 0, false, errors.NewStorageError("failed to read state[%s]", blockPersisterHeightKey, err)
 	}
 
-	if len(existing) == 0 {
-		return 0, false, nil
-	}
-
+	// A present-but-empty row (len 0) is as undecodable as a truncated one, and
+	// is the more anomalous of the two — both take the same warning.
 	if len(existing) < 4 {
 		e.logger.Warnf("state[%s] is %d bytes, too short to decode as LE uint32; leaving it untouched", blockPersisterHeightKey, len(existing))
 		return 0, false, nil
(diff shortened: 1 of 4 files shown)


ctnguyen (approved):
Review at commit [9465dd5](https://github.com/oskarszoon/teranode/commit/9465dd52007c795e957cea25ccb763a38079d858)

This is a tight, well-scoped fix for a genuinely nasty bug, and the diagnosis in the issue is exactly right: `resetBlockPersisterHeight` read the existing value and then never used it. The core guard (`existingHeight <= pf.target` → leave alone) is correct, and I checked the direction-of-harm claims in the new doc comment against the actual consumers — `services/pruner/server.go:250` seeds `lastPersistedHeight` straight from the key, and `services/p2p/Server.go:1409` / `services/legacy/peer_server.go:3549` both feed it into `DetermineStorageMode`, so a raised value really does make a node claim persistence it doesn't have. Factoring the decode into `readBlockPersisterHeight` was the right call: the `len(existing) < 4` → unknown policy now matches the `len(state) >= 4` tolerance every consumer already applies, and decoding the first 4 bytes of a longer payload matches them too. Keeping Phase 3's decision on a fresh read rather than the preflight snapshot is also the right choice — it means a key that appears mid-run above the target still gets corrected. Capturing the height in preflight to fail fast before Phase 0 is a nice bonus, the `--verify` assertion is the right kind of belt-and-braces, and the test set covers each branch of the new policy including the 2-byte and absent-key cases. Nothing here blocks merge; the notes below are refinements around the edges of the new verify/preflight wiring.

## Non-blocking issues

### ChiR1 — Phase 4 passes, and logs a false claim, when the key stops decoding

**Problem:** In `cmd/rewindblockchain/rewindblockchain/phase4_verify.go:35-39`, the assertion is gated on `known`, but the log line that follows is not. If the baseline was known and the post-run read comes back unknown (key deleted, or truncated to fewer than 4 bytes), `current` is `0`, the comparison is skipped, and `--verify` prints `verify: state[BlockPersisterHeight]=0 did not increase (was 422492)` before returning success. The value is not 0 — it is gone or undecodable, which is a state the run cannot vouch for.

**Why it matters:** `--verify` exists to catch the "shouldn't happen" transitions. A baseline-known → post-run-unknown transition is precisely one of those, and right now it is reported as a clean pass with a factually wrong height in the log. An operator reading the run output would have no signal that the key needs attention.

**Fix:** Handle `!known` explicitly rather than letting it fall through to the success log:

```go
if pf.persisterHeightKnown {
    current, known, readErr := e.readBlockPersisterHeight(ctx)
    if readErr != nil {
        return errors.NewStorageError("verify: read state[%s]", blockPersisterHeightKey, readErr)
    }

    if !known {
        return errors.NewProcessingError("verify: state[%s] was %d before the run and is now absent or undecodable",
            blockPersisterHeightKey, pf.persisterHeight)
    }

    if current > pf.persisterHeight {
        return errors.NewProcessingError("verify: state[%s] increased from %d to %d", blockPersisterHeightKey, pf.persisterHeight, current)
    }

    e.logger.Infof("verify: state[%s]=%d did not increase (was %d)", blockPersisterHeightKey, current, pf.persisterHeight)
}
```

### ChiR2 — end-to-end test for the issue shape deletes no blocks

**Problem:** `TestRewind_PersisterHeightBelowTargetLeftAlone` (`rewind_test.go:1127`) sets `TargetHeight: 4` against a 4-block chain, so target equals the tip and the test asserts `require.Zero(t, stats.BlocksDeleted)`. The issue-1340 shape is target *below* tip with a lagging persister; `TestRewindBlockchain_Integration` covers target < tip but seeds the persister at the tip (`rewind_test.go:461-465`), i.e. the rewrite-down path. So the combination "blocks actually deleted **and** persister behind target" is never run.

**Why it matters:** The one configuration that will be run against testnet-eu-1 is the one no test exercises end to end. It is also the configuration where a future change to Phase 2 or to the Phase 4 baseline could regress the fix without any test noticing.

**Fix:** This is not a one-line `TargetHeight` change — worth calling out, because the obvious edit fails. `buildLinearChain` populates `block.Subtrees` with the coinbase txid but never writes a subtree blob (its own comment at `rewind_test.go:636-637` says preflight never opens them), while `rewindOneBlock` calls `block.GetSubtrees(...)` at `phase2_blocks.go:31` and returns a `StorageError` when the blob is missing. Every existing `buildLinearChain` test therefore stops in preflight, dry-run, or Phase 4. Two workable routes:

- Reuse the richer fixture: factor the seeding in `TestRewindBlockchain_Integration` (which builds real subtree blobs via `makeSubtreeAndStore`, `rewind_test.go:112`) into a helper parameterized by the persister-height seed, then add a case that seeds it below the target instead of at the tip. Cheapest in coverage terms, since that test already asserts the full post-rewind state.
- Or seed valid subtree blobs in `buildLinearChain` (behind a flag so the preflight-only tests keep their current cost), which would unblock below-tip targets for the whole family of tests.

Either way the new assertion is the same: after a run that deletes blocks, `state[BlockPersisterHeight]` still decodes to the lagging value.

### ChiR3 — the confirmation prompt doesn't mention the persister height

**Problem:** `confirmPrompt` (`preflight.go:169-177`) prints tip, target and block count, but not what will happen to `BlockPersisterHeight`. The height is read a few lines earlier and only goes to the log; the decision the operator most needs to see for this run is the one absent from the prompt they are approving.

**Why it matters:** This is the tool's only interactive gate on a destructive, irreversible operation. After this PR the key's handling is conditional, and "left untouched because it is behind the target" vs "rewritten down to target" are materially different outcomes for an operator deciding whether the node is in the state they think it is.

**Fix:** Add one line to the summary:

```go
switch {
case !r.persisterHeightKnown:
    fmt.Fprintf(e.opts.Stdout, "  BlockPersisterHeight: not set/undecodable — left untouched\n")
case r.persisterHeight <= r.target:
    fmt.Fprintf(e.opts.Stdout, "  BlockPersisterHeight: %d (behind target) — left untouched\n", r.persisterHeight)
default:
    fmt.Fprintf(e.opts.Stdout, "  BlockPersisterHeight: %d -> %d\n", r.persisterHeight, r.target)
}
```

Worth echoing in `logStats` too, so the summary line records the outcome.

### ChiR4 — preflight states an outcome Phase 3 has not decided yet

**Problem:** `preflight.go:90` logs `state[BlockPersisterHeight] not set or undecodable; it will be left untouched` as a statement of fact, but Phase 3 makes its own decision from a fresh read. `--force-not-idle` is a supported flag, so the block persister can be live and publish the key between the two reads (it does exactly that on startup, `services/blockpersister/Server.go:322-334`) — in which case Phase 3 correctly rewrites it down, contradicting what preflight told the operator, and Phase 4 skips its assertion because the recorded baseline is `known == false`.

**Why it matters:** Small, but it is operator-facing text on a destructive tool, and it is the one line that describes what will happen to the key. Reading it as a promise and then finding the value rewritten is exactly the kind of confusion this PR is trying to remove.

**Fix:** State the observation, not the outcome — and note the write is unaudited in that case:

```go
e.logger.Infof("state[%s] not set or undecodable at preflight; Phase 3 will re-read and will only ever lower it", blockPersisterHeightKey)
```

## Recap

|---|---|---|---|
| ChiR1 | Verify passes on undecodable key |  | 30% |
| ChiR2 | End-to-end test deletes nothing |  | 20% |
| ChiR3 | Confirm prompt omits persister height |  | 15% |
| ChiR4 | Preflight log over-promises outcome |  | 10% |

icellan (approved):

PR #1347, opened 2026-07-27
p2p: run peer health checks concurrently with caching and caller context

gokutheengineer:
**Closes bitcoin-sv/teranode#4725.**

### Problem
`PeerSelector.SelectSyncPeer` was documented as "a pure function - no side effects, no network calls" but performed a synchronous HTTP health check (2s timeout) per candidate peer, sequentially, inside `isEligible`, using `context.Background()`. The pruned-node fallback pass re-probed peers already probed in the full-node pass, so selecting among N peers cost up to 2N sequential 2s round-trips on the sync coordinator's monitor loop, starving catchup, and the probes ignored shutdown. All items in the issue were confirmed still valid on main.

### Fix
- Threaded the coordinator context through `SelectSyncPeer` (the sync coordinator passes `sc.ctx`). `Stop` already cancelled `sc.ctx` for registry/blockchain RPCs on main; the contribution here is the threading, so in-flight availability probes now ride that existing cancellation and abort on shutdown instead of stalling `Stop`.
- Hoisted availability probing into a single pre-pass over basic-eligible peers: DataHub URLs are deduplicated and probed on a worker pool (at most 8 goroutines) under an overall 5s budget, so one selection tick spends at most ~5s on probes regardless of peer count (previously up to 2N x 2s). Full-node-tier URLs are enqueued ahead of pruned-tier ones, so a queue of slow pruned peers cannot exhaust the budget before a healthy full node is probed (regression-tested).
- Basic (network-free) eligibility is computed once per selection (`basicEligibleByID`) and shared by the pre-pass and both candidate passes, instead of being re-derived (and debug-logged) up to three times per tick. Eligibility is therefore frozen at selection start; the "not probed" branch in `isEligible` is unreachable today and kept purely as a fail-closed defensive guard against a future refactor that unfreezes eligibility. Its lack of test coverage is deliberate - do not try to "fix" it.
- Probe results are cached for a 10s TTL, eliminating the double probe across the full/pruned passes and across closely spaced ticks. Each entry is stamped with its own probe's completion time (not the end of the round), so a fast probe's entry cannot outlive the TTL by inheriting the slowest probe's finish time, and the write-back keeps whichever probe completed later, so a concurrent round cannot overwrite a fresher entry with a staler one. Probes cut short by ctx or the budget resolve as unhealthy for that selection but are not cached, so slow peers are not remembered as down; expired entries are pruned to keep the cache bounded.
- Deliberate drive-by, separable from the health-check work and included here from review feedback rather than split into its own PR: `SyncCoordinator.warnfUnlessStopping` downgrades RPC-failure warnings to debug once the coordinator is stopping, since `Stop`'s pre-existing cancellation makes in-flight calls fail with an expected `context.Canceled`. Applied at the RPC-failure log sites (`ListPeers`, `GetPeer`, `GetBestBlockHeader`, `RecordSyncAttempt`, `RemovePeer`, `UpdatePeerMetrics`, `RegisterPeer`, `IsPeerBanned`, `ClearAllSyncAttempts`, `ReconsiderBadPeers`) so normal shutdown does not emit warn-level noise. Byte-identical behaviour during normal operation.
- Corrected the misleading "pure function / stateless" documentation on `PeerSelector` and `SelectSyncPeer`, in `docs/topics/features/peer_registry_reputation.md` and `docs/topics/services/p2p.md`, and updated the `p2p_health_check_enabled` longdesc (real endpoint `/bestblockheader`, concurrency, budget, cache).

### Residual
The selection decision path still holds `decisionMu` across the probe pre-pass: at most one 5s probe round per 10s cache window. In the pathological case where every probe times out, with `fastMonitorInterval` at 2s that means `decisionMu` can be unavailable for roughly a third to half of wall-clock time while syncing - still far better than up to 2N x 2s per tick, but it is why moving probes outside the lock is worth doing as a follow-up. That refactor is left out of scope here.

### Tests
`services/p2p/peer_selector_health_test.go` (injectable `checkHealth` stub plus one test against real `httptest` HTTP servers): unhealthy-peer filtering, exactly one probe per URL across both selection passes, cache reuse within TTL and re-probe after expiry, unhealthy results served from cache, expired-entry pruning, health checking disabled (zero probes), shared-URL dedup, genuine concurrency with the pool bound asserted (24 peers, max in-flight in (1, 8]), full node not starved by slow pruned peers ordered ahead of it, cache entries stamped at each probe's own completion (fast vs slow probe in one round), concurrent `SelectSyncPeer` rounds exercising `healthMu` and the cache write-back under the race detector, immediate abort on cancelled caller context, and budget expiry (probe outliving a shortened budget resolves unhealthy and is not cached). `sync_coordinator_test.go` covers the `warnfUnlessStopping` downgrade after `Stop`. Existing selector tests updated for the new signature.

```
go test -race -tags "testtxmetacache" -count=1 ./services/p2p/... ./settings/...
ok  github.com/bsv-blockchain/teranode/services/p2p  17.6s
ok  github.com/bsv-blockchain/teranode/settings      2.9s
```
Health-check tests also pass `-race -cpu=1 -count=20` (no flakes).


ctnguyen (changes requested):
Review at commit [45c0f8f](https://github.com/gokutheengineer/teranode/commit/45c0f8f8c4f4adf95054ceb80a705a56332ca151)

Solid, well-scoped fix for a real problem: serial 2s probes under `context.Background()` on the sync coordinator loop were a genuine starvation vector, and bounding them with concurrency plus an overall budget, deduping via a short-TTL cache, and threading a cancellable context so `Stop` aborts in-flight probes is the right shape. Test coverage pulls its weight too — the concurrency bound, budget-expiry-not-cached, and cache reuse/expiry cases are all there. One thing needs attention before merge: the probe pre-pass doesn't preserve the full-node-first preference under the shared budget.

## Blocking issues

### ChiR1 — Health pre-pass probes all tiers under one shared budget, so slow pruned peers can starve a healthy full node's probe
**Problem:** `checkCandidatesHealth` collects the DataHub URLs of every basic-eligible peer — full and pruned alike, since `isEligibleBasic` doesn't distinguish tier — in peer-slice order, then drains them FIFO through the 8-worker pool under a single 5s `probeCtx`. The two-phase full-node preference is applied only afterward, in `getFullNodeCandidates`/`getPrunedNodeCandidates`, which merely read the precomputed `healthByURL`. So if enough slow, basic-eligible, lower-priority (pruned/header-only) URLs sit ahead of a healthy full node's URL in the slice, they occupy the workers and drain the budget first; the full node's URL is dequeued late — or after `probeCtx` has already expired, in which case `checkPeerAvailability` returns immediately on the cancelled ctx without ever contacting it — and resolves "unhealthy"/"not probed" for that tick. Phase 1 then finds no healthy full node and falls through to the pruned fallback (or returns nothing) even though a healthy full node was reachable. With ~8 workers at the 2s per-probe cap, roughly 20+ slow preceding URLs push the full node past the 5s budget. The pre-PR code probed serially with no overall budget, so a healthy full node was always eventually probed (just slowly) and never *dropped*.
**Why it matters:** This is exactly the load the PR targets — "a handful of slow/unreachable-but-advertised peers." Under that load the change can now demote a healthy full node in favor of a pruned peer (or select no peer) for a selection tick, undermining the two-phase full-node preference the selector is built around. That is a correctness regression on the core path, not just a latency one.
**Fix:** Give first-tier URLs priority for the shared workers and budget. The minimal change preserves the single pre-pass but enqueues full-node-tier URLs ahead of pruned-tier ones; better still, probe the pruned tier only when no healthy full-node candidate emerged (Phase 2 already runs only on Phase 1 miss). Add a regression test: several slow pruned peers ordered before one fast healthy full peer, asserting the full peer is still selected.

```go
// Enqueue full-node-tier URLs ahead of pruned-tier URLs so the shared
// worker pool and budget serve the preferred tier first.
var fullURLs, prunedURLs []string
now := time.Now()
for _, p := range peers {
    if _, dup := seen[p.DataHubURL]; dup {
        continue
    }
    if !ps.isEligibleBasic(p, criteria) {
        continue
    }
    seen[p.DataHubURL] = struct{}{}
    if ps.isEffectiveFullNode(p, now) {
        fullURLs = append(fullURLs, p.DataHubURL)
    } else {
        prunedURLs = append(prunedURLs, p.DataHubURL)
    }
}
urls := append(fullURLs, prunedURLs...)
```

## Non-blocking issues

### ChiR2 — Basic eligibility is recomputed up to three times per tick, multiplying debug-log noise
**Problem:** `checkCandidatesHealth` runs `isEligibleBasic` over every peer, then `getFullNodeCandidates` (`isEligibleFullNode` → `isEligible` → `isEligibleBasic`) and `getPrunedNodeCandidates` (`isEligible` → `isEligibleBasic`) recompute it, so each peer runs the banned/URL/reputation/progress/cooldown checks up to three times per selection and each failed criterion emits its `Debugf` line each pass.
**Why it matters:** Minor wasted CPU on the coordinator loop and, with debug logging on, tripled per-peer eligibility log volume that muddies the selection trace. Not a correctness issue.
**Fix:** Compute basic eligibility once per tick and reuse it — the pre-pass already visits every peer, so have it hand the candidate passes the basic-eligible set (keyed by peer ID) instead of re-deriving it, or gate the pre-pass debug lines.

### ChiR3 — `Stop` cancels `sc.ctx` before `wg.Wait`, surfacing benign context-canceled warnings from in-flight registry calls
**Problem:** `Stop` now does `close(sc.stopCh); sc.ctxCancel(); sc.wg.Wait()`. The `monitorFSM`/`periodicEvaluation` goroutines share `sc.ctx` for their registry RPCs (`ListPeers`, `GetPeer`, `RecordSyncAttempt`, …); a call in flight at shutdown returns `context.Canceled`, logged via `Warnf` (e.g. `sync_coordinator.go:234`, `:243`).
**Why it matters:** Every shutdown can now emit a few expected "…failed: context canceled" warnings. Cosmetic, but warn-level noise on a normal shutdown erodes signal and can trip alerting that greps for `Warn`.
**Fix:** Suppress the log when the coordinator context is already cancelled:

```go
if err != nil {
    if sc.ctx.Err() != nil {
        sc.logger.Debugf("[SyncCoordinator] ListPeers aborted on shutdown: %v", err)
        return nil
    }
    sc.logger.Warnf("[SyncCoordinator] ListPeers failed: %v", err)
    return nil
}
```

## Recap

| ChiR1 | Full-node starved by shared probe budget | YES | 70% |
| ChiR2 | Basic eligibility recomputed 3× | | 20% |
| ChiR3 | Shutdown context-canceled warns | | 12% |

ctnguyen (approved):
Review at commit [a6967fe](https://github.com/gokutheengineer/teranode/commit/a6967fe8dc52ceba8f4d537b386fb37d8cdd964c)

Nice follow-up — all three points from the last pass are genuinely addressed, not papered over. The full-tier-first enqueue onto the FIFO `work` channel does give the preferred tier the workers and budget first (and the new `..._FullNodeNotStarvedBySlowPrunedPeers` test pins it with a shortened budget, which is the right way to test it), `basicEligibleByID` cleanly collapses the triple eligibility recomputation into one snapshot per tick, and `warnfUnlessStopping` is a tidy one-liner that kills the shutdown warn noise without hiding real failures. The concurrency itself reads correctly: each worker writes a distinct `probeResults[i]`, `wg.Wait()` gives the happens-before, and every `healthCache` touch is under `healthMu`. What's left is small stuff — one narrow survivor of the tier-priority fix, a piece of new machinery that nothing in production actually reaches, a TTL that runs a bit longer than advertised, and a test/doc gap.

## Status of previous points

| ChiR1 | Full-node starved by shared probe budget | YES | FIXED |
| ChiR2 | Basic eligibility recomputed 3× | | FIXED |
| ChiR3 | Shutdown context-canceled warns | | FIXED |

ChiR1 is fixed for the general case; ChiR4 below covers a narrow shared-URL residual of the same failure mode.

## Non-blocking issues

### ChiR4 - A `DataHubURL` shared across tiers is classified by the first peer seen, so a full peer can still land in the pruned probe queue
**Problem:** In `checkCandidatesHealth` (`services/p2p/peer_selector.go:344-364`) the `seen` set dedupes a URL the moment the first basic-eligible peer using it is visited, and that peer alone decides the tier. If a pruned (or `FullStoragePenaltyUntil`-penalised) peer holds `DataHubURL` U and a *full* peer later in the slice holds the same U, U is appended to `prunedURLs`; the full peer is then skipped as a duplicate and reads U's pruned-tier probe result from `healthByURL`. Put slow pruned peers ahead of that pair and U is dequeued late or after `probeCtx` expires, so the healthy full node resolves unhealthy and Phase 1 misses it — ChiR1's exact failure mode, surviving behind one extra precondition. The precondition is reachable in practice: a node that restarts with a fresh libp2p ID re-registers the same DataHub URL while the registry still holds the old entry (with its old `Storage` value), and multi-instance deployments can front several peers with one asset server. The code comment at `:341-343` acknowledges the behaviour but treats it as harmless, which it isn't once the budget is what decides the tick.
**Why it matters:** It reopens the demotion this PR set out to close — a healthy full node loses to a pruned fallback, or no peer is selected, for that tick. The blast radius is much smaller than the original ChiR1 (needs a cross-tier URL collision plus a saturated budget) and it self-corrects on the next tick once the slow probes get cached, which is why it isn't a merge blocker.
**Fix:** Decide the tier per URL over all its users, promoting to the full tier if *any* basic-eligible peer on that URL is an effective full node, then keep the existing first-seen ordering:

```go
seen := make(map[string]struct{}, len(peers))
tierIsFull := make(map[string]bool, len(peers))
order := make([]string, 0, len(peers))
now := time.Now()

for _, p := range peers {
    if _, ok := basicEligible[p.ID]; !ok {
        continue
    }
    if _, dup := seen[p.DataHubURL]; !dup {
        seen[p.DataHubURL] = struct{}{}
        order = append(order, p.DataHubURL)
    }
    if ps.isEffectiveFullNode(p, now) {
        tierIsFull[p.DataHubURL] = true // any full-tier user promotes the URL
    }
}

var fullURLs, prunedURLs []string
for _, url := range order {
    if tierIsFull[url] {
        fullURLs = append(fullURLs, url)
    } else {
        prunedURLs = append(prunedURLs, url)
    }
}
urls := append(fullURLs, prunedURLs...)
```

Extend `..._FullNodeNotStarvedBySlowPrunedPeers` with a pruned peer that shares the healthy full peer's `DataHubURL` and is ordered ahead of it, and update the `:341-343` comment.

### ChiR5 - `SyncCoordinator.Stop` is never called in production, so the new cancel-on-shutdown path is inert
**Problem:** `NewSyncCoordinator` now wraps the caller's ctx in a cancellable child and `Stop` cancels it (`services/p2p/sync_coordinator.go:83`, `:398-402`), but nothing in production ever calls it: `syncCoordinator` appears in `Server.go` only at `:141`, `:433`, `:444`, `:691-692` and `:1571-1572`, and `Server.Stop` (`services/p2p/Server.go:1725`) shuts down the libp2p client, Kafka clients, Echo and the tickers without touching it. The only `sc.Stop()` call in the tree is `sync_coordinator_test.go:1080`. In-flight probes do still abort at real shutdown, but via the parent (`NewServer`'s ctx) rather than via the new code — and `sc.wg.Wait()` is never reached, so the monitor goroutines are still not joined.
**Why it matters:** The PR's stated shutdown benefit ("`Stop` now cancels that context, so in-flight probes abort on shutdown instead of stalling `Stop`") is only exercised by tests. It isn't a regression, but the machinery is dead weight and the next reader will assume the join actually happens.
**Fix:** Stop and join the coordinator at the *top* of `Server.Stop`, before its dependencies go away — `P2PClient.Close()` runs at `:1741` and `kafka.StopProducerCtx(..., s.blocksKafkaProducerClient)` at `:1764`, while the coordinator goroutines still call the registry, the blockchain client and that same blocks producer (`sendSyncTriggerToKafka`). Stopping it late would let a `handleFSMTransition`/`evaluateSyncPeer` tick publish into an already-stopping producer:

```go
func (s *Server) Stop(ctx context.Context) error {
    s.logger.Infof("[Stop] Stopping P2P service")

    // Quiesce the sync coordinator first: its goroutines use the registry,
    // the blockchain client and the blocks producer closed further down.
    if s.syncCoordinator != nil {
        s.syncCoordinator.Stop()
        s.logger.Infof("[Stop] stopped sync coordinator")
    }

    var errs []error
    // ... existing teardown
```

Guard `Stop` against a second call while you're there (`close(sc.stopCh)` panics on re-entry), e.g. with a `sync.Once`.

### ChiR6 - Cache entries are stamped after `wg.Wait`, inflating the effective TTL by up to the whole probe budget
**Problem:** `checkedAt := time.Now()` is taken once, after every worker has finished (`services/p2p/peer_selector.go:423-431`), and that single timestamp is written to every cached entry. A probe that resolved at t=0.05s is therefore stamped with the time the *slowest* probe finished — up to `healthCheckBudget` (5s) later. On a 10s TTL that means a result can be reused for as long as ~15s, and it is worst exactly when it matters least: a peer set with slow members, where the fast answers are the ones you'd want re-validated soonest.
**Why it matters:** A peer that went down right after a fast healthy probe stays "healthy" for 50% longer than the documented 10s window, so selection can keep handing sync to a dead DataHub for an extra tick or two. Small, but it silently breaks the TTL invariant the godoc and the `p2p_health_check_enabled` longdesc both state.
**Fix:** Stamp each probe with its own completion time:

```go
type probeResult struct {
    healthy   bool
    cacheable bool
    checkedAt time.Time
}
// ...in the worker, right after the probe returns:
probeResults[i] = probeResult{
    healthy:   healthy,
    cacheable: healthy || probeCtx.Err() == nil,
    checkedAt: time.Now(),
}
// ...and when writing the cache:
ps.healthCache[pending[i]] = peerHealthCacheEntry{healthy: r.healthy, checkedAt: r.checkedAt}
```

Keep the single `time.Now()` for the prune sweep.

### ChiR7 - Negative caching — the one genuinely new user-visible behaviour — has no test
**Problem:** The change makes a *completed* unhealthy probe sticky for `peerHealthCacheTTL` (`cacheable: healthy || probeCtx.Err() == nil`, `peer_selector.go:418`), which is new: before this PR every pass re-probed. The new tests cover the healthy-cached case (`..._CachedAcrossCalls`), TTL expiry (`..._CacheExpires`) and the *not*-cached case (`..._BudgetExpiryNotCached`), but nothing asserts that a peer which genuinely answered non-200 is remembered as down and skipped without a second probe inside the TTL.
**Why it matters:** This is the branch that can keep a recovered peer excluded for up to 10s (15s, per ChiR6). If someone later flips the `cacheable` expression, or decides not to cache negatives at all, the suite stays green either way.
**Fix:** Add the mirror of `..._CachedAcrossCalls`:

```go
func TestPeerSelector_HealthChecks_UnhealthyResultCached(t *testing.T) {
    ps := newHealthCheckSelectorForTest()
    bad := newPeer("peer-bad", 100, "full", 50, 0)
    good := newPeer("peer-good", 100, "full", 50, 0)
    peers := []*blockchain.PeerInfo{bad, good}

    stub := newCountingHealthStub(bad.DataHubURL)
    ps.checkHealth = stub.check

    require.Equal(t, "peer-good", ps.SelectSyncPeer(context.Background(), peers, advertisedProbeCriteria(50)))
    require.Equal(t, "peer-good", ps.SelectSyncPeer(context.Background(), peers, advertisedProbeCriteria(50)))
    require.Equal(t, 1, stub.callsFor(bad.DataHubURL), "a completed unhealthy probe must be cached, not repeated")
}
```

### ChiR8 - Stale "stateless" wording left in two places, one of them the doc page this PR edits
**Problem:** The PR retracts the misleading stateless/pure-function wording in the `PeerSelector` godoc and at `docs/topics/features/peer_registry_reputation.md:101`, but two spots still contradict it: `docs/topics/features/peer_registry_reputation.md:36` (`- **Peer Selector**: A stateless component that selects optimal peers based on reputation and other criteria`, in the same file, 65 lines above the line that *was* corrected) and `services/p2p/Server.go:140` (`peerSelector *PeerSelector // Stateless peer selection logic`).
**Why it matters:** The component-overview bullet at `:36` is what most readers of that page actually skim, so the published doc still asserts the exact claim the linked issue was filed about; `Server.go:140` is the first line anyone hits when tracing who owns the selector.
**Fix:** Bring both in line with the corrected `:101` wording — e.g. at `peer_registry_reputation.md:36`, "A component that selects optimal peers based on reputation and other criteria, optionally probing peer availability over HTTP (with a short-lived result cache)", and at `Server.go:140`, "// Peer selection logic; probes and caches DataHub availability when p2p_health_check_enabled".

## Recap

| ChiR4 | Shared URL misclassified as pruned tier | | 30% |
| ChiR5 | Coordinator Stop never invoked | | 25% |
| ChiR6 | Cache timestamp inflates TTL | | 18% |
| ChiR7 | Negative caching untested | | 14% |
| ChiR8 | Stale "stateless" wording, two spots | | 10% |

freemans13 (commented):
Reviewed the branch at `27e805edb` against `f0ffb4be1`. No correctness bug found — the core idea is right, and it does fix what the issue describes. Replacing the serial per-candidate probe inside `isEligible` with one deduplicated, bounded-concurrency, budgeted pre-pass whose results are cached is the correct shape, and the implementation holds up. Nothing below blocks the merge.

### What I ran

| Check | Result |
|---|---|
| `go vet ./services/p2p/... ./settings/...` | clean |
| `golangci-lint run services/p2p/... settings/...` | 1 hit — `prealloc` in `services/p2p/mock.go`, pre-existing on main, not in this diff |
| `go test -race -tags testtxmetacache -count=1 ./services/p2p/... ./settings/...` | pass |
| health tests, `-count=20 -cpu=1 -race` | pass, no flakes |

The single-CPU stress run is the one I cared about: `..._RunConcurrently` asserts `maxInFlight > 1` and `..._FullNodeNotStarvedBySlowPrunedPeers` depends on the work channel draining FIFO. Both survive 20 runs at `GOMAXPROCS=1`, so they shouldn't start failing on a loaded CI box.

### Worth fixing

**1. The cache timestamp is taken too late, inflating the TTL** — `services/p2p/peer_selector.go:508`

```go
wg.Wait()

checkedAt := time.Now()   // one stamp for the whole round
```

Every cacheable result inherits the moment the *slowest* probe finished, not its own. In a round that burns the full 5s budget, a probe that completed after 50ms is recorded as if it completed 5s later, so it stays reusable for nearly 15s rather than the intended 10. Net effect: a peer that went down can be remembered healthy for half again as long as the TTL suggests. Easy fix — capture the completion time inside the worker and carry it in `probeResult` alongside `healthy` and `cacheable`.

**2. The "not probed" branch is now unreachable** — `peer_selector.go:389-395`

This was live in an earlier commit but isn't any more: `basicEligibleByID` freezes eligibility *before* the pre-pass, and the pre-pass enqueues the DataHub URL of every basic-eligible peer, so by the time `isEligible` reaches the lookup the URL is guaranteed present. The code is harmless and honestly commented as defensive. The issue is that the PR description still presents it as a real behaviour:

> A peer that becomes basic-eligible only after the pre-pass (e.g. cooldown expired mid-probe) is skipped for that round with a distinct "not probed" log line

That can no longer happen. Worth rewording to say what it now is — a guard against a future refactor that unfreezes eligibility.

**3. The description overclaims the `Stop` change**

> and `SyncCoordinator.Stop` now cancels that context

`Stop` already calls `sc.ctxCancel()` on `main` — `git show upstream/main:services/p2p/sync_coordinator.go` confirms it. The branch's actual contribution is the *threading*: `sc.ctx` now reaches `SelectSyncPeer`, so probes ride a cancel that already existed for RPCs. Same end behaviour, but as written it sends the next reader hunting the diff for a `Stop` change that isn't there.

**4. Undeclared scope: `warnfUnlessStopping`** — `sync_coordinator.go:278-287` plus 13 call sites

Commit `a6967fe8d` adds a helper that downgrades `Warnf` to `Debugf` once `sc.ctx` is cancelled, and applies it across thirteen log statements unrelated to health checking (`ListPeers`, `GetPeer`, `GetBestBlockHeader`, `RecordSyncAttempt` ×3, `RemovePeer`, `UpdatePeerMetrics`, `RegisterPeer`, `IsPeerBanned`, `ClearAllSyncAttempts`, `ReconsiderBadPeers`).

The behaviour is benign — `sc.ctx.Err()` is only non-nil after `Stop`, so normal operation is byte-identical, and `sc.ctx` is never nil since `NewSyncCoordinator` sets it. But it's ~35 lines of the diff, it has nothing to do with concurrent health checks, and the description doesn't mention it. It landed under "address review nits", so it was probably suggested here — which makes it more important to name explicitly, not less.

**5. The residual is characterised optimistically**

> that stall is now bounded at ~5s total instead of growing linearly with peer count

True per selection, but the duty cycle is the part that bites. `fastMonitorInterval` is 2s while actively syncing and the cache TTL is 10s, so in the pathological case where every probe times out you get one 5s probe round per 10s cache window — `decisionMu` is unavailable for roughly a third to a half of wall-clock time. Still a large improvement on up to 2N × 2s, and I'd ship it. But "at most one 5s probe round per 10s cache window" is the honest phrasing, and it makes clearer why moving probes outside the lock is worth doing later.

### Minor

**Shared DataHub URL defeats the tier ordering** — `peer_selector.go:431-446`. A URL is classified into the full or pruned queue by whichever peer holding it appears first in the slice, so a pruned peer listed ahead of a full peer sharing that URL puts it in the pruned tier. Already commented honestly; two peers advertising one DataHub URL across different storage tiers is exotic. Noting for completeness.

**`append` aliasing** — `peer_selector.go:447`, `urls := append(fullURLs, prunedURLs...)` writes into `fullURLs`' backing array. Correct today since nothing reads `fullURLs` afterwards, but it's the classic `appendAssign` shape. A fresh `make([]string, 0, len(fullURLs)+len(prunedURLs))` costs nothing and removes the question.

### Test gaps

The new test file is genuinely good — filtering, single-probe-per-URL across both passes, cache reuse and expiry, real concurrency with the pool bound asserted, caller-context abort, budget expiry, plus one end-to-end case against real `httptest` servers. Gaps I'd still want, none blocking:

| Gap | Why it matters |
|---|---|
| Negative results are cached | `CachedAcrossCalls` only proves a *healthy* result is reused. The unhealthy-and-cached path (`cacheable = healthy \|\| probeCtx.Err() == nil`) is the one that saves work on a down peer, and nothing asserts it. |
| Cache pruning | Lines 517-522 delete expired entries; no test exercises the delete, so a regression letting the map grow unboundedly would pass CI. |
| Health checking disabled | Nothing asserts `HealthCheckEnabled: false` yields a nil map and zero probes — the setting an operator reaches for under trouble. |
| URL deduplication | Two peers sharing one DataHub URL should produce one probe; asserted only indirectly via the two-pass test. |
| `warnfUnlessStopping` | Untested — no case covers the downgrade after `Stop`. |

### Summary

The two I'd actually want addressed before this lands are the timestamp inflation (a real, if small, behaviour bug) and the description corrections — the `Stop` claim in particular. Everything else is polish or follow-up.

services/p2p/peer_selector.go
```diff
@@ -300,27 +368,177 @@ func (ps *PeerSelector) isEligible(p *blockchain.PeerInfo, criteria SelectionCri
...
+
+	wg.Wait()
+
+	pruneCutoff := time.Now()
+
+	ps.healthMu.Lock()
+	for i, r := range probeResults {
+		results[pending[i]] = r.healthy
+		if r.cacheable {
+			ps.healthCache[pending[i]] = peerHealthCacheEntry{healthy: r.healthy, checkedAt: r.checkedAt}
```
freemans13, line 536: The write-back stores this round's result unconditionally, so a concurrent selection whose probe finished *later* can have its entry overwritten by an older one.

This is benign today — every caller reaches `SelectSyncPeer` under the coordinator's `decisionMu`, so there is only ever one round in flight. And the godoc already warns that concurrent calls may double-probe. But it doesn't mention that a concurrent round can also clobber a newer entry with a staler one, which is the more surprising half.

Two ways to close it, and I'd lean to the second:

```go
if existing, ok := ps.healthCache[pending[i]]; !ok || r.checkedAt.After(existing.checkedAt) {
    ps.healthCache[pending[i]] = peerHealthCacheEntry{healthy: r.healthy, checkedAt: r.checkedAt}
}
```

or just extend the godoc sentence to say so. The guard defends against a caller that doesn't exist yet, so documenting the constraint is arguably the more honest fix — but if the intent is that the cache stays correct under future concurrent callers, the guard is cheap.

services/p2p/peer_selector.go
```diff
@@ -300,27 +368,177 @@ func (ps *PeerSelector) isEligible(p *blockchain.PeerInfo, criteria SelectionCri
...
+
+		pending = append(pending, url)
+	}
+	ps.healthMu.Unlock()
+
+	if len(pending) == 0 {
+		return results
+	}
+
+	probeCtx, cancel := context.WithTimeout(ctx, ps.healthCheckBudget)
```
freemans13: A zero `healthCheckBudget` would make this context expire immediately, which marks every peer unhealthy and quietly selects nothing — a hard-to-diagnose failure, since nothing errors, selection just always returns `""`.

Unreachable today, because `NewPeerSelector` is the only thing that builds the struct and it always sets the constant. I mention it only because this file already takes exactly that precaution one function over, in `randomIndex`'s guard on `randIntN`, so there's a local precedent:

```go
budget := ps.healthCheckBudget
if budget <= 0 {
    budget = peerHealthCheckBudget
}
probeCtx, cancel := context.WithTimeout(ctx, budget)
```

Entirely your call whether the consistency is worth two lines of diff.

services/p2p/peer_selector.go
```diff
@@ -14,6 +15,26 @@ import (
 	"github.com/bsv-blockchain/teranode/util/health"
 )
 
+const (
+	// peerHealthCheckConcurrency bounds how many peer availability probes run at once.
+	peerHealthCheckConcurrency = 8
+
+	// peerHealthCheckBudget caps the total wall-clock time one selection spends on
+	// availability probes; probes still pending when it expires resolve as unhealthy.
+	peerHealthCheckBudget = 5 * time.Second
```
freemans13, line 24: Not asking for a change in this PR — the constants are defensible as compile-time values while this is new. But of the three, the budget is the one that bites operationally, and I think it wants to become a `P2PSettings` field before this meets a large or flaky peer set.

The reason is the interaction you already flag in the description: this 5s is spent while holding `SyncCoordinator.decisionMu`. With the default periodic evaluation interval at 30s, a tick where every probe times out blocks every sync decision for a sixth of the evaluation window, and there is no way to dial that down in the field without a rebuild. An operator on a bad network has no lever.

Smaller point in the meantime: the `p2p_health_check_enabled` longdesc now mentions the 5s budget but not that it is held under the decision lock. That consequence is the part an operator actually needs to know.

services/p2p/peer_selector.go
```diff
@@ -80,8 +122,16 @@ func (ps *PeerSelector) SelectSyncPeer(peers []*blockchain.PeerInfo, criteria Se
 		return ""
 	}
 
+	// Basic (network-free) eligibility is computed once per selection and
+	// shared by the availability pre-pass and both candidate passes below.
+	basicEligible := ps.basicEligibleByID(peers, criteria)
+
+	// Probe availability of every basic-eligible candidate once, up front, so
+	// the eligibility checks below do no network I/O.
+	healthByURL := ps.checkCandidatesHealth(ctx, peers, basicEligible)
```
freemans13, line 131: Follow-up thought, explicitly not a request for this PR.

The pre-pass probes pruned-tier URLs too, even when Phase 1 finds a healthy full node and the fallback never runs. That's not a regression — `main` did the same thing, because `getFullNodeCandidates` called `isEligible` and hence `checkPeerAvailability` for every peer — so this PR is strictly better either way.

But the pre-pass structure you've built here makes a genuinely two-stage probe cheap: probe the full tier, run Phase 1, and only probe the pruned tier if Phase 1 comes back empty. On a healthy network that roughly halves steady-state probe volume, and it shortens the window `decisionMu` is held in the common case, which is the same pressure point as the budget comment above. Might be worth folding into the same follow-up that moves probing outside the lock.

services/p2p/peer_selector_health_test.go
```diff
@@ -0,0 +1,358 @@
...
+
+	ps.SelectSyncPeer(context.Background(), []*blockchain.PeerInfo{peerB}, advertisedProbeCriteria(50))
+
+	ps.healthMu.Lock()
+	_, stillCached := ps.healthCache[peerA.DataHubURL]
+	ps.healthMu.Unlock()
+	require.False(t, stillCached, "expired entry for a churned peer URL must be pruned")
+}
+
+func TestPeerSelector_HealthChecks_CacheExpires(t *testing.T) {
```
freemans13, line 260: Two gaps worth closing, both cheap given the harness you've already built here.

**The per-probe `checkedAt` stamping has no direct test.** This test ages one entry by hand, so it would still pass if entries were stamped once per round with the slowest probe's finish time — which is exactly the bug the stamping change fixed. A test with one fast and one slow probe in the same round, asserting the fast URL's entry expires on its own schedule rather than the round's, would pin the actual behaviour.

**Nothing drives two concurrent `SelectSyncPeer` calls on one selector.** That's the only thing that would genuinely exercise `healthMu` under `-race`, and it's the exact case the `SelectSyncPeer` godoc calls out as permitted. Right now the concurrency tests exercise concurrency *within* a round, not *across* rounds.

Separately, and fine as-is: the "not probed" branch in `isEligible` is uncovered because it's unreachable — `isEligible` only reaches that lookup after the basic-eligibility check passes, and every basic-eligible peer's URL is guaranteed to be in `results`. I'd keep the branch (it's fail-closed and costs one map lookup), but it's worth a line in the PR description saying the lack of coverage is deliberate, so the next person doesn't try to "fix" it.

services/p2p/sync_coordinator.go
```diff
@@ -273,6 +275,18 @@ func (sc *SyncCoordinator) isUnprovenProbeCandidate(p *blockchain.PeerInfo, now
 	return !sc.peerHasRecentFullBlockDelivery(p, now)
 }
 
+// warnfUnlessStopping logs at Warn level, downgraded to Debug once the
+// coordinator context is cancelled: Stop cancels sc.ctx before waiting for the
+// worker goroutines, so registry/blockchain calls in flight at shutdown fail
+// with an expected context.Canceled that should not surface as a warning.
+func (sc *SyncCoordinator) warnfUnlessStopping(format string, args ...any) {
```
freemans13, line 282: Flagging rather than objecting: this is separable from the health-check work.

The `sc.ctx` cancel-on-`Stop` that this reacts to is pre-existing on `main` — the description says as much about the cancellation — so the shutdown warn-spam it suppresses isn't caused by this PR's context threading. It would be just as correct and just as desirable as its own two-file PR. It's small, it's right, and it's tested, so I wouldn't pull it out at this point; but a reviewer coming to this cold will wonder why 13 call sites changed in a PR about probe concurrency. Either split it or say plainly in the description that it's a deliberate drive-by.

One tiny thing while you're here: this dereferences `sc.ctx` with no nil check, and `server_helpers_test.go:163` constructs a bare `&SyncCoordinator{}` literal, so a future test path reaching this would panic. Low priority — `boundedRPCContext` has the identical pre-existing exposure, so this is at least consistent with the file rather than a new hazard.

freemans13 (commented):
Reviewed locally at `79c3321`. Short version: the concurrency work holds up. I went looking for a correctness bug in the worker pool, the budget context, the cache, and the tier ordering and didn't find one, and the package stays green under `-race` at 20 repeats. Everything inline below is a judgement call or a follow-up rather than a defect — the one thing I'd call a factual correction is the docs claim, at the bottom.

### Checks run locally

| Check | Result |
|---|---|
| `make lint` (golangci-lint, `--new-from-rev`) | PASS — 0 issues |
| `go test -race -tags testtxmetacache ./services/p2p/... ./settings/...` | PASS — p2p 14.0s, settings 1.8s |
| `go vet ./services/p2p/... ./settings/...` | PASS |
| `-race -run 'Health\|Selector' -count=20` | PASS — no flakes |

### Claims in the description, checked rather than taken on trust

| Claim | Verdict |
|---|---|
| Probes bounded at 8 concurrent | Verified — `min(concurrency, len(pending))` workers, each running one probe at a time; the test asserts in-flight max in (1, 8] with 24 peers |
| Budget capped at 5s regardless of peer count | Verified — `WithTimeout` with `defer cancel()` on the next line, no leak; the probe is ctx-aware, so queued entries fail immediately once it fires rather than each waiting out a 2s client timeout |
| Full-node tier probed before pruned, deterministically | Verified, and it is a real guarantee rather than luck: URLs are concatenated full-then-pruned, indices are pushed into a buffered channel in that order, and buffered-channel receives are FIFO |
| Cache entries stamped with their own probe completion time | Verified — `time.Now()` is evaluated inside the worker right after the probe returns, and the comparisons carry monotonic readings, so a wall-clock jump can't corrupt the TTL |
| ctx/budget-cut probes not cached | Verified — `cacheable: healthy \|\| probeCtx.Err() == nil` is sound in both directions, and errs conservative (a genuine unhealthy verdict landing just before expiry is also dropped, which is the safe direction) |
| Expired entries pruned, cache bounded | Verified — swept under `healthMu` at the end of any round that had probes; all-cache-hit rounds skip the sweep but add no entries |
| Exactly one probe per unique URL per selection | Verified — `seen` dedupes and fresh-cached URLs never enter `pending` |
| `warnfUnlessStopping` byte-identical in normal operation | Verified — same level, same format string, same args, no extra allocation before cancellation |
| Declared residual: `decisionMu` held across the pre-pass | Confirmed accurate — every path into `SelectSyncPeer` runs under that lock, so a selection can hold it for the full budget |

One nuance on the budget: workers still *drain* the remaining queue indices after expiry, so there's an O(N) tail of fast-failing `http.Client` constructions. Microseconds, and not worth changing — just not literally zero work.

### The docs claim is not quite complete

The rename to `isEligibleBasic` and the removal of the "stateless" wording each left a line behind. None of these are in this PR's diff, so GitHub won't let me attach them inline:

- `services/p2p/Server.go:153` — `peerSelector *PeerSelector // Stateless peer selection logic`. This is the field comment for the very type whose godoc this PR corrects.
- `docs/references/services/p2p_reference.md:45` — the same `// Stateless peer selection logic` line, mirrored in the reference doc.
- `services/p2p/Server.go:1082` — the blacklist-at-point-of-use comment still sends readers to `PeerSelector.isEligible`; that filter now lives in `isEligibleBasic`.
- `services/p2p/sync_coordinator.go:167` — same stale `isEligible` reference in the `isViableSyncCandidate` godoc (just outside the diff hunk).

All four are one-line comment/docs changes.

### Two pre-existing things this PR now amplifies

Neither is yours to fix here, but worth knowing. `util/health/http_health.go` builds a fresh `http.Client` (and therefore a fresh `Transport`) per probe and closes the response body without draining it, so keep-alive never engages and every probe is a new TCP+TLS handshake — unchanged by this PR, but now fanned out 8-wide. And `docs/p2p-libp2p-review.md:83` points at `PeerSelector.isEligible()` at a line number that was already stale on `main`.

---

*Review assisted by Claude Code. Findings were verified against the code at `79c3321`; the local test runs above are real, but the usual caveat applies that 20 race-enabled repeats is evidence against ordinary interleavings, not a proof.*

services/p2p/peer_selector_health_test.go
```diff
@@ -0,0 +1,426 @@
...
+	ps.healthMu.Unlock()
+
+	require.True(t, fastCached)
+	require.True(t, slowCached)
+	gap := slowEntry.checkedAt.Sub(fastEntry.checkedAt)
+	require.GreaterOrEqual(t, gap, slowDelay/2,
+		"the fast probe's entry must be stamped at its own completion, not at the end of the round")
+}
+
+func TestPeerSelector_HealthChecks_ConcurrentSelectionsAreSafe(t *testing.T) {
```
freemans13, line 225: This is a real test, but not of the thing the commit message credits it with — it doesn't cover the cache write-back guard.

I checked by mutation rather than by eye. Reverting the guard to the old unconditional store — the actual clobber bug — and running this `-count=10`: **passes every time.**

That's structural rather than bad luck. Every stub probe here returns `healthy=true`, so a stale cache entry is indistinguishable from a fresh one, and the only assertion (`emptySelections == 0`) can't observe which entry won. There is no state the clobber can perturb that this test looks at.

It *is* a genuine test of something else, and worth keeping on those grounds: stripping the `healthMu` lock around the write-back and prune makes it fail immediately under `-race`, 3 runs out of 3. So it's solid lock-coverage.

To make it cover the guard, the probe stub needs to return *different* health per round — e.g. round 1 unhealthy and slow, round 2 healthy and fast — and then assert the cache ends up holding the later verdict. If you'd rather not, that's fine, but I'd retitle it (`...SelectionsAreRaceFree`, or similar) so that a year from now nobody reads the name and assumes the guard is regression-tested when it isn't.

For contrast, `CacheEntriesStampedPerProbe` just above is excellent and I verified it the same way: stamping every entry with a single round-end timestamp makes it fail with exactly the message it was written to give (`"0s" is not greater than or equal to "125ms"`), and the 250ms-against-125ms margin didn't flake over 20 repeats.

settings/p2p_settings.go
```diff
@@ -93,5 +93,5 @@ type P2PSettings struct {
...yncCoordinatorPeriodicEvaluationInterval time.Duration `key:"p2p_sync_coordinator_periodic_evaluation_interval" desc:"Periodic evaluation interval for sync coordinator" default:"30s" category:"P2P" usage:"How often to evaluate sync status" type:"duration" longdesc:"### Purpose\nControls how often SyncCoordinator evaluates current sync peer performance and health.\n\n### How It Works\nHealth checks performed each interval:\n- Peer still connected\n- Reputation above threshold (20)\n- No validated block progress within the no-progress timeout = failure\n- Caught up to peer (switch to better peer if available)\n\nIf all eligible peers attempted: enters exponential backoff (2s, 4s, 8s...up to 64s), then clears sync attempt timestamps to allow retry.\n\n### Trade-offs\n| Setting | Benefit | Drawback |\n|---------|---------|----------|\n| Shorter (10s) | Quicker detection of peer issues | Higher CPU overhead |\n| Longer (60s) | Lower overhead | Slower response to peer problems |\n\n### Recommendations\n- **30s** (default) - Balances responsiveness and overhead\n- **10s** - Fast networks or testing requiring quick peer switching\n- **60s** - Stable networks to reduce evaluation overhead"`
```
freemans13, line 93: The new line is right that the probe round holds the decision lock — I traced every production path and they all do. It's the second half I'd soften: the 5-second budget doesn't quite bound the stall.

One `decisionMu` hold can chain two probe rounds. `evaluateSyncPeer` calls `selectNewSyncPeer()`, and on success goes on to `triggerSyncLocked()` → `selectAndActivateNewPeer` → a second `SelectSyncPeer`, all under the same acquisition.

Normally the second round is a pure cache hit, since the 10s TTL comfortably outlives the 5s budget, so this costs nothing. But it fails to save you in precisely the bad case: if the first round was budget-cut, those results are deliberately *not* cached (correctly so), which means the second round re-probes from scratch and pays the budget again. Worst case is nearer 10s than 5s — exactly when an operator is already having a bad day.

No code change needed. Either say "bounded by the 5-second budget per probe round", or leave the longdesc and note the chaining in the `checkCandidatesHealth` godoc, which is where someone debugging a stall would actually look.

services/p2p/peer_selector.go
```diff
@@ -300,27 +368,189 @@ func (ps *PeerSelector) isEligible(p *blockchain.PeerInfo, criteria SelectionCri
...
+	close(work)
+
+	var wg sync.WaitGroup
+
+	for range min(peerHealthCheckConcurrency, len(pending)) {
+		wg.Go(func() {
+			for i := range work {
+				url := pending[i]
+
+				healthy, err := ps.checkHealth(probeCtx, url)
```
freemans13, line 500: Small one, take it or leave it. Once the budget fires with URLs still queued, the workers keep draining the channel and calling `checkHealth` for every remaining index. Each of those builds a request on a dead context and fails instantly — correct, but it's churn proportional to the queue length, and it emits a Debugf per URL, so a bad tick on a large peer set produces a burst of log lines that all say the same thing.

Cheap to skip, and the semantics already line up:

```go
for i := range work {
    if probeCtx.Err() != nil {
        probeResults[i] = probeResult{}
        continue
    }
    url := pending[i]
    ...
}
```

The zero value is `healthy:false, cacheable:false`, which is exactly what a budget-cut probe should resolve to, so nothing downstream changes.

---

While I'm here, on the "should the budget be a setting" question from the first review — I've come round to your position, but for a sharper reason than "not yet". The pre-pass probes every basic-eligible peer's URL, so with N eligible peers you issue N requests per tick, bounded at 8 concurrent × 2s; the 5s budget covers roughly 20 URLs. Past that the tail is *always* cut, always resolves unhealthy, and because cut results are deliberately not cached, the same tail is re-attempted and re-starved on the next tick, with no progress ever made. A larger budget doesn't fix that — it just moves the cliff. The fix is scheduling: rotate the pending order between rounds, or remember which URLs got starved and deprioritise them next time. Definitely a follow-up, not this PR, but worth capturing somewhere before this meets a big peer set.

freemans13 (approved):
Second pass, at `9f09a8f`. Approving — `34da758` addresses all three code points from the first review, and I checked the code rather than the commit message. Everything below is either evidence for that, or small stuff you can take or leave.

The one thing I'd genuinely like you to look at before merge is the new concurrency test, which doesn't cover what its commit message says it covers. Details inline.

### Checks at this head

| Check | Result |
|---|---|
| `make lint` | PASS — 0 issues |
| `go vet ./services/p2p/... ./settings/...` | PASS |
| `go test -race -tags testtxmetacache -count=1 ./services/p2p/... ./settings/...` | PASS — p2p 14.4s, settings 2.0s |
| `-race -run 'Health\|Selector' -count=20` | PASS — no flakes |

### The response commit, audited

**Cache write-back guard — fixed, and no regression.** The guard keeps whichever probe completed later, and it holds under either lock-acquisition order: whoever writes second only overwrites when strictly newer, so the freshest entry survives regardless of who wins the mutex.

I want to call out one thing you got right, because it's the kind of thing that usually goes wrong. Inverting `if r.cacheable {…}` to `if !r.cacheable { continue }` had the potential to skip the line above it, and if it had, budget-cut probes would have dropped out of `results` entirely rather than resolving as `false` — a silent behaviour change in selection, not a crash. `results[pending[i]] = r.healthy` still runs first, so it's fine.

**Zero budget — fixed.** `<= 0` rather than `== 0` also catches a negative override, which `WithTimeout` would otherwise turn into an instantly-dead context. I grepped the package: no unguarded read of `ps.healthCheckBudget` remains.

**Longdesc — accurate on the lock claim.** I traced every production path to `SelectSyncPeer` and they all do hold `decisionMu`. One caveat on the second half of that sentence, inline on `p2p_settings.go`.

### Still open from the first review

Four stale comment lines, all still present at this head. Two are in files this PR doesn't touch, so I can't attach them inline:

- `services/p2p/Server.go:153` — `// Stateless peer selection logic`, the field comment for the very type whose godoc this PR corrects
- `docs/references/services/p2p_reference.md:45` — the same line mirrored in the reference doc
- `services/p2p/Server.go:1082` — blacklist comment still points at `PeerSelector.isEligible`, now `isEligibleBasic`
- `services/p2p/sync_coordinator.go:167` — same stale `isEligible` in the `isViableSyncCandidate` godoc

The last one is in a file already in your diff, so it's close to free. None of these block anything.

Also still open, and both fine to leave: `warnfUnlessStopping` remains separable from the health-check work (flagging, not objecting — the test that came with it is well-targeted), and I now agree with your judgement on *not* promoting the budget to a setting, for a reason I've put inline on `peer_selector.go`.

### Invariants re-verified at this head

At most 8 concurrent probes; a 5s budget per round; full-node URLs enqueued ahead of pruned ones; exactly one probe per unique URL per selection; ctx- and budget-cut probes not cached; expired entries pruned so the cache stays bounded. The concurrency core reads clean — each `probeResults[i]` is written by exactly one worker, `wg.Wait()` happens-before every read, the buffered-and-closed channel guarantees each index is consumed once, workers can't write after the collector runs, and `defer cancel()` fires only after `wg.Wait()`, so budget expiry leaks no goroutines.

---

*Review assisted by Claude Code. The mutation tests described inline were actually run; the broader "no races" conclusion rests on code reading plus 20 race-enabled repeats, which is evidence against ordinary interleavings rather than a proof.*

PR #1348, opened 2026-07-27
p2p: Implement p2p Client.GetPeers and surface peer decode errors

gokutheengineer:
**Closes bitcoin-sv/teranode#4721.**

### Problem
`Client.GetPeers` made the gRPC call, discarded the response, and unconditionally returned an empty slice, silently violating the `ClientI` contract - callers paid for the RPC and could not distinguish "no peers" from "not implemented". Additionally, `convertFromAPIPeerInfo` swallowed `peer.Decode` and `chainhash.NewHashFromStr` errors (`peerID, _ := ...`) and returned a silent empty `PeerInfo` for unknown types, so corrupt data surfaced as zero peer IDs / nil hashes instead of errors. Both items were confirmed still present.

### Fix
- `Client.GetPeers` now converts the `GetPeersResponse` into native `[]*PeerInfo` (new `*p2p_api.Peer` case in the converter: ID, height, ban score, bytes received) and marks results `IsConnected`, since the RPC only ever reports connected peers.
- `convertFromAPIPeerInfo` returns `(*PeerInfo, error)`; decode failures are wrapped in a `ProcessingError` via small `decodePeerID` / `parseOptionalBlockHash` helpers, and the unknown-type default returns an error instead of an empty struct.
- Error handling is deliberately split by call shape: list endpoints (`GetPeers`, `GetPeersForCatchup`, `GetPeerRegistry`) log a warning and skip the bad entry so one corrupt registry record cannot blank out catchup peer selection or the asset `/peers` endpoint, while `GetPeer` (single-peer lookup) propagates the error.
- Beyond the proposed fix, `Server.GetPeers` only populated `Id`/`Addr`/`Banscore`, which would have left client-side `Height` permanently zero and kept `cmd/diagnose`'s peer-height check inert; the registry path now also fills `CurrentHeight` and `BytesReceived`. While touching that block, the libp2p address lookup was hoisted out of the per-peer loop (it was O(peers x libp2p peers) per call).
- Stale `ClientI.GetPeers` doc (peer "addresses" are not part of `PeerInfo`) corrected.

Side effect worth noting: RPC `getinfo`'s `connections` count previously always added 0 for teranode p2p peers; it now includes them. Rolling upgrade is graceful in both directions: an old server simply reports `CurrentHeight`/`BytesReceived` as 0 (the diagnose height check guards on `> 0`), and an old client ignores the new fields.

### Residual
- `cmd/monitor` renders `ReputationScore`, which the legacy `p2p_api.Peer` message cannot carry, so monitor shows `rep:0` for `GetPeers` results; carrying reputation would need a proto change (or pointing monitor at `GetPeerRegistry`), left out of scope.
- Pre-existing direct field access on gRPC responses elsewhere in `Client.go` (no nil guards) was left untouched to keep the diff minimal; the new code uses nil-safe `resp.GetPeers()`.

### Tests
- `TestSimpleClientGetPeers`: real conversion assertions (ID, height, ban score, bytes received, `IsConnected`), gRPC error propagation, and invalid-entry skip behavior.
- `TestConvertFromAPIPeerInfoErrors`: invalid peer ID and invalid block hash for both `PeerInfoForCatchup` and `PeerRegistryInfo`, plus the unsupported-type case.
- `TestServerGetPeersReturnsConnectedPeersWithHeight`: server-level test (local registry) asserting connected-only filtering and that `CurrentHeight` reaches the response.

```
ok  github.com/bsv-blockchain/teranode/services/p2p  14.006s
```

### DIFF
--- a/services/p2p/Client.go
+++ b/services/p2p/Client.go
@@ -99,15 +99,27 @@ func (c *Client) Close() error {
 //   - []*PeerInfo: Slice of peer information
 //   - error: Any error encountered during the operation
 func (c *Client) GetPeers(ctx context.Context) ([]*PeerInfo, error) {
-	_, err := c.client.GetPeers(ctx, &emptypb.Empty{})
+	resp, err := c.client.GetPeers(ctx, &emptypb.Empty{})
 	if err != nil {
 		return nil, err
 	}
 
-	// Convert p2p_api response to native PeerInfo slice
-	// Note: The p2p_api.GetPeersResponse contains legacy SVNode format
-	// For now, return empty slice as the actual implementation uses GetPeerRegistry
-	return []*PeerInfo{}, nil
+	peers := make([]*PeerInfo, 0, len(resp.GetPeers()))
+
+	for _, apiPeer := range resp.GetPeers() {
+		peerInfo, err := convertFromAPIPeerInfo(apiPeer)
+		if err != nil {
+			c.logger.Warnf("skipping peer with invalid data: %v", err)
+			continue
+		}
+
+		// The GetPeers RPC only reports currently connected peers.
+		peerInfo.IsConnected = true
+
+		peers = append(peers, peerInfo)
+	}
+
+	return peers, nil
 }
 
 // BanPeer implements the ClientI interface method to ban a peer.
@@ -523,7 +535,12 @@ func (c *Client) GetPeersForCatchup(ctx context.Context) ([]*PeerInfo, error) {
 	// Convert p2p_api peer info to native PeerInfo
 	peers := make([]*PeerInfo, 0, len(resp.Peers))
 	for _, apiPeer := range resp.Peers {
-		peerInfo := convertFromAPIPeerInfo(apiPeer)
+		peerInfo, err := convertFromAPIPeerInfo(apiPeer)
+		if err != nil {
+			c.logger.Warnf("skipping catchup peer with invalid data: %v", err)
+			continue
+		}
+
 		peers = append(peers, peerInfo)
 	}
 
@@ -717,7 +734,13 @@ func (c *Client) GetPeerRegistry(ctx context.Context) ([]*PeerInfo, error) {
 	// Convert p2p_api peer registry info to native PeerInfo
 	peers := make([]*PeerInfo, 0, len(resp.Peers))
 	for _, apiPeer := range resp.Peers {
-		peers = append(peers, convertFromAPIPeerInfo(apiPeer))
+		peerInfo, err := convertFromAPIPeerInfo(apiPeer)
+		if err != nil {
+			c.logger.Warnf("skipping registry peer with invalid data: %v", err)
+			continue
+		}
+
+		peers = append(peers, peerInfo)
 	}
 
 	return peers, nil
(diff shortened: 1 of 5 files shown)


ctnguyen (approved):
Review at commit [022bf36](https://github.com/gokutheengineer/teranode/commit/022bf3678e59c5e1ccc3eb4ee9f712ead7b4e6ec)

Nice, focused fix. `Client.GetPeers` was a stub that made the RPC and threw the result away, and the converter was quietly eating decode errors — both are now handled properly, and the diff stays tight around exactly those two problems. I like the split you landed on: list endpoints skip-and-log a single bad record while `GetPeer` propagates, which is the right asymmetry (one corrupt registry entry shouldn't blank out catchup selection). The `decodePeerID`/`parseOptionalBlockHash` helpers keep the three converter cases readable, the server-side libp2p lookup hoist is a genuine O(peers × libp2p) → O(peers) win, and the test coverage is solid — real conversion assertions, gRPC error propagation, skip behavior, and a server-level connected-only check. The error-wrapping convention (trailing `err`, no `%w`) is also correct against the recent CI guard. No blocking concerns; just two small notes below.

## Non-blocking issues

### ChiR1 — Malformed block hash now drops the whole peer from catchup selection
**Problem:** In the `PeerInfoForCatchup` / `PeerRegistryInfo` cases, a non-empty but unparseable `BlockHash` now returns an error, so `GetPeersForCatchup`/`GetPeerRegistry` skip that peer entirely. Previously the peer was kept with `BlockHash == nil`. In `blockvalidation/peer_selection.go` a nil `BlockHash` is carried into `PeerForCatchup` without a guard, so such a peer was previously still a selectable catchup source (it passes the height and DataHub-URL filters on its own merits).
**Why it matters:** A single corrupt tip-hash string on an otherwise healthy, high-DataHub peer now removes it from the catchup pool rather than degrading gracefully to "no known tip". That is a slightly harsher failure mode than the peer-ID case (a bad peer ID is genuinely unusable; a bad tip hash is not). It's defensible and it's logged, so this is a judgment call rather than a defect — worth a conscious decision.
**Fix:** If you want to preserve the old graceful degradation for the tip hash specifically, treat an unparseable hash like an empty one (nil + warn) instead of failing the peer:
```go
blockHash, err := chainhash.NewHashFromStr(hash)
if err != nil {
    // corrupt tip hash: keep the peer, just drop the hash
    return nil, nil // or log and continue at the call site
}
```
A bad peer ID should still hard-skip. If you'd rather keep the current strictness, that's fine — just confirm it's intentional.

### ChiR2 — Server test doesn't assert the new `BytesReceived` field
**Problem:** `TestServerGetPeersReturnsConnectedPeersWithHeight` asserts `CurrentHeight` reaches the response but not `BytesReceived`, even though both fields were added to the registry path in the same block.
**Why it matters:** `BytesReceived` is new wiring in `Server.GetPeers` and a silent regression (e.g. a future edit dropping the field) wouldn't be caught here. The client-side test covers it, but the server-side population is untested.
**Fix:** Seed a non-zero byte count on the connected peer and assert it, e.g.:
```go
require.Equal(t, uint64(2048), resp.Peers[0].BytesReceived)
```

## Recap
| ChiR1 | Bad tip hash drops catchup peer | | 25% |
| ChiR2 | BytesReceived untested server-side | | 10% |

icellan (approved):
Approving. Correct and well-scoped. Checked every caller of `ClientI.GetPeers` (`cmd/monitor:297`, `cmd/diagnose/health_checks.go:233,658,756`, `services/rpc/handlers.go:1666`, `daemon/test_daemon.go:2239`) and `GetPeer` (`blockvalidation/catchup_get_block_headers.go:503`, `blockvalidation/Server.go:1238`) — nothing breaks on a now-non-empty result and there's no nil-deref path on the fields the legacy `p2p_api.Peer` can't carry. The Server-side address-map hoist is behaviourally equivalent to the old inner loop. `go build` clean, new tests pass.

Three non-blocking notes:

- `services/p2p/Client.go:117` — `IsConnected = true` is right for this endpoint, but `Server.go:1763` lists with `excludeBanned=false` and `p2p_api.Peer` has no `isBanned` field, so a banned-but-still-connected peer is now counted in RPC `getinfo`.connections and shown as a healthy peer. The client has no way to filter it. Worth passing `excludeBanned=true` on the `GetPeers` path.
- `services/p2p/Client.go:112` (same at `:540`, `:739`) vs `:790` — list endpoints skip a bad registry entry, single lookup errors on it. A corrupt row silently disappears from asset `/peers`, `getpeerinfo`, and the asset peer-auth tier cache (`services/asset/httpimpl/peer_auth_middleware.go:116`), where the dropped peer falls to `tierUnverified` and gets the strictest rate limit — while `GetPeer` on that same row returns an error. Defensible for availability, but it should increment a metric so it isn't invisible. Note `cmd/diagnose` polls `GetPeers` repeatedly, so a persistently corrupt row will warn on every poll.
- `cmd/monitor/monitor.go:934` — now that `GetPeers` returns real peers, all of them render `rep:0` in red `errorStyle` because `ReputationScore` can't cross the legacy message. You call this out in the description; either point monitor at `GetPeerRegistry` or render unknown reputation as `-` rather than 0.

PR #1349, opened 2026-07-27
fix(p2p): randomize sync-peer tiebreak to resist peer-ID grinding

gokutheengineer:
**Closes bitcoin-sv/teranode#4723.**

### Problem

`selectFromCandidates` in `services/p2p/peer_selector.go` sorted candidates with a fully deterministic comparator whose final tiebreak was `candidates[i].ID < candidates[j].ID`, then always picked index 0. libp2p peer IDs are cheap to grind, so a Sybil attacker who matches the higher-priority criteria (reputation starts high for new peers; advertised height is peer-controlled) could mint an ID that sorts first and deterministically win sync-peer selection, biasing the node toward attacker-controlled catchup sources.

The issue was confirmed still valid on current `main`. Since it was filed the comparator gained two stronger leading criteria (proven recent full-block delivery and locally validated chain work), which already implement the issue's "weight by proven successful-interaction history" suggestion, but the grindable ID tiebreak at the bottom remained.

### Fix

- Extracted the sort comparator into a three-way `comparePeerCandidates` (proven delivery, validated chain work, reputation, response time, ban score, validated height) and removed the peer-ID tiebreak entirely.
- Candidates that tie on every merit criterion form the top band, and the winner is drawn uniformly at random from that band. Randomness comes from `math/rand/v2.IntN` (per-thread ChaCha8 seeded from OS entropy, not observable or predictable by a remote attacker, so crypto/rand is unnecessary); the source is injectable on `PeerSelector` for deterministic tests, with a nil-guard fallback so a future struct-literal construction cannot panic.
- The previous peer is excluded from selection whenever an alternative candidate exists, so a tied previous peer cannot be re-drawn from the top band. Production callers already strip the previous peer in `filterEligiblePeersWithTip`, so this is defence-in-depth at the selector layer, not a behavior change.
- Selection log now includes `topBandSize` so operators can tell "randomised among N equals" from "single clear winner".
- Updated stale docs that specified the removed ID tiebreak as design: `docs/topics/features/peer_registry_reputation.md` (selection criteria list was also stale in other ways) and the diagram source `docs/topics/features/img/peer_selection_sequence.puml`.

### Residual

- `docs/topics/features/img/peer_selection_sequence.svg` still shows the old "Peer ID" tiebreak; the `.puml` source is updated but plantuml is not available locally to regenerate the SVG.
- Randomisation engages only on exact ties across all six criteria. With N merit-tied Sybils against H merit-tied honest peers, capture probability per selection drops from 100% to N/(N+H) - proportional, not eliminated. The stronger bound remains the unproven-probe budget and the proven-delivery/validated-work criteria an attacker must genuinely satisfy to reach the top band.

### Tests

Added:

- `TestPeerSelector_SelectFromCandidates_RandomTiebreakCoversWholeTopBand` - stubs the random source; asserts the top band contains exactly the merit-tied peers, every band member is reachable (distinct results across indices), and a strictly worse peer never is.
- `TestPeerSelector_SelectSyncPeer_GrindableIDCannotCaptureSelection` - among 4 tied peers including an attacker ID that sorts lexicographically first, every peer must win at least once over 200 rounds (flake probability ~1e-24); catches any fixed-index selection.
- `TestPeerSelector_ComparePeerCandidates_Antisymmetric` - property check that the hand-written three-way comparator satisfies compare(a,b) == -compare(b,a) and compare(a,a) == 0, keeping `sort.Slice` and the band scan sound.
- `TestPeerSelector_SelectSyncPeer_PreviousPeerExcludedFromTiedTopBand` and `TestPeerSelector_SelectSyncPeer_PreviousPeerKeptWhenOnlyCandidate` - previous-peer exclusion semantics.
- `TestSyncCoordinator_SelectNewSyncPeer_MeritTiedSybilCannotCapture` - end-to-end through the real coordinator selection path (`selectNewSyncPeer` with registry, eligibility filtering, and criteria construction): a merit-tied Sybil ID that sorts first must not capture selection; every tied peer wins at least once over 100 rounds.

Ran (green):

```
go vet ./services/p2p/
```

### DIFF
--- a/docs/topics/features/img/peer_selection_sequence.puml
+++ b/docs/topics/features/img/peer_selection_sequence.puml
@@ -31,8 +31,8 @@ end note
 PS -> PS: Filter for "full" storage mode
 
 alt Full nodes available
-  PS -> PS: Sort by:\n1. Reputation (desc)\n2. Ban score (asc)\n3. Height (desc)\n4. Peer ID
-  PS -> PS: Select top candidate
+  PS -> PS: Rank by:\n1. Proven full-block delivery\n2. Validated chain work (desc)\n3. Reputation (desc)\n4. Response time (asc)\n5. Ban score (asc)\n6. Validated height (desc)
+  PS -> PS: Random pick among\ntop-ranked ties\n(no peer-ID tiebreak)
   PS --> BV: Selected full node peer
 else No full nodes available
 
@@ -40,13 +40,8 @@ else No full nodes available
 
   alt Fallback enabled
     PS -> PS: Filter non-full peers
-    PS -> PS: Sort by:\n1. Reputation (desc)\n2. Ban score (asc)\n3. Height (asc)\n4. Peer ID
-    note right
-      Prefer youngest pruned
-      nodes to minimize
-      UTXO pruning risk
-    end note
-    PS -> PS: Select top candidate
+    PS -> PS: Rank with same criteria\nas full nodes
+    PS -> PS: Random pick among\ntop-ranked ties\n(no peer-ID tiebreak)
     PS --> BV: Selected pruned node peer
   else Fallback disabled
     PS --> BV: No peer available
--- a/docs/topics/features/peer_registry_reputation.md
+++ b/docs/topics/features/peer_registry_reputation.md
@@ -218,27 +218,30 @@ The peer selector uses a two-phase approach for optimal selection:
 #### Phase 1: Full Node Selection
 
 1. Filter for peers that explicitly announce as "full" storage mode
-2. Sort candidates by:
+2. Rank candidates by:
 
-   - Reputation score (highest first) - **primary**
-   - Ban score (lowest first) - **secondary**
-   - Block height (highest first) - **tertiary**
-   - Peer ID (for deterministic ordering) - **quaternary**
-3. Select the top candidate (or second if top was previous peer)
+   - Proven recent full-block delivery - **primary**
+   - Locally validated chain work (highest first) - **secondary**
+   - Reputation score (highest first) - **tertiary**
+   - Average response time (lowest first, peers with measurements first) - **quaternary**
+   - Ban score (lowest first) - **quinary**
+   - Validated block height (highest first) - **senary**
+3. Exclude the previously selected peer when any alternative exists
+4. Select uniformly at random among the top-ranked candidates that tie on
+   every criterion above
+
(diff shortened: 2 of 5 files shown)


ctnguyen (approved):
Review at commit [6235d3b](https://github.com/gokutheengineer/teranode/commit/6235d3bb709bb1e7ec15a3887b1be00167aa352f)

Nice, tightly-scoped fix that removes the last grindable hook — the peer-ID tiebreak — and replaces it with a uniform draw across the merit-tied top band, backed by a faithful comparator refactor and genuinely thorough tests (full-band coverage, an end-to-end grind-can't-capture path, and an antisymmetry property check). The previous-peer exclusion is true defence-in-depth and the residuals are honestly disclosed.

liam (commented):
Reviewed the comparator soundness, band computation, randomization, and previous-peer exclusion. **Looks correct - no blocking issues.**

- **Comparator is transitive by construction.** `comparePeerCandidates` is a lexicographic composition of six single-element key extractions, each a total order (proven-delivery bool, `CompareChainWork`, reputation, response time, ban score, validated height). Because each key is a function of one element only, the composition is antisymmetric *and* transitive, so `sort.Slice` and the band scan are well-defined. The NaN trap doesn't apply here: `ReputationScore` is clamped to [0,100] on every path in `updateReputationScore`, and `AvgResponseTimeMs` is `int64` (can't be NaN).
- **Band computation is correct** - `peer_selector.go` compares each successor against `candidates[0]` (not adjacent pairs), so the contiguous equal-prefix after sorting is captured exactly; no merit-equal peer dropped, no worse peer included.
- **Grinding resistance is sound** - uniform draw over the top band removes positional bias, and `math/rand/v2` (ChaCha8, OS-seeded, unobservable) is adequate; the `crypto/rand`-unnecessary argument holds. The residual `N/(N+H)` proportional capture is honestly disclosed.
- **Previous-peer exclusion** uses the safe `[:0]` filter idiom and only reassigns when `0 < len(filtered) < len(candidates)`, so the set is never emptied and `randomIndex` is never called with n<=0.

**Non-blocking nit:** the tests cover antisymmetry but not transitivity. Transitivity holds by construction today, but a 3-peer property check (`compare(a,b)<=0 && compare(b,c)<=0 => compare(a,c)<=0`) would harden the band logic against future criterion additions.

liam (approved):

PR #1350, opened 2026-07-27
batch peer-registry updates and bound gossip handler concurrency in p2p

gokutheengineer:
**Closes bitcoin-sv/teranode#4722.**

### Problem

All issue items confirmed still valid on main:

- `subscribeToTopic` ran one goroutine per topic invoking the handler synchronously per message, so all gossip for a topic serialized behind the handler's blocking RPCs (head-of-line blocking).
- Each gossip message issued ~6-9 synchronous peer-registry gRPC calls: `updatePeerLastMessageTime` (RegisterPeer + UpdateConnectionState + UpdateLastMessageTime for the sender, RegisterPeer + UpdateLastMessageTime for the originator), `updateBytesReceived` (UpdatePeerMetrics x2), `addPeer` (RegisterPeer), and `shouldSkipBannedPeer` (IsPeerBanned) - a cheap gossip flood amplified into ~8x registry load, and `addConnectedPeer` re-registered already-connected peers on every message.

### Fix

- **New `peerRegistryBatcher`** (`services/p2p/peer_registry_batcher.go`): gossip handlers enqueue per-peer updates under a mutex and return immediately; a background goroutine flushes at most one small batch of RPCs per peer per interval (`p2p_peer_registry_batch_interval`, default 1s). Registration data merges to the most meaningful values (heights are max-merged with their paired block hash, so concurrent workers cannot regress an advertised tip within a flush window), byte deltas accumulate, and RegisterPeer/UpdateConnectionState are re-asserted at most once per 60s unless new registration data arrives. Backpressure: the pending map is bounded at 10k peers; beyond that, updates for new peers are dropped and counted. `removePeer` leaves a tombstone in the batcher so a peer removed mid-flush is neither resurrected nor skipped for re-registration on its next message (the tombstone map is count-bounded like the other two, and a flush-scoped removal set keeps a forget → re-enqueue interleaving inside one cycle from flushing the stale pre-removal snapshot). Any failed flush RPC requeues its intent for the next flush - the whole coalesced batch when RegisterPeer fails, or just the failed intents (byte deltas, last-message touch, connection state, storage) when individual follow-up RPCs error against a degraded registry.
- **Bounded worker pool per topic** in `subscribeToTopic` (`p2p_gossip_handler_concurrency`, default 4), replacing the single blocking goroutine, with a per-message `recover` so a crafted message that drives a handler into a panic is logged and dropped instead of crashing the node.
- **`IsPeerBanned` TTL cache** (5s, same pattern and lifecycle as the existing `reputationCache`/`ipBanCache`), with immediate overwrite on local ban transitions (`onPeerBanned`), invalidation on `ResetReputation`, expiry sweeps in `cleanupPeerMaps`, and an error breaker: a failed lookup is cached fail-open for the TTL, so a degraded registry is hit (and logged) at most once per peer per 5s instead of once per message.
- **`getLocalHeight` 1s cache** (improvement over the proposed fix's scope): `sanitizeAdvertisedTip` called `GetBestBlockHeader` on the blockchain service per gossip message - the same head-of-line/amplification class as the registry calls, just against a different service. Failures are cached with a shorter 200ms TTL: long enough to shed the per-message RPC storm during an outage, short enough that recovery (and correct advertised-tip capping) resumes quickly.
- Servers constructed without a batcher (unit tests) fall back to the previous direct RPC path; production always gets the batcher via `NewServer`, which clamps a non-positive configured interval to the default so the batcher's synchronous test mode is unreachable in production.

### Review round 1 (ChiR1-4 + bot note)

- **ChiR1 (blocking)**: `registryBatcher.stop(ctx)` now takes the `Stop` context, so the whole shutdown path - waiting out an in-flight flush and the final flush itself - is bounded by the service manager's per-service stop budget, mirroring `kafka.StopProducerCtx`. If the budget is exhausted while a flush is wedged, stop logs and returns instead of overrunning; the final flush is capped at `min(stop budget, registryFlushTimeout)`. Periodic flushes still detach from parent-ctx cancellation (bounded by `registryFlushTimeout`) so a flush in progress when shutdown starts completes.
- **ChiR2**: error breaker on `IsPeerBanned` (see above), flood-tested against a failing registry client (100 messages = 1 RPC).
- **ChiR3**: per-message panic recovery in the topic workers (see above).
- **ChiR4**: shorter TTL (200ms) for failed local-height reads so recovery is picked up quickly; the cap-to-maxLead side effect while an error entry is served is documented at the constant.
- **Bot note**: `lastAsserted` is now count-bounded at the same 10k cap as the pending map (stale-sweep on insert when full; if still full the assertion is simply not recorded and the peer re-registers next flush), and removal tombstones are pruned after 2x `registryFlushTimeout` instead of 10 minutes - a spoofed-ID flood can no longer grow either map without bound.

### Review round 2 (concurrency review)

- **Requeue on any RPC failure**: previously only a failed RegisterPeer requeued. Now each failed follow-up RPC (UpdateConnectionState, UpdateLastMessageTime, UpdatePeerMetrics, UpdateStorage) collects its intent into a partial update that is requeued for the next flush, so byte deltas / freshness / storage are no longer silently dropped when individual unary RPCs time out against a degraded registry. Tested with a fail-once UpdatePeerMetrics client.
- **Tombstone map count-bounded**: `forget()` now applies the same swept-then-cap discipline as `lastAsserted` - at the 10k cap, expired tombstones are swept and, if still full, the tombstone is skipped (re-registration stays guaranteed via the `lastAsserted` deletion). Tested at the cap with fresh and expired tombstones.
- **Tombstone-clear interleaving closed**: a flush-scoped removal set (not cleared by re-enqueue, reset at end of each cycle) now backs the persistent tombstones, so forget → re-enqueue inside one flush can no longer let the loop push the peer's stale pre-removal snapshot; the fresh data flushes next cycle. Tested with a blocking-hook interleaving that pins the exact sequence.

### Residual

- Registry-side eviction (TTL/LRU in the centralized registry, currently deferred to a follow-up per the in-code comments) could remove a peer the batcher still considers asserted; the peer would be absent from the registry for up to the 60s re-assert TTL. Active peers keep their `LastMessageTime` fresh every flush, so TTL eviction cannot hit them; only future LRU-under-pressure eviction could, and it self-heals within 60s.
- A ban originating outside this node's ban-transition path (e.g. another service writing ban score directly to the registry), or a lookup failing against a degraded registry, is masked fail-open by the cached `IsPeerBanned` entry for up to 5s - the same staleness contract as the pre-existing reputation cache.
- A peer removed mid-flush can still be transiently re-created in the registry by the in-flight RegisterPeer (the RPCs cannot be un-sent); the batcher state stays correct and the entry is corrected by the peer's next message or normal registry lifecycle.
- Topic workers are not drained before the batcher's final flush; handlers still running during shutdown may enqueue updates that never flush. The loss bound is one in-flight message per worker - up to `GossipHandlerConcurrency` x topics messages' worth of stats at process exit.
- The height merge is monotonic only within a flush window: two near-simultaneous messages split across a flush boundary can reach the registry out of order, leaving a transiently regressed advertised height until the peer's next message (the old strictly-ordered path never regressed from reordering). This is deliberate - a hard monotonic clamp would also suppress legitimate decreases (a peer reorging to a shorter chain, or restarting from a lower height), which must keep propagating; advertised height only feeds sync-peer selection and self-heals on the next gossip message.

### Tests

Added `services/p2p/peer_registry_batcher_test.go`:

- end-to-end amplification guard: 100-message block-gossip flood from one peer = 1 IsPeerBanned on the hot path and exactly 1 RegisterPeer + 1 UpdateConnectionState + 1 UpdateLastMessageTime + 1 UpdatePeerMetrics at flush, with the latest height surviving coalescing
- coalescing, latest-data merge, monotonic height merge, re-assert TTL skip, new-info forces register, forget/re-register, backpressure drop at the cap, stop-flushes-pending, synchronous mode, stop-without-start (no deadlock), final flush after parent context cancellation, requeue on RegisterPeer failure, forget racing an in-flight flush (blocking-hook interleaving), worker pool bound + drain (deterministic barrier, timeout fails the test)
- review round: stop honors an exhausted stop budget while a flush is wedged, assert-state map count bound (fresh entries reject, stale entries swept), handler panic recovered with subsequent messages still processed, IsPeerBanned error breaker (100 failing lookups = 1 RPC), getLocalHeight error entries expire on the shorter TTL

Updated `server_helpers_test.go` for the ban-status cache (staleness contract + immediate local-ban effect via `applyBanScore`).

Commands run (green, affected packages only; branch rebased on latest main with the `startInvalidBlockConsumer` removal from the invalid-blocks consumer rewiring taken as-is):

```
SETTINGS_CONTEXT=test go test -race -count=1 ./settings/...
go vet ./services/p2p/... ./settings/...
```

Concurrency-sensitive tests additionally stressed with `-count=8 -race`.


ctnguyen (changes requested):
Review at commit [ff00edc](https://github.com/gokutheengineer/teranode/commit/ff00edccf3950ed5d727534f472ba0c969ed2309)

Nice piece of work — the diagnosis of the amplification/head-of-line problem is spot on, and the batcher is carefully built: monotonic height merge, a removal tombstone with a post-RPC re-check, requeue-on-failure so accumulated metrics aren't dropped, and a final flush that survives the cancelled parent context. Test coverage is unusually thorough, the `forget`-during-flush interleaving especially. The one thing to fix before merge is the batcher's shutdown path ignoring the service stop budget; the rest are smaller degraded-registry and hardening notes.

## Blocking issues

### ChiR1: Batcher shutdown ignores the service-manager stop budget
**Problem:** `registryBatcher.stop()` takes no context. It waits for any in-flight flush and then runs a final `flushOnce`, each bounded only by the batcher's own 30s `registryFlushTimeout` (via `context.WithoutCancel`, which deliberately detaches from all cancellation). But the service manager calls every `Stop(stopCtx)` under a `ServiceManagerStopTimeout` deadline (default 30s), and the setting is explicit that a service's Stop must fit inside it — "Must be long enough for Kafka flushes and batcher drains" and "each Stop() runs under a context with this timeout." The Kafka producers one screen up in `Server.Stop` honor this via `kafka.StopProducerCtx(ctx, ...)`; the batcher can't — it never sees the deadline. Under a wedged registry the p2p `Stop` can run ~30s (or ~60s if a flush is already mid-cycle), overrunning the per-service budget and tripping the daemon's outer 30s shutdown watchdog (`daemon.go`), which then abandons the in-progress shutdown. Separately, the topic worker `WaitGroup` is local to `subscribeToTopic`, so `Stop` can't drain in-flight handlers before the final flush; one still running can enqueue after it.
**Why it matters:** this breaks the documented per-service stop contract. A shutdown during a registry outage overruns the budget, and once the outer watchdog fires the node is force-stopped mid-flush — cutting off exactly the graceful drain this code is trying to perform, and eating into the shutdown window of services still queued behind p2p.
**Fix:** give `stop` the `Stop` ctx and bound the final flush by it, using `registryFlushTimeout` only as a ceiling, mirroring `kafka.StopProducerCtx`. Optionally hold the topic `WaitGroup`(s) on the `Server` and wait on them (also ctx-bounded) before the final flush.
```go
// Server.Stop
if s.registryBatcher != nil {
    s.registryBatcher.stop(ctx) // honor the ServiceManagerStopTimeout budget
}
// in stop(ctx): derive the final-flush context from the caller's deadline,
// capped at registryFlushTimeout, instead of context.WithoutCancel(b.ctx).
```

## Non-blocking issues

### ChiR2: Ban check re-hits the registry (and re-logs) per message when the lookup errors
**Problem:** `shouldSkipBannedPeer` writes `banStatusCache` only after a *successful* `IsPeerBanned`. On a registry error it logs `Warnf` and returns without caching, so every subsequent gossip message re-issues the synchronous RPC *and* re-logs. Under a flood while the registry is degraded, that's one blocking RPC and one log line per message — on the path this PR set out to relieve. It mirrors the pre-existing `reputationCache` error path (`shouldSkipUnhealthyPeer` also skips caching on error) and isn't a regression, but the sibling `getLocalHeight` cache was deliberately hardened against exactly this and the ban path wasn't.
**Why it matters:** a degraded registry under gossip load becomes a per-message log flood and sustained RPC pressure on a component that is already failing.
**Fix:** put a short, bounded error breaker on the lookup — on an `IsPeerBanned` error, record the failure with a brief TTL and, until it expires, skip both the RPC and the log for that peer. Fail open during the window (treat as not-banned), which briefly delays enforcing a *new* ban under a registry outage — the same staleness contract the reputation/ban caches already accept. If you'd rather not touch enforcement timing at all, narrow this to rate-limiting the failure log alone (aggregate per flush interval, like the batcher already does for dropped updates). Either way, a flood test against a failing registry client locks the bound in.

### ChiR3: A panicking gossip handler crashes the whole node
**Problem:** each worker runs `handler(ctx, msg.Data, msg.FromID)` in a bare `for msg := range topicChannel` loop with no `recover`. An unrecovered panic in a handler does not merely kill that worker — in Go it takes down the entire process. (The single-goroutine version had the same exposure; converting to a pool is the natural moment to harden it.) Gossip handlers parse untrusted network input, so a single crafted message reaching a panicking path crashes the node.
**Why it matters:** a crash-on-message is a cheap remote DoS if any handler path can be driven to panic.
**Fix:** wrap the call in a per-message recover so a bad message is logged and dropped instead of crashing:
```go
func() {
    defer func() {
        if r := recover(); r != nil {
            s.logger.Errorf("[%s] gossip handler panic recovered: %v", topicName, r)
        }
    }()
    handler(ctx, msg.Data, msg.FromID)
}()
```

### ChiR4: `getLocalHeight` caps advertised tips to maxLead for up to 1s after a blockchain read recovers
**Problem:** failures are cached as height 0 for the full 1s TTL. This is intentional load-shedding — the comment says so, so a blockchain outage doesn't turn back into a per-message RPC storm, and that's the right call. The side effect is that once the blockchain recovers mid-window, `sanitizeAdvertisedTip` still sees `localHeight == 0` and caps every advertised tip to `MaxUnvalidatedAdvertisedHeightLead` (default 10000) until the entry expires.
**Why it matters:** minor and self-healing, but during a flapping blockchain it can briefly understate peer tips that feed sync-peer selection.
**Fix:** keep the failure-caching, but consider a shorter TTL for the failure entry than for a good read so recovery is picked up faster — or just note the cap side effect in the comment. Not a blocker either way.

## Recap

| ChiR1 | Batcher stop ignores service stop budget | YES | 65% |
| ChiR2 | Ban check un-cached on registry error | | 30% |
| ChiR3 | Handler panic crashes process | | 20% |
| ChiR4 | getLocalHeight failure-cache trade-off | | 10% |

ctnguyen (approved):
Review at commit [a822aa3](https://github.com/gokutheengineer/teranode/commit/a822aa3c29dda5ec2013198f72be94453a18fdc6)

Round 2 looks good — the blocking point is properly closed, not papered over. `stop(ctx)` now threads the service-manager budget all the way through (and I checked: `ServiceManager.Wait` builds `stopCtx` from `context.Background()` with the timeout, so the final flush genuinely survives shutdown while still being bounded), the wedged-flush test locks the behaviour in, and the `IsPeerBanned` error breaker plus per-message `recover` are both real fixes with tests that would fail without them. Nice touch putting `P2PClient.Close()` ahead of the batcher stop so the topic workers are already unwinding by the time the final flush runs. The remaining notes are all things you can take or leave — mostly about the flush goroutine becoming the new serialization point, and a couple of narrow races around the new caches.

## Status of previous points

ChiR1's blocking half — `stop` ignoring the per-service budget — is fully fixed. Its secondary, explicitly-optional half (holding the topic `WaitGroup` on the `Server` so `Stop` can drain in-flight handlers) was not done and is listed as a residual; in practice `s.P2PClient.Close()` runs earlier in `Server.Stop` and closes the topic channels, so the exposure is one message's worth of stats at exit. ChiR4 asked for a shorter failure TTL *or* a comment on the cap side effect and got both (`localHeightErrorCacheTTL` plus the note at `server_helpers.go:464-470`), so it is closed; ChiR6 below is a separate, stronger finding in the same area, not leftover work from ChiR4.

| ChiR1 | Stop ignores service budget | YES | FIXED |
| ChiR2 | Ban check un-cached on error | | FIXED |
| ChiR3 | Handler panic crashes process | | FIXED |
| ChiR4 | getLocalHeight failure-cache | | FIXED |

## Non-blocking issues

### ChiR5 - Flush is fully serial, and a cut-short cycle drops updates instead of requeueing them
**Problem:** `flushOnce` walks `pending` in a single goroutine, issuing up to five sequential RPCs per peer (`RegisterPeer`, `UpdateConnectionState`, `UpdateLastMessageTime`, `UpdatePeerMetrics`, `UpdateStorage`). At the 1s default interval and, say, 200 active gossiping peers that is ~800 serial round-trips per cycle — a 1.25ms per-RPC budget before the flusher stops keeping up; at the 10k `registryBatcherMaxPending` cap it is up to 50k. When it can't keep up, the `ctx.Err() != nil` branch (`peer_registry_batcher.go:323-326`) `return`s and every unvisited peer's coalesced batch is discarded — accumulated `bytesReceived`, the `touchLastMessage` intent, and registration data all gone. That is inconsistent with the `RegisterPeer`-failure path at `peer_registry_batcher.go:356`, which deliberately calls `requeue` precisely so accumulated metrics aren't dropped.
**Why it matters:** the batcher removes registry latency from the gossip hot path by concentrating it on one goroutine, which becomes the new bottleneck. Sustained cut-short flushes mean `LastMessageTime` stops advancing for a share of peers, and `PeerRegistryTTL` eviction treats them as idle even though they are actively gossiping.
**Fix:** requeue on the cut-short path so it matches the failure path, and give the per-peer batches bounded parallelism:
```go
for peerID, u := range pending {
    if ctx.Err() != nil {
        b.requeue(peerID, u) // same contract as the RegisterPeer-failure path
        continue
    }
    ...
}
```
plus an `errgroup` with `SetLimit(8)` (or similar) over peers — the per-peer RPC order that `RegisterPeer`-first depends on is preserved inside each peer's batch, and only distinct peers run concurrently.

### ChiR6 - An unknown local height is written into the registry as a `maxLead`-capped peer height
**Problem:** `getLocalHeight` returns `0` both for "height is genuinely 0" and for "the read failed", and now caches the failure for `localHeightErrorCacheTTL`. `sanitizeAdvertisedTip` then computes `maxAcceptedHeight = 0 + MaxUnvalidatedAdvertisedHeightLead` (default 10000) and the *capped* value is what `handleBlockTopic` (`server_helpers.go:120`) and `handleNodeStatusTopic` (`Server.go:1140`) hand to `addPeer` — so it lands in the shared registry. On mainnet a peer advertising ~900k is persisted as `Height: 10000` paired with its real tip hash. `peerEligibleForAdvertisedProbe` gates on `p.Height > localHeight` (`sync_coordinator.go:218`), so those peers drop out of catchup probing until each one's next gossip message rewrites a good height — and the batcher's max-merge only protects within a flush window. This is pre-existing in kind (the old code capped on every failing message too), but caching turns "unknown" into a state that persists, which is what makes it worth separating from the load-shedding decision itself.
**Why it matters:** a blockchain-client blip during catchup can transiently write heights *below* the local tip for every gossiping peer, which is exactly when sync-peer selection matters most.
**Fix:** make "unknown" distinguishable and skip the cap rather than capping against a height you don't have:
```go
func (s *Server) localHeight() (uint32, bool) { /* returns ok=false for error/no-client */ }

// sanitizeAdvertisedTip: when the local height is unknown, pass the advertised
// height through unchanged instead of capping it to maxLead.
```
Cheapest variant if you'd rather not touch the signature: skip the `addPeer` height/hash write when the height entry is an error entry.

### ChiR7 - A stale negative `IsPeerBanned` result can clobber `onPeerBanned`'s immediate ban entry
**Problem:** `shouldSkipBannedPeer` does an unconditional `s.banStatusCache.Store(from, {banned: <result>})` after the lookup returns. If `onPeerBanned` stores `banned: true` while that gRPC call is in flight, the returning `false` overwrites it — the peer's gossip is then accepted for the full `reputationCacheTTL` (5s), silently defeating the "make the ban effective immediately" guarantee the new `onPeerBanned` write exists to provide. The window is one `IsPeerBanned` round-trip, and the worker pool makes it easier to hit: one worker can be mid-lookup while another worker (or the invalid-blocks Kafka consumer, or the `AddBanScore` RPC) crosses the threshold. The same clobber applies to the error-breaker store on line 953. `TestServerHelpers_ShouldSkipBannedPeer_LocalBanImmediate` is single-threaded so it can't see this.
**Why it matters:** it reopens for 5s the enforcement gap that `onPeerBanned`'s cache write was added to close, on the path that drops traffic from banned peers before any further processing.
**Fix:** never let a lookup result downgrade a fresh ban, and make the check-then-write atomic — a plain `Load` followed by `Store` just moves the race, since `onPeerBanned` can land in between. Either take a small dedicated mutex around every `banStatusCache` write, or use a CAS retry loop:
```go
func (s *Server) storeBanStatus(peerID string, banned bool) {
    entry := banStatusCacheEntry{banned: banned, expiresAt: time.Now().Add(reputationCacheTTL)}
    if banned {
        s.banStatusCache.Store(peerID, entry) // a ban always wins
        return
    }
    for {
        prev, loaded := s.banStatusCache.Load(peerID)
        if !loaded {
            if _, dup := s.banStatusCache.LoadOrStore(peerID, entry); !dup {
                return
            }
            continue // someone inserted concurrently; re-inspect it
        }
        if e := prev.(banStatusCacheEntry); e.banned && time.Now().Before(e.expiresAt) {
            return // a ban recorded while our lookup was in flight wins
        }
        if s.banStatusCache.CompareAndSwap(peerID, prev, entry) {
            return
        }
        // lost the CAS: the entry changed under us, retry against the new value
    }
}
```
`banStatusCacheEntry` is comparable (`bool` + `time.Time`), so `CompareAndSwap` is safe here. Route both the success store and the error-breaker store through the helper, and have `onPeerBanned` use it too so the "ban wins" rule lives in one place.

### ChiR8 - Advertised height can regress across flush windows now that a topic is processed out of order
**Problem:** `merge`'s max-height rule only holds *within* one flush window. `CentralizedPeerRegistry.Register` overwrites unconditionally on `info.Height > 0` (`services/blockchain/peer_registry.go:308`) rather than max-merging, so worker A enqueuing height 100 in window N and worker B — still working through an older message — enqueuing 99 in window N+1 leaves the registry at 99 until the peer's next message. The single per-topic goroutine made this impossible before, so the PR description's "concurrent workers cannot regress an advertised tip" is narrower than it reads.
**Why it matters:** a regressed height can briefly drop a peer below `p.Height > localHeight` and out of advertised-probe eligibility. Low impact and self-healing, but it is a genuine behavioural change from strictly-ordered processing.
**Fix:** a hard monotonic guarantee has to live registry-side — `existing.Height = max(existing.Height, info.Height)` in `Register`, with the block hash following the winning height. If you'd rather keep the registry as-is, have the batcher remember the last flushed height per peer (it already keeps `lastAsserted`) and drop a lower one; either way, please soften the claim in the description.

### ChiR9 - Recovered gossip panics log no stack trace
**Problem:** `handleTopicMessage` logs `%v` of the recover value only. A panic recovered five frames deep inside `handleBlockTopic` leaves you with `gossip handler panic recovered (peer X): runtime error: index out of range [3]` and no way to find the line. The repo already pairs recover with `debug.Stack()` elsewhere (`services/blockassembly/subtreeprocessor/SubtreeProcessor.go:3086`, `stores/utxo/aerospike/create.go:334`).
**Why it matters:** the whole point of ChiR3's fix is to survive a crafted message; surviving it without being able to diagnose it means the underlying bug stays open, and a flood of such messages produces an unbounded, information-free log stream.
**Fix:** log the stack alongside the recover value:
```go
s.logger.Errorf("[%s] gossip handler panic recovered (peer %s): %v\n%s", topicName, fromID, r, debug.Stack())
```
Consider reusing the same per-peer TTL breaker pattern you added for `IsPeerBanned` to rate-limit the line under a flood.

### ChiR10 - `pruneAssertState` does a full O(n) scan under the enqueue mutex on every flush
**Problem:** every `flushOnce` — including the empty-pending fast path — calls `pruneAssertState`, which holds `b.mu` while iterating all of `lastAsserted` (up to 10k) and all of `removed`. Entries only age out after `registryAssertStatePruneAge` (10 min), so at the 1s default interval roughly 599 of every 600 scans find nothing, and `b.mu` is the same mutex every gossip handler takes to enqueue.
**Why it matters:** a recurring, avoidable latency blip on the hot path the PR set out to relieve; it grows linearly with peer count.
**Fix:** gate it on a `lastPruned` timestamp so it runs at most once per `registryAssertStatePruneAge/2` — the count bound in `recordAssertStateLocked` already handles the memory-pressure case, so the periodic sweep only needs to be occasional.

## Recap

| ChiR5 | Serial flush, drops on timeout |  | 32% |
| ChiR6 | Unknown height capped into registry |  | 28% |
| ChiR7 | Negative lookup clobbers fresh ban |  | 24% |
| ChiR8 | Height regresses across windows |  | 15% |
| ChiR9 | Recover logs no stack trace |  | 8% |
| ChiR10 | Prune scans under hot mutex |  | 5% |

liam (commented):
Concurrency-focused review of the batcher, worker pool, caches, and shutdown. **The concurrency primitives are sound** - no data races (all `pending`/`dropped`/`lastAsserted`/`removed` access under `b.mu`, `banStatusCache`/`localHeightCache` correct), `b.mu` is never held across a gRPC call, worker-pool panic recovery correctly isolates one message without leaking a slot, `pending`/`lastAsserted` are genuinely count-bounded at 10k, and all four `stop()` states were traced with no deadlock/double-close/send-on-closed. Nice work on the deterministic barrier-based tests.

One real gap worth fixing before merge, plus two low-severity follow-ups:

**1. (Medium) Non-`RegisterPeer` RPC failures silently drop accumulated data.** In `flushOnce`, only the `RegisterPeer` branch calls `requeue()`. If `RegisterPeer` succeeds (or is skipped because `sendRegister` is false) but `UpdatePeerMetrics` / `UpdateLastMessageTime` / `UpdateStorage` fails independently - plausible under a degraded registry where individual unary RPCs time out - that peer's coalesced `bytesReceived` / last-message / storage intents are discarded with the local `u`; nothing re-adds them. This contradicts the PR description ("byte deltas... not lost on flush failure") and `requeue()`'s own doc comment. `UpdateConnectionState` self-heals (gated by `connectedAt`/re-assert TTL), but the other three don't. It's metrics/freshness drift, not a crash, and it's untested (`TestPeerRegistryBatcher_RequeueOnRegisterFailure` only covers the RegisterPeer path). Suggest requeuing on any RPC failure.

**2. (Low) `removed` tombstone map is time-bounded but not count-bounded.** `forget()` does `b.removed[peerID] = time.Now()` with no `>= registryBatcherMaxPending` cap, unlike `pending` and `lastAsserted`; it's only pruned by age (60s). The "both maps bounded" note misses this third map. Lower risk since filling it requires real connect/disconnect churn (a libp2p handshake per entry), not a string-only flood, but it's a genuine gap in the stated invariant - same swept-then-cap fix applies.

**3. (Low) Narrow tombstone-clear interleaving.** If `flushOnce` snapshots pending, then `forget(X)` tombstones X, then X re-enqueues (clearing the tombstone) before the loop reaches X, the loop sees `isRemoved=false` and flushes X's stale pre-removal snapshot. Consequence is benign (a slightly-stale RegisterPeer; new data flushes next cycle) but it weakens the documented "neither resurrected nor skipped" guarantee and isn't covered by the existing forget-during-flush test.

**Nit:** the "at most one message's worth of stats" shutdown-loss framing undersells the bound - workers aren't drained before the final flush, so it's closer to `GossipHandlerConcurrency x topics`.

liam:
Thanks for the quick turnaround on b76148be3 - I re-reviewed it and all three points are resolved correctly, each with a dedicated regression test:

- **Finding 1 (requeue any failed RPC):** the per-RPC `failed` batch + `requeue` is right. `merge` accumulates `bytesReceived` (`+=`) and the requeue folds the failed batch *under* newer observations (`merge(u)` then `merge(existing)`), so a byte delta is neither lost nor double-counted and any fresh re-enqueued data still wins. Good call guarding `requeue` with `isRemovedLocked` so a peer forgotten between snapshot and requeue isn't resurrected. `TestPeerRegistryBatcher_RequeueOnMetricsFailure` covers it.
- **Finding 2 (tombstone bound):** `forget()` now sweeps-then-caps `removed` at `registryBatcherMaxPending`, matching `pending`/`lastAsserted`; the `lastAsserted` delete still forces re-registration if the tombstone is skipped. `TestPeerRegistryBatcher_TombstoneMapCountBounded` locks it in.
- **Finding 3 (forget→re-enqueue during flush):** `removedDuringFlush` (checked in `isRemovedLocked`, reset in the defer) closes the interleaving cleanly - nice that it's separate from the persistent tombstone so a re-enqueue can't clear it mid-cycle. `...ReenqueueDuringFlushDoesNotFlushStaleSnapshot` is a good deterministic guard.

One precise note (not a request): this fixes the per-RPC-failure drop; the separate cut-short path (`if ctx.Err() != nil { return }`) still drops unvisited peers rather than requeueing them - that's @ctnguyen's ChiR5, already flagged non-blocking, so just noting the two are distinct and that path is unchanged.

Resolves my review. 👍

liam (approved):

Issue #1346, opened 2026-07-25
Publish an operator runbook for teranode-cli rewindblockchain


oskarszoon:
## What

`teranode-cli rewindblockchain` is reachable on a deployed node as of #1345, but there is no operator runbook for it. It is the most dangerous command in the CLI — destructive, irreversible, and with preconditions it does not fully enforce — and `dangerousCommands` is empty tool-wide, so there is no typed-confirmation backstop either. The CLI reference entry added in #1345 states the preconditions inline and says explicitly that no runbook is published yet.

A 344-line runbook was written and then pulled back out of #1345, because review found two P0s and six P1s in it. Every one was about the operator's environment rather than the tool, and each is a sentence, a flag, or one extra command — but shipping it half-right is worse than not shipping it, since an operator following a wrong step lands in an unrecoverable state *through the document's own safeguard*.

This issue tracks doing it properly. The findings below are all verified against the source, not speculative.

## Design already decided

- **Docker: route through [teranode-quickstart](https://github.com/bsv-blockchain/teranode-quickstart).** `minersHowToStopStartDockerTeranode.md` and `minersHowToInstallation.md` both send miners there, and it loads `.env`, selects Compose profiles, and handles FSM transitions. Note the wrinkle: `stop.sh` runs `docker compose down`, which removes the datastore containers the rewind reads and writes through, so the procedure needs an explicit deviation — bring the datastores back before the rewind, or stop only the Teranode services.
- **Kubernetes: state requirements, do not prescribe a manifest.** Operators own how they run one-off pods. The runbook should say what the pod needs — an image containing the subcommand, the node's configmap, the same PVC mounts the Teranode pods used, datastores up — and let them choose the mechanism.

## P0

**1. The settings check cannot detect a wrong or missing subtree-blob mount, and the draft told operators to trust it.**

`teranode-cli settings | grep -iE 'blockchain_store|utxostore|subtreestore'` looks like a go/no-go gate but is not one. `blockchain_store` and `utxostore` print network URLs, correct by construction, so they prove nothing about mounts. `subtreestore` prints `file://./data/subtreestore` — a container-relative path, byte-identical whether the real data directory is mounted, a wrong one is, or nothing is. `deploy/docker/mainnet/.env` sets `DATA_PATH=${PWD}/data`, and `${PWD}` resolves from the invoking shell's cwd, not the project directory.

[...]

PR #1352, opened 2026-07-27
fix(rewindblockchain): open the subtree store with the node's hashPrefix (#1351)

### LINKED ISSUE #1351: rewindblockchain: subtree store opened without WithHashPrefix — every subtree read misses, Phase 2 cannot rewind any file:// deployment
## What

`cmd/rewindblockchain` opens the subtree blob store without `options.WithHashPrefix`, so it resolves flat filenames while the node writes hash-prefix-sharded ones. Every subtree read misses with `NOT_FOUND`, and Phase 2 dies on the first block. The tool cannot rewind any `file://` subtree store as shipped.

`cmd/rewindblockchain/rewindblockchain/rewind.go:142`:

```go
subtreeStore, err := blob.NewStore(logger, s.SubtreeValidation.SubtreeStore)   // no options
```

vs what every node does — `daemon/daemon_stores.go:454-482`:

```go
hashPrefix := 2                                       // default when the URL has no ?hashPrefix
if subtreeStoreURL.Query().Get("hashPrefix") != "" { … }
d.mainSubtreeStore, err = blob.NewStore(logger, subtreeStoreURL,
    options.WithHashPrefix(hashPrefix),
    options.WithBlockHeightCh(ch),
    options.WithBlobDeletionScheduler(blobDeletionScheduler),
    options.WithStoreType(storetypes.SUBTREESTORE))
```

No option → `HashPrefix = 0` → `stores/blob/options/Options.go:418` omits the shard directory from the constructed path.

Adding `?hashPrefix=2` to the URL does **not** work around it: `stores/blob/factory.go` parses only `batch`, `logger`, `sizeInBytes` and `writeKeys` from the query. `hashPrefix` is read by the daemon and passed as an option; the factory never looks at it. Any caller that doesn't pass the option explicitly gets a flat layout.

Same omission likely applies to any other blob store the tool opens or gains later — worth auditing `resolveStores` against `daemon_stores.go` as a whole (blockstore/persister at `daemon_stores.go:428,551,604,656` all pass `WithHashPrefix`).

## Repro

testnet-eu-1, v0.15.8-beta-1, rewinding 1749337 → 1749330:

```
2026-07-27T08:53:09Z | WARN | model/Block.go:1243 | [BLOCK][000000004600bfcd…][ID 1951427] failed to get subtree
  8379abf4004a170d5fff985b02d3fb461c2256f650e14a7d4e2d5bb0d0adb644 (given up after 3 attempts): NOT_FOUND (3): not found
Error executing command: Phase 2 failed -> rewinding block 000000004600bfcd… (height 1749337)
  -> GetSubtrees -> [BLOCK][…][ID 1951427] failed to get subtree 8379abf4… -> NOT_FOUND (3): not found
```

The file is present, sharded:

```
root@block-persister:/data/teranode-testnet-eu-1/subtreestore/83# ls 8379abf4004a170d5fff985b02d3fb461c2256f650e14a7d4e2d5bb0d0adb644*
8379abf4….subtree          2184 bytes
8379abf4….subtreeData     11995
8379abf4….subtreeMeta      1980
8379abf4….subtreeToCheck   2184
(+ .sha256 for each)
```

Resolved config in the same pod: `subtreestore[.operator]=file:///data/teranode-testnet-eu-1/subtreestore` (no query params, so the node's default `hashPrefix=2` applies). The asset server serves the same subtree fine (44 nodes, matching the block's `txCount: 44`), confirming the data is intact and the tool is looking in the wrong place.

Reproduced from two different pods, so it is not a volume-mount issue.

## Why it shipped

`rewind_test.go:189,482` injects `memory.New()` via `Options.Stores`, which bypasses `resolveStores` entirely. The production store-construction path has no test coverage. `#728` added the tool; `#1341` proposes shipping it — that would ship something that cannot work against a real `file://` deployment.

## Fix

In `resolveStores`, mirror the daemon:

```go
subtreeStoreURL := s.SubtreeValidation.SubtreeStore

// Match daemon.GetSubtreeStore: hashPrefix defaults to 2 and is read from the
// URL query there — blob.NewStore never parses it itself.
hashPrefix := 2
if v := subtreeStoreURL.Query().Get("hashPrefix"); v != "" {
    if hashPrefix, err = strconv.Atoi(v); err != nil {
        return nil, false, errors.NewConfigurationError("subtreestore hashPrefix config error: %w", err)
    }
}

subtreeStore, err := blob.NewStore(logger, subtreeStoreURL, options.WithHashPrefix(hashPrefix))
if err != nil {
    return nil, false, errors.NewConfigurationError("failed to open subtree blob store: %w", err)
}
```

Only `WithHashPrefix` is needed — the daemon's other options drive DAH-managed lifecycle, and the tool wants raw `Del`s.

Better still: have `blob.NewStore` read `hashPrefix` from the URL query the way it already reads `batch`/`sizeInBytes`, so no caller can get this wrong again. That changes behaviour for every caller, so it needs its own review; the local fix above is the safe one for the v0.15 line.

## Test

`resolveStores` needs coverage against a real `file://` store: write a subtree via a store built the daemon's way (`WithHashPrefix(2)`), then assert the tool's own `resolveStores` handle can read and delete it. A test that only exercises `Options.Stores` cannot catch this class of bug.

Related: #1340, #1341 (both `rewindblockchain`).

oskarszoon:
Closes #1351.

## What

`resolveStores` opened the subtree blob store with no options, so `Options.HashPrefix` stayed `0` and `CalculatePrefix` returned `""`. The tool resolved flat filenames while the node writes hash-sharded ones: every subtree read missed with `NOT_FOUND` and Phase 2 died on the first block. `rewindblockchain` could not rewind any `file://` deployment as shipped — which is every deployment.

Mirrors `daemon.GetSubtreeStore` (`daemon/daemon_stores.go:454-479`): hashPrefix defaults to 2, the `?hashPrefix=` query overrides it, and it reaches the store as `options.WithHashPrefix`. Only that option — the daemon's others drive DAH-managed lifecycle, and this tool wants raw `Del`s.

**Correction from review** (thanks @ordishs, @freemans13): the original version of this description — and the comment in the code — claimed `?hashPrefix=2` on the URL could not work around this, because `stores/blob/factory.go` parses only `batch`, `logger`, `sizeInBytes` and `writeKeys`. That is true of the factory, but the **file backend parses the query itself** (`stores/blob/file/file.go:446-462`) and applies it *after* the store options, so on `file://` the query would in fact have won.

The fix stands, for two better reasons now stated in the code instead:

- No shipped `subtreestore` setting carries a query — they are all bare `file://` URLs — so with the default configuration the option is the only source of the node's `hashPrefix=2`.
- `s3` does not read the query at all (it derives its path via `CalculatePrefix` from the store options alone), so the option is the only route that works across backends.

Also now recorded in the code: neither this nor `daemon.GetSubtreeStore` reads `?hashSuffix`, which the file backend turns into a negative `HashPrefix` and which would therefore override what is passed. Inherited parity gap, not introduced here.

## Why nothing caught it

Every existing test injects `memory.New()` via `Options.Stores`, which returns before `resolveStores` constructs anything. The production store-construction path had no coverage at all.

Five tests now cover it, against a real `file://` store in a temp dir plus direct coverage of the prefix resolution:

- **`TestResolveStores_SubtreeStoreReadsWhatTheNodeWrote`** — writes a blob through a store built the daemon's way (`WithHashPrefix(2)`), asserts the fixture really is sharded, then drives `resolveStores` itself and requires `Exists`/`Get`/`Del` all land on it.
- **`TestNewSubtreeStore_FlatLayoutMissesShardedBlob`** — pins the mechanism, so the fix can't be "fixed" by something that merely happens to work.
- **`TestSubtreeHashPrefix`** — the prefix resolution itself: the default of 2 on a bare URL, the `?hashPrefix=` override, and a non-numeric value rejected.
- **`TestNewSubtreeStore_RejectsBadHashPrefix`** / **`TestNewSubtreeStore_RejectsNilURL`** — the error surfaces through the constructor, and the nil-URL guard (`blob.NewStore` dereferences `storeURL.Scheme` before any nil check, so a missing setting used to panic).

**Also from review:** the first two of those were originally written as layout assertions and *could not fail* — because the file backend honours `?hashPrefix` itself, they passed with the entire parse block deleted, and `RejectsBadHashPrefix` passed because `file.newStore` rejects `banana` with a message that also contains the substring `hashPrefix`. They now assert on the resolved value and on this tool's own `subtreestore hashPrefix config error` text. Verified by deleting the parse: both fail.

Verified the regression test actually bites: reintroducing the bug as `WithHashPrefix(0)` fails it with

```
Error:    Should be true
Messages: resolveStores must open the subtree store with the node's hashPrefix;
          without it every subtree read misses and Phase 2 cannot rewind
```

## Separate bug found by the audit

The issue asks to audit `resolveStores` against `daemon_stores.go` as a whole. The subtree store is the only blob store it opens, so that's the only `WithHashPrefix` gap — but the audit turned up something else.

`deleteUTXOPersisterLastProcessed` (`phase3_finalize.go`) has never worked:

| | Persister writes (`services/utxopersister/Server.go:772,826`) | Tool deletes |
|---|---|---|
| Store | **block** store | **subtree** store — different directory |
| Filename | `"lastProcessed"` → `lastProcessed.dat` | `"lastProcessed.dat"` → `lastProcessed.dat.dat` (`ConstructFilename` appends `"." + fileType`) |
| Prefix | `WithNoHashPrefix()` | none |

The `Del` is best-effort and its miss is swallowed by a `Debugf`, which is why it went unnoticed. Fixing it needs the block store, which `resolveStores` does not open — so it's documented in place and left for its own change rather than widened into this one.

Filed as #1353.

What this PR does do is pass `WithNoHashPrefix()` on that `Del`. That keeps the path byte-identical now the store carries a prefix — and, as review pointed out, avoids a second effect: without it `CalculatePrefix` would derive `"la"` from `"lastProcessed.dat"`, and `ConstructFilename` `MkdirAll`s the prefix folder on every call including `Del` (`Options.go:397-401`), so each rewind would leave a stray empty `<subtreestore>/la/` behind while still deleting nothing.

## Testing

`go test ./cmd/rewindblockchain/... ./cmd/teranodecli/...` 57/57, same with `-race`. `go vet` clean. `make lint` 0 issues.

## Scope

`main` only. Backports are handled separately.

## Follow-ups from review

- `newSubtreeStore` is now near-verbatim `daemon_stores.go:454-462`. A shared settings-level helper would stop the two drifting apart the next time the default moves — which is exactly the class of bug being fixed here. Worth its own change.
- Now that subtree deletes actually land, scheduled deletions in `stores/blockchain/sql/blob_deletion.go` deserve a look across a rewind: a row scheduled for subtree X survives, and if the node re-syncs and rewrites X the pruner could delete a live blob when the chain re-crosses that height. Independent of this diff — the tool never schedules deletions — but the runbook (#1346) should cover it.

### DIFF
--- a/cmd/rewindblockchain/rewindblockchain/phase3_finalize.go
+++ b/cmd/rewindblockchain/rewindblockchain/phase3_finalize.go
@@ -154,9 +154,23 @@ func (e *env) resetBlockPersisterHeight(ctx context.Context, pf *preflightResult
 // deleteUTXOPersisterLastProcessed removes the file the utxo persister uses
 // to track its position. On next startup it will recompute from block
 // data — which is the correct behaviour post-rewind.
+//
+// NOTE: this Del does not currently match what the persister writes, and never
+// has. services/utxopersister/Server.go:772,826 uses the BLOCK store with
+// options.WithFilename("lastProcessed") and options.WithNoHashPrefix(), which
+// resolves to <blockstore>/lastProcessed.dat. This call uses the SUBTREE store
+// with filename "lastProcessed.dat", and ConstructFilename appends "." +
+// fileType, so it targets <subtreestore>/lastProcessed.dat.dat — a different
+// store and a different name. The Del is best-effort, so the miss is swallowed
+// by the Debugf below. Fixing it needs the block store, which resolveStores
+// does not open; tracked separately rather than widened into the hashPrefix fix.
+//
+// WithNoHashPrefix is passed so that giving the subtree store the node's
+// hashPrefix (see newSubtreeStore) leaves this path exactly as it was.
 func (e *env) deleteUTXOPersisterLastProcessed(ctx context.Context) error {
 	key := []byte(utxoPersisterLastHeightFn)
-	if err := e.subtreeStore.Del(ctx, key, fileformat.FileTypeDat, options.WithFilename(utxoPersisterLastHeightFn)); err != nil {
+	if err := e.subtreeStore.Del(ctx, key, fileformat.FileTypeDat,
+		options.WithFilename(utxoPersisterLastHeightFn), options.WithNoHashPrefix()); err != nil {
 		// Best-effort. Log and continue.
 		e.logger.Debugf("delete utxo-persister lastProcessed.dat: %v", err)
 	}
--- a/cmd/rewindblockchain/rewindblockchain/phase3_finalize.go
+++ b/cmd/rewindblockchain/rewindblockchain/phase3_finalize.go
@@ -163,10 +163,14 @@ func (e *env) resetBlockPersisterHeight(ctx context.Context, pf *preflightResult
 // fileType, so it targets <subtreestore>/lastProcessed.dat.dat — a different
 // store and a different name. The Del is best-effort, so the miss is swallowed
 // by the Debugf below. Fixing it needs the block store, which resolveStores
-// does not open; tracked separately rather than widened into the hashPrefix fix.
+// does not open; tracked in #1353 rather than widened into the hashPrefix fix.
 //
-// WithNoHashPrefix is passed so that giving the subtree store the node's
-// hashPrefix (see newSubtreeStore) leaves this path exactly as it was.
+// WithNoHashPrefix does more than keep this path byte-identical now that the
+// store carries the node's hashPrefix (see newSubtreeStore). Without it
+// CalculatePrefix would derive "la" from the filename "lastProcessed.dat", and
+// ConstructFilename MkdirAlls the prefix folder on every call including Del
+// (stores/blob/options/Options.go:397-401) — so each rewind would leave a stray
+// empty <subtreestore>/la/ behind while still deleting nothing.
 func (e *env) deleteUTXOPersisterLastProcessed(ctx context.Context) error {
 	key := []byte(utxoPersisterLastHeightFn)
 	if err := e.subtreeStore.Del(ctx, key, fileformat.FileTypeDat,
--- a/cmd/rewindblockchain/rewindblockchain/rewind.go
+++ b/cmd/rewindblockchain/rewindblockchain/rewind.go
@@ -2,11 +2,13 @@ package rewindblockchain
 
 import (
 	"context"
+	"strconv"
 	"time"
 
 	"github.com/bsv-blockchain/teranode/errors"
 	"github.com/bsv-blockchain/teranode/settings"
 	"github.com/bsv-blockchain/teranode/stores/blob"
+	"github.com/bsv-blockchain/teranode/stores/blob/options"
 	"github.com/bsv-blockchain/teranode/stores/blockchain"
 	"github.com/bsv-blockchain/teranode/stores/utxo"
 	utxofactory "github.com/bsv-blockchain/teranode/stores/utxo/factory"
@@ -139,9 +141,9 @@ func resolveStores(ctx context.Context, logger ulogger.Logger, s *settings.Setti
 		return nil, false, errors.NewConfigurationError("failed to open utxo store", err)
 	}
 
-	subtreeStore, err := blob.NewStore(logger, s.SubtreeValidation.SubtreeStore)
+	subtreeStore, err := newSubtreeStore(logger, s)
 	if err != nil {
-		return nil, false, errors.NewConfigurationError("failed to open subtree blob store", err)
+		return nil, false, err
 	}
 
 	return &Stores{
@@ -149,6 +151,43 @@ func resolveStores(ctx context.Context, logger ulogger.Logger, s *settings.Setti
 		UTXO:       utxoStore,
 		Subtree:    subtreeStore,
 	}, true, nil
+}
+
+// newSubtreeStore opens the subtree blob store the same way the node does.
+//
+// blob.NewStore does NOT read hashPrefix from the URL query — stores/blob/factory.go
+// parses only batch, logger, sizeInBytes and writeKeys. hashPrefix reaches the store
+// solely as an option, which the daemon supplies (daemon/daemon_stores.go:454-479).
+// Without it Options.HashPrefix stays 0, CalculatePrefix returns "", and every key
+// resolves to a flat path while the node writes hash-sharded ones — so every read
+// misses with NOT_FOUND and Phase 2 cannot rewind a file:// deployment at all.
+//
+// Only WithHashPrefix is mirrored here: the daemon's other options drive
+// DAH-managed lifecycle, and this tool wants raw Dels.
+func newSubtreeStore(logger ulogger.Logger, s *settings.Settings) (blob.Store, error) {
+	subtreeStoreURL := s.SubtreeValidation.SubtreeStore
+	if subtreeStoreURL == nil {
+		return nil, errors.NewConfigurationError("subtreestore URL is not configured")
+	}
+
+	// Same default and query override as daemon.GetSubtreeStore.
+	hashPrefix := 2
+
+	if v := subtreeStoreURL.Query().Get("hashPrefix"); v != "" {
+		parsed, err := strconv.Atoi(v)
+		if err != nil {
+			return nil, errors.NewConfigurationError("subtreestore hashPrefix config error", err)
+		}
+
+		hashPrefix = parsed
+	}
+
+	subtreeStore, err := blob.NewStore(logger, subtreeStoreURL, options.WithHashPrefix(hashPrefix))
+	if err != nil {
+		return nil, errors.NewConfigurationError("failed to open subtree blob store", err)
+	}
+
+	return subtreeStore, nil
 }
 
 // env bundles the resolved stores and counters shared across phases.
--- a/cmd/rewindblockchain/rewindblockchain/rewind.go
+++ b/cmd/rewindblockchain/rewindblockchain/rewind.go
@@ -2,6 +2,7 @@ package rewindblockchain
 
 import (
 	"context"
+	"net/url"
 	"strconv"
 	"time"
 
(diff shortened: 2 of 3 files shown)


icellan (approved):

ordishs (approved):
Approving the fix — it's correct, minimal, and the important test genuinely bites. Two asks below are on the comments and two of the tests, not on the behavioural change.

I verified the core regression test fails when the fix is neutralised (`WithHashPrefix(hashPrefix)` → `WithHashPrefix(hashPrefix*0)`):

```
--- FAIL: TestResolveStores_SubtreeStoreReadsWhatTheNodeWrote (0.03s)
    rewind_test.go:1556: Should be true
        resolveStores must open the subtree store with the node's hashPrefix; without it every subtree read misses and Phase 2 cannot rewind
FAIL
```

`go vet ./cmd/rewindblockchain/...` clean, `go test -race ./cmd/rewindblockchain/...` → `ok ... 5.709s`.

## 1. The load-bearing comment is wrong for the `file://` backend

`rewind.go:158-160` — repeated in the PR description and in two test comments — says:

> `blob.NewStore` does NOT read hashPrefix from the URL query — `stores/blob/factory.go` parses only batch, logger, sizeInBytes and writeKeys.

True of `factory.go`, but `stores/blob/file/file.go:445-462` parses both `hashPrefix` and `hashSuffix` from the query, and because it runs *after* `options.NewStoreOptions(opts...)` the query value **overrides** the store option:

```go
storeOpts := options.NewStoreOptions(opts...)
if hashPrefix := storeURL.Query().Get("hashPrefix"); len(hashPrefix) > 0 { ... storeOpts.HashPrefix = int(val) }
if hashSuffix := storeURL.Query().Get("hashSuffix"); len(hashSuffix) > 0 { ... storeOpts.HashPrefix = -int(val) }
```

The bug and the fix stand — with no query param (every real deployment) the option is the only source, so `HashPrefix` stayed 0. But as written the comment will mislead the next reader, and it is the stated justification for the code beneath it. Suggest narrowing it to: *`factory.go` does not parse hashPrefix; the file backend does, but only when the query param is present — with the default (absent) it must arrive as a store option, which is what the daemon supplies.*

Worth adding the real reason to keep the explicit parsing: `s3` (`s3.go:548` uses `CalculatePrefix`) does **not** read the query, so without it an `s3://` subtree store would ignore a configured `hashPrefix`.

## 2. Two of the four tests cannot fail

Following from the above: for a `file://` store the query is honoured by `file.New` regardless of what the tool does. Removing the `strconv.Atoi` block from `newSubtreeStore` entirely:

```
=== RUN   TestNewSubtreeStore_HonoursHashPrefixQuery
--- PASS: TestNewSubtreeStore_HonoursHashPrefixQuery (0.04s)
=== RUN   TestNewSubtreeStore_RejectsBadHashPrefix
--- PASS: TestNewSubtreeStore_RejectsBadHashPrefix (0.00s)
PASS
```

Both pass with the branch they claim to cover deleted. `RejectsBadHashPrefix` passes because `file.New` returns `"[File] failed to parse hashPrefix"`, which also satisfies the `Contains(err, "hashPrefix")` assertion. Given the standard this PR sets for itself — "pins the mechanism, so the fix can't be 'fixed' by something that merely happens to work" — these two don't meet it. In order of preference:

1. Assert on the resolved option rather than on observable file layout (type-assert the returned store, or have `newSubtreeStore` also return the resolved prefix) — discriminates for every backend.
2. Tighten to `require.Contains(err.Error(), "subtreestore hashPrefix config error")`, which only the tool's path produces.
3. At minimum, drop the claim in the description that these cover the override branch.

`TestResolveStores_SubtreeStoreReadsWhatTheNodeWrote` and `TestNewSubtreeStore_FlatLayoutMissesShardedBlob` are both solid — the `require.DirExists` fixture guard is exactly right.

## Smaller points

- **`WithNoHashPrefix()` does more than preserve behaviour** — worth saying so. Without it, `ConstructFilename` would derive prefix `"la"` from `"lastProcessed.dat"` and `MkdirAll` a stray `<subtreestore>/la/` on every rewind; the `MkdirAll` fires on the `Del` path too (`Options.go:396-401`).
- **Cite the tracking issue.** The `NOTE` says "tracked separately" — that's #1353, which exists. An uncited "tracked separately" rots.
- **Drift risk.** `newSubtreeStore` is now a near-verbatim copy of `daemon_stores.go:454-462`. A shared helper (e.g. a settings-level `SubtreeStoreHashPrefix()`) would stop the two diverging next time the default moves — which is precisely the class of bug being fixed here. Fine as a follow-up.
- **The nil-URL guard is a real improvement** and isn't mentioned in the description — `blob.NewStore` would previously have nil-dereferenced on `storeURL.Scheme`.
- `strconv.Atoi` silently accepts negative (`-2` → hash *suffix*) and oversized values. That is exact parity with the daemon, so correct to leave alone here.

## Conventions

gci-correct import grouping, typed `errors.NewConfigurationError` throughout, single-line logs, `require` over `assert`, no `t.Parallel()`. Consistent with AGENTS.md. No performance or security surface change — the prefix is derived from a hex-encoded hash and `ConstructFilename`'s traversal validation is untouched.

## Out of scope, worth its own look

Now that subtree deletes actually land, it would be worth confirming how scheduled deletions in `stores/blockchain/sql/blob_deletion.go` behave across a rewind: a row scheduled for subtree X survives the rewind, and if the node re-syncs and re-writes X, the pruner deletes a live blob when the chain re-crosses that height. That hazard is independent of this diff (the tool never schedules deletions), so it is not a reason to hold this PR — but the rewind runbook (#1346) should probably cover it.

freemans13 (commented):
Reviewed this against `main`. The fix is correct and the bug is real — I reproduced it by forcing the option back to `WithHashPrefix(0)` and watching `TestResolveStores_SubtreeStoreReadsWhatTheNodeWrote` fail exactly as described. Two things about the reasoning around it, though.

## The URL query *is* read — by the file backend

The PR argues in three places (the comment at `rewind.go:158-160`, a test comment, and the description) that `hashPrefix` can only reach the store as an option, and therefore that `?hashPrefix=2` on the setting is not a workaround. That's true of the factory, but `stores/blob/file/file.go:446-453` parses it for itself — and applies it *after* the options, so the URL takes precedence:

```go
storeOpts := options.NewStoreOptions(opts...)

if hashPrefix := storeURL.Query().Get("hashPrefix"); len(hashPrefix) > 0 {
	val, err := strconv.ParseInt(hashPrefix, 10, 32)
	...
	storeOpts.HashPrefix = int(val)
}
```

Verified against a real temp-dir `file://` store: opened with **no** `WithHashPrefix` but `?hashPrefix=4` in the URL, the blob shards four deep and `Exists` finds it; opened with a deliberately conflicting `options.WithHashPrefix(0)` and the same URL, still sharded, still found. So on `file://` that workaround would have worked.

This doesn't make the change wrong — it's still the right fix, just for two better reasons the comment could give instead: real deployment URLs carry no query at all, so the option is the only source for the node's default of 2; and `s3`, `http` and `memory` genuinely don't parse the query, so the option is the only portable route.

Related, and pre-existing rather than introduced here: `file.go:455-462` also reads `?hashSuffix=` and sets `HashPrefix` to its negative, which would silently override whatever `newSubtreeStore` passes. `daemon.GetSubtreeStore` has the same blind spot, so this is parity — but the "mirrors the daemon" claim inherits it.

## Two of the four new tests pass against the un-fixed code

Re-running all four with the fix neutralised (`WithHashPrefix(0)`, equivalent to passing no option per the probe above):

| Test | Pre-fix | Guards the fix? |
|---|---|---|
| `TestResolveStores_SubtreeStoreReadsWhatTheNodeWrote` | **FAIL** | Yes — the real regression test, and a good one |
| `TestNewSubtreeStore_FlatLayoutMissesShardedBlob` | PASS | By design, and its comment says so — fine |
| `TestNewSubtreeStore_HonoursHashPrefixQuery` | PASS | No |
| `TestNewSubtreeStore_RejectsBadHashPrefix` | PASS | No |

`HonoursHashPrefixQuery` is green either way for the reason above — the file store reads `?hashPrefix=4` itself, so the option is never load-bearing there, and the failure message ("hashPrefix from the URL query must reach the store as an option") asserts something the test can't observe. Suggest testing `newSubtreeStore`'s parse directly rather than store behaviour.

`RejectsBadHashPrefix` is green pre-fix because `file.newStore` also rejects `banana`, with a message that happens to contain the substring `hashPrefix` the assertion greps for. Tightening it to `errors.ErrConfiguration` or the distinctive `subtreestore hashPrefix config error` text would pin the tool's own parse.

## Smaller things

- The query now gets parsed twice by two different parsers raising two different error types (`Atoi` → `ConfigurationError` here, `ParseInt` → `StorageError` in the file store). They agree on every input today; worth a note so nobody "simplifies" the tool-side parse away later.
- `TestResolveStores_...` gets `ownedByUs == true` — the signal that the caller owns closing — and never closes the three stores. Harmless at this scale, but a `t.Cleanup` would match the contract.
- The `phase3_finalize.go` note says "tracked separately" with no issue number, so it will rot. I checked that finding rather than take it on trust and it's accurate: the persister writes `<blockstore>/lastProcessed.dat`, the tool deletes `<subtreestore>/lastProcessed.dat.dat`. Leaving it out of scope is the right call.

## Confirmed good

The claim that `WithNoHashPrefix()` leaves that `Del` byte-identical holds up — `options.MergeOptions` (`stores/blob/options/options.go:241-259`) copies the store's `HashPrefix` first and applies per-call file options after, and the file store calls `MergeOptions` on every operation.

The nil-URL guard is a genuine improvement, not just tidiness: `blob.NewStore` dereferences `storeURL.Scheme` before any nil check, so a missing `subtreestore` setting used to panic. And the tool has exactly two subtree-store call sites (`phase2_blocks.go:146`, `phase3_finalize.go:172`), both accounted for — introducing a store-level hash prefix is the kind of change that quietly breaks unrelated filename-keyed calls, and this one demonstrably doesn't.

`go vet`, `go test`, `go test -race` and `golangci-lint` all clean on `./cmd/rewindblockchain/...`. I did not run the smoke or sequential suites, and did not exercise the tool against a real `file://` deployment.

PR #1358, opened 2026-07-28
chore: SonarQube shell test operator (S7688) + remaining S1192 constants

liam:
Two small SonarQube cleanups in one PR.

## 1. `shelldre:S7688` (reliability, HIGH) - shell test operator

`scripts/check_error_wrap_verbs.sh:37` used `[ -n "$matches" ]`. Switched to `[[ -n "$matches" ]]` - the safer, more feature-rich construct. The script is `#!/usr/bin/env bash`, so `[[` is available; `bash -n` passes.

## 2. `go:S1192` - the 7 files held back from #1331

The main S1192 sweep (#1331) deliberately excluded 7 files whose literal-substitutions fall *inside pre-existing duplicated blocks*, because touching those lines inflates the `new_duplicated_lines_density` metric. Since that metric is **not a required gate**, those files are included here:

- `stores/txmetacache/improved_cache.go`
- `util/grpc_helper.go`
- `services/blockassembly/BlockAssembler.go`
- `daemon/daemon_stores.go`
- `stores/blockchain/sql/blob_deletion.go`
- `services/utxopersister/UTXOSet.go`
- `services/blockassembly/subtreeprocessor/SubtreeProcessor.go`

Each duplicated literal is now an unexported package-level constant, byte-identical to the literal it replaces. Only exact standalone tokens were replaced (prefixed/suffixed variants left untouched). Purely mechanical, no logic, no `_test.go`, no generated code.

**Note:** SonarQube's `new_duplicated_lines_density` condition will likely go red on this PR (the 7 files' edits sit in pre-existing duplicated blocks - this is not new duplication introduced by the change). That condition is not a required check.

## Verification

- `bash -n` on the shell script - clean
- `go build` across all 7 affected trees + `go vet` - clean
- `gofmt -l` on all changed files - clean
- Test files in affected packages compile; dedup spot-checked (each literal now appears once)

### DIFF
--- a/services/blockassembly/subtreeprocessor/SubtreeProcessor.go
+++ b/services/blockassembly/subtreeprocessor/SubtreeProcessor.go
@@ -57,9 +57,9 @@ import (
 // workers; 16K buckets at roughly the same total footprint reduces collision
 // probability ~4×. Exported to package-internal callers (tests) so they can
 // construct fresh maps without re-deriving the value.
-const errAddingNodeToSubtree = "error adding node to subtree"
-
 const splitMapBuckets = 16 * 1024
+
+const errAddingNodeToSubtree = "error adding node to subtree"
 
 // splitMapBucketsMax is the inclusive upper bound for
 // BlockAssembly.SplitMapBuckets — driven by the uint16 internal
--- a/services/blockassembly/subtreeprocessor/SubtreeProcessor.go
+++ b/services/blockassembly/subtreeprocessor/SubtreeProcessor.go
@@ -57,6 +57,8 @@ import (
 // workers; 16K buckets at roughly the same total footprint reduces collision
 // probability ~4×. Exported to package-internal callers (tests) so they can
 // construct fresh maps without re-deriving the value.
+const errAddingNodeToSubtree = "error adding node to subtree"
+
 const splitMapBuckets = 16 * 1024
 
 // splitMapBucketsMax is the inclusive upper bound for
@@ -2110,7 +2112,7 @@ func (stp *SubtreeProcessor) addNode(node subtreepkg.Node, parents *subtreepkg.T
 	}
 
 	if err = stp.currentSubtree.Load().AddSubtreeNode(node); err != nil {
-		return errors.NewProcessingError("error adding node to subtree", err)
+		return errors.NewProcessingError(errAddingNodeToSubtree, err)
 	}
 
 	if stp.currentSubtree.Load().IsComplete() {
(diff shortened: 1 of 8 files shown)


ctnguyen (approved):
Review at commit [e6aa0ac](https://github.com/liam/teranode/commit/e6aa0acc497948d1465621719225bc8a4538145f)

About as low-risk as a cleanup PR gets, and the description does the reviewer a favour: splitting these 7 files out of #1331 was the right call, and flagging up front that `new_duplicated_lines_density` will go red — with the reason — beats letting someone find it on the checks tab. I walked each constant against its call sites: values are byte-identical (verbs and the trailing space in `"chunks: "` included), each literal now appears exactly once per file, no name collides in its package, and the existing string assertions in `util/grpc_helper_test.go` still hold. The `[` → `[[` swap is safe under the `bash` shebang.

services/blockassembly/subtreeprocessor/SubtreeProcessor.go
```diff
@@ -57,6 +57,8 @@ import (
 // workers; 16K buckets at roughly the same total footprint reduces collision
 // probability ~4×. Exported to package-internal callers (tests) so they can
 // construct fresh maps without re-deriving the value.
+const errAddingNodeToSubtree = "error adding node to subtree"
+
 const splitMapBuckets = 16 * 1024
```
ordishs: This insertion point splits the nine-line doc comment above (which begins "splitMapBuckets is the fallback bucket count...") from the constant it documents. As written, godoc now attaches that tuning rationale to `errAddingNodeToSubtree`, and `splitMapBuckets` is left undocumented. Moving the new constant below `splitMapBuckets` reattaches the comment:

```suggestion
const splitMapBuckets = 16 * 1024

const errAddingNodeToSubtree = "error adding node to subtree"
```

ordishs (changes requested):
One small placement issue, otherwise this is a faithful mechanical cleanup — every constant is byte-identical to the literal it replaces, the `NewInvalidArgumentError` vs `NewProcessingError` distinction in `improved_cache.go` is preserved, and `msgChunks` keeps its trailing space so output is unchanged. Happy to approve once the constant below is relocated.

liam:
Thanks @ordishs - fixed in 3d1e1484c. Moved `errAddingNodeToSubtree` below `splitMapBuckets` so the nine-line tuning-rationale comment reattaches to `splitMapBuckets`. Also scanned the other 6 files for the same detachment pattern - the rest either went into existing const blocks or sit cleanly after the import block, so this was the only spot. Build + gofmt clean.

ordishs (approved):
The relocation in 3d1e1484c is exactly right — the nine-line tuning comment reattaches to `splitMapBuckets`, and the blank line keeps `splitMapBucketsMax`'s doc comment correctly associated. Verified the package builds and gofmt is clean at the new head. All other substitutions were already confirmed byte-identical in the first pass. Approved.

PR #1360, opened 2026-07-28
Merge feat/teranode-native-ops into main

ordishs:
## Summary

Merges `feat/teranode-native-ops` into `main`. The branch is 134 commits ahead, but nearly all of that content already landed in `main` via separately squash-merged PRs. The net delta is **33 files, ~2,200 insertions / ~310 deletions**, scoped almost entirely to the Aerospike UTXO store:

- `stores/utxo/aerospike/native_op.go` (new) — native Aerospike operate-path replacing UDF/Lua calls, with msgpack encode/parse and routing in `teranode.go`, `spend.go`, `set_mined.go`, `un_spend.go`, `locked.go`, `conflicting.go`
- Pruner: parent updates and `addDeletedChildren` routed through the native-op builder
- New `batch_diagnostics.go` and reworked `alert_system.go`
- Roughly half the delta is new tests (native-op wire contract, unspend fencing, pruner combined-cleanup, batcher shutdown races)
- `go.mod`: adds `vmihailenco/msgpack/v5` only — the BSV aerospike-client-go fork bump is already in `main`

## Conflict resolution

One conflict, in `settings/settings.go`: the branch adds a `MaxOrphanTxs` wiring line that `main` already has from a separate PR. Resolved by keeping `main`'s side.

## Verification

On the merged tree:
- `go build ./...` — clean
- `go vet` on touched packages — clean
- `go test ./settings/... ./stores/utxo/aerospike/...` — all pass, including the full TestContainers aerospike suite and pruner tests

Full CI on this PR should be the remaining gate.

PR #1364, opened 2026-07-28
test: harden subtree-meta fixture and catchup mock expectations

ordishs:
Extracted from `feat/teranode-native-ops` (PR #1360). #1361–#1363 have merged, so this now stands alone on `main`.

## Summary

Hardens the subtree-meta test fixtures in `model/` so they drive the real parent-validation path, and documents two mock/fixture traps. Test files plus one comment-only change in `stores/utxo/aerospike/aerospike.go`.

### `model/Block_test.go`

**The fixture produced metadata validation always rejected.** `createValidSubtreeMetadata` gave every non-coinbase node an empty `TxInpoints`. That serializes fine — `Serialize` only rejects a *nil* `ParentTxHashes`, and `NewTxInpoints` returns a non-nil empty slice — but it round-trips back nil, because `TxInpoints.deserializeFromReader` returns early on `parentCount == 0`. `validateTransaction` then rejected the first non-coinbase node with "could not be found in tx meta data", so no parent logic ever ran.

**Each node now gets a distinct inpoint** — one shared dummy parent, per-node vout. `Inpoint` identity is `{Hash, Index}`, so varying the vout is sufficient and leaves only one parent to seed. A single shared inpoint would trip `checkDuplicateInputs` and reject node 1 onward with "has duplicate inputs", which is why the "should iterate through all 4 nodes" subtest previously stopped at node 1.

`seedParent` / `seedDummyParent` store a parent as mined into a given block ID so tests that expect a clean validation resolve it; leaving it unseeded is how a test asks for the missing-parent path.

**`nodeIndex` fix.** `parallel_processing_with_multiple_missing_parents` and `oldBlockIDsMap_population` passed `nodeIndex: 1` to a *single-node* subtree. `createSubtreeMetadataWithParents` only writes indexes below `subtree.Length()`, so index 1 was never reached and the supplied parent hashes were silently dropped — neither subtest exercised what its name claims. Both now target node 0.

**Assertions replace `_ = err`.** Subtests whose names claim an outcome now assert it, and the assertions are strong enough to hold the invariants rather than just the error class:

- the multiple-parents case requires the error to name one of the *supplied* parents and not the fixture's fallback;
- the `oldBlockIDsMap` case seeds a second parent at a distinct block ID, so the recorded IDs distinguish it from the fallback.

Each of the four regressions — empty `TxInpoints`, shared inpoint, and either wrong `nodeIndex` — was reintroduced and confirmed to turn the suite red.

Also folded the same dummy-parent fallback into `createSubtreeMetadataWithParents`, whose non-target nodes had the identical latent bail, and dropped the unused exported `CreateValidSubtreeMetadata`, which documented the opposite rule.

### `services/blockvalidation/catchup_fork_handling_test.go`

Documents why the mock must return `ErrBlockNotFound` for fork blocks in `TestCatchup_NoMinedSetChurnOnRejectedFork`: `findCommonAncestor` conveys existence through `GetBlockHeader`, so a permissive catch-all would make the peer's whole chain look already-present, leaving zero offered headers for the secret-mining check.

### `stores/utxo/aerospike/aerospike.go`

Comment-only: records why the batcher shutdown guard returns `ErrServiceUnavailable` rather than `ErrUnknown` — only the former is inside `errors.IsTransientLocalError`, so `ErrUnknown` would make netsync reject the block as invalid and disconnect the delivering peer for a fault that is ours.

### DIFF
--- a/model/Block_test.go
+++ b/model/Block_test.go
@@ -3900,19 +3900,34 @@ func (m *mockSubtreeStore) GetIoReader(ctx context.Context, key []byte, fileType
 	return nil, errors.NewBlobNotFoundError("mock error")
 }
 
-// createValidSubtreeMetadata creates valid subtree metadata that won't trigger retries
+// createValidSubtreeMetadata creates valid subtree metadata that won't trigger retries.
+// Non-coinbase nodes get a dummy parent because go-subtree v1.4.0's SubtreeMeta.Serialize
+// rejects empty TxInpoints on non-coinbase entries ("parent tx hashes are not set for node N").
+// The coinbase placeholder at index 0 keeps an empty TxInpoints, matching the on-wire shape.
 func createValidSubtreeMetadata(subtree *subtreepkg.Subtree) ([]byte, error) {
-	// Create SubtreeMeta with proper structure
 	subtreeMeta := subtreepkg.NewSubtreeMeta(subtree)
 
-	// Initialize TxInpoints array for all nodes up to Length()
-	for i := 0; i < subtree.Length(); i++ {
-		// Create empty TxInpoints for all nodes (including root)
-		txInpoints := subtreepkg.NewTxInpoints()
-		subtreeMeta.TxInpoints[i] = txInpoints
+	var dummyParent chainhash.Hash
+	dummyParent[0] = 0xde
+
+	dummyInput := &bt.Input{PreviousTxOutIndex: 0}
+	if err := dummyInput.PreviousTxIDAdd(&dummyParent); err != nil {
+		return nil, err
 	}
 
-	// Serialize the metadata
+	dummyInpoints, err := subtreepkg.NewTxInpointsFromInputs([]*bt.Input{dummyInput})
+	if err != nil {
+		return nil, err
+	}
+
+	for i := 0; i < subtree.Length(); i++ {
+		if i == 0 && len(subtree.Nodes) > 0 && subtree.Nodes[0].Hash == *subtreepkg.CoinbasePlaceholderHash {
+			subtreeMeta.TxInpoints[i] = subtreepkg.NewTxInpoints()
+			continue
+		}
+		subtreeMeta.TxInpoints[i] = dummyInpoints
+	}
+
 	return subtreeMeta.Serialize()
 }
 
--- a/model/Block_test.go
+++ b/model/Block_test.go
@@ -4514,14 +4514,26 @@ func TestBlock_ValidateSubtree_NodeIteration(t *testing.T) {
 
 		defer func() { block.txMap = txmap.NewSplitSwissMapUint64(10) }()
 
-		// Create subtree metadata with parent tx hashes to trigger missing parent logic
-		subtreeMetaSlice, err := createSubtreeMetadataWithParents(subtree, 1, []chainhash.Hash{*tx2Hash})
+		// Node 0 (tx1Hash) spends tx3Hash, which is in neither txMap nor the store, so it
+		// becomes a missing parent. Node 1 (tx2Hash) takes the dummy-parent fallback, which
+		// is seeded below and resolves. Both nodes therefore contribute an entry to
+		// checkParentTxHashes and the group fans out over two parents with mixed outcomes —
+		// the accumulation across nodes that the single-node sibling subtest cannot reach.
+		//
+		// Do not attach a node's own hash here: a same-block parent whose txMap index is not
+		// greater than the child's takes the `continue` branch in checkParentTransactions,
+		// contributing nothing, which leaves the fan-out empty and validateSubtree returning
+		// nil no matter how the store is seeded.
+		subtreeMetaSlice, err := createSubtreeMetadataWithParents(subtree, 0, []chainhash.Hash{*tx3Hash})
 		require.NoError(t, err)
 
 		mockStore.data[string(subtree.RootHash()[:])] = subtreeMetaSlice
 
+		utxoStore := createTestUTXOStore(t)
+		seedDummyParent(t, utxoStore)
+
 		deps := &validationDependencies{
-			txMetaStore:           createTestUTXOStore(t),
+			txMetaStore:           utxoStore,
 			subtreeStore:          mockStore,
 			currentChain:          []*BlockHeader{},
 			currentBlockHeaderIDs: []uint32{1, 2},
@@ -4534,12 +4546,16 @@ func TestBlock_ValidateSubtree_NodeIteration(t *testing.T) {
 			parentSpendsMap:             NewSplitSyncedParentMap(4),
 		}
 
-		// This should trigger the parallel parent checking logic:
+		// Exercises the parallel parent checking logic:
 		// if len(checkParentTxHashes) > 0 { ... parentG.Go(...) ... }
 		err = block.validateSubtree(ctx, ulogger.TestLogger{}, deps, validationCtx, subtree, 0)
+		require.ErrorIs(t, err, errors.ErrBlockIncomplete)
 
-		// Test exercises the parallel parent processing logic
-		_ = err // May succeed or fail but exercises the parallel processing
+		// tx3Hash is the parent that must be reported missing; the seeded fallback must not
+		// be, which is what proves the fan-out actually had a resolving member rather than
+		// bailing on node 0 alone.
+		require.Contains(t, err.Error(), tx3Hash.String())
+		require.NotContains(t, err.Error(), dummyParentHash.String())
 	})
 
 	t.Run("single node subtree", func(t *testing.T) {
--- a/model/Block_test.go
+++ b/model/Block_test.go
@@ -3900,43 +3900,96 @@ func (m *mockSubtreeStore) GetIoReader(ctx context.Context, key []byte, fileType
 	return nil, errors.NewBlobNotFoundError("mock error")
 }
 
-// createValidSubtreeMetadata creates valid subtree metadata that won't trigger retries.
-// Non-coinbase nodes get a dummy parent because go-subtree v1.4.0's SubtreeMeta.Serialize
-// rejects empty TxInpoints on non-coinbase entries ("parent tx hashes are not set for node N").
-// The coinbase placeholder at index 0 keeps an empty TxInpoints, matching the on-wire shape.
+// dummyParentTx is the stand-in parent that every non-coinbase node built by
+// createValidSubtreeMetadata / createSubtreeMetadataWithParents spends from.
+// Tests that expect validateSubtree to succeed must seed it via seedDummyParent;
+// leaving it unseeded is how a test asks for the missing-parent path.
+var (
+	dummyParentTx   = newTx(99)
+	dummyParentHash = *dummyParentTx.TxIDChainHash()
+)
+
+// seedDummyParent stores dummyParentTx as mined into block ID 1, so
+// getParentTxMetaBlockIDs resolves it instead of returning BlockIncompleteError.
+// WithCreateOnly is required because dummyParentTx's own input is not in the
+// store; SpendAndCreate is used because utxo.Store exposes it but not Create.
+//
+// BlockID 1 is error-free for both map shapes these tests use: when
+// currentBlockHeaderIDsMap contains 1 the parent is found exactly once, and when
+// the map is empty the minBlockID > 0 branch defers to the validator — neither
+// reaches ErrCheckParentExistsOnChain.
+func seedDummyParent(t *testing.T, store utxo.Store) {
+	t.Helper()
+
+	_, _, err := store.SpendAndCreate(context.Background(), dummyParentTx, 1,
+		utxo.WithMinedBlockInfo(utxo.MinedBlockInfo{BlockID: 1, BlockHeight: 1}),
+		utxo.WithCreateOnly())
+	require.NoError(t, err)
+}
+
+// dummyInpointsForNode builds a single-input TxInpoints spending dummyParentHash
+// at vout idx. Inpoint identity is {Hash, Index}, so varying the vout is what
+// keeps each node in a fixture distinct.
+func dummyInpointsForNode(idx int) (subtreepkg.TxInpoints, error) {
+	input := &bt.Input{PreviousTxOutIndex: uint32(idx)} // nolint:gosec
+	if err := input.PreviousTxIDAdd(&dummyParentHash); err != nil {
+		return subtreepkg.TxInpoints{}, err
+	}
+
+	return subtreepkg.NewTxInpointsFromInputs([]*bt.Input{input})
+}
+
+// createValidSubtreeMetadata creates subtree metadata that survives the whole
+// validateSubtree path. Two constraints shape it:
+//
+//   - Every non-coinbase node needs at least one parent inpoint. A zero-parent
+//     entry serializes fine — Serialize only rejects a nil ParentTxHashes, and
+//     NewTxInpoints returns a non-nil empty slice — but it round-trips back nil,
+//     because TxInpoints.deserializeFromReader returns early on parentCount == 0.
+//     validateTransaction then rejects the node with "could not be found in tx
+//     meta data" before any parent logic runs.
+//   - Each node's inpoint must be distinct. checkDuplicateInputs pushes every
+//     inpoint through parentSpendsMap.SetIfNotExists, so reusing one inpoint
+//     makes every node after the first fail with "has duplicate inputs".
+//
+// The coinbase placeholder at index 0 keeps an empty TxInpoints because
+// validateSubtree skips that node. Callers that want validation to succeed must
+// also seed the parent — see seedDummyParent.
 func createValidSubtreeMetadata(subtree *subtreepkg.Subtree) ([]byte, error) {
 	subtreeMeta := subtreepkg.NewSubtreeMeta(subtree)
 
-	var dummyParent chainhash.Hash
-	dummyParent[0] = 0xde
-
-	dummyInput := &bt.Input{PreviousTxOutIndex: 0}
-	if err := dummyInput.PreviousTxIDAdd(&dummyParent); err != nil {
-		return nil, err
-	}
-
-	dummyInpoints, err := subtreepkg.NewTxInpointsFromInputs([]*bt.Input{dummyInput})
-	if err != nil {
-		return nil, err
-	}
-
 	for i := 0; i < subtree.Length(); i++ {
-		if i == 0 && len(subtree.Nodes) > 0 && subtree.Nodes[0].Hash == *subtreepkg.CoinbasePlaceholderHash {
+		if i == 0 && subtree.Nodes[0].Hash.Equal(subtreepkg.CoinbasePlaceholderHashValue) {
 			subtreeMeta.TxInpoints[i] = subtreepkg.NewTxInpoints()
 			continue
 		}
-		subtreeMeta.TxInpoints[i] = dummyInpoints
+
+		txInpoints, err := dummyInpointsForNode(i)
+		if err != nil {
+			return nil, err
+		}
+
+		subtreeMeta.TxInpoints[i] = txInpoints
 	}
 
 	return subtreeMeta.Serialize()
 }
 
-// createSubtreeMetadataWithParents creates subtree metadata with parent tx hashes
+// createSubtreeMetadataWithParents creates subtree metadata where nodeIndex spends
+// parentHashes. Every other non-coinbase node falls back to the shared dummy
+// parent rather than an empty TxInpoints — an empty entry round-trips back nil and
+// validateTransaction would reject that node before reaching nodeIndex at all. See
+// createValidSubtreeMetadata for the full reasoning.
 func createSubtreeMetadataWithParents(subtree *subtreepkg.Subtree, nodeIndex int, parentHashes []chainhash.Hash) ([]byte, error) {
 	subtreeMeta := subtreepkg.NewSubtreeMeta(subtree)
 
 	// Initialize TxInpoints for all nodes
 	for i := 0; i < subtree.Length(); i++ {
+		if i == 0 && subtree.Nodes[0].Hash.Equal(subtreepkg.CoinbasePlaceholderHashValue) {
+			subtreeMeta.TxInpoints[i] = subtreepkg.NewTxInpoints()
+			continue
+		}
+
 		// Add parent hashes to specific node
 		if i == nodeIndex && len(parentHashes) > 0 {
 			// Build mock inputs — vout 0 for each parent hash.
@@ -3956,9 +4009,16 @@ func createSubtreeMetadataWithParents(subtree *subtreepkg.Subtree, nodeIndex int
 			}
 
 			subtreeMeta.TxInpoints[i] = txInpoints
-		} else {
-			subtreeMeta.TxInpoints[i] = subtreepkg.NewTxInpoints()
+
+			continue
 		}
+
+		txInpoints, err := dummyInpointsForNode(i)
+		if err != nil {
+			return nil, err
+		}
+
+		subtreeMeta.TxInpoints[i] = txInpoints
 	}
 
 	return subtreeMeta.Serialize()
@@ -4071,9 +4131,10 @@ func TestBlock_ValidateSubtree_ComprehensiveCoverage(t *testing.T) {
 			parentSpendsMap:             NewSplitSyncedParentMap(4),
 		}
 
-		// Test - should pass since coinbase is skipped
+		// The only node is the coinbase placeholder, which validateSubtree skips, so
+		// there is nothing left to validate and no parent to resolve.
 		err = block.validateSubtree(context.Background(), ulogger.TestLogger{}, deps, validationCtx, subtree, 0)
-		_ = err // May error but exercises the coinbase skip logic
+		require.NoError(t, err)
 	})
 }
 
@@ -4134,18 +4195,31 @@ func TestBlock_ValidateSubtree_MissingParents(t *testing.T) {
 		err = subtree.AddNode(*txHash, 1, 100)
 		require.NoError(t, err)
 
+		// validateTransaction rejects any node absent from txMap before it looks at
+		// parents, so the node has to be registered to reach the parent lookup.
+		err = block.txMap.Put(*txHash, 1)
+		require.NoError(t, err)
+
+		defer func() { block.txMap = txmap.NewSplitSwissMapUint64(10) }()
+
 		// Create valid subtree metadata that won't trigger retries
 		subtreeMetaSlice, err := createValidSubtreeMetadata(subtree)
 		require.NoError(t, err)
 
 		mockStore.data[string(subtree.RootHash()[:])] = subtreeMetaSlice
 
-		// This should not error but should handle the missing parent gracefully
-		err = block.validateSubtree(ctx, ulogger.TestLogger{}, deps, validationCtx, subtree, 0)
+		// Fresh parent-spends map: production builds one per block validation, and
+		// sharing it across subtests makes the identical dummy inpoint collide.
+		subtestCtx := &validationContext{
+			currentBlockHeaderHashesMap: make(map[chainhash.Hash]struct{}),
+			currentBlockHeaderIDsMap:    map[uint32]struct{}{1: {}, 2: {}},
+			parentSpendsMap:             NewSplitSyncedParentMap(4),
+		}
 
-		// The error handling depends on the actual implementation
-		// This test verifies the code path is exercised
-		_ = err // May or may not error depending on implementation details
+		// The dummy parent is deliberately not seeded, so it is absent from the store.
+		// getParentTxMetaBlockIDs classifies that as retryable rather than invalid.
+		err = block.validateSubtree(ctx, ulogger.TestLogger{}, deps, subtestCtx, subtree, 0)
+		require.ErrorIs(t, err, errors.ErrBlockIncomplete)
 	})
 
 	t.Run("parent transaction ordering error", func(t *testing.T) {
@@ -4236,16 +4310,37 @@ func TestBlock_ValidateSubtree_MissingParents(t *testing.T) {
 		err = subtree.AddNode(*txHash, 1, 100)
 		require.NoError(t, err)
 
+		err = block.txMap.Put(*txHash, 1)
+		require.NoError(t, err)
+
+		defer func() { block.txMap = txmap.NewSplitSwissMapUint64(10) }()
+
 		// Create valid subtree metadata that won't trigger retries
 		subtreeMetaSlice, err := createValidSubtreeMetadata(subtree)
 		require.NoError(t, err)
 
 		mockStore.data[string(subtree.RootHash()[:])] = subtreeMetaSlice
 
-		err = block.validateSubtree(ctx, ulogger.TestLogger{}, deps, validationCtx, subtree, 0)
+		// Seeding the dummy parent is what makes this the "no missing parents" case:
+		// every node's parent resolves, so validateSubtree completes cleanly.
+		seededStore := createTestUTXOStore(t)
+		seedDummyParent(t, seededStore)
 
-		// Should handle empty missing parents gracefully
-		_ = err // No missing parents to process
+		seededDeps := &validationDependencies{
+			txMetaStore:           seededStore,
+			subtreeStore:          mockStore,
+			currentBlockHeaderIDs: []uint32{1, 2},
+			oldBlockIDsMap:        txmap.NewSyncedMap[chainhash.Hash, []uint32](),
+		}
+
+		subtestCtx := &validationContext{
+			currentBlockHeaderHashesMap: make(map[chainhash.Hash]struct{}),
+			currentBlockHeaderIDsMap:    map[uint32]struct{}{1: {}, 2: {}},
+			parentSpendsMap:             NewSplitSyncedParentMap(4),
+		}
+
+		err = block.validateSubtree(ctx, ulogger.TestLogger{}, seededDeps, subtestCtx, subtree, 0)
+		require.NoError(t, err)
 	})
 
 	t.Run("multiple invalid parents", func(t *testing.T) {
@@ -4259,16 +4354,29 @@ func TestBlock_ValidateSubtree_MissingParents(t *testing.T) {
 		err = subtree.AddNode(*parentHash, 2, 100)
 		require.NoError(t, err)
 
+		err = block.txMap.Put(*txHash, 1)
+		require.NoError(t, err)
+		err = block.txMap.Put(*parentHash, 2)
+		require.NoError(t, err)
+
+		defer func() { block.txMap = txmap.NewSplitSwissMapUint64(10) }()
+
 		// Create valid subtree metadata that won't trigger retries
 		subtreeMetaSlice, err := createValidSubtreeMetadata(subtree)
 		require.NoError(t, err)
 
 		mockStore.data[string(subtree.RootHash()[:])] = subtreeMetaSlice
 
-		err = block.validateSubtree(ctx, ulogger.TestLogger{}, deps, validationCtx, subtree, 0)
+		subtestCtx := &validationContext{
+			currentBlockHeaderHashesMap: make(map[chainhash.Hash]struct{}),
+			currentBlockHeaderIDsMap:    map[uint32]struct{}{1: {}, 2: {}},
+			parentSpendsMap:             NewSplitSyncedParentMap(4),
+		}
 
-		// Should handle multiple parent validation errors
-		_ = err // Multiple validation paths
+		// Both nodes carry distinct inpoints, so they clear checkDuplicateInputs and
+		// both reach the parent lookup — where the unseeded parent is reported missing.
+		err = block.validateSubtree(ctx, ulogger.TestLogger{}, deps, subtestCtx, subtree, 0)
+		require.ErrorIs(t, err, errors.ErrBlockIncomplete)
 	})
 }
 
@@ -4345,8 +4453,11 @@ func TestBlock_ValidateSubtree_NodeIteration(t *testing.T) {
 
 		mockStore.data[string(subtree.RootHash()[:])] = subtreeMetaSlice
 
+		utxoStore := createTestUTXOStore(t)
+		seedDummyParent(t, utxoStore)
+
 		deps := &validationDependencies{
-			txMetaStore:           createTestUTXOStore(t),
+			txMetaStore:           utxoStore,
 			subtreeStore:          mockStore,
 			currentChain:          []*BlockHeader{},
 			currentBlockHeaderIDs: []uint32{},
@@ -4359,11 +4470,11 @@ func TestBlock_ValidateSubtree_NodeIteration(t *testing.T) {
 			parentSpendsMap:             NewSplitSyncedParentMap(4),
 		}
 
-		// This should iterate through all 4 nodes in the subtree
+		// All 4 nodes must validate. A shared inpoint across nodes would stop this at
+		// node 1 with "has duplicate inputs", so a clean result is what proves the
+		// iteration actually reached every node.
 		err = block.validateSubtree(ctx, ulogger.TestLogger{}, deps, validationCtx, subtree, 0)
-
-		// Test exercises the node iteration logic: subtreeNode := subtree.Nodes[snIdx]
-		_ = err // May succeed or fail but exercises the iteration
+		require.NoError(t, err)
 	})
 
 	t.Run("subtree with missing parents parallel processing", func(t *testing.T) {
@@ -4435,8 +4546,11 @@ func TestBlock_ValidateSubtree_NodeIteration(t *testing.T) {
 
 		mockStore.data[string(subtree.RootHash()[:])] = subtreeMetaSlice
 
+		utxoStore := createTestUTXOStore(t)
+		seedDummyParent(t, utxoStore)
+
 		deps := &validationDependencies{
-			txMetaStore:           createTestUTXOStore(t),
+			txMetaStore:           utxoStore,
 			subtreeStore:          mockStore,
 			currentChain:          []*BlockHeader{},
 			currentBlockHeaderIDs: []uint32{},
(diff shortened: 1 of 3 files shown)


ctnguyen (changes requested):
Review at commit [b3a0169](https://github.com/ordishs/teranode/commit/b3a01694158b9c7c9d161f3af8bde695ebd6b266)

Good stack hygiene here — the net change really is just the two test files, the description says plainly which commits drop out once #1361–#1363 land, and that makes this an easy one to read. The catchup half is the stronger of the two: the comment is accurate (`findCommonAncestor` does signal existence through `GetBlockHeader`, `services/blockvalidation/catchup.go:531-541`) and it names the exact trap — a permissive catch-all leaves zero offered headers for the secret-mining check — that the next person editing that mock would otherwise walk into. The subtree fixture change also identifies something real: the old empty-`TxInpoints` shape genuinely was broken. It just doesn't finish the job — block validation still rejects the metadata, for two new reasons instead of the old one, and the tests using it discard the result so nothing goes red either way.

## Blocking issues

### ChiR1 - Fixture still can't drive a successful validation
**Problem:** Before this change, every non-coinbase node serialized with zero parents, came back from `NewSubtreeMetaFromReader` with a nil `ParentTxHashes` (`inpoints.go:334-336` returns early on `parentCount == 0`; `model/Block.go:1684`), and `validateTransaction` rejected the *first* non-coinbase node with `BlockInvalidError "could not be found in tx meta data"` (`model/Block.go:1360-1363`). The new fixture (`model/Block_test.go:3910-3928`) clears that, and then hits two new walls:
1. *Duplicate inputs.* Every non-coinbase node is assigned the same `dummyInpoints` value, so each reports the identical inpoint `{0xde…, 0}`. `checkDuplicateInputs` pushes every inpoint through `parentSpendsMap.SetIfNotExists` (`model/Block.go:1382-1396`, `model/map.go:48-61`), which returns `false` on a repeat `{Hash, Index}` — so any subtree with two or more non-coinbase nodes now fails at `snIdx == 1` with `BlockInvalidError "has duplicate inputs"`. `TestBlock_ValidateSubtree_NodeIteration/subtree_with_multiple_nodes_iteration` (`model/Block_test.go:4302-4367`), whose comment says "This should iterate through all 4 nodes", gets to node 1 and stops.
2. *Off-chain parent.* `0xde…` is absent from the `sqlitememory` store built by `createTestUTXOStore`, so a single-node subtree — the one case that clears the duplicate check — reaches `checkParentsExistOnChain`, where `getParentTxMetaBlockIDs` maps the missing parent to `BlockIncompleteError`, the retryable classification (`model/Block.go:1454-1459`). That is what `/single_node_subtree` (`model/Block_test.go:4418-4456`) now gets.
**Why it matters:** The PR's purpose is a fixture that validates cleanly, and a helper named `createValidSubtreeMetadata` documented as producing metadata "that won't trigger retries" still produces neither. The shared-inpoint design means it can never validate a multi-node subtree, so the trap stays in place for the next test that puts more than one tx in `txMap`.
**Fix:** Make the inpoints distinct and the parent resolvable. Uniqueness needs only a per-node vout, since `Inpoint` identity is `{Hash, Index}`; deriving the dummy parent from a real throwaway tx (`newTx` already exists in this package, `model/update-tx-mined_test.go:208`) lets tests seed it without `WithTXID`:
```go
// dummyParentTx is the parent every non-coinbase fixture node points at. Tests
// that expect validateSubtree to succeed must seed it via seedDummyParent.
var (
	dummyParentTx   = newTx(99)
	dummyParentHash = *dummyParentTx.TxIDChainHash()
)

for i := 0; i < subtree.Length(); i++ {
	if i == 0 && subtree.Nodes[0].Hash.Equal(subtreepkg.CoinbasePlaceholderHashValue) {
		subtreeMeta.TxInpoints[i] = subtreepkg.NewTxInpoints()
		continue
	}

	// Distinct inpoint per node: one shared inpoint makes checkDuplicateInputs
	// reject every node after the first as a duplicate input.
	dummyInput := &bt.Input{PreviousTxOutIndex: uint32(i)} // nolint:gosec
	if err := dummyInput.PreviousTxIDAdd(&dummyParentHash); err != nil {
		return nil, err
	}

	dummyInpoints, err := subtreepkg.NewTxInpointsFromInputs([]*bt.Input{dummyInput})
	if err != nil {
		return nil, err
	}

	subtreeMeta.TxInpoints[i] = dummyInpoints
}
```
Seeding must go through `SpendAndCreate` — `utxo.Store` exposes that but not `Create` (`stores/utxo/Interface.go:375-447`), and `createTestUTXOStore` returns the interface, so the concrete-store `Create` call used at `model/Block_test.go:1651` won't compile here. `WithCreateOnly` is required because `newTx` has an input:
```go
func seedDummyParent(t *testing.T, store utxo.Store) {
	_, _, err := store.SpendAndCreate(context.Background(), dummyParentTx, 1,
		utxo.WithMinedBlockInfo(utxo.MinedBlockInfo{BlockID: 1, BlockHeight: 1}),
		utxo.WithCreateOnly())
	require.NoError(t, err)
}
```
`BlockID: 1` is error-free for both map shapes these tests use: when `currentBlockHeaderIDsMap` contains 1 the parent is found exactly once; when the map is empty, `minBlockID > 0` takes the "defer to the validator" branch, which returns old block IDs and no error (`model/Block.go:1303-1316`, `1433-1449`). A store-free alternative is to point node `i` at node `i-1`'s hash so `checkParentTransactions` resolves the parent from `txMap`, but node 0 still needs a seeded parent, so it doesn't remove the seeding.

### ChiR2 - Affected subtests discard the validation result
**Problem:** Every subtest exercising this fixture ends in `_ = err` with a comment like "May succeed or fail but exercises the iteration" (`model/Block_test.go:4366`, `4456`, `4076`). That predates the PR, and it is why the old fixture could sit there rejected by `validateTransaction` for however long without anyone noticing — and why the half-fix in ChiR1 looks green now.
**Why it matters:** A fixture-hardening change that cannot fail when the fixture is still invalid has no way to demonstrate it worked, and no barrier against the next regression.
**Fix:** For the subtests whose names or comments assert an outcome, assert it. `TestBlock_ValidateSubtree/successful_validation_with_coinbase_placeholder` (`model/Block_test.go:4021-4076`) can take `require.NoError(t, err)` today — the coinbase skip makes it return nil already. `TestBlock_ValidateSubtree_NodeIteration/subtree_with_multiple_nodes_iteration` and `/single_node_subtree` should take it once ChiR1 lands; that pair is the actual proof the fixture is valid. Where an error is genuinely the expected outcome, pin it (`require.ErrorIs(t, err, errors.ErrBlockInvalid)`) instead of dropping it.

## Non-blocking issues

### ChiR3 - Justification comment names a failure the pinned go-subtree can't produce
**Problem:** The new doc comment (and the PR description) says go-subtree v1.4.0's `SubtreeMeta.Serialize` "rejects empty `TxInpoints` on non-coinbase entries". `go.mod:17` pins v1.4.2, and there `Serialize` guards on `s.TxInpoints[i].ParentTxHashes == nil` (`subtree_meta.go:175`), while `NewTxInpoints()` deliberately returns a non-nil-but-empty slice — its own doc says `Meta.Serialize` distinguishes nil from empty precisely so zero-parent non-coinbase nodes stay serializable. The old fixture therefore serialized fine; what rejected it was the nil-after-deserialize path described in ChiR1.
**Why it matters:** The comment exists to stop someone reverting the fixture. Pointing at the wrong library function and a version the repo doesn't pin means the next reader checks `Serialize`, sees empty slices pass, and concludes the comment is obsolete.
**Fix:** Retarget it at the real constraint, e.g.:
```go
// createValidSubtreeMetadata gives every non-coinbase node a distinct parent
// inpoint. A zero-parent entry serializes fine but round-trips back with a nil
// ParentTxHashes (TxInpoints.deserializeFromReader returns early on
// parentCount == 0), and validateTransaction then rejects nil parents with
// "could not be found in tx meta data". The coinbase placeholder at index 0 is
// exempt because validateSubtree skips it.
```

## Recap

| ChiR1 | Fixture still rejected by validation | YES | 60% |
| ChiR2 | Subtests discard validation result | YES | 50% |
| ChiR3 | Comment cites wrong reject path/version |  | 30% |

oskarszoon (changes requested):
Reviewed the top commit `b3a016941` only (the stacked #1361–#1363 content is out of scope here).

The catchup comment is accurate — `findCommonAncestor` does convey existence through `GetBlockHeader` + `ErrBlockNotFound` (`catchup.go:528-540`). Two problems in the subtree-meta fixture.

**`model/Block_test.go:3907` — the shared dummy inpoint makes multi-node fixtures block-invalid.** `dummyInpoints` is built once and assigned to every non-coinbase index, so `checkDuplicateInputs` (`model/Block.go:1387-1406`) trips `SetIfNotExists` on the second non-coinbase node and validation returns `BLOCK_INVALID … has duplicate inputs`. Replicating the 4-node subtest of `TestBlock_ValidateSubtree_NodeIteration` and printing the error `validateSubtree` returns:

```
pre-PR    BLOCK_INVALID    [...]:0 transaction ...a0 could not be found in tx meta data
this PR   BLOCK_INVALID    [...]:1 transaction ...a1 has duplicate inputs
distinct  BLOCK_INCOMPLETE parent transaction ...03de of tx ...a3 not found in txMetaStore
```

So the early bail moves from node 0 to node 1 — the subtest that says "should iterate through all 4 nodes" still doesn't, and it's invisible because that subtest does `_ = err`. Only the third variant (dummy parent varied per index, e.g. `chainhash.Hash{0xde, byte(i), byte(i >> 8)}`) walks all four nodes. Please vary the parent per index; that also removes the slice aliasing you get from copying one `TxInpoints` value into N slots.

**`model/Block_test.go:3903` — the stated rationale doesn't hold for the pinned go-subtree.** `go.mod` pins v1.4.2, where the guard is `i != 0 && s.TxInpoints[i].ParentTxHashes == nil` (`subtree_meta.go:174`) — index 0 exempt unconditionally, not coinbase-specific — and `NewTxInpoints()` returns a non-nil empty slice (`inpoints.go:79`). I measured it: the pre-PR fixture serializes fine (`err=<nil>`, 40/52 bytes for 1/4 leaves), and all three affected tests pass with the pre-PR fixture restored. The actual mechanism is downstream — empty `TxInpoints` deserializes back to a nil `ParentTxHashes`, and `Block.go:1362` rejects that with `could not be found in tx meta data`. Worth correcting in the comment/commit message, because it's also what makes the index-0 coinbase check load-bearing rather than dead code.

Smaller: `createSubtreeMetadataWithParents` (`:3935`) still assigns empty `TxInpoints` to every non-target node, so it has the same latent bail — the two fixtures in one file now disagree on what "valid metadata" means. And `len(subtree.Nodes) > 0` in the index-0 guard is unreachable-false, since the loop bound is `subtree.Length()` == `len(Nodes)`.

Verified: targeted `./model/` and `TestCatchup_NoMinedSetChurnOnRejectedFork` tests pass at HEAD, `go vet` clean on both packages.

oskarszoon:
Re-reviewed at `4e5f46ea9`. The rebase is clean — stack merged, single commit, 3 files. Diffing the three files against the previously reviewed `b3a016941`: `model/Block_test.go` and `catchup_fork_handling_test.go` are byte-identical, and the only new content is the 4-line comment on `recoverBatcherShutdown`. So both P1s stand. This time I ran the PR's own `createValidSubtreeMetadata` (not a replica) against the new base, go-subtree still pinned v1.4.2:

```
PREMISE leaves=1 pre-PR fixture Serialize err=<nil> bytes=40
PREMISE leaves=4 pre-PR fixture Serialize err=<nil> bytes=52

FIXTURE=pre-PR-empty      BLOCK_INVALID    [...]:0 transaction ...a0 could not be found in tx meta data
FIXTURE=PR-shared-parent  BLOCK_INVALID    [...]:1 transaction ...a1 has duplicate inputs
FIXTURE=distinct-parent   BLOCK_INCOMPLETE parent transaction ...03de of tx ...a3 not found in txMetaStore
```

**`model/Block_test.go:3925`** — one `dummyInpoints` value assigned to every non-coinbase index, so `checkDuplicateInputs` (`model/Block.go:1396`) trips at node 1. The early bail moves from node 0 to node 1; `TestBlock_ValidateSubtree_NodeIteration/subtree_with_multiple_nodes_iteration` still doesn't iterate all four nodes, and it's silent because that subtest does `_ = err`. Varying the parent per index (`chainhash.Hash{0xde, byte(i), byte(i >> 8)}`) is the only variant that reaches `...a3`.

**`model/Block_test.go:3902`** — the rationale is still wrong for the pinned version. `Serialize` guards on `ParentTxHashes == nil` with index 0 exempt unconditionally (`subtree_meta.go:174`), and `NewTxInpoints()` returns a non-nil empty slice (`inpoints.go:79`) — the pre-PR fixture serializes fine, measured again above. The real mechanism is the deserialize round-trip producing nil, rejected at `model/Block.go:1361`. That also matters because it's what makes the index-0 coinbase check load-bearing instead of dead code.

New P2 spotted on this pass: `CreateValidSubtreeMetadata` at `model/Block_test.go:4848` — exported, one letter away from the helper you changed — deliberately assigns empty `TxInpoints` "to avoid serialization errors", i.e. asserts the opposite of this PR's comment. Two helpers, contradictory stories. And the fixture at `:5290-5315` already builds a realistic parent mix; that's the pattern worth borrowing.

The new `aerospike.go:143-146` comment is accurate: `ErrServiceUnavailable` is in `IsTransientLocalError` (`errors/error_utils.go:74`) and `handleSpendPanic` does use `NewUnknownError` (`spend.go:335`). No objection — but it makes the PR body's "Test-only change" false, and the stack list there is stale now that #1361–#1363 have merged.

Verified at this head: targeted `./model/` tests, `TestCatchup_NoMinedSetChurnOnRejectedFork`, and `go vet` on all three packages all pass.

ordishs:
Thanks both — all findings addressed at `f7ff4de15`. @oskarszoon your last pass landed against `4e5f46ea9`, just before the first fix push, so it predates the changes below.

Confirming your measurements independently: the pre-PR fixture serializes fine (40/52 bytes for 1/4 leaves), and the shared-parent variant bails at node 1. We agree on the diagnosis.

| Finding | Resolution |
|---|---|
| ChiR1 / `:3925` — shared inpoint, unresolvable parent | Distinct inpoint per node + `seedParent`/`seedDummyParent` |
| ChiR2 — subtests discard the result | `require.NoError` / `require.ErrorIs` at every fixture call site |
| ChiR3 / `:3902` — rationale wrong for pinned v1.4.2 | Comment retargeted at the deserialize round-trip |
| `createSubtreeMetadataWithParents` same latent bail | Non-target nodes now take the dummy-parent fallback |
| `len(subtree.Nodes) > 0` unreachable-false | Removed |
| `CreateValidSubtreeMetadata` contradicts its sibling | Deleted — it had zero callers |
| PR body: "test-only" false, stack list stale | Body rewritten |

Two deliberate deviations from the literal suggestions:

**Vout, not parent hash.** @oskarszoon you suggested varying the parent per index (`{0xde, byte(i), byte(i>>8)}`). I varied the **vout** off one shared parent instead. `Inpoint` identity is `{Hash, Index}`, so uniqueness is equivalent, but it leaves a single parent to seed rather than N — and it removes the slice aliasing you flagged either way.

**`ErrBlockIncomplete`, not `ErrBlockInvalid`.** @ctnguyen ChiR2 suggested pinning error cases with `ErrBlockInvalid`. A missing parent is classified retryable, not invalid (`getParentTxMetaBlockIDs`, #1031), so the honest assertion for the unseeded-parent subtests is `ErrBlockIncomplete`.

**Declined:** borrowing the realistic parent-mix pattern from the fixture at `:5290-5315` into the shared helper. Two reasons. It isn't a coverage gap — every block in `checkParentTransactions` (`Block.go:1411-1430`, including the `foundInSameBlock` ordering check and its error return) is already covered via that test. And pointing node *i* at node *i−1* only works when the caller has put its nodes into `txMap` in ascending order; the inline fixture builds `allTxHashes` deliberately, but several of the six `createValidSubtreeMetadata` callers don't, and they'd start tripping the ordering error instead of testing what they're named for. Happy to revisit if you disagree.

**Also fixed, found while here:** `parallel_processing_with_multiple_missing_parents` and `oldBlockIDsMap_population` passed `nodeIndex: 1` to a *single-node* subtree, so `createSubtreeMetadataWithParents` never reached that index and the supplied parent hashes were silently dropped — neither subtest exercised its name. Both now target node 0, with assertions strong enough to hold it: the first requires the error to name one of the supplied parents and not the fallback; the second seeds a second parent at a distinct block ID so the recorded IDs tell them apart.

On ChiR2's "no barrier against the next regression" — each of the four regressions was reintroduced and confirmed to turn the suite red:

```
empty TxInpoints   -> BLOCK_INVALID ...:0 could not be found in tx meta data  (4 subtests fail)
shared inpoint     -> BLOCK_INVALID ...:1 has duplicate inputs                (2 subtests fail)
nodeIndex 1 (multi)-> error names the fallback parent, not tx2/tx3            (1 subtest fails)
nodeIndex 1 (oldID)-> oldBlockIDsMap = [1], expected [5]                      (1 subtest fails)
```

Verified at `f7ff4de15`: `go test -race ./model/` and `go test ./services/blockvalidation/` pass, `go vet` clean on both, `golangci-lint` unchanged from baseline (6 + 9 pre-existing `prealloc`, all in untouched files).

Re-review when you get a chance.

ctnguyen (changes requested):
Review at commit [f7ff4de](https://github.com/ordishs/teranode/commit/f7ff4de15cfdaa3456c1deb519628627cabe2120)

This is a solid follow-up. The fixture now genuinely drives the parent-validation path, the distinct-inpoint-off-one-shared-parent trick is a clean way to satisfy `checkDuplicateInputs` while leaving a single parent to seed, and the strongest new assertions earn their keep — the `oldBlockIDsMap` case pinning `[]uint32{5}` and the multi-parent case rejecting the fallback hash both fail loudly if `nodeIndex` ever drops the parent list again. The `seedParent`/`seedDummyParent` helpers are well-documented, and the explanatory comments check out against the code (`IsTransientLocalError`, `errors/error_utils.go:68-77`; `GetBlockHeader` existence signalling, `catchup.go:655-660`; the deserialize round-trip for go-subtree v1.4.2). The `model` suite is green under `-race` and `-shuffle`, and two of the three prior points are cleanly resolved. One `createSubtreeMetadataWithParents` call site got left behind, though — and it's the one the PR's own fallback change made more fragile, so it's worth a close look.

## Status of previous points

| ChiR1 | Fixture can't drive successful validation | YES | FIXED |
| ChiR2 | Subtests discard validation result | YES | PARTIAL FIXED |
| ChiR3 | Comment cites wrong reject path/version | | FIXED |

ChiR1 and ChiR3 are done: distinct per-node inpoints plus `seedDummyParent`, the multi-node subtest now takes `require.NoError` (`Block_test.go:4494`), and the comment is retargeted at the deserialize round-trip (accurate for the pinned go-subtree v1.4.2). ChiR2 is only partial — most fixture call sites now assert, but one that uses a fixture this PR changed still ends in `_ = err`, which is ChiR4 below.

## Blocking issues

### ChiR4 - `subtree with missing parents parallel processing` is structurally vacuous and discards its result
*Same as [@oskarszoon's point 3 (29-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1364#pullrequestreview-4806120466)* — the non-target-node latent-bail sub-point; ChiR4 extends it (see Problem).

**Problem:** The subtest `TestBlock_ValidateSubtree_NodeIteration/subtree_with_missing_parents_parallel_processing` (`Block_test.go:4497-4543`) cannot exercise the parallel parent processing its name claims, and hides that behind `_ = err` (`:4542`). It builds a 2-node subtree — node 0 = `tx1Hash` (`txMap` index 2), node 1 = `tx2Hash` (`txMap` index 1) (`:4510-4513`) — then calls `createSubtreeMetadataWithParents(subtree, 1, []chainhash.Hash{*tx2Hash})` (`:4518`), attaching `tx2Hash` as the parent of node 1, which **is** `tx2Hash`. In `checkParentTransactions` (`Block.go:1414-1424`) a same-block parent whose index (`1`) is not greater than the child's (`1`) takes the `continue` branch, so node 1 contributes nothing to `checkParentTxHashes`. Node 0 gets the dummy-parent fallback this PR introduced for non-target nodes (`Block_test.go:4033`), and this subtest's `deps` uses a bare `createTestUTXOStore(t)` (`:4524`) — unlike the seeded sibling sites (`:4567`, `:4664`). So at head it bails at node 0 on the unseeded fallback (`BLOCK_INCOMPLETE` naming `dummyParentHash`), and `_ = err` swallows it. Critically, **seeding the fallback does not rescue it**: with node 1's self-parent skipped there is no missing parent left, so `validateSubtree` then returns `nil` and no fan-out over `parentG` ever happens. The subtest is vacuous under every seeding because its supplied parent hash is the node itself — a third instance of the same family as ChiR1/ChiR2 and the `nodeIndex` bug the author fixed at the other two `createSubtreeMetadataWithParents` call sites. This extends oskarszoon's point 3: he flagged that the helper left non-target nodes on a latent bail; the PR swapped that empty-`TxInpoints` bail for an unseeded fallback parent, and at this call site it is still both unseeded and unasserted — plus vacuous for the deeper self-parent reason his pre-fix pass did not reach.

**Why it matters:** The PR exists to remove exactly this trap — a subtest that passes green while testing nothing its name claims — yet it ships a fresh instance on a fixture it just modified, which is why ChiR2 is only PARTIAL and the description's claim that assertions replace `_ = err` at every fixture call site is not fully true. Because the fallback here is the one place left unseeded and the result is discarded, no future change to the fallback or the node-0 path can make this subtest go red.

**Fix:** The supplied parent has to change, not just the seeding. Point node 0 at a parent that is genuinely absent from both `txMap` and the store, seed node 1's fallback, and assert the missing-parent classification:
```go
// node 0 = tx1Hash (txMap idx 2) spends a parent absent from txMap and the store;
// node 1's fallback is seeded so only the intended parent is missing.
subtreeMetaSlice, err := createSubtreeMetadataWithParents(subtree, 0, []chainhash.Hash{oldParentHash})
require.NoError(t, err)
mockStore.data[string(subtree.RootHash()[:])] = subtreeMetaSlice

utxoStore := createTestUTXOStore(t)
seedDummyParent(t, utxoStore)

deps := &validationDependencies{
    txMetaStore: utxoStore,
    // …rest unchanged…
}

err = block.validateSubtree(ctx, ulogger.TestLogger{}, deps, validationCtx, subtree, 0)
require.ErrorIs(t, err, errors.ErrBlockIncomplete)
require.Contains(t, err.Error(), oldParentHash.String())
```
Equally valid: the sibling `parallel processing with multiple missing parents` (`:4588-4638`) already covers the missing-parent fan-out with real assertions, so this subtest can simply be deleted — or repurposed to the path its `txMap` layout actually sets up. Note that `tx2Hash` sits at index 1 and `tx1Hash` at index 2; attaching `tx1Hash` as node 1's parent would give `parentTxIdx 2 > txIdx 1` and let it assert the same-block ordering error (`ErrBlockInvalid`, "comes before parent transaction", `Block.go:1418-1420`) — genuinely distinct coverage from the sibling.

oskarszoon (approved):
Re-reviewed at `f7ff4de15`. Both P1s are fixed at the mechanism, and I checked the fixes are real guards rather than assertions that happen to pass.

Confirmed fixed: distinct inpoints per node via `dummyInpointsForNode(i)` — `Inpoint` is `{Hash, Index}` (`inpoints.go:17`) and `SetIfNotExists` keys on both (`disk_parent_spends_map.go:109`), so varying the vout is sufficient; the rationale comment now states the actual mechanism, and `deserializeFromReader` does return early on `parentCount == 0` leaving `ParentTxHashes` nil (`inpoints.go:334`); `CreateValidSubtreeMetadata` deleted with no remaining references; `createSubtreeMetadataWithParents` brought in line; the unreachable `len(subtree.Nodes) > 0` clause gone; catchup comment deduped and de-"now"-ed.

Two things beyond what I asked for that are worth calling out: replacing the `_ = err` swallowing with real assertions, and catching that `nodeIndex` pointed at a node the subtree doesn't have, silently dropping the parent list.

I verified sensitivity by reverting each fix. Reverting the vout variation:

```
--- FAIL: TestBlock_ValidateSubtree_MissingParents/multiple_invalid_parents
      BLOCK_INVALID (11): [...]:1 transaction fedcba09... has duplicate inputs
--- FAIL: TestBlock_ValidateSubtree_NodeIteration/subtree_with_multiple_nodes_iteration
      BLOCK_INVALID (11): [...]:1 transaction fedcba09... has duplicate inputs
```

Reverting `nodeIndex` 0 → 1:

```
Error: "...parent transaction a424e16c... not found in txMetaStore" should not contain "a424e16c..."
Messages: parentHashes were dropped; the fallback parent was validated instead
--- FAIL: .../oldBlockIDsMap_population   expected: []uint32{0x5}
```

Both bite, and they're sensitive to the specific bug rather than to "some error". That's the right shape.

One residual, not a blocker — **`model/Block_test.go:4518`**, `subtree with missing parents parallel processing`, is the last `_ = err` in the two touched functions and it's vacuous for the same reason you just fixed at the two neighbouring call sites. `createSubtreeMetadataWithParents(subtree, 1, []chainhash.Hash{*tx2Hash})` points node 1 — which *is* `tx2Hash` — at itself; `tx2Hash` sits in `txMap` at index 1 and `checkParentTransactions` only errors on `parentTxIdx > txIdx`, so a self-parent at equal index yields neither an ordering error nor a `missingParentTx`. Swapping `_ = err` for `require.NoError` shows what actually drives it:

```
BLOCK_INCOMPLETE (17): parent transaction a424e16c... of tx abcdef12... not found in txMetaStore
- model.getParentTxMetaBlockIDs() model/Block.go:1459
```

`a424e16c…` is `dummyParentHash`, `abcdef12…` is `tx1Hash` (node 0) — the fallback on node 0 is the whole result. Either point node 0 at a genuinely absent parent and assert the error names it, or give node 1 a real distinct parent so the ordering case it describes runs.

Nits: in `createSubtreeMetadataWithParents`, `nodeIndex == 0` on a coinbase-placeholder subtree silently drops `parentHashes` (coinbase branch wins) — same silent-drop class, worth an error return. And `dummyInpointsForNode` varies only the vout, so node i of subtree A and node i of subtree B share an inpoint; nothing exercises that today, but a multi-subtree block through this fixture with a shared `parentSpendsMap` would trip duplicate-inputs — worth a line in the doc-comment. The PR body still says "Test-only change" and still lists the merged #1361–#1363 stack.

Verified at this head: `go test ./model/` (whole package), `go test -race` on the ValidateSubtree/ValidateTransaction tests, `go test -run TestCatchup ./services/blockvalidation/`, and `go vet` on all three packages all pass.

ordishs:
@ctnguyen ChiR4 fixed at `726789047`. Your analysis was exactly right, including the part that isn't obvious — I confirmed it before changing anything:

```
seeded fallback + self-parent at nodeIndex 1  =>  err = <nil>
```

So the subtest was vacuous under every seeding, as you said: with node 1's parent being node 1 itself, `checkParentTransactions` takes the `continue` branch, nothing reaches `checkParentTxHashes`, and no group runs.

**Fix.** Node 0 now spends `tx3Hash`, absent from both `txMap` and the store; node 1 keeps the fallback, which is seeded so it resolves. Both nodes contribute an entry, so the group genuinely fans out over two parents with mixed outcomes — the accumulation across nodes that the single-node sibling can't reach, so this keeps distinct coverage rather than duplicating it.

Assertions pin both parent identities, not just the class:

```go
require.ErrorIs(t, err, errors.ErrBlockIncomplete)
require.Contains(t, err.Error(), tx3Hash.String())
require.NotContains(t, err.Error(), dummyParentHash.String())
```

The `NotContains` is the load-bearing one: it shows the fan-out had a *resolving* member rather than bailing on node 0 alone. Only one parent can fail, so `parentG.Wait()`'s error is deterministic — verified with `-count=30` and `-shuffle=on`.

Reverting to the self-parent fails as it should:

```
--- FAIL: .../subtree_with_missing_parents_parallel_processing
    Expected error with "BLOCK_INCOMPLETE (17): block incomplete" in chain but got nil.
```

**On the two alternatives you offered:** I didn't repurpose it to the same-block ordering error, because that path is already asserted at `Block_test.go:2966` and `services/blockvalidation/BlockValidation_test.go:2155`, so it would duplicate rather than add. And I didn't delete it, since the 2-node mixed case above is coverage the single-node sibling genuinely can't provide.

**ChiR2 now complete.** All nine call sites of the two fixtures this PR touches assert their result:

| Fixture | Sites | Assertion |
|---|---|---|
| `createValidSubtreeMetadata` | 6 | `require.NoError` ×3, `require.ErrorIs(ErrBlockIncomplete)` ×3 |
| `createSubtreeMetadataWithParents` | 3 | `ErrBlockIncomplete` ×2 (+ parent-identity checks), `NoError` ×1 (+ `oldBlockIDsMap == [5]`) |

**Remaining `_ = err`, flagged not fixed —** three subtests still discard their result: `parent transaction ordering error`, `parent from genesis block`, `parent from older blocks`. All three hand-roll `subtreeMetaSlice := []byte{0x01}` rather than using either fixture, so they fail at deserialize and never reach the genesis / older-block logic their names claim. Same family of problem, but outside this PR's fixtures and it needs a call on what each should assert. Happy to take it here or in a follow-up — your preference.

Verified at `726789047`: `go test -race ./model/` green, `-shuffle=on` green, `go vet` clean, `golangci-lint` unchanged from baseline (6 pre-existing `prealloc`, untouched files). `services/blockvalidation` is untouched by this commit and was green at `a550f2922`.

ctnguyen (approved):
Review at commit [7267890](https://github.com/ordishs/teranode/commit/72678904737bd051f250f2c24712a039e7ef5da2)

The last commit lands ChiR4 properly, and it's the right shape rather than the cheap one: node 0 now spends `tx3Hash`, which really is absent from both `txMap` and the store, while node 1 keeps the seeded fallback — so `checkParentTxHashes` genuinely accumulates two entries across two nodes, exactly the case the single-node sibling can't reach, and because only one member can fail, `parentG.Wait()`'s error is deterministic rather than a coin flip. Reading the path through `checkParentTransactions` and `checkParentsExistOnChain`, the `require.NotContains(err, dummyParentHash)` is the load-bearing assertion: it is the only thing that distinguishes "the fan-out had a resolving member" from "node 0 bailed alone", and your recorded revert-to-self-parent run at this head ([comment](https://github.com/bsv-blockchain/teranode/pull/1364#issuecomment-5131276851)) shows the subtest going red with `got nil`, which is the regression the commit targets. All four prior points hold up at head. The explanatory comments also check out against the code — `catchup.go:654-666` does convey existence through `GetBlockHeader` + `ErrBlockNotFound`, `ErrServiceUnavailable` is in `IsTransientLocalError` (`errors/error_utils.go:68-77`) while `ErrUnknown` isn't (`spend.go:335`), and `deserializeFromReader` does return early on `parentCount == 0` leaving `ParentTxHashes` nil in the pinned go-subtree v1.4.2 (`inpoints.go:334-336`). What's left is smaller and mostly of the same family the PR set out to kill: the helper can still drop parents silently, and two subtests now assert something their names don't describe.

## Status of previous points

ChiR2 is judged FIXED on the scope it was raised about: all nine call sites of the two fixtures this PR touches now assert (`require.NoError` at `Block_test.go:4154`, `4360`, `4494`, `4601`, `4702`; `require.ErrorIs` at `:4239`, `:4396`, `:4552`, `:4644`). Three sibling subtests in the same two test functions still discard their result — `parent transaction ordering error` (`:4265-4270`), `parent from genesis block` (`:4295`), `parent from older blocks` (`:4319`) — but all three hand-roll `subtreeMetaSlice := []byte{0x01}` and never touch either fixture, and you flagged them explicitly as a separate call. Not counted against ChiR2.

| ChiR1 | Fixture can't drive successful validation | YES | FIXED |
| ChiR2 | Subtests discard validation result | YES | FIXED |
| ChiR3 | Comment cites wrong reject path/version | | FIXED |
| ChiR4 | Parallel-parent subtest vacuous, discards result | YES | FIXED |

## Non-blocking issues

### ChiR5 - `createSubtreeMetadataWithParents` can still drop `parentHashes` silently

*Same as [@oskarszoon's point 2 (30-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1364#pullrequestreview-4818996174).*

**Problem:** He flagged the coinbase case; the drop is broader than that. `createSubtreeMetadataWithParents` (`model/Block_test.go:4000-4042`) applies `parentHashes` only when the loop reaches `i == nodeIndex` (`:4011`), and the loop is bounded by `subtree.Length()` (`:4004`). Three inputs therefore discard the caller's parents with no signal at all:
- `nodeIndex >= subtree.Length()` — the exact bug just fixed at two call sites;
- `nodeIndex < 0`;
- `nodeIndex == 0` on a coinbase-placeholder subtree, where the guard at `:4005-4008` wins.

It stays silent rather than panicking because `NewSubtreeMeta` sizes `TxInpoints` to `subtree.Size()` (cap, not len — `go-subtree@v1.4.2/subtree_meta.go:34-38`) and `Meta.Serialize` only validates indices below `Subtree.Length()` (`subtree_meta.go:172-177`). So an out-of-range write target is simply never written, the metadata serializes cleanly, and every node ends up on the dummy fallback.

**Why it matters:** This helper is the mechanism behind two of the four regressions this PR fixes. The PR corrected the two symptomatic call sites but left the helper able to reproduce the fault the moment someone changes a subtree's leaf count or points the fixture at a coinbase subtree — and the only thing standing between that and a green vacuous test is the per-call-site `NotContains(dummyParentHash)` assertions, which the next author has no reason to know are load-bearing.

**Fix:** Refuse the input instead of silently reinterpreting it:
```go
func createSubtreeMetadataWithParents(subtree *subtreepkg.Subtree, nodeIndex int, parentHashes []chainhash.Hash) ([]byte, error) {
	if len(parentHashes) > 0 {
		if nodeIndex < 0 || nodeIndex >= subtree.Length() {
			return nil, fmt.Errorf("nodeIndex %d out of range for subtree of length %d: parentHashes would be dropped",
				nodeIndex, subtree.Length())
		}

		if nodeIndex == 0 && subtree.Nodes[0].Hash.Equal(subtreepkg.CoinbasePlaceholderHashValue) {
			return nil, fmt.Errorf("nodeIndex 0 is the coinbase placeholder: parentHashes would be dropped")
		}
	}
	// …unchanged…
```
`fmt` is already imported (`:10`), and every call site already does `require.NoError` on the returned error, so this turns the whole class into an immediate, self-describing failure.

**Criticality: 35%**

### ChiR6 - `empty missing parents list` can no longer have an empty list, and the new comment says it does

**Problem:** The subtest at `model/Block_test.go:4322-4361` is the one that covers the `len(checkParentTxHashes) == 0` early return (`model/Block.go:1205-1209`). Reading the path, it can no longer reach it. Its single node spends `dummyParentHash`, which no caller ever puts into `block.txMap`, so `checkParentTransactions` finds no `foundInSameBlock` match and appends a `missingParentTx` (`Block.go:1414-1427`). The list is length 1, so `checkParentsExistOnChain` runs, the seeded parent resolves through the `minBlockID > 0` branch (`Block.go:1305-1317`), and the result is nil. The added comment at `:4341-4342` — "Seeding the dummy parent is what makes this the 'no missing parents' case: every node's parent resolves" — conflates "resolves on chain" with "never entered the missing-parents list". The `require.NoError` is real, but on this reading it asserts the fan-out success path that `single node subtree` (`:4561-4602`) already covers, not the empty-list branch its name promises.

**Why it matters:** This is the failure mode the PR exists to remove: a green test whose name and comment describe a path it does not touch. It also means the `len(checkParentTxHashes) > 0` false branch is reachable only through the coinbase-only subtest at `:4098`, i.e. the non-coinbase empty-list case has no coverage at all.

**Fix:** Put the parent in the same block at a lower index, which is how production reaches that branch — and it drops the store dependency entirely:
```go
// dummyParentHash sits in the same block below the child, so checkParentTransactions
// takes its foundInSameBlock/continue branch and the missing-parent list really is
// empty: validateSubtree returns at the len(checkParentTxHashes) > 0 guard without
// ever consulting the store.
err = block.txMap.Put(dummyParentHash, 0)
require.NoError(t, err)
```
With that, `seededStore` / `seedDummyParent` / `seededDeps` (`:4343-4351`) are no longer needed here, and the subtest name becomes accurate. If you'd rather keep the current mechanics, rename it to something like `parent resolves outside the block` and move the empty-list case into its own subtest.

**Criticality: 30%**

### ChiR7 - `dummyInpointsForNode` collides across subtrees of the same block

*Same as [@oskarszoon's point 3 (30-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1364#pullrequestreview-4818996174).*

**Problem:** `dummyInpointsForNode(idx)` (`model/Block_test.go:3947-3957`) keys uniqueness on the vout alone, so node `i` of subtree A and node `i` of subtree B both produce `{dummyParentHash, i}`. The shared-state half of this is confirmed in the source: `validOrderAndBlessed` builds exactly one `parentSpendsMap` per block and hands the same `validationContext` to every `validateSubtree` goroutine (`model/Block.go:1105-1146`), and `SetIfNotExists` keys on the whole `Inpoint` struct (`model/map.go:48-62`, `Inpoint` is `{Hash, Index}`). A fixture-built block with two or more subtrees would therefore trip `checkDuplicateInputs` at the second subtree. That the mechanism bites rather than being theoretical is already on record twice: your own regression sweep at `f7ff4de` reports `shared inpoint -> BLOCK_INVALID …:1 has duplicate inputs (2 subtests fail)` ([comment](https://github.com/bsv-blockchain/teranode/pull/1364#issuecomment-5130062618)), and @oskarszoon reproduced the same failure independently by reverting the vout variation ([review](https://github.com/bsv-blockchain/teranode/pull/1364#pullrequestreview-4818996174)). Both cover the within-subtree case; this is the same collision one level up. Nothing triggers it today — all six `createValidSubtreeMetadata` call sites build one subtree and call `validateSubtree` directly — so it is latent, not broken.

**Why it matters:** It is the trap the PR just removed at the node level, reappearing at the subtree level, and the failure it would produce (`has duplicate inputs` on a fixture that looks obviously fine) is precisely the one that cost two review rounds to diagnose. The doc comment currently states the invariant as though it were absolute ("varying the vout is what keeps each node in a fixture distinct"), which reads as a guarantee it doesn't provide across fixtures.

**Fix:** Make the vout process-unique instead of node-unique, which keeps the single-parent-to-seed property that motivated the design:
```go
// dummyInpointVout hands out a process-unique vout, so no two fixture nodes ever
// share an inpoint — including nodes in different subtrees of the same block,
// which share one parentSpendsMap (Block.go:1129-1146).
var dummyInpointVout atomic.Uint32

func dummyInpointsForNode() (subtreepkg.TxInpoints, error) {
	input := &bt.Input{PreviousTxOutIndex: dummyInpointVout.Add(1)}
	if err := input.PreviousTxIDAdd(&dummyParentHash); err != nil {
		return subtreepkg.TxInpoints{}, err
	}

	return subtreepkg.NewTxInpointsFromInputs([]*bt.Input{input})
}
```
(needs `sync/atomic`; nothing asserts a specific vout). If you'd rather not lose the `node i ↔ vout i` readability, the minimum is a line in the doc comment stating the limit explicitly.

**Criticality: 25%**

### ChiR8 - The new hard assertions run under one 2s context shared by every subtest

**Problem:** `TestBlock_ValidateSubtree_MissingParents` and `TestBlock_ValidateSubtree_NodeIteration` each create a single `context.WithTimeout(…, 2*time.Second)` at function scope (`model/Block_test.go:4160-4161` and `:4402-4403`) and pass it to all six subtests. Pre-PR that was harmless because the results were discarded. Now `:4239`, `:4396`, `:4552` and `:4644` assert `require.ErrorIs(err, errors.ErrBlockIncomplete)`. If the shared deadline expires part-way through the suite, `txMetaStore.Get` returns a context error, which is not `ErrTxNotFound`, so `getParentTxMetaBlockIDs` wraps it as a `StorageError` (`model/Block.go:1462`) rather than `BlockIncompleteError` — and the assertion fails with "target error not in chain" pointing at the parent-classification logic instead of at a timeout. The `require.NoError` sites fail the same way. Each subtest also stands up its own `sqlitememory` store inside that shared budget (`:4343`, `:4473`, `:4532`, `:4582`, `:4679`), and this PR adds a `SpendAndCreate` seeding round-trip to most of them.

**Why it matters:** It converts a pre-existing timing smell into a suite that can go red intermittently and be misdiagnosed. Your recorded run at this head reports `go test -race ./model/` and `-shuffle=on` green ([comment](https://github.com/bsv-blockchain/teranode/pull/1364#issuecomment-5131276851)), so there is no evidence of a problem on your machine — the concern is the margin on a loaded runner, where six store creations plus seeding inside one 2-second window is not generous, and the shape of the failure when the margin is lost.

**Fix:** Give each subtest its own deadline, so a slow neighbour can't consume it and a timeout reports as a timeout:
```go
t.Run("missing parent not found in store", func(t *testing.T) {
	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
	defer cancel()
	// …
```
Dropping the per-test timeout entirely and relying on `go test -timeout` is equally fine here — nothing in these subtests depends on cancellation.

**Criticality: 18%**

### ChiR9 - `multiple invalid parents` asserts neither multiple nor invalid

**Problem:** The subtest at `model/Block_test.go:4363-4397` builds two nodes, both of which take the `createValidSubtreeMetadata` fallback, so both point at the *same* parent (`dummyParentHash`, vouts 0 and 1) and it is unseeded. `checkParentTxHashes` gets two entries naming one parent hash, and the assertion at `:4396` is `require.ErrorIs(err, errors.ErrBlockIncomplete)` — retryable, not invalid. So there is one parent, and it is missing rather than invalid; the new comment at `:4393-4394` ("the unseeded parent is reported missing", singular) describes the real behaviour accurately while the subtest name contradicts it.

**Why it matters:** Same class as ChiR6 — the name is what the next reader trusts when deciding whether the invalid-parent path is covered, and `ErrBlockInvalid` from `ErrCheckParentExistsOnChain` (`model/Block.go:1319-1321`) has no coverage in these two test functions.

**Fix:** Cheapest is a rename that matches the assertion, e.g. `unseeded parent shared by multiple nodes`. If you want the name to stay and the coverage to appear, seed the parent into two block IDs that are both in the cached set, which should drive `len(foundInPreviousBlocks) == 2` and the invalid classification:
```go
utxoStore := createTestUTXOStore(t)
seedDummyParent(t, utxoStore) // mined in block 1
_, err = utxoStore.SetMinedMulti(ctx, []*chainhash.Hash{&dummyParentHash},
	utxo.MinedBlockInfo{BlockID: 2, BlockHeight: 2})
require.NoError(t, err)
// currentBlockHeaderIDsMap is {1, 2}, so the parent is found twice →
// ErrCheckParentExistsOnChain returns BLOCK_INVALID "… found 2 times".
```

**Criticality: 14%**

## Recap

| ChiR5 | Helper still drops parents silently | | 35% |
| ChiR6 | Empty-parents subtest can't be empty | | 30% |
| ChiR7 | Inpoints collide across subtrees | | 25% |
| ChiR8 | Shared 2s context gates assertions | | 18% |
| ChiR9 | Subtest name contradicts assertion | | 14% |

PR #1366, opened 2026-07-28
perf(aerospike/pruner): route pruner writes through the native-op builder

ordishs:
Extracted from `feat/teranode-native-ops` (PR #1360), the last of a six-PR stack. The other five (#1361–#1365) have all merged, so this now applies directly to `main` and its diff is only its own changes.

## Summary

Routes the pruner's combined-cleanup parent updates and `addDeletedChildren` through the store's native-op builder when `aerospike_use_native_teranode_ops` is active, falling back to the Lua UDF otherwise. The provider injects the builder, so the `pruner` package stays decoupled from the store's native-op internals — it never learns about wire op 200, sub-op IDs, or msgpack.

The bug this fixes: `executeBatchCleanupCombined` keyed the parent-record choice on `luaPackage != ""` alone, so native-op deployments running with defensive mode **off** kept emitting a per-block `batch_sub_udf` burst. `buildParentUpdateRecords` is now shared by the combined and two-call paths so they cannot silently diverge again.

## Review changes

Parent-update result handling now **fails closed**. A missing `SUCCESS` bin, an unreadable response shape, or a status that is neither `OK` nor `ERROR` previously all resolved to `parentSuccess++`, so a genuine native-dispatcher failure could be tallied as success while the child records in the same batch were deleted — breaking exactly the referential integrity the parent-update phase exists to preserve. `classifyParentUpdateResult` reports those as failures describing what it saw, and both paths share it.

Other fixes:

- Missing parents are counted as skips whether they arrive as `KEY_NOT_FOUND` (the native path under the `UPDATE_ONLY` policy #1365 installs) or as an `ERROR`/`TX_NOT_FOUND` response map (the Lua path). Neither creates the record.
- Responses are read from `message`, the key `teranode.lua` and `ParseLuaMapResponse` actually use; `errorMessage` is kept as a tolerated alias. The server-supplied message reaches the log again on the two-call path.
- The client's ordered-map (`[]aerospike.MapPair`) encoding is accepted alongside both map shapes.
- Error and log labels are path-neutral, since a failing record may now be a native batch write rather than a Lua UDF.
- `Options.ObserveNativeOpError` lets parent-update failures reach `demoteNativeOnUnsupported`. The pruner runs its own `BatchOperate`, so it was the one native call site that could not trigger the runtime fallback — and since it runs every block, it is a likely first victim of a mixed-version cluster that passed the single-partition startup probe.

## Tests

- `parent_update_response_test.go` — response parsing across all three map shapes, the fail-closed classifier, result tallying, and the native-op error observer
- `combined_cleanup_native_op_test.go` — cleanup prefers the injected native builder over raw UDF
- `native_op_pruner_record_test.go` — the record the pruner emits is a native `BatchWrite` under `UPDATE_ONLY` (never creating a missing parent), falls back to `BatchUDF` when native ops are off, encodes the hash list as a single argument, and demotes to UDF on `PARAMETER_ERROR` while ignoring `KEY_NOT_FOUND` / timeouts
- `native_op_add_deleted_children_integration_test.go` — `aerospike`-tagged; see the caveat below

## Test coverage caveat

**The native `addDeletedChildren` path has no automated verification in CI.** `native_op_add_deleted_children_integration_test.go` needs a Teraspike server (the BSV fork, wire op 200) and selects it via `AEROSPIKE_CONTAINER_IMAGE`, which is set nowhere in this repo — so the test always hits its skip. Everything else here is covered by unit tests, but the two things only a real fork server can settle are:

1. that the dispatcher's response for this op is a shape the pruner can read — now load-bearing, because unreadable responses fail closed rather than being counted as success; and
2. that a missing parent is not created.

To run it against a fork build:

```bash
AEROSPIKE_CONTAINER_IMAGE=<teraspike-image> go test -tags aerospike \
  -run TestNativeOp_AddDeletedChildren_Integration ./stores/utxo/aerospike/
```

The test asserts both points, plus that `batch_sub_udf_complete` does not move while the parent is mutated — and it now requires that statistic to exist on some node, so a renamed or version-gated counter fails loudly instead of comparing 0 to 0.

## Verification

```
gofmt -l stores/utxo/aerospike/           → (empty)
go build ./...                            → exit 0
go vet ./stores/... ./settings/...        → exit 0
go vet -tags aerospike ./stores/...       → exit 0
go test -count=1 ./stores/utxo/aerospike/...
    ok  .../stores/utxo/aerospike         267.183s
    ok  .../stores/utxo/aerospike/pruner   43.842s
golangci-lint run stores/utxo/aerospike/...
    1 issue, pre-existing, in longest_chain_production_bench_test.go (untouched)
```

The setting defaults to off, so the native path is opt-in.

### DIFF
--- a/stores/utxo/aerospike/native_op_add_deleted_children_integration_test.go
+++ b/stores/utxo/aerospike/native_op_add_deleted_children_integration_test.go
@@ -1,0 +1,143 @@
+//go:build aerospike
+
+package aerospike
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"os"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/bsv-blockchain/aerospike-client-go/v8"
+	"github.com/bsv-blockchain/go-bt/v2/chainhash"
+	"github.com/bsv-blockchain/teranode/stores/blob/memory"
+	"github.com/bsv-blockchain/teranode/stores/utxo/fields"
+	"github.com/bsv-blockchain/teranode/ulogger"
+	"github.com/bsv-blockchain/teranode/util/test"
+	"github.com/stretchr/testify/require"
+)
+
+// sumNamespaceStat reads a numeric Aerospike namespace statistic, summed across
+// every node in the cluster.
+func sumNamespaceStat(t *testing.T, db *Store, stat string) int64 {
+	t.Helper()
+
+	cmd := "namespace/" + db.namespace
+
+	var total int64
+
+	nodes := db.client.Cluster().GetNodes()
+	require.NotEmpty(t, nodes, "cluster must expose at least one node")
+
+	infoPolicy := aerospike.NewInfoPolicy()
+
+	for _, nd := range nodes {
+		info, err := nd.RequestInfo(infoPolicy, cmd)
+		require.NoError(t, err)
+
+		for _, kv := range strings.Split(info[cmd], ";") {
+			name, value, found := strings.Cut(kv, "=")
+			if !found || name != stat {
+				continue
+			}
+
+			v, perr := strconv.ParseInt(value, 10, 64)
+			require.NoError(t, perr)
+
+			total += v
+		}
+	}
+
+	return total
+}
+
+// TestNativeOp_AddDeletedChildren_Integration verifies against a real Teraspike
+// server (the BSV fork of aerospike-server with native operate-path support,
+// wire op 200) that the pruner's addDeletedChildren parent-update — the op that
+// PR #1269 re-routed onto the native builder in the combined cleanup path — runs
+// as a NATIVE batch write (zero batch_sub_udf) and correctly mutates the parent's
+// deletedChildren map.
+//
+// It builds the record via exactly the call GetPrunerService's
+// BuildAddDeletedChildrenRecord uses, so it exercises the shared native path the
+// pruner relies on. Skips cleanly on a stock Aerospike image (native-op probe
+// fails); point at a Teraspike build with AEROSPIKE_CONTAINER_IMAGE to run it.
+func TestNativeOp_AddDeletedChildren_Integration(t *testing.T) {
+	InitPrometheusMetrics()
+
+	logger := ulogger.NewErrorTestLogger(t)
+	ctx := context.Background()
+
+	container, err := runAerospikeTestContainer(ctx)
+	if err != nil {
+		t.Skipf("Skipping: Aerospike container not available (%v)", err)
+	}
+
+	t.Cleanup(func() { require.NoError(t, container.Terminate(ctx)) })
+
+	host, err := container.Host(ctx)
+	require.NoError(t, err)
+
+	port, err := container.ServicePort(ctx)
+	require.NoError(t, err)
+
+	tSettings := test.CreateBaseTestSettings(t)
+	tSettings.Aerospike.UseNativeTeranodeOps = true
+
+	aeroURL, err := url.Parse(fmt.Sprintf("aerospike://%s:%d/test?set=test&externalStore=file://./data/externalStore", host, port))
+	require.NoError(t, err)
+
+	db, err := New(ctx, logger, tSettings, aeroURL)
+	require.NoError(t, err)
+
+	db.SetExternalStore(memory.New())
+
+	if !db.useNativeTeranodeOps {
+		t.Skipf("native ops not enabled (image %q is not a Teraspike build); set AEROSPIKE_CONTAINER_IMAGE to a native-op server", os.Getenv("AEROSPIKE_CONTAINER_IMAGE"))
+	}
+
+	// Seed a parent record so addDeletedChildren has a record to update
+	// (the op returns TX_NOT_FOUND for a missing record and mutates nothing).
+	var parentHash chainhash.Hash
+	parentHash[0] = 0xAB
+
+	key, err := aerospike.NewKey(db.namespace, db.setName, parentHash[:])
+	require.NoError(t, err)
+	require.NoError(t, db.client.PutBins(nil, key, aerospike.NewBin("seed", 1)))
+
+	var childHash chainhash.Hash
+	childHash[0] = 0xCD
+
+	childList := []interface{}{childHash.String()}
+
+	udfBefore := sumNamespaceStat(t, db, "batch_sub_udf_complete")
+
+	// Build the record exactly as GetPrunerService.BuildAddDeletedChildrenRecord does.
+	rec := db.teranodeBatchRecord(aerospike.NewBatchUDFPolicy(), LuaPackage, key, subOpAddDeletedChildren, "addDeletedChildren", childList)
+
+	_, builtAsUDF := rec.(*aerospike.BatchUDF)
+	require.False(t, builtAsUDF, "a native-op store must build addDeletedChildren as a BatchWrite, not a NewBatchUDF")
+
+	require.NoError(t, db.client.BatchOperate(nil, []aerospike.BatchRecordIfc{rec}))
+	require.NoError(t, rec.BatchRec().Err, "native addDeletedChildren batch operation must succeed")
+
+	// The parent's deletedChildren map must now contain the child hash — proof
+	// the native subOpAddDeletedChildren dispatcher executed server-side.
+	parent, err := db.client.Get(nil, key, fields.DeletedChildren.String())
+	require.NoError(t, err)
+	require.NotNil(t, parent)
+
+	deletedChildren, ok := parent.Bins[fields.DeletedChildren.String()].(map[interface{}]interface{})
+	require.Truef(t, ok, "deletedChildren must be a map, got %T", parent.Bins[fields.DeletedChildren.String()])
+	require.Equal(t, true, deletedChildren[childHash.String()], "child hash must be recorded in the parent's deletedChildren map")
+
+	udfAfter := sumNamespaceStat(t, db, "batch_sub_udf_complete")
+
+	// The decisive check: the parent was mutated (native dispatcher ran) yet the
+	// server's Lua-UDF sub-transaction counter did not move at all. The native
+	// wire-op-200 path is deliberately not counted under batch_sub_udf.
+	require.Equalf(t, udfBefore, udfAfter, "addDeletedChildren must NOT invoke a Lua UDF on a native-op store (batch_sub_udf_complete went %d -> %d)", udfBefore, udfAfter)
+}
--- a/stores/utxo/aerospike/native_op_add_deleted_children_integration_test.go
+++ b/stores/utxo/aerospike/native_op_add_deleted_children_integration_test.go
@@ -22,12 +22,20 @@ import (
 
 // sumNamespaceStat reads a numeric Aerospike namespace statistic, summed across
 // every node in the cluster.
+//
+// It requires the statistic to exist on at least one node. Returning a silent 0
+// for an unknown stat name would make the caller's before/after comparison pass
+// vacuously if the counter were ever renamed or version-gated away — the
+// comparison would be 0 == 0 while proving nothing.
 func sumNamespaceStat(t *testing.T, db *Store, stat string) int64 {
 	t.Helper()
 
 	cmd := "namespace/" + db.namespace
 
-	var total int64
+	var (
+		total int64
+		found bool
+	)
 
 	nodes := db.client.Cluster().GetNodes()
 	require.NotEmpty(t, nodes, "cluster must expose at least one node")
@@ -39,8 +47,8 @@ func sumNamespaceStat(t *testing.T, db *Store, stat string) int64 {
 		require.NoError(t, err)
 
 		for _, kv := range strings.Split(info[cmd], ";") {
-			name, value, found := strings.Cut(kv, "=")
-			if !found || name != stat {
+			name, value, hasValue := strings.Cut(kv, "=")
+			if !hasValue || name != stat {
 				continue
 			}
 
@@ -48,18 +56,20 @@ func sumNamespaceStat(t *testing.T, db *Store, stat string) int64 {
 			require.NoError(t, perr)
 
 			total += v
+			found = true
 		}
 	}
+
+	require.Truef(t, found, "namespace statistic %q not reported by any node — the before/after comparison would be vacuous", stat)
 
 	return total
 }
 
 // TestNativeOp_AddDeletedChildren_Integration verifies against a real Teraspike
 // server (the BSV fork of aerospike-server with native operate-path support,
-// wire op 200) that the pruner's addDeletedChildren parent-update — the op that
-// PR #1269 re-routed onto the native builder in the combined cleanup path — runs
-// as a NATIVE batch write (zero batch_sub_udf) and correctly mutates the parent's
-// deletedChildren map.
+// wire op 200) that the pruner's addDeletedChildren parent-update runs as a
+// NATIVE batch write (zero batch_sub_udf), correctly mutates the parent's
+// deletedChildren map, and answers with a response the pruner can actually read.
 //
 // It builds the record via exactly the call GetPrunerService's
 // BuildAddDeletedChildrenRecord uses, so it exercises the shared native path the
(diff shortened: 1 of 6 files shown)


ctnguyen (changes requested):
Review at commit [d4677c7](https://github.com/ordishs/teranode/commit/d4677c7954847c3a68f9514a8dcc39d862f58bd0)

Nice last step in the stack — the injection-through-`Options` approach is the right call here: the `pruner` package never learns about wire op 200, sub-op IDs, or msgpack, and the store hands it a builder that already knows how to fall back. Pulling `buildParentUpdateRecords` out so the combined and two-call paths share one record builder fixes a genuine divergence (the combined path was keying purely on `luaPackage` and quietly kept emitting `batch_sub_udf` on native-op deployments with defensive mode off), and the unit tests cover the three routing branches plus both response-map shapes. My concerns are all on the response/policy side of the new path rather than the routing itself: the response handling fails open in several places, and the native record is built with a create-if-missing write policy that neither path it replaces used.

## Blocking issues

### ChiR1 - Parent-update response handling fails open: anything not understood counts as success
**Problem:** Every "I could not read this response" outcome in the two parent-update loops resolves to success or to a silent skip, and this PR is what puts a second, unverified producer behind those loops. Concretely, for a parent record with no `batchRec.Err`: a missing `SUCCESS` bin falls through to `parentSuccess++` (`pruner_service.go:1755`) / `successCount++` (`:1889`); a `SUCCESS` bin that `addDeletedChildrenStatus` (`:1918`) cannot read — it accepts exactly `map[interface{}]interface{}` and `map[string]interface{}` — does the same; and a *parseable* response whose `status` is neither `"OK"` nor `"ERROR"` hits neither `switch` case and is then `continue`d, so it is counted as nothing at all. On the pure UDF path that was tolerable, since `teranode.lua` is the only producer and always emits `map[interface{}]interface{}` with `status` ∈ {`OK`,`ERROR`}. The native path's response shape is not pinned the same way: the only native response verified to parse anywhere in the tree is the `setLocked` probe in `detectNativeTeranodeOpSupport` (`native_op.go:281`), the store's own normaliser `luaResponseMap` (`teranode.go:452-474`) carries a reflect-based fallback for map kinds beyond those two, and the client's unpacker can hand back `[]aerospike.MapPair` rather than a Go map when a map particle carries the ordered-map ext header (`unpacker.go:128-177`, `isMapCDT`) — a shape neither the helper nor `luaResponseMap` accepts. I am not claiming the dispatcher emits that shape; the point is that nothing here establishes what it emits. The `map[string]interface{}` case the comment and `parent_update_response_test.go` label as "the native-op shape" is itself unverified — `unpackMapNormal` always builds `map[any]any` — and the new `aerospike`-tagged integration test asserts the record mutation and the zero-`batch_sub_udf` count but never inspects the returned bin, so a mismatch would pass CI green.
**Why it matters:** A genuine `ERROR` from a native parent update can be tallied as a success: `flushCleanupBatches` returns `nil`, the child records in the same batch are deleted even though the parent's `deletedChildren` map was never updated (referential integrity is what the parent-update phase exists to preserve), and `prometheusUtxoParentsUpdated` over-reports while `...Skipped` reads zero. The store fails *closed* on the same unknown — `ParseLuaMapResponse` returns an error and callers log it; the pruner fails *open* and stays quiet.
**Fix:** Make every unreadable outcome fail closed and say what it saw, so a shape or vocabulary mismatch shows up as a loud pruner error instead of a clean run. In both loops:

```go
resp, hasResp := batchRec.Record.Bins["SUCCESS"]
if !hasResp {
	parentErrors++
	if firstParentErr == nil {
		firstParentErr = errors.NewProcessingError("parent update returned no %q bin", "SUCCESS")
	}
	continue
}
status, errCode, errMsg, ok := addDeletedChildrenStatus(resp)
switch {
case !ok:
	parentErrors++ // unreadable response shape — do not assume success
	if firstParentErr == nil {
		firstParentErr = errors.NewProcessingError("unparseable parent update response (%T)", resp)
	}
case status == "OK":
	parentSuccess++
case status == "ERROR" && errCode == "TX_NOT_FOUND":
	parentNotFound++
case status == "ERROR":
	parentErrors++
	if firstParentErr == nil {
		firstParentErr = errors.NewProcessingError("parent update returned ERROR (code=%q, message=%q)", errCode, errMsg)
	}
default:
	parentErrors++ // unknown status — currently swallowed entirely
	if firstParentErr == nil {
		firstParentErr = errors.NewProcessingError("parent update returned unknown status %q", status)
	}
}
```

and add a defensive `[]aerospike.MapPair` case to `addDeletedChildrenStatus` so an ordered-map response is read rather than rejected. Then pin the shape with a real assertion in `TestNativeOp_AddDeletedChildren_Integration`, which is the cheapest way to close the open question:

```go
resp := rec.BatchRec().Record.Bins[nativeOpResultBin]
parsed, perr := db.ParseLuaMapResponse(resp)
require.NoErrorf(t, perr, "native addDeletedChildren response bin is not parseable (%T)", resp)
require.Equal(t, LuaStatusOK, parsed.Status)
```

### ChiR2 - Native parent updates use a create-if-missing write policy, unlike both paths they replace
**Problem:** `pruner_provider.go:64-73` routes `addDeletedChildren` through `teranodeBatchRecord`, which builds `aerospike.NewBatchWrite(s.nativeOpBatchWritePolicy, ...)` (`native_op.go:119-139`). `NewBatchWritePolicy()` defaults `RecordExistsAction` to `UPDATE` — create the record if it is absent. Both alternatives for this exact job deliberately are not create-if-missing: the Lua `addDeletedChildren` returns `TX_NOT_FOUND` and never calls `aerospike:update` when `aerospike:exists(rec)` is false (`teranode.lua:1156-1165`), and the plain MapPutItems path sets `RecordExistsAction = UPDATE_ONLY` (`pruner_service.go:1635` and `:1940`). The new `Options` doc even states "the native-op branch synthesises its own write policy", so dropping the policy is intentional — but the consequence looks unconsidered, because the pruner is the one native call site where a missing target record is the *routine* case (the code counts it as `parentNotFound`, and the missing parent is normally one the pruner itself already deleted). Whether that write resurrects a partial record now rests entirely on server-fork behaviour the client cannot verify, and nothing exercises it: the integration test seeds the parent with `PutBins` first (`native_op_add_deleted_children_integration_test.go:109`).
**Why it matters:** A resurrected parent would carry only `deletedChildren` — no `txid`, no `deleteAtHeight` — so the pruner's own DAH scan will never reclaim it (unbounded namespace growth on every pruned block), and any later read of it fails in `extractTxHash` with "invalid or missing txid". This is the same class of unverifiable-server-enforcement risk that motivated fencing `subOpUnspend` to the UDF path in the core PR, applied to the one op whose target is expected to be absent.
**Fix:** Either confirm the dispatcher's missing-record semantics and record that in the `Options` contract, backed by a test that runs the op against a **non-existent** key and asserts no record is created; or make the intent explicit client-side with a per-call `UPDATE_ONLY` policy copy (same pattern `teranodeBatchRecord` already uses for `FilterExpression`). If you take the `UPDATE_ONLY` route, both parent-update loops must first learn to tolerate `KEY_NOT_FOUND`, because a missing parent is never an error for this op and neither loop tolerates it on a mod-teranode path today:

```go
// pruner_service.go:1718 — combined cleanup: drop the !parentUsesModTeranode guard
if batchRec.Err.Matches(aerospike.ErrKeyNotFound.ResultCode) {
	parentNotFound++
	continue
}

// pruner_service.go:1860 — executeBatchParentUpdatesUDF (defensive-mode two-call path)
// currently counts ANY batchRec.Err as an error before the SUCCESS map is read,
// so it needs the same tolerance added:
if batchRec.Err != nil {
	if batchRec.Err.Matches(aerospike.ErrKeyNotFound.ResultCode) {
		notFoundCount++
		continue
	}
	...
}
```

Without that second change, switching to `UPDATE_ONLY` would turn every routine missing parent into a `StorageError` and fail the cleanup batch on each defensive-mode pruning cycle.

## Non-blocking issues

### ChiR3 - `errorMessage` is read from a key the op never sets, and the log that used to carry it was dropped
**Problem:** `addDeletedChildrenStatus` reads `m["errorMessage"]`, but the response field is `message` — `teranode.lua:83` (`FIELD_MESSAGE = "message"`) and `addDeletedChildren` sets `response[FIELD_MESSAGE]`; the store's own parser reads `respMap["message"]` (`teranode.go:375`). So the synthesised `firstParentErr` at `:1747` always renders `message=""`. The wrong key predates this PR, but the diff makes it worse in two ways: the refactor replaced `logger.Errorf("Parent update Lua error for key %v: %v", batchRec.Key, respMap)` with `code=%q` (`:1880`), removing the one place the message actually reached an operator; and the new unit test bakes the wrong contract in (`parent_update_response_test.go:58-68` asserts `errMsg == "boom"` from an `errorMessage` key that no producer emits).
**Why it matters:** A non-`TX_NOT_FOUND` parent-update failure now surfaces as a bare error code with no server-supplied detail, on the path that returns a `StorageError` and aborts the cleanup batch.
**Fix:** Read `message` (keeping `errorMessage` as a tolerated alias if you like), thread it into the two-call path's log line, and update the test:

```go
errorMessage, _ = m["message"].(string)
if errorMessage == "" {
	errorMessage, _ = m["errorMessage"].(string)
}
...
s.logger.Errorf("Parent update error for key %v: code=%q message=%q", batchRec.Key, errCode, errMsg)
```

### ChiR4 - "UDF"/"Lua" labels on failures that may have come from the native path
**Problem:** `pruner_service.go:1747` synthesises `"UDF parent update returned ERROR ..."` and `:1880` logs `"Parent update Lua error ..."`, but with `buildAddDeletedChildrenRecord` injected the failing record is a wire-op-200 `BatchWrite`, not a Lua UDF invocation.
**Why it matters:** Points on-call at the UDF executor for a failure that never went through it — small, but this is the primary diagnostic for a path that has no other observability.
**Fix:** Make the labels path-neutral (`"parent update returned ERROR (code=%q, message=%q)"`, `"Parent update error for key %v"`), or include which builder produced the record.

## Recap

| ChiR1 | Parent-update parsing fails open | YES | 75% |
| ChiR2 | Create-if-missing policy on parent update | YES | 65% |
| ChiR3 | Wrong response key for message | | 20% |
| ChiR4 | Misleading UDF/Lua log labels | | 10% |

oskarszoon (approved):
Reviewed the tip commit only (`d4677c795`, 5 files) — #1361–#1365 as their own PRs.

The bug this fixes is real and worth the extraction: keying the parent-record choice on `luaPackage != ""` alone meant native-op deployments with defensive mode **off** kept emitting a per-block `batch_sub_udf` burst. Sharing `buildParentUpdateRecords` between the combined and two-call paths is the more durable half of the change — they can't silently diverge again.

Checked the refactor for behaviour equivalence line by line. `addDeletedChildrenStatus` stays correct only because `ok` is bound to the *status* assertion rather than the map assertion, preserving the old two-level guard's fall-through to `parentSuccess++`. Load-bearing and easy to break later, but right as written. Walked the configuration space too: the provider always sets both `LuaPackage` and `BuildAddDeletedChildrenRecord`, so the plain-`MapPutItems` branches are unreachable for this store and `parentUsesModTeranode` matches the record type actually built in every combination. No silent misclassification of missing parents.

`go build ./...`, `go vet` (with and without `-tags aerospike`), and `go test ./stores/utxo/aerospike/pruner/...` (54 tests) all pass locally.

Two P2s, neither blocking:

**1. `errorMessage` is not a key any producer emits — `pruner_service.go:1923,1927`**

The Lua side uses `message`: `teranode.lua:83` (`FIELD_MESSAGE = "message"`), set at `teranode.lua:1162` on the not-found branch. The store's own parser agrees — `teranode.go:375` reads `respMap["message"]`, backing `LuaErrorInfo.Message` (`json:"message"`, `teranode.go:265`). So `errorMessage` is always `""` and the synthesised error at `:1747` always reads `message=""`.

The wrong key is pre-existing, but this PR makes it stickier and loses a field:

- `parent_update_response_test.go:62` asserts `"errorMessage": "boom"` round-trips. A green test now protects a field the server never sends.
- `pruner_service.go:1880` regressed on the two-call path (defensive mode **on**): it used to log the whole `respMap`, which *included* `message`; it now logs `code=%q` and discards `errMsg` via `_`. The helper already has the value in hand.

Read `m["message"]`, thread it through at `:1880`. The strings at `:1747` and `:1880` also still say "UDF"/"Lua" on what is now equally the native path.

**2. The integration test asserts its key contract in a comment — `native_op_add_deleted_children_integration_test.go`**

`:103` states "the op returns TX_NOT_FOUND for a missing record and mutates nothing", then `:109` seeds the parent, so that branch never runs. It's the pruner's routine case — a parent DAH-pruned before its child is exactly what `prometheusUtxoParentsUpdatedSkipped` counts. Pre-change the guard was client-inspectable (`aerospike:exists(rec)`, `teranode.lua:1159`); post-change it's the server-fork dispatcher's, unverified from here. ~6 lines: same `teranodeBatchRecord` call against an unwritten key, assert `ERROR`/`TX_NOT_FOUND` and that `client.Get` still returns nil.

Two smaller things in the same file: `:142`'s `require.Equalf(udfBefore, udfAfter)` is billed as the decisive check, but `sumNamespaceStat` returns `0` when the stat name doesn't match anything, so a renamed or version-gated `batch_sub_udf_complete` makes it pass vacuously — require the stat was found at least once. And `AEROSPIKE_CONTAINER_IMAGE` is set nowhere in the repo, so the test always hits the skip at `:99`; intentional, but it means the native `addDeletedChildren` path has no automated verification in CI. Worth saying in the PR body. (`require.False(builtAsUDF)` at `:122` is sound — that's the check carrying the test.)

Nits, take or leave: the `Options.BuildAddDeletedChildrenRecord` doc at `:113-114` says the policy arg is ignored on the native branch, but `native_op.go:129-132` reads `FilterExpression` off it and copies the policy when set — harmless for the pruner's nil-filter policy, misleading for the next caller. No test covers the combined path's result-*parsing* side (all three cases pass `keys=nil`, so only record-building is exercised). And no case for a status that is neither `"OK"` nor `"ERROR"` — the one input where the `switch` falls through without touching any counter.

ctnguyen (approved):
Review at commit [3ea6eddb](https://github.com/ordishs/teranode/commit/3ea6eddb115dfc498aefd42ee05011b47335acf6)

This is a solid response to the last round — all four previous points are genuinely addressed in the tree, not just in the commit message. The fail-open handling is gone: `classifyParentUpdateResult` now treats a missing `SUCCESS` bin, an unreadable response shape, and an unrecognised status as failures with a message describing what it saw, and the combined and two-call paths share that one classifier so they can't drift apart again. The `updateOnly` policy override is the right call for the one native call site whose target is routinely absent, and both loops now tolerate `KEY_NOT_FOUND`, so a missing parent stays a skip rather than a resurrected partial record. The response parser reads `message` (with `errorMessage` as a tolerated alias), the labels are path-neutral, and the new unit tests cover the classifier, the tallies, the ordered-map shape, and the `UPDATE_ONLY` policy without leaking into the shared policy. The integration test also picked up the missing-parent case and a real `ParseLuaMapResponse` assertion. Nice work. What's left is one substantive interaction worth a look — the fail-closed change gives the combined path a new way to skip external-file cleanup — plus a few documentation and coverage residuals. None blocks merge, but the first one is close, so I've put it at the top.

## Status of previous points

| ChiR1 | Response parsing fails open | YES | FIXED |
| ChiR2 | Create-if-missing parent policy | YES | FIXED |
| ChiR3 | Wrong `message` response key | | FIXED |
| ChiR4 | Misleading UDF/Lua labels | | FIXED |

## Non-blocking issues

### ChiR5 - Fail-closed parent updates can leak external `.tx`/`.outputs` blobs on the combined path
**Problem:** On the defensive-off (combined) path, `flushCleanupBatches` runs parent updates and record deletions in one non-atomic `BatchOperate` via `executeBatchCleanupCombined`, then only reaches `executeBatchExternalFileDeletions` **after** it returns nil (`pruner_service.go:1591-1605`). When any parent update in the chunk fails, `executeBatchCleanupCombined` returns the parent error at `:1741-1742` — before external-file cleanup. But the pruned tx's own master record was already deleted in that same batch (`allDeletions`, built at `:1188`, then combined at `buildCombinedCleanupRecords:1686-1699`), and the parent that failed is a *different* record, so the delete lands even though the parent update did not. The external file was captured in memory during the scan (`:1174-1185`), so nothing re-reads Aerospike — but the master record is now gone, so the next DAH index scan never resurfaces this tx, and the blob store no longer self-expires blobs (`stores/blob/server.go:158`: "setCurrentBlockHeight removed - DAH cleanup now handled by pruner service"; external blobs are written `WithDeleteAt(0)` at `create.go:1096-1099` precisely because "the pruner service will manage" them). The `.tx`/`.outputs` file is orphaned permanently. One failed parent aborts the whole flush, so it orphans *every* external blob in that chunk, not just the affected tx. This early-return shape is **pre-existing** — a genuine `status=ERROR` response already triggered it — but this PR widens it: the fail-closed classifier now routes three formerly-benign outcomes (missing `SUCCESS` bin, unreadable shape, unknown status) into the same parent-error return, and pre-PR those counted as success and let external cleanup proceed.
**Why it matters:** On a native-op deployment whose dispatcher response drifts from what `addDeletedChildrenStatus` accepts (the unverified contract in ChiR7), every combined cleanup cycle now both errors *and* leaks external blobs, with no reclaim path — unbounded external-store growth. It is non-blocking only because the trigger is an already-malfunctioning server that also makes the pruner fail loudly every cycle (an operator will notice and can disable native ops), and the leak is disk, not consensus state.
**Fix:** Decouple external-file cleanup from parent-update success on the combined path — a failed parent update should not strand the blobs of records whose master deletes succeeded. Either run `executeBatchExternalFileDeletions` for the external files whose deletions succeeded before returning the parent error, or return per-record deletion status from `executeBatchCleanupCombined` so `flushCleanupBatches` can still delete those blobs:

```go
// flushCleanupBatches — don't let a parent-update error skip external cleanup
combinedErr := s.executeBatchCleanupCombined(ctx, parentUpdates, keys)
if len(externalFiles) > 0 {
    if delErr := s.executeBatchExternalFileDeletions(ctx, externalFiles); delErr != nil && combinedErr == nil {
        combinedErr = delErr
    }
}
return combinedErr
```

(guarding so external deletion is attempted only for records whose master delete actually landed).

### ChiR6 - `executeBatchParentUpdates` godoc still describes the pre-native tradeoff it now reverses
**Problem:** The dispatcher's doc comment at `pruner_service.go:1783-1788` still reads "Uses a Lua UDF (addDeletedChildren) instead of BatchWrite+MapPutItemsOp so that missing parent records are handled server-side **without generating KEY_NOT_FOUND errors in Aerospike client metrics**" and "Missing parents are silently skipped by the Lua UDF (TX_NOT_FOUND)". On a native-op deployment the parent update is now a wire-op-200 `BatchWrite` with `RecordExistsAction = UPDATE_ONLY` (`native_op.go:170-171`, `pruner_provider.go:73-82`), and the routine missing-parent case surfaces as a real `KEY_NOT_FOUND` batch result — exactly the client-metrics cost this comment claims the design avoids. The comment describes the opposite of what the code now does.
**Why it matters:** This is the load-bearing rationale a future reader will trust when deciding how the pruner treats missing parents; leaving it inverted hides a deliberate tradeoff. Client-side result-code accounting was measured as a real hot-path cost (see [bsv-blockchain/teranode#1001](https://github.com/bsv-blockchain/teranode/issues/1001), the motivation for the `aerospike_enable_client_metrics` switch in [#1363](https://github.com/bsv-blockchain/teranode/pull/1363)), so choosing correctness (no resurrected parent) at the price of `KEY_NOT_FOUND` noise on the native path is a decision worth stating rather than contradicting. The choice itself is right — this is only about the stale prose.
**Fix:** Rewrite the comment to match the two paths that actually run and name the tradeoff, e.g.:

```go
// On a native-op deployment the update is a wire-op-200 BatchWrite with
// UPDATE_ONLY; on a UDF deployment it is addDeletedChildren, which guards on
// aerospike:exists(rec). A missing parent (the routine case) is a skip either
// way — KEY_NOT_FOUND on the native path (which does count against client-side
// result-code metrics), an ERROR/TX_NOT_FOUND response map on the UDF path.
```

### ChiR7 - The native `addDeletedChildren` path has no CI verification
*Same as [@oskarszoon's point 2 (29-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1366#pullrequestreview-4807642800).*
**Problem:** `TestNativeOp_AddDeletedChildren_Integration` does the right thing when it runs — it asserts the real dispatcher's response parses via `ParseLuaMapResponse` with `status == "OK"` and that `UPDATE_ONLY` leaves a missing parent uncreated. The problem is that it only runs against a Teraspike build: `runAerospikeTestContainer` (`container_helper_internal_test.go:11-24`) falls back to a stock image when `AEROSPIKE_CONTAINER_IMAGE` is unset, the native-op probe then fails, and the test hits `t.Skipf` at `:109`. That env var is configured nowhere in the repo or CI, so in every automated run those real-server assertions skip and the native path's behaviour against the fork encoding goes unverified. oskarszoon's other two sub-points here were addressed (the missing-parent subtest was added, and `sumNamespaceStat` now requires the stat to exist); this CI-configuration gap is the residual. It also lines up with SonarCloud's 64.3% new-code coverage on the PR.
**Why it matters:** If the fork's real response shape diverges from what the parser accepts, every successful native parent update would be classified as a failure (the intended, safe fail-closed direction) and the cleanup batch would error on each pruning cycle — but no automated run would catch that before deploy, because the one test that would is always skipped. Fail-closed contains the blast radius to loud errors rather than silent corruption, which is why this is non-blocking, but the gap is worth recording.
**Fix:** Note in the PR body that the native `addDeletedChildren` path is verified only when `AEROSPIKE_CONTAINER_IMAGE` points at a Teraspike build, and if a fork image is reachable in CI, wire it into the aerospike-tagged job so the integration test actually runs rather than skips.

### ChiR8 - Combined-cleanup test covers routing but not the `parentEnd` split or result processing
*Same as [@oskarszoon's point "no test covers the combined path's result-parsing side" (29-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1366#pullrequestreview-4807642800).*
**Problem:** All three cases in `combined_cleanup_native_op_test.go` call `s.buildCombinedCleanupRecords(updates, nil)` with `keys == nil`, so they only exercise which record type the parent updates route to. The child-deletion region is never built, the `parentEnd` boundary is never checked against a non-empty deletion tail, and `executeBatchCleanupCombined` — where `parentEnd` splits the results into `batchRecords[:parentEnd]` (tally as parents) and `batchRecords[parentEnd:]` (tally as deletions) — has no unit coverage at all, since it needs a live `BatchOperate`. `tallyParentUpdateResults` and `tallyChildDeletionResults` are each well covered in isolation, but their composition behind the split is not.
**Why it matters:** An off-by-one or mis-ordered `parentEnd` would misclassify a child deletion as a parent update (or vice versa) and slip through the whole suite green — the combined path is the default (defensive-mode-off) hot path, so this is the arrangement that actually runs in production.
**Fix:** Add a case that passes both `updates` and `keys` and asserts the split, e.g.:

```go
records, parentEnd, _ := s.buildCombinedCleanupRecords(makeParentUpdates(t, 2), []*aerospike.Key{k1, k2})
require.Equal(t, 2, parentEnd)
require.Len(t, records, 4)
for i := parentEnd; i < len(records); i++ {
    require.False(t, isUDF(records[i]), "child deletions must not be UDF records")
}
```

## Recap

| ChiR5 | External blob leak on combined path | | 45% |
| ChiR6 | Stale parent-update godoc | | 30% |
| ChiR7 | Native path unverified in CI | | 25% |
| ChiR8 | Combined-path split untested | | 15% |

PR #1367, opened 2026-07-28
fix(p2p): key gossip peer maps by canonical hash, parse subtree hash before use

gokutheengineer:
**Closes bitcoin-sv/teranode#4726.**

### Problem

The issue reported two defects in `handleBlockTopic`: the announcing peer was registered with a nil block hash, and the raw (unvalidated, up-to-32KB) message hash string was stored as a `blockPeerMap` key before any validation.

Since the issue was filed, PR bsv-blockchain/teranode#1201 added `sanitizeAdvertisedTip`, which parses and validates the hash before `addPeer` and `storePeerMapEntry` run - so the nil-hash registration and the unbounded-garbage-key vectors are already closed on main, and those items are skipped.

What remained of the issue: the peer maps were still keyed by the **raw** message string rather than the canonical form. `chainhash.NewHashFromStr` accepts non-canonical hex (uppercase, truncated, odd length), while the ban lookups - `ReportInvalidBlock`, `processInvalidBlockMessage`, `ReportInvalidSubtree` - use the canonical `hash.String()` produced by the validation services. A peer announcing an invalid block or subtree with e.g. an uppercase hash string therefore created a map entry that never matched the lookup, evading the invalid-block/invalid-subtree ban.

The fresh-eyes review also found that `handleSubtreeTopic` still had the exact use-before-parse pattern the issue describes: the raw, unvalidated hash string was forwarded to WebSocket subscribers and credited as peer activity before `parseHash` ran.

### Fix

- `handleBlockTopic` and `handleSubtreeTopic` now key `blockPeerMap` / `subtreePeerMap` by the canonical `hash.String()` (the issue's proposed fix; the hash is already parsed at that point on all paths).
- `handleSubtreeTopic` now parses the hash before any use, mirroring `handleBlockTopic`: a malformed hash is rejected before the WebSocket notification, `updatePeerLastMessageTime`, and `updateBytesReceived`, and the notification carries the canonical hash.
- Updated the map-key comments in `Server.go` and `docs/references/services/p2p_reference.md` to state the canonical-key contract.

### Residual

Pre-existing behavior deliberately left unchanged (flagged by review, out of scope for this issue):

- Degenerate-but-hex strings (`""`, `"abc"`) are normalized by `chainhash.NewHashFromStr` (zero-padded) rather than rejected at ingress; an empty announce hash still normalizes to the zero hash, as it did before.
- Readers (`getPeerFromMap` and the `Delete` call sites) do not re-normalize their input; all producers were verified to pass canonical `hash.String()` values.
- `storePeerMapEntry` is last-writer-wins, so the most recent announcer of a hash absorbs the ban.

### Tests

Added in `services/p2p/server_helpers_test.go`:

- `TestHandleBlockTopic_PeerMapKeyedByCanonicalHash` - uppercase announce is stored under the canonical key, not the raw string.
- `TestHandleSubtreeTopic_PeerMapKeyedByCanonicalHash` - same for the subtree map, plus the notification carries the canonical hash.
- `TestHandleBlockTopic_NonCanonicalAnnouncerStillBanned` - end-to-end: uppercase announce, then `processInvalidBlockMessage` with the canonical hash bans the announcer and removes the entry.
- `TestHandleSubtreeTopic_RejectsMalformedHash` - malformed hash produces no notification, no map entry, no peer-activity credit.
- Strengthened `TestHandleBlockTopic_RejectsMalformedAdvertisedHash` to assert the map stays empty.

Run (affected package only):

```
go vet ./services/p2p/
```

Both green.

### DIFF
--- a/docs/references/services/p2p_reference.md
+++ b/docs/references/services/p2p_reference.md
@@ -38,8 +38,8 @@ type Server struct {
     invalidSubtreeTopicName           string             // Kafka topic for invalid subtrees
     nodeStatusTopicName               string             // pubsub topic for node status messages
     topicPrefix                       string             // Chain identifier prefix for topic validation
-    blockPeerMap                      sync.Map           // Map to track which peer sent each block (hash -> peerMapEntry)
-    subtreePeerMap                    sync.Map           // Map to track which peer sent each subtree (hash -> peerMapEntry)
+    blockPeerMap                      sync.Map           // Map to track which peer sent each block (canonical chainhash.Hash.String() -> peerMapEntry)
+    subtreePeerMap                    sync.Map           // Map to track which peer sent each subtree (canonical chainhash.Hash.String() -> peerMapEntry)
     startTime                         time.Time          // Server start time for uptime calculation
     peerRegistry                      *PeerRegistry      // Central registry for all peer information
     peerSelector                      *PeerSelector      // Stateless peer selection logic
--- a/services/p2p/Server.go
+++ b/services/p2p/Server.go
@@ -134,8 +134,8 @@ type Server struct {
 	invalidSubtreeTopicName           string                         // Kafka topic for invalid subtrees
 	nodeStatusTopicName               string                         // pubsub topic for node status messages
 	topicPrefix                       string                         // Chain identifier prefix for topic validation
-	blockPeerMap                      sync.Map                       // Map to track which peer sent each block (hash -> peerMapEntry)
-	subtreePeerMap                    sync.Map                       // Map to track which peer sent each subtree (hash -> peerMapEntry)
+	blockPeerMap                      sync.Map                       // Map to track which peer sent each block (canonical chainhash.Hash.String() -> peerMapEntry)
+	subtreePeerMap                    sync.Map                       // Map to track which peer sent each subtree (canonical chainhash.Hash.String() -> peerMapEntry)
 	startTime                         time.Time                      // Server start time for uptime calculation
 	peerRegistry                      blockchain.PeerRegistryClientI // gRPC client for the centralized peer registry hosted by the blockchain service
 	peerSelector                      *PeerSelector                  // Stateless peer selection logic
--- a/services/p2p/server_helpers.go
+++ b/services/p2p/server_helpers.go
@@ -203,13 +203,20 @@ func (s *Server) handleSubtreeTopic(_ context.Context, m []byte, fromID string) 
 
 	s.logger.Debugf("[handleSubtreeTopic] received subtree %s from %s", subtreeMessage.Hash, subtreeMessage.PeerID)
 
+	// Parse the hash before any use, mirroring handleBlockTopic: a malformed
+	// hash must not reach WebSocket subscribers or count as peer activity.
+	hash, err = s.parseHash(subtreeMessage.Hash, "handleSubtreeTopic")
+	if err != nil {
+		return
+	}
+
 	now := time.Now().UTC()
 
 	select {
 	case s.notificationCh <- &notificationMsg{
 		Timestamp:  now.Format(isoFormat),
 		Type:       "subtree",
-		Hash:       subtreeMessage.Hash,
+		Hash:       hash.String(),
 		BaseURL:    subtreeMessage.DataHubURL,
 		PeerID:     subtreeMessage.PeerID,
 		ClientName: subtreeMessage.ClientName,
@@ -232,12 +239,6 @@ func (s *Server) handleSubtreeTopic(_ context.Context, m []byte, fromID string) 
 
 	// Skip notifications from unhealthy peers
 	if s.shouldSkipUnhealthyPeer(fromID, "handleSubtreeTopic") {
-		return
-	}
-
-	hash, err = s.parseHash(subtreeMessage.Hash, "handleSubtreeTopic")
-	if err != nil {
-		s.logger.Errorf("[handleSubtreeTopic] error parsing hash: %v", err)
 		return
 	}
 
--- a/services/p2p/server_helpers.go
+++ b/services/p2p/server_helpers.go
@@ -108,8 +108,12 @@ func (s *Server) handleBlockTopic(_ context.Context, m []byte, fromID string) {
 
 	now := time.Now().UTC()
 
-	// Store the peer ID that sent this block
-	s.storePeerMapEntry(&s.blockPeerMap, blockMessage.Hash, fromID, now)
+	// Store the peer ID that sent this block, keyed by the canonical hash
+	// string. Ban lookups (ReportInvalidBlock, processInvalidBlockMessage) use
+	// hash.String() from block validation, so keying by the raw message string
+	// would let a peer evade the invalid-block ban by announcing a
+	// non-canonical hex form (uppercase, truncated).
+	s.storePeerMapEntry(&s.blockPeerMap, hash.String(), fromID, now)
 
 	s.logger.Debugf("[handleBlockTopic] storing peer %s for block %s", fromID, blockMessage.Hash)
 
@@ -237,8 +241,11 @@ func (s *Server) handleSubtreeTopic(_ context.Context, m []byte, fromID string) 
 		return
 	}
 
-	// Store the peer ID that sent this subtree
-	s.storePeerMapEntry(&s.subtreePeerMap, subtreeMessage.Hash, fromID, now)
+	// Store the peer ID that sent this subtree, keyed by the canonical hash
+	// string so the ReportInvalidSubtree lookup (which uses hash.String() from
+	// subtree validation) matches even when the announcer sent a non-canonical
+	// hex form.
+	s.storePeerMapEntry(&s.subtreePeerMap, hash.String(), fromID, now)
 	s.logger.Debugf("[handleSubtreeTopic] storing peer %s for subtree %s", fromID, subtreeMessage.Hash)
 
 	if s.subtreeKafkaProducerClient != nil { // tests may not set this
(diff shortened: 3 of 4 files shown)


ctnguyen (approved):
Review at commit [4b90612](https://github.com/gokutheengineer/teranode/commit/4b90612db102aac18e56b092232370395c39a55c)

Nice, tight fix, and the diagnosis is right: the lookup side keys off the canonical `hash.String()` the validation services emit, so a raw-string key left an `ABCD…` announcer simply unfindable. I checked both reader paths and both Kafka producers — they really are canonical, so the key change closes the loop. Parsing at the top of `handleSubtreeTopic` is the better half of the change, and the PR body is refreshingly straight about which parts of issue 4726 #1201 had already closed and what residual behavior it leaves alone. Nothing below blocks merge.

## Non-blocking issues

### ChiR1 - Malformed announce hash costs the sender nothing
**Problem:** Both handlers treat an unparseable hash as a silent drop. `handleSubtreeTopic` returns at `server_helpers.go:217-220` and `handleBlockTopic` returns on `sanitizeAdvertisedTip` failure (`server_helpers.go:85-88`), neither calling `addProtocolViolation` — inconsistent with the sibling malformed-field checks in the same functions, where peer-ID spoofing (`:189`) and an invalid `DataHubURL` (`:202`) both apply ban score. The reorder makes it starker on the subtree side: the early return now sits ahead of `updateBytesReceived`, so a peer flooding malformed-hash announcements is no longer even visible in the bandwidth counters it is consuming.
**Why it matters:** A sender of syntactically invalid gossip can repeat it indefinitely at zero reputational cost and, for subtrees, with no accounting trail either.
**Fix:** Score the parse failure on both paths. The block side is unconditional — `sanitizeAdvertisedTip` is only reached on the non-self branch, so there is no self-ban risk:
```go
advertisedHeight, hash, ok = s.sanitizeAdvertisedTip(blockMessage.PeerID, blockMessage.Height, blockMessage.Hash, s.getLocalHeight())
if !ok {
    s.addProtocolViolation(fromID)
    return
}
```
The subtree parse now sits *above* the `isOwnMessage` return, so guard it so a malformed own message can't self-ban:
```go
hash, err = s.parseHash(subtreeMessage.Hash, "handleSubtreeTopic")
if err != nil {
    if !s.isOwnMessage(fromID, subtreeMessage.PeerID) {
        s.addProtocolViolation(fromID)
    }
    return
}
```
Then extend `TestHandleBlockTopic_RejectsMalformedAdvertisedHash` and `TestHandleSubtreeTopic_RejectsMalformedHash` to assert the violation is scored. Note both currently use a peer that was never registered, and `AddBanScore` mirrors the score onto `PeerInfo` only for peers already in the registry (`services/blockchain/peer_registry.go:788-791`) — so register the peer (or land one valid announce) first, then assert `info.BanScore` is positive.

### ChiR2 - Canonical-key contract is enforced only by comment
**Problem:** The invariant now lives in three prose comments (`Server.go:137-138`, `server_helpers.go:113-117`, `:254-257`) and the doc page, but the helpers still take a bare `string`: `storePeerMapEntry(peerMap *sync.Map, hash string, ...)` (`:987`) and `getPeerFromMap(peerMap *sync.Map, hash string, ...)` (`:996`). Readers pass whatever arrives on the wire — `processInvalidBlockMessage` forwards `invalidBlockMsg.GetBlockHash()` (`:657`, `:678`) and `ReportInvalidSubtree` forwards `m.SubtreeHash` from Kafka (`Server.go:2095`, `:2129`). Today's in-tree producers are canonical, so this is latent, not live; but a new producer, an external publisher on the same topic, or a hand-crafted report re-opens exactly the mismatch this PR fixes, and it fails silently as a "no peer found" warning.
**Why it matters:** The failure mode is invisible and asymmetric — a missed lookup on the block map skips an `AddBanScore`, and on the subtree map it skips the failed-interaction reputation hit in `ReportInvalidSubtree` — so a compiler-enforceable invariant shouldn't depend on every future caller reading a comment.
**Fix:** Make the writer take the parsed type and have the readers normalize before lookup, which also validates their input:
```go
func (s *Server) storePeerMapEntry(peerMap *sync.Map, hash *chainhash.Hash, from string, timestamp time.Time)
func (s *Server) getPeerFromMap(peerMap *sync.Map, hash *chainhash.Hash, mapType string) (string, error)
```
then in `processInvalidBlockMessage` / `ReportInvalidSubtree` parse the incoming string once and use the result for both the lookup and the `Delete`.

### ChiR3 - Post-store log lines print the raw string, not the key that was stored
**Problem:** `server_helpers.go:120` and `:259` still log `blockMessage.Hash` / `subtreeMessage.Hash` immediately after storing under `hash.String()`. For the very case this PR is about — a non-canonical announce — the log says `storing peer <id> for block ABCD…` while the key is `abcd…`. Same mismatch in the channel-full warnings (`:102`, `:234`) and at `:125`.
**Why it matters:** These are the lines an operator greps when an attribution goes missing, and they no longer correspond to the map contents.
**Fix:** Log the canonical form on the store-adjacent lines, keeping the raw string only where the point is "what the peer actually sent" (`:74`, `:208`):
```go
s.logger.Debugf("[handleSubtreeTopic] storing peer %s for subtree %s", fromID, hash.String())
```

### ChiR4 - Test gaps: truncated hex, and the subtree reader path
**Problem:** Two gaps. First, both code comments call out "uppercase, truncated" and `parseHash` really does accept truncation — `chainhash.Decode` errors only above 64 chars, left-zero-pads odd lengths, and zero-fills short input, so `""` becomes the zero hash and `"abc"` becomes `0…0abc` — yet the new tests only cover `strings.ToUpper(canonical)`. Second, the block side gets a genuine end-to-end test through `processInvalidBlockMessage`, but the subtree side only asserts a direct `getPeerFromMap` lookup and the notification hash; `ReportInvalidSubtree` — the actual consumer of `subtreePeerMap` — is never exercised after a non-canonical announce.
**Why it matters:** The zero-padding path is the surprising one and is listed as deliberately unchanged residual behavior, so nothing pins it in either direction; and the subtree half of the fix has no test proving the reader it exists for actually resolves the peer.
**Fix:** Add a truncated-announce case:
```go
short := "abc"
want, err := chainhash.NewHashFromStr(short)
require.NoError(t, err)
// ... announce with Hash: short, then:
_, err = s.getPeerFromMap(&s.subtreePeerMap, want.String(), "subtree")
require.NoError(t, err)
```
and mirror `TestHandleBlockTopic_NonCanonicalAnnouncerStillBanned` on the subtree side: uppercase announce, then `require.NoError(t, s.ReportInvalidSubtree(ctx, canonical, "", "reason"))`, asserting the peer picked up the failed interaction and that `s.subtreePeerMap.Load(canonical)` is gone.

### ChiR5 - `hash.String()` recomputed several times per gossip message
**Problem:** `handleBlockTopic` calls `hash.String()` five times (`:95`, `:118`, `:146`, `:151`, `:160`) and `handleSubtreeTopic` four (`:228`, `:258`, `:263`, `:275`); the map-key call is new in this diff. Each call reverses the 32 bytes and hex-encodes into a fresh 64-byte string, and the `Debugf` at `:151` pays for it even when debug logging is off, because arguments are evaluated eagerly.
**Why it matters:** Subtree and block announcements are the highest-rate gossip path in the node; small per message but pure waste, and hoisting it also removes the chance of the key and the Kafka payload drifting apart later.
**Fix:** Compute once next to the parse and reuse:
```go
hashStr := hash.String()
...
s.storePeerMapEntry(&s.subtreePeerMap, hashStr, fromID, now)
```

## Recap

| ChiR1 | Malformed hash unpenalized, unaccounted |  | 38% |
| ChiR2 | Canonical-key contract not type-enforced |  | 32% |
| ChiR3 | Store logs show raw hash |  | 18% |
| ChiR4 | Truncated hex and subtree reader untested |  | 15% |
| ChiR5 | Repeated `hash.String()` on hot path |  | 10% |

liam (commented):
Traced every writer and reader of `blockPeerMap`/`subtreePeerMap` plus the upstream ban-report producers. **The fix genuinely closes the ban-evasion - no blocking issues.**

- **Write/read sides now agree on canonical keys.** The only two production writers (`handleBlockTopic` `server_helpers.go:112`, `handleSubtreeTopic` `:241`) both go through `storePeerMapEntry` and are patched to `hash.String()`. Every ban-lookup reader already keys canonically: `processInvalidBlockMessage`/`ReportInvalidBlock` from `block.Hash().String()` (`BlockValidation.go:2125`), `ReportInvalidSubtree` from `subtreeHash.String()` (`subtreevalidation/Server.go:1048`, all callers). No other map writers exist, so the non-canonical announce key was the sole mismatch and it's removed.
- **`handleSubtreeTopic` parse-before-use reorder is correct** - `parseHash` now precedes the WebSocket notify, `updatePeerLastMessageTime`, and `updateBytesReceived`; no use remains before the parse, and the notification now carries the canonical hash.
- **Edge cases are pre-existing and benign** - empty/short hex normalizing to the zero hash can never match a real ban lookup; last-writer-wins is unchanged and can't help an attacker dodge their own ban.

**Non-blocking notes:**
1. Blocks have a true end-to-end ban test (`TestHandleBlockTopic_NonCanonicalAnnouncerStillBanned`), but the subtree side only asserts keying, not the full `ReportInvalidSubtree` -> peer-banned path. A subtree e2e ban test would fully mirror the block guarantee.
2. Out of scope for this issue, but worth a separate follow-up: a distinct evasion remains where a peer announces a hash that *differs* from the real object's hash (not merely a non-canonical encoding of it) - the map key (announced) and ban key (fetched object's real hash) never match. This PR neither introduces nor worsens it.

liam (approved):

PR #1369, opened 2026-07-28
fix(p2p): re-validate resolved IPs when fetching peer-supplied URLs

gokutheengineer:
**Refs bitcoin-sv/teranode#4724** — deliberately not `Closes`: see *Scope* below for the one
address class in the issue this does not block.

### Problem

`validateDataHubURL` (`services/p2p/server_helpers.go`) only inspects IP literals, so a
peer-supplied hostname that resolves to a loopback, link-local or cloud-metadata address
passes validation. Confirmed still valid on `main`.

Where that mattered depended on the client:

- **Peer availability probe** (`peer_selector.go` → `health.CheckHTTPServer`): a plain
  `http.Client` with no dial-time check. This was live — a peer could make the node issue
  HTTP requests to `169.254.169.254`, `127.0.0.1` or any internal host just by advertising a
  hostname, with the probe firing automatically during peer selection.
- **Subtree meta regeneration** (`model/subtree_meta_regenerator.go`): also completely
  unguarded (no `ValidateURL`, no dial check) when this branch started, and reached with
  peer-supplied `baseURL`s from `BlockValidation.go`. Not named in the issue, but the stronger
  primitive of the two since it parses the response body. `#1577` has since routed that fetch
  through `util.DoHTTPRequestBodyReaderWithRetry` — the shared guarded client — which fixes it
  upstream and better than this branch did (pooling plus 503 retry), so on merging that commit
  this branch dropped its own client there and kept only the regression test.
- **Block/subtree fetches** via `util.DoHTTPRequest*`: already had a dial-time guard, so no
  change was needed there.

### Fix

The issue proposes a re-validating `DialContext`, which is the right approach; the guard
already existed in `util/http.go` for one client, so rather than duplicate it this exposes it:

- `util.NewSSRFSafeDialContext(policy)` and `util.NewSSRFSafeHTTPClient(timeout, policy)`,
  with `util.DefaultSSRFDialPolicy` as the shared policy. The dialer resolves the hostname,
  rejects the dial if **any** resolved address is unsafe (so a mixed public/internal answer
  can't smuggle a target through failover), then connects to the already-validated IP rather
  than re-resolving — closing the DNS-rebinding TOCTOU window. `CheckRedirect` re-checks each
  hop, so a peer can't bounce the request to an internal address or off http/https.
- The p2p peer probe uses that client, wired into `#1347`'s `checkHealth` seam. `util/health`
  gained `CheckPeerHTTPServer`, which **requires** a guarded client and fails closed on nil
  rather than defaulting to an unguarded one, so a future peer probe cannot silently lose the
  guard by calling the plain constructor.
- One redirect rule for the threat: `ssrfCheckRedirect` is shared by the package client and
  every `NewSSRFSafeHTTPClient`, checking hop count, scheme, userinfo and blocked IP literals.
  Previously the two clients carried divergent redirect checks (`ValidateURL` vs the dial
  policy) for the same threat; a test asserts both now refuse the same targets with the same
  reason.
- Because dialing by validated IP loses `net.Dialer`'s dual-stack fast fallback, each address
  gets a slice of the remaining deadline, floored at 500ms and clamped to what remains.
  Without the slice a blackholed first address would consume the whole 2s probe budget; without
  the floor a peer with four A records would give the working address only 500ms where a plain
  sequential dial would have let it use the lot.
- `checkPeerAvailability` now surfaces the status message on a non-2xx response, which
  previously logged as `unhealthy: <nil>`.

### Scope — one class from the issue is deliberately not blocked

The issue lists three address classes reachable through the hostname bypass; this blocks
loopback, link-local (the cloud-metadata range) and unspecified, but **permits RFC1918/ULA**,
so it is filed as `Refs` rather than `Closes`.

The reason is that the probe only decides whether to fetch from a peer, and the block/subtree
fetch path allows private ranges by explicit documented design (`util/http.go`) because peers,
k8s pods and private miner interconnects live on private networks. A probe stricter than the
fetch it gates buys no security — the richer primitive stays open — while costing availability:
a node whose peers resolve into private space would find *no* sync peer at all, since the probe
failure drops them from both the full-node and pruned candidate lists.

Blocking RFC1918 for peer-supplied URLs is therefore a repo-wide policy decision that has to
change the fetch path too, with its own deployment review. It is not something this PR should
do unilaterally. A test (`TestPeerProbeAllowsPrivateAddresses`) pins the probe policy to the
fetch policy so the two cannot drift.

Also updated the text that now misdescribed behaviour: the `AllowPrivateIPs` and
`HealthCheckEnabled` longdescs (the former's security scope was undocumented, the latter still
claimed the probe hits `/health`), and the "stateless, pure-function" description of
`PeerSelector`, which now holds an HTTP client.

### Residual

- With `HTTP_PROXY`/`HTTPS_PROXY` set, the transport dials the proxy, so the address check
  sees the proxy rather than the peer target and the proxy fetches on the node's behalf. The
  new clients inherit `http.DefaultTransport`'s proxy behaviour rather than diverging from
  `util`'s existing shared fetch client, which has always worked this way; setting
  `Proxy = nil` on only the two new clients would leave the largest peer-fetch surface
  proxied while these two were not. Now stated in the code comment, the `AllowPrivateIPs`
  longdesc and the settings doc instead of being claimed away.
- `validateDataHubURL` stays a static pre-filter and deliberately performs no DNS resolution —
  doing so would let a peer trigger a blocking lookup per announcement. Resolution-time
  enforcement lives in the dialers instead; the doc comment now says so.
- `p2p_allow_private_ips` short-circuits the **static** `validateDataHubURL` check entirely,
  so with it set an announced loopback URL is accepted into the registry — but the
  connection-time guard still refuses loopback, so such a peer is accepted and then never
  selected. That asymmetry is deliberate (no configuration should let a peer aim the node at
  localhost or at the metadata endpoint) and is now stated in the setting's longdesc, the
  settings doc and a test rather than being described as symmetric, which it is not. Single-host
  loopback sync would not work regardless: the fetch path refuses loopback independently of
  this setting.

### Tests

New regression coverage, each verified to fail against the pre-fix code and pass after:

- `services/p2p/peer_selector_ssrf_test.go` — a peer hostname resolving to loopback is
  refused after resolution and the target server sees **zero** requests (pre-fix: probed
  successfully); every blocked class driven through the real client in IPv4 and IPv6 form
  (`[::1]`, `[fe80::1]`, `0.0.0.0`); private addresses reaching the dial rather than being
  refused, under both settings of `AllowPrivateIPs` (verified to fail if the private block is
  reintroduced); the deliberate probe-vs-static asymmetry under the flag; the probe still
  working against a reachable peer (right URL join, 2xx only); empty and malformed URLs; and a
  test pinning that the static check accepts peer hostnames, which is why the dial guard has
  to exist.
- `model/subtree_meta_regenerator_test.go` — peer fetch refused for a hostname resolving to
  loopback and for a metadata literal, with the target serving valid data and recording no
  hits (pre-fix: fetched successfully). The three pre-existing httptest-based tests use a
  helper that toggles the documented `SetSSRFProtection` escape hatch.
- `util/http_test.go` — caller-supplied policy enforced at resolution time; the global test
  toggle bypasses the guard without even consulting the policy; a blackholed first address
  still fails over to a reachable second one within the budget; the redirect guard rejecting
  link-local, loopback, non-http schemes and over-long chains; and a real request through
  `NewSSRFSafeHTTPClient` to a hostname resolving to loopback, proving the transport is
  wired to the guard and the target is never reached.

`services/p2p` includes `#1347`'s health-check tests, which this branch merged and integrated
with (the probe client is wired into that commit's `checkHealth` seam).

Scoped runs (affected packages only, all green):

```
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/p2p/
SETTINGS_CONTEXT=test go test -race -count=1 ./util/ ./util/health/
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./model/ ./settings/
```


ctnguyen (changes requested):
Review at commit [f7593d7](https://github.com/gokutheengineer/teranode/commit/f7593d71601fb22edf12e877b813f5f472b689fe)

Nice piece of work, and the right shape for the fix — exposing the guard that already lived in `util/http.go` beats writing a second dialer, and the two details that usually get botched are both handled: rejecting the dial if *any* resolved address is unsafe, and dialing the already-validated IP instead of re-resolving, which closes the rebinding TOCTOU rather than narrowing it. Finding the completely unguarded `subtree_meta_regenerator` path while working on the probe is the kind of thing that makes a security fix worth merging, since that one parses the response body. Tests are genuinely good too: verified to fail pre-fix, and asserting the target saw **zero** hits rather than just that an error came back. My findings are about the edges of the policy, not the core.

## Blocking issues

### ChiR1 - RFC1918-via-hostname is still reachable, so the issue must not be closed
**Problem:** `dataHubDialPolicy` (`services/p2p/peer_selector.go:332`) falls through to `util.DefaultSSRFDialPolicy`, which allows RFC1918/ULA. Issue #&NoBreak;4724 names `10.x` as one of three address classes reachable through the hostname bypass, and this PR is marked `Closes` on it. A peer advertising `peer.attacker.example` resolving into the node's private network still gets the node to `GET http://<internal-host>:<port>/api/v1/bestblockheader`, with the probe outcome acting as an observable reachability oracle through peer selection. The rationale given — that blocking private ranges in the probe alone would strand legitimate private peers the fetch path talks to happily — holds for hostname-based peers, but it is asymmetric with the node's own DataHubURL policy: `validateDataHubURL` → `isUnsafeIP` (`services/p2p/server_helpers.go:329`) already rejects RFC1918 **literals** at announcement time. So the project treats private DataHub targets as unsafe for literals while this change accepts them for hostnames — which is precisely the literal-vs-hostname gap the issue was filed about.
**Why it matters:** Merging as-is auto-closes a security issue with one of its three named classes still exploitable, and the internal-network probe primitive survives with no record anywhere that it was consciously accepted.
**Fix:** Either enforce it, and set `p2p_allow_private_ips=true` in the docker/k8s settings contexts that legitimately resolve to RFC1918:
```go
func (ps *PeerSelector) dataHubDialPolicy(ip net.IP) string {
	if ps.settings != nil && ps.settings.P2P.AllowPrivateIPs {
		if ip.IsLoopback() || ip.IsPrivate() {
			return ""
		}
		return util.DefaultSSRFDialPolicy(ip)
	}

	if ip.IsPrivate() {
		return "private address" // matches isUnsafeIP, used by validateDataHubURL
	}

	return util.DefaultSSRFDialPolicy(ip)
}
```
Or, if the deviation stands, drop the `Closes` linkage to `Refs` and post the accepted residual on the issue so it stays open and tracked. Documenting it only in this PR description is not enough — the issue is the artifact people will check.

### ChiR2 - The `AllowPrivateIPs` loopback carve-out relaxes the guard for no gain
**Problem:** `dataHubDialPolicy` returns `""` for loopback under `AllowPrivateIPs`, commented as "so single-host local deployments keep working". They don't: block/subtree fetches go through `util.DoHTTPRequest*` → the shared `httpClient` → `DefaultSSRFDialPolicy`, which blocks loopback unconditionally regardless of the setting (`services/blockvalidation/get_blocks.go:738,795,836,885`, `services/subtreevalidation/check_block_subtrees.go:140,279`). So with the default `asset_httpAddress = http://localhost:${ASSET_HTTP_PORT}/api/v1`, the probe reports the peer healthy, the coordinator selects it, and every fetch is then refused with `SSRF dial check: resolved address 127.0.0.1 ... is a loopback address`. What the carve-out does buy is a loopback-probe oracle: on a node with `p2p_allow_private_ips=true` — documented as the setting for isolated networks and local dev — a peer can make the node issue GETs to `127.0.0.1:<any port>` and infer reachability of localhost admin/RPC services from selection behaviour. This is not a regression versus `main`, where the probe was unguarded in every configuration; it is a relaxation that the fix deliberately keeps, with no functional benefit to weigh against it. The new `settings/p2p_settings.go` longdesc and `docs/references/settings/services/p2p_settings.md` both tell operators loopback targets are accepted, without mentioning that the data path still refuses them.
**Why it matters:** A security fix should not ship a hole it does not need, and the config documentation added here points operators at a setting that cannot deliver what it promises.
**Fix:** Delete the carve-out — `dataHubDialPolicy` collapses to `util.DefaultSSRFDialPolicy` and the method disappears with it — and drop the loopback sentences from the longdesc and the settings doc. If single-host loopback sync is genuinely a requirement, it has to be threaded through the fetch client too (pass a policy into the catchup fetch path rather than hardcoding `DefaultSSRFDialPolicy`); relaxing only the probe cannot make it work.

## Non-blocking issues

### ChiR3 - Proxy environments route around the dial policy while the docs say "never"
**Problem:** `NewSSRFSafeHTTPClient` (`util/http.go:179`) clones `http.DefaultTransport`, which carries `Proxy: ProxyFromEnvironment`, and `Clone()` preserves it. With `HTTP_PROXY`/`HTTPS_PROXY` set, the transport dials the **proxy** — so `NewSSRFSafeDialContext` validates the proxy's address and the proxy then fetches the peer-supplied target on the node's behalf, metadata endpoints included. The PR description lists this as a residual, but the code and the shipped docs do not: the `dataHubDialPolicy` comment and the new `AllowPrivateIPs` longdesc both state a peer "can never steer the node at a cloud metadata endpoint". The inverse also bites — a local egress proxy on `127.0.0.1` makes the policy reject *every* dial.
**Why it matters:** An operator reading the setting documentation gets an absolute guarantee that a single environment variable removes, and the two clients this PR adds are exactly the ones handling peer-supplied URLs.
**Fix:** For peer-supplied URLs, the direct dial is the honest default — peer block/subtree fetches are node-to-node traffic, not proxied web browsing:
```go
transport := http.DefaultTransport.(*http.Transport).Clone()
transport.DialContext = NewSSRFSafeDialContext(policy)
// Peer-supplied URLs must not be proxied: with a proxy configured the transport
// dials the proxy, so policy would validate the proxy rather than the target.
transport.Proxy = nil
```
If operators do need an egress proxy for peer fetches, keep `Proxy` and instead reword the guarantee in both the code comment and the longdesc to "unless an HTTP proxy is configured", so the claim matches the code.

### ChiR4 - Hard-partitioning the deadline can mark a reachable multi-address peer unhealthy
**Problem:** `dialAttemptContext` (`util/http.go:122`) splits the remaining deadline evenly across remaining candidates. Under the 2s probe budget, a peer whose hostname resolves to three or four addresses gets ~500-650ms per attempt after DNS — below a single TCP handshake RTT for a genuinely distant peer. Every attempt then times out and the peer is reported unhealthy on every selection round, permanently, even though it is reachable. Go's own dialer avoids this by staggering attempts concurrently (Happy Eyeballs, ~300ms offset) rather than partitioning, precisely because a partition converts "slow" into "unreachable". The shared fetch client is unaffected in practice — its 60s/5min budgets exceed `ssrfSafeDialer.Timeout` anyway — so this bites only the path the split was added for.
**Why it matters:** A silent, self-reinforcing false negative: the peer is excluded from sync forever, and the only symptom is a `Debugf` line, so it will not be diagnosed from production logs.
**Fix:** Floor the per-attempt slice so the split can only shorten an attempt that still has room for a handshake:
```go
const minDialAttemptBudget = 750 * time.Millisecond

func dialAttemptContext(ctx context.Context, remainingCandidates int) (context.Context, context.CancelFunc) {
	deadline, ok := ctx.Deadline()
	if !ok || remainingCandidates <= 1 {
		return context.WithCancel(ctx)
	}

	remaining := time.Until(deadline)

	budget := remaining / time.Duration(remainingCandidates)
	if budget < minDialAttemptBudget {
		budget = minDialAttemptBudget
	}

	if budget <= 0 || budget >= remaining {
		return context.WithCancel(ctx) // nothing to gain from splitting
	}

	return context.WithTimeout(ctx, budget)
}
```

### ChiR5 - A fresh `http.Transport` per regenerator, created once per block validation
**Problem:** `NewSubtreeMetaRegenerator` (`model/subtree_meta_regenerator.go:60`) now calls `util.NewSSRFSafeHTTPClient`, which does `http.DefaultTransport.(*http.Transport).Clone()`. Previously `&http.Client{Timeout: 30 * time.Second}` used the shared `http.DefaultTransport`, so its connection pool was reused across regenerators. `createMetaRegenerator` runs per block (`services/blockvalidation/BlockValidation.go:1792,1891,2342`), so every validated block now gets a private pool: reuse survives within one block's subtrees but is lost between blocks, and each transport that opened a connection holds it idle for `IdleConnTimeout` (90s). Separately, that type assertion is unchecked and now runs at request time rather than at package init, so it panics if `http.DefaultTransport` is ever replaced by a non-`*http.Transport` (plain `httpmock.Activate()`). Nothing in the tree does that today — every call site uses `ActivateNonDefault(util.HTTPClient())` — but the assertion moved from a place where it was impossible to a place where it isn't.
**Why it matters:** Extra TCP/TLS handshakes per block on the recovery path during catchup, plus a latent panic in a constructor block validation calls for every block.
**Fix:** Build the guarded client once and share it — a package-level `var peerFetchClient = util.NewSSRFSafeHTTPClient(peerFetchTimeout, util.DefaultSSRFDialPolicy)` in `model`, or accept an `*http.Client` in the constructor so `BlockValidation` holds one. While there, make the clone defensive:
```go
base, ok := http.DefaultTransport.(*http.Transport)
if !ok {
	base = &http.Transport{}
}
transport := base.Clone()
```

### ChiR6 - Regenerator tests reach for the process-global SSRF toggle
**Problem:** `allowLoopbackPeers` (`model/subtree_meta_regenerator_test.go:29`) flips `util.SetSSRFProtection(false)` for the whole process so three pre-existing tests can talk to their httptest servers. That toggle also short-circuits `util.ValidateURL` and every other guarded client in the binary, and the cleanup restores a hardcoded `true` rather than the prior value. It is needed only because `NewSubtreeMetaRegenerator` hardcodes its client and policy with no seam.
**Why it matters:** A test-only global that disables the guard everywhere can mask a genuine SSRF regression in an unrelated test in the same package, and the hardcoded restore is a trap if the default ever becomes `false`.
**Fix:** Give the constructor the seam ChiR5 already wants — accept an `*http.Client` (or a `util.SSRFDialPolicy`) — and have these three tests pass a loopback-permitting policy, keeping the relaxation scoped to the tests that need it and leaving `TestSubtreeMetaRegenerator_RejectsInternalPeer` on the real default. If the global stays for now, save and restore the previous value instead of assuming `true`.

### ChiR7 - `SSRFDialPolicy`'s doc comment describes the opposite of the implemented policy
**Problem:** The comment on `SSRFDialPolicy` (`util/http.go:37`) says "the p2p peer health check, for example, must also reject RFC1918 addresses to match its DataHubURL validation" — the opposite of what `dataHubDialPolicy` does in this same PR, and of the deviation the PR description argues for. It reads like a leftover from an earlier draft.
**Why it matters:** That comment is the first thing a maintainer adding a third guarded client will read, and it will send them to the wrong conclusion about what the shared policy is for.
**Fix:** Reword it to describe the actual escape hatch, e.g. "…so a caller can layer extra ranges on top of `DefaultSSRFDialPolicy`, or relax it — the p2p peer probe relaxes loopback under `p2p_allow_private_ips`." If ChiR1 or ChiR2 lands, update the example to whatever the policy then does.

## Recap

| ChiR1 | RFC1918 hostname bypass remains | YES | 70% |
| ChiR2 | Loopback carve-out with no benefit | YES | 55% |
| ChiR3 | Proxy bypasses dial policy | | 40% |
| ChiR4 | Dial budget split starves attempts | | 35% |
| ChiR5 | New transport per block | | 20% |
| ChiR6 | Global SSRF toggle in tests | | 15% |
| ChiR7 | Stale policy doc comment | | 10% |

ctnguyen (approved):
Review at commit [b15fa05](https://github.com/gokutheengineer/teranode/commit/b15fa057953ceb5b7e84eaadd22e48800edd8910)

Both blockers are properly closed, and closed the way I'd hoped rather than the cheap way. `dataHubDialPolicy` now rejects RFC1918/ULA by default so the probe enforces the same classes as `isUnsafeIP`, the loopback carve-out is gone entirely, and the new `TestDataHubDialPolicy` pins the two against each other class by class instead of just asserting a happy path. Nice touch that ChiR7 fixed itself: the `SSRFDialPolicy` doc comment that used to describe the opposite of the code is now simply true. The proxy residual went the honest route — the caveat is in the `NewSSRFSafeHTTPClient` comment, the `AllowPrivateIPs` longdesc and the settings doc, rather than being claimed away. I re-ran `go vet` over `util`, `services/p2p` and `model` and it's clean. What's left is small: three of the previous non-blocking points are untouched, and I found one operational sharp edge, one comment that claims a symmetry the code doesn't have, and one check the new clients dropped relative to the shared one.

## Status of previous points

ChiR4, ChiR5 and ChiR6 are unchanged in the code and there's no comment from you on them, so I've marked them NOT FIXED rather than assuming they were consciously declined — they were non-blocking then and still are, so this isn't a merge gate either way. Worth noting that ChiR6's pattern now also appears in the new `services/p2p/peer_selector_ssrf_test.go` (`allowLoopbackProbes`), so the process-global toggle spread rather than shrank.

| ChiR1 | RFC1918 hostname bypass remains | YES | FIXED |
| ChiR2 | Loopback carve-out with no benefit | YES | FIXED |
| ChiR3 | Proxy bypasses dial policy |  | FIXED |
| ChiR4 | Dial budget split starves attempts |  | NOT FIXED |
| ChiR5 | New transport per block |  | NOT FIXED |
| ChiR6 | Global SSRF toggle in tests |  | NOT FIXED |
| ChiR7 | Stale policy doc comment |  | FIXED |

## Non-blocking issues

### ChiR8 - A peer refused by the private-address policy looks exactly like a peer that timed out
**Problem:** With the defaults (`p2p_health_check_enabled=true`, `p2p_allow_private_ips=false`), a peer whose DataHubURL is a *hostname* resolving into RFC1918/ULA is now permanently ineligible: `isEligible` (`services/p2p/peer_selector.go:320-329`) drops it and logs a single `Debugf("Peer %s is unhealthy: %v")`. That's the intended consequence of ChiR1 and I'm not re-litigating it — but the fetch path disagrees, since block/subtree fetches and the meta regenerator both use `util.DefaultSSRFDialPolicy`, which allows private ranges (`util/http.go:147-151`, `model/subtree_meta_regenerator.go:60`). So the node refuses to *select* a peer it would happily *fetch* from, and the only evidence is a debug line whose text is identical in shape to a genuine connection timeout. `p2p_allow_private_ips=true` is the documented remedy, but it is coarse: `validateDataHubURL` short-circuits on it (`services/p2p/server_helpers.go:410-412`), so turning it on to keep private peers selectable also switches off the whole announcement-time static check. Only `compose/settings_test.conf` sets the flag today; `deploy/kubernetes/` and `deploy/docker/base/settings_local.conf` don't, and `daemon/test_daemon.go:160` sidesteps the whole thing with `SetSSRFProtection(false)`, so in-repo test coverage won't surface this either.
**Why it matters:** A node peering over a private interconnect — a deployment shape `DefaultSSRFDialPolicy`'s own comment says is real — stops syncing after upgrade with no log above debug level pointing at the setting that fixes it.
**Fix:** Make the policy rejection say so, in the one place that knows the setting exists:
```go
func (ps *PeerSelector) dataHubDialPolicy(ip net.IP) string {
	if ip.IsPrivate() {
		if ps.settings != nil && ps.settings.P2P.AllowPrivateIPs {
			return ""
		}

		ps.logger.Warnf("[PeerSelector] refusing peer probe to private address %s; set p2p_allow_private_ips=true if peers are reachable only over a private network", ip)

		return "private address"
	}

	return util.DefaultSSRFDialPolicy(ip)
}
```
If the per-resolved-IP log rate is a worry, hoist it into `isEligible` instead and key it off the peer ID so it fires once per peer per round.

### ChiR9 - `AllowPrivateIPs` does not "match its effect on validateDataHubURL"
**Problem:** The comment at `services/p2p/peer_selector.go:410-411` says "P2P.AllowPrivateIPs relaxes the private ranges, matching its effect on validateDataHubURL", and the new test asserts `require.Equal(t, isUnsafeIP(ip), strict.dataHubDialPolicy(ip), "the dial policy must agree with validateDataHubURL for %s", ipStr)`. Neither is quite right. `validateDataHubURL` doesn't relax private ranges under the flag — it returns `nil` before reaching any IP check at all (`services/p2p/server_helpers.go:409-412`), so with the flag on it also accepts loopback and link-local *literals* and the `localhost` hostname. The test only ever compares against `isUnsafeIP`, i.e. the flag-off branch, so it can't catch the difference. The behavioural consequence is small but real: on a node with `p2p_allow_private_ips=true`, a peer announcing `http://169.254.169.254/api/v1` passes validation, is registered, and incurs no protocol violation (`checkDataHubURL`, `server_helpers.go:431-444`) — it only dies later at dial time. The dial policy being *stricter* than the static check is the safe direction, so this is an accuracy problem, not a hole.
**Why it matters:** The comment is the thing the next person reads before touching this policy, and it currently claims a symmetry that would be wrong to preserve.
**Fix:** Say what actually happens, and keep the test message honest:
```go
// P2P.AllowPrivateIPs relaxes the private ranges here. Note it is deliberately
// narrower than the setting's effect on validateDataHubURL, which skips the static
// check entirely when the flag is set - loopback and link-local literals are refused
// at dial time in every configuration.
```
and change the assertion message to reference `isUnsafeIP` (the flag-off branch of `validateDataHubURL`) rather than the function as a whole.

### ChiR10 - The new clients drop the userinfo check on both the initial request and redirects
**Problem:** `ValidateURL` rejects credentials embedded in a peer-supplied URL (`util/http.go:455-459`, "Userinfo has no legitimate use here and can be used to bypass auth or confuse logging"), and the shared `httpClient` enforces that on *both* hops: on the initial request via `executeHTTPRequest` (`util/http.go:504`, and `util/http.go:680` for the streaming variant) and on redirects via `CheckRedirect` (`util/http.go:256`). The two clients this PR adds enforce it on neither. `NewSSRFSafeHTTPClient`'s `CheckRedirect` (`util/http.go:190-213`) checks only hop count, scheme and IP literals, and both consumers build the request and call `Do` directly with no `ValidateURL` (`model/subtree_meta_regenerator.go:111-116`, `util/health/http_health.go:47-53`). The probe is where that is reachable end to end today: `validateDataHubURL` has no userinfo check either (`services/p2p/server_helpers.go:387-423`), so a peer can announce `http://user:pass@peer.example/api/v1`, and `http.Client` turns the userinfo into an `Authorization: Basic` header on the probe request. The regenerator happens to be covered upstream, since `baseURL` is `ValidateURL`'d at `services/blockvalidation/Server.go:1332` before it reaches `createMetaRegenerator` — but that is the caller's doing, not the client's, so the next consumer of `NewSSRFSafeHTTPClient` inherits the gap. To be clear about the ceiling here: both the host and the credentials are peer-chosen, so no secret of ours leaks and no internal auth is bypassed (the dial policy still refuses internal addresses). What it costs is the invariant, plus the deception value of `http://legit-peer.example@attacker.example/api/v1` in the operator-facing logs that print DataHubURL verbatim (`services/p2p/sync_coordinator.go:1048,1061`, `services/p2p/peer_selector.go:268`).
**Why it matters:** Three guarded clients over the same peer-controlled input with three different rulesets is how the next real gap gets in; patching only the redirect hop would leave the new clients still divergent from the one they were modelled on.
**Fix:** Close both hops. In `validateDataHubURL`, match what `ValidateURL` already does — and put it before the `AllowPrivateIPs` short-circuit so it holds in every configuration:
```go
if parsed.User != nil {
	return errors.NewInvalidArgumentError("DataHubURL must not contain userinfo (credentials)")
}
```
and in `NewSSRFSafeHTTPClient`'s `CheckRedirect`, alongside the scheme check:
```go
if req.URL.User != nil {
	return errors.NewInvalidArgumentError("SSRF redirect check: target must not contain userinfo")
}
```
Worth a test on each: a peer announcement carrying userinfo rejected by `validateDataHubURL`, and a redirect to a userinfo URL rejected by the client. If you'd rather the client carry the guarantee itself rather than relying on callers, a thin `RoundTripper` wrapper that runs `ValidateURL(req.URL.String())` before delegating covers the initial request too, and then the regenerator stops depending on `Server.go:1332` being there.

## Recap

| ChiR8 | Private-peer refusal undiagnosable | | 35% |
| ChiR9 | AllowPrivateIPs symmetry claim wrong | | 30% |
| ChiR10 | Userinfo unchecked in new clients | | 20% |

icellan (commented):
Re-reviewed after the merge from main. The mechanism is right and genuinely closes the hostname→internal gap: resolve once, reject if any resolved address is unsafe, dial the validated IP without re-resolving, and re-check each redirect hop and scheme. No nil-deref or race issues; `PeerSelector` is only constructed via `NewPeerSelector`, so the guarded client can't be bypassed, and the global `SetSSRFProtection` toggle is restored correctly in the new tests.

The remaining risk isn't the plumbing — it's the policy tightening on the peer probe, which can silently make peers unselectable in private-network deployments.

## Availability regression: private ranges on the probe path

`services/p2p/peer_selector.go:652` — `dataHubDialPolicy` rejects RFC1918/ULA whenever `p2p_allow_private_ips` is false, which is the default (set true only in `compose/settings_test.conf`). Before this PR the probe used a plain client and connected fine.

Scenario: a k8s/internal deployment where peers advertise a cluster-internal DataHub *hostname*. It passes `validateDataHubURL` (no DNS there, by design), resolves to 10.x, every probe now fails, and `isEligible` (`peer_selector.go:405-419`) drops the peer from both the full-node and pruned candidate lists. The node ends up able to select no sync peer at all — while the actual block/subtree fetch path (`util`'s shared client with `DefaultSSRFDialPolicy`) still permits private and would have worked. The probe refuses a target the node would happily fetch from.

Either match the fetch path on private ranges, or make this a distinct opt-out so a private-network operator isn't silently wedged.

## Claimed symmetry with `validateDataHubURL` doesn't hold

`services/p2p/peer_selector.go:640-651`, `settings/p2p_settings.go:87`, `docs/references/settings/services/p2p_settings.md:171` — the comment and both docs say `AllowPrivateIPs` "relaxes the private ranges, matching its effect on `validateDataHubURL`". It doesn't: `validateDataHubURL` short-circuits with `return nil` when the flag is set (`services/p2p/server_helpers.go:410`), skipping the loopback/link-local/unspecified checks too.

Scenario: single-host dev/CI stack with `p2p_allow_private_ips=true` and a DataHubURL of the form the compose `docker.host` context uses (`http://localhost:PORT/api/v1`, cf. `compose/settings_test.conf:45`). The announcement is accepted, but the probe now permanently fails with "loopback address" and the peer is never selected. `TestDataHubDialPolicy` only asserts class-by-class agreement for the `AllowPrivateIPs=false` path, so the claimed symmetry is untested exactly where it breaks.

## Lower priority

- `util/http.go:122-134` — `dialAttemptContext` slices the *whole-request* deadline by the number of remaining candidates. For a peer hostname with several A records under the 2s probe timeout, the address that actually works gets 2s/N (500ms with 4 records); previously each sequential attempt could use the full remaining deadline. A legitimate but slightly distant multi-A-record peer now reads unhealthy where a single-address peer at the same RTT passes. A floor would help: `min(remaining, max(500ms, remaining/N))`.
- `model/subtree_meta_regenerator.go:22-30` — the comment justifies the package-level client with "a client each would give each block its own connection pool". Backwards: the old `&http.Client{Timeout: 30s}` had a nil Transport and used the shared `http.DefaultTransport` pool. The *new* client is the one introducing a separate pool (`Clone()` of DefaultTransport → `MaxIdleConnsPerHost=2`, vs 100 on `util`'s shared client). Behaviour is fine either way, but the stated reason will mislead the next reader.
- `util/health/http_health.go:35-37` — `CheckHTTPServerWithClient(nil, …)` silently falls back to an unguarded client. The doc says peer-URL callers "must" pass an SSRF-safe client, but nothing enforces it, so a future peer-address probe added via `CheckHTTPServer` loses the guard with no signal. A `CheckPeerHTTPServer(client *http.Client, …)` that rejects nil would make the misuse impossible.
- `util/http.go:252-260` — the shared `httpClient`'s `CheckRedirect` still uses `ValidateURL`, which allows loopback and private literals, while `NewSSRFSafeHTTPClient` re-checks redirect targets against the actual dial policy. Net effect is the same (the dialer stops loopback), but the file now carries two divergent redirect checks for the same threat. Folding the shared client onto `DefaultSSRFDialPolicy` would make the redirect layer real defense-in-depth rather than decorative.

icellan (commented):
Re-reviewed at 49ef41e (13 files, +733/−75 from merge-base 8bba344). `go vet ./util/... ./services/p2p ./model` clean; the SSRF/health/regenerator/peer-selector tests in `./util`, `./util/health`, `./model` and `./services/p2p` pass locally.

`a3ee8a9` resolves both blocking findings, and resolves them the right way rather than papering over them:

- The probe now shares `util.DefaultSSRFDialPolicy` with the fetch path (`peer_selector.go:97`), `dataHubDialPolicy` is gone, and the comment states the actual reason — a stricter probe policy drops peers the fetch path would talk to. The private-network wedge is closed.
- The `AllowPrivateIPs` docs no longer claim symmetry with `validateDataHubURL`. Both the settings tag and `p2p_settings.md` now split it correctly: the flag governs the static announcement check (which short-circuits entirely) and does *not* affect the connection-time guard. `TestPeerProbeIgnoresAllowPrivateIPs` pins that asymmetry, including the `localhost` case that previously only existed as an untested claim. Accepting an announcement and being reachable are now clearly separate questions, which is the distinction the earlier version blurred.

No new findings. The four lower-priority items from the last pass are unchanged and still stand:

- `util/http.go:122-134` — `dialAttemptContext` still slices the *whole-request* deadline by remaining candidate count. Under the 2s probe timeout a peer hostname with 4 A records gives the working address 500ms; previously each sequential attempt could use the full remaining deadline. A floor (`min(remaining, max(500ms, remaining/N))`) keeps the failover intent without penalising multi-A-record peers at otherwise-fine RTT.
- `model/subtree_meta_regenerator.go:22-30` — the rationale is still backwards. The old `&http.Client{Timeout: 30s}` had a nil Transport and shared `http.DefaultTransport`'s pool; the *new* client is the one that introduces a separate pool (`Clone()` → `MaxIdleConnsPerHost=2` vs 100 on `util`'s shared client). Sharing one client is right, the stated reason isn't.
- `util/health/http_health.go:35-37` — `CheckHTTPServerWithClient(nil, …)` still falls back to an unguarded client. The doc says peer-URL callers must pass an SSRF-safe client but nothing enforces it, so a future peer-address probe wired through `CheckHTTPServer` loses the guard silently. A `CheckPeerHTTPServer` that rejects nil makes the misuse unrepresentable.
- `util/http.go:252-260` — the shared `httpClient`'s `CheckRedirect` still uses `ValidateURL` (allows loopback and private literals) while `NewSSRFSafeHTTPClient` re-checks redirect targets against the dial policy. Same net effect, since the dialer stops loopback, but the file carries two divergent redirect checks for one threat. Folding the shared client onto `DefaultSSRFDialPolicy` makes that layer real defense-in-depth.

Nit: `isBlockedDialIP` (`util/http.go:221`) is now referenced only from `util/http_test.go` — the tests assert a wrapper nothing ships. Point them at `DefaultSSRFDialPolicy` and drop it.

None of these block merge.

icellan (commented):
Re-reviewed at 4f9326a (13 files, +882/−89 from merge-base fa90b23). `go vet` clean; `go test -race ./util ./util/health` and full `./services/p2p ./model` green locally.

`4f9326a` closes every outstanding item from the last pass, and the fixes are better than what I asked for:

- `dialAttemptContext` now floors the per-address share at `minDialAttemptBudget` (500ms) and returns the parent context unchanged when the floor covers what's left, so the sub-deadline can never exceed the parent budget. The multi-A-record starvation case is gone without weakening the failover intent.
- The two divergent redirect checks are now one `ssrfCheckRedirect(policy)`, used by both `NewSSRFSafeHTTPClient` and the shared `httpClient`. It's strictly stricter than the old `ValidateURL`-based version — adds scheme validation and loopback-literal rejection, keeps the hop cap, and rejects userinfo. That layer is real defense-in-depth now rather than decorative.
- `CheckPeerHTTPServer` replaces the nil-tolerant path and fails closed on a nil client instead of silently dialling unguarded. Exactly the right posture for a peer-supplied address.
- `isBlockedDialIP` is fully removed, no dead code left behind.

Two low-severity items left, neither blocking:

- `util/http.go:57` — a nil `SSRFDialPolicy` handed to `NewSSRFSafeDialContext` (or `NewSSRFSafeHTTPClient`, line 241) panics on the nil func call inside the transport's dial goroutine, i.e. at the first peer fetch, rather than failing closed. That's the opposite posture to the `CheckPeerHTTPServer` you just added, which deliberately rejects nil rather than defaulting. A future caller mirroring the old nil-tolerant `CheckHTTPServer` habit and writing `NewSSRFSafeHTTPClient(30*time.Second, nil)` compiles fine and panics the process on the first peer-supplied URL. Default to `DefaultSSRFDialPolicy` on nil, or assert it at construction.
- `services/p2p/peer_selector_ssrf_test.go:33` and `model/subtree_meta_regenerator_test.go:113` — both helpers restore the process-global SSRF flag to a hardcoded `true` in cleanup rather than the prior value, and the same helper is duplicated in two packages. Latent today, since neither package can host a `daemon.NewTestDaemon` (which sets it false at `daemon/test_daemon.go:160`) without an import cycle. But the first test binary that legitimately runs with protection disabled process-wide has these helpers silently re-enabling it for everything after. One `util` test helper that saves and restores the prior value (needs a getter) fixes both.

Otherwise the design holds up: validate-every-resolved-address then dial the validated IP still closes the rebinding window, and `checkPeerAvailability`'s new error path only changes log text — the cache decision still keys off `probeCtx.Err()`.

icellan (approved):

Issue #1368, opened 2026-07-28
[P1][v0.16] catchup stalls on cached empty subtree_data; failures misattributed to the healthy primary peer


oskarszoon:
**Priority: P1 · Target: v0.16**

**Component:** `services/blockvalidation` — `get_blocks.go` (`fetchAndStoreSubtreeAndSubtreeData`), `catchup.go` (`releaseCatchupLock`), `catchup_performance.go` (`GetPeersAtMaxHeight`)

## Summary

A single poisoned `subtree_data` response (**HTTP 200 with an empty body, served from cache**) permanently stalls post-checkpoint catchup, and teranode's error handling makes it nearly impossible to diagnose: the whole 8795-block batch is aborted, the failure is attributed to the **catchup primary peer** instead of the peer that actually failed, and the error text surfaced belongs to whichever *alternative* peer was tried last.

Net effect on a real teratestnet node: **every catchup against every peer fails (0% success across all peers), the node cannot sync past the checkpoint, and the healthiest peers accumulate 100% failure records for another peer's HTTP 404s.**

Observed on `bsva-ovh-teranode-ttn-eu-3` (teranode-quickstart, p2p-only, latest beta) syncing teratestnet from genesis. Stuck at height 18635 against a tip of 27431.

### 1. All peers show 0 successes; the broken peer is penalized least

`GET /api/v1/peers` on the stalled node:

| peer | height | attempts | successes | failures | reputation |
|---|---|---|---|---|---|
| `teranode-ttn-eu-1` | 27431 | 20 | **0** | 20 | 89.6 |
| `teranode-ttn-eu-2` | 27431 | 18 | **0** | 18 | 82.4 |
| `Devin-TTN-XServer` (162.43.7.61) | 27431 | 4 | **0** | 4 | 50.9 |
| `teranode` | 4411 | 0 | 0 | 0 | — (no DataHubURL) |
| `teranode-02` | 9325 | 0 | 0 | 0 | — (below max height) |

All three tip peers are at max height, connected, unbanned — i.e. all eligible. `Devin-TTN-XServer` is the peer that actually cannot serve the data, yet it has the *fewest* attempts and therefore absorbs the least penalty, while the two peers that CAN serve the data carry 38 failures between them.

### 2. The peers being blamed actually have the data

For the exact subtree named in the failure (`513e6b6c425547b8ec42d20d22fd82e379e27eb5d1c7405ceee21a0980f92798`):

[...]

PR #1371, opened 2026-07-28
fix(blockvalidation): survive a peer serving cached empty subtree_data, attribute catchup failures correctly

oskarszoon:
Part of #1368 — the client-side half. Asset/nginx serving fixes and the log-storm fix follow in separate PRs.

## Why

A single peer serving `subtree_data` as `HTTP 200` with an **empty body**, replayed from its nginx `proxy_cache` (`x-cache-status: HIT`, `content-length: 0`), permanently stalled post-checkpoint catchup on a teratestnet node — height pinned at 18635 against a tip of 27431. Every catchup against every peer failed, and the diagnostics pointed at the wrong peer: two healthy peers accumulated 20/20 and 18/18 failure records carrying a *third* peer's 404.

The serving fleet can't be updated on our schedule, so the client has to make progress against peers that will keep replaying a poisoned cache entry indefinitely. That constraint shaped this PR: it's the deliverable, not the serving-side fix.

## What was actually broken

**An empty body parsed clean.** go-subtree's `serializeFromReader` hits `io.EOF` on its first read and returns `nil` with every tx nil, so the failure only surfaced later inside `Serialize()` as a generic `ErrSubtreeLengthMismatch` naming no peer, no URL, no byte count.

**The surfaced error belonged to the wrong peer.** `fetchAndStoreSubtreeAndSubtreeData` kept one `lastErr`, overwritten by each alternative, so the persisted and displayed cause was whichever alternative failed *last* — unrelated to the primary. The primary's real reason was `Warnf`'d and each alternative's was `Debugf`'d, i.e. invisible at INFO.

**Failures were charged to the catchup primary.** Two sites: `releaseCatchupLock` had no `ErrExternal` case, so all-peers-failed fell through to `unknown_error` with `isPeerError = true`; and `processCatchupChItem`'s `ErrExternal` branch charged `c.peerID` unconditionally. Reputation-based selection was being steered *away* from the peers that worked.

## What this changes

- **Detect the poison explicitly** (`get_blocks.go`). An empty or short `subtree_data` body becomes a distinct `ErrExternal` naming the peer, the URL, the bytes read and how many txs are missing. The predicate mirrors what go-subtree's `Serialize` safely tolerates, so nothing that succeeds today starts failing.
- **Retry that peer once with `?cachebust=<n>`.** A peer's cache key includes `$request_uri` while nginx location matching ignores the query string, so a busted URL reaches the same handler and misses the cache, forcing regeneration — no peer-side change. `proxy_cache_bypass` is pinned to `0` in the asset cache conf, so a request header can't do this; the query parameter is the only lever available against an unpatched peer. Fires only after a detected poisoning.
- **Keep every peer's error.** The all-peers-failed error carries `primary <id> (<url>)=…; alternative <id> (<url>)=…` for every attempt and wraps the *primary's* error as the cause. Alternative failures moved `Debugf` → `Warnf`.
- **Charge each peer for its own failure, once per cycle.** A deduplicated `failedPeers` set on `CatchupContext`, drained in `releaseCatchupLock`, plus an `ErrExternal` case in that switch. `recordCatchupPeerFailure` skips local errors, so our own cancellations and storage failures can't land on a peer.

## Behaviour changes worth a reviewer's attention

1. The two "Local error fetching subtree" / "…subtreeData" messages collapse into one — the wrapped cause still identifies the stage.
2. Alternative-peer failures are now `Warnf`, and the pre-attempt `Debugf` is gone. Every alternative attempt still terminates in either a `Warnf` or the success `Infof`.
3. `ReportPeerFailure` on the all-peers-failed path charges the sync peer one extra interaction failure. That's deliberate: the call is the peer-*rotation* signal (`blockchain` → `NotificationType_PeerFailure` → `HandleCatchupFailure` → clears `currentSyncPeer`), not just blame. Gating it to avoid the charge silently degraded rotation to the 5-minute `SyncPeerNoProgressTimeout` — in this issue's own failure mode. For a peer with history the charge moves reputation ~0.6 points, nowhere near the `< 20` rotation gate. Do not re-add the gate; there's a comment saying so.
4. Subtree-fallback peers now receive a `CatchupFailure` with no paired `CatchupAttempt`, so the displayed attempts/failures pairing for a fallback-only peer isn't a rate. Reputation is computed from the separate `Interaction*` counters, which `RecordCatchupFailure` keeps consistent, so selection math is unaffected.

## Ordering / load

The bypass ceiling is `FetchNumWorkers` × `SubtreeFetchConcurrency` = up to 128 concurrent forced regenerations against a fleet-wide poisoned cache — exactly one per subtree per peer, no retry loop. The backstop is the serving peer's own non-blocking admission control returning 503, which the client honours with `Retry-After` backoff.

## Also fixed here

A remotely-triggerable panic, found by review and verified by execution: a 1-node subtree whose `Nodes[0]` is not the coinbase placeholder, plus an empty `subtree_data` body, made go-subtree's `Serialize` dereference a nil `*bt.Tx` inside a `g.Go` goroutine with no `recover()` on that stack. Reachable for any block whose tx count is ≡ 1 (mod subtree size). Pre-existing, but on the exact path this PR hardens.

## Testing

Full `services/blockvalidation` package green; regression tests for the empty body, the short body, the coinbase-only subtree that must *not* be flagged, the cache-bypass retry (asserting the busted URL was requested exactly once), the one-node non-coinbase panic shape, and `releaseCatchupLock` cases covering `ErrExternal` classification, the local-error skip, the mixed-cycle charge, and the incomplete-block double-charge guard. `go vet` and `make lint` clean.

## Not in this PR

- Serving side: asset must not emit a cacheable empty 200, and `asset-cache-nginx.conf` needs `proxy_http_version 1.1` — nginx defaults to HTTP/1.0 upstream, where a body ends at connection close, which makes `streamOrAbort`'s hijack-without-chunked-terminator defense inert in production. Separate PR.
- The `context canceled` log storm (6k–47k lines per cycle). Separate PR.
- Deferred to follow-up issues: re-queuing the affected block instead of cancelling the shared errgroup; per-resource-class peer deprioritization on 404; root-hash verification in `fetchAndStoreSubtree`; the pre-existing cross-site charge duplication across `releaseCatchupLock` / `processCatchupChItem` / `peer_selection.go`.

### DIFF
--- a/services/blockvalidation/Server.go
+++ b/services/blockvalidation/Server.go
@@ -1811,10 +1811,17 @@ func (u *Server) processCatchupChItem(ctx context.Context, c processBlockCatchup
 			u.logger.Warnf("[catchup] All peers failed for block %s (attempt %d/%d), clearing markers and reporting peer failure to allow retry from a different peer: %v", c.block.Hash().String(), attempts, u.settings.BlockValidation.CatchupMaxAttemptsPerBlock, err)
 			u.processBlockNotify.Delete(*c.block.Hash())
 			u.catchupAlternatives.Delete(*c.block.Hash())
-			u.reportCatchupFailure(ctx, c.peerID)
 
-			if reportErr := u.blockchainClient.ReportPeerFailure(ctx, c.block.Hash(), c.peerID, "catchup", err.Error()); reportErr != nil {
-				u.logger.Errorf("[catchup] failed to report peer failure for block %s peer %s: %v", c.block.Hash().String(), c.peerID, reportErr)
+			// Peers that actually failed were charged individually at the point of
+			// failure (recordCatchupPeerFailure / markCatchupFailureReported). Charging
+			// c.peerID here would blame the catchup primary for another peer's failure,
+			// which is what drove healthy peers to 0% success in issue 1368.
+			u.reportCatchupFailureForError(ctx, c.peerID, err)
+
+			if !catchupFailureAlreadyReported(err) {
+				if reportErr := u.blockchainClient.ReportPeerFailure(ctx, c.block.Hash(), c.peerID, "catchup", err.Error()); reportErr != nil {
+					u.logger.Errorf("[catchup] failed to report peer failure for block %s peer %s: %v", c.block.Hash().String(), c.peerID, reportErr)
+				}
 			}
 
 			return
--- a/services/blockvalidation/Server.go
+++ b/services/blockvalidation/Server.go
@@ -1815,13 +1815,40 @@ func (u *Server) processCatchupChItem(ctx context.Context, c processBlockCatchup
 			// Peers that actually failed were charged individually at the point of
 			// failure (recordCatchupPeerFailure / markCatchupFailureReported). Charging
 			// c.peerID here would blame the catchup primary for another peer's failure,
-			// which is what drove healthy peers to 0% success in issue 1368.
+			// which is what drove healthy peers to 0% success in issue 1368. The
+			// already-reported marker check inside reportCatchupFailureForError
+			// suppresses exactly that duplicate *reputation* charge.
 			u.reportCatchupFailureForError(ctx, c.peerID, err)
 
-			if !catchupFailureAlreadyReported(err) {
-				if reportErr := u.blockchainClient.ReportPeerFailure(ctx, c.block.Hash(), c.peerID, "catchup", err.Error()); reportErr != nil {
-					u.logger.Errorf("[catchup] failed to report peer failure for block %s peer %s: %v", c.block.Hash().String(), c.peerID, reportErr)
-				}
+			// ReportPeerFailure is deliberately NOT gated on catchupFailureAlreadyReported.
+			// It is not a reputation call — it is the sync-peer ROTATION signal, and it has
+			// to fire on every all-peers-failed cycle:
+			//
+			//   blockchain.Server.ReportPeerFailure broadcasts NotificationType_PeerFailure
+			//   -> p2p.Server routes failure_type "catchup" to
+			//      syncCoordinator.HandleCatchupFailure (and its subscription listener
+			//      deliberately bypasses the "skip notifications while syncing" filter for
+			//      this type — "needed to switch peers on catchup failure")
+			//   -> SyncCoordinator clears currentSyncPeer and calls triggerSyncLocked, i.e.
+			//      immediate re-selection of a different sync peer.
+			//
+			// A marker gate here is dead code in production: fetchAndStoreSubtreeAndSubtreeData
+			// applies markCatchupFailureReported unconditionally to every all-peers-failed
+			// error and is the only producer of this ErrExternal in the package, so the gate
+			// is always false and rotation never happens. A stuck node would then depend on
+			// the sync coordinator's 30s periodicEvaluation, which only rotates on reputation
+			// below 20 or the 5-minute SyncPeerNoProgressTimeout — a peer with history (say
+			// 100 successes and 1 failure) trips neither, so recovery costs the full five
+			// minutes per stuck cycle.
+			//
+			// The price is one extra interaction-failure charge against the primary. That is
+			// accepted, and consistent with the round-2 adjudication documented in
+			// catchup.go's failedPeers drain: an over-charge is directionally accurate while
+			// an under-charge hides a real failure, and the duplicate increment is a
+			// telemetry-precision cost rather than a reputation-math break. Losing peer
+			// rotation to save one charge is the worse trade — do not re-add the gate.
+			if reportErr := u.blockchainClient.ReportPeerFailure(ctx, c.block.Hash(), c.peerID, "catchup", err.Error()); reportErr != nil {
+				u.logger.Errorf("[catchup] failed to report peer failure for block %s peer %s: %v", c.block.Hash().String(), c.peerID, reportErr)
 			}
 
 			return
--- a/services/blockvalidation/Server.go
+++ b/services/blockvalidation/Server.go
@@ -251,6 +251,10 @@ type Server struct {
 	blocksFetched   atomic.Int64
 	blocksValidated atomic.Int64
 
+	// cacheBustCounter produces the unique token appended to a peer request URL when
+	// a previous response from that peer looked poisoned. See peer_cache_bypass.go.
+	cacheBustCounter atomic.Uint64
+
 	// previousCatchupAttempt stores details about the last failed catchup attempt.
 	// This is used to display in the dashboard why we switched from one peer to another.
 	// Protected by activeCatchupCtxMu for thread-safe access.
--- a/services/blockvalidation/block_processing_retry_test.go
+++ b/services/blockvalidation/block_processing_retry_test.go
@@ -351,6 +351,37 @@ func TestProcessCatchupChItem(t *testing.T) {
 		mockBC.AssertCalled(t, "ReportPeerFailure", mock.Anything, mock.Anything, mock.Anything, "catchup", mock.Anything)
 	})
 
+	// The two cases above build ErrExternal chains WITHOUT the already-reported
+	// marker. Production never emits that shape: fetchAndStoreSubtreeAndSubtreeData
+	// applies markCatchupFailureReported to every all-peers-failed error, and it is
+	// the only producer of this ErrExternal in the package. This case pins the real
+	// shape — ReportPeerFailure is the sync-peer ROTATION signal (blockchain
+	// PeerFailure notification -> p2p HandleCatchupFailure -> currentSyncPeer cleared
+	// + triggerSyncLocked), so it must still fire even though the marker suppresses
+	// the duplicate reputation charge in reportCatchupFailureForError. Gating it on
+	// the marker silently disabled rotation and left recovery to the 5-minute
+	// SyncPeerNoProgressTimeout.
+	t.Run("marked external error (production shape) still reports peer failure for rotation", func(t *testing.T) {
+		// Exactly what get_blocks.go returns, then the wraps it picks up on the way up.
+		marked := markCatchupFailureReported(errors.NewExternalError("all 3 peer attempts failed to fetch subtree aa"))
+		svcWrap := errors.NewServiceError("failed to fetch subtree data for block bb", marked)
+		chain := errors.NewProcessingError("worker failed for block bb", svcWrap)
+
+		require.True(t, catchupFailureAlreadyReported(chain), "precondition: production marks every all-peers-failed error")
+		require.True(t, errors.Is(chain, errors.ErrExternal))
+
+		u, _ := newServer(3, chain)
+		b := testBlock()
+		u.processBlockNotify.Set(*b.Hash(), true, ttlcache.DefaultTTL)
+
+		u.processCatchupChItem(ctx, item(b))
+
+		require.Nil(t, u.processBlockNotify.Get(*b.Hash()))
+
+		mockBC := u.blockchainClient.(*blockchain.Mock)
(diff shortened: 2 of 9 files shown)


ctnguyen (approved):
Review at commit [58daaf7](https://github.com/bsv-blockchain/teranode/commit/58daaf7f8a618b6fc253c3bce00bdef55c5e2799)

This is a well-executed client-side fix for a nasty field bug, and the diagnosis behind it clearly did the hard work. The poisoned-body predicate is genuinely correct against the go-subtree v1.4.2 pinned in `go.mod:17`: `serializeFromReader` sizes `Txs` to `subtree.Length()` (`subtree_data.go:286`) and skips index 0 only when `Nodes[0]` is the coinbase placeholder (`subtree_data.go:281-282`), which is exactly the one case the predicate exempts — so with `missing == 0` neither `Serialize()` branch can reach the nil deref at `subtree_data.go:103` that its own `i != 0` guard (`subtree_data.go:94`) lets through. Consolidating the primary/alternative fetch into `tryPeerForSubtree` removes a chunk of duplicated logic while adding behaviour, and the tests are unusually good for a change like this: the coinbase-only-must-not-flag case and the one-node non-coinbase panic shape are the two I'd have asked for, and the cache-bypass test really does pin that the retry re-requests only `subtree_data`. Nothing here blocks merge; the items below are attribution and diagnostics polish on top of a sound fix.

## Non-blocking issues

### ChiR1 — The new `ErrExternal` case also captures subtreevalidation's peer failures, dropping their stored error text

**Problem:** `catchup.go:397` classifies *any* `ErrExternal` anywhere in the terminal error chain as `peer_data_unavailable` with `isPeerError = false`. `services/subtreevalidation` also produces `ErrExternal` on peer-fetch paths — `SubtreeValidation.go:299` (`getMissingTransactionsBatch`, "failed to do http request" against `<baseURL>/subtree/<hash>/txs`) and `SubtreeValidation.go:1021` (`getSubtreeTxHashes`, "failed to do http request on host %s"). Both are reachable during catchup: `BlockValidation.go:2444` → `CheckBlockSubtrees` → `check_block_subtrees.go:681` `validateMissingSubtreesWithOrderedRetry` → `ValidateSubtreeInternal`, returned through `errors.WrapGRPC`/`UnwrapGRPC`, which rebuilds one native link per detail and preserves each link's code (`errors/errors.go:978-1019`), so `errors.Is(*err, ErrExternal)` matches by the time the error reaches `releaseCatchupLock` via `validateBlocksOnChannel` (`catchup.go:1344`). These are not all-peers-failed errors, and the failing peer is never in `failedPeers` — only `tryPeerForSubtree` writes to it.

To be precise about the blast radius, because it is narrower than it first looks: `isPeerError` gates exactly one call, `u.reportCatchupError(rpcCtx, peerID, errorMsg)` at `catchup.go:543-545`, which stores the peer's `LastCatchupError` (`services/blockchain/peer_registry.go:1305`). It never charged a failure counter. The counter for this class still fires — `catchupFunc` is `u.catchup` (`Server.go:397`), so `processCatchupChItem` at `Server.go:1771` sees the same error object, a subtreevalidation-origin `ErrExternal` carries no `markCatchupFailureReported` marker, and `reportCatchupFailureForError` at `Server.go:1821` therefore falls through to `reportCatchupFailure(ctx, c.peerID)`. So the two real deltas versus pre-PR are: the peer's stored catchup error text is no longer updated, and `errorType` is mislabelled `peer_data_unavailable`.

**Why it matters:** diagnosability is this PR's headline goal — issue 1368 took manual endpoint probing to diagnose precisely because the stored per-peer error was misleading. For subtreevalidation-origin peer HTTP failures that field now goes stale rather than wrong, and the dashboard reports `peer_data_unavailable` for a failure that has nothing to do with subtree-data availability. Failure counting and reputation math are unchanged and sync is unaffected, which is why this is not a merge blocker.

**Fix:** mark the one error this classification is for, at its single production site, and key the case on that marker rather than on the bare code:

```go
// peer_cache_bypass.go — alongside the existing cache-bypass marker
const allPeersFailedKey = "all_peers_failed_subtree"

// markAllPeersFailed / isAllPeersFailed mirror markCacheBypassRetryable /
// isCacheBypassRetryable exactly: SetData on the native link, then a
// WrappedErr() walk bounded at depth 32.

// get_blocks.go:791
return "", markAllPeersFailed(markCatchupFailureReported(errors.NewExternalError(...)))

// catchup.go:397
case errors.Is(*err, errors.ErrExternal) && isAllPeersFailed(*err):
```

Gating on the existing `catchupFailureAlreadyReported(*err)` instead is a one-line alternative that works today, but it couples this classification to a marker also applied to header-fetch failures at `catchup_get_block_headers.go:271`, `:300` and `:306` — those are not `ErrExternal` today, so nothing breaks, but the coupling is invisible and easy to break later. Note also that the pre-existing dispatch at `Server.go:1806` (`if errors.Is(err, errors.ErrExternal)`) has the same over-breadth: a subtreevalidation-origin `ErrExternal` takes its "All peers failed for block" branch and log line. That is not a regression from this PR, but if you add the marker it is worth using at both sites so the two dispatches cannot drift.

### ChiR2 — The local-error abort in the alternatives loop throws away the attempts summary

**Problem:** `get_blocks.go:760-762` appends the attempt and then returns `errors.NewServiceError(..., altErr)` without `formatSubtreeFetchAttempts(attempts)`. Everything collected so far, including the primary's error, is discarded on that path.

**Why it matters:** it's a hole in the fix for Defect A. If the local failure lands while trying alternative 2, the operator sees only "local error fetching subtree X" plus a storage or context cause, with no record that the primary and alternative 1 failed first — the same blind spot the PR set out to close, on a rarer branch.

**Fix:** include the accumulated summary in the abort message, exactly as the all-peers-failed path does.

```go
if errors.IsLocalError(altErr) {
    return "", errors.NewServiceError("[catchup:fetchAndStoreSubtreeAndSubtreeData] Local error fetching subtree %s (aborting peer retry) [%s]",
        subtreeHash.String(), formatSubtreeFetchAttempts(attempts), altErr)
}
```

### ChiR3 — `recordCatchupPeerFailure` can attribute a non-catchup fetch to the active catchup cycle

**Problem:** `catchup.go:584-586` resolves the target through the global `u.activeCatchupCtx` rather than a context threaded from the caller. `fetchSubtreeDataForBlock` has a second, non-catchup caller: `Server.go:1247` (`RevalidateBlock`). If an operator triggers `RevalidateBlock` while a catchup is in flight, its per-subtree fetch failures land in that catchup's `failedPeers` and are charged by the drain when the cycle ends in error.

**Why it matters:** small blast radius, since `RevalidateBlock` is operator-triggered, but it is the same class of mis-attribution the PR is fixing. The doc comment at `catchup.go:571-573` ("with no active catchup context … the call is a no-op") only holds when no catchup is running, so the reasoning as written does not cover this case.

**Fix:** pass the `*CatchupContext` (nil-able) down through `fetchSubtreeDataForBlock` → `fetchAndStoreSubtreeAndSubtreeData` → `tryPeerForSubtree` so attribution is explicit, and drop the global read. If threading it is too invasive here, amend the comment to name the `RevalidateBlock` overlap so the next reader isn't misled.

### ChiR4 — The all-peers-failed message repeats the primary's error and is stored and broadcast unbounded

**Problem:** `get_blocks.go:791` embeds `formatSubtreeFetchAttempts(attempts)` in the message *and* wraps `primaryErr` as the cause. `(*errors.Error).Error()` appends the wrapped chain after the message (`errors/errors.go:106-108`), so the primary's full chain appears twice. That string becomes `previousCatchupAttempt.ErrorMessage`, the stored catchup error for every peer in the drain (`catchup.go:531`, `catchup.go:536`), and the `reason` field of the `PeerFailure` blockchain notification (`Server.go:1850`).

**Why it matters:** with three peers and the usual 4–5 link chains this is comfortably multi-KB, duplicated, in a value that is persisted per peer, rendered in the dashboard, and fanned out to every notification subscriber.

**Fix:** either drop the trailing `primaryErr` — the summary already names it, and the `ErrExternal` classification does not depend on the wrapped cause — or keep the wrap and render only a bounded slice of each attempt's text:

```go
parts = append(parts, fmt.Sprintf("%s %s (%s)=%s", attempt.role, attempt.peerID, attempt.baseURL,
    truncate(attempt.err.Error(), 256)))
```

## Recap

|---|---|---|---|
| ChiR1 | Over-broad ErrExternal classification | | 40% |
| ChiR2 | Local-error abort loses attempts summary | | 30% |
| ChiR3 | Global catchup ctx mis-attribution | | 25% |
| ChiR4 | Duplicated, unbounded error text | | 15% |

ordishs (approved):
Approving. I verified the load-bearing claims independently rather than taking the description on trust:

- **go-subtree panic**: confirmed in `go-subtree@v1.4.2/subtree_data.go` — with a non-coinbase `Nodes[0]`, `txStartIndex = 0` while the nil check is guarded by `i != 0`, so a nil `Txs[0]` passes validation and `SerializeBytes()` panics. The coinbase-at-zero-only exemption in `get_blocks.go` is exactly right, and the `SingleNonCoinbaseNodeWithEmptyBodyIsRejectedNotPanicking` test pins it.
- **nginx claims**: `asset-cache-nginx.conf` keys the cache on `$request_uri` (query string included) and pins `proxy_cache_bypass 0`, so the query-parameter bust is indeed the only client-side lever against an unpatched peer.
- **Dispatch ordering**: `ErrExternal` before `IsNetworkError` in `releaseCatchupLock` and before `ErrServiceError` in `processCatchupChItem` — both correct, and `TestMarkedExternalError_SurvivesProductionWrapChain` guards the chain-walk against foreign-wrapper regressions.
- **`IsLocalError` guard**: matches only context and storage errors, so 404s/empty bodies are recorded against peers while our own cancellations never are.
- **Tests**: ran every new/modified test in the PR — all pass. Coverage is exemplary; the recorder asserting *which* peer was charged is exactly the right test shape for this bug.

Comments (follow-up issues filed, none blocking):

1. **Mid-cycle peer failures are dropped when the cycle succeeds.** `failedPeers` is copied and drained only inside the `*err != nil` branch of `releaseCatchupLock`. If the primary serves a poisoned body but an alternative rescues the block, the cycle ends clean and the poisoned peer is never charged — it stays fully reputable, gets selected again next cycle, and every cycle pays the bypass-retry latency. If a clean cycle deliberately fires no reputation RPCs, the `failedPeers` field comment should say so; it currently states unconditionally that the drain "charges each of them once."

2. **`cacheBustCounter` restarts at zero across process restarts.** The first bypass after every restart is `?cachebust=1`, and the busted response is itself cacheable under the same `$request_uri` key. If a bypass regeneration also produced a poisoned 200, a post-restart bypass for the same subtree/peer replays that cached `cachebust=1` entry — the exact failure mode the mechanism exists to escape. Seeding the counter from startup time (or a per-bust nanosecond timestamp) makes the token unique across restarts at no cost.

3. **Cross-cycle attribution edge via `context.WithoutCancel`.** `fetchAndStoreSubtreeData` intentionally survives cancellation, but `recordCatchupPeerFailure` resolves `u.activeCatchupCtx` at call time, so a fetch goroutine outliving its cycle either no-ops or records into the *next* cycle's `failedPeers`. The blamed peer is still the peer that failed, so this is cycle-attribution noise only — worth a sentence in `recordCatchupPeerFailure`'s doc comment, not a code change.

4. **Comment style, no issue filed.** The invariant-bearing comments ("do not re-add the gate", the ordering notes, the under-charge rationale) are genuinely valuable and should stay. The provenance narration around them ("round-2 adjudication", "review round 2, change 1", "an earlier version tried…") talks to this PR's reviewers and will be meaningless in six months — worth stripping to the invariant plus the failure it prevents next time someone touches these files. Same for the `[catchup:fetchAndStoreSubtreeAndSubtreeData]` log tag inside `tryPeerForSubtree`.

icellan (approved):

PR #1375, opened 2026-07-29
fix(aerospike): guard the remaining batcher enqueue sites and re-panic unrelated panics

ordishs:
Follow-up to #1361. These changes were pushed to that branch after it had already been merged at an earlier commit, so they never reached `main` — reopening them here rather than leaving the gap.

## Why this is needed

#1361 landed the guard on two enqueue sites (get and outpoint). `Store.Close` closes **seven** batchers, so on current `main` a `Spend`, `Create` or `SetLocked` in flight when `Close` runs still panics `send on closed channel` and takes the process down — on the hottest write path. This was raised as a blocking item (ChiR2) in the review of #1363.

Verified against `main` as it stands:

| file | guarded | unguarded |
|---|---|---|
| spend.go | 0 | 4 |
| create.go | 0 | 1 |
| locked.go | 0 | 1 |
| get.go | 3 | 0 |

## What this changes

- Adds `safeBatcherPutBatchCtx` and routes the six remaining sites through the guards: `Spend`, `Create` and `SetLocked` (reachable from external callers during shutdown), plus `SetDAHForChildRecords`, `handleExtraRecords` and `IncrementSpentRecords` (internal today, guarded so a future external caller or a change to `Close`'s ordering cannot reintroduce the crash).
- **Completion accounting:** `PutBatch*` enqueues a whole group in one channel send, so a rejected send rejects every item. Each batch call site completes all of its items, letting the existing post-`Wait` result read surface the shutdown error. Without this the shared `group.Wait` parks for its full budget and then reports a misleading timeout — `TestSetLocked_ClosedBatcherDoesNotParkForFullTimeout` pins that, and deleting the completion loop makes it fail with `"set locked did not complete within 30s" does not contain "shutting down"` after a 30s park.
- **Narrows the recover** to the send-on-closed-channel panic and re-panics anything else. A blanket recover reported a nil batcher or any future go-batcher failure mode as an orderly shutdown *and* as a retryable error, on a path that only runs when something is already wrong. Matching is on the rendered text, not the concrete type, because the runtime panics with a `runtime.plainError`. Covered by `TestSafeBatcherPut_RepanicsUnrelatedPanic`.
- Corrects the test comment that claimed coverage of spend/locked before those paths were actually wired.

## On the error code

The guard deliberately keeps `ERR_SERVICE_UNAVAILABLE` rather than the `ERR_UNKNOWN` that `handleSpendPanic` (`spend.go:340`) chose for its own recovered panic. That looks inconsistent but is not. `ERR_SERVICE_UNAVAILABLE` places the error inside `IsTransientLocalError`, which `netsync/manager.go:1776` and `peer_server.go:1204` consume in lock-step; with `ERR_UNKNOWN` both invert, so netsync would `PushRejectMsg(..., wire.RejectInvalid, ...)` — telling a peer a valid block is invalid — and `shouldDisconnectOnBlockErr` would drop the delivering peer, both because *our* node was shutting down. `quick_validate.go:1385` would likewise route it to `hardFail` instead of the bounded retry list. `handleSpendPanic` is right to differ because its cause is unknown and possibly a code bug; this guard's cause is precisely identified. Narrowing the recover is what keeps the two conditions from overlapping.

## Verification

`go build ./...`, `go vet` clean; full package suites green on the merged result:

```
ok  	github.com/bsv-blockchain/teranode/stores/utxo/aerospike	254.401s
ok  	github.com/bsv-blockchain/teranode/stores/utxo/aerospike/pruner	39.367s
```

After this, no bare `s.*Batcher.Put*(` remains in non-test code.

### DIFF
--- a/stores/utxo/aerospike/aerospike.go
+++ b/stores/utxo/aerospike/aerospike.go
@@ -124,20 +124,40 @@ type batcherIfc[T any] interface {
 	Close()
 }
 
-// safeBatcherPutCtx enqueues item into b, converting the "send on closed channel"
-// panic — which go-batcher v2.0.6 raises when Put is called after Close — into a
-// returned error. Store.Close closes the batchers during graceful shutdown while
-// external callers (block validation, assembly, …) may still be enqueuing; that
-// race must abort the operation, not crash the process. who labels the call site.
+// shutdownPanicText is the runtime's message when a send targets a closed
+// channel. That is how go-batcher v2 surfaces Put-after-Close; see
+// batcher.Batcher.Close, which documents that Put must not be called after it.
+const shutdownPanicText = "send on closed channel"
+
+// recoverBatcherShutdown turns the send-on-closed-channel panic into a returned
+// error and re-panics everything else. Store.Close closes the batchers during
+// graceful shutdown while external callers (block validation, assembly, …) may
+// still be enqueuing; that race must abort the operation rather than crash the
+// process. Any other panic — a nil batcher, a future go-batcher failure mode — is
+// a genuine bug and must not be relabelled as an orderly shutdown.
 //
-// The recovered value is pre-formatted with fmt.Sprintf: the runtime's panic
-// value is a runtime.plainError, which implements error, and errors.New consumes
-// a trailing error argument as the wrapped error rather than formatting it —
-// orphaning the %v verb and mislabelling the result as wrapping an UNKNOWN (0).
+// Matching is on the rendered text, not the concrete type: the runtime panics
+// with a runtime.plainError, so a type switch would pass in tests that fake the
+// panic with a string and miss in production. Rendering also has to happen before
+// the value reaches errors.New, which consumes a trailing error argument as the
+// wrapped error instead of formatting it — orphaning the %v verb and mislabelling
+// the result as wrapping an UNKNOWN (0).
+func recoverBatcherShutdown(r any, who string, err *error) {
+	text := fmt.Sprint(r)
+
+	if !strings.Contains(text, shutdownPanicText) {
+		panic(r)
+	}
+
+	*err = errors.NewServiceUnavailableError("[%s] aerospike batcher unavailable (store shutting down): %v", who, text)
+}
+
+// safeBatcherPutCtx enqueues item into b, converting a Put-after-Close panic into
+// a returned error. who labels the call site.
 func safeBatcherPutCtx[T any](b batcherIfc[T], ctx context.Context, item *T, who string) (err error) {
 	defer func() {
 		if r := recover(); r != nil {
-			err = errors.NewServiceUnavailableError("[%s] aerospike batcher unavailable (store shutting down): %v", who, fmt.Sprintf("%v", r))
+			recoverBatcherShutdown(r, who, &err)
 		}
 	}()
 
@@ -150,11 +170,27 @@ func safeBatcherPutCtx[T any](b batcherIfc[T], ctx context.Context, item *T, who
 func safeBatcherPut[T any](b batcherIfc[T], item *T, who string) (err error) {
 	defer func() {
 		if r := recover(); r != nil {
-			err = errors.NewServiceUnavailableError("[%s] aerospike batcher unavailable (store shutting down): %v", who, fmt.Sprintf("%v", r))
+			recoverBatcherShutdown(r, who, &err)
 		}
 	}()
 
 	b.Put(item)
+
+	return nil
+}
+
+// safeBatcherPutBatchCtx is the PutBatchCtx counterpart. PutBatch* enqueues the
+// whole group in a single channel send, so a rejected send rejects every item:
+// callers must complete all of them, or the shared group.Wait parks for its full
+// timeout and then reports a misleading timeout instead of the shutdown.
+func safeBatcherPutBatchCtx[T any](b batcherIfc[T], ctx context.Context, items []*T, who string) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			recoverBatcherShutdown(r, who, &err)
+		}
+	}()
+
+	b.PutBatchCtx(ctx, items)
 
 	return nil
 }
--- a/stores/utxo/aerospike/closed_batcher_group_test.go
+++ b/stores/utxo/aerospike/closed_batcher_group_test.go
@@ -1,0 +1,140 @@
+package aerospike
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/bsv-blockchain/go-bt/v2"
+	"github.com/bsv-blockchain/go-bt/v2/chainhash"
+	"github.com/stretchr/testify/require"
+)
+
+// TestSetLocked_ClosedBatcherDoesNotParkForFullTimeout is the regression test for
+// the batch-enqueue accounting. PutBatchCtx sends the whole group in one channel
+// send, so a Put-after-Close rejects every item at once. If the guard returned
+// its error without completing them, group.Wait would park for the full
+// batcherWait and then report a timeout — hiding the shutdown and stalling the
+// caller for the whole budget.
+//
+// batcherWait is set high enough that a parked wait is unmistakable.
+func TestSetLocked_ClosedBatcherDoesNotParkForFullTimeout(t *testing.T) {
+	s := newTestStoreForGet(t)
+	s.batcherWait = 30 * time.Second
+	s.lockedBatcher = sendOnClosedBatcher[batchLocked]{}
+
+	hashes := []chainhash.Hash{{0x01}, {0x02}, {0x03}}
+
+	start := time.Now()
+
+	var err error
+
+	require.NotPanics(t, func() {
+		err = s.SetLocked(context.Background(), hashes, true)
+	}, "SetLocked must not propagate the batcher's send-on-closed-channel panic")
+
+	elapsed := time.Since(start)
+
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "shutting down")
+	require.Less(t, elapsed, 5*time.Second,
+		"every item must be completed on a rejected batch send; parking for batcherWait means the accounting is wrong")
+}
+
+// TestCreate_ClosedBatcherSurfacesShutdown drives the guard through the real
+// Create path. Store.Close closes the store batcher FIRST, so it is dead for the
+// whole remaining drain — a Create racing shutdown is the widest window of any
+// site, and previously crashed the process.
+func TestCreate_ClosedBatcherSurfacesShutdown(t *testing.T) {
+	s := newTestStoreForGet(t)
+	s.batcherWait = 30 * time.Second
+	s.storeBatcher = sendOnClosedBatcher[BatchStoreItem]{}
+
+	tx, err := bt.NewTxFromString("010000000000000000ef011c044c4db32b3da68aa54e3f30c71300db250e0b48ea740bd3897a8ea1a2cc9a020000006b483045022100c6177fa406ecb95817d3cdd3e951696439b23f8e888ef993295aa73046504029022052e75e7bfd060541be406ec64f4fc55e708e55c3871963e95bf9bd34df747ee041210245c6e32afad67f6177b02cfc2878fce2a28e77ad9ecbc6356960c020c592d867ffffffffd4c7a70c000000001976a914296b03a4dd56b3b0fe5706c845f2edff22e84d7388ac0301000000000000001976a914a4429da7462800dedc7b03a4fc77c363b8de40f588ac000000000000000024006a4c2042535620466175636574207c20707573682d7468652d627574746f6e2e617070d2c7a70c000000001976a914296b03a4dd56b3b0fe5706c845f2edff22e84d7388ac00000000")
+	require.NoError(t, err)
+
+	start := time.Now()
+
+	var createErr error
+
+	require.NotPanics(t, func() {
+		_, createErr = s.Create(context.Background(), tx, 900000)
+	}, "Create must not propagate the batcher's send-on-closed-channel panic")
+
+	require.Error(t, createErr)
+	require.Contains(t, createErr.Error(), "shutting down")
+	require.Less(t, time.Since(start), 5*time.Second,
+		"the item must be completed on a rejected send; parking for batcherWait means the accounting is wrong")
+}
+
+// TestSetDAHForChildRecords_ClosedBatcherSurfacesShutdown covers the per-item
+// guard in the setDAH loop. Each child is its own channel send, so a shutdown
+// race rejects them individually; completing each one lets the existing
+// aggregate result read report the failure rather than waiting out batcherWait.
+func TestSetDAHForChildRecords_ClosedBatcherSurfacesShutdown(t *testing.T) {
+	s := newTestStoreForGet(t)
+	s.batcherWait = 30 * time.Second
+	s.setDAHBatcher = sendOnClosedBatcher[batchDAH]{}
+
+	txID := &chainhash.Hash{0x07}
+
+	start := time.Now()
+
+	var err error
+
+	require.NotPanics(t, func() {
+		err = s.SetDAHForChildRecords(txID, 3, 900000)
+	})
+
+	require.Error(t, err)
+	require.Less(t, time.Since(start), 5*time.Second,
+		"each rejected child must be completed; parking for batcherWait means the accounting is wrong")
+}
+
+// TestIncrementSpentRecords_ClosedBatcherSurfacesShutdown covers the
+// single-item guard on the increment batcher. SpendWaitTimeout is left at zero
+// so the method's own 30s fallback applies — a parked wait would be obvious.
+func TestIncrementSpentRecords_ClosedBatcherSurfacesShutdown(t *testing.T) {
+	s := newTestStoreForGet(t)
+	s.incrementBatcher = sendOnClosedBatcher[batchIncrement]{}
+
+	txID := &chainhash.Hash{0x08}
+
+	start := time.Now()
+
+	var err error
+
+	require.NotPanics(t, func() {
+		_, err = s.IncrementSpentRecords(txID, 1, 900000)
+	})
+
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "shutting down")
+	require.Less(t, time.Since(start), 5*time.Second)
+}
+
+// TestSpend_ClosedBatcherSurfacesShutdown drives the guard through the real Spend
+// path — the hottest write path and the one this guard exists for. PutBatchCtx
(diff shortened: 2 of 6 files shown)


ctnguyen (approved):
Review at commit [91fd4fa](https://github.com/ordishs/teranode/commit/91fd4fab42f8769c95dfc29a2a3d9403ccde499e)

This closes the gap properly rather than patching around it, and the parts that were easy to get wrong are right. I checked the premise the whole change rests on against the pinned dependency rather than the comment: `PutBatchCtx` in go-batcher v2.0.6 snapshots the slice and does exactly one `b.enqueue(itemEnvelope{items: snapshot})`, and the oversize split happens later in the worker's collector — so "one send carries the whole group, a rejected send rejects every item" is accurate, and the doubt raised about it on this PR can be dropped. Completion accounting holds at all six new sites: `Spend`'s circuit-breaker fast-fails plus `toEnqueue` sum to exactly `len(spends)`, `SetLocked` completes every hash so `onError`/`failFast` captures `firstErr`, `IncrementSpentRecords` uses the correct two-argument `complete(nil, err)`, and every `complete` is CAS-guarded — which matters because `completion.Group.Done` panics on an over-count. A repository-wide search for a bare `s.*Batcher.Put*(` in non-test code now returns nothing; all eight sites funnel through the guards. Narrowing the recover is the right call and the reasoning in "On the error code" checks out — `ErrServiceUnavailable` really does survive `Spend`'s `NewUtxoError(..., JoinCapped(...))` wrap at `spend.go:509`, so `IsTransientLocalError` still holds and `netsync/manager.go:1776` / `peer_server.go:1204` stay in lock-step. And `TestSetLocked_ClosedBatcherDoesNotParkForFullTimeout` is the right shape: 30s budget against a 5s bound makes a failure unmistakable rather than marginal. Nothing below blocks the merge; the two I'd most want read are the fact that a guarded abort is still completely invisible, and that none of the new tests pin the error identity the design rationale depends on.

## Non-blocking issues

### ChiR1 - A guarded abort still emits no metric and no log, now on the two hottest write paths
**Problem:** `recoverBatcherShutdown` (`stores/utxo/aerospike/aerospike.go:145-153`) converts the panic and returns — no counter, no log line. Nothing downstream makes up for it: `Create` reads `item.result` and returns it raw (`create.go:280-283`) with no log and no metric; `SetLocked` returns `firstErr` directly (`locked.go:115-117`); `Spend` logs each failed input at `Debugf` only (`spend.go:568`). The nearest precedent goes the other way: the sibling batcher-failure exits *do* count themselves — `get.go:448,452` (`ContextCanceled`, `BatchTimeout`), `spend.go:472-474` (`Spend`/`BatchTimeout`), `spend.go:1296-1298` (`IncrementSpentRecords`/`BatchTimeout`) — so the new guard is the odd one out among the paths it sits beside. Worth being precise about the scope: plenty of other exits in this package have no counter either (`create.go:195,207`, the store batch timeout at `create.go:275`, the `SetLocked` timeout at `locked.go:121-122`), so the wider gap is pre-existing and not this PR's doing. This was raised twice against the predecessor PR and left open pending direction; the reason to settle it now is that the guard just went from 2 sites to 8, including the two highest-volume externally-driven writes.
**Why it matters:** A shutdown-race storm on `Spend`/`Create` — the exact condition this change exists for — leaves nothing to alert on and nothing to search the logs for, so the only evidence that a graceful shutdown dropped in-flight work is the caller's own error handling, which at these sites is a returned error and nothing else.
**Fix:** Deliberately *not* a single `Inc` inside the shared helper: this package uses two different counter families and the helper cannot tell which applies. Reads and the store/create path use `prometheusTxMetaAerospikeMapErrors` (`metrics.go:126`) with a CamelCase function label — `("Get", …)` at `get.go:448,452,459`, `("Store", …)` at `create.go:841-845` — while the spend/reset side uses `prometheusUtxoMapErrors` (`metrics.go:251`) with `("Spend", …)` at `spend.go:473`, `("IncrementSpentRecords", …)` at `spend.go:1297` and `("Reset", …)` in `un_spend.go`. Incrementing the UTXO-write counter from the helper would file failed reads under write errors, and passing `who` through as the label would introduce lowercase ad-hoc values (`"get"`, `"setDAH"`) alongside the existing CamelCase ones. Record at the call site instead, where both the family and the label are already known:

```go
// create.go:260
if enqueueErr := safeBatcherPutCtx(s.storeBatcher, ctx, item, "store"); enqueueErr != nil {
	prometheusTxMetaAerospikeMapErrors.WithLabelValues("Store", "BatcherClosed").Inc()
	s.logger.Warnf("[Create][%s] store batcher refused the enqueue (store shutting down)", txHash.String())
	item.complete(enqueueErr)
}

// spend.go:427
if enqueueErr := safeBatcherPutBatchCtx(s.spendBatcher, ctx, toEnqueue, "spend"); enqueueErr != nil {
	prometheusUtxoMapErrors.WithLabelValues("Spend", "BatcherClosed").Inc()
	s.logger.Warnf("[SPEND][%s] spend batcher refused %d inputs (store shutting down)", tx.TxID(), len(toEnqueue))
	for _, item := range toEnqueue {
		item.complete(enqueueErr)
	}
}
```

If you would rather keep it centralised, give the helper an explicit recorder — `safeBatcherPutCtx(b, ctx, item, who, onRejected func())` — so each site still supplies its own family and label rather than the helper guessing.

### ChiR2 - The re-panic test fakes the panic with a string, and the one path whose behaviour actually flipped is untested
**Problem:** `panickingBatcher` panics with the string literal `"nil map write"` (`get_closed_batcher_test.go:149-158`). That is the same anti-pattern the predecessor PR removed: the original error-corruption bug survived review precisely because the doubles panicked with a `string` instead of the `runtime.plainError` the runtime actually produces, which is why `sendOnClosed()` (`get_closed_batcher_test.go:18-22`) drives a real `close(ch); ch <- v`. The re-panic path is where the value class matters most — `recoverBatcherShutdown` renders `r` through `fmt.Sprint` and matches on text — yet no test exercises it with a real runtime fault. The realistic trigger is a nil batcher, and exactly one path changed behaviour here: `PreviousOutputsDecorate` (`get.go:1300-1323`) does not nil-check `s.outpointBatcher` the way `get()` checks `s.getBatcher` at `get.go:426`, so the narrowed recover now re-panics a nil `outpointBatcher` where the predecessor relabelled it as shutdown. The six sites added by this PR are *not* part of that: they were bare batcher calls on `main`, so a nil `storeBatcher`/`spendBatcher`/`lockedBatcher`/`setDAHBatcher`/`incrementBatcher` already panicked before this change. `New` is also the only non-test constructor (`aerospike.go:328`) and wires all seven unconditionally (`aerospike.go:421-508`), so none of this is a live production path — it is an untested behaviour change and a trap for anything hand-building the struct.
**Why it matters:** The regression test for "we re-panic real bugs" never sees a real bug's value type, so it cannot catch a future change to the matching logic that only misbehaves on `runtime.Error`; and the one error-to-crash flip in the diff goes in unrecorded.
**Fix:** Panic through a genuine runtime fault, and cover the path that changed:

```go
func (panickingBatcher[T]) PutCtx(context.Context, *T, ...int) {
	var m map[string]int
	m["boom"] = 1 // real runtime.Error, not a string
}

require.PanicsWithError(t, "assignment to entry in nil map", func() {
	_ = safeBatcherPutCtx[batchGetItem](bad, context.Background(), &batchGetItem{}, "get")
})

// the path whose behaviour flipped: nil outpointBatcher now crashes
s := newTestStoreForGet(t) // outpointBatcher is nil
require.Panics(t, func() { _ = s.PreviousOutputsDecorate(context.Background(), tx) })
```

If crashing there is not wanted, nil-check `s.outpointBatcher` at `get.go:1320` the way `get.go:426` does.

### ChiR3 - No test pins the error identity the design rationale depends on, and `SetDAHForChildRecords` already drops it
**Problem:** Every new shutdown assertion checks message text only — `require.Contains(err.Error(), "shutting down")` at `closed_batcher_group_test.go:38-39` (`SetLocked`), `:64-65` (`Create`), `:111-112` (`IncrementSpentRecords`), and the three helper checks at `get_closed_batcher_test.go:115-126`. `TestSpend_ClosedBatcherSurfacesShutdown` (`closed_batcher_group_test.go:121-140`) is weaker still: `require.Error` plus the elapsed bound, no message and no code. That is the site where it matters most, because `Spend` does not return the guard's error — it re-wraps as `errors.NewUtxoError(..., errors.JoinCapped(maxAggregatedSpendErrs, failedSpends...))` at `spend.go:509`. I verified the link survives (`Join` chains it, `(*Error).Is` walks the chain, `IsTransientLocalError` at `errors/error_utils.go:69-77` returns true), but nothing holds it in place, and the whole "On the error code" argument rests on it. The convention already exists in this package — `spend_test.go:454` asserts `errors.Is(err, errors.ErrServiceUnavailable)` for the `IncrementSpentRecords` timeout — the new tests just did not follow it. Separately, `SetDAHForChildRecords` has already lost the identity: `spend.go:1107-1109` returns a bare `errors.NewStorageError("[setDAHForChildRecords][%s] failed to set DAH for one or more child records", txID.String())` with no wrapped cause, so `ErrServiceUnavailable` is gone from the chain and the message never says "shutting down". It still satisfies `IsTransientLocalError` only because `ErrStorageError` is separately a member of that set — and that is visibly why its own test (`closed_batcher_group_test.go:89-91`) is the only one that cannot assert `"shutting down"`.
**Why it matters:** A future refactor of `Spend`'s aggregation — flattening the chain, switching to a plain `errors.New`, capping differently — would silently reclassify a shutdown race as a peer fault: `netsync/manager.go:1776` would stop suppressing `PushRejectMsg(..., wire.RejectInvalid, ...)` and `peer_server.go:1204` would drop the delivering peer, both because *our* node was shutting down. It would pass the suite as written, and message-text assertions would not notice.
**Fix:** Assert the class, not the wording, on every one of these — cheap, and it is the property the PR is arguing for:

```go
// Spend — survives the JoinCapped aggregation
require.True(t, errors.IsTransientLocalError(spendErr),
	"a shutdown-race spend must stay transient-local, or netsync rejects the block to the peer")

// Create / SetLocked / IncrementSpentRecords — the guard's error surfaces directly
require.ErrorIs(t, err, errors.ErrServiceUnavailable)
```

and carry the cause in the child aggregation so the shutdown reason is not erased (`firstChildErr` being the first non-nil `item.result`):

```go
// spend.go:1107
return errors.NewStorageError("[setDAHForChildRecords][%s] failed to set DAH for one or more child records", txID.String(), firstChildErr)
```

### ChiR4 - The new `handleExtraRecords` guard has no closed-batcher test
**Problem:** The drift-clear branch itself is covered — `TestHandleExtraRecords_DAHSetWithDrift` (`verify_children_test.go:60-115`) and `TestDriftedCounterDoesNotSetDAH` (`spend_test.go:457-549`) both drive it — but both are cluster-backed and neither exercises a refused enqueue, so the guard added at `spend.go:1171-1173` is the only one of the six with no test. The behaviour as written is correct (`item.complete(enqueueErr)`, `group.Wait` then returns immediately, and `item.result` is logged at `spend.go:1177-1178`), and returning `nil` afterwards is deliberate best-effort cleanup that matches the pre-conversion behaviour — so this is a coverage gap, not a defect.
**Why it matters:** This is the branch that exists to undo a counter-drift DAH; if a later refactor broke the completion here it would park for `batcherWait` or leave the group unbalanced, and the only tests touching the branch need a cluster to run. The five sibling guards each got a fast unit test; this one is the odd omission.
**Fix:** Same shape as the others, using the existing double:

```go
s := newTestStoreForGet(t)
s.setDAHBatcher = sendOnClosedBatcher[batchDAH]{}
s.batcherWait = 30 * time.Second
// drive the drift branch; assert it does not panic and returns well inside batcherWait
```

Worth distinguishing the two causes in the log line too — `"failed to clear drifted master DAH (enqueue refused): %v"` — since a refused enqueue and a failed write are very different operator situations.

### ChiR5 - `SetDAHForChildRecords` repeats the panic/recover and the error log once per child after the first refusal
**Problem:** The guard sits inside the loop (`spend.go:1070-1083`) with no memory of a previous refusal. These are per-child `Put` calls, so once the batcher is closed every remaining child repeats the full send → panic → recover → `fmt.Sprint` → `strings.Contains` cycle, and because each one sets `item.result`, the post-wait loop at `spend.go:1099-1105` then emits one `logger.Errorf` per child for the same cause. The aggregate result is still correct — every child is completed exactly once and `group.Wait` returns promptly — so this is cost, not a defect.
**Why it matters:** Panic/recover is orders of magnitude more expensive than a channel send, and it lands during shutdown when the log is already busy; a paginated tx turns one event into `childCount` recoveries and `childCount` identical Error lines.
**Fix:** Make the refusal sticky, and de-duplicate the *logging* only — not the scan. Because the sends are per-child, a `Close` can land mid-loop, so `items` can legitimately hold a mix: children the batcher already accepted (whose results the drain still delivers, possibly a genuine write failure) followed by children it refused. Skipping the scan on `enqueueErr != nil` would swallow those real failures, so keep the scan and suppress only the repeated lines. Recording the index of the first refusal makes the split exact, with no error comparison:

```go
var (
	enqueueErr  error
	refusedFrom = -1 // first child the batcher refused; sticky from there on
)

for i := uint32(0); i < uint32(childCount); i++ { // nolint: gosec
	item := &batchDAH{txID: txID, childIdx: i + 1, deleteAtHeight: dah, group: group}
	items[i] = item

	if refusedFrom >= 0 { // already refused; don't re-panic per child
		item.complete(enqueueErr)
		continue
	}

	if err := safeBatcherPut(s.setDAHBatcher, item, "setDAH"); err != nil {
		enqueueErr, refusedFrom = err, int(i)
		item.complete(err)
	}
}

// … after group.Wait …

var firstErr error

for i, item := range items {
	if item.result == nil {
		continue
	}

	if firstErr == nil {
		firstErr = item.result
	}

	// Refused children all share one cause; summarised once below. Children the
	// batcher accepted before the refusal can still carry a real write failure
	// and must each be reported.
	if refusedFrom >= 0 && i >= refusedFrom {
		continue
	}

	s.logger.Errorf("[setDAHForChildRecords][%s] failed to set DAH for child record %d: %v", txID.String(), item.childIdx, item.result)
}

if refusedFrom >= 0 {
	s.logger.Errorf("[setDAHForChildRecords][%s] setDAH batcher refused %d of %d child records: %v", txID.String(), childCount-refusedFrom, childCount, enqueueErr)
}

if firstErr != nil {
	return errors.NewStorageError("[setDAHForChildRecords][%s] failed to set DAH for one or more child records", txID.String(), firstErr)
}
```

Carrying `firstErr` into the returned error also covers the `SetDAHForChildRecords` half of ChiR3, since a shutdown refusal then reaches the caller with its cause intact.

### ChiR6 - The single-send invariant the guard depends on is not documented where a reader looks for it
**Problem:** `safeBatcherPutBatchCtx`'s contract — "PutBatch* enqueues the whole group in a single channel send, so a rejected send rejects every item" (`aerospike.go:182-185`) — is asserted but not sourced, and the `batcherIfc` doc 70 lines above (`aerospike.go:111-115`) says a group larger than the configured size "is split into full-size batches" without saying *where*. Read together they look like they might contradict: if the split were realized as multiple sends, a `Close` landing mid-loop would enqueue some sub-batches before panicking and "rejects every item" would not hold. It does hold — in go-batcher v2.0.6 `PutBatchCtx` copies the slice and performs one `b.enqueue(itemEnvelope[T]{items: snapshot, sc: ...})`, and the collector does the splitting on the worker goroutine — but establishing that costs a trip to the module cache.
**Why it matters:** This is the invariant the completion accounting at three call sites is built on. The apparent contradiction has already been queried once on this PR, so the next reader will query it too, and if go-batcher ever moved the split to the producer side the guard would break silently.
**Fix:** One clause on the `batcherIfc` doc:

```go
// PutBatch / PutBatchCtx enqueue a whole group of items in a single channel
// send instead of one send per item. […] Batch size is always honoured — a
// group larger than the configured size is split into full-size batches by the
// worker's collector, never by the producer, so the enqueue itself stays a
// single send (which is what lets safeBatcherPutBatchCtx treat a refused send
// as rejecting every item).
```

### ChiR7 - Stale go-batcher version in the same file this PR rewrites
**Problem:** `aerospike.go:706` still says "go-batcher v2.0.4"; `go.mod:197` pins v2.0.6. It is now the only remaining occurrence in the package — `get.go:471` and `get_closed_batcher_test.go:25,83` say v2.0.6, and the guard doc block this PR rewrites (`aerospike.go:127-144`) correctly says "go-batcher v2". The author offered to fix this in passing during the predecessor review.
**Why it matters:** It reads as "verified against 2.0.4" to whoever next audits whether the upstream panic still exists — which is the question that decides if these guards can eventually be deleted — and it is now the odd one out in a file whose other version references were just corrected.
**Fix:** Change it to "go-batcher v2", matching the wording the new block uses, so it does not go stale again.

## Recap

| ChiR1 | Guarded abort emits nothing | | 32% |
| ChiR2 | Re-panic faked with string | | 30% |
| ChiR3 | Error class unpinned by tests | | 28% |
| ChiR4 | No closed-batcher test for drift clear | | 13% |
| ChiR5 | Per-child recover and log after refusal | | 11% |
| ChiR6 | Single-send invariant undocumented | | 10% |
| ChiR7 | Stale go-batcher version ref | | 6% |

ordishs:
Closing as redundant — the content is already on `main`.

This PR existed because #1361 was merged at an earlier commit than the branch tip, so the follow-up commit (guarding the remaining six enqueue sites, plus narrowing the recover) never reached `main`. It has since arrived by another route: #1363's branch was stacked on top of this one, so squash-merging #1363 (`0673ae2ae`) carried this change in with it.

Verified rather than assumed — every file this PR touches is now byte-identical on `main` (compared via `git hash-object`):

| file | vs main |
|---|---|
| stores/utxo/aerospike/aerospike.go | identical |
| stores/utxo/aerospike/spend.go | identical |
| stores/utxo/aerospike/create.go | identical |
| stores/utxo/aerospike/locked.go | identical |
| stores/utxo/aerospike/get_closed_batcher_test.go | identical |
| stores/utxo/aerospike/closed_batcher_group_test.go | identical |

And the crash this addressed is gone from `main` — no bare `s.*Batcher.Put*(` remains in non-test code:

| file | guarded | unguarded |
|---|---|---|
| spend.go | 4 | 0 |
| create.go | 1 | 0 |
| locked.go | 1 | 0 |
| get.go | 3 | 0 |

Nothing is lost by closing: the guards, the completion accounting on batch rejection, the narrowed recover, and all the tests (including the coverage work that took new-code coverage from 79.3% to 96.6% here) are on `main`.

One note for future stacked PRs, since this whole detour came from it: GitHub reported a stale head for #1361 for a sustained period, and merging against that head silently dropped a commit. Worth checking `gh pr view <n> --json headRefOid` against the local SHA before merging a stacked PR.

PR #1377, opened 2026-07-29
fix(asset): stop serving and caching an empty subtree_data as a valid 200

oskarszoon:
Part of #1368 — the serving side. The client-side half merged in #1371; the log-storm fix follows separately.

## Why

A peer answered `subtree_data` with `HTTP 200` and an **empty body**, its nginx `proxy_cache` stored that as a valid response, and it replayed it (`x-cache-status: HIT`, `content-length: 0`, `cache-control: public, max-age=300`) to every requester for the whole TTL. A 32768-byte subtree is 1024 tx hashes, so its `subtree_data` cannot legitimately be empty.

`streamOrAbort` already defends against *truncated* streams: on a mid-stream failure it hijacks the connection and closes it **without** the chunked terminator, so nginx sees an incomplete chunked response and refuses to commit it to cache. Two holes let the empty case through anyway.

## Hole 1 — the status line was committed before the first byte was read

`streamOrAbort` called `WriteHeader(200)` up front, so a source that yielded zero bytes — an on-demand generation that failed before its first write, or an empty stored file — went out as a perfectly cacheable `200` with an empty body. The truncation defense never engaged because there was nothing to truncate.

Now the helper peeks one byte with `io.ReadFull` *before* committing, and an empty source returns `500` instead. `proxy_cache_valid` covers only `200`, so a 500 is never stored. The peeked byte is written, not swallowed.

This applies to `subtree`, `subtree_data`, `block_legacy` and the legacy mining candidate — none of which can legitimately return an empty body. The mid-stream hijack path is untouched.

## Hole 2 — nginx was talking HTTP/1.0 upstream, so the truncation defense was inert in production

`asset-cache-nginx.conf` never set `proxy_http_version`, and nginx defaults to **HTTP/1.0** for `proxy_pass`. On HTTP/1.0 a body without a Content-Length ends at connection close — which makes a deliberately aborted stream **indistinguishable from a complete one**. nginx cached truncated and empty bodies as valid 200s regardless of the hijack.

With `proxy_http_version 1.1` the response is chunk-framed, so closing without the terminator produces `upstream prematurely closed connection while reading upstream` and nginx discards the cache entry — the behaviour `streamOrAbort` was written for. Scoped to the cached location block so the blast radius stays on the endpoints that actually cache.

`proxy_set_header Connection ""` comes along per the usual 1.1 pairing. Note this conf has no `upstream {}` block with `keepalive`, and `proxy_pass http://$asset_backend` uses a variable, which disables the upstream connection cache anyway — so the directive is harmless here but buys no reuse. The comment says so, to save the next reader the same dig.

## What this does not fix

**This PR alone cannot unstick an already-poisoned peer, and that is structural, not a deployment-timing artefact.** `proxy_cache_use_stale ... http_500` means a peer holding a poisoned entry serves the **stale empty 200** rather than the new 500 once the TTL lapses, `proxy_cache_background_update on` keeps refreshes in the background, and each access refreshes the `inactive=10m` timer. Escaping it needs either the client-side cache-bust from #1371 — a busted key has no stale entry — or an operator flushing `/tmp/nginx/cache` on the affected peer.

Related: if a zero-byte `subtreeData` file was ever committed to a peer's store, that peer now returns 500 for that hash permanently instead of an empty 200. Better for clients (explicit failure, peer charged, alternatives tried) but it turns a silently-wrong condition into one that needs cleanup.

## Testing

`services/asset/...` green under `-race`; `streamOrAbort` unit tests cover the empty body (asserting the status line stays uncommitted and the recorder body is empty), the single-byte boundary, and a handler-level empty-pipe case; the four pre-existing `streamOrAbort` tests are unmodified and still pass. `nginx -t` reports the conf syntactically valid. `go build`, `go vet`, `make lint` clean.

**Outstanding manual verification:** live cache-behaviour against a running stack was not performed — it needs real chain data. The check that matters is `upstream prematurely closed connection while reading upstream` appearing in nginx's error log on an aborted stream, with no cached empty entry afterwards; also worth capturing `uht=` in the `cache_log` before and after, since the deferred header commit shifts first-byte latency into upstream header time. If someone has a stack with chain data, that is the one thing left to confirm.

## Known limitation, not addressed here

`GetSubtreeData` and `GetSubtree` both increment their `OK`/`200` Prometheus label before streaming, so an empty-body 500 is still counted as OK. That mislabelling already existed for mid-stream failures; fixing it is a separate change.

### DIFF
--- a/deploy/docker/base/asset-cache-nginx.conf
+++ b/deploy/docker/base/asset-cache-nginx.conf
@@ -88,8 +88,14 @@ http {
       # signal only exists on HTTP/1.1. With 1.1, nginx logs "upstream prematurely
       # closed connection while reading upstream" and discards the cache entry.
       #
-      # Connection "" clears the hop-by-hop Connection header so keepalive is not
-      # forced closed per request.
+      # Connection "" clears the hop-by-hop Connection header, the standard idiom
+      # paired with proxy_http_version 1.1. It buys no connection reuse HERE, though:
+      # nginx's keepalive pool needs a named upstream {} block with "keepalive N;",
+      # and proxy_pass to a variable (http://$asset_backend) disables the upstream
+      # connection cache outright — nginx re-resolves DNS and opens a fresh TCP
+      # connection per request either way. The directive's only effect is that nginx
+      # stops sending an explicit "Connection: close" to the backend. Harmless and
+      # correct, but do not read a keepalive saving into it.
       proxy_http_version 1.1;
       proxy_set_header Connection "";
 
--- a/deploy/docker/base/asset-cache-nginx.conf
+++ b/deploy/docker/base/asset-cache-nginx.conf
@@ -99,6 +99,16 @@ http {
       # If cache lock times out, serve from backend anyway
       proxy_cache_lock_timeout 5s;
 
+      # Narrower than the http{}-level default (which includes http_500): a 500 from
+      # asset on these routes is a content fault — an empty or failed generation for
+      # one specific hash — not a transient backend outage, so re-invoking the backend
+      # cannot change the answer. With proxy_next_upstream_tries 3 that would cost up
+      # to three backend invocations per client request. Harmless in the default
+      # single-container topology, where proxy_pass to a variable yields one peer and
+      # the retry cannot fire, but real once asset is scaled and Docker DNS returns
+      # several A records — and worst exactly while the fleet is under catchup load.
+      proxy_next_upstream error timeout http_502 http_503 http_504;
+
       # Talk HTTP/1.1 upstream so responses are chunk-framed.
       #
       # nginx defaults to HTTP/1.0 for proxy_pass, where a body without a
--- a/deploy/docker/base/asset-cache-nginx.conf
+++ b/deploy/docker/base/asset-cache-nginx.conf
@@ -35,6 +35,29 @@ http {
   proxy_no_cache 0;      # 0 = do cache (do not disable caching)
   proxy_cache_bypass 0;  # 0 = use cache (do not bypass)
 
+  # Freshness label for the cached location, derived from the upstream status.
+  #
+  # This nginx never stores a failure (proxy_cache_valid lists 200 only), but that is
+  # a property of THIS hop, not of the response. RFC 9111 section 3 lets a shared
+  # cache store a 5xx precisely when explicit freshness information is present, so
+  # labelling a 500 "public, max-age=300" licenses any second cache in the path (an
+  # operator CDN or ingress in front of this container, a second asset-cache layer, a
+  # proxy between peers) to store and replay it for five minutes. The asset service
+  # now answers a zero-byte source with 500 rather than a cacheable empty 200
+  # (services/asset/httpimpl/stream.go streamOrAbort), which makes 5xx an expected
+  # outcome of these routes, so the failure path has to be labelled no-store — issue
+  # 1368's suggested fix 0.
+  #
+  # The "" arm is required: on a cache HIT no upstream is contacted and
+  # $upstream_status is empty, which must stay cacheable. A comma-separated
+  # $upstream_status (possible when proxy_next_upstream retried) and a revalidated
+  # 304 both fall to default, i.e. no-store — conservative, never the other way.
+  map $upstream_status $asset_cache_control {
+    default "no-store";              # 4xx/5xx: never store, never replay
+    200     "public, max-age=300";
+    ""      "public, max-age=300";   # cache HIT — no upstream was contacted
+  }
+
   # Set timeouts for upstream connections
   proxy_connect_timeout 5s;
   proxy_send_timeout 60s;  # Increased for large request bodies
@@ -107,7 +130,11 @@ http {
       proxy_set_header X-Forwarded-Proto $scheme;
 
       add_header X-Cache-Status $upstream_cache_status always;
-      add_header Cache-Control "public, max-age=300" always;
+
+      # always: attach on every status, including the ones nginx generates itself.
+      # The value is conditional (see the $asset_cache_control map) so a 4xx/5xx is
+      # labelled no-store instead of inheriting the 300s freshness lifetime.
+      add_header Cache-Control $asset_cache_control always;
     }
 
     location ~ ^/api/v1/(txs|subtree/[0-9a-fA-F]+/txs)$ {
--- a/deploy/docker/base/asset-cache-nginx.conf
+++ b/deploy/docker/base/asset-cache-nginx.conf
@@ -76,6 +76,23 @@ http {
       # If cache lock times out, serve from backend anyway
       proxy_cache_lock_timeout 5s;
 
+      # Talk HTTP/1.1 upstream so responses are chunk-framed.
+      #
+      # nginx defaults to HTTP/1.0 for proxy_pass, where a body without a
+      # Content-Length ends at connection close. That makes an aborted upstream
+      # stream indistinguishable from a complete one, so nginx caches truncated and
+      # empty bodies as valid 200s — issue 1368, where a 0-byte subtree_data was
+      # cached and replayed (x-cache-status: HIT) until catchup stalled. The asset
+      # service already defends against this by closing the connection WITHOUT the
+      # chunked terminator (services/asset/httpimpl/stream.go streamOrAbort), but that
+      # signal only exists on HTTP/1.1. With 1.1, nginx logs "upstream prematurely
+      # closed connection while reading upstream" and discards the cache entry.
+      #
+      # Connection "" clears the hop-by-hop Connection header so keepalive is not
+      # forced closed per request.
+      proxy_http_version 1.1;
+      proxy_set_header Connection "";
+
       proxy_pass http://$asset_backend;
 
       proxy_set_header Host $host;
--- a/deploy/docker/base/asset-cache-nginx.conf
+++ b/deploy/docker/base/asset-cache-nginx.conf
@@ -52,6 +52,14 @@ http {
   # $upstream_status is empty, which must stay cacheable. A comma-separated
   # $upstream_status (possible when proxy_next_upstream retried) and a revalidated
   # 304 both fall to default, i.e. no-store — conservative, never the other way.
+  #
+  # Accepted tradeoff: when proxy_cache_use_stale serves a stale-but-good 200
+  # because the upstream attempt failed, $upstream_status is that failed attempt's
+  # 5xx, so a valid body goes out labelled no-store and downstream caches will not
+  # store it. Only for the duration of the upstream-failure window, and it fails in
+  # the safe direction. Distinguishing it needs a composite
+  # "$upstream_cache_status|$upstream_status" key, which is more ways to get the
+  # labelling wrong than the marginal cacheability is worth.
   map $upstream_status $asset_cache_control {
     default "no-store";              # 4xx/5xx: never store, never replay
     200     "public, max-age=300";
--- a/services/asset/httpimpl/GetLegacyBlock.go
+++ b/services/asset/httpimpl/GetLegacyBlock.go
@@ -92,9 +92,17 @@ func (h *HTTP) GetLegacyBlock() func(c echo.Context) error {
 		// ever closes the read end.
 		defer r.Close()
 
+		// Outcome counted after streaming — see the note in GetSubtreeData for why, and
+		// for which cases this still labels as OK.
+		if err := streamOrAbort(c, http.StatusOK, echo.MIMEOctetStream, r); err != nil {
+			prometheusAssetHTTPGetBlockLegacy.WithLabelValues("ERROR", http.StatusText(http.StatusInternalServerError)).Inc()
+
+			return err
+		}
+
 		prometheusAssetHTTPGetBlockLegacy.WithLabelValues("OK", "200").Inc()
 
-		return streamOrAbort(c, http.StatusOK, echo.MIMEOctetStream, r)
+		return nil
 	}
 }
 
@@ -152,8 +160,15 @@ func (h *HTTP) GetRestLegacyBlock() func(c echo.Context) error {
 		// goroutine when the response write failed and streamOrAbort hijacked.
 		defer r.Close()
 
+		// Outcome counted after streaming — see the note in GetSubtreeData.
+		if err := streamOrAbort(c, http.StatusOK, echo.MIMEOctetStream, r); err != nil {
+			prometheusAssetHTTPGetBlockLegacy.WithLabelValues("ERROR", http.StatusText(http.StatusInternalServerError)).Inc()
+
+			return err
+		}
+
 		prometheusAssetHTTPGetBlockLegacy.WithLabelValues("OK", "200").Inc()
 
-		return streamOrAbort(c, http.StatusOK, echo.MIMEOctetStream, r)
+		return nil
 	}
 }
--- a/services/asset/httpimpl/GetLegacyBlock.go
+++ b/services/asset/httpimpl/GetLegacyBlock.go
@@ -84,6 +84,14 @@ func (h *HTTP) GetLegacyBlock() func(c echo.Context) error {
 			}
 		}
 
+		// The reader is a *io.PipeReader fed by a goroutine. Closing it unblocks that
+		// goroutine with io.ErrClosedPipe, which it already handles. Without this, a
+		// stream that fails because the write to the response failed (client gone) leaves
+		// the producer blocked in Write forever, holding its arena and any file-store read
+		// permit — streamOrAbort hijacks and returns nil in that case, so nothing else
+		// ever closes the read end.
+		defer r.Close()
+
 		prometheusAssetHTTPGetBlockLegacy.WithLabelValues("OK", "200").Inc()
 
 		return streamOrAbort(c, http.StatusOK, echo.MIMEOctetStream, r)
@@ -139,6 +147,10 @@ func (h *HTTP) GetRestLegacyBlock() func(c echo.Context) error {
 				return echo.NewHTTPError(http.StatusInternalServerError, errors.NewProcessingError("error getting block", err).Error())
 			}
 		}
+
+		// See GetLegacyBlock: closing the pipe read end is what unblocks the producer
+		// goroutine when the response write failed and streamOrAbort hijacked.
+		defer r.Close()
 
 		prometheusAssetHTTPGetBlockLegacy.WithLabelValues("OK", "200").Inc()
 
--- a/services/asset/httpimpl/GetSubtree.go
+++ b/services/asset/httpimpl/GetSubtree.go
@@ -124,7 +124,11 @@ func (h *HTTP) GetSubtree(mode ReadMode) func(c echo.Context) error {
 			return echo.NewHTTPError(http.StatusBadRequest, errors.NewInvalidArgumentError("invalid hash string", err).Error())
 		}
 
-		prometheusAssetHTTPGetSubtree.WithLabelValues("OK", "200").Inc()
+		// The outcome is counted per return path below rather than once up front. Counting
+		// OK/200 here filed every later failure — including the empty-source 500 the
+		// streaming modes can now return — in the success bucket. Every path that used to
+		// be counted by this single increment still increments exactly once; the two hash
+		// validation errors above return before it and were never counted, unchanged.
 
 		// sign the response, if the private key is set, ignore error
 		// do this before any output is sent to the client, this adds a signature to the response header
@@ -135,14 +139,20 @@ func (h *HTTP) GetSubtree(mode ReadMode) func(c echo.Context) error {
 		if mode == JSON {
 			offset, limit, err := h.getLimitOffset(c)
 			if err != nil {
+				prometheusAssetHTTPGetSubtree.WithLabelValues("ERROR", http.StatusText(http.StatusBadRequest)).Inc()
+
 				return echo.NewHTTPError(http.StatusBadRequest, err.Error())
 			}
 
 			subtree, offset, totalNodes, err := h.repository.GetSubtreePage(ctx, hash, offset, limit)
 			if err != nil {
 				if errors.Is(err, errors.ErrNotFound) || strings.Contains(err.Error(), "not found") {
+					prometheusAssetHTTPGetSubtree.WithLabelValues("ERROR", http.StatusText(http.StatusNotFound)).Inc()
+
 					return echo.NewHTTPError(http.StatusNotFound, err.Error())
 				} else {
+					prometheusAssetHTTPGetSubtree.WithLabelValues("ERROR", http.StatusText(http.StatusInternalServerError)).Inc()
+
 					return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
 				}
 			}
@@ -163,10 +173,14 @@ func (h *HTTP) GetSubtree(mode ReadMode) func(c echo.Context) error {
 				},
 			}
 
+			prometheusAssetHTTPGetSubtree.WithLabelValues("OK", "200").Inc()
+
 			return c.JSONPretty(200, response, "  ")
 		}
 
 		if mode != BINARY_STREAM && mode != HEX {
+			prometheusAssetHTTPGetSubtree.WithLabelValues("ERROR", http.StatusText(http.StatusBadRequest)).Inc()
+
 			return echo.NewHTTPError(http.StatusBadRequest, errors.NewInvalidArgumentError("bad read mode").Error())
 		}
 
@@ -175,21 +189,45 @@ func (h *HTTP) GetSubtree(mode ReadMode) func(c echo.Context) error {
 		subtreeReader, err := h.repository.GetSubtreeNodeHashesReader(ctx, hash)
 		if err != nil {
 			if errors.Is(err, errors.ErrNotFound) || strings.Contains(err.Error(), "not found") {
+				prometheusAssetHTTPGetSubtree.WithLabelValues("ERROR", http.StatusText(http.StatusNotFound)).Inc()
+
 				return echo.NewHTTPError(http.StatusNotFound, err.Error())
 			} else {
+				prometheusAssetHTTPGetSubtree.WithLabelValues("ERROR", http.StatusText(http.StatusInternalServerError)).Inc()
+
 				return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
 			}
 		}
 		defer subtreeReader.Close()
 
 		if mode == BINARY_STREAM {
-			return streamOrAbort(c, http.StatusOK, echo.MIMEOctetStream, subtreeReader)
+			// Outcome counted after streaming — see the note in GetSubtreeData for why, and
+			// for which cases this still labels as OK.
+			if err = streamOrAbort(c, http.StatusOK, echo.MIMEOctetStream, subtreeReader); err != nil {
+				prometheusAssetHTTPGetSubtree.WithLabelValues("ERROR", http.StatusText(http.StatusInternalServerError)).Inc()
+
+				return err
+			}
+
+			prometheusAssetHTTPGetSubtree.WithLabelValues("OK", "200").Inc()
+
+			return nil
 		}
 
-		// mode == HEX (validated above)
+		// mode == HEX (validated above). This route commits the status line before its
+		// first byte, so it cannot report an empty source as anything but a 200 — see the
+		// scope note on streamOrAbort. It is outside the cached nginx location.
 		c.Response().Header().Set(echo.HeaderContentType, echo.MIMETextPlainCharsetUTF8)
 		c.Response().WriteHeader(http.StatusOK)
-		_, err = io.Copy(hex.NewEncoder(c.Response()), subtreeReader)
-		return err
+
+		if _, err = io.Copy(hex.NewEncoder(c.Response()), subtreeReader); err != nil {
+			prometheusAssetHTTPGetSubtree.WithLabelValues("ERROR", http.StatusText(http.StatusInternalServerError)).Inc()
+
+			return err
+		}
+
+		prometheusAssetHTTPGetSubtree.WithLabelValues("OK", "200").Inc()
+
+		return nil
 	}
 }
--- a/services/asset/httpimpl/GetSubtreeData.go
+++ b/services/asset/httpimpl/GetSubtreeData.go
@@ -78,8 +78,23 @@ func (h *HTTP) GetSubtreeData() func(c echo.Context) error {
 		}
 		defer r.Close()
 
+		// Count the outcome AFTER streaming, not before. Incrementing OK/200 up front
+		// filed every empty-source 500 in the success bucket, leaving no signal for
+		// "this node is serving 500s for subtree_data" — the first thing an operator
+		// would look at when validating the issue 1368 fix in production.
+		//
+		// Still not exact: streamOrAbort also returns nil on the mid-stream hijack path
+		// and on the client-gone-before-first-byte path, so those are counted as OK.
+		// Labelling them needs the helper to report which path it took, which is a
+		// bigger change than this metric fix warrants.
+		if err := streamOrAbort(c, http.StatusOK, echo.MIMEOctetStream, r); err != nil {
+			prometheusAssetHTTPGetSubtreeData.WithLabelValues("ERROR", http.StatusText(http.StatusInternalServerError)).Inc()
+
+			return err
+		}
+
 		prometheusAssetHTTPGetSubtreeData.WithLabelValues("OK", "200").Inc()
 
-		return streamOrAbort(c, http.StatusOK, echo.MIMEOctetStream, r)
+		return nil
 	}
 }
--- a/services/asset/httpimpl/GetSubtreeData_test.go
+++ b/services/asset/httpimpl/GetSubtreeData_test.go
@@ -123,4 +123,28 @@ func TestGetSubtreeData(t *testing.T) {
 		assert.Equal(t, http.StatusInternalServerError, echoErr.Code)
 		assert.Equal(t, "PROCESSING (4): error getting subtree", echoErr.Message)
 	})
+
+	t.Run("empty body is a server error, never a cacheable 200", func(t *testing.T) {
+		httpServer, mockRepo, echoContext, responseRecorder := GetMockHTTP(t, nil)
+
+		reader, writer := io.Pipe()
+
+		go func() {
+			// An on-demand generation that produced nothing: closed with no bytes.
+			_ = writer.Close()
+		}()
+
+		mockRepo.On("GetSubtreeDataReader", mock.Anything, mock.Anything, mock.Anything).Return(reader, nil)
+
+		echoContext.SetPath("/subtree_data/:hash")
+		echoContext.SetParamNames("hash")
+		echoContext.SetParamValues("9d45ad79ad3c6baecae872c0e35022d60c3bbbd024ccce06690321ece15ea995")
+
+		err := httpServer.GetSubtreeData()(echoContext)
+
+		echoErr := &echo.HTTPError{}
+		require.True(t, errors.As(err, &echoErr))
+		require.Equal(t, http.StatusInternalServerError, echoErr.Code)
+		require.Empty(t, responseRecorder.Body.String())
+	})
 }
--- a/services/asset/httpimpl/mining_candidate_legacy_block.go
+++ b/services/asset/httpimpl/mining_candidate_legacy_block.go
@@ -63,5 +63,9 @@ func (h *HTTP) handleMiningCandidateLegacyBlock(c echo.Context) error {
 		return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
 	}
 
+	// See GetLegacyBlock: closing the pipe read end is what unblocks the producer
+	// goroutine when the response write failed and streamOrAbort hijacked.
+	defer r.Close()
+
 	return streamOrAbort(c, http.StatusOK, echo.MIMEOctetStream, r)
 }
--- a/services/asset/httpimpl/stream.go
+++ b/services/asset/httpimpl/stream.go
@@ -50,7 +50,10 @@ import (
 // truncation as io.ErrUnexpectedEOF in its body parse). A source that
 // yields zero bytes now returns an *echo.HTTPError with status 500 and
 // never commits a status line, so callers must not have written headers
-// before calling.
+// before calling. A client that disconnects before the first byte returns
+// nil through the same hijack-and-close, again without committing a status
+// line — that is a client fault, not a server one, and surfacing it as a
+// 500 would only log an error twice for a peer that is already gone.
 //
 // # HTTP/2 limitation
 //
@@ -75,9 +78,29 @@ func streamOrAbort(c echo.Context, code int, contentType string, r io.Reader) er
 	// 500, which the cache configuration (proxy_cache_valid 200 5m) never stores.
 	var first [1]byte
 
+	// With a one-byte buffer io.ReadFull returns a nil error only when n == 1, and
+	// io.ErrUnexpectedEOF only for 0 < n < 1 — impossible. So inside n == 0, readErr is
+	// always non-nil and the condition is simply "which failure": a client that went
+	// away, or an exhausted source. Grow the buffer and both assumptions break.
 	n, readErr := io.ReadFull(r, first[:])
 	if n == 0 {
-		if readErr == nil || errors.Is(readErr, io.EOF) || errors.Is(readErr, io.ErrUnexpectedEOF) {
+		// Client (or proxy) went away before the source produced anything. Not a server
+		// fault: the request context is cancelled, the producer returns ctx.Err() or the
+		// pipe is closed, and the blocked read above surfaces that. Answering 500 would
+		// make customHTTPErrorHandler log an ERROR and then fail to write JSON to a peer
+		// that is already gone — a second ERROR — for every cancelled in-flight stream,
+		// and catchup cancels in batches. Mirror the debug-level "client gone"
+		// segregation in dualStreamWithFileCreation instead: hijack and close, which
+		// leaves the response uncommitted so nothing cacheable is finalised.
+		if errors.IsContextError(readErr) || errors.Is(readErr, io.ErrClosedPipe) {
+			if conn, _, hjErr := c.Response().Hijack(); hjErr == nil {
+				_ = conn.Close()
+				return nil
+			}
+			// h2: no hijack available, fall through to the 500.
+		}
+
+		if errors.Is(readErr, io.EOF) {
 			return echo.NewHTTPError(http.StatusInternalServerError, "empty response body")
 		}
 
--- a/services/asset/httpimpl/stream.go
+++ b/services/asset/httpimpl/stream.go
@@ -2,7 +2,9 @@ package httpimpl
 
 import (
 	"io"
+	"net/http"
 
+	"github.com/bsv-blockchain/teranode/errors"
 	"github.com/labstack/echo/v4"
 )
 
@@ -45,7 +47,10 @@ import (
 // echo signals "response already finalised, don't touch it." The error
 // from io.Copy is consumed inside this helper (an abrupt hijack + close
 // is sufficient signalling to the client, which will surface the
-// truncation as io.ErrUnexpectedEOF in its body parse).
+// truncation as io.ErrUnexpectedEOF in its body parse). A source that
+// yields zero bytes now returns an *echo.HTTPError with status 500 and
+// never commits a status line, so callers must not have written headers
+// before calling.
 //
 // # HTTP/2 limitation
 //
@@ -59,11 +64,31 @@ import (
 // hijack path is the one that runs; only a client (or caching proxy)
 // speaking h2 directly to the asset service hits the fallback.
 func streamOrAbort(c echo.Context, code int, contentType string, r io.Reader) error {
+	// Peek one byte BEFORE committing the status line. A source that yields zero bytes
+	// — an empty subtreeData file, or an on-demand generation that failed before its
+	// first write — would otherwise be sent as "200 OK + empty body", which a caching
+	// reverse proxy stores as a valid response and replays for the whole TTL. That is
+	// the poisoning in issue 1368: a 32768-byte subtree (1024 tx hashes) whose
+	// subtree_data came back 200 with content-length 0, cached, stalling every
+	// requester's catchup. None of the endpoints using this helper can legitimately
+	// return an empty body, so an empty source is a server-side fault: report it as
+	// 500, which the cache configuration (proxy_cache_valid 200 5m) never stores.
+	var first [1]byte
+
+	n, readErr := io.ReadFull(r, first[:])
+	if n == 0 {
+		if readErr == nil || errors.Is(readErr, io.EOF) || errors.Is(readErr, io.ErrUnexpectedEOF) {
+			return echo.NewHTTPError(http.StatusInternalServerError, "empty response body")
+		}
+
+		return echo.NewHTTPError(http.StatusInternalServerError, readErr.Error())
+	}
+
 	h := c.Response().Header()
 	h.Set(echo.HeaderContentType, contentType)
 	c.Response().WriteHeader(code)
 
-	_, copyErr := io.Copy(c.Response(), r)
+	copyErr := writeStreamRemainder(c, first[:n], r)
 	if copyErr == nil {
 		// Happy path — let echo / Go finalise the chunked stream normally.
 		return nil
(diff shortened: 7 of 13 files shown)


ordishs (approved):
The logic is sound, the scope is minimal, and the tests are meaningful. Verified locally: all four call sites (`GetSubtreeData`, `GetSubtree` binary mode, `GetLegacyBlock`/`GetRestLegacyBlock`, mining candidate) genuinely cannot return a legitimate empty body, and none commit headers before calling the helper. `go test -race ./services/asset/httpimpl/` passes fresh. The single-byte boundary test and the "status line must not be committed" assertion cover exactly the cases that matter for cacheability.

The `proxy_http_version 1.1` change is arguably the more important half — without it the hijack-without-terminator defense was inert in production — and the `Connection ""` comment pre-empting the keepalive question is appreciated.

Suggested follow-ups, none blocking:

1. Drop `http_500` from `proxy_next_upstream` — a deterministic empty source (e.g. a committed zero-byte subtreeData file) now means nginx re-invokes the backend up to 3 times per client request; retrying a deterministic server fault buys nothing.
2. Simplify the unreachable branch at `stream.go` — `io.ReadFull` returns a nil error only when `n == len(buf)`, and `io.ErrUnexpectedEOF` only occurs for `0 < n < len(buf)`, so with a 1-byte buffer the `n == 0` condition reduces to `errors.Is(readErr, io.EOF)`.
3. Add `defer r.Close()` in `GetLegacyBlock`/`GetRestLegacyBlock`/mining-candidate handlers — pre-existing, but the new early-return path makes it worth tidying if those repository readers hold file handles.
4. Fix the OK/200 metrics increment-before-streaming mislabel already acknowledged in the description.

icellan:
Both holes are real and the fixes are the right shape. The peek-before-`WriteHeader` is the minimal way to keep an empty source out of the cache, and spotting that nginx defaults to HTTP/1.0 for `proxy_pass` — which made the existing hijack-without-terminator defense inert — is the load-bearing finding here. The Go change is correct in isolation. What follows is about what the nginx half switches on.

Verified locally: `go build ./services/asset/...` clean, `go test -race -run 'TestStreamOrAbort|TestGetSubtreeData' ./services/asset/httpimpl/` green.

## P1

### 1. Successful `block_legacy` streams take the abort path, and `proxy_http_version 1.1` makes it bite

`services/asset/httpimpl/stream.go:91`, triggered by `services/asset/repository/GetLegacyBlock.go:186`

`GetLegacyBlockReader`'s goroutine ends its **success** path with `w.CloseWithError(io.ErrClosedPipe)`. The reader therefore returns `io.ErrClosedPipe` rather than `io.EOF` after the entire body has been copied, so `io.Copy` inside `writeStreamRemainder` returns non-nil on a perfectly good response and `streamOrAbort` hijacks and closes without the chunked terminator — on every successful legacy block.

Reproduced standalone with that exact close sequence:

```
copied=15 err=io: read/write on closed pipe
```

Pre-existing (unchanged since 2025-04, and `streamOrAbort` landed in #1209), but this PR is what makes it observable:

- Before: invisible at nginx — an HTTP/1.0 body ends at connection close, so a missing terminator means nothing.
- After: nginx sees `upstream prematurely closed connection while reading upstream` on every `block_legacy` fetch, never caches it, and for responses still sitting in `proxy_buffers` turns a 200 into a 502.

It already breaks `services/legacy/peer_server.go:2057` → `NewRawBlockMessage` → `io.ReadAll`, which surfaces the unclean end as an error, so pushing a legacy block to an SV peer fails at the end of the transfer.

Fix is one line in the producer: `w.Close()` on the success path, keeping `CloseWithError(err)` for real failures. Worth doing in this PR since this PR is what changes the proxy's behaviour.

### 2. The zero-tx path still commits a 0-byte `subtreeData`, so the new 500 is sticky

`services/asset/repository/GetSubtreeData.go:274` + `services/asset/repository/GetLegacyBlock.go:287`

`writeTransactionsViaSubtreeStoreStreaming` returns `nil` — success — when the subtree header reports `numLeaves == 0`. On the on-demand path that means `storer.Close(gCtx)` commits an empty `subtreeData` blob with a DAH, and `httpWriter.Close()` hands the handler a clean EOF, so the peek returns 500. Every later request takes the `subtreeDataExists` fast path, reads 0 bytes, and 500s again for the life of the DAH.

That is the "zero-byte file committed → permanent 500" caveat in the description, except it is reachable from the service's own code, not only from historical bad data. `storer.Abort()` when the generator produced zero bytes keeps the failure recoverable instead of persisting it.

## P2

### 3. Raw error text to the caller, and client-gone reported as a server fault

`services/asset/httpimpl/stream.go:84`

`echo.NewHTTPError(500, readErr.Error())` puts internal error text (store paths, aerospike detail, `context canceled`) into a body served to an untrusted peer. It also reclassifies the common case: if the client or proxy disconnects before the first byte, `readErr` is a context error and the peer gets a 500 that logs as a server fault — the repository layer deliberately segregates exactly this case at `GetSubtreeData.go:263` via `errors.IsContextError`. Fixed message, log the detail, and don't 500 on a context error.

### 4. The new 500s advertise themselves as publicly cacheable

`deploy/docker/base/asset-cache-nginx.conf:110`

`add_header Cache-Control "public, max-age=300" always` applies to every status code, so the 500 this PR introduces leaves labelled cacheable for 5 minutes. `proxy_cache_valid 200` keeps it out of *this* proxy's cache, which is what the PR relies on, but that guarantee is local — anything else in the path takes the header at face value. Drop `always` here, or emit `no-store` for non-200.

### 5. `errReader` loses its prefix now that the helper peeks a byte

`services/asset/httpimpl/stream_test.go:27`

`Read` does `n := copy(p, e.prefix); e.consumed = true` — with the new 1-byte peek buffer it hands over 1 byte and discards the remaining 255 (and 63). Both mid-stream-failure tests now exercise "1 byte, then truncation" rather than "a real chunk, then truncation", which is also why the first has to tolerate "a net error if the connection dropped before any body bytes hit the wire". `e.prefix = e.prefix[n:]`, failing only once drained, restores the coverage.

## Checked, no action

- `n == 0 && readErr == nil` is unreachable with a 1-byte buffer; harmless.
- No reader leak on the new early return — `GetSubtreeData`/`GetSubtree` both `defer r.Close()`, and the on-demand path releases both semaphore permits in its goroutine.
- No pipe deadlock from the peek: `io.Pipe` writes block until fully consumed, but the helper goes straight into `io.Copy`.
- No new client-side timeout from deferring the header commit — `util/http.go` sets no `ResponseHeaderTimeout`, so time-to-first-byte stays bounded by the same request context. nginx's `proxy_read_timeout 120s` now bounds it too, which is worth watching but not a defect.
- `proxy_http_version` is valid in a `location` context; only the cached location is affected.
- `deploy/docker/base/asset-cache-nginx.conf` is the only nginx config in the repo, so nothing is left half-fixed here — deployment configs living outside the repo still need the same directive.

ctnguyen (changes requested):
Review at commit [0fc9f36](https://github.com/oskarszoon/teranode/commit/0fc9f3609c27f6c54f9f73de4c7dc49dda163683)

Solid, tightly-scoped fix, and deferring the status line until a byte has actually been read is the right shape for it — the peeked byte is written rather than swallowed, with a boundary test to prove it. Spotting that `proxy_pass` defaults to HTTP/1.0, which made the existing truncation defence inert in production, is the sharper half of the PR; my findings below are mostly about the blast radius of the new 500, and about what can still create an empty source in the first place.

## Blocking issues

### ChiR1 — The new failure response goes out labelled `Cache-Control: public, max-age=300`

**Problem:** `deploy/docker/base/asset-cache-nginx.conf:110` is `add_header Cache-Control "public, max-age=300" always;`. The `always` flag attaches that header to *every* status the cached location returns, so the new "empty response body" 500 is emitted with an explicit, positive freshness lifetime of exactly the same 300s window as the poisoned entry this PR removes. The asset service sets no `Cache-Control` header of its own on these routes, so nginx's value is the only one on the wire and nothing contradicts it. What is inherited: the `always` header pre-dates this PR, and so do the 404/503 responses it already mislabels. What is new: this PR makes a 5xx an expected, routine outcome of the cached endpoints, and it rests its entire correctness argument on "a failure is never stored". As written, that argument is not a property of the response — it is a property of `proxy_cache_valid 200 5m` on one hop. RFC 9111 §3 permits a shared cache to store a 500 precisely when explicit freshness information is present, so any second cache in the path (an operator CDN or ingress in front of the asset-cache container, a second asset-cache layer, a proxy between peers) is licensed to store and replay the 500 for five minutes. Issue 1368's suggested fix 0 asked specifically for `no-store` on the failure path; that half is missing.

**Why it matters:** The PR leaves the same bug class reachable one layer out with no code change at all — just a deployment topology nobody re-audited. Fixing it while the failure path is being introduced is far cheaper than diagnosing it again from a second stalled node.

**Fix:** Make the freshness label conditional on the upstream status, at `http {}` level:

```nginx
map $upstream_status $asset_cache_control {
    default "no-store";                  # 4xx/5xx: never store, never replay
    200     "public, max-age=300";
    ""      "public, max-age=300";       # cache HIT — no upstream was contacted
}
```

then in the cached location replace the unconditional header with:

```nginx
add_header Cache-Control $asset_cache_control always;
```

The `""` arm is required or every HIT would be labelled `no-store`. A comma-separated `$upstream_status` (possible if `proxy_next_upstream` retried) falls to `default`, which is the conservative direction. Worth confirming with `curl -i` against a 404 and a 500 after the change.

### ChiR2 — The on-demand path can still commit a zero-byte `subtreeData`, and the new 500 then sticks for the whole DAH window

**Problem:** `writeTransactionsViaSubtreeStoreStreaming` returns `nil` after writing nothing in at least two cases. `services/asset/repository/GetLegacyBlock.go:283-289` derives `totalTxs` from `header.numLeaves` and short-circuits with `return nil` when it is zero; and `writeChunkToWriter` is called with `block == nil` from this path (`services/asset/repository/GetSubtreeData.go:253`), so a coinbase-placeholder node emits no bytes at all (`services/asset/repository/GetLegacyBlock.go:504-516`). On a `nil` return, `dualStreamWithFileCreation` calls `storer.Close(gCtx)` (`services/asset/repository/GetSubtreeData.go:274`), and `FileStorer.Close` (`services/utxopersister/filestorer/FileStorer.go:156`) finalises the blob unconditionally — there is no zero-length guard. A zero-byte `subtreeData` is therefore persisted, and because `Exists` then returns true for that hash, every later request takes the stored-file branch and never retries generation: it just returns the new 500.

How long that lasts depends on which route created the blob. On-demand files are created with `options.WithDeleteAt(dah)` where `dah = current height + GetSubtreeValidationBlockHeightRetention()` (`services/asset/repository/GetSubtreeData.go:203`), which resolves to 288 blocks by default (`settings/interface.go:93`, `settings.conf:610`), and none of the `subtreestore` URLs in `settings.conf` set `disableDAH`, so the blob is genuinely scheduled for deletion — roughly two days at a 10-minute block target, not forever. Only block-persister-promoted files are DAH=0 and permanent, per the comment at `services/asset/repository/GetSubtreeData.go:165-166`; a zero-byte blob already in the store from that route is the one that needs manual cleanup.

The premise that these endpoints cannot legitimately return an empty body does hold on the assembly side: `getIncompleteSubtreeDataChan` rejects `currentSt.Length() <= 1` (`services/blockassembly/subtreeprocessor/SubtreeProcessor.go:735`) and empty blocks get `generateEmptyBlockCandidate` with `SubtreeHashes: [][]byte{}` (`services/blockassembly/BlockAssembler.go:1625`), so a placeholder-only subtree never reaches a block. The reachable route is a corrupt or truncated subtree file whose stream header reports `numLeaves == 0`.

**Why it matters:** This is the other half of the same defect, on the write side. Before the change a peer in this state served an empty 200; after it serves a hard 500 for that hash until DAH pruning catches up — on the order of days, with no regeneration attempt in between, for exactly the `subtree_data` a stalled catchup is blocked on. The description accepts that as collateral, but the code that manufactures the condition sits in the same subsystem as the fix and the guard is a handful of lines.

**Fix:** Refuse to finalise an empty blob — a counting writer in `dualStreamWithFileCreation` is enough:

```go
counted := &countingWriter{w: multiWriter}

err := repo.writeTransactionsViaSubtreeStoreStreaming(gCtx, counted, nil, subtreeHash)
if err == nil && counted.n == 0 {
    // No endpoint can serve an empty subtreeData, so never commit one: that would
    // turn a transient generation fault into a 500 for this hash until DAH pruning.
    err = errors.NewProcessingError("[GetSubtreeDataReader] generated subtreeData for %s is empty", subtreeHash.String())
}

if err != nil {
    // Abort discards the temp file — services/utxopersister/filestorer/FileStorer.go:196
    storer.Abort(err)
    _ = httpWriter.CloseWithError(err)
    …
}
```

Nothing is committed, so the handler's 500 stays retryable. Worth a repository-level test that a zero-leaf subtree leaves no `subtreeData` blob behind.

### ChiR3 — A client that disconnects before the first byte is now an ERROR-logged 500

**Problem:** `services/asset/httpimpl/stream.go:78-85` treats every non-EOF read error at offset 0 as a server fault, and on the on-demand `subtree_data` path that includes client disconnects. The sequence: the request context is cancelled, `writeChunkToWriter` returns `ctx.Err()` (`services/asset/repository/GetLegacyBlock.go:500`), `dualStreamWithFileCreation` calls `httpWriter.CloseWithError(err)` (`services/asset/repository/GetSubtreeData.go:256`), and the blocked `io.ReadFull` returns `(0, context.Canceled)`. That is not EOF, so `streamOrAbort` returns `echo.NewHTTPError(500, "context canceled")`, `customHTTPErrorHandler` logs it via `logger.Errorf` (`services/asset/httpimpl/http.go:705`), then attempts `c.JSON(500, …)` on a connection whose peer is already gone — which fails and logs a second `Errorf` at `services/asset/httpimpl/http.go:717`. This is newly introduced, not inherited: before the peek, the same disconnect committed the 200, took the hijack path and returned `nil`, producing zero error logs. The repository code immediately below does the opposite deliberately — `services/asset/repository/GetSubtreeData.go:257-267` classifies exactly this condition as "client gone", logs it at **Debug** and segregates the metric, with a comment noting that "one cancelled batch can cascade tens of in-flight streams" and that the operational signal must stay clean. The window is not a sliver either: on the on-demand path the first byte only appears after a whole chunk of tx metadata has been fetched, so "cancelled before byte 0" covers the slow part of the request.

**Why it matters:** Catchup cancels in batches — that cascade is documented in issue 1368 — so this manufactures serving-side ERROR noise, two lines per cancelled stream, on exactly the nodes an operator is trying to diagnose, and in a PR whose sibling work is explicitly about a log storm. It also records a client-caused abort as a server fault in the one place a reader would look to tell them apart.

**Fix:** Classify the read error before deciding it is a server fault, and route the client-gone case through the same hijack-and-close the mid-stream path already uses — that keeps the response uncommitted and uncacheable without logging an error:

```go
n, readErr := io.ReadFull(r, first[:])
if n == 0 {
    // Client (or proxy) went away before the source produced anything. Not a
    // server fault: mirror the debug-level "client gone" segregation in
    // dualStreamWithFileCreation rather than emitting an ERROR-logged 500 per
    // cancelled in-flight stream.
    if errors.IsContextError(readErr) || errors.Is(readErr, io.ErrClosedPipe) {
        if conn, _, hjErr := c.Response().Hijack(); hjErr == nil {
            _ = conn.Close()
            return nil
        }
        // h2: no hijack available, fall through to the 500.
    }

    if errors.Is(readErr, io.EOF) {
        return echo.NewHTTPError(http.StatusInternalServerError, "empty response body")
    }

    return echo.NewHTTPError(http.StatusInternalServerError, readErr.Error())
}
```

`errors.IsContextError` already exists at `errors/error_utils.go:217` and handles both bare and teranode-wrapped context errors. Returning plain `nil` here is *not* an alternative — with the response uncommitted, echo/net-http would finalise it as a 200 with `Content-Length: 0`, which is the bug being fixed.

## Non-blocking issues

### ChiR4 — The peek truncates `errReader`'s prefix, so the two mid-stream tests now exercise a 1-byte body

**Problem:** `errReader.Read` (`services/asset/httpimpl/stream_test.go:27-35`) returns `copy(p, e.prefix)` and unconditionally sets `consumed = true`. `io.ReadFull` now calls it first with a **1-byte** buffer, so it copies one byte, returns `n == 1`, and discards the remaining 255 (resp. 63) bytes of prefix; the next `Read`, from `io.Copy`, returns the error immediately. `TestStreamOrAbort_MidStreamFailure_ClientSeesUnexpectedEOF` and `TestStreamOrAbort_FailureConnIsActuallyClosed` therefore no longer exercise "256 bytes then failure" — they exercise "1 byte then failure". They still pass, because `(*http.response).Hijack` flushes the chunk writer when the header has been written, so the status line, headers and that single 1-byte chunk reach the wire before the close; that is why the change leaves no visible trace.

**Why it matters:** The description states that "the four pre-existing `streamOrAbort` tests are unmodified and still pass", which is true — but two of them now cover a materially weaker scenario. A one-byte body is a degenerate chunked stream, so the multi-chunk truncation case those tests were written for is no longer covered by anything; a future partial-write bug in `writeStreamRemainder` would slip through.

**Fix:** Make `errReader` serve its prefix across however many reads it takes, instead of dropping the tail on the first short read:

```go
func (e *errReader) Read(p []byte) (int, error) {
	if len(e.prefix) > 0 {
		n := copy(p, e.prefix)
		e.prefix = e.prefix[n:]
		return n, nil
	}

	return 0, e.err
}
```

The `consumed` field then becomes unused and can go.

### ChiR5 — `proxy_next_upstream … http_500` retries a deterministic server fault

*Same as [@ordishs's point 1 (29-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1377#pullrequestreview-4807657736).*

**Problem:** `deploy/docker/base/asset-cache-nginx.conf:51` lists `http_500`, with `proxy_next_upstream_tries 3` on `deploy/docker/base/asset-cache-nginx.conf:53`, so a deterministic empty source can cost up to three backend invocations per client request. How real the amplification is depends on DNS: because `proxy_pass http://$asset_backend` uses a variable and there is no `upstream {}` block, nginx builds the peer list per request from the resolver answer. With a single A record — the default single-container compose stack — there is no alternative peer and the retry cannot fire. If `asset` is scaled and Docker DNS returns multiple A records, it can.

**Why it matters:** Bounded, and zero in the default topology, but wrong in principle: retrying a deterministic content fault buys nothing, and on a scaled deployment it triples the cost of every poisoned hash while the fleet is already under catchup load.

**Fix:** Drop `http_500`, scoped to the cached location so the change stays contained:

```nginx
# A 500 from asset is a content fault (empty or failed generation), not a
# transient backend outage: re-invoking the backend cannot change the answer.
proxy_next_upstream error timeout http_502 http_503 http_504;
```

The `proxy_cache_use_stale … http_500` interaction on `deploy/docker/base/asset-cache-nginx.conf:30` is already covered in the description's "What this does not fix", so I am not raising it separately.

### ChiR6 — Two of the three arms in the `n == 0` branch are unreachable

*Same as [@ordishs's point 2 (29-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1377#pullrequestreview-4807657736).*

**Problem:** `services/asset/httpimpl/stream.go:80`. With a 1-byte buffer, `io.ReadFull` delegates to `ReadAtLeast(r, buf, 1)`, which clears the error only when `n >= min`; so inside `n == 0`, `readErr` is always non-nil, and `io.ErrUnexpectedEOF` (which requires `0 < n < min`) cannot occur. Both `readErr == nil` and `errors.Is(readErr, io.ErrUnexpectedEOF)` are dead.

**Why it matters:** Clarity only — but dead defensive arms in a helper whose whole job is a subtle wire-level invariant invite the next reader to assume the buffer might be larger than one byte, and hide the fact that the condition is simply "EOF".

**Fix:** `if errors.Is(readErr, io.EOF) { … }`, as folded into the ChiR3 snippet. If the intent is to leave room for a future multi-byte peek, say so in a comment rather than encoding it as unreachable branches.

### ChiR7 — The `subtree` hex route still commits 200 before the first byte

**Problem:** The doc comment and description say the fix covers `subtree`, but only the binary route does — `services/asset/httpimpl/GetSubtree.go:190-193` (`mode == HEX`) still calls `WriteHeader(http.StatusOK)` and then copies into a hex encoder, so a zero-byte `GetSubtreeNodeHashesReader` yields a committed 200 with an empty body. Exposure is small: the cached location's regex is anchored with `$`, so `/api/v1/subtree/<hash>/hex` does not match it and this nginx never stores the response. The other endpoint families the regex does match (`header`, `headers`, `blocks`, `block`, `headers_{to,from}_common_ancestor`) use `c.Blob(200, …)` on a fully materialised slice, so they always carry a `Content-Length` and neither the truncation defence nor the HTTP/1.1 change is relevant to them.

**Why it matters:** Only that the helper's contract is now half-true for `subtree`, so someone auditing "which endpoints can emit a cacheable empty 200" gets the wrong answer from the comment.

**Fix:** Either route the hex path through the same peek (wrapping the encoder, since `streamOrAbort` writes raw bytes), or narrow the doc comment at `services/asset/httpimpl/stream.go:67-75` to say `subtree` **binary** mode and note that the hex route is outside the cached location.

### ChiR8 — Three of the five call sites never close their reader

*Partially the same as [@ordishs's point 3 (29-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1377#pullrequestreview-4807657736) — extending it with which path actually leaks.*

**Problem:** `services/asset/httpimpl/GetSubtreeData.go:79` and `services/asset/httpimpl/GetSubtree.go:183` both `defer r.Close()`; `services/asset/httpimpl/GetLegacyBlock.go:89`, `services/asset/httpimpl/GetLegacyBlock.go:145` and `services/asset/httpimpl/mining_candidate_legacy_block.go:66` do not. On the *new* early-return path this is harmless, and the reason matters: all three of those readers are `*io.PipeReader`, and `io.ReadFull` on a pipe can only return `n == 0` if the write end is already closed — `GetLegacyBlockReader` and `GetMiningCandidateLegacyBlockReader` both `CloseWithError`/`Close` on every exit — so the writer goroutine has already finished. The leak is on the *pre-existing* hijack path: if `io.Copy` fails because the write to the response failed (client gone) rather than because the source failed, the pipe writer is still mid-`Write`, `streamOrAbort` hijacks and returns `nil`, and with no `r.Close()` that goroutine blocks on `w.Write` forever, holding its arena and any file-store read permit.

**Why it matters:** A real goroutine-plus-buffer leak under exactly the client-cancellation load pattern in issue 1368, and it is one line per handler.

**Fix:** `defer r.Close()` immediately after each of the three successful `GetLegacyBlockReader` / `GetMiningCandidateLegacyBlockReader` calls, matching `services/asset/httpimpl/GetSubtreeData.go:79`. Closing the read end unblocks the writer with `io.ErrClosedPipe`, which those goroutines already handle.

### ChiR9 — The `OK`/`200` counter is incremented before streaming, so the new 500 is counted as a success

*Same as [@ordishs's point 4 (29-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1377#pullrequestreview-4807657736), and acknowledged in the description.*

**Problem:** `services/asset/httpimpl/GetSubtreeData.go:81`, `services/asset/httpimpl/GetSubtree.go:127`, `services/asset/httpimpl/GetLegacyBlock.go:87` and `services/asset/httpimpl/GetLegacyBlock.go:143` all increment `WithLabelValues("OK", "200")` before calling `streamOrAbort`, so every empty-source 500 lands in the OK bucket. Agreed that this is a pre-existing mislabel and a fair follow-up rather than a blocker — but the follow-up now costs more than it did, because the empty-source 500 is a brand-new failure mode with no metric of its own. There is no dashboard signal for "this node is serving 500s for subtree_data", which is the first thing an operator would look for when validating this fix in production.

**Why it matters:** The PR's own outstanding verification step — no cached empty entry after an aborted stream — is far easier to confirm from a counter than from nginx logs, and right now the counter cannot show it.

**Fix:** A separate PR is fine, but consider landing the narrow half here: move the four `OK`/`200` increments to after `streamOrAbort` returns `nil`, and add an `ERROR`/`500` increment on the error return. The mid-stream hijack path also returns `nil`, so that alone does not fix every case, which is why the full change belongs on its own.

## Recap

| --- | --- | --- | --- |
| ChiR1 | 500 labelled publicly cacheable | YES | 70% |
| ChiR2 | Zero-byte subtreeData still committed | YES | 58% |
| ChiR3 | Client abort becomes ERROR 500 | YES | 52% |
| ChiR4 | Peek truncates test reader prefix | | 40% |
| ChiR5 | `next_upstream` retries deterministic 500 | | 22% |
| ChiR6 | Unreachable read-error arms | | 18% |
| ChiR7 | Hex subtree route uncovered | | 14% |
| ChiR8 | Legacy-block readers never closed | | 12% |
| ChiR9 | OK/200 counted before streaming | | 10% |

oskarszoon:
All addressed, `0fc9f36..9d4fbe10`, one commit per finding, plus `upstream/main` merged in (`e71e2efd3`) so this sits on current main. @ctnguyen ChiR1–9 and @icellan items 1–5 overlap on four of them, so mapping both below.

## Bot review comment on the `Cache-Control` map — accepted tradeoff, documented

`9d4fbe10f`. Correct, and thanks for working it through: with `proxy_cache_use_stale … http_500`, when nginx serves a stale-but-good 200 because the upstream attempt failed, `$upstream_status` is that failed attempt's 5xx, so the map labels a valid body `no-store` and downstream caches will not store it.

Keeping the behaviour as-is. It lasts only as long as the upstream-failure window and it fails in the safe direction — a downstream cache re-fetches rather than serving something it shouldn't. Distinguishing the case needs a composite `"$upstream_cache_status|$upstream_status"` key, and that is more ways to mislabel a response than the marginal cacheability during a failure window is worth. Written into the map's comment so the next person doesn't have to re-derive it.

## @icellan item 1 — successful `block_legacy` takes the abort path (P1, new)

`f0622578f`. Confirmed and fixed, and the diagnosis was exactly right: this PR is what turns a dormant bug into a live one. Two details worth recording beyond the report:

- It was **both** success paths, not one — the `len(block.Subtrees) == 0` branch after the coinbase write as well as the normal path after all subtrees stream. Fixing only `:186` would have left the empty-subtree case broken.
- `GetMiningCandidateLegacyBlockReader` already did `w.Close()` on success. The correct pattern was sitting in a sibling producer feeding the same helper.

**Six existing assertions had codified the bug** — three `require.ErrorIs(err, io.ErrClosedPipe)` after a complete stream, one `assert.Equal(io.ErrClosedPipe, err)` at end of stream, and one in the backpressure test reading `require.ErrorIs(t, err, io.ErrClosedPipe, "stream did not complete cleanly (truncated/aborted)")` — a message asserting the opposite of what the check enforced. They now assert a clean end. The handler-level tests never caught any of it because their mock producer closes with `writer.Close()`, i.e. correctly.

New tests: `TestGetLegacyBlockReader_SuccessEndsWithCleanEOF` (both success paths, RED with `io: read/write on closed pipe`, GREEN with a clean `io.EOF` and a complete body) and `TestStreamOrAbort_CompleteStreamDependsOnHowTheProducerCloses` pinning the coupling at the helper boundary.

**Analysis you asked for, deliberately not changed here.** The failure paths do mask the real error, and worse than it looks: `pipe.readCloseError()` returns `werr` only when `rerr == nil`, so the paired `r.CloseWithError(err)` forces the literal `io.ErrClosedPipe` regardless — passing the real error to `w.CloseWithError` is a no-op while that call stands. The real error is also logged nowhere, since the `errgroup` is never `Wait()`ed and the return value is discarded. Not a misclassification risk under the new peek: a pre-first-byte failure on these producers is only reachable via a failed pipe `Write`, which means the consumer genuinely went away. Wants its own commit and its own analysis; happy to file it.

## Blocking items

**ChiR1 / @icellan 4 — 500 labelled publicly cacheable** (`c9a7fe83f`). `map $upstream_status $asset_cache_control` at `http {}`, `add_header Cache-Control $asset_cache_control always`. The `""` arm is load-bearing exactly as described; without it every HIT is labelled `no-store`. Curl matrix: 200 MISS and 200 HIT → `public, max-age=300`; 500 and 404 → `no-store`. The observation that the guarantee was a property of one hop's `proxy_cache_valid` rather than of the response is what I should have caught writing the description — suggested fix 0 asked for `no-store` and I read it as applying only to the asset service.

**ChiR2 / @icellan 2 — zero-byte `subtreeData` still committed** (`2b211addd`). Counting writer; an empty generation now aborts instead of committing, so nothing persists and the 500 stays retryable. Right that this is the write-side half of the same defect — I traced `totalTxs == 0 → return nil` early on, dropped it when the observed subtree turned out to have 1024 leaves, and never came back. Repository test asserts a zero-leaf subtree leaves no blob behind.

**ChiR3 / @icellan 3 — client-gone reported as a server fault** (`9f268f4d0`). Classified via `errors.IsContextError` / `io.ErrClosedPipe` and routed through hijack-and-close: uncommitted, uncacheable, no ERROR line. Fixed message to the caller instead of raw `readErr.Error()`, detail logged server-side. A genuine self-inflicted regression — the peek converted a silent correct outcome into two ERROR lines per cancelled stream, in a change whose sibling PR exists to kill a log storm. Collapses ChiR6 / @icellan's "checked, no action" note: with a 1-byte buffer the `n == 0` condition really is just `errors.Is(readErr, io.EOF)`.

## Non-blocking

`350588eb2` ChiR4 / @icellan 5 — `errReader` serves its prefix across reads, so both mid-stream tests are back to multi-chunk truncation; `consumed` deleted. `847a10a20` ChiR5 — `http_500` dropped from `proxy_next_upstream`. `47800f9fe` ChiR7 — comment narrowed to `subtree` **binary**, noting the hex route sits outside the cached location. `40e81a06d` ChiR8 — `defer r.Close()` on all three legacy handlers; the diagnosis that the leak is on the pre-existing hijack path rather than the new early return is what made it worth doing now. `8b698c0d3` ChiR9 — narrow half only.

Three limits I'd rather state than bury:

- ChiR9 needed distributing across `GetSubtree`'s return paths, not moving one line, or JSON and HEX would have stopped being counted. The mid-stream hijack path still returns `nil` and still counts OK — commented; full fix stays a follow-up.
- ChiR4's guard is server-side deliberately: a client-side byte-count assertion flaked 30 out of 30 runs.
- ChiR5's multi-A-record retry, ChiR1's revalidated-304 arm, and the end-to-end nginx/SV-peer effects of the item-1 fix are reasoned from the clean EOF, not observed — no scaled deployment or chain data here. The live cache-behaviour check remains outstanding, unchanged from the description.

`go test -race ./services/asset/...` green across 6 packages (816 tests), `go build ./services/legacy/...` clean, `make lint` 0 issues, `nginx -t` successful.

ctnguyen (approved):
Review at commit [9d4fbe1](https://github.com/oskarszoon/teranode/commit/9d4fbe10f7539def4bd79fa8809e6bbfb3f093ea)

Second pass, and this one reads well: all nine points from the last round came back as actual code rather than comments standing in for fixes, one commit each, with a test pinned to every behavioural claim. The pick of the round isn't even one of mine — chasing @icellan's item 1 turned up six existing assertions that had codified `CloseWithError(io.ErrClosedPipe)`-as-success, one of them with a failure message asserting the opposite of what it enforced, and fixing that *before* flipping nginx to HTTP/1.1 is the right order to do it in. The counting-writer guard, the client-gone hijack, and the per-path metric increments are all the shape I was hoping for, and the nginx comments now explain the tradeoffs instead of leaving the next reader to re-derive them. Nothing below blocks the merge; the two worth real time are the zero-byte blob that may already be sitting on disk somewhere, and how "client gone" is being decided.

## Status of previous points

Two caveats that belong outside the table. ChiR1's fix is in and works for the failure this PR introduces, but it labels off the *upstream's* status rather than the one nginx sends, which leaves nginx's own generated statuses on the cacheable arm — that residue is ChiR12 rather than a partial fix of ChiR1. ChiR3's fix landed exactly as prescribed, but the response note also claims a fixed message replaced the raw `readErr.Error()`; the diff only does that for the `io.EOF` arm, so the raw-text half of @icellan's item 3 is still open and is ChiR13. ChiR9 landed the narrow half, which is what was agreed.

| --- | --- | --- | --- |
| ChiR1 | 500 labelled publicly cacheable | YES | FIXED |
| ChiR2 | Zero-byte subtreeData committed | YES | FIXED |
| ChiR3 | Client abort became ERROR 500 | YES | FIXED |
| ChiR4 | Peek truncated test prefix | | FIXED |
| ChiR5 | `next_upstream` retried 500 | | FIXED |
| ChiR6 | Unreachable read-error arms | | FIXED |
| ChiR7 | Hex route scope in doc | | FIXED |
| ChiR8 | Legacy readers never closed | | FIXED |
| ChiR9 | OK/200 counted before streaming | | FIXED |

## Non-blocking issues

### ChiR10 — A zero-byte `subtreeData` already in the store is still a permanent 500, with no self-heal

**Problem:** The counting writer in `dualStreamWithFileCreation` (`services/asset/repository/GetSubtreeData.go`) stops this PR's code from *creating* a zero-byte blob, and the new repository test pins that. It does nothing for a zero-byte blob that is already there — which is the population that matters, because every peer still running the old code can commit one, and block-persister-promoted files carry DAH=0, i.e. no expiry at all. For such a hash `Exists` returns true, the handler takes the stored-file branch, the peek reads `io.EOF`, and the endpoint answers 500 for that hash indefinitely with no attempt to regenerate. The description accepts this, and it is strictly better than the empty 200 it replaces — but the branch that chooses the stored file sits a few lines from the guard that was just added, and the recovery is mechanical: a zero-length blob carries no information, so it should be treated as absent rather than as authoritative.

**Why it matters:** After this PR that is the only remaining route to a permanent 500, and it lands on exactly the `subtree_data` a stalled catchup is blocked on. Left as-is it becomes a runbook step ("flush `/tmp/nginx/cache`, then go hunt zero-byte blobs") on nodes that are already hard to diagnose, and the honest reading of issue 1368 is that nobody knows yet whether such blobs exist in the field.

**Fix:** Treat a zero-length stored blob as a miss in `GetSubtreeDataReader`, before committing to the stored-file branch:

```go
// A zero-byte subtreeData cannot be legitimate (see streamOrAbort) and carries no
// information. Left in place it pins this hash to a 500 for the life of the blob —
// permanently for block-persister-promoted files, which have no DAH. Discard and
// regenerate instead, so the failure stays transient.
if subtreeDataExists {
	if size, err := repo.SubtreeStore.Size(ctx, subtreeHash[:], fileformat.FileTypeSubtreeData); err == nil && size == 0 {
		repo.logger.Warnf("[GetSubtreeDataReader] discarding zero-byte subtreeData for %s and regenerating", subtreeHash.String())

		_ = repo.SubtreeStore.Del(ctx, subtreeHash[:], fileformat.FileTypeSubtreeData)
		subtreeDataExists = false
	}
}
```

If the store has no cheap size call, the same thing works by peeking one byte off the opened reader and falling through to generation on `io.EOF`.

### ChiR11 — "Client gone" is inferred from the error value, so a server-side fault takes the same silent path

**Problem:** The `n == 0` branch in `services/asset/httpimpl/stream.go` decides client-gone from `errors.IsContextError(readErr) || errors.Is(readErr, io.ErrClosedPipe)` and then returns `nil`: no status line, no log line, and the caller counts the request `OK`/`200`. Two server-side faults get that treatment.

`io.ErrClosedPipe` is not a client signal on these producers — it is what a *failure* surfaces as. The failure paths close the read end with the error too, and `io.Pipe`'s `readCloseError` returns the write-side error only while the read side has none, so the reader sees the literal `io.ErrClosedPipe` no matter what was handed to `w.CloseWithError`. That masking is documented in the response note itself; the conclusion drawn there is that a pre-first-byte failure on these producers is only reachable through a failed pipe `Write`, which may well hold today, but it is a property of code in another package with nothing asserting it.

`context.DeadlineExceeded` is enshrined as a disconnect by `TestStreamOrAbort_ClientGoneBeforeFirstByte`, but a deadline anywhere in the server's own context chain — a store or gRPC per-operation timeout — produces the same error with the client still connected and waiting. That peer gets a bare connection close, nginx turns it into a 502, and the asset node logs nothing at all.

**Why it matters:** These cases were silent before the PR too, so this is not a regression — but the change installs a comment and a test asserting they are client faults, which is what an operator will trust while chasing the very 502s `proxy_http_version 1.1` makes visible. A swallowed server fault counted as `OK` is a worse signal than the ERROR pair ChiR3 removed.

**Fix:** Gate on the request context, which is the only authoritative signal, and keep the error value for the log line:

```go
	// net/http cancels the request context when the peer goes away — that is the
	// authoritative signal. An error value cannot separate a disconnect from a deadline
	// in our own context chain, and io.ErrClosedPipe is what a producer-side failure
	// surfaces as once the read end has been closed with an error.
	if c.Request().Context().Err() != nil {
		if conn, _, hjErr := c.Response().Hijack(); hjErr == nil {
			_ = conn.Close()
			return nil
		}
		// h2: no hijack available, fall through to the 500.
	}
```

Then flip the table test to match: cancel the request context for the disconnect rows, and add a row where the read fails with `context.DeadlineExceeded` while the request context is still live, asserting a 500. A debug line on the swallowed case would also help, though that needs a logger threaded into the helper — worth pairing with ChiR13.

### ChiR12 — The empty arm of the freshness map labels nginx's own failures publicly cacheable

*Same as [@icellan's item 4 (29-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1377#issuecomment-5117602035).*

**Problem:** Their prescription was "emit `no-store` for non-200"; this extends it to the non-200s the implemented map does not reach. `map $upstream_status $asset_cache_control` (`deploy/docker/base/asset-cache-nginx.conf`, ~line 63) keys off what the backend returned, and any status nginx produces without ever obtaining an upstream status leaves that variable unset — so it matches the `""` arm, the one intended for cache hits, and goes out `public, max-age=300`. That covers request-level rejections on these routes (400, 405, 408, 411, 413, 414) and any nginx-internal 500. The map's own comment says "4xx/5xx: never store, never replay", and for this class it does the opposite.

Scope, so this is not read as bigger than it is: the failure this PR introduces is *not* affected — an asset 500 arrives as `500` and is labelled correctly, which is the part that mattered in ChiR1. Worth confirming one adjacent case against the curl matrix you already have, though: an aborted stream where nginx never received an upstream response header should render `$upstream_status` as `-` and fall to `default`, but if it renders empty it takes the HIT arm.

**Why it matters:** Small in impact — a downstream cache holding a 400 or a 414 for five minutes is not the stall from issue 1368. It is worth closing because the block now carries a comment promising behaviour the map does not implement, and the next person to touch this will trust the comment.

**Fix:** Key on the status actually being sent, which is the only single-variable way to distinguish "no upstream was contacted because we hit cache" from "no upstream was contacted because the request never got that far":

```nginx
# $status is the status nginx sends: 200 on a cache HIT, and the real code for anything
# nginx generates itself.
map $status $asset_cache_control {
    default "no-store";              # 4xx/5xx: never store, never replay
    200     "public, max-age=300";
    304     "public, max-age=300";   # revalidation must not tell a cache to drop a valid entry
}
```

Side effect worth knowing about rather than a reason to do it: keyed this way, the stale-but-good 200 and the revalidated 304 recorded in the comment as accepted tradeoffs stop being mislabelled too, since `$status` is 200 in both. That behaviour is already documented as deliberate and safe-direction, so treat retiring it as your call, not a finding. If you would rather keep the current keying and the tradeoffs as they are, the residue above is narrow enough to accept — say so in the comment so it does not read as a promise.

### ChiR13 — The last arm still puts raw internal error text on the wire

*Same as [@icellan's item 3 (29-07-2026)](https://github.com/bsv-blockchain/teranode/pull/1377#issuecomment-5117602035).*

**Problem:** Their item had two halves; the client-gone half is fixed, and this is the other one, which the response note reports as done. The final return of the `n == 0` branch in `services/asset/httpimpl/stream.go` is still `echo.NewHTTPError(http.StatusInternalServerError, readErr.Error())`, so whatever the store or the producer produced — a filesystem path, aerospike or gRPC detail, `context canceled` on the h2 fall-through — is serialised into a body served to an untrusted peer, and nothing is logged with the subtree hash. Only the `io.EOF` arm got the fixed string. Flagging it mainly so item 3 doesn't get closed on the strength of the note; the rest of the package already answers this way (`GetSubtree.go` returns `err.Error()` for both its 404 and its 500), which is why it is not blocking.

**Why it matters:** These routes are reachable by any peer, this is the one place internal topology leaks on the new path, and the operator-side detail — which hash, which store — is exactly what the new failure mode needs and currently drops.

**Fix:** Fixed text outbound, detail server-side:

```go
	// Detail belongs in the log, not in a body served to an untrusted peer.
	h.logger.Errorf("[streamOrAbort] source failed before first byte for %s: %v", c.Request().URL.Path, readErr)

	return echo.NewHTTPError(http.StatusInternalServerError, "failed to read response body")
```

The helper has no logger today, so this means either passing one in or returning a typed sentinel the handler logs — the same plumbing ChiR11 wants, so they land well together.

### ChiR14 — Nothing asserts the clean-close property for the other readers feeding `streamOrAbort`

**Problem:** This is a coverage gap, not a diagnosed defect — those readers are not in this diff and I am not claiming any of them is broken. What the diff does establish is that the helper's happy-versus-abort decision is *entirely* how the producer closed its pipe (`TestStreamOrAbort_CompleteStreamDependsOnHowTheProducerCloses` pins exactly that), and that the property was silently violated in one producer for over a year while six assertions codified the violation. The close fix lands in `services/asset/repository/GetLegacyBlock.go` only, and the new `TestGetLegacyBlockReader_SuccessEndsWithCleanEOF` asserts it only for that producer, so `GetSubtreeNodeHashesReader` and the two `GetSubtreeDataReader` branches have no test standing behind the invariant they now have to satisfy.

**Why it matters:** Cheap insurance on a property this PR has just made load-bearing: before it, an unclean close was invisible at nginx; after it, an unclean close means the response is never cached. That is worth an assertion rather than an assumption, particularly since the first pass over the one producer anyone did examine missed one of its two success paths.

**Fix:** Generalise the helper the new test already introduces, over every reader that can reach `streamOrAbort`:

```go
for _, tc := range []struct {
	name string
	open func(t *testing.T, ctx *testCtx) io.ReadCloser
}{
	{"subtree node hashes", func(t *testing.T, ctx *testCtx) io.ReadCloser { /* GetSubtreeNodeHashesReader */ }},
	{"subtree_data, stored file", ...},
	{"subtree_data, generated on demand", ...},
	{"mining candidate legacy block", ...},
} {
	t.Run(tc.name, func(t *testing.T) {
		r := tc.open(t, setup(t))
		defer func() { _ = r.Close() }()

		// A complete stream must end in io.EOF: streamOrAbort's happy/failure decision is
		// exactly how the producer closed, and an abort drops the chunked terminator.
		drainAndAssertCleanEOF(t, r)
	})
}
```

### ChiR15 — The status label now carries two vocabularies on the same metric

**Problem:** The new failure increments use `http.StatusText(...)`, so `prometheusAssetHTTPGetSubtree` gains label values `Not Found`, `Bad Request` and `Internal Server Error`, while every success path keeps the numeric `"200"` (`GetSubtree.go`, `GetSubtreeData.go`, `GetLegacyBlock.go`). No single PromQL grouping now covers both forms, and a text status neither sorts nor joins with a code. Two smaller inaccuracies ride along: `streamOrAbort`'s h2 fall-through returns the raw copy error, which is not a 500, yet gets labelled one; and the HEX route counts `Internal Server Error` for a copy failure on a response already committed as 200. The JSON path also counts `OK`/`200` just before `c.JSONPretty`, which can still fail.

**Why it matters:** Dashboard-level only, but the whole point of moving these increments was to give an operator a usable "this node is serving 500s for `subtree_data`" signal, and a mixed vocabulary is the one thing that makes such a query awkward to write.

**Fix:** Keep the label numeric to match the existing `"200"`:

```go
prometheusAssetHTTPGetSubtree.WithLabelValues("ERROR", strconv.Itoa(http.StatusNotFound)).Inc()
```

Worth a quick look at what the asset dashboards already query before settling on the form.

## Recap

| --- | --- | --- | --- |
| ChiR10 | Zero-byte blob has no self-heal | | 40% |
| ChiR11 | Client-gone inferred from error | | 34% |
| ChiR12 | Empty-upstream arm labels failures cacheable | | 26% |
| ChiR13 | Raw error text to caller | | 22% |
| ChiR14 | Clean-close property untested elsewhere | | 18% |
| ChiR15 | Mixed status label vocabularies | | 12% |

icellan:
Re-reviewed at `9d4fbe10f` (13 files, +650/−34). Verified each fix against the code rather than the commit messages. `go build`, `go vet`, `go test -race` green on `services/asset/httpimpl` and `services/asset/repository`; `nginx -t` ok, plus live behaviour tests below.

All five prior findings are fixed, and the fixes brought their own regression tests.

1. **block_legacy success-as-failure** (`f0622578f`) — both success paths now `w.Close()`. Checked the remaining eight `CloseWithError(io.ErrClosedPipe)` sites in that file: all are genuine failure paths. `TestGetLegacyBlockReader_SuccessEndsWithCleanEOF` pins `io.EOF` on both shapes. The `defer r.Close()` additions (`40e81a06d`) also close a producer-goroutine leak on the hijack path that I had missed — the producer was blocked in `Write` forever holding its arena and file-store permit.
2. **Zero-byte `subtreeData` commit** (`2b211addd`) — `countingWriter` + `storer.Abort`, test asserts the blob is absent afterwards. No race on the counter: `writeChunkToWriter` only ever runs on the single result-collector goroutine, and `-race` is green.
3. **Client-gone as a server fault** (`9f268f4d0`) — hijack-and-close with the response uncommitted. Partially fixed, see item 2 below.
4. **`Cache-Control` on failures** (`c9a7fe83f`) — verified live against nginx 1.29 with the real conf:

   | case | status | X-Cache-Status | Cache-Control |
   |---|---|---|---|
   | upstream 200, first request | 200 | MISS | `public, max-age=300` |
   | same again | 200 | HIT | `public, max-age=300` |
   | upstream 500 | 500 | MISS | `no-store` |
   | backend down (nginx 502) | 502 | MISS | `no-store` |
   | blackholed backend (nginx 504) | 504 | — | `no-store` |

   The `""` arm is not the hole it looks like: nginx keeps 502/504 in `$upstream_status` for its own generated failures, so they land on `default`.
5. **`errReader` prefix truncation** (`350588eb2`) — drains incrementally, and `TestStreamOrAbort_MidStreamFailureWritesWholePrefix` asserts all 256 bytes reach the recorder, which is a better guard than the socket-level test can be.

Also confirmed end to end that the core defense actually fires. With a backend that sends 200 headers, one chunk, then closes without the chunked terminator:

```
error.log:  upstream prematurely closed connection while reading upstream
cache dir:  0 files
```

And the metric fan-out in `8b698c0d3` is complete: all nine return sites in `GetSubtree` after hash validation increment exactly once, no double counts, no path missed.

## Remaining

### 1. P2 — `proxy_cache_use_stale` still replays a poisoned body indefinitely, and the new map labels it `public`

`deploy/docker/base/asset-cache-nginx.conf:30`, inherited by the cached location — which now overrides `proxy_next_upstream` but not this.

Reproduced locally with `proxy_cache_valid 200 2s`: prime the entry with a 200, flip the backend to 500, request again.

```
1) prime cache (backend 200):   200  X-Cache-Status: MISS   good-body
   [backend flipped to 500]
2) after TTL, upstream 500:     200  X-Cache-Status: STALE  Cache-Control: public, max-age=300
3) again:                       200  X-Cache-Status: STALE
4) again:                       200  X-Cache-Status: STALE
```

Two things fall out, one of which contradicts a comment added in this PR.

- The stale entry is served **every** time, not once, so on an already-poisoned peer the 0-byte body outlives this fix for as long as requests keep arriving — each access also refreshes the `inactive=10m` timer. The description is right that escaping needs #1371's cachebust or an operator flush, but a `proxy_cache_use_stale` override in the location this PR already edits (at minimum dropping `error` and `http_500`) is the difference between a peer that self-heals and one that needs a human.
- The map's "accepted tradeoff" paragraph has it backwards. It says a stale-but-good 200 goes out labelled `no-store` because `$upstream_status` is the failed attempt's 5xx. It isn't: with `proxy_cache_background_update on` the client's request contacts no upstream at all (`uct="-" uht="-" urt="-"` in the access log), so `$upstream_status` is empty, the `""` arm applies, and the stale body ships as `public, max-age=300`. Harmless when the stale body is good — but when it is the poisoned empty one, this hop is telling downstream caches to store it. Fixing the point above fixes this too; re-keying the map does not, since a stale 200 is a 200 whichever variable you switch on.

### 2. Low — raw error text still reaches the caller

`services/asset/httpimpl/stream.go`, final return of the peek block.

The context-error reclassification landed, but the fallback is still `echo.NewHTTPError(500, readErr.Error())`, so arbitrary store/aerospike text goes to an untrusted peer. Observability is fine — `customHTTPErrorHandler` logs the full error and the new empty-generation error names the hash — so this is only about what leaves the box. A fixed message plus the existing log covers it.

### 3. Nit — the `DeadlineExceeded` arm infers "client gone" from the error's shape

`services/asset/httpimpl/stream.go`, and `TestStreamOrAbort_ClientGoneBeforeFirstByte` codifies it.

A client disconnect cancels the request context with `context.Canceled`. `DeadlineExceeded` at the peek would mean something below the handler timed out — a server fault — and it would be hijacked, closed, and counted `OK/200` with nothing logged, in the same PR that fixed the metric so operators can see failures. Unreachable today, which is why this is a nit: the request context carries no deadline (`heavyMW` is only a rate limiter), and the file blob store deliberately converts its own 25s permit deadline into `ServiceUnavailableError` (`stores/blob/file/file.go:358-362`) rather than letting a context error escape. Gating on `c.Request().Context().Err() != nil` would make the check authoritative instead of inferential, and stays correct if a store-level deadline is ever added.

Item 1 is the only one worth a decision before merge.

PR #1378, opened 2026-07-29
fix(p2p): stop blocking the websocket notification processor on per-client blockchain gRPC

gokutheengineer:
**Closes bitcoin-sv/teranode#4727.**

### Problem
Every new websocket client was handled inline on the single `startNotificationProcessor` goroutine: `sendInitialNodeStatuses` synchronously called `getNodeStatusMessage(context.Background())`, which performs several blockchain gRPC round-trips (best block header, FSM state, block persister state, block assembly state). A burst of connects serialised N gRPC calls on that one goroutine, and one slow/stalled blockchain call froze all notification broadcasts and dead-client reaping. `context.Background()` also detached the calls from shutdown. All items in the issue were confirmed still valid.

### Fix
- `Server` gains a `latestNodeStatus` atomic cache, refreshed on every `getNodeStatusMessage` run (the 10s periodic publisher, best-block-change publishes, and the cold-start fallback below). New clients are served the cached copy synchronously - no I/O on the processor goroutine.
- The cache is warmed in `Start` before the HTTP surface (and its `/p2p-ws` route) binds. This closes a hazard beyond the issue's proposed fix: the asset service (centrifuge) and the dashboard pin the current node's identity to the *first* `node_status` they receive, and an asynchronously computed initial status could otherwise lose the race against a remote peer's `node_status` broadcast during startup, permanently mis-pinning the node identity.
- Every fallible lookup in `getNodeStatusMessage` (best header, FSM state, ListPeers, block assembly state, BlockPersisterHeight) now carries forward the corresponding fields of the last cached status when it fails (ChiR1): the zero-value defaults (height 0, `fsm_state` UNKNOWN, storage "pruned", 0 counts) are indistinguishable from real values, and this message is cached, broadcast, and published to gossip, where peers use `storage` for sync-peer selection. This also covers a context deadline expiring mid-computation: the remaining lookups fail and their fields fall back individually. The BlockPersisterHeight lookup logs at debug level because a missing state key is normal on nodes without a block persister.
- Because failed fields carry forward instead of gating the cache write, the cache, the periodic broadcast, the gossip publish, and the initial status for new clients always serve the same message (ChiR2) - there is no path that broadcasts a status the cache rejected, and no frozen cold-cache entry: every tick stores fresh timestamp/uptime with last-known-good values only for the fields whose lookup failed.
- The empty-cache fallback (only reachable for servers that never ran `Start`, i.e. tests) computes the status on a separate goroutine with a 5s timeout context derived from the processor's context instead of `context.Background()`.
- Each publish tick in `publishNodeStatus` runs under a per-tick timeout, so a wedged blockchain call can't stall the publisher (and freeze the cache) forever. Within a tick, the status computation is bounded to half the interval so the P2P publish always has budget left (ChiR5), and the local websocket fan-out now happens before (and regardless of) the P2P publish: a failed gossip publish can no longer freeze long-lived dashboard/centrifuge clients on a stale status while newly connecting clients get a fresh copy from the cache.

### Residual
- Startup computes the node status twice (warm-up + the publisher's immediate publish, ChiR4). Deliberate: the warm-up runs before the pubsub topics are joined, so routing it through `handleNodeStatusNotification` would publish into a not-yet-established mesh; the cost is ~5 duplicated gRPC calls once at startup.
- Fields whose dependency stays down carry forward for the duration of that outage (timestamp/uptime/sync-peer fields stay fresh); there is no explicit staleness marker per field (ChiR8). Adding one would extend the wire format; deferred unless the team wants it.
- Untested: the warm-up call inside `Start` (requires full service scaffolding); it is a plain call into the fully tested `getNodeStatusMessage`.
- In the test-only cold-start fallback, first-message ordering relative to broadcasts is not guaranteed; in production the warmed cache makes that path unreachable.
- A dropped initial status (client channel full) is still only logged, with no metric.

### Tests
Added in `services/p2p/HandleWebsocket_test.go`:
- `TestStartNotificationProcessor_SlowBlockchainDoesNotStallProcessor` - regression test; verified it fails against the old inline implementation.
- `TestStartNotificationProcessor_InitialStatusPrecedesBroadcasts` - encodes the first-node_status-identifies-our-node contract relied on by centrifuge and the dashboard.
- `TestGetNodeStatusMessage_CarriesForwardLastKnownGoodOnFailure` - pins the carry-forward invariant (ChiR1/ChiR3): best-header (hash/height/chainwork/miner), FSM, ListPeers (via a small failing peer-registry stub), and block-assembly failures fall back to cached fields, recovery serves fresh values, the persister-height success path decodes, and the cache never regresses to zeros.
- `TestSendNodeStatusToClient_DropsWhenChannelFull` and `TestSendNodeStatusToClient_MarshalError` (ChiR3).
- `TestPublishNodeStatus_BoundsEachTick` - per-tick deadline on both the initial and ticker-driven publish (`nodeStatusPublishInterval` is now a test-overridable var, same pattern as `maxConcurrentBroadcasts`), a failed P2P publish is logged and survived, and the publisher shuts down on context cancellation (ChiR3).
- `TestSendInitialNodeStatuses_SendsCachedStatusSynchronously` and `TestGetNodeStatusMessage_PopulatesCache`.
- `TestHandleNodeStatusNotification_FanOutSurvivesPublishFailure` and `TestHandleNodeStatusNotification_PublishBudgetSurvivesSlowCompute` (ChiR5): a failed P2P publish still reaches local websocket clients, and a wedged status computation cannot hand Publish an expired context.

Ran (affected package only):
```
```
Result: `ok github.com/bsv-blockchain/teranode/services/p2p` (full suite green, race detector on).


ctnguyen (changes requested):
Review at commit [2bb4ced](https://github.com/gokutheengineer/teranode/commit/2bb4ced37eb925dd1c1bbb9ecbbb3a930877f666)

This is a good fix for a real problem, and the diff is pleasantly small for what it buys. Moving the per-client blockchain round-trip off `startNotificationProcessor` is clearly right, and the extra step of warming the cache in `Start` *before* `setupHTTPServer` registers `/p2p-ws` is the part I liked most — I checked `services/asset/centrifuge_impl/centrifuge.go:470-489` and the asset service really does pin `currentNodePeerID` to the first `node_status` it sees and then only refreshes on a matching peer ID, so the ordering hazard you closed is genuine and would have been nasty to debug. `setupHTTPServer` is only reached from `Start` (`services/p2p/Server.go:645`), so the async fallback is indeed unreachable in production, and I confirmed the cached `*notificationMsg` is never mutated after `Store` — every other `notificationCh` producer allocates a fresh message, and `policyFromSettings` returns a fresh `FeePolicy`, so there's no aliasing race hiding here. The regression test that fails against the old inline implementation is the right shape of test. My concerns are all about what the new cache is allowed to hold, not about the caching idea itself.

## Blocking issues

### ChiR1 - Degraded gate only covers the best-header call, so a partly-failed status is cached as authoritative
**Problem:** `degraded` is derived from a single call (`services/p2p/Server.go:1208-1212`), but `getNodeStatusMessage` makes four more calls whose failures are absorbed into defaults that read as real values: `GetFSMCurrentState` → `fsm_state: "UNKNOWN"` (`:1224-1233`, `Warnf`), `peerRegistry.ListPeers` → `connected_peers_count: 0` (`:1387-1399`, `Warnf`), `GetBlockAssemblyState` → `tx_count`/`subtree_count: 0` (`:1404-1411`, `Debugf`), and `GetState(ctx, "BlockPersisterHeight")` → `blockPersisterHeight = 0` (`:1416-1420`, **no else branch and no log at all**), which `util.DetermineStorageMode` turns into `"pruned"` unconditionally (`util/storage_mode.go:26-28`). Because `degraded` is false in all four cases, the store gate at `:1470-1472` accepts the message and overwrites a good cache. There are two ways in:

- *Independent failure of a later dependency* (a real gRPC error while `GetBestBlockHeader` succeeded). The context is still valid, so this message is cached **and** published to the gossip topic (`:1510`, `:1533`) and broadcast to websocket clients (`:1541`). The publish-side misreport is pre-existing; the cache overwrite is new in this PR.
- *Shared-deadline exhaustion*, which this PR introduces: the per-tick deadline at `:1178` (`context.WithTimeout(ctx, nodeStatusPublishInterval)`) and the warm-up deadline at `:641` (`initialNodeStatusTimeout`) are spent by `GetBestBlockHeader` first, so under a loaded blockchain service it can *succeed* while every subsequent call fails with `DeadlineExceeded` — `degraded` stays false and the defaults above are cached. In this variant `P2PClient.Publish` receives the same expired context, so gossip publication and the websocket broadcast will usually fail and return early at `:1533-1534`; I can't prove that from the tree here, so the failure I'm asserting is the definite one: the cache is overwritten either way, and the corrupted copy is then the only thing new clients are served.

**Why it matters:** The cached copy is what every newly connecting websocket client receives for up to a full publish interval, so one slow or partly failing tick misreports FSM state, peer count, block-assembly counts and storage class to the dashboard and to the asset service — exactly the misreport the new guard was added to prevent. In the independent-failure variant it also reaches remote peers, who apply it via `updateStorage` (`:1108-1110`), and a peer whose `Storage != "full"` is dropped from full-node catchup eligibility (`services/p2p/peer_selector.go:259-261`). The warm-up runs the same code, so the very first cache entry is exposed to this too.
**Fix:** Treat any failed dependency as degrading, and additionally refuse to cache a message computed under an expired context:

```go
	fsmState := "UNKNOWN"
	if s.blockchainClient != nil {
		currentState, fsmErr := s.blockchainClient.GetFSMCurrentState(ctx)
		if fsmErr != nil {
			degraded = true
			s.logger.Warnf("[getNodeStatusMessage] error getting FSM state: %s", fsmErr)
		} else if currentState != nil {
			fsmState = currentState.String()
		}
	}
	// ...same for ListPeers, GetBlockAssemblyState and GetState("BlockPersisterHeight"),
	// which currently swallows its error silently.

	// A context that expired mid-computation leaves later fields at their zero
	// defaults, which are indistinguishable from real values (storage="pruned",
	// 0 connected peers, 0 txs).
	if ctx.Err() != nil {
		degraded = true
	}

	if !degraded || s.latestNodeStatus.Load() == nil {
		s.latestNodeStatus.Store(msg)
	}
```

## Non-blocking issues

### ChiR2 - Cache staleness is unbounded and disagrees with what the broadcast path sends
**Problem:** `latestNodeStatus` is only ever stored, never invalidated or aged out (`services/p2p/Server.go:1470-1472` is the sole writer, `services/p2p/HandleWebsocket.go:277` the sole reader). Three consequences follow. First, while `GetBestBlockHeader` keeps failing, the degraded gate keeps the last good entry indefinitely, so a new client is served an arbitrarily old `timestamp`, `best_height`, `uptime`, `fsm_state` and `sync_peer_*` with nothing marking it stale — not the "up to one publish interval (10s) old" bound stated in the PR description. Second, the two delivery paths contradict each other: `handleNodeStatusNotification` broadcasts the *returned* `msg` regardless of the caching decision (`:1479` then `:1541`), so a degraded status that was just rejected from the cache is still pushed to every connected client on the same tick, and the asset service adopts it because the peer ID matches (`services/asset/centrifuge_impl/centrifuge.go:480-485`). The guard therefore only protects a client's *first* message. Third, the `|| s.latestNodeStatus.Load() == nil` clause deliberately admits a fully degraded status into a cold cache, and the gate then blocks any *fresher* degraded refresh from replacing it — so if the blockchain is unreachable during the `Start` warm-up (`:641-643`), early clients are served the frozen genesis/height-0 copy until the first fully successful publish, where previously each connect recomputed and could pick up a recovered blockchain sooner.
**Why it matters:** Silent unbounded staleness is harder to diagnose than the obviously-wrong height 0 it replaces, and having the cached copy and the broadcast copy report different things about the same node within the same tick makes any dashboard bug report ambiguous.
**Fix:** Make the two paths agree rather than letting height-0 back into the cache — swapping a degraded status in once the good one ages out would just reintroduce the misreport the guard exists to prevent. The coherent version is to suppress the degraded broadcast as well, so clients keep one consistent view, and to make the degradation visible in the log instead of in the data:

```go
	msg, degraded := s.nodeStatusMessage(ctx)
	if degraded {
		// Don't contradict the cache: clients keep the last known-good status.
		s.logger.Warnf("[handleNodeStatusNotification] degraded node status not cached or broadcast; serving last known-good copy (age %s)", ageOfCached())
		return nil
	}
```

If the team would rather have clients *see* the degradation, then the cache should serve it too — the point is that one of the two paths has to change. Either way, worth logging the warm-up outcome distinctly (today a failed warm-up only surfaces as the generic `[handleNodeStatusNotification] error getting best block header` line at `:1214`, which doesn't say a cold cache was pinned to it), and worth correcting the "at most 10s stale" claim in the PR description.

### ChiR3 - New behavioural branches are untested (coverage gate failed at 53.7% against the required 80%)
**Problem:** The four added tests cover the happy paths (cached copy sent synchronously, cache populated, processor not stalled, initial status precedes broadcasts) but none of the new rules the PR description sells as the fix. Specifically untested: the degraded-does-not-overwrite branch at `services/p2p/Server.go:1470` — the single most important new invariant — plus the warm-up in `Start` (`:635-643`), the per-tick timeout in `publishNodeStatus` (`:1177-1184`), and both failure branches of `sendNodeStatusToClient` (`services/p2p/HandleWebsocket.go:295-304`). The failed coverage gate on new code lines up with that.
**Why it matters:** The degraded gate is a two-line condition that silently changes what every new client sees; without a test it can be inverted or dropped in a later refactor with nothing failing. It is also the branch ChiR1 shows to be incomplete, so it deserves a test that pins down what "degraded" is meant to cover.
**Fix:** A table test over `getNodeStatusMessage` needs no new infrastructure — `blockchain.Mock` is already used elsewhere in this package (`services/p2p/sync_coordinator_test.go:66`, `services/p2p/server_helpers_test.go:45`), so the justification comment at `HandleWebsocket_test.go:847-850` is more than sufficient. Roughly:

```go
// good status cached, then a best-header failure must not overwrite it
mockBlockchain.On("GetBestBlockHeader", mock.Anything).Return(hdr, meta, nil).Once()
require.Equal(t, uint32(100), s.getNodeStatusMessage(ctx).BestHeight)

mockBlockchain.On("GetBestBlockHeader", mock.Anything).Return(nil, nil, assert.AnError).Once()
require.Equal(t, uint32(0), s.getNodeStatusMessage(ctx).BestHeight, "returned message is degraded")
require.Equal(t, uint32(100), s.latestNodeStatus.Load().BestHeight, "cache keeps the good status")
```

Plus a channel-full case for `sendNodeStatusToClient` (full channel in, assert nothing is delivered and nothing blocks), and — once ChiR1 is fixed — a case where best-header succeeds but a later dependency fails.

### ChiR4 - Node status is computed twice during startup
**Problem:** The warm-up at `services/p2p/Server.go:641-643` computes a complete status (its result is retained in `latestNodeStatus`, so nothing is lost), and `publishNodeStatus` — started shortly after at `:715` — recomputes it from scratch via `publish()` (`:1187`) purely to have something to publish. That is roughly five duplicated gRPC round-trips (best header, FSM state, `ListPeers`, block-assembly state, `GetState`) within the same startup, and `getNodeStatusMessage` also repeats its `syncConnectionTimes` bookkeeping (`:1280-1295`) on both passes.
**Why it matters:** Minor and purely a cost question, but it is avoidable work at the moment the node is busiest.
**Fix:** Let the warm-up double as the initial publish and drop the publisher's immediate call — the warm-up runs before the notification processor exists, so the buffered `notificationCh` message is simply broadcast to zero clients:

```go
// in Start
if err := s.handleNodeStatusNotification(warmCtx); err != nil {
	s.logger.Warnf("[Start] initial node status publish failed: %v", err)
}

// in publishNodeStatus: remove the immediate publish() and rely on the ticker
```

## Recap

| ChiR1 | Incomplete degraded gate | YES | 60% |
| ChiR2 | Unbounded staleness, split paths | | 40% |
| ChiR3 | New branches untested | | 30% |
| ChiR4 | Duplicate startup computation | | 12% |

ctnguyen (changes requested):
Review at commit [c3f8fe7](https://github.com/bsv-blockchain/teranode/commit/c3f8fe7e325b5e487845b182fc4287b6b43f4d47)

This has come a long way since the last pass. Dropping the degraded gate in favour of per-field carry-forward was the better call than the fix I suggested — it removes the whole class of "which copy is authoritative" question, and carrying hash/height/chainwork/miner together as a unit (`Server.go:1375-1381`) means the advertised tip stays internally coherent instead of pairing a fresh hash with a stale height. `TestGetNodeStatusMessage_CarriesForwardLastKnownGoodOnFailure` is a genuinely good test: it walks good → failed → recovered across four independent dependencies and asserts the cache never regresses, which is exactly the invariant that was unpinned before. The `flakyPeerRegistry` stub is a nice touch for something that would otherwise have needed a mock. I re-checked the ordering hazard too: centrifuge still clears `currentNodePeerID` on every websocket read error (`services/asset/centrifuge_impl/centrifuge.go:448-455`) and re-pins on the next first message, so the warm-up in `Start` protects reconnects and not just cold start — worth more than the two lines it costs. What I found this round is mostly about the paths *around* the cache rather than the cache itself.

## Status of previous points

ChiR2's three limbs split. Of the five fields I named as able to go arbitrarily stale, `timestamp`, `uptime` and `sync_peer_*` are now recomputed on every call (the sync-peer block at `Server.go:1383-1432` reads local state and never carries forward), so only `best_height` and `fsm_state` — plus the other dependency-backed fields — still carry forward with no marker, which is ChiR8. The cold-cache freeze is gone, and the distinct warm-up log I asked for exists at `:678`. The cache-versus-broadcast divergence is gone in the form I described, but a new one appears through a different door (ChiR5), which is why this stays PARTIAL rather than FIXED. ChiR3 is covered in the shape I sketched, inverted to carry-forward polarity; the `Start` warm-up ordering and the `storage` carry-forward assertion are still open (ChiR9), and I can't confirm the coverage-gate number from the tree, so that part of the original point is unverified rather than judged. For ChiR4 the refutation holds: the warm-up at `:676` runs before `subscribeToTopic` joins the node-status topic at `:724`, so routing it through `handleNodeStatusNotification` really would publish into an unjoined topic, and dropping the publisher's immediate publish would delay the first gossip announcement by a full interval. Two extra startup round-trips is the right trade — I'd leave it as is.

| ChiR1 | Incomplete degraded gate | YES | FIXED |
| ChiR2 | Unbounded staleness, split paths | | PARTIAL FIXED |
| ChiR3 | New branches untested | | PARTIAL FIXED |
| ChiR4 | Duplicate startup computation | | NOT FIXED |

## Blocking issues

### ChiR5 - The new per-tick deadline makes a failed gossip publish silently stop the websocket broadcast

**Problem:** `publish()` derives `tickCtx` with a timeout of `nodeStatusPublishInterval` and hands that one context to `handleNodeStatusNotification` (`Server.go:1263-1270`), which uses it for both the status computation and `P2PClient.Publish`. When the blockchain service is wedged, `getNodeStatusMessage` consumes essentially the whole budget on its first call — `GetBestBlockHeader` blocks until the deadline, then the remaining lookups fail instantly and carry forward. `Publish` then receives an already-expired context, and the pubsub publish path is context-aware (the final `select` in `Topic.validate` returns `ctx.Err()`), so it errors. `handleNodeStatusNotification` returns at `Server.go:1665-1667` — **before** the `s.notificationCh <- msg` push at `:1672`. Meanwhile `getNodeStatusMessage` has already stored the message in the cache at `:1604`. Net effect: already-connected clients receive nothing, while a client connecting one second later gets a fresh status from the cache. That contradicts the PR's own claim that the cache, the periodic broadcast, the gossip publish and the initial status "always serve the same message". Note this is the inverse of what silent mode deliberately does at `:1647-1655`, where the publish is skipped but the websocket fan-out still happens. The early return itself is pre-existing; the per-tick deadline introduced here is what makes it systematically reachable (before this PR the publisher passed the unbounded service context, so `Publish` only failed on genuine pubsub errors).

**Why it matters:** This bites in exactly the degraded scenario the PR is hardening for. The dashboard and centrifuge freeze on their last received status with no indication anything is wrong — centrifuge only refreshes `cachedCurrentNodeStatus` when a node_status with a matching peer ID arrives (`services/asset/centrifuge_impl/centrifuge.go:480-485`), so the asset service serves a frozen copy too. A reconnecting client gets fresh data and a long-lived one does not, which is the worst shape for a bug report: two operators looking at the same node see different heights.

**Fix:** Local monitoring must not depend on gossip succeeding, and the publish needs budget left over. Make the fan-out unconditional:

```go
func (s *Server) handleNodeStatusNotification(ctx context.Context) error {
	msg := s.getNodeStatusMessage(ctx)
	...
	// Local websocket clients must see the status even if gossip fails.
	defer func() {
		select {
		case s.notificationCh <- msg:
		default:
			s.logger.Warnf("[handleNodeStatusNotification] notification channel full, dropped node_status for %s", msg.PeerID)
		}
	}()

	if err = s.P2PClient.Publish(ctx, s.nodeStatusTopicName, msgBytes); err != nil {
		return errors.NewError("nodeStatusMessage - publish error", err)
	}

	return nil
}
```

and reserve part of the tick for the publish rather than letting the computation eat all of it:

```go
	// Bound the computation to a fraction of the tick so Publish always has budget.
	computeCtx, cancelCompute := context.WithTimeout(ctx, nodeStatusPublishInterval/2)
	msg := s.getNodeStatusMessage(computeCtx)
	cancelCompute()
```

## Non-blocking issues

### ChiR6 - Two concurrent producers, so a stale carry-forward can clobber a newer status in the cache *and* on the wire

**Problem:** `getNodeStatusMessage` reads `cached := s.latestNodeStatus.Load()` at `Server.go:1297` and writes `s.latestNodeStatus.Store(msg)` at `:1604`, with up to five gRPC round-trips in between. Two production goroutines call it concurrently: the publisher (`go s.publishNodeStatus(ctx)`, `:755`) and the blockchain subscription listener via `handleBlockNotification` → `handleNodeStatusNotification` (`:1244`, goroutine started at `:735`). Interleaving: the publisher tick loads `cached` at height 100 and blocks in `GetBestBlockHeader`; a block at 101 arrives, that path computes fresh, stores height 101 and gossips it; the publisher's call then fails on its deadline, carries forward height 100, and stores *that*. The cache regresses and the node gossips 100 after having gossiped 101. The same interleaving can resurrect any other superseded field — including `storage`, since `persisterHeightFailed` reads the same stale `cached`. The carry-forward is what makes this newly reachable: before this PR a failed best-header lookup produced an obviously-bogus genesis message, and the store gate would have rejected it.

Note that the cache is not the only exposure, so a guard on the `Store` alone would not be enough: the value returned at `:1606` is what `handleNodeStatusNotification` converts and publishes to the gossip topic (`:1665`) and pushes to `notificationCh` (`:1673`). Any fix has to decide the freshness question for the whole compute → cache → publish → fan-out sequence, not just for the cache write.

**Why it matters:** The gossiped height and chainwork feed remote peers' sync-peer ranking and land in their registry (`sanitizeAdvertisedTip` caps but does not reject a lower height, `server_helpers.go:568-592`), so a node that appears to move backwards gets de-prioritised for no real reason. Locally, new websocket clients are served the older height for up to a full publish interval.

**Fix:** The cleanest answer is to remove the second producer rather than arbitrate between them. Have `handleBlockNotification` poke the publisher instead of computing a status itself, so `getNodeStatusMessage` has exactly one caller on the hot path and the race cannot form. This also fixes ChiR7 and removes the duplicated work, and a burst of blocks coalesces into one refresh:

```go
// field on Server, made in New: buffered depth 1
nodeStatusRefreshCh chan struct{}

// in handleBlockNotification, replacing the inline handleNodeStatusNotification call.
// A nil channel is never ready, so this stays safe for servers built in tests.
select {
case s.nodeStatusRefreshCh <- struct{}{}:
default: // a refresh is already queued; coalesce
}

// in publishNodeStatus
	for {
		select {
		case <-ctx.Done():
			s.logger.Infof("[publishNodeStatus] node status publisher shutting down")
			return
		case <-ticker.C:
			publish()
		case <-s.nodeStatusRefreshCh:
			publish()
		}
	}
```

If you'd rather keep both producers, the minimal-diff version is a compare-and-swap that also fixes up the return value — and it needs no new field, because every computation allocates a fresh `*notificationMsg`, so pointer identity is already a version tag:

```go
	// Publish this computation only if no newer one landed while we were computing.
	if !s.latestNodeStatus.CompareAndSwap(cached, msg) {
		// We lost the race, and our copy may hold carried-forward fields that the
		// winner has already superseded. Serve the winner instead: this return
		// value is also what the caller gossips and fans out to clients.
		if winner := s.latestNodeStatus.Load(); winner != nil {
			s.logger.Debugf("[getNodeStatusMessage] newer node status landed while computing; serving it instead")
			return winner
		}
	}

	return msg
```

Don't reach for a mutex around the whole computation: that would serialise the publisher and the block-notification path and reintroduce head-of-line blocking on the subscription listener, which is the bug this PR exists to remove.

### ChiR7 - The other caller of `handleNodeStatusNotification` is still unbounded and can wedge the blockchain subscription listener

**Problem:** The PR bounds the publisher's ticks but leaves the second caller alone: `handleBlockNotification` calls `s.handleNodeStatusNotification(ctx)` at `Server.go:1244` with the service-lifetime context, no deadline. That runs inline on `blockchainSubscriptionListener`, which is a single goroutine processing notifications one at a time (`:1762-1793`, `processBlockchainNotification` is called synchronously at `:1789`). A wedged blockchain gRPC inside `getNodeStatusMessage` therefore blocks every subsequent Block, Subtree and PeerFailure notification for the p2p service indefinitely — no block or subtree gossip announcements, and no catchup peer switching on failure.

**Why it matters:** This is the same defect class the PR fixes, one function over, and it is arguably worse: stalling the notification processor costs dashboard updates, whereas stalling this loop costs block and subtree propagation and the PeerFailure path that rotates a bad sync peer. It also undercuts the claim that a wedged blockchain call can no longer freeze the cache — it can, just from this goroutine.

**Fix:** Bound it the same way the publisher now does:

```go
	// Also send a node_status update when best block changes
	statusCtx, cancel := context.WithTimeout(ctx, nodeStatusPublishInterval)
	err = s.handleNodeStatusNotification(statusCtx)
	cancel()

	if err != nil {
		ctxLogger.Warnf("[handleBlockNotification] error sending node status update: %v", err)
	}
```

The refresh-channel variant of ChiR6 resolves this outright — handing the work to the publisher goroutine means this loop never makes a blockchain call for node status at all — so if you take that route you can skip this one.

### ChiR8 - Carry-forward is unbounded and stamped with a fresh timestamp, so a wedged dependency is undetectable

**Problem:** This is the remaining limb of ChiR2. Every message gets `Timestamp: time.Now()` and a fresh `Uptime` (`Server.go:1571-1585`) even when all five lookups failed and every substantive field came from `cached`. There is no per-field marker and no age bound, so a consumer cannot distinguish "height 850000, checked just now" from "the blockchain client has been unreachable for an hour, last known height 850000". Pre-PR the failure at least announced itself as a genesis/height-0 message; now it is invisible. On the gossip side this is load-bearing: `storage` gates full-node eligibility on remote peers (`peer_selector.go:331`, `p.Storage == "full"`). Worth noting how the `storage` branch actually behaves — `GetState` returns an error for an absent key (`stores/blockchain/sql/State.go:22-46`), so on a node without a block persister `persisterHeightFailed` is true on *every* call and `storage` is never recomputed after the first one. Today that pins the correct `"pruned"` (cold cache → `blockPersisterHeight == 0` → `"pruned"`), so there is no live bug, but the field is structurally frozen rather than derived: one anomalous first tick pins the wrong value for the life of the process, and a full node whose state row becomes unreadable keeps advertising `"full"` forever, even after pruning has removed the data.

**Why it matters:** Silent indefinite staleness is harder to diagnose than an obviously wrong value, and for `storage` and `best_height` the staleness leaves the local process and changes what remote peers do.

**Fix:** Bound the carry-forward by age so a long outage eventually surfaces as unknown rather than as confident stale data. Record when each cached copy was computed and stop trusting it after a few intervals:

```go
const maxCarryForwardAge = 5 * nodeStatusPublishInterval

// alongside latestNodeStatus
cachedAt atomic.Int64 // unix nanos of the last successful computation

cached := s.latestNodeStatus.Load()
if cached != nil && time.Since(time.Unix(0, s.cachedAt.Load())) > maxCarryForwardAge {
	s.logger.Warnf("[getNodeStatusMessage] cached node status older than %s; not carrying fields forward", maxCarryForwardAge)
	cached = nil
}
```

For `storage` specifically, consider counting consecutive `GetState` failures and treating a persistent one as authoritative-zero (i.e. `"pruned"`), which is the safe direction for peers.

### ChiR9 - The `storage` carry-forward is the one field remote peers act on, and it is neither asserted nor logged above debug

**Problem:** `TestGetNodeStatusMessage_CarriesForwardLastKnownGoodOnFailure` mocks `GetState` to fail twice then succeed (`HandleWebsocket_test.go:1006-1010`) but never asserts anything about `Storage`, so `if persisterHeightFailed && cached != nil { storage = cached.Storage }` (`Server.go:1564-1566`) can be inverted or deleted with the suite still green. Separately, the failure that triggers it is only `Debugf` (`:1547`), so on a production node at info level there is no record that `storage` was pinned rather than computed.

**Why it matters:** `storage` is the field that changes remote behaviour (full-storage sync eligibility), which makes it the one most worth pinning down in a test and the one most worth seeing in the logs when it stops being derived from live data.

**Fix:** Assert it, and log when the carry-forward actually changes the answer:

```go
	if persisterHeightFailed && cached != nil {
		if cached.Storage != storage {
			s.logger.Warnf("[getNodeStatusMessage] BlockPersisterHeight unavailable; carrying forward storage=%q instead of computed %q", cached.Storage, storage)
		}

		storage = cached.Storage
	}
```

```go
	require.Equal(t, first.Storage, second.Storage, "failed persister-height lookup must carry forward storage")
```

### ChiR10 - `TestPublishNodeStatus_BoundsEachTick` asserts from a non-test goroutine and races on the interval restore

**Problem:** The `require.True(t, ok)` at `HandleWebsocket_test.go:1097` runs inside the mock's `Run` callback, which executes on the `publishNodeStatus` goroutine. `FailNow` must be called from the goroutine running the test; from elsewhere it calls `runtime.Goexit` on that goroutine instead of failing the test cleanly. Related: the three `t.Fatal` paths return while the publisher goroutine is still live, and the deferred `nodeStatusPublishInterval = oldInterval` (`:1088`) then races with that goroutine's read of the same package-level var in `publish()` — a race report layered on top of the real failure.

**Why it matters:** Only bites on failure, but that is precisely when you want a clean, readable signal rather than a race report and a Goexit.

**Fix:** Report the assertion over the existing channel and assert on the test goroutine; restore the interval only after the publisher has exited.

```go
	mockBlockchain.On("GetBestBlockHeader", mock.Anything).Run(func(args mock.Arguments) {
		callCtx, ok := args.Get(0).(context.Context)
		hasDeadline := false
		if ok {
			_, hasDeadline = callCtx.Deadline()
		}

		select {
		case deadlineSeen <- hasDeadline:
		default:
		}
	}).Return(nil, nil, assert.AnError)
```

```go
	defer func() { cancel(); <-done; nodeStatusPublishInterval = oldInterval }()
```

## Recap

| ChiR5 | Publish failure starves websocket fan-out | YES | 55% |
| ChiR6 | Concurrent producers clobber newer status | | 45% |
| ChiR7 | Block-notification caller still unbounded | | 40% |
| ChiR8 | Unbounded carry-forward, fresh timestamp | | 35% |
| ChiR9 | Storage carry-forward unasserted, debug-only log | | 20% |
| ChiR10 | Assertion on non-test goroutine | | 10% |

freemans13 (approved):
Approving.

The core change is the right shape: moving the per-client blockchain gRPC round-trips off the single notification-processor goroutine and onto an atomically-cached status directly fixes the stall, and warming that cache in `Start` before the HTTP listener binds is a good catch — it preserves the "first node_status a client sees is our own node" contract that centrifuge and the dashboard both depend on, which the issue's own proposed fix would have broken.

Verified locally against the branch:

- `SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./services/p2p/...` — ok, 13.5s, no races
- `go vet ./services/p2p/...` — clean
- `golangci-lint run ./services/p2p/...` — clean apart from one pre-existing `prealloc` hit in `mock.go` that this PR doesn't touch

I also checked the things that looked risky and they're fine: the new `bestBlockHeader = nil` path is nil-guarded at every downstream use, the fallback goroutine in `sendInitialNodeStatuses` can't panic on a closed channel (client channels are only removed from the map, never closed), and every new context has a matching cancel.

One out-of-scope observation while tracing the carry-forward through to the receiving side, noted here only so it doesn't get lost — it is pre-existing and this PR neither causes nor worsens it. `PeerSelector.isEligible` never consults `LastMessageTime` or `LastSeen`, even though both are maintained on every node_status and block message and already TTL-filtered when loading a persisted registry snapshot. Combined with the receiver gating registry updates on `BestHeight > 0`, that means a peer which dies or wedges stays in every other node's full-node candidate list indefinitely. Worth its own issue rather than anything in this diff.

ctnguyen (approved):
Review at commit [1b16fe1](https://github.com/gokutheengineer/teranode/commit/1b16fe1a36a103febbfe4fcb92626e6da6a401f7)

The one blocking item from last pass is genuinely closed. `b7db153a2` does both halves of ChiR5 cleanly: it moves the local `notificationCh` fan-out ahead of the P2P publish and makes it unconditional — so a failed gossip publish can no longer starve already-connected websocket clients — and it bounds the status computation to `nodeStatusPublishInterval/2` via a dedicated `computeCtx`, leaving the publish real budget on the same tick. It also picked up the ChiR9 test assertion I asked for. As a side benefit the same `computeCtx` now bounds the compute even on the block-notification path, which nudges ChiR7 forward. Nothing new surfaced this pass; the remaining open points are the non-blocking ones already on record, and freemans13's approval at this head lines up with what I see. Good work bringing it home.

## Status of previous points

Caveats before the table. ChiR4 stays NOT FIXED but is settled by agreement — the prior pass accepted the author's refutation (the warm-up runs before the topic is joined, so routing it through `handleNodeStatusNotification` would publish into an unjoined mesh); it is not expected to change. ChiR5 is the substantive movement and I verified both halves against `handleNodeStatusNotification` at `Server.go:1609-1682`: the fan-out at `:1626-1630` runs before the silent-mode check and before `Publish`, unconditionally, and the compute budget split is at `:1613-1616`. ChiR7 moves to PARTIAL: `computeCtx` now bounds the *compute* to 5s even when `handleBlockNotification` calls with the unbounded service context (`:1244`), but the `Publish` at `:1675` still uses that raw unbounded `ctx`, so a wedged pubsub publish can still stall the subscription listener from this path. ChiR9 moves to PARTIAL: the storage carry-forward is now asserted (`HandleWebsocket_test.go`, `require.Equal(t, first.Storage, second.Storage, …)`), but the carry-forward is still only `Debugf` at `Server.go:1587`. ChiR1/ChiR2/ChiR3 are untouched since `c3f8fe7`; the storage-assertion sub-part of ChiR3 is the same one closed under ChiR9. ChiR6, ChiR8 and ChiR10 are untouched. (The two new tests use `require.True` inside mock callbacks, but `handleNodeStatusNotification` is called synchronously there, so those assertions run on the test goroutine and do not replicate ChiR10's defect.)

| ChiR1 | Incomplete degraded gate | YES | FIXED |
| ChiR2 | Unbounded staleness, split paths | | PARTIAL FIXED |
| ChiR3 | New branches untested | | PARTIAL FIXED |
| ChiR4 | Duplicate startup computation | | NOT FIXED |
| ChiR5 | Publish failure starves fan-out | YES | FIXED |
| ChiR6 | Concurrent producers clobber status | | NOT FIXED |
| ChiR7 | Block-notification caller unbounded | | PARTIAL FIXED |
| ChiR8 | Unbounded carry-forward, fresh timestamp | | NOT FIXED |
| ChiR9 | Storage carry-forward unasserted, debug log | | PARTIAL FIXED |
| ChiR10 | Assertion on non-test goroutine | | NOT FIXED |

PR #1380, opened 2026-07-30
p2p: harden /p2p-ws websocket against DoS (connection cap, deadlines, read pump, origin allow-list)

gokutheengineer:
**Closes bitcoin-sv/teranode#4728.**

### Problem

All items in the issue were confirmed still valid against `services/p2p/HandleWebsocket.go`:

- `CheckOrigin` always returned true and the P2P HTTP server set CORS `AllowOrigins: ["*"]`, so any website could read the node-status feed cross-origin.
- No cap on concurrent upgrades: each request allocated a 100-slot channel and spawned goroutines with nothing bounding live connections.
- `WriteMessage` ran with no write deadline. A client that stopped reading filled its channel; the broadcaster evicted the channel after 1s, but the writer goroutine stayed blocked on `WriteMessage` forever and the socket was never closed (goroutine + fd leak per malicious connection).
- No read pump, read deadline, or pong handler, so half-open connections and control frames were never handled.

### Fix

- **Connection caps**: new setting `p2p_websocket_max_connections` (default 1000, 0 disables). A slot is claimed before the upgrade; requests beyond the cap get HTTP 503. A per-source cap additionally prevents a single host from taking the whole pool - tunable via `p2p_websocket_max_connections_per_source` (0 = auto max(4, cap/20), -1 disables it for deployments where a proxy/NAT legitimately funnels many clients through one source address). Sources in `p2p_websocket_trusted_source_cidrs` bypass both caps; the default is **loopback only, in all contexts, by design**: in the shipped k8s/operator and docker topologies the asset-service bridge connects from its own pod/container IP and is already protected by its per-source budget, while trusting a broader range (e.g. RFC1918) would make the caps inert there - external traffic arrives via the L7 Traefik ingress or docker's NAT gateway, whose addresses sit inside RFC1918, so every remote client would take the bypass. Behind an L7 proxy, source-IP caps fundamentally cannot distinguish external clients (they collapse into the proxy's per-source budget); the conf and longdesc state this topology limitation explicitly rather than implying the cap polices it. Sources are keyed on the transport-level `RemoteAddr`, never forwarded-for headers, which are caller-supplied. Rejections warn at most once per minute (time-throttled, not once-per-process) so an ongoing saturation stays visible at default log level for the life of the process, and the same throttle governs the warning that fires when live trusted-bypass connections exceed the per-source budget an untrusted host would get (default 50 - far before exhaustion, not at the global cap) - the signature of a reverse proxy fronting `/p2p-ws` from a trusted address and silently voiding the caps. The sentinel value `none` disables the bypass entirely (the settings loader treats an empty value as "use the default", so an empty list is otherwise inexpressible - needed when a reverse proxy terminates on the same host, whose loopback source would silently bypass the caps).
- **Write deadlines + keepalive pings**: every write (data and ping) is preceded by a 10s write deadline; the writer pings on a 54s period and exits on any write error.
- **Read pump**: 1KB read limit, 60s read deadline refreshed by a pong handler, and an inbound frame budget (100 frames per pong window) so an admitted client cannot spend server read-path budget without bound - the endpoint is publish-only, though small benign frames (e.g. the dashboard wstest page) stay tolerated. The budget covers **control frames too**: gorilla answers pings and consumes pongs inside `ReadMessage` without surfacing them, so the ping handler (which mirrors gorilla's default pong reply) and the pong handler both count against the same budget - otherwise a ping flood would reflect pongs unmetered and a pong flood would refresh the idle deadline forever. The upgrader also gets a 10s `HandshakeTimeout`.
- **Slow-loris closure at the HTTP layer**: all four HTTP server timeouts (read header, read, write, idle) are now set on the p2p echo server. The timeout code itself landed on main via #1544 (this branch had the same fix and adopted main's const-based version on merge); this PR contributes the regression test that pins all four so none can silently regress. Without `ReadHeaderTimeout` a client dribbling header bytes held an fd and a net/http goroutine forever; without `ReadTimeout` the same applied to a withheld request body. Neither is counted against the websocket cap. Hijacked `/p2p-ws` connections are unaffected: net/http clears its deadlines on hijack.
- **Origin allow-list**: `p2p_websocket_allowed_origins` gates the upgrade-time `CheckOrigin`, and the echo CORS middleware now reuses the exact same matcher via `AllowOriginFunc` (exact, case-insensitive, bare `*` wildcard) so both surfaces share one strict semantic instead of echo's looser glob/subdomain matching. Empty (default) preserves the historical allow-all behaviour; requests without an `Origin` header are always allowed.
- **Close on every exit path**: a single teardown path deregisters the client channel synchronously, cancels the per-connection context, closes the socket, and joins both pump goroutines, so no goroutine outlives the connection slot it is accounted against. A broadcast eviction (client too slow to consume notifications, yet protocol-healthy) now **cancels and closes** the evicted connection instead of leaving it muted forever holding a slot - the registry stores each connection's cancel func for exactly this.

Design notes beyond the issue's proposed fix:

- Client registration/removal is synchronous in the connection handler rather than via `newClientCh`/`deadClientCh` queues; the channel-based design allowed out-of-order add/remove (leaking reader-less channels) and a blocking teardown send that could wedge holding a slot.
- The initial `node_status` lookup is bounded by a 5s timeout so a hung blockchain backend cannot wedge connection setup.
- Keepalive timeouts are per-`Server` fields (`wsTimeouts`) so tests shrink them without racing on globals; `websocketTimeouts()` floors non-positive overrides to the defaults and clamps ping period below pong wait, so a bad override can never hand `time.NewTicker` a zero period (goroutine panic = process down) or evict every healthy connection each ping cycle (P3).
- Docs updated alongside: settings reference table, `settings.conf`, the websocket section of `docs/topics/services/p2p.md`, and both plantuml websocket diagrams (sequence + activity) regenerated to show the current synchronous-registration flow instead of the removed queues.

### Residual

- No Prometheus metrics for active/rejected websocket connections; the p2p service has no collector infrastructure today, so adding it belongs in a separate change.
- The trusted-source carve-out is CIDR-based rather than a separate internal-only listener; behind an L4 load balancer that rewrites source addresses, operators should route internal consumers directly to the node (documented in the setting's longdesc).

### Tests

Added:

- `TestWSConnLimiter` - admission unit tests: total cap, per-source cap (auto, fixed override, and -1 disable), release accounting, trusted-CIDR bypass (IPv4/IPv6), invalid CIDR entries ignored, cap disabled, the trusted-bypass overload warning (live bypass count above the untrusted per-source budget flags the caps as non-binding; a bridge-sized population does not), and the `none` sentinel (clears the trust list, loopback becomes subject to the caps).
- `TestHandleWebSocket_ConnectionCap` / `_ConnectionCapDisabled` / `_PerSourceCap` / `_TrustedSourceBypassesCap` - 503 beyond the caps, slot released on disconnect, loopback bypass under a saturated cap.
- `TestHandleWebSocket_BroadcastEvictionClosesConnection` - regression for the eviction path: with both read/write deadlines at 30s and a cap of 1, flooding a non-draining client must close its connection and free the slot within seconds; only the eviction path can pass this.
- `TestHandleWebSocket_InboundFrameLimits` - oversized message, data-frame budget exhaustion, and ping/pong control-frame floods all close the connection; the assertion rejects client-side read timeouts so the server not closing cannot false-pass.
- `TestSetupHTTPServer_Hardening` - pins all four HTTP server timeouts and the CORS matcher semantics (allowed origin echoed, disallowed and subdomain origins get no CORS header).
- `TestHandleWebSocket_OriginRestriction`, `_SlowClientEvicted`, `TestHandleClientMessages_Ping`, `TestOriginAllowed` - from the earlier rounds: origin checks, silent-client eviction via read deadline, ping/write-deadline discipline.
- `TestWebsocketTimeouts_PingPeriodClamped` - extended to pin that zero/negative timeout overrides floor to positive defaults (a zero ping period would panic `time.NewTicker` in the writer goroutine).

Run (green, race detector on):

```
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./settings/...
```


ctnguyen (changes requested):
Review at commit [7041aec](https://github.com/gokutheengineer/teranode/commit/7041aec4c94b3c0e365d29c236b6ddb717601155)

This is a genuinely good piece of hardening work on an endpoint that badly needed it. The write-deadline + read-pump + pong-handler triangle is put together correctly, the single teardown `defer` at `services/p2p/HandleWebsocket.go:428-434` really does join both pumps without deadlocking (write pump exits on `connCancel`, read pump on `ws.Close`), and dropping `newClientCh`/`deadClientCh` in favour of synchronous registration removes a class of ordering and wedge bugs rather than papering over them. The client that stops reading entirely — the exact case in the issue — is now handled properly: the 10s write deadline trips, the writer returns, the socket closes, the slot comes back. Bounding the initial `node_status` lookup and moving the keepalive knobs onto the `Server` instead of package globals are both the right calls, and the tests are real tests that dial actual sockets through echo. The blocking items below are all about coverage gaps at the edges of that design rather than about the design itself.

## Blocking issues

### ChiR1 — Slow-header clients never reach the cap, so the slow-loris vector stays open

**Problem:** `activeConns.Add(1)` happens inside the echo handler (`services/p2p/HandleWebsocket.go:397`), i.e. only after net/http has fully read the request headers and routed the request. Everything before that point is unbounded: `echo.New()` allocates a zero-value `Server: new(http.Server)` (`echo.go:364`) with no `ReadHeaderTimeout`, `ReadTimeout`, or `IdleTimeout`, `setupHTTPServer` never sets them (`services/p2p/Server.go:572-598`), and `StartHTTP` hands that server straight to `Serve` (`services/p2p/Server.go:1682`). `wsHandshakeTimeout` does not help either — gorilla applies `HandshakeTimeout` server-side only to writing the 101 response (`server.go:259-276` in `gorilla/websocket@v1.5.4-...`: `SetWriteDeadline` around `netConn.Write`, then cleared), never to reading the request. The comment at `HandleWebsocket.go:188-190` claims it stops "a stalled client [from holding] a connection slot without completing the upgrade"; it does not.

**Why it matters:** An attacker who opens sockets to `:9906` (bound on all interfaces by default, `settings.conf:901`) and dribbles header bytes holds one fd plus one net/http goroutine per connection, indefinitely, and is never counted against `p2p_websocket_max_connections`. That is precisely the "classic slow-loris against a public endpoint" wording from #4728, and it survives this PR untouched — the cap only governs clients well-behaved enough to finish their headers. The fix is also not exotic: the sibling public HTTP server in this repo already sets all four timeouts (`services/asset/httpimpl/http.go:628-631`), so the p2p server is the outlier.

**Fix:** Set the timeouts on the p2p server in `setupHTTPServer`, mirroring the asset service, and correct the misleading comment:

```go
e := echo.New()
e.HideBanner = true
e.HidePort = true
e.Server.ReadHeaderTimeout = 10 * time.Second
e.Server.IdleTimeout = 120 * time.Second
```

`ReadHeaderTimeout` is the one that closes the gap; `IdleTimeout` bounds idle keep-alive connections between requests. Neither endangers a live `/p2p-ws` socket, and for the record neither would `ReadTimeout`/`WriteTimeout`: `net/http` arms those deadlines before invoking the handler (`server.go:992-997`), but `hijackLocked` clears both with `rwc.SetDeadline(time.Time{})` (`server.go:325`) the moment gorilla hijacks, after which the websocket path manages its own deadlines. They are simply not needed here — the only non-hijacked route is `/health`.

### ChiR2 — A protocol-healthy client evicted by the broadcast timeout is muted forever and keeps its slot

**Problem:** `clientChannelMap.broadcast` evicts a backed-up client with a bare `cm.remove(ch)` (`services/p2p/HandleWebsocket.go:152-156`) and nothing tells the connection handler, which only returns on `connCtx.Done()`, `writeDone`, or `readDone` (`:468-472`). This is not the "client stopped reading" case — that one is handled correctly by the write deadline. It is the narrower case of a client whose socket keeps draining and which keeps answering pings, but whose 100-slot channel filled because its *application*-level consumption stalled past the 1s broadcast timeout. Once evicted, the writer drains the residual backlog, the ping ticker keeps succeeding, the pong handler keeps refreshing the read deadline, and the connection settles into a permanently subscribed-to-nothing state that holds one of the `WebSocketMaxConnections` slots until the process exits (`activeConns` is only decremented by the handler's `defer` at `:408`).

**Why it matters:** This is the one exit path the PR's own teardown comment ("Runs on every exit path", "no goroutine outlives the connection slot it is accounted against") does not cover, and the new cap converts a pre-existing mute into a permanent capacity leak that accrues over uptime. The consumer that hurts most is the asset service bridge (`services/asset/centrifuge_impl/centrifuge.go:311`): its `readMessages` loop blocks in `ReadMessage()`, gorilla handles ping frames internally and never returns them, so the bridge sees no error, never flips `clientConnected`, and never redials. One eviction silently kills the node's live block/`node_status` feed with no log on either side.

**Fix:** Give the registry a teardown handle so eviction closes the connection instead of orphaning it:

```go
type clientEntry struct{ cancel context.CancelFunc }

type clientChannelMap struct {
    sync.RWMutex
    channels map[chan []byte]clientEntry
}

// in broadcast, on timeout:
case <-timer.C:
    logger.Errorf("Timeout sending data to client, closing connection")
    cm.removeAndCancel(ch) // delete from map, then entry.cancel()
```

`connCancel` already unblocks the writer, which trips the handler's `select` and runs the existing teardown.

Please pair it with a regression test, and note that the obvious shape of that test is a false green. Asserting only `clientChannels.contains(ch) == false` passes at PR head. So does "connect a client that never drains, then assert the socket closed and the slot freed" — that client eventually trips the 10s write deadline and is torn down correctly today, so the test would pass for a reason unrelated to eviction. The assertion has to be **promptness**, with the deadline paths ruled out by construction:

```go
s.wsTimeouts = &wsTimeouts{writeTimeout: 30 * time.Second, pongWait: 30 * time.Second, pingPeriod: 27 * time.Second}
// cap of 1, via newWebSocketTestServer
// connect a client, stop draining it, push >100 notifications to overflow
// its channel and trigger the 1s broadcast timeout, then:
require.Eventually(t, dialSucceeds, 5*time.Second, 50*time.Millisecond,
    "eviction must close the connection and release the slot")
```

With both deadlines at 30s, nothing but the eviction path can close that connection inside the 5s bound, so the test fails at PR head and passes with the fix. A slow-but-draining reader reproduces the exact production shape more faithfully, but it is harder to drive deterministically and the timing assertion above already discriminates correctly.

### ChiR3 — Global cap with no trusted-source carve-out lets an attacker keep the asset bridge from reconnecting

**Problem:** `activeConns` is one process-wide counter, first-come-first-served (`services/p2p/HandleWebsocket.go:397-408`), on an unauthenticated endpoint bound to all interfaces by default. There is no per-source sub-cap, no reserved capacity, and no idle timeout — a connection that merely answers pings is held forever, so 1000 cheap sockets from one host occupy the whole default cap.

**Why it matters:** The cap is admission-only, so an already-established bridge connection is not displaced — the exposure is at connect time. But that is exactly when it matters: on asset-service restart, on a p2p-service restart, or after any transient drop (including one caused by ChiR2). If the bridge cannot get in, `cachedCurrentNodeStatus` stays nil and `authMiddleware` rejects **every** centrifuge websocket client with 503 (`services/asset/centrifuge_impl/centrifuge.go:533-537`), so a p2p-side saturation becomes a total outage of the asset service's realtime API. Cost to the attacker is ~1000 sockets, well below what fd/goroutine exhaustion required before this PR.

**Fix:** Do not key the carve-out on `ip.IsPrivate()` — behind an ingress, NAT, or an L7 load balancer, `RemoteAddr` is private for *all* traffic including the internet, so that check either grants the reservation to everyone or to no one depending on topology. Use an explicit trust boundary instead, in rough order of robustness:

1. Bind a second, internal-only listener for the bridge (separate `p2p_websocket_internal_listen_address`), and apply the public cap only to the public one. Cleanest, no trust inference at all.
2. Failing that, an operator-configured trusted-CIDR list (`p2p_websocket_reserved_source_cidrs`) that reserves `N` slots, matched against `RemoteAddr` — explicit, auditable, and correct behind ingress only if the operator says so.
3. Independently of the above, a per-`RemoteAddr` sub-cap (e.g. `max(4, maxConns/20)`) so no single source can take the whole pool.

Whichever is chosen, match on `c.Request().RemoteAddr`, not `c.RealIP()` — see ChiR6.

## Non-blocking issues

### ChiR4 — Read pump accepts unlimited inbound frames, and the 1KB limit is an untested hard disconnect

**Problem:** The read pump is an unthrottled `for { ws.ReadMessage() }` (`services/p2p/HandleWebsocket.go:444-452`) that discards whatever it gets. Data frames do not refresh the read deadline (only the pong handler does, `:440-442`), so a client must answer pings to stay alive — trivial — but having done so it may stream 1KB frames at line rate forever, with a gorilla allocation and parse per frame. At the other end, `SetReadLimit(wsMaxReadBytes)` means a single message over 1KB makes gorilla send a Close(1009) and return `ErrReadLimit` (`conn.go:932-936`), which tears the connection down.

**Why it matters:** `/p2p-ws` is publish-only, so there is no reason for an admitted client to be able to spend server read-path budget without bound. And the 1KB ceiling is an undocumented protocol contract that disconnects on violation — the repo's own test page does send an inbound message on this endpoint (`ui/dashboard/src/routes/wstest/+page.svelte:40`), so the "clients may send benign messages" path is live, and nothing pins the boundary behaviour.

**Fix:** Keep tolerating small benign frames (closing on any data frame would break the wstest page), but bound them — a simple counter that closes the connection after N frames per `pongWait` window is enough:

```go
inbound, windowEnd := 0, time.Now().Add(to.pongWait)

for {
    if _, _, err := ws.ReadMessage(); err != nil {
        return
    }

    if now := time.Now(); now.After(windowEnd) {
        inbound, windowEnd = 0, now.Add(to.pongWait)
    }

    if inbound++; inbound > wsMaxInboundFrames {
        s.logger.Debugf("Client exceeded inbound frame budget, closing connection")
        return
    }
}
```

Add a test for the oversized-frame path too, so the 1KB contract is asserted rather than incidental.

### ChiR5 — One setting, two different origin-matching semantics

**Problem:** `p2p_websocket_allowed_origins` feeds both `originAllowed` (`services/p2p/HandleWebsocket.go:241-253`: exact match, case-insensitive, only a bare `"*"` is a wildcard) and echo's CORS middleware (`services/p2p/Server.go:581-589`). Echo compiles each entry into an anchored regex with `*` → `.*` and additionally runs `matchSubdomain`, and its exact comparison is case-sensitive. The two agree on nothing except a plain lowercase origin.

**Why it matters:** `p2p_websocket_allowed_origins = https://*.example.com` is a natural thing for an operator to write. Echo's CORS accepts it; the websocket upgrade 403s every browser on that domain. The failure is invisible from the config's point of view and surfaces only as a dead dashboard.

**Fix:** Reuse one matcher for both — pass `AllowOriginFunc: func(o string) (bool, error) { return originAllowed(o, allowedOrigins), nil }` to `CORSWithConfig` so the CORS path inherits the strict semantics. Smaller diff than teaching `originAllowed` globs, and keeps the stricter behaviour as the single source of truth.

### ChiR6 — `c.RealIP()` is caller-controlled on this server

**Problem:** The cap-rejection logs use `c.RealIP()` (`services/p2p/HandleWebsocket.go:402` and `:404`). The p2p echo instance never sets `IPExtractor` (`services/p2p/Server.go:572-589`; only `services/asset/httpimpl/http.go:165-167` does), so `RealIP()` falls back to echo's legacy behaviour of trusting `X-Forwarded-For`, then `X-Real-IP`, before `RemoteAddr` (`echo/context.go:296-317`).

**Why it matters:** On a directly exposed port those headers are attacker-supplied, so the one warn line an operator gets when the cap is first hit can be made to name any address the attacker picks. It matters more once ChiR3 introduces per-source accounting: keying that on `RealIP()` would make the sub-cap trivially bypassable.

**Fix:** Log the peer from `c.Request().RemoteAddr`, or set `e.IPExtractor = echo.ExtractIPFromXFFHeader(trustOpts...)` on the p2p server if a trusted proxy really is in front of it.

### ChiR7 — Cap-rejection warning fires once per process; teardown logs at INFO on every disconnect

**Problem:** Two opposite log-volume problems on the new paths. `capWarnedOnce` is a `sync.Once` created when the handler closure is built (`services/p2p/HandleWebsocket.go:385-388`), so the cap-exhaustion warning is emitted at most once for the process lifetime — a second, unrelated incident hours later is invisible unless debug logging is on. Meanwhile `handleClientMessages` logs `Infof("Closing WebSocket connection due to context cancellation")` at `:275`, which under the new per-connection `connCtx` fires on **every** normal disconnect (teardown calls `connCancel()` before joining the writer), where previously it only fired at process shutdown.

**Why it matters:** The signal an operator actually needs — the cap is saturated, right now — is rate-limited to nothing, while routine dashboard churn produces one INFO line per disconnect on a public endpoint.

**Fix:** Replace `capWarnedOnce` with a time-based gate (an `atomic.Int64` holding the last warn time, warn again at most once a minute), and drop `:275` to `Debugf`.

### ChiR8 — Empty origin allow-list leaves no operator signal

**Problem:** `WebSocketAllowedOrigins` defaults to empty, which means allow-all in both `originAllowed` (`:242-244`) and `setupHTTPServer` (`services/p2p/Server.go:581-584`). Nothing indicates at startup that a node with a publicly bound `p2p_httpListenAddress` is running wide open.

**Why it matters:** The allow-all default is a deliberate, documented backwards-compatibility choice and the right one — breaking existing dashboards on upgrade would be worse. The gap is purely that an operator gets no prompt to tighten it, so the new control tends to stay unused on exactly the internet-facing nodes that want it.

**Fix:** Log once at startup in `setupHTTPServer` when the list is empty and the listen address is not loopback, e.g. `p2p_websocket_allowed_origins is empty - /p2p-ws and the P2P HTTP CORS policy accept all browser origins; set it to your dashboard origin(s) on internet-facing nodes`.

### ChiR9 — `TestHandleWebSocket_SlowClientEvicted` races on the single connection slot between subtests

**Problem:** The server is configured with `WebSocketMaxConnections: 1` (`services/p2p/HandleWebsocket_test.go:1005`). The first subtest's `require.Eventually` closes its probe connection inside the closure and returns immediately; the second subtest then dials with a bare `require.NoError` (`:1037-1038`). Releasing the slot requires the server to observe the FIN, exit the read pump, and complete the teardown join — all after the client-side `Close()` has returned.

**Why it matters:** The window is small but real, and widens under `-race` and loaded CI. When it loses, the second subtest gets a 503 and fails for a reason unrelated to what it is testing.

**Fix:** Wrap the second dial in `require.Eventually` the same way the first one is, or give this server a cap of 2 so the subtests cannot contend.

### ChiR10 — Websocket diagrams still show the deleted `newClientCh`/`deadClientCh` flow

**Problem:** `docs/topics/services/p2p.md` was rewritten to describe synchronous registration, but `docs/topics/services/img/plantuml/p2p/p2p_websocket_sequence_diagram.puml` (14 hits) and its rendered `.svg` still depict the removed queues; `p2p_websocket_activity_diagram.puml` is in the same state.

**Why it matters:** The prose and the image embedded directly beneath it now contradict each other, which is worse than either being stale alone.

**Fix:** Update the `.puml` sources to the registration → node_status → broadcast → teardown flow and regenerate the SVGs.

## Recap

| ChiR1 | Slow-header clients bypass cap | YES | 80% |
| ChiR2 | Eviction leaks slot, mutes client | YES | 76% |
| ChiR3 | Cap blocks asset bridge reconnect | YES | 68% |
| ChiR4 | Unbounded inbound frame budget | | 48% |
| ChiR5 | Origin matching differs from CORS | | 45% |
| ChiR6 | Spoofable RealIP in cap logs | | 30% |
| ChiR7 | Cap warn once; INFO per disconnect | | 28% |
| ChiR8 | No signal on empty origin list | | 24% |
| ChiR9 | Slot race between eviction subtests | | 16% |
| ChiR10 | Stale websocket diagrams | | 10% |

ctnguyen (changes requested):
Review at commit [0107475](https://github.com/bsv-blockchain/teranode/commit/0107475e74b987242903d56c5f3c2d67c3291034)

This round is a solid response to the last one. Seven of the ten previous points are genuinely closed in the code, not just in the description: the eviction path now cancels and closes the connection (`clientChannelMap.evict`), `ReadHeaderTimeout`/`IdleTimeout` are on the p2p echo server, the CORS middleware and the upgrade share one matcher through `AllowOriginFunc`, `RealIP()` is gone from the admission path, and the diagrams actually match the code now. The new `wsConnLimiter` is clean, small and unit-tested on its own terms, and `TestHandleWebSocket_BroadcastEvictionClosesConnection` is the right shape — 30s deadlines so only the eviction path can free the slot inside the window, which is exactly the discriminating assertion the old design would have failed. `TestHandleWebSocket_InboundFrameLimits` rejecting client-side read timeouts is a nice touch too; that class of test usually false-passes. The findings below are residuals at the edges of the new admission and read paths, plus one slow-loris variant the `ReadHeaderTimeout` fix does not reach.

## Status of previous points

`ChiR4` is partial because the budget added covers data frames only — inbound control frames never surface from `ReadMessage`, so they are not counted (see `ChiR12`). `ChiR7` is partial because the teardown log did drop to `Debugf` (`HandleWebsocket.go:392`), but `capWarnedOnce` is still a `sync.Once` built once per handler, so the cap warning is still emitted at most once for the whole process lifetime — the time-gated warn was the point of that item.

| ChiR1 | Slow-header clients bypass cap | YES | FIXED |
| ChiR2 | Eviction leaks slot, mutes client | YES | FIXED |
| ChiR3 | Cap blocks asset bridge reconnect | YES | FIXED |
| ChiR4 | Unbounded inbound frame budget | | PARTIAL FIXED |
| ChiR5 | Origin matching differs from CORS | | FIXED |
| ChiR6 | Spoofable RealIP in cap logs | | FIXED |
| ChiR7 | Cap warn once; INFO per disconnect | | PARTIAL FIXED |
| ChiR8 | No signal on empty origin list | | NOT FIXED |
| ChiR9 | Slot race between eviction subtests | | FIXED |
| ChiR10 | Stale websocket diagrams | | FIXED |

## Blocking issues

### ChiR11 — `ReadTimeout` is still unset, so a withheld request body holds a goroutine and fd forever

**Problem:** `setupHTTPServer` sets only two of the four timeouts (`services/p2p/Server.go:616-617`); `ReadTimeout` and `WriteTimeout` stay zero. `ReadHeaderTimeout` closes the header-dribbling case from ChiR1, but it is disarmed the moment the headers are complete: `net/http`'s `readRequest` sets the header deadline, then replaces it with `wholeReqDeadline`, which is the zero `time.Time` when `ReadTimeout` is 0 — i.e. no deadline at all for the rest of the request. After the handler returns, `chunkWriter.writeHeader` sees a non-zero `req.ContentLength` with unread bytes below `maxPostHandlerReadBytes` and drains it with `io.CopyN(io.Discard, w.reqBody, maxPostHandlerReadBytes+1)`, a blocking read with no deadline. So `GET /health HTTP/1.1` with `Content-Length: 1` and the byte never sent parks the serving goroutine in that `CopyN` indefinitely. This is not the `/p2p-ws` route — that one hijacks and `hijackLocked` clears the deadlines — it is `/health`, or any 404/405 path on the same listener.

**Why it matters:** One fd plus one `net/http` goroutine per socket, unbounded, never counted against `p2p_websocket_max_connections`, on a port bound to all interfaces by default. That is the same "classic slow-loris against a public endpoint" outcome named in the issue this PR closes (see [bitcoin-sv/teranode#4728](https://github.com/bitcoin-sv/teranode/issues/4728)); the header phase is now closed, the body phase is not. The PR describes these timeouts as "mirroring the asset service", but the asset server sets all four (`services/asset/httpimpl/http.go:627-630`), and `ReadTimeout = 30s` is precisely the one that covers this. `TestSetupHTTPServer_Hardening` pins the two that were added, which makes the gap look intentional.

**Fix:** Add the whole-request read deadline alongside the two already there. It is safe for the websocket route for the same reason given last round — `hijackLocked` calls `rwc.SetDeadline(time.Time{})` before gorilla touches the socket:

```go
e.Server.ReadHeaderTimeout = 10 * time.Second
e.Server.ReadTimeout = 30 * time.Second
e.Server.IdleTimeout = 120 * time.Second
```

and extend the `HTTP server timeouts set` subtest to `require.Positive(t, e.Server.ReadTimeout, ...)` so it cannot silently regress. `WriteTimeout` is optional here (no large responses on this listener), but adding it costs nothing and completes the mirror.

## Non-blocking issues

### ChiR12 — The inbound frame budget counts only data frames, so a ping flood walks straight past it

**Problem:** The budget lives in the read loop and increments only when `ws.ReadMessage()` returns (`services/p2p/HandleWebsocket.go:569-582`). Gorilla consumes control frames inside `advanceFrame` and never returns them to the caller (`conn.go:957-965`), so ping, pong and close frames are invisible to the counter. Worse, the connection uses gorilla's default ping handler, which answers every inbound ping with a `WriteControl(PongMessage, ...)` — so each unaccounted ping frame costs a parse plus a server-side control write. And because the pong handler refreshes the read deadline on *any* pong, solicited or not (`:560-562`), a client can hold the connection open indefinitely with unsolicited pongs while never sending a single data frame.

**Why it matters:** The stated purpose of this budget is that "an admitted client cannot spend server read-path budget without bound". As written, an admitted client can spend it without bound — it just has to use control frames instead of text frames. There is no meaningful byte amplification, so this is a CPU/syscall nuisance bounded by the connection cap rather than a hard DoS, but it does mean the ChiR4 mitigation does not cover the cheapest way to abuse the read path.

**Fix:** Move the accounting into a shared counter and charge control frames too. Both handlers are invoked synchronously on the read goroutine, so a plain closure over locals is race-free, and returning a non-nil error from a handler aborts `ReadMessage` and drops into the existing teardown:

```go
inbound, windowEnd := 0, time.Now().Add(to.pongWait)

charge := func() error {
    if now := time.Now(); now.After(windowEnd) {
        inbound, windowEnd = 0, now.Add(to.pongWait)
    }
    if inbound++; inbound > wsMaxInboundFrames {
        return errors.NewProcessingError("websocket inbound frame budget exceeded")
    }
    return nil
}

ws.SetPongHandler(func(string) error {
    if err := charge(); err != nil {
        return err
    }
    return ws.SetReadDeadline(time.Now().Add(to.pongWait))
})

ws.SetPingHandler(func(msg string) error {
    if err := charge(); err != nil {
        return err
    }
    _ = ws.WriteControl(websocket.PongMessage, []byte(msg), time.Now().Add(to.writeTimeout))
    return nil
})
```

with the read loop calling the same `charge()`. A third subtest in `TestHandleWebSocket_InboundFrameLimits` that floods `WriteControl(PingMessage, ...)` would pin it. While you are there: the budget path closes the socket with no close frame, unlike the oversized-message path which sends 1009 — a `CloseMessage` with `ClosePolicyViolation` would make the two consistent and tell the client why.

### ChiR13 — Trusted CIDRs are an unlimited bypass, and the setting's own guidance pushes operators to widen them

**Problem:** A source matching `p2p_websocket_trusted_source_cidrs` returns `func() {}, true` before the mutex is ever taken (`services/p2p/HandleWebsocket.go:335-341`) — it is not counted, and there is no secondary ceiling of any kind. The setting's `longdesc` recommends "Add the cluster/pod CIDR when the asset service connects cross-host (e.g. Kubernetes)" (`settings/p2p_settings.go`), which for a typical cluster means a `/16`.

**Why it matters:** In the deployment the docs describe, an L7 ingress terminating public traffic runs inside that same pod CIDR, so every internet-sourced connection arrives with a trusted `RemoteAddr` and both caps are silently off for the whole public surface — the exact unbounded-connection condition this PR exists to remove, reintroduced by following the documentation. The previous round suggested a *reservation* (trusted sources get guaranteed slots) rather than a *bypass* precisely so the pool stays bounded either way; the bypass gives the asset bridge its guarantee at the cost of removing the ceiling.

**Fix:** Keep the bypass semantics for the public caps but give the trusted tier its own bound, so a wide CIDR degrades to "generous" rather than "unlimited":

```go
// in wsConnLimiter
trustedTotal    int64
maxTrustedTotal int64 // e.g. 2 * maxTotal, or its own setting

// in acquire, on a trusted match:
l.mu.Lock()
defer l.mu.Unlock()
if l.maxTrustedTotal > 0 && l.trustedTotal >= l.maxTrustedTotal {
    return nil, false
}
l.trustedTotal++
return func() { l.mu.Lock(); l.trustedTotal--; l.mu.Unlock() }, true
```

Either way, the `longdesc` should say out loud that entries here are exempt from every limit, so widening the list to a pod CIDR is a deliberate decision rather than a side effect of getting the bridge connected.

### ChiR14 — The per-source cap keys on the exact IP, which an IPv6 client can dilute for free

**Problem:** `acquire` keys `perSource` on the host part of `RemoteAddr` verbatim (`services/p2p/HandleWebsocket.go:330-355`). For IPv4 that is a real constraint — addresses cost money. For IPv6 it is not: a client with an ordinary residential or hosting `/64` can source every connection from a different `/128` at zero cost, so each one is a fresh "source" with a fresh `max(4, cap/20)` allowance.

**Why it matters:** ChiR3 asked for the per-source cap so "no single source can take the whole pool". Against an IPv4 attacker that holds; against an IPv6 attacker it reduces to arithmetic — at the default cap, twenty addresses out of a `/64` fill all 1000 slots, and the per-source cap contributes nothing. The global cap is still the real defence, which is fine, but the per-source layer should not be assumed to be doing work it is not doing.

**Fix:** Normalise the key to the prefix that actually costs an attacker something — the address for IPv4, the `/64` for IPv6:

```go
func sourceKey(host string) string {
    ip := net.ParseIP(host)
    if ip == nil {
        return host
    }
    if v4 := ip.To4(); v4 != nil {
        return v4.String()
    }
    return ip.Mask(net.CIDRMask(64, 128)).String() + "/64"
}
```

and add an IPv6 case to `TestWSConnLimiter`'s `Per-source cap` subtest showing that two addresses in one `/64` share an allowance.

### ChiR15 — The per-source cap is derived, not configurable, and collapses the effective limit behind NAT or an ingress

**Problem:** `maxPerSource` is computed as `max(4, maxTotal/20)` inside `newWSConnLimiter` (`services/p2p/HandleWebsocket.go:304-306`) with no setting behind it, and it is keyed on the transport-level `RemoteAddr` by design. When the p2p HTTP listener sits behind an ingress, an L7 proxy, or carrier NAT, every legitimate client presents the same host, so the effective ceiling for the entire node is 50 at the default cap, not 1000. The only escape hatches are setting `p2p_websocket_max_connections = 0` (which also removes the global cap) or adding the proxy to the trusted CIDRs (which removes both caps for it — see ChiR13). There is no way to keep the global cap and relax the per-source one.

**Why it matters:** This is a silent availability cliff: the 51st dashboard or monitoring client gets a 503, and the only operator-visible signal is a single warn line emitted once per process lifetime (the still-open half of ChiR7). Nothing in `settings.conf` or the `longdesc` hints that the practical limit behind a proxy is `cap/20`.

**Fix:** Expose it, defaulting to the current derivation and treating 0 as "disabled":

```go
WebSocketMaxConnectionsPerSource int `key:"p2p_websocket_max_connections_per_source" default:"0" ...`
// 0 => max(4, maxTotal/20); negative => per-source cap disabled
```

and note in the `p2p_websocket_max_connections` longdesc that the per-source cap keys on the transport source address, so behind a source-rewriting proxy it applies to all clients collectively.

### ChiR16 — Cap saturation is invisible: one log line per process lifetime and no counter

**Problem:** The new admission control rejects with a 503 and nothing else. `capWarnedOnce` is a `sync.Once` built when the handler closure is created (`services/p2p/HandleWebsocket.go:506, 518-521`), so the warn fires at most once for the whole process and every later rejection goes to `Debugf`; there is no counter for active or rejected connections. The PR defers metrics on the grounds that the service has no collector infrastructure, and that rationale does check out at this commit — nothing under `services/p2p/` imports Prometheus, and `handle_catchup_metrics.go` is peer-registry RPC accounting rather than a collector.

**Why it matters:** Saturation is precisely the state an operator has to be able to see. From outside the node it is indistinguishable from "clients stopped connecting", and it is the trigger condition for both of the escape hatches above — widening the trusted CIDRs (ChiR13) or relaxing the per-source cap (ChiR15). With the ChiR7 gate still un-fixed, a saturation event hours after startup produces no output at all at default log levels, so the first two knobs this PR adds are ones an operator has no way to know they need. Worth noting too that a separate open PR on this service (#&NoBreak;1389) is already adding `services/p2p/metrics.go`, so the deferral's rationale has a short shelf life — this reads better as a tracked follow-up than as an open-ended residual.

**Fix:** Two cheap steps, neither of which needs a collector. Time-gate the warn so repeat incidents are visible (the still-open half of ChiR7), and include the live count so the line is actionable — `active()` below is a small mutex-guarded accessor over `l.total` that `wsConnLimiter` does not have yet:

```go
var lastCapWarn atomic.Int64 // unix nanos

// on rejection:
now := time.Now().UnixNano()
if prev := lastCapWarn.Load(); now-prev > int64(time.Minute) && lastCapWarn.CompareAndSwap(prev, now) {
    s.logger.Warnf("Rejecting websocket connection from %s: %d/%d slots in use (max %d per source)",
        c.Request().RemoteAddr, limiter.active(), limiter.maxTotal, limiter.maxPerSource)
}
```

Then, once `services/p2p/metrics.go` exists on main, wire that same `limiter.active()` plus a rejection counter into it. Please leave the residuals section of the PR description pointing at that follow-up so it does not get lost.

## Recap

| ChiR11 | Missing ReadTimeout, body slow-loris | YES | 70% |
| ChiR12 | Control frames bypass frame budget | | 45% |
| ChiR13 | Trusted CIDRs bypass all caps | | 42% |
| ChiR14 | Per-source cap diluted over IPv6 | | 38% |
| ChiR15 | Per-source cap fixed, breaks behind NAT | | 30% |
| ChiR16 | Cap saturation not observable | | 18% |

ctnguyen (approved):
Review at commit [4b4aa4e](https://github.com/gokutheengineer/teranode/commit/4b4aa4efe330bd5ff0ba11c357a449b149500e27)

The last blocker is genuinely closed. Commit 646239c9b sets `ReadTimeout = 30s` and `WriteTimeout = 120s` on the p2p echo server (`services/p2p/Server.go:621-622`), which is exactly the whole-request read deadline ChiR11 asked for — the header phase closed by `ReadHeaderTimeout` no longer leaves the body phase unbounded, so the `GET /health` withheld-body park is gone. The "mirroring the asset service" wording now actually holds: the asset server uses the same four values (`services/asset/httpimpl/http.go:628-631`), 30/120/120/10s, and this listener now matches them. And the fix came with the regression guard it needed — `TestSetupHTTPServer_Hardening`'s "HTTP server timeouts set" subtest gained `require.Positive` assertions on both `ReadTimeout` and `WriteTimeout` (`services/p2p/HandleWebsocket_test.go:1357-1362`), so the gap can't silently reopen. With that, there are no blocking issues left on this PR. The non-blocking residuals from the previous round are unchanged — documented residual risks already captured in the status table below — and this round adds only one small log-hygiene note.

## Status of previous points

Only one substantive commit landed since review-2 (0107475): 646239c9b, which addresses ChiR11. The two merge commits (a07387198, 4b4aa4efe) brought in unrelated blockassembly/blockvalidation/model changes; the only `settings/` and `settings.conf` edits among them add the blockvalidation `subtree_data_fetch_timeout` key and alter no p2p websocket setting (grep for `websocket` in the settings diff is empty), and `services/p2p/` and the websocket docs are untouched. So every ChiR other than ChiR11 is byte-for-byte identical to when it was raised and carries its review-2 verdict forward: ChiR4 and ChiR7 remain PARTIAL for the exact reasons stated last round (ChiR4's budget still counts data frames only; ChiR7's teardown log dropped to `Debugf` but `capWarnedOnce` is still a per-process `sync.Once`), and ChiR12–ChiR16 — all new in review-2 and none touched — remain open. ChiR11 is verified FIXED against the current head, including the test assertion the fix required.

| ChiR1 | Slow-header clients bypass cap | YES | FIXED |
| ChiR2 | Eviction leaks slot, mutes client | YES | FIXED |
| ChiR3 | Cap blocks asset bridge reconnect | YES | FIXED |
| ChiR4 | Unbounded inbound frame budget | | PARTIAL FIXED |
| ChiR5 | Origin matching differs from CORS | | FIXED |
| ChiR6 | Spoofable RealIP in cap logs | | FIXED |
| ChiR7 | Cap warn once; INFO per disconnect | | PARTIAL FIXED |
| ChiR8 | No signal on empty origin list | | NOT FIXED |
| ChiR9 | Slot race between eviction subtests | | FIXED |
| ChiR10 | Stale websocket diagrams | | FIXED |
| ChiR11 | Missing ReadTimeout, body slow-loris | YES | FIXED |
| ChiR12 | Control frames bypass frame budget | | NOT FIXED |
| ChiR13 | Trusted CIDRs bypass all caps | | NOT FIXED |
| ChiR14 | Per-source cap diluted over IPv6 | | NOT FIXED |
| ChiR15 | Per-source cap fixed, breaks behind NAT | | NOT FIXED |
| ChiR16 | Cap saturation not observable | | NOT FIXED |

## Non-blocking issue

### ChiR17 — Broadcast eviction logs at `Errorf` on a path that is now normal and remotely triggerable

**Problem:** When a client's 100-slot channel stays full past the 1s broadcast timeout, `broadcast` logs `logger.Errorf("Timeout sending data to client, closing connection")` and then evicts (`services/p2p/HandleWebsocket.go:172-176`). Before the ChiR2 fix that line marked a genuine anomaly — the writer wedged and the slot leaked. After the fix, eviction is a clean, self-healing teardown: `cm.evict` cancels the connection, the writer unblocks, the socket closes, the slot returns. Yet it still logs at ERROR, and any client on the unauthenticated, all-interfaces-by-default `/p2p-ws` endpoint can trigger it just by consuming slower than the broadcast rate for one second — a dashboard on a poor network qualifies.

**Why it matters:** This is the same log-hygiene family as ChiR7/ChiR16 but on a line neither named. It puts remote-triggerable, self-recovering events into the ERROR stream: one ERROR line per evicted connection, bounded by the connection cap but easily a burst when a notification fans out to many backed-up clients at once. It both cries wolf to operators who page on ERROR and gives an attacker a cheap way to colour the error log, on the very endpoint this PR is hardening.

**Fix:** Drop it to `Warnf` (or `Debugf`), matching the level the teardown path already settled on in ChiR7, so the level reflects "expected, handled" rather than "something failed":

```go
case <-timer.C:
    logger.Warnf("Timeout sending data to client, closing connection")
    cm.evict(ch)
```

icellan (commented):
Re-reviewed at `caec5ce53`. Verified locally: `go vet ./services/p2p/... ./settings/...` clean, and the websocket hardening tests pass under `-race`.

The lifecycle rewrite is sound. Teardown is single-path and the defer order (teardown → connCancel → release) means the slot is released only after both pumps join; evict-with-cancel closes instead of muting; write deadlines, read limit, pong-refreshed read deadline and the frame budget all behave as documented. No correctness bug in teardown, limiter accounting, or the broadcast/evict interaction. What's left is config and observability.

**P1 — trusted-CIDR default is wrong for the operator and docker contexts**

`settings.conf:1013` ships `p2p_websocket_trusted_source_cidrs = 127.0.0.1/32|::1/128` with no per-context override, but `settings.conf:944-946` put the asset service's bridge on a non-loopback source in exactly the deployments that matter:

```
p2p_httpAddress.operator = peer.${clientName}.svc.cluster.local:${P2P_HTTP_PORT}
p2p_httpAddress.docker.m = peer:${P2P_HTTP_PORT}
```

In k8s/operator and docker.m the bridge connects from a pod/container IP, isn't trusted, and is subject to both caps. Attacker holds all 1000 slots across ≥20 source IPs → the asset service's centrifuge listener drops and its redial gets 503 forever. That's the lockout the setting exists to prevent. Add `.operator` / `.docker.m` overrides, or key the bypass on something other than source IP.

**P2 — per-source cap keyed on RemoteAddr collapses behind a proxy/NAT**

`services/p2p/HandleWebsocket.go:347`. `maxPerSource = max(4, maxTotal/20)` = 50 at default. Front /p2p-ws with an ingress, L7 proxy or NAT and every client shares one `RemoteAddr` host — the 51st legitimate dashboard gets 503 while 950 slots sit idle. Inverse is worse: a proxy terminating on the same host has source 127.0.0.1, the default trusted CIDR, so *all* caps are bypassed and the hardening silently does nothing. RemoteAddr over XFF is the right call; the gap is there's no way to disable the per-source cap, and no warning when a loopback fronting proxy voids it.

**P2 — saturation is invisible after the first rejection, for the process lifetime**

`services/p2p/HandleWebsocket.go:525,537`. `capWarnedOnce` is a `sync.Once` per route registration, so only the first-ever rejection logs at WARN, everything after is DEBUG. One stray rejection during startup burns it and a real saturation event weeks later produces nothing at default log level — and there are no metrics either. Time-throttled warn (at most once per minute) instead of one-shot.

**P3 — `websocketTimeouts()` doesn't guard non-positive durations**

`services/p2p/HandleWebsocket.go:249-260`. The clamp only enforces `pingPeriod < pongWait`. An override with zero `pongWait` yields `pingPeriod = 0`, and `time.NewTicker(0)` panics inside `handleClientMessages` — a goroutine panic, so it takes the process down rather than being caught by echo's Recover. Test-only field today, but guarding bad overrides is what the function is for. Clamp to a positive floor.

icellan (commented):
Re-reviewed at `cbf707a25` (fetched from origin). Verified locally: `go build ./...`, `go vet ./services/p2p/` clean, `go test -race -count=1 ./services/p2p/...` green.

All four points from my last round are genuinely closed in the code: per-context trusted-CIDR overrides for exactly the three contexts where `p2p_httpAddress` is non-loopback (`operator`, `docker.m`, `docker.ss.teranode1` — I checked, that's the complete set; the e2e/host compose contexts run p2p and asset in one container, so loopback covers them), `p2p_websocket_max_connections_per_source` with `-1` to disable, the time-throttled cap warning (CAS is correct), and `websocketTimeouts()` now floors every duration so `time.NewTicker(0)` can't panic. Three residuals, none blocking the lifecycle work.

**P1 — the `.operator` trusted CIDRs make the caps inert in the shipped k8s topology**

`settings.conf:1025` trusts `10.0.0.0/8|172.16.0.0/12|192.168.0.0/16`, but `deploy/kubernetes/teranode/teranode-cr.yaml:112-122` exposes `/p2p-ws` and `/p2p-wss` through a Traefik ingress. Traefik is L7: the backend connection always originates from the Traefik pod's IP, which is in the pod CIDR, i.e. inside RFC1918. So every external websocket client takes the trusted bypass in `acquire` (`HandleWebsocket.go:373-391`) — `total` stays 0, `perSource` stays empty, nothing ever gets a 503. An attacker opening 5,000 connections through the ingress hits neither cap; goroutines and fds grow exactly as they did before the PR. The one signal is `bypassWarned`, which only fires once `trustedLive > maxTotal` — 1,000 concurrent bypassed connections, i.e. after the exhaustion the cap exists to prevent.

I asked for the `.operator` override, so this is on me too, but RFC1918 is too wide a lever for it. Narrow `.operator` to the asset-service pod/service CIDR (or the specific bridge source), and treat "fronted by an L7 ingress" as a topology the source-IP cap simply cannot police — worth saying so explicitly in the longdesc rather than implying the cap protects the k8s deployment.

**P2 — the inbound frame budget doesn't cover control frames**

`HandleWebsocket.go:642-661`. gorilla handles ping/pong/close inside `advanceFrame` (`conn.go:957-968`) and loops; `ReadMessage` never returns them. So `wsMaxInboundFrames` budgets data frames only:

- A client streaming ping frames at line rate keeps the read goroutine and the default ping handler's `WriteControl(PongMessage, ...)` busy for the whole 60s `pongWait` window, reflecting a frame back for each one, and `inbound` never increments.
- Unsolicited pong frames hit `SetPongHandler` (`:638-640`), which refreshes the read deadline. A client that sends nothing but pongs holds its slot indefinitely without ever touching the budget — the idle timeout is trivially defeated.

Neither is fatal given the per-source cap, but the comment at `:212-214` claims an admitted client "must not be able to spend server read-path budget without bound", and as written it can. A `SetPingHandler` that counts against the same budget (and a pong handler that does too) closes it. This is the `ChiR12` residual from the earlier review, still open.

**P3 — `bypassWarned` is a one-shot for the process lifetime**

`HandleWebsocket.go:316` / `:379-382`. Same failure mode you just fixed for the rejection warning: one transient spike during startup burns the flag and a genuine "the caps are not binding" condition weeks later is silent, with no metric behind it either. Reuse the `wsCapWarnInterval` throttle here instead of the bool.

icellan:
Re-reviewed at cbf707a25 against `origin/main` (merge-base 8bba344cb). Tests re-run green with `-race`.

The Go hardening itself is sound: slot-before-upgrade admission, one teardown path that joins both pumps, write deadlines on every write, read deadline refreshed by pongs, eviction that cancels and closes. Two things undercut it.

### P1 — `settings.conf:1025-1027`: RFC1918 trust makes the caps inert in the shipped k8s/docker topologies

`p2p_websocket_trusted_source_cidrs.operator` / `.docker.m` / `.docker.ss.teranode1` trust `10.0.0.0/8|172.16.0.0/12|192.168.0.0/16`. `deploy/kubernetes/teranode/teranode-cr.yaml:113-121` exposes `/p2p-ws` through a Traefik ingress (`wsIngress`/`wssIngress`), so every externally originated websocket arrives from the Traefik pod's RFC1918 address, takes the trusted bypass, and lands in `trustedLive` instead of counting against `maxTotal`/`maxPerSource`. In the operator context concurrent connections are therefore unbounded — the DoS this PR is closing.

This is also the opposite of what the PR body states ("the default is loopback only, in all contexts, by design ... trusting a broader range (e.g. RFC1918) would make the caps inert there"). The body wasn't updated after cbf707a25 added the overrides. Either narrow those entries to the actual asset-service pod/container CIDR, or drop them and let the bridge live inside its per-source budget — which is what the body argues for.

### P1 — `services/p2p/HandleWebsocket.go:638-661`: control frames are not counted against the inbound budget

Only `SetPongHandler` is installed; there is no `SetPingHandler`. gorilla consumes ping/pong inside `advanceFrame` and never surfaces them from `ReadMessage`, so `inbound++` in the read pump sees data frames only. An admitted client streaming ping frames at line rate gets a `WriteControl(PongMessage, ...)` reply per ping from the read goroutine, contending on the connection write lock with the notification writer, and never trips `wsMaxInboundFrames`. A pong flood is worse in one respect: every pong runs `SetReadDeadline(now+pongWait)`, so the idle timeout never fires.

The PR body claims the opposite ("The budget covers control frames too ... the ping handler ... and the pong handler both count against the same budget") and lists a test for it ("ping/pong control-frame floods all close the connection"). `TestHandleWebSocket_InboundFrameLimits` only has the oversized-message and data-frame subtests. Wrap both handlers so they increment a shared counter and close on overrun, and add the missing subtests.

### P2 — `settings/settings.go:562`: the trusted bypass can't be turned off

gocore's `GetMulti` returns the default whenever the configured value is empty, so `p2p_websocket_trusted_source_cidrs =` yields `127.0.0.1/32|::1/128`, not an empty list. An operator with a reverse proxy terminating on the same box — the exact case the `WebSocketMaxConnectionsPerSource` longdesc warns about — can only express "no trusted sources" by faking it with an unroutable CIDR. Document the workaround or accept a sentinel.

### P3

- `HandleWebsocket.go:315,379` — `bypassWarned` is a one-shot bool and only fires when `trustedLive > maxTotal`. Behind a trusted proxy with under 1000 live connections the "caps are not binding" condition is never surfaced, and once logged it never repeats. The body says this warning shares the per-minute rejection throttle; it doesn't.
- `docs/topics/services/p2p.md:313` — "(default loopback)" omits the RFC1918 context overrides that the settings reference table does mention.
- `HandleWebsocket.go:149-181` — `broadcast` still serialises 1s-timeout sends in batches of 256 behind `wg.Wait()`. At the default cap of 1000, a set of admitted non-draining clients adds ~4s to every notification before eviction clears them. Self-limiting now that eviction closes, but worth a line next to the memory-ceiling docs.
- `docs/topics/services/p2p.md:317` describes registration before the initial `node_status` queue; the code queues first, then registers, and that ordering is the whole point of the code.

icellan (commented):
Re-reviewed at `2929e80b4`. `go vet ./services/p2p/` clean, websocket/limiter/CORS/timeout tests green under `-race` (8.2s).

`6eb9c9d8b` closes everything from my last round, verified against the code rather than the commit message:

- Control-frame budget is real now. `SetPingHandler` exists at `HandleWebsocket.go:695` and both handlers share the `overBudget()` counter, mutated only on the read goroutine (gorilla invokes the handlers from inside `ReadMessage`), so the lock-free comment is accurate. Handler errors propagate out of `ReadMessage` and drop the read pump, so a ping or pong flood now closes the connection instead of buying unmetered pong reflection and an indefinitely refreshed read deadline.
- Trusted CIDRs are back to `127.0.0.1/32|::1/128` with no per-context overrides, so the RFC1918 bypass that made both caps inert under the operator and docker topologies is gone.
- `wsTrustNoneSentinel` gives a way to express "trust nobody", which `gocore.GetMulti`'s empty-means-default behaviour otherwise made impossible.
- The bypass warning is throttled by `wsCapWarnInterval` and its threshold is now `maxPerSource` rather than `maxTotal` (`:355-357`), so a fronting proxy surfaces at ~50 bypassed connections instead of after the 1000-connection exhaustion the cap exists to prevent.

One thing left, non-blocking:

**P2 — `HandleWebsocket.go:427`: per-source cap keys on the exact host, so IPv6 clients aren't bound by it**

`l.perSource[host]` uses the literal address from `SplitHostPort`. A routed /64 is the standard IPv6 allocation for a residential line or a cheap VPS, so a single client can present 2^64 distinct source addresses, each getting its own fresh per-source budget. The per-source cap contributes nothing against an IPv6 source and only the global cap of 1000 binds — which is the cap the per-source limit exists to keep one host from eating. IPv4 is unaffected.

Bucketing IPv6 by /64 before the map lookup (`ip.Mask(net.CIDRMask(64, 128))`) restores the intent. Worth a line in the `WebSocketMaxConnectionsPerSource` longdesc either way, since "per source" currently means something different depending on address family.

icellan (approved):

Issue #1306, opened 2026-07-20
Asset server panics (HTTP 500, empty body) reconstructing a partially-spent transaction from the UTXO set


Ladybug7117:
## Summary

When the Asset server serves a transaction whose full `.tx` blob has been
pruned to a UTXO-set snapshot (`.outputs` blob present, `.tx` absent), and the
transaction is **partially spent**, reconstruction produces a `*bt.Tx` whose
output vector contains **nil `*bt.Output` entries at the spent indices**. The
subsequent size/serialize of that tx dereferences the nil output and panics;
the HTTP middleware recovers it into a **500 with an empty body**.

So instead of an honest "this transaction is no longer fully retained," the
caller gets an opaque crash with no diagnostic — and the same reconstruction
failure blocks any path that must re-materialize the transaction (e.g.
validating a spend that references it).

## Environment

- Teranode Asset server (`services/asset`), checked against `v0.15.5`.
- `go-bt/v2` `v2.6.5` (the version pinned by the release line). The nil-pointer
  panic itself is in go-bt `(*Output).Size()` / `(*Output).appendTo()`; see the
  companion go-bt report. This issue tracks the teranode-side cause (why a nil
  `*Output` is produced) and the operational impact.

## Root cause / code path

All references are to the public tree at `v0.15.5`; line numbers are
approximate (look for the symbols).

1. `GetTransaction` / `GetTransactionMeta`
   (`services/asset/repository/repository.go`) call `UtxoStore.Get` and then
   size/marshal the returned tx.
2. In the aerospike store, `getExternalTransaction`
   (`stores/utxo/aerospike/get.go`) fetches the external `.tx` blob; when the
   tx has been pruned, that fetch returns `ErrNotFound` and the code falls back
   to the `.outputs` (UTXO-set) blob.
3. Reconstruction pads the output vector to `maxIndex+1` (`PadUTXOsWithNil`)
   and then fills **only the indices still present in the UTXO set** (the
   unspent outputs) — the fill loop at roughly `get.go:1529-1537`. Spent
   indices are left as **nil `*bt.Output`**.
4. Because the vector is sized to `maxIndex+1`, whenever the **highest-index
   output is still unspent**, the nil holes at lower spent indices remain
   in-bounds and are dereferenced downstream by go-bt.

Notably, teranode **already** backfills nil outputs with an empty-script
placeholder in the *input-decoration* path (around `get.go:1399`) — the
[...]

PR #1383, opened 2026-07-31
fix(asset): recover per-goroutine in errgroup fan-outs, bound GetUTXOsByTxID

icellan:
Refs #1306 — one of four PRs from that investigation; see the issue for the full set.

Echo's `middleware.Recover` only wraps the request goroutine, not the ones `errgroup` spawns. `GetTransactions` and `GetUTXOsByTxID` both fanned out bare goroutines that call into the store, so a single panic took the whole asset process down instead of returning a status.

Both endpoints take their keys from caller input — `GetTransactions` reads txids straight from the request body — so the panicking record is attacker-chosen, not incidental. `heavyMW()` is rate limiting, not auth, and peer auth is operator opt-in.

`GetUTXOsByTxID` also fanned out one goroutine per output with no limit, so a single request for a large-output transaction could exhaust the process on its own. Capped at `utxosFanoutLimit`, matching `GetUTXOs`.

The guards follow the existing pattern in `GetUTXOs.go:127-137`, which already documents this exact hazard.

## Why this is a P0 on its own

There is a reachable panic behind these endpoints today: a transaction whose `.tx` external blob is absent but whose `.outputs` blob is present reconstructs with nil `*bt.Output` holes, and every go-bt serialization entry point dereferences them. That happens on any node bootstrapped from a UTXO-set snapshot. One known txid was enough to restart the asset process on demand, repeatedly, unauthenticated.

This PR does not fix that reconstruction — see the companion PR for #1306. It stops the panic from taking the process down, which is worth landing independently of the root cause.

## Testing

`services/asset/httpimpl/goroutine_panic_test.go`. Without the fix these tests do not fail — they abort the test binary, which is precisely the production behaviour being asserted against.

`go test -race ./services/asset/...` green.


liam (commented):
**Verdict: APPROVE-WITH-NITS**

The core fix is correct, minimal and well-targeted. Confirmed the premise against `golang.org/x/sync` errgroup: it deliberately does **not** propagate child panics (`errgroup.go:81-89`), so a bare `g.Go` panic genuinely crashes the process and per-goroutine recover is the right fix. `Group.done()` releases the semaphore token via defer, so a recovered goroutine that returns normally releases its slot - no `Wait` hang.

- **Recover → error → 500: correct.** Both fan-outs use `func() (retErr error)` with the named return set inside the deferred `recover()`; errgroup captures it and the handler emits a proper 500. Matches the merged `GetUTXOs.go:127-137` reference.
- **Fan-out cap: correct.** `GetUTXOsByTXID.go:172` adds `util.SafeSetLimit(h.logger, g, utxosFanoutLimit)` (=1024), same shape and constant as `GetUTXOs.go:117`. `SafeSetLimit` clamps `<1` to `NumCPU()`, so no zero-capacity-semaphore deadlock. Base `GetUTXOsByTXID` had no limit, so the unbounded-per-output claim is accurate.
- **No loop-variable capture bug:** `GetUTXOsByTXID.go:178` shadows with `safeI, safeOutput := i, output`; `GetTransactions.go` declares a fresh `hash` per iteration.

**Nits (non-blocking):**
1. `GetUTXOsByTXID.go:221` double-wraps the error: the recovered `retErr` is already an `*echo.HTTPError(500)`, and `g.Wait()`'s result is then re-wrapped into a second `echo.NewHTTPError(500, ...)` that embeds the inner echo error's JSON as a string. Net status is still 500, but it is inconsistent with `GetTransactions.go` / `GetUTXOs.go`, which return the recovered error directly. Consider returning `err` straight through.
2. The "bound" half is untested - the tests prove panic containment (they abort the binary without the fix) but nothing asserts `utxosFanoutLimit` actually caps concurrency, so that DoS-mitigation half could silently regress. Consider a case with `> utxosFanoutLimit` outputs.

No correctness or security defects found.

liam (approved):

ctnguyen (changes requested):
Review at commit [44e6cb5](https://github.com/icellan/teranode/commit/44e6cb5a5a3c8c1f996df97af5a77baa6f070a82)

This is the right instinct and a nicely surgical diff. Splitting the crash containment out from the reconstruction fix in [#1384](https://github.com/bsv-blockchain/teranode/pull/1384) is a good call — it lands the availability win without waiting on the harder correctness argument. The mechanics are clean too: the named-return `retErr` form is the correct way to convert a panic into an errgroup error, the per-iteration capture is genuinely safe (`hash` is redeclared inside the `for` body in `GetTransactions.go:140`, `safeI`/`safeOutput` are copied in `GetUTXOsByTXID.go:184`), and the response body deliberately omits the panic value while the log keeps it — which matches the property `GetUTXOs.go:79-81` already documents. The test comment being upfront that these tests abort the binary rather than fail cleanly is the honest thing to write, and it's the strongest signal available for this class of bug. My main concern is that the same panic still has open routes out of the process on this very service, which undercuts the P0 claim the PR is making for itself.

## Blocking issues

### ChiR1 — the same reachable panic still kills the asset process through the other fan-outs

**Problem:** The only non-test `recover()` calls anywhere in `services/asset` are the three added here — `GetUTXOs.go:133`, `GetTransactions.go:167` and `GetUTXOsByTXID.go:191`. Every remaining request-path and streaming fan-out is still bare: `services/asset/repository/GetLegacyBlock.go:70`, `:322`, `:424`, `:562`, and `services/asset/repository/GetSubtreeData.go:261`. (`Server.go:263` and `:287` are also bare, but they are one-shot lifecycle goroutines that start the HTTP and centrifuge servers — not on any per-request path, so they are out of scope here.) Three of the five nest on one path: `:70` (and `:261`) run `writeTransactionsViaSubtreeStoreStreaming`, whose `:322` scheduler goroutine wraps `scheduleSubtreeChunkFetches`, which in turn owns the `:424` chunk-worker fan-out. A recover is needed at each level, because a panic in an inner errgroup goroutine is not observable by the outer one. The serialization hazard sits at the outermost level — `writeChunkToWriter` writes store-sourced transactions directly:

```go
// services/asset/repository/GetLegacyBlock.go:524
if _, err := chunkMetaSlice[i].Tx.WriteTo(w); err != nil {
```

`chunkMetaSlice` comes from `repo.getTxs` → the UTXO store, i.e. the exact reconstruction that leaves nil `*bt.Output` holes (`stores/utxo/aerospike/get.go:531` and `:1685` both `make([]*bt.Output, n)` and fill only the indices that are present). `Tx.WriteTo` dereferences every element, so this is the same nil-deref as the one this PR is guarding against in `GetTransaction`. Worse, both of these goroutines are detached background streamers — `GetLegacyBlockReader` and `GetSubtreeDataReader` return the pipe reader and let the goroutine outlive the handler — so `middleware.Recover` (`http.go:172`) can't reach them even in principle. Reachable unauthenticated via `GET /api/v1/block_legacy/:hash` (`http.go:376`), `GET /api/v1/subtree_data/:hash` (`http.go:347`) and `GET /rest/block/:hash.bin` (`http.go:319` — note this one hangs off `e.Group("/rest")` at `http.go:318`, not the configurable `asset_apiPrefix` group); the mining-candidate path (`mining_candidate_legacy_block.go:77`) reaches the same writer.

**Why it matters:** The PR's stated value is "It stops the panic from taking the process down, which is worth landing independently of the root cause." On a snapshot-seeded node — the condition established in the [#1306](https://github.com/bsv-blockchain/teranode/issues/1306) thread as permanent for every pre-snapshot tx — an attacker who cannot reach `POST /subtree/:hash/txs` can still restart the process with a plain `GET /block_legacy/:hash`, and doesn't even need to know a specific txid. Merging this as-is risks the P0 being marked contained when it isn't. `heavyMW()` doesn't help here: one request is enough.

**Fix:** Since the block is now written verbatim three times, lift it into a helper and apply it at every fan-out site, including the two detached streamers:

```go
// util/recover.go
func RecoverToError(logger ulogger.Logger, retErr *error, label string, onPanic func(r any) error) func() {
	return func() {
		if r := recover(); r != nil {
			logger.Errorf("[%s] recovered panic: %v\n%s", label, r, debug.Stack())
			*retErr = onPanic(r)
		}
	}
}
```

For the detached streamers there is no HTTP status to return, so the recover must instead close the pipe with an error so the consumer sees a failed stream rather than a truncated body. `:322` and `:424` only need the plain error-returning form — at `:322` a recover composes with the existing `defer close(resultsChan)`, which already unblocks the consumer's `range` loop during unwinding:

```go
// services/asset/repository/GetLegacyBlock.go:70
// note: the pipe reader is already bound to `r` here, so name the panic value
g.Go(func() (err error) {
	defer func() {
		if p := recover(); p != nil {
			repo.logger.Errorf("[GetLegacyBlockReader][%s] recovered panic: %v\n%s", block.Hash(), p, debug.Stack())
			err = errors.NewProcessingError("[GetLegacyBlockReader][%s] internal error streaming block", block.Hash())
			_ = w.CloseWithError(err)
			_ = r.CloseWithError(err)
		}
	}()
	...
```

## Non-blocking issues

### ChiR2 — `/utxos/:hash/json` has no heavy-tier limiter, so the new cap only mirrors half of `GetUTXOs`

**Problem:** The new comment says "Mirrors GetUTXOs, which caps at the same limit", but `GetUTXOs`' own route comment says the cap is only half the story:

```go
// services/asset/httpimpl/http.go:402-410
// Routed through heavyMW() so a single request (up to 1024-way Aerospike fan-out) is
// gated by the same tiered rate limiter that protects /subtree/:hash/txs ...
apiGroup.POST("/utxos", h.GetUTXOs(BINARY_STREAM), heavyMW()...)
```

`GetUTXOsByTxID` is registered without it, at `http.go:400`:

```go
apiGroup.GET("/utxos/:hash/json", h.GetUTXOsByTxID(JSON))
```

So it falls back to the global limiter only. Defaults are `asset_httpRateLimit` = 1024 req/s per IP and `asset_httpHeavyRateLimit` = 10 (`settings/asset_settings.go:29-30`) — a two-order-of-magnitude difference for what is now, by the PR's own framing, a heavy fan-out endpoint.

**Why it matters:** `SafeSetLimit` bounds one request at 1024 in-flight lookups; nothing bounds the number of concurrent requests. At the default global rate a single IP can hold on the order of 1024 × 1024 goroutines parked on `semGetUtxo` (`repository.go:912`), each with its stack, plus a fully materialised transaction per request. That is the same "exhaust the process" outcome the PR is trying to close, just moved from the per-request axis to the cross-request one — and it's a `GET`, so it is reachable from a browser or a crawler with no request body at all.

**Fix:** Register the route through the heavy-tier limiter like its sibling bulk endpoint.

```go
apiGroup.GET("/utxos/:hash/json", h.GetUTXOsByTxID(JSON), heavyMW()...)
```

### ChiR3 — the fan-out cap bounds goroutines, not the per-request heap the description attributes to it

**Problem:** The description says "a single request for a large-output transaction could exhaust the process on its own. Capped at `utxosFanoutLimit`." The cap only applies to concurrency. The dominant per-request cost is untouched: `GetUTXOsByTXID.go:149` loads the full extended tx bytes, `:160` parses a second full copy via `bt.NewTxFromBytes`, and `:180` allocates `make([]*UTXOItem, len(tx.Outputs))` with one heap-allocated `UTXOItem` per output — all before the first goroutine starts, and all held until the response finishes streaming.

**Why it matters:** For a transaction with a very large output set that is still tens to hundreds of MB resident per in-flight request, and it scales linearly with the number of concurrent requests — which ChiR2 leaves unbounded. Landing the goroutine cap is the right first step, but the comment currently reads as though the resource question is settled, which will discourage anyone from revisiting it.

**Fix:** Narrow the comment to what the cap actually does, and note the residual:

```go
// Bound the concurrent per-output store lookups. Note this does not bound the
// per-request heap: the full tx is materialised twice (raw + parsed) and one
// UTXOItem is allocated per output before the fan-out starts.
util.SafeSetLimit(h.logger, g, utxosFanoutLimit)
```

If a hard bound is wanted later, rejecting or paginating above an output-count threshold is the cheap option.

### ChiR4 — the recovered error is re-wrapped at `g.Wait()`, so the client sees a nested 500

**Problem:** The recover produces an `*echo.HTTPError` with a 500, and `GetUTXOsByTXID.go:237-239` wraps it in a second one:

```go
if err = g.Wait(); err != nil {
	return echo.NewHTTPError(http.StatusInternalServerError, errors.NewProcessingError("[Asset_http][%s] error getting utxos", hash.String(), err).Error())
}
```

`customHTTPErrorHandler` (`http.go:693-699`) returns `he.Message` verbatim to the client, so the body ends up carrying the inner `code=500, message=...` string inside the outer message. The same wrap also turns a `context.Canceled` from a client disconnect into a 500. `GetTransactions.go:193-196` gets this right by returning the errgroup error unchanged.

**Why it matters:** Only cosmetic for the status code, but it makes the JSON body awkward for clients to parse and inflates the disconnect-driven 500 rate in the access-log metrics, which is the signal you'd want clean when watching for the panic path firing in production.

**Fix:** Pass an `*echo.HTTPError` through unchanged and only wrap what isn't one, matching `GetTransactions`.

```go
if err = g.Wait(); err != nil {
	var he *echo.HTTPError
	if errors.As(err, &he) {
		return err
	}

	return echo.NewHTTPError(http.StatusInternalServerError, errors.NewProcessingError("[Asset_http][%s] error getting utxos", hash.String(), err).Error())
}
```

### ChiR5 — the tests don't cover the two properties that are easiest to regress

**Problem:** Three gaps in `goroutine_panic_test.go`:

1. Neither test asserts that the panic value is absent from the response body. That's the documented security property (`GetUTXOs.go:79-81`: "the panic is logged but the response body does not include the panic value") and the one a future refactor would silently break by switching to `errors.NewProcessingError("...", r)`.
2. Nothing exercises the new `SafeSetLimit` on `GetUTXOsByTxID` — both fixtures use `testTX1RawBytes`, which is a one-output transaction, so the fan-out is width 1 and neither the cap nor the "one goroutine panics, its siblings still complete" case is covered.
3. `echoContext.SetPath("/utxos/txid/:hash")` names a route that doesn't exist; the real one is `/utxos/:hash/json` (`http.go:400`). Harmless to the assertion since `SetParamValues` does the real work, but it will send the next reader to the wrong place.

**Why it matters:** Panic-recovery code is exactly the kind of thing that gets "cleaned up" later by someone who doesn't know why the message is deliberately vague, and the fan-out bound has no coverage at all right now.

**Fix:** Assert the negative on the body, and add a multi-output fixture so one panicking index sits alongside successful ones:

```go
require.NotContains(t, httpErr.Message, "nil pointer dereference")
require.NotContains(t, httpErr.Message, "runtime error")
```

## Recap

| ChiR1 | Other asset fan-outs still unguarded | YES | 82% |
| ChiR2 | Missing heavyMW on /utxos/:hash/json | | 58% |
| ChiR3 | Cap bounds goroutines, not heap | | 42% |
| ChiR4 | Double-wrapped 500 at g.Wait | | 26% |
| ChiR5 | Test gaps: leak, fan-out width, path | | 18% |

icellan:
@ctnguyen — addressed in `63dc4a630`. Verified each finding against the code before acting; all five hold.

**ChiR1** — fixed. The reading is correct, including that an outer recover cannot observe an inner errgroup goroutine's panic.

- `util.RecoverToError` logs the panic value, converts it to an error naming only the fan-out site, and runs optional panic-only cleanup with that error.
- Guards at `GetLegacyBlock.go:70/322/424/562` and `GetSubtreeData.go:261`. The two detached streamers also fail their pipe, so a panic surfaces as a failed stream rather than leaving the consumer blocked. The subtree-data streamer additionally calls `storer.Abort`, so a panic mid-generation cannot finalise a partial subtreeData, and counts the failure under `write_failed` instead of dropping it.
- At `:322` the recover is registered *after* `defer close(resultsChan)` so it runs first: `err` is set before the channel closes, and the consumer's range exits into `g.Wait()`.
- Tests inject a panic at four of the five frames, each verified to abort the test binary without its guard. `fetchSubtreeChunk` (`:424`) has no injection point of its own today — a `BatchDecorate` panic lands in the `getTxs` worker below it — so its guard is there because a panic in that frame would otherwise be unrecoverable, not because a test drives it. Flagging that rather than claiming coverage I don't have.
- The three `httpimpl` guards keep their inline form: they must return an `*echo.HTTPError` carrying the status, which doesn't belong in `util`, and two of the three are already merged on main.
- I left `debug.Stack()` out of the helper's log — `docs/references/codingConventions.md` requires single-line log messages, and the merged `GetUTXOs.go` guard logs the panic value alone. Say the word if you'd rather have stacks; that's a repo-wide call, not one for this PR.

**ChiR2** — not in this PR. Moving that route from the 1024 req/s global tier to the 10 req/s heavy tier is a visible behaviour change for existing callers of the GET, and an ops decision with a blast radius wider than a panic-containment fix. The cross-request axis you describe is real and unbounded — worth its own change.

**ChiR3** — comment narrowed to what the cap actually bounds, and it now states the residual: the tx is materialised twice and one `UTXOItem` is allocated per output before the fan-out starts, and bounding that needs an output-count threshold this doesn't add.

**ChiR4** — fixed in `e0a18930a`, slightly differently from the suggestion. Rather than an `errors.As` branch at `g.Wait()`, the one remaining raw-error path in the goroutine (`util.UTXOHash`) now builds its own `echo.HTTPError`, so every error out of the fan-out carries its own status and `g.Wait()` returns it unchanged — same shape as `GetTransactions`. Passing through unconditionally without that would have degraded the `UTXOHash` case to echo's generic "Internal Server Error" body.

**ChiR5** — all three:

1. Both panic tests now assert the panic value is absent from `httpErr.Message`.
2. The `GetUTXOsByTxID` fixture is 8 outputs panicking on vout 3, asserting the other 7 complete. Separately, `TestGetUTXOsByTXID_FanOutIsBounded` drives 1088 outputs and asserts in-flight lookups never exceed `utxosFanoutLimit` — mutation-verified (1088 in flight with `SafeSetLimit` removed). A fixed sleep did not catch it: testify's own lock staggers goroutine entry, so the probe releases itself once every output is in flight.
3. Route path corrected to `/utxos/:hash/json`.

Copilot's loop-variable finding on `GetTransactions.go:173` stays a false positive — `hash` is declared inside the loop body.

`go vet` and `go test -race` green across `./services/asset/...` and `./util/`. Re-review please.

ctnguyen (changes requested):
Review at commit [63dc4a6](https://github.com/icellan/teranode/commit/63dc4a6305f94f0c50968ee52518c4a98b1ef1b8)

This is a strong second pass. The blocking concern from last time is genuinely closed: every repository fan-out the previous review named now recovers per goroutine, the two detached streamers fail their pipe so a panic surfaces as a broken stream instead of a hung consumer, and the subtree-data streamer additionally aborts the temp blob and counts the failure — exactly the right shape given #1377. The `util.RecoverToError` helper is a clean way to lift the pattern out of five sites, the defer ordering at `GetSubtreeData.go:262-278` is thought through (RecoverToError runs first, then quorum release, then the two permits), and the new tests are better than most — the bound test is mutation-verified and drives 1088 outputs, and the author is honest that `fetchSubtreeChunk` (`:440`) has no injection point of its own rather than claiming coverage they don't have. The remaining issue is that the new containment on the legacy-block path swaps one failure mode for a quieter, more persistent one: it recovers the panic but leaks the file-store read permit held across that frame.

## Status of previous points

ChiR2 was declined rather than fixed — the author's reasoning (moving the GET off the 1024 req/s global tier onto the 10 req/s heavy tier is a visible behaviour change for existing callers and an ops decision) is sound, and the cross-request axis it describes was non-blocking before and stays out of scope for a containment PR. ChiR1's guards are all present; the only gap is that the `fetchSubtreeChunk` (`GetLegacyBlock.go:440`) guard has no test driving it, which the author discloses.

| ChiR1 | Other asset fan-outs unguarded | YES | FIXED |
| ChiR2 | No heavyMW on /utxos/:hash/json | | NOT FIXED |
| ChiR3 | Cap bounds goroutines, not heap | | FIXED |
| ChiR4 | Double-wrapped 500 at g.Wait | | FIXED |
| ChiR5 | Test gaps: leak, width, path | | FIXED |

## Blocking issues

### ChiR6 — the new legacy-block guard leaks the file-store read permit on panic

**Problem:** In the detached streamer at `GetLegacyBlock.go:70`, the subtree-data reader is opened inside the loop and released only by the three explicit `subtreeDataReader.Close()` calls on the normal/error paths (`:147`, `:166`, `:180`):

```go
subtreeDataReader, err = repo.SubtreeStore.GetIoReader(ctx, subtreeHash[:], fileformat.FileTypeSubtreeData)  // :122
...
bufferedReader := bufio.NewReaderSize(subtreeDataReader, 32*1024)
for {
    tx := &bt.Tx{}
    if _, err = tx.ReadFromWithArena(bufferedReader, arena); err != nil { ... }  // :138
    ...
    if _, err = tx.WriteTo(w); err != nil { ... }                                // :163
}
```

The reader is a `*semaphoreReadCloser` from `stores/blob/file/file.go`, whose `Close()` is the *only* thing that returns the process-global `readSemaphore` permit (`file.go:187-193`, `defaultReadLimit = 768`). The new guard's `onPanic` closure (`:75-78`) closes only the pipe ends `w` and `r`; it cannot close `subtreeDataReader`, which is declared later at `:104-107` and is out of the closure's scope, and there is no `defer …Close()` for it. So any panic raised after the reader is opened at `:122` and before the loop reaches a `Close` — a store-driver fault in `Read`, or a panic inside the `ReadFromWithArena`/`WriteTo` go-bt serialization the PR itself names — is recovered and converted to an error, but the read permit is never given back.

**Why it matters:** This is the one path where the containment is a net regression. Before the guard, a panic here crashed the process and a restart reset the in-memory semaphore; now the process survives but bleeds a global permit each time. The permit is shared by *every* blob read in the asset server, and `acquireReadPermit` waits 25 s before returning `ServiceUnavailableError` (`file.go:350-367`); enough recovered panics (768, or fewer under concurrency) and every blob read — legacy blocks, subtree data, transactions — stalls 25 s then 503s, for the life of the process. The leak is not tied to the #1306 nil-`*bt.Output` trigger specifically: that trigger is contained at blob-generation time by the `GetSubtreeData` guard, so a *finalised* subtreeData blob feeding this loop shouldn't carry holes. It is tied to *any* panic that lands in this frame — and the guard exists precisely because the team can't enumerate those panics, so shipping one that exhausts the server's read capacity undercuts the containment claim rather than delivering it.

**Fix:** Release the reader on the panic path with a deferred close, and null it out after each existing explicit close so the defer only fires on the path that still holds an open reader:

```go
var (
    subtreeDataExists bool
    subtreeDataReader io.ReadCloser
)
defer func() {
    if subtreeDataReader != nil {
        _ = subtreeDataReader.Close() // panic path only: releases the file-store read permit
    }
}()
// ...and after each existing explicit Close (:147, :166, :180):
_ = subtreeDataReader.Close()
subtreeDataReader = nil
```

`semaphoreReadCloser.Close` already guarantees the permit is released exactly once (`sync.Once` around `releaseReadPermit`), so correctness would not break if the deferred close also ran after an explicit one — but the `sync.Once` guards only the permit release, not the wrapped `ReadCloser.Close()`, which is invoked on every call and returns `os.ErrClosed` on the second (the call sites discard it). Nulling the reference after each explicit close avoids that redundant second close entirely, so the defer fires only on the panic/early-return path where the reader is genuinely still open.

## Non-blocking issues

### ChiR7 — RecoverToError leaks a trailing `error` argument into the returned error

**Problem:** The helper builds the returned error with `errors.NewProcessingError("internal error in "+format, args...)` (`util/recover.go:45`). `errors.New` treats a trailing `error`/`*Error` param specially: it extracts it as the wrapped error and renders it into the message via `Error()`'s `" -> "` join (`errors/errors.go:356-380`). So a caller that ends its args with an error — e.g. `RecoverToError(logger, &err, nil, "batch %d", i, cause)` — gets `"internal error in batch <i> -> <cause.Error()>"`: the `%d` is still satisfied by `i` (no `%!d(MISSING)`), but `cause`'s message is folded into the client-facing error, defeating the helper's stated contract that it "names only the fan-out site, so no panic detail reaches a client." (A dangling verb like `%!s(MISSING)` only appears in the narrower case where stripping the trailing error leaves a verb with no surviving arg, e.g. `"batch %d %s", i, cause`.) None of the five current call sites pass a trailing error — they pass only chainhash strings and ints — so this is latent, not an active bug, but this is new shared `util` meant to be reused.

**Why it matters:** The helper's whole value is that callers can trust it to keep internal detail out of the returned error. A reuse that quietly inverts that property is the kind of thing that gets copy-pasted onto a client-facing path later. Note the naive defensive rewrite does not fix it: `fmt.Sprintf(format, args...)` with the same call leaks `cause` back through fmt's `%!(EXTRA …)` rendering.

**Fix:** Make the contract explicit and enforce it rather than pre-formatting. Document that `args` must be format parameters only, and drop/reject any trailing error before it reaches either `fmt` or `errors.New`, with a test pinning that a trailing error neither leaks nor mangles the message:

```go
// contract: args are format params for `format` only — never a trailing error/panic value.
if n := len(args); n > 0 {
    if _, isErr := args[n-1].(error); isErr {
        args = args[:n-1] // keep it out of the client-facing message; it stays in the log via %v below
    }
}
```

### ChiR8 — a recovered panic gives an operator almost nothing to localise it

**Problem:** Because the error is now constructed inside `util.RecoverToError`, `runtime.Caller(2)` (`errors/errors.go:389`) records `util/recover.go` as the file/line/function on *every* recovered-panic error, regardless of which of the five sites — or which underlying frame — actually panicked. Combined with the deliberate omission of `debug.Stack()`, the total information about a recovered panic is the panic value plus the coarse fan-out label (e.g. `getTxs batch at 0`). The label disambiguates the guard site but not where inside it the panic came from.

**Why it matters:** This is a path the PR itself frames as a P0, and the error is the artifact an operator will pivot on when the panic path starts firing in production. The single-line-log convention is a fair reason to keep the log terse, so this isn't a correctness defect — but the Caller misattribution means the error metadata actively points at the wrong file, so without a stack there is no way to narrow a recovered panic beyond the label.

**Fix:** At minimum note in the helper doc that the error's location metadata points at `RecoverToError`, not the panic site. If diagnosis on this path matters, log a bounded stack at recovery time (the convention constrains log *lines*, and a stack can be a separate, clearly-marked field or a debug-level line) so the panic frame is recoverable without waiting for a repro.

### ChiR9 — `append(args, r)` can scribble on the caller's backing array

**Problem:** The log call at `util/recover.go:43` is `logger.Errorf("recovered panic in "+format+": %v", append(args, r)...)`. `append(args, r)` writes into `args`'s backing array in place whenever there is spare capacity. Every current call site passes discrete arguments, so `args` arrives with `len == cap` and `append` allocates a fresh slice — no aliasing today. But a future caller that passes a pre-built slice with spare capacity — `a := make([]any, 1, 4); a[0] = i; RecoverToError(logger, &err, nil, "batch %d", a...)` — would have the panic value `r` written to `a[1]` on the panic path, mutating the caller's array just past the length it handed in.

**Why it matters:** It is minor and latent, but it is a footgun baked into a new exported helper: a variadic parameter looks safe to reuse from the caller's side, and this makes it not so. Cheap to close now.

**Fix:** Build the log args on a fresh slice so the panic value never lands in the caller's backing array:

```go
logArgs := append(append(make([]any, 0, len(args)+1), args...), r)
logger.Errorf("recovered panic in "+format+": %v", logArgs...)
```

## Recap

| ChiR6 | Read-permit leak on panic | YES | 70% |
| ChiR7 | Trailing-error arg leaks internal detail | | 35% |
| ChiR8 | Recovered panic hard to localise | | 20% |
| ChiR9 | append aliases caller backing array | | 12% |

icellan:
@ctnguyen — ChiR6 through ChiR9 addressed in `163408519`. All four were valid; ChiR6 in particular was caused by the guard I added last round.

**ChiR6** — fixed. Verified the mechanism before touching it: `subtreeDataReader` is a `*semaphoreReadCloser`, its `Close()` is the only path to `releaseReadPermit` (`stores/blob/file/file.go:187-193`), 768 global permits, and `acquireReadPermit` waits 25s then returns `ServiceUnavailableError`. The `onPanic` closure closed only the pipe ends, so a panic in `ReadFromWithArena`/`WriteTo` skipped all three explicit `Close()` calls. A test injecting a panic in the deserialization frame confirmed it — panic recovered, reader never closed — and fails without the fix.

Fixed as suggested: a deferred backstop that closes the reader if one is still open, plus `subtreeDataReader = nil` after each explicit close so the defer never double-closes. Your read that this was a net regression is right — a crash resets the semaphore, a leak does not.

**ChiR7** — fixed. Confirmed at `errors/errors.go:356-371`; the test showed it concretely before the fix: `internal error in batch 7 -> PROCESSING (4): aerospike said no for record 0xdeadbeef`. A trailing `error` in `args` is now dropped before the message is built, the contract is documented, and a test pins both halves — the cause does not reach the message and the format is not left mangled.

**ChiR8** — you were right and I was wrong last round. `grep debug.Stack()` finds six existing recovered-panic sites logging `%v\n%s` (`aerospike/create.go:339`, `batch_completion.go:47`, `aerospike/spend.go:348`, `sql/sql.go:1834`, `SubtreeProcessor.go:890` and `:3088`), so the single-line log convention plainly does not govern recovered panics — the precedent settles it. The helper now logs the stack, and the doc records that the error's `runtime.Caller` metadata points at `RecoverToError` where it is constructed, not at the panicking frame.

**ChiR9** — fixed. Reproduced first: with `args := make([]any, 1, 4)` the panic value landed in `args[1]`. Log args are now built on a fresh slice.

**ChiR2** stays out of scope, as agreed.

Worth flagging for your assessment of this PR's framing: #1384 has since merged to main, so the reconstruction that made the panic reachable is fixed at source. This PR is defence-in-depth from here rather than closing a live P0 — happy to reword the description if you'd rather it said so explicitly.

`go vet` and `go test -race` green on `./services/asset/...` and `./util/`; CI green on `163408519` (test, 7 sequential shards, 3 smoketest shards). Each new test was confirmed failing before its fix. Re-review please.

ctnguyen (approved):
Review at commit [fc5f46b](https://github.com/icellan/teranode/commit/fc5f46ba4d15956f766b4437b7662aed0038af44)

This has converged. Every point from both prior passes is either fixed at this head or settled by agreement, and the fixes are the real thing rather than a coat of paint. ChiR6 — the read-permit leak that last round's own guard introduced — is closed cleanly: a deferred backstop closes the reader only when it's genuinely still open, with the reference nilled after each explicit close so the defer never double-closes. The three smaller edges on the shared helper (ChiR7/8/9) are all handled, each with a test that pins the exact property it protects. I re-walked the panic paths independently: the defer ordering holds, no read permit, quorum lock, or pipe end is left dangling on any unwind, and no panic value reaches a client. With #1384 merged, this is now defence-in-depth rather than a live-P0 fix — worth reflecting in the description as the author already offered, but not a blocker. Nice work; nothing new to raise.

## Status of previous points

ChiR2 was declined rather than fixed, and that call still holds: moving `/utxos/:hash/json` off the 1024 req/s global tier onto the 10 req/s heavy tier is a visible behaviour change for existing callers and an ops decision with a blast radius wider than a containment fix. The cross-request exhaustion axis it names is real and unbounded, but it was non-blocking before and stays out of scope here — worth its own change. Status remains NOT FIXED.

| ChiR1 | Other asset fan-outs unguarded | YES | FIXED |
| ChiR2 | No heavyMW on /utxos/:hash/json | | NOT FIXED |
| ChiR3 | Cap bounds goroutines, not heap | | FIXED |
| ChiR4 | Double-wrapped 500 at g.Wait | | FIXED |
| ChiR5 | Test gaps: leak, width, path | | FIXED |
| ChiR6 | Read-permit leak on panic | YES | FIXED |
| ChiR7 | Trailing-error arg leaks internal detail | | FIXED |
| ChiR8 | Recovered panic hard to localise | | FIXED |
| ChiR9 | append aliases caller backing array | | FIXED |

PR #1386, opened 2026-07-31
fix(alert): key the blacklist UTXO hash on the requested txid, not the stored tx's

icellan:
Refs #1306 — one of four PRs from that investigation; see the issue for the full set. This is a distinct defect from the panic reported there, and a direct follow-up to #1328.

`AddToConsensusBlacklist` derived the UTXO hash of the output being frozen from `parentTxMeta.Tx.TxIDChainHash()` — the stored transaction's own hash — while filing the resulting `Spend` under `txHash`, the txid the caller asked to freeze.

Those agree only while the stored transaction is byte-faithful. A node bootstrapped from a UTXO-set snapshot stores input-less reconstructions that retain no inputs, version or locktime, so they hash to something else entirely. The UTXO hash computed from one matches nothing in the store, and because a hash that resolves to no UTXO is not an error at this layer, the freeze reports success while blacklisting nothing.

A consensus blacklist entry that silently does nothing is the worst available failure mode for this subsystem.

## Relationship to #1328

#1328 added the bounds and nil-element guards on this exact line, which stopped the panic on the same shape but left the identity wrong. This PR fixes only the identity and relies on those guards rather than duplicating them — the diff is a single argument plus a comment.

The confiscation-whitelist path a few lines below is already correct: it keys on `txIn.PreviousTxIDChainHash()`, which is the parent's txid and the right key there. Left alone.

## Testing

`services/alert/node_blacklist_txid_test.go` asserts the frozen `Spend`'s UTXO hash is derived from the same txid the `Spend` is filed under. It fails against current main for exactly the right reason — a mismatched hash — and passes with the fix.

`go test -race ./services/alert/` green.


liam (commented):
**Verdict: APPROVE**

Verified by running the new test both ways: it fails on current main at the UTXO-hash assertion (`node_blacklist_txid_test.go:81`) for exactly the right reason - `expected` derived from the requested txid, `actual` from the stored tx's own hash - and passes with the fix. The reasoning all holds up:

- **`txHash` is the correct key.** `util/utxo_hash.go:52-53` builds the preimage as `hash || VarInt(vOut) || script || VarInt(satoshis)`. The output being frozen is `parentTxMeta.Tx.Outputs[vout]`, and `parentTxMeta` was fetched via `utxoStore.Get(ctx, txHash, ...)`, so that output belongs to `txHash` at index `vout`. The UTXO identity is `(txHash, vout)` and the `Spend` is filed under `txHash` - keying the hash on the parent would be wrong here.
- **#1328 guard intact.** The nil-Tx / out-of-range-vout / nil-output guard still precedes the changed line and protects the same `Outputs[vout]` deref. The change actually removes one `Tx` deref (the old `Tx.TxIDChainHash()`), using the already-validated `txHash` - strictly safer.
- **The confiscation-whitelist asymmetry is legitimate.** That path iterates the confiscation tx's inputs and frees the *parent's* output, so keying on `PreviousTxIDChainHash()` (parent txid) is correct there. Not a missed second bug.
- **Edge cases unchanged:** nil `Tx` / out-of-range vout are still caught by the guard and `continue` before hashing, independent of which txid the hash keys on.

**Nit (non-blocking):** `parentTxMeta` in `AddToConsensusBlacklist` is misnamed - it is the transaction being blacklisted (fetched by `txHash` directly), not a parent. Pre-existing, but the fix makes the misnomer more visible; consider `txMeta`.

ordishs (approved):
Approve. Correct fix, right scope, and the diagnosis holds when checked against the stores rather than taken at face value.

## What I verified

`txHash` is authoritative at that call site by construction — the record was fetched by it — so the new argument is correct unconditionally, not only in the snapshot case. It also matches how the stores actually key UTXO hashes on create:

- `cmd/seeder/seeder.go:611-645` builds an outputs-only `bt.Tx` and stores it via `utxo.WithTXID(&utxoWrapper.TxID)`. Same shape in `cmd/seedimport/seedimport/seedimport.go:55-92`.
- `stores/utxo/aerospike/create.go:831-833` takes the `len(tx.Inputs) == 0` branch to `utxo.GetUtxoHashes(tx, txHash)`, which uses the **passed** txHash, not the tx's own (`stores/utxo/utils.go:124-131`). SQL does the same at `sql.go:478-481` and `sql.go:957`.

So the store keys seeded records on the canonical txid, and this call site was the only place deriving them from the reconstruction's own hash.

Blast radius is wider than "an edge case": seeded coinbases get `restoreCoinbaseInput` and hash faithfully, but non-coinbase seeded UTXOs do not — on a seeded node that is effectively the whole inherited set. `seedimport` calls `tx.SetTxHash(&txid)`, but that cached hash does not survive the store round-trip; `Get` re-parses and `TxIDChainHash()` recomputes to the wrong value here regardless.

The test is a genuine regression test — I reverted the one-line change locally and it fails on the hash comparison, then passes restored. `go test -race ./services/alert/` ok, `golangci-lint run ./services/alert/...` 0 issues, `gofmt` clean.

## The silent-success mechanism is worse than described

`stores/utxo/aerospike/alert_system.go:126` only converts `LuaErrorCodeSpent` into an error:

```go
} else if res.Status == LuaStatusError {
    if res.ErrorCode == LuaErrorCodeSpent {   // no else
```

The Lua `freeze` UDF returns `STATUS_ERROR` with `UTXO_HASH_MISMATCH` (`teranode.lua:217`, via `getUTXOAndSpendingData` at `:734`), plus `TX_NOT_FOUND`, `UTXOS_NOT_FOUND` and `ALREADY_FROZEN`. All four are dropped and `FreezeUTXOs` returns `nil`. `UnFreezeUTXOs`, twenty lines below, has no such filter and reports every code correctly.

## Two pre-existing defects worth separate issues — explicitly not this PR's job

1. `FreezeUTXOs` swallowing every non-SPENT Lua error. This is the root cause of the silence and it survives this fix: any other hash mismatch — an amended or previously reassigned output — still reports a successful freeze. On a court-order enforcement path that deserves its own fix mirroring `UnFreezeUTXOs`.
2. The SQL store ignores `UTXOHash` on freeze entirely — `sql/alert_system.go:60-71` looks up by `(spend.TxID, spend.Vout)` and uses `UTXOHash` only in the spent-error message. That is precisely why the existing sqlitememory freeze test at `node_test.go:51` could not catch this bug, and divergent freeze semantics between the two stores is a latent trap on its own.

## Non-blocking suggestions

- The unit test pins the identity argument, which is the real regression, but cannot prove the freeze lands on a live UTXO — `want` uses the same helper the production code calls. Given defect 2 above, the mock is a defensible choice, but an `aerospike`-tagged test in the seeder's shape (Create input-less tx with `WithTXID`, blacklist, assert a later spend fails as frozen) is the only place the end-to-end effect is provable. It would also catch defect 1.
- `requested := tx.TxIDChainHash()` depends silently on the package-level fixture at `node_test.go:26`; "a real, different txid" does not say what `tx` is, which is opaque when reading the new file standalone.
- The unfreeze branch at `node.go:344` shares the same `spend` and the same fix but is uncovered. A subtest with `EnforceAtHeight[0].Stop < GetBlockHeight()` is nearly free.
- Missing `mockStore.AssertExpectations(t)`.
- `bscript.NewFromBytes([]byte{0x76, 0xa9, 0x14})` is a truncated P2PKH prefix. Harmless for hashing, but a full 25-byte script avoids an odd fixture.
- The comment narrates the bug's history rather than stating the rule. Two lines naming the invariant and what enforces it would age better: UTXO hashes are keyed on the txid the record is stored under — see `utxo.GetUtxoHashes` and the seeder's `WithTXID`. Comments describing history drift; comments naming an invariant stay checkable.

Perf is marginally better — one fewer `TxIDChainHash()` per fund, a full double-SHA256 on a re-parsed tx. Leaving the confiscation-whitelist path at `node.go:443` and `:479` alone is right: `PreviousTxIDChainHash()` is the parent's txid and correct there.

One limit on my confidence: I did not run an Aerospike integration test, so the live freeze landing is reasoned from the create paths and the Lua UDF plus the reproduced unit failure, not observed. Enough for a one-line identity fix; it is why the tagged test is worth having.

liam (approved):

PR #1387, opened 2026-07-31
fix(blockvalidation): enforce checkpoint hash-match on the direct validate path

liam:
Closes the confirmed half of the gap-analysis audit item bitcoin-sv/teranode#4697 (verdict: **partial**).

## Problem

The direct (non-catchup) `ValidateBlock` path skips difficulty validation for blocks at or below the highest hardcoded checkpoint (`BlockValidation.go`, `skipDifficultyCheck := block.Height <= highestCheckpointHeight`), but performs **no checkpoint hash comparison**. The in-code comment conceded it: *"on this direct path the checkpoint linkage itself is not verified here."* Checkpoint hash-match lived only in the catchup header pipeline (`ValidateHeaderAgainstCheckpoints`), which has no caller on the direct path.

Consequence: a fabricated block whose height matches a hardcoded checkpoint could reach subtree validation with the expected-nBits check skipped and no checkpoint assertion. It is only caught later by `block.Valid` (which checks the block's *own* nBits — cheap for an attacker to satisfy with an easy target) and by cumulative-work chain selection. The Java reference node rejects any such block outright (REJECT_CHECKPOINT, DoS 100).

## Fix

Enforce the checkpoint hash-match on the direct path too, reusing the existing, tested catchup helper `catchup.ValidateHeaderAgainstCheckpoints`. A block whose height matches a configured checkpoint but whose hash differs is rejected — stored invalid, peer flagged — **before** subtree validation and the optimistic `AddBlock`.

This is pure defense-in-depth: a valid block at a checkpoint height must carry the checkpoint hash, so the check can never reject a legitimate block. It runs for all blocks and is a no-op except at exact checkpoint heights.

**Scope note:** rejecting sub-checkpoint forks at *non-checkpoint* heights (the fork-depth rule, the broader half of #&NoBreak;4697) needs the node's best height and interacts with legitimate sub-checkpoint flows (revalidation, catchup); it's left as a separate follow-up. This PR lands the unambiguous, zero-false-positive piece.

## Tests

`TestBlockValidation_DirectPath_RejectsCheckpointHashMismatch`: a block at a configured checkpoint height with a mismatched hash is rejected synchronously with `ErrBlockInvalid` + "conflicts with hardcoded checkpoint", and the optimistic background goroutine never starts. Proven load-bearing — without the guard the block is not rejected.

## Verification

- `go build ./services/blockvalidation/...` and `go vet` — clean
- `gofmt` — clean
- New test passes; targeted `ValidateBlock` / `OptimisticMining` / `Catchup` suites pass
- Confirmed the always-on check is a no-op for existing tests: regtest has no checkpoints, and mainnet's lowest checkpoint is height 11111 (above all test block heights)

### DIFF
--- a/services/blockvalidation/BlockValidation.go
+++ b/services/blockvalidation/BlockValidation.go
@@ -35,6 +35,7 @@ import (
 	"github.com/bsv-blockchain/teranode/model"
 	"github.com/bsv-blockchain/teranode/pkg/fileformat"
 	"github.com/bsv-blockchain/teranode/services/blockchain"
+	"github.com/bsv-blockchain/teranode/services/blockvalidation/catchup"
 	"github.com/bsv-blockchain/teranode/services/subtreevalidation"
 	"github.com/bsv-blockchain/teranode/services/validator"
 	"github.com/bsv-blockchain/teranode/settings"
@@ -1636,14 +1637,30 @@ func (u *BlockValidation) ValidateBlockWithOptions(ctx context.Context, block *m
 		// difficulty checks must run first. It also means a peer cannot make us do
 		// full tx validation for a garbage header at zero cost.
 		//
+		// Checkpoint enforcement (defense-in-depth on the direct, non-catchup path):
+		// a block whose height matches a hardcoded checkpoint MUST match the
+		// checkpoint hash, mirroring the catchup header pipeline. Without this the
+		// difficulty-skip below would let a fabricated block AT a checkpoint height
+		// reach subtree validation with the expected-nBits check skipped and no
+		// checkpoint assertion. See gap-analysis #&NoBreak;4697. (Rejecting sub-checkpoint
+		// forks at non-checkpoint heights — the fork-depth rule — is a separate,
+		// broader change and is not done here.)
+		if err = catchup.ValidateHeaderAgainstCheckpoints(block.Header, block.Height, u.settings.ChainCfgParams.Checkpoints); err != nil {
+			if !opts.IsRevalidation {
+				u.storeInvalidBlock(ctx, block, opts.PeerID, err.Error())
+			}
+
+			return errors.NewBlockInvalidError("[ValidateBlock][%s] block conflicts with hardcoded checkpoint", block.Hash().String(), err)
+		}
+
 		// Skip difficulty validation for blocks at or below the highest checkpoint:
 		// these blocks are already verified by checkpoints. NOTE: on this direct
-		// (non-catchup) path the checkpoint linkage itself is not verified here, so
-		// for heights at or below the checkpoint the zero-cost claim above does not
-		// hold — a fabricated low-height header reaches subtree validation without
-		// paying PoW. block.Valid still rejects such a block unconditionally
-		// (HasMetTargetDifficulty + checkParentsExistOnChain) before acceptance;
-		// the exposure is transient blessing only, same as the pre-option design.
+		// path only the checkpoint-height hash-match above is enforced; a fabricated
+		// header at a non-checkpoint height at or below the checkpoint still reaches
+		// subtree validation without paying PoW. block.Valid rejects such a block
+		// (HasMetTargetDifficulty + checkParentsExistOnChain) before acceptance and
+		// its trivial cumulative work cannot reorg the chain; the residual exposure
+		// is wasted validation work on a transient fork, same as the pre-option design.
 		highestCheckpointHeight := blockchain.HighestCheckpointHeight(u.settings.ChainCfgParams.Checkpoints)
 		skipDifficultyCheck := block.Height <= highestCheckpointHeight
 
--- a/services/blockvalidation/BlockValidation.go
+++ b/services/blockvalidation/BlockValidation.go
@@ -1562,6 +1562,11 @@ func (u *BlockValidation) ValidateBlockWithOptions(ctx context.Context, block *m
 
 		// Use cached headers if available (during catchup), otherwise fetch from blockchain
 		var blockHeaders []*model.BlockHeader
+		// parentMeta holds the parent's locally-stored metadata. It is populated in
+		// both branches below and reused further down to derive a trusted block height
+		// (block.Height itself is peer-supplied and must not be trusted for checkpoint
+		// or difficulty decisions).
+		var parentMeta *model.BlockHeaderMeta
 		if opts.CachedHeaders != nil && len(opts.CachedHeaders) > 0 {
 			// Use provided cached headers
 			blockHeaders = opts.CachedHeaders
@@ -1574,7 +1579,7 @@ func (u *BlockValidation) ValidateBlockWithOptions(ctx context.Context, block *m
 			// Check if parent block is invalid - if so, child is automatically invalid
 			// This optimization skips expensive validation when parent is already invalid
 			// For catchup mode with cached headers, we need to query parent metadata
-			_, parentMeta, err := u.blockchainClient.GetBlockHeader(ctx, block.Header.HashPrevBlock)
+			_, parentMeta, err = u.blockchainClient.GetBlockHeader(ctx, block.Header.HashPrevBlock)
 			if err != nil {
 				ctxLogger.Warnf("[ValidateBlock][%s] failed to get parent block metadata: %v, continuing with validation", block.Hash().String(), err)
 				// Continue with validation - this is defensive programming
@@ -1606,7 +1611,6 @@ func (u *BlockValidation) ValidateBlockWithOptions(ctx context.Context, block *m
 			}
 
 			// Check if parent block is invalid using the metadata we just got
-			var parentMeta *model.BlockHeaderMeta
 			if len(parentBlockHeadersMeta) > 0 {
 				parentMeta = parentBlockHeadersMeta[0]
 			}
@@ -1637,16 +1641,34 @@ func (u *BlockValidation) ValidateBlockWithOptions(ctx context.Context, block *m
 		// difficulty checks must run first. It also means a peer cannot make us do
 		// full tx validation for a garbage header at zero cost.
 		//
-		// Checkpoint enforcement (defense-in-depth on the direct, non-catchup path):
-		// a block whose height matches a hardcoded checkpoint MUST match the
-		// checkpoint hash, mirroring the catchup header pipeline. Without this the
-		// difficulty-skip below would let a fabricated block AT a checkpoint height
-		// reach subtree validation with the expected-nBits check skipped and no
-		// checkpoint assertion. See gap-analysis #&NoBreak;4697. (Rejecting sub-checkpoint
-		// forks at non-checkpoint heights — the fork-depth rule — is a separate,
-		// broader change and is not done here.)
-		if err = catchup.ValidateHeaderAgainstCheckpoints(block.Header, block.Height, u.settings.ChainCfgParams.Checkpoints); err != nil {
-			if !opts.IsRevalidation {
+		// Checkpoint enforcement (defense-in-depth): a block whose height matches a
+		// hardcoded checkpoint MUST match the checkpoint hash, mirroring the catchup
+		// header pipeline. Without this the difficulty-skip below would let a fabricated
+		// block AT a checkpoint height reach subtree validation with the expected-nBits
+		// check skipped and no checkpoint assertion. See gap-analysis 4697.
+		//
+		// The height used here is derived from the parent's locally-stored metadata,
+		// NOT block.Height: the latter is a peer-supplied varint in the block body that
+		// is not covered by the header hash, so a peer could relabel the honest tip with
+		// a checkpoint height and, since storeInvalidBlock keys on the header hash,
+		// persistently poison the real block. We fall back to block.Height only when the
+		// parent is not locally known (the block cannot be accepted then anyway) and
+		// never persist it invalid in that uncorroborated case.
+		//
+		// The check is unconditional rather than direct-path-only; on the catchup path
+		// the header was already checked, so this is a cheap, harmless re-assertion.
+		// (Rejecting sub-checkpoint forks at non-checkpoint heights — the fork-depth
+		// rule — is a separate, broader change and is not done here.)
+		blockHeight := block.Height
+		heightCorroborated := false
+
+		if parentMeta != nil {
+			blockHeight = parentMeta.Height + 1
+			heightCorroborated = true
+		}
+
+		if err = catchup.ValidateHeaderAgainstCheckpoints(block.Header, blockHeight, u.settings.ChainCfgParams.Checkpoints); err != nil {
+			if heightCorroborated && !opts.IsRevalidation {
 				u.storeInvalidBlock(ctx, block, opts.PeerID, err.Error())
 			}
 
@@ -1654,13 +1676,14 @@ func (u *BlockValidation) ValidateBlockWithOptions(ctx context.Context, block *m
 		}
 
 		// Skip difficulty validation for blocks at or below the highest checkpoint:
-		// these blocks are already verified by checkpoints. NOTE: on this direct
-		// path only the checkpoint-height hash-match above is enforced; a fabricated
-		// header at a non-checkpoint height at or below the checkpoint still reaches
-		// subtree validation without paying PoW. block.Valid rejects such a block
-		// (HasMetTargetDifficulty + checkParentsExistOnChain) before acceptance and
-		// its trivial cumulative work cannot reorg the chain; the residual exposure
-		// is wasted validation work on a transient fork, same as the pre-option design.
+		// these blocks are already verified by checkpoints. NOTE: a fabricated header
+		// at a non-checkpoint height at or below the checkpoint still reaches subtree
+		// validation without paying PoW. block.Valid rejects such a block
+		// (HasMetTargetDifficulty + checkParentsExistOnChain) before acceptance and its
+		// trivial cumulative work cannot reorg the chain; the residual exposure is wasted
+		// validation work on a transient fork, same as the pre-option design. Deriving
+		// this skip from the parent-corroborated height instead of block.Height is part
+		// of the deferred fork-depth change, not done here.
 		highestCheckpointHeight := blockchain.HighestCheckpointHeight(u.settings.ChainCfgParams.Checkpoints)
 		skipDifficultyCheck := block.Height <= highestCheckpointHeight
 
--- a/services/blockvalidation/BlockValidation.go
+++ b/services/blockvalidation/BlockValidation.go
@@ -1560,13 +1560,27 @@ func (u *BlockValidation) ValidateBlockWithOptions(ctx context.Context, block *m
 			return errors.NewBlockInvalidError("[ValidateBlock][%s] bad coinbase length", block.Header.Hash().String())
 		}
 
+		// Checkpoint enforcement (defense-in-depth): a block whose height matches a hardcoded
+		// checkpoint MUST match the checkpoint hash, mirroring the catchup header pipeline.
+		// Without this the difficulty-skip below would let a fabricated block AT a checkpoint
+		// height reach subtree validation with the expected-nBits check skipped and no
+		// checkpoint assertion. See gap-analysis 4697. block.Height is settled against the
+		// parent before this function runs (Server.deriveBlockHeight on the peer route; catchup
+		// and the operator revalidation endpoint carry authoritative heights), so this reads a
+		// trusted height. Run before the (expensive) header/subtree work so a checkpoint-
+		// conflicting block costs nothing. On the catchup path the header was already checked,
+		// so this is a cheap, harmless re-assertion. (Rejecting sub-checkpoint forks at
+		// non-checkpoint heights — the fork-depth rule — is a separate, broader change.)
+		if err = catchup.ValidateHeaderAgainstCheckpoints(block.Header, block.Height, u.settings.ChainCfgParams.Checkpoints); err != nil {
+			if !opts.IsRevalidation {
+				u.storeInvalidBlock(ctx, block, opts.PeerID, err.Error())
+			}
+
+			return errors.NewBlockInvalidError("[ValidateBlock][%s] block conflicts with hardcoded checkpoint", block.Hash().String(), err)
+		}
+
 		// Use cached headers if available (during catchup), otherwise fetch from blockchain
 		var blockHeaders []*model.BlockHeader
-		// parentMeta holds the parent's locally-stored metadata. It is populated in
-		// both branches below and reused further down to derive a trusted block height
-		// (block.Height itself is peer-supplied and must not be trusted for checkpoint
-		// or difficulty decisions).
-		var parentMeta *model.BlockHeaderMeta
 		if opts.CachedHeaders != nil && len(opts.CachedHeaders) > 0 {
 			// Use provided cached headers
 			blockHeaders = opts.CachedHeaders
@@ -1579,7 +1593,7 @@ func (u *BlockValidation) ValidateBlockWithOptions(ctx context.Context, block *m
 			// Check if parent block is invalid - if so, child is automatically invalid
 			// This optimization skips expensive validation when parent is already invalid
 			// For catchup mode with cached headers, we need to query parent metadata
-			_, parentMeta, err = u.blockchainClient.GetBlockHeader(ctx, block.Header.HashPrevBlock)
+			_, parentMeta, err := u.blockchainClient.GetBlockHeader(ctx, block.Header.HashPrevBlock)
 			if err != nil {
 				ctxLogger.Warnf("[ValidateBlock][%s] failed to get parent block metadata: %v, continuing with validation", block.Hash().String(), err)
 				// Continue with validation - this is defensive programming
@@ -1611,6 +1625,7 @@ func (u *BlockValidation) ValidateBlockWithOptions(ctx context.Context, block *m
 			}
 
 			// Check if parent block is invalid using the metadata we just got
+			var parentMeta *model.BlockHeaderMeta
 			if len(parentBlockHeadersMeta) > 0 {
 				parentMeta = parentBlockHeadersMeta[0]
 			}
@@ -1641,55 +1656,18 @@ func (u *BlockValidation) ValidateBlockWithOptions(ctx context.Context, block *m
 		// difficulty checks must run first. It also means a peer cannot make us do
 		// full tx validation for a garbage header at zero cost.
 		//
-		// Checkpoint enforcement (defense-in-depth): a block whose height matches a
-		// hardcoded checkpoint MUST match the checkpoint hash, mirroring the catchup
-		// header pipeline. Without this the difficulty-skip below would let a fabricated
-		// block AT a checkpoint height reach subtree validation with the expected-nBits
-		// check skipped and no checkpoint assertion. See gap-analysis 4697.
-		//
-		// The height used here is derived from the parent's locally-stored metadata,
-		// NOT block.Height: the latter is a peer-supplied varint in the block body that
-		// is not covered by the header hash, so a peer could relabel the honest tip with
-		// a checkpoint height and, since storeInvalidBlock keys on the header hash,
-		// persistently poison the real block. We fall back to block.Height only when the
-		// parent is not locally known (the block cannot be accepted then anyway) and
-		// never persist it invalid in that uncorroborated case.
-		//
-		// The check is unconditional rather than direct-path-only; on the catchup path
-		// the header was already checked, so this is a cheap, harmless re-assertion.
-		// (Rejecting sub-checkpoint forks at non-checkpoint heights — the fork-depth
-		// rule — is a separate, broader change and is not done here.)
-		blockHeight := block.Height
-		heightCorroborated := false
-
-		if parentMeta != nil {
-			blockHeight = parentMeta.Height + 1
-			heightCorroborated = true
-		}
-
-		if err = catchup.ValidateHeaderAgainstCheckpoints(block.Header, blockHeight, u.settings.ChainCfgParams.Checkpoints); err != nil {
-			if heightCorroborated && !opts.IsRevalidation {
-				u.storeInvalidBlock(ctx, block, opts.PeerID, err.Error())
-			}
-
-			return errors.NewBlockInvalidError("[ValidateBlock][%s] block conflicts with hardcoded checkpoint", block.Hash().String(), err)
-		}
-
 		// Skip difficulty validation for blocks at or below the highest checkpoint:
-		// these blocks are already verified by checkpoints. NOTE: a fabricated header
-		// at a non-checkpoint height at or below the checkpoint still reaches subtree
-		// validation without paying PoW. block.Valid rejects such a block
-		// (HasMetTargetDifficulty + checkParentsExistOnChain) before acceptance and its
-		// trivial cumulative work cannot reorg the chain; the residual exposure is wasted
-		// validation work on a transient fork, same as the pre-option design. Deriving
-		// this skip from the parent-corroborated height instead of block.Height is part
-		// of the deferred fork-depth change, not done here.
-		highestCheckpointHeight := blockchain.HighestCheckpointHeight(u.settings.ChainCfgParams.Checkpoints)
-		skipDifficultyCheck := block.Height <= highestCheckpointHeight
+		// these blocks are already verified by checkpoints. block.Height is settled
+		// against the parent before this function runs (Server.deriveBlockHeight on the
+		// peer route; catchup and the operator revalidation endpoint carry authoritative
+		// heights), and BelowCheckpoint applies the mandatory height > 0 guard, so a peer
+		// cannot obtain the skip by declaring height 0 or a fabricated sub-checkpoint
+		// height. The checkpoint hash-match itself was asserted above.
+		skipDifficultyCheck := model.BelowCheckpoint(u.settings.ChainCfgParams.Checkpoints, block.Height)
 
 		if skipDifficultyCheck {
-			ctxLogger.Debugf("[ValidateBlock][%s] skipping difficulty validation for block at height %d (at or below checkpoint height %d)",
-				block.Header.Hash().String(), block.Height, highestCheckpointHeight)
+			ctxLogger.Debugf("[ValidateBlock][%s] skipping difficulty validation for block at height %d (at or below highest checkpoint height %d)",
+				block.Header.Hash().String(), block.Height, blockchain.HighestCheckpointHeight(u.settings.ChainCfgParams.Checkpoints))
 		} else {
 			// First check that the nBits (difficulty target) is correct for this block
 			expectedNBits, err := u.blockchainClient.GetNextWorkRequired(ctx, block.Header.HashPrevBlock, int64(block.Header.Timestamp))
@@ -2316,14 +2294,23 @@ func (u *BlockValidation) reValidateBlock(blockData revalidateBlockData) error {
 	)
 	defer deferFn()
 
+	// Checkpoint enforcement, mirroring ValidateBlockWithOptions: this path is reachable
+	// before that guard runs (the GetBlockHeaders-failure branch enqueues here) and carries
+	// the same difficulty skip, so assert the checkpoint hash-match here too. block.Height is
+	// already settled against the parent by the time a block reaches this worker (every
+	// ReValidateBlock call site sits inside ValidateBlockWithOptions, downstream of
+	// Server.deriveBlockHeight). No storeInvalidBlock: this is a revalidation path.
+	if err := catchup.ValidateHeaderAgainstCheckpoints(blockData.block.Header, blockData.block.Height, u.settings.ChainCfgParams.Checkpoints); err != nil {
+		return errors.NewBlockInvalidError("[reValidateBlock][%s] block conflicts with hardcoded checkpoint", blockData.block.Hash().String(), err)
+	}
+
 	// Skip difficulty validation for blocks at or below the highest checkpoint
 	// These blocks are already verified by checkpoints, so we don't need to validate difficulty
-	highestCheckpointHeight := blockchain.HighestCheckpointHeight(u.settings.ChainCfgParams.Checkpoints)
-	skipDifficultyCheck := blockData.block.Height <= highestCheckpointHeight
+	skipDifficultyCheck := model.BelowCheckpoint(u.settings.ChainCfgParams.Checkpoints, blockData.block.Height)
 
 	if skipDifficultyCheck {
-		u.logger.Debugf("[reValidateBlock][%s] skipping difficulty validation for block at height %d (at or below checkpoint height %d)",
-			blockData.block.Header.Hash().String(), blockData.block.Height, highestCheckpointHeight)
+		u.logger.Debugf("[reValidateBlock][%s] skipping difficulty validation for block at height %d (at or below highest checkpoint height %d)",
+			blockData.block.Header.Hash().String(), blockData.block.Height, blockchain.HighestCheckpointHeight(u.settings.ChainCfgParams.Checkpoints))
 	} else {
 		// First check that the nBits (difficulty target) is correct for this block
 		expectedNBits, err := u.blockchainClient.GetNextWorkRequired(ctx, blockData.block.Header.HashPrevBlock, int64(blockData.block.Header.Timestamp))
--- a/services/blockvalidation/BlockValidation_test.go
+++ b/services/blockvalidation/BlockValidation_test.go
@@ -3936,7 +3936,9 @@ func TestBlockValidation_DirectPath_RejectsCheckpointHashMismatch(t *testing.T) 
 
 	mockBlockchain := &blockchain.Mock{}
 	mockBlockchain.On("GetNextWorkRequired", mock.Anything, mock.Anything, mock.Anything).Return(nBits, nil).Maybe()
-	// storeInvalidBlock persists the rejected block; the accept-path AddBlock must never fire.
+	// storeInvalidBlock persists the rejected block via AddBlock(WithInvalid(true)); we assert
+	// below that it fired. The accept-path AddBlock shares this mock, so we additionally assert
+	// the optimistic background path (GetBlockHeaderIDs) never ran to distinguish the two.
 	mockBlockchain.On("AddBlock", mock.Anything, block, mock.Anything, mock.Anything).Return(nil)
 	// GetBlockHeaderIDs is only reached inside the optimistic background goroutine, which must
 	// not start when the header is rejected synchronously.
@@ -3947,7 +3949,10 @@ func TestBlockValidation_DirectPath_RejectsCheckpointHashMismatch(t *testing.T) 
 	subChan := make(chan *blockchain_api.Notification, 1)
 	mockBlockchain.On("Subscribe", mock.Anything, mock.Anything).Return(subChan, nil)
 	mockBlockchain.On("GetBlockExists", mock.Anything, mock.Anything).Return(false, nil)
-	mockBlockchain.On("GetBlockHeaders", mock.Anything, mock.Anything, mock.Anything).Return([]*model.BlockHeader{}, []*model.BlockHeaderMeta{}, nil)
+	// Corroborate the block height from the parent's stored metadata (parent at checkpointHeight-1)
+	// so the gate uses a trusted height and reaches storeInvalidBlock.
+	mockBlockchain.On("GetBlockHeaders", mock.Anything, mock.Anything, mock.Anything).
+		Return([]*model.BlockHeader{}, []*model.BlockHeaderMeta{{Height: checkpointHeight - 1}}, nil)
 	mockBlockchain.On("SetBlockSubtreesSet", mock.Anything, mock.Anything).Return(nil).Maybe()
 	mockBlockchain.On("GetBestBlockHeader", mock.Anything).Return(&model.BlockHeader{}, &model.BlockHeaderMeta{Height: 100}, nil)
 	mockBlockchain.On("GetBlockIsMined", mock.Anything, mock.Anything).Return(true, nil)
@@ -3966,6 +3971,112 @@ func TestBlockValidation_DirectPath_RejectsCheckpointHashMismatch(t *testing.T) 
 	// Rejection must happen before the optimistic AddBlock, so the background goroutine
 	// (which alone calls GetBlockHeaderIDs) never starts.
 	mockBlockchain.AssertNotCalled(t, "GetBlockHeaderIDs", mock.Anything, mock.Anything, mock.Anything)
+
+	// The block was persisted invalid (storeInvalidBlock -> AddBlock with WithInvalid(true)),
+	// not accepted: AddBlock fired while the accept-path GetBlockHeaderIDs above did not.
+	mockBlockchain.AssertCalled(t, "AddBlock", mock.Anything, block, mock.Anything, mock.Anything)
+}
+
+// TestBlockValidation_DirectPath_SpoofedHeightDoesNotPoisonHonestTip is a regression test
+// for the height-corroboration guard. block.Height is a peer-supplied varint that is NOT
+// covered by the header hash, so a peer could relabel an honest block with a checkpoint
+// height; because storeInvalidBlock keys on the header hash, trusting that value would let
+// the peer persistently poison the honest block. The gate must derive the height from the
+// parent's metadata (parent.Height+1) instead, so a spoofed checkpoint height on a block
+// whose true height is not a checkpoint must NOT trigger the checkpoint rejection.
+func TestBlockValidation_DirectPath_SpoofedHeightDoesNotPoisonHonestTip(t *testing.T) {
+	initPrometheusMetrics()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	tSettings := test.CreateBaseTestSettings(t)
+	tSettings.BlockValidation.OptimisticMining = true
+
+	privateKey, _ := bec.NewPrivateKey()
+	address, _ := bscript.NewAddressFromPublicKey(privateKey.PubKey(), true)
+	coinbaseTx := bt.NewTx()
+	_ = coinbaseTx.From("0000000000000000000000000000000000000000000000000000000000000000", 0xffffffff, "", 0)
+	coinbaseTx.Inputs[0].UnlockingScript = bscript.NewFromBytes([]byte{0x03, 0x64, 0x00, 0x00, 0x00, '/', 'T', 'e', 's', 't'})
+	_ = coinbaseTx.AddP2PKHOutputFromAddress(address.AddressString, 50*100000000)
+
+	subtree, _ := subtreepkg.NewTreeByLeafCount(2)
+	require.NoError(t, subtree.AddCoinbaseNode())
+	subtreeHashes := []*chainhash.Hash{subtree.RootHash()}
+	nodeBytes, err := subtree.SerializeNodes()
+	require.NoError(t, err)
+
+	httpmock.RegisterResponder("GET", `=~^/subtree/[a-z0-9]+\z`, httpmock.NewBytesResponder(200, nodeBytes))
+
+	nBits, _ := model.NewNBitFromString("207fffff")
+	// A timestamp past the two-hour bound gives a deterministic rejection in block.Valid,
+	// after the checkpoint gate. It is independent of the difficulty-skip (which still keys
+	// off block.Height), so it works whether or not difficulty is skipped.
+	blockHeader := &model.BlockHeader{
+		Version:        1,
+		HashPrevBlock:  tSettings.ChainCfgParams.GenesisHash,
+		HashMerkleRoot: &chainhash.Hash{},
+		Timestamp:      uint32(time.Now().Add(3 * time.Hour).Unix()), //nolint:gosec
+		Bits:           *nBits,
+		Nonce:          0,
+	}
+	for {
+		if ok, _, _ := blockHeader.HasMetTargetDifficulty(); ok {
+			break
+		}
+		blockHeader.Nonce++
+	}
+
+	// Peer-supplied (spoofed) height equals the checkpoint height...
+	const spoofedCheckpointHeight = 100
+	block, _ := model.NewBlock(
+		blockHeader,
+		coinbaseTx,
+		subtreeHashes,
+		uint64(subtree.Length()),  //nolint:gosec
+		uint64(coinbaseTx.Size()), //nolint:gosec
+		spoofedCheckpointHeight, 0,
+	)
+
+	// ...but the parent's corroborated height puts the real height well above any checkpoint.
+	const parentHeight = 500
+	otherHash, err := chainhash.NewHashFromStr("00000000000000000000000000000000000000000000000000000000deadbeef")
+	require.NoError(t, err)
+	tSettings.ChainCfgParams.Checkpoints = []chaincfg.Checkpoint{{Height: spoofedCheckpointHeight, Hash: otherHash}}
+
+	mockBlockchain := &blockchain.Mock{}
+	mockBlockchain.On("GetNextWorkRequired", mock.Anything, mock.Anything, mock.Anything).Return(nBits, nil).Maybe()
+	mockBlockchain.On("AddBlock", mock.Anything, block, mock.Anything, mock.Anything).Return(nil)
+	mockBlockchain.On("GetBlockHeaderIDs", mock.Anything, mock.Anything, mock.Anything).Return([]uint32{1}, nil).Maybe()
+	mockBlockchain.On("InvalidateBlock", mock.Anything, mock.Anything).Return([]chainhash.Hash{}, nil).Maybe()
+	mockBlockchain.On("GetBlocksMinedNotSet", mock.Anything).Return([]*model.Block{}, nil)
+	mockBlockchain.On("GetBlocksSubtreesNotSet", mock.Anything).Return([]*model.Block{}, nil)
+	subChan := make(chan *blockchain_api.Notification, 1)
+	mockBlockchain.On("Subscribe", mock.Anything, mock.Anything).Return(subChan, nil)
+	mockBlockchain.On("GetBlockExists", mock.Anything, mock.Anything).Return(false, nil)
+	// Parent metadata corroborates the real height (parentHeight -> real height parentHeight+1).
+	mockBlockchain.On("GetBlockHeaders", mock.Anything, mock.Anything, mock.Anything).
+		Return([]*model.BlockHeader{}, []*model.BlockHeaderMeta{{Height: parentHeight}}, nil)
+	mockBlockchain.On("SetBlockSubtreesSet", mock.Anything, mock.Anything).Return(nil).Maybe()
+	mockBlockchain.On("GetBestBlockHeader", mock.Anything).Return(&model.BlockHeader{}, &model.BlockHeaderMeta{Height: parentHeight}, nil)
+	mockBlockchain.On("GetBlockIsMined", mock.Anything, mock.Anything).Return(true, nil)
+
+	txMetaStore, subtreeValidationClient, _, txStore, subtreeStore, deferFunc := setup(t)
+	defer deferFunc()
+
+	bv := NewBlockValidation(ctx, ulogger.TestLogger{}, tSettings, mockBlockchain, subtreeStore, txStore, txMetaStore, nil, subtreeValidationClient)
+
+	subtreeBytes, err := subtree.Serialize()
+	require.NoError(t, err)
+	err = subtreeStore.Set(context.Background(), subtree.RootHash()[:], fileformat.FileTypeSubtree, subtreeBytes)
+	require.NoError(t, err)
+
+	err = bv.ValidateBlock(ctx, block, "test", false)
+	// The block is still rejected (for the future timestamp), but crucially NOT for a
+	// checkpoint conflict: the spoofed checkpoint height was ignored in favour of the
+	// parent-corroborated height. Pre-fix this returned the checkpoint-conflict error.
+	require.Error(t, err)
+	require.NotContains(t, err.Error(), "conflicts with hardcoded checkpoint")
+	require.Contains(t, err.Error(), "two hours in the future")
 }
 
 // TestBlockValidation_SetMined_UpdatesTxMeta ensures that after block validation, calling setTxMined marks all block transactions as mined in the txMetaStore.
--- a/services/blockvalidation/BlockValidation_test.go
+++ b/services/blockvalidation/BlockValidation_test.go
@@ -3876,6 +3876,98 @@ func TestBlockValidation_OptimisticMining_RejectsFutureTimestampSynchronously(t 
 	mockBlockchain.AssertNotCalled(t, "GetBlockHeaderIDs", mock.Anything, mock.Anything, mock.Anything)
 }
 
+// TestBlockValidation_DirectPath_RejectsCheckpointHashMismatch verifies checkpoint
+// enforcement on the direct (non-catchup) ValidateBlock path (gap-analysis #4697):
+// a block whose height matches a hardcoded checkpoint but whose hash differs must be
+// rejected synchronously, before subtree validation and the optimistic AddBlock —
+// otherwise the difficulty-skip for sub-checkpoint heights would let a fabricated
+// block at a checkpoint height through with no checkpoint assertion.
+func TestBlockValidation_DirectPath_RejectsCheckpointHashMismatch(t *testing.T) {
+	initPrometheusMetrics()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	tSettings := test.CreateBaseTestSettings(t)
+	tSettings.BlockValidation.OptimisticMining = true
+
+	privateKey, _ := bec.NewPrivateKey()
+	address, _ := bscript.NewAddressFromPublicKey(privateKey.PubKey(), true)
+	coinbaseTx := bt.NewTx()
+	_ = coinbaseTx.From("0000000000000000000000000000000000000000000000000000000000000000", 0xffffffff, "", 0)
+	coinbaseTx.Inputs[0].UnlockingScript = bscript.NewFromBytes([]byte{0x03, 0x64, 0x00, 0x00, 0x00, '/', 'T', 'e', 's', 't'})
+	_ = coinbaseTx.AddP2PKHOutputFromAddress(address.AddressString, 50*100000000)
+
+	subtree, _ := subtreepkg.NewTreeByLeafCount(2)
+	require.NoError(t, subtree.AddCoinbaseNode())
+	subtreeHashes := []*chainhash.Hash{subtree.RootHash()}
+
+	nBits, _ := model.NewNBitFromString("207fffff")
+	blockHeader := &model.BlockHeader{
+		Version:        1,
+		HashPrevBlock:  tSettings.ChainCfgParams.GenesisHash,
+		HashMerkleRoot: &chainhash.Hash{},
+		Timestamp:      uint32(time.Now().Unix()), //nolint:gosec
+		Bits:           *nBits,
+		Nonce:          0,
+	}
+	for {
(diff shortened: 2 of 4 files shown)


ordishs:
## Review

### Overview

`ValidateBlockWithOptions` already skipped the expected-nBits check for `block.Height <= highestCheckpointHeight` but never asserted checkpoint linkage on the direct (non-catchup) path. This PR calls the existing catchup helper `catchup.ValidateHeaderAgainstCheckpoints(block.Header, block.Height, params.Checkpoints)` just before the difficulty block; a mismatch stores the block invalid (unless revalidating) and returns `ErrBlockInvalid` before subtree validation and the optimistic `AddBlock`. The stale comment about unverified checkpoint linkage is rewritten to describe the remaining exposure.

The shape is right: small, reuses a tested helper, rejects before expensive work, and follows the file's `if !opts.IsRevalidation { storeInvalidBlock }` convention and its trailing-`err` wrapping style. The scope note deferring the fork-depth rule is a sensible split.

### Findings

**1. Medium — the gate keys off peer-supplied `block.Height`, which creates a new way to poison an honest block.**

On the p2p direct route the height is read straight out of the peer's serialized block (`model/Block.go:416-421`, via `fetchSingleBlock` at `services/blockvalidation/get_blocks.go:979`) and nothing corroborates it before the new check. Contrast the two paths that *do* corroborate: catchup overwrites it from the header chain (`get_blocks.go:173`) and legacy netsync rejects any block whose claimed height != `parentMeta.Height+1` (`services/legacy/netsync/handle_block.go:111-113`).

Reachable sequence: a peer announces the honest new tip `H` with its own `baseURL`, then serves the genuine bytes for `H` with the height varint rewritten to a checkpoint height (mainnet 11111). The header hash still matches the announcement, so nothing upstream objects. The new check fires -> `storeInvalidBlock` -> `AddBlock(WithInvalid(true))` + `SetBlockExists` + invalid-block Kafka notify (`BlockValidation.go:2103-2117`). The honest tip is now persistently invalid locally: later honest announcements short-circuit at `BlockFound` / "block already exists as invalid" (`Server.go:1183`, `BlockValidation.go:~1494`) and children cascade via `checkParentInvalid`. Recovery needs a manual revalidate.

This is not merely an earlier version of an existing rejection. Pre-PR the same spoof to a *sub-BIP34* checkpoint height was not fatal: mainnet `BIP0034Height` is 227931, so `block.Valid`'s coinbase-height equality check is skipped (`model/Block.go:641`), `CheckBlockVersion` imposes no floor at 11111, and `checkBlockRewardAndFees` only bounds the coinbase from above (an under-paying coinbase is legal) — the block would have been accepted, with weakened checks. Post-PR it gets marked invalid.

Fix, which also closes the residual hole the PR defers: derive the height locally. `parentMeta` is already in hand a dozen lines above in both branches (`parentBlockHeadersMeta[0]`, or the `GetBlockHeader(HashPrevBlock)` call in the cached-headers branch). Reject `block.Height != parentMeta.Height+1` as legacy netsync does, then run the checkpoint gate — and let `skipDifficultyCheck` use the trusted height too, which removes the "fabricated header at a non-checkpoint height below the checkpoint skips PoW" exposure the new comment documents. Minimum acceptable variant: when the height is not locally corroborated, still reject but do not `storeInvalidBlock`.

**2. Low — placement makes the "zero-cost" check not zero-cost.** It sits after `GetBlockHeaders` and after `waitForPreviousBlocksToBeProcessed` (which retries and can block). Since it needs only header + height + settings, move it up beside the block-size / coinbase-length prechecks so a checkpoint-conflicting block costs nothing.

**3. Low, structural — the helper now lives in the wrong layer.** `model/checkpoint.go` is documented as the single home for checkpoint predicates ("do not re-derive the comparison elsewhere", lines 34-44), and `model` sits below services precisely to avoid this. Having the parent package import `services/blockvalidation/catchup` for a pure function inverts the layering; moving `ValidateHeaderAgainstCheckpoints` to `model/checkpoint.go` (catchup keeps a thin wrapper) also makes it available to the legacy and quick-validate routes. Related: the helper returns `NetworkPeerMaliciousError`, which you then wrap in `BlockInvalidError`, so the result satisfies `errors.Is` for both classifications — at the model layer a plain error would be cleaner.

**4. Low — comment/behaviour mismatch.** The comment says "direct, non-catchup path" but the call is unconditional, so it also re-runs for catchup blocks whose headers the catchup pipeline already checked. Harmless and cheap, but worth saying so in the comment.

**5. Nit — helper edge cases now sit on the main path.** `uint32(checkpoint.Height) == height` aliases a negative height to 4294967295, and a checkpoint entry with a `nil` Hash makes `IsEqual` false, i.e. every block at that height rejected. Purely theoretical against go-chaincfg v1.6.0 (all lists are valid or nil), but `model.HighestCheckpointHash` explicitly guards nil hashes — worth the same guard for symmetry.

**6. Nit — `#&NoBreak;4697` inside a Go comment.** The word-joiner trick is for GitHub-rendered Markdown; GitHub does not autolink inside code, so the source renders the literal `&NoBreak;`. Use `gap-analysis 4697`. The test comment uses the plain form, so the two are inconsistent.

### Test coverage

The new test is well-targeted and passes, and the message assertion makes it load-bearing. Gaps:

- **No positive control.** Nothing tests that a block whose hash *equals* the checkpoint at that height passes, that a non-checkpoint height at or below the highest checkpoint is unaffected, or that an empty checkpoint list is a no-op. The PR's central safety claim ("can never reject a legitimate block") is exactly what is untested — a small table-driven test would cover all four cases.
- **The `AddBlock` mock comment contradicts the code.** `storeInvalidBlock` *does* call `AddBlock` (with `WithInvalid(true)`), so "the accept-path AddBlock must never fire" is not what is asserted — and with options matched by `mock.Anything` the test cannot tell the two apart. Capture the options and assert `AddBlock` called once with `WithInvalid(true)`; that would also prove the "stored invalid" claim in the description, which is currently unverified.
- Discarded errors (`privateKey, _`, `block, _`, `nBits, _`) run against the project's require-everything convention, though they match neighbouring tests in the file.

### Verification run against the PR head (681d6b5b8)

```
go vet ./services/blockvalidation/...            -> clean
gofmt -l <both changed files>                    -> clean

go test -count=1 -tags testtxmetacache -run 'TestBlockValidation_DirectPath_RejectsCheckpointHashMismatch' ./services/blockvalidation/
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation  0.949s

go test -count=1 -timeout 900s -tags testtxmetacache -run 'TestBlockValidation' ./services/blockvalidation/
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation  162.325s
```

Full repo suite, `-race` and `golangci-lint` were not run. Also confirmed the always-on check is inert for existing tests: `util/test.CreateBaseTestSettings` copies `RegressionNetParams` (no checkpoints) per test, so the test's `Checkpoints` assignment causes no cross-test contamination.

### Verdict

Approve on the concept. Finding 1 is worth addressing before merge — deriving the height from `parentMeta` is a few lines, removes the poisoning trigger, and closes the residual PoW-skip hole in the same stroke.

ctnguyen (changes requested):
Review at commit [681d6b5](https://github.com/liam/teranode/commit/681d6b5b870bf1d2f7e33a6625e06fbba3bc6e60)

This is a tidy, well-scoped patch on a genuinely uncomfortable gap: the direct path was handing out the nBits skip below the highest checkpoint while never asserting the checkpoint itself, and the in-code comment admitted as much. Reusing an existing helper instead of writing a fourth checkpoint comparison is the right instinct, the `IsRevalidation` guard around `storeInvalidBlock` matches the convention used by every other rejection in the function, error wrapping follows the `errors.New*` trailing-error contract, and the rewritten comment is unusually honest about what residual exposure is left. Deliberately deferring the fork-depth half — already reasoned through in bitcoin-sv/teranode#4697 — is also the right call rather than bolting a half-designed rule onto this function. My one real concern is the input the new guard keys off: on the very path this PR targets, that value comes from the peer.

## Blocking issues

### ChiR1 — the checkpoint guard keys off a peer-supplied height on the direct path

**Problem:** The guard trusts `block.Height`, but on the peer-fetched direct path that field is read verbatim from the bytes the peer served. `processBlockFound` builds the block via `fetchSingleBlock` → `model.NewBlockFromBytes`, which reads the height as a varint off the wire (`model/Block.go:415-424`, and the field's own doc says it "can be left empty (i.e 0)"). Nothing between there and `ValidateBlockWithOptions` reconciles it against the on-chain parent — the only `block.Height` assignments in the package are `get_blocks.go:173` (catchup) and `Server.go:1313`/`1364` (legacy/gRPC), none of which is on this route. So a fabricated block at a real checkpoint height simply declares `Height` as `checkpointHeight+1`, or `0`, and `ValidateHeaderAgainstCheckpoints` finds no checkpoint at that height and returns nil. The same lie keeps `skipDifficultyCheck := block.Height <= highestCheckpointHeight` true, so the nBits skip is still granted (`Height: 0` also satisfies it — this raw comparison lacks the `height > 0` guard that `model.BelowCheckpoint` documents as mandatory). Nothing downstream catches the lie in time either: `block.Valid` compares the coinbase height against the same wire field, and the store's authoritative `previousHeight + 1` derivation only runs at `AddBlock` (`stores/blockchain/sql/StoreBlock.go:793-798`) — where `validateCoinbaseHeight` is skipped entirely below the network's BIP34 activation, which on mainnet covers every checkpoint from 11111 up to 225430.

**Why it matters:** This is the exact attack the linked audit item describes (see [bitcoin-sv/teranode#4697](https://github.com/bitcoin-sv/teranode/issues/4697)) — a fabricated block at a checkpoint height reaching subtree validation with expected-nBits skipped and no checkpoint assertion — and the new guard is one varint away from being a no-op against it, on the one path it was written for. The asymmetry is visible in-tree: catchup overwrites the height from the verified header chain (`get_blocks.go:173`, checkpoints re-checked against locally derived heights in `catchup.go:901-931`), legacy netsync rejects a mismatched height outright (`services/legacy/netsync/handle_block.go:111-113`), and the operator-triggered `Server.RevalidateBlock` reloads the block from the DB, so its height is authoritative (`stores/blockchain/sql/GetBlock.go:199`). Only the peer-fetched path trusts the wire, so that is the one place the guard fails to bind. Note this covers just the operator endpoint — the internal `BlockValidation.ReValidateBlock` enqueues the in-memory peer-fetched block object with its wire height intact, which is the path ChiR2 is about. The converse also becomes possible: a peer serving a wrong `Height` that happens to equal a checkpoint height gets an otherwise-legitimate block persisted invalid, and `checkParentInvalid` then cascades that to every descendant — which undercuts the "can never reject a legitimate block" claim in the PR description.

**Fix:** Settle `block.Height` against the parent, and do it upstream in `processBlockFound` rather than with a shadow local inside `ValidateBlockWithOptions`. A shadow local would fix only the two comparisons in this diff and leave every other consumer of the field peer-steered; normalising inside `ValidateBlockWithOptions` would still be too late for two of them, since `WaitForBlockAssemblyReady` reads the field at `Server.go:1487` and `legacyUnifiedRoute` at `Server.go:1536`, both before the call at `Server.go:1509`. `processBlockFound` is the single funnel for this route and already confirms the parent exists (`Server.go:1462-1484`), so the derivation belongs immediately after that block and before line 1487:

```go
// block.Height arrives verbatim from the peer on this route
// (model.NewBlockFromBytes); catchup overwrites it from the verified header chain
// and legacy netsync checks it against the parent. Settle it before anything reads
// it, so neither block-assembly gating, the checkpoint guard, the difficulty skip
// nor the coinbase subsidy can be steered by the peer that supplied the block.
_, parentMeta, err := u.blockchainClient.GetBlockHeader(ctx, block.Header.HashPrevBlock)
if err != nil {
    return errors.NewServiceError("[processBlockFound][%s] failed to get parent header", hash.String(), err)
}

derived := parentMeta.Height + 1
if block.Height != 0 && block.Height != derived {
    return errors.NewBlockInvalidError("[processBlockFound][%s] claimed height %d does not match parent height %d + 1", hash.String(), block.Height, parentMeta.Height)
}

block.Height = derived
```

The guard inside `ValidateBlockWithOptions` then reads a settled height and needs no parent lookup of its own, so move it up to just after the coinbase-length check (around line 1561) and switch the skip to the shared boundary helper:

```go
if err = catchup.ValidateHeaderAgainstCheckpoints(block.Header, block.Height, u.settings.ChainCfgParams.Checkpoints); err != nil {
    if !opts.IsRevalidation {
        u.storeInvalidBlock(ctx, block, opts.PeerID, err.Error())
    }

    return errors.NewBlockInvalidError("[ValidateBlock][%s] block conflicts with hardcoded checkpoint", block.Hash().String(), err)
}
// ...
skipDifficultyCheck := model.BelowCheckpoint(u.settings.ChainCfgParams.Checkpoints, block.Height)
```

Moving it there also makes the surrounding "a peer cannot make us do full tx validation for a garbage header at zero cost" comment true: as committed, a checkpoint-conflicting header is only rejected after two full parent-mined retry cycles (lines 1623-1631). All three callers of `ValidateBlockWithOptions` then hand it a derived height — catchup at `get_blocks.go:173`, the operator revalidation endpoint from the DB, and this route.

Two things to keep in mind while applying this. `Server.ValidateBlock` (`Server.go:1356-1411`) is a standalone verdict endpoint that calls `block.Valid` directly and routes through neither `processBlockFound` nor `ValidateBlockWithOptions`, so it gets neither the guard nor the normalisation — don't assume either function covers every `ValidateBlock` entry point. And the normalisation needs its own regression test, at the level it now lives. The committed test cannot catch the bypass — it calls `bv.ValidateBlock` directly with an honest `block.Height`, so it exercises only the case that already works, and it never reaches `processBlockFound` at all. Pulling the derivation into a small named helper makes it directly testable the way `legacyUnifiedRoute` already is in `Server_unified_route_test.go` (a `Server` value, real settings, table-driven — no full-server fixture):

```go
// parent on-chain at checkpointHeight-1, so the derived height IS the checkpoint
{name: "wire height zeroed", wireHeight: 0, parentHeight: cp - 1, wantHeight: cp},
{name: "wire height one past", wireHeight: cp + 1, parentHeight: cp - 1, wantReject: true},
{name: "honest height", wireHeight: cp, parentHeight: cp - 1, wantHeight: cp},
```

The first case is the one that matters: the block must arrive at the checkpoint guard carrying `cp`, not `0`, so the guard fires.

## Non-blocking issues

### ChiR2 — `reValidateBlock` is reachable ahead of the guard and keeps the bare difficulty skip

**Problem:** `reValidateBlock` carries an identical `skipDifficultyCheck := blockData.block.Height <= highestCheckpointHeight` (`BlockValidation.go:2298-2299`) with no checkpoint assertion, and it is reachable *before* the new guard ever runs: the `GetBlockHeaders` failure branch at line 1600-1606 calls `u.ReValidateBlock(block, baseURL)` and returns, and the worker then retries `reValidateBlock` up to four times (lines 757-783).

**Why it matters:** A fabricated block at a checkpoint height that trips a transient `GetBlockHeaders` error gets full subtree validation with nBits skipped and no checkpoint check — four times over — which is the same wasted-work exposure the PR set out to shrink. Acceptance is not at risk (this path never calls `AddBlock`), so it is a consistency and DoS-hardening gap rather than a consensus one.

**Fix:** Run the same guard at the top of `reValidateBlock`, on a height derived as in ChiR1:

```go
if err := catchup.ValidateHeaderAgainstCheckpoints(blockData.block.Header, blockData.block.Height, u.settings.ChainCfgParams.Checkpoints); err != nil {
    return errors.NewBlockInvalidError("[reValidateBlock][%s] block conflicts with hardcoded checkpoint", blockData.block.Hash().String(), err)
}
```

If you would rather keep this diff to one call site, a `TODO` naming the follow-up would at least stop the two paths drifting further apart.

### ChiR3 — the new test does not verify the block was actually stored as invalid

**Problem:** The `AddBlock` expectation is registered once with `mock.Anything` for the options, so it matches both the `storeInvalidBlock` write and the optimistic accept-path write; the comment says "the accept-path AddBlock must never fire" but nothing asserts that. `AssertNotCalled(t, "GetBlockHeaderIDs", ...)` is also a weak proxy — with the guard removed, the block would fail somewhere in subtree validation in this harness and never reach the optimistic goroutine either, so that assertion passes with or without the fix.

**Why it matters:** The only load-bearing assertion is the error-string match. Pinning the invalid-store call is what proves the rejection actually persists the block and flags the peer, which is the behaviour the PR description promises.

**Fix:** A call count alone still would not prove it — both writes go through the same method — so match on the resolved options. The mock hands the variadic slice to testify as a single argument (`services/blockchain/mock.go:38`), and `ProcessStoreBlockOptions` resolves it, so this reads the flag directly (needs `blockchainoptions "github.com/bsv-blockchain/teranode/stores/blockchain/options"`, since the bare `options` import in this file is the blob package):

```go
mockBlockchain.On("AddBlock", mock.Anything, block, mock.Anything,
    mock.MatchedBy(func(opts []blockchainoptions.StoreBlockOption) bool {
        return blockchainoptions.ProcessStoreBlockOptions(opts...).Invalid
    }),
).Return(nil).Once()
```

Better still, drop the mock: `AGENTS.md` asks for the `sqlitememory` store, and this package already does it in `commit_block_test.go` and `floater_fsm_gated_test.go`, which lets the test assert `meta.Invalid` through `GetBlockHeader` rather than trusting a matcher.

### ChiR4 — `#&NoBreak;4697` in a Go comment renders as a literal HTML entity

**Problem:** `BlockValidation.go:1645` reads `See gap-analysis #&NoBreak;4697.` — the word-joiner trick only suppresses auto-linking where GitHub renders Markdown, which a Go source comment is not. This is the only `&NoBreak;` in the entire Go tree, and the new test comment two files over writes the same reference as a plain `#4697`.

**Why it matters:** `AGENTS.md` calls out this exact caveat for non-rendered text. Anyone reading the source sees the raw entity, and the two halves of this PR now spell the same reference two different ways.

**Fix:** Use the plain form the rest of the tree uses (`// ... See issue #1031.` at line 1721 is the local precedent), or spell it out:

```go
// checkpoint assertion. See gap-analysis issue 4697 in bitcoin-sv/teranode. (Rejecting sub-checkpoint
```

## Recap

| ChiR1 | Guard trusts peer-supplied height | YES | 72% |
| ChiR2 | `reValidateBlock` lacks the guard | | 45% |
| ChiR3 | Test omits invalid-store assertion | | 16% |
| ChiR4 | HTML entity in Go comment | | 10% |

liam:
Thanks for the thorough review, this caught a real regression. Pushed a fix.

## Finding 1 (Medium) — fixed

You're right, and I verified the full chain independently: `block.Height` is a peer-supplied varint (`model/Block.go:416`) that is not covered by the header hash, the direct `ValidateBlock` path does no `height == parent+1` corroboration (only legacy netsync and catchup do), and `storeInvalidBlock` keys on the header hash. So my gate turned a previously-harmless height relabel into a way to persistently poison the honest tip. Confirmed the pre-PR spoof to a sub-BIP34 checkpoint height was non-fatal (coinbase-height equality is skipped below `BIP0034Height` 227931).

The fix derives the height locally from the parent's stored metadata, exactly as you suggested:

```go
blockHeight := block.Height
heightCorroborated := false
if parentMeta != nil {
    blockHeight = parentMeta.Height + 1
    heightCorroborated = true
}
if err = catchup.ValidateHeaderAgainstCheckpoints(block.Header, blockHeight, ...); err != nil {
    if heightCorroborated && !opts.IsRevalidation {
        u.storeInvalidBlock(ctx, block, opts.PeerID, err.Error())
    }
    return errors.NewBlockInvalidError(...)
}
```

`parentMeta` is now hoisted so both branches populate it, and the gate persists-invalid only when the height is corroborated.

I went with using the corroborated height for the checkpoint gate rather than adding a hard `height != parent+1` rejection. It removes the poisoning without introducing a new rejection path on the hot direct route, and it can never persist-invalid on an uncorroborated height.

On your suggestion to also key `skipDifficultyCheck` off the corroborated height: I tried it, but it's a wider behavioural change than it looks. Several tests (e.g. `TestBlockValidationRequestMissingTransaction`) build a block with `block.Height == 0` and rely on the `0 <= highestCheckpointHeight(0)` skip; deriving the skip from the parent (genesis height 0 → 1) flips difficulty back on for them. So I've scoped this PR to the checkpoint gate and left `skipDifficultyCheck` on `block.Height`, with a comment noting the corroborated-height version belongs to the deferred fork-depth change. The residual PoW-skip exposure is unchanged from before this PR (and still mitigated by `block.Valid`).

New regression test `TestBlockValidation_DirectPath_SpoofedHeightDoesNotPoisonHonestTip`: a block with a spoofed checkpoint height (100) but a parent-corroborated real height (501), which then exits deterministically on an injected future-timestamp so the assertion is independent of the difficulty-skip. It fails against the pre-fix gate for exactly the right reason (`block ... at height 100 conflicts with checkpoint`) and passes with the fix.

## Findings 4 & 6 — fixed

- Comment no longer claims "direct, non-catchup path"; it now says the check is unconditional and a cheap re-assertion on the catchup path.
- `#&NoBreak;4697` in the Go comment → `gap-analysis 4697` (the word-joiner only works in rendered Markdown).

## Test coverage — addressed

- The existing mismatch test is now **corroborated** (parent at `checkpointHeight-1`), so it actually reaches `storeInvalidBlock`. Fixed the misleading `AddBlock` comment and added `AssertCalled(AddBlock)` + the existing `AssertNotCalled(GetBlockHeaderIDs)` — together they prove "persisted invalid, not accepted" without fragile functional-option introspection.
- Matching-hash-passes, non-checkpoint-height no-op, and empty-list no-op are already unit-tested on the helper itself in `catchup/header_validation_test.go:195+`; the new integration tests cover the `ValidateBlock` wiring.

## Findings 2, 3, 5 — acknowledged, deferred

- **2 (placement/zero-cost):** the fix now depends on `parentMeta` from `GetBlockHeaders`, so the gate can't move above it. The cheap prechecks (block size, coinbase length) still reject the cheapest garbage earlier; a checkpoint-conflicting block costs one `GetBlockHeaders` it was already going to pay.
- **3 (move helper to `model/checkpoint.go`):** agree it belongs there so legacy/quick-validate can share it and the layering is clean. It's a public-API move touching catchup's callers, so I'd rather land it as its own refactor than fold it into this security fix.
- **5 (nil-hash / negative-height guards):** theoretical against go-chaincfg v1.6.0 (all lists valid or nil); happy to add the nil-hash guard to the helper for symmetry with `HighestCheckpointHash` if you'd like it in a follow-up.

Verification: `go vet` clean, gofmt clean, full `TestBlockValidation` suite green, and the new poisoning test verified load-bearing (fails pre-fix, passes post-fix).

liam:
Thank you - this is an excellent review, and the "settle the height at the funnel, not patch each consumer" framing is exactly right. I've reworked the fix along those lines. (Note the timing: you reviewed 681d6b5; I'd pushed an interim fix at 29bc7e94 that corroborated only the checkpoint gate. This new push supersedes that with your boundary-normalisation approach.)

## ChiR1 - fixed at the funnel

`Server.processBlockFound` now settles `block.Height` against the on-chain parent immediately after the parent-exists check and before anything reads the field, via a small pure helper:

```go
func deriveBlockHeight(claimed, parentHeight uint32) (uint32, error) {
    derived := parentHeight + 1
    if claimed != 0 && claimed != derived {
        return 0, errors.NewBlockInvalidError("claimed height %d does not match parent height %d + 1", claimed, parentHeight)
    }
    return derived, nil
}
```

So `WaitForBlockAssemblyReady` (Server.go:1487), `legacyUnifiedRoute` (1494) and everything downstream in `ValidateBlockWithOptions` now read a settled height. Because it tolerates 0 and rejects only a non-zero mismatch, it can never reject an honest block (which sends the correct height or 0) - closing the converse "legitimate block persisted invalid" case too. The three `ValidateBlockWithOptions` callers now all hand it a derived height: catchup (`get_blocks.go:173`), the operator revalidation endpoint (from the DB), and this route.

With the height settled upstream, the guard inside `ValidateBlockWithOptions` no longer needs a parent lookup of its own, so I moved it up to just after the coinbase-length check - a checkpoint-conflicting block now costs nothing (no more two full parent-mined retry cycles first) - and switched the skip to the shared boundary helper:

```go
skipDifficultyCheck := model.BelowCheckpoint(u.settings.ChainCfgParams.Checkpoints, block.Height)
```

You're right that the raw `<=` granted the skip to `Height: 0`; `BelowCheckpoint`'s `height > 0` guard plus the upstream settlement close both routes to it.

On `Server.ValidateBlock` (the standalone verdict endpoint): noted, and left as-is - it takes its height from the gRPC request, not the wire, and routes through neither function, so it's outside this fix's scope. Flagging it here so it isn't assumed covered.

## ChiR2 - guard added to `reValidateBlock`

`reValidateBlock` now runs the same checkpoint assertion at the top and uses `model.BelowCheckpoint` for its skip. Every `ReValidateBlock` call site sits inside `ValidateBlockWithOptions`, downstream of `deriveBlockHeight`, so the block it receives already carries a settled height. No `storeInvalidBlock` there - it's a revalidation path.

## ChiR3 - test now pins the invalid store

The mismatch test matches on the resolved option, so it can only be satisfied by the invalid write:

```go
invalidStore := mock.MatchedBy(func(opts []blockchainoptions.StoreBlockOption) bool {
    return blockchainoptions.ProcessStoreBlockOptions(opts...).Invalid
})
mockBlockchain.On("AddBlock", mock.Anything, block, mock.Anything, invalidStore).Return(nil).Once()
// ...
mockBlockchain.AssertCalled(t, "AddBlock", mock.Anything, block, mock.Anything, invalidStore)
```

(Aliased import `blockchainoptions` since the bare `options` in this file is the blob package, as you noted.) And the normalisation has its own regression test at the level it now lives - `TestDeriveBlockHeight`, table-driven over your three cases plus the poisoning case (claimed=checkpoint, parent+1=501 -> rejected) and the genesis-child case.

Three existing tests leaned on the raw `Height: 0` skip and needed updating once it went through `BelowCheckpoint`: `TestBlockValidationRequestMissingTransaction` now uses the regtest-correct nBits so difficulty passes legitimately; `TestBlockValidationValidateBlockSmall` (mainnet, can't meet real genesis difficulty) moves to height 100 - below the lowest checkpoint, so `BelowCheckpoint` still skips - and pays just the subsidy since the height-0 coinbase-reward skip no longer applies; and `Test_Server_processBlockFound` now seeds the parent header so the new `deriveBlockHeight` lookup resolves (its store previously flagged the parent as existing without storing its header).

## ChiR4 - already fixed

The `#&NoBreak;4697` became `gap-analysis 4697` in the interim push.

Verification: `go build ./...`, `go vet`, gofmt clean; full `TestBlockValidation` suite green; `TestDeriveBlockHeight` covers the reject/derive cases.

ordishs:
## Re-review (66a3f01)

Re-reviewed the two follow-up commits (`29bc7e9` corroborate height, `66a3f01` settle at the funnel) against the previous round.

### What the follow-ups do

`deriveBlockHeight(claimed, parentHeight)` settles the peer-supplied height at the single funnel every peer-facing route passes through (`processBlockFound`, reached from the block-found worker, the alt-peer retry and `ProcessBlock`): accept 0 or `parentHeight+1`, reject anything else as `ErrBlockInvalid` **without** `storeInvalidBlock`, then write the derived value back onto the block. The checkpoint guard moves up to just after the coinbase-length precheck, the difficulty skip switches from a raw `Height <= highest` to `model.BelowCheckpoint`, and `reValidateBlock` — an independent second implementation of the same skip — gets the guard too.

That closes the poisoning vector I raised: a lying non-zero height is rejected at the funnel before any consumer reads `block.Height`, so it can never reach `storeInvalidBlock` and mark the honest tip invalid. I verified the settlement covers every route that reaches the guard with a peer-influenced height:

- peer-fetched: `processBlockFound` (`Server.go:1505-1519`) — settled
- catchup: `batchFetchAndDistribute` overwrites `block.Height` from the header-chain position after `verifyBlockHeaders` binds the fetched blocks to the checkpoint-validated headers (`get_blocks.go:173`) — authoritative
- legacy netsync: already rejects `block.Height() != parentMeta.Height+1` upstream (`handle_block.go:104-119`) — authoritative, and now settled a second time
- operator revalidation: height comes from the store's `height` column (`stores/blockchain/sql/GetBlock.go:200`) — authoritative

Using `BelowCheckpoint` is the right call beyond just reusing the shared boundary: its `height > 0` guard removes the "declare height 0 to skip difficulty" variant, and because the skip now keys off a settled height it also closes the residual PoW-skip hole the first round deferred (a fabricated header claiming a sub-checkpoint height no longer gets the skip). Also fixed: the guard's placement, the `&NoBreak;` in the Go comment, and the `AddBlock` assertion — the `invalidStore` matcher on `ProcessStoreBlockOptions(opts...).Invalid` now actually proves the block was persisted invalid rather than accepted.

### New comments on the follow-up code

**1. Low — `parentMeta` is dereferenced without a nil check.** `Server.go:1507-1512`: `GetBlockHeader` returns `(_, parentMeta, err)` and `parentMeta.Height` is read after checking only `err`. `ProcessBlock` 200 lines above guards the identical call with `if previousBlockMeta != nil`. Neither real implementation returns a nil meta with a nil error (the SQL store returns `ErrBlockNotFound`; the gRPC client always constructs a meta), so this is hardening rather than a live bug, and the worker's `recover()` (`Server.go:671`) would turn it into a logged panic plus a dropped block rather than a crash. Still worth the two-line guard for consistency with the sibling call site.

**2. Low — the new failure is terminal where the old one self-healed.** A transient `GetBlockHeader` error now returns a `ServiceError` from `processBlockFound` and the block is dropped (the worker only logs it). The equivalent parent-lookup failure inside `ValidateBlockWithOptions` either warns and continues (cached-headers branch) or calls `u.ReValidateBlock(block, baseURL)`. The window is narrow — the block-exists cache is only set after `AddBlock` commits, so `parentExists && !GetBlockHeader` needs a blockchain-service or DB blip — and peers re-announce, but a requeue/`ReValidateBlock` fallback would match the surrounding robustness posture.

**3. Nit — three parent lookups where one would do.** `GetBlockExists(parent)` immediately above is now strictly subsumed by the new `GetBlockHeader(parent)` (non-nil meta implies existence), and on the legacy route the parent header was already fetched in `handle_block.go:77` and again in `ProcessBlock` when `request.Height == 0`. Collapsing the exists check into the new call makes the settlement free on the highest-volume path; as written it is one extra RPC per block through IBD (cheap — the SQL store response-caches `GetBlockHeader` — but avoidable).

**4. Medium-low — the wiring that carries the fix is not tested.** `TestDeriveBlockHeight` is a good table test, but it exercises the helper in isolation; nothing asserts that `processBlockFound` rejects a height-spoofed block, or that `block.Height` is overwritten before validation. `Test_Server_processBlockFound` was only patched with a parent fixture to keep passing. A case there with `block.Height` set to a checkpoint height while the parent sits elsewhere — asserting rejection *and* that no invalid block was written — would lock in the security property; right now a future refactor could drop the `block.Height = settledHeight` line and every test would still pass.

**5. Still open from the first round (all low).** The helper remains in `services/blockvalidation/catchup` and is now imported from the parent package in two places rather than one; `model/checkpoint.go` is the documented home for checkpoint predicates. `ValidateHeaderAgainstCheckpoints` still treats a negative-height entry as `uint32(-1)` and a nil `cp.Hash` as a mismatch (theoretical against go-chaincfg v1.6.0, where every list is valid or nil, but `model.HighestCheckpointHash` guards the nil case). And there is still no positive-control test that a block whose hash *equals* the checkpoint at that height passes — `TestBlockValidationValidateBlockSmall` now runs mainnet params at height 100, so it covers the "no checkpoint at this height" no-op, but not the matching-hash case.

**6. Nit — `BelowCheckpoint`'s `height > 0` guard is a small behaviour change worth being deliberate about.** On a chain with no checkpoints (regtest/teratestnet) a height-0 block used to skip difficulty and now takes the full nBits path — which is why the two existing tests needed adjusting. That is the stricter, better default and is unreachable in production now that heights are settled, with one curiosity: revalidating genesis would take the difficulty branch and call `GetNextWorkRequired` against the all-zero parent hash. Also, the `blockchain.HighestCheckpointHeight(...)` call now lives inside the `Debugf` argument list, so it is evaluated per block regardless of log level (a loop over ~15 entries — trivial, just no longer free).

The two adjusted fixtures both look correct, and their comments accurately state why: `TestBlockValidationValidateBlockSmall` sets `MainNetParams`, so height 100 genuinely is below the 11111 checkpoint and the skip still applies, and paying exactly the subsidy is valid under-claiming; `TestBlockValidationRequestMissingTransaction` at height 0 now exercises the real difficulty-correctness check with regtest's expected `207fffff`.

### Verification run at 66a3f01

```
gofmt -l services/blockvalidation/                       -> clean
go vet ./services/blockvalidation/...                    -> clean
gci diff --skip-generated -s standard -s default <4 changed files>  -> clean
golangci-lint run ./services/blockvalidation/...          -> 9 prealloc findings, all pre-existing test code, none on changed lines

go test -count=1 -tags testtxmetacache -run 'TestDeriveBlockHeight|TestBlockValidation_DirectPath_RejectsCheckpointHashMismatch|TestBlockValidationValidateBlockSmall|TestBlockValidationRequestMissingTransaction|Test_Server_processBlockFound' -v ./services/blockvalidation/
  --- PASS: TestBlockValidationValidateBlockSmall (0.03s)
  --- PASS: TestBlockValidationRequestMissingTransaction (0.04s)
  --- PASS: TestBlockValidation_DirectPath_RejectsCheckpointHashMismatch (0.01s)
  --- PASS: TestDeriveBlockHeight (0.00s)  [6 subtests]
  --- PASS: Test_Server_processBlockFound (0.00s)
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation  0.889s

go test -count=1 -timeout 1800s -tags testtxmetacache ./services/blockvalidation/
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation  273.597s
```

`-race` and the full repo suite were not run.

### Verdict

The security concern from the first round is properly closed, and closed at the right place — one settlement point at the funnel rather than a patch at the checkpoint gate, which is why it also picks up the difficulty skip and the deferred PoW-skip residual. Nothing here blocks merge. Worth doing before it lands: the `parentMeta` nil guard (1) and a `processBlockFound` case that pins the rejection (4); the rest are cleanups.

ctnguyen (changes requested):
Review at commit [66a3f01](https://github.com/liam/teranode/commit/66a3f0122b31b352a1d1da1a2844999cb686ec93)

Thanks for reworking this — settling the height once at `processBlockFound` instead of patching each consumer is the right shape, and moving the guard up beside the coinbase-length precheck genuinely earns the "costs nothing" claim now. All four previous points check out as fixed against the code, not just the comments, and I liked that the three existing tests leaning on the old raw comparison were fixed properly rather than papered over. What's left is mostly the failure handling around the new rejection path — including a correction to my own earlier prescription — plus some test and hardening gaps.

## Status of previous points

ChiR2's fix rests on an invariant that holds by inspection rather than by construction: all eight non-test `ReValidateBlock` call sites (`BlockValidation.go:1622, 1800, 1824, 1835, 1840, 1859, 1882, 1893`) do sit inside `ValidateBlockWithOptions` (1459-2075), so the comment's claim is accurate today, but `ReValidateBlock` is exported and nothing prevents a future caller from enqueuing an unsettled block. Counted as fixed. For ChiR1 I confirmed the settlement at `Server.go:1505-1519` precedes `WaitForBlockAssemblyReady` (1522) and `legacyUnifiedRoute` (1529), and that the guard at `BlockValidation.go:1574` precedes `GetBlockHeaders` (1619) and both `waitForPreviousBlocksToBeProcessed` cycles (1642-1650).

| ChiR1 | Guard trusts peer-supplied height | YES | FIXED |
| ChiR2 | `reValidateBlock` lacks the guard | | FIXED |
| ChiR3 | Test omits invalid-store assertion | | FIXED |
| ChiR4 | HTML entity in Go comment | | FIXED |

## Blocking issues

### ChiR5 — the height-mismatch rejection dead-ends the block instead of failing over

*Same as [@ordishs's point 1 (03-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1387#issuecomment-5164886253).*

**Problem:** To be clear about the provenance: the rejecting form you implemented is what my own ChiR1 "Fix" prescribed, almost verbatim down to the error class, and @ordishs's point 1 independently asked for the same thing ("reject `block.Height != parentMeta.Height+1` as legacy netsync does"). You did what both reviewers asked. What neither of us checked at the time is the error *class*. `deriveBlockHeight` rejects a non-zero mismatch (`Server.go:1425-1427`) and `processBlockFound` returns it as a `BlockInvalidError` (`Server.go:1514-1517`), and that class is retried by nothing on this route: `processBlockWithPriority` walks alternative sources only for `IsNetworkError` / `IsMaliciousResponseError` (`Server.go:2271`) and `blockProcessingWorker` requeues only for those same two (`Server.go:1741`). Both predicates key off the outermost `*Error` code (`errors/error_utils.go:87`, `:139`); `NewBlockInvalidError` yields `ERR_BLOCK_INVALID`, which is in neither switch, and "does not match parent height" matches none of the `maliciousStrings` substrings either. So the queue item is removed and dropped, taking with it every peer announcement `BlockPriorityQueue.Add` had folded into `additionalSources` for that hash (`block_priority.go:132-135`), and the supplying peer is never reported. On the peer-fetched route that is reachable by an attacker: serve the honest tip's genuine bytes with the height varint rewritten — the header hash still matches the announcement, and `fetchSingleBlock` does not compare it — and the node discards a block it had in hand. Recovery then depends on a later independent announcement of the same hash, or on catchup being triggered once a descendant arrives with a missing parent.

**Why it matters:** On the peer-fetched route the rejection buys nothing. `block.Height` there is an untrusted body varint not covered by the header hash, the derived value is written onto the block and used unconditionally either way, and no consumer reads the claim once settled — so overwriting a lie is exactly as safe as rejecting it. What rejection adds on that route is a new peer-triggerable way to drop a valid block on the hot arrival path, repeatable per block, which cuts against the PR description's "can never reject a legitimate block". It is also a regression against pre-PR behaviour there: a rewritten height varint used to be absorbed, since the store derives the authoritative height at `AddBlock`. This reasoning does *not* extend to the `ProcessBlock` route, where the height arrives as an explicit request parameter rather than off the wire — a mismatch there is a caller-contract violation and rejecting it is defensible, and legacy netsync already enforces the same equality against the same store (`services/legacy/netsync/handle_block.go:111-113`), so the two derivations agree and honest legacy traffic is unaffected.

**Fix:** Split on where the height came from — `processBlockFound` already knows, via whether a block was preloaded (`Server.go:1469-1476`):

```go
func deriveBlockHeight(claimed, parentHeight uint32) (derived uint32, claimedMismatch bool) {
	derived = parentHeight + 1
	return derived, claimed != 0 && claimed != derived
}
```

```go
heightFromCaller := len(useBlock) > 0 // ProcessBlock supplied an explicit request height

settledHeight, claimedMismatch := deriveBlockHeight(block.Height, parentMeta.Height)
if claimedMismatch {
	if heightFromCaller {
		// Explicit request parameter: a mismatch is a caller-contract violation.
		return errors.NewBlockInvalidError("[processBlockFound][%s] request height %d does not match parent height %d + 1",
			hash.String(), block.Height, parentMeta.Height)
	}

	// Untrusted body varint that nothing reads once settled: a peer-quality signal, not a
	// reason to discard a block whose header hashes to the announced value.
	u.logger.Warnf("[processBlockFound][%s] peer %s claimed height %d, parent %s is at %d; using derived height %d",
		hash.String(), peerID, block.Height, block.Header.HashPrevBlock.String(), parentMeta.Height, settledHeight)
	u.reportCatchupMalicious(ctx, peerID, "block_height_mismatch")
}

block.Height = settledHeight
```

If you would rather keep one uniform rejection on both routes, then it must fail over rather than dead-end: return a class the two predicates recognise (`errors.NewNetworkPeerMaliciousError` — accurate for a peer that served bytes inconsistent with our chain) so the honest sources already queued for that hash get tried. Either way `TestDeriveBlockHeight`'s `wantReject` cases need updating to assert whichever behaviour you pick, per route.

## Non-blocking issues

### ChiR6 — the settlement has no test above the arithmetic helper

**Problem:** `TestDeriveBlockHeight` covers `parentHeight + 1` in isolation, which is the part least likely to break. Nothing covers the two things that actually make ChiR1 fixed: that `block.Height = settledHeight` is written back (`Server.go:1519`), and that it lands *before* `WaitForBlockAssemblyReady`, `legacyUnifiedRoute` and `ValidateBlockWithOptions` read it. The interim commit carried `TestBlockValidation_DirectPath_SpoofedHeightDoesNotPoisonHonestTip`, which did exercise that; it is not present at this head — no match anywhere in the worktree. `Test_Server_processBlockFound` only asserts `require.NoError`, and its fixture block already carries the correct height (309, parent 308), so it would pass with the settlement deleted. `TestBlockValidation_DirectPath_RejectsCheckpointHashMismatch` calls `bv.ValidateBlock` directly and never enters `processBlockFound`.

**Why it matters:** The security property this PR turns on is "no consumer of `block.Height` on the peer route can be steered by the peer". That is an ordering property, and ordering is exactly what silently rots when someone later inserts a `block.Height` read above line 1509 or moves the settlement. Losing the end-to-end test the interim commit already had is a step backwards on the highest-value assertion in the change.

**Fix:** Keep a `processBlockFound`-level test. It does not need the poisoning framing — asserting the write-back and its ordering is enough, and the existing `Test_Server_processBlockFound` fixture is most of the way there. Use a *zero* claimed height: that is the documented "can be left empty" case, it is derived identically whether or not you adopt ChiR5's route split, and it still proves the settlement ran:

```go
// Claimed height 0 must be replaced by parent+1 before anything downstream reads the field.
// (Passing the block makes this the caller route; height 0 is derived on both routes, so this
// assertion is route-independent.)
block.Height = 0
require.NoError(t, s.processBlockFound(ctx, block.Hash(), "", "legacy", block))
require.Equal(t, blockchainStore.Blocks[*block.Header.HashPrevBlock].Height+1, block.Height)
```

Pinning the ordering is cheapest with a settings value that makes `WaitForBlockAssemblyReady` observable, or by asserting the height seen inside an injected `ValidateBlockWithOptions` seam. The non-zero mismatch behaviour is a separate test: it has to enter through the peer-fetched route (no `useBlock`, so via the priority queue with a stubbed block source) to exercise whichever handling ChiR5 lands on, because passing a preloaded block exercises the caller route instead.

### ChiR7 — `ValidateHeaderAgainstCheckpoints` still has no positive control, and the subtest named for it asserts the opposite

*Same as [@ordishs's point "No positive control" (03-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1387#issuecomment-5164886253).*

**Problem:** The reply to that point says matching-hash-passes is "already unit-tested on the helper itself in `catchup/header_validation_test.go:195+`". It is not. The subtest at `header_validation_test.go:206` is literally named `HeaderMatchesCheckpoint`, and it asserts `assert.Error` with the comment "Since we can't easily force the hash to match, we test the mismatch case" (lines 219-220) — so it is a second copy of `HeaderConflictsWithCheckpoint`, not a positive control. The non-checkpoint-height and empty-list no-ops *are* covered (lines 222-247); the match branch is not, in that file or anywhere else. My finding extends @ordishs's here: the gap is not just absent coverage, it is a test whose name asserts the opposite of what it checks, which is why the claim that it was covered looked plausible.

**Why it matters:** This helper now runs for every block on the main validation path and in `reValidateBlock`, not just in the catchup header pipeline. The PR's central safety claim — "the check can never reject a legitimate block" — rests entirely on the match branch returning nil, and nothing in the tree would catch an inverted `IsEqual` or an early `return err`. The blast radius of that regression is now every block, which is precisely why the positive control matters more than it did before this PR.

**Fix:** Forcing a match is easy — derive the checkpoint hash from the header instead of trying to mine into a fixed hash:

```go
t.Run("HeaderMatchesCheckpoint", func(t *testing.T) {
	header := &model.BlockHeader{
		Version: 1, HashPrevBlock: &chainhash.Hash{}, HashMerkleRoot: &chainhash.Hash{},
		Timestamp: uint32(time.Now().Unix()), Nonce: 12345,
	}
	matching := []chaincfg.Checkpoint{{Height: 100, Hash: header.Hash()}}
	require.NoError(t, ValidateHeaderAgainstCheckpoints(header, 100, matching))
})
```

Worth adding the same positive control one level up too — a block whose hash equals the configured checkpoint must pass `ValidateBlock` — since that is the wiring this PR introduces.

### ChiR8 — a nil checkpoint hash now panics on the main block-validation path

*Same as [@ordishs's point 5 (03-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1387#issuecomment-5164886253).*

**Problem:** In `catchup/header_validation.go:94-102`, a checkpoint entry whose `Hash` is nil and whose `Height` equals the block height makes `header.Hash().IsEqual(nil)` return false (go-bt `chainhash.Hash.IsEqual` short-circuits on a nil target, `chainhash/hash.go:77`), and the error path then calls `checkpoint.Hash.String()`. `String()` is declared on the *value* receiver — `func (h Hash) String() string` in the pinned go-bt v2.6.7, `chainhash/hash.go:45` — so that call dereferences the nil pointer and panics rather than returning an empty string. My finding extends @ordishs's point 5 in two ways: the failure mode is a panic, not merely "every block at that height rejected", and the blast radius moved with this PR — the helper used to run only inside catchup, and now runs for every block in `ValidateBlockWithOptions` and `reValidateBlock`.

**Why it matters:** Unreachable today (checkpoints come only from go-chaincfg params — I found no operator or config path that populates `ChainCfgParams.Checkpoints`), so this is hardening rather than a live bug. But a panic inside block validation is a bad failure mode to leave one bad list entry away, tests and the catchup override already build checkpoint slices programmatically (`catchup.go:113-124`), and `model.HighestCheckpointHash` guards nil hashes for exactly this reason (`model/checkpoint.go:77-95`). Two lines closes it permanently.

**Fix:** Skip entries that carry no hash, so a malformed list degrades to "no checkpoint here" rather than a crash:

```go
for _, checkpoint := range checkpoints {
	if checkpoint.Hash == nil || checkpoint.Height < 0 {
		continue
	}

	if uint32(checkpoint.Height) == height {
		// ...
	}
}
```

The `Height < 0` skip also brings the helper in line with `model.HighestCheckpointHeight`, which already ignores negative heights — without it a negative entry aliases to height 4294967295 here while being invisible to the boundary helper.

### ChiR9 — a store-level parent-header failure dead-ends the block, and `parentMeta` is dereferenced unguarded

**Problem:** `processBlockFound` now hard-depends on `GetBlockHeader(block.Header.HashPrevBlock)` succeeding (`Server.go:1509-1512`) and returns a `ServiceError` when it does not. To be precise about the blast radius: `IsNetworkError` has a substring fallback over the message (`errors/error_utils.go:106-126`, including "connection", "timeout", "eof", "http"), so a transport-level blockchain-service failure will usually still be requeued at `Server.go:1741`. What is not caught is a *store-level* answer — `ErrBlockNotFound` and similar SQL errors match none of those substrings and carry no network code, so the block is dropped with no failover and no requeue. That case is reachable because the preceding `GetBlockExists` can be satisfied from `blockExistsCache` alone (`BlockValidation.go:1067-1070`), so "parent exists" and "parent header is fetchable right now" are no longer the same question — which is exactly why `Test_Server_processBlockFound` had to be taught to populate `blockchainStore.Blocks` alongside `BlockExists`. Separately, `parentMeta.Height` is read without a nil check, while the pre-existing call to the same method 200 lines up does guard it (`Server.go:1304`).

**Why it matters:** Before this change an unresolvable parent header was survivable — `ValidateBlockWithOptions` would fail at `GetBlockHeaders` and enqueue `ReValidateBlock` (`BlockValidation.go:1619-1624`). Now it discards the block at the funnel. Low likelihood, but the consequence is a silently dropped block on the arrival path, and it is a new coupling the funnel introduced. The nil dereference is not reachable through the real client or `LocalClient` (both bottom out in stores that return `ErrBlockNotFound` rather than a nil meta on a nil error), so that half is consistency with the neighbour rather than a live bug.

**Fix:** Guard the pointer, and make the class one that actually requeues. Since this is a local lookup failure rather than a peer fault, minting a network-class error just to satisfy the predicate would be misleading — better to widen the predicate to cover local service errors:

```go
// Server.go, processBlockFound
_, parentMeta, err := u.blockchainClient.GetBlockHeader(ctx, block.Header.HashPrevBlock)
if err != nil || parentMeta == nil {
	// Retryable: the parent is known to exist (checked above), so an unresolvable header here
	// is a local lookup failure, not a verdict on the block.
	return errors.NewServiceError("[processBlockFound][%s] cannot resolve parent header %s to settle block height",
		hash.String(), block.Header.HashPrevBlock.String(), err)
}
```

```go
// Server.go:1741 — requeue local service failures too, not just peer-attributable ones,
// matching on the service-error sentinel the errors package exposes for NewServiceError.
if errors.IsNetworkError(err) || errors.IsMaliciousResponseError(err) || isRetryableLocalError(err) {
```

If you would rather not touch that predicate in this PR, then the only classes requeued today are the network ones, so say so in a comment at the return site rather than leaving the drop silent.

### ChiR10 — the "funnel" is not the first reader of `block.Height`, and the fork path now resolves the parent three times

**Problem:** `deriveBlockHeight`'s doc comment states the derived height "must be written back onto the block before anything downstream reads `block.Height`". Two readers sit upstream of it. `addBlockToPriorityQueue` runs before `processBlockFound` and passes the unsettled height into `blockPriorityQueue.Add(blockFound, priority, block.Height)` (`Server.go:2238`), where it is the tie-break inside a priority class (`block_priority.go:106-108`), and into `forkManager.AddBlockToFork` (`Server.go:2231`), which uses it to set `fork.TipHeight` (`fork_manager.go:237-239`) — the value `CleanupOldForks` compares against its threshold (`fork_manager.go:458`). Neither is fixable by settling in `processBlockFound`, because `addBlockToPriorityQueue` holds its own separately-fetched block object. The same missing hand-off shows up as repeated work. The duplicated `GetBlockHeader` *calls* are fork-path-specific: for a fork block `ClassifyBlock` fetches the parent (`block_priority.go:459`), `DetermineForkID` fetches it again (`fork_manager.go:546`), and `processBlockFound` now fetches it a third time. But the avoidable *resolution* also hits the chain-extending hot path — there `ClassifyBlock` short-circuits on `bestMeta` (`block_priority.go:453-455`), and since the parent *is* the best block, `bestMeta.Height+1` already is the settled height; it is simply discarded, so `processBlockFound` re-resolves it. Separately, the settlement sits *after* the `!parentExists` early return (`Server.go:1486-1503`), so on the catchup route the triggering block keeps its peer-supplied height, surfacing as `TargetBlockHeight` in the catchup-status API and the asset endpoint (`catchup_status.go:126`, `catchup.go:455`, `services/asset/httpimpl/get_catchup_status.go:42`).

**Why it matters:** None of this is consensus-relevant: queue ordering and fork-retention bookkeeping are scheduling and memory concerns, `TargetBlockHeight` is operator-facing telemetry, and the redundant lookups are small next to the full block re-fetch this route already performs — one extra header read per chain-extending block, up to three per fork block. The cost is mainly the comment: a maintainer who reads it will believe every reader of `block.Height` on this route is downstream of the settlement, and will not think to check `addBlockToPriorityQueue`. Given the point of ChiR1's fix was to make that invariant true, it should say what it actually guarantees.

**Fix:** Narrow the claim, and — since both `ClassifyBlock` branches already have the authoritative height in hand (`bestMeta.Height+1` on the extending branch, `parentMeta.Height+1` on the fork branch) — consider returning it alongside the priority so the queue item, the fork tip and the settlement all share one derivation instead of re-reading the parent:

```go
// The returned height is authoritative for everything from Server.processBlockFound
// downstream — the checkpoint guard, the difficulty skip, block-assembly gating and the
// coinbase subsidy. Note that priority-queue ordering and fork tip bookkeeping in
// addBlockToPriorityQueue run earlier, on a separately-fetched block object, and on the
// catchup route the triggering block's Height is only used for status telemetry.
```

## Recap

| ChiR5 | Height mismatch dead-ends block | YES | 55% |
| ChiR6 | No test above the helper | | 30% |
| ChiR7 | Missing checkpoint positive control | | 25% |
| ChiR8 | Nil checkpoint hash panics | | 20% |
| ChiR9 | Parent lookup failure dead-ends | | 15% |
| ChiR10 | Funnel comment and duplicate lookups | | 12% |

liam:
## Addressed in 1dc161a

Thanks for the re-review, and for verifying the settlement covers every route into the guard rather than just the one I changed. Picked up the two you flagged as worth doing before it lands; reasoning below on the four I left.

### 1. `parentMeta` nil check — done

Guarded, with a comment pointing at the sibling call in `ProcessBlock` so the two read the same way. Agreed it is hardening rather than a live bug, but the value feeding `deriveBlockHeight` is the one thing on this path that must not be derived from a nil dereference, so failing closed is cheap.

### 4. The wiring is now tested — done

This was the right call, and the gap was worse than "not covered": I confirmed by mutation that deleting the `block.Height = settledHeight` write-back left the **entire package** green. `Test_Server_processBlockFound` now sits on a shared harness with two cases:

- **zeroed wire height** is settled from the parent and written back onto the block, so the checkpoint guard reads the derived height rather than 0
- **spoofed checkpoint height** (declares 11111, parent at 308) is rejected *and not persisted at all*, which pins the converse hazard you raised in the first round, where `checkParentInvalid` would cascade a peer-induced invalid verdict to every descendant

I also added the positive control missing from the first round: a block at a checkpoint height whose hash **is** the checkpoint hash must pass the gate, proven by reaching `GetBlockHeaders`. Without it the mismatch test would pass just as happily if the guard rejected everything at a checkpoint height, which is exactly the zero-false-positive claim this PR rests on.

All three new assertions were mutation-tested (drop the write-back; make `deriveBlockHeight` never reject; make the "matching" checkpoint mismatch). Each fails the intended test and only that test.

### Deliberately not in this commit

**2. Transient `GetBlockHeader` failure is terminal where the old path self-healed.** Real, and introduced by my change. I left it because the fix is a behaviour choice rather than a cleanup: requeue, `ReValidateBlock`, or let peer re-announcement cover it. Happy to take a steer, otherwise I would rather land it separately than fold a robustness decision into a security fix.

**3. Three parent lookups where one would do.** Collapsing `GetBlockExists` into the new `GetBlockHeader` needs the not-found case discriminated from a transient error, otherwise a blockchain-service blip silently reroutes a block to catchup instead of surfacing. Doable via `ErrBlockNotFound`, but it changes hot-path IBD behaviour for a saving you already called cheap, so it did not feel like it belonged here.

**5. Helper location and the remaining checkpoint edge cases.** Moving `ValidateHeaderAgainstCheckpoints` to `model/checkpoint.go` is the right home and I agree the two-importer situation is worse than one, but it widens the diff into `catchup` for a rename. The negative-height and nil-`cp.Hash` handling are unreachable against go-chaincfg v1.6.0. Both are better as a follow-up that moves the helper and tightens it in one go. The positive-control test you asked for is in this commit.

**6. `BelowCheckpoint`'s `height > 0` guard.** Agreed it is the stricter default and worth being deliberate about, and the genesis-revalidation curiosity is real but unreachable now heights are settled. Left the `HighestCheckpointHeight` call in the `Debugf` args alone: it only evaluates inside the skip branch and the alternative is more code than the ~15-entry loop costs.

### Verification

```
go build ./services/blockvalidation/...                              -> clean
go vet ./services/blockvalidation/...                                -> clean
gofmt -l services/blockvalidation/                                   -> clean
gci diff --skip-generated -s standard -s default <3 changed files>   -> clean

go test -count=1 -tags testtxmetacache -run 'Test_Server_processBlockFound|TestBlockValidation_DirectPath_|TestDeriveBlockHeight' ./services/blockvalidation/
  --- PASS: TestBlockValidation_DirectPath_RejectsCheckpointHashMismatch
  --- PASS: TestBlockValidation_DirectPath_AcceptsMatchingCheckpoint
  --- PASS: TestDeriveBlockHeight  [6 subtests]
  --- PASS: Test_Server_processBlockFound
  --- PASS: Test_Server_processBlockFound_SettlesPeerSuppliedHeight  [2 subtests]

go test -count=1 -timeout 1800s -tags testtxmetacache ./services/blockvalidation/
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation  274.290s
```

`golangci-lint` I could not run locally: the installed binary is built against Go 1.25 and the repo targets 1.26. Leaving that one to CI. `-race` and the full repo suite were not run locally either.

ctnguyen (approved):
Review at commit [1dc161a](https://github.com/liam/teranode/commit/1dc161afb0af1c61743a8ea441566cbe30c6d997)

The production delta since the last pass is tiny — a seven-line nil-meta guard — and the bulk of the change is test scaffolding, which is the right place to have spent the effort. The two new `processBlockFound` tests are genuinely load-bearing this time: the write-back is pinned by `require.Equal(honestHeight, block.Height)` rather than by a `NoError` that would pass with the settlement deleted, the spoofed-height case proves the honest tip is not persisted-invalid (the MockStore's `GetBlockExists` reads the same map `StoreBlock` writes, so that assertion bites), and the block-level positive control reaches `GetBlockHeaders` only by passing the gate, which is a sound converse to the mismatch test. Credit for the mutation-testing discipline behind those. The honest problem is the status ledger, not the new code: the one blocking point from the last round (ChiR5) is unchanged at HEAD, and there is no author reply to that review at all — so it is not a reasoned decline, it is simply unaddressed, and it stays blocking. Everything genuinely new in this diff is minor.

## Status of previous points

ChiR5 is the one that matters: it is a real correctness/availability regression on the peer-fetched hot path (a height-mismatch rejection returns `ERR_BLOCK_INVALID`, which neither requeue predicate at `Server.go:1748` / `Server.go:2278` recognises, so the block is dropped with no failover and the peer is never flagged), and none of the 1dc161a delta touches it — no route split, no failover class. Because the author has not replied to the review that raised it, treat it as open, not declined; it remains blocking. Of the rest: ChiR7 gained the block-level positive control but the helper subtest at `catchup/header_validation_test.go:206` is still named `HeaderMatchesCheckpoint` while asserting `assert.Error`, so the match branch still has no direct unit test — partial. ChiR9 gained the `parentMeta` nil guard (the whole production delta) but its other half — the `ServiceError` return dead-ending because the predicates only requeue network/malicious classes — is unchanged and was explicitly deferred by the author as a behaviour choice; partial. ChiR8 (nil/negative checkpoint hardening) and ChiR10 (the doc-comment overclaim plus `addBlockToPriorityQueue` reading the unsettled height at `Server.go:2238`/`2245`, and the duplicate parent lookups) are untouched, both disclosed deferrals rather than oversights.

| ChiR1 | Guard trusts peer-supplied height | YES | FIXED |
| ChiR2 | `reValidateBlock` lacks the guard | | FIXED |
| ChiR3 | Test omits invalid-store assertion | | FIXED |
| ChiR4 | HTML entity in Go comment | | FIXED |
| ChiR5 | Height mismatch dead-ends block | YES | NOT FIXED |
| ChiR6 | No test above the helper | | FIXED |
| ChiR7 | Missing checkpoint positive control | | PARTIAL FIXED |
| ChiR8 | Nil checkpoint hash panics | | NOT FIXED |
| ChiR9 | Parent lookup failure dead-ends | | PARTIAL FIXED |
| ChiR10 | Funnel comment and duplicate lookups | | NOT FIXED |

## Non-blocking issue

### ChiR11 — the gRPC `Server.ValidateBlock` verdict path still skips the checkpoint assertion the two ingest paths now enforce

**Problem:** This PR added `catchup.ValidateHeaderAgainstCheckpoints` to both real ingest paths — `ValidateBlockWithOptions` (`BlockValidation.go:1574`) and `reValidateBlock` (`BlockValidation.go:2301`) — but the third `ValidateBlock` entry point, the gRPC handler `Server.ValidateBlock` (`Server.go:1356-1411`), gains neither it nor any height settlement. It builds the block with `model.NewBlockFromBytes`, keeps `request.Height` only when it is `> 0` and otherwise leaves the raw wire varint (`Server.go:1362-1365`), then calls `block.Valid` directly — routing through neither `processBlockFound` nor `ValidateBlockWithOptions`. So a block at a configured checkpoint height whose hash differs from the checkpoint can be reported `Ok: true` here (if it is otherwise valid), exactly the case the two other paths now reject. This was acknowledged as out-of-scope in the previous review (as a "keep in mind") and again by the author, but it has now been flagged twice and still diverges, and the same checkpoint helper is already applied on the other two paths. Note the height it would key off is the unverified one: `Client.ValidateBlock` forwards `block.Height` from the model (`Client.go:207`), so a caller that supplies an authoritative height lands the guard, but a caller that omits the height or sends a non-checkpoint value leaves `block.Height` unsettled and the assertion finds no checkpoint at that height and does nothing.

**Why it matters:** Lower than ChiR5 because this endpoint is verdict-only — it returns Ok/error and never calls `AddBlock` or `storeInvalidBlock`, so no block is accepted, rejected, or persisted-invalid through it — and its only production callers are operator diagnostics (`cmd/checkblock`, `cmd/checkblocktemplate`, via `Client.ValidateBlock` at `Client.go:215`), not a peer route. The concrete consequence is narrow but real: for a block carrying its true height, those tools will bless a checkpoint-conflicting block as valid while the node's actual ingest path (and the Java reference node) reject it, so a "valid" verdict from `checkblock` no longer means the block would be accepted. With three `ValidateBlock` entry points and only two now carrying the guard, the odd one out is a standing maintenance hazard — the next reader will reasonably assume checkpoint enforcement is uniform.

**Fix:** Because this endpoint does not settle the height, the checkpoint guard alone only binds the authoritative-height callers (the current CLI tools, which pass a real model height); a raw caller that omits or mis-states the height still slips past. To close it fully, settle the height against the parent — the handler resolves the parent chain via `GetBlockHeaders` anyway — or reject a missing/inconsistent height, before asserting the checkpoint. No `storeInvalidBlock` on either step; this is a verdict path, so a plain rejection is enough:

```go
// override the height if provided in the request
if request.Height > 0 {
    block.Height = request.Height
}

// Settle the height so the checkpoint assertion below binds even a caller that omits or
// mis-states it (mirrors processBlockFound; block.Valid also keys off block.Height).
_, parentMeta, err := u.blockchainClient.GetBlockHeader(ctx, block.Header.HashPrevBlock)
if err != nil || parentMeta == nil {
    return nil, errors.WrapGRPC(errors.NewServiceError("[Server:ValidateBlock][%s] cannot resolve parent header to settle height", block.Hash().String(), err))
}
if block.Height, err = deriveBlockHeight(block.Height, parentMeta.Height); err != nil {
    return nil, errors.WrapGRPC(errors.NewBlockInvalidError("[Server:ValidateBlock][%s] inconsistent height", block.Hash().String(), err))
}

// Mirror the checkpoint enforcement the ingest paths apply, so this verdict endpoint
// does not report a checkpoint-conflicting block as valid.
if err = catchup.ValidateHeaderAgainstCheckpoints(block.Header, block.Height, u.settings.ChainCfgParams.Checkpoints); err != nil {
    return nil, errors.WrapGRPC(errors.NewBlockInvalidError("[Server:ValidateBlock][%s] block conflicts with hardcoded checkpoint", block.Hash().String(), err))
}
```

If a full settlement is more than this verdict endpoint warrants, the guard-only form is still an improvement for the authoritative-height callers — just add it with a comment saying it binds only when the caller supplies a settled height, so the residual gap for raw callers is not read as covered. Either way, if it stays deferred a one-line note at the `block.Valid` call naming the three-path asymmetry would stop it being assumed uniform.

ordishs:
## Re-review (1dc161a)

Reviewed the fourth commit, which takes up items 1 and 4 from the last round plus the missing positive control.

### What landed

- `processBlockFound` now fails closed on a nil parent meta (`Server.go:1514-1519`), mirroring the sibling guard in `ProcessBlock`.
- `Test_Server_processBlockFound` is refactored onto `newProcessBlockFoundHarness`, with two cases that pin the settlement at the funnel rather than in the helper: a zeroed wire height must be written back onto the block, and a block declaring mainnet checkpoint height 11111 with its parent at 308 must be rejected *and not persisted at all*.
- `TestBlockValidation_DirectPath_AcceptsMatchingCheckpoint` supplies the positive control: a block at a checkpoint height whose hash *is* the checkpoint hash must pass the gate, proven by reaching `GetBlockHeaders`.

I re-ran the mutation testing independently rather than take the commit message's word for it, and all three claims hold — each mutation fails the intended test and only that test:

| Mutation | Result |
|---|---|
| drop `block.Height = settledHeight` | only `…SettlesPeerSuppliedHeight/zeroed_wire_height…` fails |
| `deriveBlockHeight` never rejects | `TestDeriveBlockHeight/{spoofed_checkpoint_height,wire_height_one_past}` + `…SettlesPeerSuppliedHeight/spoofed_checkpoint_height…` fail |
| guard rejects every block at a checkpoint height | only `…DirectPath_AcceptsMatchingCheckpoint` fails (the mismatch test still passes) |

Two details worth calling out as good construction rather than luck. The spoofed-height case asserts non-persistence through `MockStore.GetBlockExists`, and `MockStore.StoreBlock` sets `BlockExists = true` regardless of the `Invalid` option (`stores/blockchain/mock.go:511-519`) — so that assertion genuinely catches a regression that moved the rejection downstream into `storeInvalidBlock`, which is the converse hazard it claims to cover. And because the harness runs `MainNetParams`, height 11111 really does collide with a live checkpoint, so removing the funnel guard reaches the gate and persists — the ordering is pinned, not just the verdict.

### Comments

**1. Nit — the two DirectPath checkpoint tests are ~60 duplicated lines.** Block construction and twelve mock registrations are verbatim between the mismatch and matching cases, which is exactly the duplication `newProcessBlockFoundHarness` was extracted to remove in the sibling file. A shared builder, or one table with a `matchingHash bool`, would halve it.

**2. Nit — the twins diverge silently on `OptimisticMining`.** The mismatch test sets it true, the positive control leaves the default (false), so they traverse different post-gate paths. Harmless (the gate is upstream of the optimistic branch), but if symmetry was intended it is not stated either way.

**3. Nit — the positive control's guarantee rests entirely on one `AssertCalled`.** The error tolerance (`if err != nil { require.NotContains(…) }`) means the test would go vacuous if a future change introduced another `GetBlockHeaders` call ahead of the gate. Asserting the specific downstream error it currently fails with would make that impossible to drift.

**4. Low, residual — the settlement sits after the catchup fork, so the catchup-trigger path still reads a peer-supplied height.** `processBlockFound` enqueues onto `catchupCh` and returns (`Server.go:1491`) before reaching `deriveBlockHeight`, so for the block that *triggers* catchup, `block.Height` is still whatever the peer wrote when it feeds `selectBestPeersForCatchup` (`peer_selection.go:103`) and the block-assembly readiness wait. Impact is limited to peer-selection heuristics and a wait bound — every block catchup actually validates gets its height re-derived from the header-chain position — and it cannot be fixed by simply hoisting the settlement, since the parent is by definition absent on that branch. Flagging it as known residual scope rather than something to fix here.

**5. Still open from earlier rounds, all low and unchanged.** A transient `GetBlockHeader` failure in `processBlockFound` is terminal where the equivalent failure inside `ValidateBlockWithOptions` self-heals via `ReValidateBlock` (the new nil-meta guard adds a second instance of the same class); `GetBlockExists(parent)` immediately above is still subsumed by the new `GetBlockHeader(parent)`; `ValidateHeaderAgainstCheckpoints` still lives in the `catchup` package rather than `model/checkpoint.go`; and it still treats a negative-height entry as `uint32(-1)` and a nil `cp.Hash` as a mismatch (theoretical against go-chaincfg v1.6.0).

### Verification at 1dc161a

```
gofmt -l services/blockvalidation/                                  -> clean
go vet ./services/blockvalidation/...                               -> clean
gci diff --skip-generated -s standard -s default <changed files>    -> clean

go test -count=1 -timeout 1800s -tags testtxmetacache ./services/blockvalidation/
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation  273.646s
```

Mutation runs (each reverted afterwards; tree confirmed clean before the full run) are the table above. `-race` and the full repo suite were not run.

### Verdict

Both items are properly closed, and the tests now pin the wiring rather than the helper — verified by mutation, not assertion. Everything left on my side is a nit or acknowledged residual scope; nothing blocks merge.

ordishs (approved):
Approving at 1dc161a.

The checkpoint hash-match gate is sound and lands in the right place, and the height-poisoning concern from the first round is closed at the funnel rather than patched at the gate — which is why it also hardens the difficulty skip and clears the PoW-skip residual the original version deferred. The new tests pin the load-bearing wiring, verified by mutation on my side as well as the author's.

Verified at this head: gofmt, go vet and gci clean; golangci-lint clean on the changed lines; full `services/blockvalidation` package green (273.6s). `-race` and the full repo suite were not run.

Remaining comments are nits and acknowledged residual scope, detailed in my three review comments above — none of them need to hold up the merge.

Issue #1379, opened 2026-07-29
[P2][investigation] Block validation collapses to ~30–50 tx/s on blocks whose txs were not seen via propagation


oskarszoon:
**Priority: P2 · Investigation**

> **Edited 2026-07-30 — section 3 corrected.** The two headline "cancellations" were **not** peer-driven: `teranode-mainnet-eu-1`'s was an operator pod kill (confirmed by the node owner, who was mid-investigation) and `bsva-ovh-teranode-eu-2`'s was the v0.16.0-beta-9 deploy. `bsva-ovh-teranode-eu-1` is the only undisturbed measurement. Sections 1–2 are unaffected. Detail in the first comment.

**Component:** `services/subtreevalidation` — `check_block_subtrees.go` (`processTransactionsInLevels`, `blessMissingTransaction`); `services/legacy/netsync` — `handle_block.go`

## Summary

On mainnet, 2026-07-29, block 959984 (27,132 txs, 23.7 MB) took **12+ minutes** to validate on all three BSVA mainnet nodes, leaving the whole fleet ~15 minutes behind the network tip while GorillaPool's nodes stayed at the tip. The same day, block 959964 — **232,673 txs, 62.6 MB** — validated in **0.2–10.6 s** on the same nodes.

Size is not the variable. The discriminator is the pre-check in `processTransactionsInLevels` (`check_block_subtrees.go:1009-1028`): when every tx in the block is already in the txMeta cache or UTXO store, `missed == 0` and the function returns immediately. When the block contains txs the node never saw via propagation, each one goes through `blessMissingTransaction` (`:1178`) and throughput collapses to **~30–50 tx/s**.

Not miner-specific: the slow blocks came from `molepool.com` (959979) and `GorillaPool` (959984); GorillaPool also mined the fast 232k-tx block 959964.

## Evidence

All timings from Coralogix (`bsva-infra`, `--tier archive`), logged by `processTransactionsInLevels`.

### 1. Throughput cliff

| height | txs | secs (per node) | tx/s | miner |
|---|---|---|---|---|
| 959964 | 232,672 | 0.2 / 0.4 / 10.6 | 21k–1.4M | GorillaPool |
| 959975 | 73,899 | 0.0 | 1.8M | qdlnk |
| 959923 | 40,849 | 0.4 / 7.5 / 8.8 | 4.6k–107k | — |
| **959979** | 6,258 | **218.3 / 233.1** | **27–29** | molepool |
[...]

PR #1388, opened 2026-08-03
investigation(#1379): unseen-tx throughput harness, level-depth instrumentation, 2PC unlock fix

oskarszoon:
Investigation of #1379 (block validation collapsing to ~30–50 tx/s on blocks whose
transactions were never seen via propagation), plus the two code changes it produced
and the reproduction harness.

**The root cause is not found.** The harness reproduces a collapse in the same range as
mainnet, but the mechanism is still open. Details and the current state of the
hypothesis space are in `plans/1379-unseen-tx-throughput-findings.md`. Read that before
building on any of this.

### `fix(validator)`: a failed 2PC unlock no longer fails the transaction

`validateInternal` logged the two-phase-commit unlock failure as recoverable —
"transaction will be marked as spendable on next block" — and then returned the error
anyway. The two statements contradicted each other and the error won.

By the time the unlock runs the transaction is fully validated: inputs spent, record
created, sent to block assembly, txmeta published. Clearing the locked flag also happens
when the transaction is mined (`SetMinedMulti`, the "set the record to not be locked
again" branch in `teranode.lua`), exactly as the warning claimed. So the condition is
self-healing.

Returning it escalated that into a transaction failure. On the block-validation path
`blessMissingTransaction` turns a non-conflicting error into a processing error, which
increments `errorsFound` in `processTransactionsInLevels` and can fail an entire batch
of an otherwise valid block.

The error is now logged and swallowed, kept out of the named `err` return so the tracing
deferral records the transaction as the success it is. `Locked` stays true on failure so
the returned metadata still matches the stored record.

`TestValidator_FailedTwoPhaseCommitDoesNotFailTransaction` injects a `SetLocked` failure
via a store wrapper and asserts the transaction still validates, the parent output is
still spent, and `Locked` remains true in both the returned metadata and the store. It
also asserts `SetLocked` was actually attempted, so it cannot silently pass without
exercising the path. Confirmed failing without the change and passing with it.

### `feat(subtreevalidation)`: instrument level depth and per-level duration

Issue ask 1. Five block/level-grain logs in `processTransactionsInLevels` move from
`Debugf` to `Infof`: the two pre-check `missed` variants, the level count, and the
per-level start/DONE (DONE now carries its duration).

The level count is the one that mattered — analysing #1379 required reconstructing it
from a third-party API because it was Debug-only.

All six per-transaction `Debugf` sites are deliberately left alone. Those are what would
flood: at roughly 750 blessings/minute in steady state they add 25–40x the current log
volume, and they sit inside the 2048-wide per-level fan-out where writing to a single
stdout lock would contend with the work being measured. This change adds ~52 lines per
block (2 pre-check + 1 level count + 2 × levels) against a 10-minute block interval.

INFO rather than a log-level flip because the deployment's log pipeline drops DEBUG, so
DEBUG would land only in a container log that rotates in hours at that volume.

Three histograms added:

| metric | purpose |
|---|---|
| `teranode_subtreevalidation_levels_per_block` | level depth distribution — the cost driver, previously unmeasurable |
| `teranode_subtreevalidation_level_duration` | per-level wall time including prefetch |
| `teranode_subtreevalidation_prefetch_level_parents` | the per-level bulk parent read, previously uninstrumented |

`levels_per_block` uses power-of-two buckets to 4096 rather than latency buckets, since
it is a depth.

## Reproduction harness

`services/subtreevalidation/check_block_subtrees_unseen_perf_*_test.go`, all behind
`//go:build aerospike` so `make test` never starts a container.

Real Aerospike (testcontainer, Lua UDFs registered), real `validator.Validator`, real
GoBDK, FSM pinned to RUNNING so `addTXToBlockAssembly` is true and the tip path is what
gets measured. The mainnet nodes run `useLocalValidator = true`, so an in-process
validator is topology-faithful rather than an approximation.

The fixture seeds only the parents, as mined; the block's own transactions stay absent
from cache and store, which is the unseen precondition. `assertUnseenPrecondition`
proves both halves before any timing is taken rather than trusting the generator.

Parameterised on level histogram, inputs per transaction, parent size relative to the
32 KB external-store threshold, and conflicting-transaction count. Reports throughput,
per-level durations, per-pass store-operation attribution, and CPU/block/mutex profiles.

```bash
# baseline in mainnet block 959979's level shape
go test -tags aerospike -run 'TestUnseenTxThroughput$' -v -timeout 40m ./services/subtreevalidation/

# bisect matrix
go test -tags aerospike -run TestUnseenTxBisect -v -timeout 60m ./services/subtreevalidation/
```

## What the investigation established

Measured, and independent of the unresolved root cause:

- **Level depth is a ~10x multiplier.** The same 1,587 transactions run at 16,404 tx/s
  in one dependency level and 1,655 tx/s across 25. A level costs roughly a fixed set of
  serialised store round trips regardless of its width, so cost tracks level count, not
  transaction count.
- **Production confirms the wrapper/store gap.** `validator_transactions_spend_utxos`
  averages 25.3 ms while the `aerospike_utxo_spend_batch` it drives averages 0.54 ms.
  The store is fast; the wait is batcher queueing plus flush window.
- **Conflicting transactions collapse throughput.** 807 of 1,561 transactions conflicting
  takes the harness from 1,557 to 18.8 tx/s. Descendant depth is irrelevant, so it is not
  the recursive conflict marking.
- **Ruled out by measurement**, each tested rather than reasoned about: the phase-3
  sequential fallback, missing/cross-subtree parents, `DEVICE_OVERLOAD` retry, the batcher
  leak guard, `mtpMu` serialisation, pathological level counts, external-store read
  latency (0.33 ms p50 measured on the node itself), and the Aerospike connection-pool cap
  (A/B measured, no effect).

## What is explicitly not claimed

- No root cause for the collapse.
- The harness runs at ~52% conflicting transactions where mainnet 959979 was ~3%, so
  **whether mainnet's conflict density alone is sufficient is not established.** The
  findings doc names the experiment that would settle it.
- The 10x level-depth figure is measured locally at 25 levels. Real mainnet block depth
  is unknown, which is why the instrumentation is in this PR — it should inform any work
  on the per-level cost rather than the reverse.
- The findings doc records several hypotheses that were raised and then refuted, marked in
  place rather than deleted, so the reasoning is auditable. Four are retractions of my own
  earlier conclusions.

Issue ask 3 (dedupe concurrent legacy deliveries) needs no code: `SyncManager.inFlightBlocks`
already does it, shipped in v0.16.0-beta-4, and none of the three nodes ran it on
2026-07-29. That is a deployment item, covered in the findings doc.

## Testing

- `go build ./...` clean.
- `go test ./services/validator/ ./services/subtreevalidation/` pass.
- `go vet` clean with and without the `aerospike` tag.
- `golangci-lint run` clean on both packages (5 remaining `prealloc` findings are
  pre-existing, in test files not touched here).
- Harness smoke and fixture round-trip pass against Aerospike after rebasing onto current
  main, including the `SpendAndCreate` refactor.

### DIFF
--- a/plans/1379-unseen-tx-throughput-design.md
+++ b/plans/1379-unseen-tx-throughput-design.md
@@ -1,0 +1,191 @@
+# Reproducing the unseen-tx block validation throughput collapse
+
+Design for a local reproduction harness and bisect for
+[issue #1379](https://github.com/bsv-blockchain/teranode/issues/1379):
+block validation drops to ~30–50 tx/s on blocks whose transactions never
+arrived via propagation.
+
+## Problem
+
+On mainnet 2026-07-29, block 959984 (27,132 txs, 23.7 MB) took 12+ minutes to
+validate on all three BSVA mainnet nodes. The same day, block 959964 —
+232,673 txs, 62.6 MB — validated in 0.2–10.6 s on the same nodes.
+
+Size is not the variable. The discriminator is the pre-check in
+`processTransactionsInLevels` (`services/subtreevalidation/check_block_subtrees.go:1009-1028`):
+when every tx in the block is already in the txMeta cache or UTXO store,
+`missed == 0` and the function returns immediately. When the block contains txs
+the node never saw via propagation, each one goes through
+`blessMissingTransaction` and throughput collapses.
+
+The cost scales with the count of *unseen* txs, not block size, and behaves
+identically on v0.15 and v0.16.0-beta-1. A block carrying ~1M never-broadcast
+txs would take roughly 6–9 hours per node. Any miner can construct one.
+
+## What code reading already settled
+
+Three of the issue's own hypotheses were resolved by reading the code, before
+any measurement.
+
+**TX_LOCKED retry backoff is not the cause.** `services/validator/Validator.go:479-510`
+caps backoff at 10/20/40 ms over 3 retries, hard-clamped at 10 attempts. Maximum
+~70 ms per tx. It cannot produce the observed ~8.7 s per level.
+
+**The issue quoted the wrong batcher numbers.** It cited
+`utxostore_spendBatcherSize = 100` and `spendBatcherDurationMillis = 100`; those
+are the Go struct defaults in `settings/settings.go:445-446`, used only when the
+key is absent. `settings.conf` overrides them to 1024 and 10. The real
+constraint on that path is `utxostore_spendBatcherConcurrency = 4`
+(`settings.conf:1298`) — four concurrent Lua batch operations.
+
+**`prefetchLevelParents` is not the root cause.** It landed in `7091af082`,
+first tagged v0.16.0-beta-1. The v0.15 nodes were equally slow, so it cannot
+explain the collapse. It remains a real latent regression: `BatchDecorate`
+(`stores/utxo/aerospike/get.go:697-742`) walks batch results serially and calls
+`GetTxFromExternalStore` inline, converting external-store reads that were
+previously parallel across the level's goroutines into serial ones.
+
+Two further observations shape the harness:
+
+**The existing histogram already exists.** `prometheusSubtreeValidationBlessMissingTransaction`
+is observed at `services/subtreevalidation/SubtreeValidation.go:412`. The issue's
+ask for "a latency histogram around `blessMissingTransaction`" is already
+satisfied in code; the gap is purely that nothing scrapes it on the mainnet
+nodes.
+
+**The validator is in-process.** `useLocalValidator = true` (`settings.conf:1255`).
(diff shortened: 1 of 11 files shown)


oskarszoon:
Closing as superseded.

#1393 merged the main O(N²) conflict-cone fix, and #1394 carries the release/v0.15 backport. This PR's final investigation confirms that same mechanism, while its description still says the root cause was not found.

Two independent changes remain worth considering as focused PRs:
- A failed 2PC unlock should not fail an already-validated transaction.
- Level-depth and per-level-duration metrics, plus the related INFO logs.

The harness and findings remain useful historical evidence, but no longer justify merging this large, stale branch.

#1379 remains open for reassessment. The smaller conflict-path cost and the cancelled HandleBlockDirect work remain unexplained.

PR #1389, opened 2026-08-03
feat(p2p): declarative per-FSM-state outbound publish gate

gokutheengineer:
**Closes bitcoin-sv/teranode#4716.**

### Problem

Outbound P2P gating on the blockchain FSM state was scattered: `isBlockchainSyncingOrCatchingUp` checks in `blockchainSubscriptionListener` and `rejectedTxHandler`, plus per-handler `ListenMode` checks, with no central gate in front of `P2PClient.Publish`. Any new code path could leak block/subtree/rejected-tx announcements during CATCHINGBLOCKS without anything failing loudly. Confirmed against current main: exactly 4 `P2PClient.Publish` sites, 3 scattered FSM checks, no violation detection.

### Fix

- `services/p2p/publish_gate.go`: declarative per-FSM-state allow-list (`outboundTopicsAllowed`) plus a single choke point `publishToNetwork` wrapping all 4 publish sites. Suppressed messages are dropped and counted in a new `teranode_p2p_publish_blocked_total{topic,fsm_state,stage}` counter (`services/p2p/metrics.go`, the service's first Prometheus metric): `stage="precheck"` counts expected skips at the handler pre-checks, `stage="chokepoint"` counts publishes that leaked past them (error-logged - this is the alertable fail-loud signal). Unknown topics fail closed; unknown FSM states (e.g. a newer blockchain enum) fall back to node_status-only; direct FSM-fetch errors fail open - though note the common degraded-client case never takes that branch, since `blockchain.Client` caches IDLE with no error as its safety fallback and thus lands in the node_status-only IDLE bucket.
- The listener and rejected-tx handler pre-checks now derive from the same map (`canSendToNetwork` / `shouldSkipNotification`), so there is one source of truth; a publish reaching the choke point in a disallowed state is a genuine leak and logs at error level.
- `ListenMode` policy (silent publishes nothing, listen-only publishes only node_status) is also enforced centrally in the choke point, in addition to the existing handler checks.
- A convention test (`TestPublishGateIsOnlyPublishCallSite`) fails if any non-test file in the p2p package contains a literal `P2PClient.Publish` call outside `publish_gate.go` (package-local, literal match - future-proofing, not a complete static guarantee; the limits are documented in the test).
- CATCHINGBLOCKS allows only node_status (as before via the scattered checks); node_status was never FSM-gated and remains allowed in every state. PeerFailure notifications still pass during catchup.
- **Behavior change**: IDLE is now node_status-only, where the old code announced blocks, subtrees, and rejected txs in IDLE (the scattered checks only ever treated CATCHINGBLOCKS as syncing). This is deliberate (review feedback): IDLE doubles as the blockchain client's "state unknown" safety fallback - `fetchAndRestoreFSMState` and the heartbeat paths store IDLE with no error when the FSM fetch fails - so a permissive IDLE would open the gate silently exactly when the client is degraded, and the docs' IDLE contract is "no network participation". It is also unobservable in practice: p2p's `Start` blocks on `WaitUntilFSMTransitionFromIdleState` before any publish path is wired, and LocalClient-backed test daemons report RUNNING.
- Improvements over the issue's suggested sketch: no FSM-state cache was added - `blockchain.Client.GetFSMCurrentState` is already a push-updated local read, so the gate adds no gRPC round-trips; and notification-level pre-checks were kept (map-driven) so catchup does not pay a `GetBlockHeader` call per caught-up block only to have the publish dropped.
- Cache-and-replay from the Java POC was not ported, per the issue (pull-based catchup covers it).
- Docs: allow-list documented in `stateManagement.md`, metric in `prometheusMetrics.md`, file listing in `p2p.md`.

### Residual

- The old 15s startup retry in `isBlockchainSyncingOrCatchingUp` is deliberately not carried into `currentFSMState` (single attempt, 5s ceiling, fail open); p2p's `Start` already waits for the FSM to leave IDLE, so the window is largely closed and the worst case is brief log noise during blockchain-service startup.
- While the blockchain client is degraded (reads as IDLE), Kafka-driven rejected-tx announcements are suppressed for the duration; block/subtree gossip arrives over the same broken stream so there is little to announce anyway. Distinguishing "genuinely idle" from "state unknown" belongs upstream in `blockchain.Client`.

- `invalidSubtreeHandler` keeps its `isBlockchainSyncingOrCatchingUp` check: it gates inbound ban-score processing, not an outbound publish. The issue's optional "defer (not drop) ban-score events during sync" is not implemented.
- `blockchain.Mock` is used for the state-matrix tests because `LocalClient` cannot be driven into CATCHINGBLOCKS (it hardcodes RUNNING); a `LocalClient`-based happy-path test is included, and this mirrors existing practice in `sync_coordinator_test.go`.

### Tests

New `services/p2p/publish_gate_test.go`: allow-list coverage of every FSM state, per-state × per-topic publish matrix, blocked-counter increment, unknown-topic fail-closed, unknown-FSM-state fallback, listen-mode matrix, fail-open on FSM error, nil blockchain client, LocalClient happy path, notification-skip matrix (incl. PeerFailure during catchup and non-announcement types), and the choke-point convention scan.

```
go vet ./services/p2p/
```

Full p2p package green with `-race`.


ctnguyen (changes requested):
Review at commit [7be9650](https://github.com/bsv-blockchain/teranode/commit/7be96504cb85a8a4b70aee4a393ea0f2586b4268)

Nice piece of work — collapsing three scattered `isBlockchainSyncingOrCatchingUp` checks into one declarative table and one choke point is the right shape for this problem, and I like that you resisted the issue's cache-and-replay suggestion instead of porting it wholesale. I verified the choke-point claim holds: `Publish` is the only outbound method on the message-bus client interface and the only two call sites left in the tree are inside `publish_gate.go`, so the gate really is complete for pubsub. `TestOutboundTopicsAllowedCoversAllFSMStates` is a genuinely good guard, and the mock-vs-`sqlitememory` exception is justified here — `LocalClient.GetFSMCurrentState` hardcodes RUNNING (`services/blockchain/LocalClient.go:423`) so the state matrix is not expressible any other way, and you paired it with a `LocalClient` happy path. The main thing I'd push back on is the IDLE row, which I think is wrong for a reason that isn't obvious from the p2p package alone.

## Blocking issues

### ChiR1 — IDLE is the blockchain client's "state unknown" fallback, so a fully permissive IDLE row opens the gate exactly when it should close

**Problem:** `outboundTopicsAllowed[FSMStateType_IDLE]` allows block, subtree and rejected-tx (`services/p2p/publish_gate.go:33-38`), justified by the comment "in IDLE no blocks or subtrees are being processed anyway". That premise doesn't hold, because IDLE is not only a real FSM state — it is also the value the blockchain client *stores locally when it does not know the state*:

- `fetchAndRestoreFSMState` runs on every subscription establishment, including the first one, and on failure does `c.fmsState.Store(&idleState)` with the comment "setting to IDLE for safety" (`services/blockchain/Client.go:1553-1568`, caller at `:1396`).
- The stale/absent-heartbeat paths do the same (`services/blockchain/Client.go:1462-1475`).

`GetFSMCurrentState` then returns that cached IDLE with `err == nil` (`services/blockchain/Client.go:1825-1829`), so `currentFSMState` never surfaces an error and neither the fail-open warning nor the blocked counter fires. Concrete sequence: p2p starts, `Server.Start` clears `WaitUntilFSMTransitionFromIdleState` (a *server-side* wait, so it returns as soon as the real FSM is non-IDLE), the client's initial FSM fetch blips and caches IDLE, the real FSM is CATCHINGBLOCKS — and the gate publishes everything until the next FSM *transition* repopulates the cache. The Kafka-driven `rejectedTxHandler` leaks immediately; block/subtree leak as soon as the subscription recovers without a transition. The second IDLE case is a deliberate `Stop`, and the very file this PR edits says of IDLE: "The node is not participating in the network in any way", "❌ Relay blocks" (`docs/topics/architecture/stateManagement.md:108-123`) — so the new paragraph at `:170` documents the CATCHINGBLOCKS row while the code silently contradicts the IDLE contract three sections above it.

**Why it matters:** This is the exact leak the PR exists to prevent, and it now fires silently — no error log, no `publish_blocked_total` increment — because the value looks like a legitimate permissive state rather than an error. Codifying it in a declarative table and a docs page makes it the *intended* policy rather than an accident, which is harder to walk back later. The stated reason for keeping IDLE permissive ("test daemons without an explicit FSM state sit in IDLE") doesn't survive checking: `Server.Start` blocks on `WaitUntilFSMTransitionFromIdleState` before any publish path starts (`services/p2p/Server.go:634`), and `LocalClient`-backed daemons report RUNNING — so nothing that publishes today can be observing a genuinely-idle FSM. Tightening IDLE is cheaper than it looks. (Background: [bitcoin-sv/teranode#4716](https://github.com/bitcoin-sv/teranode/issues/4716) asks for a fail-loud allow-list; a state that both means "stopped" and "unknown" is the one row where fail-open defeats that.)

**Fix:** Treat IDLE like the unknown-state fallback and fix the comment plus the docs paragraph to state the full table, not just the CATCHINGBLOCKS row:

```go
// IDLE means either a deliberate Stop (docs: the node must not participate in
// the network at all) or "state unknown" — the blockchain client stores IDLE
// as its safety fallback when the FSM fetch or the heartbeat fails. Either
// way, publish nothing but node_status.
blockchain_api.FSMStateType_IDLE: {
    topicKindNodeStatus: true,
},
```

If you'd rather not change behaviour in this PR, then at minimum distinguish the two meanings before merge — e.g. have the gate consult the client's subscription health, or keep a "last state seen with confidence" — because as written the gate's guarantee is void whenever the blockchain client is degraded, and `TestPublishToNetworkFailsOpenOnStateError` tests a path (`err != nil`) that the production client almost never takes.

## Non-blocking issues

### ChiR2 — no test covers the cached-IDLE fallback, which is the failure mode behind ChiR1

**Problem:** The suite exercises two FSM-read outcomes: an explicit error (`TestPublishToNetworkFailsOpenOnStateError`) and a mock that returns a state cleanly (`mockBlockchainInState`). Neither reproduces production's third outcome — a `blockchain.Client` whose FSM restore failed, which then serves a cached IDLE with `err == nil` indefinitely. Because `blockchain.Mock` returns whatever the test hands it, an IDLE row in `TestPublishToNetworkPerState` is indistinguishable from that degraded case, so the matrix currently *asserts* the leak is correct rather than catching it. Nothing in the p2p package pins the client-side contract the gate depends on either, so a future change to the client's fallback (say, to RUNNING "so the node keeps working") would silently widen the gate with every p2p test still green.

**Why it matters:** The gate's whole guarantee rests on an undocumented cross-package contract about what `GetFSMCurrentState` returns when the client is degraded. Once ChiR1 is fixed, flipping the IDLE matrix rows proves the map, but not that the map is the *right* map for the degraded case — and that is the regression worth locking down, since it spans two services and neither package's tests currently mention it.

**Fix:** Two focused tests. In `services/blockchain`, pin the contract using the hook that already exists on `mockBlockClient` (`fnGetFSMCurrentState`, `services/blockchain/mock.go:939` and `:1284-1294`, constructed directly in `client_test.go` today):

```go
// A failed restore must leave the cache at a state the callers treat as
// "do not gossip", and must keep reporting it without an error.
state, err := client.GetFSMCurrentState(ctx)
require.NoError(t, err)
require.Equal(t, blockchain_api.FSMStateType_IDLE, *state)
```

Then in `services/p2p`, assert the consequence rather than the mechanism — that the state the client falls back to is not permissive:

```go
require.Equal(t, nodeStatusOnly, allowedTopicsForState(blockchain_api.FSMStateType_IDLE),
    "IDLE is the blockchain client's degraded-state fallback; it must not permit gossip")
```

### ChiR3 — one counter conflates routine pre-check skips with genuine choke-point leaks

**Problem:** `canSendToNetwork` and `publishToNetwork` increment the same `prometheusP2PPublishBlocked` with the same labels (`services/p2p/publish_gate.go:118` and `:187`). The pre-check path is the normal, expected path during catchup — every block and subtree notification bumps the counter without any publish ever having been attempted — while the choke-point path is described in the PR as "a genuine leak". It also over-counts: `canSendToNetwork(topicKindRejectedTx)` runs *before* the `m.PeerId != ""` external-rejection early return (`services/p2p/Server.go:851` vs `:868`), so Kafka messages that would never have been published at all are counted as blocked publishes.

**Why it matters:** `teranode_p2p_publish_blocked_total > 0` is then unusable as a leak alert — it climbs into the tens of thousands during any normal initial block download — and the docs describe it as "outbound P2P publishes dropped" (`docs/references/prometheusMetrics.md:199`), which is not what the pre-check path measures. The only place the two are distinguishable is the log level (`Debugf` vs `Errorf`), which is not what operators alert on.

**Fix:** Add a `stage` label so the two are separable, and keep the docs honest about what is counted:

```go
[]string{"topic", "fsm_state", "stage"} // stage: "precheck" | "chokepoint"
```

### ChiR4 — a benign FSM transition between pre-check and publish logs at error level as a leak

**Problem:** Each notification reads the FSM state twice — once in `shouldSkipNotification` and again inside `publishToNetwork` — with `GetBlockHeader` and JSON marshalling in between (`services/p2p/Server.go:1188-1213`). A RUNNING → CATCHINGBLOCKS transition landing in that window makes the pre-check pass and the choke point drop, producing `[publishToNetwork] dropping block publish: not allowed in FSM state CATCHINGBLOCKS` at **error** level plus a counter bump. The same happens between the block publish and the nested `handleNodeStatusNotification` call at `:1218`.

**Why it matters:** The design's contract is "an error-level drop means a code path bypassed the allow-list". This race makes that contract false on every catchup entry, so the signal gets learned as noise and a real bypass gets ignored.

**Fix:** Keep the choke point's own read authoritative — it must stay the last word, or a stale pre-check verdict would let a block through after the FSM has already moved to CATCHINGBLOCKS. Pass the pre-check's observed state down as *classification metadata only*, and branch on whether it differs from the authoritative read:

```go
// observedState is used ONLY to classify the drop. The read below stays
// authoritative, so a transition after the pre-check still blocks the publish.
state, err := s.currentFSMState(ctx)
...
if !allowedTopicsForState(state)[kind] {
    if state != observedState {
        // benign: FSM moved between the caller's check and here
        s.logger.Warnf("[publishToNetwork] dropping %s publish: FSM moved %s -> %s", kind, observedState, state)
    } else {
        // the caller claims it checked this same state and allowed it: real bypass
        s.logger.Errorf("[publishToNetwork] dropping %s publish: not allowed in FSM state %s", kind, state)
    }
    return nil
}
```

Callers with no pre-check keep the existing entry point and the existing error-level treatment.

### ChiR5 — an unregistered topic is dropped but reported as success

**Problem:** `publishToNetwork` fails closed on an unknown topic and returns `nil` (`services/p2p/publish_gate.go:160-166`), so the caller sees a successful publish. Combined with `TestPublishGateIsOnlyPublishCallSite` forcing every publish through the gate, adding a fifth pubsub topic and forgetting to extend `topicKindForName` silently disables it in production with only an error log; no compile-time check and no test catches it.

**Why it matters:** Fail-closed is the right default, but reporting the failure as success removes the caller's only chance to react and turns a one-line omission into a silent feature outage. Today all four topic fields are mapped, so this is latent rather than live.

**Fix:** Return an error so it propagates into the existing `errors.NewError("... publish error", err)` wrappers, and add a test that walks the topic fields to assert every pubsub topic maps to a kind:

```go
return errors.NewProcessingError("[publishToNetwork] topic %s is not in the outbound allow-list", topicName)
```

### ChiR6 — the choke-point guard is a substring scan and is easy to defeat

**Problem:** `TestPublishGateIsOnlyPublishCallSite` globs `*.go` in the package directory and greps for the literal `.P2PClient.Publish(`, skipping `publish_gate.go` by filename. It misses `c := s.P2PClient; c.Publish(...)`, a publish from a subpackage or another package holding the client, and it silently stops guarding anything if `publish_gate.go` is ever renamed or split.

**Why it matters:** This test is the enforcement mechanism the PR relies on to keep the invariant true over time; a guard that a refactor can disable without failing is weaker than it reads.

**Fix:** Either parse with `go/ast` and match the selector on the resolved receiver type, or push it into the linter — a `forbidigo` pattern on `\.Publish\(` with `publish_gate.go` excluded, or a `depguard`-style rule — so it applies repo-wide rather than to one directory's filenames.

### ChiR7 — redundant state reads allocate a timer per call on the announcement hot path

**Problem:** `currentFSMState` builds a `context.WithTimeout` on every call (`services/p2p/publish_gate.go:93`) for what, once the client's cache is warm, is an `atomic.Pointer.Load()` that cannot block. It is called two to three times per blockchain notification and twice per rejected-tx Kafka message. `handleNodeStatusNotification` also reads the state directly at `services/p2p/Server.go:1272`, then `publishToNetwork` reads it again a few lines later in the same call.

**Why it matters:** Each `WithTimeout` allocates a context and arms a runtime timer; at subtree-announcement and rejected-tx rates this is pure overhead on a path that is otherwise a single atomic load, and the timeout only ever helps on the cold-cache gRPC fallback. Small, but it's on the hot path.

**Fix:** Hoist the bare literal into a named constant (`const fsmStateReadTimeout = 5 * time.Second`), and remove the genuinely redundant read — `handleNodeStatusNotification` already has the state a few lines above its publish, so pass it in as the observed state from ChiR4 rather than fetching it twice. Don't collapse the choke point's own read to save a call: per ChiR4 that one has to stay authoritative.

## Recap

| ChiR1 | IDLE permissive when state unknown | YES | 70% |
| ChiR2 | Cached-IDLE fallback untested | | 45% |
| ChiR3 | Blocked counter conflates pre-check and leak | | 40% |
| ChiR4 | Benign FSM race logs as leak | | 30% |
| ChiR5 | Unknown topic drop returns success | | 25% |
| ChiR6 | Choke-point guard is a grep | | 15% |
| ChiR7 | Redundant state reads, timer per call | | 10% |

freemans13 (commented):
Review of the publish gate. No blocking defects — the gating logic is correct for all three FSM states, the choke point genuinely holds, and the package passes under `-race`. Five inline comments below; two of them (the counter semantics and the IDLE/fail-open asymmetry) I'd like your view on before merge, the rest are take-or-leave.

### What I verified rather than took on trust

| Check | Result |
|---|---|
| `SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./services/p2p/...` | ok, 13.4s |
| `go vet ./services/p2p/` | clean |
| `golangci-lint run ./services/p2p/...` | 1 issue, in `mock.go` — pre-existing, not touched by this PR |

And the structural claims, which all hold:

- **The choke point is real.** Grepping the whole repo for `.Publish(` on the p2p client returns exactly two sites, both inside `publish_gate.go`.
- **The retained old check is correctly scoped.** `isBlockchainSyncingOrCatchingUp` survives only at `Server.go:818` for `invalidSubtreeHandler`, which gates inbound ban-score processing rather than an outbound publish.
- **The safety argument for IDLE holds.** `Start` blocks on `WaitUntilFSMTransitionFromIdleState` at `Server.go:634`, before the rejected-tx consumer (649), the subscription listener (718) and the node-status ticker (738). No publish path is wired while the FSM is still IDLE.
- **node_status really does keep flowing during catchup**, via its own 10s ticker in `publishNodeStatus` (`Server.go:1226`), independent of the notification listener that `shouldSkipNotification` gates.

One framing note: the IDLE restriction is a genuine behaviour change, not just a tightening. The old helper treated *only* `CATCHINGBLOCKS` as syncing (`server_helpers.go:827`), so in IDLE the node previously did announce blocks, subtrees and rejected txs. Worth saying that plainly in the description, since it currently reads as a review tweak.

### Two informational notes, no action needed

**Lost retry patience, deliberately.** The old `isBlockchainSyncingOrCatchingUp` retried for up to 15s specifically to cover "blockchain service isn't ready yet" (`server_helpers.go:795-825`); `currentFSMState` makes one attempt with a 5s ceiling. Since both now fail open, the only consequence is some extra log noise during blockchain-service startup — notifications get processed, then `GetBlockHeader` fails and logs. And `Start` already waits for the FSM to leave IDLE before wiring the listener, so that window is largely closed. Recording it so the removed retry doesn't look accidental to a future reader.

**Cosmetic.** The new table in `docs/references/prometheusMetrics.md` has a separator row a couple of characters wider than its header row. Renders fine in GFM — pure nit.

### What I liked

`TestOutboundTopicsAllowedCoversAllFSMStates` is the right kind of test: it turns "someone adds an FSM state and forgets the gate" from a silent leak into a build failure, and separately asserts node_status stays allowed everywhere. Fail-closed on unknown topics and node-status-only for unknown states are both the conservative choice. And keeping the notification-level pre-checks rather than letting everything reach the choke point avoids a `GetBlockHeader` gRPC call per caught-up block during IBD — a real saving on exactly the path where it matters.

services/p2p/publish_gate.go
```diff
@@ -0,0 +1,192 @@
...
+	state, err := s.currentFSMState(ctx)
+	if err != nil {
+		s.logger.Warnf("[canSendToNetwork] allowing %s publish, error getting blockchain FSM state: %v", kind, err)
+		return true
+	}
+
+	if !allowedTopicsForState(state)[kind] {
+		initPrometheusMetrics()
+		s.logger.Debugf("[canSendToNetwork] %s not allowed in FSM state %s", kind, state.String())
+		prometheusP2PPublishBlocked.WithLabelValues(string(kind), state.String()).Inc()
```
freemans13: **The counter measures two different things under one name, which blunts the fail-loud property.**

The gate's headline benefit is that a publish reaching the choke point in a disallowed state is a genuine leak — hence the error-level log and the counter. But this pre-check increments the *same* series, once per block and subtree notification.

During any IBD catchup that dominates completely, so `teranode_p2p_publish_blocked_total{topic="block", fsm_state="CATCHINGBLOCKS"}` climbing is the normal case, and the leak you actually want to detect is invisible inside it. The two can also double-count the same logical event if the FSM flips between this pre-check and the publish.

You already made exactly this argument in the default branch of `shouldSkipNotification` below — *"without polluting the blocked-publish counter"*. It applies equally to the Block and Subtree branches, which do pollute it via this function.

Suggestion: either add a `stage` label (`precheck` / `chokepoint`) so the two are separable, or count only at the choke point and give pre-check skips their own `notifications_skipped_total`. As it stands the metric's help string says "publishes dropped", which isn't what it counts.

services/p2p/publish_gate.go
```diff
@@ -0,0 +1,192 @@
...
+			s.logger.Debugf("[publishToNetwork] dropping %s publish in listen mode %s", kind, mode)
+			return nil
+		}
+	}
+
+	state, err := s.currentFSMState(ctx)
+	if err != nil {
+		s.logger.Warnf("[publishToNetwork] allowing %s publish, error getting blockchain FSM state: %v", kind, err)
+		return s.P2PClient.Publish(ctx, topicName, msgBytes)
+	}
```
freemans13, line 184: **This fails open on an error path that almost never fires, while the path that does fire fails closed.**

The intent here is right and it's tested — a briefly unreachable blockchain service shouldn't silence the node. The catch is that `blockchain.Client` hardly ever returns that error. On a subscription break with a missing or stale heartbeat it stores IDLE (`Client.go:1467` and `:1473`), and on a failed post-reconnect fetch it stores IDLE too (`Client.go:1561`). `GetFSMCurrentState` then reads the cached value and returns **IDLE with a nil error** (`Client.go:1826-1829`).

So the degraded-client case never reaches this branch — it lands in the IDLE bucket, which under this PR is node_status-only. That's precisely the silencing this branch exists to prevent.

To be fair about the real exposure: it's narrower than it first looks. Block and subtree notifications arrive over the same stream that just broke, so there's little to announce during the outage anyway, and `handleBlockNotification` needs a working `GetBlockHeader` regardless. Recovery is automatic on the 1s reconnect loop. The genuinely exposed path is **rejected transactions** — Kafka-driven, so they keep arriving independently of the blockchain gRPC stream, and their announcements get dropped for the duration of the blip.

No code change strictly required. But I'd make the comment and the `stateManagement.md` paragraph say plainly that a degraded blockchain client reads as IDLE and therefore stops gossip, rather than leaving a reader thinking this fail-open covers it. If you want the fail-open to actually apply, the fix belongs upstream in `blockchain.Client` — distinguishing "genuinely idle" from "state unknown" instead of overloading IDLE for both.

services/p2p/publish_gate.go
```diff
@@ -0,0 +1,192 @@
...
+// read rather than a gRPC round-trip. A nil blockchain client reports
+// RUNNING so setups without a blockchain stay permissive, matching the
+// previous behavior of isBlockchainSyncingOrCatchingUp. The returned state
+// is meaningless when err is non-nil.
+func (s *Server) currentFSMState(ctx context.Context) (blockchain_api.FSMStateType, error) {
+	if s.blockchainClient == nil {
+		return blockchain_api.FSMStateType_RUNNING, nil
+	}
+
+	fsmCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
freemans13, line 92: Nit, no change requested. Every state read allocates a 5s timeout context, and the rejected-tx path calls it twice — once via `canSendToNetwork` at `Server.go:851` and again inside `publishToNetwork` at `:894`. The read is normally a local atomic load, so the timer is pure overhead on a Kafka-rate hot path.

The double call is defensible (the pre-check saves the marshal/unmarshal work in between), and the real fix — only paying for a context when the cached state is absent — belongs in `blockchain.Client` rather than here. Mentioning it for completeness.

services/p2p/publish_gate_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	for _, f := range files {
+		if strings.HasSuffix(f, "_test.go") || f == "publish_gate.go" {
+			continue
+		}
+
+		src, err := os.ReadFile(f)
+		require.NoError(t, err)
+		require.NotContains(t, string(src), ".P2PClient.Publish(",
+			"%s calls P2PClient.Publish directly; route it through publishToNetwork so the FSM allow-list applies", f)
```
freemans13, line 80: **This test is narrower than the PR description claims.**

`filepath.Glob("*.go")` is the package directory only, not recursive, and the match is a literal substring on `.P2PClient.Publish(`. A publish added from another package, or through a local variable (`c := s.P2PClient; c.Publish(...)`), slips straight through. The description says it fails "if any non-test file calls `P2PClient.Publish` outside `publish_gate.go`", which is broader than what it does.

The invariant genuinely holds today — I grepped and found no other call sites anywhere in the repo — so this is future-proofing, not a live gap. Either note the limitation in the test comment, or widen the match to `.Publish(` within the package with the gate file allow-listed.

services/p2p/publish_gate_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	return &Server{
+		logger:              ulogger.TestLogger{},
+		blockchainClient:    blockchainClient,
+		P2PClient:           p2pClient,
+		blockTopicName:      "test-block",
+		subtreeTopicName:    "test-subtree",
+		rejectedTxTopicName: "test-rejected-tx",
+		nodeStatusTopicName: "test-node_status",
+	}, p2pClient
```
freemans13, line 38: The helper leaves `settings` nil, which is why `publishToNetwork` needs its `if s.settings != nil` guard. In production `NewServer` always populates it, so that guard has no real trigger — it's defensive code shaped by the test rather than by the code path.

The practical consequence is that the whole FSM state matrix in `TestPublishToNetworkPerState` skips the listen-mode branch entirely. Setting `settings: &settings.Settings{P2P: settings.P2PSettings{ListenMode: settings.ListenModeFull}}` here would let the nil guard go and make the matrix exercise the real path.

ctnguyen (approved):
Review at commit [486d5f8](https://github.com/gokutheengineer/teranode/commit/486d5f82873209a53fa42406134853c366fac7c3)

The one blocking point is resolved cleanly. `486d5f8` restricts IDLE to `node_status`-only across the code, the docs, and the test matrix, and the accompanying comment now states exactly why IDLE is treated as "stopped or unknown" rather than permissive — which was the crux of the earlier objection. That was the change that had to happen before merge, and it happened in the right place with the right reasoning. Everything else from the prior pass was non-blocking and stands as it was; the notes below are follow-ups, not gates.

## Status of previous points

Only one commit landed since the previous review, and it addressed ChiR1 alone, so the non-blocking items are largely untouched. ChiR2 is split: the p2p-side consequence is now pinned by the matrix (the IDLE rows assert non-permissive), but the cross-package half — a blockchain-side test proving `GetFSMCurrentState` serves a cached IDLE with `err == nil` after a failed restore — is still absent, which was the regression ChiR2 most wanted locked down. freemans13 independently raised overlaps with ChiR1/ChiR2 (degraded client caches IDLE, fail-open branch never sees it), ChiR3 (counter measures two things), ChiR7 (timer per read), and ChiR6 (grep-based guard) at the old head; those remain live below or in this table.

| ChiR1 | IDLE permissive when state unknown | YES | FIXED |
| ChiR2 | Cached-IDLE fallback untested | | PARTIAL FIXED |
| ChiR3 | Counter conflates pre-check and leak | | NOT FIXED |
| ChiR4 | Benign FSM race logs as leak | | NOT FIXED |
| ChiR5 | Unknown-topic drop returns success | | NOT FIXED |
| ChiR6 | Choke-point guard is a grep | | NOT FIXED |
| ChiR7 | Redundant reads, timer per call | | NOT FIXED |

## Non-blocking issue

### ChiR8 — the nil-`settings` guard in the choke point exists only for the test helper, and the state matrix never exercises the listen-mode branch
*Same as [@freemans13's point (10-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1389#discussion_r3751214800).*

**Problem:** `publishToNetwork` wraps its listen-mode check in `if s.settings != nil` (`services/p2p/publish_gate.go:170`), but a production `Server` always has settings — `NewServer` takes `tSettings` and stores it. The only caller that reaches this path with `s.settings == nil` is the test helper `newGateTestServer`, which builds a `Server` literal without a `settings` field (`services/p2p/publish_gate_test.go:21-40`). As a result `TestPublishToNetworkPerState`, `TestPublishToNetworkUnknownFSMStateFallsBackToNodeStatusOnly`, and the blocked-counter test all run with the listen-mode branch dead, so the interaction of listen mode with FSM gating is only ever seen in `TestPublishToNetworkListenMode`, which sets `settings` by hand and pins RUNNING.

**Why it matters:** A defensive `nil` check that is only reachable from tests reads as a real production guard, and the matrix that looks like it covers "every state × every topic" silently skips the mode dimension on that helper. If listen-mode and FSM gating ever interact wrongly (e.g. an ordering change between the two checks), the broad matrix stays green because it never populates settings.

**Fix:** Populate `settings` in `newGateTestServer` (e.g. `&settings.Settings{P2P: settings.P2PSettings{ListenMode: settings.ListenModeFull}}`) and drop the `if s.settings != nil` guard from `publishToNetwork`, so the choke point runs the same code in tests as in production and the matrix exercises the full path.

freemans13 (approved):
Approving. This does what it says: three scattered "are we syncing?" checks collapse into one declarative allow-list plus a single choke point in front of all four `P2PClient.Publish` sites, and a leaked publish becomes loud instead of silent. I checked the description's load-bearing claims against the code rather than taking them on trust — the four publish sites (only `publish_gate.go:183`/`:193` remain; the other `.Publish(` hits in the package are Kafka producers), the cached FSM read (`blockchain/Client.go:1825-1837`), `IDLE` as the client's no-error safety fallback (`Client.go:1467`, `:1473`, `:1561`), and the startup ordering that makes `IDLE` unobservable (`WaitUntilFSMTransitionFromIdleState` at `Server.go:634` precedes the rejected-tx consumer at `:649`, the subscription listener at `:718` and `publishNodeStatus` at `:738`). All hold. The residual section is unusually honest about what it costs.

One non-blocking thing worth considering, because it degrades the signal this PR exists to provide.

**The chokepoint stage will fire falsely on ordinary state transitions.** The pre-check (`shouldSkipNotification` → `canSendToNetwork`) and the choke point each read the FSM state independently, and real work happens in between — `handleBlockNotification` makes a `GetBlockHeader` gRPC call and marshals JSON before reaching the publish. If the FSM moves `RUNNING` → `CATCHINGBLOCKS` in that window, which is exactly what entering catchup is, `publishToNetwork` drops the message, counts it as `stage="chokepoint"` and logs at error level saying it leaked past the pre-checks. Nothing leaked; the pre-check was simply correct when it ran. Since the description designates `stage="chokepoint"` as the alertable fail-loud signal, every entry into catchup with notifications in flight produces a burst that looks identical to a genuine leaking code path — and a signal that cries wolf on every catchup stops being useful for what it was built to catch.

Cheapest fix: thread the pre-checked state (or just a `preChecked bool`) into `publishToNetwork`, and when the choke point's state differs from the pre-checked one, log at warn and count under a third stage, e.g. `raced` — leaving the `chokepoint` bucket as a clean "this genuinely had no pre-check" signal. If you'd rather not change signatures, documenting the race in the metric help text and `prometheusMetrics.md` would at least tell anyone writing an alert to expect a low-rate baseline around transitions.

Minor notes, take or leave:

- `publish_gate.go:115` — `initPrometheusMetrics()` sits inside the blocked branch of `canSendToNetwork`, but at function entry in `publishToNetwork`. Both are safe today; the asymmetry is the kind of thing a later edit trips over into a nil-map panic.
- `publish_gate.go:162` — the unknown-topic drop labels both `topic` and `fsm_state` as `"unknown"`, so the metric can't say which topic leaked. The error log carries the name and keeping the label bounded is right for cardinality, so this is just noted.
- Coverage of the gate itself is thorough (every state, every topic, listen modes, fail-open, fail-closed, the skip matrix). What isn't covered is that the call sites actually reach it — the convention test is a textual scan whose limits you document honestly. A test driving a `Block` notification through `blockchainSubscriptionListener` with the FSM mocked into `CATCHINGBLOCKS`, asserting the P2P mock saw no `Publish`, would close the gap the scan can't. Relatedly, `TestPublishToNetworkWithLocalBlockchainClient` can only ever exercise `RUNNING`, since `LocalClient.GetFSMCurrentState` hardcodes it.

Testing note on my side: I could not run the suite locally (my build machine is out of disk, so the p2p test binary fails at link). This approval rests on reading the code plus CI, which is green on this head commit — lint, unit tests, all seven sequential shards, all three smoketest shards, the benchmark suite and the security scans.

PR #1392, opened 2026-08-04
fix(subtreevalidation): walk the conflicting descendant cone once, not once per member (#1391)

### LINKED ISSUE #1391: utxo: GetConflictingChildren walks the full descendant spend cone — mainnet fleet wedged at 960,827
## Summary

Every mainnet Teranode stalled at height **960,827** on 2026-08-04 while SV Node advanced normally (960,830+). All versions in the fleet are affected — v0.15.6, v0.15.8 and v0.16.0 — because the defect is in code identical across all of them and on `main`.

Block **960,828** (`00000000000000000ef4cd35a6fdb54181498e28f881ac2849530de30fac704a`, 2391 txs) contains tx `005a1b5877c31e0203cac356bc303966ab356c7e0695a8d480d55a12530f85d7`, which the local UTXO store flags `conflicting`. Blessing it enters `checkCounterConflictingOnCurrentChain` → `utxo.GetConflictingChildren`, which does not complete: it ran **26m21s** and ended only when its context was cancelled. The block never reaches block validation, the legacy peer message handler is held until the 10-minute `block` timeout, the peer is dropped, the block is re-fetched, and the cycle repeats indefinitely.

## Observed behaviour

`teranode-mainnet-eu-1`, all timestamps UTC:

```
10:34:46.257  legacy  HandleBlockDirect [00000000…704a 960828] 2391 txs, peer 51.222.249.3:8333
10:34:46.268  legacy  extendTransactions DONE in 46.515633ms
10:34:46.315  legacy  createSubtrees created 1 subtrees
10:34:46.457  legacy  writeSubtree [16f61cff…768a] DONE in 141.536062ms
10:34:46.457  legacy  checkSubtreeFromBlock [16f61cff…768a]        ← gRPC into subtree-validator
10:34:46.465  stval   [CheckSubtree] Processing priority subtree message for 16f61cff…768a from legacy
10:34:46.484  stval   WARN stored tx meta maybe too big for txmeta cache, size: 43703, parent hash count: 102
10:34:46.544  stval   WARN [blessMissingTransaction][…/16f61cff…768a][005a1b58…85d7] transaction is conflicting
              (26 minutes, no further log lines for this subtree)
11:01:08.048  stval   ERROR [validateSubtree][16f61cff…768a] failed to bless missing transaction: 005a1b58…85d7
                        → failed to check counter conflicting tx on current chain
                        → [checkCounterConflictingOnCurrentChain][005a1b58…85d7] failed to get child transactions
                        → UNKNOWN (0): context canceled
11:01:08.050  stval   ERROR [checkSubtree] … DONE in 26m21.592467684s with error: (as above)
11:01:10.xxx  stval   [checkSubtree] called for subtree 16f61cff…768a … 960828     ← loop restarts
```

Legacy side, same block, showing the retry loop:

```
10:44:46  Disconnecting (51.222.249.3:8333) reason: Timeout processing message 'block' from peer, waited 10m0s
10:45:26  [startSync] Syncing from block height 960827 to block height 960828 using peer 15.235.232.121:8333
10:55:28  Disconnecting (15.235.232.121:8333) reason: Timeout processing message 'block' from peer, waited 10m0s
```

The node is otherwise healthy throughout — block assembly keeps producing subtrees, p2p gossip and the asset service are unaffected. Only the block-advancing path is wedged.

`005a1b5877c31e0203cac356bc303966ab356c7e0695a8d480d55a12530f85d7` on chain: 6 inputs, 1 output, 726 bytes, `nLockTime=0`, all inputs `nSequence=0xffffffff`. Nothing unusual about the transaction itself.

## Root cause

`GetConflictingChildren` BFS-walks **`SpendingDatas`** — every spender of every output — in addition to `ConflictingChildren`:

- `stores/utxo/process_conflicting.go:1092-1110` (v0.15.8)
- `stores/utxo/process_conflicting.go:1038` (`main`, unchanged in substance)

```go
if txMeta.SpendingDatas != nil {
    for _, spendingData := range txMeta.SpendingDatas {
        if spendingData != nil {
            child := *spendingData.TxID
            if _, ok := visited[child]; !ok {
                visited[child] = struct{}{}
                nextLevel = append(nextLevel, child)   // ← recurses into ordinary spenders
            }
        }
    }
}
```

So the walk is not bounded to the conflicting sub-DAG — it traverses the entire descendant spend cone of the counter-conflicting transaction across chain history. The `visited` map makes it cycle-safe but does nothing to bound its size.

Two aggravating factors:

1. The per-level `errgroup` (`process_conflicting.go:1024`) has **no `SetLimit`**, unlike comparable errgroups elsewhere in the codebase — so each level fans out one unbounded batch of Aerospike `Get`s.
2. There is no node budget, depth cap or deadline on the walk. It terminates only on context cancellation, which is why it surfaces as a silent 26-minute stall rather than an error.

Call path:

```
subtreevalidation.blessMissingTransaction            services/subtreevalidation/SubtreeValidation.go:448
  → checkCounterConflictingOnCurrentChain            services/subtreevalidation/SubtreeValidation.go:467
    → utxo.GetCounterConflictingTxHashes             stores/utxo/process_conflicting.go:1126
    → utxo.GetConflictingChildren                    stores/utxo/process_conflicting.go:1002
```

Trigger condition: any block containing a transaction the local UTXO store has flagged `conflicting`, where the counter-conflicting transaction has a large descendant cone. Deterministic given the same store state, which is why the whole fleet stopped on the same block.

## Impact

- P0 for mainnet operation: node cannot advance past the block, and the condition is self-perpetuating across restarts (verified — the 10:45 restart of the sync cycle re-wedged identically).
- Fleet-wide, version-independent: v0.15.6, v0.15.8, v0.16.0 all stalled at 960,827.
- Each retry cycle burns ~26 minutes and issues an unbounded volume of Aerospike reads.
- A restart is not a workaround: the block is re-fetched and re-wedges.

## Proposed fix

1. **Bound the recursion to the conflicting sub-DAG.** Keep ordinary spenders in the returned result set so callers' frozen-child checks still fire on direct children, but only enqueue `ConflictingChildren` into `nextLevel`. This is the behavioural fix; it is consensus-adjacent and wants review from someone who owns the frozen/mined semantics of `checkCounterConflictingOnCurrentChain`.
2. **Add a safety net regardless of 1.** `SafeSetLimit` on the per-level errgroup, plus a node/depth budget that fails closed with a distinct error when exceeded. On its own this only converts a 26-minute silent wedge into a fast deterministic block rejection — it does not restore progress — but it stops the class of failure from ever presenting as a stall again.
3. **Regression test:** a conflicting transaction whose counter-conflicting transaction has a wide/deep descendant cone must be blessable in bounded time.

## Open question, worth settling before fixing

Is `005a1b58…85d7` genuinely conflicting, or is the flag spurious — a stale or ghost `spendingData` on one of the six parent outputs? If the flag is wrong, the walk explosion is a symptom and the primary defect is wherever the flag was set. Determining this needs the UTXO record for the transaction and the recorded spender of each of its parent outputs; the asset API returned `401 Authentication required` during triage, so it is unresolved.

Related but distinct: #1214 (tx record deleted while parent `SpendingDatas` still reference it), #1228 (`SetConflicting` descendant-scan race), #645 (prior production incident — conflicting flag not cleared after being mined), #1379 (block validation throughput collapse).

## Incidental findings from the same log window

Not the cause of the stall, filed here only so they are not lost:

- `legacy/peer_server.go:1597` — recovered `nil pointer dereference` in `pushTxMsg`, repeating for txs `628835a6f68d89b8099264673dcee8989b088300853ec23e670cde121d6b5426` and `a8fe823ede6b3f7b7c9da06d82ae293fd294abbc56cc2228ca257ae867da630d` from 10:35 onward.
- Chronic `SPEND_BATCH_LUA … invalid spend for vout N: Invalid spend` bless failures on unrelated subtrees, running for hours before the stall.

## Environment

- Affected: `bsva-mainnet-eu-1` (v0.15.8), `bsva-mainnet-eu-2` (v0.15.6), `teranode-fde3` (v0.15.6), `bsva-mainnet-docker-eu-2` (v0.16.0), `mainnet.gorillanode.io` / `mainnet2.gorillanode.io` (v0.15.6) — all at 960,827.
- UTXO store: Aerospike.
- Evidence is from `teranode-mainnet-eu-1` only; the other nodes wedging on the same transaction is inferred from identical code and identical stall height, not directly observed.

oskarszoon:
Closes #1391.

## Problem

Every mainnet Teranode is wedged at 960,827 — still wedged as of writing. Block 960,828 contains `005a1b58…85d7`, which the local UTXO store flags conflicting. Blessing it enters `checkCounterConflictingOnCurrentChain`, which ran 26m21s and ended only when its context was cancelled. The block never reaches block validation, the legacy peer message handler is held until `legacy_peerProcessingTimeout` fires — 10m in the committed `settings.conf:725`, which is the `waited 10m0s` in the incident log, not the separate 5m `stallResponseTimeoutBlocks` — the peer is dropped, the block is re-fetched, and the cycle repeats. Fleet-wide across v0.15.6, v0.15.8 and v0.16.0 — the code is identical on all of them and on `main`.

## Root cause

The loop at `services/subtreevalidation/SubtreeValidation.go:503-514` walked the full descendant cone once per element of `counterConflictingTxHashes`:

```go
for _, counterConflictingTxHash := range counterConflictingTxHashes {
    childTransactionHashes, err := utxo.GetConflictingChildren(ctx, u.utxoStore, counterConflictingTxHash)
    ...
}
```

That slice already *is* the transitive closure of those descendants. `GetCounterConflictingTxHashes` calls `GetConflictingChildren` on each recorded spender and unions every member of the returned closure into the same `counterConflictingMap`. So the loop ran one full descendant BFS per element of the set that is the descendant set, with `visited` scoped per call, no sharing between iterations, and the outer loop fully serial.

(Citations here are by symbol rather than by line. The metrics var block this PR adds shifts `process_conflicting.go` by ~49 lines, so any file:line written against the pre-fix file is wrong at HEAD — the commit that writes the coordinates is the commit that invalidates them.)

The production shape makes it worse than O(N²) alone suggests. The counter-conflicting transaction is `62e564c8…7a6a`: 1 input, 2 outputs, vout 1 paying back to the same P2PK script with an incrementing OP_RETURN counter — an automated self-spend chain, never mined, absent from public mempools. Verified live against `mainnet-eu-1`: `blockIDs: []`, and each hop spends exactly the previous transaction's vout 1. The chain is strictly linear, so every BFS level holds exactly one node and a walk degenerates to N serial store round trips with no fan-out at all.

The production error text pins which phase burned the time — `failed to get child transactions` is `:506`, inside the deleted loop, not the first walk. The first walk completed; the 26 minutes went into re-walking.

Two consequences worth stating, because they shape the fix:

- The chain grows by roughly a hundred transactions per block for as long as the node cannot advance past the block that would demote it, so N is still increasing.
- The conflicting flag is correct. All six parents of `005a1b58…85d7` were mined in 960,827 and all six `vout 1` are spent on chain by it. This is a genuine double-spend that the block resolved in favour of the sweep, and the node correctly flagged the mined transaction as conflicting against the continuation it had already accepted.

## Fix

One walk instead of N. The detection set is unchanged, and the argument for that is:

- The cone is transitively closed, so `cone(m) ⊆ cone(Yᵢ)` for every member `m` — re-walking a member only re-derives a subset of what is already in hand.
- A frozen *descendant* of a counter already errors on the `FrozenBytesTxHash` check inside `GetCounterConflictingTxHashes`, which is applied to the full BFS closure rather than one level.
- A *directly*-frozen counter `Yᵢ` is covered only by the sentinel check in the errgroup above the deleted loop, and that check is load-bearing: `GetCounterConflictingTxHashes` inserts each `Yᵢ` into the map unconditionally with no frozen check of its own, and if `Yᵢ` is itself the sentinel then its walk early-returns `(nil, nil)`, so the closure check never sees it. Neither of these two covers depended on the deleted loop.
- The one term no earlier walk covers is the descendant cone of the transaction under check. `counterConflictingMap` is seeded with that transaction itself, which is why the deleted loop walked its cone — and why the fix keeps one walk rather than deleting the check outright.

Since the transaction under check was always a member of `counterConflictingTxHashes`, the surviving call is one the old loop already made — the new rejection set is a strict subset of the old, so no block that current code blesses can now be rejected.

Note for anyone reading the issue's original write-up: `FrozenBytesTxHash` and `CoinbasePlaceholderHashValue` are byte-identical (both 32 × `0xFF`, `go-subtree/coinbase_placeholder.go`). There is no sentinel discrepancy between `:510` and `:1125` to reconcile.

## Telemetry

`GetConflictingChildren` and `GetAndLockChildren` had no ceiling on per-level width — one concurrent store read per level member. Both now `SetLimit(4096)`, matching the aerospike get batcher's `getBatcherSize` so a level still fills a batch in a single wave instead of waiting out `getBatcherDurationMillis`. A lower ceiling starves that batcher: at 128, a 10,000-node level goes from ~3 fill-triggered round trips to ~78 timer-triggered waves.

Two histograms on completed walks, `teranode_utxo_conflicting_walk_nodes` and `teranode_utxo_conflicting_walk_depth`, plus a caller-side warn when the counter-conflicting set crosses 1000. The cone is unbounded by construction and its size is the leading indicator of this condition; the `stores/utxo` free functions have no logger, which is why the metric sits in the walk and the log in the caller.

## What this deliberately does not do

- **No fail-closed node or depth budget.** A cap trips on exactly this linear-chain shape and converts a 26-minute stall into a permanent fast rejection. The node still never advances — strictly worse than slow.
- **No narrowing of the walk to conflicting children only.** `ProcessConflicting` uses the counter set as its demotion input, and demoting a conflicting loser requires its full descendant set.

## Tests

`services/subtreevalidation/conflicting_walk_cost_test.go` — a map-backed `utxo.Store` counting `Get` calls, with the real walk delegated to the production free functions rather than stubbed. Plants the incident's shape: a linear self-spend chain of 1000.

- The cost test failed pre-fix at 501,503 `Get` calls (1002 to build the counter-conflicting set + Σ(1..1000) for the re-walks + 1 for re-walking the transaction under check, which is itself a member of the set) and passes post-fix at exactly 1003, pinned with `require.Equal` — the defect being guarded is a complexity regression, so the bound is the deterministic value rather than a multiple of it.
- Four equivalence guards pass on both sides of the fix, which is the point: the frozen sentinel at each of three placements (the counter itself, deep in the loser cone, and in the tx-under-check's own cone — that last one fails if the surviving walk is deleted), and the counter-mined-on-our-chain rejection.

`stores/utxo/conflicting_walk_metrics_test.go` — one observation per completed walk, with the right node count and depth, read as deltas since the histograms are process-global.

`go test -race ./services/subtreevalidation/... ./stores/utxo/...` green (1442 passed in 14 packages for `stores/utxo` alone), `go vet` clean.

## Residual risk

- One walk still costs N serial round trips on a linear cone, roughly 10 ms per level from the batcher's flush timer. At the estimated current N that is seconds, and it grows about 1 s per further 100 blocks of wedge. Bounded, but it argues for deploying sooner rather than later.
- The production logs cannot split the 26m21s between the first walk and the re-walks — there is no log line between them. So the hard upper bound on the surviving walk is only "completed within 26 minutes". The linear-chain arithmetic puts it at seconds and reproduces the observed magnitude, but that is a model, not a measurement. The `conflicting_walk_nodes` histogram is what replaces the estimate once this is deployed.
- Real-store latency is unverified: the cost test pins the complexity class, not aerospike round-trip behaviour.
- The deleted loop re-read every cone member as a walk root, which on the v0.15 line fails closed on root absence. A member whose record disappears between the first walk and the loop no longer fails closed. Race-window only — the first walk read it successfully microseconds earlier.

## Backports

`release/v0.15` lands separately; its `stores/utxo/process_conflicting.go` has diverged (ghost-spender tolerance from #1325) and the metrics var block conflicts. `release/v0.16` tracks `main`, so no separate backport.

### DIFF
--- a/services/subtreevalidation/SubtreeValidation.go
+++ b/services/subtreevalidation/SubtreeValidation.go
@@ -62,6 +62,12 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+// conflictingConeWarnThreshold is the counter-conflicting set size above which the
+// cone is worth a log line. The set is the loser's full descendant closure, which
+// keeps growing for as long as the node cannot advance past the block that would
+// demote it (#1391), so its size is the leading indicator of that condition.
+const conflictingConeWarnThreshold = 1000
+
 // missingTx represents a transaction that needs to be retrieved and its position in the subtree.
 //
 // This structure pairs a transaction with its index in the original subtree transaction list,
@@ -470,6 +476,10 @@ func (u *Server) checkCounterConflictingOnCurrentChain(ctx context.Context, txHa
 	counterConflictingTxHashes, err := utxo.GetCounterConflictingTxHashes(ctx, u.utxoStore, txHash)
 	if err != nil {
 		return errors.NewProcessingError("[checkCounterConflictingOnCurrentChain][%s] failed to get counter conflicting tx hashes", txHash.String(), err)
+	}
+
+	if len(counterConflictingTxHashes) > conflictingConeWarnThreshold {
+		u.logger.Warnf("[checkCounterConflictingOnCurrentChain][%s] counter-conflicting set is %d transactions", txHash.String(), len(counterConflictingTxHashes))
 	}
 
 	g, gCtx := errgroup.WithContext(ctx)
--- a/services/subtreevalidation/SubtreeValidation.go
+++ b/services/subtreevalidation/SubtreeValidation.go
@@ -510,20 +510,20 @@ func (u *Server) checkCounterConflictingOnCurrentChain(ctx context.Context, txHa
 	}
 
 	// One walk, not one per element. counterConflictingTxHashes is already the
-	// transitive closure of the counter-conflicting transactions' descendants
-	// (GetCounterConflictingTxHashes unions GetConflictingChildren of every spender
-	// into it at process_conflicting.go:1129), so re-walking each element only
+	// transitive closure of the counter-conflicting transactions' descendants —
+	// GetCounterConflictingTxHashes unions each spender's GetConflictingChildren
+	// result into counterConflictingMap — so re-walking each element only
 	// re-derives subsets of what is already in hand: N serial BFS walks over the set
 	// that IS the descendant set, one store round trip per level. With a linear
 	// self-spend chain for a cone that is O(N^2) reads, which wedged the mainnet
 	// fleet at height 960,827 for 26 minutes per attempt — see #1391.
 	//
-	// The frozen sentinel is already rejected on the counter side: inside the counter
-	// walk for descendants (process_conflicting.go:1125) and in the errgroup above for
-	// a directly-frozen counter. The one term no earlier walk covers is the descendant
-	// cone of the transaction under check itself, so that is the single walk that
-	// remains. It is cheap — the transaction has just arrived in a block, so it
-	// normally has no descendants in the store at all.
+	// The frozen sentinel is already rejected on the counter side: by the
+	// FrozenBytesTxHash check inside GetCounterConflictingTxHashes, and in the
+	// errgroup above for a directly-frozen counter. The one term no earlier walk
+	// covers is the descendant cone of the transaction under check itself, so that
+	// is the single walk that remains. It is cheap — the transaction has just
+	// arrived in a block, so it normally has no descendants in the store at all.
 	childTransactionHashes, err := utxo.GetConflictingChildren(ctx, u.utxoStore, txHash)
 	if err != nil {
 		return errors.NewProcessingError("[checkCounterConflictingOnCurrentChain][%s] failed to get child transactions", txHash.String(), err)
--- a/services/subtreevalidation/SubtreeValidation.go
+++ b/services/subtreevalidation/SubtreeValidation.go
@@ -499,17 +499,29 @@ func (u *Server) checkCounterConflictingOnCurrentChain(ctx context.Context, txHa
 		return errors.NewProcessingError("[checkCounterConflictingOnCurrentChain][%s] failed to get counter conflicting tx meta", txHash.String(), err)
 	}
 
-	// check whether the child transactions of the counter-conflicting transactions are frozen
-	for _, counterConflictingTxHash := range counterConflictingTxHashes {
-		childTransactionHashes, err := utxo.GetConflictingChildren(ctx, u.utxoStore, counterConflictingTxHash)
-		if err != nil {
-			return errors.NewProcessingError("[checkCounterConflictingOnCurrentChain][%s] failed to get child transactions", txHash.String(), err)
-		}
+	// One walk, not one per element. counterConflictingTxHashes is already the
+	// transitive closure of the counter-conflicting transactions' descendants
+	// (GetCounterConflictingTxHashes unions GetConflictingChildren of every spender
+	// into it at process_conflicting.go:1129), so re-walking each element only
+	// re-derives subsets of what is already in hand: N serial BFS walks over the set
+	// that IS the descendant set, one store round trip per level. With a linear
+	// self-spend chain for a cone that is O(N^2) reads, which wedged the mainnet
+	// fleet at height 960,827 for 26 minutes per attempt — see #1391.
+	//
+	// The frozen sentinel is already rejected on the counter side: inside the counter
+	// walk for descendants (process_conflicting.go:1125) and in the errgroup above for
+	// a directly-frozen counter. The one term no earlier walk covers is the descendant
+	// cone of the transaction under check itself, so that is the single walk that
+	// remains. It is cheap — the transaction has just arrived in a block, so it
+	// normally has no descendants in the store at all.
+	childTransactionHashes, err := utxo.GetConflictingChildren(ctx, u.utxoStore, txHash)
+	if err != nil {
+		return errors.NewProcessingError("[checkCounterConflictingOnCurrentChain][%s] failed to get child transactions", txHash.String(), err)
+	}
 
-		for _, childTransactionHash := range childTransactionHashes {
-			if childTransactionHash.Equal(subtreepkg.CoinbasePlaceholderHashValue) {
-				return errors.NewProcessingError("[checkCounterConflictingOnCurrentChain][%s] child transaction is frozen", txHash.String())
-			}
+	for _, childTransactionHash := range childTransactionHashes {
+		if childTransactionHash.Equal(subtreepkg.CoinbasePlaceholderHashValue) {
+			return errors.NewProcessingError("[checkCounterConflictingOnCurrentChain][%s] child transaction is frozen", txHash.String())
 		}
 	}
 
--- a/services/subtreevalidation/SubtreeValidation_test.go
+++ b/services/subtreevalidation/SubtreeValidation_test.go
@@ -1015,6 +1015,7 @@ func Test_checkCounterConflictingOnCurrentChain(t *testing.T) {
 
 		// Create a mock Server struct
 		s := &Server{
+			logger:    ulogger.New("test"),
 			utxoStore: utxoStore,
 		}
 
@@ -1043,6 +1044,7 @@ func Test_checkCounterConflictingOnCurrentChain(t *testing.T) {
 
 		// Create a mock Server struct
 		s := &Server{
+			logger:    ulogger.New("test"),
 			utxoStore: utxoStore,
 		}
 
--- a/services/subtreevalidation/conflicting_walk_cost_test.go
+++ b/services/subtreevalidation/conflicting_walk_cost_test.go
@@ -1,0 +1,292 @@
+// Package subtreevalidation provides functionality for validating subtrees in a blockchain context.
+package subtreevalidation
+
+import (
+	"context"
+	"strconv"
+	"sync"
+	"sync/atomic"
+	"testing"
+
+	"github.com/bsv-blockchain/go-bt/v2"
+	"github.com/bsv-blockchain/go-bt/v2/bscript"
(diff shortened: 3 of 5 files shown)


oskarszoon:
Closing this in favour of #1393, which reached the same conclusion independently and has since absorbed everything here that mattered.

The core change is the same in both: delete the per-member re-walk in `checkCounterConflictingOnCurrentChain` and replace it with a single walk of the blessed tx's own cone. Diffing the two function bodies, the only deltas were the `>1000` cone warn and the threaded budget argument — so the O(N²) removal was never the differentiator.

#1393 additionally has the per-input dedupe in `GetCounterConflictingTxHashes` (which I deferred here as a follow-up, and which is the dominant remaining term once the re-walk is gone), the SQL frozen-sentinel fix, and a node budget. As of `825d31b10` it also ports both histograms and the cone warn from this PR, bounds `GetAndLockChildren`, and adds a `conflicting_walk_duration` histogram this PR did not have — including the tracer `WithHistogram` wiring that was missing here.

Keeping both open only creates merge pain: they now conflict in three files, and because the ported metric blocks are textually identical, both pairs of `Observe` calls merge clean outside any conflict marker — a resolved merge would compile, vet clean, and record every walk twice.

Work continues in:
- #1393 — `main`
- #1394 — `release/v0.15` backport

Two items from this PR's review that #1393 does not inherit, tracked for follow-up rather than lost: the `SetLimit(128)` on the `GetMeta` errgroup in `checkCounterConflictingOnCurrentChain` (unchanged in both PRs, and the batcher argument here applies to it), and the fan-out constant question — #1393 argues 128 on SQL connection-pool grounds, which is a better-supported position than the 4096 here, but neither derives it per backend.

PR #1397, opened 2026-08-05
fix(utxo/sql): honour BlockHeights and SubtreeIdxs when asked for on their own

freemans13:
## What happened

The SQL UTXO store fills `BlockIDs`, `BlockHeights` and `SubtreeIdxs` from a single query against the `block_ids` table, but it only ran that query when the caller asked for `BlockIDs`. Ask for `BlockHeights` or `SubtreeIdxs` on their own and you got nothing back and no error — indistinguishable from a transaction that has genuinely never been mined.

On the batched path it was worse than simply returning nothing, and thanks to @ordishs and @ctnguyen for spotting this. `sendGetBatch` (`sql.go:1493-1509`) collects the union of the field sets of every request in the batch and hands that union to `BatchDecorate`, which then makes the decision on the union. So before this change, a `BlockHeights`-only `Get` returned the right answer *if and only if* some other request that happened to be batched alongside it had asked for `BlockIDs`. Since `settings.conf` sets `utxostore_getBatcherSize = 4096`, every `Get` goes through that batcher in the default configuration. The old behaviour was therefore load-dependent: right sometimes, wrong sometimes, decided by what else was in flight. A test written against it would have been flaky rather than red, which is a good part of why this survived.

No production caller trips over it today, because both callers that want block heights ask for block IDs alongside them (`services/validator/Validator.go:1132` and `services/subtreevalidation/check_block_subtrees.go:1263`). So this is a trap rather than a live fault. It is an expensive one to leave lying around, though, and I found it by nearly walking into it.

Block heights feed the input script block height handed to the script verifier, which is what BIP68 relative locktime reads. If someone later removed the apparently-unused `BlockIDs` from one of those two field lists — a very reasonable-looking cleanup, since nothing reads the returned value — then `Get` would return no block heights for any parent, the validator would take its "unconfirmed parent" branch for every input, and consensus-relevant behaviour would change on every sqlite-backed store. Including the `sqlitememory` store our own testing guidance tells people to prefer. It would pass sync tests, because the sentinel resolves to the candidate height on block paths, and only misbehave once a node is caught up.

One caller does already ask for `BlockHeights` alone: `services/subtreevalidation/legacy_real_validator_integration_test.go:113`, against a real `sqlitememory` store. It asserts the parent is *not* mined, so it passed before and still passes — but before this change that assertion was vacuous. It would have passed even if the parent had been mined. The fix turns it into a real check.

## The change

Both places that decide whether to run the `block_ids` query now ask whether any of the three fields it serves was requested, rather than only `BlockIDs`. That question is now asked once, by a `needsBlockIDsQuery` helper (`sql.go:1786`) that both call sites use. The condition was originally copied into both places, which is precisely the drift that produced the bug, so stating it once is the point rather than a tidy-up.

The two call sites are `getUnbatched` (`sql.go:1653`) and the `needBlockIDs` decision in `batchDecorateChunk` (`sql.go:3696`) — the latter being the one the validator's parent prefetch actually uses. This follows the convention already in use two lines above the second site, where `needInputs` and `needOutputs` each OR together every field served by the same query. `needBlockIDs` was the odd one out.

`Get`'s doc comment now records that the three placement fields travel together and are index-aligned, since that grouping previously existed only in the shape of one SQL statement. It also drops `parentTxHashes`, which has not been a member of `fields.FieldName` for some time — the field is `TxInpoints`.

No caller behaviour changes: asking for `BlockIDs` still returns all three fields exactly as before, which is what everything in the tree relies on today. Widening the condition can only add truthful data, never turn a mined transaction into a seemingly unmined one.

## Testing

`TestGetBlockPlacementFieldsIndependently` checks each of the two read paths against each of the two fields that were previously broken. It calls `getUnbatched` and `BatchDecorate` directly rather than going through `Get`, because which path a `Get` takes is decided by configuration — with `getBatcherSize = 4096` from `settings.conf` every `Get` is batched, and a developer running with the struct-tag default of 1 would move the coverage onto the other path rather than adding to it. One case does drive the public `Get`, to confirm the routing reaches the query at all, and a final case pins the existing behaviour that `BlockIDs` alone still returns all three.

I verified each part fails on its own, rather than checking that the group fails together:

| What I broke | What failed |
|---|---|
| `BlockHeights` term in the helper | both `BlockHeights` cases, plus the public `Get` case |
| `SubtreeIdxs` term in the helper | both `SubtreeIdxs` cases |
| `getUnbatched` call site narrowed back to `BlockIDs` | both unbatched cases |
| `batchDecorateChunk` call site narrowed back to `BlockIDs` | both batched cases, plus the public `Get` case |

Failures are `[]uint32(nil)` where `[]uint32{0x63}` is expected, which is the silent-empty symptom described above. The public `Get` case failing in the fourth experiment and not the third confirms it travels the batched path in the default configuration.

The first version of this PR did not do this. Its test reached only the batched path, so reverting the `getUnbatched` half left all four subtests passing and half the production change was unguarded — caught in review by @ordishs and @ctnguyen, and the reason the test is now structured around direct calls.

Limits worth stating: the `stores/utxo/sql` package has ten pre-existing `_Postgres` test failures that need a live postgres server, which I did not have running. In my environment they fail rather than skip, with `rootless Docker not found`. I checked the names either side of this change and they are identical, so they are unrelated, but I have not run those particular tests green. I also have not exercised this against a real aerospike cluster.

## Two things a reviewer may notice, neither introduced here

Both surfaced in a review pass over this branch. I am recording them so they are not mistaken for new faults.

Neither `block_ids` loop checks `rows.Err()` once `rows.Next()` returns false — `sql.go:1672` in `getUnbatched`, and `sql.go:3872` in `batchDecorateChunk`. If the driver fails part-way through iteration, the caller gets a short but still index-aligned placement list and a nil error, which reads as a transaction mined in fewer blocks than it really was. This change does not create that gap, but it does make those two loops run more often. It is a house-wide pattern rather than a local slip: only five of the thirteen row loops in `sql.go` check `rows.Err()`. Fixing two of the eight unguarded ones here would be arbitrary, so I have left it for a separate sweep of `getUnbatched` and the three `batchDecorate*` helpers together.

The second is a visibility change rather than a behaviour change. `sendGetBatch` hands every member of a batch the union of the batch's field sets, so a caller that asked only for `Tx` can now come back with the placement fields populated because something else in the same batch asked for one of them — where previously the query would not have run at all. The data is truthful, and every downstream reader I checked treats populated placement as "mined", which is the correct reading. This is the same union behaviour described at the top of this description, seen from the other side.

## A note on the two backends

An earlier version of this description claimed the change makes the SQL and aerospike stores agree. That was too strong, and @ctnguyen was right to push back on it.

What is true is that `stores/utxo/aerospike/get.go:601-608` adds `BlockHeights` and `SubtreeIdxs` whenever `BlockIDs` is requested, and that aerospike serves a standalone `BlockHeights` or `SubtreeIdxs` request correctly, because each is its own bin read independently. So aerospike never had this bug.

What is not true is that the two stores now return identical field sets. Aerospike expands only *from* `BlockIDs`, so asking it for `BlockHeights` alone leaves `BlockIDs` nil, whereas SQL fills all three once the query runs. That is not a defect in either — `Interface.go:412-414` says the fields parameter selects what to *retrieve*, and both stores now honour a request for any placement field, which is the contract. The accurate claim is the narrower one: neither backend silently returns empty for a field you asked for.

### DIFF
--- a/stores/utxo/sql/sql.go
+++ b/stores/utxo/sql/sql.go
@@ -1650,7 +1650,10 @@ func (s *Store) getUnbatched(ctx context.Context, hash *chainhash.Hash, bins []f
 		}
 	}
 
-	if contains(bins, fields.BlockIDs) {
+	// This one query fills BlockIDs, BlockHeights and SubtreeIdxs together, so asking
+	// for any of the three has to run it. Gating on BlockIDs alone meant a caller that
+	// asked only for BlockHeights or SubtreeIdxs got an empty slice and no error.
+	if contains(bins, fields.BlockIDs) || contains(bins, fields.BlockHeights) || contains(bins, fields.SubtreeIdxs) {
 		q := `
 			SELECT
 			    block_id,
@@ -3683,7 +3686,10 @@ func (s *Store) batchDecorateChunk(ctx context.Context, items []*utxo.Unresolved
 
 	needInputs := contains(bins, fields.Tx) || contains(bins, fields.Inputs) || contains(bins, fields.TxInpoints) || contains(bins, fields.Utxos)
 	needOutputs := contains(bins, fields.Tx) || contains(bins, fields.Outputs) || contains(bins, fields.Utxos)
-	needBlockIDs := contains(bins, fields.BlockIDs)
+	// One query fills BlockIDs, BlockHeights and SubtreeIdxs together, so asking for
+	// any of the three has to run it. Gating on BlockIDs alone meant a caller that
+	// asked only for BlockHeights or SubtreeIdxs got an empty slice and no error.
+	needBlockIDs := contains(bins, fields.BlockIDs) || contains(bins, fields.BlockHeights) || contains(bins, fields.SubtreeIdxs)
 
 	// Query 2: Bulk fetch inputs
 	if needInputs {
--- a/stores/utxo/sql/sql.go
+++ b/stores/utxo/sql/sql.go
@@ -1446,8 +1446,10 @@ func (s *Store) GetMeta(ctx context.Context, hash *chainhash.Hash, data *meta.Da
 //   - tx: Full transaction data
 //   - inputs: Transaction inputs
 //   - outputs: Transaction outputs
-//   - blockIDs: Block references
-//   - parentTxHashes: Previous transaction hashes
+//   - blockIDs, blockHeights, subtreeIdxs: where this transaction has been mined,
+//     index-aligned — entry i of each describes one placement. All three come from
+//     a single query, so requesting any one of them runs it and fills all three.
+//   - txInpoints: parent outpoints referenced by this transaction
 func (s *Store) Get(ctx context.Context, hash *chainhash.Hash, fields ...fields.FieldName) (*meta.Data, error) {
 	bins := utxo.MetaFieldsWithTx
 	if len(fields) > 0 {
@@ -1650,10 +1652,7 @@ func (s *Store) getUnbatched(ctx context.Context, hash *chainhash.Hash, bins []f
 		}
 	}
 
-	// This one query fills BlockIDs, BlockHeights and SubtreeIdxs together, so asking
-	// for any of the three has to run it. Gating on BlockIDs alone meant a caller that
-	// asked only for BlockHeights or SubtreeIdxs got an empty slice and no error.
-	if contains(bins, fields.BlockIDs) || contains(bins, fields.BlockHeights) || contains(bins, fields.SubtreeIdxs) {
+	if needsBlockIDsQuery(bins) {
 		q := `
 			SELECT
 			    block_id,
@@ -1781,6 +1780,16 @@ func contains(slice []fields.FieldName, item fields.FieldName) bool {
 	}
 
 	return false
+}
+
+// needsBlockIDsQuery reports whether any field served by the single block_ids
+// query was requested. BlockIDs, BlockHeights and SubtreeIdxs all come from that
+// one query, so asking for any of them has to run it. Both read paths call this
+// so the grouping is stated once rather than copied.
+func needsBlockIDsQuery(bins []fields.FieldName) bool {
+	return contains(bins, fields.BlockIDs) ||
+		contains(bins, fields.BlockHeights) ||
+		contains(bins, fields.SubtreeIdxs)
 }
 
 // parseInsertedAtMillis converts the inserted_at column value into Unix
@@ -3686,10 +3695,7 @@ func (s *Store) batchDecorateChunk(ctx context.Context, items []*utxo.Unresolved
 
 	needInputs := contains(bins, fields.Tx) || contains(bins, fields.Inputs) || contains(bins, fields.TxInpoints) || contains(bins, fields.Utxos)
 	needOutputs := contains(bins, fields.Tx) || contains(bins, fields.Outputs) || contains(bins, fields.Utxos)
-	// One query fills BlockIDs, BlockHeights and SubtreeIdxs together, so asking for
-	// any of the three has to run it. Gating on BlockIDs alone meant a caller that
-	// asked only for BlockHeights or SubtreeIdxs got an empty slice and no error.
-	needBlockIDs := contains(bins, fields.BlockIDs) || contains(bins, fields.BlockHeights) || contains(bins, fields.SubtreeIdxs)
+	needBlockIDs := needsBlockIDsQuery(bins)
 
 	// Query 2: Bulk fetch inputs
 	if needInputs {
--- a/stores/utxo/sql/sql_test.go
+++ b/stores/utxo/sql/sql_test.go
@@ -3625,3 +3625,78 @@ func TestMinimalCreate_CoinbaseMaturity_OutpointOnlySpend(t *testing.T) {
 		utxo.IgnoreFlags{IgnoreLocked: true, SkipUTXOHashCheck: true})
 	require.NoError(t, err, "spending at maturity height must succeed")
 }
+
+// TestGetBlockHeightsWithoutBlockIDs pins a field-selection bug that was silent.
+//
+// BlockIDs, BlockHeights and SubtreeIdxs all come from one query against the
+// block_ids table, and that query used to run only when the caller asked for
+// BlockIDs. A caller that asked for BlockHeights or SubtreeIdxs on their own got an
+// empty slice back and no error at all — indistinguishable from a transaction that
+// genuinely has not been mined.
+//
+// Nothing in the tree hits it today, because both callers that want BlockHeights
+// happen to ask for BlockIDs alongside it (Validator.go and
+// check_block_subtrees.go). That makes this a trap rather than a live fault, and
+// an expensive one: BlockHeights feeds the input script block height used for
+// BIP68 relative locktime, so a future caller that dropped the seemingly-unused
+// BlockIDs from its field list would have silently changed consensus-relevant
+// behaviour on every sqlite-backed store — including the sqlitememory store the
+// project's own guidance tells people to use in tests.
+//
+// The aerospike store has the opposite habit: it ADDS BlockHeights and SubtreeIdxs
+// whenever BlockIDs is requested (aerospike/get.go addAbstractedBins). Making the
+// two backends agree is the point of the fix.
+func TestGetBlockHeightsWithoutBlockIDs(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	utxoStore, tx := setup(ctx, t)
+
+	_, err := utxoStore.Create(ctx, tx, 0)
+	require.NoError(t, err)
+
+	_, err = utxoStore.SetMinedMulti(ctx, []*chainhash.Hash{tx.TxIDChainHash()}, utxo.MinedBlockInfo{
+		BlockID:        7,
+		BlockHeight:    99,
+		SubtreeIdx:     3,
+		OnLongestChain: true,
+	})
+	require.NoError(t, err)
+
+	t.Run("Get with only BlockHeights", func(t *testing.T) {
+		data, err := utxoStore.Get(ctx, tx.TxIDChainHash(), fields.BlockHeights)
+		require.NoError(t, err)
+		require.Equal(t, []uint32{99}, data.BlockHeights,
+			"asking for BlockHeights alone must return them, not an empty slice")
+	})
+
+	t.Run("Get with only SubtreeIdxs", func(t *testing.T) {
+		data, err := utxoStore.Get(ctx, tx.TxIDChainHash(), fields.SubtreeIdxs)
+		require.NoError(t, err)
+		require.Equal(t, []int{3}, data.SubtreeIdxs,
+			"asking for SubtreeIdxs alone must return them, not an empty slice")
+	})
+
+	// BatchDecorate reaches the same block_ids query through a separate gate, so it
+	// needs its own case; fixing only the single-Get path would leave the batched
+	// path — the one the validator's parent prefetch actually uses — still wrong.
+	t.Run("BatchDecorate with only BlockHeights", func(t *testing.T) {
+		items := []*utxo.UnresolvedMetaData{{Hash: *tx.TxIDChainHash(), Idx: 0}}
+
+		require.NoError(t, utxoStore.BatchDecorate(ctx, items, fields.BlockHeights))
+		require.NoError(t, items[0].Err)
+		require.NotNil(t, items[0].Data)
+		require.Equal(t, []uint32{99}, items[0].Data.BlockHeights,
+			"batched read must honour BlockHeights on its own too")
+	})
+
+	// The existing behaviour must be untouched: BlockIDs alone still returns all
+	// three, which is what every caller in the tree relies on today.
+	t.Run("Get with only BlockIDs still returns all three", func(t *testing.T) {
+		data, err := utxoStore.Get(ctx, tx.TxIDChainHash(), fields.BlockIDs)
+		require.NoError(t, err)
+		require.Equal(t, []uint32{7}, data.BlockIDs)
+		require.Equal(t, []uint32{99}, data.BlockHeights)
+		require.Equal(t, []int{3}, data.SubtreeIdxs)
+	})
+}
--- a/stores/utxo/sql/sql_test.go
+++ b/stores/utxo/sql/sql_test.go
@@ -3626,27 +3626,21 @@ func TestMinimalCreate_CoinbaseMaturity_OutpointOnlySpend(t *testing.T) {
 	require.NoError(t, err, "spending at maturity height must succeed")
 }
 
-// TestGetBlockHeightsWithoutBlockIDs pins a field-selection bug that was silent.
+// TestGetBlockPlacementFieldsIndependently pins a field-selection bug that was silent.
 //
-// BlockIDs, BlockHeights and SubtreeIdxs all come from one query against the
-// block_ids table, and that query used to run only when the caller asked for
-// BlockIDs. A caller that asked for BlockHeights or SubtreeIdxs on their own got an
-// empty slice back and no error at all — indistinguishable from a transaction that
-// genuinely has not been mined.
+// BlockIDs, BlockHeights and SubtreeIdxs are index-aligned placement fields served by
+// one query against the block_ids table, and that query used to run only when the
+// caller asked for BlockIDs. Asking for either of the other two on its own returned an
+// empty slice and no error — indistinguishable from a transaction never mined.
 //
-// Nothing in the tree hits it today, because both callers that want BlockHeights
-// happen to ask for BlockIDs alongside it (Validator.go and
-// check_block_subtrees.go). That makes this a trap rather than a live fault, and
-// an expensive one: BlockHeights feeds the input script block height used for
-// BIP68 relative locktime, so a future caller that dropped the seemingly-unused
-// BlockIDs from its field list would have silently changed consensus-relevant
-// behaviour on every sqlite-backed store — including the sqlitememory store the
-// project's own guidance tells people to use in tests.
+// It matters because BlockHeights feeds the input script block height used for BIP68
+// relative locktime, so a caller that dropped the apparently-unused BlockIDs from its
+// field list would have changed consensus-relevant behaviour with no failure to see.
 //
-// The aerospike store has the opposite habit: it ADDS BlockHeights and SubtreeIdxs
-// whenever BlockIDs is requested (aerospike/get.go addAbstractedBins). Making the
-// two backends agree is the point of the fix.
-func TestGetBlockHeightsWithoutBlockIDs(t *testing.T) {
+// Each read path is checked against each of the two previously-broken fields. The two
+// paths reach the query through separate checks, and each check ORs the two fields
+// separately, so any one of the four could be removed without the others noticing.
+func TestGetBlockPlacementFieldsIndependently(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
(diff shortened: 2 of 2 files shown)


ordishs (approved):
## Correctness — the fix is right and complete

- Both gates are covered and there is no third one. Repo-wide, the only other `BlockIDs`-keyed expansion is aerospike's `addAbstractedBins` (`stores/utxo/aerospike/get.go:601`), which is a superset-add rather than a gate:
  - `stores/utxo/sql/sql.go:1656` (`getUnbatched`)
  - `stores/utxo/sql/sql.go:3692` (`batchDecorateChunk`)
- Widening the gate can only add truthful data — it cannot flip a mined transaction into looking unmined — so no existing caller can regress. Asking for `BlockIDs` returns byte-identical results.
- Style matches the two lines above it, where `needInputs`/`needOutputs` each OR together every field their query serves. `needBlockIDs` was genuinely the outlier.
- No import changes (so no `gci` concern), static SQL, no new parameters.

One thing the description understates: on the batched path the old behaviour was not merely wrong, it was **nondeterministic**. `sendGetBatch` (`sql.go:3465-3480`) unions the field sets of every item in the batch and hands that union to `BatchDecorate`, and `batchDecorateChunk` gates on the union. So before this fix, a `BlockHeights`-only `Get` returned heights *iff* some co-batched item happened to ask for `BlockIDs` — batch-composition-dependent, therefore load-dependent. That is a stronger argument for the change than "silently empty", and worth a line in the description.

## Test coverage — one real gap

The `getUnbatched` hunk (`sql.go:1656`) is **not exercised by the new test**. Reverting that hunk alone, keeping the batched one:

```
--- PASS: TestGetBlockHeightsWithoutBlockIDs/Get_with_only_BlockHeights (0.00s)
--- PASS: TestGetBlockHeightsWithoutBlockIDs/Get_with_only_SubtreeIdxs (0.00s)
--- PASS: TestGetBlockHeightsWithoutBlockIDs/BatchDecorate_with_only_BlockHeights (0.00s)
--- PASS: TestGetBlockHeightsWithoutBlockIDs/Get_with_only_BlockIDs_still_returns_all_three (0.00s)
PASS
ok  github.com/bsv-blockchain/teranode/stores/utxo/sql 0.583s
```

Reverting only the *batched* hunk fails three subtests. So all four failures attributed to "reverting the fix" come from that single hunk; the single-`Get` gate has nothing holding it.

Cause: `test.CreateBaseTestSettings` picks up `utxostore_getBatcherSize = 4096` from `settings.conf` (the struct-tag default of 1 is not what applies here), so `getBatcher != nil` and `s.get` (`sql.go:1466`) always routes to `getBatched` → `BatchDecorate`. `getUnbatched` is only reached in production when `getBatcherSize <= 1`, via `GetMeta`, or when `ConflictingChildren`/`Utxos` are in the field list.

Same package, so either of these works:

```go
t.Run("unbatched Get with only BlockHeights", func(t *testing.T) {
    data, err := utxoStore.getUnbatched(ctx, tx.TxIDChainHash(), []fields.FieldName{fields.BlockHeights})
    require.NoError(t, err)
    require.Equal(t, []uint32{99}, data.BlockHeights)
})
```

or build a second store with `tSettings.UtxoStore.GetBatcherSize = 1` to go through the public `Get`. Given the PR's own framing — "fixing only one would have left the batched path still wrong" — the symmetric case deserves the same guard.

Smaller test notes:

- The batched subtest covers only `BlockHeights`; add `SubtreeIdxs` (or table-drive the four cases) so the batched gate's second field is pinned too.
- The 20-line test doc comment restates the PR description almost verbatim. Three or four lines (the invariant, plus why it matters for BIP68) would age better; the narrative belongs in the PR.
- Conventions followed: `require` not `assert`, `sqlitememory`, no `t.Parallel()`.

## Maintainability

The same three-way condition plus a near-identical three-line comment now lives at two sites — exactly the drift that produced the bug. Worth collapsing:

```go
// needsBlockIDsQuery reports whether any field served by the single block_ids
// query was requested. BlockIDs, BlockHeights and SubtreeIdxs all come from it.
func needsBlockIDsQuery(bins []fields.FieldName) bool {
    return contains(bins, fields.BlockIDs) || contains(bins, fields.BlockHeights) || contains(bins, fields.SubtreeIdxs)
}
```

Longer term the cleaner shape is what aerospike already has: a shared field-expansion helper in `stores/utxo/fields`, so the grouping is declared once and both backends consume it. Follow-up, not a blocker.

Nit: the `Get` doc comment at `sql.go:1445-1451` already enumerates field semantics and is a natural place to state that `blockIDs`/`blockHeights`/`subtreeIdxs` are always returned as a group.

## Performance and security

- No cost today: nothing in the tree asks for these fields without `BlockIDs`, so no new query is issued anywhere. Future callers pay for one query they previously "saved" by receiving wrong data.
- No consensus behaviour change now. It removes a latent consensus-relevant trap (the input script block height feeding BIP68 relative locktime in `services/validator/Validator.go`) rather than altering any current outcome.

## Verification

```
$ go test -count=1 ./stores/utxo/sql/
ok  	github.com/bsv-blockchain/teranode/stores/utxo/sql	19.000s
$ go vet ./stores/utxo/sql/
(no output)
```

The package was fully green in my environment — the ten `_Postgres` failures mentioned in the description did not appear (those tests skip without a server), so I can neither corroborate nor contradict that; it does not affect this change either way.

## Verdict

Approving. The one thing I would still like is a subtest that actually reaches `getUnbatched`, since half the diff is currently unguarded by the test written to guard it. Everything else above is a nit.

ctnguyen (changes requested):
Review at commit [813dd2e](https://github.com/freemans13/teranode/commit/813dd2eac2ffc4add3d34ba0e94bc487991abd65)

Nice catch, and a well-argued one. I chased the diagnosis through the code and it holds up: both field-selected read paths really were gated on `BlockIDs` alone, the trap that leaves behind is real rather than theoretical, and widening the gate can only add truthful data. The fix itself is two lines that match the `needInputs`/`needOutputs` shape immediately above them, and the description is refreshingly honest about what it did and didn't run. My comments are all about the guard rather than the fix — chiefly that the test doesn't reach half of what it was written to protect.

## Blocking issues

### ChiR1 - The `getUnbatched` gate is never executed by the new test
*Same as [@ordishs's point "the getUnbatched hunk is not exercised by the new test" (05-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1397#pullrequestreview-4863895348).*
**Problem:** All four subtests land on `batchDecorateChunk`; the `getUnbatched` gate at `sql.go:1656` — half the production diff — is never run. `settings.conf:1284` sets `utxostore_getBatcherSize = 4096` as a global default (the file has no `[context]` sections, so it applies unconditionally), `test.CreateBaseTestSettings` builds straight from `settings.NewSettings()` which reads it, and `setup()` (`sql_test.go:69-105`) never overrides it. `New` therefore creates `s.getBatcher` (`sql.go:235`, `getBatchSize > 1`), and `s.get` (`sql.go:1466`) routes every field set that excludes `ConflictingChildren`/`Utxos` to `getBatched` → `sendGetBatch` → `BatchDecorate`. The repo states this itself at `close_drain_postgres_test.go:78-80`, which forces `GetBatcherSize = 1` precisely because "getBatcher is only created when GetBatcherSize > 1". Two consequences beyond the point already raised: the description's "I confirmed the test has teeth by reverting the fix … all three new subtests fail" is only true of the batched hunk, and *which* gate the `Get` subtests hit is decided by configuration — a developer whose `settings_local.conf` sets `utxostore_getBatcherSize = 1` (the struct-tag default in `settings/utxostore_settings.go:45`, and what `compose/chainintegrity/chain_integrity.go:366` uses) flips all three `Get` subtests onto `getUnbatched` and leaves the batched gate held only by the single explicit `BatchDecorate` subtest. So neither gate is deterministically pinned.
**Why it matters:** The entire value of this PR is a regression guard against a silent, consensus-adjacent field-selection bug — the code fix itself changes no behaviour today. `getUnbatched` is the live path whenever `GetBatcherSize <= 1`, for `GetMeta` (`sql.go:1432`), and for any field list containing `ConflictingChildren` or `Utxos`. Merging the fix with nothing holding that gate leaves in place exactly the drift that produced the bug, and does so in the one file where a reader would reasonably assume it is covered.
**Fix:** Pin both gates explicitly and independently of config, and cover **both** added OR terms on each — the unbatched gate ORs in `BlockHeights` *and* `SubtreeIdxs`, so a `BlockHeights`-only case alone would still let the `SubtreeIdxs` term be deleted unnoticed. Same package, so a direct call is the cheapest option:
```go
for _, tc := range []struct {
    name  string
    field fields.FieldName
    check func(*testing.T, *meta.Data)
}{
    {"BlockHeights", fields.BlockHeights, func(t *testing.T, d *meta.Data) { require.Equal(t, []uint32{99}, d.BlockHeights) }},
    {"SubtreeIdxs", fields.SubtreeIdxs, func(t *testing.T, d *meta.Data) { require.Equal(t, []int{3}, d.SubtreeIdxs) }},
} {
    t.Run("unbatched Get with only "+tc.name, func(t *testing.T) {
        data, err := utxoStore.getUnbatched(ctx, tx.TxIDChainHash(), []fields.FieldName{tc.field})
        require.NoError(t, err)
        tc.check(t, data)
    })
}
```
and assert the routing the batched subtests depend on, so a settings change cannot silently relocate coverage:
```go
require.NotNil(t, utxoStore.getBatcher, "batched subtests below require the get batcher")
```
Alternatively build a second store with `tSettings.UtxoStore.GetBatcherSize = 1` and drive the public `Get`, as `close_drain_postgres_test.go:80` already does. Also worth correcting the description's verification claim to match what the reverts actually show.

## Non-blocking issues

### ChiR2 - Batched subtest leaves `SubtreeIdxs` unpinned
*Same as [@ordishs's point "add SubtreeIdxs" (05-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1397#pullrequestreview-4863895348).*
**Problem:** Only `BlockHeights` is exercised through `BatchDecorate` (`sql_test.go:3683-3691`). The batched gate at `sql.go:3692` ORs three terms, and dropping the `contains(bins, fields.SubtreeIdxs)` term leaves every subtest green.
**Why it matters:** It is the same asymmetry the PR argues against for the two gates — "fixing only one would have left the batched path still wrong" applies just as well within a gate. `SubtreeIdxs` is not a hypothetical selection: it is read alongside heights in the asset service (`services/asset/httpimpl/GetTransactionMeta.go:139-142`).
**Fix:** Table-drive so each of the two added fields is asserted on each path — the same table as in ChiR1, reused for `Get` and `BatchDecorate`:
```go
t.Run("BatchDecorate with only "+tc.name, func(t *testing.T) {
    items := []*utxo.UnresolvedMetaData{{Hash: *tx.TxIDChainHash(), Idx: 0}}
    require.NoError(t, utxoStore.BatchDecorate(ctx, items, tc.field))
    require.NoError(t, items[0].Err)
    require.NotNil(t, items[0].Data)
    tc.check(t, items[0].Data)
})
```

### ChiR3 - Duplicated gate condition, and the "backends agree" sentence overstates the diff
*Same as [@ordishs's point "the same three-way condition ... now lives at two sites" (05-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1397#pullrequestreview-4863895348).*
**Problem:** Two separate small things, both about the same grouping. First, his point: the three-way condition plus a near-identical three-line comment now exists at `sql.go:1656` and `sql.go:3692`, which is the drift that produced the bug in the first place. Second, and going slightly further than his review: the checked-in test comment at `sql_test.go:3646-3648` says aerospike "ADDS BlockHeights and SubtreeIdxs whenever BlockIDs is requested … Making the two backends agree is the point of the fix." The first clause is accurate (`aerospike/get.go:601-609`), but the conclusion is not: `addAbstractedBins` expands only *from* `BlockIDs`, so asking aerospike for `BlockHeights` alone still returns `BlockIDs == nil` while SQL now returns all three (`sql.go:1684-1686` fills them unconditionally once the gate opens). That is not a store defect — `Interface.go:412-414` says the fields parameter selects what *to retrieve*, and both stores now honour a request for any placement field, which is the contract — but the sentence claims a parity the diff does not establish, and it will be read as settled the next time someone touches either store.
**Why it matters:** The comment is the only place in the tree that records what the grouping means, and it is checked in. A reader who takes it literally will assume any one placement field yields all three on every backend, which is only true of SQL. The duplicated condition compounds it: the invariant lives in two copies of a comment rather than in one named thing.
**Fix:** Collapse the condition and narrow the claim. A package-private helper is enough here:
```go
// needsBlockIDsQuery reports whether any field served by the single block_ids
// query was requested. BlockIDs, BlockHeights and SubtreeIdxs all come from it.
func needsBlockIDsQuery(bins []fields.FieldName) bool {
    return contains(bins, fields.BlockIDs) || contains(bins, fields.BlockHeights) || contains(bins, fields.SubtreeIdxs)
}
```
used at both gates. Then reword the comment to what is actually true — "both backends now return a placement field when it is requested on its own; SQL fills all three from one query, aerospike reads only the bins asked for" — and drop the "making the two backends agree" line. His longer-term suggestion of a shared field-expansion helper in `stores/utxo/fields` is a reasonable follow-up if the grouping is ever to be declared once for both stores, but it is out of scope for a SQL-only fix.

### ChiR4 - The grouping invariant is not documented where callers look
*Same as [@ordishs's point "the Get doc comment at sql.go:1445-1451" (05-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1397#pullrequestreview-4863895348).*
**Problem:** `Get`'s doc comment (`sql.go:1444-1450`) enumerates field semantics but says nothing about the group, and it is already stale — it lists `parentTxHashes`, which is not a member of `fields.FieldName` (the field is `TxInpoints`). The only statement of the invariant is a comment duplicated at the two gates, plus a test named `TestGetBlockHeightsWithoutBlockIDs` that also covers `SubtreeIdxs`.
**Why it matters:** This bug existed because the grouping lived only in the shape of one SQL statement. Widening the gates fixes today's behaviour but doesn't tell the next caller that the three fields travel together, which is the knowledge that prevents a repeat.
**Fix:** State it where the selection is made, e.g.
```go
//   - blockIDs, blockHeights, subtreeIdxs: block references, index-aligned —
//     entry i of each describes one placement. All three come from one query, so
//     requesting any of them runs it.
//   - txInpoints: parent outpoints referenced by this transaction
```
and drop the `parentTxHashes` line. Renaming the test to something like `TestGetBlockPlacementFieldsIndependently` would also stop it reading as covering only two of the three.

### ChiR5 - Old batched behaviour was load-dependent, not just empty
*Same as [@ordishs's point "nondeterministic" (05-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1397#pullrequestreview-4863895348).*
**Problem:** `sendGetBatch` (`sql.go:1490-1509`) unions the field sets of every item in the batch and hands the union to `BatchDecorate`, and `batchDecorateChunk` gates on that union. Since `utxostore_getBatcherSize = 4096` in `settings.conf`, the store's `Get` goes through that batcher in every default configuration, so pre-fix a `BlockHeights`-only `Get` returned heights *iff* some co-batched request happened to ask for `BlockIDs`. The description and the test comment both characterise the old behaviour as "an empty slice and no error".
**Why it matters:** The description and commit body are what the next person reads. "Returns empty" implies a reproducible failure; "returns the right answer some of the time depending on concurrent load" is both a stronger argument for the change and an explanation of why a test would have been flaky rather than red.
**Fix:** One line in the description and the commit body noting the batch-composition dependence on the batched path; the test comment can then shrink to the invariant plus the BIP68 reason and leave the narrative to the PR.

## Recap

| ChiR1 | Unbatched gate untested | YES | 80% |
| ChiR2 | SubtreeIdxs unpinned on batched path | | 35% |
| ChiR3 | Duplicated gate condition, parity claim | | 25% |
| ChiR4 | Grouping invariant undocumented | | 20% |
| ChiR5 | Old behaviour was load-dependent | | 15% |

freemans13:
Thanks both — this was a genuinely useful review. You each independently found the same hole, and you were right: the test reached half of what it was written to protect, which for a PR whose entire value is a regression guard is the whole ball game. All five points are addressed in 241c9726, and the description has been rewritten.

## ChiR1 / the `getUnbatched` path never ran — fixed

Confirmed exactly as you both described. `settings.conf:1284` sets `utxostore_getBatcherSize = 4096`, `setup()` never overrides it, so `s.get` routed every one of those subtests to `getBatched` → `BatchDecorate`. Reverting the `getUnbatched` half on its own left all four green.

I took the direct-call option rather than a second store. `getUnbatched` and `BatchDecorate` are now each called directly, so neither the batcher's presence nor the field list can move what is covered. @ctnguyen's point about a developer with `getBatcherSize = 1` was the deciding factor — my first attempt at this reached `getUnbatched` through the public `Get` by adding `ConflictingChildren` to the field list, which works today but would relocate silently the moment `BatchDecorate` learned to serve that field. Direct calls have no such dependency.

I kept one case driving the public `Get`, purely to confirm the routing reaches the query at all, since the direct calls deliberately bypass it.

One small correction to the write-up, and it does not change the conclusion: the batched decision *was* deterministically pinned before this commit, by the explicit `BatchDecorate` subtest, which reaches `batchDecorateChunk` regardless of configuration. What was configuration-dependent was the three `Get` subtests. So it was one side unpinned rather than neither. The fix is the same either way.

## ChiR2 / `SubtreeIdxs` unpinned — fixed

Table-driven over both fields on both paths, so all four combinations are asserted. Your framing was the persuasive bit: "fixing only one would have left the batched path still wrong" applies just as much inside one condition as across two. Thanks for the `GetTransactionMeta.go:139-142` pointer — `SubtreeIdxs` being a real standalone selection rather than a hypothetical one is what makes it worth pinning.

## ChiR3 / duplicated condition and the parity claim — both fixed

Collapsed to `needsBlockIDsQuery` (`sql.go:1786`), used at both sites, near-verbatim from your suggestion. Duplicating that condition was the drift that produced the bug, so it should not have shipped as two copies.

The parity sentence is gone, and you were right that it was the more serious half of that point — a checked-in comment asserting something false about another backend is worse than a duplicated condition. `addAbstractedBins` expands only *from* `BlockIDs`, so aerospike still returns `BlockIDs == nil` for a heights-only request where SQL now fills all three. The description now says what is actually true: neither backend silently returns empty for a field you asked for. I also took your `Interface.go:412-414` reading — the fields parameter selects what to retrieve, and both stores honour that, so this was never an aerospike defect.

The shared field-expansion helper in `stores/utxo/fields` is the right long-term shape, but I agree it is out of scope for a SQL-only fix.

## ChiR4 / the grouping was undocumented — fixed

`Get`'s doc comment now states that the three placement fields travel together and are index-aligned, roughly as you drafted it. `parentTxHashes` is gone; good catch, it has not been a `fields.FieldName` in some time — `TxInpoints` is the field. Test renamed to `TestGetBlockPlacementFieldsIndependently`, since the old name described two of the three fields it covers.

## ChiR5 / the old behaviour was load-dependent — fixed, and this improved the PR

This was the most valuable point in the review and I had it wrong. I verified it: `sendGetBatch` (`sql.go:1493-1509`) unions the field sets across the batch and `batchDecorateChunk` decides on the union, so pre-fix a `BlockHeights`-only `Get` returned heights if and only if a co-batched request happened to ask for `BlockIDs`. With `getBatcherSize = 4096` that is every `Get` in the default configuration.

"Right sometimes, wrong sometimes, depending on what else is in flight" is a much stronger argument than "returns empty", and it also explains why this was never caught: a test against the old behaviour would have been flaky rather than red. It now leads the description.

## Verification

Each part was broken on its own, rather than checking the group fails together:

| Broken | Failed |
|---|---|
| `BlockHeights` term in the helper | both `BlockHeights` cases + public `Get` |
| `SubtreeIdxs` term in the helper | both `SubtreeIdxs` cases |
| `getUnbatched` site narrowed to `BlockIDs` | both unbatched cases |
| `batchDecorateChunk` site narrowed to `BlockIDs` | both batched cases + public `Get` |

All failures are `[]uint32(nil)` against an expected `[]uint32{0x63}`. The public `Get` case failing in the fourth experiment but not the third is independent confirmation of the routing you both described.

`go build ./...`, `go vet`, `golangci-lint` (0 issues) and `staticcheck` are clean; the new test passes under `-race`.

@ordishs — one environment difference worth noting on the postgres tests. You saw the package fully green; in my environment those ten `_Postgres` tests *fail* rather than skip, with `run postgres: generic container: get provider: rootless Docker not found`. Same ten names either side of this change, so unrelated to it, but the skip is evidently conditional on something my setup does not satisfy. Not this PR's problem, though it may be worth a look if others hit it.

## Follow-up, not in this PR

The reason this class of bug survived across both backends is that the shared conformance suite in `stores/utxo/tests/tests.go` only ever calls `db.Get(ctx, hash)` with no field list, which defaults to `MetaFieldsWithTx` — and that contains `BlockIDs`. So no backend has ever been tested with the field combination that exposes this. A "each field on its own" case there would cover both stores at once, and pairs naturally with @ordishs's shared field-expansion helper. Happy to raise it separately if you think it is worth doing.

ctnguyen (approved):
Review at commit [241c972](https://github.com/freemans13/teranode/commit/241c97267ade59aef94c47c895c1b0a698cdd892)

This round lands the fix well. The regression guard now has real teeth: `getUnbatched` and `BatchDecorate` are called directly, so no settings value can relocate what's covered, and each of the four gate terms is independently pinned — I reproduced the author's mutation table exactly, and every single-edit revert is caught. The duplicated condition is collapsed into one named `needsBlockIDsQuery` helper, the overstated "backends agree" parity claim is gone, and the `Get` doc comment now states the grouping invariant and drops the stale `parentTxHashes`. That's a clean, honest response to a review whose whole point was that the first version guarded only half of what it changed. What remains is small and all in prose: a couple of test-comment sentences that the refactor left behind, and the commit trailers.

## Status of previous points

ChiR5 is partial only because of prose: the PR description now leads with the load-dependence argument (done), but the checked-in test doc comment still characterises the old behaviour as "an empty slice and no error" with no load-dependence caveat — and that comment is the artifact that outlives the PR page. The commit-body half is cosmetic here since the repo squash-merges (branch commit bodies don't reach main's history), so the actionable residual is the test comment; see ChiR6.

| ChiR1 | Unbatched gate untested | YES | FIXED |
| ChiR2 | SubtreeIdxs unpinned on batched path | | FIXED |
| ChiR3 | Duplicated condition, parity claim | | FIXED |
| ChiR4 | Grouping invariant undocumented | | FIXED |
| ChiR5 | Old behaviour was load-dependent | | PARTIAL FIXED |

## Non-blocking issues

### ChiR6 - Test doc comment describes the pre-refactor gate shape and understates the old behaviour
*Partially overlaps [@ordishs's point "the 20-line test doc comment restates the PR description almost verbatim" (05-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1397#pullrequestreview-4863895348), and carries the residual of prior ChiR5.*

**Problem:** The test doc comment on `TestGetBlockPlacementFieldsIndependently` ends with: "The two paths reach the query through separate checks, and each check ORs the two fields separately, so any one of the four could be removed without the others noticing." The same commit collapsed both checks into the single `needsBlockIDsQuery` helper (`sql.go:1789`), so there is no longer a check that "ORs the two fields" — there is one OR expression plus two call sites. The mechanism the sentence describes is the pre-refactor shape. Separately, and carried over from ChiR5, the comment still says asking for either field alone "returned an empty slice and no error — indistinguishable from a transaction never mined", with no mention that on the batched path (every default `Get`, `getBatcherSize = 4096`) the answer was actually load-dependent — right iff a co-batched request happened to ask for `BlockIDs`.

**Why it matters:** The comment is the durable record of what this test protects — it lands in the tree and outlives the PR page, whereas the corrected description does not. A reader who takes the "each check ORs the two fields separately" sentence literally will go looking for two OR-ing checks and find one helper, and the "empty slice" framing reads as a reproducible failure rather than the flaky-not-red behaviour that explains why this survived. The conclusion the sentence draws — four independently-removable edit points — is still true and I verified it; only the stated mechanism is wrong.

**Fix:** Reword to the post-refactor shape and fold in the load-dependence caveat in one pass, e.g.

```go
// The helper is used at both read paths (getUnbatched and batchDecorateChunk) and
// ORs all three placement fields, so the field selection has four independently
// removable edit points: either OR term, and either call site. Each is pinned below.
//
// Before the fix, asking for BlockHeights or SubtreeIdxs alone returned an empty
// slice and no error on the unbatched path; on the batched path (every default Get,
// getBatcherSize = 4096) it returned them iff a co-batched request asked for
// BlockIDs — right sometimes, wrong sometimes, which is why a test against it would
// have been flaky rather than red.
```

Trimming the comment as @ordishs suggested would also remove the stale sentence, so the two fit together.

### ChiR7 - Commit bodies carry a non-standard session-URL trailer
**Problem:** Both branch commits (`813dd2eac`, `241c97267`) end with a `*-Session:` trailer carrying an external `https://…/code/session_<id>` URL. Across the last 400 commits on `main` that trailer key appears zero times; the `Co-authored-by:` trailer the same commits carry is by contrast an established style (~20 of the last 200).

**Why it matters:** The repo squash-merges, so the `*-Session:` trailer normally isn't carried into the merge commit, but it is visible permanently on the PR's commit list regardless. It is a non-standard trailer key that leaks an external, tool-specific URL into project history for no reader benefit.

**Fix:** Strip the `*-Session:` trailer from both commits. History hygiene only; the diff is unaffected. (The `Co-authored-by:` line is left to the author's own attribution convention — no change required.)

Follow-up, not a finding in this diff: `TxMetaCache.BatchDecorate` (`stores/txmetacache/txmetacache.go:567`) keys "not mined" on `BlockIDs` being nil, so it depends on the very "placement fields travel together" invariant this PR only half-declares. It is latent today and strictly improved by this change on SQL, but it is the strongest case for @ordishs's shared field-expansion helper in `stores/utxo/fields` plus an "each field on its own" case in the shared conformance suite (`stores/utxo/tests/tests.go`) — worth a separate issue.

## Recap

| ChiR6 | Stale test-comment mechanism + ChiR5 residual | | 20% |
| ChiR7 | Non-standard commit trailer | | 15% |

PR #1398, opened 2026-08-05
refactor(utxo): remove the unused BlockIDs field from utxo.Spend

freemans13:
## What happened

`utxo.Spend` carries a `BlockIDs` field, described in its comment as "the list of blocks the transaction has been mined into". No store has ever populated it and no caller has ever read it. The only code in the tree that touches it is `Clone`, which faithfully allocates a fresh slice and deep-copies a field that is always nil.

This came out of a call-site census of the whole `utxo.Store` interface, looking for places where the store returns or fetches data that nothing consumes.

Worth being explicit about what is *not* changing, because it is the more tempting target and it would be a bad idea. The per-input `[]*Spend` that `SpendAndCreate` returns is discarded at every success-path call site, which makes it look removable. It is not: it carries the per-input `Err` that distinguishes an output that is already spent from one that never existed, and callers depend on that distinction. Weakening it would let a store report generic absence where it currently reports `ErrSpent`, which would silently reject valid blocks on the ordinary block-arrival path. So the return stays exactly as it is; this removes one dead field inside it, so that the response carries only what something actually reads.

## The change

Deletes the field declaration and the corresponding branch in `Clone`. Eight lines, one file.

Nothing serialised changes either. The json tag is `blockIDs,omitempty` and the value is always nil, so the field has never appeared in any encoded `Spend`.

Note that `BlockIDsRemoval.BlockIDs` in the same file is a different type used by `RemoveBlockIDs`, and is untouched. So is `meta.Data.BlockIDs`, which is the one that genuinely carries this information.

## Testing

There is no test that can fail without this change, and I would rather say so than add one that passes either way. A field that nothing references cannot be observed at runtime, so the compiler is the proof: the whole tree builds, and `go vet ./...` compiles the test files too, so any lingering reference in a test would have surfaced as a build failure.

Limits worth stating: `go vet ./...` reports the same four pre-existing issues in `test/utils` either side of this commit, and `stores/utxo/sql` has ten pre-existing `_Postgres` test failures that need a live postgres server I did not have running. Both sets are identical before and after, but I have not run those particular tests green. `stores/utxo` itself passes.

### DIFF
--- a/docs/references/stores/utxo_reference.md
+++ b/docs/references/stores/utxo_reference.md
@@ -39,9 +39,6 @@ type Spend struct {
     // ConflictingTxID is the transaction ID that conflicts with this UTXO
     ConflictingTxID *chainhash.Hash `json:"conflictingTxId,omitempty"`
 
-    // BlockIDs is the list of blocks the transaction has been mined into
-    BlockIDs []uint32 `json:"blockIDs,omitempty"`
-
     // error is the error that occurred during the spend operation
     Err error `json:"err,omitempty"`
 }
--- a/stores/utxo/Interface.go
+++ b/stores/utxo/Interface.go
@@ -144,9 +144,6 @@ type Spend struct {
 	// ConflictingTxID is the transaction ID that conflicts with this UTXO
 	ConflictingTxID *chainhash.Hash `json:"conflictingTxId,omitempty"`
 
-	// BlockIDs is the list of blocks the transaction has been mined into
-	BlockIDs []uint32 `json:"blockIDs,omitempty"`
-
 	// error is the error that occurred during the spend operation
 	Err error `json:"err,omitempty"`
 }
@@ -180,11 +177,6 @@ func (s *Spend) Clone() *Spend {
 	if s.ConflictingTxID != nil {
 		clone.ConflictingTxID = &chainhash.Hash{}
 		*clone.ConflictingTxID = *s.ConflictingTxID
-	}
-
-	if s.BlockIDs != nil {
-		clone.BlockIDs = make([]uint32, len(s.BlockIDs))
-		copy(clone.BlockIDs, s.BlockIDs)
 	}
 
 	return clone


ordishs (approved):
Approving. The claim checks out and the removal is complete.

Verified on the PR head:

```
$ go build ./...
exit 0

$ go test ./stores/utxo/
ok  	github.com/bsv-blockchain/teranode/stores/utxo	1.107s

$ go vet -tags "aerospike,testtxmetacache" ./stores/utxo/...
exit 0
```

That last one is the run worth adding to the description. Plain `go vet ./...` does not compile build-tagged test files, and the aerospike spend tests are the most likely place a lingering `Spend.BlockIDs` reference could have hidden. The tagged vet closes that gap; the untagged one does not. I also reproduced the stated baseline: exactly the four pre-existing `test/utils` lock-copy findings, identical either side.

Every other `BlockIDs` in the tree is a distinct symbol — `meta.Data.BlockIDs`, `fields.BlockIDs`, `utxo.BlockIDsRemoval`, `aerospike.SpendState.BlockIDs`, `aerospike.LuaMapResponse.BlockIDs`, and the blockchain proto field. The serialisation claim holds too: `utxo.Spend` is never marshalled in `stores/utxo` and has no proto mapping, so an always-nil `omitempty` field could never have appeared on the wire. Docs coverage is complete — `utxo_reference.md` was the only place the field was documented, and `docs/topics/stores/utxo.md:240` is `meta.Data`, correctly untouched.

Agreed on adding no test, and on saying so rather than adding one that passes either way.

One follow-up, same species, one layer down — and unlike the struct field this one is not free:

- `stores/utxo/aerospike/spend_expressions.go:246` — `buildSpendOperations` issues `GetBinOp(fields.BlockIDs.String())` on every spend operation.
- `spend_expressions.go:42-47` — `parseSpendState` decodes that bin into `SpendState.BlockIDs`, allocating a `[]uint32`.
- Nothing reads it. `parseSpendState` has one caller (`spend_expressions.go:461`) and it consumes `TotalExtraRecs`, `External`, `Old/NewDeleteAtHeight`, `SpentUtxos`, `RecordUtxos` — never `BlockIDs`.

Same story on the UDF transport: `teranode.lua:454-455` puts `blockIDs` into the `spendMulti` response, but `parseLuaMapResponseInto` is only called from `set_mined.go:504`, so on the spend path those bytes ship and get dropped. `LuaMapResponse.BlockIDs` is genuinely live for `SetMinedMulti`, so only the `SpendState` field and the `GetBinOp` are removable.

That is a per-spend bin read, msgpack decode and allocation on the hot path feeding a dead field. Worth its own PR with a real before/after rather than a rider on this one — but worth filing while the census context is fresh.

ctnguyen (approved):
Review at commit [f912c59](https://github.com/freemans13/teranode/commit/f912c59ca0f3120fe3d30d3e4d5f5cfb8f010d21)

Clean, well-scoped deletion, and I could not find a live reference anywhere: every surviving `BlockIDs` in the tree is a different symbol, nothing builds a `Spend` positionally, and nothing marshals one — so the always-nil field really was unobservable, and the keyed literal in `Clone` means dropping it can't silently shift anything. I especially liked the description spelling out what it deliberately is *not* touching; the note on `[]*Spend` carrying the per-input `Err` that separates `ErrSpent` from plain absence is exactly what stops someone deleting the return next. Everything below is optional and none of it should hold up the merge.

## Non-blocking issues

### ChiR1 - Description overstates what plain `go vet ./...` proves
*Same as [@ordishs's point "the tagged vet closes that gap" (05-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1398#pullrequestreview-4863921421).*
**Problem:** The Testing section argues "`go vet ./...` compiles the test files too, so any lingering reference in a test would have surfaced as a build failure." That inference doesn't hold: `go vet` skips files excluded by a build constraint, so a reference behind `//go:build aerospike` would not have surfaced. The conclusion is still correct — the gap it left was narrow, partly because several aerospike files that read as tagged actually open with a double-commented `// //go:build aerospike`, so they compile unconditionally and were covered by the untagged run regardless.
**Why it matters:** The claim is load-bearing here, since the whole argument for adding no test is that the compiler is the proof. A proof that doesn't cover build-tagged files is the reasoning that gets reused on the next deletion — one where the constraint is live and the missed reference is real.
**Fix:** Swap the inference for the command @ordishs ran, which is reproducible and does close the gap:

```
$ go vet -tags "aerospike,testtxmetacache" ./stores/utxo/...
exit 0
```

### ChiR2 - `Spend.Clone` has no production caller left
**Problem:** `Clone` is the method this diff edits, and nothing outside tests calls it. Enumerating every distinct `.Clone()` receiver in the tree turns up exactly one that resolves to `utxo.Spend.Clone` — `spend.Clone()` at `stores/utxo/aerospike/aerospike_server_test.go:209`, where the clone is only used as a `GetSpend` lookup key. Nothing in `services/`, `model/`, `cmd/` or any store reaches it. Flagging this as a census datapoint, not as something to change here: dropping an exported method is a wider API decision than this PR is making, and it shouldn't ride along on a field removal.
**Why it matters:** It's the same species the census is hunting, one level up, and it carries the maintenance cost that let the removed field survive since it landed — `Clone` is what a contributor must remember to update when a field is added to `Spend`, and nothing exercises it if they forget.
**Fix:** Nothing in this PR. Either fold it into the same follow-up as the aerospike `SpendState` finding, or, if it's being kept deliberately for future 2PC work, record that in the doc comment so the next census doesn't have to re-derive that it's unused.

### ChiR3 - Iterator signature in the touched doc is wrong
**Problem:** `docs/references/stores/utxo_reference.md:147` documents `UnminedTxIterator.Next` as returning `(*UnminedTransaction, error)`; the code returns `([]*UnminedTransaction, error)` and batches. Same file, two blocks below the one this PR corrects. (`IgnoreFlags` at lines 160-163 is also stale — it omits `SkipUTXOHashCheck` — but that one is fair to leave for a docs sweep.)
**Why it matters:** This is the drift that actually costs someone time: a reader coding against the doc writes a single-item loop against a batching iterator and gets a type error at best, a silently truncated scan at worst if they index `[0]`. Raising it because this PR's second commit exists precisely to keep this file matching the structs, so it's a natural rider rather than unrelated cleanup.
**Fix:** One line, take it or leave it:

```go
// Next advances the iterator and returns a batch of unmined transactions, or nil if iteration is done
Next(ctx context.Context) ([]*UnminedTransaction, error)
```

## Recap

| ChiR1 | Untagged vet claim in description | | 25% |
| ChiR2 | `Spend.Clone` production-dead | | 12% |
| ChiR3 | Stale iterator signature in doc | | 10% |

PR #1453, opened 2026-08-05
fix(blockvalidation): verify a served block is the one requested, bound peer error bodies

liam:
Two independent hardening fixes on the peer-fetch path, both confirmed against `main` before writing any code.

Closes bitcoin-sv/teranode#4744. Addresses the second finding of bitcoin-sv/teranode#4742.

## 1. `fetchSingleBlock` accepted any block the peer chose to return

`fetchSingleBlock` requested `/block/<hash>` from a peer-supplied `baseURL` and returned whatever the body parsed to, with no check that it was the block asked for. The sibling batch path already does exactly this check in `verifyBlockHeaders`, twelve lines below.

The consequence is worse than a mismatched return value, because callers treat the requested and served identities interchangeably:

- `addBlockToPriorityQueue` sets the in-flight marker on the **requested** hash (`Server.go`, `processBlockNotify.Set(*blockFound.hash, ...)`)
- the item handed to `catchupCh` carries the **served** block
- all nine downstream `Delete` calls key on `c.block.Hash()`, the served block

So when served differs from requested, the marker under the requested hash is never deleted. From then on the enqueue gate takes its suppression branch for that hash, which only appends to `catchupAlternatives`, and that list is drained solely from a `catchupCh` item for that hash which now never arrives. The block stops being processable, and the peer chose which hash to do that to.

One check upstream collapses the whole chain: once served is guaranteed to equal requested, the marker leak, the suppression and the stall all become unreachable. That is why this fixes the reported issue without touching the identity handling in the caller, which is better addressed with bitcoin-sv/teranode#4746.

**Also disables touch-on-hit on `processBlockNotify`.** The 10 minute TTL exists as the safety net for exactly the missed-`Delete` case above, and a safety net only holds if reads cannot keep pushing the expiry out. The enqueue gate reads this entry on every duplicate announcement, so with touch-on-hit a stream of duplicates held the suppression open past the TTL indefinitely. `blockCatchupAttempts` already disables it, with a comment describing the same hazard.

This is a trade rather than pure hardening, and the cost side is worth stating: a catchup that legitimately runs longer than 10 minutes now loses its suppression marker, so the next duplicate announcement re-enqueues onto `catchupCh`. The blast radius is bounded by `CatchupChBufferSize`, and the duplicate hits `ErrCatchupInProgress` and returns without consuming an attempt, so the exchange is a bounded re-enqueue on slow catchups against an unbounded suppression window on fast ones.

## 2. The non-2xx error body was read unbounded, defeating `DoHTTPRequestBounded`

`buildHTTPError` read the entire failed-response body with `io.ReadAll` and interpolated it into the error message.

That path is reached from `DoHTTPRequestBounded` too, because a non-2xx status short-circuits into `buildHTTPError` before the caller's cap is ever applied. A hostile peer therefore bypassed the cap completely: answer with an error status, then stream forever. `DoHTTPRequestBounded`'s own doc comment says callers fetching peer-controlled data must bound the allocation, so the error path was quietly exempting itself from the guarantee the function advertises.

Now bounded at 2KB and `%q`-escaped. The escaping is the second half of the same finding: the snippet is logged verbatim and forwarded to the peer registry, so bounding the length while leaving the bytes raw still let a peer embed newlines to forge log lines, or terminal escapes. `%q` also keeps the message single-line, which the project convention requires. The cap is deliberately small because `%q` expands its input, up to four characters per control byte, so the constant bounds bytes read rather than the length of the resulting message.

## Verification

```
go build ./...                                                      -> clean
go vet ./util/ ./services/blockvalidation/                           -> clean
gofmt -l <changed files>                                             -> clean
gci diff --skip-generated -s standard -s default <changed files>     -> clean

go test -count=1 ./util/
  ok  github.com/bsv-blockchain/teranode/util                        7.690s
go test -count=1 -timeout 1800s -tags testtxmetacache ./services/blockvalidation/...
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation    274.861s
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation/catchup  13.338s
```

All five new assertions are mutation-verified, each failing its intended test and only that test:

| Mutation | Result |
|---|---|
| disable the requested-hash check | only `TestFetchSingleBlock_RejectsSubstitutedBlock` fails |
| make the hash check always reject | only the pre-existing `Successful_Fetch` positive controls fail, the new test still passes |
| remove `WithDisableTouchOnHit` | only `TestServer_processBlockNotifyDoesNotTouchOnHit` fails |
| restore the unbounded error-body read | only `TestBuildHTTPErrorBodyIsBounded` fails, both subtests |
| revert `%q` to `%s` | `TestBuildHTTPErrorBodyIsEscaped` fails, plus the three pre-existing empty-body assertions |

The error-body test asserts on markers at **both** ends of an oversized body, so it distinguishes a real truncation from a body that merely happened to be short, and it covers the bounded caller separately to pin that an error status cannot bypass the cap.

`-race` and the full repository suite were not run locally.

## Residual risk and scope

The 2KB truncation and the `%q` escaping apply to every HTTP caller in the repo, not just the peer-fetch paths. Error bodies are diagnostic strings and nothing parses them, so the exposure is a shorter, quoted message when a server returns more than 2KB of error text.

Deliberately **not** included, since each needs a number agreed rather than a mechanical fix:

- **First finding of bitcoin-sv/teranode#4742**, the unbounded `io.ReadAll` on the two block downloads (`/block/` and `/blocks/`). Worth noting that the issue's suggested fix, converting them to `DoHTTPRequestBounded`, is not the right shape: BSV has no maximum block size, so there is no byte cap to choose. It also is not needed. A `model.Block` on the wire is a header, three varints, N subtree hashes and a coinbase, not the transactions, so the response size is unrelated to the size of the block, and `readBlockFromReader` already caps the one attacker-controlled dimension (it bounds the speculative `Subtrees` pre-allocation at 1024 and relies on `append`). The actual defect is that both fetchers buffer the whole body before parsing something that is already parseable from a stream: `fetchBlocksBatch` does `io.ReadAll` and then wraps the result in `bytes.NewReader` for `model.NewBlockFromReader`. Streaming the response body into the parser bounds the allocation structurally with no constant to agree on. Two details for whoever picks it up: `RecordBytesDownloaded` currently uses `len(blockBytes)` and would need a counting reader, and the batch loop reads until EOF so it needs to stop at the `n` blocks it asked for.
- **bitcoin-sv/teranode#4743**, `context.WithoutCancel` stripping the deadline from the subtree_data fetch. Confirmed: `WithoutCancel` yields a nil `Done()` channel, so the retry loop's only abort guard can never fire and the loop runs to all 6 attempts, each installing a fresh 10 minute streaming timeout. Note that fix 2 above substantially defuses it, since the uncapped error body was what let an attempt consume its full 10 minutes.
- **bitcoin-sv/teranode#4746**, the unbounded `catchupAlternatives` list. Fix 1 does not close it: duplicate announcements of a genuinely valid hash from distinct peers still accumulate a full `*model.Block` each, drained serially on the single goroutine that services `catchupCh`.

One thing found while verifying that none of the issues mention: `services/blockvalidation/catchup/helpers.go` fetches headers with the unbounded `DoHTTPRequest` as well, a third site alongside the two block downloads.

### DIFF
--- a/services/blockvalidation/Server.go
+++ b/services/blockvalidation/Server.go
@@ -376,7 +376,14 @@ func New(
 		// when catchup completes or fails, but a missed Delete on any error/early-return
 		// branch would otherwise leak the entry permanently. Mirrors catchupAlternatives,
 		// the sibling cache for the same in-flight block.
-		processBlockNotify:  ttlcache.New[chainhash.Hash, bool](ttlcache.WithTTL[chainhash.Hash, bool](10 * time.Minute)),
+		processBlockNotify: ttlcache.New[chainhash.Hash, bool](
+			ttlcache.WithTTL[chainhash.Hash, bool](10*time.Minute),
+			// Do not extend the window on reads, for the same reason as
+			// blockCatchupAttempts below: the enqueue gate reads this entry on every
+			// duplicate announcement, so touch-on-hit would let a stream of duplicates
+			// hold the suppression open past the safety-net TTL.
+			ttlcache.WithDisableTouchOnHit[chainhash.Hash, bool](),
+		),
 		catchupAlternatives: ttlcache.New[chainhash.Hash, []processBlockCatchup](ttlcache.WithTTL[chainhash.Hash, []processBlockCatchup](10 * time.Minute)),
 		blockCatchupAttempts: ttlcache.New[chainhash.Hash, int](
 			ttlcache.WithTTL[chainhash.Hash, int](10*time.Minute),
--- a/services/blockvalidation/Server_test.go
+++ b/services/blockvalidation/Server_test.go
@@ -512,6 +512,30 @@ func TestServer_processBlockNotifyHasTTL(t *testing.T) {
 	require.True(t, item.ExpiresAt().After(time.Now()), "processBlockNotify entry expiry must be in the future")
 }
 
+// TestServer_processBlockNotifyDoesNotTouchOnHit pins the companion property to the
+// TTL above: the safety-net expiry only helps if reads cannot keep pushing it out.
+// The enqueue gate in addBlockToPriorityQueue reads this entry on every duplicate
+// announcement of an in-flight block, so with touch-on-hit a peer that keeps
+// announcing would hold the suppression open indefinitely and the TTL would never
+// fire. blockCatchupAttempts already disables touch-on-hit for the same reason.
+func TestServer_processBlockNotifyDoesNotTouchOnHit(t *testing.T) {
+	tSettings := test.CreateBaseTestSettings(t)
+	tSettings.ChainCfgParams = &chaincfg.MainNetParams
+
+	s := New(ulogger.TestLogger{}, tSettings, nil, nil, nil, nil, nil, nil, nil, nil)
+
+	var hash chainhash.Hash
+	originalExpiry := s.processBlockNotify.Set(hash, true, ttlcache.DefaultTTL).ExpiresAt()
+
+	// Any non-zero gap is enough: touch-on-hit recomputes the expiry from "now", so a
+	// read after the clock has moved would push it out measurably.
+	time.Sleep(10 * time.Millisecond)
+
+	item := s.processBlockNotify.Get(hash)
+	require.NotNil(t, item, "entry must still be present")
+	require.Equal(t, originalExpiry, item.ExpiresAt(), "reads must not extend the suppression window")
+}
+
 func TestServer_processBlockFoundChannel(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
--- a/services/blockvalidation/get_blocks.go
+++ b/services/blockvalidation/get_blocks.go
@@ -986,6 +986,17 @@ func (u *Server) fetchSingleBlock(ctx context.Context, hash *chainhash.Hash, pee
 			hash.String(), len(blockBytes))
 	}
 
+	// The peer chooses the response body, so a well-formed block is not necessarily the
+	// block that was asked for. Callers treat the two identities interchangeably (the
+	// in-flight marker is keyed on the requested hash while every later lookup keys on
+	// the served block), so a substitution would leave that marker undeletable and
+	// suppress every subsequent honest announcement of the requested hash. Reject the
+	// substitution here instead, mirroring verifyBlockHeaders on the batch path.
+	if !block.Hash().IsEqual(hash) {
+		return nil, errors.NewProcessingError("[catchup:fetchSingleBlock][%s] peer served block %s for a different hash",
+			hash.String(), block.Hash().String())
+	}
+
 	// Reputation is credited post-validation in validateBlocksOnChannel via reportValidBlockForPeers
 
 	return block, nil
--- a/services/blockvalidation/get_blocks_test.go
+++ b/services/blockvalidation/get_blocks_test.go
@@ -898,6 +898,44 @@ func TestFetchSingleBlock_CurrentBehavior(t *testing.T) {
 	})
 }
 
+// TestFetchSingleBlock_RejectsSubstitutedBlock pins the requested-hash check.
+//
+// The response body is entirely peer-chosen, so parsing to a well-formed block says
+// nothing about it being the block that was asked for. Returning a substituted block
+// is not a cosmetic mismatch: callers key the in-flight catchup marker on the
+// requested hash but every later lookup on the served block, so accepting one leaves
+// that marker undeletable and silently suppresses every subsequent honest
+// announcement of the requested hash.
+func TestFetchSingleBlock_RejectsSubstitutedBlock(t *testing.T) {
+	suite := NewCatchupTestSuite(t)
+	defer suite.Cleanup()
+
+	blocks := testhelpers.CreateTestBlockChain(t, 3)
+	requestedHash := blocks[2].Header.Hash()
+	substitutedHash := blocks[1].Header.Hash()
+	require.False(t, requestedHash.IsEqual(substitutedHash), "test needs two distinct blocks")
+
+	httpmock.ActivateNonDefault(util.HTTPClient())
+	defer httpmock.DeactivateAndReset()
+
+	// The peer answers the request for blocks[2] with a valid, parseable block that is
+	// simply a different one.
+	substitutedBytes, err := blocks[1].Bytes()
+	require.NoError(t, err)
+
+	httpmock.RegisterResponder(
+		"GET",
+		fmt.Sprintf("http://test-peer/block/%s", requestedHash.String()),
+		httpmock.NewBytesResponder(200, substitutedBytes),
+	)
+
+	fetchedBlock, err := suite.Server.fetchSingleBlock(suite.Ctx, requestedHash, "12D3KooWL1NF6fdTJ9cucEuwvuX8V8KtpJZZnUE4umdLBuK15eUZ", "http://test-peer")
+	require.Error(t, err)
+	require.Nil(t, fetchedBlock, "a substituted block must not reach the caller")
+	require.Contains(t, err.Error(), "for a different hash")
+	require.Contains(t, err.Error(), substitutedHash.String(), "error should name what was actually served")
+}
+
 // Phase 2: Tests for optimized batch fetching and ordered delivery
 func TestFetchBlocksConcurrently_OptimizedBehavior(t *testing.T) {
 	t.Run("Ordered_Delivery_With_Batching", func(t *testing.T) {
--- a/util/http.go
+++ b/util/http.go
@@ -440,12 +440,22 @@ func executeHTTPRequest(ctx context.Context, cancelFn context.CancelFunc, rawURL
 	return resp.Body, cancelFn, nil
 }
 
+// maxHTTPErrorBodyBytes bounds how much of a non-2xx response body is read for the
+// error message. The body is peer-supplied and is read on every failure path,
+// including the ones reached from DoHTTPRequestBounded: without a bound, a hostile
+// peer defeats that function's cap simply by answering with an error status and then
+// streaming indefinitely. An error message only needs enough to be diagnosable.
+const maxHTTPErrorBodyBytes = 8 * 1024
+
 // buildHTTPError constructs an appropriate error from a non-OK HTTP response.
 //
 // The error type is chosen to let callers branch with errors.Is:
 //   - 404 → ErrNotFound
 //   - 503 → ErrServiceUnavailable (typically retryable; see DoHTTPRequestBodyReaderWithRetry)
 //   - other → generic ServiceError
+//
+// The body is read up to maxHTTPErrorBodyBytes; anything beyond that is discarded
+// rather than retained in the error string.
 func buildHTTPError(resp *http.Response, rawURL string) error {
 	errFn := errors.NewServiceError
 	switch resp.StatusCode {
@@ -460,7 +470,7 @@ func buildHTTPError(resp *http.Response, rawURL string) error {
 			_ = resp.Body.Close()
 		}()
 
-		b, readErr := io.ReadAll(resp.Body)
+		b, readErr := io.ReadAll(io.LimitReader(resp.Body, maxHTTPErrorBodyBytes))
 		if readErr != nil {
 			return errFn("http request [%s] returned status code [%d]", rawURL, resp.StatusCode, readErr)
 		}
--- a/util/http.go
+++ b/util/http.go
@@ -445,7 +445,11 @@ func executeHTTPRequest(ctx context.Context, cancelFn context.CancelFunc, rawURL
 // including the ones reached from DoHTTPRequestBounded: without a bound, a hostile
 // peer defeats that function's cap simply by answering with an error status and then
 // streaming indefinitely. An error message only needs enough to be diagnosable.
-const maxHTTPErrorBodyBytes = 8 * 1024
+//
+// Kept small because the snippet is %q-escaped below, and escaping expands: a body of
+// control bytes becomes up to four characters each, so this is the bound on bytes read,
+// not on the length of the resulting message.
+const maxHTTPErrorBodyBytes = 2 * 1024
 
 // buildHTTPError constructs an appropriate error from a non-OK HTTP response.
 //
@@ -455,7 +459,10 @@ const maxHTTPErrorBodyBytes = 8 * 1024
 //   - other → generic ServiceError
 //
 // The body is read up to maxHTTPErrorBodyBytes; anything beyond that is discarded
-// rather than retained in the error string.
+// rather than retained in the error string. The snippet is %q-escaped because this
+// message is logged verbatim and forwarded to the peer registry: raw peer bytes would
+// otherwise let a peer embed newlines to forge log lines, or terminal escapes, and
+// would break the single-line log convention.
 func buildHTTPError(resp *http.Response, rawURL string) error {
 	errFn := errors.NewServiceError
 	switch resp.StatusCode {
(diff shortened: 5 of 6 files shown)


ctnguyen (approved):
Review at commit [06c08ec](https://github.com/liam/teranode/commit/06c08eccc773a136c83baddbe0e8954bda0daf78)

Two small, well-argued hardening fixes on the peer block-fetch path, and both land where they should. Putting the requested-hash check inside `fetchSingleBlock` covers all three call sites (`Server.go:925`, `:1479`, `:2128`) in one place and mirrors `verifyBlockHeaders` on the batch path, so the marker-leak/stall chain collapses at a single point — and it's parity-safe against the bitcoin-sv block validator, which I confirmed in code (the detail lives in ChiR1). The `WithDisableTouchOnHit` change is real, not cosmetic: the enqueue-gate read is a bare `processBlockNotify.Get` at `Server.go:2175` with no accompanying `Set`, so unlike the `catchupAlternatives` no-op that #4746 settled, disabling touch-on-hit here genuinely stops a duplicate stream from pushing the safety-net TTL out (confirmed against ttlcache v3.3.0). The error-body bound is a clean one-liner that closes the bypass, and the tests pin exactly what their comments claim, including the mutation behaviour. The two notes below are hardening deltas, not merge blockers.

## Non-blocking issues

### ChiR1 — Mismatch is a plain `ProcessingError`, so the priority-queue path can't fail over to an alternative source

**Problem:** The rejection returns `errors.NewProcessingError(…"peer served block %s for a different hash"…)` (`get_blocks.go:995-998`). At `processBlockWithPriority` the failover guard is `errors.IsNetworkError(err) || errors.IsMaliciousResponseError(err)` (`Server.go:2285`), and a `ProcessingError` satisfies neither — its code is in neither set and its message contains none of the trigger substrings (`errors/error_utils.go:95-102`, `:147-164`). So when the primary peer answers a GET for `H` with a valid block that simply isn't `H`, teranode returns the error up without trying an alternative source it may already hold in the priority queue. An ordinary network fetch failure, by contrast, matches on the `"http"` substring and does fail over.

**Why it matters:** Recovery is slower than it needs to be, but it is not a stall and not a correctness bug — and that conclusion is the block-level parity check, so it belongs here. bitcoin-sv carries a single block identity, keyed on the block's own hash (`net_processing.cpp:4051`, `block_download_tracker.cpp:73`), so the requested-vs-served split behind #4744 cannot arise there at all; teranode's is a self-inflicted divergence this PR papers over correctly. Rejecting a mismatched response does not change block-*acceptance* semantics: the reject writes no per-hash state (site 2128 returns at `Server.go:2142`, before `processBlockNotify.Set` at :2203, the `catchupAlternatives` append and `recordCatchupAttempt`; sites 925/1479 set nothing), `GetBlockExists(H)` stays false, so the same block stays fetchable and acceptable later under its own hash — a valid block can never be made permanently unacceptable by this check. teranode is only enforcing the request/response contract of a hash-specific HTTP GET, which bitcoin-sv's unsolicited p2p push doesn't have (there an unrequested valid block is kept, `net_processing.cpp:4045-4049`). Do not push this toward a reputation/DoS charge either: bitcoin-sv DoS-scores a peer only for an *invalid* block, never for a merely-mismatched one (`block_download_tracker.cpp:113-128`), and a mismatch has benign causes — a caching proxy, a stale response after a reorg, a DataHub serving a different fork's block. (see [bitcoin-sv/teranode#4744](https://github.com/bitcoin-sv/teranode/issues/4744))

**Fix:** If failover is wanted, engage it on the `processBlockWithPriority` path without also flagging the peer malicious there. `NewNetworkError` makes `IsNetworkError` true (failover fires at `:2285`) while `IsMaliciousResponseError` stays false:
```go
if !block.Hash().IsEqual(hash) {
    return nil, errors.NewNetworkError("[catchup:fetchSingleBlock][%s] peer served block %s for a different hash",
        hash.String(), block.Hash().String())
}
```
Scope note: this governs only the failover guard at `:2285`; it does not make the reject reputation-free everywhere. The `addBlockToPriorityQueue` call site already records a catchup failure for *any* fetch error regardless of class (`Server.go:2132-2136`), so a mismatch on that path charges the peer today and would continue to — that pre-existing attribution is out of this PR's scope. Accepting the current behaviour is also defensible given the self-heal. Durability note (not a defect here; #4746 is rightly deferred): the reference avoids this class structurally by keying all in-flight bookkeeping on the served block's own hash — the single-identity shape the deferred caller-side unification should take.

### ChiR2 — The error body is bounded in size but not sanitized before it is logged and shipped to the peer registry

**Problem:** `buildHTTPError` now caps the read at 8KB but still interpolates the raw peer bytes straight into the error string via `[%s]` / `string(b)` (`util/http.go:479`). That string is logged verbatim and shipped over gRPC into the peer registry via `reportCatchupError` on the fetch paths, so a peer controls newlines, control characters and terminal escapes in operator logs and registry entries.

**Why it matters:** [bitcoin-sv/teranode#4742](https://github.com/bitcoin-sv/teranode/issues/4742)'s proposed fix for exactly this finding asked to "sanitize/truncate before interpolating it into the error string" — this PR does the truncate half but not the sanitize half, leaving the log/registry-injection surface open. Embedded newlines also break the project convention that log messages stay single-line. It is low severity (nothing parses the message), which is why it is non-blocking. (Draining the remainder before `Close`, the other clause in that issue, is deliberately *not* worth doing: against an endless hostile body a full drain would reintroduce the very unbounded read the cap removes, so losing connection reuse on an oversized error body is the correct trade.)

**Fix:** Escape the snippet to a single-line, printable form rather than interpolating raw bytes. `%q` handles newlines, tabs, other control bytes and invalid UTF-8 in one step:
```go
// %q escapes newlines/tabs/control bytes and invalid UTF-8, so a peer-controlled body
// cannot forge log lines or inject terminal escapes, and the message stays single-line.
return errFn("http request [%s] returned status code [%d] with body %q", rawURL, resp.StatusCode, string(b))
```

## Recap

|---|---|---|---|
| ChiR1 | Mismatch blocks alternative-source failover | | 28% |
| ChiR2 | Error body unsanitized in logs | | 20% |

ordishs (approved):
## Overview

Two independent hardening fixes on the peer block-fetch path:

1. **`fetchSingleBlock` now rejects a block whose hash != the requested hash** (`get_blocks.go:995`), plus `processBlockNotify` gets `WithDisableTouchOnHit`.
2. **The non-2xx error body is bounded at 8KB** (`util/http.go`), closing a bypass of `DoHTTPRequestBounded`'s cap.

Both diagnoses hold up against the code. I verified the three `fetchSingleBlock` call sites (`Server.go:925`, `:1479`, `:2128`) and the marker/delete asymmetry described: `addBlockToPriorityQueue` sets the marker on `blockFound.hash` (`:2203`) while every `Delete` keys on `c.block.Hash()` — so an upstream identity check does collapse the whole chain. `verifyBlockHeaders` (`get_blocks.go:1013`) does exactly this check on the batch path, so the fix is consistent with the sibling code rather than novel.

Independent verification on this branch:

```
gofmt -l <6 changed files>                         -> clean (exit 0)
go vet ./util/ ./services/blockvalidation/         -> clean (exit 0)

go test -count=1 -run 'TestBuildHTTPErrorBodyIsBounded|TestDoHTTPRequestNotFound|TestDoHTTPRequestServerError|TestDoHTTPRequestBounded' ./util/
  ok  github.com/bsv-blockchain/teranode/util  0.856s

go test -count=1 -tags testtxmetacache -run 'TestFetchSingleBlock|TestServer_processBlockNotify' ./services/blockvalidation/
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation  1.185s
```

Full suite and `-race` not re-run here — as the description already states.

## Correctness

- **Fix 1 is minimal and complete for its stated scope.** All three callers get the guarantee, so the marker key and the `catchupCh` item key are now provably the same hash. Rejecting rather than reconciling is the right call; the caller-side identity handling is genuinely a separate change.
- **The touch-on-hit removal is better justified than the description says.** `processCatchupChItem` has a branch (`Server.go:1822`) that logs "requeueing" on `ErrCatchupInProgress` and then just `return`s — no requeue, no `Delete`. The TTL is the *only* cleanup for that path, so a stream of duplicate announcements previously held it open indefinitely. This change makes the safety net actually load-bearing there.
- **One behavioural consequence not acknowledged:** a legitimately in-flight catchup running longer than 10 minutes now loses its suppression marker, so the next duplicate announcement re-enqueues onto `catchupCh`. Blast radius is small (bounded by `CatchupChBufferSize`; the duplicate hits `ErrCatchupInProgress` and returns without consuming an attempt), but the change is framed as pure hardening and this is a real trade — worth a line in the description.
- **Asymmetry left behind:** `catchupAlternatives` keeps touch-on-hit, and the suppression branch `Set`s it on every duplicate (`:2188`/`:2193`), refreshing its TTL regardless. So an alternatives list — each entry a full `*model.Block` — can now outlive its `processBlockNotify` marker. It converges (the next enqueue re-sets the marker, completion deletes both), so not a leak, but disabling touch-on-hit there too would make the pair symmetric and match the comment's claim that they mirror each other.

## Security

- **`%s` on peer-controlled bytes is a log-injection vector, and this is the natural place to fix it.** `buildHTTPError` still interpolates up to 8KB of attacker-chosen bytes verbatim into the error string, which is then logged (`Server.go:2130` -> `logger.Errorf`). A peer can embed newlines and forge log lines — and the project convention is explicitly "log messages: always a single line". Suggest `%q` instead of `%s`, and consider dropping the cap to 1–2KB; a quoted 1KB excerpt is just as diagnosable and cannot break out of its line.
- **Reputation reporting for the substitution is not uniform.** `addBlockToPriorityQueue` degrades the peer on fetch error (`:2131`–`:2135`), but the other two call sites do not. Serving a different block for an explicit `/block/<hash>` request is a provable protocol violation, and `reportCatchupMalicious` already exists for that class (`catchup_get_block_headers.go:328` uses it for malformed header responses). Reporting once inside `fetchSingleBlock` — `peerID` is already in scope — would cover all three sites and make the attack cost something.

## Style / nits

- `if b != nil` (`util/http.go:477`) is always true: `io.ReadAll` returns a non-nil empty slice, so an empty error body yields `with body []`. Pre-existing, not a regression, but the line is being touched — `len(b) > 0` is the intended predicate.
- Connection reuse: closing an undrained body prevents keep-alive reuse, but only for error bodies over 8KB (`ReadAll` on the `LimitReader` still hits the real EOF below the cap). Not a concern.

## Test coverage

Both new tests are well-constructed — the head/tail marker pair genuinely distinguishes truncation from a short body, and the bounded-caller subtest pins the specific bypass. The mutation table is the right kind of evidence.

Two gaps:

- **No test pins the invariant the fix actually protects.** Both new tests sit upstream of the bug: one proves the fetch rejects, none proves the marker and `catchupCh` keys agree. Since the stall issue is closed on the strength of that downstream reasoning, a small server-level test asserting the hash placed on `catchupCh` equals the `processBlockNotify` key would pin it against a future regression in the caller — which is exactly where the identity handling is being deferred to.
- **`NewCatchupTestSuite` builds `processBlockNotify` with no options** (`catchup_test_suite.go:146`) — no TTL, touch-on-hit enabled. Pre-existing, but it now diverges from production in the specific dimension this PR hardens, so any future marker-lifecycle test written against the suite would pass for the wrong reason. Worth aligning the suite or noting it there.

## Verdict

Both fixes are correct, minimal, and in keeping with the surrounding code. Nothing blocking.

The one change I would like before merge is `%q` in `buildHTTPError` — the PR bounds the peer-controlled body but leaves it unescaped on a path that feeds the logger, which is the same class of exposure being fixed. The peer reporting and the long-catchup note are follow-ups rather than blockers.

liam:
## Addressed in f8bd8dc

Thanks both. You independently converged on the escaping, which is a good signal that I stopped half way through the finding: I bounded the allocation and left the bytes raw on a path that feeds the logger, which is the same class of exposure the commit claimed to fix.

### Done: `%q` escaping, plus a smaller cap

`buildHTTPError` now interpolates with `%q`. `TestBuildHTTPErrorBodyIsEscaped` asserts the message carries no raw newline, ESC or NUL while the content survives in escaped form, using a body deliberately well under the cap so it can only fail on escaping and never on truncation. Reverting `%q` to `%s` fails it, along with the three pre-existing empty-body assertions that now read as a quoted empty string.

I also dropped the cap from 8KB to 2KB, which neither of you asked for in quite these terms. `%q` expands its input, up to four characters per control byte, so an all-control-byte body at the old cap could have produced a 32KB message. The constant bounds bytes read, not the length of the result, so it needs to be smaller once escaping is in play. 2KB quoted is still ample to diagnose from.

Also agreed on the `Close`-without-drain point: against an endless hostile body a full drain would reintroduce exactly the unbounded read the cap removes, so losing keep-alive reuse on an oversized error body is the right trade.

### You two disagree on the reputation question, and I would rather you settle it than pick for you

Both reviews reach the opposite conclusion on whether a hash mismatch should cost the peer anything:

> **ChiR1:** Do not push this toward a reputation/DoS charge either: bitcoin-sv DoS-scores a peer only for an *invalid* block, never for a merely-mismatched one (`block_download_tracker.cpp:113-128`), and a mismatch has benign causes: a caching proxy, a stale response after a reorg, a DataHub serving a different fork's block.

> **Reputation reporting is not uniform:** Serving a different block for an explicit `/block/<hash>` request is a provable protocol violation, and `reportCatchupMalicious` already exists for that class. Reporting once inside `fetchSingleBlock` would cover all three sites and make the attack cost something.

You do agree on the diagnosis, that the current `ProcessingError` gives neither failover nor uniform attribution, and you agree the current behaviour is defensible in the meantime. The split is purely on whether the substitution is protocol violation or benign divergence.

For what it is worth, the benign causes listed in ChiR1 look real to me, and a caching proxy or a post-reorg stale response would be indistinguishable from an attack at this layer, which argues against a malicious charge specifically. But the parity claim cuts both ways: bitcoin-sv has no requested-versus-served split to police because it carries a single block identity, so its DoS policy is not obviously the right precedent for an HTTP GET contract it does not have.

Left as-is in this PR either way, since both proposed changes alter behaviour beyond the reported issue. Happy to implement whichever you land on as a follow-up, including the `NewNetworkError` variant to engage failover at `Server.go:2285` without a reputation charge.

### Not taken here, with reasons

**`if b != nil` should be `len(b) > 0`.** Correct, and I had to touch those three assertions anyway, so I considered folding it in. I left it because the predicate change alters the message for every empty-body non-2xx response repo-wide, dropping the `with body` clause entirely rather than just re-rendering it, and three existing tests assert the current shape deliberately. That is a wider behavioural change than the escaping and unrelated to the security finding. Trivial to do as its own commit if you would rather have it now.

**The long-catchup trade is now in the description.** A catchup legitimately exceeding 10 minutes loses its suppression marker and the next duplicate re-enqueues. Bounded by `CatchupChBufferSize`, and the duplicate hits `ErrCatchupInProgress` and returns without consuming an attempt. You are right that framing it as pure hardening was misleading.

**`catchupAlternatives` keeps touch-on-hit, so an alternatives list can now outlive its marker.** Agreed it converges rather than leaks, and agreed the pair should end up symmetric. That belongs with bitcoin-sv/teranode#4746, which is already about that cache's growth, rather than here.

**Both test gaps are fair.** Nothing pins that the `catchupCh` key and the `processBlockNotify` key agree, which is the invariant the fix actually protects, and `NewCatchupTestSuite` builds `processBlockNotify` with no options at all, so it now diverges from production in exactly the dimension this PR hardens. The second one worries me more than the first, because a future marker-lifecycle test written against that suite would pass for the wrong reason. I would rather fix the suite and add the key-agreement test together in the follow-up that takes the caller-side identity work, so the test lands next to the code it constrains.

### Verification at f8bd8dc

```
gofmt -l <changed files>                                          -> clean
go vet ./util/                                                    -> clean
gci diff --skip-generated -s standard -s default <changed files>  -> clean

go test -count=1 ./util/
  ok  github.com/bsv-blockchain/teranode/util                              7.690s
go test -count=1 -timeout 1800s -tags testtxmetacache ./services/blockvalidation/...
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation          274.861s
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation/catchup   13.338s
```

`blockvalidation` re-run because the error-message format change is repo-wide, not because that package changed. I checked for other assertions on the old `with body [` shape and there are none outside `util`.

`-race` and the full repository suite still not run locally.

PR #1454, opened 2026-08-05
fix(blockvalidation): back off on HTTP 429 and prefer archival peers during p2p IBD catchup

### LINKED ISSUE #1174: p2p catchup wedges during IBD: no backoff on HTTP 429, no archival-aware peer selection
## Summary

During IBD over the p2p (teranode-native) path, block-validation catchup fetches subtrees / subtree_data / blocks over HTTP from peers' asset endpoints. When a peer returns HTTP 429 (the asset heavy-endpoint limiter, `asset_httpHeavyRateLimit`, default **10 req/s**), catchup does not back off — it retries the same large concurrent fan-out immediately, re-trips the limit, and never makes progress. With no preference for peers that actually hold the data (pruned peers return 404), p2p IBD deterministically wedges.

## Impact

- p2p-only IBD on teratestnet stalls with no height progress for 30+ min at a fixed height.
- Reproduced on **v0.16.0-beta-1** and **main @`40ba9af`** — both stall at the *identical* height (5837) under the same network conditions.
- legacy sync path is unaffected (full blocks over BSV wire protocol, doesn't touch the HTTP subtree endpoints).

## Evidence

Controlled benchmark: fresh sync to height 20000, teratestnet, `COMPOSE_PROFILES=p2p`. Catchup rotated across every available peer, each failing:

| peer | role | result |
|------|------|--------|
| eks-ttn-eu-1 | archival | `429 rate limit exceeded` on `/subtree`, `/subtree_data` |
| ovh-ttn-eu-2 | pruned | `404 NOT_FOUND` on `/subtree` |
| stuartfreemansoftworks | archival | `429` on `/subtree_data` and `/blocks?n=100` batch |

Dozens of subtree fetches fire within the same second (burst), instantly exceeding the peer's heavy limiter (default 10 req/s, burst 10). Catchup then logs retry and re-issues the same burst → same 429 → zero progress until the node gives up.

```
WARN  get_blocks.go:651 [catchup:fetchAndStoreSubtreeAndSubtreeData] ... failed to fetch subtree ...
      returned status code [429] with body [{"message":"rate limit exceeded"}]
WARN  get_blocks.go:651 ... returned status code [404] with body [{"message":"NOT_FOUND (3): not found"}]
WARN  Server.go:1682   [catchup] Local service error during catchup ... (attempt 1/5),
      clearing markers to allow retry ... /api/v1/blocks/...?n=100 returned status code [429]
ERROR catchup.go:1036  refusing RUN: chain tip height 5837 is below highest checkpoint 18000 for teratestnet
```

## Root cause (client side)

- `services/blockvalidation/get_blocks.go` (`fetchSubtreeFromPeer` / `fetchSubtreeDataFromPeer` / `fetchBlocksBatch`): 429 is treated as a generic fetch error — no `Retry-After` handling, no exponential backoff, and the retry reissues the full concurrent fan-out.
- Peer selection rotates onto peers that just returned 404 (don't have the data) and re-hammers peers that returned 429, instead of preferring peers that successfully served data.

## Proposed fix

1. **Backoff on 429**: honor `Retry-After` if present, else exponential backoff with jitter; do not re-burst. Reduce per-peer heavy-fetch concurrency on repeated 429.
2. **Archival-aware peer selection**: deprioritize/skip peers returning 404 for historical data; prefer peers that have served the requested range; spread heavy fetches rather than fan out to one peer.
3. **Bound concurrent heavy fetches per peer** so a single block's subtree set can't exceed a peer's burst in one tick.

## Notes

Server-side, operators serving IBD can raise `asset_httpHeavyRateLimit` (default 10) as a mitigation — but that's a workaround. A syncing node should not wedge against a rate-limited or partially-pruned peer set; it should back off and rebalance.

Environment: teratestnet, v0.16.0-beta-1 (and main @`40ba9af`), single-node quickstart, `p2p` profile, fresh sync to 20000.

oskarszoon:
Fixes #1174 — p2p-only IBD deterministically wedges.

During p2p catchup, block-validation fetches subtree / subtree_data / block data over HTTP from peers' asset endpoints. Two failure modes stalled IBD at a fixed height, neither addressed upstream:

1. **No backoff on HTTP 429.** When a peer returned 429 (the asset heavy-route limiter, default 10/s), catchup retried the same large concurrent fan-out immediately, re-tripping the limit with no progress.
2. **No archival-aware peer selection.** Catchup rotated onto pruned peers that 404 on archival subtree data and re-hammered rate-limited ones.

## What changed

1. **`util/http`** — `buildHTTPError` maps 429 to the retryable `ErrServiceUnavailable` class; a shared `retryHTTP` driver adds equal jitter and honors a server `Retry-After` clamped to a 30s ceiling (above the 5s backoff cap), with an overflow-safe pre-jitter clamp so a hostile hint can't wrap negative. A cumulative 60s budget bounds the time spent *sleeping between attempts* — it does **not** bound total wall-clock, so the catchup fetch path additionally carries an explicit deadline (see 3). Peer URLs are redacted to `scheme://host` before they reach any error text. Asset advertises `Retry-After: 1` on the 429 path.
2. **`model` / `settings`** — the catchup transport path decodes blocks through a bounded reader (`NewBlockFromReaderWithDeclaredSizeLimit`, whose sole non-test caller is that path — it hardens catchup, not every decode caller); the coinbase parse shared by *every* block decoder (`readTransactionAllocationSafe`) now takes an explicit allocation budget (8 MiB) instead of the 1 GiB go-bt arena, so a wire-controlled length prefix can't pre-allocate gigabytes before the transport cap is consulted. The new `blockvalidation_max_incoming_block_bytes` transport cap is an **aggregate per-response** budget (shared across a multi-block batch, not `n × cap`) and is validated at startup; the new `blockvalidation_per_peer_fetch_rate` pacing knob is **not** startup-validated — a non-positive value is a documented "disable pacing" sentinel, so there is no invalid value to reject (both knobs are now in `settings.conf`). A response truncated mid-block (peer restart, TCP RST, proxy close) is classified as an external peer fault rather than a block-validation failure, so an honest peer cut mid-stream is not marked malicious.
3. **`blockvalidation` / `p2p`** — client-side per-peer token-bucket pacing so the fan-out can't burst into a peer's limiter; the catchup fetch path is now bounded by a wall-clock deadline (`CatchupIterationTimeout`) so a slow or silent peer can't hang an iteration; pruned peers deprioritized to a last-resort failover tail (not excluded — they still hold recent data for warm followers), computed from a once-per-block peer snapshot; bounded per-subtree failover breadth decoupled from the retry count; peer `Storage` class carried through the catchup API; the sync peer is switched only when a failure names the current one.

Builds on upstream #1371 (cache-bypass retry + catchup failure attribution) — reused, not duplicated.

## Supersedes #1176

Recut fresh against current `main` to drop that PR's stale multi-round merge history after #1371 landed the overlapping failure-attribution rework. The failure-attribution core is ceded to #1371; this PR is the remaining, self-contained #1174 fix.

## Testing

`go build ./...`, `go vet`, `go test -race` (util, settings, model, services/asset/httpimpl, services/p2p, services/blockvalidation), `staticcheck`, and `golangci-lint` (vs main) — all clean.


ordishs (changes requested):
### 1. Peer-supplied URL text can still forge a "local" classification

`util/http.go:206, 208, 212, 215, 217, 220, 831, 834, 836, 849`
`services/blockvalidation/get_blocks.go:800, 1042`
`services/blockvalidation/peer_cache_bypass.go:108, 110, 138`
`services/blockvalidation/catchup.go:393`

Scope note: the error constructors that interpolate a raw `baseURL` are pre-existing (`peer_cache_bypass.go` last changed in #1371; `get_blocks.go:800` appears only as a context line in this diff, `:1042` not at all). What this PR contributes is the new `RedactPeerURL` mechanism applied incompletely, plus five new or widened `IsLocalError` gates that turn a cosmetic leak into a trust decision. It is raised here because this PR is what makes the classification load-bearing.

- `RedactPeerURL` is new in this PR and is applied only in `buildHTTPError` (`util/http.go:505`) and at three call sites in `get_blocks.go` (`:92`, `:1031`, `:1100`). Every other constructor listed above still interpolates the raw peer-gossiped `DataHubURL`.
- Redacting the outer message does not help: `(*Error).Error()` renders the whole wrapped chain, and `(*Error).Is` falls back to `strings.Contains(e.Error(), target.Error())` for non-`*Error` targets — which is exactly what `context.Canceled` / `context.DeadlineExceeded` are. So `IsContextError` → `IsLocalError` is decided by substring match on peer-controlled bytes.
- `validateDataHubURL` (`services/p2p/server_helpers.go:381`) and `util.ValidateURL` check scheme and SSRF host only. `url.Parse("https://evil.example/dh/context canceled")` succeeds and `http.NewRequest` escapes the space on the wire, so it is a fully working base URL. Reproduced against the branch with that baseURL interpolated into the constructor shape these sites use — `ProcessingError -> ServiceError("...from %s", baseURL) -> ExternalError`:

```
rendered: PROCESSING (4): [catchup] all peers failed -> SERVICE_ERROR (59):
  failed to fetch subtree from https://evil.example/dh/context canceled -> EXTERNAL (8): peer returned garbage
Is(err, ErrExternal) = true
IsContextError(err)  = true
IsLocalError(err)    = true
```

The same chain with a benign `https://honest.example/dh` yields `IsLocalError == false`. So an error explicitly classified `ErrExternal` is treated as local purely because of peer-controlled URL text.
- This PR grows the number of production gates that verdict controls to six, five of them newly added or widened here (only `peer_selection.go:158` predates it): `catchup.go:393` is now the **first** switch case in `releaseCatchupLock` (`isPeerError = false`, no `CatchupFailure`, no rotation signal); `get_blocks.go:946`/`:987` return instead of trying alternatives; `peer_selection.go:158` and `Server.go:2035` `break` the alternatives loop; `Server.go:2191` and `catchup_get_block_headers.go:292` skip `reportCatchupFailure`; `Server.go:1955` takes the silent local-retry branch with no `ReportPeerFailure`.
- Net effect: one peer with a crafted URL serves garbage repeatedly, accrues no reputation penalty, and blocks failover to honest peers — the #1174 wedge this PR is meant to close. `peer_cache_bypass.go:138` is the worst of these because it embeds every attempted peer's raw baseURL into the terminal all-peers-failed error for the whole block.

Actions:
- Thread an already-redacted display string into `readBodyWithCtx` and `doRequestReaderWithRetryAfter` and use it in every constructor there, including the bare `errors.NewServiceError("failed to do http request", err)` wrap of the `*url.Error`. Same at `get_blocks.go:800`, `:1042` and `peer_cache_bypass.go:108`, `:110`, `:138`.
- Make `IsContextError`/`IsLocalError` match on error code (`ERR_CONTEXT_CANCELED` / `ERR_CONTEXT`) and on `errors.Is` against the stdlib sentinels via native links; drop the substring fallback for those targets. Message text should never decide a trust question.
- Add a regression test: a `DataHubURL` containing `context canceled` routed through `fetchSubtreeFromPeer` must yield `IsLocalError == false`.

### 2. 8 MiB coinbase cap applies to every block decode, not just peer transport

`model/Block.go:426, 465-470`

- `readBlockFromReaderWithLimits` calls `readTransactionAllocationSafe(buf, maxCoinbaseTxBytes)` unconditionally, and `readBlockFromReader` (`:365`) reaches it with no limits, so `NewBlockFromBytes` and `NewBlockFromReader` inherit the cap. That covers the consensus entry points (`blockvalidation/Server.go:1319`, `:1393`), `subtreevalidation/check_block_subtrees.go:360`, `blockpersister/persist_block.go:53`, `blockassembly/Client.go:770`, and six `blockchain/Client.go` sites that read blocks back out of our own store.
- The justifying comment bounds the coinbase by scriptSig size (2-100 bytes) and input count (1). Neither determines coinbase size — output count and output script size are what matter, and BSV consensus bounds neither.
- Reproduced against the branch. A block whose coinbase has one null-outpoint input and 300,000 standard 25-byte P2PKH outputs (10,200,059 serialized bytes) is rejected by `model.NewBlockFromBytes`:

```
small coinbase:    395 bytes -> err=<nil>
large coinbase: 10200059 bytes -> err=BLOCK_INVALID (11): error reading coinbase tx
  -> BLOCK_INVALID (11): expected 8 bytes, got 5 -> UNKNOWN (0): EOF
errors.Is(err, ErrBlockInvalid) = true
```
- Consequences: that block is consensus-valid and accepted by SVNode, so a miner can fork Teranode nodes off the longest chain for the cost of ~10 MB of block space. Independently of any attacker, once such a block is persisted locally we can no longer read it back to serve or reprocess it. The failure is reported as `ErrBlockInvalid` — a consensus verdict, not a resource error — and there is no setting to raise it.

Actions:
- Keep the allocation budget on the untrusted transport path only: thread it in from `decodeBoundedBlock` alongside `maxDeclaredBytes`/`enforceDeclaredSize`, and pass an unbounded budget from `readBlockFromReader`/`NewBlockFromBytes`.
- If a transport-side cap stays, make it a setting next to `blockvalidation_max_incoming_block_bytes`, derive the default from `Policy.ExcessiveBlockSize` rather than a magic constant, and return a distinct non-`ErrBlockInvalid` class.

### 3. Subtree and subtree_data fetches have no wall-clock bound

`services/blockvalidation/get_blocks.go:714, 822-830, 1008, 1081`

- `withCatchupFetchTimeout` has exactly two call sites: `fetchBlocksBatch` (`:1271`) and `fetchSingleBlock` (`:1351`). `fetchSubtreeFromPeer` (`:1008`) and `fetchSubtreeDataFromPeer` (`:1081`) did not get it.
- `fetchAndStoreSubtreeData` derives its download context at `:714` with `context.WithCancel(shutdownCtx)` — cancel-only, no deadline. Nothing above it supplies one either: `fetchBlocksConcurrently` gets the plain catchup ctx (`catchup.go:1036`), and the only `context.WithTimeout` in the package is the header-fetch iteration at `catchup_get_block_headers.go:223`. So `doRequestReaderWithRetryAfter` mints a fresh `httpStreamingTimeout` (300s) per attempt with no ceiling.
- The comment at `:826-829` claims each alternative fetch is "bounded (the per-attempt HTTP timeout times the retry helper's attempt count)". That product is 6 × 300s = 30 minutes.
- Concrete abuse: a peer accepts the connection, holds the request ~299s, returns `503` with no `Retry-After`. `buildHTTPError` maps 503 to `ErrServiceUnavailable`, `retryHTTP` retries after only the jittered exponential backoff (250ms..5s, so the 60s `maxBackoffTotal` never trips). Six attempts ≈ 30 minutes for one subtree from one peer, then up to `maxSubtreeFailoverPeers = 10` alternatives, plus a possible second cache-bypass round in `tryPeerForSubtree`. Since the block path *is* bounded at 30s, an attacker just concentrates on subtree assignments — `DistributeSubtreesAcrossPeers` round-robins, so a hostile peer gets a share of every block's subtrees.

Actions:
- Wrap both subtree fetches in `withCatchupFetchTimeout`, or wire up `blockvalidation_subtree_fetch_timeout` (present at `settings.conf:422` but bound to no Settings field and used nowhere), so all retry attempts plus the streaming read share one wall clock.
- Bound `retryHTTP` by elapsed time, not only by cumulative sleep — attempt duration is what the peer controls.
- Correct the comment at `:826-829`.

### 4. Unattributed catchup failures now rotate nothing; the call-site contract comment is stale and `registryRecordSyncFailure` is dead

`services/p2p/sync_coordinator.go:309, 655-682`
`services/blockvalidation/Server.go:1902-1928`

- `HandleCatchupFailureForPeer` returns early when `peerID == ""` and when the peer is not the current sync peer, and the unconditional `registryRecordSyncFailure(failedPeer)` charge was dropped. Restricting rotation to the named peer is a defensible fix for #1368, but the empty-peer case now records nothing anywhere — `reportCatchupFailureWithKind` is also a no-op for an empty peerID.
- That case is reachable: `blockvalidation/Server.go:1793-1797` builds a retry `processBlockFound{baseURL: SourceTypeRetry, peerID: ""}` with the comment "Clear peer ID so any peer can be used", and `catchup_get_block_headers.go:57-59` already anticipates empty peerIDs by falling back to `identifier = baseURL`. When such a cycle fails after trying every peer, we log "Ignoring catchup failure without peer attribution", record no failure, and do not re-select. Recovery then waits on the 30s `periodicEvaluation`, which only rotates on reputation < 20 or the 5-minute `SyncPeerNoProgressTimeout` — a peer with 100 successes and 1 failure trips neither.
- `registryRecordSyncFailure` (`sync_coordinator.go:309`) now has zero callers anywhere in the tree, including tests. `.golangci.json` disables the `unused` linter, so nothing flags it, and the registry's sync-failure metric is no longer incremented by anything.
- The comment at `blockvalidation/Server.go:1902-1928` still names `syncCoordinator.HandleCatchupFailure` (removed by this PR), still states rotation "has to fire on every all-peers-failed cycle", and still warns that the fallback "costs the full five minutes per stuck cycle" — describing a guarantee the code no longer provides.

Actions:
- Keep the compare-and-clear for the named peer, but add a fallback: when `peerID` is empty or names a non-current peer, still check whether the current sync peer has made progress and clear it if not — at minimum call `triggerSyncLocked()` so re-selection runs.
- Delete `registryRecordSyncFailure` and document that catchup failures are charged solely by `recordCatchupPeerFailure`, or re-wire it.
- Rewrite the `Server.go:1902-1928` comment to name `HandleCatchupFailureForPeer` and describe the actual behaviour.

oskarszoon:
All four addressed in f009704c5. `go test -race` green across errors, util, model, settings, services/p2p, services/blockvalidation; build + vet clean.

**1 — URL-text forgery of a LOCAL classification.** Root-caused to `IsContextError`'s `strings.Contains` fallback on the rendered chain. Dropped it. Genuine cancellations are now detected by `ERR_CONTEXT_CANCELED`/`ERR_CONTEXT` code, by native stdlib-sentinel identity (`New` now preserves `context.Canceled`/`DeadlineExceeded` in the Unwrap chain by identity rather than flattening to a message), and by gRPC status code. `RedactPeerURL` completed across `readBodyWithCtx`, `doRequestReaderWithRetryAfter` (unwrapping the `*url.Error` so its embedded raw URL doesn't leak), `get_blocks.go:800/1042`, and `peer_cache_bypass.go`. Regression: a `DataHubURL` containing the literal `context canceled` is not `IsLocalError` (`errors/error_utils_test.go`).

**2 — coinbase cap on every decode (consensus).** Agreed, this was the real bug. `readTransactionAllocationSafe` treats `maxBytes<=0` as unbounded; the budget is threaded only from `decodeBoundedBlock` (the transport envelope), and `readBlockFromReader`/`NewBlockFromReader`/`NewBlockFromBytes` pass unbounded, so store read-back and the consensus entry points accept any consensus-valid coinbase. Regression decodes a 300k-output (~10 MB) coinbase block via `NewBlockFromBytes` (`model/Block_test.go`).

**3 — subtree fetches unbounded.** Wired the dead `blockvalidation_subtree_fetch_timeout` (default 120s) to a settings field. `fetchSubtreeFromPeer` bounds inline; `subtree_data` is bounded by its download context (`fetchAndStoreSubtreeData`, now `WithTimeout`) rather than a defer-cancel inside the fetch — that function returns a streaming body the caller reads after return, so cancelling on return truncates it. Corrected the `maxSubtreeFailoverPeers` comment.

**4 — unattributed failures rotate nothing.** `HandleCatchupFailureForPeer` now, for the empty/non-current-peer cases, checks whether the current sync peer has stalled and clears it if so, otherwise re-runs selection — instead of dropping the failure and waiting on the 5-minute no-progress timer. Removed the dead `registryRecordSyncFailure` and rewrote the stale `Server.go` rotation comment to name the actual handler and behaviour.

ordishs (changes requested):
Re-reviewed at `0ece851d3`. Findings 2 and 3 are genuinely closed — the coinbase consensus fix is exactly right, confirmed by decoding the 10,200,059-byte block that failed last round on both `NewBlockFromBytes` and `NewBlockFromReader`, with the transport path still rejecting it under a tighter budget. Findings 1 and 4 are incomplete, and the two fixes that changed shared machinery each introduced a new defect.

Suites at this head are green with `-race` (`errors`, `model`, `util`, `settings`, `blockvalidation` 302.7s, `catchup` 70.6s, `p2p` 14.2s) — no regressions in the existing tests.

Items marked **[measured]** were confirmed by executing code, not by reading it.

### 1. The new subtree wall clock blames peers for local store stalls — P1 [measured]

`services/blockvalidation/get_blocks.go:720, 809-822`

The deadline context now spans the local `subtreeStore.Set`, and `classifyPeerFetchCtxErr` maps `ctx.Err() == DeadlineExceeded` to a peer-attributed `NewNetworkTimeoutError`. That branch was unreachable before: a `WithCancel` context can only yield `Canceled`.

With an expired deadline and a purely local `StorageError` from `Set`:

```
classified as: NETWORK_TIMEOUT (113): [catchup:fetchAndStoreSubtreeData] subtree data timed out
IsLocalError          = false
Is(ErrNetworkTimeout) = true
```

The cancel-only control classifies the identical failure `CONTEXT_CANCELED`, `IsLocalError = true`. Because the result is not local, `recordCatchupPeerFailure` charges the peer and the caller fans out to up to `maxSubtreeFailoverPeers = 10` alternatives — each re-triggering the same slow local write and taking the same misattributed penalty. One storage hiccup can damage an entire honest peer set.

- Close `dlCancel` before `Set` and pass `shutdownCtx` to the store write, or route `Set` failures through the `StorageError` path unconditionally.

### 2. Transport coinbase budget is effectively 8 GiB — P1 [measured]

`services/blockvalidation/get_blocks.go:1219`

Removing the blanket 8 MiB cap was correct; replacing it with `limits.maxTransportBytes` (default 8 GiB) is not a bound. Two compounding reasons:

- `maxTransportBytes` is the whole aggregate response envelope, so one block's coinbase may claim the entire batch budget.
- `readTransactionAllocationSafe` tries to tighten to the reader's *remaining* allowance via `if outer, ok := r.(*io.LimitedReader)`. That assertion never fires — the caller passes `bufio.NewReaderSize(limited, ...)` (`:1325`, `:1402`).

Measured amplification decoding a coinbase of minimal 9-byte outputs:

```
wire bytes for 1000000 minimal outputs: 9000059 (9.00 bytes/output)
heap allocated decoding it:            183583096 bytes
AMPLIFICATION: 20.4x wire size  =>  an 8 GiB envelope implies ~163 GiB heap
```

The pre-fix bound was ~163 MiB. The consensus fix was necessary, but it moved the DoS ceiling by three orders of magnitude.

- Derive the coinbase budget from `limits.maxDeclaredBytes` (excessiveblocksize) instead — a coinbase cannot exceed its block, so this cannot reject a consensus-valid block while restoring a tight bound.
- Separately, pass `limited` (or unwrap the `bufio.Reader`) so the remaining-budget tightening actually works.

### 3. Pacing-wait deadline now aborts failover entirely — P1

`services/blockvalidation/get_blocks.go:76-97, 952`

`awaitPeerFetchSlot` returns a `ContextCanceledError` when `rate.Wait` reports "would exceed context deadline". The new 120s wall clock makes that reachable on the subtree path for the first time, and it is `IsLocalError == true`, which `:952` treats as unfixable by other peers:

```go
if errors.IsLocalError(err) {
    return "", errors.NewServiceError("... Local error fetching subtree %s (not retrying with other peers)", ...)
}
```

A peer paced hard by the local token bucket (`blockvalidation_per_peer_fetch_rate`, default 8/s) under a busy fan-out exhausts the budget in the limiter queue, and the subtree fetch aborts for the whole fleet rather than failing over to an unpaced peer — a self-inflicted stall of the same class as #1174.

- Classify a pacing-wait deadline breach as retryable-with-another-peer, or check limiter headroom before committing to the peer.

### 4. The invariant justifying the retained substring classifier does not hold — P1 [measured]

`errors/error_utils.go:10-17`, `services/blockvalidation/catchup_get_block_headers.go:67, 81, 202, 215, 271`

`RedactPeerURL` itself is sound — 12 hostile inputs tested (path, query, fragment, userinfo, host, percent-encoded, IPv6, unparseable); none carries a sentinel through, and it fails closed to `[redacted url]`. `buildHTTPError` correctly drains the body without embedding it.

But the comment justifies keeping the substring classifier by asserting "no peer bytes can carry a space-delimited sentinel like 'context canceled' into the rendered chain", and `catchup_get_block_headers.go` was never touched. It still interpolates a raw peer-gossiped `baseURL` at `:271` and `:202`, and `identifier` — which falls back to `baseURL` when `peerID` is empty (`:57-59`) — at `:67`, `:81`, `:215`.

With baseURL `https://evil.example/dh/context canceled`:

```
rendered: PROCESSING (4): [catchup] iteration failed -> NETWORK_TIMEOUT (113):
  peer https://evil.example/dh/context canceled timed out after 3s during iteration 2
IsContextError = true
IsLocalError   = true          (benign-URL control: false)

circuit-breaker  IsLocalError = true
marked-malicious IsLocalError = true
```

A peer already flagged **malicious** classifies as local and escapes the reputation charge.

- Redact at those five sites and add the regression test, or stop letting message text decide a trust question: match on code and on `errors.Is` against the stdlib sentinels via native links, keeping the substring fallback for the gRPC shutdown gates only, behind a helper the reputation path does not use.

### 5. Rotation fallback is inert whenever a sync peer is active — P1

`services/p2p/sync_coordinator.go:646-691`, comment mirrored at `services/blockvalidation/Server.go:1911-1916`

For a non-stalled current peer the fallback calls `triggerSyncLocked()`, which reaches:

```go
func (sc *SyncCoordinator) selectAndActivateNewPeer(localHeight uint32, oldPeer string) error {
	if oldPeer != "" {
		sc.logger.Debugf("[SyncCoordinator] Sync peer %s already active; skipping new activation", oldPeer)
		return nil
	}
```

So whenever a sync peer is active and inside the no-progress window, the fallback does nothing. Both comments state a better peer "can preempt immediately instead of waiting on the periodic evaluator / no-progress timer"; that cannot happen. A `SourceTypeRetry` all-peers-failed cycle (`Server.go:1793-1796`, `peerID: ""`) against a slow-but-not-stalled sync peer still records nothing and changes nothing.

Genuinely fixed here: dead `registryRecordSyncFailure` deleted, the stale `HandleCatchupFailure` comment corrected, and lock discipline is right (`syncPeerNoProgressTimedOut` takes `sc.mu`; `clearNoProgressSyncPeer` documents that `decisionMu` is held and the new call site holds it — no deadlock).

- Either implement real preemption, or downgrade both comments to what the code does: an early no-progress check.

### Test gaps blocking confidence

- No test calls `HandleCatchupFailureForPeer("")` — that is the whole of fix 4.
- `withCatchupSubtreeFetchTimeout` has no test at all: not the bound, not the non-positive fallback, not the interaction with the streaming read.
- `decodeBoundedBlock`'s test helper (`get_blocks_test.go:100-103`) layers readers differently from production, so the new `maxCoinbaseBytes` parameter is never exercised against the `bufio` layering that defeats the tightening.
- `sync_coordinator_test.go:228-237` — `"ignores failure for non-current peer"` no longer describes the behaviour and passes only because zeroed timestamps make the progress check return false.

`TestNewBlockFromBytes_LargeCoinbaseDecodes` is a good requirement-derived regression test — it asserts the output count, not merely "no error".

### Docs

- `blockvalidation_subtree_fetch_timeout` is missing from `docs/references/settings/services/blockvalidation_settings.md` (73 other blockvalidation keys are documented).
- `MaxIncomingBlockBytes` longdesc still claims "a legitimate per-block envelope is single-digit MB regardless of the Bitcoin block size" (`settings/blockvalidation_settings.go:67`, mirrored at `settings.conf:424`). Finding 2 established a consensus-valid coinbase can be ~10 MB alone; an operator tuning down on that guidance would fail over from honest peers.

ordishs:
No new commits since the last review — head is still `0ece851d3`, so the five outstanding P1s stand as written. Two notes on the evidence rather than the code.

### The stated proof for finding 1 was withdrawn in the same series

The summary comment above describes `f009704c5`, which dropped the `strings.Contains` fallback and added a regression test asserting that a `DataHubURL` containing the literal `context canceled` is not `IsLocalError`. The follow-up commit `0ece851d3` reverted the classifier change **and deleted that test**:

```
0ece851d3  errors/error_utils.go      | 31 +++++-------------------
           errors/error_utils_test.go | 23 -----------------------
           errors/errors.go           |  8 --------
```

Its commit message substitutes "TestRedactPeerURL guards it". `TestRedactPeerURL` exercises the helper in isolation, which is sound as far as it goes — I tested 12 hostile inputs against `RedactPeerURL` and none carries a sentinel through; it fails closed to `[redacted url]`. But that is not the invariant the retained classifier depends on. The end-to-end claim — "no peer bytes can carry a space-delimited sentinel into the rendered chain" — has no test at head, and it does not hold: `catchup_get_block_headers.go:67, 81, 202, 215, 271` still interpolate a raw peer-gossiped `baseURL`, and I measured the result classifying `IsLocalError == true`, including for a peer explicitly marked malicious.

So the PR description currently asserts a guarantee that neither the code nor any test provides. Worth either restoring the deleted regression test in a form that covers the header path, or amending the description.

### The third-party-error concern raised inline is not currently exploitable

Following up the note on `errors/error_utils.go:242` about `subtreepkg.NewSubtreeDataFromReader` / `subtreeData.Serialize()` possibly interpolating peer bytes: I checked go-subtree v1.4.2. `subtree_data.go` interpolates only wrapped errors (`%w`) and integer indices (`%d`); no `%s`/`%v` on peer-supplied strings or byte slices. The `recover()` paths in `subtree.go:777, 810, 1015` format the panic value with `%v`, but that is runtime-generated text (bounds/nil messages), not attacker-chosen. Hashes render via `.String()` as hex, which cannot contain a space.

So there is no current injection path there — it is a valid hardening suggestion rather than a live gap, and it reinforces the same structural point: the invariant is enforced pointwise at each call site rather than by construction, so it needs a test that asserts it across the catchup error paths, not just at the helper.

icellan:
Reviewed the full diff (33 files). The direction is right, but the hardening leaks peer-path assumptions into shared code and misclassifies several failure modes. P0/P1 below; nits omitted.

## P0

**`model/Block.go:426` — the 8 MiB coinbase cap landed in the shared decoder, not the catchup path**
`readTransactionAllocationSafe(buf, maxCoinbaseTxBytes)` sits inside `readBlockFromReaderWithLimits`, so every block deserialization in the node is capped: ValidateBlock/ProcessBlock gRPC (`services/blockvalidation/Server.go:1319`, `:1393`), blockpersister (`persist_block.go:53`), subtreevalidation (`check_block_subtrees.go:360`), and `blockchain.Client.GetBlock` reading blocks back out of the local store (`Client.go:407`, `:1100`). `maxCoinbaseTxBytes` is a compile-time constant, unrelated to `excessiveblocksize` and not gated by `enforceDeclaredSize`. BSV caps coinbase scriptSig at 100 bytes but places no limit on coinbase output count — a consensus-valid coinbase over 8 MiB (~930k outputs) makes this node reject the block, mark the honest peer malicious, and fork. Worse: an already-stored block over the cap becomes permanently unreadable from the store. Gate the cap on the peer-supplied path only, or tie it to `excessiveblocksize`.

**`services/blockvalidation/get_blocks.go:1042` — incomplete `RedactPeerURL` rollout leaves the forgery channel open**
Line 1031 was changed to `util.RedactPeerURL(url)`; the empty-subtree `NewNotFoundError` eleven lines later still interpolates the raw peer URL (same at `catchup.go:393`). A peer gossips `DataHubURL = "http://evil.example/context canceled"` (`url.Parse` accepts the space, `ValidateURL` only checks scheme+SSRF host) and serves a zero-byte 200. `(*Error).Is` falls back to `strings.Contains` for non-`*Error` targets (`errors/errors.go:177`), so the error classifies as `IsLocalError`: `recordCatchupPeerFailure` skips the reputation charge (`:905`/`:916`), `fetchAndStoreSubtreeAndSubtreeData:946` refuses to try another peer, and `processCatchupChItem:1955` only warns. One hostile peer wedges catchup past that block with its reputation intact — the #1174 wedge, now peer-triggerable on demand. Redact everywhere, and stop letting substring matching decide locality.

**`services/blockvalidation/get_blocks.go:813` — local blob-store timeout gets blamed on the peer**
`classifyDownloadErr` runs on the `subtreeStore.Set` failure and its third branch (`errors.Is(err, context.DeadlineExceeded)`) returns a non-local `NetworkTimeoutError`, bypassing the `NewStorageError` on line 816. When S3/the subtree store times out, the honest serving peer is charged a catchup failure, catchup churns across the fleet, and the loud `ErrStorageError` gate in `Server.go:1943` never fires. The operator sees peer churn instead of "storage is down". Don't run peer classification over store errors.

**`services/blockvalidation/get_blocks.go:961` — `sync.Once` peer snapshot caches its error**
`catchupPeerSnapshot.get()` memoises `(peers, primaryPruned, err)`. One transient `GetPeersForCatchup` blip (p2p restart, gRPC deadline) sets `snapshotErr` for every subsequent caller, so `alternativePeers` stays nil for every subtree of that block and `fetchSubtreeDataForBlock:480` loses parallel distribution. Previously each subtree re-issued `GetPeersAtMaxHeight` and recovered. Result: "all 1 peer attempts failed", primary peer reputation-charged for our own RPC blip, block retried to the cap. Cache the success, retry the error.

## P1

- **`services/p2p/sync_coordinator.go:673`** — `HandleCatchupFailureForPeer` now rotates only on an exact `currentSyncPeer` match and no longer calls `registryRecordSyncFailure` (now dead code). blockvalidation reports the peer that *originated* the notification (`Server.go:1929`, `:1972`), which often isn't the sync peer. Then: no rotation *and* no failure metric for anyone. `Server.go:1908-1921` documents this outcome as unacceptable; fallback is a 5-minute `SyncPeerNoProgressTimeout` stall per stuck cycle.
- **`services/blockvalidation/get_blocks.go:1271`** — `withCatchupFetchTimeout` caps the whole batch fetch (all attempts + body read) at `CatchupIterationTimeout` (30s), down from `http_timeout` (60s), and the expiry surfaces as a peer-charging `NetworkTimeoutError`. The retry config's own budgets (`maxAttempts 6`, `maxRetryAfter 30s`, `maxBackoffTotal 60s`) are unreachable under a 30s ceiling — two honoured `Retry-After: 5` responses plus a slow body read blows it, and the peer gets degraded for the 429 backoff this PR added working as designed.
- **`services/blockvalidation/peer_selection.go:158`** — the new break-on-`IsLocalError` justifies itself with "no deadline is set here", which this same PR invalidates: `fetchBlocksBatch`/`fetchSingleBlock` derive a 30s ctx (`get_blocks.go:1144`, `:1268`, `:1352`) and `awaitPeerFetchSlot:84`/`:92` turns a deadline hit during pacing into a local `ContextCanceledError`. One slow alternative peer then aborts the loop over every remaining healthy peer. Twin at `Server.go:2035`.
- **`services/blockvalidation/get_blocks.go:59`** — the client limiter is a flat `rate.NewLimiter(8, 8)` per data-hub URL, ignoring the tier the peer actually grants signed requests (`tierPeer` = 10×5 = 50/s; `tierMiner` with `HTTPMinerRateLimit=0` is exempt). A 512-subtree block needs 1024 heavy requests: 128s of pure pacing instead of ~20s. That's a >6× IBD slowdown from the fix meant to unblock IBD.
- **`util/http.go:505`** — `buildHTTPError` dropped the response body and redacts the URL to scheme://host, so a failing peer now logs `returned status code [400] (137 body bytes, omitted)` with no path and no server reason. The two cases the code itself calls out — 401 auth failure and the ModSecurity WAF 400 on binary POST (`http.go:407-419`) — are now indistinguishable without a packet capture. Separately `maxErrorBodyBytes` went 2 KiB → 64 KiB, so every failed heavy fetch reads 64 KiB of attacker-chosen bytes it discards.
- **`services/blockvalidation/catchup_performance.go:354`** — `filterMaxHeightPeers` computes `maxHeight` *after* dropping no-DataHubURL/banned/low-reputation peers; `GetPeersAtMaxHeight` computed it over the full list first. During deep IBD the tip peers are exactly the ones just knocked below reputation 20, so the bar collapses to a far-behind peer, which then gets subtree assignments and 404s every request — burning failover attempts on a peer that never had the data.

oskarszoon:
Updated in 5e20ef48f. Merged `upstream/main` at a49c431d2 and resolved all six conflicts, preserving upstream sync, shutdown, and SSRF handling.

Fixes from the ticket and PR feedback:

- **Peer discovery and selection:** cache successful snapshots only, retry transient discovery failures, and require subtree peers to reach the requested block height. Archival preference and pruned-peer fallback remain intact.
- **Pacing and failover:** a saturated peer queue can fall back to another peer without a reputation penalty. Caller cancellation still stops the operation. Unattributed/non-current sync failures now use upstream's guarded preemption and stalled-peer eviction rules, without double-counting failures.
- **Timeout attribution:** local storage writes use the catchup parent context and retain local error classification. Downloads keep their own retry/streaming deadlines. Block responses now have a separate configurable two-minute budget, allowing time beyond retry backoff; subtree data retains the upstream ten-minute download budget.
- **Block receive bounds:** coinbase scanning respects the accepted block-size policy and actual remaining response bytes, including buffered bytes. Receive-limit exhaustion no longer carries an invalid-block classification that could mark an honest peer malicious. Trusted decoding still supports large valid coinbases; structurally invalid encodings remain rejected.
- **Error handling and documentation:** redact peer URLs before header-error classification, reduce discarded HTTP error bodies to 2 KiB, and document timeout scopes, transport versus heap limits, and pacing for known authenticated fleets. The default remains 8 requests/sec because signing alone does not establish the remote allowance.

Added regression coverage, updated assertions for upstream behavior, and fixed missing metric initialization in an isolated retry test.

Validation: production `go build ./...`, affected-package unit and race tests, `go vet`, `staticcheck`, and changed-line `golangci-lint` passed. `gosec` reports no findings on changed lines; existing code findings and dependency/toolchain findings from `govulncheck` remain. A full network IBD/load benchmark has not been run.

icellan:
Re-reviewed at `5e20ef48f` (36 files, +4090/−743, three-dot from `merge-base(origin/main, HEAD)`). `go build ./...` and `go vet` clean; `./model ./util ./settings ./errors ./services/p2p ./services/blockvalidation` pass, including `-race` on the new decode paths.

## Corrections to my earlier comment

Three things in it no longer hold, or were wrong when I wrote them:

- **The `RedactPeerURL` forgery P0 was overstated.** The path is closed by URL syntax regardless of redaction — a raw space can't survive `url.Parse`, and URL encoding escapes it. The redaction is belt-and-braces, not load-bearing. It still matters for a different reason, below.
- **The 30s catchup-ceiling P1 is gone.** The new wall clocks (`block_fetch_timeout` 2m, `subtree_fetch_timeout` 120s) are all *looser* than the pre-PR effective bound, which was `http_timeout`=30s applied to an unbounded `io.ReadAll`. Nothing got tighter. Withdrawn.
- **The flat 8/s limiter P1 was wrong on throughput.** `asset_httpHeavyRateLimit` already defaults to 10/s per unverified IP and `asset_peerAuthAllowlist` is empty by default, so the server was already the binding constraint. Not an IBD regression. What survives is a misclassification problem, not a rate problem — see the pacing cluster below.

I have **not** individually re-verified the remaining items from that comment (the shared-decoder coinbase cap, the `sync.Once` cached error, `classifyDownloadErr`, the sync-coordinator rotation gap, the `buildHTTPError` diagnostics loss, `filterMaxHeightPeers`). Absence from this pass isn't evidence they're fixed. Also worth knowing: two independent high-effort passes over this same head returned almost disjoint finding sets, so treat any single pass — including this one — as a floor rather than a sweep.

## Peer gets blamed for our own faults

This is the theme of the new findings, and it cuts against the PR's stated goal.

- **`services/blockvalidation/peer_selection.go:159`** — `shouldStopPeerFailover` deliberately returns `false` for an error carrying `peerFetchPacingKey`, so pacing exhaustion falls into the `else` and calls `reportCatchupFailureForError`, which skips only on `ErrBlockIncomplete`/already-reported — not on `IsLocalError`. The peer gets a real `RecordCatchupFailure` for a request it never saw. This contradicts the comment three lines above ("Neither should penalize a peer") and the `blockvalidation_per_peer_fetch_rate` longdesc ("permits failover to another peer **without a reputation penalty**"). The subtree path at `get_blocks.go:1022` gets this right — `recordCatchupPeerFailure` filters on `IsLocalError` — so the two paths now disagree on the same policy.
- **`services/blockvalidation/Server.go:2117`** — same defect in the cached-alternatives loop: the comment above the branch says pacing saturation does not stop failover, then the fall-through blames the peer anyway.
- **`services/blockvalidation/get_blocks.go:88`** — `awaitPeerFetchSlot` tags *every* non-cancel `rate.Limiter.Wait` failure with `peerFetchPacingKey`, but `x/time/rate` returns "would exceed context deadline" on a completely idle limiter whenever only a few ms remain on the caller's context. So a near-expired parent deadline is mislabeled as pacing saturation, `shouldStopPeerFailover` returns `false`, and catchup walks the whole alternative list charging each peer (above) until `ctx.Err()` trips. A shutdown burns reputation across every healthy peer. Compare `lim.Reserve().Delay()` against the remaining deadline before tagging.
- **`services/blockvalidation/catchup_get_block_headers.go:290`** — the new `if !errors.IsLocalError(err)` guard landed on `reportCatchupFailure` (line 296) but not on `circuitBreaker.RecordFailure()` six lines earlier. Repeated local cancellations (FSM transitions, a preempting catchup, shutdown) accumulate breaker failures against an honest peer; past `FailureThreshold` (default 5) the breaker opens and refuses that peer for the timeout, for faults that were entirely ours.

## Attribution lost in the other direction

- **`services/blockvalidation/get_blocks.go:997`** — a `GetPeersAtMaxHeight` failure now returns `NewServiceUnavailableError("local peer discovery unavailable during subtree failover", snapshotErr)` and **discards `primaryErr`**; before, it logged a warning and fell through to the terminal `NewExternalError(... primaryErr)`. `IsTransientLocalError` matches `ErrServiceUnavailable`, so `processCatchupChItem` takes the silent `isLocalCatchupFault` branch: no `ErrExternal`, no peer blame, no `ReportPeerFailure`, no rotation. Concretely: p2p restarts for a few seconds while the primary is serving poisoned subtree_data — the bad primary is never attributed, and catchup keeps re-selecting it until the per-block attempt cap trips a cooldown. Log the discovery failure and continue to the terminal `ErrExternal` carrying `primaryErr`.
- **`services/p2p/sync_coordinator.go:711`** — `registryRecordSyncFailure` was deleted, and the remaining `RecordCatchupFailureWithKind` updates only the catchup-specific counters. `ReputationScore` is what `evaluateSyncPeerLocked` uses for its `< 20.0` eviction and what `PeerSelector.basicEligible` gates on, and the new "unattributed / non-current failure" branch does nothing but call `evaluateSyncPeerLocked` — so on that branch a repeatedly-failing peer produces no reputation movement at all. The 1-minute `LastSyncAttempt` cooldown keeps it from hot-looping, so this is degradation rather than a wedge, but the comment block this PR removed argued the opposite trade. Confirm it's intentional.

## Local-fault handling

- **`services/blockvalidation/get_blocks.go:838`** — `dlCancel()` now runs inline before `subtreeStore.Set`, but the later-registered `defer subtreeDataReader.Close()` still runs after it, and `countingReadCloser.Close` → `onClose` → `tracing.DecoupleTracingSpan(ctx, …)` returns **the same ctx unchanged** when tracing is off (`tracing_enabled` defaults false, `util/tracing/tracing.go:636`). So `RecordBytesDownloaded` always runs on a cancelled context on the *success* path. Pre-PR the cancel lived in a `defer` registered before the reader's, so LIFO ran `Close()` while the ctx was live. Effect: N `failed to record … bytes downloaded from peer` warnings per N-subtree block, and peer `BytesReceived` never advances. Record before cancelling, or use `context.WithoutCancel`.
- **`services/blockvalidation/get_blocks.go:579`** — `fetchAndStoreSubtreeData` gained an explicit `if ctx.Err() != nil` guard before its `NewStorageError` (727-733); its sibling `fetchAndStoreSubtree` did not, so `findLocalSubtreeFile`/`subtreeStore.Get` failures are wrapped as `NewStorageError` unconditionally (579, 588, 652). With a ctx-honouring backend, a clean shutdown mid-catchup makes every in-flight subtree goroutine return `ErrStorageError`, which hits the new branch at `Server.go:2011` and logs `Errorf("Local STORAGE error during catchup … node may fall behind until storage recovers")`. That trains operators to ignore the one alarm this PR added to make real storage outages visible.

## Invariant and hygiene

- **`util/http.go:574` and `:585`** — `errors/error_utils.go:239` now *asserts* "every peer URL is redacted to scheme://host before entering an error", and uses that to justify the `strings.Contains` sentinel fallback in `(*Error).Is`. But `executeHTTPRequest` wraps the raw `*url.Error` (renders `Get "<full url>": …`) and interpolates unredacted `rawURL` in the HTML branch. The retry twin `doRequestReaderWithRetryAfter:964` unwraps `*url.Error` and uses `displayURL` for exactly this reason. Reached from `catchup.FetchHeadersWithRetry` and subtreevalidation, both on gossiped `DataHubURL`s. Not exploitable today, but the documented invariant is false and the next classifier written under it inherits the hole.
- **`util/http.go:677`** — `const maxRetryAfterSeconds = 1 << 33` is compared against an `int`. On a 32-bit target that constant isn't representable as `int` and the comparison is a compile error. Moot if nothing builds 32-bit; use `int64(secs)` or a ceiling ≤ `MaxInt32` if anything does.
- **`util/http.go:866`** — `doHTTPRequestWithRetry` has no production caller; only `util/http_test.go` references it. Every live call site uses `DoHTTPRequestBodyReaderWithRetryFunc` or `DoHTTPRequestBoundedWithRetry`. Wire it up or drop it.

## Checked and cleared

`readTransactionAllocationSafe`/`scanSerializedTransaction` is byte-for-byte compatible with go-bt v2.6.9 `Tx.ReadFrom`/`Input.readFromWithArena`, including the `0xEF` extended marker, extended field ordering, and the `inputCount==0 && outputCount!=0` case; `NewTxFromBytes` round-trips it. `newBlockFromReader` keeps the `recover()` net `NewBlockFromBytes` had. `fetchBlocksBatch` requiring exactly `n` blocks is not a behaviour change — `batchFetchAndDistribute:263` already rejected short batches. `HandleCatchupFailureForPeer` doesn't re-enter `decisionMu` via `evaluateSyncPeerLocked`. `shouldStopPeerFailover`'s pacing-marker walk survives non-`*Error` links in the chain.

icellan:
Re-reviewed at `d09818529` (merge-base `99b8579b4`, 44 files, 3,316 lines of non-test production code). Build, vet, and the full suites for `util`, `model`, `errors`, `settings`, `blockvalidation`, `blockvalidation/catchup` all pass. No P0.

Most of what I raised earlier is fixed — the coinbase cap is now a threaded parameter so only the catchup reader is bounded, `subtreeStore.Set` no longer routes through the peer classifier, and `HandleCatchupFailureForPeer` falls through to re-evaluation instead of returning early on a non-current peer. Two things worth a second look, both of which cut against what I said last round.

**`services/blockvalidation/catchup_performance.go:465` — the snapshot fix overcorrected.**
The original bug is genuinely gone: `get()` no longer memoises the error, so a transient `GetPeersForCatchup` blip can't disable failover for the whole block. But the replacement never caches failures *and* runs `load()` while holding `s.mu`. So on a p2p outage mid-block, up to `SubtreeFetchConcurrency` (32) subtree goroutines each fail over and each blocks on its own sequential gRPC attempt, serialised behind the mutex, inside the per-subtree wall clock. One fast failure becomes 32 serialised RPC timeouts. `errorOnce` suppresses the repeated log but not the repeated calls. A short negative-cache window — cache the failure for a second or two rather than for the block — sits between the two failure modes.

**`services/blockvalidation/get_blocks.go:41` (with `settings/settings.go:404`) — the 8/s pacing default is the binding constraint after all.**
I withdrew this one earlier on the grounds that `asset_httpHeavyRateLimit` already caps at 10/s server-side, so the client's 8/s couldn't bite. That was wrong arithmetic. Each subtree costs **two** limiter tokens — `fetchSubtreeFromPeer` → `/subtree` and `fetchSubtreeDataFromPeer` → `/subtree_data` — and the bucket is keyed on `baseURL`, so all 32 goroutines from `subtree_fetch_concurrency` share one 8/s budget. That's 4 subtrees/sec against a server limit of 10/s: the client is the tighter of the two, and the concurrency setting is neutralised.

Concretely, catching up 1,000 blocks × 100 subtrees from a single peer — a small or private network, or just the common case where `filterMaxHeightPeers` yields one peer at max height — spends ≥25,000s (~7h) purely waiting in `awaitPeerFetchSlot`, on top of transfer time. Nothing errors: `subtree_fetch_timeout` (120s) and `subtree_data_fetch_timeout` (10m) are both per-request, so the node just goes slow, silently. Either raise the default or charge one token per subtree instead of one per request.

Four smaller findings from the same pass (`buildHTTPError` diagnostics loss now affecting every `DoHTTPRequest*` caller tree-wide, the per-attempt timeout suppressed by the new whole-response deadlines, `readTransactionAllocationSafe` taxing the trusted store-read-back path, and a misplaced `markCatchupFailureReported` on the local-error branch) — say the word and I'll write those up too.

ordishs (changes requested):
### Overview
The PR fixes the p2p-only IBD wedge (issue 1174). It maps HTTP 429 to the retryable `ErrServiceUnavailable` class, adds jittered backoff that honours `Retry-After`, and adds client-side per-peer pacing. It also adds wall-clock bounds for block and subtree fetches, a bounded coinbase scan on the transport path, and pruned-peer deprioritisation. The retry driver, pacing reservation logic, aggregate batch byte budget and circuit-breaker probe release are careful and well tested.

One defect is a regression against the PR's own goal. The PR states that peer text can no longer forge a "local" classification, and it adds new `IsLocalError` gates on that basis. But `unwrapHTTPURLError` keeps the raw `net/http` transport cause, and that cause can embed peer bytes verbatim. `(*Error).Is` then substring-matches the rendered chain.

### Findings

**services/blockvalidation/catchup_get_block_headers.go:303-306** (root cause: util/http.go:1086-1088, util/http.go:607, errors/error_utils.go:238-248) — P1
**Issue**: The header-fetch failure path now skips the circuit breaker and the reputation charge when `errors.IsLocalError(err)` is true. `IsLocalError` → `IsContextError` → `(*Error).Is(context.Canceled)` does `strings.Contains(e.Error(), "context canceled")`. The transport error that `DoHTTPRequest` / `doRequestReaderWithRetryAfter` wrap via `unwrapHTTPURLError(err)` keeps the `net/http` message, and that message quotes peer-sent header bytes. The new comment at `errors/error_utils.go:238-245` says this cannot happen. It can.
**Failure**: A peer answers `/headers_from_common_ancestor/...` with `HTTP/1.1 200 OK\r\ncontext canceled\r\n\r\n`. The error is `SERVICE_ERROR: http request [...] failed -> net/http: HTTP/1.x transport connection broken: malformed MIME header: missing colon: "context canceled"`, and `IsLocalError` returns true. I reproduced this against `DoHTTPRequest`, `DoHTTPRequestBoundedWithRetry` and `DoHTTPRequestBodyReaderWithRetryFunc` (all three return `IsLocal=true`). Then:
1. `recordFailure()` and `reportCatchupFailure` are skipped. On main, both always ran on this path.
2. The error is returned `markCatchupFailureReported`, so the top level does not charge the peer either.
3. `processCatchupChItem` puts the `ServiceError` chain in the silent `isLocalCatchupFault` bucket, so `ReportPeerFailure` never fires and the sync peer is not rotated.
The hostile sync peer keeps its reputation and a closed breaker, and it re-wedges catchup every cycle. The same forged text also makes `shouldStopPeerFailover` return true on the subtree path, which aborts failover to honest peers. The subtree-path gate existed on main, but this PR's comments claim that gate is now safe.
**Fix**: Never wrap raw transport error text in errors from peer fetches. In `executeHTTPRequest` and `doRequestReaderWithRetryAfter`, map the `httpClient.Do` error to a fixed-message typed error, for example:
- `net.Error.Timeout()` → `NewNetworkTimeoutError`
- `syscall.ECONNREFUSED` → `NewNetworkConnectionRefusedError`
- anything else → `NewNetworkError("transport failure")`
Keep no wrapped cause. Also add a regression test in `util/http_redaction_test.go` with a raw-TCP server that sends a malformed header line containing `context canceled`. Assert `IsLocalError == false`. As defence in depth, the catchup gates could check context codes only (`errors.Is(err, errors.ErrContextCanceled)` plus stdlib `errors.Is` on the unwrapped chain) instead of the substring fallback.

**util/http.go:950 / services/blockvalidation/catchup.go:432 / services/blockvalidation/Server.go:2043** — P2
**Issue**: When 429 retries run out by attempt count, `retryHTTP` returns `NewServiceUnavailableError("http request still failing after %d attempt(s)")`. With `Retry-After: 1` this is the normal outcome: about 6 s of sleep, well inside the 60 s budget and the 2 m deadline. On the block-batch path, `releaseCatchupLock` labels this chain `local_service_unavailable` (`isPeerError=false`), and `processCatchupChItem` sends it to the silent local bucket with no `ReportPeerFailure`. The peer-attribution branch that the `retryHTTP` comment describes ("a peer that 429-spams us ... evades any reputation penalty") only runs on the rarer deadline-during-backoff path.
**Failure**: A sync peer whose `/blocks` route keeps returning 429 gets no charge and no rotation. Catchup retries the same peer until the per-block attempt cap or the 5-minute no-progress eviction. This is the issue-1174 symptom on the block path. (Before the PR, 429 was a plain `ServiceError`, which also went to the silent bucket. So this is not a regression, but the PR's stated fix does not cover it.)
**Fix**: Return a peer-attributed class when retries are exhausted on a peer URL, for example `NewNetworkError` wrapping `lastErr`, or a dedicated rate-limited marker. Then catchup charges or rotates as it does in the deadline branch.

### Verdict
**REQUEST CHANGES** — The backoff, pacing and bounded-decode work is solid, and the tests are extensive. But the new `IsLocalError` gates on the header path rely on a claim that peer bytes cannot reach the classifier, and that claim is false. With a malformed HTTP header, a hostile sync peer avoids the breaker, reputation and rotation, which recreates the wedge this PR fixes. Sanitise the transport cause (fixed-message typed errors) and add the raw-TCP regression test.

PR #1482, opened 2026-08-05
fix(bump): compute BUMP leaf offsets in 64-bit so large-block SPV proofs cannot silently wrap

### LINKED ISSUE #1427: [HARDEN-0805] Compute BUMP leaf offsets in 64-bit so large-block SPV proofs cannot silently wrap
## What can go wrong

The node can emit a merkle/SPV proof (BUMP format) that points at the *wrong* transaction, or that fails to verify at all, once blocks get large enough. A third-party SPV client verifying the proof is what would notice — the node itself raises no error — so a wrong proof ships silently.

## Where

`util/bump/format.go:90` (commit `27a0178fd`), in `ConvertToBUMP`. The global leaf offset is computed in 32-bit arithmetic:

```go
globalOffset := (uint32(proof.SubtreeIndex) << uint(subtreeLevels)) | uint32(proof.TxIndexInSubtree) //nolint:gosec
```

`globalOffset` is a `uint32`, and it then drives every sibling offset (`siblingOffset := (globalOffset >> uint(levelIdx)) ^ 1`, line 95) and the txid node offset (line 116). When a block's padded leaf count reaches 2^32 (~4.3 billion leaves), the left shift `SubtreeIndex << subtreeLevels` overflows the 32-bit value and wraps to a small number, so two transactions get colliding or wrong offsets. The `//nolint:gosec` on line 90 is what suppresses the automated overflow warning.

## Why it matters

At 1M TPS a 10-minute block is roughly 600 million transactions — within one order of magnitude of the 2^32 ceiling, and per-subtree power-of-two padding pushes the padded leaf count higher still. A wrapped offset produces a BUMP that verifies against the wrong leaf, silently, and the only automated signal (a gosec overflow warning) has been turned off with the nolint. The node emits these proofs for external ecosystem tools, so the wrong answer lands outside the node where it is hardest to trace back.

## How to reproduce / test

Unit test: call `ConvertToBUMP` with a `merkleproof.MerkleProof` whose `SubtreeIndex` and `subtreeLevels` (len of `SubtreeProof`) make `SubtreeIndex << subtreeLevels` exceed 2^32, and assert the resulting offset is either correct in a 64-bit type or a loud error — never a wrapped small number. This is a minutes-long spike; no external services needed.

## Suggested fix

Compute `globalOffset` and all derived offsets in `uint64` (BUMP offsets are conceptually unbounded leaf indices), and remove the `//nolint:gosec`. If the BUMP wire/JSON representation constrains an offset to 32 bits, add an explicit bounds check that returns a loud error when the leaf index exceeds what the format can represent, rather than silently wrapping.

## Confidence

Confirmed — the unguarded 32-bit shift was read directly on main at the cited line.

---

freemans13:
## What happened

A BUMP (BRC-74) merkle proof describes a transaction's position as a leaf index in one flat merkle tree over the whole block, and those offsets travel as VarInts — up to 64 bits — on the wire. Teranode's `ConvertToBUMP` computed the global leaf offset in 32-bit arithmetic, with the gosec overflow warning suppressed by a `nolint`. Once a block's leaf offsets cross 2^32 the subtree index shifts straight past the top bit and wraps: the node then emits an SPV proof that verifies against the *wrong* leaf, and nothing on the node notices — the failure surfaces in whichever third-party SPV client tries to use the proof.

To be clear about the scale involved, because it decides how this change should be read: with the default 32,768 transactions per subtree, the global leaf offset is close to the transaction's own position in the block, so wrapping needs roughly 4.3 billion transactions in a single block. At 1M TPS a ten-minute block is about 600 million — the same order of magnitude, about a factor of seven short. So the 32-bit arithmetic on the transaction path is **latent, not currently reachable**: this is hardening ahead of scale, and the `nolint` was hiding a real ceiling rather than a false positive.

The bug that *is* live today is smaller and was surfaced by the new input guards. Subtree proofs — the asset server's fallback when a queried hash turns out to be a subtree root rather than a transaction — carry a sentinel `TxIndexInSubtree` of −1, which the old conversion folded to `0xFFFFFFFF`. Every subtree-proof BUMP the HTTP API has ever served carried garbage offsets at every level.

This PR corrects those offsets, but **it does not make subtree-root proofs verifiable**, and I want to be plain about that rather than leave it implied. Two further defects sit on that same path: `ConstructSubtreeMerkleProof` leaves `TxID` as the zero hash, so `ConvertToBUMP` emits no level-0 leaf node and go-bc rejects the proof with "the BUMP does not contain the txid"; and its block path is built over raw stored subtree roots with no coinbase-placeholder substitution, so it would not reconcile to the header's merkle root even with a leaf. Both are written up with a repro in #1500, along with the design question of whether a BUMP proving a non-txid leaf is meaningful at all. The transaction path — the one that matters — is correct and stays covered by `coinbase_placeholder_crosscheck_test.go`.

Closes #1427.

## Changes

- **`util/bump/format.go`** — `Node.Offset` is now `uint64` and the whole offset computation runs in 64-bit; the `nolint:gosec` is gone. The binary encoder already wrote offsets as 64-bit VarInts, so the wire format is unchanged for all existing values. Inputs the arithmetic cannot represent are rejected loudly instead of wrapping: more than 64 total proof levels, negative positions, a tx index that doesn't fit its subtree's leaf count, or a subtree index that doesn't fit the block's subtree count — the last two previously smeared bits across the two segments of the flat offset and corrupted the proof silently. The −1 subtree-proof sentinel is now recognised, so its offsets are correct (the leaf being proven is the subtree root itself); the rest of that path is #1500.
- **Tests (`util/bump/format_64bit_test.go`)** — a 32-subtree-level proof puts the subtree index entirely above bit 32 and asserts every level's offset exactly, plus the full 9-byte VarInt in the binary encoding; a subtree-proof sentinel test pins the asset-server fallback shape; and each new guard has a rejection test. One existing fixture was internally inconsistent (tx index 5 in a 4-leaf subtree, subtree index 2 in a 2-subtree block — exactly the bit-smearing the guards now reject) and was made consistent, with expectations recomputed.
- **Tests (`util/bump/crosscheck_64bit_test.go`)** — converts a 34-level proof whose offsets exceed 2^32, encodes it, and re-parses it with the go-bc reference implementation, checking that a 9-byte VarInt offset survives the round trip and that Teranode's per-level offsets satisfy go-bc's own `(index >> height) ^ 1` sibling rule at every level.

## Trade-offs and testing

- Behavior change on invalid input: proofs with out-of-range indices now fail conversion with a clear error instead of producing a silently wrong BUMP. All production constructors (`ConstructMerkleProof`, `ConstructSubtreeMerkleProof`) produce in-range values, so no valid path is affected.
- JSON consumers see `offset` values beyond 2^32 only for blocks that actually have that many leaves — the same values the binary VarInt encoding always allowed.
- On what the crosscheck test does and does not prove: its sibling hashes are arbitrary bytes with no real merkle tree behind them, so it pins 64-bit VarInt interop and go-bc's per-level sibling rule, not the truth of the offset itself. The genuinely independent check — real subtrees folded to a block header's own merkle root — is the pre-existing `coinbase_placeholder_crosscheck_test.go`, at small scale. Both test comments now say this rather than overclaiming independence.
- Ran: `go test ./util/bump/ ./services/asset/... ./util/merkleproof/` green, `-race` on the bump package, full `go build ./...`, `go vet`, `golangci-lint run --new-from-rev upstream/main` (0 issues). Mutation-verified twice: truncating the arithmetic back to 32 bits fails both new tests, and an off-by-one in the segment shift fails five tests including the end-to-end placeholder crosscheck.


freemans13:
Review round 1 (BRC-74/SPV conformance with go-bc, Go correctness, integration — adversarially verified): 6 raw findings, 0 confirmed, 6 refuted. Two refuted items still produced action:

1. **go-bc crosscheck added (60117cef5)** — the conformance lens noted the new 64-bit path had no external verification (its prototype crosscheck passed during review). The suite now converts a 34-level proof whose leaf offset exceeds 2^32, encodes it, parses it with the go-bc reference implementation, and requires the folded merkle root to match an independent manual sibling fold.
2. **Follow-up issue filed as #1483** — three lenses independently converged on a pre-existing gap: subtree-proof BUMPs (the asset server's fallback) never contain the leaf they prove, because `ConvertToBUMP` doesn't consume `SubtreeRoot` and the sentinel proof carries a zero `TxID` — so the reference verifier can't verify them end-to-end. This PR fixed their garbage offsets; the missing-subject-leaf gap predates it and is deliberately not bundled.

Other refutations worth recording: the JS `2^53` JSON-precision concern requires a block with ~9 quadrillion padded leaves (physically unreachable, and every representable value is bit-identical to or strictly more accurate than main's output); and the new guards duplicating constructor invariants was judged defense-in-depth rather than fragile coupling — an out-of-range index now 500s loudly instead of serving a silently wrong proof.

freemans13:
Review round 2 (fresh reviewer, adversarial verification available): zero findings — converged after two rounds. CI green. Ready for human review.

freemans13:
Review round 3 (adversarial third pass, after two rounds had already converged to zero). The arithmetic survived the hardest checks I could aim at it; the finding worth having was about what this PR *claims*, not what it computes.

**No arithmetic bug.** Derived the offset rule from first principles and matched it against go-bc: level `d`'s sibling is `(globalOffset >> d) ^ 1`, character-for-character go-bc's own `(index >> height) ^ 1`. The segment boundary neither loses nor duplicates a level — for block level `i`, `(subtreeIndex << subtreeLevels | txIndex) >> (subtreeLevels + i)` reduces exactly to `subtreeIndex >> i`, so parity agrees on both sides of the join. The `totalLevels <= 64` cap is right rather than off by one: 64 levels means a max offset of 2^64−1, which fits, and each extreme forces the other segment to zero width (Go's defined `x << 64 == 0` then discards nothing live). Checked that no guard rejects a legitimate proof: every producer derives its level count as `ceil(log2(count))` with `index < count`, so `index < 2^levels` always holds.

**Two commits added.**

- `73809de` drops a `nolint:gosec` from the test fixture — gosec is disabled repo-wide and G115 excluded outright, so it suppressed a check that never runs, and it read oddly beside a production change that *deleted* its own gosec suppression.
- `dd1d208` pins the large offset as a hand-derived literal instead of re-writing the production expression in the test, and corrects the crosscheck's comment. That comment claimed the go-bc fold was "independent of Teranode's own arithmetic"; it isn't, because the sibling hashes are arbitrary bytes with no merkle tree behind them. It now says what it actually pins — 64-bit VarInt interop and go-bc's per-level sibling rule — and points at `coinbase_placeholder_crosscheck_test.go`, which does fold real subtrees to a header's own merkle root.

**Filed #1500, and corrected this PR's description.** The subtree-proof path is the important one. This PR fixes its *offsets* — the `-1` sentinel no longer folds to `0xFFFFFFFF` — but the resulting BUMP is still unverifiable, and the description implied otherwise. Verified end to end: `ConstructSubtreeMerkleProof` leaves `TxID` zero, so `ConvertToBUMP` emits no level-0 leaf and go-bc fails with `the BUMP does not contain the txid` even though the offsets (`2`, `0`) are now correct. Separately, that function builds its block path over raw `block.Subtrees` with no coinbase-placeholder substitution, so it would not reach the header merkle root even with a leaf. Both are in #1500 with a repro, along with the question of whether a BUMP proving a non-txid leaf is meaningful under BRC-74 at all.

I also corrected the scale framing. With 32,768 transactions per subtree the global offset tracks the transaction's own position, so wrapping needs roughly 4.3 billion transactions in one block — about seven times a 1M TPS ten-minute block. The 32-bit transaction path is latent, not currently reachable; the sentinel bug is the one live today. The old wording also claimed power-of-two padding pushed the count higher, which is wrong for the offsets actually emitted, since they are bounded by the real transaction's position.

**Verification.** `go build ./...`, `go vet`, `go test -race ./util/bump/...` and `./util/merkleproof/...` (98 tests green after merging upstream/main), `golangci-lint` clean on changed files (one pre-existing `prealloc` in `crosscheck_test.go`, an upstream file this branch doesn't touch). Mutation-tested twice: 32-bit truncation fails both new tests, and an off-by-one in the segment shift fails five, including the end-to-end placeholder crosscheck.

ordishs (approved):
Reviewed the 64-bit offset rework in `util/bump/format.go` plus both new test files, across four
reviewer lenses (Go correctness, BRC-74 protocol, adversarial input, conventions/coverage), with each
finding adversarially re-checked against current code.

**No P0 or P1 issues.** What I verified rather than assumed:

- **The arithmetic is sound.** `format.go:150` — the two segment guards at `:142`/`:146` together
  imply `globalOffset < 2^totalLevels`, so the OR is provably disjoint and cannot smear bits between
  the subtree-index and tx-index segments. The untyped `1` in `1<<uint(subtreeLevels)` takes `uint64`,
  so there is no constant overflow at 63 levels, and the only shift-by-64 cases are those where
  `totalLevels <= 64` forces the opposite segment's index to zero.
- **The new guards cannot reject a real proof.** `subtreeLevels` is `ceil(log2(subtree nodes))` from
  go-subtree's `GetMerkleProof`, `blockLevels` is `ceil(log2(len(subtrees)))` from
  `GenerateBlockMerkleProof`, and both indices are already bounds-checked upstream
  (`merkle_proof.go:170`, `:213`, `:593`). The incomplete-final-subtree lift at `merkle_proof.go:342-355`
  only raises `subtreeLevels`. No availability regression on the asset endpoint.
- **The wire format really is unchanged below 2^32.** `writeVarInt` already took a `uint64` and sizes
  by runtime magnitude, so existing values encode byte-identically; only offsets above 2^32 grow to
  the 9-byte `0xFF` form.
- **The sentinel branch matches its only producer.** `ConstructSubtreeMerkleProof` always pairs
  `TxIndexInSubtree: -1` with an empty `SubtreeProof` (`merkle_proof.go:457-468`), and a `-1` arriving
  with a non-empty `SubtreeProof` fails closed via the negative-position guard at `:138`.
- **The changed fixture in `format_test.go:41-86` was genuinely impossible, not weakened.** Old values
  were tx index 5 in a 4-leaf subtree and subtree index 2 in a 2-subtree block. The replacement is in
  range and all four offsets recompute by hand (`global = (1<<2)|3 = 7` → `6, 7, 2, 0`), with
  exact-value assertions retained.
- **The tests are load-bearing.** I re-introduced both mutations the description claims to have tested.
  Truncating to 32-bit fails `TestConvertToBUMPLargeBlockOffsets`,
  `TestConvertToBUMPAcceptsSixtyFourLevels` (both subtests) and `TestConvertToBUMP64BitCrosscheck`.
  An off-by-one in the segment shift additionally fails `TestConvertToBUMP` and
  `TestBUMPCoinbasePlaceholderEndToEnd` — the pre-existing test that folds real subtrees to a real
  header merkle root — so the offsets are pinned against genuine merkle arithmetic, not just the new
  synthetic fixtures.

Verification: `go build ./...` exit 0, `go vet ./util/bump/` exit 0,
`go test -count=1 ./util/bump/ ./util/merkleproof/` ok, `go test -race ./util/bump/` ok,
`go test ./services/asset/...` 5 packages ok / 0 fail.

Non-blocking follow-ups (P3, no action needed before merge):

- `format.go:173-174` — a zero-level proof still returns an empty `Path` with a nil error, a shape the
  package's own `Validate` rejects. Two instances, both pre-existing: legacy sync persists a 2-byte
  `CoinbaseBUMP` (nothing in-tree parses it back), and a subtree-hash request against a single-subtree
  block yields HTTP 500 rather than 404/501. Natural companion to 1500.
- `format_64bit_test.go:154-179` — adding `Validate` and `EncodeBinary` calls to the 64-level accept
  test would fully pin the "guards and `Validate` must agree" invariant documented at `format.go:67-70`;
  today changing `Validate`'s bound to `>=` would leave the suite green.
- Worth rewording `format.go:86`'s "their offsets are correct" for whoever picks up 1500 — they are
  correct in top-tree subtree-index coordinates, which is not the leaf-offset space BRC-74 defines.

ctnguyen (approved):
Review at commit [edc909c](https://github.com/freemans13/teranode/commit/edc909c3313ca95c48685f51c25443646dc19168)

This is a clean, well-scoped fix. The 64-bit offset rework is correct: the `SubtreeIndex << subtreeLevels | txIndex` OR is genuinely disjoint given the two segment guards, the maximum offset is `2^totalLevels - 1` so nothing overflows at any level boundary, and the shift-by-full-width cases are safe because the guards force the opposite segment's index to zero first. Dropping the `nolint:gosec` in favour of loud guards is the right call — out-of-range input now errors at both callers (`GetMerkleProof.go:204`, `coinbase_bump.go:51`) instead of quietly serving a wrong-leaf proof, and the wire format is unchanged since `writeVarInt` was already 64-bit. I also appreciate how honestly the scope is drawn: the docstring and PR body are explicit that subtree-root BUMPs stay unverifiable and belong to #1500 rather than pretending this closes them, and the tests are load-bearing (mutation-verified, plus a real go-bc crosscheck through a >2^32 offset). One small, very-latent asymmetry is worth a note; nothing blocking.

## Non-blocking issues

### ChiR1 — JSON offset ceiling stays at 2^53 while the binary path now reaches 2^64
**Problem:** Widening `Node.Offset` to `uint64` lifts the binary VarInt ceiling to 2^64, but the field still serializes to JSON as a bare number (`GetMerkleProof.go:223`, `c.JSONPretty`). A JavaScript SPV client parses that with `JSON.parse`, which represents it as a float64 — exact only to 2^53 — so a JSON BUMP whose leaf offset exceeds 2^53 is silently rounded on the consumer side.
**Why it matters:** It is the same failure mode this PR exists to close — an offset that resolves to the wrong leaf with nothing local noticing, surfacing only in a third-party verifier — just relocated from 2^32 to 2^53. After this change the binary and hex encodings are safe to 2^64 while the JSON encoding is not, and that gap is undocumented. Reaching it needs a block with more than 2^53 padded leaves, far beyond even the ~2^32 the PR itself calls "not currently reachable", which is why this is non-blocking rather than a defect in the fix.
**Fix:** Either document on `Node.Offset` that the JSON representation is exact only to 2^53 (binary/hex being authoritative), or emit the offset as a JSON string above the float64-safe range. No action is needed for any currently-reachable block.

Issue #1051, opened 2026-06-08
Unit test suite (`make test`) peaks ~110GB+ — reduce per-package memory to fit standard runners


oskarszoon:
## Summary

The `test` CI job (`make test` → `go test -race -coverpkg=./...` across the whole repo) intermittently dies with `Error: Process completed with exit code 143`. 143 = SIGTERM: the host runs out of memory mid-run and the job's process tree is killed (it is **not** the compile-time OOM — tests are already passing when `make` is terminated).

It is the **only** CI job affected because it is the only one that runs `-race -coverpkg=./...` over the entire module in a single `go test` invocation. A handful of test binaries have very high peak RSS, and at the default parallelism (`-p` = number of cores) enough of them overlap to exhaust the runner.

## Evidence — measured peak RSS per test binary

Captured via `VmHWM` (the kernel's monotonic peak-RSS high-water mark, from `/proc/<pid>/status`) on a 32-core / 128GB run:

```
BINPEAK 34.2G sql.test
BINPEAK 31.4G blockassembly.test
BINPEAK 19.6G subtreeprocessor.test
BINPEAK 14.8G txmetacache.test
BINPEAK 10.8G model.test
BINPEAK  4.9G aerospike.test
```

At `-p=32` these overlap and the job-wide peak reaches **~110GB+**, which is why a 16-core/64GB runner OOMs (intermittently — it depends on which heavy binaries happen to run concurrently).

Amplifiers, both of which multiply the underlying allocation cost:
- **`-race`** — roughly 5–10× memory plus the race detector's own shadow memory (mmapped by the C tsan runtime).
- **`-coverpkg=./...`** — every test binary is instrumented for the entire repo (bigger binaries + more memory), not just its own package.

## What these "groupings" are

Each `<name>.test` is the compiled Go **test binary for one package** — `go test` builds one per package (`<pkgdir>.test`) and runs several concurrently. Mapping the offenders to their packages:

| binary | package(s) | test files | what it exercises |
|---|---|---|---|
| `sql.test` | `stores/blockchain/sql` **and** `stores/utxo/sql` | 62 / 15 | SQL-backed stores: blockchain store (Postgres/SQLite testcontainers) and the SQL UTXO backend. **Two packages compile to the same binary name** — the 34.2G is the larger of the two, almost certainly `stores/blockchain/sql`. |
| `blockassembly.test` | `services/blockassembly` | 23 | Block assembly service: builds blocks/subtrees in memory. |
[...]

PR #1484, opened 2026-08-05
fix(model): carry the block transaction count in 64-bit so a huge block validates instead of retrying forever

### LINKED ISSUE #1428: [HARDEN-0805] Carry the block transaction count in 64-bit so a huge block validates instead of retrying forever
## What can go wrong

A block carrying more than about 4.29 billion transactions (2^32) is neither accepted nor cleanly rejected — the node treats it as a transient failure and re-fetches and re-validates it forever, making no progress. Because the trigger is deterministic (the same oversized block), every node on the fleet would wedge on it at the same time. This is the same "restart doesn't help / silent stall" shape as the 4 August outage.

## Where

`model/Block.go:974` (commit `27a0178fd`), in `checkDuplicateTransactions`:

```go
transactionCountUint32, err := safeconversion.Uint64ToUint32(b.TransactionCount)
if err != nil {
    return errors.NewProcessingError("[checkDuplicateTransactions][%s] failed to convert transaction count to int", b.String(), err)
}
```

`TransactionCount` is a `uint64`, but the duplicate-check path narrows it to `uint32`; a count above `math.MaxUint32` fails the conversion and returns a **processing** (transient) error. (A sibling conversion at `model/Block.go:799` in `releaseTxMap` is guarded with `if ... err == nil`, so it silently skips the pool return rather than erroring — that one is harmless; the dangerous one is line 974, which returns the error up the validation path.)

## Why it matters

Post-Genesis BSV has no block-size limit, and SV Node stores the transaction count as a CompactSize (up to 2^64) in a `size_t` vector, so a block with more than 2^32 transactions is consensus-valid and SV Node accepts it. If Teranode returns a processing error that block validation retries, the node loops on a block the rest of the network has accepted — a liveness stall, not a clean verdict.

## How to reproduce / test

Construct a block whose `TransactionCount` exceeds `math.MaxUint32` (stub subtree nodes so the structure claims a count of 2^32+1), run the duplicate-check / validation path, and assert it validates with the count handled in 64-bit end to end — never returning a retryable processing error. Cheap unit probe.

## Suggested fix

Carry the transaction count in `uint64` through the duplicate-check path (map sizing, concurrency bookkeeping) so a block with more than 2^32 transactions validates. Do **not** convert the ceiling into a Teranode-only block-invalid rejection — that would split Teranode from BSV on a block SV Node considers valid, the exact hazard being avoided. Any true hard cap must be a BSV-coordinated consensus change, never a unilateral Teranode reject.

## Confidence

Confirmed that the count is narrowed to `uint32` and that a mismatch returns a `ProcessingError` (code read on main). Strong hypothesis on the "retries forever" consequence, which depends on the block-validation caller treating a `ProcessingError` from `checkDuplicateTransactions` as retryable — confirm by tracing how that caller handles the returned error before sizing the fix.

---

freemans13:
## What happened

Post-Genesis BSV has no block-size limit, and the wire format carries a block's transaction count as a CompactSize — up to 2^64. Teranode stores the count as `uint64`, but the duplicate-transaction check narrowed it to `uint32` and returned a *processing* error when the conversion failed. Block validation treats processing errors as transient, so a block with more than about 4.29 billion transactions would be refetched and revalidated forever — never accepted, never cleanly rejected. Because the trigger is deterministic, every node in the fleet wedges on the same block at the same time, while SV Node accepts it and moves on. This is the same rare-state, silent-stall shape as the August 4th outage class the hardening map targets.

Closes #1428.

## Changes

- **`model/block_validation_pools.go`** — the txMap pool API (`GetTxMap`, `PutTxMap`, `txMapClassIdxFor`) now takes the entry count as `uint64`. Counts beyond the largest size class allocate a fresh map with the preallocation hint bounded by `txMapAllocHint`, instead of failing. The bound sizes the initial allocation only — the swiss maps grow on insert, so a bounded hint does not cap capacity.
- **`model/Block.go`** — the failing conversion and its error path are gone. `checkDuplicateTransactions` no longer sizes anything from `b.TransactionCount` at all: a new `txMapEntryCount()` sums `len(subtree.Nodes)` over the loaded subtree bodies, which is an exact upper bound on what the duplicate check inserts (bar the coinbase placeholder it skips). That value is stashed as `b.txMapCount` so the pool release key cannot drift from the value used at `GetTxMap` time, and the disk-backed `FilterCapacity` derives from it too. The release path (`releaseTxMap`) uses `b.txMapCount`, removing a now-dead conditional: the old skip-on-conversion-failure branch was unreachable, because the same narrowing had already failed earlier and left the map unassigned — no production leak existed.
- **`model/Block.go`, parent-spends sizing** — `validOrderAndBlessed` sizes its map from the same body-derived count via a new `parentSpendsCapacity()`, so a claimed count can no longer drive that allocation either. The multiply degrades to the unmultiplied count on overflow; the remaining overflow route is the operator-supplied `block_parentSpendsCapacityMultiplier`, since the body-derived count is bounded by what fits in memory.
- **No new rejection**: per the issue's warning, the ceiling is not converted into a Teranode-only block-invalid verdict — that would split Teranode from BSV on a block SV Node considers valid. A genuine hard cap must be a coordinated consensus change.
- **Tests (`model/block_txcount_64bit_test.go`)** — the issue's own scenario is now pinned end to end: `TestCheckDuplicateTransactionsAboveUint32` drives the real duplicate-check path with `TransactionCount = MaxUint32 + 1` and asserts no error. It costs a four-node map, because the claimed count no longer sizes anything — which is what previously made this untestable without the tens-of-GB allocation this package documents avoiding under issue 1051. The size classes are shrunk for that test regardless, so a future regression to claimed-count sizing fails the assertion instead of preallocating for 2^32 entries. Alongside it: the pure classification and hint functions across counts up to 2^64, a pooled round trip on the 64-bit key, `parentSpendsCapacity` including its overflow fallback, and `txMapEntryCount` on the claim-exceeds-body and zero-subtree shapes. Both new guards are mutation-proved.

## Trade-offs and testing

- The disk-backed map path already carried the count as `uint`/`uint64` and is unchanged.
- **Sizing comes from the block body, not the claimed count** (the substance of ctnguyen's review — see the thread below). An intermediate revision bounded the hint instead, which limited the blast radius of a peer-chosen number rather than declining to trust it. Both approaches are kept: the hint is still bounded by the largest pooled size class, because `MaxUint32` sits past the point where the map constructor's own per-bucket arithmetic — `(length + length/5)`, computed in `uint32` — silently wraps, which would preallocate ~859M entries by wraparound rather than by design. The bound is now defence in depth behind an input that is no longer attacker-chosen.
- **This fix removes the retry loop; it does not make a genuinely >2^32-transaction block validatable in memory mode.** A block that really carries that many transactions needs a multi-hundred-GB txMap (~46 bytes per entry), so on a memory-budgeted node the outcome is an OOM rather than a successful validation, and `block_diskMapDirs` is the real answer at that scale. What changes is that the block is no longer failed by a *type conversion* and refetched forever. Note the memory cost now tracks what a block actually carries rather than what it claims, so a block merely *claiming* a huge count is cheap.
- Two other `uint32` narrowings remain in `model` (block height, coinbase parts loop); both are height- or protocol-bounded, not transaction-count-bounded, and are untouched.
- Ran: `go build ./...`, `go vet ./model/`, `go test ./model/` (549 green), `-race` on the new tests, `golangci-lint run --new-from-rev upstream/main` (0 issues). Peak test RSS for the new file measured under 1GB (the original round-1 version peaked at ~44GB and was reworked). The round-5 overflow guard was mutation-proved: it fails against the previous `MaxUint32` clamp.

### DIFF
--- a/model/Block.go
+++ b/model/Block.go
@@ -795,10 +795,9 @@ func (b *Block) releaseTxMap() {
 		// Return the pooled in-memory map for reuse on the next block.
 		// b.TransactionCount was set in GetAndValidateSubtrees before
 		// checkDuplicateTransactions ran, so it matches the value used at
-		// GetTxMap time and the map lands in the correct size-class pool.
-		if n, err := safeconversion.Uint64ToUint32(b.TransactionCount); err == nil {
-			PutTxMap(poolable, n)
-		}
+		// GetTxMap time and the map lands in the correct size-class pool
+		// (counts above every size class are dropped by PutTxMap).
+		PutTxMap(poolable, b.TransactionCount)
 	} else if closer, ok := b.txMap.(io.Closer); ok {
 		_ = closer.Close()
 	}
@@ -971,11 +970,6 @@ func (b *Block) checkDuplicateTransactions(ctx context.Context, logger ulogger.L
 	g := new(errgroup.Group)
 	util.SafeSetLimit(logger, g, concurrency)
 
-	transactionCountUint32, err := safeconversion.Uint64ToUint32(b.TransactionCount)
-	if err != nil {
-		return errors.NewProcessingError("[checkDuplicateTransactions][%s] failed to convert transaction count to int", b.String(), err)
-	}
-
 	// set the expected subtree size based on the first subtree in the block
 	subtreeSize := 0
 	if len(b.SubtreeSlices) > 0 {
@@ -1001,8 +995,10 @@ func (b *Block) checkDuplicateTransactions(ctx context.Context, logger ulogger.L
 	} else {
 		// Draw the txMap from a size-class pool so the (potentially multi-GB)
 		// backing storage is reused across blocks. PutTxMap is called at
-		// release time below, keyed by the same transactionCountUint32.
-		b.txMap = GetTxMap(transactionCountUint32)
+		// release time below, keyed by the same 64-bit transaction count —
+		// counts above every size class allocate fresh with a clamped
+		// preallocation hint instead of failing the block (issue 1428).
+		b.txMap = GetTxMap(b.TransactionCount)
 	}
 	for subIdx := 0; subIdx < len(b.SubtreeSlices); subIdx++ {
 		subIdx := subIdx
@@ -1013,7 +1009,7 @@ func (b *Block) checkDuplicateTransactions(ctx context.Context, logger ulogger.L
 		})
 	}
 
-	if err = g.Wait(); err != nil {
+	if err := g.Wait(); err != nil {
 		// return the error from above without wrapping it
 		return err
 	}
--- a/model/Block.go
+++ b/model/Block.go
@@ -797,11 +797,14 @@ func (b *Block) releaseTxMap() {
 		_ = diskMap.Close()
 		ClearTxMapStats()
 	} else if poolable, ok := b.txMap.(*txmap.SplitSwissMapUint64); ok {
-		// Return the pooled in-memory map for reuse on the next block.
-		// b.txMapCount is the value checkDuplicateTransactions sized the map
-		// from, stashed there for exactly this purpose, so the map lands in the
-		// size-class pool it came from (counts above every size class are
-		// dropped by PutTxMap).
+		// Return the pooled in-memory map for reuse on the next block. The
+		// invariant this relies on is narrow and local: checkDuplicateTransactions
+		// assigns b.txMapCount immediately before GetTxMap and nothing between
+		// there and here writes it, so the Put key equals the Get key and the map
+		// lands in the pool it came from (counts above every size class are
+		// dropped by PutTxMap). Deliberately not stated in terms of
+		// b.TransactionCount, which GetAndValidateSubtrees only recomputes when
+		// Valid took the `subtreeStore != nil && len(b.Subtrees) > 0` branch.
 		PutTxMap(poolable, b.txMapCount)
 	} else if closer, ok := b.txMap.(io.Closer); ok {
 		_ = closer.Close()
--- a/model/Block.go
+++ b/model/Block.go
@@ -108,6 +108,11 @@ type Block struct {
 	subtreeLength   uint64
 	subtreeSlicesMu sync.RWMutex
 	txMap           txmap.TxMap
+	// txMapCount is the entry count txMap was sized from, and the pool key it
+	// must be returned with. Derived from the loaded block body by
+	// txMapEntryCount, not from the peer-supplied TransactionCount. Stashed so
+	// the release key cannot drift from the one used at GetTxMap time.
+	txMapCount      uint64
 	medianTimestamp uint32
 	// nodeAllocator, if non-nil, supplies pooled backing slices for the
 	// per-subtree Node arrays during GetAndValidateSubtrees. Only the
@@ -793,11 +798,11 @@ func (b *Block) releaseTxMap() {
 		ClearTxMapStats()
 	} else if poolable, ok := b.txMap.(*txmap.SplitSwissMapUint64); ok {
 		// Return the pooled in-memory map for reuse on the next block.
-		// b.TransactionCount was set in GetAndValidateSubtrees before
-		// checkDuplicateTransactions ran, so it matches the value used at
-		// GetTxMap time and the map lands in the correct size-class pool
-		// (counts above every size class are dropped by PutTxMap).
-		PutTxMap(poolable, b.TransactionCount)
+		// b.txMapCount is the value checkDuplicateTransactions sized the map
+		// from, stashed there for exactly this purpose, so the map lands in the
+		// size-class pool it came from (counts above every size class are
+		// dropped by PutTxMap).
+		PutTxMap(poolable, b.txMapCount)
 	} else if closer, ok := b.txMap.(io.Closer); ok {
 		_ = closer.Close()
 	}
@@ -976,10 +981,14 @@ func (b *Block) checkDuplicateTransactions(ctx context.Context, logger ulogger.L
 		subtreeSize = b.SubtreeSlices[0].Size()
 	}
 
+	// Size both map variants from the loaded block body rather than the
+	// peer-supplied TransactionCount, and keep the value for the release key.
+	b.txMapCount = b.txMapEntryCount()
+
 	if len(diskMapDirs) > 0 {
-		// An empty (coinbase-only) block has TransactionCount == 0, but the
-		// mmap-backed table rejects a zero capacity — clamp to 1.
-		filterCapacity := b.TransactionCount
+		// An empty (coinbase-only) block inserts nothing, but the mmap-backed
+		// table rejects a zero capacity — clamp to 1.
+		filterCapacity := b.txMapCount
 		if filterCapacity == 0 {
 			filterCapacity = 1
 		}
@@ -995,10 +1004,10 @@ func (b *Block) checkDuplicateTransactions(ctx context.Context, logger ulogger.L
 	} else {
 		// Draw the txMap from a size-class pool so the (potentially multi-GB)
 		// backing storage is reused across blocks. PutTxMap is called at
-		// release time below, keyed by the same 64-bit transaction count —
-		// counts above every size class allocate fresh with a clamped
-		// preallocation hint instead of failing the block (issue 1428).
-		b.txMap = GetTxMap(b.TransactionCount)
+		// release time below, keyed by the same 64-bit count held in
+		// b.txMapCount — counts above every size class allocate fresh with a
+		// bounded preallocation hint instead of failing the block (issue 1428).
+		b.txMap = GetTxMap(b.txMapCount)
 	}
 	for subIdx := 0; subIdx < len(b.SubtreeSlices); subIdx++ {
 		subIdx := subIdx
@@ -1015,6 +1024,33 @@ func (b *Block) checkDuplicateTransactions(ctx context.Context, logger ulogger.L
 	}
 
 	return nil
+}
+
+// txMapEntryCount returns the number of entries the duplicate check will insert,
+// summed over the loaded subtree bodies. It is exact bar one:
+// checkDuplicateTransactionsInSubtree skips the coinbase placeholder at
+// subIdx 0/txIdx 0 and nothing else.
+//
+// This deliberately does NOT use b.TransactionCount. That field is read straight
+// off a wire varint by readBlockFromReader, and GetAndValidateSubtrees only
+// recomputes it from the real subtree contents inside the
+// `subtreeStore != nil && len(b.Subtrees) > 0` guard in Valid, while the
+// duplicate check runs unconditionally. Sizing an allocation from it therefore
+// trusts a peer-chosen integer on paths where nothing has reconciled it against
+// the body (issue 1501). The slices below are already loaded by the time any
+// caller needs this, so the honest count costs one len() per subtree: for a
+// truthful block the derived number *is* TransactionCount, and for a block
+// claiming more than it carries the map is sized for what it actually carries.
+func (b *Block) txMapEntryCount() uint64 {
+	var n uint64
+
+	for _, subtree := range b.SubtreeSlices {
+		if subtree != nil {
+			n += uint64(len(subtree.Nodes))
+		}
+	}
+
+	return n
 }
 
 // checkDuplicateTransactionsInSubtree checks for duplicate transactions in a subtree.
@@ -1064,6 +1100,35 @@ func (b *Block) checkDuplicateTransactionsInSubtree(subtree *subtreepkg.Subtree,
 	return nil
 }
 
+// parentSpendsCapacity returns the capacity hint for the parent-spends map:
+// entryCount (taken from the loaded block body) times the assumed average
+// inputs per transaction. A multiplier of 0 is treated as 1, and the result is
+// never 0 because the mmap-backed table rejects a zero capacity.
+//
+// entryCount is bounded by the subtrees actually held in memory, so it cannot
+// overflow the product on its own. The multiplier can: it is operator-supplied
+// (block_parentSpendsCapacityMultiplier) and unvalidated. On overflow this
+// returns entryCount unmultiplied — losing the headroom but keeping a real
+// number — because NewSplitSyncedParentMap computes
+// uint32((e + e/5) / nrOfBuckets), which turns a wrapped or saturated hint into
+// an arbitrary per-bucket size rather than merely a small one.
+func parentSpendsCapacity(entryCount, multiplier uint64) uint64 {
+	if multiplier == 0 {
+		multiplier = 1
+	}
+
+	capacity := entryCount * multiplier
+	if capacity/multiplier != entryCount {
+		capacity = entryCount
+	}
+
+	if capacity == 0 {
+		return 1
+	}
+
+	return capacity
+}
+
 type validationDependencies struct {
 	txMetaStore           utxo.Store
 	subtreeStore          SubtreeStore
@@ -1083,20 +1148,19 @@ func (b *Block) validOrderAndBlessed(ctx context.Context, logger ulogger.Logger,
 		return errors.NewStorageError("[validOrderAndBlessed][%s] txMap is nil, cannot check transaction order", b.String())
 	}
 
-	// Size the parent-spends map at TransactionCount * multiplier (assumed
+	// Size the parent-spends map at transaction count * multiplier (assumed
 	// average inputs/tx). For the disk-backed map this is a hard cap, so the
 	// multiplier is configurable (block_parentSpendsCapacityMultiplier); a
 	// consolidation-heavy block exceeding it overflows a segment and halts
 	// (fail-safe). 0 is treated as 1.
-	if parentSpendsCapacityMultiplier == 0 {
-		parentSpendsCapacityMultiplier = 1
-	}
-	expectedInpoints := b.TransactionCount * parentSpendsCapacityMultiplier
-	if expectedInpoints == 0 {
-		// Empty (coinbase-only) block: TransactionCount == 0, but the
-		// mmap-backed table rejects a zero capacity — clamp to 1.
-		expectedInpoints = 1
-	}
+	//
+	// The transaction count comes from the loaded block body, not from the
+	// peer-supplied b.TransactionCount — see txMapEntryCount for why (issue
+	// 1501). That alone bounds this product: the node count is limited by the
+	// subtrees actually held in memory, so a claimed count can no longer drive
+	// it. The remaining overflow route is the multiplier itself, which is
+	// operator-supplied (block_parentSpendsCapacityMultiplier) and unvalidated.
+	expectedInpoints := parentSpendsCapacity(b.txMapEntryCount(), parentSpendsCapacityMultiplier)
 
 	var psMap ParentSpendsMap
 	if len(diskMapDirs) > 0 {
--- a/model/Block.go
+++ b/model/Block.go
@@ -1163,6 +1163,13 @@ func (b *Block) validOrderAndBlessed(ctx context.Context, logger ulogger.Logger,
 	// subtrees actually held in memory, so a claimed count can no longer drive
 	// it. The remaining overflow route is the multiplier itself, which is
 	// operator-supplied (block_parentSpendsCapacityMultiplier) and unvalidated.
+	//
+	// Recomputed here rather than read from b.txMapCount, deliberately: that
+	// field is only set by checkDuplicateTransactions, and callers that invoke
+	// this method directly (rather than through Valid, which always runs step 11
+	// first) would otherwise size from a zero. Recomputing is one len() per
+	// subtree and keeps this method self-contained — please do not "tidy" it into
+	// reading the field.
 	expectedInpoints := parentSpendsCapacity(b.txMapEntryCount(), parentSpendsCapacityMultiplier)
 
 	var psMap ParentSpendsMap
--- a/model/block_txcount_64bit_test.go
+++ b/model/block_txcount_64bit_test.go
@@ -40,16 +40,36 @@ func TestTxMapClassIdx64Bit(t *testing.T) {
 	}
 }
 
-// TestTxMapAllocHintClamp pins the preallocation-hint clamp: counts beyond
-// uint32 cap the hint at MaxUint32 (preallocation only — the map resizes on
-// insert), everything else passes through unchanged.
+// TestTxMapAllocHintClamp pins the preallocation-hint bound: any count above
+// the largest pooled size class caps the hint at that class (preallocation only
+// — the map resizes on insert), everything else passes through unchanged.
 func TestTxMapAllocHintClamp(t *testing.T) {
+	largestClass := txMapSizeClasses[len(txMapSizeClasses)-1]
+
 	require.Equal(t, uint32(0), txMapAllocHint(0))
 	require.Equal(t, uint32(1<<20), txMapAllocHint(1<<20))
-	require.Equal(t, uint32(math.MaxUint32), txMapAllocHint(math.MaxUint32))
-	require.Equal(t, uint32(math.MaxUint32), txMapAllocHint(math.MaxUint32+1))
-	require.Equal(t, uint32(math.MaxUint32), txMapAllocHint(1<<40))
-	require.Equal(t, uint32(math.MaxUint32), txMapAllocHint(math.MaxUint64))
+	require.Equal(t, largestClass, txMapAllocHint(uint64(largestClass)))
+
+	for _, n := range []uint64{uint64(largestClass) + 1, math.MaxUint32, math.MaxUint32 + 1, 1 << 40, math.MaxUint64} {
+		require.Equal(t, largestClass, txMapAllocHint(n), "count %d must be bounded to the largest class", n)
+	}
+}
+
+// TestTxMapAllocHintAvoidsConstructorOverflow pins the bound below the point
+// where the map constructor's own per-bucket arithmetic wraps. It computes
+// per-bucket size as (hint + hint/5) in uint32, so any hint above ~3.58e9
+// overflows and preallocates an arbitrary amount unrelated to the count
+// (MaxUint32 wraps to ~859M entries). Every hint this package can produce must
+// stay in the range where that arithmetic is exact.
+func TestTxMapAllocHintAvoidsConstructorOverflow(t *testing.T) {
+	for _, n := range []uint64{0, 1 << 20, uint64(txMapSizeClasses[len(txMapSizeClasses)-1]), math.MaxUint32, 1 << 40, math.MaxUint64} {
+		hint := txMapAllocHint(n)
+
+		// The constructor's computation, in uint32 as it performs it, must equal
+		// the same computation carried out without truncation.
+		require.Equal(t, uint64(hint)+uint64(hint)/5, uint64(hint+hint/5),
+			"hint %d (from count %d) overflows the constructor's per-bucket arithmetic", hint, n)
+	}
 }
 
 // TestTxMapPoolRoundTrip64BitKey pins Get/Put with the 64-bit key on pooled
--- a/model/block_txcount_64bit_test.go
+++ b/model/block_txcount_64bit_test.go
@@ -1,0 +1,60 @@
+package model
+
+import (
+	"context"
+	"math"
+	"testing"
+
+	"github.com/bsv-blockchain/go-bt/v2/chainhash"
+	subtreepkg "github.com/bsv-blockchain/go-subtree"
+	"github.com/bsv-blockchain/teranode/ulogger"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCheckDuplicateTransactionsAboveUint32 pins issue 1428: a block whose
+// TransactionCount exceeds 2^32 must go through duplicate checking in 64-bit,
+// not fail a uint32 narrowing with a retryable processing error. Post-Genesis
+// BSV has no block-size limit and SV Node carries the count as a CompactSize,
+// so such a block is consensus-valid; a conversion error here made block
+// validation refetch and revalidate the same block forever, wedging the whole
+// fleet deterministically.
+func TestCheckDuplicateTransactionsAboveUint32(t *testing.T) {
+	subtree, err := subtreepkg.NewTreeByLeafCount(4)
+	require.NoError(t, err)
+	require.NoError(t, subtree.AddCoinbaseNode())
+
+	for i := byte(1); i <= 3; i++ {
+		hash := chainhash.HashH([]byte{i, 0xdd})
+		require.NoError(t, subtree.AddNode(hash, 1, 0))
+	}
+
+	block := &Block{
+		Header:           &BlockHeader{HashPrevBlock: &chainhash.Hash{}, HashMerkleRoot: &chainhash.Hash{}},
+		TransactionCount: math.MaxUint32 + 2, // the count the wire format allows but uint32 cannot hold
+		SubtreeSlices:    []*subtreepkg.Subtree{subtree},
+	}
+
+	err = block.checkDuplicateTransactions(context.Background(), ulogger.TestLogger{}, 4, nil)
+	require.NoError(t, err, "a transaction count above 2^32 must not fail duplicate checking")
+
+	// The pooled map must be releasable with the same 64-bit count.
+	block.releaseTxMap()
+	require.Nil(t, block.txMap)
+}
+
+// TestTxMapPool64Bit pins the pool API on 64-bit counts: any count must yield
+// a usable map (counts above the largest size class allocate fresh with a
+// clamped preallocation hint — the swiss maps grow on demand), and returning
+// it with the same count must not panic.
+func TestTxMapPool64Bit(t *testing.T) {
+	for _, n := range []uint64{0, 1, 1 << 20, 1 << 31, math.MaxUint32 + 1, 1 << 40} {
+		m := GetTxMap(n)
+		require.NotNil(t, m, "count %d", n)
+
+		hash := chainhash.HashH([]byte{byte(n), byte(n >> 8), 0xab})
+		require.NoError(t, m.Put(hash, 1))
+		require.True(t, m.Exists(hash))
+
+		PutTxMap(m, n)
+	}
+}
--- a/model/block_txcount_64bit_test.go
+++ b/model/block_txcount_64bit_test.go
@@ -11,14 +11,68 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// TestCheckDuplicateTransactionsAboveUint32 pins issue 1428: a block whose
-// TransactionCount exceeds 2^32 must go through duplicate checking in 64-bit,
-// not fail a uint32 narrowing with a retryable processing error. Post-Genesis
-// BSV has no block-size limit and SV Node carries the count as a CompactSize,
-// so such a block is consensus-valid; a conversion error here made block
-// validation refetch and revalidate the same block forever, wedging the whole
-// fleet deterministically.
-func TestCheckDuplicateTransactionsAboveUint32(t *testing.T) {
+// These tests pin issue 1428 — a block whose TransactionCount exceeds 2^32
+// must be handled in 64-bit, not fail a uint32 narrowing with a retryable
+// processing error that made block validation refetch the same
+// consensus-valid block forever.
+//
+// Deliberately NOT tested by materialising a >2^32-entry map: GetTxMap
+// preallocates eagerly from its hint, and a single such call costs tens of
+// GB of RSS (the same trap TestGetTxMap_OversizedAllocatesFresh documents
+// removing after it dominated CI memory, issue 1051). The narrowing seam
+// itself no longer exists at the type level — GetTxMap/PutTxMap take uint64 —
+// so the pure classification and clamp logic carries the behavioural pin.
+
+// TestTxMapClassIdx64Bit pins the size-class classification on 64-bit counts:
+// everything at or below the largest class resolves to a pool class, and
+// every count above it — including counts beyond uint32, which previously
+// could not even be expressed — resolves to the allocate-fresh path.
+func TestTxMapClassIdx64Bit(t *testing.T) {
+	largestClass := uint64(txMapSizeClasses[len(txMapSizeClasses)-1])
+
+	require.Equal(t, 0, txMapClassIdxFor(0))
+	require.Equal(t, 0, txMapClassIdxFor(1))
+	require.Equal(t, len(txMapSizeClasses)-1, txMapClassIdxFor(largestClass))
+
+	for _, n := range []uint64{largestClass + 1, math.MaxUint32, math.MaxUint32 + 1, 1 << 40, math.MaxUint64} {
+		require.Equal(t, -1, txMapClassIdxFor(n), "count %d must take the allocate-fresh path", n)
+	}
+}
+
+// TestTxMapAllocHintClamp pins the preallocation-hint clamp: counts beyond
+// uint32 cap the hint at MaxUint32 (preallocation only — the map resizes on
+// insert), everything else passes through unchanged.
+func TestTxMapAllocHintClamp(t *testing.T) {
+	require.Equal(t, uint32(0), txMapAllocHint(0))
+	require.Equal(t, uint32(1<<20), txMapAllocHint(1<<20))
+	require.Equal(t, uint32(math.MaxUint32), txMapAllocHint(math.MaxUint32))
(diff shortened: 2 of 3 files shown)


freemans13:
Review round 1 (consensus/DoS, Go, test-adequacy lenses — adversarially verified): 4 raw findings, 2 confirmed, 2 refuted.

Both confirmed findings were about the new tests, and they were measured, not speculated: driving `GetTxMap` past the largest size class with real >2^32 counts allocated tens of GB per call (up to ~44GB peak RSS across the file; ~16GB for one test under `-race` on a different machine) — because the map constructor preallocates eagerly from its hint, which also made the PR's 'grows on demand' comment wrong. This is the exact CI-memory trap the package documents removing under issue 1051, reintroduced. Fixed in 5d4b326f5: the clamp moved into an extracted `txMapAllocHint` with an honest comment, and the tests now pin the behaviour through the pure classification/clamp functions plus a cheap real-path round trip — peak test RSS drops from ~44GB to under 1GB. The type-level guarantee (the narrowing seam no longer exists) is unchanged.

Refutations worth recording: the disk-backed branch cannot be wedged by a wire-claimed huge count because the count is derived from actual subtree contents for any block that has subtrees; and the ~40GB eager allocation reachable from a claimed count of exactly 4294967295 is pre-existing on main — this PR neither introduces nor worsens it (the clamp produces the same constructor argument main already allowed).

freemans13:
Review round 3 (fresh reviewer): one documentation nit confirmed, zero code findings. The PR body (and the lead commit message) described the old release path's skipped pool-return as a production leak; the reviewer proved that branch was unreachable on main — the same narrowing failed earlier and left the map unassigned, so no leak existed. The body is corrected (the commit message stands as history; this comment is its correction). The new unconditional pool return is simpler and correct, unchanged.

freemans13:
Review round 4 (final convergence check): zero findings. Converged across four rounds — 2 confirmed in round 1 (test memory, wrong comment), 1 in round 2 (weak release assertion), 1 doc nit in round 3 (overstated leak claim), 0 in round 4 — with every confirmed finding fixed and verified. Ready for human review.

freemans13:
Review round 5 (fresh reviewer, independent of rounds 1-4): one confirmed code finding, fixed in f47ed4780, plus two body corrections.

**The clamp landed on the one value where the constructor misbehaves.** `txMapAllocHint` returned `math.MaxUint32` for counts beyond uint32. go-tx-map v1.4.1 derives its per-bucket size as `(length + length/5)` — computed in `uint32` ([`tx_map.go:1062`](https://github.com/bsv-blockchain/go-tx-map/blob/v1.4.1/tx_map.go#L1062)) — which silently wraps for any hint above ~3.58e9. `MaxUint32` is past that point, so a >2^32-transaction block preallocated ~859M entries by wraparound rather than by design:

| hint | per-bucket | preallocated entries | intended per-bucket | wrapped? |
|---|---|---|---|---|
| 1<<30 | 157,286 | 1.29e9 | 157,286 | no |
| 3,579,139,412 | 524,287 | 4.30e9 | 524,287 | no |
| **4,294,967,295** | **104,857** | **859M** | 629,145 | **yes** |

Not a correctness bug in the narrow sense — the map still holds every entry, since capacity grows on insert — but the clamp's effect was accidental, and at ~46 bytes per entry it was still tens of GB. The hint is now bounded by the largest pooled size class: in the range where the constructor's arithmetic is exact, and no larger than the biggest allocation this pool design already sanctions. The new test pins the property (every producible hint must survive that arithmetic without truncation) rather than the constant, and was mutation-proved to fail against the previous `MaxUint32` clamp.

Incidentally, counts in the [1<<30, 2^32) band previously passed through unbounded, where a hint near 3.58e9 preallocates ~4.3e9 entries; those are bounded now too.

**Two body corrections.** The trade-off bullet claimed the clamp meant "less preallocation than ideal ... relies on map growth" — the error ran in the opposite direction: `MaxUint32` was a *maximal* hint that eagerly preallocated tens of GB. And "validates instead of retrying forever" was over-readable: a >2^32-transaction block genuinely needs a multi-hundred-GB txMap, so on a memory-budgeted node the outcome becomes an OOM, not a successful validation. What this PR actually removes is the retryable *type-conversion* failure and the fleet-wide silent stall it caused; `block_diskMapDirs` is the real answer at that scale. Both are now stated plainly in the body.

**Verified clean on the rest:** pool-key symmetry between `GetTxMap`/`PutTxMap`, the unconditional release path (and that the removed conditional was genuinely unreachable), the round-2 release assertion, and no other `TransactionCount` narrowing on this path. `go build ./...`, `go vet`, `go test ./model/` (549 green), `-race` on the new tests, `golangci-lint --new-from-rev upstream/main` all clean.

**Worth a separate issue, pre-existing on main and not a defect in this PR:** `GetAndValidateSubtrees` is conditional (`model/Block.go:661`) but `checkDuplicateTransactions` runs unconditionally (`model/Block.go:716`), so on a zero-subtree or nil-store block `TransactionCount` is never reconciled against actual subtree contents and keeps the raw wire varint from `readBlockFromReader` (`model/Block.go:357`) — which then reaches `GetTxMap`. There is no consistency guard in blockvalidation. Main is affected more severely than this branch (unbounded hints up to ~4.3e9 entries there, bounded here), so this PR narrows it rather than widening it, but the underlying "trust a claimed count for sizing" shape deserves its own ticket.

freemans13:
The unreconciled-transaction-count follow-up from round 5 is now tracked as #1501. Scope note for reviewers of this PR: #1501 is pre-existing on `main` and more severe there — this branch narrows it on the txMap path by bounding the preallocation hint, and does not widen it anywhere. It is not a blocker for #1484.

One detail worth surfacing from writing it up: the count field lives in Teranode's own block serialization, not the 80-byte header, so it is not covered by proof-of-work — and the identity check just merged in #1453 compares `block.Hash()`, which is header-only, so a substituted count passes it. Same "the peer chooses the response body" reasoning as #1453, different field.

ctnguyen (changes requested):
Review at commit [f47ed47](https://github.com/freemans13/teranode/commit/f47ed4780ff491c6b49818968b93c34d6d6a5df9)

This is a well-scoped fix for a nasty failure mode: a deterministic, fleet-wide stall where the node neither accepts nor rejects a block is about the worst outcome available, and widening the pool API to `uint64` removes the narrowing seam at the type level rather than patching around it. I like that the PR deliberately refuses to turn the ceiling into a Teranode-only invalid verdict — that restraint is the right call and matches what [bsv-blockchain/teranode#1428](https://github.com/bsv-blockchain/teranode/issues/1428) asked for. The round-5 correction is also genuinely good work: I re-derived the `go-tx-map` constructor locally (`tx_map.go:1062`, `perBucket := (length + length/5) / uint32(m.nrOfBuckets)`) and the wrap point at ~3.58e9 is real, so bounding the hint at the largest size class instead of `MaxUint32` is correct, and the property test that pins it is the right shape. The diff is small and the comments explain the *why*, not the *what*. My main concern is not with what the code does but with what it now lets through: the fix bounds the blast radius of a peer-chosen number instead of replacing it with the number the code already has in hand.

## Blocking issues

### ChiR1 — Size the txMap from the loaded block body, not from the claimed transaction count

**Problem:** `GetTxMap(b.TransactionCount)` still derives the allocation from a value that, on the path this PR newly opens, has never been reconciled against the block body. Before this change a count ≥ 2^32 returned a cheap error *before* any allocation; after it, that same count reaches `txmap.NewSplitSwissMapUint64(1<<30, 8192)`, which eagerly builds 8193 sub-maps at `perBucket = (1073741824 + 214748364) / 8192 = 157286` — capacity for ~1.29e9 entries. The exact byte cost is architecture-dependent, since `dolthub/swiss` uses 16-slot groups on amd64 and 8-slot groups on the fallback build, but it comes out around 60 GB of heap either way, of which the control arrays (roughly 1.5–3 GB, same dependency) are written immediately in `swiss.NewMap`'s `for i := range m.ctrl` loop. Meanwhile `sum(len(b.SubtreeSlices[i].Nodes))` is a tight upper bound on what `checkDuplicateTransactions` will actually insert — exact to within one entry, since `checkDuplicateTransactionsInSubtree` skips the coinbase placeholder at `subIdx == 0 && txIdx == 0` and nothing else — and every one of those slices is already loaded when `GetTxMap` is called.

**Why it matters:** `readBlockFromReader` (`model/Block.go:357`) takes the count straight off a wire varint, and `GetAndValidateSubtrees` only recomputes it inside the `subtreeStore != nil && len(b.Subtrees) > 0` guard while `checkDuplicateTransactions` runs unconditionally — so a zero-subtree block carries an arbitrary claimed count all the way to the allocator (see [bsv-blockchain/teranode#1501](https://github.com/bsv-blockchain/teranode/issues/1501), filed as a follow-up rather than resolved). The end-to-end catchup reachability is recorded there as unconfirmed, so I won't call this a proven remote DoS — but the PR's stated goal is "validates instead of retrying forever", and as written the ≥2^32 band trades an infinite retry for a multi-tens-of-GB allocation attempt on a peer-supplied integer. On a memory-budgeted node that is an OOM kill and a crash loop, which is not obviously better than the stall it replaces. The PR body positions `block_diskMapDirs` as the answer at that scale, but that branch takes the same unreconciled number (`FilterCapacity: uint(b.TransactionCount)` → `mmaphash.Expected`), so it inherits the problem rather than solving it. Sizing from the body makes the fix unconditionally correct: for an honest 2^32-tx block the derived number *is* the huge number and nothing changes; for a bogus one the map is tiny.

**Fix:** derive the count once from the loaded slices, use it for both `GetTxMap` and `PutTxMap`, and stash it on the block so the release key cannot drift:

```go
// Block struct, "local" section
txMap      txmap.TxMap
txMapCount uint64 // sizing hint and Get/Put pool key for txMap

// txMapEntryCount bounds the number of entries checkDuplicateTransactions will
// insert (exact bar the skipped coinbase placeholder), taken from the loaded
// block body rather than the peer-supplied TransactionCount, which is not
// reconciled against the body on every path.
func (b *Block) txMapEntryCount() uint64 {
	var n uint64
	for _, subtree := range b.SubtreeSlices {
		if subtree != nil {
			n += uint64(len(subtree.Nodes))
		}
	}
	return n
}

// checkDuplicateTransactions
b.txMapCount = b.txMapEntryCount()
b.txMap = GetTxMap(b.txMapCount)

// releaseTxMap
PutTxMap(poolable, b.txMapCount)
```

Consider giving the disk branch the same treatment (`filterCapacity := b.txMapCount`, still clamped to a minimum of 1) in the same change, since it is two lines away and shares the hazard.

## Non-blocking issues

### ChiR2 — Removing the early error newly exposes the parent-spends sizing to counts above 2^32

**Problem:** with `checkDuplicateTransactions` no longer failing, a block with `TransactionCount ≥ 2^32` can now reach `validOrderAndBlessed` on any caller that runs that stage — it is not unconditional, since `Block.Valid` gates it on `txMetaStore != nil` and may skip it entirely via `skipOrderAndBlessedBelowCheckpoint` on the below-checkpoint outpoint-only fast path. Where it does run, `expectedInpoints := b.TransactionCount * parentSpendsCapacityMultiplier` (`model/Block.go:1094`, default multiplier 2) sizes the parent-spends map from the same unreconciled value. That multiply can wrap `uint64`, and `NewSplitSyncedParentMap` then computes `perBucket = uint32((e + e/5) / nrOfBuckets)` — a second truncation. At a count of 2^32 the untruncated result asks for ~1e10 inpoint slots, comfortably past anything a node can hold.

**Why it matters:** the wedge this PR removes is replaced by a failure one step further down rather than by a successful validation, so the outcome for the targeted scenario is still "node does not make progress". This line is pre-existing on `main` and is recorded in [bsv-blockchain/teranode#1501](https://github.com/bsv-blockchain/teranode/issues/1501), and the peak allocation was already reachable on `main` at a count just under 2^32 — so this is a widening of the trigger band, not a new ceiling. It is fair to leave for the follow-up; it is worth saying out loud in the PR body that the ≥2^32 band now lands here.

**Fix:** either bound `expectedInpoints` the way the txMap hint is bounded, or, better and consistent with ChiR1, derive it from `b.txMapCount * parentSpendsCapacityMultiplier` with a saturating multiply:

```go
expectedInpoints := b.txMapCount * parentSpendsCapacityMultiplier
if parentSpendsCapacityMultiplier != 0 && expectedInpoints/parentSpendsCapacityMultiplier != b.txMapCount {
	expectedInpoints = math.MaxUint64 // saturate rather than wrap
}
```

### ChiR3 — The #1428 scenario has no end-to-end test, and the substitute's release-key guarantee is weaker than its comment claims

**Problem:** the issue asked for a test that drives a block with `TransactionCount > math.MaxUint32` through the duplicate-check path and asserts it does not return a retryable processing error. The PR substitutes unit tests of `txMapClassIdxFor`/`txMapAllocHint` on the stated grounds that materialising such a map costs tens of GB. That constraint is real today, but it is an artefact of the hint coming from the claimed count — `txMapSizeClasses` is a package-private var, so a test can shrink it for its own duration and run the *real* path at 2^32+1 for a few kilobytes. Separately, `TestCheckDuplicateTransactionsUsesFullCount` says a "wrong or stale release key… would silently drop the map instead of pooling it", but at `TransactionCount: 4` any wrong key below `1<<30` still resolves to a pool class and still clears the map, so the assertion only catches keys above the largest class.

**Why it matters:** the classification and clamp tests pin the helpers, but nothing pins the property the issue actually cares about — that the full path returns no error for a >2^32 count. That is the regression a future refactor would reintroduce. Note that the CI-memory constraint behind this ([bsv-blockchain/teranode#1051](https://github.com/bsv-blockchain/teranode/issues/1051)) is a genuine project rule, so "just allocate the real map" is not the ask here.

**Fix:** run the real path cheaply by swapping the size classes for the test, and soften the release-key comment to what it proves:

```go
func TestCheckDuplicateTransactionsAboveUint32(t *testing.T) {
	orig := txMapSizeClasses
	txMapSizeClasses = []uint32{1 << 4}
	t.Cleanup(func() { txMapSizeClasses = orig })

	// ... build the 4-node subtree as in TestCheckDuplicateTransactionsUsesFullCount ...
	block.TransactionCount = math.MaxUint32 + 1

	// Previously a ProcessingError from the uint32 narrowing, which block
	// validation retries forever (issue 1428).
	require.NoError(t, block.checkDuplicateTransactions(context.Background(), ulogger.TestLogger{}, 4, nil))
}
```

The shrink is what keeps it cheap — `txMapAllocHint` bounds to the largest class, so with the real classes in place the same call still asks for `1<<30`. Shrinking the slice is safe because `txMapPools` is built at init from the original classes, so every surviving index still resolves to a real (and cheap) pool. Because it mutates package-global state, the `t.Cleanup` restore is mandatory and the test must not use `t.Parallel()` (in line with the project's default), otherwise it makes the rest of the package order-dependent. If ChiR1 lands, the hint stops coming from the claimed count and the shrink is no longer needed at all — the test just sets a huge `TransactionCount` on a four-node block.

## Recap

|---|---|---|---|
| ChiR1 | Size txMap from block body | YES | 70% |
| ChiR2 | Parent-spends sizing now reachable | | 45% |
| ChiR3 | Missing end-to-end >2^32 test | | 35% |

ordishs (changes requested):
The direction is right and the txMap work is well-reasoned, but the change is incomplete in a way that makes the targeted failure worse rather than better. The `Uint64ToUint32` gate in `checkDuplicateTransactions` was load-bearing for **three** downstream allocations; this PR removes it and re-bounds only one of them.

## What's right

- 64-bit `txMapClassIdxFor` / `GetTxMap` / `PutTxMap` is the correct signature change — the count *is* 64-bit and the narrowing was a type-level lie. Dropping the `safeconversion` round-trip in `releaseTxMap()` also removes a silent-drop path where `err != nil` meant the map was never pooled.
- `err :=` at `Block.go:1012` is required, not cosmetic — `err` was declared by the deleted statement.
- The `txMapAllocHint` doc comment is genuinely useful: the `(hint + hint/5)` uint32 wrap above ~3.58e9 is a real trap in go-tx-map and worth writing down.

## Must fix — the removed gate protected more than the txMap

**1. `validOrderAndBlessed` (`model/Block.go:1094`) has no equivalent bound.**

Step 12 runs on the same block, in the same `Valid()` call, and sizes its map from the same count:

```go
expectedInpoints := b.TransactionCount * parentSpendsCapacityMultiplier   // unchecked uint64 multiply
… NewSplitSyncedParentMap(parentSpendsBuckets, expectedInpoints)          // no clamp
```

`model/map.go:36` computes `perBucket = uint32((n + n/5) / uint64(nrOfBuckets))` with no bound of any kind, and `parentSpendsSizeClasses` tops out at `1<<32`, so anything above that allocates fresh and unclamped.

A genuine 5e9-tx block — exactly the block issue 1428 is about — now clears step 11 and asks for roughly 8192 × 2.2e6 swiss slots at step 12, i.e. hundreds of GB. That is a Go runtime out-of-memory fatal error, which takes the whole process down. Pre-PR the same block got a (bad, retryable) processing error at step 11 and the node stayed up. So for the block this PR is meant to unblock, issue 1428 isn't fixed — it's relocated ~100 lines down and its failure mode is upgraded from "retry forever" to "node dies".

**2. The disk-backed branch (`model/Block.go:989`) is newly unbounded.**

The deleted guard sat *above* `if len(diskMapDirs) > 0`, so it capped `FilterCapacity` at 2^32 for free. That branch still passes `uint(b.TransactionCount)` straight into `mmaphash.New` via `NewDiskTxMapUint64`, and it is the sizing mode intended for precisely these blocks. Either the mmap sizing fails and `checkDuplicateTransactions` wraps it in `NewProcessingError` — reinstating the eternally-retried error this PR set out to remove, one branch over — or a sparse multi-TB backing file exhausts the data disk.

Both items point at the same design question, which I think is worth answering explicitly rather than site-by-site: was the uint32 narrowing an *accidental* plausibility gate on `TransactionCount`? If so, replace it with a deliberate one high up — `SizeInBytes` bounds the count directly, since a tx can't be smaller than ~60 bytes — rather than clamping each allocation site independently and hoping all of them were found.

## Worth noting: the in-memory path actually improves

For the record, so this doesn't get over-corrected: the clamp *lowers* the in-memory ceiling. Pre-PR a count of 3.5e9 (expressible in uint32, so it passed the gate) reached `NewSplitSwissMapUint64(3.5e9, 8192)` → ~197 GB. Post-PR every over-max count clamps to 2^30 → ~59 GB. The unvalidated-count exposure on the zero-subtree path (`Block.Valid` skips `GetAndValidateSubtrees` when `len(b.Subtrees) == 0`, leaving `TransactionCount` as the raw `wire.ReadVarInt` from `Block.go:357`) is real but pre-existing, and this PR shrinks it. No change needed there.

## Should fix

- **`model/block_txcount_64bit_test.go:123` — the assertion doesn't guard what its comment claims.** With `TransactionCount: 4`, every plausible wrong release key (including the `0` a stale field would hold) still resolves to class 0, so `PutTxMap` still clears and pools, and `require.Equal(t, 0, pooled.Length())` passes regardless. Make the count straddle a class boundary (just above `1<<12`, released with `0`) to give it teeth.
- **Same line — it reads `pooled` after handing it back to a package-global `sync.Pool`.** Deterministic today only because the package's tests run sequentially; `TestGetTxMap_DifferentSizeClassesAreSeparate` and `TestGetTxMap_OversizedAllocatesFresh` both draw the same 4K class. Any `t.Parallel()` there makes this flaky and race-reportable. Assert before release, or draw a fresh `GetTxMap(4)` and observe it empty.
- **`model/Block.go:796` states a premise the code doesn't hold.** `GetAndValidateSubtrees` only runs when `subtreeStore != nil && len(b.Subtrees) > 0`. The conclusion (Put key == Get key) still holds; please state the actual invariant — nothing mutates `b.TransactionCount` between `GetTxMap` and `releaseTxMap` — instead of a false one.
- **The `txMapAllocHint` comment should acknowledge the `(2^30, 3.58e9]` band.** Counts there previously got an *exact* preallocation and now get a 2^30 hint, i.e. a guaranteed mid-fill rehash across all buckets with a transient peak around 1.5× the final size. I think that's the right trade — an untrusted number shouldn't drive an exact multi-GB allocation — but the comment justifies the bound only via the wrap threshold and never mentions the cost it accepts.

## Optional

- `txMapAllocHint`'s `return uint32(n)` is unreachable: its only call site is inside `if idx < 0`, which holds exactly when `n > largestClass`. Two test cases pin behaviour no caller can reach. Either drop the branch or clamp the class index at the call site.
- Over-max maps are built with `hint == largestClass` and `txMapBuckets` — identical in shape to `txMapPools[last].New` — yet bypass that pool on Get and are dropped on Put, so back-to-back huge blocks re-allocate ~59 GB each. Dropping them is defensible (a map grown to 3.5e9 entries would otherwise retain its oversized backing in the pool), but that rationale isn't recorded anywhere.
- `TestTxMapAllocHintAvoidsConstructorOverflow` asserts a local transcription of go-tx-map's `(hint + hint/5)` against itself, so a dependency bump that changes the headroom factor drops the real wrap threshold below the chosen `1<<30` and the test stays green. Pinning against observable constructor output, or a single shared threshold constant, would keep it tied to the dependency.

## Verification run

```
$ go test -race -run 'TestTxMap|TestCheckDuplicateTransactionsUsesFullCount|TestGetTxMap' ./model/
ok  	github.com/bsv-blockchain/teranode/model	5.167s
```

The new tests pass, including under `-race` — the ownership issue above is latent, not currently firing.

freemans13:
Thanks @ctnguyen — ChiR1 is right and it is a better fix than what I had. All three addressed in 909f48518.

**ChiR1 — sizing from the body: accepted as specified.** The framing that convinced me is yours: bounding the blast radius of a peer-chosen number is weaker than not trusting it, and the body is already in hand at that point. `txMapEntryCount()` sums `len(subtree.Nodes)` over the loaded slices (skipping nil entries), the result is stashed as `b.txMapCount` so the release key cannot drift, and both `GetTxMap` and `PutTxMap` use it. I took the disk-branch suggestion too — `filterCapacity` derives from `b.txMapCount`, still clamped to a minimum of 1 — since as you say it is two lines away and shares the hazard. For an honest block the derived number *is* `TransactionCount`, so nothing changes; for a bogus one the map is sized for what the block actually carries.

**ChiR2 — done rather than deferred, because ChiR1 made it cheap.** `expectedInpoints` now derives from the same body-derived count, which removes the route from a ≥2^32 claimed count into that allocation entirely rather than bounding it. The multiply is extracted into `parentSpendsCapacity(entryCount, multiplier)` and unit-tested.

One correction to your suggested fix, though: saturating to `math.MaxUint64` is not safe. `NewSplitSyncedParentMap` computes `perBucket = uint32((e + e/5) / nrOfBuckets)` (`model/map.go:36`), and at `MaxUint64` the inner `e + e/5` wraps *before* the divide, so the result truncates to an arbitrary per-bucket size — potentially a large one, not a small one. I wrote it your way first, caught it while checking that constructor, and changed the fallback to the unmultiplied `entryCount`: a real number that loses only the inputs-per-tx headroom. Worth noting the residual overflow route is no longer the block at all but the operator-supplied `block_parentSpendsCapacityMultiplier`, which is unvalidated — the body-derived count is bounded by what fits in memory. `TestParentSpendsCapacity` asserts every value the helper can return survives that constructor's arithmetic without truncating.

**ChiR3 — both parts.** `TestCheckDuplicateTransactionsAboveUint32` drives the real path at `MaxUint32 + 1` and asserts no error, which is the pin the issue asked for; it costs a four-node map now that the claimed count sizes nothing. I kept your size-class shrink anyway, for a reason beyond cost: it means a future regression to claimed-count sizing fails the assertion instead of preallocating for 2^32 entries. `t.Cleanup` restores, no `t.Parallel()`. The release-key comment is softened to what it proves — that the key still resolves to *a* pool class, not that it is the exact key.

Both new guards are mutation-proved: removing the overflow fallback fails `TestParentSpendsCapacity`, and reverting the sizing to `b.TransactionCount` fails `TestCheckDuplicateTransactionsAboveUint32` on `expected: 0x4, actual: 0x100000000` — cheaply, thanks to the shrink.

On your framing of what was left through: I agree, and I have updated #1501 accordingly — this PR now resolves the txMap and parent-spends sizing sites, so what remains there is the narrower point that `TransactionCount` is still never reconciled against the body at all, which is a validation question rather than an allocation one.

Ran: `go build ./...`, `go vet ./model/`, `go test ./model/` (all green), `-race ./model/`, `golangci-lint --new-from-rev upstream/main` (0 issues), plus `./services/blockvalidation/`, `./services/blockpersister/...` and `./services/subtreevalidation/` since both sizing sites sit on `Block.Valid`.

freemans13:
Thanks @ordishs — and apologies for replying to @ctnguyen's review first and leaving yours sitting; the overlap is real but several of your points are distinct and two of them survived that round. Addressed in 909f48518 and c65a0ff22.

Your framing is the one I've adopted: **the `Uint64ToUint32` gate was an accidental plausibility gate, and the fix is to stop trusting the number rather than to clamp each site.** `checkDuplicateTransactions` no longer sizes anything from `b.TransactionCount`; a new `txMapEntryCount()` sums `len(subtree.Nodes)` over the loaded slices — an exact upper bound bar the skipped coinbase placeholder — stashed as `b.txMapCount`. All three allocations you identified now derive from it: the in-memory txMap, the disk-backed `FilterCapacity`, and the parent-spends map.

**Must-fix 1 — `validOrderAndBlessed`.** Two changes. It derives from the body count, which removes the inflated-count route; and `parentSpendsAllocHint` now bounds the hint to the largest pooled size class, mirroring `txMapAllocHint`, with the pool class still chosen from the unbounded value so an over-max map is dropped on Put rather than retained.

But I want to be straight about what that second part does *not* buy, because I think your must-fix overstates the remedy available here. The largest parent-spends class is `1<<32`, which preallocates ~8192 × 629k slots — around 200 GB. Bounding to it makes the allocation predictable and removes the `uint32` truncation path; it does not make the block survivable. **For a genuine 5e9-tx block your analysis holds after my fix as much as before it**: hundreds of GB is inherent in holding that many inpoints, so every in-memory bound just moves which allocation dies. I don't think there is a site-local fix that changes that, which is why I've resisted presenting one.

Where I'd push back slightly: pre-PR that block got a retryable `ProcessingError` and the node stayed up, but stayed up *permanently unable to follow the chain*, refetching forever. I'd call that a different liveness failure rather than a strictly better one — and it's the one issue 1428 was filed about. What I agree with completely is that a crash loop is worse operationally, and that neither is acceptable.

So on your design question, my answer is **yes, and it needs a decision I don't want to take unilaterally.** A deliberate plausibility gate high up is the right shape, and `SizeInBytes` bounding the count via a ~60-byte floor is a good instrument. Two caveats: on the zero-subtree path `SizeInBytes` is *also* the raw `wire.ReadVarInt` (`Block.go:363`), so it only catches self-inconsistent liars, not genuine huge blocks — the body-derived sizing already covers the liar case; and a genuine block above what the node can hold needs a *clean, non-retryable, non-block-invalid* outcome, which is a new error class, not a clamp. That's the part I'd want your and @ctnguyen's agreement on before building, since issue 1428 explicitly warns against anything that reads as a Teranode-only rejection. Happy to spec it as a follow-up on #1501 (whose scope I've already narrowed to the unreconciled-count question).

**Must-fix 2 — disk branch.** `filterCapacity` derives from `b.txMapCount`, so it is no longer inflatable. The residual you describe is real and unfixed: a genuinely huge block still asks `mmaphash.New` for a very large backing, and if that fails it is wrapped in `NewProcessingError` and retried forever — the same failure this PR set out to remove, one branch over. That is the error-class half of the design question above, and I've left it rather than guess at the semantics.

**Should-fix items.** Three of four done as asked:
- `Block.go:796` now states the actual local invariant — `txMapCount` is assigned immediately before `GetTxMap` and nothing between there and `releaseTxMap` writes it — and explicitly notes it is *not* stated in terms of `TransactionCount`, for the reason you gave.
- `txMapAllocHint`'s comment now records the cost it accepts: counts in the band below the wrap point lose their exact preallocation and rehash mid-fill at ~1.5× transient peak. Agreed it's the right trade; it should have said so.
- The release assertion documents that it observes a map after returning it to a package-global pool, that the assertion must stay adjacent to the release, and that the test must never get `t.Parallel()`.

On the fourth I've done something different, and I think your suggested change wouldn't have worked. Making the count straddle a class boundary and releasing with `0` still clears and pools, because class 0 exists — so `require.Equal(t, 0, pooled.Length())` still passes and the assertion gains no teeth. The only thing that would discriminate is pool *identity*, and `TestGetTxMap_ReusesPooledInstance` deliberately declines to assert that ("sync.Pool is allowed to drop entries"). So instead the exact key is now pinned directly by asserting `b.txMapCount` is the body-derived value that `releaseTxMap` passes, and the clear-check is documented as proving only that it took the pooled branch rather than being dropped. Mutation-proved: reverting the sizing to `b.TransactionCount` fails on `expected: 0x4, actual: 0x100000000`.

**Optional items.** Two taken, one declined:
- The over-max pool bypass rationale is now recorded where you'd look for it: such a map has been *filled* past the class it was sized for, so pooling it would retain an oversized backing for every later block drawing that class; the cost is that consecutive over-max blocks each allocate fresh.
- The dependency-coupling point was the sharpest of these and you were right — the test transcribed `(hint + hint/5)` and compared it against itself, so a changed headroom factor would have moved the real threshold below `1<<30` with the test green. There is now a `txMapConstructorWrapThreshold` constant, pinned by test to be the genuine last-exact value (3579139413) with one past it asserted to wrap. Mutation-proved: moving the constant to 4e9 fails the test. It still mirrors rather than imports the factor, since go-tx-map doesn't export it — but a bump that changes it now breaks a test instead of silently invalidating the bound.
- I've kept `txMapAllocHint`'s `return uint32(n)` branch. You're right it's unreachable from the sole call site; I'd rather the function stay total so it's correct for any input than have it depend on its caller's guard, and it's now documented as deliberately a min. Say the word if you'd still prefer it hoisted and I'll change it.

I also added `TestGetParentSpendsMapBoundsPreallocation`, which pins the new bound *at the call site* rather than only in the helper — my first attempt tested only the helper and a mutation removing the bound from `GetParentSpendsMap` went undetected. It observes the underlying swiss map's `Capacity()` (the preallocation limit on a fresh map) with the size classes shrunk so the over-max branch is cheap, and asserts the bounded and unbounded hints actually differ so the assertion can't pass vacuously.

Ran: `go build ./...`, `go vet ./model/`, `go test ./model/`, `go test -race ./model/`, `golangci-lint --new-from-rev upstream/main` (0 issues), plus `./services/blockvalidation/`, `./services/blockpersister/...`, `./services/subtreevalidation/` — all green.

ctnguyen (approved):
Review at commit [c65a0ff](https://github.com/freemans13/teranode/commit/c65a0ff225d21803edaa6717c9b63271391204ee)

Strong follow-through since the last round — all three prior points landed, and landed correctly rather than by the letter. The pivot to sizing every map from the loaded block body (`txMapEntryCount`) instead of the peer-supplied `TransactionCount` is exactly the right shape: for an honest block the derived number *is* `TransactionCount`, and for a bogus one the allocation collapses to what the block actually carries, so the ≥2^32 claimed-count route into a multi-tens-of-GB allocation is gone rather than merely bounded. I re-derived the `go-tx-map` wrap point locally: `(length + length/5)` in `uint32` (`tx_map.go:1062`) stays exact up to `3579139413` and overflows at `3579139414`, so the `txMapConstructorWrapThreshold` constant is correct and the largest txMap size class (`1<<30`) sits well below it — the property test that pins this is the right instrument. I also like that the author caught the bug in my own ChiR2 suggestion (saturating the multiplier to `MaxUint64` wraps *inside* the constructor's `e + e/5`) and used the unmultiplied fallback instead. The new tests are cheap and real, and the comments now state invariants the code actually holds. What remains is the genuinely-huge-block residual both reviews already flagged, which the thread has agreed to defer — see below.

## Status of previous points

|---|---|---|---|
| ChiR1 | Size txMap from block body | YES | FIXED |
| ChiR2 | Parent-spends sizing reachable | | FIXED |
| ChiR3 | Missing end-to-end >2^32 test | | FIXED |

ChiR1: `b.txMapCount = b.txMapEntryCount()` sizes the in-memory `GetTxMap`, keys `PutTxMap` in `releaseTxMap`, and derives the disk `FilterCapacity` (clamped to 1) — the claimed count no longer sizes anything on this path. ChiR2: `validOrderAndBlessed` sizes from `parentSpendsCapacity(b.txMapEntryCount(), multiplier)`, with an overflow fallback to the unmultiplied count that survives the constructor's per-bucket arithmetic. ChiR3: `TestCheckDuplicateTransactionsAboveUint32` drives the real duplicate-check path at `MaxUint32 + 1` and asserts no error, cheaply; the release-key comment was softened to what it proves and the exact key is pinned via `b.txMapCount`.

## Non-blocking issue

### ChiR4 — The disk-backed branches still take a body-derived huge count straight into `mmaphash.New`, so a genuinely >2^32-tx block re-enters retry-forever one branch over

*Same as [@ordishs's point "Must fix 2" (06-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1484#pullrequestreview-4874784543).*
*Partially overlaps [@ordishs's point "Must fix 1" (06-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1484#pullrequestreview-4874784543): his covers the in-memory `validOrderAndBlessed` OOM; this finding extends the same unbounded-sizing concern to the disk-backed parent-spends branch, where the failure surfaces as a retryable `ProcessingError` rather than an OOM.*

**Problem:** ChiR1 removed the *claimed-count* route into a huge allocation, which is the important win. But for a block that genuinely carries >2^32 transactions, `b.txMapCount` is legitimately huge, and on the disk path it flows unbounded into the mmap constructors — `FilterCapacity: uint(filterCapacity)` at `Block.go:1001` (`NewDiskTxMapUint64`) and `FilterCapacity: uint(expectedInpoints)` at `Block.go:1173` (`NewDiskParentSpendsMap`). Whatever the inner error type, both call sites re-wrap a constructor failure in a `ProcessingError`:

```go
return errors.NewProcessingError("[checkDuplicateTransactions][%s] failed to create disk tx map", b.String(), diskErr)
```

Block validation treats a `ProcessingError` as transient and refetches — which is precisely the fleet-wide silent stall issue 1428 set out to remove, now sitting on the disk branch instead of the old `Uint64ToUint32` seam. (The in-memory path degrades to an OOM at that scale rather than a stall; neither is a clean outcome.)

**Why it matters:** the PR's headline is "validates instead of retrying forever". For the specific block the issue is about — one that really does carry that many transactions — the type-conversion wedge is gone but the retry-forever outcome can recur via the disk path. This is not a fresh defect the PR introduces carelessly: the author acknowledges it in the [14:02 response](https://github.com/bsv-blockchain/teranode/pull/1484#issuecomment-5205698870), and it is tracked in [bsv-blockchain/teranode#1501](https://github.com/bsv-blockchain/teranode/issues/1501) as a deferred design decision. It is non-blocking here because the discussion has converged that the real remedy is a *coordinated* one, not a clamp — a clamp only changes which allocation dies — and issue 1428 explicitly forbids turning the ceiling into a Teranode-only invalid verdict. Flagging so it stays visible and does not get read as fully closed.

**Fix:** none in this PR — correctly deferred. The real fix is the error-class decision in #1501: a genuinely-too-big block needs a *clean, non-retryable, non-block-invalid* outcome that does not exist today (the choices are a retryable `ProcessingError` or an OOM). In the interim, the PR body would be more honest saying the ≥2^32 genuine-block case now surfaces as a disk-path `ProcessingError` / OOM rather than the old early type-conversion error, rather than implying the retry loop is gone in all cases.

ordishs (approved):
Both blockers from my previous review are fixed, and fixed better than what I asked for. I asked for a plausibility bound on the claimed count; instead `txMapEntryCount` stops trusting it for sizing at all, which removes the problem rather than capping it.

Re-verified against the current head rather than taking the fixes on trust:

- `Subtree.Length()` is `len(st.Nodes)` (go-subtree v1.1.2 `subtree.go:188`) and `GetAndValidateSubtrees` accumulates `subtree.Length()` (`Block.go:1669`), so `txMapEntryCount()` is **numerically identical** to `TransactionCount` on the path where that field was already reconciled. The change is a no-op where the count was trustworthy and strictly safer everywhere else — that's the property that makes this safe to take.
- Inserts ≤ `txMapEntryCount()` (the loop skips only the `subIdx==0 && txIdx==0` placeholder), so the disk map's `FilterCapacity` hard cap is still sufficient.
- `parentSpendsCapacity`'s `capacity/multiplier != entryCount` guard is correct, and `multiplier` is normalised to ≥1 before the division.
- `txMapConstructorWrapThreshold = 3579139413` is exact: `3579139413 + /5 = 4294967295` (= MaxUint32), `+1` wraps to 0. Matches go-tx-map v1.4.1 `tx_map.go:1064`.
- `parentSpendsAllocHint`'s `1<<32` bound keeps `uint32((n + n/5) / 8192)` exact at 629,145.
- No other allocation in `Block.go` is sized from `b.TransactionCount`.
- `gofmt` and `go vet` clean; `go test -race -count=1 ./model/` → `ok  github.com/bsv-blockchain/teranode/model  12.470s`.

The rationale added to `PutTxMap` for dropping over-max maps — they've been *filled* past the class they were sized for, so pooling one would retain an oversized backing for every later largest-class block — is a better answer than the reuse I suggested. Withdrawn.

Non-blocking, take or leave:

- `block_txcount_64bit_test.go:60-84` — the comment says a go-tx-map bump that changes the headroom factor makes the first loop fail, but `/5` is hardcoded in the test too, so the loop only checks the constant against its own transcription. Under a `/3` factor the true threshold drops to ~3.22e9, the constant would be wrong, and both loops still pass. Low impact (the real clamp is `1<<30`, safe under either factor — only the documented constant goes stale), but it's the same shape as the `releaseTxMap` comment from last round, which got fixed by being made honest. Either soften it, or do what `TestGetParentSpendsMapBoundsPreallocation` already does and compare against real constructor output: `txmap.NewSplitSwissMapUint64(h, txMapBuckets).Map()[0].Map().Capacity()` for a small `h` against the transcribed `(h + h/5)/8192`. The threshold itself can't be constructed cheaply, so the low-`h` factor check is the tractable version.
- `Block.go:1166` — `validOrderAndBlessed` recomputes `txMapEntryCount()` rather than reading the stashed `b.txMapCount`. Correct for the direct callers that skip step 11, but worth a sentence so nobody later "tidies" it.
- The two tests that mutate `txMapSizeClasses` / `parentSpendsSizeClasses` are safe as written (`t.Cleanup`, no `t.Parallel`), but they make global mutable state load-bearing for test correctness while the pools were built from the original tables at init. A test-only seam would remove the hazard.

freemans13:
Thanks @ordishs — and thanks for re-deriving rather than taking the fixes on trust. The `Subtree.Length()` == `len(st.Nodes)` check is the one that matters most: "numerically identical where the count was already reconciled, strictly safer everywhere else" is exactly the property I was relying on, and I had verified it only from the accumulation side, not from go-subtree's definition.

All three non-blocking notes taken, in 688c1cfce.

**The headroom-factor point was a genuine hit, and it's the third time in this PR that a comment claimed a guarantee the code didn't provide** — same shape as the `releaseTxMap` premise and the trade-off bullet in the body. You're right that `/5` was transcribed in the test too, so the loops only proved the transcription was self-consistent.

Taken your second option rather than softening it. The factor is now observed from real constructor output: the threshold itself can't be probed (tens of GB), but with a single bucket the per-bucket size is the whole `(hint + hint/5)`, so a fresh map's `Capacity()` exposes the factor for a few hundred KB. Mutation-proved by simulating a `/3` factor — fails with `expected: 1337, actual: 1204`. The comment now says plainly that the first two checks would be self-referential alone and that the third is what makes them mean something.

**`Block.go` recompute** — documented, with the reason: `b.txMapCount` is only set by `checkDuplicateTransactions`, so a caller reaching `validOrderAndBlessed` directly (rather than through `Valid`, which always runs step 11 first) would size from a zero. Explicitly asks not to be tidied into reading the field.

**Global-state seam** — the swaps now go through `withTxMapSizeClasses` / `withParentSpendsSizeClasses`, so the mandatory restore lives in one place instead of being re-remembered per test, along with why shrinking is safe against pools built at init. I kept it as a test helper rather than a production seam: injecting the tables into `GetTxMap`/`GetParentSpendsMap` would put a parameter on a hot path purely for tests. Happy to go further if you'd prefer the real seam.

Verification on the current head: `go build ./...`, `go vet ./model/`, `go test ./model/`, `go test -race -count=1 ./model/` (21.0s), `golangci-lint --new-from-rev upstream/main` (0 issues), plus `./services/blockvalidation/`, `./services/subtreevalidation/`, `./services/blockpersister/...` — all green.

Note the two design questions from your previous review remain open and are recorded on #1501 rather than resolved here: the deliberate plausibility gate, and — the harder half — what a node should *do* with a block it cannot hold in memory, given that it must be non-retryable and not block-invalid. Neither is a blocker for this PR, but I didn't want the approval to imply they were settled.

PR #1481, opened 2026-08-05
fix(model): evaluate median-time-past against the 11 headers ending at the block's parent, not the oldest of the fetched run

freemans13:
## What happened

Every block's timestamp must be strictly greater than the median timestamp of the previous 11 blocks — the 11 ending at the block's parent. Teranode's block-validation path evaluated that median against the wrong blocks. The validator fetches up to 100 previous headers newest-first (the `GetBlockHeaders` order, with the parent at index 0) and passes them to `CheckHeaderContextual`, which took the *tail* 11 of the run — the oldest headers, roughly 90 to 100 blocks below the parent. On mainnet that median sits about 15 hours behind the correct one, so the check accepted block timestamps far into the past that SV Node (and any correct implementation) rejects.

This is a silent consensus divergence: honestly-stamped blocks satisfy both the wrong window and the right one, so nothing ever surfaced in normal operation. It only bites when someone produces a block in the gap — at which point Teranode nodes accept a block the rest of the network refuses. Past-stamped blocks are also the classic difficulty-manipulation lever, which is exactly what the median-time-past rule exists to bound. Both the synchronous receipt-time check and the full background `block.Valid` used the same wrong window, so every path that evaluates the rule was affected. (Correction, from ctnguyen's review: an earlier version of this description also claimed catchup's cached headers reached the check. They do not, before or after this PR — every caller passing `CachedHeaders` also sets `DisableOptimisticMining`, and the non-optimistic path re-fetches from the store. See #1499.)

Found during the adversarially-verified review of PR #1466 and filed as #1467; two independent review agents re-derived it from the code, which is what prompted this fix.

Closes #1467.

## Changes

- **`model/Block.go` — `CheckHeaderContextual` anchors the median window on the block's parent.** The window is now the 11 headers adjacent to the parent, found at whichever end of the supplied run the parent sits (newest-first and oldest-first both work; short runs near genesis use what exists, as before). A run with the parent at *neither* end is refused with a processing error rather than silently measured: that situation means the caller supplied somebody else's chain, and it is a context error on the caller's side, not evidence the block is bad — so the block is not marked invalid for it. That distinction matters because the block-validation service persists invalid verdicts.
[...]

PR #1466, opened 2026-08-05
fix(blockassembly): floor the mining-candidate timestamp at median-time-past+1 so block assembly never hands miners a block its own rule rejects

freemans13:
## What happened

Every block must carry a timestamp strictly greater than the median timestamp of the previous 11 blocks (the "median time past" rule). Block assembly, however, stamped mining candidates with the raw local wall clock — `time.Now()` with no floor. If the node's clock lags, or if recent blocks were stamped into the future (allowed up to 2 hours) and dragged the median above local time, the assembler hands miners a candidate at or below the median. The miner solves and submits it, and the network refuses the block. The damage is not caught locally: the submit path calls `block.Valid` with no header chain, which skips the median-time rule, so the bad block is accepted by the producing node and rejected by every peer that validates it. Because a whole fleet typically shares one NTP source, a skew event hits every assembler at once, and the fleet quietly mines blocks the rest of the network throws away until wall-clock time catches back up.

SV Node avoids this by construction: its `UpdateTime` sets a candidate's time to `max(median-time-past + 1, adjusted-now)`. This change brings Teranode's two candidate paths (the normal subtree path and the empty-block path) up to the same guarantee.

Closes #1445.

## Changes

- **New `candidateTime` helper (`services/blockassembly/candidate_time.go`)** — computes the wall clock floored at median-time-past+1 of the candidate's parent chain, using the same `model.CalculateMedianTimestamp` the validator uses. Headers are fetched by parent hash (so a side-chain candidate gets the median of *its own* chain) in two steps: a batched `GetBlockHeaders` call whose result is verified to be anchored at the parent and correctly linked, with a fallback to a hash-keyed walk of the parent chain (one `GetBlockHeader` per header) when that verification fails. The fallback matters because the store's batched lookup ranges over main-chain flags and can race a reorg; the per-hash reads are immutable and immune. This is the same two-step strategy as the candidate-parent MTP helpers in `subtreevalidation` and `legacy/netsync`. Only when both attempts fail does the candidate request error, failing that one miner poll.
- **Both candidate paths now use it (`BlockAssembler.go`)** — `GetMiningCandidate`'s main path and `generateEmptyBlockCandidate` (which gains a `context.Context` parameter). The floored time also feeds `getNextNbits`, so the testnet 20-minute difficulty rule sees the same timestamp the candidate carries — the same ordering SV Node uses. When the floor engages, a warning is logged once per tip (not once per miner poll — pools poll several times a second).
[...]

PR #1485, opened 2026-08-05
fix(utxo): publish block height and median block time as one atomic snapshot

### LINKED ISSUE #1443: [HARDEN-0805] Read block height and median-block-time as one genuine atomic snapshot, or delete the false "atomic snapshot" comment
## What can go wrong

The finality decision (does a transaction's nLockTime allow it into a block/template yet) assumes the block height and the median-block-time it evaluates against both come from the *same* chain tip. The UTXO store's `GetBlockState` is supposed to return that pair as one consistent snapshot. It does not: it reads two independent atomic values one after the other. The two values are written by two separate setter calls driven by blockchain notifications. A reader that lands between the two writes sees a *new* height paired with a *stale* median time (or, on a reorg that lowers both, the reverse). The finality check can then evaluate a transaction using a height from block N and a median time from block N-1 in the same decision.

A doc-comment on the wrapper explicitly claims this returns "an atomic snapshot" that "prevents race conditions" — which is false for two separate atomic loads, and discourages anyone from testing it.

## Where (verified on `main` @ `27a0178fd`)

- `stores/utxo/aerospike/aerospike.go:706-711` — `GetBlockState` builds its result from two separate atomic loads: `Height: s.blockHeight.Load()` (line 708) and `MedianTime: s.medianBlockTime.Load()` (line 709). Nothing snapshots the two together. The two fields are updated by separate methods — `SetMedianBlockTime` (lines 695-700) stores only `medianBlockTime`, and block height is stored by its own separate setter — so a reader can observe one updated and the other not.
- `services/validator/Validator.go:370-372` — the misleading comment: `GetBlockState` "returns an atomic snapshot of both block height and median block time ... This prevents race conditions ...".
- Consumer: `services/validator/Validator.go:689` reads the pair once (`blockState := v.GetBlockState()`), then at lines 691-692 derives `blockHeight = blockState.Height + 1` when the caller passed height 0, and later feeds `blockState.MedianTime` into the finality comparison — so a single torn pair drives both the height and the median in one evaluation.
- **Blast radius is already reduced on `main`.** The consensus block-validation path is guarded: `selectFinalityComparisonTime` (lines 404-428) hard-errors (lines 423-424) when the post-CSV path lacks a caller-supplied `CandidateParentMedianTime`, rather than falling back to the racy `blockState.MedianTime`. So the worst case (a non-final tx admitted into the node's own invalid block) is closed for block validation. The torn read survives on the *policy/mempool intake* path only: `!SkipPolicyChecks` returns `blockState.MedianTime` directly (lines 407-411).

## Why it matters

This is an unenforced invariant with a comment that actively asserts it holds. The residual on `main` is narrow — a policy-admission edge case for a transaction sitting exactly on the finality boundary, not a consensus-invalid block — but it is still a wrong answer produced silently, and the false "atomic snapshot" comment means the next person who adds a consumer of this pair will reasonably trust it to be consistent and reintroduce the worse consequence. The window is nanoseconds between two stores, which is exactly why it will never surface in serial tests and needs a deliberate concurrency test.

## How to reproduce / test

A focused concurrency test on the store's block-state getters/setters (no external services needed if you unit-test the atomic pair directly, or use the Aerospike TestContainers store):

1. From one goroutine, drive the two setters with a known monotonic sequence of (height, medianTime) pairs — always advancing both together, e.g. (100, 1000), (101, 1010), (102, 1020), ...
2. From another goroutine, call `GetBlockState` in a tight loop and record every pair returned.
3. Assert every observed pair is one of the *emitted* pairs — never a cross like (101, 1000) that pairs a new height with an old median.

With today's two-load getter this catches a crossed pair. It passes once the read is made atomic.

## Suggested fix

Make the pair genuinely atomic. Either pack height and median-time into a single value updated as a unit (e.g. one `atomic.Uint64` holding both 32-bit fields, or an `atomic.Pointer[BlockState]` swapped on each update), or guard both the setters and `GetBlockState` with a small mutex so a reader always sees a matched pair. Update the setter side so height and median-time advance together under the same protection. And correct the `Validator.go:370-372` comment so it describes the actual guarantee rather than claiming an atomicity the code does not currently provide.

## Confidence

**Confirmed.** The two-load getter, the false comment, the torn-pair consumer, and the block-validation hard-error guard are all read directly in the code above.

---
_Filed from the 2026-08-05 production-hardening forward-looking hazard review (RACE-2). Evidence verified against `bsv-blockchain/teranode@27a0178fd`. Session tag: `hardening-map-2026-08-05`._

---
*Part of the `[HARDEN-0805]` production-hardening sweep (label `hardening-map-2026-08-05`), grounded in `main` @ `27a0178fd`. Priority: Low · Size: S · This task is self-contained and does not depend on any other ticket.*

---

## The median-time-past family — who owns what

Nine tickets touch the same consensus quantity: the median timestamp of the eleven blocks ending at a given block. Most were raised out of the review of #1466, at different times, and they are close enough to each other that starting one without reading the set is how two people end up fixing the same line. **This map is duplicated in each of the nine; if you change the split, change it everywhere.**

| Ticket | Stage of the pipeline | What it owns |
|---|---|---|
| #1445 (PR #1466) | Producing blocks | Floor the mining-candidate timestamp at median-time-past+1 |
| #1467 (PR #1481) | Validating blocks | Measure the median over the eleven headers ending at the block's parent |
| #1443 (PR #1485) | Judging transactions | Publish block height and median block time as one atomic snapshot — **this ticket** |
| #1505 | Producing blocks | Apply the median rule on the submit path — `block.Valid` is called with a nil `currentChain`, so the rule is structurally skipped |
| #1506 | Producing blocks | Floor against network-adjusted time rather than the raw local clock, so skew self-corrects instead of stopping mining |
| #1528 | Producer and validator | Declare the two-hour future bound once, in `settings` |
| #1529 | Producing blocks | Memoize floor-lookup failures, not only successes |
| #1541 | All of them | Consolidate `model.CalculateMedianTimestamp`'s seven call sites, which disagree about anchoring and about short runs |
| #1542 | Producing blocks | The candidate-time single-flight runs on its leader's context, so a cancelled leader costs its followers a floor for one poll |

None of these impose a merge order on the others and they share almost no files, so the sweep boilerplate calling a ticket "self-contained" still holds. The overlap is in subject matter: they reason about the same eleven-block median, so a change made without reading the set is easy to get subtly wrong or to duplicate.

#1541 is the exception worth calling out: it touches every site the other eight reason about, so it is the one most likely to collide with work in progress. Read it before starting any of the others, and land it either well before or well after them.

Two things worth knowing before touching any of these.

**The windows coincide by construction, not by accident.** Once #1445, #1467 and #1443 have all landed, every path measures the median over the same eleven blocks — the chain tip and its ten ancestors. Block validation's window ends at the block's parent, which for the next block on the tip is the tip itself; block assembly computes its floor from the candidate's parent chain, the same set; and the store's median comes from `GetBestHeightAndTime`, which fetches eleven headers inclusive of the best header. A future change that "aligns" one of these to another would be moving it off the correct rule, not onto it.

**Producer-side and validator-side fixes are not substitutes for each other.** #1445 stops this node handing miners a block its own rule rejects. #1467 stops this node accepting a block the rest of the network rejects. With only one of them, a correctly floored candidate still sails through a validation window too lenient to notice anything, or validation is right while assembly still produces candidates that fail it.

### LINKED ISSUE #1445: [HARDEN-0805] Floor the mining-candidate timestamp at median-time-past+1 so block assembly never hands miners a block its own rule rejects
## What can go wrong

Every block's timestamp must be strictly greater than the median timestamp of the previous 11 blocks (the BIP113 "median time past" rule). SV Node computes a candidate block's time as `max(median-time-past + 1, adjusted-now)` precisely to guarantee this. Teranode's block-assembly candidate builder instead sets the candidate's time to the raw local wall-clock (`time.Now()`) with no such floor.

If the node's clock lags, or if recent blocks were stamped slightly into the future (allowed up to the 2-hour future window) and dragged the median above the local clock, the candidate's timestamp comes out at or below the median-time-past. A miner then solves and submits that block — and it fails the median-time check that Teranode itself enforces downstream, so the node produces a block it (and the rest of the network) rejects. Any delay between building the candidate and submitting the solution only makes it worse, because the fixed timestamp falls further behind an advancing median.

## Where (verified on `main` @ `27a0178fd`)

- `services/blockassembly/BlockAssembler.go:1517` — `timeNow := time.Now().Unix()`, converted at line 1518, then assigned straight to `candidate.Time` at line 1548. No comparison against the previous blocks' median time.
- The empty-block path has the same gap: `generateEmptyBlockCandidate` sets `timeNow := time.Now().Unix()` (line 1593) and assigns it to the candidate's `Time` (line 1620) with no floor.
- The rule the candidate can violate is enforced only later, in `model/Block.go`: the 2-hour-future check at line 525, and the median-time-past check at line 558 — `if b.Header.Timestamp <= b.medianTimestamp { ... return ...InvalidError }`. So the candidate builder can emit a time the validator will reject.

## Why it matters

This is an invariant ("a candidate's timestamp is above the median-time-past") that the builder assumes but never enforces, while the same codebase enforces it strictly at validation. It stays invisible because clocks are normally NTP-synced and the median tracks real time, so `now > median` almost always. But the whole fleet typically shares an NTP source, so a skew event (or a run of forward-stamped blocks near the 2-hour edge) hits every block-assembly node at once — and valid block production could stall fleet-wide until wall-clock time catches back up to the median, with the only symptom being self-rejected blocks.

## How to reproduce / test

1. Build a chain whose last 11 blocks carry timestamps giving a median-time-past above a controlled "now" (either advance the block timestamps forward within the allowed future window, or drive the assembler with a clock set behind the median).
2. Call `GetMiningCandidate` (and the empty-block path).
3. Assert the candidate's `Time` is strictly greater than the median-time-past of the previous 11 blocks. Today it can come back at or below the median.

## Suggested fix

Floor the candidate timestamp exactly as SV Node does: compute the median-time-past of the previous 11 blocks and set the candidate time to `max(medianTimePast + 1, time.Now())`. Apply it in both the main candidate path (around line 1517/1548) and the empty-block path (around line 1593/1620). The median is already computed at validation time (`model/Block.go`), so the same helper can be reused on the assembly side.

## Confidence

**Strong hypothesis.** The raw-clock assignment in both candidate paths and the downstream median-time enforcement are read directly above. The part that needs a spike is confirming the skew/forward-stamp conditions that push `now <= median` are actually reachable in production.

---
_Filed from the 2026-08-05 production-hardening forward-looking hazard review (RACE-4). Evidence verified against `bsv-blockchain/teranode@27a0178fd`. Session tag: `hardening-map-2026-08-05`._

---

---

## The median-time-past family — who owns what

Nine tickets touch the same consensus quantity: the median timestamp of the eleven blocks ending at a given block. Most were raised out of the review of #1466, at different times, and they are close enough to each other that starting one without reading the set is how two people end up fixing the same line. **This map is duplicated in each of the nine; if you change the split, change it everywhere.**

| Ticket | Stage of the pipeline | What it owns |
|---|---|---|
| #1445 (PR #1466) | Producing blocks | Floor the mining-candidate timestamp at median-time-past+1 — **this ticket** |
| #1467 (PR #1481) | Validating blocks | Measure the median over the eleven headers ending at the block's parent |
| #1443 (PR #1485) | Judging transactions | Publish block height and median block time as one atomic snapshot |
| #1505 | Producing blocks | Apply the median rule on the submit path — `block.Valid` is called with a nil `currentChain`, so the rule is structurally skipped |
| #1506 | Producing blocks | Floor against network-adjusted time rather than the raw local clock, so skew self-corrects instead of stopping mining |
| #1528 | Producer and validator | Declare the two-hour future bound once, in `settings` |
| #1529 | Producing blocks | Memoize floor-lookup failures, not only successes |
| #1541 | All of them | Consolidate `model.CalculateMedianTimestamp`'s seven call sites, which disagree about anchoring and about short runs |
| #1542 | Producing blocks | The candidate-time single-flight runs on its leader's context, so a cancelled leader costs its followers a floor for one poll |

None of these impose a merge order on the others and they share almost no files, so the sweep boilerplate calling a ticket "self-contained" still holds. The overlap is in subject matter: they reason about the same eleven-block median, so a change made without reading the set is easy to get subtly wrong or to duplicate.

#1541 is the exception worth calling out: it touches every site the other eight reason about, so it is the one most likely to collide with work in progress. Read it before starting any of the others, and land it either well before or well after them.

Two things worth knowing before touching any of these.

**The windows coincide by construction, not by accident.** Once #1445, #1467 and #1443 have all landed, every path measures the median over the same eleven blocks — the chain tip and its ten ancestors. Block validation's window ends at the block's parent, which for the next block on the tip is the tip itself; block assembly computes its floor from the candidate's parent chain, the same set; and the store's median comes from `GetBestHeightAndTime`, which fetches eleven headers inclusive of the best header. A future change that "aligns" one of these to another would be moving it off the correct rule, not onto it.

**Producer-side and validator-side fixes are not substitutes for each other.** #1445 stops this node handing miners a block its own rule rejects. #1467 stops this node accepting a block the rest of the network rejects. With only one of them, a correctly floored candidate still sails through a validation window too lenient to notice anything, or validation is right while assembly still produces candidates that fail it.

### LINKED ISSUE #1467: [HARDEN-0805] Median-time-past is checked against the oldest 11 of 100 fetched headers, not the 11 ending at the parent
## What can go wrong

The median-time-past rule says a block's timestamp must be strictly greater than the median timestamp of the **previous 11 blocks** — the 11 ending at the block's parent. Teranode's block-validation path evaluates that median against the wrong 11 blocks: the oldest of the 100 headers it fetches, roughly 90–100 blocks below the parent. On mainnet that median is ~15 hours older than the correct one, so the check accepts timestamps far into the past that SV Node (and any correct implementation) rejects. A miner can stamp a block up to ~15 hours behind the true median-time-past and Teranode will accept it while the rest of the network refuses it — a consensus divergence and chain-split risk in the direction that is invisible in normal operation, because honestly-stamped blocks satisfy both windows.

## Where (verified on `main` @ `97adafe66`)

Three legs, each verified directly:

1. `services/blockvalidation/BlockValidation.go:1619` — `blockHeaders, _, err = u.blockchainClient.GetBlockHeaders(ctx, block.Header.HashPrevBlock, previousBlockHeaderCount)` fetches `blockvalidation_previous_block_header_count` headers (default 100, `settings/settings.go:362`).
2. `GetBlockHeaders` returns headers **newest-first**: element 0 is the requested parent hash and heights descend from there (the SQL store orders by height descending; the anchor checks in `services/subtreevalidation/check_block_subtrees.go:1556` and PR 1466's `candidateTime` both depend on element 0 being the requested hash).
3. `model/Block.go:540` — `CheckHeaderContextual` documents `currentChain` as "the run of previous headers ending at the block's parent" (i.e. ascending, parent last) and takes the **tail**: `lastTimeStamps := currentChain[currentChainLength-pruneLength:]`. With a newest-first slice of 100, the tail is the *oldest* 11 headers — heights parent−89 … parent−99 — not the 11 ending at the parent.

The same newest-first slice is also passed to the background `block.Valid(...)` call (`BlockValidation.go:1812`), which runs the same `CheckHeaderContextual`, so both the synchronous receipt-time check and the full validation use the wrong window. The 2-hour-future check is unaffected (it doesn't use `currentChain`).

Note the check is *correct* in the one configuration where at most 11 headers are supplied, since the tail is then the whole (order-insensitive) set — which is why unit tests that hand `CheckHeaderContextual` exactly 11 headers pass regardless of order and never caught this.

## Why it matters

- The enforced lower bound on block timestamps is ~15 hours looser than consensus. Combined with the timestamp-based difficulty rules, past-stamped blocks are the classic difficulty-manipulation lever; the median-time-past rule exists precisely to bound it.
- It is a silent divergence from SV Node: no honest block ever trips either window, so nothing surfaces in production or in differential operation until someone produces a block in the gap — at which point Teranode nodes follow a chain the rest of the network rejects (or vice versa).

## How to reproduce / test

Unit test: build 100 linked headers with strictly increasing timestamps, pass them to `CheckHeaderContextual` newest-first (exactly what `ValidateBlock` does), and give the block a timestamp above the median of the oldest 11 but below the median of the newest 11. The check passes today; a correct implementation must reject it.

## Suggested fix

In `ValidateBlock`, order (or re-slice) the headers so the median window is the 11 headers ending at the block's parent — either reverse the `GetBlockHeaders` result before passing it as `currentChain`, or change `CheckHeaderContextual` to select the 11 headers adjacent to the parent explicitly (and assert the parent is actually in the run, which would also have caught this). Add the reproduction test above, and audit the two other `CheckHeaderContextual` call sites for the same ordering assumption.

## Confidence

Confirmed on all three legs by direct code reading; found during the adversarially-verified review of PR 1466 (issue 1445), where the reviewer's coverage/mutation analysis surfaced it as a pre-existing defect. Not introduced or fixed by that PR.

---

## The median-time-past family — who owns what

Nine tickets touch the same consensus quantity: the median timestamp of the eleven blocks ending at a given block. Most were raised out of the review of #1466, at different times, and they are close enough to each other that starting one without reading the set is how two people end up fixing the same line. **This map is duplicated in each of the nine; if you change the split, change it everywhere.**

| Ticket | Stage of the pipeline | What it owns |
|---|---|---|
| #1445 (PR #1466) | Producing blocks | Floor the mining-candidate timestamp at median-time-past+1 |
| #1467 (PR #1481) | Validating blocks | Measure the median over the eleven headers ending at the block's parent — **this ticket** |
| #1443 (PR #1485) | Judging transactions | Publish block height and median block time as one atomic snapshot |
| #1505 | Producing blocks | Apply the median rule on the submit path — `block.Valid` is called with a nil `currentChain`, so the rule is structurally skipped |
| #1506 | Producing blocks | Floor against network-adjusted time rather than the raw local clock, so skew self-corrects instead of stopping mining |
| #1528 | Producer and validator | Declare the two-hour future bound once, in `settings` |
| #1529 | Producing blocks | Memoize floor-lookup failures, not only successes |
| #1541 | All of them | Consolidate `model.CalculateMedianTimestamp`'s seven call sites, which disagree about anchoring and about short runs |
| #1542 | Producing blocks | The candidate-time single-flight runs on its leader's context, so a cancelled leader costs its followers a floor for one poll |

None of these impose a merge order on the others and they share almost no files, so the sweep boilerplate calling a ticket "self-contained" still holds. The overlap is in subject matter: they reason about the same eleven-block median, so a change made without reading the set is easy to get subtly wrong or to duplicate.

#1541 is the exception worth calling out: it touches every site the other eight reason about, so it is the one most likely to collide with work in progress. Read it before starting any of the others, and land it either well before or well after them.

Two things worth knowing before touching any of these.

**The windows coincide by construction, not by accident.** Once #1445, #1467 and #1443 have all landed, every path measures the median over the same eleven blocks — the chain tip and its ten ancestors. Block validation's window ends at the block's parent, which for the next block on the tip is the tip itself; block assembly computes its floor from the candidate's parent chain, the same set; and the store's median comes from `GetBestHeightAndTime`, which fetches eleven headers inclusive of the best header. A future change that "aligns" one of these to another would be moving it off the correct rule, not onto it.

**Producer-side and validator-side fixes are not substitutes for each other.** #1445 stops this node handing miners a block its own rule rejects. #1467 stops this node accepting a block the rest of the network rejects. With only one of them, a correctly floored candidate still sails through a validation window too lenient to notice anything, or validation is right while assembly still produces candidates that fail it.

freemans13:
## What happened

The UTXO store's `GetBlockState` returns the pair that drives an nLockTime finality decision: the current block height and the median block time. Its doc comment described "an atomic snapshot" that "prevents race conditions", but every store built the result from two independent atomic loads, and the two values were written by two separate setter calls fired one after the other from blockchain notifications. A reader landing between the two writes could get a new height paired with a stale median time, or the reverse during a reorg.

The consensus block-validation path is unaffected — it hard-errors unless the caller supplies its own median — so a torn pair only reaches the policy/mempool intake path.

Closes #1443.

### How much this matters, stated plainly

A fair question on a change like this is whether anything was actually going wrong, and the honest answer is that the exposure is narrower than the issue title suggests.

The two fields are not read in the same branch of `selectFinalityComparisonTime`: the policy branch uses the median time alone, and the height appears only in an error string. They are used jointly a level up, across `validateInternal`, where the height drives era, checkpoint and DAH handling while the median time drives finality. So the torn pair does reach one transaction's evaluation, but not the single comparison.

The torn window is also a subset of a wider staleness that this PR does not address: the notification listener updates the store after the tip has already moved, so the pair can be one tip behind regardless. Closing the tear does not make the pair current, only self-consistent.

That makes this a correctness tidy-up rather than a fix for an observed incident. The fix is cheap and the invariant is easy to state, and the more concrete gain is that the interface comment no longer promises a guarantee the code did not provide.

## Changes

- **`stores/utxo/block_state.go` (new)** — a small `BlockStateHolder` keeps the pair behind one atomic pointer. `SetPair` publishes both values as a single snapshot; `SetHeight`/`SetMedianTime` exist for callers that legitimately update one field, carrying the other forward with a compare-and-swap loop so concurrent single-field writers do not erase each other.
- **`stores/utxo/Interface.go`** — a new `SetBlockState(height, medianTime)` method is the write side of the guarantee: callers holding both values for the same tip publish them together. The `GetBlockState` doc now states what is actually guaranteed — the read is one atomic load, and the pair is as consistent as its writer. `SetBlockState` documents its height-zero precondition.
- **Store implementations** (aerospike, sql, nullstore, mock, plus the logger and txmetacache wrappers) — the sql, aerospike and null stores embed `utxo.BlockStateFields`, which holds the pair and supplies all six block-state methods over one holder, so every store shares one implementation rather than keeping its own copy. Each store had also carried a second, independent height atomic beside the holder; that is gone, and the DAH, maturity, catchup and alert paths that read it directly now go through `GetBlockHeight`. The stores keep thin wrappers only where they need more than the state change: a debug line on the setters, and aerospike's mirroring of the height into its external blob store for that store's retention bookkeeping.
- **`stores/utxo/factory/utxo.go`** — the blockchain-notification listener, the only production writer of this pair, already fetches height and median time together from `GetBestHeightAndTime`; it now publishes them with one `SetBlockState` call at both the bootstrap and per-notification sites.
- **`services/validator/Validator.go`** — the comment now describes the real mechanism and its limits.
- **Tests** — `stores/utxo/block_state_test.go` covers the holder in isolation. `stores/utxo/tests/tests.go` adds shared cases that run against real stores: `SetBlockStateContract` for the write side and `SetBlockStateSnapshotUnderConcurrency` for the read side, both now also wired into the null store.

## Why the fix holds end to end

Publishing the pair atomically in the store would not help if the pair arrived already torn, so that was checked too: `GetBestHeightAndTime` reads the best block header once and then anchors its 11-header median window on that same header's hash (`services/blockchain/Server.go:3091` and `:3097`). Height and median time therefore do not come from two different tips before they reach the store either.

## Known limits

- **The rewind CLI still moves height alone.** `cmd/rewindblockchain/rewindblockchain/rewind.go:85` calls `SetBlockHeight` on its own, leaving that process's in-memory pair as a rewound height beside a stale median time. `SetHeight` carries the old median forward rather than zeroing it, and that process never validates transactions and holds its own store instance, so nothing consumes the inconsistent pair. It is worth knowing about rather than mistaking later for a gap in this fix.
- **The individual setters still exist**, for tests and height-only bootstraps, so a pair assembled through two separate single-field calls is only as consistent as the caller's sequencing. That is now documented at the interface. Production uses the combined setter.
- **The duplicate height is gone, and with it one limit this section used to record.** Each store used to carry the height twice — the holder plus a pre-existing single-field atomic that the DAH, catchup and alert paths read directly — kept in step by single-writer discipline rather than by a lock. That framing turned out to be too generous to the code: because `GetBlockHeight` read the atomic while `GetBlockState().Height` read the holder, a *reader* calling both back to back could see them disagree, and no amount of writer discipline closes a window that sits between two reads. Consolidating every reader onto the holder removes it, which is what `BlockStateFields` now does. The cost is one extra pointer hop on the paths that read the cached height; the snapshot is allocated once per block notification and stays cache-resident.
- **The concurrency cases guard the regression with more than one processor, not on one core.** On a single processor a reader can only interleave with the writer where the scheduler preempts it, so no test of this shape reliably observes a tear at `-cpu=1`. The cases are meaningful there — they gate on readers having actually sampled, so they neither pass vacuously nor fail on correct code — but the regression guard itself depends on CI running with `GOMAXPROCS > 1`. This is noted on the test.

## Testing

Three review rounds ran over this branch, and each found something real.

The first found that the new production write path had no test on either real store. `SetBlockState` carries the obligation of keeping the pre-existing single-field atomics in step, since DAH and maturity logic read them directly, so a refactor dropping those stores would have frozen DAH heights at process start with the suite green. `SetBlockStateContract` now pins that.

The second found four more. The factory test still primed the two setters the listener no longer calls and had nothing primed for `SetBlockState`. `NullStore` backed the pair with two plain unsynchronised fields, so the interface's promise was not true of every implementation. The holder's concurrency test called `require.Equal` from inside its reader goroutines, where testify's `FailNow` is not valid. And the shared contract case left a suite-wide store at height 4242 for the aerospike subtests that follow it, several of which derive their height from the store's current one — it now restores prior state and runs last.

The second round also noted that nothing tested the read side on a real store: torn pairs only appear while a writer is mid-update, so a store whose `GetBlockState` went back to two independent loads would have kept every test green. `SetBlockStateSnapshotUnderConcurrency` closes that.

The third round, from @ordishs and @ctnguyen, found that the concurrency cases could still pass without a reader ever sampling — the writer loop ran a fixed number of publications, and if the readers were never scheduled during that window the torn-pair assertion held vacuously. That is the same shape of defect this branch exists to fix, so it was worth getting right. Counting samples and asserting the total afterwards turned out not to be enough: on correct code that fails at `-cpu=1`, because the readers genuinely do not run during a fixed-length writer loop. The writer now keeps publishing until the readers have sampled, with a safety deadline and an early break once a tear is found.

That was verified against a mutation reverting the SQL store's `GetBlockState` to two independent atomic loads — the original defect:

| | clean code | mutation reinstated |
|---|---|---|
| default `GOMAXPROCS` | passes 3/3 | fails 3/3 in 0.02s, torn pairs one tip apart |
| `-cpu=1` | passes 3/3 | passes |

The bottom-right cell is the limit recorded above. The same round also found that the reference doc still carried the wording the interface had dropped and omitted `SetBlockState` entirely, and that `NullStore` accepted a height of zero where the other stores reject it — it was the one store the shared suite never ran against, which is how that survived. Both are now fixed, and the null store is wired into the shared cases.

A fourth round followed the two approvals, prompted by the one check still red: SonarCloud's quality gate, which failed on duplication rather than on anything functional. Coverage, bugs, vulnerabilities and security hotspots all passed; what tripped it was 37 of 381 new lines duplicated, 9.7% against a 3% threshold, flagged as `aerospike.go:694-721` against `sql.go:290-316` — the block-state trio, near-identical between the two stores down to a doc comment that differed by one clause.

That was a real duplication rather than a metric artefact, and the honest fix was the one this description had previously ruled out as too wide: sharing a single implementation. Doing it turned out to be roughly size-neutral (153 insertions against 144 deletions) and it closed a defect at the same time — the reader-side height disagreement described under Known limits. It also surfaced a small wrong-log bug: the setters logged before validating, so a rejected zero height emitted a debug line claiming it had been set. The wrappers now validate first.

ChiR7 is also fixed: `TestNewStore_BlockchainClientError` primed a `SetBlockState` expectation on a path its own setup makes unreachable, since the test fails blockchain-client creation and `NewStore` returns before the listener starts. The line is dropped.

Ran green: `stores/utxo` and its `aerospike`, `aerospike/pruner`, `sql`, `sql/pruner`, `factory`, `logger`, `nullstore` packages, plus `stores/txmetacache`, `services/validator`, `services/blockpersister`, and the packages that use the null store — `services/subtreevalidation`, `services/legacy/netsync`, `services/blockvalidation` and `model`. `-race` on the holder tests and the store-level cases. `go build ./...` and `go vet` clean including under the `aerospike` build tag, and `golangci-lint` clean on the changed files.

## Related work and review scope: the median-time-past family

Three open PRs touch the same consensus quantity — the median timestamp of the eleven blocks ending at a given block — at three different points in the pipeline. They share no files and impose no merge order on each other, but they only add up to a coherent median-time story together:

- **#1466 — producing blocks.** Block assembly stamped mining candidates with the raw local wall clock, so a lagging clock, or recent blocks stamped into the future dragging the median up, could hand miners a candidate at or below the median — a block the network then refuses. The candidate time is now floored at median+1. Closes #1445.
- **#1481 — validating blocks.** `CheckHeaderContextual` measured the median over the wrong eleven headers: the oldest of the roughly one hundred it fetches, some ninety blocks below the parent instead of the eleven ending at it. On mainnet that median sits about fifteen hours behind the correct one, so Teranode accepted past-stamped blocks that SV Node rejects. The window is now anchored on the block's parent. Closes #1467.
- **#1485 (this PR) — judging transactions.** The UTXO store published the median block time and the block height as two separate writes, so the pair a transaction's evaluation reads could mix values from two different chain tips. The pair is now one atomic snapshot. Closes #1443.

It would help if each PR were reviewed against its own scope, since the three were kept as separate minimal diffs deliberately — something that looks missing here is often another PR's subject.

- **In scope here:** how the height/median-time pair is published and read inside the UTXO store, and the interface contract around it. This PR changes only how an already-computed pair is stored and handed out.
- **Belongs to #1481:** whether the median value itself is measured over the right eleven blocks.
- **Belongs to #1466:** whether block assembly floors its candidate timestamp.
- **Probably not a defect, though happy to be shown otherwise:** `GetBestHeightAndTime` anchors its eleven-header window on the chain tip rather than on a parent. Judging whether an incoming transaction is final asks about the next block, whose parent is the tip, so the tip-anchored window looks like the right one here and aligning it to block validation's parent-anchored window would change its meaning. Both implementations agree (`services/blockchain/Server.go:3091-3097` and `services/blockchain/LocalClient.go:521-528`).

**Cross-cutting, and owned by none of the three.** `model.CalculateMedianTimestamp` already has six call sites on `main` — `model/Block.go:548`, `services/blockchain/Server.go:3107`, `services/blockchain/LocalClient.go:538`, `services/blockchain/median_time_past.go:37`, `services/rpc/handlers.go:1596` and `stores/blockchain/sql/StoreBlock.go:1035` — each doing its own header fetch and anchor handling, and #1466 adds a seventh. Consolidating them behind one helper is a fair suggestion, but it is a pre-existing condition and larger than any of these diffs, so it may be better as its own issue than folded into one of these three reviews.

Two further things worth knowing if you are looking at more than one of these.

**The three windows coincide by construction.** After all three land, every path measures the median over the same eleven blocks — the chain tip and its ten ancestors. Block validation's window ends at the block's parent, which for the next block on the tip is the tip itself. Block assembly computes its floor from the candidate's parent chain, the same set. And the store's median comes from `GetBestHeightAndTime`, which fetches eleven headers inclusive of the best header. So a later change that aligned one of these to another would be moving it off the rule rather than onto it.

**Neither producer-side nor consumer-side alone is sufficient.** #1466 stops this node handing miners a block its own rule rejects; #1481 stops this node accepting a block the rest of the network rejects. With only #1466 merged, a correctly floored candidate still passes a validation window too lenient to notice anything. With only #1481 merged, validation is right but assembly can still produce candidates that fail it. #1485 is the transaction-level counterpart: the same median value, used to decide whether an individual transaction is final rather than whether a block is.

#1443 and #1445 came out of the same production-hardening sweep. #1467 is different — it was surfaced by the adversarial review of #1466, which is why that PR flags a pre-existing defect it deliberately did not bundle.


freemans13:
Review round 1 (concurrency, consensus, integration lenses — adversarially verified): 4 raw findings, 1 confirmed, 3 refuted.

The confirmed finding: `SetBlockState` — now the sole production writer of block state — had no test on a real store, and its easy-to-drop obligation (keeping the single-field atomics in step for the DAH/maturity readers) was unpinned; a refactor removing those stores would freeze DAH at process-start height with the suite green. Fixed in cccc695d4: a shared conformance case (`tests.SetBlockStateContract`) asserts the snapshot and both single-field getters agree after one call and that height zero is rejected, wired into the Docker-free sqlite smoke suite and the containerized aerospike suite.

Refutations worth recording: the nanosecond window where `GetBlockHeight` and `GetBlockState` disagree during a tip update is not a usable invariant loss (no decision combines the two APIs, and cross-call lockstep never existed on main either); the reader-goroutine `require` usage was empirically shown to fail cleanly (reconstructed old layout → 4 clear error records, no hang, no flood); and the 3-line duplication between `SetBlockState` and the individual setters is deliberate — the obvious dedup (calling the public setters) would republish a torn intermediate pair, reintroducing the exact bug.

Also fixed this round: a gofmt/gci formatting failure CI caught in the blockpersister test mock.

freemans13:
Review round 2 (closure audit + fresh eyes): zero findings — converged. The closure audit mutation-checked the exact hazard round 1 described (removing the single-field store from SetBlockState fails the new conformance case) and re-ran the affected suites green. Ready for human review.

freemans13:
### Reviewers: please read the scope note before filing findings

This is one of three open PRs on the same consensus quantity — the median block time. See **"Related work and review scope: the median-time-past family"** in the description above for the full picture.

This PR owns **only** how the height/median-time pair is published and read inside the UTXO store. It changes nothing about how the median value is computed — it takes an already-computed pair and stops it being torn.

Three things that look like findings here but are not, so please don't file them against this PR:

- Whether the median is measured over the right eleven blocks → **#1481**.
- Whether block assembly floors its candidate timestamp → **#1466**.
- `GetBestHeightAndTime` anchoring its eleven-header window on the chain **tip** rather than a parent. That is correct for this use: judging whether an incoming transaction is final asks about the next block, whose parent *is* the tip. It should not be "aligned" to block validation's parent-anchored window — both implementations agree already (`services/blockchain/Server.go:3091-3097`, `LocalClient.go:521-528`).

And one cross-cutting item owned by none of the three: `model.CalculateMedianTimestamp` has six call sites on `main` each doing its own header fetch. Consolidation is a fair suggestion but predates all three — please raise it as its own issue.

ordishs (approved):
Reviewed across four lenses (Go concurrency, BSV consensus/finality, security, code quality), each finding adversarially re-checked against current code. No P0 or P1 survived — the mechanism is correct and minimal.

Verified locally on the PR head:

- `go build ./...` → exit 0; `go vet -tags aerospike ./stores/... ./services/validator/... ./services/blockpersister/...` → exit 0
- `go test -race -run TestBlockStateHolder ./stores/utxo/` → ok 2.314s
- `go test -race -run 'Test_SmokeTests/set_block_state' ./stores/utxo/sql/` → ok 1.815s
- Reintroduced the original bug (reverted `sql.Store.GetBlockState` to two independent atomic loads) → `set_block_state_snapshot_under_concurrency` fails with four torn pairs, each off by exactly one tip. The new test pins what it claims to pin.

Two things worth taking, neither blocking:

**stores/utxo/tests/tests.go:1025-1058** (and the same shape at `stores/utxo/block_state_test.go:32-65`)
- Nothing counts reader iterations, so "no torn pairs" is indistinguishable from "no snapshots observed".
- Demonstrated, not hypothetical: with the pre-PR two-load `GetBlockState` reinstated, `go test -count=1 -cpu=1 -run 'Test_SmokeTests/set_block_state_snapshot_under_concurrency' ./stores/utxo/sql/` passes 3/3 — the writer's 20,000 in-memory publications finish before the readers are ever scheduled, and issue 1443 walks through the guard built to catch it. No CI exposure (nothing passes `-cpu`, runners are ≥4-core); it bites a developer running `-cpu=1` by hand.
- Fix: a shared `atomic.Uint64` incremented by each reader plus `require.Positive(t, reads.Load(), "readers never observed a snapshot")` after `wg.Wait()`. Two lines, in both tests.

**docs/references/stores/utxo_reference.md:258-273** (and the summary line at `:326`)
- The reference doc mirrors the `Store` interface verbatim: it omits `SetBlockState` and still carries the old `GetBlockState` comment — "prevents race conditions that could occur when reading these values separately" — which is the exact overclaim this branch exists to correct in the code. Copy the new wording across from `stores/utxo/Interface.go:591-604`.

Nits, in passing: `defer close(stop)` at `tests.go:1051` so a writer-loop failure can't leak four spinning readers; document the height-zero precondition on `SetBlockState` (`Interface.go:591-598`) since two stores enforce it and the conformance test asserts it; drop the dead `SetBlockState` mock expectation at `factory/utxo_test.go:355` (its test fails before the listener runs).

One clarification worth adding to the description: with `SkipPolicyChecks` set, `selectFinalityComparisonTime` (`services/validator/Validator.go:406`) never reads `blockState.MedianTime` — it either skips finality or hard-errors for a caller-supplied value. The torn pair could only ever have mis-decided policy/mempool admission, which is a narrower blast radius than the issue narrative implies and makes the fix easier to justify as low-risk.

ctnguyen (changes requested):
Review at commit [b43c42a](https://github.com/bsv-blockchain/teranode/commit/b43c42a7f6b974c4b7cd5732873729d088e5e997)

This is a tidy fix for a real problem, and the diagnosis is the good part: the old `GetBlockState` promised a snapshot while doing two independent loads, and the write side made it worse by publishing the pair as two back-to-back setter calls. Putting the pair behind one `atomic.Pointer` and giving the notification listener a single `SetBlockState` call is the right shape — it fixes both halves rather than just the read. I particularly liked that the CAS carry-forward in `SetHeight`/`SetMedianTime` was thought through instead of hand-waved, that the listener was the only production writer and was actually converted, and that the read path in `ValidateInternal` genuinely takes one snapshot and derives `blockHeight` from it, so the guarantee survives all the way to `selectFinalityComparisonTime`. The diff also stays small and doesn't drag the duplicate-atomics cleanup along with it. My findings are all about the promise's edges — the regression test that guards it, and the places where the documentation now claims more than the code delivers.

## Blocking issues

### ChiR1 — The new concurrency tests can pass without a reader ever observing a snapshot

*Same as [@ordishs' point "stores/utxo/tests/tests.go:1025-1058" (07-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1485#pullrequestreview-4883338433).* My finding extends theirs to the shape of the fix: a bare read counter turns a silent no-op into a single-core flake, so the writer loop has to be gated on the counter rather than just asserted against afterwards.

**Problem:** Neither `SetBlockStateSnapshotUnderConcurrency` (`stores/utxo/tests/tests.go:1001-1059`) nor `TestBlockStateHolderSnapshot` (`stores/utxo/block_state_test.go:17-66`) records how many snapshots the readers actually sampled. The writer loop is a bounded run of in-memory publications — 20,000 in the store case, 200,000 in the holder case — and every one of them completes in well under the 10 ms async-preemption quantum. If the reader goroutines are never scheduled during that window, `torn` is empty and `require.Empty` passes. @ordishs demonstrated exactly this: with the pre-PR two-load `GetBlockState` reinstated, `-cpu=1` passes 3/3.

**Why it matters:** These two cases are the entire reason commit b43c42a7 exists — the sequential `SetBlockStateContract` cannot see a tear, so this is the only guard against a store's `GetBlockState` regressing to two independent loads, which is issue [#1443](https://github.com/bsv-blockchain/teranode/issues/1443) itself. A guard that reports "no torn pairs" indistinguishably from "no pairs read" gives the next engineer a green suite while the bug is back. That is the same failure mode the PR is fixing one layer down: something asserting a property it never checks.

**Fix:** Count reads and make the writer keep publishing until the readers have actually sampled, so the assertion can't be vacuous *and* can't flake on a single-core runner:

```go
var reads atomic.Uint64
// in each reader, before the relation check:
reads.Add(1)

// writer: don't stop at the iteration count until readers have sampled
const minReads = 1_000
for i := baseHeight + 1; ; i++ {
    require.NoError(t, db.SetBlockState(i, i+offset))
    if i-baseHeight >= iterations && reads.Load() >= minReads {
        break
    }
}

close(stop)
wg.Wait()

require.GreaterOrEqual(t, reads.Load(), uint64(minReads), "readers never sampled the tear window")
require.Empty(t, torn, "GetBlockState returned torn pairs: %v", torn)
```

Same change in `block_state_test.go`. Worth re-running the mutation check (SQL `GetBlockState` reverted to two loads) at `-cpu=1` afterwards to confirm it now fails there too.

## Non-blocking issues

### ChiR2 — `SetBlockState`'s height-zero precondition is neither documented nor uniform across implementations

*Same as [@ordishs' point "document the height-zero precondition on SetBlockState" (07-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1485#pullrequestreview-4883338433).* Mine extends it past the doc gap: the precondition is also not implemented uniformly, so the new conformance helper would fail against one in-tree store.

**Problem:** `sql.Store.SetBlockState` (`stores/utxo/sql/sql.go:309-312`) and `aerospike.Store.SetBlockState` (`stores/utxo/aerospike/aerospike.go:712-715`) reject height 0 with `ErrInvalidArgument`; `NullStore.SetBlockState` (`stores/utxo/nullstore/nullstore.go:63-67`) accepts it silently, and the interface (`stores/utxo/Interface.go:591-598`) says nothing about it. So the conformance helper `SetBlockStateContract`, which asserts `require.ErrorIs(err, errors.ErrInvalidArgument)`, would fail if it were ever pointed at `NullStore` — and it is currently wired only into the sql and aerospike suites.

**Why it matters:** This method was just promoted to the interface as the one blessed way to publish the pair, so its contract is now load-bearing for every future implementer. A precondition that lives only in two of three implementations and in a test helper that never runs against the third is a contract by coincidence. `NullStore` was reworked in this very diff to make the interface's promise uniform, which makes the remaining asymmetry easy to close here.

**Fix:** State the precondition on the interface method and give `NullStore` the same guard, so `SetBlockStateContract` can be run against every implementation:

```go
	// SetBlockState publishes ... (issue 1443).
	//
	// Returns ErrInvalidArgument if height is zero, matching SetBlockHeight.
	// medianTime is unconstrained: zero is accepted, as with SetMedianBlockTime.
	SetBlockState(height, medianTime uint32) error
```

### ChiR3 — The rewritten `Validator.GetBlockState` comment ends on an unconditional claim

**Problem:** `services/validator/Validator.go:370-374` now reads: "the store keeps the pair behind a single atomic pointer written by SetBlockState, so a reader can never pair a new height with a stale median time **mid-read** (issue 1443) — both values in one finality decision come from the same chain tip." The first clause is correctly scoped by "mid-read" and is true. The trailing clause is not: it asserts same-tip provenance unconditionally, and the holder does not provide that. `SetBlockHeight` still feeds the holder via `BlockStateHolder.SetHeight`, which carries the *previous* median time forward (`stores/utxo/block_state.go:39-52`), so a pair published that way is a new height beside a stale median time — durably, not mid-read. `cmd/rewindblockchain/rewindblockchain/rewind.go:85` publishes exactly such a pair. The interface doc draws the line correctly — "the pair is only as consistent as its writer" (`stores/utxo/Interface.go:600-604`) — and the validator comment, on the line issue [#1443](https://github.com/bsv-blockchain/teranode/issues/1443) cited as the false one, omits it.

**Why it matters:** Narrow, but it is the one line in the diff where the old failure mode could re-form: the premise of this branch is that an unqualified comment let the next engineer skip a check, and same-tip provenance is a property of the *writer*, not of the atomic pointer this sentence attributes it to. The comment also asserts store internals from the consumer side, where `v.utxoStore` may be a `TxMetaCache`, the logger wrapper, or a mock — `logger_test.MockStore.GetBlockState` builds the pair from two separate calls (`stores/utxo/logger/logger_test.go:57-62`), so even the "single atomic pointer" half is not true of every implementation the validator can be handed.

**Fix:** Describe what the validator relies on and defer the mechanism to the interface:

```go
// GetBlockState returns the block height and median block time from the UTXO
// store as one read, so a single finality decision cannot mix values from two
// different chain tips as long as the store's writer publishes them together
// (issue 1443). See utxo.Store.GetBlockState for the exact guarantee and its
// limits.
```

### ChiR4 — The "prevents race conditions" overclaim survives in two other places

*Same as [@ordishs' point "docs/references/stores/utxo_reference.md:258-273" (07-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1485#pullrequestreview-4883338433).* Mine adds the in-tree occurrence: the `BlockState` type doc in the same file the PR edits.

**Problem:** The exact wording this PR deleted from the method is still present at `stores/utxo/Interface.go:50-53`, on the `BlockState` type itself: "an atomic snapshot ... preventing race conditions that could occur when reading them separately". And `docs/references/stores/utxo_reference.md` mirrors the interface verbatim: `:270-273` still carries the old `GetBlockState` comment, `:325-326` still summarises it as "an atomic snapshot", and `SetBlockState` is absent from the mirrored block between `:268` and `:270`.

**Why it matters:** The type comment sits some 550 lines above the method comment in the same file, so a reader who lands on `BlockState` first gets the claim the PR exists to retire, in the stronger wording. The reference doc is what an integrator reads before implementing `utxo.Store`, and it now mirrors an incomplete interface — anyone following it would not satisfy `utxo.Store` — while still promising the guarantee the real interface has stopped making.

**Fix:** Update the `BlockState` doc to match the qualified wording now on `GetBlockState`, and re-sync the mirrored interface block plus the `:326` summary line in `utxo_reference.md`, adding `SetBlockState`.

### ChiR5 — The duplicated representation can diverge permanently, not just transiently

**Problem:** Each store now holds the pair twice — `s.blockHeight`/`s.medianBlockTime` plus `s.blockState` — with no lock tying them together. `SetBlockState` writes the single-field atomics with plain stores and then the holder with `SetPair`, while `SetBlockHeight` writes its atomic and then the holder with a CAS. Two writers interleaving can leave them permanently, not transiently, inconsistent: `SetBlockState(10, 1010)` stores `blockHeight = 10`, a concurrent `SetBlockHeight(9)` then stores `blockHeight = 9` and CAS-es the pair, and the first call's trailing `SetPair(10, 1010)` lands last — final state `GetBlockHeight() == 9`, `GetBlockState().Height == 10`, and nothing corrects it until the next write. The PR description states this skew as "transiently disagree by one block", which holds only because production happens to have a single writer goroutine.

**Why it matters:** The invariant that keeps the two copies in step is "exactly one writer", and it is nowhere written down — not on the setters, not on the interface. `SetBlockStateSnapshotUnderConcurrency` uses one writer too, so nothing would catch a second one being added. Today's blast radius is small because production really does have one writer (the notification listener; the rewind CLI is a separate process), but the single-field readers are not all bookkeeping — besides DAH stamping (`services/blockassembly/Server.go:532`, `services/subtreevalidation/SubtreeValidation.go:902`) they include catchup's height comparisons (`services/blockvalidation/catchup.go:646`, `catchup_get_block_headers.go:134`) and alert enforcement (`services/alert/node.go:339`).

**Fix:** Either drop the duplication — point the single-field getters at the holder and delete the extra atomics, which removes the class entirely:

```go
func (s *Store) GetBlockHeight() uint32     { return s.blockState.Load().Height }
func (s *Store) GetMedianBlockTime() uint32 { return s.blockState.Load().MedianTime }
```

(one atomic pointer load plus a cached deref, versus one atomic load — immaterial even on `effectiveBlockHeight`'s per-spend path) — or, if the diff should stay minimal, document on `SetBlockState`/`SetBlockHeight`/`SetMedianBlockTime` that they are single-writer and that concurrent writers can desynchronise the snapshot from the single-field getters.

### ChiR6 — `close(stop)` isn't deferred, so a writer-loop failure leaks four spinning readers

*Same as [@ordishs' nit "defer close(stop) at tests.go:1051" (07-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1485#pullrequestreview-4883338433).*

**Problem:** In `SetBlockStateSnapshotUnderConcurrency` (`stores/utxo/tests/tests.go:1055`), `close(stop)` sits after the writer loop, and the loop body is `require.NoError(t, db.SetBlockState(i, i+offset))`. A single failing publish calls `FailNow` → `runtime.Goexit`, so `close(stop)` never runs and `wg.Wait()` is never reached; the four reader goroutines then busy-spin on `GetBlockState()` with no yield for the remainder of the test binary. `TestBlockStateHolderSnapshot` (`stores/utxo/block_state_test.go:62`) copies the pattern but its writer loop is a bare `holder.SetPair`, so it has no `FailNow` path — only a panic could leak there. The fix is worth applying to both for consistency, but the live hazard is the store-level helper.

**Why it matters:** On the shared aerospike store — where this case runs against a suite-wide instance — four readers keep hammering a live store while the rest of the run proceeds, turning one clean failure into slow or flaky neighbours and a confusing diagnosis.

**Fix:** The normal path must still stop and reap the readers *before* `require.Empty`, so a bare `defer close(stop)` won't do — dropping the explicit close would leave `wg.Wait()` blocking on readers that are still waiting for a channel closed only after the function returns. Make the stop-and-reap idempotent and call it on both paths:

```go
stop := make(chan struct{})

var stopOnce sync.Once
stopReaders := func() {
    stopOnce.Do(func() { close(stop) })
    wg.Wait()
}

defer stopReaders() // failure path: FailNow/Goexit still stops the readers

for i := baseHeight + 1; i <= baseHeight+iterations; i++ {
    require.NoError(t, db.SetBlockState(i, i+offset))
}

stopReaders() // normal path: reap before asserting; the deferred call is a no-op

require.Empty(t, torn, "GetBlockState returned torn pairs: %v", torn)
```

`sync.Once` makes the double invocation safe, and the second `wg.Wait()` returns immediately.

## Recap

| --- | --- | --- | --- |
| ChiR1 | Concurrency tests can pass vacuously | YES | 55% |
| ChiR2 | Height-zero precondition non-uniform | | 35% |
| ChiR3 | Validator comment overclaims guarantee | | 25% |
| ChiR4 | Stale overclaim in type doc and reference | | 22% |
| ChiR5 | Duplicated pair can diverge permanently | | 15% |
| ChiR6 | Reader goroutines leak on writer failure | | 8% |

freemans13:
@ctnguyen thank you — this was a careful review and ChiR1 in particular was worth more than it might have looked.

**ChiR1 was right, and my first attempt at it was wrong in exactly the way you predicted.** I initially took the view that a bare sample counter would be enough and that gating the writer loop was belt-and-braces, so that is what I implemented: count reads, assert the total after `wg.Wait()`. Then I tested it, and on **correct** code it fails at `-cpu=1` — `readers observed no snapshots at all`. Which is precisely what your review said would happen: *"a bare read counter turns a silent no-op into a single-core flake, so the writer loop has to be gated on the counter rather than just asserted against afterwards."* You had already ruled out the fix I reached for, and I had to reproduce the failure to see why.

It is now gated as you described — the writer keeps publishing until the readers have sampled — with two additions: a 30s safety deadline so a genuinely starved run fails cleanly instead of hanging until the test binary times out, and an early break once a tear is recorded, so a real failure still reports in milliseconds rather than chasing `minSamples` that the exited readers will never reach.

Verified against the mutation you suggested, SQL `GetBlockState` reverted to two independent atomic loads:

| | clean code | mutation reinstated |
|---|---|---|
| default `GOMAXPROCS` | passes 3/3 | fails 3/3 in 0.02s, torn pairs one tip apart |
| `-cpu=1` | passes 3/3 | passes |

One honest caveat on the bottom-right cell. Gating does not make `-cpu=1` catch the regression, and I do not think anything of this shape can: on a single processor the reader only interleaves where the scheduler preempts it, and the mutation's two loads sit adjacent, so landing between them is a coin flip nobody should rely on. What the gating does fix is the two failure modes that actually matter — the assertion can no longer pass vacuously because the readers never ran, and it no longer fails on correct code. The regression guard itself lives at `GOMAXPROCS > 1`, which is what CI runs. That limit is now written on the test and in the PR description rather than left implicit. If you think the case is not worth keeping on those terms, I am open to that argument.

The rest, in order:

**ChiR2 — done, both halves.** The precondition is now stated on the interface method, and `NullStore` enforces it. I took your framing that a contract holding in two implementations out of three is a contract by coincidence, and went slightly further: `NullStore` was the only store the shared suite never ran against, which is how the divergence survived in the first place, so it is now wired into `SetBlockHeightZero`, `SetBlockStateContract` and the snapshot case rather than just patched. Both its setters take the guard, since tightening only `SetBlockState` would have left the store inconsistent with itself, as you noted. Checked that nothing relied on the old laxity: every `SetBlockHeight(0)` call site in the tree is a sql store deliberately discarding the error.

**ChiR3 — was already fixed** in `dd13173a6`, which landed after the `b43c42a7` you reviewed. The comment now stops at the mid-read guarantee and hands same-tip provenance back to the writer. Your diagnosis was correct against the commit you had.

**ChiR4 — the in-tree half was fixed in `dd13173a6`; the reference doc was not, and you were right that it mattered more.** `docs/references/stores/utxo_reference.md` still carried the retired wording in three places and omitted `SetBlockState` entirely, so anyone implementing `utxo.Store` from the docs would not have satisfied it. Now mirrors `Interface.go`.

**ChiR5 — partly addressed, and I have taken the documentation option rather than the consolidation one.** `dd13173a6` deleted the `medianBlockTime` shadow outright, since its only reader was its own getter, which removes half the divergence surface. The `blockHeight` shadow stays: `effectiveBlockHeight`, `deleteAtHeightFor`, the `set_mined` paths, `un_spend.go`, `conflicting.go`, `alert_system.go` and `get.go` all read it directly, and pointing them at the holder is a wider diff than this PR should carry. Your permanent-divergence scenario is correct as written, and the invariant really is "exactly one writer" — that is now stated on the setters. If you would rather see the consolidation done here than documented, say so and I will do it; I read your "if the diff should stay minimal" as leaving that open.

**ChiR6 — the live hazard was already gone**, though not in a way the `b43c42a7` you reviewed made obvious. The writer loop records `writeErr` and breaks rather than calling `require.NoError` inline, so there is no `FailNow` between launching the readers and `close(stop)`, and the leak is unreachable. `close(stop)` is still not deferred. I have left it, since your own fix note is right that a bare `defer close(stop)` would deadlock `wg.Wait()` and the stop-and-reap helper is more machinery than an unreachable path warrants — but I am happy to add it if you would rather have the belt.

Separately, and in the spirit of the branch: your review prompted me to re-check the PR's own impact claim, and it was overstated. The description said both torn values "flowed into the same finality evaluation", but `selectFinalityComparisonTime` never reads `Height` and `MedianTime` in the same branch — they are used jointly a level up in `validateInternal`, and the torn window sits inside a wider pre-existing staleness, since the listener updates after the tip has moved. Softened accordingly. It would have been a poor look for a PR about a comment that overclaimed to ship under a description that did the same.

Changes are in `2fa222334` (the gofmt/gci failure), `6bed70ac4` (ChiR1) and `70bfa077a` (ChiR2, ChiR4). Re-ran green: all of `stores/utxo` including the aerospike, sql, factory, logger and nullstore packages, plus `stores/txmetacache`, `services/validator`, `services/blockpersister`, and the packages that use the null store — `services/subtreevalidation`, `services/legacy/netsync`, `services/blockvalidation` and `model`. `-race` on the holder and store-level cases, `go build ./...` and `go vet` clean, `golangci-lint` clean on the changed files.

ctnguyen (approved):
Review at commit [70bfa07](https://github.com/freemans13/teranode/commit/70bfa077a9e8793a3fa9a0879964f0907963454a)

Good round of follow-ups. The one blocking point from last time — the concurrency guard that could pass without a reader ever sampling — is properly closed now: both the holder test and the store-level case gate the writer loop on an actual sample count and bound it with a deadline so it can't hang. You also went further than asked on a couple of the nits (wiring `NullStore` into the shared suite rather than just patching it, removing the median-time shadow outright), and I appreciate the honesty of softening the PR's own impact narrative in the same spirit as the fix. The mechanism is sound and the diff stayed tight. Most of what I raised is resolved; what's left is two partials and one stray test line.

## Status of previous points

Three statuses carry caveats. **ChiR3** is a partial: the trailing same-tip overclaim is now correctly conditioned on the writer, but the comment (`services/validator/Validator.go:370-376`) still asserts a store internal from the consumer side — "the store keeps the pair behind a single atomic pointer" — and that half was part of the original finding (the suggested fix deferred the mechanism to the interface precisely because `v.utxoStore` may be a `TxMetaCache`, the logger wrapper, or a mock). `logger_test.MockStore.GetBlockState` (`stores/utxo/logger/logger_test.go:57-62`) builds the pair from two separate loads, so the single-pointer claim is not true of every implementation the validator can hold; it is marked PARTIAL FIXED for that reason. **ChiR1** is closed in substance — the writer loop genuinely gates on `samples.Load() >= minSamples`, which is the fix the point demanded, and it is mutation-verified — but the post-loop assertion is `require.NotZero(samples)`, not `>= minSamples`, so the 30s deadline is a hang guard rather than a strength guard: a pathologically starved run that sampled only a handful of times before the deadline would still pass. On a multi-core runner the loop exits via the sample gate long before the deadline, so this is an edge of the safety valve rather than the regression guard itself, which is why it stays FIXED — but tightening the final assertion to `>= minSamples unless a tear was already found` would close it fully. **ChiR5** is a genuine partial: the `medianBlockTime` shadow is gone, which removes half the divergence surface, and the remaining `blockHeight` shadow is now documented as single-writer on `SetBlockState` — but the duplication itself stays and the warning lives only on `SetBlockState`, not on `SetBlockHeight`, the other writer in the divergence scenario. **ChiR6**'s literal suggestion (`defer close(stop)`) was not applied, but the hazard is nonetheless eliminated: the writer loop records `writeErr` and breaks instead of calling `require.NoError` inline, so there is no `FailNow`/`Goexit` between launching the readers and `close(stop)`, and the leak is unreachable.

| --- | --- | --- | --- |
| ChiR1 | Concurrency tests can pass vacuously | YES | FIXED |
| ChiR2 | Height-zero precondition non-uniform | | FIXED |
| ChiR3 | Validator comment overclaims guarantee | | PARTIAL FIXED |
| ChiR4 | Stale overclaim in type doc and reference | | FIXED |
| ChiR5 | Duplicated pair can diverge permanently | | PARTIAL FIXED |
| ChiR6 | Reader goroutines leak on writer failure | | FIXED |

## Non-blocking issue

### ChiR7 — Dead `SetBlockState` mock expectation left in the factory test

*Same as [@ordishs' nit "drop the dead SetBlockState mock expectation at factory/utxo_test.go:355" (07-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1485#pullrequestreview-4883338433).*

**Problem:** `TestNewStore_BlockchainClientError` (`stores/utxo/factory/utxo_test.go:355`) primes `mockStore.On("SetBlockState", mock.Anything, mock.Anything).Return(nil).Maybe()`, but the test deliberately clears `BlockChain.GRPCAddress` to make blockchain-client creation fail, so `NewStore` returns before it ever subscribes or starts the listener — `SetBlockState` is never called. The `.Maybe()` keeps the unmet expectation from failing the test, but the line is dead: it documents an interaction the test's own setup guarantees cannot happen.

**Why it matters:** Small, and non-blocking. It is misleading rather than harmful: a reader takes the expectation as evidence the listener path is exercised here, when the test exists to assert the opposite — that construction bails out early. It also survived the same churn that touched the surrounding block-state wiring, so it is worth clearing while the file is open.

**Fix:** Drop the line. Nothing in this test reaches the setter, so no expectation is needed:

```go
	// Register mock database initializer
	mockStore := &MockUTXOStore{}

	availableDatabases["memory"] = func(...) (utxo.Store, error) {
		return mockStore, nil
	}
```

PR #1489, opened 2026-08-06
[Blocked: bsv-blockchain/go-bt#174] fix(subtreevalidation): reject non-minimal CompactSize prefixes instead of silently canonicalizing peer bytes

### LINKED ISSUE #1421: [HARDEN-0805] Reject non-minimal CompactSize length prefixes so we cannot accept a transaction SV Node throws out
## What can go wrong

Bitcoin and SV Node insist that a length prefix (a "CompactSize" variable-length integer, or varint) be written in its shortest possible form. Writing the value 1 as three bytes (`fd 01 00`) instead of one byte (`01`) is a hard parse error there. Teranode's native transaction parser, supplied by the `go-bt` library, reads those prefixes with no such check and happily returns the value regardless of how it was encoded.

If a block ever carries a single transaction whose length is written non-minimally, an SV-Node fleet rejects that block at parse time while a Teranode fleet accepts it. Teranode then keeps building on a chain the SV majority orphans. Because the block is already written to disk, restarting the node does not clear it — the node stays on the wrong chain. That is a silent consensus split: no error is raised on either side, the two fleets simply disagree about which chain is real.

**Important caveat (this is why the fix is gated on a spike, not just a test).** A transaction's txid is the hash of its *re-serialized* bytes. If `go-bt` re-serializes the non-minimal prefix back into canonical (shortest) form when it rebuilds the transaction, then Teranode's recomputed txid no longer matches the one the miner committed to, and the block is rejected at the merkle-root check instead — a different, non-silent rejection. The silent-accept split therefore only holds if `go-bt` round-trips the exact non-minimal bytes faithfully. That round-trip fidelity must be confirmed before we call this a fleet split.

## Where (verified on main @ 27a0178fd)

- `go-bt/v2@v2.6.7/varint.go` — `VarInt.ReadFrom` at line 109. The `0xfd` branch (lines 132-138) reads two bytes and returns the value with no check that it is >= 0xfd (i.e. that it actually needed two bytes); the `0xfe`/`0xff` branches (lines 116-130) do the same for 4- and 8-byte forms. There is no minimality/canonicality check anywhere in the function. This is the varint reader used throughout `go-bt`'s transaction deserialization, which is Teranode's native parse path for transactions in propagation, subtree validation, and block validation.
- Contrast `go-wire@v1.2.10/common.go` — `ReadVarInt` at line 534 *does* enforce minimality: the `0xff` case rejects a value below `0x100000000` (lines 552-559) and the `0xfe` case rejects a value below `0x10000` (lines 571-578), returning a `messageError("ReadVarInt", ...)` non-canonical error. This is the legacy wire parser, so the two parsers already disagree on the same bytes.

(Both pins are current in `main`'s go.mod: `go-bt/v2 v2.6.7` at line 14, `go-wire v1.2.10` at line 205.)

## Why it matters

The two parsers reaching different verdicts on the same bytes is the raw material for a chain split, and it is exactly the kind of adversarial-only input honest miners never emit — so it has never appeared organically and no test guards it. It sits in Theme 1 (silent consensus divergence), the most dangerous class, because the first violation is also the first symptom and the block persists across restarts.

## How to reproduce / test

1. Take a real, valid transaction. Rewrite one of its length prefixes (for example the input count or a script length) from its minimal form into a longer non-minimal form (e.g. encode a small count as `fd XX 00`). Keep the rest of the bytes identical.
2. Feed those bytes to the `go-bt` transaction parser and confirm it parses without error and returns the same logical value — proving Teranode accepts the malformed encoding.
3. Feed the same bytes to `go-wire`'s `ReadVarInt` path and confirm it returns the non-canonical error — proving the two parsers disagree. This asymmetry alone is the unit-test deliverable.
4. **Spike (do this before treating it as a fleet split):** re-serialize the `go-bt`-parsed transaction and compare the bytes to the input. If they come back non-minimal (faithful round-trip), the recomputed txid matches the miner's and the malformed block would be silently accepted — a real split. If they come back canonical, the txid diverges and the block is instead rejected at the merkle-root check — still a bug (two implementations disagree) but not a silent accept.

## Suggested fix

If the spike shows a faithful non-minimal round-trip: add a minimality check to Teranode's transaction parse path so a non-minimal CompactSize is rejected at parse time, matching SV Node and `go-wire`. The cleanest place is at the `go-bt` varint reader (upstream fix) or a validating wrapper Teranode applies on the untrusted parse path before a transaction is admitted. Do not silently canonicalize and continue — that would let a malformed transaction change identity mid-flight; reject it.

## Confidence

**Confirmed** that the two parsers disagree (go-bt accepts non-minimal, go-wire rejects it) — read directly on main. **Needs a spike** on the exploitation chain (whether go-bt's round-trip preserves the non-minimal bytes and therefore whether the outcome is a silent accept or a merkle-check rejection).

---
*Part of the `[HARDEN-0805]` production-hardening sweep (label `hardening-map-2026-08-05`), grounded in `main` @ `27a0178fd`. Priority: High · Size: S · This task is self-contained and does not depend on any other ticket.*

freemans13:
## What happened

Bitcoin and SV Node require every CompactSize length prefix to be written in its shortest form — writing the value 1 as `fd 01 00` instead of `01` is a hard parse error there. The `go-bt` varint reader Teranode uses has no such check.

The issue asked for a spike before any fix, because the exploitation chain depended on an open question: does `go-bt` round-trip the non-minimal bytes faithfully? **The spike answered no.** `go-bt` accepts the over-long prefix and then *re-serializes it canonically*, so the txid Teranode computes is byte-identical to the canonical transaction's. Measured directly: a 95-byte canonical transaction rewritten with a 3-byte input-count prefix parses to 97 wire bytes, re-serializes to the original 95, and yields the same txid.

That answer cuts both ways, and it took a review round to see the second edge. Because the txid is unchanged, the merkle root is unchanged too — so **a non-minimal prefix is a property of how a peer delivered the bytes, not a property of the block.** The same subtree fetched from an honest peer parses to exactly the same transactions and validates fine. There is no such thing as a "non-minimal block" on these paths; there is only a peer that served an over-long encoding of a block that is otherwise perfectly good.

That reframing matters because it also tempers the fleet-split story the issue was written around. Subtree data is a Teranode-internal distribution format: a receiving node canonicalizes on parse and relays canonical bytes onward, so SV Node never sees the non-minimal form by this route. The legacy p2p path, which SV Node *does* share, is already safe — `go-wire`'s `ReadVarInt` enforces minimality.

So the case for this change is not "stop a fleet split", it is the plainer one: **fail closed on malformed peer input instead of laundering it to disk.** Today the parse silently normalizes the payload, stores the canonicalized bytes, and discards the evidence — after which nothing can tell that this node accepted bytes SV Node rejects at parse. Every hash-based check downstream passes, precisely because canonicalization defeats them. The parse path is the only place that can still see it.

Closes #1421.

## Changes

- **`services/subtreevalidation/canonical_tx.go` (new)** — `checkCanonicalTxEncoding` compares the bytes consumed from the wire against the transaction's own serialization. Any non-minimal prefix makes the wire form longer, so a single length comparison detects every such encoding without a re-parse. The transaction is **rejected**, not accepted in canonical form: silently canonicalizing would let a malformed transaction change identity mid-flight.

- **The failure is a `ProcessingError`, deliberately not a `TxInvalidError`** — this is the substantive correction from review, and the code now carries a comment saying so at both sites. `ErrTxInvalid` routes through `BlockValidation.go` into `storeInvalidBlock`, marking the block permanently invalid and surviving a restart, and `isUnvalidatablePeerError` then stops the node trying any other source. Given that a non-minimal prefix says nothing about the block, that would have handed any peer we fetch subtree data from a cheap way to permanently poison a valid block — no mining required, strictly worse than the hazard being closed. Wrapping did not contain it: Teranode's `errors.Is` walks the chain by code, so a `TxInvalidError` inside a `ProcessingError` still matches `ErrTxInvalid`, which was verified rather than assumed. `ProcessingError` matches the sibling "peer served bytes whose hash doesn't match" check a few lines away in the same function, and the branch `BlockValidation.go` already treats as "do not mark the block as invalid". The peer still gets a failure attributed against it; it is just not marked malicious. The tests assert the error is **not** `ErrTxInvalid`, so the regression cannot return silently.

- **Applied at all four peer-supplied parse points** — the block-subtree data stream (`readTransactionsFromSubtreeDataStream`) and the single-transaction fetch (`readTxFromReader`) check per transaction, where the parser exposes a byte count. The bulk subtree-data fetch (`getSubtreeMissingTxs`) and the catchup prewarm (`fetchAndStoreSubtreeData`) check over the whole payload via `util.CheckCanonicalSubtreeData`, because `go-subtree` discards the per-transaction counts. The last of those runs *before* block validation and stores the re-serialized form, so leaving it out would have laundered a non-minimal payload to disk ahead of every downstream check.

- **Extended-format handling** — Teranode's subtree data may carry *extended* transactions (each input also carrying its parent's satoshis and locking script), which serialize longer than `tx.Size()` reports; a file written from parsed transactions keeps whatever form they were in, while the asset service's on-demand generator writes the standard form. The comparison is therefore made against whichever serialization the transaction actually uses, or every legitimate extended transaction is rejected (caught by the existing suite before this PR was opened). That size is computed **arithmetically** via the new `util.ExtendedTxSize`, not by serializing: this loop was deliberately built allocation-free (arena for script bytes, pooled reader, reused hash scratch), and `ExtendedBytes()` would have allocated and copied every transaction twice just to read a length. The arithmetic already existed unexported in the Aerospike store; it is lifted to `util` so a consensus-adjacent size calculation has one implementation, with the store's existing equivalence tests still passing against it.

- **Coinbase accounting** — `go-subtree` parses a leading coinbase into `Txs[0]` for a subtree whose first node is the coinbase placeholder, but writes the file from index 1. The parser therefore consumes bytes the serialization never re-emits, and the whole-payload comparison has to add the coinbase's canonical size back or it rejects every legitimate first subtree of a block. Its size is *added* rather than the consumed bytes being ignored, so a non-minimal coinbase is still caught.

- **Tests** — the canonical form is accepted; the two-byte and eight-byte encodings of a one-byte count are rejected; the error is asserted to be `ErrProcessing` and explicitly **not** `ErrTxInvalid`; and a test documents the property that makes this worth guarding at all, asserting the non-minimal bytes yield the *same* txid as the canonical transaction, so the merkle check provably cannot catch it.

## Trade-offs and testing

- The check is applied on the peer-supplied subtree parse paths, not on every `bt.NewTxFromBytes` call in the repo. A non-canonical transaction arriving via propagation is canonicalized into the mempool and mined canonically, and the legacy wire path is already covered by `go-wire`'s minimality check, so the parse points where a block's transactions are read are the surface that matters.
- One path is knowingly left unguarded: `model/subtree_meta_regenerator.go` parses peer subtree data without the check. It derives subtree *meta* only and never persists the payload, and since `go-bt` canonicalizes, the derived meta is identical to what canonical bytes would give — so it is not a block-acceptance bypass. Noted rather than fixed to keep the diff on the paths that matter.
- The cost is one integer comparison plus an allocation-free size computation per transaction — measured neutral against the unmodified loop (the first cut, which serialized, cost +22% time and +174% allocations and was replaced after review).
- The arithmetic size is pinned against the real serialization by tests covering input/output counts, nil previous scripts, 3- and 5-byte script-length prefixes, unset previous-txid hashes and the no-inputs case, plus an allocation assertion. That equivalence is the one thing that would silently break the check if go-bt's extended layout ever changed, so it is tested rather than assumed.
- A known `go-bt` quirk is documented rather than worked around: `IsExtended()` tests `Inputs == nil` before the extended flag, so an extended transaction with no inputs is measured in its standard form and would be rejected with a misattributed reason. Harmless, because a transaction with no inputs is consensus-invalid regardless.
- **Testing limit worth stating:** the coinbase-included branch of the catchup prewarm is covered by the `util`-level unit test but not end-to-end through either real path, and all end-to-end fixtures carry standard-format transactions rather than extended ones. That coinbase line is the one an earlier commit on this branch had to fix after it caused a false rejection, so it is the highest regression risk here and deserves an end-to-end case.
- The proper long-term fix is upstream in `go-bt`'s varint reader, so every consumer benefits; this is the Teranode-side guard for the untrusted path in the meantime.
- Ran: `make lint` (0 issues), `go build ./...`, and the full `./util/ ./services/subtreevalidation/ ./services/blockvalidation/ ./stores/utxo/aerospike/` suites green, plus `-race` on the new tests.


freemans13:
Review round 1 (consensus/completeness, Go cost, test adequacy — adversarially verified): 8 raw findings, 3 confirmed, 5 refuted. All three confirmed findings were the same issue, reached independently by all three lenses and backed by benchmarks:

**The check undid the optimisation of the loop it sits in.** Subtree data carries extended transactions, so my `ExtendedBytes()` call ran for essentially every transaction of every block, allocating and copying the whole transaction twice (go-bt sizes its buffer from the *standard* size, then regrows for the larger extended form) purely to learn a length. Measured: 157ns and 1088 B per call; +22% time and +174% allocations against a loop deliberately built allocation-free with an arena, a pooled reader and a reused hash scratch buffer. Fixed in 3425a49dd: the allocation-free arithmetic already existed unexported in the Aerospike store, so it is lifted to `util.ExtendedTxSize` and shared — one implementation of a consensus-adjacent size calculation, pinned against the real serialization by tests across input/output counts, nil previous scripts, 3- and 5-byte length prefixes, unset previous-txid hashes and the no-inputs case, plus an allocation assertion. The store's existing equivalence tests still pass against it.

Refutations worth recording: the coinbase and catchup/revalidation paths were both argued to bypass the check and both were refuted on tracing; `IsExtended()`'s nil-inputs ordering was shown not to skew the size; and the coverage complaints about the extended branch are now answered by the new equivalence tests.

One process note: a reviewer's mutation experiment briefly inverted the comparison in my working tree — verified restored, and the committed check reads `bytesRead != canonicalSize` as intended.

freemans13:
Review round 2 (closure audit + fresh eyes): round-1 fix verified closed, and one new major finding confirmed with a working probe — fixed in 8eee72d7f.

**The bulk subtree-data fetch bypassed the check entirely.** When more than `percentageMissingGetFullData` (default 20%) of a subtree is missing locally — routine during catch-up, and trivially forced by an attacker who simply withholds transactions — the node takes the `/subtree_data/` branch instead of the per-transaction fetch. That branch parses via go-subtree, which discards the per-transaction byte counts my check relies on. Worse, it then stored the **re-serialized canonical form**, silently normalising the payload and discarding the evidence — precisely what the check's own doc comment forbids. The subtree was then blessed, and the block referencing it skipped the per-transaction parse entirely (`CheckBlockSubtrees` early-returns when the subtree file already exists), so nothing downstream could catch it. The reviewer demonstrated this end to end rather than arguing it, and the verifier reproduced it independently.

The parser's consumption is now counted and compared against what the parsed data serializes back to — the same property, applied over the whole payload — and a mismatch rejects rather than storing the canonicalised form.

The closure audit also confirmed round 1 is genuinely closed: the check is now allocation-neutral against the unmodified loop, and it could not find a transaction shape where `util.ExtendedTxSize` disagrees with the real serialization.

freemans13:
Review round 3 (exhaustive coverage re-audit + fresh eyes): 2 confirmed, 0 refuted — both fixed in f1f2ff009. One of them was a defect I introduced in round 2.

**Critical, mine: the payload check double-counted the coinbase.** It assumed the parser's consumption equals what the parsed data re-serializes to. For the *first* subtree of a block those legitimately differ by exactly the coinbase — go-subtree parses a leading coinbase into `Txs[0]` but writes the file from index 1 — so a perfectly canonical payload was rejected and `getSubtreeMissingTxs` hard-aborted with no fallback, reported as the peer serving malformed data. Reproduced in-package: 180 consumed vs 85 serialized, the 95-byte gap being the coinbase. Whether a peer's response carries the coinbase depends on whether that peer persisted the block, so this would have looked intermittent in the field. The coinbase's canonical size is now added back rather than its bytes ignored, so a non-minimal coinbase is still caught.

**Major: the catchup prewarm had the same gap round 2 closed at tip.** `fetchAndStoreSubtreeData` (blockvalidation) parses the same peer endpoint unchecked — and it is worse there, because it runs *before* validation and stores the re-serialized form, laundering the bytes to disk so the per-transaction check downstream only ever sees canonical data and can never fire. Demonstrated with a probe: 798 non-minimal bytes in, 796 canonical bytes stored, byte-identical to an honest serialization. Now checked with the same helper.

The check, the byte counter and the extended-size arithmetic now live in `util` so all three call sites share one implementation. Note this is the second sibling path found by tracing rather than by test — worth a reviewer's eye on whether any remaining `NewSubtreeDataFromReader` caller (the meta regenerator, flagged as lower severity since it only serves merkle proofs) should be covered before merge.

freemans13:
Review round 4 (false-positive hunt + exhaustive coverage re-audit): **zero confirmed findings** — converged after four rounds (3 → 1 → 2 → 0 confirmed, every one fixed).

The false-positive lens specifically hunted for further shapes where an honest canonical payload would now be rejected — subtrees without a coinbase placeholder, coinbase-only subtrees, empty payloads, trailing bytes, extended/standard mixes, and whether the catchup path's `coinbaseAtZero` flag is exactly equivalent to go-subtree's own condition — and found none. That matters because round 3 caught exactly such a defect in my own round-2 fix, so this was the check that most needed to come back clean.

The one refuted finding produced a genuinely useful conclusion rather than a dismissal: the coinbase in the Teranode block message is parsed by go-bt without a minimality check, verified by probe — but it cannot cause a split, because that message is an internal asset-service format no SV Node parses, `Block.Bytes()` re-serializes canonically from the store, and the legacy P2P route decodes Bitcoin-format blocks with go-wire, which is strict. Filed with the meta-regenerator site as #1490 (consistency, not a live hazard), including the one thing the reviewer flagged as read-not-run: the legacy decode path's strictness was established by reading go-wire rather than end-to-end, and that should be confirmed before acting on it.

freemans13:
CI follow-up: the SonarQube gate flagged the new catchup path as uncovered — fair, since that fix arrived from a reviewer's throwaway probe rather than a test. Made permanent in 18313b334: a canonical payload is accepted and stored, and one whose first transaction carries a 3-byte input-count prefix is rejected with `ErrTxInvalid` **and leaves nothing in the store**. That last assertion is the important one — the laundering step is what made the original bypass invisible to every downstream check, so it needed pinning explicitly rather than just asserting the error.

blockvalidation suite: 823 green.

freemans13:
CI note on the `smoketest (shard 0)` failure (`TestBIP68_AtExactCSVHeight`, "should sync to height 119"): this is **not** caused by the change, and I checked rather than assumed, because a false positive stalling sync is exactly the failure mode this PR could plausibly have introduced.

Three independent reasons:
1. The previous CI run passed the same smoketest with **byte-identical production code** — `git diff f1f2ff009 18313b334 -- ':!*_test.go'` is empty; the only new commit adds a test file.
2. That test syncs Teranode from an SV Node over **legacy P2P**, which decodes with go-wire. Every check this PR adds is on the Teranode-to-Teranode HTTP `/subtree_data/` paths; `grep` confirms `services/legacy/` has no `NewSubtreeDataFromReader` or `subtree_data` call site.
3. If the check were falsely rejecting, it would surface as a `non-canonical subtree data` / `not canonically encoded` error in the logs. Neither string appears anywhere in the failing job's output.

Consistent with the known flakiness of this suite under CI contention. Re-running.

freemans13:
The smoketest passed on re-run, confirming the flake diagnosis above (production code was byte-identical to the previously-passing run).

Remaining was the Sonar gate at 79.7% coverage on new code against an 80% threshold. Rather than chase the number, I filled the gap a round-1 reviewer had already named: the per-transaction check was only tested through its helper, never through the real block-path ingestion. 83ace4331 now drives `readTransactionsFromSubtreeDataStream` itself — a canonical stream parses, and one carrying a 3-byte input-count prefix is rejected. That is the parse the whole PR turns on, since the canonicalized txid still satisfies every hash-based check downstream.

freemans13:
## Draft — waiting on an upstream fix in go-bt

Converting this to draft. Reviewing it now would mean reviewing ~700 lines that should be deleted rather than merged, and the reason is worth setting out properly.

**This is a `go-bt` bug, and the fix belongs there.** `(*VarInt).ReadFrom` reads the discriminant, decodes the value and returns it without ever checking whether a narrower form would have done. There is no minimality concept in the reader at all. Meanwhile `go-wire` — same organisation, same ecosystem — has enforced exactly this rule in its `ReadVarInt` for years, complete with an `errNonCanonicalVarInt`. `go-bt` is simply inconsistent with its sibling.

Upstream PR: **bsv-blockchain/go-bt#174** → https://github.com/bsv-blockchain/go-bt/pull/174

### Why the fix upstream is ~40 lines and this PR is ~700

`VarInt.ReadFrom` is the single choke point for all transaction decoding: `NewTxFromBytes` goes through `Tx.ReadFrom` → `ReadFromWithArena`, and every input count, output count and script length is read through it. One check there covers everything.

This PR cannot see the varint, so it detects the same problem indirectly — by observing that the transaction came out shorter than it went in. That forces the entire edifice: wrap readers to count consumed bytes (`CountingReader`), compute what the canonical size should have been (`CanonicalTxSize`), which requires distinguishing standard from extended form (`ExtendedTxSize`, plus its own test suite), which then trips over go-subtree's omitted-coinbase asymmetry (`OmittedCoinbaseTx`), applied at four call sites that have to be found and kept in sync. Every line of that is scaffolding for working one layer above the problem.

The upstream fix also covers ground this PR does not: the fifth parse point in `model/subtree_meta_regenerator.go` that this PR knowingly skips, and the transaction-submission ingress in propagation, validator and rpc, which this PR never touched.

### Evidence that it supersedes rather than duplicates

Against the patched `go-bt`, Teranode builds clean and its transaction-parsing suites — `util`, `model`, `validator`, `propagation` — all pass unchanged. The **only** failing tests anywhere are the four in this PR, and they fail because they can no longer construct their own fixtures:

```
non-minimal varint encoding: 1 read as 3 bytes, minimal encoding is 1
```

`TestNonMinimalTxKeepsCanonicalTxID` is the clearest case. It exists to document the hazard by asserting that non-minimal bytes yield the *same* txid as the canonical transaction — the property that makes every downstream hash check useless here. It can no longer demonstrate that, because `go-bt` refuses the bytes before a txid exists. The hazard is closed one layer down.

The full existing `go-bt` suite also passes unchanged, which is the compatibility signal that mattered most: nothing in the library depended on the lenient behaviour.

### Plan

1. Land bsv-blockchain/go-bt#174 and cut a tag.
2. Bump `go-bt` here.
3. Reduce this PR to that bump, deleting `CountingReader`, `CanonicalTxSize`, `OmittedCoinbaseTx` and all four call-site guards — roughly 700 lines.

`util.ExtendedTxSize` is the one piece worth keeping on its own merits: the arithmetic predates this PR, having been lifted out of `stores/utxo/aerospike/create.go`, and the Aerospike store still uses it to size records without serializing.

Closes nothing yet — issue #1421 stays open until the bump lands.

PR #1496, opened 2026-08-06
fix(blob/file): fit file concurrency under the open-file limit the OS actually grants

### LINKED ISSUE #1431: [HARDEN-0805] Raise and validate the open-file limit at startup against the configured file-descriptor concurrency
## What can go wrong

The node never raises its own open-file-descriptor limit at startup, and never checks that its configured file-operation concurrency actually fits under the operating system's limit. If a host's real `ulimit -n` sits below the node's configured file-descriptor budget, the kernel's "too many open files" (EMFILE) can still stop file operations — and the node neither raises the limit nor fails loudly to warn the operator. It is a silent misconfiguration failure.

## Where

`stores/blob/file/file.go` (commit `27a0178fd`). The file blob store bounds concurrent file operations with two global weighted semaphores — read (default 768) and write (default 256), total 1024 (constants at lines 198-201; the "ULIMIT PROTECTION" design comment at 204-240). `InitSemaphores` (line 322), called from `cmd/teranode/daemon.go:73`, validates the configured read/write limits **only** against the hardcoded constants `MinSemaphoreLimit` (1) and `MaxSemaphoreLimit` (256000) — it never reads the actual OS limit (getrlimit `RLIMIT_NOFILE`), never compares the semaphore total against it, and never fails when the OS limit is below the configured total. A repo-wide grep for `setrlimit` / `Rlimit` / `RLIMIT` across `.go` files returns nothing, so startup never raises the limit either. The default total (1024) is exactly a stock `ulimit -n` with zero headroom for the sockets, log files, and other descriptors the process also opens.

## Why it matters

The concurrency half is genuinely handled — opens are hard-bounded with backpressure by the semaphore (a blocking, timeout-aware acquire in `acquireReadPermit`/`acquireWritePermit`), so this is not an "EMFILE under load" hazard. The residual gap is a *misconfiguration* one: on a box whose real limit is below the configured semaphore total plus the descriptors the rest of the process needs, the node can still hit EMFILE, and nothing at startup raises the limit or refuses to start with a clear message. Production boxes have been hand-tuned with high limits, which is why this has never bitten — but it is one misconfigured host away from a silent failure.

## How to reproduce / test

Start the node with `ulimit -n` set below the configured semaphore total (for example 512 against a 1024 total) and observe that it neither raises the limit nor fails loudly — it starts anyway and can later hit EMFILE from the non-semaphore descriptors. A unit test can call the new startup validation with a stubbed `RLIMIT_NOFILE` below the configured total and assert it returns a loud configuration error.

## Suggested fix

At startup, read `RLIMIT_NOFILE`; attempt `setrlimit` to raise the soft limit toward a computed target (up to the hard limit); then assert that the read+write semaphore total plus a headroom margin (for sockets, logs, and other descriptors) fits under the effective OS limit, failing loudly with a clear configuration error if it does not. Keep the existing semaphores as the concurrency bound.

## Confidence

Confirmed — the semaphore is present, `InitSemaphores` validates only against hardcoded constants, and there is no `setrlimit`/`getrlimit` anywhere in startup; all read directly on main.

---
*Part of the `[HARDEN-0805]` production-hardening sweep (label `hardening-map-2026-08-05`), grounded in `main` @ `27a0178fd`. Priority: Low · Size: S · This task is self-contained and does not depend on any other ticket.*

freemans13:
## What happened

The file blob store hard-bounds concurrent file operations with two semaphores (read 768 + write 256 = 1024 by default). That bounds *our* concurrency — but nothing checked it against what the operating system actually allows. The default budget of 1024 is *exactly* a stock `ulimit -n`, leaving zero descriptors for the sockets, gRPC connections, database pools and log files the process also opens, none of which the semaphores see.

Worse, the one place that did consult a system limit consulted the wrong one. Startup worked out how much concurrency to allow by shelling out to `ulimit -u` — the max **user processes** limit, not `ulimit -n`, the open **files** limit. So the file store was sized from an unrelated kernel resource, and with `filestore_use_system_limits` defaulting to true, that was the path a stock node took.

This PR reads the real limit and fits the store's concurrency underneath it.

Closes #1431.

## What it does

**`util/fdlimit` (new)** — `Budget()` reads `RLIMIT_NOFILE` and reports the descriptors spare for file operations: the soft limit, less a reserve. It never fails because the limit is small; the caller is expected to bound its own concurrency rather than refuse to run.

**The reserve is the part that matters.** `Headroom` (512) is held back for precisely the descriptors the semaphores *don't* bound. Sizing the store against the raw limit would let it consume everything and starve the sockets instead — the same failure wearing a different hat. Where the whole limit is below twice the reserve, half the limit is held back instead, because 512 out of 512 leaves nothing to work with. At 100,000 the reserve is still exactly 512.

**`stores/blob/file/file.go`** — `InitSemaphores` scales read and write concurrency down proportionally to fit that budget, keeping at least one of each, and returns the limits actually in force. `filestore_use_system_limits` is wired to that clamp: on, and concurrency is reduced to fit; off, and the configured values stand with the risk accepted — both halves of what the setting documents.

**`cmd/teranode/daemon.go`** — the `ulimit -u` shell-out is gone, and a reduction is now logged as a warning naming both the applied and the configured values.

**Windows and friends are handled by build tag.** There is no `RLIMIT_NOFILE` to read, so `Budget` reports the limit as unknown and startup carries on. The semaphores still bound concurrency there, so an unreadable limit is not a reason to refuse to start.

## It does not raise the limit, and that is deliberate

An earlier revision of this PR raised the soft limit toward what the store needed. It turned out to be work already done.

Go's runtime raises `RLIMIT_NOFILE` to the hard limit in a `syscall` package `init()` that runs before `main()`, on every unix platform — and on Darwin it applies `adjustFileLimit`, capping the request at `kern.maxfilesperproc`. Verified live: this shell reports a soft limit of 1,048,576, but a Go binary reads 245,760 at `main()`, which is exactly `kern.maxfilesperproc`.

So by the time application code runs, the soft limit is already at `hard-1`. Working through the branches, our raise could only ever move it to `hard` — a gain of one descriptor — and where Go's init had skipped because soft already equalled hard, it did nothing at all.

That one descriptor was paying for most of the package: `setRlimit`, the `assignRlim` shim that existed solely because FreeBSD and DragonFly model `rlim_t` as signed, a macOS read-back guarding against a cap the runtime already applies, a saturating headroom addition with its own overflow case, and the `required` parameter, which only ever fed the raise target. All of it is gone. `fdlimit.go` went from 107 lines to 49 and its tests from 314 to 85.

There is a principle underneath the arithmetic, too. The hard limit is where `ulimit -Hn`, systemd `LimitNOFILE` and container runtimes express operator policy. Reading it and fitting underneath is the more honest posture than negotiating with it.

## Clamp, not refuse

An earlier revision also **refused to start** when the budget did not fit. Review reproduced why that was wrong: under a 1024 limit, the very process it refused could hold 1020 files open at once.

The semaphores are a *ceiling* on concurrent operations, not a reservation — a descriptor is held only for the duration of one operation — so a node whose ceiling exceeds the OS limit runs fine at any realistic load. To actually hit EMFILE you would need 1024 file operations genuinely in flight at once, with nothing left for anything else.

And the failure it was preventing is already handled gracefully: an operation waits up to 25 seconds for a permit and then returns `ServiceUnavailable` (`stores/blob/file/file.go:411`). Bounded, per-operation, recoverable. Refusing to start replaced that with total unavailability from boot — and because the caller panics under a restarting systemd unit, a crash loop. It also fired on real, deliberate configurations: any host whose hard limit is below the budget, such as an explicit `LimitNOFILE=1024` unit or a limits.conf-hardened image.

The rule the code now follows is: **panic on broken invariants, degrade on resource pressure.** A tight descriptor budget means the machine is smaller than we would like; nothing is incorrect, throughput is just lower. This file already agreed with itself on that — low disk space returns `ServiceUnavailable` from the health check rather than killing the process.

## Telling the operator

Degrading quietly would be its own failure, so the reduction is reported where someone will see it.

It used to be a `fmt.Printf` to stdout, because `InitSemaphores` ran before the logger existed — invisible to structured logs and to anything that could alert on it. `InitSemaphores` now runs immediately after the logger is created (nothing between there and `daemon.Start()` opens a file store), and a clamp is a `Warnf` naming both the applied and the configured values, and pointing at all three places the ceiling can come from: `ulimit -Hn`, systemd `LimitNOFILE`, and macOS `kern.maxfilesperproc`.

## Two behaviour changes worth naming

On hosts where the old `ulimit -u` code worked at all, concurrency now stays at the configured 768/256 instead of growing to three quarters of the max-user-processes limit. Scaling concurrency *upward* off a system limit was the opposite of what the setting documents ("may reduce configured concurrency"), and wrong in direction regardless.

And on Debian and Ubuntu it never worked: `/bin/sh` is dash, dash has no `ulimit -u`, so the command errored and the budget silently fell back to the configured values — which is now what it does deliberately.

As a side effect, startup also got faster. The deleted fork-and-exec of `/bin/sh` benchmarks at ~3.0 ms on this machine; the single `getrlimit` that replaces it costs ~100 ns.

## Trade-offs

- On a clamped host the node runs with less file concurrency than configured. That is the correct trade — reduced throughput beats not booting — and the warning says so explicitly, naming both numbers.
- The clamp keeps at least one read and one write permit, so the store stays functional however small the budget. At a budget of one this deliberately overshoots by one descriptor, borrowing from the reserve; that is the price of a store that can do anything at all.
- `Headroom` is a fixed 512 with no setting behind it. On a node with a very large peer set plus several database pools that could be tight — but it only binds when the OS limit is itself small, where there is no better answer anyway. A setting here would be one more thing to misconfigure.
- If you genuinely want fail-fast rather than degrade, that is a reasonable thing to want, but it should be opt-in rather than the default. Not in this PR.

## Testing

Ran: `go build ./...`, `go vet`, `go test -race -shuffle=on` on `./util/fdlimit/`, `./stores/blob/file/` and `./settings/` (all green), `golangci-lint` over the touched packages (0 issues), and a GOOS/GOARCH cross-build of `./util/fdlimit/` across every pair `go tool dist list` reports (**47 ok, 0 fail**) — the `unix` / `!unix` split covers all of them.

Tests pin: the reserve across the whole range, including the boundary at 1023/1024/1025 where the proportional reserve hands over to the fixed one; that an unreadable limit is distinguishable from a genuine budget of zero, and leaves concurrency alone; and that the clamp preserves the read/write ratio, fits the budget, keeps a usable store at budgets of 0, 1 and 2, and leaves an ample budget untouched. `Budget()` is at 100% statement coverage — the only uncovered statement in the package is a real `Getrlimit` failure, which cannot be provoked in a test.

**Limits of this testing, stated openly.** The full `make test` was not run locally — it is too heavy for this checkout — so coverage outside the touched packages rests on CI. `InitSemaphores` itself is still not covered end to end: it is guarded by a `sync.Once` and runs once per process, so the decision is tested through `resolveConcurrency`, which it delegates to. Cross-compiling the whole binary from macOS fails on the `bdk` CGO dependency, which is pre-existing and unrelated.


freemans13:
Review round 1 (false-refusal hunt, Go/portability — adversarially verified): 1 critical confirmed, and it was my design error. Fixed in fd3d13e4c.

**Refusing to start was worse than the problem it solved.** The reviewer reproduced it: under a 1024 descriptor limit, the very process my change refused to start could hold **1020 files open simultaneously**. The semaphores are a *ceiling* on concurrent operations, not a reservation — a descriptor lives only for one operation — so a node whose ceiling exceeds the OS limit runs fine at any realistic load. My change converted a bounded, already-handled condition (an operation waits 25s, then returns `ServiceUnavailable`) into total unavailability from boot; and since the caller panics under a restarting unit, a 5-second crash loop.

It now **clamps** read/write concurrency proportionally to what the OS allows, keeping at least one of each, and startup prints the applied values with how to raise them. That makes EMFILE structurally impossible rather than merely predicted — which is what the existing use-system-limits mode already promises operators.

Two things from the verification worth recording:
- The trigger was narrower than first stated but real: only hosts whose **hard** limit is below the budget (an explicit `LimitNOFILE=1024` unit, a limits.conf-hardened image, `--ulimit nofile=1024:1024`). Low *soft* limits with generous hard limits — macOS, modern containerd, systemd defaults — are silently raised and were never at risk. Our own Hetzner fleet reports 1048576 and was unaffected.
- Separately, the pre-existing `filestore_use_system_limits` path shells out to `ulimit -u` — max user *processes* — not `-n`. So the 'safe mode' operators rely on is reading the wrong number entirely. Not this PR's defect; worth its own issue.

freemans13:
Review round 2 (regression re-hunt, Go correctness): 1 minor confirmed, fixed in db2cb4caa. No remaining way this makes a node worse off than before the PR — the regression lens specifically re-checked that after the clamp redesign.

**The only test of the raise path never ran.** My first test asks for a budget beyond the hard limit, which pins the soft limit at the hard limit and never restores it; the raise test then sees no headroom and skips. So the useful half of the feature — actually raising a too-low soft limit — had no live coverage on any platform. Each test now snapshots and restores the limit, since `Ensure` mutates a process-global resource and one test was quietly changing what every later one saw.

The verifier added something I would have got wrong: fixing only the leak would have flipped that test from *skipping* to *failing* on Linux, because the Go runtime starts a process at soft = hard−1, leaving no room for the raise the assertion expects. So the test now drives from a deliberately lowered soft limit rather than the ambient one. That was source-derived rather than reproduced (no Linux host available here), and worth flagging as such.

Also corrected a comment that credited the zero-budget check with handling Windows, when the error check does that.

**One refutation worth recording**, because it argued against a change I might otherwise have made: on hosts whose entire limit is ≤512, the clamp deliberately does nothing. Clamping to 1 read / 1 write there would serialise all block and subtree file I/O while the descriptor exhaustion still arrived from the unbounded side — sockets, gRPC, database pools — since `Headroom` exists precisely because those are not covered by these semaphores. Such a host is already doomed; crippling throughput on the way would be strictly worse than the current no-op. The PR leaves those hosts exactly as it found them.

freemans13:
**Review round 3 (convergence): 0 confirmed findings.** A fresh full-diff pass over the final state — the raise logic, the clamp arithmetic, the applied-limits accessor, the startup reporting, build tags and tests — turned up one candidate, which verification refuted.

The candidate: `//go:build unix` covers FreeBSD and DragonFly, but `syscall.Rlimit` uses `int64` fields there rather than `uint64`, so those two platforms no longer type-check. The compile fact is real and PR-introduced. The claimed consequence is not: `GOOS=freebsd go build ./...` already failed on main, because the BDK consensus library (the C++ script engine every validating path depends on) publishes no FreeBSD build. The node categorically cannot run there, CI is Linux-only with no cross-compile matrix, and nothing in the repo mentions either platform. That makes it build-tag hygiene rather than a defect — worth a one-line tidy if someone is in the file anyway, not worth a change here.

Three rounds, cumulatively: the original refuse-to-start design was killed in round 1 (a refused process could hold 1020 files open under a 1024 limit — semaphores are a ceiling, not a reservation), round 2 fixed a leaking rlimit that made the only test of the raise path silently skip, and round 3 found nothing new. Ready for human review.

ordishs (approved):
No P0 or P1 finding. The direction is right, and the clamp arithmetic holds on Linux with the shipped 768 + 256 defaults.

Verified by hand:

- `Ensure` — the `need` saturation, the hard-limit cap, the `min(Headroom, effective/2)` reserve, and the read-back fallback. Each one errs toward a smaller budget.
- `clampSemaphoreLimits` — the `int64` product and the one-permit floor.
- `resolveConcurrency` — the decision table against both halves of `filestore_use_system_limits`.
- The removal of the `ulimit -u` shell-out. It read the max-user-processes limit and scaled concurrency the wrong way.

One item worth a change before merge.

**util/fdlimit/fdlimit.go:65-84** — P2. The read-back does not catch the Darwin under-grant it names. Current macOS stores the requested soft limit verbatim, and `getrlimit` echoes it back. The kernel applies `min(rlim_cur, kern.maxfilesperproc)` only at descriptor allocation time. Measured on darwin/arm64: a raise to 300512 reads back as 300512, and `dup(0)` then returns EMFILE after 245757 descriptors. So `filestore_read_concurrency` = 200000 with `filestore_write_concurrency` = 100000 — both under `MaxSemaphoreLimit` — satisfies `budget >= total`, skips the clamp, and permits 300000 concurrent operations against an enforced cap of 245760. Older macOS defaults (`kern.maxfilesperproc` = 10240) open the same gap at 12288 + 4096. Either bound `effective` by `kern.maxfilesperproc` in a darwin build file, or drop the Darwin claim from the comment and from the doc comment on `TestEnsureReportsTheGrantedLimitNotTheRequest`, which mocks a kernel that lowers `cur`.

Six P3 items, all comments, messages, or coverage rather than behaviour:

- `stores/blob/file/file.go:374-376` — "Clamping instead makes EMFILE structurally impossible" overstates the guarantee. The write path holds two descriptors per permit: `createTempSibling` keeps the `os.OpenRoot` handle open at file.go:1350 and then opens the file at file.go:1367. `openFileWithFallback` releases the read permit at file.go:1132-1145 and then returns a live longterm-store reader. The clamped pair the new test pins (read 384, write 128, budget 512) therefore peaks at 640 descriptors, before any socket. `util/fdlimit/fdlimit.go:9-14` states the accurate version in the same change. settings.conf:375 carries the overstatement to operators.
- `cmd/teranode/daemon.go:66-73` — the comment asserts that a clamp always means the hard limit binds. `Ensure` breaks that on three exits: a refused `set`, a failed read-back, and an under-grant. The message then names `ulimit -Hn` when the soft limit or the kernel per-process cap is the real constraint. Consider returning the observed soft, hard, and granted values from `Ensure` and picking the advice from them. Send this line to the logger as well as stdout, so a clamp reaches log aggregation.
- `util/fdlimit/fdlimit.go:88-98` — the zero-budget rationale describes the caller wrongly. A zero budget with a nil error does not skip the clamp: `resolveConcurrency` falls through to `clampSemaphoreLimits`, which returns the one-read, one-write floor. The subtest "a zero budget still yields a usable store" already pins that. Same wording at fdlimit_test.go:153-157 and clamp_test.go:107-109.
- `stores/blob/file/file.go:1635` — "borrowing a single descriptor from the 512 that fdlimit reserves". A reserve of 512 needs an effective limit of 1024 or more, which yields a budget of 512 or more. In the budget-of-one case the reserve is one descriptor. Same figure at clamp_test.go:38.
- `stores/blob/file/file.go:1649-1651` — the package builds for no 32-bit target. `GOOS=linux GOARCH=arm go build ./stores/blob/file/` fails in `go-safe-conversion` and `go-wire`, and `GOARCH=386` fails identically. Keep the `int64` product; justify it as overflow safety rather than as support for a target.
- `stores/blob/file/file.go:385-393` — no test calls `InitSemaphores`, so the wiring is unverified. A mutation that assigns the configured values to `appliedReadLimit` and `appliedWriteLimit` instead of the clamped ones passes the whole suite. Only the two startup `Printf` lines report wrong numbers, so the blast radius is small. One test with non-default limits (600 and 200) would close it. Note two costs: `fdlimit` exports no seam, and the test conflicts with `TestAppliedSemaphoreLimitsBeforeInit` inside one binary.

gokutheengineer (approved):

Issue #1424, opened 2026-08-05
[HARDEN-0805] Detect merkle-tree mutation (CVE-2012-2459) so a duplicated-transaction block cannot be silently accepted or used to mark a real block invalid


freemans13:
## What can go wrong

A block must have exactly one transaction ordering that produces the merkle root in its header. The classic CVE-2012-2459 malleation abuses the "duplicate the last item when a tree level has an odd number of nodes" rule: by duplicating real transactions, an attacker can build a *different* transaction list that reproduces an honest block's merkle root. SV Node detects this — it sets a "mutated" flag and rejects a transaction list that reached an existing root by duplicating identical children.

Teranode's merkle builder implements the odd-level duplicate rule but never records whether a duplication combined two *identical real* children — which is precisely the mutation signal SV Node keys on. Teranode's only structural de-duplication happens at subtree-root granularity (in the top-level tree that combines subtree roots), not at transaction granularity within the final incomplete subtree, which is exactly where the odd-level duplication occurs.

If an attacker can shape the final subtree so a malleated transaction list collides with an honest block's merkle root, SV Node rejects it and Teranode may not — a consensus split. There is also a nastier variant: the malleated block could get one Teranode node to mark the *real* block invalid (cache poisoning), wedging that node off the honest chain.

## Where (verified on main @ 27a0178fd)

- `go-subtree@v1.4.2/merkle_tree.go` — `BuildMerkleTreeStoreFromBytes` at line 45 and the pair-hashing helper `calcMerkle` at line 136. The odd-child duplicate rule is right there (lines 144-150: when there is no right child, hash the left child concatenated with itself). Nothing anywhere in the function records that a duplication happened or that the two combined children were identical — there is no mutation flag.
- `model/Block.go` — the top-level merkle composition de-duplicates only on subtree *root* hashes: the `seen` map at lines 1798-1811 rejects a duplicate subtree root, but that is subtree-root granularity. It does not detect a duplicated *transaction* inside the final incomplete subtree, which is where the odd-level duplication lives.

## Why it matters

A silent disagreement with SV Node on block validity is Theme 1 — the most dangerous class — because the two fleets build different chains with no error on either side, and the offending block sits on disk so restarts do not help. It has not bitten because the power-of-two / complete-subtree structure confines the odd-duplication to the final incomplete subtree (shrinking the attack surface) and no adversary has probed it.

## How to reproduce / test

[...]

PR #1390, opened 2026-08-03
MvP-4692 Re-download corrupt block bodies instead of poisoning them

ctnguyen:
Resolves https://github.com/bitcoin-sv/teranode/issues/4692

## The problem

A block body that failed a body-derived check was returned as `ErrBlockInvalid`, which
persists `invalid=true`, cascades invalidation to descendants, and penalises the peer. A
re-announcement of the same hash was then rejected outright as *"block already exists as
invalid"*.

So a **single peer serving a corrupted body for a valid block hash could make the node
reject the honest chain's tip and every block built on it — permanently.** A transport
corruption from any relay was enough; no attacker was required.

The received body is not sufficient evidence to condemn the block hash. It cannot
distinguish an honest block corrupted or substituted in transit from an attacker's junk.
Condemning is catastrophic in the first case, while declining to condemn is cheap in the
second: producing such a block at all costs a full block proof-of-work at network
difficulty and forfeits the reward, since the block never connects.

This also violated an invariant the code relies on. `checkParentInvalid` documents that *a
parent marked invalid is invalid for a genuine consensus reason* — which is what licenses
cascading to children. A merkle mismatch broke that premise while still cascading.

## Alignment with svnode

This is the position svnode already takes. `CheckBlock` reports both a merkle-root mismatch
(`bad-txnmrklroot`) and a CVE-2012-2459 mutation (`bad-txns-duplicate`) through
`CorruptionOrDoS`, which applies a full DoS penalty to the connection **and** sets
`corruptionPossible`. That flag makes `AcceptBlock` and `InvalidBlockFound` skip
`ModifyStatusWithFailed`, so the block index is never marked failed, the block stays a
candidate, and no invalid-chain cascade runs.

svnode is explicit about the ordering, too:

> *All potential-corruption validation must be done before we do any transaction
> validation, as otherwise we may mark the header as invalid because we receive the wrong
> transactions for it.*

- [`src/validation.cpp`](https://github.com/bitcoin-sv/bitcoin-sv/blob/master/src/validation.cpp)
- [`src/consensus/validation.h`](https://github.com/bitcoin-sv/bitcoin-sv/blob/master/src/consensus/validation.h)

Note the mechanism lives entirely in block-index and chain-candidate bookkeeping, not in
mempool state, so it ports cleanly to Teranode, which has no mempool by design.

## What changed

A dedicated `ERR_BLOCK_CORRUPT` code, deliberately **not** matched by
[...]

PR #1504, opened 2026-08-06
fix(blockvalidation): add the missing duplicate-transaction check to the quick-validation path

freemans13:
Relates to #1424. **Scope reduced, and held as draft** — see below.

> **Draft pending #1390.** Round 3 established that this check on its own converts "the quick path silently accepts a doctored body" into "the quick path wedges the node on the honest hash", because the mutant and the honest block share a hash by construction and the catchup fallback condemns that shared hash. #1390 carries the error classification that makes the rejection safe. This branch is being kept correct and ready rather than merged ahead of it.

## What happened

A block is identified by the hash of its **header**. The header carries a single fingerprint (the merkle root) committing to the transactions, but the transactions themselves are not part of the hash. So a peer can hand us an honest header together with a *wrong body*, reusing the honest block's mining work at zero cost.

This PR originally set out to fix that whole problem. Partway through review it became clear that #1390 (MvP-4692) already fixes it, across a much wider surface — 53 files against this branch's 10 — and does so with a better design. This PR has therefore been cut down to the one gap #1390 does not cover, and the overlapping work has been dropped in its favour.

## Why #1390's design is the right one

#1390 introduces a dedicated `ERR_BLOCK_CORRUPT` code that `errors.Is(err, ErrBlockInvalid)` deliberately does **not** match, so a guard site that has not yet been taught about corruption declines to poison the hash rather than poisoning it by default. Its description names and rejects the alternative this branch had taken — tagging `ErrBlockInvalid` with side-band data — on the grounds that a single forgotten guard then poisons.

This branch is the evidence for that argument. Tagging means every site reacting to `ErrBlockInvalid` must be taught individually. Two such sites existed in `ValidateBlockWithOptions`. One was taught. The optimistic-mining branch — the one that actually handles gossip blocks, since `blockvalidation_optimistic_mining` defaults to `true` and is only disabled for the legacy and catch-up routes — was missed, and three further commits went by without closing it. #1390 makes that class of miss impossible by construction.

#1390 also already covers both open review findings against this branch: it disables optimistic mining on peer-served and catch-up paths behind a new `blockvalidation_optimistic_mining_peer_blocks` setting, and it strikes the serving peer **by peer ID** instead of through the hash-keyed lookup, which attributes to whoever *announced* the block. That second point matters more than it looks: the doctored block and the honest block share a hash by construction, so honest peers announcing the genuine block populate the same map entry an attacker's bad body would be charged against.

## What is left here

**The duplicate-transaction check on the quick-validation path** (`services/blockvalidation/quick_validate.go`).

The quick-validation pipeline runs **no** duplicate-transaction check of its own, and it is default-on for every block at or below the highest checkpoint. The legacy netsync route does check before handing a block over (`handle_block.go:251` and `:484`), and #1390 reclassifies those existing calls — but the catch-up route reaches the same pipeline with nothing in front of it, and neither PR adds a check inside the pipeline itself.

Without it, a CVE-2012-2459 mutant below the checkpoint is **committed**, not merely accepted. The merkle root matches by construction — the merkle tree duplicates the last node whenever a level holds an odd number of entries, so duplicating trailing transactions produces a different transaction list with a byte-identical fingerprint, and therefore the same header and the same block hash. `createAndSpendUTXOsForBatch` then absorbs the duplicate create via `ErrTxExists`, so nothing else stops it before `commitBlock`.

The check runs in `scanBatchForDuplicates`, per batch, **before that batch is applied** — before its UTXO creates and spends, and before its subtree files are written. Three things about that placement are load-bearing, and round 3 was right that the original placement had none of them:

- **Before the writes, not after.** It first ran in `validateSubtrees`, once every batch had already been created and spent. A rejected mutant therefore left the whole block's UTXOs written and, on default settings, locked — and nothing clears them. The only unlock pass is `commitBlock`'s, which a rejection never reaches, and the catchup fallback re-validates through subtree validation with `WithIgnoreLocked(true)`, so it spends *through* the flag without ever clearing it. An ordinary propagated transaction spending one of those outputs is later rejected `TX_LOCKED`. An earlier revision of this description claimed the state self-healed on the honest retry; that was wrong, because `tryQuickValidation` falls back to **normal** validation, which never calls `commitBlock`.
- **Before the fan-out, not inside `createAndSpendUTXOsForBatch`.** Both pipeline variants run the UTXO writes and the subtree-file writes concurrently in one errgroup, so scanning inside the former would still race the latter and leave a doctored subtree on disk under an honest subtree hash — which `subtreeWriteWorker` writes with `WithAllowOverwrite(true)`, after catchup's cleanup pass has run.
- **Against the applied list, not the slices.** It scans `batch.batchTxs`, the exact transactions about to be written. `block.SubtreeSlices` is the wrong list: on a retry it is deserialized from a local `.subtree` file (`buildSubtreeAndQueueWrite`), which can differ from the peer body being applied — so a duplicate in the peer's body could pass unnoticed, and symmetrically a stale local file could reject an honest peer's body.

Dedup state lives in a new `model.DuplicateTxScanner`, held per block so a duplicate split across two batches is still caught, and drawn from the same size-class pool or disk-backed table `Block.checkDuplicateTransactions` uses on the full validation path, honouring the same `DiskMapDirs` setting. The previous plain Go map held one entry per transaction for the whole block — hundreds of megabytes at 4M transactions, gigabytes at 100M, on precisely the path that exists to make large blocks fast.

**The two entry-point handlers** in `quickValidateBlock` and `quickValidateBlockAsync` re-wrap the rejection as `BlockInvalid` rather than passing it through bare. `(*Error).Is` (`errors/errors.go:179`) walks the wrapped chain, so an `ErrBlockInvalid` buried inside a `ProcessingError` is still matched by every `errors.Is` consumer — nothing is lost by wrapping. What matters is the error's *own* code: re-wrapped as `ProcessingError` the rejection additionally satisfies `ErrProcessing`, so any consumer classifying on the top-level code reads a consensus rejection as a transient fault. Re-wrapping as `BlockInvalid` keeps the correct code *and* restores the block hash and entry point, which a bare pass-through dropped — the callee's message carries only a txid, which on a stalled sync gives no block, height or peer to act on.

**The mutant-pair test** (`model/merkle_mutation_pair_test.go`), which #1390 has no equivalent of in `model/`. It builds a real CVE-2012-2459 mutant pair from actual transactions through the pinned go-subtree, rather than hand-faked hashes, and hangs both bodies off a **single shared header**, so they necessarily carry the same block hash — one proof-of-work, two different transaction lists. It pins that `CheckMerkleRoot` **passes on both** — so the fingerprint alone provably cannot catch this — and that `Block.Valid` rejects only the mutant. The body-classification assertions are gone; everything else stands.

## Dropped in favour of #1390

The `NewBlockInvalidBodyError` / `IsBlockInvalidBody` error kind and its tests; the body-fault reclassifications in `model/Block.go` and `model/check_duplicate_txs.go`; the `storeInvalidBlock` carve-out and the notify changes in `BlockValidation.go`; and the service-level no-persist test. All of it is covered by #1390 with the stronger error model.

## Worth recording

The ticket's central claim was wrong. #1424 said Teranode never de-duplicates at transaction level. It does — `Block.Valid` runs an unconditional scan. A probe enumerated 136 distinct same-root mutations across subtree capacities {2,4,8,16} and honest transaction counts 2..40; that scan flagged all 136, and so did SV Node's check. Full agreement, so on the main validation path Teranode already matched SV Node exactly. The real gap was the quick path, which is what remains here.

#1424 should be closed against #1390 rather than this PR, since #1390 is what addresses the cache-poisoning variant the ticket mentions in passing.

## Testing

`go build ./...`, `go vet` on the touched packages, `make lint` / `golangci-lint` (0 issues), `scripts/check_filenames.sh` (clean), `go test ./model/...` and `go test ./services/blockvalidation/...` — all pass.

Every test here was mutation-tested rather than merely run.

`go build ./...`, `go vet`, `golangci-lint` (no findings in the changed files), `scripts/check_filenames.sh`, `go test ./model/... ./services/blockvalidation/...` (all pass), and `go test -race` over the changed paths.

- Removing the `scanBatchForDuplicates` calls makes both sub-tests of `TestQuickValidateBlock_DuplicateIsBlockInvalidNotProcessing` fail **for the right reason** — the duplicate assertion is what breaks, falling through to the merkle check.
- Deleting **both** entry-point handlers makes the same tests fail specifically on `require.False(errors.Is(err, ErrProcessing))`, confirming that is the load-bearing assertion and that the `ErrBlockInvalid` one alone would not catch the regression.
- `TestScanBatchForDuplicates` covers the seam directly, including the **cross-batch** case — a mutant whose repeat sits in a later subtree than the original, which a batch-local check would miss.
- The end-to-end test additionally asserts that **nothing was written** for the duplicated transaction and that **no subtree write jobs were queued**. Those are the assertions that pin the new placement rather than merely the check's existence; they fail if the scan moves back after the writes.

Two test defects found and fixed during review, both mine:

- The pass-throughs had **zero** coverage — deleting both left the whole package green. Now covered end to end.
- The round-2 timestamp fix hollowed out the merkle-pair test. With the roots already asserted equal and every other header field identical, the honest-vs-mutant hash comparison could not fail unless the root assertion already had, and it ground the same nonce twice to reach it. One shared header now states the property directly, and a fixed timestamp makes the test deterministic rather than merely non-flaky.

## Limits

Not verified end to end — no mutated body was served to a running node over the catch-up path. Everything below is read from the code and pinned by tests, not observed on a live node.

**The `#1390` dependency is the real one, and it is why this is a draft.** The mutant and the honest block share a block hash by construction. `tryQuickValidation` (`catchup.go:1396-1426`) swallows the rejection, deletes the `.subtree` files and falls back to normal validation, which condemns that shared hash — so a single doctored body from a single peer, at no proof-of-work cost, can stall sync at that height until an operator intervenes. Rejecting the body is still strictly better than committing it, but the rejection is only *safe* once #1390's classification distinguishes a corrupt body from an invalid block. Merging this ahead of #1390 trades a silent-accept for a wedge.

**Known gaps left open, deliberately:**

- `validateSubtrees` has two other body-fault returns that are still mis-typed. The subtree size mismatch is a bare `NewProcessingError` wrapping nothing, so `errors.Is(err, ErrBlockInvalid)` is outright `false`. The merkle-root mismatch wraps a genuine `BLOCK_INVALID` in `ProcessingError` — the same demotion this PR fixes for duplicates, left in place for merkle. Both are pre-existing and belong with #1390's classification work rather than here.
- Both quick-path entry points guard on `len(block.Subtrees) > 0` and otherwise go straight to `commitBlock`. A zero-subtree block is therefore committed with no `validateSubtrees` at all — so no dedup, and **no `CheckMerkleRoot`**, meaning no binding of body to header. Dedup is vacuous for such a block so this is not a gap in this fix, but the merkle gap is real and pre-existing. Worth a separate issue.
- The scan runs after the batch's transactions have been read and extended, not before. Reading is not a store mutation, so nothing is left behind by a rejection, but a mutant still costs the read and extend work before it is rejected.

**Unrelated pre-existing bug found while writing the tests, reported not fixed:** the sequential quick-validation variant (`blockvalidation_subtree_batch_prefetch_depth=0`, a documented supported value — *"Set to 0 for memory-constrained or debugging scenarios"*) panics with an index-out-of-range. `processSubtreeBatch` does not allocate `SubtreeProcessingBatch.fullSubtreeExists`, which its pipeline counterpart `prefetchSubtreeBatch` does, so `writeSubtreeFilesForBatch` indexes a nil slice on the first subtree it writes. It reproduces on `main` and is nothing to do with this change, so the end-to-end tests run on the default pipeline variant. All three variants call `scanBatchForDuplicates` before their fan-out, so the floor itself holds on each. Filing separately.

https://claude.ai/code/session_011uBQ1ehZrbX9mSo1SznbHt


freemans13:
**Review round 1: 4 confirmed of 12 raised.** Three fixed here, one split out. Marked draft while the rounds run.

**Fixed — body faults that run before the merkle check were still condemning the honest hash.** The classification stopped short of the ordering property it depends on. SV Node runs the merkle-root and duplicate checks first and says why (`src/validation.cpp:5630`): *"All potential-corruption validation must be done before we do any transaction validation, as otherwise we may mark the header as invalid because we receive the wrong transactions for it."* Teranode inverts that order, so several checks ran before anything tied the body to the header. Now reclassified: the coinbase shape checks, the BIP34 coinbase-height checks, and the subtree nil/length rules in `GetAndValidateSubtrees`.

The last pair was the sharpest catch — directly inconsistent with this PR's own reasoning, which had already reclassified the equivalent shape rules inside `CheckMerkleRoot` one step later.

Worth noting the reviewer's headline attack was refuted while verifying it. Flipping the height push in the coinbase does reach the rejection, but the SQL store's own `validateCoinbaseHeight` aborts the invalid-block insert before the Invalid flag is read, so nothing is persisted. The coinbase *shape* variant does land, though: `IsCoinbase()` checks more than the outer guard does, so rewriting the coinbase input while leaving its length and height push intact passes every gate and reaches the rejection. I classified the height sites anyway — relying on one store incidentally refusing the write is not a guarantee.

**Fixed — the quick path's new duplicate check had no test.** A coverage profile showed the rejection branch at zero: the existing tests execute the call but never drive a duplicate through it, so deleting the check left the whole package green. That was the most novel change in the PR and the least pinned. Now tested at the `validateSubtrees` seam all three pipelines share, and mutation-verified — with the check removed the duplicate reaches the merkle check instead and the test fails.

**Split out to #1507 — two routes classification cannot fix.** Both were confirmed and both are real; neither is fixable by tagging errors, which is why they are not patched here.

Transaction validation runs before anything binds the body to the header, so a peer can serve the honest header plus a fabricated subtree list containing one transaction that fails validation and get the honest hash condemned. Reclassifying that would be *wrong*: a body that does reproduce the header's merkle root and contains an invalid transaction is genuinely invalid, and recording it is correct. The distinction isn't in the error — it's in whether the binding has happened yet.

Optimistic mining, which is default-on for new tips, stores the block before validating the body. Suppressing the invalidation there would be worse than the wedge: the attacker's body is already stored under the honest hash, so skipping it leaves a doctored body in the chain as an accepted block. There is no rollback; `InvalidateBlock` is the only mechanism.

The general cure for both is SV Node's ordering — verify the merkle root and duplicates before transaction validation and before the optimistic store. That needs a design decision (it means loading subtree contents synchronously at receipt time, exactly the work optimistic mining defers to start mining sooner), so it is #1507 rather than scope creep here. Catchup and the legacy route both disable optimistic mining, so this PR does cover the sync path where peers serve bodies in bulk.

Refuted and not changed, briefly: the dedup scan's cost on the IBD path (measured, immaterial), the claim that skipping the persist drops the only peer-punishment signal (it doesn't — the error severity that drives banning is deliberately preserved), and that catchup would stop trying other peers.

Verification: 1021 tests in `errors`/`model`, 1030 in `blockvalidation`, vet clean, `make lint` 0 issues.

freemans13:
**Review round 3: 1 confirmed of 5 raised** — and inside it, a defect I had introduced in round 2's own fix.

**The description had drifted badly from the code, and that mattered more than it sounds.** It still said eight rejection sites when sixteen had been reclassified, and it did not mention the notification commit at all. The reviewer's point about *which* sites were missing is the sharp one: the undocumented group is precisely the pre-merkle-check group, and "which checks run before the body is bound to the header" is the entire argument this PR rests on. A description that omits exactly those sites undercuts the reviewability of the change. Rewritten to state all sixteen in two named groups, both notification commits, the refreshed counts, and — plainly — that the notification commits have no test of their own.

**The real find, buried in that same report: restoring the notification dropped the revalidation guard.** The persist it replaced was wrapped in `if !opts.IsRevalidation`; my replacement was not. So an operator re-checking a stored block would publish to the invalid-block topic and cost whichever peer is recorded against that block a ban score — for a check the operator themselves asked for. Guarded at all three sites now. That was mine, introduced two commits ago while fixing something else, and it is exactly the kind of thing that only shows up when someone re-reads the final state rather than the latest diff.

One refutation worth recording, because it argued against a change: the claim that the two quick-path pass-throughs are inert because `errors.Is` already walks through the processing wrapper. The unwrapping does find the error either way, but the outermost type is what the caller classifies on, so returning it unchanged is not redundant — and the legacy unified route has no fallback, which is where that matters.

Verification after the guard fix: 1030 tests in `blockvalidation`, `make lint` 0 issues.

freemans13:
**Review round 4: 0 confirmed. Converged** — marking ready for review.

Two candidates, both refuted, both worth a line because each argued *against* changing something.

The description filed two of its reclassified sites in the wrong ordering group (the first-subtree-empty and coinbase-placeholder checks run before the merkle check, not at it). Genuinely wrong, and now corrected — but the verifier's reason for refusing it as a finding is the interesting part: the error ran in the *conservative* direction. Group two is labelled "the group that matters most to the argument", so misfiling those two into group one understated their importance rather than overstating it. No reader was led to accept anything unsafe. The verifier also caught the reviewer's own supporting claim being wrong, which is why the fix is a reword rather than a straight move.

The second: catchup's log line still tells an operator the block was "already stored as invalid", which is now untrue for the de-persisted classes. Real, but left alone deliberately — it was already untrue on `main` for peer-served duplicate transactions on a route this PR doesn't touch, and `catchup.go` isn't in the diff. Widening the set of cases a stale sentence is wrong about isn't a reason to pull an unrelated file into a consensus change. Worth a tidy by whoever next edits that file.

Four rounds, cumulatively: round 1 found that the classification stopped short of the ordering property it depends on, and split two unfixable-by-classification routes out to #1507. Round 2 found that peer punishment on the gossip path did *not* survive as claimed. Round 3 found that fixing round 2 had dropped a revalidation guard, and that the description had drifted from the code in exactly the places a reviewer needs. Round 4 found nothing.

Final verification: 1021 tests in `errors`/`model`, 1030 in `blockvalidation`, `go vet` clean, `make lint` 0 issues.

icellan:
Same fix as #1390 — the 15 reclassified sites in `model/Block.go` are line-for-line identical (this PR rewrites them to `NewBlockInvalidBodyError`, #1390 to `NewBlockCorruptError`), and 6 of the 10 files here are shared, so these can't both merge without hand-resolving every site. The mechanisms fail in opposite directions: keeping `ERR_BLOCK_INVALID` means a guard that doesn't know about the tag poisons, while #1390's dedicated code is deliberately not matched by `errors.Is(err, ErrBlockInvalid)`, so an unaware guard declines to poison. #1390 also covers the optimistic branch, catchup and legacy sync, and strikes by peer ID. The one thing unique to this PR is the quick-path duplicate scan, which is a real gap. Suggest rebasing onto #1390 and keeping only that.

P0

- `services/blockvalidation/BlockValidation.go:1835` — `IsBlockInvalidBody` has exactly one production call site (1976, the non-optimistic branch). The optimistic-mining handler is default-on (`settings/blockvalidation_settings.go:27`) and still gates on bare `errors.Is(err, ErrBlockInvalid)`, which the new constructor keeps true, so `markBlockAsInvalid` → `InvalidateBlock` still condemns the honest hash and cascades via `checkParentInvalid` on the default gossip path. The new service test sets `OptimisticMining = false` to reach the fixed branch, so nothing pins the default configuration.
- `services/blockvalidation/BlockValidation.go:2414` — the `reValidateBlock` worker has the same unguarded `InvalidateBlock`. Any transient failure on the optimistic path (1813, 1856, 1889) routes the block into `ReValidateBlock`, so a doctored body that hits a transient error first gets condemned on the retry instead.
- `services/blockvalidation/quick_validate.go:763` — the new check runs after `createAndSpendUTXOsForBatch` on all three pipelines, against the contract in `model/check_duplicate_txs.go:9-14` ("Every such path MUST call it before creating or spending UTXOs"). `AssignBlockID`, UTXO creation with `WithMinedBlockInfo`, parent spends and `SetMinedMultiChunked` all land before the reject, and neither caller rolls them back — catchup deletes only the `.subtree` files (`catchup.go:1414-1424`), the legacy route (`Server.go:1545`) cleans up nothing. The PR body acknowledges the ordering; the mined-stamp residue and the wasted create+spend are the concrete cost.
- `services/blockvalidation/BlockValidation.go:1976` — dropping the persist drops the only rejection memo. `GetBlockExists` stays false, so the short-circuit at 1485-1504 never fires and `runOncePerBlock` dedups for only 100ms (1388-1391). Re-announcing the same header+body drives the full path each time: subtree fetch from the peer, subtree validation, `block.Valid`, merkle/duplicate scan.

P1

- `services/blockvalidation/BlockValidation.go:1542`, `:1982` — `kafkaNotifyBlockInvalid` is entirely inside `if u.invalidBlockKafkaProducer != nil` (2175), and the producer is left nil when `kafka_invalidBlocks` is unset (default empty) or initialisation fails (344 logs and continues). At the three converted sites that means no ban score, no persisted row, and not even the `Warnf` that `storeInvalidBlock` always emitted. Where it does fire, `p2p.processInvalidBlockMessage` (`services/p2p/server_helpers.go:743`) resolves the hash through `blockPeerMap`, charging the peer that announced the hash rather than the one that served the body — and with nothing persisted, each replay repeats the charge. #1390's peer-ID attribution is the fix.
- `services/blockvalidation/quick_validate.go:786` — `CheckSubtreeSlicesForDuplicateTxs` presizes a plain `map[chainhash.Hash]struct{}` to the block's full tx count, on the pipeline that deliberately streams batch-by-batch and closes each batch. `Block.Valid` uses the pooled/disk-backed txmap (`settings.Block.DiskMapDirs`) for the same check; this call site uses neither.
- `services/blockvalidation/quick_validate.go:774` — still `NewProcessingError` for the subtree-shape rule that `model/Block.go:1651` now classifies body-attributable, so punishment and retry behaviour depend on which route the block took. #1390 reclassifies this one and the merkle-root shadowing next to it.

Smaller: nothing asserts `kafkaNotifyBlockInvalid` actually fires at 1542/1573/1982 — `util/kafka/kafka_producer_async_mock.go` already exports `NewKafkaAsyncProducerMock()` and is used by `services/p2p/invalid_blocks_consumer_test.go` among others, so the stated blocker doesn't hold. And the `ErrBlockInvalid` pass-through at `quick_validate.go:320` is dead: `tryQuickValidation` (`catchup.go:1396`) checks only `ErrBlockIncomplete` and returns `(true, nil)` for everything else, discarding the error.

icellan (changes requested):
Round 3, at `a8d3d8eb1`. The comment fixes landed and the merkle-wrap contradiction is gone. Nothing structural moved, and the timestamp fix hollowed out the test it was fixing.

P0

- `services/blockvalidation/quick_validate.go:796` — still the dominant issue, unchanged since round 2. The mutant's block hash equals the honest block's by construction (your own model test proves exactly that), and `tryQuickValidation` (`catchup.go:1396-1426`) discards the error, deletes the `.subtree` files and returns `tryNormalValidation=true`. Normal validation then reaches `storeInvalidBlock` on that shared hash and `reportCatchupMalicious` on the peer, so every later announcement of the genuine block is short-circuited and sync is stuck at that height until an operator intervenes. One doctored body from one peer, no PoW cost. Don't merge this ahead of #1390: without its classification the check converts "quick path silently accepts a doctored body" into "quick path wedges the node on the honest hash".
- `services/blockvalidation/quick_validate.go:789` — the rejected block's UTXOs stay locked with nothing left to clear them. `SetLocked(false)` on this path has exactly one caller, `commitBlock` -> `unlockSubtreeTransactionsIfNeeded` (`:1054`), which the error path skips; the catchup fallback spends through subtreevalidation with `validator.WithIgnoreLocked(true)` (`check_block_subtrees.go:652`) and never unlocks; and the validator's own `SetLocked(false)` (`Validator.go:1043`) is its two-phase commit for transactions it created, not a heal for these. Block-based spends ignore the flag, so this surfaces later: an ordinary propagated tx spending one of those outputs is rejected `TX_LOCKED`.

P1

- `services/blockvalidation/quick_validate.go:797` — the check reads `block.SubtreeSlices`, which on the `fullSubtreeExists` branch is deserialized from the local `.subtree` file (`:106-117`, `:978-990`), not the peer's `.subtreeToCheck` node list that `createAndSpendUTXOsForBatch` actually created and spent. A mandatory floor that can scan a different transaction list than the one applied to the store isn't a floor: a duplicate in the peer's body can pass unnoticed and commit, and symmetrically a stale local file carrying a duplicate rejects an honest peer's body.
- `services/blockvalidation/quick_validate.go:324` — `g.Wait()` only waits for jobs to be queued on `writeJobsChan`. The cleanup `Del` loop at `catchup.go:1417-1424` has no drain barrier and `subtreeWriteWorker` writes with `WithAllowOverwrite(true)` (`:78`), so a queued doctored subtree can be written after the delete. Because the mutant subtree hashes to the honest root, the survivor is a doctored body under a legitimate subtree hash — read back by the immediate fallback (so the block is condemned even from an honest peer) and served onward by the asset service.
- `services/blockvalidation/quick_validate.go:238` — the legacy unified route has no cleanup counterpart to catchup's `Del` loop. A rejection returns straight to `Server.processBlockFound`, leaving the `.subtree` files written by `writeSubtreeFilesForBatch` (`:958-990`) in the store under honest hashes plus the locked, half-spent UTXOs; the next attempt finds `fullSubtreeExists` and reuses the stored file instead of the peer's data.
- `services/blockvalidation/quick_validate.go:796` — `make(map[chainhash.Hash]struct{}, totalNodes)` over every node in the block, held while all `SubtreeSlices` are resident. `model/Block.go:1000-1010` routes the same check through `NewDiskTxMapUint64` or the size-class pool for exactly this reason. ~200MB at 4M txs, GBs at 100M — on precisely the blocks the quick path exists to accelerate.
- `services/blockvalidation/quick_validate.go:239` — the pass-through drops the `ProcessingError` wrap that carried the block hash and the `[quickValidateBlock]` tag, and the callee's message contains only a txid. A stalled legacy sync logs a duplicate txid with no block, height or peer.
- `model/merkle_mutation_pair_test.go:170` — the flake fix removed the test's content. With the timestamp shared and `honestRoot.IsEqual(mutantRoot)` already asserted at `:129`, the two headers are bit-identical by construction, so this assertion cannot fail unless `:129` already did — and it grinds the same proof-of-work twice to reach it. Build ONE header from the shared root and hang both bodies off it, the shape `model/check_duplicate_txs_test.go:190-206` already uses.

Smaller: `model/check_duplicate_txs.go:14` still states the precondition as absolute while this call site documents 14 lines of why it doesn't follow it — reconcile the callee doc or the call site, don't leave both. And neither new test enters `quickValidateBlock`/`quickValidateBlockAsync`, so deleting either pass-through branch leaves the package green.

Most of the above collapses into one change: run the scan before `createAndSpendUTXOsForBatch`, against the list actually applied to the store. That settles the locked UTXOs, the stale-slice mismatch, both cleanup gaps and the callee contract at once. Then either make `tryQuickValidation` act on the verdict or drop the two pass-throughs — one is documented as inert today and neither is tested.

freemans13:
**Round 3: 7 confirmed of 9, and the synthesis was right.** Taken, at `72c095214`. Marked draft.

**The synthesis was the whole review.** Running the scan before `createAndSpendUTXOsForBatch` against the applied list does settle four of the findings at once, and I've done that rather than patch them individually. One correction to the shape: putting it *inside* `createAndSpendUTXOsForBatch` is not sufficient, because both pipeline variants run the UTXO writes and the subtree-file writes concurrently in the same errgroup (`:604-620`, `:737-753`). Scanning inside the UTXO half still races the write half. It now runs in `scanBatchForDuplicates`, before the fan-out, so P1-2's queued-write window closes too — nothing doctored is ever queued rather than being queued and then racing a delete.

**P0-2 confirmed, and my description was wrong about it.** I claimed the locked UTXOs self-healed on the honest retry. They don't: `tryQuickValidation` returns `tryNormalValidation=true`, and the normal path never calls `commitBlock`, which is the only caller of `unlockSubtreeTransactionsIfNeeded` → `SetLocked(false)` on this path. Your `WithIgnoreLocked(true)` point (`check_block_subtrees.go:652`) is the reason the fallback doesn't clear them either. Moot now, since nothing is written before the rejection, and the end-to-end test asserts exactly that — the duplicate's txid must be absent from the store afterwards.

**P1-1 confirmed and it was the sharper of the two.** `SubtreeSlices` on the `fullSubtreeExists` branch is deserialized from the local `.subtree` file (`:106-117`, `:977-987`), so the check could scan a different list than the one applied — in both directions. It now scans `batch.batchTxs`.

**P1-4 confirmed.** Replaced with a `model.DuplicateTxScanner` drawing from the same size-class pool or disk-backed table `Block.checkDuplicateTransactions` uses, honouring `DiskMapDirs`. Held per block, so the cross-batch duplicate is still caught — there's a test for that case specifically, since a batch-local check would see two clean batches.

**P1-5 confirmed.** Re-wrapped as `BlockInvalid` rather than passed through bare, which keeps the correct top-level code *and* restores the block hash and entry point.

**P1-6 confirmed — you were right that my own fix hollowed it out.** With the roots asserted equal and every other field identical, the hash comparison couldn't fail unless the root assertion already had, and it ground the same nonce twice. Rebuilt on one shared header per your suggestion. Timestamp is now fixed rather than sampled, so it's deterministic rather than merely non-flaky.

**Callee contract reconciled.** `CheckSubtreeSlicesForDuplicateTxs`'s doc now says which shape it suits and points batched callers at the scanner, so the precondition and the call site agree instead of contradicting.

**P0-1 accepted as the reason not to merge.** I'm not contesting it — this is now a draft pending #1390, and the description says so. Rejecting a doctored body is still better than committing it, but the rejection is only *safe* once the classification distinguishes a corrupt body from an invalid block, and merging ahead of that trades a silent-accept for a wedge.

**One correction on the closing line:** the pass-throughs *are* tested now — `TestQuickValidateBlock_DuplicateIsBlockInvalidNotProcessing`, added in `04b383f87`, after this review was written. Your point stood when you made it: deleting both left the package green. The load-bearing assertion turned out to be the negative one, since `(*Error).Is` walks the chain and `errors.Is(err, ErrBlockInvalid)` stays true even when re-wrapped. Verified by mutation in both directions.

**Unrelated bug found while writing the tests, filing separately:** the sequential variant (`blockvalidation_subtree_batch_prefetch_depth=0`) panics on index-out-of-range. `processSubtreeBatch` never allocates `SubtreeProcessingBatch.fullSubtreeExists`, which `prefetchSubtreeBatch` does, so `writeSubtreeFilesForBatch` indexes a nil slice. Reproduces on `main`.

Verification: `go build ./...`, `go vet`, `golangci-lint` (nothing in the changed files), full `model` + `blockvalidation` + `catchup` suites, and `-race` over the changed paths. Both new placements mutation-tested — removing the scan calls fails the end-to-end tests on the duplicate assertion, falling through to the merkle check.

icellan (changes requested):
Round 4, at `04b383f87`. The pass-through coverage landed and holds up — the new tests do drive `quickValidateBlock`/`quickValidateBlockAsync` with a real sqlitememory UTXO store, so the duplicate genuinely survives create+spend to reach the guard. That one is closed.

The 42-line production change is byte-identical to round 3, so both structural asks from that review are still open, and two things got worse.

P0

- `services/blockvalidation/quick_validate.go:796` — unchanged and still the blocker. Re-verified on the new merge-base `9de8693c2`: the body-fault classification is not on main. `tryQuickValidation` (`catchup.go:1412-1426`) discards the verdict and falls back, normal validation re-detects the duplicate, `storeInvalidBlock` (`BlockValidation.go:1953-1955`) records the *honest* block hash invalid, and `checkParentInvalid` (`:2135`) cascades to every descendant. One packet, no PoW cost, manual `RevalidateBlock` to recover. Land #1390 first; without its classification this check converts "quick path silently accepts a doctored body" into "one packet permanently wedges the node".
- `services/blockvalidation/quick_validate.go:324` — the doctored `.subtree` now demonstrably outlives the rejection. The legacy unified route has no cleanup counterpart to catchup's `Del` loop, and on the catchup route `g.Wait()` only waits for jobs to be *queued* on `writeJobsChan`, so `subtreeWriteWorker` (`WithAllowOverwrite(true)`, `:75-83`) can write after `catchup.go:1417-1424` has already deleted. `model/Block.go:1619` prefers `FileTypeSubtree` over `FileTypeSubtreeToCheck`, so every later validation of the honest block deserializes the attacker's duplicated body from local disk and condemns it. The wedge persists after the attacker disconnects.
- `services/blockvalidation/quick_validate.go:790` — the comment now acknowledges the locked-UTXO leak rather than fixing it. Default `quickValidateSkipsUtxoLock=false`, so `createAndSpendUTXOsForBatch:1334` creates every tx `WithLocked(true)`; `SetLocked(false)` has exactly one caller, `commitBlock` -> `unlockSubtreeTransactionsIfNeeded` (`:367`), which this path skips; the catchup fallback and subtreevalidation spend with `WithIgnoreLocked(true)` (`:1426`, `subtreevalidation/check_block_subtrees.go:652`) and never unlock. A propagated transaction later spending one of those outputs is rejected as locked, with no operator-visible cause.
- `services/blockvalidation/quick_validate_duplicate_test.go:147` — your test comment is right that the sequential-pipeline panic is pre-existing; I verified it at `9de8693c2` (`processSubtreeBatch` never allocates `SubtreeProcessingBatch.fullSubtreeExists` while `prefetchSubtreeBatch` does, so `:1488` indexes a nil slice). But `blockvalidation_subtree_batch_prefetch_depth <= 0` is a value the settings docs recommend for memory-constrained use, and on it every below-checkpoint block panics in the block-found worker — so the CVE floor this PR adds is unreachable on that configuration. Fix it here or open an issue; a test comment isn't a resolution.

P1

- `services/blockvalidation/quick_validate.go:797` — still scanning `block.SubtreeSlices`. `findLocalSubtreeFile` prefers `.subtreeToCheck` (`find_local_subtree.go:25-31`) for the data that drives `extendBatch` and `createAndSpendUTXOsForBatch`, while `fullSubtreeExists` is set independently from `Exists(FileTypeSubtree)` (`:904-906`). Since a CVE mutant's subtree roots equal the honest ones, any locally present `.subtree` for those roots makes `:106-117` / `:977-987` overwrite `block.SubtreeSlices` with the honest body — the floor then passes on honest data while the attacker's create+spend has already been applied, and the block commits.
- `services/blockvalidation/quick_validate.go:796` — unpooled `make(map[chainhash.Hash]struct{}, totalNodes)` (`model/check_duplicate_txs.go:51`) over every node in the block, held for the whole scan, where `model/Block.go` routes the same check through `NewDiskTxMapUint64` or the size-class pool precisely to avoid that. On a memory-capped catchup pod that is a GC storm or an OOM kill on exactly the blocks the quick path exists to accelerate.
- `services/blockvalidation/quick_validate.go:238` — the pass-through returns the callee's text unwrapped, so what reaches gRPC (`Server.go:1343-1345`) and netsync (`handle_block.go:328-334`) names only a txid: no block hash, no height, no call site. Every other error leaving `quickValidateBlock` carries the block hash.
- `model/merkle_mutation_pair_test.go:176` — with the timestamp now shared, both headers are field-for-field equal once `honestRoot.IsEqual(mutantRoot)` is asserted at `:135`, so this assertion cannot fail independently and the second nonce grind buys nothing. Build ONE header from the shared root and hang both bodies off it, the shape `model/check_duplicate_txs_test.go:190-206` already uses.

Smaller: `validateSubtrees`' doc comment (`:770`) still describes only subtree sizes and the merkle root, though it is now the single seam carrying the CVE floor for all three pipelines. And the CVE rationale is now restated in full for the fifth time in the tree (`model/check_duplicate_txs.go:9-41`, `netsync/handle_block.go:237-244` and `:472-484`, the 14-line block here, both new test doc comments), with the `(*Error).Is` paragraph duplicated verbatim between `:231-239` and the test at `:238-249` — point at the helper's doc comment instead.

Same two asks as round 3: land after #1390, and move the scan ahead of `createAndSpendUTXOsForBatch` against the list actually applied to the store. That second move settles the locked UTXOs, the stale-slice mismatch, both cleanup gaps and the callee-contract contradiction in one change.

PR #1508, opened 2026-08-06
feat(aerospike): configurable write CommitLevel, relax pruner deletes to COMMIT_MASTER

icellan:
`writePolicy.CommitLevel` and `batchWritePolicy.CommitLevel` were hard-coded to `COMMIT_ALL` with no way to change them. `COMMIT_ALL` is also the aerospike client default, so `aerospike_useDefaultPolicies` did not relax it either.

## Configurable commit level

`aerospike_writePolicy` gains a `CommitLevel` query parameter (`all` | `master`, case-insensitive, defaults to `all`), parsed alongside the other write-policy tunables. It governs both `GetAerospikeWritePolicy` and `GetAerospikeBatchWritePolicy` — there is no separate batch-write policy URL, and it is the same durability decision.

The package global holding it defaults to `COMMIT_ALL` via its zero value (`COMMIT_ALL CommitLevel = iota`), so callers running before the policy URL is parsed keep full durability rather than silently relaxing.

## Pruner always uses COMMIT_MASTER for removals

Independent of the setting. Pruning only drops records already provably safe to drop, and it is idempotent: a replica that misses the write is re-pruned on the next partition scan, so waiting for a full-replication ACK buys no correctness.

The pruner's delete paths never went through the util helpers anyway — they built raw `NewBatchDeletePolicy()` / `NewBatchWritePolicy()` inline, silently inheriting `COMMIT_ALL`. A settings knob on the util helpers alone would have done nothing for the pruner.

**Parent updates deliberately keep `COMMIT_ALL`** — they mutate records that survive the prune. Mixing commit levels inside the single `BatchOperate` of `executeBatchCleanupCombined` is safe because `CommitLevel` is carried per record (`batch_attr.go:219/275/306`), not on the shared `BatchPolicy`.

This sits cleanly alongside the native-op reroute in #1366, which changed only the parent-update path (`buildAddDeletedChildrenRecord` → native `subOpAddDeletedChildren`). Child deletions still route through the shared builder.

## Also

Deduplicates the child-deletion record building, which was copy-pasted between `executeBatchDeletions` and `executeBatchCleanupCombined` (−35 lines in `pruner_service.go`). The extraction is what makes the durability assertion testable against the records actually put on the wire, rather than against the policy constructors in isolation.

## Caveat

`CommitLevel` only has an effect on namespaces with `replication-factor` > 1. If the utxo namespace runs Aerospike in strong-consistency mode, the server may ignore client `COMMIT_MASTER` — worth confirming against the live namespace config before flipping the global to `master` and expecting a throughput win. The pruner override is correct either way.

## Testing

Written test-first. New coverage:

- `getQueryCommitLevel` parsing: `all`, `master`, case-insensitivity, absent, empty, and rejection of unknown/numeric values
- the parsed level reaching both policy builders, and `useDefaultPolicies` bypassing it
- the deletion batch records actually carrying `COMMIT_MASTER`, with TTL touches keeping `UPDATE_ONLY` + `Expiration=1`
- a guard that parent-update policies stay `COMMIT_ALL`

Run on the rebased branch: `go build ./...`, `go vet`, `golangci-lint` (0 issues), `go test` on `util`, `settings` and the full `stores/utxo/aerospike/...` suite, plus `-race` on the changed packages. The container-backed pruner tests were confirmed to actually execute rather than skip.


freemans13 (approved):

ordishs (approved):
Mechanism looks right — the two-branch mapper, the shared `buildDeletionBatchRecords`, and
leaving parent updates at `COMMIT_ALL`. Three things about the safety argument this rests on
("a missed removal is re-found by the `delete_at_height` scan next session and re-pruned"):

**P1 — `settings/settings.go:590`: relaxation is default-on, unconditional, unlogged**
- `getBool("pruner_relaxRemovalCommitLevel", true, ...)`, `settings.conf:589` and the
  `default:"true"` tag all agree — every existing rf>1 deployment loses full-replica ACK on
  upgrade with no opt-in.
- `NewService` applies it with no replication-factor check and no startup log of the effective
  commit level, so the operator's first signal is the symptom (records reappearing, storage not
  shrinking) with nothing in their own config to point at.
- The setting's own longdesc says it is "only meaningful on namespaces with replication-factor
  > 1" — i.e. the risky value is the default and the safe value is opt-in.
- Suggest defaulting to `false` (or gating on rf) plus one startup log line.

**P1 — `stores/utxo/aerospike/pruner/prune_policies.go:55-56`: extra records have no `delete_at_height`**
- The relaxed level is applied to every key in the batch, including the pagination keys derived
  from `totalExtraRecs`.
- `create.go:936-940` appends `DeleteAtHeight` only to `batches[0]`, and
  `SetDAHForChildRecordsMulti` never runs for a tx that never becomes "all spent" (e.g. a
  `>utxoBatchSize` OP_RETURN tx).
- If the node owning an extra record's partition restarts after ack but before replication (a
  routine rolling restart), that record resurrects while the master record — the only pointer to
  it — is gone. No `delete_at_height` ⇒ the range-filter scan never returns it: an unreclaimable
  orphan for the life of the namespace.
- Under `COMMIT_ALL` that window only opened on genuine multi-node data loss.

**P1 — `stores/utxo/aerospike/pruner/pruner_service.go:1632,1719`: permanent blob delete paired with a master-only ACK**
- `flushCleanupBatches` removes the record with `COMMIT_MASTER`, then
  `executeBatchExternalFileDeletions` deletes the blob irreversibly.
- A resurrected record then has `external=true` with no blob. A later read — legacy sync,
  `getrawtransaction`, or block validation fetching a parent during a reorg inside the retention
  window — hits `get.go:753-759`, where the item comes back **with an error rather than a clean
  not-found**, so callers see a hard storage error for a tx that logically does not exist, until
  the next prune session re-removes the record.
- The pruner's own re-prune path tolerates the missing blob, so nothing else surfaces it.

Smaller items (`pruner_service.go:389` wiring untested and `COMMIT_ALL` being the zero value;
`util/aerospike_test.go:107` passing with the production pins deleted; the `aerospike_writePolicy`
longdesc still advertising CommitLevel) are in the full review rather than here.

Issue #1400, opened 2026-08-05
[HARDEN-0805] Pin and document the deep-reorg cap, and stop flagging an honest heavier chain as malicious


freemans13:
## Summary

During catchup, Teranode refuses to follow any chain fork that is deeper than the coinbase-maturity limit (100 blocks by default), and it does so *before* it ever weighs how much proof-of-work the competing chain carries. On top of refusing the chain, it records the peer that offered it as "malicious". This means that if a genuinely heavier chain ever forks more than 100 blocks back, Teranode will reject it — regardless of the fact that it is the better chain by work — and will penalise the honest peer that delivered it. The 100-block cap itself is a deliberate safety choice (a reorg deeper than coinbase maturity could invalidate coinbase outputs that were already spent), but two things around it are unguarded: the cap is applied *ahead of* the work comparison so it wins even against a heavier chain, and the peer gets a malicious mark for what the code's own design notes call a policy decision that should not carry a reputation penalty.

## Why it matters

There are two concrete failures, both currently silent (they log, but nothing tests or asserts the behaviour):

1. **Wrong-chain stranding.** If a real, heavier reorg deeper than 100 blocks ever occurs on the network, an SV Node follows the heaviest valid chain, but Teranode's catchup rejects it at the depth check and never gets to the work comparison. Two Teranode nodes that experience this would stay on the lighter chain and could partition permanently from the rest of the network. Whether that trade-off is acceptable is a real BSV-policy question — but right now it is neither documented as an intended divergence from SV Node nor pinned by a parity test, so nobody can tell "intended" from "accidental".

2. **Honest peer mislabelled malicious.** The depth check flags the delivering peer as malicious ("coinbase_maturity_violation") even when that peer served a perfectly valid, heavier chain. The code's own design comment (see evidence) states the malicious verdict must be gated on *work*, not depth, precisely so that an honest peer serving a heavier deep chain is followed and left unpenalised. The depth check violates that stated intent: it penalises reputation for what the same code base elsewhere calls a pure policy decision.

## Evidence in current code

All references at commit `27a0178fd`.

`services/blockvalidation/catchup.go:724-735` — the cap rejects on depth alone, no work is consulted, and it records the peer as malicious:

```go
if catchupCtx.forkDepth > uint32(u.settings.ChainCfgParams.CoinbaseMaturity) {
    u.logger.Errorf("[catchup][%s] fork depth (%d blocks) exceeds coinbase maturity (%d blocks)", ...)
[...]

Issue #1145, opened 2026-06-25
[TN-006][medium] blockvalidation: catchup clears own-chain mined_set before fork-depth/secret-mining checks


ordishs:
**Severity:** medium · **Confidence:** high · **Pass:** consensus / error-handling
**Component:** `services/blockvalidation/catchup.go`

### Summary
Catchup "Step 3.5" clears `mined_set` and emits `BlockMinedUnset` on the node's own still-canonical blocks **before** Step 4 (`validateForkDepth`) and Step 5 (`checkSecretMining`), either of which can then reject the fork without performing the reorg — so the honest chain's tx-mined state is churned for a fork that is then refused.

### Evidence
- `catchup.go:185-220` — Step 3.5 (`forkDepth>0` → `GetBlockHeadersByHeight(commonAncestor+1, currentHeight)` → `ClearBlockMinedSet` + `SendNotification(BlockMinedUnset)`).
- `:222-225` `validateForkDepth` (rejects depth > CoinbaseMaturity=100); `:227-230` `checkSecretMining`. Both after Step 3.5.
- `stores/blockchain/sql/ClearBlockMinedSet.go` sets `mined_set=false`; `model/update-tx-mined.go` (`unsetMined=true`) un-stamps txs.
- Self-healing via `GetBlocksMinedNotSet.go` periodic re-stamp.

### Why it matters
A peer merely announcing a deep fork header chain makes this node un-stamp its own confirmed txs, then reject the fork — a repeatable transient-inconsistency / churn (low-grade DoS) on the honest chain. Self-healing, so medium not high.

### Suggested fix
Move Step 3.5 to after `validateForkDepth`/`checkSecretMining` pass, ideally only once the reorg is actually committed (`on_main_chain` flipped).

### Suggested tests
Catchup with common ancestor at `currentHeight-101` → assert `validateForkDepth` errors AND `ClearBlockMinedSet`/`BlockMinedUnset` never invoked on a main-chain block. Repeat for secret-mining depth.

---
_Filed from the Teranode system review. Full context, evidence and dismissed-finding audit trail in `reviews/llm-system-review/` (findings.md, component-coverage.yaml, recommended-next-actions.md)._

PR #1515, opened 2026-08-06
fix(blockvalidation): measure fork depth from the accepted chain, not a lagging counter

### LINKED ISSUE #1438: [HARDEN-0805] Stop UTXO-store height lag from inflating catchup fork depth and mislabelling an honest peer as secret-mining
## What can go wrong

Catchup computes the common ancestor between our chain and a peer's using the **UTXO-store height** as the ceiling — only blocks at or below that height count as candidate ancestors. But the UTXO-store height is updated asynchronously and can trail the blockchain-store height by many blocks under load. When it lags, the common ancestor is forced lower than reality, which inflates the computed fork depth. That inflated depth then feeds both the deep-reorg gate and the secret-mining check. An adversary who deliberately induces UTXO lag (a spend burst) and then presents an honest, *shallow* fork can push the computed depth past the threshold — so a valid fork is rejected and the honest peer that delivered it is recorded as attempting secret mining. Because all nodes share the lag under the same load, the misclassification is correlated across the fleet.

## Where (verified on main, commit 27a0178fd)

- `services/blockvalidation/catchup.go:637` — `findCommonAncestor`.
- `services/blockvalidation/catchup.go:646` — `currentHeight := u.utxoStore.GetBlockHeight()`; the walk breaks whenever `meta.Height > currentHeight`, so only blocks at/below the (possibly lagging) UTXO height are eligible ancestors. The code comment even notes blocks ahead of UTXO height 'exist in blockchain store but aren't fully processed yet'.
- The resulting fork depth feeds `validateForkDepth` (the deep-reorg gate, `catchup.go` ~line 715) and the secret-mining check (`catchup.go` ~lines 1454-1526), which reports the peer via `recordMaliciousAttempt`.

## Why it matters

The fork-depth and secret-mining decisions are meant to reflect the true divergence point between the two chains, not an artefact of how far the async UTXO height happens to trail. A lag-induced inflation both rejects a valid shallow fork and defames an honest peer. It has not bitten because UTXO-store lag is normally small and nobody has combined deliberately-induced lag with a fork offer. This is a distinct root cause from the existing deep-reorg-cap ticket (#1400, which concerns depth-versus-work ordering and the malicious mark) — here the depth itself is computed wrong because of an async-lag ceiling.

## How to reproduce / test

Force the UTXO-store height well behind the blockchain-store height, then run a genuinely shallow honest fork through common-ancestor resolution, fork-depth validation, and the secret-mining check; assert it is *not* classified as too-deep or secret-mining and the peer is *not* flagged. Self-contained: the blockvalidation suite already provides a mock/sqlitememory blockchain client and header/block builders; the UTXO store height can be held back in the test double.

## Suggested fix

Compute the common ancestor and fork depth against a height that reflects the node's true accepted chain (the blockchain store's height) rather than the lagging UTXO-store height — or otherwise account for the known UTXO lag so a shallow fork is not inflated into a deep one. At minimum, do not feed a lag-inflated depth into the malicious/secret-mining verdict.

## Confidence

Needs a spike. The UTXO-height ceiling in `findCommonAncestor` and its flow into fork-depth and secret-mining are confirmed on main; the exploitability (how far lag can be induced and whether it reliably crosses the threshold for a shallow fork) should be confirmed before choosing the exact remediation.

> **Related but distinct from #1400.** #1400 fixes the deep-reorg cap's ordering (cap applied before weighing work) and the malicious-peer mark. This ticket is a different root cause: the fork *depth* itself is mis-computed because the common-ancestor search uses a lagging UTXO-store height as its ceiling, so even a shallow honest fork can be scored as too-deep.

---
*Part of the `[HARDEN-0805]` production-hardening sweep (label `hardening-map-2026-08-05`), grounded in `main` @ `27a0178fd`. Priority: Medium · Size: M · This task is self-contained and does not depend on any other ticket.*

freemans13:
Closes #1438

## What happened

When a peer announces a block we don't have, catchup works out where our chain and theirs diverge, and how deep that fork is. That depth decides two things: whether to reject the fork as too deep to be safe, and whether to record the peer as having withheld a chain it mined in secret.

The divergence point was found by walking the peer's headers against a ceiling taken from the **UTXO store's height**. That is not a chain height — it is a counter refreshed asynchronously each time a block notification arrives, re-reading the true height at that moment. It trails the accepted chain whenever that subscription is starved or during bulk sync, and nothing bounds how far.

Taking both the ceiling *and* the depth baseline from that counter made the fork depth wrong in **both directions**.

**Understated — the more serious half, and one the ticket missed entirely.** The ancestor was pinned at the lagging height and the depth measured from that same height, so the result was (true depth − lag). A genuinely 150-deep fork with 60 blocks of lag scored 90 and passed a coinbase-maturity gate set at 100. A consensus-safety check, silently bypassed by an unrelated counter falling behind.

**Overstated in the secret-mining check.** That check reads the *real* tip, so it measured from the real tip down to a lagging ancestor — roughly the lag itself rather than the divergence. An honest peer offering an ordinary shallow fork could be rejected *and* recorded as malicious, pinning its reputation below the eligibility threshold so it is routed around for future catchups. Because the whole fleet shares the lag under the same load, the misjudgement would be correlated across nodes, and the retry path then tries other peers and can flag a set of honest ones in a single pass.

**The ticket's attack story is wrong and I've dropped it.** It claimed an adversary could induce the lag with a burst of spends. That counter moves only on block notifications and re-reads the absolute height each time, so transaction volume cannot move it and a single notification collapses the lag to nothing. The lag is self-inflicted — bulk sync, or a starved subscription — not adversary-induced. The consequence is real; the exploit is not.

## Changes

**`services/blockvalidation/catchup.go`** — source the baseline from the blockchain store's tip, which is what divergence is a property of, and which the secret-mining check already reads. Both height sources collapse to one, so the depth is the true depth for every consumer: the coinbase-maturity gate, the secret-mining check, the mined-set cleanup range, and the header filter.

**The two heights were never two quantities.** Review turned up the detail that makes this simpler than I first described it: the UTXO store's height is set from `blockchainClient.GetBestHeightAndTime` on each `NotificationType_Block` (`stores/utxo/factory/utxo.go:139-195`). It is a lagging *cache of the very value* `GetBestBlockHeader` returns. It never carried the meaning its old comment claimed — "blocks above this aren't fully processed yet" — because it was never a watermark of applied UTXO effects, only a stale copy of the chain tip. So this is not the substitution of one semantic for another. It is the removal of a lag, and the fix is to read the value directly instead of through a cache that has no invalidation guarantee.

That also disposes of the question of whether the ancestor's UTXOs need to be present. Nothing here needed them, on either reading: UTXO effects are applied before a block is added to the blockchain store on both paths, and catchup never unspends or rewinds anything — it validates forward and lets the reorg happen when the tip moves. "Where do the chains diverge" was the only question being asked, and the counter was the wrong instrument for it.

A failure reading our own tip returns a `ServiceError` — the one error class the catchup caller retries *without* reporting peer failure. A processing error there would penalise an innocent peer for our own RPC fault, which is the same category of mistake this fix exists to remove.

**Two consequences downstream of the re-sourced height, both unremarked until review.** `currentHeight` is stashed on the catchup context and read by two later steps, and its meaning changed under both.

The mined-set cleanup at step 5.5 clears `ancestor+1 .. currentHeight`. That range used to stop at the lagging counter, leaving the blocks between it and the real tip still stamped as mined after a reorg took them; it now covers the whole reorged stretch. `GetBlockHeadersByHeight` walks the best chain by chain work, so those are exactly the canonical blocks being forked away — a latent gap closed rather than a widening, but it is a behaviour change and it was not called out.

The other is `verifyCheckpointsInHeaderChain`, which bails out whenever `forkDepth > 0`. Since the depth is now non-zero in cases where the lag previously flattened it to zero, quick validation is skipped in a few situations where it used to run. That is the safe direction — quick validation is the riskier path — but it is a silent change and worth naming.

**The ceiling itself stays, re-sourced.** Deleting it outright was my first attempt and it broke the Sybil-resistance test, which is worth recording because it is a real constraint and not a test artefact: the walk asks whether a block is *in our store*, which is true of a side-chain block we received and did not adopt — and such a block can sit above our tip. Selecting one makes the ancestor exceed the chain we are measuring from, which the secret-mining check asserts cannot happen and aborts on. So the ceiling is load-bearing; what was wrong was only where its height came from.

**The tip is now read once, which review showed this fix had made necessary.** Re-sourcing the baseline left the tip being read twice — once in `findCommonAncestor`, again in `checkSecretMiningFromCommonAncestor` — and that turned out to matter more after the change than before it.

While the baseline was the lagging counter, the relationship was `ancestor ≤ utxoHeight ≤ tip`. The ancestor sat below the counter, which sat below the tip, so the secret-mining check's invariant — *the ancestor must not exceed the height we measure from* — had a margin it never used. Measuring from the tip removes that margin by design: the ancestor can now equal the tip exactly. So any decrease in the tip between the two reads trips the invariant, and its failure path returns a `ProcessingError`, which `releaseCatchupLock` charges to the peer. A local reorg, billed to whoever happened to be serving us. That is the same category of mistake this PR exists to remove, reachable through a narrower door the PR itself opened.

`findCommonAncestor` now stashes the tip on `CatchupContext` and `checkSecretMining` hands it on, so ancestor selection, the depth trigger and the work gate all reason about one fixed tip. The window is closed by construction rather than narrowed. The check also loses an RPC it never needed — it already required that same header's chainwork for the work comparison, so it was fetching the header twice to use two halves of it. A missing tip means the caller skipped `findCommonAncestor`, which is our fault and not the peer's, so it aborts as a `ServiceError`. The invariant stays as a guard on that arrangement and on the `uint32` subtraction beneath it, with its comment corrected to say it can no longer trip. It is also reclassified from `ProcessingError` to `ServiceError`, which review turned up as the last loose end here: both heights it compares are read from our own blockchain store, so a trip is by definition our state disagreeing with itself and can never be the peer's doing. Leaving it as a `ProcessingError` meant the one path that survives the fix would still have charged the peer — a reputation hit and a sync-peer rotation for our own fault, which is exactly the failure this PR is about.

I want to be honest about the size of this: a tip that *decreases* mid-catchup means a local reorg landed inside a two-RPC window, so it is a narrow race rather than a routine failure. It is in scope because this PR is what makes it reachable, and because the fix is smaller than the bug.

**Three comment-only commits follow the fix**, correcting statements this change falsified rather than leaving them to mislead the next reader. One is a test scenario that documented itself as a depth-1 fork — true while the baseline was the UTXO height of 1018 against an ancestor at 1017, but the mocked tip *is* that ancestor, so the depth is now 0 and the fork-cleanup mocks below it are no longer exercised (the dedicated coverage for that path is in `catchup_fork_handling_test.go`). One is the Sybil-resistance test, whose comment credited our UTXO height for stopping the ancestor walk; the fixture mocks both heights at 1000 so the outcome is unchanged, but the mechanism named is the one this PR removed. The third is the block locator's cap, discussed below.

**Two further test comments are reworded** because they cited numbers their fixtures do not contain. One said "with the accepted chain at 200 and the counter still at 100" three lines above a comment explaining that the counter is *deliberately left un-mocked*; the other cited a "lag 60" that exists only in the old arithmetic. Both were counterfactual explanations reading as descriptions of the setup in front of them. Small, but this PR's whole argument is that a statement no longer true invites the next reader to trust it, and these were that failure in miniature inside the commits making the argument.

## Testing

`go build ./...`, `go vet ./services/blockvalidation/...`, and the full package suite — `blockvalidation` ok in 275s, `catchup` ok in 69s. `golangci-lint --new-from-rev upstream/main` reports 0 issues.

Two existing tests pinned the old, buggy contract, so they are reversed rather than worked around. One of them — asserting that no ancestor is found in exactly the lag scenario — described its own setup as *"the divergence scenario that causes the bug"*, so the behaviour was known and pinned anyway. It is now the regression test: with the accepted chain at 200, the ancestor at 152 must be found and the depth must be 48, where the old code would have stopped the walk below a counter lagging at 100 and found nothing.

New tests:

- The depth is not understated by lag, and a 150-deep fork is therefore rejected by the coinbase-maturity gate rather than waved through at 90. This is the consensus-safety direction the ticket missed.
- A stored block above our tip is not selected as the ancestor, while one at *exactly* our tip is, giving a depth of 0. This pins the ceiling I nearly removed, and pins its comparison as `>` rather than `>=`.
- A failure reading our own tip is a service error and not a processing error, so the peer is not charged for it.
- The tip is read exactly once across the ancestor search and the secret-mining check, asserted by call count — and, separately, that a tip which moves *down* between where the two reads used to be no longer discards a sound catchup. The call count alone was not enough: every other fixture mocks a tip that never changes, so a second read is behaviourally invisible to them and only the count would have noticed one returning.
- A fork exactly at the secret-mining threshold returns early, and one block past it reaches the work gate. The subtraction feeding that boundary is the direct consumer of the height this PR rewired, and no existing fixture sat near it.
- A registry tip naming a block we already hold is ignored; one naming a block we lack is still followed; a lookup or existence-check failure falls back to the announced block.
- The ordinary case (divergence below the tip) still resolves as before.

The lag-scenario test deliberately leaves the UTXO store's height un-mocked, so if the search ever consults it again the mock panics on the unexpected call rather than quietly passing.

Every one of those claims was checked by mutation rather than by reading — the fix broken deliberately, the suite re-run, the failure confirmed to name the right contract. That found two tests passing for the wrong reason, both now fixed. The *"missing chain tip"* case asserted only that *some* service error came back that mentioned no secret mining and flagged nobody; with the guard deleted a zero-value tip fell through to the ancestor-height check one line below, which satisfies all four of those assertions, so the guard could have been removed with the subtest still green. It now names the string only the guard emits. And `TestFindCommonAncestor_IgnoresStoredBlocksAboveOurTip` used heights of 90 and 120 against a tip of 100, which cannot tell `>` from `>=`; a header at exactly 100 was added.

The secret-mining tests now pass the tip as an argument rather than mocking the lookup, which surfaced something worth recording: the *"common ancestor ahead of our height"* case never mocked `GetBestBlockHeader` at all. It was passing on an expectation left behind by the subtest before it, through the shared mock. Supplying the tip explicitly removes that dependency, and the subtest is renamed since "UTXO height" no longer describes what it asserts. The old *"chainwork lookup failure"* case tested a lookup that no longer happens there, so it moves to the same contract at the new boundary — a missing tip must abort without flagging the peer, and must be a service error — while the lookup failure itself stays covered where it now lives.

## Trade-offs and limits

**Not reproduced against a live lagging node.** Every step is verified against the source and the arithmetic is pinned by tests, but I did not stand up a node with a starved notification subscription and offer it a fork. The lag mechanism, its lack of a bound, and both mis-measurements were read from the code. The same holds for the registry-tip route: the phantom depth and the malicious verdict are reproduced in the test suite, but I did not drive a real stale `node_status` entry through a live catchup. How often the two quantities are elevated together in practice I can reason about, not measure.

The same applies to the two-reads race: it is closed by construction and the single read is asserted by call count, but I did not drive a live tip decrease through the window to watch the invariant trip. That the window existed is a property of the code; how often it would be hit in practice I can only reason about, not measure.

**The locator cap stays, but its justification was false and is corrected.** The request we send a peer is capped at the same lagging counter, and the comment justifying that cap said `findCommonAncestor` would reject every header returned from a tip-based locator. Re-sourcing the ceiling removed that failure mode, so the cap's stated reason no longer exists — a comment asserting a contract that has gone, which invites the next reader either to delete the cap on the strength of it or to preserve it believing the old behaviour.

What the cap does now is start the peer's header stream lower than necessary, so the ancestor walk climbs back through headers we already hold — one `GetBlockHeader` lookup per block of lag, where the old ceiling stopped the walk after roughly two, since any header above the UTXO height was rejected outright. Against that, the old code re-downloaded and re-validated `lag` *full blocks*, because `filterHeaders` takes everything after the ancestor with no existence check, so the trade is still comfortably the better side.

An earlier revision of this description claimed the cap changes no outcome. That is not quite true, and I am correcting it rather than leaving it to mislead. The header fetch truncates at `CatchupMaxAccumulatedHeaders` (100,000 by default), so once the lag exceeds that, the served stream ends *below* our tip, the walk takes its ancestor at that point, and the depth is overstated by `lag - 100,000` — the same over-measurement this PR exists to remove, reached by a different route. It takes a 100,000-block lag to get there, which is why I have kept the cap rather than pulling it out alongside the ceiling fix, and the code comment now states the trade instead of asserting harmlessness. Removing the cap remains the right eventual tidy: it closes that route and collapses the O(lag) walk back to about two lookups.

**The sibling routes to the same failure, one of which this PR made worse before closing it.** A fork depth which describes no fork is reachable three independent ways. This PR closes the first outright, and closes the reachable half of the third after review showed the fix had made it *worse*, not neutral.

#1516 — the walk keeps the last header that *exists in our store*, and the header lookup has no best-chain condition, so a block from a fork we validated and did not adopt can be selected as the ancestor. Verified verbatim identical on main, and this change never decreases the computed depth, so it cannot make that route easier to hit.

#1517 — the walk stops on the first header we lack, which is sound, but it also stops when the served list simply ends, and it does not distinguish the two. Exhausting the list means no divergence was observed, yet a depth is computed anyway: the mined flag is cleared on our own canonical blocks and an honest peer can be recorded as violating coinbase maturity. Reproduced on main with no lag at all. That ticket also records why the obvious fix — forcing the depth to zero when the list is exhausted — would reintroduce the understatement this PR removes.

**An earlier revision of this description implied this PR was neutral toward that route. It is not, and the correction is the last commit here.** The depth computed on the exhaustion route is `tip − top-of-served-list`, so re-sourcing the baseline raises it by exactly the current UTXO lag. That is enough to carry an honest peer over the coinbase-maturity gate it previously sat under. What makes the list end early is the header-fetch target, and that target is the peer's P2P registry entry, refreshed from its `node_status` only every ten seconds — and `sanitizeAdvertisedTip` caps entries that are too *high* while passing anything too *low* straight through. So the entry can name a block behind the one the peer just announced, the served stream ends below our own tip, every header in it is one we already hold, and the walk takes its ancestor at the end of a list in which nothing diverged.

Writing G for how far the registry entry trails our tip and L for the UTXO lag: whenever L ≥ G, the old ceiling stopped the walk at the lagging height and the depth was **0**, with no action taken. Measuring from the real tip it becomes **G**, and once G exceeds 100 blocks `validateForkDepth` records a `coinbase_maturity_violation` against a peer sitting on our own chain. Both quantities are elevated together during bulk sync, which is exactly when catchup runs. It stays out of reach in steady state — at the mainnet tip ten seconds is a fraction of a block, so G is 0 or 1 — but reaching it needs no 100,000-block lag, only a peer that advanced more than 100 blocks inside one status interval.

I reproduced it on this branch before fixing it: a peer serving nothing but blocks we already hold scores a fork depth of 150 and is flagged.

The fix is to ignore a registry tip naming a block we already hold. Holding it is precisely the tell that the entry is stale, since it can teach us nothing we do not have, whereas the announced block is by definition one we lack — `catchupGetBlockHeaders` returns early if it exists. The choice of target moves into `catchupTargetHash` so it can be tested directly rather than through HTTP scaffolding, and every form of doubt (a registry lookup failure, an existence check we could not complete) takes the announced block. A genuinely more advanced registry tip is still preferred, which is what the registry is for.

This leaves the locator-cap route open, and deliberately: it truncates the stream for the same reason but needs a 100,000-block lag to do it, and the trade is documented at the cap rather than fixed here.

Fixing any one of the three leaves the others intact, which is why they are separate tickets rather than one.

**Adjacent but distinct from #1400**, which concerns the deep-reorg cap being applied before weighing work. This is a different root cause: the depth itself was mis-measured.


freemans13:
**Review round 1: 0 confirmed of 6 raised.** Two refutations were substantial enough to act on anyway — one became issue #1516, the other a small test-hygiene change.

**#1516 — the ancestor need not be on our best chain.** The walk keeps the last header that *exists in our store*, and the header lookup is keyed on hash alone with no chain-membership condition. So a block from a fork we validated and did not adopt, sitting below our tip, is a valid candidate — and the depth measured from it describes no reorg at all. Same class of failure as #1438, reached independently: this one is about the *ancestor*, not the baseline, so a perfectly current tip does not help.

The verifier refused it as a finding here on solid ground, and did the work to prove it: the selection logic is verbatim identical on main, and across several fork layouts this change never *decreases* the computed depth — raising the baseline by the lag can raise the selected ancestor by at most the lag, so the bypass cannot become easier to hit. In the lagging case it moves 50 points closer to the truth. Blocking a change that strictly narrows the gap because it does not also close an older orthogonal route would be the wrong call. Filed with the concrete fix, since the membership call already exists on the interface and the header lookup already returns the id needed to batch it.

**Test hygiene.** My depth test set the coinbase-maturity gate to 100 before asserting rejection. The value is already 100 in the fixture, so assigning it implied it might differ and would have silently retuned the test if the fixture changed. Now asserted as a precondition instead. I checked whether the mutation could leak across tests — it cannot, the fixture is rebuilt per call — but asserting is the honest form either way.

The other refutation worth a line: two independent tip reads mean an ancestor selected at exactly the tip trips a "should not happen" abort if our tip moves down mid-walk, and that error charges the honest peer. Real, but the verifier established main has *zero* margin here in the healthy case — at tip with a live subscription the counter equals the tip, so main already selects an ancestor at the tip and the same race exists. The margin only appears in the lagging regime, where the tip is rising, not falling.

One lens (test quality) returned nothing, so I checked its intended subject myself rather than treat silence as a pass — that is where the settings-mutation change above came from.

Verification: 1029 tests in `blockvalidation`, `make lint` 0 issues.

freemans13:
**Review round 2: 0 confirmed of 4 raised.** One refutation became issue #1517; another proved a test does what it claims, which is worth stating because I had only asserted it.

**The tripwire is real, and now demonstrated rather than claimed.** The lag test deliberately leaves the UTXO store's height un-mocked, on the theory that if the search ever consults it again the mock panics instead of quietly passing. A verifier tested exactly that: it built a scratch worktree, kept the corrected baseline but put the walk's ceiling back on the lagging counter, and the test failed immediately with an unexpected-call panic naming the reintroduced line. With the probe reverted it is green again — so it is a true differential, not a test that passes either way. That is the strongest form of confirmation available for a guard whose whole purpose is to catch a future refactor.

**#1517 — running out of peer headers is read as a fork.** The walk stops on the first header we don't hold, which is sound. But it also stops when the served list simply ends, and it does not distinguish the two. Exhausting the list means no divergence was observed at all — yet a depth is computed from the last header anyway. Consequences: the mined flag is cleared on our own canonical blocks with an un-stamp notification for each (the churn #1145 exists to prevent), and an honest peer whose list ends far enough below our tip is recorded as violating coinbase maturity. Reachable with no lag at all: a stale registry tip for a peer is enough.

The verifier refused it as a finding here and was right on both counts. It reproduced the whole thing on main in the ordinary no-lag regime, so this branch does not introduce it — what main did inside the lagging window was pin the depth to zero for *every* fork, genuine deep ones included, which is the bypass this PR removes rather than a protection it loses. And it showed the reviewer's proposed fix — force the depth to zero when the list is exhausted — would reintroduce that very understatement, because the accumulation loop truncates precisely when a long genuine fork is being served. The ticket says so explicitly, because that trap is the interesting part.

That makes three independent routes to the same underlying failure, a fork depth that describes no fork: the baseline from a lagging counter (this PR), the ancestor not being on our best chain (#1516), and the walk not knowing why it stopped (#1517). Fixing any one leaves the others intact.

Also refuted: that the locator cap's justification is now stale (true, but that file is untouched and the cap is behaviourally inert either way — a handful of extra headers).

Verification: 1029 tests in `blockvalidation`, `make lint` 0 issues.

freemans13:
**Review round 3: 0 confirmed. Converged** — marking ready for review.

Worth noting how this round was read, because the first attempt at it failed with a network error and returned an empty findings list. An empty list from a crashed run is indistinguishable at a glance from an empty list from a clean one, so I re-ran it rather than bank the result. The second attempt completed and the journal shows the reviewer genuinely returned no findings, which is what converged looks like.

Three rounds, cumulatively: round 1 found that the divergence point need not be on our best chain at all, which became #1516 — refused as a finding here only after a verifier worked the arithmetic across several fork layouts and showed this change never *decreases* the computed depth. Round 2 demonstrated that the lag test's un-mocked height is a real tripwire, by reintroducing the old height source and watching the test panic, and found the third route to the same failure (#1517). Round 3 found nothing.

Final verification: 1029 tests in `blockvalidation`, `go vet` clean, `make lint` 0 issues.

liam (approved):
Approving. The central factual claim holds, and I verified it rather than taking it on trust.

## The load-bearing claim is correct

Everything rests on "the UTXO store's height is a lagging cache of the chain tip, not a watermark of applied UTXO effects". If that were wrong, this would be swapping one semantic for another inside a consensus-relevant gate. It is not: `stores/utxo/factory/utxo.go` sets it from `blockchainClient.GetBestHeightAndTime(ctx)` on every `NotificationType_Block`, both on the init path and in the subscription goroutine. It is literally a stale copy of the value `GetBestBlockHeader` returns. The reframing in the description -- that this is lag removal rather than substitution -- is right, and it is what makes the change safe.

I derived both mis-measurement directions independently and they come out as claimed:

- **Deep fork** (tip 1000, utxo 940, divergence 850): the ancestor sits under the ceiling, depth is measured as `940 - 850 = 90` against a true 150, and slips under a maturity gate of 100.
- **Shallow fork** (tip 1000, utxo 940, divergence 990): the walk breaks at the ceiling so the ancestor pins at 940, and the secret-mining check reads `1000 - 940 = 60` against a true depth of 10 -- the lag itself, exactly as described.

The "never decreases the computed depth" argument, which is what keeps this from making the #1516 route easier, also holds: raising the ceiling only admits more candidates, so the ancestor is greater than or equal to its old value, and a walk that stopped on a missing header picks the same ancestor against a higher tip.

## Other claims checked

- **Both heights were always one quantity.** The only consumers of `catchupCtx.currentHeight` are the fork-depth calculation and the mined-set cleanup range. Neither wanted a UTXO watermark, so collapsing to one source loses nothing.
- **The `ServiceError` routing is real.** `Server.go:1908` clears markers and returns *before* `reportCatchupFailureForError` and `ReportPeerFailure`, so a local RPC fault genuinely does not charge the peer.
- **The locator-cap trade is real.** `filterHeaders` takes everything after the ancestor index with no existence check, so a lower ancestor really did mean re-downloading full blocks rather than re-walking headers. The corrected justification is the honest one.

Build clean; `go test -race ./services/blockvalidation/` green (302s).

## Two things worth raising

**The mined-set clearing blast radius grows by exactly the lag.** Step 5.5 clears `mined_set` over `[ancestor+1, currentHeight]`, and raising `currentHeight` from the UTXO height to the tip extends that range by the lag. The extension is *correct* -- those blocks are being orphaned -- but step 5.5 still runs before the fork is adopted, since catchup validates forward and the reorg happens when the tip moves. So if the catchup then fails partway, `lag` more still-canonical blocks have had their mined_set cleared than before. The issue-1145 ordering fix moved this after the safety checks but not after adoption. Worth naming, because the enlargement lands precisely in the starved-subscription scenario this PR targets.

**The maturity gate returns `ServiceError`** (`catchup.go:735` on main, untouched here), which is the retry-locally-do-not-charge-the-peer class. Pre-existing, but this PR is what makes it reachable: the gate previously could not fire in the lagging case, and now it will. A peer offering a genuinely too-deep fork gets retried against, bounded by `CatchupMaxAttemptsPerBlock`, rather than routed around. Out of scope to change here, but it is a behaviour this PR switches on, so it belongs in the description or a follow-up ticket.

---

The honesty about what was verified versus reasoned, and the explicit correction of both the ticket's attack story and an earlier revision's "changes no outcome" claim, is what makes this reviewable at all at this size.

ordishs (approved):
## Code Review — teranode PR #1515

### Overview
The fix moves the fork-depth baseline in `findCommonAncestor` from the UTXO store's lagging height counter to the blockchain store's real tip (`GetBestBlockHeader`), stashed once on `CatchupContext` and reused by `checkSecretMiningFromCommonAncestor`. This closes a two-RPC race where a local reorg between separate tip reads could trip an invariant and misclassify the resulting error as chargeable to the peer. A related fix (`catchupTargetHash`) ignores a P2P registry tip when we already hold that block, since a stale registry entry otherwise truncates the peer's header stream and can produce a phantom fork depth against an honest peer.

Verified against the current PR head. Build, vet, and `golangci-lint --new-from-rev` are clean. Ran every new/changed test (ancestor-search, secret-mining work-gating boundary cases, single-tip-read, tip-moves-down race, `catchupTargetHash` stale-tip guard, integration scenarios) — all pass. Traced the `ServiceError` vs `ProcessingError` reclassification through to `processCatchupChItem` in `Server.go` and confirmed `ErrServiceError` returns early without charging the peer, matching the PR's description.

### Verdict
**APPROVE.** The root-cause narrative is well-supported by the code, the fix is minimal and correctly scoped, and new tests pin both the consensus-safety direction (fork depth no longer understated) and the edge cases review turned up. No correctness, security, or test-coverage gaps found worth blocking on.

PR #1518, opened 2026-08-06
fix(utxo/aerospike): decide transaction creation from the master record, and fence the creation lock

### LINKED ISSUE #1442: [HARDEN-0805] Aerospike create reports success without writing the master record, and the creation lock's release is unfenced
## Problem

Two defects in the Aerospike UTXO store's transaction creation path (`stores/utxo/aerospike/create.go`). The first bites without any concurrency; the second needs a stall.

### Create reports success without having written the master record

A transaction is written as a batch of records. Only record 0 — the master — carries the mined-state metadata: `GetBinsToStore` puts `BlockIDs`/`BlockHeights`/`SubtreeIdxs`, or `UnminedSince` when there are none, on `batches[0]` alone.

Creation success was decided by "was *any* record in the batch written". So a writer that found the master already present, and merely filled in a missing child record, reported that it had created the transaction — while another writer's mined-state metadata stood and its own was never applied.

Callers act on that signal. Block validation only collects a transaction for its mined-info repair when the store reports the transaction already exists, so a false success suppresses the repair: the master keeps the earlier writer's `UnminedSince` with no block references, and a mined transaction stays recorded as unmined. That then feeds the unmined pruner and the re-add-to-block-assembly path.

This needs no concurrency at all. A partial batch failure deliberately leaves its records in place for the next attempt to complete, so a later, entirely sequential create from a different caller finds the master present, fills in the missing child, and reports success.

### The creation lock's release is unfenced

Transactions that span more than one record — more outputs than the batch size, default 128 — take a per-transaction creation lock: an Aerospike record written `CREATE_ONLY` with a wall-clock TTL and nothing else guarding it (`calculateLockTTL`: 30s base plus 2s per record, capped at 300s).

`releaseLock` deleted that record by key, with no check that the lock being deleted was the one this caller acquired. A writer that stalls past the TTL loses the lock silently — Aerospike expires the record regardless of what the writer believes. A second writer then satisfies `CREATE_ONLY` and takes a fresh lock. When the first writer resumes, its deferred release deletes the second writer's live lock, letting a third writer in while the second still believes it holds it.

Note what this is *not*: each record is written `CREATE_ONLY`, which the database settles server-side, so two concurrent writers cannot overwrite each other's records — one wins per record and the other is told it already exists. The blob write is idempotent. The hazard here is lock eviction, not record corruption.

## Solution

**Decide creation from the master record alone.** Report "we created this transaction" only when record 0 was written by us. Extract the rule so it can be tested without an Aerospike client.

**Fence the lock with a per-acquisition token.** Stamp a fresh random token into a lock bin on acquire, and make the release conditional on that token still matching, using a filter expression so the check is server-side and the delete stays one round trip. A release that finds a different token is not ours to clean up: treat it as success, distinctly from the already-gone case, since the two mean different things when reading logs after an incident.

Do **not** use the record's generation for this. The acquire is always a first-ever `CREATE_ONLY` write, so its generation is always 1 — and a lock re-created after expiry is also 1. Generation cannot distinguish "my live lock" from "someone else's, created after mine expired", which is exactly the case being fenced.

The token fence bounds lock eviction only. It does not stop two writers both believing they hold the lock and racing to create the same transaction's records; that is bounded by the per-record `CREATE_ONLY` writes and by the master-record rule above.

## Severity

The false-success defect is reachable today on a sequential path and silently records mined transactions as unmined, so it is the half that matters. The lock fence is worth having but prevents less: the damage it averts is a third writer entering after a stale release, on a path that also requires a stall past the TTL.

Related: #1518, #1519.

---
*Part of the `[HARDEN-0805]` production-hardening sweep (label `hardening-map-2026-08-05`). Priority: High · Size: M.*

freemans13:
Closes #1442

## The problem

A transaction is written to the Aerospike UTXO store as a batch of records, and only record 0 — the master — carries the mined-state metadata. `GetBinsToStore` puts the block references, or the unmined marker when there are none, on `batches[0]` alone.

Creation success was decided by whether *any* record in the batch was written. So a writer that found the master already present and merely filled in a missing child reported "I created this transaction", while another writer's metadata stood and its own was never applied.

That success is one the caller acts on. Block validation only collects a transaction for its mined-info repair when the store says it already exists, so a false success **suppresses the repair**: the master keeps the earlier writer's unmined marker with no block references, and a mined transaction stays recorded as unmined — which then feeds the unmined pruner and the re-add-to-block-assembly path.

**It needs no concurrency whatsoever.** A partial batch failure deliberately leaves its records in place for the next attempt ("Do NOT clean up partial records"), so a later, entirely sequential create from a different caller finds the master present, fills in the missing child, and reports success.

Separately, the creation lock's release was unfenced. That lock is only taken when a transaction spans more than one record — more outputs than the batch size, default 128 — and it was deleted by key with no check that the lock being deleted was the one this caller acquired. A writer that stalls past the TTL loses the lock silently, a second writer takes a fresh one, and the first writer's deferred release then evicts it.

Worth being clear about what that second defect is not: each record is written create-only, which the database settles server-side, so two concurrent writers cannot scribble over each other — one wins per record and the other is told it already exists. The blob write is idempotent too. The hazard is lock eviction, not record corruption.

## Changes

**Decide creation from the master record alone** (`stores/utxo/aerospike/create.go`). The rule is extracted into a pure function, `classifyCreateBatchResults`, which reports whether *we* wrote record 0, plus the record indices that were already present and those that genuinely failed. The surrounding function acquires the creation lock through the concrete client before reaching this point, so extracting the rule is what makes it testable at all — see Testing.

`GetBinsToStore` now records that the mined-state bins living only on `batches[0]` is what that rule depends on, since it is the function someone edits when changing record layout.

**Fence the lock** (same file). Each acquisition carries a random token, and the release is conditional on it via a server-side filter expression, so a delete only takes effect on the lock this caller was handed. A release that finds a different token is not ours to clean up: logged and treated as success, distinctly from the already-gone case, since the two mean different things when you are reading logs after an incident.

Deliberately **not** generation-based, which is the obvious approach and does not work here: the acquire is always a first-ever create-only write, so its generation is always 1 — and a lock re-created after expiry is also 1. Generation cannot distinguish "my live lock" from "someone else's created after mine expired", which is precisely this failure mode.

The code says plainly what the fence does not do: it stops a stale release deleting another writer's live lock, but it does not stop the concurrent double-write itself. That is bounded by per-record create-only and by the master-record rule above.

**A simplification pass over the above**, no behaviour change. The classifier returns the failed indices rather than a bare boolean, so the caller no longer walks the batch results a second time to find what to log. `acquireLock` and `releaseLock` call the `isKeyExists` / `isKeyNotFound` helpers instead of re-deriving those checks inline, and `isKeyExists` matches on the `aerospike.Error` interface rather than one concrete pointer type. Two stale comments citing line numbers that pointed at unrelated code were replaced with named references.

**A second simplification pass, after review.** Three error paths in the lock code turn out to be unreachable, and each was costing a branch no test could ever enter. `crypto/rand.Read` never returns an error on Go 1.26 — it crashes the program instead — so `generateLockToken`'s error return was dead, and it propagated a second dead branch into `acquireLock`; `rand.Text` does the same job in one line, with at least 128 bits of entropy, and removes both. `releaseLock`'s `isKeyNotFound` branch cannot fire against the pinned client either, which maps the server's `KEY_NOT_FOUND` to `existed=false` with a nil error — the comment directly below it already said exactly that, so the code and the comment disagreed. The remaining filter check moved into an `isFilteredOut` helper alongside `isKeyExists`, which is where this package keeps its result-code matching. `create.go` ends up eleven lines shorter.

## Testing

`go build ./...`, `go vet`, `make lint` (0 issues), and the whole `stores/utxo/aerospike` suite under `-race` with the testcontainer harness up — both packages green in 289s. Re-run after merging `upstream/main` in, since that merge touched the same package.

The master-record rule is pinned by a table test over the combinations that matter, including the mirror case (we wrote the master, a child already existed → we *did* create it) and the distinction between already-exists and a genuine failure. Mutation-verified, before and after the simplification pass: reverting the rule to "any record counts" fails exactly the regression case and the master-failure case.

**The fence is now covered against a real server, where an earlier round of this PR said it could not be.** That claim rested on the locked path not being drivable container-*free*, which is true and still is — `storeExternallyWithLock` acquires its lock through the concrete client, so a test with no server panics before reaching anything assertable. But this package already runs container-backed tests by default through its own testcontainer harness, so "not container-free" was never the same thing as "not testable", and treating the two as equivalent is what left the fence unverified.

`TestLockFence` runs the sequence the fence exists for: writer A acquires the lock, the record goes away as an expiry would leave it, writer B acquires a fresh one, and A's stale release must leave B's lock standing. Removing the filter expression fails exactly that assertion and no other. It also pins the ordinary cycle, a second acquisition being refused while the lock is held, a double release being a no-op, and two acquisitions never sharing a token.

`TestStoreExternallyMasterRecordDecidesCreation` does the same for the master-record rule, end to end through `storeExternallyWithLock` rather than through the classifier alone — and it needs neither concurrency nor an injected error: create a multi-record transaction, delete its child to reach the state a partial failure leaves, then attempt it again. Reverting the rule to "any record counts" makes that second attempt report success, which is the defect in one line. Both mutations were run, not reasoned about.

That closes the gap #1585 tracks, and answers the question left open in it. Deleting an already-gone lock comes back as `existed=false` with a nil error, not `FILTERED_OUT` — so an ordinary expired lock logs "already gone", and only a live lock carrying someone else's token logs "held by a different token". The log wording was already right; now it is pinned rather than assumed.

Every API used was verified against the pinned client source rather than assumed: that a filter expression exists on the write policy, that the delete path honours it, the numeric result code for a filtered-out delete, and the expression constructors' signatures.

## Trade-offs and limits

**The fence is not worth much on its own** — it prevents a third writer entering after a stale release, but the damage was never there. The master-record fix is the half that matters, and it is independently justified. Both are included because the ticket asks for the fence and it is correct as far as it goes.

**This trades one narrow failure for another, and the trade is deliberate — see #1519.** Returning already-exists means the validator takes its early return, which explicitly does not announce the transaction to block assembly. For a *mined* transaction arriving via block validation that is exactly right, and it is the whole point: already-exists is what triggers the mined-info repair. For an *unmined* transaction it means one more case where the store holds a complete transaction that nothing has been told to mine — bounded by block assembly re-seeding from `UnminedSince` at startup.

That case is narrow — it needs a multi-record transaction whose master was written by an attempt that then failed, followed by a later attempt completing the missing child. But it is real, and I would rather name it than let the diff imply it away. The change also removes a double-notify in the opposite direction, where a second writer could announce a transaction whose master the first writer had already created and announced.

The reason it is not fixed here is that the fault is not in the store. The validator infers "already announced" from "already exists", and those are set by different steps; the same harm is already reachable on `main` on the far more common single-record path, where a batch-level error after the writes land makes the resubmission report already-exists. Fixing it properly means the validator checking rather than assuming, which is #1519 — and that ticket records why it must not simply undo this change, since the two callers want different things from the same signal.

**One narrowing this shares with the retry policy.** `settings.conf` sets `MaxRetries=5` on the batch policy, so a create-only sub-request that reaches the server but loses its response comes back `KEY_EXISTS` on retry — from this writer's *own* earlier attempt, not another writer's. If that lands on record 0, the master-record rule reports already-exists for a transaction this writer really did create, and the validator's early return then skips announcing it to block assembly. The transaction is still fully persisted and spendable, `clearCreatingFlag` still runs, and it is re-announced when next encountered, so the cost is a delay rather than a loss. It is not wholly new either — the same hazard already applied whenever the entire batch was retried. Worth naming because the trade is deliberate: a missed notification is cheaper than a suppressed mined-info repair, which is what the old rule risked.

**The expiry window is short.** Roughly 32 seconds for a typical transaction, growing only to a 5-minute cap. The August outage's multi-minute freeze exceeds the typical case by an order of magnitude. Lengthening it is not proposed here — a longer expiry trades one failure for another, and the fence plus the master-record fix bound the consequences rather than the trigger.

**Found while reviewing this change, filed separately:** the subtree quorum lock in `services/subtreevalidation/TryLockIfNotExists.go` has the same unfenced-release hazard, and its heartbeat never runs in production because both call sites set an absolute timeout. That is #1584.


freemans13:
**Review round 1: 0 confirmed of 4 raised.** One refutation identified a real weakness this change slightly widens — now filed as #1519 and disclosed in the description rather than left for a reviewer to find in the diff.

**The trade-off worth knowing about.** Returning already-exists sends the validator down its early return, which explicitly does not announce the transaction to block assembly. For a *mined* transaction arriving via block validation that is exactly right and is the entire point — already-exists is what triggers the mined-info repair. For an *unmined* one it means one more case where the store holds a complete transaction that nothing has been told to mine. Narrow (it needs a multi-record transaction whose master was written by an attempt that failed, then a later attempt completing the child) but real.

The verifier refused it as a finding against this PR and was right: the same harm is already reachable on `main` on the far more common single-record path, where a batch-level error after the writes land makes the resubmission report already-exists. The genuine fault is the validator inferring "already announced" from "already exists" — two things set by different steps. #1519 records that, and records why it must not simply undo this change, since block validation and the validator want different things from the same signal.

**One lens returned nothing, so I checked its subject myself** rather than count silence as a pass — the same thing happened on #1515. It was the correctness lens, which mattered most here. I verified directly that record 0 of the batch is always record 0 of the split, that the mined-state metadata is only ever placed there, that the single-record case is behaviourally identical to before, and that the early return above genuinely skips the announcement.

Also refuted: that the filtered-out release deserves a warning rather than a debug line (the baseline logged *nothing* in either branch, so this is a request for a feature upstream never had, and the condition has an instrumented signature downstream anyway); and that the lock-record documentation needs a row for the new bin — true, but that table is already a schema version behind on the very index it documents and wrong about the setting that triggers the whole pattern, so adding one accurate row to it would be a poor trade.

Verification: 815 tests in the store package, 19 skipped for want of a container, `make lint` 0 issues.

freemans13:
**Review round 2: 0 confirmed of 2 raised. Converged** — marking ready for review.

Both candidates were refuted as pre-existing or cosmetic, but one of them surfaced a process problem worth naming: **a round-1 review agent left an uncommitted edit in my worktree** instead of reporting it, despite the instruction not to touch the tree. The edit itself was correct — a doc comment named `getBinsToStore` where the function is exported as `GetBinsToStore` — so I verified it against the actual definition and folded it in rather than discarding it. The same correction is applied to the description.

That is the third time a review agent has written to the branch it was reviewing. It is benign here, but on an earlier PR one of them deleted a fix from a previous round, and on this branch I destroyed my own uncommitted work with a restore that assumed the file was committed. Anything left in a worktree by something other than the author is a hazard regardless of whether the content is right.

The other candidate: the "lock already gone" debug line can never fire, because the client returns success rather than that result code when the record is absent. Factually correct — and entirely pre-existing, since that branch and its discarded return value are both baseline; this change only added a log line inside it. Refused on the same grounds round 1 refused the argument that the filtered-out case deserves a warning: the baseline logged nothing at all, so these are requests for observability the code never had, not repairs of anything this change broke.

**One lens returned empty again**, so as on the previous two PRs I checked its subject myself rather than treat silence as a pass. It was the fence lens. I verified the token is generated before the lock is written and its failure aborts the acquire, so the token is only ever returned after a successful create — meaning the deferred release can never run with an empty token against a live lock. The reachability of the filtered-out branch and the shape of its type assertion were separately confirmed by the other lens against the pinned client source.

Verification: 815 tests in the store package, 19 skipped for want of a container, `go vet` clean, `make lint` 0 issues.

freemans13:
Two follow-ups filed from the review of this branch, both for the gaps the PR body names rather than anything it introduces:

- #1585 — cover the lock fence with a test that actually expires the lock mid-write. This is the "only structurally verified" gap under Testing above, plus the one client-source question that could not be settled by reading (whether deleting an already-expired record with a filter returns `KEY_NOT_FOUND` or `FILTERED_OUT`, which changes only the log wording).
- #1584 — the subtree quorum lock in `services/subtreevalidation/TryLockIfNotExists.go` has the same stale-release hazard this PR fences, unfenced, and its heartbeat branch never runs in production because both call sites set an absolute timeout. Found while reviewing this change; it is a separate defect and deliberately not bundled here.

#1519 already covers the validator-side trade-off described under Trade-offs and limits.

ordishs (approved):
## Review — no P0 or P1 findings

The master-record rule and the token-fenced lock release are both correct fixes for real defects. The
code comments explain the reasoning well. Nothing here blocks the merge.

Two P2 items are worth a follow-up. Details are in the full review.

- `stores/utxo/aerospike/create.go:1207-1228` — the `!masterCreated` branch merges two states. "Every
  record already existed" is a true `ErrTxExists`. "The master pre-existed and I wrote at least one
  child" is not: the earlier writer failed part-way and never notified block assembly, so this writer is
  the first one able to announce the transaction. `Validator.go:825` returns early on `ErrTxExists`, so
  both `sendToBlockAssembler` (`Validator.go:972`) and `twoPhaseCommitTransaction` (`Validator.go:1005`)
  are skipped, and the `WithLocked(true)` set at `Validator.go:1551` stays set. The only repair is
  `loadUnminedTransactions`, whose production callers are `Start` (`BlockAssembler.go:944`) and reset
  (`BlockAssembler.go:611`) — no timer. The description names this trade-off and defers it to #1519. The
  part the description does not name is the surviving locked flag: until a restart or a reorg, children
  cannot spend the transaction either. Suggest a distinguishable outcome for the "I completed a partial
  record set" case, and a store-level test for the two-writer sequence.

- `stores/utxo/aerospike/lock_token_test.go` — the lock fence has no test. `TestGenerateLockToken`
  proves properties of `crypto/rand`, not the fence. No test calls `acquireLock` or `releaseLock`, and
  none mentions the `lock_token` bin, so an inverted filter or a dropped `FilterExpression` would keep
  the suite green. The internal package already has `runAerospikeTestContainer`, so a container-backed
  test is available: `runAerospikeTestContainer` in `container_helper_internal_test.go` already serves
  `aerospike_test.go` and `index_test.go` from inside `package aerospike`. Assert that a stale-token
  release leaves the second holder's lock intact and returns nil, and that a matching-token release
  removes the record.

Smaller P3 items:

- `create.go:1108`, `:1245` still cite `processTxMetaUsingStore.go:112-122`, which is the missing-hash
  collection loop; the Creating check sits at `processTxMetaUsingStore.go:154` and `:225`. `create.go:1240`
  cites `teranode.lua:288`, the tx-not-found branch; the creating guard is at `teranode.lua:295-297`.
  `create.go:1247` cites "(line 890)", an unrelated `IsCoinbase` bin. `create.go:1218` and `:1261` still
  describe an "All exist" path that the new rule replaced.
- `create.go:1371` states one cause as fact ("now held by a different token"). Name both causes, include
  the transaction hash, and add a counter — a 32s TTL overrun currently leaves only a `Debugf`.
- `docs/topics/features/utxo_lock_records.md:157`, `:236` gained the lock-token change but never state the
  master-record rule; the retry path still reads "finds all records exist".

Verified against the vendored client (`v8.7.1-bsv5`): `Delete` honours `policy.FilterExpression`, and
`delete_command.go:66-72` maps `FILTERED_OUT` to an error while a missing record returns `existed=false`
with a nil error. `TestClassifyCreateBatchResults`, `TestIsKeyExists` and `TestGenerateLockToken` pass.

ordishs (approved):
No P0 or P1 findings. Approve.

Verified against the current tree and the pinned client `aerospike-client-go/v8@v8.7.1-bsv5`:

- `stores/utxo/aerospike/create.go:1061-1077` - record 0 really is the master.
  `GetBinsToStore` puts `BlockIDs`, `BlockHeights`, `SubtreeIdxs` and `UnminedSince` only on
  `batches[0]`, and `CalculateKeySourceInternal` maps index 0 to the bare transaction hash, on
  both the `StoreTransactionExternally` and the `StorePartialTransactionExternally` path.
- `services/blockvalidation/quick_validate.go:1294` - block validation repairs mined info only on
  `ErrTxExists`, so the old any-record success rule did suppress that repair. The fix is real.
- `stores/utxo/aerospike/create.go:1363-1390` - the lock fence matches the client contract.
  `delete_command.go` maps `FILTERED_OUT` to an error and `KEY_NOT_FOUND_ERROR` to
  `existed=false` with a nil error, and both branches handle it as a no-op.
- The token is 128 bits from `crypto/rand`, hex-encoded, and never logged.
  `util.GetAerospikeWritePolicy` builds a fresh policy per call, so the filter expression cannot
  leak into another write.
- `errors.As` over the `aerospike.Error` interface resolves at every new call site, and `Matches`
  walks the wrap chain. This is safer than the old concrete pointer assertion, because
  `Client.PutBins` chains its errors.
- `LockRecordIndex` is `0xFFFFFFFE` at `create.go:93`, so the docs correction is right.
- `go vet ./stores/utxo/aerospike/...` exits 0. `go test -short -run
  'TestClassifyCreateBatchResults|TestIsKeyExists|TestGenerateLockToken' ./stores/utxo/aerospike/...`
  reports `ok`.

One P2 and four P3 items follow, all non-blocking.

The P2 is the caller-side conflation the description already names and tracks in #1519. The store
now returns `TxExistsError` both when another writer owns the master and when this writer completed
an earlier attempt's record set. The validator skips the block assembly notification in both cases.
That state self-heals at the next unmined reload. One correction to the framing: the description
calls the case narrow on the strength of the documented 20,000-output threshold, but the real
setting is `utxostore_utxoBatchSize` with a default of 128 (`settings.conf:1373`). The multi-record
path is therefore far more common than the docs imply.

P3 items:

- `stores/utxo/aerospike/create.go:1210-1218` - the block comment still says the branch triggers
  when all records exist. The trigger is now `!masterCreated`. Line 1212 states the exact
  assumption the new rule breaks.
- `docs/topics/stores/utxo.md:356` still says `0xFFFFFFFF`. The PR fixed the same error in
  `utxo_lock_records.md`. Both files, plus `utxo_lock_records.md:339`, also cite a 20,000 threshold
  and a `utxo_store_batch_size` key that does not exist.
- `stores/utxo/aerospike/lock_token_test.go` - no test asserts the fence itself. Rename the bin in
  `acquireLock` and the whole unit suite still passes.
- `stores/utxo/aerospike/create_master_record_test.go` - `TestIsKeyExists` passes only bare errors,
  so it does not pin the `errors.As` behaviour the change introduced. The classifier test asserts
  slice lengths rather than the index values.

ctnguyen (approved):
Review at commit [1faa6a7](https://github.com/freemans13/teranode/commit/1faa6a7900b5a6c995f0693a496bda1a35a774a1)

This is a careful fix for a subtle, concurrency-free correctness bug, and the diff reads like someone who understood the failure end to end. Deciding creation from the master record instead of "any record written" is the right call, and extracting the rule into `classifyCreateBatchResults` makes it testable without a container; the table test covers the master-vs-child combinations the rule turns on. The lock fence is the right shape — a random token rather than generation, with the reasoning written down — the write policy is rebuilt per call so the filter can't leak, and the token is never logged. The description is unusually honest about what it does *not* fix and defers those to tracked issues. I don't think anything here blocks merge, but two of the items below are more than cosmetic — a locked-master consequence the description doesn't name, and a lock fence that nothing exercises — so the no-blocking call rests partly on those being followed up rather than on the change being fully covered.

## Non-blocking issues

### ChiR1 — `!masterCreated` recovery can leave a completed master locked

*Same as [@ordishs's point "the !masterCreated branch merges two states" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1518#pullrequestreview-4915704519).*
*Same as [@ordishs's point "the caller-side conflation the description already names" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1518#pullrequestreview-4915916511).*

**Problem:** The recovery branch at `create.go:1209` returns `TxExistsError` for every `!masterCreated` outcome. One state it now covers is: an *earlier attempt wrote the master (index 0) then failed before announcing*, and this writer has just completed the missing child. The validator's early return on `ErrTxExists` (`Validator.go:825`) then skips `sendToBlockAssembler` (`Validator.go:972`) and `twoPhaseCommitTransaction` (`Validator.go:1005`), so the `WithLocked(true)` set for block-assembly-bound txs at `Validator.go:1552` is never cleared. The *unannounced-transaction* half of this is the trade-off the PR already names and that [bsv-blockchain/teranode#1519](https://github.com/bsv-blockchain/teranode/issues/1519) records and disposes of — I raise it only as settled background, not to re-litigate it. What #1519 does **not** name is the *surviving locked flag*: the completed master stays locked, so children can't spend it until `loadUnminedTransactions` runs at Start/reset (`BlockAssembler.go`, no timer) — i.e. a restart or reorg. Before this PR the same scenario reported success via `createdAny`, which announced it and cleared the lock, so the locked-master outcome is new to this change.

This path is only reachable if the duplicate CREATE_ONLY outcome on record 0 surfaces as a per-record `BatchRec().Err` with a nil batch-level error — `storeExternallyWithLock` returns before the classifier when `batchOperate` itself errors. The pinned client does exactly that: an ordinary per-record result code sets the per-record error and continues (`batch_command_operate.go:157`), so the batch call returns nil. The one exception is the narrow case where the failing record is the first and its message is the last part with only a bare header, which returns a batch-level error (`batch_command_operate.go:124`) and never reaches the classifier — an edge that predates this PR.

**Why it matters:** The description frames the case as rare on the strength of a ">20,000 outputs" threshold; the real trigger is `utxostore_utxoBatchSize`, default **128** (`settings.conf:1373`), so multi-record transactions — and thus this window — are far more common than the docs imply. The locked-flag consequence is not stated anywhere in the diff. Confidence: the store side and the reachability above are confirmed against the pinned client source; the validator early-return and locked-flag path were traced statically, not executed.

**Fix:** Ideally give "I completed a partial record set" an outcome distinguishable from "another writer owns a live master" so the validator can announce/unlock the former — though the clean fix is validator-side and belongs with #1519. At minimum, add the store-level two-writer test @ordishs asked for, and document the surviving-locked consequence next to the #1519 trade-off so it is not lost.

### ChiR2 — the lock fence and its wrap-chain matching are untested

*Same as [@ordishs's point "the lock fence has no test" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1518#pullrequestreview-4915704519).*
*Same as [@ordishs's point "no test asserts the fence itself" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1518#pullrequestreview-4915916511).*
*Same as [@ordishs's point "TestIsKeyExists passes only bare errors" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1518#pullrequestreview-4915916511).*

**Problem:** `lock_token_test.go` only exercises `generateLockToken` — i.e. properties of `crypto/rand`. No test calls `acquireLock`/`releaseLock`, references the `lock_token` bin, or asserts the `FilterExpression`, so inverting the filter, dropping it, or renaming the bin in `acquireLock` all keep the suite green. Separately, `TestIsKeyExists` feeds only bare errors, so it never exercises the wrap-matching the switch to `errors.As` over the `aerospike.Error` interface introduced. I traced this to be sure it is worth pinning: `PutBins` (`client.go:462`) returns `command.Execute()`, whose execute loop threads every attempt's error through `chainErrors` (`error.go:400`), and on a retried command the returned aerospike error carries a wrapped inner (`error.go:428`) — so a chained result is reachable. I want to be honest about the limit of that, though: the chain keeps the `*AerospikeError` as the *outer* and `.Matches()` reads that outer's code, so a KEY_EXISTS result is still caught even by the old concrete `*aerospike.AerospikeError` assertion — I did not find a PutBins path that wraps a KEY_EXISTS such that the concrete assertion would miss it. So the wrapped-error test is best justified as a defensive guard that keeps `isKeyExists` robust to *any* wrapping (including teranode's own `%w` chains, per `native_op_unit_test.go:56`), rather than as proof of a known PutBins wrap path. Either way it is currently unpinned. (The classifier's own `isKeyExists` is separately safe here: per-record batch outcomes are set directly via `setError` at `batch_record.go:142`, i.e. an unwrapped Aerospike error.)

**Why it matters:** The fence is the PR's second headline change, and nothing that ran here touches it; a regression ships silently. The switch to `errors.As` over the interface at the `acquireLock`/`releaseLock` call sites is a sound robustness improvement, but its behaviour under wrapping is asserted by no test. The gap is acknowledged and tracked in [bsv-blockchain/teranode#1585](https://github.com/bsv-blockchain/teranode/issues/1585), and the new-code coverage gate is currently failing — so this is "should it block" rather than "does it exist"; I read it as non-blocking given #1585, but worth closing given the harness is already there.

**Fix:** The package already has `runAerospikeTestContainer` (`container_helper_internal_test.go`). Add: a stale-token release leaves a second holder's lock intact and returns nil; a matching-token release removes the record. And add a wrapped case to `TestIsKeyExists`, mirroring the existing pattern in `native_op_unit_test.go:56`:

```go
require.True(t, isKeyExists(fmt.Errorf("put failed: %w", aeroErr(types.KEY_EXISTS_ERROR))))
```

### ChiR3 — the `FILTERED_OUT` release log states one cause as fact

*Same as [@ordishs's point "states one cause as fact (now held by a different token)" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1518#pullrequestreview-4915704519).*

**Problem:** `create.go:1371` logs the `FILTERED_OUT` outcome as `"is now held by a different token; not ours to delete"`. But `FILTERED_OUT` also arises when the lock record carries *no* `lock_token` bin at all — e.g. a lock created by a pre-fence node during a rolling deploy — where `ExpStringBin("lock_token")` cannot match and the true cause is "old-format lock", not "a different token". The message asserts a single cause, carries no transaction hash, and has no counter/metric.

**Why it matters:** During a rolling deploy the two node versions are mixed: a new node cannot delete an old node's unfenced lock (it falls to TTL — fine), while an old node still deletes by key and can transiently evict a new node's fenced lock. That window is bounded by the deploy, but the log will misattribute it, and a genuine post-TTL stale-release after an incident leaves only a `Debugf` with nothing to grep or count.

**Fix:** Reword so the message doesn't assert a single cause — a different token *or* a pre-fence/old-format lock with no `lock_token` bin — and include the tx hash so a real stale-release is identifiable in the logs. A counter/metric would help spot post-TTL overruns but is optional for this finding.

### ChiR4 — stale line-number citations remain in the create.go comments

*Same as [@ordishs's point "still cite processTxMetaUsingStore.go:112-122" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1518#pullrequestreview-4915704519).*
*Same as [@ordishs's point "the block comment still says the branch triggers when all records exist" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1518#pullrequestreview-4915916511).*

**Problem:** The PR fixed two stale citations but left several: `create.go:1108` and `:1245` still cite `processTxMetaUsingStore.go:112-122` (the Creating check is at `:154`/`:225`); `create.go:1247` still describes completing Phase 2 "via 'All exist' path (line 890)" — the path the new `masterCreated` rule replaced; `create.go:1240` cites `teranode.lua:288`, the tx-not-found branch rather than the creating guard.

**Why it matters:** These comments are the map future maintainers follow through a genuinely intricate two-phase path; a wrong line number sends them to unrelated code, and the "All exist" wording now contradicts the rule this very PR introduced.

**Fix:** Replace with named references (as the PR already did for the two it corrected), or refresh the line numbers.

### ChiR5 — docs still carry the old lock index, threshold, and setting name

*Same as [@ordishs's point "docs/topics/stores/utxo.md:356 still says 0xFFFFFFFF" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1518#pullrequestreview-4915916511).*
*Same as [@ordishs's point "gained the lock-token change but never state the master-record rule" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1518#pullrequestreview-4915704519).*

**Problem:** The PR corrected `0xFFFFFFFF` → `0xFFFFFFFE` in `utxo_lock_records.md` but not in `docs/topics/stores/utxo.md:356`, which still shows the old value. Both docs also still describe a ">20,000 outputs" trigger and a `utxo_store_batch_size` key (`utxo_lock_records.md:27,103,339`; `utxo.md:352`) — the actual setting is `utxostore_utxoBatchSize`, default 128 (`settings.conf:1373`). And neither doc mentions the new master-record rule the retry path now depends on; `utxo_lock_records.md` still reads "finds all records exist".

**Why it matters:** These are the reference docs for exactly the mechanism this PR changed; the 128-vs-20,000 gap in particular understates how often the multi-record path (and ChiR1's window) is hit.

**Fix:** Correct the index in `utxo.md`, replace the threshold/key across both docs with `utxostore_utxoBatchSize` (128), and add a line stating that record 0 is the master and creation is decided from it.

## Recap

|---|---|---|---|
| ChiR1 | Recovery leaves master locked | | 55% |
| ChiR2 | Fence & wrap-chain untested | | 42% |
| ChiR3 | FILTERED_OUT log states one cause | | 30% |
| ChiR4 | Stale comment citations | | 20% |
| ChiR5 | Docs: index, threshold, rule | | 15% |

Issue #1416, opened 2026-08-05
[HARDEN-0805] Add goleak goroutine-leak detection to the long-lived service test suites


freemans13:
## Summary

Teranode's Go test suites do not use any goroutine-leak detection. The `go.uber.org/goleak` library is not a dependency the code actually uses — it appears only as a leftover hash pair in `go.sum` (pulled in transitively), is absent from `go.mod`, and is not imported by a single `.go` file in the repository. As a result, if one of the long-lived stateful services (block assembly, block validation, p2p, and others) starts a goroutine on `Start`/`New` and fails to stop it on shutdown or context cancellation, nothing in CI notices. The leak is simply invisible until it shows up as growing memory and goroutine counts on a running node in production.

## Why it matters

The services here are long-lived and heavily concurrent: they spawn background workers, subscribers, tickers, and fan-out goroutines that are supposed to exit cleanly when their context is cancelled or the service is stopped. A goroutine that is accidentally left running — a `for { select { ... } }` loop with no `ctx.Done()` case, a subscription channel never closed, a worker whose stop signal is dropped — does not crash anything. It just accumulates. On a node that runs for days under continuous load (exactly the IBD and steady-state regimes this project cares about), thousands of orphaned goroutines pin memory and stack space that the garbage collector can never reclaim, which is a quiet contributor to the kind of slow memory growth that only ever gets diagnosed by pulling a live goroutine profile off a struggling node.

This is a silent failure class. A leaked goroutine produces no test failure, no log line, and no error. The unit tests for these services construct a service, exercise it, and return — and they never assert that the goroutines the service started have actually gone away afterwards. `goleak.VerifyNone` (or `goleak.VerifyTestMain`) closes exactly that gap: it snapshots the goroutines still alive at the end of a test and fails if any unexpected ones survive, turning a production-only symptom into an immediate, deterministic CI failure at near-zero cost.

## Evidence in current code

Repository at commit `27a0178fd`, worktree `/Users/freemans/github/freemans13/teranode/.worktrees/harden-verify`.

goleak is not a real dependency. It is absent from `go.mod` entirely, and present in `go.sum` only as the two hash lines that survive when a module is in the build graph but unused:

- `go.sum:1288` — `go.uber.org/goleak v1.3.0 h1:...`
- `go.sum:1289` — `go.uber.org/goleak v1.3.0/go.mod h1:...`
- `grep -c goleak go.mod` → `0`

[...]

Issue #1521, opened 2026-08-07
[HARDEN-0805] Every ban notification can block forever, leaking a goroutine per ban


freemans13:
## What can go wrong

Every ban and unban spawns a goroutine to notify subscribers, and that goroutine can block forever. `Add` and `Remove` each do `go func(){ b.notifySubscribersAsync(event) }()`, and `notifySubscribersAsync` sends to each subscriber's channel through the shared safe-send helper **with no timeout**. With no timeout option that helper is a plain blocking channel send — it guards against panicking on a closed channel and nothing else.

Subscriber channels are buffered at 100. Any subscriber that stops draining — because it is slow, wedged, or simply went away without unsubscribing — costs one permanently blocked goroutine per ban or unban, forever.

## Why it matters

This is the silent-accumulation failure this whole area is about: no error, no log line, no test failure. The goroutines pin their stacks and everything they reference, and the only symptom is memory growth on a node that has been up for a while.

It is not theoretical, and the scale is startling. Turning on goroutine-leak detection over the p2p package produced **1956 leaked goroutines from a twelve-second test run, and 1900 of them were this one send**. That is a test run, not a busy node — a real node bans peers continuously.

Worth being precise about what is and is not broken: the ban list's own bookkeeping is fine, and a slow subscriber does not lose events or corrupt state. The defect is purely that the notification has no way to give up.

## Where (verified on `main` @ `25748b9c2`)

- `internal/banlist/ban_list.go` — `Add` (~174) and `Remove` (~197) each spawn `go func(){ b.notifySubscribersAsync(event) }()`.
- `internal/banlist/ban_list.go` — `notifySubscribersAsync` (~325) sends per subscriber via the safe-send helper with no timeout argument.
- `util/safe_send.go` — with zero timeout options this is a bare `ch <- t` wrapped in a panic guard.
- Subscriber channels are created buffered at 100.

## How to reproduce / test

Self-contained in `internal/banlist`, no external services. Subscribe, never drain the channel, then perform more than 100 bans. Assert the goroutine count returns to its baseline. Today it grows by one per ban beyond the buffer and never recovers. Once #1522 lands, the package has goroutine-leak detection enabled (see #1416), so the test above will fail loudly rather than needing a manual count. The gate cannot see this leak on its own: no existing test both subscribes and stops draining, so writing that test is what turns the gate into a regression net here.

## Suggested fix

[...]

PR #1522, opened 2026-08-07
test: gate two packages against goroutine leaks, and measure what the rest would cost

freemans13:
Partially addresses #1416. It gates `internal/banlist` and `stores/blob/file`; the six service
packages named in the definition of done stay open, with the measured starting counts recorded
as a comment on #1416 so the backlog survives this PR. No closing keyword, deliberately.

## What happened

Nothing in the test suites notices when a long-lived service starts a goroutine and never stops it. The leak-detection library was not a real dependency at all — it survived only as leftover hashes in `go.sum`, absent from `go.mod`, imported by zero files. So a service that fails to shut down a background worker produces no test failure, no log line, and no error. It shows up as growing memory on a node that has been up for days, diagnosed only by pulling a goroutine profile off something already struggling.

The ticket asks for the check across the long-lived service suites. **I measured before implementing, and the answer is that it is not affordable there yet:**

| package | leaked goroutines |
|---|---|
| block validation | 2087 |
| p2p | 1956 |
| block assembly | 1259 (after the cheapest central fix) |
| subtree validation | 549 |
| validator | 538 |
| propagation | 472 |

Turning the check on in those packages makes CI permanently red, and a permanently-red gate gets ignored and then disabled. This project has already had to gate a suite for exactly that reason. So the honest scope is the two packages that can be green today, plus a template and a measured backlog.

## The most valuable thing the measurement found

**A genuine production leak, filed as #1521.** Of p2p's 1956 leaked goroutines, **1900 were a single send**: every ban and unban spawns a goroutine to notify subscribers, and that notification sends to each subscriber's channel with no timeout — a plain blocking send. Any subscriber that stops draining costs one permanently blocked goroutine per ban, forever. That was from a twelve-second test run; a real node bans peers continuously.

That is a change to a long-lived service and deserves its own review, so it is not bundled here.

## Changes

**Promote the library to a real dependency**, and gate two packages.

**`internal/banlist`** is the interesting one, and the reason to believe the gate works: it was red at **around 111 leaked goroutines** on the committed defaults — 4 per store across 28 tests, all from blockchain stores the tests never closed — and is green after a three-line cleanup. With `blockchain_use_in_memory_chain_check` on it is 6 per store, around 167. The totals move by one or two between runs because a connection opener or a cache janitor can be caught mid-lifecycle; the per-store figures are the exact ones. The stores' `Close` exists and works; the tests simply were not calling it. This is also the package where the production leak of #1521 lives, so the gate is positioned to catch it once a test exercises the non-draining subscriber path. It cannot see that leak today: the notification channel is 100-deep, the four subscribing tests all drain or unsubscribe, and the one test that issues enough bans to overflow the buffer, `TestBanList_ConcurrentAccess`, registers no subscriber at all. The test that closes the net belongs with #1521.

**`stores/blob/file`** needed no changes and serves as the template for enabling the same check package by package as each is cleaned up.

**One ignore, and only one.** `gocore`'s init-time goroutine is a bare loop with no context and no exit path — genuinely unstoppable. It must be matched by name rather than by top frame: its top frame is `time.Sleep`, so ignoring the top frame would blind the check to every sleeping goroutine in the tree. That distinction is subtle enough to be worth the comment it has.

**Deliberately no ignores** for database pools, ttl caches, gRPC callback serialisers, or the Aerospike cluster worker. Every one of those exits when the object owning it is closed, so ignoring them would hide precisely the leak this exists to catch — a service holding an unclosed store. That refusal is *why* only two packages qualify today, and it is the right trade: a gate that cannot fail is worse than no gate, because it looks like coverage.

## Testing

`go build ./...`, `make lint` (0 issues), and both gated packages run **five consecutive times** under `-race` with the CI flags (`SETTINGS_CONTEXT=test`, `-tags testtxmetacache`) — 505 passing, no flakes. The repeat count is deliberate: leak detection retries for about a second before reporting, which is thin margin for a goroutine that was told to stop a tick earlier, and this repo has known flakiness under contention.

**Verified the gate can fail.** Removing the ban list cleanup turns the package red with `found unexpected goroutines`. One caveat for anyone repeating that measurement in a git worktree: gocore's config search walks *up* out of the worktree, so a `settings_local.conf` in the parent checkout is picked up silently. That is what makes the in-memory (6-per-store) figure appear locally when the committed default is off, and CI — which has no `settings_local.conf` — sees the 4-per-store case. A leak check that cannot fail is decoration, so this is the assertion that matters most.

## Trade-offs and limits

**Six of the ticket's target packages are deferred, and that is most of what was asked for.** They need per-package cleanup first, each with its own measured starting point. The counts above are the backlog; the acceptance criterion for each is the same — green with no new ignores beyond `gocore`.

**A correction worth recording**, because it inverted a conclusion mid-spike: the leak check only runs when the package's tests all *passed*. An early measurement called block assembly the cleanest package, when in fact its run had a flaky failure so the check never executed — it is the worst. That also means the gate silently produces no signal on a red run, so it cannot be relied on in a package with a flaky test. Block assembly has two known port-bind flakes; they matter more once a gate is attached.

**One eight-line central fix was tried and deliberately left out.** Adding store cleanup to two shared test helpers halved block assembly (2457 → 1259) and broke nothing, but nothing in *this* PR can prove it works. It belongs with the block assembly cleanup, where a gate makes it falsifiable.

**The gate does not cover this package's own production constructor.** `NewFromSettings` opens a blockchain store, takes the DB handle out of it, and returns without keeping or closing the store, so the same background goroutines the test cleanup here stops run for the process lifetime. Bounded rather than growing, one store per construction, but it is the exact shape this gate advertises catching, and no test exercises the constructor so the gate stays green over it. Raised by ctnguyen in review and filed as #1653 rather than fixed here: the complete fix needs `Close` on `banlist.Interface` and plumbing into the daemon and p2p shutdown paths, which is production change to a long-lived service and the same reason #1521 was split out.

**Not verified:** validator's and block validation's counts were measured once and not re-run. Propagation's 472 may be far smaller in CI — 448 of them come from an Aerospike store created only when Docker is available locally, so it may be a cheap early win. Worth checking before ranking the backlog.


freemans13:
**Review round 1: 0 confirmed of 3 raised.** One lens returned nothing, so I checked its subject myself — as on the previous three PRs, an empty answer is treated as a failure rather than a pass.

**What I verified in its place**, since it was the lens that mattered most here:

The gate can fail. Removing the ban list cleanup turns that package red with `found unexpected goroutines`.

The gate in the blob-store package is **live, not decorative** — the PR itself flagged it as the weaker of the two and offered to drop it. Removing the single ignore turns it red on exactly the `gocore` goroutine, which proves two things at once: the check is detecting goroutines there rather than passing vacuously, and the ignore is load-bearing rather than defensive. That was the honest doubt about that package and it is now answered.

Flakiness: five consecutive runs from me under `-race` with the CI flags, plus twenty-four more from a verifier at `GOMAXPROCS=1`, all green.

**The refutation is worth reading in full**, because its supporting work is better than the finding. The claim was that closing the blockchain store in cleanup races the store's own startup-rebuild goroutine, producing error-level log noise on nearly every test and risking flakiness in the gate itself. The mechanism is real — `Close` does not wait for that goroutine, and the lines are genuinely emitted.

Both consequences fell over on checking. The verifier built a control probe to establish that Go discards a passing package's output entirely unless `-v` is used, confirmed CI's runner prints one line per package and only dumps output for failures, and then showed the same error lines already appear on `main` in the store's *own* test suite — a package this PR does not touch. So it is not a new class of output, and it is invisible where it would matter. On the flakiness half they measured the retry window at roughly 430ms against a post-close path that does no I/O at all, then ran 24 executions to confirm.

That is the right shape for a refutation: not "I disagree", but "here is the control I ran and here is the number".

Verification: 505 tests across both gated packages at `-count=5` under `-race`, `make lint` 0 issues.

freemans13:
**Review round 2: 0 confirmed. Converged** — marking ready for review.

The result arrived as an empty findings list, which is ambiguous on its own: a reviewer that crashed returns exactly the same shape as one that found nothing. Checked the run's journal, which records an explicit empty findings list from a completed agent — so this is convergence, not a silent failure. Worth stating because I banked a crashed round as converged earlier in this campaign and had to re-run it.

Two rounds, cumulatively: round 1 raised three, confirmed none, and its strongest refutation did real work — building a control to show Go discards a passing package's output, then demonstrating the same log lines already appear on `main` in the store's own suite. I separately verified the two things the empty lens was meant to cover: that the gate fails without the cleanup, and that the blob-store gate is live rather than decorative (removing its single ignore turns it red).

Final verification: 505 tests across both gated packages at `-count=5` under `-race` with the CI flags, plus 24 further executions from a verifier at `GOMAXPROCS=1`, all green. `make lint` 0 issues.

liam (approved):
Approving. I checked the two things a gate like this has to get right, and it gets both: the gate genuinely fails on a real leak, and the scope is honest about what it cannot cover yet.

Measuring before implementing, and then refusing to turn the check on in the six packages where it would be permanently red, is the right instinct. A red gate that nobody can fix gets ignored and then deleted, and the measured backlog is more useful than a disabled workflow would have been.

## Verified rather than assumed

**The gate actually catches leaks.** I mutation-tested both packages:

- Removing the `t.Cleanup(store.Close)` from `newTestBanList` fails `internal/banlist`, with the stack pointing at `stores/blockchain/sql/sql.go:1826` created by `sql.New` at `:388`. So the fix and the check are genuinely coupled, not coincidentally green.
- Adding a `go func() { select {} }()` to `stores/blob/file` fails there too, naming the offending line. The template works as a template.

**The `VerifyTestMain` caveat is exact.** Worth quoting since it is the sharpest thing in the PR and it is easy to get wrong:

```go
exitCode := m.Run()
...
if exitCode == 0 {
    if err := Find(opts); err != nil { ... }
}
```

So a package with any failing test reports no leaks and looks clean. Documenting that in both files, rather than only in the PR, is what makes this safe to copy.

**The gocore ignore is justified and correctly scoped.** `gocore@v1.0.82/config.go:328` is `go func() { time.Sleep(1s); ticker := ...; for { ... } }()` started from `init`, with no context and no exit path. Genuinely unstoppable. The `IgnoreAnyFunction` versus `IgnoreTopFunction` reasoning is also right for a reason worth spelling out: that goroutine's top frame is `time.Sleep` only during its first second and something in the ticker select afterwards, so a top-frame ignore would be both fragile and far too broad. Matching the goroutine's own function name anywhere in the stack is the correct choice.

Both gated packages pass locally, and `go.sum` needed no change, which corroborates the description's point that the dependency had been sitting there unused.

## CI

`smoketest (shard 0)` is red on `TestBlockValidationWithParentAndChildrenTxs` in `test/e2e/daemon/ready`. I do not think this change can cause it: the diff is four files, no production code, and the only non-test change is a test-only dependency entering `go.mod`. Neither gated package is on the e2e path. The branch is also current with main (zero commits behind), so it is not a stale-base artifact. Worth a re-run to confirm it is a flake before merge, but I would not hold the PR on it.

## Note on the follow-up

The p2p finding in #1521, 1900 of 1956 leaked goroutines from a single unbounded send on the ban notification path, is a good argument for this gate existing at all: it was found by measurement, not by reading. Keeping it out of this PR was right.

Nothing blocking.

ctnguyen (commented):
Review at commit [31d48ba](https://github.com/freemans13/teranode/commit/31d48ba4f7ce155d858ea8336d095e353b63d3fe)

This is a well-argued PR and the measurement work behind it is the valuable part. Promoting `go.uber.org/goleak` from leftover `go.sum` hashes to a real dependency, then proving the gate has teeth on `internal/banlist` (red at ~111 leaked goroutines, green after a three-line `t.Cleanup`), is exactly the right way to introduce a check like this. The single `IgnoreAnyFunction` for gocore's init goroutine is correctly reasoned — I checked `gocore@v1.0.82` `Stat.go:60-71` and it really is a bare `for { … time.Sleep(1ms) }` with no context and no exit path, and its idle top frame really is `time.Sleep`, so `IgnoreTopFunction` would indeed blind the check to every sleeping goroutine. The refusal to add ignores for db pools, ttl caches and the Aerospike worker is the discipline that makes the gate meaningful. The exit-code caveat both new comments describe is also verified in `testmain.go`: `Find` runs only when `m.Run()` returned 0, so a red or flaky package silently reports no leaks — and it is good that the PR documents that rather than leaving it to be rediscovered. Two things need attention before merge, though: the `Closes #1416` keyword against an issue whose acceptance criteria this PR explicitly does not meet, and a production-side store leak in the very package being gated.

## Blocking issues

### ChiR1 — `Closes #1416` will auto-close an issue whose definition of done is not met

**Problem:** The PR body's first line is `Closes #1416`, so merging closes [bsv-blockchain/teranode#1416](https://github.com/bsv-blockchain/teranode/issues/1416) automatically. That issue's definition of done names three packages specifically — `services/blockassembly`, `services/blockvalidation`, `services/p2p` — and its acceptance criteria include `grep -rl goleak --include="*.go" .` returning at least those three, plus `go test` over those three passing with the check active. This PR gates `internal/banlist` and `stores/blob/file` and gates none of the three; `git grep -ln goleak -- '*.go'` at this head returns exactly the two new `main_test.go` files. The PR is honest about this in its own "Trade-offs and limits" section ("Six of the ticket's target packages are deferred, and that is most of what was asked for"), which makes the closing keyword the inconsistency rather than the scope decision.

**Why it matters:** The scope call itself is defensible — a permanently-red gate does get disabled, and the measured counts (block validation 2087, p2p 1956, block assembly 1259) support that. But the six-package backlog only exists as prose in a merged PR description if the tracking issue closes. That is the classic way hardening work evaporates: the ticket reads "done", nobody re-opens it, and the measured starting points are lost. The issue is still open today, so this is fixable by editing the body before merge.

**Fix:** Change the body's first line to a non-closing reference and make the backlog a tracked artefact:

```text
Partially addresses #1416 — gates internal/banlist and stores/blob/file;
the six service packages named in the definition of done remain open,
with measured starting counts recorded in #1416.
```

Then post the count table as a comment on #1416 (or open one issue per deferred package) so the backlog survives independently of this PR's description.

**Criticality: 62%**

### ChiR2 — `NewFromSettings` drops the blockchain store without closing it

**Problem:** `internal/banlist/ban_list.go:44-61`: `NewFromSettings` calls `blockchain.NewStore` at `:50`, extracts `store.GetDBEngine()` (`:55`) and `store.GetDB()` (`:60`), and returns without retaining or closing `store`. The only reference is dropped on return, so the store's background goroutines — the same reservation sweep, ttl-cache janitors and `database/sql` connection opener the PR's own new comment at `ban_list_test.go:36-48` enumerates — run for the process lifetime with nothing able to stop them. `SQL.Close` exists (`stores/blockchain/sql/sql.go:411`). This is precisely the "unclosed store owns background goroutines" defect the PR fixes on the test side, left standing in production code in the package the PR chose to gate.

**Why it matters:** Bounded, not accumulating — be clear on that. There are three call sites: `services/p2p/BanList.go:33`, which is inside a `sync.Once` singleton (`BanList.go:30-33`) and so contributes at most one store per process; and `daemon/daemon_services.go:1215` inside `createBanList` (`:1214`), which is called from `daemon/daemon_services.go:443` (asset ban list) and `:1069` (propagation ban list) with no `sync.Once` guarding either. So the true shape is one store per daemon service construction plus the single p2p store — bounded per construction, held until process exit, not growth. What makes it blocking is the interaction with this PR: the new gate is attached to `internal/banlist` and no test exercises `NewFromSettings`, so the gate is green while the package's production constructor contains exactly the shape the gate advertises catching. Merging the gate over an unfixed instance of its own target defect is what turns "this package is leak-free" into a claim narrower than it reads.

**Fix:** Note that adding `Close` to the concrete type alone is not sufficient — the daemon callers hold `banlist.Interface` (`internal/banlist/interface.go:11-36`), whose method set is `IsBanned`, `Add`, `Remove`, `ListBanned`, `Subscribe`, `Unsubscribe`, `Init`, `Clear`, with no `Close`. A complete fix is three parts: have `BanList` own the store, add `Close` to both the struct and the interface, and plumb the call into each caller's shutdown path.

```go
type BanList struct {
	// ...
	store blockchain.Store // owned when constructed via NewFromSettings
}

func (b *BanList) Close(ctx context.Context) error {
	if b.store == nil {
		return nil
	}
	return b.store.Close(ctx)
}
```

Add a test that constructs via `NewFromSettings` and closes, so the new gate actually covers the constructor. If the interface change and shutdown plumbing are judged too broad for a test-infrastructure PR, the minimum before merge is to file it as its own issue and drop the leak-free framing from this PR's body.

**Criticality: 55%**

## Non-blocking issues

### ChiR3 — "the gate there has immediate value" overstates what the gate can currently see

**Problem:** The PR body says of `internal/banlist`: "This package is also where the production leak above lives, so the gate there has immediate value", and [bsv-blockchain/teranode#1521](https://github.com/bsv-blockchain/teranode/issues/1521) goes further — "The package now has goroutine-leak detection enabled (see #1416), so a test that exercises this will fail loudly rather than needing a manual count." Neither holds yet at this head. The leak path is `ban_list.go:175-177` and `:198-200` (`go func(){ b.notifySubscribersAsync(event) }()` → `util.SafeSend` into a 100-buffered channel, `ban_list.go:286`, `:333-335`). Reaching a blocked send needs both a registered subscriber and more than 100 undrained events, and no test has both: the three subscribing tests (`ban_list_test.go:129`, `:221`, `:442`) each drain their channel and issue far fewer bans than the 100-deep buffer, while `TestBanList_ConcurrentAccess` (`ban_list_test.go:514-537`) does issue 500 `Add` calls — plenty to overflow the buffer — but registers no subscriber at all, so `notifySubscribersAsync` iterates an empty slice. Detection therefore needs a test that subscribes and then stops draining, which does not exist here.

**Why it matters:** The claim is the load-bearing justification for picking this package over others, and #1521 now depends on it for its own "will fail loudly" assertion. Someone fixing #1521 could reasonably assume the regression net is already in place and skip writing the test that is the actual net. Not blocking: excluding the #1521 fix from this PR is a sound split, and the gate does have real value — it caught the 111-goroutine test leak.

**Fix:** Reword to what the gate does today — "this is the package where the production leak of #1521 lives, so the gate is positioned to catch it once a test exercises the non-draining subscriber path" — and add that test as part of #1521, e.g.:

```go
ch := bl.Subscribe()          // deliberately never drained
defer bl.Unsubscribe(ch)
for i := 0; i < 150; i++ {    // past the 100-deep buffer
	require.NoError(t, bl.Add(ctx, fmt.Sprintf("10.0.0.%d", i), expiry))
}
```

**Criticality: 38%**

## Recap

|---|---|---|---|
| ChiR1 | Closes keyword on unmet issue | YES | 62% |
| ChiR2 | NewFromSettings never closes store | YES | 55% |
| ChiR3 | Gate cannot see notify leak | | 38% |

freemans13:
@ctnguyen thanks, all three land. ChiR1 and ChiR3 are fixed in the PR body, ChiR2 is filed as #1653 rather than fixed here. Details below.

**ChiR1 — closing keyword.** Fixed. The body's first line is now "Partially addresses #1416", no closing keyword, and the measured backlog is posted as a comment on #1416 (#&NoBreak;1416 comment 5407785666) so it survives independently of this description. That comment carries the six counts, the acceptance criterion, and the three caveats worth keeping: the exit-code caveat that made block assembly look like the cleanest package when it is the worst, propagation's 448 Aerospike goroutines that may vanish in CI, and the unproven eight-line central fix. You were right that the ticket reading "done" is how the backlog evaporates.

**ChiR2 — `NewFromSettings` drops the store.** Verified and filed as #1653 with your call-site analysis, including the bounded-not-growing distinction and the point that adding `Close` to the concrete type alone is insufficient because `banlist.Interface` has no `Close`. Not fixed here, for the same reason #1521 was split out: the complete fix is production change to a long-lived service, needing an interface change plus shutdown plumbing in the daemon and a singleton reset in p2p, and it deserves its own review rather than riding on a test-infrastructure PR. I took your stated fallback instead and dropped the framing: the "Trade-offs and limits" section now says outright that the gate does not cover this package's own production constructor, and why.

Provenance, since it affected the call: `NewFromSettings` is unchanged by this branch. It came in with #548 and is an ancestor of upstream/main, so this PR did not introduce it.

**ChiR3 — "immediate value" overstates the gate.** Confirmed independently and reworded. The channel is 100-deep at `ban_list.go:286`; the four subscribing tests (`ban_list_test.go:132`, `:224`, `:445`, `:473`) all drain or unsubscribe; `TestBanList_ConcurrentAccess` at `:514` issues 500 `Add` calls but registers no subscriber, so `notifySubscribersAsync` iterates an empty slice. Nothing currently reaches a blocked send. The body now says the gate "is positioned to catch it once a test exercises the non-draining subscriber path" and names the reason it cannot today.

I also edited #1521, which was the second place the claim lived. It said "the package now has goroutine-leak detection enabled, so a test that exercises this will fail loudly". It now says the gate cannot see the leak on its own and that writing the non-draining test is what turns it into a regression net. That was the more dangerous copy of the claim, since it could have led someone to skip writing the actual net.

No code changes this round. HEAD is c6284cf, which is this branch with upstream/main merged in.

freemans13:
@liam the smoketest shard 0 failure you flagged was a flake. CI is fully green on c6284cf: 36 success, 1 skipped, 0 failures, including `smoketest (shard 0)` and `TestBlockValidationWithParentAndChildrenTxs`. Your reasoning held, four test-only files cannot reach the e2e path.

Also confirms the rest of the board on this head: Sonar quality gate passed with 0 new issues and 0 security hotspots, benchmark comparison shows 0 regressions across 139 benchmarks, and Socket scores goleak 100 on all five categories. Both gated packages run green locally under `-race` with the merged upstream/main.

Copilot has not reviewed this PR at any point. Every request since 2026-08-20 has come back with the same refusal: "Copilot was unable to review this pull request because the user who requested the review has reached their quota limit." Recording that rather than letting the refusal read as a clean pass.

ctnguyen (approved):
Review at commit [c6284cf](https://github.com/freemans13/teranode/commit/c6284cfd072d2b3ce1baa8c19b7c533a1b8468c5)

Good round. All three previous points landed, and they landed in the way that keeps the information alive rather than the way that makes the review comment go away: the closing keyword is gone and the six-package backlog now lives as a substantive comment on #1416 with the counts, the acceptance criterion and the caveats intact; the constructor leak is filed as #1653 with the call-site analysis and the interface problem spelled out; and the body's claim about what the gate can see is now the accurate, narrower version, with the same correction pushed back into #1521 so the second copy of the claim doesn't outlive the first. The code itself is unchanged since the last review — same 89 lines — and it still reads well: the "why `IgnoreAnyFunction` and not `IgnoreTopFunction`" comment, the refusal to ignore pool/cache/gRPC goroutines, and the `VerifyTestMain` exit-code caveat documented in both files rather than only in the PR body are the parts a future copier of this template most needs. Nothing new survived scrutiny this round.

## Status of previous points

ChiR2 is graded against the fallback the original point itself authorised ("if the interface change and shutdown plumbing are judged too broad for a test-infrastructure PR, the minimum before merge is to file it as its own issue and drop the leak-free framing"). Both halves of that minimum landed — #1653 exists with the full analysis, and the body's "Trade-offs and limits" now discloses that the gate does not cover this package's own production constructor. The underlying `NewFromSettings` defect is untouched by design and tracked in #1653.

|---|---|---|---|
| ChiR1 | Closes keyword on unmet issue | YES | FIXED |
| ChiR2 | NewFromSettings never closes store | YES | FIXED |
| ChiR3 | Gate cannot see notify leak | | FIXED |

PR #1523, opened 2026-08-07
fix(blob/batcher): stop dropping a write when its flush fails, and bound Close

### LINKED ISSUE #1412: [HARDEN-0805] Cover blob batcher silent write loss on failed flush and unbounded Close
## Summary

The batching wrapper around the blob store (`stores/blob/batcher/batcher.go`) has two related liveness/durability problems in its write path. First, callers of `Set` always get `nil` back immediately because the blob is only enqueued for a background worker; nothing the caller sees reflects whether the blob was ever actually written. Second, when that background worker hits an overflow flush that fails, it drops the item it was holding and leaves the half-built batch in place, so blobs can be lost silently during a storage outage with only a log line to show for it. On top of that, the flush always runs under `context.Background()` and `Close` ignores the context it is handed and waits unconditionally, so a hung underlying store makes shutdown hang forever and defeats the service manager's stop timeout.

## Why it matters

This wraps a real blob store in production. `stores/blob/factory.go` wires it in via `createBatchedStore`, so any blob store configured with batching goes through this code path.

Two concrete failures:

1. Silent write loss during a transient storage outage. The worker dequeues an item, and if adding it would overflow the batch, it first flushes the accumulated batch. If that flush fails, the worker returns before appending the just-dequeued item and without resetting the batch. The dequeued item is already gone from the queue, so its bytes are lost for good. The only trace is a logged error. This is dangerous precisely because it is silent: the caller was told `nil` (success) long ago, and nothing surfaces the loss.

2. Unbounded shutdown. `Close` is supposed to honour a timeout so a stuck node can still be stopped. Here it ignores its context argument entirely and blocks on the worker goroutine, which during shutdown flushes the final batch through the underlying store under `context.Background()`. If the underlying store hangs, `Close` never returns, so the service manager's stop deadline is meaningless.

## Evidence in current code

All references are `stores/blob/batcher/batcher.go` at 27a0178fd.

`Set` enqueues and unconditionally returns `nil` (lines 363-377):
```go
func (b *Batcher) Set(_ context.Context, hash []byte, fileType fileformat.FileType, value []byte, _ ...options.FileOption) error {
	b.queue.Enqueue(BatchItem{ ... })
	select {
	case b.notifyCh <- struct{}{}:
	default:
	}
	return nil
}
```

Overflow-flush failure drops the current item and does not reset the batch (lines 179-189). On error it returns before reaching the `append` at line 189 and before the reset at lines 184-185:
```go
if currentPos+dataSize > b.sizeInBytes && currentPos > 0 {
	if err := b.writeBatch(b.currentBatch, b.currentBatchKeys); err != nil {
		return errors.NewStorageError("error writing batch", err)   // returns here
	}
	b.currentBatch = make([]byte, 0, b.sizeInBytes)                 // skipped on error
	b.currentBatchKeys = make([]byte, 0, b.sizeInBytes)             // skipped on error
}
b.currentBatch = append(b.currentBatch, batchItem.value...)        // skipped on error
```

The worker only logs that error and moves on to the next queue item (lines 125-128), so the dequeued item is never retried:
```go
err = b.processBatchItem(batchItem)
if err != nil {
	b.logger.Errorf("error processing batch item: %v", err)
}
continue
```

`writeBatch` runs under `context.Background()` (line 251):
```go
g, gCtx := errgroup.WithContext(context.Background())
```
and the actual store write uses that context (line 257): `b.blobStore.Set(gCtx, ...)`.

`Close` ignores its context and waits unconditionally (lines 309-323):
```go
func (b *Batcher) Close(_ context.Context) error {
	close(b.done)
	select {
	case b.notifyCh <- struct{}{}:
	default:
	}
	b.wg.Wait()   // no ctx, no timeout
	return nil
}
```
The worker's shutdown path flushes the final batch through `writeBatch` (line 148), which again calls the underlying store under `context.Background()`, so a hung store blocks `wg.Wait()` and hence `Close` forever.

## Not currently covered

The only test file, `stores/blob/batcher/batcher_test.go`, uses the in-memory store (`memory.New()`) exclusively and only exercises happy paths: `TestBatcher_Set`, `TestBatcher_SetFromReader`, `TestBatcher_BatchSizeLimit`, `TestBatcher_BatchingBehavior`, plus health and unsupported-operation checks. There is no store that fails a write, so nothing asserts what happens to an item when an overflow flush fails; there is no assertion that a failed batch is retried rather than dropped; and there is no store that blocks, so nothing asserts that `Close` returns within a deadline. This is genuinely new coverage, not a duplicate.

## Definition of done

Add two focused tests in `stores/blob/batcher/batcher_test.go`, each with a small local fake implementing the `blobStoreSetter` interface (`Health` and `Set`) — no shared harness needed.

1. Fail-then-recover (no silent loss). Build a fake store whose `Set` returns an error on the first call and succeeds afterwards. Configure a small `sizeInBytes` and feed enough items to force at least one overflow flush while the store is in its failing state, then let it recover, then `Close`. Assert that every blob's bytes eventually reach the store (concatenate what the fake actually received and check each value is present). With the current code this fails because the item held at the moment of the failed flush is dropped. The fix should ensure a failed flush does not lose the in-flight item — e.g. retry the flush, or keep the item so it is retried, rather than returning past the append with the batch unreset.

2. Close honours its deadline. Build a fake store whose `Set` blocks (e.g. waits on a channel that the test never closes, or sleeps well beyond the test deadline). Enqueue at least one item so a final-batch flush will run on shutdown. Call `Close` with a context that has a short timeout (e.g. `context.WithTimeout(ctx, 500ms)`) and assert `Close` returns near that deadline (guard the whole thing with its own overall test timeout so a regression fails loudly instead of hanging CI). With the current code this hangs. The fix should make `Close` respect its context (return when the context is done even if the worker is still blocked) and thread a cancellable context into `writeBatch`/the underlying `Set` instead of `context.Background()`, so shutdown cannot outlive the stop deadline.

Acceptance criteria a reviewer can check: both tests are present, both fail against the code as it stands today, and both pass after the fix; the fix does not weaken the existing happy-path tests.

## Scope

This ticket is self-contained and depends on no other hardening ticket. Both tests can be written with a minimal local fake `blobStoreSetter` inside the existing `batcher_test.go`; no shared test harness is required.

---
_Filed from the 2026-08-05 production-hardening test-map review. Evidence verified against `bsv-blockchain/teranode@27a0178fd`. Session tag: `hardening-map-2026-08-05`._

freemans13:
Closes #1412

## What happened

The batching wrapper around blob storage had a family of defects with one shape: the caller is told a write succeeded, and the bytes are then discarded. All are real, and I traced each line by line — but the honest headline is that **none of them is live**, and establishing that changed the shape of the fix.

**A write could be lost on the overflow path.** Callers are told success immediately, because the blob is only queued. The worker then dequeues an item, finds the batch would overflow, flushes the accumulated batch first — and on a failed flush returned *before* appending the item it was holding. That item was already off the queue, so its bytes were gone, leaving a log line and a caller who had been told "fine" long ago.

**And on the shutdown path.** The worker's final flush error was only logged, so `Close` returned nil while the batch it was holding had been thrown away: a caller told its store shut down cleanly having just lost everything still in memory.

**And after close.** `Set` accepted writes after `Close`, queueing them behind a worker that had already returned, so they sat there forever while the caller was told they were queued.

**And when the key offset overflowed.** Key records address each item with a `uint32` offset. Because this change retains a batch across flush failures, that batch can grow past the 4 GiB ceiling; the conversion then failed and returned with the item already dequeued. Same defect, fourth path — and one this change's own fix for the first path is what made reachable.

**Each item's recorded position could be wrong.** The offset was captured before the flush reset the buffer, so an item that triggered a flush landed at the start of the new batch but was indexed to its pre-flush position. That is corruption rather than absence, and it was not in the original ticket.

**Shutdown could hang forever.** `Close` ignored its context argument entirely and waited unconditionally, and the final flush ran detached from any deadline. Since the daemon closes stores *in sequence* against one shared budget, a single hung backend also denied the chain-state store its orderly close at the end of that list.

**`Close` was not idempotent.** Making it fallible invited retries, and a retry is exactly what panicked — close of an already-closed channel, during shutdown.

**A non-hash key panicked.** Keys were converted to a 32-byte hash unconditionally; the blockchain peer-registry store's key is the 13-byte string `peer-registry`, so pointing batching at it crashes the process rather than returning an error.

**A non-positive batch size panicked at startup**, in `make()`, rather than being rejected as the configuration error it is.

## Why this is dormant, and why that bounded the work

Batching is applied only when a store's URL carries `?batch=true`. **That string appears nowhere in the repository except as prose**: one code comment and two lines of the settings doc, one of which is an example. No `settings.conf`, no compose file, no k8s manifest, no test sets it.

More decisive: **the batcher is write-only.** Reads, existence checks, deletes and delete-height updates all return "not supported", `SetCurrentBlockHeight` is a no-op so a wrapped store's Delete-At-Height bookkeeping never advances, and no unpacker for its output format exists anywhere. So even a *successful* batched write is unreadable by Teranode. That inverts the failure story in the ticket — enabling batching on subtree or block storage does not silently lose data, it breaks the node loudly on the first read. The one store whose write pattern actually fits is the transaction store, which this project deliberately treats as best-effort and which defaults to discarding writes anyway.

So this is a documented-but-dead footgun, not a live durability bug. The defects are fixed because they are cheap and real. Deliberately **not** done: no durability machinery, no unpacker, no timed flush, and writes stay asynchronous — making them synchronous would delete the batcher's reason to exist.

## Changes

All in `stores/blob/batcher/batcher.go` and `stores/blob/factory.go`, plus the settings doc.

**The overflow path.** The flush error is now captured rather than returned early: the item is appended either way, and only a *successful* flush resets the batch and the offset. One edit fixes both the lost write and the wrong recorded position.

The trade-off is deliberate and stated in the code: against a persistently dead backend the in-memory batch now grows instead of shedding one item per flush. Memory growth beats silently discarding bytes the caller was told were written. Two consequences follow directly, and are worth stating plainly rather than leaving to be discovered. Because each write still triggers a retry of the whole accumulated buffer — a per-item retry cadence unchanged from before — the *total bytes attempted* against a dead backend grow quadratically with the outage rather than linearly. And retaining a buffer after an ambiguous failure makes writes at-least-once rather than at-most-once: if a flush partially succeeded, those bytes are sent again in the next successful flush.

**The uint32 offset ceiling.** The overflow test now lives in `shouldFlushBefore`, which also forces a flush when the offset would pass `MaxUint32` and key records are being written. The offset resets to zero and the item is appended normally.

This supersedes an earlier version of this branch, which chose to drop the item rather than append a value with no index entry addressing it. That was a real choice between two bad outcomes, but flushing first avoids both, and it caps the unbounded batch growth above at no cost to the healthy path. `Set` additionally rejects a value too large to be addressed by a key record at all, synchronously, rather than accepting it and discarding it in the worker. Between the two guards the conversion can only still fail when the backend is dead *and* the retained batch has already passed 4 GiB — a batch that is unwritable either way and that `Close` already reports as lost, so dropping one further item costs nothing that was still recoverable. That residue is documented where it happens, and it no longer swallows the flush error that says why the backend was unhealthy.

**Shutdown.** `Close` races its wait against its context and, on expiry, logs and returns an error naming how many bytes are unflushed. Abandoning is the deliberate half of a real trade-off: the data in flight is memory-only against a backend that is already hung, so waiting saves nothing, while hanging costs a clean shutdown elsewhere. The worker's final flush error is now handed to `Close` and reported alongside that byte count. Teardown runs under a `sync.Once` and later callers get the first call's result — a snapshot, so an abandoned flush that later completed is still reported as abandoned, which is documented rather than papered over. Note the daemon currently discards blob-store close errors, so the batcher's own log line is the operator-visible signal.

`Set` after `Close` returns an error. A `Set` racing `Close` concurrently can still slip through, because closing that window means serialising the write path, which costs more than this dormant path is worth — stated in the doc comment rather than implied.

**Guards.** A length check on the key turns the panic into an error. The factory rejects a non-positive `sizeInBytes` instead of letting it reach `make()`.

**Buffer ownership, now documented.** `Set` does not copy `value`; it enqueues the caller's slice and the worker appends it later. Every other blob store is synchronous and copies before returning, so this is the one backend where reusing that buffer after `Set` returns is unsafe, and nothing said so. Copying here instead would add a full copy per item to the only path this type exists to make cheap.

**Docs.** The package comment promised a timed background flush that does not exist, and claimed reads and `SetDAH` pass through to the underlying store when every one of them returns an unsupported-operation error. The settings doc warned against combining `batch=true` with `localDAHStore` — a parameter no Go file in the repo reads — described `sizeInBytes` as `int64` where the code uses `Atoi`, and said a batch is held until it *reaches* `sizeInBytes` when the flush actually fires when the *next* item would overflow. Corrected, including the claim that a zero size panics in `make()`: only a negative one does, zero instead degenerates into flushing after every item.

A later review pass found three more things the code did but the docs did not admit, each now stated where it happens: the package comment listed only two flush triggers after a third was added; the `writeBatch` orphan above; that `Close` handed an already-expired context can abandon the final batch on its first poll, because a `select` among ready cases picks at random; and that a `Set` slipping past the closed check is worse than dropped, since the queue is unbounded and its bytes are then held until the process exits.

## Testing

`go build ./...`, `go vet ./stores/blob/...`, `go test -race ./stores/blob/...` (all 10 packages), `go test -race -count=5 ./stores/blob/batcher/...` for flakes, and `make lint` (0 issues). `upstream/main` is merged in and all of the above was re-run on the merged tree.

Ten new tests, all in-package with no external services, each mutation-verified by reverting the change it pins:

- A store that fails its first write then succeeds: every value must still arrive. Reverting drops the triggering item and the test fails on exactly that value.
- With key records enabled and a batch size that forces a flush: each record's offset and size must actually address its own bytes. Reverting produces a record claiming offset 8 in an 8-byte batch.
- `Close` against a write that blocks forever must return an error inside its deadline. Reverting hangs until the test binary's own timeout fires.
- A failed shutdown flush must surface from `Close`. Reverting returns nil.
- A second `Close` must not panic. Reverting panics with "close of closed channel".
- `Set` after `Close` must error. Reverting returns nil.
- A 13-byte key — the real peer-registry key — must error rather than panic. Reverting reproduces the panic verbatim.
- A table test over the flush arithmetic, including the `uint32` ceiling. Neutering the ceiling clause fails exactly the one case that pins it. The arithmetic is tested directly rather than through `Set`, because reaching the ceiling for real would need a 4 GiB batch.
- A table test over `Set`'s oversize-value guard, whose predicate is now split out as `exceedsKeyRecordSize` for the same reason the flush arithmetic was. Relaxing the comparison to `>=` fails exactly the boundary case: a value of precisely `MaxUint32` bytes still fits a key record and must be accepted.
- A table test over `NewStore` with `batch=true` and a bad `sizeInBytes` — negative, zero, unparseable, valid, absent. There was no factory test file at all before this. It drives `NewStore` rather than `createBatchedStore`, so it exercises the path a settings URL actually takes, and asserts the accepted cases really are batched so they could not pass if `batch=true` were ignored. Removing the guard reproduces the original `makeslice: cap out of range` panic.

**One thing found while implementing, worth flagging:** the `Close` fix as first written introduced a data race, reading the batch length from the calling goroutine while the worker appended to it. The race detector caught it; it is fixed with an atomic mirror of the length. That value is approximate by nature — fine for a log line, not to be relied on for anything exact.

## Limits

**Not reproduced against a real hung backend.** The blocking-store test proves `Close` gives up on schedule, but real file or object-store hang behaviour is not exercised here.

**The `uint32` guards are not covered by an end-to-end test**, because exercising them for real needs a 4 GiB allocation. Both predicates are tested directly instead, at their boundaries.

**The configuration claim is what actually keeps this safe**, and it is not enforced by anything. If it is worth guarding, the assertion to add is that no store URL in the shipped settings carries `batch=true` — that is the property whose violation is the entire risk. I have not added it, because it is a different kind of test from the rest of this change.

**Known and deliberately left alone**, since none is a live defect and each would widen the diff past what a dormant path justifies: a batched store never closes the store it wraps, because the interface it holds exposes only `Health` and `Set`; both error returns in `createBatchedStore` abandon an already-constructed base store without closing it; an item larger than the whole batch size is not flushed on arrival, so the configured size is not strictly a memory bound; and `writeBatch` writes batch-data and batch-keys as two independent calls, so a partial failure plus a retry under a fresh key can leave an orphan file. That last one is a consequence of this change's own decision to retain the batch, so it is now documented on `writeBatch` rather than only here; making it atomic would need a two-phase write.

**The daemon still discards blob-store close errors.** Threading that through was out of scope, but it means the new error is only visible in logs.

https://claude.ai/code/session_01VYgzH1eA9ZWptkuQ1rce8d


freemans13:
**Review round 1: 0 confirmed of 4 raised.** Two refutations conceded honest caveats, which are now in the description rather than left implicit.

**The lens I most wanted returned empty**, as has happened on each of the last four PRs, so I checked its subject myself: what becomes of the worker after `Close` abandons its wait. The waiting goroutine exits when the worker does, so it only outlives `Close` when the worker is genuinely hung — which is inherent to a hung backend, not something this change causes. A worker still writing after an abandoned close is honest, because the caller was explicitly told the flush was abandoned. Closing twice would panic on an already-closed channel, but that line is verbatim on `main`, so it is pre-existing rather than introduced here.

**Both refutations did better work than the findings they overturned.** The first claimed that not resetting on failure turns every later write into a re-flush of the whole growing buffer. The verifier built a probe — a separate module compiling the old and new code side by side against an always-failing store — and measured the retry cadence as *identical*: ten full-buffer attempts either way. The per-item retry is upstream's behaviour; only the payload size is new. It also checked the escalation that would have made the finding serious, that recovery becomes impossible past S3's single-upload ceiling, and found this code routes through multipart upload, so no such wall exists.

The second claimed a partial flush failure now duplicates already-written data. Real mechanism, but it needs a second switch that is set nowhere in the repo — and the verifier noticed the finding contradicts itself: the configuration where duplication would be undetectable is precisely the one where the bug cannot occur, since there is only a single write in that shape.

**What I took from both**: the total bytes attempted against a dead backend grow quadratically with outage length rather than linearly, and retention makes writes at-least-once rather than at-most-once. Neither is a defect — both follow directly from choosing retention over dropping — but neither should be discovered by whoever first enables this. Both now stated in the description.

Verification: 298 tests across the blob packages under `-race`, `make lint` 0 issues.

freemans13:
**Review round 2: 0 confirmed. Converged** — marking ready for review, with one caveat about how that conclusion was reached.

The reviewer returned an explicit empty findings list, which is the shape of convergence. But its trace was thin — fourteen tool calls in about four minutes, for a brief that asked for three race-enabled runs of the blob packages plus vet and lint, which alone take longer than that. I do not think it did the work it was asked to do, and an empty answer from a reviewer that did not look is indistinguishable from one that looked and found nothing.

So I ran that verification myself rather than bank it: `go test -race ./stores/blob/... -count=3` green across all ten packages, `go vet` clean, `make lint` 0 issues. On the substance I had already checked the concurrency question by hand in round 1 — the fate of the worker after `Close` abandons its wait — which was the part I was least sure of.

I am recording this rather than quietly marking it ready, because "the reviewer said nothing" and "there is nothing to say" are different claims and only the second one is load-bearing. Treat this PR as having had one thorough round and one that I backfilled, not two independent ones.

Round 1, for contrast, was substantial: four findings, none confirmed, but two refutations that measured rather than argued — one compiling old and new side by side against a failing store to show the retry cadence is unchanged, the other catching that a claimed duplication bug cannot occur in the only configuration where it would matter. Two honest caveats came out of them and are now in the description.

gokutheengineer (commented):
## Review of PR 1523 (blob batcher: surface dropped-write errors)

The direction is right (stop silently dropping batched blobs), but a few issues undermine the fix. Issues 1-3 I'd consider blocking; 4-5 are strongly recommended follow-ups.

### Issue 1: Retained batch grows unbounded during a backend outage (OOM risk)
`stores/blob/batcher/batcher.go:228`

On flush failure, `processBatchItem` now retains the batch and keeps appending, with no size cap, no backpressure to `Set` callers (which never block or error), and no health signal beyond per-flush error logs. During an extended blob-store outage at production write rates, `currentBatch` grows without bound until the process is OOM-killed, losing all buffered batches plus in-flight state. That's a strictly worse outcome than the bounded data loss this change was avoiding. Suggest a cap (bytes or batch count) after which Set returns an error or blocks, so callers get backpressure instead of a dead node.

### Issue 2: IntToUint32 overflow path writes data without a key record and masks the flush error
`stores/blob/batcher/batcher.go:251`

In the `writeKeys` path, a `safeconversion.IntToUint32(currentPos)` failure returns after the value bytes were already appended to `currentBatch` but before the key record is appended, and the early return also masks `flushErr`. The unbounded-growth behavior from Issue 1 is exactly what makes `currentPos > math.MaxUint32` reachable: once the retained batch exceeds 4GiB, every subsequent blob lands in batch-data with no offset record, permanently unretrievable by key, with no error surfaced to the writer.

### Issue 3: Set/Close race still silently drops a write while returning nil
`stores/blob/batcher/batcher.go:487`

The `closed` check in `Set` is check-then-act: a goroutine can pass `b.closed.Load()`, get descheduled, and enqueue after the worker has already finished its shutdown drain. `Set` returns nil, but nothing will ever dequeue the item. That's the same "told success, dropped the bytes" defect class this PR sets out to eliminate, just with a narrower window. The code comment documents and accepts this; worth either closing the window (e.g. re-check closed after enqueue and drain once more) or at least having Close do a final post-drain queue sweep.

### Issue 4: Partial writeBatch failure duplicates batch data on retry
`stores/blob/batcher/batcher.go:305`

`writeBatch` writes batch-data and batch-keys concurrently under a freshly generated random batch key per call. If the data Set succeeds but the keys Set fails, the whole batch is retained and the next flush rewrites the same data under a different key: permanent storage now holds a duplicate batch-data blob plus an orphaned key-less one, and DAH never reclaims either. Pre-existing in this function, but the new retain-on-failure behavior makes the retry (and thus the duplication) guaranteed. Reusing the same batch key across retries of the same batch would make the flush idempotent.

### Issue 5: The newly surfaced Close error is discarded at every production call site
`stores/blob/batcher/batcher.go:421`

`daemon.closeStores` does `_ = txStoreToClose.Close(ctx)` (same for subtree/temp/block/persister stores, `daemon/daemon.go:518-541`), and blockchain `Server.Stop` only logs it at Warn (`services/blockchain/Server.go:989`). So the headline fix never changes operator-visible behavior: a failed shutdown flush still ends with the daemon exiting 0 and a clean-looking shutdown. The callers need to propagate or at least loudly surface this error for the fix to land end to end.

ctnguyen (approved):
Review at commit [7653d00](https://github.com/freemans13/teranode/commit/7653d00c68773ac08b2185753ca2ac8bbd66c06f)

This is a careful, honest piece of work. All four defects are real, the fixes are correct, and the reasoning behind each is spelled out where a future reader needs it rather than buried. The overflow fix in `processBatchItem` is especially neat: capturing `flushErr` and only resetting the batch/offset on success repairs both the dropped write and the stale key offset in one edit, and it keeps the per-item retry cadence identical to before. The tests are the strongest part — each is mutation-verified, the fakes are minimal and in-package, and the `writeKeys` offset test actually decodes the key records and addresses back into the batch data rather than trusting a length. The error constructions type-check against the `errors` package wrapping convention (trailing `error` is extracted as the wrapped error, `%d` consumes the remaining `unflushed` arg), and I confirmed the `daemon.closeStores` claims in the `Close` doc are accurate: it does use one shared 30s budget, closes sequentially, and discards blob-store close errors, so the shutdown-abandon trade-off is justified exactly as written. The author's own Limits section already names the one real follow-up — asserting no shipped store URL carries `batch=true`, the property whose violation is the whole risk — and correctly scopes it out; I have nothing to add there. Nothing here blocks merge; the single note below is optional.

## Non-blocking issue

### ChiR1: Stale factory doc comment still promises the timed flush this PR just disowned
**Problem:** `stores/blob/factory.go:93-96` (`createBatchedStore`) documents URL params `batchSize`, `batchInterval`, and `batchWorkers`, none of which the code reads — it only parses `sizeInBytes` and `writeKeys` (lines 110-123). `batchInterval` in particular describes the "maximum time before a batch is processed" timed flush that does not exist, which is the exact false claim this PR corrects in the `batcher.go` package comment.

**Why it matters:** This PR's stated intent includes making the docs describe what the code actually does. The package comment was fixed; its sibling in the factory that constructs the batcher was not, so a reader who starts at the wiring point still sees the debunked timed-flush story and three parameter names that silently do nothing.

**Fix:** Trim the comment to the params actually honoured:

```go
// The batcher is configured through URL query parameters:
//   - sizeInBytes: Maximum accumulated batch size in bytes before a flush (default: 4 MiB)
//   - writeKeys:   Whether to write a separate key index alongside each batch
```

gokutheengineer (approved):

PR #1524, opened 2026-08-07
ci: measure coverage on the legacy wire code that peers can reach

### LINKED ISSUE #1415: [HARDEN-0805] Narrow the Sonar coverage exclusion so legacy wire code is gated
## Summary

Teranode's Sonar configuration tells the coverage-based quality gate to ignore the entire `services/legacy` tree. That directory is where the node speaks the peer-to-peer wire protocol to untrusted remote nodes (the P2P server, per-peer message handling, block/tx download and relay). Because the whole subtree is on the coverage-exclusion list, any code there — including newly added or modified message-handling code — never counts toward the coverage number the quality gate checks. A pull request can add un-tested handling of attacker-supplied wire data to the most exposed part of the codebase and the gate will not notice. This ticket asks to narrow that exclusion so coverage of the genuinely maintained legacy wire code becomes visible and gated.

## Why it matters

The quality gate here is not advisory. A required commit status called "SonarQube Quality Gate" is posted on each PR's head commit and is what branch protection evaluates (see `.github/workflows/sonar-pr-analyze.yaml`). When Sonar's "coverage on new code" condition is part of that gate, a PR that adds uncovered lines normally gets blocked. But coverage exclusions remove files from that measurement entirely, so lines in an excluded path are treated as if they don't exist for coverage purposes.

The danger is that this exemption covers exactly the code an attacker can reach without authentication: `services/legacy/peer_server.go` (140 KB of peer-server logic), `services/legacy/Server.go`, the `netsync` block/transaction download loop, and the per-peer message plumbing under `services/legacy/peer`. These parse and act on bytes sent by remote peers. This is the class of code where a missing bounds check or an unhandled message shape turns into a crash or a resource-exhaustion vector. The failure is silent: nothing breaks today, no test goes red, no alarm fires. The risk simply stays invisible — a reviewer looking at the green gate has no signal that a legacy wire-handling change shipped with zero new test coverage.

There is a legitimate reason the exclusion exists: `services/legacy` also contains vendored, rarely-touched third-party libraries (`bsvec`, `bsvutil`, `txscript`, `addrmgr`, `connmgr`) forked from btcd/bsvd. Measuring coverage on those would add noise. So the fix is to *narrow* the exclusion to the vendored subpackages, not to delete it wholesale.

## Evidence in current code

`sonar-project.properties:17` (at 27a0178fd) lists the whole subtree in the coverage-exclusion set:

```
sonar.coverage.exclusions=poc/**,ui/dashboard/**,k8sresolver/**,cmd/**,compose/cmd/**,tracing/**,internal/profiling/**,compose/chainintegrity/**,**/benchmark.go,scripts/validate-markdown.py,services/legacy/**,model/TestHelper.go,util/kafka/kafka_producer_async.go
```

The `services/legacy/**` glob covers every Go file under that directory, including the attacker-facing wire code:

- `services/legacy/peer_server.go` — 140 KB peer-server message handling
- `services/legacy/Server.go` — 27 KB legacy service entry
- `services/legacy/netsync/` — block/transaction sync driven by remote peers
- `services/legacy/peer/peer.go`, `.../association_manager.go`, `.../stream_policy.go` — per-peer connection and message plumbing

That the gate is real and PR-blocking is shown in `.github/workflows/sonar-pr-analyze.yaml`: the `SonarQube Quality Gate check` step runs `sonarsource/sonarqube-quality-gate-action`, and the final step posts a commit status with `context: 'SonarQube Quality Gate'` to the PR head SHA, with `state = 'failure'` when the gate fails — i.e. branch protection can require it.

Note the exclusion is coverage-only: `services/legacy` is *not* in the general `sonar.exclusions` line (line 16), so Sonar still lints and analyses these files for bugs/smells — only their test coverage is exempted from the gate.

## Not currently covered

`services/legacy` does contain many test files (83 `*_test.go` files under the tree, e.g. `peer_server_test.go`, `peer_server_snapshot_tx_test.go`, `server_start_test.go`). Those tests run and can pass, but they do two things this ticket is *not* about:

1. They assert behaviour of specific functions. They do **not** assert anything about whether the codebase's coverage of legacy wire code is measured or gated — a config posture, not a runtime behaviour.
2. Their coverage output is discarded by the gate. Because of the `services/legacy/**` coverage exclusion, whatever these tests do or don't cover contributes nothing to the quality-gate number, and adding new *un*-covered legacy code cannot lower it.

So there is no existing test or check that fails when legacy wire coverage is invisible, and none that would flag re-broadening the exclusion. This is not duplicate work.

## Definition of done

1. **Narrow the exclusion.** In `sonar-project.properties`, replace the blanket `services/legacy/**` entry in `sonar.coverage.exclusions` (line 17) with exclusions scoped to the vendored third-party subpackages only, e.g.:

   ```
   services/legacy/bsvec/**,services/legacy/bsvutil/**,services/legacy/txscript/**,services/legacy/addrmgr/**,services/legacy/connmgr/**
   ```

   so that the maintained wire code — `services/legacy/*.go` (Server.go, peer_server.go, config.go, etc.), `services/legacy/netsync/**`, `services/legacy/peer/**`, `services/legacy/peer_api/**` — is once again measured. (Keep `**/*.pb.go` and mocks excluded via the existing generic patterns.) The exact vendored list should be confirmed against which subpackages are forked-and-frozen versus locally maintained; adjust the glob set accordingly and record the reasoning in a comment above the line.

2. **Add a guard so the exclusion cannot silently re-broaden.** Add a small Go test (e.g. `sonar_config_test.go` at repo root, or under an existing config-lint test package) that reads `sonar-project.properties`, parses the `sonar.coverage.exclusions` value, and asserts that none of the maintained legacy wire paths are excluded — concretely, that no exclusion glob matches `services/legacy/peer_server.go`, `services/legacy/Server.go`, `services/legacy/netsync/`, or `services/legacy/peer/`. The test fails if someone re-adds `services/legacy/**` (or an equivalent broad glob). Use `require` from testify.

   **Acceptance criteria a reviewer can check:**
   - The new test fails on the *current* `sonar-project.properties` (blanket `services/legacy/**` present) and passes after step 1's narrowing.
   - `services/legacy/peer_server.go` and `services/legacy/netsync` are no longer matched by any coverage-exclusion glob.
   - The vendored crypto/script subpackages remain excluded (no new coverage noise from `bsvec`/`txscript`).

3. **Report the surfaced backlog.** Expect the first Sonar run after narrowing to reveal a real coverage deficit on legacy wire code. Capture the resulting coverage figures in the PR description so the team can decide whether to set a ratcheting new-code coverage target for these packages. Raising overall legacy coverage is follow-up work, not part of this ticket's done-criteria — this ticket's job is only to make the gap visible and gated going forward.

## Scope

This ticket is self-contained and depends on no other hardening ticket. The change is a one-line edit to `sonar-project.properties` plus a small standalone Go test that parses that file; it needs no shared harness, no external services, and no coordination with other tickets. The follow-on effort of actually writing tests to raise legacy wire-code coverage is deliberately out of scope and can be tracked separately once the gate makes the backlog visible.

---
_Filed from the 2026-08-05 production-hardening test-map review. Evidence verified against `bsv-blockchain/teranode@27a0178fd`. Session tag: `hardening-map-2026-08-05`._

freemans13:
Closes #1415

## What happened

The coverage quality gate ignored the entire `services/legacy` tree. That is where the node speaks the peer-to-peer wire protocol to untrusted remote nodes — the peer server, per-peer message handling, block and transaction download. A pull request could add untested handling of attacker-supplied wire data to the most exposed code in the repository and the gate would stay green.

The exclusion exists for a legitimate reason: that tree also holds packages vendored from btcd/bsvd, where measuring coverage is noise. So this narrows it to those, rather than removing it.

## Why this is safe, established rather than assumed

The gate definition is not in this repository — it lives server-side. Rather than guess, the spike read it from the public API. It has six conditions: five on new code, and one on **overall** project coverage at an 80% threshold.

The decisive fact is *where each is evaluated*. Querying live pull requests, the overall condition comes back listed but with **no measured value at all**, while every new-code condition carries a real number. That is the service declining to evaluate it in pull-request context, not it passing — and #1518 proves the converse, failing its gate on new-code coverage of 74.4%.

So **existing uncovered legacy code cannot retroactively fail anything.** Only lines a pull request adds or changes enter the calculation. Which is exactly the behaviour the ticket wants: future changes to wire handling get gated, history does not.

The coverage data already exists — the test target instruments every package, because `make test` passes `-coverpkg=./...` — so Sonar was discarding real numbers at the exclusion line, not lacking them. Un-excluding produces measurements, not zeros.

## What changes

One line. `services/legacy/**` becomes seven explicit entries, so `netsync`, `peer` and `version` become measured while the vendored packages stay excluded by name.

Two further directories, `peer_api` and `testdata`, also stop being named by the coverage exclusion — but they contribute nothing either way, because the project-wide `sonar.exclusions` line already drops them from analysis entirely (`**/*.pb.go` and `**test**`). `peer_api` holds only generated protobuf and a `.proto` file.

There is a second, durable effect worth stating, because it outlasts the specific package list: **any new package added under `services/legacy` from now on is measured by default.** Under the old blanket pattern it would have been silently unmeasured on arrival. The exclusion now has to be widened deliberately, in a reviewable diff, rather than inherited by accident.

**Naming the vendored packages explicitly is load-bearing, and this is the trap worth knowing about.** `netsync` ships an upstream README and carries a btcsuite copyright header, despite being actively maintained here — three of its files have changed recently. Any rule derived from provenance markers, which is the obvious way to write this, would have silently kept it excluded and defeated the entire change while looking correct.

## The compromise, stated plainly

**The legacy root files stay excluded, and that is a compromise rather than a judgement about provenance.** They are the ticket's actual target: `peer_server.go` alone is 4,165 lines with 25 wire handlers, plus `Server.go`, `config.go` and `upnp.go`. Including that package takes the project under the 80% threshold and leaves the SonarCloud project permanently failing.

So this is the safe first step, and the root package needs a second one: raise its coverage first, then un-exclude. I would rather ship the part that is unambiguously safe than stall the whole thing on the hard part.

Nothing *consumes* a failing overall condition: the main-branch job runs the scanner but has no quality-gate step, and the gate is not a required check (the ruleset requires `test`, `golangci-lint`, `filename check report` and `gitleaks`, and nothing else). So it would not turn CI red — it would leave the dashboard red, which I read the never-permanently-red principle as covering in spirit. Asking an administrator to lower the threshold to make room would be weakening a gate to land a change, which is worse.

## The numbers, measured rather than modelled

An earlier version of this description carried two figures that do not survive measurement, and both corrections point the same way — there is more headroom than I claimed, and the root package is less catastrophic than I claimed, though still too expensive.

The numbers below come from the actual `coverage.out` that CI hands to Sonar, downloaded from this pull request's own run, with the project's analysis and coverage exclusion rules applied to it. Each block is counted covered if any test binary reached it, which is how a merged Go profile is read.

| Package | Statements | Coverage |
|---|---|---|
| `netsync` | 1,886 | 59.6% |
| `peer` | 1,176 | 73.8% |
| `version` | 1 | 100% |
| **un-excluded by this PR** | **3,063** | **65.1%** |
| `services/legacy` root — *stays excluded* | 1,928 | 29.8% |
| `addrmgr` — *stays excluded* | 544 | 69.7% |

**The root package is at roughly 30%, not the 14.6% I previously reported.** That earlier figure was the one place where I "corrected" an estimate with a worse measurement — almost certainly by running only that package's own tests, which misses the coverage other packages' tests contribute through `-coverpkg`. Weighting by source lines instead of statements moves it to 31.7%, not down. The decision does not change, but the stated reason for it should be right. The commit message still carries the pre-measurement figures, 27% for the root package and 1.9 points of headroom, because it was written before the profile was pulled. The table above is the measured version and supersedes it; the commit stands as written because this branch is never rebased.

**The margin is wider than I claimed, not thinner.** Main now sits at **82.7%** overall on the dashboard, up from the 81.9% this description was originally written against. On the profile above, un-excluding these three packages costs about 0.7 points of line coverage; scaled to Sonar's own denominator, which is around a third smaller than the profile's, call it **roughly one point — landing near 81.8%, with about 1.8 points of headroom.** The earlier "roughly half a point" was pessimistic by about three times.

That also revises the shape of the next step. Bringing in the root package costs somewhere around 3 points on the same scaling, which still breaks the threshold — so it still waits. **`addrmgr` costs about a tenth of a point and is now comfortably affordable.** It is the strongest candidate for the second step, ahead of the root package: it ingests `ADDR` messages from untrusted peers, so it sits squarely inside this ticket's scope, and its package documentation has been rewritten in Teranode's own style with the btcsuite copyright removed — it is not vendored in the way `connmgr` next to it still visibly is. It stays out of *this* change only to keep the diff to the packages the ticket names, not because of the margin.

`txscript` is the other judgement call and I want a reviewer to be able to disagree in the open. It is vendored in shape but has had three consensus-behaviour fixes in the last year. I have left it excluded because this ticket's scope is the wire protocol rather than the script engine — but that reasoning is a preference, not a fact.

## Testing

There is nothing to run: this is a scanner configuration change with no code impact. `git diff` is one line. The gate definition and its six conditions were read from the live API rather than assumed, the `-coverpkg` and workflow claims were checked against the Makefile and the workflow files, and the coverage figures were computed from the real CI profile as described above.

## Limits

**The gate lives outside version control.** Anyone with administrative access can change thresholds or conditions without a commit here, so the safety argument is true today and is not pinned by anything in this repository.

**This pull request's own check cannot confirm the coverage impact,** for two independent reasons. The overall condition is not evaluated in pull-request context at all; and the analysis workflow deliberately reads the scanner configuration from the *base* commit as a supply-chain precaution, so this run scans with the old exclusion list regardless. The first run that applies the new list is the post-merge scan of main. If the number comes in lower than expected, the answer is to pull `peer` or `netsync` back out — the same one-line edit in reverse — rather than to lower the threshold.

**The scaling between the profile and Sonar's own metric is approximate.** The profile counts more coverable lines than Sonar does, so the point-costs above are converted, not read off. They are good enough to rank the packages and to tell one point from three, which is what the decisions turn on, and not good enough to promise a specific post-merge figure.

https://claude.ai/code/session_01FW2wZf3xSg7feUxKUCopVT


freemans13:
**Review round 1: 0 confirmed of 2 raised** — but both refutations corrected numbers in my description, and one of the corrections matters.

**The margin is roughly half a point, not 1.9.** I had taken 81.9% as the starting point and asked what un-excluding the root package would cost. That ignores the effect of this PR itself: what it un-excludes — netsync, peer and version, 3,061 statements at 61.9% — is well below the project average, so it drags the total down to around 80.5% before anything else is considered. Still green, but with far less room than I implied, and anyone adding a third package to the measured set should know that. Now stated in the description.

**And the root package is at 14.6%, not the ~27% I wrote** — measured from a real coverage profile rather than inferred. That strengthens the decision to leave it excluded rather than undermining it, but it was wrong and is corrected.

Both findings were refused on their conclusions rather than their data, which is the right call in both cases. The first argued that the well-covered vendored packages are themselves the missing headroom, so excluding them is self-inflicted. The verifier reproduced every statement count to the digit, then showed the conclusion does not follow: the whole bundle proposed for measurement comes to 75.1%, still below the base, so it pushes the number down rather than creating room. The blocker is not that btcd's crypto packages are excluded — it is that the wire code is genuinely under-tested.

The second observed that this PR's own check cannot exercise the change, because the analysis workflow deliberately reads the scanner configuration from the base commit as a supply-chain precaution. True, and now noted in the description — though the verifier pointed out the causal story is over-determined, since the overall condition is not evaluated in pull-request context regardless.

**All four config checks came back clean, and were genuinely verified rather than assumed:** the wildcard semantics were checked against Sonar's documentation and pattern API (single `*` does not cross a directory separator, so the root-files entry does not recurse); all eleven subdirectories are accounted for exactly once, including the two that fall to other exclusion lines; a future new subdirectory would be measured by default, which is the safe direction for this gate; and the line parses with no stray or duplicate entry.

freemans13:
**Review round 2: 0 confirmed. Converged** — marking ready for review.

An explicit empty findings list from a completed reviewer (checked in the run's journal, since a crashed agent returns the same shape), with 26 tool calls against a one-line change — proportionate for re-checking a config line and a set of prose claims, unlike the thin run I backfilled on an earlier PR.

Two rounds. Round 1 raised two, confirmed none, and corrected two numbers in my description — the headroom after this change is about half a point rather than 1.9, because I had failed to account for the drag of the packages this PR itself un-excludes; and the root package measures 14.6% rather than the ~27% I had inferred. Round 2 found nothing further.

Worth restating what this does and does not do, since it is one line and easy to over-read. It makes future changes to `netsync`, `peer` and `version` count toward the coverage gate. It does **not** cover `peer_server.go`, which is the file the ticket is really about — that needs its coverage raised first, and is called out in the description as a second step rather than quietly omitted.

freemans13:
**Review round 3: 0 findings across all three surfaces. Converged.**

Merged `upstream/main` (clean, no conflicts) and re-ran the loop against the new head `bafcf26`.

- **Inline review comments:** none.
- **Claude Code Review:** complete, no issues found, no inline comments. It re-verified the directory accounting and the glob semantics against the tree.
- **Copilot:** did not review. Five consecutive attempts have come back with "Copilot was unable to review this pull request because the user who requested the review has reached their quota limit." The most recent two are from today. The re-request no longer even registers the reviewer on the PR, so this is a quota block rather than a slow response.
- **SonarCloud:** quality gate passed, 0 new issues, 0 accepted issues, 0 security hotspots.
- **CI:** 36 checks green, 1 skipped (aerospike fork image), 0 failures.
- **Benchmarks:** 139 benchmarks compared, 0 regressions, 0 improvements.

I also re-checked the load-bearing claim independently rather than taking the bot's word for it. `services/legacy` holds eleven subdirectories: `addrmgr, blockchain, bsvec, bsvutil, connmgr, netsync, peer_api, peer, testdata, txscript, version`. Six are named in the new exclusion list, plus the fourteen root `*.go` files via the non-recursive glob. That leaves `netsync`, `peer` and `version` newly measured. `peer_api` holds only `peer_api.pb.go`, `peer_api_grpc.pb.go` and a `.proto`, all dropped by the project-wide `**/*.pb.go`; `testdata` holds one Go file, dropped by `**test**`. Neither contributes coverage either way.

The one thing this PR's own run still cannot show is the coverage impact, for the two reasons already in the description: the overall condition is not evaluated in pull-request context, and the analysis workflow reads exclusions from the base commit. The post-merge scan of `main` is the first real signal.

freemans13:
**Review round 4: one minor finding fixed, everything else clean. Converged.**

Merged `upstream/main` into `stu/harden-1415-sonar-legacy` (clean, one incoming change: `services/legacy/netsync` nil-peer guard from #1682) and re-ran the loop against the new head `36b4189`.

**The one finding, from the Claude Code Review comment above.** It noted the commit message says the legacy root package is "roughly 27%" while the description's measured table says 29.8%. That is real. The commit message was written before the coverage profile was pulled from CI, and the description was corrected afterwards. This branch is never rebased, so the commit message cannot be edited. I have added a sentence to the description saying so explicitly, rather than leaving a reader to spot the gap and wonder which figure is live: the table is the measured version and supersedes the commit message's 27% and 1.9-point headroom estimate.

**Everything else on the three surfaces:**

- **Inline review comments:** none.
- **Claude Code Review:** complete, no issues found. It independently re-verified the directory accounting, the non-recursive `*` glob semantics, and the point about `version` carrying a btcsuite header while being genuinely Teranode code.
- **Copilot:** did not review. The request returned "Copilot was unable to review this pull request because the user who requested the review has reached their quota limit." That is the sixth consecutive attempt with the same refusal, so treat this PR as unreviewed by Copilot rather than passed by it.
- **SonarCloud:** quality gate passed, 0 new issues, 0 accepted issues, 0 security hotspots.
- **Benchmarks:** 139 compared, 0 regressions, 0 improvements.

**CI needed two reruns, and both failures were pre-existing flakes in packages this PR does not touch.** Worth recording since the diff is one line of scanner configuration and cannot plausibly cause either.

The first run timed out after 10 minutes in `services/blockvalidation/TestBlockValidation_ReportsInvalidBlock_OnInvalidBlock`, with a genuine data race underneath it: `stores/blob/memory` keeps `currentBlockHeight` as a plain `uint32`, written by `SetCurrentBlockHeight` (memory.go:347) from the subtree-validation blockchain listener and read unsynchronised by `ttlCleaner` (memory.go:104). That field dates to #3063 in June 2025 and is an ancestor of `upstream/main`, so it is upstream code and out of this PR's scope. It is worth a ticket of its own.

The second run failed `services/blockassembly/subtreeprocessor/Test_queueWithTime`, a wall-clock test with 200ms margins that flakes on a loaded runner under `-race`. Also upstream, from #563.

The third run was green: 36 checks pass, 1 skipped (aerospike fork image), 0 failures, and the SonarQube Quality Gate status flipped from "Upstream tests failure — Sonar not run" to "Quality Gate passed".

The limit already stated in the description still stands: this PR's own run cannot show the coverage impact, because the overall condition is not evaluated in pull-request context and the analysis workflow reads exclusions from the base commit. The post-merge scan of `main` is the first real signal.

freemans13:
**Review round 5: no findings on any surface, nothing to fix. Converged on `fd1ef83`.**

`upstream/main` was already merged into `stu/harden-1415-sonar-legacy` at `fd1ef83`, so there was nothing new to bring in and no push this round.

- **Inline review comments:** none.
- **Claude Code Review:** complete on `fd1ef83` (updated 11:30:50Z), no issues. It re-checked the directory accounting and the non-recursive `services/legacy/*.go` glob against the tree.
- **Copilot:** did not review. The re-request came back at 11:33:02Z with "Copilot was unable to review this pull request because the user who requested the review has reached their quota limit." That makes this PR unreviewed by Copilot, not passed by it.
- **SonarCloud:** quality gate passed on the analysis of 11:37:25Z, with 0 new issues, 0 accepted issues and 0 security hotspots.
- **CI:** 36 checks pass, 1 skipped (aerospike native-ops, fork image), 0 failures. `make lint` locally reports 0 issues.
- **Benchmarks:** 151 compared, 0 regressions, 0 improvements.

One Sonar reading bears on what this PR set out to do. Queried directly, the pull-request gate for 1524 lists the overall `coverage` condition with no `actualValue`, while main's gate reports overall coverage at 84.1%. That fits the description: the overall condition is not evaluated on a pull request, so existing uncovered legacy code cannot fail this one. It also means this PR's green gate says nothing about the coverage impact. The scan runs with the base commit's exclusion list, and the first scan to measure `netsync`, `peer` and `version` will be the one on main after merge. Main has moved from the 82.7% the description quotes to 84.1%, so the roughly one-point cost estimated there now leaves more headroom than the description says.

oskarszoon (approved):
Verified at `fd1ef83b1`. Does what the title says: `netsync` (6 files, 32 test files) and `peer` (10 files, 11 test files) move from unmeasured to measured, and both are already substantially tested, so this counts existing work rather than opening a coverage cliff.

Checked the two packages that fall through the new enumeration without tests — `peer_api` (1301 lines of generated protobuf) and `testdata` (a helper plus `.bin`/`.hex` fixtures). Neither can drag the ratio down: `sonar.exclusions` at line 14 is evaluated ahead of `sonar.coverage.exclusions` and already removes both via `**/*.pb.go` and `**test**`.

Two consequences worth expecting rather than acting on:

- The enumeration fails open to measurement. A future `services/legacy/<newdir>` is now included automatically where the blanket glob excluded it. Safer default for a gate, but it is a behaviour change beyond the two named packages.
- Overall project coverage will step on the first run now that `netsync` and `peer` count. Worth reading the first CI result before reacting to it.

Top-level `services/legacy/*.go` stays excluded while `netsync`/`peer` become measured, though `feeler.go` and `peer_registry_sync.go` are also peer-facing and do have tests. Fine as an incremental step — flagging only in case the intent was "all peer-reachable legacy code".

ctnguyen (approved):
Review at commit [fd1ef83](https://github.com/freemans13/teranode/commit/fd1ef83b1b464e77381265456ee403688258bd3f)

One line, and a good one. Blanket-excluding `services/legacy/**` from coverage meant the most externally-exposed code in the repo — the peer-to-peer wire path — could grow untested handling of attacker-supplied data with the gate still green, and this narrows the exclusion to the vendored packages plus the root package that the description deliberately holds back for a second step, instead of removing it wholesale. I checked the load-bearing mechanics rather than the prose: `services/legacy` has exactly the eleven subdirectories the description accounts for, six are named in the new list, `peer_api` holds only `peer_api.pb.go`, `peer_api_grpc.pb.go` and the `.proto` (all dropped by `sonar.exclusions=**/*.pb.go`), and `testdata/test_helper.go` is dropped by `**test**` — so `netsync`, `peer` and `version` are what actually becomes measured. `Makefile:180` does pass `-coverpkg=./...`, so the profile already carries real numbers for those packages and un-excluding yields measurements rather than zeros. `.github/workflows/sonar-pr-analyze.yaml:240` really does read the properties file from the base commit, so this PR's own scan cannot exercise the new list, exactly as the description says; and the main-branch `sonarqube` job ends at the `SonarQube Scan` step (`.github/workflows/teranode_main_tests.yaml:186-189`, the last lines of the file) with no quality-gate action after it. I did not run the test suite, so nothing below rests on a measured coverage figure, and I took the SonarQube single-`*` wildcard semantics and the branch ruleset's required-check list on report since neither is readable from this tree. Also worth saying: naming the vendored packages explicitly rather than deriving them from provenance markers is the right call and the description is right about why — `services/legacy/netsync/manager.go:1` carries a btcsuite copyright header and `services/legacy/netsync/README.md` is still the upstream bsvd readme, yet `git log -- services/legacy/netsync/` is wall-to-wall Teranode-specific work. A provenance-based rule would have looked correct and quietly defeated the whole change. Two notes, neither of them about the mechanics.

## Non-blocking issues

### ChiR1 — `Closes` auto-closes the hardening ticket while its stated primary target stays excluded

**Problem:** The PR body says `Closes #1415`, so merging will close the hardening issue. But the description also says of the legacy root files that "**They are the ticket's actual target**", and that the root package "needs a second one: raise its coverage first, then un-exclude". That root package is the big one: `peer_server.go` is 4,641 lines at this head, with 25 wire handlers (`grep -c "^func (sp \*serverPeer) On" services/legacy/peer_server.go`). By the author's own reading the ticket is partially delivered, and no follow-up issue is referenced anywhere in the PR or its linked items.

**Why it matters:** The remaining work is the larger and more exposed half. Besides those handlers, `addrmgr` — which the description itself rates as the strongest candidate for step two and as affordable at roughly a tenth of a point — ingests untrusted `ADDR` payloads: `serverPeer.OnAddr` at `services/legacy/peer_server.go:2047` feeds `msg.AddrList` straight into `s.addrManager.AddAddresses` at line 2088. That is squarely inside [#1415](https://github.com/bsv-blockchain/teranode/issues/1415)'s scope and it stays unmeasured. Once the ticket auto-closes, the only record that a step two was ever planned is this PR description.

**Fix:** Either open the follow-up issue for the root package and `addrmgr` and reference it in the body, or downgrade the keyword so the ticket survives the merge — `Refs #1415` / `Part of #1415` rather than `Closes`. The second is a one-word edit and costs nothing.

### ChiR2 — the new exclusion list encodes a non-obvious classification with no in-file rationale

**Problem:** `sonar-project.properties:17` now carries seven `services/legacy/...` entries with nothing to say why each is there. Two different reasons are mixed into one flat list: six subdirectory entries that mean "vendored, coverage is noise", and `services/legacy/*.go`, which the description explains is a temporary compromise pending a coverage push. Nothing in the file distinguishes them.

**Why it matters:** This is exactly the knowledge the PR establishes is not recoverable from the tree. `netsync` looks vendored and is not; `addrmgr` reads as one of the vendored six but its package doc at `services/legacy/addrmgr/addrmanager.go:1` has been rewritten in Teranode's style with the btcsuite header gone. The next person to touch this line has a list of paths, no rationale, and a broken build or a coverage dip pushing them toward "just add the failing package back". Adding `services/legacy/netsync/**` would silently undo the change and read as routine maintenance — which is the same trap the PR was written to avoid, relocated from the pattern syntax to the reviewer's head. The file already sets the precedent for explaining a decision in place: `sonar-project.properties:23-26` carries a four-line comment justifying the `go:S3776` test-file exemption.

**Fix:** Add a short comment block above the line. Something like this is enough — note the properties format needs the whole value on one physical line, so the comment has to carry the structure:

```
  sonar-project.properties, above line 17:
  Legacy exclusions are two different things. bsvec, bsvutil, txscript,
  addrmgr, connmgr and blockchain are vendored from btcd/bsvd — coverage
  there is noise. services/legacy/*.go is the root package and is a
  temporary compromise: including it drops the project under the 80%
  threshold, so raise its coverage first, then drop this entry.
  Do not add netsync or peer back — they are actively maintained here
  despite the btcsuite headers, and they are the wire path this list
  exists to keep measured.
```

## Recap

|---|---|---|---|
| ChiR1 | Auto-closes ticket with work remaining | | 35% |
| ChiR2 | Exclusion list lacks in-file rationale | | 25% |

PR #1526, opened 2026-08-07
fix(blockvalidation): bound the detached subtree_data fetch

liam:
Closes bitcoin-sv/teranode#4743.

## The problem

`fetchAndStoreSubtreeData` detaches from sibling cancellation on purpose. The existing comment explains why: the function runs in a per-subtree goroutine inside an errgroup, so using `gCtx` means one sibling failure cancels every in-flight download in the batch, and each cancellation makes the serving peer abort its on-demand creation and discard Aerospike work already paid for.

The problem is that `context.WithoutCancel` returns a context with **no deadline and a nil `Done()` channel**, and both matter downstream:

- `DoHTTPRequestBodyReaderWithRetry`'s only abort is `case <-ctx.Done()`. A nil channel can never be selected, so `time.After` always wins and the loop runs to `maxAttempts`.
- `doHTTPRequestForStreamingWithRetryAfter` installs its streaming timeout **only when the context carries no deadline**, so each attempt started a fresh one.

A peer answering 503 and then dripping bytes held one fetch for `maxAttempts x http_streaming_timeout`, which is six times ten minutes on the committed config.

## The fix

One deadline on the detached context resolves both, because the per-attempt timeout is conditional on there being no deadline:

```go
ctx, cancelDetached := context.WithTimeout(context.WithoutCancel(ctx), timeout)
defer cancelDetached()
```

`Done()` fires again for the retry guard, and every attempt now shares this single bound. The detachment from sibling cancellation is unchanged.

The repo already uses this idiom in `services/asset/centrifuge_impl/centrifuge.go`, which pairs `WithoutCancel` with a `WithTimeout`. `get_blocks.go` was the only site that detached without re-imposing a bound.

The value is a new setting defaulting to **10m**, which is what the existing comment already assumed was happening ("or hit its own `http_streaming_timeout`"). A non-positive value falls back to the default rather than meaning "no limit", so a misconfigured or unparsed value cannot silently restore the unbounded fetch.

## Two things worth flagging

**The deadline now also covers the parse and the store write**, which the detached context reached before and which previously could not be interrupted. I checked the blob store: `SetFromReader` writes via `createTempSibling` and renames, so an interrupted write leaves no partial blob for the `Exists` check at the top of the function to trip over. Worth a second opinion on the non-file backends.

**A correction to the issue, and to something I wrote on bitcoin-sv/teranode#4742.** I previously said that bounding the non-2xx error body in bsv-blockchain/teranode#1453 would substantially defuse this, because the uncapped read was what let an attempt consume its full ten minutes. That is wrong. The cap bounds memory, not time: `io.ReadAll(io.LimitReader(body, 2048))` returns once it has 2KB, so a peer that drips fewer than 2KB slowly still burns the whole per-attempt window. This issue was undiminished by that merge.

Also a refinement to the issue's description: the retry loop only iterates on 503 (`if !errors.Is(err, errors.ErrServiceUnavailable) { return nil, err }`), so the six-attempt multiplier needs the peer to answer 503 every time. Trivially achievable, but it does not happen on arbitrary failures.

## Verification

```
go build ./services/blockvalidation/... ./settings/...                -> clean
go vet ./services/blockvalidation/ ./settings/                        -> clean
gofmt -l <changed files>                                              -> clean
gci diff --skip-generated -s standard -s default <changed files>      -> clean

go test -count=1 ./settings/...
  ok  github.com/bsv-blockchain/teranode/settings                            0.139s
go test -count=1 -timeout 1800s -tags testtxmetacache ./services/blockvalidation/...
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation            275.547s
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation/catchup      8.409s
```

The test asserts on the **attempt count**, not wall-clock alone, because a timing-only assertion would still pass if the loop ran to completion quickly. Mutation: removing the deadline makes all six attempts run and fails it on exactly that assertion (`"6" is not less than "6"`, 7.86s versus 0.70s with the fix).

`-race` and the full repository suite were not run locally.

## Also in this PR

The comment at `get_blocks.go` referred to a "companion fix in `services/subtreevalidation/check_block_subtrees.go`". No `WithoutCancel` remains in that package, so the reference is stale and the rewritten comment drops it. That path uses the `CheckBlockSubtrees` server context, which already carries `CheckBlockSubtreesTimeout`, so the compounding described above never applied there.

A second commit takes the review follow-ups: an expiry of the new bound is re-attributed to the peer so it does not read as a local failure, the setting is context-overridable and its default lives in one place, and `settings.conf` gains the key. See the review reply for detail.


ordishs (approved):
Approving. The diagnosis and the fix both hold up against the code, and the worst-case hold on one detached fetch drops from ~60m to 10m. No P0/P1 findings.

Verified directly rather than taken on trust:

- `util/http.go:519-523` — `maxAttempts: 6`, 250ms doubling capped at 5s; `util/http.go:567-570` — `case <-ctx.Done()` is the loop's only abort, so a nil `Done()` really did let every attempt run.
- `util/http.go:585-588` — the per-attempt streaming timeout is installed only `if _, ok := ctx.Deadline(); !ok`, so one injected deadline correctly bounds all attempts; with `settings.conf:636 http_streaming_timeout = 600000` the pre-fix worst case was 6 × 10m.
- `get_blocks.go:580-586` — the fail-closed branch is load-bearing, not just defensive: this package builds bare `&settings.Settings{}` in several tests, where a 0 timeout would otherwise yield an already-expired context.
- `get_blocks_test.go:952` — the test discriminates against today's code: reverting only the deadline fails it with `get_blocks_test.go:981: Error: "6" is not less than "6"`. Passes 5/5 under `-race` at ~0.7s; `go test -race -run TestFetchAndStoreSubtreeData ./services/blockvalidation/` → ok, 2.919s.
- Removing the `check_block_subtrees.go` companion-fix reference is correct: no `WithoutCancel` remains in `services/subtreevalidation`, that path uses the `CheckBlockSubtrees` server context which already carries `CheckBlockSubtreesTimeout` (`Client.go:161-163`), so the nil-Done/fresh-per-attempt compounding never applied there.

On your open question — the deadline now covering parse and store write: **safe on every backend reachable from this call site.** `file` (what all committed contexts select, `settings.conf:1156-1163`) consults ctx only at `acquireWritePermit` (`file.go:723`); the write is `io.Copy` from a `bytes.Reader` into a temp sibling, fsync'd and atomically renamed (`file.go:777-797`), so the write either is refused up front or completes in full. `memory` reads ctx only in the `!AllowOverwrite` branch this call site does not take. `s3` only publishes on upload completion, and finite-DAH blobs are never routed there (`s3.go:272-274`) — this call site always passes `WithDeleteAt(dah>0)`. `http` aborts the POST, and the far end's `io.Copy` errors on the truncated body and discards the temp (`server.go:444`). `batcher.Set` ignores ctx entirely. No partial blob is ever visible to the `Exists` check at `get_blocks.go:550`.

Two P2 follow-ups (not blockers) and three nits are in the full review: chiefly that an expiry now surfaces as `context.DeadlineExceeded`, which `errors.IsLocalError` treats as ours — suppressing alternative-peer failover (`get_blocks.go:749-751`), the `failedPeers` charge (`catchup.go:580`) and sync-peer rotation (`Server.go:1908-1916`) for a peer that answers 503 slowly enough to cross the bound mid-attempt. Worth a follow-up issue mapping that expiry to an `ErrExternal`, since a `Done()` on a detached context can only mean peer-controlled I/O. Nits: append `alternativeContext...` at `settings/settings.go:359`, and swap `require.Less(t, calls, 6)` (which encodes util's private `maxAttempts`) for `require.ErrorIs(t, err, context.DeadlineExceeded)` plus an upper bound.

freemans13 (commented):
## Review

The diagnosis here is correct, and I verified every step of it against the code. `context.WithoutCancel` really does strip both the deadline and the `Done()` channel; `doHTTPRequestBodyReaderWithRetry`'s only abort is `case <-ctx.Done()` on a nil channel, so it always ran to six attempts; and `doHTTPRequestForStreamingWithRetryAfter` really does install a fresh `http_streaming_timeout` per attempt precisely because the context carried no deadline (`util/http.go:585-588`). The committed `http_streaming_timeout` is `600000` ms, so the "six times ten minutes" figure is right. The one-line fix is the right shape and lands in the right place.

There is one problem worth blocking on, though, and it is not in the mechanics of the timeout — it is in what the timeout *means* to the code above it.

**The new deadline converts a peer failure into a local failure, which switches off peer failover and stops the offending peer being charged.** That is the opposite of what you want from a fix whose whole purpose is to contain a hostile or broken peer.

The rest is smaller: the setting is not context-overridable, it never made it into `settings.conf` (where a dead near-identically-named key already sits waiting to mislead an operator), and one claim in the description contradicts the diff.

| Severity | Where | What |
|----------|-------|------|
| **High** | `services/blockvalidation/get_blocks.go:592` | Deadline expiry is classified as a *local* error, so failover to other peers is skipped and the peer is never charged |
| Medium | `settings/settings.go:359` | `getDuration` call omits `alternativeContext...`, so the setting cannot be overridden per settings-context |
| Low | `settings.conf` | New key absent from `settings.conf`; a dead, near-identically-named key (`blockvalidation_subtree_fetch_timeout = 120s`) sits there instead |
| Low | `services/blockvalidation/get_blocks.go:592` | The bound now also covers parse + serialize + store, which previously had no bound at all — interacts badly with the first finding |
| Info | three files | The 10m default is written out in three places with nothing enforcing agreement |
| Info | description | The "Not included" section says a stale comment was left alone; the diff removes it |

---

## High — the deadline turns a peer failure into a local failure

### What happens

`fetchAndStoreSubtreeAndSubtreeData` has an explicit early-out for local errors — the reasoning being that if *our* storage or *our* context broke, asking a different peer cannot help, so don't burn attempts on the fleet (`get_blocks.go:748-751`):

```go
if errors.IsLocalError(err) {
    return "", errors.NewServiceError("... Local error fetching subtree %s (not retrying with other peers) ...", ...)
}
```

`recordCatchupPeerFailure` has the matching guard (`catchup.go:580`): a local error "is ours, not the peer's, and must never land an innocent peer in `failedPeers`".

`errors.IsLocalError` returns true for any context error, and `IsContextError` unwraps through the whole chain (`errors/error_utils.go:236`). So when the new deadline fires, the retry loop returns `ctx.Err()` = `context.DeadlineExceeded`, `fetchSubtreeDataFromPeer` wraps it in a `ServiceError`, `fetchAndStoreSubtreeData` wraps that in a `ProcessingError` — and the result is still, all the way up, a local error.

### Evidence

I ran the real code path (a temporary probe test, since removed) with a 503 responder and a 600 ms `SubtreeDataFetchTimeout`:

```
err = PROCESSING (4): [catchup:fetchAndStoreSubtreeData] Failed to fetch subtreeData for 0000…cdab
   -> SERVICE_ERROR (59): [catchup:fetchSubtreeDataFromPeer] failed to fetch subtree data from http://test-peer/subtree_data/0000…cdab
   -> UNKNOWN (0): context deadline exceeded

IsLocalError   = true
IsContextError = true
```

### Why this is a regression, not just a pre-existing wart

Before this PR, the hostile-drip peer eventually exhausted all six attempts and the loop returned `errors.NewServiceUnavailableError("… still 503 after 6 attempts …")`. That is **not** a local error — I checked that too; `IsLocalError` on the equivalent wrapped shape returns `false`. So the old behaviour was: sixty minutes wasted, then failover to an alternative peer, and the peer charged in `failedPeers`.

After this PR the same peer trips the ten-minute deadline first, and the behaviour becomes: ten minutes wasted, **no failover**, **no charge**, and the subtree fetch returns an `ErrServiceError` which — per the comment already in this file at `get_blocks.go:795-800` — the catchup top-level handler "short-circuits into a silent 'clear markers, retry' loop that hides peer-data-quality issues".

So the wall-clock damage per fetch drops from 60 min to 10 min, which is the win the PR is claiming, but the node loses its ability to route around the bad peer and loses the telemetry that would have told anyone it happened. The same peer stays in rotation and can be picked again for the next subtree, and the next.

One honest caveat: this misclassification already existed for the *streaming-timeout* path — a peer stalling mid-body would already have produced a context error under the old per-attempt timeout. What is new is that the 503-exhaustion path, the one this PR's own test exercises, has moved from the attributable bucket to the unattributable one.

### Suggested fix

Distinguish "our own deadline fired" from "the parent was cancelled", and re-classify the former as a peer error. Something along these lines at the error-return sites in `fetchAndStoreSubtreeData`, keeping the original `ctx` under a different name before it is shadowed:

```go
// Our own bound firing while the parent is still alive is the peer's fault,
// not ours — keep it attributable so failover and reputation still work.
if ctx.Err() != nil && parentCtx.Err() == nil {
    return errors.NewServiceUnavailableError(
        "[catchup:fetchAndStoreSubtreeData] Peer %s (%s) exceeded the %s subtree_data bound for %s",
        peerID, baseURL, timeout, subtreeHash.String(), err)
}
```

Any non-local error code works — `ErrServiceUnavailable` fits because it is what the retry loop would have produced anyway, and it already flows correctly through `fetchAndStoreSubtreeAndSubtreeData` and `recordCatchupPeerFailure`.

The existing test should then gain an assertion on the classification, because that is the property that actually matters here:

```go
require.False(t, errors.IsLocalError(err), "a peer that exhausts the bound must stay attributable")
```

---

## Medium — the setting cannot be overridden per settings-context

`settings/settings.go:359`:

```go
SubtreeDataFetchTimeout: getDuration("blockvalidation_subtree_data_fetch_timeout", 10*time.Minute),
```

Every `get*` helper takes a trailing `alternativeContext ...string` and passes it to `gocore.Config(alternativeContext...)`. Of the 51 `getDuration` calls in `settings.go`, 47 pass it through and 4 do not — this new one is one of the four. The others are `blockvalidation_check_subtree_from_block_timeout`, `…_retry_backoff_duration` and `p2p_ban_duration`, all pre-existing.

The consequence is narrow but real: when `Settings` is built for an alternative context — `NewSettings("docker.m")` and friends — this field silently reads the default context instead, so a `blockvalidation_subtree_data_fetch_timeout.docker.m` style override would be ignored. Given the neighbouring lines have the same bug it is easy to read as house style, but it isn't.

Fix: add `, alternativeContext...` to the call.

---

## Low — nothing in `settings.conf`, and a dead lookalike key already there

The docs table gained a row, but `settings.conf` did not. That is defensible on its own — plenty of settings live only in the struct tag plus the `settings.go` default, including the two neighbours in the same block.

What makes it worth raising is what *is* in `settings.conf`, at line 422:

```
# Timeout for fetching subtrees from peers (default: 120s)
blockvalidation_subtree_fetch_timeout = 120s
```

I grepped the whole tree for `subtree_fetch_timeout` and `SubtreeFetchTimeout` across `.go`, `.conf`, `.md` and `.yaml` — nothing reads it. It is dead config, and has been for some time. Not this PR's fault, but this PR introduces `blockvalidation_subtree_data_fetch_timeout`, which differs by one word, does something adjacent, and is *not* in the file. An operator who goes looking for "the subtree fetch timeout" will find the dead 120s knob, set it, and see nothing change.

Suggestion: add the new key to `settings.conf` next to line 422 with its 10m default and a one-line comment, and flag the dead key for removal. Removing it is arguably out of scope here, but leaving both unmentioned is the worst of the three options.

---

## Low — the bound now covers parse, serialize and store

The description flags this itself and I agree with the analysis as far as the file-backed blob store goes: `SetFromReader` writes to a temp sibling and renames, so an interrupted write leaves nothing partial for the `Exists` check at the top of the function to trip over on the next attempt.

Two things to add.

First, the shape of the tightening. Previously the *download* got `http_streaming_timeout` per attempt and the parse/serialize/store were unbounded; now the whole sequence shares a single 10m budget. For a large subtree over a slow but honest link, work that used to succeed can now fail. Ten minutes is generous and this is unlikely to bite in practice — but if it does, the High finding above makes it much worse, because the honest slow peer is *also* not failed over from and *also* not charged; the node just silently retries the same slow peer.

Second, a small side effect nobody has to act on: `countingReadCloser.onClose` reports bytes downloaded through `RecordBytesDownloaded`, and `DecoupleTracingSpan` returns the context unchanged when tracing is disabled (`util/tracing/tracing.go:636-641`), which is the normal case. Defer order in `fetchAndStoreSubtreeData` puts `Close()` before `cancelDetached()`, so the defer itself is harmless — but if the *deadline* has already expired, that best-effort byte accounting now fails with a warn log where previously the detached context guaranteed it went through. Telemetry only.

---

## Info — the 10m default is written three times

`defaultSubtreeDataFetchTimeout = 10 * time.Minute` in `get_blocks.go`, `default:"10m"` in the struct tag, and `10*time.Minute` in `settings.go`. The const carries a comment saying it "must match the default declared for that setting", but nothing enforces that, and the fail-closed guard means a drift would be silent rather than loud.

The clean version is to declare the const in the `settings` package and reference it from `settings.go`, since `blockvalidation` already imports `settings` and not the other way round. Low value; worth mentioning only because the fail-closed design makes drift invisible.

---

## Info — the description contradicts the diff

The "Not included" section says:

> The stale reference at `get_blocks.go` to a "companion fix in `services/subtreevalidation/check_block_subtrees.go`" points at a file that no longer contains a `WithoutCancel`. Left alone rather than guessing at what it should now say.

The diff removes that line. Removing it is the right call — it was stale — but the description should say so rather than claim the opposite, since a reviewer reading only the description would go looking for a comment that is no longer there.

---

## What I ran

All in a worktree at the head of this branch.

```
go build ./services/blockvalidation/... ./settings/...          -> clean
go vet   ./services/blockvalidation/  ./settings/               -> clean
gofmt -l <the four changed .go files>                           -> clean
go test -count=1 ./settings/...                                 -> ok (0.385s)
go test -count=1 -race -tags testtxmetacache \
  -run TestFetchAndStoreSubtreeData_DetachedFetchIsBounded \
  ./services/blockvalidation/                                   -> ok (2.764s)
golangci-lint run ./services/blockvalidation/... ./settings/... -> 9 issues, all pre-existing
```

The nine lint issues are all `prealloc` suggestions on lines this PR does not touch (`get_blocks_test.go` at 1678/1768/2001/3307/3349, `testhelpers/block_helpers.go:268`, and three more of the same shape). Nothing the PR introduces trips the linter.

I did **not** run the full repository suite or the smoke/sequential tests — too heavy for the box I was on, and the description already reports the full `blockvalidation` package green at 275s.

## On the new test

It is a good test and it passes under `-race`. Asserting on attempt count rather than wall-clock alone is the right call, and `NewCatchupTestSuite` builds fresh settings per test via `testutil.CreateBaseTestSettings(t)`, so mutating `suite.Server.settings.BlockValidation.SubtreeDataFetchTimeout` cannot leak into other tests. The 600 ms deadline against a 250 ms → 500 ms → 1 s backoff chain gives plenty of margin in the safe direction: a loaded CI box makes *fewer* attempts happen, not more, so `calls < 6` cannot flake upward.

Its one gap is the one described in the High finding — it pins the duration but not the attribution, which is the property that determines whether the node actually recovers from the peer it just gave up on.

liam:
## Addressed in 28e1839

The High finding is right and I've taken it. It is a regression I introduced, and the framing in the review is the correct one: the wall-clock win was real but the node lost the ability to route around the peer responsible for it, which is the opposite of the point.

### The suggested fix does not work, and neither does the obvious variant

Worth recording, because the failure is silent. The sketch in the review keeps the context error as a wrapped cause:

```go
return errors.NewServiceUnavailableError("... exceeded the %s subtree_data bound ...", ..., err)
```

`IsContextError` matches a wrapped cause, so that still classifies as local. So does rendering the cause into the message with `%v` or `.Error()`. I probed all four shapes against the real helpers before choosing:

| Shape | `IsLocalError` |
|---|---|
| current (context error propagates) | `true` |
| reclassified, context error wrapped as cause | `true` |
| reclassified, cause rendered with `%v` | `true` |
| reclassified, no reference to the context error | **`false`** |
| pre-PR 503 exhaustion, for comparison | `false` |

Only the last form works. The code now carries a comment saying so, because the natural instinct on reading it is to add the cause back for diagnostics, and doing that would silently restore exactly the classification this is undoing.

### What landed

The re-attribution is a deferred check on a named return, keyed on the **parent** context rather than the detached one, so it does not depend on defer ordering relative to `cancelDetached`. It only fires while the caller's context is still alive: a genuinely cancelled parent is our shutdown, not the peer's fault, and stays local.

One residual I should name rather than let you find: if the parent is cancelled in the window between our deadline firing and the defer running, the error stays local. The window is tiny and it errs toward the previous behaviour, so I left it rather than adding synchronisation.

Also taken:

- **Medium**, `alternativeContext...` now passed. You were right that the neighbouring lines having the same bug makes it read as house style; 47 of 51 pass it.
- **Info**, the 10m default now lives once, in `settings.DefaultSubtreeDataFetchTimeout`. Your point that the fail-closed guard makes drift *silent* rather than loud is what made this worth doing. The struct tag still repeats it as a literal because tags cannot reference constants; that is noted where the constant is declared.
- **Low**, `settings.conf` gains the new key. I also marked the dead `blockvalidation_subtree_fetch_timeout` above it as dead in a comment rather than deleting it, so an operator who has already set it is not surprised, but nobody new sets it by mistake. Deleting it is a separate change.
- **Info**, the description contradiction is fixed. You were right, the diff does remove the stale `check_block_subtrees.go` reference and the description claimed the opposite.

### On the test nits

Both reviews landed on the same gap from opposite directions, and the fix for one changes the other. The test now asserts `errors.IsLocalError(err) == false` plus `ErrorIs(err, errors.ErrServiceUnavailable)`, which pins the attribution.

That also resolves the suggestion to replace `require.Less(t, calls, 6)` with `require.ErrorIs(t, err, context.DeadlineExceeded)`: after this change the error is deliberately **not** a context error, so that assertion would now be wrong. I kept the attempt-count assertion, since it is what discriminates against a loop that runs to completion quickly, and added the classification assertions alongside it rather than instead of them.

I also added `TestSubtreeDataFetchTimeout_FailsClosed`. The fail-closed resolution moved into a small nil-tolerant helper, partly because it was the uncovered branch behind the Sonar new-code coverage gate, and partly because "unset must not mean unbounded" deserves to be pinned rather than inferred.

### On the Low finding about parse and store now being bounded

Agreed, and thank you for the backend-by-backend check, which goes further than mine did. The point that an honest slow peer is made *worse* by the High finding is well taken, and is another reason attribution had to be fixed rather than deferred.

### Verification at 28e1839

```
go build ./services/blockvalidation/... ./settings/...            -> clean
go vet ./services/blockvalidation/ ./settings/                    -> clean
gofmt -l <changed files>                                          -> clean
gci diff --skip-generated -s standard -s default <changed files>  -> clean

go test -count=1 ./settings/...
  ok  github.com/bsv-blockchain/teranode/settings                          0.107s
go test -count=1 -timeout 1800s -tags testtxmetacache ./services/blockvalidation/...
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation          275.276s
  ok  github.com/bsv-blockchain/teranode/services/blockvalidation/catchup    8.403s
```

Mutation: disabling the re-attribution fails the new assertion and only that one (`Should be false: a peer that exhausts the bound must stay attributable`). Removing the deadline still fails the attempt-count assertion as before.

`-race` and the full repository suite were not run locally.

### Not taken

The P2 follow-up about mapping the expiry to an `ErrExternal` more generally: that is now partly moot for this call site, since the expiry no longer surfaces as a context error at all. The broader question, whether a `Done()` on any detached context should be classified as peer-controlled I/O, is still worth its own issue and is wider than this PR.

freemans13 (approved):
## Re-review after `28e183976` — approving

All six findings from the previous pass are addressed, and the High one is addressed properly rather than papered over. I re-verified each fix against the code and by running it, including a mutation test to confirm the new assertion actually pins the behaviour. Nothing in the follow-up commit is wrong.

Two residual points below, both minor and neither blocking, plus one genuinely separate pre-existing vulnerability that this review turned up and that should become its own issue.

| Previous finding | Status | Verified how |
|---|---|---|
| **High** — deadline expiry classified as local, killing failover and peer charging | **Fixed** | Ran the path; mutation test confirms the new assertion catches a regression |
| **Medium** — `getDuration` missing `alternativeContext...` | **Fixed** | `NewSettings("docker.m")` resolves the key |
| **Low** — key absent from `settings.conf`, dead lookalike nearby | **Fixed** | Key added and parses; dead key now labelled DEAD KEY |
| **Low** — bound covers parse + store | **Acknowledged** | Unchanged by design; see residual point 2 |
| **Info** — default written three times | **Fixed** | Now `settings.DefaultSubtreeDataFetchTimeout`, one declaration |
| **Info** — description contradicted the diff | **Fixed** | Commit message states the removal plainly |

## A correction I owe you

My suggested fix wrapped the context error as a cause:

```go
return errors.NewServiceUnavailableError("... exceeded the %s bound ...", ..., err)
```

You are right that this does not work, and it is worth recording *why*, because it is not visible from reading `IsContextError`. `errors.Is` falls through to `checkGRPCContextError`, which calls `hasGRPCCode`, which runs `status.FromError(current)` over each link in the chain (`errors/errors.go:1207-1231`). `status.FromError` parses the error's **message string**. So an error whose text merely *contains* "context deadline exceeded", with no wrapped cause at all, is classified as a context error.

I checked all four shapes:

```
A: wrapped as cause          IsContextError=true   IsLocalError=true
B: rendered with %v          IsContextError=true   IsLocalError=true
C: message mentions phrase   IsContextError=true   IsLocalError=true
D: no reference at all       IsContextError=false  IsLocalError=false
```

Only D re-classifies. The comment warning that the replacement "must not carry the context error, neither wrapped as a cause nor rendered into the message" is load-bearing, and the test's `require.False(t, errors.IsLocalError(err))` pins it, so a future diagnostic tweak that reintroduces the text fails loudly rather than silently reverting the fix. That is the right structure.

## The `parentCtx` guard is sound

The guard reads the caller's context rather than the detached one, which is right, but it is worth confirming that `parentCtx` cannot expire on its own — if it could, the re-attribution would silently stop firing on long fetches, which is exactly when it matters.

`parentCtx` is the context returned by `tracing.Tracer(...).Start(...)`. `UTracer.Start` only installs a timeout when the caller passes a `WithTimeout` option (`util/tracing/tracing.go:444-447`); this call site passes only `WithParentStat` and `WithDebugLogMessage`, so `options.Timeout` is zero and no deadline is added. `parentCtx` therefore expires only when the real caller cancels.

Defer ordering is fine too, and as your commit message says the guard does not depend on it. For the record the order is: buffered-reader return → `subtreeDataReader.Close()` → re-attribution → `cancelDetached()` → tracing `deferFn()`. `deferFn()` is called with no arguments so it cannot disturb the named return. The named return does not shadow anywhere — `subtreeDataReader, err := ...` reuses the function-scope `err` — and the `Exists` check still runs on the original context and returns before the re-attribution defer is registered, so a pre-cancelled parent still exits early and stays local.

## Residual 1 — the original error is discarded entirely (Info)

The replacement carries nothing of the original, which is necessary for the reason above. The side effect is that when the bound fires you can no longer tell whether it expired in the HTTP fetch, in the parse, or in the store write — three quite different operational meanings — because the error is dropped rather than logged.

A log line is not an error chain, so this costs nothing:

```go
defer func() {
    if err != nil && parentCtx.Err() == nil && errors.IsContextError(err) {
        u.logger.Warnf("[catchup:fetchAndStoreSubtreeData] bound expired for subtree %s from peer %s (%s): %v",
            subtreeHash.String(), peerID, baseURL, err)
        err = errors.NewServiceUnavailableError(...)
    }
}()
```

## Residual 2 — the re-attribution also covers our own store write (Low)

`errors.NewStorageError("failed to store subtreeData", ctxErr)` is a context error — `IsContextError` is true for it. So if the bound expires during `subtreeStore.Set`, the peer is charged and the node fans out to alternative peers for what may be our own slow disk.

This is genuinely ambiguous rather than plainly wrong: by the time the store write starts the peer has already had most of the ten-minute budget, so "the peer ate the time and left none for the write" is at least as likely as "our disk is slow". And the previous behaviour — nobody charged, no failover — was worse. So I would not change the logic, only say this in the comment. If you would rather fail toward not blaming peers, excluding storage errors is one line:

```go
if err != nil && parentCtx.Err() == nil && errors.IsContextError(err) &&
    !errors.Is(err, errors.ErrStorageError) {
```

## Separate issue — a peer can force the local classification with its response body

Pre-existing, not introduced here, and not a reason to hold up this PR. But it is the same bug class this PR is fixing, and this review is how it surfaced.

`hasGRPCCode` parses the error **message**, and `buildHTTPError` embeds up to 2048 bytes of peer-controlled response body into that message (`util/http.go:486`). So a peer that answers a non-2xx with the body `context canceled` produces an error that `IsLocalError` calls local — suppressing alternative-peer failover in `fetchAndStoreSubtreeAndSubtreeData` and stopping `recordCatchupPeerFailure` charging it. Verified against the exact shape `buildHTTPError` produces:

```
peer body "context canceled"          -> IsContextError=true   IsLocalError=true
peer body "context deadline exceeded" -> IsContextError=true   IsLocalError=true
peer body "boring 500"                -> IsContextError=false  IsLocalError=false
```

A peer can make itself un-failover-able and un-chargeable with a fourteen-byte response body — the same outcome the High finding described, reached deliberately instead of accidentally.

Your new defer incidentally neutralises this for the subtree_data path, since the re-attribution converts it back to `ErrServiceUnavailable` whenever the parent is alive. The `/subtree` path is still exposed: `fetchSubtreeFromPeer` uses `util.DoHTTPRequestBounded` and wraps the result at `get_blocks.go:870`, with no equivalent guard. Anywhere else peer-controlled text reaches an error message has the same exposure. The durable fix belongs in `errors` — `hasGRPCCode` should not be inferring transport semantics from arbitrary message text. Worth its own issue.

## What I ran

At `28e183976`, in a worktree left clean afterwards.

```
go build ./...                                                  -> clean
go vet   ./services/blockvalidation/  ./settings/               -> clean
gofmt -l <the four changed .go files>                           -> clean
go test -count=1 ./settings/...                                 -> ok (0.402s)
go test -count=1 -race -tags testtxmetacache \
  -run 'TestFetchAndStoreSubtreeData_DetachedFetchIsBounded|
        TestSubtreeDataFetchTimeout_FailsClosed' \
  ./services/blockvalidation/                                   -> ok (3.700s), both PASS
golangci-lint run ./services/blockvalidation/... ./settings/... -> 9 issues, all pre-existing prealloc
```

`settings.conf` parsing deserved its own check, because `getDuration` **panics** on a parse failure — a bad value in the committed config is a startup crash, not a fallback. `10m` is fine:

```
NewSettings()           -> SubtreeDataFetchTimeout = 10m0s
NewSettings("docker.m") -> SubtreeDataFetchTimeout = 10m0s
```

The second line also confirms the `alternativeContext` fix.

**Mutation test.** I disabled the re-attribution condition and re-ran the bounded test. It fails on the classification assertion alone, at the intended line:

```
--- FAIL: TestFetchAndStoreSubtreeData_DetachedFetchIsBounded (0.70s)
    Error Trace: services/blockvalidation/get_blocks_test.go:1021
    Error:       Should be false
    Messages:    a peer that exhausts the bound must stay attributable, not read as a local failure
```

So the assertion is real, not decorative.

Not run: the full repository suite, smoke and sequential tests. The full `blockvalidation` package was reported green at 275s on the previous commit and nothing in this follow-up changes a shared path, but that is an inference, not a result I have.

## On the new `TestSubtreeDataFetchTimeout_FailsClosed`

Extracting the resolution into `subtreeDataFetchTimeout` and testing it directly is a better shape than the inline fallback it replaces — the branch was previously unreachable from any test. The nil-settings case is a little defensive for production code, but the justification given is honest about why it is there, and a nil-tolerant fail-closed helper is hard to argue with when the thing it guards is peer-controlled.

PR #1531, opened 2026-08-07
Escape peer-controlled gossip strings in the dashboard raw JSON view

gokutheengineer:
**Closes bitcoin-sv/teranode#4735.**

### Problem

Both items in the issue were confirmed still valid on `main`:

1. `ui/dashboard/src/internal/components/msgbox/index.svelte` rendered websocket messages through `{@html formatJSON(message)}`, and `formatJSON` applied syntax-highlighting regexes to `JSON.stringify` output without any HTML escaping. `JSON.stringify` escapes JSON metacharacters, not HTML.
2. `handleNodeStatusTopic` (`services/p2p/Server.go`) and `handleBlockTopic` (`services/p2p/server_helpers.go`) forwarded peer-controlled strings to `notificationCh` verbatim, before the health gate, with no length or charset bound.

Net effect: an anonymous remote peer could execute script in the operator dashboard's origin by publishing one gossip message, gated only on the operator toggling the raw-JSON switch.

The fresh-eyes review found three further relay paths the issue did not mention, all fixed here:

- `best_block_hash` reached WebSocket clients raw whenever `best_height == 0`, since the parsed-hash replacement is gated on `BestHeight > 0`. Bounded only by the 64 KB message cap.
- `handleSubtreeTopic` (`server_helpers.go`) relays `ClientName` with exactly the pattern fixed in `handleBlockTopic`.
- Our **own** `miner_name` is extracted from the best block's coinbase scriptSig (`util.ExtractCoinbaseMinerRaw`), so it is chosen by whoever mined the block, and it was published to peers and forwarded to WebSocket clients unfiltered. Every node on the network renders the same block, so this path needs no gossip at all.

### Fix

**Dashboard (the actual vulnerability).** New `msgbox/json-highlight.ts` owns escaping and highlighting; the component imports it.

Rather than the escape-then-regex approach the issue proposes, this walks the JSON once with a quote-aware token regex and escapes each fragment as it is emitted. Two reasons:

- Escaping `"` (as the issue suggests) would silently kill highlighting, because all five highlighting regexes match on `"`.
- Escaping only `&<>` and keeping the regexes is *safe* but not *clean*: the review showed the regexes are quote-unaware, so `client_name = ":evil"` splices a `<span>` into another span's `class` attribute, producing malformed (though inert) markup. The single-pass version places spans structurally, so "the only tags in the output are the ones we wrote" is actually true rather than nearly true.

`escapeHtml` escapes `&`, `<`, `>` and deliberately not `"`: output is only ever interpolated into element content, never an attribute.

**Service (defence in depth).** New `services/p2p/sanitize.go`, applied at ingest so a single call covers both the WebSocket relay and the peer-registry writes:

- `Version`, `CommitHash`, `ClientName`, `MinerName`, `FSMState`, `SyncPeerID` - printable ASCII minus `<>&"'`+backtick+backslash, capped at 128. Filtered rather than rejected, since `node_status` is telemetry and dropping a whole message over an odd client name loses monitoring data.
- `ChainWork`, `SyncPeerBlockHash`, and `BestBlockHash` - hex-only, all-or-nothing (a partially stripped hex number would silently mean something else).
- `ListenMode`, `Storage` - exact enum match, unknown becomes `""`, which is the "unknown / old peer" encoding consumers already handle.

The cap is 128 rather than 64 to match the peer registry's existing `sanitizeClientName`, so the same value is not bounded to two different lengths depending on path; 64 would also have truncated the 69-character genesis coinbase tag.

`BestBlockHash` is bounded in `handleNodeStatusTopic` at the point it seeds the notification, not inside `sanitizeNodeStatusMessage`. Doing it in the shared helper regressed `TestHandleNodeStatusTopic_RejectsMalformedAdvertisedHash`: blanking a malformed hash early made `sanitizeAdvertisedTip` accept the tip (capped) instead of rejecting it outright, so validation must still see what the peer actually sent.

Other **`{@html}`** sinks audited and left alone, with reasons: `internal/components/typo/index.svelte:57` and `lib/components/typo/index.svelte:50` gate on an `html` prop that defaults false and has no app-level caller; `block-details-card` `{@html difficultyDisplay}` is numeric-derived via `formatNumberExpStr`; the logo and icon sinks render static SVG maps. Note the scope of that sweep: it searched for `{@html}`, so it does not cover DOM-library HTML sinks such as d3's `.html()` - see the first Residual item.

**CI.** `.github/workflows/dashboard_pr_smoke.yaml` ran only `npm run check` and `npm run test:integration` (Playwright). `test:unit` (vitest) is bundled only under `npm run test`, which the workflow never calls, so the escaping tests added here would never have executed on a PR - the regression guard this fix relies on was unenforced. The workflow now runs `npm run test:unit -- --run` between the two existing steps (`--run` so it cannot hang in watch mode).

### Residual

- **A second, currently-exploitable HTML sink on the forks page, not fixed here.** `ui/dashboard/src/routes/forks/helpers.ts:242-277` builds d3 node labels with `.html(...)`, interpolating `d.data.hash` raw and `d.data.miner` raw: `domain` is seeded from `d.data.miner` and only replaced on a successful `new URL()` parse, so a coinbase tag like `` (not a valid URL) falls through to the label verbatim. Same threat model as the `MinerName` path sanitized here - the string comes from the coinbase scriptSig - but it needs no raw-JSON toggle, only an operator viewing `/forks`. Pre-existing and outside this PR's gossip-string scope, so it is left for a follow-up; the fix is to escape the untrusted fragments or set them with `.text()` rather than `.html()`.
- `lib/components/textinput/index.svelte:25` defaults `footnoteHtml = true`, so a future `<TextInput footnote={apiError}>` would be an instant XSS. No caller passes `footnote` today, so changing the default is out of scope here, but it is a live footgun.
- Three handlers log raw peer strings before validation (`server_helpers.go:76`, `:212`, `:484`), which sits awkwardly with this change's own log-injection rationale. Pre-existing and log-only, not a dashboard sink.
- The `Storage` enum allowlist is a forward-compatibility trade: a future mode such as `"archive"` would be dropped by older nodes. Chosen deliberately, since every consumer only tests `== "full"`. Relatedly, an unrecognised value now leaves the previous registry value in place instead of overwriting it.
- Per-field bounds are applied at the p2p ingest points only; the broader per-field length work in the companion issue is not attempted here.

### Tests

Added `ui/dashboard/src/internal/components/msgbox/json-highlight.test.ts` (11 tests). The core invariant asserts every tag in the output is one of our own highlighting spans, exercised with the issue's `` payload, a `</span><svg/onload=...>` breakout, hostile property *names*, nesting, and the five splice payloads the review found (`:evil`, `x": 1, ":z`, tokens inside strings, lone quote, trailing backslash). One test pins that highlighting still works, so escaping `"` cannot be reintroduced silently.

Added `services/p2p/sanitize_test.go`. Unit tests per helper, plus end-to-end handler tests asserting `notificationCh` receives sanitized values for `handleNodeStatusTopic`, `handleBlockTopic`, `handleSubtreeTopic`, and the `best_height == 0` tip path. `TestSanitizeNodeStatusMessage_PreservesHonestValues` covers every `listen_mode`/`storage` combination and the genesis miner tag. `TestSanitizeNodeStatusMessage_CoversEveryStringField` walks the struct by reflection and fails when a new string field is added without either sanitizing it or adding a documented exemption - this is the guard that the missed `BestBlockHash` would have tripped.

Dropped two assertions the review proved vacuous: the peer registry runs its own `sanitizeClientName` on write, so asserting the stored name is clean passes against the unfixed code.

Coverage on the new Go code: `sanitize.go` is at 100% across all four functions, and every changed line in `Server.go` / `server_helpers.go` has a hit. The last uncovered one was the `minerName` assignment in `getNodeStatusMessage` (nothing reached it, because the shared test harness leaves `blockchainClient` nil), now covered by `TestGetNodeStatusMessage_SanitizesMinerName`.

The bidi test case is written as the escape `"teranode" + "\u202e" + "en"` rather than a literal U+202E, so the file stays pure ASCII - a bidi control in source renders differently from how it executes (Sonar S6389).

```
  ok  github.com/bsv-blockchain/teranode/services/p2p  17.000s

go vet -tags "testtxmetacache" ./services/p2p/...    # clean

cd ui/dashboard && npm run test:unit -- --run        # 40 passed (2 files)
cd ui/dashboard && npm run check                     # 800 files, 0 errors, 14 pre-existing warnings
```


freemans13 (approved):
Approving. The fix is sound, the core vulnerability is genuinely closed, and the single-pass highlighter is a better answer than the escape-then-regex approach the original issue proposed. Everything the description claims reproduced when I ran it myself.

What follows is one item I'd like to see land (in this PR or a fast follow-up — it isn't a defect in the code here), plus some consistency notes and observations. None of it blocks merge.

## Verified independently

| Check | Result |
|---|---|
| `go test -tags testtxmetacache ./services/p2p/` (full package) | `ok` 11.0s |
| `go test -race` on the sanitize/handler tests | `ok` 2.0s |
| `go vet -tags testtxmetacache ./services/p2p/` | clean |
| `golangci-lint run services/p2p/...` | clean |
| `vitest run` (dashboard) | 40 passed, 2 files |
| `npm run check` (svelte-check) | 800 files, 0 errors, 14 warnings |

Environment note for anyone else reproducing the frontend half: the dashboard requires Node >= 20.19.0, so `npm ci` fails outright on Node 18. I used 22.16.0.

I read the full files rather than the diff, and traced each peer-controlled string from the gossip handler through to the browser.

## The one worth acting on: CI never runs the dashboard unit tests

The heart of this PR is the invariant in `json-highlight.test.ts` — every tag in the highlighter's output is one of its own five spans, exercised with the issue's original payload and a `</span><svg/onload=alert(1)>` breakout. It's a good test. It doesn't run in CI.

`.github/workflows/dashboard_pr_smoke.yaml` is the only workflow triggering on `ui/dashboard/**`, and its steps are:

```yaml
- run: npm run check              # svelte-check
- run: npm run test:integration   # Playwright
```

No `npm run test:unit`. I grepped every workflow in `.github/workflows/` for `test:unit` and `vitest` — no hits anywhere. So this suite passes locally and is then never executed again by anything.

The gap is pre-existing (`theme_parity.test.ts` has been in the same position), but this PR is the first to lean on vitest for a *security* invariant, which changes what the gap costs: a future refactor back to escape-then-regex would go green.

One line after the existing `npm run check`:

```yaml
- run: npm run test:unit -- --run
```

`--run` because `test:unit` is bare `vitest` (watch mode); vitest does auto-detect CI, but explicit is cheaper than relying on it. `npm test` won't do — that's `test:integration && test:unit`, which would run Playwright twice.

## Consistency notes

### `base_url` is the sibling field that didn't get the same treatment

All three handlers put a peer-supplied URL on the notification that reaches WebSocket clients — `BaseURL: blockMessage.DataHubURL` at `server_helpers.go:99`, the same at `:228`, and `nodeStatusMessage.BaseURL` at `Server.go:1137`. Right beside it on the same struct literal, `ClientName` is now filtered. The URL isn't.

`validateDataHubURL` (`server_helpers.go:389`) checks the scheme is http/https and runs the SSRF/localhost/blacklist checks against the hostname. It says nothing about path or query. Confirmed with a standalone program that `url.Parse` accepts both of these without complaint:

```
http://evil.example/<img src=x onerror=alert(1)>   -> err=<nil> scheme="http" host="evil.example"
http://evil.example/?a="><svg onload=alert(1)>     -> err=<nil> scheme="http" host="evil.example"
```

So markup reaches `notificationMsg.BaseURL`, bounded only by the message-size cap. Inert today because the dashboard now escapes — but that is the same "bounded only by downstream escaping" position this PR argues is not good enough for `ClientName`.

I'd lean towards documenting the exemption rather than filtering, since URLs legitimately need characters the display filter strips. Either way it's worth being a decision rather than an omission.

### The "same cap on both paths" rationale is only half true

The description justifies 128 over 64 as matching `sanitizeClientName` "so the same value is not bounded to two different lengths depending on path". The number does match (`maxPeerNameLength = 128`, `services/blockchain/peer_registry.go:22`) but the semantics differ:

- `sanitizeClientName` truncates to 128 bytes **first** (`name = name[:maxPeerNameLength]`), then filters.
- `sanitizePeerDisplayString` filters **while** counting, so stripped characters never consume budget — which `TestSanitizePeerDisplayString_Truncates` deliberately pins.

`strings.Repeat("<", 200) + "teranode"` therefore becomes `""` through the registry and `"teranode"` through the new helper. Same cap, different answer. Nothing breaks; the comment's claim just doesn't quite hold. Cleanest fix would be for the registry to call the new helper — failing that, soften the comment to say the *cap* matches.

Worth noting the new filter is strictly stricter than the registry's, because it excludes the backtick and the registry doesn't. An improvement, just not "matching".

### The reflection guard covers the wrong struct by one hop

`TestSanitizeNodeStatusMessage_CoversEveryStringField` is the smartest thing here — exactly the guard that would have caught the missed `BestBlockHash`. But the struct that reaches the browser is `notificationMsg` (`HandleWebsocket.go:24`), which has string fields the guard can't see: `Hash`, `BaseURL`, `PropagationURL`, `Miner`, `PreviousHash`.

I grepped for assignments to the last two outside tests and found none — they're dead in the p2p package, so no live hole. But `Miner` is dead by accident, and a future handler populating it from block metadata would be the exact `MinerName` bug this PR just fixed, unguarded. Consider mirroring the reflection test onto `notificationMsg`, or a comment on that struct pointing at the sanitizer.

### `sanitizePeerHexString` bounds width without pinning it

Both callers pass values that are exactly 64 hex characters in practice, but the helper accepts 1 to 64 including odd lengths. So on the `best_height == 0` path a peer can still push `best_block_hash: "DEADBEEF"` — inert, but meaningless. Requiring exactly 64 (or at least an even length) would be tighter. Judgement call, not a defect, given the framing is bounding rather than validating.

### One HTML sink the `{@html}` sweep couldn't see

The audit section is accurate for `{@html}`, but that isn't the whole sink class. `ui/dashboard/src/routes/forks/helpers.ts:242` builds node labels through d3's `.html()`, interpolating `d.data.hash` raw and `d.data.miner` raw whenever `new URL(d.data.miner)` throws (the `catch` logs and falls through with `domain` still holding the raw value).

Out of scope — forks API data, not gossip, and it predates this PR. Raising it only because the audit reads as exhaustive and `innerHTML` / `.html()` are the same hazard. A line in the residual list would stop the next reader assuming otherwise.

### Storage-mode constants duplicated with no compile-time link

`storageModeFull` / `storageModePruned` restate the bare literals `util.DetermineStorageMode` returns. The forward-compatibility trade is already flagged in the description; the mechanical hardening is to export those constants from `util` and reference them, so adding a mode becomes a compile-visible change here rather than a silent drop. Minor.

## Checked and fine

Recording these so the ground isn't re-covered:

- **No ReDoS in the highlighter.** In `("(?:\\.|[^"\\])*")` the two inner branches are disjoint (one starts with a backslash, the other excludes it), so there's nothing ambiguous to backtrack through.
- **The module-level `g`-flagged regex is safe.** `matchAll` clones the regex, so there's no shared `lastIndex` between calls. Would have been a real bug with `.exec` in a loop.
- **`(\s*:)?` can't mis-fire.** In `JSON.stringify` output a value is never followed by a colon, so the optional group only attaches to genuine property names.
- **Truncation arithmetic is right.** `b.Len()` counts bytes while the loop ranges over runes, but every surviving rune is single-byte ASCII so the two coincide. Invalid UTF-8 decodes to `RuneError` (0xFFFD), which is `> 0x7E` and gets stripped.
- **Our own `chain_work` still survives the new hex allowlist.** `chainWorkStr` is `hex.EncodeToString(bestBlockMeta.ChainWork)` (`Server.go:1330`) — 32 bytes, so exactly 64 lowercase hex characters. Worth stating explicitly: a `0x` prefix or decimal encoding would have made this a silent network-wide regression where every node dropped every peer's chainwork.
- **The 64-vs-128 reasoning checks out** — the genesis coinbase tag is 69 characters, so 64 really would have truncated it.
- **Handler ordering is correct.** `ClientName` is sanitized after the peer-ID spoof check (which needs the raw claimed ID) and before the ban gate, the registry write and the notification send, in both `handleBlockTopic` and `handleSubtreeTopic`.
- **The `BestBlockHash` carve-out is genuinely necessary.** Blanking it in the shared helper would let `sanitizeAdvertisedTip` accept a capped tip instead of rejecting a malformed one. Bounding it at the notification seed is the right place, and the comment explaining why is worth keeping.
- **`FeePolicy` carries no strings** — all four fields numeric, so the relayed pointer holds no unsanitized text.
- **The five highlight CSS classes are all still defined** (`index.svelte:204-221`), so the rewrite doesn't silently lose styling.
- **The non-raw msgbox path is safe** — `{field.value}` at `index.svelte:91` is ordinary Svelte interpolation, which escapes.

## Suggested order

1. Add `npm run test:unit -- --run` to `dashboard_pr_smoke.yaml` — without it this PR's regression guard is decoration.
2. Decide on `base_url`: sanitize, or document the exemption.
3. Soften or fix the "same cap both paths" comment.
4. The rest is optional polish.

ctnguyen (changes requested):
Review at commit [eb8dadb](https://github.com/gokutheengineer/teranode/commit/eb8dadbd19f0bd89bed810d597a149127c95ca31)

Solid work — the core fix is right where it matters. The single-pass tokenizing highlighter places spans structurally and escapes each fragment as it emits, so the "only our own tags" invariant genuinely holds, leaving `"` unescaped is correct for `<pre>` content, and the tests pin that decision. The service-side sanitizers and the reflection guard are sensible defence-in-depth, and the write-up is honest about its own residuals. One thing must be sorted before merge — the suite that guards all of this never runs in CI — plus a handful of non-blocking items, most of which the PR's own framing invites.

## Blocking issues

### ChiR1 — CI never runs the vitest suite that guards this fix

*Same as [@freemans13's point "CI never runs the dashboard unit tests" (10-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1531#pullrequestreview-4898386641).*

**Problem:** Confirmed against the workflow: `.github/workflows/dashboard_pr_smoke.yaml` runs only `npm run check` and `npm run test:integration`. In `ui/dashboard/package.json`, `test:integration` is `playwright test` and the new vitest suite is a separate `test:unit` (`vitest`) — bundled only under `npm run test`, which CI never calls. So `json-highlight.test.ts`, the entire security invariant of this PR, never executes on any PR.

**Why it matters:** This PR sells its safety on those tests ("escaping `"` cannot be reintroduced silently"). It can — CI stays green through a regression that reintroduces the `{@html}` XSS. The regression guard is the mechanism this fix relies on to stay fixed, and right now it is unenforced; wiring it in is must-fix-before-merge, not a follow-up.

**Fix:** Add a vitest step to the smoke workflow, in run mode so it doesn't hang waiting for watch:

```yaml
      - run: npm run check
      - run: npm run test:unit -- --run
      - run: npm run test:integration
```

## Non-blocking issues

### ChiR2 — A second unescaped `{@html}` sink survives on the forks page

*Same as [@freemans13's point "One HTML sink the {@html} sweep couldn't see" (10-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1531#pullrequestreview-4898386641).*

**Problem:** `ui/dashboard/src/routes/forks/helpers.ts:242` builds d3 node labels with `.html(...)`, interpolating `d.data.hash` raw and `d.data.miner` raw whenever the `new URL(d.data.miner)` parse throws (line 244-252 falls back to `domain = d.data.miner`). The miner string comes from the coinbase scriptSig — the exact "chosen by whoever mined the block" source this PR sanitizes for `MinerName` on the p2p side. A coinbase tag like `` is not a valid URL, so it flows straight into `.html()`. Extending @freemans13's point: this is not merely a residual to note — it is a live, currently-exploitable XSS in the operator dashboard, and unlike the issue this PR fixes it needs no raw-JSON toggle, just an operator viewing `/forks`.

**Why it matters:** The PR's "Audited and left alone, with reasons" section reads as an exhaustive `{@html}` sweep, but this sink isn't in it. A reader trusts the audit and moves on, leaving a second XSS with the same threat model as the one just closed. It is pre-existing and outside the gossip-string scope, which is why it doesn't block — but it should not be left silently undocumented.

**Fix:** Out of scope to fix here, but add it to the Residual list, and ideally open a follow-up. The fix is to stop feeding raw strings to `.html()` — build the label with escaped text (`escapeHtml(d.data.miner)` / `escapeHtml(d.data.hash)`) or set text via `.text()` for the untrusted fragments.

### ChiR3 — `BaseURL`/`DataHubURL` still reach the dashboard as raw peer text, and the exemption comment overstates its own safety

*Same as [@freemans13's point "base_url is the sibling field that didn't get the same treatment" (10-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1531#pullrequestreview-4898386641).*

**Problem:** `validateDataHubURL` (`server_helpers.go:389`) only checks scheme and hostname — nothing in path/query/fragment — so `http://evil.example/">` passes. That value reaches `notificationMsg.BaseURL` verbatim in **all three** handlers, not just node_status: `handleNodeStatusTopic` (`Server.go:1137`), `handleBlockTopic` (`server_helpers.go:99`, `DataHubURL → BaseURL`), and `handleSubtreeTopic` (`server_helpers.go:228`). Extending @freemans13's node_status observation to the block/subtree paths. It is inert today only because the dashboard now escapes — the very cross-layer coupling this PR's opening rationale argues against ("independently of how carefully any one consumer escapes"). The bigger snag is the exemption rationale in `sanitize.go:111` and `sanitize_test.go`: `BaseURL` is documented as *"validated separately (SSRF + operator blacklist)"* / *"has its own SSRF and blacklist validation"* — which reads as "this field is handled," but SSRF/blacklist validation says nothing about markup.

**Why it matters:** A future maintainer trusts that comment and assumes `BaseURL` is display-safe. It isn't; it is display-safe only as long as the dashboard escape holds. That is precisely the invariant this PR is trying to make explicit rather than incidental.

**Fix:** Either bound the URL's renderable surface at ingest, or — at minimum — correct the exemption comment to state the truth: `BaseURL` carries unbounded peer text into the dashboard and is safe *only* because the raw-JSON view escapes, not because SSRF validation makes it inert.

### ChiR4 — The reflection guard pins the ingest struct, not the struct that reaches the browser

*Same as [@freemans13's point "The reflection guard covers the wrong struct by one hop" (10-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1531#pullrequestreview-4898386641).*

**Problem:** `TestSanitizeNodeStatusMessage_CoversEveryStringField` walks `NodeStatusMessage`, but the struct actually serialized to `/p2p-ws` clients is `notificationMsg` (`HandleWebsocket.go:24`). Its string fields `Hash`, `BaseURL`, `PropagationURL`, `PreviousHash`, `Miner` are populated in the handlers and are not covered by any equivalent guard. Today only `BaseURL` is a live gap (ChiR3) — `Hash` is always a re-parsed canonical hash, `Miner`/`PreviousHash`/`PropagationURL` are unset on the reaching paths — so this is dead-by-accident, exactly as @freemans13 noted, not a bug per se.

**Why it matters:** The guard's promise is "a newly added string field cannot reach the browser unsanitized." It doesn't deliver that: a new field added to `notificationMsg`, or a handler that starts populating `Miner`/`PreviousHash` from peer data, sails past the reflection test because the test never looks at that struct.

**Fix:** Mirror the reflection walk onto `notificationMsg` (with its own exemption map for the parsed/derived fields), or at least add a comment on `notificationMsg` noting which fields must already be sanitized before assignment and pointing at the ingest sanitizers.

### ChiR5 — "Same cap on both paths" holds for the number, not for the behaviour

*Same as [@freemans13's point "The 'same cap on both paths' rationale is only half true" (10-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1531#pullrequestreview-4898386641).*

**Problem:** The `maxPeerDisplayStringLen = 128` comment says the value "is not bounded to two different lengths depending on which path it took," matching `blockchain.sanitizeClientName`. The numeric cap matches, but the two functions don't behave alike. `sanitizeClientName` (`peer_registry.go:34`) truncates to 128 bytes *first*, then filters; `sanitizePeerDisplayString` filters *while* counting, so stripped chars don't consume budget. `strings.Repeat("<",200)+"teranode"` yields `""` via the registry path but `"teranode"` via the new helper. Extending @freemans13: they also diverge on charset — `sanitizeClientName` keeps a backtick (its `<>&'"\` exclusion list has no `` ` ``), while `sanitizePeerDisplayString` strips it — and `sanitizeClientName` trims surrounding whitespace, which the new helper doesn't. A peer's `ClientName` runs through both (ingest, then registry write in `addPeer`), so the value shown in the dashboard and the value stored can differ.

**Why it matters:** The comment invites a reader to treat the two as interchangeable. They aren't, and the divergence is a genuine (if benign today) inconsistency, not just a cap.

**Fix:** Soften the comment to claim only what's true (the cap is shared, the filtering differs), or route both paths through one helper so the claim becomes real.

### ChiR6 — `sanitizePeerHexString` bounds the width but doesn't pin it

*Same as [@freemans13's point "sanitizePeerHexString bounds width without pinning it" (10-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1531#pullrequestreview-4898386641).*

**Problem:** The helper accepts any hex string of length 1..64, including odd lengths. For hash-shaped fields — `BestBlockHash` on the `best_height == 0` path (`Server.go:1060`), where no canonical re-parse happens — a peer can push a short, non-hash value like `DEADBEEF` and have it surface as `best_block_hash`. `ChainWork` legitimately varies in width, so it's fine there; the hash fields are the ones with a fixed expected shape.

**Why it matters:** A dashboard reader sees a well-formed-looking hex value that is not a real block hash. It's a display-integrity nit, not a safety hole — hence low — but the "plausible hex" bar is looser than the field's actual shape.

**Fix:** For hash fields specifically, require exactly 64 hex digits (or at least even length); leave `ChainWork` as a width-bounded check. A judgement call, as @freemans13 flagged.

### ChiR7 — Storage-mode constants restate `util.DetermineStorageMode` literals with no compile-time link

*Same as [@freemans13's point "Storage-mode constants duplicated with no compile-time link" (10-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1531#pullrequestreview-4898386641).*

**Problem:** `sanitize.go:30-31` defines `storageModeFull = "full"` / `storageModePruned = "pruned"`, duplicating the bare literals `util.DetermineStorageMode` (`util/storage_mode.go:22,27,48`) returns. Nothing ties the two together, so if the producer's vocabulary ever changes, the enum allowlist silently starts dropping honest values.

**Why it matters:** Minor, but it's a producer/validator pair that can drift apart without a compiler complaint.

**Fix:** Export the mode strings as constants from `util` and reference them from both `DetermineStorageMode` and `sanitizePeerEnum`.

## Recap

| ChiR1 | CI skips vitest guard | YES | 65% |
| ChiR2 | Unescaped forks `.html()` XSS | | 55% |
| ChiR3 | BaseURL raw + misleading comment | | 40% |
| ChiR4 | Reflection guard wrong struct | | 30% |
| ChiR5 | "Same cap" comment overstated | | 20% |
| ChiR6 | Hex width not pinned | | 12% |
| ChiR7 | Storage constants duplicated | | 6% |

ctnguyen (approved):
Review at commit [1a5b24a](https://github.com/gokutheengineer/teranode/commit/1a5b24a5353330ea8720332c246612ccefa3fc15)

The one thing that had to be sorted before merge is sorted: the vitest suite that guards this whole fix now runs in CI, so the escaping invariant can no longer regress silently behind a green build. I re-read the actual head — the workflow, `json-highlight.ts`, the service sanitizers, and where the highlighter output lands — and the core fix is unchanged and still right: single-pass structural spans, `"` left unescaped because the output only ever lands in `<pre>` element content (confirmed at `index.svelte:83`), and the tests pin the decision. The only substantive change since the last round is the CI wiring plus a merge from `main`, which is exactly what I'd want to see for a blocking-only follow-up.

## Status of previous points

ChiR1 was the sole blocking point and it is fixed. ChiR2–ChiR7 were all non-blocking; the code between the two review commits is unchanged apart from the CI step and a merge, so — as expected for a targeted follow-up — they remain open. ChiR2–ChiR6 are documented residuals/nits the prior round already accepted as deferred; none has regressed.

| ChiR1 | CI skips vitest guard | YES | FIXED |
| ChiR2 | Unescaped forks `.html()` XSS | | NOT FIXED |
| ChiR3 | BaseURL raw + misleading comment | | NOT FIXED |
| ChiR4 | Reflection guard wrong struct | | NOT FIXED |
| ChiR5 | "Same cap" comment overstated | | NOT FIXED |
| ChiR6 | Hex width not pinned | | NOT FIXED |
| ChiR7 | Storage constants duplicated | | NOT FIXED |

No new findings at this head.

PR #1532, opened 2026-08-07
Require authentication on mutating p2p data-plane RPCs

gokutheengineer:
**Closes bitcoin-sv/teranode#4734.**

### Problem

All four items in the issue were confirmed still present on `main`:

- `adminProtectedMethods()` protected only 7 operator-facing RPCs. Ten *mutating* data-plane RPCs were reachable with no caller identity at all, validated only by whether the target peer ID was syntactically decodable - and peer IDs are cheap to mint offline.
- `ReportValidatedChainProgress` is the trust anchor sync-peer selection was hardened to use *instead of* self-reported heights. Writable unauthenticated, that hardening was bypassable: report a large height/chainwork for a Sybil, then `RecordCatchupMalicious` every honest peer (enumerated for free via the read-only `GetPeerRegistry`), and the Sybil becomes the only eligible sync peer.
- `p2p_grpcListenAddress` defaulted to `:${P2P_GRPC_PORT}` - all interfaces - in every context except `.dev`.
- `UpdateCatchupReputation` was a no-op stub kept "for legacy compile compatibility" with no remaining caller anywhere in the tree.

### Fix

**Authentication.** All 10 mutating data-plane RPCs move into the protected set, renamed `adminProtectedMethods` -> `authProtectedMethods` since it is no longer only admin RPCs. No caller changes were needed: every internal consumer goes through `p2p.NewClient`, which already attaches `grpc_admin_api_key`. Only read-only queries remain anonymous.

I took the issue's first option (one key) rather than a second service credential. A separate key only buys least-privilege if services are actually provisioned different keys, and today `p2p.Client` hands the admin key to every consumer regardless - so the split would need client-side per-method key selection to mean anything. Noted under Residual rather than half-built.

**Bind.** The committed default becomes loopback, with wide binds added exactly where `p2p_grpcAddress` is non-loopback (`docker.m`, `docker.ss`, `operator`) plus the generated split-mode compose contexts. This matters more than it looks: `settings.conf` ships `grpc_admin_api_key = testkey`, so on an un-tuned node the key authenticates nobody and the loopback bind is the control that actually closes the hole.

**Startup warning.** `warnIfGRPCExposed` flags the combination the interceptor cannot save - bound to all interfaces with a placeholder or sub-16-char key. `p2p.Client` now also warns when the key is unset, since that is where the rejections surface.

**Dead RPC.** `UpdateCatchupReputation` deleted from the proto (regenerated), client, interfaces, mocks and docs. No mixed-version risk: nothing in the tree ever invoked it, so an old client cannot hit the new `Unimplemented`.

**Beyond the issue**, from the fresh-eyes review: the loopback default would have broken the split multinode compose topology, where `compose/cmd/gennodes` emits `p2p_grpcAddress.docker.teranodeN = teranodeN-p2p:PORT` with p2p in its own container. Blockvalidation and subtreevalidation would have gotten connection-refused on `GetPeersForCatchup` and found zero catchup peers, while the container healthcheck (`nc -z localhost 9904`) kept reporting healthy. The template now emits a matching listen override.

### Residual

- **`grpc_admin_api_key = testkey` is still the committed default.** Where an operator has widened the bind without changing it, the key authenticates nobody. Changing the shipped value would break every existing deployment, so this is a warning, not a behavior change.
- **Separate service-vs-admin credentials** not implemented, for the reason above.
- **Read-only RPCs remain anonymous by design**, but `GetPeerRegistry` / `GetPeersForCatchup` dump peer IDs, DataHub URLs, heights, reputation and ban state. Documented in `p2p_reference.md` as a reason to keep a widened port cluster-internal.
- **Empty `grpc_admin_api_key` now degrades more than before**: peer reporting fails (advisory-only, logged, does not break catchup) rather than just admin RPCs. Both client and server warn.
- **The source guard is an approximation, not a proof** - it cannot see mutation through an interface value or another package. Said so in its doc comment; its own detection is regression-tested.
- **Not fixed, separate issue worth filing:** `services/legacy/Server.go` protects only `BanPeer`/`UnbanPeer` on the legacy `peer_api.PeerService`, leaving `ClearBanned` unauthenticated - the same class of bug - and it logs the generated admin API key at `Infof`.

### Known gap: the same registry writes are still unauthenticated on the blockchain service

This PR locks the P2P `PeerService` door; the equivalent side door is still open, and
that is worth stating plainly rather than leaving implied.

`services/blockchain/Server.go` registers `PeerRegistryService` on the blockchain gRPC
port with `nil` auth options, so no interceptor is installed at all. That service
exposes the same writes this PR gates one hop earlier - `RecordValidatedPeerProgress`,
`UpdatePeerMetrics` (including `recordMalicious`), `AddBanScore`, `ClearBannedPeers`,
`RemovePeer` - and with weaker validation than the p2p path: no `peer.Decode` check and
no `maxReportedChainWorkBytes` cap. `blockchain_grpcListenAddress` binds all interfaces
in every context except `.dev`, and p2p's `peerRegistry` is always a
`blockchain.PeerRegistryClient` built with `&util.ConnectionOptions{}` - no key - so the
newly protected reporters are thin forwarders to an unprotected service.

Net effect: in the split deployments where this hardening matters (`docker.m`,
`docker.ss`, `operator`), anyone who can reach the blockchain port can still write
validated height and chainwork for a Sybil peer and flag every honest peer malicious,
with no credential. Until that is closed, the p2p key narrows the attack surface but
does not eliminate the sync-peer takeover.

Fixing it means passing real auth options covering the mutating `PeerRegistryService`
methods and giving the p2p-side client the key. Narrowing
`blockchain_grpcListenAddress` the way this PR narrowed the p2p one is **not** an
alternative: `blockchain_grpcAddress` is a routable container/k8s address in exactly
those contexts (`blockchain:8087`, `blockchain-1:8087`, `k8s:///blockchain...`), so a
loopback bind would break every service that dials blockchain. It only "fixes" the
contexts that were never exposed.

Tracked as follow-up work, not addressed here.

### Also out of scope

`services/legacy/Server.go` protects only `BanPeer`/`UnbanPeer` on the legacy
`peer_api.PeerService`, leaving `ClearBanned` unauthenticated - the same class of bug.
Needs its own issue.

### Tests

New in `services/p2p/server_auth_test.go`:

- `TestPublicRPCsDoNotMutateRegistry` - the issue's ask, implemented as a source guard: parses the package, walks each public handler transitively, and fails if it reaches a write on the peer registry, ban list, registry batcher, sync coordinator, gossip caches or ban channel. Read-only methods are allow-listed, so anything new counts as a write until deliberately listed.
- `TestFindGuardedWritesDetectsEscapes` - keeps that guard honest against 13 synthetic escape classes (helper recursion, free function taking `*Server`, batcher, `sync.Map`, channel send, goroutine, method value, field replacement, pointer/argument escape, local alias, value receiver) plus 3 read-only cases.
- `TestProtectedRPCsAreTheMutatingOnes` - the reverse direction, so a read path does not stay behind the key by inertia.
- `TestReportValidatedChainProgressRequiresAuth`, `TestRecordCatchupMaliciousRequiresAuth` - wire-level over bufconn against the real `PeerService` behind the real interceptor. Both assert the rejection *and* that registry state is unchanged, then prove the attack would have worked by repeating the call with the key: validated height/chainwork get written, and one malicious flag flips `PeerSelector.isEligible` to false.
- `TestPublicRPCsStayAnonymousOverTheWire` - read-only queries still work with no key.
- `TestGRPCAuthOptionsProtectEveryMutatingRPC` - pins the wiring `Start` hands to `util.StartGRPCServer`, so dropping `ProtectedMethods` cannot leave the suite green.
- `TestBindsAllInterfaces` - wildcard detection across IPv4/IPv6 spellings.

Commands run (affected packages only):

```
SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./services/p2p/... ./settings/... ./compose/...   # ok
SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./services/blockvalidation/catchup/...            # ok
go vet -tags testtxmetacache ./services/p2p/... ./services/blockvalidation/... ./services/subtreevalidation/... ./services/rpc/... ./settings/... ./compose/... ./daemon/...   # clean
gofmt -l services/p2p services/blockvalidation settings compose   # clean
```

`services/blockvalidation`, `services/subtreevalidation` and `services/rpc` cannot be test-linked on this machine (unrelated pre-existing GoBDK cgo link failure: undefined `_TxValidator_Set*` symbols); `go vet` type-checks them including their tests, and both were exercised for compilation.


ctnguyen (changes requested):
Review at commit [f434844](https://github.com/gokutheengineer/teranode/commit/f434844c87875a7c5f2e783496cc7f3f2efb4233)

Solid, well-scoped security fix: the ten mutating data-plane RPCs move behind the API key, the default bind drops to loopback, the dead `UpdateCatchupReputation` RPC is cleanly removed, and the design fails closed. The bufconn tests are the highlight — they assert the rejection *and* that registry state is unchanged, then re-run with the key to prove the attack would otherwise have landed. One issue in the `docker.ss` config needs fixing before merge; the rest are minor.

## Blocking issues

### ChiR1 — `docker.ss` widens the p2p bind but leaves the client dialing loopback

**Problem:** The PR adds `p2p_grpcListenAddress.docker.ss = :${P2P_GRPC_PORT}` and documents `docker.ss` as a context that "runs P2P in its own container or pod," implying sibling services reach it cross-container. In `compose/docker-compose-ss.yml` they are indeed separate containers (`p2p-1`, `blockvalidation-1`, `subtreevalidation-1`, all `SETTINGS_CONTEXT=docker.ss.teranode1`), but there is no `p2p_grpcAddress.docker.ss` override, so the client address falls back to the base `p2p_grpcAddress = localhost:${P2P_GRPC_PORT}`. `blockvalidation-1`/`subtreevalidation-1` therefore dial their own loopback for `GetPeersForCatchup` and the reporters and never reach `p2p-1`. The gennodes split template gets this right — it sets both the address (`teranodeN-p2p:...`) and the listen override.

**Why it matters:** The `docker.ss` change is self-contradictory as shipped. The documented split-container safety story does not hold: `GetPeersForCatchup` gets connection-refused → zero catchup peers, while the `nc -z localhost 9904` healthcheck still reports healthy — the exact silent failure this PR set out to prevent for gennodes. Meanwhile the widened bind keeps the unauthenticated read-only peer registry (`GetPeerRegistry`, peer IDs, DataHub URLs, heights, reputation, ban state) reachable on the Docker network for a reach that never happens. (The wide bind itself is not a new exposure — `docker.ss` inherited the old global `:${P2P_GRPC_PORT}` default before this PR too — but the PR now presents it as a deliberate, load-bearing choice, and that choice is ineffective.) Since the fix is one line and touches config this PR already changed, it should land here rather than as a follow-up.

**Fix:** If catchup across containers is intended for this topology (it appears to be — `p2p_httpAddress.docker.ss.teranode1` already points at `p2p-1`), mirror that for gRPC:

```ini
p2p_grpcAddress.docker.ss.teranode1 = p2p-1:${P2P_GRPC_PORT}
```

Otherwise drop `p2p_grpcListenAddress.docker.ss` so the port stays loopback-only and the read-only registry is not offered to the Docker network for nothing.

## Non-blocking issues

### ChiR2 — `p2p_settings.go` longdesc omits `docker.ss` from the widened contexts

**Problem:** The `GRPCListenAddress` `longdesc` in `settings/p2p_settings.go` still says the bind is widened only in "the shipped docker.m and operator contexts," but `settings.conf`, the rendered `p2p_settings.md`, and `p2p_reference.md` all now list `docker.ss` as well.

**Why it matters:** No runtime effect, but this is the source-of-truth metadata that generates the settings docs, so it will render a doc that disagrees with `settings.conf`. (Whether `docker.ss` stays in the widened set at all depends on ChiR1.)

**Fix:** Add `docker.ss` to the list, e.g. "the shipped docker.m, docker.ss and operator contexts do this" — keeping it consistent with whatever ChiR1 settles on.

### ChiR3 — `interface.go` GRPCAdminAPIKey longdesc contradicts its own Warning and the runtime

**Problem:** The updated `GRPCAdminAPIKey` `longdesc` in `settings/interface.go` gained a Warning bullet saying an empty key makes "a random key is generated at startup, so every protected RPC rejects all callers," but the same field's Values bullet still reads "**empty** (default) - Admin API key requirement disabled (insecure)" and its Security bullet still says "Empty value disables admin API key requirement (insecure for production)." In the checked code an empty key is *not* a disabled requirement: both `services/p2p/Server.go` (`grpcAuthOptions`) and `services/legacy/Server.go` generate a random 32-byte key and keep the interceptor installed, so protected RPCs fail closed.

**Why it matters:** No runtime effect, but this is the source-of-truth metadata that renders the settings docs, and it now tells operators two opposite things about the security posture of an empty key — "disabled/insecure" versus "fail closed, callers rejected." The Warning points the right way, so it is minor, but the Values/Security bullets should not still describe the pre-hardening behavior.

**Fix:** Reword the Values and Security bullets to match the Warning and the code, e.g. "**empty** (default) - a random key is generated at startup; protected RPCs reject every caller until a key is set."

### ChiR4 — regenerated gRPC stub reports an older protoc than the file it replaces

**Problem:** `services/p2p/p2p_api/p2p_api_grpc.pb.go` had its header change from `protoc v6.33.2` to `protoc v6.32.0`, while `p2p_api.pb.go` carries no such downgrade — so the two generated files now disagree on the toolchain that produced them.

**Why it matters:** Nothing functional, but the next person who runs `make gen` with the pinned toolchain gets a spurious one-line diff flipping the version back, which is noise in an otherwise clean regenerate.

**Fix:** Regenerate both stubs with the repo's pinned protoc so the version headers are consistent.

## Recap

| ChiR1 | docker.ss bind vs client address mismatch | YES | 65% |
| ChiR2 | p2p_settings.go longdesc omits docker.ss | | 15% |
| ChiR3 | interface.go empty-key longdesc contradiction | | 12% |
| ChiR4 | protoc version drift in stub | | 8% |

ctnguyen (approved):
Review at commit [a2859e8](https://github.com/gokutheengineer/teranode/commit/a2859e827b0e005c90c5db977399f4f9e9bbacd5)

This round is in good shape and I don't think anything here needs to block merge. `docker.ss` now sets both halves (`p2p_grpcAddress.docker.ss.teranode1 = p2p-1:${P2P_GRPC_PORT}` at `settings.conf:911`, matching the `p2p-1` container in `compose/docker-compose-ss.yml`), and turning that class of mistake into an invariant with `settings/p2p_bind_consistency_test.go` is a better outcome than a one-line fix. The auth classification checks out: all 25 remaining `PeerService` RPCs are classified exactly once, the eight public ones are genuinely read-only (I read `GetPeers`, `GetPeer`, `GetPeerRegistry`, `IsBanned`, `ListBanned`, `GetPeersForCatchup`, `IsPeerMalicious`, `IsPeerUnhealthy` at head and none of them write), and the streaming guard from [bsv-blockchain/teranode#1257](https://github.com/bsv-blockchain/teranode/pull/1257) survives. The `UpdateCatchupReputation` removal is clean — zero references left anywhere in the tree, docs included. Two of the four previous points are still open, and the findings below are mostly about the edges of the new guard rails rather than the change itself.

## Status of previous points

ChiR1 is verified against the real compose file, not just the settings diff: `docker-compose-ss.yml` has exactly one node and its p2p container is `p2p-1`, so the single `docker.ss.teranode1` override is sufficient, and `p2p_grpcListenAddress.docker.ss` resolves for it through the context hierarchy the same way `p2p_httpListenAddress.docker.ss` already did. ChiR3 has become more load-bearing than it was, for the reason in ChiR5 below — it is the doc an operator reads right before creating that failure mode, so it is worth closing in this PR rather than later.

| ChiR1 | `docker.ss` bind/client mismatch | YES | FIXED |
| ChiR2 | `p2p_settings.go` longdesc context list | | FIXED |
| ChiR3 | `GRPCAdminAPIKey` longdesc self-contradictory | | NOT FIXED |
| ChiR4 | gRPC stub reports older protoc | | NOT FIXED |

## Non-blocking issues

### ChiR5 — an empty `grpc_admin_api_key` permanently disables validated-progress reporting, and the warning undersells it

**Problem:** `grpcAuthOptions` (`services/p2p/Server.go:922`) generates a throwaway random key when `grpc_admin_api_key` is empty, installs the interceptor anyway, and lets the service start. The `p2p.Client` in the same deployment reads the same empty setting and sends no key (`services/p2p/Client.go:59-70`), so all ten reporters fail with `Unauthenticated` on every call, forever. I traced every caller and they are genuinely advisory — `services/blockvalidation/peer_metrics_helpers.go:28,56,166,186,203,224,307,318`, `services/blockvalidation/get_blocks.go:875,946,980,1029`, `services/subtreevalidation/check_block_subtrees.go:147,301` all log at `Warnf` and continue, so catchup does not abort, which matches the Residual note. What the note misses is the second-order effect: `ReportValidatedChainProgress` never landing means `peerHasValidatedWork` (`services/p2p/sync_coordinator.go:218`) is false for every peer permanently, so `isEligibleByProgress` (`services/p2p/peer_selector.go:564-576`) can never take the validated branch and falls through to the bounded advertised-probe path on every selection round, for the life of the node.

**Why it matters:** [bsv-blockchain/teranode#1257](https://github.com/bsv-blockchain/teranode/pull/1257) left the reporters public on exactly this reasoning. Reversing that is the right call and fail-closed is the right direction, so this is not a merge blocker — `settings.conf:640` ships `grpc_admin_api_key = testkey`, no shipped context is affected, and both sides warn. But the warning text says protected RPCs "will reject callers", which reads as "admin operations stop working"; it does not say that sync-peer selection silently drops back to the unproven bootstrap tier that [bsv-blockchain/teranode#1201](https://github.com/bsv-blockchain/teranode/pull/1201) was built to graduate off. Combined with ChiR3 — `settings/interface.go:61` still tells operators empty means "Admin API key requirement disabled (insecure)" — the operator who wants to turn auth off will blank the key and get this, with no signal that it cost them anything beyond admin RPCs.

**Fix:** Close ChiR3 in this PR rather than leaving the two docs contradicting each other, and name the real consequence in both warnings, e.g. "…so peer reputation and validated chain progress are never recorded and sync-peer selection stays on the unproven bootstrap path". If you would rather make it unmissable, fail fast instead — no shipped context blanks the key, so nothing in-tree regresses:

```go
if apiKey == "" {
	return nil, errors.NewConfigurationError(
		"grpc_admin_api_key is not set; the PeerService cannot start because every mutating RPC, " +
			"including the catchup and validated-progress reporters, would reject all callers")
}
```

I'd take either; the docs half is the part I'd not skip.

### ChiR6 — the exposure warning only fires for wildcard binds, so the more careful operator gets less warning

**Problem:** `warnIfGRPCExposed` (`services/p2p/Server.go:956`) gates on `bindsAllInterfaces` (`services/p2p/Server.go:975`), which returns true only for an empty host or an unspecified IP. A bind to a specific routable address — `10.0.1.50:9904`, or a container's own DNS name like `p2p-1:9904` — returns false, so no warning fires even with the committed `testkey`. `TestWarnIfGRPCExposed` pins this deliberately: `{"specific interface with placeholder key", "10.0.1.50:9904", placeholderAdminAPIKey, false}`.

**Why it matters:** The doc comment justifies it as "an operator naming an interface on purpose", but intent to bind is not the same as safety of the bind. The condition the warning exists to catch is "this port is reachable from another host and the key authenticates nobody", and `10.0.1.50:9904` with `testkey` satisfies it exactly as much as `:9904` does. The result inverts the incentive: the operator who does the careful thing and names one LAN interface is told nothing, while the one who binds everything is warned. The shipped wide contexts all use `:`, so this gap only affects hand-tuned deployments — which are the ones running with a non-default bind in the first place.

**Fix:** Key the warning off "not loopback" rather than "wildcard", and flip the test expectation for the specific-interface case:

```go
// bindReachableOffHost reports whether the listener accepts connections from
// anywhere but this host: a wildcard bind, a routable IP, or a DNS name.
func bindReachableOffHost(listenAddress string) bool {
	host, _, err := net.SplitHostPort(listenAddress)
	if err != nil {
		return false
	}
	switch {
	case host == "":
		return true
	case host == "localhost":
		return false
	}
	if ip := net.ParseIP(host); ip != nil {
		return !ip.IsLoopback()
	}
	return true // a DNS name that is not localhost
}
```

That is the negation of `addressIsLoopback` in `settings/p2p_bind_consistency_test.go`, so it is worth putting one copy somewhere both can use and dropping the second spelling.

### ChiR7 — the source guard's field list fails open for new `Server` fields, and its doc comment misstates its own blind spots

**Problem:** Two things. `readOnlyMethodsByField` enumerates twelve fields, and `guardedField` returns "not guarded" for anything not in that map. So the per-field method allow-list fails closed — which is what the comment describes and what makes the guard worth having — but the *field set itself* fails open: `Server` has 42 fields, and a write through one of the other 30 is invisible to `findGuardedWrites`. `syncConnectionTimes` (`services/p2p/Server.go:156`) is a live example, a `sync.Map` of per-peer sync state in the same class as `blockPeerMap`/`subtreePeerMap`, which *are* listed. Separately, the comment sells the guard short: it says mutation reached through "a function that takes `*Server` as a parameter" is invisible, but `serverIdents` collects `*Server` parameters as well as the receiver and `findGuardedWrites` recurses into plain same-package functions via `callee`, and `TestFindGuardedWritesDetectsEscapes` proves it with `writeViaFreeFunc`/`freeWriter` in the `dirty` list. The actual blind spots are narrower: interface values, other packages, a `*Server` reached through anything other than a plain identifier, and unclassified fields.

**Why it matters:** The claim "anything new counts as a write until deliberately listed" is the property the next reviewer will lean on, and today it holds for new methods but not for new fields — the direction drift is most likely to come from. The comment being wrong in the other direction is smaller but compounds it: it tells a reader the guard is weaker than it is in a place where it is stronger, and stronger than it is where it is weaker.

**Fix:** Add `syncConnectionTimes` (and `notificationCh`, which a public handler should also never send on), correct the blind-spot sentence, and make the field set self-maintaining the way the RPC set already is:

```go
// inertServerFields are Server fields a public handler may touch freely:
// config, loggers, immutable names, client handles. Every other field must
// appear in readOnlyMethodsByField.
var inertServerFields = map[string]bool{"logger": true, "settings": true, /* ... */}

func TestEveryServerFieldIsClassified(t *testing.T) {
	for _, f := range serverStructFields(t) { // parse the struct from the AST you already build
		_, guarded := readOnlyMethodsByField[f]
		require.True(t, guarded || inertServerFields[f],
			"Server field %q is unclassified: add it to readOnlyMethodsByField or inertServerFields", f)
	}
}
```

### ChiR8 — the widening guidance stops at "set a strong key", but the key now rides the hot path in cleartext

**Problem:** Before this PR the admin key was attached to occasional operator RPCs. Now `p2p.Client` sends `x-api-key` on every catchup metric, every valid-block/subtree report, every `RecordBytesDownloaded` — and `util.GetGRPCClient` uses `insecure.NewCredentials()` unless `security_level_grpc` is non-zero, which defaults to `0` ("Insecure plaintext connections", per its own longdesc). The new section in `docs/references/services/p2p_reference.md` and the `warnIfGRPCExposed` message both tell the operator to set a strong key and restrict the port; neither mentions that on a widened bind the key itself is now on the wire continuously.

**Why it matters:** For the shipped contexts that widen the bind and cross a host boundary (`operator`, and `docker.ss` on the compose network), an on-path observer recovers the credential from ordinary traffic rather than having to guess it. That is a different exposure from the pre-PR pattern, and the operator-facing note is the thing they will act on.

**Fix:** Add a sentence to the widening guidance in `p2p_reference.md` and to the `GRPCListenAddress` longdesc recommending `security_level_grpc` when the bind crosses a network boundary. If you take ChiR6, "reachable off-host and `SecurityLevelGRPC == 0`" is a cheap extra clause on the same warning.

### ChiR9 — the bind-consistency test's context list has two soft spots

**Problem:** The `""` entry calls `NewSettings()` with no argument, which resolves `gocore.Config().GetContext()` — the ambient `SETTINGS_CONTEXT`. The subtest is named `default` but under `SETTINGS_CONTEXT=test` it re-tests `test`, and under the Makefile's `docker.ci` it tests `docker.ci`; the one line this PR actually changed, `p2p_grpcListenAddress = localhost:${P2P_GRPC_PORT}` at `settings.conf:921`, is never asserted under its own name. Second, `shippedP2PContexts` is hand-maintained and omits contexts in real use — `docker.ci` (the Makefile's `SETTINGS_CONTEXT_DEFAULT`), `docker.teranode2.test`/`docker.teranode3.test`, `docker.host.teranode2`/`teranode3`. They all resolve consistently today, but a hand-maintained list is how ChiR1 happened. Minor and separate: the comment in `addressIsLoopback` says the `k8s:///` form "splits on the last colon and still yields a host" — it does not, `net.SplitHostPort` rejects it with `too many colons in address`, so the operator context is covered by the error branch, not the DNS-name branch. Right answer, misleading reason.

**Why it matters:** The test is the durable part of the ChiR1 fix, so its coverage claim is what the next person relies on. Right now it does not cover the committed defaults, and adding a context to `settings.conf` does not force anyone to add it here.

**Fix:** Use a context with no `p2p_` overrides so the committed base values are what gets asserted, and fill in the missing shipped contexts:

```go
var shippedP2PContexts = []string{
	"committed-defaults-no-overrides", // no p2p_* overrides exist for this, so the base values resolve
	"dev", "test", "docker.ci",
	"docker.m", "docker.ss.teranode1",
	"docker.host.teranode1", "docker.host.teranode2", "docker.host.teranode3",
	"docker.teranode1.test", "docker.teranode2.test", "docker.teranode3.test",
	"operator", "operator.teratestnet",
}
```

Better still, derive the list by scanning `settings.conf` for `p2p_grpcAddress.*` / `p2p_grpcListenAddress.*` suffixes so a new context cannot be added without being checked. And fix the `addressIsLoopback` comment to say `SplitHostPort` errors on the scheme-qualified form.

### ChiR10 — the API key comparison is not constant time

**Problem:** `util.CreateAuthInterceptor` (`util/grpc_helper.go:562`) compares with `keys[0] != apiKey`. Pre-existing, not introduced here, but this PR makes that single comparison the sole access control on ten more RPCs including the sync-selection trust anchor.

**Why it matters:** Low practical risk over gRPC, but it is a one-line fix on a check that is now load-bearing for the data plane rather than just for occasional admin calls.

**Fix:** `subtle.ConstantTimeCompare([]byte(keys[0]), []byte(apiKey)) == 1`.

## Recap

| ChiR5 | Empty key degrades validated-progress tier | | 45% |
| ChiR6 | Warning misses non-wildcard exposed binds | | 30% |
| ChiR7 | Guard field list fails open | | 25% |
| ChiR8 | Hot-path key sent in plaintext | | 20% |
| ChiR9 | Bind-consistency test coverage gaps | | 15% |
| ChiR10 | Non-constant-time key comparison | | 10% |

settings.conf
```diff
@@ -948,10 +948,24 @@ p2p_dht_cleanup_interval = 24h
...
-p2p_grpcListenAddress             = :${P2P_GRPC_PORT}
-p2p_grpcListenAddress.dev         = localhost:${P2P_GRPC_PORT}
+# Bind loopback by default: every mutating PeerService RPC is API-key protected,
+# but an un-tuned public bind still exposes the read-only peer registry. Only the
+# contexts that run p2p in its own container/pod - i.e. where p2p is addressed by
+# something other than localhost - widen the bind. Deliberately scoped to p2p:
+# its PeerService is the one that decides sync-peer selection from caller-supplied
+# input, so it is the one worth closing by default. Split-mode compose stacks add
+# their own override (see compose/cmd/gennodes/templates/settings.conf.tmpl).
+p2p_grpcListenAddress             = localhost:${P2P_GRPC_PORT}
```
icellan, line 965: Flipping the default bind to loopback is a silent breaking change for any settings context not shipped in this repo, and nothing detects it at runtime.

The invariant "client address routable <-> bind widened" is enforced only by `settings/p2p_bind_consistency_test.go` over the nine committed contexts. A deployment with its own context name (custom `SETTINGS_CONTEXT`, operator configmap, `settings_local.conf`) that relied on the old `:9904` default and runs p2p in its own container/pod gets loopback after upgrade. Then `blockvalidation`/`subtreevalidation` dial `p2p:9904` and get connection-refused, `selectBestPeersForCatchup` (`services/blockvalidation/peer_selection.go:41`) only warns and returns the error, every reporter call only warns, and nothing in blockvalidation's health check covers the p2p client — so the node stops finding catchup peers while its healthcheck (`nc -z localhost 9904`, `compose/cmd/gennodes/main.go:249`) stays green.

Put the check the unit test encodes into startup: in `grpcAuthOptions` (`services/p2p/Server.go:923`) compare the `P2P.GRPCAddress` host against the bind and log an error, or refuse to start, when a non-loopback client address meets a loopback bind.

settings/p2p_bind_consistency_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// shippedP2PContexts are the settings contexts committed settings.conf carries
+// p2p overrides for, plus the plain defaults. Deployment-specific contexts that
+// live in generated or gitignored config are out of scope here.
+var shippedP2PContexts = []string{
```
icellan, line 13: This asserts a property of committed `settings.conf` but actually reads the developer's merged config. `NewSettings` layers `settings_local.conf` over `settings.conf` — visible in the test's own log output (`settings_local.conf:N is replacing ...`).

A developer pointing their local node at a containerised p2p sets `p2p_grpcAddress = 10.0.0.5:9904` in `settings_local.conf` and `make test` now fails in `./settings` with a bind-consistency error unrelated to anything they changed — and the fix lives in a gitignored file, so nobody else can reproduce it. The mirror case is worse: a local `p2p_grpcListenAddress` override can mask a genuine settings.conf regression.

Either read `settings.conf` directly for this assertion, or skip when the resolved value differs from the committed file.

services/p2p/server_auth_test.go
```diff
@@ -80,13 +82,377 @@ func TestAdminProtectedMethodsCoverAllRPCs(t *testing.T) {
...
+// listed here. Fields that reach peer state only through writes (the batcher,
+// the sync coordinator, the gossip caches, the ban channel) get an empty
+// allow-list: touching them at all from a public handler is a finding.
+//
+// The guard is a source-level approximation, not a proof. It follows calls on
+// the receiver within this package only; mutation reached through an interface
+// value, a function that takes *Server as a parameter, or another package is
+// invisible to it. Treat a pass as "no obvious write", and keep reviewing new
+// public handlers by hand.
+var readOnlyMethodsByField = map[string]map[string]bool{
```
icellan, line 97: The guard misses the peer-state mutator most likely to be used next: `blockchainClient`.

`readOnlyMethodsByField` covers `peerRegistry`, `banList` and the caches, but not `blockchainClient` (`services/p2p/Server.go:129`), whose interface exposes `ReportPeerFailure` (`services/blockchain/Interface.go:913`) — a peer-state write that feeds peer selection. A future public handler calling `s.blockchainClient.ReportPeerFailure(...)` passes `TestPublicRPCsDoNotMutateRegistry` and ships unauthenticated.

I checked every current public handler (`GetPeers`, `GetPeer`, `GetPeerRegistry`, `GetPeersForCatchup`, `IsBanned`, `ListBanned`, `IsPeerMalicious`, `IsPeerUnhealthy`) — none touches the field, so adding `"blockchainClient": {}` to the map is a one-line, non-breaking tightening.

icellan (commented):
Re-review of the current head. Build and vet clean, `./services/p2p/... ./settings/... ./compose/...` green including both new guard tests, and I dumped the resolved `P2P.GRPCAddress`/`GRPCListenAddress` for every shipped context — the config half is coherent (`docker.ss.teranode1` -> `p2p-1:9904` + `:9904`, `operator*` -> `k8s:///...` + `:9904`, everything else loopback/loopback). `UpdateCatchupReputation` removal is complete, zero references repo-wide.

One finding doesn't anchor to a line in this diff, so it goes here.

**In the one shipped context that widens the bind, the new auth is nominal.** `deploy/kubernetes/teranode/teranode-configmap.yaml:7` sets `SETTINGS_CONTEXT: "operator"`; this PR sets `p2p_grpcListenAddress.operator = :9904`; the CRs (`teranode-cr.yaml:108`, `teranode-cr-mainnet.yaml:117`) put a Traefik `grpcIngress` in front of the peer gRPC port; and `grpc_admin_api_key` appears nowhere in `deploy/kubernetes/` or in the install docs' `teranode-operator-secrets`, which documents only `blockchain_store` and `utxostore`. So a miner following the shipped installation runs with `grpc_admin_api_key = testkey` from settings.conf:684 — a constant published in this repo. Anyone who can reach the peer-grpc ingress sends `x-api-key: testkey`, calls `ReportValidatedChainProgress` for a peer ID they minted and `RecordCatchupMalicious` on every honest peer: exactly the Sybil sync-peer takeover the new comments give as the reason for protecting these RPCs.

`warnIfGRPCExposed` already detects this precise combination (wide bind + placeholder key) and only calls `Warnf`. Make it fail closed on mainnet/testnet — placeholder-or-short key plus wide bind should be a startup error, or fall back to the generated random key — and add `grpc_admin_api_key` to the operator secret and the install docs.

services/p2p/Server.go
```diff
@@ -1215,22 +1197,114 @@ func (s *Server) disconnectPreExistingBannedPeers(ctx context.Context) {
...
+		// Operator-facing admin RPCs.
 		"/p2p_api.PeerService/BanPeer":         true,
 		"/p2p_api.PeerService/UnbanPeer":       true,
 		"/p2p_api.PeerService/ClearBanned":     true,
 		"/p2p_api.PeerService/AddBanScore":     true,
 		"/p2p_api.PeerService/ResetReputation": true,
 		"/p2p_api.PeerService/ConnectPeer":     true,
 		"/p2p_api.PeerService/DisconnectPeer":  true,
+
+		// Internal data-plane reporters (block/subtree validation).
```
icellan, line 1343: Gating these ten reporters makes `grpc_admin_api_key` load-bearing for sync, but no shipped multi-container stack sets one.

`deploy/docker/base/settings_local.conf:18` (docker.m — the documented production Docker miner deployment, where `peer`, `blockvalidation` and `subtreevalidation` are separate containers per `deploy/docker/base/docker-teranode.yml`), `compose/docker-compose-ss.yml` (docker.ss), the gennodes split stacks, and any operator install whose secret predates this PR all run without a key. Only `test/docker-compose.e2etest*.yml` injects one. Single-process `make dev` is affected too — `daemon/daemon_stores.go:149` builds the p2p client over gRPC regardless of co-location.

`ReportValidatedChainProgress`/`ReportValidBlock` are what set `PeerInfo.ValidatedChainWork` and `BlocksReceived`, so with them rejected, `peerAheadByValidatedWork` (`services/p2p/sync_coordinator.go:222`) and `peerHasRecentFullBlockDelivery` are permanently false. Every peer stays in the budget-gated unproven-probe tier (`p2p_max_unproven_sync_probes_per_backoff_window`, default 3) instead of locking onto a proven high-work sync peer — the node never settles on a good peer, and the only signal is one server Warnf, one client Warnf, and a Warnf per rejected report.

Two things needed: ship a key in the compose stacks the way the e2e stacks do, and make the server fail loudly — Error level or refuse to start — when the protected reporters have no key, rather than degrading silently.

icellan (commented):
Re-reviewed at `f148586d`. Build and vet clean on p2p/settings/blockvalidation/rpc/subtreevalidation/daemon/cmd; `go test ./services/p2p/... ./settings/...` green.

Three of the four things from my last pass are resolved: the committed `testkey` default is gone (via #1578), `p2p_bind_consistency_test.go` now parses committed settings.conf directly instead of the merged `settings_local.conf`, and `"blockchainClient": {}` is in the guarded-field map. The operator/k8s one is partly addressed — docs, CR and configmap now say the key is required and lives in `teranode-operator-secrets`; the wide bind plus ingress remain, and enforcement is still warn-only, which is now the `Server.go:1210` comment below.

Two new blockers inline. Two smaller ones that don't anchor to a diff line:

**`services/p2p/p2p_api/p2p_api.proto` — `UpdateCatchupReputation` removed with no `reserved` entry.** No in-repo callers remain, so the live impact is limited to mixed-version rolling upgrades (old client gets `Unimplemented`; the old handler was a no-op, so log noise) and out-of-tree consumers. Worth a `reserved` entry so the name can't be silently reused with different semantics later.

**`services/p2p/p2p_api/*.pb.go` — protobuf toolchain drift.** Regenerated with protoc-gen-go v1.36.12 / protoc v7.35.1 / protoc-gen-go-grpc v1.6.2; the rest of the repo is on v1.36.11 / protoc v6.33.2-v7.34.1. Builds fine, but the next `make gen` churns it back.

**`deploy/docker/base/settings_local.conf:18` (plus `.template:114` and `docs/howto/miners/docker/*`) — Docker miner config and docs never updated for the widened requirement.** Not touched by this PR, so no inline anchor. The comment there still scopes the key to "protected gRPC admin RPCs (setban/clearbanned, connect/disconnect peer, alert-driven bans, reset reputation)" and ships it commented out, but after this PR the key also gates the ten data-plane reporters. Only the k8s docs got the "required for a working node" treatment, so an operator following the Docker guide hits the degradation above with no signal.

services/p2p/Server.go
```diff
@@ -1215,22 +1197,114 @@ func (s *Server) disconnectPreExistingBannedPeers(ctx context.Context) {
...
+// Placeholder and weak-key handling lives in util.ValidateAdminAPIKey, which is
+// shared with the legacy and RPC services; a placeholder is ignored so the
+// random-key path below applies instead. When no usable key is configured that
+// leaves every protected RPC unreachable rather than open - the safe direction,
+// since these RPCs steer sync-peer selection.
+func (s *Server) grpcAuthOptions() (*util.AuthOptions, error) {
+	listenAddress := s.settings.P2P.GRPCListenAddress
+
+	apiKey := s.settings.GRPCAdminAPIKey
+	if util.ValidateAdminAPIKey(s.logger, "P2P", apiKey, listenAddress, s.settings.SecurityLevelGRPC) {
```
icellan, line 1224: The merge dropped the short-key fail-closed guard this branch used to have.

`9db63e4e3` added `checkGRPCExposure`, which refused to start on any non-regtest network when the bind was all-interfaces **and** the key was a placeholder **or shorter than 16 chars**. The merge at `4bb106c15` (with #1578) replaced it with `util.ValidateAdminAPIKey`, which only ignores placeholders and merely `Warnf`s on a short key.

Net effect: `grpc_admin_api_key=abc` on the `operator` context — which binds `:9904` on all interfaces (`settings.conf:1021`) and is fronted by a Traefik `grpcIngress` (`deploy/kubernetes/teranode/teranode-cr.yaml:107`) — is now accepted as the real key on mainnet. Brute-force it and you get exactly the capability this PR set out to close: forge `ReportValidatedChainProgress` for a Sybil, flag honest peers malicious.

Collapsing the duplicate implementation into the shared helper is reasonable; losing the short-key fail-closed isn't. That case needs to hard-fail somewhere, whether in the helper or back here.

services/p2p/server_auth_test.go
```diff
@@ -80,13 +82,381 @@ func TestAdminProtectedMethodsCoverAllRPCs(t *testing.T) {
...
```
icellan, line 98: `blockchainClient` is covered now — `P2PClient` still isn't.

`protectedWithoutGuardedWrites` exempts `ConnectPeer`/`DisconnectPeer` precisely because the guard can't see libp2p-layer mutation, but `P2PClient` isn't a key here, so `guardedField` returns false for it entirely. A future *public* RPC that disconnects or otherwise manipulates a peer through `s.P2PClient` passes `TestPublicRPCsDoNotMutateRegistry` unnoticed.

Adding `"P2PClient"` with its read-only method allow-list closes the same hole in the other direction.

icellan:
Checked again at `447ee03b8`. One thing outside this diff that undercuts the whole change, so flagging it here rather than as an inline comment.

**The registry writes this PR gates on p2p are still reachable unauthenticated on the blockchain service's gRPC port.**

`services/blockchain/Server.go:536-540`:

```go
if err := util.StartGRPCServer(ctx, b.logger, b.settings, "blockchain", b.settings.BlockChain.GRPCListenAddress, func(server *grpc.Server) {
	blockchain_api.RegisterBlockchainAPIServer(server, b)
	blockchain_api.RegisterPeerRegistryServiceServer(server, b)
	closeOnce.Do(func() { close(readyCh) })
}, nil); err != nil {
```

That trailing `nil` is `authOptions`, so `PeerRegistryService` runs with no auth interceptor at all. `settings.conf:383-385` binds `blockchain_grpcListenAddress` to `:${BLOCKCHAIN_GRPC_PORT}` — all interfaces — for every context except `.dev`. And p2p's `s.peerRegistry` is always `blockchain.PeerRegistryClient` over gRPC (`daemon/daemon_stores.go:185`, built with `&util.ConnectionOptions{}`, no API key), so the newly protected reporters are thin forwarders to an unprotected service.

The service exposes `RecordValidatedPeerProgress`, `UpdatePeerMetrics` with `recordMalicious`, `AddBanScore`, `ClearBannedPeers` and `RemovePeer`. In exactly the split deployments where the p2p hardening matters — docker.m, docker.ss, operator — anyone who can reach the blockchain port writes validated height and chainwork for a Sybil and flags every honest peer malicious, with no credential. Same sync-peer takeover this PR set out to close, one service over, and with weaker input validation than the p2p path (no `peer.Decode` check, no `maxReportedChainWorkBytes` cap).

Two ways to resolve, either is fine by me: pass real `authOptions` covering the mutating `PeerRegistryService` methods here and give the p2p-side client the key, or narrow `blockchain_grpcListenAddress` the way this PR narrowed the p2p one. Doing it in a follow-up is reasonable, but until one of them lands the p2p key is a lock on the front door with the side door open, so it's worth being explicit about that in the PR description.

Adjacent, confirmed at the same head, and already noted as separate in your description: `services/legacy/Server.go:672` protects only `BanPeer`/`UnbanPeer`, so `ClearBanned` is unauthenticated. Worth an actual issue.

services/p2p/server_auth_test.go
```diff
@@ -80,13 +83,388 @@ func TestAdminProtectedMethodsCoverAllRPCs(t *testing.T) {
...
```
freemans13, line 98: This map is an allow-list keyed by field name, so a write to any `Server` field that is not a key here is invisible to the guard and the test still passes.

That is not the limitation the doc comment above discloses. The comment scopes the blind spot to interface values, functions taking `*Server`, and other packages. This case is a plain method call on the receiver inside this package, which is exactly what the guard says it follows.

Demonstrated rather than argued. I injected two peer-state writes into `GetPeers`, an unauthenticated handler, through `go test -overlay` so nothing on your branch changed:

```go
func (s *Server) GetPeers(ctx context.Context, _ *emptypb.Empty) (*p2p_api.GetPeersResponse, error) {
	s.syncConnectionTimes.Store("injected-sybil-peer", int64(0))
	s.liveConnCache.Store("injected-sybil-peer", liveConnCacheEntry{live: true})
```

`TestPublicRPCsDoNotMutateRegistry`, `TestProtectedRPCsAreTheMutatingOnes` and `TestAuthProtectedMethodsCoverAllRPCs` all stayed green.

To be fair about severity, neither field is much of an attack primitive today. `syncConnectionTimes` (`Server.go:169`) only supplies the sync-peer connect timestamp in the gossiped node status message. `liveConnCache` (`Server.go:183`) is a TTL cache of libp2p liveness, though the comment at `Server.go:1315` explains that marking a peer connected exempts it from registry cleanup, which is what stops a flood of fresh peer IDs growing the registry without bound.

I am raising it anyway because this is the third field to be missed, after `blockchainClient` and `P2PClient` in the last two rounds. Three in a row says something about the shape rather than about the entries. Inverting it closes the class: parse the `Server` struct in the same AST pass and require every field to appear either here or in an explicit `notPeerState` list. A new field then fails the test until somebody classifies it, which turns "remember to add it" into "the test tells you".

settings/p2p_bind_consistency_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// shippedP2PContexts are the settings contexts committed settings.conf carries
+// p2p overrides for, plus the plain defaults. Deployment-specific contexts that
+// live in generated or gitignored config are out of scope here.
+var shippedP2PContexts = []string{
```
freemans13, line 17: `shippedP2PContexts` is hand maintained, so a future `p2p_grpcAddress.<newcontext>` added to settings.conf without a matching entry here goes unchecked in silence. That is the same bug this test exists to catch, one level up.

It can be derived instead. `readCommittedSettingsConf` already parses every key, so the context set is just the suffixes of keys starting `p2p_grpcAddress.` or `p2p_grpcListenAddress.`, unioned with the empty context. A new override then cannot be added without being tested.

Deriving it would also pick up `docker.ss` itself, which is absent today. settings.conf keys the bind at the stack level (`p2p_grpcListenAddress.docker.ss = :${P2P_GRPC_PORT}`, line 1020) and the client address at the node level (`p2p_grpcAddress.docker.ss.teranode1`, line 1008). Resolving plain `docker.ss` therefore yields a widened bind against a loopback client address, the second of the two cases your doc comment describes, where the unauthenticated read-only registry is offered to the network for a reach that never happens.

Nothing is broken today. `compose/docker-compose-ss.yml` only ever sets `SETTINGS_CONTEXT: docker.ss.teranode1`, and the split key levels match existing house style, since `p2p_httpListenAddress.docker.ss` at line 1035 does the same thing. So this is about the test's reach, not about the config.

freemans13 (approved):
Approving the diff. Build, `go vet` and the full-repo lint form CI uses (`golangci-lint run --timeout=8m --disable gosec --disable prealloc`, 0 issues) are clean, and `go test -count=1 ./services/p2p/... ./settings/... ./util/` passes at `3d006b507`.

I checked the guard tests actually bite instead of assuming it. Deleting `ReportValidBlock` from `authProtectedMethods` via `go test -overlay` (so your branch was never touched) fails `TestAuthProtectedMethodsCoverAllRPCs` at `services/p2p/server_auth_test.go:61`. That is the part of this PR I trust most.

Two things I tried to break and could not, worth recording because they were the obvious worries about making the key load-bearing for sync:

**The reporters degrade, they do not storm.** Every call site logs at Warn and carries on, and none of them fails the block: `peer_metrics_helpers.go:203`, `:365`, `:376`, `get_blocks.go:875`, `:946`, `:980`, `:1029`, `check_block_subtrees.go:147`, `:301`, `SubtreeValidation.go:856`. The client retry interceptor only retries `Unavailable` and `DeadlineExceeded` (`util/grpc_helper.go:370`), so an `Unauthenticated` reply costs one round trip rather than `GRPCMaxRetries` of them. The residual cost is log volume, since `RecordBytesDownloaded` runs per subtree fetch, so a keyless split deployment gets a Warn per subtree all the way through catchup.

**`docker.host` survives the loopback default.** That compose file is `network_mode: "host"` throughout (`test/docker-compose-host.yml:3`), so its containers share one loopback. The gennodes split template pairs both overrides correctly under `{{- if not $.AllInOne }}`.

On the `reserved` request from an earlier round, your comment at `services/p2p/p2p_api/p2p_api.proto:318` is right, and I tested it rather than taking it on trust. protoc 35.1 rejects `reserved` inside a `service` body in both proto3 and edition 2023, with `Expected "rpc".` Nothing actionable there, and the comment is the correct substitute for it.

Two non-blocking notes inline. Both are about the shape of a test allow-list rather than about shipped behaviour, and neither should hold up merge.

One last thing, which is not mine to resolve. The unauthenticated `PeerRegistryService` on the blockchain service raised in the 31 August comment still stands and I think it is right. It does not belong in this diff, but until one of the two fixes lands, the description should say plainly that this key closes one of two doors onto the same registry writes.

Not checked: I did not stand up a split container stack, so everything above about `docker.m`, `docker.ss` and `operator` is read off settings.conf and the compose files, not observed.

PR #1544, opened 2026-08-10
fix(p2p): set http.Server timeouts on the p2p HTTP server

gokutheengineer:
**Closes bitcoin-sv/teranode#4737.**

### Problem

The p2p service's echo HTTP server was a zero-value `http.Server`: `ReadTimeout`, `ReadHeaderTimeout`, `WriteTimeout` and `IdleTimeout` all unset, i.e. unlimited, on a port that binds all interfaces by default. A client that dribbles request headers (slowloris) or parks idle keep-alive connections holds a connection, a goroutine and an fd forever. The validator, asset and blockpersister servers all set these; p2p was the outlier. Confirmed still present at `services/p2p/Server.go` (`setupHTTPServer` / `StartHTTP`).

### Fix

Set the timeouts in `setupHTTPServer`, where the server is built, as named constants asserted by the regression test:

- `ReadHeaderTimeout` 10s, `ReadTimeout` 30s, `WriteTimeout` 30s, `IdleTimeout` 120s.

These land on the served instance for both paths: `StartHTTP` serves via `s.e.Server.Serve` / `s.e.Server.ServeTLS` and never uses echo's `Start`/`StartTLS`, so `e.TLSServer` is never involved.

One improvement over the issue's proposal: the issue said `WriteTimeout` must stay unset or it would kill `/p2p-ws` streams. That's not how `net/http` behaves - `Hijack` clears the connection deadlines before the handler gets the conn (and gorilla/websocket clears them again on upgrade, including when a `HandshakeTimeout` is set). So none of these timeouts apply to an established websocket, and `WriteTimeout` is set too, matching every sibling service and closing the slow-read half of the same DoS. This is proven by a dedicated test (see below).

The in-code comment and the docs scope the claim precisely: the timeouts bound plain HTTP exchanges and incomplete `/p2p-ws` upgrades; they do not bound an established websocket stream, whose liveness (read deadlines, ping/pong, connection caps) is separate hardening work (see Residual).

Also fixed adjacent stale docs in `docs/topics/services/p2p.md`: the route name (`/ws` → `/p2p-ws`), `s.StartHttp` → `StartHTTP` (which binds synchronously and serves in a goroutine), and the long-gone `gocore.Config().GetURL("asset_httpAddress")` reference (now `Asset.HTTPPublicAddress` with `Asset.HTTPAddress` fallback).

### Residual

Deliberately out of scope, from the issue's "Also" list:

- **Loopback default for `p2p_httpListenAddress`**: the asset service's centrifuge client connects to `/p2p-ws` cross-pod (`p2p_httpAddress.operator = peer.${clientName}.svc.cluster.local`), and there is no `p2p_httpListenAddress.operator` override - a loopback committed default would break operator/k8s deployments on upgrade. Needs a deployment decision plus per-context overrides.
- **CORS `*` / always-true `CheckOrigin`**: needs an operator-facing allow-list setting; flipping `CheckOrigin` also breaks the dashboard's cross-origin `wstest` page. Separate change.
- **`netutil.LimitListener` connection cap**: no sibling precedent and the cap value should be a setting; separate change.
- **Post-upgrade websocket hardening** (client cap, read deadlines/ping-pong on established conns): pre-existing gap tracked by the `/p2p-ws` hardening work (PR #1380); these `http.Server` timeouts cover the request/header/idle phases, not hijacked conns. An attacker can still park established websockets; that is unchanged by this PR and explicitly not claimed by it.
- **Asset service TLS timeouts are dead code** (`h.e.StartTLS` serves from `e.TLSServer`, but the timeouts are set on `e.Server`): separate pre-existing bug, deserves its own issue rather than being bundled here.

### Tests

Added `services/p2p/server_http_timeouts_test.go`:

- `TestSetupHTTPServerTimeouts` - regression guard asserting the four fields equal the named constants, plus sanity relationships (`ReadHeaderTimeout <= ReadTimeout`, `IdleTimeout >= ReadTimeout`), so a revert or a nonsense value fails the test.
- `TestHTTPServerClosesSlowHeaderClient` - slowloris repro: raw conn, partial headers, asserts the server closes the connection when `ReadHeaderTimeout` fires instead of holding it open.
- `TestWebsocketSurvivesHTTPServerTimeouts` - proves the risky half: with all timeouts shortened to 300ms, an established `/p2p-ws` stream still delivers a notification pushed 1.2s after upgrade.

Run (affected package only):

```
```

All green, including after rebasing onto latest upstream/main.

### DIFF
--- a/docs/topics/services/p2p.md
+++ b/docs/topics/services/p2p.md
@@ -169,7 +169,8 @@ The `Start` function is responsible for commencing the P2P service:
 2. **HTTP Endpoints**:
 
     - Sets up a health check endpoint (`/health`) that responds with "OK".
-    - Adds a WebSocket endpoint (`/ws`) for real-time communication via `s.HandleWebSocket`.
+    - Adds a WebSocket endpoint (`/p2p-ws`) for real-time communication via `s.HandleWebSocket`.
+    - Configures HTTP server timeouts (read header, read, write, idle) so slow or stalled clients cannot hold connections open indefinitely; established WebSocket connections are exempt because connection deadlines are cleared on upgrade.
 
 3. **Start HTTP Server**:
 
--- a/docs/topics/services/p2p.md
+++ b/docs/topics/services/p2p.md
@@ -148,7 +148,7 @@ The P2P server's `Init` function is in charge of server setup:
 
 2. **Asset HTTP Address Configuration**:
 
-    - Retrieves the Asset HTTP Address URL from the configuration using `gocore.Config().GetURL("asset_httpAddress")`.
+    - Retrieves the Asset HTTP Address URL from the settings, preferring `Asset.HTTPPublicAddress` and falling back to `Asset.HTTPAddress`.
 
 3. **Block Validation Client Initialization**:
 
@@ -170,11 +170,11 @@ The `Start` function is responsible for commencing the P2P service:
 
     - Sets up a health check endpoint (`/health`) that responds with "OK".
     - Adds a WebSocket endpoint (`/p2p-ws`) for real-time communication via `s.HandleWebSocket`.
-    - Configures HTTP server timeouts (read header, read, write, idle) so slow or stalled clients cannot hold connections open indefinitely; established WebSocket connections are exempt because connection deadlines are cleared on upgrade.
+    - Configures HTTP server timeouts (read header, read, write, idle) so slow or stalled clients cannot hold plain HTTP exchanges (or incomplete WebSocket upgrades) open indefinitely. Established WebSocket connections are exempt: connection deadlines are cleared on upgrade, so post-upgrade liveness is handled by separate WebSocket hardening, not these timeouts.
 
 3. **Start HTTP Server**:
 
-    - Initiates the HTTP server using a goroutine, which is executed via `s.StartHttp`.
+    - Starts the HTTP server via `s.StartHTTP`, which binds the listener synchronously (so configuration errors fail startup) and then serves in a goroutine.
 
 4. **PubSub Topics Setup**:
 
--- a/services/p2p/Server.go
+++ b/services/p2p/Server.go
@@ -682,21 +682,29 @@ func (s *Server) Init(ctx context.Context) (err error) {
 //
 // Returns an error if any component fails to start, or nil on successful startup.
 
+// HTTP server timeouts for the p2p HTTP surface. They bound every phase of a
+// plain HTTP exchange (e.g. /health) plus the request/header/idle phases of a
+// /p2p-ws connection that never completes its upgrade. They do NOT bound an
+// established /p2p-ws stream: net/http clears the connection deadlines on
+// Hijack (and gorilla/websocket clears them again on upgrade), so post-upgrade
+// liveness (read deadlines, ping/pong, connection caps) is separate websocket
+// hardening work.
+const (
+	httpReadHeaderTimeout = 10 * time.Second
+	httpReadTimeout       = 30 * time.Second
+	httpWriteTimeout      = 30 * time.Second
+	httpIdleTimeout       = 120 * time.Second
+)
+
 func (s *Server) setupHTTPServer() *echo.Echo {
 	e := echo.New()
 	e.HideBanner = true
 	e.HidePort = true
 
-	// Bound every phase of a plain HTTP exchange so slow or stalled clients
-	// cannot hold connections (and their goroutines/fds) forever. The
-	// long-lived /p2p-ws stream is unaffected: net/http clears the connection
-	// deadlines on Hijack (and gorilla/websocket clears them again on
-	// upgrade), so none of these timeouts apply once a websocket is
-	// established.
-	e.Server.ReadHeaderTimeout = 10 * time.Second
-	e.Server.ReadTimeout = 30 * time.Second
-	e.Server.WriteTimeout = 30 * time.Second
-	e.Server.IdleTimeout = 120 * time.Second
+	e.Server.ReadHeaderTimeout = httpReadHeaderTimeout
+	e.Server.ReadTimeout = httpReadTimeout
+	e.Server.WriteTimeout = httpWriteTimeout
+	e.Server.IdleTimeout = httpIdleTimeout
 
 	e.Use(middleware.Recover())
 
--- a/services/p2p/Server.go
+++ b/services/p2p/Server.go
@@ -687,6 +687,16 @@ func (s *Server) setupHTTPServer() *echo.Echo {
 	e.HideBanner = true
 	e.HidePort = true
 
+	// Bound the request/header phase and idle keep-alives so slow or stalled
+	// clients cannot hold connections (and their goroutines/fds) forever.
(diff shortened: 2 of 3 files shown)


ctnguyen (approved):
Review at commit [0426368](https://github.com/gokutheengineer/teranode/commit/0426368218bc97ff1299237aa61093b6f55f6fe4)

Nice, tight fix. This closes a real gap — the p2p echo server was the one sibling running a zero-value `http.Server` with all four timeouts unbounded on an all-interfaces bind, so slowloris and idle keep-alive fd/goroutine parking were genuinely open. The values (10/30/30/120s) sit right alongside what validator, asset, and blockpersister already hardcode, so this is consistent rather than a new pattern. The part I most appreciate is that you didn't just take the issue's word that `WriteTimeout` had to stay unset: you reasoned through the hijack path (net/http zeroes the conn deadline on `Hijack`, gorilla clears it again on upgrade) and then *proved* it with `TestWebsocketSurvivesHTTPServerTimeouts` rather than asserting it — that's the exact half of the change most likely to bite silently, and it's covered. The slowloris repro and the regression guard round it out, the diff stays minimal, and the residuals (loopback bind, CORS, connection cap, post-upgrade websocket hardening via #1380) are scoped out with real reasons instead of hand-waving. The stale `/ws` → `/p2p-ws` doc correction is a good catch too. Nothing blocking or non-blocking from me.

freemans13 (commented):
## Review: sound change, would merge

I went after the two claims this PR leans on, since both are the kind that pass tests while being wrong in production.

**The timeouts are genuinely live, on both serve paths.** Echo keeps two separate `http.Server` objects (`e.Server` and `e.TLSServer`), and setting fields on the wrong one is silent dead code. P2P gets this right: `setupHTTPServer()` is called once at `services/p2p/Server.go:693`, and `StartHTTP` serves via `s.e.Server.Serve(listener)` (`Server.go:1843`) for plain HTTP and `s.e.Server.ServeTLS(...)` (`Server.go:1846`) for TLS. It never calls `e.Start`/`e.StartTLS`, so `e.TLSServer` is never touched. Worth noting because the asset service falls into exactly this trap — see the follow-up note at the bottom.

**The websocket-safety argument holds, verified against the module versions in `go.mod`.** Go's `net/http` clears the connection deadlines in `hijackLocked` (`server.go:325`), and gorilla clears them again on upgrade (`server.go:265`). I also checked the other branch: if someone later sets a `HandshakeTimeout` on the shared upgrader at `HandleWebsocket.go:185`, gorilla re-clears the write deadline after the handshake write (`server.go:273-276`), so the connection ends up deadline-free either way. The reasoning in the description is right, and it's robust to that future change rather than true by accident.

Only two routes exist on this echo instance — `/health` (`Server.go:637`) and `/p2p-ws` (`Server.go:641`). No SSE, chunked, pprof or large-blob endpoint that a 30s `WriteTimeout` could cut off.

### Verification run

| Command | Result |
|---|---|
| `SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 -run 'TestSetupHTTPServerTimeouts\|TestHTTPServerClosesSlowHeaderClient\|TestWebsocketSurvivesHTTPServerTimeouts' ./services/p2p/...` | pass, 3.47s |
| same, `-count=5` | pass, 9.15s, no race reports |
| same, `GOMAXPROCS=1 -count=10` | pass, 16.42s |
| `go vet ./services/p2p/...` | clean |
| `golangci-lint run --build-tags testtxmetacache ./services/p2p/...` | only a pre-existing `prealloc` nit at `services/p2p/mock.go:76`; neither changed file flagged |

The `GOMAXPROCS=1` run was specifically to stress the 200ms/300ms timeouts and the 1200ms sleep. Margins are 4–25x against the 5s detection deadlines, so CI flake risk looks low.

### The one thing I'd push back on

**The comment and the doc sentence promise more than the change delivers.** Both say slow or stalled clients "cannot hold connections open indefinitely". That is now true for `/health`, but `/p2p-ws` — the only route worth attacking — is exempt by design, and there is nothing behind the upgrade to enforce liveness: no read pump, no `SetReadDeadline`, no ping/pong, no per-IP connection cap anywhere in the non-test p2p code. A client can open N websockets, never read, and hold N goroutines and N fds indefinitely; `handleClientMessages` blocks on `WriteMessage` once the socket buffer fills.

This is pre-existing rather than a regression, and the description does acknowledge it under Residual as belonging to the `/p2p-ws` hardening work in #1380. The gap is only that the in-code comment (`Server.go:619-624`) and `docs/topics/services/p2p.md:172` don't carry that caveat. Narrowing both to "plain HTTP exchanges" would make the claim match what the code buys.

### Smaller points

**Two of the three tests don't guard the production values.** `TestHTTPServerClosesSlowHeaderClient` and `TestWebsocketSurvivesHTTPServerTimeouts` both overwrite the timeouts immediately after `setupHTTPServer()`, so they would still pass with the production change reverted. That leaves `TestSetupHTTPServerTimeouts` as the sole regression guard, and it asserts only `NotZero` — `1ns` would satisfy it. The test comments are honest about the shortening, so this is a design nit, not a misleading test. Asserting a sane relationship (`ReadHeaderTimeout <= ReadTimeout`, `IdleTimeout >= ReadTimeout`) or the exact values against shared constants would close it cheaply. Flagging that I inferred this by reading rather than by actually reverting the change.

**Stale bullet directly under the docs edit.** `docs/topics/services/p2p.md:176-178` still says the server starts "via `s.StartHttp`". The method is `StartHTTP`, and it now binds synchronously and only serves in a goroutine — a distinction that looks deliberate per the comment at `Server.go:695`. Line 150 of the same file also still references `gocore.Config().GetURL("asset_httpAddress")`, which is long gone. Both are adjacent to what you're already touching.

**Duplication, for the record, not as a blocker.** These four values are byte-for-byte identical to propagation's (`services/propagation/Server.go:995-998`), and validator, asset, blockpersister, blob and the daemon each hardcode their own variants. There is no settings field for HTTP server timeouts anywhere in `settings/`. So this PR matches the codebase pattern rather than deviating from it; the copy-paste is the codebase's problem, not this change's.

### Separate follow-up worth filing

The asset service has the exact bug this PR avoids. `services/asset/httpimpl/http.go:628-631` sets the same four timeouts on `h.e.Server`, but starts HTTPS via `h.e.StartTLS(...)` at line 648, and echo's `startTLS` serves from `e.TLSServer` — a different `http.Server`. Asset's timeouts are therefore dead code whenever `SecurityLevelHTTP != 0`. Deserves its own issue; shouldn't be bundled here.

Also benign but worth knowing: `e.Server.ServeTLS` auto-enables HTTP/2, under which these timeouts become per-stream rather than connection-level (Go disarms the conn write deadline and re-applies `WriteTimeout`/`ReadTimeout` per stream). No new risk from this change. Gorilla can't upgrade an h2 request at all, so `/p2p-ws` over HTTPS is likely already broken — that part is inferred, I didn't stand up a TLS listener to confirm it.

---
🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

freemans13 (approved):

PR #1546, opened 2026-08-10
fix(postgres): provision teranode/miner roles as NOSUPERUSER

icellan:
## Finding

A third-party audit found that the Postgres database roles created by the deployment configs are granted `SUPERUSER` instead of `NOSUPERUSER` in every path that is actually wired into a running deployment:

- `deploy/kubernetes/postgres/postgres-configmap.yaml` (`teranode` role, k8s deployment)
- `deploy/docker/base/docker-services.yml` (`teranode` role, embedded init.sql heredoc)
- `compose/postgres/init.sql` (`miner1`/`miner2`/`miner3` roles, local compose stack)
- `test/postgres/init.sql` (`miner1`/`miner2`/`miner3` roles, test mirror of the above)

Notably, the provisioning scripts under `scripts/postgres/init-prod-*.sql` and `init-stage-*.sql` already use `NOSUPERUSER` correctly, and the `coinbase*` roles in the compose/test init.sql files were already `NOSUPERUSER` too — proving the application does not need superuser to function; these four files were just never brought in line.

## Fix

Switch `SUPERUSER` → `NOSUPERUSER` for the `teranode` and `miner1`/`miner2`/`miner3` roles in the four files above. Purely a role-privilege change, no functional impact — these roles only ever create and own their own database/schema.

## Note

A proper long-term fix would consolidate this role-creation logic, which is currently duplicated independently across roughly five different places (the four files above plus the `scripts/postgres/init-*.sql` variants) into a single source of truth. Out of scope for this minimal fix.

---

Security: closes a least-privilege gap flagged in a third-party audit.


ordishs (approved):
Verified the change. The pre-change problem is real: eight bare `SUPERUSER` attributes across five files, while the sibling coinbase roles and every `scripts/postgres/init-prod-*.sql` already use `NOSUPERUSER`. I traced the only superuser-dependent statement in the Go codebase (`sql.go:1219`) and confirmed the genesis insert does not need it, because a row with `id=0` and `parent_id=0` satisfies the self-referencing foreign key at `sql.go:603`. The failure of the pooled `SET` statement also removes a latent hazard, where a lost `origin` restore leaves a pooled connection with all foreign keys disabled.

**P2 — existing deployments keep `SUPERUSER`. `deploy/docker/base/docker-services.yml:19-22`, `deploy/kubernetes/postgres/postgres-configmap.yaml:6`**

The init SQL runs only when the PostgreSQL data directory is empty. This PR adds no `ALTER ROLE` migration.

`docker-services.yml:15` binds `${DATA_PATH}/postgres` to `/var/lib/postgresql/data`. `postgres.yaml:41-56` mounts the configmap at `/docker-entrypoint-initdb.d/init.sql` beside a 10Gi persistent volume claim. The `postgres:17` entrypoint skips `/docker-entrypoint-initdb.d` whenever PGDATA is already initialised. An operator who pulls this change and restarts keeps a `SUPERUSER` teranode role, so the audit finding stays open on every running node.

Recommendation: add a remediation note to the PR body or the deployment README — run `ALTER ROLE teranode NOSUPERUSER;` as the postgres superuser, and the same for `miner1-3` on the compose stacks.

ctnguyen (changes requested):
Review at commit [6fd5e0f](https://github.com/icellan/teranode/commit/6fd5e0f1641d2d0776578ecf64caecbaf9f6bdd7)

Nice, tight least-privilege fix. The four provisioning files that were out of line are now consistent with the `scripts/postgres/init-*.sql` set, and I checked the rest of the tree — there is no remaining bare `SUPERUSER` anywhere in the repo, so the sweep is complete on this side. I also like that you didn't stop at the one-line change: the template got a regression test, and the `ALTER ROLE` remediation for pre-existing volumes (raised by ordishs) is now documented for both Docker and Kubernetes with a verify step and an explicit "ownership is unaffected, you can leave the node running" reassurance, which is exactly the thing operators worry about. The Kubernetes remediation checks out against the actual manifests: `postgres` is a StatefulSet so `postgres-0` is right, `teranode-operator` is the namespace the install doc uses, and miners apply `deploy/kubernetes/postgres/` directly — so that path really is fixed by this PR. One gap on the Docker path below.

## Blocking issues

### ChiR1 — The miner-facing Docker deployment still provisions `teranode` as SUPERUSER

*Partially overlaps [@ordishs's point P2 (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1546#pullrequestreview-4907182797) — his point was about existing volumes; this is about brand-new installs.*

**Problem:** `docs/howto/miners/docker/minersSecurityBestPractices.md` states "New deployments provision it as `NOSUPERUSER`", and the whole `docs/howto/miners/docker/` set explicitly names [teranode-quickstart](https://github.com/bsv-blockchain/teranode-quickstart) as the supported Docker install path. Quickstart does not consume this repo's `deploy/docker/base/docker-services.yml` — it vendors its own copy of the init heredoc, which at quickstart HEAD `8ba35c7` (commit dated 12-06-2026, checked 13-08-2026) still reads:

```sql
-- teranode-quickstart, compose/docker-services.yml:19-23
CREATE ROLE teranode LOGIN
  PASSWORD 'teranode'
  SUPERUSER INHERIT NOCREATEDB NOCREATEROLE NOREPLICATION;
```

[Permalink to line 21](https://github.com/bsv-blockchain/teranode-quickstart/blob/8ba35c743b5b367e670403f3762a9bf99e9195d0/compose/docker-services.yml#L21). So a miner who clones quickstart and runs `setup.sh` tomorrow, on a brand-new empty data directory, still gets a superuser `teranode` role. The Kubernetes half of this PR does close the loop (miners `kubectl apply -f deploy/kubernetes/postgres/`); the Docker half only fixes the in-repo dev/CI stack.

**Why it matters:** The audit finding stays open on the primary production Docker path — not just for existing nodes (P2) but for every future install. Worse, the doc added here is the one place an operator would check, and it tells them the opposite: reading "new deployments provision it as NOSUPERUSER", a fresh quickstart installer will reasonably skip the `ALTER ROLE` remediation and stay superuser indefinitely.

**Fix:** Land the same one-line change in `compose/docker-services.yml` in the teranode-quickstart repo, and until that ships, don't let the doc overstate the current state. Something like:

```markdown
Teranode's PostgreSQL role (`teranode`) owns the `teranode` database and needs no
superuser rights. Deployments provisioned from this repository's Compose and
Kubernetes manifests create it as `NOSUPERUSER`. The `teranode-quickstart` stack
still creates it as `SUPERUSER`; until that is updated, apply the remediation
below regardless of when your node was provisioned.
```

Then drop that caveat in the follow-up that syncs quickstart.

## Non-blocking issues

### ChiR2 — The regression guard covers one of the five changed provisioning paths

**Problem:** `TestWriteAll_MinerRolesAreNotSuperuser` guards a path that genuinely was at risk — `init.sql.tmpl` is fixed in this PR and the test pins the rendered `postgres/init-multinode.sql` — but it stops there. None of the four files the audit actually flagged (`compose/postgres/init.sql`, `test/postgres/init.sql`, `deploy/docker/base/docker-services.yml`, `deploy/kubernetes/postgres/postgres-configmap.yaml`) nor `scripts/postgres/init-*.sql` are guarded, and by your own note in the PR body the role-creation logic is duplicated across ~8 places, which is precisely the condition under which one copy drifts back. Separately, the regex is case-sensitive; `SUPERUSER` is a keyword, so `create role x login superuser;` is valid SQL and would sail past the check while the comment above it claims "No bare SUPERUSER attribute anywhere".

**Why it matters:** The guard covers the one file nobody edits by hand (it's generated) and leaves the four hand-maintained ones — the exact files that drifted and triggered the audit — unprotected, while reading as though the whole class is covered. The next person adding a Postgres provisioning file gets no signal at all.

**Fix:** Add a second test that sweeps every provisioning file, and use `\b` so case folding is safe. This test package sits at `compose/cmd/gennodes`, so the repo root is three levels up — or drop the test wherever you prefer and adjust the prefix:

```go
func TestProvisioningFilesAreNotSuperuser(t *testing.T) {
    // Test binaries run in their own package dir; compose/cmd/gennodes -> repo root.
    repoRoot := filepath.Join("..", "..", "..")

    // (?i) + \b: "superuser" is a keyword and may be written in any case;
    // \b will not match inside "NOSUPERUSER" since O and S are both word chars.
    bareSuperuser := regexp.MustCompile(`(?i)\bsuperuser\b`)
    for _, f := range []string{
        "compose/postgres/init.sql",
        "test/postgres/init.sql",
        "deploy/docker/base/docker-services.yml",
        "deploy/kubernetes/postgres/postgres-configmap.yaml",
        "compose/cmd/gennodes/templates/init.sql.tmpl",
    } {
        b, err := os.ReadFile(filepath.Join(repoRoot, f))
        require.NoError(t, err, "%s", f)
        require.NotRegexp(t, bareSuperuser, string(b), "%s grants SUPERUSER", f)
    }
}
```

A `filepath.Glob` over `scripts/postgres/init-*.sql` in the same loop would cover the remaining copies without having to maintain the list by hand.

While you're there, `compose/postgres/init.sql` and `test/postgres/init.sql` are byte-identical; a `require.Equal` on their contents is a two-line drift guard that would have caught this class of bug in the first place.

### ChiR3 — `docker compose down -v` does not reset the stacks these files actually serve

**Problem:** The new header in `compose/postgres/init.sql` and `test/postgres/init.sql` says "Either recreate it (`docker compose down -v`) or ... run `ALTER ROLE`". That works for `compose/docker-compose-ss.yml` (named volume `postgres-data`) and `test/docker-compose.e2etest.yml` (named volume `data-postgres`), but two of the consumers bind-mount the data directory from the host, where `-v` is a no-op:

- `compose/docker-compose-chainintegrity.yml:45` → `../data/postgres:/var/lib/postgresql/data`, mounting `compose/postgres/init.sql`
- `test/docker-compose-host.yml:55` → `../data/postgres:/var/lib/postgresql`, mounting `test/postgres/init.sql`

**Why it matters:** Someone following the comment on those stacks runs `down -v`, sees the stack come back up, and assumes the roles were recreated — the old superuser roles are still there on disk. A silently ineffective remediation is worse than none.

**Fix:** Cover both storage shapes in the comment:

```sql
-- up before the NOSUPERUSER change keeps its superuser roles on the existing
-- volume. Recreate the storage ('docker compose down -v' for named volumes,
-- or delete the bind-mounted host directory, e.g. 'data/postgres'), or, as the
-- postgres superuser, run 'ALTER ROLE <role> NOSUPERUSER;' for each role below.
```

### ChiR4 — `SET session_replication_role` in the blockchain store is now a permanently-failing silent no-op

**Problem:** `stores/blockchain/sql/sql.go:1219` and `:1233` run `SET session_replication_role = 'replica'/'origin'` around the genesis insert. That GUC is superuser-only, so after this change every Postgres deployment gets `permission denied to set parameter "session_replication_role"` — and the result is discarded via `_, _ =`, so nothing is logged. The insert itself is fine: genesis goes in with `id = 0` and `parent_id = previousBlockID = 0`, and PostgreSQL accepts a self-referencing row against `parent_id BIGSERIAL REFERENCES blocks(id)`, which matches ordishs' verification. So this is not a functional regression — but it does leave code that claims to disable FK checks and can no longer ever do so.

**Why it matters:** The next person who needs a FK-bypassed write will find this branch, assume it works, and get a silent failure. It also hides a real latent bug that this PR incidentally defuses: `SET` is per-session, and `s.db` is a pool, so under the old superuser roles the `'replica'` and `'origin'` statements could land on different connections — leaving a pooled connection with FK and trigger enforcement disabled for every subsequent write on it.

**Fix:** Drop the Postgres branch and record why it isn't needed:

```go
// turn off foreign key checks when inserting the genesis block
if s.engine == util.Sqlite || s.engine == util.SqliteMemory {
    _, _ = s.db.Exec("PRAGMA foreign_keys = OFF")
}
// Postgres needs no equivalent: genesis is inserted with id=0 and parent_id=0,
// and a self-referencing row satisfies the blocks(parent_id) -> blocks(id) FK.
// (session_replication_role is superuser-only and the role is NOSUPERUSER.)
```

If you'd rather keep it, at least surface the error with `s.logger.Warnf` instead of swallowing it.

## Recap

| --- | --- | --- | --- |
| ChiR1 | Quickstart still SUPERUSER; doc says otherwise | YES | 75% |
| ChiR2 | Guard misses four provisioning files | | 40% |
| ChiR3 | `down -v` no-op on bind-mounted stacks | | 25% |
| ChiR4 | Dead `session_replication_role` call | | 15% |

icellan:
Addressed ctnguyen's review in 89538f5:

**ChiR1 (blocking)** — Softened the doc claim in `minersSecurityBestPractices.md`. It now says deployments provisioned from this repo's Compose/Kubernetes manifests create the role NOSUPERUSER, and explicitly calls out that `teranode-quickstart` still creates it SUPERUSER, so the remediation applies regardless of provisioning path. Not fixing quickstart itself here — that's a separate repo/PR.

**ChiR2** — Added `TestProvisioningFilesAreNotSuperuser`, sweeping all five hand-maintained files (`compose/postgres/init.sql`, `test/postgres/init.sql`, `deploy/docker/base/docker-services.yml`, `deploy/kubernetes/postgres/postgres-configmap.yaml`, `compose/cmd/gennodes/templates/init.sql.tmpl`) plus a glob over `scripts/postgres/init-*.sql`, on top of the existing generated-template test. Regex is case-insensitive with a word boundary, and skips comment lines so it doesn't false-positive on prose like "needs no superuser rights". Also added `TestDevAndTestPostgresInitAreIdentical` to catch `compose/postgres/init.sql` and `test/postgres/init.sql` drifting apart, since that's the exact failure mode that caused the original audit finding.

**ChiR3** — Extended the `down -v` remediation comment in both init.sql files to also cover bind-mounted data directories (`compose/docker-compose-chainintegrity.yml`, `test/docker-compose-host.yml`), where `-v` is a no-op.

**ChiR4** — Dropped the Postgres branch of the `session_replication_role` toggle in `stores/blockchain/sql/sql.go` around the genesis insert. It was a silent no-op under NOSUPERUSER (that GUC is superuser-only) and isn't needed anyway since the genesis row is self-referencing (`id=0`, `parent_id=0`) and satisfies the FK without disabling it. Replaced with a comment explaining why.

Ran `go build`, `go vet`, and `go test -race` on `./stores/blockchain/sql/...` and `./compose/...`. All pass except the Postgres-testcontainer tests, which can't run in my environment (no Docker daemon available) — unrelated to these changes; the sqlite-backed genesis-insert path exercised by the ChiR4 change passed cleanly.

ctnguyen (approved):
Review at commit [89538f5](https://github.com/icellan/teranode/commit/89538f545931777290f2948e9109f04ef8b05954)

This round is a solid response to the last one. The doc no longer promises something the quickstart path doesn't deliver, the regression guard grew from one generated file to every hand-maintained provisioning file plus a glob over `scripts/postgres/init-*.sql`, the drift guard between the two `init.sql` copies is in (and they are byte-identical right now), and the `session_replication_role` toggle is gone with a comment that actually explains why Postgres never needed it. I re-checked the genesis path behind that comment — `StoreBlock.go:776` sets `previousBlockID = 0` for genesis, `StoreBlock.go:414-447` inserts with an explicit `id = 0`, and Postgres validates `blocks(parent_id) -> blocks(id)` (`sql.go:603`) after the row exists, so the self-reference holds and nothing is being bypassed. I also swept the tree again: no bare `SUPERUSER` survives in any provisioning file, and no other statement in the Go code needs superuser (the `ALTER SYSTEM` calls are all commented out). Two previous points have a leftover sliver each, and two new things are worth a look — none of it blocking.

## Status of previous points

Two caveats on the table below.

**ChiR2** — the sweep at `compose/cmd/gennodes/gen_test.go:124-162` covers everything the point named, and the glob keeps `scripts/postgres/` maintenance-free. Two small pieces of that point remain, and both stay under ChiR2 rather than becoming new items: `scripts/postgres/init_dev.sh:3` also creates the `teranode` role and is the one role-creating file in the repo the sweep does not see (the glob is `init-*.sql`, the file is `init_dev.sh`), so the test's own comment "sweeps every hand-maintained Postgres provisioning file" is slightly ahead of what it does. And the case-sensitivity half of the point was fixed only in the new test — `gen_test.go:107` still uses `(?:^|[^A-Z])SUPERUSER` on the rendered output. In practice the template is now covered case-insensitively by the sweep, so this is cosmetic consistency, not a hole.

**ChiR3** — the `init.sql` headers are fixed: `compose/postgres/init.sql:5-10` and `test/postgres/init.sql:5-10` now name both storage shapes, which is exactly right for `compose/docker-compose-chainintegrity.yml:46` (`../data/postgres`) and `test/docker-compose-host.yml:55`. The same advice in the docs did not get the same treatment: `docs/howto/miners/docker/minersSecurityBestPractices.md:96-100` still frames the dev/test stacks as "the `postgres-data` volume was kept (`docker compose down` without `-v`)", which describes only `compose/docker-compose-ss.yml:93`. Someone on the chainintegrity or host stack reading that paragraph will run `down -v` and believe the roles were recreated. Same sentence also tells operators to `ALTER ROLE` the `coinbase*` roles, which were already `NOSUPERUSER` before this PR — harmless, but it muddies which roles actually need the fix. Worth one edit to that paragraph; still under ChiR3.

On ChiR1: the doc claim is fixed for this repo, but the one-line change in `teranode-quickstart`'s `compose/docker-services.yml` is still owed as a follow-up — that's the PR that gets to delete the caveat paragraph again.

| --- | --- | --- | --- |
| ChiR1 | Quickstart SUPERUSER vs doc claim | YES | FIXED |
| ChiR2 | Guard misses provisioning files | | PARTIAL FIXED |
| ChiR3 | `down -v` no-op on bind mounts | | PARTIAL FIXED |
| ChiR4 | Dead `session_replication_role` | | FIXED |

## Non-blocking issues

### ChiR5 — Nothing exercises the store as a NOSUPERUSER role

**Problem:** The PR's premise is "the application does not need superuser to function", and 89538f5 now leans on that harder than before: with the Postgres branch of the FK toggle gone, a fresh Postgres deployment reaches `insertGenesisTransaction` (`sql.go:1215-1236`) with no privilege escape hatch at all. Nothing in the test suite checks that. The Postgres-backed store tests go through `test/utils/postgres/container.go`, which starts the container with `dbUser = "postgres"` (`container.go:105-109`) and creates each per-test database as that role, so `New()` at e.g. `stores/blockchain/sql/FindBlocksContainingSubtree_test.go:284` always runs as a superuser. The SQLite-backed tests never touch the Postgres path. So the schema creation, the `ALTER TABLE blocks SET (autovacuum_enabled = ...)` at `sql.go:248` (owner-only, fine in theory), and the genesis insert are all verified only as superuser.

**Why it matters:** The evidence that NOSUPERUSER is sufficient is good — `scripts/postgres/init-prod-mainnet.sql:2-10` has run that way in production, and the `coinbase*` roles always did — so this is not a "the code is broken" concern. It's that the failure mode if the premise ever stops holding is "node cannot start against a freshly provisioned database", discovered by an operator rather than by CI, and the change that makes the premise load-bearing landed in this commit with only manual reasoning behind it. Given the sweep tests now pin the *config*, pinning the *behavior* is the cheap other half.

**Fix:** In one Postgres-backed store test, hand ownership to a NOSUPERUSER role and connect as it, so the genesis insert and schema creation are proven without superuser:

```go
// after SetupTestPostgresContainer(), using an admin session:
//   CREATE ROLE teranode_test LOGIN PASSWORD 'teranode_test'
//     NOSUPERUSER INHERIT NOCREATEDB NOCREATEROLE NOREPLICATION;
//   ALTER DATABASE <dbName> OWNER TO teranode_test;
storeURL, err := url.Parse(connStr)
require.NoError(t, err)
storeURL.User = url.UserPassword("teranode_test", "teranode_test")

// The genesis insert must succeed with no session_replication_role escape hatch.
s, err := New(ulogger.TestLogger{}, storeURL, tSettings)
require.NoError(t, err)
```

### ChiR6 — The SQLite half of the FK toggle keeps the pooled-connection hazard the Postgres half just shed

*Same as [@ordishs's point "leaves a pooled connection with all foreign keys disabled" (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1546#pullrequestreview-4907182797).*

**Problem:** His observation, and my ChiR4, were both scoped to the Postgres branch, which is now gone — this is about the SQLite branch that remains, which neither of us asked to change. `sql.go:1216-1217` and `sql.go:1234-1235` run `PRAGMA foreign_keys = OFF` / `ON` on `s.db`, which is a pool of up to 5 connections (`util/sql.go:267`). That is the precise pattern `util/sql.go:249-253` warns against in its own words — "a post-open `db.Exec("PRAGMA foreign_keys = ON")` would only affect a single pooled connection" — which is why FK enforcement is set per-connection via the DSN (`util/sql.go:218`, `util/sql.go:237`) instead. Both `Exec` results are discarded, and there is a second hole: if `StoreBlock` fails, `sql.go:1226-1229` returns before the re-enable, leaving whichever connection ran the `OFF` back in the pool with FK enforcement disabled. Separately, the new comment at `sql.go:1219-1224` explains Postgres well but leaves the asymmetry unexplained — if a self-referencing row satisfies the constraint, the natural question is why SQLite still needs the bypass, and the next reader has to work that out from scratch.

**Why it matters:** In practice it works today because the pool is idle at startup and the three statements reuse the same connection, so this is fragility rather than a live bug. But it is fragility in the FK-enforcement path of the store, guarded by discarded errors, in a block this commit rewrote — and the error path genuinely can hand a FK-disabled connection back to the pool. Anyone extending this later (a second FK-bypassed write, or a startup that stores more than genesis) inherits both problems silently.

**Fix:** Make the re-enable unconditional and say why SQLite is the odd one out:

```go
// SQLite checks this FK immediately, so the self-referencing genesis row
// (id=0, parent_id=0) needs the check suspended; Postgres validates after
// the row exists and needs no equivalent. PRAGMA is per-connection and s.db
// is a pool, so keep the window as small as possible and always restore.
if s.engine == util.Sqlite || s.engine == util.SqliteMemory {
    if _, pragmaErr := s.db.Exec("PRAGMA foreign_keys = OFF"); pragmaErr != nil {
        logger.Warnf("could not disable foreign keys for genesis insert: %v", pragmaErr)
    }
    defer func() {
        if _, pragmaErr := s.db.Exec("PRAGMA foreign_keys = ON"); pragmaErr != nil {
            logger.Warnf("could not re-enable foreign keys after genesis insert: %v", pragmaErr)
        }
    }()
}
```

Wrapping the insert in a single `*sql.Conn` (`s.db.Conn(ctx)`) would close the pool question entirely, if you want the stronger version.

## Recap

| --- | --- | --- | --- |
| ChiR5 | No NOSUPERUSER test coverage | | 40% |
| ChiR6 | SQLite FK pragma on pooled conns | | 30% |

PR #1548, opened 2026-08-10
Remove stray grpcIngress from asset service in sample manifests

icellan:
## Summary

The asset service serves HTTP and WebSocket only — it runs no gRPC server. Both sample CRs (`teranode-cr.yaml` and `teranode-cr-mainnet.yaml`) had a `grpcIngress` block under `asset.spec`, which wrongly suggests the service exposes gRPC.

## Why removal instead of a comment

Checked whether `grpcIngress` is structurally required by the operator's CRD schema before touching it. This repo doesn't vendor the operator's CRD/schema — it only references the operator Helm chart by tag (`teranode-operator.yaml`). Within these same sample files, several other services with no gRPC endpoint (`blockAssembly`, `blockValidator`, `blockchain`, `rpc`, `subtreeValidator`, `alertSystem`) simply omit any ingress field and are valid entries, which indicates `grpcIngress` is an optional, per-service field rather than something the schema mandates unconditionally. That made outright removal the safer, less confusing fix rather than leaving a misleading field with an explanatory comment.

Other `grpcIngress` blocks in the same files (e.g. under `peer`, `propagation`) are left untouched — those services do run gRPC.

## Test plan

- [x] `git diff` reviewed — only the two erroneous `grpcIngress` blocks removed, nothing else touched
- [ ] Manifests validated against a running operator (not available in this environment)

### DIFF
--- a/deploy/kubernetes/teranode/teranode-cr-mainnet.yaml
+++ b/deploy/kubernetes/teranode/teranode-cr-mainnet.yaml
@@ -20,6 +20,10 @@ spec:
   asset:
     enabled: true
     spec:
+      # No grpcIngress here: the asset service serves HTTP and WebSocket only
+      # and starts no gRPC server. Omitting the field is not a statement about
+      # the other services — several of them do run gRPC listeners and simply
+      # keep them cluster-internal.
       deploymentOverrides:
         replicas: 1
         resources:
--- a/deploy/kubernetes/teranode/teranode-cr-mainnet.yaml
+++ b/deploy/kubernetes/teranode/teranode-cr-mainnet.yaml
@@ -26,11 +26,6 @@ spec:
           requests:
             cpu: 1
             memory: 1Gi
-      grpcIngress:
-        annotations:
-          traefik.ingress.kubernetes.io/router.entrypoints: ast-grpc
-        className: traefik-internal
-        host: mainnet-eu-1.example.com
       httpIngress:
         annotations:
           traefik.ingress.kubernetes.io/router.entrypoints: web
--- a/deploy/kubernetes/teranode/teranode-cr.yaml
+++ b/deploy/kubernetes/teranode/teranode-cr.yaml
@@ -17,6 +17,10 @@ spec:
   asset:
     enabled: true
     spec:
+      # No grpcIngress here: the asset service serves HTTP and WebSocket only
+      # and starts no gRPC server. Omitting the field is not a statement about
+      # the other services — several of them do run gRPC listeners and simply
+      # keep them cluster-internal.
       deploymentOverrides:
         imagePullPolicy: Never
         replicas: 1
--- a/deploy/kubernetes/teranode/teranode-cr.yaml
+++ b/deploy/kubernetes/teranode/teranode-cr.yaml
@@ -24,11 +24,6 @@ spec:
           requests:
             cpu: 100m
             memory: 256Mi
-      grpcIngress:
-        annotations:
-          traefik.ingress.kubernetes.io/router.entrypoints: ast-grpc
-        className: traefik-internal
-        host: teranode.local
       httpIngress:
         annotations:
           traefik.ingress.kubernetes.io/router.entrypoints: web


ordishs (approved):
Verified. `services/asset/` contains no `grpc.NewServer` and no `StartGRPCServer` call, `services/asset/asset_api/asset_api.proto` declares zero `service` blocks, and `settings.conf` defines only `ASSET_HTTP_PORT`. The removed `grpcIngress` pointed at a port the asset service never serves.

Two nits:

- The PR body says `blockAssembly`, `blockValidator`, `blockchain`, `rpc`, `subtreeValidator` and `alertSystem` have "no gRPC endpoint". That is false for five of the six — they all declare gRPC services. Only `rpc` is correct, and for a different reason: it serves JSON-RPC over HTTP on `TERANODE_RPC_PORT` with no proto service.
- `deploy/kubernetes/teranode/teranode-cr.yaml:27` — the claim that `grpcIngress` is optional in the operator CRD cannot be verified inside this repository. `teranode-operator.yaml` pins only the operator image tag `v0.2.8`.

ctnguyen (approved):
Review at commit [b4d392c](https://github.com/icellan/teranode/commit/b4d392c2c5adec2b96c2ac34c5d91bfac9e1ff77)

The core change is right, and I checked it rather than taking it on trust: `services/asset/` contains no gRPC server at all (`Server.Start` only brings up the echo HTTP server and the Centrifuge WebSocket server, `Server.go:256-290`), `services/asset/asset_api/asset_api.proto` declares message types with no `service` block, and the only port the service binds is `ASSET_HTTP_PORT`. So the removed `grpcIngress` blocks were routing an entrypoint at something that never listened. Nice catch, and I like that you left `peer` and `propagation` alone instead of doing a blanket sweep. Nothing here blocks merge. What is left is the paperwork around the change rather than the YAML itself — the PR description still carries a factual error another reviewer already flagged, and the same "asset speaks gRPC" myth survives in a few other places that are more likely to bite an operator than a sample manifest was.

## Non-blocking issues

### ChiR1 — PR description still claims five gRPC services have "no gRPC endpoint"

*Same as [@ordishs's point 1 (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1548#pullrequestreview-4907188404).*

**Problem:** The "Why removal instead of a comment" section justifies the change by asserting that `blockAssembly`, `blockValidator`, `blockchain`, `rpc`, `subtreeValidator` and `alertSystem` have "no gRPC endpoint". Five of those six do:

```
services/alert/alert_api/alert_api.proto:12                              service AlertAPI
services/blockassembly/blockassembly_api/blockassembly_api.proto:16      service BlockAssemblyAPI
services/blockchain/blockchain_api/blockchain_api.proto:13               service BlockchainAPI
services/blockchain/blockchain_api/blockchain_api.proto:673              service PeerRegistryService
services/blockvalidation/blockvalidation_api/blockvalidation_api.proto:9 service BlockValidationAPI
services/subtreevalidation/subtreevalidation_api/subtreevalidation_api.proto:11 service SubtreeValidationAPI
```

Only `rpc` is correct, and for a different reason than the one implied — it serves JSON-RPC over HTTP on `TERANODE_RPC_PORT` and has no proto `service` at all. Extending the original nit: the description was never edited, and the comment added in `b4d392c` only hedges around it ("several of them do run gRPC listeners") without withdrawing the claim, so the two now contradict each other.

**Why it matters:** The entire value of this PR is removing a misleading statement about which services expose gRPC, so leaving a *more* misleading statement in its own justification is self-defeating. It is also the kind of claim someone will cite later when deciding whether those five services need ingress of their own — and the PR description is where anyone reading back through this change will look first.

**Fix:** Rewrite that paragraph. Something like: "The other services in these samples — `blockAssembly`, `blockValidator`, `blockchain`, `subtreeValidator`, `alertSystem` — do run gRPC servers; they simply keep them cluster-internal and so need no ingress. `rpc` serves JSON-RPC over HTTP and declares no proto service. `asset` is the only one of these that runs no gRPC listener whatsoever." Then drop the "several other services omit it, therefore it's optional" inference entirely and replace it with the evidence in ChiR2.

### ChiR2 — `grpcIngress` optionality is verifiable after all; cite the CRD instead of inferring it

*Same as [@ordishs's point 2 (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1548#pullrequestreview-4907188404).*

**Problem:** The description argues that `grpcIngress` must be optional because other services in the same file omit it. That is an inference from sample data, not evidence about the schema — a field can be required for one service's spec and absent from another's. Extending the original nit, which stopped at "cannot be verified inside this repository": it *can* be verified outside it. The operator repo is public, and `api/v1alpha1/asset_types.go` declares the field on `AssetSpec` as an optional pointer with `omitempty` (verified against the operator repo at 60dd7fcd):

```go
GrpcIngress *IngressDef `json:"grpcIngress,omitempty"`
```

So removing the block is structurally valid and the CRs still validate. The conclusion was right; only the reasoning was weak.

**Why it matters:** This closes the one genuinely open question on the PR — whether the manifests still apply cleanly — with a citation rather than an assumption, and it stops the "other entries omit it" pattern from being reused as a justification for a field where it does not hold.

**Fix:** Replace the inference in the description with the permalink: <https://github.com/bsv-blockchain/teranode-operator/blob/60dd7fcdcd6c/api/v1alpha1/asset_types.go>. Worth noting the operator version the samples target too, since `teranode-operator.yaml` pins `v0.2.8`.

### ChiR3 — the same stray asset-gRPC references survive elsewhere in the repo (follow-up, not for this PR)

**Problem:** Out of scope for a manifest-only diff, so flagging it for a separate change rather than asking you to widen this one. The manifests were arguably the least harmful instance of this myth. Still present:

- `docs/references/services/pruner_reference.md:407` tells operators to verify the UTXO store connection with `grpcurl -plaintext localhost:8090 asset.AssetAPI/HealthGRPC`. There is no `AssetAPI` gRPC service, no `HealthGRPC` method, and 8090 is `ASSET_HTTP_PORT` — the command cannot succeed.
- `docs/howto/bugReporting.md:100-101` and `.github/ISSUE_TEMPLATE/bug_report.md:57-58` show `asset_grpcAddress=` and `asset_grpcListenAddress=:8091` in the sample settings dump. Neither key exists in `settings.conf` or `settings/asset_settings.go`.
- `docs/staging/operations/ClusterManagementAndOperations.md:58` cites `asset-grpc-ingress.yaml` as an example of "ingress rules for gRPC assets".

**Why it matters:** An operator following the pruner troubleshooting steps will paste that `grpcurl` line, get a connection-level failure against an HTTP port, and reasonably conclude the store is broken. That is a worse outcome than an unused ingress. The bug-report template is milder but it teaches every new reporter that these settings exist.

**Fix:** In a follow-up, swap the pruner health check for the endpoint that actually exists (`GET /health` is registered on the asset HTTP server at `services/asset/httpimpl/http.go:307`):

```bash
curl -s localhost:8090/health
```

and delete the two obsolete keys from the bug-report samples. Keep this diff to the manifests; a line in the description noting the follow-up is enough so it does not get lost.

### ChiR4 — the manifest comment answers a reviewer rather than an operator

**Problem:** Two of the comment's four lines explain what the omission does *not* imply about other services. That is review correspondence baked into a shipped sample. It is also duplicated verbatim across both CRs, sits above `deploymentOverrides` rather than where the field it describes used to live (between `deploymentOverrides` and `httpIngress`), and is the only "this field is deliberately absent" note in files where six other services also omit `grpcIngress` with no comment at all.

**Why it matters:** These CRs are copy-paste templates — every line ends up in users' clusters. Negative-space commentary that references a discussion nobody can see ages badly, and duplicating it in two files means two places to update if asset ever grows a gRPC listener.

**Fix:** Trim to a single factual line, placed where the field would have been:

```yaml
  asset:
    enabled: true
    spec:
      deploymentOverrides:
        imagePullPolicy: Never
        replicas: 1
        resources:
          requests:
            cpu: 100m
            memory: 256Mi
      # asset serves HTTP and WebSocket only — it starts no gRPC server, so no grpcIngress
      httpIngress:
```

## Recap

| --- | --- | --- | --- |
| ChiR1 | False gRPC claim in description | | 55% |
| ChiR2 | Cite operator CRD for optionality | | 30% |
| ChiR3 | Stray asset-gRPC docs remain | | 25% |
| ChiR4 | Comment is reviewer commentary | | 15% |

PR #1549, opened 2026-08-10
docs(grafana): document docker-base dashboards and cross-stack gaps

icellan:
## Summary
- Adds a README next to the docker-base Grafana dashboards (`deploy/docker/base/grafana_dashboards/`), mirroring the style of the existing compose stack README.
- Names "Teranode Overview" as the primary day-to-day operational dashboard for this stack.
- Documents which dashboards ship in the docker-base stack vs the compose stack, so operators of either stack can see the gap and copy JSON across manually if they want a dashboard from the other stack.

No dashboard JSON files are changed or unified between stacks — that stays out of scope for this doc-only change.

## Test plan
- [x] Doc-only change; no code affected.
- [x] Verified paths and dashboard titles against the JSON files in both `deploy/docker/base/grafana_dashboards/` and `compose/grafana/dashboards/`.

### DIFF
--- a/compose/grafana/dashboards/blockassembly-state.json
+++ b/compose/grafana/dashboards/blockassembly-state.json
@@ -25,7 +25,7 @@
     {
       "datasource": {
         "type": "prometheus",
-        "uid": "prometheus"
+        "uid": "${prometheus}"
       },
       "fieldConfig": {
         "defaults": {
@@ -113,7 +113,7 @@
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "prometheus"
+            "uid": "${prometheus}"
           },
           "editorMode": "code",
           "expr": "teranode_blockassembly_current_state",
@@ -129,7 +129,7 @@
     {
       "datasource": {
         "type": "prometheus",
-        "uid": "prometheus"
+        "uid": "${prometheus}"
       },
       "fieldConfig": {
         "defaults": {
@@ -203,7 +203,7 @@
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "prometheus"
+            "uid": "${prometheus}"
           },
           "editorMode": "code",
           "expr": "histogram_quantile(0.5, sum(rate(teranode_blockassembly_state_duration_seconds_bucket[5m])) by (state, le))",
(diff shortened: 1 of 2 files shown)


ordishs (approved):
The gap is real: `deploy/docker/base/grafana_dashboards/` had no README, while `compose/grafana/dashboards/` has one. For a docs-only PR the name accuracy is the whole job, and four of the five dashboard names do not match a provisioned dashboard title. An operator cannot find any of them in Grafana.

**P2 — "Dispatcher Batch Queue" is a panel title, not a dashboard title. `deploy/docker/base/grafana_dashboards/README.md:16`**

The `.title` of `teranode/teranode-batch-index-dashboard.json` is `Aerospike Batch Index Bottleneck Diagnostic (1024 Buffers)`, uid `stwh6cx`. "Dispatcher Batch Queue" is the first of its six panels.

**P2 — the two Aerospike names are also wrong. `deploy/docker/base/grafana_dashboards/README.md:17`**

The table says "Aerospike Latency" and "Aerospike Namespace". The real `.title` values are `Latency View` (uid `ZoeGW1DBk`) and `Namespace View` (uid `zGcUKcDZz2`).

**P2 — the compose dashboard name disagrees with the compose README. `deploy/docker/base/grafana_dashboards/README.md:23`**

This README says "BlockAssembler State Timeline". The `.title` of `compose/grafana/dashboards/blockassembly-state.json` is `BlockAssembler State Monitoring`, which is also what the compose README heading uses. "State Timeline" is only its first panel.

**P2 — the scope column mislabels an Aerospike dashboard as a Teranode queue. `deploy/docker/base/grafana_dashboards/README.md:16`**

The scope reads "Batch/queue internals". All six panels query `aerospike_node_stats_batch_index_*` filtered by `cluster_name="$cluster"`. An operator who opens it to debug a Teranode dispatcher backlog sees empty panels unless an Aerospike exporter is scraped.

**P2 — the cross-stack copy advice does not work for the dashboard it names. `deploy/docker/base/grafana_dashboards/README.md:27`**

The closing sentence says the panels are "not tied to a specific deployment", then advises a manual copy. Nine of `blockassembly-state.json`'s ten panels pin `datasource.uid` `prometheus`, the tenth uses `${prometheus}`, and its `templating.list` is empty. No datasource in `compose`, `deploy` or `test` declares `uid: prometheus`, so Grafana assigns a generated uid. Every panel reports a missing datasource and the operator has no picker to fix it. The pinned uid is a pre-existing defect in that JSON, not introduced here.

Recommendation: copy the exact `.title` values from the JSON files, or add a second column with the real title. Then drop or qualify the "not tied to a specific deployment" clause.

freemans13 (approved):

PR #1550, opened 2026-08-10
docs(plans): stop citing a design-plan file not present in the repo

icellan:
## Summary

`plans/subtree-lift-audit-notes.md` cited a companion design-plan file, `plans/subtree-lift.md`, at several points. That file was never committed to the repository, leaving dangling references for anyone reading the audit notes.

Rephrases each citation to note that the design plan is not preserved in this repository, instead of pointing at a file that doesn't exist. No design content was fabricated to fill the gap.

## Test plan

- [x] Confirmed `plans/subtree-lift.md` does not exist anywhere in the repo or its git history (`find . -name "subtree-lift.md"`, `git log --all --full-history -- plans/subtree-lift.md`)
- [x] `grep -n "subtree-lift.md" plans/subtree-lift-audit-notes.md` returns no matches after the edit
- [x] Doc-only change, no code affected

### DIFF
--- a/plans/subtree-lift-audit-notes.md
+++ b/plans/subtree-lift-audit-notes.md
@@ -1,11 +1,13 @@
 # Subtree-Lift Activation Strategy — Audit Notes
 
-This file documents the audit performed for Task 0 of `plans/subtree-lift.md`. It records
+This file documents the audit performed for Task 0 of the subtree-lift design work
+(the original design plan is not preserved in this repository). It records
 where merkle-root computation lives on the assembly side today, whether incomplete final
 subtrees ever reach finalised (mined-and-persisted) blocks, and a recommended activation
 strategy for the upcoming `RootHashPadded` lift.
 
 Source versions inspected:
+
 - `github.com/bsv-blockchain/go-subtree` v1.2.0 (from `go.mod`)
 - Teranode at HEAD (commit `efaeedf9c`)
 
@@ -173,7 +175,8 @@ header was computed against that same un-padded root by `createMerkleTreeFromSub
 
 This implies that under the planned `RootHashPadded(targetHeight)` change, the validator
 needs an explicit notion of the "intended" subtree height to lift to. The new validation
-rules in `plans/subtree-lift.md` use `targetHeight = subtrees[0].Height`, which is fine
+rules proposed in the design plan (not preserved in this repository) use
+`targetHeight = subtrees[0].Height`, which is fine
 **only when there are >1 subtrees** (since the first one defines the canonical full
 height). For the single-incomplete-subtree case — which is the only one that occurs
 today — there is no peer subtree to compare against; `len(hashes) == 1` falls into the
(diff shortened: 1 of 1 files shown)


ordishs (approved):
Verified. `plans/subtree-lift.md` appears in no ref of this repository, so all six citations were dead, and the patch removes every one. I also checked two things the body only asserts: the pinned anchor `model/Block.go:1329` at commit efaeedf9c is indeed `case len(hashes) == 1:`, and the three blank-line additions clear exactly the three MD032 violations that `.markdownlint.yml` enables.

Three wording nits, all at `plans/subtree-lift-audit-notes.md`:

- Line 4 — "not preserved in this repository" suggests the file was once present and then lost. `git rev-list --all --objects` finds no object by that name, so it was never committed.
- Lines 4, 178, 197, 213, 240, 313 — the same parenthetical repeats six times. One note in the header plus plain references to "the design plan" would read better and shrink the diff.
- Lines 213-214 and 240-241 — the wrapping is ragged after the substitution. Line 214 is the orphan fragment "predicates Strategy B on the", and lines 240-241 produce adjacent parentheses.

ctnguyen (approved):
Review at commit [730aa0a](https://github.com/icellan/teranode/commit/730aa0a9892222409bf399336db834e637947457)

Solid little cleanup, and the premise holds up: `plans/subtree-lift.md` genuinely never existed — `git log --all --full-history` and `git rev-list --all --objects` both come up empty — so those six citations were dead, not merely moved. The rewrite is meaning-preserving with no design content invented, and the blank lines clear real `MD032` violations under the repo's `.markdownlint.yml`. The head commit also resolves all three of ordishs' nits, so I haven't re-raised them; what's left is the same class of stale reference in the places this PR didn't reach.

## Non-blocking issues

### ChiR1 — Every line anchor still points at a three-month-old commit

**Problem:** The doc pins its anchors with `Teranode at HEAD (commit efaeedf9c)` (line 13), and every line number in the file is valid only at that commit — including the ones this PR rewrites around: `model/Block.go:1329` (lines 183, 198, 215-216) and `services/blockassembly/Server.go:1484` (lines 246, 286). `efaeedf9c` is dated 2026-05-13 and is 383 commits behind this branch's head. At head, `CheckMerkleRoot` has moved from `model/Block.go:1290` to 1762 and `case len(hashes) == 1:` from 1329 to 1801; `Server.go` 1383/1420/1484/1538/1591/1610/1999 now land on `select {`, an unrelated comment, `return nil, err`, and similar. So the file still walks a reader into the wrong code — the same failure mode the PR set out to remove, just in the anchors rather than the file citation.

**Why it matters:** The whole audit conclusion turns on *which* `CheckMerkleRoot` branch is exercised today, so `model/Block.go:1329` is the single most load-bearing pointer in the document, and it now resolves to unrelated helper code. "Teranode at HEAD" also reads as "current main" while the numbers are pinned to a May commit, which invites exactly the wrong reading.

**Fix:** Name the symbol next to each number and make the pin unambiguous — symbol names survive rebases, line numbers don't. For example: `Teranode at commit efaeedf9c (2026-05-13) — all line numbers below are relative to that commit, not to current main`, and `the case len(hashes) == 1 branch of model.Block.CheckMerkleRoot (Block.go:1329 at efaeedf9c)`. Clearly outside the PR's stated scope, so fine to defer to a follow-up.

### ChiR2 — PR description still carries the wording the head commit removed

*Same as [@ordishs's point 1 (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1550#pullrequestreview-4907188134).*

**Problem:** His nit landed against the file and is fixed there (line 4 now says "was never committed"), but the same superseded wording survives in the PR summary: "Rephrases each citation to note that the design plan is *not preserved in this repository*". That sentence is now stale twice over — the wording he objected to, plus the "each citation" mechanic, which the head commit deliberately replaced with one header note and plain "the design plan" references.

**Why it matters:** On squash merge the PR body becomes the commit body, so both the retracted phrasing and a description that no longer matches the diff land in permanent history. Reviewers reading top-down also hit the mismatch before they reach the file.

**Fix:** Reword the summary to match head, e.g. "Replaces the six dead citations with plain references to 'the design plan', and states once, in the header, that the plan was never committed to this repository." The test-plan boxes still hold as written — `grep -n "subtree-lift.md" plans/subtree-lift-audit-notes.md` returns nothing at head.

## Recap

|---|---|---|---|
| ChiR1 | Line anchors 383 commits stale | | 30% |
| ChiR2 | PR description stale vs head | | 15% |

PR #1551, opened 2026-08-10
Remove dead build-tx-blaster Makefile target

icellan:
## Summary

- `build-tx-blaster` built `./cmd/txblaster/`, but that directory doesn't exist anywhere in the repo, so the target could only ever fail with a build error. Removed the target and its `.PHONY` entry, plus the now-stale doc line describing it in `docs/howto/makefile.md`.
- The blaster/coinbase binaries used by `compose/docker-compose-3blasters.yml` actually come from an external `teranode-coinbase` image, not from anything built in this repo. Added a comment at the `x-teranode-coinbase-base` anchor so this isn't mistaken for an in-repo build target later.

Checked for other references to `build-tx-blaster` (CI workflows, other Makefile targets) — found none besides the doc line removed here.

## Test plan

- [ ] `make build-tx-blaster` no longer exists as a target
- [ ] `make build` / `make lint` still succeed
- [ ] compose file still parses (`docker compose -f compose/docker-compose-3blasters.yml config`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ dev-dashboard:
 	npm install --prefix ./ui/dashboard && npm run dev --prefix ./ui/dashboard
 
 .PHONY: build
-# build-blockchainstatus build-tx-blaster build-propagation-blaster build-aerospiketest build-blockassembly-blaster build-utxostore-blaster build-s3-blaster build-chainintegrity
+# build-propagation-blaster build-aerospiketest build-blockassembly-blaster build-utxostore-blaster build-s3-blaster build-chainintegrity
 build: update_config build-teranode-with-dashboard build-teranode-cli clean_backup
 
 .PHONY: update_config
@@ -140,10 +140,6 @@ build-teranode-dev:
 # .PHONY: build-blockassembly-blaster
 # build-blockassembly-blaster: set_debug_flags
 # 	go build --trimpath -ldflags="-X main.commit=${GITHUB_SHA} -X main.version=MANUAL" -gcflags "all=${DEBUG_FLAGS}" -o blockassemblyblaster.run ./cmd/blockassembly_blaster/main.go
-
-.PHONY: build-blockchainstatus
-build-blockchainstatus:
-	go build -o blockchainstatus.run ./cmd/blockchainstatus/
 
 # .PHONY: build-aerospiketest
 # build-aerospiketest:
@@ -494,8 +490,6 @@ clean_gen:
 .PHONY: clean
 clean:
 	rm -f ./teranode_*.tar.gz
-	rm -f blaster.run
-	rm -f blockchainstatus.run
 	rm -rf build/
 	rm -f coverage.out
 
--- a/Makefile
+++ b/Makefile
@@ -117,10 +117,6 @@ build-teranode-ci: set_debug_flags set_txmetacache_flag
 build-chainintegrity: set_debug_flags
 	CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS) GOARCH=$(GOARCH) go build -o chainintegrity.run ./compose/cmd/chainintegrity/
 
-.PHONY: build-tx-blaster
-build-tx-blaster: set_debug_flags
-	CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS) GOARCH=$(GOARCH) go build --trimpath -ldflags="-X main.commit=${GIT_COMMIT} -X main.version=${GIT_VERSION}" -gcflags "all=${DEBUG_FLAGS}" -o blaster.run ./cmd/txblaster/
-
 .PHONY: build-teranode-cli
 build-teranode-cli:
 	CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS) GOARCH=$(GOARCH) go build -mod=readonly -o teranode-cli ./cmd/teranodecli
--- a/compose/docker-compose-3blasters.yml
+++ b/compose/docker-compose-3blasters.yml
@@ -9,6 +9,7 @@ x-teranode-base: &teranode-base
     - ../data/test/${TEST_ID:-chainintegrity}:/app/data
 
 x-teranode-coinbase-base: &teranode-coinbase-base
+  # coinbase.run/blaster.run/blaster-tui.run come from this externally-built image; the source is not in this repository.
   image: 434394763103.dkr.ecr.eu-north-1.amazonaws.com/teranode-coinbase:latest
   # image: teranode-coinbase:local
   networks:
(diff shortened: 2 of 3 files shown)


ordishs (approved):
Verified. `cmd/txblaster/` is absent from the tree — commit 3aa0306b4 ("remove commands that are moved to separate repo") deleted it, so `make build-tx-blaster` could only fail.

Three nits:

- `Makefile:71` — the comment above the `build` target still lists `build-tx-blaster` among the optional build targets, so the removal is incomplete and the PR body claim is not quite right.
- `Makefile:501` — the `clean` target still runs `rm -f blaster.run`, and `docs/howto/makefile.md:194` still documents it. No in-repo target can produce that file after this PR.
- `Makefile:148-150` — `build-blockchainstatus` has the identical defect: `cmd/blockchainstatus/` does not exist. This PR removes only the txblaster case.

ctnguyen (approved):
Review at commit [7f83de6](https://github.com/icellan/teranode/commit/7f83de6bc9c2645b69fb429c987f43dc94c64a54)

Good, well-scoped cleanup. Both targets really were dead — `cmd/txblaster/` and `cmd/blockchainstatus/` are absent from the tree, so `make build-tx-blaster` / `make build-blockchainstatus` could only ever fail — and the removal of those two names is complete: no `build-tx-blaster`, `build-blockchainstatus` or `blockchainstatus.run` reference survives anywhere in the Makefile, docs, `.github/workflows`, `scripts/` or the `Dockerfile`. (The `blaster.run` entrypoints in the compose files stay, correctly — they run inside the external `teranode-coinbase` image, which is the point of the new comment; `compose/multinode.sh:778` confirms that repo is where it is built.) Nothing in CI invoked the removed targets, and `*.run` is gitignored, so dropping the two `rm -f` lines from `clean` has no practical fallout. The second commit also picked up all three of the earlier review nits. The remaining points are follow-up polish; none should hold the merge.

## Non-blocking issues

### ChiR1 — Five more dead build targets still advertised, in the Makefile and the docs

*Extends [@ordishs's point 1 (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1551#pullrequestreview-4907187909).*

**Problem:** The head commit removed `build-blockchainstatus` and `build-tx-blaster` from the comment at `Makefile:71`, which resolves the literal nit, but the same line still lists `build-propagation-blaster build-aerospiketest build-blockassembly-blaster build-utxostore-blaster build-s3-blaster` — all five are commented-out target blocks at `Makefile:128-146` pointing at `./cmd/propagation_blaster/`, `./cmd/aerospiketest/`, `./cmd/blockassembly_blaster/`, `./cmd/utxostore_blaster/`, `./cmd/s3_blaster/`, none of which exist in the tree. Of the six names on that line only `build-chainintegrity` is real. The docs go further: `docs/howto/makefile.md:428-434` lists those same five under "Several targets are commented out in the Makefile but **may be re-enabled as needed**" — which is not true, since re-enabling any of them just reproduces the build failure this PR set out to remove.

**Why it matters:** This is exactly the defect the PR exists to fix — the build system advertising targets whose source was moved out to a separate repo. The docs claim is the worse half: a reader following it will uncomment a target and hit the same "no such directory" error. Finishing it here is cheaper than a third PR in this series.

**Fix:** Reduce the comment to the one live target, delete the dead commented-out blocks, and either drop the docs bullet or restate it as history:

```make
.PHONY: build
# build-chainintegrity
build: update_config build-teranode-with-dashboard build-teranode-cli clean_backup
```

### ChiR2 — PR body describes only half the change, test plan unchecked

**Problem:** The description covers the `build-tx-blaster` removal and the compose comment only. It does not mention `build-blockchainstatus`, the `clean` lines, or the second doc bullet, all of which came in with `7f83de6`. It also still asserts "Checked for other references … found none besides the doc line removed here", which the second commit contradicts. All three test-plan boxes are unchecked.

**Why it matters:** On a squash merge the body becomes the commit message, so the permanent history under-describes the change and carries a claim the diff disproves. The unchecked boxes leave it ambiguous whether `make build` and the compose parse were re-run after the second commit.

**Fix:** Refresh the summary to cover both removals plus the `clean` and comment cleanups, drop the stale "found none" sentence, and tick the boxes once re-verified.

## Recap

| --- | --- | --- | --- |
| ChiR1 | Five dead targets still advertised | | 35% |
| ChiR2 | PR body stale, test plan unchecked | | 15% |

freemans13 (approved):
Approving at `7f83de6`. Independently verified rather than restating what @ordishs and @ctnguyen already covered — their points stand and I won't repeat them.

## What I checked

Both removed targets really were dead: `cmd/txblaster/` and `cmd/blockchainstatus/` are absent from the tree, and a full-repo grep finds no surviving reference to `build-tx-blaster`, `build-blockchainstatus` or `blockchainstatus.run` in the Makefile, docs, `.github/workflows`, `scripts/` or the `Dockerfile`. The Makefile still parses — `make -n build` and `make -n clean` both resolve, and `build-chainintegrity` (the one genuinely live name left on the `Makefile:71` comment line) still builds.

Two things worth recording because they look like problems and aren't:

**The compose comment is on the right anchor.** The diff's hunk header shows `x-teranode-base`, which makes it look misplaced, but that's just the nearest preceding anchor — the line lands at `compose/docker-compose-3blasters.yml:12`, attached to `x-teranode-coinbase-base`, exactly as the description says.

**Nothing is broken by the loss of `blaster.run`.** Three compose files still have `entrypoint: ["./blaster.run"]`; if any of them ran that against the in-repo `teranode:latest` image it would now fail. All three inherit `x-teranode-coinbase-base` → the external ECR `teranode-coinbase` image, so they're unaffected. And since `*.run` is gitignored, dropping the two `rm -f` lines from `clean` has no fallout — `clean` never removed `teranode.run` or `chainintegrity.run` either, so this isn't a new gap.

## One non-blocking gap

The clarifying comment landed only in `docker-compose-3blasters.yml`. `compose/docker-compose-chainintegrity.yml:11` and `test/docker-compose-host.yml:12` carry a byte-identical `x-teranode-coinbase-base` anchor, same ECR image, same `./blaster.run` entrypoints — and neither got the comment. Since the point of the comment is to stop someone mistaking these for in-repo build targets, and those two files are just as likely to be the ones read first, the same line is worth copying across:

```yaml
x-teranode-coinbase-base: &teranode-coinbase-base
  # coinbase.run/blaster.run/blaster-tui.run come from this externally-built image; the source is not in this repository.
  image: 434394763103.dkr.ecr.eu-north-1.amazonaws.com/teranode-coinbase:latest
```

Cosmetic, and not worth holding the merge for.

## Caveat on verification

I did not run `make lint` or `make test`. No Go files changed, and `make lint` only inspects files changed versus main, so neither would exercise anything in this diff — the Makefile dry-runs above are the equivalent check for what actually moved. Worth flagging so the approval isn't read as a green test suite.

PR #1553, opened 2026-08-10
docs(aerospike): correct stale UTXO pagination threshold in comments

icellan:
## Summary
- Package-level comments across `stores/utxo/aerospike/*.go` stated pagination kicks in above a hardcoded 20,000 UTXOs. That was the original default of `utxostore_utxoBatchSize`, which has since been lowered (default 128, docker=50, docker.m=512).
- Replaced the stale hardcoded number with a reference to the actual configurable setting, e.g. "Pagination Records: Additional records for transactions with more outputs than utxostore_utxoBatchSize (default 128)".
- No functional change. The pagination mechanism itself (main record + paginated child records) is correct and untouched.
- The same stale figure also appears in `stores/utxo/aerospike/README.md`; left out of this PR since fixing it triggers unrelated pre-existing markdownlint violations in that file — worth a separate small cleanup PR.

## Test plan
- [x] `go vet ./stores/utxo/aerospike/...` passes
- [x] Comment-only / doc-only diff, no code logic touched


ordishs (approved):
Verified. `settings/utxostore_settings.go:26` and `settings/settings.go:457` both set `utxostore_utxoBatchSize` to 128, `settings.conf:1392` agrees, and `stores/utxo/aerospike/aerospike.go:342` reads only that setting, so the old "20,000 UTXOs" comment was stale. The new text matches `splitIntoBatches` at `stores/utxo/aerospike/create.go:661-670`, and the diff covers all 12 Go files in the package that carried the stale phrase.

Two nits:

- The stale 20,000 figure survives in Markdown: `docs/topics/datamodel/utxo_data_model.md` lines 455, 457, 461, 482, 497, 516, 519. The repository now states two different thresholds.
- `docs/topics/features/utxo_lock_records.md:334` and `docs/topics/stores/utxo.md:352` name a setting key `utxo_store_batch_size` that does not exist anywhere in the repo. The real key is `utxostore_utxoBatchSize`.

ctnguyen (changes requested):
Review at commit [0e7f415](https://github.com/icellan/teranode/commit/0e7f415d994d75c870c5ad9bcd9f8bdcdf694cbb)

Good catch on the stale 20,000 — it was the original `utxostore_utxoBatchSize` default and it had drifted a long way from the current 128, so every one of those package headers was actively misleading. I checked the new claims against the tree and they hold up: `MaxTxSizeInStoreInBytes` really is `const … = 32 * 1024` in `stores/utxo/aerospike/aerospike.go:92` with no setting behind it, the `> utxoBatchSize` (not `>=`) framing matches `splitIntoBatches`, the reworked 300-output example lands exactly on 128/128/44 with `totalExtraRecs = 2`, and the "never change after initialisation" warning is real — `CalculateKeySource(txID, vout, utxoBatchSize)` bakes it into the key. Replacing hardcoded numbers with the setting name rather than swapping 20,000 for 128 was the right call; it won't rot again. The two nits from the earlier review are both addressed in this head commit. What's left is that the second commit reached into the externalisation story, and that part of the docs is not consistent with itself yet.

## Blocking issues

### ChiR1 — `utxo_data_model.md` still says a paginated transaction keeps its tx data in Aerospike, contradicting the bullet this PR added

**Problem:** The decision diagram at `docs/topics/datamodel/utxo_data_model.md:434-438` routes `Medium (≤32KB, > utxoBatchSize outputs)` to "Multiple Aerospike Records (Pagination)" with a master record holding "Full transaction data", and the Storage Flow list at lines 591-598 only has two branches (`≤32KB and outputs ≤ utxoBatchSize` → inline, `>32KB` → blob), so the ≤32KB-but-paginated case is documented as staying inline. The code says the opposite: `GetBinsToStore` ends with `if len(batches) > 1 { external = true }` (`create.go:946-947`), so the master record gets an `external=true` bin and no `inputs`/`outputs` bins; `sendStoreBatch` routes any multi-record item to `StoreTransactionExternally` (`create.go:433`), and `storeExternallyWithLock` writes the blob unconditionally (`create.go:1099`). A 2KB, 300-output transaction is externalised. The trigger-conditions bullet this PR added at line 557 states exactly that — so the file now contradicts itself 120 lines apart, and `stores/utxo/aerospike/README.md:52-56` (which already shows `external = true` on a paginated record) sides with the code.

**Why it matters:** Externalisation behaviour is the one thing the head commit set out to align with the code, and this is the branch a reader hits first — the ASCII diagram is the summary everyone skims. Anyone sizing blob storage, or debugging why a small transaction has `external=true` and a `.tx` file, gets the wrong answer from the diagram and the right one from the bullet, with nothing to indicate which to trust.

**Fix:** Make the Medium branch show external storage, e.g. `├─ external=true flag (no inline tx data)` / `├─ Blob Storage (.tx or .outputs)` in place of "Full transaction data", and add a step to the Storage Flow list: `If outputs > utxoBatchSize (regardless of size): write transaction to blob storage, split UTXOs across master + child records, set external=true`. Worth a one-liner on why — the multi-record write path is the two-phase-commit path, which always goes through the external store.

## Non-blocking issues

### ChiR2 — "does not fit in a single record" is no longer a true reason at 128 outputs

**Problem:** `docs/topics/features/utxo_lock_records.md:27` now reads "transactions with more outputs than `utxostore_utxoBatchSize` (default 128). Such a transaction does not fit in a single record and must be split", and `docs/topics/datamodel/utxo_data_model.md:457` keeps "…are automatically split across multiple Aerospike records to stay within database size constraints". Those rationales were plausible at 20,000 outputs; at 128 they are wrong. 128 UTXO entries are roughly 4-9KB, against the ~1MB record cap the same file cites at `utxo_lock_records.md:42`. `splitIntoBatches` splits purely on the configured `s.utxoBatchSize`, with no size input at all. The number was corrected but the sentence built around the old number was carried over.

**Why it matters:** It contradicts the record-size figure in the same document, and it tells anyone tuning `utxoBatchSize` that raising it risks `RECORD_TOO_BIG` at a few hundred outputs, which is false — the actual size guards are the 32KB inline threshold and the `RECORD_TOO_BIG` retry at `create.go:606`.

**Fix:** Attribute the split to the configured batch size, not the record limit: "…are split into records of `utxostore_utxoBatchSize` UTXOs each, so a record never grows unbounded with the output count. Actual record-size overflow is handled separately by the 32KB externalisation threshold and by the `RECORD_TOO_BIG` retry."

### ChiR3 — the stale 20,000 survives in two tracked files

*Same as [@ordishs's point 1 (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1553#pullrequestreview-4907187295).*

**Problem:** His instances in `docs/topics/datamodel/utxo_data_model.md` are all fixed at head — this extends to two files neither of you listed. `stores/utxo/aerospike/README.md` still has it at lines 36, 52, 61 and 70 (`### Paginated Transaction (> 20,000 outputs)`, `utxos - utxo[0..19999]`, `recordUtxos - 20000`, `utxos - utxo[20000..39999]`) — the README of the very package whose twelve headers this PR corrects. One further tracked doc found by grepping for the old figure still describes pagination as kicking in above 20,000 outputs as well.

**Why it matters:** After this PR the repo is *more* internally inconsistent about the number than before at these two spots, because the surrounding sources now all say 128. The README is the first thing someone opens when working on the Aerospike store.

**Fix:** The README's stale figures are two headings and three lines inside fenced code blocks, so the numeric fix itself trips no markdownlint rule; the deferral cost is only the handful of pre-existing mechanical violations the hook surfaces on any touch (`MD031`/`MD040` around the fences), which are a few blank lines and fence languages. Fixing them here is cheaper than a follow-up PR that nobody schedules — but if you'd rather keep the split, at least take the other file, which has no such excuse.

### ChiR4 — `docs/topics/stores/utxo.md` still frames externalisation as a 1MB / `RECORD_TOO_BIG` mechanism

**Problem:** This PR edits `docs/topics/stores/utxo.md:352`, but the section right above it at line 332 still says "to overcome Aerospike's record size limitation of 1MB, the system implements an external storage mechanism for large transactions", followed by "If a transaction is too large to fit in a single Aerospike record (indicated by a `RECORD_TOO_BIG` error) … the UTXO store will store the full transaction data in an external storage". The primary trigger is the 32KB `MaxTxSizeInStoreInBytes` check at `create.go:416`, plus `ExternalizeAllTransactions` and the multi-record path; `RECORD_TOO_BIG` is only the fallback retry at `create.go:606`.

**Why it matters:** It leaves a reader with an externalisation threshold about 30× too high, in the one document that is the entry point for the UTXO store — and it now disagrees with the 32KB figure this PR introduced two documents over.

**Fix:** "Transactions whose extended size exceeds `MaxTxSizeInStoreInBytes` (32KB), transactions that need more than one record, and all transactions when `utxostore_externalizeAllTransactions` is set, are stored externally. A `RECORD_TOO_BIG` response is handled as a fallback retry via the same path."

### ChiR5 — a compile-time constant in a table headed "configuration settings"

**Problem:** `docs/topics/features/utxo_lock_records.md:336` adds `| MaxTxSizeInStoreInBytes (compile-time constant, not a setting) | 32KB | Size threshold for external storage |` to a table introduced as "The lock record pattern uses these configuration settings". The annotation is accurate and the value is right, but the row leaves a reader looking for a key they can set, with no pointer to where the constant lives.

**Why it matters:** Minor, but it is the one row in that table nobody can act on, and the natural next question ("where do I change it?") has no answer in the text.

**Fix:** Drop the row from the table and put it underneath as a note: "The external-storage size threshold is not configurable: `MaxTxSizeInStoreInBytes = 32 * 1024` in `stores/utxo/aerospike/aerospike.go`."

## Recap

| ChiR1 | Paginated tx wrongly shown as inline | YES | 60% |
| ChiR2 | False "doesn't fit" rationale | | 45% |
| ChiR3 | 20,000 left in two files | | 30% |
| ChiR4 | 1MB externalisation trigger | | 22% |
| ChiR5 | Constant listed as a setting | | 10% |

freemans13 (commented):
Reviewed at head `0e7f415d9`. The direction is right and the arithmetic checks out — I verified `MaxTxSizeInStoreInBytes = 32 * 1024` at `stores/utxo/aerospike/aerospike.go:92` (pinned by the test at `aerospike_test.go:782`), the strictly-greater-than framing against `splitIntoBatches` at `create.go:661-673`, the 128/128/44 worked example, and `utxostore_externalizeAllTransactions` at `settings/settings.go:481`. `go vet ./stores/utxo/aerospike/... ./settings/...` is clean.

I've deliberately not re-litigated @ordishs's nits or @ctnguyen's ChiR1–ChiR5 — I checked them independently and agree, ChiR1 especially (`GetBinsToStore` sets `external = true` whenever `len(batches) > 1` at `create.go:946-947`, and `settings.conf:1390` says so in plain English: *"utxos per record, if larger it will externalize the transaction"*). Everything below is new.

## Finding A — the stale number survives in all twelve files this PR edits

About forty lines below the corrected `# Architecture` bullets, the same package header still reads:

```go
// Large Transaction with External Storage:
//   - Same as normal but with external=true
//   - Transaction data stored in blob storage
//   - Multiple records for >20k outputs
```

Present and unchanged in every file the PR touches:

| File | Line |
|---|---|
| `stores/utxo/aerospike/aerospike.go` | 47 |
| `stores/utxo/aerospike/aerospike_test.go` | 47 |
| `stores/utxo/aerospike/alert_system.go` | 47 |
| `stores/utxo/aerospike/create.go` | 47 |
| `stores/utxo/aerospike/delete.go` | 50 |
| `stores/utxo/aerospike/get.go` | 47 |
| `stores/utxo/aerospike/metrics.go` | 47 |
| `stores/utxo/aerospike/set_mined.go` | 50 |
| `stores/utxo/aerospike/spend.go` | 48 |
| `stores/utxo/aerospike/teranode.go` | 48 |
| `stores/utxo/aerospike/test_helper.go` | 48 |
| `stores/utxo/aerospike/un_spend.go` | 48 |

**Why all three of us missed it:** it's spelled `20k`, not `20,000` or `20000`. Every grep for the old figure — mine included, at first — sails straight past. I only found it by rendering the package doc instead of searching the source.

**Why it matters:** this isn't a distant document that happens to mention the number. It's the same comment block, in the same files, in the PR whose whole purpose is removing that number. Each header now says "more outputs than `utxostore_utxoBatchSize` (default 128)" in one section and ">20k outputs" one screen further down, with nothing to tell a reader which half is current.

**Fix:** `//   - Multiple records when outputs exceed utxostore_utxoBatchSize` in each of the twelve, then a repo-wide `grep -i '20k'` sweep. I ran that sweep — the only other hits are legitimate (subtree benchmark names, a netsync comment about 20k-tx blocks), so this is the last of it.

## Finding B — the package comment is duplicated fourteen times, and godoc concatenates them

This is the root cause of the rot, and it's the durable fix.

Fourteen files in `stores/utxo/aerospike` open with a `// Package aerospike ...` comment — twelve carry the full multi-section block, `set_mined_expressions.go` and `spend_expressions.go` carry a one-liner. There's no `doc.go`.

I assumed at first that godoc picks one and the other eleven are harmless dead weight. That's wrong, and the truth is worse — I checked on a throwaway two-file package, and `go/doc` **concatenates** every package comment it finds. Measured on this package at head:

```
$ go doc ./stores/utxo/aerospike | wc -l
     755
$ go doc ./stores/utxo/aerospike | grep -c 'Package aerospike provides'
14
$ go doc ./stores/utxo/aerospike | grep -c '# Architecture'
11
$ go doc ./stores/utxo/aerospike | grep -c '20k outputs'
11
```

So the rendered package documentation is 755 lines repeating the same Architecture / Features / Usage / Database Structure block eleven times, and the stale `>20k outputs` line appears eleven times in the published doc for the very package this PR is correcting.

The duplication is why one number was able to drift in twelve places at once: there are twelve copies and nothing keeping them in step. Fixing the number without fixing the duplication guarantees the next drift.

**Fix:** move the block into `stores/utxo/aerospike/doc.go` and delete it from the other thirteen files. That's clearly outside this PR's scope and I'm not asking you to bundle it — but it's worth an issue, it turns the next twelve-file diff into a one-file diff, and Finding A collapses into it if it lands first.

## Finding C — "(default 128)" isn't true in the two contexts most of us run locally

The stated principle here — name the setting so the text can't rot again — is the right one. But the PR then appends the literal default alongside it, roughly twenty times across the Go headers and the markdown, which reintroduces the same failure mode. And the number isn't universally true today. From `settings.conf:1392-1394`:

```
utxostore_utxoBatchSize          = 128
utxostore_utxoBatchSize.docker   = 50
utxostore_utxoBatchSize.docker.m = 512
```

Anyone on a `docker` or `docker.m` context reads "(default 128)" against a store batching at 50 or 512.

**Suggestion (not a blocker):** drop the parenthetical from the twelve package headers and just name the setting — the headers don't need a number to be useful. Keep the figure where a reader can act on it and where it's maintained anyway: the `longdesc` in `settings/utxostore_settings.go`, and `settings.conf`.

## Finding D — naming the file ChiR3 left unnamed

@ctnguyen's ChiR3 mentions "one further tracked doc found by grepping for the old figure". For the record it's `.claude/skills/debug-teranode/references/architecture.md:158`:

> **Pagination:** >20,000 outputs (`utxoBatchSize`) → master record (index 0) + child records…

That's the reference file behind the `debug-teranode` skill, so a wrong threshold there feeds straight into future debugging sessions. The other outstanding spot is `stores/utxo/aerospike/README.md` lines 36, 52, 59, 61, 70, already covered by ChiR3.

## Finding E — the externalisation threshold compares *extended* size (minor)

`docs/topics/datamodel/utxo_data_model.md:556` says external storage is used when "Transaction size > `MaxTxSizeInStoreInBytes` (32KB)". The code compares the **extended** size — `extendedTxSize(tx)` at `create.go:413`, which adds the per-input previous-output data that `tx.Size()` omits — and for a partial transaction with no inputs it sums output sizes only (`create.go:403-409`). A transaction whose plain serialised size is comfortably under 32KB can still externalise.

The Storage Flow list further down already says "Calculate transaction extended size", so the document is half-right; it's the trigger bullet that's loose. One word ("extended size") rather than a rewrite.

## Recap

| ID | Finding | Required |
|---|---|---|
| A | `>20k outputs` stale in all 12 edited files, ~40 lines below the fix | YES |
| B | Package comment duplicated 14×, godoc concatenates to 755 lines | separate PR |
| C | "(default 128)" untrue for `docker` (50) / `docker.m` (512) | suggestion |
| D | Names the file ChiR3 left unnamed | |
| E | Externalisation compares extended size, not tx size | |

Only A is really on you for this PR, and it's twelve one-line edits in files you already have open.

**Limits on this review:** I read the full diff plus surrounding code and verified every factual claim in the changed text against the tree. I did not run `make test` or `make lint` — the diff is comment-only, `go vet` is clean, and CI here is already green. A, C, D and E are objectively checkable from the line numbers above. B's severity is a judgment call; the 755-line godoc output is measured fact, but whether it justifies a refactor is yours and @ordishs's call.

icellan:
Addressed ChiR1 (blocking) — the Medium-size decision diagram in utxo_data_model.md showed a paginated transaction keeping full transaction data inline, contradicting the externalisation bullet this branch already added ~120 lines below. Updated the diagram to show external=true/blob storage and added the missing Storage Flow step.

Also fixed the stale ">20k outputs" figure freemans13 found surviving in all twelve stores/utxo/aerospike package headers this branch already touches (spelled `20k`, which is why earlier greps for `20,000`/`20000` missed it), plus the same stale figure in .claude/skills/debug-teranode/references/architecture.md.

Deferred: Finding B (consolidating the duplicated package comment into a doc.go) — explicitly flagged by the reviewer as separate follow-up material, not for this PR.

docs/topics/datamodel/utxo_data_model.md
```diff
@@ -508,31 +508,31 @@ Contain UTXO batches with minimal metadata:
...
-  - utxos[0:20000]        // First 20K UTXOs
-  - totalUtxos = 45000
+  - utxos[0:128]          // First batch of UTXOs
+  - totalUtxos = 300
   - totalExtraRecs = 2
-  - recordUtxos = 20000   // Non-nil in this record
+  - recordUtxos = 128     // Non-nil in this record
   - spentUtxos = 0
   - spentExtraRecs = 0
   - Full tx data or external=true
```
freemans13: A paginated master record never carries inline tx data — `GetBinsToStore` forces `external = true` when `len(batches) > 1` (`stores/utxo/aerospike/create.go:947`) and then writes the `external` bin instead of `inputs`/`outputs`. Since this example is 300 outputs at a batch size of 128, it is always the external case.

```suggestion
  - external=true (tx data in blob storage)
```

freemans13 (commented):
Reviewed by checking every factual claim in the diff against the code, since this PR is entirely claims about how the code behaves. The corrections are accurate and CI is green. Two things worth addressing before merge.

## One contradiction left standing

The "Pagination Example" master record still ends with `- Full tx data or external=true` (`docs/topics/datamodel/utxo_data_model.md:522`). That example is a 300-output transaction at the default batch size of 128, so it is by construction multi-record — and `GetBinsToStore` does `if len(batches) > 1 { external = true }` (`stores/utxo/aerospike/create.go:947`), with the branch immediately after writing the `external` bin *instead of* the `inputs`/`outputs` bins. A paginated master record can therefore never hold inline transaction data, so the "Full tx data or" half is impossible.

This is exactly what commit 57f07237e fixed in the decision diagram ~80 lines above ("external=true flag (no inline tx data)"); the worked example kept the old wording. Inline suggestion below.

## The README gap

`stores/utxo/aerospike/README.md` still carries four stale 20,000 references (lines 36, 52, 61, 70), plus a "For a UTXO with vout 25000" key-calculation example that only makes sense in the 20k era.

Once this merges, the twelve `.go` package headers in that directory will say "utxostore_utxoBatchSize (default 128)" while the README beside them says 20,000 — the package will contradict itself, which is arguably worse than the uniform staleness it started in.

The stated blocker is pre-existing markdownlint violations. That's honest, but it isn't gating: markdownlint appears only in `.pre-commit-config.yaml`, not in any workflow under `.github/workflows/`. The violations are also cheap — the fences need a language tag and surrounding blank lines. Worth folding in here rather than deferring to a follow-up.

## Nits, take or leave

1. The master-record table now says `totalExtraRecs` is present "If outputs > `utxoBatchSize`", but `create.go:908` writes that bin unconditionally, with value 0 for single-record transactions. The rewording carried a pre-existing inaccuracy forward.
2. "Transaction extended size > `MaxTxSizeInStoreInBytes`" is right for full transactions, but for partial (outputs-only) ones `create.go:403-408` sums output sizes instead of using extended size.
3. `.claude/skills/debug-teranode/references/architecture.md` sits outside the aerospike/utxo docs the title advertises. Same stale fact, so I'd keep it — just noting the scope creep.

## Verified correct

- Default 128 with `docker=50` and `docker.m=512` — `settings.conf:1392-1394`, `settings/settings.go:457`.
- `MaxTxSizeInStoreInBytes = 32 * 1024` is a genuine compile-time constant with no settings key overriding it (`aerospike.go:92`), so the new "not configurable" note is right.
- `utxostore_externalizeAllTransactions` is the correct key — `settings/settings.go:481`.
- Pagination triggers at outputs above the batch size — `splitIntoBatches`, `create.go:663`.
- Multi-record does force external storage — `create.go:947`.
- `RECORD_TOO_BIG` genuinely retries through the same external-storage path — `create.go:606-620`.
- The "never change after initialisation" warning matches the setting's own `longdesc`.
- The 300-outputs-into-128/128/44 arithmetic checks out.
- No stale key names (`utxo_store_batch_size`, `utxo_store_max_tx_size_in_store`, `utxo_store_externalize_all_transactions`) survive anywhere in the tree.

`go vet ./settings/...` and `go build ./stores/utxo/aerospike/...` are clean at 57f07237e, and all CI checks pass (one aerospike fork-image job skipped).

docs/topics/datamodel/utxo_data_model.md
icellan: Correct, applied in 1c64894d5. `docs/topics/datamodel/utxo_data_model.md:522` now reads `external=true (tx data in blob storage)` — `GetBinsToStore` sets `external = true` unconditionally once `len(batches) > 1`, so the 300-output/128-batch example can never be inline. Rest of the Pagination Example section checked; no other sentence implies inline storage.

ctnguyen (approved):
Review at commit [57f0723](https://github.com/icellan/teranode/commit/57f07237edb8681452de54b54771143f8d6faab7)

Third pass, and this head is clearly better than the last one. The blocking part of ChiR1 is essentially dealt with — the Medium branch of the decision diagram no longer claims a paginated transaction keeps its tx data inline, and the new Storage Flow step 4 spells out the case the list was missing. The `>20k outputs` line that survived twelve package headers (spelled `20k`, which is why three of us searched past it) is gone; the only place the stale figure still lives is `stores/utxo/aerospike/README.md`. ChiR4 and ChiR5 are cleanly closed, and the `.claude/skills/debug-teranode` reference file is corrected too. What's left: three earlier points still have a tail, and the externalisation prose has a blind spot around partial (outputs-only) transactions in two places.

## Status of previous points

Three points are only partially closed, and the remainders stay under their original IDs rather than becoming new items.

**ChiR1** — the diagram and the Storage Flow list are fixed, but the *Pagination Example* immediately below still ends the master record with `- Full tx data or external=true` (`docs/topics/datamodel/utxo_data_model.md:522`). That example is 300 outputs at a batch size of 128, so it is multi-record by construction, and `GetBinsToStore` sets `external = true` whenever `len(batches) > 1` and then writes the `external` bin *instead of* `inputs`/`outputs` (`stores/utxo/aerospike/create.go:946-956`) — the "Full tx data or" half cannot happen. @freemans13 posted the exact one-line suggestion for this ([inline comment, 18-08-2026](https://github.com/bsv-blockchain/teranode/pull/1553#discussion_r3805132539)); take it. Secondary: the Medium branch (lines 434-442) now shows `external=true` but no blob-storage destination, so a reader sees the flag with nowhere for the bytes to go — one extra `├─ Blob Storage (.tx or .outputs)` line closes it.

**ChiR2** — both sentences I quoted are fixed. The same misattribution survives twice more: `docs/topics/features/utxo_lock_records.md:42` ("Aerospike limits individual records to ~1MB; large transactions must span multiple records") and, in a file this PR already edits, the `splitIntoBatches` doc comment at `stores/utxo/aerospike/create.go:649-650` ("by splitting them into multiple records to stay within Aerospike size limits"). The function takes no size input at all — it splits on `s.utxoBatchSize` and nothing else (`create.go:661-670`).

**ChiR3** — `.claude/skills/debug-teranode/references/architecture.md:158` is fixed. `stores/utxo/aerospike/README.md` is not: lines 36 and 52 still say `20,000 outputs`, lines 59/61/70 still say `utxo[0..19999]` / `recordUtxos - 20000` / `utxo[20000..39999]`, and line 78's key-calculation example is "For a UTXO with vout 25000". After this merges, twelve headers in that directory say 128 and the README beside them says 20,000. On the stated blocker: `markdownlint` appears only in `.pre-commit-config.yaml` — nothing under `.github/` runs it — so it is not gating this merge.

| ChiR1 | Paginated tx shown as inline | YES | PARTIAL FIXED |
| ChiR2 | False "doesn't fit" rationale | | PARTIAL FIXED |
| ChiR3 | 20,000 left in two files | | PARTIAL FIXED |
| ChiR4 | 1MB externalisation trigger | | FIXED |
| ChiR5 | Constant listed as a setting | | FIXED |

## Non-blocking issues

### ChiR6 — `utxo.md` says externalisation stores and retrieves "the full transaction data"; for partial transactions it stores outputs only

**Problem:** The section this PR rewrites states it three times: "the UTXO store will store the full transaction data in an external storage" (`docs/topics/stores/utxo.md:334`), "the full transaction data is stored externally" (`:336`), and "it will then retrieve the full transaction data from the external storage using the transaction hash as a key" (`:338`). For a partial transaction — `len(tx.Inputs) == 0`, the outputs-only case — no full transaction is ever written. The store builds a `utxopersister.UTXOWrapper` (txid, height, coinbase flag, and per-output value + locking script) and writes it as `FileTypeOutputs`, on both the inline path (`stores/utxo/aerospike/create.go:462-507`) and the locked multi-record path (`StorePartialTransactionExternally`, `create.go:988-1023`). The read side matches: `getExternalTransaction` asks for `FileTypeTx` first, and on error retries as `FileTypeOutputs` and reconstructs a `bt.Tx` carrying outputs and nothing else — no inputs, no version, no locktime (`get.go:1654-1699`). Nothing is lost by that, since a partial transaction has no inputs to begin with; the document is simply describing a `.tx` blob that does not exist for these transactions.

Worth noting what I did *not* find: no bin records which file type was written, so the file type is discovered by trial — the `.tx` read must fail before `.outputs` is attempted. That costs one failed blob read per cache miss, not per lookup, since `GetTxFromExternalStore` sits behind the 10-second `externalTxCache` (`get.go:1640-1651`; `utxostore_useExternalTxCache` defaults to `true`). Same blind spot as ChiR8 below, but this one also covers the read path, which is why I'm listing it separately.

**Why it matters:** This is the entry-point document for the UTXO store, and it is the one an operator uses to work out what to look for in the blob store. As written it sends someone hunting for a `<txid>` `.tx` object that was never created, and it implies a `GetMeta`/`BatchDecorate` caller asking for `fields.Tx` on an externalised partial transaction gets a complete transaction back, when `get.go:769-791` hands them the outputs-only reconstruction.

**Fix:** Name the two shapes, matching the `FileTypeTx` / `FileTypeOutputs` split the data-model document already documents.

```markdown
In such cases the transaction data is stored externally while metadata and UTXOs stay
in Aerospike, potentially across multiple records: full transactions are written as a
`.tx` blob, and partial (outputs-only) transactions as a `.outputs` blob holding a
`UTXOWrapper`. The Aerospike record carries `external=true` in either case. On read,
the store fetches whichever of the two exists — an `.outputs` blob yields the outputs
alone, not a complete transaction.
```

### ChiR7 — master-record table's Presence column is wrong for both `*ExtraRecs` bins

*Same as [@freemans13's point 1 under "Nits, take or leave" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1553#pullrequestreview-4962352822), extended to the second row.*

**Problem:** `docs/topics/datamodel/utxo_data_model.md:484` and `:487` now say `totalExtraRecs` and `spentExtraRecs` are present "If outputs > `utxoBatchSize`". Neither is right, and they are wrong in opposite directions. `totalExtraRecs` is written unconditionally on the master record, with value `len(batches)-1` — so it exists as `0` on every single-record transaction (`stores/utxo/aerospike/create.go:909`). `spentExtraRecs` is never written at create time at all; nothing in `GetBinsToStore` or `splitIntoBatches` sets it, it appears only when Lua first increments it (`teranode.lua:1252-1271`), and both the expression and read paths explicitly default a missing bin to 0 (`set_mined_expressions.go:193-195`, `teranode.lua:986`). The PR reworded these cells and carried the pre-existing inaccuracy through.

**Why it matters:** The Presence column is the part of that table someone actually acts on — a tool or query that treats a missing `spentExtraRecs` as an error, or a present `totalExtraRecs` as proof of pagination, gets it backwards. `duplicate_spend_test.go:71-74` asserts the absence, so the code's contract is the opposite of the doc's.

**Fix:** State the real conditions.

```markdown
| `totalExtraRecs` | Number of child records (pagination records) | Always (0 if single-record) |
...
| `spentExtraRecs` | Number of child records fully spent | Created on first child spend; absent means 0 |
```

### ChiR8 — the externalisation trigger is stated in terms that don't hold for partial transactions

*Same as [@freemans13's Finding E (13-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1553#pullrequestreview-4925384031).*

**Problem:** `docs/topics/datamodel/utxo_data_model.md:556` now says external storage is used when "Transaction extended size > `MaxTxSizeInStoreInBytes` (32KB)". That is exactly right for a full transaction, but for a partial one (`len(tx.Inputs) == 0`) the code never computes an extended size — it sums `output.Size()` over the non-nil outputs and compares that (`stores/utxo/aerospike/create.go:403-409`). The new step 4 has the matching gap: "Write transaction to blob storage" is a `.tx` file for a full transaction but a serialised `UTXOWrapper` written as `.outputs` for a partial one (`create.go:988-1023`), which the "Partial Transactions" list further down gets right.

**Why it matters:** This section is where someone goes to work out whether a given transaction will land in the blob store, and the outputs-only path is the one used for the large output sets most likely to be near the threshold. It is also the only externalisation figure in the document that a reader could apply and get a wrong answer from.

**Fix:** Qualify the bullet, and point step 4 at the right file type.

```markdown
- Extended transaction size > `MaxTxSizeInStoreInBytes` (32KB) — for partial
  (outputs-only) transactions the comparison is against the summed output sizes
```

## Recap

| ChiR6 | Partial txs store outputs only | | 28% |
| ChiR7 | Wrong Presence for `*ExtraRecs` | | 15% |
| ChiR8 | Trigger wrong for partial txs | | 12% |

PR #1554, opened 2026-08-10
fix(legacy): restore plain-inventory fallback for block relay to non-sendheaders peers

icellan:
## What

In the legacy (SVNode bridge) peer server's `handleRelayInvMsg`, a newly relayed block was only announced to a connected peer when that peer had negotiated `sendheaders` (in which case a `headers` message is sent). Peers that did not negotiate `sendheaders` got no announcement for the block at all.

## Why

Before commit f022dde6e (the tx fee-filter refactor for `handleRelayInvMsg`, processing `InvTypeTx` with a fee filter check), the function ended with a single shared `sp.QueueInventory(msg.invVect)` call that ran for any inventory item not otherwise handled — covering both `InvTypeTx` and the non-`sendheaders` `InvTypeBlock` case. That refactor moved the `QueueInventory` call into the `InvTypeTx` branch only (returning immediately after), and made the block/`sendheaders` branch also return. The shared tail call, and with it the block fallback for peers without `sendheaders`, was dropped as a side effect of that reshuffle rather than as a deliberate design choice.

This is currently masked because SVNode, the only legacy peer exercised in practice, does negotiate `sendheaders`. Any other wire-compatible client that instead expects plain `inv` announcements for blocks would never be notified of new blocks over this path.

## Fix

Restore the fallback: for `InvTypeBlock` when the peer has not negotiated `sendheaders`, queue the plain inventory vector via `QueueInventory`, mirroring the pre-refactor behavior. Extracted as `handleRelayBlockInvMsg` (taking the same `serverPeerQueueInventory` interface already used by `handleRelayTxMsg`) so it can be unit tested without a live peer connection.

## Testing

Added `TestHandleRelayBlockInvMsg`, which fails to build before the fix (the function doesn't exist) and passes after. Ran `go test ./services/legacy/... -race` and `go vet ./services/legacy/...` clean.

### DIFF
--- a/services/legacy/peer_server.go
+++ b/services/legacy/peer_server.go
@@ -2523,6 +2523,13 @@ func (s *server) handleRelayInvMsg(state *peerState, msg relayMsg) {
 			return
 		}
 
+		// Peers that have not negotiated sendheaders still need the block
+		// announced via a plain inventory message.
+		if msg.invVect.Type == wire.InvTypeBlock {
+			s.handleRelayBlockInvMsg(sp, msg)
+			return
+		}
+
 		if msg.invVect.Type == wire.InvTypeTx {
 			// Don't relay the transaction to the peer when it has
 			// transaction relaying disabled.
@@ -2603,6 +2610,13 @@ func (s *server) handleRelayBlockMsg(sp *serverPeer, msg relayMsg) {
 	}
 
 	sp.QueueMessage(msgHeaders, nil)
+}
+
+// handleRelayBlockInvMsg queues a plain inventory vector for a block to a peer
+// that has not negotiated sendheaders. Peers that did negotiate sendheaders are
+// announced via a headers message instead, in handleRelayBlockMsg.
+func (s *server) handleRelayBlockInvMsg(sp serverPeerQueueInventory, msg relayMsg) {
+	sp.QueueInventory(msg.invVect)
 }
 
 // handleBroadcastMsg deals with broadcasting messages to peers.  It is invoked
--- a/services/legacy/peer_server_test.go
+++ b/services/legacy/peer_server_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/bsv-blockchain/go-bt/v2/chainhash"
 	"github.com/bsv-blockchain/go-chaincfg"
+	txmap "github.com/bsv-blockchain/go-tx-map"
 	"github.com/bsv-blockchain/go-wire"
 	"github.com/bsv-blockchain/teranode/errors"
 	"github.com/bsv-blockchain/teranode/services/blockchain"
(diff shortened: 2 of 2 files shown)


ordishs (approved):
Verified. Commit f022dde6e moved the shared `sp.QueueInventory(msg.invVect)` into the `InvTypeTx` branch, so a block reached no peer without sendheaders. The live path is `netsync/manager.go:3476`, which relays `InvTypeBlock` from the Kafka blocks-final listener. The fix restores the pre-refactor behaviour, sits after the `WantsHeaders` branch, and cannot loop or double-announce, because `manager.go:2380` and `peer.go:2275` add received invs to known inventory and `peer.QueueInventory` drops known items.

One nit: `services/legacy/peer_server_test.go:203` — the new test calls the extracted `handleRelayBlockInvMsg` directly. It does not drive `handleRelayInvMsg`, which holds the dispatch that the bug removed. The same helper-only pattern already covers `handleRelayTxMsg`, so this matches local practice.

freemans13 (approved):
Approving — the fix is correct and the regression story checks out. `git show f022dde6e^` confirms the old `handleRelayInvMsg` ended with a single shared `sp.QueueInventory(msg.invVect)` that a block heading to a non-`sendheaders` peer fell through to, and the fee-filter refactor moved that call inside the `InvTypeTx` branch behind a `return`, dropping the block fallback as a side effect. Restoring it matches btcd/SVNode behaviour.

A few things I checked beyond the diff, all clean:

- `RelayInventory` is only ever called with `InvTypeBlock` (`netsync/manager.go:3476`) and `InvTypeTx` (`peer_server.go:1938` plus the rebroadcast tick), so restoring the fallback for `InvTypeBlock` alone leaves no gap versus the old catch-all tail.
- No echo back to the peer that announced the block: inbound invs are marked via `AddKnownInventory` (`netsync/manager.go:2380`, `peer_server.go:904`/`:997`) and `peer.QueueInventory` drops anything already known.
- The advertisement is honoured — `OnGetData` serves `InvTypeBlock` via `pushBlockMsg` (`peer_server.go:1463`), so a peer that answers the new inv with `getdata` actually gets the block.
- No new stall risk on the `peerHandler` goroutine: `outputInvChan` is buffered at 5000 and this is the same path the existing tx relay already uses.

Locally: `go test -race` on the new tests passes (5/5 repeats of the pipe-based one), `go vet ./services/legacy/` clean, `golangci-lint run ./services/legacy/` reports 0 issues.

### One suggestion, non-blocking

Nothing pins the *other* side of the branch — that a peer which **did** negotiate `sendheaders` still gets a `headers` message and **not** an inv. The two branches are order-dependent: if the new `InvTypeBlock` fallback were ever moved above the `WantsHeaders()` check, `sendheaders` peers would silently start receiving inv announcements instead of headers, and every test in the package would still pass.

Given that the bug being fixed here is precisely "a refactor silently reshuffled these branches", a sibling test to `TestHandleRelayInvMsgBlockToNonSendHeadersPeer` — same real-peer setup, but with `sendheaders` negotiated, asserting a `MsgHeaders` arrives and no `MsgInv` does — would pin both halves of the dispatch and make the next reshuffle fail loudly. Worth adding while the test scaffolding is fresh, but I don't think it should hold up the merge.

### Unrelated CI note

`sequential (shard 2)` is red on `TestLongestChainForkAerospike/with_double_spend_transaction`, in `test/sequentialtest/longest_chain/`. That is an Aerospike fork/double-spend test this diff cannot reach, and it passes on other currently-open PRs, so it reads as runner-specific rather than broken on main — but it did fail 3/3 retries in that job, so probably worth a clean re-run rather than just dismissing it.

PR #1556, opened 2026-08-10
fix(blockvalidation): bind quick-validation checkpoint height to what was actually verified this run

icellan:
### The bug

`verifyCheckpointsAgainstHeaders` in `services/blockvalidation/catchup.go` set
`catchupCtx.highestCheckpointHeight` to the highest checkpoint height across the
**entire globally configured checkpoint list** (`blockchain.HighestCheckpointHeight(catchupCtx.checkpoints)`),
unconditionally - regardless of whether that specific checkpoint fell within the
current catchup range, or was ever hash-checked against a fetched header.

The verification loop only checks checkpoints whose height falls inside
`[firstBlockHeight, lastBlockHeight]` for the current catchup run, and increments
`checkpointsChecked` when at least one checkpoint's hash actually matches a header.

The quick-validation gate is:

```go
canUseQuickValidation := catchupCtx.useQuickValidation && block.Height <= catchupCtx.highestCheckpointHeight
```

Because `useQuickValidation` is set to `true` as soon as `checkpointsChecked > 0`
(i.e. any single checkpoint in the current range verified), and
`highestCheckpointHeight` was the global configured maximum rather than the
verified one, every block up to the globally highest *configured* checkpoint became
eligible for quick validation - including blocks between the last checkpoint
actually hash-verified in this catchup run and the true chain tip. Those blocks were
never cryptographically anchored during this session.

Quick validation skips full script re-execution. It is only safe under the
invariant (documented in `quick_validate.go`) that checkpoints guarantee the
blocks are valid - and that guarantee only holds for blocks at or below a
checkpoint that was actually verified during this run, not one that merely
appears in the configured checkpoint list.

### The fix

Track a local `highestVerifiedCheckpointHeight` inside
`verifyCheckpointsAgainstHeaders`, updated only when a checkpoint's hash actually
matches a fetched header during the verification loop. Set
`catchupCtx.highestCheckpointHeight` to that verified value instead of the
globally configured maximum. This strictly narrows the quick-validation
eligibility window to only what was genuinely proven during the current catchup
run - it can only make behavior more conservative, never less safe.

### Test

Added `TestQuickValidationBoundToActuallyVerifiedCheckpoint` in
`catchup_test.go`. It configures two checkpoints - one inside the current header
range (verified) and one far above it (outside the range, never checked) - and
asserts:

1. `catchupCtx.highestCheckpointHeight` equals the verified checkpoint's height,
   not the globally configured higher one.
2. A block above the verified checkpoint but below the merely-configured higher
   checkpoint correctly falls back to normal (full) validation via
   `tryQuickValidation`.

Confirmed the test fails against the pre-fix code (`highestCheckpointHeight`
reported as the configured height 50 instead of the verified height 5), and
passes after the fix.

### Verification

- `go build ./services/blockvalidation/...`
- `go vet ./services/blockvalidation/...`
- `go test ./services/blockvalidation/... -race` - all green
- `go test ./services/blockvalidation/...` (full package suite) - all green, no regressions in existing checkpoint/quick-validation tests

### DIFF
--- a/docs/topics/services/blockValidation.md
+++ b/docs/topics/services/blockValidation.md
@@ -267,6 +267,19 @@ Before fetching full block data, the catchup process verifies checkpoints to ens
 - Approximately 10x faster than full validation for historical blocks
 - Currently disabled pending additional testing (`useQuickValidation = false`)
 
+**Behaviour change — quick-validation window is bound to what this run verified:**
+
+The eligibility bound is the highest checkpoint whose hash was actually **matched against a
+header in the current catchup range**, not the highest checkpoint present in
+`ChainCfgParams.Checkpoints`. A checkpoint that is configured but sits outside the range being
+processed proves nothing about this session, so it no longer widens the window.
+
+Operational consequence: in each catchup range, the blocks above the last in-range checkpoint —
+the tail up to the peer's tip — fall back to full validation instead of quick validation. On a
+node syncing far below the highest checkpoint this is one tail per catchup range and only
+affects sync throughput, never which blocks are accepted. Sync of the checkpointed prefix is
+unaffected as long as ranges continue to span checkpoints.
+
 **Safety Guarantees:**
 
 Checkpoint verification ensures:
@@ -450,7 +463,7 @@ Validates transactions by spending their inputs in parallel:
 The quick validation path is only applied to blocks below verified checkpoints:
 
 - **Checkpoint Verification**: During catchup, checkpoints are verified in header chain (see Section 2.2.2)
-- **Eligibility Check**: `block.Height <= highestCheckpointHeight` determines if quick validation can be used
+- **Eligibility Check**: `block.Height <= highestCheckpointHeight` determines if quick validation can be used, where `highestCheckpointHeight` is the highest checkpoint **hash-verified in the current catchup run** — blocks above it fall back to full validation (see Section 2.2.2)
 - **Trust Assumption**: Blocks below checkpoints are known to be valid, allowing optimized processing
 - **Current Status**: Quick validation currently disabled (`useQuickValidation = false`) pending additional testing
 
--- a/services/blockvalidation/BlockValidation.go
+++ b/services/blockvalidation/BlockValidation.go
@@ -1656,20 +1656,48 @@ func (u *BlockValidation) ValidateBlockWithOptions(ctx context.Context, block *m
 		// difficulty checks must run first. It also means a peer cannot make us do
 		// full tx validation for a garbage header at zero cost.
 		//
-		// Skip difficulty validation for blocks at or below the highest checkpoint:
-		// these blocks are already verified by checkpoints. block.Height is settled
-		// against the parent before this function runs (Server.deriveBlockHeight on the
-		// peer route; catchup and the operator revalidation endpoint carry authoritative
-		// heights), and BelowCheckpoint applies the mandatory height > 0 guard, so a peer
-		// cannot obtain the skip by declaring height 0 or a fabricated sub-checkpoint
-		// height. The checkpoint hash-match itself was asserted above.
+		// The header hash must meet the target its OWN nBits declares. This is enforced for
+		// every block, including below-checkpoint ones: it is what makes the chainwork the
+		// block contributes (computed from those same declared bits in
+		// getCumulativeChainWork) honestly earned, and it is the cheap gate that stops a
+		// peer buying (fail-open) subtree validation with a garbage header. Unlike the
+		// expected-nBits rule below there is no historical exception to it — no block on any
+		// chain has ever been valid without meeting its own target — so enforcing it here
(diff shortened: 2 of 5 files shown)


ordishs (approved):
Verified the defect. Pre-change `catchup.go:898-899` set `catchupCtx.highestCheckpointHeight` from the global configured maximum before the verification loop ran, and the only consumer at `catchup.go:1372` gates quick validation on `block.Height <= highestCheckpointHeight`. One in-range checkpoint match therefore opened the quick path up to height 945000 on mainnet.

The fix binds the field to the highest hash-matched checkpoint of this run only. The narrowing is safe: `verifyCheckpointsInHeaderChain` runs once per catchup, and `BuildFromHeaders` / `validateChain` prove header linkage first.

Four nits:

- `services/blockvalidation/catchup_test.go:3770` — the two load-bearing assertions use `assert.Equal` and `assert.True`. AGENTS.md requires `require`, and the `assert.Equal` on `highestCheckpointHeight` is the whole point of the test.
- `services/blockvalidation/catchup.go:56` — the field comment still reads "Highest checkpoint height for validation checks". It now means "highest checkpoint hash-verified in this catchup run".
- `services/blockvalidation/catchup.go:935` — worth a release note. The tail of each catchup range above the last in-range checkpoint now falls back to full validation.
- Out of scope, pre-existing: `BlockValidation.go:1665` and `:2309` both compute `skipDifficultyCheck := model.BelowCheckpoint(...)`, which returns true for any height in 1..945000. No exploit path traced, but it is the same trust pattern this PR narrows elsewhere.

freemans13 (approved):
Approving. I re-derived both changes independently and they hold up; everything below is documentation accuracy, nothing behavioural.

On the catchup change, the properties I checked were the ones that usually break this kind of narrowing. `verifyCheckpointsAgainstHeaders` has exactly one caller (`catchup.go:874`), the accumulator is scoped to that function, and the assignment lands after the loop. Every early return in `verifyCheckpointsInHeaderChain` bails before the call, so a stale height from a prior run could in principle survive on the context — but `useQuickValidation` is reset to `false` at line 848 on entry and is the mandatory first half of the gate at line 1381, so it can never be consumed. The field has no other reader.

On the difficulty change, the `:=` at `BlockValidation.go:1672` reuses the `var err error` declared at line 1379 in the same block rather than shadowing it, so nothing downstream is affected by the reordering. Locally: `go vet ./services/blockvalidation/...` clean, `golangci-lint` clean for the files this PR touches, and `go test ./services/blockvalidation/...` green across the package and `catchup`.

Four nits, none blocking.

**`getCumulativeChainWork` doesn't exist** — `BlockValidation.go:1662` justifies the check by saying the chainwork the block contributes is "computed from those same declared bits in `getCumulativeChainWork`". There's no such symbol anywhere in the tree. The real code is `CalculateWork` / `CalcBlockWork` in `services/blockchain/work/work.go`. The argument is right, the citation just sends the reader nowhere.

**The stated justification doesn't cover the quick-validation route** — `BlockValidation.go:1666` (and the same sentence in the test doc comment) claims `model.Block.Valid` step 1 runs the identical check "on every route that permanently accepts a block". It doesn't: `quickValidateBlockAsync` (`quick_validate.go:266`) goes to `commitBlock` (line 330) and `AddBlock` without ever calling `block.Valid`. That path *is* PoW-checked, but by `catchup.ValidateHeaderProofOfWork` at `catchup.go:1606`, on the fetched header chain — a different mechanism in a different package. The safety conclusion survives intact, only the reasoning for it is wrong. Worth correcting precisely because the automated review above repeated the claim as confirmed; in a comment written to justify a consensus change, that's the sentence someone will lean on later.

**The docs still say this feature is off, in both places the PR touched** — `blockValidation.md:268` and `:468` both read "currently disabled (`useQuickValidation = false`) pending additional testing". But `catchup.go:881` sets it `true`, and `blockvalidation_catchup_allow_quick_validation` defaults to `true` (`settings/settings.go:380`). Line 268 is one line after the paragraph this PR inserts; line 468 is five lines below the bullet it rewrites. As it stands the page carefully documents the eligibility rules of a feature it also says is switched off.

**The description only covers the first commit** — the body is entirely about the checkpoint bound. `a8bd19d2d`, always enforcing the header's own target difficulty, is an independent consensus change with no writeup. It reads as a direct answer to the fourth nit on the earlier review, which would explain it, but that review's text discusses only the catchup change. Worth extending the description so both changes are on the record under the "CONSENSUS-RELEVANT" heading, and confirming the existing approval was meant to cover both.

One observation rather than a nit: both new tests set `OptimisticMining = false`, but the danger the comment describes is specifically the optimistic-mining path, where `AddBlock` puts the block on the chain before the background `Valid()` catches it. Nothing currently pins the fix in the configuration the comment calls the production default. A third case with `OptimisticMining = true`, asserting the bad-PoW below-checkpoint block never reaches `AddBlock` as valid, would close that — but the two existing tests already cover both call sites, so this is additive.

PR #1557, opened 2026-08-10
fix(legacy): honour --nobanning in addBanScore

icellan:
## Summary

`--nobanning` (`cfg.DisableBanning`) is checked before several ban-score-driven disconnects elsewhere in `services/legacy/peer_server.go`, but `addBanScore` itself had its early-return check commented out. As a result, with `--nobanning` set, ban score still accumulated (and warnings were still logged / peers could still cross the ban threshold and get disconnected) even though the flag's intent is to disable banning entirely.

This restores the early return in `addBanScore` so the flag is honoured consistently everywhere ban score is touched.

## Test plan

- Added `TestServerPeerAddBanScoreRespectsDisableBanning`, covering both `DisableBanning: false` (score accumulates) and `DisableBanning: true` (score stays at zero). Confirmed it fails against the pre-fix code and passes after uncommenting the check.
- `go test ./services/legacy/... -race` passes.
- `go vet ./services/legacy/...` passes.

### DIFF
--- a/services/legacy/peer_server.go
+++ b/services/legacy/peer_server.go
@@ -615,10 +615,14 @@ func (sp *serverPeer) OnVersion(p *peer.Peer, msg *wire.MsgVersion) *wire.MsgRej
 	// This prevents connections from BCH/BTC/BTG and other incompatible forks
 	userAgent := msg.UserAgent
 	if !strings.Contains(userAgent, "Bitcoin SV") && !strings.Contains(userAgent, "BSV") {
-		sp.server.logger.Warnf("Rejecting and banning peer %s with non-BSV user agent: %s", sp.Peer, userAgent)
-
-		// Ban the peer to prevent repeated connection attempts from incompatible clients
-		sp.server.BanPeer(sp)
+		// Ban the peer to prevent repeated connection attempts from incompatible
+		// clients, unless banning is disabled. The peer is rejected either way.
+		if cfg.DisableBanning {
+			sp.server.logger.Warnf("Rejecting peer %s with non-BSV user agent (banning disabled): %s", sp.Peer, userAgent)
+		} else {
+			sp.server.logger.Warnf("Rejecting and banning peer %s with non-BSV user agent: %s", sp.Peer, userAgent)
+			sp.server.BanPeer(sp)
+		}
 
 		reason := "Only BSV Blockchain clients are supported"
 
--- a/services/legacy/peer_server.go
+++ b/services/legacy/peer_server.go
@@ -538,9 +538,9 @@ func (sp *serverPeer) pushAddrMsg(addresses []*wire.NetAddress) {
 // disconnected.
 func (sp *serverPeer) addBanScore(persistent, transient uint32, reason string) {
 	// No warning is logged and no score is calculated if banning is disabled.
-	// if cfg.DisableBanning {
-	//	return
-	// }
+	if cfg.DisableBanning {
+		return
+	}
 	if sp.isWhitelisted {
 		sp.server.logger.Debugf("Misbehaving whitelisted peer %s: %s", sp, reason)
 		return
--- a/services/legacy/peer_server_test.go
+++ b/services/legacy/peer_server_test.go
@@ -891,6 +891,53 @@ func TestServerPeerAddBanScoreExists(t *testing.T) {
 	assert.NotNil(t, sp.addBanScore)
 }
 
+// TestServerPeerAddBanScoreRespectsDisableBanning verifies that addBanScore
+// honours cfg.DisableBanning (the --nobanning flag) by leaving the ban score
+// untouched when banning is disabled, and that it still accumulates normally
+// when banning is enabled.
+func TestServerPeerAddBanScoreRespectsDisableBanning(t *testing.T) {
+	// cfg is a package-level variable read by addBanScore; save/restore it so
+	// this test doesn't leak state into other tests in the package.
+	origCfg := cfg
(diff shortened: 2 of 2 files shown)


ordishs (approved):
Verified the defect. `OnGetData` at `services/legacy/peer_server.go:1438` accumulates ban score with no flag check, and can reach `BanPeer` plus a disconnect while banning is disabled. The early return is safe: `newServer` assigns `cfg` before any peer exists, so the new `cfg` dereference cannot panic on a production path.

Three nits:

- The title says the fix honours `--nobanning`, but the legacy command line parser is dead code. `newConfigParser` is commented out at `config.go:333`, and the ini parse at `409-419`. The flag arrives by another route, so the title is misleading.
- `services/legacy/peer_server.go:618-621` — the flag is now honoured for ban score, but not for banning as a whole. `OnVersion` rejects a non-BSV user agent and calls `sp.server.BanPeer(sp)` directly, with no flag check.
- The new `TestServerPeerAddBanScoreRespectsDisableBanning` uses `assert.Zero` / `assert.NotZero`. AGENTS.md requires `require`, although the rest of that file uses `assert`.

freemans13 (approved):
Reviewed at `fea0507e14016bdaf1565ffc149213a32c12e095`. Approving — the fix is correct and complete, and everything I could run locally is green. Comments below are non-blocking, with one request to update the PR description.

## What the change does

`addBanScore` is the funnel every misbehaviour path goes through, and its `if cfg.DisableBanning { return }` guard was sitting behind `//`. So with `--nobanning` set, ban score still accumulated, warnings were still logged, and a peer crossing the threshold was still banned and disconnected — the flag did roughly nothing on that path. The first commit uncomments the guard; the second does the same job at the other automatic ban site, the non-BSV user-agent rejection in `OnVersion`.

I checked completeness rather than taking it on faith. There are exactly two places in the legacy service that ban a peer automatically — `peer_server.go:569` (the threshold breach inside `addBanScore`) and `peer_server.go:624` (the fork-client rejection) — and both are now guarded. The gRPC `BanPeer` in `Server.go:488` is correctly left alone, since that's an operator explicitly banning someone rather than the node deciding to. `handleBanPeerMsg` does *not* check the flag itself, so guarding at the call sites is genuinely required and not belt-and-braces.

Two things I verified because they would otherwise have been quiet bugs:

- The `OnVersion` change still returns the reject message, and `peer/peer.go:2744-2748` turns a non-nil reject into a write plus a negotiation error. So a Bitcoin Cash peer is still disconnected under `--nobanning`, just not added to the ban list — exactly what the new comment claims.
- The comment in `enforceNodeBloomFlag` at line 1667 ("Even though the `addBanScore` function already examines whether or not banning is enabled…") was *stale* while the check was commented out. This PR silently makes it true again.

I also confirmed the test's claim about wiring: `legacy_config_DisableBanning` does reach the field through the reflection loop in `setConfigValuesFromSettings`, and the `nobanning` struct tag really is inert. The setting isn't set in any committed settings context, so no existing deployment changes behaviour by default.

## Verification

| Check | Result |
|---|---|
| `go vet ./services/legacy/` | clean |
| `golangci-lint run ./services/legacy/` | 0 issues |
| `go test -race -count=1 ./services/legacy/` | ok, 3.955s |
| Both new tests in isolation, `-race` | pass |

## Non-blocking comments

**1. With `--nobanning`, misbehaviour warnings now vanish entirely.** The early return at `peer_server.go:541` sits above all logging, so operators lose the `Misbehaving peer … ban score increased to N` line they used to get. Some people run `--nobanning` precisely to *watch* misbehaviour without acting on it, and this takes that away. Worth considering a `Debugf` before the return. Caveat: that would dereference `sp.server`, which the new `addBanScore` test leaves nil, so the test would need the fix in comment 4 first. This matches btcd upstream behaviour, so I'm happy either way — just flagging that it's a real loss for someone.

**2. The PR description doesn't cover the second commit.** The body only describes `addBanScore` and one test, but `fea0507e` adds the `OnVersion` change and a second test. Could you update it before merge so the changelog reads correctly?

**3. One behavioural consequence worth stating explicitly in the body.** The only remaining consumer of ban-score-driven disconnection is the getdata-flood throttle at `peer_server.go:1442`. Under `--nobanning`, a peer can now spam oversized inventory queries indefinitely with no disconnect and no log. `enforceNodeBloomFlag` is unaffected, since it disconnects regardless. This is what the flag means and it matches btcd, so I'd ship it — but it belongs in the description as a stated consequence rather than being discovered in production later.

**4. Small test fragility.** In `peer_server_test.go:930`, `sp := &serverPeer{}` has a nil `server`. It only survives because `BanThreshold: 100` gives a warn threshold of 50, and a score of 1 never reaches the logging branch. Any future log line added before that point turns a clean test failure into a nil-pointer panic. Giving it `server: &server{logger: ulogger.TestLogger{}}` is one line and removes the trap.

**5. Optional: the test could go one step deeper.** It asserts the score stays at zero, which is sufficient given the early return, but it doesn't exercise the behaviour that actually motivated the fix — a peer crossing `BanThreshold` and getting banned anyway. Driving the score past the threshold with `DisableBanning: true` and asserting `banPeers` stays empty would mirror the shape of your `OnVersion` test nicely.

One thing worth calling out positively: both new tests save and restore the package-level `cfg`. The pre-existing test at `peer_server_test.go:429` does not — it leaks `cfg = &config{whitelists: nil}` into whatever runs next. Not this PR's problem, and no test in the package uses `t.Parallel()` so there's no race today, but it's a nearby hazard someone will eventually trip over.

PR #1558, opened 2026-08-10
docs(operator): add monitoring guide, link metrics reference, fix legacy CR resource snippet

icellan:
## Summary

- Adds a deployment-agnostic monitoring guide (`docs/howto/miners/minersHowToMonitoring.md`), wired into the nav under Miners, covering the Grafana dashboards shipped in-repo (Teranode Service Overview, Aerospike batch-index/namespace/latency, BlockAssembler State Monitoring), the metrics that distinguish a healthy node from a degraded one, and a starter-alerting note (alert rules are operator-defined; none ship in-repo).
- Links `references/prometheusMetrics.md` from the Docker and Kubernetes miner installation docs, and from the new monitoring guide. Previously it was reachable only from the References nav and other reference docs, not from the miner-facing installation path.
- Fixes the legacy-service resource snippet in the Kubernetes installation doc: the `cpu: 100m` / `memory: 256Mi` / `imagePullPolicy: Never` values are Minikube/local-dev defaults, not appropriate for testnet/mainnet. Added a caveat and a pointer to the shipped mainnet CR (`cpu: 2` / `memory: 64Gi`) for real deployments.

## Out of scope

The audit also recommended converting the 5 "Production Considerations" bullets in the Kubernetes installation doc into full procedures. That's a larger doc-authoring effort and is left as a follow-up; this PR only turns the monitoring bullet into a link to the new guide.

## Test plan

- [x] Verified all new/updated relative doc links resolve to existing files
- [x] `markdownlint` passes on the new doc
- [x] Verified the legacy CR values against `deploy/kubernetes/teranode/teranode-cr-mainnet.yaml`

### DIFF
--- a/compose/grafana/dashboards/README.md
+++ b/compose/grafana/dashboards/README.md
@@ -9,53 +9,61 @@ Comprehensive monitoring dashboard for BlockAssembler internal state transitions
 ### Panels
 
 #### 1. State Timeline
+
 - **Type:** State Timeline
 - **Shows:** Visual timeline of state changes over time
 - **Colors:**
-  - Green: Running (normal operation)
-  - Blue: GetMiningCandidate (serving miners)
-  - Purple: BlockchainSubscription (processing new block)
-  - Light Blue: MovingUp (advancing chain tip)
-  - Orange: Resetting (full reset)
-  - Red: Reorging (handling chain reorg)
-  - Yellow: Starting (initialization)
+    - Green: Running (normal operation)
+    - Purple: BlockchainSubscription (processing new block)
+    - Light Blue: MovingUp (advancing chain tip)
+    - Orange: Resetting (full reset)
+    - Red: Reorging (handling chain reorg)
+    - Yellow: Starting (initialization)
+    - Blue: Reconciling (catching the tip up after startup or a missed notification)
 
 #### 2. State Durations (P50/P95/P99)
+
 - **Type:** Time Series
 - **Shows:** Percentile distribution of time spent in each state
 - **Use:** Identify performance bottlenecks
 - **Example:** If P95 BlockchainSubscription is 5s, 95% of blocks process in under 5s
 
 #### 3. State Transitions (per second)
+
 - **Type:** Time Series
 - **Shows:** Rate of state transitions
 - **Format:** `from → to`
 - **Use:** Understand state change frequency and patterns
 
 #### 4. Current State
+
 - **Type:** Stat (single value)
 - **Shows:** Current state with color coding
 - **Updates:** Real-time (5s refresh)
 
 #### 5. Time in Current State
+
 - **Type:** Stat with gauge
 - **Shows:** How long in current state
 - **Thresholds:**
-  - Green: < 5s (normal)
-  - Yellow: 5-10s (slow)
-  - Red: > 10s (investigate)
+    - Green: < 5s (normal)
+    - Yellow: 5-10s (slow)
+    - Red: > 10s (investigate)
 
 #### 6. State Distribution (Last 5 min)
+
 - **Type:** Pie Chart
 - **Shows:** Percentage of time in each state
 - **Use:** Understand where BlockAssembler spends most time
 
 #### 7. State Entries (Last 5 min)
+
 - **Type:** Stat (horizontal)
 - **Shows:** How many times each state was entered
 - **Use:** Identify high-frequency states
 
 #### 8. State Transition Matrix
+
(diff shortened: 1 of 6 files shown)


ordishs (approved):
Verified the gap this fills: `docs/howto/miners/` had no monitoring guide, and the miner install docs never linked `docs/references/prometheusMetrics.md`. I checked every metric name in the new text against the Go sources, and most of the guide is accurate. Two text errors would mislead an operator.

**P2 — the dashboard table promises automatic provisioning that does not happen. `docs/howto/miners/minersHowToMonitoring.md:14-24`**

The table lists `compose/grafana/dashboards/blockassembly-state.json` with the four `deploy/docker/base` dashboards, then says the dashboards are provisioned automatically on the Docker quickstart path. That is false for the BlockAssembler dashboard.

`deploy/docker/monitoring/docker-compose.yml` extends the grafana service from `deploy/docker/base/docker-services.yml`, which mounts only `./grafana_dashboards` (`deploy/docker/base/docker-services.yml:232`). That directory holds the two teranode dashboards and the two aerospike dashboards. `blockassembly-state.json` is mounted only by `compose/docker-compose-ss.yml:485` and `test/docker-compose-host.yml:435`. The operator opens Grafana, sees no BlockAssembler dashboard, and gets no import instruction.

Recommendation: add a column that names the stack which provisions each dashboard, and state that `blockassembly-state.json` must be imported by hand.

**P2 — the state metrics are numeric gauges, so the example alerts never fire. `docs/howto/miners/minersHowToMonitoring.md:45-58` and `:92-103`**

The guide treats `teranode_blockchain_fsm_current_state` and `teranode_blockassembly_current_state` as string-valued, and uses capitalised state names.

`services/blockchain/fsm.go:83` sets the gauge with `float64(blockchain_api.FSMStateType_value[e.Dst])`; `blockchain_api.pb.go:142` maps IDLE 0, RUNNING 1, CATCHINGBLOCKS 2. An alert written as `fsm_current_state != "RUNNING"` is not valid PromQL against a numeric gauge.

`services/blockassembly/BlockAssembler.go:915` observes `state_duration_seconds` with the `StateStrings` values: `starting`, `running`, `resetting`, `blockchainSubscription`, `reorging`, `movingUp`, `reconciling`. A query with `state="Running"` or `state="MovingUp"` returns an empty series.

Recommendation: show the numeric mapping for both gauges, use the exact lower-camel label values in any `state=` selector, and write the alerts as numeric comparisons.

freemans13 (approved):
Approving. I verified the guide's factual claims against the source at 765c8a94a rather than reading it as prose, and the accuracy is high.

Checked and correct: all seventeen metric names resolve to real `promauto` registrations with the right namespace and subsystem (including the easy-to-miss `teranode_tx_meta_cache_hits`/`_misses` under subsystem `tx_meta_cache`, which are kept fresh by a ticker, so the hit-ratio advice is sound); the label sets match (`state_transitions_total{from,to}` and `state_duration_seconds{state}` at services/blockassembly/metrics.go:344,355), and the values really do come from `StateStrings`, so the warning that `{state="Running"}` matches nothing is worth having; the block assembly table (0/1/2/4/5/6/7, 3 unused) matches BlockAssembler.go:61-88 exactly; the dashboard JSON drops the stale `GetMiningCandidate` mapping, adds `Reconciling`, and still parses; the provisioning story holds (base/docker-services.yml:231-232 mounts the four-dashboard directory that mainnet/testnet/monitoring all extend, while blockassembly-state.json is only mounted by docker-compose-ss.yml and test/docker-compose-host.yml, so "import by hand" is right); ports 3005/9090, every relative link, and the mainnet CR's `cpu: 2` / `memory: 64Gi` for legacy all check out.

One thing still outstanding, already flagged inline by the review bot after the last commit and not yet addressed — docs/howto/miners/minersHowToMonitoring.md lines 65-67. The gauge is not written on transitions only: `prometheusBlockchainFSMCurrentState` is registered at service construction (Server.go:173) and, being a plain `NewGauge`, exports `0` from that moment; it is then set from the persisted state at Server.go:442 before any transition, and only after that on each transition at fsm.go:83. So the series is never absent, the suggested `absent(...)` alert would never fire, and — the part that actually costs an operator — a `0` always means a genuine `IDLE`, not "no data yet". Worth a follow-up line along those lines.

Minor, take it or leave it: the FSM table lists 0/1/2 but doesn't note that 3 is reserved (`LEGACYSYNCING`, removed), which reads slightly asymmetrically next to the block assembly table that does call out its unused 3.

PR #1559, opened 2026-08-10
Fix service-inventory doc drift: add Pruner overlay-service section, correct P2P bootstrap description

icellan:
## Summary

- Adds a "Pruner Service" subsection to the Overlay Services section of
  `docs/topics/architecture/teranode-microservices-overview.md`. The daemon starts
  six overlay services (`daemon/daemon_services.go`), but the doc only documented
  five — Pruner was missing entirely.
- Rewrites the stale description in `docs/topics/services/p2p.md` that referred to
  a standalone "p2p bootstrap service" managing the private network. That service
  doesn't exist; bootstrapping is a configuration option (`p2p_bootstrap_peers`)
  read by the P2P service itself at startup.
- Drops the same phantom "P2P Bootstrap Service" from `docs/references/glossary.md`
  (replaced with a "Bootstrap peers" entry describing the actual config option), and
  fixes the libp2p entry's reference to it.
- Adds a guard test (`daemon/daemon_service_inventory_test.go`) asserting the
  daemon's service dispatch (`daemon_services.go`) matches the `services/<name>`
  directory layout in both directions. This turns drift between the dispatch table
  and the services/ layout into an automatic CI failure. It does not check the
  architecture docs — those still need a manual pass.

## Test plan

- [x] `go test ./daemon/...` passes
- [x] `go vet ./daemon/...` passes
- [x] Verified the guard test actually catches drift: temporarily added a fake
      service name to the test's list (no matching `services/` directory), confirmed
      the test failed with a clear error message, then reverted the temporary change
- [x] Visually reviewed the doc edits for formatting/style consistency with the
      surrounding sections

### DIFF
--- a/daemon/daemon_service_inventory_test.go
+++ b/daemon/daemon_service_inventory_test.go
@@ -1,0 +1,49 @@
+package daemon
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestServiceInventory_MatchesServicesDirectory guards against documentation and
+// code drift: every service name the daemon knows how to start (the "Formal"
+// service-name constants passed to shouldStart/AddService in daemon_services.go)
+// must have a corresponding source package under services/<name>. If a service is
+// added, renamed, or removed here without updating the services/ layout (or vice
+// versa), this test fails instead of silently drifting out of sync with the docs.
+func TestServiceInventory_MatchesServicesDirectory(t *testing.T) {
+	// The set of "Formal" service-name constants the daemon can start, as used in
+	// startServices() (daemon/daemon_services.go). Keep this list in sync with that
+	// function's shouldStart/starters list.
+	daemonServiceNames := []string{
+		serviceAlertFormal,
+		serviceAssetFormal,
+		serviceBlockAssemblyFormal,
+		serviceBlockPersisterFormal,
+		serviceBlockValidationFormal,
+		serviceBlockchainFormal,
+		serviceLegacyFormal,
+		serviceNameP2PFormal,
+		servicePropagationFormal,
+		servicePrunerFormal,
+		serviceRPCFormal,
+		serviceSubtreeValidationFormal,
+		serviceUtxoPersisterFormal,
+		serviceValidatorFormal,
+	}
+
+	servicesDir := filepath.Join("..", "services")
+
+	for _, name := range daemonServiceNames {
+		dirName := strings.ToLower(name)
+		pkgPath := filepath.Join(servicesDir, dirName)
+
+		info, err := os.Stat(pkgPath)
+		require.NoError(t, err, "daemon can start service %q but services/%s does not exist", name, dirName)
+		require.True(t, info.IsDir(), "services/%s exists but is not a directory", dirName)
+	}
+}
--- a/daemon/daemon_service_inventory_test.go
+++ b/daemon/daemon_service_inventory_test.go
@@ -12,13 +12,20 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// daemonServicesFile is the source file whose shouldStart() call sites define the
-// authoritative set of services the daemon can start.
-const daemonServicesFile = "daemon_services.go"
+// daemonPackageDir is the directory scanned for d.shouldStart(...) call sites,
+// which define the authoritative set of services the daemon can start. Scanning
+// the whole package (rather than just daemon_services.go) catches dispatch calls
+// that live elsewhere, such as daemon.go's "wait_for_postgres" switch.
+const daemonPackageDir = "."
+
+// daemonReceiverIdent is the receiver name a call must use to be treated as a
+// daemon method call (e.g. d.shouldStart(...)), so that an unrelated same-named
+// method or field on another type in the package isn't folded into the inventory.
+const daemonReceiverIdent = "d"
 
 // daemonServiceNames maps the identifier of every "Formal" service-name constant
 // that startServices() dispatches to that constant's value. The identifiers are
-// cross-checked against the real shouldStart() call sites in daemonServicesFile, so
+// cross-checked against the real shouldStart() call sites in the daemon package, so
 // this map cannot drift from the code in either direction.
 var daemonServiceNames = map[string]string{
 	"serviceAlertFormal":             serviceAlertFormal,
@@ -38,19 +45,23 @@ var daemonServiceNames = map[string]string{
 }
 
 // nonServiceShouldStartArgs holds shouldStart() arguments that are command-line
-// switches rather than services, and so are exempt from the inventory.
+// switches rather than services, and so are exempt from the inventory. Both
+// identifier names (e.g. serviceHelp) and string-literal switches (e.g.
+// wait_for_postgres) are looked up here, keyed by their literal text.
 var nonServiceShouldStartArgs = map[string]struct{}{
 	// -help only prints usage; it starts nothing and has no services/ package.
 	"serviceHelp": {},
+	// wait_for_postgres (daemon.go) is a startup gate, not a service.
+	"wait_for_postgres": {},
 }
 
-// TestServiceInventory_MatchesServicesDirectory guards against documentation and
-// code drift in both directions: the set of services startServices() can dispatch is
-// derived from the actual shouldStart() call sites in daemon_services.go, so a
-// service added to (or removed from) that function without being added to (or
-// removed from) daemonServiceNames fails the test. Every inventoried service must
-// also have a source package under services/<name>, which is what the architecture
-// docs enumerate.
+// TestServiceInventory_MatchesServicesDirectory guards the daemon's service dispatch
+// against the services/ layout in both directions: the set of services
+// startServices() can dispatch is derived from the actual shouldStart() call sites
+// in the daemon package, so a service added to (or removed from) that dispatch
+// without being added to (or removed from) daemonServiceNames fails the test. Every
+// inventoried service must also have a source package under services/<name>. This
+// does not check the architecture docs — those still need a manual pass.
 func TestServiceInventory_MatchesServicesDirectory(t *testing.T) {
 	dispatched := shouldStartServiceIdents(t)
 
@@ -58,8 +69,8 @@ func TestServiceInventory_MatchesServicesDirectory(t *testing.T) {
 	for ident := range dispatched {
 		_, listed := daemonServiceNames[ident]
 		require.True(t, listed,
-			"%s dispatches shouldStart(%s) but %s is missing from daemonServiceNames: add it here and add its services/<name> package",
-			daemonServicesFile, ident, ident)
+			"daemon package dispatches shouldStart(%s) but %s is missing from daemonServiceNames: add it here and add its services/<name> package",
+			ident, ident)
 	}
 
 	// Direction 2: nothing in the inventory may be a service the daemon no longer
@@ -67,8 +78,8 @@ func TestServiceInventory_MatchesServicesDirectory(t *testing.T) {
 	for ident := range daemonServiceNames {
 		_, ok := dispatched[ident]
 		require.True(t, ok,
-			"daemonServiceNames lists %s but %s never passes it to shouldStart: remove it here or restore the dispatch",
-			ident, daemonServicesFile)
+			"daemonServiceNames lists %s but the daemon package never passes it to shouldStart: remove it here or restore the dispatch",
+			ident)
 	}
 
 	// Every dispatched service must have a source package under services/<name>.
@@ -84,46 +95,77 @@ func TestServiceInventory_MatchesServicesDirectory(t *testing.T) {
 	}
 }
 
-// shouldStartServiceIdents parses daemonServicesFile and returns the identifiers
-// passed as the first argument to every d.shouldStart(...) call, minus the
-// non-service switches. Deriving the set from the source rather than restating it
-// keeps the guard bidirectional.
+// shouldStartServiceIdents parses every non-test .go file in daemonPackageDir and
+// returns the identifiers passed as the first argument to every
+// daemonReceiverIdent.shouldStart(...) call, minus the non-service switches.
+// Deriving the set from the source rather than restating it keeps the guard
+// bidirectional. Scanning the whole package (rather than just daemon_services.go)
+// catches dispatch calls that live elsewhere. shouldStart must be called directly
+// (e.g. d.shouldStart(...)); a method value taken as ss := d.shouldStart and
+// invoked as ss(...) is not visible to this AST match.
 func shouldStartServiceIdents(t *testing.T) map[string]struct{} {
 	t.Helper()
 
 	fset := token.NewFileSet()
 
-	file, err := parser.ParseFile(fset, daemonServicesFile, nil, 0)
-	require.NoError(t, err, "failed to parse %s", daemonServicesFile)
+	entries, err := os.ReadDir(daemonPackageDir)
+	require.NoError(t, err, "failed to read %s", daemonPackageDir)
 
 	idents := make(map[string]struct{})
+	found := false
 
-	ast.Inspect(file, func(n ast.Node) bool {
-		call, ok := n.(*ast.CallExpr)
-		if !ok {
-			return true
+	for _, entry := range entries {
+		name := entry.Name()
+		if entry.IsDir() || !strings.HasSuffix(name, ".go") || strings.HasSuffix(name, "_test.go") {
+			continue
 		}
 
-		sel, ok := call.Fun.(*ast.SelectorExpr)
-		if !ok || sel.Sel.Name != "shouldStart" || len(call.Args) == 0 {
+		path := filepath.Join(daemonPackageDir, name)
+
+		file, err := parser.ParseFile(fset, path, nil, 0)
+		require.NoError(t, err, "failed to parse %s", path)
+
+		ast.Inspect(file, func(n ast.Node) bool {
+			call, ok := n.(*ast.CallExpr)
+			if !ok {
+				return true
+			}
+
+			sel, ok := call.Fun.(*ast.SelectorExpr)
+			if !ok || sel.Sel.Name != "shouldStart" || len(call.Args) == 0 {
+				return true
+			}
+
+			recv, ok := sel.X.(*ast.Ident)
+			if !ok || recv.Name != daemonReceiverIdent {
+				return true
+			}
+
+			found = true
+
+			switch arg := call.Args[0].(type) {
+			case *ast.Ident:
+				if _, exempt := nonServiceShouldStartArgs[arg.Name]; exempt {
+					return true
+				}
+
+				idents[arg.Name] = struct{}{}
+			case *ast.BasicLit:
+				lit := strings.Trim(arg.Value, `"`)
+				_, exempt := nonServiceShouldStartArgs[lit]
+				require.True(t, exempt,
+					"shouldStart(%s) at %s is not in nonServiceShouldStartArgs: add it there if it isn't a service",
+					arg.Value, fset.Position(call.Lparen))
+			default:
+				t.Fatalf("shouldStart is called with an unrecognized argument kind at %s: this guard can only inventory constant identifiers and string literals", fset.Position(call.Lparen))
+			}
+
 			return true
-		}
+		})
+	}
 
-		ident, ok := call.Args[0].(*ast.Ident)
-		require.True(t, ok,
-			"shouldStart is called with a non-constant service name at %s: this guard can only inventory constant identifiers",
-			fset.Position(call.Lparen))
-
-		if _, exempt := nonServiceShouldStartArgs[ident.Name]; exempt {
-			return true
-		}
-
-		idents[ident.Name] = struct{}{}
-
-		return true
-	})
-
-	require.NotEmpty(t, idents, "found no shouldStart call sites in %s: has the service dispatch moved?", daemonServicesFile)
+	require.True(t, found, "found no %s.shouldStart call sites in %s: has the service dispatch moved?", daemonReceiverIdent, daemonPackageDir)
+	require.NotEmpty(t, idents, "found no service identifiers among %s.shouldStart call sites in %s", daemonReceiverIdent, daemonPackageDir)
 
 	return idents
 }
--- a/daemon/daemon_service_inventory_test.go
+++ b/daemon/daemon_service_inventory_test.go
@@ -1,6 +1,9 @@
 package daemon
 
 import (
+	"go/ast"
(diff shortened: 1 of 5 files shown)


ordishs (approved):
Verified both claims. `daemon/daemon_services.go:74` starts the Pruner service and `services/pruner` exists, so the missing section was real drift. No P2P bootstrap service exists in `services/` or `cmd/`, so the old sentence was false. The new text is accurate against the code, including the BlockPersisted-to-Block fallback at `services/pruner/server.go:6-9`.

**P2 — the same drift survives in the sibling architecture doc. `docs/topics/architecture/teranode-overall-system-design.md:126`**

Line 126 lists overlay services as: Block Persister, UTXO Persister, P2P, **P2P Bootstrap**, Legacy, RPC. A tree-wide grep finds no bootstrap service package in `services/` or `cmd/`; only `docs/topics/services/img/plantuml/bootstrap` remains. A grep for "Pruner" in this file returns nothing, although `daemon/daemon_services.go:74` starts `servicePrunerFormal`.

AGENTS.md names this file as the architecture reference, so a reader still gets the exact drift this PR sets out to fix.

Recommendation: replace "P2P Bootstrap Service" with "Pruner Service" in that list, so both architecture docs match the daemon.

Nit: the new daemon guard test asserts that each daemon service name has a matching `services/<name>` package. That does not guard the doc drift its comment promises.

ctnguyen (changes requested):
Review at commit [4f8625a](https://github.com/icellan/teranode/commit/4f8625a855663b142636fac261224e37c1700b87)

Good, well-targeted cleanup, and both core claims hold up: the daemon really does start Pruner, and there is no P2P bootstrap service anywhere in the tree. The head commit's bidirectional rewrite of the guard is a real improvement over restating the service list by hand — deriving the dispatch set with `go/ast` closes the one-directional hole, and it passes as written. The findings below are mostly about drift the PR left behind and two soft spots in the guard.

## Blocking issues

### ChiR1 — The nonexistent P2P Bootstrap Service survives in the glossary

*Partial overlap with [@ordishs's point P2 (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1559#pullrequestreview-4907183083).*

**Problem:** `docs/references/glossary.md` still documents the service as real, in two places: line 69 defines `**P2P Bootstrap Service**: Helps new nodes discover peers and join the network, using libp2p and Kademlia.`, and line 55 attributes libp2p itself to it (`used by Teranode's P2P Bootstrap Service for peer discovery`). After this PR the glossary is the only *prose* doc left containing the string "Bootstrap Service"; grepping `docs/` unfiltered turns up two more hits in a diagram asset — `docs/topics/services/img/plantuml/bootstrap/bootstrap_grpc_request.puml:25` (`participant "gRPC Server (Bootstrap Service)"`) and the same text baked into its rendered `.svg`. That asset sits in a directory of nine orphaned diagrams (`bootstrap_startup`, `bootstrap_grpc_ping`, `bootstrap_connect_websocket`, `bootstrap_high_level_overview`, …) documenting the phantom service's gRPC/WebSocket API, including a `GetNodes()` call; nothing under `docs/` embeds any of them — no `.md` references `plantuml/bootstrap`. This extends rather than repeats ordishs' P2: he found the phantom service in the sibling architecture doc and the head commit fixed exactly that (verified — `teranode-overall-system-design.md:113-128` now lists six overlay services matching the overview's `3.1`–`3.6`, and neither architecture doc mentions "bootstrap" any more). The glossary is a third location he didn't cover, and the only one that gives the phantom service a formal definition rather than a list entry.

**Why it matters:** A glossary entry reads as authoritative. Someone told the service exists will go looking for `services/p2pbootstrap`, a deployment manifest, or a port for it, and find nothing. Leaving it in means the drift named in the PR title is only partly fixed, and the next person to hit it has no reason to think it was already investigated.

**Fix:** Two edits in the same style as the p2p.md change. Line 55: "used by Teranode's P2P Service for peer discovery". Line 69: drop the entry, or replace it with the thing that actually exists:

```markdown
**Bootstrap peers**: Initial DHT/relay entry points, configured via `p2p_bootstrap_peers`
and dialled by the P2P Service at startup (libp2p/Kademlia).
```

The glossary text is the part I'd hold the merge for — it's what readers actually see. The orphaned `plantuml/bootstrap/` directory is a separate, larger cleanup (delete it, since nothing links to it) and is fine as a follow-up; I'm not asking you to bundle 18 asset files into a prose fix.

## Non-blocking issues

### ChiR2 — The guard's comments still claim doc coverage it doesn't have

*Partial overlap with [@ordishs's nit (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1559#pullrequestreview-4907183083).*

**Problem:** The rewrite fixed the direction ordishs was pointing at, but the overstatement he objected to survived it. `daemon/daemon_service_inventory_test.go:47` still opens with "guards against documentation and code drift", and line 58 still ends with "which is what the architecture docs enumerate". The test opens no file under `docs/` — it parses `daemon_services.go` and stats directories under `services/`. Delete `### 3.6 Pruner Service` from the overview and the test still passes; that is precisely the drift this PR was filed to fix. The PR description carries the same claim ("turns future service additions/removals … into an automatic CI failure, rather than something that has to be caught by manual doc review"), which is not true for the doc half.

**Why it matters:** An overstated guard is worse than no guard: the next person adding a service reads the comment, trusts CI to catch the doc gap, and ships the exact drift this PR is cleaning up. The check itself is worth keeping — it just needs to be described as what it is.

**Fix:** Say what it guards and stop there, e.g. "guards the daemon's service dispatch against the `services/` layout in both directions: the set of names passed to `shouldStart()` in `daemon_services.go` must match `daemonServiceNames` exactly, and every inventoried name must have a `services/<name>` package. This does not check the architecture docs — those still need a manual pass." Then trim the corresponding PR-description bullet.

### ChiR3 — New p2p.md text sets up a private-network trap it doesn't close

**Problem:** The replacement sentence at `docs/topics/services/p2p.md:41` opens with "The p2p peers can form a private network" and then explains that the service "reads the `p2p_bootstrap_peers` setting at startup to obtain initial DHT entry points". It never says what an empty value does. To be precise about the default, since it matters: `settings.conf:864` ships a non-empty BSVA-managed value (`/dnsaddr/${network}.bootstrap.teranode.bsvb.tech`), so out of the box operators are *not* on public IPFS peers — the `default:""` struct tag at `settings/p2p_settings.go:28` is doc-generator metadata, and the `[]string{}` at `settings/settings.go:519` only applies if the key were absent, which it isn't. The trap is on the private-network path this sentence invites: the `BootstrapPeers` longdesc explicitly recommends "Leave empty for private networks using only StaticPeers", and blanking it lands in `getBootstrapAndRelayPeers` (`go-p2p-message-bus@v0.1.17/client.go:207-227`), which substitutes `dht.GetDefaultBootstrapPeerAddrInfos()` — the default public IPFS bootstrap peers — after which `NewClient` calls `connectToManagedPeers(..., bootstrapPeerKind, ...)` unconditionally, i.e. even under `p2p_dht_mode = off`. Two smaller inaccuracies come from the same framing: bootstrap peers aren't only DHT entry points (they double as relay/circuit peers, `config.go:72-78`), and in `off` mode they are the only peer-address source.

**Why it matters:** The sentence is the natural landing spot for someone setting up a private network, and following the advice one config file over puts them on public IPFS infrastructure. One clause closes the gap, and getting it right is the point of this PR.

**Fix:** Extend the sentence:

```markdown
… reads the `p2p_bootstrap_peers` setting at startup to obtain initial DHT and relay entry
points for peer discovery. Blanking it does not isolate the node — the client then falls back
to the default public IPFS bootstrap peers — so a private network should point it at its own
bootstrap servers, alongside `p2p_static_peers` and `p2p_dht_mode = off`.
```

The "leave empty" recommendation in the `BootstrapPeers` longdesc is the root of this and deserves a follow-up.

### ChiR4 — `Block` notifications are a configured mode, not a fallback

**Problem:** The new bullet in `teranode-microservices-overview.md:432` reads "Respond to `BlockPersisted` (or `Block`, as a fallback) notifications instead of polling". There is no runtime fallback. `pruner_block_trigger` selects one of two mutually exclusive modes and each handler drops the other notification type outright: `services/pruner/server.go:157-159` skips `BlockPersisted` under `OnBlockMined`, and `server.go:201-204` skips `Block` under `OnBlockPersisted`. `triggerInitialPruning` (`server.go:321-322`) reads `lastPersistedHeight`, which stays 0 without a Block Persister, so it returns at the `currentHeight == 0` guard. Net effect on the default (`OnBlockPersisted`) with no Block Persister deployed: pruning never triggers at all. To be fair, the PR inherited this — the wording matches `docs/topics/services/pruner.md:88` and the package comment at `services/pruner/server.go:6-7`, both of which claim an automatic fallback; the settings `longdesc` for `pruner_block_trigger` gets it right ("Use when Block Persister is not available").

**Why it matters:** "Fallback" tells an operator the service degrades gracefully; in reality the wrong combination silently disables pruning, which is the unbounded-growth failure the same paragraph warns about. Copying the claim into a second doc also makes it harder to correct later.

**Fix:** In this PR, drop the misleading framing:

```markdown
- Respond to `BlockPersisted` or `Block` notifications — selected by `pruner_block_trigger`
  — instead of polling
```

The source claims at `services/pruner/server.go:6-7` and `docs/topics/services/pruner.md:88`/`132` deserve a separate fix.

### ChiR5 — The guard's AST derivation is narrower and looser than its comments imply

**Problem:** Four gaps in `shouldStartServiceIdents`, none of which bite today but all of which weaken the "cannot drift in either direction" claim.

1. *Single-file scope.* It parses only `daemon_services.go` (line 96), yet `shouldStart` is already called from elsewhere in the package: `daemon/daemon.go:284` does `d.shouldStart("wait_for_postgres", args)`. A service dispatched from another file is invisible to Direction 1 and passes silently; conversely, *moving* an existing dispatch out of `daemon_services.go` makes Direction 2 fail with "never passes it to shouldStart: remove it here or restore the dispatch", pointing the developer at the wrong repair.
2. *Receiver-blind matching.* The walk accepts any `*ast.SelectorExpr` whose selector is named `shouldStart` (lines 107-109), without checking the receiver resolves to the daemon. The name is already overloaded in this very file — `serviceStarter.shouldStart` is a `bool` field (line 44), read at line 129. That field is not called, so nothing matches today, but any same-named method or func-typed field on another type in the package would be silently folded into the inventory.
3. *Method values are missed.* A refactor to `ss := d.shouldStart; ss(serviceFoo, args)` produces no matching `CallExpr`, so `serviceFoo` never enters the dispatched set — Direction 1 goes quiet and Direction 2 then fails for every service handled that way.
4. *Exemptions can't cover literals.* The `*ast.Ident` assertion at lines 112-115 runs *before* the `nonServiceShouldStartArgs` lookup at line 117, so a `wait_for_postgres`-style string-literal switch added to `daemon_services.go` hard-fails the guard, the exemption map cannot rescue it, and the failure message doesn't mention the map exists.

(The CWD-relative paths are fine, incidentally: `go test` runs with the package directory as CWD, so both `daemon_services.go` and `filepath.Join("..", "services")` resolve.)

**Why it matters:** The comment at lines 15-16 calls these call sites "the authoritative set of services the daemon can start", and Direction 2's whole value rests on the derived set being complete and correctly attributed. Each gap turns a future refactor into either a silent pass or an opaque failure with a misleading message.

**Fix:** Walk every non-`_test.go` file in the package instead of one filename; gate on the receiver (`sel.X` being an ident whose type is the daemon, or at minimum the ident `d`); and keep the exemption lookup ahead of the assertion:

```go
if ident, ok := call.Args[0].(*ast.Ident); ok {
    if _, exempt := nonServiceShouldStartArgs[ident.Name]; !exempt {
        idents[ident.Name] = struct{}{}
    }
    return true
}
// String-literal switches are allowed only if explicitly exempted.
lit, ok := call.Args[0].(*ast.BasicLit)
require.True(t, ok, "shouldStart called with a non-constant, non-literal argument at %s", fset.Position(call.Lparen))
require.Contains(t, nonServiceShouldStartArgs, strings.Trim(lit.Value, `"`),
    "shouldStart(%s) is not a service: add it to nonServiceShouldStartArgs", lit.Value)
```

Item 3 can't be closed by AST matching alone — a comment noting that `shouldStart` must be called directly is enough.

### ChiR6 — A contradicting stale paragraph survives further down p2p.md

**Problem:** The PR corrects line 41 but leaves lines 209 and 216-217 of the same file describing DHT setup as `initDHT` / `initPrivateDHT` methods on a `P2PClient` struct, with "Standard mode" using public IPFS bootstrap nodes and "Private mode" creating "an isolated private network with custom bootstrap nodes". Neither function exists anywhere — in this repo or in `go-p2p-message-bus@v0.1.17`, where the entry point is `setupDHT` (`client.go:391`) and the only knob is `DHTMode` ∈ `server`/`client`/`off` (`config.go:102-120`). There is no private-DHT mode, and bootstrap-peer selection is driven by the setting, not by a mode switch — which is what the newly corrected line 41 says.

**Why it matters:** A reader who gets past the fixed paragraph lands on one that contradicts it and names symbols they can't grep for, which undercuts the correction. Same class of drift, same file.

**Fix:** Rewrite those bullets against the current code: DHT setup happens in `go-p2p-message-bus`, `p2p_dht_mode` selects `server` / `client` / `off`, and bootstrap peers come from `p2p_bootstrap_peers` (see ChiR3). Fine as a follow-up if you'd rather keep this PR tight.

## Recap

| ChiR1 | Glossary keeps phantom bootstrap service | YES | 70% |
| ChiR2 | Guard comment overstates doc coverage | | 50% |
| ChiR3 | Private-network bootstrap trap unmentioned | | 40% |
| ChiR4 | Pruner `Block` trigger isn't a fallback | | 35% |
| ChiR5 | Guard AST scope and matching gaps | | 25% |
| ChiR6 | Stale `initDHT` paragraph in p2p.md | | 15% |

icellan:
Addressed the blocking finding (ChiR1, ordishs P2): a nonexistent 'P2P Bootstrap Service' (removed ~2024, bootstrapping is now just the p2p_bootstrap_peers config option) still survived as a formally-defined glossary entry (two places) and in the sibling teranode-overall-system-design.md overlay-services list, which was also missing Pruner for the same reason the microservices-overview doc originally was. Fixed both.

Deferred (per reviewer's own scoping): deleting the orphaned plantuml/bootstrap/ diagram directory (~9 files) — explicitly flagged as separate follow-up cleanup, not blocking.

icellan:
Addressed:

- **ChiR1 (blocking)** — dropped the phantom "P2P Bootstrap Service" from `docs/references/glossary.md`. Line 69 is replaced with a `Bootstrap peers` entry describing the actual `p2p_bootstrap_peers` config option; line 55's libp2p entry now credits the P2P Service instead. Left the orphaned `docs/topics/services/img/plantuml/bootstrap/` diagrams alone, per your note — that's a separate cleanup.
- **ordishs P2** — `teranode-overall-system-design.md:126` was already fixed in the head commit you reviewed (lists Pruner Service, no Bootstrap). Nothing left to do there.
- **ChiR2** — reworded the guard's doc comment and the PR description to say precisely what it checks: daemon-dispatch (`daemon_services.go`, now the whole `daemon` package) against `services/<name>` in both directions, with an explicit note that it does not check the architecture docs.
- **ChiR3** — verified the claim against the vendored `go-p2p-message-bus@v0.1.17` (matches go.mod): `getBootstrapAndRelayPeers` does fall back to `dht.GetDefaultBootstrapPeerAddrInfos()` when `BootstrapPeers` is empty, and `connectToManagedPeers` dials them regardless of `DHTMode`. Added the warning clause to `docs/topics/services/p2p.md:41`.
- **ChiR4** — reworded the Pruner bullet in `teranode-microservices-overview.md` to say the trigger is selected by `pruner_block_trigger`, not a fallback.
- **ChiR5** — hardened the guard: it now walks every non-test file in the `daemon` package (not just `daemon_services.go`, so `daemon.go`'s `wait_for_postgres` dispatch is in scope), gates matches on the receiver being the daemon (`d.shouldStart(...)`), and checks string-literal arguments against the exemption map instead of hard-failing on them. `go test ./daemon/... -race` and `go vet ./daemon/...` both pass.

Deferred (your call on these, noting why):

- The orphaned `plantuml/bootstrap/` diagram directory (ChiR1's follow-up) — left as-is, you were explicit this is separate, larger cleanup.
- **ChiR3's root cause** — the `BootstrapPeers` setting's own longdesc still recommends "leave empty for private networks," which is the actual trap; that's a separate settings-doc fix.
- **ChiR4's other two claim sites** (`services/pruner/server.go:6-7` package comment, `docs/topics/services/pruner.md:88/132`) — same "fallback" framing, but you scoped the fix to the doc this PR touches.
- **ChiR6** (stale `initDHT`/`initPrivateDHT`/"Private mode" paragraph at `p2p.md:209-217`) — same drift class, but it's a real rewrite of section 2.2 against the current `go-p2p-message-bus` DHT setup, not a one-line fix. Left for a follow-up as you suggested.
- Found one more instance of the same drift while grepping, out of scope for this PR: `docs/topics/architecture/teranode-architecture-brief.md:71` still lists "Bootstrap" as a Teranode service and is also missing Pruner. Flagging for the same follow-up rather than expanding this diff further.

Not touched: ChiR5's item 3 (method-value indirection) can't be closed by AST matching per your own note; left the code comment stating shouldStart must be called directly.

ctnguyen (approved):
Review at commit [6597f1a](https://github.com/icellan/teranode/commit/6597f1adf507f9fd208a225aa9953fc3136077b4)

Solid second pass — the blocking item is genuinely closed and the guard is in much better shape. `glossary.md:55` and `:69` now describe what exists rather than a service deleted years ago, and the `Bootstrap peers` replacement entry is accurate against the library: bootstrap peers really are used as both DHT entry points and relay peers (`go-p2p-message-bus@v0.1.17/client.go:225-227`) and really are dialled at startup (`client.go:123`). The new p2p.md clause about blanking `p2p_bootstrap_peers` matches `getBootstrapAndRelayPeers` exactly (`client.go:215-223`), and `p2p_dht_mode = off` is a real value — the library handles it at `client.go:112` and `settings.conf:896` already ships it. The guard rewrite is the nicest part: walking the whole package picks up `daemon.go:284`, the receiver gate and the literal-exemption path are both in, and the comment now says plainly that docs are not covered. Two small things left: one gap in the glossary, and one more direction the guard could cheaply cover.

## Status of previous points

ChiR6 is unchanged and stays open under its own id — `p2p.md:209` and `:216-217` still describe `initDHT` / `initPrivateDHT` and a "Private mode", neither of which exists in this repo or in `go-p2p-message-bus@v0.1.17`. That was agreed as a follow-up rewrite of section 2.2, so it is recorded as NOT FIXED rather than re-raised as a new item.

Two residuals worth noting, also not renumbered. ChiR1: the prose is clean — no `.md` under `docs/` mentions a bootstrap service any more — but the orphaned assets remain (`docs/topics/services/img/plantuml/bootstrap/`, plus `img/Bootstrap_Service_Container_Context_Diagram.png` and `img/Bootstrap_Service_System_Context_Diagram.png`), which was explicitly scoped out. ChiR5: the receiver gate is name-based (`daemonReceiverIdent = "d"`), which was the prescribed minimum, but note the `Stores` type in the same package also uses `d` as its receiver (`daemon_stores.go:64-671`) — so the gate buys little discrimination here, and a dispatch written with any other receiver name goes unseen. Fine as-is; just don't treat it as type-accurate.

| ChiR1 | Phantom service in glossary | YES | FIXED |
| ChiR2 | Guard overstates doc coverage | | FIXED |
| ChiR3 | Private-network bootstrap trap | | FIXED |
| ChiR4 | `Block` trigger not a fallback | | FIXED |
| ChiR5 | Guard AST scope and matching | | FIXED |
| ChiR6 | Stale `initDHT` paragraph | | NOT FIXED |

## Non-blocking issues

### ChiR7 — The glossary is now the only service-inventory doc still missing Pruner

**Problem:** `docs/references/glossary.md` defines an entry for every other service the daemon starts — Alert (line 3), Asset (7), Block Assembly (11), Block Persister (15), Block Validation (17), Blockchain (23), Legacy (53), P2P (71), Propagation (77), RPC (83), Subtree Validation (89), UTXO Persister (99), Validator (103) — 13 of the 14 names in `daemonServiceNames`. There is no Pruner entry. This is the same omission the PR fixes in `teranode-microservices-overview.md:424` and `teranode-overall-system-design.md:128`, in a file this commit already edits for exactly that reason.

**Why it matters:** The glossary is the flat index people grep when they see an unfamiliar service name in a log line or a deployment manifest. Pruner is the newest overlay service and therefore the one most likely to be looked up, and it is the only one that returns nothing. Leaving it out reproduces, in the third doc, the drift the first two just had removed.

**Fix:** Add an entry alongside the other services, in the `P` run near line 75, keeping the trigger-selection wording consistent with the fix already made at `teranode-microservices-overview.md:432` — the two modes are mutually exclusive, each handler discarding the other notification type (`services/pruner/server.go:158-161` and `:202-205`):

```markdown
**Pruner Service**: Removes stale UTXO data — delete-at-height records and the parents of old unmined transactions — to keep the UTXO store from growing without bound. Runs on `BlockPersisted` or `Block` notifications, selected by `pruner_block_trigger`.
```

### ChiR8 — The guard never checks `services/` back against the inventory

**Problem:** The test walks dispatch→inventory (`daemon_service_inventory_test.go:69-83`) and inventory→`services/<name>` (lines 88-95), but nothing walks `services/*` back the other way. A directory under `services/` that no `shouldStart` call dispatches passes silently. That is the shape the phantom P2P Bootstrap Service would have had in reverse — a package left behind after its dispatch was removed — and it is the one drift direction a filesystem check can actually catch, unlike the doc drift the comment correctly disclaims. Today the sets happen to match exactly (14 directories, 14 inventory entries), so the check would be green on arrival.

**Why it matters:** The header at lines 58-64 sells the test as bidirectional, and it is — but only across the dispatch/inventory axis, not the axis that involves the filesystem. Closing this makes the guard fail on a half-finished removal instead of waiting for a human to notice an unreferenced package.

**Fix:** Add a third direction after the existing loop, reusing `servicesDir`:

```go
// Direction 3: every services/<name> package must be reachable from the dispatch.
inventoried := make(map[string]struct{}, len(daemonServiceNames))
for _, name := range daemonServiceNames {
    inventoried[strings.ToLower(name)] = struct{}{}
}

serviceDirs, err := os.ReadDir(servicesDir)
require.NoError(t, err, "failed to read %s", servicesDir)

for _, entry := range serviceDirs {
    if !entry.IsDir() {
        continue
    }

    _, ok := inventoried[entry.Name()]
    require.True(t, ok,
        "services/%s exists but no d.shouldStart call dispatches it: wire it into startServices or remove the package",
        entry.Name())
}
```

## Recap

| ChiR7 | Glossary missing Pruner entry | | 35% |
| ChiR8 | Guard skips `services/` reverse check | | 30% |

PR #1560, opened 2026-08-10
fix(legacy): disconnect peer on outbound write error, clamp advertised block size to wire capacity

icellan:
Two independent fixes to the legacy (SVNode bridge) service's wire-protocol limit handling:

1. **outHandler write-error recovery.** A write error to a peer's socket was logged but never triggered a disconnect, so a socket left alive by the OS after a failed write could freeze that peer's outbound send queue permanently (sendDoneQueue only drains on successful sends). Restored a disconnect call on the write-error path.

2. **Advertised excessive-block-size vs actual wire capacity.** The service unconditionally floored the configured `ExcessiveBlockSize` to `math.MaxUint64` regardless of user config, so the advertised EB user-agent comment implied an essentially unlimited block size — while the wire path actually caps real message relay at ~4GB (`wire.SetLimits`). The configured value is now respected but clamped to what the wire path can actually deliver, so advertised capacity matches real capacity.

**Out of scope, flagged for follow-up:** a related silent-truncation bug in the vendored `github.com/bsv-blockchain/go-wire` dependency — `WriteMessageWithEncodingN` returns `(0, nil)` instead of an error for payloads over `math.MaxUint32` bytes. That's third-party dependency code, not fixed here; needs an upstream fix or a local `replace` directive as a separate change.

### DIFF
--- a/services/legacy/config.go
+++ b/services/legacy/config.go
@@ -8,7 +8,6 @@ package legacy
 
 import (
 	"fmt"
-	"math"
 	"net"
 	"net/url"
 	"os"
@@ -28,26 +27,32 @@ import (
 )
 
 const (
-	defaultDataDir            = "data"
-	defaultMaxPeers           = 125
-	defaultMaxPeersPerIP      = 5
-	defaultBanDuration        = time.Hour * 24
-	defaultBanThreshold       = 100
-	defaultConnectTimeout     = time.Second * 30
-	defaultFreeTxRelayLimit   = 15.0
-	defaultTrickleInterval    = peer.DefaultTrickleInterval
-	defaultExcessiveBlockSize = math.MaxUint64
-	// maxWireMessagePayload is the largest message payload the wire protocol
-	// path can actually relay (set via wire.SetLimits in Server.go's Init).
-	// ExcessiveBlockSize is clamped to this value so the advertised EB never
-	// promises more than the transport can deliver.
-	maxWireMessagePayload = 4_000_000_000
-	defaultBlockMinSize   = 0
-	// defaultBlockMaxSize is derived from maxWireMessagePayload, not
-	// defaultExcessiveBlockSize: ExcessiveBlockSize is clamped to
-	// maxWireMessagePayload in loadConfig, so the default block max size must
-	// stay within that same ceiling or the sanity check below rejects it.
-	defaultBlockMaxSize            = maxWireMessagePayload - 1000
+	defaultDataDir          = "data"
+	defaultMaxPeers         = 125
+	defaultMaxPeersPerIP    = 5
+	defaultBanDuration      = time.Hour * 24
+	defaultBanThreshold     = 100
+	defaultConnectTimeout   = time.Second * 30
+	defaultFreeTxRelayLimit = 15.0
+	defaultTrickleInterval  = peer.DefaultTrickleInterval
+	// maxWireBlockPayload is the excessive block size handed to wire.SetLimits
+	// in Server.go's Init. go-wire keeps it as its `ebs` and returns it from
+	// MaxBlockPayload(), which is the per-message cap it enforces on `block`
+	// and `tx` messages on both read and write - so it is the largest block
+	// this node will accept over the legacy wire. It is not the overall
+	// message payload cap: go-wire derives that as ((ebs/1e6)*1MiB)*2, i.e.
+	// roughly 8.39 GB for this value (go-wire v1.2.10 message.go:38-42).
+	maxWireBlockPayload = 4_000_000_000
+	// defaultExcessiveBlockSize is only a placeholder. loadConfig always
+	// replaces it with the limit block validation actually enforces, see
+	// advertisedExcessiveBlockSize.
+	defaultExcessiveBlockSize = maxWireBlockPayload
+	defaultBlockMinSize       = 0
+	// defaultBlockMaxSize is derived from maxWireBlockPayload: there is no
+	// point mining a block larger than the legacy wire path can carry. When
+	// the enforced excessive block size is smaller, loadConfig caps this down
+	// further.
+	defaultBlockMaxSize            = maxWireBlockPayload - 1000
 	blockMaxSizeMin                = 1000
 	defaultGenerate                = false
 	defaultSigCacheMaxSize         = 100000
@@ -128,12 +133,21 @@ func minUint64(a, b uint64) uint64 {
 	return b
 }
 
-// clampExcessiveBlockSize caps a configured (or defaulted) ExcessiveBlockSize
-// at maxWireMessagePayload, the largest payload the wire protocol path can
-// actually relay (see wire.SetLimits in Server.go's Init). Values at or below
-// the ceiling are returned unchanged.
-func clampExcessiveBlockSize(v uint64) uint64 {
-	return minUint64(v, maxWireMessagePayload)
+// advertisedExcessiveBlockSize returns the excessive block size to advertise in
+// the EB user agent comment. It derives from policyExcessiveBlockSize - the
+// limit block validation enforces in ValidateBlock, i.e.
+// settings.Policy.ExcessiveBlockSize - capped at maxWireBlockPayload, the
+// largest block this node can receive over the legacy wire. The advertised
+// value is therefore never larger than the block size this node honours.
+//
+// A non-positive policy value means the policy limit is disabled (see
+// ValidateBlock), leaving the wire cap as the only effective limit.
+func advertisedExcessiveBlockSize(policyExcessiveBlockSize int) uint64 {
+	if policyExcessiveBlockSize <= 0 {
+		return maxWireBlockPayload
+	}
+
+	return minUint64(uint64(policyExcessiveBlockSize), maxWireBlockPayload)
 }
 
 // maxUint32 is a helper function to return the maximum of two uint32s.
@@ -178,7 +192,7 @@ type config struct {
 	CPUProfile              string        `long:"cpuprofile" description:"Write CPU profile to the specified file"`
 	DebugLevel              string        `short:"d" long:"debuglevel" description:"Logging level for all subsystems {trace, debug, info, warn, error, critical} -- You may also specify <subsystem>=<level>,<subsystem2>=<level>,... to set the log level for individual subsystems -- Use show to list available subsystems"`
 	Upnp                    bool          `long:"upnp" description:"Use UPnP to map our listening port outside of NAT"`
-	ExcessiveBlockSize      uint64        `long:"excessiveblocksize" description:"The maximum size block (in bytes) this node will accept. Cannot be less than 32000000."`
+	ExcessiveBlockSize      uint64        `long:"excessiveblocksize" description:"The maximum size block (in bytes) this node will accept. Derived from the excessiveblocksize policy setting, capped at what the legacy wire path can carry."`
 	MinRelayTxFee           float64       `long:"minrelaytxfee" description:"The minimum transaction fee in BSV/kB to be considered a non-zero fee."`
 	FreeTxRelayLimit        float64       `long:"limitfreerelay" description:"Limit relay of transactions with no transaction fee to the given amount in thousands of bytes per minute"`
 	NoRelayPriority         bool          `long:"norelaypriority" description:"Do not require free or low-fee transactions to have high priority for relaying"`
@@ -357,7 +371,11 @@ func parseCheckpoints(checkpointStrings []string) ([]chaincfg.Checkpoint, error)
 // The above results in bsvd functioning properly without any config settings
 // while still allowing the user to override settings with config files and
 // command line options.  Command line options always take precedence.
-func loadConfig(logger ulogger.Logger) (*config, []string, error) {
+//
+// policyExcessiveBlockSize is settings.Policy.ExcessiveBlockSize, the block
+// size limit block validation enforces. The advertised EB user agent comment is
+// derived from it so this node cannot advertise a block size it would reject.
+func loadConfig(logger ulogger.Logger, policyExcessiveBlockSize int) (*config, []string, error) {
 	// Default config.
 	cfg := config{
 		MaxPeers:                defaultMaxPeers,
@@ -618,21 +636,32 @@ func loadConfig(logger ulogger.Logger) (*config, []string, error) {
 		return nil, nil, err
 	}
 
-	// Excessive blocksize can never advertise more than the wire protocol can
-	// actually relay, regardless of what was configured (or defaulted to).
-	cfg.ExcessiveBlockSize = clampExcessiveBlockSize(cfg.ExcessiveBlockSize)
+	// The excessive blocksize we advertise is the one block validation
+	// enforces, capped at what the legacy wire path can carry, so it is never
+	// larger than the block size this node honours.
+	cfg.ExcessiveBlockSize = advertisedExcessiveBlockSize(policyExcessiveBlockSize)
 
 	// Limit the max block size to a sane value.
-	blockMaxSizeMax := cfg.ExcessiveBlockSize - 1000
-	if cfg.BlockMaxSize < blockMaxSizeMin || cfg.BlockMaxSize > blockMaxSizeMax {
-		str := "%s: The blockmaxsize option must be in between %d and %d -- parsed [%d]"
-		err = fmt.Errorf(str, funcName, blockMaxSizeMin,
-			blockMaxSizeMax, cfg.BlockMaxSize)
+	if cfg.BlockMaxSize < blockMaxSizeMin {
+		str := "%s: The blockmaxsize option may not be less than %d -- parsed [%d]"
+		err = fmt.Errorf(str, funcName, blockMaxSizeMin, cfg.BlockMaxSize)
 		logger.Errorf("%v", err)
 		logger.Errorf("%s", usageMessage)
 
 		return nil, nil, err
 	}
+
+	// Never mine a block bigger than the excessive blocksize we accept. This is
+	// capped rather than rejected because the excessive blocksize follows the
+	// enforced policy setting, which an operator can legitimately set below the
+	// default max block size.
+	blockMaxSizeMax := uint64(blockMaxSizeMin)
+	if cfg.ExcessiveBlockSize > blockMaxSizeMin+1000 {
+		blockMaxSizeMax = cfg.ExcessiveBlockSize - 1000
+	}
+
+	cfg.BlockMaxSize = minUint64(cfg.BlockMaxSize, blockMaxSizeMax)
+
 	// Limit the minimum block sizes to max block size.
 	cfg.BlockMinSize = minUint64(cfg.BlockMinSize, cfg.BlockMaxSize)
 
--- a/services/legacy/config.go
+++ b/services/legacy/config.go
@@ -28,17 +28,26 @@ import (
 )
 
 const (
-	defaultDataDir                 = "data"
-	defaultMaxPeers                = 125
-	defaultMaxPeersPerIP           = 5
-	defaultBanDuration             = time.Hour * 24
-	defaultBanThreshold            = 100
-	defaultConnectTimeout          = time.Second * 30
-	defaultFreeTxRelayLimit        = 15.0
-	defaultTrickleInterval         = peer.DefaultTrickleInterval
-	defaultExcessiveBlockSize      = math.MaxUint64
-	defaultBlockMinSize            = 0
-	defaultBlockMaxSize            = defaultExcessiveBlockSize - 1000
+	defaultDataDir            = "data"
+	defaultMaxPeers           = 125
+	defaultMaxPeersPerIP      = 5
+	defaultBanDuration        = time.Hour * 24
+	defaultBanThreshold       = 100
+	defaultConnectTimeout     = time.Second * 30
+	defaultFreeTxRelayLimit   = 15.0
+	defaultTrickleInterval    = peer.DefaultTrickleInterval
+	defaultExcessiveBlockSize = math.MaxUint64
+	// maxWireMessagePayload is the largest message payload the wire protocol
+	// path can actually relay (set via wire.SetLimits in Server.go's Init).
+	// ExcessiveBlockSize is clamped to this value so the advertised EB never
+	// promises more than the transport can deliver.
+	maxWireMessagePayload = 4_000_000_000
+	defaultBlockMinSize   = 0
+	// defaultBlockMaxSize is derived from maxWireMessagePayload, not
+	// defaultExcessiveBlockSize: ExcessiveBlockSize is clamped to
+	// maxWireMessagePayload in loadConfig, so the default block max size must
+	// stay within that same ceiling or the sanity check below rejects it.
+	defaultBlockMaxSize            = maxWireMessagePayload - 1000
 	blockMaxSizeMin                = 1000
 	defaultGenerate                = false
 	defaultSigCacheMaxSize         = 100000
@@ -119,19 +128,17 @@ func minUint64(a, b uint64) uint64 {
 	return b
 }
 
+// clampExcessiveBlockSize caps a configured (or defaulted) ExcessiveBlockSize
+// at maxWireMessagePayload, the largest payload the wire protocol path can
+// actually relay (see wire.SetLimits in Server.go's Init). Values at or below
+// the ceiling are returned unchanged.
+func clampExcessiveBlockSize(v uint64) uint64 {
+	return minUint64(v, maxWireMessagePayload)
+}
+
 // maxUint32 is a helper function to return the maximum of two uint32s.
 // This avoids a math import and the need to cast to floats.
 func maxUint32(a, b uint32) uint32 {
-	if a > b {
-		return a
-	}
-
-	return b
-}
-
-// maxUint64 is a helper function to return the maximum of two uint64s.
-// This avoids a math import and the need to cast to floats.
-func maxUint64(a, b uint64) uint64 {
 	if a > b {
 		return a
 	}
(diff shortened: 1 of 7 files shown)


ordishs (approved):
Verified both problems in the checkout. `outHandler` at `services/legacy/peer/peer.go:2358-2369` had no disconnect on a write error — `git log -L` shows commit cfe90ee87 (#2246) commented out `p.Disconnect()`, and 110d705dc later deleted the comment. `config.go:615` floored `ExcessiveBlockSize` up to `math.MaxUint64`, which made the user agent read `EB18446744073709.6`.

The fix is complete: the other six `writeMessage` callers all sit in the handshake path, where the error returns and the caller disconnects.

Four nits:

- `services/legacy/peer/peer.go:2360-2369` — the disconnect runs after the error log, unlike btcd, which disconnects first so `shouldLogWriteError` suppresses the duplicate. `DisconnectWithLogFunc` logs before the dedup counter, so each repeated write error adds another `Warnf` line.
- `services/legacy/config.go:632` — the advertised EB still does not come from the value the node enforces. Block validation rejects on `settings.Policy.ExcessiveBlockSize`.
- `services/legacy/config.go:38-41` — the name and comment of `maxWireMessagePayload` state a fact the dependency contradicts. go-wire v1.2.10 `message.go:40-42` computes `maxMessagePayload()` as `((ebs / 1000000) * 1024 * 1024) * 2`.
- `services/legacy/config_test.go` — the new tests use `t.Fatalf` rather than testify `require`, against the AGENTS.md convention, although they match the existing style of that file.

ctnguyen (approved):
Review at commit [9b33343](https://github.com/icellan/teranode/commit/9b33343a36572e980ed84302af0772615ab01fb7)

Both halves of this are correct, and the diagnosis behind the first one is exactly right — I traced the freeze end to end: with no disconnect, `outHandler` never signals `sendDoneQueue`, so `queueHandler` stays `waiting` forever, and because `p.quit` never closes nobody ever drains `pendingMsgs`; every caller blocked on a `doneChan` hangs while the peer still reports `Connected()` and keeps getting handed work. Disconnecting first fixes that properly rather than papering over it. The log-level pick before the disconnect is correct on the first error (`p.disconnect` is still 0 at that point), and the rewritten go-wire comment checks out against v1.2.10 — `MaxBlockPayload()` returns `ebs` verbatim and is enforced before the streaming external-handler delegation too, so 4e9 really is the inbound block ceiling. The new peer tests do reach the fixed path and are free of the races I looked for, though only through one of its two logging branches (ChiR2). Nothing here blocks; my notes are about logging blast radius and some leftovers.

## Non-blocking issues

### ChiR1 — Disconnect reasons are now dropped for every non-first call, service-wide

*Same as [@ordishs's point 1 (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1560#pullrequestreview-4907185773).*

**Problem:** `DisconnectWithLogFunc` (`services/legacy/peer/peer.go:2547`) now logs only when `atomic.AddInt32(&p.disconnect, 1) == 1`. This is the flip side of ordishs' point: the duplicate-`Warnf` case he described only existed *because* the disconnect was missing, and restoring the disconnect alone already closes it — `writeMessage` returns `nil` immediately once `p.disconnect != 0` (`peer.go:1319-1321`), and `outHandler` is a single goroutine, so after the first write-error disconnect the error branch is unreachable and at most one write-error line per peer is possible. What the extra dedup does buy is a loss: roughly fifty call sites across `peer_server.go`, `netsync/manager.go` and `peer.go` pass a reason, and the second and later reasons now vanish silently. The common ordering is not favourable — `inHandler` unconditionally calls `DisconnectWithInfo("Peer appears to be stalled or misbehaving, inHandler break out")` on exit (`peer.go:2144`), so an ordinary remote close usually wins the counter and the specific reason that follows (`"Misbehaving peer -- banning and disconnecting"` at `peer_server.go:570`, the ban and protocol-violation reasons in netsync) is discarded.

**Why it matters:** those reasons are the only record of *why* a peer was dropped, so ban and protocol-violation incidents become unattributable — and the noise reduction they were traded for is already provided by the guard in `writeMessage`.

**Fix:** keep the log unconditional; if repeat noise is still a concern, demote the non-first calls rather than silencing them:

```go
if atomic.AddInt32(&p.disconnect, 1) != 1 {
	p.logger.Debugf("Disconnect (%s) already in progress, additional reason: %s", p, reason)
	return
}

logFunc("Disconnecting (%s) reason: %s", p, reason)
```

**Criticality: 45%**

### ChiR2 — Write-error disconnects are logged below the shipped log level, and that branch is untested

**Problem:** the level comes from `shouldLogWriteError`, which returns `false` for any non-temporary `*net.OpError` (`peer.go:2330`) — i.e. the common cases: broken pipe, connection reset by peer, use of a closed connection. So `logFunc` is `Debugf`, and because the level is now attached to the *disconnect* line rather than to a separate error line, the whole event is emitted at Debug. `settings.conf:782` ships `logLevel = INFO`, so a peer dropped because its socket died leaves no trace at all, while every other disconnect path in the service logs at Info or Warn. This is not a regression against the pre-PR code (which logged nothing *and* did not disconnect); it is an inconsistency in how the newly added disconnect event is classified. Note also that neither new test covers this branch: `flakyWriter.Write` returns a plain `errors.New("simulated write error")` (`peer_test.go:1258`), which trips none of `shouldLogWriteError`'s suppression cases, so both tests take the `Warnf` path. The tests do exercise the disconnect fix, but the logging behaviour that will actually be seen in production is unexercised.

**Why it matters:** making this failure visible is much of the point of the change — a peer that silently disappears at the shipped log level is exactly the symptom that let the missing disconnect survive two commits (`cfe90ee87`, `110d705dc`) unnoticed.

**Fix:** log the disconnect itself at Info unconditionally and let `shouldLogWriteError` gate only whether the underlying error deserves a `Warnf`:

```go
if p.shouldLogWriteError(err) {
	p.logger.Warnf("Failed to send message to %s: %v", p, err)
}

p.DisconnectWithInfo(fmt.Sprintf("write error: %v", err))
```

Give `flakyWriter` a settable error so one case can return a non-temporary `*net.OpError` (e.g. `&net.OpError{Op: "write", Err: syscall.EPIPE}`) and assert the resulting log level.

**Criticality: 40%**

### ChiR3 — The new BlockMaxSize clamp does not hold its own invariant for small policy values

**Problem:** `config.go:658-663` floors `blockMaxSizeMax` at `blockMaxSizeMin` (1000) whenever `cfg.ExcessiveBlockSize <= 2000`, then clamps `BlockMaxSize` down to it. For a policy `excessiveblocksize` below 1000 the result is `BlockMaxSize = 1000 > ExcessiveBlockSize` — the exact opposite of the comment's "Never mine a block bigger than the excessive blocksize we accept". Between 1000 and 2000 that inequality survives but the 1000-byte slack the code deliberately reserves is silently gone. The new test states the invariant as `require.LessOrEqual(t, cfg.BlockMaxSize, cfg.ExcessiveBlockSize-1000)` but only exercises it at 64 MB, so it cannot catch either case. Two things that are fine: there is no underflow (the `> blockMaxSizeMin+1000` guard covers the subtraction), and the blast radius is currently nil because `cfg.BlockMaxSize`/`BlockMinSize` are written in `loadConfig` and read nowhere else in the tree — block assembly uses `settings.Policy.BlockMaxSize`. Also worth deciding deliberately rather than by accident: the old code *errored* on an out-of-range `blockmaxsize` and the new code silently caps. That is the right call for the ordinary range (with EB derived from policy, the old check would have failed legacy startup for any policy below ~4 GB), but it removes the only signal a misconfiguring operator would get.

**Why it matters:** the clamp is the sole enforcement of the stated relationship and it does not hold at the boundary; the moment `cfg.BlockMaxSize` is wired to something real, the violation becomes live rather than theoretical.

**Fix:** pick an explicit behaviour for a too-small enforced EBS instead of inventing a floor — clamping alone cannot satisfy both `BlockMaxSize >= blockMaxSizeMin` and `BlockMaxSize <= ExcessiveBlockSize - 1000` once EBS drops to `blockMaxSizeMin + 1000`, so one of them has to give consciously. Rejecting is the closest match to the check being replaced:

```go
// An enforced excessive block size this small cannot support a mined block of
// at least blockMaxSizeMin bytes, so treat it as a misconfiguration rather
// than silently inventing a floor above what we would accept.
if cfg.ExcessiveBlockSize <= blockMaxSizeMin+1000 {
	err = fmt.Errorf("%s: excessiveblocksize %d is too small to mine a block of at least %d bytes",
		funcName, cfg.ExcessiveBlockSize, blockMaxSizeMin)
	logger.Errorf("%v", err)

	return nil, nil, err
}

cfg.BlockMaxSize = minUint64(cfg.BlockMaxSize, cfg.ExcessiveBlockSize-1000)
```

The tradeoff is that an absurd policy value now fails legacy startup instead of being quietly normalised; if you would rather keep the service up, the alternative is to treat it as mining-disabled (`BlockMaxSize = 0`) and log it once at Warn. Either way, add boundary cases to the table test for policy EBS of 999, 1000, 2000 and 2001 so whichever rule is chosen is pinned down.

**Criticality: 28%**

### ChiR4 — "the 4GiB settings default" is not the shipped default

**Problem:** `TestAdvertisedExcessiveBlockSizeTracksEnforcedPolicy` labels 4294967296 as "the 4GiB settings default", and both `loadConfig` calls in `TestExcessiveBlockSizeUserAgentComment` pass it. 4 GiB is only the code fallback in `settings/settings.go:98`; the committed `settings.conf:619` sets `excessiveblocksize = 10737418240`, so out of the box `Policy.ExcessiveBlockSize` is 10 GiB and the shipped path is the "above the cap" case. The table does cover 10737418240 — just labelled as the non-default one. Separately, `TestExcessiveBlockSizeUserAgentComment` still only asserts `len(cfg.UserAgentComments) == 1` with every value assertion commented out; now that the advertised value is deterministic (`EB4000.0` for any policy at or above the cap, independent of `os.Args` since the `flags` parsing at `config.go:434-456` is commented out) there is nothing left blocking a real assertion.

**Why it matters:** mislabelled fixtures leave the next reader believing the default deployment advertises something it doesn't, and the one test named after the user-agent comment still doesn't check the user-agent comment.

**Fix:** relabel 4294967296 as the code fallback and add or rename a case for `settings.conf`'s 10 GiB; uncomment the assertions in `TestExcessiveBlockSizeUserAgentComment` and set them to `EB4000.0`.

**Criticality: 22%**

### ChiR5 — The reordering saves the peer, not `outHandler`; the comment overstates it

**Problem:** the comment at `peer.go:2370-2373` says the disconnect-first ordering stops a caller that "stopped receiving on its done channel" from wedging "this handler". It doesn't — `msg.doneChan <- struct{}{}` on the very next line is still an unguarded blocking send on a channel the caller may never read. What the reordering actually buys is that `p.conn` and `p.quit` are closed first, so `WaitForDisconnect` returns and the peer is reaped; `outHandler` itself can still block indefinitely, and since `stallHandler` needs both `inQuit` and `outQuit` (`peer.go:1809-1823`), a `close(p.outQuit)` that never runs leaks two goroutines per peer. `TestOutHandlerDisconnectsBeforeSignallingDoneChan` demonstrates precisely this: it has to receive from `doneChan` at the end to "release outHandler". This is unreachable in production — every real `doneChan` is buffered (`peer_server.go:1446`) — so it is a comment-accuracy point, not a live bug.

**Why it matters:** the wording invites a future reader to assume `outHandler` cannot block there, which is exactly the assumption that would justify dropping the buffering on a caller's `doneChan`.

**Fix:** reword to say the disconnect completes before any potentially blocking send, so the peer is torn down regardless, and state the caller contract explicitly (`doneChan` must be buffered or actively received). While there: have `connectedPeerPairWithFlakyWriter`'s cleanup also disconnect and `WaitForDisconnect` on `outPeer`, not just `inPeer`.

**Criticality: 18%**

### ChiR6 — Dead helper and inert default left in `config.go`

**Problem:** `maxUint32` (`config.go:153-161`) has no callers anywhere in the package. It was already dead before this PR, but the same hunk removes its sibling `maxUint64`, so this is the natural moment to take it — and nothing will flag it, since `unused` and `staticcheck` are disabled globally *and* additionally excluded for `services/legacy` in `.golangci.json`. Related: `defaultExcessiveBlockSize` is assigned at `config.go:387` and then unconditionally overwritten at `config.go:642`, which the comment openly admits ("only a placeholder").

**Why it matters:** this hunk is already the cleanup hunk; leaving behind a dead helper and a constant that exists only to be overwritten costs the next reader time for no benefit.

**Fix:** delete `maxUint32` and `defaultExcessiveBlockSize`, initialising the struct literal with `ExcessiveBlockSize: advertisedExcessiveBlockSize(policyExcessiveBlockSize)` so the assignment at line 642 disappears too.

**Criticality: 10%**

## Recap

| --- | --- | --- | --- |
| ChiR1 | Non-first disconnect reasons dropped | | 45% |
| ChiR2 | Write-error disconnect logged at Debug | | 40% |
| ChiR3 | BlockMaxSize clamp breaks own invariant | | 28% |
| ChiR4 | Test mislabels non-default EB value | | 22% |
| ChiR5 | Comment overstates reordering guarantee | | 18% |
| ChiR6 | Dead helper and inert config default | | 10% |

services/legacy/config.go
```diff
@@ -611,20 +636,32 @@ func loadConfig(logger ulogger.Logger) (*config, []string, error) {
...
+	// Never mine a block bigger than the excessive blocksize we accept. This is
+	// capped rather than rejected because the excessive blocksize follows the
+	// enforced policy setting, which an operator can legitimately set below the
+	// default max block size.
+	blockMaxSizeMax := uint64(blockMaxSizeMin)
+	if cfg.ExcessiveBlockSize > blockMaxSizeMin+1000 {
+		blockMaxSizeMax = cfg.ExcessiveBlockSize - 1000
+	}
+
+	cfg.BlockMaxSize = minUint64(cfg.BlockMaxSize, blockMaxSizeMax)
```
freemans13, line 658: This fallback inverts the invariant the comment above it states. For any policy value in `(0, 2000]` the `if` is skipped, `blockMaxSizeMax` stays at its `1000` floor, and `BlockMaxSize` comes out **larger** than `ExcessiveBlockSize` — the opposite of "never mine a block bigger than the excessive blocksize we accept", in exactly the degenerate case this branch was written to handle.

I ran the real derivation to confirm rather than eyeball it: `policy = 500` gives `ExcessiveBlockSize = 500`, `BlockMaxSize = 1000`, `BlockMinSize = 0`.

A one-line fix:

```go
blockMaxSizeMax := uint64(blockMaxSizeMin)
if cfg.ExcessiveBlockSize > blockMaxSizeMin+1000 {
	blockMaxSizeMax = cfg.ExcessiveBlockSize - 1000
}

blockMaxSizeMax = minUint64(cfg.ExcessiveBlockSize, blockMaxSizeMax)
```

or drop the `1000` floor entirely and let the clamp go to zero, if a sub-2000-byte policy should simply mean "mine nothing".

This is also the only new branch with no test coverage. `TestLoadConfigAdvertisesEnforcedExcessiveBlockSize` already asserts the right relationship (`BlockMaxSize <= ExcessiveBlockSize-1000`) — it just never tries a policy small enough to trip it. A `loadConfig(ulogger.TestLogger{}, 500)` case would catch this and would have failed before the fix.

Admittedly a 500-byte excessive block size is not a configuration anyone sane ships. But it's the one branch here that's actually wrong, and it's cheap to make right.

freemans13 (changes requested):
Both fixes are real and correctly diagnosed, and I'd be happy to see this land — I'm requesting changes only for one small bug in the new block-size clamp, detailed inline.

The check I cared most about passes: I took a scratch copy of the tree, reverted both `peer.go` hunks, and re-ran the three new tests. All three fail on the pre-fix code with exactly the symptoms claimed — "peer did not disconnect after a write error", and 3 log lines where 1 was wanted. These are genuine regression tests, not decoration. `go build`, `go vet`, `staticcheck` and `golangci-lint` are all clean on the touched files (the 6 `prealloc` hits in the package are pre-existing, in `addrmgr/` and `bsvutil/`), `-race` is clean, and 20 repeat runs of the new peer tests showed zero flakes — the `io.Pipe`/`flakyWriter` harness isn't timing-sensitive.

I also checked the go-wire figures in the new comment against `go-wire@v1.2.10` in the module cache, since they're asserted as fact: `SetLimits(uint64)` is at `message.go:34`, `maxMessagePayload()` at `message.go:38-42` is `((ebs/1e6)*1MiB)*2`, and for `ebs=4e9` that's exactly 8,388,608,000 bytes. The comment and its line reference are both correct.

On the concurrency half, which is where the real deadlock risk lives: `stallControl` has a live consumer for as long as `outHandler` runs, so the pre-write send can't deadlock; `doneChan` is buffered at every in-repo call site (`OnGetData` uses `make(chan struct{}, 1)` at `peer_server.go:1446` and `1455`, everything else passes `nil`); and once the disconnect closes `p.quit`, the existing cleanup loops drain every pending message and signal every waiting `doneChan`. That last part is precisely what unwedges the freeze this PR set out to fix. No objection there.

---

Three non-blocking notes, all about the config half.

**The shipped behaviour change is bigger than the description says.** `settings.conf:619` sets `excessiveblocksize = 10737418240` (10 GiB) for *all* contexts — not the Go default of `4294967296` that the new tests exercise. So a real mainnet node's advertised user agent goes from `EB18446744073709.6` to `EB4000.0`. That's a deliberate and defensible reduction in what we tell legacy peers we accept, but it's network-visible, and it now *understates* what `ValidateBlock` enforces, because the wire cap binds first. Worth stating plainly in the PR body. Separately: if 4 GB really is the inbound ceiling, it's worth asking whether `wire.SetLimits` should track the 10 GiB policy instead so the two limits stop diverging — bigger change, follow-up material, but the divergence deserves a note somewhere.

**The clamp's merit is undersold.** Switching `BlockMaxSize` from reject to clamp isn't leniency — without it this PR would introduce a *startup failure* for every operator running `excessiveblocksize < 4_000_000_000`. The old check never fired because `defaultBlockMaxSize` and `blockMaxSizeMax` were both `MaxUint64-1000` and therefore always equal; once EB follows policy, `defaultBlockMaxSize` (3,999,999,000) exceeds `EB-1000` for any smaller policy and `newServer` returns an error. As written the change reads as a gratuitous relaxation of a validation, which sells it short.

**Two small gaps in the "advertised ≤ enforced" story.** The `%.1f` on `EB/1e6` means any policy under 1,000,000 bytes advertises literally `EB0.0` — unreachable before, when EB was always `MaxUint64`, reachable now that it follows an operator-settable value. And `setConfigValuesFromSettings` (`peer_server.go:3530`) runs *after* `loadConfig` and can set `legacy_config_ExcessiveBlockSize` by reflection, bypassing both new clamps; it lands after the user-agent string is built, so the field and the advertised value can then disagree. The network-facing invariant survives — the advertisement still tracks policy — it's just that the doc comment's absolute phrasing ("this node cannot advertise a block size it would reject") is a bit stronger than the code.

One thing I checked and cleared, in case it comes up: `maxWireBlockPayload` is an untyped constant and every use site converts to `uint64`, so there's no 32-bit overflow. The `int`-typed test literals would overflow a 32-bit `int`, but the repo already can't build on 386 (go-wire's own `math.MaxUint32` overflows there, and `gobdk` ships no 386 files), so it's moot.

I've deliberately left out the disconnect-logging concerns — @ctnguyen covered those well and I traced the same path and agree with the demote-rather-than-silence suggestion.

icellan:
Addressed the blocking BlockMaxSize clamp bug (freemans13): the floor at blockMaxSizeMin (1000) for ExcessiveBlockSize <= 2000 could produce a BlockMaxSize larger than ExcessiveBlockSize for policy values below 1000, the exact opposite of the stated invariant. Added an explicit clamp to ExcessiveBlockSize on top of the floor, plus boundary test cases at 999/1000/2000/2001 (TestBlockMaxSizeNeverExceedsExcessiveBlockSize) that fail on the pre-fix code and pass now.

Also addressed two non-blocking items from ordishs/ctnguyen: the write-error disconnect now logs unconditionally at Info (previously gated by shouldLogWriteError, meaning the exact failure mode this PR fixes visibility for produced zero log output at the shipped level), and non-first disconnect reasons are now logged at Debug instead of silently dropped (previously ~50 call sites' ban/protocol-violation reasons could vanish if an unrelated disconnect won the dedup race first).

Deferred: relabeling the EB test fixture and the dead maxUint32/defaultExcessiveBlockSize cleanup — lower priority, happy to take a follow-up pass if wanted.

freemans13 (approved):
Approving — the clamp bug I blocked on is properly fixed in 4fc2dcf5c, and nothing new arrived with it. Thanks for the quick turnaround.

**The fix.** `blockMaxSizeMax = minUint64(cfg.ExcessiveBlockSize, blockMaxSizeMax)` at `services/legacy/config.go:656` caps the floor down to the excessive block size before the final clamp, so the promise in the comment above it — "never mine a block bigger than the excessive blocksize we accept" — now holds across the whole input range, including the `(0, 2000]` window that inverted it before. `TestBlockMaxSizeNeverExceedsExcessiveBlockSize` covers exactly the boundary (999, 1000, 2000, 2001) and asserts the relationship directly with `require.LessOrEqual(t, cfg.BlockMaxSize, cfg.ExcessiveBlockSize)` rather than only pinning hard-coded numbers. That is the right shape of test here — it checks the end state, not the mechanism.

**What I ran** on this head: `go vet ./services/legacy/...` clean; `golangci-lint run ./services/legacy/ ./services/legacy/peer/` 0 issues; both full packages pass, `services/legacy/peer` under `-race`. I also ran the two new exact-log-count tests with `-race -count=15`, because exact log-count assertions are a classic intermittent-CI source. They're safe, and it isn't luck: the only `logger.Infof` sites in `peer.go` are the block-height update (589), the duplicate-verack path (2000), and the two disconnect paths under test (2376, 2532). Neither of the first two is reachable in those tests, so `info == 1` is deterministic.

**Two claims that carry the second fix, traced into go-wire rather than taken on faith:**

The 4 GB cap survives the streaming block handler. `Server.Init` registers a streaming handler so multi-GB blocks aren't buffered whole, and my worry was that this skips go-wire's per-message check and makes the advertised figure meaningless. It doesn't — in go-wire v1.2.10 the `msg.MaxPayloadLength()` check is at `message.go:471` and the external-handler dispatch is below it at 487, so the cap fires first. Your comment in `config.go` is accurate, including "on both read and write" (write side is `message.go:371`).

The out-of-scope truncation bug is less urgent than the description implies. `WriteMessageWithEncodingN` returning `(0, nil)` above `math.MaxUint32` is real (`message.go:385-387`), but it's unreachable for `block` and `tx` at the current limit: the per-message check at `message.go:371` rejects anything above 4e9 first, and 4e9 sits below `MaxUint32` (4.295e9). It's only reachable for message types whose `MaxPayloadLength` is the overall 8.39 GB cap — `reject`, `cfcheckpoint` — and nobody builds a multi-gigabyte reject message. Worth an upstream fix, but not a live hazard, and holding the limit at 4e9 is what keeps it closed. Might be worth softening that paragraph in the description so the follow-up gets prioritised accordingly.

**Five non-blocking observations, all tidy-ups:**

1. The `BlockMaxSize` clamp guards a value nothing reads. `cfg.BlockMaxSize` and `cfg.BlockMinSize` have no consumers anywhere in the legacy package; the real mining limit is `settings.Policy.BlockMaxSize`, read at `services/blockassembly/BlockAssembler.go:1483`. Only `cfg.ExcessiveBlockSize` is actually read, for the `EB%.1f` comment. Doesn't make the fix wrong — correctness in dead config still beats a stated-but-false promise — but worth a note so a future reader doesn't take this clamp for something load-bearing.

2. `defaultExcessiveBlockSize` is now write-only: assigned at `config.go:377`, unconditionally overwritten at 632, never read in between. Documented as "only a placeholder", which is honest, but it could just be deleted along with the struct initialisation — dead constants get mistaken for real defaults later.

3. `blockMaxSizeMin` is no longer an actual floor. The `< blockMaxSizeMin` validation runs before the clamp, so the clamp can legitimately produce `BlockMaxSize = 999`, below the 1000 the error message above enforces — your own test asserts this (policy 999 gives 999). Harmless given point 1, but the check-then-clamp ordering means the documented minimum isn't one.

4. Silent clamping replaced a hard config error for an oversized `blockmaxsize`. The justification reads fine, and in practice the branch was unreachable anyway since the CLI parsing path in `loadConfig` is commented out — as your own test comments say. Fine as-is.

5. Test hygiene: the second half of `TestExcessiveBlockSizeUserAgentComment` now asserts the same thing as the first half (both `EB4000.0`), so its only remaining value is documenting that the CLI flag is ignored — worth a sentence saying so, or dropping it. And these tests mutate global `os.Args` without restoring it; a `t.Cleanup` would stop that leaking into whatever runs next in the package.

**One operator-visible change worth flagging in the description:** the advertised comment moves from `EB18446744073709.6` to `EB4000.0`. That's honest for the legacy wire, but it understates what the node accepts over the modern P2P path, where the go-wire cap doesn't apply and policy allows the shipped 10 GiB. Safe direction to err in, and the comment only ever goes out on the legacy wire — just worth saying so nobody reads `EB4000.0` as "rejects blocks over 4 GB" in general.

icellan:
CI failure diagnosis: `sequential (shard 2)` / `TestLongestChainForkPostgres/with_double_spend_transaction`.

Root cause is a pre-existing flaky test, not a regression from this PR's diff. `testLongestChainWithDoubleSpendTransaction` waits on `WaitForBlockBeingMined(block6b)` before reading the block-assembly mining candidate for `tx1`, but that call only confirms `setTxMined` has finished on the blockchain/UTXO side. Block assembly's own reorg — which moves `tx1` back into its candidate subtrees once block6b becomes the tip — runs asynchronously and can still be in flight, so `VerifyInBlockAssembly(t, tx1)` races it. A `td.WaitForBlock(t, block6b, blockWait)` call for exactly this purpose is already present, but commented out, at that line. Confirmed by reproducing locally:

- On this branch: `go test -c -race` + running the compiled binary directly (CI compiles the sequential suite with `-race`, which slows things down enough to open the window) fails 3/6 runs with the identical assertion (`Expected transaction ... to be found exactly once, but found 0 times`).
- Same repro against `main` at 9de8693c2 (before this PR): fails 1/6 runs with the same signature. So the race predates this PR; it isn't introduced by the outbound-write-error disconnect or the excessive-block-size clamp.

Fix: uncommented the existing `td.WaitForBlock(t, block6b, blockWait)` call, matching the synchronization already used for the equivalent step in the other two subtests in this file. 10/10 runs green afterward with the same `-race` binary.

Verified: `go build ./...`, `go vet ./...`, `go test -race -tags testtxmetacache ./services/legacy/...` all pass, plus the target test 10/10 with `-race`.

PR #1561, opened 2026-08-10
docs(settings): drop stale unused-topic note from validatortxsConfig longdesc

icellan:
The kafka_validatortxsConfig setting's longdesc claimed the topic is "currently unused in codebase" — it isn't: Propagation produces to it and Validator consumes from it. Removed the stale note.


ordishs (approved):
Verified. The removed note was false: `settings/settings.go:145` parses `kafka_validatortxsConfig` into `Kafka.ValidatorTxsConfig`, `daemon/daemon_kafka.go:166` builds the Validator consumer group from it, and `daemon/daemon_kafka.go:92` builds the Propagation producer from the same key. The fix is complete for its scope — a repo-wide search finds "Currently unused in codebase" only in that one struct tag, and the generated `docs/references/settings/kafka_settings.md` does not repeat it.

Two nits:

- `settings/kafka_settings.go:21` — the surviving longdesc still presents the setting as a tuning override only. It also selects the transport between Propagation and Validator.
- `daemon/daemon_kafka.go:92` — pre-existing: the producer path reads the raw gocore key, while the consumer path at line 166 reads the parsed `settings.Kafka.ValidatorTxsConfig`. The two can diverge, because `getURL` in `settings/helpers.go:65` honours `alternativeContext`.

freemans13 (approved):

PR #1562, opened 2026-08-10
fix(netsync): unskip four legacy netsync tests broken by incomplete fixtures

icellan:
## Summary

Four tests in `services/legacy/netsync` were skipped with a generic "nil pointer issue" comment. In every case the nil pointer was in the test's own `SyncManager` fixture, not in the code under test:

- `TestSyncManager_extendTransactions` - fixture never set `utxoStore`; `extendTransactions`'s phase-2 bulk decorate calls `sm.utxoStore.BatchPreviousOutputsDecorate` unconditionally. Added `utxoStore: &nullstore.NullStore{}` and strengthened the assertion to actually check inputs got extended.
- `TestSyncManager_createUtxos` - same missing `utxoStore`. Also fixed a stale assertion: the test called `utxos := sm.createUtxos(...); assert.NotNil(t, utxos)`, but `createUtxos` only returns `error` - there's no value to be non-nil. Now asserts on the actual error. The coinbase tx fixture also needed an explicit zeroed `PreviousTxID` (`IsCoinbase()` dereferences it via `util.GetFees`).
- `TestSyncManager_validateTransactions` - fixture never set `chainParams`; `validateTransactions` -> `candidateFinalityTimesForBlock` dereferences `sm.chainParams.CSVHeight`. Added regtest chain params and a non-zero block timestamp (the zero-value `time.Time` converts to a huge negative Unix time and fails the conversion).
- `TestSyncManager_prepareSubtrees` - fixture set `blockchainClient`/`validationClient`/`subtreeStore` but not `chainParams`, so `quickValidationAllowed` nil-guarded to `false`, forcing the non-quick path that needs `sm.subtreeValidation` (also never set). Rebuilt the test with a 2-tx block (a single-tx block returns from `prepareSubtrees` before touching any of this) and added chain params + a `subtreevalidation.MockSubtreeValidation`.

## TestBlockchainSync disposition: still skipped, with an accurate reason

`TestBlockchainSync` (`manager_test.go`) was skipped as "TODO: needs to be fixed" / "timing out". Investigated and found two distinct issues stacked on top of each other:

1. `startSync()` now short-circuits to `FSMStateRUNNING` and never sends `PushGetBlocksMsg` whenever the sync peer's advertised height ties ours. This test's remote peer config never sets `NewestBlock`, so it always reports height 0 = our genesis-only height, and the very first `getblocks` wait times out immediately.
2. Working around (1) (by advertising a higher peer height) gets past the handshake, but the test then times out later waiting for `QueueBlock`'s error channel during real block processing. This harness passes a `nil` block-assembly client into `New(...)` and never configures `subtreeValidation`'s block-level expectations, either of which plausibly stalls block processing deep in `HandleBlockDirect`.

Root-causing (2) needs a proper block-processing test harness (real/mocked block assembly + subtree validation wired for an actual block), which is a bigger lift than a fixture nil-pointer fix. Left skipped, with the skip reason now describing both findings instead of the stale placeholder.

## Test plan

- [x] `go test ./services/legacy/netsync/... -race -v -run 'TestSyncManager_extendTransactions|TestSyncManager_createUtxos|TestSyncManager_validateTransactions|TestSyncManager_prepareSubtrees|TestBlockchainSync'` - all pass (`TestBlockchainSync` correctly skipped)
- [x] `go test ./services/legacy/...` - full package suite passes, no regressions

### DIFF
--- a/services/legacy/netsync/handle_block_test.go
+++ b/services/legacy/netsync/handle_block_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/bsv-blockchain/teranode/services/legacy/bsvutil"
 	"github.com/bsv-blockchain/teranode/services/legacy/peer"
 	"github.com/bsv-blockchain/teranode/services/legacy/testdata"
+	"github.com/bsv-blockchain/teranode/services/subtreevalidation"
 	"github.com/bsv-blockchain/teranode/services/validator"
 	"github.com/bsv-blockchain/teranode/stores/blob/memory"
 	"github.com/bsv-blockchain/teranode/stores/utxo"
@@ -367,41 +368,64 @@ func Benchmark_createSubtrees(b *testing.B) {
 
 // Test extendTransactions to increase coverage
 func TestSyncManager_extendTransactions(t *testing.T) {
-	t.Skip("Skipping test due to nil pointer issue")
+	initPrometheusMetrics()
+
 	block, err := testdata.ReadBlockFromFile("../testdata/000000000000000009631dd3dd7357675d8a1f8925be5e7851c68255531ac5fb.bin")
 	require.NoError(t, err)
 
 	sm := &SyncManager{
-		settings: test.CreateBaseTestSettings(t),
-		logger:   ulogger.TestLogger{},
+		settings:  test.CreateBaseTestSettings(t),
+		logger:    ulogger.TestLogger{},
+		utxoStore: &nullstore.NullStore{},
 	}
 
 	txMap := txmap.NewSyncedMap[chainhash.Hash, *TxMapWrapper](len(block.Transactions()))
 	txOrder, err := sm.createTxMap(context.Background(), block, txMap)
 	require.NoError(t, err)
 
-	// Test extending transactions
+	// Test extending transactions: with a real block, non-coinbase txs whose
+	// inputs reference same-block parents get extended via phase 1 (txMap
+	// lookup); the rest fall through to phase 2's BatchPreviousOutputsDecorate.
 	err = sm.extendTransactions(context.Background(), testBlockIdent(block), txOrder, txMap, false)
-	assert.NoError(t, err)
+	require.NoError(t, err)
+
+	sawExtendedInput := false
+
+	for _, txHash := range txOrder[1:] {
+		wrapper, found := txMap.Get(txHash)
+		require.True(t, found)
+
+		for _, in := range wrapper.Tx.Inputs {
+			if in.PreviousTxScript != nil {
+				sawExtendedInput = true
+			}
+		}
+	}
+
+	assert.True(t, sawExtendedInput, "at least one input should have been extended from an in-block parent")
 }
 
 // Test createUtxos to increase coverage
 func TestSyncManager_createUtxos(t *testing.T) {
-	t.Skip("Skipping test due to nil pointer issue")
 	sm := &SyncManager{
-		settings: test.CreateBaseTestSettings(t),
-		logger:   ulogger.TestLogger{},
+		settings:  test.CreateBaseTestSettings(t),
+		logger:    ulogger.TestLogger{},
+		utxoStore: &nullstore.NullStore{},
 	}
 
-	// Create a simple coinbase transaction
+	// Create a simple coinbase transaction. IsCoinbase() (called by
+	// util.GetFees via NullStore.Create) requires a zeroed PreviousTxID, so it
+	// must be set explicitly rather than left as the input's nil default.
+	coinbaseInput := &bt.Input{
+		PreviousTxSatoshis: 0,
+		PreviousTxOutIndex: 0xffffffff,
+		SequenceNumber:     0xffffffff,
+	}
+	require.NoError(t, coinbaseInput.PreviousTxIDAdd(&chainhash.Hash{}))
+
 	coinbaseTx := &bt.Tx{
 		Version: 1,
-		Inputs: []*bt.Input{
-			{
-				PreviousTxSatoshis: 0,
-				PreviousTxOutIndex: 0xffffffff,
-			},
-		},
+		Inputs:  []*bt.Input{coinbaseInput},
 		Outputs: []*bt.Output{
 			{
 				Satoshis:      50 * 100000000,
@@ -427,13 +451,12 @@ func TestSyncManager_createUtxos(t *testing.T) {
 	block.SetHeight(100)
 
 	// Test createUtxos
-	utxos := sm.createUtxos(context.Background(), txMap, testBlockIdent(block), 0, false)
-	assert.NotNil(t, utxos)
+	err := sm.createUtxos(context.Background(), txMap, testBlockIdent(block), 0, false)
+	assert.NoError(t, err)
 }
 
 // Test validateTransactions to increase coverage
 func TestSyncManager_validateTransactions(t *testing.T) {
-	t.Skip("Skipping test due to nil pointer issue")
 	initPrometheusMetrics()
 
 	validationClient := &validator.MockValidator{}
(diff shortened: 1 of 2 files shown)


ordishs (approved):
Verified. The four nil-pointer test fixtures are the real cause, the fix is correct, and leaving `TestBlockchainSync` skipped is the honest call.

One nit: `services/legacy/netsync/handle_block_test.go:392` — the `sawExtendedInput` assertion does not isolate phase-1 behaviour the way the comment implies. `NullStore.PreviousOutputsDecorate` unconditionally overwrites `PreviousTxScript` on every input, and `extendTransactions` phase 2 always calls `BatchPreviousOutputsDecorate` on all non-coinbase txs when `outpointOnly` is false, as this test uses it. So `sawExtendedInput` would be true even if `extendFromTxMap` were fully broken.

ctnguyen (approved):
Review at commit [1cbe46b](https://github.com/icellan/teranode/commit/1cbe46b7e76d445f052afb7be73786e524b21ef5)

Good triage, and the diagnoses hold up. I traced each one: `SomeParentsInBlock` really is written in exactly one place (`handle_block.go:1658`, inside `extendFromTxMap`) and phase 2 never touches it, so the second commit genuinely answers the nit rather than papering over it; `Input.PreviousTxID()` does deref a nil `*chainhash.Hash` via `CloneBytes`, so the coinbase fixture really did need an explicit zeroed `PreviousTxID`; regtest is `CSVHeight: 576` with `Checkpoints: nil`; `txCount <= 1` really did short-circuit the old single-tx `prepareSubtrees` block before anything interesting; and `startSync` (`manager.go:777-787`) really does call `Run(...)` and return without `PushGetBlocksMsg` on a height tie — so the new `TestBlockchainSync` skip reason is accurate, and leaving it skipped with a real explanation beats forcing it green. `go vet -mod=mod ./services/legacy/netsync/` is clean (bare `go vet` fails here on stale local vendoring, unrelated to this PR); I did not re-execute the four tests in this environment, so everything below rests on source tracing rather than observed runs. My findings are all about how much these four tests actually pin down now that they run.

## Non-blocking issues

### ChiR1. `extendTransactions`: phase 2 is executed but never asserted, and the fake makes it hard to assert

*Partially overlaps [@ordishs's nit (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1562#pullrequestreview-4907185329).*

**Problem:** The rewritten assertion covers phase 1 only. Phase 2 — `BatchPreviousOutputsDecorate`, i.e. the exact call that was nil-dereferencing and the reason `utxoStore` had to be added — is unasserted: delete the phase-2 block from `extendTransactions` and this test still passes. The deeper cause is that `NullStore.PreviousOutputsDecorate` (`nullstore.go:114-125`) overwrites `PreviousTxScript`/`PreviousTxSatoshis` on *every* input, while both real stores explicitly skip already-decorated inputs (`stores/utxo/sql/sql.go:4068`, `stores/utxo/aerospike/get.go:1304`) — the contract `handle_block.go:1584` relies on. The comment in the test correctly identifies this and routes around it; the fake stays unfaithful.

**Why it matters:** With the fake clobbering phase 1, this test can never catch a regression where phase 2 overwrites the same-block parent values phase 1 resolved. That is a live failure mode with real consequences: `createSubtrees` feeds `calculateTransactionFee` off those `PreviousTxSatoshis`, so clobbering would silently stamp wrong fees into subtrees. That is precisely the gap the nit pointed at, and it survives the fix.

**Fix:** Use a `netsync`-local spy instead of `NullStore` here — `nullstore` has ~15 call sites and some may rely on the unconditional overwrite, so changing it globally is the wrong blast radius for one test. The spy records that `BatchPreviousOutputsDecorate` was called and honours the skip rule the real stores implement:

```go
type decorateSpy struct {
	utxo.Store // embed NullStore for the rest of the interface
	calls atomic.Int32
}

func (d *decorateSpy) BatchPreviousOutputsDecorate(_ context.Context, txs []*bt.Tx) error {
	d.calls.Add(1)
	for _, tx := range txs {
		for _, input := range tx.Inputs {
			if input == nil || input.PreviousTxScript != nil {
				continue // real stores skip already-decorated inputs
			}
			input.PreviousTxScript = bscript.NewFromBytes([]byte("test"))
			input.PreviousTxSatoshis = 100_000_000_000
		}
	}
	return nil
}
```

Then assert `calls == 1`; that for a tx whose parents are all out-of-block the input carries the sentinel (only phase 2 can have set it); and that a tx with `SomeParentsInBlock` kept its real parent satoshis/script.

### ChiR2. `createUtxos` fixture is a shape production cannot produce

**Problem:** The test seeds `txMap` with a coinbase, but `createTxMap` explicitly excludes it — `handle_block.go:1872`: `if !tx.IsCoinbase() { txMap.Set(hashCopy, ...) }`. So `createUtxos` never sees a coinbase in production. The consequence is that the test drives `util.GetFees` down its coinbase early-return branch and the new `coinbaseInput.PreviousTxIDAdd(&chainhash.Hash{})` exists purely to keep that unrealistic fixture from panicking.

**Why it matters:** The one branch `createUtxos` actually hits in production — a non-coinbase tx whose extended inputs must satisfy `inputFees >= outputFees` — goes untested, and the added comment teaches a future reader that `createUtxos` handles coinbases, which it never does.

**Fix:** Use a regular tx with decorated inputs (`PreviousTxSatoshis` above the output total) and drop the coinbase from the map. The `PreviousTxIDAdd` workaround then falls away with it. If a coinbase in `txMap` is worth pinning at all, it belongs in a `createTxMap` test asserting the exclusion, not here.

### ChiR3. `validateTransactions` asserts only what is already covered elsewhere

**Problem:** In this fixture `assert.NoError` can only fail via `candidateFinalityTimesForBlock` or `EnsureMTPLoaded`. `MockValidator` is a hand-written no-op (no testify expectations), and `validateTransactions` discards every `Validate` error through `classifyAndCountPrewarmError` (`handle_block.go:1479-1481`). So the test passes whether or not a single transaction is ever submitted to the validator. Meanwhile the era-selection logic the new `chainParams`/timestamp lines are there to exercise already has dedicated table tests in `median_timestamp_test.go:192-241`.

**Why it matters:** As written, the test is a "does not panic" smoke check on a path that is already covered better one file over — it will not notice if the level loop stops dispatching transactions.

**Fix:** Assert the one thing only this test can see — that each level's transactions reach the validator — using the mock's existing hook:

```go
var validated atomic.Int32
validationClient.ValidateFunc = func(_ context.Context, tx *bt.Tx) (*meta.Data, error) {
	validated.Add(1)
	return util.TxMetaDataFromTx(tx)
}
...
require.NoError(t, err)
require.EqualValues(t, 1, validated.Load())
```

### ChiR4. `prepareSubtrees` comment implies the chain params select the branch; they don't

**Problem:** The comment reads "Regtest has no checkpoints, so `quickValidationAllowed` is false and `prepareSubtrees` is forced down the non-quick path". Every clause is true, but it suggests the added `&chaincfg.RegressionNetParams` is what routes the test: `quickValidationAllowed` returns false for a nil `chainParams` too, so the branch is identical with and without this line. The change that actually made the test work was setting `subtreeValidation`.

**Why it matters:** Purely a comment-accuracy point — the test itself exercises the intended path and asserts the mock. But a comment that misattributes causality costs the next person changing this fixture a debugging cycle.

**Fix:** Reword to say the params are set for realism, and that regtest can never be below a checkpoint so the non-quick path is the only one reachable here.

### ChiR5. `assert` where the project convention is `require`

**Problem:** The new/changed assertions use `assert.True`, `assert.NoError`, `assert.Len`, `assert.Equal`, `assert.Nil`. AGENTS.md ("Testing") states: use `require` from testify, not `assert`.

**Why it matters:** Convention only here — none of these sites dereference after a failed assertion, so the practical risk is limited to noisier failure output. Flagging it because it is a written project rule and the file is being touched anyway.

**Fix:** Switch the touched assertions to `require`. The surrounding file mixes both, so a blanket cleanup is out of scope; just make the new lines follow the rule.

## Recap

| --- | --- | --- | --- |
| ChiR1 | Phase 2 unasserted; fake breaks contract | | 45% |
| ChiR2 | Coinbase fixture unreachable in production | | 30% |
| ChiR3 | Validator dispatch never asserted | | 25% |
| ChiR4 | Comment misattributes branch selection | | 10% |
| ChiR5 | `assert` instead of `require` | | 8% |

PR #1563, opened 2026-08-10
fix(seeder): detect truncated UTXO snapshot files via footer validation

icellan:
## Summary

- The seeder's UTXO-set read loop (`readUTXOWrapperFile` in `cmd/seeder/seeder.go`) only knew how to distinguish "record stream ended" from "read error" via `errors.Is(err, io.EOF)`. Teranode's `errors.Error.Is` falls back to substring matching against non-`*Error` targets, and `io.ErrUnexpectedEOF`'s message ("unexpected EOF") contains "EOF" as a substring — so a file truncated mid-record produced the exact same code path as a clean end-of-file.
- Every `.utxo-set` snapshot already carries a 16-byte trailing footer (`txCount`, `utxoCount`) written by the persister. The seeder read right past it without ever checking it.
- After the read loop ends (for any reason), compare the number of transactions/UTXOs actually processed against `utxopersister.GetFooter(f)`. A mismatch now returns an error instead of silently completing.

## Operational risk this closes

A seeder import that stops partway through a snapshot — truncated in transit, cut short by a disk/network fault, or by an interrupted copy — previously reported `All workers finished successfully` and wrote `lastProcessed.dat` as if the whole UTXO set had been imported. The node then serves that height as though its UTXO set is complete: it will treat outputs that were never imported as nonexistent and reject valid spends of them — a consensus-divergence failure that only surfaces long after the seeding step reported success, when the node is already serving traffic. The fix makes that failure loud and immediate at seed time instead.

## Test plan

- [x] New regression test `TestReadUTXOWrapperFile_TruncatedFile_ReturnsError` in `cmd/seeder/seeder_truncation_test.go`: builds a well-formed snapshot layout, cuts it off 10 bytes into the second record's TxID (no footer at all), and asserts `readUTXOWrapperFile` returns an error.
- [x] `go build ./cmd/seeder/... ./services/utxopersister/...`
- [x] `go test ./cmd/seeder/... ./services/utxopersister/... -race`

### DIFF
--- a/services/utxopersister/UTXOSet.go
+++ b/services/utxopersister/UTXOSet.go
@@ -21,7 +21,6 @@ import (
 	"encoding/binary"
 	"io"
 	"sort"
-	"strings"
 	"sync"
 
 	"github.com/bsv-blockchain/go-bt/v2"
@@ -659,37 +658,36 @@ func (us *UTXOSet) CreateUTXOSet(ctx context.Context, c *consolidator) (err erro
 				// height/coinbase + its UTXOs).
 				utxoWrapper, err := NewUTXOWrapperFromReader(ctx, previousUTXOSetReader)
 				if err != nil {
-					// CreateUTXOSet appends a 16-byte footer (txCount +
-					// utxoCount) after the final UTXOWrapper. This loop does
-					// not consult that count, so it only learns the records
-					// are exhausted when the next read either lands exactly on
-					// EOF (a bare io.EOF) or short-reads the footer, which
-					// io.ReadFull reports as io.ErrUnexpectedEOF ("unexpected
-					// EOF"). cmd/utxovalidator handles the same footer.
-					//
-					// The short read is matched by substring, not
-					// structurally: errors.New flattens a non-*Error cause to
-					// its message (errors/errors.go:334-336), discarding the
-					// io.ErrUnexpectedEOF sentinel - so errors.Is(err,
-					// io.ErrUnexpectedEOF) would itself reduce to this same
-					// strings.Contains. (And do not fold the io.EOF clause into
-					// errors.Is: "EOF" is a substring of "unexpected EOF", so
-					// it would swallow this footer error too.) A structural fix
-					// - FromReader returning a typed sentinel, and validating
-					// records-read == txCount against the footer - is tracked
-					// as a follow-up.
-					//
-					// Consequence: a genuinely truncated tail is
-					// indistinguishable from the footer and is silently
-					// accepted (pre-existing; same as utxovalidator). Matching
-					// only "unexpected EOF" - not the broader "failed to read
-					// txid" utxovalidator also matches - keeps a real non-EOF
-					// read error loud rather than swallowed.
-					if err == io.EOF || strings.Contains(err.Error(), "unexpected EOF") {
-						break OUTER
+					// CreateUTXOSet unconditionally appends a 16-byte footer
+					// (txCount||utxoCount) after the final UTXOWrapper, with
+					// no marker byte in front of it. The only way this loop
+					// can legitimately be done is if the next read lands
+					// exactly on those footer bytes - which FromReader
+					// reports via the ErrRecordBoundary sentinel - and the
+					// footer's counts agree with what was actually read
+					// here. Anything else (a bare io.EOF, which can only
+					// happen if the footer itself is missing entirely; a
+					// short read at any other offset; or an
+					// ErrRecordBoundary whose footer bytes don't match) is a
+					// genuine truncation and must fail loudly rather than
+					// silently produce a new snapshot that omits the
+					// unread tail.
+					boundary, ok := err.(*ErrRecordBoundary)
+					if !ok {
+						return errors.NewStorageError("error reading previous utxo-set (%s.%s) at iteration %d", c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, txCount, err)
 					}
 
-					return errors.NewStorageError("error reading previous utxo-set (%s.%s) at iteration %d", c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, txCount, err)
+					expectedTxCount, expectedUTXOCount, decErr := DecodeFooter(boundary.FooterBytes[:])
+					if decErr != nil {
+						return errors.NewStorageError("error decoding previous utxo-set (%s.%s) footer", c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, decErr)
+					}
+
+					if expectedTxCount != txCount || expectedUTXOCount != utxoCount {
+						return errors.NewProcessingError("previous utxo-set (%s.%s) is truncated: footer expects %d transactions/%d utxos, only %d/%d were read",
+							c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, expectedTxCount, expectedUTXOCount, txCount, utxoCount)
+					}
+
+					break OUTER
 				}
 
 				ts = readStat.AddTime(ts)
--- a/services/utxopersister/UTXOSet.go
+++ b/services/utxopersister/UTXOSet.go
@@ -588,12 +588,14 @@ func (us *UTXOSet) CreateUTXOSet(ctx context.Context, c *consolidator) (err erro
 	}
 
 	var (
-		readStat   = createStat.NewStat("readTX")
-		filterStat = createStat.NewStat("filterUTXOs")
-		writeStat  = createStat.NewStat("writeUTXOs")
-		ts         = gocore.CurrentTime()
-		txCount    uint64
-		utxoCount  uint64
+		readStat    = createStat.NewStat("readTX")
+		filterStat  = createStat.NewStat("filterUTXOs")
+		writeStat   = createStat.NewStat("writeUTXOs")
+		ts          = gocore.CurrentTime()
+		txCount     uint64
+		utxoCount   uint64
+		recordsRead uint64
+		utxosRead   uint64
 	)
 
 	if c.firstPreviousBlockHash.String() != c.settings.ChainCfgParams.GenesisHash.String() {
@@ -672,9 +674,9 @@ func (us *UTXOSet) CreateUTXOSet(ctx context.Context, c *consolidator) (err erro
 					// genuine truncation and must fail loudly rather than
 					// silently produce a new snapshot that omits the
 					// unread tail.
-					boundary, ok := err.(*ErrRecordBoundary)
-					if !ok {
-						return errors.NewStorageError("error reading previous utxo-set (%s.%s) at iteration %d", c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, txCount, err)
+					var boundary *ErrRecordBoundary
+					if !errors.As(err, &boundary) {
+						return errors.NewStorageError("error reading previous utxo-set (%s.%s) at iteration %d", c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, recordsRead, err)
 					}
 
 					expectedTxCount, expectedUTXOCount, decErr := DecodeFooter(boundary.FooterBytes[:])
@@ -682,15 +684,24 @@ func (us *UTXOSet) CreateUTXOSet(ctx context.Context, c *consolidator) (err erro
 						return errors.NewStorageError("error decoding previous utxo-set (%s.%s) footer", c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, decErr)
 					}
 
-					if expectedTxCount != txCount || expectedUTXOCount != utxoCount {
+					if expectedTxCount != recordsRead || expectedUTXOCount != utxosRead {
 						return errors.NewProcessingError("previous utxo-set (%s.%s) is truncated: footer expects %d transactions/%d utxos, only %d/%d were read",
-							c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, expectedTxCount, expectedUTXOCount, txCount, utxoCount)
+							c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, expectedTxCount, expectedUTXOCount, recordsRead, utxosRead)
 					}
 
 					break OUTER
 				}
 
 				ts = readStat.AddTime(ts)
+
+				// Count every wrapper read from the previous file,
+				// unconditionally and before deletion filtering, so the
+				// truncation check above compares against what was actually
+				// read - not against txCount/utxoCount, which only track
+				// survivors after filtering and exist to write the *new*
+				// file's own footer below.
+				recordsRead++
+				utxosRead += uint64(len(utxoWrapper.UTXOs))
 
 				// Filter UTXOs based on the deletions map
 				utxoWrapper.UTXOs = filterUTXOs(utxoWrapper.UTXOs, c.deletions, &utxoWrapper.TxID)
--- a/cmd/seeder/seeder_truncation_test.go
+++ b/cmd/seeder/seeder_truncation_test.go
@@ -88,6 +88,55 @@ func openForReading(t *testing.T, path string) (*os.File, *bufio.Reader) {
 	return f, reader
 }
 
+// TestReadUTXOWrapperFile_CompleteFile_Succeeds is the happy-path companion
+// to the truncation regression test below: a well-formed snapshot file - the
+// two records built by buildSnapshotWrappers, followed by a real 16-byte
+// footer matching the actual counts - must be accepted, and the drained
+// channel must yield exactly txCount wrappers. This guards against a future
+// change to the footer-count comparison, the skip-empty-wrapper gate, or the
+// footer layout silently making the seeder reject every valid snapshot.
+func TestReadUTXOWrapperFile_CompleteFile_Succeeds(t *testing.T) {
+	metadata, wrapperBytes, txCount, utxoCount := buildSnapshotWrappers(t)
+	require.Len(t, wrapperBytes, 2)
+
+	header := fileformat.NewHeader(fileformat.FileTypeUtxoSet)
+
+	var footer [16]byte
+	binary.LittleEndian.PutUint64(footer[0:8], txCount)
+	binary.LittleEndian.PutUint64(footer[8:16], utxoCount)
+
+	data := make([]byte, 0, len(header.Bytes())+len(metadata)+len(wrapperBytes[0])+len(wrapperBytes[1])+len(footer))
+	data = append(data, header.Bytes()...)
+	data = append(data, metadata...)
+	data = append(data, wrapperBytes[0]...)
+	data = append(data, wrapperBytes[1]...)
+	data = append(data, footer[:]...)
+
+	path := filepath.Join(t.TempDir(), "utxo-set-complete.dat")
+	require.NoError(t, os.WriteFile(path, data, 0o600))
+
+	f, reader := openForReading(t, path)
+
+	utxoWrapperCh := make(chan *utxopersister.UTXOWrapper, 10)
+
+	var received []*utxopersister.UTXOWrapper
+
+	done := make(chan struct{})
+
+	go func() {
+		defer close(done)
+		for w := range utxoWrapperCh {
+			received = append(received, w)
+		}
+	}()
+
+	err := readUTXOWrapperFile(context.Background(), ulogger.TestLogger{}, f, reader, utxoWrapperCh)
+	<-done
+
+	require.NoError(t, err, "a complete, well-formed snapshot file must not be rejected")
+	require.Len(t, received, int(txCount), "the drained channel must yield exactly txCount wrappers")
+}
+
 // TestReadUTXOWrapperFile_TruncatedFile_ReturnsError is the regression test
 // for the silent-truncation bug: a snapshot file cut off partway through a
 // UTXOWrapper record - here, mid-way through the second record's 32-byte
--- a/cmd/utxovalidator/utxo_validator.go
+++ b/cmd/utxovalidator/utxo_validator.go
@@ -11,10 +11,10 @@ import (
 	"io"
 	"os"
 	"strconv"
-	"strings"
 
 	"github.com/bsv-blockchain/go-bt/v2/chainhash"
 	"github.com/bsv-blockchain/go-chaincfg"
+	safeconversion "github.com/bsv-blockchain/go-safe-conversion"
 	"github.com/bsv-blockchain/teranode/errors"
 	"github.com/bsv-blockchain/teranode/pkg/fileformat"
 	"github.com/bsv-blockchain/teranode/services/utxopersister"
@@ -119,17 +119,45 @@ func validateUTXOSet(ctx context.Context, r io.Reader, verbose bool) (*UTXOValid
 
 	for {
 		if err = utxoWrapper.FromReader(ctx, br, false); err != nil {
-			if err == io.EOF {
-				break // End of file reached
-			}
-			// Handle unexpected EOF more gracefully - this can happen at the end of large files
-			errStr := err.Error()
-			if strings.Contains(errStr, "failed to read txid") || strings.Contains(errStr, "unexpected EOF") {
-				fmt.Printf("Reached end of file after processing %d transactions\n", processedWrappers)
-				break
+			// utxopersister writers unconditionally append a 16-byte footer
+			// (txCount||utxoCount) after the final record, with no marker
+			// byte in front of it. The only way this loop can legitimately
+			// be done is if the next read lands exactly on those footer
+			// bytes (reported via ErrRecordBoundary) and the footer's
+			// counts agree with what was actually read here. Anything else
+			// - a bare io.EOF (only possible if the footer itself is
+			// missing), a short read at any other offset, or an
+			// ErrRecordBoundary whose footer bytes don't match - is a
+			// genuine truncation and must fail loudly rather than being
+			// reported as a successful, complete validation.
+			boundary, ok := err.(*utxopersister.ErrRecordBoundary)
+			if !ok {
+				return nil, errors.NewProcessingError("error reading UTXO wrapper", err)
 			}
 
-			return nil, errors.NewProcessingError("error reading UTXO wrapper", err)
+			expectedTxCount, expectedUTXOCount, decErr := utxopersister.DecodeFooter(boundary.FooterBytes[:])
+			if decErr != nil {
+				return nil, errors.NewProcessingError("error decoding utxo-set footer", decErr)
+			}
+
+			processedTxCount, tcErr := safeconversion.IntToUint64(processedWrappers)
+			if tcErr != nil {
+				return nil, errors.NewProcessingError("error converting processed transaction count", tcErr)
+			}
+
+			processedUTXOCount, ucErr := safeconversion.IntToUint64(result.UTXOCount)
+			if ucErr != nil {
+				return nil, errors.NewProcessingError("error converting processed UTXO count", ucErr)
+			}
+
+			if expectedTxCount != processedTxCount || expectedUTXOCount != processedUTXOCount {
+				return nil, errors.NewProcessingError("utxo-set file truncated: footer expects %d transactions/%d utxos, only %d/%d were read",
+					expectedTxCount, expectedUTXOCount, processedTxCount, processedUTXOCount)
+			}
+
+			fmt.Printf("Reached end of file after processing %d transactions\n", processedWrappers)
+
+			break
 		}
 
 		result.ActualSatoshis += utxoWrapper.UTXOTotalValue
--- a/cmd/utxovalidator/utxo_validator.go
+++ b/cmd/utxovalidator/utxo_validator.go
@@ -130,8 +130,8 @@ func validateUTXOSet(ctx context.Context, r io.Reader, verbose bool) (*UTXOValid
 			// ErrRecordBoundary whose footer bytes don't match - is a
 			// genuine truncation and must fail loudly rather than being
 			// reported as a successful, complete validation.
-			boundary, ok := err.(*utxopersister.ErrRecordBoundary)
-			if !ok {
+			var boundary *utxopersister.ErrRecordBoundary
+			if !errors.As(err, &boundary) {
 				return nil, errors.NewProcessingError("error reading UTXO wrapper", err)
 			}
 
--- a/cmd/utxovalidator/utxo_validator_test.go
+++ b/cmd/utxovalidator/utxo_validator_test.go
@@ -1,6 +1,7 @@
 package utxovalidator
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"os"
@@ -9,6 +10,7 @@ import (
 
 	"github.com/bsv-blockchain/go-bt/v2/chainhash"
 	"github.com/bsv-blockchain/teranode/pkg/fileformat"
+	"github.com/bsv-blockchain/teranode/services/utxopersister"
 	"github.com/bsv-blockchain/teranode/ulogger"
 	"github.com/bsv-blockchain/teranode/util/test"
 	"github.com/stretchr/testify/assert"
@@ -58,17 +60,19 @@ func TestValidateUTXOFile_LocalDoesNotDoubleReadMagic(t *testing.T) {
 
 	// Build a minimal UTXO set file: 8-byte fileformat magic + 32-byte
 	// current block hash + 4-byte height + 32-byte previous block hash
-	// + zero UTXO wrappers. The OUTER wrapper loop in validateUTXOSet
-	// breaks on either io.EOF or "failed to read txid", so a body that
-	// ends after the metadata exercises the read path cleanly.
+	// + zero UTXO wrappers + the 16-byte zero-count footer every real
+	// writer appends. The OUTER wrapper loop in validateUTXOSet only
+	// accepts the record stream as complete once it reaches that
+	// footer and its counts agree with what was processed (0/0 here).
 	header := fileformat.NewHeader(fileformat.FileTypeUtxoSet)
-	body := make([]byte, 0, 8+32+4+32)
+	body := make([]byte, 0, 8+32+4+32+16)
 	body = append(body, header.Bytes()...)
 	body = append(body, blockHash[:]...)
 	var heightBuf [4]byte
 	binary.LittleEndian.PutUint32(heightBuf[:], 42)
 	body = append(body, heightBuf[:]...)
 	body = append(body, prevHash[:]...)
+	body = append(body, make([]byte, 16)...) // txCount=0, utxoCount=0
 
 	dir := t.TempDir()
 	path := filepath.Join(dir, "test.utxo-set")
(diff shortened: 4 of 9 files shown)


ordishs (approved):
Verified the fix. The pre-change bug is real, and I reproduced it: five distinct truncations all returned an error that `errors.Is(err, io.EOF)` matched by substring, so the old loop reported success. The footer check rejects all five and accepts a complete file. Both producers count wrappers the same way the reader does, so there is no false positive. Two gaps remain.

**P2 — two sibling read loops keep the silent-truncation logic. `services/utxopersister/UTXOSet.go:689`, `cmd/utxovalidator/utxo_validator.go:124`**

The consolidator matches `err == io.EOF || strings.Contains(err.Error(), "unexpected EOF")`. The utxovalidator also matches `"failed to read txid"`. The PR body presents the truncation risk as closed, but only the seeder read is covered.

Failure path: feed the utxopersister a previous utxo-set cut inside a record TxID, or cut exactly at a UTXO boundary. Both error shapes (`"... -> unexpected EOF"` and a bare `io.EOF`) satisfy the guard at `UTXOSet.go:689`, so the outer loop breaks early. `CreateUTXOSet` then writes a **new** snapshot that omits every UTXO after the cut, with a fresh self-consistent 16-byte footer. That laundered file passes this PR's new check.

One nuance: a cut inside a script renders as `"-> EOF"`, which the consolidator guard does not match, so that sub-case already fails loudly there.

`GetFooter` needs an `*os.File`, and both loops read through `blob.Store.GetIoReader`, so they need a different mechanism — a typed sentinel from `FromReader`, or a footer fetched separately from the store.

**P2 — the new test suite has no complete-file case. `cmd/seeder/seeder_truncation_test.go:131`**

`buildSnapshotWrappers` returns `txCount` and `utxoCount`, and the single test discards both with `_, _`. No test builds a real 16-byte footer and asserts that `readUTXOWrapperFile` accepts the file.

The current code is correct — I built the complete case against the real packages and it matches. The risk is unguarded: a future change to the counting rule, to the skip-empty-wrapper gate at `services/utxopersister/UTXOSet.go:702`, or to the footer layout would make the seeder reject every valid snapshot, and this suite would stay green.

Recommendation: add `TestReadUTXOWrapperFile_CompleteFile_Succeeds`, use the discarded counts, append the footer, then assert the drained channel yields exactly `txCount` wrappers.

Nit: the `GetFooter` doc comment still says "32-byte EOF marker".

services/utxopersister/UTXOSet.go
```diff
@@ -659,37 +658,36 @@ func (us *UTXOSet) CreateUTXOSet(ctx context.Context, c *consolidator) (err erro
...
+						return errors.NewStorageError("error reading previous utxo-set (%s.%s) at iteration %d", c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, txCount, err)
 					}
 
-					return errors.NewStorageError("error reading previous utxo-set (%s.%s) at iteration %d", c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, txCount, err)
+					expectedTxCount, expectedUTXOCount, decErr := DecodeFooter(boundary.FooterBytes[:])
+					if decErr != nil {
+						return errors.NewStorageError("error decoding previous utxo-set (%s.%s) footer", c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, decErr)
+					}
+
+					if expectedTxCount != txCount || expectedUTXOCount != utxoCount {
```
freemans13: This confirms the Critical point the Claude Code Review bot raised on 11 Aug, which came in a few hours after the approval and about two hours after the last commit, so it may not have been seen. I re-derived it independently and then reproduced it, so it isn't a bot false positive.

**The check compares the wrong two numbers, and it fires on every real consolidation — not an edge case.**

`txCount` / `utxoCount` are *survivor* counters. They only increment for wrappers that still have UTXOs left after deletion filtering, and they exist to write the **new** file's footer at lines 741/748:

```go
utxoWrapper.UTXOs = filterUTXOs(utxoWrapper.UTXOs, c.deletions, &utxoWrapper.TxID)  // 696

if len(utxoWrapper.UTXOs) > 0 {                    // 701
    ...
    txCount++                                      // 706
    utxoCount += uint64(len(utxoWrapper.UTXOs))    // 708
}
```

But `expectedTxCount` / `expectedUTXOCount` come from the **previous** file's footer, which records how many records that file actually contains. So the comparison is "records in the old file" vs "records that weren't spent" — those only agree when nothing was spent.

And at this point `c.deletions` holds *exactly* the spends of pre-existing outputs: `processAdditionsFromReader` (consolidator.go:329-330) deletes any deletion whose output was also created inside the range, so what remains is precisely outputs that came from the previous snapshot. Any block range that spends an old coin — i.e. the normal case — leaves survivors < previous total and aborts UTXO-set generation on an intact file.

**Reproduction.** A throwaway test in the package: an intact previous set of two wrappers with a matching `2/2` footer, and a consolidator whose `deletions` map contains one spend of one of those two outputs. Nothing truncated.

```
--- FAIL: TestZZProof_IntactPreviousSetWithDeletions
    Received unexpected error:
    PROCESSING (4): previous utxo-set (7ac32ab9….utxo-set) is truncated:
    footer expects 2 transactions/2 utxos, only 1/1 were read
      - CreateUTXOSet() services/utxopersister/UTXOSet.go:686
```

**Why the suite is green.** All four new `CreateUTXOSet` tests build a consolidator with an empty `deletions` map, so survivors always equal totals and the check passes by accident. `go test ./services/utxopersister/... ./cmd/seeder/... ./cmd/utxovalidator/...` passes and `go vet` is clean; the failing shape is simply never exercised.

**Suggested fix.** Keep a separate counter incremented unconditionally once per wrapper, using the pre-filter UTXO length, and compare *that* against the footer. Leave `txCount`/`utxoCount` as-is — they are correctly the survivor counts for the new file's own footer.

```go
recordsRead++
utxosRead += uint64(len(utxoWrapper.UTXOs))   // before filterUTXOs
```

then compare `expectedTxCount != recordsRead || expectedUTXOCount != utxosRead` here. Worth landing a regression test with a non-empty `deletions` map alongside it, since that's the gap that let this through.

Two smaller things while here: `err.(*ErrRecordBoundary)` on line 675 (and the equivalent in `cmd/utxovalidator`) is a bare type assertion that works only because `NewUTXOWrapperFromReader` returns the error unwrapped — `errors.As` costs nothing and survives anyone adding a wrap later. And the message on line 677 says "at iteration %d" but passes `txCount`, which is a survivor count rather than an iteration index.

For what it's worth, the seeder half of this PR (commit 1) looks correct to me: `readUTXOWrapperFile` counts every wrapper it reads, unfiltered, so the survivor/read distinction never arises there. Same for `cmd/utxovalidator`. It's specifically the `CreateUTXOSet` loop that has the problem.

icellan:
Pushed 80cde7f0d, addressing the reproduced blocker and the smaller items from this round of review.

**Blocker (freemans13, reproduced).** `CreateUTXOSet`'s previous-set read compared `txCount`/`utxoCount` against the previous file's footer, but those two only increment for wrappers that survive `filterUTXOs` - they exist to write the *new* file's own footer. Any consolidation spending a pre-existing UTXO leaves survivors below the previous total, so the check fired a false "truncated" error on every real consolidation. Added `recordsRead`/`utxosRead`, incremented unconditionally before filtering, and compare those against the previous footer instead. `txCount`/`utxoCount` are untouched, still used for the new file's footer.

Added a failing-first regression test (`TestCreateUTXOSet_PreviousSetWithDeletions_NotReportedAsTruncated`): previous set with two wrappers/footer `2/2`, one output spent via `c.deletions`, asserts `CreateUTXOSet` succeeds. Confirmed it reproduced the exact reported error before the fix.

**Smaller items, same commit:**
- `err.(*ErrRecordBoundary)` bare type assertions in `UTXOSet.go` and `cmd/utxovalidator/utxo_validator.go` swapped for `errors.As`.
- The truncation error message logged `txCount` (a survivor count) as the iteration index; now logs `recordsRead`.
- Added `TestReadUTXOWrapperFile_CompleteFile_Succeeds` in `cmd/seeder/seeder_truncation_test.go`, using the previously-discarded `txCount`/`utxoCount` from `buildSnapshotWrappers` to build a real footer and assert the drained channel yields exactly `txCount` wrappers.

**Not touched:** ordishs's P2 on the silent-truncation pattern (`io.EOF`/"unexpected EOF" substring match) in `UTXOSet.go`/`utxo_validator.go` was already closed by ecba9e93a before this pass - both now use the `ErrRecordBoundary` sentinel + footer comparison, not string matching. Verified current code, not just the commit message.

Verified: `go test ./services/utxopersister/... ./cmd/seeder/... ./cmd/utxovalidator/... -race` (all green), `go vet` on the same packages (clean), `go build ./...` (clean), `golangci-lint run` on the touched packages (0 issues).

ctnguyen (changes requested):
Review at commit [ecba9e9](https://github.com/icellan/teranode/commit/ecba9e93a4d23f0a67bbad6bf26af5bf581656d4)

The underlying diagnosis here is excellent and the failure mode is worth closing: `errors.Error.Is` falling back to substring matching means `"unexpected EOF"` satisfies `errors.Is(err, io.EOF)`, so a truncated snapshot walked straight through the seeder's happy path and wrote `lastProcessed.dat`. Replacing that with a typed `ErrRecordBoundary` that hands the candidate footer bytes back to the caller is the right structural fix — it works for non-seekable blob-store readers where `GetFooter` can't, and it kills the substring dependency at the source instead of papering over it. The comments explaining *why* each guard is shaped the way it is are genuinely useful, and the negative-path test coverage (mid-TxID cut, missing footer, mismatched footer, across three packages) is thorough. Two things hold it back: the `CreateUTXOSet` half compares the wrong pair of counters and will abort UTXO-set generation on intact files in normal operation, and the seeder detects truncation only after part of the snapshot has already been committed to the UTXO store.

## Blocking issues

### ChiR1 — `CreateUTXOSet` compares survivor counts against the previous file's record counts

*Same as [@freemans13's point "The check compares the wrong two numbers" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1563#discussion_r3768163724).*

**Problem:** At `repo/services/utxopersister/UTXOSet.go:685` the footer check is `expectedTxCount != txCount || expectedUTXOCount != utxoCount`. But `txCount`/`utxoCount` are only incremented inside the `len(utxoWrapper.UTXOs) > 0` gate at line 701, *after* `filterUTXOs` has removed spent outputs (line 696) — they are survivor counters whose only job is to produce the **new** file's footer at lines 741/748. `expectedTxCount`/`expectedUTXOCount` come from the **previous** file's footer, i.e. how many records that file actually contains. The two quantities only coincide when nothing was spent. Since `processAdditionsFromReader` (`repo/services/utxopersister/consolidator.go:329-330`) removes any deletion whose output was also created inside the range, whatever remains in `c.deletions` is precisely the set of spends against the previous snapshot — so any block range that spends a pre-existing coin makes survivors strictly less than the previous total. A wrapper whose outputs are all spent also disappears entirely from `txCount`, widening the gap further.

**Why it matters:** This is not an edge case; it is the normal case. Every incremental consolidation in `Server.processNextBlock` reads the previous set and applies that block's deletions, so the guard fires on an intact, correctly written file and returns `"previous utxo-set (...) is truncated"`. The `writeSucceeded` defer then aborts the storer, so no new set is produced at all — the persister stalls permanently at the first block that spends an old output, and the error message actively misdirects whoever debugs it toward file corruption. Note also that `txCount` at line 677 is passed as "at iteration %d", which is already a survivor count rather than an iteration index.

**Fix:** Keep a separate pair of read counters incremented unconditionally per record, before filtering, and compare those against the footer. Leave `txCount`/`utxoCount` alone — they are correctly the survivor counts for the new file.

```go
var recordsRead, utxosRead uint64
// ...
if expectedTxCount != recordsRead || expectedUTXOCount != utxosRead {
    return errors.NewProcessingError("previous utxo-set (%s.%s) is truncated: footer expects %d transactions/%d utxos, only %d/%d were read",
        c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, expectedTxCount, expectedUTXOCount, recordsRead, utxosRead)
}
// ...
ts = readStat.AddTime(ts)

recordsRead++
utxosRead += uint64(len(utxoWrapper.UTXOs)) // before filterUTXOs

utxoWrapper.UTXOs = filterUTXOs(utxoWrapper.UTXOs, c.deletions, &utxoWrapper.TxID)
```

### ChiR2 — the seeder validates the footer only after wrappers have already been committed to the UTXO store

**Problem:** The ordering in `repo/cmd/seeder/seeder.go` is: the worker pool starts first (lines 463-469), then `readUTXOWrapperFile` joins the same errgroup (line 515). Inside its loop each wrapper is handed to a worker at line 584 (`case utxoWrapperCh <- utxoWrapper:`) and only *then* are `txProcessed`/`utxosProcessed` incremented (lines 585-586). A worker receives it at line 625 and calls `processUTXO` (line 631), which calls `store.SpendAndCreate(..., utxo.WithCreateOnly(), ...)` at line 674 — one store write per wrapper, no batching and no transaction spanning the import. The footer read (`GetFooter(f)`, line 600) and the count comparison (line 605) happen only after the loop has terminated. So validation runs strictly after the channel sends, workers persist as they go, and there is no rollback on the error path: by the time the mismatch is reported, an indeterminate prefix of the snapshot is already committed to the UTXO store. How much is not knowable from the code — returning the error cancels the errgroup context, and workers take the `<-ctx.Done()` branch at line 621, so wrappers still sitting in the buffered channel (`channelSize`, default 1000) may be dropped unprocessed. The claim here is only that some already-sent wrappers are committed and none are undone, not that the whole prefix is.

**Why it matters:** The PR closes the *marker* half of the risk it describes but not the *store* half. The marker half does work and is worth keeping: the error propagates through `g.Wait()` (line 519), so `lastProcessed.dat` at line 527 is never written and `utxoTip` stays nil, which skips `writeBlockAssemblerState` at line 254. Two things also limit the damage — `WithCreateOnly()` applies no spends, and `processUTXO` short-circuits `errors.ErrTxExists` to nil (lines 683-685), so re-seeding the same or a corrected snapshot over the dirty store is idempotent per transaction. What is left is that the UTXO store holds an unknown partial prefix with nothing recording that fact: the next run's gate keys off `lastProcessed.dat` (line 402) and `blockAssemblerStateExists` (line 420), both absent after a failed import, so it cannot distinguish a fresh empty store from a half-written one, and nothing warns an operator who treats the failure as transient and starts a node against it. To be clear about scope, this is not a regression introduced here — before the change the same partial writes happened *and* were marked complete — the point is that the fix stops one step short of leaving the store in a state that is known to be partial.

**Fix:** Record a durable in-progress marker alongside `lastProcessed.dat` before any worker can write, clear it on success, and gate subsequent runs on it. This does not eliminate the write window — the counts genuinely cannot be known until the records have been read — but it makes the partial state self-describing instead of invisible.

```go
// before starting the workers and the reader
if err = blockStore.Set(ctx, nil, fileformat.FileTypeDat, []byte(utxoFile+"\n"),
    bloboptions.WithFilename("importInProgress"), bloboptions.WithNoHashPrefix()); err != nil {
    return nil, errors.NewStorageError("failed to write import-in-progress marker", err)
}
// ... after g.Wait() returns nil, delete the marker in the same step that writes lastProcessed.dat,
// and extend the pre-flight check near line 402 to refuse to proceed while it exists unless -force.
```

### ChiR3 — no positive-path test anywhere; the suite is green by construction

*Same as [@ordishs's point "the new test suite has no complete-file case" (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1563#pullrequestreview-4907182564).*

**Problem:** Every new test asserts `require.Error`. `buildSnapshotWrappers` in `repo/cmd/seeder/seeder_truncation_test.go:25` computes `txCount` and `utxoCount` and the sole test at line 103 discards both with `_, _` — nothing ever builds a real 16-byte footer and asserts `readUTXOWrapperFile` *accepts* it. The same gap is what hides ChiR1: all four `CreateUTXOSet` tests that exercise the previous-set loop (`repo/services/utxopersister/UTXOSet_test.go:220`, `:271`, `:314`, and the pre-existing `:163`) construct the consolidator via `NewConsolidator(...)` and never populate `c.deletions` — the identifier does not appear once in that file — so survivors always equal totals and the wrong comparison passes by accident. `build_utxoset_test.go`'s end-to-end builds either start at genesis (previous-set read skipped) or stage additions with no spends of prior-snapshot outputs, so they don't reach the shape either.

**Why it matters:** A check that only has failing-side tests is a check that can be inverted, over-tightened, or wired to the wrong variable without anything going red — which is exactly what happened. Given the blast radius (the persister refuses to produce any UTXO set; the seeder would reject every valid snapshot), positive-path coverage is the minimum gate before this lands.

**Fix:** Add `TestReadUTXOWrapperFile_CompleteFile_Succeeds` using the discarded counts and a real footer, asserting the drained channel yields exactly `txCount` wrappers and `readUTXOWrapperFile` returns nil. Alongside the ChiR1 fix, add a `CreateUTXOSet` test with a **non-empty** `deletions` map over an intact previous set.

```go
metadata, wrapperBytes, txCount, utxoCount := buildSnapshotWrappers(t)

var footer [utxopersister.FooterSize]byte
binary.LittleEndian.PutUint64(footer[0:8], txCount)
binary.LittleEndian.PutUint64(footer[8:16], utxoCount)

data := append(append(header.Bytes(), metadata...), bytes.Join(wrapperBytes, nil)...)
data = append(data, footer[:]...)
// ... require.NoError(t, readUTXOWrapperFile(...)) and count the drained wrappers
```

## Non-blocking issues

### ChiR4 — bare type assertions on the new sentinel

*Same as [@freemans13's point "Two smaller things while here" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1563#discussion_r3768163724).*

**Problem:** `repo/services/utxopersister/UTXOSet.go:675` and `repo/cmd/utxovalidator/utxo_validator.go:133` both use `err.(*ErrRecordBoundary)`. This only holds because `NewUTXOWrapperFromReader` and `FromReader` return the sentinel completely unwrapped today — `NewUTXOWrapperFromReader` at `repo/services/utxopersister/UTXO.go:195` does a bare `return nil, err`.

**Why it matters:** The convention everywhere else in these packages is to wrap errors with `errors.NewStorageError`/`NewProcessingError`. The first person who adds a wrap on that return path silently turns both boundary checks into "not a boundary", which converts every clean end-of-file into a hard error — a loud failure, but one introduced by a change that looks completely innocuous. `errors.As` costs nothing and is immune.

**Fix:** Use `errors.As` at both sites.

```go
var boundary *ErrRecordBoundary
if !errors.As(err, &boundary) {
    return errors.NewStorageError("error reading previous utxo-set (%s.%s) after %d records", c.firstPreviousBlockHash.String(), fileformat.FileTypeUtxoSet, recordsRead, err)
}
```

## Recap

| ChiR1 | Survivor counts vs previous footer | YES | 95% |
| ChiR2 | Store written before footer validated | YES | 80% |
| ChiR3 | No positive-path test coverage | YES | 72% |
| ChiR4 | Bare type assertion on sentinel | | 30% |

freemans13 (commented):
Reviewed at `80cde7f0d`.

**My 12-Aug blocker is fixed, and I checked the test has teeth.** `recordsRead`/`utxosRead` now increment unconditionally before `filterUTXOs` and the footer comparison uses them, leaving `txCount`/`utxoCount` to do their real job of writing the new file's footer. I didn't take `TestCreateUTXOSet_PreviousSetWithDeletions_NotReportedAsTruncated` on trust — I reverted just the comparison line back to `expectedTxCount != txCount || expectedUTXOCount != utxoCount` and re-ran it:

```
--- FAIL: TestCreateUTXOSet_PreviousSetWithDeletions_NotReportedAsTruncated
    Received unexpected error:
    PROCESSING (4): previous utxo-set (de8a40c0...d778.utxo-set) is truncated:
    footer expects 2 transactions/2 utxos, only 2/2 were read
```

That is the false "truncated" firing on an intact file, so the test reproduces the reported failure rather than passing alongside it.

Other things I checked and found sound: the counting rules agree on both sides of the format (`CreateUTXOSet` never writes a zero-output wrapper, so a reader counting every record can't drift from a writer counting only non-empty ones; `cmd/bitcointoutxoset` counts every wrapper it emits and the seeder counts every wrapper it reads); `GetFooter`'s "seek is not supported" path can't fire in the seeder because it always arrives via `os.Open`, and `FileStorer.Close` appends nothing after the payload so the footer really is the last 16 bytes; and there's no compatibility break, since both producers have always written the footer. `go test ./services/utxopersister/... ./cmd/seeder/... ./cmd/utxovalidator/... -count=1` is green.

## The one thing left: the consolidator's own input readers still swallow truncation

`processAdditionsFromReader` (`services/utxopersister/consolidator.go:310`) and `processDeletionsFromReader` (`:281`) still end on `errors.Is(err, io.EOF)` — the exact substring match this PR's description identifies as unable to tell "file ended" from "file was cut short". These are the readers behind `GetUTXOAdditionsReader`/`GetUTXODeletionsReader` at `consolidator.go:248` and `:263`, i.e. the production path for every block the persister consolidates.

I built a two-record additions stream and cut it at four points:

| Cut point | Error returned | `errors.Is(err, io.EOF)` |
|---|---|---|
| 5 bytes into the next TxID | `STORAGE_ERROR (69): failed to read txid, expected 32 bytes got 5 -> UNKNOWN (0): unexpected EOF` | **true** |
| exactly 16 bytes into the next TxID | `record stream ended exactly at the footer boundary…` | **true** |
| inside the height/count field | `STORAGE_ERROR (69): failed to read height and number of utxos, expected 8 bytes got 1 -> UNKNOWN (0): unexpected EOF` | **true** |
| exactly at an output boundary inside a record | `EOF` (bare) | **true** |

And calling the loop directly with the stream cut 5 bytes into the second record:

```
processAdditionsFromReader on truncated additions returned: <nil>
additions recorded: 1 (second record's output silently dropped: true)
```

**Why this matters more after this PR than before it.** A truncated `utxo-additions` file means the consolidator misses outputs the block created. `CreateUTXOSet` then writes a **new** snapshot without them, stamped with a footer matching what it wrote — internally consistent, so the seeder's new check, `utxovalidator` and the next consolidation's previous-set check all certify it as complete. A node seeded from that file treats outputs that exist on the network as never created and rejects valid spends of them. The net effect is a real safety net downstream while the upstream hole stays open, which reads as more end-to-end assurance than there is.

Deletions are in better shape by luck rather than design: `NewUTXODeletionFromReader` returns the raw `io.ErrUnexpectedEOF` unwrapped, so the substring fallback never runs and a mid-record cut is reported properly. Only a cut landing exactly on a 36-byte boundary is missed, and no format without counts can catch that.

**On the fix.** The cheap version — `err == io.EOF` instead of `errors.Is` — closes the first three shapes but not the fourth, because `NewUTXOFromReader` (`UTXO.go:393`) returns the reader's bare `io.EOF` when the stream ends part-way through a record's outputs; a record that has started can't legitimately end there, so that read wants converting to an unexpected-end error. Closing the boundary-aligned case too would need additions/deletions files to carry counts the way `utxo-set` files already do.

Either fix it here, or split it out — but in that case please adjust the wording: `ecba9e93a` is titled "close truncation-laundering gap in consolidator and utxo-validator readers", and what it changed is the previous-utxo-set read inside `UTXOSet.go`. The consolidator's own readers weren't touched, and anyone reading the history later will believe this is closed.

## Smaller things

- **Doc drift on the function that gained the new sentinel.** `NewUTXOWrapperFromReader` (`UTXO.go:165-185`) still says it returns "an empty UTXOWrapper with io.EOF error" when "the transaction ID matches the EOF marker (32 zero bytes)". It returns `nil`, that condition can only hold when nothing was read at all, and the doc never mentions `ErrRecordBoundary` — now the most important thing a caller needs to know about it. This PR introduces the sentinel and edits this file, so it's the natural place to correct it.
- **`cmd/filereader` will still bless a truncated snapshot.** `handleUtxoSet` (`file_reader.go:383`) and `handleUtxoAdditions` (`:430`) use the same `errors.Is(err, io.EOF)` break and then print `set contains N UTXO(s)`. Only a diagnostic tool, but it's the tool an operator reaches for to check whether a snapshot is intact.
- **Pre-existing, not this PR's:** `cmd/seeder/seeder.go:512` reads the previous block hash with a bare `reader.Read(...)` rather than `io.ReadFull`; a short read there would misalign every following record. In practice the buffer always has 32 bytes ready, and the new footer check would now make any such misalignment loud — so a tidy-up, not a defect here.

## On @ctnguyen's ChiR2 (partial writes before validation)

Correct on the facts — the footer is at the end, so the counts can't be known until the records have been read, and the workers commit as they go. My own view is that it's separable from this change: the PR already takes the failure from "silently reported as a complete import" to "loudly reported as a failure", with no marker file written and the block assembler state left alone, so the next run doesn't treat the store as seeded. What's left is that the store can't describe itself as half-written. The in-progress marker is cheap and worth doing, but it's its own piece of work with its own test surface, and I'd rather not hold the part that already works behind it.

Nothing in the branch was modified — the mutation and the two probes were local only and have been removed.

ctnguyen (approved):
Review at commit [80cde7f](https://github.com/icellan/teranode/commit/80cde7f0d567515d5f3e792a0c7fddbd807c7416)

This round landed the important one properly. `recordsRead`/`utxosRead` are now incremented unconditionally before `filterUTXOs` and compared against the previous file's footer, `txCount`/`utxoCount` are left alone for the new file's own footer, and the error message reports the read counter instead of the survivor counter — so `CreateUTXOSet` no longer aborts on every intact consolidation that spends a pre-existing coin. The regression test that pins it populates `c.deletions` for real, which is exactly the shape that was missing, and the seeder finally has a happy-path test that builds a genuine footer and asserts the drained channel yields the right number of wrappers. Nothing blocking left. What remains is scope: the sentinel this PR introduces is handed to every wrapper-stream reader in the tree, including ones that read file types with no footer at all, where the guard those readers already use accepts truncation for reasons that mostly predate this change — and the `errors.As` migration doesn't buy quite the robustness we both assumed it did.

## Status of previous points

ChiR2 is deliberately deferred: I accept the author's push-back that a durable in-progress marker plus a pre-flight gate is an import-transactionality feature, not part of this fix, and the partial-write window predates this change. It is not a merge blocker for this PR. Please file the follow-up issue so the store-side half doesn't get lost — the marker half of the risk is genuinely closed here.

| ChiR1 | Survivor counts vs previous footer | YES | FIXED |
| ChiR2 | Store written before footer validated | YES | NOT FIXED |
| ChiR3 | No positive-path test coverage | YES | FIXED |
| ChiR4 | Bare type assertion on sentinel |  | FIXED |

## Non-blocking issues

### ChiR5 - readers of footerless file types decide completeness via `errors.Is(err, io.EOF)`, which accepts any truncation
*Same as [@ordishs's point "two sibling read loops keep the silent-truncation logic" (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1563#pullrequestreview-4907182564).*

**Problem:** ordishs named the two `.utxo-set` read loops and both are fixed. This extends his point to the reader he didn't name: `processAdditionsFromReader` at `repo/services/utxopersister/consolidator.go:310-317`, which reads `utxo-additions` files. Those carry no footer — `NewUTXOSet` writes only block hash + height (`repo/services/utxopersister/UTXOSet.go:161-183`) and the storer is closed with nothing appended (`:359`) — so on that file type *any* short read is truncation and the only legitimate end-of-stream is a zero-byte read. Its guard is `errors.Is(err, io.EOF)`, and two distinct paths satisfy it:

- **(a) exact-16-byte short read — introduced by this PR.** `FromReader` returns the sentinel for every wrapper stream, not just `.utxo-set`: `repo/services/utxopersister/UTXO.go:261-266` has no notion of which file type it is reading. `ErrRecordBoundary.Is` (`UTXO.go:234-236`) then reports it equal to `io.EOF`. The wrapper's `Error()` text says "record stream ended exactly at the footer boundary" and its doc comment at `UTXO.go:206-209` asserts a footer that does not exist for this file type.
- **(b) every other short read — pre-existing, untouched by this PR.** `FromReader` returns `NewStorageError("failed to read txid, expected 32 bytes got %d", …, io.ErrUnexpectedEOF)` (`UTXO.go:268`). `(*Error).Is` falls back to `strings.Contains(e.Error(), target.Error())` for a non-`*Error` target (`repo/errors/errors.go:172-178`), and `io.EOF.Error()` is `"EOF"`, a substring of the rendered `"… -> unexpected EOF"`. So `errors.Is(err, io.EOF)` is true here as well.

Attribution matters: (b) is the older, broader hole and predates this change entirely — the PR neither created nor widened it. (a) is new, and it is not a regression either, because before the change that same 16-byte read took path (b) and was already silent. `cmd/filereader/file_reader.go:385` and `:432` have the identical guard on both file types. `stores/utxo/aerospike/get.go:1682` reads a footerless `FileTypeOutputs` blob but does not test for EOF, so it still errors — only the reported cause is now misleading. Separately, `repo/cmd/seeder/seeder.go:571` is also still on `errors.Is(err, io.EOF)`; the footer check backstops it, but a genuine mid-record read error whose message contains "EOF" is swallowed there and re-reported as `"snapshot file truncated"`, discarding the real cause.

**Why it matters:** A truncated `utxo-additions` file silently drops its tail from the newly written snapshot, which then gets a fresh, self-consistent footer and passes every check this PR adds — the same laundering shape ordishs traced for the `.utxo-set` readers, one layer upstream. Path (b) makes that true for every cut point, not just the 16-byte one, which is why removing the `Is` alias on its own would not fix it.

**Fix:** Two parts, and the first is the one that actually closes the hole. In the footerless readers, stop deciding completeness with `errors.Is`: the only legitimate terminator is the bare `io.EOF` sentinel that `NewUTXOWrapperFromReader` returns on a zero-byte read (`UTXO.go:190-193`), so test identity. That rejects both (a) and (b), since neither `*ErrRecordBoundary` nor a `*Error` is `== io.EOF`. Note the stdlib `errors.Is` is not a safe substitute — it consults `(*Error).Is`, so the substring fallback still fires.

```go
// consolidator.go:312, file_reader.go:385 and :432 — utxo-additions / utxo-set streams
// with no footer: a zero-byte read is the only clean end. Do not use errors.Is here;
// (*Error).Is substring-matches "EOF" against any wrapped "unexpected EOF".
if err == io.EOF { //nolint:errorlint // identity is the point: a wrapped short read is truncation
    return nil
}

return err
```

Then, so the alias cannot reintroduce (a) through some future `errors.Is` call: migrate `cmd/seeder/seeder.go:571` to `errors.As` on the sentinel — it is the only caller relying on the aliasing, and it doesn't need it, since the boundary hands back the footer bytes and that also removes the `GetFooter` seek and the `*os.File` parameter — and drop `ErrRecordBoundary.Is`.

```go
// cmd/seeder/seeder.go, in place of errors.Is(err, io.EOF)
var boundary *utxopersister.ErrRecordBoundary
if !errors.As(err, &boundary) {
    logger.Errorf("Failed to read UTXO: %v", err)
    return errors.NewProcessingError("failed to read UTXO", err)
}

expectedTxCount, expectedUTXOCount, decErr := utxopersister.DecodeFooter(boundary.FooterBytes[:])
if decErr != nil {
    return errors.NewProcessingError("failed to decode snapshot footer", decErr)
}
break OUTER
```

### ChiR6 - `errors.As` does not survive a wrap either; the unwrapped-return invariant is untested
*Same as [@freemans13's point "Two smaller things while here" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1563#discussion_r3768163724).*

**Problem:** Both sites now use `errors.As` (`repo/services/utxopersister/UTXOSet.go:677-680`, `repo/cmd/utxovalidator/utxo_validator.go:133-136`), which is what was asked for — but it does not deliver the property it was asked for. `errors.New` extracts a trailing cause and, when that cause is not an `*Error`, replaces it with a message-only copy: `case error: wErr = &Error{message: err.Error()}` at `repo/errors/errors.go:367-369`. The original value is discarded, so nothing downstream can recover it. Wrap the sentinel with any `errors.New*` constructor and `errors.As(err, &boundary)` fails exactly as the bare type assertion did. So the load-bearing invariant is unchanged: `FromReader` and `NewUTXOWrapperFromReader` must return `*ErrRecordBoundary` completely unwrapped (`repo/services/utxopersister/UTXO.go:195`, `:265`), and nothing in the suite pins that. `repo/services/utxopersister/build_utxoset_test.go:129` also still uses the bare `err.(*ErrRecordBoundary)`.

**Why it matters:** If someone adds a wrap on that return path, all three boundary checks decide "not a boundary" and every clean end-of-file becomes a hard `"truncated"` / `"error reading previous utxo-set"` error — the persister stops producing sets and the validator rejects every file. Loud, but triggered by a change that looks innocuous, and `errors.As` no longer signals that the constraint exists.

**Fix:** Pin the invariant with a test that fails the moment a wrap is introduced, and say so at the return site.

```go
func TestNewUTXOWrapperFromReader_BoundaryIsReachableViaErrorsAs(t *testing.T) {
    // records + footer; the read after the last record lands on the footer
    _, err := NewUTXOWrapperFromReader(context.Background(), bytes.NewReader(footerOnly))

    var boundary *ErrRecordBoundary
    require.True(t, errors.As(err, &boundary),
        "the sentinel must be returned unwrapped: errors.New flattens a non-*Error cause to its message")
}
```

### ChiR7 - `NewUTXOWrapperFromReader`'s doc still describes the removed 32-byte EOF marker
*Same as [@ordishs's point "the GetFooter doc comment still says 32-byte EOF marker" (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1563#pullrequestreview-4907182564).*

**Problem:** The nit was applied to `Footers.go` but the sibling doc it points at was left. `repo/services/utxopersister/UTXO.go:166-182` still says the function returns "an empty UTXOWrapper with io.EOF error" when "the transaction ID matches the EOF marker (32 zero bytes)". No writer emits such a marker, and it could not work if one did: 32 zero bytes read successfully, so `io.ReadFull` returns `nil` and the record is parsed as a wrapper with a zero TxID. The `uw.TxID.IsEqual(&chainhash.Hash{})` conjunct at `repo/services/utxopersister/UTXO.go:190` is dead for the same reason — `uw` is freshly allocated at line 187 and `FromReader` returns `io.EOF` only from `UTXO.go:250-252`, which is reached when `io.ReadFull` read zero bytes and therefore left `TxID` zeroed. The condition is always true on that path.

**Why it matters:** This is the doc the next person reads when deciding how the stream terminates, and it describes a mechanism that does not exist while omitting the sentinel that does — the same staleness the PR correctly fixed one file over.

**Fix:** Rewrite the doc to describe the three real outcomes (a wrapper, a plain `io.EOF` on a zero-byte read, or `*ErrRecordBoundary` on a 16-byte short read) and drop the dead conjunct:

```go
if err := uw.FromReader(ctx, r); err != nil {
    return nil, err
}
```

## Recap

| ChiR5 | Footerless readers accept truncation |  | 50% |
| ChiR6 | Unwrapped-return invariant untested |  | 30% |
| ChiR7 | Stale EOF-marker doc, dead conjunct |  | 18% |

icellan:
Rebased onto current main (04f382b8c). No merge conflicts — one commit (the seeder.go footer-truncation detection) was dropped automatically since its content already landed on main via a related merge; the remaining commits (utxovalidator pre-filter counters, utxopersister footer/UTXOSet changes and tests) applied cleanly. Verified with `go build ./...`, `go vet`, and `go test -race` across cmd/seeder, cmd/utxovalidator, services/utxopersister — all green.

PR #1564, opened 2026-08-10
Add p2p Prometheus metrics and fix pruner metric namespace

icellan:
## Summary
- Add services/p2p/metrics.go, following the same promauto pattern used by the other pipeline services: connected peer count, ban events (labelled by reason), catchup attempts/successes, and websocket subscriber count. Wired into Server.go, HandleWebsocket.go, and handle_catchup_metrics.go.
- Fix services/pruner/metrics.go: metrics were hand-writing the prefix into bare names (pruner_errors_total, pruner_active, etc.) instead of using Namespace: "teranode" / Subsystem: "pruner" like every other instrumented service. This meant pruner series didn't show up under the teranode_* prefix, so dashboard variables and alert templates filtering on that prefix silently missed the pruner.

## Breaking change

All fourteen pruner metrics are renamed from bare `pruner_*` to `teranode_pruner_*` (namespace/subsystem, not string-prefixing). The `errors_total`/`duration_seconds` `operation` label values also now match what `worker.go` actually emits (`preserve_parents`, `expire_preservations`, `dah_pruner`), not the stale set that was previously documented.

Any external Grafana dashboard, Prometheus recording rule, or alert keyed on the old `pruner_*` names or the old label values will break. This is deliberate: the whole point of the rename is to make these series visible under the `teranode_*` prefix that fleet-wide dashboards already filter on. A companion PR updating `bitcoin-sv/teranode-argocd-deployments` is being raised separately.

## Test plan
- [x] go build ./services/p2p/... ./services/pruner/...
- [x] go test ./services/p2p/... ./services/pruner/... -race
- [x] Added services/p2p/metrics_test.go covering the new metrics


ordishs (approved):
Verified. Pre-change `pruner/metrics.go` used bare names with no `Namespace` or `Subsystem`, unlike propagation and blockassembly. The fix applies `teranode` / `pruner` consistently across all fourteen metrics.

Three nits:

- `docs/references/services/pruner_reference.md` and `docs/references/settings/services/pruner_settings.md` still show the pre-fix bare metric names throughout, including in the curl and PromQL examples. These need the same rename.
- `services/p2p/handle_catchup_metrics.go:72-92` — `RecordCatchupFailure` and the block-incomplete-failure path increment no counter. No `prometheusP2PCatchupFailures` symbol exists.
- `services/p2p/Server.go:690-702` — `prometheusP2PConnectedPeers.Set` is called only inside the pre-existing 30-second NAT-logging ticker, so the gauge is stale between ticks and stops updating if that goroutine exits.

freemans13:
Had a careful read through this one. The code is in good shape — I built and ran `go test -race`, `go vet` and `make lint` over `services/p2p/...` and `services/pruner/...` at the head commit and everything is clean, and I couldn't find a correctness bug in the Go. Worth calling out that the websocket gauge is done the right way: `Set(float64(len(cm.channels)))` under the same mutex that guards the map, rather than `Inc`/`Dec`, which makes a leaked count impossible by construction instead of by careful bookkeeping. The pruner namespace fix itself is clearly correct — those series genuinely weren't showing up under the `teranode_*` prefix, so dashboard variables and alert templates filtering on it were silently skipping the pruner.

Three things I'd raise before this merges.

**Have the argocd dashboards been checked for the old pruner names?** Renaming the 14 pruner metrics is a hard break for anything querying `pruner_errors_total`, `pruner_active`, `rate(pruner_duration_seconds_bucket[5m])` and friends. I grepped this repo thoroughly — `deploy/docker/base/grafana_dashboards/`, `compose/grafana/`, `compose/prometheus/`, `deploy/docker/monitoring/`, plus all `*.json`/`*.yaml`/`*.yml`/`*.md`/`*.conf` — and found **zero** in-repo references, so nothing here breaks. But the dashboards and alert rules that matter live in `teranode-argocd-deployments`, which I can't see from this repo. The failure direction is the bad one: a pruner-stall alert would silently stop firing rather than fire falsely, so it fails quiet. If there are references over there, it's probably worth either landing the dashboard change first or adding transitional recording rules (`pruner_errors_total = teranode_pruner_errors_total`) for a release. Either way an operator-facing note in the PR description would help, since there's no CHANGELOG for it to go in.

**The package-level `init()` in `services/p2p/metrics.go`.** This is the only `metrics.go` in the repo with one — `grep -ln "^func init()" services/*/metrics.go` returns just this file. Everywhere else registration happens lazily from the constructor, so a service that isn't enabled exports nothing. Since Teranode is a single binary with per-service flags and `services/p2p` is linked unconditionally by `daemon/daemon_services.go`, the effect is that validator, propagation and blockassembly pods running with `-p2p=0` will now all export `teranode_p2p_connected_peers 0` and `teranode_p2p_websocket_connections 0` — which makes a fleet-wide alert like `teranode_p2p_connected_peers < 1` fire on every non-p2p pod. I think the `NewServer` call at `Server.go:230` is sufficient on its own; if the concern is that p2p unit tests construct `*Server` directly, adding `initPrometheusMetrics()` to `freshTestServer` and a `TestMain` solves that without changing production registration semantics.

**The documented label values don't match what the code emits.** This one is cheap to fix since the PR is already rewriting these exact lines. The docs (`pruner_reference.md:105-107,176-178`, `pruner_settings.md:847`) describe `teranode_pruner_skipped_total{reason="not_running"|"already_in_progress"|"preserve_failed"}`, but `worker.go:64,157,166,171` emits `block_assembly_timeout`, `below_min_height`, `fsm_error`, `catchup_mode` — no overlap at all. Same for `errors_total`: docs say `operation="preserve_parents"|"state_check"`, code emits `parent_preservation`, `expire_preservations`, `dah_pruner` (`worker.go:214,237,252`). So an operator copying the example alert `teranode_pruner_skipped_total{reason="not_running"} > 0` straight out of the settings doc gets an alert that can never fire, while the real signal goes unwatched. The inline comments at `metrics.go:65` and `:75` are stale in the same way.

Related and smaller: `pruner_processed_total` survives un-prefixed in three places (`pruner_dah_pruning.puml:62`, `pruner.md:245`, `pruner_settings.md:917`) — and that metric doesn't exist in the code at all, only `blob_deletion_processed_total` does. The `.puml` line reads half-converted (`teranode_pruner_duration_seconds` on one line, `pruner_processed_total` on the next) and the regenerated SVG bakes it in. `pruner_settings.md:917` tells operators to `grep pruner_processed_total` to verify the pruner is running, which will always come back empty on a healthy node.

A few minor things, take or leave:

- `prometheusP2PBanEvents` only increments in `onPeerBanned`, which is reached solely from ban-score threshold crossings. The `BanPeer` admin RPC and `handleBanEvent` both bypass it, so operator bans don't show up in `rate(teranode_p2p_ban_events_total[5m])` despite the name implying all ban events.
- In `AddBanScore`, an unrecognised non-empty `reason` is logged as a warning and then passed straight through to `WithLabelValues` — bounded in practice today since all in-repo callers use the constants, but setting `reason = ReasonUnknown` in that `default` branch would close it off permanently.
- The six new p2p metrics aren't documented anywhere; `docs/references/prometheusMetrics.md` has sections for 19 services but no P2P one. Feels in scope for a metrics PR.
- The namespace fix stops at the service boundary — `stores/utxo/aerospike/pruner/pruner_service.go:279-292` still registers `utxo_cleanup_batch_duration_seconds` and two others with no Namespace/Subsystem, and `pruner_reference.md` now documents prefixed and unprefixed metrics a few sections apart. Fine as a follow-up, just noting so the convention lands once rather than twice.

freemans13:
One correction to my own suggestion above, on the unrecognised-`reason` point.

I suggested setting `reason = ReasonUnknown` in the `default` branch of `AddBanScore`. That would also change the value passed to `peerRegistry.AddBanScore` on the next line — and the comment right there notes the registry has matching weights per reason, so that risks altering ban scoring behaviour rather than just tidying the metric.

Safer to clamp the metric label only, leaving the registry argument untouched:

```go
prometheusP2PBanEvents.WithLabelValues(normalizeBanReason(reason)).Inc()
```

with `normalizeBanReason` returning `ReasonUnknown` for anything outside the four constants in `ban_reasons.go` — the same shape as `normalizeCatchupFailureKind` in `handle_catchup_metrics.go`, which this PR already uses for the catchup-failure label.

freemans13 (commented):
Reviewed at `eb3c555`. Recording this formally as a comment review — the detail is in my two comments on this PR above, so I won't repeat it here.

Short version: the Go is clean. I built and ran `go test -race`, `go vet` and `make lint` over `services/p2p/...` and `services/pruner/...` at the head commit, all green, and I couldn't find a correctness bug. The pruner namespace fix is clearly right, and the websocket gauge is done the safe way — `Set(float64(len(cm.channels)))` under the mutex that guards the map, so a leaked count is impossible by construction.

Not blocking, but the three things I'd still want an answer on before merge are the argocd dashboard/alert references to the old bare pruner metric names (nothing in this repo, but `teranode-argocd-deployments` isn't visible from here and this fails quiet rather than loud), the package-level `init()` in `services/p2p/metrics.go` making non-p2p pods export `teranode_p2p_connected_peers 0`, and the documented pruner label values not matching what `worker.go` actually emits.

icellan:
Rebased onto main to resolve a merge conflict from an unrelated PR (#1389) that added its own services/p2p/metrics.go independently. Merged the two metric sets (kept prometheusP2PPublishBlocked from main, plus this PR's connected-peers/ban/catchup/websocket metrics) — no functional changes to this PR's own fix. go build/vet/test -race all green on services/p2p and services/pruner.

freemans13 (commented):
Re-reviewed at `de02c8b2`, following on from my pass at `eb3c555`.

**What's improved since last time:** the doc rename is now complete and correct — I grepped `.go`, `.md`, `.json` and `.yml` across the repo and no bare `pruner_*` metric names remain (the only surviving bare `pruner_` hits are settings keys like `pruner_grpcPort`, which are correct). The catchup-failure counter and the dedicated connected-peers ticker are both good additions, and moving the gauge off the NAT-diagnostics goroutine is the right call.

**Verification I ran at this head:** `go build`, `go vet`, `go test -race -count=1` and `golangci-lint run` over `services/p2p/...` and `services/pruner/...` — all green (p2p 16.0s, pruner 6.2s). The one lint hit, `prealloc` at `services/p2p/mock.go:76`, is pre-existing in a file this PR doesn't touch, so `make lint` won't see it in CI.

I agree with the three points already sitting inline from the bots — the missing rows in `docs/references/prometheusMetrics.md`, the unbounded `reason` label at `Server.go:2281`, and the phantom `pruner_processed_total` — so I won't restate them.

---

### The one thing I'd really like fixed before merge

**The pruner label *values* are wrong, in the docs and in the `metrics.go` comments this PR rewrote.**

The PR fixed every pruner metric *name* but left every label *value* untouched, and almost none of them match what `worker.go` actually emits. For a change whose whole purpose is making the pruner metrics match their documentation, that feels like the other half of the same job.

`prunerSkipped` — nothing overlaps at all:

| Source | Values |
|---|---|
| `worker.go:64,157,166,171` (actual) | `block_assembly_timeout`, `below_min_height`, `fsm_error`, `catchup_mode` |
| `metrics.go:65` comment (this PR) | `not_running`, `no_new_height`, `already_in_progress` |
| `pruner_reference.md:105-107` | `not_running`, `already_in_progress`, `preserve_failed` |

`prunerErrors` — only `dah_pruner` overlaps:

| Source | Values |
|---|---|
| `worker.go:214,237,252` (actual) | `parent_preservation`, `expire_preservations`, `dah_pruner` |
| `metrics.go:75` comment (this PR) | `preserve_parents`, `dah_pruner`, `poll` |
| `pruner_reference.md:176-178` | `preserve_parents`, `dah_pruner`, `state_check` |

`prunerDuration` — the code emits a third value nobody documents:

| Source | Values |
|---|---|
| `worker.go:216,239,254` (actual) | `preserve_parents`, `expire_preservations`, `dah_pruner` |
| `metrics.go:55` comment (this PR) | `preserve_parents` or `dah_pruner` |

The practical cost is that the troubleshooting steps at `pruner_reference.md:448` tell an operator to query `teranode_pruner_errors_total{operation="preserve_parents"}`, which will always come back empty — the same silent-miss failure mode the missing `teranode_` prefix caused, just one level down. Lines 55, 65 and 75 of `metrics.go` were rewritten by this PR, so correcting the comments while you're there is a two-minute change; the reference doc tables are a slightly longer edit.

### Still open from my previous review

- **Package-level `init()` in `services/p2p/metrics.go`.** This is the only `init()`-registered metric set in the repo; every other service registers from its constructor. Because the daemon is one binary, it means a pod running only, say, block assembly still exports `teranode_p2p_connected_peers 0` — a gauge that reads as "this node has no peers" rather than "this node isn't the p2p service". A `TestMain` calling `initPrometheusMetrics()` would solve the test problem the comment describes without changing the production path. Your call, but I'd still like it considered.
- **Dashboards outside this repo.** Nothing in-tree references the old bare names, but `teranode-argocd-deployments` isn't visible from here and this breaks quietly rather than loudly. A "breaking: pruner metrics renamed" line in the PR body would help whoever owns those dashboards.

### New, minor

- `services/p2p/metrics_test.go:168` — the second `startConnectedPeersMonitor` discards its `done` channel, so that goroutine keeps ticking and writing the global `prometheusP2PConnectedPeers` gauge after the test returns. Nothing asserts on it later and `-race` is clean, so it's harmless today; it's just inconsistent with the first half of the same test, which does wait on `done`. One line.
- `connectedPeersPollInterval = 5 * time.Second` is hardcoded. Sensible default and cheap to poll — only noting that scrape intervals are usually 15s or more, so most of those polls are discarded.

freemans13:
Following up on the one open question from my review that I couldn't answer earlier — whether anything outside this repo scrapes the old bare `pruner_*` names. I've now checked `bitcoin-sv/teranode-argocd-deployments` at current `origin/main` (`edfda333`), and the answer is yes: **merging this as-is blanks the entire pruner Grafana dashboard.**

`appsets/grafana/application/base/teranode-pruner-dashboard-configmap.yaml` is live — it's registered in `appsets/grafana/application/base/kustomization.yaml` and shipped by the Grafana appset. Of its 39 queries, 13 reference pruner metrics, and **all 13 use the bare names. Zero use `teranode_pruner_*`.** Every one of them silently returns no data the moment this PR lands, across every namespace on dev-scale.

The eleven metrics it queries, all of which need the prefix:

```
pruner_active                          pruner_deleting_children_total
pruner_current_height                  pruner_blob_deletion_processed_total
pruner_duration_seconds (bucket/sum/count)   pruner_blob_deletion_not_found_total
pruner_errors_total                    pruner_blob_deletion_pending
pruner_skipped_total                   pruner_blob_deletion_duration_seconds (bucket/sum/count)
pruner_updating_parents_total
```

Panels affected include Pruner Active, Pruner Current Block Height, the whole Blob Deletion group (rate, pending queue, p95 duration), and the parents-preserved / children-deleted counters.

Two things worth saying about scope, so this doesn't sound worse than it is:

- It's a mechanical fix — prefixing those eleven names in that one file. No PromQL restructuring, no other dashboard in that repo touches pruner metrics (I grepped the whole tree).
- **My label-values finding above does *not* affect these dashboards.** Every one of the 13 queries aggregates with `sum(...) by (namespace)` and none filters on `reason=` or `operation=`, so the wrong documented label values are a docs problem only. Worth separating the two.

The sequencing matters though: the dashboard change and this PR need to land close together, or the pruner panels go dark in the gap. Given the deployments repo is a separate ArgoCD repo, that's a coordination call rather than something to fix here — but it does mean this PR should carry an explicit "breaking: pruner metric names changed, requires a matching teranode-argocd-deployments update" note, and ideally the deployments PR should be open and linked before this merges.

Happy to raise that deployments PR if it's useful — just say the word and I'll put it up for review rather than merging anything.

icellan:
Label values fixed in `08a0bd9fd`. I enumerated every `WithLabelValues` call site in `services/pruner/worker.go` and treated that as the source of truth — no emitting code changed, docs only.

**`skipped_total{reason}`**

| Documented (wrong) | Emitted (right) | Site |
|---|---|---|
| `not_running` | `block_assembly_timeout` | `worker.go:64` |
| `no_new_height` | `below_min_height` | `worker.go:157` |
| `already_in_progress` | `fsm_error` | `worker.go:166` |
| `preserve_failed` | `catchup_mode` | `worker.go:171` |

Nothing overlapped, so all four are replacements rather than a mapping.

**`errors_total{operation}`**

| Documented (wrong) | Emitted (right) | Site |
|---|---|---|
| `preserve_parents` | `parent_preservation` | `worker.go:214` |
| `state_check` / `poll` | `expire_preservations` | `worker.go:237` |
| `dah_pruner` | `dah_pruner` (unchanged) | `worker.go:252` |

**`duration_seconds{operation}`**

| Documented | Emitted | Site |
|---|---|---|
| `preserve_parents` | `preserve_parents` (unchanged) | `worker.go:216` |
| — (missing) | `expire_preservations` | `worker.go:239` |
| `dah_pruner` | `dah_pruner` (unchanged) | `worker.go:254` |

Changed in:

- `services/pruner/metrics.go` — the label comments at the `duration_seconds`, `skipped_total` and `errors_total` definitions, plus the doc-comment summary line on `_initPrometheusMetrics`.
- `docs/references/services/pruner_reference.md` — the three label tables and examples, and the High Error Rate troubleshooting query that told operators to query `operation="preserve_parents"` on `errors_total` (always empty).
- `docs/references/settings/services/pruner_settings.md` — the Monitoring Settings line referencing `skipped_total{reason="not_running"}`.

One thing I left alone deliberately: phase 1 emits `errors_total{operation="parent_preservation"}` but `duration_seconds{operation="preserve_parents"}` for the same phase — the same phase under two different names across two metrics. Its siblings (`expire_preservations`, `dah_pruner`) are identical across both. That looks like a code-side inconsistency rather than a doc bug, so I documented what is emitted instead of changing it. Worth deciding separately whether to align the emitted values, which would be another breaking change for anything already keyed on them.

Verified at this head: `go build ./...`, `go vet ./services/pruner/...`, `go test ./services/pruner/... -race` — all pass.

icellan:
@freemans13 on the argocd dashboard — yes, please raise that PR. Take the offer up.

Scope: the 13 queries in `appsets/grafana/application/base/teranode-pruner-dashboard-configmap.yaml` still on the bare `pruner_*` names need to move to the `teranode_pruner_*` namespace.

Note the label values changed too, so it isn't a pure name substitution — `08a0bd9fd` in this PR corrects the documented values to match what `worker.go` actually emits. Anything keyed on `skipped_total{reason=...}` is affected: all four documented values were wrong (`not_running`/`no_new_height`/`already_in_progress`/`preserve_failed` → `block_assembly_timeout`/`below_min_height`/`fsm_error`/`catchup_mode`), and `errors_total{operation=...}` had two wrong (`preserve_parents` → `parent_preservation`, `state_check`/`poll` → `expire_preservations`). Worth checking the dashboard queries against that list rather than against the old docs.

Merge order: land the deployments PR first, or same day, so the dashboard isn't blank in between.

icellan:
Pushed 83e73a2a5: fixes a label-value mismatch between `errors_total` and `duration_seconds` for pruner phase 1. `errors_total` was emitting `operation="parent_preservation"` while `duration_seconds` used `operation="preserve_parents"` for the same phase — the other two phases already agree (`expire_preservations`, `dah_pruner`). Standardised on `preserve_parents` (verb+object shape matching `expire_preservations`).

Changed: `services/pruner/worker.go`, `services/pruner/metrics.go` (doc comments), `docs/references/services/pruner_reference.md`.

@freemans13 this affects the dashboard migration you're doing in teranode-argocd-deployments: `errors_total{operation="parent_preservation"}` is now `operation="preserve_parents"`. Same value as `duration_seconds{operation="preserve_parents"}`, which was already correct before this commit.

ctnguyen (changes requested):
Review at commit [de02c8b](https://github.com/icellan/teranode/commit/de02c8b2d3411b0a2c777359b69d5fd58795fa14)

The pruner half of this is unambiguously right — those fourteen series genuinely were invisible to anything filtering on `teranode_*`, and fixing it via `Namespace`/`Subsystem` rather than string-prefixing the names is the correct shape. On the p2p side, the websocket gauge is done the way I'd want it: `Set(float64(len(cm.channels)))` inside the same mutex that guards the map, so the count can't drift no matter how many times `remove` is called for the same channel (and `broadcast`'s timeout path does call it a second time). Pulling the connected-peers gauge onto its own ticker instead of piggybacking the 30s NAT-diagnostics goroutine is the right call, and `TestRecordCatchupFailure_IncrementsFailureCounter` is the shape I want from a metrics test — it drives the real gRPC handler rather than the collector. My concerns are mostly about what the metrics *mean* and what breaks downstream, not about the Go.

## Blocking issues

### ChiR1 — the connected-peers gauge disagrees with this repo's own definition of a connected peer

**Problem:** `updateConnectedPeersGauge` (`repo/services/p2p/connected_peers_monitor.go:48-53`) sets the gauge to `len(s.P2PClient.GetPeers())` — the raw length, with no filtering. Every other place in this package that answers "how many peers are we connected to" filters first, and says why:

- `getNodeStatusMessage` (`repo/services/p2p/Server.go:1609-1627`) computes the `connected_peers_count` field of `node_status` from `peerRegistry.ListPeers`, counting only `p.IsConnected`, under the comment *"The registry also holds gossiped/disconnected peers, so count only directly connected ones."* `server_helpers_test.go:148-160` pins that behaviour with the assertion "gossiped/disconnected registry peers must not be counted".
- The `GetPeers` RPC skips `!p.IsConnected` on its registry path (`Server.go:2088`).
- The same RPC's libp2p fallback — the path that actually reads `s.P2PClient.GetPeers()` — drops entries with `sp.Addrs == nil`, entries whose first address is empty, and the local node ID (`Server.go:2117-2138`).
- `Interface.go:24` spells the distinction out: `IsConnected bool // Whether this peer is directly connected (vs gossiped)`.

So the one existing consumer of `GetPeers()` in this file treats its elements as needing to be filtered down to real, addressable, non-self peers, and the new gauge applies none of those filters. Note the limit on what I can show from this tree: exactly how far `len(GetPeers())` diverges from the connected count at runtime depends on the peer bookkeeping inside `go-p2p-message-bus` (pinned at v0.1.17 in `go.mod`), which isn't checked in here. What is visible here is the disagreement itself — two in-repo readers filter, the new gauge doesn't — and that `Server.go:2117-2138` exists at all is the repo's own evidence that unfiltered elements are not all connected peers.

One consequence is already baked into the tests: `MockServerP2PClient.GetPeers` returns bare `PeerInfo{}` values with no addresses (`repo/services/p2p/mock.go:71-80`), and `TestUpdateConnectedPeersGauge_ReflectsLivePeerCount` asserts the gauge equals that raw count. Those fixtures would need real addresses once the gauge filters.

**Why it matters:** A metric named `connected_peers`, with help text "Number of peers currently connected to this node over the p2p network", that reports a different number than the `connected_peers_count` the dashboard already shows for the same node, is a metric operators will mistrust the first time they compare the two — and they will compare, because this is the gauge you alert on. It also skews in the permissive direction: unfiltered is `>=` filtered, so `teranode_p2p_connected_peers < 1` is the alert most likely to under-fire.

**Fix:** Source the gauge from the registry so it is the same number as `node_status`, and thread a context through the monitor.

```go
// updateConnectedPeersGauge sets prometheusP2PConnectedPeers to the number of
// DIRECTLY connected peers, using the same definition as the
// connected_peers_count field in node_status (Server.go:1609-1627): the
// registry also holds gossiped/disconnected peers.
func (s *Server) updateConnectedPeersGauge(ctx context.Context) {
	if s.peerRegistry == nil {
		return
	}
	if err != nil {
		// Leave the previous value rather than publish a wrong one.
		s.logger.Warnf("[updateConnectedPeersGauge] ListPeers failed: %v", err)
		return
	}
	connected := 0
	for _, p := range peers {
		if p.IsConnected {
			connected++
		}
	}
	prometheusP2PConnectedPeers.Set(float64(connected))
}
```

If you'd rather not pay a gRPC hop every `connectedPeersPollInterval` (5s is already ~3× a typical scrape interval), read the value the periodic node-status publisher has already computed and cached: `if st := s.latestNodeStatus.Load(); st != nil { prometheusP2PConnectedPeers.Set(float64(st.ConnectedPeersCount)) }`. That is free and agrees with the dashboard by construction. Staying on `GetPeers()` is defensible only if the gauge applies the same filter the RPC fallback does (`len(p.Addrs) > 0`, skip `s.P2PClient.GetID()`).

### ChiR2 — the pruner rename blanks the live pruner Grafana dashboard, with no note in the PR

*Same as [@freemans13's point (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1564#issuecomment-5329770836).*

**Problem:** Renaming the fourteen pruner series is a hard break for any query using the old bare names. Nothing in this repo references them — I confirmed there are no hits under `deploy/` or `compose/` — but the dashboard that matters lives in `bitcoin-sv/teranode-argocd-deployments`, and @freemans13 verified at `edfda333` that `appsets/grafana/application/base/teranode-pruner-dashboard-configmap.yaml` is live (registered in the base `kustomization.yaml`) and that **all 13** of its pruner queries use the bare names, none the prefixed form. Panels affected include Pruner Active, Current Block Height, the whole Blob Deletion group, and the parents-preserved / children-deleted counters. The PR body still carries no breaking-change note and links no deployments PR.

**Why it matters:** The failure direction is the bad one — a pruner-stall alert stops firing rather than fires falsely, and the panels go blank rather than red, across every namespace. There is no CHANGELOG for this to land in, so whoever owns those dashboards has nothing to react to. This is a sequencing problem rather than a code problem, but it has to be resolved before merge rather than after.

**Fix:** Put an explicit note in the PR description — "breaking: pruner metric names now prefixed `teranode_pruner_*`, requires a matching `teranode-argocd-deployments` update" — and open/link the deployments PR so the two land together. @freemans13 has offered to raise it. If the gap can't be closed to near-zero, ship transitional recording rules for one release so both names resolve:

```yaml
groups:
  - name: pruner-metric-rename-compat
    rules:
      - record: pruner_errors_total
        expr: teranode_pruner_errors_total
      - record: pruner_active
        expr: teranode_pruner_active
      # ... one per renamed series
```

### ChiR3 — package-level `init()` makes every non-p2p pod export the p2p metrics

*Same as [@freemans13's point 2 (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1564#issuecomment-5267695823).*

**Problem:** `repo/services/p2p/metrics.go:56-58` registers the whole p2p metric set from a package `init()`. This is the only `metrics.go` in `services/` that does so — every other service registers lazily from its constructor, so a service that isn't enabled exports nothing. Because Teranode is one binary and `services/p2p` is imported unconditionally by `daemon/daemon_services.go` (line 23, and again via `p2p.ClientI` in the asset, rpc and alert wiring), the `init()` runs in every pod. A validator or blockassembly pod started with `-p2p=0` will now export `teranode_p2p_connected_peers 0` and `teranode_p2p_websocket_connections 0`. Both `initPrometheusMetrics()` calls in `NewServer` (`Server.go:238` and `:459`) become dead as a result.

**Why it matters:** A gauge reading 0 is indistinguishable from "this node has no peers", so any fleet-wide rule on `teranode_p2p_connected_peers` fires on every pod that isn't running p2p. And it's one-way: once a series is exported across the fleet you can't quietly withdraw it. The stated justification — that unit tests construct `*Server` directly — is a test concern, and there's no `TestMain` in the package yet, so it can be solved without changing production registration.

**Fix:** Drop the `init()` and initialize from the test entry point instead.

```go
// services/p2p/main_test.go
func TestMain(m *testing.M) {
	initPrometheusMetrics() // tests build *Server directly, bypassing NewServer
	os.Exit(m.Run())
}
```

Keep the `initPrometheusMetrics()` call in `NewServer`; `sync.Once` still makes both paths safe.

### ChiR4 — the documented pruner label *values* don't match what `worker.go` emits, including in comments this PR rewrote

*Same as [@freemans13's point "The one thing I'd really like fixed before merge" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1564#pullrequestreview-4962287832).*

**Problem:** Every metric *name* was fixed and every label *value* was left alone. `repo/services/pruner/metrics.go:65` documents `prunerSkipped` reasons as `not_running`, `no_new_height`, `already_in_progress`; `worker.go:64,157,166,171` emits `block_assembly_timeout`, `below_min_height`, `fsm_error`, `catchup_mode` — zero overlap. Line 75 documents `prunerErrors` operations as `preserve_parents`, `dah_pruner`, `poll`; `worker.go:214,237,252` emits `parent_preservation`, `expire_preservations`, `dah_pruner` — only `dah_pruner` matches. Line 55 documents two duration operations; `worker.go:216,239,254` emits three (`expire_preservations` is undocumented). The reference doc repeats the same wrong values at `docs/references/services/pruner_reference.md:105-107` and `:176-178`, and the troubleshooting step at `:448-449` tells an operator to query `teranode_pruner_errors_total{operation="preserve_parents"}`. Lines 55, 65 and 75 of `metrics.go` were rewritten by this diff.

**Why it matters:** An operator copying `teranode_pruner_skipped_total{reason="not_running"} > 0` straight out of the docs gets an alert that can never fire, while the reason that actually trips (`block_assembly_timeout`) goes unwatched. That's exactly the silent-miss failure the missing `teranode_` prefix caused, and for a PR whose stated purpose is making the pruner metrics match their documentation it's the other half of the same job.

**Fix:** Correct the three inline comments while you're in the file, and the two reference tables.

```go
	[]string{"operation"}, // "preserve_parents", "expire_preservations", "dah_pruner"
	...
	[]string{"reason"},    // "block_assembly_timeout", "below_min_height", "fsm_error", "catchup_mode"
	...
	[]string{"operation"}, // "parent_preservation", "expire_preservations", "dah_pruner"
```

Then update `pruner_reference.md:105-107` / `:176-178` / `:448-449` and `pruner_settings.md:847` to the emitted values. Worth adding a note that these are code-derived so the next reader knows to re-check `worker.go` rather than trust the table.

## Non-blocking issues

### ChiR5 — `pruner_processed_total` survives un-prefixed and doesn't exist in the code

*Same as [@freemans13's point "Related and smaller" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1564#issuecomment-5267695823).*

**Problem:** Three docs still reference `pruner_processed_total`: `docs/topics/services/img/plantuml/pruner/pruner_dah_pruning.puml:62`, `docs/topics/services/pruner.md:245`, and `docs/references/settings/services/pruner_settings.md:917`. No such metric exists in `services/pruner/metrics.go` — the closest is `blob_deletion_processed_total`. The `.puml` line is now half-converted (`teranode_pruner_duration_seconds` immediately followed by bare `pruner_processed_total`), and the regenerated SVG bakes that inconsistency into a committed image.

**Why it matters:** `pruner_settings.md:917` instructs operators to `grep pruner_processed_total` to confirm the pruner is running; on a perfectly healthy node that comes back empty, which reads as a dead pruner. Small, but it's the one doc line whose whole purpose is "is this thing alive".

**Fix:** Replace it with a metric that exists, then regenerate the SVG.

```
Pruner -> Pruner: Record Metrics:\nteranode_pruner_duration_seconds\nteranode_pruner_deleting_children_total
```

and change the settings-doc check to `grep teranode_pruner_deleting_children_total` (or `teranode_pruner_active`, which is the more direct liveness signal).

### ChiR6 — the six new p2p metrics aren't in the metrics reference

*Same as [@freemans13's point "The six new p2p metrics aren't documented anywhere" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1564#issuecomment-5267695823).*

**Problem:** `docs/references/prometheusMetrics.md` now has a P2P Service Metrics section (`:199-203`), added by #1389, but it contains a single row for `teranode_p2p_publish_blocked_total`. None of `connected_peers`, `ban_events_total`, `catchup_attempts_total`, `catchup_successes_total`, `catchup_failures_total` or `websocket_connections` are listed. This extends @freemans13's original note, which predated the rebase and observed that there was no P2P section at all — the section exists now, so this is a six-row table edit.

**Why it matters:** That file is where an operator looks to find out what's scrapeable, and it's the same discoverability gap the pruner namespace fix is about. A metrics PR that adds six series and documents none of them leaves the reference immediately stale.

**Fix:** Add the six rows to the existing table, with the label sets (`reason` on ban events, `kind` on catchup failures) and the enumerated values.

### ChiR7 — the ban counter misses operator/external bans and takes an unbounded `reason`

*Same as [@freemans13's point (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1564#issuecomment-5268599228), extending his earlier minor notes.*

**Problem:** Two things about `prometheusP2PBanEvents.WithLabelValues(reason).Inc()` at `repo/services/p2p/Server.go:2281`. First, `onPeerBanned` is reached only from ban-score threshold crossings (`AddBanScore` at `:2245`, `applyBanScore` at `:2263`). The `BanPeer` RPC (`:2144-2151`) writes straight to `banList`, and `handleBanEvent` (`server_helpers.go:1173`) applies externally gossiped bans — neither increments, so `rate(teranode_p2p_ban_events_total[5m])` under-reports despite the name implying all ban events. Second, `AddBanScore` (`:2226-2237`) logs a warning for an unrecognised non-empty `reason` and then passes it through unchanged, so a mis-versioned caller can mint arbitrary label values.

**Why it matters:** In-repo callers all use the `ban_reasons.go` constants today, so cardinality is bounded in practice — but the label originates from a gRPC field, which is exactly the kind of boundary you don't want feeding a `CounterVec` directly. The coverage gap is the more likely operational surprise: an operator bans a peer by hand, sees nothing in the metric, and concludes the ban didn't take.

**Fix:** Clamp the label at the metric only, leaving the value handed to `peerRegistry.AddBanScore` untouched (the registry has per-reason weights, so changing that argument would alter scoring):

```go
prometheusP2PBanEvents.WithLabelValues(normalizeBanReason(reason)).Inc()
```

with `normalizeBanReason` returning `ReasonUnknown` for anything outside the four constants — same shape as `normalizeCatchupFailureKind`. For coverage, either increment in `BanPeer`/`handleBanEvent` with a distinguishing reason (`operator`, `gossip`) or narrow the help text and metric name to score-threshold bans specifically.

### ChiR8 — three of the six new metrics are asserted only by poking the collector

*Partly the same as [@freemans13's point "services/p2p/metrics_test.go:168" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1564#pullrequestreview-4962287832); his note covers the leaked monitor goroutine, the coverage gap below is additional.*

**Problem:** `TestPrometheusMetricsIncrementAndSet` (`repo/services/p2p/metrics_test.go:71-89`) is documented as exercising "the increment/set call sites used by the actual wiring", but it calls `Inc()`/`Set()` directly on the collectors and never enters production code. For three of the six new metrics that is the only assertion there is:

- `prometheusP2PCatchupAttempts` — incremented at `handle_catchup_metrics.go:51`.
- `prometheusP2PCatchupSuccesses` — incremented at `:70`.
- `prometheusP2PBanEvents` — incremented at `Server.go:2281`.

To be precise about the gap: all three increments *are* executed by existing tests — `handle_catchup_metrics_test.go:55,76` drive the real handlers, and `ban_enforcement_test.go:95` and `server_helpers_test.go:667,679` call `onPeerBanned` — but nothing asserts the counters moved. So the lines run and are covered for panic-freedom; what isn't pinned is that they run *on the success path and only there*. The other three metrics are properly covered: `catchup_failures_total` via `TestRecordCatchupFailure_IncrementsFailureCounter`, `websocket_connections` via `cm.add`/`cm.remove`, `connected_peers` via `updateConnectedPeersGauge`.

Two smaller test-hygiene items in the same file. The second `startConnectedPeersMonitor` at `:166-172` discards its `done` channel; `defer cancel2()` does stop it, but the test returns before the goroutine observes the cancel, so it can still write the global gauge after the test finishes — inconsistent with the first half of the same test, which waits on `done`. And `TestClientChannelMapUpdatesWebsocketConnectionsGauge` (`:94-113`) asserts absolute values 1/2/1/0 on a process-wide gauge fed by a per-instance `clientChannelMap`, where every other assertion in the file correctly uses a before/after delta.

**Why it matters:** The two catchup counters sit after an early `return` on the registry error (`handle_catchup_metrics.go:47-49`, `:66-68`); that ordering is the interesting property and nothing guards it. Move an `Inc()` above the error check, or drop it, and the suite stays green. The hygiene items break nothing today — no later test in the package reads these gauges, and there is exactly one `clientChannelMap` in production (`HandleWebsocket.go:310`, reached only from `Server.go:722`) — but they're latent ordering fragility.

**Fix:** Follow the pattern the failure-counter test already sets, for the two counters whose handlers are already under test:

```go
func TestRecordCatchupAttempt_IncrementsCounter(t *testing.T) {
	before := testutil.ToFloat64(prometheusP2PCatchupAttempts)

	s, reg, pid := freshTestServer(t)
	reg.Register(&blockchain.PeerInfo{ID: pid.String()})

	resp, err := s.RecordCatchupAttempt(context.Background(), &p2p_api.RecordCatchupAttemptRequest{PeerId: pid.String()})
	require.NoError(t, err)
	require.True(t, resp.Ok)
	require.Equal(t, before+1, testutil.ToFloat64(prometheusP2PCatchupAttempts))

	// and the negative case the early return is there for
	_, err = s.RecordCatchupAttempt(context.Background(), &p2p_api.RecordCatchupAttemptRequest{PeerId: "not-a-peer-id"})
	require.Error(t, err)
	require.Equal(t, before+1, testutil.ToFloat64(prometheusP2PCatchupAttempts))
}
```

An equivalent assertion around the existing `onPeerBanned` call in `ban_enforcement_test.go` covers the ban counter. Separately: wait on the second monitor (`done2 := s.startConnectedPeersMonitor(...); cancel2(); <-done2`), switch the websocket assertions to deltas, and either fix or drop the "actual wiring" wording in the `TestPrometheusMetricsIncrementAndSet` comment.

## Recap

|---|---|---|---|
| ChiR1 | Connected-peers gauge unfiltered | YES | 85% |
| ChiR2 | Rename blanks pruner dashboard | YES | 80% |
| ChiR3 | `init()` exports p2p metrics everywhere | YES | 65% |
| ChiR4 | Documented pruner label values wrong | YES | 60% |
| ChiR5 | Phantom `pruner_processed_total` |  | 45% |
| ChiR6 | New p2p metrics undocumented |  | 40% |
| ChiR7 | Ban counter gaps, unbounded label |  | 35% |
| ChiR8 | Counters asserted without call sites |  | 25% |

icellan:
Pushed 58297a088 addressing the remaining review findings.

**Already fixed at head (before this commit), no action taken:**
- ChiR4 (pruner label values in comments/docs) — `08a0bd9fd`, phase 1 further aligned in `83e73a2a5`.
- ChiR2 (external Grafana dashboard breakage) — freemans13 raising the companion `teranode-argocd-deployments` PR. Added the breaking-change note to this PR's description covering both the namespace rename and the label-value fix.

**ChiR3 (blocker) — package init() exporting p2p metrics everywhere.** Removed the `init()` in `services/p2p/metrics.go`. Registration now happens only via `initPrometheusMetrics()` from `NewServer`, plus a package `TestMain` (`services/p2p/main_test.go`) for the tests that construct `*Server` directly. Added `services/p2p/metricsregistrationtest/no_init_registration_test.go`, a standalone test package that only imports `services/p2p` (no `NewServer` call anywhere in that test binary) and asserts no `teranode_p2p_*` family is registered — this is the actual regression check for the bug, since anything inside `services/p2p`'s own test binary will always register via `TestMain`.

**ChiR1 (blocker) — connected-peers gauge unfiltered.** `updateConnectedPeersGauge` now reads `s.peerRegistry.ListPeers` and counts only `IsConnected` entries, exactly matching `getNodeStatusMessage`'s `connected_peers_count` (Server.go:1609-1627). Went with the registry read over the cached `latestNodeStatus` alternative: it's the same call `GetPeers`/`getNodeStatusMessage` already make on every poll, it doesn't depend on the status publisher having run at least once, and it fails closed (a `ListPeers` error now leaves the gauge at its previous value instead of publishing zero or a stale-cache value silently). Added a test for that error path plus a test confirming gossiped/disconnected registry entries are excluded.

**ChiR7 (nit) — ban counter gaps + unbounded label.** Bounded the `reason` label to a known set (`normalizeBanReasonLabel` in `ban_reasons.go`); the raw value passed to `peerRegistry.AddBanScore` is untouched so scoring weights aren't affected. For the missing `BanPeer` RPC path, added a direct increment in `BanPeer` (`Server.go`) under a new `operator_ban` reason. I did **not** add an increment in `handleBanEvent` as suggested — traced the ban-event channel and `onPeerBanned`'s own IP-ban follow-up call (`s.banList.Add`) pushes onto the same channel `handleBanEvent` reads from, so incrementing there would double-count every score-threshold ban already counted at `onPeerBanned`. Also, no code anywhere in the tree currently constructs a `BanEvent` with a non-empty `PeerID`, so the "gossip" branch the review assumed is reachable is dead code today. Net: operator bans are now counted, threshold bans are still counted once (at `onPeerBanned`), nothing double-counts.

**ChiR5 (nit) — phantom `pruner_processed_total`.** Confirmed via grep, no such metric exists. Replaced all three references (`.puml`, `pruner.md`, `pruner_settings.md`) with `teranode_pruner_deleting_children_total` (diagram/metrics list) and `teranode_pruner_active` (the settings-doc liveness check). Regenerated the `.svg` from the updated `.puml`.

**ChiR6 (nit) — undocumented p2p metrics.** Added all six rows to `docs/references/prometheusMetrics.md`'s P2P Service Metrics table, with label sets and enumerated values.

**ChiR8 (nit) — counters asserted without call sites.** Added `TestRecordCatchupAttempt_IncrementsCounter`, `TestRecordCatchupSuccess_IncrementsCounter`, `TestOnPeerBanned_IncrementsBanEventsCounter` and `TestBanPeer_IncrementsBanEventsCounter`, all driving the real gRPC handlers rather than poking collectors. Fixed the leaked-ticker case in `TestStartConnectedPeersMonitor_UpdatesGaugeOnItsOwnTicker` (now waits on the second monitor's `done` channel too) and switched `TestClientChannelMapUpdatesWebsocketConnectionsGauge` to a pinned-baseline delta instead of absolute values. Reworded `TestPrometheusMetricsIncrementAndSet`'s comment to stop claiming it exercises production call sites.

Verification: `go build ./...`, `go vet ./services/p2p/... ./services/pruner/...`, `go test ./services/p2p/... ./services/pruner/... -race` all green; `golangci-lint run` clean on touched files.

freemans13 (commented):
# PR 1564 — review round 3, at `83e73a2a`

*Add p2p Prometheus metrics and fix pruner metric namespace* — `icellan`,
branch `fix/audit-19-p2p-pruner-metrics`. Third pass; earlier passes were at
`eb3c555` (12 Aug) and `de02c8b2` (18 Aug). This one covers the two commits
Siggi pushed on 20 Aug and re-checks everything that was still open.

Because this is someone else's branch I've changed nothing in the working
tree — everything below is feedback.

## The headline

The one thing I asked for before merge is done. The pruner label *values* in
the docs and in the `metrics.go` comments now match what `worker.go` actually
emits, and `worker.go:214` was corrected from `parent_preservation` to
`preserve_parents` so the errors counter and the duration histogram finally
agree on the name of phase 1. I re-derived the full set from the code rather
than trusting the diff:

| Metric | Values `worker.go` emits | Now documented? |
|---|---|---|
| `prunerSkipped` | `block_assembly_timeout` (`:64`), `below_min_height` (`:157`), `fsm_error` (`:166`), `catchup_mode` (`:171`) | yes, all four |
| `prunerErrors` | `preserve_parents` (`:214`), `expire_preservations` (`:237`), `dah_pruner` (`:252`) | yes, all three |
| `prunerDuration` | `preserve_parents` (`:216`), `expire_preservations` (`:239`), `dah_pruner` (`:254`) | yes, all three |

That closes ChiR4 in Chi's review and the "one thing I'd really like fixed"
from my 18 Aug pass. Nothing else on the branch moved.

## Verification I ran at `83e73a2a`

Green: `go build`, `go vet`, and `go test -race -count=1` over
`./services/p2p/...` and `./services/pruner/...` (p2p 16.1s, pruner 6.1s).
`golangci-lint run` over both packages reports exactly one issue,
`prealloc` at `services/p2p/mock.go:76`, which is in a file this branch never
touches — pre-existing, and `make lint` (changed files only) won't see it.

I did not run the smoke or sequential suites; they're too heavy locally, so
treat CI as the authority on those.

## New this round

Two things I hadn't spotted before, both of which follow directly from the
work this PR is doing.

### N1 — two pruner skip paths still emit no metric at all

The label-value fix resolved the mismatch by deleting the reasons that didn't
exist. Worth noting that one of them described a real code path that is still
uncounted, and there's a second one beside it:

- `worker.go:150-152` — `if blockHeight <= s.lastProcessedHeight.Load() { continue }`.
  This is a duplicate or out-of-order notification, and in practice it is
  probably the most frequent skip the pruner takes. It increments nothing and
  logs nothing. The old (wrong) `no_new_height` label was describing exactly
  this, so the docs fix removed the name rather than wiring up the counter.
- `worker.go:184-186` — `if !s.waitForBlockMinedStatus(ctx, &sig.blockHash) { continue }`.
  When `mined_set` never arrives within `pruner_blockAssemblyWaitTimeout`, the
  block drops out of pruning entirely. The only trace is a `Debugf` inside
  `waitForBlockMinedStatus`, so at a normal log level this is invisible in both
  logs and metrics.

The second is the one that matters operationally: it's the "the pruner quietly
stopped pruning" failure, and there is nothing to alert on. Two extra
`prunerSkipped.WithLabelValues(...)` calls — say `already_processed` and
`mined_set_timeout` — would close it, and they belong in this PR more than
anywhere else since it's the PR that made the skip taxonomy honest. Not
blocking, but cheap.

### N2 — ChiR3 has a fix that doesn't need a `TestMain`

Chi's ChiR3 (and my earlier point 2) is about the package-level `init()` in
`services/p2p/metrics.go:56-58` making every non-p2p pod export
`teranode_p2p_connected_peers 0`. It's still the only `init()` among the twelve
`services/*/metrics.go` files — I re-checked all of them.

The proposed remedy was a `TestMain`. There's a simpler one already in this
package: `publish_gate.go:115` and `:157` call `initPrometheusMetrics()` at
the recording site, immediately before touching the collector. Doing the same
at the four new recording sites — `HandleWebsocket.go` `add`/`remove`,
`Server.go:2281`, and the three in `handle_catchup_metrics.go` — makes the
`init()` unnecessary without touching the test setup at all, and it's the
pattern this file already established. `sync.Once` makes the repeated call
almost free.

### N3 — the free fix for ChiR1 checks out

Chi suggested sourcing the connected-peers gauge from the cached node status
rather than paying a registry hop every 5s. I verified it exists:
`Server.go:169` declares `latestNodeStatus atomic.Pointer[notificationMsg]`,
`:1713` stores it, and `ConnectedPeersCount` on it is computed at `:1611-1626`
by counting `p.IsConnected` over `peerRegistry.ListPeers`. So
`if st := s.latestNodeStatus.Load(); st != nil { prometheusP2PConnectedPeers.Set(float64(st.ConnectedPeersCount)) }`
is both free and agrees with the dashboard by construction. Worth taking.

## Still open, unchanged since `de02c8b2`

None of the p2p side moved between `de02c8b2` and `83e73a2a` — I diffed it.
So Chi's list stands as written:

| ID | What | My read |
|---|---|---|
| ChiR1 | Connected-peers gauge counts `len(GetPeers())` unfiltered, while `node_status` counts only `IsConnected` | Agree. Two numbers for the same thing on the same dashboard. See N3 for the cheapest fix. |
| ChiR2 | Pruner rename blanks the live Grafana dashboard | Process, not code — see below. |
| ChiR3 | `init()` exports p2p metrics from every pod | Agree. See N2 for a lighter fix. |
| ChiR4 | Documented pruner label values wrong | **Closed** at `83e73a2a`. |
| ChiR5 | Phantom `pruner_processed_total` | Confirmed still present in three places: `pruner_dah_pruning.puml:62`, `pruner.md:245`, `pruner_settings.md:917`. The settings-doc line tells operators to `grep pruner_processed_total` to check the pruner is alive; it returns empty on a healthy node. The `.puml` line is now half-converted, and the committed SVG bakes that in. |
| ChiR6 | Six new p2p metrics undocumented | Confirmed. `docs/references/prometheusMetrics.md:199-203` has a P2P section with exactly one row (`publish_blocked_total`). None of the six new series are listed. Six-row table edit. |
| ChiR7 | Ban counter misses operator/gossip bans, takes an unbounded `reason` | Agree, non-blocking. |
| ChiR8 | Three counters asserted only by poking the collector | Agree, non-blocking. |

## On ChiR2 — the dashboard sequencing

This is the only item that can't be fixed inside this repo. The rename is a
hard break for `teranode-argocd-deployments` →
`appsets/grafana/application/base/teranode-pruner-dashboard-configmap.yaml`,
which is live and whose 13 `services/pruner` queries all use the bare names.

I've raised that side: **deployments PR 217**
(`fix/pruner-dashboard-teranode-prefix`). It prefixes exactly those 13 and
leaves the 26 `utxo_pruner_*` queries alone — those come from
`stores/utxo/aerospike/pruner`, are a different pruner, and a blanket
find-replace would turn them into `utxo_teranode_pruner_*` and break panels
that work today.

Merge order cuts both ways: landing 217 first blanks the dashboard against
today's code, landing it after blanks it against tomorrow's. They need to go
together, same day. If that coordination is awkward, PR 217's body describes a
dual-name form (`sum(rate(A)) or sum(rate(B))`) that removes the constraint
entirely.

Separately, the PR description still carries no breaking-change note. There's
no CHANGELOG for this to land in, so the description is the only place anyone
downstream will find out.

## Where I land

The pruner half is right and now internally consistent. The blocking item I
personally raised is closed. What's left that I'd genuinely want before merge
is ChiR1 — a gauge named `connected_peers` that disagrees with the
`connected_peers_count` on the same dashboard is the kind of thing that costs
someone an afternoon during an incident — and the dashboard sequencing, which
is a scheduling problem rather than a code one. ChiR3 I'd take too, now that
there's a two-line fix for it.

Everything else (ChiR5–ChiR8, N1) is worth doing but I wouldn't hold the PR
for it.

freemans13:
Adding the clickable link for the dashboard half of this, since I named it in
my review above but plain text across repos doesn't autolink:

**https://github.com/bitcoin-sv/teranode-argocd-deployments/pull/217** —
`fix(grafana): namespace the pruner dashboard queries to teranode_pruner_*`

That moves the thirteen `services/pruner` queries in
`appsets/grafana/application/base/teranode-pruner-dashboard-configmap.yaml`
onto the prefixed names. It deliberately leaves the other twenty-six alone —
those read the Aerospike-side pruner in `stores/utxo/aerospike/pruner`, whose
metrics are already named `utxo_pruner_*` in the Go source and are not part of
this rename. A blanket `pruner_` replace would have turned them into
`utxo_teranode_pruner_*` and broken panels that work today, so each of the
thirteen was matched on its full metric name.

**Merge order, which is the bit that needs a decision rather than a review:**
the two have to land together, same day. It isn't "dashboard last" — merging
217 early blanks the dashboard against today's code, merging it late blanks it
against tomorrow's. Either window is the same silent failure the namespace fix
exists to prevent, just pointed the other way.

If coordinating that turns out to be awkward, 217's description sets out the
alternative: make each query tolerate both names
(`sum(rate(A[…])) … or sum(rate(B[…])) …`). That removes the ordering
constraint entirely, at the cost of thirteen uglier expressions and a cleanup
PR once this has been deployed everywhere. Happy to switch it over if that's
easier than syncing the merges — just say.

@icellan — worth a line in this PR's description too, something like
"breaking: pruner metric names are now prefixed `teranode_pruner_*`, needs the
matching `teranode-argocd-deployments` change to land alongside". There's no
CHANGELOG for this to appear in, so the description is the only place anyone
downstream will find out.

ctnguyen (approved):
Review at commit [58297a0](https://github.com/icellan/teranode/commit/58297a088733864fd454e4d111847a1bb4c700e9)

All eight previous points hold up under a re-read of the tree — including the two I flagged as easy to get wrong. The `init()` really is gone, and `services/p2p/metricsregistrationtest/no_init_registration_test.go` is a genuine regression check rather than a test that can't fail: four of the seven collectors are plain `Gauge`/`Counter`, so a re-added `init()` would put them in `Gather()` and trip the assertion. The connected-peers gauge now makes the same `ListPeers` call with the same `IsConnected` filter as `getNodeStatusMessage` (`repo/services/p2p/connected_peers_monitor.go:57-71` vs `repo/services/p2p/Server.go:1613-1626`), so the two numbers agree by construction, and the error path leaving the previous value rather than publishing zero is the right call. On ChiR7 the declined half of my suggestion is correct and I traced it: `internal/banlist/ban_list.go:174` builds every event as `BanEvent{Action: "add", IP: ..., Subnet: ...}` with `PeerID` never set, so the `PeerID` branch in `handleBanEvent` is unreachable today, and both `onPeerBanned` and `BanPeer` reach `handleBanEvent` through `banList.Add` — incrementing there would double-count. That is a fix, not a dodge. What's left below is all small, and mostly about the new polling goroutine.

## Status of previous points

ChiR2 is fixed as far as this repo can be: the description now carries a Breaking change section covering both the rename and the label-value change, and the companion deployments PR is raised and linked ([bitcoin-sv/teranode-argocd-deployments#217](https://github.com/bitcoin-sv/teranode-argocd-deployments/pull/217)). The dashboard file itself and the same-day merge ordering live outside this repo, so neither is verifiable from here — that part is a scheduling commitment, not code. ChiR3's new guard does fail if the `init()` comes back, but only for the non-`Vec` collectors; see ChiR10.

| ChiR1 | Connected-peers gauge unfiltered | YES | FIXED |
| ChiR2 | Rename blanks pruner dashboard | YES | FIXED |
| ChiR3 | `init()` exports p2p metrics everywhere | YES | FIXED |
| ChiR4 | Documented pruner label values wrong | YES | FIXED |
| ChiR5 | Phantom `pruner_processed_total` |  | FIXED |
| ChiR6 | New p2p metrics undocumented |  | FIXED |
| ChiR7 | Ban counter gaps, unbounded label |  | FIXED |
| ChiR8 | Counters asserted without call sites |  | FIXED |

## Non-blocking issues

### ChiR9 - the connected-peers poll has no deadline, unlike the publisher it copies

**Problem:** `updateConnectedPeersGauge` passes the caller's context straight into `s.peerRegistry.ListPeers` (`repo/services/p2p/connected_peers_monitor.go:57`), and the only context it ever gets is the long-lived `Start` context (`repo/services/p2p/Server.go:812`). The code it deliberately mirrors does the opposite: `publishNodeStatus` wraps every recompute in `context.WithTimeout(ctx, nodeStatusPublishInterval)` (`repo/services/p2p/Server.go:1370`) with a comment at `:1358-1360` saying exactly why — *"Each publish is bounded to this interval so one wedged blockchain call cannot stall the publisher (and freeze the latestNodeStatus cache) forever."* Same RPC, same peer count, no bound. In a distributed deployment `ListPeers` is a gRPC call to the blockchain service (`repo/services/blockchain/peer_registry_client.go:196-218`), which sets no deadline of its own. Transport keepalive (`repo/util/grpc_helper.go:157-160`) catches a dead connection, but not a live server that is slow or blocked — and while the call is outstanding the monitor's `select` isn't running, so ticks are dropped and the gauge holds its last value with nothing recording that it has stopped moving.

**Why it matters:** A gauge that silently freezes at its last reading is the failure mode `connected_peers_monitor.go:12-20` says the dedicated ticker exists to prevent, and it's the same permissive direction as the original ChiR1: `teranode_p2p_connected_peers < 1` under-fires rather than over-fires.

**Fix:** Bound each poll the way the publisher does, so a slow registry costs one skipped sample instead of an indefinite stall:

```go
case <-ticker.C:
	pollCtx, cancel := context.WithTimeout(ctx, interval)
	s.updateConnectedPeersGauge(pollCtx)
	cancel()
```

Same for the priming call before the loop. The existing `ListPeers`-error branch already does the right thing with the result, so a `DeadlineExceeded` needs no new handling.

### ChiR10 - `CounterVec` registration is untested, in both new tests

**Problem:** Two gaps with one root cause: `Gather()` only surfaces a `*CounterVec` once a label combination has been materialized.

`TestImportingP2PPackageDoesNotRegisterMetrics` (`repo/services/p2p/metricsregistrationtest/no_init_registration_test.go:23-32`) asserts that no gathered family starts with `teranode_p2p_`, so an `init()` registering a `CounterVec` would be invisible to it. It catches today's regression only because `connected_peers`, `websocket_connections`, `catchup_attempts_total` and `catchup_successes_total` are plain collectors that appear on registration; the three `CounterVec`s — `publish_blocked_total`, `ban_events_total`, `catchup_failures_total` — would not. The file's doc comment calls itself "the actual regression check for the bug", which is broader than what it checks.

`TestPrometheusMetricsRegisteredWithTeranodeNamespace` hits the same behavior and works around it only partway. `repo/services/p2p/metrics_test.go:55-56` materializes two of the three vecs, `ban_events_total` and `catchup_failures_total`, but not `publish_blocked_total`; correspondingly the `checked` map at `:61-68` holds six entries and omits `teranode_p2p_publish_blocked_total`, while the doc comment at `:47-48` states the test "confirms all p2p metrics are registered with the default gatherer under the `teranode_p2p_` prefix". Being precise about the exposure: that collector is not untested outright — `TestPublishToNetworkIncrementsBlockedCounter` (`repo/services/p2p/publish_gate_test.go:125-139`) drives both stage labels and asserts the counts. But it does so through the Go variable, so it would pass unchanged if `Namespace`/`Subsystem` were dropped from that collector. The claim here is only that nothing asserts the *name* it registers under, not that it is unexercised.

**Why it matters:** The unasserted prefix is precisely the bug the pruner half of this PR exists to fix — a bare name invisible to anything filtering on `teranode_*` — now uncovered in the p2p set; and the next p2p metric is as likely to be a `CounterVec` as not, so if it were the only one an `init()` registered, the guard would stay green while every non-p2p pod exported it.

**Fix:** Two parts. For the guard, assert by attempted re-registration rather than by gathering — that path is name-based, so it covers `Vec` types:

```go
err = prometheus.DefaultRegisterer.Register(prometheus.NewCounterVec(
	prometheus.CounterOpts{Namespace: "teranode", Subsystem: "p2p", Name: "ban_events_total",
		Help: "Total number of peer ban events, labelled by ban reason"},
	[]string{"reason"}))
require.NoError(t, err, "importing services/p2p must not register teranode_p2p_ban_events_total")
```

Keep the `Gather()` sweep too — it's the cheap catch-all — but say in the comment that it alone doesn't cover `Vec` collectors. For the namespace test, close the seventh slot:

```go
prometheusP2PPublishBlocked.WithLabelValues("namespace_check", "namespace_check", "namespace_check").Add(0)
// ...
"teranode_p2p_publish_blocked_total": false,
```

### ChiR11 - the poll is now an RPC per tick, which changes the cost the interval was chosen for

*Same as [@freemans13's point "most of those polls are discarded" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1564#pullrequestreview-4962287832).*

**Problem:** His note was that `connectedPeersPollInterval = 5 * time.Second` (`repo/services/p2p/connected_peers_monitor.go:10`) is finer-grained than a typical 15s scrape, so most polls are discarded — cheap, but wasted. The ChiR1 fix changes the premise: the poll used to read the local `s.P2PClient.GetPeers()`, and now it is a `ListPeers` gRPC round trip returning every entry in the registry unfiltered and unpaginated (`repo/services/p2p/connected_peers_monitor.go:57`; the request carries no limit — `repo/services/blockchain/peer_registry_client.go:197-202`). p2p already makes that same call every 10s from the node-status publisher (`repo/services/p2p/Server.go:1613`, interval at `:1363`) plus once per best-block notification, so the gauge roughly triples p2p's `ListPeers` rate against the blockchain service to produce a number the publisher has already computed and cached in `latestNodeStatus` (`repo/services/p2p/Server.go:1713`).

**Why it matters:** Marshalling the full peer registry three times per 10s window on every node, to resample a value that changes on the order of minutes, is load with no observability gain — and it grows with registry size, which includes gossiped peers.

**Fix:** Either widen the interval to at least the scrape interval (15s) and drop it to a `var` so tests don't need the parameter, or keep the ticker and read the already-computed count when it's fresh, falling back to the registry only if the publisher hasn't run:

```go
if st := s.latestNodeStatus.Load(); st != nil {
	prometheusP2PConnectedPeers.Set(float64(st.ConnectedPeersCount))
	return
}
```

Your reason for preferring the registry read — no dependency on the publisher having run, and a clean fail-closed on error — is sound, and the fallback above keeps both properties.

### ChiR12 - `operator_ban` also counts alert-system bans

**Problem:** `BanPeer` increments `prometheusP2PBanEvents.WithLabelValues(ReasonOperatorBan)` (`repo/services/p2p/Server.go:2153`), and `ReasonOperatorBan`'s comment describes it as bans "applied directly by the BanPeer RPC" (`repo/services/p2p/ban_reasons.go:14-18`). Two callers reach that RPC: `setban` from the JSON-RPC surface (`repo/services/rpc/handlers.go:2393`) and the alert service, which bans on receipt of a signed alert message (`repo/services/alert/node.go:194`). Both land under `reason="operator_ban"`, and the docs row calls it "`operator_ban` (BanPeer RPC)" (`repo/docs/references/prometheusMetrics.md:205`).

**Why it matters:** `ban_events_total{reason="operator_ban"}` jumping is read as "a human did something"; an alert-system ban is a very different event to page on, and the label hides which one happened.

**Fix:** Either rename the label to describe the mechanism rather than the actor (`rpc_ban` / `direct_ban`), or thread the origin through so the two are separable — e.g. an `origin` argument on `BanPeerRequest`, defaulting to `operator`, with the alert service passing `alert`. If neither is worth it now, at least widen the doc cell to say the value covers both the `setban` RPC and alert-system bans.

### ChiR13 - leftovers from the `init()` removal

**Problem:** Two small residues. `NewServer` now calls `initPrometheusMetrics()` twice — once added by this diff at `repo/services/p2p/Server.go:238` and once already present at `:459`; `sync.Once` makes the second a no-op, so it's only noise, but a reader has to check that. And `TestInitPrometheusMetricsIdempotent`'s doc comment still explains the repeated calls as happening "once via the package init() and again from NewServer" (`repo/services/p2p/metrics_test.go:29-31`), describing the mechanism this commit deleted. Separately, and worth a decision rather than a change: the new recording sites don't self-initialize, whereas `publish_gate.go` calls `initPrometheusMetrics()` at its own recording sites (`:115`, `:157`). That's safe today — every new site is either a `*Server` method or `clientChannelMap.add`/`remove`, and the only `clientChannelMap` is built inside `HandleWebSocket`, reached solely from `setupHTTPServer` at `repo/services/p2p/Server.go:722` — so `NewServer` has always run first. It just means the two halves of the same file disagree about whether that's something to rely on.

**Why it matters:** No runtime effect; a comment that describes a deleted mechanism is the kind of thing that sends the next reader looking for an `init()` that isn't there.

**Fix:** Drop the call at `:238`, reword the comment to name the two real entry points (`NewServer` and the package `TestMain`), and either add `initPrometheusMetrics()` to the four new recording sites or note in `metrics.go` that recording sites deliberately assume `NewServer`.

## Recap

| ChiR9 | Poll RPC has no deadline |  | 45% |
| ChiR10 | CounterVec registration untested |  | 30% |
| ChiR11 | 5s poll now a gRPC hop |  | 25% |
| ChiR12 | `operator_ban` covers alert bans |  | 15% |
| ChiR13 | Init-removal leftovers |  | 10% |

icellan:
Rebased onto current main (04f382b8c). Conflicts, all resolved as unions (kept both sides' metrics, per the earlier metrics.go add/add resolution pattern):
- services/p2p/metrics.go: union of this PR's websocket_notifications_dropped/clients_evicted metrics (already on main from another merge) with this PR's own connected_peers/ban_events/catchup_*/websocket_connections metrics. A rebase auto-merge artifact after the union also silently dropped the closing brace of notificationDropped(); restored it.
- services/p2p/HandleWebsocket.go: main had since added a cancel-func eviction mechanism (map[chan []byte]context.CancelFunc) for slow websocket clients; kept that type and layered this PR's websocket_connections gauge update on top.
- services/p2p/ban_enforcement_test.go: two disjoint sets of new tests added at the same location on both sides (relay-ban-IP tests on main, ban-events-counter tests here) — concatenated both, no overlap.
- docs/references/prometheusMetrics.md: unioned both sides' new metric rows in the P2P table.
- services/p2p/metrics_test.go: fixed a stale `cm.add(ch)` call (this PR's old single-arg clientChannelMap.add signature) to match the cancel-func signature now on main.

Verified with `go build ./...`, `go vet`, and `go test -race` across services/p2p and services/pruner — all green. Confirmed the connected-peers IsConnected-filtered gauge, the removed package-level metrics init(), and the corrected pruner label values (preserve_parents/expire_preservations/dah_pruner) are all still present.

icellan:
Rebase onto main lost two edits, which broke the build across every CI job. Fixed in `3299c9182`:

- `services/p2p/metrics.go` — the closing brace of `notificationDropped` was dropped by git's auto-resolution of the add/add conflict on this file, with no conflict marker. The file ended mid-function at line 146. Confirmed against main that the function body is otherwise identical, so only the brace was lost.
- `services/p2p/metrics_test.go:125,128` — two `cm.add(ch)` call sites still on the old signature. Main grew a `context.CancelFunc` parameter on `clientChannelMap.add` (`HandleWebsocket.go:97`); both updated to pass `nil`.

Verified the substantive changes survived the rebase: no package-level `init()` in `metrics.go`, the connected-peers gauge filters on `IsConnected` via `peerRegistry.ListPeers` (`connected_peers_monitor.go:57,66`), and the corrected pruner label values (`preserve_parents`, `expire_preservations`, `dah_pruner`) are the only ones emitted by `worker.go`.

`go build ./...` clean, `go vet ./services/p2p/... ./services/pruner/...` clean, `go test ./services/p2p/... ./services/pruner/... -race -count=1` green.

icellan:
Approved and green, but holding the merge until the companion dashboard PR lands.

`bitcoin-sv/teranode-argocd-deployments` #217 moves the pruner dashboard onto the `teranode_pruner_*` names. Merging this first blanks 13 of its 39 panels in every namespace until that one goes in — no error, no warning, just empty graphs, which is the failure mode this rename exists to prevent.

#217 is approved and mergeable but currently shows BLOCKED with no checks reported, which looks like branch protection needing a review from someone with rights on that repo.

One thing checked and confirmed harmless: this PR also corrected the pruner label *values* (`skipped_total{reason}` and `errors_total{operation}`, including phase 1 moving from `parent_preservation` to `preserve_parents`). The dashboard filters only on `namespace` — zero `reason=` or `operation=` filters across all 39 queries — so the value corrections don't affect it. #217's namespace-only scope is complete.

Will merge as soon as #217 is in.

freemans13 (approved):
Approving on the current head, `3299c9182`. Worth saying explicitly why a fresh approval is useful here: the two existing approvals sit on pre-rebase commits, ordishs on `69c9b007` and ctnguyen on `58297a088`, so GitHub's green "Approved" is carried forward rather than earned on the code that would actually merge. The 31 August rebase onto three weeks of main already lost one edit silently, so a stale approval felt worth converting into a checked one.

**The rebase preserved the change exactly.** I reconstructed both diffs against their own merge bases, `25a93f8c` for the pre-rebase head and `04f382b8c` for the current one, and compared the added and removed lines. The two are identical except for the two call sites the fixup commit already covers:

```
< +     cm.add(ch1)          > +     cm.add(ch1, nil)
< +     cm.add(ch2)          > +     cm.add(ch2, nil)
```

Same 19 files, same line counts. Nothing else moved, so ctnguyen's approval still describes what is here.

**The `nil` cancel func in that test is safe, and not only because the doc comment says so.** `remove` never reads the stored cancel func, it just deletes the key and re-sets the gauge. The only code that calls it is `evict`, which guards with `exists && cancel != nil` (`HandleWebsocket.go:120`). So a nil entry cannot panic on either path the test drives.

**The one thing the rebase created that nobody had checked: this PR removes the package-level `init()`, and main's three new websocket and publish-gate metrics arrived written against a package that still had one.** That combination is how you get a nil pointer in production and a green test suite, because the package's own `TestMain` registers everything regardless. I traced all four non-test uses of the metrics main brought in:

| Use | Guarded by |
|---|---|
| `publish_gate.go:132` | `initPrometheusMetrics()` at line 130 |
| `publish_gate.go:177` | function entry, line 172 |
| `publish_gate.go:210` | same function entry, line 172 |
| `HandleWebsocket.go:163` | `initPrometheusMetrics()` at line 162 |

All four are covered, line 210 by the entry call in `publishToNetwork` rather than an adjacent one. `notificationDropped` calls init itself. No unguarded path, so the `init()` removal is safe against the code main grew in the meantime.

Also confirmed the union merge of the two independently written `metrics.go` files did not leave a metric declared but unregistered, which is the quiet failure mode there: all nine vars in the `var` block have a matching `promauto` call in `_initPrometheusMetrics`.

Local run at this head: `go build ./...` clean, `go vet` clean on both services, and `go test ./services/p2p/... ./services/pruner/... -race -count=1` green, including `metricsregistrationtest`, which is the regression check for the `init()` removal. CI is 36 green and 1 skipped, the skip being the aerospike native-ops job that always skips on a fork image.

**On the merge hold.** Agreed, and the diagnosis on the companion dashboard PR is that its commits are unsigned. `bitcoin-sv/teranode-argocd-deployments` has a repository ruleset with `required_signatures` on `main`, and both commits on that PR report `verified: false, reason: unsigned`. That fits what you saw: two approvals, mergeable, zero checks, still BLOCKED. Unsigned commits have landed on that main recently, but from an org admin who can bypass the ruleset, which I cannot. Re-signing means rewriting both commits and force-pushing, so I will confirm before touching it.

PR #1567, opened 2026-08-10
fix(asset): correct GetBlock doc examples, register by-height block routes

icellan:
The doc comments on the block-serving handlers in `services/asset/httpimpl/GetBlock.go` showed example URLs that don't match the server's real registered routes — a nonexistent `/raw` suffix, wrong content-type claims, and a never-registered `/block/height/...` path. Rewrote them to match the real routes (bare path = binary, `/hex`, `/json`).

Also: `GetBlockByHeight` was a fully implemented, tested handler that `http.go` never registered as a route — unreachable via HTTP, only exercised directly by its test. Registered it following the existing by-hash triplet pattern (binary/hex/json) rather than deleting working, tested code.

### DIFF
--- a/services/asset/httpimpl/GetBlock.go
+++ b/services/asset/httpimpl/GetBlock.go
@@ -131,14 +131,14 @@ type BlockExtended struct {
 //
 // Example Usage:
 //
-//	// Get block in JSON format
-//	GET /block/height/0
-//
 //	// Get block in raw binary format
-//	GET /block/height/0/raw
+//	GET /api/v1/block/height/0
 //
 //	// Get block in hex format
-//	GET /block/height/0/hex
+//	GET /api/v1/block/height/0/hex
+//
+//	// Get block in JSON format
+//	GET /api/v1/block/height/0/json
 func (h *HTTP) GetBlockByHeight(mode ReadMode) func(c echo.Context) error {
 	return func(c echo.Context) error {
 		ctx, _, deferFn := tracing.Tracer("asset").Start(c.Request().Context(), "GetBlockByHeight_http",
@@ -280,14 +280,14 @@ func (h *HTTP) GetBlockByHeight(mode ReadMode) func(c echo.Context) error {
 //
 // Example Usage:
 //
-//	// Get block in JSON format
-//	GET /block/hash/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
-//
 //	// Get block in raw binary format
-//	GET /block/hash/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f/raw
+//	GET /api/v1/block/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
 //
 //	// Get block in hex format
-//	GET /block/hash/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f/hex
+//	GET /api/v1/block/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f/hex
+//
+//	// Get block in JSON format
+//	GET /api/v1/block/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f/json
 func (h *HTTP) GetBlockByHash(mode ReadMode) func(c echo.Context) error {
 	return func(c echo.Context) error {
 		return h.getBlockByHash(c, mode)
--- a/services/asset/httpimpl/http.go
+++ b/services/asset/httpimpl/http.go
@@ -381,6 +381,10 @@ func New(logger ulogger.Logger, tSettings *settings.Settings, repo *repository.R
 	apiGroup.GET("/block/:hash/forks", h.GetBlockForks)
 	apiGroup.GET("/block/:hash/nearestforks", h.GetNearestForkHeights)
 
+	apiGroup.GET("/block/height/:height", h.GetBlockByHeight(BINARY_STREAM), heavyMW()...)
+	apiGroup.GET("/block/height/:height/hex", h.GetBlockByHeight(HEX), heavyMW()...)
+	apiGroup.GET("/block/height/:height/json", h.GetBlockByHeight(JSON), heavyMW()...)
+
 	apiGroup.GET("/block/:hash/subtrees/json", h.GetBlockSubtrees(JSON))
 
 	apiGroup.GET("/search", h.Search)
--- a/services/asset/httpimpl/http_test.go
+++ b/services/asset/httpimpl/http_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/bsv-blockchain/teranode/errors"
 	"github.com/bsv-blockchain/teranode/services/asset/repository"
 	"github.com/bsv-blockchain/teranode/services/blockchain"
 	"github.com/bsv-blockchain/teranode/settings"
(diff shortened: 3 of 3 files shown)


ordishs (approved):
Verified. In a clean worktree, reverting only the `http.go` route registration makes the new test fail with 404 on all three by-height paths, which proves the routes were missing. With the full diff applied the whole `services/asset/httpimpl` package passes, including the new `TestGetBlockByHeightRouteRegistered` subtests. The corrected doc comments now show `/api/v1` paths, which match the default `asset_apiPrefix` in `settings.conf` and `settings.go`. No route shadowing exists between `/block/height/:height` and `/block/:hash`, because Echo matches by segment count.

No findings.

freemans13 (commented):
Reviewed this one independently — it's correct and well-scoped, and I'd ship it as-is. Four small follow-ups below, none of them blocking.

**What I verified rather than assumed.** The risk I wanted to rule out was route shadowing: adding a static `height` segment as a sibling of the existing `/block/:hash` param route can, in some routers, stop the param route from matching. I built a throwaway Echo router with the real registration set and confirmed that on echo v4.13.3 the backtracking works — `/block/<hash>`, `/block/<hash>/hex`, `/json`, `/forks` and `/subtrees/json` all still resolve to the by-hash handlers, the three new by-height paths resolve to the new ones, and a bare `/block/height` falls through to by-hash as a (nonsense) hash rather than 404-ing oddly. I also checked the rewritten doc claims against the code, and they now hold: binary really is `application/octet-stream` via `c.Blob`, hex really is `text/plain` via `c.String`, json really is `application/json` via `c.JSONPretty`. And I swept the package for other orphaned handlers — `GetBlockByHeight` was the only one; everything else that isn't registered is a lifecycle or helper method (`Init`, `Stop`, `Sign`, `AddHTTPHandler`), so the audit looks complete.

Package tests pass, `go vet` is clean, and `golangci-lint` on `services/asset/httpimpl` reports only a pre-existing `prealloc` warning in `fsm_handler.go`, which this PR doesn't touch.

### Follow-ups

1. **`docs/references/services/asset_reference.md`** (Block Endpoints, around line 579) lists the by-hash triplet but not the new by-height routes. Given this PR is specifically about docs telling the truth about routes, leaving the reference doc stale slightly undercuts its own point — three short entries would close it.

2. **`docs/topics/services/assetServer.md:177`** has the same gap: it documents `/api/v1/block/:hash` with its hex and json variants, with no mention of by-height. One line alongside the existing entry.

3. **`services/asset/httpimpl/http.go:75-81`** — the `New()` godoc's own "Block Related" endpoint list doesn't mention the by-height route either. That list is already partial (it omits `/blocks/:hash` and others), so this is the most optional of the four, but it's in the file you're already editing.

4. **Error-path test coverage.** The new router test is a good addition and covers exactly the failure mode that motivated the PR. What's still untested is the pair of error paths that registration makes publicly reachable for the first time: a non-numeric `:height` hitting `strconv.ParseUint` should give 400, and an unknown height should give 404. `TestGetBlockByHeight` in `GetBlock_test.go` only exercises the JSON/Binary/Hex success cases. Two small subtests would be cheap insurance now that arbitrary user input reaches the parser over HTTP.

One thing worth a conscious nod rather than a change: exposing lookup by height makes sequential bulk scraping trivially scriptable in a way that lookup by hash isn't. I don't think that's a reason to hold the change — the new routes sit behind the same `heavyMW()` rate limiter as the by-hash ones, and `/api/v1/blocks/:hash` already allows walking the chain — but it's a real difference in how easy the data is to enumerate, so it's better acknowledged than discovered later.

ctnguyen (approved):
Review at commit [ad1450e](https://github.com/icellan/teranode/commit/ad1450e3d06aad51a81bd10e5527ca29e22ed17c)

Nice catch on the orphaned handler — `GetBlockByHeight` being fully implemented, tested, and simply never wired to the router is exactly the kind of thing that hides for years, and choosing to register it rather than delete it was the right call. The registration follows the existing by-hash triplet faithfully, including `heavyMW()`, and the new `TestGetBlockByHeightRouteRegistered` tests the one thing handler-level tests structurally cannot: that the route exists. The doc rewrite is a real improvement too — the old examples pointed at `/block/hash/<hash>/raw`, which never existed. One thing worth flagging so nobody duplicates work: the error paths raised in @freemans13's point 4 are in fact already covered at handler level in `GetBlock_test.go` — `Invalid height` asserts 400 (`GetBlock_test.go:121-139`), `Repository not found` asserts 404 (`GetBlock_test.go:164-185`), and `Repository error` asserts 500 (`GetBlock_test.go:141-162`). Only two follow-ups from me, both documentation, neither blocking.

## Non-blocking issues

### ChiR1 — New public routes are absent from the OpenAPI source and the committed spec

*Same as [@freemans13's points 1, 2 and 3 (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1567#pullrequestreview-4918796257).*

**Problem:** Beyond the two markdown files @freemans13 listed, the gap also hits a generated artifact that is checked in. `services/asset/httpimpl/swagger_routes.go:92-112` carries the `swagger:route` annotations for the by-hash triplet; there is no by-height entry, and `services/asset/httpimpl/swagger.json` is committed and regenerated from those annotations via `make swagger-asset` / `make swagger-validate` (Makefile:899-909). So the published description of the service is now missing three live endpoints. The same omission exists in `New()`'s own godoc endpoint list (`http.go:74-80`), in `docs/references/services/asset_reference.md:579-594`, and in `docs/topics/services/assetServer.md:177`.

**Why it matters:** The stated premise of this PR is that documentation should not lie about the routes. Shipping a route family the machine-readable spec doesn't know about reintroduces the same class of drift one layer down, and this one bites harder than a stale markdown line because clients generate stubs from `swagger.json`. Nothing here affects runtime behaviour — the routes work; they are just invisible to anyone reading the spec.

**Fix:** Add the three annotations alongside the by-hash block and regenerate the spec, then top up the prose lists.

```go
// swagger:route GET /api/v1/block/height/{height} block getBlockByHeight
// Get a block by height in binary format.
// responses:
//   200: binaryResponse
//   400: errorResponse
//   404: errorResponse
```

### ChiR2 — The edited doc comments still carry inaccurate claims

**Problem:** The PR fixed the `Example Usage` blocks but left untrue statements a few lines above them, in the same comments it was auditing. Three items, in both handlers:

- `GetBlock.go:71-72` documents a `format` query parameter that "overrides the default response format". Reading `GetBlockByHeight` end to end (`GetBlock.go:142-212`) and `getBlockByHash` (`GetBlock.go:298-370`), neither handler reads any query parameter at all — the response format comes solely from the `mode` argument fixed at registration time (`http.go:378-386`). There is no code path by which a query string changes the output.
- `GetBlock.go:128` says execution time is recorded in the `GetBlock_http` statistic, but the span is opened as `GetBlockByHeight_http` (`GetBlock.go:144`).
- `GetBlock.go:129` **and** `GetBlock.go:278` both name the Prometheus metric `asset_http_get_block`. `metrics.go:192-198` registers it with `Namespace: "teranode"`, `Subsystem: "asset"`, `Name: "http_get_block"`, which Prometheus composes as `teranode_asset_http_get_block`. Line 278 is in the `GetBlockByHash` comment, whose `Example Usage` section this PR also rewrote — so it's the same block, same audit, same miss.

**Why it matters:** These are documentation-only defects; the metric is emitted correctly and the handlers behave correctly, so nothing is broken at runtime. The cost is misdirection. The phantom `format` parameter is the worst of the three because it is actively misleading rather than merely stale — a caller can reasonably write `?format=json` against the binary route, get binary back with a 200, and have no signal that the parameter was ignored. The two wrong identifiers just send operators searching for a stat and a metric that don't exist under those names.

**Fix:** Drop the `Query Parameters` section (or state that the format is fixed by the route), and correct both identifiers in both comment blocks.

```go
// Query Parameters:
//   - none; the response format is determined by the route suffix (bare/hex/json)
...
// Monitoring:
//   - Execution time recorded in the "GetBlockByHeight_http" span
//   - Prometheus metric "teranode_asset_http_get_block" tracks successful responses
```

## Recap

| ChiR1 | Routes missing from OpenAPI spec | | 35% |
| ChiR2 | Inaccurate claims in edited comments | | 25% |

PR #1568, opened 2026-08-10
fix(docs,settings): correct stale k8s doc headers, implement TestNoMissingTags

icellan:
Two independent fixes:

- The Kubernetes operator/miner runbook docs' "Last modified" headers still showed a March 2025 date despite edits through November — corrected against actual git history.
- `TestNoMissingTags` had no real body and didn't actually check anything. Implemented a reflection walk over the settings struct that fails on any leaf field missing its documentation tag, with an explicit exemption for genuinely runtime-computed fields. Verified it catches a missing tag before restoring it.


ordishs (approved):
Verified both halves. Git log dates on all nine "Last modified" header files match the new dates exactly, including the one file left untouched. I applied the diff in a worktree and confirmed the pre-change `TestNoMissingTags` was a no-op log statement. The new test passes; stripping a real key tag from `ServiceName` makes it fail with the exact field name.

Two nits:

- `docs/howto/miners/kubernetes/minersSecurityBestPractices.md` — the diff bundles unrelated blank-line cleanup into a commit whose stated purpose is a stale date header.
- `settings/coverage_test.go:14-20` — the exemption list is a hardcoded map of dotted field paths in the test file. A sentinel tag value on the field declaration itself would not drift.

freemans13 (commented):
Reviewed both halves locally. **Looks good to me** — no blockers, no majors. Two things I'd suggest before merge, plus some nits you can take or leave.

## What I checked

The important property of this PR is that the old `TestNoMissingTags` asserted nothing at all — it just logged "Migration complete: All 398 configurable fields have struct tags" — so the whole value rests on the replacement being *able* to fail. It is. Two falsification experiments, both reverted afterwards:

- removed `key:"clientName"` from `settings/interface.go:32` → `FAIL: Should be empty, but was [ClientName]`
- removed `key:"KAFKA_HOSTS"` from `settings/kafka_settings.go:10` (nested, to prove recursion) → `FAIL: Should be empty, but was [Kafka.Hosts]`

The dotted prefix works and the message names the offending field, so it's actionable.

I also instrumented the walk with a throwaway test: it visits **615 fields** (610 tagged, 5 exempt, 0 missing), zero unexported fields anywhere in the tree, and exactly one foreign struct skipped (`ChainCfgParams → chaincfg.Params`). No slice, map, array or interface field anywhere holds a settings struct, so the "container of structs" false-negative class doesn't exist today, and no type is visited twice so there's no cycle.

**No runtime behaviour change**, which I wanted to confirm given the PR is framed as docs+test. The five newly-exempted fields previously had no `key` tag at all, so they hit the `key == ""` branch; being `string`/`bool` rather than struct or pointer-to-struct, they fell through to `continue` and were never exported. The new `key == keyTagExempt → continue` at `settings/export.go:131-133` produces the identical outcome. `git grep 'key:"-"' main -- settings/` returns nothing, so no pre-existing field's behaviour flips. `extractFields` still yields 610 entries and `TestMetadataCoverage` still passes.

All eight doc dates verified against `git log -1` on `main` — every one matches exactly.

Green locally: `go test ./settings/...`, `go test -race ./settings/...`, `go vet ./settings/...`, `golangci-lint run ./settings/...` (0 issues), `gofmt -l settings/`.

Carrying the exemption on the field declaration rather than in a skip-list is the right call — it survives renames and moves, and mirrors the `encoding/json` `-` convention so it reads naturally.

## Two suggestions

**1. The test can still pass vacuously.** `settings/coverage_test.go:107-109` asserts only `require.Empty(t, missing, …)`. A nil `missing` passes, so a walk that visits *zero* fields reports success. That isn't hypothetical, because of the foreign-package skip at line 95: if `KafkaSettings` (or any sub-struct) were moved into a sibling package, all ~30 of its fields would drop out of the check with no failure, since `fieldType.PkgPath() != settingsPkgPath` short-circuits the whole subtree. A settings-hygiene test that can pass while checking nothing is the exact failure mode this test exists to prevent. Suggest counting visited leaves and asserting a floor:

```go
require.GreaterOrEqual(t, visited, 600, "walk visited too few fields — did a subtree move out of the settings package?")
```

**2. `BlocksValidate` looks dead, not internal.** `settings/kafka_settings.go:9` exempts it as "Internal field, not directly configured", but a repo-wide grep shows it's never written and never read: `settings.go:131-165` builds the `KafkaSettings` literal without it, and no consumer references `Kafka.BlocksValidate` (the other `BlocksValidate` hits are `BlocksValidated`, an unrelated catchup counter). Deleting the field satisfies the test just as well and removes an unused exported field from a public settings struct.

## Nits

- **Tagged struct fields stop recursion** (`settings/coverage_test.go:82-84`). Any non-empty `key` tag skips the field, including struct-typed ones. Two exist today — `BlockChain.PostgresPool` (`blockchain_settings.go:20`) and `UtxoStore.PostgresPool` (`utxostore_settings.go:52`), both `*PostgresSettings` with real key tags — so their children are never inspected under those paths. Harmless only because `PostgresSettings` is also reached via the untagged top-level `Settings.Postgres`; remove or tag that one field and the whole subtree goes unchecked, silently. Same shape applies to `key:"-"` on a struct: one tag hides N descendants. Either recurse into settings-package structs regardless of tag, or assert that no struct-typed field carries a `key` tag other than `-` and document the whole-subtree exemption as intentional.
- **Stale sibling comment** (`settings/coverage_test.go:19`): `// We should have 398 settings based on the complete migration` — the assertion below is `>= 300` and the real count is 610. You deliberately removed the matching "398 configurable fields" log line from `TestNoMissingTags`; this one survived in the same file.
- **`minersSecurityBestPractices.md`** is the one doc where real content changes (9 blank lines removed), yet its header is set to `5-July-2025`, the file's pre-PR date — stale the moment it merges. Either bump it to the PR date or drop the whitespace cleanup so it's header-only like the other seven. Purely cosmetic otherwise; consecutive blank lines render identically in MkDocs.
- **`TestKeyTagExemptNotExported`** (`settings/coverage_test.go:129-134`) collects exempt names from the top level of `Settings` only, so the nested `Kafka.BlocksValidate` exemption isn't covered by the assertion at line 140. That line also matches on bare field name, so a nested tagged setting named `Version`/`Context`/`Commit` would fail spuriously — no collision today, but matching on the full dotted path would be safer.
- **Scope**, entirely optional: only the 8 `kubernetes/` docs were fixed. 11 others carry the same header and are still stale — worst is `minersHowToBackup.md` (header `6-March-2025`, git `18-February-2026`) and `minersHowToCPUMiner.md` (header `02-Aug-2025`, git `2-April-2026`, plus a different date format from every other file). Nothing in the repo enforces these headers, so all 19 will drift again; a CI check or pre-commit hook is the only thing that makes them trustworthy. Out of scope here, just flagging.

## Unrelated finding, worth a follow-up

Not caused by this PR, but squarely in the area the new test polices: `ExportMetadata` currently emits **duplicate keys**.

- `blockmaxsize` — from both `block_settings.go:15` (`MaxSize`) and `policy_settings.go:5` (`BlockMaxSize`)
- `blockvalidation_processTxMetaUsingStore_BatchSize` — from both `block_settings.go:21` and `blockvalidation_settings.go:24`

Two fields silently fighting over one config key is a worse bug than a missing tag, and the new walk already visits every field, so asserting key uniqueness alongside tag presence would be nearly free.

## Testing limits

Settings-package tests, race detector, vet and lint were run locally and passed; the heavier CI stages (benchmarks, sequential, smoke) were not — impractical locally, and this change is confined to a test file plus struct tags, so the risk there is low but not zero. The duplicate-key finding above was established by reading and grep, not by running `ExportMetadata` and diffing its output, so treat those specific pairs as high-confidence rather than proven.

ctnguyen (changes requested):
Review at commit [f420837](https://github.com/icellan/teranode/commit/f4208372e8efe3e11c083f94b163af7327fda389)

Turning `TestNoMissingTags` from a `t.Log` placeholder into something that can actually fail is a genuinely useful bit of housekeeping, and the design choice is the right one: carrying the exemption as `key:"-"` on the declaration mirrors the `encoding/json` convention, survives renames and moves, and keeps `extractFields` and the test reading from the same source of truth. Adding `TestKeyTagExemptNotExported` to police the other half of that contract is a nice touch — plenty of people would have stopped at the first test. The `extractFields` change is behaviour-neutral (the five newly-exempt fields are `string`/`bool` leaves that already fell through the `key == ""` branch to `continue`), and the docs half is a clean, low-risk diff. My main concern is that the new walk and the exporter it is meant to mirror do not use the same traversal rule, and the gap sits exactly where the struct is most likely to grow next.

## Blocking issues

### ChiR1 — The walk and `extractFields` disagree on foreign structs, leaving every `*url.URL` setting unchecked

*Partially the same as [@freemans13's point 1 (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1568#pullrequestreview-4917160582).*

**Problem:** The two walkers that are supposed to mirror each other apply different rules to an untagged struct field. `extractFields` (`settings/export.go:135-145`) has no package check at all — for `key == ""` it descends into any struct or pointer-to-struct, foreign or not. The test's walk (`settings/coverage_test.go:91-97`) bails out on `fieldType.PkgPath() != settingsPkgPath` and records nothing. `Settings.ChainCfgParams` (`settings/interface.go:62`) is the concrete present-day case: it is untagged and `*chaincfg.Params` is foreign, so `extractFields` walks the whole of `chaincfg.Params` hunting for `key` tags while the test never looks at it. The two happen to agree on the *outcome* today — go-chaincfg v1.6.0 declares no `key` tags, so the recursion emits nothing and there is nothing for the test to check — so this is a divergence in traversal rule, not a live export bug. Where the divergence has teeth is `*url.URL`. There are roughly 28 `*url.URL` settings fields — `TracingCollectorURL` (`settings/interface.go:34`), the ten `Kafka.*Config` fields (`settings/kafka_settings.go:21-31`), `Block.BlockStore`/`TxStore`/`UtxoStore` (`settings/block_settings.go:16,24,25`), `BlockChain.StoreURL` (`settings/blockchain_settings.go:15`), the four Aerospike policy URLs, and so on. All of them are tagged at this head, so nothing is broken right now. But drop a tag from one and both walkers go quiet in different ways: `extractFields` takes the untagged branch, descends into `url.URL`, finds no `key` tags and emits nothing, so the setting silently disappears from `ExportMetadata`; the test takes the `PkgPath` branch and appends nothing to `missing`, so it stays green. The docstring's justification at `settings/coverage_test.go:61-63` — structs from other packages "are not ours to tag" — is true of `chaincfg.Params` and false of these: those fields are ours, and they do carry tags. freemans13 raised the same `PkgPath` short-circuit as a hypothetical about a sub-struct being moved to a sibling package; my point is that the hole already covers the second-most-common setting type in the tree.

**Why it matters:** The whole value of this PR's Go half is that the test can fail when a tag goes missing. A guard that is structurally blind to ~5% of the fields — specifically the store and endpoint URLs, which are the fields most often added when a new backend lands — will read green while the regression it exists to catch ships. And because the exporter and the test disagree about what a foreign struct even is, the test is not verifying the traversal that actually produces the portal data.

**Fix:** Make the walk treat a settings-package struct as the only container worth descending into; everything else is a leaf that must carry a tag. To keep `ChainCfgParams` passing, give it the sentinel this PR already introduces — that is behaviour-neutral for `extractFields` (it contributes no entries either way, since `chaincfg.Params` has no `key` tags) and it brings the two walkers back into agreement, both skipping the field for the same declared reason. One consequence worth handling in the same change: the walk has no `IsExported()` guard, unlike the sibling walker `walkSensitiveTags` at `settings/redact.go:104`, so under the stricter rule an unexported field would start being flagged as missing.

```go
// settings/interface.go
ChainCfgParams *chaincfg.Params `key:"-"` // Runtime-derived; exported separately as the "network" setting

// settings/coverage_test.go — inside walk()
field := typ.Field(i)
if !field.IsExported() {
    continue
}

fieldName := prefix + field.Name

if field.Tag.Get("key") != "" {
    continue
}

fieldType := field.Type
if fieldType.Kind() == reflect.Pointer {
    fieldType = fieldType.Elem()
}

// Only a settings-package struct is a container to walk into. Anything else -
// including a foreign struct such as url.URL - is a leaf setting and must carry
// a tag or the key:"-" exemption.
if fieldType.Kind() == reflect.Struct && fieldType.PkgPath() == settingsPkgPath {
    walk(fieldType, fieldName+".")
    continue
}

missing = append(missing, fieldName)
```

Criticality: **70%**

## Non-blocking issues

### ChiR2 — A `key` tag on a struct-typed field hides its whole subtree, in both the walk and `extractFields`

*Same as [@freemans13's point "Tagged struct fields stop recursion" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1568#pullrequestreview-4917160582).*

**Problem:** `settings/coverage_test.go:82-84` treats *any* non-empty `key` tag as "nothing more to check here", including on struct-typed fields. Two exist today: `BlockChain.PostgresPool` (`settings/blockchain_settings.go:20`) and `UtxoStore.PostgresPool` (`settings/utxostore_settings.go:52`), both `*PostgresSettings` with real key tags. Their 13 children are never inspected under those paths; coverage survives only because `PostgresSettings` is also reachable through the untagged top-level `Settings.Postgres` (`settings/interface.go:54`). Remove or tag that one field and 13 settings go unchecked silently. The sentinel this PR introduces sharpens the same edge: a single `key:"-"` on a struct-typed field would exempt an arbitrarily large subtree from both the test and `ExportMetadata` (`settings/export.go:131-133`) in one stroke, with nothing recording that it happened. Extending freemans13's point slightly: the mirrored behaviour in `extractFields` is the more consequential half, since it is what determines what the portal actually shows.

**Why it matters:** The test is meant to be the tripwire for silent coverage loss, and this is a way to lose coverage silently. The blast radius per tag is unbounded, which is a bad property for a one-character annotation.

**Fix:** Either recurse into settings-package structs regardless of tag, or make the whole-subtree exemption explicit and assert that nobody adds one by accident.

```go
if k := field.Tag.Get("key"); k != "" {
    if fieldType.Kind() == reflect.Struct && fieldType.PkgPath() == settingsPkgPath {
        require.Equal(t, keyTagExempt, k,
            "struct field %s carries key %q - its %d children are then never checked; "+
                "use key:\"-\" if the whole subtree is deliberately exempt",
            fieldName, k, fieldType.NumField())
    }
    continue
}
```

Criticality: **45%**

### ChiR3 — `TestKeyTagExemptNotExported` does not cover the nested exemption this PR adds

*Same as [@freemans13's point "`TestKeyTagExemptNotExported`" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1568#pullrequestreview-4917160582).*

**Problem:** The exempt-name set at `settings/coverage_test.go:129-134` is built from a flat loop over the top level of `Settings`, so it contains only `Commit`, `Version`, `Context` and `IsAllInOneMode`. `Kafka.BlocksValidate` (`settings/kafka_settings.go:9`) — one of the five exemptions this PR introduces, and the only nested one — is absent, so the assertion at line 140 would not notice if `extractFields` regressed and started exporting it. Separately, line 140 matches on the bare `setting.Name`, which is the unqualified field name; a future nested setting called `Version` or `Context` would fail this test spuriously even though it is a perfectly legitimate setting. No such collision exists in the tree today, so that half is latent rather than broken.

**Why it matters:** This test exists to prove the sentinel is honoured end-to-end. Covering four of the five exempt fields, and missing the one that exercises the recursive path, leaves the interesting case untested.

**Fix:** Collect exempt fields with the same recursive walk `TestNoMissingTags` uses, keyed on the dotted path, and compare against a dotted path built from `metadataEntry.FieldName` (or expose the path alongside it) rather than the bare display name.

```go
exempt := map[string]bool{}

var collect func(typ reflect.Type, prefix string)
collect = func(typ reflect.Type, prefix string) {
    for i := 0; i < typ.NumField(); i++ {
        f := typ.Field(i)
        switch {
        case f.Tag.Get("key") == keyTagExempt:
            exempt[prefix+f.Name] = true
        case f.Tag.Get("key") == "":
            ft := f.Type
            if ft.Kind() == reflect.Pointer {
                ft = ft.Elem()
            }
            if ft.Kind() == reflect.Struct && ft.PkgPath() == settingsPkgPath {
                collect(ft, prefix+f.Name+".")
            }
        }
    }
}
collect(reflect.TypeOf(Settings{}), "")

require.Len(t, exempt, 5, `unexpected number of key:"-" fields`)
```

Criticality: **35%**

### ChiR4 — `Kafka.BlocksValidate` is dead, not "internal"

*Same as [@freemans13's point 2 (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1568#pullrequestreview-4917160582).*

**Problem:** `settings/kafka_settings.go:9` is now tagged `key:"-"` with the comment "Internal field, not directly configured". A repo-wide search for `BlocksValidate` returns exactly one hit — that declaration. It is never assigned (the `KafkaSettings` literal in `settings/settings.go` does not set it) and never read; every other `BlocksValidate*` match in the tree is `BlocksValidated`, the unrelated catchup counter in `services/blockvalidation/`. No config key, conf entry or env var maps to it either.

**Why it matters:** The exemption makes the test pass, but it also promotes an incorrect comment into an intentional, documented contract on a field of a public struct. The next reader will assume something somewhere populates it.

**Fix:** Delete the field. That satisfies `TestNoMissingTags` just as well and removes an unused exported member from the public `KafkaSettings` API. If there is a near-term consumer planned, keep it but change the comment to say so.

Criticality: **22%**

### ChiR5 — `minersSecurityBestPractices.md` is stale the moment it merges

*Same as [@ordishs's point 1 (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1568#pullrequestreview-4907184848) and [@freemans13's point "`minersSecurityBestPractices.md`" (12-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1568#pullrequestreview-4917160582).*

**Problem:** This is the one file in the docs half where real content changes — nine blank lines removed around lines 6, 52-54 and 56-57 of the pre-PR file — yet the header at `docs/howto/miners/kubernetes/minersSecurityBestPractices.md:3` is set to `5-July-2025`, the file's own pre-PR modification date. The commit that carries this diff will itself be the file's newest change, so the header is wrong again as soon as it lands. The other seven files in the set are header-only edits, where backdating to the last real content change is exactly right.

**Why it matters:** The PR's stated purpose is making these headers trustworthy; this is the single file where the new value is guaranteed to be wrong. It also bundles unrelated whitespace cleanup into a commit whose message is about a date header, which makes the diff harder to review at a glance.

**Fix:** Either drop the whitespace cleanup so the file matches the other seven and `5-July-2025` stays accurate, or keep the cleanup and bump the header to the merge date. Consecutive blank lines render identically in MkDocs, so the cleanup buys nothing on its own.

Criticality: **18%**

## Recap

| ChiR1 | Foreign structs skipped, URLs unchecked | YES | 70% |
| ChiR2 | Tagged struct hides whole subtree | | 45% |
| ChiR3 | Nested exemption untested | | 35% |
| ChiR4 | `BlocksValidate` is dead code | | 22% |
| ChiR5 | Security doc header stale on merge | | 18% |

icellan:
Pushed 5e7983913 addressing all five findings.

**ChiR1** (blocking) — `walk()` and `extractFields` now agree on what counts as a container: a settings-package struct is always descended into; anything else — including `*url.URL` and `*chaincfg.Params` — is a leaf that must carry a tag. `ChainCfgParams` gets `key:"-"`. Added `TestNoMissingTags_CatchesDroppedURLTag`, which builds a throwaway struct with an untagged `*url.URL` field and asserts the walk reports it missing — proves the guard now actually catches this, rather than asserting it against the real `Settings` tree where it would trivially pass either way.

**ChiR2** — recursion into a settings-package struct field now happens regardless of whether that field also carries a real `key` (e.g. `BlockChain.PostgresPool`, `UtxoStore.PostgresPool`); only the `key:"-"` sentinel itself now halts recursion. `extractFields`'s behavior is unchanged (a real key on a struct field still emits one entry and doesn't recurse — that's deliberate, it's a single override setting, not a set of independently-exported children); the test's walk is decoupled from it so it can verify the children's tags independently.

**ChiR3** — `TestKeyTagExemptNotExported` now collects exempt fields via the same recursive walk instead of a flat top-level loop, and compares against the dotted `FieldName` path `extractFields` already tracks internally (extended it to accumulate the full path, purely for this internal check — no change to exported `Key`/`Name`) instead of the bare display name. Removes the collision risk on bare names.

**ChiR4** — grepped the repo: `Kafka.BlocksValidate` is never assigned or written anywhere (the one non-test hit was the field declaration itself; every `BlocksValidate*` match elsewhere is the unrelated `BlocksValidated` catchup counter in `services/blockvalidation`). Deleted the field rather than keeping the `key:"-"` exemption with the "internal field" comment.

**ChiR5** — kept the whitespace cleanup as-is (reverting it fails the repo's `markdownlint` pre-commit hook on `MD012/no-multiple-blanks`, so "drop the cleanup" wasn't actually viable) and bumped the header to today's date instead, since this commit is now itself the file's latest content change.

Verified: `go build ./...`, `go vet ./...`, `go test ./settings/... -race`, `golangci-lint run ./settings/...`, `staticcheck ./settings/...`, `gofmt -l` all clean.

freemans13 (commented):
Re-reviewed at head `5e7983913`. This commit answers every point from the two earlier rounds, including ctnguyen's blocking ChiR1, and the guard is now genuinely load-bearing — I could not find a way to make it pass while a real setting was silently lost, **with one exception**, which is the one thing I'd want fixed before merge. Nothing below is a runtime bug: the Go half is struct tags plus a test, and I confirmed the exported metadata is unchanged apart from the intended edits.

## Where the earlier rounds stand at this head

| Point | Raised by | Status |
|---|---|---|
| ChiR1 — foreign structs skipped, ~28 `*url.URL` settings unguarded | @ctnguyen (blocking) | **Fixed.** Non-settings-package structs are now leaves that must carry a tag, `ChainCfgParams` got `key:"-"`, and `TestNoMissingTags_CatchesDroppedURLTag` pins it so it can't silently regress |
| ChiR2 — a tagged struct hides its whole subtree | @ctnguyen | **Fixed** via the recurse-regardless-of-tag option. This is what opens the new hole below |
| ChiR3 — nested exemption untested, bare-name matching | @ctnguyen, me | **Fixed.** `settingsWalkExempt` recurses, comparison is on the dotted `FieldName` |
| ChiR4 — `Kafka.BlocksValidate` dead, not internal | @ctnguyen, me | **Fixed** — deleted. Re-grepped the repo: zero remaining references, every other `BlocksValidate*` hit is `BlocksValidated`, the unrelated catchup counter |
| ChiR5 — security-doc header stale on merge | @ordishs, @ctnguyen, me | **Fixed** by bumping to today. See the nit below on what that trades away |
| My point 1 — test could pass vacuously via the `PkgPath` escape hatch | me | **Largely moot.** With foreign structs as leaves, a subtree moved to a sibling package now fails loudly as an untagged leaf instead of vanishing |
| Hardcoded skip-list in the test file | @ordishs | **Fixed** — the exemption is a `key:"-"` sentinel on the declaration |

## Should fix before merge — the two struct-level override settings are the last blind spot

Confirmed by mutation, not reasoned. `BlockChain.PostgresPool` (`settings/blockchain_settings.go:20`) and `UtxoStore.PostgresPool` (`settings/utxostore_settings.go:52`) are `*PostgresSettings` fields carrying real key tags — `blockchain_postgres_pool` and `utxostore_postgres_pool` — exported as single aggregate settings. Fixing ChiR2 by descending into settings-package structs *regardless* of the tag means the walk now steps straight past those two tags on its way to the children, so nothing checks they still exist.

I instrumented both traversals and diffed them field by field: of the 610 exported entries, exactly two are never visited as leaves — those two. So I deleted the tag from `blockchain_settings.go:20` and re-ran:

```
$ go test ./settings/ -run 'TestNoMissingTags|TestMetadataCoverage|TestKeyTagExemptNotExported'
ok      github.com/bsv-blockchain/teranode/settings     0.306s

entries 610 -> 621 ; blockchain_postgres_pool exported = false
```

All three guards stay green while the setting disappears from the portal — precisely the regression this test exists to catch. It's worse than a bare loss, too: the 12 children then start exporting under `BlockChain.PostgresPool.*`, and since `PostgresSettings` is already exported via the untagged top-level `Settings.Postgres`, that silently adds **12 duplicate keys**.

The walk can't infer this on its own, because it has no way to distinguish an intentionally untagged container (`Kafka KafkaSettings`, correctly tag-free) from an override whose tag was dropped by accident. The cheap, honest fix is to name the struct-level keys:

```go
// Struct-typed override settings are exported as a single aggregate entry, so the
// walk above steps past their tag on its way to the children. Pin them by key.
for _, k := range []string{"blockchain_postgres_pool", "utxostore_postgres_pool"} {
    require.True(t, exportedKeys[k], "struct-level override setting %q lost its key tag", k)
}
```

`exportedKeys` is one pass over `extractMetadataStructure()`, which the test file already calls. Folding key-uniqueness into the same pass would cover the duplicate-key half for free.

## Minor — the two walkers in the test file still disagree with each other

`settingsWalk` (`settings/coverage_test.go:92`) descends into a settings-package struct regardless of its tag; `settingsWalkExempt` (`settings/coverage_test.go:208-210`) bails on any non-empty tag, so it does **not** descend into `PostgresPool`. A `key:"-"` field added inside `PostgresSettings` would therefore be visited-and-skipped by the first walker but never collected by the second, quietly weakening `TestKeyTagExemptNotExported` rather than failing it.

Latent — no such field exists today. But drift between two near-identical reflection walks is the exact failure mode this PR was written to stop, and the second walker is a strict specialisation of the first: `settingsWalkExempt` can be replaced by a single `settingsWalk` call whose callback records `field.Tag.Get("key") == keyTagExempt`. That drops ~28 lines and makes the divergence impossible by construction.

## Nits

- **Dead parameter.** `settingsWalk`'s callback takes `(fieldName, field, fieldType)` and both call sites discard the third as `_` (`coverage_test.go:118`, `:141`). Drop it; `field.Type` is reachable if ever needed.
- **`redact.go` doesn't know the new sentinel.** `settings/redact.go:109` collects sensitive keys with `if k := f.Tag.Get("key"); k != ""`, which now accepts `"-"` — a future field carrying both `redact:"true"` and `key:"-"` would register the literal `"-"` in `sensitiveKeys`. Harmless today (nothing exports a key of `"-"`, so nothing is redacted spuriously), but the sentinel now has a second consumer that doesn't recognise it. `k != "" && k != keyTagExempt` closes it while the convention is new.
- **Stale count comment.** `settings/coverage_test.go:20` still reads *"We should have 398 settings based on the complete migration"* while the assertion below is `>= 300` and the real count is 610. The matching "398 configurable fields" log line was deliberately deleted from the old `TestNoMissingTags`; this sibling survived in the same file.
- **The security-doc date is now correct-today rather than correct-always.** `minersSecurityBestPractices.md:3` reads `20-August-2026`, matching the commit that set it, so ChiR5 is answered. It's still the only file in the set with real content changes (nine blank lines) and will be wrong again if this merges on a different day. The dependable version of the whole exercise is a CI check on the header rather than another manual pass — 11 more docs outside `kubernetes/` carry the same header and are still stale.
- **Commit hygiene.** That docs date change rides in `5e7983913`, whose subject is entirely about settings (`align TestNoMissingTags with extractFields, drop dead BlocksValidate field`), so the docs edit is orphaned from any message describing it.

## Follow-up, now proven: `ExportMetadata` really does emit duplicate keys

I flagged this last round as high-confidence-but-unproven. Proven now — ran `extractMetadataStructure()` and grouped by key:

```
total entries=610   distinct keys=608   duplicate keys=2
DUP blockmaxsize                                      <- Block.MaxSize , Policy.BlockMaxSize
DUP blockvalidation_processTxMetaUsingStore_BatchSize  <- Block.ProcessTxMetaUsingStoreBatchSize ,
                                                         BlockValidation.ProcessTxMetaUsingStoreBatchSize
```

Two pairs of fields fighting over one config key each is a worse defect than a missing tag. Out of scope here, but the walk already visits every field, so a uniqueness assertion alongside tag-presence is nearly free.

## What I verified

Doc half — all eight headers checked against `git log -1` on the merge-base rather than taken on trust:

| File | Header now | Last pre-PR commit |
|---|---|---|
| minersHowToConfigureTheNode.md | 15-July-2025 | 15-July-2025 |
| minersHowToInstallation.md | 1-July-2026 | 01-July-2026 |
| minersHowToResetTeranode.md | 15-June-2026 | 15-June-2026 |
| minersHowToStopStartKubernetesTeranode.md | 1-July-2026 | 01-July-2026 |
| minersHowToSyncTheNode.md | 1-July-2026 | 01-July-2026 |
| minersHowToTroubleshooting.md | 21-October-2025 | 21-October-2025 |
| minersSecurityBestPractices.md | 20-August-2026 | 05-July-2025 (file is edited by this PR) |
| minersUpdatingTeranode.md | 17-November-2025 | 17-November-2025 |

Go half:

- **No runtime behaviour change.** `metadataEntry.FieldName` now carries a dotted path; I grepped every consumer — it's read only inside `settings/export.go` and this test file, and it is **not** a member of `SettingMetadata`, so nothing crosses the API into `ui/dashboard`. The five newly-exempt fields previously hit the `key == ""` branch and, being non-struct leaves, fell through to `continue`; the new `key == keyTagExempt` branch gives the identical outcome, and the export still yields 610 entries.
- **`key` tag consumers are only `export.go`, `redact.go` and these tests** — no config loader reads it, so `key:"-"` cannot affect how a value is loaded.
- **Coverage measured, not assumed:** the walk visits 632 leaves against 610 exported entries. The 24 extra are the `PostgresPool.*` children, whose tags are exported anyway via the untagged top-level `Settings.Postgres` — duplicate paths through the same type, not dead tags.
- **Green locally:** `go test ./settings/...`, `go test -race ./settings/...`, `go vet ./settings/...`, `golangci-lint run ./settings/...` (0 issues), `gofmt -l settings/` (clean).

## Testing limits

Settings package only — the heavier CI stages (benchmarks, sequential, smoke) are impractical locally, though a test file plus struct tags carries little risk there. Worth flagging that GitHub has run almost nothing on this PR: the only checks reported are the two Socket Security jobs (one skipping, one pass), so there's no CI evidence of the Go suite passing, only my local runs. And the PR is currently **conflicting with main** — it needs a merge before any of this lands, and I reviewed the head as it stands rather than a merged result.

docs/howto/miners/kubernetes/minersHowToInstallation.md
```diff
@@ -1,6 +1,6 @@
 # How to Install Teranode with Kubernetes Helm
 
-Last modified: 29-October-2025
+Last modified: 1-July-2026
```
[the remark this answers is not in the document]
icellan: Right, and fixed in `b313e3fda` without rebasing — all four now read 7-September-2026.

Derived rather than taken on trust: `git merge-base --is-ancestor e53520fd HEAD` returns 0, and that commit touches all four files substantively (it rewrote the setfsmstate command, e.g. `--fsmstate RUNNING` to `--fsmstate catchingblocks`). Per-file `git log` confirms it is the most recent content change in each case, ignoring this PR own header-only commit. Day is unpadded to match the convention in that directory.

Worth saying plainly: this is inherent to hand-maintained "Last modified" headers — if main advances over these files again before merge, they go stale once more.

freemans13 (commented):
Re-reviewed at `b313e3fd`. The docs half is now right on every file it touches, and I could not fault the four new dates. The Go half is byte for byte what I reviewed on 20 August, so the one item I asked for then is still open; I re-proved it at this head rather than assuming it from the diff.

Not approving yet, for that one item. Nothing here is a runtime bug: the Go change is struct tags plus a test.

## Docs: the seven headers in the diff all check out

Each header checked against the last commit that changed that file's content, on the merge base `712605a5f`:

| File | Header at this head | Last content change on main |
|---|---|---|
| minersHowToConfigureTheNode.md | 15-July-2025 | `26ba6a8a0`, 15-July-2025 |
| minersHowToInstallation.md | 7-September-2026 | `e53520fdf`, 07-September-2026 |
| minersHowToResetTeranode.md | 7-September-2026 | `e53520fdf`, 07-September-2026 |
| minersHowToStopStartKubernetesTeranode.md | 7-September-2026 | `e53520fdf`, 07-September-2026 |
| minersHowToSyncTheNode.md | 7-September-2026 | `e53520fdf`, 07-September-2026 |
| minersHowToTroubleshooting.md | 21-October-2025 | `47d39300c`, 21-October-2025 |
| minersSecurityBestPractices.md | 20-August-2026 | `8d839dc38`, 20-August-2026 |

The blank-line cleanup that ChiR5 objected to is gone from `minersSecurityBestPractices.md`, and its header now names a real commit instead of the day the branch happened to be pushed. That is a better answer than the one I reviewed last round, because it stays true if this sits for another month.

## Minor: two guides in the same directory are left stale

The set this PR dates is seven of the nine guides under `docs/howto/miners/kubernetes/`. The other two are stale by the same rule the PR applies, and one of them was in the diff until the rebase:

- `minersUpdatingTeranode.md` reads `11-August-2026`. Its content last changed on 20 August in `8d839dc38`, the same commit that changed `minersSecurityBestPractices.md`, which this PR does re-date to 20-August-2026. The header and the commit disagree because `8d839dc38` wrote the date it was drafted rather than the date it landed. Earlier heads of this branch dated this file `17-November-2025`, correct at the time; the rebase took upstream's side and the file dropped out of the diff, so the fix was lost silently rather than deliberately.
- `minersHowToExposeP2P.md` reads `11-June-2026`. Its prose changed on 9 September in `99b8579b4`, which rewrote the paragraph on which port peers dial. It has never been in this PR's set.

Your point in the thread above stands, and this is the evidence for it: main advanced over this directory twice while the PR was open, and it will again. A CI check on the header is the version that holds. Nothing to do here beyond deciding whether to sweep the last two in, and I would not hold the PR for it.

## Still open from 20 August: a settings key can be deleted with every guard green

Unchanged and unanswered, so I am not restating the analysis, only confirming it survives at this head: [the detail is in my 20 August review](https://github.com/bsv-blockchain/teranode/pull/1568#pullrequestreview-4982116528).

Re-ran the same mutation on `b313e3fd`. Deleting the `key` tag from `settings/blockchain_settings.go:24`:

```
$ go test ./settings/ -run 'TestNoMissingTags|TestMetadataCoverage|TestKeyTagExemptNotExported' -count=1
ok      github.com/bsv-blockchain/teranode/settings     0.220s

exported entries 652 -> 663 ; blockchain_postgres_pool exported = false ; duplicate keys 2 -> 14
```

All three guards stay green while one setting disappears from the exported metadata and twelve of its children start exporting a second time under a different path. The two fields this applies to are still exactly two, so the hole has not widened: `settings/blockchain_settings.go:24` and `settings/utxostore_settings.go:52`. The suggested fix, asserting those two keys are present by name, is four lines in the test file.

## What I verified, and what I did not

- CI at `b313e3fd`: 29 checks green, one skipped (`test aerospike native-ops (fork image)`).
- `go test ./settings/ -count=1` green locally at `b313e3fd`, and `go build ./settings/...` clean.
- The mutation above, run twice, clean and mutated, with the tree restored afterwards.
- I did not re-read the Go diff line by line this round. I diffed the branch-authored changes under `settings/` at `5e79839` against `b313e3fd` and they are identical, so my 20 August reading still applies.
- ctnguyen's `CHANGES_REQUESTED` is still the PR's formal verdict. It was submitted against `f420837`, two heads back, and its blocking item ChiR1 was fixed at the next commit. Someone has to re-review or dismiss it before this can merge regardless of anything above.

icellan:
Guard hole closed in `45e9c342248b39d6a037fe5ddb65cd06e383ec63`.

Reproduced it before fixing rather than taking it on trust, and your numbers came back bit for bit: dropping the `key` tag from `blockchain_settings.go:24` leaves `TestNoMissingTags|TestMetadataCoverage|TestKeyTagExemptNotExported` green, exported entries go 652 -> 663, duplicate keys 2 -> 14, and `blockchain_postgres_pool` exported = false.

Root cause, for the record: `extractFields` only recurses into a nested settings struct when that field own `key` tag is empty, whereas the `TestNoMissingTags` walk always recurses so it can check the children. So the guard structurally cannot see a dropped tag on an opaque struct — it is not a gap in coverage but a blind spot in the traversal.

Added the by-name assertion inside `TestMetadataCoverage`, which already builds the registry, so it is 14 lines including the comment. Verified it fires: dropping either tag in turn now fails with the matching message, tree restored and clean between runs.

One deviation from what you suggested, and I think it matters. I tried the generalization — "every struct field carrying a nested settings struct has a `key` tag" — and it is not just harder to keep small, it is false. `Settings.Debug DebugSettings` is embedded with no tag at all and deliberately flattened through, children exported individually; that is the normal case. The two `PostgresPool` fields are the unusual one: opaque single-setting structs reused across parents. The only reflective signal distinguishing "opaque" from "flatten-through" is the very tag being tested for. So by-name for those two is the correct shape, and they are the complete set — `*PostgresSettings` has exactly two occurrences in the tree.

A no-duplicate-keys invariant would also have caught it, but the baseline already carries 2 unrelated duplicates, so that needs its own cleanup first.

On the two remaining stale guides: taking your advice and not holding the PR for them. Your CI-check point is the right conclusion — I re-dated four files by hand and one of them had already gone stale once from a rebase taking upstream side silently. That is the failure mode repeating, not an accident.

ctnguyen formal CHANGES_REQUESTED is still the blocker here and on nineteen other PRs in this batch, all pushed to since. Chasing that separately.

settings/interface.go
```diff
@@ -52,14 +52,14 @@ type Settings struct {
...
+	ChainCfgParams               *chaincfg.Params `key:"-"` // Runtime-derived; exported separately as the "network" setting
```
freemans13, line 62: **Blocker.** This sentinel, and the four others like it, now fail `TestNoNewDeadSettingKeys`, which landed on main this morning in `322f47603` (#1803).

That test collects every non-empty `key` tag and fails any whose value is never read back in the settings package. `-` is non-empty and nothing reads `("-"`, so the merged result reports all five as dead keys: `Should be empty, but was [- - - - -]`.

I merged main into `45e9c342` locally and reproduced it. One line in main's file clears it, and the settings suite then goes 320-green with `-race` and `vet` clean:

```go
// settings/blockassembly_txmapdirs_test.go, inside collectKeyTags
if key, ok := field.Tag.Lookup("key"); ok && key != "" && key != keyTagExempt {
```

settings/coverage_test.go
```diff
@@ -19,6 +21,20 @@ func TestMetadataCoverage(t *testing.T) {
...
+	// below silently stops emitting the struct-level setting and instead
+	// flattens its children into keys that already exist elsewhere - both
+	// changes TestNoMissingTags cannot see, since its walk always descends
+	// into settings-package structs regardless of their own key tag.
+	exportedKeys := make(map[string]bool, len(registry.Settings))
+	for _, setting := range registry.Settings {
+		exportedKeys[setting.Key] = true
+	}
+	require.True(t, exportedKeys["blockchain_postgres_pool"], "BlockChainSettings.PostgresPool must keep its key tag")
+	require.True(t, exportedKeys["utxostore_postgres_pool"], "UtxoStoreSettings.PostgresPool must keep its key tag")
```
freemans13, line 36: Verified both of these fire. Dropping either tag now fails `TestMetadataCoverage` naming the field, so the hole I raised on 20 August is closed.

**Residual, not blocking.** The pair is correct today but does not maintain itself: a third `*SomeSettings` override field added later reopens the same hole with every guard green, and nothing here prompts its author to extend the list.

I tried the derived version you rejected and you were right, for a sharper reason than awkwardness: it passes vacuously. "Every settings-package struct field carrying a real key tag must be in the registry" is conditioned on the tag existing, so deleting the tag deletes the assertion. I ran it against the dropped `blockchain_postgres_pool` tag and it stayed green.

A census assertion is a third option that catches both the drop and a future addition:

```go
var structKeyed []string
settingsWalkStructKeys(&structKeyed) // any settings-pkg struct field with a real key tag
require.ElementsMatch(t,
    []string{"BlockChain.PostgresPool", "UtxoStore.PostgresPool"}, structKeyed,
    `the set of struct-typed override settings changed; add the new key to the assertions above`)
```

Ran it: green clean, and fails with "the set of struct-typed override settings changed" once a tag is dropped.

settings/coverage_test.go
```diff
@@ -19,6 +21,20 @@ func TestMetadataCoverage(t *testing.T) {
 	require.NotNil(t, registry)
 	require.GreaterOrEqual(t, len(registry.Settings), 300, "Should have at least 300 settings after complete migration")
```
freemans13, line 22: Nit, carried over. The comment on line 20 still reads "We should have 398 settings based on the complete migration" while this assertion is `>= 300` and the real count at this head is 653.

The matching "398 configurable fields" log line was deliberately deleted from the old `TestNoMissingTags`; this sibling survived in the same file.

settings/coverage_test.go
```diff
@@ -54,9 +70,166 @@ func TestMetadataCoverage(t *testing.T) {
...
+		require.False(t, exemptFieldPaths[entry.FieldName], "exempt field %q was exported as a setting", entry.FieldName)
+	}
+}
+
+// settingsWalkExempt recursively collects the dotted field path of every
+// key:"-" field in typ, descending into settings-package structs the same way
+// extractFields does (a real key on a struct field stops extractFields from
+// recursing, so it stops here too - only the keyTagExempt sentinel itself is
+// what needs collecting, and it can appear at any depth).
+func settingsWalkExempt(settingsPkgPath string, typ reflect.Type, prefix string, exempt map[string]bool) {
```
freemans13, line 207: Nit, carried over and latent. This walker bails on any non-empty `key` tag, so it does not descend into `BlockChain.PostgresPool` or `UtxoStore.PostgresPool`. `settingsWalk` above (line 88) does descend into them regardless of tag.

A `key:"-"` field added inside `PostgresSettings` would therefore be visited-and-skipped by the first walker but never collected by this one, quietly weakening `TestKeyTagExemptNotExported` rather than failing it. No such field exists today.

Drift between two near-identical reflection walks is the exact failure mode this PR was written to stop, and this walker is a strict specialisation of the first: it can be replaced by a `settingsWalk` call whose callback records `field.Tag.Get("key") == keyTagExempt`.

freemans13 (changes requested):
Re-reviewed at `45e9c342`. The item I blocked on last round is fixed, and I proved it by mutation rather than by reading the diff. A different blocker turned up while you were working, and it came from main rather than from you: the `test` check is red, and merging as-is keeps it red.

Requesting changes for that one item only. The fix is four characters.

## New blocker: your `key:"-"` sentinel trips a guard main added this morning

`TestNoNewDeadSettingKeys` landed on main today in `322f47603` (#1803). It collects every non-empty `key` struct tag on `Settings` and fails any key that is never read back anywhere in the settings package, on the grounds that a tag with no lookup is a setting operators can configure to no effect.

Your sentinel is a non-empty `key` tag whose value is the literal `-`. Nothing reads `("-"`, so all five exempt fields are reported as dead keys. That is the `test` failure sitting on this PR right now:

```
=== FAIL: settings TestNoNewDeadSettingKeys (0.03s)
    Error:    Should be empty, but was [- - - - -]
    Messages: these keys are declared on Settings but never read by NewSettings,
              so configuring them does nothing: [- - - - -]
```

Five entries because five fields carry `key:"-"`: `Commit`, `Version`, `Context`, `IsAllInOneMode` and `ChainCfgParams`.

Reproduced, not inferred. I merged `bsv-blockchain/main` into `45e9c342` in a throwaway worktree and got the same five-element failure. One line in main's file clears it:

```go
// settings/blockassembly_txmapdirs_test.go, inside collectKeyTags
if key, ok := field.Tag.Lookup("key"); ok && key != "" && key != keyTagExempt {
```

On that merged tree with the line applied:

```
go test ./settings/... -count=1          Go test: 320 passed in 1 packages
go test -race ./settings/... -count=1    Go test: 320 passed in 1 packages
go vet ./settings/...                    clean
```

Neither side did anything wrong here; the sentinel and the dead-key guard just landed the same week and neither knows about the other. But this PR is the one that goes red, and it is the PR that introduces the sentinel, so the teaching-the-guard line belongs in it.

## Fixed: the hole I have been holding since 20 August

`45e9c342` closes it. Verified at this head by dropping each tag in turn, tree restored and clean between runs:

```
drop key:"blockchain_postgres_pool"  (settings/blockchain_settings.go:24)
  --- FAIL: TestMetadataCoverage    coverage_test.go:35
      BlockChainSettings.PostgresPool must keep its key tag

drop key:"utxostore_postgres_pool"   (settings/utxostore_settings.go:52)
  --- FAIL: TestMetadataCoverage    coverage_test.go:36
      UtxoStoreSettings.PostgresPool must keep its key tag
```

Clean tree green. Both fire, both name the field, and the message tells you what to do.

## You were right to reject the generalisation, and here is the proof

You said the generalisation you tried was false rather than just fiddly. It is worse than false. I built it, and it passes vacuously.

The natural derived rule is "every settings-package struct field carrying a real `key` tag must appear in the exported registry under that key". I wrote it as a throwaway test. On a clean tree it finds exactly your two fields and passes. Then I dropped the `blockchain_postgres_pool` tag and re-ran it, and it stayed **green**. Obvious once seen: the rule is conditioned on the tag existing, so deleting the tag deletes the assertion with it. Every derived form of that shape is structurally unable to catch a dropped tag, which is exactly why naming the two keys works and deriving them does not.

There is a third option that is neither, and it is worth four lines. Assert the **census** rather than the keys: the set of settings-package struct fields carrying a real key tag is exactly `{BlockChain.PostgresPool, UtxoStore.PostgresPool}`. I ran that too:

- clean tree: pass, finds `[BlockChain.PostgresPool UtxoStore.PostgresPool]`
- `blockchain_postgres_pool` tag dropped: fail, "the set of struct-typed override settings changed"

It catches the drop, and it also fires when someone adds a third field of that shape, which your two `require.True` lines cannot. That is the residual worth naming: the pair is correct today and does not maintain itself. A third `*SomeSettings` override field added later reopens the identical hole with every guard green, and nothing in the file prompts its author to extend the list.

Not blocking, and your call. Detail inline on the assertions.

## Docs half: still correct, and main has not eaten it again

Main has not touched `docs/howto/miners/kubernetes/` since the merge base `712605a5f`, so all seven headers you set in `b313e3fd` are still accurate at this head. Given the point you made in the thread about hand-maintained headers, that is luck rather than a property, but it holds today.

## Carried over from earlier rounds, still open, none blocking

- **Stale count comment.** `settings/coverage_test.go:20` still reads "We should have 398 settings based on the complete migration" while the assertion below is `>= 300` and the real count at this head is 653. The matching "398 configurable fields" log line was deliberately deleted from the old `TestNoMissingTags`; this sibling survived in the same file.
- **The two walkers in the test file still disagree.** `settingsWalk` (`coverage_test.go:88`) descends into a settings-package struct regardless of its tag. `settingsWalkExempt` (`coverage_test.go:207`) bails on any non-empty tag, so it does not descend into either `PostgresPool`. Latent, because no `key:"-"` field exists inside `PostgresSettings` today. Detail inline.
- **Dead callback parameter.** `settingsWalk`'s callback takes `(fieldName, field, fieldType)` and both call sites discard the third as `_`. `field.Type` is reachable if it is ever wanted.
- **`redact.go` does not know the sentinel.** `settings/redact.go:109` collects sensitive keys with `if k := f.Tag.Get("key"); k != ""`, which now accepts `"-"`. A future field carrying both `redact:"true"` and `key:"-"` would register the literal `"-"` in `sensitiveKeys`. Harmless today, since nothing exports a key of `"-"`, but the sentinel now has a second consumer that does not recognise it. `k != "" && k != keyTagExempt` closes it while the convention is still new.
- **Two guides in the same directory are still stale**, by the rule this PR applies. `minersUpdatingTeranode.md` reads 11-August-2026 but its content last changed on 20 August in `8d839dc38`, the same commit this PR credits for `minersSecurityBestPractices.md`. `minersHowToExposeP2P.md` reads 11-June-2026 but its prose last changed on 9 September in `99b8579b4`. I would not hold the PR for either.

## The other red check is not yours

`smoketest (shard 2)` fails on `TestCatchup` in `test/e2e/daemon/ready` with "Condition never satisfied". Evidence rather than a hunch that this PR did not cause it: the same test fails in the same window on `teraslab/integration-wip`, an unrelated branch, and of the four failing smoketest runs in the last twenty, two fail on `TestCatchup` and two do not. Nothing in this diff can reach that test either, since the `key` tag is read only by `export.go`, `redact.go` and these test files, and the `interface.go` change is struct tags plus gofmt realignment with no runtime code in it.

## What I verified, and what I did not

- Both `PostgresPool` mutations above, run at `45e9c342`, tree confirmed clean afterwards with `git status --short`.
- The derived-rule and census probes, written as throwaway test files in the settings package, run clean and mutated, then deleted. Tree clean.
- The merge against `bsv-blockchain/main` (`fbc62933d`), the `[- - - - -]` failure, the one-line fix, and the 320-test green that follows it. Throwaway worktree, since this is your branch and I do not commit to it.
- `go test ./settings/... -count=1`, `-race`, `go vet ./settings/...` and `gofmt -l settings/` on the unmerged head: all clean, 306 tests.
- Doc headers checked against `git log` on `bsv-blockchain/main` rather than taken on trust.
- I did not re-read the Go diff line by line this round. The only change since `b313e3fd` is the 14-line addition in `45e9c342`, so my 20 August reading of the rest still applies.
- I did not run the heavier CI stages locally. CI at this head is 34 pass, 1 skipping, 2 fail, and both failures are covered above.

## Formal state

ctnguyen's `CHANGES_REQUESTED` from 20 August is still this PR's recorded verdict. It was submitted against `f420837`, four heads back, and its blocking item ChiR1 was fixed at the very next commit. Someone has to re-review or dismiss it before this can merge, independently of anything above.

icellan:
Follow-up: rebased onto main at `b492d48d1` to fix a CI failure that only existed in the merge preview, not on the branch.

This PR introduces the `key:"-"` convention — main has none. Main has since gained `TestNoNewDeadSettingKeys`, whose `collectKeyTags` accepts any non-empty key tag, so on the merged tree it read `-` as a configurable key and reported all five runtime-computed fields (`Commit`, `Version`, `Context`, `IsAllInOneMode`, `ChainCfgParams`) as dead settings. Neither guard is wrong alone; they only disagree once combined, which is why the branch tested clean and the merge did not.

Fixed it in the branch that introduces the convention: `collectKeyTags` now skips `key:"-"`, the same marker `export.go` already honours. Three lines.

Checked the guard is still protective rather than merely quiet — injected a field tagged `key:"probe_never_read_anywhere"` and it failed with that key named, then passed again once removed. Relaxing a guard filter without proving it still fires would have been the worse outcome here.

The rebase itself was clean, no conflicts. Your re-review at `b313e3fd` still applies to the docs half; the only change since is the PostgresPool key-tag assertion and this one.

ordishs (approved):
The walk now enforces the tag contract, and the `key:"-"` exemption is consistent with `ExportMetadata`. `ChainCfgParams` emitted nothing before, so its output does not change. `BlocksValidate` has no references. The settings tests pass on the head and on a trial merge with current main.

One small item: #1493 merged today and changed the probe section of `minersHowToTroubleshooting.md` without a header bump. After a rebase, set that header to `23-September-2026`, or the corrected date will be stale when this merges.

settings/interface.go
icellan: We found this independently the same day — I hit it via the merge-preview CI failure and fixed it in `b492d48d1`, before seeing your comment. Your version is better and is now what is there (`03f7a42e8`): `keyTagExempt` rather than my bare `"-"` literal. Confirmed the constant lives in `settings/export.go` and swept the rest of the diff for other hardcoded `"-"` in that sense — only comments, no code.

settings/coverage_test.go
icellan: Census assertion added in `6ed6cc8e8`, and your reason for rejecting the derived version is sharper than mine — I argued it was false in general because most nested settings structs are deliberately flattened through; you showed it also passes vacuously, since conditioning on the tag means deleting the tag deletes the assertion. Both are true and yours is the one that matters.

Kept the two by-name `require.True` checks alongside the census: they give a precise failure naming the field, while the census catches the case neither of us can enumerate in advance — a third override field appearing.

One deviation. `settingsWalkStructKeys` is its own small traversal rather than a `settingsWalk` call, because `settingsWalk` only invokes `visit` for leaf fields and never for the struct fields it recurses into, so it structurally cannot report "this struct field carries a real key tag" without a new hook. It is documented as mirroring `settingsWalk`, recursion rule included.

Probed both directions: dropping `blockchain_postgres_pool` fails with `elements differ`, and adding a second `*PostgresSettings` field to `UtxoStoreSettings` fails with `extra elements in list B`.

settings/coverage_test.go
icellan: Fixed. The real count at this head is 656, so the comment now describes the assertion as a regression floor rather than an exact count, and the floor moved from 300 to 600.

Judgment call worth stating: 600 rather than something tighter to 656, so the guard is meaningful without being brittle against every settings addition. Say if you want it closer to the real number.

settings/coverage_test.go
icellan: Merged, and the drift was real — proved rather than assumed. Added a temporary `key:"-"` field inside `PostgresSettings`: the new code collects it at all three embedding sites (`BlockChain.PostgresPool.*`, `Postgres.*`, `UtxoStore.PostgresPool.*`), while the old walker, restored from a pre-change copy and run against the same probe, collected only `Postgres.*` — missing both nested sites, exactly as you described.

Your suggested mechanism did not work verbatim: `settingsWalk` filters exempt fields with `continue` before invoking `visit`, so a callback checking `key == keyTagExempt` never sees them. Extended it minimally instead — `settingsWalk` now delegates to `settingsWalkExemptAware`, which takes an optional `visitExempt` callback for `key:"-"` fields. `settingsWalk`s signature and behaviour are unchanged, so the existing callers needed no edits, and the specialised walker is gone. One traversal, as you wanted.

icellan:
Separate observation from working through the above, not part of this PR and not proposing to fix it here.

`TestBlockAssemblyTxMapDirs_DefaultEmpty` fails on any developer machine whose `settings_local.conf` sets `blockassembly_txMapDirs` — `NewSettings()` reads ambient local config with no isolation, and gocore walks up parent directories to find it, so a worktree several levels down still picks up the repo-root file. Reproduced on a byte-clean tree to confirm it is not something in this branch. CI is unaffected, since there is no `settings_local.conf` there.

It only started mattering when that key went live recently. Mentioning it because it is the same shape as what this PR is about: a guard that looks green or red for reasons unrelated to what it is guarding.

PR #1569, opened 2026-08-10
refactor(validator): consolidate the 2PC unlock into one deferred exit path

icellan:
## What this is

Not a bug fix. On `main`, both error paths between UTXO persistence and the end of `validateInternal` (tx-inpoints construction failing, `sendToBlockAssembler` failing) already keep the transaction `Locked`. That's the correct, intentional behaviour: an undelivered-to-block-assembly transaction must stay locked, or a child spending it could validate and enter a mining template before its parent, producing a block other implementations reject.

This PR:

- Consolidates the lock release into a single deferred helper, `unlockLockedTxOnExit`, registered once right after the tx is durably persisted and spent, instead of releasing the lock only at the tail of the happy path.
- Documents the invariant directly on the helper: why those two error paths must stay locked, and why staying locked is a safe, self-healing failure mode.
- Gates the unlock decision on an explicit `delivered` flag rather than inferring "was this tx delivered to block assembly" from "is `err` nil". `err` is nil for the entire window between entering the add-to-block-assembly code and `sendToBlockAssembler` returning, so a panic in that window (e.g. a nil/misconfigured block-assembly client) would incorrectly unlock an undelivered tx if the decision were based on `err == nil` alone. `delivered` only flips to `true` once `sendToBlockAssembler` has actually returned successfully, so it stays panic-safe.
- Adds unit tests: the two headline behaviours (stays locked on error, unlocks on success), a no-op case, an unlock-failure-is-surfaced case, a panic-during-delivery case (new - the deferred/flag-based approach has to get this right, the old tail-of-function code never had this failure mode), and an end-to-end test through `ValidateWithOptions` that exercises the real named-return/defer mechanism rather than calling the helper directly.

## Recovery path

The self-healing recovery is `BlockAssembler.loadUnminedTransactions`: it adds a still-locked, undelivered transaction to the subtree processor and only then clears its `Locked` flag, so a child transaction can never validate against an unlocked-but-undelivered parent. See `docs/topics/features/two_phase_commit.md`. Note this ordering guarantee has one pre-existing exception under default `OnRestartValidateParentChain`/`OnRestartRemoveInvalidParentChainTxs` settings, called out in the updated doc comment.

### DIFF
--- a/services/validator/Validator.go
+++ b/services/validator/Validator.go
@@ -944,6 +944,17 @@ func (v *Validator) validateInternal(ctx context.Context, tx *bt.Tx, blockHeight
 		return nil, err
 	}
 
+	// From this point on the tx has been durably persisted with its inputs spent and,
+	// when addToBlockAssembly is set, its own record marked Locked (the 2PC
+	// in-progress marker). Every return path below - success or error - must release
+	// that lock, or the tx is stuck forever: persisted, spent, Locked, and never
+	// delivered to block assembly. This is deliberately separate from spend rollback:
+	// reverseSpends only ever applies before persistence, and spends must stand once
+	// the tx is durably created (see spendAndCreateInUtxoStore). The argument
+	// (txMetaData) is captured now, by value of the pointer, so a later `return nil,
+	// err` in this function does not affect what the deferred call unlocks.
+	defer v.unlockLockedTxOnExit(decoupledCtx, tx, txID, txMetaData, &err)
+
 	if validationOptions.SkipUtxoCreation {
 		// create the tx meta needed for the block assembly
 		if validationOptions.OutpointOnlySpend {
@@ -1001,17 +1012,47 @@ func (v *Validator) validateInternal(ctx context.Context, tx *bt.Tx, blockHeight
 		}
 	}
 
-	if txMetaData.Locked {
-		if err = v.twoPhaseCommitTransaction(decoupledCtx, tx, txID); err != nil {
-			v.logger.Warnf("[Validate][%s] error during two phase commit, transaction will be marked as spendable on next block: %v", txID, err)
+	// Unlocking (if the tx is still Locked) happens in the deferred
+	// unlockLockedTxOnExit call registered above, on this and every other return
+	// path from this point in the function.
+	return txMetaData, nil
+}
 
-			return txMetaData, err
-		}
-
-		txMetaData.Locked = false
+// unlockLockedTxOnExit is the single point that releases a tx's two-phase-commit
+// Locked flag on the way out of validateInternal. It is deferred immediately after
+// the tx has been durably persisted with its inputs spent (see validateInternal),
+// so it runs on every return from there on - success or error - and not just the
+// happy path. Without it, an error returned between persistence and the end of
+// the function (e.g. building tx inpoints, or sendToBlockAssembler failing) would
+// leave the tx persisted, spent, and Locked forever, since nothing else ever
+// clears the flag.
+//
+// It does not reverse the spend: that is reverseSpends' job, and it only applies
+// before persistence (see spendAndCreateInUtxoStore) - once the tx is durably
+// created it is valid and its spends must stand.
+//
+// If err is nil on entry (the pure success path) and the unlock itself fails, err
+// is set to the unlock error so the caller is told the tx still needs the
+// fallback recovery ("marked as spendable on next block"). If err is already
+// non-nil (an earlier failure in validateInternal), the unlock is still attempted
+// and its failure only logged, so the tx's original, more informative error is
+// not masked by a lock-release problem.
+func (v *Validator) unlockLockedTxOnExit(ctx context.Context, tx *bt.Tx, txID string, txMetaData *meta.Data, err *error) {
+	if txMetaData == nil || !txMetaData.Locked {
+		return
 	}
 
-	return txMetaData, nil
+	if unlockErr := v.twoPhaseCommitTransaction(ctx, tx, txID); unlockErr != nil {
+		v.logger.Warnf("[Validate][%s] error during two phase commit, transaction will be marked as spendable on next block: %v", txID, unlockErr)
+
+		if *err == nil {
+			*err = unlockErr
+		}
+
+		return
+	}
+
+	txMetaData.Locked = false
 }
 
 // getTransactionInputBlockHeights returns the block heights for each input of the transaction
--- a/services/validator/Validator.go
+++ b/services/validator/Validator.go
@@ -952,9 +952,21 @@ func (v *Validator) validateInternal(ctx context.Context, tx *bt.Tx, blockHeight
 	// undelivered tx must stay locked. This is deliberately separate from spend
 	// rollback: spend rollback only ever applies before persistence, and spends must
 	// stand once the tx is durably created (see spendAndCreateInUtxoStore). The
-	// argument (txMetaData) is captured now, by value of the pointer, so a later
-	// `return nil, err` in this function does not affect what the deferred call sees.
-	defer v.unlockLockedTxOnExit(decoupledCtx, tx, txID, txMetaData, &err)
+	// argument (persistedMeta) is captured now, by value of the pointer, so a later
+	// reassignment of txMetaData (e.g. on the SkipUtxoCreation path below) does not
+	// change what the deferred call sees.
+	//
+	// delivered tracks whether the tx has actually reached block assembly, which is
+	// the fact the unlock decision must be gated on. It starts true when this tx
+	// skips block assembly entirely (nothing to deliver), and is flipped to true
+	// only after sendToBlockAssembler returns successfully below. Deliberately NOT
+	// inferred from `err == nil`: a panic between entering the addToBlockAssembly
+	// block and sendToBlockAssembler returning would leave err nil while the tx was
+	// never delivered, and unlocking on that basis would be exactly the hazard this
+	// function exists to prevent.
+	persistedMeta := txMetaData
+	delivered := !addToBlockAssembly
+	defer v.unlockLockedTxOnExit(decoupledCtx, tx, txID, persistedMeta, &err, &delivered)
 
 	if validationOptions.SkipUtxoCreation {
 		// create the tx meta needed for the block assembly
@@ -992,6 +1004,8 @@ func (v *Validator) validateInternal(ctx context.Context, tx *bt.Tx, blockHeight
 
 			return nil, err
 		}
+
+		delivered = true
 	}
 
 	// Serialize and enqueue txmeta for the subtree validation kafka topic.
@@ -1035,26 +1049,53 @@ func (v *Validator) validateInternal(ctx context.Context, tx *bt.Tx, blockHeight
 // other implementations reject.
 //
 // Staying locked is the documented safe failure mode ("money temporarily can't
-// be spent") and it is self-healing, in the right order: the block assembler's
-// unmined-transaction loader adds such a tx to the subtree processor and only
-// then clears its Locked flag (see BlockAssembler.loadUnminedTransactions).
+// be spent") and it is normally self-healing, in the right order: the block
+// assembler's unmined-transaction loader adds such a tx to the subtree processor
+// and only then clears its Locked flag (see BlockAssembler.loadUnminedTransactions).
+// One exception: under the default OnRestartValidateParentChain /
+// OnRestartRemoveInvalidParentChainTxs settings, a transaction that gets filtered
+// out of the parent-chain validation during a block-assembly restart is
+// unconditionally unlocked as part of the same batch even though it was never
+// re-added - this is pre-existing block-assembly behaviour, not introduced here.
 //
 // It does not reverse the spend: spend rollback only applies before persistence
 // (see spendAndCreateInUtxoStore) - once the tx is durably created it is valid
 // and its spends must stand.
 //
-// On the success path, if the unlock itself fails, err is set to the unlock error
-// so the caller is told the tx still needs the fallback recovery ("marked as
-// spendable on next block").
-func (v *Validator) unlockLockedTxOnExit(ctx context.Context, tx *bt.Tx, txID string, txMetaData *meta.Data, err *error) {
+// On the success path, if the unlock itself fails, err is set to the unlock error.
+// No caller currently acts on that signal specifically - the real recovery path is
+// the same self-healing loadUnminedTransactions route described above (see
+// docs/topics/features/two_phase_commit.md), not caller-side handling of this
+// error. Setting err is mainly for observability of the failure at the point it
+// happened.
+//
+// The unlock decision is gated on delivered, not on *err being nil. delivered only
+// becomes true once sendToBlockAssembler has actually returned successfully, so it
+// stays false across a panic unwinding from anywhere in that window - unlike *err,
+// which is nil for the entire window regardless of whether delivery happened. Using
+// *err as the proxy would unlock a tx that was never delivered to block assembly if
+// something panicked in that narrow window, which is the exact state this function
+// exists to prevent.
+func (v *Validator) unlockLockedTxOnExit(ctx context.Context, tx *bt.Tx, txID string, txMetaData *meta.Data, err *error, delivered *bool) {
 	if txMetaData == nil || !txMetaData.Locked {
 		return
 	}
 
 	// Not delivered to block assembly - leave it locked for the unmined-transaction
-	// loader to heal, and keep the original, more informative error.
+	// loader to heal. This covers both ordinary error returns and a panic unwinding
+	// through the window between entering the addToBlockAssembly block and
+	// sendToBlockAssembler returning.
+	if delivered == nil || !*delivered {
+		v.logger.Warnf("[Validate][%s] tx stays locked, it was not delivered to block assembly", txID)
+
+		return
+	}
+
+	// Defensive: delivered is true but an error was still recorded. This should not
+	// happen given the current control flow (every return path after delivery sets
+	// no error), but keep it locked rather than trust an inconsistent state.
 	if *err != nil {
-		v.logger.Warnf("[Validate][%s] tx stays locked, it was not delivered to block assembly: %v", txID, *err)
+		v.logger.Warnf("[Validate][%s] tx stays locked despite being delivered, due to a later error: %v", txID, *err)
 
 		return
 	}
--- a/services/validator/Validator.go
+++ b/services/validator/Validator.go
@@ -946,13 +946,14 @@ func (v *Validator) validateInternal(ctx context.Context, tx *bt.Tx, blockHeight
 
 	// From this point on the tx has been durably persisted with its inputs spent and,
 	// when addToBlockAssembly is set, its own record marked Locked (the 2PC
-	// in-progress marker). Every return path below - success or error - must release
-	// that lock, or the tx is stuck forever: persisted, spent, Locked, and never
-	// delivered to block assembly. This is deliberately separate from spend rollback:
-	// reverseSpends only ever applies before persistence, and spends must stand once
-	// the tx is durably created (see spendAndCreateInUtxoStore). The argument
-	// (txMetaData) is captured now, by value of the pointer, so a later `return nil,
-	// err` in this function does not affect what the deferred call unlocks.
+	// in-progress marker). Routing every return path below through one deferred
+	// cleanup keeps the release in a single place instead of only on the tail of the
+	// happy path. It does NOT unlock on error: see unlockLockedTxOnExit for why an
+	// undelivered tx must stay locked. This is deliberately separate from spend
+	// rollback: spend rollback only ever applies before persistence, and spends must
+	// stand once the tx is durably created (see spendAndCreateInUtxoStore). The
+	// argument (txMetaData) is captured now, by value of the pointer, so a later
+	// `return nil, err` in this function does not affect what the deferred call sees.
 	defer v.unlockLockedTxOnExit(decoupledCtx, tx, txID, txMetaData, &err)
 
 	if validationOptions.SkipUtxoCreation {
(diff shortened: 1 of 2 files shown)


ordishs (commented):
The refactor is sound, but the title and the body describe a change that is not in the diff.

**Title and body do not match the code**

`gh pr view` shows two commits. The first released the lock on the two error paths. The second reverted that, and states that the premise did not hold. The final diff therefore keeps the `Locked` flag set on both named error paths, which is exactly what `main` already does.

I checked `main` directly: both error paths (tx-inpoints build failure, `sendToBlockAssembler` failure) already returned before the `Locked` check, so the transaction already stayed `Locked` before this PR.

What the diff actually delivers:

- a refactor that consolidates the `Locked` release into one deferred `unlockLockedTxOnExit`;
- documentation that explains why the two error paths must stay `Locked`;
- unit tests for the new helper.

All three are good, and the self-healing order holds: `BlockAssembler.loadUnminedTransactions` adds the transaction to the subtree processor before it unlocks.

Recommendation: retitle and rewrite the body to describe the refactor plus documentation, not a lock-release bug fix. As written, the PR reads as a behaviour change that reviewers will look for and not find.

ctnguyen (changes requested):
Review at commit [f47a07a](https://github.com/icellan/teranode/commit/f47a07a2a753f717f256b9d3b3349897b0e74379)

Nice arc on this one — you went looking for a lock-release bug, found the premise didn't hold, and turned it into a consolidation plus the documentation that was missing. The reasoning in the `unlockLockedTxOnExit` doc comment is sound and genuinely useful — the case for keeping those two paths locked is right, and the add-then-unlock sequence it points at is really there in both `loadUnminedTransactions` variants, though it isn't quite the unconditional guarantee the comment implies (ChiR4). The named-return/pointer-to-error plumbing is right, and I walked every ordinary return path — success, `ErrTxExists`, conflicting, DAH-evicted, coinbase, `SkipUtxoCreation`, and the two named error paths — and on all of them the refactor reproduces main's behaviour exactly. The one place it diverges is panic unwinding, which is ChiR1. Two things I'd want settled before this merges.

## Blocking issues

### ChiR1 — The deferred unlock uses "no error" as a proxy for "delivered to block assembly", so a panic unlocks an undelivered tx

**Problem:** `unlockLockedTxOnExit` decides whether to release the lock by inspecting `*err` (`Validator.go:1056`). From the `defer` at `Validator.go:957` through to `return txMetaData, nil` at `:1019`, `err` is nil. If anything *panics* rather than returning an error, the deferred call runs with `*err == nil`, calls `twoPhaseCommitTransaction`, and clears `Locked`. On main the unlock was a straight-line statement at the tail of the function, so a panic simply skipped it — this is a new path, and it points the wrong way.

The window that actually matters is narrow: `Validator.go:971-994`, from entering the `addToBlockAssembly` block until `sendToBlockAssembler` returns. A panic there unlocks a tx that never reached block assembly. The concrete candidate is `v.blockAssembler.Store` on a nil or misconfigured assembler (`Validator.go:1635` — reachable when `BlockAssembly.Disabled` is false but the client was never wired), plus whatever the block-assembly client does downstream of it.

For the avoidance of doubt, two nearby windows are *not* violations and shouldn't be read as part of this: a panic in the txmeta Kafka block at `:1003-1014` unwinds with the tx already delivered, so unlocking there is exactly what the success path does; and a panic in the `SkipUtxoCreation` block at `:959-969` is harmless because the snapshot the helper holds is nil on that path (see ChiR5).

**Why it matters:** This is precisely the state your own doc comment says must never exist — "unlocking it here would let a child spending it validate and enter this node's mining template without its parent". It is not purely theoretical: with `validator_useLocalValidator`, propagation embeds the validator and recovers panics around the call (`services/propagation/Server.go:796`, recovering `processTransactionInternal`, which calls `Validate` at `:1340`). The process survives, and the node then runs with an unlocked, undelivered parent until the next restart re-runs `loadUnminedTransactions`. (If the process instead dies, restart heals it — but that's the lucky case, not the designed one.)

**Fix:** Gate on the fact you actually care about instead of on the absence of an error. Something like:

```go
persistedMeta := txMetaData   // pin the object that carries the lock
delivered := !addToBlockAssembly
defer func() { v.unlockLockedTxOnExit(decoupledCtx, tx, txID, persistedMeta, &err, delivered) }()
...
if err = v.sendToBlockAssembler(...); err != nil { ... }
delivered = true
```

and in the helper, `if !delivered { keep the lock; return }` before the `*err` check. That is panic-safe, states the invariant directly, and stops a future edit that adds an early `return txMetaData, nil` before block assembly from silently unlocking. It also resolves ChiR5.

*Criticality: 60%*

### ChiR2 — Title and body still describe the lock-release fix that this branch reverted

*Same as [@ordishs's point "Title and body do not match the code" (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1569#pullrequestreview-4907182082).*

**Problem:** The title still reads "release the 2PC lock on the two error paths that skip it" and the body still describes those transactions as "stuck: persisted, spent, permanently Locked, never delivered to block assembly, never unlocked". Commit `24f1f5d` reverted that change and the head diff keeps the lock — the same as main. The head diff delivers a deferred-unlock consolidation, the invariant documentation, and four unit tests, and nothing else. The retitle/rewrite ordishs asked for has not been applied.

**Why it matters:** On squash merge the title and body become the permanent commit message in `main`. Anyone later doing archaeology on the 2PC lock reads a commit claiming this PR fixed a stuck-lock bug on those two paths — and, worse, a body that frames the intended safe failure mode as a defect. That is the exact class of misreading that makes someone "re-fix" it later by releasing the lock, which is the consensus hazard the new doc comment exists to prevent.

**Fix:** Retitle to something like `refactor(validator): consolidate the 2PC unlock into one deferred exit path` (or `docs(validator): …`), and rewrite the body to describe the consolidation, why the two error paths must stay locked, and the self-healing route through `loadUnminedTransactions`. Drop the "stuck transaction" narrative entirely.

*Criticality: 55%*

## Non-blocking issues

### ChiR3 — The doc comment promises a caller contract that no caller implements, and contradicts the design doc

**Problem:** `Validator.go:1046-1048` says that on unlock failure "err is set to the unlock error so the caller is told the tx still needs the fallback recovery". No caller does anything of the sort. `services/legacy/netsync/manager.go:1288` turns any non-orphan, non-conflicting error into `PushRejectMsg(wire.CmdTx, wire.RejectInvalid, ...)` back to the peer — for a transaction this node validated, persisted, spent, published to Kafka, and handed to block assembly. `services/legacy/netsync/handle_block.go:1373` treats a non-retryable error as a hard block failure. And a caller that *does* retry cannot re-unlock: the retry hits the `ErrTxExists` short-circuit at `Validator.go:825-835`, which returns before the `defer` at `:957` is even registered. Separately, `docs/topics/features/two_phase_commit.md:112` and `:180` state the opposite design intent — unlock failures are "logged but not considered fatal", with `SetMinedMulti` as the fallback.

**Why it matters:** The behaviour itself is inherited from main, so this isn't a regression. But this PR's whole value is that it writes the invariants down, and this particular sentence documents a contract that does not exist and conflicts with the file it cites two paragraphs earlier. A reader trusting it will assume there is a recovery handshake on the success-path unlock failure; there is only "wait until it's mined".

**Fix:** Either reword to match reality — the unlock error is surfaced for observability, and the real recovery is `SetMinedMulti` / `loadUnminedTransactions`, not caller action — or, if surfacing it as a returned error really is intended, say so and reconcile `two_phase_commit.md:180`. Worth noting in the body that the peer-reject consequence is pre-existing and out of scope.

*Criticality: 35%*

### ChiR4 — The "self-healing, in the right order" claim has an exception under default settings

**Problem:** The claim at `Validator.go:1037-1040` is that `loadUnminedTransactions` adds the tx to the subtree processor and only then clears `Locked`. That ordering does hold for the transactions that survive the load, but `lockedTransactions` is accumulated in the worker loop at `BlockAssembler.go:2765-2769`, *before* `validateParentChain` filters `unminedTransactions` at `:2880`. The unconditional `SetLocked(ctx, lockedTransactions, false)` at `:2964-2965` then unlocks the whole accumulated set, including entries that were filtered out and never reached `AddNodesDirectly`/`AddDirectly` (`:2921`, `:2941`). Both gates default to true (`settings/settings.go:300` and `:302`).

**Why it matters:** For a transaction stuck by the very failure this comment is about, the documented healing step can unlock it without adding it — landing it back in exactly the "unlocked but undelivered" state the comment says the design prevents. It's pre-existing blockassembly behaviour, not something this PR introduces, but the comment presents the invariant as unconditional.

**Fix:** Qualify the comment (note the `OnRestartValidateParentChain` / `OnRestartRemoveInvalidParentChainTxs` exception), or file a follow-up to collect `lockedTransactions` after the filter so the unlock set is exactly the added set.

*Criticality: 30%*

### ChiR5 — The deferred call snapshots `txMetaData`, which can drift from what the function returns

**Problem:** `defer v.unlockLockedTxOnExit(..., txMetaData, &err)` at `Validator.go:957` evaluates `txMetaData` at registration time, but `Validator.go:962-965` reassigns it on the `SkipUtxoCreation` path. From then on the helper is inspecting and mutating a different object from the one the caller receives — including its `txMetaData.Locked = false` write at `:1070`. It's benign today: `SpendOnly` returns a nil meta (`stores/utxo/spend_and_create.go:71-73`), so the helper's nil guard fires, and `WithLocked(true)` is never combined with `SkipUtxoCreation` (`Validator.go:1604-1609`). The comment at `:955-956` justifies the snapshot for the `return nil, err` case, which is a real reason — but it doesn't mention the reassignment, and the invariant that keeps it safe lives in another function.

**Why it matters:** Two independent facts in two files are what stop this from being a bug, and neither is asserted anywhere. If `SpendAndCreate` ever returns a populated meta under `SpendOnly`, or the `if/else` at `:1604` becomes an `if`, this silently unlocks or silently fails to unlock, with no test catching it.

**Fix:** The `persistedMeta := txMetaData` pin from ChiR1 makes the snapshot explicit and named, and lets you say in one line why it's the right object. Alternatively assert it: `if validationOptions.SkipUtxoCreation && txMetaData != nil && txMetaData.Locked { ... }` is a state that should never exist.

*Criticality: 25%*

### ChiR6 — The new tests don't cover the one genuinely subtle part of the refactor

**Problem:** All four new tests call `v.unlockLockedTxOnExit(...)` directly with a local `&deferredErr`. The mechanism that makes the whole design work — `err` being a *named return* at `Validator.go:661`, so a deferred write lands in the value the caller receives — is never exercised. `TestValidator_UnlockLockedTxOnExit_SetsErrOnUnlockFailure` would pass unchanged if someone converted the signature to unnamed returns, at which point the success-path unlock failure would be silently swallowed by `return txMetaData, nil`.

**Why it matters:** The pre-existing end-to-end tests do pin the two headline behaviours (`TestValidate_BlockAssemblyError` for stays-locked, `TestValidator_LockedFlagChangedIfBlockAssemblyStoreSucceeds` for unlock-on-success), so coverage isn't bad — but the one thing this refactor newly depends on is the one thing untested.

**Fix:** Add a case that drives `ValidateWithOptions` to success with a store whose `SetLocked` fails (the existing `FailingUtxoStore` plus a `blockassembly.NewMock()` returning success, i.e. `TestValidator_LockedFlagChangedIfBlockAssemblyStoreSucceeds` with the failing store swapped in) and asserts the returned error is the unlock error. Two nits while you're in there: `TestValidator_UnlockLockedTxOnExit_NoopWhenNotLocked` uses `assert.NoError` where `AGENTS.md` calls for `require`, and its `utxoStore.Create(...)` setup is dead — the assertion is driven entirely by the `&meta.Data{Locked: false}` literal, and the store record is never read.

*Criticality: 15%*

## Recap

| ChiR1 | Panic unlocks undelivered tx | YES | 60% |
| ChiR2 | Title/body contradict diff | YES | 55% |
| ChiR3 | Documented caller contract absent | | 35% |
| ChiR4 | Self-healing claim has exception | | 30% |
| ChiR5 | Stale txMetaData snapshot | | 25% |
| ChiR6 | Named-return interplay untested | | 15% |

icellan:
Pushed a follow-up commit addressing the panic-safety gap: `unlockLockedTxOnExit` was gating the unlock decision on `*err == nil`, but `err` is nil for the entire window between entering the add-to-block-assembly code and `sendToBlockAssembler` returning. A panic in that window (nil/misconfigured block-assembly client, etc.) would have unlocked a tx that was never actually delivered.

Fixed by adding an explicit `delivered *bool`, set to `true` only after `sendToBlockAssembler` returns successfully (or immediately for txs that skip block assembly entirely). The unlock decision now gates on `delivered`, not on the absence of an error. Also renamed the captured meta pointer to `persistedMeta` so a later reassignment of `txMetaData` on the `SkipUtxoCreation` path can't change what the deferred call sees.

Added `TestValidate_PanicDuringBlockAssemblyDeliveryKeepsLock`, which forces a panic inside `blockAssembler.Store` and asserts the tx stays `Locked` in the UTXO store afterward.

Title/description were already retitled to drop the "stuck transaction bug fix" framing per @ordishs — this is a panic-safety hardening of the deferred-unlock consolidation, not a fix for a bug that existed on `main`.

icellan:
Retitled and rewrote the description (ChiR2 / ordishs) to describe what the diff actually does: a refactor consolidating the 2PC unlock into one deferred exit path, plus the invariant documentation and tests. Dropped the "stuck transaction" bug-fix narrative — confirmed it never applied on `main`.

Fixed ChiR1 (panic during block-assembly delivery unlocks an undelivered tx): `unlockLockedTxOnExit` no longer infers "delivered" from `*err == nil`. There's now an explicit `delivered bool`, set `true` only after `sendToBlockAssembler` actually returns successfully (`false` immediately when a tx skips block assembly entirely), and the helper checks `delivered` before it looks at `*err`. This is panic-safe: a panic mid-window leaves `delivered` false, so the tx stays locked. Also pinned `persistedMeta := txMetaData` at defer-registration time per your note, so the later `SkipUtxoCreation` reassignment can't change what the deferred call operates on.

Added a test that forces a panic inside the add-to-block-assembly window (mock block-assembly client panics in `Store()`) and asserts the tx is still `Locked` afterward — this fails against the pre-fix code (defer would unlock on `err == nil`) and passes now.

Addressed ChiR6: added an end-to-end test driving `ValidateWithOptions` with a failing-unlock store, asserting the *returned* error reflects the unlock failure, so the named-return/defer plumbing itself is exercised, not just the helper in isolation. Also swapped `assert.NoError` for `require.NoError` and dropped the dead `utxoStore.Create` call in the no-op test.

Addressed ChiR3 and ChiR4 by rewording the doc comment: it no longer claims a caller contract that doesn't exist (the real recovery path is `loadUnminedTransactions`/`SetMinedMulti`, this matches `two_phase_commit.md`), and it now notes the `OnRestartValidateParentChain`/`OnRestartRemoveInvalidParentChainTxs` exception to the self-healing ordering guarantee (pre-existing block-assembly behavior, not introduced here).

Skipped ChiR5 as a separate item — the `persistedMeta` pin from the ChiR1 fix already resolves it (the snapshot is now explicit and named at the point it matters).

`go test ./services/validator/... -race` and `go vet ./services/validator/...` are green (one pre-existing test, `TestExtendedTxa1f6...`, fails in this sandbox for an unrelated reason — no Docker/Aerospike testcontainer available locally — verified it fails identically on `main` and on the pre-fix commit, unrelated to this change).

ctnguyen (approved):
Review at commit [3d49efe](https://github.com/icellan/teranode/commit/3d49efe9afad72617a0d584e4ea177da18cc70bd)

Good turnaround on this round. The `delivered` flag is the right fix for the panic window — it states the fact the decision actually depends on instead of a proxy for it, and checking it *before* `*err` means a future edit that adds an early success return before block assembly can't silently unlock. `persistedMeta` being named and pinned at defer-registration time reads much better than the implicit snapshot, and the panic test earns its keep: forcing the mock's `Store()` to panic is exactly the shape of failure the old `*err == nil` gate got wrong. The `ValidateWithOptions` end-to-end case now covers the named-return plumbing, the doc comment carries the `OnRestartValidateParentChain` caveat, and the title/body finally describe the diff. Both blocking points are closed. What's left is polish on the documentation, which — given that documenting the invariant is most of this PR's value — is worth getting exactly right.

## Status of previous points

ChiR3 is the only one still open, and it is still non-blocking. The false caller-contract sentence is gone, but the wording that replaced it at `Validator.go:1065-1070` — "No caller currently acts on that signal specifically", "mainly for observability" — is also not what happens: the unlock error is a `ProcessingError`, so it is neither `ErrTxMissingParent` nor `ErrTxLocked`, and `services/legacy/netsync/manager.go:1254-1291` therefore takes the `else` branch, records the hash in `rejectedTxns`, logs `Errorf` at `:1284`, sends `PushRejectMsg(wire.CmdTx, wire.RejectInvalid, "rejected", ...)` at `:1288`, and returns before `AnnounceNewTransactions` at `:1304-1306` — so the transaction is rejected to the peer and never relayed, for a tx this node validated, persisted, spent, and successfully delivered to block assembly. Meanwhile `docs/topics/features/two_phase_commit.md:112` and `:180` still say unlock failures are "logged but not considered fatal", and the new `TestValidator_ValidateWithOptions_ReturnsUnlockFailureError` (`Validator_test.go:1741-1742`) now pins the error-returning side of that disagreement without reconciling it. To close ChiR3, pick a side: either stop returning the unlock error (log it, bump a counter, matching the doc), or keep returning it and replace the "observability" sentence with the actual consequence plus a note reconciling `two_phase_commit.md:180`. The behaviour itself is inherited from `main`, so nothing regresses.

| ChiR1 | Panic unlocks undelivered tx | YES | FIXED |
| ChiR2 | Title/body contradict diff | YES | FIXED |
| ChiR3 | Documented caller contract absent | | PARTIAL FIXED |
| ChiR4 | Self-healing claim has exception | | FIXED |
| ChiR5 | Stale txMetaData snapshot | | FIXED |
| ChiR6 | Named-return interplay untested | | FIXED |

## Non-blocking issues

### ChiR7 — The doc comment's headline condition is no longer what the code does

**Problem:** `Validator.go:1042` opens with "It releases the lock ONLY when validateInternal is returning success." The implemented gate is `delivered && *err == nil`, which is not the same thing: after `delivered = true` at `Validator.go:1008`, a panic unwinding through the txmeta-Kafka block at `Validator.go:1017-1028` reaches the deferred helper with `delivered` true and `*err` nil, so it performs the `SetLocked(false)` store write while `validateInternal` is *not* returning at all. That is a genuine behaviour change from `main`, where the unlock was a straight-line statement a panic simply skipped. The paragraph at `:1072-1078` explains the `delivered` gate correctly, so the comment contradicts itself — the headline sentence is the one a reader stops at.

**Why it matters:** Unlocking there is the right call and consistent with the intent already stated at `Validator.go:1012-1016` (a missing txmeta message is recoverable, a stuck lock is not), so this is not a hazard. But it is the one behaviour the refactor newly introduces relative to `main`, it has no test, and the comment that exists specifically to pin the invariant describes a different rule. The next person reasoning about panic safety here will get the wrong answer from the first sentence.

**Fix:** State the actual condition and the deliberate asymmetry.

```go
// It releases the lock only once the tx has actually reached block assembly and no
// error was recorded (delivered && *err == nil) - not "only on a successful return".
// The difference is panics: unwinding from before delivery keeps the lock (the hazard
// this function exists to prevent), unwinding from after delivery still releases it,
// because the tx is in block assembly and a missing txmeta message is recoverable
// while a stuck lock is not (see the txmeta-Kafka comment in validateInternal).
```

A companion to `TestValidate_PanicDuringBlockAssemblyDeliveryKeepsLock` that panics *after* delivery (e.g. a `sendTxMetaToKafka` seam) and asserts the tx ends up unlocked would pin both halves.

*Criticality: 25%*

### ChiR8 — Two guards in the helper can't be reached, and one of them argues against the invariant above it

**Problem:** `Validator.go:1088` tests `delivered == nil`, but the only registration site passes `&delivered` (`Validator.go:969`) and all five tests pass the address of a local, so it is never nil. More interesting is `Validator.go:1094-1101`: the comment concedes "This should not happen given the current control flow", and that holds — after `delivered = true` at `:1008` the only remaining code is the txmeta block, which routes its failure into `txMetaErr` and never touches `err`, so `*err` is nil at `return txMetaData, nil`. If a future edit *did* make it reachable, its chosen behaviour ("keep it locked rather than trust an inconsistent state") is the opposite of what the invariant 50 lines above prescribes: the tx has reached block assembly, so `:1042-1049`'s own reasoning says releasing the lock is safe, and withholding it only blocks child spends until `loadUnminedTransactions` heals it.

**Why it matters:** Dead defensive code in a function whose whole purpose is to encode one decision rule makes the rule harder to read, and a dead branch that would resolve the wrong way is worse than no branch — the next person to add an error-setting path after delivery inherits a silent, unlockable stall rather than a compile error or a clear decision.

**Fix:** Drop the nil check by capturing `delivered` in a closure instead of threading a pointer, and either delete the `*err != nil` branch or make it unlock-with-a-warning so the code matches the documented invariant.

```go
defer func() { v.unlockLockedTxOnExit(decoupledCtx, tx, txID, persistedMeta, &err, delivered) }()
...
// in the helper, after the !delivered check:
if *err != nil {
    // Delivered, so releasing the lock is safe per the invariant above; the error
    // is unrelated to delivery. Log it so an unexpected combination is visible.
    v.logger.Warnf("[Validate][%s] unlocking a delivered tx despite a later error: %v", txID, *err)
}
```

*Criticality: 20%*

### ChiR9 — The not-delivered warning is a duplicate on the two error paths but the only record on the panic path, so it needs to distinguish them

**Problem:** `Validator.go:1089` emits one `Warnf` per transaction on every exit where the tx is Locked and undelivered, and that covers two materially different situations. On the two ordinary error returns — `subtree.NewTxInpointsFromTx` at `Validator.go:989-992` and `sendToBlockAssembler` at `:996-1006` — an error is returned, `span.RecordError` captures it, and callers log it again: `services/legacy/netsync/manager.go:1284` logs `Errorf` for the same transaction. There the new line is additive. On the panic-before-delivery path the same line is the *only* per-transaction record: propagation's recover at `services/propagation/Server.go:795-800` logs "Recovered from panic in processTransactionInternal" without the tx hash, and netsync's only `recover()` (`manager.go:2946-2956`) is an unrelated shutdown channel guard, so a panic from that caller is not recovered there at all. `main` logged nothing on any of these paths.

**Why it matters:** During a block-assembly outage the returned-error case fires per transaction at ingest rate on the validator hot path, roughly doubling an already-noisy failure that a counter would express better. But a blanket demote to `Debugf` would silence the panic case, which is precisely the one where nothing else names the stuck transaction — so the obvious tidy-up is the wrong one here.

**Fix:** Keep a warn where it is the only record and demote where it is not, and add a counter so the aggregate stays visible either way.

```go
prometheusValidatorTxStaysLockedUndelivered.Inc()

if *err != nil {
    // The returned error is logged by the caller too - keep this at debug.
    v.logger.Debugf("[Validate][%s] tx stays locked, it was not delivered to block assembly: %v", txID, *err)
} else {
    // No error recorded means we are unwinding from a panic; nothing downstream
    // names this tx, so this line has to carry it.
    v.logger.Warnf("[Validate][%s] tx stays locked, panic unwound before delivery to block assembly", txID)
}
```

*Criticality: 10%*

## Recap

| ChiR7 | Headline condition mis-stated | | 25% |
| ChiR8 | Unreachable guards, wrong default | | 20% |
| ChiR9 | Warning conflates two paths | | 10% |

ordishs (approved):

PR #1571, opened 2026-08-10
fix(daemon): remove stale -faucet help text

icellan:
The in-repo daemon's help text advertised a `-faucet=<1|0>` flag, but the in-repo daemon starts no faucet service — that flag belongs to the external coinbase binary's own CLI. Confirmed no corresponding flag is parsed anywhere in the daemon package; purely stale help text, removed.


ordishs (approved):
Verified. `daemon/daemon_services.go` calls `shouldStart` for 15 services, and `faucet` is not one of them, so the removed help text described a flag that starts nothing. A grep over all Go files finds no `startFaucet` read and no faucet service package. The only `-faucet=1` uses are `./coinbase.run` invocations in compose files, which target a different binary.

Two nits:

- `daemon/daemon_test.go:202` — `TestPrintUsage` asserts only on three lines, so nothing stops the removed line returning.
- `daemon/daemon.go:284` — `printUsage()` is incomplete in the opposite direction. `shouldStart("wait_for_postgres", args)` makes `-wait_for_postgres=1` a real flag, and `serviceHelp` at `daemon.go:94` maps it, but the usage text omits it.

freemans13 (approved):
Approving. The change is correct and the test that comes with it is a genuinely good one.

I verified the central claim rather than taking it on trust. Grepping the daemon package, `shouldStart` is called 16 times — once for `help`, once each for the 14 services in `daemon_services.go`, and once for `wait_for_postgres` at `daemon/daemon.go:284`. There is no faucet call site anywhere, so the removed help text described a flag that starts nothing. Counting `-all=0`, which `shouldStart` matches inline rather than through a call, that is 17 honoured flags against 17 documented ones, which is exactly why the new test passes.

The second commit is the more interesting half, so I mutation-tested the new guard rather than assuming it works. All three drift directions fail as intended, each with a message naming the flag and the direction:

| Mutation to `printUsage()` | Message |
|---|---|
| Delete the `-pruner` entry | `shouldStart honours -pruner but printUsage does not document it` |
| Add a fabricated `-bogus` entry | `printUsage documents -bogus but no shouldStart call honours it` |
| Re-add the `-faucet` entry | fails both tests, on `should not contain "faucet"` and on the drift check |

That is a real regression guard, not decoration. The AST walk also fails closed in the places that matter: an argument it cannot resolve to a string constant calls `t.Fatalf` rather than being silently skipped, so if someone starts computing a service name the test complains instead of quietly under-collecting. Scoping the parse to the package directory is right, since `shouldStart` is unexported and cannot be called from outside it.

Both tests pass on Go 1.26.5 and under the conditions `make test` actually uses (`-race -tags testtxmetacache`, `SETTINGS_CONTEXT=test`). `gofmt` is clean. `golangci-lint run ./daemon/...` reports one `prealloc` warning in `daemon/test_daemon.go:1294`, which is pre-existing in a file this PR does not touch.

One thing worth fixing before merge, non-blocking: **the PR description is now stale.** It describes only the faucet removal and says nothing about the second commit, which adds `-help=1` and `-wait_for_postgres=1` to the usage text plus around 150 lines of test. Since the description becomes the merge commit body, a reader of `git log` would not learn that this PR changed the documented flag set in two places beyond the deletion. Two sentences would cover it.

Two nits I would not block on, both in test-only code and both inherited from the pre-existing `TestPrintUsage` pattern rather than introduced here. `documentedUsageFlags` swaps `os.Stdout` at `daemon/daemon_test.go:236-246` and restores it only on the success path — if the `require.NoError(t, w.Close())` were ever to fail, `t.FailNow` calls `runtime.Goexit` and stdout stays pointed at a closed pipe for the rest of the package's test binary. A `defer func() { os.Stdout = oldStdout }()` right after the swap is the idiomatic fix, though `Close` on a fresh pipe realistically does not fail. The read end `r` is likewise never closed in either test. Separately, both tests write the whole usage text into a pipe with no reader draining it and only read after closing the writer; at roughly 1.2 KB against a 16 KB pipe buffer there is about 13x headroom, so no problem today, but it is worth knowing the ceiling exists given this test's whole purpose is to make people add entries to `printUsage`.

PR #1572, opened 2026-08-10
test(seeder): round-trip and complete-file coverage for UTXO snapshot import

icellan:
## Summary

Adds the two remaining tests from the original audit recommendation on the seeder's UTXO snapshot import path, alongside the truncated-file regression test in PR #1563:

- `TestSeederImport_RoundTripsRealSnapshotFile` - writes a `.utxo-set` file using the persister's own `UTXOWrapper` record encoding, feeds it through the seeder's import path (`readUTXOWrapperFile` -> `processUTXO`), and checks the actual record content (values, scripts, coinbase handling) survives into the UTXO store, not just that footer counts line up. Also asserts the surviving UTXO's actual on-disk index via `GetSpend`, not just its content, since the store's compacted read view can't otherwise distinguish a correctly-indexed UTXO from one landed at the wrong index by an import bug.
- `TestSeederImport_CompleteFile_FinishesCleanly` - feeds a well-formed, complete snapshot file through the same path and asserts it finishes without error, every record lands in the store, and the footer counts match what was actually processed.

Both share the existing `buildSnapshotWrappers` / `openForReading` / `newTestUtxoStore` helpers from `seeder_truncation_test.go` and `coinbase_restore_test.go` rather than adding new scaffolding.

This PR is now test-only. A real behavioral fix (headers-truncation detection) that was previously bundled into this branch has been split out into #1604 per review feedback, so it can be reviewed on its own terms.

## Dependency

This branches off `fix/audit-14-seeder-truncation-safety` (PR #1563), which introduces `readUTXOWrapperFile` and its footer validation. That PR is not yet merged, so this diff currently also contains its commit. It should be rebased onto `main` (and this PR's diff will shrink to just the new test file) once #1563 lands.

## Test plan

- [x] `go test ./cmd/seeder/... ./services/utxopersister/... -race`
- [x] `go vet ./cmd/seeder/... ./services/utxopersister/...`
- [x] `golangci-lint run cmd/seeder/...`

### DIFF
--- a/cmd/seeder/seeder.go
+++ b/cmd/seeder/seeder.go
@@ -512,54 +512,8 @@ func processUTXOs(ctx context.Context, logger ulogger.Logger, appSettings *setti
 		return nil, errors.NewProcessingError("couldn't read previous block hash from file", err)
 	}
 
-	var (
-		txProcessed    uint64
-		utxosProcessed uint64
-	)
-
 	g.Go(func() error {
-		defer close(utxoWrapperCh)
-
-	OUTER:
-		for {
-			select {
-			case <-gCtx.Done():
-				// Context canceled, stop reading lines
-				logger.Infof("Context cancelled, stopping reading UTXOWrapper")
-				return gCtx.Err()
-
-			default:
-				var utxoWrapper *utxopersister.UTXOWrapper
-
-				utxoWrapper, err = utxopersister.NewUTXOWrapperFromReader(gCtx, reader)
-				if err != nil {
-					if errors.Is(err, io.EOF) {
-						break OUTER
-					}
-
-					logger.Errorf("Failed to read UTXO: %v", err)
-					return errors.NewProcessingError("failed to read UTXO", err)
-				}
-
-				select {
-				case <-gCtx.Done():
-					logger.Infof("Context cancelled while sending UTXO to channel, stopping UTXO processing")
-					return gCtx.Err()
-
-				case utxoWrapperCh <- utxoWrapper:
-					txProcessed++
-					utxosProcessed += uint64(len(utxoWrapper.UTXOs))
-
-					if txProcessed%1_000_000 == 0 {
-						logger.Infof("Processed %16s transactions with %16s utxos", formatNumber(txProcessed), formatNumber(utxosProcessed))
-					}
-				}
-			}
-		}
-
-		logger.Infof("FINISHED  %16s transactions with %16s utxos", formatNumber(txProcessed), formatNumber(utxosProcessed))
-
-		return nil
+		return readUTXOWrapperFile(gCtx, logger, f, reader, utxoWrapperCh)
 	})
 
 	if err = g.Wait(); err != nil {
@@ -575,6 +529,88 @@ func processUTXOs(ctx context.Context, logger ulogger.Logger, appSettings *setti
 	}
 
 	return &utxoSetTip{hash: hash, height: height}, nil
+}
+
+// readUTXOWrapperFile reads UTXOWrapper records from reader (backed by f) and
+// sends them to utxoWrapperCh until the record stream is exhausted, then
+// closes the channel.
+//
+// The underlying UTXOWrapperFromReader/errors.Is(err, io.EOF) EOF detection
+// cannot tell a clean end-of-records boundary apart from a file truncated
+// mid-record: both surface as an error whose message contains "EOF" (a real
+// truncation trips io.ErrUnexpectedEOF, whose message "unexpected EOF"
+// contains "EOF" as a substring), so errors.Is treats them identically via
+// this package's substring-matching Is fallback. To make truncation
+// detection exact, every snapshot file carries a trailing 16-byte footer
+// (utxopersister.GetFooter) recording the txCount/utxoCount it was written
+// with; once the read loop ends for any reason, that footer is compared
+// against what was actually processed, and a mismatch is reported as an
+// error instead of silently treated as success.
+func readUTXOWrapperFile(ctx context.Context, logger ulogger.Logger, f *os.File, reader *bufio.Reader, utxoWrapperCh chan<- *utxopersister.UTXOWrapper) error {
+	defer close(utxoWrapperCh)
+
+	var (
+		txProcessed    uint64
+		utxosProcessed uint64
+		err            error
+	)
+
+OUTER:
+	for {
+		select {
+		case <-ctx.Done():
+			// Context canceled, stop reading lines
+			logger.Infof("Context cancelled, stopping reading UTXOWrapper")
+			return ctx.Err()
+
+		default:
+			var utxoWrapper *utxopersister.UTXOWrapper
+
+			utxoWrapper, err = utxopersister.NewUTXOWrapperFromReader(ctx, reader)
+			if err != nil {
+				if errors.Is(err, io.EOF) {
+					break OUTER
+				}
+
+				logger.Errorf("Failed to read UTXO: %v", err)
+				return errors.NewProcessingError("failed to read UTXO", err)
+			}
+
+			select {
+			case <-ctx.Done():
+				logger.Infof("Context cancelled while sending UTXO to channel, stopping UTXO processing")
+				return ctx.Err()
+
+			case utxoWrapperCh <- utxoWrapper:
+				txProcessed++
+				utxosProcessed += uint64(len(utxoWrapper.UTXOs))
+
+				if txProcessed%1_000_000 == 0 {
+					logger.Infof("Processed %16s transactions with %16s utxos", formatNumber(txProcessed), formatNumber(utxosProcessed))
+				}
+			}
+		}
+	}
+
+	// The read loop above only knows the record stream ended - it cannot
+	// distinguish a clean footer boundary from a truncated file (see the
+	// doc comment above). Validate against the file's own footer counts so
+	// a genuinely truncated snapshot is reported as an error rather than
+	// silently accepted.
+	expectedTxCount, expectedUTXOCount, footerErr := utxopersister.GetFooter(f)
+	if footerErr != nil {
+		return errors.NewProcessingError("failed to read snapshot footer", footerErr)
+	}
+
+	if expectedTxCount != txProcessed || expectedUTXOCount != utxosProcessed {
+		return errors.NewProcessingError(
+			"snapshot file truncated: expected %d transactions/%d UTXOs, processed %d/%d",
+			expectedTxCount, expectedUTXOCount, txProcessed, utxosProcessed)
+	}
+
+	logger.Infof("FINISHED  %16s transactions with %16s utxos", formatNumber(txProcessed), formatNumber(utxosProcessed))
+
+	return nil
 }
 
 // worker processes UTXOWrapper messages from the channel and stores them in the UTXO store.
--- a/cmd/seeder/seeder_roundtrip_test.go
+++ b/cmd/seeder/seeder_roundtrip_test.go
@@ -1,0 +1,217 @@
+package seeder
+
+import (
+	"context"
+	"encoding/binary"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/bsv-blockchain/go-bt/v2/chainhash"
+	"github.com/bsv-blockchain/teranode/pkg/fileformat"
+	"github.com/bsv-blockchain/teranode/services/utxopersister"
+	utxosql "github.com/bsv-blockchain/teranode/stores/utxo/sql"
+	"github.com/bsv-blockchain/teranode/ulogger"
+	"github.com/stretchr/testify/require"
+)
+
+// writeCompleteSnapshotFile assembles a well-formed .utxo-set file byte for
+// byte in the layout CreateUTXOSet writes (services/utxopersister/UTXOSet.go):
+// fileformat header, then blockHash(32)+height(4)+previousBlockHash(32), then
+// one UTXOWrapper.Bytes() record per wrapper, then the trailing 16-byte
+// footer (txCount + utxoCount, little-endian) that GetFooter reads back. It
+// reuses the same per-record encoding (UTXOWrapper.Bytes()) that the real
+// persister writes, so the file the seeder reads back is byte-identical to
+// what production would have produced for these records.
+func writeCompleteSnapshotFile(t *testing.T, wrappers []*utxopersister.UTXOWrapper) string {
+	t.Helper()
+
+	var blockHash chainhash.Hash
+	blockHash[0] = 0xcc
+
+	var prevHash chainhash.Hash
+	prevHash[0] = 0xdd
+
+	header := fileformat.NewHeader(fileformat.FileTypeUtxoSet)
+
+	data := make([]byte, 0)
+	data = append(data, header.Bytes()...)
+	data = append(data, blockHash[:]...)
+
+	heightBytes := make([]byte, 4)
+	binary.LittleEndian.PutUint32(heightBytes, 200)
+	data = append(data, heightBytes...)
+
+	data = append(data, prevHash[:]...)
+
+	var txCount, utxoCount uint64
+
+	for _, w := range wrappers {
+		data = append(data, w.Bytes()...)
+		txCount++
+		utxoCount += uint64(len(w.UTXOs))
+	}
+
+	footer := make([]byte, 16)
+	binary.LittleEndian.PutUint64(footer[0:8], txCount)
+	binary.LittleEndian.PutUint64(footer[8:16], utxoCount)
+	data = append(data, footer...)
+
+	path := filepath.Join(t.TempDir(), "roundtrip.utxo-set")
+	require.NoError(t, os.WriteFile(path, data, 0o600))
+
+	return path
+}
+
+// importSnapshotFile drives the file through the same functions Seeder's
+// processUTXOs uses: readUTXOWrapperFile feeds a channel that processUTXO
+// drains into store, exactly as worker() does, but without the extra worker
+// pool machinery that's orthogonal to file-format correctness.
+func importSnapshotFile(t *testing.T, path string, store *utxosql.Store) error {
+	t.Helper()
+
+	f, reader := openForReading(t, path)
+
+	utxoWrapperCh := make(chan *utxopersister.UTXOWrapper, 10)
+
+	readErrCh := make(chan error, 1)
+
+	go func() {
+		readErrCh <- readUTXOWrapperFile(context.Background(), ulogger.TestLogger{}, f, reader, utxoWrapperCh)
+	}()
+
+	var processErr error
+
+	for w := range utxoWrapperCh {
+		if err := processUTXO(context.Background(), store, w, nil); err != nil && processErr == nil {
+			processErr = err
+		}
+	}
+
+	readErr := <-readErrCh
+
+	if readErr != nil {
+		return readErr
+	}
+
+	return processErr
+}
+
+// TestSeederImport_RoundTripsRealSnapshotFile writes a real .utxo-set file
+// using the persister's own UTXOWrapper record encoding, feeds it through the
+// seeder's import path (readUTXOWrapperFile -> processUTXO), and checks the
+// exact record content - not just counts - survives into the UTXO store.
+// This is the round-trip half of the two tests the audit recommended
+// alongside the truncated-file regression test.
+func TestSeederImport_RoundTripsRealSnapshotFile(t *testing.T) {
+	ctx := context.Background()
+	store := newTestUtxoStore(ctx, t)
+
+	txA := chainhash.HashH([]byte("roundtrip-tx-a"))
+	txB := chainhash.HashH([]byte("roundtrip-tx-b"))
+
+	wrappers := []*utxopersister.UTXOWrapper{
+		{
+			TxID:   txA,
+			Height: 200,
+			UTXOs: []*utxopersister.UTXO{
+				{Index: 0, Value: 1234, Script: []byte{0x76, 0xa9, 0x14}},
+				{Index: 1, Value: 5678, Script: []byte{0x51}},
+			},
+		},
+		{
+			TxID:   txB,
+			Height: 200,
+			UTXOs: []*utxopersister.UTXO{
+				{Index: 2, Value: 999999, Script: []byte{0x6a, 0x01, 0x02, 0x03}},
+			},
+		},
+	}
+
+	path := writeCompleteSnapshotFile(t, wrappers)
+
+	require.NoError(t, importSnapshotFile(t, path, store))
+
+	gotA, err := store.Get(ctx, &txA)
+	require.NoError(t, err)
+	require.NotNil(t, gotA.Tx)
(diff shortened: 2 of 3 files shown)


ordishs (approved):
Verified the core claim. `errors.Is` falls back to `strings.Contains` on the rendered message when the target is not a `*Error`, so `io.ErrUnexpectedEOF` wrapped by `NewUTXOWrapperFromReader` is indistinguishable from a clean `io.EOF` under the old check. The new tests exercise the footer-based fix correctly.

Two nits:

- `cmd/seeder/seeder.go:328` and `:755` — the same `errors.Is(err, io.EOF)` substring hazard sits unpatched in the utxo-headers read loops, with no footer validation added there.
- The diff currently bundles the still-unmerged production fix from #1563, because `readUTXOWrapperFile` is not on main. The diff shrinks to the two new test files once #1563 lands.

freemans13:
## Review — safe to merge, two behavioural notes worth a look first

Reviewed at `3235e6b81`. No critical or high findings, and I couldn't break the mechanism. Nice approach: rather than untangle the sentinel plumbing, you compare against a count a truncation can't forge — the existing 16-byte footer for `.utxo-set`, and the preamble tip height for `.utxo-headers`, which sits before any truncation point. That's the right minimal-diff call.

I verified the two load-bearing assumptions rather than taking them on trust:

- `WriteHeadersToStore` (`services/utxopersister/UTXOSet.go:867`) walks tip→genesis one block per height into a hash-keyed map and writes every entry, so `recordCount == tipHeight+1` genuinely holds. Same shape in `cmd/bitcointoutxoset/bitcoin/headers.go:150`.
- `GetFooter`'s seek-to-end-minus-16 matches both writers, and `FileStorer.Close` appends no trailer after the counts.
- Empirically, against the real committed testnet seed `seeds/tstn/000000000c130c0b….*`: headers file has 301 records for tip height 300, contiguous 0..300, all bytes consumed; set file counts 300 tx / 300 UTXOs against a footer of 300/300. Neither new check false-rejects real producer output.

I could not construct a truncated file that still passes, nor a valid complete file that gets falsely rejected.

---

### Two worth discussing before merge

**1. Truncated headers now restore *zero* coinbases instead of a partial set** — `cmd/seeder/seeder.go:461-466`

`loadCoinbaseTxs` now returns an error where it used to return a partial map, but the caller treats it as best-effort: `Warnf` plus an **empty** map. Net effect is strictly worse than before. With a headers file truncated at 90%, previously ~90% of coinbases were stored faithfully with their real input and txid; now none are, and the only signal is one warning line. The hard check in `processHeaders` never runs on the `-skipHeaders` path, so nothing is fatal there at all.

The best-effort downgrade is deliberate and pre-dates this PR (the comment at `:459` explains it) — it just interacts badly with the new error return. Is a truncation error specifically one you'd want to be fatal here?

**2. Truncated headers + intact utxo-set can strand the node without a checkpoint** — `cmd/seeder/seeder.go:238`, `:254`, `:424-429`

The passes run concurrently. UTXO pass succeeds and writes `lastProcessed.dat`; header pass now fails, so `Seeder` returns at `:238` before `writeBlockAssemblerState`. Operator repairs the headers file and re-runs — `processUTXOs` sees `lastProcessed.dat`, returns `(nil, nil)` at `:429`, so `utxoTip` is nil and the `!skipUTXOs && utxoTip != nil` guard at `:254` skips the checkpoint write *again*. Node starts with no checkpoint; only `-force` (full UTXO re-import) recovers.

The trap is structural and pre-existing — the comment at `:251-253` shows it's known for the `writeBlockAssemblerState` failure case. What's new is a much more likely trigger. At minimum worth a line in the PR description.

### Lower priority

3. `recordsRead == tipHeight+1` turns an undocumented producer property into a hard gate (`seeder.go:378`, `:840`). It holds in both producers today and on the real seed — but a future partial-range headers export would hard-fail *every* import with "truncated". A contract comment on the writer side would cost nothing.
4. `g.Wait()` wraps the reader error as `NewProcessingError("error in worker", err)` (`seeder.go:537`), so the headline message on the exact failure this PR exists to surface reads `error in worker -> snapshot file truncated: …`.
5. The three truncation tests assert only `require.Error`, never the reason, so they'd stay green if the error came from a different branch (a `GetFooter` seek failure, or `StoreBlock` erroring in `processHeaders`). They're not vacuous — all three do fail if the new checks are removed — just imprecise. `require.ErrorContains(t, err, "truncated")` would pin the branch.

### Pre-existing, in code this PR touches — follow-ups, not objections

6. `services/utxopersister/Headers.go:185` — `make([]byte, coinbaseLen)` allocates straight from an untrusted uint32 before `io.ReadFull` can fail; a corrupt length of `0xFFFFFFFF` forces a 4 GB allocation. This is the exact pattern deliberately avoided in `UTXO.go:352-363` via `io.CopyN` into a `bytes.Buffer`, and it sits in the read path this PR is hardening.
7. `cmd/seeder/seeder.go:510` — `errors.NewProcessingError("invalid file type: %s", header.FileType)` passes the method value, not the call, so it renders as `%!s(func() fileformat.FileType=0x…)`. `go vet` doesn't catch it. Lines 291 and 788 get it right.
8. `cmd/seeder/seeder.go:527` — `reader.Read(previousBlockHash[:])` is a bare `Read`; a short read misaligns every following record by up to 31 bytes (`UTXOSet.go:404` switched to `io.ReadFull` for exactly this). The new footer check actually mitigates it, but the failure would be misreported as truncation.

### Coverage gaps worth considering

- No test drives **real producer output** through the seeder's readers — both new tests hand-roll the byte layout, so a framing change in `CreateUTXOSet`/`WriteHeadersToStore` breaks production while these stay green.
- **Truncation inside the 16-byte footer** (8 of 16 bytes present) — the most realistic real-world case (interrupted write/`scp`), and the one where `GetFooter` reads record payload as half a count.
- Zero-record complete utxo-set (footer 0/0); file shorter than 16 bytes (`Seek(-16, io.SeekEnd)` → EINVAL, reported as "failed to read snapshot footer" rather than truncation); `-skipHeaders` with a truncated headers file; a headers file with *surplus* records.
- Free win: `TestLoadCoinbaseTxs_RealSeedFile` points at seed hash `000000007816d70a…`, which isn't in the repo, so it always skips — while a real V2 seed *is* committed at `seeds/tstn/000000000c130c0b…`. Repointing it gives CI real-file coverage of the new invariant for nothing.

### Clean

EOF handling is correctly diagnosed rather than papered over (confirmed at `errors/errors.go:1126` that `errors.Is(err, io.EOF)` degenerates to a substring check, exactly as the comments claim). Resource cleanup is clean — `readUTXOWrapperFile` doesn't own `f`, and `defer close(utxoWrapperCh)` fires on every return path including the new footer-error one, so workers can't deadlock. Concurrency is clean: the closure→function extraction is behaviour-preserving, and `GetFooter`'s seek only moves the offset after the read loop is done with `reader`. Bounds/overflow in the new code, nil guards, Go idioms, godoc accuracy and scope discipline all fine.

### Verification

`go build`, `go vet`, `go test ./cmd/seeder/... -race -count=1` (`ok … 2.391s`, all 7 new tests), `golangci-lint run cmd/seeder/...` (0 issues) — all pass locally, alongside the green CI here.

**Limits:** the heavy suites (benchmarks, sequential, smoketest) weren't re-run locally — relying on CI for those. The empirical file-format check used the single committed testnet seed, a 300-block toy rather than a mainnet-scale snapshot; the invariants are structural so scale shouldn't matter, but that's reasoning rather than measurement. Findings 6–8 are pre-existing and not objections to merging.

freemans13:
Addendum to my review above — one more coverage gap, this one *inside* the round-trip test rather than beside it.

**The round-trip test can't distinguish vout 2 from vout 0** — `cmd/seeder/seeder_roundtrip_test.go:151`

The interesting case the test sets up is `txB`: a single UTXO at index 2, so outputs 0 and 1 are already spent and `PadUTXOsWithNil` hands `processUTXO` `[nil, nil, output]`. That's the one place a snapshot import can go quietly wrong — a UTXO landing at vout 0 instead of vout 2 gets its hash computed over the wrong index and is permanently unspendable.

The seeder gets it right: `createOutputsBatched` (`stores/utxo/sql/sql.go:696-731`) preserves `entry.index` and derives `utxoHash` from it. But the test can't tell, because it asserts:

```go
require.Len(t, gotB.Tx.Outputs, 1)
require.Equal(t, uint64(999999), gotB.Tx.Outputs[0].Satoshis)
```

and the SQL store's `Get` rebuilds outputs with `SELECT locking_script, satoshis FROM outputs WHERE transaction_id = $1 ORDER BY idx` (`sql.go:1636`) — no `idx` in the projection and no re-padding, so the returned slice is a compacted view. Those assertions pass byte-for-byte identically if the UTXO had been written at index 0. The test walks up to the failure mode worth catching and then looks away.

Suggested fix: assert on something that carries the index — recompute `util.UTXOHashFromOutput(&txB, output, 2)` and check the store holds it, or attempt a spend of outpoint `txB:2`. One extra assertion turns "content survived" into "content *and index* survived", which is what the audit item was after.

Same shape as finding 5 above (the truncation tests asserting only `require.Error` and never the reason): the assertion is weaker than the thing the test exists to prove. Non-blocking either way.

Minor, same file: the comment on line 138 reads "output at index 1 must be present without a leading nil hole", but `txA` has outputs at 0 and 1 — there's no hole in that case.

freemans13 (changes requested):
Converting my earlier notes into a formal review, because on a closer look at what's actually on this branch I think three of them should be settled before merge rather than after. Everything else from those two comments stands as written and is non-blocking — I'm not restating it here.

First, a correction to my own earlier framing. I had been reading `seeder.go` on this branch as inherited from #1563. It isn't. This branch is `c7ce2e252` (shared with #1563) + `94bd4502b` (the round-trip test) + `3235e6b81` (the headers truncation fix). That last commit is this PR's alone — #1563 doesn't carry it, and #1563 has since moved on to `ecba9e93a` (the consolidator laundering gap), which this branch doesn't have. So the two branches have drifted apart in both directions, and the behavioural change is this PR's own.

---

### 1. The coinbase best-effort path is now strictly worse than before — `cmd/seeder/seeder.go:461-466`, `:793+`

`loadCoinbaseTxs` now returns `nil, err` when the record count doesn't match the tip height. The caller catches that, logs a warning, and substitutes an **empty** map:

```go
coinbaseTxs, err := loadCoinbaseTxs(logger, headersFile)
if err != nil {
    logger.Warnf("[processUTXOs] could not load coinbase transactions from %s; ...: %v", headersFile, err)
    coinbaseTxs = map[chainhash.Hash]*bt.Tx{}
}
```

Before `3235e6b81`, the loop broke on EOF and returned whatever it had accumulated. So on a headers file truncated at 90%, the old code stored ~90% of coinbases with their real inputs and txids; the new code stores none of them, and the only signal is a single warning line. The import still "succeeds".

The commit's goal — stop treating truncation as a clean end-of-records — is right, and I'm not arguing with it. It's the interaction with the pre-existing best-effort caller that inverts the outcome. Detection got better and the result got worse.

Either of these resolves it:

- return the partial map alongside the error (`return coinbaseTxs, err`) so the caller keeps today's degraded-but-partial behaviour while the warning tells the operator the file is damaged; or
- make truncation specifically fatal at the caller, distinguishing it from the absent/legacy/unreadable cases the best-effort comment at `:457-460` was actually written for.

The first is the smaller diff and preserves intent. The second is defensible if you'd rather a damaged headers file never seed at all — but note the hard check in `processHeaders` doesn't run on the `-skipHeaders` path, so nothing is fatal there today.

### 2. The round-trip test can't distinguish vout 2 from vout 0 — `cmd/seeder/seeder_roundtrip_test.go:151`

This is the PR's headline deliverable, so I think it's worth getting right rather than filing as a follow-up.

The interesting case the test constructs is `txB`: a single UTXO at index 2, so outputs 0 and 1 are spent and `PadUTXOsWithNil` hands `processUTXO` a `[nil, nil, output]`. That's precisely where a snapshot import can go quietly wrong — a UTXO landing at vout 0 instead of vout 2 gets its hash computed over the wrong index and is permanently unspendable.

The seeder gets it right: `createOutputsBatched` (`stores/utxo/sql/sql.go:696-731`) preserves `entry.index` and derives `utxoHash` from it. But the test can't tell, because the SQL store's `Get` rebuilds outputs with `SELECT locking_script, satoshis FROM outputs WHERE transaction_id = $1 ORDER BY idx` (`sql.go:1636`) — no `idx` in the projection, no re-padding. The returned slice is a compacted view, so:

```go
require.Len(t, gotB.Tx.Outputs, 1)
require.Equal(t, uint64(999999), gotB.Tx.Outputs[0].Satoshis)
```

passes byte-for-byte identically if the UTXO had been written at index 0. The test walks right up to the failure mode it exists to catch and then looks away.

One assertion fixes it — recompute `util.UTXOHashFromOutput(&txB, output, 2)` and check the store holds it, or attempt a spend of outpoint `txB:2`. That turns "content survived" into "content *and index* survived", which is what the audit item was after.

(Minor, same file: the comment on line 138 reads "output at index 1 must be present without a leading nil hole", but `txA` has outputs at both 0 and 1 — there's no hole in that case.)

### 3. `3235e6b81` isn't in the PR description

The title is `test(seeder):` and the description says this adds two tests and that "this PR's diff will shrink to just the new test file once #1563 lands." But `3235e6b81` is 51 lines of behavioural change to `seeder.go` that belongs to this PR and won't disappear on a rebase. A reviewer working from the description wouldn't know to look at it — and the approval here predates that commit by a day.

Splitting it into its own PR is probably cleanest, since it's a separate audit item from the tests and would let it be reviewed on its own terms. Failing that, the description and title need to say it's there.

---

### Not blocking

Everything else in my two earlier comments — the stranded-checkpoint interaction at `:238`/`:254`/`:424-429`, the `recordsRead == tipHeight+1` gate hardening an undocumented producer property, the `error in worker` wrapping burying the headline message, the truncation tests asserting only `require.Error`, the three pre-existing bugs in touched code (`Headers.go:185` unbounded `make` from an untrusted uint32, `seeder.go:510` `%s` on a method value, `seeder.go:527` bare `Read`), and the coverage gaps — all still stand, none of them justify holding the merge. Happy to raise the pre-existing three as separate issues if that's useful.

### What I verified

I re-read both paths rather than relying on my earlier notes. Findings 1 and 2 are confirmed against the code as it sits at `3235e6b81`. The mechanism in `3235e6b81` itself is sound — I couldn't construct a truncated headers file that passes the new check, nor a legitimate complete file that gets falsely rejected, and the real committed testnet seed (`seeds/tstn/000000000c130c0b…`, 301 records for tip height 300) passes cleanly. Local `go build`, `go vet`, `go test ./cmd/seeder/... -race -count=1` and `golangci-lint run cmd/seeder/...` were all green when I ran them.

**Limits:** that empirical check used a 300-block toy seed, not a mainnet-scale snapshot — the invariants are structural so scale shouldn't matter, but that's reasoning rather than measurement. I haven't re-run anything against current `main`, which has moved since. And the heavy suites (benchmarks, sequential, smoketest) I'm relying on CI for.

icellan:
Addressed the three blocking findings:

1. **Coinbase best-effort regression** — fixed `loadCoinbaseTxs` to return the partial map alongside the error on truncation, so the caller keeps today's degraded-but-partial behavior (warning fires, but data isn't discarded).
2. **Round-trip test doesn't distinguish vout index** — strengthened the txB assertion to verify the UTXO's computed hash reflects its actual index (2), not just that content survived — this now fails if a UTXO lands at the wrong index.
3. **Behavioral commit not described / should be split** — split the headers-truncation-detection fix (previously bundled here) into its own PR: https://github.com/bsv-blockchain/teranode/pull/1604. This PR (#1572) now contains only the two test files (round-trip + complete-file), as originally scoped.

Also fixed the minor doc-comment inaccuracy on txA's output-hole description.

icellan:
Correction to my earlier comment on this PR: the headers-truncation-detection commit is now fully split out to #1604 (removed from this branch entirely, not just duplicated), and I've now actually fixed the two real blocking items on this PR itself:

1. **Round-trip test didn't verify index** — the test constructed a UTXO at output index 2 specifically to catch an index-computation bug, but only asserted against `store.Get`'s returned slice, which the SQL store compacts by position — so it would pass identically whether the UTXO landed at index 2 (correct) or was silently miswritten to index 0. Added `GetSpend` checks against the actual (txid, vout) pair: asserts found at vout 2, not-found at vout 0. This now fails on a real misindexing bug instead of looking identical.
2. **Coinbase best-effort regression** — moot on this branch now that the headers-truncation commit that caused it lives in #1604 instead; that PR's own review will need to address it if it recurs there. (For the record: it doesn't recur there, since #1604 kept the partial-map-on-error behavior — see that PR.)

This PR now contains only the two original test files (round-trip + complete-file) plus the shared #1563 dependency, matching its original test-only scope.

icellan:
Addressed all three blocking items.

**#1 (coinbase best-effort regression)** — split out along with the whole behavioral commit, see below. The fix itself (return the partial map alongside the error instead of discarding it) is in #1604, with its own regression test proving a headers file truncated after N blocks still yields those N coinbases instead of none.

**#2 (round-trip test doesn't verify the index)** — `a17c4389` adds `GetSpend` checks against the actual `(txid, vout)` pair: asserts the UTXO is found at vout 2 (its real index) and absent at vout 0 (the first padded hole), so a misindexed UTXO now fails the test instead of looking identical. Also fixed the txA comment inaccuracy (both index 0 and 1 are real outputs there, no hole).

**#3 (undisclosed behavioral commit)** — split it out: #1604 now carries `3235e6b81` (headers-truncation detection) plus the item-#1 fix, with a description explaining the bug and the fix. This PR is back to test-only.

Non-blocking: the same EOF substring-matching hazard in the utxo-headers loops is exactly what #1604 fixes, so that gap is closed there rather than left as a follow-up. The #1563-bundling note resolves itself once #1563 merges and this branch rebases, as you said.

`go test ./cmd/seeder/... ./services/utxopersister/... -race` and `go vet` are green on both branches.

freemans13 (approved):
Approving. All three points from my earlier review are settled, and I checked each rather than taking them on trust.

**The two points about an undeclared behavioural change have gone away together.** Commit `3235e6b81` is off this branch and the work now lives in #1604, so the coinbase best-effort objection travels with it and belongs in that review rather than here. I confirmed the split is clean rather than assuming: `git diff c7ce2e252..HEAD -- cmd/seeder/seeder.go` comes back empty, so `seeder.go` on this branch is byte-identical to #1563's commit and the PR genuinely is test-only, as the title says.

**The round-trip point is the one I cared about, so I proved it.** `a17c4389a` adds the `GetSpend` checks against the real `(txid, vout)` pair. I broke `PadUTXOsWithNil` (`services/utxopersister/UTXOSet.go:821`) so it packs UTXOs at `padded[i]` instead of `padded[utxo.Index]` — exactly the index-computation bug the test exists to catch — and the test failed as it should:

```
--- FAIL: TestSeederImport_RoundTripsRealSnapshotFile
    seeder_roundtrip_test.go:162: expected: 0  actual: 3
    Messages: the UTXO must be recorded at vout 2, its real index
```

That mutation was invisible before `a17c4389a`. The assertion earns its place, and the stale "no leading nil hole" comment is fixed too.

None of what follows should hold the merge.

---

### The packed height/coinbase field is still untested

I mutation-tested four breakages to find out what these tests actually defend. Mis-indexing and mid-record truncation are both caught. Two others sail straight through.

`UTXOWrapper.Bytes()` squeezes the height and the coinbase flag into a single uint32 at `services/utxopersister/UTXO.go:123`:

```go
encodedValue := (uw.Height << 1) | flag
```

That is the least obvious line in the record format and the most likely to be got wrong. I changed the shift from `<< 1` to `<< 2` in the writer only, so the writer and reader now disagree about the very field this PR sets out to prove they agree on — and every test in `cmd/seeder` still passed. Separately, I made `processUTXO` pass `WithSetCoinbase(false)` unconditionally; again, all green.

The tests check values, scripts, and now indices, but never the height and never the coinbase flag. Both are cheap, because `store.Get` already returns them. I confirmed this works rather than suggesting it blind — adding

```go
require.Equal(t, []uint32{200}, gotB.BlockHeights)
```

passes on the clean tree and fails on the broken-shift mutation. For the coinbase half, one wrapper would need `Coinbase: true` with an `IsCoinbase` assertion; asserting `false` on the current wrappers cannot catch anything, since they are not coinbases either way.

One honest limit on that suggestion: `meta.Data.BlockHeights` is populated from `WithMinedBlockInfo`, not from the positional `blockHeight` argument to `SpendAndCreate`. I corrupted that argument separately (`utxoWrapper.Height+7`) and the `BlockHeights` assertion did not notice. So the line above covers the record encoding and the mined-block height, but leaves the positional argument untested. Lower value — same source value, and it feeds maturity and locktime rather than what gets stored — but worth knowing it is not as total as it looks.

### Two small corrections to the description

It says the round-trip test checks "values, scripts, coinbase handling". There is no coinbase handling in either new test: no wrapper sets `Coinbase: true`, and `processUTXO` is called with a `nil` coinbase map throughout. That path is covered in `coinbase_restore_test.go`, just not here. Worth correcting so a later reader does not trust a guarantee that is not there.

It also says both tests share the existing helpers "rather than adding new scaffolding". It does add scaffolding — `writeCompleteSnapshotFile` at `cmd/seeder/seeder_roundtrip_test.go:27` — and then the second test does not use it, assembling the same header, metadata, records and footer inline at lines 184-200. The duplication exists because the helper takes `[]*UTXOWrapper` while `buildSnapshotWrappers` hands back pre-encoded `[][]byte`. Cosmetic, but it is two copies of the file layout to keep in step if the format ever changes.

### Minor notes

The footer re-read at lines 224-231 is close to tautological: the test writes a footer from `txCount`/`utxoCount`, then asserts `GetFooter` reads those same values back. That exercises `GetFooter`'s seek-from-end parsing against the writer's layout, which is worth a little, but it is the test checking its own fixture rather than the import path.

The real value of the complete-file test, I think, is as the negative control for the truncation test in #1563 — same `buildSnapshotWrappers` fixture, spliced versus whole, so a future change that made the truncation test pass for the wrong reason would show up here. That is a stronger justification than the one in the description and worth saying out loud in the comment.

On Copilot's note about the reference at line 175: I think it is half right. GitHub auto-links issue references in prose it renders, not inside Go source in a diff, so the AGENTS.md concern being cited does not really bite here. But the reference is genuinely confusing — it means audit item 14, and now that those items map onto #1563, #1572 and #1604, a reader will read it as a PR number. Spelling it out as "audit item 14" in words fixes it.

Test isolation is fine, for the record. The shared `newTestUtxoStore` helper hardcodes `sqlitememory:///coinbase_restore`, which looks like a collision risk, but `util/sql.go:218` discards that path and generates a random 16-character in-memory database name per store. No leakage between tests.

---

**What I ran**, on `a17c4389a`, with a clean tree before and after each mutation:

- `go test -race -count=1 -run 'TestSeederImport|TestReadUTXOWrapperFile|TestProcessUTXO' ./cmd/seeder/...` — pass
- `go vet ./cmd/seeder/...` — clean
- `golangci-lint run cmd/seeder/...` — 0 issues
- four mutations, each reverted immediately, results as described above

PR #1573, opened 2026-08-10
docs(k8s): explain why HPA can't safely target the propagation deployment

icellan:
## Summary

An audit recommended shipping an example `HorizontalPodAutoscaler` manifest for the operator-managed propagation Deployment, as a convenience for operators who want automatic scale-out on top of the existing manual `deploymentOverrides.replicas` control.

Investigated whether that's actually wireable against the Teranode Operator before writing one. It isn't, so this ships a documentation note instead of a manifest that would silently misbehave.

## What I found

- The propagation service is manually horizontally scalable today via `propagation.spec.deploymentOverrides.replicas` in the Cluster CR — confirmed in `deploy/kubernetes/teranode/teranode-cr.yaml` and `teranode-cr-mainnet.yaml`.
- The operator's `Propagation` CRD (`api/v1alpha1/propagation_types.go` in `bsv-blockchain/teranode-operator`) does declare a `/scale` subresource (`+kubebuilder:subresource:scale:specpath=.spec.deploymentOverrides.replicas,...`), which is the mechanism an HPA normally uses to target a custom resource directly — so on paper this looks HPA-ready.
- In practice it isn't safe: the parent `Cluster` controller reconciles every child resource including `Propagation` (`cluster_controller.go`'s `ReconcileBatch`) both on a fixed one-minute `RequeueAfter` and whenever the `Propagation` object's own spec changes (which writing to `/scale` triggers, since it bumps `.metadata.generation` and the controller watches `Owns(&Propagation{})` with `GenerationChangedPredicate`). Each reconcile does `propagation.Spec = *cluster.Spec.Propagation.Spec` unconditionally — overwriting the replica count an HPA just set back to whatever's baked into the Cluster CR, within at most a minute.
- Net effect: pointing an HPA at the propagation Deployment (or at the Propagation CR's scale subresource) wouldn't converge — it would flap.

Given that, shipping an example HPA manifest would be worse than not shipping one: it would look like working autoscaling while actually racing the operator's own reconciler.

## Change

Added a "Scaling the Propagation Service" subsection to `docs/howto/miners/kubernetes/minersHowToInstallation.md` explaining the constraint and pointing at the supported manual-scaling path (`deploymentOverrides.replicas`).

## Test plan

- [x] Doc-only change; verified rendering and links via the existing markdown lint pre-commit hook (passed).
- [x] No manifest was added since none would actually function against the current operator.

### DIFF
--- a/docs/howto/miners/kubernetes/minersHowToInstallation.md
+++ b/docs/howto/miners/kubernetes/minersHowToInstallation.md
@@ -296,29 +296,31 @@ the propagation spec has no `deploymentOverrides` block yet — as in
 `teranode-cr.yaml` — add one.
 
 Not every service in the Cluster CR can be scaled this way. The validator can:
-`teranode-cr-mainnet.yaml` runs it with 8 replicas. Services such as block
-assembly and blockchain are single-instance — block assembly holds the mining
-jobs handed to miners and its assembler state per process, so a second replica
-would lose solved blocks and clobber that state. That is why the example CRs set
-`replicas: 1` for every service except the validator.
+`teranode-cr-mainnet.yaml` runs it with 8 replicas (`teranode-cr.yaml` ships the
+validator disabled, with an empty spec). Services such as block assembly and
+blockchain are single-instance — block assembly holds the mining jobs handed to
+miners and its assembler state per process, so a second replica would lose
+solved blocks and clobber that state. That is why every
+`deploymentOverrides.replicas` entry in both example CRs is `1`, the one
+exception being the validator in `teranode-cr-mainnet.yaml`.
 
-There is currently no supported way to put a `HorizontalPodAutoscaler` in front
-of the operator-managed propagation Deployment. The Teranode Operator's
-`Propagation` custom resource does declare a Kubernetes `/scale` subresource,
-which is the mechanism an HPA would normally use to target a custom resource
-directly. In practice this doesn't hold up: the parent `Cluster` controller
-reconciles every child resource, including `Propagation`, both on a fixed
-one-minute timer and whenever the `Propagation` object's spec changes (which an
-HPA writing to `/scale` would trigger). On each of those reconciles it
-overwrites the entire `Propagation` spec — including the replica count — with
-whatever is stored in the `Cluster` CR itself. Any replica count an HPA sets
-gets reverted within at most a minute, so autoscaling would silently flap
-rather than converge.
+A `HorizontalPodAutoscaler` that targets the `Propagation` custom resource's
+`/scale` endpoint will not work. The Teranode Operator's `Propagation` CR does
+declare a Kubernetes `/scale` subresource, with its `specReplicasPath` pointing
+at `.spec.deploymentOverrides.replicas`, which is the mechanism an HPA would
+normally use to target a custom resource directly. In practice this doesn't hold
+up: the parent `Cluster` controller reconciles every child resource, including
+`Propagation`, both on a fixed one-minute timer and whenever the `Propagation`
+object's spec changes (which an HPA writing to `/scale` would trigger). On each
+of those reconciles it overwrites the entire `Propagation` spec — including the
+replica count — with whatever is stored in the `Cluster` CR itself. Any replica
+count an HPA writes to `/scale` gets reverted within at most a minute, so
+autoscaling against that endpoint silently flaps rather than converging.
 
 Until the operator changes this reconcile behavior (e.g. by leaving
 externally-managed replica counts alone), manual scaling via
-`deploymentOverrides.replicas` remains the supported way to size the
-propagation service.
+`deploymentOverrides.replicas` in the `Cluster` CR remains the supported way to
+size the propagation service.
 
 ## Related Documentation
 
--- a/docs/howto/miners/kubernetes/minersHowToInstallation.md
+++ b/docs/howto/miners/kubernetes/minersHowToInstallation.md
@@ -14,6 +14,7 @@ Last modified: 29-October-2025
     - [Deploy Teranode](#deploy-teranode)
 - [Verifying the Deployment](#verifying-the-deployment)
 - [Production Considerations](#production-considerations)
+    - [Scaling the Propagation Service](#scaling-the-propagation-service)
 - [Other Resources](#other-resources)
 
 ## Introduction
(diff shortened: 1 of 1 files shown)


ordishs (approved):
Verified against the teranode-operator source at master. The Propagation CRD declares a scale subresource with `specReplicasPath .spec.deploymentOverrides.replicas`, `ClusterReconciler` requeues after one minute, it watches `Owns(&Propagation{})` with `GenerationChangedPredicate`, and `updatePropagation` in `internal/controller/cluster_propagation.go` does an unconditional `propagation.Spec = *cluster.Spec.Propagation.Spec`. An HPA on the CR `/scale` endpoint really does flap. The block assembly claim also holds, because `jobStore` is an in-process ttlcache and `SubmitMiningSolution` returns `NotFound` on a miss.

Two wording nits:

- "There is currently no supported way..." is wider than the mechanism that supports it. The clobber is unconditional only for an HPA that targets the Propagation CR `/scale` endpoint. Please scope the sentence to that.
- "That is why the example CRs set `replicas: 1` for every service except the validator" does not match `deploy/kubernetes/teranode/teranode-cr.yaml`.

ctnguyen (approved):
Review at commit [eae0a60](https://github.com/icellan/teranode/commit/eae0a608c60905a4d25e7c93324a241ed0eaac61)

Good call shipping a documented reason instead of the HPA manifest the audit asked for — a note explaining why the thing flaps is worth more than a manifest that flaps. I checked the section against the tree and it holds up: the two example CRs, the mainnet validator's 8 replicas, and the block-assembly rationale (`jobStore` is an in-process ttlcache and a solution submitted to the wrong replica really does come back `NotFound`). The operator behaviour checks out too, verified against the operator repo at 60dd7fcd. Two small things left, neither a blocker.

## Non-blocking issues

### ChiR1 — The "that is why" behind `replicas: 1` attributes the wrong cause

*Partial overlap with [@ordishs's point 2 (11-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1573#pullrequestreview-4907184037) — he caught the count mismatch and the head commit fixed it; this is the causal half of the same sentence, which survived the fix.*

**Problem:** The paragraph establishes that block assembly and blockchain are single-instance, then concludes "That is why every `deploymentOverrides.replicas` entry in both example CRs is `1`, the one exception being the validator in `teranode-cr-mainnet.yaml`". The count is now exactly right — I walked all 9 entries in `teranode-cr.yaml` and all 14 in `teranode-cr-mainnet.yaml`. The causal claim isn't. Propagation itself ships `replicas: 1` in `teranode-cr-mainnet.yaml:136`, and the same paragraph opens by calling propagation horizontally scalable. Asset and subtree validation are documented as multi-pod services (`docs/topics/services/assetServer.md:101`, `docs/topics/services/subtreeValidation.md:163`) and both also ship at `1`. The `1`s are there because the examples are minimal, not because the services are pinned.

**Why it matters:** A miner sizing a production cluster reads "every entry is `1` because these services are single-instance" and concludes asset, subtree validation and propagation can't be scaled either — the opposite of what the rest of the docs say, and the opposite of what this section is trying to teach. It also makes the paragraph contradict its own opening sentence.

**Fix:** Break the causal link and keep the (now correct) observation: "The example CRs are deliberately minimal, so every `deploymentOverrides.replicas` entry is `1`, the one exception being the validator in `teranode-cr-mainnet.yaml`. Those values are starting points, not per-service ceilings — block assembly and blockchain are the ones that genuinely have to stay at `1`."

**Criticality: 30%**

### ChiR2 — The caveat stops at the CR `/scale` endpoint and leaves the Deployment route unaddressed

**Problem:** After the head commit the caveat is scoped precisely to "a `HorizontalPodAutoscaler` that targets the `Propagation` custom resource's `/scale` endpoint" — the right scope for what's actually been verified. But the next thing a reader reaches for is an HPA pointed straight at the `propagation` Deployment the operator generates, and the section says nothing about it either way. The closing paragraph only says manual scaling "remains the supported way", which reads as a recommendation rather than a statement about that route.

**Why it matters:** I want to be careful here: I have not verified what happens on the Deployment route, and I'm not claiming it's broken. The evidence I have covers only the `Cluster` → `Propagation` spec overwrite; the Propagation controller's Deployment reconcile lives in the operator repo and I didn't check it. That's precisely the point — the section is written to stop someone shipping an autoscaler that flaps, and as it stands a reader can't tell whether the obvious alternative is safe, also broken, or simply untested. Leaving that open in the one document that exists to answer the question is the gap.

**Fix:** One clause either way, once someone confirms the behaviour. If the `Propagation` controller writes the Deployment's `replicas` from `spec.deploymentOverrides.replicas` on every reconcile, say so and the caveat generalises. If it doesn't, that's worth stating even more — it would be the one route that works. If nobody has tested it, "untested" is still a better answer than silence.

**Criticality: 15%**

## Recap

| --- | --- | --- | --- |
| ChiR1 | Wrong cause for `replicas: 1` | | 30% |
| ChiR2 | Deployment HPA route unaddressed | | 15% |

freemans13 (commented):
Agreed with the call this PR makes — a documented reason why the thing flaps is worth more than an example manifest that flaps. I re-checked the factual claims against the two example CRs here and against `bsv-blockchain/teranode-operator` at master, and they all hold: propagation ships disabled in both CRs, `teranode-cr.yaml` really has no `deploymentOverrides` block under propagation, the mainnet validator really runs 8, the `/scale` subresource is declared at `api/v1alpha1/propagation_types.go:45`, and the reconcile story is exactly as described (`cluster_controller.go:126` requeues after a minute, line 142 owns `Propagation`, line 151 applies `GenerationChangedPredicate`, and `cluster_propagation.go` does `propagation.Spec = *cluster.Spec.Propagation.Spec` unconditionally).

Two things I'd raise on the text itself.

**The section no longer covers the case a reader is most likely to try.** The caveat is now scoped to the CR's `/scale` endpoint, but the title still says "the propagation deployment", and pointing an HPA at the Deployment is the obvious first thing an operator reaches for. That case is broken too, by a different path: `propagation_deployment.go`'s `updateDeployment` calls `SetDeploymentOverridesWithContext`, which does `dep.Spec.Replicas = cr.DeploymentOverrides().Replicas` whenever that field is non-nil (`internal/utils/deployment.go:276`), and `propagation_controller.go:128` requeues every five seconds while status and spec disagree. So a Deployment-targeted HPA gets reverted as well — faster than the `/scale` one, if anything.

The flip side is the part that bites. `updateDeployment` only writes the full deployment spec when the object is new (`isNewDeployment := dep.ResourceVersion == ""`, line 49), and the replicas override is skipped entirely when `Replicas` is nil. So if `replicas` is simply left out of the Cluster CR, the operator never touches `dep.Spec.Replicas` after creation — and a Deployment-targeted HPA looks like it would actually converge. The opening advice here ("if the propagation spec has no `deploymentOverrides` block yet — as in `teranode-cr.yaml` — add one") is precisely what forecloses that. Worth a sentence covering the Deployment case explicitly. I'd stop short of documenting the omit-`replicas` path as supported, though: I've read that out of the operator source rather than tried it on a cluster, and it depends on an implementation detail that could change without notice.

**One causal claim overreaches.** "That is why every `deploymentOverrides.replicas` entry in both example CRs is `1`" reads the 1s as evidence that those services are single-instance. But asset, RPC, peer, block validator and subtree validator are all `1` in the examples too, and this repo's own `docs/topics/services/assetServer.md:101` says multiple asset server instances are supported. The 1s look like conservative example defaults rather than scaling constraints, so I'd soften "that is why" to something like "the examples all start at 1" and keep the single-instance argument to block assembly and blockchain, where it's actually made.

Neither is blocking — the substance of the section is right and well evidenced.

PR #1574, opened 2026-08-10
test: skip container-backed tests without Docker, fix a port-bind test flake, fix gotestsum PATH fallback

icellan:
## Summary

- Several Postgres- and Aerospike-backed unit tests called `require.NoError` directly on the container-start error, so they failed hard instead of skipping when no container runtime (Docker/OrbStack) is reachable. Brought them in line with the skip-with-reason pattern already used elsewhere in the suite (`t.Skipf` on container-start failure), across `stores/blockchain/sql`, `stores/utxo/sql`, `stores/utxo/aerospike` (and its `pruner` package), `stores/utxo/tests`, `util/usql`, `util/kafka`, `services/validator`, and `services/propagation`.
- `Test_CoinbaseSubsidyHeight` (and its siblings in the same test file) start a throwaway blockchain daemon and block-assembly gRPC server on OS-assigned free ports. Reproduced the flake under machine-wide parallel load: the gap between listen-then-close (to pick a free port) and the later gRPC bind lets another process grab that exact port, failing startup non-deterministically ("can't start GRPC server" / connection refused). Retried the whole test setup with freshly chosen ports on that failure, the same way the Postgres test-container helper already retries its own bind race.
- The `test` Makefile recipe's `gotestsum` fallback only installs the binary into `GOPATH/bin` but still invoked the bare `gotestsum` command, which fails on a fresh checkout that hasn't added `GOPATH/bin` to `PATH`. Now invokes it via `$(go env GOPATH)/bin/gotestsum`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...` (no new findings; two pre-existing unrelated findings in `test/utils/helper.go` / `test/utils/transaction_helper.go`)
- [x] `golangci-lint run` on touched packages (0 issues; 5 pre-existing `prealloc` findings in untouched files)
- [x] Reproduced the original `Test_CoinbaseSubsidyHeight` port-bind flake under real parallel load (multiple concurrent OS processes), confirmed it no longer reproduces with the fix
- [x] `go test -race -count=5 -run Test_CoinbaseSubsidyHeight ./services/blockassembly/...` — green
- [x] `SETTINGS_CONTEXT=test go test -tags testtxmetacache ./util/test/... ./util/usql/... ./util/kafka/... ./stores/blockchain/sql/... ./stores/utxo/sql/... ./stores/utxo/tests/... ./stores/utxo/aerospike/... ./services/validator/... ./services/propagation/...` — green, with the previously-hard-failing container tests now skipping cleanly when no Docker/OrbStack is reachable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,7 +185,7 @@ install-tools:
 .PHONY: test
 test:
 	@command -v gotestsum >/dev/null 2>&1 || { echo "gotestsum not found. Installing..."; $(MAKE) install-tools; }
-	SETTINGS_CONTEXT=test gotestsum --format pkgname -- -race -tags "testtxmetacache" -count=1 -timeout=10m -coverprofile=coverage.out -coverpkg=./... $$(go list ./... | grep -v github.com/bsv-blockchain/teranode/test/ | sort)
+	SETTINGS_CONTEXT=test $$(go env GOPATH)/bin/gotestsum --format pkgname -- -race -tags "testtxmetacache" -count=1 -timeout=10m -coverprofile=coverage.out -coverpkg=./... $$(go list ./... | grep -v github.com/bsv-blockchain/teranode/test/ | sort)
 
 # run tests in the test/longtest directory
 .PHONY: longtest
--- a/services/blockassembly/blockassembly_system_test.go
+++ b/services/blockassembly/blockassembly_system_test.go
@@ -74,6 +74,39 @@ func setupTest(t *testing.T) (*nodehelpers.BlockchainDaemon, *BlockAssembly, con
 	return nil, nil, nil, nil, nil
 }
 
+// startAttemptTimeout bounds how long attemptSetupTest waits, after readyCh
+// closes, for an error that Start's launcher goroutine may still be in the
+// process of delivering. Mirrors cleanupFn's own bounded wait below.
+const startAttemptTimeout = 200 * time.Millisecond
+
+// waitForEarlyStartErr waits for readyCh to close, then waits up to timeout
+// for a value on startErrCh.
+//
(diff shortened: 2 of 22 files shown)


ordishs (approved):
Verified the three problems. Several files did call `require.NoError` directly on container-start errors, the Makefile fallback did invoke a bare `gotestsum` that is not on PATH after a fresh `go install`, and the listen-then-close-then-rebind pattern in the blockassembly setup is a real race. `go build ./...` succeeds and the new `SkipIfContainerUnavailable` unit tests pass. Nits only.

- `stores/blockchain/sql/*_test.go`, `stores/utxo/sql/close_drain_postgres_test.go`, `unlock_batcher_postgres_test.go`, `util/usql/advisory_lock_test.go` — `SkipIfContainerUnavailable(t, err)` was inserted **before** the existing `require.NoError(t, err)`, not in place of it. Both `Skipf` and `Fatalf` call `runtime.Goexit()`, so the trailing line is dead. Please remove it.
- `services/blockassembly/blockassembly_system_test.go`, `attemptSetupTest` — the post-`readyCh` select uses a bare `default:` on `startErrCh`. `Start()` closes `readyCh` in a defer, and the outer goroutine only sends on `startErrCh` after `Start()` returns. The window is very narrow and a miss still fails later rather than passing, so this is defensive only. `cleanupFn` a few lines below already uses a bounded wait on the same channel — the same pattern here would close the gap.
- `services/blockassembly/load_unmined_benchmark_test.go:65,74,263` — the same hard `require.NoError` container-start pattern is left unconverted. These are benchmarks, so they do not affect `make test`. Optional follow-up.

freemans13 (approved):

PR #1575, opened 2026-08-11
Reject placeholder gRPC admin API key and remove committed testkey

gokutheengineer:
**Closes bitcoin-sv/teranode#4738.**

### Problem
`settings.conf` committed a real, non-empty admin API key (`grpc_admin_api_key = testkey`) with no context override. Because this repository is public, that value is world-readable, so the admin-RPC auth interceptor added in #1257 protected nothing on every default deployment: anyone reaching the P2P (or Legacy) gRPC port could call `ClearBanned`, `BanPeer` (durable, restart-surviving subnet self-eclipse), `ConnectPeer`, etc. The listener also runs at `securityLevelGRPC = 0`, so the key travels in cleartext.

The single shared key gates both the P2P admin RPCs and the Legacy service's protected methods, and - crucially - internal clients (`services/p2p/Client.go`, `services/legacy/peer/Client.go`, the RPC and alert services) read the **same** setting to authenticate *outbound* calls. So simply deleting the value is not safe on its own: the server would mint a random key while clients send none, silently breaking ban/unban/connect with `Unauthenticated`.

### Fix
1. **Remove the committed `testkey`** so the default is unset. Unset is fail-closed: each server mints a random key at startup and the protected admin RPCs are simply unreachable until an operator configures a real shared secret. The bare `grpc_admin_api_key =` line is dropped entirely (it defaulted to empty anyway and only fed gocore's override echo).
2. **Ignore placeholders at startup, keep the node up.** New `util.ValidateAdminAPIKey` (called from `p2p.Server.Start`/`legacy.Server.Start` where the key is resolved) detects a well-known placeholder (`testkey`, `changeme`, …; case/whitespace-insensitive), logs an `Errorf`, and falls back to the random-key path. The security outcome is identical to unset (fail-closed), but the node keeps validating/relaying blocks rather than being taken offline by a setting that only guards admin RPCs. `cmd/diagnose` still flags a placeholder as `ERROR` as the preflight gate.
3. **Warn on cleartext exposure** (not fail): when a real key is served on a non-loopback listener without verified TLS (`securityLevelGRPC <= 1`; level 1 uses `InsecureSkipVerify` and is MITM-exploitable), log a warning. Internal clusters on trusted networks legitimately run without TLS and must not be crash-looped.
4. **Stop leaking the generated key**: the Legacy server previously logged its random key at `Info` (`[Legacy] Generated admin API key: %s`) - a secret-in-logs bug that defeated the fail-closed intent. Now it logs a warning without the value, matching P2P.
5. **Surface the client-side regression**: both internal client constructors now `Warnf` when the key is empty **or a placeholder** (a placeholder is ignored by the server, so a client sending it is still rejected - the old "using API key" log was misleading). `handleClearBanned` now reports failure instead of always returning success: it tolerates an **absent** leg (a configured-but-not-running service - `DeadlineExceeded`/`Unavailable`, #591) but fails on a leg that **refused** (`Unauthenticated`, or `!Ok`) or when nothing cleared - so an unauthenticated p2p leg can't be masked by the unprotected legacy leg (see ChiR14 below). It checks the legacy `resp.Ok`, surfaces the failing leg's reason in the RPC error message, and returns `ErrRPCInternal` rather than "invalid parameter". (`TestBanListRPCE2E`, where the legacy service is configured-but-absent, drove getting the absent-vs-refused split right.)
6. **Normalize once**: trim `grpc_admin_api_key` at settings load so a whitespace-only value resolves to empty consistently for every reader (servers and clients), rather than becoming a live secret in one place and not another.
7. **No committed key for any context**: `TestDaemon` assigns a shared non-placeholder key so in-process server + clients agree. The docker e2e stacks inject their shared test key via the compose `environment:` blocks (`test/docker-compose.e2etest*.yml`), which gocore reads (via `os.Getenv`) ahead of the config files - so nothing lands in the shared `docker.teranodeN.test` contexts that `compose/docker-compose-3blasters.yml` also uses.
8. Docs corrected: `settings/interface.go` `desc`/`longdesc`, `rpc_reference.md`, `legacy_reference.md`, and `topics/services/rpc.md` (they claimed empty = "unauthenticated admin operations" and that the generated key "will be logged"), plus an explicit operator-facing note that a *configured* key is currently echoed in the startup settings dump.

Improvement over the issue's proposed fix: the issue suggested hard-failing when the key is empty **or** a placeholder while non-loopback. Hard-failing on *empty* would crash every default non-loopback deployment (docker/operator/mainnet), since empty is the safe fail-closed path.

### Review changes (from @ctnguyen)
- **ChiR1 (hard-fail → ignore)**: the earlier revision hard-failed `Init` on a placeholder, which - traced end to end - took the whole node offline with a graceful-looking exit-0 and no restart policy in the shipped compose stacks. Since the security outcome of ignore-and-fall-back-to-random is identical (fail-closed) and the value only guards admin RPCs, the placeholder is now ignored (logged) instead of fatal.
- **ChiR2 (secret-in-logs / docs)**: added the operator-facing warning that a *configured* key is echoed in the startup settings dump, kept the "(never logged)" wording scoped strictly to the *generated* key, and dropped the bare `grpc_admin_api_key =` declaration. The underlying leak lives in shared gocore startup/override logging (it ignores `redact:"true"` and leaks `rpc_pass` / `p2p_private_key` the same way) and is out of scope for this diff - see Residual.
- **ChiR3 (docker key)**: moved the shared e2e key out of `settings.conf` (where it reached `docker.teranodeN.test`, used by 3blasters too) into the test-only compose `environment:`. Note: the suggested `${TERANODE_E2E_ADMIN_KEY}` interpolation in `settings.conf` is unsafe here because gocore resolves an *unset* `${VAR}` to the literal `{UNKNOWN}` (`config.go` `replaceVariables`), which would itself be a usable world-readable key for any stack that doesn't set the var; injecting via compose env avoids that.

### Second review round (@ctnguyen / @freemans13)
- **ChiR9 (no deploy surface names the key)**: added a commented `grpc_admin_api_key` pointer to `deploy/docker/base/settings_local.conf` and its `.template`, and extended the existing "not-in-configmap" Secret note in `deploy/kubernetes/teranode/teranode-configmap.yaml`, each naming the fail-closed consequence for `setban`/`clearbanned`/alert bans/reset-reputation. (Release-note wording for the upgrade is called out below for the maintainers' release process.)
- **ChiR10 (legacy `Ok` ignored)**: `handleClearBanned` now checks `resp == nil || !resp.Ok` on the legacy leg (matching `handleSetBan`), so a soft "did not clear" no longer counts as a clear. Tests updated: mocks return `Ok: true` where the leg is meant to succeed, a real `RPC.ClientCallTimeout` is set so the legacy leg runs as in production, and a new "legacy reports not-ok → failure" case was added.
- **ChiR4 (error hides cause)**: `clearbanned` failure now returns `ErrRPCInternal` and includes the failing leg's reason in the message.
- **ChiR11 (client logs contradict server)**: both client constructors treat a placeholder as unset and warn (see point 5).
- **ChiR5 (no length floor)**: `ValidateAdminAPIKey` now warns (does not fail) when a real key is shorter than 16 chars, matching the `cmd/diagnose` threshold.
- **ChiR6 (bracketed IPv6 loopback)**: `isLoopbackListenAddress` now strips brackets from a bare `[::1]` (no port); added test cases.
- **ChiR7 (longdesc overstated protection)**: the `grpc_admin_api_key` `longdesc` no longer lists "get peer info" as protected; it now names the actual state-mutating admin RPCs and notes read-only queries stay unauthenticated.
- **ChiR8 (tautological test)**: replaced with `settings.TestNewSettingsTrimsAdminAPIKey`, which actually calls `NewSettings` and asserts a padded/whitespace env value is trimmed at load.

### Third review round (@ctnguyen / @freemans13)
- **ChiR14 (clearbanned masks a *refused* p2p leg)**: `ClearBanned` is protected on p2p but not legacy, so with the key unset and both services enabled, the p2p leg returned `Unauthenticated` while the unprotected legacy leg cleared - and "any leg succeeded" reported success while every p2p ban survived. `handleClearBanned` now distinguishes a leg that is **absent** (`DeadlineExceeded`/`Unavailable` - a configured-but-not-running service, tolerated per #591) from one that **answered and refused** (`Unauthenticated`, or `!Ok`). It fails if any reachable leg refused, even if another cleared. New `legAbsent(err)` helper via gRPC status codes (both clients surface the raw status for transport/auth failures). Tests retargeted: the old "one leg fails but other succeeds" case (which pinned the bug) is replaced by an absent-leg-tolerated case, a shipped-default case (peer off → `Unavailable`, legacy clears → success), and the ChiR14 regression guard (p2p `Unauthenticated` + legacy `Ok` → failure).
- **ChiR15 (setban error class/cause)**: `handleSetBan` now returns `ErrRPCInternal` (not `ErrRPCInvalidParameter`) and names the failing leg's reason, matching `handleClearBanned`. setban keeps any-leg-succeeds (no refused-masking hole: `BanPeer`/`UnbanPeer` are protected on *both* services, so a refused leg can't be papered over). Updated the two setban failure-path test assertions.

### Upgrade note (for release notes)
Deployments that relied on the previously committed `grpc_admin_api_key = testkey` must now set a real shared `grpc_admin_api_key` (env or secret store), or the cross-container admin RPCs (`setban`/`clearbanned`, connect/disconnect peer, alert-driven bans, reset reputation) return `Unauthenticated`. Automatic peer banning inside p2p/legacy is unaffected. `cmd/diagnose` reports an empty or placeholder key.

### Residual (recommend follow-up issues, not fixed here)
- **gocore logs configured secrets at startup**: `gocore.Config().Stats()` and the override echo print resolved values ignoring `redact:"true"`, so a configured `grpc_admin_api_key` (and `rpc_pass`, `p2p_private_key`) appears in the startup log. The fix belongs in the shared startup plumbing, not this diff; documented here and flagged for a separate issue.
- **`rpc_user = bitcoin` / `rpc_pass = bitcoin`** are still committed and guard the entire admin JSON-RPC surface **fail-open** with a world-known credential - strictly worse than the gRPC key. Left to a separate issue to keep this PR scoped to #4738.
- **Optional strict mode**: if some operators want a hard startup failure on a placeholder, that could be added later behind an explicit opt-in setting rather than as the upgrade default.

### Tests
- `util/admin_api_key_test.go`: `ValidateAdminAPIKey` (placeholder → `ignoreKey=true` + `Errorf` via a capturing logger; short-key length warn; cleartext warn at level 0/1, none at level 2/loopback), `IsPlaceholderAdminAPIKey` (exact-match, no false positives), `isLoopbackListenAddress` (IPv4/IPv6/unspecified/`localhost`/mapped/malformed/bare-`[::1]`).
- `settings/admin_api_key_test.go` (`TestNewSettingsTrimsAdminAPIKey`): padded/whitespace env value is trimmed at load.
- `cmd/diagnose/config_checks_test.go`: placeholder→ERROR, weak→WARN, strong→OK.
- `services/rpc` `TestHandleClearBannedComprehensive`: updated for the corrected contract (any-leg-succeeds; legacy `!Ok` and all-fail and no-service report failure; real client-call timeout).
- Commands run (worktree, affected packages only):
  - `SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./util/ ./settings/` (relevant tests) → **ok**
  - `go vet -tags testtxmetacache ./util/ ./settings/ ./services/p2p/ ./services/legacy/ ./services/legacy/peer/ ./services/rpc/` → **clean**
  - `go build -tags testtxmetacache` on all changed packages → **clean**
- Not run locally: the `cmd/diagnose`, p2p/legacy, rpc, and daemon test binaries fail to *link* on this machine due to the known GoBDK cgo linker issue (`Undefined symbols: _TxValidator_*`); they compile (vet passes) and run in CI.


freemans13 (approved):

ctnguyen (changes requested):
Review at commit [28adbdf](https://github.com/gokutheengineer/teranode/commit/28adbdfca04bfbac9f61635aa97aa940cfcae3e3)

This is a well-scoped fix for a genuinely nasty hole, and the core reasoning is right: deviating from the issue by keeping `empty → random key` instead of hard-failing on empty is the correct call, and I traced the fail-closed claim rather than taking it on faith — `util/grpc.go:66-71` installs the auth interceptor whenever `APIKey != ""`, and both servers feed it a 64-char hex random key when the setting is empty (`services/p2p/Server.go:843-858`, `services/legacy/Server.go:656-680`), so the protected methods really are unreachable rather than open. The placeholder matcher is exact-match after trim+lower, so there are no false positives on real keys (`util/admin_api_key.go:32-35`). The `.test` → `.test.legacy` fallback claim also checks out against gocore's `findValue` (it strips one dotted suffix at a time, `gocore@v1.0.82/config.go:647-677`), so the docker e2e contexts do resolve a shared key. Fixing the Legacy `Infof` that printed its generated key, correcting the docs that claimed empty meant "unauthenticated admin operations", and the `cmd/diagnose` ERROR path are all good touches, and the test tables are unusually thorough for a change this size. Two things need settling before merge, though — the hard-fail question you flagged, and a secret-in-logs gap that the doc rewrite walks away from rather than closes.

## Blocking issues

### ChiR1 - Placeholder key hard-fails the whole node, silently, with exit 0
**Problem:** You asked for a verdict on the hard fail, so here it is: as implemented it needs a migration path. I traced the failure shape end to end. `util.ValidateAdminAPIKey` returns a configuration error from `services/p2p/Server.go:653-656` and `services/legacy/Server.go:255-258`; `ServiceManager.AddService` propagates it (`util/servicemanager/service_manager.go:170-172`); `Daemon.Start` logs one `Errorf`, calls `sm.ForceShutdown()` and closes `doneCh` (`daemon/daemon.go:305-310`) — and then *continues past the failure*, registering Prometheus metrics and launching the health server, before returning through the later select. None of those paths carries a non-zero exit, and there is no `os.Exit` anywhere in `daemon/` or `cmd/teranode/`, so `main` returns normally and the process terminates with **status 0**. The repo's own compose stacks give the teranode containers no restart policy at all (`test/docker-compose.e2etest.yml:137,167,197` — only postgres and kafka-console carry `restart: on-failure`), so an exit-0 teranode container just stays down. The net effect for anyone who copied `testkey` out of this repo's shipped `settings.conf` into `settings_local.conf` is: node stops validating and relaying blocks after an upgrade, one ERROR line in the log, and an exit code that reads as a clean shutdown to anything watching process exits.
**Why it matters:** The blast radius is inverted — the value guards only the admin RPC surface (`adminProtectedMethods()`, `services/p2p/Server.go:1020-1030`, plus Legacy's two methods), but the failure takes the node's consensus participation offline; and the operators who trip it are precisely the ones who never treated `testkey` as a secret, i.e. almost certainly aren't using admin RPCs at all. The security outcome of your softer alternative is identical, because placeholder→empty lands on the already-verified fail-closed random-key path (see the deliberate deviation recorded in [bitcoin-sv/teranode#4738](https://github.com/bitcoin-sv/teranode/issues/4738) — the issue's own verification confirms empty is safe). Hard failure buys attention, not security, and pays for it in availability.
**Fix:** Treat a placeholder as empty and keep the node up. Keep the `cmd/diagnose` ERROR as the preflight gate, and if you want an escape hatch for strict operators, put the hard fail behind an explicit opt-in setting rather than making it the upgrade default.
```go
// ValidateAdminAPIKey now reports whether the caller should ignore the
// configured key and fall back to the random-key (fail-closed) path.
if IsPlaceholderAdminAPIKey(trimmed) {
    logger.Errorf("[%s] grpc_admin_api_key is the well-known placeholder %q; this repository is public so it protects nothing. Ignoring it: a random key is used instead, so admin RPCs are unreachable until a real secret is configured", serviceName, trimmed)
    return true, nil // ignoreKey
}
```
If you keep the hard fail instead, it needs at minimum a release-note entry naming `testkey` explicitly and a non-zero exit status, so the outage is attributable rather than looking like a graceful stop.

### ChiR2 - A configured admin key is printed in cleartext at every startup, and the doc rewrite drops the warning that said so
**Problem:** Two verified leak paths, both on a stock node. (1) `cmd/teranode/daemon.go:83-84` logs `gocore.Config().Stats()` at `Info` on every startup. `Stats()` walks every key present in gocore's merged settings map and prints its *resolved* value (`gocore@v1.0.82/config.go:1130-1153` and `1193-1202`); its only masking is for `*EHE*`-encrypted values (`config.go:57-63`), and `settings.Redact` is applied to the JSON dump in `cmd/settings/Settings.go:172` but **not** to the `Stats()` block. `grpc_admin_api_key` is in that map (declared at `settings.conf:674` and `679-681`), and resolution checks `os.LookupEnv` before the files (`config.go:620-627`) — so the operator's real secret is printed whether it came from `settings_local.conf`, any other settings file, or the environment. (2) gocore merges all settings files into one map and echoes every override with both values: `log.Printf("INFO: %s:%d is replacing %q: %q -> %q", ...)` (`config.go:210-214`), so a `settings_local.conf` key yields `grpc_admin_api_key: "" -> "<secret>"` on stdout. Meanwhile the doc rewrite removes the one operator-facing sentence that mentioned logs — `docs/references/services/rpc_reference.md:435` previously read "…not exposed in logs or configuration files in production environments" and now reads "…so the key is not exposed in transit". The new "(never logged)" parentheticals (`rpc_reference.md:431`, `docs/topics/services/rpc.md:141`, `docs/references/services/legacy_reference.md:257`) are, to their credit, correctly scoped to the *generated* key and are accurate as written — but they are the only mention of logging left, and they read to an operator as a property of the setting.
**Why it matters:** Item 4 of this PR fixes secret-in-logs for the generated key, which nobody can use anyway, while the configured shared secret — the one that actually grants admin access — is still printed at `Info` on every startup. Deleting the "not exposed in logs" warning in the same change moves operator guidance the wrong way on a PR whose entire subject is secrets hygiene for this one setting.
**Fix:** The minimum bar for merge is the documentation half, since the leak itself lives in shared startup plumbing rather than this diff: restore an explicit operator-facing warning that a configured key appears in the node's startup log, and tighten the parenthetical so it names the generated key rather than implying the setting is never logged. Also drop the now-purposeless bare declaration — `getString("grpc_admin_api_key", "", ...)` already defaults to empty, so the line buys nothing and only feeds the override echo:
```diff
-grpc_admin_api_key =
+# grpc_admin_api_key =   <- set via environment variable or secret store, never here
```
Be aware that this last change is necessary but **not** sufficient: it silences path (2) only, and the `.test` overrides below it keep the key in gocore's map, so `Stats()` still prints the resolved value. The actual fix is to honour `redact:"true"` in the gocore startup and override logging (it currently leaks `rpc_pass` and `p2p_private_key` identically) — I am not requiring that in this PR, but please file it rather than leaving it implied.

## Non-blocking issues

### ChiR3 - The docker e2e override re-commits a world-readable key that the new guard blesses
**Problem:** `settings.conf:679-681` commits `docker-e2e-test-admin-key` for `docker.teranode{1,2,3}.test`, and `util/admin_api_key_test.go:122` pins it as explicitly *not* a placeholder. Those contexts are not only e2e: `compose/docker-compose-3blasters.yml:151,222,293` uses `docker.teranode1/2/3.test` too. Anyone running those stacks with the p2p gRPC port reachable has the exact #&NoBreak;4738 exploit chain back, and the new startup check cannot catch it by construction.
**Why it matters:** The PR's premise is that a committed value in a public repo is worthless as a secret; the fix keeps that true for `testkey` while carving out an exception the denylist will never see.
**Fix:** Interpolate it from the environment instead — `settings.conf` already does this for ports (`${HEALTH_CHECK_PORT}`), and gocore's `replaceVariables` handles it:
```
grpc_admin_api_key.docker.teranode1.test = ${TERANODE_E2E_ADMIN_KEY}
```
with the value supplied by the compose files / `test/.env`. Nothing secret-shaped stays in the repo, and the e2e stack still agrees on one key.

### ChiR4 - `handleClearBanned` reports the wrong error class and hides the real cause
**Problem:** `services/rpc/handlers.go:2323-2328` returns `ErrRPCInvalidParameter` (-8) with `"Failed to clear banned list"`. `clearbanned` takes no parameters, so -8 is the wrong class (`ErrRPCMisc`/-1 is the general-purpose code, `services/rpc/bsvjson/jsonrpcerr.go:33`), and the underlying error is discarded — it only reaches a `Warnf`. Note this is now the *default* path, not an edge case: with the key empty, the p2p leg fails `Unauthenticated` (`ClearBanned` is protected, `services/p2p/Server.go:1024`) while the Legacy leg succeeds because Legacy protects only `BanPeer`/`UnbanPeer` (`services/legacy/Server.go:671-674`). So every default deployment now gets a bare "invalid parameter" from `clearbanned`, and `setban` fails the same way via `handlers.go:2490-2500`, with no hint that the cause is an unset `grpc_admin_api_key`. Separately, the all-legs-must-succeed rule here diverges from `handleSetBan`'s at-least-one-succeeded rule; both are defensible but the difference isn't recorded in the RPC reference.
**Why it matters:** This is the one place an operator actually observes the fail-closed default, and it tells them nothing actionable — turning a config problem into a debugging session.
**Fix:** Collect the leg errors and surface them:
```go
if !attempted || len(failures) > 0 {
    return nil, &bsvjson.RPCError{
        Code:    bsvjson.ErrRPCMisc,
        Message: fmt.Sprintf("Failed to clear banned list: %s (if this is Unauthenticated, grpc_admin_api_key is not configured)", strings.Join(failures, "; ")),
    }
}
```
and add a line to `docs/references/services/rpc_reference.md` stating that `clearbanned` requires *all* available ban legs to succeed while `setban` requires one.

### ChiR5 - Denylist with no entropy floor: the near-miss class stays open
**Problem:** `knownPlaceholderAPIKeys` (`util/admin_api_key.go:15-27`) is exact-match, so `test-key`, `testkey1`, `changeme123`, `teranode`, `abc123` and a one-character key all pass `ValidateAdminAPIKey`. `cmd/diagnose/config_checks.go:260` already flags anything under 16 chars as WARN, but that's an opt-in CLI that never runs at startup.
**Why it matters:** The most likely operator response to the ChiR1 startup failure is a minimal edit to the offending value — `testkey` → `testkey1` — which the denylist waves through with no signal at all. A length floor is a generic control that covers the whole class a denylist structurally cannot.
**Fix:** Add a floor alongside the denylist, warn-only if you want to avoid another availability cliff:
```go
const minAdminAPIKeyLen = 16
if len(trimmed) < minAdminAPIKeyLen {
    logger.Warnf("[%s] grpc_admin_api_key is only %d chars; use a strong random secret of at least %d (32+ recommended)", serviceName, len(trimmed), minAdminAPIKeyLen)
}
```

### ChiR6 - `isLoopbackListenAddress` misses bracketed IPv6 without a port
**Problem:** `util/admin_api_key.go:79-95`: for `"[::1]"` (no port) `net.SplitHostPort` errors, so `host` becomes the literal `"[::1]"`, `net.ParseIP` returns nil, and the function reports non-loopback — a spurious cleartext warning. The test table covers `"::1"` and `"[::1]:9904"` but not the bracketed-no-port form (`util/admin_api_key_test.go:143-146`). An empty listen address also warns even though no listener exists.
**Why it matters:** Warning-only, so this is log noise rather than a security gap — but a warning that fires on a correctly loopback-bound listener trains operators to ignore it.
**Fix:** Strip the brackets on the no-port path and add the case to the table:
```go
host = strings.TrimSuffix(strings.TrimPrefix(host, "["), "]")
```

### ChiR7 - The setting's own longdesc still claims the key protects "get peer info"
**Problem:** `settings/interface.go:61` — the line this PR edits — keeps "Protects operations like ban peer, unban peer, and get peer info." in its `longdesc`. The protected map covers only the seven mutating RPCs (`services/p2p/Server.go:1020-1030`); the peer-info reads `GetPeers`, `GetPeerRegistry` and `GetPeer` (`services/p2p/p2p_api/p2p_api.proto:315,346,352`) are not in it, and Legacy protects only `BanPeer`/`UnbanPeer` (`services/legacy/Server.go:671-674`). To be clear, I'm not questioning that those reads are open — that classification was settled deliberately in [bsv-blockchain/teranode#1257](https://github.com/bsv-blockchain/teranode/pull/1257) — only that this string still says otherwise. The PR fixed exactly this inaccuracy in `docs/references/services/rpc_reference.md:429`, so it's a missed spot rather than a disagreement.
**Why it matters:** This `longdesc` is the source for the generated settings reference, so an operator sizing up whether the p2p gRPC port can be exposed reads that peer-info queries are gated when they are in fact reachable unauthenticated.
**Fix:** Match the wording you already landed in the RPC reference:
```
Protects state-mutating peer-policy operations (ban, unban, clear bans, ban score,
reputation reset, connect/disconnect peer). Read-only peer queries are intentionally
unauthenticated.
```

### ChiR8 - `TestValidateAdminAPIKeyMatchesServerEmptyCheck` doesn't test what it claims
**Problem:** `util/admin_api_key_test.go:163-168` asserts `strings.TrimSpace("   ") == ""` — a tautology about the stdlib, not about the invariant in the comment (that `settings.NewSettings` trims at `settings/settings.go:87` and both servers compare the trimmed value against `""` at `services/p2p/Server.go:844` / `services/legacy/Server.go:657`). The comment also refers to "finding 6", which is a numbering scheme from outside the repo and means nothing to a future reader.
**Why it matters:** The test name promises a guard against the server/loader normalization drifting apart, so a reader will trust it and it won't actually fail if that drift happens.
**Fix:** Assert against the real loader, and drop the external reference. `getString` goes straight to `gocore.Config().Get`, which checks the environment first (`gocore/config.go:620-627`), so this works as a `settings`-package test (keeping it out of `util` also avoids the import cycle):
```go
t.Setenv("grpc_admin_api_key", "   ")
require.Equal(t, "", NewSettings().GRPCAdminAPIKey)
```
Or simply rename it to describe what it actually does.

## Recap

| ChiR1 | Placeholder hard-fail needs migration path | YES | 80% |
| ChiR2 | Configured key logged at startup | YES | 70% |
| ChiR3 | Committed docker e2e admin key |  | 50% |
| ChiR4 | clearbanned error hides real cause |  | 45% |
| ChiR5 | No minimum key length |  | 35% |
| ChiR6 | Bracketed IPv6 loopback miss |  | 15% |
| ChiR7 | longdesc overstates protected methods |  | 12% |
| ChiR8 | Tautological normalization test |  | 10% |

services/rpc/handlers.go
```diff
@@ -2299,6 +2308,19 @@ func handleClearBanned(ctx context.Context, s *RPCServer, cmd interface{}, _ <-c
 
 		if err != nil {
 			s.logger.Warnf("Failed to clear banned list in legacy peer service: %v", err)
+		} else {
+			success = true
```
freemans13: **This does not actually mirror `handleSetBan` — the legacy response's `Ok` field is never checked.**

The comment above says "Mirror handleSetBan: as long as one leg clears, report success", but the two functions treat the legacy response differently. `handleSetBan` (line ~2438) does:

```go
if err != nil {
    ...
} else if resp == nil || !resp.Ok {
    s.logger.Warnf("Legacy peer service did not apply ban for %s", c.IPOrSubnet)
} else {
    success = true
}
```

whereas here the response is discarded at line 2305 (`_, err := ...`) and any non-error counts as success.

`ClearBannedResponse` does carry an `Ok bool` (`services/legacy/peer_api/peer_api.pb.go:751`).

To be straight about the severity: this cannot misfire today, because the legacy server implementation returns `Ok: true` unconditionally (`services/legacy/Server.go:467-470`). So it is a latent trap rather than a live bug. But the moment `ClearBanned` grows a failure path that reports `Ok: false`, clearbanned will report success while the ban list was not cleared — which is exactly the silent-success class this PR exists to eliminate. The `_,` discard was pre-existing; counting it as success is new, which is why I am flagging it here.

Suggest capturing the response and matching the setban shape:

```go
resp, err := s.legacyP2PClient.ClearBanned(legacyCtx, &emptypb.Empty{})
cancel()

if err != nil {
    s.logger.Warnf("Failed to clear banned list in legacy peer service: %v", err)
} else if resp == nil || !resp.Ok {
    s.logger.Warnf("Legacy peer service did not clear the ban list")
} else {
    success = true
}
```

freemans13 (commented):
Good PR, and I think it should land. The hole is real — `settings.conf` shipped `grpc_admin_api_key = testkey` in a public repo, so the auth interceptor from #1257 was protecting nothing on every default deployment — and the fix keeps server and client consistent rather than just deleting the value. The write-up already anticipates most of what a reviewer would raise, including the gocore residual, which I appreciate being documented rather than papered over.

I checked out the branch and verified locally rather than reading the diff alone. Everything is green: `go build ./...` clean, `go vet` clean across all seven touched packages, and `golangci-lint run ./... --new-from-rev 537ab3638` (the Makefile's own invocation, pinned to this branch's parent so it lints only your commits) reports **0 issues**. Targeted tests all pass — `./util/` (10 + 16 + 2 subtests), `./cmd/diagnose/...`, `./settings/...`, and `./services/rpc/ -run 'TestHandleClearBanned|TestHandleSetBan'` (22 subtests). The new tests are behavioural and use a capturing logger to prove the warnings actually fire, which is the right shape.

Two caveats on that verification, in fairness: I did not run `make lint` verbatim, because it does `git fetch origin main` and `origin/main` was 25+ commits stale in my worktree, so it would have linted a large unrelated surface. And I could not execute the docker or daemon e2e suites (no stack running), so my conclusion that `TestBanListRPCE2E` and `TestBanPeerList` stay covered by the new compose env vars is static reasoning, not a test result.

Three of your claims I verified independently rather than taking on trust, and all three hold:

1. The compose env-var route genuinely works — gocore's `getInternal` (config.go:621) does `os.LookupEnv(key)` on the raw, exact-case key *before* consulting any config file, so the override really does keep the value out of the shared `docker.teranodeN.test` context.
2. The "configured key is echoed in the startup dump" warning is accurate, not defensive boilerplate. `PrintSettings` calls `settings.Redact()` for the SETTINGS JSON block and `GRPCAdminAPIKey` carries `redact:"true"`, so that half is genuinely clean — but the line above it prints `gocore.Config().Stats()`, and gocore's `maskSecrets` (config.go:57) masks *only* its own `*EHE*`-encrypted values. A plain key prints verbatim. If it helps the follow-up issue: `settings/redact.go` already has an `extractSensitiveKeys()` walker that would make masking the STATS block by key name fairly cheap.
3. The `securityLevel <= 1` reasoning matches `loadTLSCredentials` exactly — level 0 is `insecure.NewCredentials()`, level 1 on the client is `InsecureSkipVerify: true`. Treating level 2+ as safe is right.

The documented list of protected P2P methods also matches `adminProtectedMethods()` item for item.

I have left four inline comments. The one I would most like a response to before merge is the deployment gap on `settings.conf` — it is an operational consequence rather than a coding bug, but I think it needs either a commented-out placeholder line in the deploy files or an explicit note in the release notes. The `handleClearBanned` one is a genuine (if latent) correctness defect. The other two are small.

Two minor notes that did not need their own anchor:

- **Warning noise on a stock node.** With no key configured — now the default — a default node emits four warnings at startup (two servers, two clients), and `cmd/monitor` and `cmd/diagnose` both build p2p clients and will warn on every invocation even though neither calls an admin RPC. That is a fair bit of alarm for what is the *intended* state, and it risks operators "fixing" it by setting a key they do not need. Worth considering warning only from paths that actually attempt an admin RPC.
- **A stale `settings_local.conf` will fail the make targets.** The placeholder path deliberately logs at `Error`, and `Makefile:639` (and the equivalent at 776) greps container logs for `| ERROR |` and returns 1 on any hit. Anyone with a leftover `testkey` locally will see those targets fail on a log grep with a fairly confusing message. Narrow, but easy to trip over.

One last thing, clearly out of scope for this PR so I am not asking you to touch it: while tracing the alert ban path I noticed `services/alert/node.go:207` reads `if !legacyResp.Ok { banned = true }`, which looks inverted (and `banned` is unreachable in the false case anyway, since the function returns early on error). Pre-existing, untouched by this branch, but it may deserve its own issue.

settings.conf
```diff
@@ -664,8 +664,18 @@ gocore_stats_reported_time_threshold = 24h
 stats_prefix = gocore
 # @endgroup
 
-# TODO: change api key and move out of settings
-grpc_admin_api_key = testkey
+# Admin API key for protected gRPC admin RPCs (peer ban/unban, connect, etc.).
```
freemans13, line 683: **The fail-closed default silently disables alert-driven peer banning, and the operator deploy files were not given anywhere to put a key.**

This is the one I would most like a response to before merge. It is an operational consequence of the design rather than a coding bug, so I am not arguing the default should change — but I think it needs to be surfaced somewhere an upgrading operator will see it.

The alert service bans peers through the same admin-key-protected gRPC path. `daemon/daemon_services.go:578` builds it a legacy `peer.NewClient`, and `services/alert/node.go:194` and `:202` call `p2pClient.BanPeer` / `peerClient.BanPeer` — both of which `return err` immediately on failure. Before this PR the committed `testkey` meant server and client agreed, so alert-driven bans worked out of the box everywhere. After it, any deployment that does not supply a key gets a randomly generated server key against a client sending nothing, and alert bans fail with `Unauthenticated`.

I swept every file in the repo that starts a teranode process. Only the two e2e compose files you updated set a key:

| File | Starts teranode | Sets `grpc_admin_api_key` |
|---|---|---|
| `test/docker-compose.e2etest.yml` | yes | **yes** (3 nodes) |
| `test/docker-compose.e2etest.legacy.yml` | yes | **yes** (3 nodes) |
| `deploy/docker/mainnet/docker-compose.yml` | yes | no |
| `deploy/docker/testnet/docker-compose.yml` | yes | no |
| `deploy/docker/base/docker-teranode.yml` | yes (12 per-service containers) | no |
| `deploy/kubernetes/teranode/teranode-configmap.yaml` (+ CRs, operator) | yes | no |
| `compose/docker-compose-3blasters.yml` | yes | no |
| `compose/docker-compose-chainintegrity.yml` | yes | no |
| `compose/docker-compose-ss.yml`, `.localBinary.override.yml`, `test/docker-compose-host.yml` | yes | no |

`grep -rn grpc_admin_api_key deploy/ compose/ .github/` returns nothing.

The `compose/` stacks are harmless — I checked, they only generate blocks and verify chain integrity, and never call a ban RPC. The `deploy/` ones are the concern: `deploy/docker/base/docker-teranode.yml` runs each service in its own container, so there the key genuinely has to be *shared across containers*. There is no same-process shortcut and no documented place to inject it.

Suggestion: either add a commented-out `grpc_admin_api_key:` line to the `deploy/docker/{mainnet,testnet}` compose files and the k8s configmap showing operators where it goes, or state explicitly in the PR body and release notes that alert-driven bans require a configured key from this version onward. Worth pointing at `cmd/diagnose` too — `cmd/diagnose/config_checks.go:246-271` now flags a missing or placeholder key, so there *is* a detection path, it just is not named as the upgrade remedy.

services/p2p/Client.go
```diff
@@ -60,7 +60,9 @@ func NewClientWithAddress(ctx context.Context, logger ulogger.Logger, address st
 	// Include the admin API key in the connection options
 	apiKey := tSettings.GRPCAdminAPIKey
 	if apiKey != "" {
-		logger.Infof("[Legacy Client] Using API key for authentication")
+		logger.Infof("[P2P Client] Using API key for authentication")
```
freemans13: **The clients never apply the placeholder check, so their logs actively point away from the cause.**

Same applies to `services/legacy/peer/Client.go:42`.

The servers now detect a placeholder and log an `Errorf`, but the clients still only branch on empty. So an operator with `grpc_admin_api_key = testkey` left in a `settings_local.conf` sees the server log an error, and the client cheerfully log *"Using API key for authentication"* at Info — and then every admin RPC fails with `Unauthenticated`. The reassuring client line is the misleading part: it suggests the key side is fine.

`util.IsPlaceholderAdminAPIKey` is already exported and this file already imports `util`, so it is a small change — treat a placeholder the same as empty in both client constructors:

```go
if apiKey == "" || util.IsPlaceholderAdminAPIKey(apiKey) {
    logger.Warnf("[P2P Client] grpc_admin_api_key is unset or a well-known placeholder; admin RPCs (ban, unban, connect/disconnect peer) will fail with Unauthenticated because the server ignores placeholders and uses a random key")
} else {
    logger.Infof("[P2P Client] Using API key for authentication")
}
```

(`services/legacy/peer/Client.go` would need the `util` import adding.)

services/rpc/handlers.go
```diff
@@ -2299,6 +2308,19 @@ func handleClearBanned(ctx context.Context, s *RPCServer, cmd interface{}, _ <-c
...
+		}
+	}
+
+	// No applicable clear leg succeeded (every attempted leg failed - e.g. the
+	// admin API key is unset or wrong so calls are Unauthenticated - or none was
+	// available). clearbanned is an administrative control, so report the failure
+	// rather than silently returning success.
+	if !success {
+		return nil, &bsvjson.RPCError{
+			Code:    bsvjson.ErrRPCInvalidParameter,
```
freemans13: Minor, and I would not block on it: `ErrRPCInvalidParameter` is the wrong code for this failure. Nothing about the caller's parameters was invalid — the backend was unreachable or unauthenticated. `ErrRPCInternal` or `ErrRPCMisc` fits better.

The argument for leaving it is consistency: the existing `handleSetBan` uses the same code for its equivalent all-legs-failed case, and matching it is defensible. But this PR is copying that code into a new place, so it seemed worth one line either way. Your call.

ctnguyen (changes requested):
Review at commit [cd3fd0b](https://github.com/gokutheengineer/teranode/commit/cd3fd0beec53b082876ab9d24ec6af7b9b59183d)

The two blocking points from last round are properly closed, and closed in the right way. `ValidateAdminAPIKey` now returns an `ignoreKey` bool instead of an error (`util/admin_api_key.go:56-72`), so a stale `testkey` costs an `Errorf` and the fail-closed random-key path rather than the node's consensus participation — that was the availability-vs-security trade I was worried about, and you landed on the correct side of it. The docs half of the logging point is done properly too: the "configured key is echoed in the startup settings dump" warning is now in all four operator-facing places, the "(never logged)" wording stays scoped to the *generated* key, and the bare declaration is gone from `settings.conf`. Moving the e2e key into the compose `environment:` blocks is a cleaner answer than the `${VAR}` interpolation I suggested, and your reasoning about `{UNKNOWN}` is right. What's left below is mostly the ground level of shipping a fail-closed default: the deployment surfaces that operators actually read never mention that the key is now required, and the `handleClearBanned` rewrite has a hole its own new tests don't catch.

## Status of previous points

ChiR4 changed shape rather than being fixed: the all-legs-must-succeed rule I flagged as diverging from `handleSetBan` is now the any-leg-succeeds rule, but the two things the item actually asked for — a sensible error class and surfacing the leg error — are untouched (`services/rpc/handlers.go:2320-2325` still returns `ErrRPCInvalidParameter` with a bare `"Failed to clear banned list"`, and the leg errors only reach `Warnf`). [@freemans13](https://github.com/bsv-blockchain/teranode/pull/1575#discussion_r3797033158) raised the error-code half independently. ChiR8 lost the stray "finding 6" reference from its comment, but `util/admin_api_key_test.go:175` still asserts `strings.TrimSpace("   ") == ""` and the test still never touches `settings.NewSettings` or either server's empty check, which is what its name promises. ChiR5, ChiR6 and ChiR7 are unchanged in the code: no length floor in `ValidateAdminAPIKey`, `isLoopbackListenAddress("[::1]")` still reports non-loopback (`util/admin_api_key.go:83-87`, and the test table at `:145-160` still has no bracketed-no-port case), and `settings/interface.go:61` still says the key "Protects operations like ban peer, unban peer, and get peer info."

| ChiR1 | Placeholder hard-fail | YES | FIXED |
| ChiR2 | Configured key logged at startup | YES | FIXED |
| ChiR3 | Committed docker e2e admin key |  | FIXED |
| ChiR4 | clearbanned error hides cause |  | PARTIAL FIXED |
| ChiR5 | No minimum key length |  | NOT FIXED |
| ChiR6 | Bracketed IPv6 loopback miss |  | NOT FIXED |
| ChiR7 | longdesc overstates protected methods |  | NOT FIXED |
| ChiR8 | Tautological normalization test |  | PARTIAL FIXED |

## Blocking issues

### ChiR9 - No shipped deployment surface says the key is now required, so cross-container admin RPCs break on upgrade

*Same as [@freemans13's point "the operator deploy files were not given anywhere to put a key" (17-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1575#discussion_r3797033156).*

**Problem:** To be precise about what is and isn't missing: the injection points exist and work. `deploy/docker/base/docker-teranode.yml:9` mounts `./settings_local.conf` into every container, `deploy/docker/{mainnet,testnet}/docker-compose.yml` carries a shared `environment:` anchor, and `deploy/kubernetes/teranode/teranode-configmap.yaml:15-19` already documents an uncommitted `teranode-operator-secrets` Secret for exactly this class of value. What's missing is that none of them says `grpc_admin_api_key` belongs there — the string appears nowhere under `deploy/` or `compose/`. The functional consequence is not a startup failure; the stack still comes up. It is that `deploy/docker/base/docker-teranode.yml` runs each service in its own container (`peer` at `:153` with `-p2p=1`, `legacy` at `:51`, `rpc` at `:172`), so `setban`/`clearbanned` issued at the JSON-RPC port genuinely cross the auth interceptor. With no shared key the `rpc` container sends none while the `peer` container has minted its own random one, every leg is rejected, and `handleSetBan` returns `"Failed to apply ban"` (`services/rpc/handlers.go:2487-2497`). The same holds for `services/asset/httpimpl/reset_reputation.go:51`, which reaches the protected `ResetReputation` over the asset HTTP API, and for `services/alert/node.go:194-199`, which `return err`s the whole alert ban as soon as the p2p leg is rejected. Before this PR the committed `testkey` made all three work by accident.

**Why it matters:** The break is functional and shipped, but the remedy is guidance, which is why it belongs in this PR rather than a follow-up — the operator has no way to learn from the repo that a previously working command now needs a shared secret. The detection path even exists and goes unnamed: `cmd/diagnose/config_checks.go:246-252` reports the empty key, but nothing in the diff points an upgrading operator at it.

**Fix:** Add a commented pointer to each of the three surfaces and name the remedy in the release notes, so the fail-closed default is discoverable.

```conf
# deploy/docker/base/settings_local.conf
# Shared secret for the protected gRPC admin RPCs (setban/clearbanned, connect/disconnect
# peer, alert-driven bans, reset reputation). Required because these services run in
# separate containers. Unset = admin RPCs return Unauthenticated. Use 32+ random chars.
# grpc_admin_api_key.docker.m = <32+ char random secret>
```

```yaml
# deploy/kubernetes/teranode/teranode-configmap.yaml — extend the existing Secret note
  # grpc_admin_api_key is a shared secret and is NOT stored here; put it in the
  # "teranode-operator-secrets" Secret alongside blockchain_store/utxostore.
  # Without it, setban/clearbanned and alert-driven peer bans fail Unauthenticated.
```

### ChiR10 - `handleClearBanned` counts a legacy `Ok: false` as a successful clear, and every new test asserts that behaviour

*Same as [@freemans13's point "the legacy response's `Ok` field is never checked" (17-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1575#discussion_r3797033152).*

**Problem:** `services/rpc/handlers.go:2305` discards the response (`_, err := s.legacyP2PClient.ClearBanned(...)`) and sets `success = true` on any nil error, while `handleSetBan` — the function the comment at `:2287-2288` says it mirrors — does check it (`:2439`, `resp == nil || !resp.Ok`). The asymmetry is sharper than it looks, because the *p2p* leg is `Ok`-checked one layer down: `services/p2p/Client.go:243-245` turns `!resp.Ok` into an error. So the legacy leg is the only one where a "did not clear" response reads as a clear. Extending @freemans13's point: the new tests don't just fail to cover this, they certify the opposite. Every legacy mock in `TestHandleClearBannedComprehensive` returns `&peer_api.ClearBannedResponse{}` — zero value, `Ok: false` (`handlers_additional_test.go:3576`, `:3606`, `:3656`). The subtest renamed to `"one leg fails but the other succeeds - reports success"` (`:3597-3628`) therefore has no succeeding leg at all: p2p errors, legacy answers `Ok: false`, and the assertion `require.True(t, success)` passes only because the field is ignored. `"only legacy client available"` (`:3653-3674`) is the same. Applying the obvious fix would turn those two tests red, which is the signal that they are currently pinned to the wrong contract.

**Why it matters:** Item 5 of this PR exists to stop `clearbanned` reporting success when nothing was cleared. As written, the one leg that can report a soft failure is the one whose soft failure is dropped — and the test suite now locks that in, so if `Server.ClearBanned` (`services/legacy/Server.go:467-470`, currently unconditional `Ok: true`) ever grows a real failure path, nothing fails. Today it's latent; the tests are what make it durable.

**Fix:** Capture the response and match the `handleSetBan` shape, then give the mocks `Ok: true` where the subtest name says the leg succeeded and add an `Ok: false` case asserting failure.

```go
resp, err := s.legacyP2PClient.ClearBanned(legacyCtx, &emptypb.Empty{})
cancel()

if err != nil {
    s.logger.Warnf("Failed to clear banned list in legacy peer service: %v", err)
} else if resp == nil || !resp.Ok {
    s.logger.Warnf("Legacy peer service did not clear the ban list")
} else {
    success = true
}
```

While you are in there, the mocks are handed `settings.Settings{ChainCfgParams: ...}` with a zero `RPC.ClientCallTimeout`, so `context.WithTimeout(ctx, 0)` yields an already-expired context that only passes because the mocks ignore `ctx` — worth setting a real timeout so the legacy leg is exercised the way production runs it.

## Non-blocking issues

### ChiR11 - The client constructors never apply the placeholder check, so their logs point away from the cause

*Same as [@freemans13's point "The clients never apply the placeholder check" (17-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1575#discussion_r3797033157).*

**Problem:** `services/p2p/Client.go:62` and `services/legacy/peer/Client.go:41` still branch on `apiKey != ""` alone. An operator with `grpc_admin_api_key = testkey` left in a `settings_local.conf` now gets a server `Errorf` saying the key is ignored, and — in the same log — a client `Infof` saying *"Using API key for authentication"*, followed by `Unauthenticated` on every admin RPC. The reassuring line is the misleading one. One correction to @freemans13's suggested patch: `services/legacy/peer/Client.go` already imports `util` (line 10, for `util.GetGRPCClient`), so no import change is needed on either side — this is a two-line edit in each file.

**Why it matters:** The placeholder branch is now the only state where server and client disagree about what the setting means, and it is exactly the state an upgrading operator lands in. Diagnosability is the whole point of choosing ignore-and-log over hard-fail; a client log that contradicts the server log spends that benefit.

**Fix:** Treat a placeholder as unset in both constructors.

```go
if apiKey == "" || util.IsPlaceholderAdminAPIKey(apiKey) {
    logger.Warnf("[P2P Client] grpc_admin_api_key is unset or a well-known placeholder; admin RPCs (ban, unban, connect/disconnect peer) will fail with Unauthenticated because the server ignores placeholders and uses a random key")
} else {
    logger.Infof("[P2P Client] Using API key for authentication")
}
```

### ChiR12 - The p2p reference was left out of the doc pass and never mentions the admin key

**Problem:** `docs/references/services/p2p_reference.md` documents all seven RPCs that `adminProtectedMethods()` gates — `BanPeer` (`:245`), `UnbanPeer` (`:251`), `ClearBanned` (`:269`), `AddBanScore` (`:275`), `ConnectPeer` (`:281`), `DisconnectPeer` (`:287`), `ResetReputation` (`:360`), matching `services/p2p/Server.go:1021-1031` item for item — and never says that any of them requires `grpc_admin_api_key`. Its `## Security` section (`:656-658`) covers only `securityLevelHTTP` and certificate files. Two adjacent notes in the same block are also wrong, though pre-existing rather than introduced here: `:284` and `:290` state that `ConnectPeer` and `DisconnectPeer` are "available on the client only. The server-side gRPC handler is not yet implemented", while `services/p2p/Server.go:2145` and `:2171` implement both — and both sit in the protected set, so the note misses the handler and the auth requirement together.

**Why it matters:** This gap bites harder after this PR than before it. Unset is now the default, so on a stock node every one of those seven calls answers `Unauthenticated`, and the reference a reader consults for the p2p service specifically offers no explanation. The PR already fixed exactly this class of inaccuracy in `rpc_reference.md`, `legacy_reference.md` and `topics/services/rpc.md`, so this is a missed spot in the same pass rather than a disagreement — the same shape as ChiR7.

**Fix:** Extend the `## Security` section, and drop the two stale handler notes while you are in the file.

```markdown
## Security

The server supports both HTTP and HTTPS configurations based on the `securityLevelHTTP` setting.
When using HTTPS, it requires certificate and key files to be specified in the configuration.

State-mutating admin RPCs — `BanPeer`, `UnbanPeer`, `ClearBanned`, `AddBanScore`,
`ResetReputation`, `ConnectPeer` and `DisconnectPeer` — additionally require the
`grpc_admin_api_key` shared secret, supplied as `x-api-key` gRPC metadata and enforced by
`adminProtectedMethods()`. When the key is unset the server generates a random one, so these
calls answer `Unauthenticated` until an operator configures a value that the internal clients
read as well. Read-only peer queries are intentionally unauthenticated.
```

### ChiR13 - The unset-key client warning fires from two CLIs that never call a protected RPC

*Same as [@freemans13's point "Warning noise on a stock node" (17-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1575#pullrequestreview-4952425987).*

**Problem:** `cmd/monitor/monitor.go:185` and `cmd/diagnose/health_checks.go:99` each construct a p2p client for health checking only; neither appears among the callers of the seven RPCs in `adminProtectedMethods()`. With the key unset — the new default — both therefore print `"[P2P Client] grpc_admin_api_key is not set; admin RPCs (ban, unban, connect/disconnect peer) will fail with Unauthenticated"` on every invocation, describing a consequence that cannot apply to the process emitting it. On a node the same constructor warning fires three more times (`daemon/daemon_stores.go:149`, `daemon/daemon_services.go:494` and `:578`) alongside the two server warnings (`services/p2p/Server.go:852`, `services/legacy/Server.go:668`).

**Why it matters:** This is log accuracy, not behaviour — nothing malfunctions. But the constructor has no way to know whether its caller will ever attempt an admin RPC, so rewording the message cannot fix it; the message is in the wrong place.

**Fix:** Log at `Debugf` in the constructors and raise the actionable warning where a protected RPC is actually rejected — the ban legs in `services/rpc/handlers.go` and `services/alert/node.go`.

```go
} else {
    logger.Debugf("[P2P Client] grpc_admin_api_key is not set; admin RPCs will be rejected until a shared key is configured")
}
```

## Recap

| ChiR9 | Deploy surfaces omit required key | YES | 60% |
| ChiR10 | Legacy `Ok:false` counted as cleared | YES | 55% |
| ChiR11 | Clients skip placeholder check |  | 35% |
| ChiR12 | p2p reference omits admin auth |  | 22% |
| ChiR13 | Warning fires from non-admin paths |  | 15% |

ctnguyen (changes requested):
Review at commit [0aa2743](https://github.com/gokutheengineer/teranode/commit/0aa2743f2ec8340c55411e788b862b8003c506b4)

Nearly everything from the last two rounds is closed, and closed the way I'd hoped. The length floor and the bracketed-IPv6 fix landed in `util/admin_api_key.go`, the `longdesc` no longer claims peer-info reads are gated, `settings/admin_api_key_test.go` now actually drives `NewSettings()` instead of asserting a stdlib identity, both client constructors treat a placeholder as unset, and the deployment guidance finally exists in the three surfaces an operator reads (`deploy/docker/base/settings_local.conf:14-18`, its `.template:103-114`, `deploy/kubernetes/teranode/teranode-configmap.yaml:21-25`) — with the base file bind-mounted by `deploy/docker/base/docker-teranode.yml:9`, so mainnet and testnet both pick it up. ChiR10 is done properly too: the legacy `Ok` is checked, the mocks now return `Ok: true` where the subtest name says the leg succeeded, and a real `ClientCallTimeout` is wired in. One thing does need another pass, though, and it's the same silent-success class ChiR10 was about — on a node that runs both the peer and legacy services with no key configured, the any-leg rule that replaced the all-legs rule leaves `clearbanned` reporting success while the p2p ban list is untouched.

## Status of previous points

Two caveats. ChiR6's code fix is in (`util/admin_api_key.go:99-103`) and correct, but the test table (`util/admin_api_key_test.go:150-165`) still has no bare `[::1]` case, so the new bracket-stripping branch is the only path in that function with no coverage — `"::1"` and `"[::1]:9904"` both bypass it. The empty-listen-address warning noted in the same item is also unchanged. ChiR12 and ChiR13 are untouched in the tree: `docs/references/services/p2p_reference.md:656-658` still covers only `securityLevelHTTP` and never mentions the admin key for the seven RPCs it documents, and `services/p2p/Client.go:66` still warns at `Warnf` from constructors that `cmd/monitor/monitor.go:185` and `cmd/diagnose/health_checks.go:99` call without ever touching an admin RPC.

| ChiR1 | Placeholder hard-fail | YES | FIXED |
| ChiR2 | Configured key logged at startup | YES | FIXED |
| ChiR3 | Committed docker e2e admin key |  | FIXED |
| ChiR4 | clearbanned error hides cause |  | FIXED |
| ChiR5 | No minimum key length |  | FIXED |
| ChiR6 | Bracketed IPv6 loopback miss |  | PARTIAL FIXED |
| ChiR7 | longdesc overstates protected methods |  | FIXED |
| ChiR8 | Tautological normalization test |  | FIXED |
| ChiR9 | Deploy surfaces omit required key | YES | FIXED |
| ChiR10 | Legacy Ok false counted as cleared | YES | FIXED |
| ChiR11 | Clients skip placeholder check |  | FIXED |
| ChiR12 | p2p reference omits admin auth |  | NOT FIXED |
| ChiR13 | Warning fires from non-admin paths |  | NOT FIXED |

## Blocking issues

### ChiR14 - With peer and legacy both enabled and the key unset, `clearbanned` reports success while p2p bans survive

**Problem:** The any-leg-succeeds rule masks a *rejected* leg, not just an absent one, and the two legs are asymmetric: `ClearBanned` is in the p2p protected set (`services/p2p/Server.go:1044`) but **not** in the legacy one, which covers only `BanPeer`/`UnbanPeer` (`services/legacy/Server.go:672-675`). The RPC service builds both clients unconditionally (`daemon/daemon_services.go:494,501`) — that is client wiring, not which services run — so the harmful configuration is specifically **peer enabled + legacy enabled + `grpc_admin_api_key` unset**. There, the p2p leg returns `Unauthenticated` at `services/rpc/handlers.go:2298`, the legacy leg sails past its interceptor into `Server.ClearBanned`, which returns `Ok: true` unconditionally (`services/legacy/Server.go:467-470`), `success` is set at `:2329`, and `handleClearBanned` returns `true` with no error: the legacy list is cleared, every p2p ban survives, and the caller is told it worked. Bounding this honestly: on the *shipped* deploy defaults the peer service is off — `deploy/docker/{mainnet,testnet}/docker-compose.yml:71-76` comment `peer:` out, and `deploy/kubernetes/teranode/teranode-cr-mainnet.yaml:107-108` sets `peer: enabled: false` while `:98-99` enables legacy — so there the p2p leg fails `Unavailable`, legacy still answers `Ok: true`, `clearbanned` still reports success, but no p2p ban list exists to be left behind and nothing is silently lost. The configuration that does bite is a node that has enabled the peer service, i.e. the deployment this PR's threat model is about and what the e2e stacks run. The new test at `services/rpc/handlers_additional_test.go:3598-3629` is exactly that scenario — p2p errors, legacy returns `Ok: true` — and asserts success, so the suite pins the behaviour rather than catching it. I traced the code path and the wiring statically; this establishes the mechanism and the configuration that reaches it, not an observed run.

**Why it matters:** In that configuration the same misconfiguration gives two opposite answers: `setban` fails cleanly, because `BanPeer`/`UnbanPeer` are protected on both services (`services/p2p/Server.go:1042-1049`, `services/legacy/Server.go:672-675`), while `clearbanned` reports success — and the misleading answer is on the command that *removes* a security control, so an operator believes bans are gone when the p2p list still refuses those peers. Item 5 of this PR exists to stop exactly that, and ChiR10 closed only the `Ok: false` half. The `#591` justification for any-leg covers a configured-but-absent service *timing out* (see [bsv-blockchain/teranode#591](https://github.com/bsv-blockchain/teranode/issues/591)); it does not cover a leg that answered and refused.

**Fix:** Distinguish "leg absent" from "leg refused". Tolerate `DeadlineExceeded`/`Unavailable` (that is the `#591` case) and fail the command on anything else, so an `Unauthenticated` p2p leg can no longer be papered over by the unprotected legacy leg. Both legs surface the gRPC status unchanged for transport/auth failures (`services/p2p/Client.go:240-250` returns the error raw and only wraps the `!Ok` case), so a status-code check works on both.

```go
// A configured-but-absent service times out or is Unavailable by design (#591);
// tolerate that leg. Any other refusal - notably Unauthenticated when
// grpc_admin_api_key is unset - means a real ban list was left untouched.
func legAbsent(err error) bool {
    switch status.Code(err) {
    case codes.DeadlineExceeded, codes.Unavailable:
        return true
    default:
        return false
    }
}

var refused string
if s.p2pClient != nil {
    if err := s.p2pClient.ClearBanned(ctx); err != nil {
        if !legAbsent(err) {
            refused = err.Error()
        }
        s.logger.Warnf("Failed to clear banned list in P2P service: %v", err)
    } else {
        success = true
    }
}
// ... same shape for the legacy leg ...
if !success || refused != "" {
    return nil, &bsvjson.RPCError{Code: bsvjson.ErrRPCInternal.Code, Message: msg}
}
```

Then retarget `"one leg fails but the other succeeds - reports success"` at the case it is meant to cover (legacy `DeadlineExceeded`, p2p clears) and add a `codes.Unauthenticated` p2p leg + legacy `Ok: true` case asserting failure. A simpler variant if you'd rather not classify errors: require the p2p leg to succeed whenever `s.p2pClient != nil`, keeping the any-leg tolerance for the legacy leg only.

## Non-blocking issues

### ChiR15 - `setban` keeps the wrong error class and still drops the cause, so the two handlers now disagree

*Same as [@freemans13's point (17-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1575#discussion_r3797033158).*

**Problem:** Extending @freemans13's point, which was about `clearbanned`: that half is now fixed (`services/rpc/handlers.go:2343-2346` returns `ErrRPCInternal` and appends the leg reason), but `handleSetBan` was left as-is and still returns `ErrRPCInvalidParameter` with a bare `"Failed to apply ban"` / `"Failed to remove ban"` (`:2509-2519`), discarding the leg errors into `Warnf` at `:2442`, `:2460`, `:2474`, `:2491`. So the two sibling handlers now use different codes for the same "no leg applied it" failure, and the consistency argument that justified leaving `clearbanned` alone has been inverted rather than resolved. This is also the path an upgrading operator hits first: unlike `clearbanned`, `setban` fails cleanly on the new default because `BanPeer`/`UnbanPeer` are protected on *both* services, so it is the command that will actually be reported as broken — with a code that says the caller's parameters were wrong and no mention of `Unauthenticated`.

**Why it matters:** The whole reason for choosing ignore-and-log over hard-fail was diagnosability; a `-8 invalid parameter` on a well-formed `setban` spends that, and an operator has no thread to pull.

**Fix:** Give `handleSetBan` the same treatment, tracking the leg reason as `handleClearBanned` now does:

```go
if !success {
    msg := "Failed to apply ban"
    if c.Command == "remove" {
        msg = "Failed to remove ban"
    }
    if lastErr != "" {
        msg = fmt.Sprintf("%s: %s", msg, lastErr)
    }

    return nil, &bsvjson.RPCError{Code: bsvjson.ErrRPCInternal.Code, Message: msg}
}
```

### ChiR16 - `lastErr` keeps the last leg's reason, so the actionable one is overwritten

**Problem:** `services/rpc/handlers.go:2293` holds a single `lastErr`, assigned by the p2p leg at `:2300` and then overwritten by the legacy leg at `:2318`/`:2325`. When both legs fail, the reported message is whichever ran last — typically the generic `"legacy peer service reported the ban list was not cleared"` — while the p2p `Unauthenticated` that names the actual cause is dropped. The comment at `:2288-2290` says the intent is to "name the cause instead of a bare failed", which the last-write-wins choice undercuts precisely when there is more than one cause.

**Why it matters:** Minor on its own, but it compounds ChiR14: once a refused leg does produce an error, the message an operator sees should be the one that points at `grpc_admin_api_key`.

**Fix:** Collect per-leg reasons and join them, as suggested in ChiR4's original patch:

```go
var failures []string
// ... failures = append(failures, fmt.Sprintf("p2p: %v", err)) etc.
if len(failures) > 0 {
    msg = fmt.Sprintf("%s: %s", msg, strings.Join(failures, "; "))
}
```

### ChiR17 - The k8s Secret note points at an install doc that does not list the key

**Problem:** `deploy/kubernetes/teranode/teranode-configmap.yaml:21-25` now tells operators to put `grpc_admin_api_key` in the `teranode-operator-secrets` Secret, and `:19` — two lines above — refers them to `docs/howto/miners/kubernetes/minersHowToInstallation.md` ("Create the Teranode Secret") for how to create it. That section (`:170-195`) still enumerates only `blockchain_store` and `utxostore`, both in prose at `:172-176` and in the copy-paste manifest at `:180-191`. No file under `docs/howto/` mentions `grpc_admin_api_key` at all.

**Why it matters:** An operator following the documented procedure — which is a manifest to copy — creates a Secret without the key and lands on exactly the broken admin RPCs ChiR9 was about; the configmap comment is a cross-reference to a procedure that contradicts it.

**Fix:** Add the key to the manifest example and to the surrounding sentence:

```yaml
stringData:
  blockchain_store: "postgres://POSTGRES_EXAMPLE_URI_CHANGE_ME"
  utxostore: "aerospike://AEROSPIKE_EXAMPLE_URI_CHANGE_ME"
  # Shared secret for the protected gRPC admin RPCs (setban/clearbanned,
  # connect/disconnect peer, alert-driven bans, reset reputation). Omit it and
  # those RPCs fail Unauthenticated. Use 32+ random characters.
  grpc_admin_api_key: "GENERATE_A_32_PLUS_CHAR_RANDOM_SECRET"
```

## Recap

| ChiR14 | clearbanned false success with peer enabled | YES | 52% |
| ChiR15 | setban error class and cause |  | 35% |
| ChiR16 | Last leg reason overwrites first |  | 25% |
| ChiR17 | k8s Secret doc omits the key |  | 20% |

ctnguyen (approved):
Review at commit [73dd64d](https://github.com/gokutheengineer/teranode/commit/73dd64d904cb320bbfbf57e17034b55750b227ec)

The absent-vs-refused split landed exactly as asked, and it landed well: `legAbsent` keys on `codes.DeadlineExceeded`/`codes.Unavailable` only (`services/rpc/handlers.go:2374-2385`), both legs surface the raw gRPC status so the check actually works on them (`services/p2p/Client.go:240-250` returns the transport error unwrapped, and the retry interceptor at `util/grpc_helper.go:352-383` passes it through), and the test that used to pin the bug is now a regression guard for it (`handlers_additional_test.go` "p2p refuses (Unauthenticated) while legacy clears - reports failure"). `handleSetBan` picked up `ErrRPCInternal` and the leg reason too. With ChiR14 and ChiR15 closed, every blocking point from the three earlier rounds is done, and nothing below needs to hold the merge. I also checked @freemans13's [stale-`settings_local.conf` point](https://github.com/bsv-blockchain/teranode/pull/1575#pullrequestreview-4952425987) against the tree: the stacks behind the `| ERROR |` log grep at `Makefile:629` and `:766` mount `settings.conf` plus `compose/settings_test.conf`, and neither resolves a placeholder any more, so that target only trips for someone who deliberately injects one. What is left is four refinements, three of them in the leg-outcome plumbing this PR now depends on.

## Status of previous points

Four items are untouched in the tree and one is still half-done. ChiR6's bracket-stripping branch (`util/admin_api_key.go:99-103`) is correct but remains the only unexercised path in that function - the table at `util/admin_api_key_test.go:150-166` still has no bare `"[::1]"` case (`"::1"` and `"[::1]:9904"` both bypass the branch), and an empty listen address still warns. ChiR12: `docs/references/services/p2p_reference.md:656-658` still covers only `securityLevelHTTP`, and `:284`/`:290` still claim the `ConnectPeer`/`DisconnectPeer` server handlers are unimplemented. ChiR13: `services/p2p/Client.go:62-68` still warns at `Warnf` from constructors that `cmd/monitor/monitor.go:185` and `cmd/diagnose/health_checks.go:99` call without ever touching an admin RPC. ChiR16: `handlers.go:2301` is still a single `lastErr` written by both legs (`:2311`, `:2333`), and the new `refused` variable has the same last-write-wins shape (`:2315`, `:2337`, `:2344`). ChiR17: `docs/howto/miners/kubernetes/minersHowToInstallation.md:172-176` and the copy-paste manifest at `:188-190` still list only `blockchain_store` and `utxostore`, so the cross-reference at `teranode-configmap.yaml:19` still points at a procedure that contradicts the note above it.

| ChiR1 | Placeholder hard-fail | YES | FIXED |
| ChiR2 | Configured key logged at startup | YES | FIXED |
| ChiR3 | Committed docker e2e admin key |  | FIXED |
| ChiR4 | clearbanned error hides cause |  | FIXED |
| ChiR5 | No minimum key length |  | FIXED |
| ChiR6 | Bracketed IPv6 loopback miss |  | PARTIAL FIXED |
| ChiR7 | longdesc overstates protected methods |  | FIXED |
| ChiR8 | Tautological normalization test |  | FIXED |
| ChiR9 | Deploy surfaces omit required key | YES | FIXED |
| ChiR10 | Legacy Ok false counted as cleared | YES | FIXED |
| ChiR11 | Clients skip placeholder check |  | FIXED |
| ChiR12 | p2p reference omits admin auth |  | NOT FIXED |
| ChiR13 | Warning fires from non-admin paths |  | NOT FIXED |
| ChiR14 | clearbanned false success with peer enabled | YES | FIXED |
| ChiR15 | setban error class and cause |  | FIXED |
| ChiR16 | Last leg reason overwrites first |  | NOT FIXED |
| ChiR17 | k8s Secret doc omits the key |  | NOT FIXED |

## Non-blocking issues

### ChiR18 - The leg-success signal the new logic trusts can be false: `BanList.Clear()` swallows its own failure

*Same as [@freemans13's point "the legacy response's `Ok` field is never checked" (17-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1575#discussion_r3797033152).*

**Problem:** `handleClearBanned` now treats `resp.Ok`/`err == nil` from a leg as proof that a ban list was cleared, but neither server can report a failed clear. Extending @freemans13's point: they read the unconditional `Ok: true` at `services/legacy/Server.go:467-470` as the reason the missing check was a latent trap rather than a live bug - nothing could fail, so nothing could be reported wrong. The failure does exist; it is swallowed one layer below. `internal/banlist/ban_list.go:299-311` empties the in-memory map first, then issues `DELETE FROM bans`, and on error only logs - the method returns nothing, because `Clear()` in `internal/banlist/interface.go:34-35` has no error in its signature. Both callers therefore hard-code success: `services/p2p/Server.go:2392-2395` and `services/legacy/Server.go:467-470` return `Ok: true` unconditionally. So if the DELETE fails (store down, permissions, contention), the leg reports a clear, `handleClearBanned` returns `true`, the clearing service's in-memory map is empty so its entries drop out of `listbanned` - and the rows are still in the `bans` table, so those bans come back the next time that service loads the list. Nothing reloads it in the meantime: the ban lists handed to p2p and legacy come from `p2p.GetBanList` (`services/p2p/BanList.go:24-57`), which loads once in `Init` and never starts a periodic reload - `daemon/daemon_services.go:1214-1228` starts one only for the asset and propagation lists.

**Why it matters:** Item 5 of this PR exists so `clearbanned` stops reporting success when nothing was cleared, and ChiR10/ChiR14 closed the two places where the *handler* could get that wrong. This is the same defect one layer down, and it is the layer the handler's whole contract now rests on. The failure is also the worst-shaped one: the operator is told the bans are gone, sees them gone from `listbanned`, and they reappear on the next restart.

**Fix:** Give `Clear()` an error and propagate it, so the `Ok` field the handler now checks can actually be false. Note the in-memory map is wiped before the DELETE, so a failure already leaves memory and storage disagreeing - worth clearing memory only after the delete succeeds.

```go
// internal/banlist: Clear can fail, because it deletes the persisted rows.
func (b *BanList) Clear() error {
    ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
    defer cancel()

    if _, err := b.db.ExecContext(ctx, "DELETE FROM bans"); err != nil {
        return errors.NewStorageError("failed to clear bans table", err)
    }

    b.mu.Lock()
    b.bannedPeers = make(map[string]BanInfo)
    b.mu.Unlock()

    return nil
}
```

```go
// services/p2p/Server.go
func (s *Server) ClearBanned(ctx context.Context, _ *emptypb.Empty) (*p2p_api.ClearBannedResponse, error) {
    if err := s.banList.Clear(); err != nil {
        return nil, errors.WrapGRPC(err)
    }

    return &p2p_api.ClearBannedResponse{Ok: true}, nil
}
```

Criticality: 40%

### ChiR19 - `handleSetBan` counts the legacy leg's `Ok: true` as an applied ban, but it only means the request was queued

**Problem:** The comment this PR adds at `handlers.go:2442-2453` states that `success` "tracks whether at least one applicable ban leg actually applied the ban". For the legacy leg it does not. `legacy.Server.BanPeer` (`services/legacy/Server.go:488-495`) calls `banPeer`, which looks for a *currently connected* peer at that address, pushes a `banPeerForDurationMsg` onto a channel and returns nil (`:534-556`); `BanPeer` then answers `Ok: true`. The work happens later on the `peerHandler` goroutine (`services/legacy/peer_server.go:3226-3227` → `handleBanPeerForDurationMsg`, `:2580-2604`), where a failed `banList.Add` is only logged (`:2600-2603`). `UnbanPeer` is the same and says so in its own doc comment (`Server.go:515-521`, "ok means the request was received and processed, but no necessarily successfully", worker at `peer_server.go:2608-2622`). So `handlers.go:2509` sets `success = true` on receipt, and `setban` reports success even when the ban was never applied - and since the p2p and legacy services share one in-process ban list when they run in the same binary (`services/p2p/BanList.go:24-26`, "The singleton ensures that the P2P service and legacy service share the same in-process ban list instance"; `services/legacy/peer_server.go:3863` obtains it the same way), a store failure that refused the p2p leg can silently fail the queued legacy `Add` as well, with the acknowledgement already sent.

Separately, the same code never got the absent-vs-refused split, and the comment's justification - both methods are protected on both services, so a refused leg cannot hide behind an unprotected sibling - covers the `Unauthenticated` failure mode and only that one. `services/p2p/Server.go:2315-2331` returns an error whenever `banList.Add`/`Remove` fails, so any other leg-level refusal is still masked by a legacy `Ok`. In the **split-process** deployments - `deploy/docker/base/docker-teranode.yml` runs each service in its own container, and the per-service k8s CRs do the same - the two processes hold independent singletons, so there the P2P service's list genuinely never receives the entry while `setban` reports success. `listbanned` does not expose it either: it merges both legs (`handlers.go:2186-2245`, appending at `:2239`) and legacy's `ListBanned` reads its own list (`services/legacy/Server.go:450-452`), which does have the ban.

**Why it matters:** `setban` is the command that fails cleanly on the new unset-key default, so it is the one an upgrading operator exercises first, and this PR made `success` the sole determinant of what it reports. A rule that accepts a queue acknowledgement as proof of application can report success for a ban that landed nowhere - the false-positive direction on a security control, which is the direction ChiR14 was about.

**Fix:** Two small changes. Reuse `legAbsent` here so a reachable leg that refused fails the command, and stop treating the legacy acknowledgement as application - either make `legacy.Server.BanPeer`/`UnbanPeer` await the worker's outcome before answering, or narrow the comment at `:2442-2453` so the next reader is not misled about what `success` proves.

```go
var (
    success bool
    refused string
    lastErr string
)

// ... per leg:
if err := s.p2pClient.BanPeer(ctx, c.IPOrSubnet, expirationTimeInt64); err != nil {
    lastErr = err.Error()
    if !legAbsent(err) {
        refused = err.Error()
    }

    s.logger.Warnf("Error while trying to ban teranode peer: %v", err)
} else {
    success = true
}

// ... and at the end:
if refused != "" || !success {
```

Criticality: 35%

### ChiR20 - In split-process deployments a partial clear is reported as a flat failure, with the persisted rows already deleted

**Problem:** The new failure path collapses two different outcomes into one message (`handlers.go:2350-2369`): "no leg cleared anything" and "one leg cleared, another refused" both produce `Failed to clear banned list: <reason>`. The second case is the ChiR14 configuration, and what it leaves behind depends on the deployment shape. In a single-binary node the two services share one ban list instance (`services/p2p/BanList.go:24-26`; `services/legacy/peer_server.go:3863`), so the legacy leg's `Clear()` empties the very map the P2P service reads and only the message is wrong. In the split-process deployments - `deploy/docker/base/docker-teranode.yml` runs each service in its own container, and the per-service k8s CRs do the same - the two processes hold independent singletons and the `bans` table is their only shared state, so a legacy `Clear()` deletes *all* rows (`internal/banlist/ban_list.go:307`) while the P2P process keeps its in-memory list, which it never reloads (see ChiR18). The P2P bans therefore survive the failed command but disappear for good when that process next restarts, and the operator who read "Failed to clear banned list" has no reason to expect it.

**Why it matters:** The report is at least on the safe side, which is why this is not blocking - but the message drives what the operator does next, and "failed" invites a retry against state that is already half-changed, in the exact configuration ChiR14 was about and in the deployment shape ChiR9 established as the one that needs the shared key.

**Fix:** Distinguish the two outcomes in the message, and short-circuit on a refusal so the second leg's `DELETE` is not issued for a command that will report failure anyway. A refusal short-circuit does not change the shipped-default path, where the p2p leg is *absent* rather than refusing.

```go
if refused != "" || !success {
    msg := "Failed to clear banned list"
    if success {
        msg = "Ban list only partially cleared (one service cleared, another refused)"
    }

    reason := refused
    if reason == "" {
        reason = lastErr
    }

    if reason != "" {
        msg = fmt.Sprintf("%s: %s", msg, reason)
    }

    return nil, &bsvjson.RPCError{Code: bsvjson.ErrRPCInternal.Code, Message: msg}
}
```

Criticality: 28%

### ChiR21 - The two handlers' leg rules are now subtle and documented nowhere

**Problem:** `clearbanned` fails if any reachable leg refused but tolerates an absent one; `setban` succeeds if any leg applied it. Neither rule appears in the RPC reference: `docs/references/services/rpc_reference.md:1324-1353` describes `clearbanned` as taking no parameters and returning `null` on success (it actually returns `true`, `handlers.go:2371`), with no mention that it can now fail, that it fails when no ban service is reachable, or that a tolerated absent leg means part of the node's ban state was not touched. `setban` at `:1248` is the same. This is the documentation half of ChiR4, which was left out when the error-class half landed, and the semantics have changed twice since.

**Why it matters:** These two commands are the operator's only window onto the fail-closed default this PR introduces, and the error they now return is the one signal that says `grpc_admin_api_key` is unset. An operator who hits `Failed to clear banned list: ... Unauthenticated` has nothing in the reference to read next.

**Fix:** Add a short semantics note to both command sections and correct the return value.

```markdown
**Returns:** `true` on success.

Both the P2P and Legacy peer services are asked to clear their list. The command
fails (`ErrRPCInternal`) if a reachable service refuses - notably `Unauthenticated`
when `grpc_admin_api_key` is unset or a placeholder - or if no service cleared.
A configured-but-unreachable service (`Unavailable`/`DeadlineExceeded`) is
tolerated, so its portion of the ban state is left untouched. `setban` differs:
it succeeds if any applicable service applied the change.
```

Criticality: 15%

## Recap

| ChiR18 | Clear swallows delete failure |  | 40% |
| ChiR19 | setban counts a queued ban as applied |  | 35% |
| ChiR20 | Partial clear reported as total failure |  | 28% |
| ChiR21 | Leg semantics undocumented |  | 15% |

PR #1486, opened 2026-08-05
fix(model): tie the subtree-meta cache file to the subtree it is validating

freemans13:
- **`services/blockassembly/subtreeprocessor/SubtreeProcessor.go` — the reorg move-back path uses the same validated reader, and binds the subtree to its key.** This is the second production consumer of the file, and the one where the raw deserialize ran inside an `errgroup` goroutine. A later review round found the load-site binding above had closed only one of the two consumers: this path deserializes its own subtree from the committed key and never compared the two. Full subtrees in a block share a leaf count, so a foreign but well-formed `.subtree` under the right key passes both header checks alongside the committed subtree's genuine meta, and `moveBackBlockBulkBuild` then inserts the foreign node hashes into `currentTxMap` carrying the committed subtree's positional inpoints — `handleReorg` only falls back to a reset on error, so that poisons assembly state rather than failing. The comparison now happens in two places on purpose: `NewSubtreeMetaFromValidatedReader` makes it itself, since it already holds both values and every caller then inherits it; and this path checks explicitly as well, because `StoreTxInpointsForSubtreeMeta` is optional, the else branch reads no meta at all, and `subtreesNodes[idx]` is assigned straight after deserialize either way.

- **`cmd/filereader/file_reader.go` — the CLI reads through the validated reader, reports the defect, and still dumps the body.** This is the tool an operator points at a suspect meta file, and it was dying with an index out of range on exactly the corruption it was opened to diagnose. An earlier revision fixed the panic but returned on rejection, which lost the per-index dump — the thing that shows *which* entries are missing. It streams the happy path, prints the subtree's own root and node count first, names the mismatched root or count, re-opens the file to re-parse the body best-effort behind a `recover`, and then returns the error so the process exits non-zero rather than reporting success on the file it was opened to find. Recovering is right in a diagnostic tool aimed at a file already known to be malformed, and wrong in the node. Note the CLI compares against the `.subtree` file's own claimed root, because it has no block to supply a key, so a torn `.subtree` header is reported here as a meta mismatch — naming the other file of the two. A later round caught that this root was dereferenced without a nil check, and `RootHash()` is nil for a subtree claiming zero leaves; the print above it survives only because `fmt` recovers a `Stringer` panic. That is the same shape this PR set out to remove, so it is guarded and reported now.

## What happened

[...]

PR #1577, opened 2026-08-11
fix(blockvalidation): stop released blocks poisoning revalidation, and fix the subtree-meta peer fetch URL

freemans13:
## What happened

On 2026-08-11 the dev-scale-2 node permanently invalidated a perfectly valid block ("first subtree has no nodes") and froze for 10.7 hours until 20 manual `RevalidateBlock` gRPC calls unstuck it. Root-causing that incident surfaced two independent bugs that compounded into the freeze, and this PR fixes both — plus two further defects that a pre-merge review of the fix itself turned up.

The chain of events: the mining node was running with subtree-meta writing disabled, so the validating peer needed the subtree meta regenerator. The regenerator's local lookup missed (the subtree data hadn't landed yet), and its peer fallback failed — not transiently, but deterministically, because it builds the wrong URL (bug 1). That transient-looking processing error sent the block down the failure path, which returns the block's pooled node slices to a memory pool and then requeues **the same in-memory Block** for revalidation. The released subtrees still looked "loaded" to the revalidation path, so every retry failed instantly on gutted in-memory state — never re-reading the store — and escalated a transient error into a permanent on-chain `InvalidateBlock` (bug 2).

## The changes

**1. `model/Block.go` — the already-loaded early-exit now recognises released subtrees.** `GetAndValidateSubtrees` skipped reloading whenever every `SubtreeSlices` entry was non-nil. But releasing a block's nodes strips each subtree's `Nodes` while leaving the `*Subtree` pointer in place, so a requeued revalidation trusted a husk with zero nodes and `Block.Valid` failed "first subtree has no nodes" — a consensus-invalid verdict for what was actually a memory-management artifact. The early-exit now treats a non-nil subtree with zero nodes as not loaded and reloads from the store. In the incident, all four revalidation attempts completed in ~110µs via this early-exit with a stale transaction count — the regression test pins both symptoms (nodes reloaded, `TransactionCount` recomputed).

**2. `model/Block.go` + `services/blockvalidation/pools.go` — releasing a block's subtrees is now one locked, complete operation, and no longer pools an unmapped region.** The release logic moves into a new `Block.ReleaseSubtreeNodes` method which, under the block's own subtree mutex: hands heap-backed node slices to the pool, skips pooling mmap-backed slices, `Close`s every subtree (unmapping and deleting mmap backing files), and nils each `SubtreeSlices` entry so every loaded-check sees the truth. `releaseBlockNodes` delegates to it, and the cache-eviction callback's separate `Close` loop is deleted — the method now does it.

The mmap guard closes a real use-after-free, not a theoretical one. The old code justified pooling unconditionally on the grounds that "PutNodeSlice's cap-match check discards anything not produced by GetNodeSlice", and that reasoning is exactly backwards: `classIndexForCap` matches on capacity alone, go-subtree's `newFileBackedMmapNodes` returns `unsafe.Slice(...)[:0:capacity]` so cap equals the leaf capacity exactly, and leaf capacities are powers of two. A 1M-leaf subtree therefore produces cap `1<<20`, which is precisely `nodePoolClasses[5]`. The mmap region would have been accepted into the pool, munmapped by the eviction callback, and later handed out as a fresh Node slice. No evidence it ever fired, but nothing was stopping it.

**3. `model/subtree_meta_regenerator.go` — the peer fallback can now actually succeed.** It appended the *local* `Asset.APIPrefix` to peer DataHub URLs that already end in `/api/v1` (every `asset_httpAddress` / `asset_httpPublicAddress` form embeds the prefix), so every peer fetch requested `/api/v1/api/v1/subtree_data/<hash>` and 404ed — verified live against a real peer (correct URL 200, doubled URL 404). The regenerator now appends only `/subtree_data/<hash>`, exactly like every other consumer of an announced peer base URL (eight call sites across blockvalidation and subtreevalidation append only resource paths; the regenerator was the sole outlier), and the `apiPrefix` constructor parameter is gone — a local setting has no business shaping a remote peer's path. The convention this codifies: the announcing peer owns its full base URL, prefix included; consumers treat it as opaque and append only resource paths. The unit tests had encoded the opposite contract (bare test-server URL plus a separately-passed prefix), which is why this shipped green; they now use production-shaped URLs.

**4. Same file — the peer fetch retries 503s, is bounded, and failures are visible.** The fetch used a bare `http.Client` with no retry, while its siblings deliberately use `util.DoHTTPRequestBodyReaderWithRetry` because a peer's asset service 503s under admission control while generating subtree_data on demand. Switched to the shared helper (also gaining its SSRF-guarded client), with a 30-second per-peer context deadline. Without it, this inline-in-validation fetch would inherit the shared client's 5-minute streaming timeout against a hung peer. Note the 30s is a **whole-peer budget spanning all attempts**, not a per-attempt one — the old bare client gave 30s to a single attempt with no retries, so under sustained 503 backoff the last attempt now gets considerably less. Per-peer failure logs move from `Debugf` — invisible at INFO, which is why the incident log only ever showed the generic "not available locally or from peers" — up to `Warnf`, and the last peer error is wrapped into the returned error (only when one exists; the no-peers gRPC path previously would have rendered a `%!(EXTRA <nil>)` artifact).

**5. `model/Block.go` — `GetAndValidateSubtrees` no longer deadlocks on its own error path.** The function holds `b.subtreeSlicesMu` for its entire body, and both error returns in the subtree-size check formatted the message with `b.String()`, which takes that same mutex. `sync.RWMutex` is not reentrant, so the goroutine hung forever *still holding the block's mutex*. A peer reaches it with a block whose non-final subtrees have unequal lengths (e.g. `[4,2,4]`). Both returns now use `b.Hash().String()`, which reads an atomic and takes no lock.

This deadlock predates the branch. It is fixed here because it sits inside the function the branch changes, and because change 2 widens its blast radius: the eviction callback now takes `subtreeSlicesMu`, so a hung validation goroutine holding that mutex would stall every operation on the `lastValidatedBlocks` cache, not just the one block.

**6. `services/blockvalidation/BlockValidation.go` — `setTxMinedStatus` releases under the block mutex.** It closed its subtrees and cleared `SubtreeSlices` with no lock held. That was survivable while a released subtree stayed non-nil-but-gutted — a racing reader got a stale, wrong, but non-fatal value. Change 1 makes the entries nil, so the same racing reader would now dereference nil. The block is reachable via `lastValidatedBlocks` between the `Get` and the `Delete` above, and the TTL cleaner runs the same release concurrently, so the release is routed through `ReleaseSubtreeNodes` to take the mutex. Nodes are deliberately not pooled here (`put` is nil): these subtrees are read straight from the store, not allocated from the block validation node pool.

## Testing

- New regression test `TestGetAndValidateSubtrees_ReloadsAfterNodesReleased` reproduces the incident mechanism at the model layer: load, release nodes, revalidate, assert reload from store + recomputed `TransactionCount`.
- New `TestGetAndValidateSubtrees_SizeMismatchDoesNotDeadlock` drives a 2/1/2-length block through the size check on a 30s timeout. Verified honest: with change 5 reverted it fails on the timeout with the intended message; with the fix it passes in 1.8s.
- New `TestReleaseSubtreeNodes_ClosesMmapAndNilsEntries` pins the full release contract: heap slices pooled, mmap slices not pooled, mmap backing file removed by `Close`, entries nil-ed.
- New `TestReleaseBlockNodes_NilsSubtreeSliceEntries` covers the blockvalidation delegation.
- New `TestSubtreeMetaRegenerator_PeerBaseURLAlreadyHasAPIPrefix` pins the production URL contract, `TestSubtreeMetaRegenerator_RetriesOn503` pins the admission-control retry, and `TestSubtreeMetaRegenerator_NoPeers_CleanError` pins the no-peers error shape.
- All written test-first and watched fail for the intended reasons before each fix.
- Full `model` and `blockvalidation` suites green under `-race` (blockvalidation ~308s, catchup ~70s), `go build ./...`, `go vet`, `golangci-lint` clean.
- An adversarial multi-agent review ran over the diff (five independent lenses — correctness, concurrency, consensus flow, blast radius, test honesty — with every finding independently re-verified by refutation-first agents). It confirmed four distinct issues in the first version of this change, all fixed and test-covered here: the nil-error wrap artifact, the 5-minute stall ceiling, the dead-coded mmap `Close` loop, and the unlocked `SubtreeSlices` mutation. Two further hypotheses (double-release unsafety; a claimed pre-existing-race widening) were investigated and refuted with code-level proof. A second review pass over the finished diff produced changes 5 and 6.

## Trade-offs and limits

- The 30s per-peer fetch bound also caps total 503 backoff (~7.75s of delays plus request time fits, but a peer that needs longer than 30s to generate a large subtree_data on demand will time out where `check_block_subtrees`' catchup fetcher, with its 5-minute streaming window, would not). Deliberate: this fetch runs inline in block validation. It is a `const` rather than a settings key, unlike the sibling fetchers which rely on the configurable `http_streaming_timeout` — worth revisiting if a real peer ever trips it.
- The eviction callback now takes the block's subtree mutex while `expiringmap` holds its own write lock. I traced every holder of `subtreeSlicesMu` and none reaches back into `lastValidatedBlocks`, so there is no lock-order inversion. But `GetAndValidateSubtrees` holds that mutex across store I/O with retries, so evicting a block mid-reload will stall other cache operations where the callback previously never blocked. Change 5 removes the unbounded version of this hazard; the bounded one remains, tracked in #1581.
- Not every unlocked reader of `SubtreeSlices` was converted. Change 6 covers the release site that races the TTL cleaner; `quick_validate.go` and the `Valid`-path readers still touch entries without the mutex, and `setTxMinedStatus` still reallocates `SubtreeSlices` unlocked at the invalid-block branch. Those are pre-existing and unchanged in kind by this PR, but the nil entries make a future race louder — a sweep here would have meant an unreviewed widening of a consensus-path change, so it is tracked in #1580.
- `releaseBlockNodes` does not log its `Close` error. An earlier revision threaded a logger through it, which changed nine call sites — eight of them pre-existing, untested block-validation error paths. That re-dated already-untested lines as "new" and dropped the SonarQube new-code coverage gate to 79.7%, without improving the paths themselves. Since the only way `Close` fails is a munmap error (go-subtree swallows the backing-file `Remove` error), and those callers are failure paths already returning an error with no logger in scope, the wrapper now discards it explicitly with a godoc note. `ReleaseSubtreeNodes` still returns the error, and the two release sites that do have a logger — the reload path and `setTxMinedStatus` — log it.
- Out of scope, each now tracked: no automatic recovery from a transiently-invalidated block, which is what forced the incident's manual gRPC calls (#1579); the `"Subtree already exists - skipping"` handler leaving stored subtrees meta-less (#1582); and nothing validating that an announced DataHub URL actually serves the asset API (#1583).

https://claude.ai/code/session_01UMSWCexUagHvxpmU8xQ5nx


ordishs (changes requested):
The fixes here are correct and the tests are honest — reverting each one makes its test fail, and the mmap pool guard closes a real use-after-free rather than a theoretical one. One blocking item, which is a side effect of the URL repair itself.

**model/subtree_meta_regenerator.go:166-186** (P1) — reject an incomplete `Data` before you build the meta.

`NewSubtreeDataFromReader` stops on a clean `io.EOF` and reports success, so a truncated or zero-byte peer body leaves the tail `Txs` entries nil. `buildMetaFromSubtreeData` skips those entries with `continue`, and the returned `Meta` keeps zero-valued `TxInpoints`. `Meta.GetParentTxHashes` then returns nil with no error, `model/Block.go:1431` raises `BlockInvalidError`, and `BlockValidation.go:1948-1951` calls `storeInvalidBlock`. A valid block becomes permanently invalid — the same outcome this PR set out to prevent.

The doubled `/api/v1/api/v1` URL made this branch 404 on every real peer, so the path was dead. The URL fix makes it live. A zero-byte 200 is already documented at `services/blockvalidation/get_blocks.go:641-646` as a proxy cache replaying a failed on-demand generation.

**Action:** after `NewSubtreeDataFromReader`, require a non-nil transaction at every non-coinbase index, and return `errors.NewProcessingError` otherwise — mirroring the count check the sibling fetcher already has at `get_blocks.go:665-679`. Add a test that serves a body truncated at a transaction boundary, plus a zero-byte case, and assert a transient error rather than a `Meta`.

**model/subtree_meta_regenerator.go:149-185** (P2) — the same check also closes the hostile-peer truncation route. Each transaction the peer sends must still match `subtree.Nodes[txIndex]`, so a peer can truncate but never substitute. `check_block_subtrees.go:902` rejects exactly this with a `txCount != subtree.Length()` test; the regenerator has no equivalent.

Three P3 nits, none blocking: bound the peer body with an `io.LimitReader`; add a `SetNodeReleaser` so the reload path returns pooled slices instead of dropping them; and add a stalled-peer test — the current timeout test only reads the field back from the constructor, so deleting the timeout guard leaves the suite green.

freemans13:
P1 and P2 fixed in 778411dec, P3 test gap fixed with it. I verified the whole chain before implementing rather than taking it on trust, and every link holds:

- `serializeFromReader` (go-subtree v1.4.2, `subtree_data.go`) breaks on `io.EOF` and returns nil, leaving the tail `Txs` nil.
- `buildMetaFromSubtreeData` skipped those with `continue`, so `TxInpoints` stayed zero-valued.
- `Meta.GetParentTxHashes` returns `s.TxInpoints[index].GetParentTxHashes(), nil` — nil slice, **no error**.
- `validOrderAndBlessed` raises `BlockInvalidError` on `parentTxHashes == nil`, and `ValidateBlock` calls `storeInvalidBlock` when not revalidating.

And your framing of why it belongs in this PR is the part that convinced me: the doubled `/api/v1/api/v1` made the branch dead, so repairing the URL is what makes it reachable. The PR creates the exposure, so it carries the guard.

**Fix:** `buildMetaFromSubtreeData` now requires a non-nil `ParentTxHashes` at every node, starting at index 1 only when node 0 holds the coinbase placeholder — since only a block's first subtree carries it there, node 0 is a real transaction everywhere else.

**On P2** — agreed it is closed by the same check, and worth recording why: `serializeFromReader` hash-checks every transaction against `s.Subtree.Nodes[txIndex]` and returns `ErrTxHashMismatch` otherwise, so a hostile peer could only ever truncate, never substitute. Truncation is now rejected, so the route is shut.

**Tests:** `TestSubtreeMetaRegenerator_IncompletePeerBody_IsTransient` covers a zero-byte 200 and a body truncated at a transaction boundary, asserting a transient error, no `Meta`, no store write, and specifically **not** `ErrBlockInvalid`. Verified honest — with the guard disabled both subtests fail with "An error is expected but got nil", producing the bad meta exactly as you described.

**P3, the weak timeout test — you were right and it was worse than "leaves the suite green".** I wrote that test; it only read the field back from the constructor and proved nothing about behaviour. Added `TestSubtreeMetaRegenerator_StalledPeer_IsBounded`, which holds a response open and requires the fetch to be bounded. Confirmed by deleting the `context.WithTimeout`: the old test stays green, the new one fails at its 20s bound.

**One thing you should know before merging.** PR #1486 already contains this exact fix, in the same function, with the same `firstChecked` shape — I found it while checking for duplicate work. I have deliberately written this to match it so the conflict resolves by taking either side, but the two PRs **will** conflict in `buildMetaFromSubtreeData`, and whichever merges second needs to drop its copy. I implemented it here anyway rather than deferring to #1486, because this PR is the one that makes the branch reachable and could merge first, which would leave a window with the path live and unguarded. Flagging so it is a decision rather than a surprise.

**Two P3s not done, with reasoning:**

- **`io.LimitReader` on the peer body** — agreed in principle. An over-long body already fails (`serializeFromReader` returns `ErrTxIndexOutOfBounds` once `txIndex` exceeds the node count), but that does not stop a single enormous transaction or a huge bogus varint from being read into memory first, so the bound is still worth having. I did not add it because picking the constant is a policy call I did not want to invent inline, and I could not find an existing precedent to copy. Happy to add it if you name the bound, or I can derive one from the subtree node count and the max transaction size.
- **`SetNodeReleaser`** — correct that the reload path drops pooled slices instead of returning them. It is a lost optimisation rather than a leak, since the heap slices are simply collected, and it needs new API on `model.Block`. I would rather do it deliberately than bolt it onto this PR. Say the word and I will raise it as a follow-up.

Verification after the change: `go build ./...`, `go vet`, `make lint` clean at 0 issues, and `-race` green across `model` (15.8s), `blockvalidation` (302s), `catchup` (72s) and `subtreevalidation` (171s).

freemans13:
Follow-ups raised for the two P3s I did not implement, both unassigned:

- **#1590** — bound the peer `subtree_data` body. Records why the mechanism is trivial but the constant is not, lists the three candidate approaches, and points at #1586 (unvalidated length prefix in the go-subtree parse) as the same hazard one layer down whose resolution may subsume the bound chosen here.
- **#1591** — `model.Block` has `SetNodeAllocator` but no releaser, which is why the reload path passes `nil` and drops pooled slices. Flagged as a lost optimisation rather than a leak, with the suggestion that allocator and releaser be installed as a pair, since installing one without the other is what produced the asymmetry.

ordishs (changes requested):
Both incident fixes are correct and worth landing — the peer URL doubling and the gutted-husk early-exit are real bugs, properly fixed, and the `b.String()` self-deadlock fix is genuine (`model/Block.go:496` takes the same non-reentrant mutex). One blocker below.

## model/Block.go:1770 — P1 — nil-ed `SubtreeSlices` entries turn a concurrent release of an in-use block into a nil dereference

`releaseSubtreeNodesLocked` now writes `nil` into every `SubtreeSlices` entry. The mutex serialises **writers** only — no reader of `SubtreeSlices` takes `subtreeSlicesMu`. `Valid`'s long phases index it unlocked at `Block.go:936` (`subtree.Fees`), `:984` (`b.SubtreeSlices[0].Size()`), `:1081` (`len(subtree.Nodes)`), and `:1210`/`:1225` (`subtree.RootHash().String()`).

**Reachable path**

- Optimistic mining is on by default (`settings/settings.go:350`). The background validation goroutine spawned at `BlockValidation.go:1788` is never waited on — `optimisticMiningWg` has no `Wait` call.
- `BlockValidation.go:2028` sits at the same indent as `if useOptimisticMining` (line 1743), so it runs on **both** paths. The outer goroutine caches the block at line 2030 while `block.Valid` (line 1807) is still running.
- If that validation outlives the entry's 2-minute TTL, `expiringmap`'s cleaner fires the eviction function → `releaseBlockNodes` → `ReleaseSubtreeNodes`, nil-ing every entry.
- The still-running goroutine reads `b.SubtreeSlices[sIdx]` at `Block.go:1210` and panics at `Block.go:1225`: `Subtree.RootHash()` returns a nil `*chainhash.Hash` on a nil receiver (go-subtree@v1.4.2 `subtree.go:457-459`), and `chainhash.Hash.String()` has a **value** receiver (go-bt/v2@v2.6.7 `chainhash/hash.go:45`), so calling it dereferences nil. There is no `recover` in the `errgroup` goroutine, so the panic kills the process.

Before this change the same interleaving left a non-nil husk and produced a zero-length loop. The race itself pre-dates the diff; the diff changes its outcome from a silent wrong verdict to a crash. The same eviction also returns the pooled `[]Node` arrays to `PutNodeSlice` while those goroutines still read them.

The PR description acknowledges this direction ("the nil entries make a future race louder") and defers the reader sweep. The point of this finding is that the widening is reachable today, and taking the mutex on the write side does not mitigate it, because the readers never take it.

**Suggested fix**

- Add an in-use marker on `Block` — a refcount or an atomic flag held for the duration of `Valid` — and return `false` from the eviction function while it is set, so the entry survives to the next tick.
- Separately, drop the `Set` at `BlockValidation.go:2028-2030` on the optimistic path. Line 1864 already caches the block once `Valid` has returned, and that is the only `Set` with a safe lifetime.
- Either way, adding the nil-entry guard that `CheckMerkleRoot` already has at `Block.go:1837` to the four unlocked readers would contain the blast radius cheaply.

---

Also worth folding into the same pass — two new tests survive a trivial mutant, and both pin behaviour the description claims as proven:

- `model/block_release_reload_test.go:234` asserts the pooled slice **count**, not identity. Both fixture subtrees have non-nil `Nodes` of length 2, so inverting the guard at `model/Block.go:1749` to `st.IsMmapBacked()` keeps `len(pooled) == 1` and the test still passes — shipping exactly the mmap use-after-free the guard exists to prevent. Capture `heapSt.Nodes` and `mmapSt.Nodes` before the release and assert on `&pooled[0][0]` instead.
- `model/subtree_meta_regenerator.go:195`: `createTestSubtree` unconditionally writes `CoinbasePlaceholderHashValue` into `nodes[0]`, so `firstChecked` is 1 in every test and a bare `firstChecked := 1` stays green. The uncovered case is the one the comment calls out — a non-first subtree with a real transaction at node 0, served a zero-byte 200.

Seven further P2/P3 items (the new `lastValidatedBlocks.mu` → `subtreeSlicesMu` lock edge in the eviction callback; `ReleaseSubtreeNodes(nil)` at `:1312` discarding pool-drawn slices on the dominant success path, since `ValidateBlockWithOptions` installs the pooled allocator at `:1469`; the reload mutating a possibly caller-owned array; a missing settings loader-wiring test) are non-blocking and I can add them inline if useful.

freemans13:
Both P1s were right, and the first one is worse than the review states. Fixed in `8af3b5412`, `049ad1dbf` and `21c19cb31`.

## The nil-entry P1

I wrote the test before the fix, and it doesn't fail — it **takes the test binary down**. The panic lands in an errgroup goroutine, so `require.NotPanics` never sees it:

```
panic: runtime error: invalid memory address or nil pointer dereference
model.(*Block).checkDuplicateTransactionsInSubtree(...)  Block.go:1081
model.(*Block).checkDuplicateTransactions.func1()        Block.go:1020
golang.org/x/sync/errgroup.(*Group).Go.func1()
```

Your `RootHash().String()` chain is exactly right, and it isn't the only site. Fixed in two halves.

**Contain it.** Guards on every unlocked reader: `Valid`'s first-subtree check, `checkBlockRewardAndFees`, `checkDuplicateTransactions` and its per-subtree worker, `validateSubtree`, and the three reads in the merkle-root composition. Two things beyond the nil check itself:

- Each guard **captures the entry into a local, tests the local, then uses the local**. Re-reading `b.SubtreeSlices[i]` after checking leaves the identical window open, so the capture is the fix and the nil test is just how it's spelled.
- Every guard returns a **processing** error, never `BlockInvalid`. A subtree released mid-validation is a memory-management artifact and says nothing about the block; returning a consensus verdict there would rebuild the exact incident this PR exists to fix. `CheckMerkleRoot`'s existing guard already had this shape — the rest of the path now matches it, and there's a test pinning yours so a refactor can't quietly drop it.

One placement detail: the nil check for the dup-check fan-out lives in `checkDuplicateTransactionsInSubtree`, not in the loop that spawns it. Returning from the loop would abandon the goroutines already in flight; going through the errgroup means they're still waited on.

**Remove the reachability.** Your second suggestion was the better one and I took it: `BlockValidation.go`'s cache write sat outside the `if useOptimisticMining` block and so ran on both paths, publishing the block into `lastValidatedBlocks` while the background goroutine may still be inside `block.Valid` on that same `*Block`. It's now gated on `!useOptimisticMining`. Nothing is lost — the goroutine caches the block itself once `Valid` returns, immediately before the `updateSubtreesDAH` notification `setMined` waits on, so the ordering the cache protects is unchanged, and that Set is the only one on the path whose lifetime outlives its reader. The path's own comment has said so since it was written: *"We do NOT cache the block here as subtrees are not yet loaded."* The else-branch warnings moved inside the guard too, so the optimistic path stops logging a spurious "not caching block" on every block.

I did not add the refcount. With the release no longer scheduled underneath a live validation, and the readers no longer able to crash on one, it buys correctness I can't currently reach a case for — and it's a change to the block lifecycle that deserves its own PR rather than a fourth one bolted onto this.

## The pooled-slice test mutant

Also correct, and I'd have missed why it mattered if you hadn't spelled out the inversion. I ran it: with the guard flipped to `st.IsMmapBacked()`, `require.Len(pooled, 1)` still passes.

Then it got interesting. My first attempt at an identity assertion used `&pooled[0][0]`, and under the inverted guard the test **segfaulted on an unmapped address**:

```
unexpected fault address 0x10610c000
fatal error: fault
[signal SIGSEGV: segmentation violation]
runtime.memmove()
```

That is the use-after-free itself, on demand — `Close` had munmapped the region and the assertion read through the pooled pointer into it. So the assertion now compares the backing-array addresses as plain integers via `reflect`, never dereferencing, and fails cleanly rather than faulting. Verified both directions: inverted guard fails on identity, correct guard passes.

## Not done

- **No test for the one-line optimistic-path guard.** A test asserting `!optimistic && hasValidSubtrees` would be tautological, and a real end-to-end assertion that the block isn't cached early is inherently racy — the honest choices were a worthless test or a flaky one, so I took neither and am flagging it instead. It's covered indirectly by the suite's setMined flows.
- **The reader sweep and the eviction-callback stall** stay where you left them, in #1580 and #1581. The guards here contain the nil hazard; they don't convert the unlocked readers into locked ones.

## Verification

`go build ./...`, `go vet`, `go test -race ./model/...` (15.3s), `go test -race ./services/blockvalidation/...` (303.6s + 70.1s catchup), `make lint` 0 issues — all re-run after merging the upstream sync that landed on this branch mid-work, not just before it.

Also folded in `feea51ff1`, a docs fix from an earlier pass that had never made it onto the branch: the settings `longdesc` still claimed `http_streaming_timeout` defaults to 5 minutes when shipped `settings.conf` sets 600000.

ordishs (approved):
Re-reviewed at `f7513775a`. The blocker is fixed properly, and fixed in both halves rather than one. Approving.

## The P1 — nil-ed `SubtreeSlices` entries

`049ad1dbf` skips the cache `Set` on the optimistic-mining path, removing the reachability. Verified nothing is lost: the background goroutine's own `Set` (`BlockValidation.go:1876`) precedes `updateSubtreesDAH` (line 1880), which sends the `BlockSubtreesSet` notification the setMined worker waits on — so setMined always finds the block cached, and that `Set` is the only one on the path whose lifetime outlives its reader.

`8af3b5412` then guards every unlocked reader, so the hazard is contained even if a release is ever scheduled underneath a live validation again. Three things there are better than what I asked for:

- **Capture-then-test-then-use** rather than a bare nil check. Re-reading `b.SubtreeSlices[i]` after the check would have left the same window open.
- **`ProcessingError`, never `BlockInvalidError`.** A subtree released mid-validation is a memory-management artifact and says nothing about the block. I confirmed both callers honour it — the non-optimistic path returns early on `ErrProcessing` (`BlockValidation.go:1932`), and the optimistic path falls to the `else` at line 1844 and calls `ReValidateBlock`. Returning a consensus verdict here would have recreated the exact incident this branch exists to fix.
- The `checkDuplicateTransactions` guard sits in the callee rather than the loop body, so the error travels through the errgroup and in-flight goroutines are still waited on.

Checked independently: every `SubtreeSlices` deref in `model/Block.go` is now guarded or under the mutex (680, 946, 1000, 1038, 1074, 1241, 1877, 1915, 1935, 1967); no new guard sits inside the `subtreeSlicesMu` region, so none reintroduces the `b.String()` re-entrancy deadlock; `model/update-tx-mined.go:226` already had its own guard.

**Test honesty**: I removed the `validateSubtree` guard and re-ran `TestValidateSubtree_NilSubtreeIsTransient`. It reproduces the exact defect — `nil pointer dereference` at `model/Block.go:1262`, the `RootHash().String()` chain from the original finding.

## Other first-pass findings

Fixed: pooled-slice **identity** assertion in `block_release_reload_test.go` (and better than suggested — pointers compared as integers, since `Close` has already munmapped by assert time); the `put == nil` rationale; the settings docs row.

Deferred with an accurate comment, no objection: `setTxMinedStatus` still passes `nil` rather than `PutNodeSlice`, so a cached block's pool-allocated slices are dropped rather than recycled. The comment now says so plainly instead of asserting the opposite.

Still open, none blocking:

- Eviction callback still takes the block mutex under `expiringmap`'s write lock (`BlockValidation.go:377`). `049ad1dbf` narrows reachability but the lock edge stands — tracked in #1581.
- The reload path still mutates a possibly caller-owned `SubtreeSlices` array in place (blockassembly aliases `job.Subtrees`). Latent.
- Three untested claims: the `firstChecked = 0` branch (`80e1ed335` hoisted the expression but added no test, so `firstChecked := 1` is still a surviving mutant); no loader-wiring or constant-drift test for `SubtreeMetaPeerFetchTimeout`; no concurrent-release test for the release-under-mutex claim.
- `allowLoopbackHTTP` restores a hardcoded `true` rather than the previous value.

## Credit where due

`f79ad4292` adds a release before the unset-mined reload (`BlockValidation.go:1243`), closing an mmap mapping-and-backing-file leak I missed. The hoist in `80e1ed335` incidentally guards a `subtree.Nodes[0]` read that its `Length() > 0` test did not cover. And I checked `b7cd028ac`'s deletion of `TestSubtreeMetaRegenerator_PeerBaseURLAlreadyHasAPIPrefix`: genuinely duplicated — `_FromPeer` has the identical exact-path handler with `server.URL + "/api/v1"` and 404s everything else, so the URL contract stays pinned.

## Verification

```
go vet ./model/... ./services/blockvalidation/... ./settings/...   exit 0
go test ./model/... -count=1 -race                                 ok  14.179s
go test ./services/blockvalidation/ -count=1 -race                 ok 297.090s
```

liam (approved):

icellan:
Re-reviewed at f7513775a. The nil-deref guards are in and correct (Block.go:677, :1095, :1259), and the `put=nil` comment at BlockValidation.go:1311 no longer claims the slices never came from the pool. The rest of the previous round is still open, and two of them are on the exact path this PR exists to fix.

**P0 — `model/subtree_meta_regenerator.go:76`, `:89`: an incomplete body aborts regeneration instead of falling through to the next source.**
Both the local branch and the peer loop do `return r.buildAndStoreMeta(...)`. A truncated local `subtree_data` file passes `getLocalSubtreeData` (`serializeFromReader` stops at a clean `io.EOF`), the new completeness check at :202 rejects it, and the error returns — the peer loop at :84-95 is never entered even though the announcing peer holds a complete copy. Inside the loop it is the same shape: the first peer that yields a body wins, so one poisoned peer means peers 2..N are never tried. Every retry re-reads the same truncated file, so the block never validates. The test only configures a single peer, so it can't catch this.

**P1 — `model/subtree_meta_regenerator.go:138`: no cache-bust on the peer URL.**
`get_blocks.go` routes its `subtree_data` fetch through `peerResourceURL` with a cachebust parameter precisely because a peer's proxy can cache an aborted on-demand generation as `200` + 0 bytes (issue-1368, cited in this PR's own test comment). This builds a bare `%s/subtree_data/%s`. Combined with the short-circuit above, one cached empty response wedges regeneration for that subtree until the upstream entry expires.

**P1 — `model/subtree_meta_regenerator.go:128`: `DefaultPeerFetchTimeout` is 30s for a payload fetched elsewhere under 10m.**
The budget covers all 503 backoff (~7.75s with the default retry config) plus the full body stream. `subtreevalidation/check_block_subtrees.go` fetches the same payload under `blockvalidation_subtree_data_fetch_timeout` = 10m. For a mainnet-size subtree the body can't be streamed in the remaining ~22s, so `RegenerateMeta` always returns `ErrProcessing`, the block is re-queued as transient, and the URL fix is neutralised by the bound. Making it configurable helps only if `settings.conf` also raises it.

**P1 — `model/Block.go:1637`: the reload release is destructive on subtrees the block doesn't exclusively own.**
`quick_validate.go:159`/`:163` store the same `fullSubtree` pointer into both `block.SubtreeSlices[subtreeIdx]` and the `SubtreeWriteJob` queued on the shared `writeJobsChan`. If a later batch fails, earlier entries stay populated and later ones are nil; the block is then re-validated normally, `GetAndValidateSubtrees` takes the reload branch, and `releaseSubtreeNodesLocked(nil)` strips `Nodes` and `Close()`s subtrees a write worker still holds — `job.Subtree.Serialize()` then derefs a nil `RootHash()`, or writes a 0-leaf `.subtree` with `AllowOverwrite`. It's a data race on `st.Nodes` either way. `blockassembly/Server.go:1663` aliases the processor's live `job.Subtrees` into a `model.Block` and calls `Valid`; that one is safe only by the "every job subtree has ≥2 nodes" invariant, which is worth a comment naming the cross-package dependency at minimum.

**P1 — `services/blockvalidation/BlockValidation.go:373`: the evictor now takes a per-block mutex under the cache's map-wide lock.**
`expiringmap.clean()` holds `m.mu.Lock()` while calling `evictionFn`, and `releaseBlockNodes` → `ReleaseSubtreeNodes` blocks on `block.subtreeSlicesMu` — held as a write lock for the whole body of `GetAndValidateSubtrees` (:1606), which spans concurrent store reads with retries and backoff. Evicting a block that is mid-reload parks the cleaner under `m.mu`, and every `lastValidatedBlocks` Get/Set/Delete blocks behind it. The old eviction fn was lock-free.

**P1 — `services/blockvalidation/BlockValidation.go:1247`, `:1266`: the unsetMined branch still writes `SubtreeSlices` with no lock.**
The leak half is fixed (the release at :1243 before the overwrite). The lock half isn't: the slice-header write and the element assignments run outside `subtreeSlicesMu` while the TTL cleaner runs the release under it. Entries the unset path is filling get nil-ed, `UpdateTxMinedStatus` reports `missing subtree %d of %d`, and the invalid block's transactions are never un-mined — stale mined flags in the UTXO store. `-race` should flag the header write.

Two nits, non-blocking: the completeness predicate at :196-205 duplicates `get_blocks.go:665-685` including the index-0 coinbase exemption, and loses the `newPoisonedSubtreeDataError` that names the peer and byte count — one helper called from both sites would be better. And the `Debugf`→`Warnf` promotions at :79 and :93 mean several WARN lines per subtree on the routine missing-meta path, on top of the entry warn at :71.

PR #1578, opened 2026-08-12
Remove committed p2p identity keys from settings.conf and guard against recurrence

gokutheengineer:
**Refs bitcoin-sv/teranode#4739** - deliberately not `Closes`: removal from HEAD does not revoke the key (it stays reachable in history, clones, forks, and cached views), so the issue must stay open until the deployment check and any rotation recorded there are done. The issue carries the rotation action.

### Problem
`settings.conf` committed full Ed25519 identity keys in a public repo: `p2p_private_key.operator.mainnet.stage.1` (a plausible real operator key, present nowhere else in the repo) and a context-less `alert_p2p_private_key` test fixture that silently became the alert identity for every context that did not override it. Nothing guarded against the next key landing in the file: the gitleaks config allowlisted these settings by key name, and its default ruleset does not flag `*_private_key = <hex>` lines at all (verified empirically against the actual leaked key, at the CI-pinned gitleaks 8.18.4 built from source).

### Fix
- Removed `p2p_private_key.operator.mainnet.stage.1`. Only that internal context resolved it; the committed production contexts (`operator`, `docker.m`) never resolved any p2p key and keep using the read-`p2p.key`-or-generate path unchanged.
- Removed the context-less `alert_p2p_private_key` default so the alert service's documented auto-generate path runs. No committed stack loses a needed value: `docker.ss` does not run alert, compose test contexts set `startAlert.docker = false`, and production contexts disable alert.
- Removed the commented-out `tx_blaster_p2p_private_key` (192-hex, the legacy 96-byte libp2p form) and its peer ID from settings.conf, and dropped its gitleaks allowlist entry. A commented-out key is just as committed as a live one, and this value's provenance could not be established (it appeared nowhere else in the repo - the same evidential profile as the rotated operator key). The setting was inert, so removal changes no behavior; regenerate with `cmd/keygen` when needed.
- Turned alert off in the gennodes multinode rig (template `startAlert.docker.teranodeN = false`, dropped `-alert=1` from the core sidecar). Alert was non-functional there anyway - its port is unexposed, it has no local bootstrap peers, and the regtest topic has no matching genesis keys - while its libp2p host bootstraps into the public IPFS DHT. This matches what `compose/settings_test.conf` already does for the other docker rigs.
- Added `settings/committed_private_keys_test.go` as the CI guard. It is value-shape based, not key-name based: every assignment in every committed settings-format file whose value looks like key material must be a known public test fixture, and `*private_key`-named settings may never carry a context-less identity key. Review hardening (thanks to the round on 172755772):
  - Hex shape is `{64,}` minus the 66/130 public-key lengths, so the 96-byte legacy libp2p form (192 hex) that `crypto.UnmarshalEd25519PrivateKey` accepts is covered, matching the gitleaks rule.
  - Pipe-separated multi-key values (`miner_wallet_private_keys` via `getMultiString`) are split and each element checked.
  - Lines are parsed exactly as gocore's `processFile` does (comment from the first `#` anywhere, key is everything left of the first `=`), so the guard sees every literal gocore sees - no parser-divergence bypasses.
  - `${VAR}` references are resolved against same-file definitions the way gocore's `replaceVariables` does - substring substitution iterated to a fixed point (bounded), not a single whole-value hop - so nested hops and keys assembled from sub-floor fragments (`${A}${B}${C}${D}`) are checked as the value the consumer actually gets. Deliberate divergence: unresolvable tokens stay put rather than being rewritten to `{UNKNOWN}`; a guard must not rewrite what it cannot see into something benign.
  - Every key-shaped token on the RAW line is fixture-checked before comment stripping, so commented-out keys - just as committed, just as exposed - cannot hide behind a `#`.
  - Scan targets are discovered via `git ls-files` (committed files only - a filesystem walk would scan a developer's gitignored `settings_local.conf`, which legitimately holds real keys), with the five known files asserted present so a discovery bug cannot silently shrink coverage.
  - Scoping note: the context-less structural ban applies to hex identity keys. The pre-existing context-less WIF wallet default (`coinbase_wallet_private_key = ${PK1}`) is part of the PK1-PK6 wallet-key finding tracked separately on the issue; the fixture check still fails on any real WIF anywhere.
- Tightened `.gitleaks.toml`: added an explicit `*private_key = <hex>` detection rule and switched the allowlist from key names to the specific public fixture values. Verified at gitleaks 8.18.4: the rule catches the removed line's exact shape, the full-tree scan is clean, and every allowlist entry is load-bearing.
- Kept the context-scoped `docker.ss`/`dev` fixtures that are load-bearing (their peer IDs are pinned in config) and already public test data.

### Residual
- The rotation itself (tracked on the issue, which stays open): confirm whether `SETTINGS_CONTEXT=operator.mainnet.stage.1` was ever deployed; if yes, rotate and re-pin the peer ID wherever referenced. Two traps recorded there: the asset response signer reads `P2P.PrivateKey` directly and silently stops signing when it is empty (no error, no request-time log), and a fresh `p2p.key` changes the peer ID.
- `settings.conf` PK1-PK6 are committed mainnet-format WIF wallet keys, and `miner_wallet_private_keys.operator.mainnet = ${PK1} | ${PK2} | ${PK3}` resolves to them unless a deployment overrides the vars (nothing in `deploy/` does). Fund exposure, out of this issue's scope - flagged on the tracker issue for separate follow-up. The guard allowlists them explicitly with a comment so their later removal trips nothing.
- `p2p_shared_key` (settings.conf) is a committed context-less 32-byte PSK with no Go consumer found - allowlisted as pre-existing, candidate for separate removal.
- `compose/cmd/gennodes/peer_keys.json` (10 fixture keys) stays path-allowlisted in gitleaks; a real key pasted there would not be caught. Fixture-only file, low risk.
- Upstream `go-alert-system` regenerates the auto-generated alert key on every start (its `readPrivateKey` hex-decodes the file *path*) and writes it 0644 - upstream bug, not addressed here.
- Pre-existing doc inaccuracies (p2p key "stored in the blockchain database", alert key "PEM-format") left untouched to keep the diff minimal.

### Tests
- `TestNoRealPrivateKeysCommitted` guard, verified to fail on: the two original bad lines, a non-fixture scoped key, a context-suffixed variable definition (`PK9.operator.mainnet = <hex>`), an innocuously named WIF literal, a 192-hex legacy-form key, pipe-separated WIF literals, a hex value with a no-space `#comment` suffix, a dash-named definition (`my-key = <hex>`), a context-less fixture behind nested `${VAR}` hops, a commented-out 192-hex key, and a key assembled from four 32-hex fragments via `${A}${B}${C}${D}`.
- `SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./settings/... ./compose/cmd/gennodes/...` - pass
- `SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/alert/...` - pass
- `go vet ./settings/ ./compose/cmd/gennodes/` - pass
- gitleaks (CI config) over the PR commit range and the full tree: no leaks; synthetic `p2p_private_key.operator.prod = <hex>`: caught by the new rule.


liam (approved):
Verified the claims against the tree rather than the description. Approving — one request before merge, and it is about the description, not the code.

**Verified:** the removed `operator.mainnet.stage.1` key resolved only for that exact context, and no context-less `p2p_private_key` default remains, so every other context falls through to the read-`p2p.key`-or-generate path. The alert auto-generate path is real (`services/alert/server.go:424` → `createPrivateKeyDirectory` → `os.UserHomeDir` + `os.Mkdir`), and the `Dockerfile` sets no `USER`, so `$HOME` is writable in the standard image — the concern that motivated the earlier approach is genuinely low risk. The new guard test and the gennodes tests pass locally.

## The description no longer matches the diff

The body says:

> The gennodes multinode overlay now pins a per-node `alert_p2p_private_key` (the node's existing fixture key), so its alert-enabled core container never depends on a writable `$HOME`

The diff does something else. There is no pinned key. It sets `startAlert.docker.teranodeN = false` in the settings template and strips `-alert=1` from the core sidecar (`compose/cmd/gennodes/main.go:263`, plus `main_test.go`, both templates, and `MULTINODE.md`). The final commit is titled *"guard on value shape, drop vestigial gennodes alert"* — the approach changed and the body did not follow.

The new approach looks right: the template comment justifies it (unexposed port, no local bootstrap peers, no matching genesis keys on regtest), and it matches what `compose/settings_test.conf` already does for the other docker rigs. But it is a behaviour change to the multinode rig that a reviewer approving from the description would not know they were approving. Please update the body.

## Guard-test nits

- `committedSettingsFiles` is a hardcoded list. Existing entries fail loudly if renamed, so no silent drift there, but a *newly added* committed settings file is never scanned. A glob over tracked `*.conf` would be drift-proof, which is the point of the guard.
- `hexKeyValue` matches exactly 64 or 128 hex chars; the gitleaks rule uses `{64,}`. A key of some other length passes the test but trips gitleaks. Worth aligning.
- The new gitleaks rule is hex-only, so a WIF key committed under a `*_private_key` name is caught by the Go test but not by the diff scan. Fine, just asymmetric.

## On the residual you flagged

PK1-PK6 are still committed and now explicitly allowlisted in the guard. Reasonable scoping, but an allowlist entry carrying "tracked separately" in its comment tends to outlive the intention to remove it. Worth confirming that follow-up issue actually exists.

ctnguyen (changes requested):
Review at commit [1727557](https://github.com/gokutheengineer/teranode/commit/1727557724981912e5d1aaa50376062ffb9d2b10)

Good, well-scoped piece of work. Removing `p2p_private_key.operator.mainnet.stage.1` is clearly right, and the value `a4a361d2…f22e` now appears nowhere in the tree at this commit, so the removal is complete at HEAD. The reasoning for dropping the context-less `alert_p2p_private_key` holds up against the code: `settings/settings.go:192` already defaults it to `""`, `alert_settings.go` documents empty as "auto-generation of ephemeral identity", and `requireP2P` → `createPrivateKeyDirectory` (`services/alert/server.go:411-477`) is a real path. Making the guard value-shaped rather than key-name-based is the correct instinct — gocore's `replaceVariables` (`config.go:553-572`) really does resolve `${VAR}` through `findValue`, including context-suffixed definitions, so a name-based check would have been trivially escapable. The gennodes alert removal is sound too: the split entrypoint passes `-all=0` first (`docker-compose-split.yml.tmpl:141`), so dropping `-alert=1` genuinely disables it, and `startAlert.docker.teranodeN` resolves for the `docker.teranodeN.test` context because `findValue` strips segments right-to-left. The body's rotation note is the right instinct as well — my first item is about making sure it survives the merge rather than about it being missing.

## Blocking issues

### ChiR1 — Deletion does not revoke, and merging closes the only record of the rotation

**Problem:** The diff removes the key from HEAD. It does not remove it from the repository's history: the blob remains reachable from earlier commits in every clone, fork and mirror, and in the hosting provider's cached commit views. The only action that reduces exposure is rotating the identity, and this PR cannot perform it. The PR body says as much under "Rotation note (ops, outside this repo)", and [bitcoin-sv/teranode#4739](https://github.com/bitcoin-sv/teranode/issues/4739) carries the concrete instruction ("check internally whether `operator.mainnet.stage.1` runs (or ever ran) anywhere. If yes: rotate the node's identity"). But the body opens with **Closes bitcoin-sv/teranode#4739**, so merging auto-closes the one tracked item that records the outstanding rotation. Nothing in the repo carries it afterwards. I cannot determine from this tree whether that context was ever deployed — that is exactly the check the issue asks for and the reason the record needs to outlive the merge.

**Why it matters:** For a key-exposure incident, code removal is the cosmetic half and rotation is the mitigating half. Closing the tracker on the cosmetic half is how a live compromised identity ends up with no owner. Two follow-on details make the rotation itself easy to get wrong, and both are worth carrying into the ticket rather than only into a PR body that nobody re-reads:

- The asset signer reads `tSettings.P2P.PrivateKey` directly (`services/asset/httpimpl/http.go:286-301`) and the p2p service's generate-or-read path assigns the key to a **local** variable (`services/p2p/Server.go:292-341`) without writing it back to settings. So a node rotated by deleting the setting and letting `p2p.key` be generated silently loses response signing — `Sign` returns nil when `privKey` is nil (`http.go:667-681`), and the startup path only logs on decode/unmarshal failure, never on an empty value. No error, no log at request time, just missing `X-Signature` headers.
- Rotating by generating a fresh `p2p.key` changes the node's peer ID, which matters wherever that ID is pinned in peer configuration.

**Fix:** Before merge, either change `Closes` to `Refs` so the issue stays open until the deployment check and any rotation are done, or open a separate rotation ticket and reference it from both the PR body and the issue. Keeping the exposure statement explicit in that ticket is the point — something like:

```
The key is in git history; removal from HEAD does not revoke it.
Action: confirm whether SETTINGS_CONTEXT=operator.mainnet.stage.1 was ever
deployed. If yes, rotate the identity and re-pin the peer ID wherever it is
referenced. If asset_sign_http_responses is on, set p2p_private_key explicitly
— the asset signer does not fall back to the generated p2p.key and fails silent.
```

### ChiR2 — The hex shape misses the 96-byte libp2p key form that the consumers accept

*Same as [@liam's point "hexKeyValue matches exactly 64 or 128 hex chars" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1578#pullrequestreview-4960374755).*

**Problem:** `hexKeyValue` (`settings/committed_private_keys_test.go:69`) accepts exactly 64 or 128 hex characters. Liam framed the mismatch as "passes the test but trips gitleaks"; the sharper half is the other direction. Every consumer of these settings funnels into `crypto.UnmarshalEd25519PrivateKey` — `services/p2p/Server.go:385`, `services/asset/httpimpl/http.go:293`, `cmd/keygen/key_generator.go:107`, and upstream `go-alert-system`'s `readPrivateKey`. That function accepts **two** encodings: 64 bytes, and 96 bytes (`ed25519.PrivateKeySize + ed25519.PublicKeySize`, the legacy redundant-public-key form). A working, committed identity key in the 96-byte form is 192 hex characters and slips straight through the guard.

**Why it matters:** This is the identity-key shape the PR exists to guard against, and the guard is billed as the full-file complement to a scan that only sees PR diffs. A 192-hex key added today is flagged once by the new rule (`{64,}`) and then never again by the standing check — so if it is added in a commit whose diff scan is skipped, missed or overridden, nothing catches it afterwards.

**Fix:** Widen to `{64,}` to match the scan rule, and exclude the 33/65-byte public-key lengths explicitly rather than by omission, so the intent survives the change.

```go
hexKeyValue = regexp.MustCompile(`^[0-9a-fA-F]{64,}$`)

// isKeyShapedHex reports hex blobs long enough to be key material.
// 66/130 chars are compressed/uncompressed public keys (alert_genesis_keys)
// and are deliberately excluded.
func isKeyShapedHex(v string) bool {
    return hexKeyValue.MatchString(v) && len(v) != 66 && len(v) != 130
}
```

## Non-blocking issues

### ChiR3 — Pipe-separated multi-key values are invisible to the guard

**Problem:** `hexKeyValue` and `wifKeyValue` (`settings/committed_private_keys_test.go:69-70`) are anchored with `^…$` and applied to the whole right-hand side (`:107-110`). Teranode has settings that are *lists* of keys: `miner_wallet_private_keys` is read via `getMultiString("miner_wallet_private_keys", "|", …)` (`settings/settings.go:299`) and `settings.conf:848-850` already uses the `A | B | C` form. A line such as

```ini
miner_wallet_private_keys.operator.mainnet = L56TgyTpDdvL3W24SMoALYotibToSCySQeo4pThLKxw6EFR6f93Q | KyAwSjuXZNgj78w3W7mR1fVMbPFu2heaCJJkWK5Yy58NZ4xafV6k
```

matches neither regex, so the loop `continue`s and the test passes. The new scan rule does not cover it either — its value pattern is `[0-9a-f]{64,}`, so WIFs never match.

**Why it matters:** Not blocking, because this is wallet-key hardening rather than the identity-key recurrence this PR is closing, and the underlying `miner_wallet_private_keys` exposure is already tracked separately on [bitcoin-sv/teranode#4739](https://github.com/bitcoin-sv/teranode/issues/4739). But it is the one setting in `settings.conf` whose documented format the guard structurally cannot see, so replacing `${PK1} | ${PK2} | ${PK3}` with literals — the natural move when the env vars are inconvenient — would land silently.

**Fix:** Strip the comment once, then split on `|` and shape-check each part.

```go
raw := stripComment(m[3])
for _, part := range strings.Split(raw, "|") {
    value := strings.Trim(strings.TrimSpace(part), `"`)
    if value == "" || (!isKeyShapedHex(value) && !wifKeyValue.MatchString(value)) {
        continue
    }
    // …existing allowlist + context-less assertions…
}
```

Note this interacts with ChiR2: once values are split, `alert_genesis_keys` yields five 66-hex compressed public keys, so the length exclusion above has to stay.

### ChiR4 — The line parser is stricter than gocore's, which creates bypasses

**Problem:** Two divergences from `gocore.processFile` (`config.go:195-217`):

1. Comment stripping. gocore does `strings.Split(line, "#")[0]` — everything from the *first* `#` is a comment, whatever precedes it. `normalizeValue` (`settings/committed_private_keys_test.go:75-82`) only strips at `" #"`. So `p2p_private_key.operator.mainnet = a4a3…f22e#stage1` is a live key to gocore and a non-key-shaped string to the guard.
2. Key charset. `assignmentLine` (`:64`) requires the first segment to be `[A-Za-z0-9_]+`. gocore takes everything left of the first `=`. `my-key = <64 hex>` is a valid, resolvable definition (`${my-key}` resolves through `findValue`) that the regex rejects outright.

**Why it matters:** The value-shape design is justified in the PR body by its resistance to indirection — "the exact escape an adversarial review demonstrated against a name-based check". That guarantee only holds if the guard sees every literal gocore sees; each parser divergence reopens the escape it was built to close.

**Fix:** Drop the regex and parse exactly as gocore does, so parity is structural rather than maintained by hand.

```go
if i := strings.Index(line, "#"); i >= 0 {
    line = line[:i]
}
pos := strings.Index(line, "=")
if pos < 0 {
    continue
}
key := strings.TrimSpace(line[:pos])
name, context := key, ""
if i := strings.Index(key, "."); i >= 0 {
    name, context = key[:i], key[i:]
}
```

### ChiR5 — The context-less check is defeated by one level of indirection

**Problem:** The `require.NotEmpty(t, context, …)` assertion (`settings/committed_private_keys_test.go:116-119`) fires only when the name matches `private_keys?$` *and* the value on that same line is key-shaped. Both conditions fail under indirection:

```ini
ALERT_KEY             = e76c…2526     # allowlisted fixture, name is not *private_key → no context check
alert_p2p_private_key = ${ALERT_KEY}  # value is not key-shaped → line skipped entirely
```

That reinstates exactly the defect this PR removes — a context-less alert identity shared by every deployment that does not override it — and the guard stays green.

**Why it matters:** The secret-material half of the check is indirection-proof, but the structural half is not, and the issue's "Guard" section asks for both ("rejecting any non-fixture **or** context-less `*_private_key` value", see [bitcoin-sv/teranode#4739](https://github.com/bitcoin-sv/teranode/issues/4739)). Because fixture values are allowlisted by design, the reintroduction path costs one extra line.

**Fix:** Build the name→value map for the file first, resolve a single `${VAR}` hop against it before applying both checks, and keep the context assertion keyed on the *referencing* name.

```go
resolved := value
if m := varRef.FindStringSubmatch(value); m != nil { // varRef = `^\$\{([^}]+)\}$`
    if v, ok := defs[m[1]]; ok {
        resolved = v
    }
}
```

### ChiR6 — Hardcoded file list will silently stop covering new settings files

*Same as [@liam's point "committedSettingsFiles is a hardcoded list" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1578#pullrequestreview-4960374755).*

**Problem:** `committedSettingsFiles` (`settings/committed_private_keys_test.go:55-61`) enumerates five paths. Renames fail loudly (`require.NoError` on the read), so there is no silent breakage there — but a newly committed settings file is simply never scanned, with no signal. `compose/cmd/gennodes/templates/settings.conf.tmpl` is already a committed settings-format file outside the list; it carries only `{{ .PrivateKey }}` placeholders today, so nothing leaks, but it shows the class exists.

**Why it matters:** A drift-prone allowlist of scan targets undercuts a guard whose whole value is that it keeps running unattended after everyone has forgotten why it exists.

**Fix:** Discover targets instead of listing them — walk the repo root for `settings*.conf`, `settings*.conf.template` and `*.conf.tmpl`, skipping `vendor/`, `node_modules/` and generated output dirs. Keep an explicit assertion that the five known files were among those found, so a discovery bug cannot silently reduce the guard to zero files.

### ChiR7 — One key is still allowlisted by name, against this PR's own principle

**Problem:** `.gitleaks.toml:53` keeps `'''COINBASE_WALLET_PRIVATE_KEY'''` — a name-based allowlist with `regexTarget = "line"`, so *any* line mentioning that setting is exempt from every rule. The PR replaces the other name-based entries with the specific fixture values precisely because "a real key committed under any `*_private_key` setting is still caught by the rule above" (`.gitleaks.toml:38-43`). The three values behind that name are known — the same WIFs already in `allowedFixtures` (`settings/committed_private_keys_test.go:45-47`) and in `compose/settings_test.conf:402-404`. Relatedly, and as Liam noted, the new rule is hex-only, so WIFs under a `*_private_key` name are never matched by the diff scan at all.

**Why it matters:** It is the one remaining "add a real secret under this exact name and the scan stays quiet" hole, and it sits on a *wallet* key rather than an identity key. Inside the five scanned files the Go guard still catches it; anywhere else (a compose `environment:` block, a new deploy overlay) nothing does.

**Fix:** Swap the name for the three values, matching the treatment the p2p fixtures got.

```toml
  # Dev test wallet WIFs for the docker blaster stacks (compose/settings_test.conf).
  '''L56TgyTpDdvL3W24SMoALYotibToSCySQeo4pThLKxw6EFR6f93Q''',
  '''KyAwSjuXZNgj78w3W7mR1fVMbPFu2heaCJJkWK5Yy58NZ4xafV6k''',
  '''L3NVjmwg3nC7ZPrwMVF6FXiG1a1RZ89nhizmJVctGztRKLYrhtFL''',
```

Optionally add a sibling rule with the WIF shape (`[5KL][1-9A-HJ-NP-Za-km-z]{50,51}`) so the diff scan and the Go guard cover the same value shapes.

### ChiR8 — The tracker issue still describes the abandoned pinned-key approach

*Same as [@liam's point "The description no longer matches the diff" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1578#pullrequestreview-4960374755).*

**Problem:** Liam's before-merge request was that the PR body still claimed the gennodes overlay "pins a per-node `alert_p2p_private_key`". The body has since been updated and now correctly reads "Turned alert off in the gennodes multinode rig", so his literal request is satisfied. The same stale claim survives one level up: the "Status (2026-08-18)" section of [bitcoin-sv/teranode#4739](https://github.com/bitcoin-sv/teranode/issues/4739) says this PR "pins per-node alert keys in the gennodes overlay". The diff does the opposite — `templates/settings.conf.tmpl:20-23` sets `startAlert.docker.teranodeN = false` and pins no key, and `compose/cmd/gennodes/main.go:263` drops `-alert=1`.

**Why it matters:** That status section is the record that outlives the PR, and it is on the issue this PR closes. Anyone auditing later reads "alert keys were pinned" and goes looking for keys that were never added. It also matters for ChiR1: the same section is where the rotation action lives, so it is worth getting right in the same edit.

**Fix:** Update the issue's status section to match: alert is disabled in the gennodes rig (`startAlert.docker.teranodeN = false`, `-alert=1` dropped from the core sidecar), no per-node alert key is pinned.

### ChiR9 — `multinode.sh` help text still advertises alert in the core sidecar

**Problem:** `compose/multinode.sh:26` still reads `sidecar for rpc/alert/blockpersister/etc.`. The same wording was updated in `MULTINODE.md`, `main.go:174`, `main_test.go` and `docker-compose-split.yml.tmpl`, but this copy was missed.

**Why it matters:** It is the string an operator actually sees when running `multinode.sh` with no arguments, so it is the most likely place to mislead someone into looking for an alert container that is no longer generated.

**Fix:** Drop `alert/` from the usage string.

```sh
                             sidecar for rpc/blockpersister/etc.). Heavier
```

## Recap

| --- | --- | --- | --- |
| ChiR1 | Rotation tracking lost on merge | YES | 75% |
| ChiR2 | 96-byte key form not matched | YES | 60% |
| ChiR3 | Pipe-separated key lists unchecked | | 48% |
| ChiR4 | Parser diverges from gocore | | 42% |
| ChiR5 | Context-less check dodges indirection | | 36% |
| ChiR6 | Hardcoded scan target list | | 28% |
| ChiR7 | Name-based allowlist entry remains | | 22% |
| ChiR8 | Issue status describes wrong approach | | 15% |
| ChiR9 | Stale alert mention in usage text | | 10% |

ctnguyen (changes requested):
Review at commit [ccb2658](https://github.com/gokutheengineer/teranode/commit/ccb2658772d86fc779ac41fa8313a91b271f90e5)

This round did real work. The guard is a much better control than it was: `parseAssignment` (`settings/committed_private_keys_test.go:88-107`) is now a faithful copy of gocore's parse loop — first `#` wins, key is everything left of the first `=`, both sides trimmed — so the parser-divergence bypasses are genuinely closed; `hexKeyValue` at `{64,}` with the 66/130 exclusion covers the 96-byte libp2p form that `crypto.UnmarshalEd25519PrivateKey` accepts; the `|` split matches what `GetMulti` actually does; and discovery via `git ls-files` is a better answer than the filesystem walk I suggested, because it structurally cannot reach a developer's gitignored `settings_local.conf`. Switching the gitleaks allowlist from key names to specific values is the right principle too. Seven of the nine previous points are done. What is left is two loose ends from last round and one gap the design opens on purpose: the guard reads a settings file the way gocore reads it, and gocore does not read comments.

## Status of previous points

Three caveats. `ChiR5` is materially better — a one-hop `${VAR}` is now resolved before both checks, so `alert_p2p_private_key = ${FIXTURE}` does fail the context-less assertion — but gocore's `replaceVariables` (`github.com/ordishs/gocore`, pinned at `go.mod:41`; `config.go:553-572`) loops until no `${}` tokens remain and re-enters `getInternal` for each, so it resolves nested references; the guard's single hop does not, and a second hop still walks past the structural ban. `ChiR7` is untouched: `.gitleaks.toml:53` still carries `'''COINBASE_WALLET_PRIVATE_KEY'''`, a name-based entry with `regexTarget = "line"`, and no WIF-shaped rule was added. `ChiR9` is untouched: `compose/multinode.sh:26` still reads `sidecar for rpc/alert/blockpersister/etc.` while every other copy of that sentence was updated.

| ChiR1 | Rotation tracking lost on merge | YES | FIXED |
| ChiR2 | 96-byte key form not matched | YES | FIXED |
| ChiR3 | Pipe-separated key lists unchecked |  | FIXED |
| ChiR4 | Parser diverges from gocore |  | FIXED |
| ChiR5 | Context-less check dodges indirection |  | PARTIAL FIXED |
| ChiR6 | Hardcoded scan target list |  | FIXED |
| ChiR7 | Name-based allowlist entry remains |  | NOT FIXED |
| ChiR8 | Issue status describes wrong approach |  | FIXED |
| ChiR9 | Stale alert mention in usage text |  | NOT FIXED |

## Blocking issues

### ChiR10 — The guard cannot see commented-out key material, and `settings.conf` still holds a 192-hex key behind a `#`

**Problem:** `parseAssignment` starts with `strings.Cut(line, "#")` (`settings/committed_private_keys_test.go:89`). For a line that begins with `#` the remainder is empty, the `=` cut fails, and the line is dropped before any shape check runs — it never enters `defs` either. `settings.conf:1292` is exactly that line:

```ini
settings.conf:1290:  # tx_blaster_p2p_peer_id = 12D3KooWLRaBuYhHPcj8ixqwiVLtJ6YxMMSJRDwPYKVN5HWSnUWo
settings.conf:1292:  # tx_blaster_p2p_private_key = f19576f6a8ea51ebf5a1fb05cdfea7b66bfa8ecdcec58232d064f3343fb19ab29d99c9e7339b479ed4ee48d7bb7a078bcc9e2ad2355e7e6ba94297e97ecb7744
```

192 hex characters — the legacy libp2p form this PR widened `hexKeyValue` to cover — sitting in a committed public file. The other half of the control is now blind to it too: `.gitleaks.toml:44-51` adds its first 64 characters to the value allowlist, under a comment (`:38-43`) that names its provenance as "test/test_settings.go, compose/settings_test.conf, compose/cmd/genpeerkeys/main.go". I searched the tree: `f19576f6…fb19ab2` occurs in exactly two places, `settings.conf:1292` and that allowlist entry. It is in none of the three files the comment cites. That is the same evidential position the PR body uses to classify `p2p_private_key.operator.mainnet.stage.1` as a real operator secret — "present nowhere else in the repo". I cannot establish whether this value is real or a throwaway; what I can establish is that after this merge no control in the repo will look at it again.

**Why it matters:** The guard is billed as the standing full-file complement to a scan that only sees PR diffs, and commenting a key out is the single most common way one is left behind in a config file. gocore-parity comment stripping is right for *resolution* semantics and wrong for a *leak* guard: a key behind a `#` is byte-for-byte as exposed as a live one, and this is the file the PR exists to clean. The allowlist comment compounds it — it tells a future reader the value was verified against three fixture sources when it was not, which is the exact review step that would otherwise catch a real key being allowlisted.

**Fix:** Run the value-shape check over the raw line before comment stripping, and keep the parsed form only for the name/context assertions. No new imports:

```go
// Package-level, next to the other patterns:
var keyShapedToken = regexp.MustCompile(`[0-9a-zA-Z]{50,}`)

// Inside the per-line loop, before `parseAssignment`:
for _, tok := range keyShapedToken.FindAllString(raw, -1) {
    if !isKeyShapedHex(tok) && !wifKeyValue.MatchString(tok) {
        continue
    }
    require.True(t, allowedFixtures[tok],
        "%s:%d: committed private-key-shaped value that is not a known public test fixture (commented-out keys count - they are just as committed); if this leaked, rotate it",
        rel, lineNo)
}
```

Then delete `settings.conf:1290-1292` and drop the matching `.gitleaks.toml` entry — the setting is commented out, so removing it changes no behaviour — or, if the value really is a published fixture, say where it comes from next to the allowlist entry rather than folding it into a comment about three files it does not appear in.

## Non-blocking issues

### ChiR11 — `${VAR}` resolution is one hop and whole-value; gocore's is recursive and substring

**Problem:** `varRefValue` is anchored `^\$\{([^}]+)\}$` (`settings/committed_private_keys_test.go:75`) and is applied once per part (`:176-180`), so the guard resolves a value that is *entirely* a single reference, one level deep. gocore's `replaceVariables` (`github.com/ordishs/gocore`, pinned at `go.mod:41`; `config.go:553-572`) is a `for` loop over `FindAllString` that substitutes every `${…}` token found anywhere in the value and repeats until none remain, with each lookup going through `getInternal` — which re-enters resolution. The gap that matters is concatenation rather than nesting: fragments below the 64-character floor are individually invisible, and the assembled reference matches neither regex.

```ini
A = a4a361d2e1d3b73a4a50857794bf0a2d   # 32 hex - under the floor, skipped
B = d148bf18531203dc403220f2365e06e3
C = c12bcc09668d9603cd9f0e223ca101a5
D = 9e9677721bd4d85a21f24d6faf79f22e
p2p_private_key.operator = ${A}${B}${C}${D}   # not `^\${...}$`, so never resolved
```

gocore hands the consumer a working 128-hex key; the guard passes. The diff scan does not cover it either — `.gitleaks.toml:34` exempts the whole line because it contains a placeholder, and the new rule needs `{64,}` of hex immediately after the `=`. Bounding this honestly: I checked the five scanned files and no such construction exists today, so this is a hardening gap in a control the body advertises as indirection-proof, not an occurrence. The nested-hop variant (`${A}` → `${B}` → fixture) walking past the context-less assertion is the remainder of `ChiR5`, not part of this item.

**Why it matters:** The PR body sells value-shape checking on the grounds that it survives indirection, and this is the one indirection it does not survive — a gap in the guarantee rather than a live exposure, which is why it is not blocking. Worth closing while the resolver is fresh, since the cost is a few lines now and a re-derivation later.

**Fix:** Resolve the way gocore does — substring substitution, iterated to a fixed point with a small bound.

```go
varRefAny := regexp.MustCompile(`\$\{[^}]+\}`)

resolve := func(v string) string {
    for i := 0; i < 8 && strings.Contains(v, "${"); i++ {
        next := varRefAny.ReplaceAllStringFunc(v, func(m string) string {
            if r, ok := defs[m[2:len(m)-1]]; ok {
                return r
            }
            return m
        })
        if next == v {
            break
        }
        v = next
    }
    return v
}
```

Apply it to the whole value before the `|` split, then shape-check each part as now. One deliberate divergence to note in the comment: gocore replaces an unresolvable `${X}` with the literal `{UNKNOWN}`, which is how its loop terminates. Leaving the token in place, as above, is the right call for a guard — a definition the guard could not see should not be silently rewritten into something benign — so the fixed-point check is what stops the loop instead.

### ChiR12 — The placeholder allowlist exempts any line that merely *mentions* a variable

**Problem:** `.gitleaks.toml:34` is `'''\$\{[A-Za-z0-9_]+\}'''` under `regexTarget = "line"`, so a single placeholder anywhere on a line suppresses every rule for that line — including the new `teranode-settings-private-key` rule added at `:22-25`. The concrete case is the one the guard's `|` splitting was added for. `settings.conf:849` reads:

```ini
miner_wallet_private_keys.operator.mainnet     = ${PK1} | ${PK2} | ${PK3}
```

Substituting a literal for one element leaves `${PK2}` on the line, so the diff scan stays silent on the literal. Inside the five files the Go guard still catches it; in a `deploy/` overlay, a compose `environment:` block or any other file the guard does not scan, nothing does.

**Why it matters:** This PR's stated reason for rewriting this allowlist was to stop exempting by name so that "a real key committed under any `*_private_key` setting is still caught". Exempting by "the line mentions a placeholder" is the same category of hole one level up, and it sits directly under the block the PR rewrote. It is pre-existing, which is why it is not blocking.

**Fix:** Anchor the exemption to a value that is *entirely* placeholders, so it stops applying the moment a literal appears next to one.

```toml
  # ${ENV_VAR} placeholders resolved at runtime (not literal secrets). Anchored
  # to the value position: a line mixing a placeholder with a literal is scanned.
  '''=\s*"?\$\{[A-Za-z0-9_]+\}(\s*\|\s*\$\{[A-Za-z0-9_]+\})*"?\s*$''',
```

## Recap

| ChiR10 | Commented-out key material unguarded | YES | 60% |
| ChiR11 | Variable resolution shallower than gocore |  | 35% |
| ChiR12 | Placeholder line allowlist too broad |  | 28% |

ctnguyen (approved):
Review at commit [b89460d](https://github.com/gokutheengineer/teranode/commit/b89460d6ab06918c14d64a91428825243e4ab0f8)

This round closes the two points I raised last time, and it closes them properly rather than narrowly. The raw-line pass (`settings/committed_private_keys_test.go:197-205`) walks the raw line for key-shaped tokens and fixture-checks each one that survives the hex-or-WIF shape filter at `:198`, before comment stripping; the commented `tx_blaster` pair is gone from `settings.conf`; and no tracked file still carries the value `f19576f6…7744` — I grepped for it. The `.gitleaks.toml` allowlist entry went with it, and all seven remaining value entries (`.gitleaks.toml:44-50`) really are documented in `allowedFixtures` — I matched each of the seven against the map. `resolve` (`:175-189`) is now the substring-substitution-to-a-fixed-point shape that gocore's `replaceVariables` uses (`config.go:553-570` in gocore v1.0.82, pinned at `go.mod:41`), with the `{UNKNOWN}` divergence argued for rather than glossed over. A side benefit worth noticing: because the raw pass looks at tokens rather than at the parsed value, the shape check is now indifferent to quoting and to any non-alphanumeric decoration around the value, which retires a whole family of parser nits at once (`ChiR17` below is the residue: decoration that *is* alphanumeric still merges into the token). I also enumerated, across the files discovery actually reaches, the tokens that pass that same hex-or-WIF filter, and each one is in `allowedFixtures`, so the guard is consistent with the tree it guards. And the path the alert change depends on fails loudly, not silently — `createPrivateKeyDirectory` (`services/alert/server.go:464-477`) returns a configuration error rather than falling through with no key. Nothing new here must be fixed before merge; what follows is the tail, plus three of the twelve that are still open.

## Status of previous points

Three still open. `ChiR7`: `.gitleaks.toml:52` still carries `'''COINBASE_WALLET_PRIVATE_KEY'''` under `regexTarget = "line"` and no WIF-shaped rule was added — sharper than last round, `compose/docker-compose-3blasters.yml:380` and `:556` set `COINBASE_WALLET_PRIVATE_KEY: "L56Tgy…f93Q"` in an `environment:` block, that file is not among the seven the Go guard discovers, and every line naming that setting is exempt from every rule, so swapping the fixture WIF for a real one there is invisible to both halves of the control. `ChiR9`: `compose/multinode.sh:26` still reads `sidecar for rpc/alert/blockpersister/etc.`. `ChiR12`: `.gitleaks.toml:34` is unchanged and `settings.conf:865` is still the case in point. Two judgements to declare rather than bury: `ChiR6` and `ChiR11` are fixed as asked, and the residues I found in their replacements (the `git ls-files` fallback, per-file variable resolution) are raised fresh as `ChiR16` and `ChiR14` because they are properties of the new implementations, not remainders of the old defects. On `ChiR8`, the PR body now reads "Turned alert off in the gennodes multinode rig", so @liam's request is satisfied; one drift to watch is that the issue — as the context records it — still describes the guard's indirection handling as one-hop while `resolve` is now fixed-point, which is the same class of record drift and the same one-line edit. This all rests on reading the tree at `b89460d` and grepping it; I did not run the guard.

| ChiR1 | Rotation tracking lost on merge | YES | FIXED |
| ChiR2 | 96-byte key form not matched | YES | FIXED |
| ChiR3 | Pipe-separated key lists unchecked |  | FIXED |
| ChiR4 | Parser diverges from gocore |  | FIXED |
| ChiR5 | Context-less check dodges indirection |  | FIXED |
| ChiR6 | Hardcoded scan target list |  | FIXED |
| ChiR7 | Name-based allowlist entry remains |  | NOT FIXED |
| ChiR8 | Issue status describes wrong approach |  | FIXED |
| ChiR9 | Stale alert mention in usage text |  | NOT FIXED |
| ChiR10 | Commented-out key material unguarded | YES | FIXED |
| ChiR11 | Variable resolution shallower than gocore |  | FIXED |
| ChiR12 | Placeholder line allowlist too broad |  | NOT FIXED |

## Non-blocking issues

### ChiR13 — The second removed key has no exposure record anywhere

**Problem:** `settings.conf:1305-1309` replaces the commented-out pair with a note saying the value's provenance could not be established, and the PR body reaches that conclusion using the same evidential test it used on the operator key: it appeared nowhere else in the repo. HEAD is clean — I confirmed the value is gone from the tree and from the gitleaks allowlist. What is missing is the other half. [bitcoin-sv/teranode#4739](https://github.com/bitcoin-sv/teranode/issues/4739) carries a rotation action for `p2p_private_key.operator.mainnet.stage.1` and nothing else, and the PR body's Residual section lists five follow-ups without this one among them. Bounding the concern honestly: `grep -rn "tx_blaster" --include=*.go` over this tree returns nothing, so no Go code reads any `tx_blaster_*` setting — the value was inert twice over, commented out and unconsumed, and I cannot establish that it was ever used as an identity. What I can establish is that a peer ID was pinned beside it, which is the artefact of having actually run something, and that after this merge no control in the repo and no ticket outside it looks at that value again.

**Why it matters:** `ChiR1` set this PR's own standard: removal from HEAD is the cosmetic half, and the record has to outlive the merge. Applying that standard to the key with a known context and not to the key whose provenance is unknown inverts the priority — the second one is the one nobody can vouch for. The cost of closing it is a bullet on an issue that is already staying open.

**Fix:** Add the check to #&NoBreak;4739's residual list, alongside the operator-key action. Record the peer ID and the commit that removed the line rather than the key material, so the ticket does not republish it:

```
Also removed: a commented-out tx_blaster libp2p keypair, peer ID
12D3KooWLRaBuYhHPcj8ixqwiVLtJ6YxMMSJRDwPYKVN5HWSnUWo (value in the
history of settings.conf, removed in 2e30bbe). No tx_blaster_* setting
has a Go consumer, so it was never read by the node. Action: confirm
with whoever generated it whether it was ever used anywhere; if yes,
treat as compromised.
```

### ChiR14 — `${VAR}` resolution is per-file, but gocore merges the committed settings files into one table

**Problem:** `defs` is built inside the per-file loop (`settings/committed_private_keys_test.go:158-166`), so it is reset for every scan target and `resolve` (`:175-189`) can only see definitions from the file it is currently reading. gocore does not work that way: `Config` creates one map (`config.go:263`) and fills it by calling `processFile` for `settings.conf`, then `settings_test.conf`, then `settings_local.conf` (`:266`, `:290`, `:297`), and `replaceVariables` resolves each `${…}` through `getInternal` → `findValue` against that single merged map. Two of those files are committed and both are scan targets: `compose/docker-compose-3blasters.yml:168` and the other compose rigs mount `./settings_test.conf:/app/settings_local.conf` next to the `settings.conf` the Dockerfile copies in. The seam is live today — `compose/settings_test.conf:135-140` reference `${PK1}`/`${PK2}`/`${PK3}`, which are defined only at `settings.conf:855-857`, so at those six lines `resolve` returns the token unchanged and the line is skipped before either check runs. Nothing is hidden right now because those values are allowlisted fixtures, but it means the `ChiR11` concatenation bypass still works with the fragments and the reference split across the two files: put four 32-hex definitions (below the token floor, individually invisible) in `settings.conf` and `p2p_private_key.docker.teranodeN = ${A}${B}${C}${D}` in `compose/settings_test.conf`, and gocore hands the consumer a working 128-hex key that the guard never assembles.

**Why it matters:** This is the same guarantee `ChiR11` was about — the guard resolves indirection the way the consumer does — and the fix landed at the wrong scope. Cross-file is not an exotic layout here; it is how every docker rig is wired, which is why six lines already exercise it.

**Fix:** Build the definition table once across all discovered files, then run the per-line checking pass against it. Keeping the "any definition anywhere resolves" over-approximation is the right direction for a leak guard — it can only make more values visible, and the fixture allowlist absorbs the known ones.

```go
type scanFile struct {
    rel   string
    lines []string
}

var files []scanFile
defs := map[string]string{}

for _, rel := range discoverSettingsFiles(t, root) {
    content, err := os.ReadFile(filepath.Join(root, rel))
    require.NoError(t, err, "committed settings file must exist: %s", rel)

    lines := strings.Split(string(content), "\n")
    files = append(files, scanFile{rel: rel, lines: lines})

    for _, raw := range lines {
        if name, _, value, ok := parseAssignment(raw); ok {
            if v := strings.Trim(value, `"`); v != "" {
                defs[name] = v
            }
        }
    }
}
// then the existing per-line pass over `files`, resolving against the shared `defs`
```

### ChiR15 — The scan rule is gated on the setting name, so the value-first principle stops at the file boundary

*Same as [@liam's point "the new gitleaks rule is hex-only" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1578#pullrequestreview-4960374755).*

**Problem:** Liam framed this as an asymmetry between the two controls on WIFs and did not press it; the general form is worth stating. `.gitleaks.toml:22-25` fires only when the setting name ends in `private_key` or `private_keys` (optionally followed by `.context`) and the value is hex immediately after the `=`. The allowlist below it is value-based on purpose, and its comment states the reason: allowlist by value "so a real key committed under any `*_private_key` setting is still caught by the rule above" (`:38-43`). Detection is still name-based, so the principle is only half applied. The rule does not fire on a hex identity key under a name that does not end in `private_key` (`p2p_key`, `node_identity`), on a dash-spelled name (`p2p-private-key` — `[a-z0-9_]*private_keys?` needs the literal `_key`), or on a WIF under any name. Inside the seven files discovery reaches, the Go guard's value check still catches all three, because it ignores the name entirely. Outside them nothing does, and key material already lives outside them: `compose/docker-compose-3blasters.yml:380` carries a wallet WIF in an `environment:` block, and `compose/cmd/genpeerkeys/main.go` and `compose/cmd/gennodes/peer_keys.json` are path-allowlisted at `.gitleaks.toml:54-59`.

**Why it matters:** The diff scan is the only repo-wide control this PR adds; the Go guard is deliberately scoped to settings-format files. Leaving the scan name-gated means the repo-wide half is the one that depends on the attribute an author picks freely, which is the dependency the guard was rewritten to remove.

**Fix:** Add a value-shaped sibling rule that does not consult the setting name. It has to start at 128 hex rather than 64 — txids, block hashes and merkle roots are 64 hex and appear all over the repo, so a `{64,}` value rule would be pure noise. That still covers both forms `crypto.UnmarshalEd25519PrivateKey` accepts.

```toml
[[rules]]
id = "teranode-long-hex-secret"
description = "Long hex blob assigned to any setting or env var"
regex = '''(?i)^\s*[a-z0-9_.-]+\s*[:=]\s*"?(?:0x)?[0-9a-f]{128,}'''
```

The WIF half belongs with `ChiR7`, which already asks for a WIF-shaped rule alongside dropping the name-based allowlist entry.

### ChiR16 — The git-unavailable fallback narrows the scan set and skips its own coverage assertion

**Problem:** `discoverSettingsFiles` returns `knownSettingsFiles` on any `git ls-files` error (`settings/committed_private_keys_test.go:119-123`), and it returns *before* the `require.Contains` loop at `:137-139`. When git works, the same filter also yields the two tracked `*.conf.tmpl` files — `compose/cmd/gennodes/templates/settings.conf.tmpl` and `compose/cmd/gennodes/templates/aerospike.conf.tmpl`; I listed the tracked `.conf`-family files to check. So the fallback list (`:58-64`) is a strict subset of what discovery finds. In a checkout with no `.git` or no `git` binary, the guard silently drops those two files, skips the assertion whose stated job is to make a discovery bug loud, and still passes green — `t.Logf` output is not printed for a passing test unless `-v` is set. The comment at `:56-57` claims a discovery bug "can therefore never silently reduce the guard"; the fallback is the one path where it can.

**Why it matters:** Small, but it is the guard's own escape hatch, and a green result that means "scanned five of seven files" is exactly the outcome the `require.Contains` assertion exists to prevent.

**Fix:** Make the fallback non-shrinking and non-silent — add the two templates to `knownSettingsFiles`, and fail rather than log, since every CI checkout has git.

```go
out, err := exec.Command("git", "-C", root, "ls-files").Output()
require.NoError(t, err, "git ls-files must succeed: discovery cannot be verified without it")
```

### ChiR17 — A decorated hex value slips both shape checks

**Problem:** Both passes require the candidate to be *entirely* hex. `hexKeyValue` is `^[0-9a-fA-F]{64,}$` (`settings/committed_private_keys_test.go:71`) and `keyShapedToken` is `[0-9a-zA-Z]{50,}` (`:80`), which matches maximal alphanumeric runs. A `0x` prefix is alphanumeric, so `0x` followed by 128 hex is one 130-character token that is not pure hex: `isKeyShapedHex` rejects it on the `x`, `wifKeyValue` rejects it because `0` is not in `[5KL9c]`, and the parsed pass rejects the same value for the same two reasons. A line of the form

    settings.conf:  p2p_private_key.operator.prod = 0x<128 hex>

is therefore invisible to the guard, and `.gitleaks.toml:22-25` does not catch it either, since its value pattern is `"?[0-9a-f]{64,}` immediately after the `=`. To be clear about the blast radius: that value would not work as a key, because `hex.DecodeString` rejects the prefix — this is an evasion of the leak check, not a working-identity path.

**Why it matters:** `0x`-prefixed hex is a common paste form, and the guard's job is to notice committed key material regardless of whether it happens to load. A value that leaks a real key but fails to start the node is still a leaked key.

**Fix:** Strip an optional `0x`/`0X` before shape-checking, in the one helper both passes go through.

```go
func isKeyShapedHex(v string) bool {
    if len(v) > 2 && v[0] == '0' && (v[1] == 'x' || v[1] == 'X') {
        v = v[2:]
    }

    return hexKeyValue.MatchString(v) && len(v) != 66 && len(v) != 130
}
```

## Recap

| ChiR13 | No record for second key |  | 42% |
| ChiR14 | Variable table is per-file |  | 35% |
| ChiR15 | Scan rule gated on name |  | 26% |
| ChiR16 | Git fallback narrows coverage |  | 18% |
| ChiR17 | Decorated hex evades shape check |  | 10% |

PR #1589, opened 2026-08-12
test(e2e): port bitcoin-sv functional tests

sugh01:
port bitcoin-sv functional tests with a wire-peer harness and gap register

Ports bitcoin-sv functional test scripts onto Teranode's in-process TestDaemon, and records every obstacle that blocked a port in a machine-checked register rather than losing it in commit messages. 31 of 79 portable scripts are ported (1 full, 30 with at least one assertion waived and a reason recorded).

What is here:

- test/utils/wirepeer: a legacy-wire test peer replacing bitcoin-sv's mininode.py. Speaks the BSV wire protocol against a running daemon, including hand-framed messages for the malformed-input cases upstream relies on.

- test/e2e/daemon/bsv: 30 ported scripts. Where Teranode diverges from bitcoin-sv, the port asserts what Teranode actually does, usually as a tripwire subtest that fails the day the divergence is fixed, so no waiver can go stale silently.

- test/e2e/daemon/bsv/registry.yaml: the source of truth. Per-script porting status and a gaps block holding 35 findings, each with what was measured, what was only read from code, and what was not established.

- test/e2e/daemon/bsv/GAPS.md: generated from that registry. Hand edits fail `make test`, so the document cannot drift from the data.

- GOAL.md and PORTING.md: the brief and the method, including the rule that a porting change never alters node behaviour. A port that fails because Teranode is wrong records a gap; it does not fix the node.

Makefile targets: bsvportingcheck (registry validation, wired into `make test` since it needs no daemon), bsvporttest, bsvportingstatus, bsvportinggaps and bsvportinggapsdoc.

17 of the gaps are confirmed defects now filed as #4769-#4785, ranging from a consensus divergence at activation boundaries to an 89-byte block message that can drive tens of gigabytes of allocation. Their registry entries carry the issue links. 7 confirmed defects remain unfiled pending review, and the rest are product decisions, open questions, or ours to fix here.

No production code is touched.


liam:
Heads-up on the `reconsiderblock-error-contract` tripwires: #1637 fixes that gap, so they are about to fire. Flagging rather than editing this branch.

`reconsiderblock` on an unknown block now returns exactly what upstream asserts:

```
code    -5
message Block not found
```

## Three assertions break, not two

Worth being precise, because the third one is not labelled as a tripwire and reads as the safe half:

**`bsv_command_line_invalid_block_test.go:130`** — described as *"The half that matches: it fails, and it says the block was not found."*

```go
require.Contains(t, message, "BLOCK_NOT_FOUND",
    "the failure should be attributed to the block being unknown")
```

This one breaks too. It matches because the internal error code string was leaking into the RPC message — it is asserting on `BLOCK_NOT_FOUND (10)` from the wrapped chain, not on upstream's text. Once the message is bitcoind's `Block not found`, the uppercase underscored form is gone and `Contains` fails. So the half that "matches" was matching on the leak the other tripwire exists to catch.

**`:134`** and **`:137`** fire as designed:

```go
require.EqualValues(t, -25, code, ...)                          // now -5
require.Contains(t, message, "sql: no rows in result set", ...) // now absent
```

## Replacement

All three collapse into the upstream assertion the port was written against:

```go
// Upstream: assert_raises_rpc_error(-5, 'Block not found', reconsiderblock, hash).
require.EqualValues(t, -5, code)
require.Equal(t, "Block not found", message)
```

## registry.yaml

Two places, if you want them in the same pass:

- `:71` — the `reconsiderblock-error-contract` entry, `status: open`
- `:2361` — the deviation note under this port, which describes the -25 / four-level-chain behaviour and says it is tripwired on both

## Sequencing

Whichever of the two merges second needs the edit; there is no ordering constraint beyond that. If this one lands first I am happy to do it in #1637 instead — say the word and I will, rather than leaving you a broken test to discover.

For context on the fix: the message is trimmed rather than replaced, so the paths where a caller genuinely needs the reason keep it (`TX rejected: script execution failed for input 0`) while the wrapped internals drop away. Not-found classes map to bitcoind's code and text. Details in #1637 and bitcoin-sv/teranode#4778.

PR #1593, opened 2026-08-13
feat(blockassembly): limit transactions held in RAM

ordishs:
Issue: https://github.com/bitcoin-sv/teranode/issues/4459

## Problem

When block assembly is not producing blocks, validated transactions pile up in RAM without bound and the node eventually exhausts memory.

## Approach

Block assembly measures how many transactions it holds and publishes a full/not-full flag over the blockchain notification bus. The transaction ingress points read that flag and stop accepting new transactions while it is set.

Three decisions worth reviewing:

**Not an FSM state.** Fullness is orthogonal to the node lifecycle. A new FSM state would be mutually exclusive with `RUNNING` and `CATCHINGBLOCKS`, so a full node would stop syncing and `CATCHUPBLOCKS` could not fire. The flag rides the same notification plumbing and is cached in `blockchain.Client` next to `fmsState`, so reading it is an atomic load with no RPC — cheap enough for the per-transaction path. A full node stays `RUNNING`.

**Block assembly never refuses.** It only measures and publishes. Transactions already in flight complete normally and the count may exceed the limit; the node reacts at ingestion instead, which is where a client can actually be told. This also means the validator's `sendToBlockAssembler` is never made to fail, so none of the already-applied UTXO spends need unwinding.

**Two watermarks.** The flag sets at the limit and clears only once the count falls back to the resume watermark. Without that hysteresis a node sitting at the limit would flap on every transaction and storm the notification bus.

## Changes

- **Settings** — `blockassembly_maxTransactionsInMemory` and `blockassembly_maxTransactionsInMemoryResume`. Both default to `0`, which disables the limit and preserves current behaviour for existing deployments.
- **Measurement** — `TransactionsInMemory()` = transactions in subtrees + queue length. Queued transactions have to be counted, or block assembly under-reports what it holds and sails past the limit.
- **Broadcast** — new `NotificationType_BlockAssemblyFull` (enum 9), published on transition plus a 10s heartbeat so a subscriber that starts or reconnects after a transition converges rather than holding its default. The metadata contract lives in one place (`services/blockchain/block_assembly_full.go`): `NewBlockAssemblyFullNotification` builds it, `blockAssemblyFullFromNotification` reads it, so producer and consumers cannot drift on the key or its encoding. Parsing fails safe — an unparseable or missing value reads as *not full*, so a malformed notification cannot turn into an ingress outage.
- **Ingress gates** — all three client-transaction entry points are gated:

  | Path | Gated | Note |
  |------|-------|------|
  | Propagation `processTransactionInternal` (gRPC, HTTP, UDP) | Yes | Checked *before* storing, so a refused transaction costs no storage |
  | Legacy netsync `handleTxMsg` | Yes | Reaches the validator directly, bypassing propagation |
  | `sendrawtransaction` RPC | Yes | Also stores + validates directly; check sits before `txStore.Set` |
  | Transactions arriving inside a block | **No** | Blocks must validate regardless of block assembly pressure |

- **Observability** — `teranode_blockassembly_tx_ingress_full` gauge, rejection counters in all three ingress services, and `txIngressFull` on `GetBlockAssemblyState`.

A dropped legacy transaction is **not** added to `rejectedTxns`, because a full block assembly is transient rather than a fault in the transaction. The requested-transaction bookkeeping is cleared so a later inv can fetch it again once there is room.

Refusals are logged at **debug**, not error. All three propagation call sites route through `logTxProcessingError`; left at error level, a sustained inbound stream against a full node would take one error line per transaction, amplifying the exact overload the limit exists to contain and burying real errors. Classifying a refusal needs more than `errors.Is`, because the UDP worker goes through the gRPC handler (which flattens the error into a status) and `ERR_SERVICE_UNAVAILABLE` maps to `codes.Internal` rather than `codes.Unavailable`; `isTransientBackpressure` reconstructs the application code from the attached detail and treats the batch-handler admission control's bare `Unavailable` the same way. The gRPC code mapping is deliberately left alone — changing it would make the client retry gate start retrying these.

## Sizing

Budget roughly 1KB per transaction, so 1 billion needs about 1TB.

This is a **soft** limit, and the overshoot is **not** bounded by the limit itself. The ingress points stop publishing, but the validator consumer (`services/validator/Server.go:368`) keeps draining whatever is already in its Kafka topic into block assembly. The overshoot is therefore bounded by **validator consumer lag** at the moment the limit is crossed. Size the limit below available RAM by at least the transaction count that lag represents. This is documented in `settings.conf` and in the setting's longdesc.

## Scope

Ingress only. The startup load path (`loadUnminedTransactions`) is unchanged — `blockassembly_unminedTxDiskSortEnabled` and `blockassembly_unminedLoadingBatchSize` already address the restart side of the original issue. Once the cap is configured the unmined set cannot grow past it going forward.

## Testing

- `golangci-lint run ./services/... ./settings/... --new-from-rev bsv/main` → `0 issues.`
- `go build ./...` and `go vet` on the touched trees → clean.
- Feature tests under `-race -tags testtxmetacache` → all pass.

Tests cover: hysteresis across both watermarks and the disabled default; the measurement including queued transactions; the notification contract and its round trip through both `Client` and `LocalClient`; the shared parser including the fail-safe cases; the full monitor loop against a real subtree processor; the backpressure classification across all three error forms plus three must-still-be-errors cases; and all three ingress gates, each asserting the accept case as well as the refuse case.

The propagation and RPC gate tests assert that a refused transaction never reaches the transaction store, and that the same transaction is accepted once block assembly reports room. The RPC test wires a validator stub that panics if called, so the gate short-circuiting before validation is proven rather than assumed. The legacy netsync test asserts the transaction never reaches validation while full, is not marked permanently rejected, and has its requested-transaction bookkeeping cleared.

The monitor intervals are injectable fields rather than constants, so the disabled-default test dropped from 12s of wall clock to 0.45s and the monitor-loop test from a 15s budget to 0.09s.


ordishs:
Design holds up: keeping fullness off the FSM, a cached atomic read on the per-transaction path, and two watermarks to stop bus flapping. Confirmed `setTxCountFromSubtrees` really does exclude the queue, so adding `QueueLength` is a correction rather than a double count, and the count drains on forward/backward moves so a healthy node cannot creep into a permanent full state. `go vet` clean on the touched trees; the new tests pass under `-race`.

Two items to fix before merge.

**services/propagation/Server.go:1321-1324** — the refusal is returned from `processTransactionInternal`, and every caller of that path logs it at error level: gRPC at `Server.go:1108`, the batch reader at `Server.go:1199`, the UDP multicast worker at `Server.go:563`.

- State: block assembly full, limit configured. Input: a sustained inbound transaction stream at normal propagation rates. Output: one error log line per transaction, for as long as the condition lasts.
- This amplifies the exact overload the feature exists to contain — the log pipeline becomes a second failure and real errors are buried.
- The existing batch-handler backpressure at `Server.go:1097` already does this correctly: it returns `codes.Unavailable` before the logging path and only increments a counter. Either move the check above the `Errorf`, or downgrade those three log sites when `errors.Is(err, errors.ErrServiceUnavailable)`. The Prometheus counter and the single transition-level `Warnf` already carry the operator signal.

**services/rpc/handlers.go:861-869, 909-911** — `handleSendRawTransaction` writes the transaction to `txStore` and then calls `validatorClient.Validate` directly, so it reaches block assembly without passing either new gate. It is not listed in the Scope section.

- State: block assembly full and refusing on both gated paths. Input: a `sendrawtransaction` RPC. Output: the transaction is stored and validated, block assembly grows past the configured limit, and the storage the propagation gate avoids is spent anyway.
- The RPC server already holds `blockchainClient blockchain.ClientI` (`services/rpc/server.go:682`), so this is the same check before the `txStore.Set` at line 861.
- If leaving RPC ungated is deliberate, say so in Scope and in the `blockassembly_maxTransactionsInMemory` longdesc — the setting currently reads as a cap on everything.

One sizing note worth folding into the docs rather than the code: the overshoot is bounded by validator Kafka lag, not by the limit. Propagation stops publishing, but the validator consumer (`services/validator/Server.go:368`) keeps draining whatever is already in the topic into block assembly. "Can briefly exceed this value" understates that, and operators sizing against available RAM need to know.

Smaller points, none of them blocking: the new method at `services/blockchain/Client.go:1857` was inserted under the `IsFSMCurrentState` doc comment, so both methods now have wrong godoc; the `"full"` metadata parse is duplicated in `Client.go:218` and `LocalClient.go:307` and only the `LocalClient` copy is exercised by tests, since every new test drives a `LocalClient`; `assert` is used in 9 places against the AGENTS.md preference for `require`; and `TestTxIngressLimitMonitorDoesNothingWhenDisabled` waits 12s of wall clock, which injectable intervals would return to `make test`.

freemans13 (commented):
Reviewed the full branch against `upstream/main`. No blockers. Six findings, all minor — one gap in the ingress enumeration, one misleading identifier, one degenerate-config edge, one benign race, one judgement call on an RPC code, and a docs nit.

Everything I could run locally is green: `go build ./...`, `go vet` on all seven touched trees, `golangci-lint run --new-from-rev upstream/main` (0 issues), and all 26 new tests under `-race -tags testtxmetacache`.

## Things I checked rather than assumed

Recording these because they are the parts of a change like this that fail quietly, and they all came back clean.

**The notification really reaches every client.** Two things had to be true. The server fan-out (`services/blockchain/Server.go:775-795`) is type-agnostic, so there is no allowlist to add the new type to. And the client's notification listener goroutine starts unconditionally inside `NewClientWithAddress` (`Client.go:189-215`), not on first `Subscribe` — so propagation, RPC and subtree validation all receive `BlockAssemblyFull` without registering as subscribers. The new `case` also consumes it ahead of the `default` fan-out, so no existing subscriber sees an unfamiliar type.

**No retry storm.** This was my main worry: a refusal classified as retryable and then hammered in a loop against a node already short of memory. It is safe. The gRPC retry interceptor (`util/grpc_helper.go:370`) retries only `codes.Unavailable` and `codes.DeadlineExceeded`, and `ERR_SERVICE_UNAVAILABLE` maps to `codes.Internal`, so it never fires. `errors.IsRetryableError` *does* treat the code as retryable (`errors/error_utils.go:34`), but its only two callers (`blockvalidation/quick_validate.go:1385`, `legacy/netsync/handle_block.go:1367`) sit on block-validation paths that never see the gate. The decision to leave the gRPC code mapping alone is what keeps this quiet, and the PR is right about why.

**`isTransientBackpressure`'s top-level-code argument holds.** I walked every return in `processTransactionInternal`: the coinbase check and `txSanityChecks` give `TxInvalidError`, `storeTransaction` is re-wrapped as `StorageError`, `validateTransactionViaHTTP` gives `ServiceError`, the direct validator call gives `ProcessingError`. The comment does not name `validateTransactionViaKafka`, which is the fifth exit — it returns `nil` or `ProcessingError`, so the claim survives, but the enumeration is one line short.

**The netsync early return is safe.** `txMsg` carries a `reply chan struct{}` a caller can block on, so an early return could in principle have wedged a peer goroutine. It does not — the dispatcher signals `reply` after `handleTxMsg` returns (`manager.go:2586-2593`), so every return path in the handler is already fire-and-forget.

**"Hold the refusal across a restart" actually works.** Starting the monitor before `loadUnminedTransactions` only helps if the count climbs *during* the reload rather than jumping at the end. It does: the reload adds via `AddDirectly` and `AddNodesDirectly`, both of which bump `txCount` incrementally (`SubtreeProcessor.go:2506`, `:2602`). On a backlog above the limit the flag re-asserts early in the reload, well inside the 60s TTL.

**The `setTxCountFromSubtrees` change is a comment fix, not a behaviour change.** The `+7 −4` looks like logic but the body is untouched. The old comment claimed the queue was included in the count; it never was. That stale comment is exactly what would have made someone size the limit wrong, so this is a good catch in passing.

Also confirmed both settings are genuinely read in `settings/settings.go:313-314` rather than only tagged on the struct, and that `NotificationType` 9 / `StateMessage` field 11 were both previously unused with the regenerated `.pb.go` containing only the additions plus realignment.

## Findings

### 1. Legacy orphan promotion is an ungated path into block assembly

`services/legacy/netsync/manager.go:1349`

`processOrphanTransactions` calls `validationClient.Validate` with default options, so `AddTXToBlockAssembly` is true and those transactions enter the mining template. It has two callers. From `handleTxMsg` it is transitively gated — the new check returns before reaching it. But `handle_block.go:270` also calls it, from the block path, in a background goroutine, and that one is not gated.

The "transactions arriving inside a block are never gated" rationale does not cover these. They are not in the block — they are mempool transactions that were orphaned and whose missing parent happened to arrive in the block. They are a mempool admission, gated nowhere.

The RAM impact is negligible and I would not change the behaviour for it: the pool is capped by `legacy_maxOrphanTxs`, default 100 (`settings/settings.go:665`), so the worst case is 100 transactions per block against a limit sized in the hundreds of millions.

What I would change is the documentation. `Interface.go` enumerates the ingress points and tells future authors *"If you add another path that reaches block assembly, gate it here too"*. An enumeration that instructs people to keep it complete should be complete, or the next person will trust it. Either note the exception, or gate it — the cost is one atomic load on a path that runs at most once per block.

### 2. Metric identifier says transactions, counts subtrees

`services/subtreevalidation/metrics.go:57`

`prometheusSubtreeValidationTxsNotAddedToBlockAssemblyFull` is incremented once per subtree at `subtreeHandler.go:175`, not once per transaction. The exported name (`subtrees_not_added_to_block_assembly_full`) and the `Help` text both say subtrees correctly, so nothing an operator sees is wrong — only the Go identifier misleads, and it is what the next reader of the increment site will look at. Suggest `prometheusSubtreeValidationSubtreesNotAddedToBlockAssemblyFull`.

### 3. `resumeWatermark` can derive a watermark the count cannot reach

`services/blockassembly/BlockAssembler.go`, `resumeWatermark`

With `limit == 1` the guarded fallback returns `limit - 1`, which is 0. But `TransactionsInMemory()` includes the coinbase placeholder node that the first subtree always carries — its own doc comment says so — so the count can never be 0, `count <= resume` never holds, and the refusal never clears from block assembly's side. Ingress would reopen only via the client TTL, and the heartbeat would immediately re-assert it.

Degenerate config, and nobody will set a limit of 1. But it is the one input where the derived watermark is structurally unreachable, and the comment on that branch (*"only reachable for a limit of 1, where the only value below the limit is 0"*) spots the value without spotting that it is unreachable. Either floor the derived resume at 1, or say in the comment that single-digit limits are meaningless because of the placeholder.

### 4. Benign ordering race in the TTL expiry

`services/blockchain/block_assembly_full.go`, `set` / `isFull`

`set` stores `lastHeardAt` then swaps `full`; `isFull` reads `full` then `lastHeardAt`. A heartbeat landing between those two reads can lose to the expiry `CompareAndSwap`, clearing a refusal that was just refreshed.

It self-heals on the next heartbeat, so exposure is up to 10s of open ingress, and it is only reachable after a ≥60s notification gap — which is precisely the situation the code intends to fail open in anyway. Not worth a fix on its own. If you want it closed, storing `full` before `lastHeardAt` in `set` does it, since a stale-but-fresh-looking timestamp is the safe direction.

### 5. `ErrRPCOutOfMemory` (−7) deserves a deliberate decision

`services/rpc/handlers.go:884`

The reasoning against `txRejectedPrefix` and `ErrRPCInternal` is right and clearly argued: the transaction was not rejected and the node is not broken, so a broadcaster reading "TX rejected" drops it permanently, contradicting the 503 the HTTP surface returns for the same condition.

The specific replacement is the part I would want a second opinion on. In Core, −7 is `RPC_OUT_OF_MEMORY`, documented as running out of memory *during an operation*. Core's own answer for a node that will not take a transaction right now is `RPC_TRANSACTION_REJECTED` (−26) with a "mempool full" style message — which is a verdict, so the PR's objection applies to it too. There may be no clean fit in Core's code space and −7 is a defensible read, but no broadcaster's handling of −7 has been tested here, and the failure mode if one treats it as fatal is the same permanent drop the comment is trying to avoid. Worth an explicit call rather than inheriting it from the comment.

### 6. Five new metrics, no entry in the metrics reference

`docs/references/prometheusMetrics.md` gains nothing for the five new counters and gauges. That file is already drifting (39 of 45 block assembly metrics, 7 of 9 propagation metrics), so this is a pre-existing pattern rather than a regression here. Low priority, but it is five lines.

## On the three decisions the description flags

All three hold up.

**Not an FSM state** is right, and the reasoning is the important part: a mutually exclusive state would stop a full node syncing, turning a memory-pressure signal into a liveness failure. Riding the notification bus next to `fmsState` is also what makes the per-transaction read an atomic load, which is what makes gating the hot path viable at all.

**Block assembly never refuses, only measures** is the decision that keeps this change small. Because `sendToBlockAssembler` is never made to fail, no already-applied UTXO spends need unwinding — that is the part that would have been genuinely dangerous.

**Two watermarks** is necessary, and the reason bears restating: without hysteresis a node parked at the limit broadcasts a notification per transaction to every subscriber, and the server evicts subscribers whose buffers fill. Flapping would not just be noisy, it would evict the very subscribers that need the flag.

The one consequence I would keep in view operationally is the peer-announced subtree path. It does not refuse, so there is no resubmission: a transaction skipped there enters this node's template only when a block carrying it arrives, or after a restart. On a multi-node network that is where the volume is, so a long full window means this node's template diverges from its peers' for the duration. The PR states this plainly in three places, so it is a disclosed trade-off rather than a surprise — but it is the thing most likely to look like a bug to whoever is watching a full node.

icellan:
**P0 — peer-subtree gate can produce an unmineable template (child in, parent out)**
`services/subtreevalidation/subtreeHandler.go:173`

The three refusing gates are safe on this: nothing is stored or validated, so no child can follow a refused parent. This fourth gate is different — it isn't a refusal, and it breaks the invariant that keeps the template topologically valid.

- The invariant is enforced by the locked-flag two-phase commit, not by block assembly. `Validator.go:1552-1557` appends `WithLocked(true)` **only when `addToBlockAssembly` is true**; `twoPhaseCommitTransaction` (`Validator.go:1039`) unlocks only after block assembly accepted the tx; `TxLockedMaxRetries` (`Validator.go:478`) exists for the child-arrives-early race. `SubtreeProcessor.addNode` (`SubtreeProcessor.go:2115`) does no parent check at all.
- With `AddTXToBlockAssembly(false)` the tx is created **unlocked** (outputs and metadata still created — `SkipUtxoCreation` is a separate option) and never inserted into the template. Both halves of the guard are gone: nothing blocks the child, and there is no lock for the retry to observe.
- Scenario: full → peer subtree brings `P`, unlocked UTXOs, not in template, never revisited → flag clears → `C` spending `P` arrives via propagation, spends fine, enters template → we mine → `checkParentsExistOnChain` (`model/Block.go:1276`) fails, `P` has no on-chain block IDs. Peers reject the block.
- Not a narrow race: `P` stays out of the template for the process lifetime unless a block mines it, and the announcing peer may never mine that subtree.
- Every existing `AddTXToBlockAssembly(false)` site is a block or catchup context where the tx becomes mined, so its children's parents are on-chain. This is the first unmined, RUNNING-state one.

Two options: keep `utxo.WithLocked(true)` here so a child fails with `ErrTxLocked` — an honest refusal — instead of entering the template (verify that being mined clears `Locked`; `loadUnminedTransactions` already unlocks leftovers at `BlockAssembler.go:3357`); or re-enter the skipped set through the existing unmined-reload / `ResetWithInputValidation` topological path on the full→not-full transition. The second also fixes P1. The descendant case needs a test either way — the current test asserts only that the flag flips.

**P1 — skipped subtree transactions are never revisited**

Same site. No retry and no recovery until a block arrives or the process restarts, on the path the comment itself says carries most of the volume. A node that spends time at the cap mines templates materially poorer than the network's for as long as the process lives. Re-entering the set on the clearing transition fixes this and the P0 together.

**Note**

The description's table omits this gate and says "all three client-transaction entry points are gated". The one that changes semantics rather than refusing should be in that table, with the descendant caveat spelled out.

Rest of the change reads well: the hysteresis plus the warning on a discarded resume value, heartbeat + TTL so a refusal can neither be lost nor become sticky, expiring fail-open *with* a warning, starting the monitor before `loadUnminedTransactions`, suppressing only the clearing direction during a reload, and the `*errors.Error` fast path before `status.FromError`.

ordishs:
## Review — P1

**`services/blockassembly/BlockAssembler.go:406`** — the reload guard is inert after a restart.

```go
if b.txIngressFull.Load() && b.unminedTransactionsLoading.Load() {
    return true, false
}
```

The guard requires the flag to be **already set in this process**. A restarted process starts at the zero value: `NewBlockAssembler` never sets it, and `applyTxIngressCount` is the only writer. So during the reload that follows a restart the guard cannot engage.

**Failure.** Block assembly runs with `blockassembly_maxTransactionsInMemory = 1000000000` and is full. It restarts. Propagation, legacy netsync, the RPC and subtree validation each hold a cached `full=true` whose `lastHeardAt` is the last heartbeat before the process died, and `blockAssemblyFullTTL` is 60s. `loadUnminedTransactions` takes minutes on a backlog of that size. At t≈60s every ingress point expires its refusal and reopens. The monitor measures a partly-loaded assembler — below the limit — so `applyTxIngressCount` takes its `default` branch, reports `changed=false`, and publishes nothing. The node accepts unbounded new work for the rest of the reload, on top of a backlog it already cannot fit. That is the OOM the feature exists to prevent.

Confirmed at PR head with a throwaway test in package `blockassembly`: a fresh assembler with `txIngressLimit=100`, `unminedTransactionsLoading=true` and 10 transactions loaded returns `full=false` from `evaluateTxIngressFull`.

The guard is correct and live for an **in-process** reload (`Reset` → `BlockAssembler.go:939`), where the flag was already true before the subtree processor was cleared. It is dead only for the restart case that commit b70ed39 names, documents in `Start` (`BlockAssembler.go:1372-1380`), and tests.

`TestTxIngressLimitMonitorKeepsIngressRefusedDuringReload` (`tx_ingress_limit_test.go:1126`) does not catch this because it arranges the state by hand:

```go
require.True(t, ba.txIngressFull.CompareAndSwap(false, true), "arrange: this process knows it was full")
```

A restarted process never knows it was full.

**Recommendation.** Hold the refusal on the *reload*, not on a flag the new process cannot have: while a limit is configured and `unminedTransactionsLoading` is set, report full regardless of the current flag and publish it, so a restarted process re-establishes the refusal at its first evaluate tick rather than waiting for the count to climb past the limit. The cost is refusing ingress for the duration of every reload on a limit-configured node, which is the conservative direction and is bounded by the reload. Then start the test from a genuinely fresh `txIngressFull` so it fails against the current code.

---

Everything else is P2 or below. The architecture is sound — the flag on the notification bus rather than the FSM, the two watermarks, the `0` defaults. The error classification, the 503 and `ErrRPCOutOfMemory` mappings, the fail-safe parser, the consumer-side TTL and the metrics are all correct and each has a test that fails when reverted. `go build ./...` passes, and the new `blockassembly` and `blockchain` tests pass under `-tags testtxmetacache`.

services/blockassembly/BlockAssembler.go
```diff
@@ -292,6 +368,241 @@ func (b *BlockAssembler) QueueLength() int64 {
...
+	// comes up with txIngressFull at its zero value, so a guard that only suppressed clearing
+	// would be inert exactly when it is needed: every ingress point would expire its cached
+	// refusal after blockAssemblyFullTTL and reopen partway through a reload that can run for
+	// minutes, and this process would publish nothing to stop it, because it measures below the
+	// limit for most of that reload. So the refusal is established here rather than merely held.
+	//
+	// The cost is that a node with a limit configured refuses ingress for the length of its block
+	// assembly startup even when the unmined set is small. That is the conservative direction, it
+	// is bounded by the reload, and the ingress points answer it with a retryable 503.
+	if b.txIngressLimit > 0 && (b.txIngressStartupPending.Load() || b.unminedTransactionsLoading.Load()) {
```
freemans13, line 424: **The refusal established here is released by the resume watermark, not by the end of startup.**

This branch sets `txIngressFull` to true. When startup ends, `applyTxIngressCount` reads that as `wasFull` and clears only once the count falls to `txIngressResume`, which defaults to 90% of the limit. A reload that restores a count between the resume watermark and the limit therefore leaves ingress refused after startup has finished.

Confirmed at this head with a throwaway test in package `blockassembly`:

```go
ba.txIngressLimit = 1000
ba.txIngressResume = resumeWatermark(1000, 0) // 900
ba.txIngressFull.Store(true)                  // what this branch does

full, changed := ba.applyTxIngressCount(950)  // startup over, 950 reloaded
// full=true changed=false
```

The same assembler with no startup phase returns `full=false` for that count, so the refusal is inherited rather than measured.

Three claims say otherwise. The comment here says the cost "is bounded by the reload". The comment in `Start` says the same. The setting longdesc says "Ingress reopens as soon as startup finishes, if there is room", where "room" means below the resume watermark rather than below the limit, which is not how an operator will read it. The real bound is mining, so on a quiet node it is a block interval or more.

The existing test cannot catch this. `TestTxIngressEvaluateHoldsRefusalWhileUnminedTransactionsLoad` uses `limit = 5`, where `resumeWatermark` returns 4 and no integer count sits strictly between the two watermarks.

The same branch now also fires on the in-process `Reset` path, where the old rule only held a refusal that already existed. A `ResetBlockAssembly` on a node sitting at 95% of its limit now shuts ingress until it drains to 90%.

Cheapest fix: let the first evaluation after startup decide against the high watermark instead of inheriting the refusal, by clearing `txIngressFull` where `txIngressStartupPending` is cleared. The cost is at most one duplicate `full=true` notification when the reload really did restore a full backlog, since the next `applyTxIngressCount` sets it straight back.

services/subtreevalidation/subtreeHandler.go
```diff
@@ -149,6 +149,33 @@ func (u *Server) subtreesHandler(ctx context.Context, hash *chainhash.Hash, base
...
+	//
+	// Note that unlike the other three ingress points, this one does not refuse and so gets no retry:
+	// propagation, netsync and the RPC return an error and the sender resubmits, whereas a subtree
+	// transaction skipped here is not revisited when the flag clears. It reaches the mining template
+	// only when a block carrying it arrives, or on the next restart. The restart path works because
+	// the transaction is still created with UnminedSince set — that is driven by the absence of
+	// block IDs, not by this flag — so loadUnminedTransactions picks it up. The consequence to be
+	// aware of is that a transaction announced only during the full window, whose subtree is never
+	// mined by anyone, stays out of this node's template for the life of the process.
+	if u.blockchainClient != nil && u.blockchainClient.IsBlockAssemblyFull() {
```
freemans13, line 173: **icellan's P0 from 2026-08-21 is still live here, and nobody has answered it.**

The two commits pushed today expanded the comment block above and renamed the counter, but the behaviour on this path is unchanged. I checked the two structural links rather than take them on trust:

- `utxo.WithLocked(true)` is appended only inside the `addToBlockAssembly` branch (`Validator.go:2008-2010`), so a transaction created through this gate is unlocked.
- `addNode` performs no parent-presence check (`SubtreeProcessor.go:2269`); parents go into `currentTxMap` for bookkeeping only.

Both halves of the guard that normally keeps the template topologically valid are therefore absent for transactions skipped here. A parent skipped on this path stays out of the template for the life of the process. A child submitted through propagation after the flag clears spends it happily and does enter the template. The block we then mine fails `checkParentsExistOnChain` (`model/Block.go:1501`).

It is conditional rather than routine: it needs the announcing peer's subtree never to be mined by anyone. The cost when it lands is a discarded block rather than a chain split, since our own block validation rejects it too.

I am not asking for a particular fix, and the RAM argument for the gate is sound. I am asking for the decision to be recorded, either by taking one of icellan's two options or by saying here why the residual is accepted. This comment block is thorough about the fee-opportunity residual and silent about the unmineable-template one, so as it stands the next reader will conclude the second was not considered.

services/blockchain/LocalClient.go
```diff
@@ -294,6 +310,12 @@ func (c *LocalClient) GetBlockHeaderIDs(ctx context.Context, blockHash *chainhas
 }
 
 func (c *LocalClient) SendNotification(ctx context.Context, notification *blockchain_api.Notification) error {
+	// track the block assembly ingress flag locally, mirroring what the gRPC Client does when it
+	// receives this notification over its subscription
+	if notification.GetType() == model.NotificationType_BlockAssemblyFull {
+		c.blockAssemblyFull.set(blockAssemblyFullFromNotification(notification), time.Now())
```
freemans13, line 316: `set` reports whether the value changed, and the gRPC `Client` logs on that at `Client.go:223`. Here the return is dropped, so a single-process deployment gets no log line when block assembly starts or stops refusing ingress. One line closes the asymmetry.

freemans13 (changes requested):
**Request changes**, on one new defect and one earlier finding that has had no answer. The code written since my last pass is good; both objections are about where it stops.

The two commits since 2026-08-20 fix real defects. The packed atomic in `blockAssemblyFullState` closes the read-ordering race I raised, and does it by making the bad state unrepresentable rather than by reordering two stores, which is the better fix. Establishing the startup refusal instead of merely holding one closes the restart hole, and the publish retry closes a genuine gap, since a lost `full=false` had nothing to recover it.

I checked that the new tests distinguish the fix from its absence rather than just passing. Reverting the startup guard to the old `txIngressFull.Load() && unminedTransactionsLoading.Load()` rule fails six cases across `TestTxIngressEvaluateHoldsRefusalWhileUnminedTransactionsLoad`, `TestTxIngressLimitMonitorKeepsIngressRefusedDuringReload` and `TestTxIngressLimitMonitorRefusesBeforeTheReloadBegins`. That is the right shape.

## What I ran

`go build ./...` and `go vet` on the seven touched trees are clean. `golangci-lint run --new-from-rev upstream/main` reports 0 issues. The `blockassembly` and `blockchain` suites pass under `-race -tags testtxmetacache`. Upstream CI is green on this head, with no failing checks and one Aerospike fork-image job skipping.

I did not soak this on a node. The new finding below comes from a unit-level test of `applyTxIngressCount`, not from a live restart.

## Findings

**1. The startup refusal is released by the resume watermark, not by the end of startup.** New in `aae3ae303`, inline on `BlockAssembler.go`. A node whose reload restores a count between the resume watermark and the limit keeps refusing ingress after startup has finished, until mining drains it. Three places state the opposite, including the setting's own documentation. Confirmed at this head with a throwaway test.

**2. icellan's P0 from 2026-08-21 is still open and has had no reply.** Inline on `subtreeHandler.go`. The peer-subtree gate can leave a parent out of the mining template while a child that spends it goes in, and the resulting block fails `checkParentsExistOnChain`. The two new commits expanded the comments on that path and renamed its counter, but the behaviour is unchanged, so his mechanism still holds. I verified the two structural links myself rather than relaying them. I am not asking for a specific fix, only for the decision to be recorded.

**3. The PR description still says "all three client-transaction entry points are gated".** The peer-announced subtree gate is a fourth, and it is the one that changes semantics rather than refusing. It now appears correctly in the setting longdesc and in the `IsBlockAssemblyFull` interface comment, so only the description is behind. icellan asked for this on 2026-08-21.

**4. A nit on `LocalClient`,** inline.

## Where my earlier findings stand

Closed: the metric identifier, the ordering race, and the ingress enumeration, which now names the subtree path.

Two still open, both minor and neither worth blocking on. `resumeWatermark` still derives an unreachable resume of 0 for a limit of 1, since `TransactionsInMemory()` counts the coinbase placeholder and can never be 0. The five new metrics still have no entry in `docs/references/prometheusMetrics.md`.

The legacy orphan-promotion path from `handle_block.go:270` is still an ungated route into the template, and the enumeration that tells future authors to keep itself complete still does not note the exception. The RAM cost stays negligible, bounded by `legacy_maxOrphanTxs`, so this is a documentation point only. Retitling error code -7 from Bitcoin Core to SVNode answers the attribution part of my fifth finding; the choice of -7 itself I read as a decision taken.

icellan (changes requested):
Re-review at max effort against current code. The design is sound in outline and the documentation is unusually good, but two things block merge: the peer-subtree gate can build a template that peers persist as an *invalid* block, and the OOM this PR exists to fix is still reachable on restart.

The three *refusing* gates are safe on the parent/child question — a refused tx is rejected before `txStore.Set` and before the validator, so no UTXO is created and a later child fails its input lookup. The fourth gate is a different mechanism and that is where the problem is.

Verified locally: `go build`, `go vet`, `golangci-lint`, `gci diff`, `gofmt` clean; both `.pb.go` regenerate byte-identical; the new tests pass, including under `-race`. None of the below is caught by CI.

## Blocking

**1. Peer-subtree gate produces an unmineable template — `services/subtreevalidation/subtreeHandler.go:173`**

The gate creates a transaction's UTXOs while withholding it from the template. Flag true, peer subtree carries parent `P`: `WithAddTXToBlockAssembly(false)` skips only `sendToBlockAssembler`, while `spendAndCreateInUtxoStore` still creates `P`'s UTXOs with `UnminedSince` set. Flag clears. Child `C` spending `P:0` arrives via any ingress point, `P`'s UTXO exists, so `C` validates and **is** templated — `addNode` (`SubtreeProcessor.go:2115`) does no parent-presence check, it only dedupes via `currentTxMap`. We win a block: the template holds `C`, `P` is neither in the block nor on chain, and `P` can never be mined by this node because it is not in the template.

Peers then run `validOrderAndBlessed` → `checkParentsExistOnChain` → `getParentTxMetaBlockIDs` (`model/Block.go:1534`, `len(BlockIDs)==0` → `BlockIncompleteError`) → `BlockValidation.go:1938-1944` with `isCaughtUp` true → `storeInvalidBlock` + "block contains a floater". The block is persisted invalid by every caught-up peer while this node keeps building on its own tip.

All four pre-existing `WithAddTXToBlockAssembly(false)` sites are gated on `FSMStateCATCHINGBLOCKS`, where the skipped txs arrive inside blocks that go on chain — that is why they are safe and this one is not. Note also that `check_block_subtrees.go:494-498` already spends code capturing the FSM state once per block so two pipelines cannot "give divergent `WithAddTXToBlockAssembly` settings for the same block"; this gate is re-read per subtree, so subtrees of the same block can diverge.

Either refuse the subtree outright while full — the block carrying it re-presents those txs through the ungated `CheckBlockSubtrees` path — or reconcile on the clearing transition. Both preserve the invariant "if a tx's UTXOs exist unmined, it is in the template".

**2. The conditional `WithLocked(true)` removes the only enforcement of that invariant — `services/validator/Validator.go:1554`**

Topological template ordering is enforced by the locked-flag two-phase commit, not by any check in block assembly. Locked outputs are unspendable (`stores/utxo/Interface.go:297`, enforced with `ErrTxLocked` in `aerospike/spend.go:1026` and `sql/sql.go:2243`); `twoPhaseCommitTransaction` (`Validator.go:1039`) unlocks a tx only *after* block assembly accepted it; `TxLockedMaxRetries` (`Validator.go:478`) exists to absorb a child arriving just before its parent's unlock.

`WithLocked(true)` sits behind `else if addToBlockAssembly`, so with the gate engaged `P` is created **unlocked** — not spend-only, `SkipUtxoCreation` is a separate option, so outputs and metadata are fully created — and never templated. The guard is gone in both directions: nothing blocks the child, and the `ErrTxLocked` retry cannot help because there is no lock to observe. This is what makes 1 reachable.

**3. `submitMiningSolution` validates its own block blind to parent relationships — `services/blockassembly/Server.go:1669`**

`block.Valid(..., nil txMetaStore, ...)`. `model/Block.go:763` gates step 12 on `txMetaStore != nil`, so `checkParentTransactions` / `checkParentsExistOnChain` never run for a locally mined block: `Valid` returns true, `AddBlock` stores it as the local tip, `SetBlockSubtreesSet` runs, and the invalid block is announced. The line immediately above carries the pre-existing `// TODO add more checks to the Valid function, like whether the parent/child relationships are OK` — this PR turns that theoretical gap into a reachable one. The failure branch already does the right thing (`Reset(false)` + delete the job), so passing `ba.utxoStore` plus the header/ID context would both detect and self-heal.

## Before this can be relied on as an OOM fix

**4. The unmined reload bypasses the cap — `BlockAssembler.go:3069`.** A node OOMs holding N txs; all N keep `UnminedSince`. On restart the reload iterates every one into a slice preallocated for 16M and feeds them through `AddDirectly` / `AddNodesDirectly`, which bypass `AddTx` and all four gates. Grepping 3000-3400 for `txIngressLimit` / `IsTxIngressFull` / `MaxTransactionsInMemory` returns nothing. The node OOMs again at the same point. The Scope section credits `unminedTxDiskSortEnabled` and `unminedLoadingBatchSize`, but those control sort strategy and batch size, not the total loaded. Enforcing the cap inside `SubtreeProcessor` — the one component that owns the count and that both `AddTxBatch` and `AddDirectly` pass through — covers this and any future path.

**5. Nothing seeds the flag, so the loading guard cannot fire on the restart path it was written for — `BlockAssembler.go:405`.** The guard needs `txIngressFull` already true; the only writers are `applyTxIngressCount` at `:428`/`:443`, with no persistence, restore, or seeding. A process killed while full starts false, so every evaluate tick hits `default: return wasFull, false` while the reload runs for minutes — and each ingress point's cached `full=true`, timestamped by the *dead* process's last heartbeat, expires at T+60s. That is exactly what the `Start()` comment at `:1320-1327` claims to prevent. Both tests that pin this manufacture the state with `txIngressFull.CompareAndSwap(false, true)` (`tx_ingress_limit_test.go:456`, `:576`), a state no restarted process reaches, so both are green against the broken behaviour.

**6. The reload keeps the count flat, so the count cannot re-establish the flag either — `BlockAssembler.go:3296`.** The iterator appends every record, then `stableSortUnminedByCreatedAt` sorts the whole slice, and only then does the add loop run — and `txCount` is incremented only in that final phase. With 5 this removes the last fallback: the guard cannot fire because the flag is false, and `applyTxIngressCount` cannot set it because the count has not moved. Moving the monitor earlier in `Start()` does not help when the monitor has nothing to observe.

## Also P1

**7. Fifth ingress point, ungated — `services/validator/Server.go:388`.** `ValidateTransaction`/`ValidateTransactionBatch` and HTTP `/tx`, `/txs` reach block assembly with no gate, and `settings.conf:1467`/`:1473` set both listen addresses, so both surfaces are live in the shipped config. Grep for `IsBlockAssemblyFull` under `services/validator` returns zero hits, and none of the four new rejection metrics counts this traffic. `Interface.go`'s new contract enumerates four paths and says "If you add another path that reaches block assembly, gate it here too" — this path is in neither the code nor the table, and the validator Server already holds a `blockchain.ClientI` (`Server.go:108`).

**8. The netsync gate is an unrecoverable silent drop — `legacy/netsync/manager.go:1234`.** The tx is dropped before `Validate` with no reject message, so there is no `UnminedSince` record for the reload to recover. The comment says clearing `requestedTxns` "lets us fetch it again from a later inv message", but nothing produces one: legacy peers inv a mempool tx once per connection, `sm.requestedTxns` is a 10s expiring map rather than a retry trigger, the node never sends `MsgMemPool`, `processInvMsg:2430` stops requesting while full, and `peer.AddKnownInventory(iv)` has already run. `Interface.go` asserts the opposite ("The first three refuse, so the sender retries"). Two of the four gated paths silently lose transactions.

**9. Genuine store faults now return 503 instead of 500 — `propagation/Server.go:731`.** `stores/blob/file/file.go:383` returns `NewServiceUnavailableError` on a write-permit timeout; `processTransactionInternal:1348` wraps it as `NewStorageError`. `(*Error).Is` walks the chain, so the new arm matches and `POST /tx` answers "back off and resubmit" for a node whose tx store is broken. In `handleMultipleTx` it takes the new 503 arm instead of the `>= 500` arm, so a disk stall no longer forces the batch to 500, and a batch mixing one store fault with 99 permanently-invalid txs reads as wholly retryable. `isTransientBackpressure:1395` explicitly refuses `errors.Is` for this exact reason and logs the same error at Error level — one request, two contradictory verdicts. `Server_test.go`'s `hardFaultErr` has no wrapped cause, so the case is untested. Live at the default config. One predicate in `errors/error_utils.go` would settle it; if the two decisions genuinely need different sets, both comments should say so rather than one denying the other.

**10. The gRPC refusal maps to `codes.Internal`, which the shared retry gate does not retry — `propagation/Server.go:1339`.** `ErrorCodeToGRPCCode` (`errors/errors.go:1023`) has no `ERR_SERVICE_UNAVAILABLE` arm; `util/grpc_helper.go:370` retries only `Unavailable`/`DeadlineExceeded`. So a gRPC submitter drops the tx and sees a status reading "the node is broken", while the pre-existing admission-control refusal 226 lines above in the same handler uses `codes.Unavailable` and *is* retried. UDP6 is fire-and-forget with no ack, so a refused multicast tx is simply lost — and because `:563`'s `Errorf` became `logTxProcessingError`, the batch handler's `codes.Unavailable` rejection is now `Debugf`, so wholesale UDP shedding is silent at INFO.

**11. The TTL expiry CAS clobbers a concurrent heartbeat — `blockchain/block_assembly_full.go:114`.** `full` and `lastHeardAt` are independent atomics. A caller loads `full=true` and a stale `lastHeardAt`; the heartbeat then lands and runs `set(true, now)` (`changed=false`, nothing logged); the caller's `CompareAndSwap(true, false)` still succeeds because it is not conditioned on the timestamp, wiping a refusal that is 0ms old. Ingress accepts for ~10s and the `Warnf` claims "no notification for 1m0s" when one just arrived. TOCTOU on correctly-used atomics, so `-race` never sees it — reproduced deterministically. CAS the timestamp too, or hold both in one `atomic.Value`.

**12. No replay or refetch of the flag on (re)connect — `blockchain/Client.go:1862`.** `SubscribeToServer` authoritatively refetches the FSM state and the server replays the chain tip, but nothing replays the last `BlockAssemblyFull`. A propagation pod that restarts, or a subscriber the server evicts and that reconnects, starts at the zero value and admits everything for 0-10s into an assembler already at its limit, on top of the documented Kafka-lag overshoot. Rolling-restarting the ingress services during a full window reopens each in turn. This PR added `StateMessage.TxIngressFull` for exactly this question and nothing reads it.

## Lower

13. `BlockAssembler.go:405` — other direction of the same guard: `reset()`'s `postProcessFn` (`:888`) also runs the reload, so every reorg-triggered reset on a large unmined set re-pins node-wide refusal for its duration, and the 10s heartbeat keeps refreshing every client's TTL so the expiry cannot rescue it.
14. `BlockAssembler.go:463` + `blockchain/Server.go:78` — `publishTxIngressFull` has no per-call deadline and bottoms out in a blocking send on a 100-deep channel, so a stalled fan-out parks the monitor with nothing logged; separately the 10s heartbeat halves the 64-slot subscriber budget (~9 min → ~4.5 min) for all subscribers precisely when consumers are slowest. Both degradations fail open after 60s.
15. `BlockAssembler.go:463` — the monitor calling `SendNotification` on `&blockchain.Mock{}` panics unrecoverably on the goroutine and kills the whole package's test binary; masked only by the setting defaulting to 0. The same PR's `Mock.IsBlockAssemblyFull` avoids `m.Called` for this hazard; `SendNotification` did not get the same treatment.
16. `tx_ingress_limit_test.go` — no test wires the setting through `NewSettings`/`NewBlockAssembler` (zero `*_test.go` hits repo-wide; every test overwrites `ba.txIngressLimit` by hand), so a typo'd key or wrong field leaves the feature dead in every real deployment with the suite green. The subtree-gate test asserts only the spy's recorded option, never a template.
17. ~815ns and 8 allocs per refusal (uncached `TxIDChainHash`, `errors.New` stack capture), the UDP path always taking the `UnwrapGRPC` protobuf branch, ~110KB of `/txs` error messages, and `time.Now()` per transaction even with the limit disabled.
18. `BlockAssembler.go:507` — shutdown zeroes the gauge but leaves the flag set and publishes no clearing notification, so the metric, the gRPC `StateMessage` and the clients all disagree.
19. `BlockAssembler.go:340` — `resumeWatermark`'s edge branch is reachable for limits 1-9, not just 1 (the comment is wrong), and `limit=1` yields `resume=0`, which the count can never reach because the coinbase placeholder is always counted → permanent ingress outage.
20. `BlockAssembler.go:499` — the monitor registers on `b.wg` but `BlockAssembler.Wait()` has no production caller, so shutdown never joins it and the `Add(1)` advertises a guarantee that does not exist.

## Worth considering instead of more per-ingress gates

Three alternatives came up repeatedly while working through the above: gate once at the `AddTx*` choke point in `SubtreeProcessor` rather than at each ingress point (covers 4 and 7 by construction, and any future path); `PauseAll()` the validator's Kafka consumer to bound the overshoot the PR currently documents as unbounded; and a GOMEMLIMIT-fraction watermark instead of a raw transaction count, given the sizing section's own caveat that per-tx cost swings from a few hundred bytes to tens of KB.

services/blockassembly/BlockAssembler.go
freemans13, line 424: **This guard now pins ingress closed for the whole of every reorg-driven reset too, and nothing says so.**

`unminedTransactionsLoading` is set by `loadUnminedTransactions`, which runs from two places: `Start`, and the post-process step of `reset` (`BlockAssembler.go:992`). This guard keys on that flag alone, so a reset forces `txIngressFull` true node-wide for the length of the reload, whatever the assembler is actually holding.

Before `aae3ae303` that could not happen with the flag clear, because the old condition also required `txIngressFull.Load()`. So closing the restart hole extended the refusal onto a path it did not previously cover. I do not think that was intended, and it may well be the right behaviour anyway: the argument that justifies it at startup applies unchanged, in that what the assembler holds mid-reload does not describe what it is about to hold.

What is wrong either way is the documentation. Three places frame this cost as a startup-only one and none of them mentions a reset. This comment says the node "refuses ingress for the length of its block assembly startup". `Start`'s comment says the flag "is cleared once the reload returns". The setting's long description has a `### Startup` heading with no equivalent for reorgs (`blockassembly_settings.go:62`).

The size of it depends on the node. `reset` runs on a large reorg, on a reorg containing an invalid block, and from the `Reset` and `ResetWithInputValidation` RPCs (`BlockAssembler.go:2214`, `:2260`, `:1924`). On a node whose unmined set takes minutes to reload, that is minutes of every ingress point answering 503 during a reorg, which is when you would least choose to turn transactions away.

If you keep it, the fix is three comment edits. If you do not, the guard needs to distinguish the startup reload from a reset reload, which `txIngressStartupPending` already gives you.

freemans13 (changes requested):
**Request changes.** The most useful thing this round is not a new defect, it is a correction that should unblock some work: three of the twenty items in the 2026-09-10 review describe code this branch no longer contains, and can be closed without touching anything.

That review is recorded against `2d6d9e81e` because that is the head GitHub saw when it was submitted. Its content matches `2f4023e76`, the merge commit two commits earlier. The two fixes pushed on 2026-09-09 are what closed them, and in each case they do what the finding asks for.

**Finding 5, the flag is never seeded.** It quotes `if b.txIngressFull.Load() && b.unminedTransactionsLoading.Load()`. That is line 406 of `2f4023e76` and it is absent at this head. Commit `aae3ae303` replaced it with a guard that never reads the flag, so it establishes a refusal rather than only holding one. A process killed while full now refuses from its first evaluate tick, which is the hole the finding describes (`BlockAssembler.go:424`).

**Finding 11, the expiry swap clobbers a heartbeat.** It describes the cached flag and its timestamp as two independent atomics, and asks for them to be held as one value. They were two at `2f4023e76`, where the clearing compare-and-swap was `s.full.CompareAndSwap(true, false)` on line 114, the exact line the finding cites. Commit `2d6d9e81e` packed them into one `atomic.Int64`, where zero means not refusing and any other value is the nanosecond instant of the last announcement. The clearing swap now tests the exact reading the caller judged, so a heartbeat landing in that window makes it fail (`block_assembly_full.go:128`).

**Finding 8's evidence.** The comment it quotes, claiming the cleared bookkeeping "lets us fetch it again from a later inv message", was replaced in `2d6d9e81e` by six lines saying the opposite, in the same terms the finding asks for (`legacy/netsync/manager.go:1236`).

Finding 6 is a different case and should not be closed. Its mechanism is right, and it corrects something I got wrong on 2026-08-20. I claimed the in-memory transaction count climbs during the unmined reload. It does not. The reload appends every record to a slice, merges those slices across workers, sorts the result, and only then starts adding, so the count stays flat through the phase that actually allocates the memory (`BlockAssembler.go:3329-3421`). What no longer holds is the conclusion drawn from it, that this removes the last fallback, because after `aae3ae303` the flag does not depend on the count during a reload at all.

## What still blocks, verified at this head rather than relayed

**The peer-subtree gate can build a template that peers store as an invalid block.** This is the 2026-08-21 P0 and it is unchanged since; `2d6d9e81e` touched only the metric's Go identifier on that line. The gate keeps a transaction out of the mining template while still creating its unspent outputs, so a child spending that parent validates and is templated, and the parent is in neither the block nor the chain (`subtreeHandler.go:174`). The half of this that makes it reachable is a conditional that predates the branch: `WithLocked(true)` sits behind `else if addToBlockAssembly`, so with the gate engaged the parent is created unlocked, and the locked-output retry that normally absorbs a child arriving before its parent has nothing to observe (`services/validator/Validator.go:2008`). That file is not branch-authored, so this is a path the branch newly reaches rather than code it wrote, but the reachability is new.

**The unmined reload is not capped, so the restart this PR exists to survive still runs out of memory.** A node that dies holding N transactions leaves all N with `UnminedSince` set, and the reload pulls every one of them back in. Refusing ingress does not help, because the allocation is the reload's own. I checked the disk-sort path as well, since the PR description credits `unminedTxDiskSortEnabled`: it writes to a temporary store to avoid holding the unsorted slice, then reads every entry back and calls `AddDirectly` per transaction, so the resident set at the end is identical (`BlockAssembler.go:3909`). Grepping the whole `services/blockassembly` tree for `txIngressLimit`, `IsTxIngressFull` and `MaxTransactionsInMemory` returns nothing inside either reload or inside `SubtreeProcessor`.

**The validator's own listeners are a fifth ingress point with no gate.** `grep -rn IsBlockAssemblyFull services/validator` returns zero hits, and both listen addresses are set in the shipped defaults, so the gRPC and HTTP surfaces are live (`settings.conf:1629`, `:1635`). The interface comment tells future authors to gate any new path that reaches block assembly, and the setting's own table lists four gated paths, so both should name this one or gate it.

**The startup refusal is released by the resume watermark, not by the end of startup.** Still open from my 2026-09-09 review, still reproducible. A node whose reload restores a count between the resume watermark and the limit keeps refusing after startup finishes, until mining drains it. Three places state the opposite, including the setting's own documentation.

## New this round

One new finding, inline on the startup guard: that guard now also pins ingress closed for the whole of every reorg-driven reset, and nothing documents it.

## Corrections to my own earlier findings

On `resumeWatermark`, finding 19 is right and my August version was too narrow. I said the derived resume value is unreachable only at a limit of 1. The `if resume >= limit` branch is taken for every limit from 1 to 9, because integer division makes `limit/10` zero there, so `resume` equals `limit`. The comment claiming the branch is "only reachable for a limit of 1" is wrong for 2 through 9 (`BlockAssembler.go:346`).

`go build ./...` is clean. `go test -race -tags testtxmetacache -count=1` over the seven touched trees passes: 3129 tests across 15 packages. Upstream CI is green on this head, 42 checks passing with the Aerospike fork-image job skipping.

I did not soak this on a node, so nothing here is evidence about behaviour under real memory pressure. The reload and reset findings are read from the code, not observed on a restart.

One practical note: the branch no longer merges. `settings/settings.go` conflicts with `main`, and everything else merges cleanly.

freemans13:
@icellan, a note on the 2026-09-10 review, since a review body cannot be replied to in a thread.

It is recorded against `2d6d9e81e`, but its content matches `2f4023e76`, two commits earlier. Findings 5, 8 and 11 each quote code that `aae3ae303` and `2d6d9e81e` had already replaced the day before, and in all three cases with the change the finding asks for. Detail and the line evidence are in my review just now.

Finding 6 is the opposite case. Its mechanism holds and it corrects a claim I made in August about the reload's transaction count, so that one is live even though the conclusion drawn from it is not.

PR #1596, opened 2026-08-15
fix(agents): repair four unloadable agents, stop hand-listing the roster

ordishs:
## What

Two changes to the agent setup, in one place because the second explains the first.

**1. Repairs four agent definitions that never loaded.**

`api-tester`, `backend-architect`, `document-reviewer`, and `test-writer-fixer` each ended the `description` field with the word `Examples:` and then ran on across blank lines into example blocks. That bare colon opens a YAML mapping mid-scalar, so the parser stopped there and discarded the rest of the frontmatter — the `tools` and `color` fields with it. The agents never registered and could not be selected.

Each description is now a quoted single line that carries the trigger text the examples used to convey. Only the description is read when deciding whether to delegate, so nothing of value moves.

**2. Replaces the hand-written roster in `AGENTS.md` with a pointer to the directory.**

The roster is what exposed the breakage: it advertised fourteen agents when only ten worked. Every agent file already carries its own description and the directory is read automatically, so the list duplicated information that gets loaded regardless — while needing manual upkeep to stay correct. Upkeep it did not get. A pointer cannot drift.

## Evidence

Frontmatter parsed with a real YAML parser, before and after:

```
before                                  after
api-tester             FAILS            api-tester             PARSES
backend-architect      FAILS            backend-architect      PARSES
document-reviewer      FAILS            document-reviewer      PARSES
test-writer-fixer      FAILS            test-writer-fixer      PARSES
(10 others PARSE)                       (10 others PARSE)

parses: 10   fails: 4                   parses: 14   fails: 0
```

The failure in all four cases: `mapping values are not allowed in this context at line 2`.

All four agents were confirmed to register and become selectable after the change.

## Scope

Documentation and agent configuration only. No production code, no behaviour change to the node.

Every agent system prompt body is unchanged byte for byte — only frontmatter was rewritten. The removed example blocks remain in history.


liam (commented):
I verified the central claim independently rather than taking the description at face value. It holds up.

## Verdict: approve

### The diagnosis is correct

Parsing every agent file's frontmatter at `main` with a real YAML parser:

```
FAIL  api-tester.md         mapping values are not allowed here
FAIL  backend-architect.md  mapping values are not allowed here
FAIL  document-reviewer.md  mapping values are not allowed here
FAIL  test-writer-fixer.md  mapping values are not allowed here
PARSE (10 others)
```

The mechanism is exactly as stated: `description: ... Examples:` ends the line on a bare colon, which YAML reads as a key indicator, so the scalar terminates and the parse dies there. Everything after it — `tools`, `color` — is discarded with the block.

There is a stronger piece of evidence than the parser, though: **the agent roster injected into a live Claude Code session in this repo lists ten agents, and it is precisely the ten that parse.** `api-tester`, `backend-architect`, `document-reviewer` and `test-writer-fixer` are absent. The breakage is live, not theoretical, and `AGENTS.md` has been advertising four agents that could not be selected.

### The fix is faithful

- All 14 files parse on this branch.
- `tools` and `color` are recovered for the three files that had them.
- `test-writer-fixer` gets no `tools` field — correct, because it never had one on `main` (its frontmatter there was `name`, `description`, `color` only). Nothing was silently dropped, and it inherits all tools the way `bitcoin-expert` does.
- I hashed everything after the closing `---` in each of the four files, before and after: **byte-identical**. The "system prompts unchanged" claim is true, not approximate.
- No dangling references. Only `AGENTS.md` named the four agents. `.github/workflows/claude-code-review.yml:191` cites `.claude/agents/document-reviewer.md` by *path* for its methodology text — that body is unchanged, so the workflow is unaffected.

### Three notes, none blocking

**1. The `AGENTS.md` wording argues with itself.** "Claude Code loads them automatically — read the directory for the current roster" tells an agent to `ls` a directory whose contents are already in its context. Minor waste. Something like "each agent's description is loaded into context automatically; the directory is the source of truth if you need the full definition" would be tighter.

**2. Removing the roster costs something real.** `AGENTS.md` opens by saying it is read by all AI tools, and tools that do not auto-discover `.claude/agents/` lose the at-a-glance domain mapping ("bitcoin-expert — consensus, crypto"). I still think the trade is right: a list that spent an unknown stretch of time advertising four dead agents is worse than no list. Worth naming the trade rather than treating it as a pure win.

**3. Nothing stops this regressing.** The next vendored agent with an `Examples:` description fails the same silent way — no error, no warning, the agent just is not there. This PR is a one-off repair of a class of bug. A few lines in CI would make it durable:

```yaml
- name: Validate agent frontmatter
  run: |
    for f in .claude/agents/*.md; do
      sed -n '2,/^---$/p' "$f" | sed '$d' | yq -e '.name and .description' > /dev/null \
        || { echo "::error file=$f::frontmatter does not parse"; exit 1; }
    done
```

I would suggest that as a follow-up rather than expanding this PR — the current diff is well scoped and the commit split (roster pointer, then the repair that justifies it) reads cleanly.

liam (approved):

oskarszoon (approved):
Verified the parse failure independently: on `main`, `api-tester`, `backend-architect`, `document-reviewer` and `test-writer-fixer` all fail YAML frontmatter parsing with "mapping values are not allowed in this context" — the `user:`/`assistant:` lines inside the multi-line `description:` scalar. All four parse at head, the other ten agents were already fine, and the agent bodies are unchanged.

`.github/workflows/claude-code-review.yml` references `document-reviewer.md` by path, so it still resolves.

Issue #1372, opened 2026-07-29
blockvalidation: mid-cycle peer failures in failedPeers are dropped when the catchup cycle succeeds


ordishs:
Follow-up from the review of PR #1371 (part of #1368).

`releaseCatchupLock` copies and drains `CatchupContext.failedPeers` only inside the `*err != nil` branch. When the primary serves a poisoned `subtree_data` body but an alternative peer rescues the block, the cycle ends with `err == nil` and the poisoned peer is never charged — it keeps full reputation, gets selected again next cycle, and every subsequent cycle pays the detect-then-bypass-retry latency against it.

Either:
- drain and charge `failedPeers` on the success path too (a clean cycle can still carry genuine per-peer failures), or
- document the drop as deliberate in the `failedPeers` field comment, which currently states unconditionally that the drain "charges each of them once."

The first option better serves the attribution goal of PR #1371: the poisoned-peer scenario is exactly the one where the cycle usually succeeds via an alternative.

PR #1597, opened 2026-08-17
MvP-1374 docs(blockvalidation): state the attribution invariants on recordCatchupPeerFailure

### LINKED ISSUE #1374: blockvalidation: document cross-cycle attribution edge in recordCatchupPeerFailure
Follow-up from the review of PR #1371 (part of #1368). Documentation-only.

`fetchAndStoreSubtreeData` intentionally survives cancellation via `context.WithoutCancel`, but `recordCatchupPeerFailure` resolves `u.activeCatchupCtx` at call time. A fetch goroutine that outlives its catchup cycle therefore either no-ops (context already cleared) or records its failure into the *next* cycle's `failedPeers` map. The blamed peer is still the peer that actually failed, so this is cycle-attribution noise rather than a correctness problem — but the doc comment on `recordCatchupPeerFailure` should state it so a future reader doesn't rediscover it as a bug.

ctnguyen:
Resolves https://github.com/bsv-blockchain/teranode/issues/1374

Follow-up from the review of #1371. Comment-only: one doc comment in
`services/blockvalidation/catchup.go`, no behaviour change.

## Why

`recordCatchupPeerFailure` resolves the failing peer's catchup cycle from the server-wide
`u.activeCatchupCtx` rather than from its caller. Two properties define the limits of that read and
neither was written down, so the next reader has to re-derive both.

The old comment also said that a call outside catchup is a no-op. That holds only when no catchup is
running, which is not the same claim.

## Correction to the issue's premise

The issue asks the comment to say a detached fetch can outlive its cycle and be charged to the next
one. **That is not reachable on the current tree**, so the comment does not say it.

A fetch detached with `context.WithoutCancel` still runs inside the errgroup that
`fetchSubtreeDataForBlock` waits on (`get_blocks.go:430`), which `blockWorker` waits on
(`get_blocks.go:133`), which `catchup` waits on (`catchup.go:1041`) before the deferred
`releaseCatchupLock` clears `activeCatchupCtx` (`catchup.go:473`). A slow peer can delay the end of
its cycle; it cannot survive it. Writing the issue's text verbatim would have documented a
non-existent bug.

What *is* reachable through the same global read is `RevalidateBlock` (`Server.go:1254`): it runs on
its own gRPC goroutine with no interlock against a running catchup, so its per-subtree failures are
booked against whatever cycle happens to be live. The peer blamed is the one that actually failed, so
the charge is right by peer and wrong only by cycle.

The comment therefore records the join as an invariant to preserve (breaking it by making any fetch
on that path fire-and-forget would require threading the cycle from the caller) and the
`RevalidateBlock` overlap as the live caveat.

## Scope

One hunk in `services/blockvalidation/catchup.go`, comment lines only. The `failedPeers` field
comment is untouched — #1372 owns it.

## Verification

`gofmt`, `go build`, `go vet`, `make lint` (0 issues) clean. `services/blockvalidation` green with
and without `-race`, 0 data races. Inertness is proven by the diff shape: one file, every changed
line a comment.

## Follow-up

The `RevalidateBlock` attribution non-determinism deserves its own issue — the fix is to thread a
nil-able `*CatchupContext` from `fetchSubtreeDataForBlock` down to `recordCatchupPeerFailure` and
decide whether non-catchup failures need their own reporting sink. Not bundled here.

### DIFF
--- a/services/blockvalidation/catchup.go
+++ b/services/blockvalidation/catchup.go
@@ -611,17 +611,26 @@ func (u *Server) releaseCatchupLock(ctx *CatchupContext, err *error) {
 // its context (context.WithoutCancel) so an aborted fetch still finishes writing,
 // but the goroutine stays inside the errgroup fetchSubtreeDataForBlock waits on,
 // which blockWorker waits on, which catchup waits on before releaseCatchupLock
-// clears the context. A slow peer can therefore delay the end of its cycle — up to
-// the subtree_data fetch timeout — but can never outlive it, so a failure raised by
-// a catchup's own fetch is never charged to a later cycle nor dropped into a cleared
+// clears the context. A slow peer can therefore delay the end of its cycle but can
+// never outlive it. That delay is not a single subtree_data fetch timeout: the
+// deadline is installed per call with no shared budget, and one subtree makes up to
+// two calls per peer (a cache-bypass retry) across each alternative peer, so the
+// drain ceiling is a multiple of that timeout — see fetchAndStoreSubtreeAndSubtreeData.
+// The bound is larger in magnitude but still finite, so a failure raised by a
+// catchup's own fetch is never charged to a later cycle nor dropped into a cleared
 // context. Making any fetch on that path fire-and-forget breaks this and requires the
 // cycle to be threaded from the caller instead.
 //
 // Not every caller is a catchup. RevalidateBlock reaches this path through
 // fetchSubtreeDataForBlock on its own gRPC goroutine with no interlock against a
-// running catchup, so its per-subtree failures land in that cycle's failedPeers and
-// are charged when the cycle drains. The peer blamed is the one that actually failed
-// to serve the data, so the charge is right by peer and wrong only by cycle.
+// running catchup, so its per-subtree failures land in whatever catchup cycle is
+// active. releaseCatchupLock drains failedPeers only inside its *err != nil branch,
+// so those failures are charged only if that concurrent cycle itself ends in error;
+// a cycle that succeeds discards the map untouched and the RevalidateBlock failure is
+// then charged to no one. When a failure is charged, it lands on the peer that
+// returned it for that fetch, not on some other peer; that does not prove the peer
+// deserves a reputational charge, because a peer that 404s subtree data the pruner
+// removed did nothing wrong.
 func (u *Server) recordCatchupPeerFailure(peerID string, err error) {
 	if peerID == "" || err == nil || errors.IsLocalError(err) {
 		return
--- a/services/blockvalidation/catchup.go
+++ b/services/blockvalidation/catchup.go
@@ -595,13 +595,33 @@ func (u *Server) releaseCatchupLock(ctx *CatchupContext, err *error) {
 }
 
 // recordCatchupPeerFailure attributes a data-serving failure to the peer that caused
-// it, for the current catchup cycle. Best-effort: with no active catchup context
-// (e.g. a direct block fetch outside catchup) the call is a no-op.
+// it, charging it against whichever catchup cycle is active when the call is made.
+// Best-effort: with no cycle active the call is a no-op.
 //
 // errors.IsLocalError is checked here — not at each call site — so every caller
 // gets the guard for free: a context cancellation (catchup abort / peer switch /
 // shutdown landing mid-retry) or a local storage failure (subtreeStore.Set) is ours,
 // not the peer's, and must never land an innocent peer in failedPeers.
+//
+// Attribution reads the server-wide u.activeCatchupCtx rather than taking a cycle
+// from the caller. Two properties define the limits of that read, and both are worth
+// re-checking before changing anything on this path.
+//
+// A catchup cycle outlives every fetch it starts. fetchAndStoreSubtreeData detaches
+// its context (context.WithoutCancel) so an aborted fetch still finishes writing,
+// but the goroutine stays inside the errgroup fetchSubtreeDataForBlock waits on,
+// which blockWorker waits on, which catchup waits on before releaseCatchupLock
+// clears the context. A slow peer can therefore delay the end of its cycle — up to
+// the subtree_data fetch timeout — but can never outlive it, so a failure raised by
+// a catchup's own fetch is never charged to a later cycle nor dropped into a cleared
+// context. Making any fetch on that path fire-and-forget breaks this and requires the
+// cycle to be threaded from the caller instead.
+//
+// Not every caller is a catchup. RevalidateBlock reaches this path through
+// fetchSubtreeDataForBlock on its own gRPC goroutine with no interlock against a
+// running catchup, so its per-subtree failures land in that cycle's failedPeers and
+// are charged when the cycle drains. The peer blamed is the one that actually failed
+// to serve the data, so the charge is right by peer and wrong only by cycle.
 func (u *Server) recordCatchupPeerFailure(peerID string, err error) {
 	if peerID == "" || err == nil || errors.IsLocalError(err) {
 		return
--- a/services/blockvalidation/catchup.go
+++ b/services/blockvalidation/catchup.go
@@ -619,7 +619,11 @@ func (u *Server) releaseCatchupLock(ctx *CatchupContext, err *error) {
 // The bound is larger in magnitude but still finite, so a failure raised by a
 // catchup's own fetch is never charged to a later cycle nor dropped into a cleared
 // context. Making any fetch on that path fire-and-forget breaks this and requires the
-// cycle to be threaded from the caller instead.
+// cycle to be threaded from the caller instead — as would an injected
+// fetchSubtreeDataForBlockFn that does not preserve that join.
+// In adaptive-fetch optimistic mode the per-subtree fetch is skipped entirely, so on
+// that branch there is nothing to join and the guarantee holds vacuously rather than
+// by the join above.
 //
 // Not every caller is a catchup. RevalidateBlock reaches this path through
 // fetchSubtreeDataForBlock on its own gRPC goroutine with no interlock against a


ordishs (approved):
Approving. The old comment was wrong and this replaces it with something genuinely more useful, at zero behaviour risk — every changed line is a comment.

The correction to the issue's premise is right, and I checked it rather than taking it on trust. The join chain holds link by link: the detach at `get_blocks.go:593` stays inside the errgroup joined at `get_blocks.go:430`, reached from `blockWorker` (`get_blocks.go:237-243`), joined again at `get_blocks.go:133`, then `catchup.go:1047-1061`, before the deferred `releaseCatchupLock` (`catchup.go:181`) clears the context at `catchup.go:473`. I looked for an escape hatch — a bare `go func` on the path, a `ctx.Done()` return that abandons the group, a discarded `Wait` — and found none. Good call not writing the issue's text verbatim.

Two added clauses overstate what the code does. Neither blocks the merge; both point the safe way. Raising them only because precision is this comment's whole purpose.

**`services/blockvalidation/catchup.go:596-597`** — "are charged when the cycle drains" reads as unconditional, but the drain is gated on the cycle failing. `releaseCatchupLock` copies `failedPeers` only inside `if *err != nil && ctx != nil` (`catchup.go:384`), with the copy at `:442-449` and the branch closing at `:470`. If the concurrent catchup succeeds, `*err` is nil, the gate at `:479` sees an empty map, the loop at `:521` never runs, and the map dies with the `CatchupContext` at `:473`. So a `RevalidateBlock` subtree failure is charged only if an unrelated catchup also fails. That also means the closing "right by peer and wrong only by cycle" omits a third outcome: no charge at all. Under-charge, never a wrong-peer charge.

Suggested qualifier: "...land in that cycle's `failedPeers`, and are charged only if that cycle ends in an error — `releaseCatchupLock` drains `failedPeers` inside its `*err != nil` branch (`catchup.go:384-470`), so a cycle that succeeds discards them."

**`services/blockvalidation/catchup.go:588-589`** — "up to the subtree_data fetch timeout" gives a bound of one timeout, but the deadline is installed per call with no shared budget (`get_blocks.go:593`), and the path makes several sequential calls inside the one goroutine the cycle waits on. `tryPeerForSubtree` calls `fetchSubtreeAndDataFromPeer` at `:736` and again on a cache-bypass-retryable error at `:751` — two full windows for one peer — and `fetchAndStoreSubtreeAndSubtreeData` runs it once per alternative peer at `:805`. The ceiling for one subtree is about 2 × (1 + alternative peers) × timeout: with the 10m default (`settings.conf:459`) and three alternatives, roughly 80 minutes rather than 10.

Worth noting what does *not* multiply it, since the distinction is easy to lose: HTTP retries inside one call share the single bound, because `util/http.go:586` installs the per-attempt streaming timeout only when the context has no deadline and `:593` supplies one. The abort case is also one window, since the `IsLocalError` guard at `get_blocks.go:779` skips the failover. The multiplication needs a slow peer that is not cancelled. The surrounding "can never outlive it" is unaffected — only the magnitude is understated.

Suggested qualifier: "up to one subtree_data fetch timeout per attempt, and the path makes one attempt per peer plus a cache-bypass retry, so the drain is a multiple of that bound — see `fetchAndStoreSubtreeAndSubtreeData`."

One more, explicitly **not** for this PR: the `IsLocalError` paragraph at `catchup.go:576-578` is unchanged context here, so it is out of scope. But its promise that a cancellation "must never land an innocent peer in `failedPeers`" has a gap — `util/http.go:235-237` replaces a cancellation during the `/subtree` body read with `NETWORK_TIMEOUT`, which carries no cause and no "context canceled" text, so `IsLocalError` returns false and `get_blocks.go:744` charges the peer. I confirmed that return shape but not its end-to-end reachability, so treat it as plausible rather than demonstrated. It looks like a fit for the follow-up issue you already promised.

Scope discipline is good — leaving the `failedPeers` field comment to #1372 is right, though note it carries the same unconditional-drain elision at `catchup.go:62-64`.

freemans13:
Independent review — I re-derived the claims from the tree rather than reading them off the existing review, so where I agree with @ordishs below that is a second confirmation and not an echo.

**The central correction holds.** Declining to write issue 1374's text verbatim was right: a detached fetch cannot outlive its cycle, so the issue asked for a comment describing a bug that does not exist. I walked the join chain link by link. The detach at `get_blocks.go:593` happens inside the goroutine started at `get_blocks.go:415`, joined by the errgroup at `:430`; that call is synchronous inside `blockWorker` (`:242`), itself in an errgroup joined at `:133`, which `catchup` waits on before the deferred `releaseCatchupLock` clears `activeCatchupCtx` at `catchup.go:473`. I went looking for an escape hatch — a bare `go func`, a discarded `Wait`, a second detach — and there is none: `get_blocks.go` contains only `g.Go` calls and exactly one `context.WithoutCancel`. `gofmt`, `go build` and `go vet` are clean on the package, and every changed line is a comment, so there is no behaviour risk.

Two clauses still overstate what the code does. Both were already raised and neither is fixed at `8a200de2f`. Repeating them only because precision is this comment's whole purpose.

**`catchup.go:596-597` — "are charged when the cycle drains" reads as unconditional.** `releaseCatchupLock` copies `failedPeers` only inside `if *err != nil && ctx != nil` (`catchup.go:384`), with the copy at `:442-449`. A cycle that succeeds never copies the map and it dies with the `CatchupContext` at `:473`. So a `RevalidateBlock` subtree failure is charged only if an unrelated catchup also fails, and the closing "right by peer and wrong only by cycle" omits a third outcome: no charge at all.

**`catchup.go:588-589` — "up to the subtree_data fetch timeout" gives a bound of one window, but there is no shared budget.** The deadline is installed per call at `get_blocks.go:593`, and the path makes several sequential calls inside the one goroutine the cycle waits on: `tryPeerForSubtree` fetches at `:736` and again on a cache-bypass retry at `:751`, and `fetchAndStoreSubtreeAndSubtreeData` runs that whole thing once per alternative peer at `:805`. With the shipped 10m default (`settings.conf:459`, wired at `settings/settings.go:362`) and three alternatives, one subtree can hold the cycle for upwards of an hour. The surrounding "can never outlive it" is unaffected; only the magnitude is understated.

**One point I have not seen raised: "the charge is right by peer" is more generous than the code deserves.** `RevalidateBlock` fetches subtree data *precisely because the pruner removed it* — that is what the comment at `Server.go:1245-1246` says the call is for. A peer that 404s subtree data for an old invalid block is not misbehaving; it was never obliged to retain it. But a 404 is not a local error, so `recordCatchupPeerFailure` charges it and `releaseCatchupLock` reports it to the peer registry. "Right by peer" is true only in the thin sense that the peer genuinely did not hand over the bytes; reputationally it is a false charge against a peer that did nothing wrong. That seems worth naming rather than asserting away, and it looks like a natural fit for the follow-up issue the description already promises.

Two smaller notes for the next reader, neither a defect:

- "Making any fetch on that path fire-and-forget breaks this" points the reader at `get_blocks.go`, but the seam most likely to actually break the join is `fetchSubtreeDataForBlockFn` (`Server.go:195`, assigned at `:403`, swapped at `catchup_test_suite.go:157`). An injected implementation could break the guarantee without anyone touching the documented path.
- In adaptive-fetch optimistic mode (`get_blocks.go:233-235`) `fetchSubtreeDataForBlock` is skipped entirely, so on that branch the guarantee is vacuous rather than upheld.

Scope discipline is good — leaving the `failedPeers` field comment to #1372 is right. Note it carries the identical unconditional-drain wording at `catchup.go:60-65` ("drained by releaseCatchupLock, which charges each of them once"), so whoever picks that up should correct both in one pass.

ctnguyen:
Thanks for the feedbacks

@ordishs

- **O1 (drain is gated on cycle failure).** Fixed. The "Not every caller is a catchup"
  paragraph now states that `releaseCatchupLock` drains `failedPeers` only inside its
  `*err != nil` branch, so a `RevalidateBlock` subtree failure is charged only when the
  concurrent catchup cycle itself ends in error; a cycle that succeeds discards the map
  untouched and the failure is then charged to no one. The old "charged when the cycle
  drains" wording, which elided that third outcome, is gone. I intentionally dropped the
  "right by peer, wrong only by cycle" claim rather than re-asserting it — see F3 below on
  why. The underlying cross-cycle/no-cycle attribution non-determinism this documents (a
  `RevalidateBlock` failure landing in whichever cycle is active, or none) is a behaviour
  change rather than a doc fix and is tracked in the follow-up issue. The github-actions
  bot's inline comment on this line makes the same point and is addressed by this same
  edit, so that thread can be resolved.

- **O2 (timeout bound is a multiple, not one window).** Fixed. The comment no longer says
  the delay is bounded by a single subtree_data fetch timeout. It now states the deadline
  is installed per call with no shared budget, one subtree makes up to two calls per peer
  (a cache-bypass retry) across each alternative peer, so the drain ceiling is a multiple
  of that timeout — pointing at `fetchAndStoreSubtreeAndSubtreeData`. The "can never
  outlive it / never dropped into a cleared context" guarantee is kept; only the magnitude
  was understated.

- **O3 (`NETWORK_TIMEOUT` can erase a cancellation cause and defeat `IsLocalError`).**
  Deferred, per your "not for this PR". The return shape is confirmed but end-to-end
  reachability is not yet demonstrated, so asserting it either way in the comment would be
  premature. Folded into the follow-up issue on `RevalidateBlock` attribution gaps.

- **O4 (same elision on the `failedPeers` field comment).** Deferred. That field comment
  is owned by issue #1372; touching it here would widen scope. Whoever picks up #1372
  should apply the same "only on a failing cycle" qualifier there so both read
  consistently.

---

@freemans13

- **F1 (third outcome: no charge at all).** Shared with O1 — same edit, commit 1. No
  separate change; the paragraph now spells out the "charged to no one" outcome you called
  out.

- **F2 (10m default; one subtree can hold the cycle for ~an hour).** Shared with O2 —
  same edit, commit 1. Note on the citation: the live setting is
  `blockvalidation_subtree_data_fetch_timeout = 10m` (`settings.conf:467`, wired at
  `settings/settings.go:362`); `settings.conf:459` is inside a dead-key comment block for
  the unread `blockvalidation_subtree_fetch_timeout`. The magnitude arithmetic is
  unaffected. The committed comment refers to the timeout by name only, so it stays
  drift-proof.

- **F3 ("right by peer" is too generous — pruner-driven 404 is a false charge).**
  Accepted. Because of this I did *not* restore the "right by peer" claim; instead the
  comment now says the charge lands on the peer that returned the failure but that this
  does not prove a reputational charge is deserved, since a peer that 404s pruned subtree
  data did nothing wrong. The behavioural fix (excluding such 404s from attribution) is a
  code change, not a doc typo, so it is deferred to the follow-up issue.

- **F4 (name the injectable seam).** Fixed (commit 2). The fire-and-forget caution now
  also names an injected `fetchSubtreeDataForBlockFn` that fails to preserve the join as a
  way the guarantee can break, not only a change under `get_blocks.go`.

- **F5 (optimistic mode makes the guarantee vacuous).** Fixed (commit 2). Added a sentence
  noting that in adaptive-fetch optimistic mode the per-subtree fetch is skipped entirely,
  so on that branch there is nothing to join and the guarantee holds vacuously rather than
  by the join chain above.

- **F6 (same as O4 on the field comment).** Deferred, owned by issue #1372 — same item as
  O4, fixed there in one pass, not re-edited here.

liam (approved):
Approving. A comment-only PR is only worth as much as its claims are true, so I checked each one against the tree rather than reading for plausibility.

The correction to the issue's premise is the valuable part and it is right. The join really does exist, so the "a detached fetch outlives its cycle and is charged to the next one" scenario is not reachable, and writing the issue's text verbatim would have documented a bug that does not exist. Declining to do that, and saying so in the PR, is the right call.

## Claims verified

- `context.WithoutCancel` on the fetch path: `get_blocks.go:593`, wrapped in a per-call `WithTimeout`.
- The detached goroutine stays inside `fetchSubtreeDataForBlock`'s errgroup: created at `:377`, joined at `:430`.
- `releaseCatchupLock` clears `activeCatchupCtx` only after the drain block, and the `failedPeers` copy sits inside `if *err != nil && ctx != nil`. So the comment's claim that a cycle ending in success discards the map untouched is exact.
- `RevalidateBlock` calls `fetchSubtreeDataForBlock` directly on its own gRPC goroutine: `Server.go:1254`. Note it calls the method, not the `fetchSubtreeDataForBlockFn` seam, which makes the caveat sharper than the comment claims.
- The `fetchSubtreeDataForBlockFn` injection seam is real (`Server.go:195`, assigned at `:403`, read at `get_blocks.go:237`), so calling it out as a way to break the invariant is warranted rather than hypothetical.
- Optimistic-mode skip: `get_blocks.go:230`, `:234`.
- The cache-bypass retry that makes the drain ceiling a multiple of the per-call timeout: `bypassCache` threaded through `fetchAndStoreSubtree` / `fetchAndStoreSubtreeAndSubtreeData`.

Inertness is provable rather than asserted: every added line in the diff begins with `//`, and there are no non-comment additions at all.

## One precision note, non-blocking

The chain as written is "the goroutine stays inside the errgroup `fetchSubtreeDataForBlock` waits on, which `blockWorker` waits on, which `catchup` waits on". That is true transitively, but it skips two named hops, and the whole point of this comment is to save the next reader from re-deriving the chain. The actual path is:

```
fetchAndStoreSubtreeData goroutine
  -> fetchSubtreeDataForBlock errgroup      g.Wait()  get_blocks.go:430
  -> blockWorker                            (calls it synchronously)
  -> fetchBlocksConcurrently errgroup       g.Wait()  get_blocks.go:133
  -> fetchAndValidateBlocks                 catchup.go:1128
  -> catchup                                catchup.go:295  (Step 10)
  -> defer releaseCatchupLock               catchup.go:189
```

`fetchBlocksConcurrently` and `fetchAndValidateBlocks` are the two missing links. A reader who greps for where `catchup` waits on `blockWorker` will not find it, because nothing does that directly.

Related: the PR description attributes `get_blocks.go:133` to `blockWorker` ("which blockWorker waits on (get_blocks.go:133)"), but that line is `fetchBlocksConcurrently`'s wait. `blockWorker` waits at `:430`, by calling `fetchSubtreeDataForBlock` synchronously. The conclusion is unaffected; only the citation is off, and it does not appear in the shipped comment.

Adding the two names to the shipped chain would make it greppable. Your call whether that is worth another push.

services/blockvalidation/catchup.go
```diff
@@ -595,13 +595,46 @@ func (u *Server) releaseCatchupLock(ctx *CatchupContext, err *error) {
...
+// fetchSubtreeDataForBlockFn that does not preserve that join.
+// In adaptive-fetch optimistic mode the per-subtree fetch is skipped entirely, so on
+// that branch there is nothing to join and the guarantee holds vacuously rather than
+// by the join above.
+//
+// Not every caller is a catchup. RevalidateBlock reaches this path through
+// fetchSubtreeDataForBlock on its own gRPC goroutine with no interlock against a
+// running catchup, so its per-subtree failures land in whatever catchup cycle is
+// active. releaseCatchupLock drains failedPeers only inside its *err != nil branch,
+// so those failures are charged only if that concurrent cycle itself ends in error;
```
freemans13, line 632: "charged only if that concurrent cycle itself ends in error" is necessary but not sufficient. A cycle can end in error and still charge no one.

`recordCatchupPeerFailure` reads the pointer under `activeCatchupCtxMu.RLock` and releases it at `catchup.go:645`, then takes `failedPeersMu` and writes at `:651-658`. `releaseCatchupLock` holds `activeCatchupCtxMu` from `:391` through `:500` and copies the map at `:468-473`. Nothing orders those two against each other. `failedPeersMu` protects the map's contents, not the read-then-write against the drain. A `RevalidateBlock` goroutine descheduled between `:645` and `:651` writes into a map that has already been copied, or into a `CatchupContext` already cleared at `:499`, since it still holds the old pointer. Either way the failure dies with the map, on a cycle that did end in error.

What makes this a clause rather than an issue is that it cannot happen to a catchup's own fetches. The join documented in the paragraph above means the write finishes inside the errgroup goroutine, before `g.Wait` returns and long before `releaseCatchupLock` runs. It is reachable only through the caller with no interlock, which is the one this paragraph is already about.

It is narrow. The two have to interleave within a few instructions. And it lands on the "charged to no one" outcome you already name, so it widens how that outcome is reached rather than adding a fourth one. If you want the qualifier: "...only if that concurrent cycle itself ends in error, and only if the record lands before `releaseCatchupLock` copies the map".

freemans13 (approved):
Approving. Second round. The three commits land every item from the first, and I re-checked each against the tree rather than reading it off your reply.

Verification: `gofmt`, `go build` and `go vet` clean on `2cc42a789`. `services/blockvalidation` green under `-race`, 107s, no races reported. Upstream CI green across the full suite, including all seven sequential shards and three smoke shards, so this is not one of those fork PRs where the heavy jobs silently never ran.

**The drain gate (O1, F1).** Correct as written. The gate is `catchup.go:392`, the copy `:468-473`, the clear `:499`. The "charged to no one" outcome is now stated rather than left to inference.

**The timeout is a multiple, not one window (O2, F2).** Correct. `tryPeerForSubtree` (`get_blocks.go:734`) fetches at `:736` and again on a cache-bypass retry at `:751`, and `fetchAndStoreSubtreeAndSubtreeData` runs that whole thing once per alternative peer at `:805`, with the deadline installed per call at `:593`.

**Your correction to my citation is right and mine was wrong.** `settings.conf:457-461` is an annotated dead-key block for `blockvalidation_subtree_fetch_timeout`. The live key is `blockvalidation_subtree_data_fetch_timeout` at `:467`, read at `settings/settings.go:362`, defaulting to `DefaultSubtreeDataFetchTimeout` of 10m (`settings/blockvalidation_settings.go:16`). Naming the timeout in the comment instead of citing a line number is the right call for exactly this reason.

**Dropping "right by peer" (F3).** Good, and better than restoring it with a qualifier attached. `Server.go:1245-1246` says the fetch exists because the pruner removed the data, so a 404 from an honest peer is the expected case here, not the edge one.

**The injectable seam (F4).** Landed. One detail for the next reader, not a defect: the seam only covers catchup. `blockWorker` goes through `u.fetchSubtreeDataForBlockFn` (`get_blocks.go:237-239`), but `RevalidateBlock` calls `u.fetchSubtreeDataForBlock` directly (`Server.go:1254`), so an injected function cannot reach the path the second paragraph documents.

**Optimistic mode (F5).** Landed, and the claim holds more strongly than I expected. `blockWorker` sets `contributingPeers, err = nil, nil` at `get_blocks.go:234-236`, and the only two production calls to `recordCatchupPeerFailure` sit inside `tryPeerForSubtree` (`get_blocks.go:744`, `:753`), reachable only through `fetchSubtreeDataForBlock`. Nothing charges a peer on that branch at all, so "vacuously" is the precise word.

One thing worth putting in the record, because it raises what the second paragraph is worth. I enumerated the production entry points rather than sampling them: `recordCatchupPeerFailure` from `tryPeerForSubtree`, from `fetchAndStoreSubtreeAndSubtreeData`, from `fetchSubtreeDataForBlock`, which has exactly two callers, `blockWorker` and `RevalidateBlock`. So "Not every caller is a catchup. RevalidateBlock..." is the complete set on this tree and not one example of a class. That is more useful than it reads, because it tells whoever adds a third caller that they are the person who has to come back and rewrite this comment.

One new finding, inline and non-blocking, on the drain rule. The "ends in error" condition is necessary but not sufficient, and the gap sits precisely in the caller the paragraph is about.

PR #1598, opened 2026-08-17
fix(p2p): enable GossipSub peer scoring and refuse unscored peer-exchange records

gokutheengineer:
**Closes bitcoin-sv/teranode#4740.**

### Problem
The p2p service built GossipSub through `go-p2p-message-bus` v0.1.17, which passed exactly one option: `pubsub.WithPeerExchange(true)`. No peer scoring existed anywhere in the stack: no IP-colocation penalty, no behaviour penalty, and a zero `AcceptPXThreshold`, so every peer record handed to the node in a PRUNE was accepted and dialed. Because gossipsub marks peers the node dialed as outbound, unconditionally-accepted PX defeats both of gossipsub's built-in mesh defenses (the over-Dhi graft refusal and the Dout outbound quota), letting a zero-cost Sybil swarm fully capture the block/subtree/node_status/rejected_tx meshes and stall block and subtree discovery. The gossipsub v1.1 spec requires scoring whenever PX is enabled.

All items in the issue were confirmed still valid against `services/p2p/Server.go` and the v0.1.17 library source.

### Fix
- Bump `go-p2p-message-bus` v0.1.17 → v0.1.23, which adds first-class scoring support: penalty-only defaults (IP-colocation penalty, behaviour penalty for GRAFT flooding and broken IWANT promises, graylist at -8000), and static/bootstrap peers registered as GossipSub direct peers so trusted links are exempt from scoring.
- Add `p2p_enable_peer_scoring` (default **true**) and `p2p_enable_peer_exchange` (default **true**) settings, wired through a `buildP2PMessageBusConfig` seam in `services/p2p/Server.go`.
- **Refuse inbound PX records entirely while scoring is penalty-only** (`AcceptPXThreshold = 1`, above the maximum attainable score of 0). Accepting records from 0-scored peers would let the attacker get itself dialed, and PX-dialed peers register as outbound - bypassing the Dhi graft refusal and Dout quota, the strongest half of the issue's attack. PX emission stays on for other nodes; acceptance is re-enabled work for when positive scoring exists. (Addresses review ChiR1.)
- Add `p2p_peer_score_ip_colocation_threshold` (default 10, matching the library) so the colocation penalty can be tuned without disabling scoring. (ChiR2)
- **PX on with scoring off is now a startup configuration error** - the exact spec-violating state this PR eliminates cannot be reintroduced by config. (ChiR3)
- Expand the private-range colocation whitelist (applied when `p2p_allow_private_ips=true`) to RFC6598 shared space, IPv4/IPv6 link-local and IPv6 ULA, and panic on a malformed constant instead of silently shrinking the list. (ChiR4)
- `PeerScoreInspect`: warn only on transition/material change (a persistent graylisted swarm cannot flood the warn level), threshold read from the installed thresholds struct, tallies mutex-guarded. (ChiR5)
- Log the effective scoring configuration once at startup on both branches (params, thresholds, PX state), so "did scoring actually turn on, with what values" is answerable from one log line. (ChiR6)
- Bump `go-alert-system` v0.1.6 → v0.1.17: required because the message-bus bump moves kad-dht to v0.42, which changed `dht.New`'s signature; v0.1.17 is the lockstep-upgraded release. Its diff vs v0.1.6 is confined to the `dht.New` call and defensive `len()` guards in `MessageString()` methods. This bump cannot be split into its own PR - without it the module does not compile against kad-dht v0.42.
- Update `docs/p2p-libp2p-review.md`, the settings reference (including the corrected remediation guidance for colocated operators: the penalty is applied by each remote peer, so no local setting fixes it - distinct IPs or reciprocal static peers do), and `settings.conf`.

Honest framing of the guarantee: penalty-only scoring raises the cost of Sybil mesh capture and the PX-refusal closes the outbound-marking bypass, capping inbound-only Sybils at gossipsub's Dhi/Dout guards - but a patient freeloading Sybil on a fresh IP still scores 0. Per-topic delivery scoring (`TopicScoreParams`) is the follow-up hardening; it was deliberately not shipped here because `MeshMessageDeliveries` thresholds are dangerous on low-rate topics (blocks arrive ~every 10 min) and would penalize honest peers without careful per-topic tuning. It is also the prerequisite for re-enabling PX acceptance meaningfully.

### Dependency scope
Direct, intentional bumps: `go-p2p-message-bus` v0.1.23 and `go-alert-system` v0.1.17. Everything else in go.mod/go.sum is MVS fallout of those two: libp2p v0.48→v0.49, pubsub v0.15→v0.17, kad-dht v0.39→v0.42, go-sdk v1.2.24→v1.3.2, go-bt, grpc v1.81→v1.82, modernc.org/sqlite v1.47→v1.54, pgx v5.9→v5.10, gorm, prometheus/client_golang. The go-sdk `script/interpreter` and `transaction` diffs were audited (lint/cosmetic; `Flag.HasFlag` value→pointer receiver; no semantic consensus change found). The storage-driver bumps (sqlite/pgx/gorm) sit beneath `stores/blockchain` and `stores/utxo` and could not be exercised locally beyond the sqlite-backed tests (no Docker for TestContainers) - **merge should gate on the full CI suite including the Postgres TestContainers jobs**, which is the authoritative verification for those.

### Residual
- **Bootstrap peers become GossipSub direct peers network-wide.** The library direct-peers `StaticPeers ∪ BootstrapPeers` whenever scoring is on; since every node ships the same BSVA bootstrap dnsaddr, bootstrap nodes will receive O(N) copies of each message and hold connmgr-protected connections from every node. Not overridable from Teranode today; needs a library opt-out (direct-peer static peers only). Flagged in `docs/p2p-libp2p-review.md`.
- **A ≤threshold-per-IP freeloading swarm still scores 0** (see honest framing above). With PX acceptance refused it is capped by the Dhi/Dout guards at roughly 4 of 6 mesh slots per topic, so capture is partial and delivery degrades to gossip-pull latency rather than being silenced. Full closure needs per-topic delivery scoring.
- **`AppSpecificScore` is not wired to Teranode's ban/reputation state.** Banned peers are already network-disconnected and IP-banned (`server_helpers.go` ban enforcement), so this is defense-in-depth (mesh eviction + score gating before reconnect) and needs peer-ID→IP plumbing that deserves its own change.
- **6h `RetainScore`**: with penalty-only params every disconnecting peer's score entry is retained for 6h (the score>0 fast-drop path is unreachable). Entries are small and connection churn is bounded by the conn manager, but worth an upstream discussion about a shorter retention for penalty-only profiles.
- Per-topic `InvalidMessageDeliveries` penalties remain inert: the bus still does not expose `RegisterTopicValidator` (tracked as the outstanding library change in `docs/p2p-libp2p-review.md`).

### Tests
- `TestBuildP2PMessageBusConfig_MeshProtection` (services/p2p/server_bus_config_test.go): scoring/PX pass-through including the enable→disable inversion; PX-without-scoring returns a configuration error; explicit params with `AcceptPXThreshold > 0` whenever scoring is on; colocation threshold override honoured (0 keeps the library default); whitelist only under `AllowPrivateIPs`; params stay nil when scoring is off (non-nil would force-enable scoring in the bus); score inspection warns on transition only and never repeats for an unchanged graylisted set.
- `TestPrivateIPColocationWhitelist`: pins the whitelisted ranges (in and out).
- `TestP2PGossipSubMeshProtectionSettings` and `TestP2PPeerScoreIPColocationThreshold_LoaderReadsKey` (settings/p2p_settings_test.go): secure defaults plus loader-reads-overrides, written context-proof per the batcher-test precedence pattern.
- Ran (scoped):
  - `SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./settings/...` — ok
  - `SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/p2p/...` — ok
  - `SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/alert/...` — ok (covers the go-alert-system bump)
  - `SETTINGS_CONTEXT=test go test -tags "testtxmetacache" -count=1 ./stores/blockchain/...` — sqlite-backed tests ok; Postgres TestContainers tests fail locally only for "rootless Docker not found" (no Docker in this environment; see Dependency scope - CI is authoritative)
  - `go vet ./services/p2p/... ./settings/... ./services/alert/...` — clean


liam (approved):
Approving on the code. One request before merge: document the real dependency scope.

**Verified against the library source.** I pulled go-p2p-message-bus v0.1.23 into the module cache and read `scoring.go`. The thing that looked dangerous is not: `PeerScoreParams`/`PeerScoreThresholds` are only set inside the `AllowPrivateIPs` branch, but they are pointer fields, and `resolvePeerScoreConfig` substitutes `DefaultPeerScoreParams()`/`DefaultPeerScoreThresholds()` whenever `EnablePeerScoring` is true and they are nil. `ErrIncompletePeerScoreConfig` fires only when exactly one is set, which this code never does. No zero-value hazard.

Every parameter in the description checks out against the source: IP-colocation -35 / threshold 10, behaviour -16 / threshold 6, decay 0.928, `RetainScore` 6h, `AcceptPXThreshold` 0, graylist -8000, and the direct-peer exemption for static/bootstrap peers in `buildPubSubOptions`. Builds clean; `services/p2p` and `settings` pass; a sweep over `stores/... model/... services/validator/... services/blockchain/...` came back with zero failures.

## The dependency scope is wider than the description admits

The body says the alert bump's "diff vs v0.1.6 is confined to the `dht.New` call and defensive `len()` guards". That is true of alert-system's *own source*, but not of its `go.mod`. Alert v0.1.17 raises the module-graph floor, and MVS drags Teranode's own direct dependencies forward with it:

| Dependency | main | this PR |
|---|---|---|
| `bsv-blockchain/go-bt/v2` | v2.6.7 | **v2.6.8** |
| `bsv-blockchain/go-sdk` | v1.2.24 | **v1.3.2** |
| `modernc.org/sqlite` | v1.47.0 | **v1.54.0** |
| `mrz1836/go-datastore` | v1.2.6 | v1.3.7 |
| `google.golang.org/grpc` | v1.81.1 | v1.82.1 |

plus libp2p v0.48→v0.49, quic-go v0.59→v0.60, the whole pion/WebRTC stack, prometheus client, otel and `golang.org/x/*`.

In your favour, and worth saying plainly: these are **forced by MVS, not hand-edited** — I diffed alert v0.1.6's and v0.1.17's `go.mod` to confirm. That is the difference between "someone ran `go get -u`" and "this is the unavoidable cost of the alert bump", and the description should draw that distinction rather than leave a reviewer to guess. As written, nothing tells a reader that a p2p-scoring PR moves the BSV transaction library and the SQLite driver behind the blockchain and UTXO stores.

I checked the consensus-relevant surface Teranode actually imports from go-sdk (`script`, `script/interpreter`, `primitives/ec`, `chainhash`, `compat/base58`). The deltas in `script/interpreter/operations.go`, `script/script_chunk.go` and `transaction/signaturehash.go` are **cosmetic**: `//nolint` directive reformatting, gofmt alignment, named-return to explicit-return conversion, a `_` parameter rename, one capacity hint on a `make`. No behaviour change. So the risk is low — but it should be stated rather than discovered, and CI should be green on the full suite, not just p2p. Ideally the bump lands as its own PR so a future bisect points at the right change.

## IP-colocation vs multi-node operators

`p2p_allow_private_ips` is false in production, so the RFC1918 whitelist does not apply and colocation keys on the **public** IP. An operator running more than 10 nodes behind one NAT gateway gets peers 11+ penalized, and the penalty scales with the square of the surplus: at 20 peers that is 10² × -35 = **-3500**, past `GossipThreshold` (-2000) though short of graylist (-8000), so their gossip is ignored. Teranode targets exactly that class of operator. Not a defect, but it is the most likely source of a "why did my mesh degrade after upgrading" report, and it deserves a line in the settings doc and a heads-up to ops.

## Two smaller things

- An operator can set `p2p_enable_peer_exchange=true` with `p2p_enable_peer_scoring=false` — the spec-violating combination this PR exists to fix. The library logs a warning, but nothing in Teranode surfaces it. Since both settings are new here, consider refusing to start, or logging at Error, on that combination rather than relying on a library-level `Warnf`.
- `privateIPColocationWhitelist()` silently drops any CIDR that fails to parse. They are compile-time constants so it can never fire, but a `must`-style panic reads better than `if err == nil`. Also worth a comment on why `fc00::/7` and `169.254.0.0/16` are absent while `::1` is present.

ctnguyen (changes requested):
Review at commit [100eff4](https://github.com/gokutheengineer/teranode/commit/100eff4fbe82e276def9bee04d84808a2f1267ff)

This is a well-argued change and the write-up is unusually honest for a security PR — the "Residual" section names the direct-peer fan-out on bootstrap nodes, the unwired `AppSpecificScore`, and the fact that penalty-only scoring never awards positive score, instead of quietly claiming a full fix. The `buildP2PMessageBusConfig` seam is the right refactor: it turns a config literal buried 300 lines into `NewServer` into something testable, and `server_bus_config_test.go` earns its keep by pinning the enable→disable inversion on peer exchange and by asserting that score params stay nil when scoring is off. Enabling the behaviour penalty and wiring `PeerScoreInspect` are both real gains that didn't exist before. My one substantial concern is about the *values* the shipped parameters actually take versus the exploit described in the issue this PR closes; the rest is documentation accuracy and observability polish.

## Blocking issues

### ChiR1 — The shipped score parameters do not penalize the exploit in the issue this PR closes, and PX acceptance stays effectively ungated

**Problem:** Two library defaults combine to leave the issue's stated attack untouched. `IPColocationFactorThreshold` is 10 and pubsub only applies the penalty when `peersInIP > threshold` (`ipColocationFactor`: `if peersInIP > ps.params.IPColocationFactorThreshold`). The exploit in [bitcoin-sv/teranode#4740](https://github.com/bitcoin-sv/teranode/issues/4740) is *"200 libp2p peers across 20 IPs"* — exactly 10 per IP, i.e. exactly at the cliff and therefore contributing 0 from P6. The behaviour penalty only fires on re-GRAFT-before-backoff and broken IWANT promises, which a patient freeloader never does. So the literal configuration in the issue still scores 0 and remains fully mesh-eligible.

Second, `AcceptPXThreshold` is 0 (confirmed against v0.1.23 in @liam's review). pubsub's `handlePrune` accepts PX whenever `score >= acceptPXThreshold`, and under penalty-only params **no peer can ever score above 0** — so the gate excludes only peers already penalized for something else. Every 0-scored Sybil's PX records are still accepted and dialed, and `pxConnect` dials mean those peers register as *outbound*, which is precisely what defeats the `len(peers) >= Dhi && !gs.outbound[p]` graft refusal in `handleGraft` and satisfies the Dout quota. That bypass — the strongest part of the issue's argument — is unmitigated.

The description's honest framing covers half of this ("a patient freeloading Sybil on a fresh IP still scores 0") but not the Dhi/Dout consequence, and the title claims Sybil peers "cannot capture the mesh for free".

**Why it matters:** The PR is marked `Closes bitcoin-sv/teranode#4740`, so merging it retires the issue while the exploit sequence written in that issue still works unchanged, with no adaptation required from the attacker. Unlike the residuals the description already discloses, this one is reachable from Teranode today — `PeerScoreThresholds` is a Config field this code already knows how to set.

**Fix:** Raise `AcceptPXThreshold` above the maximum attainable score so PX records are never accepted while positive scoring does not exist — that is the issue's proposed interim mitigation, applied at the layer that actually gates acceptance (`WithPeerExchange(false)` only stops *emitting* PX; it does not stop `handlePrune` from dialing attacker-supplied records). Because the thresholds are currently only constructed inside the `AllowPrivateIPs` branch, that construction has to move up to cover every scoring-enabled deployment:

```go
if tSettings.P2P.EnablePeerScoring {
    thresholds := p2pMessageBus.DefaultPeerScoreThresholds()
    // Penalty-only params cap every score at 0, so any positive threshold means
    // "never accept PX". Until per-topic delivery scoring can lift honest peers
    // above 0, refuse attacker-supplied peer records outright: pxConnect dials
    // mark those peers outbound, which bypasses gossipsub's Dhi graft refusal
    // and satisfies the Dout quota.
    thresholds.AcceptPXThreshold = 1
    conf.PeerScoreThresholds = thresholds
    conf.PeerScoreParams = ... // see ChiR4 on the whitelist
}
```

If you would rather keep PX acceptance as-is, please downgrade `Closes #4740` to a `Refs`, and record in the issue that the ≤10-per-IP swarm and the outbound bypass are still open so the follow-up hardening isn't lost. Separately, consider exposing `IPColocationFactorThreshold` as a setting (see ChiR2) so it can be tightened below 10 on networks without colocated operators.

## Non-blocking issues

### ChiR2 — Colocated operators are penalized by *remote* peers, so the settings doc's remediation advice does not work

*Same as [@liam's point "IP-colocation vs multi-node operators" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1598#pullrequestreview-4960379884).*

**Problem:** This extends @liam's point in the direction that changes the advice. `ipColocationFactor` counts `len(ps.peerIPs[ip])` — peers connected to *the scoring node* from that IP. The penalty is therefore computed independently by every remote peer, against the colocated operator's nodes. `docs/references/settings/services/p2p_settings.md:68` says such deployments "will penalize each other's nodes" and offers "or disable scoring on that isolated network" as a remedy. Both understate it: the intra-operator half is only one of two, and turning scoring off locally has no effect at all on how the rest of the network scores you. The `StaticPeers` remedy also only helps the specific pairs that list each other reciprocally, not the anonymous remote peers doing the penalizing.

Note the limit on how far this reaches: the count is per-observer, so a remote peer holding only a handful of connections into the colocated set applies no penalty at all. The exposure is to peers that happen to connect to 11 or more of the same operator's nodes, which is unpredictable rather than certain.

There is also no way to relax the colocation penalty alone. The only knob is the whole-feature boolean, so an operator who trips this must give up the behaviour penalty and PX gating too — reverting to exactly the state this PR exists to fix.

**Why it matters:** Teranode targets horizontally scaled operators, and a node scoring below `GossipThreshold` at its peers loses IHAVE/IWANT service, not just mesh slots. Shipping remediation guidance that doesn't remediate turns a support ticket into a wrong fix.

**Fix:** Correct line 68 to say the penalty is applied by each remote peer that holds more than `IPColocationFactorThreshold` connections to the same source IP, that no local setting changes that, and that distinct public IPs (or reciprocal static-peer entries with the specific peers involved) are the only real mitigations. Then add a setting so the penalty can be tuned without disabling scoring, e.g.:

```go
// settings/p2p_settings.go
PeerScoreIPColocationThreshold int `key:"p2p_peer_score_ip_colocation_threshold" default:"10" ...`
```

and honour it in `buildP2PMessageBusConfig` on the same params struct ChiR1 already moves up.

### ChiR3 — `p2p_enable_peer_exchange=true` with `p2p_enable_peer_scoring=false` is accepted silently

*Same as [@liam's point "Two smaller things" bullet 1 (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1598#pullrequestreview-4960379884).*

**Problem:** `buildP2PMessageBusConfig` maps both flags straight through with no cross-check, so an operator can configure the exact spec-violating combination this PR was written to eliminate. This one is @liam's finding and I add no analysis to it — I'm seconding it because the fix is a four-line guard, and because `NewServer` already rejects a bad `listen_mode` with `errors.NewConfigurationError` (`Server.go:364-365`), so the precedent for refusing to start on an invalid P2P combination is in the same function the guard belongs in.

**Why it matters:** Both keys are new in this PR, so this is the one moment where a guard costs nothing. Relying on a library-level `Warnf` puts the only signal in a log line that nobody greps for, on the one configuration that is strictly worse than either flag alone.

**Fix:** Validate the pair in `NewServer` before building the config — either refuse to start, or log at `Errorf` if you'd rather not break an existing private deployment.

```go
if tSettings.P2P.EnablePeerExchange && !tSettings.P2P.EnablePeerScoring {
    return nil, errors.NewConfigurationError("p2p_enable_peer_exchange requires p2p_enable_peer_scoring (gossipsub v1.1 pairs PX with scoring); set one of them to match")
}
```

### ChiR4 — The colocation whitelist misses IPv6 private ranges and IPv4 link-local, and swallows parse failures

*Same as [@liam's point "Two smaller things" bullet 2 (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1598#pullrequestreview-4960379884).*

**Problem:** `privateIPColocationWhitelist` (`Server.go:199-207`) covers `127.0.0.0/8`, `::1/128` and the three RFC1918 blocks, but not `fc00::/7` (IPv6 ULA), `fe80::/10` (IPv6 link-local), `169.254.0.0/16`, or `100.64.0.0/10` (shared address space, common as cloud/CNI pod ranges). Including `::1/128` while omitting `fc00::/7` is the inconsistency @liam flagged; my concern is that it's a functional gap, not just a missing comment — a dual-stack or IPv6-only local cluster with `AllowPrivateIPs=true` gets no exemption at all, which is exactly the scenario the function exists for. Today `p2p_listen_addresses` defaults to `0.0.0.0` so inbound is IPv4 in practice, which limits the blast radius, but nothing pins that.

Separately, `if _, n, err := net.ParseCIDR(cidr); err == nil` drops any unparseable entry without a trace. It cannot fire on today's constants, but the failure mode of a future typo is a silently shorter whitelist, i.e. exactly the penalty this function is meant to prevent.

**Why it matters:** A silently incomplete whitelist makes local multi-node clusters mesh-ineligible with no error and no log, which is a confusing way to lose test-network connectivity.

**Fix:** Add the missing ranges and make a parse failure loud rather than silent.

```go
func privateIPColocationWhitelist() []*net.IPNet {
    // Loopback, RFC1918, RFC6598 shared space, IPv4 link-local, IPv6 ULA and
    // IPv6 link-local: the ranges a local/test cluster can legitimately share.
    cidrs := []string{
        "127.0.0.0/8", "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16",
        "100.64.0.0/10", "169.254.0.0/16",
        "::1/128", "fc00::/7", "fe80::/10",
    }
    nets := make([]*net.IPNet, 0, len(cidrs))
    for _, cidr := range cidrs {
        _, n, err := net.ParseCIDR(cidr)
        if err != nil {
            panic(fmt.Sprintf("privateIPColocationWhitelist: bad constant CIDR %q: %v", cidr, err))
        }
        nets = append(nets, n)
    }
    return nets
}
```

### ChiR5 — `PeerScoreInspect` re-warns on every tick, and reads its threshold from the defaults rather than the installed ones

**Problem:** The inspect callback (`Server.go:250-272`) emits a `Warnf` on every invocation for as long as any peer sits below the graylist threshold. The inspect period is chosen inside the bus and is not visible or configurable from here, so the log rate is unknown from Teranode's side. Reaching graylist (-8000) under the default colocation weight of -35 needs roughly 16 surplus peers on one IP — about 26 connections from a single host — which is cheap for a remote attacker to set up and then simply leave running, producing an indefinite warn-per-tick stream. The line is also fully re-derived each tick, so a steady state is indistinguishable from an escalating one in the log.

Second, `graylistThreshold` at `Server.go:249` is read from a fresh `DefaultPeerScoreThresholds()` rather than from the thresholds the config ends up carrying. Those agree today, on both branches, so nothing misreports right now — but ChiR1's fix changes `conf.PeerScoreThresholds` in place, and the moment any threshold is customised this line reports against a value that is no longer installed.

**Why it matters:** The whole point of wiring the inspector was attribution. An unbounded repeating line at warn level is the fastest way to get the signal filtered out or the level lowered, and a threshold read from the wrong source turns the log into a misleading one rather than a missing one.

**Fix:** Derive the threshold from the same struct that goes into the config, and log on transition and material change rather than on every tick.

```go
thresholds := p2pMessageBus.DefaultPeerScoreThresholds()
// ... any customisation (see ChiR1) ...
conf.PeerScoreThresholds = thresholds

graylistThreshold := thresholds.GraylistThreshold
var lastGraylisted int
var lastWorst peer.ID
conf.PeerScoreInspect = func(snapshots map[peer.ID]*pubsub.PeerScoreSnapshot) {
    // ... existing tally ...
    if graylisted > 0 && (graylisted != lastGraylisted || worstPeer != lastWorst) {
        logger.Warnf("[p2p] gossipsub peer scores: %d negative, %d graylisted, worst %s score %.0f", negative, graylisted, worstPeer, worst)
    } else if negative > 0 {
        logger.Debugf(...)
    }
    lastGraylisted, lastWorst = graylisted, worstPeer
}
```

Note the closure would then hold mutable state across pubsub-driven invocations; if the bus can call it concurrently, guard it with a mutex or keep the counters in an atomic.

### ChiR6 — Nothing logs whether peer scoring actually ended up active

**Problem:** `NewServer` logs silent mode, DHT mode and advertise addresses on the way to building the bus config, but `buildP2PMessageBusConfig` emits nothing. So for the security control this PR turns on by default there is no line in the startup log saying whether scoring is on, whether PX is on, or which colocation parameters are in force — and in the common production case (`AllowPrivateIPs=false`) the parameters are the bus's rather than ones visible in Teranode's own code, so an operator has no local way to read them off.

**Why it matters:** This is the first thing anyone will want when triaging a "my mesh degraded after upgrading" report — the ChiR2 scenario, precisely. Without it, confirming the effective configuration means reading two repositories at the right versions.

**Fix:** Log the effective configuration once, on both branches.

```go
if tSettings.P2P.EnablePeerScoring {
    logger.Infof("[p2p] gossipsub peer scoring enabled: ip-colocation weight %.0f threshold %d (whitelisted ranges %d), graylist %.0f, acceptPX %.0f, peer exchange %v",
        params.IPColocationFactorWeight, params.IPColocationFactorThreshold,
        len(params.IPColocationFactorWhitelist), thresholds.GraylistThreshold,
        thresholds.AcceptPXThreshold, tSettings.P2P.EnablePeerExchange)
} else {
    logger.Warnf("[p2p] gossipsub peer scoring DISABLED (p2p_enable_peer_scoring=false), peer exchange %v", tSettings.P2P.EnablePeerExchange)
}
```

This reads naturally off the explicit `params`/`thresholds` locals that ChiR1 introduces.

### ChiR7 — Dependency sweep spans consensus-adjacent and storage libraries, but only scoped tests were run

*Same as [@liam's point "The dependency scope is wider than the description admits" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1598#pullrequestreview-4960379884).*

**Problem:** Agreeing with @liam and adding the piece that worries me most concretely: the bumps that got the least verification are the storage ones. `jackc/pgx/v5` v5.9.2→v5.10.0, `mrz1836/go-datastore` v1.2.6→v1.3.7, `gorm` and `modernc.org/sqlite` v1.47→v1.54 all sit directly beneath `stores/blockchain` and `stores/utxo`, and the description states the Postgres TestContainers tests could not run locally. So the two dependency changes with the widest data-path reach are the two the author explicitly could not exercise, while the reported runs (`./settings/...`, `./services/p2p/...`, `./services/alert/...`) cover none of them.

**Why it matters:** A future bisect on a store-layer regression will land on a PR titled "enable GossipSub peer scoring", and the reviewer of that bisect has nothing in this description telling them the SQLite driver and the Postgres driver moved here.

**Fix:** Gate merge on a green full-suite CI run (Postgres TestContainers included), not the scoped runs, and add @liam's MVS-forced-versus-hand-edited distinction plus the go-sdk surface audit to the description. If the alert-system bump can be split into its own PR, that is the cleaner outcome — the p2p change then reads as a ~250-line diff instead of a ~1300-line one.

### ChiR8 — The settings reference this PR edits still omits the `silent` listen mode in both places it enumerates them

**Problem:** `docs/references/settings/services/p2p_settings.md:16` describes `ListenMode` as `Node operation mode ("full" or "listen_only")`, and line 101 in the Validation Rules table says `Must be "full" or "listen_only"`. `silent` is missing from both, and from the file entirely. `Server.go:364-365` accepts all three (`ListenModeFull`, `ListenModeListenOnly`, `ListenModeSilent`) and `settings/p2p_settings.go:15` documents `silent` in full, including its own trade-off table. This is pre-existing staleness rather than something this PR introduced — but this PR is editing this file, and the validation table specifically tells a reader what will stop the service from starting.

**Why it matters:** To be precise about the limit: nothing malfunctions. `listen_mode=silent` works, and the code's own error message on an invalid value lists all three modes. The cost is that the reference doc actively tells an operator the mode will be rejected, so the one mode aimed at security-sensitive deployments looks unsupported to anyone reading the docs rather than the source.

**Fix:** Update both rows while you're in the file.

```markdown
| ListenMode | string | "full" | listen_mode | Node operation mode ("full", "listen_only", or "silent") |
...
| ListenMode | Must be "full", "listen_only" or "silent" | Service fails to start if invalid | During server initialization |
```

## Recap

|---|---|---|---|
| ChiR1 | Issue exploit still unpenalized; PX gate inert | YES | 80% |
| ChiR2 | Remote-applied colocation penalty; wrong doc advice | | 48% |
| ChiR3 | PX-on/scoring-off combination unguarded | | 40% |
| ChiR4 | Whitelist gaps; silent CIDR parse failure | | 30% |
| ChiR5 | Repeating warn; threshold from wrong source | | 26% |
| ChiR6 | No startup log of scoring config | | 22% |
| ChiR7 | Storage dependency bumps untested | | 20% |
| ChiR8 | Settings doc omits silent mode | | 14% |

ctnguyen (approved):
Review at commit [7731070](https://github.com/gokutheengineer/teranode/commit/77310709a9ee754ef77ef778e8b77b0efb8178ff)

This round did the work. `AcceptPXThreshold = 1` now sits on the single scoring-enabled path in `buildP2PMessageBusConfig` (`services/p2p/Server.go:265-274`) rather than hiding inside the `AllowPrivateIPs` branch, which was the thing I was most worried about last time — PX acceptance is genuinely closed now, not nominally gated. The PX-without-scoring guard, the colocation threshold setting, the widened whitelist with a loud panic, the startup log on both branches: all present, all tested, and the tests actually pin the risky behaviour rather than the easy behaviour. The description's honest framing survived the rewrite too, which is rarer than it should be. What's left is mostly about the *claims* the PR makes rather than the code it ships: the residual list still stops one layer short of the resource that actually binds the attack, PX emission is left on with nothing to gain from it, and the reassurance the new settings doc offers colocated operators rests on a detail the PR itself raises and doesn't resolve. One previous point is only partly fixed and one is untouched.

## Status of previous points

`ChiR5` is the partial. The threshold is now read from the installed `thresholds` struct (`Server.go:281`) and the tallies are mutex-guarded (`Server.go:282-306`), so both of those are properly closed. The warn-suppression is the part that does not hold, and the flaw is in the wording my previous round handed over — the `graylisted != lastGraylisted || worstPeer != lastWorst` condition was the suggested fix and the author implemented it as given, so this is me revising that advice rather than the author getting it wrong. `changed` keys on `worstPeer` identity, and `worstPeer` is picked with `snap.Score <= worst` (`Server.go:294`), so when several peers hold the same worst score the winner is whichever key Go's randomised map iteration visits last. Selection across ticks is therefore non-deterministic, and the gate can retrigger the warn on a tick where nothing material changed — the persistent-graylisted-swarm case it was added to stop. Equal scores are what a single-IP colocation swarm produces *if* the colocation term is the only non-zero one under penalty-only params; that parameter structure comes from this review's context rather than from source in this worktree, so treat the tie premise as reported rather than verified, and how often the gate fails open as unquantified. That it can fail open is what the tree shows. Tie-break deterministically (`snap.Score < worst || (snap.Score == worst && pid < worstPeer)`), or key the transition on the graylisted count plus a bucketed worst score instead of on peer identity.

Two smaller things in the same callback while you are there: recovery is never warned — when `graylisted` falls back to 0 the branch drops to `Debugf`, so a warn-level watcher sees the onset of an attack and never sees it end. And the case at `services/p2p/server_bus_config_test.go:169-175` is labelled "swarm shrinks below graylist ... warn again on recovery-relevant change" but actually adds a third peer at -8500, taking the graylisted count from 1 to 2; that is an escalation, not a recovery, so the recovery path is untested.

`ChiR7` is marked fixed on the strength of the rewritten description — the MVS-forced versus hand-edited distinction, the go-sdk surface audit, and the reason the alert bump cannot be split are all there now. The full-suite CI gate it asks for is a merge-time condition, not something the tree can show. `ChiR8` is untouched: `docs/references/settings/services/p2p_settings.md:16` still reads `("full" or "listen_only")` and line 103 still says `Must be "full" or "listen_only"`.

| ChiR1 | Score params and PX gate inert | YES | FIXED |
| ChiR2 | Remote-applied colocation penalty | | FIXED |
| ChiR3 | PX-on with scoring-off unguarded | | FIXED |
| ChiR4 | Whitelist gaps, silent parse failure | | FIXED |
| ChiR5 | Repeating warn, threshold source | | PARTIAL FIXED |
| ChiR6 | No startup log of scoring | | FIXED |
| ChiR7 | Storage dependency bumps untested | | FIXED |
| ChiR8 | Settings doc omits silent mode | | NOT FIXED |

## Non-blocking issues

### ChiR9 — The residual accounting stops at the mesh, but the binding resource is the connection budget

**Problem:** The description's ceiling on what survives this PR is stated in mesh terms: "capped by the Dhi/Dout guards at roughly 4 of 6 mesh slots per topic, so capture is partial and delivery degrades to gossip-pull latency rather than being silenced". The "degrades rather than silences" half of that leans on [bitcoin-sv/teranode#4740](https://github.com/bitcoin-sv/teranode/issues/4740)'s verification note 3 — honest peers still emit IHAVE to D_lazy random *non-mesh topic peers*. But the same note continues: total eclipse "needs most of the topic-peer set, which is cheap given the bus default `maxConns=35` and an indiscriminate `TagPeer(..., 100)` that shields Sybil connections from connmgr trimming". If the swarm holds most of the 35 connection slots and cannot be trimmed, the D_lazy pool the consolation depends on is Sybil-dominated too, and the mesh-slot arithmetic is no longer the constraint that matters. This repo already carries the same observation independently: `docs/p2p-libp2p-review.md:117` and should-fix item 7 at line 134 record min=25/max=35/grace=20s, that Teranode does not override it, and that gossipsub wants D=6 per topic — 24 slots across four topics before bootstrap and DHT. Nothing this PR ships touches the connection budget or the tagging, and neither the Residual list nor `docs/p2p-libp2p-review.md`'s updated section 2 mentions that the mesh ceiling is conditional on it.

**Why it matters:** This is the one remaining piece of the linked issue's argument that the PR neither closes nor discloses, and the PR is marked `Closes` that issue — so it retires with an undisclosed dependency. Every other limit in the Residual section is named honestly, which makes this omission read as a gap in the analysis rather than a scoping choice.

**Fix:** Add it to Residual: state that the Dhi/Dout ceiling and the gossip-pull fallback both assume the connection budget is not itself captured, that connmgr limits and the indiscriminate peer tagging are library-side and not reachable from Teranode today, and link should-fix item 7 in `docs/p2p-libp2p-review.md` as the follow-up. Given the issue's note 3 names this as what makes total eclipse cheap, it is also worth deciding deliberately whether `Closes bitcoin-sv/teranode#4740` or `Refs` is the honest trailer.

### ChiR10 — PX emission stays on by default while acceptance is refused, so the harvesting half of the PX risk is retained for no local benefit

**Problem:** `DisablePeerExchange: !tSettings.P2P.EnablePeerExchange` (`Server.go:250`) with the setting defaulting to true means every node keeps emitting PX records while `AcceptPXThreshold = 1` guarantees it will never use anyone else's. `settings/p2p_settings.go:107` describes emission plainly: "when pruning a peer from the mesh, the node hands it a sample of other peers it knows". [bitcoin-sv/teranode#4740](https://github.com/bitcoin-sv/teranode/issues/4740) names two PX harms — an attacker can "both harvest the node's mesh view and seed Sybil addresses on every prune". This PR closes the seeding half and keeps the harvesting half, and an attacker can induce prunes cheaply by joining a topic and letting heartbeat mesh maintenance do the rest. The justification given is that "PX emission stays on for other nodes", but every node running this code also refuses acceptance, so within the Teranode network the emission benefits nobody; the beneficiaries are pre-scoring nodes and anyone mapping the topology. Nothing in the description, `settings.conf:910-915`, or the settings reference discloses the retained half. `listen_mode=silent` makes it sharper: `Server.go:495-498` and `534-537` suppress advertisement and DHT so the node "remains undiscoverable", and then PX emission hands out its topic peers anyway.

**Why it matters:** Peer-record emission with addresses is exactly the reconnaissance primitive a targeted eclipse wants, and it is now pure cost — the PR removed the only thing PX was buying this node. It also means the security posture is documented as "PX is on but safe" when the accurate statement is "PX is emit-only, and emit is the half we kept".

**Fix:** Either default emission off while acceptance is refused, and say so — the two flags are already independent and `scoring=true, PX=false` is a tested configuration:

```
settings.conf
p2p_enable_peer_exchange = false   # acceptance is refused; emission buys this node nothing
```

or keep the default and disclose the trade in Residual plus the `EnablePeerExchange` longdesc: emission leaks a sample of this node's topic peers with addresses to any peer it prunes, including one that induces the prune, and that is retained deliberately for the benefit of nodes still running pre-scoring builds. Consider forcing it off under `ListenModeSilent`, where it contradicts the mode's stated intent.

### ChiR11 — The new colocation guidance promises live-connection semantics, which the PR's own 6h retention residual puts in doubt

*Same as [@liam's point "IP-colocation vs multi-node operators" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1598#pullrequestreview-4960379884).*

**Problem:** This extends @liam's point past the "document it" ask into whether the documentation that landed is correct. `docs/references/settings/services/p2p_settings.md:69` closes with a reassurance: "the exposure is per-observer: a remote peer holding only a few connections into the colocated set applies no penalty" — present tense, live connections. The Residual section of this PR says the opposite may be true of the underlying state: "with penalty-only params every disconnecting peer's score entry is retained for 6h (the score>0 fast-drop path is unreachable)". Retention is conditioned on a non-positive score precisely so that a penalized peer cannot reset itself by reconnecting, and the colocation penalty is computed from a per-IP peer set (`len(ps.peerIPs[ip])`, as established in the previous round and in @liam's reading of the shipped defaults). Whether a retained entry still occupies that per-IP set decides whether the doc sentence is true. If it does, the penalty is computed against a rolling 6h peak rather than current connections, so a NAT'd operator who briefly exceeds the threshold — a rolling restart, a redeploy, a flapping link — stays penalized at that observer for hours after dropping back under it, and the remaining under-threshold nodes are penalized alongside. The PR treats the 6h retention purely as a memory-footprint question ("entries are small") and does not resolve the functional one, yet ships the operator-facing reassurance that depends on it.

**Why it matters:** This is the sentence a support engineer will quote back at a colocated operator during exactly the "why did my mesh degrade after upgrading" call @liam predicted. Getting it wrong in the reassuring direction is worse than not writing it, because it will send the investigation away from the real cause.

**Fix:** Confirm the retention behaviour against the pinned `go-libp2p-pubsub` v0.17.0 scoring implementation — specifically whether the peer-removal path releases the peer's IP entries when it retains a non-positive score, or only when it drops the entry. Then either keep line 69 as-is, or qualify it and add the consequence:

```
docs/references/settings/services/p2p_settings.md
... the exposure is per-observer, but not instantaneous: a non-positive score
entry is retained for RetainScore (6h) after disconnect, so the penalty can
reflect a recent peak in connections from that IP rather than the current count.
```

Either way it belongs in Residual next to the existing retention note, since it changes that note from a footprint concern into an operator-visible one.

### ChiR12 — `AcceptPXThreshold = 1` encodes an invariant about the library defaults that nothing pins

**Problem:** The whole PX refusal rests on one unstated premise: that the bus's `DefaultPeerScoreParams()` contains no positive-score term, so the maximum attainable score is 0 and any positive threshold means "never". The premise is true today (@liam confirmed the shipped defaults are penalty-only), and it is asserted as fact in four places — the comment at `Server.go:266-270`, `settings.conf:912`, `settings/p2p_settings.go:94` and `:107`, and `docs/references/settings/services/p2p_settings.md:72` — but nothing enforces it. The test only checks the sign: `require.Greater(t, conf.PeerScoreThresholds.AcceptPXThreshold, 0.0)` (`services/p2p/server_bus_config_test.go:86`). The PR's own roadmap is to add positive per-topic delivery scoring, and it is the same library that would add it to the defaults. On the day `DefaultPeerScoreParams()` gains any positive term, `1` stops meaning "refuse all PX" and starts meaning "accept from anyone above a trivially reachable score" — silently, with a green test suite and five pieces of documentation now saying something false.

**Why it matters:** A dependency bump would re-open the bypass this PR exists to close, with no failing test and no log line to notice it by. That is the failure mode most likely to survive review, because the diff that causes it will be a one-line version change in `go.mod`.

**Fix:** Assert the premise where it is used, so a library change that invalidates it breaks the build or the test rather than the security property:

```go
// services/p2p/server_bus_config_test.go
// AcceptPXThreshold = 1 only means "refuse all PX" while nothing can score above 0.
require.Empty(t, conf.PeerScoreParams.Topics,
    "positive per-topic scoring makes AcceptPXThreshold=1 a reachable threshold, not a refusal")
require.Zero(t, conf.PeerScoreParams.AppSpecificWeight)
```

and reference that constraint from the comment at `Server.go:266`, so whoever enables `TopicScoreParams` later is told to revisit the threshold rather than inheriting it.

### ChiR13 — The new startup-failing validation is missing from the Validation Rules table it belongs in

**Problem:** `buildP2PMessageBusConfig` now refuses to start on PX-with-scoring-off (`Server.go:232-234`), which is a new way for the p2p service to fail at boot. `docs/references/settings/services/p2p_settings.md:101-107` is the table whose entire purpose is enumerating those — it already carries `ListenMode` with "Service fails to start if invalid" — and this PR edits that file without adding a row. The condition is described in prose at line 37 and line 72, so a reader who scrolls will find it; a reader who goes straight to the validation table will not.

**Why it matters:** Both keys are new and default to a combination that starts fine, so the first person to meet this error will be an operator deliberately turning scoring off on an isolated network — and the table they would check first says nothing about it.

**Fix:** One row, while in the file (this pairs naturally with the still-open ChiR8 rows in the same table):

```markdown
| EnablePeerExchange | Requires EnablePeerScoring | Service fails to start if PX on with scoring off | During server initialization |
```

### ChiR14 — `docs/p2p-libp2p-review.md` states two different pinned library versions

**Problem:** Line 3 still reads "Library version pinned in `go.mod`: `github.com/bsv-blockchain/go-p2p-message-bus v0.1.15`", including the module-cache path at that version, while line 30 — the paragraph this PR rewrote — says v0.1.23. The document now contradicts itself about the one fact it opens with. Its `file:line` anchors into `Server.go` went stale in the same commit: line 10 points at `Server.go:332-335` for the config pass-through, which this diff moved to `236-251`, and line 11's `322-324` and `288-291` for silent mode are now `534-537` and `494-511`. That anchor drift is pre-existing rather than introduced here, and line 5 promises "concrete `file:line` references", so a +141-line shift in the file it references is worth a pass.

**Why it matters:** Small, but this file is the security reference someone will read to decide what the current posture is, and a stale version header at the top is the fastest way to have the rest of it distrusted.

**Fix:** Update line 3 to v0.1.23 and refresh the `Server.go` anchors in section 1 (and the resolved-blocker note at line 125 which points at `Server.go:334`).

### ChiR15 — Two of the three descriptions of the colocation whitelist still say loopback and RFC1918 only

**Problem:** `ChiR4` widened the whitelist and the code took it: `Server.go:201-207` now covers `127.0.0.0/8`, `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `100.64.0.0/10`, `169.254.0.0/16`, `::1/128`, `fc00::/7` and `fe80::/10`. Two of its operator-facing descriptions did not move with it. `settings/p2p_settings.go:94` still says "When p2p_allow_private_ips is true, loopback and RFC1918 ranges are whitelisted", `:102` narrows it the same way ("loopback/private ranges"), and `settings.conf:906-907` says "loopback/private ranges are exempt". The settings reference is the one that was updated — `docs/references/settings/services/p2p_settings.md:71` correctly lists loopback, RFC1918, RFC6598, link-local and IPv6 ULA — which is what makes the other two look like an oversight rather than deliberate shorthand. This is the mirror image of the usual documentation defect: the code is *more* permissive than its description, not less.

**Why it matters:** The exposed reader is an operator on an IPv6-only or CNI-pod-range cluster (`fc00::/7`, `100.64.0.0/10`) who reads the settings struct's longdesc or `settings.conf`, concludes their range is not exempt, and disables scoring to get their cluster meshing — giving up the behaviour penalty and the PX gate to solve a problem they never had. That is the exact regression this PR exists to prevent, reached by reading the docs it ships.

**Fix:** Bring both in line with the code and with the reference doc that already has it right:

```
  settings.conf:906-907
  # exempt; with p2p_allow_private_ips=true, loopback, RFC1918, RFC6598 shared
  # space, link-local and IPv6 ULA ranges are exempt from the colocation penalty.
```

and the same list in the `EnablePeerScoring` longdesc at `settings/p2p_settings.go:94` and the `PeerScoreIPColocationThreshold` longdesc at `:102`. `TestPrivateIPColocationWhitelist` already pins the nine ranges, so the code side needs nothing.

## Recap

| ChiR9 | Residual omits connection budget | | 48% |
| ChiR10 | PX emission on, acceptance refused | | 44% |
| ChiR11 | Colocation doc versus 6h retention | | 38% |
| ChiR12 | PX threshold pins unpinned invariant | | 34% |
| ChiR13 | Validation table missing new error | | 18% |
| ChiR14 | Review doc pins two versions | | 12% |
| ChiR15 | Whitelist described too narrowly | | 10% |

PR #1599, opened 2026-08-17
fix(p2p/blockchain): stop IsConnected from pinning peer-registry entries forever

gokutheengineer:
**Closes bitcoin-sv/teranode#4741.**

### Problem

Every accepted gossip message flagged its pubsub author `IsConnected=true` in the centralized peer registry (`updatePeerLastMessageTime` → `addConnectedPeer`), and nothing ever cleared the flag except the ban path (`removePeer`, whose only caller is `disconnectBannedPeerByID`) — go-p2p-message-bus exposes no libp2p disconnect callback. Since `isCleanupExempt` exempted connected peers from both TTL expiry and LRU eviction, every peer ID that ever gossiped was pinned in the registry forever: `p2p_peer_registry_max_size` and `p2p_peer_registry_ttl` could never evict attacker-created entries, the LRU target collapsed to 0, and only honest idle peers were evicted. An attacker relaying self-signed messages under fresh peer IDs (no direct connection needed — `FromID` is the pubsub author) grew the registry without bound, amplifying into `List` clones under the registry write lock and monotonic `connected_peers_count` / `GetPeers` output.

All issue items were confirmed still valid against main.

### Fix

Three layers, so the invariant holds even if one fails:

- **Liveness gate on the hot path**: `updatePeerLastMessageTime` now marks a sender connected only when it has a live libp2p connection. Liveness is answered from a per-peer cache (`liveConnCache`, `reputationCacheTTL` = 5s, swept by the existing peer-map cleanup) populated by `isPeerIPBanned`'s existing `GetPeers()` walk — which every gossip handler already runs for the same peer immediately beforehand — with a targeted walk on a miss (`Addrs` is populated from the host's open connections, the only connectedness signal the library exposes). A freshly connected neighbour is therefore flagged on its very first message — same behaviour as main, and well inside `daemon.TestDaemon.ConnectToPeer`'s 15s budget — while gossip-relayed publishers are registered as gossiped peers and stay subject to TTL/LRU, closing the sustained-flood vector, not just the connect-and-leave one.
- **Bidirectional reconciliation sweep**: `reconcileConnectionStates` runs on the existing peer-map cleanup tick (own goroutine, in-flight guard, 30s pass timeout so a slow registry cannot stall the cache sweep or wedge RPCs): clears `IsConnected` on entries with no live connection and sets it on live peers the hot path missed (messages that arrived before the snapshot included them). A new batcher method `forgetAssertState` drops only the reassert memory so a reconnecting peer is re-flagged on its next message; when the forget races an in-flight `flushOnce`, the cycle's re-record drops the whole assert snapshot — including the halves that cycle sent — because the batcher cannot tell whether the reconciler's clear landed before or after its own write, and keeping a "sent" half would mask the clear for up to `registryReassertTTL`. The reconcile pass also seeds the per-peer liveness cache (live peers true, cleared peers false) so the hot path cannot re-flag a just-cleared peer off a stale cache entry.
- **Soft exemption in Cleanup**: `isCleanupExempt` exempts connected peers only while `peerActivity` is within the TTL (banned peers stay exempt), so even a stale flag cannot pin an entry forever. `Load` additionally resets `IsConnected=false` on restore — a connection cannot survive a restart, and the registry must not depend on the p2p service's sweep for that invariant.

Improvements over the issue's proposed fix: a libp2p disconnect callback is impossible (the library exposes no host — see `networkDisconnector`), and a clear-only periodic diff was shown insufficient (a continuously gossiping flood re-flags itself between sweeps); gating the `true` write on liveness closes that and makes `connected_peers_count`, `GetPeers`, `getpeerinfo`, and the dashboard counters reflect the true neighbour set instead of accumulating monotonically. Shutdown now stops the cleanup ticker before closing the P2P client so a final tick cannot mass-clear flags against a closed host.

### Review notes

On the `Addrs`-as-liveness question: verified against go-p2p-message-bus v0.1.17 source. `GetPeers()` iterates the peer tracker's known topic peers — every peer that ever authored a message on a subscribed topic, never pruned, so the listing itself is NOT connected-only — and builds `Addrs` from `host.Network().ConnsToPeer`, i.e. open connections only, not the retained peerstore. Empty `Addrs` therefore means no live connection, and the filter is essential (it is what separates live neighbours from gossip-only authors), not redundant. `services/p2p/Client.go:117` is not inconsistent: that is the gRPC client of the p2p service's own `GetPeers` RPC, whose server side filters to `IsConnected` registry entries (`Server.go`), so marking everything it returns as connected matches that RPC's contract. The distinction is now documented on `refreshLiveConnIDs`.

### Residual

- **The unbounded-growth vector is relocated upstream, not closed.** go-p2p-message-bus's `peerTracker.topicPeers` records every message author and is never pruned — the same attacker-controlled set the registry was accumulating, one layer down. Teranode's registry now stays capped under a sustained flood (the fix works), and the walks this change adds match the cost profile the gossip path already pays (`isPeerIPBanned` walks `GetPeers()` per peer per 5s today; the reconcile snapshot adds one walk per cleanup interval) — but the underlying set still grows without bound. Needs an upstream fix: a pruned/bounded tracker, plus a `Connectedness(peer.ID)` accessor (thin wrapper over `host.Network().Connectedness`) that would make the liveness check exact and O(1), remove the walks entirely, and also close the connected-but-silent-peer caveat.
- **Under sustained flood at the size cap, LRU evicts the honest peers first.** The LRU phase orders by activity and a flood keeps its spoofed IDs continuously fresh, so under size pressure the idle honest peers — the reputation and ban history worth keeping — are what get evicted. Not introduced here (LRU always ordered this way), but this PR is what makes the LRU phase actually bite.
- `ListPeers` still has no pagination and reconcile reads the full list once a minute; with the registry now actually bounded by `p2p_peer_registry_max_size` the payload is capped, but an IDs-only / connected-only RPC would be a worthwhile follow-up.
- Peers holding many real open connections can still keep entries exempt while fresh; bounding that is the connection-limits work tracked separately (issue 1163).
- The reconciler is now the first component to reliably notice a disconnect, but it does not inform the sync coordinator. A narrow follow-up could clear and reselect the sync peer when its flag is cleared — a latency win only, since catchup RPCs to a dead sync peer already fail into reselection.

### Tests

- `TestServerHelpers_ReconcileConnectionStates_SyncsBothDirections` — clears gone/unknown peers, keeps live ones, flags live peers the hot path missed.
- `TestServerHelpers_GossipOnlyPublisherNeverFlaggedConnected` — through the real batcher path: relayed publishers are never flagged and do not flap back after a reconcile pass.
- `TestServerHelpers_NewNeighbourFlaggedOnFirstMessage` — the production ordering (service running, then a peer connects and gossips): the neighbour is flagged connected immediately on its first message, not one cleanup interval later. This is the case the review's reproduction caught; it fails on the snapshot-gate design and passes now.
- `TestPeerRegistryBatcher_ForgetAssertStateForcesReassert`, `TestPeerRegistryBatcher_ForgetAssertStateDuringFlushNotResurrected`, and `TestPeerRegistryBatcher_ForgetDuringConnectOnlyFlushZerosSnapshot` — reassert semantics, including the mid-flush race in both send configurations.
- `TestPeerRegistryBatcher_ReconnectAfterMaskedClearIsReflagged` — the round-4 blocker's observable consequence: a reconciler clear landing after the flush's connect assert must not leave a reconnecting peer invisible for a reassert TTL. Verified to fail on the pre-fix batcher.
- `TestCentralizedPeerRegistry_Cleanup_EvictsStaleConnected` — stale-connected entries are TTL-evicted; fresh-connected and banned stay exempt.
- `TestCentralizedPeerRegistry_Persistence_LoadResetsIsConnected` — no phantom connections after restart.
- `TestServerHelpers_StartPeerMapCleanup_RunsReconcile` — the ticker wiring end-to-end: a stale flag is cleared by the ticker-driven reconcile through a real batcher.
- `TestServerHelpers_ReconcileConnectionStates_ErrorAndGuardPaths` — nil guards, `ListPeers` failure, per-direction `UpdateConnectionState` failures, and the canceled-context cut-short path.
- `TestPeerRegistryBatcher_ForgetDuringConnectOnlyFlushZerosRegisteredAt` — the other half of the mid-flush race (register kept, connect sent).

Ran (green):

```
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/blockchain/... ./services/p2p/... ./settings/...
```


liam (approved):
Approving. I checked the two hazards the `isCleanupExempt` change could have introduced and both are fine; the remaining points are a cost concern and a comment that claims slightly more than the code delivers.

**Verified:** all three layers are present and coherent. There is no zero-activity connected peer that the new exemption would wrongly evict, because `Register` sets `LastSeen = now` for new peers (`peer_registry.go:285`). And the `addPeer` fallback in `updatePeerLastMessageTime` cannot clobber richer data, because `Register` only updates non-zero fields on existing entries. Builds clean, targeted `p2p` and `blockchain` tests pass, and the coverage is genuinely good — both reconcile directions, the ticker path, the guard paths, and the gossip-only-publisher case.

## The reconcile is most expensive exactly when it matters most

`peerRegistry` is a gRPC client to the blockchain service (`Server.go:152`), and `reconcileConnectionStates` calls:

```go
```

That is the **entire** registry — up to `p2p_peer_registry_max_size` (default 10,000) `PeerRegistryInfo` messages, 38 fields and 9 timestamps each — cloned under the registry lock, marshalled, shipped and unmarshalled every `p2p_peer_map_cleanup_interval` (default **1 minute**). Then one **serial** `UpdateConnectionState` round trip per changed peer.

Under the sustained flood this PR exists to defend against, the registry sits at max size and the stale-flag count is highest, so the sweep is at its most expensive precisely when it is most needed — and the 30s timeout then cuts passes short, slowing convergence. There is no hard cliff (`MaxMessageSize` defaults to 1GB in `util/grpc_helper.go:117`), so this is cost rather than breakage.

It is also avoidable: the clear direction only needs entries with `IsConnected=true`, and the flag direction only needs the `live` set, which is bounded by the actual neighbour count. A connected-only filter on `ListPeers`, or a bulk `SetConnectionStates` RPC, collapses both the payload and the round trips.

## The `Stop()` reorder does not cover the in-flight pass

Moving the ticker stop above `P2PClient.Close()` prevents *new* ticks, but a reconcile goroutine spawned by the previous tick can still be mid-flight, and `refreshLiveConnIDs()` runs before the ctx-timeout wrapper without checking ctx — so it can still call `GetPeers()` on a closing client and store an empty live set. Impact is small (shutdown only, and `Load` resets `IsConnected` on restart regardless), but the comment claims more than the code delivers. Waiting on `reconcileInFlight`, or an early ctx check in the refresh, would make them match.

## Worth stating the guarantee precisely

`Register` bumps `LastSeen` on every gossip message, so a sustained flooder's entries stay newest and LRU evicts honest idle peers first. The fix converts *unbounded* growth into *bounded but attacker-dominated* occupancy — a clear improvement, with reputation-weighted eviction as the natural follow-up. You were careful to state the limits honestly on the peer-scoring PR; the same paragraph would serve here.

services/p2p/server_helpers.go
```diff
@@ -1139,13 +1143,122 @@ func (s *Server) startPeerMapCleanup(ctx context.Context) {
...
+// host.Network().ConnsToPeer — open connections only, not the peerstore —
+// while the peer list itself is every peer that ever authored a message on a
+// subscribed topic (gossip-only publishers included, never pruned). So the
+// Addrs filter is what separates live neighbours from gossip-only authors;
+// it is not redundant with the listing. This differs from the p2p service's
+// own GetPeers RPC, which is filtered to IsConnected registry entries.
+func (s *Server) refreshLiveConnIDs() map[string]struct{} {
+	live := make(map[string]struct{})
+	if s.P2PClient != nil {
+		for _, p := range s.P2PClient.GetPeers() {
```
freemans13, line 1249: The registry is bounded now, but the attacker's memory doesn't stop accumulating — it just accumulates one layer down, and this sweep is what now walks it on a timer.

I grepped the whole of go-p2p-message-bus v0.1.17 for `delete(` and got zero hits. `peerTracker` has six per-peer maps — `names`, `topicPeers`, `lastSeen`, `malformed`, `skipMalformed`, `isRelaying` — all keyed on the gossip author, all insert-only, never pruned. The same flood of self-signed messages under fresh peer IDs still grows those forever.

That's the library's bug, not this PR's, and a tracker entry is much smaller than a `PeerInfo`, so bounding the registry is still the bigger win. Two things follow though. Teranode previously only touched that map on demand in the `GetPeers` RPC handler; this now allocates a `[]PeerInfo` sized to the entire tracker and does a `ConnsToPeer` call per entry every minute, so under exactly the flood being defended against, the defensive sweep is the thing scanning the attacker's data structure on a fixed schedule. And the Residual section reads as if 4741 is fully closed when the underlying accumulation is still live.

Worth a line in Residual plus a follow-up issue against the library. Separately, `Server.go:2093` in the `GetPeers` RPC handler already does the same `GetPeers()` + `len(Addrs) > 0` walk — it could reuse this snapshot and save a second pass over that map.

services/p2p/peer_registry_batcher.go
```diff
@@ -454,6 +480,17 @@ func (b *peerRegistryBatcher) flushOnce(ctx context.Context) {
 			// above, and recording the assertion would suppress the peer's
 			// re-registration for registryReassertTTL after its next message.
 			if !b.isRemovedLocked(peerID) {
+				// A forgetAssertState() may also have raced the RPCs: keep
+				// only the assertions this cycle actually sent so the
+				// forgotten (stale) part of the snapshot is not resurrected.
+				if _, forgotten := b.assertForgottenDuringFlush[peerID]; forgotten {
```
freemans13, line 493: Both branches here are guarded on `!sendRegister` / `!sendConnected`, so when the cycle sent both there's a third case that falls through untouched: a `forgetAssertState` racing it is discarded entirely. The registry ends at `IsConnected=false` (the reconciler wrote it) while the batcher believes it asserted connected zero seconds ago, so it won't re-assert for `registryReassertTTL`.

It self-heals — reconcile is bidirectional, so the next pass sets the flag back — so this isn't a bug. But the doc comment on `forgetAssertState` says the flush-scoped set "stops an in-flight `flushOnce` from resurrecting the pre-forget snapshot", and in the both-sent case it doesn't. `ForgetAssertStateDuringFlushNotResurrected` and `ForgetDuringConnectOnlyFlushZerosRegisteredAt` cover the two single-send halves; both-sent is the uncovered one.

Either zero unconditionally, or reword the comment to say the guarantee holds per-assertion and that the reconcile pass is what closes the rest.

services/p2p/Server.go
```diff
@@ -1996,6 +2011,14 @@ func (s *Server) Stop(ctx context.Context) error {
 		coordCancel()
 	}
 
+	// Stop the peer map cleanup ticker before closing the P2P client: a
+	// reconcile tick against a closed client sees zero live connections and
+	// would clear IsConnected on every registry peer during shutdown.
```
freemans13: The reorder stops a *new* tick, but a pass already dispatched on its own goroutine under `reconcileInFlight` is never waited for, so it can still call `GetPeers()` after `Close()`, see zero live connections, and mass-clear every flag — the outcome this comment says the reorder prevents.

The outcome is harmless, but for two other reasons: `ConnsToPeer` on a closed swarm returns empty rather than panicking, and the new `Load` reset zeroes `IsConnected` on restore anyway. So `Load` is the real backstop here, not the reorder. Either wait on `reconcileInFlight` in `Stop`, or say that in the comment so a later change to `Load` doesn't quietly remove the protection.

freemans13 (commented):
Reviewed the three branch commits against `upstream/main`. Ran `SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./services/p2p/... ./services/blockchain/... ./settings/...` (exit 0) and `make lint` (0 issues), both green.

I also checked the load-bearing claim independently, since the whole fix rests on it: in go-p2p-message-bus v0.1.17, `GetPeers()` iterates `peerTracker.getAllTopicPeers()` — every peer that ever authored on a subscribed topic — but builds `Addrs` from `host.Network().ConnsToPeer`, so open connections only. Empty `Addrs` really does mean no live connection, and the filter really is what separates neighbours from gossip-only authors. The claim holds.

One thing worth confirming for anyone else reading: gating the hot path on a snapshot refreshed once per `p2p_peer_map_cleanup_interval` means a genuinely new neighbour isn't flagged connected for up to a minute. I traced every `IsConnected` consumer — the node-status connected count (`Server.go:1631`), the `GetPeers` RPC filter (`Server.go:2099`, feeding `getpeerinfo` and the dashboard), and `asset/httpimpl/get_peers.go`. The sync coordinator does not read it. So that lag is reporting-only and won't delay sync-peer selection on cold start.

Nothing blocking. Three comments inline: one substantive (the growth moves down a layer rather than going away), two where a code comment claims more than the code delivers.

freemans13 (commented):
Reviewed the full diff plus surrounding context. **The fix is correct and the design is sound — I found no blocking defects.** Two things I'd want addressed before merge are Residual-section edits rather than code changes.

### What I ran

| Check | Result |
|---|---|
| `make lint` (golangci-lint, `--new-from-rev origin/main`) | 0 issues |
| `go vet ./services/p2p/... ./services/blockchain/...` | clean |
| New p2p tests, `-race -count=1 -tags testtxmetacache` | pass |
| New blockchain registry tests, `-race -count=1` | pass |

### Claims I checked independently

The design rests on several load-bearing claims about `go-p2p-message-bus`, so I checked them against the v0.1.17 source rather than taking them on trust. They all hold:

- `GetPeers()` really is a valid liveness signal and the `Addrs` filter really is essential — `client.go:663` iterates `peerTracker.getAllTopicPeers()` (every peer that ever authored a message on a subscribed topic), and builds `Addrs` from `c.host.Network().ConnsToPeer` at `client.go:667`. Non-empty `Addrs` means an open connection; the listing itself is not connected-only.
- `FromID` really is the pubsub author, not the immediate hop — `client.go:1086` takes `msg.GetFrom()` and passes it straight through at `client.go:1114`. The premise of the vulnerability holds.
- The `services/p2p/Client.go:117` "inconsistency" genuinely isn't one — that's the gRPC client of the p2p service's own `GetPeers` RPC, and the server side does filter to `IsConnected` entries (`Server.go:2099`).
- `addConnectedPeer` has exactly one production caller (`Server.go:1088`), so the new liveness gate really is the only place `IsConnected` is set true besides the reconciler.
- The new `addPeer` fallback can't clear a flag — `pendingPeerUpdate.merge` treats `markConnected` as sticky-true-only. That's what makes the one-minute snapshot lag benign, and it's worth knowing.
- The `Cleanup` LRU arithmetic still holds. After phase 1 every surviving connected peer is by construction fresh, so it's still LRU-exempt and `target` can't go negative. `TestCentralizedPeerRegistry_Cleanup_LRUExemptsConnectedAndBanned` remaining valid unchanged is a good sign the change is additive rather than a behaviour swap.
- The `Stop` reorder is safe — nothing between the ticker's old and new position depends on the ticker.

I also traced the `assertForgottenDuringFlush` handling carefully, since it's the fiddliest part. The set is created inside the same `b.mu` critical section as the pending swap and `st` is read under `b.mu` later, so a `forgetAssertState` landing before the read gives a zero `st` (both assertions resent) and one landing after is caught by the flush-scoped set; the `sendRegister || sendConnected` guard skips the re-record entirely when nothing was sent. I couldn't find a hole.

---

### Worth addressing before merge

**1. The unbounded-growth vector is relocated rather than closed, and this PR now walks the unbounded side every 60s.**

`peerTracker.topicPeers` (`types.go:73`) is inserted into by `recordMessageFrom` on every message author (`types.go:131`, called from `client.go:1107`) and **nothing ever deletes from it** — I grepped the whole module, there's no prune, bound or expiry. It's the same attacker-controlled set the registry was accumulating, one layer down.

`GetPeers()` walks that entire map and allocates a fresh `[]PeerInfo` over it. Before this PR that only happened on request (the `GetPeers` RPC handler, `getNodeStatusMessage`); now `refreshLiveConnIDs` does it unconditionally every `p2p_peer_map_cleanup_interval`. So under exactly the sustained flood this fixes, teranode's registry stays capped — which is the fix, and it works — but the library-side map grows without bound and the periodic sweep becomes an O(spoofed-IDs) walk plus a full-set allocation on a fixed cadence.

Not a blocker, and the vector is genuinely much narrower after this change than before. But Residual currently reads as though growth is closed. Worth saying it has moved, and filing an upstream issue for either a pruned tracker or a cheap connected-IDs accessor that doesn't materialise the whole set.

**2. LRU now evicts the honest peers in favour of the flood.**

The registry is bounded now, but `Cleanup`'s LRU phase orders by `peerActivity` and evicts oldest-first. A sustained flood keeps its spoofed IDs continuously fresh, so under size pressure the entries that actually get evicted are the idle honest peers — precisely the reputation and ban history worth keeping.

Not introduced here (the LRU always had this property), but this PR is what makes the LRU phase actually bite, so it's the right place to name it. Sits naturally next to the connection-limits note in Residual.

---

### Smaller points

**3. The ticker-stop reorder is best-effort; the comment claims more than the code guarantees.** `Server.go:2016-2020` says stopping the ticker before closing the client prevents a reconcile tick from mass-clearing flags at shutdown. `ticker.Stop()` stops future ticks — it doesn't cancel a goroutine the last tick already launched, and nothing waits on `reconcileInFlight`. The race is narrowed, not closed. It's harmless in practice, but for a reason the comment doesn't give: `Load` now resets `IsConnected = false` anyway, so a shutdown mass-clear has no persistent consequence. Either say that (cheapest, and it's the real argument) or drain the goroutine in `Stop`.

**4. `connected_peers_count` means something slightly narrower than "the true neighbour set".** Liveness is `getAllTopicPeers ∩ open connections`, and the topic-peer set is authorship-based. So a genuinely connected peer that has never authored a gossip message since process start is invisible and stays unflagged indefinitely, and for up to one cleanup interval after a peer connects its messages register it as gossip-only. Both self-heal for real teranode peers via the `node_status` heartbeat, so this is an accuracy caveat rather than a defect — the counters are vastly more truthful than the monotonic accumulation they replace. But "peers with an open connection that have gossiped at least once since startup" is the phrasing an operator debugging a peer-count discrepancy would want to read.

**5. Doc drift on `p2p_peer_map_cleanup_interval`.** `settings/p2p_settings.go:53` still describes the tick as only scanning `blockPeerMap` and `subtreePeerMap`. It now also drives the connection-state reconcile, which makes it the knob setting how long a stale `IsConnected` flag can persist. The two registry settings were updated; this one was missed.

**6. One test can't distinguish the branch it claims to cover.** In `TestServerHelpers_ReconcileConnectionStates_ErrorAndGuardPaths`, the canceled-context case asserts only that flags are untouched — satisfied identically by the in-loop `ctx.Err()` cut-short branch and by the `ListPeers` error branch. Today the local registry client ignores context so the intended branch does run, but if that ever changes the test keeps passing while covering something else. Asserting `counting.callCount("ListPeers") == 1` would pin it.

**7. The clear-then-reassert window converges — worth writing down.** If a `markConnected` was enqueued from a message that arrived while the peer was still in the old snapshot, and reconcile clears the flag before that batch flushes, the flush re-asserts `true` — and because `forgetAssertState` just zeroed `connectedAt`, that assert is *guaranteed* to be sent rather than skipped as recently-asserted. Window is at most one batch interval, corrected by the next reconcile pass. So it converges to the right answer, but that reasoning isn't written anywhere and it looks like a bug to the next reader of `forgetAssertState`.

**8. Optional follow-up: the sync coordinator isn't told about detected disconnects.** The reconciler is now the first thing in the service that reliably notices a peer has gone. Calling `HandlePeerDisconnected` outright would be wrong — it also calls `registryRemovePeer`, and a mere disconnect shouldn't destroy a reputation record. The narrower move is clearing the sync peer and re-triggering selection when the cleared peer *is* the current sync peer. Even that is only a latency win, since catchup RPCs to a dead sync peer fail into `HandleCatchupFailure` and reselect. Out of scope here — noting it because this PR creates the signal that would make it possible.

services/blockchain/peer_registry.go
```diff
@@ -1378,9 +1379,20 @@ func (r *CentralizedPeerRegistry) Cleanup(maxSize int, ttl time.Duration) (int,
 }
 
 // isCleanupExempt reports whether a peer must be retained regardless of TTL or
-// size pressure. Caller must hold the registry lock.
-func isCleanupExempt(info *PeerInfo) bool {
-	return info.IsConnected || info.IsBanned
+// size pressure. Banned peers are always retained. IsConnected only exempts a
+// peer whose activity is still within ttl: the flag is asserted by the p2p
+// service on every gossip message but there is no libp2p disconnect signal to
```
freemans13, line 1384: **[suggestion]** `isCleanupExempt` still says the p2p service asserts `IsConnected` “on every gossip message.” After this PR that is false: `updatePeerLastMessageTime` only calls `addConnectedPeer` when `hasLiveConnection` is true, and gossip-relayed publishers go through `addPeer`. The outdated sentence makes layer 3 look like the only flood defence and hides that the hot-path gate is what stops a self-signed-ID flood from ever becoming cleanup-exempt.

**Suggestion:** Reword to the actual write path, e.g. the flag is set for live libp2p neighbours (hot path plus reconcile) and there is no disconnect callback to clear it, so a stale flag must not pin an entry past `ttl`.

services/p2p/server_helpers.go
```diff
@@ -1139,13 +1143,122 @@ func (s *Server) startPeerMapCleanup(ctx context.Context) {
...
+}
+
+// reconcileConnectionStates synchronizes the registry's IsConnected flags with
+// actual libp2p connectedness, in both directions: it clears the flag on
+// entries with no live connection and sets it on live peers the hot path
+// missed (their messages arrived before the liveness snapshot included them).
+// Nothing else ever clears the flag: go-p2p-message-bus exposes no disconnect
+// callback (see networkDisconnector) and the only other clearing site is the
+// ban path (removePeer), so without this sweep every flagged peer would stay
+// cleanup-exempt in the registry for the life of the process.
```
freemans13, line 1311: **[suggestion]** The new `reconcileConnectionStates` comment says that without this sweep “every flagged peer would stay cleanup-exempt in the registry for the life of the process.” That was true when `isCleanupExempt` was `IsConnected || IsBanned`. This PR’s layer 3 already evicts connected peers once `peerActivity` exceeds TTL, so an idle connect-and-leave entry is not pinned for process lifetime; the sweep is what keeps `GetPeers` / `connected_peers_count` honest and what clears flags on still-active gossip-only authors. As written, the comment contradicts `isCleanupExempt` two packages over.

**Suggestion:** Say the sweep is the only thing that clears `IsConnected` besides `removePeer`, so without it the flag (and therefore GetPeers / connected counts) would stay true for the process lifetime; idle entries are still TTL-evictable via the soft exemption.

services/p2p/server_helpers_test.go
```diff
@@ -69,6 +70,160 @@ func mustNewPeerID(t *testing.T) peer.ID {
...
+	stale := mustNewPeerID(t)
+	s.addConnectedPeer(stale, "", 0, nil, "")
+
+	// Known topic peer with no open connection: the ticker-driven reconcile
+	// must clear its stale connected flag.
+	s.P2PClient = &MockServerP2PClient{peers: []p2pMessageBus.PeerInfo{{ID: stale.String()}}}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	s.startPeerMapCleanup(ctx)
```
freemans13, line 130: **[nit]** `TestServerHelpers_StartPeerMapCleanup_RunsReconcile` starts a 10ms `time.Ticker` via `startPeerMapCleanup` and only cancels the context. `cancel()` makes the loop return, but nothing calls `peerMapCleanupTicker.Stop()`, so the ticker keeps firing until process exit.

**Suggestion:** `t.Cleanup` that stops the ticker (and optionally waits for `reconcileInFlight` to clear) after `Eventually` succeeds.

services/p2p/server_helpers.go
```diff
@@ -1131,6 +1131,10 @@ func (s *Server) startPeerMapCleanup(ctx context.Context) {
 
 	s.peerMapCleanupTicker = time.NewTicker(cleanupInterval)
 
+	// Seed the liveness snapshot so the gossip hot path has one before the
+	// first reconcile tick.
+	s.refreshLiveConnIDs()
```
freemans13: This seed is the only refresh the gossip path gets until the first reconcile tick, and that is what creates the timing problem I've described in the review body.

At the moment this runs, a freshly started node has no peers connected yet, so the snapshot is empty. The next write to `liveConnIDs` is a full `p2p_peer_map_cleanup_interval` away — one minute by default, and not overridden anywhere in `settings.conf`. Every peer that connects in that window, and every peer that connects between any two later ticks, keeps being registered as a gossip-only publisher no matter how much it gossips.

I confirmed it by replaying that exact order in a test (empty seed, peer connects, peer gossips): it fails on this branch after 15 seconds and passes on `upstream/main`. Full test in the review body.

The concrete casualty is `daemon.TestDaemon.ConnectToPeer` (`daemon/test_daemon.go:2207`), which polls the `IsConnected`-filtered `GetPeers` RPC and gives up after 15 seconds. It only escapes CI because every caller in `test/e2e/daemon/ready` is inside a skipped test.

services/p2p/Server.go
```diff
@@ -1064,16 +1072,23 @@ func (s *Server) updatePeerLastMessageTime(from string, originatorPeerID string)
...
 		return
 	}
 
-	s.addConnectedPeer(senderID, "", 0, nil, "")
+	// Only mark the sender connected when it has a live libp2p connection.
+	// FromID is the pubsub author, not the immediate hop, so gossip-relayed
+	// messages arrive "from" peers we never dialled; flagging those connected
+	// would exempt them from registry cleanup and let a flood of self-signed
+	// messages under fresh peer IDs grow the registry without bound.
+	if s.hasLiveConnection(from) {
```
freemans13, line 1141: This is where the timing problem is cheapest to fix, and cheaper than I implied last round.

By the time control reaches this line, `shouldSkipBannedPeer` has already run for this same `from` — every calling handler does it first (`Server.go:1199`, `server_helpers.go:65`, `:201`, `:479`) — and it calls `isPeerIPBanned`, which walks `s.P2PClient.GetPeers()` looking at `p.Addrs` for this exact peer, with only a 5-second per-peer cache. The liveness answer you want is already being computed in that loop, for this peer, seconds ago at worst.

So there are two cheap options and one clean one:

- Fold the liveness lookup into that existing walk (same peer, same `p.Addrs`) and cache it alongside the ban answer. Costs nothing new and cuts the lag from 60s to 5s, comfortably inside `ConnectToPeer`'s budget.
- Or give the liveness snapshot its own short ticker, independent of the reconcile cadence. Also cheap, for the same reason — the walk already happens far more often than once a minute.
- The clean long-term fix is upstream: a `Connectedness(peer.ID)` accessor on go-p2p-message-bus, a thin wrapper over `host.Network().Connectedness`. That would make this check exact and O(1), remove the snapshot entirely, and also close the "a connected peer that has never gossiped is invisible" caveat you documented above. Worth filing alongside the pruned-tracker ask.

Whichever way you go, the ordering case needs a test — see my note on `server_helpers_test.go`.

services/p2p/server_helpers_test.go
```diff
@@ -69,6 +70,160 @@ func mustNewPeerID(t *testing.T) peer.ID {
...
+	neighbourID := mustNewPeerID(t)
+	publisherID := mustNewPeerID(t)
+
+	// Only the neighbour has an open connection; the publisher's messages
+	// arrive relayed through the mesh (FromID is the pubsub author).
+	s.P2PClient = &MockServerP2PClient{peers: []p2pMessageBus.PeerInfo{
+		{ID: neighbourID.String(), Addrs: []string{"/ip4/10.0.0.1/tcp/9905"}},
+		{ID: publisherID.String()},
+	}}
+	s.refreshLiveConnIDs()
```
freemans13: This line is why the timing problem got through both of my earlier rounds: refreshing the snapshot *before* the gossip means this test only ever covers "the snapshot already knows this peer".

That's a real case and worth keeping. But the case that happens every time in production is the opposite ordering — the snapshot is taken, and *then* a peer connects and starts gossiping. There's no test for it anywhere in the file, which is exactly why nothing went red.

Worth adding a sibling test that gossips first and only then makes the peer live in the mock client, asserting the peer becomes visible within a bound you're happy to promise. That pins the behaviour whichever way you fix the timing, and it would have caught this.

freemans13 (commented):
Re-reviewed after `b9acd9eb9`. Everything I raised last round is addressed properly — the doc drift on `p2p_peer_map_cleanup_interval`, both code comments that claimed more than the code delivered, the test that couldn't tell its own branches apart, and the two Residual additions. Good, careful responses.

Reading the whole change again rather than just the delta, though, I found one thing neither of my earlier rounds caught, and I owe you a correction on something I got wrong.

### What I ran

| Check | Result |
|---|---|
| `make lint` | 0 issues |
| `go test -race -count=1 -tags testtxmetacache ./services/p2p/... ./services/blockchain/... ./settings/...` | all ok |
| Reproduction test below, on this branch | **FAIL** |
| Same test on `upstream/main` | pass (0.85s) |

---

## A genuinely connected neighbour stays invisible for up to a minute

`liveConnIDs` is written in exactly two places: the seed in `startPeerMapCleanup`, and one refresh per reconcile tick — `p2p_peer_map_cleanup_interval`, default one minute, and not overridden anywhere in `settings.conf`.

That means the real timeline for every peer that connects to an already-running node is: the node seeded an empty snapshot at startup, the peer dials in and starts gossiping, and the snapshot still doesn't know it exists. Its messages register it as a gossip-only publisher. Nothing flags it connected until the next tick, up to 60 seconds later.

I didn't want to assert that from reading, so I replayed the real ordering — empty seed, then the peer connects with an open connection, then it gossips — and ran it on both sides:

```go
func TestReview_NewNeighbourFlaggedBeforeFirstTick(t *testing.T) {
	s, reg := newServerWithLocalRegistry(t)
	s.settings.P2P.PeerMapCleanupInterval = time.Minute // production default
	s.registryBatcher = newPeerRegistryBatcher(context.Background(), s.logger, s.peerRegistry, 0)

	// Service starts with no peers connected yet: the seed snapshot is empty.
	client := &MockServerP2PClient{peers: []p2pMessageBus.PeerInfo{}}
	s.P2PClient = client

	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()
	s.startPeerMapCleanup(ctx)

	// Now a peer dials in: the host has an open connection to it, and it
	// authors a gossip message (node_status heartbeat).
	neighbour := mustNewPeerID(t)
	client.peers = []p2pMessageBus.PeerInfo{
		{ID: neighbour.String(), Addrs: []string{"/ip4/10.0.0.9/tcp/9905"}},
	}
	s.updatePeerLastMessageTime(neighbour.String(), "")

	// ConnectToPeer's budget is 15s; poll well inside it.
		got, ok := reg.Get(neighbour.String())
		return ok && got.IsConnected
	}, 15*time.Second, 100*time.Millisecond,
		"a genuinely connected, gossiping neighbour must be visible to the GetPeers RPC inside ConnectToPeer's 15s budget")
}
```

On this branch it fails after the full 15 seconds. On `upstream/main` it passes in under a second, because the old code marked every sender connected unconditionally.

**Why 15 seconds is the number that matters:** `daemon.TestDaemon.ConnectToPeer` (`daemon/test_daemon.go:2207`) polls the `GetPeers` RPC — which filters on `IsConnected` (`Server.go:2109`) — and calls `t.FailNow()` after 15 seconds. Sixty is bigger than fifteen, so that helper can no longer succeed against a freshly started node.

CI is not red today, and it took me a while to work out why: every `ConnectToPeer` call site in `test/e2e/daemon/ready` sits inside a `t.Skip`'d test, and `make smoketest` only runs that one directory. The live call sites are in `test/e2e/daemon/bsv` and `test/e2e/daemon/wip`, which CI doesn't run. So this is latent rather than currently failing — but any new multi-node E2E test written with that helper will fail, which is a nasty thing for the next person to debug.

The production-side consequence is narrower but real: `getpeerinfo`, the dashboard peer list, `connected_peers_count` and `cmd/diagnose`'s peer-count health check all read zero connected peers for the first minute after startup, and every subsequent new neighbour is missing for up to a minute after it connects.

**Why the current tests don't catch it:** `TestServerHelpers_GossipOnlyPublisherNeverFlaggedConnected` calls `s.refreshLiveConnIDs()` *before* gossiping, so it only ever exercises "the snapshot already knows this peer". Connect-after-snapshot — which is what happens every single time in production — has no coverage. That's the gap that let this through, and it's worth closing whichever way you fix the timing.

---

## Correction: my cost objection from last round was wrong-headed

Last round I said this change now walks the unbounded `topicPeers` map every 60 seconds, and I framed that as a new cost. That framing was wrong, and it went into your Residual section on my prompting, so it's worth putting straight.

`shouldSkipBannedPeer` → `isPeerIPBanned` (`services/p2p/server_helpers.go:1006`) already does a full `GetPeers()` walk on the gossip accept path, cached for only 5 seconds per peer. Every handler that reaches `updatePeerLastMessageTime` calls `shouldSkipBannedPeer` first (`Server.go:1199`, `server_helpers.go:65`, `:201`, `:479`), so every message that gets as far as the connected check has already had that exact walk done for that exact peer within the last 5 seconds. There's a debug ticker walking it every 30 seconds on top of that.

Against that, one refresh per minute adds nothing measurable. The upstream point still holds — `topicPeers` is genuinely never pruned and that's the real unbounded thing — but the "periodic O(spoofed-IDs) walk on a fixed cadence" line in Residual reads as a new cost this change introduces, and it isn't. Worth softening to keep the honest part (the growth moved down a layer) without the part that overstates what this change added.

That correction is also what makes the fix cheap, which I've written up inline on the hot path.

freemans13 (changes requested):
## Round 4 review — the round-3 blocker is genuinely fixed, one residual gap in the same mechanism

Reviewed at `3c2cac4be` against `upstream/main` (`0c1b78349`).

### The round-3 blocker is closed, and the new test really does catch it

I asked for the liveness lookup to be folded into the existing 5s-cached `isPeerIPBanned` walk rather than a once-per-cleanup-interval snapshot, and that is exactly what `3c2cac4be` does. I verified the new test is a real regression test rather than a restatement of the new code: I checked out the pre-fix commit `b9acd9eb9` in a scratch worktree, dropped `TestServerHelpers_NewNeighbourFlaggedOnFirstMessage` in unchanged (it compiles against the old snapshot design without modification), and it fails there:

```
--- FAIL: TestRound3Repro_NewNeighbourFlaggedOnFirstMessage (0.00s)
    Messages: a connected, gossiping neighbour must be visible immediately
```

It passes at `3c2cac4be`. So the connect-after-snapshot hole is closed and pinned.

I also want to close out the cost question properly, since the TTLs now line up in a way that makes it a non-issue. `isPeerIPBanned` writes `ipBanCache` and `liveConnCache` from the same `now` with the same `reputationCacheTTL`, so the two entries expire together: on the next message the ban walk refreshes both and `hasLiveConnection` reads a fresh cache. The fallback walk inside `hasLiveConnection` therefore fires only on paths that skip `shouldSkipBannedPeer` (the `isOwnMessage` short-circuit) or when `banList`/`P2PClient` is nil. One `GetPeers()` walk per peer per 5s, which is the profile `main` already pays. That is the right shape.

Full suite green here at HEAD:

```
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/p2p/... ./services/blockchain/...
ok  github.com/bsv-blockchain/teranode/services/p2p          16.752s
ok  github.com/bsv-blockchain/teranode/services/blockchain    39.922s
```

`golangci-lint run ./services/p2p/... ./services/blockchain/... ./settings/...` reports one `prealloc` hit in `services/p2p/mock.go:76`, which is not in this PR's diff — clean as far as this change goes.

### Blocker: `forgetAssertState` can be masked, which reopens the same 60s-invisibility window it exists to prevent

In `flushOnce`, the re-record step only zeroes the halves of the assert state that this cycle did *not* send:

```go
if _, forgotten := b.assertForgottenDuringFlush[peerID]; forgotten {
    if !sendRegister {
        st.registeredAt = time.Time{}
    }
    if !sendConnected {
        st.connectedAt = time.Time{}
    }
}
```

The comment reasons that "the assertions this cycle actually sent" are still true, so they should be kept. That holds only when the reconciler's clear landed *before* the batcher's write. In the other interleaving it does not, and the reconciler is the one that wins in the registry:

1. `flushOnce` reads `st` for peer X, `sendConnected` is true (last assert older than `registryReassertTTL`), and sends `UpdateConnectionState(X, true)`.
2. X's connection drops. The reconcile pass, whose snapshot no longer contains X, sends `UpdateConnectionState(X, false)` — this lands **after** step 1, so the registry ends at `false`.
3. The reconciler calls `forgetAssertState(X)`, which is recorded in `assertForgottenDuringFlush`.
4. The flush re-records. `sendConnected` was true, so `connectedAt` is kept at `now`.

The registry now says `IsConnected=false` while the batcher believes it asserted connected a moment ago. X reconnects and gossips a few seconds later: the hot path correctly sees a live connection and enqueues `markConnected`, but `sendConnected` evaluates `now.Sub(st.connectedAt) > registryReassertTTL` as false and the assert is skipped. X stays `IsConnected=false` for up to `registryReassertTTL` (1 minute), until the next reconcile pass happens to flag it. That is the same "a live neighbour is invisible for up to a minute" failure I raised in round 3, in a narrower window — and `forgetAssertState` was introduced specifically to guarantee it can't happen.

Reproduced at `3c2cac4be`. The hook on the counting client fires *before* the inner RPC, so to get the clear to land after the batcher's write I hooked the next RPC in the same cycle (`UpdateLastMessageTime`) instead:

```go
client.onUpdateLastMessageTime = func() {
    require.NoError(t, inner.UpdateConnectionState(context.Background(), pid, false))
    b.forgetAssertState(pid)
}
b.enqueueRegister(pid, "", 0, nil, "", true)
b.enqueueLastMessage(pid)
b.flushOnce(context.Background())
// registry is false here, as expected

// peer reconnects and gossips
b.enqueueRegister(pid, "", 0, nil, "", true)
b.flushOnce(context.Background())
got, _ = reg.Get(pid)
require.True(t, got.IsConnected)   // FAILS
```

```
--- FAIL: TestRepro_ReconcileClearMaskedByKeptConnectedAt (0.00s)
    Messages: a reconnected, gossiping peer must be re-flagged connected on its next message
```

**Suggested fix** — zero both halves unconditionally when the peer was forgotten mid-flush:

```go
if _, forgotten := b.assertForgottenDuringFlush[peerID]; forgotten {
    st.registeredAt = time.Time{}
    st.connectedAt = time.Time{}
}
```

The reconciler's clear is authoritative about what the registry currently holds, and the batcher cannot tell from inside the cycle which write landed last, so forgetting both is the only safe reading of a `forgetAssertState` that raced the RPCs. The cost is one redundant `RegisterPeer` + `UpdateConnectionState` pair on the peer's next message, and only for peers the reconciler actually cleared — bounded by real disconnects, not by gossip volume. There is no re-assert loop: if the peer is genuinely gone it sends no further messages, so nothing re-enqueues it.

I applied that patch in a scratch worktree. My repro passes, and the whole `./services/p2p/` suite under `-race` is green except for one test:

```
--- FAIL: TestPeerRegistryBatcher_ForgetDuringConnectOnlyFlushZerosRegisteredAt
    peer_registry_batcher_test.go:245: Should be false
    Messages: connect assert sent this cycle is kept
```

That assertion — `require.False(t, st.connectedAt.IsZero(), "connect assert sent this cycle is kept")` — is pinning the behaviour that causes the bug, so it needs to become `require.True(..., st.connectedAt.IsZero(), ...)` along with the fix. Worth adding the reconnect-after-masked-clear case as a test too, since that is the observable consequence rather than the internal field state.

### Smaller notes, none blocking

**`reconcileConnectionStates`' set-true direction doesn't touch the batcher.** When it flags a live peer the hot path missed, it writes the registry but leaves `lastAsserted[peerID].connectedAt` untouched, so the peer's next message re-sends `UpdateConnectionState(true)` redundantly. Harmless and self-limiting — worth a line of comment saying it's deliberate, so a future reader doesn't mistake it for the same class of bug as the one above.

**`snapshotLiveConnIDs` doesn't feed `liveConnCache`.** After a reconcile clears a peer, the hot path can still answer "live" from a cache entry up to `reputationCacheTTL` old and re-flag it. The doc comment on `hasLiveConnection` already states the 5s staleness contract, so this is consistent rather than wrong — but seeding the cache from the reconcile snapshot would be nearly free and would tighten both directions.

**`liveConnCache` is the third per-peer-ID cache swept only once per `p2p_peer_map_cleanup_interval`.** Entries expire after 5s but are not deleted until the 60s sweep, so under a flood of fresh spoofed peer IDs it accumulates one entry per unique author per minute — alongside `ipBanCache` and `banStatusCache`, which already behave identically on `main`. No new order of growth, and the registry itself is now genuinely bounded, which is the point of the PR. Just noting that the same-shaped transient is now paid three times over; if the sweep interval is ever raised for CPU reasons these are what grow.

**`isCleanupExempt`'s TTL condition can now evict a connected-but-silent peer entirely.** A live peer whose `peerActivity` is older than `p2p_peer_registry_ttl` loses its exemption and is dropped, and `reconcileConnectionStates` only iterates existing entries so it won't re-add it — its reputation and ban history go with it. Unreachable for a `full`-mode neighbour, which heartbeats on `node_status` every 10 seconds against a 24h TTL. It is theoretically reachable for a neighbour in `listen_only`/`silent` mode that never publishes, if such an entry gets into the registry by another route. Fine as it stands; mentioning it because the doc table says such a peer "stays unflagged" and it is now also evictable.

### On the residuals section

The two headline residuals are both stated accurately and I agree with the framing. The upstream `peerTracker.topicPeers` growth is the real remaining hole and a `Connectedness(peer.ID)` accessor is the right ask — it would collapse both the walks and the connected-but-silent caveat. The LRU-evicts-honest-peers-first point is correctly attributed to pre-existing ordering rather than to this change, and correctly flagged as something this PR makes matter.

Once the `forgetAssertState` masking is closed and its pinning test updated, this looks good to me.

services/p2p/peer_registry_batcher.go
```diff
@@ -454,6 +487,17 @@ func (b *peerRegistryBatcher) flushOnce(ctx context.Context) {
...
```
freemans13, line 493: **Blocker.** This only forgets the halves the cycle did *not* send, which reopens the window `forgetAssertState` exists to close.

The reasoning — "the assertions this cycle actually sent are still true" — holds only when the reconciler's clear landed *before* the batcher's write. In the other ordering the reconciler wins in the registry but the batcher keeps its assert memory:

1. This cycle reads `st` with `sendConnected` true and sends `UpdateConnectionState(X, true)`.
2. X's connection drops; the reconcile pass, whose snapshot no longer holds X, sends `UpdateConnectionState(X, false)` — landing **after** step 1, so the registry ends at `false`.
3. The reconciler calls `forgetAssertState(X)`, recorded in `assertForgottenDuringFlush`.
4. We get here, `sendConnected` was true, so `connectedAt` is kept at `now`.

The registry now says disconnected while the batcher believes it just asserted connected. X reconnects and gossips a few seconds later: the hot path correctly sees a live connection and enqueues `markConnected`, but `now.Sub(st.connectedAt) > registryReassertTTL` is false, so the assert is skipped and X reads disconnected for up to a minute — until a reconcile pass happens to flag it. That is the round-3 failure class in a narrower window, inside the mechanism added to prevent it.

Reproduced at `3c2cac4be`. The counting client's hook fires *before* the inner RPC, so hooking `UpdateConnectionState` gives you the ordering that already works; hooking the next RPC of the same cycle puts the clear after the batcher's write:

```go
client.onUpdateLastMessageTime = func() {
    require.NoError(t, inner.UpdateConnectionState(context.Background(), pid, false))
    b.forgetAssertState(pid)
}
b.enqueueRegister(pid, "", 0, nil, "", true)
b.enqueueLastMessage(pid)
b.flushOnce(context.Background())
// registry is false here, as expected

// peer reconnects and gossips
b.enqueueRegister(pid, "", 0, nil, "", true)
b.flushOnce(context.Background())
got, _ = reg.Get(pid)
require.True(t, got.IsConnected)   // FAILS
```

Suggested fix — forget both halves unconditionally:

```suggestion
				if _, forgotten := b.assertForgottenDuringFlush[peerID]; forgotten {
					st.registeredAt = time.Time{}
					st.connectedAt = time.Time{}
				}
```

The reconciler's clear is authoritative about what the registry currently holds, and the batcher cannot tell from inside the cycle which write landed last, so forgetting both is the only safe reading of a `forgetAssertState` that raced the RPCs. Cost is one redundant `RegisterPeer` + `UpdateConnectionState` pair on the peer's next message, only for peers the reconciler actually cleared — bounded by real disconnects, not by gossip volume. No re-assert loop either: a peer that is genuinely gone sends no further messages, so nothing re-enqueues it.

I applied that patch in a scratch worktree. My repro passes and the whole `./services/p2p/` suite is green under `-race`, with one exception — see my comment on `peer_registry_batcher_test.go:245`.

services/p2p/peer_registry_batcher_test.go
```diff
@@ -145,6 +170,82 @@ func TestPeerRegistryBatcher_SkipsReassertWithinTTL(t *testing.T) {
...
+	b.enqueueRegister(pid, "", 0, nil, "", true)
+	b.flushOnce(context.Background())
+	counting.onUpdateConnectionState = nil
+
+	// The connect assert this cycle actually sent is kept; the not-sent
+	// register half of the forgotten snapshot must not be resurrected.
+	b.mu.Lock()
+	st = b.lastAsserted[pid]
+	b.mu.Unlock()
+	require.False(t, st.connectedAt.IsZero(), "connect assert sent this cycle is kept")
```
freemans13, line 245: This assertion pins the behaviour that causes the blocker I raised on `peer_registry_batcher.go:493`, so it is the one test that fails once that is fixed:

```
--- FAIL: TestPeerRegistryBatcher_ForgetDuringConnectOnlyFlushZerosRegisteredAt
    peer_registry_batcher_test.go:245: Should be false
    Messages: connect assert sent this cycle is kept
```

It needs to invert to `require.True(t, st.connectedAt.IsZero(), ...)` alongside the fix. Worth also adding the reconnect-after-masked-clear case as a test, since the registry flag a reconnecting peer ends up with is the observable consequence — the internal field state is only the mechanism.

services/p2p/server_helpers.go
```diff
@@ -1192,13 +1213,153 @@ func (s *Server) startPeerMapCleanup(ctx context.Context) {
...
+				continue
+			}
+			// Drop the batcher's reassert memory so a peer that reconnects
+			// gets re-marked connected on its next message instead of being
+			// skipped as recently asserted.
+			if s.registryBatcher != nil {
+				s.registryBatcher.forgetAssertState(info.ID)
+			}
+			cleared++
+		case !info.IsConnected && isLive:
```
freemans13, line 1349: Non-blocking: the set-true direction writes the registry but leaves `lastAsserted[info.ID].connectedAt` untouched, so the peer's next message re-sends `UpdateConnectionState(true)` redundantly. Harmless and self-limiting, but worth a line of comment saying it is deliberate — the clear-direction branch just above deliberately *does* poke the batcher via `forgetAssertState`, so the asymmetry reads like an oversight rather than a choice.

services/p2p/server_helpers.go
```diff
@@ -1192,13 +1213,153 @@ func (s *Server) startPeerMapCleanup(ctx context.Context) {
...
+// snapshotLiveConnIDs returns the set of peer IDs that currently have an
+// open libp2p connection. Liveness comes from P2PClient.GetPeers(): verified
+// against go-p2p-message-bus v0.1.17 (client.go GetPeers), Addrs is built
+// from host.Network().ConnsToPeer — open connections only, not the peerstore
+// — while the peer list itself is every peer that ever authored a message on
+// a subscribed topic (gossip-only publishers included, never pruned). So the
+// Addrs filter is what separates live neighbours from gossip-only authors;
+// it is not redundant with the listing. This differs from the p2p service's
+// own GetPeers RPC, which is filtered to IsConnected registry entries.
+func (s *Server) snapshotLiveConnIDs() map[string]struct{} {
```
freemans13, line 1246: Non-blocking: this snapshot does not feed `liveConnCache`, so right after a reconcile clears a peer the hot path can still answer "live" from a cache entry up to `reputationCacheTTL` old and re-flag it. The doc comment on `hasLiveConnection` already states that staleness contract, so this is consistent rather than wrong — but seeding the cache from the snapshot you have just built would be close to free and would tighten both directions at once.

services/p2p/server_helpers.go
```diff
@@ -946,6 +946,21 @@ func (s *Server) cleanupPeerMaps() {
 		s.ipBanCache.Delete(key)
 	}
 
+	// Evict expired liveConnCache entries: the liveness checks insert one
```
freemans13, line 949: Non-blocking observation on shape rather than a defect. Entries expire after `reputationCacheTTL` (5s) but are not deleted until this sweep runs, once per `p2p_peer_map_cleanup_interval` (60s), so under a flood of fresh spoofed peer IDs the map accumulates one entry per unique author per minute. `ipBanCache` and `banStatusCache` immediately below behave identically on `main`, so there is no new order of growth here, and the registry itself is now genuinely bounded — which is the point of the PR.

Just flagging that the same-shaped transient is now paid three times over: if this interval is ever raised for CPU reasons, these three maps are what grow.

services/blockchain/peer_registry.go
```diff
@@ -1378,9 +1379,20 @@ func (r *CentralizedPeerRegistry) Cleanup(maxSize int, ttl time.Duration) (int,
...
+// clear it, so a stale flag must not pin an entry in the registry forever.
+// Caller must hold the registry lock.
+func isCleanupExempt(info *PeerInfo, now time.Time, ttl time.Duration) bool {
```
freemans13, line 1387: Non-blocking: the TTL condition means a peer with a live connection whose `peerActivity` is older than `ttl` loses its exemption and is dropped entirely, taking its reputation and ban history with it — and `reconcileConnectionStates` only iterates existing entries, so it will not re-add it.

Unreachable for a `full`-mode neighbour, which publishes `node_status` every 10 seconds against a 24h default TTL, so I am not asking for a change. Raising it because the docs table describes a connected-but-silent peer as one that "stays unflagged", and it is now also evictable — theoretically reachable for a `listen_only`/`silent` neighbour that never publishes, if such an entry reaches the registry by another route.

freemans13 (commented):
Same round-4 findings as my review above, now anchored inline so each one sits on the code it concerns. Re-checked against the current head `24d53e5d3`: the branch-authored code these anchor to is byte-identical to `3c2cac4be`, the merge only shifted line numbers in `server_helpers.go`. The blocker is the first comment on `peer_registry_batcher.go`; the rest are non-blocking.

For context on the headline result: the round-3 snapshot-gate regression is genuinely fixed by `3c2cac4be`, and I verified `TestServerHelpers_NewNeighbourFlaggedOnFirstMessage` is a real regression test by dropping it unchanged into a scratch worktree at `b9acd9eb9`, where it fails.

services/p2p/server_helpers.go
```diff
@@ -1192,13 +1213,170 @@ func (s *Server) startPeerMapCleanup(ctx context.Context) {
...
+
+	live := s.snapshotLiveConnIDs()
+
+	// Seed the per-peer liveness cache from the snapshot just built: nearly
+	// free (bounded by open connections), and it tightens the hot path right
+	// after this pass — a peer this pass is about to clear must not be
+	// re-flagged off a stale cached "live" for the next few seconds.
+	now := time.Now()
+	for id := range live {
+		s.cacheLiveConn(id, true, now)
```
freemans13, line 1325: The comment says the seed exists so "a peer this pass is about to clear must not be re-flagged off a stale cached `live`". That can't be what this loop does. It writes `true` only for IDs **in** `live`, and the clear branch below fires for `info.IsConnected && !isLive` — peers **absent** from `live`. The two sets are disjoint, so no peer this pass is about to clear can be touched here. The line that actually provides that protection is the `s.cacheLiveConn(info.ID, false, …)` you added on the clear branch.

What the loop does do is skip the targeted `GetPeers()` walk inside `hasLiveConnection` for the first message from each live peer within `reputationCacheTTL` of a pass — about 5 seconds in every 60, which is marginal given `isPeerIPBanned` already walks and caches on the gossip path.

Against that, it widens the stale-true window slightly, in the one direction this PR exists to narrow. Before this loop a live-but-silent peer had no cache entry at all, so a message claiming `FromID = X` fell through to a fresh walk and got the truth. Now every live peer carries a cached `true` for 5s after each pass whether or not it has ever gossiped. If X's connection drops inside that window, a spoofed `FromID = X` reads the cached `true`, the hot path enqueues `markConnected`, and X reads connected until the next pass clears it — up to a minute — during which `isCleanupExempt` also keeps its entry LRU-exempt.

`cacheLiveConn` stores unconditionally with no expiry comparison, so the loop can clobber a fresher and truer entry as well: a concurrent gossip walk that observes X's connection has just gone writes `false` with a later expiry, and this `Store` then overwrites it with `true`.

All of it is bounded and reporting-only — `IsConnected` still isn't read by the sync coordinator — so nothing here blocks. But since the loop can't deliver the protection its comment claims, and does carry that cost, I'd drop it and keep the clear-branch line.

services/p2p/server_helpers.go
```diff
@@ -1192,13 +1213,170 @@ func (s *Server) startPeerMapCleanup(ctx context.Context) {
...
+			if err := s.peerRegistry.UpdateConnectionState(ctx, info.ID, false); err != nil {
+				s.logger.Warnf("[reconcileConnectionStates] UpdateConnectionState %s false failed: %v", info.ID, err)
+				continue
+			}
+			// Overwrite any stale cached "live" so the hot path cannot
+			// re-flag this peer off a pre-disconnect cache entry, and drop
+			// the batcher's reassert memory so a peer that reconnects gets
+			// re-marked connected on its next message instead of being
+			// skipped as recently asserted.
+			s.cacheLiveConn(info.ID, false, time.Now())
```
freemans13, line 1356: Two small things on this line, neither blocking.

**Ordering.** The poison lands after `UpdateConnectionState(ctx, info.ID, false)` returns, so for the duration of that RPC a message for this peer can still read the pre-disconnect cached `true`, enqueue `markConnected`, and have the flush re-assert `true` after the clear. Moving this line above the RPC closes that gap for free, and it is safe on the error path: if the RPC fails and we `continue`, a cached "not live" for a peer that genuinely isn't live is the right answer anyway.

**Coverage.** This line has no test. I deleted it and the seed loop above and re-ran the full `./services/p2p/` suite — green, so nothing in the suite distinguishes their presence from their absence. Of the two, this is the one holding a correctness property, so it is the one worth pinning. The shape: registry holds the peer connected, the snapshot does not contain it, the liveness cache holds a fresh `true`, run a pass, then assert a following `updatePeerLastMessageTime` does **not** re-flag it. Without this line that assertion fails, which is exactly the property you want nailed down.

freemans13 (approved):
## Round 5 — the round-4 blocker is closed and properly pinned. Approving.

Reviewed at `3d63a4b17`. Only one branch-authored commit since my last round (`28523c9e6`); the rest of the range is merged `main`, which I've left out of scope.

### The blocker is genuinely fixed, and the new tests really do catch it

`28523c9e6` takes the suggested fix as-is — `st = registryAssertState{}` — and since that struct holds nothing but `registeredAt` and `connectedAt`, zeroing the whole thing is exactly the two-field reset and nothing more. The pinning assertion in `TestPeerRegistryBatcher_ForgetDuringConnectOnlyFlushZerosRegisteredAt` was inverted and the test renamed to `…ZerosSnapshot`, and you added the reconnect case as its own test, which is the right call — the field state is the mechanism, the reconnect is the consequence anyone actually cares about.

I checked those are real regression tests rather than restatements of the new code, by putting the old conditional back in a scratch worktree at this same head and running them there:

```
--- FAIL: TestPeerRegistryBatcher_ForgetDuringConnectOnlyFlushZerosSnapshot
    Messages: connect assert must be forgotten even though this cycle sent it
--- FAIL: TestPeerRegistryBatcher_ReconnectAfterMaskedClearIsReflagged
    Messages: a reconnected, gossiping peer must be re-flagged connected on its next message
```

Both fail with the fix reverted and pass with it in place, and `TestPeerRegistryBatcher_ForgetAssertStateDuringFlushNotResurrected` passes either way — so the two new tests are carrying the weight, not an existing one that happens to overlap. That closes the last of the "a live neighbour reads disconnected for up to a minute" family across rounds 3 and 4.

The three non-blocking notes from round 4 are all addressed properly too: the set-true branch now says in a comment that skipping the batcher poke is deliberate, the doc row records that an unflagged peer is evictable once idle (which matches `isCleanupExempt` — no `IsConnected`, no exemption, so it falls to the LRU like any idle entry), and `reconcileConnectionStates` now seeds the liveness cache. That last one is where my two remaining notes are.

### What I ran

| Check | Result |
|---|---|
| `SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./services/p2p/... ./services/blockchain/...` | all ok (p2p 24.5s, blockchain 42.1s) |
| `golangci-lint run ./services/p2p/... ./services/blockchain/... ./settings/...` | 1 hit, `prealloc` at `services/p2p/mock.go:76` — not in this branch's diff, so pre-existing |
| Fix reverted in a scratch worktree, both new tests re-run | both FAIL, as above |
| Both new `cacheLiveConn` lines deleted, full `./services/p2p/` suite re-run | **ok** — see the coverage note below |

### Two non-blocking notes, both on the new cache seeding

Neither blocks. Details are inline, but in short:

**The seed loop can't do what its comment says it does.** It writes `true` for every ID in `live`, and the peers this pass is about to clear are by definition the ones *absent* from `live` — the two sets are disjoint. The protection against a stale cached "live" re-flagging a cleared peer comes entirely from the separate `cacheLiveConn(info.ID, false, …)` on the clear branch, which is the line that actually does the work. What the loop buys on its own is skipping a `GetPeers()` walk on the first message from each live peer in the 5s after a pass, and it costs a small widening in the direction this PR exists to close. My suggestion is to drop the loop and keep the clear-branch line.

**Neither cache line has any coverage.** I deleted both and ran the whole `./services/p2p/` suite: green. Nothing in the suite can tell whether either exists, which is worth closing given the clear-branch line is the one holding a correctness property.

Approving on the strength of the blocker fix — these two are cleanup, and I'm happy either way on them.

PR #1602, opened 2026-08-18
config(policy): raise maxscriptsizepolicy default to 100MB

galt-tr:
## What

Raises the default for `maxscriptsizepolicy` from **500000** (500KB) to **100000000** (100MB).

## Why

`maxscriptsizepolicy` caps the size of an individual locking or unlocking script during transaction validation. It is pushed into the BDK script engine at validator startup (`services/validator/ScriptVerifierGoBDK.go`, `se.SetMaxScriptSizePolicy(...)`).

The key is not present in `settings.conf`, so nothing overrides it — the effective default came from the Go fallback in `settings/settings.go` at 500KB. Meanwhile `settings.conf` already ships `maxtxsizepolicy = 100000000` (100MB), so the script cap was 200x tighter than the transaction cap and was the binding constraint on otherwise-acceptable large-script transactions.

This change lines the two up.

## Safety

- **Policy only, not consensus.** This limit is bypassed during block validation (`SkipPolicyChecks=true`), so it cannot cause a consensus split — nodes running different values still accept the same blocks.
- **Other bounds still apply.** Script execution memory remains bounded by `maxstackmemoryusagepolicy` (100MB) and total transaction size by `maxtxsizepolicy` (100MB). This drops a redundant sub-limit rather than removing all bounds.
- **Still operator-tunable.** Anyone wanting the old behaviour sets `maxscriptsizepolicy = 500000` in their config; the override path is covered by a test.

## Changes

- `settings/settings.go` — the effective default: `getInt("maxscriptsizepolicy", 100000000, ...)`.
- `settings/policy_settings.go` — the `default:`/`desc:`/`longdesc:` struct-tag metadata that feeds `ExportMetadata()` (the settings portal), so it tracks the code default.
- `docs/references/settings/policy_settings.md` — reference table and the "Default BSV Configuration" example. The "Conservative Policy Configuration" example is deliberately left at a lower value.
- `settings/settings_test.go` — pins the resolved default and the env-override path, following the existing `TestMaxRawTxFee_*` pattern.

`settings.conf` is intentionally untouched — the Go fallback is the single authoritative default, and duplicating the value there would just create two places to keep in sync.

## Not touched

- `services/legacy/txscript/engine.go` `MaxScriptSize = 110000` — a constant in the legacy Go script engine whose `NewEngine` has no production callers (test-only); it does not gate validation and is not configuration.
- `stores/utxo/utils.go` `maxScriptSizeBeforeGenesis = 10000` — a pre-Genesis consensus constant used to identify provably-unspendable outputs; unrelated.

## Testing

```
go test ./settings/... ./services/asset/httpimpl/... ./services/p2p/...   # ok
go test ./services/validator/...                                          # logic tests ok
go vet ./settings/... ./services/validator/...                            # clean
make lint                                                                 # 0 issues
```

Verified end-to-end that the value resolves live against the committed `settings.conf`: `maxscriptsizepolicy` reports `100000000`, and the settings-portal metadata reports the same for both default and current value.

One pre-existing failure in `./services/validator/...` (`TestExtendedTxa1f6a...`) is a testcontainers/Aerospike Docker startup problem — it reproduces identically on a clean tree at `origin/main`, so it is unrelated to this change.

https://claude.ai/code/session_01CqFYSRx3YW2X2ximnrNmJV


ordishs (approved):
Approving. Correct, minimal, well-scoped. Verified the claims independently rather than taking the description at face value; three minor nits below, none blocking.

## Verified

| Claim | Result |
| --- | --- |
| Key absent from `settings.conf`, so the Go fallback is the effective default | Confirmed — only `maxtxsizepolicy = 100000000` is present (`settings.conf:832`); no `maxscriptsizepolicy` line |
| Value is pushed into BDK at validator startup | Confirmed — `services/validator/ScriptVerifierGoBDK.go:183` |
| Policy-only, bypassed on the block path, so no consensus split | Confirmed — block path uses `validator.WithSkipPolicyChecks(true)` (`services/subtreevalidation/Server.go:903,958`; `check_block_subtrees.go:649,1073`) |
| Other bounds still apply | Confirmed — `maxstackmemoryusagepolicy` 104857600 (`settings/settings.go:112`), and total tx size caps any single script, so the sub-limit really was redundant |
| Tests pass | Confirmed — `go test -count=1 -run TestMaxScriptSizePolicy ./settings/` 2 PASS; `./services/asset/httpimpl/... ./services/p2p/...` ok |

One risk the description does not mention, which I checked: `newScriptVerifierGoBDK` **panics** if any BDK setter rejects its value, and the shipped `libGoBDK` archive contains the string `"Policy value for max script size must not exceed consensus limit of "`. A default above that consensus ceiling would therefore panic the validator at startup, not merely warn. I built a throwaway test constructing the verifier with the new default against real BDK — it passes, so 100MB is accepted and there is no startup-panic regression. (Test removed afterwards.)

I also grepped for stale copies of the old value: no other production code, config, docs, or UI references the 500KB default. Every remaining `500000` in tests (`services/asset/httpimpl/get_policy_test.go`, `services/p2p/fee_policy_test.go`) sets the field explicitly, so those are unaffected.

## Nits

1. `settings/policy_settings.go:9` — "matches the maxtxsizepolicy ceiling" holds only for the committed `settings.conf`. `MaxTxSizePolicy`'s own default is `10485760` (10MB), so an operator whose config omits `maxtxsizepolicy` gets a 10MB tx cap with a 100MB script cap. Harmless (the script cap is then inert), but the wording overstates it — suggest "matches the 100MB `maxtxsizepolicy` shipped in `settings.conf`".
2. `docs/references/settings/policy_settings.md` — the "Default BSV Configuration" block now lists `maxtxsizepolicy = 10485760` directly above `maxscriptsizepolicy = 100000000`, i.e. a script cap 10x the tx cap. It is just the two Go defaults side by side, but it reads like a mistake. Same for the table rows near line 20-22.
3. Worth stating explicitly in the Safety section that 500000 is also bitcoin-sv node's post-Genesis policy default, so after this change Teranode accepts and relays scripts that default-configured svnode peers reject. That is relay/mempool asymmetry, not consensus, and there is precedent (`maxtxsizepolicy` already diverges at 100MB vs svnode's 10MB) — but it is the one real behavioural consequence and it belongs in the writeup.

## Non-blocking

- `test/e2e/daemon/ready/smoke_test.go:879-926` derives its oversized script from the **live** setting (`maxScriptSize+1000` bytes plus a `maxScriptSize` padding push), so with the new default it would allocate roughly 200MB. It is `t.Skip()`'d at line 855 so it cannot bite today, but it is a trap for whoever re-enables it. It should pin its own override — `compose/settings_test.conf:372` already defines `maxscriptsizepolicy.dev.system.test.oversizedscripttest = 10000`, currently unused by any test.
- `TestMaxScriptSizePolicy_Default` goes through `NewSettings()`, so a developer's uncommitted `settings_local.conf` setting this key fails it locally. Same property as the existing `TestMaxRawTxFee_*` tests it mirrors, so consistent with house style.

icellan (approved):

PR #1603, opened 2026-08-18
feat(utxo): add an embedded Pebble UTXO store (experimental)

ordishs:
## Status: experimental / spike

A complete `utxo.Store` implementation over an embedded [Pebble](https://github.com/cockroachdb/pebble) LSM (CockroachDB's storage engine, also go-ethereum's default backend). Zero external infrastructure: no database server, no cluster — the store is a directory. Selectable via `utxostore = pebble:///path/to/dir`. Additive: no existing store is touched.

Opened for visibility and discussion rather than immediate merge — see the promotion checklist below for what this still owes. It explores the single-binary node deployment shape: an SVNode-style `./teranode` with no Aerospike cluster or Postgres server to operate.

## Design

The packed transaction-centric layout from #1594 mapped onto a keyspace:

- One master record per transaction (flags, counters, mined refs, page-0 spend slots), overflow page records past 64 outputs, an immutable payload record (inputs/outputs as offset-indexed blobs), and per-page utxo-hash records.
- Secondary access paths are dedicated key ranges — unmined-since, delete-at-height, preserve-until, conflicting, conflicting-children (with reverse index), per-output overrides, and the conflict-resolution WAL — every index maintained in the **same atomic WriteBatch** as its mutation.
- Concurrency: 1024 stripe locks acquired in sorted order; `SpendAndCreate` is atomic by construction (one batch), and the `ErrTxExists` leave-spends-in-place contract falls out naturally by committing the batch with only the spends staged.
- Pruning walks the delete-at-height index and removes records with range-delete tombstones that compaction reclaims.
- Page size is stamped into the store on first open and validated on every subsequent open.
- Commits use `pebble.Sync` (WAL fsync; Pebble group-commits concurrent syncs).

## Testing

- The full shared conformance suite from `stores/utxo/tests` — the same 26 tests the Aerospike and packedsql stores run (conflict WAL crash recovery, spend error taxonomy, block-state contracts, the whole SpendAndCreate family) — passes in ~2s with no containers. `-race` clean, `golangci-lint` 0 issues.

## Benchmarks (shared `tests.Benchmark` harness, same machine — Apple M5 Max, indicative)

```
pebble (NoSync)      0.31 ms/op
packedsql (PG16)     1.23 ms/op   (PR #1594)
sql store (PG16)    24.5  ms/op
pebble (Sync)       13.2  ms/op
```

Read the Sync row carefully: it is one *real* macOS fsync per operation on a **serial** harness — the worst case by construction, since Pebble amortises WAL syncs across concurrent commits and this harness offers none. The Postgres containers meanwhile never pay a true fsync (Docker's VM absorbs it), so NoSync-vs-container is the like-for-like comparison. A concurrent benchmark is on the checklist.

## Constraints

- **Single-process**: an embedded store binds to one process, so this fits the `-all=1` deployment, not split-services topologies.
- **Fresh-sync only**, like any new store backend.

## Promotion checklist (deliberately not done in the spike)

- [ ] Extract the slot/blob codec shared with #1594 into a common package (duplicated for now so neither PR blocks the other)
- [ ] Crash-recovery testing (kill mid-batch, reopen, verify invariants)
- [ ] Concurrent benchmark to characterise group-commit under load
- [ ] Mainnet sync validation
- [ ] Pebble options tuning (cache size, compaction concurrency, WAL settings)

## Relationship to other work

Third point in the same design space as #1594 (packedsql, in-place patch + synchronous bookkeeping on Postgres) and #684 (postgres store, append-only spends + deferred sweep). All three now run the same layout philosophy against the same conformance contract; this one trades multi-process access for the smallest operational footprint and the best single-box numbers so far.


liam (commented):
Reviewed as a spike rather than a merge candidate, so I focused on three things: what affects people who are not using it, claims in the description that overstate, and design observations worth having before the promotion work starts.

## The locking design is sound

This is the part a UTXO store gets wrong, and it is right here. Spends mutate the *parent's* master record, and the lock set matches: `Spend` locks one stripe per distinct parent group, `SpendAndCreate` locks those plus the self hash, deduped and sorted before acquisition so there is no lock-order inversion. The single-batch atomicity claim is genuine, and the `ErrTxExists` leave-spends-in-place contract really does fall out of committing with only the spends staged -- I read that path and it is as described rather than special-cased.

The conformance wiring is real too: `conformance_test.go` calls the same `tests.Store` / `tests.Spend` / `tests.Restore` entry points the other stores use, and 27 subtests pass. Build clean, `-race` green in ~3s.

## The spike is less isolated than "additive" suggests

`factory/pebble.go` registers via an unconditional `init()`, and `stores/utxo/factory` is imported by `daemon/daemon_stores.go`, `cmd/seeder` and `cmd/rewindblockchain`. So **Pebble and its entire transitive tree compile into every Teranode binary** whether or not anyone sets `utxostore = pebble://`. That tree includes `getsentry/sentry-go`, reachable from the package per `go list -deps`. It is linked rather than necessarily active -- `cockroachdb/errors` only reports when a client is configured -- but a telemetry SDK entering a consensus node's dependency graph as a side effect of an experiment seems worth a deliberate decision.

A `//go:build pebble` tag on `factory/pebble.go` would make the spike genuinely opt-in and match the "no existing store is touched" framing.

Related and concrete: **`go.mod` was not tidied.** `github.com/cockroachdb/pebble/v2 v2.1.6` is marked `// indirect`, but `create.go`, `get.go` and `spend.go` import it directly -- the whole Pebble block landed in the indirect section.

## The headline benchmark number is not reachable

`New()` hardcodes `sync: pebble.Sync`, and nothing reads a query parameter or setting. The only way to get `NoSync` is `BenchmarkSuiteNoSync` poking the unexported field from inside the package. So the `0.31 ms/op` row -- and the framing that "NoSync-vs-container is the like-for-like comparison" -- describes a configuration an operator cannot select. The only number a deployment can actually produce today is the 13.2 ms/op Sync row.

That is about three lines to fix (a `?sync=false` on the store URL, which this config style already supports elsewhere), and it would make the comparison real rather than hypothetical.

## Lock footprint scales with distinct parent count

Stripe locks are held across `batch.Commit(pebble.Sync)`, which is a real fsync. The number of stripes taken is the number of *distinct parent transactions*, so a transaction spending many distinct parents holds a large fraction of the 1024 stripes for the duration of that fsync, and a few thousand distinct parents saturates the stripe space outright.

Teranode specifically targets very large transactions, so this is worth characterising in the concurrent benchmark already on the checklist. It may be the thing that limits the group-commit amortisation the Sync row is counting on, since group commit only helps when commits are genuinely concurrent.

## One small robustness gap

`validateMeta` does `binary.LittleEndian.Uint32(val)` with no length check, so a short or corrupt meta record panics rather than returning a storage error. One line, and it is the first thing touched on every open.

---

For a spike the correctness core is in better shape than I expected -- passing the same conformance contract as Aerospike and packedsql is meaningful evidence, not a formality. My only ask before this sits on `main` in any form is the build tag, since that is the one item that costs everyone else something.

stores/utxo/pebble/misc.go
```diff
@@ -0,0 +1,639 @@
...
+	var deleted int64
+
+	for i := range victims {
+		h := victims[i]
+
+		unlock := p.store.lockStripes(h[:])
+
+		batch := p.store.db.NewBatch()
+
+		err := p.store.stageDelete(batch, h[:])
```
freemans13: The victim list was built at line 589 with no lock held, and `stageDelete` deletes unconditionally (`get.go:411` reads the master only to find which index keys to clean up). Taking the stripe lock here does not help, because nothing rechecks the condition the lock was taken for.

Between the scan and this iteration, `SetMinedMulti` with `UnsetMined` (`mined.go:43`), `Unspend` (`spend.go:467`) and `SetLocked(_, true)` (`misc.go:194`) all set `deleteAtHeight = 0`, and all three run in services outside the pruner worker. The first is the reorg path, so the transaction gets deleted at exactly the moment its outputs have to survive.

The window is the whole pass, one fsync per victim, so against a real backlog it is seconds. SQL has no equivalent window: `deleteTombstoned` evaluates `delete_at_height <= $1` inside the DELETE itself (`stores/utxo/sql/pruner/pruner_service.go:144`).

Suggested fix, here rather than in `stageDelete` since the public `Delete` must stay unconditional:

```go
m, err := p.store.getMaster(&h)
if err != nil || m.deleteAtHeight == 0 || m.deleteAtHeight > int64(height) {
    unlock()
    continue
}
```

stores/utxo/pebble/misc.go
```diff
@@ -0,0 +1,639 @@
...
+	if prunerServiceInstance != nil {
+		return prunerServiceInstance, nil
+	}
+
+	prunerServiceInstance = &prunerService{store: s}
+
+	return prunerServiceInstance, nil
+}
+
+type prunerService struct {
```
freemans13, line 648: `prunerService` holds no settings, and `Prune` uses only its `height` argument, so every `pruner_*` setting is a no-op on this backend.

`pruner_utxoDefensiveEnabled` is the one worth caring about. On SQL it swaps in a delete that will not remove a parent while any spending child is unmined or shallower than the safety window (`sql/pruner/pruner_service.go:152`). Set it on a Pebble node and nothing changes, with nothing logged to say so.

It defaults to false in production, so this is not live today. But a setting that quietly does nothing on one backend is the sort of thing found during an incident rather than before one. Either honour it or refuse to start when it is set.

stores/utxo/pebble/create.go
```diff
@@ -0,0 +1,275 @@
...
+	if built.master.flags&flagConflicting != 0 {
+		if err := batch.Set(conflictIdxKey(built.hash[:]), nil, nil); err != nil {
+			return errors.NewStorageError("pebble: failed to stage conflict index for %s", built.hash, err)
+		}
+
+		createdAt := make([]byte, 8)
+		putInt64(createdAt, built.master.createdAt)
+
+		for _, parent := range built.children {
+			if err := batch.Set(childrenKey(parent, built.hash[:]), createdAt, nil); err != nil {
```
freemans13, line 110: This writes under the *parent's* hash, but `Create` locked only the child's stripe at line 36. `SetConflicting` has the same shape (`misc.go:67` locks `txHash`, `misc.go:107` writes under each parent).

`stageDelete` range-deletes that same keyspace holding only the parent's stripe (`get.go:437`). Since `stripeOf(child)` and `stripeOf(parent)` are unrelated, `Create(child)` and `Delete(parent)` run concurrently over the same keys: the forward entry can land after the range delete and dangle, or be swept while its `childrenRevKey` reverse entry survives.

Not UTXO loss, but it does mean `GetConflictingChildren` can miss a child or report one against a parent that is gone, and the lock set no longer matches the mutation set. Adding the parent hashes to the `lockStripes` call in `Create` and `SetConflicting` would close it.

stores/utxo/pebble/mined.go
```diff
@@ -0,0 +1,474 @@
...
+
+	batch := s.db.NewBatch()
+	defer func() { _ = batch.Close() }()
+
+	for i := range expired {
+		h := expired[i]
+
+		m, err := s.getMaster(&h)
+		if err != nil {
+			continue
```
freemans13: This `continue` swallows every error, not just not-found. `PreserveTransactions` two functions up gets this right (`mined.go:380` matches `errors.ErrTxNotFound` specifically and returns anything else).

On a genuine read failure the preserve-index key is never cleared, so the same hash is rescanned on every later call and `preserveUntil` stays set forever, which keeps `deleteAtHeight` at 0 and the transaction unprunable. The storage error is also invisible: nothing logs it and `ProcessExpiredPreservations` still returns nil.

Same treatment as `PreserveTransactions`: skip on `ErrTxNotFound`, return otherwise.

stores/utxo/pebble/mined.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+func (it *heightIndexIterator) Next(ctx context.Context) ([]*utxo.UnminedTransaction, error) {
+	if it.done || it.err != nil {
+		return nil, it.err
+	}
+
+	lower := []byte{it.prefix}
+	if it.lastKey != nil {
+		lower = append(it.lastKey, 0)
```
freemans13: `lower` aliases `it.lastKey`'s backing array here. `it.lastKey` was filled by `append(it.lastKey[:0], key...)` at line 172, which for a 41-byte key allocates a 48-byte array, so `append(it.lastKey, 0)` has spare capacity and returns a slice over that same array rather than a copy.

Line 172 then runs again inside the loop and rewrites bytes 0..40 of that array while the iterator holding `lower` as its `LowerBound` is still open. `consistencyIterator` has the identical pair at lines 284 and 302.

I could not make this misbehave: forward `Next()` does not reconsult the lower bound, so today it looks latent rather than live. Still worth the one-line copy, since it is the kind of thing that only surfaces when a seek gets added to the resume path.

```go
lower = append(append([]byte(nil), it.lastKey...), 0)
```

stores/utxo/pebble/codec.go
```diff
@@ -0,0 +1,321 @@
...
+	binary.LittleEndian.PutUint32(b[26:], m.pagesSpent)
+	binary.LittleEndian.PutUint32(b[30:], m.version)
+	binary.LittleEndian.PutUint32(b[34:], m.lockTime)
+	binary.LittleEndian.PutUint64(b[38:], m.fee)
+	binary.LittleEndian.PutUint64(b[46:], m.sizeInBytes)
+	binary.LittleEndian.PutUint64(b[54:], uint64(m.createdAt))      //nolint:gosec
+	binary.LittleEndian.PutUint64(b[62:], uint64(m.deleteAtHeight)) //nolint:gosec
+	binary.LittleEndian.PutUint64(b[70:], uint64(m.unminedSince))   //nolint:gosec
+	binary.LittleEndian.PutUint64(b[78:], uint64(m.preserveUntil))  //nolint:gosec
+	binary.LittleEndian.PutUint16(b[86:], uint16(len(m.blockRefs))) //nolint:gosec
```
freemans13, line 250: The `//nolint:gosec` hides a truncation that would corrupt spend state rather than just lose a block reference. `decodeMaster` splits `blockRefs` from `spends` at this length (`codec.go:254-260`), so a wrapped value misaligns the page-0 spend slots, and page 0 is where `spentCount` and every DAH eligibility decision live.

At 12 bytes per reference the ceiling is 5461 block references on one transaction, and `appendBlockRef` never bounds the list, so this is not reachable in practice. But the failure mode is silent UTXO corruption rather than an error, which is worth more than a nolint. Either widen it to `uint32` (the record is 88 bytes, two more will not be noticed) or return a storage error when the length does not fit.

freemans13 (changes requested):
Read the whole package rather than the diff, and ran the build, vet, race tests and coverage locally. Requesting changes on one item only: the pruner can delete a transaction whose delete-at-height was cleared while the prune pass was already in flight. Everything else below is promotion-checklist material and none of it blocks the spike as a spike.

liam already covered the build tag, the untidied `go.mod`, the unreachable NoSync number, the stripe footprint under wide transactions and the `validateMeta` length check. I read that review first, agree with all five, and have not repeated them.

## The pruner deletes from a snapshot it never rechecks

`prunerService.Prune` collects every victim from the delete-at-height index with no lock held (`misc.go:589`), then walks that list taking one stripe lock per transaction and calling `stageDelete` (`misc.go:605`). `stageDelete` deletes unconditionally. It reads the master record only to work out which index keys to clean up (`get.go:411`).

Anything that clears delete-at-height between the scan and that transaction's turn in the loop is therefore ignored. Three writers do exactly that, and all three live in services that run alongside the pruner worker rather than inside it:

- `SetMinedMulti` with `UnsetMined` sets `deleteAtHeight = 0` (`mined.go:43`). That is the reorg path.
- `Unspend` sets it to 0 whenever it clears a spend (`spend.go:467`).
- `SetLocked(hashes, true)` sets it to 0 (`misc.go:194`).

The stripe lock does not close this, because the pruner never re-evaluates the condition it locked for. The window is the whole prune pass, and every victim costs an fsync, so against a real delete-at-height backlog that is seconds rather than microseconds. The SQL store has no equivalent window: `deleteTombstoned` is a single `DELETE ... WHERE delete_at_height <= $1` (`stores/utxo/sql/pruner/pruner_service.go:144`), so the condition is evaluated at delete time.

The bad case is a reorg un-mining a transaction, which is the exact moment its outputs have to survive, and an in-flight prune pass deleting it anyway. That is unspent UTXOs gone, and children that can never be revalidated.

One condition fixes it, in the pruner loop after the stripe lock and before staging: reread the master, skip when `deleteAtHeight == 0 || deleteAtHeight > height`. It belongs in the pruner and not in `stageDelete`, because the public `Delete` must still delete unconditionally.

## The pruner reads no settings

`prunerService` is `{store, observers, observersMu}` and `Prune` uses only the height argument. Across the whole package, exactly five settings are read: `CoinbaseMaturity`, `GenesisActivationHeight`, `GetUtxoStoreBlockHeightRetention`, `ConflictingChildrenMaxNodes` and `ReAssignedUtxoSpendableAfterBlocks`. Every `pruner_*` setting is therefore a no-op on this backend.

`pruner_utxoDefensiveEnabled` is the one that matters. On SQL it swaps in a delete that refuses to remove a parent while any spending child is unmined or shallower than the safety window (`sql/pruner/pruner_service.go:152`). Set it on a Pebble node and nothing changes, with no log line saying so. It defaults to false in production so this is not a live hazard today, but a setting that silently does nothing on one backend is the kind of thing that gets discovered during an incident. Worth either honouring it or refusing to start when it is set.

## Two writers reach outside their own stripe

The lock design holds for master records and spend slots. It does not hold for the conflicting-children keyspace.

`Create` with `WithConflicting` locks only the child's stripe (`create.go:36`) and then writes `childrenKey(parent, child)`, which lives under the parent's hash (`create.go:101`). `SetConflicting` does the same, locking only `txHash` (`misc.go:67`) before writing under each parent (`misc.go:107`). Meanwhile `stageDelete` range-deletes that whole keyspace for a transaction while holding only that transaction's stripe (`get.go:437`).

So `Create(child)` and `Delete(parent)` take disjoint stripes and run concurrently over the same keys. The forward entry can land after the range delete and dangle, or be swept while its `childrenRevKey` reverse entry survives. The result is either a child reported against a parent that no longer exists, or a child that `GetConflictingChildren` can no longer reach. Not a UTXO-loss bug, but it undercuts the claim that the lock set matches the mutation set, and the double-spend traversal is the thing that depends on it.

## Testing

Everything below is from this branch at `00e828a9e`, on this machine. No live node, no concurrent workload.

```
go build ./stores/utxo/...                       clean
go vet ./stores/utxo/pebble/...                  clean
go test -race -count=1 ./stores/utxo/pebble/...  ok, 5.733s
go test -count=1 -cover ./stores/utxo/pebble/... 83.2% of statements
```

The conformance suite really does run against the same `tests.Store` entry points as the other backends, and it passes under `-race`, which is meaningful for a store this young.

Two things I could not check. The pruner race is read from the code, not reproduced: it needs a reorg concurrent with a prune pass and I have no node running this store. And the coverage numbers disagree, `go test -cover` says 83.2% of statements while the SonarQube gate reports 60% on new code and fails against its 80% threshold. Somebody should work out which measurement is right before promotion, because one of them says the gate is already satisfied.

Inline comments below carry three smaller items: a swallowed storage error that leaves a preserve-index entry to be rescanned forever, an iterator lower bound that the resume path mutates while the iterator is live, and a `uint16` length prefix whose overflow would misalign the page-0 spend slots.

icellan (changes requested):
Reviewed with a focus on atomicity and parity with the Aerospike store.

**Writes are genuinely atomic** — one `pebble.Batch` per operation, `pebble.Sync`, index and master always in the same batch, and `SpendAndCreate` is a single batch, which is better than the shared sequential helper. The conflict-intent WAL ordering is correct.

The problems are all on the **isolation** side. The 1024-way stripe mutex is the only concurrency control and it is not re-entrant, so every precondition read (`s.Get`, `GetSpend`, the pruner's index scan, `getMaster`) is deliberately hoisted *above* the lock. Readers take no lock and no snapshot at all. That single decision produces five of the findings below. Aerospike gets record-level atomicity free from its Lua UDFs / native ops, which is why several of these are pebble-specific rather than model-level.

One thing to settle before anything else: `stores/utxo/factory/pebble.go` `init()` registers `pebble` unconditionally — no build tag, no feature flag, no settings entry, no docs. Despite the SPIKE doc comment this is selectable from plain config today.

## Blocking

1. **`stores/utxo/pebble/misc.go:609` — the pruner deletes revived records.** `prunerService.Prune` snapshots victim hashes from an unlocked DAH-index scan (`misc.go:589`) and then calls `stageDelete` unconditionally, never re-checking under the stripe lock that `deleteAtHeight` is still set and still due. Reproduced: stamp a tx `deleteAtHeight=150`, clear the master's DAH to 0 (the state a concurrent writer leaves), `Prune(200)` → deleted, `Get` returns `TX_NOT_FOUND`. The pruner runs once per block concurrently with block validation, and the per-victim fsync loop holds the window open for minutes. A reorg `Unspend`, `PreserveTransactions` (`mined.go:387`) or `SetLocked(true)` all clear the DAH inside it, and the record is destroyed with its live UTXOs. Every later tx spending it fails `ErrTxNotFound` permanently. `sql/pruner/pruner_service.go:144-147` makes the predicate part of the delete — one `DELETE ... WHERE delete_at_height <= $1`. Aggravating: unlike `aerospike/pruner_provider.go:35`, pebble's `GetPrunerService` ignores `utxostore_disableDAHCleaner`, so an operator cannot switch it off to recover.

2. **`stores/utxo/pebble/spend.go:359` — spending an OP_RETURN stamps live UTXOs for deletion.** `spentCount` is incremented for any page-0 slot write with no check that the slot was ever counted in `page0Count`. Reproduced: tx with outputs `[P2PKH, OP_RETURN, P2PKH]` mined at 100 → `outputCount=3 page0Count=2 spentCount=0`; one outpoint-only `Spend` of vout0 + vout1 → `spentCount=2 page0Count=2 dah=389`, while vout2 is a live unspent UTXO destroyed at height 389. Reachable because `SupportsOutpointOnlySpend() == true` (`pebble.go:118`), so `services/validator/Validator.go:1549` and `blockvalidation/quick_validate.go:1384` pass `SkipUTXOHashCheck`. The hash gate at `spend.go:322` is the only thing distinguishing a real UTXO slot from a data-output slot. Worse, `utils.go:277` sets `UTXOHash` to a non-nil **zero** hash on that path, which equals the zero hash stored for a non-spendable slot — so `spend.go:322` matches even with the check enabled. Second route: `pageFullySpent` (`spend.go:241-250`) counts raw non-nil slots against `spendableCount`, so one phantom write to a non-spendable slot in a mixed page donates a false completion credit and pushes `pagesSpent` to `pagesTotal` one real spend early. Aerospike stores a hash only when `ShouldStoreOutputAsUTXO` and `teranode.lua:194-208` nil-checks the slot, returning `UTXO_NOT_FOUND`; pebble additionally *accepts* a tx spending an unspendable output that Aerospike rejects. Aerospike returns `SupportsOutpointOnlySpend() == false`, so the right comparison here is SQL — which is also not vulnerable, because it stores a real `utxo_hash` row per output.

3. **`stores/utxo/pebble/pebble.go:102` — `Close()` panics the process instead of draining.** It flips a flag and calls `db.Close()` with no drain and no synchronisation, and no store method checks `s.closed` before touching `s.db`. pebble v2.1.6 `getInternal` (`db.go:563`), `applyInternal` (`db.go:835`) and `newIter` (`db.go:1061`) all begin `if err := d.closed.Load(); err != nil { panic(err) }`. On SIGTERM the UTXO store is closed while validator / block-validation goroutines are still inside `Get`, `BatchDecorate` or `batch.Commit`; the next call panics with `pebble: closed` and takes the process down, losing any staged-but-uncommitted batch. This violates `stores/utxo/Interface.go:386-402` verbatim: "implementations MUST wait for outstanding batched writes to complete before returning... Returning before pending writes have committed risks silently losing UTXO state". `Close(ctx)` also ignores `ctx`, though the contract says the context bounds the drain. Same class as the Aerospike client-level drain at `aerospike.go:738-780`.

4. **`stores/utxo/pebble/get.go:35` — reads are not isolated; half-applied batches are observable.** Every read API issues 3-6 independent `s.db.Get` / `NewIter` calls for one logical record with no stripe lock and no snapshot: `recordToMeta` reads `m|H` (`get.go:35`), then `p|H` (`get.go:56`), then `k|H|*` (`get.go:142`), then `g|H|1..N` (`get.go:227`) and `o|H|*` (`get.go:252`), each at a different sequence number. Torn spend: a reader sees page-0 slots empty, a `Spend` batch commits atomically writing both `m|P` and `g|P|1`, the reader then reads `g|P|1` and returns a `SpendingDatas` state that never existed. `GetCounterConflictingTxHashes` (`process_conflicting.go:1204`) builds its parent set from exactly that array, so the page-0 spender is never added to the counter-conflicting set and a double-spend becomes invisible to the conflict machinery. Torn delete, reproduced: with the master present and the payload key gone (the interleaving `Prune`/`Delete` creates), `Get` returns `STORAGE_ERROR (69): pebble: failed to read payload` with `errors.Is(err, ErrTxNotFound) == false`, so the validator's DAH-evicted-parent branch never fires and a valid tx is rejected as a hard store failure. Fix: hold `s.db.NewSnapshot()` across the whole read set.

5. **`stores/utxo/pebble/create.go:252` — nil-output panic on the validator hot path.** `for i, output := range tx.Outputs { outputItems[i] = output.Bytes() }` has no nil check, while the hash loop 55 lines earlier in the same function does (`create.go:197`). go-bt's `(*Output).Bytes()` dereferences the receiver unconditionally. Nil-padded output slices are an anticipated shape here: `services/utxopersister/UTXOSet.go:828` exists as `UnpadSlice` for it, `aerospike/create.go:1002` skips nil outputs citing "transaction outputs received before inputs", and `sql/sql.go:1322` does the same. Pebble is the only one of the three that dropped the guard.

6. **`stores/utxo/pebble/misc.go:527` — `ReAssignUTXO` silently no-ops.** It writes the override record with a bare `s.db.Set` and never sets `flagHasOverrides` on the master, but both readers gate override lookup on that flag (`spend.go:294`, `get.go:306`). Reproduced: `Create(tx, h, utxo.WithFrozen(true))` sets `flagFrozen` but not `flagHasOverrides`; `GetSpend` still reports FROZEN from the tx-level flag so `ReAssignUTXO`'s precondition passes and it returns nil — after which `flagHasOverrides` is still false and `GetSpend` still returns FROZEN. The reassigned hash and `spendableIn` are dead bytes: a court-ordered reassignment reports success while the output stays unspendable by the new owner forever. It only appears to work today because `FreezeUTXOs` happens to set the flag (`misc.go:404-408`, correctly, in one batch). Second route with no exotic setup: a concurrent `UnFreezeUTXOs` clears `flagHasOverrides` between `ReAssignUTXO`'s unlocked `GetSpend` and its write. Aerospike's reassign UDF (`teranode.lua:870-965`) rewrites the utxos entry in place so no flag can desync. Related: nothing in the package ever clears `flagFrozen`.

## Should fix before this is usable

7. **`pebble.go:56`** — `New()` passes `storeURL.Path` to `pebble.Open` verbatim, discarding `url.Host` and ignoring `tSettings.DataFolder`. `pebble://data/utxostore` parses to `Host="data", Path="/utxostore"`, so the store opens at filesystem root and the `data` component vanishes. The direct analogue of the committed idiom `utxostore.dev = sqlite:///utxostore` (`settings.conf:1314`) is `pebble:///utxostore`, which also resolves to `/utxostore`; `util/sql.go:220-223` joins `DataFolder` with the path. Under a root container this silently succeeds and writes the UTXO store to `/`. `conformance_test.go:20` masks it by overwriting `storeURL.Path = dir` after parsing, so no test exercises a host-bearing URL.

8. **`get.go:508`** — `BatchPreviousOutputsDecorate` calls `payloadBlobs()` once per *input* rather than once per parent, and `payloadBlobs` heap-copies the parent's entire payload (`getValue` does `append([]byte(nil), val...)`, `pebble.go:244`, and `create.go:255` packs inputs and outputs into one key). A parent with ~1.1M minimal outputs is a ~14MB payload; a child with 240k inputs spending it costs 240k Gets and ~3.4TB of memcpy from ~20MB of attacker data. Decoration at `quick_validate.go:1214` and `netsync/handle_block.go:1601` runs *before* script validation, so the scripts need not verify. The errgroup at `get.go:496` discards `gCtx` and the inner loop never checks ctx, so it is not cancellable and the liveness probe fires. `sql.go:3895-3944` deliberately groups by parent and dedups outpoints, with a comment about "whole-block-sized input sets".

9. **`mined.go:449`** — `ProcessExpiredPreservations` swallows every `getMaster` error with a bare `continue`, treating a corrupt master or an I/O error identically to a benign miss. `preserveUntil` is never cleared and `stageMaster` never called, so the `prefixPreserveIdx` entry survives and is re-read and re-failed on every block forever, while `dahEligible` (`spend.go:21`) will never let the tx be pruned. This is the only read-error path in the package that does not return — `mined.go:102` and `mined.go:379` both do.

10. **`misc.go:62`** — `SetConflicting` reads the tx meta via `s.Get` *before* taking the stripe lock and derives both the children edges and the returned BFS frontier from that stale snapshot. `MarkConflictingRecursively` (`process_conflicting.go:953`) drives its BFS purely from that second return value, so a spender committed inside the window is never appended to `spendingTxHashes` (`misc.go:136-140`), the BFS terminates early, and a conflicting parent is left with a non-conflicting child that stays eligible for block assembly. Aggravating: `SetConflicting` commits one batch per tx hash and returns on the first failure, so a partially-committed cascade is never unwound and those txs keep both `flagConflicting` and the `deleteAtHeight` stamped at `misc.go:86`.

11. **`get.go:275`** — a frozen override unconditionally *replaces* the recorded spender with the `FrozenBytes` sentinel in both readers (`get.go:275-278`, `get.go:332-335`) rather than merging, and `FreezeUTXOs` validates "not spent" in an unlocked pre-pass (`misc.go:373-391`) before taking the stripe in pass 2 (`misc.go:393-422`). If a spend commits between the passes the slot holds both the spender and a frozen override, the spender becomes permanently invisible to `GetCounterConflictingTxHashes` / `GetConflictingChildren`, and a later `UnFreezeUTXOs` resurrects it out of nowhere. Aerospike's freeze UDF does the spent-check and the write inside one server-side record operation and returns SPENT (`alert_system.go:131`).

12. **`get.go:387`** — `GetSpend` returns a hard `ProcessingError` for a vout beyond the tx's output count, which the asset handler turns into a 500. `aerospike/get.go:230-244` handles the identical input as `Status_NOT_FOUND` with a comment naming the crash it fixed: a caller passing an oversized vout via `/api/v1/utxo` and `/api/v1/utxos`, which no longer pre-validate vout, "would index past the slice and panic ... crashing the asset process when the panic surfaces in an errgroup goroutine outside echo's recover middleware." A pebble-backed node re-opens that hole for any unauthenticated caller.

13. **`get.go:437`** — `stageDelete` range-deletes the forward conflicting-children edges `k|hash|child` but never removes the paired reverse edges `K|child|hash`. Reproduced: create parent P and conflicting child C, `Delete(P)` → `forward=0 reverse=1`. The `DeleteRange` loop covers edges where `hash` is the parent; the iterator at `get.go:444-455` cleans the pair only where `hash` is the child. On a node that has processed many double-spends the `K|` keyspace grows monotonically with dead entries. `TestDeleteRemovesEveryKey` only asserts on master/page/hashes/payload/override, so the name overclaims and it cannot catch this.

14. **`pebble.go:92` and `codec.go:109`** — fixed-offset reads before length validation. `stored := binary.LittleEndian.Uint32(val)` on a 1-byte value under key `M` panics with `index out of range [3] with length 1`, on the `New()` startup path — a truncated meta key crash-loops the node at boot instead of returning the `ConfigurationError` the function exists to produce. `offsetBlobItem` validates only `i < n` where `n` is read from the blob's own first 4 bytes, then reads `blob[4+4*i:]` unchecked: a 4-byte blob claiming count=1000 panics, even though `codec.go:112-114` contains a corruption guard that is unreachable for the shortest truncations. Reached from `Get` / `BatchDecorate` / `PreviousOutputsDecorate` / `recordToUnmined`, so one bit-rotted payload value crash-loops the node with no way to skip past it. `decodeMaster`, `decodePage` and `decodeOverride` in the same file all bounds-check correctly — `offsetBlobItem` is the outlier.

15. **`spend.go:521`** — `Spend`/`SpendAndCreate` hold one stripe mutex per distinct parent across the entire read phase *and* the fsync'd commit. With `stripeOf = Uint16(hash[:2]) % 1024`, N random distinct parents holds `1024*(1-(1023/1024)^N)` stripes: N=3000 → 967 (94%), N=7000 → all of them. A ~300KB tx with 7000 inputs from distinct parents freezes every Create/Spend/Delete/SetMinedMulti/prune in the process for 7000 random point reads plus a WAL fsync — and the tx need not be valid, since `stageSpendGroup` errors are collected into `firstErr` while the loop keeps iterating every group (`spend.go:529-533`), so an unspendable tx pays the same lock cost and can be replayed to hold the store frozen. `PreserveTransactions` (`mined.go:369`) has the same shape over an uncapped mempool-sized parent set from `pruner_unmined.go:110`, once per block. Compounding: `Prune` commits one `pebble.Sync` batch per victim (`misc.go:602-615`) — ~1M serialised fsyncs per cycle, which the pruner can never keep up with; `SetConflicting` (`misc.go:112`), `FreezeUTXOs` (`misc.go:412`) and `UnFreezeUTXOs` (`misc.go:465`) repeat the one-item-one-fsync pattern.

## Smaller

- `go mod tidy` was not run: `cockroachdb/pebble/v2` is marked `// indirect` despite six direct importers, and go.sum is missing 12 lines.
- `codec.go:2` justifies the duplicated codec by citing `stores/utxo/packedsql`, which does not exist in the repo — the only grep hit is the comment itself.
- `BatchDecorate` has no nil-item guard and no ctx check; SQL has both.
- `conformance_test.go` omits `tests.MinedThenSpendAllPrunes` — the one shared guard for the DAH lifecycle pebble reimplements from scratch. It passes when wired up, but the pebble substitute `TestPrunerDeletesStampedTransactions` hand-writes `deleteAtHeight=150` and never calls `Spend()`, so nothing tests that `dahEligible`/`pagesSpent` actually *computes* the stamp — the exact machinery finding 2 breaks.
- No test drives concurrent goroutines through `lockStripes`, the store's only isolation mechanism; `-race` cannot catch a too-short critical section.
- No Close-and-reopen durability test that writes real UTXO data first. Pebble is the first backend whose state lives only on local disk between restarts, and `TestPageSizeIsImmutable` reopens an empty database.
- `spentCount`/`pagesSpent` are never read back after an Unspend/re-Spend cycle, so counter drift is invisible to the suite.
- `encodeMaster` truncates `len(blockRefs)` to uint16 — silent corruption past 5,461 block refs.
- `SetMinedMulti`'s `fullySpent` branch (`mined.go:62`) hand-copies `dahEligible` but drops its `flagConflicting == 0` guard; a duplicate hash within one `SetMinedMulti` call reads stale state and silently loses the first update.
- `pebble.Open` with a bare `&pebble.Options{}` gives every Store its own private 8MB block cache.

Things I chased and refuted, so nobody else spends time on them: the `append(it.lastKey, 0)` / `IterOptions.LowerBound` aliasing (pebble copies bounds at `NewIter`; a 2,500-record pagination probe showed no loss or duplication); `heightIndexIterator` truncating on a window of skipped rows (the loop caps output, not scan); `PreserveTransactions` skipping records with neither DAH nor `preserveUntil` (matches the deliberate gate in both `aerospike.go:1143` and `sql.go:5365`); `QueryOldUnminedTransactions`' 1000 cap (SQL has the same `LIMIT 1000`); the coinbase-maturity boundary (identical to `teranode.lua:326`); the `GetPrunerService` package singleton (copied verbatim from sql/aerospike); `old := *m` shallow-copy aliasing (only scalars are read back); `encodeMaster`/`decodeMaster` offsets (checked field by field, symmetric).

stores/utxo/pebble/get.go
```diff
@@ -0,0 +1,540 @@
...
+	if err := batch.Commit(s.sync); err != nil {
+		return errors.NewStorageError("pebble: failed to commit delete for %s", hash, err)
+	}
+
+	return nil
+}
+
+// DeleteComplete removes a transaction and all its associated data. stageDelete
+// already removes the master record, the payload, and every page, hash, override
+// and conflicting-child record in one atomic batch, so Delete leaves nothing
```
freemans13: `leaves nothing behind` holds for every keyspace here except one.

The range-delete loop below covers `k|hash|*`, the forward conflicting-child edges where this hash is the **parent**. The reverse iterator after it covers `K|hash|*`, the reverse edges where this hash is the **child**, and for each of those it cleans both directions. Nothing ever looks for `K|*|hash`. So deleting a parent removes its forward edges and leaves the matching reverse edge `K|child|parent` behind.

icellan reported that mechanism as finding 13 on 28 August, so the gap itself is not news. What is new is that this comment now asserts the opposite, on the one method whose interface contract turns on that exact phrase: "a SUCCESSFUL DeleteComplete leaves nothing behind" (`stores/utxo/Interface.go:416`).

The leak is bounded rather than corrupting. `prefixChildrenRev` is read in exactly one place, this function's own iterator, so an orphan `K|child|parent` misleads no reader and is swept when the child is itself deleted. It grows monotonically for as long as the child outlives the parent.

Two ways out. Either collect the children by iterating `k|hash|*` before the range delete and issue `batch.Delete(childrenRevKey(child, hash))` for each, or leave the code alone and change the comment to say the reverse edge survives and why that is acceptable. What should not stand is a comment claiming completeness that the code does not deliver.

freemans13 (changes requested):
Re-reviewed after today's push. The delta since the two blocking reviews is fifteen lines of store code, and exactly one of the open findings has been closed.

**What changed.** `DeleteComplete` was added, and the two `DeleteThenUnspendRestoresParent` conformance cases were wired into `conformance_test.go`. Both exist because a new method landed on the shared store interface upstream, so this reads as rebase compatibility work rather than a response to review. Separately `go mod tidy` was run: `cockroachdb/pebble/v2` now sits in the direct require block instead of being marked indirect, which closes the item liam, icellan and the review bot each raised.

**What is still open.** Everything else. There is still no build tag on `stores/utxo/factory/pebble.go`, so Pebble and its transitive tree, Sentry included, compile into every Teranode binary whether or not anyone selects the store. `New()` still hardcodes `pebble.Sync` with no way to reach `NoSync` from a store URL, so the headline benchmark row still describes a configuration an operator cannot choose. All fifteen numbered findings in icellan's 28 August review and all six of my inline comments from 24 August land on code that has not been touched since.

**Verified locally.** `go test -race -count=1 -run TestConformance ./stores/utxo/pebble/` passes in 4.8s on head 2ae5ec9, both new paginated cases included. Upstream CI was still running when I looked, so I am not quoting it as evidence either way.

One new finding, inline on `get.go`: the doc comment added above `DeleteComplete` claims the delete leaves nothing behind, and the reverse conflicting-child edge is the exception. Small on its own, but it now contradicts the interface contract in prose, which is worse than the leak.

Changes requested, unchanged.

icellan (changes requested):
Follow-up to my earlier review. A second independent pass reproduced the two data-loss findings from the first (the OP_RETURN spend count and the pruner deleting revived records) and turned up seven more that I had not reported. One of them is a constraint on the deployment model rather than a bug, and I would settle that one before spending effort on the rest.

## Deployment constraint — settle this first

**`stores/utxo/pebble/pebble.go:56` — Pebble takes an exclusive OS directory lock, so the store is single-process.** `LockDirectory` (pebble `open.go:129`, documented at `open.go:1166-1172` as "preventing another process from opening the database") means exactly one process may hold the store open. Teranode deploys `utxo.Store` consumers as separate processes: `compose/docker-compose-ss.yml` runs validator-1, blockassembly-1, subtreevalidation-1, blockvalidation-1, asset-1 and blockpersister-1 as distinct containers under one `SETTINGS_CONTEXT`, each resolving the same `utxostore` URL through `daemon/daemon_stores.go`. In that topology — and in the equivalent per-service k8s pod model — the second container dies inside `pebble.Open` with a file-lock error. Nothing warns the operator: `stores/utxo/factory/pebble.go:13` registers `availableDatabases["pebble"]` in a plain `init()`, while `stores/utxo/factory/aerospike.go` carries a commented-out `//go:build aerospike`, so the gating idiom exists here and was not used. `settings.conf` has no `pebble://` entry and `docs/references/settings/stores/utxo_settings.md:108-114` lists five schemes without it. This is not a bug to fix so much as a boundary to state explicitly and enforce.

Also on `New()`: `&pebble.Options{}` inherits `NoFilterPolicy` (`options.go:454`) — no bloom filters, so `Create`'s guaranteed-miss existence check at `create.go:54` pays full LSM read amplification — plus an 8MB cache, a 4MB memtable with `MemTableStopWritesThreshold=2`, and one compaction goroutine. `s.sync` is hardcoded to `pebble.Sync` with no URL or settings knob, and `validateMeta`, `BeginConflictIntent` and `CompleteConflictIntent` bypass it anyway.

## Additional blocking findings

1. **`stores/utxo/pebble/spend.go:160` — missing the "parent pruned, spending tx already blessed" fallback that both mature stores implement.** Aerospike (`stores/utxo/aerospike/spend.go:568-581`) and SQL (`stores/utxo/sql/sql.go:1954-1976`) both catch `ErrTxNotFound` on a spend, `Get` the *spending* tx, and clear the error if it already exists — "parent tx not found, but tx already exists in store, assuming already blessed". That is the normal state after a parent is DAH-pruned. Pebble has no such path: `getMaster` returns `ErrTxNotFound`, `setAll` stamps it on every spend in the group, and `Spend` returns `UTXO_ERROR -> TX_NOT_FOUND`. Reproduced side by side: `Create(ParentTx)`, `SpendAndCreate(child)`, `Delete(ParentTx)` (= pruned), then re-validate the same child — sqlitememory returns `TX_EXISTS`, which the validator treats as already-processed; pebble returns `UTXO_ERROR (79): pebble: spend failed -> TX_NOT_FOUND (30)`. In production that is block validation or catchup re-validating an already-blessed tx: the tx fails permanently, the subtree never validates, the block never finalises, and the node wedges at that height — the exact wedge class the Aerospike fallback was written for.

2. **`stores/utxo/pebble/spend.go:470` — `Unspend(spends, false)` is a silent no-op, so conflict-demoted parents stay locked forever.** The tri-state `flagAsLocked` is collapsed to a bool and the write is `if setLocked { m.flags |= flagLocked }` with no else. `utxo.ProcessConflicting` locks affected parents via `s.Unspend(ctx, affectedParentSpends, true)` (`stores/utxo/process_conflicting.go:298`) and `utxo.ReverseProcessConflicting` unlocks them via `s.Unspend(ctx, demotedSpends, false)` (`process_conflicting.go:499`). SQL implements exactly that contract — `flagLockedProvided := len(flagAsLocked) > 0` then `UPDATE transactions SET locked = $2` (`stores/utxo/sql/sql.go:3044-3095`) — with a comment naming those two callers. Pebble can apply the lock but never release it, so a reorg that demotes a winner leaves the parents `flagLocked` and every subsequent spend of them fails `TxLockedError` at `spend.go:177`; block assembly wedges on the restored chain. Note the three backends now disagree three ways — Aerospike's `Unspend` accepts `flagAsLocked` and ignores it entirely (`stores/utxo/aerospike/un_spend.go:98-130`) — and pebble picked the only asymmetric option. Nothing in `stores/utxo/tests` pins the locked flag after `Unspend`, which is why the conformance suite is green.

3. **`stores/utxo/pebble/mined.go:34`, `misc.go:186`, `mined.go:101` — one missing hash discards the update for every other transaction in the batch.** `SetMinedMulti`, `SetLocked` and `MarkTransactionsOnLongestChain` all return on the first `ErrTxNotFound` *before* `batch.Commit`. `BlockAssembler.unlockConflictParents` (`services/blockassembly/BlockAssembler.go:1367`) builds the union of the parents of every winner and calls `SetLocked(ctx, parents, false)`; its own code at line 1342 acknowledges records may be pruned. One pruned parent means none of the others get unlocked, and `utxo.setLockedWithRetry` (`process_conflicting.go:912`) then burns every backoff retry on the same deterministic failure, so the step-5 unlock never lands. `fixUnminedSinceInconsistencies` (`BlockAssembler.go:2687`) passes thousands of hashes from a non-snapshot-isolated scan; one tx pruned between scan and mark writes zero records and errors out at startup, leaving every genuinely inconsistent record with its stale `unminedSince`. And `SetMinedMultiChunked` (`stores/utxo/set_mined_multi_chunked.go:53`) aborts the errgroup on the first chunk error, so one stale hash in a fat block leaves thousands of txs stuck unmined. SQL absorbs missing rows naturally (`UPDATE ... WHERE hash = $1` matching zero rows); Aerospike applies the per-record UDFs that succeeded and reports only the failures (`aerospike/set_mined.go:509-515`, `longest_chain.go:120-140`).

4. **`stores/utxo/pebble/mined.go:155` — `heightIndexIterator` resumes from the height-index key rather than the txid, so a reorg emits a transaction twice or drops it.** The cursor is `prefix||BE(unminedSince)||txid` and each `Next()` opens a fresh unsnapshotted iterator. `SetMinedMulti{UnsetMined}` (`mined.go:47`) and `MarkTransactionsOnLongestChain(false)` (`mined.go:111`) *move* a row within the `u` index — the ordinary reorg path. Verified both directions on the real store: a tx already emitted at `unminedSince=100` and re-keyed forward to 500 is emitted a second time by the next `Next()`, and `BlockAssembler.loadUnminedTransactions` then adds the same tx twice to the subtree processor, producing a duplicate in the block candidate; a not-yet-visited tx at 300 re-keyed backwards to 150, behind the cursor, is never emitted at all and is silently dropped from block assembly's mempool with no error. Aerospike keys its scan by txid partition, so a bin update cannot move a record relative to the cursor. Separately, at `mined.go:189-192`, if `Prune` or `Delete` commits between reading the index key and the `getMaster` point lookup, `it.err` is latched and every later `Next()` returns that error forever — `loadUnminedTransactions` turns it into "error getting unmined transaction" and aborts, so block assembly cannot start or reset while the pruner runs.

5. **`stores/utxo/pebble/get.go:414` — zombie DAH index entries, re-processed and re-counted on every block forever.** `stageDelete` returns nil when the master is already gone *without* deleting the DAH index key that led the pruner there, and `Prune` (`misc.go:622`) does `deleted++` regardless. Reproduced: two DAH index entries for one tx (heights 140 and 150) → `Prune(200)` reports `deleted=2` for a single transaction and leaves one entry behind; a second `Prune(200)` reports `deleted=1` having deleted nothing, and will keep doing so on every block. The finding-2 generator is concrete: `stageDelete` removes the index key at the master's *current* `deleteAtHeight`, so when `SetMinedMulti`'s DAH-extension branch moves the DAH after the pruner's scan, the pruner deletes the new key and orphans the scanned one. `p.observers`' `OnPruneComplete(height, deleted)` drives the pruning metrics and backpressure, so those counts are permanently inflated by phantom deletions, and each zombie costs a `getMaster`, a stripe-lock acquisition and an fsynced empty batch commit per block. Fix: delete the scanned index key explicitly in the `Prune` loop rather than deriving it from the master, and count a deletion only when a master was actually removed.

6. **`stores/utxo/pebble/conformance_test.go:14` — every test in the package runs at block height 0.** `newTestStore` never calls `SetBlockHeight` or `SetBlockState`, so all 26 wired conformance subtests, all of `store_test.go` and both benchmarks run with `GetBlockHeight() == 0`. Six height-keyed branches are only ever exercised at zero: `SetConflicting`'s DAH stamp (`misc.go:86`), `SetLocked(false)`'s re-stamp (`misc.go:199`), `MarkTransactionsOnLongestChain`'s `unminedSince` (`mined.go:111`), `SetMinedMulti`'s `UnsetMined` `unminedSince` (`mined.go:46`, always exactly 1), `ReAssignUTXO`'s `spendableIn` (`misc.go:523`), and `GetSpend`'s `CalculateUtxoStatus` plus its IMMATURE comparison (`get.go:330, 345`). A wrong-direction comparison in any of them passes today. It also masks a live defect: this store overloads `unminedSince == 0` as the "mined" sentinel — `stageMaster` writes the unmined index only when `> 0`, `recordToMeta` sets `UnminedSince` only when `> 0`, `dahEligible` requires `== 0` — where SQL and Aerospike use NULL and can represent height 0. So `MarkTransactionsOnLongestChain(false)` or `Create(tx, 0)` executed before the first FSM height notification silently records a transaction as mined-and-on-longest-chain with no block reference, invisible to `GetUnminedTxIterator`, `GetPrunableUnminedTxIterator` and `QueryOldUnminedTransactions`, and now prunable. A reorg during startup drops those transactions from block assembly permanently. Adding `require.NoError(t, store.SetBlockHeight(N))` to `newTestStore` for a realistic N turns two findings in this review into failing assertions.

## Test coverage is the root cause of several of these

`conformance_test.go` wires 26 shared cases but omits `tests.MinedThenSpendAllPrunes` — the DAH/prune lifecycle test that Aerospike and both SQL backends run — and `tests.Sanity`. Pebble is also absent from `tests.GetStoreTestCases()`, `TestGetAndLockChildren` and the `TestUnminedTxIterator*` registries. `TestPrunerDeletesStampedTransactions` hand-stamps `deleteAtHeight` via `stageMaster`, so the production DAH path is never executed — the store could set DAH nowhere at all and that test would still pass. `MinedThenSpendAllPrunes` passes when wired up, so this is pure wiring, not a blocked test.

## Corrections to my earlier review

`SetConflicting(value=false)` never removing the children/childrenRev edges written on `value=true` is real and reproduced, but it is **pre-existing and repo-wide**, not introduced here: Aerospike's `updateParentConflictingChildren` also only ever adds, and `RemoveFromConflictingChildren` has no caller anywhere in `process_conflicting.go` or `services/`. Not this PR's problem.

Two further correctness items I am reporting but deliberately not ranking as blocking: `get.go:157` — requesting `fields.Inputs` without `fields.Tx` returns `Data.Tx == nil`, so `BlockAssembler.validateUnminedTxInputs` rejects every unmined tx and `reset(ctx, reorgFailed)` discards the mempool; Aerospike sets `Data.Tx` for this case, SQL does not, and the new store copied the broken sibling, with the caller's own comment at `BlockAssembler.go:1861-1865` documenting it as broken. And `mined.go:304` — `consistencyIterator.scanned` is a plain `int64` raced by `fixUnminedSinceInconsistencies`' progress goroutine, visible under `-race`; Aerospike uses `atomic.Int64`. Neither iterator ever consults `ctx` despite the interface doc requiring it, and `Close()` writes `it.done` while `Next` reads it.

## Reuse

`utxo.BlockStateFields` (`stores/utxo/block_state.go:93`, embedded by aerospike, sql and nullstore, extracted specifically to fix the torn height/median-time pair) is reimplemented here as 54 lines. `spend.SpendingData.Bytes()`, the conflict-WAL intent codec (a third byte-identical copy), and hand-rolled equivalents of `binary.{Big,Little}Endian` and `slices.Contains` are all duplicated. `misc.go` is a 639-line grab-bag, the filename `docs/references/codingConventions.md:119` names as an anti-pattern. `Health()` never touches the DB — SQL runs `SELECT 1` — so a corrupt store reports 200. The `packedsql` package the codec header cites as its in-flight sibling does not exist in the tree or in git history; the repo corrected that same pattern in b66c0a57c.

## Performance, if this proceeds toward promotion

Every spend re-encodes the whole master (172-2404 bytes to change 36). `SetMinedMulti` saturates all 1024 stripes for a large block in one unbounded batch — a store-wide stall. `BatchDecorate` is a sequential `Get` loop defaulting to the heaviest field set (`MetaFieldsWithTx`, against Aerospike's `defaultDecorateBins`) and never clears `item.Data` on error, breaking the `Data == nil means not-found` contract. `stageSpendSlot` re-reads the hashes page per input instead of caching it the way `pages` is cached. `bench_test.go`'s working set is a single key that never leaves the memtable, so the benchmark is structurally incapable of observing any of this.

icellan (approved):
Re-reviewed at `f1c641b57`. I verified each previously-flagged area against the current code rather than the commit messages. The fix-up commits land, so this supersedes my two earlier change requests.

**Snapshot isolation.** Every multi-key read path — `Get`, `GetSpend`, `decorateOne`, `stageDelete`, `BatchPreviousOutputsDecorate`, and both iterators — now assembles its logical record from a single `*pebble.Snapshot`. The torn-read class is closed, including the `GetCounterConflictingTxHashes` vector.

**Close guard.** `enter()`/`leave()` wraps every DB touch, and the guard is held for the lifetime of iterators and snapshots. I specifically audited for a nested `enter()`, which would deadlock against a waiting `Close`, and found none — the inner read helpers all take a `reader` and call `readValue` directly. Lock order is consistently stripes → close-guard, never the reverse.

**Stripe coverage.** All 14 `stageMaster`/`stageCreate`/`stageDelete` call sites hold the relevant stripe, including the parents' stripes for the conflicting-child edge writes in `Create` and `SpendAndCreate`.

**Reverse conflicting-child edges.** `get.go:486-515` clears both directions now.

Build tag on the factory registration and the `?sync=false` knob are both as described. Codec offsets round-trip (`masterFixedLen` 88, encode/decode field order matches), key-prefix bounds do not collide, all slice-to-array conversions are length-safe, and `dahEligible`/`pagesSpent`/`spentCount` bookkeeping is consistent across spend, unspend, mine and preserve. I ran the conformance suite plus a concurrent Spend/Get/SetMinedMulti/Prune stress under `-race`: clean.

Four things I suspected were divergences turned out to match `sql`/`aerospike` exactly, so they are deliberate and I am not reporting them: `GetSpend` on a data output, `RemoveBlockIDs` not restoring `unminedSince`, `PreserveTransactions` skipping txs with no DAH, and the unconditional lock-clear in `SetMinedMulti`.

Three minor items left, none blocking.

**`stores/utxo/pebble/conformance_test.go:14` — medium.** `newTestStore` uses a bare `settings.NewSettings()`, so it inherits whatever `pruner_utxoDefensiveEnabled` the developer's or CI context has set. `GetPrunerService` (`misc.go:631`) deliberately hard-errors on that flag, so `TestPrunerDeletesStampedTransactions` (`store_test.go:565`) and `TestConformance/mined_then_spend_all_prunes` fail on any machine where it is on. Reproduced here: `go test -race ./stores/utxo/pebble/...` fails with `CONFIGURATION (5): pebble: pruner_utxoDefensiveEnabled is not supported by the pebble utxo store`, and passes with the flag off. Pin `tSettings.Pruner.UTXODefensiveEnabled = false` in `newTestStore`.

**`stores/utxo/pebble/mined.go:53` — low.** `unminedSince` is derived three different ways: `GetBlockHeight()+1` here, `GetBlockHeight()` in `MarkTransactionsOnLongestChain` (`mined.go:120`), and `blockHeight` in `buildTx` (`create.go:239`). Past the off-by-one, the two that omit the `+1` write `unminedSince = 0` when the store's cached height is still 0, before the blockchain client has called `SetBlockHeight`. Zero is the codec's "mined" sentinel, so such a record is never emitted by `GetUnminedTxIterator` or `QueryOldUnminedTransactions`, and `ScanInconsistentUnminedTxs` will not catch it either since it only looks for `unminedSince > 0 && len(blockRefs) > 0`. Narrow window, permanent effect on the record. Pick one convention.

**`stores/utxo/pebble/create.go:259` — low.** This is my earlier nil-output finding, and the fix went in the wrong direction. The comment claims outputs can arrive nil-padded and the payload loop must tolerate it, but `buildTx` calls `util.TxMetaDataFromTx`/`TxMetaDataFromTxNoFee` at lines 130-132, *before* the loop, and both panic on a nil output (`util.GetFees` → `fees.go:43`, and `(*bt.Tx).Size` → `output.go:114`; confirmed both with a probe test). So the guard is unreachable. It would not be a working defence even if it were reached: `packOffsetBlob` writes a zero-length entry for the nil output, and `(*bt.Output).ReadFrom` on that returns `satoshis(8): got 0 bytes: EOF`, so `Create` succeeds and every later `Get(fields.Tx|Outputs)` and `PreviousOutputsDecorate` on that tx fails permanently. Either drop the dead guard and its comment, or reject the create explicitly.

stores/utxo/pebble/spend.go
```diff
@@ -0,0 +1,595 @@
...
+}
+
+func (s *Store) stageSpendGroup(batch *pebble.Batch, g *parentGroup, blockHeight uint32, ig utxo.IgnoreFlags) error {
+	m, err := s.getMaster(&g.hash)
+	if err != nil {
+		for _, sp := range g.spends {
+			sp.Err = err
+		}
+
+		return err
```
freemans13, line 171: Aerospike and SQL both have a fallback at this point that Pebble lacks. When the parent is not found, they check whether the *spending* transaction already exists. If it does, they clear the error, on the reasoning that the parent was pruned after the child had already been accepted (`aerospike/spend.go:568-581`, `sql/sql.go:1952-1976`).

Without that fallback, `SpendAndCreate` on an existing transaction whose parent has been pruned returns `UTXO_ERROR -> TX_NOT_FOUND` here, where both other stores return `TX_EXISTS`. The validator tests `ErrUtxoError` before its own `ErrTxNotFound` fallback (`Validator.go:1121`, `:1199`), so a valid resubmit is rejected.

Suggested shape: in `Spend` and `SpendAndCreate`, when a group fails with `ErrTxNotFound`, look up the spending transaction's master once. If it is present, clear `sp.Err` for that group and move on. `SpendAndCreate` then reaches `stageCreate`, which returns `ErrTxExists`, and that is the answer the validator already handles.

freemans13 (approved):
Re-reviewed at `f1c641b57`, and approving. Every finding from my two change requests is closed, and I checked each one against the code rather than the commit messages.

**My findings, now closed**

- The pruner rereads the master record under the stripe lock and skips a transaction whose delete-at-height was cleared or moved after the scan, so a reorg that un-mines a transaction mid-pass now wins (`misc.go:695-716`).
- `pruner_utxoDefensiveEnabled` now fails the pruner service loudly instead of being ignored, and `utxostore_disableDAHCleaner` returns no service, the same as the Aerospike provider (`misc.go:618-635`). `services/pruner/server.go:122-128` turns both into a startup error, so neither is silent.
- `Create`, `SpendAndCreate` and `SetConflicting` all lock the parents' stripes before writing conflicting-child keys under them. `SetConflicting` also commits one batch now rather than one per hash.
- `stageDelete` removes both directions of the conflicting-child edges, whether the deleted hash is the parent or the child, so the `DeleteComplete` doc comment is now true (`get.go:449-524`).
- `ProcessExpiredPreservations` returns read errors instead of skipping them. Both iterators copy their resume bound. `encodeMaster` refuses a block-refs length that would overflow its uint16 prefix.
- The factory registration sits behind `//go:build pebble`, and `?sync=false` reaches `NoSync` with a warning.

**One divergence still reproduces. Not blocking, but worth fixing here**

icellan's 9 September finding about a pruned parent still holds at this head. Say a transaction is already in the store and its parent has since been deleted by the pruner. If that transaction is submitted again, Aerospike and SQL both see that the spending transaction exists and answer `TX_EXISTS`, which the validator treats as an idempotent success. Pebble fails in the spend phase with `UTXO_ERROR -> TX_NOT_FOUND` and never reaches the create. The validator's own fallback for a pruned parent doesn't catch it either, because the `ErrUtxoError` branch is tested first and the `ErrTxNotFound` branch is its `else` (`Validator.go:1121`, `:1199`). So the resubmit is rejected. Same sequence run on both stores at this head:

```
sqlitememory: TX_EXISTS (33): Transaction already exists in sqlite store
pebble:       UTXO_ERROR (79): pebble: spend failed -> TX_NOT_FOUND (30): pebble: transaction d580e67e… not found
```

I confirmed the rejection on the validator path. I did not trace whether block or subtree validation can reach it, which is the node-wedge case icellan described, so I'm not holding the approval on it. The fix is about ten lines mirroring `aerospike/spend.go:568-581`. More detail inline.

**Two I rechecked and am dropping**

- `Unspend(spends, false)` still never clears the lock flag, so Pebble differs from SQL here. Both production callers either run `SetLocked(false)` afterwards or act on parents that `ProcessConflicting` already unlocked at its step 5 (`process_conflicting.go:354`, `:887`). Aerospike ignores the flag in both directions. I found no path that leaves a parent locked unless that step-5 unlock itself failed.
- `SetLocked` giving up on the whole batch when one hash is missing matches sqlitememory, which also errors and leaves the present hash locked.

**Verified locally at `f1c641b57`**

```
go build ./stores/utxo/...                        clean
go build -tags pebble ./stores/utxo/factory/...   clean
go vet ./stores/utxo/pebble/...                   clean
go test -race -count=1 ./stores/utxo/pebble/...   ok  7.0s
go test -count=1 -cover ./stores/utxo/pebble/     83.6% of statements
```

Upstream CI shows 37 passing and one skipped (the Aerospike fork-image job). I ran the probe through `go test -overlay`, so nothing was added to the branch. My settings context has `pruner_utxoDefensiveEnabled` off, which is why the race run passed here and failed for icellan. Their note about pinning that flag in `newTestStore` still stands. I tested with no live node and no concurrent workload beyond the package's own stress tests.

PR #1605, opened 2026-08-18
fix(p2p): report the height a peer advertised, not the sync cap

ordishs:
## Problem

`sanitizeAdvertisedTip` caps a peer's advertised height at `localHeight + p2p_max_unvalidated_advertised_height_lead` (default 10000) so an unvalidated claim cannot steer catchup. That cap is correct and stays.

The capped value was also used for operator-facing telemetry. A node still catching up therefore reported **every peer at `localHeight + 10000`** — a number no peer ever sent — shown next to that peer's genuine tip hash and chainwork.

Observed on a node syncing mainnet from genesis, with peers correctly advertising 962710:

```
[sanitizeAdvertisedTip] capped advertised height for peer 12D3KooWA4EX...
from 962710 to 104791 (local height 94791, max lead 10000)
```

That warning fired 6,504 times in one run. Every peer's displayed height tracked our own catch-up progress in lockstep:

| Time | Local node | Peers shown as |
|---|---|---|
| 12:08:18 | 94,326 | 104,361 |
| 12:08:38 | 94,490 | 104,494 |
| 12:08:48 | 94,567 | 104,569 |

Neighbouring fields were never capped, so each row contradicted itself: the real tip hash `00000000...bf798a1b`, the real chainwork, and `sync_peer_height: 962710` all sat beside a fabricated height.

## Change

Keep the cap on the sync-decision path; report the raw claim wherever an operator reads it.

- `node_status` and `block` gossip notifications now carry the advertised height. The peer registry still receives the capped value.
- `PeerRegistryInfo` gains `advertised_height` in both `blockchain_api` and `p2p_api`, carried through the registry, the p2p client, and the asset peers response.
- The dashboard peers page shows the advertised height, with the capped sync height in the tooltip. Its sort follows the displayed value rather than the local ceiling every peer shares during catchup.
- Corrected the setting's own `longdesc`, which claimed the cap governed "peer displays".

Peer selection, catchup probing, and the `ListPeers` height filter all still read the capped `Height`, so the protection added in #1201 is unchanged. Older peers that do not send the field fall back to the existing value.

## Testing

New coverage at each hop: registry storage (raw vs capped), batcher merge including the capped-tie case, `blockchain_api` proto round-trip, `p2p_api` conversion, and the p2p client conversion the peers endpoint consumes.

```
make test        12746 tests, 49 skipped, 1 failure
go test -race ./services/p2p/... ./services/blockchain/... ./services/asset/...   all ok
npm run check    800 FILES 0 ERRORS
npm run build    done
npm run test     40 passed (40) + 16 passed
golangci-lint    2 prealloc issues, both in untouched files
```

The single failure is `daemon TestDaemon_Start_AllServices` ("no valid listen address"). It is pre-existing and environmental: it reproduces identically on unmodified `main`, and is caused by a local node already holding port 8333. It should pass in CI.

## Notes for reviewers

- `make gen` regenerates every proto. The local `protoc` (33.2) is older than the one that produced some checked-in files (34.1), which rewrites version stamps across ~21 unrelated `.pb.go` files. Only the two protos changed here were regenerated; there is no stamp drift in the diff. Worth pinning protoc separately.
- `services/asset/httpimpl/swagger.json` is stale on `main` — regenerating it also pulls in `/api/v1/utxos` and `/api/v1/utxos/hex`, which exist in code but never reached the spec. That unrelated drift is deliberately **not** bundled here; only the one new field was added, verified identical to the generator's output for that model. A separate `make swagger-asset` run would close the gap.


gokutheengineer (commented):
The core change is correct and safely scoped: I traced `AdvertisedHeight` through the whole tree and it reaches only telemetry surfaces (registry storage, proto conversions, asset endpoint, dashboard). Sync decisions (sync coordinator, peer selector, catchup, `ListPeers` filter) all still read the capped `Height`, and the only non-p2p consumers of `node_status` use the hash or just relay. Proto field numbers and the rolling-upgrade fallback check out.

That said, there is one real bug in the new merge logic, one removed safety bound that needs an explicit decision, and a few surfaces where the symptom this PR fixes still shows.

### Must fix

**1. Tie-break pairs the surviving height with the losing observation's hash** - `services/p2p/peer_registry_batcher.go:97`

In `merge()`'s capped-tie branch, only the `advertisedHeight` update is guarded by "higher raw claim wins"; the `blockHash` overwrite below it runs unconditionally. When the incoming observation loses the tie-break, its hash still replaces the winner's, so the registry stores a height/hash pair no peer ever sent - the exact mismatch this PR's description complains about, one layer down. The branch's own comment says "take the paired hash with it", which the code doesn't do. Merge order across concurrent topic workers is nondeterministic, so this is not a theoretical corner. The new tie test uses nil hashes and cannot catch it.

Fix: move the hash update inside the `from.advertisedHeight > u.advertisedHeight` guard (keeping take-newest when the raw claims are equal), and extend the tie test with distinct hashes.

**2. The displayed height is now unbounded - decide and document** - `ui/dashboard/src/routes/peers/+page.svelte:246`

`sanitizeAdvertisedTip` rejects only an unparseable hash. A peer gossiping `BestHeight=4294967295` with any valid-format hash passes, gets no ban score, and its raw claim flows uncapped through the registry, the asset endpoint, and the page's headline value and sort key (`(row.advertised_height ?? 0) || (row.height ?? 0)`). Pre-PR the displayed value was bounded at local+10000; this PR removes that bound with no mitigation, so a single malicious peer can permanently pin the top of every operator's height-sorted table. If that's accepted risk, say so in the PR; otherwise add a display-side plausibility ceiling or a ban score for absurd claims.

### Same symptom, unfixed surfaces

**3. `getpeerinfo` still reports the cap** - `services/rpc/handlers.go:1311`

`CurrentHeight` and `StartingHeight` both come from `p.Height`, so during catchup the RPC shows every peer at localHeight+10000 while the dashboard shows raw claims for the same peers. The conversion this code reads already carries `AdvertisedHeight` after this PR (`services/p2p/Client.go` `convertFromAPIPeerInfo`), so it's a one-field fix.

**4. Peer detail modal disagrees with its own table row** - `ui/dashboard/src/routes/peers/+page.svelte:1027`

The modal's "Height" metric renders `selectedPeer.height` (capped). During catchup, a row headlines advertised 962,710 and clicking it opens a modal showing 104,791 - the same self-contradicting display this PR exists to remove, now within one page.

**5. Fallback capped height is labeled "Advertised by peer"** - `ui/dashboard/src/routes/peers/+page.svelte:543`

The tooltip is built from `display` (the fallback-resolved value), not from what the peer advertised. When `advertised_height` is 0 (older peer version, mixed-version rolling upgrade), the capped registry height is shown with the tooltip asserting "Advertised by peer: 10,100" - presenting the locally derived cap as the peer's claim.


`sanitizeAdvertisedTip` still emits an unconditional per-message `Warnf` on every cap. Under this PR's own framing (the cap is "a bound on what an unvalidated claim may influence, not a statement about the peer"), a node syncing from genesis warning on essentially every gossip message for hours is the wrong level. Should be `Debugf` or a rate-limited / once-per-peer warn.

### Minor / consistency

**7. Registry is last-write-wins on `AdvertisedHeight`** - `services/blockchain/peer_registry.go:315`

`Register()` overwrites on any value > 0, so the batcher's tested "capped tie must not regress" invariant only holds within one 1s flush window: a higher claim flushed in window N is overwritten by a lower one arriving alone in window N+1, and the displayed height ticks down then up during catchup. This mirrors the registry's pre-existing newest-wins `Height` semantics and self-heals within seconds, so it may be intended - but the batcher and registry currently encode contradictory policies for the same question. Worth a comment either way.

**8. Adjacent merge branches apply opposite survival policies** - `services/p2p/peer_registry_batcher.go:91`

The `from.height > u.height` branch overwrites `advertisedHeight` unconditionally, so if the cap ceiling advances mid-window, a stale lower raw claim sanitized against the newer ceiling beats a fresher higher claim - the opposite of the tie branch three lines below. Narrow and self-healing, but the two branches answer "which raw claim survives" differently. (I checked the zeroing hazard: no caller can produce height > 0 with advertisedHeight = 0, so there is no zero-overwrite path.)

**9. Two adjacent swappable positional `uint32`s across four signatures** - `services/p2p/server_helpers.go:631`

`registerPeer` / `addPeer` / `addConnectedPeer` / `enqueueRegister` now take `height, advertisedHeight`; a transposed call compiles cleanly and silently swaps the sync and display values. The batcher already has a struct-based `enqueue(peerID, *pendingPeerUpdate)` one layer down; passing a struct through would also have avoided touching ~30 test call sites, churn that repeats on the next field.

**10. Dead raw-height initializers** - `services/p2p/server_helpers.go:82` and `services/p2p/Server.go` (`sanitizedBestHeight`)

Both raw initializers are dead on every path that reaches the registry (the `isSelf` path returns before `addPeer`; the node_status path is gated on `sanitizedTipOK`), but the raw-initialized form implies an uncapped-to-registry path exists and forces an auditor to trace the guard chains to prove otherwise. Zero-value declarations (`var registryHeight uint32`) would make "registry height is always sanitized" true by construction.

gokutheengineer (approved):

ctnguyen (changes requested):
Review at commit [0befc61](https://github.com/ordishs/teranode/commit/0befc614b3ff369e70942fdb7e8f9c857354354a)

This is a well-diagnosed bug with a well-chosen fix. Splitting "what the peer claimed" from "what we let an unvalidated claim influence" is the right shape — the cap from #1201 stays exactly where it belongs, and I traced `AdvertisedHeight` end to end and it really does only land on telemetry surfaces: the sync coordinator, `PeerSelector`, catchup and the `ListPeers` filter all still read the capped `Height`. The field-number choices are additive, the older-peer fallback is thought through, and the test coverage is unusually good for a telemetry change — a proto round-trip assertion at each hop plus the two merge branches, each with a comment saying *why* the assertion exists. The dashboard sort following the displayed value rather than the shared local ceiling is a nice catch that most people would have missed. Below: one real defect in the new merge branch, one safety bound that this PR removes without replacing, and a handful of surfaces where the symptom the PR fixes is still visible (some inside the diff, some adjacent).

## Blocking issues

### ChiR1 — Capped-tie merge keeps the winning advertised height but the losing observation's hash

*Same as [@gokutheengineer's point 1 (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1605#pullrequestreview-4960959221).*

**Problem:** In `services/p2p/peer_registry_batcher.go:95-105`, the capped-tie branch guards the `advertisedHeight` update with "higher raw claim wins" but leaves the `blockHash` overwrite ungated:

```go
} else if from.height == u.height {
    if from.advertisedHeight > u.advertisedHeight {
        u.advertisedHeight = from.advertisedHeight
    }
    if from.blockHash != nil {   // runs even when `from` lost the tie-break
        u.blockHash = from.blockHash
    }
}
```

When the incoming observation *loses* on the raw claim, its hash still replaces the winner's, so the registry ends up holding `AdvertisedHeight` from observation A next to `BlockHash` from observation B. The branch comment says "take the paired hash with it", which is precisely what the code does not do. Extending @gokutheengineer's point: the pre-PR code was deliberately newest-hash-wins on a height tie, and there was no second height field to disagree with it, so the mispairing is new with this diff — and the fix has to state a policy for the `from.advertisedHeight == u.advertisedHeight` case too, otherwise the tie-hash behaviour silently changes for peers on older versions where `advertisedHeight` is always 0 and every tie is an equality.

**Why it matters:** This reproduces the exact defect the PR exists to remove — a height shown next to a hash the peer never paired it with — one layer further down, and now inside the persisted registry rather than just a rendered cell. Concurrent topic workers make the merge order nondeterministic (the function's own doc comment says so), so it is reachable in normal operation, not a corner case. The new tie test passes `nil` hashes for both observations and therefore cannot fail on it.

**Fix:** Move the hash update inside the raw-claim guard and keep newest-wins only when the raw claims are equal, then give the tie test distinct hashes.

```go
} else if from.height == u.height {
    // Same capped height: the capped value cannot break the tie, so prefer
    // the higher raw claim and carry the hash it arrived with.
    if from.advertisedHeight > u.advertisedHeight {
        u.advertisedHeight = from.advertisedHeight
        if from.blockHash != nil {
            u.blockHash = from.blockHash
        }
    } else if from.advertisedHeight == u.advertisedHeight && from.blockHash != nil {
        u.blockHash = from.blockHash
    }
}
```

### ChiR2 — The displayed and sorted height is now attacker-controlled with no bound

*Same as [@gokutheengineer's point 2 (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1605#pullrequestreview-4960959221).*

**Problem:** `sanitizeAdvertisedTip` (`services/p2p/server_helpers.go:580-604`) rejects only an unparseable hash; an absurd height is capped and logged, never scored. `sanitizeNodeStatusMessage` (`services/p2p/sanitize.go:118-132`) bounds only strings. So a peer gossiping `BestHeight = 4294967295` with any well-formed hash returns `ok == true`, and `Server.go:1292` then hands the raw value straight to `addPeer` as `advertisedHeight`. From there it reaches the registry, the peers endpoint, and the page's headline value and sort key (`+page.svelte:246`). Pre-PR every operator-visible height was bounded at `local + 10000`; this PR removes that bound and adds nothing in its place.

Two refinements on top of @gokutheengineer's version. First, the WebSocket side of this is not new — `node_status` already relayed `ChainWork` and `SyncPeerHeight` unbounded (`Server.go:1266-1268`), which is exactly why the PR description could quote a truthful `sync_peer_height: 962710`. What *is* new is that the raw claim now reaches the **registry**. Second, `PeerInfo` has no `json:"-"` tags and `peer_registry_persistence.go:28,68` marshals `[]*PeerInfo` wholesale, so the bogus value is written to the operator's persisted peer blob and survives a restart until that peer's next message.

**Why it matters:** A single peer, at zero cost and with no ban-score consequence, permanently pins the top of every height-sorted peer table on every node that hears it — degrading the monitoring surface precisely during the incident conditions where an operator needs it. The protection this replaces was added deliberately in [bsv-blockchain/teranode#1201](https://github.com/bsv-blockchain/teranode/pull/1201) for the same threat model.

**Fix:** Either state in the PR that an unbounded display value is accepted risk, or bound the claim at ingest with a ceiling far wider than the sync cap so honest peers are unaffected, and score the peer:

```go
// A claim beyond any plausible network tip is a protocol violation, not a
// telemetry datum: drop it rather than storing it for display.
if uint64(advertisedHeight) > maxAcceptedHeight+uint64(implausibleAdvertisedLead) {
    s.logger.Warnf("[sanitizeAdvertisedTip] rejecting implausible advertised height %d from peer %s (local height %d)", advertisedHeight, peerID, localHeight)
    s.applyBanScore(peerID, ReasonProtocolViolation)
    return 0, nil, false
}
```

### ChiR3 — Tooltip presents the capped fallback as the peer's own claim

*Same as [@gokutheengineer's point 5 (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1605#pullrequestreview-4960959221).*

**Problem:** `ui/dashboard/src/routes/peers/+page.svelte:539-547` builds the tooltip from `display`, which is the fallback-resolved value, not from `advertised`:

```js
const display = advertised || syncHeight
let tooltip = `Advertised by peer: ${display.toLocaleString()}`
```

When `advertised_height` is 0 the label asserts that the peer advertised the locally derived cap. That is not a hypothetical path: it is every peer during a mixed-version rolling upgrade (the fallback the PR description explicitly supports), every peer when the blockchain service predates field 46 and drops it on `RegisterPeer`, and every peer loaded from a persisted registry blob written before this change.

**Why it matters:** The PR's whole point is that the dashboard should stop showing a number no peer sent. In the fallback case it still shows exactly that number, and this diff adds a label that positively claims the peer sent it — worse than the silent version it replaces, because an operator can no longer tell the two apart.

**Fix:** Label from `advertised`, and say so plainly when the peer did not report one.

```js
let tooltip = advertised
  ? `Advertised by peer: ${advertised.toLocaleString()}`
  : `Sync height (capped): ${syncHeight.toLocaleString()}\nPeer did not report an advertised height.`
if (advertised && syncHeight && advertised !== syncHeight) {
  tooltip += `\nSync height (capped): ${syncHeight.toLocaleString()}`
  tooltip += '\nCapped while this node catches up; used for sync decisions only.'
}
```

### ChiR4 — Peer detail modal contradicts the row that opened it

*Same as [@gokutheengineer's point 4 (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1605#pullrequestreview-4960959221).*

**Problem:** `+page.svelte:1027` renders the modal's Height metric from `selectedPeer.height` — the capped registry value — while the table cell at `:534-557` now renders `advertised_height`. During catchup a row headlines 962,710 and one click opens a panel reading 104,791 for the same peer.

**Why it matters:** Pre-PR the two agreed (both wrong, but consistent). This diff makes them disagree, so it introduces a fresh instance of the self-contradicting display it was written to eliminate, this time one click apart inside a single page rather than across two pages. Fixing the table but not the detail view for that table is the kind of gap that gets shipped and then puzzled over in an incident channel.

**Fix:** Resolve the same way the cell does and show both values explicitly.

```svelte
<div class="metric-item">
  <span class="metric-label">Height (advertised)</span>
  <span class="metric-value">#{(selectedPeer.advertised_height || selectedPeer.height)?.toLocaleString() || '0'}</span>
</div>
{#if selectedPeer.advertised_height && selectedPeer.height && selectedPeer.advertised_height !== selectedPeer.height}
  <div class="metric-item">
    <span class="metric-label">Sync height (capped)</span>
    <span class="metric-value">#{selectedPeer.height.toLocaleString()}</span>
  </div>
{/if}
```

## Non-blocking issues

### ChiR5 — The `GetPeers` RPC contract still carries only the capped height, so every non-browser consumer stays wrong

**Problem:** The PR extends `PeerRegistryInfo` but not `Peer`. `p2p_api.Peer` (`services/p2p/p2p_api/p2p_api.proto:8-36`) has no `advertised_height` — its highest field number is still 27 — so `Server.GetPeers` can only send the capped value even though it is reading straight out of the registry that now holds both:

```go
allPeers, err := s.peerRegistry.ListPeers(ctx, nil, 0, 0, false, false)   // Server.go:2070
...
resp.Peers = append(resp.Peers, &p2p_api.Peer{
    Id:            p.ID,
    CurrentHeight: p.Height,   // Server.go:2097 — p.AdvertisedHeight is available and dropped
```

`convertFromAPIPeerInfo`'s `*p2p_api.Peer` branch (`services/p2p/Client.go:822-833`) therefore leaves `AdvertisedHeight` at zero for every `GetPeers` caller. Two operator tools are affected. `cmd/monitor/monitor.go:945` prints `h:%d` from `peer.Height` under a panel titled "Top Peers by Height" (`:917`). `cmd/diagnose/health_checks.go:675-684` builds `max_peer_height` from it, and `:761-788` goes further and *derives* a number from it — `behind := int64(maxPeerHeight) - int64(chainHeight)` — which during catchup is pinned at exactly the configured lead, feeding the `behind > 10 && isRunning` FAIL branch.

Limits, stated so this is not read as broader than it is: this is pre-existing on `main`, not a regression introduced here. `services/rpc/handlers.go:1666` also calls `GetPeers` but uses only `len()` for `connections`, so `getinfo` is unaffected. The registry-unavailable fallback in `Server.GetPeers` (`:2105` onward) has no registry heights at all, so it is out of scope either way. And this is distinct from ChiR6: `getpeerinfo` reads `GetPeerRegistry` and is a one-field fix, whereas nothing downstream of `GetPeers` can be fixed at the call site alone.

**Why it matters:** The PR's stated goal is to report the raw claim wherever an operator reads it. Leaving the `GetPeers` contract untouched means that goal is met only for the browser: the two command-line tools an operator reaches for when the dashboard is unreachable still show the fabricated `local + 10000`, and `diagnose` turns it into a derived "blocks behind" figure that looks plausible and carries no hint that it is an artefact.

**Fix:** Fix it at the contract so all consumers are corrected at once, rather than per call site — add the field to `Peer`, populate it in `Server.GetPeers`, and carry it through the conversion.

```proto
  int64 feeFilter = 27;
  // Raw height the peer advertised, before the unvalidated-lead cap that
  // produces currentHeight. Telemetry only.
  uint32 advertisedHeight = 28;
```

```go
// Server.go GetPeers
CurrentHeight:    p.Height,
AdvertisedHeight: p.AdvertisedHeight,

// Client.go convertFromAPIPeerInfo, case *p2p_api.Peer
Height:           p.CurrentHeight,
AdvertisedHeight: p.AdvertisedHeight,
```

Then have `monitor` and `diagnose` prefer `AdvertisedHeight` with `Height` as the fallback. Pointing both tools at `GetPeerRegistry` instead would also work — both hold a `p2p.ClientI` (`monitor.go:153`, `health_checks.go:45`) — but it is two call-site changes that leave the next `GetPeers` consumer with the same bug.

### ChiR6 — `getpeerinfo` still reports the cap

*Same as [@gokutheengineer's point 3 (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1605#pullrequestreview-4960959221).*

**Problem:** `services/rpc/handlers.go:1311-1312` fills both `CurrentHeight` and `StartingHeight` from `p.Height`. That path already goes through `GetPeerRegistry` (`handlers.go:1239`), so `AdvertisedHeight` is sitting in the struct unused. After this PR the RPC and the dashboard report different heights for the same peer at the same moment.

**Why it matters:** `getpeerinfo` is the scripted, node-to-node-comparable view; leaving it on the cap means the fix is visible only to humans looking at a browser, and any tooling built on the RPC keeps consuming the fabricated value.

**Fix:** Read the advertised claim with the capped value as the fallback.

```go
advertised := p.AdvertisedHeight
if advertised == 0 {
    advertised = p.Height
}
CurrentHeight:  int32(advertised),
StartingHeight: int32(advertised),
```

### ChiR7 — `PeerSelector` debug logs print the capped height under the label `advertised_height`

**Problem:** `services/p2p/peer_selector.go:175-176` and `:191-192` log `advertised_height=%d` with `p.Height`; `:299-300` and `:303-304` do the same. Before this PR the label was loose but unambiguous — there was only one height. Now `AdvertisedHeight` is a real field with exactly that name and a different value, so the log line names one thing and prints another.

**Why it matters:** These lines are the primary evidence trail for sync-peer selection decisions, and the PR's own framing makes "advertised" mean specifically *uncapped*. Someone correlating a selection log against the peers page during catchup will see two different numbers under the same name and reasonably conclude something is inconsistent in the selector.

**Fix:** Rename the label to match what is printed.

```go
ps.logger.Debugf("[PeerSelector] Full node candidate: %s (validated_height=%d, sync_height=%d, mode=%s)",
    p.ID, p.ValidatedHeight, p.Height, p.Storage)
```

### ChiR8 — Four signatures now take two adjacent swappable positional `uint32`s

*Same as [@gokutheengineer's point 9 (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1605#pullrequestreview-4960959221).*

**Problem:** `registerPeer`, `addPeer`, `addConnectedPeer` (`services/p2p/server_helpers.go:610,629,638`) and `enqueueRegister` (`peer_registry_batcher.go:276`) all now take `height, advertisedHeight uint32` side by side. A transposed call compiles cleanly and silently swaps the sync value with the display value — the one mistake that would reintroduce the bug this PR fixes, with the cap landing in `AdvertisedHeight` and the raw claim in `Height`.

**Why it matters:** No compiler or test catches a transposition, and the batcher already has a struct-based `enqueue(peerID, *pendingPeerUpdate)` one layer down, so the pattern exists. Threading a small struct through would also have avoided touching ~30 test call sites — churn that repeats verbatim on the next field.

**Fix:** Pass a small value struct so the two heights are named at every call site, rather than ordered.

### ChiR9 — Per-message cap warning is the wrong level under this PR's own framing

*Same as [@gokutheengineer's point "still emits an unconditional per-message `Warnf` on every cap" (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1605#pullrequestreview-4960959221).*

**Problem:** `server_helpers.go:599` still emits an unconditional `Warnf` on every capped message. The PR description reports 6,504 of them in a single run.

**Why it matters:** The diff explicitly reframes the cap as "a bound on what an unvalidated claim may influence, not a statement about the peer". If capping is normal and expected for the whole duration of an initial sync, warning on essentially every gossip message contradicts that framing and buries genuine warnings. This is the same log line that surfaced the bug, so it has earned its keep — but as a `Debugf` or a once-per-peer warn.

**Fix:** Drop to `Debugf`, or keep `Warnf` behind a per-peer or per-interval gate so the first cap per peer is loud and the rest are not.

### ChiR10 — Swagger attaches the shared doc comment to `height`, leaving `advertised_height` undocumented

**Problem:** The doc comment in `services/asset/httpimpl/get_peers.go:18-20` sits above `Height`, so the generator hangs it entirely off `height` (`swagger.json:2963`) while the new `advertised_height` property (`swagger.json:2884-2888`) ships with no description at all — even though the text is mostly *about* `AdvertisedHeight`.

**Why it matters:** `swagger.json` is the published contract for `/api/v1/peers`; the field an integrator is most likely to misread is the one with no description, and the field they will read the description on is the one that no longer means what the headline number shows.

**Fix:** Split the comment so each field documents itself.

```go
// Height is capped against local progress and is what sync decisions use.
Height uint32 `json:"height"`
// AdvertisedHeight is the peer's raw claim. It differs from Height only
// while this node is catching up, and is 0 for peers on older versions.
AdvertisedHeight uint32 `json:"advertised_height"`
```

## Recap

|---|---|---|---|
| ChiR1 | Tie merge mispairs height and hash | YES | 72% |
| ChiR2 | Displayed height now unbounded | YES | 64% |
| ChiR3 | Capped fallback labeled as advertised | YES | 56% |
| ChiR4 | Modal contradicts its own row | YES | 50% |
| ChiR5 | `GetPeers` contract drops advertised height | | 44% |
| ChiR6 | `getpeerinfo` still reports cap | | 36% |
| ChiR7 | Selector log label vs value | | 24% |
| ChiR8 | Swappable positional height args | | 18% |
| ChiR9 | Per-message cap warning | | 12% |
| ChiR10 | New swagger field undocumented | | 8% |

ordishs:
@ctnguyen @gokutheengineer — all points from both reviews are addressed across two follow-up commits (`db0c344a6`, `af26811cb`).

**Blocking / must-fix**
- **ChiR1 / point 1** (tie merge mispairs height and hash): fixed in `peer_registry_batcher.go`. The hash update now lives inside the `advertisedHeight` comparison, and the equal-claims case takes the newest hash explicitly. Added a test with distinct hashes on each side of the tie (the old test used `nil` hashes and couldn't catch this).
- **ChiR2 / point 2** (unbounded displayed height): added `maxPlausibleAdvertisedHeight` (500,000,000) in `sanitizeAdvertisedTip` — a display-only ceiling independent of local height, so a genesis-sync peer still shows its real claim but a `uint32`-max-style claim gets clamped. Went with the ceiling-only option rather than adding a ban score for implausible claims; flagging that as a deliberate scope choice in case either of you wants to push back on it.
- **ChiR3 / point 5** (tooltip mislabels the capped fallback as "advertised"): tooltip now reads "Sync height (capped)" when `advertised_height` is 0, instead of asserting the peer advertised the capped value.
- **ChiR4 / point 4** (modal contradicts its own row): the detail modal's Height field now resolves `advertised_height || height`, same as the table cell.

**Non-blocking**
- **ChiR5** (`GetPeers` contract / `monitor` / `diagnose` still on the cap): added `advertisedHeight` to `p2p_api.Peer`, wired it through `Server.GetPeers` and `Client.convertFromAPIPeerInfo`, and pointed `cmd/monitor` and `cmd/diagnose` at it (with the capped value as fallback for older peers).
- **ChiR6 / point 3** (`getpeerinfo` still reports the cap): fixed, same fallback pattern.
- **ChiR7** (`PeerSelector` logs `advertised_height` while printing capped `p.Height`): relabeled those four log lines to `sync_height`.
- **ChiR8 / point 9** (swappable positional height args): `registerPeer`/`addPeer`/`addConnectedPeer`/`enqueueRegister` now take a `peerHeightClaim{Height, AdvertisedHeight}` struct instead of two adjacent `uint32`s.
- **ChiR9** (per-message cap warning): dropped to `Debugf`.
- **ChiR10** (swagger doc glued to `height`): split the comment so `advertised_height` documents itself in `swagger.json`.

Full `go build ./...`, `go vet`, and the `p2p`/`rpc`/`blockchain`/`asset`/`diagnose` suites with `-race` all pass, plus `npm run check`/`npm run test` on the dashboard side.

ctnguyen (approved):
Review at commit [af26811](https://github.com/ordishs/teranode/commit/af26811cb5213c170af6f35ad87cabeff8834f46)

Two follow-up commits close all ten points from the last round, and most of them properly rather than minimally. The tie-merge restructure is exactly what ChiR1 asked for and the new test genuinely distinguishes the two hashes rather than passing `nil` on both sides. `peerHeightClaim` is a better answer to ChiR8 than the one that was suggested — the doc comment names the transposition hazard, which is the part that keeps it alive through the next field. ChiR5 was fixed at the proto contract instead of per call site, so `monitor`, `diagnose` and `getpeerinfo` all got corrected at once. And the `var sanitizedBestHeight uint32` / `var registryHeight uint32` conversions, each with a comment explaining why the zero value is unreachable, were not asked for by anyone and are the kind of thing that makes the next audit of this file cheap. One previous point I would hold open, plus four new notes below — all four from the two commits no reviewer had seen.

## Status of previous points

ChiR2 is the one I would not close yet. `services/p2p/server_helpers.go:707` adds `maxPlausibleAdvertisedHeight = 500_000_000` and `:720-724` clamps the display value to it. That is half of what ChiR2's `Fix` asked for: the ceiling is there, the ban score is not, and the author flags the omission openly. But the harm ChiR2 described was that a single peer, at zero cost and with no ban-score consequence, pins the top of every height-sorted peer table — and a peer claiming 499,999,999 still does exactly that. That is roughly 500× any honest mainnet tip, and the value is still stored in the registry, still persisted, still served by `/api/v1/peers`, and still wins the sort. The clamp also *stores* the ceiling as though the peer had advertised it, so a clamped row renders 500,000,000 beside the peer's real tip hash — the shape of the bug this PR exists to remove. The other branch of ChiR2's `Fix` — stating in the PR that an unbounded display value is accepted risk — would close it just as well; what is not closed is doing neither. (The clamp's own `Warnf` at `:722` fires only on implausible claims, so it does not reintroduce ChiR9's volume during honest catchup — which is the scenario ChiR9 was about. For a peer that lies deliberately, every message is an implausible claim and the warning is per-message again; see ChiR14.)

Two rows read FIXED where the remedy is narrower than the `Fix` block implied, so to be explicit about what FIXED means. ChiR3: `ui/dashboard/src/routes/peers/+page.svelte:594-596` now labels from `advertised` rather than from `display`, which removes the false assertion — but the fallback branch reads only `Sync height (capped): N`, without the "peer did not report an advertised height" sentence, so the operator still has to infer *why* the label changed. ChiR4: the modal at `:1142` resolves `advertised_height || height` and therefore agrees with the row that opened it — but the second `Sync height (capped)` metric row was not added, and the modal value still carries no tooltip where the table cell does. In both cases the contradiction is gone; the fuller presentation is not.

|---|---|---|---|
| ChiR1 | Tie merge mispairs height and hash | YES | FIXED |
| ChiR2 | Displayed height now unbounded | YES | PARTIAL FIXED |
| ChiR3 | Capped fallback labeled as advertised | YES | FIXED |
| ChiR4 | Modal contradicts its own row | YES | FIXED |
| ChiR5 | `GetPeers` contract drops advertised height | | FIXED |
| ChiR6 | `getpeerinfo` still reports cap | | FIXED |
| ChiR7 | Selector log label vs value | | FIXED |
| ChiR8 | Swappable positional height args | | FIXED |
| ChiR9 | Per-message cap warning | | FIXED |
| ChiR10 | New swagger field undocumented | | FIXED |

## Non-blocking issues

### ChiR11 — `diagnose` now escalates an unvalidated peer claim into a FAIL verdict

**Problem:** `cmd/diagnose/health_checks.go:766-780` builds `maxPeerHeight` from `peer.AdvertisedHeight` (capped `Height` only as the older-peer fallback), then `:783` derives `behind := int64(maxPeerHeight) - int64(chainHeight)` and `:787` turns `behind > 10 && isRunning` into `StatusFAIL` with `behind=%d blocks` in the message. Before this delta the same loop read `peer.Height`, which `#1201` bounds at `localHeight + p2p_max_unvalidated_advertised_height_lead` — an operator-configurable ceiling, deliberately chosen to limit what an unvalidated claim can influence. The new input is bounded only by the hardcoded `maxPlausibleAdvertisedHeight`. This is a defect in what the previous review's ChiR5 `Fix` told the author to do ("have `monitor` and `diagnose` prefer `AdvertisedHeight` with `Height` as the fallback"); that advice was right for `monitor.go:948`, which only prints, and wrong here, where the value feeds a verdict. `services/p2p/p2p_api/p2p_api.proto:36-38` states the intended contract in as many words — "Telemetry only; currentHeight still drives sync decisions" — and `health_checks.go` is now the one consumer that breaks it.

**Why it matters:** Stating the limit plainly: a false FAIL was already reachable before this PR, because the capped value gave `behind = 10000` and `10000 > 10`. What changes is that the operator can no longer bound it through settings, and that `behind=` is now a figure the lying peer chooses rather than one derived from local progress. During honest catchup the new reading is a genuine improvement — `behind` finally means what it says instead of being pinned at the configured lead. The problem is only the escalation: `teranode-cli diagnose` is what an operator runs when they already suspect something, and a red "Chain vs Peers / behind=499,037,290 blocks" line on a healthy node sends them the wrong way at the worst moment. `:678` has the same shape for `max_peer_height` but only prints it, so it is fine as-is.

**Fix:** Keep reporting the advertised figure — it is the honest answer to "how far behind are we" — but stop letting it decide the verdict. Reserve `StatusFAIL` for a locally validated input and let a gap that rests only on peer claims be reported, with its provenance, without turning the check red:

```go
  // behind is derived from peer-advertised heights, which nothing local has
  // validated. Report it, but do not fail on it: one peer claiming an
  // implausible tip would otherwise turn a healthy node red.
  behind := int64(maxPeerHeight) - int64(chainHeight)
  msg := fmt.Sprintf("our_height=%d, best_peer=%d (peer-reported, unvalidated)", chainHeight, maxPeerHeight)
  if behind > 0 {
      msg += fmt.Sprintf(", behind=%d blocks", behind)
  } else {
      msg += ", in sync"
  }
  results = append(results, HealthResult{Service: "Chain vs Peers", Address: "-", Status: StatusOK, Message: msg})
```

`CheckStatus` currently has only `StatusOK` / `StatusFAIL` / `StatusSKIP` (`cmd/diagnose/diagnose.go:22-24`); if the signal is worth keeping visible, adding a warn level there and using it here is the fuller version. Either way the `isRunning` special case can go — it only existed to suppress this same figure during catchup.

### ChiR12 — The two merge branches still answer "which raw claim survives" in opposite ways

*Same as [@gokutheengineer's point 8 (18-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1605#pullrequestreview-4960959221).*

**Problem:** `services/p2p/peer_registry_batcher.go:89-94` adopts `u.advertisedHeight = from.advertisedHeight` unconditionally whenever `from.height > u.height`, so a lower raw claim replaces a higher one whenever the capped ceiling advanced between the two sanitizations. Three lines below, `:95-111` — the branch ChiR1 just tightened — now carries an explicit comment that a losing observation must not displace the winner's value or its hash. The contrast is sharper after this PR than before it: the tie branch orders by the raw claim, the branch above it orders by the capped height, and the capped height is the derived value of the two (`min(raw, localHeight + maxLead)` against a ceiling that moves during catchup).

**Why it matters:** Telemetry only, self-healing on the next flush, and the zeroing hazard genuinely does not exist (no caller produces `height > 0` with `advertisedHeight == 0`, since `sanitizeAdvertisedTip` always returns `displayHeight >= registryHeight` for any realistic local height). So this is low-impact on its own. It matters because the file now documents one policy in a comment and implements the other ten lines above it, and the next person to touch `merge()` has to work out which was intended.

**Fix:** Decide `advertisedHeight` and its hash by the raw claim in both cases, keeping the existing highest-capped-height rule for `height` itself. This does not disturb the four pinned merge assertions, and it extends the same within-window monotonicity the doc comment at `:77-84` already accepts for `height`:

```go
  if from.height > u.height {
      u.height = from.height
  }
  // The raw claim orders the observations in both cases: the capped value is
  // derived from a ceiling that moves during catchup, so it cannot decide
  // which claim a peer actually made last.
  switch {
  case from.advertisedHeight > u.advertisedHeight:
      u.advertisedHeight = from.advertisedHeight
      if from.blockHash != nil {
          u.blockHash = from.blockHash
      }
  case from.advertisedHeight == u.advertisedHeight:
      if from.blockHash != nil {
          u.blockHash = from.blockHash
      }
  }
```

If the current split is deliberate, a sentence on the upper branch saying so is enough.

### ChiR13 — The new `/api/v1/peers` field descriptions are false for legacy peers

**Problem:** `services/asset/httpimpl/get_peers.go:35-40` documents `height` as "capped against local progress and is what sync decisions use" and `advertised_height` as "0 for peers on older versions that do not report it", and `swagger.json:2884-2889` / `:2963-2968` publish both. But this endpoint lists peers of *both* transports — `:174` calls `ListPeers` with a nil filter, and `:156-159` says so explicitly. Wire-protocol peers are registered by `services/legacy/peer_registry_sync.go:150-158`, which sets `Height: snap.height` straight from the peer's `LastBlock()` (`services/legacy/Server.go:231-234`): never through `sanitizeAdvertisedTip`, never capped, and never read by the p2p sync path, which excludes wire peers entirely. That path also never sets `AdvertisedHeight`, so `advertised_height` is 0 for every legacy peer at every version. This wording came verbatim from the previous review's ChiR10 `Fix` block, which was written against the libp2p path only; this round revises that advice. The dashboard is unaffected — `+page.svelte:349-354` routes legacy rows to a separate card that renders `height` plainly — so this is a contract-document problem, not a rendering one.

**Why it matters:** `swagger.json` is the published contract for `/api/v1/peers`, and both statements mislead in a way an integrator cannot detect from the payload: they will read a legacy peer's uncapped wire height as a locally capped sync input, and read `advertised_height: 0` as evidence that the peer runs an old Teranode build when in fact no legacy peer will ever populate it.

**Fix:** Scope each sentence to the transport it describes.

```go
  // Height is the peer's latest known block height. For a libp2p peer it is
  // capped against local progress and is what sync decisions use; for a
  // legacy (wire-protocol) peer it is the height the peer reported, uncapped,
  // and no sync decision reads it.
  Height uint32 `json:"height"`
  // AdvertisedHeight is a libp2p peer's raw claim, differing from Height only
  // while this node is catching up. It is 0 for legacy peers, which never
  // report it, and for libp2p peers on versions that predate the field.
  AdvertisedHeight uint32 `json:"advertised_height"`
```

Regenerating the model is enough to carry this into `swagger.json`.

### ChiR14 — The ChiR9 fix and the new clamp warning disagree inside one commit

**Problem:** ChiR9 was closed by demoting the cap log from `Warnf` (`main:server_helpers.go:711`) to `Debugf` (`:742`), on the stated grounds that an unconditional per-message warning is the wrong level for something a peer can trigger on every message. Twenty lines above it, the same commit adds a new unconditional per-message `Warnf` at `:722` for the plausibility clamp. `sanitizeAdvertisedTip` runs once per gossip message from `server_helpers.go:111` and `Server.go:1575`; nothing in `services/p2p` rate-limits it, the call precedes `shouldSkipUnhealthyPeer` (`Server.go:1604`) so only an already-banned peer is filtered, and no path on this branch applies a ban score. Stating the limit plainly, because it matters for how this is read: **this is not a new attack surface.** The invalid-hash `Warnf` at `:716` is unchanged from `main:695` and is equally unconditional, equally per-message, and equally unscored, so a peer that wanted to fill the log could already do so from inside this same function. The point is only that one commit demoted one such line and added another.

**Why it matters:** During honest catchup the new line is silent, so ChiR9's 6,504-per-run problem does not come back — that part of the fix holds. The inconsistency shows up for the peer this line exists to describe: for a peer deliberately claiming an implausible height, every message is implausible, so the log fills at gossip rate with a warning that carries no consequence for the peer and no new information for the operator after the first one. Given that the same round decided the cap log should be per-peer-quiet, the clamp log should have been held to the same standard.

**Fix:** Apply ChiR9's own remedy to the line next to it — `Debugf`, or keep `Warnf` behind a first-occurrence-per-peer gate so the operator learns the peer is claiming nonsense once rather than continuously. If instead the intent is that a clamped claim is genuinely worth a warning every time, the two lines are now making opposite statements about the same class of event and one sentence saying why would settle it.

## Recap

|---|---|---|---|
| ChiR11 | Diagnose FAIL on unvalidated claim | | 40% |
| ChiR12 | Merge branches disagree on survivor | | 20% |
| ChiR13 | Peers API docs ignore legacy | | 10% |
| ChiR14 | Clamp warning contradicts ChiR9 fix | | 8% |

PR #1609, opened 2026-08-18
feat(svp2p): new SVNode-faithful P2P bridge service, phase 1 (transport + handshake)

ordishs:
## What

Phase 1 of the legacy-service rewrite (svp2p): a new `services/svp2p` service that connects to legacy Bitcoin peers, completes the SVNode handshake, holds healthy idle connections, and serves the existing `peer_api` gRPC surface. The service registers behind a `-svp2p=1` daemon flag and is mutually exclusive with `-legacy=1`.

Design spec and implementation plan live in `docs/superpowers/` (gitignored, available on request).

## Why

The current legacy service embeds an extraction of bsvd's p2p/netsync stack. It is over-abstracted and diverges from the reference implementation. The rewrite ports the semantics of bitcoin-sv C++ (`net.cpp` / `net_processing.cpp`) into Go, with goroutine-per-stream I/O as the one licensed deviation. This PR is the first slice: transport and handshake only. Block/tx ingestion (the bridge) comes in Phase 2.

## How

- `transport`: go-wire framing, one reader and one writer goroutine per connection, byte-budgeted droppable send lane plus a reserved priority lane, receive backpressure via a bounded channel.
- `protocol`: pure handshake state machine verified line-by-line against SVNode `net_processing.cpp` (`ProcessVersionMessage`, `ProcessVerAckMessage`, `ProcessProtoconfMessage`). Every ported decision carries a comment naming its C++ source. Deliberate non-carries for now: `authch`, `sendcmpct` (compact blocks arrive in a later phase and must not be advertised before they can be served).
- Peer runtime: SVNode ping cadence (2 min), idle timeout from `legacy_peerIdleTimeout`, misbehavior threshold 100.
- Ban list: manual bans (IP or CIDR) with JSON persistence, enforced on accept and dial, exposed through the existing `peer_api` Ban/Unban/IsBanned/ListBanned/ClearBanned methods with the same admin-key protection.
- Reuses the `legacy_*` settings keys so operators change nothing.

## Testing

- Table-driven handshake tests derived from SVNode behavior (message order, misbehavior scores, protoconf violations, self-connection, obsolete versions).
- Transport framing round-trips, close/backpressure semantics, race detector clean.
- End-to-end integration test: two full svp2p stacks handshake over real TCP, hold idle without flapping, and ban-then-refuse-reconnection works.
- `make lint` clean; full unit suite: 12,806 tests with one failure that pre-exists this branch (`TestDaemon_Start_AllServices` binds port 9091, occupied on the dev machine by a running node; verified identical on the base commit).

## Tracking

Phase 1 of 6 — tracker: #1610. Dark-launched: `-svp2p=1` only, mutually exclusive with `-legacy=1`. Not a QA target until the tracker's drop-in criterion is met.


ordishs:
Superseded by https://github.com/bsv-blockchain/teranode/pull/1675, which consolidates the svp2p phases into one service PR (everything svp2p-specific under services/svp2p, off by default via startSvp2p / -svp2p=1). The per-phase review history stays on this PR's branch.

PR #1611, opened 2026-08-18
feat(legacy): register legacy peers in the Teranode peer registry for dashboard visibility

ordishs:
## Purpose

Teranode connects to two separate peer networks. The p2p service uses libp2p. The
legacy service uses the original Bitcoin p2p network. The dashboard showed libp2p
peers only, so an operator could not see legacy peers at all.

This registers connected legacy peers in the centralized peer registry, flags them
as wire-protocol peers, and surfaces eight wire-protocol-only fields in the
dashboard.

## Scope

This is a read-only visibility path. It adds no decision path.

The legacy service keeps its own sync engine in `services/legacy/netsync`. That
engine discovers peers, compares chain tips and selects a sync peer without any
reference to the peer registry. This change does not touch that behaviour. It only
reports what the legacy service already does.

## Approach

The peer registry now lives in the blockchain service, fronted by
`blockchain.PeerRegistryClientI`. The legacy service already holds a
`blockchain.ClientI`, and the daemon already owns a shared registry client. The
legacy service therefore writes to the registry directly. No libp2p hop is
involved.

`TransportType.TRANSPORT_WIRE_PROTOCOL` already existed in the proto and was
described as belonging to the legacy service. Until now no production code wrote
it.

### 1. Registry and proto

`PeerInfo` gains a nested `Legacy *LegacyPeerInfo` block with eight fields:
direction, protocol version, service flags, ping, clock offset, starting height,
sync-peer flag and wire connect time.

The block is nested rather than eight loose fields because `Register` copies only
fields that carry "meaningful new data". `Inbound` and `IsSyncPeer` are meaningful
when false, so a per-field merge could not tell false from unset. A nil pointer
states "this update carries no legacy data" once.

`persistedRegistryVersion` moves from 2 to 3, following the existing precedent, so
a downgraded binary refuses to re-save and silently drop the new fields.

### 2. Reconcile loop

A goroutine in the legacy service ticks every `legacy_peerRegistrySyncInterval`
(default 10s), snapshots the connected peers, and pushes what changed. An
unchanged peer costs no RPC. The loop touches no wire path and no handshake path.
A registry failure logs a warning and the next tick retries.

Four behaviours the tests pin down, because each is a way this could look right
and be wrong:

1. `getPeers()` returns nil when its query channel is full or its reply times out.
   That must not read as "every peer disconnected".
2. `UpdatePeerMetrics` takes byte deltas, while the legacy snapshot supplies
   absolute totals. A reconnect resets those counters, so a backwards delta clamps
   to zero rather than wrapping.
3. `Register` refreshes `LastSeen`, which feeds registry TTL cleanup. A vanished
   peer is marked disconnected once and then dropped from tracking, so its entry
   can age out.
4. `Register` assigns `ConnectedAt` from its own clock and discards the caller's
   value. The true wire connect time therefore rides in `Legacy.TimeConnected`.

### 3. Keeping legacy peers out of libp2p paths

`ConnectedPeersCount` in the node-status gossip message keeps its established
libp2p-only meaning, because other nodes already consume and chart it. Legacy
peers get a separate `legacy_connected_peers_count` field, which is additive JSON.

The sync coordinator and the two p2p peer read RPCs now pass an explicit
`TRANSPORT_HTTP` filter. Catchup fetches blocks and subtrees over HTTP from a
peer's DataHub, which a wire peer cannot serve, so the exclusion was already
correct — but it rested on a wire peer happening to have an empty `DataHubURL`.
Filtering on transport states the real reason, so a future change that relaxed the
DataHub precondition cannot silently admit legacy peers.

### 4. Read path

The asset service reads the registry directly instead of hopping through the p2p
service. The registry keys legacy peers by a `legacy:host:port` string; a libp2p
`peer.ID` round trip would base58-encode those bytes and corrupt them. The hop
added no logic, and `/api/v1/peers` no longer depends on the p2p service.

Each peer gains a `transport` label. A legacy peer gains a nested `legacy` object.

### 5. Dashboard

A **Legacy Peers (Bitcoin P2P)** card lists wire-protocol peers in their own
table. A sync badge marks the peer the legacy sync manager selected, and a
disconnected peer renders greyed.

The Teranode table keeps its one-minute freshness filter and now excludes legacy
peers. The legacy card uses a wider window rather than the same one, because a
legacy peer pings every two minutes and a one-minute last-message bound would
hide a healthy peer between pings. A connected peer is always listed; a
disconnected one is kept only while its last message is within ten minutes.

That bound matters because an inbound legacy peer is keyed on its remote
address, which carries an ephemeral source port, so each inbound reconnect
produces a new registry entry rather than reusing the old one. Without the bound
a flapping peer would accumulate a row per reconnect until the registry TTL
expired them.

An earlier revision of this description claimed the card filtered on connection
state. It did not: the filter selected on transport alone and `is_connected`
only drove cell dimming, so the table had no staleness bound at all. Corrected
in c9b0879fb.

The peers smoke fixture previously carried an `{ok,data}` shape the page never
read, so only the empty state had ever been exercised. It now returns the real
`{peers,count}` payload, and a new smoke test asserts the transport split, the
sync badge and the greyed disconnected row.

## Settings

| Key | Default |
|---|---|
| `legacy_peerRegistryEnabled` | `true` |
| `legacy_peerRegistrySyncInterval` | `10s` |

Both are wired through `NewSettings`. The flag is a kill switch.

## Verification

```
go build ./...                                      exit 0
go vet ./...                                        4 issues, all pre-existing in test/utils/
make test                                           DONE 12790 tests, 45 skipped, 1 failure in 497.971s
                                                    0 data races
golangci-lint --new-from-rev upstream/main          0 issues.
npm run check                                       801 FILES 0 ERRORS 14 WARNINGS
npm test                                            17 smoke passed, 40 unit passed
npm run build                                       done
```

The single `make test` failure is `daemon TestDaemon_Start_AllServices`, which
fails on `listen tcp 127.0.0.1:9091: bind: address already in use`. It is a
pre-existing environmental failure: the same test fails identically on a clean
checkout of `main`, because another process on the development machine held that
port.

`go vet` reports four issues, all in `test/utils/`, which this branch does not
touch.

## Residual risk

The reconcile loop is tested against a real `CentralizedPeerRegistry`. The
snapshot adapter `legacyPeerSnapshots` is not yet exercised against a live legacy
peer, so an end-to-end check against a real stack is still worth doing before this
is relied on operationally.

The svp2p legacy bridge rewrite also touches `services/legacy`, so the two may
need reconciling.


services/asset/httpimpl/get_peers.go
```diff
@@ -48,77 +70,111 @@ type PeersResponse struct {
...
+		BytesReceived:   peer.BytesReceived,
+		LastBlockTime:   peer.LastBlockTime.Unix(),
+		LastMessageTime: peer.LastMessageTime.Unix(),
+
+		// Catchup-specific counters. The timestamps remain the generic
+		// interaction ones; catchup-scoped timestamps are not tracked.
+		CatchupAttempts:        peer.CatchupAttempts,
+		CatchupSuccesses:       peer.CatchupSuccesses,
+		CatchupFailures:        peer.CatchupFailures,
+		CatchupLastAttempt:     peer.LastInteractionAttempt.Unix(),
```
galt-tr: **[bug]** Switching `/api/v1/peers` from `p2p.GetPeerRegistry` to the blockchain registry dropped p2p's `timeToUnix` zero-guard (`if t.IsZero() { return 0 }`). `time.Time{}.Unix()` is `-62135596800`, which is truthy in JS, so the catchup modal (`selectedPeer.catchup_last_attempt ? new Date(...) : 'Never'`) renders year 1 instead of "Never" for any peer that has never run catchup. This hits no-legacy nodes: they still serve this endpoint, and most libp2p peers have unset catchup timestamps. `LastCatchupErrorTime`, `LastBlockTime`, and `Legacy.TimeConnected` have the same conversion.

**Suggestion:** Add a local helper matching p2p (`if t.IsZero() { return 0 }; return t.Unix()`) and use it for every timestamp in `peerInfoToResponse`. Cover an unset catchup timestamp in `TestPeerInfoToResponse_Libp2pPeerOmitsLegacy`.

ui/dashboard/src/routes/peers/LegacyPeersCard.svelte
```diff
@@ -0,0 +1,212 @@
...
+<Card contentPadding="16px">
+  <div class="legacy-header">
+    <Typo variant="title" size="h4" value="Legacy Peers (Bitcoin P2P)" />
+    <span class="legacy-count">{connectedCount} connected / {peers.length} known</span>
+  </div>
+
+  {#if peers.length === 0}
+    <div class="legacy-empty">
+      <p>No legacy peers</p>
+      <p class="sub">The legacy service is not connected to any Bitcoin P2P peer.</p>
```
galt-tr: **[suggestion]** `LegacyPeersCard` is always mounted. When `startLegacy` is false the registry has no wire peers, so operators who never enabled the legacy service still see a card whose empty copy says "The legacy service is not connected to any Bitcoin P2P peer." That implies the service is running. The same copy also appears for the first sync interval after legacy starts, because `run()` waits for the ticker and does not reconcile immediately.

**Suggestion:** Either hide the card when `peers.length === 0`, or change the subtitle so it does not assume the legacy service is up (e.g. "No Bitcoin P2P (legacy) peers are registered."). Optionally fire one reconcile on start so a live legacy node is not empty for 10s.

ui/dashboard/src/routes/peers/LegacyPeersCard.svelte
```diff
@@ -0,0 +1,212 @@
...
+                {peerAddress(peer)}
+                {#if peer.legacy?.is_sync_peer}
+                  <span class="badge sync">SYNC</span>
+                {/if}
+                {#if peer.is_banned}
+                  <span class="badge banned">BANNED</span>
+                {/if}
+              </td>
+              <td class="agent" title={peer.client_name || ''}>{peer.client_name || '-'}</td>
+              <td>{peer.legacy?.inbound ? 'in' : 'out'}</td>
```
galt-tr: **[suggestion]** `{peer.legacy?.inbound ? 'in' : 'out'}` treats a missing `legacy` object as outbound. Reconcile always attaches the nested block, but a persisted or partial payload with `transport === 'legacy'` and `legacy == null` would show "out" rather than unknown.

**Suggestion:** Render `-` (or similar) when `peer.legacy` is absent; only then branch on `inbound`.

galt-tr (commented):
## Summary

The PR mirrors connected Bitcoin-wire peers into the centralized registry for dashboard visibility, and the no-legacy path is isolated correctly: the reconcile loop lives only in `legacy.Start` (so `startLegacy=false` never runs it), Asset `/api/v1/peers` still lists libp2p registry entries, `ListPeers(TRANSPORT_HTTP)` uses proto `TRANSPORT_HTTP = 0` so unset TransportType still matches libp2p peers and does not admit wire peers into catchup, and `connected_peers_count` stays libp2p-only with `legacy_connected_peers_count` omitempty. Merge/TTL of the nested Legacy block, uint64 delta clamping, nil-vs-empty `getPeers` snapshots, and the persistence version bump to 3 all look right.

Focused on a node that has **not enabled the legacy service**. Isolation looks correct. The one real regression is that Asset `GetPeers` now calls `time.Time.Unix()` on zero timestamps without the zero-guard p2p used, so a no-legacy node's catchup modal shows year-1 dates instead of "Never".

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 0

services/asset/httpimpl/get_peers.go
ordishs: Confirmed and fixed in 4e00dbbd6.

Verified the guard did exist: `peerInfoToP2PProto` (`services/p2p/Server.go:2472`) defines `timeToUnix` as a local closure and applies it to seven timestamps, and `convertFromAPIPeerInfo` then turns those zeros into `time.Unix(0, 0)`, so the old chain served 0. Reading the registry directly dropped it.

Added `timeToUnix` in `get_peers.go` and applied it to all eight timestamps — the seven generic ones plus `Legacy.TimeConnected`. Tests: `TestPeerInfoToResponse_UnsetTimestampsAreZero` failed with exactly `-62135596800` before the change, `TestPeerInfoToResponse_SetTimestampsSurvive` guards against a guard that swallows real values, and `TestPeerInfoToResponse_UnsetLegacyTimeConnectedIsZero` covers the nested block.

Also covered the `GetPeers` handler itself, which had no test: both transports from one read, an empty registry, a missing registry client (503), and a registry failure (500).

services/legacy/peer_registry_sync.go
```diff
@@ -0,0 +1,194 @@
...
+
+		if err := p.registry.UpdateConnectionState(ctx, id, false); err != nil {
+			p.logger.Warnf("[LegacyPeerRegistry] disconnect %s failed: %v", id, err)
+			continue
+		}
+
+		// Drop the tracking entry so no later tick registers this peer again.
+		// Register refreshes LastSeen, which feeds registry TTL cleanup; a peer
+		// that kept being registered would never age out.
+		delete(p.lastSeen, id)
```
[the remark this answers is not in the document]
ordishs: Confirmed and fixed in 81c52f38a.

Reproduced it first. The failing test is the snapshot-gap shape rather than a full TCP reconnect: `getPeers()` only returns peers whose `Connected()` is true, so a peer can leave one snapshot and return with the same object and its counters still climbing. With 1000 bytes reported, one absent tick, then 1200, the entry read 2200 instead of 1200 — and it compounds on every gap, so the drift is unbounded.

Fix: on a first sighting the loop now reads the stored totals back via `GetPeer` and baselines against them, so the surviving entry is never added to itself. One extra RPC per newly-seen peer only.

I also reworked `byteDelta` rather than taking the clamp-to-zero suggestion. Clamping fixes the over-report but silently drops the new connection's bytes. Legacy counters only ever climb within one connection, so a backwards counter means the connection object was replaced and the current total again belongs entirely to the new connection — reporting it is correct in both shapes:

- gap, counters continued: 1000 then 1200 → 1200
- gap, counters reset: 1000 then 50 → 1050

Both are covered (`TestReconcile_ReconnectDoesNotDoubleCountBytes`, `TestReconcile_ReconnectWithResetCountersDoesNotRegress`), plus a baseline-lookup failure degrading to reporting the whole total. This changed one existing expectation: `TestReconcile_ByteCountersAreDeltas` previously asserted a backwards counter added nothing, and now asserts it contributes its new bytes.

ui/dashboard/src/routes/peers/LegacyPeersCard.svelte
ordishs: Fixed in a83a08c6f, taking the copy option rather than hiding the card.

The subtitle now reads "No Bitcoin P2P (legacy) peers are registered." — it states only what is known. I kept the card mounted deliberately: hiding it on an empty list makes "legacy is enabled with no peers yet" indistinguishable from "this build has no legacy view", which is the question an operator is usually asking.

The first-interval blank is fixed too — `run()` now reconciles once before entering the ticker loop, so a live legacy node populates immediately instead of after 10s. If the internal server is not answering yet the snapshot is nil and that tick is skipped harmlessly. Covered by `TestRun_ReconcilesImmediatelyAndStopsOnContextCancel`, which uses a one-hour interval so only the up-front call can satisfy it.

One disclosure: that change rode along in 81c52f38a because it touches the same file, so its commit message does not mention it.

ui/dashboard/src/routes/peers/LegacyPeersCard.svelte
ordishs: Fixed in a83a08c6f. Renders `-` when the nested block is absent, and only branches on `inbound` when it is present.

ordishs:
Pushed four commits addressing the review. Per-thread replies are inline; two items had no thread:

**`GetPeersForCatchup` transport filter** (a83a08c6f) — applied. Worth noting the function's own doc comment already claimed "HTTP transport" while passing `nil`, so this makes the code match its stated contract rather than only adding consistency. It was safe before purely because of the `DataHubURL == ""` guard, which is exactly the implicit reliance this PR set out to remove elsewhere.

**Coverage gate (73%, needs 80%)** — the shortfall was three functions at 0%, all now covered:

| | before | after |
|---|---|---|
| `GetPeers` handler | 0% | 93.8% |
| `peerSnapshotFrom` / `legacyPeerSnapshots` | 0% | 100% (mapping) |
| `run` | 0% | 88.9% |
| `peerInfoToResponse` | 85.7% | 100% |
| `reconcile` | 83.3% | 83.3% |

Rather than bolt on tests for coverage's sake, e356d7d60 extracts `peerSnapshotFrom` and a `legacyPeerStats` interface out of `legacyPeerSnapshots`. That closes the residual risk flagged in the PR description — the field-for-field mapping the dashboard renders was previously unverifiable without a live legacy server, and is now tested directly, including the sync-peer badge, skipped multistream sub-peers, rejected addresses, and the negative-last-block clamp.

`legacyPeerSnapshots` itself stays at 0%: what remains is nil guards and a loop over the now-covered mapping, and exercising it needs a full legacy server. Its nil-means-no-data contract — the precondition the reconcile loop depends on — is covered separately.

**Verification on the current tip:**

```
go build ./...                                exit 0
go test ./services/{legacy,asset,blockchain,p2p}/... ./settings/...   all ok
golangci-lint --new-from-rev upstream/main    0 issues.
npm run check                                 802 FILES 0 ERRORS 14 WARNINGS
npm test                                      17 smoke passed, 40 unit passed
npm run build                                 done
```

The 14 svelte-check warnings and the `npm run lint" Prettier failures are pre-existing repo-wide; my files are Prettier-clean.

One behaviour change to flag for a second look: `byteDelta` now reports a backwards counter's total as new bytes instead of clamping to zero. Reasoning is in the thread on the byte-counter comment — clamping fixed the over-report but dropped the reconnected peer's bytes. This flipped one existing test expectation, deliberately.

galt-tr (approved):

oskarszoon (changes requested):
Reviewed at `963aa0af`. The mechanical work here is clean — the proto change is purely additive (`legacy = 46`, nothing renumbered), the committed `.pb.go` is a verbatim regeneration, the `Register` merge clones behind a fresh `*LegacyPeerInfo` with no aliasing, the `uint64` byte-delta clamp is right in both directions, the reconcile goroutine exits cleanly on ctx cancel, and the nil-vs-empty `getPeers()` distinction is genuinely honoured. `TRANSPORT_HTTP` being enum 0 means the new filters don't hide pre-existing untagged entries from catchup, so that reasoning holds too.

One bug makes the feature inert, plus a body/code mismatch.

**P1 — `reconcileConnectionStates` clears `IsConnected` on every legacy peer, `services/p2p/server_helpers.go:1378`**

Four of the five p2p `ListPeers` call sites were converted to `transportHTTPFilter()`. The one that got missed is the only one that *writes*:

```go
```

`live` comes from `snapshotLiveConnIDs()` → `s.P2PClient.GetPeers()`, libp2p IDs only, so a `legacy:host:port` key can never be a member. `case info.IsConnected && !isLive:` at `:1391` then fires `UpdateConnectionState(..., false)`.

Nothing repairs it: `peer_registry_sync.go:142` asserts connection state only under `if !known`, and that once-only behaviour is a pinned invariant (`TestReconcile_UnchangedPeerCostsNoRegisterCall`). `Register`'s existing-entry branch never writes `IsConnected`.

So ~60s after a wire peer connects (`p2p_peer_map_cleanup_interval` default 1m): `legacy_connected_peers_count` gossips 0 (`Server.go:1984-1991` skips `!p.IsConnected`), the card header reads 0 connected, every live row renders dimmed, and `/api/v1/peers` reports `is_connected: false` for connected peers.

The line is unchanged from `main`, but on `main` a libp2p-only liveness set was a correct oracle — this PR introduces the first entries the sweep structurally cannot see. Fix is `transportHTTPFilter()` at `:1378`. `legacy_peer_isolation_test.go` covers `getNodeStatusMessage`, `SyncCoordinator`, `GetPeerRegistry` and `GetPeers` but never `reconcileConnectionStates` with a wire peer — worth a case there.

**P2 — the legacy card does not filter on connection state, `ui/dashboard/src/routes/peers/+page.svelte:325`**

```js
legacyData = fetched.filter(peer => peer.transport === 'legacy')
```

The comment and the PR body both say legacy peers are filtered on connection state "instead of" the freshness window. They aren't — `is_connected` only drives `cellClass()` dimming. So the legacy table has no staleness bound at all. That matters because inbound legacy peers key on `p.conn.RemoteAddr()` (`peer.go:2948`, ephemeral source port), not the stable dialled address the comment at `peer_registry_sync.go:23` describes, so a flapping inbound peer accumulates a fresh dimmed row per reconnect. Either add a recency bound or fix the comment and the body.

**P2, worth a look before merge**

- `peer_registry_sync.go:171` — a failed `UpdatePeerMetrics`/`UpdateLastMessageTime` falls through to the unconditional `p.lastSeen[snap.id] = snap` rebaseline at `:184`, unlike the `RegisterPeer` path which `continue`s. One transient RPC failure permanently drops that tick's byte delta rather than retrying it.
- `peer_registry_sync.go:187` — no first-tick reconcile against what the registry already holds, so a legacy-service restart with peers connected leaves phantom `IsConnected=true` entries for up to the 24h TTL. The P1 above currently masks this; fixing the P1 unmasks it.
- `peer_registry_persistence.go:50` — the v3 bump behaves as documented, but `Load` returns before repopulating `r.peers` and `r.banScores`, so a rollback runs with an empty registry: no persisted bans, no reputation for catchup selection. That's a steep price for a field the legacy sync re-derives in 10s.
- `legacy_settings.md` isn't updated with the two new keys.

services/legacy/Server.go
```diff
@@ -162,9 +170,113 @@ func New(logger ulogger.Logger,
...
+	}
+
+	serverPeers := s.server.getPeers()
+	if serverPeers == nil {
+		return nil
+	}
+
+	var syncPeerID int32
+	if s.server.syncManager != nil {
+		syncPeerID = s.server.syncManager.SyncPeerID()
```
freemans13, line 274: **`SyncPeerID()` has no timeout, no select and no context, so a reconcile tick can outlive shutdown.**

This is the one call in the snapshot path with no bound on how long it waits, and the contrast with the line three above it is what gives it away. `s.server.getPeers()` is careful about exactly this hazard: a non-blocking send with a `default` arm, a 5 second receive timeout, and a documented nil return meaning "no answer" (`peer_server.go:3801`). `SyncPeerID()` has none of it:

```go
func (sm *SyncManager) SyncPeerID() int32 {
	reply := make(chan int32)
	sm.msgChan <- getSyncPeerMsg{reply: reply}

	return <-reply
}
```

`reply` is unbuffered, and the only goroutine that ever answers it is `blockHandler`. That loop leaves on `case <-sm.quit: break out` and then drains nothing, and `SyncManager.Stop()` closes `sm.quit` (`netsync/manager.go:3131`, `:2531`, `:3064`).

So if a reconcile tick is in flight when netsync stops, `<-reply` never returns. `reconcile` never returns. `run`'s `case <-ctx.Done()` is never reached, and the goroutine is parked for the life of the process holding its snapshot slice. Same shape, less severe, if the block handler is merely slow: one tick waits behind whatever block is being validated.

Two things make it easy to miss.

Before this PR `SyncPeerID` had no production caller at all. Outside its own definition, `grep -rn 'SyncPeerID()' services/` hits `netsync/manager_test.go` and this line, nothing else. It has never been on a live path, so nothing has ever forced it to grow a guard.

`TestRun_ReconcilesImmediatelyAndStopsOnContextCancel` does assert clean shutdown on context cancel, but it injects `func() []peerSnapshot { return []peerSnapshot{snap} }`, so `legacyPeerSnapshots` never runs and neither does this call (`peer_snapshot_test.go:150`). It pins the loop's shape, not the path production takes.

**The fix is cheap: drop the channel.** The value already sits behind a mutex that is safe to read concurrently. The handler for `getSyncPeerMsg` does nothing but

```go
case getSyncPeerMsg:
	var peerID int32
	if sp := sm.loadSyncPeer(); sp != nil {
		peerID = sp.ID()
	}
	msg.reply <- peerID
```

and `loadSyncPeer` is an `RLock` read, with `storeSyncPeer` the only writer taking the same lock (`manager.go:2618`, `:583`, `:619`). A direct accessor returns the same value with no wait:

```go
// SyncPeerIDNoWait returns the current sync peer's ID, or 0 if there is none,
// without a round trip through the block handler's message channel.
func (sm *SyncManager) SyncPeerIDNoWait() int32 {
	if sp := sm.loadSyncPeer(); sp != nil {
		return sp.ID()
	}

	return 0
}
```

That also clears up something smaller. The PR description says the loop "touches no wire path and no handshake path", and the `legacy_peerRegistryEnabled` longdesc says the mirror "feeds no sync, catchup or peer-selection decision". Both are true about decisions. But `SyncPeerID()` does put a message on `sm.msgChan` every 10 seconds, and that channel is the sync manager's single serialization point. Your own comment at `manager.go:2454` spells out why it matters: `blockHandler` is its sole consumer, so anything stalling it stalls disconnects, sync-peer rotation, inv, headers and tx dispatch for every peer. Each message also costs `blockHandler` one `GetFSMCurrentState` call. Tiny in absolute terms, and I am not claiming it hurts today. A dashboard feature just should not be in there when a lock read gives the same answer.

services/legacy/peer_registry_sync.go
```diff
@@ -0,0 +1,216 @@
...
+				Legacy:           &legacyCopy,
+			}
+
+			if err := p.registry.RegisterPeer(ctx, info); err != nil {
+				p.logger.Warnf("[LegacyPeerRegistry] RegisterPeer %s failed: %v", snap.id, err)
+				continue
+			}
+		}
+
+		if !known {
```
freemans13, line 170: Confirming oskarszoon's P1 from my own read rather than re-reporting it, because this line is the half of it that lives on branch-authored code.

`reconcileConnectionStates` clears `IsConnected` on every legacy peer. Its liveness set comes from `snapshotLiveConnIDs()`, which walks `P2PClient.GetPeers()` and therefore holds libp2p IDs only, so a `legacy:host:port` key can never be a member and the `info.IsConnected && !isLive` arm fires (`services/p2p/server_helpers.go:1293`, `:1378`, `:1391`). `transportHTTPFilter()` at `:1378` is the fix.

Nothing repairs it from this side, and that is what this line decides. `UpdateConnectionState(..., true)` runs only under `if !known`, and `known` stays true for as long as the peer keeps turning up in snapshots. `Register`'s existing-entry branch never writes `IsConnected` either (`services/blockchain/peer_registry.go:390` onwards). So the flag goes false once and stays false while the peer is still connected, and `legacy_connected_peers_count` reports 0.

Whichever way you fix it, please make the test assert the end state rather than the argument. A case in `legacy_peer_isolation_test.go` that puts a wire peer in the registry, runs `reconcileConnectionStates`, and asserts `IsConnected` is still true afterwards will fail if the filter is ever dropped again. A test that only checks the filter passed to `ListPeers` will not.

services/legacy/peer_registry_sync.go
```diff
@@ -0,0 +1,216 @@
...
+				ID:               snap.id,
+				TransportType:    blockchain_api.TransportType_TRANSPORT_WIRE_PROTOCOL,
+				TransportTypeSet: true,
+				ClientName:       snap.userAgent,
+				NetworkAddress:   snap.addr,
+				Height:           snap.height,
+				Legacy:           &legacyCopy,
+			}
+
+			if err := p.registry.RegisterPeer(ctx, info); err != nil {
```
freemans13: Minor, and separate from the connection-state bug. The four registry calls in `reconcile` pass the loop's context straight through, and `PeerRegistryClient` adds no deadline of its own (`services/blockchain/peer_registry_client.go:196` and the wrappers around it).

The p2p side bounds the identical RPCs on purpose: every `SyncCoordinator` registry call goes through `boundedRPCContext()` (`services/p2p/sync_coordinator.go:299`). An unresponsive blockchain service will not wedge this loop permanently, since context cancel still works here, but one tick can hang for as long as the gRPC connection takes to give up, and with a 10 second ticker that is a visible gap in the dashboard. Worth matching the p2p bound.

freemans13 (changes requested):
Requesting changes. oskarszoon's P1 is real, I confirmed it independently, and I found one more of the same shape on the legacy side.

The new one: `legacyPeerSnapshots` calls `SyncManager.SyncPeerID()`, which is a blocking send on the sync manager's message channel followed by a receive on an unbuffered reply channel, with no select, no timeout and no context. Once `SyncManager.Stop()` closes `sm.quit` nothing drains that channel, so a reconcile tick in flight at shutdown parks forever and the loop's `case <-ctx.Done()` is never reached. Before this PR `SyncPeerID` had no production caller at all, only tests, which is why it has never needed a guard. The fix is a lock read instead of a channel round trip, since `loadSyncPeer()` is already concurrency-safe. Detail in the thread on `services/legacy/Server.go:258`.

Both of these are the same class: a path that is correct in isolation and wrong once this PR puts new kinds of entry, or a new caller, in front of it. The registry code itself is fine. It is the two seams into existing services that need another look.

Everything else I checked came out clean, and some of it is careful work worth saying so about. The proto change is purely additive at field 46, nothing renumbered. The `filter_transport` boolean sentinel means passing `TRANSPORT_HTTP`, which is enum 0, actually arrives as a filter rather than being erased by proto3's zero-value omission, so the new exclusions really do exclude (`services/blockchain/peer_registry_grpc.go:53`). The persisted envelope marshals `[]*PeerInfo` directly, so the nested legacy block genuinely is written to disk and the version 3 bump does what its comment claims. `clonePeerInfo` clones the legacy pointer, so the registry never aliases a caller's struct. Both new settings are loaded in `settings/settings.go:699-700`, not merely struct-tagged, which is a failure I have seen three times in this repo.

Two smaller things in the threads: the reconcile loop's registry RPCs carry no deadline where the p2p sync coordinator bounds the identical calls, and I have added the branch-side detail to the connection-state finding.

Verification I ran on `963aa0af5`:

```
go build ./...                                             exit 0
go test -race -count=1 ./services/legacy/ \
    ./services/asset/httpimpl/ ./services/blockchain/ \
    ./services/p2p/ ./settings/                            all ok, 0 races
```

What I did not do: I never ran the dashboard suite, and I never put this in front of a live legacy peer. The shutdown hang is a code-reading result, not an observed one.

services/legacy/peer_registry_sync.go
ordishs: Confirmed and fixed in 3d3e1f3df. Both halves of your read check out.

`reconcileConnectionStates` arrived here in the upstream merge (963aa0af5), after this branch's p2p work, which is why the filter was never applied to it. `live` comes from `snapshotLiveConnIDs()` walking `P2PClient.GetPeers()`, so a `legacy:host:port` key can never be a member and `info.IsConnected && !isLive` fires for every legacy peer on every sweep.

Your second half is the part that makes it stick, and it is right: `UpdateConnectionState(..., true)` runs only under `if !known`, `known` stays true while the peer keeps appearing in snapshots, and `Register`'s existing-entry branch never writes `IsConnected`. So the flag goes false once and stays false with the peer still connected.

Applied `transportHTTPFilter()` at the `ListPeers` call. The sweep's own doc comment already said it syncs "actual libp2p connectedness", so this makes it match its stated contract.

Test asserts the end state as you asked — `TestReconcileConnectionStates_LeavesWirePeersConnected` registers two connected wire peers plus a live and a dead libp2p peer, runs the sweep, and asserts the wire peers are still `IsConnected` while the dead libp2p peer is still cleared. It failed on the pre-fix code with `legacy:203.0.113.7:8333 is a wire peer: a libp2p connectedness sweep must not clear it`, and the three pre-existing sweep tests still pass, so the libp2p half is intact.

I did not add a self-heal on the legacy side. With the filter applied nothing else clears these entries, and re-asserting per tick would put an RPC back on every peer on every tick. Happy to add a bounded version if you would rather not depend on that.

services/legacy/Server.go
ordishs: Confirmed and fixed in 1413369de. Verified each link in the chain rather than taking it on trust:

- `SyncPeerID` (`manager.go:3131`) — unbuffered `reply`, send on `sm.msgChan`, then `<-reply`.
- Handler at `:2618` is exactly as you quoted.
- `loadSyncPeer` (`:583`) is an `RLock` read; `storeSyncPeer` (`:619`) is the only writer and takes the same lock.
- `blockHandler` leaves on `<-sm.quit` and drains nothing.
- `grep` for production callers returns exactly one hit: `services/legacy/Server.go:258`. This branch put it on a live path for the first time, which is why it had never needed a guard.

I went further than the additive accessor you sketched and converted `SyncPeerID` itself, removing `getSyncPeerMsg` and its handler arm. Reasoning: `getSyncPeerMsg` had exactly one sender, the netsync tests call the accessor rather than the message, and leaving a blocking method beside a safe one preserves the trap for the next caller. All eight existing `SyncPeerID` assertions in `manager_test.go` pass unchanged, which is decent evidence the value is the same.

Two new tests, both proven non-vacuous by restoring the old body: they fail after the full 5s guard (`SyncPeerID blocked: it must not depend on the block handler running` / `SyncPeerID blocked after Stop`). One uses a manager that was never started, the other one that has been stopped — your two shapes.

You were also right about my claim being loose, and I would rather own that than let it stand. "Touches no wire path" and "feeds no sync, catchup or peer-selection decision" were true about decisions but not about `sm.msgChan`, which is the serialization point your own comment at `manager.go:2454` warns about, and it cost `blockHandler` a message plus a `GetFSMCurrentState` every 10s. After this change the loop genuinely does not touch it, so the claim is now accurate rather than needing to be softened.

services/legacy/peer_registry_sync.go
ordishs: Fixed in 51e38af06. Verified the asymmetry: `boundedRPCContext` is `context.WithTimeout(sc.ctx, sc.rpcTimeout)` with `defaultRPCTimeout = 5 * time.Second` (`sync_coordinator.go:295`, `:132`), and none of the `PeerRegistryClient` wrappers add a deadline.

All six registry calls in `reconcile` now derive a 5s per-call deadline from the loop context, matching that value. Explicit `cancel()` rather than `defer` since they are inside the per-peer loop.

`TestReconcile_BoundsEachRegistryCall` uses a client that blocks until its context is done, and asserts `reconcile` returns and the call was cut at the timeout rather than left to the transport. `TestReconcile_DefaultRPCTimeoutIsSet` guards the wiring, since an unset timeout would make every call fail immediately.

I kept it per-call to match the p2p precedent. Worth noting that leaves a tick's total unbounded in peer count; if you would prefer a whole-tick cap instead, say so.

services/legacy/peer_registry_sync.go
```diff
@@ -0,0 +1,216 @@
...
+
+		if err := p.registry.UpdateConnectionState(ctx, id, false); err != nil {
+			p.logger.Warnf("[LegacyPeerRegistry] disconnect %s failed: %v", id, err)
+			continue
+		}
+
+		// Drop the tracking entry so no later tick registers this peer again.
+		// Register refreshes LastSeen, which feeds registry TTL cleanup; a peer
+		// that kept being registered would never age out.
+		delete(p.lastSeen, id)
```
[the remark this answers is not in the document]
ordishs: Agreed, and verified rather than assumed. On the current tip `Load` sets `info.IsConnected = false` on every restored entry (`peer_registry_persistence.go:251`) and `isCleanupExempt` only exempts a connected peer whose activity is inside the TTL (`peer_registry.go:1432`). A persisted-connected legacy peer that never reconnects is therefore cleared on load and reapable, so no first-tick registry reconciliation is needed.

Related but separate, and it did need fixing: `reconcileConnectionStates` was clearing `IsConnected` on live legacy peers because it listed the registry unfiltered while deriving liveness from libp2p IDs. Fixed in 3d3e1f3df — see the thread on `server_helpers.go`.

ordishs:
Three more fixes pushed, all from @freemans13's review. Per-thread detail is inline; summary and the state of the two bot items:

| Finding | Status |
|---|---|
| Connection sweep cleared wire peers | Fixed 3d3e1f3df — **was silently zeroing the feature** |
| `SyncPeerID` could park the reconcile goroutine | Fixed 1413369de |
| Registry RPCs unbounded | Fixed 51e38af06 |
| Ghost persisted-connected peer (bot) | Resolved upstream, verified on the current tip |
| Coverage gate | 73% → **94.6%** (threshold 80), gate now OK |

The first one is the significant one: `reconcileConnectionStates` came in with the upstream merge (963aa0af5) after this branch's p2p work, listed the registry with no transport filter, and derived liveness from `P2PClient.GetPeers()` — libp2p IDs only. Every connected legacy peer was cleared on every sweep, so `legacy_connected_peers_count` read 0 and every legacy row greyed out. The feature's main output was broken and none of my tests caught it, because they asserted the filter I passed rather than the state that resulted. The new test asserts end state.

Two deliberate departures from what was suggested, both argued in-thread:

1. **`SyncPeerID`** — converted the existing accessor to a mutex read and deleted `getSyncPeerMsg` and its handler arm, rather than adding `SyncPeerIDNoWait` alongside. It had exactly one sender and the netsync tests call the accessor, so leaving a blocking method beside a safe one just preserves the trap. All eight existing assertions pass unchanged.
2. **`byteDelta`** (from the earlier round) — reports a backwards counter's total as new bytes instead of clamping to zero.

Also correcting my own claim: the PR description said the loop "touches no wire path", and the `legacy_peerRegistryEnabled` longdesc says it "feeds no sync, catchup or peer-selection decision". True about decisions, but until 1413369de the loop put a message on `sm.msgChan` every 10s — the sync manager's single serialization point. It no longer does, so both statements are now accurate rather than needing to be weakened.

**One thing I have not done.** Sonar's remaining new issue is `go:S107` on `services/legacy/Server.go:147` — `legacy.New` now takes 11 parameters, over the authorised 10, because of the `peerRegistry` argument. It breaches no gate condition (gate is OK), and the clean fix is to group the dependencies into a struct, which touches the daemon and every test call site. I did not want to bundle that refactor into a review round unasked — say the word and I will, or it can be a follow-up.

**Verification on 51e38af06:**

```
go build ./...                                exit 0
go vet ./services/{legacy,p2p}/...            clean
go test ./services/{legacy,p2p,asset,blockchain}/...   all ok
golangci-lint --new-from-rev upstream/main    0 issues.
npm run check                                 802 FILES 0 ERRORS 14 WARNINGS
npm test                                      17 smoke passed, 40 unit passed
npm run build                                 done
```

`TestHandleWebSocket_InboundFrameLimits` failed once at exactly its 5s deadline during the parallel whole-tree run, then passed 3/3 in isolation and again for the whole p2p package. It asserts server-side close versus client read deadline, so it is load-sensitive rather than related to these changes.

ordishs:
Pushed 442b6055e, closing out the `go:S107` I flagged as outstanding.

`legacy.New` goes from 11 parameters to three — logger, settings, and a `legacy.Dependencies` struct carrying the nine service clients and stores. The `Server`'s own fields are unchanged, so this touches the constructor boundary only rather than rippling `s.blockchainClient` style accesses through the file.

Both call sites now name what they pass, which is worth having independently of the rule: the daemon wiring reads as a list of named dependencies, and the peer test no longer positions five bare nils.

One thing worth flagging for whoever reviews this, since it cuts against the change. No other service here uses a deps struct — `p2p.NewServer` takes 11 positional parameters and `subtreevalidation.New` takes 12, both already over the threshold and unflagged only because they are not new code. So `legacy.New` is now the odd one out stylistically. I think the struct is the better shape and it is what was asked for, but if the preference is consistency with the siblings over satisfying S107, reverting this one commit is clean and leaves everything else intact.

**Verification on 442b6055e:**

```
go build ./...                                exit 0
go vet ./services/legacy/... ./daemon/...     clean
go test ./services/legacy/... ./daemon/...    all ok
golangci-lint --new-from-rev upstream/main    0 issues.
```

Also worth correcting an earlier report: `daemon TestDaemon_Start_AllServices` now passes. I had reported it as a pre-existing environmental failure on `listen tcp 127.0.0.1:9091: bind: address already in use`, proven by reproducing it on a clean checkout of main. That port is free on this machine now and the test passes untouched, which confirms the diagnosis.

ordishs:
Follow-ups filed for the rest of the `go:S107` sites, so the `legacy.Dependencies` change in this PR is not left as a one-off.

I scanned the tree with an AST walk rather than grep, over exported non-method functions in non-generated, non-test files. Five sites exceed the 10-parameter limit:

| Site | Params | Issue |
|---|---|---|
| `services/asset/Server.go:87` `NewServer` | 12 | #1686 |
| `services/subtreevalidation/Server.go:175` `New` | 12 | #1687 |
| `services/p2p/Server.go:399` `NewServer` | 11 | #1688 |
| `services/legacy/netsync/manager.go:3152` `New` | 11 | #1689 |
| `services/blockvalidation/catchup/helpers.go:238` `CreateCatchupResultWithLocator` | 11 | #1690 |

`legacy.New` is now at three and off the list.

Two things I would rather state than leave to be discovered:

**One of these is partly this PR's doing.** `asset.NewServer` was at 11 parameters on `main`; commit b90e5e753 here added `peerRegistryClient` so `/api/v1/peers` could read the registry directly, taking it to 12. The rule was already breached, but this PR made it one worse. That is recorded in #1686.

**#1690 is a different fix.** `CreateCatchupResultWithLocator` is not a constructor taking service dependencies, so a `Dependencies` struct does not apply — it needs a parameter struct for its result inputs, or to be inlined at its call sites. Filed separately for that reason rather than lumped in.

Each issue carries the exact production and test call sites, and each repeats the counter-argument: no service in this repo uses a deps struct today, so the first few conversions make things temporarily less consistent. Worth doing as a sweep or not at all. `#1687` (subtreevalidation) and `#1688` (p2p) are the strongest candidates on their merits — both have runs of consecutive same-typed Kafka client parameters, where a transposition compiles clean and fails at runtime.

ordishs:
Correction to my previous comment: the five per-service issues are consolidated into a single tracker, **#1691**, and #1686 through #1690 are closed as duplicates of it.

The table in that earlier comment now points at closed issues. Read #1691 instead — it carries the same content: the five sites, their exact production and test call-site counts, the same-typed-parameter hazard for subtreevalidation and p2p, the note that this PR took `asset.NewServer` from 11 params to 12, and the note that `CreateCatchupResultWithLocator` needs a parameter struct rather than a deps struct and should not ride along with the sweep.

Consolidating was the right call for the reason I had already flagged as the counter-argument: no service in this repo uses a deps struct today, so converting them piecemeal makes the codebase temporarily less consistent rather than more. One tracker makes that go/no-go decision once instead of five times, and closing it with "we are leaving these as they are" is a legitimate outcome — better than the sweep half-landing across five separate issues.

Nothing in this PR changes. Tip is still 442b6055e.

ordishs:
@oskarszoon — apologies, I worked from @freemans13's confirmation of your P1 and did not read your review itself until now. Your review carries no inline comments, so it never appeared in the `pulls/comments` listing I was polling, and I only ever saw P1 referenced second-hand. Everything below P1 went unaddressed for two rounds as a result. That is my process failure, not an ambiguity in your review.

All of it verified against the code. Four fixed, one answered rather than changed.

**P1 — `reconcileConnectionStates`** — fixed in 782f82cf7, before I read this. Your diagnosis was exact, including that the missed call site is the only one that *writes*, and that `peer_registry_sync.go` could not repair it because the connection assert is once-only and pinned by `TestReconcile_UnchangedPeerCostsNoRegisterCall`. The test you asked for is in `legacy_peer_isolation_test.go` and asserts end state, not the filter argument.

**P2a — the card does not filter on connection state** — fixed in c9b0879fb. You were right on both counts, and both were my claims rather than the code's: the comment and the PR body said connection state, the filter selected on transport alone, and `is_connected` only drove `cellClass()` dimming.

Your supporting point checks out too and falsified a second comment of mine. `peer.go:2948` sets `p.addr = p.conn.RemoteAddr().String()` for inbound peers, so the comment at `peer_registry_sync.go:23` claiming the address is stable across reconnects was wrong for exactly the inbound case that makes accumulation possible. Corrected there as well.

I took both halves of your "either/or": a connected peer is always listed, a disconnected one only while its last message is within ten minutes, and the comment and PR body now describe that.

Worth flagging what this uncovered: the smoke fixture served timestamps 445 days old, so the *Teranode* table had been rendering zero rows and only the legacy card was ever exercised. The mock now rebases non-zero peer timestamps so the newest lands on now, preserving the fixture's relative ages — a zero stays zero, since the UI reads that as "Never". Both tables populate, and a third legacy peer that is disconnected and stale pins the bound. I checked that assertion is not vacuous by removing the bound and watching the row count fail.

**P2b — failed push then rebaselined** — fixed in e20ff63d5. Confirmed the asymmetry you describe: `RegisterPeer` continued on error, the metrics and last-message calls only logged and then flowed into the unconditional `p.lastSeen[snap.id] = snap`. What counts as pushed is now tracked per dimension, so a rejected call carries its previous baseline forward and the next tick reports its own bytes plus the ones that did not land. Two tests, both failing before the change.

**P2e — no first-tick reconcile, phantom `IsConnected`** — fixed in e20ff63d5. Your prediction held precisely: fixing P1 unmasked it. The first tick that produces real data now clears stale connected wire entries. It is gated on a non-nil snapshot, since nil means "no data" and not "nothing is connected", and it runs once rather than every tick. Both cases are tested, including that a libp2p entry is left alone.

**P2f — `legacy_settings.md`** — fixed in 55bfcc854. Both keys are in the table, with a section on what the mirror does and does not do.

**P2c — the v3 bump's rollback cost — I am not changing this, and want to argue it rather than quietly leave it.**

Your reading of the mechanism is right: a v3 blob read by a v2 binary returns `errFutureRegistryVersion`, and `Load` returns at `peer_registry_persistence.go:237` before repopulating `r.peers` and `r.banScores`, so the older binary runs with an empty registry — no persisted bans, no reputation for catchup selection.

Where I disagree is that not bumping is cheaper. Without the bump a v2 binary loads the v3 blob successfully, drops the `Legacy` field it cannot unmarshal, and then **saves that truncated envelope back**, silently destroying the persisted legacy block for the v3 binary you roll forward to. The existing version-history comment says the version-2 bump was made for exactly this reason. So the choice is a loud, self-healing degradation on rollback versus a silent, persistent data loss, and I would rather have the loud one.

Two things reduce the cost in practice: `saveDisabled` means the rollback does not overwrite the operator's blob, so rolling forward recovers the full registry intact, and the sentinel is logged at error level rather than swallowed.

If you would rather the rollback kept bans and reputation, the right fix is in the persistence layer, not this bump — have `Load` repopulate what it *can* parse before returning the sentinel. That is a change to shared registry behaviour affecting libp2p peers too, so I would rather do it as its own PR than fold it in here. Happy to open one if you agree that is the shape.

**Verification on 55bfcc854:**

```
go build ./...                                              exit 0
go vet ./services/legacy/...                                clean
go test -race -count=1 ./services/legacy/... \
    ./services/p2p/... ./services/asset/...                 all ok, 0 races
golangci-lint --new-from-rev upstream/main                   0 issues.
npm run check                                                802 FILES 0 ERRORS 14 WARNINGS
npm test                                                     17 smoke passed, 40 unit passed
npm run build                                                done
```

Full `make test` on the previous tip was 13634 tests, 46 skipped, 0 failures, 0 races. Still not put in front of a live legacy peer — that gap is unchanged and remains the honest limit on all of this.

freemans13 (approved):
Approving. All three of my findings are fixed, and I checked each one by breaking the fix in a scratch copy and watching the new test fail, rather than reading the diff and taking the result on trust.

**`SyncPeerID`.** You went further than I sketched and converted the accessor itself, deleting `getSyncPeerMsg` and its handler arm. That is the better call and I was wrong to propose the additive version. A safe method sitting beside a blocking one with the same job leaves the trap loaded for whoever calls it next, and `getSyncPeerMsg` had one sender. Restoring the old blocking body fails both new tests at the full five second guard, with `SyncPeerID blocked: it must not depend on the block handler running` and `SyncPeerID blocked after Stop`, so they cover both shapes.

**The connection sweep.** Reverting `transportHTTPFilter()` back to `nil` at `server_helpers.go:1383` fails `TestReconcileConnectionStates_LeavesWirePeersConnected` with exactly the message you quoted. The test asserts the resulting state, so it will still fail if the filter is dropped again, which is what I asked for.

**Registry RPC deadlines.** Every one of the six calls cancels immediately after its RPC rather than deferring, so nothing leaks a context across the `continue` arms in the per-peer loop. Per-call rather than per-tick is right; a whole-tick cap would make the bound depend on peer count.

I then went looking for what the fixes themselves might have broken, since three of these six commits are new code written under review pressure.

**The constructor refactor is the one I checked hardest, and it is correct.** Turning eleven positional arguments into a struct is where a silent swap lives, because the compiler cannot see it. `SubtreeStore` and `TempStore` are both `blob.Store`, so transposing them compiles clean and fails at runtime. Daemon passes `subtreeStore` and `tempStore` in that order and `New` reads them back in that order, so the mapping holds. Every other field pair differs by type. `daemon` tests pass, which exercises the real wiring.

**`adoptExistingEntries` avoids the mirror image of P1.** A first-tick pass that clears connected flags is one unfiltered `ListPeers` away from doing to libp2p peers exactly what the sweep was doing to wire peers. It filters on `TRANSPORT_WIRE_PROTOCOL`, so it cannot reach them. It also retries rather than marking itself done when the list call fails, so a blockchain service that is slow to come up does not skip the pass.

**The two remaining unfiltered `ListPeers` calls are both correct.** `getNodeStatusMessage` counts both transports on purpose, and `getPeerIDFromDataHubURL` matches on a non-empty DataHub URL, which a wire peer never has.

**The dashboard bound is not vacuous.** Removing `peer.is_connected || peer.last_message_time > legacyStaleBefore` takes the legacy table from two rows to three and fails the smoke test.

Verification on `55bfcc854`, in a throwaway worktree, nothing written to your branch:

```
go build ./...                                            exit 0
go vet ./services/legacy/... ./services/p2p/... ./daemon/...   clean
go test -race -count=1 ./services/legacy/ ./services/legacy/netsync/ \
    ./services/asset/httpimpl/ ./services/blockchain/ ./services/p2p/ \
    ./settings/ ./daemon/                                 all ok, 0 races
npm run test:unit                                         40 passed
npm run test:integration                                  17 passed
```

That closes the gap I left open last round. I said then that I had not run the dashboard suite; I have now, including the legacy card test.

Two things that change nothing in the code.

The branch was rewritten rather than added to, so `3d3e1f3df`, `1413369de` and `51e38af06`, cited in your thread replies, are not reachable from the branch. The same commit messages are there under `782f82cf7`, `6150461f8` and `39a5f6dca`. GitHub still resolves the old identifiers, so the links work, but anyone fetching the branch and running `git log` will not find them.

`reviewDecision` will stay `CHANGES_REQUESTED` on my approval alone, because oskarszoon's review predates these fixes. Their P1 and P2a are addressed and I have verified both, so it is worth asking them to take another look rather than waiting.

The remaining risk is the one your own description names, and I have not touched it: none of this has run against a live legacy peer. Everything above is tests and code reading.

oskarszoon (approved):
Re-reviewed at `55bfcc85`. The P1 is fixed at source and I verified the three observables rather than taking the commit at its word: `reconcileConnectionStates` is scoped to `transportHTTPFilter()` (`server_helpers.go:1383`), no unfiltered registry writer is left in the tree, and across three sweeps `LegacyConnectedPeersCount` holds, wire entries keep `IsConnected=true`, and the libp2p half still clears. The byte-delta fix is a correct rollback-on-failure rebaseline — neither loses nor double-counts. `adoptExistingEntries` lands in the same commit set as the sweep fix, so the phantom-peer window never opens.

Both scope changes check out. `6150461f8` is a real fix, though not the one its message describes: `loadSyncPeer`/`storeSyncPeer` already locked `syncPeerMu` on both sides, so `-race` was clean before and after — the defect was an unbounded wait on an unbuffered channel with no responder once `blockHandler` exits. Reverting it makes both new tests *time out* rather than race-fail. The locking is right (correct pre-existing mutex, no lock held across `sp.ID()`, no inversion), and the blast radius is one caller — `SyncPeerID` had no production callers at all at the merge-base. `442b6055e` is behaviour-preserving field-for-field, including the two same-typed `blob.Store` args where a positional-to-named refactor could have silently swapped. The `+518` in `+page.svelte` is ~95% Prettier; the Teranode table's predicate is unchanged and now has a smoke assertion guarding it.

Approving. Nothing below blocks, but the first two are worth picking up.

**`adoptExistingEntries` marks itself done even when its clears failed — `peer_registry_sync.go:306`**

The `ListPeers` failure path correctly returns without setting the flag. A per-peer `UpdateConnectionState` failure at `:306-313` only logs and `continue`s, while `p.adoptedExisting = true` at `:318` is unconditional. The pass never runs again, so that phantom sits at `IsConnected=true` until the TTL expires it — 24h by default, which is the outcome the commit exists to prevent. Reproduced: a client that fails only the clear on tick 1 and recovers on tick 2 leaves `adoptedExisting == true` and the peer still connected after tick 2. Track whether any clear failed and only set the flag when all succeeded.

**The new `SyncPeerID` tests are flaky under `-race` at full-suite parallelism — `sync_peer_id_test.go:22`**

Both use a hardcoded 5s deadline. `go test -race ./services/legacy/... ./services/p2p/...` failed both once ("SyncPeerID blocked") and passed on an immediate re-run; in isolation they finish in under 100ms. The locking is sound, so this is `-race` scheduling contention across ~15 parallel packages, not a deadlock — but CI runs exactly that command, so it'll flake and get rerun-normalised. 10–15s, or drop the fixed wall-clock budget.

**Smaller things**

- `peer_registry_sync.go:103` — the 5s bound is per-RPC, not per-tick. A first-sighting peer issues 5 sequential RPCs, so against a registry that accepts but never replies, 20 peers gives a ~500s single `reconcile()`. `legacy_settings.md` says this "delays a tick rather than stalling the mirror"; bounded-not-infinite is a real improvement, but ~8 minutes against a 10s cadence is a different claim. Wrap `reconcile` in one `WithTimeout(ctx, p.interval)`, or fix the doc.
- `peer_registry_sync.go:170` — the mirror asserts `IsConnected` only on first sighting, and every registry mutator no-ops silently on a peer the registry doesn't hold. Since `Load` forces `IsConnected=false`, a blockchain-service restart leaves a live wire peer flagged disconnected for the life of its connection, with no self-heal — libp2p peers recover via the batcher's reassert TTL. Pre-existing, not introduced here, and cosmetic (rows stay listed; `registrationEqual` sees `PingMicros` move every ~2min, so the entry is recreated if lost). A periodic bidirectional reconcile would close it; resetting `lastSeen` on RPC errors would not, since this path returns no error.
- The body still says netsync "does not touch that behaviour" — true behaviourally, but the sync engine's message loop was edited, and that assertion is what scoped my last review. Section 3 also still lists only four filter sites; the fifth, `reconcileConnectionStates`, is the actual P1 fix and the reason the mirror now solely owns wire-entry connection state. Worth the same correction you gave the card-filter claim.
- `routes.spec.ts:143` — the 600s bound is only tested at 120s and 7200s. Nothing near the boundary, and the predicate isn't unit-testable (`legacyRetentionSeconds` is a private const). Fixtures at 590s/610s would catch an off-by-one or a 4× drift.
- Still open from last pass, all non-blocking: the v3 downgrade still empties `r.peers`/`r.banScores` before returning; `LegacyAddressCell` still has no width constraint so the ellipsis won't fire; disconnected rows are still opacity-only with no ARIA; `legacyPeerSnapshots`' nil path is still untested.

PR #1612, opened 2026-08-18
fix(p2p): don't derive IP bans from relay-circuit multiaddrs

gokutheengineer:
**Closes bitcoin-sv/teranode#4748.**

### Problem

`extractIPFromMultiaddr` documented that it returns `""` for relay circuits, but the body just returned the first `/ip4` or `/ip6` value. For a circuit-relayed connection libp2p reports `/ip4/<RELAY-IP>/tcp/9905/p2p/<relayID>/p2p-circuit[/p2p/<peerID>]`, so banning a relayed peer wrote the RELAY's IP to the persisted ban list - collaterally banning the relay (e.g. the bootstrap node) and every peer behind it via gossip filtering (`isPeerIPBanned`), the periodic ban sweep (`disconnectPeersOnBanList`), and the shared HTTP 403 ban-list middleware. It also let an attacker redirect its own IP ban onto an attacker-chosen relay IP (IP-ban evasion via peer-ID rotation). `checkMultiaddrBanned` had the same defect in the dial path: it refused to dial any peer reachable only through a banned relay IP, and DNS-resolved the relay's `/dnsaddr` component of circuit addresses.

### Fix

- The circuit rule lives in one place: a new `isRelayCircuit(maddr)` helper (detected via `ValueForProtocol(ma.P_CIRCUIT)`, position-independent and immune to lookalike DNS names like `p2p-circuit.example.com`). `extractIPFromMultiaddr` now returns `""` for any relay-circuit multiaddr, so only directly-observed transport addresses can produce or match an IP ban. This fixes all four call sites at once: `onPeerBanned`, `isPeerIPBanned`, `disconnectPeersOnBanList`, and `checkMultiaddrBanned`. Peer-ID bans still cover circuit peers. A peer with both a circuit and a direct address still gets its direct IP banned.
- `checkMultiaddrBanned` parses the multiaddr once and returns `(false, nil)` for circuits before any IP or DNS check, so the relay's `/dnsaddr`/`/dns4` component is no longer resolved and checked against the ban list. Parse-failure behavior is unchanged (fail closed with an error), now pinned by a test.
- Comments on both functions updated to match actual behavior.

An independent review pass verified the premise in the deps (`circuitv2/client` builds `RemoteMultiaddr` as `<relay transport addr>/p2p/<relayID>/p2p-circuit`; the bus fills `PeerInfo.Addrs` from `conn.RemoteMultiaddr()` with relay enabled), confirmed no other multiaddr-IP extraction exists in the repo (legacy ban paths use `host:port`, HTTP/gRPC middleware use their own address sources), and ran a mutation check: every new test fails against the pre-fix code.

### Residual

- A banned peer with only circuit addresses now produces no IP-ban entry at all; it remains banned by peer ID (registry + `banStatusCache` + `IsPeerBanned` enforcement), which is the correct scope - now asserted by the tests.
- Deliberate tradeoff: `ConnectPeer` will now dial *through* an explicitly banned relay IP (the circuit's relay hop still opens a transport connection to that IP). The alternative - checking the pre-circuit components in the dial path - would reintroduce the self-eclipse this fixes (refusing every peer behind a banned relay). `ConnectPeer` is admin-only gRPC.
- Pre-existing, unchanged: an operator IP-only ban has no effect on a peer reaching the node through a relay, because the peer's real IP is not observable in a circuit address. Closing that via self-reported peerstore addresses would be spoofable and worse.

### Tests

Added to `services/p2p/ban_enforcement_test.go`:

- `TestExtractIPFromMultiaddr` extended with circuit cases (ip4/ip6 relay addrs, with and without target peer), plus a guard that circuit rows parse AND carry a literal IP component - so `""` provably comes from the circuit rejection.
- `TestOnPeerBanned_RelayedAddrDoesNotBanRelayIP` - mixed circuit + direct addrs: only the direct IP is banned.
- `TestOnPeerBanned_CircuitOnlyAddrsBanNoIP` - circuit-only peer produces no IP ban, while the peer-ID ban still takes full effect (banStatusCache primed, peer removed from the registry).
- `TestDisconnectPeersOnBanList_IgnoresRelayedPeersOfBannedRelayIP` - sweep no longer evicts peers behind a banned relay IP.
- `TestServer_ConnectPeer_AllowsCircuitThroughBannedRelayIP` and `TestServer_ConnectPeer_AllowsCircuitViaDNSAddrOfBannedRelay` - dial path no longer blocks circuits through a banned relay IP or resolves the relay's DNS name.
- `TestShouldSkipBannedPeer_RelayedPeerOfBannedRelayIPNotSkipped` - gossip from peers behind a banned relay is not dropped.
- `TestServer_ConnectPeer_FailsClosedOnInvalidMultiaddr` - pins the fail-closed ordering now that the parse is the first gate in `checkMultiaddrBanned`.

Run (affected package only):

```
```

Result: `ok github.com/bsv-blockchain/teranode/services/p2p 20.5s`.


ctnguyen (approved):
Review at commit [0ecbc86](https://github.com/gokutheengineer/teranode/commit/0ecbc8678d9d0a1b5d46c742bb5c8c9bc8e5b9f3)

This is a tight, well-scoped fix and the core of it is right. Putting the circuit rule in one `isRelayCircuit` helper built on `ValueForProtocol(ma.P_CIRCUIT)` is the correct primitive — it is genuinely position-independent and cannot be fooled by a hostname that merely contains the token, which a substring check would have been. The premise rests on two properties of the dependencies rather than on anything in this repository — that `PeerInfo.Addrs` is filled from live `conn.RemoteMultiaddr()` values rather than from self-reported peerstore addresses, and that circuit-v2 always joins the `/p2p-circuit` component onto the relay's transport address — both of which the linked issue's verification traced in the dependency source; given them, every relayed address the ban logic can observe is caught and only directly-observed transports can still produce an IP ban. Reordering `checkMultiaddrBanned` so the parse is the first gate is both more correct and cheaper than the old parse-twice shape, and the tests are specific rather than decorative. My notes below are about what the change gives up and what the description claims — not about the logic, which I could not break.

## Non-blocking issues

### ChiR1 - Relay-reached peers become unrestrainable by ID rotation
**Problem:** A peer whose only observed address is a circuit now produces no `banList` entry at all (`services/p2p/Server.go:2289-2310`), so enforcement rests entirely on the peer ID. Before this change the relay's IP landing in `banList` had a second effect that the Residual section does not mention: because `isPeerIPBanned` (`services/p2p/server_helpers.go:1020-1031`) and `disconnectPeersOnBanList` (`services/p2p/server_helpers.go:1305-1324`) match on the *extracted* IP, that entry also matched every *other* identity arriving over the same relay. An attacker spamming from behind a relay it controls was therefore stopped even after rotating its peer ID. After this change, rotating the Ed25519 key yields a peer with no registry score, no cached ban status, and a circuit address that matches nothing — the 5-6-message ban cycle restarts from scratch at zero cost.
**Why it matters:** The IP-ban-evasion half of the linked issue's impact survives the fix; the node no longer has any enforcement granularity coarser than a peer ID for peers it can only see through a circuit, so one host can spam indefinitely.
**Fix:** Keep the change — the collateral it removes is strictly worse than what it gives up — but state this in the Residual section instead of describing the peer-ID ban as "the correct scope", and track a follow-up. The workable direction is a throttle rather than a ban, so it cannot self-eclipse: a shared, non-persisted concurrency or score budget keyed on the circuit's relay hop, applied to all peers arriving via that hop, e.g.

```go
// relayHopKey returns a grouping key for peers reached over the same relay.
// It must never be written to the ban list - it names the relay, not the peer.
func relayHopKey(maddr ma.Multiaddr) (string, bool) {
	if !isRelayCircuit(maddr) {
		return "", false
	}
	// ... relay transport component, used only for rate limiting
}
```

### ChiR2 - The `Closes` trailer retires a two-finding issue on one finding
**Problem:** The PR body opens with `Closes bitcoin-sv/teranode#4748`, which auto-closes that issue at merge. The issue carries two findings and this diff addresses only the second. The first — the ban list is never plumbed into a libp2p `ConnectionGater` — is untouched, and the code being edited still documents the consequence: `listenForBanEvents` polls precisely because "we cannot refuse dials from banned addresses" (`services/p2p/server_helpers.go:1155-1157`), and `networkDisconnector` remains unimplemented by the bus (`services/p2p/server_helpers.go:1327-1334`) — this PR even widens that comment's version window rather than removing it.
**Why it matters:** Merging silently closes a tracked security finding that the change does not fix.
**Fix:** Make the trailer non-closing and name the sub-finding, e.g. `Addresses finding 2 of bitcoin-sv/teranode#4748`, leaving the issue open for the gater work. If the intent is that the gater gap is already tracked elsewhere (issue 1163 is cited as the duplicate), say that explicitly in the body so the closure is a deliberate hand-off rather than an accident.

### ChiR3 - `ConnectPeer` doc comment now overstates the ban check
**Problem:** `services/p2p/Server.go:2156-2157` still reads "Banned addresses are refused." With the new early return, any multiaddr carrying a `/p2p-circuit` component is waved through without a single ban lookup (`services/p2p/server_helpers.go:1259-1261`), so `/ip4/<banned-ip>/tcp/9905/p2p/<anyID>/p2p-circuit/p2p/<anyID>` is dialed — and by the PR's own Residual section, the circuit's relay hop still opens a transport connection to that banned IP. The PR updated the comments on `extractIPFromMultiaddr` and `checkMultiaddrBanned` for exactly this reason but left this one behind.
**Why it matters:** It is the contract for an admin-facing RPC, and it now promises a guarantee the code intentionally no longer provides.
**Fix:** Correct the comment to match the new exemption:

```go
// ConnectPeer connects the P2P client to the peer at the given multiaddr.
// Addresses on the IP/subnet ban list are refused. Relay-circuit addresses are
// exempt: their IP names the relay hop, not the dial target, so the dial may
// still transit a banned relay IP (see checkMultiaddrBanned).
```

### ChiR4 - Two verification claims outrun what the tests actually pin
**Problem:** Two things. First, the description states "every new test fails against the pre-fix code", but `TestServer_ConnectPeer_FailsClosedOnInvalidMultiaddr` (`services/p2p/ban_enforcement_test.go:369-381`) passes on both sides: pre-fix, `extractIPFromMultiaddr("not-a-multiaddr")` already returned `""`, so control fell through to the same `ma.NewMultiaddr` error and the same fail-closed response. It is a reasonable regression pin, just not a mutation-detecting one. Second, the "immune to lookalike DNS names like `p2p-circuit.example.com`" claim has no covering case — every circuit row in `TestExtractIPFromMultiaddr` carries a real `/p2p-circuit` component, and the new guard itself selects rows with `strings.Contains(tt.addr, "/p2p-circuit")` (`services/p2p/ban_enforcement_test.go:87`), so a substring-based implementation would keep the whole suite green.
**Why it matters:** The two properties the fix is sold on — parse-order fail-closed and lookalike immunity — are the two the suite does not discriminate on.
**Fix:** Soften the mutation-check wording, and add one case a substring implementation would fail. A hostname containing the token is not a circuit and must still be resolved and fail closed, mirroring the existing `.invalid` pattern:

```go
func TestServer_ConnectPeer_FailsClosedOnLookalikeCircuitDNSName(t *testing.T) {
	s, _ := newServerWithLocalRegistry(t)
	s.banList = &recordingBanList{}
	client := &MockServerP2PClient{}
	s.P2PClient = client

	resp, err := s.ConnectPeer(context.Background(), &p2p_api.ConnectPeerRequest{
		PeerAddress: "/dns4/p2p-circuit.does-not-exist.invalid/tcp/9905/p2p/12D3KooWAfBVdmphtMFPVq3GEpcg3QMiRbrwD9mpd6D6fc4CswRw",
	})
	require.NoError(t, err)
	require.False(t, resp.Success, "a hostname containing p2p-circuit is not a circuit and must still fail closed")
	client.AssertNotCalled(t, "Connect", mock.Anything, mock.Anything)
}
```

## Recap

| ChiR1 | Relay peers evade by ID rotation |  | 40% |
| ChiR2 | Closes trailer retires unfixed finding |  | 32% |
| ChiR3 | Stale ConnectPeer ban comment |  | 15% |
| ChiR4 | Test claims outrun coverage |  | 12% |

services/p2p/server_helpers.go
```diff
@@ -1447,26 +1461,40 @@ func extractIPFromMultiaddr(addrStr string) string {
...
-	if ip := extractIPFromMultiaddr(addrStr); ip != "" {
-		return s.banList.IsBanned(ip), nil
-	}
-
 	maddr, err := ma.NewMultiaddr(addrStr)
 	if err != nil {
 		return false, errors.NewInvalidArgumentError("invalid multiaddr %s: %v", addrStr, err)
 	}
 
+	if isRelayCircuit(maddr) {
```
liam, line 1490: Nit, and it does not change my view of the PR: this bypass is syntactic rather than semantic. Appending `/p2p-circuit/p2p/<id>` to any address makes `checkMultiaddrBanned` skip the ban check, with no actual relay involved.

`ConnectPeer` is admin-only and an admin can call `ClearBanned` anyway, so it is not an escalation. But it is a slightly stronger statement than the "will now dial *through* an explicitly banned relay IP" framing in the PR body, which reads as though a real relay hop is required. Might be worth a sentence in the residual section so nobody later reads that guarantee as tighter than it is.

services/p2p/server_helpers.go
```diff
@@ -1428,14 +1428,28 @@ func (s *Server) handleBanEvent(ctx context.Context, event BanEvent) {
...
 		return ""
 	}
 
+	return extractIPFromParsedMultiaddr(maddr)
+}
+
+// extractIPFromParsedMultiaddr is extractIPFromMultiaddr for an
+// already-parsed multiaddr; it applies the same relay-circuit rejection.
+func extractIPFromParsedMultiaddr(maddr ma.Multiaddr) string {
+	if isRelayCircuit(maddr) {
```
liam, line 1449: Minor: on the `checkMultiaddrBanned` path this `isRelayCircuit` check is redundant, since that function already returns early for circuits before calling here. Harmless defence in depth for the other caller, just flagging that it is a second call to `ValueForProtocol` on a path that already answered the question.

liam (commented):
Reviewed the diff, traced the surrounding code, and ran `go vet` plus `-race` tests on the branch (`./services/p2p/...` green).

**This one looks right to me.** Things I checked rather than assumed:

- `ValueForProtocol(P_CIRCUIT)` returns `("", nil)` when the component is present (go-multiaddr v0.16.1, `multiaddr.go:245`), so `err == nil` is a correct presence check for a zero-size protocol. That is not obvious from the name and it is the load-bearing detail of the fix.
- All four call sites are covered (`server_helpers.go:1097`, `:1463`, `:1512`, `Server.go:2642`), and there is no other multiaddr-to-IP extraction anywhere in the repo.
- The security-critical direction holds: the addresses feeding `onPeerBanned` / `isPeerIPBanned` come from `P2PClient.GetPeers()`, i.e. `conn.RemoteMultiaddr()`, which libp2p observes rather than the peer asserting. So a peer cannot fabricate "I am a circuit" to shed its own IP ban. That was my main worry and it checks out.
- The `checkMultiaddrBanned` reordering preserves fail-closed on parse error, and the new test pins it.

Two nits inline, neither blocking.

liam (approved):

PR #1614, opened 2026-08-18
docs: remove orphaned FoundationDB sharding diagrams

icellan:
## Summary
- Removes `docs/sharding/write.puml` and `docs/sharding/topology.puml`, leftovers from a 2023 FoundationDB sharding spike that was never pursued.
- Confirmed nothing else in `docs/` references this directory, so nothing is broken by removing it.
- These diagrams could otherwise mislead a reader into thinking sharding via FoundationDB is a current design direction.

## Test plan
- Doc-only deletion; no build/test impact.


ctnguyen (approved):
Review at commit [785eed5](https://github.com/icellan/teranode/commit/785eed51ed2582599789cbda7646bec332b1cb43)

Nice bit of low-risk housekeeping, and the rationale holds up on inspection. Nothing in the tree references `docs/sharding`, `topology.puml`, or `write.puml` — not the docs pages, `mkdocs.yml`'s nav, the `Makefile`, `scripts/`, `docs/hooks/`, or any workflow; in fact no `.puml` path is referenced from any markdown, config, or script in the repo, because pages embed committed image files instead. There's also no build-time diagram-rendering step: `docs-deploy.yml` installs mkdocs-material and runs `mkdocs gh-deploy`, and `exclude_docs` only drops `hooks/`, so these sources were being copied to the site as inert static files. And the two deleted files had no committed render beside them — `docs/sharding/` held exactly these two sources, which is why the directory vanishes entirely with this diff and no orphaned image is left behind. The FoundationDB framing checks out too: the `Stateless Proxy` / `Storage Server` / `Shard` vocabulary is FoundationDB's own, and no other trace of it survives anywhere in the repo. Two clean deletions, no collateral, and the history keeps the files if the idea is ever revisited — nothing to flag.

ordishs (approved):

PR #1619, opened 2026-08-18
Remove unused id-token: write permission from CI workflows

icellan:
## Summary

An earlier fix removed AWS credential access from CI entirely (moved off long-lived keys), but several workflow files kept declaring an `id-token: write` permission left over from before that change. Nothing in these workflows consumes it — no `aws-actions/configure-aws-credentials` or other OIDC-consuming action is present.

- Removed unused `id-token: write` from:
  - `.github/workflows/teranode_main_tests.yaml`
  - `.github/workflows/nightly_tests.yaml`
  - `.github/workflows/teranode_pr_tests.yaml`
  - `.github/workflows/teranode_pr_smoketests.yaml`
  - `.github/workflows/claude-code-review.yml`
  - `.github/workflows/chainintegrity.yaml`
  - `.github/workflows/teranode_main_build.yaml` (`test_and_lint` job's own permission block, granted only for the now-cleaned-up local callee)
- Deleted a dead commented-out `chainintegrity` job block in `teranode_main_build.yaml` left over from the same cleanup.

Left alone: the `build-and-deploy` job in `teranode_main_build.yaml`, which calls the reusable workflow `bsv-blockchain/teranode-shared-workflows/.github/workflows/build_and_deploy.yaml@main`. That permission is passed through to a workflow defined in a different repository, so whether it's actually consumed there can't be verified from this repo — left untouched pending confirmation from that repo's maintainers.

## Follow-up (out of scope here)

`bsv-blockchain/teranode-shared-workflows` reportedly has ~3 more instances of the same leftover `id-token: write` permission. That's a separate repository, so cleaning those up is a follow-up for that repo's maintainers.

## Test plan

- [x] Confirmed via grep that none of the touched workflows reference `aws-actions/configure-aws-credentials` or any other OIDC-consuming action
- [x] Validated YAML syntax for all touched files (`YAML.load_file`)

### DIFF
--- a/.github/workflows/chainintegrity.yaml
+++ b/.github/workflows/chainintegrity.yaml
@@ -6,7 +6,6 @@ on:
 permissions:
   contents: read
   pull-requests: read
-  id-token: write
   packages: read
 
 env:
@@ -79,26 +78,26 @@ jobs:
           REQUIRED_HEIGHT=120
           MAX_ATTEMPTS=120  # 5 minutes with 5s sleep
           SLEEP=5
-          
+
           # Function to check for errors in all teranode container logs at once
           check_errors() {
             # Get current time for this check
             local current_time
             current_time=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
-            
+
             # Check for errors - if last_check_time is empty, it will check all logs
             local since_param=""
             if [ ! -z "$last_check_time" ]; then
               since_param="--since=$last_check_time"
             fi
-            
+
             # Single command pattern that works for both initial and subsequent checks
             local errors
             errors=$(docker compose -f compose/docker-compose-chainintegrity.yml logs --no-color $since_param teranode1 teranode2 teranode3 | grep -i "| ERROR |" || true)
-            
+
             # Update timestamp for next check
             last_check_time=$current_time
-            
+
             if [[ ! -z "$errors" ]]; then
               echo "ERROR: Found error logs in teranode containers:"
               echo "$errors"
@@ -106,22 +105,22 @@ jobs:
             fi
             return 0
           }
-          
+
           # Initialize empty for first check to get all logs
           last_check_time=""
-          
+
           for ((i=1; i<=MAX_ATTEMPTS; i++)); do
             h1=$(curl -s http://localhost:18090/api/v1/bestblockheader/json | jq -r .height)
             h2=$(curl -s http://localhost:28090/api/v1/bestblockheader/json | jq -r .height)
             h3=$(curl -s http://localhost:38090/api/v1/bestblockheader/json | jq -r .height)
             echo "Attempt $i: heights: $h1 $h2 $h3"
-            
+
             # Check for errors in all teranode containers
             if ! check_errors; then
               echo "Errors found in container logs. Exiting."
               exit 1
             fi
-            
+
             if [[ -z "$h1" || -z "$h2" || -z "$h3" ]]; then
               if [[ $i -gt 10 ]]; then
                 echo "Error: One or more nodes are not responding after 10 attempts. Exiting."
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -21,7 +21,6 @@ on:
 permissions:
   contents: read
   pull-requests: write
-  id-token: write
 
 # Cancel any in-flight CCR run for the same PR when a new one starts.
 # Prevents zombie "In Progress" sticky comments when two reviews race on the same PR.
--- a/.github/workflows/nightly_tests.yaml
+++ b/.github/workflows/nightly_tests.yaml
@@ -13,7 +13,6 @@ env:
 permissions:
   contents: read
   pull-requests: read
-  id-token: write
   packages: write
 
 jobs:
--- a/.github/workflows/teranode_main_build.yaml
+++ b/.github/workflows/teranode_main_build.yaml
@@ -20,7 +20,6 @@ jobs:
     permissions:
       contents: read
       pull-requests: read
-      id-token: write
     secrets: inherit
 
   # Calculate version information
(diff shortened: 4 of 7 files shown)


ctnguyen (approved):
Review at commit [5a5a29f](https://github.com/icellan/teranode/commit/5a5a29f8e31e006e721f2cea6de3d2d9da464d2b)

Nice bit of least-privilege housekeeping, and it's the kind of cleanup that usually never happens. The reasoning holds up: `id-token: write` only does anything if a step asks the runner for an OIDC token, and after this change the only declaration left in the repo is on `build-and-deploy` (`.github/workflows/teranode_main_build.yaml:48`) — so the sweep is complete for everything that can be judged from here. I also like that you didn't strip the cross-repo callee just because it matched the same string; leaving it and saying plainly in the description that it can't be confirmed from this repo is the right call, and the description is generally honest about what was and wasn't verified. Nothing here blocks merge. What I'd raise is that the same file you edited still carries a second stale grant, the same pattern shows up on other keys in the other touched files, and one deletion took a useful piece of context with it.

## Non-blocking issues

### ChiR1 — `teranode_main_build.yaml` keeps a stale `pull-requests: read` in the block this PR edits

**Problem:** This PR removes `id-token: write` from the `test_and_lint` permission block but leaves `pull-requests: read` beside it (`.github/workflows/teranode_main_build.yaml:22`), and the same scope sits in the workflow-level block at line 14. Nothing consumes it. The workflow triggers only on `push` (lines 3-10: `v*` tags and `main` / `staging` / `release/*`), so there is no pull request in context to begin with. In the caller, `calculate-version` only checks out and runs `scripts/determine-git-version.sh`, and `build-and-deploy` declares its own block (lines 45-48) that doesn't include the scope. In the callee reached through `test_and_lint`, none of the four jobs touches pull-request data either: checkout, Docker Hub login, `setup-go`, the local `go-cache` action (`actions/cache` only, `.github/actions/go-cache/action.yml:37-77`), `make test`, artifact upload/download, the two shell check scripts, `golangci-lint-action` (invoked without `only-new-issues`, so no diff fetch), and the Sonar scan, which authenticates with `SONAR_TOKEN` rather than the workflow token.

**Why it matters:** This is precisely the leftover this PR set out to remove, one line above the line it removed, in a file already open in the diff. Leaving it means the next person doing this sweep has to re-derive the same conclusion — and a dead scope in a push-triggered workflow is the sort of thing that quietly justifies itself later ("it's always been there").

**Fix:** Drop the scope from both sides in the same commit. GitHub caps a called workflow's permissions to what the caller grants, so trimming only the caller's block while `teranode_main_tests.yaml:16` still asks for `pull-requests: read` leaves the callee requesting something it is no longer granted — at best redundant, and I'd rather not rely on it being silently intersected:

```yaml
# teranode_main_build.yaml — workflow level and the test_and_lint job
permissions:
  contents: read

# teranode_main_tests.yaml — the callee, same commit
permissions:
  contents: read
```

### ChiR2 — The same leftover pattern on other permissions in the other touched files

**Problem:** `id-token: write` isn't the only grant in these workflows without a consumer:

- `nightly_tests.yaml:16` grants `packages: write`, but the workflow only builds locally (`load: true` at lines 51-57, no `push:` and no `docker push` anywhere) and logs into GHCR with `GITHUB_TOKEN` to pull; the `type=gha` cache goes to the Actions cache, not the package registry. `packages: read` covers everything it does.
- `teranode_pr_tests.yaml:17` grants `pull-requests: write`, but nothing in the file calls the pull-request API — the only step that would have (`npx github-actions-ctrf`) is commented out at lines 144-146, and the Sonar handoff is a separate `workflow_run` workflow with its own permissions.
- `teranode_pr_smoketests.yaml:14` and `chainintegrity.yaml:8` grant `pull-requests: read`, but the only PR data used is from the event payload (`github.event.pull_request.head.repo.full_name`), which needs no token scope.

**Why it matters:** `pull-requests: write` in `teranode_pr_tests.yaml` is the one I'd act on: that workflow checks out and executes branch code via `make test`. Fork PRs get a read-only token regardless of the block, so the grant is inert there, but for internal branch PRs it hands a write-scoped token to a job whose entire purpose is running arbitrary test code from that branch. Same class of unnecessary standing capability as the one this PR is removing.

**Fix:** Trim them in the same sweep, since you're editing these exact blocks:

```yaml
# nightly_tests.yaml
permissions:
  contents: read
  packages: read

# teranode_pr_tests.yaml
permissions:
  contents: read
```

### ChiR3 — Deleting the commented-out job also deleted the only record of why it was disabled

**Problem:** The removed block was the only place in the repo that said why the chain-integrity reusable workflows aren't wired up: they depend on the private `teranode-commands` / `teranode-coinbase` packages. To be clear about causation — this PR doesn't make anything unreachable. `chainintegrity.yaml` and `chainintegrity-3blasters.yaml` are `on: workflow_call` only (`.github/workflows/chainintegrity.yaml:3-4`, `.github/workflows/chainintegrity-3blasters.yaml:3-4`) and a commented-out `uses:` was never a caller, so they had no live caller before this PR either. What changes is that the explanation is now gone while ~200 lines of uncalled YAML in each file remain, with nothing in the repo pointing at them (their scenario is also already covered by the inline job in `nightly_tests.yaml:20-178`).

**Why it matters:** Uncalled workflow files still appear live in the Actions listing and keep resurfacing in future sweeps — this PR is itself an example, since it spent a line trimming a permission in a file that nothing invokes. Without the "why disabled" note, whoever next wonders whether to delete or revive them has to rediscover the private-package dependency from scratch.

**Fix:** Either delete both files here (cleanest end state, given `nightly_tests.yaml` already covers the scenario), or keep the rationale where it survives — a header comment in each file, plus a tracking issue:

```yaml
# Not called by any workflow: depends on the private teranode-commands /
# teranode-coinbase packages. See <issue link> before deleting or reviving.
name: Chain Integrity Test
```

## Recap

| ChiR1 | Stale PR scope in edited block | | 35% |
| ChiR2 | Other unused permission grants | | 30% |
| ChiR3 | Lost disabled-job rationale | | 20% |

ordishs (approved):

PR #1625, opened 2026-08-18
docs(chainintegrity): correct comment on harness relationship

icellan:
## Summary
- The Go-based chain integrity test's header comment claimed it "replaces" the compose-based `docker-compose-3blasters.yml` nightly test. That's not accurate: the compose-based harness still runs nightly (`nightly_tests.yaml`, cron `0 2 * * *`) against a full docker-compose stack, while the Go harness (`test/e2e/chainintegrity`) runs on every PR via `make chainintegrity` in `teranode_pr_smoketests.yaml`. They exercise overlapping scenarios in different environments and are complementary, not one superseding the other.
- Reworded the comment to describe the relationship accurately.

## Note
Separately, the reusable `chainintegrity.yaml` / `chainintegrity-3blasters.yaml` workflows are currently disabled (commented out) in `teranode_main_build.yaml`. That's an unrelated, pre-existing state and not something this PR touches — if those are genuinely dead weight, retiring them would be a separate decision.

## Test plan
- [x] `go vet ./test/e2e/chainintegrity/...` (comment-only change)

### DIFF
--- a/test/e2e/chainintegrity/chain_integrity_test.go
+++ b/test/e2e/chainintegrity/chain_integrity_test.go
@@ -33,7 +33,9 @@ const (
 )
 
 // TestChainIntegrity3Nodes tests chain integrity across 3 nodes with transaction load
-// This test replaces the docker-compose-3blasters.yml approach with a pure Go implementation
+// This is a pure Go implementation of the same scenario as docker-compose-3blasters.yml
+// (which still runs nightly against a full docker-compose stack); the two are
+// complementary, not a replacement of one by the other.
 func TestChainIntegrity3Nodes(t *testing.T) {
 	if testing.Short() {
 		t.Skip("Skipping chain integrity test in short mode")
--- a/test/e2e/chainintegrity/chain_integrity_test.go
+++ b/test/e2e/chainintegrity/chain_integrity_test.go
@@ -33,9 +33,12 @@ const (
 )
 
 // TestChainIntegrity3Nodes tests chain integrity across 3 nodes with transaction load
-// This is a pure Go implementation of the same scenario as docker-compose-3blasters.yml
-// (which still runs nightly against a full docker-compose stack); the two are
-// complementary, not a replacement of one by the other.
+// This is the in-process Go counterpart to the compose/docker-compose-3blasters.yml harness
+// (still run nightly): the two overlap but are complementary, not one superseding the other.
+// This test trades scale and real container-level P2P for fast, deep structural verification
+// (per-node block hashes, UTXO store, subtree fees, topological order, persister files); the
+// compose harness runs a much longer chain with an SV node in the network and establishes
+// node agreement by log hashing instead.
 func TestChainIntegrity3Nodes(t *testing.T) {
 	if testing.Short() {
 		t.Skip("Skipping chain integrity test in short mode")


ordishs (approved):
Approve. Comment-only 3-line change; all factual claims in the PR body verified:

- Compose-based `docker-compose-3blasters.yml` still runs nightly — `.github/workflows/nightly_tests.yaml` job `chainintegrity`, cron `0 2 * * *`.
- Go harness runs on every PR — `.github/workflows/teranode_pr_smoketests.yaml` job `chainintegrity` → `make chainintegrity` (`Makefile:264-268`).
- The reusable `chainintegrity*.yaml` workflows are indeed commented out in `teranode_main_build.yaml` with a stated reason (private `teranode-commands`/`teranode-coinbase` packages) — pre-existing and correctly out of scope here.
- `go vet ./test/e2e/chainintegrity/...` clean locally.

The old wording ("This test replaces the docker-compose-3blasters.yml approach") was genuinely stale, so this is a real fix.

Minor, non-blocking nits:

1. "the same scenario as" reads as parity, but the Go harness is a scaled-down analogue (`targetBlocks = 10`, `txsPerBlock = 10`, `numWorkers = 2`) versus real block-generator containers on a full stack. The PR description's own phrasing — "overlapping scenarios in different environments" — is more precise and would be worth pulling into the comment.
2. Comments that reference CI wiring go stale exactly the way the one being fixed did. Dropping the "which still runs nightly" clause and keeping just the "complementary" statement would be more durable.

No behavioural change, no risk.

ctnguyen (changes requested):
Review at commit [d2d5561](https://github.com/icellan/teranode/commit/d2d5561041588facc6d55f070d1a46942e09af6e)

Good instinct here — a stale "this replaces X" comment is exactly the kind of thing that quietly justifies deleting a job that's still earning its keep, and the diff is minimal and self-contained. I checked the two claims behind the change and both hold: `nightly_tests.yaml` really does still run `compose/docker-compose-3blasters.yml` on `cron: '0 2 * * *'` (`.github/workflows/nightly_tests.yaml:4,66`), and `make chainintegrity` really does run the Go harness on every PR (`.github/workflows/teranode_pr_smoketests.yaml:167` → `Makefile:265-268`). The description's note about the disabled reusable workflows is accurate too, and correctly scoped out. My one objection is that the replacement wording swaps one overstatement for another, and it's a smaller claim than the one you already wrote in the PR description.

## Blocking issues

### ChiR1 — "the same scenario" overstates the overlap

**Problem:** `test/e2e/chainintegrity/chain_integrity_test.go:36` now says the Go test is "a pure Go implementation of the same scenario as docker-compose-3blasters.yml". The two harnesses do not run the same scenario, in several ways that matter:

- **Scale.** The Go test caps out at `targetBlocks = 10` with `coinbaseMaturity = 5` and `txsPerBlock = 10` across `numWorkers = 2` (lines 29-32). The compose stack's `block-generator-1` is launched with `--initialBlocks 100 --stopHeight 1000` (`compose/docker-compose-3blasters.yml:334-354`), and the nightly job gates on `REQUIRED_HEIGHT=120` (`.github/workflows/nightly_tests.yaml:71,125-130`). Two orders of magnitude apart.
- **What's in the network.** The compose stack also brings up `svnode1`, a legacy SV node peer, with its profile deliberately commented out so it starts by default (`compose/docker-compose-3blasters.yml:98-107`). The Go test has three `TestDaemon` instances and nothing else.
- **Isolation.** Compose gives each node its own Aerospike, Prometheus and container boundary with real cross-process P2P; the Go test runs three daemons in one process, sharing a parent data dir (`SkipRemoveDataDir: nodeNumber > 1`, line 117).
- **Verification method — the biggest one.** These two harnesses answer different questions. The compose side never asserts that the three nodes agree on a chain: its wait loop only polls each node's `bestblockheader` and checks the three heights are individually numeric and `>= REQUIRED_HEIGHT` (`.github/workflows/nightly_tests.yaml:105-130`), and its sole agreement signal comes afterwards from `chainintegrity.run` comparing *filtered log-file hashes* for majority consensus (`.github/workflows/nightly_tests.yaml:156-167`, `Makefile:684-693`). The Go test does no log comparison at all: it compares every block hash across all three nodes directly (lines 247-259) and then runs deep structural checks over UTXO store spends, subtree fees, topological ordering and block-persister files (lines 262-282).

Your own description gets this right — "They exercise overlapping scenarios in different environments and are complementary". The comment is the weaker version of that sentence.

**Why it matters:** The whole point of this PR is that a comment nudged a reader toward "the nightly job is redundant, drop it." "Same scenario, one runs per PR and one runs nightly" nudges toward the identical conclusion, just one step further down the road — the reader now thinks the nightly run is a slower duplicate rather than the only thing covering 1000-block runs, real container-level P2P, SV-node presence, and log-hash consensus. Since the comment is the entire deliverable of this PR, it's worth landing on the accurate claim.

**Fix:** Use the description's wording, say what each side actually buys, and give the compose file its `compose/` path so the reference is checkable:

```go
// TestChainIntegrity3Nodes tests chain integrity across 3 nodes with transaction load
// This is the in-process Go counterpart to the compose/docker-compose-3blasters.yml harness
// (still run nightly): the two overlap but are complementary, not one superseding the other.
// This test trades scale and real container-level P2P for fast, deep structural verification
// (per-node block hashes, UTXO store, subtree fees, topological order, persister files); the
// compose harness runs a much longer chain with an SV node in the network and establishes
// node agreement by log hashing instead.
```

icellan:
ChiR1 — fixed in 493153e55. Checked all four differences before rewriting:

- scale: `targetBlocks = 10` / `txsPerBlock = 10` / `numWorkers = 2` vs `--initialBlocks 100 --stopHeight 1000` and `REQUIRED_HEIGHT=120`
- `svnode1` is in the compose stack with its `profiles:` block commented out, so it starts by default; the Go test has three `TestDaemon` instances and nothing else
- in-process daemons sharing a parent data dir vs container-level P2P
- verification: the nightly wait loop only checks three heights are numeric and above the threshold, agreement comes from `chainintegrity.run` comparing filtered log-file hashes; the Go test compares every block hash directly and then checks UTXO spends, subtree fees, topological order and persister files

Took your replacement wording as-is, including the `compose/` path prefix.

`go build ./...` and `go vet ./test/e2e/chainintegrity/...` both clean.

ctnguyen (approved):
Review at commit [493153e](https://github.com/icellan/teranode/commit/493153e55828f6b8f60a3750aa5d1b9d5408d223)

The rewrite landed well. The new comment at `repo/test/e2e/chainintegrity/chain_integrity_test.go:35-41` is the wording from the last round, and it holds up against the code: the block-hash comparison across all three nodes is at lines 247-262, the UTXO/subtree-fee/topological checks at 307-371 (`subtreeFees != subtree.Fees` at 480, `topological order violations` at 371), and the persister-file verification at 581 — so "per-node block hashes, UTXO store, subtree fees, topological order, persister files" is a fair inventory rather than a slogan. The compose side of the sentence checks out too: `block-generator-1` runs `--initialBlocks 100 --stopHeight 1000` with no `profiles:` key, so it starts by default (`repo/compose/docker-compose-3blasters.yml:334-354`), `svnode1` has its `profiles:` block commented out and `legacy_connect_peers.docker.teranode1.test = svnode1:18333` wires teranode1 to it (`repo/compose/docker-compose-3blasters.yml:98-105`, `repo/compose/settings_test.conf:415`), and node agreement really does come from filtered log-file hashes rather than from any chain comparison — the comparison is implemented in `repo/compose/chainintegrity/chain_integrity.go:123-227`, which sha256s each node's filtered log file (`sha256.Sum256(filteredData)` at :209) and decides majority consensus from those digests alone; the nightly workflow only wires it in and gates on its output (`repo/.github/workflows/nightly_tests.yaml:156-167`). ChiR1 is fixed — the overstatement is gone and nothing overstated took its place.

PR #1606, opened 2026-08-18
feat(legacy): download blocks from every eligible peer, and keep the ones that arrive early

freemans13:
## What happened

Legacy sync downloads every block body from a single peer. `fetchHeaderBlocks` opened with
`sp := sm.loadSyncPeer()`, built one `getdata` and sent it to that one peer, and across the whole
service there were only two places a block `getdata` was ever constructed. So however many peers the
node was connected to, one TCP connection was the hard ceiling on how fast it could catch up — and
whenever that peer was judged slow, the node disconnected it and threw away every header it had
downloaded since the last checkpoint.

SV Node has no sync peer for block bodies at all. Every eligible peer runs `FindNextBlocksToDownload`
each pass and claims its own slice against a shared in-flight ledger, up to 16 blocks each inside a
1024-block window. With eight outbound peers that is up to 128 blocks moving at once.

This PR closes that gap. It also has to fix the things that made multi-peer unsafe, and one of those
turned out to be a live data race on `main` — so the diff is larger than the headline suggests, and the
commits are ordered so each can be read on its own.

It then grew a second half, because the first half was measured on mainnet and the holding area it
introduced turned out to be where the time went. Spreading downloads over several peers makes
out-of-order arrival the normal case, so the park is not a corner of this feature, it is on the hot
path. On the Hetzner mainnet box the node sat at one height for **nineteen minutes** with 58 blocks
already downloaded and waiting, having committed nothing: a goroutine profile caught the one goroutine
that commits blocks in order sitting inside the park's disk write, and in that window it took on 64 park
admissions and 39 deletions instead of committing anything. So the second half takes the park's work off
that goroutine and replaces the rule that was throwing downloaded blocks away.

## The change, in the order the commits land

**A single owner for the header list.** `blockHandler` dispatches `go sm.handleHeadersMsg(msg)` — a new
goroutine per headers message — and that function mutates `sm.headerList` in five places while the
block-queue consumer removes its front on a different goroutine. `container/list` is not goroutine-safe
and there was no mutex for it. Two headers messages arriving together already raced each other. This is
a pre-existing bug, not one multi-peer introduces, and everything else here is built on that list, so it
goes first. The lock is dropped and re-taken around `GetBestBlockHeader`, which can block for minutes
during sync, because holding it there would wedge the one goroutine that commits blocks in order.

**A hash index over the header list**, so a header can be found without walking.

**One in-flight ledger.** There were two request maps that disagreed: a global one and a per-peer one,
[...]

PR #1632, opened 2026-08-19
feat(legacy): probe unused addresses so the node has good ones to dial when a peer drops

### LINKED ISSUE #1607: legacy: add feeler connections to keep the address book healthy
Teranode's legacy peer layer has no equivalent of svnode's feeler connections, and nothing else in the node performs the job they do.

## What svnode does

Once the outbound tier is at target, `ThreadOpenConnections` reserves a small number of extra slots (`nMaxFeeler`) and makes a short-lived probe roughly every two minutes (`FEELER_INTERVAL = 120`, `src/net/net.h:88`). The probe connects, confirms the address answers, and lets the address manager promote it from the "new" table into "tried". The slots are budgeted explicitly alongside the outbound ones:

```cpp
semOutbound = std::make_shared<CSemaphore>(std::min((nMaxOutbound + nMaxFeeler), nMaxConnections));
```

and inbound capacity is reduced to match (`src/net/net.cpp:1261`), so feelers are paid for rather than borrowed.

The design goal is stated in the source: *"Increase the number of connectable addresses in the tried table."*

## Why it matters here

Teranode dials automatic outbound peers from `newAddressFunc`, which draws from the address manager and skips any candidate whose netgroup is already represented. Nothing ever verifies an address that is not currently being dialled for real. So the tried table only improves as a side effect of connections the node happened to want anyway, and it degrades whenever addresses go stale.

That matters most during a long initial block download, which is exactly where the peer-supply problems in #1601 showed up: a node that churns peers draws repeatedly from a book whose quality nothing is maintaining. Every failed dial costs an address-manager penalty, and there is no counterweight promoting good addresses back up.

PR #1601 fixed the accounting bugs that lost outbound slots — leaked pending entries, over-dialling past the target, named peers consuming automatic slots and netgroups. Those were all about spending the budget correctly. Feelers are the other half: keeping the pool of addresses worth spending it on healthy. It was deliberately left out of that PR because it is new functionality rather than a correctness fix.

## Rough shape of the work

- A feeler budget setting, defaulting the way svnode does, with the slots reserved from the peer cap rather than taken out of the automatic tier.
- A probe path in the connection manager that dials, confirms the handshake, and disconnects, without registering the peer as an ordinary outbound peer or claiming a netgroup — feelers must not consume automatic capacity or diversity.
- Address-manager integration for the probe — likely none needed. See the verified findings below.
- Interval pacing similar to svnode's Poisson-spaced ~120s, and only while the automatic tier is already at target, so probing never competes with real connection supply.

## Verified: how the address manager is wired today

Checked rather than assumed, since the design above depends on it. Teranode integrates with `addrmgr` at exactly three points:

| Call | Site | When |
|---|---|---|
| `Attempt()` | `peer_server.go:3015`, `outboundPeerConnected` | after a TCP connection is established |
| `Good()` | `peer_server.go:680`, `OnVersion` | outbound peers only, after the version exchange and the BSV user-agent check |
| `Connected()` | `peer_server.go:2475`, `handleDonePeerMsg` | peers that completed the handshake |

**The good news: promotion works, and a feeler would get it for free.** `Good()` is the new-to-tried trigger (`addrmanager.go:1046`) and it already fires on a successful outbound version exchange. A probe that connects and handshakes through the ordinary outbound path would promote its address with no new address-manager code at all. That removes most of the integration work this issue originally anticipated.

**The other half is now fixed.** When this issue was opened, failed dials were never recorded: `Attempt()` ran only after a TCP connection had already succeeded, so `attempts` stayed at zero, `chance()` never reduced a dead address's weight, and `isBad()` could never condemn it. That has since been folded into PR #1601, following svnode's guards — a failure counts only when the node is demonstrably connected elsewhere, and at most once per address between successes.

So the node can now learn that an address is dead. Feelers are how it learns that a dormant one is alive again, which is the piece still missing: nothing verifies an address the node is not already dialling for real, so a working address that has gone quiet is never promoted back into the tried table.

## References

- `src/net/net.cpp:1771` `ThreadOpenConnections`, feeler branch and `FEELER_INTERVAL` pacing
- `src/net/net.cpp:1261` inbound capacity accounting
- `src/net/net.cpp:2547` semaphore sizing
- Comparison notes written while reviewing #1601, covering every other difference found between the two implementations

freemans13:
> **#1601 has merged**, so this branch is no longer stacked and the diff has
> collapsed to just the feeler work: thirteen files, all of them listed below.

Closes #1607.

## What happens today

The node keeps two lists of peer addresses: ones it has merely been told about, and ones it has actually connected to and exchanged greetings with. Only the second list is worth much when a peer drops and a replacement is needed in a hurry.

An address only ever moves from the first list to the second as a side effect of a connection the node wanted anyway. Nothing ever checks an address the node is not already using.

## Why that is a problem

Hosts go away, and the address book has no way of finding out. The list of addresses known to answer therefore only shrinks, while the list of unchecked ones grows. When a peer is lost — which during a long initial block download happens constantly — the replacement is drawn from a pool nothing maintains, and the node can spend a long time dialling hosts that stopped answering months ago. That is where the peer-supply trouble in #1601 showed up.

#1601 was about spending the connection budget correctly, and part of that was teaching the address book that addresses die. This is the other half: teaching it that a dormant one is alive again.

## What this adds

A feeler, copied in behaviour from SV Node. Roughly once every two minutes, and only while the node already has all the real peers it is aiming for, it picks one address it has never confirmed, connects, waits for the other end to say who it is, notes that the address works, and hangs up. SV Node states the goal in one line of its own source: increase the number of connectable addresses in the tried table.

The probe is paid for rather than borrowed. SV Node carves its feeler's allowance out of the room it keeps for incoming peers and never out of its outgoing target, so probing can never cost it a peer it chose to dial. Teranode has a single combined ceiling over incoming and automatic outgoing peers rather than SV Node's two, so the faithful translation is to lower that one ceiling by the feeler allowance and leave the outgoing target completely alone. With the shipped defaults — a cap of 20 peers and a target of 8 outgoing — the one reserved slot comes entirely out of the incoming share, taking it from 12 to 11, which is exactly SV Node's shape. (The cap is 20, not bsvd's compiled-in 125: `settings.conf` sets `legacy_config_MaxPeers = 20` and the reflection loader in `config.go` applies it to that field on every real run. An earlier version of this description said 125, and two tests asserted against it.) The runtime `addnode` path gets the same figure, so the two doors cannot disagree about the cap.

**The probe claims nothing.** It never registers as an ordinary peer and never goes near the connection manager, and both matter. Being in the outgoing peer list *is* the claim on a network segment and *is* the peer count, so a probe registered as a peer would take a slot and a segment away from a real one. And the connection manager's whole job is keeping connections alive: handed a probe it would hold the slot for as long as the probe lived and then dial a replacement the moment it hung up. SV Node can afford to route its feeler through the ordinary path because its probes live well under a second; ours is asked to claim nothing at all, so it dials directly and drives a bare connection.

**It draws only from the unchecked list.** This is the one place the issue's "no address-manager changes needed" was too optimistic. It is right that promotion comes free — the same call the ordinary path already makes does it. But the existing selector tosses a coin between the two lists, and for a probe whose entire purpose is to move an address into the second one, half those draws would be wasted. SV Node restricts its feeler's draw the same way. The first commit adds that restricted draw and leaves the existing selector's behaviour untouched.

**It checks who answered before believing them.** The ordinary peer path rejects non-BSV clients, and the probe applies the same test, now shared between them. This is not a nicety: marking an address as good can evict an existing good one, so promoting a BTC or BCH node that happened to answer would push out a real BSV peer and degrade the very pool the probe exists to improve.

**It records failures too.** The failure bookkeeping #1601 added is wired into the connection manager's dial, which a direct dial bypasses, so the probe records its own. Without that, the feeler would only ever teach the book good news.

**The gap between probes is randomised** around the configured average rather than fixed. A fixed two-minute period is a fingerprint: it identifies the node across address changes and predicts the next probe, and it would put a fleet started together in lockstep. SV Node randomises for the same reason.

Three settings, all live end to end. `legacy_maxFeelerPeers` (default 1, matching SV Node's single feeler) is the allowance and the single off switch — zero stops the probing and releases the reserved slot together. `legacy_feelerInterval` (default 120s, SV Node's figure) is the pacing; a non-positive value falls back to the default with a warning rather than acting as a second, disagreeing off switch, because "disabled but still paying for a slot" is exactly the sort of half-state this series has shipped before. `legacy_feelerHandshakeTimeout` (default 25s) is the probe's own deadline, and has to stay inside the peer package's 30-second negotiate timeout for the reason set out in round three below.

Three places where I deliberately differ from SV Node, each explained at the point in the code where the difference lives: an address in a segment we already occupy is skipped rather than costing the whole two-minute slot; banned addresses are filtered when picking rather than discovered at dial time; and there is no filter on advertised services, because Teranode's ordinary dial path has none and a probe stricter than the thing it feeds would verify addresses the node then declines to use.

## Testing

Thirty-one tests. The important one drives the real dial path against a real listener and watches an address move between the two lists through public API alone — the unchecked list empties while the total count is unchanged, which is precisely what promotion is. Its counterpart answers with a non-BSV greeting and shows the address stays put but is charged an attempt.

Every test was then checked by breaking the thing it claims to cover. That found four production lines I could delete with the suite staying green — the user-agent check, the restricted draw, the failure recording, and the at-target condition where it is actually applied rather than where it is defined — plus one test that was asserting a moment too early to mean anything. All five are fixed, in their own commits, and the re-run confirms each mutation now fails.

**One line is not covered and I am not going to pretend otherwise:** the single call that starts the feeler, inside `peerHandler`. Deleting it leaves all 885 tests green. `peerHandler` starts the sync manager, which needs a blockchain client, a validator, a UTXO store, a subtree store and three validation clients, and this package has no fixture for that. The mitigations are that everything inside the started function is covered end to end, the launch is one line next to the connection manager's, and each probe logs a single line at info level so the wiring is checkable on a running node in one command. I will paste that output from teratestnet with a short interval before this leaves draft.

Also stated openly:

Probe goroutines are not waited on at shutdown. The dial has a thirty-second timeout and no cancellation, and holding shutdown for thirty seconds on a dead address is worse than the alternative. A probe checks for shutdown after its dial and stops without touching the address book, so the worst case is at most one lingering goroutine per reserved slot, each ending on its own within thirty seconds.

A probe and a real dial can still race to the same host. The connection manager dedups only against its own books and cannot see a probe; the probe reads a snapshot of registered peers and cannot see a dial in flight. The window is now much narrower than it was: the probe re-queries the peer set after its dial and gives the host up if a peer has taken it, which is what SV Node does at the same point. What is left is a dial already in flight, which is in neither book. Closing that properly needs a reservation shared across two subsystems, which is not worth it for a probe. An earlier version of this description put the cost at "one wasted probe"; that understated it, because two connections to one host risks losing the first, and the first is the real peer.

The reservation bounds registered peers, not raw sockets. Teranode applies its peer cap after the handshake rather than at accept, so the live socket count can briefly exceed the cap regardless of anything here. SV Node's semaphore is a harder guarantee than ours; this reproduces the accounting property within the accounting Teranode actually has.

A probe's own peer logs at debug, never at warning. That is not just about tidiness: the mainnet soak showed that probing a fork client reliably produced an error plus a `Disconnecting ... reason: malformed message` warning — the exact line the disconnect-rate measurements count — because the BSV parser meets a command it does not know and tears the connection down before the feeler's deliberate quiet hang-up can. The log-once guard cannot arbitrate that; whichever teardown wins owns the log. So the probe peer is given a logger that moves its warnings and errors down to debug. An earlier version of this section described the problem as "a rare cosmetic wart" affecting a microsecond-wide race, which was wrong in both frequency and volume.

## Note for the owner of #1601

Five lines of yours are touched, none of them reverted or changed in behaviour. With the feeler allowance at zero the tree behaves exactly as it does today. They are: the peer-cap comparand in `handleAddPeerMsg` and the matching one in the runtime `addnode` path, both gaining a term that is zero unless feelers are budgeted; the user-agent condition in `OnVersion`, extracted to a named function and otherwise identical; the new start line in `peerHandler`; and a read-only accessor on the connection manager. The comment #1601 left at the peer-cap check already names the feeler budget as part of that arithmetic, which is what this fills in.

---

## Review findings, fixed before this was opened

Three adversarial reviews ran against this branch and all three said do not ship.
Each finding was demonstrated rather than argued, and all three are fixed in the
final commit.

1. **The reservation could starve the outbound tier.** The budget only checked
   that reserving left room for one ordinary peer, when it has to leave room for
   the whole automatic tier. Turning `MaxPeers` down or the target up left the
   node permanently below target - dialling, being refused by its own door, and
   dialling again. It now gives up the probe instead, with a warning naming the
   numbers. Giving up the probe is the right way to lose that argument.

2. **The slot cap was untested.** Deleting the token acquisition left the whole
   package green, and that cap is the entire mechanism behind "reserved rather
   than borrowed". The test written for it did not actually close the gap - see
   the second review round below, which caught that and fixed it properly.

3. **One disconnect was logged twice**, on 19.5% of *successful* probes rather
   than the tight race described above. The log ran before the guard that makes
   teardown idempotent, so a probe hanging up on purpose at debug level was still
   reported at warning by its own read loop. Fixed by moving the log inside the
   guard - a general improvement to every peer teardown in the legacy layer, and
   it should make the disconnect counting in #1601 more accurate rather than
   merely quieter.

---

## Second review round

A further review ran after the above, and found four things. All are fixed on
the branch; every fix was then checked by breaking it and watching the test go
red.

1. **The slot cap was still untested, despite the claim above.** Replacing the
   whole token acquisition in `feelerHandler` with a bare `go s.feelerProbe()` -
   removing the cap entirely - left the package green. The test written for it
   built its own channel and re-implemented the select by hand, so it was
   exercising Go's channel semantics rather than the production line and could
   not see that line disappear. `TestFeelerHandlerWaitsForASlotToken` now drives
   the real loop with its only slot already spoken for, shows no probe runs, then
   returns the token and shows one does. The second half is what makes the first
   half mean anything: a loop that had simply died would otherwise pass.

2. **The starvation guard was judged against the wrong number.** `feelerBudget`
   was handed the outbound target computed from configuration, before
   `connmgr.New` had a say - and `New` substitutes its own default of eight for a
   configured zero. The guard therefore compared against zero in exactly the case
   it was written for, since no reservation can look like it starves a tier
   aiming for nothing. With `targetoutboundpeers` unset and `maxpeers` at eight
   the node reserved a slot anyway, dropped its admission ceiling to seven, and
   sat below target for ever, dialling for an eighth peer its own door refused.
   The budget is now set after the manager exists, from its effective target,
   which is what that accessor's own documentation asks callers to do and what
   `feelerAllowed` already did.

3. **Two claims in comments were wrong, and five svnode citations pointed at the
   wrong lines.** The citations were checked against a local v1.1.0 checkout: the
   connect loop's 500ms sleep is `net.cpp:1802`, not 1932; the pre-dial jitter
   1934, not 1931; the service filter 1902, not 1898; the ban check 2113, not
   2115. The connect-only justification claimed the probe's address source is not
   installed in that mode, which is false - `addrmgr.New` is called
   unconditionally; what connect-only omits is `newAddressFunc`. And
   `peerAdmissionCeiling` claimed named peers are not bounded by the ceiling:
   they are not *counted* against it, but `handleAddPeerMsg` still applies it to
   them, so a full node turns a named peer away and the reservation makes that
   bite one peer sooner.

4. **Two smaller edges.** The probe writes to the address book on both arms of
   the dial, but only the success arm checked for shutdown; one check now covers
   both. And `poissonNext` converted a float to a `Duration` without checking it
   fits - harmless at the two-minute default, but the mean is settable and
   `ExpFloat64` reaches about 745, so a mean beyond roughly 140 days overflows.
   Both architectures were measured rather than assumed, and they fail in
   opposite directions: amd64 returns `MinInt64`, a negative gap that probes on
   every tick, while arm64 saturates to `MaxInt64` and stops the feeler for the
   life of the process. That is also why the guard is tested as its own function
   rather than through `poissonNext` - on arm64 an unguarded saturation and the
   guarded fallback are the same value, so the bug is invisible from there.

One finding was raised and **not** acted on. `poissonNext` returns exactly zero
for a very small positive interval, which is true and measurable, but a
nanosecond mean means "probe as fast as you can", the one-second poll already
caps the rate, and re-rolling from a zero deadline is harmless. The one-second
floor that poll puts under any configured interval is now written into
`legacy_feelerInterval`'s own documentation instead.

---

## Third review round

A third review ran after the above and found that a claim the second round had
accepted was false. Fixing it turned out to be a behaviour fix rather than a
comment fix, so this round changes what the node does.

1. **Every silent host was being counted as a lost peer.** The comment on the
   feeler's handshake deadline said the peer package's own negotiate timeout
   "does not cover this case: it starts once negotiation begins". It does cover
   it. `Peer.start` races the whole negotiation against
   `time.After(negotiateTimeout)`, so it bounds total elapsed handshake time,
   which is exactly the failure the feeler's own deadline was written for - a
   host that accepts a TCP connection and then says nothing at all. Both
   deadlines sat at thirty seconds, so they tied, the peer package won, and the
   probe's deliberately quiet hang-up was quiet only when the far side actually
   answered. Every mute address the feeler touched was reported at warning level
   as a lost peer, on the very measurement this series exists to clean up.

   The deadline is now a third setting, `legacy_feelerHandshakeTimeout`,
   defaulting to twenty-five seconds so the feeler's timer settles the teardown
   with five seconds in hand. A non-positive value falls back to the default and
   anything at or beyond the peer timeout is brought back under it, both with a
   warning, because a value that reintroduces the warning hang-up is not one
   anyone means to choose. `negotiateTimeout` is exported as
   `peer.NegotiateTimeout` so the guard reads the real number instead of a copy
   that can drift, and its own comment now says elapsed rather than inactivity,
   which is what it has always measured.

2. **A non-BSV host is now banned, not just hung up on.** This is a policy
   change and the widest thing in the PR, so it is worth stating plainly rather
   than leaving in the diff. Without it the address stays in the new table and is
   drawn again, so the probe spends its whole allowance rediscovering the same
   BTC and BCH nodes. It mirrors what `OnVersion` already does to such a peer,
   and svnode's own `-banclientua` handling. But the blast radius is wider than
   the feeler: the ban list is a process-wide singleton shared with the p2p
   service and notifies its subscribers, so a probe now bans a host node-wide
   for twenty-four hours and triggers disconnect sweeps in both layers. The
   `-nobanning` switch turns it off, and that path is tested.

3. **All three new mechanisms shipped unpinned, and the second hop was dead to
   tests.** Deleting the peer-timeout guard, deleting the non-positive fallback,
   and severing the link between the loaded setting and the probe's timer each
   left the package green. The last is the one that matters: a setting that loads
   but never reaches the code is dead, and nothing was watching for that. All
   three now go red when broken, the second hop through a listener that accepts
   TCP, reads the version and then stays silent.

4. **Three comments had been shortened past the point of usefulness** and are
   restored. The worst loss was the measured per-architecture overflow behaviour
   in `boundedDuration` - amd64 yields `MinInt64` and probes on every tick, arm64
   saturates and stops the feeler for the life of the process. That is a
   measurement, not something a reader can recover from the code, and it is the
   entire reason the guard is tested as its own function rather than through
   `poissonNext`.

Also corrected: both the comment and the setting's documentation claimed
`AssociateConnection` starts the peer's timer *first*. It starts it in a
goroutine, microseconds either side of the probe building its own, so neither
text can promise an order. What is true is that a disconnect is only ever
reported once, so whichever timer fires first owns the teardown - and the
five-second margin is what puts the answer beyond doubt.

Still open, neither blocking: the quiet-teardown behaviour is now three pieces -
the log-once guard, the twenty-five second default, and the two edge guards -
each pinned alone but never together, so one test still owes the whole path. And
`newFeelerTestServer` builds its ban list with a nil database, which `Add`
panics on rather than erroring; harmless today because the one test that bans
overrides it, impossible in production because `GetBanList` fails closed, but a
landmine for the next person writing a feeler test.

---

## Fourth review round

A fourth review ran against the merged head: a correctness and concurrency pass,
a mutation pass over the whole new test suite, and a claim-and-citation pass
against the local SV Node v1.1.0 checkout. It also found an unsubmitted draft
review sitting on this PR from an earlier session, whose findings are folded in
below. Nothing severe turned up in the feature itself; what did turn up was one
real coverage hole, one real race, three gaps the draft review had already
spotted, and a cluster of claims that had drifted.

1. **The three shutdown guards were completely untested.** Neutralising all three
   at once left the whole `services/legacy` package green in twelve seconds, and
   `feeler_test.go` never mentioned shutdown at all. That is the guard whose
   justification runs to three paragraphs, including round two's own fix
   extending it to both arms of the dial.

   The guard after the dial is now tested through a real probe on both arms. The
   other two are only reachable when shutdown begins *mid-probe*, which no test
   can land on purpose, so they were given seams: the attempt guard is
   `attemptIfRunning` and the promotion guard lives in `judgeVersion`. Both are
   tested directly, and each is paired with its running counterpart, because
   "nothing was written" is only evidence if something is written when the node is
   up.

2. **The probe could throw away a version it already had, on a coin toss.** A
   version can land in the same instant the far side hangs up or the deadline
   expires, and a `select` picks uniformly among cases already ready. The
   hang-up and timeout arms each decided their own outcome, so either could
   discard a version in hand. The missed promotion is a nuisance; the missed
   **ban** is the problem, because the address stays in the new table and the
   feeler spends its whole allowance rediscovering the same BTC and BCH nodes —
   the exact loop round three added the ban to close.

   Measured, with a delay widening the window the probe reaches the select
   through: 23 of 40 probes of a BSV host failed to promote it, and 20 of 40
   probes of a BTC host failed to ban it. A clean coin toss. With the fix, 0 of 40
   on both. Each arm now records only why its own wait ended, and the verdict is
   reached once downstream, so no arm can be written that skips it.

3. **The host filter was blind to every Tor peer.** `connectedHosts` keys on the
   host part of a peer's dial string; the filter keyed on `na.IP.String()`. Those
   agree for IPv4 and IPv6 literals, so nothing showed — but for an OnionCat
   address they never do, because the dial string carries the `.onion` name and
   the raw IP the `fd87:d87e:eb43:` form. So "never a host we are already talking
   to" did not apply to Tor peers at all. Both halves now share one predicate.

4. **The occupancy snapshot was stale by the time the socket was up**, and SV Node
   re-checks at exactly that point for exactly that reason (`FindNode` in
   `OpenNetworkConnection`, `net.cpp:2118`). See the corrected disclosure above.

5. **The probe promoted on user agent alone**, so a BSV node announcing a protocol
   below `MinAcceptableProtocolVersion` was verified into the tried table — where
   promotion can evict an existing entry — even though the ordinary outbound path
   refuses to dial it. The check has to live in the probe: the peer package
   applies the same rule, but only *after* invoking the `OnVersion` listener the
   probe promotes from, so `res.done` is already closed by the time it rejects.
   Not banned, unlike a non-BSV user agent — a BSV node on an old protocol is
   unusable, not hostile.

6. **Six statements were untrue and are corrected.** `setFeelerBudget` claimed
   `handleAddPeerMsg` and `handleQuery` are the *only* readers of `feelerSlots`;
   there are four, and the two omitted are in `feeler.go` itself. The SV Node
   jitter citation claimed its 0-1s pre-dial sleep exists "purely to break up the
   half-second granularity of its own connect loop", where SV Node's own comment
   at `net.cpp:1932` says "to avoid synchronization" — the anti-lockstep reason
   this description already gives for the randomised gap; and the justification
   for omitting it was backwards, because our firing is quantised to whole seconds
   by the poll tick and so is *coarser* than SV Node's 500ms loop. The
   `-banclientua` citation pointed at the guard at `net_processing.cpp:1722`
   rather than the `Misbehaving` call at `:1725`. `peerAdmissionCeiling` said the
   two doors' disagreement over permanent peers is tracked by the TODO at the
   check; that TODO asks a different question, and nothing tracks the
   disagreement. `feelerProbe` said it "never goes near the connection manager",
   where it never *registers* with it but does read it. And a comment claimed SV
   Node can route its feeler through the ordinary path because "its probes live
   well under a second" — they use the ordinary handshake and can sit for sixty
   seconds, contradicting this file's own citation of that constant twenty lines
   earlier.

7. **The three settings were documented nowhere outside their struct tags.**
   `legacy_settings.md` gains the rows and a section, and `legacy_maxFeelerPeers`
   now documents the two conditions besides a configured zero that switch feelers
   off — connect-only mode, and a cap too tight to hold the reservation. An
   operator who tightened the peer cap and found feelers off could not previously
   learn why from the setting's own documentation.

Also fixed: `newFeelerTestServer` built its ban list with a nil database, which
`Add` panics on rather than erroring. Production was never exposed, because
`GetBanList` fails closed, but it left any test whose probe met a non-BSV user
agent one edited string literal away from a panicking binary instead of a failing
assertion. It now builds a writable one.

Every fix in this round was checked by breaking it and watching the test go red,
and each was committed before the mutation was applied.

**One finding was raised and rejected.** Copilot reported that
`legacy_feeler_settings_test.go` misuses `require.Equal` with format specifiers
because "testify does not format the message for Equal". It does:
`messageFromMsgAndArgs` in testify v1.11.1 runs
`fmt.Sprintf(msgAndArgs[0].(string), msgAndArgs[1:]...)` whenever more than one
argument is passed. `Equalf` is the explicit variant, not the only working one.
The godoc above that test *was* wrong in a different way, and is fixed.

**Still open, neither blocking.** The quiet-teardown behaviour is three pieces —
the log-once guard, the twenty-five second default and the two edge guards — each
pinned alone but never together, so one test still owes the whole path. Probe
goroutines are still unwaited at shutdown; the correctness pass confirmed this
cannot reach a use-after-close, because `AddrManager.Stop` sets no closed flag,
`BanList.Stop` never closes its database handle and `ConnManager.Stop` leaves its
config intact, so the worst case really is a lingering goroutine and a lost write.
And with `legacy_maxFeelerPeers > 1` two concurrent probes can still select the
same address, since nothing marks a candidate as claimed by a probe in flight;
the cost is a wasted slot, and it is dormant at the shipped default of one.

Not caused by this branch, but currently red: CI's `test` job fails on
`services/blockassembly TestGetBlockAssemblyBlockCandidate/with_10_txs`, which
expects a coinbase of `5000000045` and gets `5000000000` — the candidate came back
with no fees and no transactions after the test had already watched `TxCount`
reach 11. It passed three of three locally and this branch touches no file under
`blockassembly`. SonarQube's failure is downstream of it.

---

## Third review round

An approving human review raised three non-blocking findings. All three were real,
all three are fixed, and each fix was checked by reverting it and watching the
tests go red.

1. **An address the probe could not resolve spent a whole two-minute slot in
   silence.** The probe resolved after selection and gave up on failure, before it
   had counted an attempt or written anything, so the entry came out of it
   unchanged and equally likely to be drawn next time. OnionCat is a class that
   always takes that path: the address book admits it on purpose, and this layer
   has no onion dial path at all. Selection now resolves and skips what it cannot
   dial, so the pass keeps looking. The resolve goes ahead of the ban check,
   because `BanList.IsBanned` logs at error when handed something that is not an
   IP, and a pass that kept looking would otherwise have turned one silent slot
   burn into up to a hundred error lines.

2. **The selection pass read address-book entries with no lock held.**
   `UnverifiedAddress` released the manager's mutex before returning and the
   caller then read the entry, racing `Attempt` and `Good`. It now returns a
   snapshot assembled inside the lock rather than the entry itself, so there is no
   handle left to race. A new test drives 200 selection passes against a writer
   and failed under `-race` before the fix.

3. **A startup misconfiguration was reported as permanent runtime noise.** The
   handshake-deadline guards were re-run on every probe, so a bad value warned
   every two minutes for the life of the process and broke the one-info-line-per-
   probe property this description leans on. It is settled once in `startFeeler`
   now, and the deadline it settled on is on the startup line.


services/legacy/feeler.go
```diff
@@ -0,0 +1,635 @@
...
+		// that an address is dead. recordFailedDial is wired into the connection
+		// manager's own Dial closure, which a direct dial bypasses, so without
+		// this call the probe would only ever teach the book good news.
+		s.recordFailedDial(netAddr)
+		s.logger.Debugf("[Feeler] Dial %s failed: %v", addrString, err)
+
+		return
+	}
+
+	if s.banList.IsBanned(conn.RemoteAddr().String()) {
```
freemans13: **[bug]** After TCP connect the probe only re-checks the ban list, not whether the host (or its netgroup) is now held by a real peer. SV Node does that check in `OpenNetworkConnection` (`net.cpp:2118`, `FindNode`) because the selection-time `setConnected` snapshot is stale by the time the socket is up. Teranode's snapshot is taken once at the start of `feelerCandidate` and never refreshed; worse, an in-flight feeler is deliberately not in `outboundPeers`, so `newAddressFunc` cannot see it either. The race is concrete during IBD: feeler is allowed only at target, so a probe starts; an outbound then drops; `AutomaticOutboundCount` is established-only so replenishment starts immediately; `GetAddress` can still draw the same new-table address the probe is sitting on (Good has not run yet). Two connections to one host is the failure mode the host filter's own comment calls "a good way to lose the first one", and unlike SV Node the replacement outbound is the connection that can be dropped. The 25s handshake timeout makes the window large.

**Suggestion:** After the successful dial (alongside the existing ban check), re-query `feelerPeerSnapshot` and abort if the host or netgroup is now occupied. Also publish in-flight feeler targets to the ordinary dial path (or skip `GetAddress` results that match a live probe) so connmgr cannot pick the same new-table address while the probe still holds the TCP connection.

services/legacy/feeler.go
```diff
@@ -0,0 +1,635 @@
...
+			outcome = "abandoned, shutting down"
+		case !isBSVUserAgent(res.userAgent()):
+			outcome = "answered but is not a BSV node"
+			s.banNonBSVHost(addrString)
+		default:
+			// Promotes the address from new to tried, which is the entire point
+			// of the exercise. svnode's feeler clears the same bar at the same
+			// moment, in ProcessVersionMessage rather than on verack, and so
+			// does teranode's own outbound path in OnVersion.
+			s.addrManager.Good(na)
```
freemans13: **[suggestion]** Promotion keys only on `isBSVUserAgent`. SV Node's feeler rides `ProcessVersionMessage`, which refuses obsolete protocol versions and unexpected services before `MarkAddressGood` (`net_processing.cpp:1699` and `1684`, then `1851`). Teranode's ordinary `serverPeer.OnVersion` also skips `Good` when `msg.ProtocolVersion < MinAcceptableProtocolVersion`. The feeler `OnVersion` listener (line 540) captures the UA and closes `done` with neither check, so a BSV UA on an unusable protocol still moves the address into tried — and `Good` can evict an existing tried entry. `MinAcceptableProtocolVersion` is `wire.MultipleAddressVersion` (ancient), so this is unlikely in the wild, but it is a tried-table degradation the ordinary path already refuses and that the PR's "check who answered before promoting" claim does not currently pin.

**Suggestion:** In the `<-res.done` success path, refuse promotion unless the version meets `peer.MinAcceptableProtocolVersion` (pass the protocol version through `feelerResult` next to the UA). Do not `Good` a peer the ordinary outbound path would drop.

services/legacy/feeler.go
```diff
@@ -0,0 +1,635 @@
...
+		if s.banList.IsBanned(addrmgr.NetAddressKey(na)) {
+			continue
+		}
+
+		// Never a host the node is already talking to. A second connection to a
+		// peer we are mid-download from is a good way to lose the first one.
+		// The netgroup set below cannot stand in for this: it is derived from
+		// the automatic outbound list alone, so inbound and named peers are
+		// invisible to it.
+		if _, held := snap.hosts[na.IP.String()]; held {
```
freemans13: **[suggestion]** `connectedHosts` keys on `net.SplitHostPort(sp.Addr())` (the dial string's host, which for onion is the `.onion` name via `ipString`/`NetAddressKey`). The candidate filter looks up `na.IP.String()`, which for OnionCat is `fd87:d87e:eb43:…`. Those keys never meet, so the "never a host we are already talking to" rule does not apply to Tor peers even though addrmgr stores and groups them. IPv4/IPv6 literals happen to match; onion is the hole.

**Suggestion:** Key both maps the same way — `addrmgr.ipString(na)` / the host part of `NetAddressKey(na)` — so the filter is format-independent.

services/legacy/feeler.go
```diff
@@ -0,0 +1,635 @@
...
+// records what it learned, and hangs up.
+//
+// It never builds a serverPeer and never goes near the connection manager.
+// Membership of state.outboundPeers is the netgroup claim and the peer count,
+// so a probe registered as an ordinary peer would take both from a real one.
+// The connection manager's job is to keep connections alive: handed a probe,
+// it would count it against the outbound target and dial a replacement when
+// it hung up.
+//
+// svnode can afford to route its feeler through the ordinary path because its
```
freemans13: **[suggestion]** Production comments in this file are multi-paragraph design essays (why the reservation exists, why connmgr must not own the probe, PR/branch archaeology). Line 336's "svnode can afford to route its feeler through the ordinary path because its probes live well under a second" is also factually off: SV Node feelers use the ordinary handshake and can sit until `DEFAULT_P2P_HANDSHAKE_TIMEOUT_INTERVAL` (60s). They are often fast when the peer is up; they are not bounded under a second. The real reason not to go through connmgr is accounting (`TargetOutbound` / `AutomaticOutboundCount`), which the surrounding comments already state. Tests carry the same history leak (`feeler_test.go:314`, "the earlier sketch on stu/legacy-svnode-align").

**Suggestion:** Keep a short WHY (not through connmgr because it would count against the outbound target; handshake timeout must beat `peer.NegotiateTimeout`). Drop branch names, prior-sketch recaps, and the sub-second claim.

freemans13:
Round four is pushed (2f8b325e7). Replying to the three bot findings here rather than inline, because an unsubmitted draft review of my own is currently blocking review replies on this PR.

**On `settings/legacy_feeler_settings_test.go` — `require.Equal` with format specifiers.** Not a defect. `messageFromMsgAndArgs` in testify v1.11.1 runs `fmt.Sprintf(msgAndArgs[0].(string), msgAndArgs[1:]...)` whenever more than one argument is passed, so these assertions render exactly as intended; `Equalf` is the explicit variant, not the only working one. Left as is. The godoc above that test *was* wrong in a different way — it said "the two feeler settings" when the test covers three, and described a context gate that is not what the code does — and that is fixed in 9441a1b10.

**On the `[Feeler] Disabled` log message.** Fair, and fixed from the other end. The message no longer claims a cause, but it did not name one either. `legacy_maxFeelerPeers` now documents all three routes to a zero budget — a configured zero, connect-only mode, and a peer cap too tight to hold the reservation — and the last two already log their own reason at startup. An operator who tightens `legacy_config_MaxPeers` and finds feelers off can now learn why from the setting itself.

**On the `select` between `res.done` and `gone`.** Correct, and the consequence is worse than a missed promotion: the non-BSV check and the ban lived only on the `res.done` arm, so a BTC or BCH host that identifies itself and drops at once had a coin-flip chance of being recorded as "hung up before its version" — never banned, left in the new table, and drawn again. That is the loop the ban was added to close.

Measured with a delay widening the window the probe reaches the select through: 23 of 40 probes of a BSV host failed to promote it, and 20 of 40 probes of a BTC host failed to ban it. With the fix, 0 of 40 on both. Each arm now records only why its own wait ended and the verdict is reached once downstream in `judgeVersion`, which re-reads `res.done` rather than trusting it to have won — so no arm can be written that skips it. Thanks for that one.

Full round-four notes are in the description.

freemans13:
## Mainnet soak

Running on the Hetzner mainnet node, merged with #1606 so both soak together, at
the **shipped defaults** — one slot, 120s mean, 25s handshake deadline — rather
than the shortened interval an earlier draft of the description promised. Window
12:39:37Z to 14:32:48Z, 1h53m, **42 probes**.

First, the line no unit test covers:

```
[Feeler] Starting with 1 slot(s), mean interval 2m0s
```

That is `startFeeler` inside `peerHandler`, reached on a real node.

### What 42 probes found

| Outcome | Count |
|---|---|
| dial timed out (i/o timeout) | 25 |
| connection refused | 4 |
| no route to host | 3 |
| **verified** | **4** |
| answered but is not a BSV node (all banned) | 4 |
| handshake deadline expired | 1 |
| hung up before its version | 1 |

**32 of 42 addresses — 76% — were dead or unreachable.** That is the case for the
feature stated in numbers: three quarters of what sat in this node's unverified
table was rubbish, and nothing was ever going to find that out except a probe.

The four that verified were `/Bitcoin SV:1.2.2/`, `/Bitcoin SV:1.2.1/` (twice) and
`/Bitcoin SV:1.2.0/` — real addresses promoted into the tried table, so a peer lost
later is replaced from hosts that answered minutes ago.

Of the ten hosts that answered at all, **four were fork clients** — two
`/Bitcoin Cash Node:29.0.0(EB32.0)/`, one `(EB32.0; BCH-ATG)`, one
`/Bitcoin ABC:0.33.9(EB32.0)/`. All four banned. That is a 40% fork-client rate
among responsive hosts, and it is why the ban is not optional: without it the
probe redraws them for ever and spends its whole allowance rediscovering the same
four nodes.

Observed pacing is one probe per 161s against a configured mean of 120s. The
difference is the at-target gate correctly declining to probe on ticks where the
outbound tier had dipped below target, which during IBD it regularly does.

### Three things the soak found that review did not

**1. Most probes were invisible.** The first sample logged two probe lines whose
own counters read `attempted 8`. Every early return after the attempt counter
logged at debug, so a failed dial — the single most interesting outcome, and 76%
of this sample — left no trace. The description's claim that "each probe logs a
single line at info" was true only of probes that reached a handshake. Fixed: all
of them now report through one helper.

**2. Probing a fork client inflated the disconnect count.** The first BCH probe
produced:

```
ERROR | peer.go:2028 | Can't read message from 178.105.86.117:8333 ... unhandled command
WARN  | peer.go:2652 | Disconnecting (178.105.86.117:8333 ...) reason: malformed message
```

That WARN is the exact line the disconnect-rate measurements key on, so the feeler
was inflating the number this series exists to drive down, on every probe of the
clients it is built to find. The log-once guard could not help: the read loop met a
command the BSV parser does not know, tore down first and won the race, so the
guard suppressed the feeler's own quiet line and let the warning through.

The probe peer now gets a logger that moves its warnings and errors to debug.
Verified on all four fork clients in this window: **zero warnings and zero errors
between them**, only the two info lines each.

This replaces the description's "rare cosmetic wart" disclosure, which was wrong in
both frequency and volume — not one warning in a microsecond-wide race, but an
error plus a counted disconnect, reliably.

**3. The ordinary dial path has the same noise, and the ban reduces it.** One of
the banned hosts had been hit by the *ordinary* path four days earlier:

```
2026-08-17T20:53:58Z ERROR ... unhandled command [sendaddrv2]
2026-08-17T20:53:58Z WARN  ... reason: malformed message
```

So fork clients were already polluting the disconnect count through the normal
outbound path, independently of this PR. Banning them on discovery stops the
connection manager re-dialling them at all, which makes this a small net
improvement to that measurement rather than merely a neutral one. Not a claim I
had expected to be able to make.

### Also confirmed live

- A silent host was hung up by the feeler's own 25s deadline and reported at info,
  not at warning as a lost peer — round three's `legacy_feelerHandshakeTimeout`
  doing its job on a real host.
- Node health over the window: no ERROR, FATAL or panic lines; 162 warnings per
  20,000 lines, none from the feeler; 14 disconnects per 50,000 lines; 1.5 GB
  resident; sync advancing at roughly 5 blocks/second through height 131k+.

### Unrelated, but worth someone's attention

The pruner logs `phase 1b: failed to expire preservations: STORAGE_ERROR (69)` on
essentially every block on this node — first seen 2026-08-17, 182,106 times since,
so it long predates this branch and nothing here touches the pruner or the UTXO
store. Flagging it because the pruner is load-bearing.

services/legacy/feeler.go
```diff
@@ -0,0 +1,486 @@
...
+//
+// Separated from peerHandler so that everything except the single call line is
+// reachable from a test. peerHandler itself cannot be constructed in a unit
+// test: it starts the sync manager, which needs a blockchain client, a
+// validator, a UTXO store, a subtree store and three validation clients.
+func (s *server) startFeeler() {
+	if s.feelerSlots <= 0 {
+		s.logger.Infof("[Feeler] Disabled (legacy_maxFeelerPeers is not set above zero)")
+		return
+	}
```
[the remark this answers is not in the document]
freemans13: Fixed in 667a2f394. You and the Claude reviewer landed on the same thing from different sides: the docs promise the reason is logged, and `feelerBudget` only kept that promise for the tight-cap case.

`configured <= 0` and `connectOnly` shared one silent branch, so a node in either state produced only `startFeeler`'s bare `[Feeler] Disabled`, which reports that the loop did not start and nothing about why. The branch is now split and each case logs a line naming the setting that caused it, at info. The configured-zero test runs first so the deliberate rollback lever is what gets named when it is combined with connect-only.

`TestFeelerBudgetNamesTheReasonItRefused` asserts the operator-visible text for all three refusals plus that precedence, and that a granted reservation logs nothing about being disabled. Reverting the split fails three of its four subtests.

settings/legacy_feeler_settings_test.go
```diff
@@ -0,0 +1,61 @@
...
+		gocore.Config().Set(winBudget, "")
+		gocore.Config().Set(winInterval, "")
+	})
+
+	loaded := NewSettings()
+
+	require.Equal(t, 4, loaded.Legacy.MaxFeelerPeers,
+		"NewSettings must read %s under context %q", budgetKey, ctx)
+	require.Equal(t, 7*time.Second, loaded.Legacy.FeelerInterval,
+		"NewSettings must read %s under context %q", intervalKey, ctx)
```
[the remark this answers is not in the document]
freemans13: Not changing this one. testify does format the message when extra arguments are passed, so `require.Equal` is correct here.

`assert.messageFromMsgAndArgs` (testify v1.11.1, `assert/assertions.go:315`):

```go
if len(msgAndArgs) > 1 {
    return fmt.Sprintf(msgAndArgs[0].(string), msgAndArgs[1:]...)
}
```

The `Equalf` variants exist for compile-time vet checking of the format string, not because `Equal` leaves it unrendered. These three calls each pass two extra args, so they take that branch and render as intended.

freemans13:
Round five is pushed (667a2f394, on top of a merge of upstream/main).

**The one open finding is fixed.** `feelerBudget` now names its reason on every refusal, not just the tight-cap one. `configured <= 0` and `connectOnly` shared a silent branch, so the only line either produced was `startFeeler`'s bare `[Feeler] Disabled` — which reports that the loop did not start and nothing about why. Each has its own branch and its own info line now, naming the setting responsible. The configured-zero test comes first, so a node that is both in connect-only mode and has the lever pulled reports the lever.

That makes `legacy_settings.md:60` ("Each logs its reason at startup") and the `legacy_maxFeelerPeers` longdesc true as written, so the docs are unchanged.

`TestFeelerBudgetNamesTheReasonItRefused` covers all three refusals, the precedence between the first two, and that a granted reservation logs nothing about being disabled. Reverting the split fails three of its four subtests, which I checked rather than assumed.

Copilot's two inline findings are answered on their threads: the misleading-disabled-log one is the same finding, fixed by the same commit; the `require.Equal` vs `Equalf` one is rejected, with the testify source that decides it (`messageFromMsgAndArgs` formats whenever more than one arg is passed).

Copilot itself has refused the last several review requests with the quota-limit message, so it has not looked at anything since 2026-08-19.

freemans13:
Round six is pushed (1cfbc75eb). All 37 checks green on that head.

One more line, from re-reading my own fix rather than from a finding. `setFeelerBudget` had a fourth route to a zero budget: no connection manager, because the outbound target the reservation is judged against comes off the manager. It is unreachable on the real path, since `newServer` assigns `connManager` immediately above the only call, but it is the one way the ordering contract in that function's doc comment can be broken, and a silent zero there reads exactly like an operator having pulled `legacy_maxFeelerPeers`. It names itself now too, so every route to a zero budget says which route it was. `TestSetFeelerBudgetNamesAMissingConnectionManager` covers it.

On the three items flagged for a second look, no change and here is why each:

- **24h process-wide ban on a non-BSV host.** This is the ordinary path's rule, not a new one. `judgeVersion` tests the protocol version before the user agent, which is the order `serverPeer.OnVersion` uses, so a BSV node on a protocol we do not speak is dropped and never banned; only a non-BSV user agent is banned, exactly as an ordinary inbound or outbound connection to that host would be. Making the probe softer than the path it feeds would verify addresses the node then declines to use.
- **Probe goroutines unwaited at shutdown.** Bounded at one per reserved slot, so one at the shipped default, and each is capped by `legacy_feelerHandshakeTimeout` at 25s. Joining them would mean holding shutdown open for a host that is already not answering.
- **`quietProbeLogger` bypassed if the peer package starts deriving its own logger.** Correct as a future hazard, and worth remembering. It does not derive one today; the probe peer gets the injected value and nothing in `peer` wraps it.

Copilot has refused every request since 2026-08-19 with the quota-limit message, three more times today, so it has not reviewed anything on this branch since then.

SonarQube: quality gate passed, 3 new issues (advisory), 100% coverage on new code, 0% duplication. Benchmarks: 0 regressions, 0 improvements, 139 unchanged.

ctnguyen (approved):
Review at commit [1cfbc75](https://github.com/freemans13/teranode/commit/1cfbc75ebce0c71d5a12b9b218c4ba1aef90c568)

This is a genuinely good piece of work, and unusually easy to review because the code explains itself. The design decision that carries the most weight — that the probe claims nothing, drives a bare peer and calls the address book itself rather than going through the connection manager — is the right one, and the reasoning for it at `services/legacy/feeler.go:356-375` holds up against the accounting in `handleAddPeerMsg`. The reservation arithmetic checks out: `peerAdmissionCeiling` is applied at both doors with the same comparand, all four readers of `feelerSlots` named in `setFeelerBudget`'s doc comment are the only ones in the tree (`git grep -n feelerSlots` over tracked files), and at a budget of zero every one of them reduces to the pre-PR expression, so the "behaves exactly as today" claim is true rather than merely asserted. The `selectNew` extraction is faithful line for line — the refactored `GetAddress` can still only reach it when `nNew > 0`, which is the invariant the loop needs. I spot-checked the svnode claims against a local checkout of a later point release than the v1.1.0 the citations were taken from, so I can vouch for the substance rather than the line numbers: the `semOutbound` sizing as `min(nMaxOutbound + nMaxFeeler, nMaxConnections)`, the `nMaxInbound` subtraction, the at-target gate on the feeler block, the `break` on a netgroup collision, the "avoid synchronization" wording on the pre-dial jitter and both `net.h` constants are all as described. The in-tree citations (`peer.go:2774`/`:2783`, `peer_server.go:797`, the `handleAddPeerMsg` TODO) land on exactly the right lines. Test coverage is serious, and the habit of pairing a guard test with its running counterpart is the reason I believe the guards. Three things below, none of them blocking.

## Non-blocking issues

### ChiR1 — An address the probe cannot resolve burns a whole probe interval, silently

**Problem:** `feelerProbe` resolves its chosen address at `services/legacy/feeler.go:386` and gives up at `:388` if that fails — before `s.feelerAttempted.Add(1)` at `:392`, before any address-book write, and without an info line. By then `feelerHandler` has already re-rolled the deadline at `feeler.go:344`, so the slot is spent for the full interval. There is a reproducible class of address that always takes this path: OnionCat. `IsRoutable` deliberately admits OnionCat addresses (`services/legacy/addrmgr/network.go:230`), so gossip puts them in the new table; `NetAddressKey` renders them as `<base32>.onion:port` (`services/legacy/addrmgr/addrmanager.go:915-919`); and `bsvdLookup` refuses any `.onion` host unconditionally, whatever the proxy configuration (`services/legacy/config.go:793-796`). This layer has no onion dial path at all — `cfg.dial` is either `net.DialTimeout` or a SOCKS dialer (`config.go:745`, `:764`) and `bsvdDial` only ever receives an already-resolved `net.Addr` (`config.go:783`). Because no `Attempt` is recorded, `ka.lastattempt` never moves, `chance()` keeps its full weight, and the same entry stays eligible for the next draw at `feeler.go:816`.

**Why it matters:** I am claiming the mechanism, not a rate — I have no way from the tree to say what fraction of a real address book is OnionCat. But the failure is invisible in exactly the way this PR spent commits closing elsewhere: no info line, neither counter moves, and the outcome is a lost two minutes rather than a logged one. It also sits oddly beside `feelerHostKey` (`feeler.go:643-662`), which goes out of its way to key OnionCat peers by their `.onion` name so the occupancy filter matches — that care implies onion entries are expected to be live, and the round-three fix to the host filter was written on the same assumption. Either they can be probed, in which case the resolve needs a path, or they cannot, in which case they should not be reaching the probe at all.

**Fix:** Reject at selection so the pass keeps looking, next to the ban check in `feelerCandidate`:

```go
  // No onion dial path in this layer: bsvdLookup refuses .onion outright
  // (config.go:794), so such a candidate can only waste the slot.
  if addrmgr.IsOnionCatTor(na) {
      continue
  }
```

  The general form is to do the resolve inside the selection loop and skip any candidate that will not resolve, returning the `net.Addr` alongside `na`; that costs nothing for ordinary entries, since every other address in the book renders as an IP literal and takes the `net.ParseIP` short-circuit in `addrStringToNetAddr` (`services/legacy/peer_server.go:4328`).

### ChiR2 — The selection pass reads KnownAddress fields with no lock held

**Problem:** `UnverifiedAddress` releases `AddrManager.mtx` before returning (`services/legacy/addrmgr/addrmanager.go:993-994`), and `feelerCandidate` then reads the returned entry unlocked: `ka.NetAddress()` at `feeler.go:821` and `ka.LastAttempt()` at `feeler.go:844`. Neither accessor takes the mutex (`services/legacy/addrmgr/knownaddress.go:31`, `:49`), while the same fields are written under it from other goroutines — `Attempt` sets `ka.lastattempt` at `addrmanager.go:1081` and `Good` at `:1129`. The type says as much itself at `knownaddress.go:40-43`: "NOT safe for concurrent access ... the same is true of every other accessor on this type ... Read it from a goroutine that is not racing the address manager — in practice, tests."

**Why it matters:** the feeler is a new long-lived goroutine doing this up to `feelerCandidateTries` times per pass while peer read loops and the connection manager are writing those same entries. `time.Time` is a multi-word value, so a torn read makes `time.Since` return something arbitrary and the ten-minute freshness filter admits or rejects the wrong candidate. The blast radius is small — a wasted probe, not a corrupted book — but it is a real race and `-race` will report it once a test drives selection concurrently with address-book writes. In fairness this is a pattern the PR extends rather than invents: `newAddressFunc` does the same thing at `peer_server.go:4104`. It is worth fixing in the code this PR adds, since that is where the new accessor pair was introduced.

**Fix:** let the manager answer under its own lock rather than handing out an entry to be read later:

```go
  // UnverifiedAddress returns the candidate and the freshness the caller needs,
  // both read under a.mtx.
  func (a *AddrManager) UnverifiedAddress() (*KnownAddress, time.Time) {
      a.mtx.Lock()
      defer a.mtx.Unlock()

      if a.nNew == 0 {
          return nil, time.Time{}
      }

      ka := a.selectNew()

      return ka, ka.lastattempt
  }
```

  A locked `LastAttempt` accessor on `AddrManager` would do just as well; the point is that the read happens on the manager's side of the mutex.

### ChiR3 — The handshake-timeout guard re-validates, and re-warns, on every probe

**Problem:** `feelerProbe` calls `feelerHandshakeTimeout(s.logger, s.settings.Legacy.FeelerHandshakeTimeout)` inline when it builds its timer (`services/legacy/feeler.go:475`), and both guards inside that function log at warn (`feeler.go:205` and `:210`). The setting is fixed at startup, so on a node configured with a non-positive or too-large value those warnings repeat for the life of the process — about once every two minutes at the shipped pacing. The sibling setting is handled the other way round: `feelerHandler` resolves the interval and warns once before entering its loop (`feeler.go:320-324`).

**Why it matters:** a startup-time misconfiguration is reported as ongoing runtime noise, and it lands on the warn level this series is trying to keep clean — the same concern that justified giving the probe peer a `quietProbeLogger` (`feeler.go:535-545`). It also quietly breaks the "one info line per probe" property the description leans on for checking the feature on a running node, since a misconfigured node emits two or three lines per probe instead.

**Fix:** resolve it once alongside the interval and carry it, leaving `feelerHandshakeTimeout` a pure function so the table test at `feeler_test.go:1225` is unaffected:

```go
  // feelerHandler, beside the interval resolution
  handshake := feelerHandshakeTimeout(s.logger, s.settings.Legacy.FeelerHandshakeTimeout)
  ...
  case <-s.feelerTokens:
      go s.feelerProbe(handshake)
```

  Holding it on the server in `startFeeler` works equally well and keeps `feelerProbe`'s signature, which matters for the existing `runOneProbe` helper.

## Recap

| ChiR1 | Unresolvable onion candidate wastes probe | | 40% |
| ChiR2 | Unlocked KnownAddress reads during selection | | 30% |
| ChiR3 | Handshake timeout warning repeats forever | | 12% |

> *Required* is `YES` for blocking items that must be fixed before merge and is empty for
> non-blocking items. *Criticality* is a subjective 0%-100% hint to the author: 0% means not
> critical at all, and 100% means extremely critical. It is secondary to the
> blocking/non-blocking classification, which is decided on each issue's own merits rather
> than by a score cutoff. Items are listed in decreasing Criticality, and every blocking item
> is more critical than every non-blocking item.

freemans13:
All three fixed, one commit each, on `stu/legacy-feeler-connections` at 2131a8ee6. Every claim in the three findings held up against the tree, and two of them turned out to be worse than described once I chased them.

Thank you for the OnionCat chain in particular. Every link is there: `IsRoutable` exempts OnionCat from the RFC4193 rejection at `addrmgr/network.go:230`, `ipString` turns it into `<base32>.onion` at `addrmanager.go:915`, `bsvdLookup` refuses any `.onion` suffix before it ever reaches `cfg.lookup` at `config.go:794`, and `cfg.dial` is only ever `net.DialTimeout` or a SOCKS dialer handed an already-resolved `net.Addr`. I also grepped for `onionAddr` and it is constructed exactly once in the tree, in `peer_server_test.go:472`. This layer has no onion dial path at all, and I would not have found that from the probe.

### ChiR1, unresolvable candidate burns a probe interval, af3d05863

Took the general form rather than the `IsOnionCatTor` check. `feelerCandidate` now resolves inside the selection loop and skips a candidate that will not resolve, returning the `net.Addr` alongside the `*wire.NetAddress`. The short-circuit you cited is there (`peer_server.go:4328`), and since `NetAddressKey` renders every non-OnionCat address as an IP literal, an ordinary entry pays a `net.ParseIP` and no lookup.

One thing the general form exposed that the narrow one would have hidden. I put the resolve **before** the ban check, not beside it. `BanList.IsBanned` falls through to `net.ParseIP` and logs at error when that fails (`internal/banlist/ban_list.go:228`), so an OnionCat address reaching it costs an error line. With the old code that was one per burnt slot; with the pass now continuing it would have been up to `feelerCandidateTries` of them per pass. I watched that happen: on the first cut of this fix, running the new test printed a hundred `ERR_LEVEL invalid IP address passed to IsBanned: aaaqaaqaamaaiaaf.onion` lines. So the fix would have traded a silent slot burn for an error storm, which is not a trade this series should be making.

Evidence: `TestFeelerCandidateSkipsWhatItCannotResolve`, three subtests. Reverting just the `continue` so the loop falls through with the unresolved address fails all three:

```
--- FAIL: .../an_unresolvable_address_is_the_only_thing_in_the_book
      Expected nil, but got: &wire.NetAddress{... IP:net.IP{0xfd, 0x87, 0xd8, 0x7e, ...}}
--- FAIL: .../the_pass_keeps_looking_past_one
      expected: "8.8.8.8:8333"
      actual  : "aaaqaaqaamaaiaaf.onion:8333"
--- FAIL: .../an_unresolvable_address_never_reaches_the_ban_list
```

On the coherence point: the legacy layer can never hold an onion peer. `connectedHosts` is built from `serverPeer.Addr`, which is `conn.RemoteAddr` for an inbound peer and a `net.Addr` from `addrStringToNetAddr` for an outbound one, and all three connection-manager entry points go through `addrStringToNetAddr` (`peer_server.go:2948`, `:4116`, `:4180`). So neither side of the comparison in `feelerHostKey` can be a `.onion` name. I left the function alone, because it has no onion-specific branch to remove: the rendering happens inside `NetAddressKey`, and matching whatever `NetAddressKey` produces is the property that matters. What I changed is the comment, which claimed the pairing covered a case this layer cannot reach.

The probe now has exactly one path out with no info line, and it is the honest one: nothing was drawn at all, so there is no address to report an outcome against. It logs at debug.

### ChiR2, unlocked KnownAddress reads during selection, 819c2bb90

Real, and reproducible. I wrote the test first and it failed under `-race` before any fix, naming exactly the pair you did:

```
WARNING: DATA RACE
Read at 0x00c000b3cb08 by goroutine 104:
      services/legacy/feeler.go:902          <- ka.LastAttempt()
Previous write at 0x00c000b3cb08 by goroutine 103:
      services/legacy/addrmgr/addrmanager.go:1129   <- Good

WARNING: DATA RACE
Write at 0x00c000b3ccc8 by goroutine 103:
      services/legacy/addrmgr/addrmanager.go:1081   <- Attempt
Previous read at 0x00c000b3ccc8 by goroutine 104:
      services/legacy/feeler.go:902
```

I did not take the `(ka, lastattempt)` sketch, because it fixes half of it. The caller still holds the `*KnownAddress`, and reading `ka.na` off it afterwards is the same unlocked read. `UnverifiedAddress` now returns a snapshot type instead of the entry, assembled entirely inside the lock, so there is no handle left to race.

On your implied question about the pointer: I checked every write to a `KnownAddress`'s `na` before deciding, and returning the pointer is safe rather than lucky. The manager never mutates a published `wire.NetAddress` in place. All three sites that change one copy the struct, edit the copy and swap `ka.na`: `updateAddress:312`, `Connected:1101`, `SetServices:1219`, two of them saying so in a comment. If any of them had written through the pointer the fix would have had to be a copy, and the snapshot type makes that a one-line change if it ever does.

The snapshot carries `Attempts` as well as `NetAddress` and `LastAttempt`. Selection does not read it, but it is the third field that decides whether an address is worth dialling and a snapshot missing it is a trap for the next caller.

Proof it is gone: `go test -race ./services/legacy/... ./settings/...` is clean. Narrowing the new lock scope so the three fields are read after `Unlock` brings both reports straight back, which is what stops the test from merely describing the current shape of the code.

`newAddressFunc` at `peer_server.go:4104` still does the same thing. I left it: it is the connection manager's path, it predates this PR, and widening the diff to reach it is how this branch grows another review round.

### ChiR3, handshake-timeout guard re-warns on every probe, 2131a8ee6

Took the server-field option, for the reason you gave: `runOneProbe` and every probe test drive `feelerProbe` directly, and a parameter would have those tests supplying the value the production path is supposed to be responsible for. `startFeeler` resolves it once, beside the token channel it already sets up there. `feelerHandshakeTimeout` stays a pure function, so the table test at `feeler_test.go:1225` is untouched. The field is written before the goroutine that reads it exists and never again, the same ordering argument `feelerSlots` already relies on.

The resolved deadline is now on the startup line, so the value a misconfigured node fell back to is still reported once, where a startup mistake belongs:

```
[Feeler] Starting with 1 slot(s), mean interval 2m0s, handshake deadline 25s
```

Before and after, on a node with `legacy_feelerHandshakeTimeout` set to a non-positive value. Before: nothing at startup, then `[Feeler] legacy_feelerHandshakeTimeout must be positive, using 25s` at warn on every probe, for ever. After: that line once, at startup, and one info line per probe from then on.

Evidence: `TestFeelerHandshakeTimeoutIsSettledOnce`. Putting the resolution back inside the probe fails both subtests:

```
--- FAIL: .../startFeeler_warns_once_and_keeps_the_corrected_value
      Messages: the corrected deadline has to be carried, not recomputed
--- FAIL: .../a_probe_never_re-validates_it
      Should be zero, but was 1
      Messages: a startup misconfiguration must not be re-reported per probe:
                [Feeler] legacy_feelerHandshakeTimeout must be positive, using 25s
```

The second subtest uses an answering listener rather than the mute one, so the probe ends on the version instead of on the deadline and the warn count is the only thing separating the two versions.

### Elsewhere

`docs/references/settings/services/legacy_settings.md` gained the selection-skips-unresolvable rule and the "warned once, at startup" wording, in the same commits. Nothing in the PR description was falsified; ChiR3 makes its "each probe logs a single line at info" claim true of misconfigured nodes too, which it was not before.

`go test`, `go test -race` and `go vet` are clean over `./services/legacy/...` and `./settings/...`, and `make lint` reports 0 issues.

freemans13:
Amended after posting the reply above: 24fe7bd73 is the head now, a one-line comment fix in `feelerCandidate`. The doc comment still pointed at the end of the loop for the reasoning behind the resolve, which was true of my first cut; it leads the other tests now, for the ban-list reason set out above. No behaviour change.

On the previous head's `golangci-lint` failure: it was infrastructure, not code. The job died in 28 seconds at `golangci-lint config verify`, which could not fetch `golangci-lint.run/jsonschema/golangci.v2.10.jsonschema.json` before its client timeout, so no linter ever ran and no report was produced. `golangci-lint run --timeout=10m --disable gosec --disable prealloc ./...`, the same arguments the workflow uses, reports 0 issues locally over the whole repo.

ctnguyen (approved):
Review at commit [24fe7bd7](https://github.com/freemans13/teranode/commit/24fe7bd735510fd52ba1f581ae493a242c42234a)

Four commits since the last round, one per point, and all three are properly closed. The delta is small and easy to follow, and in two places it is better than what the previous review asked for. ChiR1 took the general form — the resolve moved into the selection loop and a candidate that will not resolve now costs one try out of `feelerCandidateTries` instead of the whole interval (`services/legacy/feeler.go:900-905`) — and the ordering of the resolve ahead of the ban check turns out to have a reason the previous review did not have. ChiR2 came back as a snapshot type rather than the `(ka, lastattempt)` pair the previous review sketched, which is the right call for the reason given: that sketch left the caller holding the `*KnownAddress`. ChiR3 is settled once at startup and the resolved value now shows up on the `[Feeler] Starting` line, which is where a startup mistake belongs. The new tests are the same quality as the rest of the suite — the "keeps looking past one" subtest in particular is what makes the onion assertion mean anything, and the ban-list subtest pins the ordering rather than just the outcome. One new note below, non-blocking and latent rather than live; nothing else in the delta needs attention.

## Status of previous points

Two of the three came back in a better shape than the previous review proposed, and that is worth stating plainly: on ChiR2 the previous review's suggested `(ka, lastattempt)` return was an incomplete fix, and the snapshot type is the correct shape. On ChiR1 the resolve leads the ban check for a reason the previous review did not have — `BanList.IsBanned` falls through to `net.ParseIP` and logs at error when that fails, so with the pass continuing rather than returning, an OnionCat draw would have cost an error line per try. Ordering it first costs an ordinary entry nothing, since `NetAddressKey` renders every non-OnionCat address as an IP literal that `addrStringToNetAddr` answers without a lookup.

| ChiR1 | Unresolvable candidate wasted a probe | | FIXED |
| ChiR2 | Unlocked KnownAddress reads in selection | | FIXED |
| ChiR3 | Handshake warning repeated per probe | | FIXED |

## Non-blocking issues

### ChiR4 — The settled handshake deadline has no fallback where it is used

**Problem:** `feelerProbe` builds its deadline straight off the field, with nothing between the two: `timer := time.NewTimer(s.feelerHandshake)` (`services/legacy/feeler.go:497`). `feelerHandshakeTimeout`'s own doc comment states what a non-positive value costs — "A timer built from zero or less fires immediately, so every probe would report a timeout before the far side could possibly answer" (`feeler.go:188-190`) — but that guard now runs in `startFeeler` (`feeler.go:294`), so the zero value of the field is the one input to the timer nothing checks. The consequence is that a zero deadline leaves the probe unable to wait at all: `timer.C` is ready the moment the select at `feeler.go:516-527` is reached, so unless a version has landed in that same instant the timeout arm is the only ready case, and the probe gives up before the far side could answer. `judgeVersion` then returns the fallback because `res.done` is not closed (`feeler.go:615-619`), so there is no ban and no `Good`. That last step is also what keeps the reading narrow: even when the timeout arm wins the select, `judgeVersion` re-reads `res.done` and would still judge a version that had arrived (`feeler.go:610-613`), so the "verifies nothing" outcome follows from the probe never waiting rather than from which arm happened to win. But `s.attemptIfRunning(na)` has already run after the TCP connect (`feeler.go:485`), so `Attempt` moves `lastattempt` and increments `attempts` when the node is connected elsewhere (`addrmgr/addrmanager.go:1114-1120`). Every probe would connect, instantly report "timed out", write a discouragement against the address and verify nothing.

**Why it matters:** Production is safe today and I am claiming no more than a latent trap. The only non-test caller of `feelerProbe` is `feelerHandler` (`git grep -n feelerProbe` over tracked files gives `feeler.go:365`), and `startFeeler` writes the field at `feeler.go:294` before starting that goroutine at `:298`. What makes it worth a line is that the invariant is held by a comment and a call-graph fact rather than by the code, and the tree already has two sites stepping around it: `runOneProbe` backfills the field when it is zero (`feeler_test.go:1118-1119`), and `TestFeelerHandlerWaitsForASlotToken` starts `feelerHandler` directly with the field unset (`feeler_test.go:1350`), so it already drives the zero-value path. Its only assertion is `srv.feelerAttempted.Load() > 0`, and that counter is incremented at `feeler.go:411`, before the dial at `:413` and long before the timer at `:497` — so the assertion holds whatever the timer does. The path is untested rather than protected. The failure shape is also the least visible one for this feature: inert while degrading the book it exists to improve, since `chance()` multiplies weight by 0.01 for ten minutes after an attempt and divides by 1.5 per counted failure (`addrmgr/knownaddress.go:66-73`) and `isBad` condemns a never-successful address at three (`knownaddress.go:103`). To be clear about where this comes from: the server-field shape is the one the previous review's own ChiR3 fix offered, so this is a cost that suggestion carried rather than something introduced here.

**Fix:** One line at the read site, which also lets `runOneProbe` drop its backfill:

```go
  // Settled by startFeeler. The fallback is for any path that reaches a probe
  // without having gone through it.
  handshake := s.feelerHandshake
  if handshake <= 0 {
      handshake = defaultFeelerHandshakeTimeout
  }

  timer := time.NewTimer(handshake)
```

  Settling the field in `newServer` beside `setFeelerBudget` (`peer_server.go:4184`) would close it too, but it would emit `feelerHandshakeTimeout`'s warnings on a node that has feelers switched off, so the guard at the read site is the cheaper of the two.

freemans13:
ChiR4 is fixed in `1ab956586`. Every claim in it held, and chasing it found one more thing.

**The finding.** Confirmed end to end before touching anything: the read site at `feeler.go:497` was `time.NewTimer(s.feelerHandshake)` with nothing between the field and the timer; `attemptIfRunning` runs at `:485`, after `AssociateConnection` and before the timer is ever built, so the discouragement is already written by then; and `feelerAttempted.Add(1)` is at `:411`, before the dial. That last one is the part worth repeating back, because it is what makes the path untested rather than merely unprotected: `TestFeelerHandlerWaitsForASlotToken` starts `feelerHandler` with the field unset and asserts only `feelerAttempted.Load() > 0`, a counter that moves before the dial and long before the timer, so the test drives the zero-value path and cannot see anything it does.

**The fix.** Your guard at the read site, for your reason. Settling it in `newServer` beside `setFeelerBudget` would emit `feelerHandshakeTimeout`'s warnings on a node with feelers switched off, which is the thing ChiR3 just removed. It is silent for the same reason.

`runOneProbe`'s backfill is dropped, as you suggested. That turned out to matter more than a tidy-up: **it broke `TestFeelerProbeUsesTheConfiguredHandshakeTimeout`**, which set `FeelerHandshakeTimeout` to 250ms and relied on the helper to carry it to the probe. With the backfill gone the field stayed zero, the new fallback gave it the 25s default, and the test blew its 10s budget. The helper had been the thing under test rather than the production path. It now settles the field the way `startFeeler` does, which is the hop the test's own doc comment says it is pinning.

**Evidence.**

```
--- FAIL: TestFeelerProbeWaitsWhenItsDeadlineWasNeverSettled (0.01s)
        Error:      Not equal:
                    expected: 0x1
                    actual  : 0x0
        Messages:   a probe whose deadline was never settled must still wait for the version
```

That is with the guard removed and nothing else changed. The 0.01s is the finding itself: the probe connected, timed out and gave up before the listener could speak. The test asserts the end state rather than the resolved duration, against a listener that holds its version back 250ms, so the question is only whether the address was promoted. With the guard it passes under `-race`.

`services/legacy`, `services/legacy/addrmgr` and `settings` pass with `go test` and `go test -race`. `make lint` reports 0 issues.

**On where this came from.** Noted, and I would rather have the trap than the alternative: the server-field shape is still right, because a parameter would have `runOneProbe` supplying the value production owns. ChiR4 is the cost of that choice made explicit, which is a better place for it than a comment.

Thanks for both rounds. The onion chain in ChiR1 was the pick of them: four files deep, and the only reason the `IsBanned` error-per-try problem surfaced at all was that the general form you pushed for let the pass keep going where the narrow check would have quietly returned.

liam (approved):
Approving. Race-clean, the accounting is right, and the guards I tried to break mostly broke loudly.

## Verified

**The reservation arithmetic does what the description claims.** `feelerBudget` refuses when `maxPeers-configured < targetOutbound`, which is the whole automatic tier and not merely one peer, so the starvation case from your own first round cannot come back quietly. With the shipped defaults (cap 20, feeler 1, target 8) the admission ceiling becomes 19 and the inbound share goes 12 to 11, which is the svnode shape the description says it is. Reading the target off the connection manager rather than recomputing it is also right for the reason given: `connmgr.New` substitutes its own default, so a recomputed zero would clear every count.

**The token lifecycle cannot leak or block.** `defer func() { s.feelerTokens <- struct{}{} }()` sits at the top of `feelerProbe`, so it covers every return path including a panic, and the channel is created at `feelerSlots` capacity and pre-filled with exactly that many, so the release can never block. The `default:` arm on the acquire means a tick with every slot busy is skipped rather than queued.

**Mutation results.** Deleting the BSV user-agent check fails four tests (`TestFeelerDoesNotPromoteNonBSVPeer`, `TestFeelerProbeOfANonBSVHostStaysQuiet`, `TestJudgeVersionHonoursAVersionThatRacedTheTeardown`, `TestServerPeerOnVersionRespectsDisableBanning`). Disabling the outbound-target guard fails three (`TestFeelerBudget`, `TestFeelerBudgetNamesTheReasonItRefused`, `TestSetFeelerBudgetUsesTheManagersEffectiveTarget`). Both clean.

`go test -race` is green across the whole legacy tree and `settings`.

## One thing to re-check

Your first round records that the slot cap was untested, and that the second round "fixed it properly". I ran what I think is the faithful mutation, deleting both the token acquisition and its release so there is no cap at all, and the suite **hung** rather than failing. It never produced a result inside seven minutes.

A hang is caught by a CI timeout, so this is not a hole exactly, but it is a much weaker signal than an assertion: it gives no name, no line, and on a busy runner it reads as flakiness rather than a broken invariant. I could not cleanly separate "the test detects the missing cap" from "removing the release lets probes spawn unbounded and wedges the package", so I am reporting the observation rather than a diagnosis.

Worth confirming the cap fails with an assertion rather than by wedging. If it only wedges, the test that is supposed to pin the reservation is pinning it by accident.

## Noted, not objections

The disclosed limits all look correctly judged to me: one lingering probe goroutine per reserved slot at shutdown bounded by the dial timeout, the residual probe-versus-dial race that would need a cross-subsystem reservation to close properly, and the cap bounding registered peers rather than raw sockets. Each is the right call given what Teranode's accounting actually is, and each is stated rather than buried.

Moving the probe peer's warnings down to debug is also right, and the reasoning about the fork-client teardown owning the log is the sort of thing that only shows up in a soak. Good that it did.

Nothing blocking.

freemans13:
Reproduced your hang, and it separates cleanly. The cap is pinned by an assertion, not by wedging. No code change.

**What your mutation actually removed.** Deleting the acquisition *and* the release takes out two different things. The acquisition is the cap. The release is what `drainFeelerProbes` reads to know a probe has finished (`feeler_test.go:1039`), and it waits 45s per token before calling `t.Error`. So the release deletion makes six tests that have nothing to do with the cap each sit for 45 to 90 seconds:

```
--- FAIL: TestFeelerRecordsAFailedDial (45.01s)
--- FAIL: TestFeelerDoesNotPromoteNonBSVPeer (90.05s)
--- FAIL: TestFeelerHandlerWaitsForASlotToken (46.05s)
--- FAIL: TestFeelerProbeUsesTheConfiguredHandshakeTimeout (45.27s)
--- FAIL: TestFeelerHandshakeTimeoutIsSettledOnce (45.02s)
--- FAIL: TestFeelerProbeWaitsWhenItsDeadlineWasNeverSettled (45.27s)
panic: test timed out after 7m0s
```

That is your seven minutes: six helper timeouts in series, not one hang. Note `TestFeelerHandlerWaitsForASlotToken` is in that list and **did** fail with its own message, so even under the compound mutation the cap was named. It was just buried under the slow ones.

**The faithful mutation.** Delete the acquisition only, so probes are uncapped but a finished probe still returns its token:

```
--- FAIL: TestFeelerHandlerWaitsForASlotToken (1.04s)
        Error:      Condition satisfied
        Messages:   with its only slot taken the loop must skip its turn rather than probe anyway
FAIL    github.com/bsv-blockchain/teranode/services/legacy       9.431s
```

1.04 seconds, named, and the rest of the package is unaffected. That is the signal you were looking for.

**Why the split matters for reading the test.** `TestFeelerTokensCapProbesInFlight` exercises the channel directly and says in its own doc comment that deleting `feelerHandler`'s acquisition would leave it green, pointing at `TestFeelerHandlerWaitsForASlotToken` as the one that drives the real loop. The two are deliberately a pair, and the loop half is the one that carries the assertion.

One thing your run did expose that I would not have found otherwise: `drainFeelerProbes` is a poor failure reporter when the release is what broke, since it burns 45s and reports "a feeler probe never finished" rather than pointing at the release. That is a test-helper diagnostic, not a production path, so I have left it alone rather than add churn to a PR that is otherwise ready. Happy to tighten it in a follow-up if you think it is worth the noise.

Thanks for running the mutations, and for doing it across the tree rather than only on the code under review.

Issue #556, opened 2026-03-04
Cohort-based mined state: replace the per-transaction mined writes with a cohort→block map


freemans13:
## The problem

Every time a block is found, teranode writes "you have been mined" onto **every transaction record in the block**: BlockIDs, BlockHeights, SubtreeIdxs, plus clearing unminedSince and Locked — hundreds of millions of Aerospike updates per block. Measured on scale-1 (2026-08-14, 40 blocks averaging ~357M txs, peaks ~570M): the pass runs at ~3.4M records/s and takes **1–3 minutes per block**, and the next block cannot finish processing until it completes (`waitForPreviousBlocksToBeProcessed` gates on `mined_set`, which is only set after the full pass). The deeper cost is the volume: this pass is **roughly a third of all UTXO-store writes**, it scales linearly with TPS (~600M writes/block at 1M TPS, ~900M at 1.5M), and every one of those writes exists only to record a fact that is identical for huge groups of transactions. The design below replaces it with a per-block cost that is effectively constant.

## The idea

Stop writing mined-state onto transactions entirely.

Instead, give every transaction **one tiny label at creation** — which one-second slice it was born in. Call that group of transactions a **cohort**. When a block is found, record **which cohorts the block contains**: a few hundred small rows in the blockchain DB instead of hundreds of millions of record updates.

> **A transaction's blocks = the `cohort_map` rows for its cohort. It is mined iff one of those rows is on the main chain. No rows = not mined.**

That one rule answers "is this transaction mined, and in which block, at what height" for normal blocks, fork blocks, and reorgs — no special cases. Transactions created after the miner snapshotted its candidate belong to cohorts the block simply doesn't map, so the largest class of "leftovers" costs **zero writes** to handle.

## How it works

**Stamping.** The validator computes `cohort = floor(wall_clock / 1s)` locally at create — no counter, no coordination between the ~17 embedded validator instances — writes it as a 4-byte bin on the record (the *only* mining-related field a record ever carries), and sends it with the transaction to block assembly (~5 bytes on the wire). Two clock guards: never stamp at or below the newest mapped cohort, and never map a cohort younger than ~7s; mapped cohorts naturally trail the clock by 5–20s, so these almost never fire.

[...]

PR #1635, opened 2026-08-19
feat(utxo): add the cohort spine for cohort-based mined state

freemans13:
## What happened

Issue #556 describes a complete end state: stop writing mined-state onto every transaction in a block, and instead give each transaction one 4-byte "cohort" label at create time and record, per block, which cohorts it contains. That end state is roughly nine interlocking mechanisms spread across six services, and the 2026-08-14 design review left three blockers and nine majors open against the write-side protocol (the counter/flag/heal machinery). Building all of that in one go would produce a diff nobody could review and a node nobody could trust.

So this PR builds the *spine*: the pieces every later phase has to read, chosen so that they are self-contained, testable without containers, and completely inert until a feature flag is turned on. Nothing reads a cohort yet. No consensus path changes. With `utxostore_cohortStamping` off — the default — no transaction is stamped at all and the stored label stays zero.

Three parts.

### 1. `util/cohort` — the rules, with no I/O

A new pure package holding the things the rest of the design has to agree on, so they are defined once and tested hard rather than re-derived at each call site.

- **The ID scheme.** `ID` is a `uint32` split into three ranges that partition the space exactly: sentinels (`Unset` 0, `Historical` 1, `BornMined` 2), synthetic numbers `[3, 1231006505)` for cohorts minted by splits, and wall-clock seconds from the Bitcoin genesis timestamp upward. Predicates, `String()`, and conversion to and from time, with the range errors.
- **The stamper and its clock guard.** Once a clock cohort has been mapped to a block, a transaction stamped into that same second would read as mined without ever having been in the block — so the stamper keeps a floor and never issues at or below it, falling back to `floor+1` and counting the event rather than failing when the clock is slow, frozen or running backwards. `ObserveMapped` deliberately **ignores synthetic IDs**: they are allocation counters from a sub-genesis range and are not magnitude-comparable with clock cohorts, so letting one set the floor would switch the guard off entirely. That was a specific finding of the design review; there is a test proving it.
- **`CanMap`** — the other side of the same guard: a clock cohort may not be mapped until it is a few seconds old, so in-flight writes are not stranded outside a block that claims their second.
- **`Classify`** — a cohort is wholly inside a block, wholly outside it, or straddling.
- **`PlanSplit`** — the crash-safety property of the whole design lives in one ordering: insert the fresh cohort's inherited map rows *first*, while the members about to move are still covered by them; then re-stamp the smaller side; then insert the new block's row. The planner returns that as an ordered list of typed steps rather than prose, so an executing caller cannot get the order wrong, and the doc comment walks through why a crash between any two steps leaves every transaction reading correctly — worst case "not mined yet", which is the safe direction. There is a test named after the worked example in the issue.

Statement coverage of the package is 100%.

### 2. The blockchain DB — two insert-only tables

`cohort_map(cohort, block_id, member_count, verified)` and `cohort_split_allocations(source_cohort, block_hash, new_cohort)`, in both the Postgres and the SQLite DDL, with `RecordCohortMap`, `CohortBlocks` and `AllocateSplitCohort` on the blockchain store.

Two things are worth calling out.

**The headline test.** `cohort_map_test.go` builds a chain, maps a cohort to a block on it, then takes that block off the main chain, and asserts the cohort's answer flips purely through `blocks.on_main_chain` while the contents of `cohort_map` stay byte-identical. That is the property the whole design rests on — a reorg becomes a flag flip with zero per-transaction writes — and it is now pinned by a test. The insert-only property is pinned too, and mutation-tested: changing the SQLite `INSERT OR IGNORE` to `INSERT OR REPLACE` makes that test fail.

**`CohortBlocks` deliberately does not answer "is it mined".** It returns `OnMainChain` and `Invalid` as data. `blocks.on_main_chain` can be transiently false for a block that *is* on the best chain — `CheckBlockIsInCurrentChain` documents this at length and refuses to answer false from the flag alone, because a false negative there gets escalated into a permanent block invalidation. Baking a verdict into the store would inherit that risk on a consensus-relevant answer, so the caveat is in the doc comment and the decision is left to whoever writes the reader slice.

**And a live bug fixed on the way.** `createPostgresSchema` consults `isBlockchainSchemaCurrent` first and, if it returns true, runs no DDL at all. That probe checked the `blocks` and `block_id_reservations` tables and a list of columns and indexes — but not `scheduled_blob_deletions`, which means an existing Postgres deployment that predates that table can never gain it. Adding `cohort_map` without touching the probe would have reproduced the same defect, so the probe now checks every table the DDL creates, `scheduled_blob_deletions` and `state` included.

### 3. The label on the record

`Cohort` on `meta.Data`, `fields.Cohort`, `CreateOptions.Cohort` and `WithCohort`, persisted and read back by both stores: a `cohort BIGINT` column on `transactions` threaded through all three SQL insert paths (plain, CTE, batched) and both read paths, and a `cohort` bin on the Aerospike master record with the matching case in the decorate switch.

**The cohort is deliberately absent from all four `meta.Data` binary codecs.** `Bytes()` ends with an unlengthed, unversioned run of 4-byte block IDs and `NewDataFromBytes` reads the whole remainder to EOF as exactly that — so four bytes appended there is not a parse error, it silently decodes as *an extra BlockID*, which is mined-state corruption on the consensus path. `MetaBytes`/`NewMetaDataFromBytes` have the same shape behind a fixed 17-byte header with no version field. There is a test that round-trips a `Data` carrying a cohort through `Bytes()` and asserts the BlockIDs come back unchanged, documenting both the reason and the consequence: anything that goes through those codecs reads the cohort back as zero.

Stamping is wired at the two validator create sites, which is where mempool transactions are born. Transactions created as part of a block already carry `WithMinedBlockInfo` — they are mined at the moment they are created and their cohort is never consulted — so they get the `BornMined` sentinel, and snapshot imports get `Historical`. Every stamp site is behind the flag.

## Deliberately not in this PR

- **The gRPC/`ClientI` surface for the cohort tables.** Nothing outside the blockchain service needs them yet; adding the proto, service, client and four mock implementations would double the diff for no current consumer.
- **The block-assembly cohort tally**, which the issue's phase 0 asks for as a measurement. Mapping it turned up a problem worth recording: block assembly cannot see the cohorts of transactions it was not already holding. The block side of the movement walk has only hashes — `CreateTransactionMap` parses the 48-byte subtree leaf and keeps the 32 bytes of hash — and for a block this node mined the walk takes the own-block branch and never reaches the remainder path at all. So a tally built there has two systematic blind spots and the metric would not reveal either. Widening the subtree leaf is not an option: `subtree.Node` lives in `go-subtree` and its size is a network-visible format. Block assembly will need to carry the cohort in its own in-memory working set, which interacts with the `currentTxMap` double-buffer swap — a design question, not a wiring job.
- **Everything on the write-side protocol** the design review left open: the create counters, the flags, the heals, the repair scan, the manifest. Those need the blockers answered first.

## Testing

Run in the worktree, on this change set:

- `go build ./...` — clean.
- `go vet ./...` — four `copylocks` findings, all pre-existing in `test/utils/`, identical before and after.
- `go test -race -count=1` over `util/cohort`, `model`, `stores/blockchain/...`, `stores/utxo/...`, `settings`, `cmd/seeder`, `cmd/seedimport`, `services/validator`, `services/blockvalidation/...`, `services/legacy/netsync`, `services/blockassembly/subtreeprocessor` — all pass.
- `util/cohort`: 100% statement coverage, and `-race -count=5` clean.
- `golangci-lint` and `staticcheck` over every touched package — the 20 `prealloc` findings are all in files this change does not touch (checked with `comm` against the changed-file list).
- Three mutation tests, to prove the tests are load-bearing rather than decorative: `INSERT OR IGNORE`→`INSERT OR REPLACE` breaks the insert-only test; forcing the SQL store's cohort to zero breaks the round-trip test; replacing the settings loader line with a literal `false` breaks the settings test. All three reverted and re-verified.

### Limits of that testing, stated plainly

**Docker was not available on this machine, so nothing container-backed ran.** Three parts of this change have therefore never been executed:

1. The Postgres schema probe — all eight `SchemaCurrent_PostgreSQL` tests skip. This is the change with the worst failure mode (silently no DDL) and it is Postgres-only by construction.
2. The Postgres-only UTXO insert paths, `createCTE` and the batched create. Covered here only by a static check that the shared statement's `$1..$27` placeholders match the 27 arguments at both call sites, plus the `sqlmock` DDL step list.
3. The Aerospike cohort read path — the new `case fields.Cohort` in `BatchDecorate`. The write side has container-free coverage (`cohort_bin_test.go`); the read side does not.

A CI run with Postgres and Aerospike is needed before anyone trusts those three.

Two further gaps, both deliberate and both mattering to the reader slice rather than this one: `txmetacache` round-trips `meta.Data` through the codecs that omit the cohort, so a cache hit reads it back as zero; and `stores/utxo/unmined_serialization.go` builds an `UnminedTransaction` that carries no cohort at all.

## Two decisions worth a second opinion

- `meta.Data.Cohort` is a plain `uint32` while the API surface uses `cohort.ID`, so every store boundary converts. That keeps the store types free of the new package, at the cost of the type system not stopping someone assigning a block height into the field.
- `IsSentinel(Unset)` returns true, which keeps the three ranges exactly exclusive. If a later slice wants `IsSentinel` to mean "a recorded sentinel, so not Unset", it is cheaper to change now.


freemans13:
## Review round 1

Copilot could not review this PR (quota limit), and the Claude Code Review workflow skips drafts, so I ran a review of the diff myself instead of reporting a clean round off two no-shows. Sonar's quality gate passed with 0 new issues. All CI checks pass on 94a7539.

**Copilot's exact refusal**, twice, on 94a7539: "Copilot was unable to review this pull request because the user who requested the review has reached their quota limit."

### What the review found

No bug in the plumbing. No data race, no truncation of the 4-byte id, the `$26`/`$27` argument positions in `createCTESQL` line up with the 25 preceding args, the flag-off default is `cohort.Unset` on every path, and the schema migrations follow the existing `pg_attribute` / `pragma_table_info` patterns and are idempotent.

What it did find is that several comments assert design properties the code does not yet hold, and one mock is more permissive than the store it stands in for. Fixed in 22539ab.

### Fixed: one behaviour divergence

`MockStore.RecordCohortMap` accepted the unset cohort that `SQL.RecordCohortMap` rejects, so a test written against the mock could pass on a call the real store fails. It now rejects it before any write, so a batch carrying one bad row writes none of them. `MockStore.AllocateSplitCohort` keyed its allocation map on `cohort.ID.String()`, a human label; it now keys on the numeric id.

New test in `stores/blockchain/mock_cohort_test.go` asserts the end state rather than the call: after a rejected batch, `CohortBlocks` must come back empty for the valid row that shared it. Verified by removing the guard:

```
--- FAIL: TestMockStoreRecordCohortMapRejectsUnset (0.00s)
    mock_cohort_test.go:29:
        Error:      An error is expected but got nil.
```

Restored, and `go test -race ./util/cohort/... ./model/... ./stores/blockchain/... ./settings/...` is green:

```
ok  github.com/bsv-blockchain/teranode/util/cohort          1.324s
ok  github.com/bsv-blockchain/teranode/model               16.946s
ok  github.com/bsv-blockchain/teranode/stores/blockchain    3.646s
ok  github.com/bsv-blockchain/teranode/stores/blockchain/sql 66.201s
ok  github.com/bsv-blockchain/teranode/settings            (cached)
```

### Fixed: comments that claimed more than the code delivers

- **`BornMined` is reorg-blind, and two of its three creators run at the tip.** A born-mined transaction has no map row, so its mined answer is chain-independent, which is only sound for a block that cannot leave the main chain. Quick-validate qualifies, it runs below a checkpoint. Legacy netsync does not: `createUtxos` is reached from `ValidateTransactionsLegacyMode` for every legacy block, and the below-checkpoint flag there only selects `outpointOnly`. Nor does the block-assembly coinbase, which is created at the tip by definition and is precisely the transaction whose mined state has to reverse on a reorg.
- **The stamper's floor is per-process and never armed.** `ObserveMapped` has no production caller, so the floor stays at `Unset` for the process lifetime. It is also per `*Validator`, and a cluster runs a validator inside every propagation pod as well as the validator service, so several independent floors exist where poisoning a cohort has to be global. And `Stamp` does not order against a concurrent `ObserveMapped`: it can load the floor, decide its candidate clears it, and return while another goroutine raises the floor past that candidate.
- **`DefaultMinMapAge = 7s` has to cover create latency, and nothing bounds it.** The window runs from `Stamp()` returning through to the record being durable, which on the validator path goes through the create batcher and whatever backpressure is on it.
- **The cohort is populated inconsistently and dropped by the meta cache.** The SQL store is not field-gated: `getUnbatched` and `batchDecorateChunk` both select a fixed column list and populate `Cohort` on every read. Aerospike populates it only when `fields.Cohort` is named. One layer up, neither `MetaBytesInto` nor `metadataOnly` carries the field, so on the SQL store a `GetMeta` returns the real cohort on a cache miss and zero on a cache hit.
- **The nullable-scan comment justified a case that cannot occur.** The ALTER adds the column `NOT NULL DEFAULT 0` and backfills every existing row, so no row this store wrote can scan NULL.
- **Inherited split rows had no defined member count.** Only a subset of the source's transactions moves to the fresh cohort, so copying the source row's count across would describe the wrong cohort, and `CohortMapRow` freezes `Verified` at first insert, so a wrong count can never be corrected. `StepInheritBlocks` now specifies `MemberCount: 0, Verified: false`.
- **`CohortRoundTrip` claimed no hot read path pays for the cohort column.** The SQL store does, unconditionally.

### Not fixed, deliberately, and needing a call before the executor lands

1. **Where `BornMined` goes at the tip.** Giving legacy netsync and the coinbase a real cohort *now* would be strictly worse than the current state: with no map-row writer, they would read as never mined rather than always mined. The fix belongs with that writer, so the blocks-table flags govern them the way they govern everything else. Documented at the constant for now.
2. **Making the floor global.** It probably belongs behind the blockchain store, which already knows which cohorts are mapped. Same PR as the writer.
3. **Justifying `DefaultMinMapAge`.** Either measure stamp-to-durable latency and derive the constant, or close the gap structurally by refusing to map a cohort while any stamp at or below it is outstanding. Worth instrumenting the latency now so the follow-up has a number.
4. **The cohort store methods have no client-side route.** `RecordCohortMap`, `CohortBlocks` and `AllocateSplitCohort` are on `blockchain.Store` only, with nothing on `blockchain.ClientI` and no protobuf. In a normal multi-process deployment the validator, block validation and block assembly all reach the blockchain through the client, so only code running inside the blockchain service can touch the map. The gRPC surface is where the mandatory step order in `split.go` gets much harder, so it is worth budgeting for explicitly.

One thing checked and not reported: `create.go` writes the bin as `aerospike.NewIntegerValue(int(cohortID))`. On a 32-bit build, clock cohorts from 2038 on would wrap negative and be discarded by the `raw < 0` guard in `get.go`. The neighbouring `CreatedAt` bin already stores `UnixMilli` through the same `int` cast and so already assumes 64-bit, and Teranode is 64-bit in practice, so this is theoretical and consistent with the file rather than a regression.

Issue #1591, opened 2026-08-12
model.Block has an allocator but no releaser, so the reload path drops pooled node slices


freemans13:
Raised as a P3 in the review of #1577 and deliberately not fixed there, because it needs new API on `model.Block` rather than a local change.

## What happens

`model.Block` can be given an allocator that draws pooled `[]Node` backing slices, but has no matching way to give them back. The result is a one-way street: block validation installs `SetNodeAllocator(NodeAllocFromPool)` so subtree deserialization draws from the pool, but any release path inside `model` has to drop the slices instead of returning them.

Concretely, the reload path added in #1577:

```go
// model/Block.go:1591-1598
// Nodes are not pooled here: the pool's put function lives in
// blockvalidation and this path has no access to it.
if closeErr := b.releaseSubtreeNodesLocked(nil); closeErr != nil {
```

The `nil` is the whole issue — `ReleaseSubtreeNodes` takes a `put func([]subtreepkg.Node)`, and callers in `services/blockvalidation` pass `PutNodeSlice`, but `model` has nothing to pass.

## Why it matters, and why it is not urgent

This is a **lost optimisation, not a leak**. The dropped slices are ordinary heap memory and are collected normally; nothing is retained and nothing is corrupted. What is lost is the pooling benefit on that specific path, which exists precisely because these slices are large: the pool comment notes a 654M-tx block across ~624 subtrees of ~1M nodes is roughly 29 GB of allocation per block.

The reload path runs when a block is revalidated after its nodes were released, so it is not the hot path — but it is exactly the path that allocates a fresh full set of subtree nodes, which is when the pool is worth the most.

## Where

- `model/Block.go:117-147` — `nodeAllocator`, `SetNodeAllocator`, `NodeAllocator`
- `model/Block.go:1591-1598` — the reload path passing `nil`
- `services/blockvalidation/pools.go` — `NodeAllocFromPool`, `PutNodeSlice`, `releaseBlockNodes`

## Suggested shape

A `SetNodeReleaser` mirroring `SetNodeAllocator`, stored under the same `subtreeSlicesMu` and read the same way, so `releaseSubtreeNodesLocked` can default to the installed releaser when the caller passes nil. Worth deciding at the same time whether the allocator and releaser should be installed as a pair, since installing one without the other is what produced this asymmetry.

Issue #1590, opened 2026-08-12
Bound the peer subtree_data body in the subtree meta regenerator


freemans13:
Raised as a P3 in the review of #1577 and deliberately not fixed there, because choosing the bound is a policy call rather than a mechanical change.

## What can go wrong

`SubtreeMetaRegenerator.getSubtreeDataFromPeer` streams a peer's `subtree_data` response straight into `subtreepkg.NewSubtreeDataFromReader` with no size bound. The peer is remote and, on this path, not fully trusted.

Two existing checks limit the damage but neither bounds memory:

- `serializeFromReader` returns `ErrTxIndexOutOfBounds` once `txIndex` exceeds the subtree's node count, so a peer cannot stream an unbounded *number* of transactions.
- Each transaction is hash-checked against `s.Subtree.Nodes[txIndex]`, so a peer cannot substitute content.

Neither stops a **single** oversized transaction, or a bogus length prefix inside one, from being read into memory before anything rejects it. The count check fires only after a transaction has been fully parsed.

## Where

`model/subtree_meta_regenerator.go` — `getSubtreeDataFromPeer`, the `NewSubtreeDataFromReader(subtree, body)` call.

Note this fetch runs **inline in block validation**, so memory pressure here lands on the validation path rather than on a background fetcher.

## Why it was not done in #1577

The mechanism is trivial — wrap `body` in an `io.LimitReader`. The bound is not. A value that is too low breaks legitimate large subtrees; too high buys nothing. No existing precedent in the codebase was found to copy, and inventing a constant inline seemed worse than raising it deliberately.

Candidate approaches:

1. derive it from the subtree node count times a maximum plausible transaction size
2. reuse whatever policy bounds a block body elsewhere, if one exists
3. a configurable setting, matching the convention used for `blockvalidation_subtree_meta_peer_fetch_timeout`

## Related

**#1586** — "go-subtree: subtree-meta body parse allocates from an unvalidated length prefix" — is the same class of hazard one layer down, and its resolution may inform or subsume the bound chosen here. Worth reading first.

Also **#1421** and **#1490** on non-minimal CompactSize prefixes.

PR #1636, opened 2026-08-20
fix(model): try every subtree_data source before giving up, and stop the subtree release reaching past what a block owns

### LINKED ISSUE #1581: Cache eviction holds a block's subtree mutex while expiringmap holds its own write lock, stalling the whole cache
## What can go wrong

Evicting one block from `lastValidatedBlocks` can stall every other operation on that cache, including operations on unrelated blocks.

`expiringmap.clean()` holds the map's own write lock while it calls the eviction function. Since #1577 that callback takes `block.subtreeSlicesMu`, and `GetAndValidateSubtrees` holds the same mutex across store I/O — three retries with backoff, per subtree. So evicting a block that happens to be mid-reload blocks the cleaner, which is still holding `m.mu`, which blocks every concurrent `Get`, `Set`, `Delete` and `Len` on the cache.

Before #1577 the callback only returned node slices to a pool and never blocked.

## Where

- `services/blockvalidation/BlockValidation.go:365-372` — the eviction callback
- `util/expiringmap/expiringmap.go:239-248` — `clean()` invoking `evictionFn` under `m.mu`
- `model/Block.go` — `GetAndValidateSubtrees` holds `subtreeSlicesMu` across the store reads

## This is not a deadlock

I traced every holder of `subtreeSlicesMu` — `GetAndValidateSubtrees`, `String`, `SetNodeAllocator`, `SubtreesLoaded`, `GetSubtreeSlicesCount` — and none reaches back into `lastValidatedBlocks`. There is no lock-order inversion today, and the stall is bounded by store I/O.

It was worse before: a pre-existing self-deadlock in `GetAndValidateSubtrees`, where both error returns formatted with `b.String()` while holding the non-reentrant `subtreeSlicesMu`, would have held that mutex forever and frozen the cache permanently. #1577 fixed that, so what remains is a bounded stall rather than an unbounded one.

## Why leave it open

Both candidate fixes are policy decisions that belong to whoever owns the cache's eviction behaviour:

1. Document the invariant at the callback — "nothing reachable under `subtreeSlicesMu` may touch `lastValidatedBlocks`" — and rely on review to hold the line. Cheap, but it only prevents the inversion, not the stall.
2. Use `TryLock` and veto the eviction by returning `false` when the block is busy. Removes the stall, but changes eviction semantics: a busy block survives its TTL and is retried on the next tick, so memory is held longer than the cache contract implies.

## Provenance

Found in the pre-merge review of #1577 and recorded in its limits section.

## Prior art — the sibling case was already accepted and fixed

**#1152, "util/expiringmap: blocking send to evictionCh under write lock (latent)", is the same hazard through the other door** and was fixed. It reported that `clean()` does a blocking send to `evictionCh` while holding `m.mu`, so a slow consumer would block the ticker under the lock and deadlock every `Get`/`Set`/`Delete`/`Len`. It was rated low only because it was latent: it explicitly noted that "no prod code wires `WithEvictionChannel` (only `WithEvictionFunction` is used)".

That is precisely the gap this issue sits in. The channel path was hardened; the eviction-*function* path — the one actually used in production — can now block under the same lock, because #1577 made the callback take a block mutex that `GetAndValidateSubtrees` holds across store I/O.

So the argument for acting was already accepted for the latent variant. This is the live variant of it.

freemans13:
## What happened

This PR was opened against a main that did not yet have a fix for the regenerator's source fall-through. While it was in review, #1486 landed one, arriving at the same conclusion independently and, on two points, with a better answer. So this branch has been merged with main and rewritten to build on main's version rather than compete with it.

What that means concretely: main's `tryBuild` closure and its local-then-peers fall-through are now the base. Main's `isFirstSubtree` parameter, threaded through `SubtreeMetaRegeneratorI`, `RegenerateMeta`, `buildAndStoreMeta` and `buildMetaFromSubtreeData`, is the mechanism for the index-0 exemption. Main's `storeRegeneratedMeta` returning `(cached, error)` stays, along with its two completeness checks and its per-source `Warnf` logging. This branch's own versions of all of that are gone.

The original P0 that opened this PR is therefore fixed by #1486, not by this branch. What remains here is four things #1486 does not do, plus one it does not touch at all.

## What this PR adds on top of main

**1. A cache-busting second fetch.** Main's peer loop gives up on a peer the moment its body fails to build a meta. For the failure this path exists to survive, that discards the peer most likely to be able to help. Peers front the asset service with an nginx `proxy_cache` that stores any 200 for its TTL, so an aborted on-demand generation reaches the client as "200 with an empty body" and is replayed to every byte-identical request (#1368). With one announcing peer, which is the usual case, that wedges the subtree until the upstream entry expires however many times we ask.

A peer that answered now gets one more fetch on a cache-busting URL before the loop moves on. The cache key includes the query string while nginx location matching ignores it, so the busted URL reaches the same handler and misses the cache. Only a peer that *answered*: a refused connection or an expired budget left no body for a cache to be blamed for, and a second request would just spend the budget again. `getSubtreeDataFromPeer` reports that separately from the error, because a body that failed to deserialize is still a body the cache may be replaying.

The token is process-wide and clock-seeded rather than per-regenerator. `createMetaRegenerator` builds a fresh `SubtreeMetaRegenerator` for every `ValidateBlock` and `ReValidateBlock`, so a counter on the struct would restart at zero each time and every retry would ask for the identical `?cachebust=1` URL, which nginx caches under its own key like any other. The clock seed covers the same weakness across a restart. This is the model-package equivalent of the uniqueness `services/blockvalidation/Server.go`'s sibling counter gets from living on the long-lived `Server`.

**2. The per-fetch timeout at 10m rather than 30s.** `check_block_subtrees.go` fetches the identical payload from the identical endpoint under `blockvalidation_subtree_data_fetch_timeout = 10m`. The regenerator had 30s, which after ~7.75s of 503 backoff cannot stream a mainnet-size body, so it would fail on exactly the blocks that need it. Both copies of the default move together: `settings.DefaultSubtreeMetaPeerFetchTimeout` and `model.DefaultPeerFetchTimeout`, held to each other by `TestSubtreeMetaPeerFetchTimeout_ConstantsDoNotDrift`.

The wiring was checked rather than assumed, because a tagged and documented setting can still be dead. The key is read at `settings/settings.go:365` into `BlockValidation.SubtreeMetaPeerFetchTimeout`, passed at `services/blockvalidation/BlockValidation.go:349` into `NewSubtreeMetaRegenerator`, stored on the struct, and applied by `getSubtreeDataFromPeer`'s `context.WithTimeout`. `TestSubtreeMetaPeerFetchTimeout_IsWired` pins that end to end.

**3. The poisoned-file repair.** Falling through to a peer fixes this node's own validation and nothing else. The file the consequence lives on is `subtree_data`, not `.subtreeMeta`. The asset service's `GetSubtreeDataReader` (`services/asset/repository/GetSubtreeData.go`) checks only `Exists(hash, FileTypeSubtreeData)` and then streams the file straight back on `GET /api/v1/subtree_data/:hash`, with no validation of the body, and it is the same route this regenerator fetches from. So a node that routes around its own unusable copy goes on serving it to every peer that asks for as long as the file's DAH lasts, turning a loud local failure into a silent outward one. This PR exists because a poisoned `subtree_data` copy stalled other nodes, so closing only our half would be the wrong half.

`getLocalSubtreeData` now reports separately that a `subtree_data` file exists and cannot satisfy its subtree, and `RegenerateMeta` says so once at Warn naming the subtree. When a peer then yields a body that does satisfy it, that body is written back over the local file with `WithAllowOverwrite` and the same DAH the meta gets. Three boundaries are deliberate:

- Both unusable shapes count. A body stopping at a clean `io.EOF` deserializes "successfully" and is caught by `MissingSubtreeDataTxs`; one truncated mid-transaction fails inside `NewSubtreeDataFromReader`. The file is on disk either way, so it is served outward either way.
- An **absent** local file is not written back. It is not an outward poison at all, because the asset service regenerates it on demand from the subtree.
- The repair declines on a subtree with no nodes. `Data.Serialize` indexes `Nodes[0]` before it validates anything, so that shape panics rather than erroring, inside a `validOrderAndBlessed` errgroup goroutine no `recover()` covers.

Best effort throughout, every failure a Warn: the regeneration it runs inside has already succeeded, and failing the block over a failed repair write would turn a recovered block back into a stalled one.

**4. The subtree release and mmap ownership work.** Entirely orthogonal to #1486 and unchanged by the reconciliation.

- `GetAndValidateSubtrees`' reload closed **every** survivor before reallocating `SubtreeSlices`. A block does not exclusively own all of its subtrees: `quick_validate.go` stores the same `*Subtree` into both `block.SubtreeSlices` and the `SubtreeWriteJob` it queues on the shared write channel, and `blockassembly/Server.go` aliases the processor's live `job.Subtrees` into a `model.Block`. Gutting one of those under a write worker races its `Nodes` and makes `Serialize` write a 0-leaf `.subtree` over a good one. `closeMmapSubtreesLocked` now closes only mmap-backed survivors, which are the only ones that need it. A heap subtree needs nothing but its reference dropped.
- A closed mmap subtree's `Nodes` are detached before the mapping goes. `Subtree.Close` unmaps the region and removes the backing file and never touches `Nodes`, so dropping the old `ReleaseNodes()` call to stop damaging shared heap subtrees also stopped detaching the mmap ones, leaving every closed survivor holding a slice into memory that had just been unmapped.
- Cache eviction attempts the release instead of forcing it (`tryReleaseBlockNodes`, closes #1581). `expiringmap.clean()` holds the map's write lock for its whole sweep while `GetAndValidateSubtrees` holds a block's subtree mutex across store reads with retries, so evicting a block mid-reload parked the cleaner and queued every `lastValidatedBlocks` operation behind one block's I/O. `clean()` already `continue`s on a veto and leaves the expiry untouched, so declining defers the eviction rather than abandoning it.
- The invalid-block reload writes under the block's mutex, via a new `Block.ReplaceSubtreeSlices`, with the store reads deliberately left outside it so this does not become another hold-across-I/O.

**5. The completeness predicate, reworked for main's world.** `model.MissingSubtreeDataTxs` still consolidates the check that used to exist twice, but its index-0 exemption is now the caller's decision rather than an inference from the placeholder hash. The regenerator passes the subtree's real position, because a meta has to agree with `validateSubtree`, which skips node 0 only for `sIdx 0`. The catchup fetcher passes `true`, because it is not building a meta: it is keeping `Data.Serialize` from walking into a nil `*bt.Tx`, and `Serialize`'s rule is the hash at index 0 whatever the position. Making catchup stricter was tried first and rejected: it failed two existing tests that build synthetic blocks whose every subtree carries a placeholder at node 0, and it would charge a peer for a body `Serialize` would have handled.

## What was dropped as superseded

This branch's own ChiR4 commit (`fail the source when a regenerated meta will not serialize`) is gone. Main covers both halves and covers them better.

The bug was that `Meta.Serialize` exempts index 0 unconditionally, so an unset entry at index 0 of a subtree with no coinbase placeholder would serialize and get persisted. This branch fixed the symptom, adding a walk over the built meta that exempted index 0 only for a genuine placeholder. Main fixes the cause: `isFirstSubtree` makes the exemption agree with `validateSubtree` rather than inferring it, so the two predicates cannot disagree in the first place, and `buildAndStoreMeta` already returns the serialize refusal rather than logging it and handing the meta back anyway. Main's version is also threaded through the public interface, so a future caller has to answer the question rather than inherit a guess.

This branch's aggregated failure logging is also gone, in favour of main's per-source `Warnf` inside `tryBuild`. That is a straight loss of a nicety, kept only because the merge rule was that main's design wins on every overlap and re-litigating logging shape was not worth a divergence.

## Testing

Each re-applied fix was broken in-tree and the guarding test watched failing, then restored.

- Repair: removing the `repairLocalSubtreeData` call leaves `_PoisonedLocalData_IsRepairedFromThePeerBody` asserting the 85-byte truncated body against the 170-byte complete one. `_AbsentLocalData_IsNotWrittenBack` and `_RepairFailure_DoesNotFailRegeneration` pin the other two boundaries.
- Cache-bust: removing the retry fails `_PoisonedPeer_RetriedWithCacheBust` ("the poisoned entry must be bypassed"), `_CacheBustTokenIsUniqueAcrossRegenerators` ("attempt 0 must get past the poisoned entry") and `_PoisonedPeer_FallsThroughToNextPeer` (hit count 1, not 2). Removing the clock seed fails `_NewCacheBustCounter_IsClockSeeded` with `"0" is not greater than or equal to "1787836062912557000"`. `_UnreachablePeer_IsNotRetried` pins the other direction.
- Mmap detach: removing `_ = st.ReleaseNodes()` from `closeMmapSubtreesLocked` fails `TestGetAndValidateSubtrees_ReloadClosesSurvivingSubtrees` with "a closed mmap subtree must not keep a slice into the region Close just unmapped".
- Nodes snapshot: **no committed test distinguishes it.** Reverting it to `Length()`-then-index leaves the whole `model` package green. That is reported rather than papered over, and the reason is in the limits below.
- `make lint` clean (0 issues). `go build ./...` clean. `go vet` clean over model, blockvalidation, blockassembly, settings and util. `go test ./model/...` green, `go test -race ./model/...` green in 21.3s. `go test` green over blockvalidation, blockvalidation/catchup, blockassembly, blockassembly/mining, blockassembly/subtreeprocessor, subtreevalidation, settings and util with its subpackages.
- Two existing tests changed rather than were added to. `_PlaceholderExemptionMatchesValidateSubtree`'s assertion follows the rejection one layer up, from `buildMetaFromSubtreeData` into `getLocalSubtreeData`, because the poisoned classification has to happen before the peer loop can decide whether to write a body back. The verdict and the node it names are unchanged. `TestMissingSubtreeDataTxs` gained a subtest for each direction of the exemption argument.

## Limits, stated plainly

- **None of this has ever run on a node.** No soak, no testnet, no mainnet. Everything here is verified by unit test and by mutation of the code under test.
- **The regenerator fall-through cannot be exercised by a plain soak.** The trigger is a missing or torn `.subtreeMeta` on a block whose `subtree_data` is also unusable, which is rare enough that a normal soak would tell you nothing either way. Proving it live needs a deliberately poisoned file.
- **The cache-bust has never met a real nginx.** Its test asserts against an `httptest` server that behaves the way the issue-1368 write-up says a proxy cache does. Whether a real peer's `proxy_cache` keys and matches the way that assumes is untested here.
- **The 10m budget is per fetch, not per peer.** `getSubtreeDataFromPeer` creates its own `context.WithTimeout` on every call and the loop can call it twice for the same peer, so the worst case for one peer is 20m and for one `RegenerateMeta` call it is that again per configured peer. The call runs inside a `validOrderAndBlessed` errgroup goroutine holding a pooled parent-spends map for its whole duration. A single deadline cannot distinguish a hung peer from a slow one; doing that needs a stall detector on the stream.
- **The `Nodes` snapshot removes a reproducible panic without making the access synchronised.** A test racing the predicate against `ReleaseNodes` over 50k rounds does distinguish it, failing every run with `index out of range [0] with length 0`, but `-race` rejects that test for a real reason: it is an unsynchronised read of a field another goroutine writes. One read of a slice header cannot tear on any architecture Go supports, so this is benign in a way the race detector cannot express, but the fix is a mitigation and not a cure. Closing it properly means `ReleaseNodes` taking the subtree's own mutex in go-subtree.
- **Not fixed, found while testing that:** go-subtree's `Subtree.NodeIndex`, reached through `Meta.SetTxInpointsFromTx`, reads `st.nodeIndex` outside the lock, builds it under the lock, then reads the built map outside the lock again. A test racing `buildMetaFromSubtreeData` against `ReleaseNodes` dies there even with the snapshot applied. That is a race in the dependency, not in this PR. Worth its own issue against go-subtree.
- **`blockassembly`'s safety rests on a condition the compiler does not enforce.** With `blockassembly_subtreeMmapDir` set, `SubtreeProcessor.newSubtree` returns `NewTreeByLeafCountMmap`, and `submitMiningSolution` aliases those straight into a `model.Block` before calling `Valid`. What keeps that safe is only that `GetAndValidateSubtrees` takes its "already loaded" early exit and never reaches the release. Tighten that check and block validation would unmap and delete files the live processor is reading. Copying the slice does not help, because `Close` acts on the `*Subtree` and a clone holds the same pointers. Recorded at all three sites rather than folded in, because the durable fix is a behaviour change to a mining path this PR does not otherwise touch.
- **Declining an eviction can defer it indefinitely** if a block's mutex is never released. That would be a hung validation, a worse problem with its own symptoms, and the entry simply persists rather than anything being corrupted.
- **`ReplaceSubtreeSlices` closes nothing before it swaps.** Its only caller is safe because `setTxMinedStatus` releases the block immediately beforehand. Call-site discipline, not something the function enforces.
- **Left for follow-up, none of it on a path this PR changes.** The sibling cache-bust counter on `services/blockvalidation/Server.go` is a bare `atomic.Uint64` with no clock seed, so after a restart it asks for `?cachebust=1` again; consolidating the two URL builders would fix it but three pre-existing tests assert the exact token values. `expiringmap`'s capacity path treats a veto as "do not insert" and silently drops the new entry where `clean()` retries it, and `Stop()` runs no final sweep. `fetchAndStoreSubtree` can return an mmap-backed subtree that `fetchAndStoreSubtreeData` uses and drops without `Close`, leaking the mapping on the catchup path. Siggi's `io.LimitReader` bound stays tracked in #1590 and the pooled-slice releaser in #1591.

https://claude.ai/code/session_0173u8CBAKHNAKMPuvrxKEef


ordishs (approved):
Reviewed against merge-base `0111d8a93` (16 files, matches the PR file list).

The three concurrency fixes hold up. I checked the reasoning the comments rely on rather than taking it on trust:

- `quick_validate.go:118-127` — the mmap-capable branch really does queue a job with no `Subtree` field, so the entry it shares is heap-backed.
- `blockassembly/Server.go:1603` — `len(jobSubtreeHashes)` always equals `len(job.Subtrees)`, so that path takes the early exit and never reaches the release.
- `expiringmap.clean()` — `continue` on a veto, expiry untouched, retried next tick.
- `buildHTTPError` returns ServiceError / NotFoundError / ServiceUnavailableError, none matching `ErrExternal` by code, so the cache-busting retry fires only on a short body and not on ordinary HTTP failures.

Mutation-tested the release fix: swapping `closeMmapSubtreesLocked` back to `releaseSubtreeNodesLocked(nil)` fails both `TestGetAndValidateSubtrees_ReloadLeavesSharedHeapSubtreeUsable` and `_ReloadDoesNotWriteIntoACallerOwnedArray`. Load-bearing.

Locally: `go build ./...` clean, `go vet` clean over model/blockvalidation/util, and both new test sets green under `-race`.

One item worth settling before merge.

**`model/subtree_meta_regenerator.go:251`, `settings.conf:461`** — the peer fetch budget

The budget rises from 30s to 10m and bounds one fetch, not the call. Nothing caps `RegenerateMeta` as a whole, and `tryPeer` can make two fetches, each starting the budget afresh.

Failure: a peer that accepts the connection and stalls after the response headers now blocks `model/Block.go:1283` for 10 minutes instead of 30 seconds. A peer answering 200 with a short body slowly spends the budget twice — 20 minutes. This runs inline in `Block.Valid`, inside a `validateSubtree` errgroup goroutine that holds a pooled parent-spends map for its whole duration. The errgroup cancels its siblings on the first error, so a block stays bounded near one budget rather than N of them, but that is still a validation that used to cap at 30 seconds.

Recommendation: keep the 10m per-fetch budget — it has to cover a mainnet-size body, and the sizing argument is right — and add an aggregate deadline on the whole `RegenerateMeta` call. Then the number of sources and cache-busting retries cannot multiply the wall-clock cost, and the setting keeps a plain meaning for operators.

Smaller notes, none blocking:

- `model/Block.go:1281` and `1285` — "nothing at WARN on the success path" is not reached. Removing the `Warnf` from `buildAndStoreMeta` leaves the call site emitting two WARN lines per regenerated subtree, and those lines are outside this diff.
- `util/expiringmap/expiringmap.go:142` — this is the first eviction function that can return false. `clean()` handles a veto correctly; the capacity path in `Set` does not, and `evictOldestLocked` only ever considers the single oldest entry. Unreachable today because `lastValidatedBlocks` never calls `WithMaxSize`, but it is a new trap for whoever adds one.
- `model/peer_subtree_data.go:47` — `MissingSubtreeDataTxs` returns 0 for a nil subtree or nil data, and 0 means complete. Unreachable from both callers, but the wrong default direction for a shared predicate on a consensus path.
- `model/subtree_meta_regenerator.go:263` and `services/blockvalidation/peer_cache_bypass.go:95` — the change consolidates the completeness predicate so two copies cannot drift, then adds a second copy of the cache-bust URL builder beside the existing one.
- `services/blockvalidation/Server.go:256` — the sibling counter is a bare `atomic.Uint64` with no clock seed, so after a restart it asks for `?cachebust=1` again. That is the exact weakness the new model counter was hardened against; seeding it the same way is a one-line change.

freemans13:
Thanks — four of your items are fixed in 77fe669fa..017a51304, and reviewing change 4 the way you did turned up a regression I had introduced and the tests were not catching.

**A real defect, mine, found tracing your `closeMmapSubtreesLocked` reasoning further than I had.** `Subtree.Close` calls only its closer: it unmaps the region and removes the backing file, and never touches `Nodes`. The old `releaseSubtreeNodesLocked` did `ReleaseNodes()` before `Close()`. Dropping that call to stop damaging shared heap subtrees also stopped detaching the mmap ones, so every closed mmap survivor was left holding a slice into memory that had just been unmapped — a fault on any read, not a clean panic. The tell was in this branch's own test: `_ReloadDoesNotWriteIntoACallerOwnedArray` compares pointers as integers, with a comment saying that formatting the value would kill the test binary. I had written that off as a testing trap rather than reading it as the symptom it was. `closeMmapSubtreesLocked` now detaches inside the mmap branch and drops the slice rather than pooling it — its backing *is* the mapped region — and heap survivors are still untouched.

**On the peer fetch budget.** I think the compounding is narrower than either of us wrote, and worth pinning down before adding machinery for it. Every production call site passes exactly one peer: `createMetaRegenerator([]string{baseURL})` at `BlockValidation.go:1857`, `:1956` and `:2421`, and `nil` at `Server.go:1399`. So "that again per configured peer" in my doc comment reads far worse than reality — the worst case is 2 × 10m for one stalled peer, not 2 × 10m × N. You are right that a 20-minute inline stall where it used to be 30 seconds is the real change, and I have said so plainly.

Where I have not taken the recommendation is the aggregate deadline, because at N=1 I do not think it buys anything. To be safe it has to be at least twice the per-fetch budget — otherwise a legitimately slow first fetch that fails at 9m leaves the cache-busting retry a minute, which reintroduces exactly the "budget too small to finish" failure the 10m sizing exists to avoid. Two fetches is already the ceiling, so an aggregate of 2× is the current worst case restated. It starts earning its keep the moment the peer list grows past one, and I would rather add it then, together with whatever makes the list longer, than add a constant now that only documents a bound the code already has. The honest fix for a hung-versus-slow peer is a stall detector on the stream, which is more than this PR should carry.

**`MissingSubtreeDataTxs` fails closed now**, and it was worse than the nil default you flagged. It walked `data.Txs` while both callers report the denominator as `Subtree.Length()`, which agree only because `serializeFromReader` allocates `Txs` at `Subtree.Length()` before it reads anything. It counts unfilled subtree nodes now, so a short slice counts the shortfall and nil data means every node is unfilled. It also had no direct test at all, so the decision its comment spends thirty lines defending was unasserted; `TestMissingSubtreeDataTxs` pins it, and catches both mutations — exempting index 0 the way `Data.Serialize` literally does, and walking the body's entries instead of the subtree's nodes.

**The two WARN lines you spotted at `model/Block.go:1281` and `:1285` are Debug.** You were right that my claim was not reached: removing the `Warnf` from `buildAndStoreMeta` took a successful regeneration from three WARN lines to two, not to zero, and both survivors were outside the part of the diff I was looking at. `RegenerateMeta` owns the reporting now — one Info naming the source that worked, one aggregated Warn on total failure.

**The eviction test pins the wiring now.** You and I both had this as a known limit; it is closed by naming the closure. `evictLastValidatedBlock` lives in `pools.go`, `NewBlockValidation` registers it, and the test delegates to it instead of building its own. Reverting it to a forced release now fails `TestLastValidatedBlocksEviction_DoesNotStallTheCache` after 10.9s where it used to leave the suite green.

**Not taken, and recorded in the body instead.** Seeding the sibling counter and sharing one cache-bust URL builder is the right change and fixes both of your notes at once, but `peer_cache_bypass_test.go:25-26` and `get_blocks_test.go:3864` assert the literal tokens `cachebust=1` and `cachebust=2`, so it means rewriting pre-existing tests from inside a consensus-path change. Same for the `expiringmap` capacity-path veto — real trap, wrong PR to widen into `util/expiringmap`. Both are in the trade-offs section now, along with two things this review turned up: `expiringmap.Stop()` runs no final sweep, so a block still cached at shutdown is never released; and the cache bust only fires on a short-but-parseable body, since one truncated mid-transaction fails inside `NewSubtreeDataFromReader` with an error that is not `ErrExternal`. The sibling fetcher has that last gap too.

Green after the changes: `make lint` 0 issues, `go build ./...` and `go vet` clean, and under `-race` `model` 15.9s, `blockvalidation` 307.4s, `blockassembly`, `util` and `settings` all passing.

ctnguyen (changes requested):
Review at commit [8e2f163](https://github.com/freemans13/teranode/commit/8e2f1630ac0519c5ae85b0e98f505bb28ba5ca50)

This is a strong follow-up. The core restructure in `subtree_meta_regenerator.go` is the right shape: making fetch-and-validate part of each attempt, so an unusable source falls through exactly like an unreachable one, is a much better answer than special-casing the truncated file, and `tryLocal` / `tryPeer` / `fetchCompleteFromPeer` read cleanly. The ownership analysis behind change 4 is the part I'd single out — narrowing the release to mmap survivors, and then saying plainly at all three sites that `blockassembly` is safe because of an early exit rather than an invariant, is more useful to the next person than a confident claim would have been. I re-checked the four production call sites myself (`BlockValidation.go:1950`, `:2049`, `:2703` and `Server.go:1404`): they really do pass one peer or none, and because `validOrderAndBlessed` cancels its siblings on the first error, one stalled peer stays bounded near one budget rather than multiplying across a block's subtrees — so the reasoning for declining the aggregate deadline holds up. Catching the missing detach in `closeMmapSubtreesLocked` while tracing someone else's review comment is the kind of find that only comes from actually following the reasoning through, and the comment explaining why the slice is dropped rather than pooled is exactly right. One thing does need fixing before this merges: the fall-through turns a poisoned local `subtree_data` from a loud local failure into a silent outward one, because the asset service keeps serving that file to peers. Details in ChiR1; the other three are yours to take or leave.

## Blocking issues

### ChiR1 — Routing around a poisoned subtree_data turns a loud local failure into a silent outward one

**Problem:** The poisoned local file predates this PR — I am not claiming the change creates it. What the change does is alter what happens next, and it removes the only thing that made it visible. Before change 1, a short local `subtree_data` failed the regeneration, `RegenerateMeta` returned an error, and the block did not validate: loud, attributable, and impossible to miss. After it, `tryLocal` rejects the body (`model/subtree_meta_regenerator.go:155-157`), a peer supplies a complete one, the block validates, and the node carries on — while still serving the bad body outward. `services/asset/repository/GetSubtreeData.go:85-104` does `SubtreeStore.Exists(..., FileTypeSubtreeData)` and, when the file is present, streams it straight back through `GetIoReader` with no validation of the body at all, on the same `GET /api/v1/subtree_data/:hash` route (`services/asset/httpimpl/http.go:347`) that this very regenerator fetches from. Nothing repairs the file, nothing deletes it, nothing logs that this node is now a poisoned source, and `tryPeer` discards a complete, hash-validated body that could have replaced it. Two consequences follow. A peer's cache-busting retry bypasses only its own nginx, so the busted request lands on our asset service and gets the same short file back — the bypass this PR adds cannot break a stall whose source is a peer's on-disk file. And locally, `services/subtreevalidation/check_block_subtrees.go:309-315` takes the "SubtreeData exists, extract transactions from stored file" branch whenever the file is present, so `extractAndCollectTransactions` keeps failing the count check at `:830-832` with no re-fetch. For completeness on where such a file comes from: `processSubtreeDataStream` tees the peer body into `SetFromReader` (`:857-858`) and closes the pipe with `pw.Close()` rather than `CloseWithError` on a clean parse (`:886`), so a 200 with a cleanly-EOF-terminated short body commits successfully and only *then* trips the count check at `:902-904`, with nothing removing what was written.

**Why it matters:** This PR exists because one poisoned `subtree_data` copy stalled other nodes — that is the whole content of [bsv-blockchain/teranode#1368](https://github.com/bsv-blockchain/teranode/issues/1368) and the reason [bsv-blockchain/teranode#1577](https://github.com/bsv-blockchain/teranode/pull/1577) and this change exist. Making our own validation survive the poison while leaving the node able to *be* the poison closes the half of the theme that only helps us. The trade is a bad one in observability terms too: the pre-change behaviour was a stalled block, which is exactly the signal an operator acts on; the post-change behaviour is a healthy-looking node quietly handing a short body to every peer that asks, for as long as the file's DAH lasts. The description's "Nothing needs to delete the bad local file: the peer-built meta is stored, so the next call finds the meta and never enters the regenerator at all" is correct, but it is a statement about the `.subtreeMeta` file. The file the asset service serves is `subtree_data`, and the disclosure says nothing about it — a disclosure covering one file while a second file carries the consequence.

**Fix:** When a peer source yields a complete body, overwrite the local copy. The regenerator already holds the validated bytes and already computes the same DAH expression for the meta in `storeRegeneratedMeta`:

  ```go
  func (r *SubtreeMetaRegenerator) tryPeer(...) (*subtreepkg.Meta, error) {
      // ... existing fetch + cache-bust retry ...
      r.repairLocalSubtreeData(ctx, subtreeHash, data) // best-effort, warn on failure
      return r.buildAndStoreMeta(ctx, subtreeHash, subtree, data)
  }
  ```

  with `repairLocalSubtreeData` doing `data.Serialize()` then `Set(..., FileTypeSubtreeData, ..., options.WithAllowOverwrite(true), options.WithDeleteAt(dah))`. Deleting the short file when `tryLocal` rejects it is a smaller variant that also stops the outward propagation, though it loses the repair. Either way, `tryLocal`'s rejection should not be Debug-only: a local body that cannot satisfy its own subtree means this node is serving corrupt data, which is worth one Warn naming the subtree.

## Non-blocking issues

### ChiR2 — Nothing pins the mmap detach in closeMmapSubtreesLocked

**Problem:** `_ = st.ReleaseNodes()` at `model/Block.go:2075` is the whole of the use-after-unmap fix, and no test in the tree fails if it is removed. `TestGetAndValidateSubtrees_ReloadClosesSurvivingSubtrees` (`model/block_release_reload_test.go:133-181`) asserts only that the backing file is gone; `_ReloadDoesNotWriteIntoACallerOwnedArray` asserts only the array element's pointer identity; `TestReleaseSubtreeNodes_ClosesMmapAndNilsEntries` exercises `releaseSubtreeNodesLocked`, which detaches on its own account. The commit message for `79818ad7` records the verification as "temporarily removing the detach and watching a probe on the closed subtree's Nodes field flip from nil to a live slice" — a manual probe, not something committed. That is the one change in this PR with no automated guard, and the PR's own adversarial pass lists three concurrency fixes as load-bearing; this is not among them.

**Why it matters:** The failure mode is a read through a slice into munmapped memory, which faults the process rather than panicking cleanly — the reason the sibling test has to compare pointers as integers in the first place. It is currently unreachable only because the single caller replaces the slice header on the next line; a future caller, or a reordering inside `closeMmapSubtreesLocked`, reintroduces it silently. Everything else in this PR is mutation-tested, which makes this the odd one out rather than a general gap.

**Fix:** One assertion in the existing test, next to the backing-file check, since the closed subtree is still reachable through the local variable:

  ```go
  require.Nil(t, mmapSurvivor.Nodes,
      "a closed mmap subtree must not keep a slice into the region Close just unmapped")
  ```

  `Nodes` is only read for nil-ness here, never dereferenced, so this is safe after the munmap.

### ChiR3 — MissingSubtreeDataTxs indexes Nodes outside the subtree's lock

**Problem:** `model/peer_subtree_data.go:64-65` takes the length through `subtree.Length()` (which acquires the subtree's own `RWMutex`) and then indexes `subtree.Nodes[0]` directly, unlocked:

  ```go
  nodes := subtree.Length()
  coinbaseAtZero := nodes > 0 && subtree.Nodes[0].Hash.Equal(subtreepkg.CoinbasePlaceholderHashValue)
  ```

  Those are two separate reads of `Nodes`. `Subtree.ReleaseNodes` sets `st.Nodes = nil` without taking that mutex, so a release landing between them turns line 65 into an index-out-of-range on a zero-length slice. The regenerator reaches this with the `*Subtree` that `validOrderAndBlessed` captured out of `SubtreeSlices` without holding `subtreeSlicesMu` — the same window `Block.go:1415` and `:1252` already guard against by name, and `releaseSubtreeNodesLocked` (`Block.go:2009`) is what calls `ReleaseNodes` on it.

**Why it matters:** It panics inside a per-subtree errgroup goroutine that nothing recovers, so it kills the node rather than failing the block — the identical outcome the function's own doc comment spends thirty lines arguing against for the `Data.Serialize` case. The two copies this consolidated had the same shape, so it is not a regression, but consolidating into one shared predicate on a consensus path is exactly the moment to fix it once instead of preserving it.

**Fix:** Snapshot the slice header once and derive both the guard and the index from it, so the two cannot disagree:

  ```go
  nodesSlice := subtree.Nodes
  nodes := len(nodesSlice)
  coinbaseAtZero := nodes > 0 && nodesSlice[0].Hash.Equal(subtreepkg.CoinbasePlaceholderHashValue)
  ```

  `buildMetaFromSubtreeData` (`model/subtree_meta_regenerator.go:296`) computes `hasCoinbasePlaceholder` the same way and wants the same treatment.

### ChiR4 — buildAndStoreMeta can return a meta its own serializer rejects

**Problem:** The final assertion in `buildMetaFromSubtreeData` changed target. It used to walk the built meta (`meta.TxInpoints[i].ParentTxHashes == nil` for every checked node); it now re-runs the body predicate (`model/subtree_meta_regenerator.go:317-319`). Those are not the same question, because `SetTxInpointsFromTx` writes at `Subtree.NodeIndex(txid)` rather than at the body index, and the constructor it goes through yields a nil `ParentTxHashes` for a transaction with no inputs: `NewTxInpointsFromTx` (`go-subtree@v1.4.2/inpoints.go:88-90`) delegates to `newSizedFromInputs`, which returns a bare `TxInpoints{}` when `len(inputs) == 0` (`inpoints.go:374-378`). The distinction is load-bearing and easy to trip over, because the *other* constructor behaves differently: `NewTxInpoints()` (`inpoints.go:79-83`) deliberately returns a non-nil-but-empty slice, and its own doc comment says why — `Meta.Serialize` uses nil versus empty to tell "no inpoints set yet" from "this node has no parents". So nil genuinely means unset, which is what makes the `Meta.Serialize` guard meaningful, and a fully populated `data.Txs` can still leave a meta entry in that unset state. The only remaining detector is `Meta.Serialize`, which refuses exactly that (`i != 0 && ParentTxHashes == nil`), but `storeRegeneratedMeta` logs its error and returns (`:330-334`) while `buildAndStoreMeta` hands the meta back to the caller anyway (`:289`). Worse for index 0: `Meta.Serialize` exempts it unconditionally, whereas the body predicate only exempts a genuine coinbase placeholder, so an unset entry at index 0 of a non-first subtree serializes and gets persisted with a DAH. I have not found a path where a block that would otherwise be valid reaches this, so treat it as defence-in-depth rather than a live bug.

**Why it matters:** The comment above the check calls it a "final assertion … a body that cannot fill every node must never become a meta", and the downstream consequence it names is the one that condemns valid blocks: `GetParentTxHashes` returning nil with no error reads as "transaction not found". A check whose whole job is to be the last line of defence should not be the one place where a rejection is downgraded to a log line, and dropping `Serialize`'s verdict is now the only thing between an unset entry and a stored meta.

**Fix:** Keep both assertions — the body check for the source fall-through, plus the meta check that was removed — or, cheaper, make `storeRegeneratedMeta` report a serialize failure so `buildAndStoreMeta` can fail the source and let `RegenerateMeta` move to the next one:

  ```go
  metaBytes, err := meta.Serialize()
  if err != nil {
      return errors.NewProcessingError("[storeRegeneratedMeta][%s] meta is incomplete", subtreeHash.String(), err)
  }
  ```

  A `Set` failure should stay a warning as it is today; it is only the serialize refusal that says the meta itself is unusable.

## Recap

|---|---|---|---|
| ChiR1 | Poisoned subtree_data served to peers | YES | 75% |
| ChiR2 | Mmap detach has no test | | 40% |
| ChiR3 | Unsynchronised Nodes index read | | 30% |
| ChiR4 | Rejected meta still returned | | 20% |

freemans13:
All four are fixed, on `75fbcdf65`. Two of them turned out to be partly wrong in ways worth knowing about, so read R2 and R3 below rather than taking them as applied-as-written.

## R1: poisoned `subtree_data` served outward. Fixed in `d18194d15`.

Verified independently, all of it. `GetSubtreeData.go:85-104` really does `Exists(hash, FileTypeSubtreeData)` and then streams `GetIoReader` straight back with no validation of the body, on the `GET /api/v1/subtree_data/:hash` route registered at `http.go:347`, which is the route this regenerator fetches from. `check_block_subtrees.go:309-315` takes the stored-file branch on `Exists`, `:830-832` is the count check with no re-fetch, and `:886` really is a plain `pw.Close()` on a clean parse, so a short body commits before `:902-904` trips.

Took the repair variant, not the delete one. `tryLocal` now reports separately that a `subtree_data` file exists and cannot satisfy its subtree, and says so once at Warn naming the subtree as you asked. When a peer yields a body that does satisfy it, `repairLocalSubtreeData` writes that body back over the local file with `WithAllowOverwrite(true)` and `WithDeleteAt(getBlockHeight() + blockHeightRetention)`, the same DAH expression `storeRegeneratedMeta` uses, so the repaired body expires with the meta built from it.

Three things I did differently from your sketch:

Both poisoned shapes count, not just the short one. A body truncated mid-transaction fails inside `NewSubtreeDataFromReader` rather than reaching `MissingSubtreeDataTxs`, and it is on disk just the same, so it is served outward just the same. `tryLocal` treats "the file opened and did not yield a usable `Data`" as poisoned regardless of which of the two failed.

An absent local file is deliberately not written back. Writing one is not a repair: it is not an outward poison at all, since the asset service regenerates it on demand from the subtree, and writing it would resurrect a file retention may have deleted and pay a full-body write on the ordinary missing-meta path.

`repairLocalSubtreeData` declines when `data.Subtree` is nil or has no nodes. `Data.Serialize` indexes `Subtree.Nodes[0]` before it validates anything, so those shapes panic rather than returning an error, and this runs in an errgroup goroutine with no recover.

Tests, each broken and watched failing:

- `_PoisonedLocalData_IsRepairedFromThePeerBody` asserts the store, not the call: it reads the local `subtree_data` back through `GetIoReader` and requires `MissingSubtreeDataTxs` to be zero on it, plus the `AllowOverwrite` flag and the DAH on the write. Removing `r.repairLocalSubtreeData(ctx, subtreeHash, data)` fails it with the bytes diff `([]uint8) (len=170)` expected against `([]uint8) (len=85)` actual, the truncated body still sitting there.
- `_AbsentLocalData_IsNotWrittenBack` fails if the repair is made unconditional.
- `_RepairFailure_DoesNotFailRegeneration` fails with `PROCESSING (4): ... subtreedata not available locally or from peers -> STORAGE_ERROR (69): store is read only` if a failed repair write is propagated instead of warned.

The mock store now routes a `subtree_data` `Set` back into the map `GetIoReader` reads from, the way a real blob store does. Without that the repair would be unobservable and the test could only assert that a write happened.

Your two consequences (a) and (b) both still stand and are now stated in the PR description. The cache-bust still cannot break a stall whose source is a peer's on-disk file, and `check_block_subtrees.go` still has no re-fetch. Neither is on a path this PR changes.

## R2: nothing pinned the mmap detach. Fixed in `ea1b21f39`.

Correct: `_ReloadClosesSurvivingSubtrees` asserted only the backing file, `_ReloadDoesNotWriteIntoACallerOwnedArray` only pointer identity, and `_ClosesMmapAndNilsEntries` exercises `releaseSubtreeNodesLocked`, which detaches on its own account.

Your assertion does detect the regression, but not the way you expected, and the file's own existing comment predicted it. `require.Nil` formats a failure as `Expected nil, but got: %#v`, which walks the elements. With `_ = st.ReleaseNodes()` removed it does not report an assertion at all. It dies:

```
unexpected fault address 0x112fa4000
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x2 addr=0x112fa4000]
...
fmt.(*pp).printValue(...)  /src/fmt/print.go:770
github.com/stretchr/testify/assert.Nil(...)
```

So "Nodes is only read for nil-ness here, never dereferenced" holds for the success path and not for the failure path, which is the one that matters. Same trap `_ReloadDoesNotWriteIntoACallerOwnedArray` already records for `require.Same`. Committed it as `require.True(t, mmapSurvivor.Nodes == nil, ...)`, which hands testify only the bool. The same mutation then gives a clean `Should be true: a closed mmap subtree must not keep a slice into the region Close just unmapped`.

## R3: `Nodes` indexed outside the lock. Fixed in `e2d9acce3`, and your fix is not sufficient.

`ReleaseNodes` confirmed at go-subtree `subtree.go:855-864`: it nils `st.Nodes` with no mutex, while `Length()` at `:253-259` takes `st.mu.RLock()`. So the two reads can disagree exactly as you describe. Applied the snapshot at both sites.

But it does not make the access safe, and I could not commit a test for it. A test racing the predicate against `ReleaseNodes` over 50k rounds does distinguish the fix: reverting to `Length()`-then-index fails it with `panic: runtime error: index out of range [0] with length 0` on every run. It cannot be committed because `-race` rejects it, and for a real reason rather than a test artifact:

```
WARNING: DATA RACE
Write at 0x00c00067c508 by goroutine 58:
  go-subtree.(*Subtree).ReleaseNodes()  subtree.go:861
Previous read at 0x00c00067c508 by goroutine 57:
  model.MissingSubtreeDataTxs()  peer_subtree_data.go:74
```

One read of a slice header cannot tear on any architecture Go supports, so this is benign in a way the race detector cannot express, but it is still an unsynchronised read of a field another goroutine writes. Closing it properly means holding `subtreeSlicesMu` across the regenerator's use of the subtree, or `ReleaseNodes` taking the subtree's own mutex upstream. Both are wider than this PR. So change 10 is verified by mutation and documented as such rather than shipped with a test that would fail CI.

Second finding from the same probe, which bears on how much your fix buys at the regenerator site. A test racing `buildMetaFromSubtreeData` against `ReleaseNodes` still dies **with** the snapshot applied, at go-subtree's `Subtree.NodeIndex` (`subtree.go:561-580`) reached through `Meta.SetTxInpointsFromTx`:

```
panic: runtime error: invalid memory address or nil pointer dereference
go-subtree.(*Subtree).NodeIndex()  subtree.go:567
go-subtree.(*Meta).SetTxInpointsFromTx()  subtree_meta.go:128
model.(*SubtreeMetaRegenerator).buildMetaFromSubtreeData()  subtree_meta_regenerator.go:394
```

`NodeIndex` reads `st.nodeIndex` outside the lock, builds it under the lock, then reads the built map outside the lock again. So the snapshot at that site removes one of two hazards rather than making the function safe. It still goes in, since it costs nothing, but the function is not safe against a concurrent release and the comment does not claim it is. Worth an issue against go-subtree.

## R4: a meta its own serializer rejects. Fixed in `75fbcdf65`.

Every link verified. `NewTxInpointsFromTx` (`inpoints.go:88-90`) delegates to `newSizedFromInputs`, which returns a bare `TxInpoints{}` at `len(inputs) == 0` (`inpoints.go:374-378`); `NewTxInpoints()` (`inpoints.go:79-83`) returns a non-nil-but-empty slice with the doc comment saying why; `Meta.Serialize` (`subtree_meta.go:163-176`) refuses `i != 0 && ParentTxHashes == nil` and exempts index 0 unconditionally; `SetTxInpointsFromTx` (`:127-141`) writes at `Subtree.NodeIndex(*tx.TxIDChainHash())`.

Took both of your options, because neither covers the other and a test proves it. `buildMetaFromSubtreeData` walks the built meta again alongside the body check, exempting index 0 only for a genuine coinbase placeholder; and `storeRegeneratedMeta` returns the serialize refusal rather than logging it, serializing before the store check because the serialize is the assertion rather than a step towards the write. A `Set` failure stays a warning.

The subtests separate them. With the meta walk removed, `_MetaThatWillNotSerialize_FailsTheSource/unset entry at index 0 of a subtree with no coinbase placeholder` fails with `An error is expected but got nil` while `unset entry past index 0` still passes on the serialize refusal alone. That is the index-0 gap you named, reproduced. `_StoreRegeneratedMeta_NilStore` fails with `a meta that will not serialize must be reported even when there is no store to write it to` when the serialize is moved back behind the store check, and `_StoreFailure_StillReturnsTheMeta` fails if a `Set` failure is promoted to a returned error.

Consumers, since this turns a log into a returned error. `storeRegeneratedMeta` has one caller, `buildAndStoreMeta`. `buildAndStoreMeta` has two, `tryLocal` and `tryPeer`, and both return the error to `RegenerateMeta`, which moves to the next source. `RegenerateMeta` has one production caller, `validateSubtree` at `Block.go:1443`, which already treats its error as "error getting subtree meta slice". No currently valid block becomes rejected: an unset entry makes `GetParentTxHashes` return nil, which `Block.go:1632` already turns into "transaction could not be found in tx meta data" and a `BlockInvalidError`. Failing the source is strictly better than that, because another source now gets tried. Agreed on treating it as defence in depth; I found no reachable path either.

## What I could not verify

The `-race` limit on R3 above, restated because it is the one open thing: the snapshot removes the crash, and I have no committed test proving it, only a mutation I ran locally. The two go-subtree hazards (`ReleaseNodes` without the mutex, `NodeIndex` reading its map outside the lock) are unfixed and unfixable from this repo.

## Aggregate deadline

Thanks for checking that independently. One peer or none at all four call sites, plus `validOrderAndBlessed` cancelling siblings on first error. Not reopening it.

Local: `make lint` 0 issues, `go vet ./model/...` clean, `go build ./...` clean, `go test ./model/...` green, `go test -race ./model/...` green in 16.9s, `go test ./services/blockvalidation/` green in 284.9s.

freemans13:
Heads-up on CI, unrelated to the review above.

PR #1486 merged to `main` at 12:48 BST and rewrote the same three files this PR touches: `model/Block.go`, `model/subtree_meta_regenerator.go` and its test (+175 and +481 lines respectively). #1636 is now `CONFLICTING`, so GitHub cannot build the merge ref and no Actions workflow has started on `75fbcdf65`. The only checks on the head are the two Socket Security ones. That is why this is not green rather than anything in the four fixes.

I stopped short of resolving it, because it is not a textual clash. #1486 independently fixed the same defect as R4 and fixed it better: `storeRegeneratedMeta` there returns `(bool, error)` with the serialize refusal returned rather than logged, and the meta walk gates the node-0 exemption on a new `isFirstSubtree` parameter threaded through `RegenerateMeta`, `buildAndStoreMeta` and `buildMetaFromSubtreeData`, so it agrees with `validateSubtree` instead of guessing from the placeholder hash. That is a public interface change and it is already on `main`. #1486 also implemented the source fall-through independently, as a `tryBuild` closure, which is this PR change 1 by another route, and it keeps per-source `Warnf` where change 8 moved to Debug plus one aggregate Warn.

So reconciling means deciding which of two overlapping designs survives across roughly 700 conflicted lines, on the consensus path, in code Chi reviewed 30 minutes ago. That is a design call rather than a merge, so I have left the branch exactly as pushed rather than picking one.

R1, R2 and R3 are orthogonal to #1486 and apply unchanged: nothing upstream repairs the poisoned local `subtree_data`, nothing pins the mmap detach, and upstream still reads `subtree.Length()` and then indexes `subtree.Nodes[0]`. R4 is the one that should probably be dropped in favour of the upstream version.

Locally on `75fbcdf65`, pre-merge: `make lint` 0 issues, `go vet` clean over model, blockvalidation, blockassembly, settings and util, `go build ./...` clean, and green under `-count=1`: `model` 3.2s, `model` under `-race` 16.9s, `blockvalidation` 284.9s, `blockassembly` 304.9s, `subtreevalidation` 36.1s, `asset` and its subpackages, `settings`.

Copilot refused again on its quota limit (review 5040548748, on this head), which is the eleventh time.

freemans13:
@ctnguyen this branch has been rebuilt on top of main, so the code you reviewed is not the code that is here now. Three of your four findings survive as you wrote them. The fourth is dropped, because #1486 landed a better fix for it while we were reviewing.

**What changed and why.** #1486 landed its own source fall-through in the regenerator, reaching the same conclusion independently. Rather than hand-merge two versions of the same fix, I merged main in and took main's design on every point of overlap: its `tryBuild` closure, its local-then-peers loop, its `isFirstSubtree` parameter, its `storeRegeneratedMeta` returning `(cached, error)`, its completeness checks and its per-source logging. Then I re-applied only what this branch genuinely adds. The commits after the merge are one per re-applied piece, so you can read them separately.

**R1, the poisoned local file, survives unchanged.** All three of the deviations from your sketch are still there and still deliberate: both poisoned shapes count, including a body truncated mid-transaction that fails inside `NewSubtreeDataFromReader` and never reaches `MissingSubtreeDataTxs`; an absent file is not written back, because the asset service regenerates it on demand and it is not an outward poison; and the repair declines on a zero-node subtree, because `Data.Serialize` indexes `Nodes[0]` before validating anything. It is re-homed onto main's structure rather than this branch's `tryLocal`/`tryPeer` split, which is gone.

**R2, the test assertion, survives, but not in the form you suggested.** `require.Nil` does not work here and it is worth knowing why. It formats a failure as `Expected nil, but got: %#v`, which walks the slice elements. The slice points into a region `Close` has just unmapped, so with the fix reverted the assertion dies in `fmt.(*pp).printValue` with `unexpected fault address` and `fatal error: fault`, killing the test binary instead of reporting anything. I tried it that way first and watched it happen. `require.True(t, mmapSurvivor.Nodes == nil, ...)` gives a clean "Should be true" on the same mutation, which is what is committed.

**R3, the two reads of `Nodes`, survives, and I want to be precise about what it buys.** It removes the panic: one read of the slice header now feeds both the length guard and the index, so a `ReleaseNodes` landing between them can no longer turn the index into an out-of-range on a zero-length slice. It does **not** make the read synchronised. A test racing the predicate against `ReleaseNodes` over 50k rounds does distinguish the fix, failing every run with `index out of range [0] with length 0`, but `-race` rejects that test for a real reason: it is an unsynchronised read of a field another goroutine writes. One read of a slice header cannot tear on any architecture Go supports, so this is benign in a way the race detector cannot express, but the fix is a mitigation rather than a cure. Closing it properly means `ReleaseNodes` taking the subtree's own mutex in go-subtree. I have said so in the PR body and in the commit, and I have reported honestly that no committed test distinguishes this one.

Second thing from that investigation, which is more your area than mine: go-subtree's `Subtree.NodeIndex`, reached through `Meta.SetTxInpointsFromTx`, reads `st.nodeIndex` outside the lock, builds it under the lock, then reads the built map outside the lock again. A test racing `buildMetaFromSubtreeData` against `ReleaseNodes` dies there even with your R3 fix applied. That is a race in the dependency, not in this PR, and it deserves its own issue against go-subtree.

**R4, the meta that will not serialize, is dropped.** #1486's fix is better and I would rather have theirs. You found that `Meta.Serialize` exempts index 0 unconditionally, so an unset entry at index 0 of a subtree with no coinbase placeholder would serialize and get persisted. This branch fixed the symptom, adding a walk over the built meta that exempted index 0 only for a genuine placeholder. Main fixes the cause: it threads `isFirstSubtree` through the interface, so the exemption agrees with `validateSubtree` (which skips node 0 only for `sIdx 0`) rather than being inferred from the hash, and the two predicates cannot disagree in the first place. Main's `buildAndStoreMeta` also already returns the serialize refusal rather than logging it and handing the meta back, which was the other half of what you found. Porting this branch's version on top would have been two mechanisms for one bug.

The one place your R4 reasoning still shows up is `model.MissingSubtreeDataTxs`. Its index-0 exemption is now the caller's decision instead of an inference: the regenerator passes the subtree's real position, catchup passes `true` because it is protecting `Data.Serialize` and not building a meta. I tried making catchup strict too and backed it out. It failed `TestFetchSubtreeDataForBlock/MultipleSubtrees` and `TestSubtreeFunctions/fetchAndStoreSubtreeAndSubtreeData_Success`, both of which build synthetic blocks whose every subtree carries a placeholder at node 0, and in production it would charge a peer for a body `Serialize` would have handled fine.

Everything is green: `make lint` 0 issues, `go build ./...`, `go vet` over the five affected trees, `go test` and `go test -race` on `model`, and `go test` on blockvalidation, blockvalidation/catchup, blockassembly and its subpackages, subtreevalidation, settings and util. The three fixes above were each broken in-tree and their guarding tests watched failing before being restored, except R3, which nothing can distinguish for the reason above.

ctnguyen (approved):
Review at commit [2a265d6](https://github.com/freemans13/teranode/commit/2a265d624515186f0a65778849960b7116ab18df)

Taking main's design on every overlap and then re-applying only the genuinely new parts, one commit per piece, was the right call and it makes this round readable in a way a hand-merge of two competing regenerators never would have been. All four of my previous points land, and two of them land better than I wrote them — the `require.Nil` trap in the mmap-detach assertion is a real find, and saying plainly that the `Nodes` snapshot is a mitigation rather than a cure is more useful than a confident claim would have been. The eviction work is the part I'd single out: `TryReleaseSubtreeNodes` plus `evictLastValidatedBlock` named rather than inline, and `TestLastValidatedBlocksEviction_DoesNotStallTheCache` driving the function `NewBlockValidation` actually registers — that test does the thing most eviction tests don't, which is fail if the production wiring is reverted. `MissingSubtreeDataTxs` is also better than the two copies it replaces: making the index-0 exemption an argument instead of an inference is the correct shape, and `TestMissingSubtreeDataTxs` pins both directions of it. Nothing here needs to block the merge. What I have is four notes on the newly re-applied code, and the first one is a case where the retry does the opposite of what its own comment says it does.

## Status of previous points

One clarification on what FIXED means for ChiR3, because it is narrower than "the race is closed". The snapshot landed at both sites — `model/peer_subtree_data.go:102-104` and `model/subtree_meta_regenerator.go:401-403` — and both now derive the length guard and the index from one read of the slice header, which is exactly what ChiR3 asked for and which removes the out-of-range panic from the two reads disagreeing. It does not make the read synchronised, and ChiR3 did not ask it to: `Subtree.ReleaseNodes` still nils `Nodes` without taking the subtree's mutex, so a release concurrent with either call is still a data race, now yielding a stale-but-valid slice header instead of a crash. Closing that properly means the lock moving into go-subtree's `ReleaseNodes`, which is out of this repo's reach. The code says so at both sites and the PR body says so too; FIXED here means the item as written, not the underlying race.

ChiR4 was dropped in favour of main's version, so I checked whether main's mechanism as it stands in this tree actually covers what ChiR4 identified. It does, on both halves. `storeRegeneratedMeta` now returns the serialize refusal instead of logging it (`model/subtree_meta_regenerator.go:487-490`) and `buildAndStoreMeta` propagates it (`:363-366`), which closes the "rejection downgraded to a log line" half. The index-0 half is closed by the meta walk at `:445-457`: `hasCoinbasePlaceholder` is `isFirstSubtree && nodes > 0 && nodesSlice[0].Hash.Equal(...)` (`:403`), so an unset `ParentTxHashes` at index 0 of a subtree that is not the block's first, or of a first subtree carrying no placeholder, now returns an error rather than serializing through `Meta.Serialize`'s unconditional `i != 0` exemption. That is the exact gap ChiR4 named, and it is covered by the cause rather than the symptom. Agreed it was the better fix to keep.

|---|---|---|---|
| ChiR1 | Poisoned subtree_data served to peers | YES | FIXED |
| ChiR2 | Mmap detach has no test | | FIXED |
| ChiR3 | Unsynchronised Nodes index read | | FIXED |
| ChiR4 | Rejected meta still returned | | FIXED |

## Non-blocking issues

### ChiR5 — The cache-busting retry fires on a fetch that ran out of its own budget

*Same as [@ordishs's point "the peer fetch budget" (20-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1636#pullrequestreview-4986043250).*

**Problem:** This is not the aggregate deadline, which is settled — it is the trigger for the second fetch. The retry condition is `!ok && answered` (`model/subtree_meta_regenerator.go:173`), and the comment two lines above it states the exclusion explicitly: "A refused connection or an expired budget left no body for the cache to be blamed for, and a second request would just spend the budget again before moving on" (`:170-172`). The code does not implement the "expired budget" half. `getSubtreeDataFromPeer` installs its own deadline at `:332`, then returns `answered = false` only when `util.DoHTTPRequestBodyReaderWithRetry` itself failed (`:340-343`), and `answered = true` for every failure of `NewSubtreeDataFromReader` (`:348-351`). That helper returns a body reader as soon as it has a 200 (`util/http.go:746-759`); a non-200 or a pre-header expiry comes back as an error (`:683-689`). So a deadline that fires *before* the response headers gives `answered = false` and no retry, exactly as documented — but a deadline that fires *during* the body stream surfaces as a read failure inside `NewSubtreeDataFromReader`, which is `answered = true`, and the retry goes out with a fresh full budget from the same undeadlined parent context. That is the one case where a second fetch cannot possibly help: the first attempt did not fail because the peer served a poisoned entry, it failed because 10 minutes was not enough for the body, and the retry gets 10 minutes for the same body.

**Why it matters:** The default moved from 30s to 10m in this diff, and the PR body discloses 2 × 10m as the per-peer worst case. What it does not say is that a slow peer is the *ordinary* way to reach that ceiling rather than an adversarial one — a legitimately large `subtree_data` that does not fit one budget now costs 20 minutes inline in `Block.Valid`, inside a `validOrderAndBlessed` errgroup goroutine holding a pooled parent-spends map for the whole duration, and both halves are guaranteed to fail. It also weakens the signal the retry is meant to carry: the "(cache-busted)" source label set at `:175` and printed by `tryBuild`'s `Warnf` (`:123`, `:132`) will name a peer whose problem was never a cache. The 10m sizing itself I have no argument with — the same payload from the same endpoint gets 10m on the catchup path (`services/blockvalidation/get_blocks.go:540-544`, applied at `:589`).

**Fix:** Let the fetch report whether its own budget is what ended it. The fetch context already knows, and this needs no error-chain sniffing:

  ```go
  func (r *SubtreeMetaRegenerator) getSubtreeDataFromPeer(ctx context.Context, ..., bypassCache bool) (*subtreepkg.Data, bool, error) {
      fetchCtx, cancel := context.WithTimeout(ctx, r.peerFetchTimeout)
      defer cancel()
      // ... request on fetchCtx ...
      data, err := subtreepkg.NewSubtreeDataFromReader(subtree, body)
      if err != nil {
          // A body we ran out of budget on is not evidence of a poisoned cache
          // entry, and a fresh budget cannot finish what this one could not.
          return nil, fetchCtx.Err() == nil, err
      }

      return data, true, nil
  }
  ```

  `fetchCtx.Err() != nil` is true only when this fetch's deadline fired or the parent was cancelled, which are precisely the two cases the existing comment already argues should not be retried.

### ChiR6 — Any failure to read the local subtree_data is reported as this node poisoning its peers

**Problem:** `getLocalSubtreeData` returns `localPoisoned = true` for *every* error out of `NewSubtreeDataFromReader` (`model/subtree_meta_regenerator.go:275-278`), not only for a body that cannot satisfy the subtree (`:280-282`). The file having opened is the only evidence involved; whether the failure came from the bytes on disk or from the read of them is not distinguished. That matters because two things hang off the flag. First, `RegenerateMeta` states the conclusion as fact — "the local subtree_data cannot satisfy this subtree, so this node is serving a bad body to every peer that asks for it" (`:152-154`) — which is an operator-facing accusation that a transient read failure has not earned. A store read can fail for reasons that are not the file: the regenerator is wired to whatever `u.subtreeStore` is (`services/blockvalidation/BlockValidation.go:346-349`), which includes the HTTP-backed and S3-backed implementations under `stores/blob/`, where a mid-stream failure is a network event rather than a corrupt body. Second, the same flag is the sole gate on the write-back (`:192-194`), so a healthy local file gets a full-body overwrite it did not need, with the DAH consequence in ChiR7 attached. The same line also promises "a complete peer body will be written back over it" before any peer has been contacted, and it is emitted unconditionally — including on the gRPC `ValidateBlock` path, where `createMetaRegenerator(nil)` (`services/blockvalidation/Server.go:1404`) means `r.peerURLs` is empty and no write-back can ever happen.

**Why it matters:** The whole value of ChiR1's fix is that the Warn is trustworthy — an operator who greps for it should be able to conclude their node is serving corrupt data to the fleet and act on it. A line that also fires on a blob-store hiccup, and that promises a repair on a call site that has no peers to repair from, is the kind of alarm that gets filtered out within a week, at which point the real case goes unnoticed too. The needless overwrite is the smaller half but it is not free: it re-serializes and rewrites the entire body, which for a mainnet-size subtree is a real cost paid on a file that was fine, and it is the same cost the absent-file case is deliberately kept away from by that branch returning `false` (`:266-269`).

**Fix:** Two independent narrowings, both small. Keep the classification to failures the bytes are responsible for, and move the reporting to where the outcome is known:

  ```go
  data, err := subtreepkg.NewSubtreeDataFromReader(subtree, reader)
  if err != nil {
      // A read that did not complete says nothing about what is on disk, and a
      // context error is our shutdown rather than the file's fault.
      return nil, !errors.IsContextError(err) && ctx.Err() == nil, err
  }
  ```

  and log the Warn only once there is something to say — either at the repair site (`repairLocalSubtreeData` already logs its own outcome at `:229`, `:237` and `:242`) or after the peer loop, so the "will be written back" claim is never made when `len(r.peerURLs) == 0`. If distinguishing a torn file from a failed read is not reliably possible for a given store, then say less in the line rather than more: "cannot be read or cannot satisfy this subtree" is defensible on the evidence the code actually has.

### ChiR7 — The repair restamps the file it overwrites with a finite DAH

**Problem:** `repairLocalSubtreeData` writes the replacement with `options.WithAllowOverwrite(true), options.WithDeleteAt(r.getBlockHeight() + r.blockHeightRetention)` (`model/subtree_meta_regenerator.go:234-236`). The DAH is computed unconditionally, so the repaired file carries a finite expiry whatever the file it replaced carried. This is the only `subtree_data` write in the tree that replaces a file it has just confirmed exists — every other writer either checks `Exists` and skips (`services/blockvalidation/get_blocks.go:564-567`, `services/subtreevalidation/check_block_subtrees.go:253-258`, `services/subtreevalidation/SubtreeValidation.go:1312-1320`) or deletes first (`services/blockpersister/streaming_process_subtree.go:80-82`). And there is a writer that makes such a file permanent: blockpersister promotes an existing `subtree_data` with `SetDAH(..., 0)` once its block reaches persistence (`streaming_process_subtree.go:85`), gated only on the file deserializing (`:75`) — which is exactly the check this PR's own comment says a short-but-parseable body survives (`model/subtree_meta_regenerator.go:186-189`). So a file can legitimately be both permanent and incomplete, and the repair fixes its contents while silently giving it an expiry. I could not find a path where a promoted file actually reaches the repair, because the regenerator runs during validation and revalidation while promotion happens at persistence, so treat this as defence-in-depth rather than a live bug — the same standing ChiR4 had.

**Why it matters:** The failure would be quiet and delayed: the block validates, the file looks repaired, and `blockHeightRetention` blocks later a persisted block's `subtree_data` is pruned. Nothing on the ingest path notices, and the recovery story for a persisted block's data file is thinner than for a recent one, since on-demand regeneration needs the transactions to still be retrievable. The comment above the repair call site (`:183-191`) already enumerates the other three writers and reasons about each; blockpersister's DAH promotion is the one property of that enumeration that the write itself then undoes, which is worth either handling or naming.

**Fix:** The regenerator's `SubtreeStoreWriter` is `GetIoReader` plus `Set` (`:29-38`), so it cannot read the existing DAH and cannot restore one. Either widen it enough to make the decision explicitly, or record the decision where the write is:

  ```go
  // The replacement inherits this DAH rather than the old file's, which the
  // store interface here cannot read back. Safe while every subtree_data the
  // regenerator can reach is still expiring; a file blockpersister has
  // promoted with SetDAH(.., 0) would be downgraded to an expiring one.
  dah := r.getBlockHeight() + r.blockHeightRetention
  ```

  Naming it is enough for now given I have no reachable path; the interface change is the durable answer if one turns up.

### ChiR8 — The 10m budget's stated precedent is not the code that carries it

**Problem:** `DefaultPeerFetchTimeout`'s doc comment justifies the 20× rise as "Sized to match settings.DefaultSubtreeDataFetchTimeout, which bounds the same subtree_data payload fetched from the same peer endpoint by check_block_subtrees.go" (`model/subtree_meta_regenerator.go:303-305`), and the PR description makes the same claim. `check_block_subtrees.go` does not read that setting. Its `subtree_data` fetch is `util.DoHTTPRequestBodyReaderWithRetry(fetchCtx, url)` (`services/subtreevalidation/check_block_subtrees.go:279`) on the durable top-level request context, and its own comment says the bound is "its own http_streaming_timeout" per attempt (`:274-278`). `SubtreeDataFetchTimeout` is read in exactly one place: `services/blockvalidation/get_blocks.go:540-544`, applied to `fetchAndStoreSubtreeData` at `:589`. The precedent is real, the conclusion is sound — that path is the same payload from the same endpoint under the same 10m — it is just a different file, and one with a materially different mechanism (a single whole-fetch deadline versus a fresh per-attempt streaming window).

**Why it matters:** This is the load-bearing sentence for a 20× increase to a bound that runs inline on the validation path, and it is the sentence a reviewer or an operator will follow to decide whether 10m is the right number. Pointed at `check_block_subtrees.go` they find a fetch with no such bound and a comment describing a per-attempt window, which either reads as the justification not holding up or as evidence that per-attempt windows are the house convention here — the opposite of what this constant is doing.

**Fix:** Name the fetch that carries the constant, and keep the mechanism distinction visible:

  ```go
  // Sized to match settings.DefaultSubtreeDataFetchTimeout, which bounds the
  // same subtree_data payload from the same peer endpoint for catchup's
  // fetchAndStoreSubtreeData (services/blockvalidation/get_blocks.go). Both are
  // whole-fetch deadlines covering retries plus the body stream, unlike
  // check_block_subtrees.go's fetch, which is bounded per attempt by
  // http_streaming_timeout.
  ```

  The PR description's point 2 wants the same correction.

## Recap

|---|---|---|---|
| ChiR5 | Retry fires on budget expiry | | 45% |
| ChiR6 | Read failure misreported as poison | | 35% |
| ChiR7 | Repair restamps DAH unconditionally | | 25% |
| ChiR8 | Timeout precedent names wrong file | | 15% |

icellan (approved):

PR #1639, opened 2026-08-21
Generalise settings-doc drift guard to every settings doc; delete dead kafkaWorkers settings

icellan:
## Summary

Part 1: generalises the single-file blockvalidation settings-doc drift guard
into a shared mechanism that walks every file under
`docs/references/settings/` and cross-checks each settings-table row's
documented default against the real default (`ExportMetadata()`), plus a
second check that every key used in a doc's configuration examples actually
exists. Fixes every genuine mismatch it surfaced:

- transposed `ProcessTxMetaUsingStore` batch-size/concurrency defaults and a
  misspelled example key in `blockvalidation_settings.md`
- nine struct-tag defaults in `settings/interface.go` and
  `settings/blockassembly_settings.go` that had drifted from the literal
  passed to `getBool`/`getInt`/`getDuration`/`getString` in `settings.go`
  (in every case the doc was right and the tag was stale)
- one doc bug in the other direction (`legacy_allowBlockPriority` documented
  as `false`, code default `true`)
- several misspelled/nonexistent example keys across `alert_settings.md`,
  `pruner_settings.md` and `utxo_settings.md`
- two settings-table rows and one prose section documenting features that
  were never implemented (`utxostore_prunerMaxConcurrentOperations`,
  `pruner_jobTimeout`), removed
- a factually wrong "Environment Variable Overrides" section in
  `pruner_settings.md` claiming gocore uppercases keys automatically (it
  does an exact-string `os.LookupEnv`, nothing is transformed)

Rows that are genuinely computed at runtime from another setting
(`utxostore_unminedTxRetention`, `utxostore_parentPreservationBlocks`,
`utxostore_blockHeightRetention` - the struct tag is a static snapshot
because tags can't reference the runtime constant/variable), or where one
env-var key backs two different struct fields (`blockmaxsize`, shared by
`PolicySettings.BlockMaxSize` and `BlockSettings.MaxSize` with different
defaults - a real settings.go ambiguity, flagged not fixed here since fixing
it changes key-resolution behaviour), are listed as explicit, justified
exemptions rather than papered over by loosening the comparison.

Part 2: deletes `validator_kafkaWorkers`, `blockvalidation_kafkaWorkers` and
`block_kafkaWorkers` - defined and loaded but never read anywhere outside the
settings package. Kafka consumer concurrency is actually controlled by the
`consumer_ratio` URL parameter per topic; docs that described these settings
as the real lever are corrected.

**Release note**: the settings loader (gocore) has no schema and does not
reject unknown keys - an operator with one of the three deleted keys still
set in `settings_local.conf` or deployment config gets silence, not a
startup error. The value is simply never looked up again after this change.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./settings/...`
- [x] `go test ./settings/... -race`
- [x] `golangci-lint run ./settings/...`, `staticcheck ./settings/...`
- [x] confirmed the guard bites: introduced a wrong table default and a
      fabricated example key one at a time, verified the test fails naming
      the exact file/key, reverted
- [x] `grep -rn -i kafkaworkers .` returns nothing


icellan:
Converted `docs/references/settings/services/pruner_settings.md` to the standard
`| Setting | Type | Default | Environment Variable | Usage |` table format, so it is no longer the
one file exempted from the guard this PR adds. `minCheckedRowsPerDoc` for it goes 0 -> 24; the
parser now cross-checks 27 rows (24 `Pruner` keys + 3 `UTXOStore` keys). `blob_settings.md` keeps
its 0-row exemption - it documents blob store URL query params, not settings keys.

Drift the newly-enabled coverage exposed:

- `startPruner` documented with default `true`. There is no code default: `daemon.shouldStart()`
  calls `gocore.Config().GetBool` with no fallback, so it is `false`. `settings.conf` ships `true`
  at the base level but `false` for `docker.m` and `operator` - the doc's context table and its
  per-context examples both claimed `true` for those two.
- `pruner_grpcPort` documented as an integer setting with default 8096. No such key exists;
  `PRUNER_GRPC_PORT` is a gocore template variable interpolated into
  `pruner_grpcAddress`/`pruner_grpcListenAddress`.
- `utxostore_unminedTxRetention` claimed a typical value of ~7200 blocks (~50 days). Default is
  `globalBlockHeightRetention/2` = 144.
- `utxostore_parentPreservationBlocks` claimed ~14400 blocks (~100 days) while its own worked
  example said 1440 (10 days). Default is `blocksInADayOnAverage*10` = 1440.
- `pruner_min_block_height` and `pruner_utxoPrunedSetMaxEntries` were undocumented.
- Three `settings.conf` values that differ from the Go defaults were undocumented:
  `pruner_utxoDefensiveBatchReadSize` (1024 vs 10000), `pruner_utxoChunkSize` (1024 vs 1000),
  `pruner_utxoChunkGroupLimit` (1 vs 10), plus `pruner_utxoPartitionQueries.docker.m = 8`.
- `docker.host` overrides for both address settings were documented as shipped; `settings.conf`
  has none.

Surfaced, not fixed: `pruner_skipDuringCatchup`, `pruner_force_ignore_block_persister_height` and
`pruner_skipProcessExpiredPreservations` carry `key` struct tags and are read by `services/pruner`,
but `settings.NewSettings()` never populates those fields - they always hold `false`, so the keys
have no effect. Wiring them changes runtime behaviour, so it stays out of a doc conversion; the doc
now states it under "Settings not wired to configuration". Separately, `pruner_jobTimeout` still
sits in `settings.conf` even though this PR deleted the doc row for it - dead config line.

Verified: `go test ./settings/... -race`, `go build ./...`, `go vet ./settings/...`, markdownlint
via pre-commit. Coverage proof: changing the `MinBlockHeight` row's default to 7 fails with
`pruner_settings.md documents MinBlockHeight (pruner_min_block_height) with default "7", but the
settings package defaults it to "0"`; reverted.

icellan:
Extended `TestSettingsDocDefaultsMatchCode` to a three-way check: doc default vs struct tag default vs the value `NewSettings` actually produces at runtime, built in an isolated gocore context with every env var and confs entry that could shadow a `getX(...)` fallback stripped out first.

Why the tag alone wasn't enough: this PR already found nine struct tags in `interface.go` where the tag had drifted from the literal passed to `getBool`/`getInt`/`getDuration` in `settings.go`, and in every case the doc was right and the tag was wrong. A doc-vs-tag comparison is blind to that class of bug by construction - if both sides drift the same way, they still compare equal. Anchoring to the runtime value closes that gap.

Running the runtime anchor surfaced more of the same pattern beyond those nine:

- `FetchNumWorkers` / `SubtreeFetchConcurrency` (blockvalidation) and `SpendBatcherConcurrency` (legacy): doc and tag agreed with each other but were both stale against the real `getInt(...)` literal. Fixed to match the literal.
- Six `PolicySettings` fields (`MaxOrphanTxSize`, `LimitAncestorCount`, `LimitCPFPGroupMembersCount`, `MaxStdTxValidationDuration`, `MaxNonStdTxValidationDuration`, `MaxTxChainValidationBudget`) are entirely unwired - their `getInt(...)` calls are commented out in `NewSettings`, so the field is always its Go zero value regardless of any config. Doc and tag both documented a non-zero default that was never live. Corrected both to 0, with a note pointing at the commented-out loader line.

The three `utxo_settings.md` exemptions and the `blockmaxsize` key-collision exemption carry forward with their original justifications, now expressed as which of the three anchors is unreliable for that row instead of a blanket skip - the utxo rows document a runtime formula, not its resolved number, so they still can't be text-compared, but that's now explicit rather than just "skip everything". Four rows with a genuinely CPU-count-dependent default (`max(4, NumCPU()/2)`) are newly exempted from the runtime comparison only; doc and tag still have to agree on the formula.

A failing row now names exactly which pair(s) disagree and prints all three values, e.g.:

```
blockvalidation_settings.md documents FetchNumWorkers (blockvalidation_fetch_num_workers): [doc/tag doc/runtime] disagree - doc default "1", struct tag default "1", runtime default "16"
```

icellan:
Two additions on top of the doc-drift sweep:

**Dead config removal**: `pruner_jobTimeout` is not read anywhere in Go (`getInt`/`getDuration` on that key doesn't exist), so removed the settings.conf line plus its now-orphaned mentions in `docs/references/services/pruner_reference.md` and `docs/topics/services/pruner.md`. Also deleted `stores/utxo/aerospike/pruner/README.md` - it documented the `JobManager`/`Options{WorkerCount, MaxJobsHistory}`/`GetJobs()`/`UpdateBlockHeight()` API that the partition-based pruner rewrite removed wholesale; nothing in it was accurate anymore.

**key: tag uniqueness guard**: added `TestSettingsKeyTagsAreUnique` in `settings/export_test.go`. It walks the `Settings` struct and fails, naming both dotted field paths, whenever two fields share a `key:` tag - the exact bug class that let two fields silently clobber each other in `ExportMetadata()`'s flat map.

Running it surfaced two collisions, both exempted but categorised differently so it's obvious which is which:

- `blockmaxsize` - **accidental, a bug**. `PolicySettings.BlockMaxSize` (default 0, controls mining) and `BlockSettings.MaxSize` (default 4294967296, controls acceptance) are different settings that happen to share a key. **This exemption must be deleted once #1640 (`fix/remove-dead-block-maxsize-setting`) lands** - that PR removes `BlockSettings.MaxSize`.
- `blockvalidation_processTxMetaUsingStore_BatchSize` - **intentional, permanent**. `BlockSettings.ProcessTxMetaUsingStoreBatchSize` is read by blockpersister, `BlockValidationSettings.ProcessTxMetaUsingStoreBatchSize` by subtreevalidation/asset; both resolve from the same `getInt` call so they never diverge, and the sharing is already documented as `**SHARED**` in `blockpersister_settings.md`. Not touching this - collapsing it into one field (à la `AdaptiveFetch`) is a separate, deliberate refactor if anyone wants it.

`go build ./...`, `go vet ./settings/...`, `go test ./settings/... -race` all green.

ctnguyen (changes requested):
Review at commit [d6a8375](https://github.com/icellan/teranode/commit/d6a83750c7c01297a39495a337914497e6ede447)

The three-way anchor is the right instinct and the part of this PR I'd keep even if everything else changed — a doc-vs-tag check really is blind by construction to the nine tags that had drifted. The isolation machinery holds up too; I opened gocore v1.0.82 under `$GOPATH/pkg/mod` and read it rather than inferring: `Config(ctx)` allocates a fresh map and copies `confs` entry by entry (config.go:449-452, inside the alternative-context branch at 442-458), and `getInternal` consults `os.LookupEnv(key)` before touching `confs` (config.go:620-627). So the two things worth isolating are the two things isolated, and `GetAll()` returns a fresh map, meaning the `Unset` loop is not mutating what it iterates. The four `runtime`-unreliable exemptions map one-for-one onto the four `max(4, runtime.NumCPU()/2)` call sites in `settings/settings.go` (343, 348, 372, 612), and those are the only machine-dependent defaults in the package — grepping `settings.go` and `helpers.go` for `runtime.`, `os.Hostname`, `os.Getenv`, `GOMAXPROCS` and `time.Now` turns up nothing else. The retagging carries no behavioural risk: `ExportMetadata()`'s only non-test caller in the repo is the read-only handler at `services/asset/httpimpl/settings_handler.go:75`. `git grep -in kafkaworkers` over tracked content at this commit returns nothing at all, so no tracked file still mentions the deleted keys. Every `settings.conf` claim in the pruner conversion checks out against `settings.conf:566-642` and `:1121-1126`, and the rewritten "Environment Variable Overrides" section is now actually correct about gocore's exact-string lookup. `go vet -mod=mod ./settings/...` is clean here. I did not run `go test`, so everything below is reasoned from reading the guard, not from executing it — and where I land differently from the PR description is on how far it can be trusted: it validates the rows it can resolve and silently walks past the rows it cannot, which is the largest class of doc drift, with seventeen live instances still in the tree.

## Blocking issues

### ChiR1 — A doc table row naming a key that does not exist is silently skipped, not failed

**Problem:** `settings/settings_doc_test.go:530-538` drops any row whose `Environment Variable` cell is not a known key:

```go
codeDefault, ok := codeDefaults[row.key]
if !ok {
    continue
}
```

Nothing else covers that case. `TestSettingsDocExampleKeysExist` only inspects `key = value` assignment lines (`exampleKeyRe`, line 643), never table cells. So a settings-table row documenting a misspelled, renamed or fabricated key passes both tests. That is precisely the drift this PR fixed by hand — `pruner_grpcPort` and `utxostore_prunerMaxConcurrentOperations` were both nonexistent keys — and the guard will not stop either from coming back. Seventeen such rows survive at HEAD, each verified with a per-key `git grep`:

- `docs/references/settings/services/blockchain_settings.md:27-38` — twelve rows naming `blockchain_postgres_pool_postgres_maxOpenConns`, `..._circuitBreakerEnabled` and so on. The string `blockchain_postgres_pool_postgres` appears nowhere in Go. There are two real Postgres pool key families and the documented names belong to neither: a **per-service** family built by concatenation inside `getPostgresPoolSettings` (`settings/helpers.go:159-165`) — `blockchain_postgres_maxOpenConns` and six siblings, each defaulting to 0 so the pointer stays nil unless one is explicitly set — and a **global** family read at `settings/settings.go:71-79` (`postgres_maxOpenConns` default 50, `postgres_maxIdleConns` default 10, and five more). For a row sitting under a `blockchain_postgres_pool` heading the key an operator actually wants is the per-service one. Five of the twelve are worse than misnamed: the `circuitBreaker*` rows have no equivalent in either family, because those keys exist only as tags on `PostgresSettings` (`settings/postgres_settings.go:18-22`) and no loader ever reads them (see ChiR2).
- `docs/references/settings/stores/utxo_settings.md:48-51` — four rows for `utxostore_prunerParentUpdateBatcherSize`, `..._prunerParentUpdateBatcherDurationMillis`, `..._prunerDeleteBatcherSize`, `..._prunerDeleteBatcherDurationMillis`. `git grep -n "prunerDeleteBatcher\|prunerParentUpdateBatcher" -- '*.go'` returns nothing. These sit directly adjacent to the row this PR deleted, in a block of the file this PR edits.
- `docs/references/settings/services/subtreevalidation_settings.md:18` — `subtreevalidation_subtreeBlockHeightRetention`. The value is actually derived in `settings/interface.go:112-119` from `global_blockHeightRetention` plus `subtreevalidation_blockHeightRetentionAdjustment`.

**Why it matters:** A wrong default and a nonexistent key fail the same way for an operator — the line in `settings_local.conf` is silently ignored and the node runs on the built-in value. The guard is being merged as the thing that makes that class of bug loud, and the PR body sells it that way, but the check is one-sided: it validates the *value* of rows it recognises and says nothing about rows it does not. That also means the only backstop against a key rename is `minCheckedRowsPerDoc`, which is deliberately slack (see ChiR5). Worse, the skip path is indistinguishable between legitimate and illegitimate rows: `global_settings.md:16` (`SETTINGS_CONTEXT`), `pruner_settings.md:40` (`startPruner`) and `pruner_settings.md:49` (`pruner_IndexName`) all correctly document things that are not settings-package keys, and all three take the same silent `continue` — so the carefully written `false (no code default - ...)` cell on the `startPruner` row is never checked against anything either.

**Fix:** Fail on an unresolvable key behind an explicit exemption list, the same shape as `exampleKeyExemptions`. The rows needing an entry are exactly the ones already listed in `exampleKeyExemptions`, currently three — `SETTINGS_CONTEXT` (`settings/settings_doc_test.go:656`), `startPruner` (`:661`) and `pruner_IndexName` (`:677`). I enumerated that by extracting the `Environment Variable` cell from every settings table in `docs/references/settings/` and diffing against every `key:"..."` tag in `settings/*.go`; the two column layouts in use are the standard five-column one and `kafka_settings.md`'s four-column variant, and every key in the kafka tables resolves.

```go
codeDefault, ok := codeDefaults[row.key]
if !ok {
    if _, exempt := lookupNonSettingsKey(base, row.key); !exempt {
        rowFailures = append(rowFailures, fmt.Sprintf(
            "%s (%s) is documented as a settings key, but no such key exists", row.setting, row.key))
    }
    continue
}
```

Then correct the seventeen rows. For the Postgres block that means documenting both families in the place each belongs — the per-service `blockchain_postgres_*` and `utxostore_postgres_*` keys under their own services, and the global `postgres_*` fallback family under Global Settings, where it is currently absent entirely, which is why `utxo_settings.md:53` points at a section that has nothing to point at.

## Non-blocking issues

### ChiR2 — An unwired setting is invisible when its intended default is the zero value, and the tag-vs-runtime check that would catch the rest only runs on documented rows

**Problem:** The runtime anchor catches an unwired field only when some anchor claims a non-zero default. When the intended value *is* the zero value, doc, tag and runtime all read zero and compare equal, so the row passes while the setting does nothing. This PR's own headline correction is that case: the six `PolicySettings` loaders are commented out (`settings/settings.go:105`, `114`, `115`, `120`, `121`, `122`), and the fix was to move doc and tag down to `0` to match — which is right as documentation, but it also means all three anchors now agree on a number that is not a default but an accident, and the guard will defend it indefinitely. The seventh commented-out loader, `settings/settings.go:123` (`// ValidationClockCPU: getBool("validationclockcpu", false, ...)`), got no annotation at all, and it is invisible for exactly this reason: its intended value is `false`, so `docs/references/settings/policy_settings.md:47` presents it as a live default and every comparison passes.

The gap is wider than the zero-value case, though, because the guard only ever compares tag against runtime for keys that appear in a doc table (`settings/settings_doc_test.go:529-566`, inside the per-row loop). Both anchors are available for every key without any doc at all. The `PostgresSettings` circuit breaker is the worked example: `settings/postgres_settings.go:18-22` tags five keys with non-zero defaults (`postgres_circuitBreakerFailureThreshold` `5`, `...HalfOpenMax` `3`, `...Cooldown` `30s`, `...FailureWindow` `10s`, `...Enabled` `false`), and no loader sets any of them — neither the global `Postgres:` literal at `settings/settings.go:71-79` nor `getPostgresPoolSettings` at `settings/helpers.go:158-181`, both of which populate the same seven non-circuit-breaker fields. So `tSettings.Postgres.CircuitBreakerEnabled` is permanently `false`, the global fallback at `util/sql.go:132` can never fire, and the block guarded by `if poolSettings.CircuitBreakerEnabled` at `util/sql.go:163` is unreachable. A tag-versus-runtime comparison would have flagged four of those five immediately — `5` against `0`, `30s` against `0s` — but it never runs, because the only rows that document them are the nonexistent composite names in ChiR1.

**Why it matters:** "This setting looks configurable and is not" is the most valuable class of finding in this PR, and the note added under "Settings not wired to configuration" is the right pattern for reporting it. But it was found by hand, so the coverage is whatever the author happened to notice, and a reader will reasonably infer the list is exhaustive. Meanwhile a documented Postgres resilience control that cannot be switched on is the sort of thing an operator discovers during an outage. I am deliberately not putting a number on how many keys are in this state: the obvious method — checking whether a tagged key appears as a literal in `settings.go` — is unsound, because `getPostgresPoolSettings` builds its keys by concatenation and so `blockchain_postgres_pool` (`settings/blockchain_settings.go:20`) and `utxostore_postgres_pool` (`settings/utxostore_settings.go:52`) are correctly wired at `settings/settings.go:327` and `:484` despite never appearing as literals. Every example above is individually verified; the population size is not.

**Fix:** Run the tag-versus-runtime comparison over every entry in `ExportMetadata()`, independent of whether a doc mentions it. That is the sound version of the check — it observes what `NewSettings` actually produced rather than pattern-matching source text, so helper-built keys are handled for free. The exemptions it needs are ones this file already carries: the four `runtime.NumCPU()` keys and the `blockmaxsize` collision. Expect it to surface more than the circuit breaker family on first run — `p2p_peer_registry_ttl` is tagged `24h` and resolves to `0s`, for one — so land it with the triage that implies rather than as a silent bulk exemption. Separately, annotate `validationclockcpu` for parity with its six siblings.

### ChiR3 — Three of the four `both`-unreliable exemptions are avoidable, and this PR proves it

**Problem:** `settings/settings_doc_test.go:302-328` marks `utxostore_unminedTxRetention`, `utxostore_parentPreservationBlocks` and `utxostore_blockHeightRetention` as `both`, disabling all three comparisons, on the grounds that the doc states a formula that "cannot be reduced to the same comparable literal without a formula evaluator this test does not have". But the same three settings are documented twice. `docs/references/settings/stores/utxo_settings.md:10-12` writes the bare formulas (`globalBlockHeightRetention/2`, `blocksInADayOnAverage*10`). `docs/references/settings/services/pruner_settings.md:61-62`, added by this PR, writes `144 (globalBlockHeightRetention/2, ...)` and `1440 (blocksInADayOnAverage*10)` — resolved number plus formula in a parenthetical, which `canonicalDefault`'s trailing-paren strip reduces to `144` and `1440` and which therefore passes the full three-way check.

**Why it matters:** `both` is the strongest exemption in the file — it turns the row off entirely — and it is being spent on a formatting choice rather than a real limitation. The result is two documentations of the same key with opposite guard coverage: change `utxostore_parentPreservationBlocks`'s tag to 9999 and the pruner doc fails while the utxo doc, which is the canonical reference for that setting, stays silent. All three values happen to agree with reality today (144/1440/288), so nothing is currently wrong — this is guard strength, not a doc bug.

**Fix:** Restate `utxo_settings.md:10-12` in the `pruner_settings.md` style and delete the three exemptions. That leaves `blockmaxsize` as the only `both` entry, which is a genuine key collision rather than a formatting artifact.

### ChiR4 — The example-key regex cannot see an indented code block

**Problem:** `exampleKeyRe` at `settings/settings_doc_test.go:643` is anchored `(?m)^`, so the key must sit at column 0. A fenced block nested inside a numbered list is indented four spaces and is therefore never scanned. There are three such lines in the file this PR converted: `docs/references/settings/services/pruner_settings.md:469`, `:470` and `:484`. All three happen to name real keys, so nothing is broken — they are simply outside the check.

**Why it matters:** "Every key used in a doc's configuration examples exists" is one of the two guarantees this PR advertises, and the troubleshooting sections — where indented list-nested snippets naturally occur — are the examples an operator is most likely to copy verbatim under pressure. The same anchoring also means the regex has no notion of fence language, which is why `PreserveUntil` needed a hand-written exemption at `settings/settings_doc_test.go:672-675`: any future Go, SQL or shell snippet containing an assignment will demand another one.

**Fix:** Drop the column-0 anchor in favour of leading-whitespace tolerance, and restrict matching to inside ```` ```conf ````/```` ```bash ```` fences so the language-mismatch exemptions can go away:

```go
var exampleKeyRe = regexp.MustCompile(`(?m)^[ \t]*([A-Za-z][A-Za-z0-9_]*)((?:\.[A-Za-z0-9_]+)*)[ \t]*=[^=]`)
```

### ChiR5 — The row-count floors are slack enough to hide a silent parse failure, which is the failure mode they exist to catch

**Problem:** `minCheckedRowsPerDoc` (`settings/settings_doc_test.go:613-636`) is deliberately set "a little below the row count actually produced". For `pruner_settings.md` the comment on commit `567e077` puts the parseable count at 27 and the floor at 24, so three rows can stop being checked without a failure. The parser is easy to trip into exactly that: any non-`|` line resets `inTable` (`settings/settings_doc_test.go:133-136`), so one prose line interposed in a table truncates it and every row below is re-read as a header row and dropped. That is not hypothetical — it is the bug this PR fixes at `docs/references/settings/stores/utxo_settings.md:53`, where a `**Note**` line sat mid-table and cut off the rows beneath it. Exempted rows also increment `checked` (`settings/settings_doc_test.go:540-545`), so a row whose comparisons are all skipped still counts toward the floor.

**Why it matters:** The floor is the only thing standing between a reformat and the test quietly becoming a partial no-op, and a three-row margin is roughly the size of the truncation this PR just repaired.

**Fix:** Make the floors exact and require an update when a row is added or removed — `require.Equal` rather than `require.GreaterOrEqual`. The churn is one line per doc edit and is the same discipline the exemption staleness checks already impose. Failing that, exclude exempted rows from `checked` so the floor measures rows actually compared.

### ChiR6 — Deleting the `pruner_jobTimeout` doc row left the config line that sets it

**Problem:** This PR removed the documentation for `pruner_jobTimeout` on the correct grounds that the feature does not exist — `git grep -in jobTimeout` over tracked Go files returns nothing. The committed config that sets it, and describes its behaviour in comments, is untouched:

    settings.conf:601:  # Timeout for waiting for pruner job completion before coordinator moves on
    settings.conf:602:  # Large pruning operations (multi-million rows) may exceed this timeout, which is normal
    settings.conf:603:  # The pruner continues in background and will be re-queued immediately
    settings.conf:604:  # Default: 10m
    settings.conf:605:  pruner_jobTimeout = 10m

**Why it matters:** This is the wrong half to remove. Before this PR an operator reading either the doc or the config file got the same (wrong) story; after it, the doc is silent and `settings.conf` is the only remaining description of a feature the PR has just established does not exist — four lines of comments asserting background re-queue behaviour that no code implements. Deleting the doc row without the config line converts a consistent error into an inconsistent one, and moves the surviving copy to the place nothing in CI looks at. I checked the other row this PR deleted the same way: `git grep -in prunerMaxConcurrentOperations` returns nothing anywhere, so `pruner_jobTimeout` is the only sibling in this state.

**Fix:** Delete `settings.conf:601-605` in this PR. It is five lines and squarely in scope for a change whose part 2 is deleting dead settings. As a smaller follow-up, the same file could grow a companion test that parses `settings.conf`, strips context suffixes and asserts every base key resolves — which is what would make the release note's "operator gets silence" a one-time migration note rather than a standing property.

### ChiR7 — There is a second duplicate `key:` tag, unexempted and unflagged

**Problem:** The `blockmaxsize` exemption (`settings/settings_doc_test.go:330-340`) correctly diagnoses a key used by two struct fields and correctly identifies which one wins the flat map (`Policy` is field 63 of `Settings`, `Block` is field 68, so `BlockSettings.MaxSize` is visited last — confirmed). It is not the only one. `blockvalidation_processTxMetaUsingStore_BatchSize` is declared twice, at `settings/block_settings.go:20` and `settings/blockvalidation_settings.go:36`. Both tags say `1024` and both `getInt` calls (`settings/settings.go:249` and `:342`) pass `1024`, so the collision is currently invisible and harmless. Nothing in the guard would tell you if that stopped being true.

**Why it matters:** The `blockmaxsize` exemption text calls this shape "a genuine settings.go ambiguity worth fixing separately". If it is worth fixing, it is worth detecting, and detecting it is three lines. Right now the second instance is only discoverable by doing what I did — `grep -roh 'key:"[A-Za-z0-9_]*"' settings/*.go | sort | uniq -d`.

**Fix:** Assert in this test that `ExportMetadata().Settings` has no duplicate `Key`, with `blockmaxsize` and `blockvalidation_processTxMetaUsingStore_BatchSize` as named, reasoned exemptions until the fields are reconciled. That also protects the Settings portal, which currently serves both entries for a colliding key.

### ChiR8 — The `settings.conf ships N` parentheticals are stripped before comparison and never verified

**Problem:** `canonicalDefault`'s trailing-paren strip (`settings/settings_doc_test.go:192`, `228-230`) discards the annotation wholesale, so the three claims this PR adds at `docs/references/settings/services/pruner_settings.md:19-21` — `10000 (Go default; settings.conf ships 1024)`, `1000 (... ships 1024)`, `10 (... ships 1)` — are free text as far as the guard is concerned. I checked them against `settings.conf:614`, `:618` and `:619` and all three are correct today.

**Why it matters:** These annotations are the most operationally useful cells in the table and the ones most likely to rot, because `settings.conf` changes on a different cadence than the Go defaults. `settings.conf` is committed, so unlike a formula this is mechanically checkable — which makes it the one place where stripping the parenthetical costs something.

**Fix:** Recognise the `(Go default; settings.conf ships X)` form specifically and assert `X` against the base-context value parsed out of `settings.conf`, leaving all other parentheticals stripped as they are now.

## Recap

| ChiR1 | Doc table keys never validated | YES | 78% |
| ChiR2 | Unwired settings stay invisible | | 58% |
| ChiR3 | Avoidable utxo doc exemptions | | 55% |
| ChiR4 | Indented examples not scanned | | 30% |
| ChiR5 | Row floors hide silent drops | | 25% |
| ChiR6 | Dead pruner jobTimeout config block | | 22% |
| ChiR7 | Second duplicate key unflagged | | 20% |
| ChiR8 | Config annotations unverified | | 12% |

icellan:
Pushed `0536970f9` (rebased onto `origin/main`, was previously DIRTY/conflicting — resolved, see below). Disposition of ChiR1-ChiR8:

**ChiR1 (blocking)** — fixed. `TestSettingsDocDefaultsMatchCode` now fails on a doc-table row whose key doesn't resolve against `ExportMetadata()`, behind an explicit `nonSettingsKeyExemptions` list (same shape as the existing example-key exemptions). Corrected all seventeen surviving rows:
- `blockchain_settings.md` / `utxo_settings.md`: the twelve fabricated `blockchain_postgres_pool_postgres_*` names replaced with the real per-service `blockchain_postgres_*`/`utxostore_postgres_*` keys (each exempted — real config-facing keys built by string concatenation in `getPostgresPoolSettings`, invisible to `ExportMetadata()` because the pointer field's own `key:` tag makes it one opaque entry). Added a new "PostgreSQL Connection Pool" section under Global Settings documenting the `postgres_*` family these fall back to, including a note that the five circuit-breaker keys are currently unwired (see ChiR2).
- `utxo_settings.md`: deleted the four fabricated pruner-batcher rows (`PrunerParentUpdateBatcherSize/DurationMillis`, `PrunerDeleteBatcherSize/DurationMillis`) — no backing field anywhere in the codebase.
- `subtreevalidation_settings.md`: replaced the fabricated `SubtreeBlockHeightRetention`/`subtreevalidation_subtreeBlockHeightRetention` row with the real `BlockHeightRetentionAdjustment`/`subtreevalidation_blockHeightRetentionAdjustment`.

**ChiR2** — fixed. Added `TestSettingsTagMatchesRuntimeForAllKeys`, comparing every `ExportMetadata()` key's struct-tag default against its isolated `NewSettings()` value, independent of any doc. Also annotated the seventh unflagged commented-out `PolicySettings` loader (`ValidationClockCPU`) for parity with its six siblings.

Fixed two `formatValue`/`canonicalDefault` false positives the sweep surfaced (not real bugs): non-string slice formatting (`dashboard_devServerPorts` was rendering as `"[2 items]"` instead of its actual values) and nil-pointer-to-struct defaults written as the literal `"nil"` (`utxostore_postgres_pool`).

Real mismatches, triaged against the open PRs you named:
- `postgres_circuitBreaker{Enabled,FailureThreshold,HalfOpenMax,Cooldown,FailureWindow}` — exempted, closed by #1642.
- `p2p_peer_registry_{ttl,cleanup_interval,max_size}` — exempted, closed by #1645.
- `blockchain_peerRegistrySaveInterval`, `blockchain_subscription_timeout` — exempted, closed by #1643.

No keys found outside those five PRs' coverage — #1646/#1641 didn't surface anything in this sweep (their target keys already matched tag/runtime, or aren't yet tagged).

**ChiR3** — fixed. Restated `utxo_settings.md`'s three retention rows in the resolved-number style `pruner_settings.md` already uses (`144 (globalBlockHeightRetention/2)` etc.) and deleted the three now-unnecessary `both` exemptions. `blockmaxsize` is the only one left — genuine key collision.

**ChiR4** — fixed. `exampleKeyRe` now tolerates leading whitespace, and matching is scoped to `conf`/`bash`/`text` fenced blocks (not raw doc text), which both picks up the three previously-invisible indented lines in `pruner_settings.md` and removes the `PreserveUntil` exemption (the ```go sample it needed the exemption for no longer matches at all, since `go` isn't in the fence allowlist).

**ChiR5** — fixed. `minCheckedRowsPerDoc` floors replaced with an exact `checkedRowsPerDoc` map (`require.Equal`), and a fully `both`-exempted row no longer increments the counter.

**ChiR6** — the primary fix (dead `pruner_jobTimeout` config block + stale pruner README) was already done at `128c4d5af` on this branch, confirmed still gone at head. For the companion follow-up: I did not implement a full settings.conf-vs-`ExportMetadata()` resolver — a first pass surfaces ~68 unresolved base keys, the large majority legitimate (gocore `${VAR}` interpolation targets, the `start<App>` pattern, etc.), several genuinely worth a look (e.g. `securityLevelGRPC` in settings.conf vs the real `security_level_grpc` tag) but that's a separate audit, not something to triage safely inside this diff. Instead added a narrow regression test (`TestSettingsConfHasNoDeadKeys`) asserting the specific keys this PR already found dead (`pruner_jobTimeout`, `utxostore_prunerMaxConcurrentOperations`) stay dead. Flagging the ~68-key list to you separately rather than guessing at exemptions for it here.

**ChiR7** — already done at `128c4d5af`: `TestSettingsKeyTagsAreUnique` in `settings/export_test.go` covers this collision with a typed, reasoned exemption. Confirmed it satisfies the ask (proved it bites: introduced a fake duplicate tag, test failed correctly, reverted).

**ChiR8** — fixed. The `"N (Go default; settings.conf ships X)"` form is now recognised specifically and `X` is asserted against settings.conf's actual base-context value (comment-stripped, quote-stripped), instead of being discarded by the generic trailing-parenthetical strip.

Verification: `go build ./...`, `go vet ./settings/...` (and `./...` — pre-existing, unrelated `lostcancel`/copylocks warnings in `test/utils` only), `go test ./settings/... -race`, `golangci-lint run ./settings/...` all clean. For ChiR1/ChiR2/ChiR5/ChiR6/ChiR8 I introduced the specific defect each check targets, confirmed the failure names the right file/key, then reverted — details in the session log if useful.

ordishs (approved):
Two items to fix before merge. The guard design is right, and the twenty tag-default fixes plus the three `*_kafkaWorkers` deletions are correct and complete.

**docs/references/settings/services/blockchain_settings.md:27,36** — also `docs/references/settings/stores/utxo_settings.md:54,64` and `docs/references/settings/global_settings.md:123`

- All three PR-added sections say `PostgresPool` stays `nil` only "if none of the seven are set", and two annotate the `RetryEnabled` row "(falls back to global if unset)". Neither claim holds for `RetryEnabled`.
- `getPostgresPoolSettings` omits `retryEnabled` from the all-zero test (`settings/helpers.go:168-169`), and the global merge in `util/sql.go:110-146` gives a fallback to the other six fields but never to `RetryEnabled`.
- Case A: an operator sets only `blockchain_postgres_retryEnabled = true`. The pool stays `nil`, the global default `false` applies, retry stays off.
- Case B: an operator sets global `postgres_retryEnabled = true` plus `blockchain_postgres_maxOpenConns = 100`. The pool is now non-nil with `RetryEnabled=false`, so one pool-size knob silently disables Postgres retry resilience. The global flag alone enables it. Both cases reproduce on the `utxostore` prefix.
- Fix: reword "none of the seven" to "none of the six numeric or duration settings", and state that `RetryEnabled` has no global fallback and is ignored unless one of those six is also set. Fix all three files together. Adding `|| retryEnabled` to the nil guard would instead make the docs true, but that is a behaviour change for its own commit.

**settings/settings_doc_test.go:89-99, 167-170, 947**

Three parser blind spots let injected drift pass. A dropped row never reaches `checked++`, so the per-file exact-count assertion cannot detect the loss — the counts in `checkedRowsPerDoc` were recorded after the drops.

- `splitTableRow` splits on every `|` and ignores the markdown `\|` escape, so an escaped pipe in a cell shifts every later column. `p2p_settings.md:49` documents `p2p_websocket_trusted_source_cidrs` as `127.0.0.1/32\|::1/128`; the Default cell becomes `127.0.0.1/32\` and the key cell `::1/128`, which fails `docKeyRe`. Rewriting that documented default to `0.0.0.0/0\|999.999.999.999/1` still passes. The neighbouring row, which has no escape, fails loudly on the same drift.
- The row loop takes a bare `continue` when the key cell fails `docKeyRe`, so the row reaches neither the key-existence failure nor the three default comparisons. `blockassembly_settings.md:30` documents `DoubleSpendWindow` as `BlockTime * 6` with key `N/A`; the real key is `double_spend_window_millis` and the real default is `0`. With `N/A` the guard passes; with the real key it reports the mismatch.
- `exampleKeyRe` anchors the key at line start, so it cannot match a shell `export KEY=value` line even though `configFenceLanguages` includes `bash`. `pruner_settings.md:419-426` is a ```bash fence of four `export` lines — the exact lines whose misspelled keys this PR hand-fixed. Injecting `export utxostore_unminedTxRetentoin=10000` still passes; the same typo without the `export` prefix fails as intended.
- Fix: split on unescaped pipes only then unescape each cell; make a non-identifier key cell a row failure gated by an explicit exemption list for rows that genuinely have no key; allow an optional prefix in `exampleKeyRe` (`^[ \t]*(?:export[ \t]+)?([A-Za-z][A-Za-z0-9_]*)`). Then correct the two rows the fix exposes and bump `checkedRowsPerDoc["p2p_settings.md"]` to 41.

PR #1640, opened 2026-08-21
Remove dead BlockSettings.MaxSize setting

icellan:
## Summary

`BlockSettings.MaxSize` (key `blockmaxsize`, in `settings/block_settings.go`) has zero readers anywhere in the codebase. It is set from config in `settings/settings.go` and never read again. Its documentation is also wrong: it claims "Blocks exceeding this size are rejected during validation" and warns of forks from mismatched limits, but no validation path reads this field.

The real acceptance-side limit is `PolicySettings.ExcessiveBlockSize` (`excessiveblocksize`), enforced in `services/blockvalidation/BlockValidation.go`. The real mining-side limit is `PolicySettings.BlockMaxSize`, which shares the config key `blockmaxsize` with the field removed here (added independently, seven days apart, by different commits/authors, with no cross-reference) and has real readers: `services/blockassembly/BlockAssembler.go` (`filterSubtreesByMaxSize`, gating the mining candidate) and `services/rpc/handlers.go` (exposed as `maxminedblocksize`).

Because both fields carried the `key:"blockmaxsize"` tag, `settings/export.go`'s reflection-based settings-portal export was emitting two entries for `blockmaxsize` with contradictory descriptions and defaults. After this change it emits one.

This is deletion only:
- `PolicySettings.BlockMaxSize` is untouched.
- `PolicySettings.ExcessiveBlockSize` is untouched.
- No renaming, no re-wiring.

No deprecated-alias period is needed: an operator with `blockmaxsize` set in config keeps exactly the behavior the `Policy` field already gives them (unchanged), and the deleted field had no behavior of its own to lose.

## Verification

- `go build ./...`
- `go vet ./...` (pre-existing unrelated vet warnings in `test/utils` confirmed present on `main` before this change, untouched by this diff)
- `go test ./settings/... -race` — pass
- `go test ./services/blockassembly/... ./services/blockvalidation/... ./services/rpc/... -race` — pass
- `git grep -n "blockmaxsize"` — only the `Policy` field, `settings.conf`, `test/config/*.conf`, and unrelated `services/legacy` config remain

## Note for a concurrent PR

A separate open PR (`fix/settings-doc-drift-sweep`) touches `settings/` and adds a doc-drift guard test with an exemption for the `blockmaxsize` key collision between `BlockSettings.MaxSize` and `PolicySettings.BlockMaxSize`. Once both this PR and that one have landed, that specific exemption for the `blockmaxsize` collision should be removed from the doc-drift guard test, since the collision it was written for no longer exists.


ctnguyen (approved):
Review at commit [acfd63c](https://github.com/icellan/teranode/commit/acfd63ce31de95966e1607a7729921a13fb9acae)

This is the kind of cleanup I like reviewing: a two-line deletion with a PR body that actually does the archaeology. The dead-field claim checks out — `git grep -n 'Block\.MaxSize'` and a search for `BlockSettings.*MaxSize` in `*.go` both come back empty at this head, so the field really had zero readers and the deletion cannot change any runtime decision. The diagnosis of *why* it was dead is right too: `PolicySettings.BlockMaxSize` and `PolicySettings.ExcessiveBlockSize` are the fields with real enforcement behind them, and having a second field silently squatting on the `blockmaxsize` key with contradictory documentation was a genuine trap for anyone reading the settings reference. Removing the duplicate export entry is a real improvement, not just tidiness — the settings API was serving that key twice with contradictory text, and now it serves it once. Nothing here needs to block the merge; what follows is documentation and follow-up hygiene around the edges of an otherwise clean deletion.

## Non-blocking issues

### ChiR1 — The settings API listed `blockmaxsize` twice; say which entry survives and what its text loses

**Problem:** `ExportMetadata` at `settings/export.go:43` appends one `SettingMetadata` per cached entry (`:53-73`) with no key-uniqueness check. The consumer does not add one either: `services/asset/httpimpl/settings_handler.go:75-103` calls `ExportMetadata()`, runs the result through `redactSensitiveSettings`, then filters and sorts and appends every surviving entry — there is no dedup step anywhere on that path. So before this diff the settings API genuinely returned **two** `blockmaxsize` entries in a single response, with contradictory defaults (`4294967296` vs `0`), contradictory usage hints (`"4GB default (BSV)"` vs `"0 means unlimited (use excessiveblocksize)"`) and contradictory descriptions (acceptance-side "Blocks exceeding this size are rejected during validation" vs mining-side "Maximum block size to create when mining"). After this diff it returns one, and the survivor is the Policy field — the one with real readers. The PR body's "After this change it emits one" is accurate. What it does not say is *which* one, or that the entry going away carried the acceptance-side prose an operator may have been reading as the documentation for that key. I have not read whatever renders this API response, so I make no claim about what a UI did with the duplicate.

**Why it matters:** This is the most user-visible consequence of the change and the body under-sells it in one direction while leaving a gap in the other. Under-sells: "two entries with contradictory descriptions" reads like a tidiness issue, when in practice an operator querying the settings API for `blockmaxsize` got two rows disagreeing about both the default and the semantics, with nothing indicating which governed anything. The gap: the row that disappears is the one whose text described block *acceptance*, and the correct acceptance-side key — `excessiveblocksize` — is a different entry entirely. Anyone who had anchored on the removed description will find the acceptance semantics no longer documented under the key they were looking at, even though nothing about acceptance behaviour changed.

**Fix:** Spell out the before/after in the summary — something like: "The settings API (`services/asset/httpimpl/settings_handler.go`, which does not dedup by key) returned two `blockmaxsize` entries with contradictory defaults and descriptions; it now returns one, the mining-side `PolicySettings.BlockMaxSize` (`default:\"0\"`). The acceptance-side text that the removed entry carried was wrong for this key — acceptance is governed by `excessiveblocksize`. No runtime behaviour changes." Worth a separate follow-up: have `ExportMetadata` fail loudly or log on a duplicate `Key` rather than emitting both, so the next accidental collision surfaces at build or startup instead of in an API response.

### ChiR2 — The note for the sibling PR says "that specific exemption", but #1639 adds two

**Problem:** The PR body's closing note tells the reader that once both changes land, "that specific exemption for the `blockmaxsize` collision should be removed from the doc-drift guard test" — singular, one exemption, one test. Checked against #1639's own diff, it adds two distinct exemptions: a `fieldPaths: []string{"PolicySettings.BlockMaxSize", "BlockSettings.MaxSize"}` entry in the export-test exemption map, and a separate `key: "blockmaxsize"` exemption in the `settings_doc_test.go` defaults exemptions. Whoever picks up the follow-up from this sentence will remove one and leave the other. That the two PRs are coordinated at all is settled between them — #1639's exemption text already names "the pending fix/remove-dead-block-maxsize-setting change" — so this is purely about the note being an accurate work item (see [bsv-blockchain/teranode#1639](https://github.com/bsv-blockchain/teranode/pull/1639)).

**Why it matters:** A dangling exemption in a drift guard is worse than no exemption: it keeps a check permanently suppressed for a collision that no longer exists, so a future re-introduction of a duplicate `blockmaxsize` key passes the guard silently. The note is also the only record of the follow-up — there is no issue tracking it — so its precision is the whole safety net. There is a merge-order wrinkle too: #1639 is still open and was updated after this PR, so if this lands first, #1639's `fieldPaths` exemption references `BlockSettings.MaxSize`, a field that no longer exists; whether that is inert or breaks the exemption match depends on how the map is keyed, which I have not read.

**Fix:** Make the note enumerate both, and name the merge order:

```
Once both this PR and #1639 have landed, remove BOTH blockmaxsize exemptions
added by #1639: the fieldPaths entry {"PolicySettings.BlockMaxSize",
"BlockSettings.MaxSize"} in the export-test exemption map, and the
key: "blockmaxsize" entry in settings_doc_test.go's defaults exemptions.
If this PR lands first, #1639 should drop the BlockSettings.MaxSize half
before merging.
```

Better still, open a tracking issue and reference it from both PRs rather than relying on a paragraph in a merged description.

### ChiR3 — `settings.conf` context overrides still carry the deleted field's 4 GB default

**Problem:** The config file ships the only inline explanation of this key:

```
settings.conf:385:  # Set maximum block size in bytes we will mine. Size of the mined block will never exceed the maximum block size we will accept (excessiveblocksize)
settings.conf:386:  blockmaxsize          = 0
settings.conf:387:  blockmaxsize.docker.m = 4294967296
settings.conf:388:  blockmaxsize.operator = 4294967296
```

The comment describes the mining-side semantics correctly, matching the surviving Policy field. But `blockmaxsize.docker.m` and `blockmaxsize.operator` both set 4294967296 — the same value as the deleted field's `default:"4294967296"`. With the acceptance-side reading of this key gone, those context overrides now unambiguously mean "cap mined blocks at 4 GB", which is a real mining-policy statement rather than a mirror of a default.

**Why it matters:** Non-blocking because the runtime values are unchanged by this diff — the Policy field read those same overrides before. But this PR is the moment where the key's meaning becomes unambiguous, and it is worth confirming the docker and operator contexts actually intend a 4 GB mining cap rather than having inherited the deleted field's default by coincidence. I have not established that they were copied from it; the values simply coincide.

**Fix:** Confirm the intent of `blockmaxsize.docker.m` / `blockmaxsize.operator`; if 4 GB is deliberate, no change. If they were tracking the removed field's default, drop the overrides so those contexts fall through to the documented `0` (unlimited). Either way the PR body's "No renaming, no re-wiring" list could note that config values were reviewed and left intentionally unchanged.

## Recap

| ChiR1 | Duplicate settings API entry removed | | 32% |
| ChiR2 | Sibling PR note undercounts exemptions | | 30% |
| ChiR3 | Confirm intent of 4GB context overrides | | 15% |

ordishs (approved):

PR #1641, opened 2026-08-21
fix(pruner): wire up unwired settings, guard nil blockchain client, drop dead flag

icellan:
## Summary

Three pruner settings carried `key:` struct tags and appeared in the reference docs, but were never populated by `settings.NewSettings()`, so setting them in `settings.conf` silently did nothing.

- Wire up `pruner_skipDuringCatchup` and `pruner_skipProcessExpiredPreservations` in the Pruner settings block, matching the population style of their siblings. Defaults stay `false`, preserving current behaviour.
- Fix a missing nil guard on the catchup-skip check: it dereferenced `blockchainClient` with no nil check, unlike the analogous `blockAssemblyClient` check a few lines below. Test harnesses that construct `&Server{}` with a nil `blockchainClient` would panic the moment this setting was turned on.
- Delete `pruner_force_ignore_block_persister_height`. It has zero read sites - it was wired once, then deliberately removed and replaced by `pruner_block_trigger` (which is fully wired), then reintroduced as a tag-only field with no implementation behind it. Replaced its doc section with a pointer to `pruner_block_trigger=OnBlockMined`, which is the supported way to get that behaviour.

Also noted in the settings.conf comment and reference docs: skipping pruning during catchup defers the prune burst rather than eliminating it - a node that skips a long catchup hits one much larger DAH sweep on entering RUNNING.

## Test plan

- [x] Added `settings/pruner_settings_test.go`: proves the loader actually reads `pruner_skipDuringCatchup` and `pruner_skipProcessExpiredPreservations` from config (override + default-false cases), not just that the fields work when set directly on the struct.
- [x] Added tests in `services/pruner/worker_test.go` covering both FSM branches (CATCHINGBLOCKS skips and records the metric, RUNNING proceeds normally) using the existing `blockchain.Mock`, plus a regression test that reproduces the nil-pointer panic pre-fix and passes post-fix.
- [x] `go build ./...`
- [x] `go vet ./settings/... ./services/pruner/...`
- [x] `go test ./settings/... ./services/pruner/... -race`
- [x] `golangci-lint run ./settings/... ./services/pruner/...`
- [x] `staticcheck ./settings/... ./services/pruner/...`
- [x] `settings/coverage_test.go` still passes after the deletion


ctnguyen (approved):
Review at commit [de78de0](https://github.com/icellan/teranode/commit/de78de043215a965aedb02479d3d522f19fd935f)

Small, competent settings-hygiene fix, and the diagnosis behind it is right: three `key:`-tagged fields that `NewSettings()` never populated, so an operator could set them and get silence. Wiring the two live ones next to their siblings, deleting the one with no implementation behind it, and redirecting the docs to `pruner_block_trigger=OnBlockMined` is the correct shape. The best part is `settings/pruner_settings_test.go`, which explains why a default-only assertion cannot catch a relapse — both defaults are `false`, which is also the Go zero value — and then tests the override path, so it actually guards the bug rather than restating it. The `worker_test.go` additions cover both FSM branches rather than only the happy one. Nothing here needs to block merge; the notes below are about what enabling `pruner_skipDuringCatchup` now means operationally, plus some test cleanup. (Incidental: adding `SkipProcessExpiredPreservations` re-aligned the whole `Pruner:` struct literal, so the `settings/settings.go` diff is 22 rewritten lines for 2 added ones — gofmt doing its job, but worth knowing if this needs a rebase.)

## Non-blocking issues

### ChiR1 — An FSM lookup failure defers pruning for that notification

**Problem:** At `services/pruner/worker.go:166-170`, an error from `GetFSMCurrentState` logs at `Warnf`, increments `prunerSkipped{fsm_error}` and `continue`s, so the prune cycle for that notification does not run. This branch was unreachable before this PR — `SkipDuringCatchup` could never be true — so wiring the setting is what makes it live. This is not data loss and not invisible: the component's own contract at `worker.go:134-136` states that "the atomic target height always reflects the latest notification, so intermediate heights are naturally skipped", and with the `blockHeight <= s.lastProcessedHeight.Load()` guard at `:150-152` a later, higher notification still drives a later sweep. The narrow point is that a blockchain-client error is now treated the same as "we are catching up", when the two mean different things: one is a known state, the other is an unknown one.

**Why it matters:** The direction is fail-safe for data — skip rather than delete — so this is a design question, not a defect. But an operator who enables the setting for its stated purpose (avoiding a catchup race) also silently opts into "pruning pauses while the blockchain client is unhealthy", and nothing in the `longdesc` or `settings.conf` comment mentions that. If FSM lookups fail persistently, the DAH sweep stalls while the node keeps taking blocks, and the only signal is a `Warnf` plus a counter nobody may be graphing.

**Fix:** Either treat an unknown state as "proceed" — matching how a nil `fsmState` already falls through at `worker.go:171` — or keep the skip and document it. If keeping it, `Errorf` reads better than `Warnf` for a condition that suspends pruning:

```go
fsmState, err := s.blockchainClient.GetFSMCurrentState(ctx)
if err != nil {
    // Unknown state, not a known catchup state: proceed rather than
    // deferring the prune on a transient client error.
    s.logger.Errorf("[pruner][%s:%d] failed to get FSM state, proceeding: %v", blockHashStr, blockHeight, err)
    prunerErrors.WithLabelValues("fsm_state").Inc()
} else if fsmState != nil && *fsmState == blockchain.FSMStateCATCHINGBLOCKS {
    // ... existing skip
}
```

Whichever way it goes, add a line to the `pruner_skipDuringCatchup` `longdesc` saying that enabling it makes FSM-client health a dependency of pruning.

**Criticality: 28%**

### ChiR2 — The nil `blockchainClient` guard is test-only scaffolding in a production path

**Problem:** `worker.go:164` now reads `if s.settings.Pruner.SkipDuringCatchup && s.blockchainClient != nil`, with a comment saying it mirrors the `blockAssemblyClient != nil` check at `worker.go:184`. The convention match is real, but the two cases are not equivalent: `s.blockchainClient` is dereferenced unconditionally on the startup path at `services/pruner/server.go:143` (`Subscribe`), and again at `:221`, `:250` and `:267`, plus `blob_deletion_worker.go:94`, `:177` and `:330`. A nil client panics in `Init` long before `prunerProcessor` ever runs, so on any production path the new conjunct is always true. Where it does fire — a hand-built `&Server{}` in a test — the effect is that `SkipDuringCatchup` becomes a no-op with no log line saying so.

**Why it matters:** Low severity precisely because the nil case is unreachable in production; this is not a defect. But it reproduces, one level down, the exact failure mode the PR is fixing: a setting that reads as enabled and quietly does nothing. Also worth correcting the PR description, which calls this "a missing nil guard" that "would panic the moment this setting was turned on" — the panic is reachable from a test harness, not from a deployed node, and the changelog reads more alarmingly than the code warrants.

**Fix:** Keep the guard, it is cheap and conventional, but make the no-op audible:

```go
if s.settings.Pruner.SkipDuringCatchup {
    if s.blockchainClient == nil {
        s.logger.Warnf("[pruner][%s:%d] pruner_skipDuringCatchup=true but no blockchain client; not checking FSM state", blockHashStr, blockHeight)
    } else {
        // ... existing FSM check
    }
}
```

**Criticality: 20%**

### ChiR3 — New settings tests leave keys behind in the global config

**Problem:** `settings/pruner_settings_test.go:46-47` sets a key on the process-global `gocore.Config()` and "restores" it with `Set(key, "")`, which leaves the key present as an empty string rather than removing it. `TestPrunerSettings_UnwiredKeysDefaultToFalse` (`:57-60`) writes the same empty strings and has no cleanup at all. This matches the established convention in the package — `settings/asset_settings_test.go:89-90`, `settings/batcher_max_concurrent_test.go:45-46`, `settings/batcher_tickinterval_test.go:98-99` and `settings/helpers_test.go:13-14`, `:64` all do the same thing — so the new file is consistent, not novel. The cleanup pattern itself is the weak part, and there is a better option available: `gocore.Configuration.Unset` exists (`gocore@v1.0.82/config.go:527`).

**Why it matters:** Latent order-dependence rather than a present bug: leftover empty-string keys mean a later test in the same package sees "set to empty" instead of "absent". It also slightly weakens the default-false test's own claim, since it asserts against keys it just blanked — so it proves "empty string parses to false" as much as it proves "the default is false", and this file exists specifically to be a durable relapse guard.

**Fix:** Use `Unset` on cleanup, and give the default-false test cleanup it currently lacks:

```go
t.Cleanup(func() { gocore.Config().Unset(tc.key) })
```

For the default case, drop the `Set(key, "")` loop and `Unset` the keys before calling `NewSettings()` so the assertion tests a genuinely absent key. Converting the older tests to `Unset` is out of scope here, but worth a follow-up if the convention is going to spread.

**Criticality: 12%**

## Recap

|---|---|---|---|
| ChiR1 | FSM error defers prune cycle | | 28% |
| ChiR2 | Nil guard is test-only scaffolding | | 20% |
| ChiR3 | Config keys blanked not unset | | 12% |

ordishs (approved):
Approving. The change fixes a real bug: `pruner_skipProcessExpiredPreservations` is read at `services/pruner/worker.go:234` and `:243`, but `settings.NewSettings()` never populated it, so the documented Phase 1b kill-switch could never be turned on. Same shape for `pruner_skipDuringCatchup` at `worker.go:163`.

Verified independently:

- Tag/wire parity for the whole Pruner block is now exact. A mechanical diff of the `key:` tags in `settings/pruner_settings.go` against the populate calls in the Pruner literal of `settings/settings.go` gives an empty set in both directions.
- `ForceIgnoreBlockPersisterHeight` and `pruner_force_ignore_block_persister_height` have zero references left anywhere in the tree — no Go code, no chart, no manifest, no doc.
- The ~20 realigned lines in the Pruner literal are whitespace only. Field names, keys and defaults are unchanged.
- No generator writes `docs/references/settings/**`, so the added note survives regeneration.

```
$ go build ./...                                       # exit 0
$ go vet ./settings/... ./services/pruner/...           # exit 0
$ go test -count=1 -race ./settings/... ./services/pruner/...
ok  github.com/bsv-blockchain/teranode/settings          1.690s
ok  github.com/bsv-blockchain/teranode/services/pruner   4.830s
```

No P0 or P1 findings. One item worth acting on:

**`settings/settings.go:592,608` — release note, not a code change**

Wiring these keys is a behaviour change for any deployment that already set them, got no effect, and now gets the real behaviour on upgrade. `pruner_skipProcessExpiredPreservations=true` is the one that matters: the struct tag itself says preserved parents then "accumulate and are never pruned (unbounded retention)".

The blast radius is small. `settings.conf:593` pins `pruner_skipDuringCatchup = false`, and `pruner_skipProcessExpiredPreservations` is absent from `settings.conf` entirely, so no deployment inherits `true` from the shipped defaults — only an operator's own override reaches the new path. `worker.go:244` also logs at Info on every prune cycle when Phase 1b is skipped, and names the key.

Suggest one line in the release notes: these two keys were inert and are now honoured, so audit `settings_local.conf` and environment overrides before upgrading.

Minor, none blocking:

- `settings/pruner_settings_test.go:46-49,58-60` — both new tests drive the loader only through `gocore.Config().Set`, but gocore reads `os.LookupEnv(key)` before the `confs` map, so an exported key beats the `Set` and the `""` cleanup. Reproduced with `pruner_skipDuringCatchup=true go test -run TestPrunerSettings ./settings/` → `--- FAIL: TestPrunerSettings_UnwiredKeysDefaultToFalse`. This is the standing convention of the package (`asset_settings_test.go:89-90`, `batcher_tickinterval_test.go:98-99`, `p2p_settings_test.go`), which all break the same way, so it belongs in a package-wide follow-up with `t.Setenv` rather than here.
- `services/pruner/worker.go:163` — the new `&& s.blockchainClient != nil` guard cannot fire in production. `daemon/daemon_services.go` propagates the error from `GetBlockchainClient`, which never returns `(nil, nil)`, and `Init` derefs the client unguarded at `server.go:143` (`Subscribe`) before `prunerProcessor` ever runs. The guard is test-only, and it makes a safety flag silently inert with no log or metric, where the adjacent FSM-error branch fails closed and records `prunerSkipped("fsm_error")`. It does match the file's own convention (`worker.go:181`, `server.go:450`), so this is a preference. A `blockchain.Mock` returning an error would reach the `fsm_error` skip and cover the same harness case without a production branch. If the guard stays, a `prunerSkipped.WithLabelValues("fsm_unavailable")` plus a Warnf would make the inert flag visible.
- `services/pruner/worker.go:163-176` — two of the four FSM outcomes stay uncovered: `GetFSMCurrentState` returning an error, and a `(nil, nil)` return that falls through and prunes. Both are pre-existing context lines, and cheap to add with the `blockchain.Mock` already imported.
- `docs/references/settings/services/pruner_settings.md:190` — the replacement sentence now sits under `### pruner_block_trigger` and restates the `OnBlockMined` bullet two lines above. Suggest dropping it.

PR #1643, opened 2026-08-21
fix(settings): wire five unpopulated blockchain/legacy settings

icellan:
## Summary

An audit found struct fields carrying `key:` tags that `NewSettings()` never populates, so operators setting the key get silence instead of an effect. This handles the five zero-risk cases; the remaining findings need follow-up review.

- Deletes `blockchain_subscription_timeout` (`BlockChainSettings.SubscriptionTimeout`): confirmed no reader anywhere in the codebase (`git grep -n '\.SubscriptionTimeout\b'` returns nothing). History shows the tag was resurrected after a deliberate removal that also dropped its population and reader; this is not the same field as `BlockAssemblySettings.BlockchainSubscriptionTimeout`, which is populated and left untouched.
- Wires four settings, all default-off so production behaviour is unchanged:
  - `blockchain_peerRegistryStore` / `blockchain_peerRegistrySaveInterval` — gate the peer registry blob-store persistence block in the blockchain service (`services/blockchain/Server.go`). These ship together: the save interval is only reached once the store URL is set, so with the default empty URL the whole block still never runs.
  - `legacy_upnp` — reaches the UPnP branch already present in the legacy peer server; also removes the stale `// c.Upnp = true // TODO set from settings` comment now that it's done. Enabling this activates a UPnP code path that has otherwise never executed in this codebase; that path itself is out of scope here.
  - `blockchain_raw_miner_tag` — reaches the SQL blockchain store's miner-tag sanitization toggle. The existing test named `TestRawMinerTagSettingWiring` asserted the struct field directly, bypassing `NewSettings()`, so it could never have caught the missing wiring; it now drives the setting through `gocore.Config().Set` + `NewSettings()`.

## Test plan

- [x] Added `settings/blockchain_settings_test.go` and `settings/legacy_settings_test.go`: set each key via `gocore.Config().Set`, call `NewSettings()`, assert the field reflects the override, plus a defaults-unchanged assertion.
- [x] Fixed `stores/blockchain/sql/rawminertag_test.go` to drive the setting through config + `NewSettings()` instead of assigning the struct field.
- [x] `go build ./...`
- [x] `go vet ./settings/... ./services/blockchain/... ./services/legacy/... ./stores/blockchain/...`
- [x] `go test` and `go test -race` over `./settings/...`, `./services/blockchain/...`, `./services/legacy/...`, `./stores/blockchain/...`
- [x] `settings/coverage_test.go` tag-coverage tests pass after the deletion
- [x] `golangci-lint run` over the touched packages: 0 issues


ctnguyen (approved):
Review at commit [6d5ac15](https://github.com/icellan/teranode/commit/6d5ac159e0832416d16edf86e1ddb5470b9af84e)

This is a careful piece of settings hygiene. The audit framing is honest — five keys carried `key:` tags that `NewSettings()` never populated, so an operator editing them got silence instead of an effect — and the fix stays tightly scoped: four `getX` calls, one dead field removed, and tests that would actually have caught the original bug. Two things stand out as real quality moves rather than box-ticking. First, the `SubscriptionTimeout` deletion is backed by a zero-reader check and correctly distinguishes the field from `BlockAssemblySettings.BlockchainSubscriptionTimeout`, which is left alone — that's the kind of homework that makes a deletion safe to approve. Second, `TestRawMinerTagSettingWiring` was rewritten to drive the value through `gocore.Config().Set` + `NewSettings()` (`stores/blockchain/sql/rawminertag_test.go:46-51`) instead of assigning `tSettings.BlockChain.RawMinerTag` directly; the old form structurally could not have detected the missing wiring, and the new one can. The new tests in `settings/` follow the same discipline — they assert an *override*, not just a default, because a default assertion on a zero-valued field passes whether or not the loader reads the key, and both files say so in their comments. Nothing below is a blocker; two modest notes.

## Non-blocking issues

### ChiR1 — no runtime signal at the point `legacy_upnp` takes effect

**Problem:** `settings/settings.go:697` wires `Upnp: getBool("legacy_upnp", false, …)`, and `services/legacy/peer_server.go:3784-3786` turns that into `cfg.Upnp = true`. Default-off is airtight (`getBool` with a `false` default), and the PR description discloses up front that this path has not executed in this codebase and that the path itself is out of scope — that deferral is reasonable and I'm not asking for it to be reopened. The narrow gap is that a *running* operator who flips the key gets nothing at the point it takes effect: the config change is only visible in the config file, not in the logs of the process whose behaviour it changed.

**Why it matters:** The disclosure covers the reviewer of this PR, and the `longdesc` covers someone reading the generated settings docs. Neither covers the person tailing logs after a port-mapping oddity appears, since the symptom surfaces far from the line that caused it. Compare the sibling peer-registry block, which does narrate itself — `Infof` on load and a `Warnf` at `services/blockchain/Server.go:501` when the interval isn't positive. One line brings the UPnP switch up to the same standard.

**Fix:** Log where the flag is consumed:

```go
if tSettings.Legacy.Upnp {
    cfg.Upnp = true
    logger.Warnf("[legacy] legacy_upnp enabled; UPnP port mapping is not exercised in this deployment")
}
```

**Criticality: 12%**

### ChiR2 — `getURL` parse failures are indistinguishable from unset (shared helper, follow-up)

**Problem:** `getURL` returns nil both when the key resolves to `""` (`settings/helpers.go:85-87`) and when `ParseMultiHostURL` fails (`:89-92`). Because `services/blockchain/Server.go:481` gates on `storeURL != nil`, a malformed `blockchain_peerRegistryStore` produces exactly the behaviour of not setting it: persistence silently off. To be clear about scope — this is a property of the shared helper, not of this diff. `getURL(` appears 28 times in `settings/settings.go` (32 across `settings/*.go`), so every URL-valued setting in the repo already behaves this way; this PR adds one more subscriber to an existing contract, and the two new lines at `settings/settings.go:332` and `:323` are correct as written.

**Why it matters:** It's the same failure shape this PR set out to remove — a key that reads as configured but has no effect — one layer down in the helper. Worth naming here because the diff surfaced it, not because the diff caused it. Blast radius per setting varies; for the peer registry it's bounded (reputation history isn't persisted, the node runs fine) but invisible, and the operator's mental model will say persistence is on.

**Fix:** Nothing in this PR. As a follow-up against the helper, a `Warnf` in the `err != nil` branch would make a set-but-unparseable value visible for all ~28 callers at once:

```go
u, err := urlutil.ParseMultiHostURL(str)
if err != nil {
    // log once here rather than at each of the ~28 call sites
    return nil
}
```

Changing the nil-return contract itself would touch every caller and is rightly out of scope.

**Criticality: 8%**

## Recap

| ChiR1 | No log when UPnP enabled |  | 12% |
| ChiR2 | Shared getURL swallows parse errors |  | 8% |

icellan:
Rebased onto current main (fd830d1de).

Conflict: `settings/settings.go` only, two hunks in `defaultSettings()`.

- `BlockChain` block: main had gained `StoreDBTimeoutMillis` (now via `DefaultBlockchainStoreDBTimeoutMillis` constant) and `InitializeNodeInState` (now `strings.TrimSpace`-wrapped). Kept both of those plus this PR's three new population lines (`RawMinerTag`, `PeerRegistryStore`, `PeerRegistrySaveInterval`) — union, no lines dropped from either side.
- `Legacy` block: main had added `PeerRegistryEnabled` / `PeerRegistrySyncInterval` population lines (unrelated feature). Kept those and added this PR's `Upnp` line. No conflicting UPnP work landed on main — the `Upnp` struct field and the `peer_server.go` reference were untouched elsewhere, so this was a straight textual conflict, not two competing fixes.

`services/legacy/peer_server.go` and `settings/blockchain_settings.go` auto-merged cleanly (struct fields for the three settings already existed on main from other work; only the `settings.go` population lines were missing, which is what this PR adds).

Verified after resolution:
- `SubscriptionTimeout` deletion intact, `BlockAssemblySettings.BlockchainSubscriptionTimeout` untouched.
- All four wirings present (`blockchain_peerRegistryStore`, `blockchain_peerRegistrySaveInterval`, `legacy_upnp`, `blockchain_raw_miner_tag`).
- `rawminertag_test.go` still drives the setting through `NewSettings()`/gocore config, not direct struct assignment.
- `go build ./...`, `go vet` and `go test` / `go test -race` over `settings/`, `services/blockchain/`, `services/legacy/`, `stores/blockchain/` all green, including `settings/coverage_test.go`.

Force-pushed the rebased branch.

ordishs (approved):
Verified the description's claims and they hold up:

- `git grep '\.SubscriptionTimeout\b'` returns nothing, and no `blockchain_subscription_timeout` string exists in `settings.conf` or the docs tree. The deletion is safe.
- The four new loader lines match the key strings and default values in the struct tags exactly.
- `getURL` returns `nil` for an empty string (`settings/helpers.go:81-93`), so the `require.Nil` assertion is correct and the persistence block at `services/blockchain/Server.go:481` stays unreached by default.
- `test.CreateBaseTestSettings` calls `NewSettings()` (`util/test/helpers.go:20`), so the rewritten `rawminertag_test.go` does drive the real loader.

I commented out the four new loader lines and re-ran the new tests. Both fail. The tests genuinely guard the wiring, which the previous `rawminertag_test.go` could not. `go build ./...`, `go vet`, `go test` and `go test -race ./settings/...` all pass.

One finding, about what the change switches on rather than about the diff itself.

**`settings/settings.go:332` with `services/blockchain/Server.go:478-505`** — P1

This PR makes peer registry persistence reachable. The same code block reads three more settings that `NewSettings()` still does not populate: `p2p_peer_registry_ttl`, `p2p_peer_registry_cleanup_interval` and `p2p_peer_registry_max_size`. All three appear only in `settings/p2p_settings.go:56-58` as `key:` tags. Grepping `settings.go` for `PeerRegistryTTL`, `PeerRegistryCleanupInterval` or `PeerRegistryMaxSize` returns nothing, so the fields stay at the Go zero value.

Failure: an operator sets `blockchain_peerRegistryStore` to enable persistence, as this PR now permits, and sets `p2p_peer_registry_max_size=10000` as the tag documents. `NewSettings()` never reads that key, so the field stays 0, and `PeerRegistryCleanupInterval` stays 0 for the same reason. `Server.go:505` then calls `StartCleanup(ctx, 0, ttl, 0)`, and `peer_registry.go:934-937` returns immediately on a non-positive interval. No TTL eviction and no LRU bound ever run at runtime. The registry grows for every peer ID seen through `UpdatePeerMetrics`, and this PR now writes that unbounded map to a blob store every 60 seconds. Only the 24h fallback TTL applied at `Load` prunes it, and only on restart.

Recommendation: wire the three `p2p_peer_registry_*` keys in this change. Their tag defaults are 24h, 1h and 10000, so they restore the documented bound rather than leave it off. I understand the PR scopes itself to default-off cases and these three do change behaviour. If you prefer to keep that scope, please gate the persistence block on a populated cleanup interval, or note the gap in the PR so the follow-up lands first.

Separately, and not blocking: the suite still has no systemic guard for this bug class. `TestMetadataCoverage` counts `key:` tags but never checks the loader reads them. Comparing every `key:` tag against every literal key passed to a `get*` helper, 15 tagged keys still have no read — the three above, five `postgres_circuitBreaker*`, three `pruner_*`, two `*_subtreeMmapDir`, `blockassembly_txMapDirs` and `aerospike_enable_preserve_filter_expressions`. One reflection-based test that sets each tagged key and asserts the field changed would catch all 15 and retire the per-key tests. I left two P3 test-hygiene notes out of this comment.

PR #1645, opened 2026-08-21
fix(settings): wire peer registry TTL/LRU cleanup settings into NewSettings

icellan:
## Summary

Three P2P settings — `p2p_peer_registry_max_size`, `p2p_peer_registry_ttl`, `p2p_peer_registry_cleanup_interval` — carried `key:`/`default:` struct tags on `P2PSettings`, but `settings.NewSettings()` never populated them. They stayed at Go-zero regardless of what an operator put in `settings.conf`.

`services/blockchain/Server.go` feeds the cleanup interval straight into `CentralizedPeerRegistry.StartCleanup`, which returns immediately when the interval is `<= 0`. So the peer registry's TTL+LRU eviction loop has never actually run: the registry grows without bound for every peer ID ever seen, degrading every O(n) scan over it (catchup peer selection, reputation queries).

This adds the three missing `getInt`/`getDuration` calls, matching the sibling `PeerMap*` wiring style already in the same block. Defaults (`10000` / `24h` / `1h`) match the struct tags, so a deployment that never set these keys sees no behaviour change from the wiring itself — but a deployment on a long-running node will now actually start evicting stale registry entries on the configured cadence, which is the intended fix.

`docs/p2p-libp2p-review.md` claimed this was already resolved and pointed at the wrong file (`services/p2p/peer_registry.go` instead of `services/blockchain/peer_registry.go`); corrected.

## Safety of enabling eviction

Read `StartCleanup`/`Cleanup` (`services/blockchain/peer_registry.go`) in full before wiring this:

- TTL phase and LRU phase both exempt `IsConnected || IsBanned` peers unconditionally.
- LRU eviction is strictly oldest-first by `peerActivity` (max of `LastSeen`/`LastMessageTime`), and every successful catchup interaction (`RecordCatchupSuccess` → `recordReceivedSuccessLocked`) refreshes `LastSeen` to now, so an actively-serving peer's recency keeps resetting and it never becomes the LRU eviction candidate.
- Catchup/sync peer selection (`services/p2p/peer_selector.go` `isEligibleBasic`) does not gate on `IsConnected` — a gossiped-but-not-directly-connected peer can be selected as a sync source. That peer is not covered by the connected/banned exemption. In practice this is a narrow residual risk, not a correctness bug: `RecordCatchupAttempt` alone doesn't refresh recency, only success does, so a peer mid-flight on its very first catchup attempt (no success yet) is theoretically evictable before TTL — but TTL defaults to 24h and eviction of a registry entry does not abort in-flight HTTP fetches from its `DataHubURL`; `RecordCatchupSuccess`/`Failure` no-op safely if the entry is gone, and the sync coordinator just re-selects from the next `GetPeers` snapshot. No crash path found.
- `maxSize=0` disables the LRU phase and is a documented sentinel (`Cleanup` comment, settings longdesc), not a bug — left as-is.
- `cleanupInterval<=0` disables the whole loop and is likewise a documented sentinel (`StartCleanup` comment) for operators who only want TTL-on-load semantics — left as-is. No in-code fallback added for either, since a fallback would silently override an intentional, documented opt-out. TTL keeps its existing `Server.go` fallback to 24h for the same zero, since TTL has no such documented sentinel and a literal zero there would evict almost the entire non-exempt registry on the first tick.

## Test plan

- [x] `TestP2PPeerRegistrySettings_LoaderReadsAllKeys` / `_DefaultsMatchTheStructTags` (`settings/p2p_settings_test.go`) — proves the loader reads overrides and matches the struct tag defaults.
- [x] `TestCentralizedPeerRegistry_StartCleanup_WiredFromSettings` (`services/blockchain/peer_registry_test.go`) — drives `StartCleanup` with values read back from `settings.NewSettings()` (not hand-written constants), confirms the loop evicts a stale entry and that an actively-connected peer survives both the TTL and LRU-over-max-size phases.
- [x] Existing `TestCentralizedPeerRegistry_Cleanup_TTL`, `_LRUExemptsConnectedAndBanned`, and related `Cleanup`/`StartCleanup` tests already cover TTL expiry and LRU-over-max-size separately at the unit level.
- [x] `go build ./...`, `go vet ./settings/... ./services/blockchain/... ./services/p2p/...`
- [x] `go test` and `go test -race` over `./settings/... ./services/blockchain/... ./services/p2p/...`
- [x] `golangci-lint run`, `staticcheck` clean on the changed packages


ctnguyen (approved):
Review at commit [7b17a75](https://github.com/icellan/teranode/commit/7b17a754e94bae82ff375dde7e3b426a3a75139a)

Three lines of production code that fix a bug which has been live in every deployment. I traced the chain: `StartCleanup` returns immediately when `interval <= 0` (`services/blockchain/peer_registry.go:935-937`), its only caller feeds it `b.settings.P2P.PeerRegistryCleanupInterval` (`services/blockchain/Server.go:478`, passed at `:512`), and `NewSettings()` never populated that field — so the value was always the zero `time.Duration` and the TTL+LRU eviction goroutine never started, no matter what an operator put in `settings.conf`. The registry accumulated an entry for every peer ID ever seen, and the O(n) scans over it (catchup selection, reputation queries) paid for all of them. That is a genuine latent bug, and the diff is the minimal correct fix: `settings/settings.go:538-540` with defaults matching the struct tags at `settings/p2p_settings.go:56-58`, so a deployment that never set these keys keeps the tag defaults, while the loop finally starts and an explicitly configured value is finally honoured. The safety analysis in the PR body is what makes this reviewable at a glance rather than a leap of faith — it reads `Cleanup` end to end, establishes that connected and banned peers are exempt in both phases, argues from `recordReceivedSuccessLocked` why an actively-serving peer never becomes the LRU candidate, and then raises a residual risk *against its own change* (`isEligibleBasic` not gating on `IsConnected`, and `RecordCatchupAttempt` not refreshing recency) before bounding the consequence. Deliberately leaving `cleanupInterval` and `maxSize` without a fallback, on the grounds that both zeros are documented opt-out sentinels while TTL has no such sentinel, is the right call and coherent with `Server.go:474-477`. The test that drives `StartCleanup` from `settings.NewSettings()` output rather than literals is the only shape that would catch a relapse, and it is the one that was written. Two notes below, both documentation text rather than behaviour.

## Non-blocking issues

### ChiR1 — Shipped longdesc still names a file that does not exist

**Problem:** The PR corrects the wrong-file reference in `docs/p2p-libp2p-review.md` (`services/p2p/peer_registry.go` → `services/blockchain/peer_registry.go`) but leaves the identical error in the `longdesc` struct tag of one of the three fields it is shipping. `settings/p2p_settings.go:56` still reads:

```
settings/p2p_settings.go:56 — longdesc: "### Purpose\nUpper bound on the per-peer record map (services/p2p PeerRegistry). ..."
```

`services/p2p/peer_registry.go` does not exist at this head; the implementation is `CentralizedPeerRegistry` in `services/blockchain/peer_registry.go`.

**Why it matters:** `longdesc` is the text the settings portal renders for operators, so of the two copies of this mistake it is the one with the wider and less technical audience — and the one someone reads while deciding what to set `p2p_peer_registry_max_size` to. Fixing the reference in an internal review doc while leaving it wrong in the operator-facing description of the very field being wired inverts the priority. It is also cheapest to fix now, while the field is already being touched, rather than after the legacy `services/p2p` registry is gone and the reference points at nothing recognisable at all.

**Fix:** Update the `longdesc` on `PeerRegistryMaxSize` to name the real owner, and check the sibling `PeerRegistryTTL` / `PeerRegistryCleanupInterval` longdescs at `settings/p2p_settings.go:57-58` for the same phrasing while you are in there:

```go
PeerRegistryMaxSize int `key:"p2p_peer_registry_max_size" default:"10000" longdesc:"### Purpose\nUpper bound on the per-peer record map (services/blockchain CentralizedPeerRegistry). ..."`
```

**Criticality: 25%**

### ChiR2 — Wiring comment understates what changes at runtime

**Problem:** The comment introduced at `settings/settings.go:534-537` says the defaults match the struct tags "so wiring these keys does not change behaviour for a deployment that never set them." Every clause is true, but the sentence reads as *no behaviour change*, when the actual effect is twofold: on a default-configured long-running node, TTL+LRU eviction begins running — hourly — for the first time ever; and on a node that did set one of these keys, the value stops being silently discarded and starts taking effect.

**Why it matters:** This is the copy of the reasoning that survives in the tree. Someone bisecting a future change in registry memory footprint or peer-selection behaviour will read it and reasonably deprioritise this commit, when it is in fact the commit that turned the eviction loop on. The PR body draws the distinction correctly ("a deployment on a long-running node will now actually start evicting stale registry entries on the configured cadence, which is the intended fix"); the in-tree comment kept the reassuring half and dropped the operative one.

**Fix:** Carry the operative half into the comment, and scope the reassurance to deployments that never set the keys rather than to configured values:

```go
// Centralized peer registry (services/blockchain) TTL+LRU cleanup
// configuration. Defaults match the struct tag defaults, so a deployment
// that never set these keys sees the same values as before — but the
// cleanup loop previously never started at all (StartCleanup returns early
// on a zero interval), so such a node now begins evicting stale registry
// entries hourly. And any explicitly configured value was previously
// ignored entirely; it is now honoured. That is the fix.
```

**Criticality: 10%**

## Recap

|---|---|---|---|
| ChiR1 | Longdesc names nonexistent file | | 25% |
| ChiR2 | Comment understates behaviour change | | 10% |

icellan:
Rebased onto current `origin/main` (`4241a96c7`, includes #1674 and the other recently-merged work). New head: `5f11b7fac0`.

Conflicts, both in the single commit:

- `settings/settings.go` (P2P block): main had added `SeenHash*` fields adjacent to where this PR adds `PeerRegistry*` fields. Union — kept both.
- `docs/p2p-libp2p-review.md`: two hunks.
  - The "PeerRegistry has no eviction policy — Resolved" bullet: main's wording had been updated independently by #1599 (`stop IsConnected from pinning peer-registry entries forever`), which changed the actual exemption semantics in `isCleanupExempt` (connected peers are now exempt only while their activity is within TTL, not unconditionally, plus a p2p-side reconciliation sweep for stale `IsConnected` flags). Kept main's more accurate description and folded in this PR's specific note about the settings-loader bug being the actual fix. Not a real behavioural conflict — main's text was just more current than what this branch had when it was written.
  - Nice-to-have list renumbering: pure renumber collision from unrelated items main had added elsewhere in the same list; no content lost either side.

Note for reviewers: main gained real peer-registry work since this branch was authored (#1390, #1599, #1611, #1664), most relevantly #1599 changing `isCleanupExempt`'s connected-peer exemption from unconditional to TTL-bounded, with a reconciliation sweep clearing stale `IsConnected` flags. That's a stricter/safer semantic than this PR's original analysis assumed, and it's already covered by its own tests on main. Wiring the three settings here just turns the (already more carefully gated) cleanup loop on; no additional risk from that change as far as I can tell.

Verified after rebase: the three settings wirings and all three call-out tests (`TestP2PPeerRegistrySettings_LoaderReadsAllKeys`, `TestP2PPeerRegistrySettings_DefaultsMatchTheStructTags`, `TestCentralizedPeerRegistry_StartCleanup_WiredFromSettings`) are intact and pass. `go build ./...`, `go vet ./settings/... ./services/blockchain/... ./services/p2p/...`, and `go test` / `go test -race` over those three packages are all green.

icellan:
The prior analysis holds: at the shipped defaults (10000 / 24h / 1h) this is safe. A connected peer republishes `node_status` every ~10s, which refreshes `LastSeen` via `Register`, and the LRU phase can't structurally evict an actively-connected peer (it's exempt while its activity is within TTL) — four orders of magnitude of slack. This commit doesn't fix a live bug; it guards the tuning range this PR newly opens up.

**Floor added:** `p2p_peer_registry_ttl` below 5 minutes is now coerced up to 5 minutes at startup, logged as a warning naming both the configured and effective values (`resolvePeerRegistryTTL` in `services/blockchain/Server.go`). Reasoning: a connected peer's registry entry is only refreshed when a `node_status`/gossip update reaches the registry, and the gossip-handler batcher (`p2p_peer_registry_batch_interval`-driven) can coalesce that for up to a minute (`registryReassertTTL`) before re-asserting. 5 minutes is 5x that batcher window and 30x the publish interval — enough margin for a missed beat or two of either without meaningfully narrowing the operator's tuning range (still four orders of magnitude below the 24h default).

**`maxSize` and `cleanupInterval` are untouched, deliberately.** Both `<=0` are documented, permanent opt-outs (`Cleanup`/`StartCleanup` doc comments): 0 disables LRU entirely, a non-positive interval disables the cleanup loop (TTL-on-load only). Neither has a dangerous-but-positive range: a connected peer is exempt from LRU regardless of how small `maxSize` is, and a short-but-positive `cleanupInterval` only costs extra sweep CPU/lock time, not correctness — the TTL exemption for connected peers doesn't depend on sweep frequency. So only the TTL gets a floor.

**Tests added:**
- `services/blockchain/peer_registry_test.go`: `TestCentralizedPeerRegistry_Cleanup_LRUAtProductionScale` — `Cleanup(10000, 24h)` with 10010 entries, asserting strict oldest-activity-first eviction and that a connected peer survives.
- `services/blockchain/server_test.go`: `TestResolvePeerRegistryTTL_FloorsTooShortValue` (30s → coerced to 5m, warning logged and asserted), `TestResolvePeerRegistryTTL_LeavesLegitimateSentinelAlone` (0 → 24h default, no warning), `TestResolvePeerRegistryTTL_LeavesValidValueAlone` (7m passes through unchanged).
- Also tightened a pre-existing flaky test (`TestCentralizedPeerRegistry_StartCleanup_WiredFromSettings`): its 10ms TTL raced against however long the test scheduler took to reach the first cleanup tick, since `UpdateConnectionState` doesn't refresh `LastSeen` and the connected peer was never touched again after registration. Backdating the stale peer immediately makes eviction deterministic on the first sweep instead of depending on wall-clock timing.

`go build ./...`, `go vet`, `go test` and `go test -race` all green over `settings/...`, `services/blockchain/...`, `services/p2p/...`. `golangci-lint run` clean.

ordishs (approved):

PR #1646, opened 2026-08-21
fix(settings): wire mmap-backed subtree and disk-backed TxMap dir settings

icellan:
## Summary

- `blockassembly_subtreeMmapDir`, `blockassembly_txMapDirs` and `blockvalidation_subtreeMmapDir` had `key:` struct tags but `settings.NewSettings()` never populated them, so the mmap-backed-subtree / disk-backed TxMap block-RAM-reduction feature could not be enabled by configuration regardless of what the consuming code did with it.
- Wires all three fields into `settings/settings.go`, matching sibling style (`getString` for the two directory strings, `getMultiString` with the existing `|`-separated convention for the directory list).
- Defaults stay empty, so production behaviour is unchanged unless an operator explicitly configures a directory.
- Adds fail-fast startup validation (new `util.ValidateWritableDir`, verified by creating and removing a real file rather than just checking permission bits): `NewBlockAssembler` and `blockvalidation.Server.Init` now return a configuration error if a configured directory does not exist or is not writable, instead of relying on a downstream silent heap fallback that gives no signal the feature never actually turned on.
- `blockvalidation_subtreeMmapDir` is read on every subtree load (`get_blocks.go`'s `fetchAndStoreSubtree`, not just once at construction like the block-assembly pair), so its startup validation matters more; the existing runtime mmap-to-heap fallback in `subtreeFromBytesWithMmap` now logs a warning instead of failing silently, so a later transient failure (e.g. disk fills up after startup) stays visible without aborting the block.
- Documents all three settings, including the RAM/disk tradeoff and the multi-directory striping behaviour of `TxMapDirs` (BadgerDB shard chosen by hashing the tx hash across the configured paths), under `docs/references/settings/services/`.
- Resets these three settings in the shared test-settings helpers (`util/test/helpers.go`, `test/nodeHelpers/blockchainDaemon.go`, `daemon/test_daemon.go`) so the shared test suite never depends on directories that only exist on one developer's machine (this surfaced while validating the fix, since a local `settings_local.conf` had these keys set for manual testing of the feature).

## Test plan

- [x] `settings/mmap_dir_settings_test.go` — config reaches each of the three fields via `NewSettings()`, both default (off) and env-override.
- [x] `services/blockassembly/subtreeprocessor/options_test.go` — `WithMmapDir`/`WithTxMapDirs` actually set the processor's fields.
- [x] `services/blockassembly/mmap_dir_wiring_test.go` — a configured, writable directory reaches `NewBlockAssembler` without error; a missing/unwritable directory fails construction.
- [x] `services/blockvalidation/init_mmap_dir_test.go` — same fail-fast contract for `blockvalidation_subtreeMmapDir` via `Server.Init`.
- [x] `services/blockvalidation/quick_validate_mmap_test.go` — new `TestReadSubtree_MmapEnabled_MatchesHeapResult` proves block validation with a configured mmap dir produces an identical subtree/subtree-data result to the heap-only path.
- [x] `util/writable_dir_test.go` — the new validator against empty/valid/missing/not-a-directory/not-writable directories.
- [x] `go build ./...`, `go vet ./...`, `staticcheck`, and `golangci-lint run --new-from-rev origin/main` all clean.
- [x] New tests run `-race -count=10`, stable.
- [x] Full `go test [-race] ./services/blockassembly/... ./services/blockvalidation/... ./settings/... ./util/...` green.


ctnguyen (approved):
Review at commit [0230f10](https://github.com/icellan/teranode/commit/0230f1093e685b9f98f7eaa9739301f9d03f2ecf)

This is a clean fix for a real dead-config bug: three fields carried `key:` tags that `NewSettings()` never populated, so the off-heap subtree/TxMap feature was unreachable no matter how it was configured. The wiring matches sibling style, the docs are unusually good (the striping explanation and the RAM/disk tradeoff are genuinely useful to an operator), and `util.ValidateWritableDir` proving writability by actually creating and removing a file rather than reading permission bits is the right call — it catches read-only mounts that a stat-based check would wave through. Replacing the silent heap fallback with a logged warning is the other good instinct here: a feature that quietly doesn't engage is worse than one that complains. Test coverage is proportionate — option-level, settings-level, construction-level, plus a behaviour-neutrality test that the mmap path yields an identical subtree to the heap path. Nothing below is a correctness problem; the notes are about upgrade disclosure and operator ergonomics.

## Non-blocking issues

### ChiR1 — Newly-live keys make an existing config value startup-fatal, undisclosed

**Problem:** These three keys had `key:` tags but were never populated, so any value an operator already has for them has been inert. This PR makes them live (`settings/settings.go:315`, `:316`, `:413`) and adds fail-fast validation at three call sites (`services/blockassembly/BlockAssembler.go:245-260`, `services/blockvalidation/Server.go:647-649`). One unchanged config file therefore produces two changes in a single upgrade: the feature turns on, and startup becomes fatal if the directory is missing or unwritable. The PR body says "production behaviour is unchanged unless an operator explicitly configures a directory", which holds for the empty-config case and does not cover the case where a value is already present but was never read. I make no claim about how many deployments carry such a value — nobody in this review knows that. The PR's note that a local `settings_local.conf` had these keys set for manual testing is an illustration that configurations of this shape get written, not evidence of production prevalence.

**Why it matters:** The change reads as behaviour-preserving, so an operator has no prompt to audit these keys before restarting. Fail-fast itself is a defensible design choice, and the comment at `services/blockvalidation/Server.go:643-646` argues for it well — a silent heap fallback on a per-subtree path gives no signal the feature never engaged. The gap is disclosure, not design: no release note, no deprecation window, and no warn-and-continue option.

**Fix:** Add an upgrade note stating that these three keys become live in this release and that a stale or now-invalid value will block startup, and adjust the PR body's "behaviour is unchanged" wording to name that case. If a softer landing is wanted, log the value before validating so the failing key is named in the log even as the process exits:

```go
if dir := u.settings.BlockValidation.SubtreeMmapDir; dir != "" {
    u.logger.Infof("[Init] blockvalidation_subtreeMmapDir is configured (%s); off-heap subtree loading enabled", dir)

    if err := util.ValidateWritableDir(dir); err != nil {
        return errors.NewConfigurationError("blockvalidation_subtreeMmapDir is not usable", err)
    }
}
```

**Criticality:** 35%

### ChiR2 — `TxMapDirs` accepts the same directory twice, subdividing the shard set

**Problem:** The validation loop at `services/blockassembly/BlockAssembler.go:253-257` checks each entry independently, so `blockassembly_txMapDirs=/mnt/nvme0/txmap|/mnt/nvme0/txmap` passes — both entries exist and are writable — and the duplicated list reaches `subtreeprocessor.WithTxMapDirs` at `:259`. Downstream, `services/blockassembly/subtreeprocessor/disk_tx_map.go:93-100` takes `numDisks: len(paths)` straight from that list, `:122-129` builds one `diskShard` per path with its own store, its own write batch, its own `writeCh` sized `writeChBuffer/len(paths)`, and a distinct `prefix` of `fmt.Sprintf("%s-disk%d", prefix, i)`, and `:186-188` routes by `diskOf(hash) = int(shardOf(hash)) % m.numDisks`. A duplicated path therefore produces several independent logical shards rooted at the same directory: the hash space is divided as if there were N disks when fewer physical devices back it, and each duplicate takes its own slice of the write buffer. The distinct per-shard prefix means this is not a correctness problem — the shards do not collide.

**Why it matters:** The striping described in the PR body ("BadgerDB shard chosen by hashing the tx hash across the configured paths") becomes misleading rather than absent: the operator sees N configured directories and reasonably expects roughly N× write throughput, while the extra shards contend for one device and the write buffer is split more finely than the hardware warrants. This validation block exists precisely to turn misconfiguration into a startup error, and a repeated path is the most likely hand-editing slip in a `|`-separated list.

**Fix:** Dedupe-check in the same loop, on the cleaned path so `/mnt/x` and `/mnt/x/` collide:

```go
seen := make(map[string]struct{}, len(tSettings.BlockAssembly.TxMapDirs))

for _, dir := range tSettings.BlockAssembly.TxMapDirs {
    clean := filepath.Clean(dir)
    if _, dup := seen[clean]; dup {
        return nil, errors.NewConfigurationError("blockassembly_txMapDirs contains duplicate directory %s", clean)
    }

    seen[clean] = struct{}{}

    if err := util.ValidateWritableDir(dir); err != nil {
        return nil, errors.NewConfigurationError("blockassembly_txMapDirs is not usable", err)
    }
}
```

**Criticality:** 25%

### ChiR3 — Per-subtree mmap fallback warning has no rate limit or counter

**Problem:** The new warning at `services/blockvalidation/BlockValidation.go:307-309` fires inside `subtreeFromBytesWithMmap`, which is on a per-subtree-load path (`fetchAndStoreSubtree` calls it for each subtree, `services/blockvalidation/get_blocks.go:464`). If the configured directory becomes unwritable after startup — the disk filling during catchup is the case the comment names — every subtree in every block emits its own warning line, with no aggregation and no metric.

**Why it matters:** A sustained degraded state should be one actionable signal, not a line per subtree. High-volume repeated warnings are exactly what gets filtered out of log pipelines, and with no counter there is nothing to alert on, so the feature can silently be running heap-only at scale.

**Fix:** Rate-limit or aggregate — a first-failure warning (`sync.Once` on the `BlockValidation` instance, or a periodic sampled log) plus a Prometheus counter for mmap-fallback events gives an alertable signal without the volume.

**Criticality:** 15%

## Recap

| --- | --- | --- | --- |
| ChiR1 | Newly live keys undisclosed upgrade risk | | 35% |
| ChiR2 | Duplicate txMapDirs entries accepted | | 25% |
| ChiR3 | Unthrottled per-subtree fallback warning | | 15% |

icellan:
Rebased onto current `main` (was on `0230f1093`, now on `26d1f5ed1`).

Two conflicts, both clean unions (adjacent additions, not competing implementations):

- `settings/settings.go`: this PR's `BlockAssembly.SubtreeMmapDir`/`TxMapDirs` population lines conflicted with main's unrelated `MaxQueueItems`/`QueueFullWaitTimeout` addition to the same struct literal. Kept both.
- `services/blockvalidation/Server.go` (new conflict, not present in the last rebase in this queue): main added a comment above the `NewBlockValidation` construction explaining the P2P-client wiring for corrupt-block peer attribution (#1390-era). This PR inserted the fail-fast `ValidateWritableDir` check + comment at the same spot. Kept both — the mmap validation runs first, then the existing "only create if not already set" construction with main's updated comment.

No competing mmap/writable-dir implementation appeared on main — checked `git log` for any other `ValidateWritableDir` or mmap-startup-validation work; none found.

Confirmed present after rebase: `blockassembly_subtreeMmapDir`, `blockassembly_txMapDirs`, `blockvalidation_subtreeMmapDir` wiring; fail-fast `util.ValidateWritableDir` calls in `blockvalidation.Server.Init` and `blockassembly.NewBlockAssembler`; the `Warnf` on runtime mmap-to-heap fallback in `subtreeFromBytesWithMmap`; and the three test-settings resets (`test.CreateBaseTestSettings`, `nodeHelpers.NewBlockchainDaemon`, `daemon.NewTestDaemon`).

Verification:
- `go build ./...` clean.
- `go vet ./settings/... ./services/blockassembly/... ./services/blockvalidation/... ./util/...` clean.
- Full `go test` over that set: all green except a pre-existing failure in `services/blockvalidation` (`TestValidateBlock_UnusableHeaderRunIsRebuiltByHashWalk`, from `header_run_walk_fallback_test.go`, added by #1481) — reproduced identically on a clean `origin/main` checkout with no changes from this PR, so it's unrelated to this rebase.
- The mmap-specific tests (`settings`, `services/blockassembly`, `services/blockassembly/subtreeprocessor`, `services/blockvalidation`, `util`) at `-race -count=10`: all green, no flakes.

New head: `26d1f5ed1`.

ordishs (changes requested):
**services/blockassembly/subtreeprocessor/SubtreeProcessor.go:4442-4450, 4689, 4713** — P0

- `moveForwardBlock` captures `originalCurrentSubtree` (4689) and `chainedSubtrees` (4671). `processBlockSubtrees` returns the same pointers `stp.chainedSubtrees` holds.
- `resetSubtreeState` then closes all of them: `cs.Close()` (4442) and `stp.closeChainedSubtrees()` (4450).
- `moveForwardBlock` passes those closed subtrees into `processRemainderTransactionsAndDequeue` (4713), which reads `currentSubtree.Nodes` and `subtree.Nodes`.
- `Close()` is a no-op for a heap subtree, so this is harmless today. For an mmap subtree it calls `munmap` and unlinks the backing file (go-subtree v1.4.2 `mmap_store.go:26-40`), while `Nodes` still points into the unmapped region.
- Failure: set `blockassembly_subtreeMmapDir`; block assembly moves forward; `processOwnBlockSubtreeNodes` reads unmapped memory; the process takes SIGSEGV. This is the success path, not an error path. The rollback closures at 830-831, 3270-3271 and 3334-3335 reinstate the same closed subtrees.
- Fix: close the subtrees at the commit point, after remainder processing finishes and the rollback defers are disarmed.

**services/blockvalidation/get_blocks.go:464, 713-718** — P1

- `fetchAndStoreSubtree` returns an mmap-backed subtree. `fetchSubtreeAndDataFromPeer` passes it to `fetchAndStoreSubtreeData`, which only reads it, never stores it and never closes it. It is a local, so `model.Block.ReleaseSubtreeNodes` never sees it.
- Failure: set `blockvalidation_subtreeMmapDir`. During catchup the `localExists` branch (456) is the common case, so each load leaks one mapping and one `subtree-nodes-*` file of `capacity * 48` bytes. After ~65530 subtrees the process hits the default `vm.max_map_count`; every later mmap fails and the new warning at `BlockValidation.go:308` fires per subtree, blaming disk or permissions. The files survive a restart, because only `Close()` unlinks them.
- Fix: deferred `Close()` in `fetchSubtreeAndDataFromPeer`. The quick-validate path already does this correctly via `SubtreeProcessingBatch.Close()` (`quick_validate.go:1056`), so catchup is the only gap.

**services/blockassembly/subtreeprocessor/SubtreeProcessor.go:4417-4421** — P1

- The in-memory path preserves rollback data with a double buffer (`currentTxMap` / `currentTxMapShadow`, undone by `swapCurrentTxMapBack`). The `DiskTxMap` path instead calls `Clear()` in place, which cancels the write batch, rebuilds every cuckoo filter and opens a new Badger store. No copy survives.
- The rollback closure restores the same, now empty, pointer. `swappedHere` stays false on the disk branch, so the defer at 4411-4415 does nothing either.
- Failure: set `blockassembly_txMapDirs`. `resetSubtreeState` drops every TxInpoints entry for the in-flight transactions. A later `moveForwardBlock` step fails. Rollback reinstates chained subtrees that still list thousands of hashes, against an empty map. Every later lookup misses.
- Fix: defer the `Clear()` to commit, or keep two Badger generations.

**services/blockassembly/BlockAssembler.go:252-257** — P1

- The new loop validates each `TxMapDirs` entry with `util.ValidateWritableDir`, which returns nil for `""` because empty means "feature disabled" (`util/writable_dir.go:17-20`). That rule does not hold per list entry.
- `getMultiString` splits on `|` and only trims; it does not drop empty items (gocore v1.0.82 `config.go:700-704`). So an empty entry passes the new fail-fast check.
- Failure: `blockassembly_txMapDirs = /mnt/nvme0/txmap|` (trailing pipe, doubled pipe, or a whitespace-only entry). `NewDiskTxMap` gets `["/mnt/nvme0/txmap", ""]` with `numDisks = 2`. `tempstore.New` maps an empty base path to `os.TempDir()` (`stores/tempstore/badger.go:42-44`), so `diskOf` routes about half the inpoints to a Badger store under `/tmp` — a RAM-backed tmpfs on a typical node. This inverts the setting's purpose, with no error and no warning.
- Fix: reject an empty entry explicitly before validating. Add a test with `TxMapDirs = []string{t.TempDir(), ""}`.

joe-legalist (approved):

joe-legalist (approved):

PR #1647, opened 2026-08-21
fix(blockassembly): correct two claims the stall signal made about itself, and de-flake its park test

freemans13:
Follow-ups on #1520, which merged earlier today. A review pass over that branch turned up three things after the approvals had already landed — none of them a behaviour bug in production, but two of them are wrong claims in comments and one is a test that can fail for a reason unrelated to what it tests. Splitting them out here rather than leaving them, since the whole point of that change was to make a signal an operator can trust, and a docstring that misdescribes how the level is chosen undermines exactly that.

Worth noting up front: Copilot's last two review runs on #1520 both came back "unable to review this pull request because the user who requested the review has reached their quota limit", so the final commits went in without automated review. That is what prompted the extra pass.

### The docstring said depth picks the level. For half the arms, it does not.

`reportDequeueStall`'s comment claimed a clean split: the lifecycle state picks the message, and whether the queue has held work picks the level. That is true of only two of the four arms. A consumer that has not started reports at info however much is queued behind it, because that is routine startup either way; a consumer that has exited reports at error even on an empty queue, because the service is dead regardless. Only the two middle arms consult the `sawQueuedWork` latch.

The behaviour was already right — this is the description of it that was wrong, and it is the description someone will read before changing the severity rules.

### The reported stall duration is floored, and the comment no longer blames NTP for it.

The closing edge computes `(now - staleness) - stalledSince` and used to carry a comment asserting the result "cannot go negative: staleness here is at most the threshold, and the staleness that opened the incident exceeded it". That claim is right, but it rests on an assumption it never states — that the clock does not run backwards between the two ticks — so the first pass here replaced it with an NTP-step argument and a floor.

**The NTP argument was itself wrong, and a second pass caught it.** `now` comes from `time.Now()`, `Time.Add` carries the monotonic reading through, and `Time.Sub` between two readings that both carry one ignores the wall clock entirely. The running service cannot report a negative duration here, whatever NTP does. Verified rather than assumed: `Add(-100 years)` still keeps the monotonic reading, and a `Sub` with the wall clock stripped out of both operands returns the same answer as one without.

The floor stays, and the comment now says which caller the hazard actually belongs to. `observeDequeueStall` is pure specifically so the transition table can drive it with `time.Date` values — which carry no monotonic reading, and so put the wall clock back in charge. Handing a caller "was stalled for -5m0s" for an input the table can construct is not worth the two lines saved. Pinned by a new case that does exactly that; delete the floor and only that case fails.

### The park test could fail without anything being wrong.

`TestLastDequeueTime_StopsAdvancingWhileConsumerParked` confirmed the consumer had actually parked by waiting for `LastDequeueTime` to fall 150ms behind, reasoning that the consumer stamps once per loop iteration and idles only `IdleSleepDuration` (10ms). That is a timing guess, and roughly fifteen missed iterations satisfy it — which CPU starvation during this package's three-minute race run can produce with the park not having happened at all. The test then captures `stalledAt` against a still-live consumer, which stamps during the following 300ms sleep, and the central assertion fails. A false red, not a false green, but a flake in a test whose whole job is to be believable about a stall.

`lengthCh` is unbuffered, so the test's send returns only once the consumer has executed `case lengthCh := <-stp.lengthCh`, and from there it is committed to the unbuffered `respCh` send with nothing in between that can stamp the timestamp. Closing a channel when the send returns turns the wait from a guess into a handshake.

## Testing

`go build ./...`, `go vet ./services/blockassembly/...`, and `go test -race -count=1 ./services/blockassembly/...` — all three packages green.

Whole-package `golangci-lint run ./services/blockassembly/...` reports four `prealloc` hints, all in files neither this PR nor #1520 touches: three in `BlockAssembler_test.go` and `load_unmined_disk_sort_test.go`, one in `mining/BuildBlockHeader.go`. Pre-existing; left alone. `golangci-lint run` over just the two files this PR changes is clean.

**Mutation-verified where it is possible.** Removing the negative floor fails the new backwards-clock case and nothing else.

**Honest limit on the third fix: it is not pinned by a failing test.** Deleting the handshake wait entirely leaves the test passing, because in the ordinary case the sending goroutine wins the race and the consumer parks before the enqueue anyway. The wait only matters when it does not, which is the starvation case that cannot be provoked on demand. So its correctness rests on the unbuffered-channel argument above rather than on a test that goes red — which is the same reason the old guard's flakiness never showed up in CI either.

I also re-checked the concurrency claims of #1520 while in there, and they hold: every input the detector samples is a plain atomic load, so none can block on the select loop being diagnosed; `GetCurrentLength`, the channel round-trip that would, appears nowhere in that path; and the `ConsumerStarted`-before-`LastDequeueTime` read order is right in the shipped code and does fail its test when swapped. The unchecked `atomic.Value` assertion in `GetCurrentRunningState` is safe because `NewSubtreeProcessor` stores a state before returning and is the only construction site.

https://claude.ai/code/session_01QajtAi7pwfsEQfnMdYeARo


ordishs (approved):
All three claims check out against the code, and the two behaviour-relevant ones are verified rather than taken on trust.

- The new `reportDequeueStall` docstring matches the four switch arms. `!consumerStarted` logs info, `ConsumerExited()` logs error, and only the `sawQueuedWork` and `default` arms read the latch. The old text was wrong.
- `go test -race -count=1 -timeout 500s ./services/blockassembly/...` — ok on all three packages (327s / 3s / 156s).
- Mutation check on the floor: with `if stalledFor < 0` removed, exactly one test fails — `TestObserveDequeueStall_Transitions/backwardsClockDoesNotReportANegativeDuration`, `expected: 0s, actual: -5m1s`.
- The park handshake is sound. `lengthCh` is unbuffered, and `SubtreeProcessor.go:981-983` has nothing between the receive and the unbuffered `respCh` send that stamps `lastDequeueMillis`.

Two comment-accuracy points, neither blocking. They are prose, but this PR exists to fix comment accuracy, so they belong in the same pass.

**services/blockassembly/dequeue_stall.go:158-161**

The comment credits the non-negativity to the monotonic clock alone: "not because the wall clock behaves ... so an NTP step cannot invert it". The conclusion is true, the reason incomplete.

`staleness` is a wall-clock quantity. `Server.go:347` computes `now.Sub(ba.blockAssembler.LastDequeueTime())`, and `SubtreeProcessor.LastDequeueTime` returns `time.UnixMilli(stp.lastDequeueMillis.Load())` (`SubtreeProcessor.go:1972-1974`), which carries no monotonic reading. `Time.Sub` therefore falls back to the wall clock for that subtraction, so a clock step does reach the result through `staleness`, even though it cannot reach the final `Sub`.

Write the difference as `(m2 - staleness2) - (m1 - staleness1) = (m2 - m1) + (staleness1 - staleness2)`. The monotonic readings give `m2 >= m1`. The branch condition gives `staleness2 <= dequeueStallThreshold`, and the rising edge required `staleness1 > dequeueStallThreshold`, so `staleness1 - staleness2 > 0`. The result stays positive for any clock step. The guarantee needs both facts, and the comment now credits only the one it added.

Suggestion: keep the monotonic sentence, and restore the threshold half the old comment had. That is the part a clock step cannot touch. Line 165, "restore every wall-clock hazard the service itself is immune to", overstates for the same reason — the service is immune in the final subtraction, not in `staleness`.

**services/blockassembly/Server.go:391**

The retained paragraph still opens "The level comes from the incident-wide latch, not from this tick's depth." The paragraph immediately above now establishes that two of the four arms never read the latch, so read in sequence this reads as a universal rule the new text has just qualified. Narrowing it to the arm it describes would close the gap.

On the third fix: the note that it is not pinned by a failing test is accurate, and acceptable. The replacement is strictly stronger than what it removes — the old guard proved a staleness gap, the new one proves the consumer reached a specific select arm — and the argument rests on unbuffered-channel semantics plus the absence of a stamp between `SubtreeProcessor.go:981` and `:983`, both checkable by reading rather than by timing.

ctnguyen (approved):
Review at commit [379bf86](https://github.com/freemans13/teranode/commit/379bf86dd89e4cb5f0d7680be7494dc126beeaac)

Nice follow-up to do rather than shrug at. Splitting three small honesty fixes out of a merged branch instead of leaving them is the right instinct, especially for a signal whose whole value is that an operator believes it. The `reportDequeueStall` docstring rewrite is a genuine correction — the four arms at `Server.go:405-452` really do split two-fix-their-own-level / two-read-the-latch, and the old text claimed otherwise. The floor in `observeDequeueStall` is small, pinned by exactly one new table case, and honestly motivated by the function being pure. And swapping the 150ms staleness guess in the park test for a close-on-send handshake is strictly stronger than what it removes; the unbuffered-`lengthCh` argument is checkable by reading `SubtreeProcessor.go:981-983`. Two comment-level notes below, neither blocking.

## Non-blocking issues

### ChiR1 — Non-negativity argument credits the monotonic clock but drops the threshold half

*Same as [@ordishs's point 1 (24-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1647#pullrequestreview-5007408091).*

**Problem:** `dequeue_stall.go:154-160` now says the difference "cannot go negative, and not because the wall clock behaves", crediting only the monotonic reading. The conclusion holds, the reason as written does not cover the whole expression. Writing it out: `(now₂ − staleness₂) − (now₁ − staleness₁) = (now₂ − now₁) + (staleness₁ − staleness₂)`. Monotonic gives `now₂ ≥ now₁` for the first term only. `staleness` is a wall-clock quantity — `Server.go:347` computes `now.Sub(ba.blockAssembler.LastDequeueTime())`, and `LastDequeueTime` returns `time.UnixMilli(stp.lastDequeueMillis.Load())` (`SubtreeProcessor.go:1972-1974`), which carries no monotonic reading, so that `Sub` falls back to the wall clock. A step therefore *does* reach the result, through `staleness`. What actually pins the second term is the old comment's threshold argument: this branch requires `staleness₂ <= dequeueStallThreshold` and the rising edge at `:128` required `staleness₁ > dequeueStallThreshold`, so `staleness₁ − staleness₂ > 0`. Extending their point in the running-service direction, that threshold guard is also what disposes of a forward wall-clock step: a forward step inflates `staleness₂`, and an inflated `staleness₂` fails the `staleness <= dequeueStallThreshold` guard, so the ended branch is not entered at all. A backward step pushes the other way — `now` lands earlier than the already-stamped `lastDequeueMillis` and `staleness` goes negative — and that one is caught by the floor at `Server.go:348-350`, though its own comment at `:343-346` motivates it by the sub-tick sampling race (`now` is sampled before the timestamp is read) rather than by clock steps, so a backward step is a second thing it happens to catch rather than its stated purpose. Line 165's "restore every wall-clock hazard the service itself is immune to" overstates for the same reason as the rest: the service is immune in the final `Sub`, not in `staleness`. The same defect appears in miniature in the new table case's own comment — `dequeue_stall_test.go:215` says the negative-duration input is "Reachable only from this table", but `dequeue_stall_tick_test.go` also drives the tick with `time.Date` values (`:58`, `:183`, `:315`, `:368`) and so could construct monotonic-free inputs; it just never walks time backwards. "Constructed only here" would have been exact.

**Why it matters:** This is a comment whose only job is to tell the next person which invariant they must not break. Someone who reads it as "monotonic covers this" could later change the exit condition (say, exit on staleness below half the threshold) believing the subtraction is still protected, when the term that actually protected it came from the threshold relationship the comment no longer mentions. Given the PR's own premise — a docstring that misdescribes the mechanism undermines the signal — the fix that removed a wrong argument shouldn't also drop the correct one it replaced.

**Fix:** Keep the monotonic sentence, restore the threshold half, and soften line 165:

```go
  // Both ends of that subtraction are a clock reading minus the
  // staleness at that reading, so each is the moment of a dequeue, and
  // the later dequeue follows the earlier one. Two facts, not one, make
  // that hold. The monotonic reading covers (now2 - now1): now comes
  // from time.Now(), Add carries the reading through, and Sub between
  // two readings that both carry one ignores the wall clock, so an NTP
  // step cannot invert it. The threshold covers the rest: this branch
  // requires staleness <= dequeueStallThreshold while the rising edge
  // required staleness > it, so (staleness1 - staleness2) > 0. That
  // second half is the one that matters, because staleness itself is a
  // wall-clock quantity - Server.go derives it from LastDequeueTime,
  // a time.UnixMilli value with no monotonic reading. A forward step
  // inflates staleness and so fails the guard above rather than
  // reaching here; a backward one drives it negative and is floored at
  // the caller.
  //
  // Floored anyway, because the clock here is only a parameter. This
  // function is pure precisely so the table test can drive it with
  // time.Date values, which carry no monotonic reading and can walk
  // backwards in a way the service's own inputs cannot.
```

and while you are in the test, `dequeue_stall_test.go:215` reads better as "Constructed only here".

### ChiR2 — Retained "the level comes from the latch" paragraph now reads as a rule the text above just qualified

*Same as [@ordishs's point 2 (24-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1647#pullrequestreview-5007408091).*

**Problem:** `Server.go:392-395` still opens "The level comes from the incident-wide latch, not from this tick's depth." That sentence was accurate against the old paragraph it followed, which claimed depth picks the level for every arm. The rewritten paragraph directly above (`:380-390`) now establishes the opposite for half the switch: `!consumerStarted` at `:406` reports info regardless of `queueLength`, and `ConsumerExited()` at `:425` reports error regardless — neither reads `sawQueuedWork` at all. So read top to bottom, the diff fixes the "depth decides the level" overclaim and then immediately follows it with an unscoped "the latch decides the level" claim about the same four arms. Same defect class the PR exists to remove, in the paragraph directly beneath the one it rewrote.

**Why it matters:** The paragraph is load-bearing for a change nobody should make: it's the recorded reason not to swap `sawQueuedWork` for a live depth read, which is explicitly wrong because `dequeueDuringBlockMovement` drains the queue from inside the branch blocking the consumer (settled on [bsv-blockchain/teranode#1520](https://github.com/bsv-blockchain/teranode/pull/1520)). An unscoped claim sitting under text that just contradicted it invites a reader to discount both, which is the worst outcome for the one warning that shouldn't be weakened.

**Fix:** Scope it to the arm it actually describes:

```go
  // For that third case the level comes from the incident-wide latch, not
  // from this tick's depth. Depth is a single sample of a number the
  // blocking handler is itself draining, so keying the level on it directly
  // would let a live incident report at info on the ticks that land just
  // after a drain.
```

## Recap

| --- | --- | --- | --- |
| ChiR1 | Non-negativity reason incomplete | | 35% |
| ChiR2 | Unscoped latch claim retained | | 25% |

PR #1651, opened 2026-08-24
p2p: make websocket notification fan-out non-blocking and add drop/eviction metrics

gokutheengineer:
**Closes bitcoin-sv/teranode#4752.**

### Problem

Of the three findings in the issue, two were already fixed on main by the recent websocket hardening:

- Broadcast-timeout removal turning slow clients into permanent zombies: `evict()` now cancels the per-connection context, and the handler tears the socket down.
- The fd leak on the `<-done` teardown branch: the handler now closes the websocket in a single `defer` covering every exit path.

What remained valid from the first finding:

- The notification fan-out still ran inline on the single notification processor goroutine, spawning a goroutine per client behind a 256-slot semaphore and waiting up to 1s per slow client before `wg.Wait()` returned. A pool of slow readers (each keeping its 100-slot buffer full but accepting sends just under the 1s timeout) stalls every broadcast by ceil(N/256) seconds indefinitely, backing the 1000-slot `notificationCh` up into silent producer-side drops of block/subtree/node_status notifications.
- Those producer-side drops were logged at Warn only, with no metric.

### Fix

- `clientChannelMap.broadcast` now does a synchronous non-blocking send per client. A client whose buffer is full is evicted immediately via `evict()` (connection context cancelled, socket closed, slot released) instead of being waited on. The per-client buffer is the entire grace a slow consumer gets; the semaphore/timer/goroutine machinery is deleted. A broadcast now costs microseconds regardless of client behavior, so the processor can never be stalled and `notificationCh` cannot back up because of slow websocket clients.
- Added `teranode_p2p_websocket_notifications_dropped_total{type}`, incremented at all four `notificationCh` producer drop sites (block, subtree, node_status x2), and `teranode_p2p_websocket_clients_evicted_total`, incremented on fan-out eviction.

This improves on the issue's proposed fix by removing the 1s timeout entirely rather than moving it off the processor goroutine: the timeout was the attack surface (slow readers riding just under it were never evicted), and client registration/removal already no longer flows through the processor.

**Operator-visible behavior change:** a client with a full buffer previously got up to 1s of grace per notification; it is now evicted on the first broadcast that finds its buffer full. An honest dashboard on a briefly congested link that would have survived a burst under the old grace can now be disconnected (it reconnects and re-receives the initial node_status). The eviction metric makes this visible. Eviction logging is a single summary line per broadcast (count of evicted clients), not one line per client, so a mass-eviction event cannot stall the fan-out on a slow log sink.

Docs updated: `docs/topics/services/p2p.md` (fan-out description) and `docs/references/prometheusMetrics.md` (new metrics).

### Tests

- `TestBroadcastMessage` rewritten to exercise the real `broadcast`: responsive clients receive, backed-up clients are evicted and their connection cancel funcs invoked.
- `TestBroadcast_NonBlockingFanout` added: 64 backed-up clients plus one responsive client; asserts the fan-out completes far below the old 1s floor, every backed-up client is evicted and cancelled, the eviction metric counts each one, and the responsive client still receives.
- `TestBroadcast_BoundedPool` and `TestBroadcast_NonPositivePoolSizeDoesNotDeadlock` removed with the semaphore machinery they tested.
- Existing regression tests (`TestBroadcast_SequentialTimeoutDoS`, `TestHandleWebSocket_BroadcastEvictionClosesConnection`, eviction/cap/deadline suites) pass unchanged in behavior.

Ran: `SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/p2p/...` - ok.

### Residual

- A fresh-eyes review noted two pre-existing test-hygiene nits left as-is (out of scope for this change): some websocket tests don't explicitly join their background notification-processor goroutines, and `TestHandleWebSocket_BroadcastEvictionClosesConnection`'s flooding goroutine has no explicit join. Neither can affect correctness or the new metric assertions.


liam (approved):
Reviewed against a checkout rather than the diff alone, and mutation-tested the new tests rather than trusting that they pass. This is good work. Approving, with two minor points and one design thought that is explicitly not a request.

The reasoning in the description is the part I most agree with: the 1s timeout was the attack surface, not the mitigation. A slow reader riding just under it was never evicted and cost every other client a full second per broadcast. Removing it outright is the right call, and doing the eviction synchronously means the processor goroutine can no longer be stalled by client behaviour at all.

## What I verified

The three hazards this pattern usually has are all absent, and I checked each rather than assuming:

- **Send on a closed channel.** Snapshotting under `RLock` and sending outside it would panic if a client channel were closed concurrently. Client channels are never closed anywhere in the file (only `readDone` / `writeDone` are); a removed channel is simply abandoned to GC. Safe.
- **Lock re-entry.** `broadcast` releases the read lock before the send loop and `evict` takes the write lock, so no deadlock.
- **Cold-start eviction.** A client registered before its drainer exists would be evicted by the first burst. `handleClientMessages` is started before `clientChannels.add()`, and `sendInitialNodeStatuses` queues into the buffer before registration, so there is no window.

All four `notificationCh` producer drop sites are instrumented (`server_helpers.go` block and subtree, `Server.go` both node_status paths), label cardinality is bounded to three values, the metric names follow the `_total` convention with the right namespace and subsystem, and the docs table matches the declarations.

On the tests: replacing `cm.evict(ch)` with a silent drop fails `TestBroadcastMessage/Some_backed-up_clients`, `TestBroadcast_NonBlockingFanout` and `TestBroadcast_SequentialTimeoutDoS`. That last one runs in 0.15s normally and 7s mutated, so it is genuinely guarding the behaviour rather than passing incidentally. `go test -race -tags testtxmetacache ./services/p2p/...` ok, `golangci-lint run --new-from-rev upstream/main` 0 issues, and no references to the removed `maxConcurrentBroadcasts` machinery survive anywhere in `services/` or `docs/`.

## Minor

**Eviction accounting counts clients that had already left.** `evict()` returns nothing, so `evicted++` fires unconditionally:

```go
default:
    cm.evict(ch)
    evicted++
```

If a backed-up client disconnects normally between the snapshot and the send, its handler has already run `clientChannels.remove(ch)`, so `evict` finds `exists == false` and does nothing, but the broadcast still counts it and logs `Evicted 1 websocket clients with full send buffers and closed their connections`. A normal disconnect is then reported as an eviction, in the log and in the metric that exists to make evictions visible. Narrow race, easy fix: have `evict` return whether it removed anything and gate the increment on that.

**Log level.** `Errorf` for the eviction summary. A slow client is not a node fault, and since this policy evicts more readily than the old one did, it will be louder than before. `Warnf` reads more accurately. Entirely subjective, ignore if you disagree.

## Design thought, not a request

The buffer is 100 and eviction is now on the first broadcast that finds it full, which the description is admirably direct about. Subtrees are the high-rate notification type, so a dashboard on a congested link can plausibly fall 100 behind during a burst and be flapped. The reconnect path makes that survivable and the new metric makes it visible, which is the right way round.

If it does show up in the metric, a middle ground that keeps the fan-out strictly non-blocking is a per-channel strike counter: evict on the Nth consecutive full-buffer broadcast rather than the first. That is not a timeout, it stays O(1) with no waiting, and it tolerates bursts while still evicting a genuinely stuck client promptly. Worth having in mind for later rather than doing here.

Nothing above blocks merge.

ctnguyen (approved):
Review at commit [e23b74b](https://github.com/gokutheengineer/teranode/commit/e23b74b4f63e9004ed3547db8939fb0b52a8149b)

This is a clean, well-argued simplification. Deleting the semaphore/timer/goroutine machinery in favour of a plain non-blocking `select` is the right call, and the reasoning in the PR body — that the 1s timeout *was* the attack surface, because a reader riding just under it was never evicted — is more convincing than the fix the linked issue originally proposed. The diff is net negative on production lines while closing all three findings in [bitcoin-sv/teranode#4752](https://github.com/bitcoin-sv/teranode/issues/4752), and `liam` has already approved it after mutation-testing the new tests rather than trusting a green run. I also like that the eviction/drop metrics landed in the same change as the policy that makes them matter, that the operator-visible behaviour change is called out explicitly rather than buried, and that the comments at `HandleWebsocket.go:135-143` and `:156-158` record *why* rather than *what*. What I have left is one precision defect in the new eviction metric, below; separately I agree with [@liam's point 2 (24-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1651#pullrequestreview-5008460049) that `Warnf` reads more accurately than `Errorf` at `:162`, but that is taste and I am not filing it.

## Non-blocking issues

### ChiR1 — Eviction counter and log count clients that had already disconnected

*Same as [@liam's point 1 (24-08-2026)](https://github.com/bsv-blockchain/teranode/pull/1651#pullrequestreview-5008460049).*

**Problem:** `broadcast` snapshots the channel set under `RLock` (`services/p2p/HandleWebsocket.go:127-133`), then releases the lock before the send loop. In the loop's `default` branch it calls `cm.evict(ch)` and increments `evicted` unconditionally (`:150-153`). But `evict` (`:114-123`) returns nothing: it reads `cancel, exists := cm.channels[ch]`, deletes, and only uses `exists` to decide whether to call `cancel`. If a client disconnected normally in the window between the snapshot and its turn in the send loop, its connection handler has already run `clientChannels.remove(ch)` in the teardown defer (`:632`), so `evict` finds `exists == false` and does nothing at all — yet the broadcast still counts it into `prometheusP2PWebsocketClientsEvicted` (`:161`) and into the summary log line (`:162`). Note the deregistered channel is also still a full 100-slot buffer that nobody drains, so the `default` branch is exactly the path such a channel takes: a backed-up client that disconnects on its own gets reported as an eviction. Extending `liam`'s observation slightly: the same window makes the log message wrong in its own terms — it claims the connections were closed, and for these entries no connection was touched.

**Why it matters:** The metric exists so operators can see how often the new no-grace policy disconnects clients. When ordinary disconnects of backed-up clients also land in it, the counter conflates "we cut a client off" with "a client left", which are the two cases an operator most needs to separate, and it makes the number unusable for judging whether the 100-slot buffer is too tight. Worth being straight about the classification: I first raised this as blocking, on the argument that the PR's safety case for a harsher eviction policy rests on this metric being trustworthy. On examination `liam`'s minor classification is the better call. The bug inflates the counter rather than suppressing it, so every genuine eviction still shows up — it degrades precision, it does not defeat the signal, and it cannot hide the thing the metric was added to reveal. It also needs a channel that is both full *and* disconnecting inside a narrow window, and the metric is new in this PR, so no existing alert contract is at stake. Still worth fixing before anyone builds a dashboard on it, but not a merge gate.

**Fix:** Have `evict` report whether it removed anything and gate both the counter and the log on that.

```go
// evict removes a client channel and cancels its connection. Reports whether
// the channel was still registered, so callers do not account for a client
// that had already disconnected.
func (cm *clientChannelMap) evict(ch chan []byte) bool {
	cm.Lock()
	cancel, exists := cm.channels[ch]
	delete(cm.channels, ch)
	cm.Unlock()

	if exists && cancel != nil {
		cancel()
	}

	return exists
}
```

and in the send loop:

```go
		default:
			if cm.evict(ch) {
				evicted++
			}
```

A test for it fits naturally alongside `TestBroadcast_NonBlockingFanout`: register a backed-up channel, `cm.remove` it before broadcasting, then assert the metric delta is 0.

Two notes that are not findings. `initPrometheusMetrics()` inside the `evicted > 0` branch (`:160`) is a `sync.Once.Do` (`services/p2p/metrics.go:20-22`), so after the first call it is an atomic load and cheap enough for this loop — worth knowing rather than changing, though hoisting it to service startup would make the hot path obviously free instead of requiring the reader to go check. And on `liam`'s point 3 (a per-channel strike counter before eviction): I agree with the framing there — shipping the harsher policy with a metric first and tuning on evidence is the right order, and a strike counter is a follow-up if the metric shows honest clients being flapped, not something this PR should carry.

PR #1655, opened 2026-08-25
settings: default maxopsperscriptpolicy to UINT32_MAX, matching bitcoin-sv

### LINKED ISSUE #1654: [FEAT] Raise the MaxOpsPerScriptPolicy default to match bitcoin-sv (UINT32_MAX) now that maxscriptsizepolicy is 100 MB
## Is your feature request related to a problem? Please describe

`settings/settings.go` now defaults `maxscriptsizepolicy` to 100,000,000 (up from 500,000), but `maxopsperscriptpolicy` is still 1,000,000. Both are pushed into the BDK engine (`services/validator/ScriptVerifierGoBDK.go`, `SetMaxScriptSizePolicy` / `SetMaxOpsPerScriptPolicy`) and BDK enforces the op count the same way bitcoin-sv's `EvalScript` does: every opcode above `OP_16` increments `nOpCount`, and exceeding the policy value fails the script with `SCRIPT_ERR_OP_COUNT`. Pushes are free, so the count is bounded above by script size, and for opcode-dense scripts it runs at roughly 0.5 counted ops per byte (for example `OP_1 OP_DROP` repeated, or hash-heavy covenant code).

That means a script above roughly 2 MB of executed opcodes is rejected by Teranode's policy path no matter what `maxscriptsizepolicy` says. The size raise is 200x; the op cap did not move, so the new size default is unreachable for real scripts.

It is also a divergence from the reference implementation. bitcoin-sv's post-Genesis default is `DEFAULT_OPS_PER_SCRIPT_POLICY_AFTER_GENESIS = UINT32_MAX` (`src/policy/policy.h`), so an SV node relays and mines scripts that a default-configured Teranode rejects.

## Describe the solution you'd like

Default `maxopsperscriptpolicy` to 4,294,967,295 (UINT32_MAX), the bitcoin-sv default, and update the field's `longdesc` and `docs/references/settings/policy_settings.md` to say that only non-push opcodes count and that the count is bounded by script size. Operators keep the knob; only the default changes. PR attached.

## Describe alternatives you've considered

* `0` (Teranode already documents 0 as unlimited). Matching SV's finite constant seemed cleaner for cross-implementation parity and avoids a special case.
* Scaling the op cap with `maxscriptsizepolicy` automatically. More moving parts for the same effect, since ops <= bytes always holds.

## Additional context

* DoS angle, stated so it can be weighed: the op count bounds the number of opcodes, not the work. Any single counted op can operate on a stack item up to `maxstackmemoryusagepolicy` (100 MB), so 1,000,000 ops was never a CPU bound; the bounds that actually limit validation work are script size, stack memory, and fees. The duration settings (`maxnonstdtxvalidationduration` etc.) are placeholders per their own `longdesc` and are commented out in `settings/settings.go`, so nothing changes on that axis either way.
* `services/validator/TxValidator_test.go` `TestMaxOpsPerScriptPolicy` shows the cap is enforced through BDK (it sets the policy to 2 and expects rejection).
* The legacy Go engine (`services/legacy/txscript`, `MaxOpsPerScript = 201`) is untouched by this request.
* Not executed against a running Teranode; derived from the BDK wiring above and bitcoin-sv's `EvalScript` op-count rule. Happy to add a functional test if you point me at the right harness (`docs/topics/functionalRequirementTests.md` mentions one for this policy).

Calgooon:
## What changed

* `settings/settings.go`: `maxopsperscriptpolicy` default 1,000,000 -> 4,294,967,295 (UINT32_MAX).
* `settings/policy_settings.go`: struct tag default and `longdesc` updated. The description now states that only non-push opcodes (> `OP_16`) are counted, so the count is bounded above by script size, and records the previous default with what it rejected.
* `docs/references/settings/policy_settings.md`: table row updated.

## Why

`maxscriptsizepolicy` moved to 100 MB but `maxopsperscriptpolicy` stayed at 1,000,000. Both reach the BDK engine (`ScriptVerifierGoBDK.go`), and BDK counts non-push opcodes exactly as bitcoin-sv's `EvalScript` does, failing with `SCRIPT_ERR_OP_COUNT` above the policy value. Opcode-dense scripts run at about 0.5 counted ops per byte, so the old cap rejected anything above roughly 2 MB of executed opcodes regardless of the size policy. bitcoin-sv's post-Genesis default for this policy is `UINT32_MAX` (`DEFAULT_OPS_PER_SCRIPT_POLICY_AFTER_GENESIS`), so this also removes a relay/mining divergence between a default SV node and a default Teranode. Closes #1654.

## Testing

* `gofmt -l settings/` clean; `go build ./settings/` and `go test ./settings/` pass.
* Not run here: the cgo BDK build and `services/validator` tests (no BDK static libs on this machine). `TestMaxOpsPerScriptPolicy` sets the policy explicitly, so it is unaffected by the default.

## Impact / risk

Policy default only; consensus untouched, operators keep the knob. The op count never bounded validation work (one counted op can act on a 100 MB stack item), so raising it does not open a new DoS axis; script size, stack memory and fees remain the limits. Nodes that pin `maxopsperscriptpolicy` in config are unaffected.


ctnguyen (approved):
Review at commit [4ffbc48](https://github.com/Calgooon/teranode/commit/4ffbc48d2beb5104ff0745d0b27f7aabe7bc7387)

This is the right change and it is argued well. I followed the value through to the C++ that consumes it and the parity claim holds: `settings/settings.go:109` and the struct tag at `settings/policy_settings.go:10` now both carry 4294967295, which is exactly `DEFAULT_OPS_PER_SCRIPT_POLICY_AFTER_GENESIS` (`bitcoin-sv/src/policy/policy.h:147`), and the counting site (`bitcoin-sv/src/script/interpreter.cpp:466`) really does skip everything at or below `OP_16`, so the old 1,000,000 cap was biting well before `maxscriptsizepolicy` ever could. The "consensus untouched" claim also checks out — `GetMaxOpsPerScript` returns `MAX_OPS_PER_SCRIPT_AFTER_GENESIS` on the consensus branch (`bitcoin-sv/src/configscriptpolicy.cpp:72-74`) and never reads the policy field, and teranode has no consensus-side counterpart to this key. Nice touch that both definition sites moved together rather than one; that mismatch is an easy one to leave behind. Of the seven touched lines in `settings/settings.go` only one is real — the other six are comment realignment the wider literal forces, so there is nothing to say about them. Worth knowing too, if it helps your confidence about the un-run BDK build: `NewPolicySettings()` returns a zero struct (`settings/policy_settings.go:36-40`), and the C++ setter maps 0 to UINT32_MAX, so the existing GoBDK tests at `services/validator/ScriptVerifierGoBDK_test.go:79` and `:355` have already been exercising the engine at this value. Everything I found is on the operator-facing side of the change rather than in the behaviour.

## Non-blocking issues

### ChiR1 — New default sits exactly on the setter ceiling, so raising the knob panics at startup

**Problem:** 4294967295 is not just bitcoin-sv's default, it is also the largest value the setter accepts. `ConfigScriptPolicy::SetMaxOpsPerScriptPolicy` returns `false` for anything above `MAX_OPS_PER_SCRIPT_AFTER_GENESIS` (`bitcoin-sv/src/configscriptpolicy.cpp:172-176`), and teranode converts that refusal into a hard crash at `services/validator/ScriptVerifierGoBDK.go:167-169`:

  ```go
    if err := se.SetMaxOpsPerScriptPolicy(po.MaxOpsPerScriptPolicy); err != nil {
        panic(err)
    }
  ```

  That runs inside `newScriptVerifierGoBDK`, called from `NewTxValidator` (`services/validator/TxValidator.go:115`), i.e. at validator construction. Nothing between the config file and that call bounds the value: `getInt` passes it straight through (`settings/helpers.go:37`), `PolicySettings.SetMaxOpsPerScriptPolicy` is a bare assignment (`settings/policy_settings.go:66-68`), and there is no validation hook at all — neither `Settings` nor `PolicySettings` has a `Validate` method anywhere in the tree. Before this PR the default was 1,000,000, three orders of magnitude below the ceiling, so upward tuning was safe; now the headroom is zero and *any* upward adjustment is fatal. The off-by-one is easy to hit: `settings/settings.go:99` sets `excessiveblocksize` to 4294967296 — one greater, ten lines above the new line. The field's own trade-off table still lists a `Higher` row as a supported direction.

**Why it matters:** svnode handles the same rejection cleanly — `bitcoin-sv/src/init.cpp:3409-3418` wraps it in `InitError(err)` and refuses to start with the message. Teranode reaches the same outcome through a panic and a stack trace, which is a worse first clue during a config change, and this PR is what moves that edge from "unreachable in practice" to "adjacent to the default".

**Fix:** don't clamp — clamping would diverge from svnode, which rejects. Surface it as a startup error naming the key and the ceiling, and drop the `Higher` row from the trade-off table since it now describes a configuration that cannot exist. The existing `getUint64AtLeast` in this package (`settings/helpers.go:72-80`) is the precedent for validating a policy value at load time rather than deep inside a service; a ceiling-checking sibling would keep the diagnosis in `settings` where the operator can read it.

**Criticality:** 40%

### ChiR2 — The `0` option and the new default are the same value, listed as two different behaviours

**Problem:** the `Values` list in the new `longdesc` keeps `- **0** - Unlimited (use with caution)` alongside the new `- **4294967295** (default)`. Those are the same setting. `bitcoin-sv/src/configscriptpolicy.cpp:180-182` maps a zero input to `MAX_OPS_PER_SCRIPT_AFTER_GENESIS`, which is `UINT32_MAX` (`bitcoin-sv/src/consensus/consensus.h:43`) — the exact number the default now carries. So the list presents one behaviour as two options, and hangs a "use with caution" warning on a value identical to the recommended default while the default line carries no such warning. The repo already leans on this equivalence, incidentally: `NewPolicySettings()` returns a zero struct (`settings/policy_settings.go:36-40`, still marked `// TODO set defaults`), so every test built on it has been running at UINT32_MAX.

**Why it matters:** this text is a live operator surface, not a code comment — `ExportMetadata` copies `longdesc` into the settings registry (`settings/export.go:140`) and the asset service serves it at `services/asset/httpimpl/settings_handler.go:75`. It also bears on the rationale recorded in [teranode#1654](https://github.com/bsv-blockchain/teranode/issues/1654), which rejected defaulting to `0` because matching a finite constant "avoids a special case": there is no special case to avoid, since the C++ treats the two identically. That doesn't change the decision — the explicit constant is still the more self-documenting default, and matching the reference literal is worth something on its own — but the `0` bullet should stop implying a distinct, riskier mode.

**Fix:** reword to say what it actually does, e.g. `- **0** - Accepted and treated as UINT32_MAX by the engine; prefer the explicit default`.

**Criticality:** 30%

### ChiR3 — Retained op-counting text contradicts the bullet this PR just added

**Problem:** the new bullet says the count "is bounded above by script size", which is correct, but three surrounding statements survived the edit and say otherwise. Against `bitcoin-sv/src/script/interpreter.cpp`:

- `- Op counting includes all executed operations, not just total opcodes in script` is backwards. The increment at `:466` is `if((opcode > OP_16) && !IsValidMaxOpsPerScript(++nOpCount, params))`, and it is not guarded by `fExec` — `fExec` is computed at `:459` and only consulted further down for execution. Opcodes inside a not-taken `OP_IF` branch still count. The tally is the *static* non-push opcode count of the script, never more than it.
- `- Loops count each iteration toward the limit` describes something that does not exist. `nOpCount` is declared at `:432`, the scan is `while (pc < pend)` with `pc` only advancing through `GetOp`, and there is no opcode that moves it backwards — no byte is ever counted twice.
- The Purpose line, "Maximum number of opcodes that can execute in a script", has the same two problems: pushes never count, and opcodes that do not execute do.

  Separately, the crossover figure for the old default is loose by about 2x. `- **1000000** - ... (about 2 MB of opcode-dense script)` assumes roughly half a counted op per byte, but every counted opcode is a single byte — `OP_16` is `0x60` (`bitcoin-sv/src/script/opcodes.h:36`), so single-byte opcodes above it such as `OP_DROP` (`0x75`) and `OP_DUP` (`0x76`) each cost one byte and one count. A script can carry up to one counted op per byte, so ~1 MB was already enough to trip the old cap.

**Why it matters:** the case for this PR is "ops cannot exceed bytes", and a doc that tells operators loops multiply the count asserts the opposite in the same bullet list. Same operator-facing surface as ChiR2, so this is what someone reads while deciding whether to lower the knob — and ChiR1's advice to move `maxopsperscriptpolicy` and `maxscriptsizepolicy` together only makes sense if the bounded-by-size relationship is stated consistently.

**Fix:** delete the two stale bullets, reword Purpose to "Maximum number of non-push opcodes in a script (counted whether or not the branch executes)", and change "about 2 MB" to "as little as ~1 MB".

**Criticality:** 20%

## Recap

|---|---|---|---|
| ChiR1 | Default sits on panic ceiling | | 40% |
| ChiR2 | Zero equals the new default | | 30% |
| ChiR3 | Stale op-counting description | | 20% |

galt-tr (approved):

galt-tr:
@Calgooon please fix the failing lint check

PR #1659, opened 2026-08-26
fix(legacy): advertise feefilter in satoshis/kB, not a truncated BSV float

Calgooon:
## What

`resetFeeFilterToDefault` built the wire `feefilter` message from `int64(Policy.MinMiningTxFee)`. `MinMiningTxFee` is configured in BSV/kB (default 0.000005), so the message always carried `MinFee: 0` and every legacy peer was told this node wants all transactions, regardless of its configured fee policy. The bug was invisible internally because the guard and the stored `currentFeeFilter` marker already used the `SatoshiPerBitcoin` conversion; only the value put on the wire was truncated.

## Fix

* Convert once, with `math.Round`, in one helper (`policyFeeFilterSatoshisPerKB`) used by the reset path and the new-peer path. Rounding matters because many decimal rates sit just below their integer number of satoshis in float64: 0.00000003 BSV/kB is 2.9999999999999996 sat/kB and truncates to 2 (79,946 of the million rates between 0.00000001 and 0.01000000 lose a satoshi to truncation). A negative rate advertises 0; a non-finite rate (settings parsing accepts NaN and Inf) advertises 0 with a warning rather than a platform-dependent cast; a finite rate beyond int64 clamps to MaxInt64.
* Tell every peer. The reset only runs on the transition to RUNNING, so a peer connecting to a node that had been running for a while was never told anything. `handleNewPeerMsg` now sends the policy filter whenever the node is not catching up, and registers the peer before deciding its filter so a concurrent reset finds it.
* Claim the marker before broadcasting. The reset's load, broadcast and store were not atomic against the catch-up raise stored for a late peer on another goroutine, which could bury that raise and leave the peer at 1 BSV/kB for good. The marker is now taken with a compare-and-swap first; a raise stored after the claim stays visible and the next reset lowers that peer.

## What the filter means

A deliberate choice, recorded in the helper's comment: the advertised value is `minminingtxfee`, the rate below which this node's validator rejects a transaction outright, so under BIP133 the filter states what the node accepts. The one cost is a zero-fee consolidation, which the node accepts but which a filtering peer will not relay to it; such transactions still arrive by direct submission and from peers that do not filter. svnode advertises its mempool's rolling minimum instead because it admits sub-floor transactions and only declines to mine them; Teranode has no such mempool and no lower floor to advertise.

## Testing

* `TestResetFeeFilterToDefault_AdvertisesSatoshisPerKB` asserts the remote peer receives `MinFee: 3` for a 0.00000003 BSV/kB configuration. With `math.Round` replaced by truncation the test fails (an earlier revision used 0.00000250, which truncates correctly and could not tell the two apart; the review caught that by mutation).
* `TestPolicyFeeFilterSatoshisPerKB` tables the conversion: rounding, negative, NaN, both infinities, and two rates beyond int64.
* `TestHandleNewPeerMsg_SendsPolicyFilterWhenRunning` asserts a peer connecting while RUNNING receives the policy filter and the marker is untouched.
* `TestResetFeeFilterToDefault_LowersARaiseStoredAfterTheReset` pins the compare-and-swap ordering: a peer raised after a reset is lowered by the next one.
* The existing catch-up, nil-FSM-state and disconnected-peer tests still pass.
* `go test -race ./services/legacy/netsync/`, `go build ./...`, `go vet`, `golangci-lint run --new-from-rev upstream/main`, `scripts/check_error_wrap_verbs.sh`, `scripts/check_filenames.sh`: green.

Also corrects the same wrong float example in `ScriptVerifierGoBDK.go`, where it came from (three comment lines).

### DIFF
--- a/services/legacy/netsync/manager.go
+++ b/services/legacy/netsync/manager.go
@@ -12,6 +12,7 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"math"
 	"math/rand/v2"
 	"net"
 	"net/url"
@@ -854,8 +855,23 @@ func (sm *SyncManager) startSync() {
 }
 
 func (sm *SyncManager) resetFeeFilterToDefault() {
-	if sm.currentFeeFilter.Load() != uint64(bsvutil.SatoshiPerBitcoin*sm.settings.Policy.MinMiningTxFee) {
-		feeFilter := wire.NewMsgFeeFilter(int64(sm.settings.Policy.MinMiningTxFee)) // nolint:gosec
+	// MinMiningTxFee is configured in BSV/kB while the wire feefilter message
+	// carries satoshis/kB. Convert with math.Round, not truncation: IEEE-754
+	// cannot represent some decimal rates exactly (0.00000250 stores as
+	// 0.0000024999...), and truncating the raw float dropped the value to 0
+	// sat/kB for every real configuration, so peers were told this node wants
+	// all transactions regardless of its fee policy. This mirrors the
+	// conversion pushed into BDK by newScriptVerifierGoBDK.
+	satoshisPerKB := int64(math.Round(sm.settings.Policy.MinMiningTxFee * bsvutil.SatoshiPerBitcoin))
+	if satoshisPerKB < 0 {
+		// A negative rate is rejected at validator startup, but the legacy
+		// service can run in a process without a validator; never advertise a
+		// wrapped-around filter.
+		satoshisPerKB = 0
+	}
+
+	if sm.currentFeeFilter.Load() != uint64(satoshisPerKB) {
+		feeFilter := wire.NewMsgFeeFilter(satoshisPerKB)
 
 		for p := range sm.peerStates.Range() {
 			if p == nil {
@@ -869,7 +885,7 @@ func (sm *SyncManager) resetFeeFilterToDefault() {
 			p.QueueMessage(feeFilter, nil)
 		}
 
-		sm.currentFeeFilter.Store(uint64(bsvutil.SatoshiPerBitcoin * sm.settings.Policy.MinMiningTxFee))
+		sm.currentFeeFilter.Store(uint64(satoshisPerKB))
 	}
 }
 
--- a/services/legacy/netsync/manager.go
+++ b/services/legacy/netsync/manager.go
@@ -854,38 +854,98 @@ func (sm *SyncManager) startSync() {
 	})
 }
 
-func (sm *SyncManager) resetFeeFilterToDefault() {
-	// MinMiningTxFee is configured in BSV/kB while the wire feefilter message
-	// carries satoshis/kB. Convert with math.Round, not truncation: IEEE-754
-	// cannot represent some decimal rates exactly (0.00000250 stores as
-	// 0.0000024999...), and truncating the raw float dropped the value to 0
-	// sat/kB for every real configuration, so peers were told this node wants
-	// all transactions regardless of its fee policy. This mirrors the
-	// conversion pushed into BDK by newScriptVerifierGoBDK.
-	satoshisPerKB := int64(math.Round(sm.settings.Policy.MinMiningTxFee * bsvutil.SatoshiPerBitcoin))
-	if satoshisPerKB < 0 {
-		// A negative rate is rejected at validator startup, but the legacy
-		// service can run in a process without a validator; never advertise a
-		// wrapped-around filter.
-		satoshisPerKB = 0
+// policyFeeFilterSatoshisPerKB is the feefilter this node advertises to legacy
+// peers whenever it is not catching up: minminingtxfee, converted from the
+// BSV/kB it is configured in to the satoshis/kB the wire message carries.
+//
+// The conversion rounds rather than truncates. Many decimal rates sit just
+// below their integer number of satoshis in IEEE-754 (0.00000003 BSV/kB is
+// 2.9999999999999996 satoshis/kB, which truncates to 2), and before this
+// conversion existed the raw BSV/kB float was truncated, which is 0 for every
+// real configuration. newScriptVerifierGoBDK pushes the same rounded value into
+// BDK.
+//
+// A non-finite rate advertises 0, the accept-everything filter: settings
+// parsing accepts "NaN" and "Inf" without error, and casting such a value is
+// platform-dependent (int64 of +Inf is MaxInt64 on arm64, which would silently
+// stop every peer relaying anything, and MinInt64 on amd64). A finite rate
+// beyond the int64 range is clamped to MaxInt64, the only representation of
+// "reject everything" the message has. A negative rate, rejected at validator
+// startup but reachable in a process that runs legacy without a validator,
+// advertises 0 rather than a wrapped value.
+//
+// What the filter means is a deliberate choice (PR 1659 review). BIP133 says a
+// peer will not relay transactions paying less than the filter, and
+// minminingtxfee is the rate below which this node's validator rejects a
+// transaction outright (BDK's fee floor), so the filter states what the node
+// accepts. The one exception is a free consolidation (isFreeConsolidationTxn in
+// the validator), which this node accepts with no fee and which a filtering
+// peer will therefore not relay to it; such transactions still arrive by
+// direct submission and from peers that do not filter. svnode derives its
+// filter from the mempool's rolling minimum instead, but svnode admits
+// sub-floor transactions to its mempool and only declines to mine them, so its
+// relay floor and its mining floor are different numbers; Teranode has no
+// mempool of that kind and no lower floor to advertise.
+func (sm *SyncManager) policyFeeFilterSatoshisPerKB() int64 {
+	rate := sm.settings.Policy.MinMiningTxFee
+	if math.IsNaN(rate) || math.IsInf(rate, 0) {
+		sm.logger.Warnf("[policyFeeFilter] minminingtxfee %v is not a finite number, advertising a feefilter of 0", rate)
+
+		return 0
 	}
 
-	if sm.currentFeeFilter.Load() != uint64(satoshisPerKB) {
-		feeFilter := wire.NewMsgFeeFilter(satoshisPerKB)
+	satoshisPerKB := math.Round(rate * bsvutil.SatoshiPerBitcoin)
+	if satoshisPerKB < 0 {
+		return 0
+	}
 
-		for p := range sm.peerStates.Range() {
-			if p == nil {
-				continue
-			}
+	// float64(math.MaxInt64) is 2^63 exactly, one above the largest int64, so
+	// anything at or above it does not fit.
+	if satoshisPerKB >= float64(math.MaxInt64) {
+		return math.MaxInt64
+	}
 
-			if !p.Connected() {
-				continue
-			}
+	return int64(satoshisPerKB)
+}
 
-			p.QueueMessage(feeFilter, nil)
+// resetFeeFilterToDefault tells every connected legacy peer the policy fee
+// filter, once per change of the advertised value. It is called on reaching
+// current from more than one goroutine (the block-queue consumer in
+// handleBlockMsg and the message handler in blockHandler), and it races the
+// catch-up raise handleNewPeerMsg stores for a peer that connects while the
+// node is still catching up. The marker is therefore claimed with a
+// compare-and-swap BEFORE the broadcast: two resets broadcast once, and a raise
+// stored after the claim is not overwritten unseen but leaves the marker
+// raised, so the next reset lowers that peer too. (Storing the marker after the
+// broadcast, as this used to, could bury such a raise under the policy value
+// and leave the peer filtered at 1 BSV/kB for good.)
+func (sm *SyncManager) resetFeeFilterToDefault() {
+	satoshisPerKB := sm.policyFeeFilterSatoshisPerKB()
+	want := uint64(satoshisPerKB) //nolint:gosec // never negative, see policyFeeFilterSatoshisPerKB
+
+	for {
+		old := sm.currentFeeFilter.Load()
+		if old == want {
+			return
 		}
 
-		sm.currentFeeFilter.Store(uint64(satoshisPerKB))
+		if sm.currentFeeFilter.CompareAndSwap(old, want) {
+			break
+		}
+	}
+
+	feeFilter := wire.NewMsgFeeFilter(satoshisPerKB)
+
+	for p := range sm.peerStates.Range() {
+		if p == nil {
+			continue
+		}
+
+		if !p.Connected() {
+			continue
+		}
+
+		p.QueueMessage(feeFilter, nil)
 	}
 }
 
@@ -986,26 +1046,38 @@ func (sm *SyncManager) handleNewPeerMsg(peer *peerpkg.Peer) {
 	// Initialize the peer state
 	isSyncCandidate := sm.isSyncCandidate(peer)
 
-	// While catching up, ask every newly-connected peer to hold back
-	// transaction announcements to reduce load during sync. The raise is queued
-	// per-peer; the global currentFeeFilter is only the marker the reset path
-	// (resetFeeFilterToDefault) checks, so it must NOT gate the per-peer queue —
-	// otherwise only the first peer to connect during catch-up would be told.
-	// The filter is restored to the policy default once we reach RUNNING.
-	if state, ferr := sm.blockchainClient.GetFSMCurrentState(sm.ctx); ferr != nil {
-		sm.logger.Errorf("[handleNewPeerMsg] failed to get current FSM state: %v", ferr)
-	} else if state != nil && *state == teranodeblockchain.FSMStateCATCHINGBLOCKS {
-		feeFilter := wire.NewMsgFeeFilter(bsvutil.SatoshiPerBitcoin)
-		peer.QueueMessage(feeFilter, nil)
-		sm.currentFeeFilter.Store(bsvutil.SatoshiPerBitcoin)
-	}
-
+	// Register the peer BEFORE deciding on its feefilter, so a
+	// resetFeeFilterToDefault running concurrently on another goroutine finds
+	// it in peerStates and lowers whatever is queued for it below.
 	sm.peerStates.Set(peer, &peerSyncState{
 		syncCandidate:   isSyncCandidate,
 		requestQueue:    txmap.NewSyncedSlice[wire.InvVect](maxRequestedBlocks),
 		requestedTxns:   expiringmap.New[chainhash.Hash, struct{}](10 * time.Second), // allow the node 10 seconds to respond to the tx request
 		requestedBlocks: expiringmap.New[chainhash.Hash, struct{}](60 * time.Minute), // allow the node 1 hour to respond to the requested blocks, needed for legacy sync/checkpoints
 	})
+
+	// While catching up, ask every newly-connected peer to hold back
+	// transaction announcements to reduce load during sync. The raise is queued
+	// per-peer; the global currentFeeFilter is only the marker the reset path
+	// (resetFeeFilterToDefault) checks, so it must NOT gate the per-peer queue,
+	// otherwise only the first peer to connect during catch-up would be told.
+	// The filter is restored to the policy default once we reach RUNNING.
+	//
+	// Otherwise tell the peer the policy fee filter now. The reset path only
+	// runs on the transition to RUNNING, so before this a peer connecting to a
+	// node that had been running for a while was never told anything and kept
+	// relaying every transaction (PR 1659 review). The marker is left alone:
+	// it records what the reset path last broadcast, and this is a per-peer
+	// message.
+	if state, ferr := sm.blockchainClient.GetFSMCurrentState(sm.ctx); ferr != nil {
+		sm.logger.Errorf("[handleNewPeerMsg] failed to get current FSM state: %v", ferr)
+	} else if state != nil && *state == teranodeblockchain.FSMStateCATCHINGBLOCKS {
+		feeFilter := wire.NewMsgFeeFilter(bsvutil.SatoshiPerBitcoin)
+		peer.QueueMessage(feeFilter, nil)
+		sm.currentFeeFilter.Store(bsvutil.SatoshiPerBitcoin)
+	} else {
+		peer.QueueMessage(wire.NewMsgFeeFilter(sm.policyFeeFilterSatoshisPerKB()), nil)
+	}
 
 	// Start syncing by choosing the best candidate if needed.
 	if isSyncCandidate && sm.loadSyncPeer() == nil {
--- a/services/legacy/netsync/manager_test.go
+++ b/services/legacy/netsync/manager_test.go
@@ -1826,3 +1826,60 @@ func TestHandleNewPeerMsg_SkipsDisconnectedPeer(t *testing.T) {
 
 	require.False(t, sm.peerStates.Exists(disconnectedPeer), "disconnected peer must not be registered in peerStates")
 }
+
+// TestResetFeeFilterToDefault_AdvertisesSatoshisPerKB verifies that the
+// feefilter message sent to legacy peers carries the configured policy minimum
+// converted to satoshis/kB. MinMiningTxFee is configured in BSV/kB; before the
+// fix the raw float was truncated to int64, so every real configuration
+// advertised a filter of 0 sat/kB and peers relayed all transactions regardless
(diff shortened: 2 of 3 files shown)


services/legacy/netsync/manager_test.go
```diff
@@ -1826,3 +1826,60 @@ func TestHandleNewPeerMsg_SkipsDisconnectedPeer(t *testing.T) {
...
 }
+
+// TestResetFeeFilterToDefault_AdvertisesSatoshisPerKB verifies that the
+// feefilter message sent to legacy peers carries the configured policy minimum
+// converted to satoshis/kB. MinMiningTxFee is configured in BSV/kB; before the
+// fix the raw float was truncated to int64, so every real configuration
+// advertised a filter of 0 sat/kB and peers relayed all transactions regardless
+// of this node's fee policy. The rate 0.00000250 is chosen deliberately: it is
+// stored as 0.0000024999... in IEEE-754, so this also pins the math.Round
+// (rather than truncating) conversion, matching newScriptVerifierGoBDK.
```
freemans13: This explanation is backwards for the rate the test uses, and the test therefore does not pin what it says it pins.

In float64, `0.00000250 * 1e8` is `250.00000000000002842`. The product lands just above 250, not just below, so `int64(...)` of it is already 250 and truncation gives the same answer as `math.Round`.

Swapping `math.Round(x)` for plain `int64(x)` in `resetFeeFilterToDefault` leaves this test green. I ran it.

`math.Round` does matter, so this is only a choice-of-rate problem. Of the 1,000,000 rates from 0.00000001 to 0.01000000 at 8 decimal places, 79,946 land below their integer and lose a satoshi to truncation. The smallest is 0.00000003, whose product is `2.9999999999999995559`: truncation 2, rounding 3. Using that rate (or 0.00000024, product `23.999999999999996447`) would make the test fail without `math.Round`, which is what the comment promises.

services/legacy/netsync/manager.go
```diff
@@ -854,8 +855,23 @@ func (sm *SyncManager) startSync() {
 }
 
 func (sm *SyncManager) resetFeeFilterToDefault() {
-	if sm.currentFeeFilter.Load() != uint64(bsvutil.SatoshiPerBitcoin*sm.settings.Policy.MinMiningTxFee) {
-		feeFilter := wire.NewMsgFeeFilter(int64(sm.settings.Policy.MinMiningTxFee)) // nolint:gosec
+	// MinMiningTxFee is configured in BSV/kB while the wire feefilter message
+	// carries satoshis/kB. Convert with math.Round, not truncation: IEEE-754
+	// cannot represent some decimal rates exactly (0.00000250 stores as
+	// 0.0000024999...), and truncating the raw float dropped the value to 0
```
freemans13: Same claim as in the test, and same problem: `0.00000250 * 1e8` is `250.00000000000002842`, above 250 rather than below, so this rate is not an example of truncation dropping a satoshi.

Rounding is still the right call. 79,946 of the 1,000,000 rates between 0.00000001 and 0.01000000 do lose a satoshi to truncation, the smallest being 0.00000003 (product `2.9999999999999995559`, truncates to 2). Worth citing one of those instead so the comment matches the arithmetic.

For what it's worth the same wrong example sits in `ScriptVerifierGoBDK.go:208-210`, which is presumably where it came from. Not this PR's job to fix, but worth not propagating.

services/legacy/netsync/manager.go
```diff
@@ -854,8 +855,23 @@ func (sm *SyncManager) startSync() {
...
+	// conversion pushed into BDK by newScriptVerifierGoBDK.
+	satoshisPerKB := int64(math.Round(sm.settings.Policy.MinMiningTxFee * bsvutil.SatoshiPerBitcoin))
+	if satoshisPerKB < 0 {
+		// A negative rate is rejected at validator startup, but the legacy
+		// service can run in a process without a validator; never advertise a
+		// wrapped-around filter.
+		satoshisPerKB = 0
+	}
+
+	if sm.currentFeeFilter.Load() != uint64(satoshisPerKB) {
```
freemans13: This guard means the fix only reaches peers that were already connected when the node transitioned to RUNNING.

All three callers of `resetFeeFilterToDefault` fire on that transition (`manager.go:785`, `manager.go:1893`, `manager.go:2563`), and once `currentFeeFilter` holds the policy value this branch never runs again. `handleNewPeerMsg` queues a feefilter only in the CATCHINGBLOCKS branch just below, so a peer connecting during steady-state RUNNING is sent nothing at all.

Before this PR that was harmless, because the advertised value was 0 and BIP133 treats an absent feefilter the same way. Now the two diverge, and on a long-lived node most peers will be later arrivals holding no filter while a shrinking set holds 500 sat/kB.

Sending the stored `currentFeeFilter` to every new peer in `handleNewPeerMsg`, whatever the FSM state, would close it. If you would rather keep the change minimal, I would at least drop the scope into the commit message so the next person does not read "legacy peers are told our fee policy" as covering all of them.

freemans13 (changes requested):
The core fix is right and worth having. `resetFeeFilterToDefault` really was building the wire message from `int64(MinMiningTxFee)`, which is 0 for every rate below 1 BSV/kB, so legacy peers were told this node accepts anything. Converting once and reusing the value for the guard, the message and the marker is the correct shape.

Two things I would want changed before merge, and two notes that are not blocking.

## The test does not pin the rounding, and its comment says it does

Both `manager.go` and the test say the rate 0.00000250 is stored as 0.0000024999... in IEEE-754, so `math.Round` is needed where truncation would yield 249. That is backwards for this particular rate. In float64, `0.00000250 * 1e8` is 250.00000000000002842, which sits just *above* 250, so truncation also gives 250.

I checked by swapping `math.Round(x)` for plain `int64(x)` in `resetFeeFilterToDefault` and running the test. It still passes. The test cannot tell the fix from its absence on the one behaviour it was written to pin.

`math.Round` is genuinely load-bearing, so this is a test-value problem rather than a code problem. Sweeping every rate from 0.00000001 to 0.01000000 at 8 decimal places, 79,946 of the 1,000,000 land below their integer and truncate one satoshi low. The smallest is 0.00000003, where the product is 2.9999999999999995559: truncation gives 2, rounding gives 3. Any of those as the test rate would make the test do what its comment claims.

The same wrong explanation is in `ScriptVerifierGoBDK.go:208-210`, which is where this one was copied from. That is not yours and not this PR's job, but it is worth a follow-up so the claim stops spreading.

## A peer that connects while the node is already RUNNING never receives the filter

All three callers of `resetFeeFilterToDefault` fire on the transition into RUNNING (`manager.go:785`, `manager.go:1893`, `manager.go:2563`), and the `currentFeeFilter` guard stops it repeating. `handleNewPeerMsg` only queues a feefilter inside the CATCHINGBLOCKS branch (`manager.go:995-1001`). So the policy reaches exactly the peers that were connected at the moment of the transition, and nobody else.

That structure predates this PR, but it had no effect before, because the advertised value was 0 and BIP133 treats "no feefilter sent" as the same thing. Now the two differ: peers present at the transition hold 500 sat/kB, later arrivals hold nothing, and on a node that has been up for a while that is most of them. I would either send the current filter to every new peer in `handleNewPeerMsg` regardless of state, or say in the commit message that the fix only covers peers connected at the transition.

## Not blocking

Copilot's point about non-finite input is real, and I confirmed one concrete outcome. Replied in that thread with the numbers rather than repeating it here.

`services/p2p/message_types.go:42` truncates the same conversion (`uint64(feeInSatoshis)`) where legacy now rounds. At 0.00000003 BSV/kB the HTTP status endpoint would report 2 sat/kB while legacy peers are told 3. Cosmetic, out of scope here, but the argument for rounding applies there too.

## What I ran

Local, on darwin/arm64, at 76eb1c2:

- `go test -race -run TestResetFeeFilterToDefault_AdvertisesSatoshisPerKB ./services/legacy/netsync/` passes.
- `go test -count=1 ./services/legacy/netsync/` passes, 8.7s.
- `golangci-lint run --new-from-rev=upstream/main services/legacy/netsync/...` reports 0 issues, so dropping the old `nolint:gosec` is fine.
- The mutation described above, reverted straight after. Your working tree is untouched.

I did not run the legacy service against a real peer, so the assertion that peers now honour the 500 sat/kB filter is from reading BIP133 and the wire message, not from observed relay behaviour.

services/legacy/netsync/manager.go
```diff
@@ -854,8 +855,23 @@ func (sm *SyncManager) startSync() {
...
+	// sat/kB for every real configuration, so peers were told this node wants
+	// all transactions regardless of its fee policy. This mirrors the
+	// conversion pushed into BDK by newScriptVerifierGoBDK.
+	satoshisPerKB := int64(math.Round(sm.settings.Policy.MinMiningTxFee * bsvutil.SatoshiPerBitcoin))
+	if satoshisPerKB < 0 {
+		// A negative rate is rejected at validator startup, but the legacy
+		// service can run in a process without a validator; never advertise a
+		// wrapped-around filter.
+		satoshisPerKB = 0
+	}
```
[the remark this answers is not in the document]
freemans13: Agreeing with this, with a concrete outcome rather than a general worry.

`strconv.ParseFloat("NaN", 64)` returns NaN with no error at all, and teranode's `getFloat64` (`settings/helpers.go:101-105`) throws away gocore's ok flag, so a junk value in settings.conf reaches `MinMiningTxFee` instead of falling back to the default. Reachable, not theoretical.

What happens next depends on the machine, which is the part worth fixing. On darwin/arm64 I measured `int64(math.Round(math.Inf(1)*1e8))` as `9223372036854775807`, so `+Inf` sails past the `< 0` clamp and advertises a feefilter of MaxInt64 sat/kB. Every legacy peer then withholds every transaction, silently. On amd64 the same expression is `INT64_MIN`, the clamp catches it, and we advertise 0. Opposite behaviours from one config value.

NaN happens to land on 0 on arm64 and on `INT64_MIN` on amd64, so that one is survivable either way. `+Inf` is not.

The guard shape already exists in the repo at `services/p2p/message_types.go:37-45`. An `math.IsNaN || math.IsInf` check before the multiply, plus an upper bound on the product, would match it.

Low likelihood, since it needs a malformed config. Worth guarding anyway because the failure is total and silent.

icellan (changes requested):
Conversion is correct and the test probe is well chosen (0.00000250 is 249.9999… in float64, so it actually distinguishes round from truncate). Two things before this lands:

- `manager.go:873` — the `currentFeeFilter` marker gate means this fix mostly doesn't reach peers. `resetFeeFilterToDefault` is the only non-catch-up `NewMsgFeeFilter` call site. If the node reaches RUNNING with no legacy peers connected, the marker stores 500 and zero messages go out; peers connecting later only get a feefilter from `handleNewPeerMsg` when the FSM is in `CATCHINGBLOCKS`, and every later reset short-circuits on the marker. Those peers are never told anything and keep relaying everything — the symptom this PR is fixing. Sending the policy filter from `handleNewPeerMsg` in the non-catch-up branch closes it.
- `manager.go:865` — `MinMiningTxFee` isn't the acceptance floor. `settings/policy_settings.go:22` and BDK's `IsFreeConsolidation` exempt consolidation transactions from the fee floor, so advertising 500 sat/kB makes BIP133-compliant peers withhold zero-fee consolidations this node would accept and mine. bitcoin-sv derives feefilter from the relay minimum, not the mining minimum. Teranode has no separate relay setting, so it's a call to make — just make it deliberately rather than as a side effect of a units fix.

Pre-existing, now live: the load-check/broadcast/store in `resetFeeFilterToDefault` isn't atomic against the catch-up raise at `:998` (different goroutines — block-queue consumer vs `blockHandler`). A peer can register mid-reset, get the 1e8 raise, and be stranded at 1e8 sat/kB forever since the marker then reads 500. Harmless while the reset stored 0; worth a `CompareAndSwap` now.

Calgooon:
Thank you both, and apologies for the wait on this one. Every point below was checked before it was changed.

@freemans13, you were right on both counts, and the mutation was the right way to show it:

1. The test could not tell rounding from truncation, because `0.00000250 * 1e8` is `250.00000000000003` in float64, just above 250. It now uses `0.00000003` (product `2.9999999999999996`, truncating to 2), and with `math.Round` swapped for truncation both it and a new table test fail. The comments in `manager.go` and the test that repeated the wrong example are corrected, and so is the same wrong example in `ScriptVerifierGoBDK.go`, which is where it came from; that is a small separate commit here rather than a new PR, since it is the same three lines. (The same example had spread into PR 1660 as well; that will be corrected there.)
2. A peer connecting while the node is already running is now told the policy filter from `handleNewPeerMsg`, whatever the FSM state other than catching up, and is registered before its filter is decided so a concurrent reset finds it. Test added for the running case.
3. The non-finite guard is in: NaN and Inf advertise 0 with a warning, and a finite rate beyond int64 is clamped to MaxInt64. Your arm64 versus amd64 numbers are in the comment. The truncation in `services/p2p/message_types.go` is left for another change, as you suggested.

@icellan, thank you as well:

1. The new-peer gap, as above.
2. On what the filter should mean, I made the call and wrote it down in the helper's comment. This node's validator rejects a transaction below `minminingtxfee` outright (BDK's floor), so under BIP133 that rate is what the node accepts and the filter states it. The cost is exactly the one you named: a zero-fee consolidation, which this node accepts, is not relayed to it by a filtering peer, and svnode's relay check uses the raw fee rate (`TxMempoolInfo` builds `feeRate` from `entry.GetFee()`), so that includes svnode peers. Such transactions still arrive by direct submission and from peers that do not filter, and I would rather take that than tell peers to relay everything, which is what the node validates and rejects. On the svnode comparison, one small correction: `-minrelaytxfee` was removed and `MinFeePerKB` is set to 0 in `init.cpp`, so svnode's filter is `max(mempool rolling minimum, 0)` and it advertises nothing under normal load. That fits svnode, which admits sub-floor transactions to its mempool and only declines to mine them; Teranode has no such mempool, so there is no lower floor to advertise. If a separate relay-floor setting is wanted, that is a small follow-up and I am glad to do it.
3. The marker is now claimed with a compare-and-swap before the broadcast. A raise stored by a late peer after the claim leaves the marker raised, and the next reset lowers that peer; a test pins that sequence. Storing after the broadcast, as before, buried the raise under the policy value.

One thing to flag, kindly: the review note says `0.00000250` is `249.9999` in float64 and distinguishes round from truncate. It is `250.00000000000003`, so it does not, which is why the test rate changed. Rounding is still the right call: 79,946 of the million rates between 0.00000001 and 0.01000000 truncate a satoshi low, `0.00000003` among them.

The branch is rebased on main. Happy to change any of it.

icellan:
Re-reviewed at 866179efc.

Both structural findings are fixed: the non-catch-up branch in `handleNewPeerMsg` now sends unconditionally, and CAS-before-broadcast closes the lost-update race — registration precedes the per-peer queue, so either the reset's `Range` picks the peer up or the raise lands after the claim and the marker stays raised for the next reset. The conversion helper also picked up NaN/Inf/negative handling and a correct `>= float64(math.MaxInt64)` clamp.

One new blocker, introduced by the fix to the first finding:

- `manager.go:1079` — the unguarded `feefilter` disconnects every peer below protocol 70013. `MsgFeeFilter.BsvEncode` (go-wire v1.2.11, `msg_fee_filter.go`) returns an error when `pver < FeeFilterVersion` (70013), but `peer.MinAcceptableProtocolVersion = wire.MultipleAddressVersion` = 209 (`peer.go:44`), so the handshake accepts 209..70012. In `outHandler` a `writeMessage` error is fatal — `DisconnectWithLogFunc("write error: ...")` at `peer.go:2376`. An old SV node or light client connecting to a RUNNING Teranode is dropped on connect, every time. The hazard existed before on the single reset broadcast; making the send unconditional puts it on every connection. Guard all three send sites with `p.ProtocolVersion() >= wire.FeeFilterVersion` — line 1079, the catch-up raise at 1076, and the reset loop at 948.

Minor:

- `manager.go:1073` — on `GetFSMCurrentState` error the branch only logs and sends nothing. A transient blockchain-service blip during a connection burst leaves those peers with no filter for the life of the connection, and the reset path won't help them since it early-returns while the marker already equals the policy value. Fall through to `sm.policyFeeFilterSatoshisPerKB()` on error.

On the consolidation exemption: the code comment argues the trade-off is deliberate, which is fine as far as it goes, but enforcement including the exemption lives in BDK (`services/validator/TxValidator.go:224-226`) and previously nothing was advertised outside the reset transition — this PR is what turns the trade-off on for the general case. That wants a maintainer sign-off rather than a comment. Not blocking from me.

services/legacy/netsync/manager_test.go
```diff
@@ -1826,3 +1827,221 @@ func TestHandleNewPeerMsg_SkipsDisconnectedPeer(t *testing.T) {
...
+	require.True(t, sm.peerStates.Exists(p), "peer must be registered")
+}
+
+// TestResetFeeFilterToDefault_LowersARaiseStoredAfterTheReset pins the
+// compare-and-swap ordering. A peer that connects while the FSM still reads
+// CATCHINGBLOCKS is given the raised filter and the marker is set to it, even
+// if a reset has just run on another goroutine. Because the reset claims the
+// marker before broadcasting, that raise is not buried under the policy value:
+// the marker reads raised, and the next reset lowers the peer. With the old
+// store-after-broadcast ordering the marker would read 500 and the peer would
```
freemans13, line 1998: The last sentence of this comment is not true of this test. I reverted the ordering, putting the check back to a plain `Load` and moving the store after the broadcast loop, and the test still passes.

The reason is that the test is sequential. The reset runs to completion, including its store, before `handleNewPeerMsg` starts. The bug the compare-and-swap prevents needs the raise to be stored between the reset's broadcast and the reset's own store, and this ordering never produces that.

The code is right, and I checked the invariant separately rather than taking it on trust. Registration always precedes the per-peer store, so any peer left holding the raise is also visible to the reset's `Range`, and the marker ends raised in every interleaving I could construct.

Two ways out. Either drop the last sentence and rename the test to what it does check, which is that a raise stored after a completed reset survives and is lowered by the next reset. Or add a test hook between the claim and the broadcast so the interleaving can be driven deterministically.

I could not build a clean deterministic version without such a hook. `peerStates` is keyed on the concrete `*peer.Peer` so there is no seam to fake, and `QueueMessage` is asynchronous, so there is nowhere to block the reset mid-broadcast.

services/legacy/netsync/manager.go
```diff
@@ -970,27 +1046,39 @@ func (sm *SyncManager) handleNewPeerMsg(peer *peerpkg.Peer) {
...
+	// it records what the reset path last broadcast, and this is a per-peer
+	// message.
 	if state, ferr := sm.blockchainClient.GetFSMCurrentState(sm.ctx); ferr != nil {
 		sm.logger.Errorf("[handleNewPeerMsg] failed to get current FSM state: %v", ferr)
 	} else if state != nil && *state == teranodeblockchain.FSMStateCATCHINGBLOCKS {
 		feeFilter := wire.NewMsgFeeFilter(bsvutil.SatoshiPerBitcoin)
 		peer.QueueMessage(feeFilter, nil)
 		sm.currentFeeFilter.Store(bsvutil.SatoshiPerBitcoin)
+	} else {
+		peer.QueueMessage(wire.NewMsgFeeFilter(sm.policyFeeFilterSatoshisPerKB()), nil)
```
freemans13, line 1079: Adding @icellan's protocol version guard here is worth doing. I do not think it blocks, and here is why.

`MsgFeeFilter.BsvEncode` does refuse below 70013, and the write error is fatal. A peer below 70013 never gets this far, though. `AssociateConnection` queues a protoconf to every peer 10ms after the handshake with no version check, at `peer.go:2889`, and `MsgProtoconf.BsvEncode` refuses at the same 70013 floor.

I connected pairs at 70012, 70013 and the default 70016, with no netsync and no fee filter sent anywhere. Only the 70012 pair was disconnected, both ends, within 500ms of the handshake. The other two stayed up.

So the guard removes a second cause of a drop that already happens on main. It is still worth having. It makes the disconnect log name the real culprit, and it stops this path depending on protoconf's version floor staying where it is.

services/legacy/netsync/manager.go
```diff
@@ -970,27 +1046,39 @@ func (sm *SyncManager) handleNewPeerMsg(peer *peerpkg.Peer) {
...
 	// otherwise only the first peer to connect during catch-up would be told.
 	// The filter is restored to the policy default once we reach RUNNING.
+	//
+	// Otherwise tell the peer the policy fee filter now. The reset path only
+	// runs on the transition to RUNNING, so before this a peer connecting to a
+	// node that had been running for a while was never told anything and kept
+	// relaying every transaction (PR 1659 review). The marker is left alone:
+	// it records what the reset path last broadcast, and this is a per-peer
+	// message.
 	if state, ferr := sm.blockchainClient.GetFSMCurrentState(sm.ctx); ferr != nil {
```
freemans13, line 1072: Falling through to the policy filter when this errors, as @icellan suggests, is right. It will not cover the case it is aimed at.

`GetFSMCurrentState` returns the client's cached state and only makes an RPC when the cache is empty, at `Client.go:1825-1837`. When the notification stream breaks, the client stores IDLE into that cache and keeps serving it, at `Client.go:1467`, `Client.go:1473` and `Client.go:1561`. A blockchain service blip therefore hands you IDLE with a nil error, and the error branch above is never taken.

IDLE already lands in the else branch, so those peers do get the policy filter. The practical effect runs the other way. While the node is genuinely catching up with a broken stream, new peers are told the policy filter instead of the 1 BSV/kB raise, so the catch-up load shedding quietly stops.

That is not a regression, since before this pull request such peers were sent nothing at all, which BIP133 treats as no filter. It is worth knowing that the only three states are IDLE, RUNNING and CATCHINGBLOCKS, and that IDLE means "unknown" at least as often as it means "idle".

freemans13 (approved):
Approving. Both changes I asked for are in, and mutation testing confirms the tests can now tell each fix from its absence. One test claims to pin something it cannot, and that is the only thing I would still change. It is not a merge blocker.

The `CHANGES_REQUESTED` on this pull request is @icellan's, not mine, so it stays until they clear it. I do have evidence that the blocker behind it describes an outcome that is already true on main, so I think it can be downgraded.

## The pre-70013 disconnect is not new

The mechanism is real. `MsgFeeFilter.BsvEncode` returns an error below protocol version 70013, and a write error is fatal in `outHandler`, which calls `DisconnectWithLogFunc` at `peer.go:2376`.

The outcome is not new, because such a peer is already disconnected before any fee filter can reach it. After a successful handshake `AssociateConnection` queues a protoconf message to every non-stream peer unconditionally, 10ms later, at `peer.go:2889`. `MsgProtoconf.BsvEncode` refuses at the same floor: `ProtoconfVersion` and `FeeFilterVersion` are both 70013 in go-wire v1.2.11.

I measured it with no netsync involved and no fee filter sent anywhere. Three connected pairs, checked 500ms after the handshake:

| negotiated protocol version | both peers still connected |
|---|---|
| 70012 | no |
| 70013 | yes |
| 70016 (the default) | yes |

An old SV node or light client is therefore dropped on connect today, on main, by the protoconf. Adding the version guard at the three fee filter sites is still worth doing, because it removes a second cause of the same drop and keeps the disconnect log honest about which message failed. It does not change which peers survive.

## What the mutations say

I reverted each part of the fix in turn and re-ran the tests.

| mutation | tests |
|---|---|
| `math.Round` replaced by `math.Trunc` | fail |
| the RUNNING per-peer send removed | fail |
| compare-and-swap replaced by the old store-after-broadcast | pass |

The first two are the findings I raised last round, and both are now genuinely pinned. The rounding case fails on the exact value it was written for, 3 against 2. Thank you for picking a rate that does the job.

## The compare-and-swap test does not pin the ordering

Detail is in the inline comment. The code is correct, so this is a claim and coverage problem rather than a behaviour problem.

## Not blocking

@icellan's point about the `GetFSMCurrentState` error path is right, and falling through to the policy filter on error is the correct call. It is not sufficient on its own, because a broken notification stream gives you IDLE with a nil error rather than an error. Detail inline.

## What I ran

Local, darwin/arm64, at 866179efc. Upstream CI has not run on this commit. Every workflow sits at `action_required` waiting for a maintainer, and `gh pr checks` reports only the two Socket Security checks as passing, so the check list is not evidence of anything.

- `go test -count=1 ./services/legacy/netsync/` passes, 9.5s.
- `go test -count=1 -race` over the four fee filter tests passes, 4.2s.
- `golangci-lint run --new-from-rev=upstream/main` over both changed packages reports 0 issues.
- `go vet ./services/legacy/netsync/` clean, and `services/validator` builds.
- The three mutations above, each reverted straight after. Your branch is untouched and the probe test files I added were deleted.

I did not run the legacy service against a real peer. The protocol version numbers above come from in-process peers connected over a pipe, not from a live SV node.

PR #1660, opened 2026-08-26
feat(validator): optional per-script fee tiers for script size and ops (minminingtxfeebyscriptsize, minminingtxfeebyscriptops)

### LINKED ISSUE #1658: [FEAT] Let miners price scripts beyond the size and ops policy caps (per-script fee tiers)
## Is your feature request related to a problem? Please describe

Teranode's fee policy is a single scalar: `minminingtxfee` (BSV/kB), pushed into BDK once at startup (`services/validator/ScriptVerifierGoBDK.go`, `SetMinMiningTxFee`). Every transaction pays the same rate per byte regardless of the validation load its scripts put on a node.

Real workloads on mainnet today carry large, hash-dense executed scripts in the megabyte range. They are consensus-valid, and the ones under the current policy caps have been mined; the rest never relay, because two policy settings gate them with hard ceilings rather than prices:

* `maxscriptsizepolicy` (upstream default now 100,000,000, but mainnet nodes still commonly advertise 500,000)
* `maxopsperscriptpolicy` (1,000,000; issue 1654 proposes UINT32_MAX to match SV Node)

Raising those caps is the right parity move, but it hands miners nothing in return: a script that was previously rejected now validates at the same flat fee rate as a simple payment. Miners have asked for the ability to configure different accepted fee rates for such transactions, so the scripts the caps used to forbid become payable instead of free.

## Describe the solution you'd like

Two optional policy settings, each pricing exactly the metric its cap gates, applied per executed script and marginally, like tax brackets:

* `minminingtxfeebyscriptsize = 500000:10` pairs with `maxscriptsizepolicy`: script bytes beyond 500,000 must pay at least 10 sat/kB extra
* `minminingtxfeebyscriptops = 1000000:10` pairs with `maxopsperscriptpolicy`: counted ops beyond 1,000,000 (the same rule EvalScript counts: every opcode above OP_16, pushes free) must pay at least 10 satoshis per 1000 ops extra

Executed scripts are each input's unlocking script, the locking script it spends, and a legacy P2SH redeem script; a transaction's own output scripts are priced when they are later spent. Ops are counted statically, which equals the executed count since Bitcoin script has no loops.

Design constraints the implementation honours:

1. BDK exposes only the scalar fee rate, so the schedules are enforced in Go (`services/validator/script_fee_tiers.go`), immediately before BDK validation, in policy mode only (`SkipPolicyChecks` bypasses them like every policy rule).
2. The surcharges are additive on top of the `minminingtxfee` floor, which BDK keeps enforcing unchanged, so the Go check only ever tightens policy. Empty settings (the default) disable it entirely; default behaviour is byte-identical to today.
3. BDK exempts free consolidation transactions from its fee floor, so the same exemption is honoured via a Go port of bitcoin-sv `policy.cpp` `IsFreeConsolidationTxn`, pinned by a differential test that runs real signed transactions through GoBDK and asserts BDK and the Go predicate agree in both directions.
4. Rates are integer satoshis per 1000 units, avoiding the IEEE-754 hazards already documented for the float BSV/kB conversion.

Observability: `teranode_validator_script_tiered_fee_rejections` and `teranode_validator_script_tiered_fee_consolidation_exemptions` counters, and `teranode-cli diagnose` warns when a cap is at or below the first tier threshold of its paired setting (scripts would then be rejected before ever reaching a fee tier). PR attached.

## Describe alternatives you've considered

* Extending BDK with per-script fee APIs. Cleaner in one engine, but a cross-repo change to bitcoin-sv/bdk plus a gobdk release, and the Go validator already owns node-context policy checks (`checkInputs`). The enforcement could migrate into BDK later without changing the settings' contract.
* Whole-transaction size tiers (an earlier revision of the attached PR). Simpler, but it prices the wrong thing: a transaction full of small scripts is cheap to validate per byte, while one 3 MB hash-dense script is not. Per-script metrics match both the cost and the caps being raised.
* Fee classes by script type (mAPI feespec style `standard`/`data`). The ecosystem deprecated mAPI, ARC's policy schema is closed to new fields (`additionalProperties: false`), and type taxonomies invite classification disputes. Metric thresholds are objective.

## Additional context

* Strictly opt-in: empty schedules short-circuit before any work, and no existing configuration changes behaviour.
* Admission control only: block assembly is FIFO by creation time, so the settings change what the validator accepts, not transaction ordering.
* Advertisement: the ARC-format `GET /v1/policy` response is unchanged (its schema forbids new fields); `miningFee` keeps reporting the floor. Operators should advertise tiers out of band for now; a follow-up could extend the Teranode-owned p2p `node_status` `fee_policy` message.
* While writing this I found that the legacy `feefilter` message advertises 0 sat/kB regardless of configuration (`int64` truncation of the BSV/kB float in `services/legacy/netsync/manager.go`, while the internal guard and marker correctly use the satoshi conversion). Separate fix PR opened for that: #1659.

Calgooon:
Closes #1658

## What

Adds two optional policy settings that let a miner require a higher minimum fee for demanding executed scripts, each pricing exactly the metric its policy cap gates:

* `minminingtxfeebyscriptsize = 500000:10` pairs with `maxscriptsizepolicy`: script bytes beyond each threshold pay at least the tier rate (sat/kB)
* `minminingtxfeebyscriptops = 1000000:10` pairs with `maxopsperscriptpolicy`: counted ops beyond each threshold pay at least the tier rate (satoshis per 1000 ops)

Both are pipe-separated `<threshold>:<satoshisPerK>` pairs, applied marginally like tax brackets, per executed script (each input's unlocking script, the locking script it spends, and a legacy P2SH redeem script for coins created before Genesis). Empty (the default) disables a setting; default behaviour is byte-identical to main.

The motivation is the cap raises: with `maxscriptsizepolicy` at 100 MB and `maxopsperscriptpolicy` heading to UINT32_MAX (#1654), scripts the caps used to forbid become minable. These settings make them payable instead of free.

## Design

* **Enforcement lives in Go** (`services/validator/script_fee_tiers.go`), immediately before the BDK call in `TxValidator.ValidateTransaction`, gated on `!SkipPolicyChecks`. BDK exposes only the scalar `SetMinMiningTxFee`, so per-script schedules cannot be pushed into it.
* **The Go check only tightens.** The surcharges are additive on top of the `minminingtxfee` floor, which BDK keeps enforcing unchanged, and tier rates are non-negative, sorted, unique and non-decreasing (validated at parse time), so the two layers cannot disagree. Fee arithmetic saturates rather than wraps, so a parse-valid but bad configuration can only raise the required fee.
* **Ops are counted statically and match svnode's executed count**, in the coin's era, with one documented exception. Every opcode above OP_16 counts and pushes are free; a top-level `OP_RETURN` ends the count and a nested one keeps counting; `OP_CHECKMULTISIG` is charged its key count where that is statically certain (IF-depth zero, still executing, a literal push immediately before it) and one op otherwise, which under-counts. Genesis and Chronicle both change the grammar (`OP_VERIF`/`OP_VERNOTIF` are conditionals from Chronicle on), so the walk takes the coin's era from its height exactly as BDK derives its flags. The invariant the design rests on: never count more than svnode for a script BDK accepts, since only that can strand a legitimate coin; under-counting costs the miner revenue and nothing else.
* **Scripts beyond the caps BDK enforces are left unpriced** so BDK's own cap rejection surfaces instead of a misleading insufficient-fee error. For a post-Genesis coin those caps are `maxopsperscriptpolicy` and `maxscriptsizepolicy`; for a pre-Genesis coin they are the fixed 500-op and 10000-byte limits BDK applies instead.
* **Free consolidations stay free of the floor, not of the surcharge.** BDK exempts `IsFreeConsolidationTxn` transactions from its fee floor; `isFreeConsolidationTxn` ports that predicate (bitcoin-sv `policy.cpp`) with standardness following svnode's `Solver` in the coin's era, and the exemption waives only the `minminingtxfee` floor. The surcharge is always due, so a large-script output cannot be created cheaply and then "consolidated" to dodge it. The predicate is held to never being stricter than BDK; where it is more permissive the cost is bounded to the floor, which BDK re-checks one call later.
* **Integer satoshis-per-1000-units rates**, avoiding the documented IEEE-754 hazards of the float BSV/kB unit. The required fee accumulates the floor and all per-script band products in satoshi-thousandths and divides once, so it is monotone.
* **Rejections are wrapped the way BDK's insufficient-fee rejection is**, so gRPC/HTTP mappings and the rejected-transaction Kafka fan-out treat them identically. `SkipScriptValidation` without `SkipPolicyChecks` is rejected so the schedule cannot be bypassed from a request flag.
* **Zero allocations on the per-transaction path** at 2, 20 and 100 inputs, and the ops walk is skipped when a script is shorter than the first ops threshold (unless it contains a multisig opcode, whose key count can exceed the length).
* **Operator guardrail**: `teranode-cli diagnose` warns when a cap is at or below the first tier threshold of its paired setting, since scripts would then be rejected by the cap before ever reaching a fee tier.

## Not in this PR

* Advertisement. The ARC-format `GET /v1/policy` schema forbids new fields (`additionalProperties: false`), so it keeps reporting the floor as `miningFee`. The gap is stated in the setting's own description so an operator sees it before enabling a schedule; extending the Teranode-owned p2p `node_status` `fee_policy` message is the intended follow-up.
* Runtime reload. Policy settings are startup-only across the codebase; the schedules follow the same rule.

## Testing

BDK exposes no op-count getter, so `maxopsperscriptpolicy` is used as an oracle: BDK rejects with `SCRIPT_ERR_OP_COUNT` exactly when its executed count exceeds the cap, so accepting at a cap of N and rejecting at N-1 pins the count to exactly N. Every such test carries a working-oracle control. The same idea with a zero-fee transaction and a non-zero floor measures the consolidation exemption.

* Differential tests against the real GoBDK engine (`script_fee_tiers_differential_test.go`) pin: top-level versus nested `OP_RETURN`; multisig key counts for small-constant and data-push encodings; a nested `OP_RETURN` suppressing later key counts; the coin's era deciding whether a P2SH redeem runs (a pre-Genesis coin spent at a post-Genesis tip still runs it; a post-Genesis coin never does; a coin at an unrecorded height 0 does); `OP_VERIF`/`OP_VERNOTIF` as conditionals from the Chronicle activation height on, with the version comparison; the fixed pre-Genesis caps; `maxscriptsizepolicy` 0 as unlimited; the 0 to 6 rewrite of `minconfconsolidationinput`; and consolidation standardness for P2SH at heights 0, pre-Genesis and post-Genesis and for bare multisig with the Solver's pubkey push sizes, all with signed transactions and the Go predicate agreeing with BDK in both directions.
* Two fuzzers compare the static count with BDK's on random scripts, alternating between a post-Chronicle coin and a post-Genesis pre-Chronicle one. `TestCountScriptOpsFuzzDifferentialBDK` requires exact agreement (20000 scripts, 18643 pinned exactly, zero divergences); `TestCountScriptOpsNeverOverCountsFuzzBDK` covers false branches, `OP_NOTIF`, `OP_RETURN` in a branch that never runs and both truth values of `OP_VERIF`, asserting the walk never counts more than svnode (19060 of 20000 verified). Corpus size is 120 by default for CI; `FEE_TIER_FUZZ_SCRIPTS=20000` runs the deep search.
* Unit tests: the ops rule vector by vector in both grammars, tier arithmetic and saturation, every consolidation disqualifier, the standard templates including the reviewer's two panic reproducers, the fee-tier surcharge always being due, the era plumbing, `SkipPolicyChecks` bypass and the `SkipScriptValidation` guard.
* Benchmarks: the disabled path, 2, 20 and 100 inputs, and `countScriptOps` with `SetBytes` including a 10 MB all-ops case; 0 allocations throughout.
* Full `./services/validator/...`, `./settings/...`, `./cmd/diagnose/` suites with `-race`: green. `go build ./...`, `go vet`, `golangci-lint run --new-from-rev`, `scripts/check_error_wrap_verbs.sh`, `scripts/check_filenames.sh`: green.

## Docs

* `docs/references/settings/policy_settings.md`: new setting rows, format paragraph, validation rows, the cap rule per era.
* `docs/topics/services/validator.md` and `docs/references/services/validator_reference.md`: these still documented the removed `checkFees`; updated to describe BDK floor enforcement plus the new Go-side check.
* `docs/references/prometheusMetrics.md`: the two new counters.
* `settings.conf`: commented example block next to `minminingtxfee`.

## Changes since the first review

All 15 items from @icellan and the 3 from @raspi-user are addressed; the mapping is in the review thread. Beyond those, the walker was put through several rounds of adversarial review with every finding measured against BDK before being accepted, which found and fixed: a three-byte `<push 17> OP_CHECKMULTISIG` counting 18 ops and slipping under the length guard; push lengths assembled in a signed int (a 32-bit hazard on attacker bytes); `lastPush` treating `OP_1` to `OP_16` as non-pushes, which left an `OP_1 <redeem>` scriptSig unpriced; `OP_VERIF`/`OP_VERNOTIF` being conditionals for coins created since Chronicle; the policy caps being applied to pre-Genesis coins BDK caps at 500 ops and 10000 bytes; and two places where the consolidation predicate was stricter than svnode's `Solver` (height 0, and pubkey push sizes).

## History note

The first revision of this branch priced whole-transaction size. It was reworked to the per-script metrics after clarifying the requirement: a transaction full of small scripts is cheap to validate per byte, while a single multi-megabyte hash-dense script is not, and per-script metrics are what the caps being raised actually gate. The commits show the evolution, and the later ones each name the finding they answer; squash-merge is fine.

## Residual risk

Deliberate under-counts, each bounded and each measured: a multisig whose key count arrives on the stack from the other script of the pair is charged one op; a nested `OP_RETURN` in a branch that never runs suspends later key counts where svnode would not; the pre-Chronicle grammar suspends key counts after any `OP_VERIF`; a P2SH redeem at an unrecorded height 0 is not billed (about 520 bytes at most, since pre-Genesis pushes are capped there). Deliberate over-permissiveness in the consolidation predicate (bare multisig up to n of 16 where svnode allows 3, data carriers without the datacarrier flag) costs at most the byte-rate floor, which BDK re-checks. Neither direction can reject a transaction BDK accepts for more than that floor, and no accepted script is counted higher than svnode counts it.


raspi-user:
I’ve reviewed this with focus on policy safety rather than style. The overall design is sensible: the new settings are opt-in, policy-mode only, layered above BDK’s existing `minminingtxfee`, and the tests cover a good set of boundary cases, including P2SH redeem-script charging and a BDK differential check for normal free consolidation behaviour.

I would still ask for changes before merge.

1. **Unchecked arithmetic can silently weaken the configured fee policy.**
   In `services/validator/script_fee_tiers.go`, `tierExcessThousandths` multiplies script units by `SatoshisPerK`, and `checkScriptTieredFees` then adds that to the base floor before dividing by 1000. The parser accepts very large `uint64` thresholds and `int64` rates, so a bad but parse-valid config can wrap `uint64` and produce a lower required fee than intended. The same class of issue exists in the consolidation checks where `factor * len(outputs)` and `factor * sumOutputLockingScriptBytes` are computed unchecked. This is policy/security-sensitive code, so I think it should either use checked/saturating arithmetic or reject configurations whose worst-case calculations cannot be represented safely.

2. **The new fee policy is not discoverable by clients or peers.**
   The PR explicitly leaves `/v1/policy` and P2P `fee_policy` unchanged. I understand the ARC schema issue, but operationally this means a node can reject transactions for a fee rule that wallets and propagation peers cannot discover from the usual policy surfaces. At minimum I would want this called out very clearly in operator docs and tied to a follow-up issue before enabling this in production. Ideally the Teranode-owned P2P policy message should be extended soon, otherwise routing and fee estimation will be guesswork.

3. **The free-consolidation exemption should be fail-closed unless the BDK match is proven for the edge cases.**
   The normal P2PKH differential test is useful, but the implementation states that `isStandardPrevoutScript` is intentionally more permissive than svnode/BDK in some cases. If Go says “free consolidation” where BDK would not, a transaction that pays the base fee but not the script-tier surcharge can be accepted. That may be an intentional tradeoff, but it needs either tighter matching or explicit tests covering the permissive cases so the risk is deliberate and bounded.

I could not run the Go tests locally because this environment does not currently have `go` on `PATH`, so this is a static/context review only. The PR is current and mergeable, and the only reported checks I saw were Socket security checks passing.

icellan (changes requested):
Blocking. Three of these are reproduced, not theorised — details and measurements below. `git status` clean, no benchmarks were committed.

**P0**

1. `services/validator/script_fee_tiers.go:280` — remote validator panic. `isStandardPrevoutScript` calls go-bt's `IsP2PK()`/`IsMultiSigOut()` on an attacker-controlled prevout script; both index `parts[i][0]` with no length check (`bscript/script.go:293`) and `DecodeParts` can emit an empty part. Reproduced in-package: prevout script `01 AA 4C 00` → `index out of range [0] with length 0`. Second variant `4C 00 AE AE` panics in `IsMultiSigOut`. No `recover()` on the gRPC (`util/grpc.go`) or Kafka (`Server.go:368`) paths. Reachable at 1:1 wire cost via extended-format submission, since `Validator.go:1105` trusts a wire-supplied `PreviousTxScript` and the UTXO-hash check is later (`Validator.go:851`). This line is the only use of those two helpers in the repo outside tests. Replace them with prefix/length checks — that fixes P0-2 too.
2. `script_fee_tiers.go:280` — memory amplification. The same chain runs `DecodeParts` twice over the full prevout script, ~289 bytes allocated per script byte. Measured: 10 MB prevout → 2,885 MB allocated, 21 M mallocs, 636 ms, 721 MB peak `HeapInuse`; 32 MB → 8,815 MB / 1.65 s. `Validator.go:1219` fills `PreviousTxScript` from the UTXO store, so a ~200-byte tx spending a pre-created 10 MB-locking-script outpoint costs 0.6 core-seconds — free (rejected, no fee) and repeatable with distinct children of the same outpoint. `checkInputs` (`TxValidator.go:462`) only bounds this when `maxcoinsviewcachesize > 0`, default 0.
3. `script_fee_tiers.go:78` — op count diverges from svnode at `OP_RETURN`. `interpreter.cpp:857` returns `SCRIPT_ERR_OK` post-Genesis on a top-level `OP_RETURN` and never reaches the increment at `:464`; `countScriptOps` keeps walking. `OP_DUP OP_HASH160 <h20> OP_EQUALVERIFY OP_CHECKSIG OP_RETURN <10MB>`: svnode counts 5, this counts ~10,000,005. With your documented `1000000:10` that's a ~90,000 sat surcharge on a ~200-byte spend — rejected here, accepted by BDK one call later and by every other node. The trailing bytes are attacker-chosen, so this is a griefing vector: pay someone into a data-tailed P2PKH script and their coin is unspendable on tier-enabled Teranode nodes. Falsifies the comment at `:44-49` and `settings.conf:911`.

**P1**

4. `script_fee_tiers.go:317` — the exemption is always claimable, so the surcharge is not collectible on any path. The donation branch sets `factor = len(tx.Inputs)`, making its own UTXO-reduction test `N < N*1` — false for every N>=1, so a 1-in/1-out tx that consolidates nothing qualifies; `minConf = 0` drops both confirmation checks. The ratio test is satisfied *more* easily the larger the priced script. Working evasion: create a `0x6a`+10 MB locking script (go-bt `IsData()` has no length bound, and `executedScripts` excludes own outputs so creation pays only the floor), spend it immediately with `0x51` and one `00 6a 04 'dust'` output — ~95,000 sat of surcharge collects 0. This shape is faithful to `policy.cpp:41-58`, so it's not a port bug; the problem is reusing an exemption written for a flat dust-sized floor to waive an unbounded surcharge. The "cost falls on the future spender" promise at `:139-144` is unenforceable.
5. `script_fee_tiers.go:279` — `isStandardPrevoutScript` is era-blind; `policy.cpp:123` judges standardness at `coinHeight`, so post-Genesis `Solver` maps `IsP2SH` to `TX_NONSTANDARD` and `TX_NULL_DATA` needs the `OP_FALSE OP_RETURN` prefix + `IsPushOnly` + `GetDataCarrier`. go-bt's `IsData()` accepts any `OP_RETURN`-prefixed script. 20 x `OP_RETURN <100MB>` outputs spent with 1-byte `OP_TRUE` unlockers into one P2PKH output passes every rule → zero surcharge for 2 GB of executed script, recorded as a legitimate consolidation.
6. `script_fee_tiers.go:79` — `OP_CHECKMULTISIG` priced at one op. `interpreter.cpp:1536` does `nOpCount += nKeysCount` from the stack, invisible to a static walk. With `MaxPubKeysPerMultisigPolicy` default 0, a 1-of-1,000,000 multisig lock has svnode ops = 1,000,001 and `countScriptOps` = 1 → surcharge 0 on the single densest shape the setting exists to price. No static walk can be exact here (same construct in a dead branch adds 1), so the doc comment should say "approximation that under-counts CHECKMULTISIG", not "the exact rule EvalScript counts".
7. `script_fee_tiers.go:327` — `MaxConsolidationInputScriptSize` and `MinConfConsolidationInput` are read raw, but `config.cpp:335`/`:354` rewrite 0 to 150 and 6, and `ScriptVerifierGoBDK.go:227`/`:231` push the raw values in. With `maxconsolidationinputscriptsize = 0`, BDK calls a 20-in P2PKH consolidation free while `:358` rejects it — exactly what the file header promises cannot happen. With `minconfconsolidationinput = 0` the divergence flips. Normalise, or read the effective values back from BDK.
8. `script_fee_tiers.go:164` — P2SH redeem branch has no Genesis/BIP16 gate. `VerifyScript` gates the third `EvalScript` on `!(flags & SCRIPT_UTXO_AFTER_GENESIS)`. A post-Genesis `OP_HASH160 <20> OP_EQUAL` output spent with `OP_PUSHDATA4 <20MB preimage>` bills the same 20 MB twice (once at `:156`) against a script no BSV node runs. The sigop code this branch cites as its model gates correctly — `services/legacy/txscript/script.go:702` requires `ScriptBip16`. For genuine pre-Genesis P2SH the double count *is* correct parity, so the fix is the era gate on the UTXO height (`utxoHeights` already has it), not removing the double count. Confirm against BDK whether P2SH still applies post-Genesis for pre-Genesis coins and say so in a comment either way.
9. `script_fee_tiers.go:222` — unbounded walk ahead of BDK's own caps. Measured 1.48 ms / 15.5 ms / 146.9 ms at 1/10/100 MB of counted opcodes (~700 MB/s), single-threaded, before the BDK call at `TxValidator.go:247`. svnode aborts with `SCRIPT_ERR_OP_COUNT` at the 1,000,000 cap, so BDK spends O(cap) and this spends O(length). Also mislabels the rejection: an over-cap script now reports "insufficient fee", so the submitter retries with more fee forever. Fix: stop counting at `maxopsperscriptpolicy+1` and skip the surcharge. Stopping at the last tier threshold is *not* sound — `tierExcessThousandths` sets `upper = value` for the final tier. Your "BDK has no per-script fee API" claim checks out (`bdkcgo/txvalidator_cgo.h` exposes only `GetSigOpCount` and the policy setters), so Go enforcement is the right call.
10. `script_fee_tiers.go:221` — the ops walk is provably skippable on the common path. Every counted opcode advances `i` by exactly 1, so `ops <= len(script)` strictly; with `tiers` sorted ascending, `len(script) <= opsTiers[0].Threshold` guarantees a zero surcharge. Go evaluates the `countScriptOps(script)` argument first, so the early break inside `tierExcessThousandths` buys nothing — the guard must be at the call site. With your own documented `1000000:10` that's every script under 1 MB walked byte-by-byte for a guaranteed zero. Measured with the guard + the `:146` alloc removed: 2-in 67.5→11.7 ns (5.8x), 20-in 492→97.6 ns, 100-in 2285→475 ns, all to 0 allocs; the 1 MB all-ops case 1.479→0.450 ms. A companion size guard isn't worth it (`tierExcessThousandths` inlines). Separately, the switch tests the counted-opcode case last, costing five compares per byte on exactly the scripts these tiers price — hoisting `if opcode > bscript.Op16 { ops++; continue }` to the top is 3.36x (696→2340 MB/s), counts verified identical across empty, dust-return, P2PKH, well-formed PUSHDATA1/2/4, all three truncated-push cases and 100 NOPs.
11. `script_fee_tiers.go:253` — tier rejections never reach the rejected-tx Kafka topic. `ScriptVerifierGoBDK.go:358` wraps BDK's insufficient-fee as `NewTxInvalidError(errMsgInvalidTx, policyErr)`; `Validator.go:546` gates the publish on `errors.Is(err, errors.ErrTxInvalid)` and `errors/errors.go:181` matches codes exactly. A bare `NewTxPolicyError` never matches, so the byte-identical underpayment is published when BDK catches it and dropped when the tiers catch it. The PR body says the rejection plumbing is unchanged; it isn't. Wrap it the same way.
12. `services/validator/TxValidator.go:226` — `?skipScriptValidation=true` defeats the whole schedule. The `SkipScriptValidation` branch returns at `:226`, before the `!SkipPolicyChecks` gate at `:236`, and the flag comes straight off the request (`Server.go:511-517`, `Server.go:848`). Nothing requires `SkipPolicyChecks` alongside it; the one internal caller pairs them (`handle_block.go:1358`+`1367`), which is why tests miss it. Move the check above the early return, or reject `SkipScriptValidation && !SkipPolicyChecks` in `validateInternal` next to the existing `OutpointOnlySpend` guards. `validator_reference.md:208` and `validator.md:296` both document it as unconditional.
13. `script_fee_tiers.go:146` — per-tx heap allocation, and wrong capacity for P2SH. Escape confirmed (`-gcflags='-m -m'`: "escapes to heap ... from return scripts"); it's the return value, so escape analysis will never help. Measured 96 B/1 alloc at 2 inputs, 1024 B at 20, 4864 B at 100 — ~96 MB/s of garbage at 1M tx/s. And `:155`/`:162`/`:166` append three entries per P2SH input against capacity for two, so majority-P2SH txs hit `growslice` (93.4 ns / 288 B / 2 allocs at 2 inputs). The slice is consumed once at `:216`; inline the loop and both problems go away. Don't reach for `sync.Pool` or a `TxValidator` scratch field — `Validator.go:1800` runs concurrently from block-validation errgroups, so that's a race. In fairness this is ~25% on top of an allocation the BDK adapter already makes per tx (`ScriptVerifierGoBDK.go:285`), not a new dominant cost; it's worth removing because removing it is free.
14. `script_fee_tiers.go:230-236` — the discarded `util.GetFees` error serialises and double-SHA256s the tx. `util/fees.go:48` formats `btTx.TxID()` into a message this call site throws away. Measured 4.34 ms / 10.5 MB on a 10 MB tx, 42.7 ms / 100 MB on a 100 MB tx (only ~592 ns on a small one, which is why it looks harmless). Reachable with `PreviousTxSatoshis = 0` + a 1-sat output + one large script; stacks with item 9 for ~190 ms and 100 MB per tx. Sum in/out inline instead.
15. `services/validator/script_fee_tiers_test.go` — no benchmarks on a change whose premise is the per-tx path. The existing `Benchmark_validateInternal` passes `SkipPolicyChecks: true` (`Validator_test.go:962`), so nothing in the package can regress this code in either configuration. Gate on: disabled-path (0 allocs, ~2.1 ns baseline); enabled 2/20/100-in (0 allocs once `:146` is fixed); `BenchmarkCountScriptOps` with `b.SetBytes` so the switch ordering is visible; and an adversarial 10 MB all-ops case, whose 15.5 ms is the number that should have to be justified in review.

P2s and nits (uint64 underflow on an unsorted tier slice, unbounded `SatoshisPerK`, 32-bit overflow in both walkers — unreachable on supported builds, nil counters in struct-literal test validators, `settings.conf:922`'s ops example being inert and tripping your own diagnose WARN, the schedule being invisible to `settings/export.go` and unadvertised to wallets) are in a separate pass; say the word and I'll add them.

Cleared so nobody chases them: `tx.Size()` doesn't serialise (5.5 ns, 0 allocs); `blockHeight` is correctly tip+1 and matches `policy.cpp:94`; `utxoHeights` length is always right where the check runs; the `height == 0` carve-out is a real svnode rule (`policy.cpp:95`); `initPrometheusMetrics` can't double-register; and keeping the hand-rolled walk is correct — `DecodeParts`, `DefaultOpcodeParser.Parse` and `legacy/txscript.ParseScript` all allocate per push. `parseFeeTiers` is the strongest code in the PR: runs once in `NewSettings`, sort order guaranteed, ten malformed-input cases all panic with the setting name.

Calgooon:
Thank you both. This was a generous review: reproduced, measured, and with the non-issues cleared so nobody chased them, and it made the change a great deal better. Everything below was measured against the real GoBDK engine before it was changed, using `maxopsperscriptpolicy` as an oracle for the executed op count (accepted at a cap of N and rejected at N-1 pins it exactly).

@icellan, all 15 items are addressed, in your numbering:

1 and 2: `IsP2PK` and `IsMultiSigOut` are gone; the templates are matched on the bytes, with both of your reproducers kept as tests.
3: a top-level `OP_RETURN` ends the count and a nested one keeps counting, both measured.
4: the exemption now waives only the `minminingtxfee` floor; the surcharge is always due.
5: standardness follows svnode's `Solver` in the coin's era (P2SH before Genesis only, `OP_FALSE OP_RETURN` after, any 33 to 65-byte push as a pubkey), with differential tests against BDK.
6: `OP_CHECKMULTISIG` is charged its key count where that is statically certain (depth zero, still executing, a literal push immediately before it) and one op otherwise, and the comments now say "under-counts" where they said "exact". You were right that no static walk can be exact; the fuzz described below measures how far it sits from svnode.
7: zero-valued `maxconsolidationinputscriptsize` and `minconfconsolidationinput` are normalised to 150 and 6; the 0 to 6 rewrite is measured.
8: the redeem is billed only for coins created before Genesis. On the question you left open: the coin's height decides, not the spending tip. A pre-Genesis coin still runs its redeem when spent today and a post-Genesis coin never does, both measured, and the comment says so.
9: a script beyond the caps BDK enforces is left unpriced so BDK's own rejection surfaces. For a pre-Genesis coin those are the fixed 500-op and 10000-byte limits, which BDK applies in place of the policy caps (measured).
10: the length guard and the switch order are in, with 0 allocations at 2, 20 and 100 inputs.
11: tier rejections are wrapped the way BDK's insufficient-fee rejection is, so they reach the rejected-tx topic.
12: `SkipScriptValidation` without `SkipPolicyChecks` is now rejected, next to the `OutpointOnlySpend` guards.
13 and 14: the executed-script slice and the `GetFees` call are gone.
15: benchmarks cover the disabled path, 2, 20 and 100 inputs, and `countScriptOps` with `SetBytes`, including the 10 MB all-ops case.

I would be glad to have the P2 list whenever you have a moment.

@raspi-user, thank you as well:

1: tier and consolidation arithmetic saturates instead of wrapping, so a parse-valid but bad configuration can only raise the required fee, never lower it.
2: the gap is documented in the setting's own description so an operator sees it before enabling the schedule, and extending the P2P `fee_policy` message is the intended follow-up.
3: the exemption predicate is now held to "never stricter than BDK" and measured that way. Where it is more permissive, the cost is bounded to the byte-rate floor, which BDK re-checks one call later, and those cases are pinned in tests.

After that I put the walker through several rounds of adversarial review by separate models and measured every finding against BDK. That turned up four more defects, all fixed and pinned: a multisig that counted more ops than it had bytes and slipped under the length guard, push lengths assembled in a signed int, `lastPush` treating `OP_1` to `OP_16` as non-pushes, and, for coins created since the Chronicle upgrade, `OP_VERIF` and `OP_VERNOTIF` being conditionals. Two fuzzers now compare the static count with BDK's on 20000 random scripts per run in both eras, one requiring exact agreement and one asserting it never counts more than svnode.

The branch is rebased on main. Squash-merge is fine by me, and if any of this would be better done differently I am happy to change it.

icellan (changes requested):
Re-reviewed at `85126b0e0`. **All 18 round-1 items are genuinely fixed** — I re-derived each against current code rather than trusting the commit messages. All five new byte matchers and `lastPush` are index-safe, so both P0s are dead; the surcharge is always due; the `NewTxInvalidError(_, NewTxPolicyError(_))` wrap does reach `ErrTxInvalid` through `errors.Is`; the arithmetic saturates; the hot path is 0-alloc. I also reran both of your new fuzzers at 20,000 scripts against real GoBDK and got your exact counts with zero divergences. Good response.

Round 2 finds a different class. The four sharpest are measured, not argued. Numbers below are from probes against real GoBDK (`v1.2.5-...cdfa7814ee5d`) and benchmarks I ran locally; nothing was committed.

One root cause ties several of these together: **three BDK limits are guessed in Go — max ops per script, max script size, max pubkeys per multisig — while BDK exports a getter for each** (`bdkcgo/txvalidator_cgo.h:68,70`, `script/txvalidator.go:237`), all narrowed away by the `bdkValidator`/`bdkNativeTxValidator` interfaces (`ScriptVerifierGoBDK.go:246,250`). That is also why 500/10000 are hardcoded at `script_fee_tiers.go:118-119`. Items 1 and 15 share one fix shape.

**P0**

1. `script_fee_tiers.go:91,101` — the zero-cap model disagrees with BDK. `opsCap`/`scriptSizeCap` map a `0` cap to `math.MaxUint64`; probed, `SetMaxOpsPerScriptPolicy(0)` then `GetMaxOpsPerScript(true,false)` returns **4294967295**, and the size pair behaves the same. With `maxopsperscriptpolicy = 0`, a script whose counted ops exceed 2^32-1 gets `capExceeded = false`, so it is priced and rejected as "insufficient fee" while BDK rejects it with `SCRIPT_ERR_OP_COUNT` — the retry-forever misdiagnosis the cap carve-out was added to prevent. `TestScriptSizeCapZeroDifferentialBDK:786-788` asserts `require.Equal(math.MaxUint64, scriptSizeCap(...))` and claims "0 is not a rewrite to some default"; that is false, and the test is swap-proof — its 600,000-byte script is accepted under both candidate values, so substituting the BDK-accurate 4294967295 leaves it green. `opsCap`'s identical zero-cap logic has no differential or unit coverage at all.
2. `script_fee_tiers.go:470` — the `opCap` early-out is vacuous in exactly the configuration this feature exists to accompany. The ops longdesc says raising the cap to UINT32_MAX removes the ceiling and this setting replaces it with a price, so the intended pairing is `maxopsperscriptpolicy` at 0 or 4294967295 plus a tier — at which point `if ops > opCap` can never fire for any submittable transaction. Measured `BenchmarkCountScriptOps/adversarial_10mb_ops` = 17.05 ms at 614.89 MB/s, so ~85 ms for a 50 MB all-ops unlocking script and ~170 ms for 100 MB, single-threaded, before BDK. `validator_httpBodyLimit` defaults to 100 MB and the validator's own endpoint (`Server.go:1025`) applies no `maxtxsizepolicy` pre-check — that limit is enforced only inside BDK, which runs after. ~6 req/s pins a core, rejected, no fee. With the *default* cap of 1000000 the walk self-limits at ~1.6 ms, so this is purely the recommended-config case. Needs a length bound as well as a cap bound, since counted ops exceed script length only via the multisig key count.
3. `script_fee_tiers.go:656` + `TxValidator.go:236` — the check answers with a retryable verdict for transactions BDK classifies as permanently invalid. `bdkPolicyRelatedScriptError` (`ScriptVerifierGoBDK.go:390-397`) lists exactly six codes: `OP_COUNT`, `SCRIPTNUM_OVERFLOW`, `SCRIPTNUM_MINENCODE`, `SCRIPT_SIZE`, `PUBKEY_COUNT`, `STACK_SIZE`. Everything else — `SCRIPT_ERR_EVAL_FALSE` from a garbage signature, `SCRIPT_ERR_UNBALANCED_CONDITIONAL`, disabled opcodes — falls to `:387` `NewTxInvalidError(_, bdkCause(ERR_TX_INVALID, _))`, so `errors.Is(err, ErrTxPolicy)` is false and `publishPolicyRejectedTx` does not fire for them today. The tier check runs first (`TxValidator.go:236` precedes `:247`) and returns a `TxPolicyError`, for which `errors.Is(err, ErrTxPolicy)` **is** true. So a transaction with a bad signature that also underpays is told "insufficient fee: N paid, M required" and retries forever, re-walking its scripts each time, for something that can never be valid — and its raw bytes are published to `KAFKA_TX_POLICY_REJECTED` (`Validator.go:591`), documented at `:599-602` as consensus-valid transactions rejected by local policy, for subtree-validation pods to cache. The cap carve-out reasoning generalises: only price after BDK, or only for verdicts BDK itself classifies as policy.
4. `services/validator/Validator.go:629` vs `:640` — the publish guard and the payload measure different things. `tx.Size()` (go-bt `tx.go:766`) computes version+locktime+varints+inputs+outputs with no `PreviousTxScript`; `SerializeBytes()` (`tx.go:418`) returns `ExtendedBytes()` when the tx is extended, which it is by the time `publishPolicyRejectedTx` runs. A transaction spending 100 previously-created outputs each carrying a 1 MB locking script has `tx.Size()` ~15 KB — far under the `validator_kafka_maxMessageBytes` default of 1048500, so the guard passes — and a ~100 MB payload, copied again by `proto.Marshal`: ~200 MB of allocation per rejected submission on the concurrent validation hot path, before `TryPublish` silently fails or the broker rejects it. The tx is rejected so its inputs are never spent, and the same bytes can be resubmitted without limit and without ever paying a fee. The publisher bug is pre-existing (reachable today by underpaying `minminingtxfee` by a few satoshis), but this PR widens the trigger to the intended common case once a schedule is enabled.

**P1**

5. `cmd/diagnose/config_checks.go:666` — `checkFeeTiers` gates its WARN on `capValue > 0`, so `maxopsperscriptpolicy = 0` — the one configuration where the walk bound is gone — reports INFO all-clear. An operator setting the cap to 0 per the setting's own longdesc guidance, pairing it with `minminingtxfeebyscriptops = 1000000:10` and running `teranode-cli diagnose` is told the pairing is fine while `opsCap` has silently become `MaxUint64`. Only the cap-too-low direction is covered.
6. `cmd/teranodecli/teranodecli/cli.go:158` — `NewSettings()` runs before the command switch (diagnose is dispatched at `:301-307`), and `settings/settings.go:105-106` calls `parseFeeTiers`, which panics at `settings/policy_settings.go:37/42/47/61/65` on any malformed entry. So an operator who writes `minminingtxfeebyscriptsize = 500000` (missing `:rate`), `500000:10|400000:20` (decreasing rate) or `0:10` and runs `teranode-cli diagnose --config` to find out why gets an unrecovered panic with a stack trace — the new `checkFeeTiers` reporting is unreachable in exactly the case it exists for. The same panic also takes down every unrelated subcommand (filereader, aerospikereader, export-blocks).
7. `script_fee_tiers.go:472` — `capExceeded` returns 0 for the **whole** script rather than just the ops tier, so enabling `minminingtxfeebyscriptops` can *lower* the fee `minminingtxfeebyscriptsize` requires for the same transaction. Measured twice. Post-Genesis: with `sizeTiers=[{1000,10}]` alone, a 100,006-byte script whose counted ops exceed the coin's op cap (`<push4 0x7fffffff> OP_CHECKMULTISIG` plus 100,000 bytes) prices at 990,060 satoshi-thousandths; adding `opsTiers=[{1000,10}]` makes `priceScript` return 0 for the same script. Pre-Genesis: with `sizeTiers={100:1000}`, `opsTiers={10:1000}` and the fixed 500-op cap, a 2000-byte script reaching ops=501 prices at 0 where the size tier alone charges 1900 satoshis. The comment justifies the early-out purely as not pre-empting BDK's op-count rejection, which argues only for skipping the ops term. Sound today only because Go's count is never above BDK's on a script BDK accepts — and items 13 and 15 are where that breaks, at which point this line converts an over-count into a total fee waiver across both metrics. The docs present the two schedules as independent.
8. `settings/export.go:221` — `formatValue`'s `reflect.Slice` case handles only `Elem().Kind() == reflect.String` and otherwise falls through to `fmt.Sprintf("[%d items]", val.Len())`. These are the repo's first `[]<struct>` settings, so an operator who sets `minminingtxfeebyscriptsize = 500000:10|5000000:50` and opens the asset settings endpoint (`settings_handler.go:75` -> `ExportMetadata`) or the dashboard sees CurrentValue `"[2 items]"` — no thresholds, no rates. Combined with the schedule being absent from `/v1/policy` (`get_policy.go:53`) and from the Teranode-owned p2p `fee_policy` (`services/p2p/message_types.go:52`, whose struct already carries `MaxScriptSizePolicy`/`MaxTxSizePolicy`/`MaxTxSigopsCountsPolicy`, so the seam exists), the schedule that decides rejections is readable from no running-node surface.
9. `script_fee_tiers.go:641` — the exemption counter counts the wrong thing. The `Inc()` sits inside `if fee >= required { if exempt { ... } }`, so it fires for a consolidation paying ten times the requirement that never needed the exemption; and the whole block is unreachable when `surchargeThousandths == 0` (early return at `:618`), i.e. for every consolidation whose scripts sit below the thresholds — which is all of them, by the file header's own argument. Meanwhile `metrics.go:70-74` says "Underpaying transactions accepted ... because they qualify as free consolidations" and tells the operator a surge relative to rejections is worth a look, and `prometheusMetrics.md` repeats it. So a node with normal large-script consolidation traffic shows a permanently rising counter with no abuse occurring, and shows nothing when waiving the floor is what actually admitted the tx. The correct gate compares `fee` against the un-waived `(floorThousandths+surchargeThousandths)/1000`.
10. `script_fee_tiers.go:656` — the rejection borrows `errMsgInvalidTx` = "GoBDK fail to ValidateTransaction" for a check BDK never ran. That constant (`ScriptVerifierGoBDK.go:24`) is otherwise used only by the BDK adapter — its own `Warnf` prefix at `:307` and the outer message on all four real BDK rejection paths (`:357,:360,:366,:383`), with `ScriptVerifierGoBDK_test.go:310` pinning `errors.UserMessage` to exactly that string — and `checkScriptTieredFees` returns before `tv.bdk.ValidateTransaction` is called. The resulting "GoBDK fail to ValidateTransaction: insufficient fee: 3000 satoshis paid, 4200 required by the per-script fee tiers" is what `Validator.go:571` puts into `KafkaRejectedTxTopicMessage.Reason` and broadcasts, and what lands in logs, so an operator triaging a rejection spike greps for GoBDK and starts in the cgo bridge and the BDK version when the cause is a local setting. The Kafka parity needed the `NewTxInvalidError(_, NewTxPolicyError(_))` shape, not the borrowed message.
11. `services/validator/Validator.go:776` — this is my fault as much as yours: my round-1 finding offered "move the check above the early return, **or** reject the flag combination in `validateInternal`", and the second option doesn't close the hole at the layer that owns the policy. Three problems. (a) `TxValidator.go:156`'s `if SkipScriptValidation { ... return nil }` still precedes the `if !SkipPolicyChecks` block at `:234-238`, so any caller not going through `validateInternal` bypasses the tier — `test/consensus/validator_integration.go:84` and `test/longtest/services/validator/input_test.go:41` call `TxValidator.ValidateTransaction` directly, and `TxValidator_test.go:156-170` now asserts `NoError` for `Options{SkipPolicyChecks:false, SkipScriptValidation:true}`, pinning the bypass. (b) `ProcessingError` maps to `codes.Internal` (`errors/errors.go:1098`) and is off `publicCauseCodes`, so a client sending `skip_script_validation=true` without `skip_policy_checks` is told "error processing" — indistinguishable from a node outage — and its retry logic treats a permanently invalid request as transient; `NewInvalidArgumentError` maps to `codes.InvalidArgument`. (c) `validator_api.proto:76`, `Server.go:848`'s `?skipScriptValidation` query param and `options.go:52-58` all still describe the flags as independently settable, unlike the sibling `outpoint_only_spend` field one proto line below which documents its invariant plus a deployment-skew note — so during a rolling upgrade the same request succeeds on old instances and fails on new ones behind one load balancer.
12. `script_fee_tiers.go:516` — `txFee`'s doc comment asserts the totals are bounded by the money-range checks BDK or the Go backstop applies. That does not hold on the path that reaches it: `tv.checkOutputValues` and `tv.checkInputValuesAndInflation` (`TxValidator.go:257/290`) run only inside the `SkipScriptValidation` branch, which returns at `:226`; `checkInputs` (`:445`) checks only cumulative prevout size, and only when `maxcoinsviewcachesize > 0` (default 0); BDK runs after at `:247`. go-bt's `IsExtended` checks `PreviousTxScript != nil` and never `PreviousTxSatoshis`, so a submitter posting an already-extended tx controls both sums — two inputs of 2^63 satoshis wrap `in` to 0. Then `out > in` makes `ok=false`, `checkScriptTieredFees` returns nil at `:626`, and the surcharge, the rejection counter and the exemption counter are all skipped; the tx dies later as bad-txns-in-belowout, so the operator sees zero `script_tiered_fee_rejections` while the tier check never ran on a class of submissions it was meant to price. `ok=false` should defer the fee comparison, not waive the surcharge and both counters.
13. `script_fee_tiers.go:255` — an `OP_ENDIF` at `ifDepth == 0` is silently ignored where svnode fails the script with `SCRIPT_ERR_UNBALANCED_CONDITIONAL`, producing a measured **over**-count. `OP_1 OP_1 OP_IF OP_ENDIF OP_ENDIF OP_RETURN <50 OP_NOP>`: `countScriptOps` returns 4 — the second `OP_ENDIF` is swallowed by the `if ifDepth > 0` guard at `:256`, leaving `ifDepth` at 0, so the following `OP_RETURN` is read as top level and counted — while svnode counts 3 (IF, ENDIF, ENDIF) and then errors, never fetching the `OP_RETURN`. This script does reach BDK, so it falsifies the header claim at `:26-33` that the walk matches svnode's executed count exactly for every script that reaches BDK with `OP_CHECKMULTISIG` as the one documented exception. Over-counting is the direction the file says must never happen, and it is harmless today only because BDK rejects every script that can reach this state — the invariant is enforced by coupling to BDK's conditional-balance check, not by the walk itself. The bogus `ifDepth` also persists for the rest of the walk, so a genuinely nested `OP_RETURN` downstream can be read as top level. Combined with item 7, any future relaxation turns a phantom depth into a total fee waiver.
14. `script_fee_tiers.go:570` — this answers the round-1 question about who pays for output scripts, and the answer is nobody. With the documented `minminingtxfeebyscriptsize=500000:10` and the shipped `minminingtxfee`, a transaction creating a 100 MB `OP_FALSE OP_RETURN` output pays the byte-rate floor alone and no surcharge, whereas *spending* a 100 MB locking script owes roughly 995,000 satoshis. Because `OP_FALSE OP_RETURN` outputs are provably unspendable, that surcharge is never collected from anyone — so the settings structurally cannot price the shape that most bloats blocks and the UTXO commitment. `maxscriptsizepolicy` is not create-side either, so the pairing is internally consistent; the problem is that `settings.conf` and the `policy_settings` longdesc say the setting prices the metric its policy cap gates and that a large-script output cannot be created cheaply and then consolidated to dodge it. It can be created cheaply and simply never spent. Say so in the operator docs, or price the create side.
15. `script_fee_tiers.go:267` — the multisig key count is added to the tally with no check against the pubkey limit BDK enforces, so a surcharge is priced off a number BDK's interpreter never computes. Measured from real BDK: `GetMaxPubKeysPerMultiSig(genesis=false, cons=false)` = 20 (the fixed pre-Genesis limit); `(genesis=true, cons=false)` = 4294967295 with `maxpubkeyspermultisigpolicy` at its default 0. Consequence: the 3-byte pre-Genesis script `{0x01, 21, OP_CHECKMULTISIG}` makes `countScriptOps` return ops=22 (1 opcode + key count 21), well under the pre-Genesis `opCap` of 500, so `capExceeded` is false and an ops surcharge is priced off 22. Real BDK rejects the identical script at the opcode with "Pubkey count negative or limit exceeded" — the same verdict at `opCap` 500 and at 21, proving its counter never advances past the bare fetch (1). A 21-op over-count on three bytes: the second falsification of the header's exactness claim, more reachable than the `OP_ENDIF` case, and via item 3 the submitter is told to pay more instead of being told their multisig has too many keys. Fix mirrors what the op and size caps already do: clamp against `GetMaxPubKeysPerMultiSig` and decline to price when it is exceeded.

**Refuted, so nobody chases them**

The suspected `bdkBlockHeight` -1 off-by-one: probed at minConf=6 across 4/5/6/7 confirmations, exact agreement. The extended-format prevout "bypass": `SkipUTXOHashCheck` is OutpointOnlySpend-only, so the spend-time hash check rejects any lie. The `postChronicle=false` invariant: holds on all six go-chaincfg v1.6.2 networks.

**Smaller items, not blocking**

`checkFeeTiers` adds 34 lines of branching with zero tests, though `config_checks_test.go` already has a `findResult` helper built for it (AGENTS.md: "Write or adjust tests with the change"). `isPreGenesisCoin` (`:608`) and `isBDKPreGenesisCoin` (`:578`) disagree on the same coin three lines apart on sentinel and height 0 — under-charge only, but two same-shaped names with opposite semantics. `script_fee_tiers_fuzz_test.go:94` cites `TestCheckMultiSigUnderCountDifferentialBDK`, which does not exist; the real test is `TestCheckMultiSigKeyCountDifferentialBDK`. The ops longdesc example at `settings/policy_settings.go:94` still equals the `maxopsperscriptpolicy` default, so it is inert and trips the WARN this PR added, against the same longdesc's Recommendations bullet — `settings.conf:926` gets it right. `smallIntPushData`/`negOnePushData:347-360` avoid an allocation for a value provably always priced at zero, and `lastPush:384` hands slices of that shared mutable state to a concurrent hot path. The push decoder is copy-pasted between `countScriptOps:289-331` and `lastPush:394-428` — the signed-int and PUSHDATA4 fixes already had to land in both. `isP2PKHScript`/`isP2SHScript`/`isDataScript` duplicate go-bt helpers that are already byte-only and panic-safe, so the round-1 P0 rationale does not apply to them; `:608` already calls `IsP2SH()` for the question `:753` asks with the private copy. `//nolint:gosec` at `:937` answers the sign, not the `int -> uint32` truncation G115 flags. ~7% of each fuzz corpus is unmeasurable by the op-count oracle, so that slice is unpinned in both directions, and the fixed seed means the 120-script CI corpus explores the same prefix forever. `validator_reference.md:209` still lists `pushDataCheck`, which is not in the tree.

The CI reviewer's thread r3892731039 (pre-Genesis P2SH redeem bytes paying the size tier roughly twice, ~520 bytes bounded) is still open and pinned by no test either way.

PR #1663, opened 2026-08-27
feat(utxo): add utxoset, a delete-on-spend UTXO store

freemans13:
## What happened

The postgres UTXO store records a spend as the **presence** of a row in an append-only table, so the thing it indexes is every spend ever made. That set only grows. It reached 392,884,647 rows and roughly 30 GB of index against 16 to 20 GB of cache, and once an index stops fitting in memory every lookup becomes a disk seek. Sync rate halved, then halved again: 11,563 blocks a day, then 5,317, then 2,529.

This store inverts what is indexed. It holds one row per live unspent output, a spend is a `DELETE`, and a double spend is detected by the row not being there. The set it indexes is the live one, which is roughly flat rather than growing.

That single `DELETE` does four jobs at once. It arbitrates the double spend, because zero rows affected means the coin is already gone. It is the decorate fetch, because `RETURNING` hands back the satoshis and locking script, so script validation never parses a parent transaction. It is the reclaim, because the row is gone. And it is the write.

**This is additive.** 37 files, 5,861 lines, nothing deleted or modified. Two of those files register the store in the factory; the rest are the new package. Selecting it requires an explicit `utxoset://` URL, so nothing changes for anyone who does not opt in.

## What is here

**The coin table.** One row per spendable output, split eight ways by the first byte of the transaction id. The key is the first 12 bytes of the transaction id plus the output number, and it is deliberately **not** unique: a 96-bit prefix can collide, and a unique index on it would be a consensus bug, because roughly 2^48 of attacker work would make a legitimate output fail to be created. Identity is the full 32 bytes carried on the row and rechecked in every predicate, so a collision costs one extra heap visit and never a wrong answer.

**The spend journal**, which is svnode's `rev*.dat` in a table. A delete-on-spend store destroys the row, so a reorg has nothing to restore from unless the payload is captured at the moment of deletion. The journal row is written in the same **statement** as the delete, not merely the same transaction, so there is no window in which a coin is gone with nothing recording how to put it back. It is also the prune engine, and that is why it has no off-switch: without it nothing can be reclaimed during the entire initial sync.

**The transaction tables.** One permanent row per transaction holding identity, block membership and the bounded fields, plus a separate table holding just the serialized bytes, split by height so old ones are discarded by dropping a file rather than deleting rows. That split matters: the permanent row is pinned for as long as any of its outputs is unspent, at any age, so keeping the bytes there too would make the transaction archive permanent for the whole pinned population.

**The pruner session.** Each retiring journal partition is read as a work list before it is dropped, because dropping it destroys the only record of which transactions had an output spent in that window. A transaction row is deleted only when it holds no unspent output, is itself on the main chain, and every transaction that spent it is buried past the depth at which the node could un-mine it.

## Trade-offs and limits, stated plainly

**The transaction table cannot be reclaimed by dropping partitions**, only row by row, because a row dies when its last output is spent and that spend settles, which happens at any age. Measured on PostgreSQL 18.6, its index is 79.9 bytes per row when freshly built and 125.6 after eight cycles of 50% turnover, still climbing. At the projected tip that is 24 to 38 GB against a 25 GB index budget, before the coin table's own projected 40 GB. **This design does not solve that.** It is the store's known unsolved cost.

**Nothing here has run against real load.** An earlier version of this branch synced mainnet from genesis and reached height 686,010 with a 5.54 GB index against 24 GB of memory and the unspent set flat at 45 to 47 million rows across 240,000 blocks. That run exercised the coin table only. No transaction table existed then, so none of the transaction-window work in this PR has seen a real block.

**Storage figures are extrapolated.** They were measured at 300,000 to 43,000,000 rows in a container whose page cache held every object built, so no measurement observed a real disk read. Page-fetch counts should transfer; timings will be worse on real hardware.

**Spender identity is bounded by the journal's retention.** A double spend names the transaction that took the coin, read from the journal. Beyond retention the store genuinely cannot say who took it, and reports the double spend without naming a winner.

**Two claims were removed from the package comment rather than softened**, because neither had a source: a projected index size at the tip, and a write-amplification ratio against append-only. The tip UTXO count is the least certain input in the whole design and nobody has run `gettxoutsetinfo` on a synced SV Node to settle it.

## Testing

- 64 tests in the package, no skips, run against PostgreSQL 18.6, which is the version the mainnet box runs.
- The store-agnostic conformance suite in `stores/utxo/tests` passes in full: Store, Spend, Freeze, SetMined, Conflicting. That suite is the specification, and it is the same one the sql and aerospike stores are held to. It is not gated or opted out of.
- Every test written before its implementation and watched fail first. Where a test passed for the wrong reason it was tightened and re-watched failing; one asserted only "an error" where a missing table satisfied it, and now requires the specific constraint violation.
- `go build ./...`, `go vet`, `-race` and `golangci-lint` clean.
- Run on two separate databases each time, one shared and one created fresh, because a stale schema in a shared database masked two failures earlier in this work.

**Not tested:** any of this under concurrency at load, the reclaimer against a long-running reader holding a snapshot open, autovacuum behaviour under sustained churn, or recovery from a crash mid-session.

https://claude.ai/code/session_01P1s8eyLfgGLmDcdzKHu6HG

## Update 2026-09-06: the SV Node shape, and no transaction bodies below the checkpoint

The branch now carries the store as it runs on the Hetzner mainnet box, 50 commits on from the description above. The shape moved to what SV Node's chainstate does: coins in the database, bodies elsewhere.

**Membership by block, not by a mutable flag.** A transaction's block membership is a row per (transaction, mined height, block id) in a table split into 288-block windows; mempool identity lives in its own table; the spend journal carries the coin's block facts so the pruner can read a retiring window as a work list. The reclaimer that this replaced leaked half of what it should have freed and cost more CPU than the writes it cleaned up after.

**Statement shapes that survive postgres's planner.** Per-leaf scalar lookups over the identity table, `unnest` plus `CROSS JOIN LATERAL ... LIMIT 1 OFFSET 0` over the partitioned tables, every coin write re-checking the transaction id, bit masks written as `(flags & n) < 1`, and `plan_cache_mode = force_custom_plan`, each there because the generic plan or the bit-mask estimate walked a whole table on the sixth execution.

**Conflict-intent WAL, reassign, mined-transaction conflict reads, the legacy sibling-block flag, a Close that stops every batcher, and all 28 shared conformance suites** are implemented; nothing is stubbed.

**Batcher settings honoured.** The store now reads the same `utxostore_*Batcher*` drain-mode and greedy-accumulate settings the other stores read.

**No transaction body below the checkpoint** (`utxostore_skipTxBodyBelowCheckpoint`, default off). At mainnet height 705k the store was serialised on postgres's WAL write lock, and 80 percent of WAL bytes were the raw transaction bodies being inserted; the coin, membership and journal rows are about 100 bytes each. SV Node's chainstate holds no bodies, and this store's own pruner was already dropping body windows past a 288-block horizon, so the bodies were a two-day rolling buffer that cost a full WAL write per transaction. With the setting on, a transaction mined at or below the highest hardcoded checkpoint gets coins, membership and journal rows but no body row, decided by the shared `model.BelowCheckpoint` predicate (genesis keeps its body). Decorate reads coin rows, so spends at the tip are unaffected. What is given up: block persister must find the subtree data file legacy sync wrote (288-block retention) or its rebuild fails; the asset service's legacy block endpoint cannot serve those blocks; a non-extended child of a body-less parent fails validation above the checkpoint; there is no backfill. On mainnet the change took body inserts from 73,500 a minute to zero and WAL per transaction down about tenfold; the WAL write lock stopped being the dominant wait.

**Measured on Hetzner mainnet (32 GB, one disk), from genesis:** the store branch merged with the IBD pipeline branch (bsv-blockchain/teranode#1699) reached height 721k on 2026-09-06 with no restarts and the pruner keeping up. Throughput ranged 13k to 21k tps through 640k, fell to 1k to 5k tps in the heavy era above 700k while WAL-bound, and read 10.6k to 11.8k tps at 721k after the body write was switched off.

**Limits.** Fresh sync only; the schema is not a migration of the earlier utxoset layout. Tests run on postgres via testcontainers. The 288-block body horizon and the no-backfill rule are operator-facing and documented on the setting.

**Follow-up 2026-09-07: presence is the record, not the body.** The first restart after switching the body write off produced a false "MANUAL INTERVENTION REQUIRED: coinbase divergence" alarm from block assembly: its coinbase presence check read the coinbase with the body field and treated a nil body as a missing coinbase, so every coinbase created since the switch read as missing and the walk-back found no good floor. Nothing was repaired or damaged, but startup divergence detection had disabled itself. Every reader of the body field in the services was then classified: three used the body as a presence or validity proxy and are fixed (coinbase recovery now asks for block ids; conflict-parent unlocking and unmined-input validation read the stored inpoints first, fall back to the body, and return an error when a record has neither), one panicked-then-recovered in the asset service's legacy block writer and now returns a named error, and four error messages now name the retention reason and the setting. Two things surfaced on the way: unmined-input validation had been dead on every SQL-backed store (it returned false on the first line, dropping every unmined transaction at reload), and making it live exposed a read-inside-write-transaction shape in the SQL store's `SetConflicting` that self-deadlocked on SQLite's shared cache and cost postgres two pool connections per call; the reads are hoisted out of the transaction, with the widened non-atomic window stated in the comment, and the SQLite cascade test that had been suppressed for that deadlock is enabled. A note for readers of the utxoset schema: the mined create path writes NULL inpoints for every block transaction below the checkpoint, so consumers must never rely on inpoints or body for those records; presence means the record exists.


freemans13:
Sonar: both new issues addressed in 097b37a8b. They were `go:S1186` on `journalPruner.Start` and `journalPruner.AddObserver` — the doc comments above each already said why they do nothing, but the rule reads inside the braces, and so does anyone looking at the body rather than the signature. Comments moved in.

The 0.2% coverage on new code in the earlier run was not a measurement artefact, and it is worth naming properly. Every test in `stores/utxo/utxoset` was skipping in CI, and so was `TestUtxosetSchemeIsRegistered` in `stores/utxo/factory`. Both helpers connected to `localhost:5441` and called `t.Skipf` when nothing answered; CI has nothing listening there, so all 66 tests skipped on every run and the store was verified nowhere but my laptop. The `test` check was green throughout, because a package that skips is a package that passes.

Fixed in cf4585e3c and 33aff0e87: the DSN now comes from `UTXOSET_TEST_DSN` when set, and otherwise from the shared `postgres:16-alpine` container the rest of the repo's integration tests use. The reachability skip is gone. The only skip left is the repo-wide `SkipIfContainerUnavailable`, which fires when Docker itself is missing and hard-fails for every other reason.

Verified rather than assumed — from the CI `test` job log on 33aff0e87:

```
✓  stores/utxo/factory (5.968s)
✓  stores/utxo/utxoset (14.033s)
```

zero `SKIP` lines for either, against 0.4s of pure skipping before. Sonar now reports 80.4% coverage on new code and the quality gate passes.

Also in this round: 130e9e2b4 fixes the build against the `SetBlockState` method upstream added to `utxo.Store`. The package was carrying its own height and median-time atomics beside a `GetBlockState` that returned an empty `utxo.BlockState` no matter what had been set, so anything reading the tip through the snapshot saw height 0. It now embeds `utxo.BlockStateFields` like the other stores, which deletes both atomics and makes `GetBlockHeight` and `GetBlockState` read the same memory.

Benchmark comparison: 0 regressions, 0 improvements, 136 unchanged — expected, since nothing outside the new package is touched.

freemans13:
Brought this branch up to date with main and fixed a validator gap that stops a node running this store.

## What happened

Main merged the fix for GHSA-v76m-6vc7-g7c7. The UTXO commitment hashes a locking script followed immediately by its value, with nothing marking where the script ends, so a forged shorter script with a larger value can match a real one. The fix makes the validator always overwrite each input's script and value from the node's own copy of the parent, and never use values a transaction arrives carrying.

The validator gets that copy by asking the store for the parent's outputs. This store answers without a body for a parent it no longer holds a body for: bodies are dropped 288 blocks after creation, and with `utxostore_skipTxBodyBelowCheckpoint` on they are never written below the highest checkpoint. The validator then rejected the transaction as "body not retained". That already applied on this branch to every transaction that arrived without parent values; main's fix extends it to the ones that arrived with them. The practical effect was that a node on this store rejects nearly every spend of an output more than about two days old, and on mainnet would stop at the first block above 945,000.

## Each change and why

**Merge of upstream/main.** Two conflicts. In the validator, main's unconditional overwrite is kept. main also added `DeleteComplete` to the store interface for backends where `Delete` leaves child records behind; this store keeps none, so it delegates to `Delete` exactly as the SQL store does.

**`fix(validator): fill body-less parents from the utxoset coin table`.** When the store returns a parent without a body, the validator fills those inputs through `PreviousOutputsDecorate`, which on this store reads value and script straight from each coin row by transaction id and output index. It clears the values the transaction arrived with first, because decorate skips any input that already carries a script and would otherwise keep a forged pair. It decorates a private copy holding only that parent's inputs and copies the results back, because the validator resolves every parent of a transaction concurrently against the same transaction and decorate walks every input it is given. An input left empty afterwards fails the transaction: a spent or missing coin as a missing parent, a reassigned coin as a processing error.

The fix is in the validator rather than the store because the coin table is indexed only on the packed transaction id and output index, with no index on the id alone and no stored output count, so a store lookup handed only a transaction id cannot find its outputs without a scan. The validator already knows exactly which outputs it needs.

## Testing

- Validator unit tests: coin found and supplied values overwritten; coin gone; coin reassigned; several parents at once with one body-less, under `-race`; a parent with a body unchanged; the existing GHSA regression tests unmodified and passing. Mutation checks on the clearing step, the private copy and the empty-input guard each fail the test that names them.
- Real store: `TestValidatorFillsBodylessParentFromCoinTable` builds a real utxoset store on a Postgres testcontainer with body-skipping on, mines a signed parent below the checkpoint, confirms it has no body, and validates an honest signed child end to end through `ValidateWithOptions`. The package's own decorate tests pass against Postgres too. These tests skip silently unless testcontainers can find Docker; on OrbStack that needs `DOCKER_HOST=unix://$HOME/.orbstack/run/docker.sock`.
- Behaviour difference worth knowing: a spend of an already-spent output from a body-less parent now reports as a missing parent, where a parent with a body reports the output as spent. Delete-on-spend leaves no coin row to distinguish the two.
- Two `services/blockassembly` tests, `TestReplayPendingConflictIntents_ReverseReplayCompletes` and `TestReplayRecoveryAerospikeAssemblyEndToEnd`, fail identically on this branch before the merge and on main.

https://claude.ai/code/session_01C6jxW7ai4L7pvtdZGEuWE9

PR #1664, opened 2026-08-28
p2p: make catchup malicious reports visible to IsPeerMalicious and escalate to ban score

gokutheengineer:
**Closes bitcoin-sv/teranode#4754.**

### Problem

`RecordCatchupMalicious` recorded malicious behaviour only as `MaliciousCount++` in the peer registry (which pins reputation to 5.0), while `IsPeerMalicious` answered from ban state alone. Since no catchup malicious path ever added ban score, the recorded flag was write-only: every `isPeerMalicious` gate in blockvalidation catchup (header-fetch abort points, `processBlockFound`, `tryAlternativePeersForCatchup`) stayed open for a peer caught serving malicious headers, violating checkpoints, or secret mining. `ReconsiderBadPeers` later reset `MaliciousCount` to 0, so even the indirect reputation pin expired with no escalation. All cited locations confirmed still present on main.

### Fix

Both halves of the issue's proposed fix, since each covers a hole the other leaves:

- `IsPeerMalicious` now also reports `IsMalicious: true` (distinct reason) when the peer's `MaliciousCount > 0`, after the existing ban check. This is what lets catchup abort on a peer flagged mid-operation, and it auto-recovers when the sync coordinator's `ReconsiderBadPeers` sweep clears the record after its (exponentially growing) cooldown.
- `RecordCatchupMalicious` now also raises the ban score under a new `catchup_malicious` reason (50 points; ban threshold 100, decay 1/min) via the shared `applyBanScore` helper, so repeat offenders cross into a real ban instead of being re-admitted forever. The charge is throttled to one per peer per 10-minute window, because a single invalid block is reported through up to three blockvalidation paths in the same catchup cycle (direct report, deferred lock release, unvalidatable-peer handler) and would otherwise insta-ban on one offense. The score write is best-effort (logged on failure); a new ban transition triggers the standard `onPeerBanned` disconnect path.
- `catchup_malicious` added to `DefaultBanConfig().ReasonPoints`, the p2p `AddBanScore` known-reason switch, and the `p2p_ban_threshold` setting's penalty table.
- Updated the now-stale "banning not yet implemented" log lines and the Peer Banning sections in `docs/topics/services/blockValidation.md` / `docs/references/services/p2p_reference.md`.

### Design notes (from review)

- **Recovery invariant**: the whole recovery story leans on `UpdateMetrics(recordMalicious=true)` setting `LastInteractionFailure` and pinning reputation to 5.0 - together those clear both `ReconsiderBadPeers` eligibility gates (`ReputationScore < 20`, non-zero last failure). If either were removed, a malicious record would become a permanent exclusion. Now stated as a comment on the `recordMalicious` branch in `peer_registry.go` so a future change trips over it.
- **Effective strike count (ChiR2)**: with 1 pt/min decay and the 10-minute throttle, charges run 50 -> 90 -> 130, so a ban takes three reported offenses over 20+ minutes, not two. The real timeline is now pinned where the decay clock is reachable, in `blockchain`'s `TestCatchupMaliciousPointsOutrunWindowDecay`, which advances `LastDecay` between charges and asserts the third crosses the threshold; it also asserts the points-outrun-decay coupling (`points > window/decayInterval * decayAmount`). The p2p test was renamed to `TestRecordCatchupMalicious_ChargesAgainAfterWindow` and re-asserted to pin only that the throttle releases after the window (BanScore rises by another charge), not a strike count. A comment next to `catchupMaliciousChargeWindow` records the coupling.
- **Failed charge no longer consumes the window (ChiR3)**: `shouldChargeCatchupMalicious` stamps the throttle before the charge is attempted; if `AddBanScore` then errors (e.g. a transient blockchain-service gRPC failure), the stamp is now rolled back via `clearCatchupMaliciousCharge` so the next report retries instead of being suppressed for a full window. `applyBanScore` now returns its error for this one caller; statement-context callers keep their fire-and-forget shape. Covered by `TestRecordCatchupMalicious_FailedChargeReleasesThrottle`.
- **Two registry round-trips in `IsPeerMalicious`**: considered and kept. The ban table deliberately outlives peer removal, so `IsPeerBanned` can be true while `GetPeer` reports found=false; collapsing into a single `GetPeer` would miss exactly those peers. Documented in code. If the added RPC rate on the block-found / header-batch paths shows up in practice, a `banStatusCache`-style TTL cache is the follow-up.
- **Ban-score escalation reuses `applyBanScore`** (with its success log folded in) instead of duplicating the AddBanScore/onPeerBanned sequence, so the two escalation paths cannot drift.

### gitleaks allowlist

The `p2p_ban_threshold` struct tag's key name clears generic-api-key's entropy bar, so touching that settings line turned the PR scan red. The allowlist entry is now bounded: it only matches a settings tag whose `default:"..."` value is <=20 characters, so a real secret committed as a settings default is too long to satisfy it and stays reported by every rule. Verified empirically with the CI-pinned gitleaks 8.18.4: a full-history scan diff against the pre-change config shows zero findings un-suppressed and exactly the two settings-tag false positives suppressed; a synthetic 64-hex secret in a tag default is not maskable by the entry (it is also, notably, invisible to the ruleset even without any allowlist - a pre-existing baseline gap, not one this entry introduces); conf-style `*_private_key = <hex>` lines are still caught by `teranode-settings-private-key`.

On the review's alternatives, tested against 8.18.4 specifically: a rule-scoped `[rules.allowlist]` requires redefining upstream `generic-api-key` wholesale (a frozen ~1500-line copy including its stopword list - unmaintainable); `regexTarget = "match"` is shared by the whole `[allowlist]` table and empirically un-suppresses 57 historical findings the existing line-target entries legitimately cover; `paths` in 8.18.4's single allowlist ORs with regexes rather than restricting them. The bounded-default line regex is the tightest option the pinned version can express. Happy to lift this hunk into its own PR if preferred - it is separable.

### Auth boundary (ChiR1) - resolved by rebasing onto current main

The original review at `3ecaef3` flagged that `RecordCatchupMalicious` was an unauthenticated route to a ban. That is now moot: this branch is merged up to current `main`, which includes "Require authentication on mutating p2p data-plane RPCs" (#1532). That change moved all ten data-plane reporters - `RecordCatchupMalicious` included - into `authProtectedMethods()`, so the RPC now requires the gRPC API key, same as `AddBanScore`. The stale "deliberate boundary crossing" comment this PR had added to the old `adminProtectedMethods()` was dropped in the merge (the function was renamed and rewritten upstream). ChiR1's remaining sub-point - that the ID need not be a registered peer - is now bounded to authenticated in-deployment callers (block/subtree validation), the same trust level as every other reporter.

### Residual

- The eventual consequence is wider than the reputation pin it replaces: `onPeerBanned` also writes the peer's IP to the ban list for 24h, enforced by the shared HTTP 403 middleware. The malicious-report predicates (`shouldReportConsensusMalicious` / `isLocalCatchupFault`) exclude local faults (disk, aerospike timeouts, own shutdown), and the 3-strike timeline above bounds the blast radius of a residual false positive.
- A recorded malicious flag persists across restarts (registry persistence serializes `PeerInfo` wholesale) and is only cleared by the `ReconsiderBadPeers` sweep, which runs while the FSM is in a syncing-capable state, or by operator `ResetReputation`. Inherited behavior, now consequential; noted in the `IsPeerMalicious` doc comment.
- `onPeerBanned` -> `removePeer` deletes the registry `PeerInfo` (including the just-written `MaliciousCount`); the gate still holds because ban entries deliberately outlive peer removal. Pre-existing ban-path behavior, unchanged here.
- `services/blockvalidation` unit tests cannot link in this local environment (known GoBDK cgo issue); the package builds, `go vet` is clean, its `catchup` subpackage tests pass, and the blockvalidation changes are two log strings and a comment. CI runs the full package.

### Tests

Added in `services/p2p/handle_catchup_metrics_test.go`, against the real in-memory `CentralizedPeerRegistry`:

- `TestIsPeerMalicious_RecordedMaliciousPeer` - one report makes `IsPeerMalicious` true before any ban.
- `TestIsPeerMalicious_ReconsideredPeerRecovers` - `ReconsiderBadPeers` clears the flag.
- `TestRecordCatchupMalicious_AddsBanScore` - one report adds 50 ban points, no ban yet.
- `TestRecordCatchupMalicious_SameWindowReportsChargeOnce` - three same-cycle reports bump `MaliciousCount` to 3 but charge 50 points once.
- `TestRecordCatchupMalicious_ChargesAgainAfterWindow` - once the throttle window elapses, a distinct offense is charged again (BanScore rises); high threshold so it observes the charge without the ban+removal. Not a strike count - see ChiR2 above.
- `TestRecordCatchupMalicious_FailedChargeReleasesThrottle` - a failed `AddBanScore` rolls back the throttle stamp so the next report retries (ChiR3).
- `blockchain.TestCatchupMaliciousPointsOutrunWindowDecay` - pins the real 3-charge-over-20min timeline and the points-outrun-decay coupling, advancing the decay clock instead of sleeping.

Ran (green):

```
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/blockchain/...
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./settings/...
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/blockvalidation/catchup/...
gitleaks 8.18.4 (CI version): PR-range scan green; full-history diff vs pre-change config: 0 un-suppressed / 2 FPs suppressed
```


.gitleaks.toml
```diff
@@ -28,6 +28,11 @@ regex = '''(?i)[a-z0-9_]*private_keys?(?:\.[a-z0-9_.-]+)*\s*=\s*"?[0-9a-f]{64,}'
 description = "Localhost/docker dev-default DB credentials and env placeholders"
 regexTarget = "line"
 regexes = [
+  # Go settings struct tags (`key:"p2p_ban_threshold" desc:"..."`): the tag's
+  # key NAME can clear generic-api-key's entropy bar and get flagged as the
+  # secret whenever a PR touches the line. The tag never holds a secret value —
+  # defaults live in settings.conf, which is still scanned.
+  '''key:"[a-z0-9_]+" desc:"''',
```
liam: This is broader than the file's own stated policy, and I think it should be tightened or split out.

It lands in the **global** `[allowlist]` (a bare `[allowlist]` table header is root-scoped in TOML, not attached to the `[[rules]]` block above it) with `regexTarget = "line"`. So any line matching `key:"..." desc:"` is exempt from *every* rule, including this repo's own `teranode-settings-private-key` rule and the whole upstream default set.

Settings struct tags carry `default:"..."` on the same line. So a secret committed as a settings default would now be silently exempted, which is the exact scenario the comment at `.gitleaks.toml:40-43` says is covered:

> Known public P2P test fixtures, allowlisted by VALUE, not key name, so a real key committed under any `*_private_key` setting is still caught by the rule above.

This entry is a key-name / structural pattern, which is the thing that comment rules out.

Suggestions, any one of which works: move it to a rule-scoped `[rules.allowlist]` on the rule that is actually misfiring; or keep it global but add `paths` limited to `settings/.*_settings\\.go`; or switch to `regexTarget = "match"` so it exempts the matched token rather than the whole line.

Separately, this is a CI unblock rather than part of the fix, and it changes secret-scanning coverage for the whole repo. I would split it into its own PR so it gets reviewed as a security-config change on its own terms.

services/p2p/handle_catchup_metrics.go
```diff
@@ -468,20 +525,28 @@ func (s *Server) IsPeerMalicious(ctx context.Context, req *p2p_api.IsPeerMalicio
...
+				Reason:      "peer is banned",
+			}, nil
+		}
 
-	if banned {
-		return &p2p_api.IsPeerMaliciousResponse{
-			IsMalicious: true,
-			Reason:      "peer is banned",
-		}, nil
+		info, found, err := s.peerRegistry.GetPeer(ctx, req.PeerId)
```
liam, line 549: This adds a second registry round-trip to a hot path. `IsPeerMalicious` was one `IsPeerBanned` call; it is now `IsPeerBanned` + `GetPeer`.

The caller, `blockvalidation.isPeerMalicious` (`peer_metrics_helpers.go:302`), is uncached and runs per block-found Kafka message (`blockvalidation/Server.go:826`) and per catchup header batch (`catchup_get_block_headers.go:209`). In a distributed deployment `peerRegistry` is a gRPC client to the blockchain service, so this doubles the RPC rate on both paths.

A single `GetPeer` could plausibly serve both checks since `PeerInfo` carries `IsBanned`, but you would need to confirm it has the same expiry semantics as `IsBannedPeer` before collapsing them, which is probably why it is written this way. Either collapse it or note that the extra round-trip was considered and is acceptable.

services/p2p/handle_catchup_metrics.go
```diff
@@ -174,9 +174,61 @@ func (s *Server) RecordCatchupMalicious(ctx context.Context, req *p2p_api.Record
 		return &p2p_api.RecordCatchupMaliciousResponse{Ok: false}, errors.WrapGRPC(errors.NewServiceError("update peer metrics", err))
 	}
 
+	// The malicious counter alone never escalates to a ban, and
+	// ReconsiderBadPeers eventually resets it, so a repeat offender would be
+	// re-admitted indefinitely. Raise the ban score as well so repeated
+	// malicious reports cross the ban threshold. Best-effort: the malicious
+	// record above already landed and is what the catchup gates consume.
+	if s.shouldChargeCatchupMalicious(req.PeerId) {
+		score, banned, err := s.peerRegistry.AddBanScore(ctx, req.PeerId, ReasonCatchupMalicious, 0)
```
liam: This block reimplements `applyBanScore` (`Server.go:2593`): same `AddBanScore`, same warn-on-error, same `onPeerBanned` on a new ban transition. The only delta is the extra info-level success log. Worth reusing the helper, or folding the info log into it, so the two ban-score escalation paths cannot drift.

services/blockchain/peer_registry.go
```diff
@@ -159,6 +159,7 @@ func DefaultBanConfig() BanConfig {
 			"spam":               50,
 			"invalid_block":      10,
 			"catchup_failure":    30,
+			"catchup_malicious":  50,
```
liam, line 195: Not an objection, just making sure the blast radius is a deliberate choice: at 50 points against a threshold of 100, **two** reports past the 10-minute throttle produce a 24h ban, and `onPeerBanned` also writes the peer's **IP** to the ban list, which the shared HTTP 403 middleware enforces. That is materially wider than the reputation pin it replaces.

I did check the false-positive exposure and it looks bounded: the malicious reports at `blockvalidation/catchup.go:641` and `:1508` are gated behind `shouldReportConsensusMalicious` / `isLocalCatchupFault`, which do exclude local faults (disk, aerospike timeouts, our own shutdown). So this is a judgment call, not a defect.

If there is any residual doubt about those predicates, 25 points (four strikes) would give the same eventual outcome with a gentler first failure mode, and the throttle already prevents one offense from stacking.

liam (commented):
Reviewed the diff and traced the recovery path through the registry. Ran `go vet` and `-race` tests on the branch: `./services/p2p/...` and `./settings/...` both green.

The core change is sound, and I want to call out one thing I verified because the whole design leans on it and it is not stated in the PR body: `UpdateMetrics(recordMalicious=true)` sets `LastInteractionFailure` (`peer_registry.go:398-404`) and pins reputation to 5.0 (`:643`), so a flagged peer clears **both** `ReconsiderBadPeers` gates (`ReputationScore < 20` and a non-zero last failure). If either had been missing, `IsPeerMalicious` returning true on `MaliciousCount > 0` would have been a permanent exclusion with no recovery. It is worth writing that down in the PR description, since a future change to either field silently turns this into a peer-liveness bug.

Two things I would want changed, plus some smaller notes inline.

services/blockchain/peer_registry.go
liam: Correcting myself: **"two reports produce a ban" is wrong**, and my suggestion to drop to 25 points falls away with it. The automated review on the test file got this right.

`AddBanScore` applies decay before adding (`peer_registry.go:749-758`), and `DefaultBanConfig` is 1 pt/min against a threshold of 100 (`:150-155`). Since the throttle forces at least 10 minutes between charges, at least 10 points decay in between:

- t=0: 0 + 50 = 50
- t=10m: 50 - 10 = 40, + 50 = 90 (below threshold, no ban)
- t=20m: 90 - 10 = 80, + 50 = 130 -> ban

So it is effectively a 3-strike rule over 20+ minutes, not 2 strikes. That is a reasonable place to land and I withdraw the suggestion to lower the points.

The part of my comment that still stands is the blast radius: `onPeerBanned` writes the peer's **IP** to the ban list for 24h, which the shared HTTP 403 middleware enforces, so the eventual consequence is wider than the reputation pin it replaces. Given the malicious predicates are properly gated against local faults, I am comfortable with that at 3 strikes.

Worth noting that `TestRecordCatchupMalicious_RepeatedOffensesBanPeer` ages the throttle map but not the registry decay clock, so it bans on the 2nd charge in a way production timing cannot reproduce. The test is still valid as an escalation check, but a comment saying so would stop the next reader inferring a 2-strike rule the way I did.

liam (approved):
Re-reviewed against the current head (`2d6564e30`, merged with main). Every one of my previous findings is addressed, and I re-verified the load-bearing claims rather than trusting the description.

**Verified independently:**

- The recovery invariant is real. `calculateAndUpdateReputation` (`services/blockchain/peer_registry.go:648-651`) early-returns 5.0 **whenever** `MaliciousCount > 0`, not just at record time, so the `ReputationScore < 20` gate in `ReconsiderBadPeers` genuinely holds for the life of the flag. The new comment on the `recordMalicious` branch is accurate.
- `applyBanScore` reuse is done and the info log is folded in. One side effect worth knowing: that `Infof` now also fires on the three existing `ReasonProtocolViolation` call sites (`Server.go:1391/1409/1417`, `server_helpers.go:311`). Benign, just new log volume.
- The blast radius is narrower than I implied last round. `handleBlockFound` (`blockvalidation/Server.go:826`) only warns and still queues the block; it is `processBlockFound:1478` that refuses to fetch. So a flagged peer stops being a source, but its announcements still reach alternative sources.

`go vet` clean, and `-race -tags testtxmetacache ./services/p2p/... ./settings/...` green on the branch.

Two refinements, neither blocking.

**1. The gitleaks tightening does what it claims for the default value, but not for the rest of the line.**

The `default:"[^"]{0,20}"` bound genuinely stops a secret being the default. But `regexTarget = "line"` in the root `[allowlist]` exempts the *whole line* from *every* rule, and in these struct tags `longdesc:"..."` comes after `default:`. I ran the regex over `settings/`: 263 lines across 25 files now match and are globally exempt.

Nothing on them is secret-shaped today (I grepped the matched lines for 32+ char hex/base64 tokens; every hit is a camelCase identifier in prose). So the config is clean as it stands, and the residual is narrow: a secret pasted into a `longdesc` example on any settings line would be silently suppressed. That still supports splitting this hunk into its own PR so it gets read as a security-config change.

**2. The documented recovery path is not a fixed cooldown.**

`ReconsiderBadPeers` requires `cooldown * 3^ReputationResetCount` (`peer_registry.go:1021-1029`), where `cooldown` is itself `5m * max(1, backoffMultiplier/2)`. `ReputationResetCount` only ever increments (`:1034`), is persisted across restarts with the rest of `PeerInfo`, and is zeroed only by operator `ResetReputation` (`:1064`) or by peer removal.

So automatic clearing of a malicious flag is roughly 5 minutes for a fresh peer, and hours to days for one that has recovered a few times. The residual section says the flag "is only cleared by the `ReconsiderBadPeers` sweep" without that caveat. Worth one line, since it is the same class of thing the recovery-invariant comment exists to protect: a reader checking whether recovery is bounded will find the sweep and stop there.

Approving. Both notes are description/comment accuracy, not defects.

ctnguyen (changes requested):
Review at commit [3ecaef3](https://github.com/gokutheengineer/teranode/commit/3ecaef364d59a2ef1d42ea3323c6d8aea5356eb1)

This is the right shape for the problem: it takes both halves of the fix the linked issue proposes rather than just the cheap one, so `MaliciousCount` stops being write-only *and* the catchup paths finally reach a ban instead of logging "banning not yet implemented" forever (see [bitcoin-sv/teranode#4754](https://github.com/bitcoin-sv/teranode/issues/4754)). The throttle is well-reasoned and well-explained — collapsing the several report paths that fire for one bad block is exactly the trap a naive version of this would fall into, and the comment on `catchupMaliciousChargeWindow` says why. The test set is honest about what it covers: record-visible, reconsider-recovery, single-charge-per-window, and the repeat-offense ban. The residuals in the PR body are disclosed rather than buried, and the doc/log-string cleanup actually removes the stale "not yet implemented" claims instead of leaving them to rot. Three things below — one is the auth boundary you flagged, and I think it needs a decision before this merges rather than after.

## Blocking issues

### ChiR1 — an unauthenticated RPC now escalates to a 24h ban against any peer ID, and the comment recording that tradeoff understates it

**Problem:** The charge added at `services/p2p/handle_catchup_metrics.go:182-192` runs inside `RecordCatchupMalicious`, which is not in `adminProtectedMethods()` (`services/p2p/Server.go:1240-1250`) — that set protects `BanPeer`, `UnbanPeer`, `ClearBanned`, `AddBanScore`, `ResetReputation`, `ConnectPeer` and `DisconnectPeer`. The other two escalation sites in the package are the admin-protected `AddBanScore` handler (`services/p2p/Server.go:2581-2605`) and `applyBanScore`, reached from the gossip handlers (`services/p2p/Server.go:2609-2621`), so this is the first unauthenticated gRPC route to a ban.

The reach is wider than a peer the node is already talking to. `peer.Decode` at `services/p2p/handle_catchup_metrics.go:169` only checks that the id is well-formed; the registry write behind it does not require the peer to exist. `UpdateMetrics` returns silently for an unknown id (`services/blockchain/peer_registry.go:385-388`), and both client paths report success anyway (`services/blockchain/peer_registry_client.go:434-437`, `services/blockchain/peer_registry_grpc.go:29-41`), while `AddBanScore` creates a `banScores` entry for any string it is handed (`services/blockchain/peer_registry.go:720-724`). So ban score accumulates against peer IDs the node has never seen, and the early-return at `services/p2p/handle_catchup_metrics.go:173-175` does not bound the target set the way it looks like it does. The same gap makes the "the malicious record above already landed" premise in the comment at `services/p2p/handle_catchup_metrics.go:177-181` conditional rather than guaranteed.

The comment added at `services/p2p/Server.go:1232-1236` justifies the boundary as being the "same as they could already poison its reputation via the other reporting RPCs". That equivalence fails three ways:

- Reputation poisoning is auto-recovered; a ban is not. A malicious record pins `ReputationScore` to 5.0 (`services/blockchain/peer_registry.go:643-647`), and `ReconsiderBadPeers` clears both the score and `MaliciousCount` (`services/blockchain/peer_registry.go:999-1033`), driven by the sync coordinator on a 5-minute base cooldown (`services/p2p/sync_coordinator.go:1404-1431`). That sweep never touches `banScores` or `IsBanned`, so a ban stands for the full `BanDuration`, 24h by default.
- A ban does strictly more. Crossing the threshold runs `onPeerBanned` (`services/p2p/Server.go:2626-2674`): it seeds the ban-status cache, adds every one of the peer's IPs to the local ban list until `until`, and disconnects the peer. Reputation poisoning does none of that.
- The credential asymmetry runs the wrong way. With `grpc_admin_api_key` unset the server generates a random key precisely so the admin RPCs are unreachable (`services/p2p/Server.go:1047-1055`). On that default the operator cannot reach `AddBanScore` at all, while this path needs no credential.

**Why it matters:** Being careful about reach — nothing in the tree establishes that the p2p gRPC port is reachable from outside the deployment, and the comment states it must not be exposed publicly, so this is not a remote-exploitation claim. What it is: for anything that can already reach that port, the available effect grows from a recoverable reputation penalty to a 24h ban plus an IP-level ban-list entry plus a forced disconnect, none of which the reputation sweep undoes. And the residual disclosed in the PR body — "anyone who can reach the p2p gRPC port can therefore drive a peer toward a ban" — reads as a peer already in a catchup relationship with this node, whereas the actual capability covers any well-formed peer ID, registered or not. The gap between the disclosed tradeoff and the real one is what makes this a code decision rather than a wording one.

**Fix:** Decide the boundary, then record the decision accurately. Two implementations both satisfy this. Either keep the record unauthenticated — it is what the catchup gates consume — and suppress the address-level half for unauthenticated callers:

```go
  if banned {
      // Registry ban only: the address-level ban list and forced disconnect
      // stay behind the admin-authenticated AddBanScore path.
      s.logger.Warnf("[RecordCatchupMalicious] Peer %s reached the ban threshold via catchup reports", req.PeerId)
  }
```

Or authenticate the reporting RPCs as a group, as you suggest in the PR body, if the in-tree callers can carry credentials. Whichever is chosen, the comment at `services/p2p/Server.go:1232-1236` then needs to describe it — including the actual delta over reputation poisoning and the fact that the target need not be a registered peer — so the next person deciding whether some other RPC can safely escalate is reading something true.

## Non-blocking issues

### ChiR2 — the ban needs three charges over 20+ minutes, and the repeat-offense test asserts two

**Problem:** `catchup_malicious` is priced at 50 against a threshold of 100 with `DecayAmount: 1` per minute (`services/blockchain/peer_registry.go:150-165`), and `AddBanScore` applies decay from the wall clock before adding points (`services/blockchain/peer_registry.go:750-760`). Since the throttle spaces charges by at least 10 minutes (`services/p2p/handle_catchup_metrics.go:197-203`), at least 10 points decay between any two charges:

```text
  t=0min    0 -> +50            = 50
  t=10min   50 - 10 = 40  -> +50 = 90    (below the 100 threshold)
  t=20min   90 - 10 = 80  -> +50 = 130   -> banned
```

`TestRecordCatchupMalicious_RepeatedOffensesBanPeer` (`services/p2p/handle_catchup_metrics_test.go:283-307`) reaches a ban on the second charge because it back-dates `s.catchupMaliciousLastCharge[pid]` by `catchupMaliciousChargeWindow` without advancing the registry's `LastDecay`. Both are `time.Now()` in production and cannot diverge that way, so the test passes on a timeline the running node cannot produce.

**Why it matters:** Two separate costs. The test's name and assertion encode a two-strike rule the code does not have, so a reader calibrating operator expectations from it will be wrong about time-to-ban. More durably, the escalation only works at all because the points per charge (50) exceed the decay across one window (10 minutes x 1/min = 10) — and nothing pins that relationship. The window is a `const` in `services/p2p` while the points, decay and threshold live in `DefaultBanConfig` in `services/blockchain`, with the threshold additionally operator-tunable via `p2p_ban_threshold`. Widen the window to 50 minutes in a later change and the score never accumulates and no repeat offender is ever banned, with no test failing.

**Fix:** Rename and re-assert the p2p test so it pins what it actually exercises — that the charge is not throttled once the window has elapsed — by checking `BanScore` rather than the ban transition. Then pin the real invariant where the clock is reachable, with an in-package `services/blockchain` test that back-dates `LastDecay` between charges and asserts the third one crosses the threshold. A one-line comment next to `catchupMaliciousChargeWindow` noting that it must stay below `points / DecayAmount` would stop the coupling from being invisible.

### ChiR3 — a failed charge consumes the throttle window

**Problem:** `shouldChargeCatchupMalicious` (`services/p2p/handle_catchup_metrics.go:208-230`) records `now` and returns true before the caller has attempted anything; only afterwards does the caller run `AddBanScore`, and on error it logs a `Warnf` and moves on (`services/p2p/handle_catchup_metrics.go:183-185`). The timestamp stands regardless, so the next report for that peer is throttled away for the full window even though no points were ever applied.

**Why it matters:** `peerRegistry` is a gRPC client to the blockchain service in a distributed deployment, so a transient failure there is precisely the case this hits — and it is the case where the charge is dropped silently. Nothing retries it, nothing else fires within the window, and there is no counter or metric behind the `Warnf`, so a run of failed charges is invisible unless someone is reading logs at the time. (It also interacts with the accumulation in ChiR2, but the defect stands on its own.)

**Fix:** Record the charge only when it lands — split the decision from the recording:

```go
  if s.shouldChargeCatchupMalicious(req.PeerId) {
      score, banned, err := s.peerRegistry.AddBanScore(ctx, req.PeerId, ReasonCatchupMalicious, 0)
      if err != nil {
          s.clearCatchupMaliciousCharge(req.PeerId) // let the next report retry
          s.logger.Warnf(...)
      }
      ...
  }
```

The map pruning already in `shouldChargeCatchupMalicious` keeps this bounded either way.

## Recap

| ChiR1 | Unauthenticated RPC escalates to ban | YES | 75% |
| ChiR2 | Throttle and decay coupling untested | | 40% |
| ChiR3 | Failed charge consumes throttle window | | 25% |

ctnguyen (approved):
Review at commit [57ef493](https://github.com/gokutheengineer/teranode/commit/57ef493cf41ae7911452a95732002b93094d392d)

All three points from the last round are closed, and two of them are closed the way I'd hoped rather than the cheap way: the p2p repeat-offense test now asserts what it actually exercises instead of a strike count the production clock cannot produce, and the real 50/90/130 timeline is pinned on the blockchain side where the decay anchor is reachable. The throttle rollback in `RecordCatchupMalicious` is implemented exactly as sketched, with a test that wraps the real registry rather than a full mock — nice. The design notes in the PR body are unusually honest about what each test does and does not cover, which made this round quick. Four follow-ups below, all non-blocking; two of them are residuals of the fixes themselves rather than anything that was there before, and two are places where a comment claims more than the code delivers.

## Status of previous points

`ChiR1` is fixed by the merge up to current `main`, not by a change in this branch: #1532 moved `RecordCatchupMalicious` into `authProtectedMethods()` (`services/p2p/Server.go:1379`), and the stale comment the item objected to is gone with the function it lived on. `ChiR2` and `ChiR3` are judged against their own `Fix` fields, and both landed literally — rename plus `BanScore` assertion, a blockchain-side test that back-dates `LastDecay`, the coupling comment, and the stamp rollback. Their leftovers are new items below rather than status downgrades, because neither leftover is something those `Fix` fields asked for.

| ChiR1 | Unauthenticated RPC escalates to ban | YES | FIXED |
| ChiR2 | Throttle and decay coupling untested |  | FIXED |
| ChiR3 | Failed charge consumes throttle window |  | FIXED |

## Non-blocking issues

### ChiR4 — the throttle rollback trades a dropped charge for a possible double charge on one offense
**Problem:** `services/p2p/handle_catchup_metrics.go:186-190` clears the throttle stamp on *any* error out of `applyBanScore`, and `applyBanScore` returns whatever `s.peerRegistry.AddBanScore` returned (`services/p2p/Server.go:2839-2843`). In a distributed deployment that is a gRPC unary call (`services/blockchain/peer_registry_client.go:221-230`), so a client-side error — deadline exceeded, transport reset — does not establish that the server-side mutation did not happen. The score can have landed and only the response been lost.

That matters here because of the very thing the throttle exists for: one bad block is reported through three distinct catchup paths in the same cycle — the deferred lock-release report in `releaseCatchupLock` (`services/blockvalidation/catchup.go:641-642`), the direct report in `validateBlocksOnChannel` (`services/blockvalidation/catchup.go:1508-1511`), and the unvalidatable-peer handler (`services/blockvalidation/Server.go:2035-2041`) — all funnelling into `reportCatchupMalicious` at `services/blockvalidation/peer_metrics_helpers.go:280-305`. If report 1's response is lost after the points landed, the stamp is released and report 2 — seconds later, same offense — charges again. 50 + 50 = 100, which is the default `Threshold` (`services/blockchain/peer_registry.go:183-198`), so the peer is banned on a single offense.

**Why it matters:** The failure the rollback fixed under-punished, which is self-correcting; this one over-punishes, and a ban is not: `onPeerBanned` writes the peer's IPs to the local ban list for `BanDuration` and disconnects it. The code comment at `services/p2p/handle_catchup_metrics.go:199-200` still states the property flatly — "one offense scores once and only genuinely repeated offenses ban" — which is now conditional on the registry RPC never losing a response, and no test covers the lost-response case. (`ChiR7` is a second, unrelated way that same sentence overstates the guarantee.)

**Fix:** Keep the retry but don't make it immediate — back-date the stamp instead of deleting it, so the retry lands outside the same report burst but still well inside the window:

```go
  // Retry soon, but not inside the same report burst: the charge may have
  // landed server-side and only the response been lost.
  const catchupMaliciousRetryDelay = time.Minute

  if s.shouldChargeCatchupMalicious(req.PeerId) {
      if err := s.applyBanScore(req.PeerId, ReasonCatchupMalicious); err != nil {
          s.deferCatchupMaliciousRetry(req.PeerId, catchupMaliciousRetryDelay)
      }
  }
```

where `deferCatchupMaliciousRetry` sets the stamp to `now.Add(catchupMaliciousRetryDelay - catchupMaliciousChargeWindow)` under the existing mutex. If you'd rather keep the immediate clear, then say in the comment that the charge is at-least-once and that a lost response can double-charge one offense — the current comment reads as a guarantee.

### ChiR5 — the comment says the blockchain test pins the window coupling; it pins the other half
**Problem:** `services/p2p/handle_catchup_metrics.go:202-207` tells the reader that if `catchupMaliciousChargeWindow` "is ever widened past points/DecayAmount ... the score can never accumulate and no repeat offender is banned", and that "the blockchain side pins that invariant with `TestCatchupMaliciousPointsOutrunWindowDecay`". That test declares its own copy of the window — `const chargeWindow = 10 * time.Minute` at `services/blockchain/peer_registry_ban_test.go:133-138`, labelled "Mirror of services/p2p catchupMaliciousChargeWindow (not importable here without a cycle)". Widening the real constant therefore fails nothing: the mirror stays at 10 minutes. The test pins the blockchain-side half (50 points, 1/min, 100 threshold) genuinely, and `git grep -n catchupMaliciousChargeWindow -- services/` shows the only other use of the real constant is `handle_catchup_metrics_test.go:327`, which back-dates by the constant itself and so self-adjusts to any value.

The cycle argument only runs one way. `services/p2p` already imports `services/blockchain` in non-test code (`services/p2p/handle_catchup_metrics.go:10`, `services/p2p/Server.go:40`), and `handle_catchup_metrics_test.go:10` imports it too, so the assertion can read both the real constant and `blockchain.DefaultBanConfig()`.

**Why it matters:** This is the one coupling in the feature whose violation is silent — widen the window and the whole escalation quietly becomes a no-op, with `MaliciousCount` back to being the only consequence. The comment currently reassures a future editor that a test will catch it.

**Fix:** Move the coupling assertion to the p2p side, where both operands are real, and drop the mirrored constant from the blockchain test so it only claims the points/decay/threshold timeline:

```go
  // services/p2p/handle_catchup_metrics_test.go
  func TestCatchupMaliciousChargeOutrunsWindowDecay(t *testing.T) {
      cfg := blockchain.DefaultBanConfig()
      decayPerWindow := int32(catchupMaliciousChargeWindow/cfg.DecayInterval) * cfg.DecayAmount
      require.Greater(t, cfg.ReasonPoints[ReasonCatchupMalicious], decayPerWindow,
          "one charge must exceed the decay across a throttle window, else no offender is ever banned")
  }
```

Then point the comment at that test instead.

### ChiR6 — the doc comment on IsPeerMalicious describes recovery as a sweep, with no hint that its cooldown grows without bound
*Same as [@liam's point 2 (01-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1664#pullrequestreview-5076470048).*
**Problem:** The new doc comment at `services/p2p/handle_catchup_metrics.go:518-524` says the malicious record "is cleared again when the sync coordinator's reputation-recovery sweep runs `ReconsiderBadPeers` (only while the FSM is in a syncing-capable state), or by an operator `ResetReputation`". `ReconsiderBadPeers` also requires `now.Sub(info.LastReputationReset) >= cooldown * 3^ReputationResetCount` (`services/blockchain/peer_registry.go:1062-1070`), and `ReputationResetCount` only ever increments (`:1075`), is serialized with the rest of `PeerInfo`, and is zeroed only by `ResetReputation` (`:1105`) or peer removal. The base `cooldown` is itself `5m * max(1, backoffMultiplier/2)` (`services/p2p/sync_coordinator.go:1414-1423`). So automatic clearing is ~5 minutes for a fresh peer and hours to days for one that has recovered a handful of times — 5m × 3⁵ is just over 20 hours at five prior resets.

Where I'd extend liam's point: he framed it as description accuracy, and on the previous head it was, because the record only pinned reputation. On this head `IsPeerMalicious` returns true on `MaliciousCount > 0` (`services/p2p/handle_catchup_metrics.go:553-558`), so the same unbounded timer now decides how long a peer is excluded by every catchup gate — including `tryAlternativePeersForCatchup`, where per the linked issue's verification notes the `isPeerMalicious` check is the only malicious exclusion in the path (see [bitcoin-sv/teranode#4754](https://github.com/bitcoin-sv/teranode/issues/4754)). The PR body's *Fix* bullet does say "(exponentially growing) cooldown"; the shipped comment and the body's *Residual* section do not.

**Why it matters:** The comment is where the next reader checks whether the new exclusion is bounded, and as written the answer looks like "yes, a few minutes". For a peer with a history of resets it can be a day or more, which is longer than the 24h ban this feature escalates to — the softer state outlasting the harder one is worth one sentence.

**Fix:** One line on the comment, and the same caveat in the PR body's *Residual* list:

```go
  // state), or by an operator ResetReputation. Note the sweep is not a fixed
  // cooldown: eligibility needs cooldown * 3^ReputationResetCount, and that
  // count only ever grows, so for a peer that has recovered before the record
  // can stand for hours or days.
```

### ChiR7 — the comment states a property of the default threshold as a property of the mechanism
**Problem:** `services/p2p/handle_catchup_metrics.go:199-200` says the window means "one offense scores once and only genuinely repeated offenses ban", unconditionally. It holds only at the default threshold. Production builds `BanConfig` with `Threshold: int32(tSettings.P2P.BanThreshold)` while leaving `ReasonPoints` at the defaults (`services/blockchain/Server.go:198-199`), and `AddBanScore` bans on `entry.Score >= r.banConfig.Threshold` (`services/blockchain/peer_registry.go:832`). At `p2p_ban_threshold = 50` — a value the setting's own trade-off table advertises as "Lower (50) | Strict enforcement, faster bad peer isolation" (`settings/p2p_settings.go:33`) — the first `catchup_malicious` report is an immediate 24h ban.

To be clear about scope: this is not an objection to the 50-point price, to the three-strike arithmetic at the default, or to the IP-ban blast radius. It is also worth saying plainly that one-shot-at-threshold-50 is not a class this PR introduces — `spam` is already priced at 50 in the same table. The new part is only that this comment asserts the opposite.

**Why it matters:** An operator who lowers the threshold on the strength of that trade-off table gets single-report bans from a mechanism whose in-tree documentation promises the reverse, and the comment is the natural place someone checks before tuning.

**Fix:** Qualify the sentence rather than change the behaviour:

```go
  // charge so one offense scores once and only genuinely repeated offenses ban
  // at the default threshold (100). Operators who lower p2p_ban_threshold to
  // the points value or below get a ban on the first report, as they already
  // do for spam, which is priced the same.
```

## Recap

| ChiR4 | Failed charge may double charge |  | 40% |
| ChiR5 | Window coupling not pinned |  | 30% |
| ChiR6 | Recovery cooldown grows unbounded |  | 25% |
| ChiR7 | Comment ignores tunable threshold |  | 20% |

PR #1665, opened 2026-08-28
p2p: latch sync-slot verdicts to the FSM completion edge and fix the backoff multiplier ratchet

gokutheengineer:
**Closes bitcoin-sv/teranode#4753.**

### Problem
Both findings confirmed on current main (line numbers drifted slightly from the issue).

1. `handleFSMTransition` had no FSM-state latch. Block validation runs the whole catchup header phase with the FSM still in RUNNING (CATCHINGBLOCKS only starts at the block-body phase), so once the sync peer had validated chainwork credited from a previous run, every 2s RUNNING tick re-judged the in-flight sync: "considered failed" if the peer was still ahead, or a spurious "completed successfully" (with a backoff reset) if it was level. Either way the slot was cleared and re-selected, burning each peer's 1-minute sync-attempt cooldown within seconds, leaving rejected duplicate triggers' block hashes suppressed in block validation's dedup cache, resetting the no-progress clock on every re-claim, and flapping `syncPeerID` in gossiped node_status.

2. `resetBackoff`'s multiplier reset was dead code on the monitor path: it was gated on `allPeersAttempted`, but `checkAndClearExpiredBackoff` always clears that flag (and doubles the multiplier) when the window expires, before any selection - and both production `resetBackoff` call sites run after that. The multiplier could only ratchet to the 32x cap, making every future exhaustion backoff maximal and inflating the reputation-recovery cooldown from 5 to 80 minutes until restart.

### Fix
1. Sync-slot completion verdicts are keyed to signals that can be attributed to a peer, never to FSM observation alone:
   - **The authoritative settler is `SyncCoordinator.HandleCatchupSuccess`**, wired from the p2p `RecordCatchupSuccess` gRPC handler. The report names the peer and carries the catchup duration, so a verdict cannot be misattributed after a slot handover; a stale report whose catchup started before the current claim is ignored. Still-ahead settles as "completed but peer is still ahead; re-selecting" (a designed outcome - the header fetch truncates at its accumulated cap on a node far behind), level-or-behind settles as success and de-escalates backoff. It is gated on a new `catchup_completed` proto field (older blockvalidation versions used this RPC to credit individual header batches, which must not settle the sync - rolling-upgrade safe in both directions), guarded against running during shutdown, and run async so settling (which may select and health-check a new peer) never blocks the RPC. De-escalation of backoff happens only here.
   - **The FSM CATCHINGBLOCKS -> RUNNING edge never issues a verdict.** An edge says a catchup finished, not whose: the header phase runs entirely in RUNNING, so a slot that changed hands during another peer's header phase would be judged by that peer's outcome, and the coordinator cannot attribute an excursion (catchups are also enqueued by gossip, not only by the coordinator's trigger). The edge instead arms a bounded window (`pendingCompletionPeer`, expiring after 30s) that holds the fallback release off while the authoritative report is in flight; the window is disarmed by the report, by a slot change, by leaving RUNNING, or by expiry.
   - **A fallback release** covers completions that emit no signal at all - a trigger rejected by block validation's single-flight guard, or catchup's already-synced short-circuits (`doCatchup` returns nil without reporting success when the header fetch yields no headers, or when filtering leaves none and the target block is already known locally). At steady-state block spacing this release is an ordinary outcome, not an escape hatch. A holder that is level or behind by validated work AND has gone the preemption guard (half the no-progress timeout) *without delivering a validated block* is quietly released: keyed on progress age, not claim age, so an actively delivering level peer keeps the slot (matching `evaluateSyncPeer`'s keep-while-caught-up policy); no verdict is fabricated - no `resetBackoff`, no probe-budget refill, no sync attempt recorded against the peer. Level is tested with `chainWorkGreater` directly, so a peer ahead by raw chainwork but demoted by an active full-storage penalty (header work it failed to back with a block body) is never released as finished - it is left to the no-progress deadline, which also benches it.

   Per-tick housekeeping (registry existence, block-progress recording, the no-progress deadline) runs unlatched on every RUNNING tick, so the existing eviction paths are untouched.

2. `resetBackoff` now resets the multiplier unconditionally (the issue's minimum fix), so a genuinely settled successful sync de-escalates it; `considerReputationRecovery`'s recovery path now also actually resets. The issue's optional "reset on local chainwork advance" was deliberately NOT implemented: local tip advances come from gossip by any peer, and resetting on every advance would pin the multiplier at 1 during IBD and let each ~2s selection-failure cycle re-enter backoff - whose entry clears every peer's sync-attempt cooldown - reintroducing the cooldown-burning churn through the other door. A comment in `resetProbeBudgetIfLocalChainWorkAdvanced` documents this.

### Behavior change worth calling out
On main, `handleFSMTransition` returned true (short-circuiting `checkFSMState`) on most slot-held RUNNING ticks, which incidentally suppressed `considerReputationRecovery` and `handleRunningState` in steady state. With the latch those ticks fall through, so reputation recovery would have run (and issued a `ReconsiderBadPeers` RPC) every 2s. It is now explicitly throttled to a 30s sampling interval (`considerReputationRecoveryThrottled`); recovery cooldowns are minutes, so the sampling loses nothing, and the RPC load is bounded regardless of FSM state or slot occupancy. `handleRunningState` remains unthrottled (it early-returns when a slot is held).

### Residual
Reviewed by an independent adversarial pass plus two collaborator review rounds; known, deliberate gaps:
- A slot holder still *ahead* by validated work whose catchup died silently (e.g. rejected with `ErrCatchupInProgress`) is bounded by the existing eviction paths: opportunistic preemption (needs a strictly-higher-work rival and 2.5m without delivery) or the 5-minute no-progress deadline, plus rotation via `HandleCatchupFailure` / periodic evaluation. The instant recycling was the bug being fixed; those are the designed paths.
- A peer served by a blockvalidation that does not set `catchup_completed` (rolling upgrade) never settles with a verdict. Worst-case waits are explicit: a level holder is released by the fallback once its progress age passes 2.5m, plus up to 30s when an edge window was armed; a still-ahead holder waits for preemption or the 5m deadline. Backoff is simply not de-escalated by those releases.
- The latch has sampling gaps in when it can observe edges, all bounded and none able to misattribute (edges carry no verdicts): the normal 2s tick; `checkFSMState`'s early return during an active backoff window (up to 64s at the 32x cap, and a slot can be claimed inside it via `TriggerSync`/failure/disconnect paths); and `monitorFSM`'s caught-up slow tick (15s, or no `checkFSMState` call at all), which can swallow a whole gossip-driven excursion - that window is only open with an empty slot, where no verdict is owed. A lost edge only means the fallback release is not held off for that completion.
- The `HandleCatchupSuccess` goroutine is guarded by `stopCh` but not wg-tracked (matching the existing best-effort stance on late callbacks documented in `Stop`), and is spawned per RPC without dedup: a burst of completion reports queues goroutines on `decisionMu` rather than coalescing. Both are bounded by the real call rate of one report per completed catchup; the async spawn deliberately avoids both blocking the RPC and serializing it against the monitor tick.

### Tests
Added regression tests in `services/p2p/sync_coordinator_test.go` and `handle_catchup_metrics_test.go`:
- steady RUNNING ticks no longer recycle a slot whose holder has higher validated work (the 2s churn),
- a completion edge never judges the slot holder (the handover-misattribution shape: claim predating the excursion),
- an armed edge window holds the fallback release off, expires after its bound without touching the slot, and a late credit produces no edge-derived verdict,
- a slot claimed mid-catchup survives the excursion end untouched,
- the fallback releases a quiet level peer after the grace (asserting backoff, probe budget and the peer's sync-attempt record are untouched), keeps an actively delivering level peer regardless of claim age, never releases an ahead peer, and never releases a penalty-demoted peer that is ahead by raw chainwork,
- `HandleCatchupSuccess` settles without an FSM edge, ignores non-sync peers and stale reports (catchup started before the claim), disarms the edge window even when its own verdict defers, and a trailing edge cannot evict the freshly re-claimed slot,
- the gRPC handler settles the slot only when `catchup_completed` is set (legacy header-batch shape leaves it alone),
- a -race test drives the FSM path and the success signal concurrently over the latch fields,
- `resetBackoff` de-escalates a ratcheted multiplier under the production ordering (flag already cleared),
- a chainwork advance refreshes the probe budget but does not de-escalate the multiplier,
- the reputation-recovery throttle samples at its interval.

Ran (all green):
```
go vet ./services/p2p/...
go build ./services/blockvalidation/... ./services/legacy/...
```


services/p2p/sync_coordinator.go
```diff
@@ -826,59 +862,205 @@ func (sc *SyncCoordinator) checkFSMState() {
...
+		// a finished sync (and the fast monitor tick) until the no-progress
+		// deadline. Once the holder is level or behind by validated work the
+		// sync has achieved its goal, so settle it as success after a grace
+		// period. The grace (the preemption guard, half the no-progress timeout)
+		// covers an in-flight header phase that has not yet re-credited a peer
+		// whose stale validated work sits at or below local, so an active sync
+		// is not settled prematurely.
+		if localWorkOK && peerHasValidatedWork(peerInfo) &&
+			!peerAheadByValidatedWork(peerInfo, localChainWork) &&
+			now.Sub(syncStart) > sc.preemptionProgressGuard() {
```
liam: This is my main concern with the PR.

The condition uses `now.Sub(syncStart)`, time since the slot was **claimed**, but reuses `preemptionProgressGuard()`, whose documented meaning is time gone **without delivering a validated block** (`sync_coordinator.go:210-215`, and the `p2p_sync_peer_no_progress_timeout` longdesc, which says "the incumbent must go this long without delivering a validated block"). `lastSyncProgressTime` is refreshed two lines up by `recordSyncPeerBlockProgress` on line 932 and then never consulted.

The comment above says the grace exists so "an active sync is not settled prematurely", but claim age does not measure that. I probed it on this branch: peer level by validated work, `lastSyncProgressTime = time.Now()` (progress as fresh as it can be), claim age 5 minutes:

```
handled=true peer="current" backoffMultiplier 16->1 probeBudget 0->2 syncStart moved=true LastSyncAttempt=true
```

So it settles a peer that is actively progressing, declares "Sync completed successfully", and immediately re-claims the same peer with a fresh `syncStartTime`. That cycle then repeats every `preemptionProgressGuard()` (2.5 min by default) indefinitely. Two consequences:

1. `syncPeerID` still flaps in gossiped `node_status`, at 2.5 min instead of 2s. That is a smaller version of a symptom this PR sets out to remove, not its removal.
2. `settleSyncCompletion` calls `resetBackoff`, which refills `unprovenProbeBudgetRemaining`. So `p2p_max_unproven_sync_probes_per_backoff_window` (default 3) gets topped up on a timer by a completion that was never observed.

Suggested fix: gate on progress age instead of claim age. `syncPeerNoProgressTimedOut` already computes `progressAge` on line 933, so it is nearly free:

```go
if localWorkOK && peerHasValidatedWork(peerInfo) &&
    !peerAheadByValidatedWork(peerInfo, localChainWork) &&
    progressAge > sc.preemptionProgressGuard() {
```

And separately, consider not calling `resetBackoff` from this inferred-success path at all, reserving de-escalation for the two observed signals (the FSM edge and `HandleCatchupSuccess`). The fallback is a guess about a completion; the other two are evidence of one.

services/p2p/sync_coordinator.go
```diff
@@ -826,59 +862,205 @@ func (sc *SyncCoordinator) checkFSMState() {
...
+			return sc.settleSyncCompletion(currentPeer, peerInfo, localChainWork)
```
liam: Related to the comment above: this fallback also contradicts the policy the periodic evaluation already implements for the identical condition.

`evaluateSyncPeer` at `sync_coordinator.go:1160-1170` handles `!peerAheadByValidatedWork` ("Caught up to sync peer %s by validated work") and deliberately **keeps the slot**, rotating only if `selectNewSyncPeer()` returns a strictly better peer:

```go
if !peerAheadByValidatedWork(peerInfo, localChainWork) {
    sc.logger.Infof("[SyncCoordinator] Caught up to sync peer %s by validated work", currentPeer)
    // Don't clear peer yet, but look for better peer.
    if betterPeer := sc.selectNewSyncPeer(); betterPeer != currentPeer && betterPeer != "" {
```

This new path clears and re-triggers unconditionally on the same condition. So two code paths now disagree about what "caught up" means for the slot, and the 2s one wins by frequency.

Making the fallback do the same rotate-only-if-better check would resolve this and most of the churn in the comment above, and would keep the two paths from drifting further.

services/p2p/sync_coordinator.go
```diff
@@ -826,59 +862,205 @@ func (sc *SyncCoordinator) checkFSMState() {
...
 		}
+		return false
```
liam: Behaviour change worth adding to the PR description, since it is not obvious from the diff.

On main, `handleFSMTransition` returned `true` whenever a sync peer was held and validated work was available, which short-circuited `checkFSMState` before it reached `considerReputationRecovery()` and `handleRunningState()` (`:821-823`). With the latch it returns `false` in that case, so both now run on **every** 2s RUNNING tick in steady state.

`handleRunningState` is cheap (`selectAndActivateNewPeer` early-returns when `oldPeer != ""`). But `considerReputationRecovery` issues a `ReconsiderBadPeers` RPC to the blockchain service every 2s where it previously ran rarely, which is both a load change and a real change in how fast bad peers recover their reputation.

Probably fine, possibly even desirable, but it should be an intentional call rather than a side effect of the latch.

services/p2p/sync_coordinator.go
```diff
@@ -826,59 +862,205 @@ func (sc *SyncCoordinator) checkFSMState() {
...
-			if err != nil {
-				sc.warnfUnlessStopping("[SyncCoordinator] GetPeer %s failed: %v", currentPeer, err)
-				return false
-			}
+	completionEdge := sc.lastObservedFSMStateSet &&
+		sc.lastObservedFSMState == blockchain_api.FSMStateType_CATCHINGBLOCKS &&
+		*currentState == blockchain_api.FSMStateType_RUNNING
+	if *currentState == blockchain_api.FSMStateType_CATCHINGBLOCKS &&
+		(!sc.lastObservedFSMStateSet || sc.lastObservedFSMState != blockchain_api.FSMStateType_CATCHINGBLOCKS) {
+		sc.fsmCatchingSince = time.Now()
```
liam: Minor, and not among the tradeoffs listed in the PR body: `fsmCatchingSince` records when CATCHINGBLOCKS was *observed*, not when it was entered, so the causality guard has a sampling blind window of up to one tick (~2s). A slot claimed between the real transition and the tick that observes it will have `syncStart < fsmCatchingSince`, pass the guard, and be judged by a catchup it did not start. Low impact given the other guards, just worth a line in the residual section alongside the excursion-keying note.

services/p2p/handle_catchup_metrics.go
```diff
@@ -57,14 +57,26 @@ func (s *Server) RecordCatchupSuccess(ctx context.Context, req *p2p_api.RecordCa
...
 	}
 
-	if _, err := peer.Decode(req.PeerId); err != nil {
+	decodedPeer, err := peer.Decode(req.PeerId)
+	if err != nil {
 		return &p2p_api.RecordCatchupSuccessResponse{Ok: false}, errors.WrapGRPC(errors.NewProcessingError(errInvalidPeerIDFormat, err))
 	}
 
-	if err := s.peerRegistry.RecordCatchupSuccess(ctx, req.PeerId, req.DurationMs); err != nil {
+	if err := s.peerRegistry.RecordCatchupSuccess(ctx, decodedPeer.String(), req.DurationMs); err != nil {
```
liam: This is an unrelated behaviour change bundled into the PR. Switching from `req.PeerId` to `decodedPeer.String()` normalizes the registry key: `peer.Decode` accepts several encodings (CIDv1 base32 among them) and `String()` emits base58, so a caller sending a non-canonical ID now writes to a different key than before.

It is probably a fix rather than a regression, but it is orthogonal to the sync-latch work, and the sibling handlers in this same file (`RecordCatchupAttempt`, `RecordCatchupFailure`, `RecordCatchupMalicious`, `UpdateCatchupError`) still pass `req.PeerId` unnormalized. So this leaves the file internally inconsistent.

I would either normalize all of them in a separate PR, or drop it here and pass `req.PeerId` (the new `HandleCatchupSuccess` call on line 77 genuinely does want the canonical form, so keep `decodedPeer.String()` there).

services/p2p/handle_catchup_metrics.go
```diff
@@ -57,14 +57,26 @@ func (s *Server) RecordCatchupSuccess(ctx context.Context, req *p2p_api.RecordCa
...
 
+	// Let the sync coordinator settle the completed sync (mirrors the catchup
+	// FSM completion edge; this is the authoritative signal for catchups whose
+	// FSM excursion the monitor tick never observes). Only for reports flagged
+	// as whole completed catchups: older blockvalidation versions used this RPC
+	// to credit individual header batches, which must not settle the sync. Run
+	// asynchronously: settling may select and health-check a new peer, which
+	// must not block this RPC.
+	if req.CatchupCompleted && s.syncCoordinator != nil {
+		go s.syncCoordinator.HandleCatchupSuccess(decodedPeer.String(), time.Duration(req.DurationMs)*time.Millisecond)
```
liam, line 80: Confirming the shutdown caveat you already flagged, and adding one detail: this is spawned per RPC with no dedup, and immediately blocks on `decisionMu`, which `checkFSMState` holds across blockchain RPCs bounded at 5s. So a burst of completion reports queues goroutines on the mutex rather than coalescing.

Both that and the untracked-goroutine-past-`stopCh` case are bounded and low given the real call rate (one per completed catchup), so I am not asking for a change. Worth one sentence in the residual noting the queueing, since the async choice is doing double duty here: not blocking the RPC, and not serializing against the monitor tick.

liam (commented):
Reviewed the diff, traced the surrounding coordinator code, ran `go vet` and `-race` tests on the branch (`./services/p2p/...` green), and wrote a throwaway probe test to check one behaviour empirically.

**Both underlying bugs are real and I confirmed them independently.** The missing edge detection is plain in the pre-diff code. The ratchet is real too: `resetBackoff` has exactly two production call sites (`sync_coordinator.go:1000`, `:1617`), and both are reachable only from `checkFSMState` *after* `checkAndClearExpiredBackoff` has already set `allPeersAttempted = false` (`:1318`), so the multiplier reset was unreachable. The unconditional reset is the right fix.

The `catchup_completed` proto gate is also the right shape, and safe in both upgrade directions. I confirmed `RecordCatchupSuccess` has exactly one caller (`blockvalidation/catchup.go:303`, whole-catchup completion), so setting the flag unconditionally in `Client.go` is correct today. The stale-report guard direction is right as well: in normal flow `syncStart < catchupStart` so it passes, and a mid-catchup re-claim correctly moves `syncStart` past `catchupStart` and is rejected.

My substantive concern is the fallback settle, detailed inline with the probe output. Short version: it keys off claim age rather than progress age, so it settles peers that are actively delivering blocks, and it manufactures a "sync completed successfully" verdict on a 2.5-minute cycle. I would hold on that part.

liam (approved):
Re-reviewed against the current head (`74b3ea6ec`, merged with main). All six of my previous findings are addressed, and the fixes are the right ones rather than comments papering over the behaviour.

**Verified independently:**

- The fallback now keys on `progressAge` (`sync_coordinator.go:966`), calls no `resetBackoff`, refills no probe budget, and records no sync attempt. `TestSyncCoordinator_HandleFSMTransition_LevelQuietPeerFallbackReleasesAfterGrace` asserts all three, which is exactly the guard I wanted.
- `considerReputationRecoveryThrottled` documents that it needs `decisionMu`; `checkFSMState:831` does hold it.
- `decodedPeer.String()` is scoped to the coordinator call only, with the registry call left on `req.PeerId` matching the siblings in that file.
- The stale-report guard is on the right quantity: I re-confirmed the single caller (`blockvalidation/catchup.go:303`) passes `time.Since(catchupCtx.startTime)`, the whole-catchup duration, so `syncStart.After(now - catchupDuration)` really is comparing the claim against the catchup start. Rolling-upgrade safe in both directions.

`-race -tags testtxmetacache ./services/p2p/...` green on the branch.

Three notes, none blocking.

**1. At mainnet block spacing the fallback is the routine steady-state exit, not a rare escape hatch.**

`preemptionProgressGuard()` is 2.5 minutes by default (5m timeout / 2); interblock time is about 10 minutes. So on a caught-up node a perfectly healthy level sync peer crosses the guard between blocks and gets released. What follows is: `triggerSyncLocked` -> `selectAndActivateNewPeer` -> `filterEligiblePeersWithTip` finds nothing ahead (`:1196-1206`) -> `enterBackoffMode()` -> `registryClearAllSyncAttempts()` plus a WARN, on a roughly 2.5-minute cadence.

This is still a large improvement on main, which recycled the slot every 2s, and it self-heals on the next block: the peer goes ahead, the claim is probe-budget-exempt, catchup runs, `HandleCatchupSuccess` fires and `resetBackoff` lands. So I am not asking for a behaviour change.

But the comment at `:950-964` and the residual both read as though the fallback only fires for signal-less completions, and `TestSyncCoordinator_HandleFSMTransition_FallbackKeepsActivelyDeliveringLevelPeer` exercises it with `lastSyncProgressTime = time.Now()`, which is fresher than production ever gets. Either raise the guard on this path, additionally require the local tip to be stale, or just say in the comment that releasing a quiet level peer between blocks is expected rather than exceptional.

**2. A second sampling blind window, alongside the `fsmCatchingSince` one already in the residual.**

`checkFSMState` returns at `:838` while `checkAndClearExpiredBackoff()` is true, so `handleFSMTransition` never runs and `lastObservedFSMState` goes stale for the whole backoff window. At the 32x cap that is 64 seconds in which a complete CATCHINGBLOCKS excursion can begin and end unobserved, losing the edge entirely.

`HandleCatchupSuccess` covers it, which is precisely why wiring it was the right call, but it belongs in the residual next to the existing observed-vs-entered note.

**3. `p2p_api_grpc.pb.go` carries only a toolchain comment bump** (protoc-gen-go-grpc 1.6.1 -> 1.6.2, protoc 6.33.2 -> 7.35.1). No CI job regenerates protos, so it is harmless, but it will churn back for anyone still on the older toolchain. Fine to leave, or revert those two lines.

Approving. Note 1 is the only one describing real behaviour rather than documentation, and it describes an improvement over what is on main today; I would just like the comment to match what actually happens.

ctnguyen (changes requested):
Review at commit [aeff358](https://github.com/gokutheengineer/teranode/commit/aeff358cfa33bce998086cf6e1017241b0b43de3)

Both defects here are diagnosed precisely, and the diagnoses are the hard part. The missing FSM latch is exactly right — block validation runs the whole header phase with the FSM still in `RUNNING`, so re-judging the in-flight sync on every 2s tick was always going to produce a verdict about a catchup that had not happened yet, and latching to a real `CATCHINGBLOCKS -> RUNNING` edge is the correct shape of fix. The `allPeersAttempted` gate on the multiplier reset being dead code on both production call sites is a genuinely non-obvious find, and the new comment on `resetBackoff` explains *why* it was dead rather than just asserting it. The causality guard and the deferred evaluation are both thoughtful: keying `fsmCatchingSince` on the excursion rather than the individual catchup deliberately errs toward suppressing verdicts, which is the right direction for this class of bug, and the `pendingCompletionPeer` deferral is invalidated on every state change that could invalidate it. The `catchup_completed` proto gate is properly rolling-upgrade safe in both directions, and declining the issue's optional third change with the reasoning written into `resetProbeBudgetIfLocalChainWorkAdvanced` is the right call — that option really would have reopened the churn through the backoff-entry door. The test list is unusually thorough for state-machine work, including a `-race` test that drives the FSM path and the success signal concurrently over the new latch fields.

The one thing I'd want changed before merge is the fallback settle added in the second commit. It is the only new mechanism that can produce a verdict with no completion signal behind it, and it reaches for `resetBackoff` when it does.

## Blocking issues

### ChiR1 — Fallback settle can declare success on a sync still in its header phase, and de-escalates backoff when it does

**Problem:** The fallback settle (`services/p2p/sync_coordinator.go:938-955`) settles the slot as a success on two conditions, neither of which is evidence that a catchup finished: the holder is "not ahead by validated work", and it has held the slot longer than `preemptionProgressGuard()`. Trace the clock — the guard is `syncPeerNoProgressLimit() / syncPeerPreemptionGuardDivisor` (`:242-244`), so 2.5m on the 5m default (`:147`, `:169`), and it is measured against `syncStartTime`, i.e. when the slot was *claimed*. Trace the credit — a catchup emits its validated-header credit exactly once, at `services/blockvalidation/catchup.go:284`, after the whole header chain has been fetched, common-ancestor-resolved, fork-depth-checked, continuity-verified and checkpoint-verified; only afterwards does it leave `RUNNING` for `CATCHINGBLOCKS`, at `catchup.go:1178`. So for the entire window between the claim and that single credit point the FSM is `RUNNING`, and a holder whose *previous* credit sits at or below local reads as "not ahead" — which is precisely the stale-credit state the original bug turned on. That window's design ceiling is not small: header fetch runs up to `CatchupMaxAccumulatedHeaders / maxBlockHeadersPerRequest` = 100000/10000 = 10 iterations (`settings/settings.go:381`, `services/blockvalidation/catchup.go:30`), each bounded only by `blockvalidation_catchup_iteration_timeout`, default 30s (`settings/settings.go:379`), with retries on top, and the Kafka trigger plus the catchup queue sit in front of it before `startTime` is taken. **I have traced the mechanism and that ceiling; I have not established how often a real header phase actually exceeds 2.5 minutes, so treat the rate as unquantified — the mechanism is what I am asserting.**

Two further ways the condition is not evidence of success, independent of the clock:

- `!peerAheadByValidatedWork` is also true for a peer that *is* ahead by chainwork but sits inside a block-incomplete / full-storage penalty window (`:250-254`, `:262-264`) — which is exactly a peer that served header work it could not back with a block body. Such a peer can hold the slot via the advertised-probe tier, and the fallback then calls it a completed sync.
- The condition never consults delivery at all. A holder that has delivered zero blocks under this claim satisfies it as readily as one that delivered the whole chain; neither `lastSyncProgressTime` nor `peerInfo.BlocksReceived` is in the test. Note also that `preemptionProgressGuard()` is defined, documented and used everywhere else as a *progress-age* threshold (`:1370`, and its own doc comment at `:237-241`); this is the only place it is applied to claim age.

**Why it matters:** The settle path calls `sc.resetBackoff()` (`:1000`), which unconditionally drops `backoffMultiplier` to 1 and also refills `unprovenProbeBudgetRemaining` (`:1525`). So a sync that has not completed produces a real de-escalation of the exhaustion escalation Fix 2 exists to make work, and refills the very budget that bounds how many unproven peers may be probed per window — for a holder that was never required to deliver a block. The slot is then cleared and re-triggered while the original catchup is still running, so the completion signal that eventually arrives is dropped at `:1028` (the peer is no longer the holder), and `syncPeerID` flaps in gossiped `node_status`. Because the retrigger can re-pick the same peer once its 1-minute sync-attempt cooldown has passed — and 2.5 minutes have, by construction — the cycle *can* repeat on a 2.5-minute period with a backoff reset each time, kept viable by the probe-budget refill. That is the original bug's shape on a longer clock, which is the specific failure this PR is meant to close. It also contradicts the policy the periodic evaluation applies to the very same state: in `evaluateSyncPeer` at `:1346-1355` a level holder is deliberately *not* cleared ("Don't clear peer yet, but look for better peer") and no backoff reset is issued.

**Fix:** The minimal change that keeps the fallback's purpose — unpinning a finished sync that emitted no signal — while removing the harm is to stop treating it as a *verified* success: clear and re-trigger, but do not `resetBackoff` (and so do not refill the probe budget), and log it as what it is rather than as "Sync completed successfully ... by validated work". Something like:

  services/p2p/sync_coordinator.go — split the two outcomes

```go
  func (sc *SyncCoordinator) settleSyncCompletion(currentPeer string, peerInfo *blockchain.PeerInfo,
      localChainWork []byte, verified bool) bool {
      if peerAheadByValidatedWork(peerInfo, localChainWork) { /* unchanged */ }
      if !verified {
          // No completion signal and no credit attributable to this claim: release
          // the slot so it is not pinned to the no-progress deadline, but do not
          // de-escalate backoff or refill the probe budget on unproven evidence.
          sc.logger.Infof("[SyncCoordinator] Releasing idle sync slot held by %s; level by validated work with no completion signal", currentPeer)
          sc.clearSyncPeerIfCurrent("")
          _ = sc.triggerSyncLocked()
          return true
      }
      // ... existing verified-success branch, including resetBackoff
  }
```

with `handleFSMTransition:953` passing `false` and the completion-edge and `HandleCatchupSuccess` call sites passing `true`.

If you would rather keep the fallback authoritative, then gate it on evidence attributable to *this* claim rather than on claim age — `peerInfo.LastValidatedAt.After(syncStart)` is available and round-trips over the registry gRPC (`services/blockchain/peer_registry.go:91`, `services/blockchain/peer_registry_grpc.go:239`/`:309`), so it directly answers "did this claim produce a credit". Two caveats worth weighing: `LastValidatedAt` is stamped on the blockchain service's clock while `syncStartTime` is on the p2p side, so the comparison crosses processes; and requiring evidence means the already-synced short-circuit — the case with no credit and no delivery — stops settling here and falls back to the no-progress deadline, which is the latency the fallback was added to avoid. That tradeoff is why I would take the first option. Either way, excluding the penalty case explicitly (test `chainWorkGreater(peerInfo.ValidatedChainWork, localChainWork)` directly rather than going through `peerAheadByValidatedWork`) is worth doing, and a test for a level holder whose catchup is still in the header phase would pin whichever choice you make.

## Non-blocking issues

### ChiR2 — "considered failed" is logged on the authoritative-success path for a legitimate partial catchup

**Problem:** `settleSyncCompletion`'s first branch logs `Sync with peer %s considered failed; peer still has higher validated work` (`services/p2p/sync_coordinator.go:992`), and `HandleCatchupSuccess` reaches it (`:1063`) on a report that block validation raised as a *success*. "Succeeded but still ahead" is not an error state there, it is a designed outcome: when the header fetch hits the accumulated-header cap it truncates and breaks with `stopReason = "Memory limit reached"` (`services/blockvalidation/catchup_get_block_headers.go:401-410`), `reachedTarget` stays false, nothing downstream treats that as an error, and `doCatchup` runs to `reportCatchupSuccess` (`services/blockvalidation/catchup.go:303`). A node further behind than `CatchupMaxAccumulatedHeaders` (100000 by default, `settings/settings.go:381`) reaches that cap by construction, and the catchup then completes successfully with the peer legitimately still ahead.

**Why it matters:** The behaviour is defensible — there is more chain to fetch and re-selection is a reasonable response — but the operator-visible line directly contradicts the success block validation just logged for the same catchup. That is the kind of log pair that sends someone hunting a peer problem that does not exist. I have not tried to establish how often deployments take this path; the wording defect stands regardless of its frequency.

**Fix:** Split the wording by call site, or pass the origin down, so the authoritative path says what actually happened — e.g. `Catchup with peer %s completed but peer is still ahead by validated work; re-selecting` for the `HandleCatchupSuccess` caller, keeping "considered failed" for the FSM-edge caller where the sync really did fall short with no success report. This pairs naturally with the `verified` parameter suggested in ChiR1 if you take that route.

### ChiR3 — Two new-guard branches go untested, one of them under a test named for it

**Problem:** Two things:

- `TestSyncCoordinator_HandleCatchupSuccess_ConsumesEdgeSoFreshSlotSurvives` (`services/p2p/sync_coordinator_test.go:1928`) is named and commented for the `sc.pendingCompletionPeer = ""` consumption in `HandleCatchupSuccess` (`sync_coordinator.go:1040`), but in this configuration that line is a no-op: arming `pendingCompletionPeer` requires a `CATCHINGBLOCKS -> RUNNING` edge, and the test's only prior tick is the `CATCHINGBLOCKS` observation, so nothing is pending when the success report runs. What the assertions actually pin is the causality guard at `:958` — on the trailing `RUNNING` tick `completionEdge` is true, `pendingCompletionPeer` is re-armed to `next` at `:904`, and `:958` fires because `next` was claimed after `fsmCatchingSince`, consuming the edge and returning false. So the test is mislabelled rather than redundant: it does exercise `:958`, but by a different route than `TestSyncCoordinator_HandleFSMTransition_SlotClaimedDuringCatchupNotJudgedByItsCompletion` (`:1856`) — through the success signal rather than a mid-catchup claim — while the consumption it is named for goes untested.
- The `pendingCompletionExpiry` branch (`sync_coordinator.go:970-976`) has no test. `git grep -n 'pendingCompletionExpiry' -- services/` returns only its definition and that one use, and no test touches `pendingCompletionSince`, so the 30s bound — the thing that stops a late credit settling a long-dead edge — is unexercised. `TestSyncCoordinator_HandleFSMTransition_DeferredCompletionSettlesWhenValidatedWorkArrives` (`:1876`) covers the defer-then-settle direction only.

**Why it matters:** The consumption at `:1040` and the expiry bound at `:970-976` are both guards against misattributing a verdict, which is the failure class this PR exists to remove, and a later refactor that weakens either of those two would go green. The first case is also misleading to read: the test name and comment point at the edge consumption, so someone maintaining `HandleCatchupSuccess` would reasonably believe that line is covered when the coverage is really on the causality guard.

**Fix:** For the first, arrange for the edge not to be re-armed on the trailing tick — observe an intervening `RUNNING` tick before the success report, so `lastObservedFSMState` is `RUNNING` when the post-settle tick runs and only the consumed pending state can decide the outcome — or retitle it to what it pins. For the second, arm a completion edge with the peer uncredited, push `pendingCompletionSince` back past 30s, and assert the pending state is dropped and the peer is left alone.

## Recap

| ChiR1 | Fallback settle unproven success and backoff reset | YES | 70% |
| ChiR2 | Considered-failed log on success path | | 30% |
| ChiR3 | New guards not pinned by tests | | 20% |

ctnguyen (changes requested):
Review at commit [0c48268](https://github.com/gokutheengineer/teranode/commit/0c48268730bb32f66df3570e86c02478aff7ed1a)

This round is a clear step forward. All three of my previous points are addressed, and the rework of the fallback in `2ae8136f2` is a better answer than the one I suggested: keying the release on progress age instead of claim age, and dropping the fabricated success verdict entirely (no `resetBackoff`, no probe-budget refill, no sync attempt against the peer) is both simpler and more honest than threading a `verified` flag through `settleSyncCompletion`. Testing the penalty case with `chainWorkGreater` directly in `82aba6831` closes the demoted-peer hole cleanly. The new tests are real ones too — I traced `ConsumesDeferredEdge`, `DeferredCompletionExpires` and `FallbackDoesNotReleasePenalizedAheadPeer` through the state machine and each fails if the line it names is removed, which is not something I can say about every regression test. What I found this round is one gap in the causality guard itself, plus two smaller notes.

## Status of previous points

`ChiR1` is fixed, but it is worth being precise about what fixed means: the fallback still releases a level slot holder whose catchup may be mid-header-phase, and it still clears the slot and re-triggers while that catchup runs. What changed is that the release no longer claims success — `sync_coordinator.go:970-978` no longer reaches `settleSyncCompletion`, so there is no `resetBackoff`, no probe-budget refill and no recorded sync attempt — and it now keys on `progressAge` rather than claim age, so an actively delivering level peer keeps the slot. That was the substance of the point, and it is closed.

| ChiR1 | Fallback settle unproven success and backoff reset | YES | FIXED |
| ChiR2 | Considered-failed log on success path | | FIXED |
| ChiR3 | New guards not pinned by tests | | FIXED |

## Blocking issues

### ChiR4 — The completion-edge causality guard cannot attribute an excursion, so a slot handover during another peer's header phase gets that peer's verdict

**Problem:** The guard at `sync_coordinator.go:982` is `syncStart.After(sc.fsmCatchingSince)`, and `fsmCatchingSince` is stamped at `:890` when a `CATCHINGBLOCKS` excursion is first observed. That makes the block-phase start a proxy for the catchup start — but a catchup runs its entire header phase in `RUNNING`, because `setFSMCatchingBlocks` is reached only from `fetchAndValidateBlocks` (`services/blockvalidation/catchup.go:1178`), step 10 of `catchup`. A slot claimed during peer A's header phase therefore predates the excursion A's catchup later produces, the guard reads it as "claimed before that catchup started", and the current holder B is judged by A's outcome.

This is not the sampling slop the PR already discloses. That residual is observed-versus-entered: `fsmCatchingSince` names the right instant to within one 2s tick. This is a different quantity — block-phase start versus *catchup* start — and the gap between them is the whole header phase. I read the constants: the fetch accumulates up to `blockvalidation_max_accumulated_headers` (default 100000) at `maxBlockHeadersPerRequest` (10000) per iteration before truncating at `catchup_get_block_headers.go:401-410`, and each iteration is bounded by `blockvalidation_catchup_iteration_timeout` (default 30s, applied at `catchup_get_block_headers.go:220-224`), with retries on top. That is the ~10-iteration, ~5-minute design ceiling the previous review's `ChiR1` put on record — against 2 seconds for the disclosed window.

The sequence, with both interleavings:

  - t0 — A claims the slot, its catchup starts, header phase runs in `RUNNING`.
  - t1 — B takes the slot. `preemptSyncPeer` sets `syncStartTime = t1` (`:1457`); blockvalidation keeps fetching from A under its catchup lock, which issue [bitcoin-sv/teranode#4753](https://github.com/bitcoin-sv/teranode/issues/4753) records is not aborted by a coordinator clear.
  - t2 > t1 — A's catchup enters `CATCHINGBLOCKS`; a tick stamps `fsmCatchingSince = t2`.
  - t3 — A's catchup finishes. `restoreFSMState` is deferred inside `fetchAndValidateBlocks`, so the FSM is back in `RUNNING` before `reportCatchupSuccess` is even sent at `catchup.go:303`.
  - t4 — a tick sees the `CATCHINGBLOCKS -> RUNNING` edge, arms it for B at `:912-914`, and `t1.After(t2)` is false, so the guard passes and `:1006-1007` applies A's outcome to B.

Both orderings of t3 and t4 produce the misattribution, and the edge-first one is the more likely: the FSM returns to `RUNNING` strictly before the success report is sent, and the report then has a gRPC hop and a `go` dispatch to traverse, so a tick landing in between arms and settles within that single call without ever consulting the report. `HandleCatchupSuccess` does hold the evidence that would settle this — block validation names A — but it returns at `:1062` because A is not the current peer, without touching the latch.

On reachability: the three timer-driven handover paths all require the header phase to outlast their threshold without a validated block delivery — preemption needs `progressAge > preemptionProgressGuard()` (`:1404`), which is `syncPeerNoProgressLimit()/2`, 2.5m on the 5m default (`:151`, `:239-253`); the no-progress eviction needs the full 5m (`:654`); the new fallback release needs the same 2.5m (`:972`). So this item's reachability is tied to the same unquantified header-phase duration `ChiR1` discussed, with the design ceiling above sitting right at those thresholds. A ban or a disconnect (`:1505`, `:701`) can also hand the slot over with no threshold at all. `HandleCatchupFailure` does *not* support the scenario, and it is worth saying why: a catchup that fails after entering the block phase returns the FSM to `RUNNING` through the same deferred `restoreFSMState` and only then reports the failure, so the handover happens after `fsmCatchingSince` is stamped and the existing guard suppresses it correctly. I have traced all of this statically against this tree; I have not established a rate.

**Why it matters:** The outcome is exactly the failure class this PR exists to remove. If B is still ahead it is logged "considered failed" and recycled for a catchup it never ran. If B has become level by the time A's catchup lands local work, `settleSyncCompletion` takes the success branch at `:1033-1034` — "Sync completed successfully with peer B" plus `resetBackoff()`, which de-escalates the multiplier and refills `unprovenProbeBudgetRemaining` on a non-event, the same de-escalation-without-evidence `ChiR1` was about arriving through a different door. `git grep -n 'fsmCatchingSince' -- services/` shows no test exercises a claim predating the excursion, so nothing pins this either way.

**Fix:** Stop letting the FSM edge issue verdicts at all, and make `HandleCatchupSuccess` the only thing that settles. The edge would still arm `pendingCompletionPeer`, which holds the fallback release off at `:948`; it just would not settle. The change is bigger than muting `:1006-1007`, and the reason is the expiry: `pendingCompletionExpiry` is checked at `:994`, *nested inside* the missing-validated-work branch at `:993`, so it can only ever bound an edge that is still waiting for work. Mute only the settle and the armed edge survives every following tick — `:948` skipped because the pending peer is the holder, `:993` skipped because the work is available — leaving `pendingCompletionPeer` set indefinitely and the fallback release suppressed with it. So the expiry has to be hoisted above `:993`, and `:993-1007` then collapses, because once the edge cannot settle there is no deferred evaluation left to retry:

```go
  // An armed edge must be bounded whether or not validated work is
  // available, so this check moves above the missing-work branch it
  // currently sits inside.
  if now.Sub(sc.pendingCompletionSince) > pendingCompletionExpiry {
      sc.pendingCompletionPeer = ""
      return false
  }
  // The edge says a catchup finished, not whose. Hold the fallback release
  // off and wait for HandleCatchupSuccess, which names the peer.
  return false
```

Traced forward from the arming tick: the edge stays armed for at most `pendingCompletionExpiry`, and while it is armed the only thing suppressed is the fallback release — the no-progress eviction at `:943` still runs, and `:915-919` still disarms if the slot changes hands. On the tick where the expiry fires the pending state is cleared with the slot untouched, and on the next tick `:948` is taken again, so the fallback release is eligible once more under its own conditions. That suppression has a real cost, and it is worth naming because the two clocks are independent: `pendingCompletionSince` is stamped when the edge is armed, `progressAge` runs from the holder's last validated block delivery, and nothing couples them. So when an edge is observed while the holder is *already* past the 2.5m guard, the release that would have fired on that tick is deferred until the expiry clears the pending state, at most `pendingCompletionExpiry` later. The timeline above is how a holder gets into that state — it delivers nothing across another peer's excursion, so its progress age keeps growing — but it does not establish that the guard has been crossed by the time the edge arrives: that depends on how long after the handover the excursion completes, which is unquantified. Where `HandleCatchupSuccess` fires it settles sooner; where it never fires, the expiry is the only thing that clears the arming.

Every verdict then carries a report naming a peer, checked against the slot holder at `:1062` and against the real catchup start at `:1065`, so the misattribution class is removed rather than one instance patched. Two consequences to accept, both traced: the `reportedSuccess` parameter added in `82aba6831` becomes constant-`true`, so the `ChiR2` wording collapses to one branch; and a peer served by a blockvalidation that does not set `catchup_completed = true` — the rolling-upgrade case this PR's own proto gate exists for — would never settle via the edge. The worst-case wait for that slot is then explicit: a level holder is released by the fallback once its progress age passes 2.5m, plus up to `pendingCompletionExpiry` when an edge was armed while it was already past the guard, and a still-ahead holder — the "completed but still ahead" header-cap case — waits either for preemption at `:1404`, which needs the same 2.5m *and* a strictly-higher-work rival, or for the 5m no-progress deadline; `evaluateSyncPeer` on its own does not rotate a level holder unless it finds a better peer (`:1380-1389`). Genuine catchup *failures* are unaffected: `services/blockvalidation/Server.go:1918` reports them through `ReportPeerFailure`, which the comment there documents as the sync-peer rotation signal routed to `HandleCatchupFailure`, independent of the edge.

I rejected the alternative of a slot-generation guard — recording the holder as of the start of the preceding `RUNNING` stretch and suppressing when the holder changed. It fails on the legitimate case: the peer whose catchup produces the excursion claims the slot *during* that stretch, since the claim is what triggers the catchup, and in the normal flow the slot is empty at the stretch start because the previous settle emptied it. Comparing edge-time holder against stretch-start holder therefore suppresses the ordinary verdict too, which is the proposal above reached by accident and without its honesty. Underneath that, the coordinator has no identity for "the peer blockvalidation is currently catching up from": `processCatchupChItem` is fed by every enqueue path for a block with a missing parent, not only by the coordinator's Kafka trigger, so no purely coordinator-side timestamp can attribute an excursion. Only a report that names the peer can. `ChiR6` below is a further reason the stretch boundary itself is not reliably observable.

## Non-blocking issues

### ChiR5 — The fallback comment attributes the signal-less completion to "an older blockvalidation", which is not what this tree does

*Same as [@liam's point 1 (01-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1665#pullrequestreview-5076470312).*

**Problem:** `sync_coordinator.go:949-953` says the completions the release exists for are "a trigger rejected by block validation's single-flight guard, or an already-synced short-circuit **on an older blockvalidation**". The qualifier is wrong on the blockvalidation in this repo. `git grep -n reportCatchupSuccess -- services/` gives exactly one call site, `services/blockvalidation/catchup.go:303`, at the very end of `catchup`. Two earlier paths return `nil` — a clean success to the caller — without reaching it:

  - `catchup.go:197-200`: `len(catchupCtx.headersFetchResult.Headers) == 0` logs "block already exists or no headers needed" and returns `nil`.
  - `catchup.go:263-264` hands off to `handleNoNewHeaders`, which returns `nil` at `catchup.go:328-329` when `GetBlockExists(blockUpTo)` is true.

So the already-synced short-circuit is current behaviour, not legacy behaviour. This extends liam's point rather than repeating it: he asked for the comment to match what happens because the guard arithmetic makes the release routine at mainnet block spacing; the two live short-circuits are a second, independent reason the release is an ordinary outcome rather than an escape hatch.

**Why it matters:** The comment is the only place a future reader learns when this branch is expected to fire, and "older blockvalidation" reads as dead-code-for-compatibility. Anyone tuning `preemptionProgressGuard()` or debugging a stream of `Releasing sync slot held by ...` lines will start from a false premise, and the author disclosure in the PR description carries the same qualifier.

**Fix:** Drop the version qualifier and name the current paths, e.g. "a trigger rejected by block validation's single-flight guard, or an already-synced short-circuit (`catchup.go` returns nil without reporting success when the header fetch yields nothing, or when filtering leaves no new headers and the target block is already known locally)". While there, say plainly — as liam asked — that releasing a quiet level peer between blocks is the expected steady-state outcome, not an exceptional one.

### ChiR6 — A third FSM sampling window, opened by the caught-up slow tick, is not in the residual

*Same as [@liam's point 2 (01-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1665#pullrequestreview-5076470312).*

**Problem:** `monitorFSM` at `:811-816` does not call `checkFSMState()` at all when `isCaughtUp()` is true — it only resets the timer to `slowMonitorInterval` (15s, `:150`). `isCaughtUp()` returns true once the slot is empty and no peer is ahead by validated work or advertised above local (`:480-505`). So `lastObservedFSMState` goes stale for the whole slow tick, and a complete `CATCHINGBLOCKS` excursion driven by ordinary gossip can begin and end inside it, losing the edge. The mechanism is deliberate — the comment at `:497` already reasons about this back-off — so this is a residual note, not a defect.

It extends liam's point 2 rather than repeating it. His window is opened by `checkFSMState` returning early at `:837` during backoff, is bounded by the backoff duration (up to 64s at the 32x cap), and can be open while a slot is *held*, because a peer can be claimed inside the window through `TriggerSync`, `HandleCatchupFailure` or `HandlePeerDisconnected`. This one is opened a level higher, in `monitorFSM` rather than inside `checkFSMState`, and it is newly reachable right after the fallback release empties the slot — the state this PR creates.

I traced its consequence and it is narrower than either of the other two, which is worth saying rather than leaving implied: because `isCaughtUp()` returns false whenever the slot is non-empty (`:490-492`, and the same check on the tip-unavailable path at `:513`), this window can only be open while the slot is empty, and with an empty slot `handleFSMTransition` returns at `:907-909` with no verdict owed. So no attribution is lost — only the latch's continuity.

**Why it matters:** The PR's residual enumerates blind windows and lists one, bounded by a 2s tick; liam asked for his to be added next to it. A reader auditing the latch will take that enumeration as complete and reason about `lastObservedFSMState` as though it is sampled on a fixed 2s cadence, when a caught-up node samples it at 15s or not at all.

**Fix:** Add both windows to the residual, with their bounds and the distinction above: the backoff one can drop an edge for a held slot, this one cannot. No code change is needed for this item.

## Recap

| ChiR4 | Edge verdict misattributed after slot handover | YES | 55% |
| ChiR5 | Fallback comment blames older blockvalidation | | 30% |
| ChiR6 | Third undisclosed FSM sampling window | | 15% |

ctnguyen (approved):
Review at commit [908b438](https://github.com/gokutheengineer/teranode/commit/908b4382ce832299e00d806c34a63c251034b389)

The only new work since my last round is `c9b67a105`, and it does exactly what `ChiR4` asked for rather than the minimum that would close the wording. `fsmCatchingSince` is gone from the tree, the FSM edge is now purely arm-only, the expiry check was hoisted out of the missing-validated-work branch (which then collapsed entirely, as predicted), `reportedSuccess` disappeared along with the log branch it existed for, and `HandleCatchupSuccess` is the single caller of `settleSyncCompletion` (`sync_coordinator.go:1077`). The no-progress eviction still runs ahead of the armed-window branch, so nothing got hidden behind the suppression. I also like that the tests were rewritten rather than patched — the old ones asserted behaviour that no longer exists, and deleting `DeferredCompletionSettlesWhenValidatedWorkArrives` instead of bending it is the right call. One thing the new shape introduced, below, and it is a small one.

## Status of previous points

`ChiR4` is fixed and fixed structurally: the edge cannot issue a verdict at all now, so the misattribution class is removed rather than one instance patched. `ChiR1`-`ChiR3` are unregressed — the fallback release at `:982-989` still fabricates no verdict (no `resetBackoff`, no probe-budget refill, no recorded sync attempt), the "considered failed" wording is now absent from `services/` altogether (`git grep -n 'considered failed' -- services/` returns nothing) because the branch carrying it was deleted with `reportedSuccess`, and the fallback guards are still pinned by `FallbackKeepsActivelyDeliveringLevelPeer`, `FallbackDoesNotSettleAheadPeer` and `FallbackDoesNotReleasePenalizedAheadPeer`.

Two of the statuses are narrower than they look, so worth being explicit. `ChiR5` asked only for the comment: drop the "older blockvalidation" qualifier, name the live short-circuits, and say plainly that releasing a quiet level peer is an ordinary outcome. All three landed at `:959-966`. The consequence chain @liam traced past that — the release driving `enterBackoffMode` — is still undisclosed, but he explicitly declined to ask for it, so it is not a residual against this item. `ChiR6` asked for no code change at all, only the two extra sampling windows in the residual with their bounds and the held-slot-versus-empty-slot distinction; the PR body's residual now carries all three windows with exactly that distinction, so a documentation-only answer is a complete one here.

| ChiR1 | Fallback settled unproven success | YES | FIXED |
| ChiR2 | Considered-failed log on success path | | FIXED |
| ChiR3 | New guards not pinned by tests | | FIXED |
| ChiR4 | Edge verdict misattributed after handover | YES | FIXED |
| ChiR5 | Fallback comment blames older blockvalidation | | FIXED |
| ChiR6 | Third undisclosed FSM sampling window | | FIXED |

## Non-blocking issues

### ChiR7 — Every completion edge re-stamps the armed window, so the 30s expiry does not bound the suppression
**Problem:** `sync_coordinator.go:911-913` re-stamps `pendingCompletionSince = time.Now()` on *every* `CATCHINGBLOCKS -> RUNNING` edge seen while the same peer holds the slot, and the armed branch at `:947-956` measures the expiry from that last stamp. Edges carry no peer identity — that is this commit's own reason for making them verdict-free, since catchups are enqueued by gossip and not only by the coordinator's trigger — so an excursion belonging to some other peer's catchup re-arms the window for the quiet holder and restarts its 30s. `HandleCatchupSuccess` cannot correct it either: a report naming a peer that is not the slot holder returns at `:1042`, before the disarm at `:1054`. While edges keep arriving less than `pendingCompletionExpiry` apart, the window stays armed.

What makes the deferral accumulate is that the two clocks are independent. `pendingCompletionSince` restarts on each edge, while `progressAge` runs off the holder's own `BlocksReceived`: `recordSyncPeerBlockProgress` (`:626-635`) returns early unless the peer is the slot holder, so blocks another peer's catchup delivers do not refresh the holder's progress clock even though they advance local chainwork — which only makes the holder more level, i.e. more eligible for the release being suppressed.

This is new in `c9b67a105`. Before it, the armed branch settled and cleared `pendingCompletionPeer` on the first tick where validated chainwork was available, so a re-stamp was self-limiting; the window is now purely a suppression that lives out its full expiry. I traced the mechanism statically at this head and I have not established a rate for gossip-driven excursions, only that the path exists; and the outcome is bounded by the 5-minute no-progress deadline, not unbounded.

**Why it matters:** The comment at `:948-952` says the bound exists so that a report which never arrives "cannot suppress the release indefinitely", and across repeated edges that does not hold. The cost is that a level, quiet holder the fallback was written to release *without blame* once `progressAge > preemptionProgressGuard()` (2.5m on the 5m default) instead reaches the no-progress eviction at `:939-944`, where `clearNoProgressSyncPeer` records a sync attempt against it (`:658`) and logs a WARN — a cooldown burn on a peer that, in the fallback comment's own words at `:973-974`, "did nothing wrong".

**Fix:** Stamp only when arming a window that is not already armed for this holder, so the bound covers the window rather than the most recent edge:

```go
  if completionEdge {
      // Bound the window from when it was armed, not from the last edge: an
      // edge produced by another peer's catchup must not extend it.
      if sc.pendingCompletionPeer != currentPeer {
          sc.pendingCompletionPeer = currentPeer
          sc.pendingCompletionSince = time.Now()
      }
  } else if sc.pendingCompletionPeer != "" && sc.pendingCompletionPeer != currentPeer {
```

The legitimate case survives: a second catchup for the same holder arms a fresh window, because the report for the first one already cleared `pendingCompletionPeer` at `:1054`. The tradeoff to accept is the rolling-upgrade shape where no report ever arrives — there a genuinely in-flight later report can lose the race to the fallback release by up to one expiry, which is tolerable given the release fabricates no verdict. On tests, `TestSyncCoordinator_HandleFSMTransition_ArmedEdgeHoldsFallbackUntilExpiry` (`sync_coordinator_test.go:1839`) drives arm → expire → release with a single arming; a second edge inside the window would pin this.

PR #1661, opened 2026-08-27
feat(svp2p): full bridge — multi-peer download, serving, relay, addrman and parity harness (phase 3)

ordishs:
## What

Phase 3 of the legacy-service rewrite (svp2p): the full bridge. On top of the Phase 1 transport/handshake and the Phase 2 header index + headers-first IBD, this phase adds multi-peer block download, block/tx serving, block and tx relay, tx ingestion with an orphan pool, a `CAddrMan` port with `peers.json`, addrman-driven outbound connections, SVNode-faithful ban scoring, and an in-process parity harness that drives legacy and svp2p from the same scripted peer.

The service stays dark-launched: `-svp2p=1` only, mutually exclusive with `-legacy=1`. Operators change no settings.

Design spec, plan and execution ledger live in `docs/superpowers/` (gitignored, available on request).

## Why

Phase 2's testnet gate showed IBD well below bitcoin-sv, attributed to single-peer block download. Phase 3 was commissioned to close that gap and to finish the bridge so the service can hold at tip and serve peers.

## How

**Multi-peer block download (Tasks 1-7, 6b)**
- Work-based header index, contextual difficulty (DAA / cw-144) and checkpoint fence on every new header from every peer.
- Sync-peer rotation releases its block window; eager availability with SVNode's 24 h near-tip relaxation; blocks scheduled across every useful peer; parallel fetch of the same block from up to `legacy_blockDownloadMaxParallelFetch` peers.
- Per-block download timeout kept unconditional as in SVNode, with SVNode's three percentages exposed as settings (see table).

**Serving and relay (Tasks 8-13)**
- `getheaders` / `getblocks` answered from the ACTIVE tip (never the header-index tip); empty `headers` reply kept as SVNode does.
- `getdata` served with streaming block sends; blocks over 4 GiB answer `notfound` until Phase 4's extmsg.
- Block announcements from the blocks-final topic, tx announcements from txmeta, tx inv round-trip over the legacy inv topic.

**Ingest (Tasks 14-15, 21)**
- Transactions ingested through the validator; orphan transaction pool; parent-missing blocks retained in the TempStore and replayed when the parent lands (no re-fetch).

**Peers (Tasks 17-20, 23)**
- `CAddrMan` port with `peers.json` persistence; `getaddr` answered for inbound peers once per connection; addrman-driven outbound dialer up to `legacy_targetOutboundPeers`.
[...]

Issue #1610, opened 2026-08-18
svp2p: SVNode-faithful rewrite of the legacy P2P bridge — phase tracker


ordishs:
Tracker for the rewrite of the legacy P2P bridge as `services/svp2p`, a Go implementation modeled on the SVNode reference (`net.cpp` / `net_processing.cpp`) instead of the extracted bsvd stack.

The new service is inert on main: it runs only with `-svp2p=1`, is mutually exclusive with `-legacy=1`, and the daemon default remains the legacy service until cutover. Phases merge to main dark and developer-verified.

## Phases

- [x] **Phase 1 — transport + handshake** (#1609). Framing, flow control, SVNode-faithful version/verack/protoconf handshake, ban list, peer manager, `peer_api`, daemon flag. Verified against a live testnet SVNode peer.
- [ ] **Phase 2 — header index + IBD scheduling + bridge relocation.** Exit: syncs testnet end to end.
- [ ] **Phase 3 — serving + relay.** getdata/getheaders service, tx/block announcements. Exit: full-bridge behavior, parity harness green.
- [ ] **Phase 4 — associations/multistreams + extended messages (>4 GiB).**
- [ ] **Phase 5 — compact blocks** (receive path, then high-bandwidth), behind its own flag, default-off. Additive — the legacy service never had compact blocks, so this does not gate drop-in status.
- [ ] **Phase 6 — survivor relocation, soak, cutover, deletion** of the bsvd-derived packages (~45k lines).

## QA entry criterion

The test team engages only when svp2p is a drop-in replacement for the legacy service:

- Phases 1–4 merged.
- Parity harness green across the recorded corpus (old vs new service, identical externally visible behavior).
- Testnet IBD reaches the same chain tip as the legacy service.
- An operator can switch `-legacy=1` to `-svp2p=1` with zero settings changes.

When every box above is ticked, this issue gets the QA handoff comment and the `not-QA-ready` labels come off. Until then, nothing here is a QA target.

## PR convention

Every phase PR: title carries `svp2p` and the phase number, labels `svp2p` + `not-QA-ready`, and links back to this tracker.

PR #1666, opened 2026-08-28
feat(svp2p): multistreams and the extended header for blocks over 4 GiB (phase 4, stacked on #1661)

ordishs:
## What

Phase 4 of the legacy-service rewrite (svp2p): SVNode multistreams (associations, the `BlockPriority` DATA1 stream) and the extended message header for blocks over 4 GiB. Stacked on the Phase 3 PR (#1661); this diff shrinks to Phase 4 alone once that lands.

Still dark-launched: `-svp2p=1` only, mutually exclusive with `-legacy=1`. No new settings keys — multistreams are gated by the existing `legacy_allowBlockPriority` and the block cap by `excessiveblocksize`.

## Why

svp2p ran one TCP stream per peer and answered `notfound` for any block a 32-bit length field cannot declare. SVNode (and legacy) open a second stream for blocks/pings so a large block never queues behind relay traffic, and frame >4 GiB payloads with `extmsg`. Both were parity gaps; the second was watch-list scenario 9, the only scenario without a verdict.

## How

**transport**
- `PeerConn` interface: the surface `protocol.Peer` consumes; `*Conn` and the new `*Association` both satisfy it, so no state machine changed.
- `Association`: one `Conn` per stream type (GENERAL always, DATA1 optional), merged inbound with per-stream backpressure, outbound routed by policy, teardown of every stream on any stream's failure (the PR 1286 invariant), byte counters summed.
- Policies ported from `stream_policy.cpp`: `Default`; `BlockPriority` routes exactly SVNode's `IsHighPriorityMsg` set (block, cmpctblock, blocktxn, getblocktxn, headers, getheaders, hdrsen, gethdrsen, ping, pong) to DATA1.
- Extended header (`protocol.cpp:220-263`): read side recognises `extmsg` + `0xffffffff` and streams the 64-bit-length block; send side frames above `math.MaxUint32 - 1` with a zero checksum (one pass, no hashing); both gated on negotiated version ≥ 70016. Extended non-block frames are refused (every legal one exceeds our receive limit). `Config.MaxBlockPayload` carries `excessiveblocksize` into the transport; `wire.SetLimits` now reads the same setting instead of a 4,000,000,000 literal.

**protocol**
- Our association ID rides in `version` (outbound generates one; inbound echoes the dialer's, as SVNode does). Policy chosen after the peer's protoconf (`GetPreferredStreamPolicyName`).
- Inbound: a fresh connection's first message is classified — `version` → peer; `createstream` → `MoveStream` (same-IP check with a 24 h ban on mismatch, refuse unknown ID / duplicate type / unknown policy with `reject 0x60`, ack on the new stream, no `Peer` for the husk); `streamack` unsolicited → reject; anything else → disconnect; silent husks time out.
- Outbound: after a BlockPriority handshake, a single drain goroutine dials the peer again, sends `createstream` first, awaits `streamack` (5 s), attaches DATA1; failures leave the peer single-stream.
- `getdata` serves blocks over 4 GiB to 70016 peers; `notfound` only for older peers.

## Testing

- Transport: extended header round trips at 4 GiB − 1 / 4 GiB / 4 GiB + 1 with generated bodies (≈110 s under `-race`); association routing, merge, teardown, close-releases-parked-reader; policy tables.
- Protocol: inbound classification (attach + ack on the new stream; other-IP → ban; duplicate/unknown/off → reject; extmsg or over-length first message → disconnect; silent husk timeout); outbound DATA1 open, no-dial cases, silent-ack does not stall the queue, Stop abandons a parked dial.
- Integration: two svp2p stacks open DATA1 to each other; a real block download travels on DATA1 (blocks on DATA1 > 0, none on GENERAL after the ack).
- Parity: scenario 9 adjudicated — svp2p frames 4,294,967,297 bytes with the extended header and the peer receives every byte; legacy reads the body and sends nothing (go-wire drops the frame). New rows 13 (BlockPriority negotiation) and 14 (`headers`/`getheaders` routing: legacy's set is narrower than SVNode's). Scenario 8 (cold start) and 12 (user-agent fence) were closed in the Phase 3 PR.
- Live, regtest against bitcoin-sv (Docker): outbound DATA1 negotiated and acked by SVNode, blocks arrive on DATA1 (per-socket byte counters); inbound DATA1 after the association-ID echo fix; a genuine 4.63 GB block mined on SVNode arrived on DATA1 with the extended header and was ingested in 62 s (after a first attempt hit a local backoff — see residuals).
- `make test` on the final head (`eb928c0bb`): 14,574 tests, 47 skipped, 0 failures. `go test ./services/svp2p/... ./daemon/ ./settings/ -race -count=1`: all ok (parity 312 s, transport 120 s, protocol 117 s). `golangci-lint run services/svp2p/...`: 0 issues. One Phase 3 test (`TestServerStopFlushesLegacyInvProducerDespiteAnEarlyReturn`) flaked once in an earlier full run and passed 6/6 in isolation and in the final run — carried as a flake to harden.

### Known residuals (carried, not fixed here)

- **Phase 3 scheduler:** when the admission gate refuses a block on backoff (`SkipForBackoff`), the ingest path still drains the whole declared payload off the socket and `BlockFailed` re-offers the hash with no spacing, so the same peer re-serves it every sync tick until the backoff elapses. On the live 4.63 GB regtest block that was ~5× the block's bytes across one failed attempt. Phase 4 does not change the mechanism, only the ceiling (`excessiveblocksize`). Fix shape: defer the re-offer until `Admission.BackoffRemaining` via the existing `retryAfter` map. This should land before >4 GiB blocks are enabled in production.
- `excessiveblocksize=0` is documented as "unlimited" but the wire layer uses the 4 GiB default (go-wire cannot size messages from 0).
- The two basic-path 4 GiB round-trip subtests cost ~100 s under `-race`; they skip under `-short`.

## Tracking

Phase 4 of 6 — tracker: #1610. Stacked on #1661 (Phase 3). Not a QA target until the tracker's drop-in criterion is met.


ordishs:
Superseded by https://github.com/bsv-blockchain/teranode/pull/1675, which consolidates the svp2p phases into one service PR (everything svp2p-specific under services/svp2p, off by default via startSvp2p / -svp2p=1). The per-phase review history stays on this PR's branch.

PR #1668, opened 2026-08-31
fix(p2p): charge the peer that served invalid subtree bytes, not the gossip announcer

gokutheengineer:
**Closes bitcoin-sv/teranode#4756.**

### Problem
`ReportInvalidSubtree` resolved the peer to charge from `subtreePeerMap` first and used `peerURL` only as a fallback. The map is populated from gossip, so it holds whoever ANNOUNCED the subtree hash, while `peerURL` is the DataHub the invalid bytes were actually fetched from. Whenever the two differed (announcements come from the miner, bodies from whichever peer's DataHub the block came from), the honest announcer absorbed the `InteractionFailures` charge and the peer that served the bad bytes recorded nothing. Since gossip-only peers accrue no `InteractionSuccesses`, a single mischarged failure drops them from reputation 50 to 5, below both the gossip filter (<20) and the unhealthy gate (<40).

### Fix
- `KafkaInvalidSubtreeTopicMessage` gains a `peer_id` field carrying the serving peer's ID, as recommended in the issue's verification notes. subtreevalidation plumbs `ValidateSubtree.PeerID` (already paired with `BaseURL` at both origins: the subtree gossip announcement and the block-validation request) through all six `publishInvalidSubtree` sites.
- p2p `ReportInvalidSubtree` attribution order is now: explicit peer ID from the message, then the registry owner of `peerURL`, and the gossip announcer in `subtreePeerMap` only when no URL was supplied at all. An unresolvable URL means no charge instead of charging the announcer. When the announcer and the charged peer differ, the discrepancy is logged; when an explicit peer ID does not own the reported URL, a warning makes the producer-side pairing invariant self-enforcing.
- The map entry is consumed on every outcome (single snapshot at the top), so an uncharged entry can no longer linger against the map cap or misattribute a later report.
- Mixed-version rollout is safe in both directions: an old producer leaves `peer_id` empty and the new consumer falls back to URL resolution; a new producer's field is ignored by an old consumer.

Improvement over the issue's proposed fix: instead of only re-resolving the URL owner at report time (exact-string registry scan, fragile), the serving peer ID travels in the Kafka message itself, per the verification-note recommendation.

Deliberate behavior change worth noting: on the legacy path (`BaseURL == "legacy"`) and any other unresolvable URL, nobody is charged where the old code charged the gossip announcer. A drop in `InteractionFailures` volume after deploy is expected, not a regression.

### Residual
- `CheckSubtreeFromBlockRequest` carries no `peer_id`, so `CheckSubtree`-driven reports publish an empty peer ID and rely on URL-owner resolution. Extending that gRPC API is out of scope here.
- `getPeerIDFromDataHubURL` remains an exact string match over `ListPeers`; URL normalization (trailing slash, scheme case) would remove a false-negative class but is pre-existing behavior.
- The invalid-subtree dedup map is still keyed by subtree hash only, so within the 1-minute TTL only the first serving peer of a given bad subtree is charged (pre-existing).

### Tests
New `services/p2p/invalid_subtree_report_test.go` (real `CentralizedPeerRegistry`, no mocks) pins the bug — each attribution test fails on the old map-first logic:

- announcer in map + explicit serving peer ID → serving peer charged, announcer untouched, map entry consumed
- announcer in map + URL owned by a different registered peer, no explicit ID → URL owner charged
- unresolvable URL → nobody charged, map entry still consumed
- no URL and no ID → falls back to the announcer
- explicit ID with empty URL → charged
- undecodable peer ID → nobody charged, no error, map entry consumed
- Kafka consumer path end-to-end: `peer_id` from `KafkaInvalidSubtreeTopicMessage` wins over the map announcer

`Server_coverage_test.go` asserts `PeerId` round-trips through the published Kafka message.

Ran (affected packages only):

```
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/p2p/...        # ok
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./util/kafka/...          # ok
go vet ./services/p2p/... ./services/subtreevalidation/... ./util/kafka/...                    # clean
```

`services/subtreevalidation` tests compile (`go vet` covers the `_test.go` files) but cannot link locally due to the known GoBDK cgo linking issue on this machine; CI covers them.


services/subtreevalidation/invalid_subtree_test.go
```diff
@@ -239,7 +239,7 @@ func TestInvalidSubtreeReporting_PeerCannotProvideData(t *testing.T) {
 		httpmock.NewBytesResponder(http.StatusNotFound, []byte(errors.New(errors.ERR_NOT_FOUND, "not found").Error())))
 
 	stat := gocore.NewStat("test")
-	_, err = server.getSubtreeTxHashes(ctx, stat, &subtreeHash, baseURL)
+	_, err = server.getSubtreeTxHashes(ctx, stat, &subtreeHash, baseURL, "")
```
freemans13, line 242: This is where the missing test is cheapest. A few lines below, this test already unmarshals the published Kafka message and asserts `SubtreeHash`, `PeerUrl` and `Reason`. Pass a real peer ID here instead of `""` and add `assert.Equal(t, "<that id>", msg.PeerId)` next to the existing assertions. Same for the `getMissingTransactionsBatch` call at line 214.

That pins four of the six `publishInvalidSubtree` sites: three inside `getMissingTransactionsBatch` and one inside `getSubtreeTxHashes`. The other two sit behind `processMissingTransactions` and `getSubtreeMissingTxs`, which no existing test reaches with a Kafka producer attached, so I would not ask for those here.

Why it matters more than a coverage number: an empty peer ID is not a visible failure. It falls back to URL resolution and every assertion in the suite still passes, so the plumbing this PR adds is the part with no test that can tell the fix from its absence.

services/p2p/Server.go
```diff
@@ -160,7 +160,7 @@ type Server struct {
 	nodeStatusTopicName               string                         // pubsub topic for node status messages
 	topicPrefix                       string                         // Chain identifier prefix for topic validation
 	blockPeerMap                      cappedPeerMap                  // Which peer sent each block (canonical hash -> peerMapEntry); insert-capped, issue 1409
-	subtreePeerMap                    cappedPeerMap                  // Which peer sent each subtree (canonical hash -> peerMapEntry); insert-capped, issue 1409
+	subtreePeerMap                    cappedPeerMap                  // Which peer ANNOUNCED each subtree via gossip, not necessarily who served its bytes (canonical hash -> peerMapEntry); insert-capped, issue 1409
```
freemans13, line 163: Observation, not a request, and possibly a follow-up issue rather than anything for this branch.

After this change the only reader of the subtree half of this map is the `peerID == "" && peerURL == ""` fallback in `ReportInvalidSubtree`. Most publish sites cannot reach it: `getMissingTransactionsBatch` rejects a baseURL that is not http or https before it publishes, and `getSubtreeTxHashes` errors out on an empty one. The fallback is left firing only for the two deeper sites, and only when `ValidateSubtree.BaseURL` is itself empty.

Meanwhile the map is still written on every gossip subtree announcement, which the `p2p_peer_map_max_size` documentation sizes for roughly 16 subtrees per second, and it carries eviction attribution plus a cleanup sweep that holds the maps' mutex against incoming announcements for the length of its walk. None of that machinery can actually be deleted, because `blockPeerMap` keeps it alive for `ReportInvalidBlock` either way. The narrower question is whether the subtree writes still earn their share of it.

util/kafka/kafka_message/kafka_messages.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.11
-// 	protoc        v7.34.1
+// 	protoc-gen-go v1.36.12
```
freemans13, line 3: This file is now the only generated file in the repo at protoc-gen-go v1.36.12 and protoc v7.35.1. Fourteen others sit at v1.36.11 and v7.34.1.

The wire format is unaffected and I would not rerun anything for this alone. It is only that the next person to run `make gen` on the pinned toolchain will churn these two lines back and get a diff they did not ask for.

freemans13 (approved):
Approving. The attribution reorder is right, and the p2p side is pinned by tests that genuinely fail on the old map-first logic rather than merely covering it.

I walked the execution path rather than the diff, since the risk here is a peer ID that quietly arrives empty. Both origins that populate `ValidateSubtree.PeerID` pair it with a DataHub URL (the HTTP endpoint a peer serves subtree bytes from) that the same peer owns. The gossip announcement authenticates the claimed `PeerID` against the real libp2p sender before pairing it with that sender's DataHubURL, at `services/p2p/server_helpers.go:227`. `CheckBlockSubtrees` takes both from the block source, at `services/subtreevalidation/check_block_subtrees.go:381`. All four `ValidateBlockWithOptions` call sites pair `opts.PeerID` with a `baseURL` from the same peer, and `RevalidateBlock` goes further by resolving the URL from `GetPeer(peerID).DataHubURL`, at `services/blockvalidation/Server.go:1250`. All six `publishInvalidSubtree` sites carry the ID.

One consumer worth naming, because it closes a question the reorder opens: the explicit peer ID is the first attribution source that can name a peer the registry has never heard of, where the old map and URL lookups could only produce peers it already held. `UpdateMetrics` returns early on an unknown ID, at `services/blockchain/peer_registry.go:384`, so there is no way to grow the registry by reporting arbitrary valid peer IDs.

The one thing I would add before merge, and it does not block: nothing proves the peer ID actually reaches the Kafka message from the validation path. All eleven updated subtreevalidation call sites pass `""`, and the only assertion on `PeerId` calls `publishInvalidSubtree` directly with a literal. If one of the six threading sites had been left empty or wired to the wrong variable, the suite stays green and the fix degrades silently to URL-only resolution, which is the behaviour you are replacing. There is a cheap version of the test, detailed inline.

Two smaller observations are inline as well: the subtree half of `subtreePeerMap` is close to vestigial after this change, and the regenerated `.pb.go` carries a protoc toolchain version no other generated file in the repo uses.

Verified locally on 9d4f6d8. `go test -race -tags testtxmetacache -count=1 ./services/p2p/...` passes in 13.4s, including the six new attribution tests. `TestInvalidSubtreeReporting|TestPublishInvalidSubtree` in `services/subtreevalidation` passes here, so the GoBDK linking problem in the description looks local to your machine rather than to the branch. `go vet` over p2p, subtreevalidation and util/kafka is clean. CI is green across all 38 checks. What I did not check: none of this ran against a live peer set, so the expected drop in `InteractionFailures` volume after deploy is a prediction, not a measurement.

liam (approved):
Reviewed independently on top of the existing review round. I checked the load-bearing claims rather than reading the diff alone, and ran the affected suites on the branch.

**Verified:**

- The plumbing actually reaches the paths that matter, which is the part the fix stands or falls on. `ValidateSubtree.PeerID` is populated at both real origins: `subtreeHandler.go:147` (gossip) and `check_block_subtrees.go:669` (block validation, from `request.PeerId`). The only origins leaving it empty are the two `CheckSubtree` sites (`Server.go:849`, `:942`), which is exactly the documented `CheckSubtreeFromBlockRequest` residual. Nothing else silently degrades to URL resolution.
- `ReportInvalidSubtree` is a `Server` method in no interface, so the signature change is contained, and `getPeerFromMap` is still live for the block path, so nothing is left orphaned.
- The unconditional map consumption at the top is a real behaviour change and the right one.
- `SETTINGS_CONTEXT=test go test -race -tags testtxmetacache ./services/p2p/... ./util/kafka/...` green; `go vet` clean.

Three notes, none blocking.

**1. Worth one sentence on why this PR and 1669 order attribution differently.**

This PR demotes the gossip map below URL resolution; 1669 keeps it above for blocks. Read side by side that looks like drift, and it is not. Both `handleBlockTopic:75-80` and the subtree handler reject on `fromID != message.PeerID`, so both maps hold the verified originator rather than a relay. For blocks the announcement and the fetch are the same event by the same peer, so map and URL agree and the ordering is moot. For subtrees they are two different events, the miner announces and the block's DataHub serves, which is the entire bug being fixed here. Stating that in the description would stop the asymmetry reading as an oversight in review.

**2. Proto field naming.**

`peer_id` is snake_case in a message whose other fields are camelCase (`subtreeHash`, `peerUrl`, `reason`). Harmless on the wire, but protobuf JSON lowerCamelCases the field name, so the serialized name is `peerId` while the doc example in `kafkaMessageFormat.md` writes `"peer_id"`. Naming the field `peerId` matches its siblings and makes the example correct under either convention. 1669 has the same thing with `peer_id` / `peer_url`.

**3. This PR and 1669 conflict with each other.**

`git merge-tree pr1668 pr1669` reports conflicts in `services/p2p/Server.go` and `util/kafka/kafka_message/kafka_messages.pb.go`. The `.proto` auto-merges cleanly; the generated `.pb.go` does not, because both regenerate the rawDesc bytes. Whoever lands second should rerun `make gen` rather than hand-resolve conflict markers inside the generated file.

I also back the earlier request for a peer-ID assertion in the subtree-validation test. An empty peer ID is not a visible failure, it falls through to URL resolution and every existing assertion still passes, so without that assertion there is no test that can tell this fix from its absence.

Approving.

PR #1669, opened 2026-08-31
p2p: fair-share peer map eviction and invalid-block provenance for ban attribution

gokutheengineer:
**Closes bitcoin-sv/teranode#4755.**

### Problem

The issue's four items, checked against current main:

- Unbounded insertion into the gossip peer maps: already fixed on main (`cappedPeerMap`, inline cap at insert) - skipped.
- Dead `p2p_peer_map_*` settings keys: already wired in `settings/settings.go` - skipped.
- Cross-peer attribution washout: still open. The map is one space shared by all peers, so a flooder spraying distinct hashes ages other peers' entries out of the cap before validation reports, and the 10-point `invalid_block` score is never applied.
- Missing attribution fallback: still open. `processInvalidBlockMessage` and `ReportInvalidBlock` gave up with "no peer found for block" when the map entry was gone - unlike the subtree path, there was no URL fallback - so a peer could void its own ban by flooding its entry out (the issue's exploit sequence).

### Fix

**Fair-share eviction (cross-peer washout).** `cappedPeerMap` now threads each peer's entries onto a per-peer insertion-order list alongside the global one. At capacity, an inserting peer already holding its fair share of the map (cap divided by the number of peers present) evicts its own oldest entry; only a peer below that share evicts the global oldest. Past its share, a flooder's pressure lands on its own junk instead of other peers' attribution, with no new tuning knob - the share adapts to the peers actually in the map. Eviction stays O(1) per insert.

**Provenance in the invalid-block report (self-eviction / issue exploit).** `KafkaInvalidBlockTopicMessage` gained `peer_id` and `peer_url`. Block validation fills them from `opts.PeerID`/`baseURL` at every invalidation site that knows them (the gossip, catchup and revalidation paths all do; the `setTxMined` path deliberately passes empty, and the `"legacy"` URL sentinel is stripped). The p2p consumer resolves attribution message-peer-ID first, then the peer map, then a DataHub-URL registry lookup - so a washed-out map entry no longer voids the ban, closing the issue's trigger sequence. This is stronger than the proposed URL-only fallback: the peer ID is the spoof-checked gossip announcer, carried end-to-end and immune to map eviction. `ReportInvalidBlock` (and its unused `InvalidBlockHandler` interface) gained the same URL fallback, mirroring `ReportInvalidSubtree`.

**Scoring idempotence.** With attribution now carried in the message, the old implicit dedupe (ban gated on a map entry deleted after scoring) no longer applied, and Kafka is at-least-once - a redelivered message would have scored the peer again (10 points each, 100 threshold). A TTL-bounded `reportedInvalidBlocks` map (same `cappedPeerMap` type, same cap/TTL/sweep) now records scored hashes after a successful `AddBanScore`, so one invalid block scores once per TTL, not once per delivery, while consumer retries of a failed `AddBanScore` still go through.

An independent fresh-eyes review of the change (including a randomized invariant fuzz over the three-structure bookkeeping - map, global list, per-peer lists - with live cap changes) found no bookkeeping, locking, or termination bugs; its findings on wording, idempotence, the `"legacy"` sentinel, a stale test mock, and doc examples are fixed in the second commit. Docs updated where they described the old semantics (`kafkaMessageFormat.md` including examples, `p2p_peer_map_max_size` longdesc, the at-capacity log line, code comments).

### Residual

- Fair share bounds cross-peer washout, it does not eliminate it: while ramping up to its share on a full map, a flooder still displaces up to one share of other peers' oldest entries (previously: the entire map), and a sybil flood can wash more at one live libp2p connection per share. A full fix needs connection admission / gossip rate limiting (tracked separately). The provenance fields make both residuals harmless for the invalid-block ban path. The code comments and settings longdesc state these bounds explicitly.
- `ReportInvalidBlock` / `InvalidBlockHandler` have no callers in the repo; the method keeps map-then-URL precedence (it has no peer-ID input) while the live Kafka path prefers the message's peer ID. Left as-is rather than deleted - removing the interface is a separate decision.
- Pre-existing and out of scope: the `checkOldBlockIDs` -> `ErrBlockInvalid` branch in `ValidateBlockWithOptions` invalidates via `blockchainClient.InvalidateBlock` directly, publishing no invalid-block event, so that path never scores a peer. `opts.PeerID`/`baseURL` are in scope there if this should change.
- `services/blockvalidation` unit tests cannot link on this machine (known local GoBDK cgo issue, final link step only); the package and its tests are typechecked via `go vet -tags testtxmetacache` and the `catchup` subpackage suite passes. CI runs the full suite.

### Tests

New tests:

- `TestCappedPeerMapFloodEvictsFlooderNotVictim` - honest attribution survives a 50x-cap flood; the flooder's own entries are the ones evicted.
- `TestCappedPeerMapFairShareAdapts` - the share denominator is the peers present: a peer at its share self-evicts, a new peer takes the global oldest.
- `TestStorePeerMapEntryFloodCannotWashOutAttribution` - the original exploit end-to-end through the production helpers.
- `TestProcessInvalidBlockMessage_PeerIdSurvivesMapWashout`, `_PeerIdBeatsMapEntry`, `_PeerUrlFallback`, `TestReportInvalidBlock_PeerURLFallback` - the new attribution resolution order.
- `TestProcessInvalidBlockMessage_DuplicateDeliveryScoresOnce` - redelivery does not re-score.
- `requireMapConsistent` extended to assert the per-peer lists stay in step with the map and global list (including the empty-map and re-attribution paths).

Commands run (affected packages only):

```
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/p2p/...      # ok
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./util/kafka/... ./interfaces/... ./settings/ ./services/blockvalidation/catchup/  # ok
go vet -tags "testtxmetacache" ./services/blockvalidation/ ./services/p2p/ ./util/kafka/... ./interfaces/... ./settings/  # clean
```


settings/p2p_settings.go
```diff
@@ -48,7 +48,7 @@ type P2PSettings struct {
...
...t peer instead.\n\nA higher cap also lengthens the cleanup sweep's walk, which holds the maps' mutex against every incoming announcement for its duration — negligible at 10000, less so an order of magnitude above it. Size for peak legitimate announcement rate over one TTL window, not for the worst flood observed.\n\n### Trade-offs\n| Setting | Benefit | Drawback |\n|---------|---------|----------|\n| Higher | Better attribution for more blocks | More memory usage |\n| Lower | Lower memory | May lose attribution for older blocks |\n\n### Recommendations\n- **10000** (default) - Handles large networks at mainnet subtree rates\n- Lower for memory-constrained environments\n- Raise only against a measured, peer-spread at-capacity warning. Note that the cleanup sweep walks every entry under the maps' mutex, and the gossip handlers that block on it are a bounded pool (p2p_gossip_handler_concurrency), so the walk rather than the memory is what sets the practical ceiling.\n- 0 or negative selects the default rather than disabling the bound: there is no unbounded mode, and the coercion is logged at startup. This differs from p2p_peer_registry_max_size, where 0 does disable LRU enforcement."`
```
freemans13, line 51: The opening line of this longdesc still says the cap covers `blockPeerMap, subtreePeerMap`. As of this PR it governs three maps: `applyPeerMapLimits` now also sets it on `reportedInvalidBlocks` (`services/p2p/Server.go:738`). An operator sizing this key would want to know the dedupe records share the same budget, even though at 10000 entries against a 10 minute TTL they will never come close to competing for it.

The same drift is in `Stop`, where the comment reads `Clear the peer maps to free memory` above three `Clear()` calls, and in the at-capacity warning text, which says `peer maps` but only reports two of the three.

services/p2p/Server.go
```diff
@@ -161,6 +161,7 @@ type Server struct {
 	topicPrefix                       string                         // Chain identifier prefix for topic validation
 	blockPeerMap                      cappedPeerMap                  // Which peer sent each block (canonical hash -> peerMapEntry); insert-capped, issue 1409
 	subtreePeerMap                    cappedPeerMap                  // Which peer sent each subtree (canonical hash -> peerMapEntry); insert-capped, issue 1409
+	reportedInvalidBlocks             cappedPeerMap                  // Invalid blocks already scored (canonical hash -> scoring record); dedupes at-least-once Kafka redelivery so one invalid block is scored once per TTL, not once per delivery
```
freemans13, line 165: Reusing `cappedPeerMap` here is a sensible way to get the TTL sweep and the bound for free, but it also inherits the fair-share eviction policy, and that policy is written for a completely different input. In the gossip maps the per-peer list key is the peer whose announcement is being recorded, and self-eviction past a peer's share is the point. Here the key is the peer that was scored, so a repeat offender crosses its share first and drops its own oldest dedupe records first, which is exactly the peer whose records you would rather keep.

The effect is benign, since losing a dedupe record only means a redelivery scores an already-guilty peer a second time. It is worth one sentence on this field saying so, because the whole `cappedPeerMap` type comment argues the opposite direction and a future reader will try to map it onto this use.

services/p2p/server_helpers.go
```diff
@@ -822,11 +822,43 @@ func (s *Server) processInvalidBlockMessage(message *kafka.KafkaMessage) error {
...
-		s.logger.Warnf("[processInvalidBlockMessage] %v", err)
-		return nil // Not an error, just no peer to ban
+	// Kafka is at-least-once and the revalidation path can republish the same
+	// block, so a block must be scored once, not once per delivery. The old
+	// code got that for free — the ban was gated on a peer-map entry deleted
+	// after scoring — but the message now carries its own attribution, so the
+	// dedupe has to be explicit. The record is TTL-bounded like the peer maps:
+	// a redelivery after the TTL can re-score, which matches the attribution
+	// window the map-gated path always had.
+	if prev, ok := s.reportedInvalidBlocks.Load(blockHash); ok {
```
freemans13, line 832: The dedupe key is the block hash on its own, so within the TTL window only the first peer to be scored for a given invalid block is ever scored for it. A second peer that serves the same bad block inside those 10 minutes gets a free pass.

Most of the time that never comes up, because `storeInvalidBlock` writes the block with the invalid flag and the next announcement of the same hash short-circuits at the `GetBlockExists` check in `blockvalidation/Server.go` before validation runs again. But `storeInvalidBlock` only logs when its `AddBlock` fails (`BlockValidation.go:2250`) and publishes to Kafka regardless. On that path the block is not stored, `GetBlockExists` stays false, and a second peer announcing the same block does get revalidated and republished. The dedupe then swallows the score for that second peer.

Keying the record on hash plus peer ID would keep the at-least-once redelivery guarantee this is here for, and still score a genuinely distinct peer. If you have already decided that per-hash is the behaviour you want, a line saying so would settle it, since the comment above currently only argues the redelivery case.

services/p2p/server_helpers.go
```diff
@@ -916,6 +952,7 @@ func (s *Server) cleanupPeerMaps() {
 	ttlCutoff := now.Add(-s.peerMapTTLOrDefault())
 	blockExpired := s.blockPeerMap.DeleteExpired(ttlCutoff)
 	subtreeExpired := s.subtreePeerMap.DeleteExpired(ttlCutoff)
+	s.reportedInvalidBlocks.DeleteExpired(ttlCutoff)
```
freemans13, line 955: Eviction pressure on this third map is invisible. The `DeleteExpired` return value is discarded here, unlike the two lines above it, and `EvictionsSinceLastRead` is never called on `reportedInvalidBlocks` anywhere, so the at-capacity warning below covers only the block and subtree maps.

In practice the map cannot fill: it would take 10000 distinct invalid blocks inside one 10 minute TTL. But if it ever did, the dedupe would silently degrade back to scoring once per delivery with nothing in the log to say why the ban scores jumped. Either fold the count into the existing sweep log line, or add a comment stating that this map is not expected to reach its cap and that is why it is not reported.

freemans13 (approved):
Approving. The fair-share eviction and the end-to-end provenance both do what they claim, and I could not find a correctness or concurrency defect. Four low-severity notes inline, none blocking.

## What I checked beyond reading the diff

**The peer ID the Kafka message now carries is authenticated.** That is the load-bearing claim of the whole change, because the PR makes it outrank the peer map. `handleBlockTopic` rejects any gossip message whose declared `PeerID` differs from the libp2p sender that delivered it, and scores a protocol violation for the attempt (`services/p2p/server_helpers.go:77`). So the `PeerId` written into the block topic message at `server_helpers.go:164` can only ever be the real announcer. Preferring it over the map is safe and is not a new spoofing surface.

**The DataHub URL fallback actually resolves.** `getPeerIDFromDataHubURL` matches the peer registry by exact string equality, which only works if both sides hold the same string. They do: the same `blockMessage.DataHubURL` value is written to the registry at `server_helpers.go:141` and into the Kafka message at `:164`. The `"legacy"` strip in `kafkaNotifyBlockInvalid` is correct too, since that sentinel has no URL scheme and is never a registry DataHub URL.

**The dedupe cannot race itself.** `kafkaNotifyBlockInvalid` publishes with `Key: block.Hash().String()`, so every redelivery of one block hashes to the same Kafka partition and is handled by one consumer in sequence. That is what makes the non-atomic Load-then-Store in `processInvalidBlockMessage` safe. It is worth a sentence in the code, because the safety comes entirely from the partition key and not from anything visible at the dedupe site.

**Provenance is not silently lost on the revalidation retries.** `ReValidateBlock` drops `opts.PeerID` (`BlockValidation.go:2538`), which looked like a hole in the new attribution. It is not: that path sets `useFullValidation=false` and lands in `reValidateBlock`, which never calls `storeInvalidBlock` or `markBlockAsInvalid`. No invalid-block message is ever published with the peer ID quietly dropped by a retry.

**The interface change is inert.** `NewBlockValidation` takes the handler through `opts ...interface{}` and never reads it, so nothing in production implements or calls `InvalidBlockHandler`. Widening the signature costs nothing, but note that the direct `Server.ReportInvalidBlock` path has no dedupe at all, unlike the Kafka consumer. If that handler is ever wired up, the two paths will not behave the same.

## Verification run locally

- `go build ./...` and `go vet` over the four touched trees: clean.
- `go test -race -count=1 ./services/p2p/...`: pass, 13.8s.
- `go test -race -count=1 ./services/blockvalidation/...`: pass, 174s.
- Repo CI on this head is green on all checks, including golangci-lint and the Sonar quality gate.

What I did not do: no smoke or sequential suites locally, and no soak on a live node. The fair-share behaviour is exercised only in memory at caps of 10 and 100, so its cost and its eviction mix under real mainnet announcement rates are unmeasured.

I have not repeated the protoc toolchain drift on `kafka_messages.pb.go` that the bot already raised. I agree with it that the churn is cosmetic.

liam (approved):
Reviewed independently on top of the existing review round, with a focus on the fair-share bookkeeping. Ran the affected suites on the branch.

**Verified:**

- The three-structure bookkeeping (map, global list, per-peer lists) stays consistent across `Store` re-attribution, `Delete`, `DeleteExpired` and `Clear`. The drain loop terminates because each iteration removes exactly one entry.
- `limit/len(m.perPeer)` cannot divide by zero, but only because `pl != nil` short-circuits ahead of it and implies `len(m.perPeer) >= 1`. That is load-bearing and the comment does not mention it: anyone hoisting the divisor into a local or reordering that condition gets a panic on the gossip hot path. Worth a clause on `fairShareVictimLocked`.
- The observability gap on `reportedInvalidBlocks` raised earlier is observability only, not a leak. `recordEvictorLocked` caps `evictors` at `evictorTrackLimit` and decays on overflow, so never calling `EvictionsSinceLastRead` on that map grows nothing.
- `SETTINGS_CONTEXT=test go test -race -tags testtxmetacache ./services/p2p/... ./util/kafka/... ./interfaces/...` green; `go vet -tags testtxmetacache` clean.

Three notes, none blocking.

**1. The fair-share denominator is dynamic, so the residual's bound is slightly optimistic on skewed distributions.**

The residual says a flooder displaces up to one share's worth of other peers' entries. That is exactly right when entries are spread evenly. But the share divides by `len(perPeer)`, meaning peers *currently holding entries*, and global-oldest eviction drains the smallest, oldest holders out of that denominator first, which raises the flooder's own share mid-flood.

Probed on this branch at cap 100:

```
10 peers x 10 entries  -> flooder ends with  9 of 100  (honest lost  9)
 4 peers x 25 entries  -> flooder ends with 20 of 100  (honest lost 20)
 2 peers x 50 entries  -> flooder ends with 33 of 100  (honest lost 33)
50 peers x  2 entries  -> flooder ends with  1 of 100  (honest lost  1)
```

All match limit/(n+1), as documented. Skewed:

```
P1 holds 1 (oldest), P2 holds 99  -> flooder ends with 50 of 100
```

The flooder drains P1's single entry, `len(perPeer)` falls from 3 to 2, and its share rises from 33 to 50. So the real bound is limit divided by the number of peers still holding entries when the flood settles, a denominator the attacker partly controls by draining small holders first. It still converges, still bounds what used to be the whole map, and is harmless for the block path now that provenance travels in the message. It is just worth saying the denominator moves, since a reader will otherwise compute the bound from the peer count at flood start.

**2. Proto field naming.**

`peer_id` and `peer_url` are snake_case in a message whose other fields are camelCase (`blockHash`, `reason`). Harmless on the wire, but protobuf JSON lowerCamelCases the name, so the serialized names are `peerId` / `peerUrl`. Naming them that way in the `.proto` matches the siblings and keeps the doc examples honest. 1668 has the same thing with its `peer_id`.

**3. This PR and 1668 conflict with each other.**

`git merge-tree pr1668 pr1669` reports conflicts in `services/p2p/Server.go` and `util/kafka/kafka_message/kafka_messages.pb.go`. The `.proto` auto-merges cleanly; the generated `.pb.go` does not, because both regenerate the rawDesc bytes. Whoever lands second should rerun `make gen` rather than hand-resolve conflict markers inside the generated file.

On the earlier point about the hash-only dedupe key: I traced the `storeInvalidBlock` AddBlock-failure path behind it and it is genuinely reachable, so that one is worth answering rather than waving through.

Approving.

PR #1674, opened 2026-08-31
Fix dead securityLevelGRPC settings.conf key

icellan:
## Problem

`settings.conf` shipped `securityLevelGRPC = 0`. Nothing reads that key. The loader
reads `security_level_grpc` (`settings/settings.go`), and the struct tag is
`key:"security_level_grpc"` (`settings/interface.go`). The committed line is dead.

It is harmless at its current value because the real default is also `0` - but an
operator who edits it to `1` expecting to enable TLS on inter-service gRPC gets
silent plaintext instead: `util/grpc_helper.go` `GetGRPCClient()` never sees the
change and keeps picking `insecure.NewCredentials()`.

The likely cause: the sibling setting genuinely is camelCase - `SecurityLevelHTTP`
carries `key:"securityLevelHTTP"` (`settings/interface.go`) - so `securityLevelGRPC`
looked correct by analogy. `securityLevelHTTP` itself is untouched by this PR; the
broader camelCase/snake_case inconsistency across the settings package is a
separate, larger question and remains a follow-up.

## Fix

- `settings.conf`: `securityLevelGRPC` -> `security_level_grpc`, value and comments
  unchanged.
- Same wrong spelling also appeared in operator-facing recommendation text, which
  told operators to set the dead key - corrected in `cmd/diagnose/config_checks.go`,
  `util/admin_api_key.go`, `docs/howto/diagnose.md`, and
  `docs/references/services/rpc_reference.md`.
- Searched `deploy/`, `kubernetes/`, `compose/`, and every `.conf`/`.yml`/`.yaml`/`.env`
  file in the repo for other occurrences of the wrong spelling (including
  context-suffixed variants like `securityLevelGRPC.docker`) - found none beyond the
  ones listed above.

## Regression test

Added `TestSettingsConfNoShadowedKeys` (`settings/settings_conf_dead_key_test.go`),
which fails on any `settings.conf` key whose case/underscore-normalized form
collides with a real `key:` tag while the literal key does not match it exactly -
the precise failure mode that let `securityLevelGRPC` through silently, since
gocore's config lookup is exact-case and exact-string.

This is deliberately narrower than a full "every settings.conf key resolves to
something real" sweep. `TestSettingsConfHasNoDeadKeys` (added on `main` separately)
already covers a related but different case - asserting that two specific,
already-removed keys stay removed - and does not attempt the full sweep either,
since most unresolved base keys are legitimate (interpolation `${VAR}` targets,
`startXxx` bootstrap flags, keys read outside the reflection-tagged struct). This
test targets only the mechanically-detectable signature of a copy-paste-by-analogy
typo.

Six existing normalized collisions are exempted in the test with individual
reasoning. Four are genuinely distinct ALL_CAPS interpolation constants
(`P2P_PORT`, `P2P_PORT_COINBASE` - confirmed referenced via `${...}` elsewhere in
settings.conf; `LOG_LEVEL`, `COINBASE_GRPC_ADDRESS` - same naming block, no
confirmed Go reader). The other two look like the same bug class as
`securityLevelGRPC` but are left for the settings owner to triage rather than fixed
here, to keep this change scoped:

- `kafka_unitTest` (settings.conf, builds a full Kafka URL) is never read by any Go
  code; the real setting is the ALL_CAPS `KAFKA_UNITTEST`, a bare topic-name string.
- `coinbase_Store.docker` is a lone context-only override with no base assignment,
  so it cannot resolve to anything regardless; the real key is `coinbase_store`.

## Verification

- `go build ./...`
- `go vet ./settings/... ./util/...`
- `go test ./settings/... -race`
- `go test ./util/ -race` (admin API key + gRPC TLS/security-level tests)
- `golangci-lint run` / `staticcheck` on the touched packages - no new issues (three
  pre-existing `prealloc` warnings confirmed present on `main` before this change)
- End-to-end contrast, run as a standalone binary against isolated `settings.conf`
  fixtures so gocore's config singleton can't carry state between runs:
  - `security_level_grpc = 1` -> `NewSettings().SecurityLevelGRPC == 1`
  - `securityLevelGRPC = 1` (old spelling) -> `NewSettings().SecurityLevelGRPC == 0`
- Reintroduced the misspelling to confirm `TestSettingsConfNoShadowedKeys` fails and
  names the dead key, then reverted.

### DIFF
--- a/cmd/diagnose/config_checks.go
+++ b/cmd/diagnose/config_checks.go
@@ -208,7 +208,7 @@ func checkSecurity(s *settings.Settings) []ConfigResult {
 			Severity:    severity,
 			Check:       "gRPC TLS",
 			Value:       "disabled (level 0)",
-			Recommended: "Set securityLevelGRPC >= 1 for production",
+			Recommended: "Set security_level_grpc >= 1 for production",
 		})
 	} else {
 		results = append(results, ConfigResult{
--- a/docs/howto/diagnose.md
+++ b/docs/howto/diagnose.md
@@ -71,7 +71,7 @@ Checks connectivity to each gRPC service by creating a client and calling its `H
 
 If a service address is empty, the check is skipped. If the service is disabled in settings (e.g. `startValidator=false` for the current context), the skip message says so explicitly:
 
-```
+```text
 Validator gRPC  -  SKIP  -  disabled (startValidator=false)
 ```
 
@@ -123,7 +123,7 @@ These catch subtle problems where services silently fall out of sync:
 
 ### Example Output
 
-```
+```text
 Service Health Checks
 =====================
 SERVICE                  ADDRESS                STATUS  LATENCY  MESSAGE
@@ -175,7 +175,7 @@ Listen addresses checked: Blockchain gRPC/HTTP, Block Assembly gRPC, Block Valid
 
 | Check | Condition | Severity |
 |-------|-----------|----------|
-| gRPC TLS | `securityLevelGRPC = 0` | WARN in prod, INFO in dev |
+| gRPC TLS | `security_level_grpc = 0` | WARN in prod, INFO in dev |
 | HTTP TLS | `securityLevelHTTP = 0` | WARN in prod, INFO in dev |
 | TLS cert files | TLS enabled but `server_certFile` or `server_keyFile` empty | ERROR |
 | gRPC admin API key | Empty or < 16 chars | WARN in prod |
@@ -232,14 +232,14 @@ Checks that the data folder exists, is a directory, and is writable. Catches per
 
 ### Example Output
 
-```
+```text
 Configuration Checks
 ====================
 SEVERITY  CHECK                       VALUE                                       RECOMMENDED
 INFO      Configuration context       dev
 INFO      Network                     mainnet
 OK        Port conflicts              none (16 listen addresses checked)
-INFO      gRPC TLS                    disabled (level 0)                          Set securityLevelGRPC >= 1 for production
+INFO      gRPC TLS                    disabled (level 0)                          Set security_level_grpc >= 1 for production
 INFO      HTTP TLS                    disabled (level 0)                          Set securityLevelHTTP >= 1 for production
 OK        gRPC admin API key          32 chars
 OK        Kafka hosts                 localhost:9092
--- a/docs/references/services/rpc_reference.md
+++ b/docs/references/services/rpc_reference.md
@@ -432,7 +432,7 @@ For GRPC services, certain administrative operations require additional API key 
 - **Usage**: API key must be included in GRPC requests as metadata with the key `x-api-key`.
 
 !!! warning "Security Note"
-    The API key provides administrative access to peer-policy operations. Keep it secret, never commit it, and prefer `securityLevelGRPC >= 2` (verified TLS) when the gRPC listener is not loopback-bound so the key is not exposed in transit. Note that a *configured* key is currently echoed in the node's startup settings dump, so treat startup logs as sensitive and prefer supplying the key via the environment or a secret store.
+    The API key provides administrative access to peer-policy operations. Keep it secret, never commit it, and prefer `security_level_grpc >= 2` (verified TLS) when the gRPC listener is not loopback-bound so the key is not exposed in transit. Note that a *configured* key is currently echoed in the node's startup settings dump, so treat startup logs as sensitive and prefer supplying the key via the environment or a secret store.
 
 ## General Format
 
--- a/settings.conf
+++ b/settings.conf
@@ -1161,7 +1161,7 @@ rpc_pass = bitcoin
 
 rpc_user = bitcoin
 
-securityLevelGRPC = 0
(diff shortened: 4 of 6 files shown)


ctnguyen (approved):
Review at commit [bd6376f](https://github.com/icellan/teranode/commit/bd6376f44f4084af7466d060e42c6f9b3b6fb09e)

The core change is plainly right and unusually well evidenced. `settings/interface.go:59` carries `key:"security_level_grpc"`, `settings/settings.go` reads that string, and gocore's lookup is exact — `os.LookupEnv(key)` then a case-sensitive map probe in `findValue` — so the old `settings.conf` line really was inert, and the "typed by analogy to the genuinely-camelCase `securityLevelHTTP`" diagnosis is convincing. The blast radius is correctly identified too: `git grep` over tracked files shows the old spelling now survives only in the new test's own explanatory comments, and every operator-facing string that told people to set the dead key is caught. The regression test targets the right signature — normalized collision without exact match — rather than reaching for a blanket "every key resolves" sweep it would have had to exempt half of `settings.conf` to pass; the six exemptions are individually reasoned and the two that admit to being the same bug class are honestly labelled as deferred rather than quietly dropped. I traced the two `COINBASE_GRPC_ADDRESS` claims and both halves hold: nothing in Go reads that literal key, and since gocore looks up the *requested* key name, the `docker-compose-3blasters.yml` env var never reaches `coinbase_grpcAddress`. Adding a base assignment equal to the struct default also can't shadow anything — env still wins, and `findValue` walks context suffixes before falling back to base, so the only observable change is the reported source in the settings dump.

Nothing here should hold up the merge. Both notes below are about the neighbourhood the fix lands in: now that the key resolves and the diagnose output names it correctly, operators are much more likely to actually turn gRPC TLS on, and the two places that meet them when they do are a little out of step with the rest of the codebase.

## Non-blocking issues

### ChiR1 — Diagnose cert-file check gates only on HTTP, so a gRPC TLS misconfiguration passes

**Problem:** `cmd/diagnose/config_checks.go:237` guards the cert-file check with `s.SecurityLevelHTTP > 0 && (s.ServerCertFile == "" || s.ServerKeyFile == "")` — `SecurityLevelGRPC` is not consulted. But `util/grpc.go:48-57` returns a `ConfigurationError` (`server_certFile is required for security level %d`) whenever `SecurityLevelGRPC > 0` and the cert or key path is empty, so every gRPC service refuses to start. The docs table this commit touched describes the check generically — `docs/howto/diagnose.md:180` reads `TLS cert files | TLS enabled but server_certFile or server_keyFile empty | ERROR` — which overstates what it covers. This path was always possible to reach, since gocore layers `settings_local.conf` and environment overrides ahead of the struct-tag default; what this commit changes is that the shipped base config now exposes the key and the diagnose recommendation names it correctly, so an operator is far more likely to walk into it.

**Why it matters:** An operator who acts on the gRPC TLS recommendation without cert paths configured gets a green `gRPC TLS` row from diagnose and then a node that will not come up. Catching exactly that before a restart is what the config checks are for.

**Fix:** Include the gRPC level in the condition:

```go
if (s.SecurityLevelHTTP > 0 || s.SecurityLevelGRPC > 0) && (s.ServerCertFile == "" || s.ServerKeyFile == "") {
```

Both server paths read the same `ServerCertFile` / `ServerKeyFile`, so one combined check is sufficient.

### ChiR2 — Diagnose reports level 1 as OK while the same commit warns about it elsewhere

**Problem:** `cmd/diagnose/config_checks.go:213-219` reports any non-zero `SecurityLevelGRPC` as `SeverityOK` with the value `level 1` and no caveat. On that same configuration, `util/admin_api_key.go:76` — changed in this commit — warns that `security_level_grpc=1` "does not provide verified transport security, so the admin key can be harvested in transit". So two parts of the node disagree about whether the operator is in good shape. The recommendation strings show the same split: `config_checks.go:211` now says `Set security_level_grpc >= 1 for production`, while the two other strings this commit touched say `>= 2` (`util/admin_api_key.go:76` and `docs/references/services/rpc_reference.md:435`). Level 1 client-side is `credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})`, which `util/grpc_helper.go:389` and `:426` describe as MITM-vulnerable by design.

**Why it matters:** Level 1 is a legitimate supported mode, so recommending it is not wrong in itself — but a green `OK` on it is, because it is the one signal that tells an operator to stop looking. Having the diagnostic bless a configuration the runtime warns about undercuts both messages, and since this commit is what makes all three strings take effect, it is the natural place to line them up.

**Fix:** Report level 1 at the same severity the runtime uses, and match the other two call sites on the recommended floor:

```go
Recommended: "Set security_level_grpc >= 2 for production (level 1 does not verify certificates)",
```

with the non-zero branch treating level 1 as a warning rather than `SeverityOK`. The sample output at `docs/howto/diagnose.md:242` needs the same edit.

## Recap

| ChiR1 | Cert check ignores gRPC level | | 40% |
| ChiR2 | Level 1 reported as OK | | 30% |

ordishs (approved):

PR #1680, opened 2026-09-01
chore: ignore local .mcp.json MCP server config

ordishs:
A `.mcp.json` in the repo root holds per-developer MCP server endpoints, internal hostnames among them. It is not shared configuration and it is not tracked, so it shows up as an untracked file in every working tree that has one. Ignore it.

No behaviour change; one line in `.gitignore`.

**Verified:** `git check-ignore -v .mcp.json` matches at `.gitignore:124`, and a real `.mcp.json` placed in the tree no longer appears in `git status`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,4 @@ teranode-dev
 .worktrees
 docs/superpowers/
 compose/generated/
+.mcp.json


oskarszoon (approved):
Clean. `.mcp.json` has never been tracked on any ref (`git log --all -- .mcp.json` is empty), so the entry is preventative and needs no `git rm --cached` alongside it — worth confirming, since an ignore added over a tracked file is a silent no-op. Placement and form match the existing local-config block.

liam (approved):
LGTM. One-line ignore, no tracked .mcp.json in the repo.

PR #1681, opened 2026-09-01
fix(utxo): tolerate a dangling spender reference under a recent parent (#1214)

ordishs:
## The wedge

`SequentialSpendAndCreate` is spend-first. It records a transaction's spends on its parents, then creates the transaction's own record:

```go
spends, err = s.Spend(ctx, tx, blockHeight, options.IgnoreFlags)
// ...
md, err := s.Create(ctx, tx, blockHeight, opts...)
```

A crash in that window, or a rollback that exhausts its three `unspendWithRetry` attempts, leaves a parent output slot naming a spender the store has no record of — a dangling spender reference.

`GetCounterConflictingTxHashes` then walks that absent spender. `GetConflictingChildren` fails on its root read, the error propagates out of `checkCounterConflictingOnCurrentChain`, and block validation wedges on a block that SVNode-following peers accept.

Create-first (#1355) would close the window at source, but it is still open and unmerged. #1393 bounded and deduped the walk; it did not touch the absent-record case.

## The guard

An absent record has no `BlockIDs`, so it is definitionally not mined on our chain. Tolerance is gated on the parent's confirmation depth:

- Every parent output slot naming the spender confirmed **within** the retention window, or unmined → no counter mined on those slots could have been pruned yet, so the absence must be a never-created loser. Drop the spender from the counter set and continue.
- Any slot **below** that window → a mined-then-pruned counter cannot be ruled out. Fail closed; SVNode would reject a block double-spending a confirmed output.
- Parent depth unknown, or `retention == 0` → fail closed.

Tolerance is ANDed across every slot naming the spender, not taken from the first seen, so one unprovable slot is enough to reject.

The guard is scoped to `ErrTxNotFound` / `ErrNotFound`; every other walk error still propagates. The tip height is read lazily, only when an absent record is hit, so a store carrying no dangling reference is never asked for it.

Tolerated absences increment `teranode_utxo_dangling_spender_ref_tolerated_total`, so the inconsistency stays visible rather than silent.

## Proof

The end-to-end tests build the dangling reference on a real `sqlitememory` store — `WithSpendOnly()` with no matching `WithCreateOnly()` is exactly the spend-first window — then run `checkCounterConflictingOnCurrentChain`.

With the guard disabled, identical inputs, pre-fix semantics:

```
--- FAIL: TestCheckCounterConflictingOnCurrentChain_ToleratesDanglingRefUnderRecentParent
--- PASS: TestCheckCounterConflictingOnCurrentChain_FailsClosedOnDanglingRefUnderBuriedParent
```

With the guard:

```
--- PASS: TestCheckCounterConflictingOnCurrentChain_ToleratesDanglingRefUnderRecentParent
--- PASS: TestCheckCounterConflictingOnCurrentChain_FailsClosedOnDanglingRefUnderBuriedParent
```

The fail-closed case passes in **both** configurations: the guard does not weaken the consensus check, it only narrows what wedges.

## Verification

```
go test ./stores/utxo/... ./services/subtreevalidation/...   14 packages ok, exit 0
go test -race ./stores/utxo/ ./services/subtreevalidation/ ./stores/utxo/sql/...   ok, exit 0
go vet ./...                                                 exit 0
golangci-lint run stores/utxo/... services/subtreevalidation/...   exit 0
```

The `go vet` and lint output that remains is pre-existing, in `test/utils` and older test files this PR does not touch.

## Prune horizon — checked

The tolerance is sound only if the store never deletes a transaction mined on the longest chain before `mined_height + retention`. That assumption is now verified rather than assumed, and `stores/utxo/sql/prune_horizon_test.go` pins it against the real pruner service.

The DAH arithmetic agrees across both backends: the Aerospike Go expression (`set_mined_expressions.go`) and the Lua UDF (`teranode.lua:988`) both compute `currentBlockHeight + blockHeightRetention`, and SQL's `SetMinedMulti` uses `minedBlockInfo.BlockHeight + retention` (`sql.go:3346`). The pruner deletes on `delete_at_height <= tip`. The spend path stamps `tip + 1 + retention`, which is strictly later, so the earliest stamp a mined transaction can carry comes from `SetMinedMulti`.

Tests, on a real store with the real pruner:

- mined at 1000, outputs already spent → **survives** pruning at `1000 + retention - 1`
- same → **deleted** at `1000 + retention`
- flagged conflicting at 1000 (stamped `1000 + retention`), then mined at 1500 → `SetMinedMulti` **bumps the stale stamp forward**, and it survives to `1500 + retention - 1`

The boundary is exact, with no slack: the guard tolerates while `parent_height + retention > tip`, and a counter mined at `h >= parent_height` survives while `h + retention > tip`.

## Residual risk

**Aerospike does not bump a stale conflicting stamp.** `teranode.lua` returns early in its conflicting branch (`if rec[BIN_CONFLICTING] then ... return "", nil`), and the Go expression yields `ExpUnknown` for a conflicting record that already has a DAH. SQL bumps it forward; Aerospike does not. For that to shorten the horizon a record would have to be conflicting-flagged *and* mined on the longest chain at once, which is contradictory in the node's own model — conflicting means "we consider this a loser". I could not construct the state (the store refuses to spend a conflicting transaction's outputs with `TX_CONFLICTING`), but I did not prove it unreachable, and I did not write the Aerospike-side equivalent of these tests.

**Unrelated pre-existing bug found while testing.** `Store.SetConflicting` deadlocks on SQLite: it opens a transaction at `sql.go:4325`, then calls `s.GetSpend` on the pool at `sql.go:4383`, which cannot get a connection until that transaction commits. Reproduced as a hang until the 600s test timeout. Every existing `sql_test.go` call passes an empty hash slice, so no current test reaches it. Not touched by this PR — the horizon test writes the conflicting flag with direct SQL to route around it. Worth its own issue.

### DIFF
--- a/services/subtreevalidation/SubtreeValidation.go
+++ b/services/subtreevalidation/SubtreeValidation.go
@@ -473,7 +473,7 @@ func (u *Server) blessMissingTransaction(ctx context.Context, blockHash chainhas
 func (u *Server) checkCounterConflictingOnCurrentChain(ctx context.Context, txHash chainhash.Hash, blockIds map[uint32]bool) error {
 	// the tx is conflicting, check whether the counter-conflicting transactions have already been mined on our chain
 	// first get the parent transactions and check if they were spent
-	counterConflictingTxHashes, err := utxo.GetCounterConflictingTxHashes(ctx, u.utxoStore, txHash, u.settings.UtxoStore.ConflictingChildrenMaxNodes)
+	counterConflictingTxHashes, err := utxo.GetCounterConflictingTxHashes(ctx, u.utxoStore, txHash, u.settings.UtxoStore.ConflictingChildrenMaxNodes, u.settings.GetUtxoStoreBlockHeightRetention())
 	if err != nil {
 		return errors.NewProcessingError("[checkCounterConflictingOnCurrentChain][%s] failed to get counter conflicting tx hashes", txHash.String(), err)
 	}
--- a/stores/utxo/process_conflicting.go
+++ b/stores/utxo/process_conflicting.go
@@ -55,6 +55,20 @@ var (
 		},
 	)
 
+	// prometheusDanglingSpenderRefTolerated counts absent counter-conflicting records
+	// that the parent-depth guard in GetCounterConflictingTxHashes tolerated (excluded
+	// from the counter set) instead of hard-erroring. A non-zero value means the store
+	// carries dangling spender references (a never-created loser, #1214) that would
+	// otherwise have wedged block validation — surface it so the inconsistency is visible.
+	prometheusDanglingSpenderRefTolerated = promauto.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: "teranode",
+			Subsystem: "utxo",
+			Name:      "dangling_spender_ref_tolerated_total",
+			Help:      "Number of absent counter-conflicting records tolerated by the parent-depth guard in the counter-conflicting walk",
+		},
+	)
+
 	// prometheusUtxoConflictingWalkDuration replaces the store-method duration
 	// histogram (e.g. aerospike txmeta_get_conflicting) for the walks that
 	// GetCounterConflictingTxHashes now runs via the package-level function —
@@ -1170,7 +1184,12 @@ func GetConflictingChildren(ctx context.Context, s Store, hash chainhash.Hash, m
 // counter-conflicting transaction) and that spender's full descendant set.
 // maxNodes bounds each descendant walk (see GetConflictingChildren); <= 0
 // means unbounded.
-func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhash.Hash, maxNodes int) ([]chainhash.Hash, error) {
+//
+// retention is the UTXO store block-height retention window. It gates the
+// parent-depth guard that tolerates an absent counter-conflicting record (see
+// parentDepthInfo); pass 0 to disable the guard and fail closed on every absent
+// record.
+func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhash.Hash, maxNodes int, retention uint32) ([]chainhash.Hash, error) {
 	ctx, _, deferFn := tracing.Tracer("utxo").Start(ctx, "GetCounterConflictingTxHashes")
 
 	defer deferFn()
@@ -1191,10 +1210,18 @@ func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhas
 		parentTxs[*input.PreviousTxIDChainHash()] = nil
 	}
 
+	// parentDepth records, per parent, the confirmation depth used to gate tolerance
+	// of an absent counter-conflicting record (see the absent-counter branch below).
+	// Captured from the parent record we already fetch, so no extra round trip.
+	parentDepth := make(map[chainhash.Hash]parentDepthInfo, len(parentTxs))
+
 	for parentTx := range parentTxs {
 		parentTxHash := &parentTx
 
-		parentTxMeta, err := s.Get(ctx, parentTxHash, fields.Utxos)
+		// fields.BlockIDs is requested alongside fields.BlockHeights because the SQL
+		// backend only populates BlockHeights when the block-id join is loaded; the
+		// parent-depth guard below relies on BlockHeights being present for a mined parent.
+		parentTxMeta, err := s.Get(ctx, parentTxHash, fields.Utxos, fields.BlockHeights, fields.BlockIDs, fields.UnminedSince)
 		if err != nil {
 			return nil, err
 		}
@@ -1210,6 +1237,7 @@ func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhas
 		}
 
 		parentTxs[*parentTxHash] = spendingTxIDs
+		parentDepth[*parentTxHash] = newParentDepthInfo(parentTxMeta)
 	}
 
 	// validate every input and collect the unique counter-spenders in first-seen
@@ -1220,8 +1248,15 @@ func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhas
 	seenSpenders := make(map[chainhash.Hash]struct{}, len(txMeta.Tx.Inputs))
 	uniqueSpendingTxIDs := make([]chainhash.Hash, 0, len(txMeta.Tx.Inputs))
 
+	// spenderParents records every parent output slot that names a given spender.
+	// The absent-record guard below needs all of them: one slot it cannot prove
+	// recent is enough to fail closed for that spender.
+	spenderParents := make(map[chainhash.Hash][]chainhash.Hash, len(txMeta.Tx.Inputs))
+
 	for _, input := range txMeta.Tx.Inputs {
-		parenTxIDS, ok := parentTxs[*input.PreviousTxIDChainHash()]
+		parentHash := *input.PreviousTxIDChainHash()
+
+		parenTxIDS, ok := parentTxs[parentHash]
 		if ok {
 			// check the length of the spending txs, if it's less than the index, then the input is not spent
 			if len(parenTxIDS) <= int(input.PreviousTxOutIndex) {
@@ -1231,7 +1266,7 @@ func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhas
 
 			spendingTxID := parenTxIDS[input.PreviousTxOutIndex]
 			if spendingTxID != nil {
-				counterConflictingMap[*spendingTxID] = struct{}{}
+				spenderParents[*spendingTxID] = append(spenderParents[*spendingTxID], parentHash)
 
 				if _, ok := seenSpenders[*spendingTxID]; !ok {
 					seenSpenders[*spendingTxID] = struct{}{}
@@ -1241,13 +1276,53 @@ func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhas
 		}
 	}
 
+	// tipHeight is resolved lazily; nil means "not yet read from the store"
+	var tipHeight *uint32
+
 	for _, spendingTxID := range uniqueSpendingTxIDs {
 		// call the package-level walk directly (not the Store method) so the
 		// caller-chosen maxNodes budget flows into the BFS
 		childHashes, err := GetConflictingChildren(ctx, s, spendingTxID, maxNodes)
 		if err != nil {
+			// A counter-conflicting record that is absent from the store has no
+			// BlockIDs, so it is definitionally not mined on our chain. Tolerate it
+			// (exclude from the counter set) only when every parent output slot it
+			// occupies is confirmed within the retention window of tip: there, no
+			// counter mined on that slot could yet have been pruned, so the absence
+			// must be a never-created loser — a spend recorded on the parent whose
+			// own record was never written (#1214). The block is valid; legacy
+			// SVNode-following peers accept it. Below that window we cannot rule out
+			// a mined-then-pruned counter, so we fail closed: SVNode would reject a
+			// block double-spending a confirmed output.
+			if errors.Is(err, errors.ErrTxNotFound) || errors.Is(err, errors.ErrNotFound) {
+				// the tip height is read only on this path, so a store that never
+				// carries a dangling reference is never asked for it
+				if tipHeight == nil {
+					h := s.GetBlockHeight()
+					tipHeight = &h
+				}
+
+				tolerable := true
+
+				for _, parentHash := range spenderParents[spendingTxID] {
+					if !parentDepth[parentHash].withinRetention(*tipHeight, retention) {
+						tolerable = false
+						break
+					}
+				}
+
+				if tolerable && len(spenderParents[spendingTxID]) > 0 {
+					prometheusDanglingSpenderRefTolerated.Inc()
+					continue
+				}
+			}
+
 			return nil, err
 		}
+
+		// admitted only once its record was readable; a tolerated absent spender is
+		// deliberately left out of the counter set
+		counterConflictingMap[spendingTxID] = struct{}{}
 
 		for _, childHash := range childHashes {
 			if childHash.Equal(subtree.FrozenBytesTxHash) {
@@ -1267,4 +1342,67 @@ func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhas
 	// fmt.Printf("counterConflicting: %v\n", counterConflicting)
 
 	return counterConflicting, nil
+}
+
+// parentDepthInfo captures a parent tx's confirmation depth relative to the
+// pruning horizon, so the counter-conflicting walk can decide whether an absent
+// spender of that parent is provably a never-created loser (safe to tolerate) or
+// a possibly mined-then-pruned counter (must fail closed).
+type parentDepthInfo struct {
+	// unmined is true when the parent carries no mined height (UnminedSince set),
+	// so it is trivially at the top of the chain — no mined spender of it could
+	// yet have been pruned.
+	unmined bool
+	// minHeight is the lowest block height the parent is mined at (valid only when
+	// mined is true).
+	minHeight uint32
+	// mined is true when a concrete mined height is known for the parent.
+	mined bool
+}
+
+func newParentDepthInfo(parent *meta.Data) parentDepthInfo {
+	if parent == nil {
+		return parentDepthInfo{}
+	}
+
+	if parent.UnminedSince != 0 {
+		return parentDepthInfo{unmined: true}
+	}
+
+	if len(parent.BlockHeights) == 0 {
+		return parentDepthInfo{}
+	}
+
+	minHeight := parent.BlockHeights[0]
+	for _, h := range parent.BlockHeights[1:] {
+		if h < minHeight {
+			minHeight = h
+		}
+	}
+
+	return parentDepthInfo{minHeight: minHeight, mined: true}
+}
+
+// withinRetention reports whether the parent is confirmed within retention blocks
+// of tipHeight (or is unmined). When true, no counter-conflicting tx mined on the
+// parent's output slot could yet be prune-eligible — a mined spender's delete-at-height
+// is mined_height + retention, and mined_height >= parent_height — so an absent spender
+// must be a never-created loser. When we cannot prove recency (no mined height known
+// and not flagged unmined), or retention is 0, we return false and fail closed.
+func (d parentDepthInfo) withinRetention(tipHeight, retention uint32) bool {
+	if retention == 0 {
+		return false
+	}
+
+	if d.unmined {
+		return true
+	}
+
+	if !d.mined {
+		return false
+	}
+
+	// Equivalent to minHeight > tipHeight - retention, written as addition to
+	// avoid unsigned underflow when tipHeight < retention.
+	return d.minHeight+retention > tipHeight
 }
--- a/stores/utxo/sql/prune_horizon_test.go
+++ b/stores/utxo/sql/prune_horizon_test.go
@@ -1,0 +1,211 @@
+package sql
+
+import (
+	"context"
+	"testing"
+
+	"github.com/bsv-blockchain/go-bt/v2"
+	"github.com/bsv-blockchain/go-bt/v2/chainhash"
+	utxostore "github.com/bsv-blockchain/teranode/stores/utxo"
+	"github.com/bsv-blockchain/teranode/stores/utxo/pruner"
+	"github.com/bsv-blockchain/teranode/stores/utxo/tests"
+	"github.com/stretchr/testify/require"
+)
+
+// The parent-depth guard in utxo.GetCounterConflictingTxHashes tolerates an absent
+// counter-conflicting record when the parent output slot it occupied is confirmed
+// within `retention` blocks of tip. That is only sound if the store never deletes a
+// transaction mined on the longest chain before mined_height + retention: a counter
+// mined on the parent's slot has mined_height >= parent_height, so its deletion
+// height is at or beyond the parent's.
+//
+// These tests pin that horizon at its exact boundary, on a real store driven by the
+// real pruner service. The earliest DAH the code can produce for a mined transaction
+// comes from SetMinedMulti (sql.go: newDAH = minedBlockInfo.BlockHeight + retention);
+// the spend path stamps tip+1+retention, which is strictly later. So a transaction
+// mined with all outputs already spent is the worst case, and it is what is built here.
+func buildMinedFullySpent(ctx context.Context, t *testing.T, store *Store, tx *bt.Tx, mineHeight uint32) *chainhash.Hash {
+	t.Helper()
+
+	_, _, err := store.SpendAndCreate(ctx, tx, mineHeight, utxostore.WithCreateOnly())
+	require.NoError(t, err)
+
+	txHash := tx.TxIDChainHash()
+
+	// Spend every output BEFORE the mined transition, so SetMinedMulti sees an
+	// all-spent record and stamps the earliest DAH the store can produce.
+	for i, out := range tx.Outputs {
+		spendTx := bt.NewTx()
+		require.NoError(t, spendTx.From(txHash.String(), uint32(i), out.LockingScript.String(), out.Satoshis))
+		require.NoError(t, spendTx.PayToAddress("1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa", 1000))
+
+		_, _, err = store.SpendAndCreate(ctx, spendTx, mineHeight, utxostore.WithSpendOnly())
+		require.NoError(t, err, "spending output %d", i)
+	}
+
+	_, err = store.SetMinedMulti(ctx, []*chainhash.Hash{txHash}, utxostore.MinedBlockInfo{
+		BlockID:        100,
+		BlockHeight:    mineHeight,
+		SubtreeIdx:     0,
+		OnLongestChain: true,
+	})
+	require.NoError(t, err)
+
+	_, err = store.Get(ctx, txHash)
+	require.NoError(t, err, "setup: tx must exist before pruning")
+
+	return txHash
+}
+
+func horizonStore(ctx context.Context, t *testing.T) (*Store, pruner.Service, uint32) {
+	t.Helper()
+
+	ResetPrunerServiceForTests()
+	t.Cleanup(ResetPrunerServiceForTests)
+
+	store, _ := setup(ctx, t)
+
+	provider, ok := any(store).(pruner.PrunerServiceProvider)
+	require.True(t, ok)
+
+	prunerSvc, err := provider.GetPrunerService()
+	require.NoError(t, err)
+	prunerSvc.Start(ctx)
+
+	retention := store.settings.GetUtxoStoreBlockHeightRetention()
+	require.NotZero(t, retention, "retention must be configured for this test to mean anything")
+
+	// tests.Tx spends tests.ParentTx; the parent must exist for SetConflicting to
+	// record conflicting_children against it.
+	_, _, err = store.SpendAndCreate(ctx, tests.ParentTx, 1, utxostore.WithCreateOnly())
+	require.NoError(t, err)
+
+	return store, prunerSvc, retention
+}
+
+// One block BEFORE the horizon the guard relies on, the mined transaction must
+// still be present. If this fails, the guard can tolerate an absent counter that
+// was in fact mined and pruned, and the tolerance is unsound.
+func TestPruneHorizon_MinedTxSurvivesUntilMinedHeightPlusRetention(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	store, prunerSvc, retention := horizonStore(ctx, t)
+
+	const mineHeight uint32 = 1000
+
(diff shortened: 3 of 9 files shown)


icellan (changes requested):
The wedge is real and the shape of the guard (depth-gated, ANDed across slots, scoped to not-found) is defensible. But the safety proof — "an absent record has no BlockIDs, so it is definitionally not mined on our chain" — holds only for a miss on the walk **root**, and the code applies it to a not-found from anywhere in `GetConflictingChildren`'s BFS cone. That turns a liveness fix into a consensus fail-open. Three further paths reach the same fail-open, the tolerance is also armed on the `ProcessConflicting` demotion path where it converts a clean abort into a half-committed mutation, and both end-to-end tests pass an empty `blockIds` map, so the branch that actually decides accept/reject is unreachable in them — every finding below passes the new tests unchanged.

## Blocking

- `stores/utxo/process_conflicting.go:1297` — root and descendant not-found are indistinguishable. `GetConflictingChildren` reads the root and every descendant through the same `s.Get(...)` (`:1126`) and returns the raw error. Parent P mined at 900, tip 1000, retention 288; P:0 names counter X whose record **exists** with `BlockIDs=[7]`; X names a grandchild Y that is absent. The walk 404s on Y, the guard checks only P's depth and `continue`s, X never enters `counterConflictingMap`, so `checkCounterConflictingOnCurrentChain` (`SubtreeValidation.go:527-534`) never matches `blockIds[7]` and the node accepts a block double-spending an output already spent by a confirmed tx. Reproduced independently twice. Only tolerate when the **root** read is the not-found.

- `process_conflicting.go:1407` — `withinRetention` has no zero/lagging-tip guard, and a low tip is the permissive direction: `minHeight+retention > 0` is true for every mined parent. Reproduced with the PR's own buried case (parent at height 10, `GetBlockHeight()==0`) — tolerated, `err == nil`. The tip is published only by `stores/utxo/factory/utxo.go:139-157`, which logs-and-skips on error or 0, never starts with `startBlockchain=false`, and refreshes only on block notifications, so it lags during catchup — and lagging low widens the window. `set_mined_expressions.go:301-305` already stopped trusting this cached value for DAH stamping for the same reason. Minimum: `if tipHeight == 0 { return false }`.

- `process_conflicting.go:1368` — `newParentDepthInfo` returns `unmined:true` on `UnminedSince != 0` before looking at `BlockHeights`, and unmined tolerates unconditionally at any depth. A record legitimately carries both: Aerospike `MarkTransactionsOnLongestChain(false)` writes only the unminedSince bin (`longest_chain.go:62`) and never clears blockHeights; SQL does the same (`sql.go:4590-4598`). After any reorg or fork promotion a parent has `BlockHeights=[100]` and `UnminedSince=95` at tip 1000, and the guard declares it top-of-chain and tolerates a counter mined on that buried slot and pruned 600 blocks ago. Prefer the mined height whenever `len(BlockHeights) > 0`.

- `process_conflicting.go:1325` — tolerance is also armed on the demotion path (`aerospike/conflicting.go:28`, `sql/sql.go:4260`), where dropping the loser does not avoid the wedge, it relocates it after two committed mutation steps. `MarkConflictingRecursively` derives affected spends from W's own inputs; `Unspend` matches on `AND spending_data = $3` (`sql.go:3014-3028`) so the slot holding X is a no-op; `SpendAndCreate(W, WithSpendOnly, ...)` then hits `sql.go:2255-2264` "already spent by a different transaction" and returns `UtxoSpentError` — there is no ignore-spent flag (`Interface.go:187-194`). Pre-PR that input aborted cleanly with zero mutation; now it fires the compensating rollback and, on rollback failure, "MANUAL INTERVENTION REQUIRED" (`:213`) — on the path whose own comment says a failure there wedges block assembly on the block forever. No test covers `ProcessConflicting` with the new retention argument.

## Should fix before merge

- `process_conflicting.go:1328` — the tolerate `continue` skips the frozen-sentinel scan over the spender's cone, and since #1393 this loop is the only place counter-spender cones are frozen-checked (`SubtreeValidation.go:513-518` states that invariant explicitly). Combined with the first finding the cone is partially walkable and can hold an alert-system frozen sentinel, so a block spending a frozen UTXO is accepted.

- `process_conflicting.go:1297` — `errors.Is` on teranode errors matches a code anywhere in the wrap chain (`errors/errors.go:180-200`), so a `StorageError` wrapping a blob `NotFound` — exactly what `aerospike/get.go:1938-1941` returns when a large tx's external blob is unreadable or pruned — is classified as a never-created loser and tolerated. That is a data-availability fault swallowed as absence. `DoesNotTolerateNonNotFoundWalkError` uses a bare `ProcessingError` that wraps nothing, so it does not cover this.

- `process_conflicting.go:1392` — the premise "a mined spender's DAH is `mined_height + retention`" is false for conflicting-marked records, which is the normal state for cone members after `MarkConflictingRecursively`. `teranode.lua:990-1001` sets DAH once for conflicting records and never raises it, `buildDeleteAtHeightExpression` mirrors that guard (`set_mined_expressions.go:184-187`), and SQL `SetConflicting` stamps from the cached tip (`sql.go:4280`). A record marked while the cache reads 500 against a real tip of 1200 gets DAH 789 and is prune-eligible immediately — a record mined on the current chain and pruned inside the parent's window, which is what the guard claims cannot exist. This is the residual risk the description flags; the audit fails.

- `process_conflicting.go:1297` — `GetConflictingChildren` fans a BFS level over `conflictingWalkFanOut` concurrent Gets and errgroup keeps only the first error, so whether the walk surfaces as NotFound (tolerated) or a storage error (propagated) is a scheduling race when a level mixes a transient Aerospike failure with a genuine dangling ref. Same store state, different accept/reject across runs, and the racy outcome is the fail-open one.

- `services/subtreevalidation/counter_conflicting_dangling_test.go:103` — both end-to-end tests pass `map[uint32]bool{}`, so the branch that returns `ErrTxInvalid` (`SubtreeValidation.go:530-536`) can never fire: the tolerate test asserts only that the walk did not error, the fail-closed test only that it did. Missing cases: a counter whose record is present and mined in a block inside `blockIds` (must reject); a not-found originating from a descendant rather than the spender (must reject); `ProcessConflicting` over a dangling slot. Also `util/test/helpers.go:28` sets `GlobalBlockHeightRetention = 10`, so these tests run a 10-block window while the chosen heights and comments read as though 288 applied, and the retention actually in force is never asserted.

- `process_conflicting.go:1224` — adding `fields.BlockHeights` to the parent Get introduces a new hard failure on Aerospike: `processBlockHeights` returns a `StorageError` when the bin is absent (`get.go:1195-1199`), where `processBlockIDs` returns empty and nil for the same condition (`get.go:1149-1153`). A parent record from an older node version or a snapshot restore previously read fine under `fields.Utxos` alone and now fails `ERR_STORAGE`, which is not in the tolerated class, so the block wedges. A change made to enable a fail-open guard adds a new fail-closed wedge on the production backend, and both new test files are sqlitememory/mock only.

- `stores/utxo/sql/sql.go:4260` — unguarded `s.settings.GetUtxoStoreBlockHeightRetention()`. `sql.CreateMockStore` (`sql/mock.go:155-161`) builds `&Store{db, logger}` with nil settings, and `sql.go:3337-3339` already guards the identical call. Unreached today, so CI stays green over a live nil-deref path.

Also worth a look, non-blocking: `retention == 0` is documented as the off switch but cannot be set in production without disabling UTXO pruning cluster-wide, and is semantically inverted (in that deployment every absent record really is a never-created dangling ref, yet the guard rejects all of them); the comment at `:1221` justifying the extra `fields.BlockIDs` is wrong — `needsBlockIDsQuery` (`sql/sql.go:1784-1788`) already returns true for `BlockHeights` alone, and on Aerospike `addAbstractedBins` (`get.go:703-711`) drags `SubtreeIdxs` off the wire on every parent read; `spenderParents` (`:1269`) appends one hash per input with no dedup, so a large fan-in tx builds megabytes of duplicates and repeats the same map lookup, paid on every call including the common no-absence case.

stores/utxo/process_conflicting.go
```diff
@@ -1268,3 +1343,66 @@ func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhas
...
+		return true
+	}
+
+	if !d.mined {
+		return false
+	}
+
+	// Equivalent to minHeight > tipHeight - retention, written as addition to
+	// avoid unsigned underflow when tipHeight < retention.
+	return d.minHeight+retention > tipHeight
```
freemans13, line 1407: This is where the shelf life comes from. `tipHeight` is the validating node's current tip, so the accept-or-reject answer for one unchanged store state flips as the chain grows. Measured on this commit with the parent mined at 900 and retention 288: tolerated at tip 1000 and 1187, rejected at 1188 and 1200.

Because nothing clears the dangling reference from the parent's output slot, a parent that validates fine today simply stops validating once the tip passes 1188, and stays that way.

stores/utxo/process_conflicting.go
```diff
@@ -1191,10 +1210,18 @@ func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhas
...
+	parentDepth := make(map[chainhash.Hash]parentDepthInfo, len(parentTxs))
+
 	for parentTx := range parentTxs {
 		parentTxHash := &parentTx
 
-		parentTxMeta, err := s.Get(ctx, parentTxHash, fields.Utxos)
+		// fields.BlockIDs is requested alongside fields.BlockHeights because the SQL
+		// backend only populates BlockHeights when the block-id join is loaded; the
+		// parent-depth guard below relies on BlockHeights being present for a mined parent.
+		parentTxMeta, err := s.Get(ctx, parentTxHash, fields.Utxos, fields.BlockHeights, fields.BlockIDs, fields.UnminedSince)
```
freemans13, line 1224: Worth noting where the recurrence actually lands. A parent whose outputs are all spent gets a delete-at-height of `mined_height + retention`, the same boundary `withinRetention` uses below, so the pruner deletes the parent at the exact moment tolerance expires.

This read then returns TX_NOT_FOUND and line 1226 returns it raw, with no guard on it at all. That is unchanged from before the PR, but it means the fail-closed branch further down is not the path the recurrence takes.

freemans13 (commented):
Not ready to merge. I worked this independently of the review already sitting on it, so I will skip everything icellan raised and cover two things instead: what I could actually reproduce of the central objection, and one consequence I have not seen raised anywhere.

## The root-versus-descendant confusion is real, and here is a repro

icellan's first blocking point is that `GetConflictingChildren` reads the spender and every one of its descendants through the same `s.Get` call and hands back the raw error, so the guard cannot tell "this spender was never created" from "something three levels down is missing". I did not take that on trust. I wrote a throwaway test against `MockUtxostore` on this commit:

- parent P mined at height 900, node tip 1000, retention 288, so the guard is armed
- P's output 0 names spender X
- X's own record **reads fine** and carries `BlockIDs = [7]`
- X names a child Y, and Y is absent

Result:

```
err = <nil>
result = [27ad8841f3b101a25986fef816a5c72425a5a16c0bff8cc1f0d687e7db8b37a3]   // the tx itself, nothing else
spenderX (mined in block 7) present in counter set? false
```

X is a transaction that exists and is mined in block 7, and the walk dropped it. `checkCounterConflictingOnCurrentChain` never fetches its meta, so the `blockIds[7]` test at `SubtreeValidation.go:530` never runs, and the block is accepted. That is a double-spend getting through, not a theoretical hole. Confirmed.

## The guard has a shelf life, and nothing repairs the reference

This is the part I have not seen raised. Tolerance is computed against the node's current tip, so one unchanged store state gives different answers as the chain grows. Parent mined at 900, retention 288, measured on this commit:

```
tip=1000   ->  ACCEPT (tolerated)
tip=1187   ->  ACCEPT (tolerated)
tip=1188   ->  REJECT
tip=1200   ->  REJECT
```

Nothing here removes the dangling reference from the parent's output slot. The PR counts it in `dangling_spender_ref_tolerated_total` and moves on. So the reference is still sitting there at height 1188, and every later conflicting transaction touching that parent hits the same absent record, now outside the window, and block validation stops dead again. The fix buys `retention` blocks. At 288 on mainnet that is roughly two days.

Past that point the failure does not even reach the guard. Once a parent's outputs are all spent it gets a delete-at-height of `mined_height + retention`, which is the same boundary `withinRetention` uses, so the pruner deletes the parent at the moment tolerance expires. The next walk then fails at the parent read, and that read has no guard on it:

```
parent itself pruned -> err = TX_NOT_FOUND (30): parent pruned (guard never consulted)
```

That path is unchanged from before this PR, so it is not a regression. It does mean the fail-closed branch the whole design is built around is not where the recurrence lands.

I would rather see the spend-first window closed at source (#1355), or a repair that clears the orphaned slot, than a time-limited tolerance that hides the inconsistency until it runs out.

## What I did not check

I ran only the two packages this PR touches, `./stores/utxo/` and `./services/subtreevalidation/`, both green on 8bd866a, plus the two throwaway probes above, which I have deleted. No Aerospike tests and no long tests, so icellan's Aerospike-specific claims about the demotion path are unverified by me. I did read `processBlockHeights` at `stores/utxo/aerospike/get.go:1195` and it does return a `StorageError` where `processBlockIDs` returns empty and nil for the same missing bin, so that asymmetry is at least real.

I have not run any of this on a node.

Posting as a comment rather than a second changes-requested, since icellan's block already stands on this commit.

stores/utxo/sql/prune_horizon_test.go
```diff
@@ -0,0 +1,211 @@
...
+		BlockID:        200,
+		BlockHeight:    mineHeight,
+		SubtreeIdx:     0,
+		OnLongestChain: true,
+	})
+	require.NoError(t, err)
+
+	require.NoError(t, store.db.QueryRowContext(ctx,
+		`SELECT delete_at_height FROM transactions WHERE hash = $1`, txHash[:]).Scan(&stamped))
+	require.GreaterOrEqual(t, stamped, int64(mineHeight+retention),
```
freemans13, line 195: This holds for the SQL store. It does not hold for Aerospike, and the Lua branch named two lines up is where it stops holding.

The mined path calls `setDeleteAtHeight(rec, blockHeight, blockHeightRetention)` at `teranode.lua:663`, and that function opens with:

```lua
if rec[BIN_CONFLICTING] then
    if not existingDeleteAtHeight then
        rec[BIN_DELETE_AT_HEIGHT] = newDeleteHeight
        ...
    end
    return "", nil
end
```

It writes a stamp only when there is none, then returns. No bump. So the bump this test proves, and that the guard depends on, does not happen on the production backend for exactly the records at issue. The Aerospike native-ops job on this run reports `skipping`, so nothing here would catch it.

stores/utxo/sql/prune_horizon_test.go
```diff
@@ -0,0 +1,211 @@
...
+// If it is later mined on the longest chain at a greater height, that earlier
+// stamp must not survive — otherwise a mined transaction becomes deletable
+// before mined_height + retention and the parent-depth guard's assumption breaks.
+//
+// The conflicting flag and its early stamp are written with direct SQL rather
+// than through Store.SetConflicting: that method opens a transaction at
+// sql.go:4325 and then calls s.GetSpend on the pool at sql.go:4383, which
+// deadlocks on SQLite. What is under test here is the DAH arithmetic in
+// SetMinedMulti, and this sets up its precondition exactly.
+func TestPruneHorizon_ConflictingThenMinedDoesNotKeepEarlierStamp(t *testing.T) {
```
freemans13, line 147: All three tests flag conflicting before mining. The reverse order is the one that bites, and it is uncovered.

`SetConflicting` stamps `GetBlockHeight() + 1 + retention` from the store's cached tip, which lags (published on block notifications only, and 0 when the store starts with `startBlockchain` false). Probe against a real store and the real pruner, retention 10:

```
after SetMinedMulti(height=1500, unspent outputs): delete_at_height = <nil>
after conflicting flag with cached tip 500:        delete_at_height = 511
PRUNED at tip 515, which is 995 blocks BEFORE mined_height+retention (1510)
```

Outputs left unspent on purpose: that is what keeps `delete_at_height` null, and null is the only case where the `COALESCE(delete_at_height, $3)` in `SetConflicting` writes anything. `SetMinedMulti` has already run by then, so nothing bumps it back.

freemans13 (commented):
Still not mergeable, and the update does not move it. The new commit is test-only. `git diff --name-only 8bd866aa f79e2ce9` lists one file, `stores/utxo/sql/prune_horizon_test.go`, so every blocking finding on 8bd866aa stands untouched, including the root-versus-descendant fail-open that icellan, the review bot and I have now each found separately. The bot's note on that one says it could not execute anything and asks for "a targeted case (mined root + absent descendant)". That case is already on this PR, with output, in my 3 September review.

What the new commit does address is the residual risk paragraph in the description, the claim that the pruner never deletes a mined transaction before `mined_height + retention`. I ran the three tests on f79e2ce9 and they pass. They do not settle the question, for two reasons.

## The audit is SQL-only, and its result inverts on Aerospike

The third test proves that `SetMinedMulti` bumps a stale flag-height stamp forward, so a transaction flagged conflicting and then mined keeps a deletion height at or beyond `mined_height + retention`. That is true of the SQL store. It is not true of Aerospike, which is the production backend, and the test comment itself points at the Lua branch where it breaks.

The mined path calls `setDeleteAtHeight(rec, blockHeight, blockHeightRetention)` at `teranode.lua:663`. The first thing that function does with a conflicting record is this:

```lua
if rec[BIN_CONFLICTING] then
    if not existingDeleteAtHeight then
        rec[BIN_DELETE_AT_HEIGHT] = newDeleteHeight
        ...
    end
    return "", nil
end
```

It sets the stamp only when there is no stamp, then returns unconditionally. There is no bump. So the exact behaviour the third test relies on to close the hazard is absent on the backend that matters, and the tests cannot catch it because they run against SQL. Worth noting the Aerospike native-ops job on this run reports `skipping`.

## Mined first, then flagged conflicting, is not covered

The three tests all flag conflicting before mining. The reverse order behaves differently on SQL too, because `SetConflicting` stamps `GetBlockHeight() + 1 + retention` from the store's cached tip, and that cache lags. It is published on block notifications only and sits at 0 when the store starts with `startBlockchain` false.

Probe on f79e2ce9, retention 10, real store, real pruner service:

```
after SetMinedMulti(height=1500, unspent outputs): delete_at_height = <nil>
after conflicting flag with cached tip 500:        delete_at_height = 511
the guard assumes this tx survives to mined_height + retention = 1510
PRUNED at tip 515, which is 995 blocks BEFORE mined_height+retention (1510)
```

A transaction mined on the longest chain at 1500 is deleted at 515. The outputs are left unspent on purpose, because that is what leaves `delete_at_height` null, and null is the one condition under which the `COALESCE(delete_at_height, $3)` in `SetConflicting` writes anything at all. Nothing bumps it afterwards, since `SetMinedMulti` already ran.

Whether a *counter-conflicting* transaction specifically can reach that ordering is the question the audit needs to answer. A reorg demoting a previously mined transaction is the obvious candidate. The audit as written does not answer it either way.

## The SQLite deadlock is real, and it needs its own issue

I checked the claim in the test comment rather than taking it. `Store.SetConflicting` opens a write transaction with `s.db.Begin()`, issues the `UPDATE` through it, and then calls `s.GetSpend` on the pool, which cannot get a connection until that transaction commits. Called with one real hash it does not return within 5 seconds.

That is pre-existing and this PR is right not to fix it here. But a production-code deadlock currently exists only as a comment inside a test file on an unrelated PR, and the existing SQL tests miss it because they all pass an empty hash slice. It should be an issue in its own right, otherwise it is lost the moment this branch merges or closes.

## What I did not check

I ran `./stores/utxo/sql/` on f79e2ce9, the three new tests pass, plus two throwaway probes which I have deleted. Nothing on Aerospike, so the Lua reading above is static only, not executed. Nothing on a node.

The one red check, `legacy-sync`, fails because an SVNode container could not bind RPC port 48332 on the runner. That is a port collision, unrelated to this branch.

icellan:
Re-reviewed at f79e2ce9a. The new commit is 211 lines of test in `stores/utxo/sql/prune_horizon_test.go` and no production change, so the four blocking findings are all still open and still at the same lines.

On the one they do engage with — the prune-horizon premise — the result holds for SQL and not for Aerospike.

SQL bumps a stale stamp forward, as the commit message says: `setMinedMultiChunk` (`sql.go:3361-3364`) has `WHEN delete_at_height IS NOT NULL AND delete_at_height < $1 THEN $1`. So a record flagged conflicting while unmined does get lifted to `mined_height + retention` when it is later mined, and the horizon the guard depends on holds there.

Aerospike refuses that bump on both paths:

- `teranode.lua:990-999` handles conflicting first and returns early — `if rec[BIN_CONFLICTING] then if not existingDeleteAtHeight then ... end; return "", nil end`. An existing DAH is never raised.
- `set_mined_expressions.go:241` — `ExpAnd(isConflicting, dahNotExists)` sets the DAH, everything else falls through to `ExpUnknown()`. Same semantics.

The comment at `sql.go:3350-3353` says the SQL CASE "Mirrors aerospike Lua setDeleteAtHeight logic". It does not: it diverges on exactly the case this commit pins. Combined with `SetConflicting` stamping from the cached tip (`sql.go:4280`, and the Aerospike equivalent), a record marked conflicting while the cache reads 500 against a real tip of 1200 keeps DAH 789 through being mined at 1200 and is pruned inside a parent's tolerate window — the state the guard asserts cannot exist.

So the horizon is pinned on the backend where it already held, and the commit message states the conclusion generally. The test is `stores/utxo/sql` only, so the Aerospike coverage gap noted against `process_conflicting.go:1224` is also unchanged.

Separately, worth its own issue rather than anything here: the deadlock you documented in the commit message is real and pre-existing — `SetConflicting` opens a transaction at `sql.go:4325` and then calls `s.GetSpend` on the pool at `sql.go:4383`, which cannot get a connection until that transaction commits. The existing sql tests miss it because they all pass an empty hash slice.

stores/utxo/process_conflicting.go
```diff
@@ -1241,14 +1276,54 @@ func GetCounterConflictingTxHashes(ctx context.Context, s Store, txHash chainhas
...
 		// call the package-level walk directly (not the Store method) so the
 		// caller-chosen maxNodes budget flows into the BFS
 		childHashes, err := GetConflictingChildren(ctx, s, spendingTxID, maxNodes)
 		if err != nil {
+			// A counter-conflicting record that is absent from the store has no
+			// BlockIDs, so it is definitionally not mined on our chain. Tolerate it
+			// (exclude from the counter set) only when every parent output slot it
+			// occupies is confirmed within the retention window of tip: there, no
+			// counter mined on that slot could yet have been pruned, so the absence
+			// must be a never-created loser — a spend recorded on the parent whose
```
freemans13, line 1292: "must be a never-created loser" is the load-bearing claim, and #1214's own evidence contradicts it.

The issue describes its mainnet instance `fbdebded4b373806c6b6883b5b16c91d7d6e6ea77aa56e077f74b165b4a0dabb` as "a never-mined first-seen loser whose record got **reaped**", and names the mechanism as suspect 1: `SetConflicting` stamps a delete-at-height, the pruner deletes the record. That record was created, then deleted. It is not the spend-first window this comment describes.

The stamp that deletes it is unrelated to the parent's height. `SetConflicting` writes `GetBlockHeight() + 1 + retention` from the store's lagging cached tip (`sql.go:4280`). I measured a transaction mined at 1500 pruned at tip 515 on this branch, because the flag was written while the cache read 500.

So the recency of the parent proves nothing about whether the absent record ever existed. It proves only that a counter carrying the *mined* delete-at-height (`mined_height + retention`) has not been pruned. A record marked conflicting first is deleted on a different clock, and on Aerospike that stamp is never raised afterwards (`teranode.lua:990-999`), so it survives being mined.

services/subtreevalidation/SubtreeValidation.go
```diff
@@ -473,7 +473,7 @@ func (u *Server) blessMissingTransaction(ctx context.Context, blockHash chainhas
 func (u *Server) checkCounterConflictingOnCurrentChain(ctx context.Context, txHash chainhash.Hash, blockIds map[uint32]bool) error {
 	// the tx is conflicting, check whether the counter-conflicting transactions have already been mined on our chain
 	// first get the parent transactions and check if they were spent
-	counterConflictingTxHashes, err := utxo.GetCounterConflictingTxHashes(ctx, u.utxoStore, txHash, u.settings.UtxoStore.ConflictingChildrenMaxNodes)
+	counterConflictingTxHashes, err := utxo.GetCounterConflictingTxHashes(ctx, u.utxoStore, txHash, u.settings.UtxoStore.ConflictingChildrenMaxNodes, u.settings.GetUtxoStoreBlockHeightRetention())
```
freemans13, line 476: The guard is threaded into this call, but the second walk in the same function is left bare.

`utxo.GetConflictingChildren(ctx, u.utxoStore, txHash, ...)` at line 519 walks the winner's own descendant cone. It reads its root and every descendant through the same `s.Get` (`process_conflicting.go:1126`) and returns the error unchanged, so a dangling spender reference anywhere in that cone gives `TX_NOT_FOUND`, gets wrapped into a `ProcessingError` at line 520, and wedges block validation exactly as before this PR.

#1214 names `GetConflictingChildren` as one of the three readers that hit `TX_NOT_FOUND` on a dangling reference. Guarding only the counter walk closes one of them.

freemans13 (commented):
Still blocked, and this round has two findings that are new rather than a restatement. The code has not moved: `git diff --name-only 8bd866aa f79e2ce9` is one test file, so icellan's four blockers all stand at the same lines and I will not relitigate them.

## The guard's premise is contradicted by the evidence in #1214 itself

The comment at `process_conflicting.go:1287-1296` says an absent record inside the parent's retention window "must be a never-created loser". The issue it cites says otherwise. #1214's documented mainnet instance, `fbdebded4b373806c6b6883b5b16c91d7d6e6ea77aa56e077f74b165b4a0dabb`, is described there as "a never-mined first-seen loser whose record got reaped". Reaped, not never created. Suspect 1 in that issue names the mechanism: `SetConflicting` stamps a delete-at-height on the loser, and the pruner then deletes it.

That stamp has nothing to do with the parent's height. `SetConflicting` writes `GetBlockHeight() + 1 + retention` from the store's cached tip (`sql.go:4280`), and that cache lags. I measured it on this branch on 3 September: a transaction mined at 1500 was pruned at tip 515, because the conflicting flag was written while the cache read 500.

So a record can vanish from well inside the parent's window without ever having been a never-created loser, and the guard tolerates it. The window the guard proves is the mined-counter pruning horizon, `mined_height + retention`. The window it needs to prove is that nothing which could carry `BlockIDs` has been deleted, and the conflicting stamp goes straight through that.

The residual-risk section says this would need a record "conflicting-flagged and mined on the longest chain at once, which is contradictory in the node's own model". It is not contradictory, it is a sequence. `MarkTransactionsOnLongestChain(false)` sets `unmined_since` and never clears `block_ids` (`sql.go:4673-4678`, and Aerospike `longest_chain.go:62`). A transaction demoted by a reorg, flagged conflicting, then re-mined when that fork comes back carries `BlockIDs` and a stale conflicting stamp at the same time. On Aerospike that stamp is never raised (`teranode.lua:990-999`), so the record is reaped inside the parent's window, the walk 404s, the guard tolerates, and its `BlockIDs` are never tested at `SubtreeValidation.go:530`. That is a counter mined on our chain dropped from the check.

## Two of the three readers #1214 names are still unguarded

#1214 states it plainly: "every reader that walks that ref (`GetCounterConflictingTxHashes`, `GetConflictingChildren`, `ProcessConflicting`) hits `TX_NOT_FOUND`". This PR guards the first.

The clearest gap sits four lines below the guarded call, in the same function. `checkCounterConflictingOnCurrentChain` walks the winner's own descendant cone at `SubtreeValidation.go:519` through a bare `utxo.GetConflictingChildren`. That walk reads its root and every descendant through the same `s.Get` (`process_conflicting.go:1126`) and returns the error unchanged, so a dangling spender reference anywhere in the winner's cone still yields `TX_NOT_FOUND`, still gets wrapped into a `ProcessingError` at `:520`, and still wedges block validation on the block. Same inconsistency, same function, no tolerance.

The two store-method entry points are unguarded for the same reason: `aerospike/conflicting.go:37` and `sql/sql.go:4271`. So is the parent read at `process_conflicting.go:1224` inside the guarded function, which I raised on 3 September and which is where the recurrence actually lands once the pruner takes the parent.

For a change whose entire justification is liveness, closing one door of four leaves the node wedging on the same class of store state.

## #1214 is a write-side issue, and it parked the read-side track

This one is a call for whoever owns the issue, not a review argument. #1214's status block says the read-side recovery approach (#1213, and PRs #1217 and #1239) was closed, that "we're not carrying that fix through more review", and that preventing the reference from forming at the write side "is the durable fix and the only track going forward". Its five acceptance criteria are all write-side: enumerate every delete path, verify each one unspends and clears `conflictingChildren` before deleting, decide the delete-at-height policy for conflicting losers, add an invariant check with a metric, and reproduce the deletion path in a test.

This PR is a read-side tolerance, which is the track that was parked. The description does not claim to close the issue, which is right, but the title carries `(#1214)` and that reference lands in the merge commit and on the issue timeline. If it merges as titled, the write-side work reads as done when none of it has been started.

My preference has not changed since 3 September: close the spend-first window at source (#1355), or repair the orphaned slot, rather than tolerate it for `retention` blocks and wedge anyway once they run out.

## What I checked

I read the diff at `f79e2ce9a` against `2935344e0`, all three comment surfaces, issue #1214, and the call sites named above. I ran nothing locally this round, so everything here is code reading plus the measurements from my earlier rounds, which I have labelled as such. CI on this head is green except `test aerospike native-ops (fork image)`, which reports `skipping`, so nothing in that run touches the production backend where the prune-horizon argument fails.

PR #1683, opened 2026-09-03
p2p: remove dead 10MB gossip cap, ceiling topics at 10KB

gokutheengineer:
**Closes bitcoin-sv/teranode#4757.**

### Problem
- `maxP2PMessageSize = 10MB` in `services/p2p/Server.go` was dead code: two hits in the repo (comment + declaration), and its comment claimed a pre-parse safety net that does not exist. Confirmed still present.
- The other claims are confirmed but not fixable here: `go-p2p-message-bus` (now v0.1.23) still exposes neither `pubsub.WithMaxMessageSize` nor `RegisterTopicValidator`, and does not expose the underlying `pubsub.PubSub`. So the libp2p 1MiB default is the only pre-relay size check, and the per-topic caps (and the ban check) run in the subscription handlers after gossipsub has fanned the message out. `docs/p2p-libp2p-review.md` already tracks this as the top outstanding hardening item, blocked on the library.
- Additionally, the per-topic caps were looser than the payloads justify: block was 32KB (headroom for a `Coinbase` field no Teranode version has ever populated and nothing consumes) and node_status was 16KB. Real gossip messages are ~1KB.

### Fix
- Delete `maxP2PMessageSize` and its false comment.
- Add `maxGossipMessageSize = 10KB` as the documented ceiling every per-topic cap must stay under, with an honest comment that it is not enforced on the wire and why.
- Lower block and node_status caps to the 10KB ceiling. Subtree and rejected_tx stay at 8KB. The egress side (`publishToNetwork` cap check, node_status URL-drop fallback) already enforces the same table, so a local misconfiguration fails loudly rather than being dropped by peers.
- Update `docs/p2p-libp2p-review.md` (three lines still described a 10MB safety net / cap).
- Reconcile the Coinbase comments in `gossip_field_validation.go` and its test with the new no-headroom stance, and state the tradeoff: another implementation populating Coinbase with a tx above ~3.8KB (hex-doubled) would now have that announcement dropped unscored.

### Residual
- Pre-relay enforcement (`WithMaxMessageSize`, per-topic `ValidatorEx`, pre-relay ban check) needs a `go-p2p-message-bus` change first; not addressable in this repo.

### Tests
- New `TestTopicKindCaps_WithinGossipCeiling`: all four subscribed topic kinds have a cap and every cap is at or under 10KB.
- `TestFullyPopulatedMessagesFitUnderCaps` now really is fully populated: adds `FeePolicy` and `MinMiningTxFee`, which a real node sets (marshals to ~5.9KB against the 10KB cap).
- `TestEscapeHeavyURLsStillFitUnderNodeStatusCap` adjusted: it built two 2KB URLs made half of `&` (14.8KB marshalled) and called that realistic. It now uses a realistic query-string density at the full URL bound; the pathological case remains covered by `TestHandleNodeStatusNotification_OversizedURLsDroppedNotSilenced`.
- Existing oversized-drop and fully-populated-fits-under-cap tests pass against the new constants.

```
ok  	github.com/bsv-blockchain/teranode/services/p2p	19.993s
```


ctnguyen (approved):
Review at commit [9cd054a](https://github.com/gokutheengineer/teranode/commit/9cd054aecc61e3d95b62d2b6165a6b64a85b1a9b)

Good cleanup, and the honesty is the best part of it. Deleting a constant whose comment advertised a pre-parse safety net that was never wired to anything is worth doing on its own; replacing it with a comment that says out loud that the new ceiling is *not* enforced on the wire, and why, is what stops the next reader from re-discovering the same gap from scratch. `TestTopicKindCaps_WithinGossipCeiling` is a good addition too: its ceiling loop ranges over the live `topicKindCaps` map rather than restating the four constants, so raising any cap past the ceiling fails the test wherever that cap happens to be declared. The doc edits track the code, and I found no stale reference to the removed 10MB net anywhere in `docs/` or `services/p2p/`. Three notes below, none of them blocking.

## Non-blocking issues

### ChiR1 — the URL field bound and the tightened topic caps are no longer mutually satisfiable, and the test that pinned the worst case was relaxed alongside them

**Problem:** `maxGossipURLLen = 2048` (`services/p2p/gossip_field_validation.go:21`) and `maxBlockMessageSize = maxGossipMessageSize` = 10240 (`services/p2p/Server.go:107`) are no longer mutually satisfiable. `encoding/json` HTML-escapes `<`, `>` and `&` into six-byte `\uXXXX` escapes (one byte in, six bytes out), so a field of length `L` carrying `k` escapable bytes marshals to `L + 5k`. `BlockMessage` has no struct tags, so its marshalled shape is

  `services/p2p/message_types.go:98` — `{"PeerID":"…","ClientName":"…","DataHubURL":"…","Hash":"…","Height":N,"Header":"…","Coinbase":"…"}`

  which is 91 bytes of braces, keys, colons, commas and value quotes. With every field at its bound and no escapable byte: 91 + 128 + 128 + 2048 + 64 + 10 + 160 = **2629 bytes**, which is what `TestFullyPopulatedMessagesFitUnderCaps` asserts. Make the 2048-byte `DataHubURL` all `&` and it becomes 2629 + 5×2048 = **12869**; do the same to `PeerID` (`checkGossipString` permits any printable rune) and it is **13509**. One convention for every number in this review: I escape only the fields `validateFields` actually bounds — `PeerID` and the URLs, since hex fields cannot hold an escapable byte — and leave `ClientName` plain, because `sanitizePeerDisplayString` strips `<`, `>` and `&` outright rather than bounding them. Escaping `ClientName` too adds 640 bytes (block 14149, subtree 13943) and changes no conclusion below. The crossover is exactly 1523 escapable bytes in the URL: 2629 + 5×1523 = 10244 > 10240, while 1522 gives 10239 and fits. Under the old 32KB cap 13509 could never breach, so `block` was the last topic where passing `validateFields` *guaranteed* fitting under the cap — `subtree` (8KB, worst case 13303) and `node_status` (two URLs, worst case ~25KB) were already in this state before the PR.

  The second half is the test. `TestEscapeHeavyURLsStillFitUnderNodeStatusCap` was the only test exercising a field-bound worst case, and it had to change to stay green: the old `"a&"` URL marshals that message to 14838 bytes — the PR description's own "14.8KB" — which is fine under 16KB and 4.6KB over 10KB. The new `"key=val&"` density lands at ~7.2KB. The relaxation is argued as a realism improvement, and that argument is fair for node_status egress, but it is also the change that let the tightened cap pass, and it leaves nothing in the suite exercising the worst case the field bounds actually admit.

**Why it matters:** Three cases, and they are not equally serious. Hostile ingress is the mildest: no DataHub URL contains 1523 of `<>&`, so an attacker crafting one only gets their own announcement dropped before parse and unscored (`services/p2p/server_helpers.go:38`) — the cost there is to the invariant, not to any traffic. Local egress is the case the PR designs for: `handleBlockNotification` has no degraded form, so it hard-fails at the cap check in `publishToNetwork` (`services/p2p/publish_gate.go:185`) and, because that error returns at `services/p2p/Server.go:1700`, also skips the piggybacked `node_status` publish. That is the deliberate fail-loudly-locally design the comment describes, but it means the relationship between the two bounds now has to be maintained on purpose instead of holding by construction.

  The third case is the one I'd actually weigh at merge time, because the emitter is neither an attacker nor a misconfigured local node: during a rolling upgrade, an un-upgraded peer still enforces 16KB on its own node_status egress, so it will happily publish an escape-heavy but in-bounds status that upgraded peers now drop before unmarshal at `services/p2p/Server.go:1471`. Upgraded nodes then stop learning that peer's height and tip until it is upgraded. node_status is also the easiest of the three to trip: from the 5934-byte plain baseline in `ChiR2`, only 862 escapable bytes across the two URLs cross 10240, against 1523 for `block`. Worth noting because the PR reasons about exactly this hazard one constant away — `maxRejectedTxMessageSize` keeps its 8KB explicitly because "un-upgraded peers publish the untruncated validator error chain, so keep headroom for those during mixed-version operation" (`services/p2p/Server.go:117-120`) — while node_status gives up 6KB of headroom here with no equivalent note.

**Fix:** Pick one and write it down. Lowering `maxGossipURLLen` only half-works (≤1503 makes block's worst case fit, but two URLs on node_status cannot be made to fit at any useful length), so the cheaper option is to keep the bounds and pin the crossover explicitly, asserting the breach rather than a fit:

  ```go
  // The per-field bounds do NOT imply the topic cap: json escaping can expand
  // an in-bounds URL 6x. Pin the crossover so a future cap change is deliberate
  // and handleBlockNotification's hard-fail stays the documented outcome.
  func TestWorstCaseInBoundsBlockExceedsCap(t *testing.T) {
      b, err := json.Marshal(BlockMessage{
          PeerID: strings.Repeat("p", maxGossipPeerIDLen),
          ClientName: strings.Repeat("d", maxPeerDisplayStringLen),
          DataHubURL: strings.Repeat("&", maxGossipURLLen),
          Hash: strings.Repeat("0", maxGossipHashLen), Height: ^uint32(0),
          Header: strings.Repeat("0", maxGossipHeaderLen),
      })
      require.NoError(t, err)
      require.Greater(t, len(b), maxBlockMessageSize,
          "in-bounds fields can still breach the block cap; the publisher must hard-fail, not publish")
  }
  ```

  and narrow the "the property the tightened caps rest on" phrasing at `services/p2p/gossip_field_validation_test.go:630-633`, which now overstates its reach: the property is broken on the three topics that carry a URL, and `maxGossipURLLen` is the whole cause. It still holds for `rejected_tx`, structurally — with no URL field, its bounds give 51 + 128 + 128 + 64 + 1024 = 1395 bytes plain, and 7795 even with `PeerID`, `Reason` and `ClientName` all escapable, comfortably under the 8192 cap.

### ChiR2 — the fully-populated ceiling is computed but never recorded, so the remaining margin is unpinned

**Problem:** `TestFullyPopulatedMessagesFitUnderCaps` asserts only `require.Less(t, len(nodeStatus), maxNodeStatusMessageSize)` (`services/p2p/gossip_field_validation_test.go:659`). Summing the real bounds — 25 fields giving 361 bytes of quoted keys and colons plus 24 commas and 2 braces, 5212 bytes of string values at bound (2×2048 URLs, 5×128 display strings, 2×128 peer IDs, 3×64 hex, the two enums) plus their 30 quotes, 100 bytes of numerics, and the 205-byte `fee_policy` sub-object — gives **5934 bytes** against a 10240 cap. That is 58% of the cap and 4306 bytes of slack that nothing in the tree states. Minor accuracy note on the same test: `maxFee := 1e300` marshals as `1e+300` (6 bytes), which is not the widest float64 encoding (`math.MaxFloat64` marshals as `1.7976931348623157e+308`, 23 bytes), so "fully populated" is ~34 bytes short across the two float fields.

**Why it matters:** This test is the only artefact in the PR that answers the maintainer's second ask on the linked issue — "suspect message around 64KB > verify!" (see [bitcoin-sv/teranode#4757](https://github.com/bitcoin-sv/teranode/issues/4757)). The answer is reassuring: the largest message class tops out at ~5.9KB with every field at its bound, nowhere near 64KB. But it exists only as a number a reader has to recompute by running the test, so the verification the maintainer asked for is not recorded anywhere they will find it. Separately, a bare `Less(cap)` gives no signal as headroom is consumed — `NodeStatusMessage` gained `FeePolicy` and `MinMiningTxFee` recently enough that this PR had to add them to the test, so the field list demonstrably grows.

**Fix:** Assert against a stated margin, so the number lives in the source and a future field addition fails with a message that explains itself:

  ```go
  // The field-bound ceiling for node_status is ~5.9KB (25 fields, 2x2048B URLs,
  // 5x128B display strings, the fee_policy sub-object) — nothing legitimate is
  // anywhere near 64KB. Keep ~35% of the 10KB cap free for fields added later.
  require.Less(t, len(nodeStatus), maxNodeStatusMessageSize*65/100,
      "fully populated node_status must fit under its cap with room for future fields")
  ```

### ChiR3 — the dead `Coinbase` field has no `omitempty`, so it pads every block announcement

**Problem:** `BlockMessage` carries no struct tags at all, so `Coinbase` (`services/p2p/message_types.go:105`) has neither a json name nor `omitempty` and every block announcement this node publishes spends a literal `,"Coinbase":""` — 14 bytes on the wire for a field with no producer and no consumer. Over tracked content, `git grep -n Coinbase -- services/p2p/` returns only the declaration, three comments and `TestHandleBlockTopic_NonHexCoinbaseIgnored` (which sets it); a repo-wide `git grep -n '\.Coinbase\b'` finds no `BlockMessage.Coinbase` access anywhere — every hit belongs to the UTXO wrapper, the mining solution or the `Coinbase` settings block. This PR grows the field's footprint rather than shrinking it: it now costs a comment at `services/p2p/Server.go:105-106`, a rewritten comment at `services/p2p/gossip_field_validation.go:112-118` carrying the "~3.8KB (hex-doubled)" tradeoff, and a comment at `services/p2p/gossip_field_validation_test.go:632-633`. (That `~3.8KB` figure checks out, for the record: 10240 − 2629 = 7611 hex characters = 3805 raw bytes, i.e. the worst case with every other field at its bound. Keep the `~` — with a realistic short DataHub URL the headroom is nearer 4.8KB of raw transaction.)

**Why it matters:** 14 bytes against a 10KB cap is not a capacity problem, so the payload is the cheap part. The cost is that three comment sites and a documented byte budget have to stay in sync with a field no code path exercises, and this PR already needed a follow-up commit (`9cd054a`, "reconcile coinbase comments") to resync exactly those comments after `7a77678` changed the cap. Tagging the field stops the wire waste in one line, and no code path anywhere reads what that removes; the comments then have one place to live instead of three.

**Fix:** Add the tag, keeping the field name exactly as it is today:

  ```go
  Coinbase string `json:"Coinbase,omitempty"` // Hexadecimal representation of the coinbase transaction
  ```

  The capital `C` is deliberate — the struct is otherwise tagless, so Go already marshals this field as `Coinbase`, and the more natural-looking `json:"coinbase,omitempty"` would silently rename it on the wire. The *encoding* does change, and that is the whole point: `omitempty` stops the key being emitted at all, which is where the 14 bytes go — announcements simply no longer carry `"Coinbase":""`. That is safe because nothing reads it, and because the two forms are indistinguishable to any typed receiver: `json.Unmarshal` leaves an absent key at its zero value, so a `BlockMessage` decoded from a payload without the key holds the same `""` it would hold from `"Coinbase":""`. The only receiver that could tell them apart is a loosely-typed non-Teranode parser indexing the key directly, which would find it missing rather than empty — and since Teranode has only ever emitted `""`, such a parser already has to handle the no-coinbase case. The larger alternative is to delete the field outright and fold the three comments into one line at the size check: `json.Unmarshal` ignores unknown fields, so a peer that does populate `Coinbase` is handled exactly as it is today (accepted, unscored, bounded only by the topic cap) and `TestHandleBlockTopic_NonHexCoinbaseIgnored` keeps its meaning; against that, the declaration is the only place this wire field is written down for another implementation, which is a fair reason to keep it tagged rather than gone.

## Recap

| --- | --- | --- | --- |
| ChiR1 | URL bound exceeds topic caps |  | 30% |
| ChiR2 | Fully populated margin unpinned |  | 15% |
| ChiR3 | Dead Coinbase field lacks omitempty |  | 10% |

liam (approved):
Reviewed independently on top of the existing review round. I reproduced the size arithmetic with a throwaway probe on the branch rather than taking it on trust, and ran the suite.

**Verified:** the ChiR1 arithmetic is correct, to the byte in two places.

- block fully populated with plain URLs: **2629 bytes**
- block crossover: **exactly 1523** escapable bytes in the URL gives 10244 > 10240
- block with an all-`&` URL: 12769; node_status with two: 26293
- node_status fully populated: 6013 using `math.MaxFloat64`, against 5934 using `1e300`. The ~79-byte delta is exactly the float-width point that review makes itself.

So the claim stands: `maxGossipURLLen = 2048` and a 10KB block cap are no longer mutually satisfiable.

`SETTINGS_CONTEXT=test go test -race -tags testtxmetacache ./services/p2p/...` green, `go vet -tags testtxmetacache` clean.

## Where I would weight ChiR1 lower

The numbers that decide severity are the realistic ones, and they are not in the review. With an operator-shaped URL (`http://asset.node1.example.com:8090`) and a real libp2p peer ID, a block announcement is **540 bytes** and a node_status is **1723 bytes**, against 10240 caps. That is 6x to 19x headroom. Tripping the tightened cap needs about 1523 escapable bytes in one DataHub URL, or about 860 across two. No operator-configured URL looks like that.

Two mitigations also go uncredited, and both cut the cases the review weighs at merge time:

1. **node_status egress does not hard-fail.** `Server.go:2309-2311` drops the optional URLs and republishes when the marshalled form exceeds the cap. Only block hard-fails, and block is the topic needing 1523 escapable bytes rather than 860.
2. **The piggybacked node_status skip costs one tick, not the update.** `nodeStatusPublishInterval = 10 * time.Second` (`Server.go:1799`) publishes node_status on its own timer, so a block-publish hard-fail delays node_status by at most 10 seconds rather than suppressing it until the next block.

The rolling-upgrade case survives as real but narrow. I agree with the recommended shape: keep the bounds and pin the crossover in a test that asserts the breach. Chasing the two bounds into agreement does not work anyway, since two URLs on node_status cannot be made to fit at any useful URL length.

## One finding I do not think has been raised

**The new guard test cannot catch the drift it was written for, on the two topics it names.**

`maxBlockMessageSize = maxGossipMessageSize` (`services/p2p/Server.go:112`) and `maxNodeStatusMessageSize = maxGossipMessageSize` (`:126`), so `require.LessOrEqual(capBytes, maxGossipMessageSize)` is tautological for both. Raising `maxGossipMessageSize` to 1MiB would raise block and node_status to 1MiB with it, and `TestTopicKindCaps_WithinGossipCeiling` would still pass. That is exactly the scenario its own comment claims to guard: "a cap drifting back up towards the libp2p 1MiB default would silently widen the pre-parse abuse surface on that topic."

The test guards the caps against the ceiling; nothing guards the ceiling. One line closes it:

```go
require.LessOrEqual(t, maxGossipMessageSize, 16*1024,
    "the gossip ceiling itself must stay tight; raising it raises every aliased topic cap with it")
```

The test is worth keeping either way. Its real reach today is subtree and rejected_tx, plus catching a future edit that breaks the alias. The comment just over-claims.

## Corroborating the automated doc comment

Worth actioning rather than skipping, because this PR's own edit created the inconsistency. `docs/p2p-libp2p-review.md:76` now reads "a 5 MB blob on `node_status` is rejected before JSON parsing" immediately before the newly added "the only pre-relay size check is libp2p's 1MiB default". A 5MB message never reaches the handler, so the per-topic cap never sees it. The band the 10KB ceiling actually polices is 10KB to 1MiB. It is a one-number fix.

## Verdict

Good change. The dead-constant removal plus a comment that says out loud the new ceiling is not enforced on the wire is the valuable part, and the tightening sits well clear of what real traffic uses. Everything above is comment or test accuracy.

Approving.

PR #1684, opened 2026-09-03
p2p: always bind p2p_port and reject listen addresses libp2p cannot honour

gokutheengineer:
**Closes bitcoin-sv/teranode#4758.**

### Problem
Both items confirmed still valid on `upstream/main` (line numbers had shifted to `Server.go:360-363` and `Server.go:413-420`).

1. `conf.Port` was only handed to go-p2p-message-bus inside `if len(advertiseAddresses) > 0`. Committed settings ship no advertise addresses and `p2p_share_private_addresses = false`, so the bus bound `/ip4/0.0.0.0/tcp/0` + `/ip6/::/tcp/0`: a random ephemeral port that changes every restart, while `p2p_port` was validated as mandatory and then discarded. Every firewall rule, k8s Service and compose mapping for 9905 pointed at nothing.
2. `p2p_listen_addresses` never reached libp2p. The bus has no listen-address option and hardcodes the dual-stack wildcard bind, so `127.0.0.1` (dev) or a narrowed interface was silently ignored while the node still listened on every interface. Its `nil` check was dead (`getMultiString` returns a non-nil empty slice), and with `p2p_share_private_addresses = true` the raw value was copied into `AnnounceAddrs`: for a hand-written bare `0.0.0.0` the bus rejects it as an invalid multiaddr at startup, and for the committed `docker.host` multiaddr it announced an undialable wildcard. No committed context ships the crashing form.

### Fix
- `buildP2PMessageBusConfig` always sets `conf.Port = tSettings.P2P.Port`; `AnnounceAddrs` is set independently when advertise addresses exist.
- New `settings.ValidateP2PListenAddresses(addrs, port)`: accepts only the wildcard bind the bus actually performs (`0.0.0.0`, `::`, `/ip4/0.0.0.0[/tcp/<p2p_port>]`, `/ip6/::[/tcp/<p2p_port>]`). Empty, a specific interface, a foreign port or a non-IP multiaddr fail startup with a message saying what the bus binds and to use a firewall or `p2p_advertise_addresses` instead. This is the issue's fallback option, since the proper fix (a `ListenAddrs` field) lives in go-p2p-message-bus; once the bus grows one, the validator is the single place to relax.
- Advertise-address selection extracted into `resolveAdvertiseAddresses` (replacing a test-only mirror copy of the logic). `SharePrivateAddresses` no longer copies listen addresses into `AnnounceAddrs`: a wildcard is not dialable and crashes the bus. With empty announce addrs libp2p already advertises every bound interface address, private ones included, which is what that setting intended.
- `cmd/diagnose` uses the same validator, so a narrowed value is reported as ERROR instead of OK.
- `settings.conf`: dropped the inert `p2p_listen_addresses.dev = 127.0.0.1` (would now fail startup), documented the accepted forms.
- `p2p_share_private_addresses` was already ineffective (it either announced the undialable wildcard or crashed the bus); its `longdesc`, `settings.conf` comment and docs now say plainly that it has no effect on announcement with the current bus, instead of promising privacy it cannot deliver.
- Setting `longdesc`s and the P2P reference docs updated; the stale "Port only used when AdvertiseAddresses are specified" comment corrected.

### Residual
- **Silent listen mode now binds `p2p_port` deterministically** instead of the ephemeral port it got by accident from the old bug. It still does not announce itself (DHT off, no announce override, no node_status), but peers it dials learn its bound addresses over Identify with a stable port, and an unsolicited dial to `p2p_port` succeeds. The `listen_mode` description and reference doc now say so and point at a firewall for inbound; a genuinely inbound-free mode needs the same bus-side listen option as item 2.
- The proper fix for item 2 is a `ListenAddrs` option in go-p2p-message-bus; until then the validator is the single place to relax when the bus gains one. `p2p_share_private_addresses` likewise needs a bus-side address filter to mean anything.
- `p2p_advertise_addresses` entries are still handed to the bus unvalidated, so a bare `host:port` fails inside `NewClient` as a service error rather than a configuration error at startup. Left for a follow-up; the reference doc that recommended that format is corrected here.
- Deployment-visible: a `settings_local.conf` still carrying the old shipped `p2p_listen_addresses.dev = 127.0.0.1` now fails startup with an explicit message, and two default-configured nodes on one host now collide on 9905 instead of one silently falling back to a random port. Both are the intended fail-loud behaviour.
- In-process multi-node daemon tests now really bind and dial `p2p_port` (previously the dialled port had nothing listening). Smoke and `test/e2e/daemon` suites cannot link locally (GoBDK cgo), so CI is the first run of those with this change.

### Tests
- `settings`: `TestValidateP2PListenAddresses` (accepted and rejected forms), `TestCommittedP2PListenAddressesAreValid` (committed `settings.conf` passes the startup check; run under `test`, `dev`, `docker.m`, `operator` contexts).
- `services/p2p`: `TestBuildP2PMessageBusConfig_MeshProtection/port is set without advertise addresses` (regression for item 1), `TestResolveAdvertiseAddresses` (silent / explicit / share-private never announces the wildcard), `TestNewServerRejectsNarrowedListenAddresses` (real constructor fails with a configuration error for `127.0.0.1`, `10.0.1.5`, port mismatch, unset).
- `cmd/diagnose`: `TestCheckP2PConfigListenAddresses` (narrowed value, port mismatch and zero port report ERROR).
- Review round (ctnguyen, freemans13) added: the committed-config guard layers `compose/settings_test.conf` over `settings.conf` for the `docker.host` contexts, the only committed values carrying an explicit port, and pins that the layered value is the port-bearing one; the `host:port` slip (`0.0.0.0:9905`) gets a hint naming the accepted spellings; diagnose cells are asserted free of the rendered error-code chain.
- Fresh-eyes review round added: duplicate-component multiaddrs (`/ip4/0.0.0.0/tcp/9905/ip4/127.0.0.1/tcp/9905`) are rejected, and the committed-config guard reads `settings.conf` directly across every shipped context instead of going through `NewSettings`.

```
SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./settings/        # ok
SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./services/p2p/... # ok
go vet ./cmd/diagnose/                                                                # ok (test binary cannot link locally: GoBDK cgo)
```


ctnguyen (changes requested):
Review at commit [94518cc](https://github.com/gokutheengineer/teranode/commit/94518cc955699d349f6f6cbe908e788706e8a2d4)

This is a good fix for a genuinely nasty class of bug: config that was validated as mandatory and then thrown away. Moving `conf.Port = tSettings.P2P.Port` out of the `len(advertiseAddresses) > 0` branch is the right one-line change, and refusing a listen address the bus cannot honour is issue [bitcoin-sv/teranode#4758](https://github.com/bitcoin-sv/teranode/issues/4758)'s own recommended fallback rather than an invented design. I liked the details: the validator walks multiaddr components instead of using `ValueForProtocol`, so `/ip4/0.0.0.0/tcp/9905/ip4/127.0.0.1/tcp/9905` cannot sneak through on its leading wildcard, and the test table covers exactly that. Deleting the test-only mirror copy of the advertise-address logic in favour of a real `resolveAdvertiseAddresses` is a straight improvement, `cmd/diagnose` is kept in sync so it stops reporting an inert value as OK, and the longdesc rewrites replace an aspirational promise with what the code actually does. Test coverage is proportionate to the risk — a real-constructor rejection test plus a committed-config guard is more than most config PRs bring.

The findings below are mostly about the documentation half of the change, which is fair game here because truthful configuration documentation *is* this PR's deliverable. One security-oriented setting had its behaviour changed by this diff without its documentation following.

## Blocking issues

### ChiR1 — Silent listen mode now binds the well-known port, and its documentation still promises stealth

**Problem:** Before this change, `resolveAdvertiseAddresses` returned empty for silent mode by construction, so `conf.Port` was never assigned and the bus bound `/ip4/0.0.0.0/tcp/0` + `/ip6/::/tcp/0` — an ephemeral port that changed on every restart. After `services/p2p/Server.go:364` sets `conf.Port` unconditionally, a silent node binds `p2p_port` (9905 by default) on both stacks, deterministically. The `listen_mode` longdesc at `settings/p2p_settings.go:15` is not touched by this PR and still reads *"does not advertise itself (no DHT advertisement, no address announcement, no node_status publishing). Minimises attack surface for monitoring or security-sensitive deployments"*, with a trade-off row of *"Maximum stealth, minimal inbound exposure"* and a recommendation for *"nodes that must minimise inbound connections"*. Nothing in the diff, including the Residual section, mentions silent mode.

Separately, `docs/references/settings/services/p2p_settings.md` — which this PR *does* edit (+7/-6) — under-documents silent in two rows. The settings row at `:16` gives `listen_mode` as *Node operation mode ("full" or "listen_only")*, and the Validation Rules table at `:108` reads `| ListenMode | Must be "full" or "listen_only" | Service fails to start if invalid | During server initialization |`. The *validation* half of that row is accurate — `services/p2p/Server.go:482` does reject anything outside the three constants — but the *enumeration* is incomplete in both rows: `silent` is a valid third value and appears in neither.

**Why it matters:** The bind change is correct and is the point of the PR — this is not a request to revert it. But two of silent mode's documented properties are now weaker than the text claims. First, with `AnnounceAddrs` empty libp2p's Identify hands every peer the node dials the addresses it actually bound; those addresses now carry a fixed port, so a silent node gives each outbound peer a dial-back address that stays valid across restarts instead of one that expires with the next process. Second, an unsolicited dial to the node's `p2p_port` now succeeds at a predictable, documented port rather than at a random high port that rotated per restart. Non-advertisement (DHT off, no announce addrs, no `node_status` publish) still holds, so the delta is discoverability-by-scan and durability of the dial-back address, not a new announcement channel. No committed context sets `listen_mode = silent` — `settings.conf:853-854` ship `full` and `dev` gets `listen_only`, and `deploy/docker/base/settings_local.conf:10` ships `listen_only` — so the affected population is operators who deliberately opted in, which is precisely the "security-conscious deployments" audience the longdesc addresses. A PR that took the trouble to rewrite three neighbouring longdescs because they overstated behaviour should not leave this one overstating it, and an operator who cannot find `silent` documented in the reference table it belongs in has no way to price the trade-off at all.

**Fix:** Update the `listen_mode` longdesc to say what silent does and does not buy, and add a line to the PR's Residual list so the change is disclosed. Something like:

  `- **silent** - Connects outbound to configured peers and syncs data, but does not advertise itself (no DHT advertisement, no AnnounceAddrs override, no node_status publishing). libp2p still binds p2p_port on all interfaces and still reports its bound addresses over Identify to peers it dials, so silent removes announcement, not inbound reachability: restrict that with a firewall on p2p_port.`

Change the trade-off row from *"Maximum stealth, minimal inbound exposure"* to something like *"No self-advertisement; inbound still reachable on p2p_port"*. In the reference doc, add `silent` to the settings row at `:16` and widen the Validation Rules row at `:108` to `Must be "full", "listen_only" or "silent"`, leaving its Impact column as is. If a genuinely inbound-free mode is wanted, that needs a bus-side listen option and belongs in the same follow-up as the `ListenAddrs` work already named in Residual.

## Non-blocking issues

### ChiR2 — A single-address-family entry is accepted while the bus binds both stacks

**Problem:** `validateListenAddress` (`settings/p2p_listen_addresses.go:40-91`) accepts each entry independently, so `{"/ip4/0.0.0.0/tcp/9905"}`, `{"/ip6/::"}` and bare `{"::"}` all pass — they are in the accepted list at `settings/p2p_listen_addresses_test.go:20-28`. The bus binds both families unconditionally: `go-p2p-message-bus` v0.1.23 `createHost` appends `libp2p.ListenAddrStrings` for `/ip4/0.0.0.0/tcp/<port>` *and* `/ip6/::/tcp/<port>`, which is also what this PR's own longdesc asserts. The new longdesc then labels the two-entry form as *"explicit dual-stack"*:

  `settings/p2p_settings.go:13:  - **/ip4/0.0.0.0/tcp/9905|/ip6/::/tcp/9905** - explicit dual-stack`

Calling the pair "explicit dual-stack" implies the single-entry forms are single-stack, which they are not.

**Why it matters:** This is the same failure mode the validator exists to close, one axis over. An operator on an IPv6-hostile network who writes `p2p_listen_addresses = /ip4/0.0.0.0/tcp/9905` to keep libp2p off IPv6 gets a clean startup and an IPv6 listener anyway — "silently ignored" behaviour that the reference doc now explicitly promises will not happen (`docs/references/services/p2p_reference.md:602`: *"a narrower value fails startup instead of being silently ignored"*). The practical harm is bounded — the IPv6 bind is the pre-existing behaviour, not something this PR introduced — but the new documentation turns a pre-existing quirk into a stated guarantee it does not keep.

**Fix:** Cheapest correct option is to stop implying family selection: drop *"explicit dual-stack"* and say the address family is not selectable, e.g. in the Format section state *"the /ip4 and /ip6 forms are interchangeable confirmations of the same bind; both families are always bound"*. If you would rather make the validator match the promise, reject a list that names exactly one family:

  ```go
  var sawIP4, sawIP6 bool
  // ... set from each entry's family while validating
  if sawIP4 != sawIP6 {
      return errors.NewConfigurationError(
          "p2p_listen_addresses names only one address family; libp2p always binds both 0.0.0.0 and :: on p2p_port %d", port)
  }
  ```

### ChiR3 — The committed-config guard never reads the one committed file that could fail it

**Problem:** `TestCommittedP2PListenAddressesAreValid` (`settings/p2p_listen_addresses_test.go:71`) iterates `shippedP2PContexts`, which includes `docker.host.teranode1` (`settings/p2p_bind_consistency_test.go:17-27`), but resolves every context through `readCommittedSettingsConf`, i.e. `settings.conf` only. For `docker.host.teranode1` that yields the context-free `p2p_listen_addresses = 0.0.0.0` (`settings.conf:1043`) against `p2p_port = ${P2P_PORT}` → 9905, and a bare `0.0.0.0` passes at any port, so the subtest is vacuous. The override that actually carries a port for that context lives in the other committed settings file:

  `compose/settings_test.conf:197:  p2p_listen_addresses.docker.host = /ip4/0.0.0.0/tcp/${PORT_PREFIX}${P2P_PORT}`
  `compose/settings_test.conf:177:  p2p_port.docker.host = ${PORT_PREFIX}${P2P_PORT}`

with `PORT_PREFIX.docker.host.teranode1 = 1` (`compose/settings_test.conf:27`). That file is committed and is mounted as `/app/settings_local.conf` by `compose/docker-compose-3blasters.yml` and `compose/docker-compose-chainintegrity.yml`.

**Why it matters:** Those two lines agree today, so nothing is broken — this is a coverage gap, not a live failure. But they are the only committed pair carrying an explicit port, therefore the only committed pair the new validator can reject on a port mismatch, and they are exactly what the guard does not look at. Change `PORT_PREFIX` handling in one line and not the other and the new validator turns it into a hard startup failure that the guard added in this PR would still report green. The doc comment justifies reading `settings.conf` directly so a developer's gitignored `settings_local.conf` cannot fail the test — sound for a gitignored file, but `compose/settings_test.conf` is neither gitignored nor a developer's. The repo already has the machinery: `settings/committed_private_keys_test.go:118-142` discovers every committed settings-format file via `git ls-files` and pins `compose/settings_test.conf` in `knownSettingsFiles` at line 60.

**Fix:** Reuse that discovery. Parse each committed settings file into the same `settingsConf` map, layer `compose/settings_test.conf` over `settings.conf` the way the compose mount does, and validate the resolved `(p2p_listen_addresses, p2p_port)` pair per context — or, if that is more than you want here, at least add a second subtest that resolves `docker.host.teranodeN` against `compose/settings_test.conf` so the port-bearing value is covered:

  ```go
  composeConf := readSettingsFile(t, filepath.Join("..", "compose", "settings_test.conf"))
  for _, ctx := range []string{"docker.host.teranode1", "docker.host.teranode2", "docker.host.teranode3"} {
      // resolve p2p_port / p2p_listen_addresses with composeConf layered over conf
  }
  ```

### ChiR4 — The NAT traversal guide still shows an advertise address that cannot start the node

**Problem:** This PR corrects `docs/references/services/p2p_reference.md:603` to say advertise entries must be complete multiaddrs and that a bare `host:port` is rejected. It also edits `docs/P2P_NAT_TRAVERSAL.md`, replacing the `p2p_share_private_addresses` warning at line 84 — but leaves the advertise example 37 lines above it untouched:

  `docs/P2P_NAT_TRAVERSAL.md:47:  p2p_advertise_addresses = ["/ip4/203.0.113.1/tcp/9905"]`

gocore reads that value literally, brackets and quotes included, so it reaches the bus as `["/ip4/203.0.113.1/tcp/9905"]`. Per the PR description the bus parses each entry with `multiaddr.NewMultiaddr` and returns an error from `NewClient`, which `services/p2p/Server.go:593` turns into a service error, so a node configured from this example does not start.

**Why it matters:** The section is headed "Option 2: Explicitly Set Advertise Address" and is the guide an operator behind NAT follows, which is exactly the audience this PR's reference-doc correction was written for. The failure is loud rather than silent, so nobody loses data — but a copy-paste from the NAT guide produces a node that will not boot, and the file is already open in this diff, so fixing it costs nothing.

**Fix:** Make the example a real multiaddr, matching the corrected reference doc:

  `docs/P2P_NAT_TRAVERSAL.md:47:  p2p_advertise_addresses = /ip4/203.0.113.1/tcp/9905`

Follow-up, outside this diff: the same unusable wording survives in two shipped templates this PR does not touch — `deploy/docker/base/settings_local.conf:8` and `deploy/docker/base/settings_local.conf.template:81` both suggest `p2p_advertise_addresses.docker.m` as a bare hostname, under a comment describing it as "the IP or domain exposed to the public internet with P2P_PORT". Those are the files a miner copies, so they are worth a separate sweep rather than being bundled here.

### ChiR5 — Diagnose puts the raw wrapped error chain in the operator-facing Recommended column

**Problem:** `cmd/diagnose/config_checks.go:426` builds the recommendation with `fmt.Sprintf("%v. Set p2p_listen_addresses to 0.0.0.0 or /ip4/0.0.0.0/tcp/%d", err, s.P2P.Port)`. `err` is an `*errors.Error` carrying `validateListenAddress`'s error as its wrapped cause, and `Error()` (`errors/errors.go:92-131`) renders the code name, the numeric code and every wrapped cause joined by `->`. For `p2p_listen_addresses = 10.0.1.5` the cell becomes roughly `CONFIGURATION (<n>): p2p_listen_addresses entry "10.0.1.5" is not supported (libp2p always binds 0.0.0.0 and :: on p2p_port 9905; use p2p_advertise_addresses or a firewall to control reachability) -> CONFIGURATION (<n>): binding a specific interface is not supported. Set p2p_listen_addresses to 0.0.0.0 or /ip4/0.0.0.0/tcp/9905`.

**Why it matters:** `diagnose` output is read by operators, and this is the one check in `checkP2PConfig` whose recommendation carries internal error codes and a wrap chain into a table cell. The advice is also partly redundant with the message it wraps, and when `p2p_port` is 0 the tail reads "or /ip4/0.0.0.0/tcp/0", which is not something anyone should set. Purely a readability point, but it is newly introduced by this diff in a tool the PR is otherwise improving.

**Fix:** Use the innermost message, or write the recommendation directly rather than interpolating the error. For example keep the validator's detail out of the cell and phrase the advice yourself:

  ```go
  Recommended: fmt.Sprintf("libp2p always binds 0.0.0.0 and :: on p2p_port %d; set p2p_listen_addresses to 0.0.0.0 or /ip4/0.0.0.0/tcp/%d and restrict exposure with a firewall", s.P2P.Port, s.P2P.Port),
  ```

### ChiR6 — The silent-mode log line asserts no advertise addresses are configured when some are

**Problem:** `services/p2p/Server.go:387` logs `"[silent mode] No advertise addresses configured"` on the silent branch. That branch is reached whether or not `p2p_advertise_addresses` is set, and its whole purpose is to discard a configured value — `TestResolveAdvertiseAddresses/silent mode suppresses explicit advertise addresses` covers exactly that path.

**Why it matters:** An operator who set `p2p_advertise_addresses` and then switched to silent mode gets a log line telling them nothing was configured, which hides the override rather than reporting it. The replaced message ("Address advertisement suppressed") at least named the suppression, even if its "node will not be discoverable" tail overclaimed — the accompanying comment now correctly attributes discoverability to the DHT being off, so only the message needs to carry the override.

**Fix:** Report the suppression and what was suppressed:

  ```go
  case tSettings.P2P.ListenMode == settings.ListenModeSilent:
      if len(tSettings.P2P.AdvertiseAddresses) > 0 {
          logger.Infof("[silent mode] p2p_advertise_addresses %v suppressed - nothing is announced in silent mode", tSettings.P2P.AdvertiseAddresses)
      } else {
          logger.Infof("[silent mode] no advertise addresses announced")
      }
      return nil
  ```

## Recap

| ChiR1 | Silent mode exposure undocumented | YES | 60% |
| ChiR2 | Single address family silently ignored | | 45% |
| ChiR3 | Guard skips compose settings file | | 28% |
| ChiR4 | NAT guide advertise example invalid | | 22% |
| ChiR5 | Diagnose output carries error chain | | 18% |
| ChiR6 | Silent mode log states falsehood | | 12% |

settings/p2p_settings.go
```diff
@@ -46,7 +46,7 @@ type P2PSettings struct {
...
+	SharePrivateAddresses bool `key:"p2p_share_private_addresses" desc:"Advertise private/local IP addresses to peers" default:"true" category:"P2P" usage:"Enable for local/test environments" type:"bool" longdesc:"### Purpose\nControls whether to advertise RFC1918 private IP addresses (10.x, 172.16.x, 192.168.x) to peers.\n\n### Current behaviour\nThis setting has **no effect** on what is announced with the current go-p2p-message-bus. When AdvertiseAddresses is empty, libp2p advertises every interface address it bound, private ones included, whatever this value is; the bus offers no address filter short of a full AdvertiseAddresses override. The wildcard ListenAddresses are never announced themselves: 0.0.0.0 is not a dialable address.\n\n### Recommendations\n- Local peers on one network can already dial each other; no change needed for local dev\n- To keep internal addresses private, set AdvertiseAddresses to the public address you want published\n- AllowPrivateIPs still governs whether this node dials private addresses"`
```
freemans13: **SF1 (blocking).** These two closing lines are wrong under the shipped defaults:

> Local peers on one network can already dial each other; no change needed for local dev
> AllowPrivateIPs still governs whether this node dials private addresses

`p2p_allow_private_ips` is assigned in no committed settings file, so a local node takes the struct-tag default `false` at `:87`. With it false the bus installs a connection gater blocking 10/8, 172.16/12, 192.168/16, 127/8, 169.254/16, 100.64/10, fc00::/7, fe80::/10 and ::1 (go-p2p-message-bus v0.1.23 `client.go:247-286`, installed at `:324-332`). That gater is libp2p's `conngater.BasicConnectionGater`, which blocks both directions from one subnet set: `InterceptAddrDial` for the outbound dial, `InterceptAccept` for the inbound connection (go-libp2p v0.49.0 `p2p/net/conngater/conngater.go:285` and `:310`).

So two peers on one private network running defaults can neither dial nor accept each other. The compose stacks set `p2p_allow_private_ips=true` for precisely this reason (`compose/settings_test.conf:196`).

Suggested replacement for the first bullet:

    - Local dev on one private network needs p2p_allow_private_ips = true. With the shipped default false, libp2p's connection gater refuses RFC1918, loopback and link-local addresses outbound and inbound alike, so private peers can neither dial this node nor be dialled by it

And the `AllowPrivateIPs` longdesc at `:87` needs the matching correction in the same pass, since this PR is what makes it false: *"Both AllowPrivateIPs and SharePrivateAddresses must be true for full local network connectivity"* is no longer true of `SharePrivateAddresses`, which the rest of this PR proves does nothing.

docs/P2P_NAT_TRAVERSAL.md
```diff
@@ -76,14 +76,12 @@ p2p_enable_relay = true
...
 
-For local development or private networks:
-
 ```conf
 # In settings_local.conf
 p2p_share_private_addresses = true
 ```
 
-⚠️ **Warning**: Only use this in controlled environments where peers can actually reach private IPs.
+ℹ️ With the current go-p2p-message-bus this setting has no effect on what is announced: when `p2p_advertise_addresses` is empty, libp2p advertises every bound interface address, private ones included, regardless of this value. Local peers on the same network can therefore already dial each other. To publish only a public address, set `p2p_advertise_addresses` explicitly.
```
freemans13: **SF1 (blocking), same claim as `settings/p2p_settings.go:49`.** "Local peers on the same network can therefore already dial each other" does not hold with the shipped `p2p_allow_private_ips = false`. libp2p's connection gater then refuses RFC1918 in both directions, so this paragraph tells a local-dev reader no change is needed when the change they need is `p2p_allow_private_ips = true`. See the detail in the review body.

settings/p2p_listen_addresses.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	return nil
+}
+
+func validateListenAddress(addr string, port int) error {
+	if !strings.HasPrefix(addr, "/") {
+		ip := net.ParseIP(addr)
+		if ip == nil {
+			return errors.NewConfigurationError("not an IP address or multiaddr")
```
freemans13, line 62: **SF2 (non-blocking).** `0.0.0.0:9905` is the mistake operators will actually make here, and this branch sends them somewhere unhelpful. `net.ParseIP` rejects it, so they get "not an IP address or multiaddr" wrapped in the caller's "use p2p_advertise_addresses or a firewall to control reachability" at `:33`, when what they need is "drop the port, or write `/ip4/0.0.0.0/tcp/9905`".

The repo teaches that format: `legacy_listen_addresses` documents `host:port` with `0.0.0.0:8333` as its worked example (`settings/legacy_settings.go:11`). Anyone configuring both settings in one sitting carries it across.

One branch covers it:

```go
ip := net.ParseIP(addr)
if ip == nil {
    if host, p, splitErr := net.SplitHostPort(addr); splitErr == nil {
        return errors.NewConfigurationError("host:port is not accepted; write %s or /ip4/%s/tcp/%s", host, host, p)
    }

    return errors.NewConfigurationError("not an IP address or multiaddr")
}
```

`SplitHostPort` errors on bare `0.0.0.0` and on bare `::`, so it cannot misfire on an accepted form. It reaches the diagnose cell too, so it pairs with ChiR5.

services/p2p/Server.go
```diff
@@ -410,16 +440,19 @@ func NewServer(
...
 	if p2pPort == 0 {
 		return nil, errors.NewConfigurationError("p2p_port not set in config")
 	}
 
+	// go-p2p-message-bus always binds 0.0.0.0 and :: on p2pPort. Reject any
+	// listen address that asks for something else rather than ignoring it: an
+	// operator who narrowed the bind to one interface must not be left believing
+	// it took effect.
+	if err := settings.ValidateP2PListenAddresses(tSettings.P2P.ListenAddresses, p2pPort); err != nil {
+		return nil, errors.NewConfigurationError("invalid p2p listen configuration", err)
```
freemans13: **SF3 (non-blocking).** This wrap adds a third layer to an error that already describes itself. `ValidateP2PListenAddresses` returns an `*errors.Error` with the configuration code and a message naming both the setting and the offending entry, and `Error()` joins every wrapped cause with `->` (`errors/errors.go:92-131`), so `p2p_listen_addresses = 10.0.1.5` starts the node's last log line three codes deep:

    CONFIGURATION (n): invalid p2p listen configuration -> CONFIGURATION (n): p2p_listen_addresses entry "10.0.1.5" is not supported (...) -> CONFIGURATION (n): binding a specific interface is not supported

`return nil, err` reads better and keeps `TestNewServerRejectsNarrowedListenAddresses` green unchanged: the inner error is already `ErrConfiguration` and already contains every substring the test asserts on. Same family as ChiR5.

freemans13 (changes requested):
Review at commit [94518cc](https://github.com/bsv-blockchain/teranode/pull/1684/commits/94518cc955699d349f6f6cbe908e788706e8a2d4).

The core fix is right, and I checked its premise in the bus source rather than taking the description's word for it. `go-p2p-message-bus` v0.1.23 hardcodes `libp2p.ListenAddrStrings("/ip4/0.0.0.0/tcp/<port>", "/ip6/::/tcp/<port>")` and documents `Port` zero as "a random port will be chosen", with no listen-address option anywhere in `Config` (`client.go:359-361`, `config.go:65-66`). So moving `conf.Port` out of the advertise-address branch fixes a real bug, and refusing a narrower listen value is the honest thing to do until the bus grows the option.

I am requesting changes on one item, SF1. It is the same class as Chi's ChiR1, and the same class as the bug this PR fixes: configuration text that promises something the code does not do. Truthful config documentation is this PR's deliverable, so a new sentence that gets it wrong matters more here than it would elsewhere.

## Blocking

### SF1 - the new share-private guidance sends local-dev operators down a dead end, and contradicts the setting three fields below it

The rewritten `p2p_share_private_addresses` longdesc closes with two claims (`settings/p2p_settings.go:49`):

- *"Local peers on one network can already dial each other; no change needed for local dev"*
- *"AllowPrivateIPs still governs whether this node dials private addresses"*

Both are wrong under the shipped defaults, for exactly the reader that paragraph addresses.

`p2p_allow_private_ips` is assigned in no committed settings file. `settings.conf` mentions it only inside an unrelated comment at `:956`, so a local node takes the struct-tag default, `false` (`settings/p2p_settings.go:87`). The compose stacks set `p2p_allow_private_ips=true` explicitly (`compose/settings_test.conf:196`), which is the tell that local multi-node work needs it.

With it false the bus installs a connection gater blocking 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, 127.0.0.0/8, 169.254.0.0/16, 100.64.0.0/10, fc00::/7, fe80::/10 and ::1 (`client.go:247-286`, installed at `:324-332`). That gater is libp2p's `conngater.BasicConnectionGater`, and it blocks both directions from one blocked-subnet set: `InterceptAddrDial` refuses the outbound dial, `InterceptAccept` refuses the inbound connection (go-libp2p v0.49.0, `p2p/net/conngater/conngater.go:285` and `:310`). The bus's own doc comment says as much, "prevents connections to/from".

So two peers on one private network running shipped defaults can neither dial each other nor accept each other, and the setting the new text hands the reader off to governs inbound as well as outbound.

This file already said the right thing and still does, three settings down: *"Both AllowPrivateIPs and SharePrivateAddresses must be **true** for full local network connectivity"* (`settings/p2p_settings.go:87`). That line is now stale in its own way, because the rest of this PR proves `SharePrivateAddresses` does nothing. The two longdescs disagree, and neither is right on its own.

**Fix.** Name the setting local dev actually needs, and say the gater works in both directions:

    - Local dev on one private network needs p2p_allow_private_ips = true. With the shipped default false, libp2p's connection gater refuses RFC1918, loopback and link-local addresses outbound and inbound alike, so private peers can neither dial this node nor be dialled by it
    - To keep internal addresses private, set AdvertiseAddresses to the public address you want published

Then correct `AllowPrivateIPs` at `:87` in the same pass: *"Both AllowPrivateIPs and SharePrivateAddresses must be true"* becomes *"AllowPrivateIPs must be true for local network connectivity; SharePrivateAddresses does not affect it"*. That line sits outside the diff, but this PR is what turned it from imprecise into false.

The same sentence is in the NAT guide at `docs/P2P_NAT_TRAVERSAL.md:84`, in a paragraph this PR wrote, and needs the same correction.

## Non-blocking

### SF2 - `0.0.0.0:9905` is the mistake operators will make, and the error points them elsewhere

`validateListenAddress` sends any non-multiaddr string that `net.ParseIP` rejects to "not an IP address or multiaddr" (`settings/p2p_listen_addresses.go:44`), and `ValidateP2PListenAddresses` wraps it in "use p2p_advertise_addresses or a firewall to control reachability" (`:33`). Write `p2p_listen_addresses = 0.0.0.0:9905` and you are told about firewalls, when what you need is "drop the port, or write `/ip4/0.0.0.0/tcp/9905`".

That is the mistake worth anticipating because the repo teaches it. `legacy_listen_addresses` documents the `host:port` form with `0.0.0.0:8333` as its worked example (`settings/legacy_settings.go:11`). An operator configuring both settings in one sitting carries the format across.

One branch in the `ip == nil` path covers it:

```go
ip := net.ParseIP(addr)
if ip == nil {
    if host, p, splitErr := net.SplitHostPort(addr); splitErr == nil {
        return errors.NewConfigurationError("host:port is not accepted; write %s or /ip4/%s/tcp/%s", host, host, p)
    }

    return errors.NewConfigurationError("not an IP address or multiaddr")
}
```

`SplitHostPort` errors on both bare `0.0.0.0` and bare `::`, so this cannot misfire on an accepted form. It lands in the diagnose cell too, so it pairs with ChiR5.

### SF3 - the outer wrap at `Server.go:453` adds a third layer to an already self-describing error

`ValidateP2PListenAddresses` returns an `*errors.Error` carrying the configuration code and a message that already names the setting and the offending entry. Wrapping it again in "invalid p2p listen configuration" buys nothing and costs a layer, because `Error()` renders every wrapped cause joined by `->` (`errors/errors.go:92-131`). The startup line for `p2p_listen_addresses = 10.0.1.5` comes out three codes deep:

    CONFIGURATION (n): invalid p2p listen configuration -> CONFIGURATION (n): p2p_listen_addresses entry "10.0.1.5" is not supported (...) -> CONFIGURATION (n): binding a specific interface is not supported

`return nil, err` reads better and keeps `TestNewServerRejectsNarrowedListenAddresses` passing unchanged: the inner error is already `ErrConfiguration` and already contains every substring the test checks. Same family as ChiR5, so fix them together or leave both.

## What I checked and agree with

- **The premise, from the bus rather than the description.** Dual-stack wildcard bind hardcoded on `config.Port`, zero means random, no listen option in `Config`.
- **"SharePrivateAddresses has no effect on announcement" is accurate.** The bus installs an `AddrsFactory` only when `AnnounceAddrs` is non-empty (`client.go:336-350`); with it empty, libp2p announces its bound addresses unfiltered.
- **The regression test earns its place.** `TestBuildP2PMessageBusConfig_MeshProtection/port is set without advertise addresses` asserts `conf.Port == 9906` for both a nil and an empty advertise list, so moving the assignment back inside the old `if` fails it. That is the property that matters, not the call.
- **The new failure mode really is loud.** `p2p.NewServer`'s error returns straight out of `startP2PService`, and the starter loop returns it up, so the daemon exits rather than logging and carrying on (`daemon/daemon_services.go:368-370` and `:128-133`).
- **Mainnet announcement does not change.** `settings.conf:1081` ships `p2p_share_private_addresses = false`, so the old code already took the empty-announce branch there.

One correction to the description while we are here. It says the raw `0.0.0.0` "was copied into `AnnounceAddrs`, which the bus rejects at startup". No committed config does that. The only context shipping `p2p_share_private_addresses = true` is `docker.host` (`compose/settings_test.conf:198`), and it pairs that with a proper multiaddr at `:197`, which the bus parses happily and announces as an undialable wildcard. The crash is real for a hand-written config; nobody should go hunting for a shipped one.

Chi's six findings all hold up against the code and I have nothing to subtract from them.

## What I did not check

I ran no tests locally. CI on 94518cc is green across all 38 checks, including three smoketest shards, seven sequential shards, chainintegrity and legacy-sync, which answers the description's "CI is the first run of those with this change". I did not run a node, so nothing above is evidence about bind or announcement behaviour on a live network.

ctnguyen (changes requested):
Review at commit [94caf9f](https://github.com/gokutheengineer/teranode/commit/94caf9f9944dd9a7181e66c0eeeb2ed3df07cb29)

All six points from the last round are addressed, and addressed thoughtfully rather than minimally — the `listen_mode` longdesc now says plainly that silent removes announcement and not inbound reachability, the reference tables list the third mode, the family-selection wording is gone, and the committed-config guard layers `compose/settings_test.conf` the way the compose stacks mount it, with a `require.Contains(listen, "/tcp/")` pin so the new subtest cannot quietly go vacuous. That last touch is the kind of thing that keeps a guard honest a year from now. The `rootCause` helper is also correct, including the part I expected to be wrong: `As(nil, target)` really does not report false in this errors package, because `isGRPCWrappedError(nil)` is true (`status.FromError(nil)` returns `(nil, true)`) and `UnwrapGRPC` then boxes a typed-nil `*Error` into the interface, so the guard and its comment both earn their place. Two things came out of reading the new commit: a test expectation that the output-format change left behind, and the `host:port` hint, which is right for the one input it is tested on and wrong for the rest.

## Status of previous points

Two of these are FIXED on a narrower reading than the headline suggests, and both are narrower because that is what the item asked for. `ChiR2`'s `Fix` named the documentation route as the cheapest correct option in preference to tightening the validator; the longdesc now states the address family is not selectable and the example label changed to "same bind, written out for both families" (`settings/p2p_settings.go:13`), so the validator is unchanged by design and that is the offered route taken. `ChiR5`'s `Fix` supplied a `Recommended` string, and `cmd/diagnose/config_checks.go:456` adopted it verbatim — which means the `/ip4/0.0.0.0/tcp/0` wording it produces when `p2p_port` is 0 survives, since the snippet itself carried that. Not worth a new id: the same check emits a separate ERROR row telling the operator to set `p2p_port` (`:431-438`). `ChiR3` is fully resolved — the layered lookup resolves `p2p_port.docker.host` and `p2p_listen_addresses.docker.host` to `19905` and `/ip4/0.0.0.0/tcp/19905` for `docker.host.teranode1`, so the pair is really exercised now. `ChiR4` verified with `git grep` over tracked `docs/`: no bracketed `_addresses = [` example remains.

| ChiR1 | Silent mode exposure undocumented | YES | FIXED |
| ChiR2 | Single address family silently ignored | | FIXED |
| ChiR3 | Guard skips compose settings file | | FIXED |
| ChiR4 | NAT guide advertise example invalid | | FIXED |
| ChiR5 | Diagnose output carries error chain | | FIXED |
| ChiR6 | Silent mode log states falsehood | | FIXED |

## Blocking issues

### ChiR7 — The new reason suffix breaks an expectation in the test the same commit edits
**Problem:** The `test` check on this head is red. `TestCheckP2PConfigListenAddresses/narrowed_interface_is_an_error` fails at `cmd/diagnose/config_checks_test.go:79` ([run](https://github.com/bsv-blockchain/teranode/actions/runs/34042058759/job/101510495256)):

  ```text
  FAIL: cmd/diagnose TestCheckP2PConfigListenAddresses/narrowed_interface_is_an_error (0.00s)
      config_checks_test.go:79:
  DONE 13730 tests, 47 skipped, 2 failures in 347.233s
  ```

  The reason is the reason-suffix change: `cmd/diagnose/config_checks.go:446-450` now appends the root cause to the error-path cell, so for `p2p_listen_addresses = 10.0.1.5` at `p2p_port = 9905` the `ConfigResult.Value` is `10.0.1.5 (binding a specific interface is not supported)`, while the table row for that exact input still carries the pre-change expectation — `:65` sets `value: "10.0.1.5"` and `:79` asserts it with `require.Equal` against `res.Value`. The correct string is already asserted two subtests below, at `:92-99`, which the same commit added. The other rows are unaffected: `empty is an error` goes down the `value == ""` branch, which does not append, and `port mismatch is an error` leaves `value` unset.

**Why it matters:** `cmd/diagnose` cannot merge red. The log's second failure is this subtest's parent, so those two lines are the whole of it — nothing else in the PR is failing, and the fix is one line rather than a rethink of the cell format.

**Fix:** Let the new subtest own the full string and drop the stale expectation from the row:

  ```go
  {name: "narrowed interface is an error", addrs: []string{"10.0.1.5"}, severity: SeverityERROR},
  ```

  If you would rather keep the row asserting something, make the table check a prefix match instead of equality, since every error-path cell now begins with the raw value.

## Non-blocking issues

### ChiR8 — The host:port hint names spellings the same validator rejects
**Problem:** `settings/p2p_listen_addresses.go:44-48` fires whenever `net.SplitHostPort` succeeds, and echoes the operator's own host and port back as the two suggested spellings. The branch came in from @freemans13's SF2 snippet, which carried the same `/ip4/%s`, so this refines that suggestion rather than disputing it. The hint is correct for the single input the table pins (`settings/p2p_listen_addresses_test.go:49`, `0.0.0.0:9905` at `p2p_port` 9905). For everything else that reaches it, one or both halves name a value this same function refuses:

  - `127.0.0.1:9905` or `10.0.1.5:9905` yields "write 127.0.0.1 or /ip4/127.0.0.1/tcp/9905"; both halves then fail at `:52-54` and `:72-74` with "binding a specific interface is not supported".
  - `[::]:9905` splits to host `::`, so the hint reads "/ip4/::/tcp/9905". go-multiaddr's ip4 transcoder parses with `net.ParseIP(s).To4()` (`go-multiaddr@v0.16.1/transcoders.go:84-87`), so that fails at `:58-61` with "invalid multiaddr". The bare `::` half is fine.
  - `localhost:9905` or `foo:bar` — `SplitHostPort` does not require a numeric port — yields "write foo or /ip4/foo/tcp/bar", where neither half is a usable value.
  - `0.0.0.0:1234` at `p2p_port = 9905` echoes 1234 into the multiaddr, which is then refused at `:80-82` for the port mismatch.

**Why it matters:** The string also reaches the operator-facing diagnose cell through `rootCause` (`cmd/diagnose/config_checks.go:449`), so `p2p_listen_addresses = 127.0.0.1:9905` prints "write 127.0.0.1" in the very table whose job is to tell the operator that 127.0.0.1 is not accepted. Nothing is lost — the failure stays loud and the second attempt gets the accurate "binding a specific interface is not supported" message — but a hint whose whole purpose is to shorten the operator's loop instead adds a bounce, and `127.0.0.1` is the value this PR deliberately deleted from `settings.conf`, so it is the one it can least afford to recommend.

**Fix:** Emit the multiaddr half only when the host really is the wildcard, pick the family from the parsed host, and use `p2p_port` rather than the port the operator typed:

  ```go
  if host, _, splitErr := net.SplitHostPort(addr); splitErr == nil {
      if hostIP := net.ParseIP(host); hostIP != nil && hostIP.IsUnspecified() {
          proto := "ip4"
          if hostIP.To4() == nil {
              proto = "ip6"
          }

          return errors.NewConfigurationError("host:port is not accepted; write %s or /%s/%s/tcp/%d", host, proto, host, port)
      }

      return errors.NewConfigurationError("host:port is not accepted, and %q is not the wildcard bind; write 0.0.0.0 or /ip4/0.0.0.0/tcp/%d", host, port)
  }
  ```

  Worth adding `127.0.0.1:9905`, `[::]:9905` and `localhost:9905` to the rejected table next to the existing `0.0.0.0:9905` row, since it is the single accepted-host case that hid this.

## Recap

| ChiR7 | Diagnose test expectation left stale | YES | 65% |
| ChiR8 | Hint names rejected listen values | | 40% |

ctnguyen (approved):
Review at commit [bb783fdc](https://github.com/gokutheengineer/teranode/commit/bb783fdce7332dc226c8d7a802e2a283e14d79b0)

Both points from the last round are fixed, and fixed the way they were asked for rather than approximately: the stale `value:` field is gone from the diagnose table row so the dedicated subtest at `cmd/diagnose/config_checks_test.go:92-99` owns the full string, and the `host:port` hint now echoes the operator's host back only when it really is the wildcard, always naming `p2p_port` instead of whatever port was typed. The four rejected-table rows at `settings/p2p_listen_addresses_test.go:50-53` pin exactly the inputs that hid the old defect, including the `[::]:9905` case that used to be told to write `/ip4/::/tcp/9905`. CI on this head is green across all 37 check-runs, so the red `test` job that ChiR7 cited is resolved too. The unreviewed delta is three files and about twenty lines, and it is clean.

The two new items below are both in the documentation half, which is where this PR's remaining exposure is: `docs/P2P_NAT_TRAVERSAL.md` still carries a section describing the behaviour this PR removed, and the `host:port` branch has one spelling left where the diagnosis half of the message is wrong. Neither blocks merge.

## Status of previous points

Two of these were recorded as FIXED on a deliberately narrow reading in round 2, and both caveats still hold on this head, so they are repeated rather than widened. `ChiR2` was fixed by the documentation route its own `Fix` named as the cheapest correct option — the longdesc at `settings/p2p_settings.go:13` now states the address family is not selectable and labels the two-entry example "same bind, written out for both families" — so `validateListenAddress` still accepts a single-family entry, by design. `ChiR5`'s `Recommended` string was adopted verbatim from `ChiR5`'s own snippet (`cmd/diagnose/config_checks.go:456`), so the `/ip4/0.0.0.0/tcp/0` wording it produces when `p2p_port` is 0 survives; round 2 ruled that not worth a new id because the same check emits a separate ERROR row telling the operator to set `p2p_port` (`config_checks.go:431-438`), and that row is still there. `ChiR4` re-verified with `git grep` over tracked `docs/`, `settings.conf`, `compose/` and `deploy/`: no bracketed `_addresses = [` example remains. `ChiR3`'s guard still layers `compose/settings_test.conf` over `settings.conf` for the three `docker.host.teranodeN` contexts (`settings/p2p_listen_addresses_test.go:113-121`) and still carries the `require.Contains(t, listen, "/tcp/")` pin at `:119`, so it cannot go vacuous.

| ChiR1 | Silent mode exposure undocumented | YES | FIXED |
| ChiR2 | Single address family silently ignored | | FIXED |
| ChiR3 | Guard skips compose settings file | | FIXED |
| ChiR4 | NAT guide advertise example invalid | | FIXED |
| ChiR5 | Diagnose output carries error chain | | FIXED |
| ChiR6 | Silent mode log states falsehood | | FIXED |
| ChiR7 | Diagnose test expectation left stale | YES | FIXED |
| ChiR8 | Hint names rejected listen values | | FIXED |

## Non-blocking issues

### ChiR9 — The NAT guide still explains the mechanism this PR deleted, and now contradicts its own corrected paragraph

**Problem:** This PR edits `docs/P2P_NAT_TRAVERSAL.md` in two places — the advertise example at `:47` and the Option 5 warning, replaced by a new paragraph at `:84`. That new paragraph states:

  `docs/P2P_NAT_TRAVERSAL.md:84:  ... when p2p_advertise_addresses is empty, libp2p advertises every bound interface address, private ones included, regardless of this value.`

The document's own Problem statement and Implementation Details section, both untouched, assert the opposite:

  `docs/P2P_NAT_TRAVERSAL.md:8:  2. go-p2p v1.1.18 aggressively filters private IPs when advertiseAddresses is empty`
  `docs/P2P_NAT_TRAVERSAL.md:10:  4. Nodes end up invisible to the network`
  `docs/P2P_NAT_TRAVERSAL.md:143:  The go-p2p library AddrsFactory aggressively filters private IPs when advertiseAddresses is empty. This prevents the observed address mechanism from working because:`
  `docs/P2P_NAT_TRAVERSAL.md:145:  - The node has no addresses to advertise`
  `docs/P2P_NAT_TRAVERSAL.md:149:  The solution is to always pass listen addresses to override this filtering, allowing the observed address mechanism to function properly.`

Two problems, plus one line that needs rewording, all inside the same file the diff is already open in:

1. `:8-10` and `:143-147` are the direct negation of `:84`, of the `SharePrivateAddresses` longdesc rewritten at `settings/p2p_settings.go:49`, of `docs/references/services/p2p_reference.md:609`, and of the new `resolveAdvertiseAddresses` doc comment at `services/p2p/Server.go:388-391` ("An empty result leaves announcement to libp2p, which advertises the interface addresses it actually bound (private ones included) and whatever public address peers observe via Identify"). One of the two readings is wrong and this PR asserts, in four places, that it is the one in the unedited text.
2. `:149`'s prescribed remedy — pass listen addresses to override the filtering — was implemented by precisely the branch this diff removes from `NewServer`: `else if tSettings.P2P.SharePrivateAddresses { advertiseAddresses = listenAddresses }`. Nothing puts the configured `p2p_listen_addresses` into `AnnounceAddrs` any more, and no setting can restore that, so following `:149` is now a no-op. (An operator can still put a wildcard into the announce set the long way round, by naming it in `p2p_advertise_addresses`, which `resolveAdvertiseAddresses` returns at `services/p2p/Server.go:408-410` and `buildP2PMessageBusConfig` assigns unvalidated at `:381-382` — but that is not what the guide's remedy describes, and the PR's Residual already flags advertise entries as reaching the bus unvalidated.)
3. `:130` ("Nodes advertise their listen addresses (including private IPs) to enable libp2p's observed address detection") is right in substance and wrong in mechanism. Private interface addresses really are advertised, but they are advertised because libp2p announces what it bound, not because `p2p_listen_addresses` was passed through — and that setting is now wildcard-only and never announced, so the phrase "their listen addresses" points the reader at a setting that cannot affect announcement at all.

Verified with `git grep`: the stale `AddrsFactory` / `v1.1.18` / "invisible to the network" claims are confined to this one tracked file, so the fix is local. Two neighbouring lines are *not* part of this — `:135` ("Nodes behind NAT advertise their private addresses") is still accurate on this head, and `:139`'s peer-list filtering is a separate `GetPeers`-side mechanism this diff does not touch.

**Why it matters:** This is the guide an operator behind NAT reads, the same audience `ChiR4` was about, and it is the file where the PR chose to correct one paragraph. An operator who follows `:149` has nothing to do: no setting passes the configured listen addresses to the announce set any more, and the only listen value that starts the node is a wildcard. Worse, an operator who reads `:8-10` and then `:84` gets two incompatible statements about what happens with an empty `p2p_advertise_addresses` seventy-six lines apart, and no way to tell which is current — which is exactly the "documentation promising behaviour the code does not have" failure this PR set out to remove. The behaviour change is the PR's own (see [bitcoin-sv/teranode#4758](https://github.com/bitcoin-sv/teranode/issues/4758) finding 2 and the PR's Residual note that `p2p_share_private_addresses` needs a bus-side address filter to mean anything), so the disclosure in Residual covers the setting being inert but not this file still teaching the removed mechanism as the fix.

**Fix:** Retire the two stale blocks. At `:8-10`, state the current premise instead:

  ```text
  2. libp2p advertises the interface addresses it bound, private ones included, when
     p2p_advertise_addresses is empty
  3. Behind NAT those are private addresses no external peer can dial, so the node
     depends on the observed address peers report via Identify
  ```

Replace the "The go-p2p v1.1.18 Issue" section at `:141-149` with what actually governs reachability now — a public `p2p_advertise_addresses` entry, or `p2p_allow_private_ips = true` on a private network. Then reword `:130` rather than deleting it, so it attributes the advertisement to the right mechanism:

  `docs/P2P_NAT_TRAVERSAL.md:130:  1. **Advertisement Stage**: Nodes advertise the interface addresses libp2p bound (including private IPs) to enable libp2p's observed address detection`

If you would rather not rewrite the section in this PR, the minimum that stops it being actively misleading is deleting `:149` and the `:141-147` block and adding a one-line pointer to `:84`.

### ChiR10 — The bare `:port` spelling lands in the wrong half of the new hint and is told it is not the wildcard

**Problem:** The `host:port` branch came in from @freemans13's `SF2` snippet, and its current wording was taken character-for-character from `ChiR8`'s own suggested `Fix`; this item revises that advice for the one input it got wrong. `settings/p2p_listen_addresses.go:49-60` routes on whether `net.ParseIP(host)` yields an unspecified address. For `p2p_listen_addresses = :9905`, `net.SplitHostPort(":9905")` succeeds with `host == ""` — SplitHostPort does not require a host, which is why `net.Listen("tcp", ":9905")` works — and `net.ParseIP("")` is nil, so `hostIP != nil && hostIP.IsUnspecified()` is false and the input falls through to `:59`:

  ```text
  p2p_listen_addresses entry ":9905" is not supported (libp2p always binds 0.0.0.0 and ::
  on p2p_port 9905; use p2p_advertise_addresses or a firewall to control reachability) ->
  host:port is not accepted, and "" is not the wildcard bind; write 0.0.0.0 or /ip4/0.0.0.0/tcp/9905
  ```

The advice half is right. The diagnosis half is not: `:9905` is the wildcard bind, on exactly `p2p_port`, and the empty `%q` renders as a bare `""` that names no host at all. Traced by reading `:49-60` against Go's documented `SplitHostPort`/`ParseIP` behaviour; the four sibling rows added in this commit pin the other inputs but none covers an empty host.

**Why it matters:** `:port` is not a hypothetical slip in this repo — it is the idiom `settings.conf` teaches for "bind all interfaces", in the same P2P block, immediately above the setting that now rejects it. `git grep` over `settings.conf` finds five committed values in that form: `p2p_grpcListenAddress.docker.m|docker.ss|operator` at `:1019-1021` and `p2p_httpListenAddress` plus its `docker.m` override at `:1031` and `:1034`, twelve lines above `p2p_listen_addresses = 0.0.0.0` at `:1043`. `HTTPListenAddress`'s own longdesc documents it explicitly under `### Format` as "**:port** - Binds to all interfaces". That is the same argument shape as `SF2`'s `legacy_listen_addresses` premise, one setting closer. The string also reaches the operator-facing diagnose cell through `rootCause` (`cmd/diagnose/config_checks.go:449`), so the cell reads `:9905 (host:port is not accepted, and "" is not the wildcard bind; ...)`. Nothing is lost — the failure stays loud, the outer message still says what the bus binds, and the suggested spelling is correct — but the operator is told the wildcard they wrote is not the wildcard, which costs them the loop the hint exists to shorten.

**Fix:** Treat an empty host as the wildcard and route it to the same branch, spelling the suggestion `0.0.0.0` since there is no host to echo. Guarding on `host == ""` first keeps `hostIP` off the nil path:

  ```go
  if host, _, splitErr := net.SplitHostPort(addr); splitErr == nil {
      hostIP := net.ParseIP(host)
      if host == "" || (hostIP != nil && hostIP.IsUnspecified()) {
          proto, shown := "ip4", host
          switch {
          case shown == "":
              shown = "0.0.0.0"
          case hostIP.To4() == nil:
              proto = "ip6"
          }

          return errors.NewConfigurationError("host:port is not accepted; write %s or /%s/%s/tcp/%d", shown, proto, shown, port)
      }

      return errors.NewConfigurationError("host:port is not accepted, and %q is not the wildcard bind; write 0.0.0.0 or /ip4/0.0.0.0/tcp/%d", host, port)
  }
  ```

Worth a fifth row alongside the four this commit added:

  ```go
  {"bare port is the wildcard shorthand", []string{":9905"}, "write 0.0.0.0 or /ip4/0.0.0.0/tcp/9905"},
  ```

Accepting `:<p2p_port>` outright as a valid spelling of the wildcard bind would also be defensible and closer to what the rest of the P2P block teaches, but it is a scope increase — it would want the same treatment in `ValidateP2PListenAddresses`' accepted set and its longdesc — so the message fix is the cheaper correct move here.

## Recap

| ChiR9 | NAT guide teaches removed mechanism | | 24% |
| ChiR10 | Bare port form misdiagnosed | | 20% |

docs/P2P_NAT_TRAVERSAL.md
```diff
@@ -76,14 +76,12 @@ p2p_enable_relay = true
...
 
-For local development or private networks:
-
 ```conf
 # In settings_local.conf
 p2p_share_private_addresses = true
 ```
 
-⚠️ **Warning**: Only use this in controlled environments where peers can actually reach private IPs.
+ℹ️ With the current go-p2p-message-bus this setting has no effect on what is announced: when `p2p_advertise_addresses` is empty, libp2p advertises every bound interface address, private ones included, regardless of this value. What local dev on a private network does need is `p2p_allow_private_ips = true`: with the shipped default `false`, libp2p's connection gater refuses RFC1918, loopback and link-local addresses in both directions, so private peers can neither dial this node nor be dialled by it. To publish only a public address, set `p2p_advertise_addresses` explicitly.
```
freemans13, line 84: **SF4 (non-blocking).** This paragraph is a real correction, and it sits 10 to 50 lines below three options that are entirely inert. `docs/P2P_NAT_TRAVERSAL.md` is the only file in the repository that mentions `p2p_enable_nat_service` (`:32`, `:71`, `:121`), `p2p_enable_nat_port_map` (`:62`, `:72`, `:122`), `p2p_enable_hole_punching` (`:73`) or `p2p_enable_relay` (`:74`). None is read by `settings/settings.go`, none has a struct tag, none appears in `settings.conf`.

They were deleted by `fb3d000c` (#&NoBreak;4135, "Simplify p2p server", 3 October 2025), which replaced the first three with the single `p2p_enable_nat` gate and made relay unconditional. The guide was never updated, so Options 1, 3 and 4 tell an operator behind NAT to write four keys that gocore ignores in silence, while `p2p_enable_nat` stays at its shipped `false`. The old defaults were `true`, so the reader is left believing they turned on something that shipped off.

See the review body for the suggested replacement, which is one setting plus the shared-hosting warning that `settings/p2p_settings.go:78` carries and Option 4 does not.

freemans13 (approved):
Review at commit [bb783fdc](https://github.com/bsv-blockchain/teranode/pull/1684/commits/bb783fdce7332dc226c8d7a802e2a283e14d79b0).

Approving, which lifts the `CHANGES_REQUESTED` I left on 94518cc. All three of my previous items are fixed at the source rather than papered over, and I re-checked the PR's premise against the message-bus library on this head instead of carrying the earlier verification forward.

## Status of my previous points

| ID | Previous point | Blocking | Status |
| SF1 | share-private guidance sends local dev to a dead end | YES | FIXED |
| SF2 | `0.0.0.0:9905` slip points at firewalls | | FIXED |
| SF3 | third error layer at the `NewServer` call site | | FIXED |

`SF1` is fixed in all three places it lived. The `SharePrivateAddresses` longdesc now names `p2p_allow_private_ips` and says the gater blocks both directions (`settings/p2p_settings.go:49`), the `AllowPrivateIPs` longdesc no longer claims `SharePrivateAddresses` is needed for local connectivity (`:87`), and the NAT guide paragraph carries the same correction (`docs/P2P_NAT_TRAVERSAL.md:84`). `SF2` landed and was then improved twice past my snippet, by `ChiR8` and `ChiR10`, which is the right direction. `SF3` is `return nil, err` now (`services/p2p/Server.go:471-473`), so the startup line is two codes deep rather than three.

## Non-blocking

### SF4 - three of the NAT guide's five options tell operators to set settings that were deleted eleven months ago

`docs/P2P_NAT_TRAVERSAL.md` names four p2p configuration keys that exist nowhere else in the repository:

- `p2p_enable_nat_service` at `:32` (Option 1), `:71` (Option 4) and `:121`
- `p2p_enable_nat_port_map` at `:62` (Option 3), `:72` and `:122`
- `p2p_enable_hole_punching` at `:73`
- `p2p_enable_relay` at `:74`

`git grep` finds each of them in that one file and nowhere else. None appears in `settings/settings.go`, which is the only place a key is actually read, and none has a struct tag or a `settings.conf` line. gocore ignores an unknown key silently, so an operator behind NAT who follows Option 1, 3 or 4 writes four settings that do nothing at all.

They were real once, which is why they read so plausibly. Commit `19162edc` (#&NoBreak;3640, "add P2P config for NAT, Hole punching a relay") introduced them, and `fb3d000c` (#&NoBreak;4135, "Simplify p2p server", 3 October 2025) deleted that whole batch. The behaviour did not go away, the switches did: go-p2p-message-bus v0.1.23 gates `libp2p.NATPortMap()`, `EnableNATService()` and `EnableHolePunching()` on a single `config.EnableNAT` (`client.go:365-378`), and calls `EnableRelay()` unconditionally (`:380-382`). Teranode feeds `config.EnableNAT` from `p2p_enable_nat` at `services/p2p/Server.go:298`. So Option 4's four lines collapse to one real setting plus a relay that is always on, and Options 1 and 3 collapse to the same setting.

Two things make this worse than ordinary doc rot.

The old defaults were `true` and the replacement default is `false` (`settings/settings.go:553`). The guide therefore does not merely fail to help. It leaves a reader who followed it believing they enabled something that shipped off, which is the exact failure mode this PR set out to remove.

And the one real replacement carries a warning the guide does not. `p2p_enable_nat`'s own longdesc calls the UPnP gateway scan a **CRITICAL RISK** on shared hosting, up to account suspension, and says cloud and VPS deployments should always leave it false (`settings/p2p_settings.go:78`). Option 4 sells the same behaviour as a "Full NAT Traversal Suite" for "maximum connectivity" with no caveat at all.

**Fix.** Fold Options 1, 3 and 4 into the one switch that exists, and carry the warning across from the setting's own description:

```conf
# In settings_local.conf
# UPnP/NAT-PMP port mapping, the AutoNAT service and hole punching are one switch.
# Circuit relay is always on and needs no setting.
# Do not enable this on cloud or shared hosting: the gateway scan reads as network
# reconnaissance and has produced abuse reports.
p2p_enable_nat = true
```

`:121-122` need the same correction. `:108` already greps for `p2p_enable_nat`, so it is the one part of that section that survives unchanged.

Non-blocking because it predates this PR by eleven months and the code half here is correct and green. But it is the same defect class the PR exists to remove, in a file the PR already has open, and it pairs with `ChiR9`, which covers the stale go-p2p v1.1.18 filtering story at `:8-10`, `:130` and `:141-149`. Taken together, everything in that guide except Option 2 and Option 5 is now wrong, and one more docs commit here would finish it.

## What I verified on this head, rather than carrying forward

- **The premise still holds.** `go.mod:202` still pins go-p2p-message-bus v0.1.23, and `client.go:359-363` still hardcodes `libp2p.ListenAddrStrings("/ip4/0.0.0.0/tcp/<Port>", "/ip6/::/tcp/<Port>")`, with `Config.Port` documented as "if zero, a random port will be chosen" (`config.go:65-66`). The validator's accepted set is exactly that pair, so it neither over- nor under-constrains the bind.
- **"libp2p announces bound addresses, private ones included" is accurate.** This PR now asserts it in five places, so it is worth being sure. The bus installs an `AddrsFactory` only inside `if len(config.AnnounceAddrs) > 0` (`client.go:336-350`), so with the announce set empty libp2p's default identity factory applies and nothing is filtered. The private-IP connection gater (`client.go:324-333`) blocks connections rather than announcements, so it does not qualify the claim.
- **Silent mode's new firewall advice is not undercut by circuit relay.** The bus calls `EnableRelay()` and `EnableAutoRelayWithStaticRelays()` unconditionally, which made me check whether a relayed inbound connection could arrive past a firewall on `p2p_port`. It cannot today: the relay candidates are the bootstrap peers (`client.go:231-233`), those are other teranode nodes running this same bus, and the bus never calls `libp2p.EnableRelayService()`, so no teranode node grants a reservation. Worth recording because the new `listen_mode` text is only true while that stays so.
- **The empty-entry branch is live, not dead code.** gocore's `GetMulti` splits and trims but does not drop empty elements (`gocore@v1.0.82/config.go:700-703`), so a trailing `|` really does reach `p2p_listen_addresses contains an empty entry`.
- **The end state is tested, not just the config value.** `TestBuildP2PMessageBusConfig_MeshProtection/port is set without advertise addresses` asserts `conf.Port`, which is one step short of a bind. The bind itself is covered in CI: `daemon.TestDaemon.ConnectToPeer` dials `/dns/127.0.0.1/tcp/<P2P.Port>/p2p/<PeerID>` (`daemon/test_daemon.go:2299`) and fails the test if the peer never shows up in `getpeerinfo`, and `reorg_test`, `smoke_test` and `bsv_invalidblockrequest_test` all use it. Before this change that dial had nothing listening, so those suites are the real regression test for the headline fix.
- **The committed blast radius is two values and both are guarded.** `git grep` over the tracked tree finds `p2p_listen_addresses` assigned only at `settings.conf:1043` and `compose/settings_test.conf:197`. Nothing under `deploy/` sets it, so `ChiR3`'s layered guard covers everything shipped.
- **Multi-node in-process tests cannot collide on the now-deterministic port.** `daemon/test_daemon.go:259-264` allocates each node a free port before assigning `P2P.Port`.

```
SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./settings/ ./cmd/diagnose/   ok
SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./services/p2p/               ok (14.4s)
```

CI on bb783fdc is 36 green and one skipped (`test aerospike native-ops (fork image)`), so the daemon, smoke and chainintegrity suites that exercise the real bind have run, just not on my machine. I did not run a node, so nothing above is evidence about bind or announcement behaviour on a live network. `ChiR9` and `ChiR10` are both still open on this head and I have nothing to subtract from either.

PR #1693, opened 2026-09-04
fix(blockchain): honor fresh-node boot state

### LINKED ISSUE #1671: Fresh-node boot into CATCHINGBLOCKS is wrong for seeded production nodes; wire blockchain_initializeNodeInState
## Problem

Since #1135 a fresh node (no persisted FSM state) boots straight into `CATCHINGBLOCKS`, hardcoded, with no supported way for an operator to ask for a quiescent start:

```go
// services/blockchain/Server.go, Init()
if stateStr == "" { // no persisted state: this is a fresh node
    b.finiteStateMachine.SetState(blockchain_api.FSMStateType_CATCHINGBLOCKS.String())
    b.logger.Infof("[Blockchain][Init] fresh node, booting FSM into %v (catch-up mode)", b.finiteStateMachine.Current())
    b.store.SetFSMState(ctx, b.finiteStateMachine.Current())
}
```

That is the right default for the case #1135 targeted — a fresh install syncing from the network, where the old `IDLE` default cost several minutes of doing nothing. It is the wrong default for a **seeded** production node.

## Why seeding is different

The documented seeding flow runs with the stack **down** (`docs/howto/miners/docker/minersHowToSyncTheNode.md`):

```bash
./stop.sh
./clean.sh --data-only
./seed.sh <block-hash> [dir|url]
./start.sh
```

`cmd/seeder` writes the UTXO set, headers, and `state["BlockAssembler"]` directly against the stores. It does **not** write `state["FSMState"]` — there are no `SetFSMState` calls in `cmd/seeder/seeder.go`. So the first `./start.sh` after a seed hits exactly the `stateStr == ""` branch above and the node begins dialing peers and fetching blocks immediately.

The operator has just written a multi-GB UTXO snapshot from an external source into a production node's stores, and gets no window to confirm it landed correctly before the node starts consuming network and mutating that state. Before #1135 the `IDLE` default gave that window for free: start the stack, verify, then `setfsmstate --fsmstate running`.

Things an operator reasonably wants to check post-seed, pre-catchup:

- the seeded tip height/hash matches the snapshot they intended to load
- `state["BlockAssembler"]` points where they expect
- services come up healthy and connect to their stores
- config/context is right for the network the snapshot belongs to (a testnet snapshot on a mainnet context is a real mistake to make)

## Why there is no workaround today

Neither existing lever is usable:

1. **`blockchain_initializeNodeInState`** — the setting exists and is plumbed into `settings.Settings`, but is referenced nowhere else. Its own `longdesc` says so:

   > **UNUSED** — This setting is defined but not referenced anywhere in the codebase. This setting was intended to force the blockchain service to initialize in a specific FSM state (IDLE, RUNNING, CATCHINGBLOCKS) instead of using default initialization logic, but it is not currently implemented.

   ```
   settings/settings.go:329:  InitializeNodeInState: getString("blockchain_initializeNodeInState", "", alternativeContext...),
   ```

   Read, never used.

2. **`-localTestStartFromState`** (`daemon/daemon_services.go:256-275`) — the escape hatch the setting's note points at. The name says `localTest`; it is not documented for operators and not something to put in a production runbook.

And once the node is up in `CATCHINGBLOCKS` there is no way back to `IDLE`: the FSM has no `CATCHINGBLOCKS -> IDLE` edge (#1670). The only route is via `RUNNING`, which switches on live subtree validation and block-assembly tx feeding — on a node whose seed has not been verified yet. That is the opposite of what the operator is asking for.

## Proposed fix

Wire the setting that already exists. `blockchain_initializeNodeInState` is the natural home:

- When set to a valid state name, `Init()` uses it for the fresh-node (`stateStr == ""`) branch instead of the hardcoded `CATCHINGBLOCKS`.
- When unset, keep #1135's `CATCHINGBLOCKS` default — a fresh network-sync install should not regress.
- Reject an invalid value loudly at startup rather than silently falling through.
- Update the `longdesc`, which currently documents the setting as unused.

That gives seeded deployments `blockchain_initializeNodeInState = IDLE` in their settings context, and leaves every other deployment unchanged.

Worth deciding as part of this:

- Whether the seeding tooling should write `state["FSMState"] = IDLE` itself, so a seeded store is self-describing and no separate setting is needed. Arguably cleaner: the seeder knows it just seeded, and the operator does not have to remember a config key. It does mean `cmd/seeder` starts owning FSM state, which it deliberately does not today.
- Whether `-localTestStartFromState` should be folded into the setting rather than living alongside it.

## Relationship to #1670

Independent but compounding, and both worth fixing:

- #1670 — add a `CATCHINGBLOCKS -> IDLE` edge, so a node already up can be quiesced.
- this issue — let a seeded node come up quiescent in the first place, so the operator does not need to quiesce it after the fact.

Fixing only #1670 leaves a race: the node still starts catching up at boot and the operator is chasing it. Fixing only this one leaves any node that has already entered `CATCHINGBLOCKS` stuck.

## Acceptance

- [ ] `blockchain_initializeNodeInState = IDLE` makes a fresh node boot into `IDLE` and stay there
- [ ] Unset preserves #1135 behaviour (fresh node boots `CATCHINGBLOCKS`)
- [ ] An invalid value fails startup with a clear error instead of being ignored
- [ ] A node with persisted FSM state still restores that state, ignoring the setting
- [ ] Seeded-node path covered by a test: seed writes `state["BlockAssembler"]` but no FSM state, node boots to the configured state
- [ ] `longdesc` for `blockchain_initializeNodeInState` no longer says UNUSED
- [ ] Seeding docs (`docs/howto/miners/docker/minersHowToSyncTheNode.md`, kubernetes equivalent) show the verify-before-catchup flow

## Refs

- #1135 — hardcoded the fresh-node boot state to `CATCHINGBLOCKS`
- #1670 — no `CATCHINGBLOCKS -> IDLE` edge
- #1346 — operator runbook for `teranode-cli rewindblockchain` (same "get to IDLE" problem)

oskarszoon:
Fresh production nodes previously began catchup immediately after seeding, before operators could inspect imported data. This selects the configured boot state only when no FSM state is stored, defaulting `operator` and `docker.m` to IDLE and retaining automatic CATCHINGBLOCKS elsewhere.

- Persisted state takes precedence without validating unused boot configuration. Fresh-node values remain uppercase IDLE/CATCHINGBLOCKS/RUNNING; invalid values abort startup.
- Configured and restored RUNNING remain checkpoint-gated. Configured RUNNING below checkpoint aborts without fallback. Persisted RUNNING migrates to CATCHINGBLOCKS only after a successful below-checkpoint height comparison.
- Tip-read failures or missing metadata abort startup without rewriting persisted state. FSM-state read/write failures also abort.
- Unrecognized persisted names fail startup without writes instead of entering a state with no outgoing transitions. The known LEGACYSYNCING migration remains supported.
- Operator docs explain explicit catchup, configured RUNNING's seeded-tip requirement, unknown-state recovery and unavailable Docker listeners during the IDLE inspection window.

The seeder does not write FSM state. Runtime RUN behavior and checkpoint enforcement remain unchanged. Replaces the boot-policy half of #1673; checkpoint prerequisite #1685 is merged.

Validation: regressions failed before both corrections. Final blockchain/settings package tests passed with race detection (39.959s / 2.669s); package vet and changed-code lint passed; strict MkDocs passed (8.62s). Narrow independent checks found no blockers. Final-head `make test` passed: 13,744 tests, 46 skipped, 504.698s, with race detection and repository-wide coverage. Final CI follow-up: 37 successful checks and one expected skip; Sonar passed with zero new issues. Claude reviewed the final fixes with no new issues; Copilot could not review the final push because its quota was exhausted. Human re-review remains requested. Docker healthcheck behavior was checked in source, not on a running deployment.

Closes #1671


services/blockchain/Server.go
```diff
@@ -407,28 +407,29 @@ func (b *Blockchain) Init(ctx context.Context) error {
 		return nil
 	}
 
+	bootState, err := b.fsmBootState()
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
oskarszoon: Updated the `Init` godoc in eb8c741ee to document configured fresh-node boot selection and both RUNNING checkpoint-gate outcomes. Invalid configured boot state remains intentional fail-fast behavior, covered by `TestInit_InvalidConfiguredBootStateFailsWithPersistedState`.

services/blockchain/Server.go
```diff
@@ -407,28 +407,29 @@ func (b *Blockchain) Init(ctx context.Context) error {
...
+
 	// Set the FSM to the latest persisted state
 	stateStr, err := b.store.GetFSMState(ctx)
 	if err != nil {
-		b.logger.Errorf("[Blockchain][Init] Error getting FSM state: %v", err)
+		return errors.NewStorageError("[Blockchain][Init] failed to get persisted FSM state: %v", err)
 	}
```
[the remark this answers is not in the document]
oskarszoon: Fixed in eb8c741ee. Removed the orphaned `%v` verbs from all three `NewStorageError` calls while retaining trailing errors for wrapping. Added regression assertions covering literal `%v` and `%!v(MISSING)` output.

docs/howto/miners/minersHowToInteractWithFSM.md
```diff
@@ -1,6 +1,16 @@
...
+deployments (`operator` and `docker.m` settings contexts) start in `IDLE`, giving
+an operator a safe inspection window. Other contexts keep the automatic
+`CATCHINGBLOCKS` default. Configure this with
+`blockchain_initializeNodeInState`; it accepts `IDLE`, `CATCHINGBLOCKS`, or
+`RUNNING`, and invalid values fail startup.
+
+The setting applies only when no FSM state is persisted. Restarts normally
+restore the persisted state. A persisted `RUNNING` state that no longer passes
+the active network's checkpoint gate is safely persisted and resumed as
+`CATCHINGBLOCKS` instead.
```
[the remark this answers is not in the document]
oskarszoon: Added the explicit first-run instruction in 93de83d40 next to the production `IDLE` explanation. It directs operators to `teranode-cli setfsmstate --fsmstate catchingblocks`, explains that `RUNNING` is refused below the checkpoint, and notes automatic promotion after catch-up.

services/blockchain/Server.go
```diff
@@ -453,6 +468,29 @@ func (b *Blockchain) Init(ctx context.Context) error {
...
+	configured := ""
+	if b.settings != nil {
+		configured = strings.TrimSpace(b.settings.BlockChain.InitializeNodeInState)
+	}
+
+	if configured == "" {
+		return blockchain_api.FSMStateType_CATCHINGBLOCKS.String(), nil
+	}
+
+	switch configured {
```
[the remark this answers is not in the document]
oskarszoon: Keeping uppercase-only configuration values intentionally. Issue #1671 and the settings contract define exact `IDLE`, `CATCHINGBLOCKS`, and `RUNNING` values, and `TestInit_InvalidConfiguredBootStateFails` pins lowercase forms as invalid. CLI input normalization does not define the configuration-file contract; the startup error lists all accepted values.

icellan (changes requested):
Direction is right — a fresh node shouldn't sit in IDLE, and RUNNING should be checkpoint-gated. But the boot block conflates "tip is below the checkpoint" with "couldn't read the tip", and the new config value can abort `Init` on nodes where it's irrelevant. Two of these are fleet-wide availability risks, not edge cases.

**`services/blockchain/Server.go:451` — a transient store error durably demotes persisted RUNNING.**
`guardRunBelowHighestCheckpoint` fails closed on store errors as well as on a low tip: `GetBestBlockHeader` erroring (Postgres blip, or `blockchain_store_dbTimeoutMillis` expiring against a cold pool — i.e. exactly at startup) and `meta == nil` both return a `StateError` (Server.go:2978-2984). The persisted branch treats *any* error as "unsafe RUNNING" and then **persists** CATCHINGBLOCKS. A synced mainnet miner restarts, its first tip read times out, persisted RUNNING is overwritten — mining and live validation stay off until some later catchup happens to hit `restoreFSMState` (blockvalidation/catchup.go:1421). The rest of this PR takes the opposite posture for `GetFSMState`/`SetFSMState`: DB error aborts `Init`. Do the same here — return the error on a read failure, demote only on a genuine height comparison. `TestInit_PersistedRunningHonorsCheckpointGate/"missing tip resumes catchup"` currently locks the wrong behaviour in.

**`services/blockchain/Server.go:478` (+ `settings/settings.go:329`) — boot state is case-sensitive and validated even when unused.**
`fsmBootState()` accepts only exact `IDLE`/`CATCHINGBLOCKS`/`RUNNING`, while the operator CLI every doc in this PR points at is case-insensitive (`cmd/setfsmstate/set_fsm_state.go:46` lowercases the input; the docs write `--fsmstate catchingblocks`). So `blockchain_initializeNodeInState.operator = catchingblocks` gets a `ConfigurationError`. And because `fsmBootState()` runs before `GetFSMState` (Server.go:413), that error aborts `Init` on nodes that already have a persisted state, where the setting does nothing — a configmap typo rolled out to the fleet crashloops every already-synced node on its next restart. `strings.ToUpper` the trimmed value, and/or validate only on the fresh-store path.

**`services/blockchain/Server.go:425` — configured RUNNING on an empty store is unbootable rather than rejected.**
On any checkpointed network a fresh store's tip is genesis, so the guard always fails and `Init` returns an error → the blockchain pod crashloops with no persisted state. The semantically identical persisted case at line 451 merely degrades to CATCHINGBLOCKS. Meanwhile the setting doc (`settings/blockchain_settings.go:19`) advertises RUNNING as a normal value, which reads as fallback, not refusal. Either degrade to CATCHINGBLOCKS here too, or state in the setting doc and the `settings.conf` comment that RUNNING is only valid for a pre-seeded store above the highest checkpoint.

Checked and fine: settings-context wiring (the `.operator` / `.docker.m` keys do resolve via gocore suffix-trimming), the `errors.New*` format strings, the LEGACYSYNCING migration ordering, the docs' FSM claims against `fsm.go:16-40`, and test isolation.

services/blockchain/Server.go
```diff
@@ -440,7 +443,18 @@ func (b *Blockchain) Init(ctx context.Context) error {
 			stateStr = blockchain_api.FSMStateType_CATCHINGBLOCKS.String()
 
 			if setErr := b.store.SetFSMState(ctx, stateStr); setErr != nil {
-				b.logger.Errorf("[Blockchain][Init] error persisting migrated FSM state: %v", setErr)
+				return errors.NewStorageError("[Blockchain][Init] failed to persist migrated FSM state", setErr)
```
freemans13, line 447: **The persisted FSM state is never validated, and the only documented recovery from a bad one goes away with this PR.**

`fsmBootState()` now rejects anything outside `IDLE`, `CATCHINGBLOCKS`, `RUNNING` for the configured value. The persisted value takes a different route: past this LEGACYSYNCING special case, whatever string came out of the store goes straight into `b.finiteStateMachine.SetState(stateStr)` at line 462. `FSMStateType_value` has exactly three entries (`blockchain_api.pb.go:142-146`) and looplab's `SetState` accepts any string, so an unrecognised persisted name leaves the FSM sitting on a state with no outgoing edges. `AvailableEventsForState` returns an empty slice for it, both `RUN` and `CATCHUPBLOCKS` fail, and the node is parked with no way out short of editing the store by hand.

The branch you are adding this comment on is itself the evidence that this happens: state names do get retired, and the proto now carries `reserved 3; reserved "LEGACYSYNCING";` (`blockchain_api.proto:975-981`). A rollback across a rename, or a hand-edited row, lands in the same place as LEGACYSYNCING did.

Meanwhile this PR removes the one line that told operators how to dig out. `docs/howto/miners/docker/minersHowToSyncTheNode.md` used to say "If a startup race leaves the node in `INIT`, set the state manually" with a `setfsmstate` example; the rewritten "Verify a Seed Before Catch-up" section drops it and nothing replaces it.

Suggestion: give the general case the same treatment as LEGACYSYNCING. After the migration block, if `blockchain_api.FSMStateType_value[stateStr]` has no entry, log it, set `stateStr` to `CATCHINGBLOCKS` and persist, so the node self-heals the way it already does for the one retired name you handle explicitly.

I could not find a code path in this repo that writes a non-enum value today, so this is robustness rather than a live defect.

settings.conf
```diff
@@ -388,7 +388,12 @@ blockchain_grpcListenAddress.dev                    = localhost:${BLOCKCHAIN_GRP
 blockchain_httpListenAddress             = :${BLOCKCHAIN_HTTP_PORT}
 blockchain_httpListenAddress.dev         = localhost:${BLOCKCHAIN_HTTP_PORT}
 
-blockchain_initializeNodeInState =
+# FSM state for a fresh node with no persisted FSM state. Empty preserves
+# automatic CATCHINGBLOCKS startup. Production operator and single-node Docker
+# contexts start in IDLE so a seed can be inspected before catchup begins.
+blockchain_initializeNodeInState          =
+blockchain_initializeNodeInState.operator = IDLE
+blockchain_initializeNodeInState.docker.m = IDLE
```
freemans13, line 400: **This leaves two of the shipped Docker quickstart's own containers permanently unhealthy on a fresh store.**

Six services block on `WaitUntilFSMTransitionFromIdleState` before they bind anything. Two of them are enabled in the mainnet quickstart and carry port-listen healthchecks:

- **subtreevalidation** waits at `services/subtreevalidation/Server.go:555` and only starts its gRPC server at `:588`. Healthcheck: `netstat -ltn | grep -c 8086` (`deploy/docker/base/docker-teranode.yml:74-77`).
- **pruner** waits at `services/pruner/server.go:267` and starts gRPC last. Healthcheck greps `8096` (`docker-teranode.yml:194-197`).

`grep -c` exits 1 on zero matches, so both containers report unhealthy for as long as the node sits in IDLE. With this line that is now the default first-boot state for `docker.m`, and the new section in `docs/howto/miners/docker/minersHowToSyncTheNode.md` sends the operator to `./status.sh` at exactly that moment. Someone following the guide sees two unhealthy containers in the step whose whole purpose is confirming the seed looks right, and reasonably concludes the seed failed.

No deadlock, to be clear: nothing in `deploy/docker/mainnet/docker-compose.yml` has `depends_on: condition: service_healthy` on subtreevalidation or pruner, and the blockchain container binds 8087 regardless of FSM state, so the rest of the stack comes up and `./cli.sh setfsmstate` works. blockvalidation and asset already avoid the problem by binding before the wait, and blockvalidation says why in a comment at `services/blockvalidation/Server.go:1033-1035`.

Either fix works: say in the doc that those two stay unhealthy until catch-up starts, or move the listener ahead of the IDLE wait in those services the way blockvalidation did.

Two related notes. legacy, validator, propagation, p2p, blockpersister and utxopersister have the same pre-bind wait, and propagation (8084) and peer (9905) are healthchecked too, so any stack that enables them inherits this. And the same shape applies to whatever readiness probes the operator chart puts on these services under the `operator` context on line 395, but that chart is not in this repo and I have not checked it.

freemans13 (commented):
Feedback only, I have not touched the branch.

**Verdict:** not approving while icellan's blocking review is open, but I am not adding a second block. The two items below are should-fix, neither is a blocker on its own.

I agree with all three of icellan's points, the first one especially. `guardRunBelowHighestCheckpoint` returns the same `StateError` for "tip is below the checkpoint" and for "could not read the tip", and the persisted branch turns both into a durable write of CATCHINGBLOCKS. The docs make that harder to diagnose rather than easier: `minersHowToInteractWithFSM.md` now describes the demotion purely as failing the checkpoint gate, so an operator whose synced miner drops to catch-up after a Postgres hiccup has nothing in the guide pointing at a failed tip read.

Two things I did not see raised, both about paths this change runs on rather than the change itself. Details inline.

1. **Persisted FSM state is still unvalidated, and this PR deletes the one documented recovery for it.** The configured value now fails fast on anything outside the three enum names; the persisted value goes straight into `SetState` unchecked, and an unrecognised name has no outgoing FSM edges.
2. **`docker.m` defaulting to IDLE leaves two of the quickstart's own containers permanently unhealthy.** subtreevalidation and pruner both block on the IDLE wait before binding their gRPC ports, and their compose healthchecks are port-listen probes on those exact ports.

Worth keeping, and nobody has said it: the `GetFSMState` fail-fast is the strongest part of this change and it fixes a real bug that predates it. The old code logged a read error and fell through with `stateStr == ""`, which is the fresh-node branch, so a transient store error made a synced node re-initialise as fresh and overwrite its own persisted state. That is now an abort. Same for the `SetFSMState` paths.

**Verification.** `go test ./settings/... ./services/blockchain/...` and `go vet` on the same packages are green locally at 7977bf6d (blockchain package 28.0s, no failures). Upstream CI on this head is 37 checks passing and 1 skipped (`test aerospike native-ops (fork image)`). I did not run the Docker quickstart, so item 2 is read off the compose healthchecks and the service start order, not observed on a running stack.

services/blockchain/Server.go
freemans13: Resolved in c299b58f, and I am happy with the route you took rather than the one I suggested.

You abort on an unrecognised persisted name instead of self-healing to CATCHINGBLOCKS. Refusing to guess whether the operator meant catch-up or mining is the better call for a state this consequential, and the ordering is right: LEGACYSYNCING still migrates before the enum check, so the one name you do understand is still handled silently.

The half I actually cared about was recovery, and the new Troubleshooting Sync section covers it, including the point that `teranode-cli` cannot repair this because blockchain startup is what is failing. That is better than the `INIT` line it replaces.

I also went and checked how bad the abort could be in practice. The only runtime writer of persisted FSM state is `SendFSMEvent`, and it writes `b.finiteStateMachine.Current()` after a transition the FSM already accepted, so it can only ever store one of the three enum names (`services/blockchain/Server.go:2939-2942`). Nothing in the repo can write a value that trips your new check. It is defence against a version rollback or a hand-edited row, which is exactly what it should be.

`TestInit_UnknownPersistedStateAbortsWithoutRewriting` asserting `setFSMStateCalls` is zero is the right assertion, since the whole point is that the bad value survives for someone to look at.

settings.conf
freemans13: Resolved in c299b58f. The new paragraph names subtreevalidation and pruner, says why their gRPC ports are not listening, and says the blockchain listener stays up so the CLI still works, which is the bit an operator actually needs at that moment.

The line I did not expect and think is the best part is "do not dismiss a continuing failure as expected". Documenting an expected-unhealthy window usually creates a second problem, where a genuine failure gets waved through because the guide said unhealthy was normal. Telling the reader to re-check after leaving IDLE closes that.

You also caught something I missed: the plain network-sync path at the top of the guide needed the explicit `setfsmstate --fsmstate catchingblocks` too, not just the seeded path. A fresh `docker.m` node with no seed at all sits in IDLE for the same reason and I had only looked at the seed window.

Still unchecked, and still not in this repo: whatever readiness probes the operator chart puts on these services under the `operator` context. Worth a glance from someone who has that chart, since the pod-level consequence there is harsher than an unhealthy container in `docker compose ps`.

services/blockchain/Server.go
oskarszoon: Addressed in c299b58f571ddd6c8fcbef044e304fdf79b26b62. Unknown persisted names now fail Init before SetState or any write, with an error naming the value. The known LEGACYSYNCING migration remains supported.

I chose fail-fast rather than migrating arbitrary unknown values to CATCHINGBLOCKS: an unrecognized value is insufficient evidence that the operator intended synchronization. The Docker guide now documents recovery with the stack stopped, including compatible backups/store repair and why the CLI cannot repair a failed startup.

TestInit_UnknownPersistedStateAbortsWithoutRewriting covers INIT, RUNNIG, idle and FUTURE_STATE, verifies zero writes and preservation of the stored value. The regression failed before the fix; package race tests and the final full unit/race suite passed.

settings.conf
oskarszoon: Addressed with the documentation option in c299b58f571ddd6c8fcbef044e304fdf79b26b62. The Docker seed-inspection section now explains that subtreevalidation and pruner wait before binding gRPC in IDLE, so listener-based healthchecks can remain starting/unhealthy during that window. It directs operators to inspect the FSM, tip and logs, start CATCHINGBLOCKS explicitly, then recheck health and investigate failures that persist after startup.

The network-sync instructions also include the explicit catchup command. Listener ordering is unchanged. Verified against the service startup paths and bundled compose definitions; strict MkDocs passed. I did not run a Docker deployment for this correction.

freemans13 (approved):
Approving on c299b58f. Both of my items are resolved and I have replied in each thread. Feedback only throughout, I have not touched the branch.

icellan's review is still the open block and only they can clear it, so here is what changed against each of their three points, to make that re-check quick.

**1. Transient store error durably demoting persisted RUNNING. Fixed properly.** `evaluateRunCheckpoint` now returns whether the gate error came from an actually observed tip height, and the persisted branch demotes only in that case. A read failure or nil meta returns the error and `Init` aborts with the stored RUNNING untouched. The test that locked in the wrong behaviour is gone: the `"missing tip"` case is renamed `"genesis tip"`, which is what it always really was, since an empty store still has genesis and height 0 is a genuine comparison below the checkpoint. The real unreadable-tip case is now covered separately, across a store error, a deadline, a cancellation and nil metadata, and each asserts zero writes plus a successful retry once the store recovers. That is the end state, not the call.

**2. Boot state validated even when unused. Half fixed, half declined with reasons.** `fsmBootState()` moved inside the fresh-store branch, so a configmap typo no longer crashloops already-synced nodes. Case sensitivity is unchanged and now documented as deliberate in four places, including "Values are uppercase" in the setting doc and `settings.conf`.

**3. Configured RUNNING unbootable rather than rejected. Taken via the doc route you offered.** Both the setting longdesc and the `settings.conf` comment now say RUNNING needs a pre-seeded store at or above the highest checkpoint and that startup fails with no fallback below it.

**One small thing, no need to block on it.** Moving the validation inside the fresh branch also means a bad value is now completely silent on every node that has a persisted state. A typo rolled out to the fleet is discovered only when some node is next reset. A `Warnf` when `blockchain_initializeNodeInState` is set but ignored would restore that without putting the abort back.

**One observation, pre-existing and out of scope.** One commit here is called "fail on FSM state writes", and after it the one write that still does not fail is the runtime one: `SendFSMEvent` logs a `SetFSMState` error and returns success anyway (`services/blockchain/Server.go:2941-2946`, last touched in an upstream commit). So `setfsmstate --fsmstate catchingblocks` can report success while nothing was persisted, and the next restart resumes the old state. Worth a follow-up issue rather than widening this PR.

**Verification.** `go test ./settings/... ./services/blockchain/` and `go vet` on the same packages are green locally at c299b58f (blockchain package 27.9s, no failures). Upstream CI on this head is passing with one skip (`test aerospike native-ops (fork image)`). Note that the Copilot review sitting on this head is its quota-limit refusal, not a review, so 1705d0e1 is the last head it actually looked at. I still have not run the Docker quickstart, so the healthcheck behaviour behind the new doc paragraph is read from the compose definitions and service start order rather than observed.

ordishs (approved):
- **services/blockchain/Server.go:487-500** — P1: `fsmBootState` accepts only uppercase names and aborts startup otherwise. Every operator-facing FSM command here uses the lowercase form, including the docs this PR rewrites (`setfsmstate --fsmstate catchingblocks`) and `cmd/setfsmstate/setfsmstate.go:48`, which lowercases its input before matching. An operator who writes `blockchain_initializeNodeInState.operator = catchingblocks` gets `invalid blockchain_initializeNodeInState "catchingblocks"` on the first start of a new production node, and the blockchain service never starts — CrashLoopBackOff on Kubernetes, an endless `restart: on-failure` retry under the bundled compose. `TestInit_InvalidConfiguredBootStateFails` proves this for `"idle"`. Suggest `strings.ToUpper(configured)` before the switch, keeping the abort for genuinely unknown names.

- **settings/settings.go:329, services/blockchain/Server.go:478-500** — P2: the boot state is validated only in the no-persisted-state branch, so a malformed value on a node that already has persisted FSM state is never read and never reported. The typo then aborts the first boot of the next fresh node, possibly months later and on a different machine from the one where the change was made. Validation-only-on-use is defensible; silence on an unparseable value is the part worth changing. Suggest validating in `NewSettings`, or logging a warning at `Init` when a non-empty value is ignored.

- **docs/howto/miners/docker/minersHowToSyncTheNode.md:170-178** — P2: the IDLE-window healthcheck note names only `subtreevalidation` and `pruner`. `propagation` has the same pre-listener FSM wait (`services/propagation/Server.go:325` before `:366`) and the same port-listen healthcheck on 8084 (`deploy/docker/base/docker-teranode.yml:145`). An operator running `./status.sh` sees `propagation` unhealthy, and the doc's own "do not dismiss a continuing failure as expected" then points at a non-problem. Please name `propagation` in the list.

- **services/blockchain/Server.go:456-462** — P2, no code change requested: every restart of a node persisted in RUNNING now performs a `GetBestBlockHeader` read during `Init`, and a failed read aborts startup where previously the node restored RUNNING with no store read. Fail-closed is the right choice and the persisted state is preserved for the retry (`TestInit_PersistedRunningCheckpointReadFailurePreservesState`). Please confirm the Kubernetes operator retries such a startup abort rather than backing off indefinitely, and consider a docs line stating the abort is expected to clear on retry.

- **Kubernetes probes** — P2, unverified from this repository: the k8s manifests here define no probes for Teranode pods; the operator supplies them and lives elsewhere. If it attaches a *liveness* probe to the gRPC port of an FSM-gated service, those pods restart forever during the IDLE window and the inspection window is unusable. The pre-existing k8s docs already claimed a fresh node starts in IDLE, which suggests the operator tolerated IDLE historically. Given the PR body notes Docker healthcheck behaviour was checked in source only, the same caveat applies here with more consequence — worth one live check before release.

icellan (approved):
Re-reviewed at `c299b58f5`. All three earlier findings are addressed:

- `evaluateRunCheckpoint` now separates "observed a low tip" from "couldn't read the tip", and only the former demotes persisted RUNNING (Server.go:456-459). This was the important one.
- `fsmBootState()` moved inside the fresh-store branch (Server.go:420), so a bad config value can no longer abort `Init` on nodes that already have persisted state.
- The RUNNING semantics — pre-seeded store at or above the checkpoint, no fallback — are now stated plainly in the setting doc.

Approving. Five non-blocking items worth a follow-up:

1. Exit code on startup failure. This PR makes blockchain `Init` able to fail for the first time, but the path is log + `ForceShutdown` + `close(doneCh)` (daemon/daemon.go:306-310), then `RunDaemon` returns and `main.go:76` returns normally — **exit 0**. Under `restart: on-failure` in `deploy/docker/base/docker-teranode.yml`, a transient store blip at boot leaves the container stopped and never retried. Either `restart: unless-stopped` for `blockchain`, or exit non-zero on startup failure. k8s `restartPolicy: Always` is unaffected.
2. Four docs still tell operators to run `setfsmstate --fsmstate RUNNING`, which is the command the new gate refuses on a fresh below-checkpoint node: `docs/howto/miners/minersHowToInteractWithFSM.md:78,102`, `docs/howto/miners/docker/minersHowToStopStartDockerTeranode.md:69`, `docs/howto/miners/docker/minersHowToTroubleshooting.md:59`. The k8s guides were swept; these were missed.
3. The IDLE-window warning names `subtreevalidation` and `pruner` but not `propagation`, which also doesn't bind while IDLE.
4. `docs/howto/miners/kubernetes/minersHowToSyncTheNode.md` Step 6 has no IDLE-window warning at all — those pods sit `0/1 NotReady` until someone runs `setfsmstate`, which reads as a broken deploy.
5. `settings/blockchain_boot_state_context_test.go:55` asserts `"IDLE"` under the `operator` context, whose default is already `IDLE` — the test passes even if the env var were ignored. Use `" CATCHINGBLOCKS "`.

Also: only `LEGACYSYNCING` is migrated, so a store persisted under `STOPPED`/`MINING`/`CATCHINGTXS`/`RESTORING` (dropped in `1b18b0c69`) now hits the new unknown-state abort with no CLI recovery. Those names predate current releases, so it's theoretical — but the mappings are unambiguous and extending the map is cheaper than the datastore-repair runbook.

Issue #1670, opened 2026-08-31
blockchain FSM: no CATCHINGBLOCKS -> IDLE edge, so a fresh v0.16 node cannot be set idle for rewind


oskarszoon:
## Problem

There is no `CATCHINGBLOCKS -> IDLE` edge in the blockchain FSM, so `teranode-cli setfsmstate --fsmstate idle` fails when the node is catching up. Since #1135 a fresh node *boots* into `CATCHINGBLOCKS`, which makes this the default state a new node sits in.

`services/blockchain/fsm.go` — `STOP` accepts exactly one source on `main`/`release/v0.16`:

```go
{
    Name: blockchain_api.FSMEventType_STOP.String(),
    Src:  []string{ blockchain_api.FSMStateType_RUNNING.String() },
    Dst:  blockchain_api.FSMStateType_IDLE.String(),
},
```

A test added in #1135 records the consequence rather than treating it as a defect:

```go
// services/blockchain/server_test.go, Test_Idle
expectError: true, // Fresh node starts in CATCHINGBLOCKS; STOP event has no CATCHINGBLOCKS->IDLE edge
```

### What changed, and why it matters more now

|                  | `STOP -> IDLE` sources          | Fresh-node boot state          |
| ---------------- | ------------------------------- | ------------------------------ |
| `release/v0.15`  | `RUNNING`, `LEGACYSYNCING`      | `IDLE` (constructor default)   |
| `release/v0.16`  | `RUNNING` only                  | `CATCHINGBLOCKS`               |

The edge was already missing on `release/v0.15`, but it was hard to reach: a fresh node booted `IDLE`, and `LEGACYSYNCING` had its own edge to `IDLE`. On `release/v0.16` the `LEGACYSYNCING` state is gone and fresh nodes boot straight into `CATCHINGBLOCKS`, so the only way to `IDLE` is a two-step through `RUNNING`:

```
teranode-cli setfsmstate --fsmstate running
teranode-cli setfsmstate --fsmstate idle
```

Routing through `RUNNING` to get to `IDLE` is the wrong shape for an operator who wants the node to stop doing work. `RUNNING` switches on live subtree validation and block-assembly tx feeding — which is exactly what #1135 was careful to avoid on a not-yet-caught-up node — and then immediately switches it back off.

### Operational impact

`cmd/rewindblockchain` gates on `IDLE`:

```go
// cmd/rewindblockchain/rewindblockchain/preflight.go
[...]

PR #1695, opened 2026-09-05
fix(blockchain): bound FSM work and survive RPC cancellation

oskarszoon:
Superseded by #1696, which includes the bounded-transition hardening together with persistence-before-notification, uncertain-write reconciliation, and bounded promotion retries. Review and merge #1696 next.

An RPC cancellation during `looplab/fsm` v1.0.2's transition can leave the FSM permanently in transition, rejecting later events until restart. A checkpoint lookup can also hold the transition lock indefinitely when the caller supplies no deadline.

Complete accepted in-process transitions independently of caller cancellation, and bound checkpoint reads and accepted-state persistence with `blockchain_store_dbTimeoutMillis` (5 seconds for nonpositive values). Checkpoint admission still respects an earlier caller deadline and fails closed. Classify an already-pending transition as a state error.

This is a prerequisite for durable operator pause in #1670. It consumes the checkpoint gate merged in #1685 and leaves boot policy in #1693. Runtime persistence failure reporting and notification ordering are handled in #1696.

Validation: regression tests observed failing before implementation; blockchain package tests, race tests, `go vet`, and `golangci-lint` passed (zero lint issues). Tests use real SQLite plus a narrow blocking-read wrapper and cover cancellation followed by another transition, checkpoint timeout with preserved state and released lock, earlier caller deadline, and zero/negative timeout settings.

Accepted work may finish after the client disconnects. Store timeouts depend on the store honoring context cancellation.

Store contexts are canceled immediately after persistence, before the optional response delay. Configuration and the nonpositive FSM fallback share one database-timeout default. A regression reproduces the former timer lifetime issue and verifies prompt cancellation. The updated blockchain/settings race suites, vet, and changed-code lint pass.

### DIFF
--- a/services/blockchain/Server.go
+++ b/services/blockchain/Server.go
@@ -2932,7 +2932,7 @@ func (b *Blockchain) SendFSMEvent(ctx context.Context, eventReq *blockchain_api.
 func (b *Blockchain) fsmStoreContext(ctx context.Context) (context.Context, context.CancelFunc) {
 	timeout := time.Duration(b.settings.BlockChain.StoreDBTimeoutMillis) * time.Millisecond
 	if timeout <= 0 {
-		timeout = 5 * time.Second
+		timeout = time.Duration(settings.DefaultBlockchainStoreDBTimeoutMillis) * time.Millisecond
 	}
 	return context.WithTimeout(ctx, timeout)
 }
--- a/services/blockchain/Server.go
+++ b/services/blockchain/Server.go
@@ -2879,11 +2879,15 @@ func (b *Blockchain) SendFSMEvent(ctx context.Context, eventReq *blockchain_api.
 		}
 	}
 
-	err := b.finiteStateMachine.Event(ctx, eventReq.Event.String())
+	// Once admitted, complete the local transition even if the RPC is cancelled.
+	// looplab/fsm v1.0.2 leaves a pending transition behind when its context is
+	// cancelled during Event, rejecting every subsequent event until restart.
+	transitionCtx := context.WithoutCancel(ctx)
+	err := b.finiteStateMachine.Event(transitionCtx, eventReq.Event.String())
 	if err != nil {
 		b.logger.Debugf("[Blockchain Server] Error sending event to FSM, state has not changed.")
 		switch err.(type) {
-		case fsm.InvalidEventError, fsm.NoTransitionError:
+		case fsm.InvalidEventError, fsm.NoTransitionError, fsm.InTransitionError:
 			return nil, errors.WrapGRPC(errors.NewStateError("[Blockchain Server] FSM event %s rejected in state %s", eventReq.Event.String(), priorState, err))
 		default:
 			return nil, errors.WrapGRPC(err)
@@ -2893,7 +2897,9 @@ func (b *Blockchain) SendFSMEvent(ctx context.Context, eventReq *blockchain_api.
 	state := b.finiteStateMachine.Current()
 
 	// set the state in persistent storage
-	err = b.store.SetFSMState(ctx, state)
+	storeCtx, cancel := b.fsmStoreContext(transitionCtx)
+	defer cancel()
+	err = b.store.SetFSMState(storeCtx, state)
 	// check if there was an error setting the state
 	if err != nil {
 		b.logger.Errorf("[Blockchain Server] Error setting the state in blockchain db: %v", err)
@@ -2921,6 +2927,16 @@ func (b *Blockchain) SendFSMEvent(ctx context.Context, eventReq *blockchain_api.
 	return resp, nil
 }
 
+// fsmStoreContext bounds store operations performed while fsmMu is held.
+// Non-positive configuration falls back to the default database timeout.
+func (b *Blockchain) fsmStoreContext(ctx context.Context) (context.Context, context.CancelFunc) {
+	timeout := time.Duration(b.settings.BlockChain.StoreDBTimeoutMillis) * time.Millisecond
+	if timeout <= 0 {
+		timeout = 5 * time.Second
+	}
+	return context.WithTimeout(ctx, timeout)
+}
+
 // guardRunBelowHighestCheckpoint blocks the RUN transition when the local
 // chain tip has not yet reached the highest hard-coded checkpoint for the
 // active network.
@@ -2938,7 +2954,9 @@ func (b *Blockchain) guardRunBelowHighestCheckpoint(ctx context.Context) error {
 		return nil
 	}
 
-	_, meta, err := b.store.GetBestBlockHeader(ctx)
+	storeCtx, cancel := b.fsmStoreContext(ctx)
+	defer cancel()
+	_, meta, err := b.store.GetBestBlockHeader(storeCtx)
 	if err != nil {
 		return errors.NewStateError("cannot read best block header to evaluate RUN gate", err)
 	}
--- a/services/blockchain/Server.go
+++ b/services/blockchain/Server.go
@@ -2898,8 +2898,8 @@ func (b *Blockchain) SendFSMEvent(ctx context.Context, eventReq *blockchain_api.
 
 	// set the state in persistent storage
 	storeCtx, cancel := b.fsmStoreContext(transitionCtx)
-	defer cancel()
 	err = b.store.SetFSMState(storeCtx, state)
+	cancel()
 	// check if there was an error setting the state
 	if err != nil {
 		b.logger.Errorf("[Blockchain Server] Error setting the state in blockchain db: %v", err)
--- a/settings/blockchain_settings.go
+++ b/settings/blockchain_settings.go
@@ -5,6 +5,10 @@ import (
 	"time"
 )
 
+// DefaultBlockchainStoreDBTimeoutMillis supplies both the configured database
+// timeout default and the FSM fallback for non-positive configuration.
+const DefaultBlockchainStoreDBTimeoutMillis = 5000
+
 // BlockChainSettings configures the blockchain state management service.
 type BlockChainSettings struct {
 	GRPCAddress           string            `key:"blockchain_grpcAddress" desc:"gRPC address for Blockchain service" default:"localhost:8087" category:"BlockChain" usage:"Address for inter-service communication" type:"string" longdesc:"### Purpose\nSpecifies the client connection address for the Blockchain gRPC service.\n\n### How It Works\nOther services connect to this address to query blockchain state, headers, and chain information.\n\n### Used By\n- **Block Validation** - Get headers, check chain\n- **Block Assembly** - Get best block\n- **Asset Server** - Query blocks\n- **RPC** - getblock, getblockheader\n\n### Format\n**host:port** - Where host is a hostname, IP address, or DNS name.\n\n### Examples\n- **localhost:8087** - Local development (default)\n- **blockchain.internal:8087** - Internal DNS\n- **10.0.1.50:8087** - Direct IP address\n\n### Recommendations\n- Use internal DNS or service mesh addresses in production\n- Use localhost only for local development"`
--- a/services/blockchain/server_fsm_hardening_test.go
+++ b/services/blockchain/server_fsm_hardening_test.go
@@ -1,0 +1,138 @@
+package blockchain
+
+import (
+	"context"
+	"net/url"
+	"testing"
+	"time"
+
+	"github.com/bsv-blockchain/go-chaincfg"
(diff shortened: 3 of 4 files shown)


services/blockchain/Server.go
```diff
@@ -2938,7 +2954,9 @@ func (b *Blockchain) guardRunBelowHighestCheckpoint(ctx context.Context) error {
 		return nil
 	}
 
-	_, meta, err := b.store.GetBestBlockHeader(ctx)
+	storeCtx, cancel := b.fsmStoreContext(ctx)
```
freemans13, line 2957: Worth naming explicitly, because it is an operator-visible change and not just a resource tidy-up: RUN now refuses on a slow store where it previously waited on one.

Before this line, `GetBestBlockHeader` ran on the caller's context. A blockchain store that took twelve seconds to answer still answered, and the gate then evaluated the real tip height. Now anything past `blockchain_store_dbTimeoutMillis`, five seconds by default, comes back as a context deadline. `guardRunBelowHighestCheckpoint` turns that into `cannot read best block header to evaluate RUN gate` and RUN is refused.

Failing closed is the right call for a checkpoint gate, and the operator can simply retry, so I am not asking you to change the bound. The reason I am raising it is that the path is reachable on a real node. On a mainnet node backed by PostgreSQL during initial block download, the store periodically saturates its disk and even single-row reads stall well past five seconds. In that window `setfsmstate --fsmstate running` starts returning an error whose wording points at the checkpoint gate when the actual cause is a slow database.

Cheap fix, if you think it is worth it: distinguish the deadline case in the message, something like `cannot read best block header to evaluate RUN gate: store did not respond within blockchain_store_dbTimeoutMillis`, when `errors.Is(err, context.DeadlineExceeded)` holds. That saves whoever hits it from going looking for a checkpoint problem that is not there.

settings/blockchain_settings.go
```diff
@@ -5,6 +5,10 @@ import (
 	"time"
 )
 
+// DefaultBlockchainStoreDBTimeoutMillis supplies both the configured database
+// timeout default and the FSM fallback for non-positive configuration.
+const DefaultBlockchainStoreDBTimeoutMillis = 5000
```
freemans13, line 10: This is the first consumer of `blockchain_store_dbTimeoutMillis` anywhere in the tree, and the description it inherits does not match what it now does.

On `upstream/main`, `BlockChain.StoreDBTimeoutMillis` is loaded in `settings/settings.go` and read nowhere else. I grepped the whole repository for non-test readers and found only the settings package itself. The key is also absent from `settings.conf`, so every deployment has been taking the 5000 default without it meaning anything. Making a dead setting live is a good outcome, so this is not a complaint about the choice to reuse it.

The problem is that two pieces of documentation now describe it wrongly. The struct tag below says queries exceeding the timeout "are cancelled and trigger retry logic if retries remain", and neither of the two new call sites retries. The settings reference table calls it "Database operation timeout (store-level)", but it is applied server-side to exactly two calls in the blockchain service's FSM path rather than to blockchain store operations in general. That table is at `docs/references/settings/services/blockchain_settings.md:17`, and it already flags neighbouring dead settings as `**UNUSED**`, so it is clearly meant to be kept honest.

The consequence is concrete rather than cosmetic. An operator who reads "Decrease for faster failure detection in high-performance environments" and sets this to 200ms is really capping FSM state persistence at 200ms. When `SetFSMState` then misses the deadline, the error is only logged, the RPC still returns success carrying the new state, and the node restarts into the stale stored state, because `Init` restores the FSM from the store at `services/blockchain/Server.go:411-450`. That is the divergence you already acknowledged and deferred to #1696, but the new timeout is what makes it reachable through configuration.

Updating the `longdesc` and the docs row to say what the knob actually governs would make it safe for someone to tune.

freemans13 (approved):
Approving. The premise holds up against the library source, and every one of the four behaviour changes is covered by a test that genuinely fails without it. My two findings are about documentation and operator experience, not correctness, and neither needs to block the merge.

## The wedge is real, and it is permanent

I checked the claim in the new comment rather than taking it on trust, because it is the whole justification for `context.WithoutCancel`. It is accurate, and the mechanism is worth writing down.

In looplab/fsm v1.0.2, `Event` stores the pending transition in `f.transition` and then runs it. The closure begins `if ctx.Err() != nil { e.Err = ctx.Err(); return }` and returns without clearing `f.transition` (`fsm.go:344-352`). The caller, `transitionerStruct.transition`, does not clear it either (`fsm.go:423-429`). So a cancellation landing in that window leaves `f.transition` non-nil forever. Every later `Event` then trips the `if f.transition != nil { return InTransitionError{event} }` check at `fsm.go:312`, and `Can` returns false because it is defined as `ok && (f.transition == nil)`. The FSM is dead until the process restarts. Nothing in the library clears the field on a later call.

That also makes the `fsm.InTransitionError` case in the error switch the right classification rather than a cosmetic tidy-up. Teranode registers only an `enter_state` callback and no `leave_state` or `before_event` ones (`services/blockchain/fsm.go:66-86`), so the cancel-check inside the transition closure is the only place the wedge can form, and this PR closes the only route to it. `finiteStateMachine.Event` has exactly one non-test caller in the tree.

## Mutation testing: all four fixes are load-bearing

A passing test proves nothing unless it fails when the fix is removed, so I reverted each change in turn and re-ran the new suite. Every fix is covered, and no test is redundant.

| Fix reverted | Test that failed |
|---|---|
| `transitionCtx := context.WithoutCancel(ctx)` back to `ctx` | `TestSendFSMEvent_AcceptedCancellationDoesNotWedge` |
| `fsm.InTransitionError` dropped from the error switch | `TestSendFSMEvent_InTransitionReturnsStateError` |
| `cancel()` moved back to a `defer` | `TestSendFSMEvent_ReleasesStoreContextBeforeResponseDelay` |
| checkpoint read back on the raw `ctx` | `TestSendFSMEvent_CheckpointTimeoutReleasesTransitionLock`, `TestGuardRunBelowHighestCheckpoint_NonPositiveTimeoutStillBounded` |

The tests assert end state rather than call shape, which is what makes them worth having. The cancellation test checks that the store holds `RUNNING` and that a following `STOP` still works, not merely that no error came back.

I also confirmed the ordering is sound. The checkpoint guard runs on the cancellable `ctx` and `context.WithoutCancel` is only applied afterwards, so admission still fails closed on a caller deadline while the accepted transition survives. Reusing `fsmStoreContext` for both call sites is correct because `context.WithTimeout` takes the shorter of the two deadlines, which is what `TestGuardRunBelowHighestCheckpoint_RespectsEarlierDeadline` pins down.

## Two findings, both left inline

**RUN now refuses on a slow store where it used to wait.** This is a real behaviour change on a real node, not a theoretical one, and I think it deserves a clearer error message rather than a code change. Details on `services/blockchain/Server.go`.

**`blockchain_store_dbTimeoutMillis` is documented as something it is not.** This PR is its first consumer anywhere in the tree, and both the struct tag and the settings reference describe behaviour that does not match the two paths it now governs. Details on `settings/blockchain_settings.go`.

## Verification

Run locally on `8f95b6e`:

- `go test -count=1 -race ./services/blockchain/` passed in 37.9s.
- `go test -count=1 ./settings/` passed.
- `go vet ./services/blockchain/ ./settings/` clean.
- The four mutations above, each reverted and restored individually.

Upstream CI is fully green, 38 checks including the complete `test`, `smoketest` and `sequential` shards, so the suite really ran rather than being skipped.

## What I did not check

I did not exercise this on a live node, so the slow-store behaviour described in the inline comment is reasoned from the code path and from how the PostgreSQL-backed mainnet node behaves under load, not measured against a stalling store. Nothing here tests the interaction with the durable operator pause in #1670 that this is a prerequisite for.

icellan (changes requested):
Verified the looplab/fsm v1.0.2 leak this fixes — `Event` sets `f.transition` at fsm.go:372 and `doTransition` returns early on `ctx.Err() != nil` (fsm.go:347-352) without clearing it, so a cancelled caller context wedges the FSM until restart. `WithoutCancel` at the admission point is the right fix.

Two things on the new timeout semantics:

- `services/blockchain/Server.go:2900` — `SetFSMState` is now capped at `StoreDBTimeoutMillis` (5s) where it previously inherited a deadline-less caller context, and its error is still only logged. Under store load the in-memory FSM goes `RUNNING` and the RPC succeeds while the persisted state stays `CATCHINGBLOCKS`; the next restart restores the stale state in `Init` and the node silently reverts. A bounded write that can be dropped needs the error returned or retried, not logged.
- `settings/blockchain_settings.go:10` — `blockchain_store_dbTimeoutMillis` had no consumer anywhere in the tree before this PR, and its `longdesc` still describes a generic query timeout with retry logic ("decrease for faster failure detection in high-performance environments"). Following that advice now drops FSM state writes and makes the RUN gate refuse, with no retries behind either. Either document the new role or add a separate FSM-specific key.

Lower priority:

- `services/blockchain/Server.go:2957` — the RUN gate fails closed on the tip read, and its only automatic caller (`services/blockvalidation/catchup.go:1432`) logs and returns with no retry, so a slow read leaves the node in `CATCHINGBLOCKS` until some later catchup calls `restoreFSMState` again.
- `services/blockchain/Server.go:2890` — recognising `InTransitionError` improves the message but nothing clears `f.transition`: `Event` refuses to start a transition (fsm.go:312-314) and `SetState` doesn't touch it (fsm.go:223), so `setfsmstate` can't unwedge it either and the node still needs a restart. Recreating the FSM under `fsmMu` (`NewFiniteStateMachine()` + `SetState(priorState)`) would make it self-healing; the new test currently locks in "clear error, no remedy".

Nit: `WithoutCancel` leaves the blocking `b.notifications <- req` (Server.go:2515, buffer 100) as the one genuinely unbounded piece of work done under `fsmMu`. Not a regression — it already used `context.Background()` — but it sits awkwardly with "bound FSM transition work".

ordishs (approved):
Verified the `looplab/fsm` v1.0.2 wedge in the module source: `transitionFunc` returns early on `ctx.Err() != nil` without resetting `f.transition`, so every later `Event` returns `InTransitionError` until restart. `SendFSMEvent` is the only `Event` caller, so `context.WithoutCancel` covers the whole surface. The store-timeout half correctly bounds the two calls that run under `fsmMu`. Reverting each half of the fix in turn makes the matching tests fail, so the regressions are genuine.

One item to settle before merge.

**services/blockchain/Server.go:2900-2906** — P1

- The new 5-second deadline on `SetFSMState` adds a fresh trigger for silent state divergence. The error is only logged. The RPC still returns the in-memory state and a `nil` error. `stores/blockchain/sql/SetFSMState.go:62` issues a single `ExecContext` with no retry, so a deadline is terminal.
- Failure: an operator sends `STOP` before maintenance. The upsert on a loaded PostgreSQL store exceeds 5 seconds. The FSM moves to `IDLE`, the RPC reports `IDLE`, and the store still holds `RUNNING`. On the next restart, `Init` reads the persisted state (`Server.go:445-450`) and boots `RUNNING`. Live subtree validation and block-assembly tx feeding come up during maintenance.
- Before this change the write blocked until it completed, so this path needed caller cancellation — which the rest of the PR now prevents. #1695 on its own widens the window it closes.
- Recommendation: return the persistence error, retry once, or raise the log to a distinct alertable message that names the divergence. If #1696 owns the reporting, please make the merge order explicit.

Smaller items, not blocking:

- **Server.go:2891-2892** — `fsm.InTransitionError` now shares the message `FSM event RUN rejected in state IDLE` with an ordinary invalid transition. An operator cannot tell a bad request from an FSM that needs a restart. Give it its own text.
- **Server.go:2956-2960** — the RUN gate returns `cannot read best block header to evaluate RUN gate` for both a store fault and a deadline. The cold-cache case is reachable: `SetFSMState` calls `ResetResponseCache`, so the tip query after any transition misses the response cache. Distinguish `context.DeadlineExceeded` so the operator knows to retry.
- **settings/blockchain_settings.go:22** — `blockchain_store_dbTimeoutMillis` had no consumer anywhere in the tree before this PR, and its first reader is the service rather than the store. The `longdesc` still promises "trigger retry logic if retries remain", and `docs/references/settings/services/blockchain_settings.md:17` still calls it "store-level". Update both, or add a distinct FSM key.
- **server_fsm_hardening_test.go** — no test covers a `SetFSMState` timeout. `fsmHardeningWriteStore` records the context but never blocks, so the persistence branch of the new timeout is untested.

Verification run locally: `go vet ./services/blockchain/... ./settings/...` clean; `go test -race -count=1 ./services/blockchain/ ./settings/` passes; full package tests pass.

PR #1696, opened 2026-09-05
fix(blockchain): make FSM transitions durable and retry promotion

oskarszoon:
Runtime FSM transitions can report success before their destination is durable. A failed or ambiguously acknowledged write can leave memory and storage disagreeing, and a failed catchup promotion can leave a fully caught-up node without mining.

Persist the destination before changing memory, notifying services, or updating metrics. Return failed writes to callers, retain the prior in-memory state, and log that the database may already contain the destination. Serialize convenience RPC checks with transitions; after an uncertain write, an already-current target succeeds only after a new acknowledged persistence write. Clients always consult the server for these commands.

Bound checkpoint reads and accepted persistence writes with `blockchain_store_dbTimeoutMillis` (5000 ms; nonpositive values use that default). Reads retain an earlier caller deadline; accepted writes finish independently of caller cancellation. These are separate per-operation budgets, not a bound on notification delivery or the entire RPC. Document timeout and pending-transition restart diagnostics.

Automatic RUN and CATCHUPBLOCKS both refuse operator IDLE under the transition lock. This closes direct promotion and the two-step IDLE → CATCHINGBLOCKS → RUNNING escape identified in review. Explicit operator resume uses SendFSMEvent; RUN retains its checkpoint gate. Normal automatic catchup from RUNNING and reconciliation of an already-current target remain available. Legacy RUN callers prefilter cached CATCHINGBLOCKS to avoid redundant requests, while the server remains authoritative.

On an IDLE catchup refusal, validation workers do not start and no completion RUN is scheduled. The consumer logs the need for explicit resume and clears processing markers without consuming the per-block retry budget. Cleanup classifies the refusal as local and preserves any earlier genuine peer failures. Catchup attempt telemetry may already have been recorded before admission. A later notification can retry after explicit resume; immediate replay is not guaranteed. Daemon test setup and operator gRPC examples use explicit events when leaving IDLE.

Catchup completion makes at most three promotion attempts for transient failures with cancellable one-second backoff. Shutdown cancellation is quiet; deliberate IDLE refusal logs at Info; failed confirmation warns without inventing the current FSM state.

Supersedes #1695 and builds on merged #1693. This is the durability prerequisite for #1670 and draft #1698. The CATCHINGBLOCKS-to-IDLE pause transition, work admission during an active catchup, and queue retention/resumption remain in that draft. A write error cannot prove the database stayed unchanged; no compensating rollback is attempted. IDLE is not proof that services have stopped writing, and rewind still requires shutdown.

Validation uses real SQLite and the blockchain authority, with fault injection at store boundaries. Regressions cover persistence failures, uncertain-write reconciliation, ordering, cancellation/timeouts, promotion retries, stale cache, the automatic two-step escape after STOP, concurrent STOP/catchup, explicit resume, validation refusal and cleanup, and concurrent/idempotent daemon setup. Relevant regressions were observed failing before the fixes. Final validation results are recorded below.

Validation: real-authority regressions were observed failing before the fixes. Full affected-package race suites, vet, changed-code lint (zero issues), and strict MkDocs passed. The full local repository suite on `66ab77d5d` passed: 14,129 tests, 47 skipped, 596.838 seconds, with race detection and repository-wide coverage.

Sonar flagged increased handler complexity from the logging branch. Final head `2aee82d80` extracts that unchanged logging into a helper; focused consumer/FSM/release race tests passed in 18.471 seconds, with clean vet and lint. Independent review confirmed equivalent behavior.

The required post-push check completed at 13:59:48 UTC on September 8: final-head CI has 37 successful checks and one skip, with none pending. Fresh Sonar analysis has zero new issues and 98.1% new-code coverage; the complexity finding is cleared. The final benchmark reports zero regressions across 139 comparisons; the report does not identify the baseline commit. Claude reported no new issues; Copilot could not review because of its quota.

The final local full-suite run exposed an existing Kafka startup/shutdown race in `TestSyncPeerID_SurvivesStop`: `KafkaAsyncProducer.startInMemory` calls the first `publishWg.Add(1)` concurrently with `Stop` calling `Wait`. The same race reproduced on unmodified main `82771682f` with `go test -race -tags testtxmetacache -count=100 -run '^TestSyncPeerID_SurvivesStop$' ./services/legacy/netsync`. The relevant Kafka code, test, and manager lifecycle functions are unchanged by this PR. The reproduction and lifecycle-race diagnosis are documented for separate follow-up.


services/blockchain/fsm.go
```diff
@@ -60,10 +61,22 @@ func AvailableEventsForState(state string) []string {
...
 	// Define callbacks
 	callbacks := fsm.Callbacks{
+		"before_event": func(ctx context.Context, e *fsm.Event) {
+			storeCtx, cancel := b.fsmStoreContext(context.WithoutCancel(ctx))
+			defer cancel()
+			if err := b.store.SetFSMState(storeCtx, e.Dst); err != nil {
+				// Cancel before looplab installs its pending transition or changes
+				// state, so an explicit retry remains possible. A failed write may
+				// have committed: never attempt an unsafe compensating rollback.
+				e.Cancel(errors.NewStorageError("failed to persist FSM transition from %s to %s", e.Src, e.Dst, err))
```
freemans13, line 81: Worth tracing who consumes this new error before merge. The one that matters is `restoreFSMState` in `services/blockvalidation/catchup.go:1431`: it logs the error from `Run` and returns, so a store write that fails here leaves the node in CATCHINGBLOCKS with nothing to retry it.

Block assembly gates mining on `IsFSMCurrentState(..., FSMStateRUNNING)` (`services/blockassembly/Server.go:712`, `:800`, `:1454`), and blockvalidation and subtreevalidation each park a goroutine on `WaitForFSMtoTransitionToGivenState(..., FSMStateRUNNING)` (`services/blockvalidation/Server.go:1087`, `services/subtreevalidation/Server.go:572`). Catchup only runs when the node is behind, and a node that has just caught up is at the tip, so there is no natural second attempt.

A bounded retry in `restoreFSMState`, or an explicit `Warnf` naming the consequence, would close it. Either way, a test that asserts the FSM is still CATCHINGBLOCKS after the write fails would pin whichever behaviour you choose, where a test asserting only that `Run` returned an error would not.

services/blockchain/Server.go
```diff
@@ -2879,29 +2882,30 @@ func (b *Blockchain) SendFSMEvent(ctx context.Context, eventReq *blockchain_api.
...
+	// Once admitted, complete the local transition even if the RPC is cancelled.
+	// looplab/fsm v1.0.2 leaves a pending transition behind when its context is
+	// cancelled during Event, rejecting every subsequent event until restart.
+	transitionCtx := context.WithoutCancel(ctx)
+	err := b.finiteStateMachine.Event(transitionCtx, eventReq.Event.String())
 	if err != nil {
 		b.logger.Debugf("[Blockchain Server] Error sending event to FSM, state has not changed.")
-		switch err.(type) {
-		case fsm.InvalidEventError, fsm.NoTransitionError:
+		switch eventErr := err.(type) {
```
freemans13, line 2944: This commit removed the `Errorf` that read `Error setting the state in blockchain db`, which was the blockchain service's only record of a failed FSM write. What is left on this path is the `Debugf` on the line above, visible only at debug level, and in the committed-then-errored case your description calls out it says something untrue: the persisted state did change even though memory did not.

The callers do log it (`services/blockvalidation/catchup.go:1431`, `services/legacy/netsync/manager.go:800`, `services/asset/httpimpl/fsm_handler.go:171`), so nothing is fully silent. But an operator reading the blockchain service log after a restart that came back in an unexpected state now has no line to find. An `Errorf` in the new `fsm.CanceledError` branch, naming source, destination and the fact that the database may already hold the destination, would put it back where it is most useful.

services/blockchain/Server.go
```diff
@@ -2818,14 +2818,17 @@ func (b *Blockchain) IsFullyReady(ctx context.Context) (bool, error) {
 }
 
 // SendFSMEvent sends an event to the finite state machine and returns the state
-// reached by an accepted transition.
+// reached by an accepted, persisted transition. On a persistence error, memory
```
freemans13, line 2868: This paragraph tells the operator to retry the failed event, and for that event's own destination it works. It does not hold in the opposite direction, which is the one an operator is likely to take.

Take a RUN whose write commits and then reports an error: memory is IDLE, the database holds RUNNING. The operator sees the error and calls `Idle` to be sure the node is parked. `Idle` short-circuits on `b.finiteStateMachine.Is(IDLE)` (`Server.go:3055`), returns success without writing, and the node restarts into RUNNING with block assembly enabled and the checkpoint gate never consulted, because `Init` restores with `SetState` rather than a transition. `Run` and `CatchUpBlocks` have the same shape (`:2991`, `:3011`).

Your description says the convenience RPCs do not reconcile an ambiguous write, so this may be a deliberate deferral. If you want it cheap: set a flag under `fsmMu` when a persist fails, and have the three short-circuit branches rewrite the state once while the flag is set, rather than writing on every no-op call.

settings/blockchain_settings.go
```diff
@@ -5,6 +5,10 @@ import (
 	"time"
 )
 
+// DefaultBlockchainStoreDBTimeoutMillis supplies both the configured database
+// timeout default and the FSM fallback for non-positive configuration.
+const DefaultBlockchainStoreDBTimeoutMillis = 5000
```
freemans13, line 10: Nit: the struct tag on `StoreDBTimeoutMillis` still carries `default:"5000"` as a literal. `settings/export.go:138` reads that tag to generate the settings documentation, so the comment's claim that this constant supplies both defaults holds for `settings.go:328` and the FSM fallback but not for the exported docs. Three places, two of them now linked.

freemans13 (changes requested):
Requesting changes on one item. The persistence ordering itself is right and I could not break it.

I checked the looplab v1.0.2 mechanics against the source rather than trusting the comments. In `fsm.Event`, the `f.transition != nil` guard and the `f.current == dst` shortcut both sit outside the window where `before_event` runs, so cancelling there cannot leave the pending transition that wedges the FSM, and no entry in `FSMTransitions` has a source state equal to its destination, so the `NoTransitionError` path is unreachable anyway. `CanceledError` is a value type carrying `Err`, so the unwrap in `SendFSMEvent` gets the storage error and its gRPC code intact. The bounded store context, the detachment from caller cancellation, and the refusal to roll back an ambiguous write all read correctly.

The blocking item is a consumer of the new failure mode rather than a line this PR touches.

## 1. The only automatic promotion to RUNNING now has an unretried failure path

`restoreFSMState` logs the error from `Run` and returns (`services/blockvalidation/catchup.go:1431`). Before this PR a failed FSM write was logged inside the server and the RPC still returned success, so the node reached RUNNING in memory. Now the same store blip makes `Run` return an error and the node stays in CATCHINGBLOCKS, in memory and on disk.

What that costs: block assembly gates mining on `IsFSMCurrentState(..., FSMStateRUNNING)` (`services/blockassembly/Server.go:712`, `:800`, `:1454`), and blockvalidation and subtreevalidation each park a goroutine on `WaitForFSMtoTransitionToGivenState(..., FSMStateRUNNING)` (`services/blockvalidation/Server.go:1087`, `services/subtreevalidation/Server.go:572`). Catchup is the thing that promotes the node, and it only runs when the node is behind. A node that has just finished catching up is at the tip, which is exactly when nothing triggers catchup again. So a one second database hiccup at the end of catchup can leave a node out of RUNNING and not mining until someone notices.

I am not arguing the old behaviour was better. Reaching RUNNING in memory with a stale row on disk is the bug you are fixing. But this needs to be a decision rather than a side effect: a bounded retry in `restoreFSMState`, or at minimum a `Warnf` that names the consequence, plus a test that asserts the end state (FSM still CATCHINGBLOCKS, promotion not silently assumed) rather than asserting that `Run` returned an error.

## Non-blocking

Three smaller things are inline: the convenience RPCs can confirm a state the database does not hold, the blockchain service's own log no longer records a failed FSM write, and the shared timeout default is shared by two of the three places that carry it.

Locally on `fix/fsm-durable-transitions` at `1be11dde7`: the new FSM tests pass, pass again under `-race`, and `go vet ./services/blockchain/... ./settings/...` is clean. I did not run the full suite, and I did not exercise this against a real PostgreSQL blockchain store. That matters for one case: the committed-then-errored write your description builds on is a lost acknowledgement over a network, which SQLite cannot really produce, so `TestSendFSMEvent_AmbiguousWriteCanRetry` simulates it rather than reproducing it.

I also did not review `7112401c6` (bound FSM transition work), since that is #1695's commit riding along here.

icellan (changes requested):
The core change is sound. `before_event` cancelling via `e.Cancel` returns `CanceledError` before looplab installs `f.transition` (`fsm@v1.0.2/fsm.go:330-333, 372`), so there's no wedge, and `context.WithoutCancel` closes the pre-existing cancellation-wedge hole. Build, vet and the full `services/blockchain` package pass locally.

What I want addressed is the behavioural fallout of turning a logged-and-ignored persistence failure into a hard abort, plus the timeout semantics.

## Blocking

**`services/blockvalidation/catchup.go:1431`** — `restoreFSMState` is the only production sender of RUN after catchup, and it only logs the error, no retry. Before this PR a transient `SetFSMState` failure was logged and the FSM still reached RUNNING. Now the transition aborts, so a single DB blip at catchup completion parks the node in CATCHINGBLOCKS indefinitely — no live subtree validation, nothing feeding block assembly — until an operator runs `setfsmstate`. If strict persistence is the goal, this caller needs a bounded retry.

**`services/blockchain/Server.go:2955`** — `guardRunBelowHighestCheckpoint` now bounds `GetBestBlockHeader` at `StoreDBTimeoutMillis` (default 5s); it was unbounded before. On a loaded Postgres store a >5s header read turns into `cannot read best block header to evaluate RUN gate` and RUN is refused, feeding the same no-retry wedge above. Confirm 5s is comfortably above p99 for that query on production stores.

**`services/blockchain/fsm.go:70`** — `context.WithoutCancel(ctx)` strips the caller's deadline as well as cancellation, and `fsmStoreContext` then applies the full `StoreDBTimeoutMillis`. With a tuned-up value (say 60000) `fsmMu` *and* looplab's `eventMu` stay held for up to 60s after the gRPC client has given up, and every subsequent `SendFSMEvent` queues behind it. Either `min(callerRemaining, storeTimeout)` with a floor, or document that this key must stay small.

## Non-blocking

- `Server.go:2877` + `fsm.go:73` — one `SendFSMEvent(RUN)` can now consume two independent `StoreDBTimeoutMillis` budgets (guard read, then FSM write), so worst-case lock hold is 2x the configured DB timeout. The setting's `longdesc` reads as a per-operation bound.
- `settings/blockchain_settings.go:10` — `fsmStoreContext` treats non-positive as "use 5000", but sibling `coinbase_store_dbTimeoutMillis` documents `0` as "no timeout (wait indefinitely)". `blockchain_store_dbTimeoutMillis` had no consumer before this PR, so this PR is defining its semantics for the first time — and silently inverts the convention for anyone who sets `0`. Honour `0` as unbounded, or say in the `longdesc` that `0` means the default, not "off".
- `Server.go:2822` / `fsm.go:73` — persist-before-transition inverts the divergence direction on an ambiguous write: the store can now be ahead of memory (exactly what `TestSendFSMEvent_AmbiguousWriteCanRetry` asserts — DB `IDLE`, memory `RUNNING`). A crash in that window brings the node back up in a state it never actually entered. Previously divergence went the other way, which is the safer direction for restart. The comment acknowledges it; I'd rather see an explicit sign-off than a comment.
- `Server.go:402,430,442` — `Init` still only logs `SetFSMState` failures on fresh-node boot and the LEGACYSYNCING migration path, so the node can come up believing it's in CATCHINGBLOCKS while the store says otherwise, and the next restart re-derives a different state. With every runtime transition now strict, this asymmetry is the odd one out.
- `fsm.go:70` — the `before_event` callback dereferences `b.store` / `b.settings`, so any `*Blockchain` constructed without a store that fires an FSM event nil-panics. That's why `fsm_test.go` had to swap `New(ctx, logger, settings, nil, nil)` for the store-backed helper. `cmd/fsm_visualizer/main.go:14` builds the FSM on a nil `*Blockchain` and is safe today only because it never fires an event — one line away from a panic. A nil guard, or a doc note on `NewFiniteStateMachine`.

## Nits

- `Server.go:2896` folds `fsm.InTransitionError` into the "event rejected in state X" `StateError`, but `InTransitionError` means the FSM is wedged — a sticky condition, not a rejected transition. Callers that shrug off `StateError` will now silently swallow it.
- `TestSendFSMEvent_PersistenceTimeoutDoesNotWedge` mutates `b.settings.BlockChain.StoreDBTimeoutMillis`; fine only because each test builds its own settings.

ordishs (changes requested):
The ordering fix is correct. I verified the looplab/fsm v1.0.2 diagnosis against the module source: `beforeEventCallbacks` returns `CanceledError{e.Err}` before `f.transition` is installed, so `e.Cancel` in `before_event` cannot wedge the machine, and the pre-existing wedge on a cancelled context is real. The transition table has no self transitions, so the looplab no-op path that would write on an error return is unreachable. `go vet`, the full `services/blockchain` and `settings` packages, the focused tests under `-race`, and `golangci-lint` all pass on `1be11dde`.

Two items to address before merge, and one that can follow.

**services/blockchain/Server.go:2891-2901**, **services/blockchain/fsm.go:70-79** — P1: a failed FSM persist writes no error to the server log.

- The diff deletes `b.logger.Errorf("[Blockchain Server] Error setting the state in blockchain db: %v", err)`.
- The replacement path logs only `b.logger.Debugf("[Blockchain Server] Error sending event to FSM, state has not changed.")`, which is debug level and omits `err`. The `before_event` callback logs nothing.
- Failure: make the store reject writes, send STOP. The RPC returns the storage error, but the service log records nothing at INFO or ERROR about the database failure. That weakens the operational trust this PR sets out to build.
- Recommendation: log at error level in the `fsm.CanceledError` branch or in `before_event` before `e.Cancel`, with the event, source, destination, and wrapped error.

**services/blockchain/fsm.go:70-79**, **services/blockvalidation/catchup.go:1421-1435** — P1: the automated CATCHINGBLOCKS-to-RUNNING promotion has no retry, so one failed write can strand a node.

- `restoreFSMState` logs `failed to send RUN event` and returns. Nothing retries.
- Before this change a store write error was logged and the transition still completed in memory. Now the transition fails outright.
- Failure: the store rejects one `SetFSMState` as catchup finishes. The node stays in CATCHINGBLOCKS. Per the Init comment at Server.go:419-427, RUNNING is what enables live subtree validation and block-assembly transaction feeding. Recovery needs an operator RUN or a later catchup cycle, and a later cycle needs a new chain gap to appear. The old behaviour was self-healing across a restart, because the restart restored the stale CATCHINGBLOCKS state and catchup ran again.
- Recommendation: add a bounded retry for `ERR_STORAGE_ERROR` in `restoreFSMState`, or retry the write inside `before_event` before cancelling, and escalate the log.

**services/blockchain/fsm.go:73**, **services/blockchain/Server.go:448** — P1: the write can now lead the in-memory state, and the boot path never re-evaluates the checkpoint gate.

- Init restores with `b.finiteStateMachine.SetState(stateStr)`, which fires no callbacks and runs no guard.
- Before this change the store could only lag the FSM, never lead it. An ambiguous write now leaves RUNNING in the store while memory stays at the source state and the RPC reports failure.
- Failure: latent — no single-step reachable path found. It needs an ambiguous RUN write above the highest checkpoint, then a later drop of the tip below it through `cmd/rewindblockchain` or a checkpoint bump in go-chaincfg. The next restart boots into RUNNING with the gate bypassed, which is the property `40faeb0f0` added. The argument that commit made against depending on a boot default applies to depending on the persisted state too.
- Recommendation: run `guardRunBelowHighestCheckpoint` in Init when the restored state is RUNNING, and fall back to CATCHINGBLOCKS when it rejects.

Missing test scenarios: no test asserts a failed persist reaches the log; none covers a storage error at the catchup promotion boundary; none asserts Init refuses to restore RUNNING below the highest checkpoint.

P2 and P3 notes on the `StoreDBTimeoutMillis` `longdesc` (it still claims retry logic that does not exist, and the setting had no production consumer before this PR) and on the persistence test helper (it shares the global `chaincfg.RegressionNetParams` by pointer, and hand-builds settings so every test exercises the non-positive fallback rather than the configured default) are minor and can be folded in at your convenience.

icellan (approved):
Re-reviewed at `5020e8f44`. All three blockers from my last pass are addressed:

- `catchup.go:1429` — `restoreFSMState` now retries with cancellable backoff and a `retryableFSMPromotionError` classifier that correctly treats Canceled as non-retryable. Good.
- `Server.go:3017` — the checkpoint read passes the caller's `ctx` through `fsmStoreContext`, so the caller deadline is retained rather than replaced, and a refusal is now retried by the caller above.
- `settings/blockchain_settings.go:22` — the `WithoutCancel` behaviour is unchanged but now an explicit contract: two independent budgets on RUN, "a caller may time out while a write is still completing", "large values extend lock hold after a caller disconnects", and `0`/negative documented as "use the default; never disables the bound". That resolves the semantics ambiguity. Documented contract is fine here.

`go vet` on the affected packages is clean apart from the four pre-existing `test/utils` copylocks warnings the PR body already calls out; `-race` on the FSM and promotion tests is green.

Approving. Two things I'd like fixed before merge, both cosmetic-but-operator-facing:

**`services/blockvalidation/catchup.go:1457`** — the terminal `Warnf` fires unconditionally on every non-success exit, including two cases where the text is actively wrong:
- with an already-cancelled ctx (the normal shutdown teardown of `defer u.restoreFSMState(ctx, ...)` at line 1187) it logs "not confirmed after **0** attempts" having never called `Run`;
- on the deliberate `automatic RUN refused from IDLE` refusal it tells the operator the node "may remain CATCHINGBLOCKS with mining disabled" when the node is IDLE by the operator's own choice.

Branch the message on `attempts == 0` and on the IDLE refusal.

**`services/blockchain/Server.go:3050`** — same root cause. In the uncertain-reconcile branch for `RUNNING`, a checkpoint-gate failure returns an error even though the node is already RUNNING in memory. `restoreFSMState` sees a non-retryable `StateError`, stops after one attempt, and emits the same misleading warning. Fail-closed is the right call; the message isn't.

Non-blocking, for whenever:

- `services/blockchain/Server.go:3062` — `automatic RUN refused from IDLE` is a real behaviour change for any deployment whose boot state is IDLE (`operator`, `docker.m`). Legacy netsync reaching "current" used to self-promote such a node and now cannot, and all three `netsync` call sites only `Errorf`/`Infof` the refusal, so there's no operator-visible signal beyond logs that the node is parked. `blockchain_reference.md` covers it; `docs/howto/miners/minersHowToInteractWithFSM.md` doesn't and probably should.
- `services/blockchain/Client.go:2007` — dropping the cached-state short-circuit from `Run` means `services/legacy/netsync/manager.go:1867` issues a gRPC round-trip plus an `Infof` on every block message once `sm.current()` is true, and `manager.go:774` on every `startSync`. Each takes server-side `fsmMu`, which a RUN transition can hold for two `StoreDBTimeoutMillis` budgets. Volume is negligible at block cadence, but coupling a no-op keepalive RPC to a slow-store transition is new.
- `services/blockchain/fsm.go:72` — the nil-receiver contract is now a doc comment, but `before_event` still dereferences `b.store` and `b.settings` unconditionally, and `fsm_visualizer/main.go:15` builds the FSM on a nil `*Blockchain`. It's safe only because it never fires an event. `if b == nil || b.store == nil { return }` would make the contract enforced rather than documented.

services/blockchain/fsm.go
oskarszoon: Fixed in `5020e8f44`. `restoreFSMState` now makes at most three attempts with cancellable one-second backoff for transient storage/transport failures, including wrapped checkpoint-read failures. SQLite-backed tests assert recovery and that exhausted writes leave both memory and storage in CATCHINGBLOCKS. A retry after a concurrent operator STOP preserves IDLE. The focused race regressions pass. The later warning-wording refinements requested by icellan are separate remaining followups before merge.

services/blockchain/Server.go
oskarszoon: Fixed in `5020e8f44`. The persistence callback now logs at Errorf in the blockchain service, naming source, destination, unchanged in-memory state, and the possibility that the database already contains the destination. Failed reconciliation also logs at Errorf. The debug message no longer claims the persisted state is unchanged.

services/blockchain/Server.go
oskarszoon: Fixed in `5020e8f44`. A failed write marks persistence uncertain under `fsmMu`. Run, Idle and CatchUpBlocks serialize their authoritative state checks with transitions and rewrite an already-current target while uncertain; only an acknowledged write clears the flag. Clean no-ops do not write or notify. Removed the client cached-success bypasses as well. The real client/server plus SQLite regression covers all three commands after commit-then-error, failed reconciliation, recovery and no extra notifications. RUNNING reconciliation retains the checkpoint gate; merged #1693 also checks RUNNING on startup.

settings/blockchain_settings.go
oskarszoon: Fixed in `5020e8f44`. Go struct tags remain literals, so `TestBlockchainStoreTimeoutDocumentationDefault` reads the exported default tag and checks it against `DefaultBlockchainStoreDBTimeoutMillis`. The comment now accurately says the constant supplies the loader/fallback defaults and the test keeps the documentation tag in sync. The settings regression passes.

services/blockchain/fsm.go
freemans13: Confirmed fixed. Three attempts with cancellable one-second backoff, and `retryableFSMPromotionError` correctly excludes the below-checkpoint and IDLE refusals so a permanent rejection does not burn retries on something that can never succeed. The tests assert the persisted store contents rather than the call: `TestRestoreFSMState_ExhaustedPersistenceRetriesRemainCatching` pins both memory and storage at CATCHINGBLOCKS, and `TestRestoreFSMState_RetryPreservesOperatorIdle` covers the STOP-during-retry race I had not thought of. All pass here under `-race`.

One leftover, minor. With an already-cancelled context the loop breaks before the first attempt and still reaches the final warning, so a shutdown during catchup logs `RUN promotion not confirmed after 0 attempts; node may remain CATCHINGBLOCKS with mining disabled`. That reads as an incident during an ordinary event. An early return when `ctx.Err() != nil` at entry would keep the warning meaningful (`services/blockvalidation/catchup.go:1433`, `:1458`).

services/blockchain/Server.go
freemans13: Confirmed fixed. The `Errorf` in `before_event` names the source, the destination, the unchanged in-memory state and the possibility that the database already holds the destination, which is exactly the line an operator needs after an unexpected restart state. The debug message no longer claims the persisted state is unchanged. Nothing outstanding from me on this one.

services/blockchain/Server.go
freemans13: Confirmed fixed, and the design is better than what I suggested. Holding `fsmMu` across the authoritative state check and the reconciling write closes a race I did not raise, and dropping the client-side cached-state short circuits means a stale cache can no longer answer on behalf of the authority. `TestClient_ConvenienceCallsReconcileUncertainPersistence` drives the real client through the real server, asserts the persisted state, proves a failed reconciliation cannot report success, and asserts no extra notification and no extra write on an acknowledged clean no-op.

One consequence to look at, raised in the review rather than here because the file is outside this diff: refusing automatic RUN from IDLE leaves three callers in `services/legacy/netsync/manager.go` asking on every block, header and peer event with no backoff.

settings/blockchain_settings.go
freemans13: Confirmed fixed. `TestBlockchainStoreTimeoutDocumentationDefault` reads the exported tag and compares it to the constant, so the documentation value cannot drift silently. The rewritten `longdesc` also drops the retry claim that had no implementation and states the zero-or-negative semantics explicitly, which was the point about the coinbase key documenting zero as unbounded.

services/blockchain/Server.go
```diff
@@ -3018,49 +3045,51 @@ func HighestCheckpointHeight(checkpoints []chaincfg.Checkpoint) uint32 {
...
+		}
 		return &emptypb.Empty{}, nil
 	}
-
-	req := &blockchain_api.SendFSMEventRequest{
-		Event: blockchain_api.FSMEventType_RUN,
+	// Automatic promotion must not undo an operator STOP, including between
+	// attempts after a transport error. Explicit operator RUN uses SendFSMEvent.
+	if event == blockchain_api.FSMEventType_RUN && current == blockchain_api.FSMStateType_IDLE.String() {
+		return nil, errors.WrapGRPC(errors.NewStateError("automatic RUN refused from IDLE; use SendFSMEvent for an explicit operator RUN"))
```
freemans13: This refusal is right, and it is the change that finally makes an operator STOP stick. But it turns a call that used to succeed into one that can never succeed, and three callers were not told.

`services/legacy/netsync/manager.go` calls `Run` at `:774` when a peer is at our height, at `:1867` on a block message once `sm.current()` is true, and at `:2547` inside the `blockHandler` message loop. All three are automatic promotions, all three only log the failure, and none backs off.

The one at `:2547` is the one to look at. It is guarded by "current FSM state is not RUNNING", and that guard used to close itself, because the RUN succeeded and the node became RUNNING. It cannot close now. So a node an operator has parked in IDLE, with legacy peers still connected, issues a blockchain round trip and logs a failure on every peer, header and block message for as long as it stays parked. Transaction volume is not in play, since inv processing is already gated on RUNNING at `:2280`, so this is per block and per peer event rather than a flood. It is still an unbounded repetition of a call whose answer cannot change.

The cheap fix is at the callers: only attempt automatic promotion when the authoritative state is CATCHINGBLOCKS, which is the condition `restoreFSMState` used to check before you moved admission into the server. Alternatively have those three sites treat the IDLE refusal as terminal and stop asking until the state changes.

test/utils/helper.go
```diff
@@ -1266,28 +1267,19 @@ func SendEventRun(ctx context.Context, blockchainClient blockchain.ClientI, _ ul
...
 			return errors.NewError("Timeout waiting for Blockchain service", err)
 		default:
-			err = blockchainClient.Run(ctx, "test")
+			// Test setup is an explicit operator action and may leave IDLE.
+			// Automatic Run intentionally preserves operator IDLE.
+			state, stateErr := blockchainClient.GetFSMCurrentState(ctx)
+			if stateErr == nil && state != nil && *state == blockchain.FSMStateRUNNING {
+				return nil
+			}
+			err = blockchainClient.SendFSMEvent(ctx, blockchain_api.FSMEventType_RUN)
```
freemans13, line 1276: This is the right fix, and there is a sibling that did not get it. `daemon/test_daemon.go:601` still does `blockchainClient.Run(ctx, "test")` under `require.NoError` when a test asks the daemon for `FSMState: RUNNING`, which is explicit test setup expressed through the automatic promotion RPC, exactly the pattern you replaced here.

It does not fire today, because a fresh node boots into CATCHINGBLOCKS and RUN from there is still allowed. It becomes a hard test failure the moment a test boots into IDLE, and `blockchain_initializeNodeInState` documents IDLE as a supported value in this PR's own settings text. Same two-line change as here.

freemans13 (approved):
Approving. Both items I blocked on are properly fixed, and the reconciliation design is better than the flag-and-rewrite I suggested.

What I verified on `5020e8f44` rather than took on trust: the build is clean, `go test -race` passes for the blockchain FSM tests, the settings tests, and the new catchup promotion tests. The tests assert end state rather than the call, which is the part that matters. `TestRestoreFSMState_ExhaustedPersistenceRetriesRemainCatching` pins memory and storage together, `TestClient_ConvenienceCallsReconcileUncertainPersistence` drives the real client through the real server and checks that a failed reconciliation cannot report success, and both prove the absence case as well: no extra notification, no extra write on an acknowledged no-op.

One thing worth saying out loud, because I do not think it has been named. Before this change, an operator STOP did not stick on a node running legacy sync at all. The sync manager promoted the node back to RUNNING on the next block, because RUN from IDLE was a valid transition and nothing distinguished an automatic promotion from an operator's own request. Refusing automatic RUN from IDLE is what finally makes STOP mean STOP, which is the durable-STOP goal behind #1670. That is the strongest argument for the change and it deserves a line in the description.

Three non-blocking findings, all consequences of that new refusal rather than defects in the code you wrote.

## 1. Three legacy sync callers now ask forever

Inline on the refusal. Short version: `services/legacy/netsync/manager.go` calls `Run` from three places, all of them automatic, and all of them only log the error. The one in the message loop is guarded by "state is not RUNNING", a guard that used to close itself because the call succeeded. It cannot close now, so a parked node repeats a refused round trip on every block, header and peer event, forever, with no backoff.

## 2. The sibling test helper was not updated

Inline on the helper you did fix. `daemon/test_daemon.go:601` still uses `Run` for explicit test setup under `require.NoError`, which is the pattern you replaced in `test/utils/helper.go`.

## 3. A normal shutdown now logs an incident

Raised in the thread on the retry loop. An already-cancelled context breaks before the first attempt and still reaches the final warning, so shutting down during catchup prints `RUN promotion not confirmed after 0 attempts; node may remain CATCHINGBLOCKS with mining disabled`.

## On the other reviews

ordishs still shows CHANGES_REQUESTED against `1be11dde7`, which no longer exists after the rebase, so that verdict is stale rather than outstanding. His third P1, that `Init` restores a persisted RUNNING without re-running the checkpoint gate, is genuinely resolved: `e53520fdf` from #1693 is now the tip of main and gates a restored RUNNING through `evaluateRunCheckpoint`, falling back to CATCHINGBLOCKS. That came from the rebase, not from this PR.

icellan's lock-hold and two-budget points are documented in the rewritten `longdesc` rather than bounded in code. That is a defensible call and he has approved, but it is a documented behaviour, not a fixed one.

## What I did not check

I did not run the full suite, and I still have not exercised this against a real PostgreSQL blockchain store, which is the only place the lost-acknowledgement case this design turns on can actually happen.

services/blockchain/Server.go
oskarszoon: Fixed in `c65f284f5`. All three recurring legacy callers now use a CATCHINGBLOCKS prefilter, so observed IDLE/RUNNING and missing/unknown state do not send redundant RUN RPCs. The cached observation is only an optimization: the server remains the authority and still rejects stale-cache attempts against actual IDLE. Later events recheck the cache, so promotion resumes after explicit catchup and cache recovery without latching a pause or dropping work. Real SQLite/server tests exercise repeated message-loop/startSync events, cache recovery and stale CATCHINGBLOCKS against operator IDLE. Focused and full netsync race suites pass.

test/utils/helper.go
oskarszoon: Fixed in `c65f284f5`. Daemon RUNNING setup now sends explicit `SendFSMEvent(RUN)`. A StateError is confirmed through the authority with Run so already-RUNNING setup remains idempotent, including a concurrent promotion; other errors propagate. There is no cached-success shortcut. Real SQLite/authority tests cover IDLE, CATCHINGBLOCKS and RUNNING, stale cached values in both directions, and repeated setup. The full daemon race suite passes.

services/blockchain/fsm.go
oskarszoon: The shutdown followup is fixed in `c65f284f5` as well. An already-cancelled context returns before attempting promotion, and cancellation during backoff does not emit a terminal incident warning. Expected operator-IDLE refusal logs at Info. Other failures say RUNNING was not durably confirmed without claiming the current state is CATCHINGBLOCKS, also addressing icellan's uncertain-RUNNING reconciliation case. All three diagnostic regressions were observed failing before the fix and now pass in the promotion race suite.

ordishs (approved):
Re-reviewed at `c65f284f5` (base `e53520fdf`). All three round-1 P1s are addressed, and the new work is verified rather than asserted.

- **Failed-persist logging** — resolved. `services/blockchain/fsm.go:79` logs source, destination and cause at error level before cancelling; `Server.go:3065` does the same for reconciliation.
- **Promotion retry** — resolved. `services/blockvalidation/catchup.go:1429-1483` retries three times with cancellable backoff, classified by type rather than message. I confirmed the classification survives the error round trip: a `StateError` wrapping `context.DeadlineExceeded` still matches after `WrapGRPC`/`UnwrapGRPC`, storage errors match, the IDLE refusal does not. `TError.wrappedError` carries the cause over real transport, so this holds on the wire.
- **Boot checkpoint gate** — resolved, though `Server.go:456-470` is already in the merge-base, so credit belongs to the boot-policy work. `fsmPersistenceUncertain` narrows the remaining window usefully.
- **Settings documentation and the duplicated tag default** — resolved, with `settings/blockchain_timeout_default_test.go` keeping the tag in step.

One P1 remains.

**services/blockchain/Server.go:3048-3081**, **services/blockvalidation/catchup.go:1184** — P1: automatic promotion can still leave operator IDLE, in two steps.

- `sendFSMConvenienceEvent` refuses only the direct RUN edge. `CatchUpBlocks` has no equivalent guard and `IDLE -> CATCHINGBLOCKS` is legal. `catchup()` calls `setFSMCatchingBlocks` unconditionally with no FSM precondition, then `restoreFSMState` requests RUN from CATCHINGBLOCKS, which is permitted.
- Reachable at runtime: `services/blockvalidation/Server.go:1100` waits for the FSM to leave IDLE only once at startup, so after a later operator STOP the Kafka consumer keeps running and a gap still starts catchup.
- Observed against the real server and SQLite:

```
step1 direct Run refused: rpc error: code = Internal desc = automatic RUN refused from IDLE; ...
step2 automatic CatchUpBlocks from IDLE err=<nil> current=CATCHINGBLOCKS
step3 automatic Run from CATCHINGBLOCKS err=<nil> current=RUNNING
RESULT current=RUNNING persisted=RUNNING
```

- The docs state the stronger property: `blockchain_reference.md` says Run "refuses automatic promotion from operator `IDLE`", and `minersHowToInteractWithFSM.md` says automatic requests "cannot leave operator `IDLE`" and that legacy sync "no longer reverses an operator STOP". The first is true of the single edge; the operational claim is not delivered.
- Recommendation, pick one and make the docs match: (1) refuse automatic CATCHUPBLOCKS from IDLE too, and have block validation report that catchup cannot start while parked — this delivers the documented property; or (2) keep the behaviour and narrow the docs to say an operator STOP is not preserved once catchup starts. Option 1 is what the rest of the change argues for, but it needs a decision about what catchup does when refused, so it is worth an explicit choice rather than a silent doc edit. Either way, add a test that starts from IDLE, drives the automatic pair and asserts the intended outcome — there is none today.

Minor, non-blocking: the netsync handler at `manager.go:2554-2564` reads the FSM state and then `runIfCatchingBlocks` reads it again (cached, so cheap, but redundant); the comment at `test/nodeHelpers/blockchainDaemon.go:153` still describes an IDLE-to-RUNNING `Run` that is no longer possible; and `server_fsm_persistence_test.go:47` still shares `&chaincfg.RegressionNetParams` by pointer and hand-builds settings, so every test there exercises the non-positive timeout fallback rather than the configured value.

The new tests are the right shape — `server_fsm_reconciliation_test.go` routes the real client into the real server over SQLite and asserts a clean no-op performs no write; `catchup_promotion_test.go` forwards `LocalClient` to the real authority instead of asserting against a mock.

Verification on `c65f284f5`: `go build ./...` ok; `go vet` clean on the changed packages; `go test` green for `services/blockchain/...`, `settings/...`, `services/blockvalidation/...`, `services/legacy/netsync/...` and `daemon/...`; `go test -race -run 'FSM|Fsm|fsm'` green across those five packages; `golangci-lint` reports 10 prealloc issues, all in files this PR does not touch.

oskarszoon:
@ordishs Fixed the automatic catchup escape in `66ab77d5d` (rebased onto `82771682f`). Both automatic RUN and CATCHUPBLOCKS now refuse authoritative IDLE under the transition lock. Explicit SendFSMEvent(CATCHUPBLOCKS) still resumes a parked node; normal automatic catchup from RUNNING remains available.

An IDLE refusal returns before fetch/validation workers start or completion RUN is deferred. Consumer handling clears markers without consuming the per-block retry budget. Cleanup classifies the refusal as local, preserving earlier genuine peer failures and their error messages. Catchup attempt telemetry can already have been recorded before admission; immediate replay after resume remains outside this PR.

Real-server/SQLite regressions cover the automatic two-step pair after STOP, a catchup request arriving during STOP persistence, explicit resume, the consumer and cleanup paths, and concurrent/idempotent daemon setup. The relevant regressions failed before the fixes. All four affected package race suites passed, plus vet, changed-code lint (zero issues), and strict MkDocs. Full repository tests and CI are running.

Could you re-review this boundary fix?

ordishs (approved):
Re-reviewed at `2aee82d80` (base `82771682f`). The earlier five commits are byte-identical after the rebase, so this covers `66ab77d5d` and `2aee82d80` only.

**Round-2 P1 — automatic promotion could leave operator IDLE in two steps — resolved.** `Server.go:3073-3077` now refuses both RUN and CATCHUPBLOCKS from IDLE under the transition lock. Same probe as round 2, re-run against the real server and SQLite:

```
step1 automatic Run from IDLE     err=automatic RUN refused from IDLE; ...            current=IDLE
step2 automatic CatchUpBlocks     err=automatic CATCHUPBLOCKS refused from IDLE; ...  current=IDLE
step3 automatic Run again         err=automatic RUN refused from IDLE; ...            current=IDLE
RESULT current=IDLE persisted=IDLE
```

The consumer side is handled rather than left to fail noisily: `blockvalidation/Server.go:1876-1883` logs the refusal at info, `Server.go:1931` clears the markers without charging the peer, and `catchup.go:483-488` classifies it `local_fsm_refusal`. I checked that new switch case against the file's own ordering rules — it sits above the `validation_failure` branch that sets `reportMalicious`, as the neighbouring comments require — and confirmed it cannot shadow a peer-attributable error: the only `NewStateError` producer in the catchup path is `setFSMCatchingBlocks` at `catchup.go:1419`, which runs before any peer interaction. The docs match the code, including the grpcurl examples, which correctly move the operator off `Run`/`CatchUpBlocks` onto `SendFSMEvent`.

One follow-up, non-blocking.

**services/legacy/netsync/manager.go:798-802** — P2: the legacy sync path was not updated for the widened refusal.

```go
if bestPeer.LastBlock()-bestBlockHeightInt32 > 10 {
    if err = sm.blockchainClient.CatchUpBlocks(sm.ctx); err != nil {
        sm.logger.Errorf("[startSync] failed to set blockchain state to catching blocks: %v", err)
    }
}
```

Operator IDLE is now an expected refusal, not a fault, so a parked node that is more than 10 blocks behind logs an error line for a state the operator chose. This is the same case `logCatchupFSMRefusal` demotes to info in block validation and `catchup.go:1471` handles for the RUN edge. The caller also lacks the cached prefilter that `runIfCatchingBlocks` applies three lines earlier. `startSync` returns early when a sync peer is already held, so this fires once per sync-peer election rather than on every 30-second tick; re-election while parked is likely but I did not prove a cadence. Suggest a `catchUpIfNotIdle` sibling, or classify at the call site. No test covers it — `fsm_promotion_test.go` covers only the RUN edge.

Minor: `daemon/test_daemon.go:167-177` falls back to `CatchUpBlocks` on any `ErrStateError`, so a genuine store failure during the explicit event surfaces as the IDLE refusal instead of the real cause (test diagnostics only; `setTestDaemonRunning` has the same shape). And `server_fsm_persistence_test.go:47` still shares `&chaincfg.RegressionNetParams` by pointer and hand-builds settings, so every test there exercises the non-positive timeout fallback — carried from round 1.

`catchup_idle_test.go` is well built: it drives the real authority, so the message-string coupling in `logCatchupFSMRefusal` is genuinely covered, and it asserts the properties that are easy to get wrong — no warning, no peer failure, attempt budget neither spent nor reset, markers cleared for a later retry. `TestCatchupEntry_OperatorIdleDoesNotStartValidation` notes that the old code accepted IDLE there, which makes it a real regression test.

Verification on `2aee82d80`: `go build ./...` ok; `go vet` clean on the five changed packages; `go test` green for `services/blockchain/...`, `settings/...`, `daemon/...`, `services/blockvalidation/...` and `services/legacy/netsync/...` (exit 0); `go test -race -run 'FSM|Fsm|fsm|Idle|Catchup'` green across four packages (exit 0); `golangci-lint` reports 10 prealloc issues, all in files this PR does not touch.

PR #1697, opened 2026-09-05
fix: recover assembly after FSM feed suppression

oskarszoon:
A lost FSM subscription can make a RUNNING node observe synthetic IDLE. Subtree validation then stores transactions without feeding assembly, and reconnecting does not replay them. This change adds automatic recovery without requiring a new block, restart or reorg, while preserving conservative feeding and pruning during IDLE or uncertainty.

Block assembly discovers the unmined index on a fixed-delay timer (default `blockassembly_unminedRecoveryInterval=1h`). A dedicated uncached `ReadFSMState` RPC must confirm ready, persistence-confirmed RUNNING. Deferred or failed recovery retries after `min(interval, 1m)`; zero configuration uses the default; a negative duration (for example `-1s`) disables new automatic passes. Apply configuration changes with a service restart. An already-started repair keeps its retry responsibility and mining gate. Upgrade blockchain before relying on this RPC. An older server returning Unimplemented produces an informational upgrade message; other authority/storage failures remain warnings.

The processor snapshots queued and assembled work, rechecks fresh transaction metadata and full anchored-chain membership, hydrates ancestors, and rebuilds in parent-before-child order. Recovery is read-only with respect to the UTXO store: it does not run the startup loader, unlock in-flight transactions or change mined/conflict flags. Locked transactions require prior accepted-handoff evidence. Selection failures preserve the current template and queue; successful rebuilding consumes only the captured queue prefix. Later arrivals remain queued.

A failure after memory replacement retains retry responsibility and blocks new mining and dequeue. Reset RPCs return their actual result and visibly refuse every reset variant while repair is pending. The `teranode_subtreeprocessor_recovery_pending_seconds` gauge measures pending duration across failed retries. Each queued reset executes its own options; accepted resets retain the service lifetime if their RPC caller times out. If the chain advances, recovery repairs the original assembly tip before normal reconciliation resumes mining. Existing mining jobs and async storage retain mmap ownership across replacement; accepted queued storage survives recovery completion. Replaying existing work does not distort adaptive subtree sizing.

The consumer changes require positively observed RUNNING for pruning when `pruner_skipDuringCatchup=true` and for assembly feeding across block, legacy gRPC, peer gRPC and Kafka paths. Nonlegacy subtree checks continue validating on state-read errors with feeding disabled. Expected catchup skips do not inflate suppression accounting or consume the warning budget. Upgrade notes explain that previously configured `pruner_skipDuringCatchup=true` now takes effect and can increase retained disk usage.

Validation uses real SQLite stores, the native blockchain client and in-memory Kafka. Recovery regressions cover no-new-block repair, authoritative readiness, parent ordering, mined exclusion, queue ownership, partial failure followed by chain advance, mmap mining-job leases, and accepted storage lifetime. The review follow-up also covers IDLE/catchup avoiding pending-block waits, negative-interval opt-out, retaining incomplete repair while disabled, actual reset RPC results, per-request reset options, cancellation/shutdown, and concurrent pending-duration scrapes. Full affected race suites passed: block assembly 439.562s, subtree processor 493.160s, settings 4.013s. Focused reset RPC and legacy-fixture suites passed; vet, changed-code lint and strict docs checks are clean. Independent bounded review found no blocker in the follow-up. Final-head CI passed 14,323 tests with 46 skips. The post-push check at 10:30:52 UTC verified `bdf2241d3`: 37 successful CI checks, one skip, no pending/failed checks, and zero unresolved inline threads. Fresh Sonar reports zero issues, 90.6% new-code coverage and no duplication. Claude's final review confirms all three findings are addressed with no blockers. The benchmark report found no significant regressions in 139 comparisons (baseline commit unspecified).

The recommended day-long mining-node soak remains outstanding; CI is not a live-node soak.

Operational cost: recovery scans the unmined index and rebuilds metadata/order in memory, consuming CPU, RAM and storage I/O while queue consumption pauses. The hourly default limits frequency; it is not a benchmark-derived capacity guarantee. IDLE remains control-plane intent, not global quiescence; rewind requires service shutdown.

Consumer-safety prerequisite for #1698 under #1670, replacing part of #1672. Includes merged #1693 and #1696. Siggi's recovery finding is implemented here and needs human re-review; draft #1698 remains separate.


settings/pruner_settings.go
```diff
@@ -6,7 +6,7 @@ import "time"
...
...ed" default:"false" category:"Pruner" usage:"Suppress pruning in catchup, IDLE, or unknown FSM states" type:"bool" longdesc:"### Purpose\nWhen enabled, starts pruning only after the blockchain FSM is positively known to be RUNNING. CATCHINGBLOCKS, IDLE, missing states, unknown states, and FSM read errors suppress the cycle.\n\n### How It Works\n- When enabled, the pruner checks FSM state before starting preservation, expiry, transaction deletion, or blob-deletion scheduling\n- Catchup writes already in progress may finish after STOP enters IDLE, so IDLE must not authorize a new pruning cycle\n- Once the node transitions to FSMStateRUNNING, the pruner executes normally with Phase 1 (preservation) before Phase 2 (deletion)\n\n### Trade-offs\n| Setting | Benefit | Drawback |\n|---------|---------|----------|\n| true | Pruning starts only in RUNNING | Temporary storage growth during catchup or IDLE |\n| false (default) | Storage efficient during catchup | Only safe with retention >= 288 blocks |\n\n### Recommendations\n- **false** (default) - Recommended for production deployments with retention >= 288 blocks\n- Set to **true** for development/testing with aggressive retention values"`
```
freemans13, line 9: This tag is the whole reason the pruner change cannot run. `settings/settings.go` has no `SkipDuringCatchup:` entry in its `Pruner: PrunerSettings{...}` block, and settings there are loaded by explicit `getBool` calls rather than by reflection over these tags. The field stays false however the key is set.

Verified on this branch by setting `pruner_skipDuringCatchup=true` in config and loading settings: `SkipDuringCatchup` came back `false`, while `SkipPreserveParents`, which does have a loader call, came back `true`.

One line beside the other pruner keys fixes it:

```go
SkipDuringCatchup: getBool("pruner_skipDuringCatchup", false, alternativeContext...),
```

`SkipProcessExpiredPreservations` on line 10 is dead in the same way, and `ForceIgnoreBlockPersisterHeight` has no loader and no reader at all. Both predate this PR.

services/pruner/worker.go
```diff
@@ -158,17 +158,17 @@ func (s *Server) prunerProcessor(ctx context.Context) {
...
 				fsmState, err := s.blockchainClient.GetFSMCurrentState(ctx)
 				if err != nil {
 					s.logger.Warnf("Failed to get FSM state, skipping pruner: %v", err)
 					prunerSkipped.WithLabelValues("fsm_error").Inc()
 					continue
 				}
-				if fsmState != nil && *fsmState == blockchain.FSMStateCATCHINGBLOCKS {
-					s.logger.Debugf("[pruner][%s:%d] skipping during catchup", blockHashStr, blockHeight)
-					prunerSkipped.WithLabelValues("catchup_mode").Inc()
+				if fsmState == nil || *fsmState != blockchain.FSMStateRUNNING {
```
freemans13: This branch never executes on a real node, because the `SkipDuringCatchup` guard two lines up is always false (no loader call in `settings/settings.go`). So `fsm_not_running` and `fsm_error` are never emitted, and the dashboard migration note added to `pruner_reference.md` asks operators to move off a label they never had.

The logic itself is what I would want once the setting is wired. Skipping a notification defers pruning rather than losing it, since the work is height-relative and the next notification at a higher height picks it up, so widening the skip window from CATCHINGBLOCKS to everything-but-RUNNING does not accumulate debt.

services/subtreevalidation/Server.go
```diff
@@ -913,11 +913,11 @@ func (u *Server) checkSubtreeFromBlock(ctx context.Context, request *subtreevali
...
-		// While catching up blocks, disable adding transactions to block
-		// assembly: bulk-history txs do not belong in our template. In
-		// RUNNING, assembly stays enabled so txs from a legacy-bridge tip
-		// block survive in the mempool if the block loses a reorg.
-		if *currentState == blockchain.FSMStateCATCHINGBLOCKS {
+		// Only known RUNNING state permits feeding block assembly. Admitted
+		// catchup writes may finish in IDLE; their bulk-history transactions
+		// must not enter the template. RUNNING retains reorg resilience for
+		// transactions from legacy-bridge tip blocks.
+		if currentState == nil || *currentState != blockchain.FSMStateRUNNING {
```
freemans13: Worth hoisting this gate above the `BaseUrl == "legacy"` branch rather than leaving it inside. The peer-facing branch at line 941 passes no `WithAddTXToBlockAssembly` at all, and the default is `true`, so it would feed assembly in every FSM state including CATCHINGBLOCKS.

Not a live bug today: the only production caller is `services/legacy/netsync/handle_block.go:652`, which hardcodes `"legacy"`. But `CheckSubtreeFromBlock` is a gRPC entry point, so the day someone wires a non-legacy caller the rule this PR establishes silently does not apply to it.

freemans13 (changes requested):
Request changes, on one line of wiring. The subtree-validation half is solid and I verified it end to end. The pruner half cannot run on any node.

## The pruner guard is inert

`pruner_skipDuringCatchup` has a struct tag, a `settings.conf` entry on line 615, and now three paragraphs of documentation. Nothing in `settings/settings.go` reads the key. Settings there are populated by explicit `getBool`/`getInt` calls rather than by reflection over the tags, so the field stays at Go's zero value whatever an operator writes. `s.settings.Pruner.SkipDuringCatchup` is always false, and the FSM check at `services/pruner/worker.go:162` never runs in production.

I proved it against this branch by setting the key to `true` in config and loading settings, with a wired pruner key alongside as a control:

```
config says true for all three keys; loaded values are:
  Pruner.SkipDuringCatchup               = false
  Pruner.SkipProcessExpiredPreservations = false
  Pruner.SkipPreserveParents (control)   = true
```

The gap predates this PR, so it is not something you broke. It matters here because this is the PR that documents the behaviour. `docs/references/services/pruner_reference.md` now tells operators to migrate dashboards from `catchup_mode` to `fsm_not_running`, and neither label has ever been emitted by a running node. I also checked #1698, since the body points at it as the follow-up: its only `settings/settings.go` change is an unrelated blockchain timeout constant, so the key stays dead there too.

`TestPrunerSkipDuringCatchupRequiresRunning` cannot catch this. It sets the field directly on the struct at line 60, which is the right way to test the branch, and is exactly why the wiring gap stays invisible.

The fix is one line beside the other pruner keys in `settings/settings.go`:

```go
SkipDuringCatchup: getBool("pruner_skipDuringCatchup", false, alternativeContext...),
```

Then an override test shaped like `TestP2PPeerMapSettings_LoaderReadsAllKeys` in `settings/p2p_settings_test.go`, which exists to guard this exact bug class. A default-value assertion will not do it, because the dead value and the default are both false.

While you are in there, `SkipProcessExpiredPreservations` is dead the same way, and `ForceIgnoreBlockPersisterHeight` has no loader and no reader anywhere. Both predate this PR and are your call.

## The subtree-validation change is right

There are only three FSM states, so the change is precisely "IDLE stops counting as safe", and all three assembly-feeding sites move together. I followed the option to where it lands rather than stopping at the gate: `subtreesHandler` forwards its variadic options to `ValidateSubtreeInternal`, which passes them to `processMissingTransactions`, and `ProcessOptions` applies them in order over the defaults. Nothing downstream appends a competing `WithAddTXToBlockAssembly`, so yours is the last writer. The comment claiming the closure and `processTransactionsInLevels` share one captured value is accurate, both read the same `addTXToBlockAssembly`.

The new tests assert the resolved validator options rather than the fact that a gate was called, which is the assertion worth having. Moving `BlocksOnly` above the FSM read is a real improvement, and the `require.Zero(t, client.reads.Load())` assertion pins it properly.

One consumer question I chased and cleared. Transactions validated in IDLE no longer reach the template, but block assembly reloads unmined transactions on start and reset, so they come back rather than being dropped from the mempool.

## What I ran

`go test ./services/pruner/` green, and `./services/subtreevalidation/` green under `-race -tags testtxmetacache` for the new FSM tests, both on 84fbb25. CI is 37 of 38 passing with `test aerospike native-ops` skipped on the fork image. I did not run the smoke or sequential suites, and none of this has been soaked on a live node.

icellan:
Reviewed. The change itself is coherent — the nil-safety on the old `*state` derefs was a genuine panic risk, moving the `BlocksOnly` short-circuit ahead of the FSM read is right, and `findMissingSubtrees` is a pure extraction (diffed line-for-line, quorum/exists/ordering semantics identical). `go vet` clean, new tests pass. No blockers, but two things on the fail-closed flip:

- `services/subtreevalidation/check_block_subtrees.go:448` — `GetFSMCurrentState` serves a cache that is deliberately stamped `IDLE` on notification-stream loss (`services/blockchain/Client.go:1467`/`:1473`/`:1561`, "setting FSM to IDLE for safety"), corrected only on reconnect via `fetchAndRestoreFSMState`. With the gate now requiring a positive RUNNING, a RUNNING node whose stream has broken but not yet reconnected validates every accepted block's txs with assembly feeding off for the whole window — silently dropping them from the mining template and losing the reorg resilience the old comment explicitly defended. Nothing logs or counts the not-RUNNING branch, so it's invisible in production. Add a warn/counter there, and consider a direct unary read when the cached value reads non-RUNNING.

- `services/subtreevalidation/Server.go:921` — same exposure on the legacy-bridge path, and it matters more there: the comment block immediately above (~865-905) documents that `handleBlockMsg → HandleBlockDirect` runs in *every* FSM state and that a restarted node with FSM restored to RUNNING catches up over this bridge. A cached `IDLE` during that catch-up suppresses assembly feeding for exactly the tip-block txs that comment says must stay in the mempool.

Smaller things:

- `services/pruner/worker.go:169` — the RUNNING check is admission-only and sits before `waitForBlockMinedStatus`, which can block for `pruner_blockAssemblyWaitTimeout` (default 10m). A cycle admitted in RUNNING can run its whole preserve/expire/delete sequence inside a later CATCHINGBLOCKS/IDLE window — the overlap the setting exists to prevent. Re-checking the state after the mined-status wait, before Phase 1, closes most of it.
- `services/subtreevalidation/Server.go:953` — the non-legacy branch of `checkSubtreeFromBlock` still passes no assembly option, so it defaults to `AddTXToBlockAssembly=true` in CATCHINGBLOCKS and IDLE alike. The PR body claims coverage of the peer-announcement validation paths, but only the Kafka handler was gated. In-repo the only caller is netsync with `baseURL == "legacy"`, so it's reachable only by direct gRPC today — either gate it for consistency or note why it's deliberately left open.
- `services/subtreevalidation/subtreeHandler.go:69` — state is read once per Kafka message and bound into the option, but validation runs async via `g.Go`. On an IDLE→RUNNING transition, queued messages validate with `AddTXToBlockAssembly(false)` although the node is RUNNING by bless time. Recoverable via `UnminedSince`, but only on the next `loadUnminedTransactions`, not promptly.
- `services/subtreevalidation/assembly_fsm_test.go:74` — unlike its sibling `subtree_handler_fsm_test.go:64` it never sets `tSettings.SubtreeValidation.QuorumPath`, so `New()` creates the quorum directory at whatever `subtreevalidation_quorumPath` resolves to. All 12 subtests fail locally with `STORAGE_ERROR (69): Failed to create quorum path ... permission denied` on a box with a non-default data dir. `tSettings.SubtreeValidation.QuorumPath = t.TempDir()` fixes it. (`TestUnconfirmedParentsOptionDoesNotLeakToPeerPaths` has the same gap already.)

ordishs (approved):
**services/subtreevalidation/subtreeHandler.go:65-70** (same pattern at `Server.go:918-923` and `check_block_subtrees.go:448`)

The new gate cannot distinguish a paused node from an unavailable blockchain subscription, and the downgrade it applies is silent and one-way.

- `blockchain.Client.GetFSMCurrentState` serves a cached value (`services/blockchain/Client.go:1826`). That cache is set to IDLE as a fail-safe when the notification stream breaks or the heartbeat goes stale (`services/blockchain/Client.go:1465-1475`), and the blockchain server reports IDLE while `subscriptionManagerReady` is false (`services/blockchain/Server.go:2753-2755`). A genuinely RUNNING node therefore reads IDLE for the length of a blockchain-service restart.
- The old test (`*state == CATCHINGBLOCKS`) kept assembly feeding on in that window. The new test turns it off.
- The loss does not recover. The consumer runs with `kafka.WithLogErrorAndMoveOn()` (`services/subtreevalidation/Server.go:581`), so the offset commits regardless, and once the subtree is stored `TryLockIfFileNotExists` returns `SubtreeExistsError` (`subtreeHandler.go:137-145`), so re-announcement is dropped.
- Nothing logs the downgrade and no metric counts it.

Failure: node is RUNNING; blockchain service restarts for 20 s; cached state becomes IDLE; peer announces subtree S; `addToAssembly` is false; S validates, stores, commits. Transactions this node learned about only through S never reach its mining template, and re-announcing S is a no-op. No log line, no counter.

Scope: transactions already received on the direct propagation path are unaffected, so the loss is a subset. The effect is template completeness and mining revenue, not consensus.

Recommendation:

1. Log at Warn (rate-limited) and increment a counter labelled by the observed state whenever a subtree path disables assembly feeding. The pruner has `fsm_not_running`; these paths have nothing.
2. Consider a live FSM read that bypasses the cached fail-safe IDLE before the downgrade, so a transport fault is not read as a pause. Kafka redelivery is not available here — `WithLogErrorAndMoveOn` commits on error.

freemans13:
Confirming @ordishs's and @icellan's finding, and withdrawing the paragraph in my review that called the subtree-validation change correct. It is not correct. I traced the option forward to the validator and never asked where the FSM value came from, which is precisely where the bug lives.

Two people reached it independently, so the mechanism is not in doubt and I will not restate it. I verified each step on this branch and it holds.

## The recommended fix does not close the window

Both reviews suggest a direct unary read that bypasses the client's cached IDLE. That is not enough, for two reasons.

The server's own handler returns IDLE whenever `subscriptionManagerReady` is false, whatever the state machine actually holds. A direct read lands on that same branch, so it reports IDLE during blockchain-service startup just as the cache does. See `services/blockchain/Server.go:2736-2755`.

There is also nothing to call. Every read on the client returns the cached pointer first, at `Client.go:1826`, `:2059` and `:2119`, and the blockchain interface exposes no read-through variant. So "bypass the cache" means adding a method to the client and fixing the server's ready-gate, not swapping one call for another.

## The underlying shape

IDLE carries two meanings that cannot be told apart at any of these call sites. One is "an operator paused the node". The other is "I cannot reach the state machine". This PR needs the first and silently gets the second.

Separating them belongs at the source, not at each gate. Until something distinguishes them, the fail-open test this PR replaced is the safer shape, because it reads "cannot tell" as "carry on". If the fail-closed direction is kept anyway, then the warn plus counter both reviews ask for is the minimum, since the downgrade is currently invisible.

## The two pruner findings interact

@icellan is right that the RUNNING check is admission-only, and that a cycle admitted in RUNNING can run its whole preserve, expire and delete sequence inside a later CATCHINGBLOCKS or IDLE window, because the wait for mined status can block for ten minutes at the default.

That cannot happen today, for the reason in my review: `pruner_skipDuringCatchup` has no loader call in `settings/settings.go`, so the guard is never on and the branch never runs. Both need fixing. The wiring has to land first, or the re-check cannot be exercised on a real node.

The rest of my review stands, including the ungated non-legacy branch of `checkSubtreeFromBlock` that @icellan also flagged.

icellan (changes requested):
Re-reviewed at `00c534092` (16 files, +776/-110 vs merge-base `e53520fdf`). The counter and rate-limited warning address the visibility gap, the pruner re-check is correctly placed after both waits and before `blobNotify`/Phase 1, and the `state != nil` guard removes the pre-existing nil-deref. `go vet` clean on all three packages, all seven new/changed tests pass under `-race`, and `findMissingSubtrees` is a semantics-preserving extraction.

One blocker remains, plus three smaller things.

**`services/subtreevalidation/metrics.go:248` — the suppression isn't self-healing.**
`blockchain.Client` *stores* a synthetic `FSMStateIDLE` when the notification stream breaks or the heartbeat goes stale (`services/blockchain/Client.go:1467`, `:1473`). On a RUNNING node that means every subtree validated in that window gets `AddTXToBlockAssembly(false)`. The comment points at "normal unmined transaction reloads" for recovery, but that is `loadUnminedTransactions`, and it is only called from `BlockAssembler.Start` (`BlockAssembler.go:1111`) and reorg `reset` (`:662`) — there is no periodic reload. So those txs stay out of the mining template until a BA restart or a reorg. Pre-PR only CATCHINGBLOCKS suppressed feeding, so this is a new fee-loss window rather than a pre-existing one. Either make the non-RUNNING read verify against a direct unary call before suppressing, or give the unmined set a periodic reload so the window closes on its own.

Smaller:

- `services/subtreevalidation/subtreeHandler.go:67` — the gate runs before the CATCHINGBLOCKS early return, so ordinary catchup both inflates `assembly_feeding_suppressed_total{observed_state="catchingblocks"}` and emits the "may indicate subscription loss; check blockchain connectivity" warning every minute for a completely expected state. The limiter is a single instance-wide timestamp (`metrics.go:266`) shared across all four paths and all states, so a genuine synthetic-IDLE suppression on `check_block_subtrees` in the same minute never gets logged at all. Key it by `(path, observed_state)`, or skip the warning for catchup.
- `services/subtreevalidation/Server.go:826` — the FSM read moved out of the legacy-only branch, so the peer (non-legacy) `CheckSubtreeFromBlock` path now hard-fails on an FSM read error where it previously never touched the FSM. Before the first notification the client cache is nil and the call goes to gRPC, so a brief blockchain-service blip turns a valid subtree check into a `ProcessingError`. Failing closed on feeding doesn't require failing the validation — treat a read error as `addToAssembly=false` here.
- `settings/settings.go:620` — `pruner_skipDuringCatchup` was never loaded by `NewSettings` before this PR (confirmed absent at the merge base), i.e. the flag was dead. Anyone who already has it set true gets the guard activated for the first time on upgrade, now also covering IDLE/unknown/read-error, with unbounded storage growth over a long catchup. Worth an upgrade note.
- nit: real admission-gating logic in `metrics.go` is hard to find; it belongs next to the handlers.

settings/pruner_settings.go
oskarszoon: Fixed in the published branch: SkipDuringCatchup is loaded explicitly from pruner_skipDuringCatchup, with regression coverage for true/false values and context overrides. The unrelated pre-existing settings are left for separate cleanup.

services/pruner/worker.go
oskarszoon: The published fix wires the setting loader and rechecks the FSM after both readiness waits, before blob notification and Phase 1. Real-store tests cover the active guard. The upgrade documentation clarification is being prepared with the remaining review fixes.

services/subtreevalidation/Server.go
oskarszoon: Fixed in the published branch: both legacy and peer CheckSubtreeFromBlock paths receive the same explicit assembly-feeding option, enabled only for observed RUNNING. Regression coverage exercises both paths. The separate synthetic-IDLE recovery concern remains open and is being worked on before this PR is ready.

services/subtreevalidation/subtreeHandler.go
```diff
@@ -52,20 +52,24 @@ func (u *Server) subtreeMessageHandler(ctx context.Context) func(msg *kafka.Kafk
...
 
+		// In BlocksOnly mode, skip processing peer-announced subtrees (only process subtrees from blocks)
+		if u.settings.SubtreeValidation.BlocksOnly {
+			return nil
+		}
+
 		state, err := u.blockchainClient.GetFSMCurrentState(gCtx)
 		if err != nil {
 			return errors.NewProcessingError("[subtreeMessageHandler] failed to get FSM current state", err)
 		}
```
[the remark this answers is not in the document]
oskarszoon: The divergence preserves the existing Kafka behavior. This error return predates the PR, and the consumer uses WithLogErrorAndMoveOn(): with a live consumer context it logs the failure and advances the offset, rather than requesting redelivery. The gRPC change removes the validation dependency newly introduced by moving its FSM read above the legacy/peer split. I am retaining the existing Kafka policy here; the separate assembly-recovery blocker remains open, and we are not relying on Kafka retry to solve it.

icellan (dismissed):
Re-reviewed at `d7adb5ff7` against merge-base `3e60ee119` (17 files, +820/-110). `go vet` clean on subtreevalidation/pruner/settings, both packages green, FSM/assembly/suppression tests green. My blocker is addressed — approving.

What the last round fixed:

- `services/subtreevalidation/assembly_admission.go:31` — `catchingblocks` now returns before taking the warning token, so ordinary catchup no longer monopolises the budget and a genuine synthetic-IDLE suppression on the block/legacy paths actually gets logged. Counter still records every observation.
- `services/subtreevalidation/subtreeHandler.go:67` — the CATCHINGBLOCKS early return now precedes the gate, so `assembly_feeding_suppressed_total{path="kafka_subtree"}` no longer conflates "dropped without validating" with "validated but not fed".
- `services/subtreevalidation/Server.go:826` — the peer (non-legacy) path no longer aborts validation on an FSM read error; it sets `currentState = nil` and fails closed only on feeding, which matches the stated posture. Legacy still errors, which is right.
- Admission logic moved out of `metrics.go` into its own file.

On the self-healing point: the comments no longer claim automatic recovery ("recovery may require an unmined transaction reload"), so "txs validated in a suppression window stay out of the template until a BA restart or reorg" is now an accepted tradeoff rather than a wrong claim. Fine by me. Triggering `loadUnminedTransactions` on the RUNNING re-observation edge would close it properly if you want a follow-up.

Two non-blocking leftovers:

- `services/pruner/worker.go:124` — with `pruner_skipDuringCatchup=true`, a cached synthetic IDLE or a blockchain-pod restart makes every cycle skip with only a `Debugf` plus `prunerSkipped{reason="fsm_not_running"}`. Disk grows unbounded with nothing operator-visible, which is inconsistent with the rate-limited `Warnf` this PR added on the subtreevalidation side. Same treatment there would be worth it.
- `services/subtreevalidation/Server.go:1039` — `publishInvalidSubtree` still does a bare `*state ==` after only an `err` check. The premise at the three other call sites is that `GetFSMCurrentState` can hand back `(nil, nil)`, and this PR's own tests inject exactly that. Pre-existing and not reachable through the real `*Client`, but it's a one-liner that keeps the file internally consistent.

oskarszoon:
@icellan, automatic recovery is implemented in `e343801f4` (current head `b5e2cca70`, including the test-double compatibility fix). Your approval was on `d7adb5ff7`; could you re-review this added implementation?

The fixed-delay recovery timer uses a new ready, persistence-confirmed `ReadFSMState` RPC. It rebuilds eligible unmined, queued and assembled work in parent order without changing UTXO state. Exact queue-tail ownership and accepted Locked records survive retries. A partial failure keeps mining closed through repair and chain reconciliation, while leases preserve existing mmap mining jobs and queued storage writes.

Real-store regressions cover recovery without a new block, unavailable authority, actual IDLE/catchup, failed rebuild followed by chain advance, more than 1,000 chain headers, parent ordering, later queue arrivals, and submitting a mining job issued before recovery. Full affected race suites, vet, changed-code lint and strict docs passed. Independent bounded re-review found no remaining blocker in the reviewed recovery/lifetime scopes. Repository-wide tests and final-head CI are being checked; the first full run exposed the missing blockpersister test-mock method now fixed in `b5e2cca70`.

The default interval is one hour, with deferred/failed attempts retried after at most one minute. The PR documents scan/rebuild CPU, RAM and I/O cost, queue pauses, and the requirement to upgrade blockchain for the new RPC. Your two explicitly nonblocking pruner-warning/nil-guard suggestions remain recorded as follow-ups.

services/blockassembly/unmined_recovery.go
```diff
@@ -0,0 +1,156 @@
...
+
+// recoverUnminedTransactions repairs missing template entries at the current tip.
+// The channel listener owns serialization with block notifications and resets.
+func (b *BlockAssembler) recoverUnminedTransactions(ctx context.Context) (bool, error) {
+	if err := ctx.Err(); err != nil {
+		return false, err
+	}
+	// State and tip reads must not hold the assembly listener indefinitely.
+	readCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	if err := b.subtreeProcessor.WaitForPendingBlocks(readCtx); err != nil {
```
freemans13: The expensive wait runs before the cheap gate, so a node in catchup pays for a pass that cannot succeed.

`WaitForPendingBlocks` retries with exponential backoff and infinite attempts, bounded only by the five-second `readCtx` created two lines up. During catchup there are usually blocks whose mined flag is not yet set, so it burns the whole budget and returns an error. The FSM check on the next line would have rejected the pass anyway, for free.

The cost lands on the block assembly channel listener, which is the same goroutine that handles block notifications, resets and reconciles. A deferred pass shortens the timer to one minute, so this repeats every sixty seconds for the whole of catchup, with a `Warnf` each time.

Reading the state first skips all of it:

```go
state, err := b.blockchainClient.ReadFSMState(readCtx)
if err != nil {
	cancel()
	return false, err
}
if state != blockchain.FSMStateRUNNING {
	cancel()
	return false, nil
}
if err := b.subtreeProcessor.WaitForPendingBlocks(readCtx); err != nil {
	cancel()
	return false, err
}
```

The wait still precedes the tip read, so the guarantee it exists for is unchanged.

The same one-minute cadence turns any lasting unavailability into a per-minute warning. A rolling upgrade that puts this block assembly against an older blockchain service is the case I would expect to see first. The new RPC returns Unimplemented, and the node logs `Unmined transaction recovery deferred` every minute until the rollout finishes.

services/blockassembly/BlockAssembler.go
```diff
@@ -491,6 +506,9 @@ func (b *BlockAssembler) triggerReconcile() {
 // Returns:
 //   - error: Any error encountered during reset
 func (b *BlockAssembler) reset(ctx context.Context, validateInputs ...bool) error {
+	if b.subtreeProcessor.RecoveryPending() {
```
freemans13, line 514: A wedged rebuild has no operator escape hatch, and the one they will reach for reports success while doing nothing.

Once `recoveryPending` is set it is cleared only by a successful publish at the end of `recoverUnmined`. While it is set, this guard refuses `reset`, the subtree processor refuses its own reset, reorg and move-forward, `processNewBlockAnnouncement` returns early, and `GetMiningCandidate` errors. That is a coherent fail-closed posture, and `TestUnminedRecoveryPendingRejectsBlockMovementAndLegacyReset` shows it is deliberate rather than accidental.

The problem is what an operator sees. `ResetBlockAssembly` returns an empty message and a nil error unconditionally, and `resetWithOptions` runs the reset in a goroutine that only logs a failure. So the operator gets a success response, block assembly stays wedged, and the sole evidence is one `Errorf` line. A process restart is then the only fix.

Reaching this state needs `addDirectly` to fail after the destructive phase has begun. Allocating a new memory-mapped subtree inside `completeSubtree` is the plausible trigger, so a full disk under `blockassembly_subtreeMmapDir` gets you there.

Either of these would close it. Let a full reset override `recoveryPending`, since a reset rebuilds from the unmined index anyway and is strictly more thorough than the repair. Or keep refusing it and make the refusal visible, by returning the error from the reset RPCs and adding a gauge for time spent with `recoveryPending` set.

The settings doc says a failed rebuild keeps mining and dequeue closed until a retry succeeds. It does not say reset is refused as well, and that is the part an operator needs to read before they lose an hour to it.

services/blockassembly/unmined_recovery.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/bsv-blockchain/teranode/stores/utxo"
+)
+
+const unminedRecoveryRetryDelay = time.Minute
+
+func (b *BlockAssembler) unminedRecoveryInterval() time.Duration {
+	if interval := b.settings.BlockAssembly.UnminedRecoveryInterval; interval > 0 {
+		return interval
+	}
+	return settings.DefaultUnminedRecoveryInterval
```
freemans13, line 25: No value disables recovery. A non-positive interval falls back to the one-hour default, so `blockassembly_unminedRecoveryInterval = 0` reads as on and hourly rather than off. The settings doc says so plainly, so this is a choice rather than an oversight.

It matters more here than for an ordinary interval. This pass gates the production mining path, and a failed rebuild stops block assembly following the chain until a retry succeeds. An operator who hits trouble with it on a mining node has no configuration route out and has to redeploy a build.

Suggest either treating a negative value as disabled while keeping zero as the default, or adding a boolean beside the existing service-level `blockassembly_disabled` kill switch.

freemans13 (approved):
Approve. The blocker I raised on 2026-09-07 is fixed. The pruner catchup guard now has a loader call beside the other pruner keys, plus an override test that would catch the regression, and the key reads back as set when I load settings against this branch.

One thing worth knowing before weighing the two standing approvals. The branch was force-pushed six minutes after icellan approved, and the whole recovery feature landed in that push: 35 files, +2,966/-249 on top of the state that was approved, including a new gRPC method on the blockchain service. So neither approval covers the recovery pass, the mining-snapshot leases, or `ReadFSMState`. oskarszoon has already asked icellan to look again.

I read the new code and found no correctness defect. Three operability findings, all inline, none of which I would hold the merge for.

1. The recovery tick does its expensive work before its cheap gate, so catchup pays five seconds and one warning a minute for a pass that cannot succeed. One-line reorder.
2. A wedged rebuild has no operator escape hatch, and the reset they will reach for returns success while doing nothing.
3. The recovery interval has no value meaning off.

Four things I chased and cleared, so nobody repeats the work:

- The mining-snapshot lease, which keeps memory-mapped subtree files alive while a caller reads them, is released on every path out of `GetMiningCandidate`. The incomplete-subtree branch needs no lease at all, because `createIncompleteSubtreeCopy` builds a heap-backed tree and the retain is a no-op on it.
- Cache eviction racing a mining submission is handled. The eviction callback marks the lease released, `Retain` then fails, and the caller reports the job expired rather than reading unmapped memory.
- The parent-first selection walk terminates on cycles and never dereferences an entry it excluded.
- `LocalClient` refusing to answer `ReadFSMState` reaches no production path. Its only constructors are test helpers.

What I ran, on 7b695ed58. `go build ./...` and `go vet` are clean. Under `-race -tags testtxmetacache` the block assembly, subtree processor, mining, subtree validation, pruner and settings packages are green, as are the new authoritative-FSM-read tests in the blockchain package. Upstream CI is 37 passing with one skip, and Sonar reports 90.9% coverage on new code.

What I did not do. None of this has been soaked on a live node. Given that the recovery pass now gates the production mining path and can stop block assembly following the chain when a rebuild fails, I would want a mining node running it for a day before it reaches production, watching for the wedge described in finding 2.

settings/pruner_settings.go
freemans13: Confirmed on 7b695ed58. The loader call is there beside the other pruner keys, and `TestPrunerSkipDuringCatchupLoaderReadsOverride` is the right shape to catch a regression, since it asserts an override rather than the default. Agreed on leaving `SkipProcessExpiredPreservations` and `ForceIgnoreBlockPersisterHeight` to a separate cleanup. My blocker is cleared and I have approved.

services/subtreevalidation/Server.go
freemans13: Confirmed on 7b695ed58. The FSM read is hoisted above the legacy branch and both paths now take the same explicit option, so the rule this PR establishes applies to the gRPC entry point as well. The peer path failing open on a read error while legacy still errors is the right split. The synthetic-IDLE recovery you flagged as still open is now the recovery pass in this same PR; I have reviewed it and left three operability notes, none blocking.

services/blockassembly/unmined_recovery.go
oskarszoon: Fixed in [bdf2241d3](https://github.com/bsv-blockchain/teranode/commit/bdf2241d3d0274ff3c10d837be60a1afb8017256). Recovery reads authoritative FSM state before WaitForPendingBlocks, so IDLE/catchup defer immediately; RUNNING still waits before reading the tip. Real-store regressions use a block with subtrees ready and mined flags pending: IDLE/catchup return without waiting, while RUNNING still respects the wait. Older-server Unimplemented now logs an informational upgrade message; genuine authority/storage/rebuild failures remain warnings.

services/blockassembly/BlockAssembler.go
oskarszoon: Fixed in [bdf2241d3](https://github.com/bsv-blockchain/teranode/commit/bdf2241d3d0274ff3c10d837be60a1afb8017256) using the visible-refusal option. All three reset RPCs wait for and return their actual result; queued requests execute their own options instead of receiving synthetic success. Partial recovery still refuses every reset variant before any consistency scan or UTXO mutation. Added teranode_subtreeprocessor_recovery_pending_seconds, sampled on scrape: retries retain the original pending age, successful publication clears it. Regression coverage includes pending-recovery refusal, queued requests, cancellation/shutdown, successful resets, and metric lifecycle/concurrent scrapes. An accepted reset keeps the service lifetime if its caller times out; the caller then has an unknown outcome. Settings documentation now covers refused operations, fixing the underlying fault, and stop/restart recovery. The recommended mining-node soak is still outstanding; CI does not substitute for it.

services/blockassembly/unmined_recovery.go
oskarszoon: Fixed in [bdf2241d3](https://github.com/bsv-blockchain/teranode/commit/bdf2241d3d0274ff3c10d837be60a1afb8017256). A negative blockassembly_unminedRecoveryInterval (for example -1s) disables new automatic recovery passes and their timer; zero keeps the 1h default. The setting loads through the normal configuration path. Regression tests verify no index scan when disabled and that disabling new passes cannot discard an already-started repair or open its mining gate. Apply the setting with a service restart; startup reload remains active. Documented both the opt-out and the missed-template-entry tradeoff.

icellan (changes requested):
Re-reviewed at `bdf2241d3` against merge-base `99b8579b4` — 54 files, +4361/-416. I dismissed my earlier approval: that was given at `d7adb5ff7` when this was 17 files / +820, and the scope since then (unmined-recovery subsystem, mining snapshots, subtree storage lease, new `ReadFSMState` RPC with a proto change) is not what I signed off on.

The FSM gating and the lease/refcount machinery look sound, and the recovery subsystem is a real answer to the self-healing gap I raised. One blocker in it.

**The recovery pass can permanently lose queued transactions.** `services/blockassembly/subtreeprocessor/unmined_recovery.go:172`

`snapshotPublished` (`:89`) puts every queued-prefix hash into `accepted`, and all of `accepted` goes to `prepare` (`:109`). But `classifyRecoveryMetadata` (`unmined_recovery_selection.go:133-136`) returns early with `entry.tx` unset for `data.Creating` and for `ErrTxNotFound`, so those stay `recoveryExcluded` (`:69`, `:190`) and never reach `selected`. Then `discardThrough(boundary)` dequeues the entire prefix unconditionally — `queue.go:41-48` loops `dequeueBatch` to the boundary with no per-hash check.

Scenario: validator is mid multi-record 2PC write on a large paginated tx (`Creating=true`); block assembly has it queued; a recovery pass snapshots the prefix, reads it as `Creating`, excludes it, and discards the prefix; the validator finalises a moment later. That tx is now in neither the queue nor the rebuilt template, and stays out of mining until a restart or a reset — along with every descendant the DFS evicts by ancestry. This is the same transaction-loss class the PR exists to fix. Either restrict the discard to hashes actually included/mined/conflicting, or re-enqueue the transiently-excluded ones.

**Reset regression.** `services/blockassembly/BlockAssembler.go:1682` — `executeResetRequest` now returns early when `fixUnminedSinceInconsistencies` fails. Pre-PR (`99b8579b4:410-415`) that error was logged and `b.reset(...)` still ran. So a single transient UTXO-store read error turns `ResetBlockAssemblyFully` into a no-op that reports failure, on the RPC whose purpose is unwedging a bad template — and which is now also the only escape from a latched `recoveryPending`. Keep logging-and-continuing, or gate the abort on a non-transient error.

Rest, in rough order:

- `services/blockassembly/unmined_recovery.go:93` — `recoveryMiningBlocked.Swap(true)` is set before the read-only selection phase (`waitForSubtreeStorage`, `ReadFSMState`, `GetBestBlockHeader`, `GetBlockHeaderIDs`, `BatchDecorate` over the whole ancestor closure). Nothing mutates until `prepare` returns, so `GetMiningCandidate` fails for the entire scan window rather than the rebuild. On a large unmined set that is a multi-minute mining outage per pass. `recoveryPending` at `subtreeprocessor/unmined_recovery.go:149` already covers the mutation window.
- `services/blockassembly/unmined_recovery.go:89` — the pass runs on the BlockAssembler channel-listener goroutine with no overall deadline (only the FSM/tip reads get the 5s `readCtx`). While `scanUnminedRecoveryHashes` iterates the full unmined index, the listener services neither `blockchainSubscriptionCh` nor `resetCh`: block notifications can be dropped once the subscriber channel fills, and reset RPCs stall. `triggerReconcile` recovers it, but a long scan quietly degrades chain following on a mining node.
- `services/blockassembly/unmined_recovery.go:164` — `scanUnminedRecoveryHashes` checks `tx.Node == nil` before `tx.Skip`, so an iterator Skip sentinel aborts the whole pass with "missing transaction metadata". Both stores emit that sentinel with a nil Node (`aerospike/unmined_iterator.go:461`, `sql/unmined_iterator.go:147`) and every pre-existing consumer checks `Skip` first (`BlockAssembler.go:3008/3412/3534`). No current write path puts `unminedSince` on a coinbase so it is latent, but it is a one-line reorder. The test that covers this (`startup_recovery_test.go:207`) builds the sentinel with a non-nil Node, which is not what either store produces — it is the only `Skip: true` in the package's tests, so nothing can catch it.
- `services/blockchain/fsm_read.go:26` — `fsmMu.TryLock()` on a plain `sync.Mutex` (`Server.go:121`) makes two concurrent `ReadFSMState` calls fail one with `Unavailable`, and any in-flight `SendFSMEvent` fail the read outright. With the 1-minute recovery backoff, a node transitioning often defers recovery repeatedly. `RWMutex` + `TryRLock`, or a short bounded wait.
- `services/blockassembly/subtreeprocessor/subtree_storage_lease.go:92` — `waitForSubtreeStorage` has no timeout and runs on the dispatcher goroutine (`SubtreeProcessor.go:968`), so a hung blob store halts admission, `moveForwardBlock`, reorg and reset. The existing `getSubtrees` path has the same exposure, so this is a second entry point rather than a new class — still wants a deadline.
- `services/blockassembly/unmined_recovery_selection.go:125` — a parent whose UTXO record returns `ErrTxNotFound` stays `recoveryExcluded`, and `orderRecoveryTransactions` evicts its whole unmined descendant cone from the rebuilt template, re-evicting every pass. Default pruner settings keep the parent alive, so this needs preservation disabled or failed — but the pre-PR code only *skipped* such txs at startup, where this now actively evicts. Missing parent record should classify as settled.
- `services/blockassembly/subtreeprocessor/unmined_recovery.go:149` — a deterministically failing rebuild leaves `recoveryPending` latched, and `BlockAssembler.go:514/1682` refuse both reset RPCs by design, so the only exit is a process restart. The transient case self-heals on the 60s retry; the deterministic case has no operator escape.
- `services/blockassembly/unmined_recovery.go:66` — the 5s `readCtx` is shared with `WaitForPendingBlocks`, which is documented to retry indefinitely (1s/2s/4s backoff). On a node with pending blocks it reliably returns `DeadlineExceeded`, giving a `Warnf` every 60s with no operator action available. Treat pending blocks as a quiet deferral, or give it its own budget.
- `services/blockassembly/Server.go:2271` — `ResetBlockAssembly`/`ResetBlockAssemblyFully`/`ResetBlockAssemblyValidateInputs` return raw errors while every other handler in the file uses `errors.WrapGRPC(...)`, so clients cannot unwrap them into teranode error types. These also went fire-and-forget → synchronous, so a caller timeout now returns `ctx.Err()` while the reset keeps running; worth documenting on the RPC.
- `services/blockassembly/unmined_recovery.go:19` — `blockassembly_unminedRecoveryInterval = 0` does not disable recovery; 0 maps to the 1h default and only a negative value turns it off. That inverts the usual convention here.
- `services/blockassembly/subtreeprocessor/mock.go:82` — the `IsMethodCallable` gate means an `On("RecoveryPending").Return(true)` without `.Times(...)` fires once and then silently falls back to `false`. Nothing mocks it today, but `RecoveryPending()` is consulted from `reset`, `processNewBlockAnnouncement`, `GetMiningCandidate` and `nextUnminedRecoveryDelay`, so the first test that does will quietly exercise the unblocked path. A plain `atomic.Bool` on the mock avoids it.

oskarszoon:
The September 14 findings are addressed in `a1676fc96`, integrated with current main. The queue-loss regression and five mining-admission race cases failed before the fixes and pass afterward.

- Accepted transactions with Creating/missing metadata, or incomplete/unproven ancestry, now defer the whole pass before template or queue mutation. The real dispatcher regression verifies retained ownership and exactly-once recovery after metadata becomes available.
- Full reset again logs a consistency-scan failure and continues the template reset. Reset RPCs preserve typed errors and document that accepted work continues if the caller stops waiting.
- Mining stays available throughout read-only selection. The gate begins before replacement; candidate requests recheck it after snapshot acquisition and before returning, including requests that started before the pass. Rejected mmap candidates release their leases.
- `blockassembly_unminedRecoveryTimeout` supplies a configurable total deadline (default 5m), including the dispatcher storage wait. SQL query admission and row iteration use that context, including through the logging wrapper. Aerospike iterator setup retains its client info timeout; iteration observes cancellation. Custom context-free constructors still require their own bound. A deadline during replacement retains repair responsibility; the runbook covers increasing the budget if necessary.
- Real nil-Node Skip sentinels are handled before metadata validation. Pending mined flags quietly defer the periodic pass instead of entering the startup/reset retry loop.
- `ReadFSMState` now uses `TryRLock`; concurrent readers succeed while transitions, unready authority and uncertain persistence still fail closed.
- The mock recovery flag is atomic and remains stable across repeated reads.

Two safety boundaries remain deliberate. An absent parent record cannot establish canonical-chain membership: conflicting records are also pruned. Accepted descendants now retain their existing template/queue rather than being evicted. A failed destructive rebuild still cannot be bypassed by reset; bounded retries, the pending-duration metric and the documented repair/restart path remain. The interval convention is retained as documented: zero uses the default and a negative duration disables new passes; already-started repair is never abandoned.

Local integrated `make test` passed: 15,407 tests, 46 skipped, 559.134s, with race detection and repository-wide coverage. The pinned Go 1.27 linter reports zero new issues; affected-package vet and strict docs pass. Independent bounded integration checks found no lost main-branch safety checks. CI for this new head is running; older CI results are not used as evidence for it.

Could you re-review the current head? The mining-node soak remains outstanding; it is not represented by CI.

ordishs (changes requested):
### Overview
The PR adds a periodic unmined-transaction recovery pass to block assembly. The pass repairs templates that lost transactions while subtree validation fed with a synthetic IDLE state. It also adds an uncached `ReadFSMState` RPC and mmap snapshot leases for mining jobs and async subtree storage. It changes assembly feeding and pruner admission to require a positively observed RUNNING state. Reset RPCs now return the real result.

The safety design is careful. Selection is read-only, accepted work is never dropped on unknown eligibility, the queue prefix is consumed only after success, and mining is gated during publication. The lease lifecycle for mmap subtrees closes real use-after-unmap windows. My concerns are about cost and liveness. The pass runs by default every hour on every healthy mining node. When it succeeds, it replaces the whole template, even when no transaction is missing.

### Findings

**services/blockassembly/subtreeprocessor/unmined_recovery.go:110-167** — P1
**Issue**: A successful pass always clears and rebuilds the whole template. There is no check that the selection contains a transaction missing from `currentTxMap` or the queue. The rebuild calls `stp.addDirectly(ctx, tx.Node, tx.TxInpoints, true, true)` with `skipNotification=true`. The new subtrees are written to the blob store but peers never receive announcements for them. The order also changes (unmined-index scan order, then map iteration of accepted hashes), so the subtree roots differ from the roots that peers already validated.
**Failure**: A healthy RUNNING mining node with no missing transactions runs the default hourly pass. Its announced, peer-validated subtrees are discarded. Its next block references only unannounced subtrees. Every peer takes the `CheckBlockSubtrees` missing-subtree path for the whole block. The comment in `findMissingSubtrees` says that path is "expensive Aerospike reconstruction, which can fail under load and cascade into CATCHINGBLOCKS mode". This raises block validation latency and orphan risk for the miner once per hour. It also rewrites every subtree file each hour.
**Fix**: Compute `missing = selected − currentTxMap − queued prefix` on the dispatcher. If `missing` is empty, return without mutation. Where possible, append only the missing transactions parent-first to the live template, and keep the full rebuild as a fallback. If a full rebuild is still needed, announce the rebuilt subtrees (`skipNotification=false`) so peers can pre-validate them.

**services/blockassembly/unmined_recovery.go:56-61, subtreeprocessor/unmined_recovery.go:149-166** — P1
**Issue**: The whole pass (index scan, selection and rebuild) shares one `blockassembly_unminedRecoveryTimeout` budget (default 5m). A timeout after `recoveryPending.Store(true)` leaves the processor pending. Every retry repeats the same scan, selection and rebuild under the same budget. No path gives more time or falls back to the startup-style unbounded reload.
**Failure**: A healthy node has a template where scan + selection + rebuild takes more than 5m. Rebuild sends every completed subtree through `newSubtreeChan` to blob storage, so a large template can do this. The default hourly pass starts the rebuild and times out. After that, `GetMiningCandidate` returns `errMiningRecoveryPending`, `processNewBlockAnnouncement` returns early on every block, and reorg/reset are refused. The retry every minute never finishes. The node stops mining until an operator raises the timeout and restarts. Before this PR, the same node kept mining. Finding 1 makes this apply to healthy nodes, not only to nodes that lost feeds.
**Fix**: Do not apply the timeout after the destructive point, or give it a separate, much larger budget. Alternatively, after N failed repair attempts, escalate to the existing startup reload path instead of repeating the same bounded attempt. At minimum, do the scan and selection before the budget starts for the destructive phase. Also add a default-off rollout for the periodic timer until the mining-node soak is complete.

### Verdict
**REQUEST CHANGES** — The safety reasoning for selection, queue handoff and leases is strong. But the default-on hourly pass replaces the whole template on healthy nodes, which strips subtree pre-announcement from the next block. A rebuild that exceeds the default timeout can also stop mining permanently. Please make a successful pass a no-op when nothing is missing, and remove the timeout wedge, before this ships enabled by default.

icellan:
Reviewed at `25b709a7a`, reached via #1698 — I captured that PR's diff from the merge-base with `main` and these findings turned out to sit in this branch, not that one. Posting them here where they belong. One of them (the assembly gate) I already raised on #1698 by mistake.

## P0

**`services/blockassembly/subtreeprocessor/unmined_recovery.go:149` — `recoveryPending` is a one-way latch on failure**

It is set `true` immediately before the destructive phase (`closeChainedSubtrees`, the `currentSubtree` swap, `currentTxMap.Clear()`) and cleared only in `publishRecoveredMiningData` (`:182`) on full success. The replay loop returns early on `ctx.Err()` or any `addDirectly` error.

A node with a large template hits the default `UnminedRecoveryTimeout` of 5m during replay and returns with `recoveryPending == true`. From then on `moveForwardBlock`, `reorgBlocks` and `reset` are all rejected (`SubtreeProcessor.go:896`, `:913`, `:983`), `processNewBlockAnnouncement` returns early, `GetMiningCandidate` errors, and the dispatcher's `default:` branch sleeps 1 ms instead of dequeuing ingest. The 60 s retry re-runs under the same 5 m bound, so a template that cannot be replayed inside the budget wedges block assembly permanently until the process restarts.

The latch needs to clear on the failure paths too, with the destructive phase either made resumable or not entered until the replay is known to fit.

**`services/blockassembly/subtreeprocessor/SubtreeProcessor.go:978` — the recovery pass blocks the dispatcher goroutine**

The handler runs inline on the single dispatcher goroutine. The whole pass executes with the dispatcher blocked: the storage wait, the `accepted` set built by walking every node of every chained subtree, the `prepare` callback with its `BatchDecorate` over the entire unmined set, then the replay. Transaction ingest dequeue is halted for the full window.

`blockassembly_maxQueueItems` defaults to `0` — unbounded. This is exactly the unbounded-queue growth path that the setting's own documentation describes as having caused a 558 GB RSS OOM, and `settings.conf` now triggers it hourly by default.

## P1

**`services/blockassembly/BlockAssembler.go:429` — recovery also occupies the channel-listener goroutine**

`recoverUnminedTransactions` runs on the goroutine that drains `blockchainSubscriptionCh` and `resetCh`. For the duration of the pass, block assembly consumes no block notifications and no reset requests. On by default: `settings.conf` adds `blockassembly_unminedRecoveryInterval = 1h`.

**`services/blockassembly/unmined_recovery.go:104` — the pass is unconditional**

There is no cheap pre-check for whether anything is actually missing. Every hour it runs a full `GetUnminedTxIterator` scan, a full metadata `BatchDecorate` over all candidates plus their ancestor frontier, and a complete teardown and rebuild of the mining template, even when the template is already correct. A cheap comparison up front would skip the common case entirely.

**`services/subtreevalidation/subtreeHandler.go:60` — RUNNING-only assembly gate on a state that is synthesized as IDLE on any subscription blip**

(This is the finding I mis-filed on #1698.)

`Client.GetFSMCurrentState` (`blockchain/Client.go:1828`) returns the cache without contacting the server, and that cache is deliberately forced to IDLE on failure: no heartbeat ever received (`Client.go:1469`), heartbeat stale past `heartbeatTimeout` (`:1475`), reconnect state fetch failed (`:1563`).

The old test was `*state == CATCHINGBLOCKS` → disable, which a synthetic IDLE could not trip. `allowAssemblyForObservedFSM` (`assembly_admission.go:14`) is `state != RUNNING` → suppress, which it can. A transient blockchain-service gRPC blip on a genuinely RUNNING node stops feeding transactions into block assembly for the whole outage: the node keeps validating, looks healthy, and mines short blocks. Same predicate at `subtreevalidation/Server.go:826` and `check_block_subtrees.go:486`.

The `assembly_feeding_suppressed_total` metric and the one-per-minute warning fix the "nothing logs it" half, which was my main complaint last round. The predicate itself still rides an untrustworthy read. Either route these gates through `ReadFSMState`, or treat unknown/synthetic as RUNNING and reserve fail-closed for a confirmed IDLE.

**`services/blockchain/fsm_read.go:26` — `ReadFSMState` fails fast on lock contention, and its caller treats that as fatal**

It returns `codes.Unavailable` whenever `fsmMu.TryRLock()` fails (a writer holds or is pending) or `subscriptionManagerReady` is false. The consumer added in #1698 retries at 1 Hz with a budget of 5, so a ~5 s blockchain-service restart, or a slow `SetFSMState` persistence window overlapping several probes, aborts the in-flight catchup unit with a `ServiceError` rather than waiting it out.

Worth noting the write side is still the real hazard: `sendFSMEventLocked` holds `fsmMu` across `SetFSMState`, `time.Sleep(FSMStateChangeDelay)` (`Server.go:2995`) and the unbounded blocking send `b.notifications <- req` (`Server.go:2594` — the non-blocking form at `:788` is still the only guarded one). If `startSubscriptions` has exited and notifications accumulate, that send blocks forever with `fsmMu` held, and `ReadFSMState` then returns `Unavailable` permanently.

## P2

**`services/subtreevalidation/Server.go:826` — FSM read added to the non-legacy path, failing closed on error**

`checkSubtreeFromBlock` now issues the FSM read on every call, including the non-legacy peer/block path where previously no read happened at all. On the non-legacy error path it sets `currentState = nil`, which makes `allowAssemblyForObservedFSM` return false: a transient blockchain-service blip silently drops those transactions from the mining template rather than surfacing the failure. Recovery is left to the new hourly pass, which is itself gated on an authoritative RUNNING.

## P3

**`services/blockassembly/Server.go:594` — sequence gap makes subtree notifications silently lossy**

`subtreeStorageWorker`'s `ctx.Done()` branch now `continue`s after `work.request.Release()` without emitting a `subtreeStorageResult`, and the `resultChan <- result` send below can also be abandoned on `ctx.Done()`. Either leaves a permanent gap in `subtreeNotificationSender`'s `nextSeq` sequence (`Server.go:~690`), so every subsequent subtree's notification sits in `pending` and is never sent. Shutdown-only today, so nothing hangs in practice — but it turns the ordering guarantee into something that fails silently instead of loudly.

PR #1698, opened 2026-09-05
fix(fsm): preserve catchup work across durable pause

oskarszoon:
Draft status after human review (2026-09-07): not ready to merge. #1693 is merged; #1696 is the next review/merge target and supersedes closed #1695. #1697 has new consumer fixes. This draft has not yet integrated their latest heads.

Remaining corrections: distinguish actual FSM/admission state from startup readiness and synthetic cached IDLE; prevent authoritative reads being held behind unbounded notification delivery without weakening STOP serialization; bound and classify catchup-entry failures; revisit synthetic-IDLE peer replacement behavior. Integration must retain #1696's uncertain-write reconciliation and bounded promotion retries alongside this draft's automatic RUN/CATCHUP refusal from IDLE. Prerequisite propagation alone does not address these review findings.

STOP now pauses CATCHINGBLOCKS directly and persists IDLE before reporting success. Automatic Run/CatchUpBlocks calls cannot undo that pause; operators resume explicitly with `teranode-cli setfsmstate --fsmstate catchingblocks`.

Catchup workers retain queued targets, alternate sources and progress markers while paused. Each target, fork cleanup pair, validation block and subtree-prefetch block waits for an uncached admission snapshot serialized with STOP at the blockchain authority. Admission leaves the operating state unchanged. Peer evaluation suspends during IDLE/unavailable state without counting that time as lack of progress.

Supersedes #1672. Retains its useful STOP transition expectations and updates the FSM diagrams/operator instructions, replacing its cached-IDLE/drop-work approach with authoritative admission and preserved queue ownership. Refs #1670, #1346 and #1685. Draft for review of the pause contract and implementation; does not close the operational runbook ticket.

Dependencies and review scope:

- Includes earlier transition-hardening and persistence commits; update from combined #1696, which supersedes #1695.
- Includes an earlier version of #1697. Integrate its current loader, post-wait guard, feeding options, counters, and warning fixes before enabling STOP from CATCHINGBLOCKS.
- Boot-policy PR #1693 is merged. Rebase after prerequisite review; resolve overlaps through explicit conflict review.
- New native `ClientI.AdmitCatchupWork` uses the existing GetFSMCurrentState RPC; no protobuf/wire change. Custom Go implementations of ClientI must implement the new method. Automatic Run/CatchUpBlocks callers must use an explicit SendFSMEvent to resume IDLE deliberately.

Scope: admission starts one catchup unit. Previously admitted units and their asynchronous writes may finish after STOP. Normal direct ProcessBlock intake and background recovery remain outside this pause contract. IDLE is not a global drain barrier; rewind requires verified service shutdown. In-memory pending work is retained while the process lives; restart reconstructs synchronization from durable chain progress. An ambiguous store commit followed by an error requires retry of the failed event, as documented in #1696.

Validation on the existing draft head, before the outstanding human findings above:

- Regression tests cover durable STOP/failure/retry, automatic-origin rejection, explicit resume, stale client state, state-neutral admission, readiness recovery, admitted-work drain/next-unit pause, fork/prefetch boundaries, marker lifetime, queue capacity/shutdown, retry ownership and peer timeout suspension.
- Focused race suites, relevant package vet, independent implementation review and protocol/consumer safety review passed.
- Final blockchain race package passed (52.372s), blockvalidation broad/race suites passed (99.578s/9.514s), and P2P race with repository-wide coverage passed (21.114s). Updated blockpersister test implementation passed its race package (6.316s).
- Changed-code lint: zero issues. Relevant package vet and strict MkDocs build passed; regenerated FSM SVG visually checked.
- Final repository `make test` passed on head `288ebf210`: 13,728 tests, 46 skipped, 470.385s. This target runs the unit suite with race detection and repository-wide coverage.

Review cleanup includes the shared database-timeout default and #1697's accurate pruner metric label, relaxed test polling, and subtree-existence helper. The integrated changes passed the full suite above and the strict documentation build.

Peer-announced subtree validation also passes a known-RUNNING assembly-feeding option, closing an additional IDLE bypass while preserving ordinary validation and existing catchup skipping. Actual Kafka-entry and block-only-option isolation regressions cover the correction.

BlocksOnly peer announcements now skip before querying FSM state; a regression verifies zero state reads and no validation even when the FSM observation would fail.


icellan:
The core mechanism holds up. `context.WithoutCancel` around `looplab.Event` is the right fix — verified against looplab/fsm v1.0.2 that a ctx cancel mid-`Event` really does leave `f.transition` non-nil forever, and the added `InTransitionError`/`CanceledError` cases are correct. `before_event` runs after the transition-validity checks, so no invalid event gets persisted. Boot (`Init` → fresh node into CATCHINGBLOCKS) and operator resume (`setfsmstate --fsmstate catchingblocks`) both still work. `go build ./...` and `go vet` on the touched packages are clean.

My problem is with the premise. The description argues at length that an observed IDLE is untrustworthy, then builds four new fail-closed RUNNING-only gates on exactly that untrustworthy read — and makes the "authoritative" read itself return a synthetic IDLE.

## P1

**`services/subtreevalidation/subtreeHandler.go:70` — RUNNING-only assembly gate on a state that is synthesized as IDLE on any subscription blip**

`Client.GetFSMCurrentState` returns the cache without contacting the server, and that cache is deliberately forced to IDLE on failure: no heartbeat ever received (`Client.go:1465`), heartbeat stale past `heartbeatTimeout` (`:1471`), reconnect state fetch failed (`:1559`).

The old test was `*state == CATCHINGBLOCKS` → disable, which a synthetic IDLE could not trip. The new test is `*state == RUNNING` → enable, which it can. So a transient blockchain-service gRPC blip on a genuinely RUNNING node stops feeding transactions into block assembly for the whole outage: the node keeps validating, looks healthy, and mines short blocks. Nothing logs the flip. Same inversion at `services/subtreevalidation/Server.go:920` and `check_block_subtrees.go:448`.

Either route these gates through the authoritative RPC, or treat unknown/synthetic as RUNNING for the assembly decision and reserve fail-closed for a confirmed IDLE.

## P2

**`services/blockchain/Client.go:2059` — `AdmitCatchupWork` is not authoritative**

`Blockchain.GetFSMCurrentState` (`Server.go:2757`) returns IDLE whenever `!subscriptionManagerReady.Load()`, regardless of real state. So the doc comment ("state-neutral snapshot serialized with STOP") is wrong for the whole blockchain-service startup window, and admission cannot distinguish "operator STOPped" from "not ready yet". It also disagrees with `sendFSMEventLocked`, which gates on the unmasked `Current()`: with readiness unset and real state CATCHINGBLOCKS, `CatchUpBlocks` returns no-op success while `AdmitCatchupWork` refuses.

**`services/blockchain/Server.go:2739` — FSM reads now serialise on `fsmMu` behind two 5 s store ops, a sleep, and an unbounded channel send**

`sendFSMEventLocked` holds `fsmMu` across `guardRunBelowHighestCheckpoint` → `GetBestBlockHeader` (5 s), `before_event` → `SetFSMState` (5 s), `enter_state` → `b.notifications <- req` (`Server.go:2515`, a blocking send with no select/default — the heartbeat path at `:742` already uses the non-blocking form), and `time.Sleep(FSMStateChangeDelay - duration)` at `:2932`.

A RUN against a slow store therefore blocks every `GetFSMCurrentState` RPC for ~10 s. `AdmitCatchupWork` uses a 5 s RPC timeout, so every parked catchup unit's admission times out and re-queues, and `CheckFSM` (wired into 14 services) 503s on a cache miss. Worse: if `startSubscriptions` has already exited and 100 notifications accumulate, the `enter_state` send blocks forever with `fsmMu` held — permanently wedged FSM read path. Wants a separate RWMutex for the read, or a snapshot released before the metrics work, plus a bounded send in `enter_state`.

**`services/p2p/sync_coordinator.go:792` — suspension removes stalled-sync-peer rotation entirely**

`activeSyncStateLocked` (same cached read as the P1) now gates `evaluateSyncPeer` (`:1153`) and `triggerSyncLocked` (`:642`), not just `handleRunningState`. `evaluateSyncPeer` previously ran unconditionally, so a stalled, low-reputation or vanished sync peer got cleared and replaced. Under a synthetic IDLE that recovery path is gone — `HandleCatchupFailure`'s `clearSyncPeerIfCurrent` + rotation becomes a no-op and the node stops following until the client cache recovers via reconnect. Silent, too: `warnfUnlessStopping` only fires on an RPC error, the nil-client and non-active-state branches log nothing.

**`services/blockvalidation/catchup.go:1408` — `setFSMCatchingBlocks` retries every error forever, silently**

It uses the bare `waitForCatchupAdmission(ctx, u.blockchainClient.CatchUpBlocks, …)` helper, which loops at 1 Hz on any error until ctx cancellation — including a persistent `ErrStorageError` from a failed `SetFSMState` in `before_event`. Unlike `(*Server).waitForCatchupAdmission` it logs nothing. A store that cannot persist the CATCHINGBLOCKS transition wedges the catchup session indefinitely while holding the catchup lock, so every other announced block gets `ErrCatchupInProgress` and silence is the only symptom. Wants a periodic warning, and a "paused → retry" vs "storage broken → bail" split.

ordishs (approved):
### P1 — services/p2p/sync_coordinator.go:642, 792-806, 1153

The two new sync gates read the *cached* client FSM state, which the rest of this PR deliberately bypasses.

- `triggerSyncLocked` (642) and `evaluateSyncPeer` (1153) now return early on `activeSyncStateLocked() == nil`. That reads `blockchainClient.GetFSMCurrentState`, which on `*blockchain.Client` short-circuits on a cached pointer and never reaches the wire (`Client.go:1826-1829`).
- The cache holds a synthetic IDLE that no real transition produced: `Client.go:1467`, `1473` (broken notification stream) and `Client.go:1561` (post-reconnect `fetchAndRestoreFSMState` failure).
- `AdmitCatchupWork` bypasses this cache on purpose — "Bypass fmsState: either IDLE or CATCHINGBLOCKS there can be stale" (`Client.go:2057`) — as does `restoreFSMState` (`catchup.go:1420`). The new p2p gate is the one place that trusts it, and it gates every sync decision.
- Failure: blockchain service restarts; p2p stores synthetic IDLE; on reconnect `fetchAndRestoreFSMState` reads IDLE because the server has not yet set `subscriptionManagerReady` (`Server.go:2757`, set at `Server.go:770`). The fetch reports success, so nothing retries. The real FSM is RUNNING and never transitions again, so no `FSMState` notification refreshes the cache (`Client.go:210-216` is the only other writer). `activeSyncStateLocked` then always returns nil, and when the sync peer disconnects, `HandlePeerDisconnected` → `TriggerSync` → `triggerSyncLocked` no-ops with no path left to select a replacement. The node stops following the chain until p2p restarts, explained only by a Debug line.
- This is a regression introduced here: before the PR both call sites were FSM-independent, so eviction and re-selection still worked under a stale cached IDLE.

Suggested: use an uncached read for this gate as `AdmitCatchupWork` does, or fail open. Cheaper containment — suspend only on an IDLE observed from the wire, and leave `evaluateSyncPeer`'s eviction path ungated so a dead sync peer is always released.

### P1 — services/p2p/sync_coordinator.go:800-816

A transient FSM read failure resets the sync-peer stall clock, so a stalled peer is never evicted.

- Line 804 folds observed IDLE, a nil state, and a read error into one `syncSuspended = true` branch. On the next successful read, all three are treated as an operator resume and stamp `lastSyncProgressTime = time.Now()`.
- `activeSyncStateLocked` runs on the 2s `monitorFSM` tick and the evaluation tick; the default stall limit is 5 minutes. One synthetic-IDLE blip per stall window holds the clock at zero indefinitely.
- Failure: a peer that serves headers and never delivers block bodies keeps the sync slot forever, defeating the protection documented at `sync_coordinator.go:583-600`. Not attacker-controlled — the peer cannot cause the blips.

Suggested: reset only when the previous suspension came from a positively observed IDLE. Better, freeze instead of reset — record the suspension start and on resume advance `lastSyncProgressTime` by the suspended duration. A test that alternates one failed read with a RUNNING read and asserts the stalled peer is still evicted would lock this in.

### P1 — settings/settings.go:599-621 (guard at services/pruner/worker.go:162)

`pruner_skipDuringCatchup` is never read in `NewSettings`, so the pruner guard this PR hardens cannot be enabled.

- Struct tags are documentation only in this repo. The `Pruner: PrunerSettings{...}` literal wires 20 other pruner keys via `get*` helpers and omits `SkipDuringCatchup`, so the field stays `false` and `worker.go:162` never executes.
- The gap predates the PR, but the PR rewrites that tag line (`settings/pruner_settings.go:9`), rewrites the guard, and rewrites both docs pages — and this guard is the pruner-side fix for the IDLE bypass.
- Failure: an operator sets `pruner_skipDuringCatchup = true` per `settings.conf:615` and `docs/references/settings/services/pruner_settings.md:138`. The value stays `false`, `prunerSkipped{reason="fsm_not_running"}` never increments, and the pruner runs Phase 1/1b/2 plus blob-deletion scheduling while the FSM is CATCHINGBLOCKS or operator-paused IDLE. The default is `false`, so there is no regression on the default path — the defect is a documented knob that silently does nothing.
- `services/pruner/worker_fsm_test.go:60` assigns the struct field directly, so it cannot detect this.

Suggested: add `SkipDuringCatchup: getBool("pruner_skipDuringCatchup", false, alternativeContext...),` to the `Pruner` literal, plus a test that builds settings through `NewSettings`. A reflection test asserting every `key:` tag is reachable through `NewSettings` would catch this whole silent class.

oskarszoon:
The three September 9 findings are addressed in `c898a3668`, integrated with current main and #1697.

- Both P2P gates now use `ReadFSMState`, so synthetic cached IDLE cannot permanently prevent peer replacement. Disconnected/failed peer removal remains possible during authority loss.
- Only confirmed IDLE suspends the stall clock. Resume adds the paused duration instead of resetting accumulated age; transient failures never reset it. Regressions cover alternating failed/successful reads and stale cached IDLE.
- `pruner_skipDuringCatchup` is loaded through `NewSettings` by the integrated consumer prerequisite, with constructor-level coverage. This retains current main's pruner settings and expiry behavior.

The related admission corrections are also included: ready/persistence-confirmed reads, a distinct confirmed-pause sentinel, bounded transient retry handling, immediate permanent-error return, and retained main-branch promotion retries with separate RPC deadlines. Real SQLite/gRPC tests cover STOP persistence, blocked notifications and ambiguous committed STOP. Integration also preserves main's prefetch memory/provenance/cleanup changes and its three-alternative-peer cap.

The final integrated local run covered 15,476 tests (46 skipped, 534.490s). All FSM/catchup/consumer packages passed under race detection and repository-wide coverage. Its only failure was an unrelated Docker host-port collision starting Kafka's TestFlushBytesRegression_MessageTooLarge; the unchanged Kafka package passed separately with the same instrumentation (24.490s). An earlier full run exposed an existing catchup-lock test scheduling assumption; it now uses barriers, passed100 race/coverage repetitions (38.553s), and passed the integrated rerun. Pinned Go1.27 lint reports zero new issues, affected-package vet passes, and strict MkDocs passes. Final-head CI is pending.

Could you re-review this head? It remains draft while #1697 is awaiting merge. The documented contract permits already admitted work to finish; IDLE is not a rewind barrier, and production soak is still outstanding.

ordishs (approved):
Re-review at `4ea6254e2`, covering the part of this PR on top of #1697.

The earlier P1 items are resolved. The P2P gate now reads `ReadFSMState` and does not use the cached state. Unavailable reads no longer reset the stall clock, and removal of a vanished peer runs before the gate. Build, vet and `go test -race` pass for `services/blockchain`, `services/blockvalidation/...` and `services/p2p`.

There are no blocking findings. Smaller follow-ups:
- The fork mined-clear/notify admission point at `catchup.go:304-307` has no test. When I deleted the guard, the catchup/fork tests still passed.
- `processBlockNotify` no longer has a decision reader. The comments at `Server.go:187` and `Server.go:2307` still describe it as the duplicate guard.
- The operator FSM docs say that later units wait after STOP. Legacy netsync and direct block intake do not wait. Please state this in the docs.

icellan:
Re-reviewed at head `4ea6254e2` (my earlier comment was against `288ebf210`).

**Correction first:** the P1 in my previous comment — the RUNNING-only assembly gate in `services/subtreevalidation/` — is #1697's code, not this PR's. I captured the diff from the merge-base with `main` while #1697 was still open, so its commits were inside the range. The defect is real and still present at this head, but it belongs on #1697 and I've filed it there. Same goes for `services/blockchain/fsm_read.go`.

**Status of the rest of that comment:**

- `AdmitCatchupWork` not authoritative — **fixed**. `Client.go:2062` calls `ReadFSMState`, which returns `Unavailable` rather than a synthetic IDLE, so "blockchain not ready" and "operator STOPped" are now distinguishable.
- `GetFSMCurrentState` serialising on `fsmMu` — **fixed for that RPC** (`Server.go:2815` takes no lock), but the underlying hazard survives: `sendFSMEventLocked` still holds `fsmMu` across `SetFSMState`, `time.Sleep(FSMStateChangeDelay)` (`Server.go:2995`) and the unbounded blocking `b.notifications <- req` (`Server.go:2594`; the non-blocking form at `:788` is still the only guarded one). It now manifests as `ReadFSMState` returning `Unavailable` forever, which feeds the first P1 below.
- Sync-peer rotation — **partially fixed**, see P2 below.
- `setFSMCatchingBlocks` looping forever — **fixed**, and the fix creates the second P1 below.
- `catchup_queue.go` alternatives wipe — **unchanged**, restated as a P2 below.

## P1

**`services/blockvalidation/catchup_admission.go:87` — `Unimplemented` classified as permanent, on an RPC this PR series introduces**

`transientCatchupAdmissionError` treats `codes.Unimplemented` as non-transient, but `ReadFSMState` is new in `blockchain_api.proto`. In a rolling upgrade where blockvalidation rolls before blockchain, every `AdmitCatchupWork` gets `Unimplemented`, `waitForCatchupAdmission` returns a `ServiceError` immediately, and every catchup unit aborts — `Server.go:2138`, `catchup.go:1704`, `get_blocks.go`. `catchupTransitionError` has the same exposure since it returns `readErr` verbatim.

The skew is anticipated elsewhere in the series: `logUnminedRecoveryError` explicitly special-cases `Unimplemented`. This path should too — treat it as transient, or fall back to the old `GetFSMCurrentState` when the new RPC is absent.

**`services/blockvalidation/catchup_admission.go:48` — a 5 s blockchain blip now aborts a catchup unit**

Bounding the retry loop at `catchupAdmissionMaxFailures = 5` is the right fix for the infinite-retry problem, but `ReadFSMState` returns `Unavailable` whenever the write lock is held or pending. Combined with the retry at 1 Hz, a blockchain-service restart, or a slow `SetFSMState` window overlapping five probes, aborts the in-flight catchup unit with `ServiceError` instead of waiting it out. Either give `Unavailable` a longer budget than the other transient classes, or back off rather than counting.

## P2

**`services/blockvalidation/catchup_queue.go:58` — queue-full rollback wipes alternatives it did not add**

The channel-full branch of `enqueueCatchup` calls `u.catchupAlternatives.Delete(hash)` unconditionally. Alternatives legitimately exist with no queued entry: the `ErrBlockPolicyDeclined` branch (`Server.go:2323`) preserves them deliberately as the only recovery route, and `releaseCatchupOwnership` restores their TTL rather than deleting them. An announcement arriving while `catchupCh` is full now destroys that route. The code this replaced only deleted `processBlockNotify` on a full channel. Remove only what this call added.

**`services/blockvalidation/Server.go:2323` and `:2351` — re-announcements silently dropped inside the ownership window**

Both branches call `u.processBlockNotify.Delete(...)` directly instead of `finishCatchupTarget(c)`, so `catchupQueued[hash]` retains ownership until the deferred `releaseCatchupOwnership` runs — which is after `reportCatchupFailureForError`, `ReportPeerFailure` and `reportCatchupMalicious`, all RPCs. A re-announcement landing in that window is absorbed as an "alternative" rather than enqueued, then discarded when ownership is released. The block is dropped until some peer announces it again.

(The `processBlockNotify` deletes themselves are now cosmetic — that cache has no functional reader left anywhere in the repo since this PR removed `addBlockToPriorityQueue`'s dedup `Get`. Worth dropping the cache or restoring a reader; the `ttlcache.NoTTL` pinning and the `Server.go:454` comment currently guard nothing.)

**`services/p2p/sync_coordinator.go:884` — IDLE never latched, so a resume reads as a completed catchup**

`checkFSMState` returns before `handleFSMTransition` whenever `activeSyncStateLocked()` is nil, so IDLE never reaches `lastObservedFSMState` (`:942`). Node in CATCHINGBLOCKS → operator STOP → IDLE → resume to RUNNING: the next tick computes `completionEdge` against the stale CATCHINGBLOCKS latch (`:939`) and treats the resume as a finished catchup, arming `pendingCompletionPeer` and suppressing the fallback slot release for a catchup that never completed. The `pendingCompletionPeer = ""` reset for the excursion never fires either.

**`services/p2p/sync_coordinator.go:1404` — stalled-peer rotation still behind the gate, and still mostly silent**

Moving vanished-peer removal (`:1377`) and low-reputation rotation (`:1385`) ahead of the gate at `:1395` fixes most of what I raised last time. The no-progress deadline path is still gated, so under a confirmed IDLE or an unavailable authority a stalled but existing, good-reputation peer is never rotated. On logging: `:840` now warns on RPC error, but the nil-client branch (`:833`), the IDLE branch and the other-state branch still emit nothing, and `triggerSyncLocked` (`:681`) returns nil as though it had acted.

PR #1704, opened 2026-09-08
p2p: rate-limit and dedup the rejected-tx re-broadcast, send error codes not text

gokutheengineer:
**Closes bitcoin-sv/teranode#4759.**

### Problem
`rejectedTxHandler` re-broadcast every internally rejected transaction (one Kafka message per validator rejection) on the `rejected_tx` gossip topic with no rate limit and no txid dedup. Each rejection costs an attacker one invalid transaction on the propagation surface, while each re-broadcast is fanned out by gossipsub to every mesh peer at this node's expense: a several-fold egress amplification plus a network-wide flood of a topic recipients treat as informational (`handleRejectedTxTopic` takes no action on it). The `Reason` field also carried the validator's raw `err.Error()`, unbounded and shaped by the rejected input, truncated at 1024 bytes on the way out.

The size cap and FSM-gate parts of the original report were already covered on `main` (`publishToNetwork` enforces `maxRejectedTxMessageSize` outbound; `canSendToNetwork` is a cached atomic load) and are not touched here.

### Fix
- `services/p2p/rejected_tx_gate.go`: `rejectedTxEgressGate`, applied in `rejectedTxHandler` before any publish work. A per-txid dedup on the existing `seenHashCache` primitive (this node as the single announcer) followed by a refundable token bucket. Dedup runs first so repeats never spend a token. `allow` returns a `rejectedTxGrant`; a publish that did not reach the network (`publishToNetwork` returned `sent == false`, validation or marshal failure) calls `grant.publishFailed()`, which returns **both** charges: the dedup slot so the txid can retry, and the rate token since nothing was sent. The bucket is the package's own rather than `golang.org/x/time/rate` because that limiter cannot cancel a reservation once its act time has passed, which is always the case by the time a publish is known to have failed. A refund never lifts the bucket above burst. No unbounded mode: a zero-value gate applies the package defaults on first use, matching `cappedPeerMap` and `seenHashCache`.
- The dedup cache joins the `cleanupPeerMaps` sweep and its size line, so it releases at TTL like the block and subtree seen-hash caches instead of staying pinned at `SeenHashMaxSize` after a distinct-txid flood.
- New settings `p2p_rejected_tx_publish_rate` (default 10/s) and `p2p_rejected_tx_publish_burst` (default 100), wired through `applyPeerMapLimits`; the dedup shares the `p2p_seen_hash_*` bounds. Non-positive values select the defaults. Reference docs updated.
- New counter `teranode_p2p_rejected_tx_publish_suppressed_total{reason="rate_limited"|"duplicate"}`; suppression is counted and debug-logged, not warn-logged per message, for the same reason as `gossipPublishDropped`.
- `errors.Walk(err, visit)`: bounded walk over the teranode links of an error chain, at most `maxChainWalkDepth` (64) links visited regardless of what the visitor keeps, so same-coded chains, foreign wrappers and a cycle that got past `SetWrappedErr` all terminate. `errors.CodeChain(err, max)` is built on it: codes outermost first, deduplicated, `ERR_UNKNOWN` skipped, capped at `max`.
- `util/rejectedtx`: the reason format, shared by its producer and the egress chokepoint. Grammar `CODE("/"CODE)*(": "DETAIL)?`: the `CodeChain` joined with `/` (at most 4 codes; enum names print without `ERR_`), followed by `: <detail>` when a link's message is one of the fixed literals in the closed allowlist. `Reason(err)` builds it, `Valid`/`Normalize` check it. The two GoBDK outer literals are defined here and the script verifier builds its errors from them, so verifier and allowlist cannot drift.
- `services/validator`: the rejected-tx Kafka `Reason` is `rejectedtx.Reason(err)`. The full error is still returned to the caller and logged locally as before.
- `services/p2p`: `rejectedTxHandler` runs the reason it received through `rejectedtx.Normalize` before publishing, so the guarantee holds at the chokepoint and not only at the producer: during a rolling upgrade a pre-upgrade validator still publishes `err.Error()` on the in-cluster topic, and that now leaves the node as `TX_INVALID`, not as 1KB of truncated script text.
- `seenHashCache.ttlLocked` clamps the TTL to the 15s publish window. `Check`'s expiry branch runs before the window is consulted, so a `p2p_seen_hash_ttl` below 15s re-published every repeat and silently switched the dedup off for all three consumers, the rejected-tx egress included.

### Decision: what the reason carries
Codes alone were near-binary on the wire: every message-only `NewTxInvalidError` in `TxValidator.go` is one link, and the BDK mapper's outer `TX_INVALID` over inner `TX_INVALID` collapses to one code, so the receiving node's log read the same for a low-fee tx, a negative output and a bad signature. The discriminating information at those sites is the `bad-txns-*` literal, which is Bitcoin's standard reject vocabulary and carries no attacker bytes. So the reason keeps one such literal, chosen by exact match against the closed allowlist in `util/rejectedtx`: the `bad-txns-*` strings, `coinbase transactions are not supported`, `transaction fee is too low`, `bad-txns-inputs-too-large`, and the two fixed GoBDK outer literals. Deepest match wins, in the spirit of `DeepestPublicCause`, but within `Walk`'s 64-link bound: a literal buried deeper is knowingly lost and the reason falls back to codes alone. That bound is a deliberate truncation, unlike `maxIsChainDepth`, which is cycle insurance; the comment on it now says so. Formatted messages (input indexes, heights) and the BDK cause text are not on the list and never travel; a message that merely contains an allowlisted string does not match. Examples: `TX_INVALID: bad-txns-in-belowout`, `TX_INVALID/TX_POLICY: transaction fee is too low`, `TX_INVALID` (formatted or unlisted message).

This allowlist is stricter than `errors.publicCauseCodes`, which lets the same codes surface their message verbatim at the public error boundary. There the text goes back to the submitter who supplied the input it describes; here it is fanned out to every peer at this node's expense. The package doc records that so the literal list is not widened casually.

### Dedup period
The seen-hash publish window is a 15s package constant. With one announcer the observed period is two windows: a repeat inside the window is refused on the spent budget, the first repeat after rollover is refused too (the cache withholds the rollover retry grant from last window's grantee, which is us), and the next one is granted. A txid the validator keeps rejecting is therefore re-broadcast about every 30s until its entry ages out at `p2p_seen_hash_ttl` (now never below the window). Neither new setting changes this; it is documented on the gate, in the `p2p_rejected_tx_publish_rate` and `p2p_seen_hash_ttl` longdescs, the metric help text, and pinned by `TestRejectedTxEgressGate_RepeatPeriodIsTwoPublishWindows`.

### Clock
The gate owns one clock (`time.Now` unless a test injects one) read by both `allow` and `publishFailed`, so no bucket operation reads the wall clock behind a synthetic timeline. The handler-level flood test runs on a frozen clock and asserts the exact burst instead of a wall-clock upper bound.

### Settings text
The `p2p_rejected_tx_topic` and `KAFKA_REJECTEDTX` longdescs described a reason vocabulary (`double_spend`, `invalid_signature`, `Missing parent`, …) and a "filter future announcements" behaviour that never existed; both now describe the actual flow, the informational nature of the topic, the egress cap and the reason format. Note the longdescs surface in the dashboard settings portal via `ExportMetadata`, not in `docs/references/settings/`, whose tables carry the short `desc` only.

### Residual
- The dedup is secondary: a distinct-txid flood is bounded by the bucket alone, at burst plus rate per second. Peers only read this topic for display, so the defaults are deliberately low; raise them if a real mesh shows honest rejection bursts hitting `reason="rate_limited"`. Not exercised against real volume.
- Inbound rate limiting of the topic (a hostile peer flooding us) remains the open design item in internal #2870; this PR is the outbound path only.
- The validator side has no dedup of repeated rejections of the same txid, so p2p still receives one Kafka message per repeat and drops it there. Fine as is: the Kafka hop is in-cluster and the p2p dedup is the cheaper place.
- The reason tests now live in `util/rejectedtx`, which is cgo-free and runs locally. `services/validator` itself still cannot link locally (GoBDK cgo) and is verified by `go vet` here; CI runs it.
- Wiring the `clear()` on `Stop` is not pinned end to end: `Stop` needs a fully built server. The gate-level `TestRejectedTxEgressGate_ClearKeepsBucket` pins what `clear` does, including the deliberate "bucket is not reset".

### Tests
- `services/p2p`: `TestRejectedTxHandler_DistinctTxIDFloodIsRateLimited` (500 distinct rejections through the real Kafka handler on a frozen clock with rate 1 / burst 5 publish exactly 5), `TestRejectedTxHandler_RepeatedTxIDPublishedOnce`, `TestRejectedTxHandler_PublishFailureAllowsRetry` (burst 1: the retry can only succeed if the failed attempt's token came back), `TestRejectedTxHandler_GateDropReturnsDedupGrant`, `TestRejectedTxHandler_NormalizesReasonAtChokepoint` (pre-upgrade error text and unlisted details leave as `TX_INVALID`; well-formed reasons pass intact), `TestRejectedTxHandler_InternalRejectionReplacesFreeTextReason`, `TestRejectedTxHandler_ExternalRejectionsDoNotTouchTheGate` (50 external rejections leave the single burst token for our own), `TestApplyPeerMapLimits_ConfiguresRejectedTxEgress` (rate, burst, seen-hash size and TTL all in force through the settings path; delete `Server.go`'s `setLimits` call and it fails), `TestCleanupPeerMaps_SweepsRejectedTxDedup`, `TestRejectedTxEgressGate_ZeroValueUsesDefaults`, `TestRejectedTxEgressGate_Reasons`, `TestRejectedTxEgressGate_RefusalDoesNotDebitBucket`, `TestRejectedTxEgressGate_PublishFailedReturnsToken`, `TestRejectedTxEgressGate_PublishFailedUsesGateClock`, `TestRejectedTxEgressGate_RepeatPeriodIsTwoPublishWindows`, `TestRejectedTxEgressGate_SubWindowTTLStillDeduplicates`, `TestRejectedTxEgressGate_ClearKeepsBucket`, `TestRejectedTxEgressGate_NonPositiveLimitsSelectDefaults`, `TestTokenBucket_RefundCapsAtBurstAndTimeIsMonotonic`, `TestRejectedTxMessage_ReasonShapesPassIngressBounds` (every allowlisted detail and every reason shape through the real `sanitizePeerDisplayString`, unchanged), `TestSeenHashCache_TTLIsClampedToPublishWindow`.
- `errors`: `TestCodeChain` (ordering, dedup, cap, foreign outer link, flattened foreign wrapped error, explicit unknown link), `TestCodeChain_SameCodedChainIsBoundedByDepth` (50k same-coded links visit exactly 64), `TestCodeChain_CycleTerminates`, `TestWalk_SkipsForeignLinksAndStopsOnRequest`.
- `settings`: loader cases for both new keys.
- `util/rejectedtx`: `TestReason` (join, detail selection, deepest match, formatted and substring non-matches, fallback; every produced reason passes `Valid`), `TestDetails_AreFixedLiterals`, `TestValid` (accepts exactly the grammar; rejects pre-upgrade error text, unknown or duplicate codes, over-long lists, unlisted details, stray separators).

```
SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./errors/ ./settings/ ./util/rejectedtx/ ./services/p2p/   # ok
go vet ./services/validator/                                                                                                 # ok (test binary cannot link locally: GoBDK cgo)
golangci-lint run ./errors/... ./services/p2p/... ./services/validator/... ./settings/... ./util/rejectedtx/...              # only the pre-existing prealloc hit in p2p/mock.go
```


icellan:
Reviewed the diff. Premise holds up: `handleRejectedTxTopic` is informational, `rejectedTxConfig` has one consumer, and nothing depends on the free-text `Reason`, so the code-list swap breaks nothing. Traced the gate's dedup/retry interaction with `seenHashCache` by hand across first-publish, repeat, rate-limited-first-attempt and failed-publish — every path lands where the comments claim, and `publishFailed` is never called twice for one grant. New tests pass locally. No blockers, one thing worth fixing:

**`errors/error_utils.go:344`** — `CodeChain`'s `max` bounds *collected codes*, not *visited links*. A duplicate or `ERR_UNKNOWN` link hits `continue` without growing `codes`, so the `len(codes) >= max` break is unreachable on such chains. That leaves it without cycle insurance, unlike the rest of the package: `(*Error).Is` and the walk at `errors.go:852` both carry `maxIsChainDepth = 1<<20` explicitly against a cycle slipping past `SetWrappedErr`'s guards, and `Error()` truncates at 20. A cyclic chain makes `rejectedTxReason` spin forever, hanging the validator goroutine on the per-rejection path. It also walks the whole chain for same-coded chains — the 10k+-link shape this package documents from the block-820116 mass-spend stall is exactly the error that reaches `rejectedTxReason` (measured 50k links at 320us/call). A `depth` counter alongside `codes` fixes both.

Two low-severity notes:

**`services/p2p/Server.go:772`** — `rejectedTxEgress.seen` is the third `seenHashCache` on the Server and the only one missing from the `DeleteExpired` sweep at `server_helpers.go:1018-1019` (and from the stats line at `:1108`). After a flood of distinct txids it stays pinned at `SeenHashMaxSize` entries, each with its `announcers`/`grantees`/`prevGrantees` maps, for the process lifetime — retention rather than a leak, but the sibling caches release at TTL.

**`services/p2p/Server.go:1248`** — the `!sent` path returns the dedup grant but not the rate-limit token, so the two limits recover asymmetrically. A drop on the FSM branch (`publish_gate.go:216`) or the fail-open branch after a `Publish` error still burns a token, so a burst during a brief mesh outage empties the bucket and the first rejections that could actually be published get throttled at 10/s. `canSendToNetwork` pre-filters most FSM cases, which is why this is low. Either return the token alongside `publishFailed`, or note in a comment that it's deliberately kept.

freemans13 (commented):
Reviewed the outbound gate and the reason change against the paths that feed them. No correctness defect found in the gate itself. I traced the dedup and refund interaction with `seenHashCache` across first publish, repeat, rate-limited first attempt and failed publish, and every path lands where the comments say it does. `applyPeerMapLimits` is reached from `NewServer` (`services/p2p/Server.go:644`), so both new settings are genuinely wired rather than just tagged, which is worth stating because a tagged and documented key in this repo is not always a read one.

One thing I would want settled before merge, and it is a design question rather than a bug: the reason that now travels the network is the constant `TX_INVALID` for very nearly every rejection this node produces.

- All 13 `NewTxInvalidError` calls in `services/validator/TxValidator.go` pass a message and no wrapped error, so the chain is one link and the reason is `TX_INVALID`. That set is the whole `bad-txns-*` vocabulary: `bad-txns-in-belowout`, `bad-txns-vout-negative`, `bad-txns-inputvalues-outofrange` and the rest. Those are Bitcoin's standard reject strings, fixed literals carrying no attacker-controlled bytes.
- The main script path collapses as well. `mapBDKValidationError` returns `NewTxInvalidError(errMsgInvalidTx, bdkCause(errors.ERR_TX_INVALID, errVerify))` at `services/validator/ScriptVerifierGoBDK.go:366` and `:387`, an outer `TX_INVALID` wrapping an inner `TX_INVALID`, and your own `repeated codes deduplicated` case in `TestCodeChain` proves that pair collapses to one code.
- What survives is the policy branch, `TX_INVALID/TX_POLICY`. The field is therefore close to binary on the wire.

The only consumer of `Reason` anywhere in the tree is a `Debugf` on the receiving node (`services/p2p/server_helpers.go:596`). That is exactly the operator who wants to know which rejection it was, and the line now reads the same for a low-fee transaction, a negative output and a bad signature. The unbounded, input-shaped part you are right to strip is the BDK cause text. The 13 fixed literals are not that, so an enumerated reason for the message-only sites would keep the field useful and still exclude the text. If the answer is that the field is being deliberately retired to a category marker, that is a fine answer, it just belongs in the PR body as a decision.

Either way, two shipped setting descriptions now state something false about the field this PR changed, and neither is in the diff:

- `settings/p2p_settings.go:24`, the `RejectedTxTopic` longdesc: "Message contains TX hash, rejection reason (double_spend, invalid_signature, etc.), and Peer ID."
- `settings/kafka_settings.go:16`, the `KAFKA_REJECTEDTX` longdesc: lists "Invalid signature / Missing parent / Double-spend / Policy violation / Insufficient fee" as the reasons carried.

Both are the longdesc source that generates `docs/references/settings/`, which this PR regenerated for the two new keys.

I also agree with the depth bound @icellan asked for on `CodeChain`, and read it the same way: `max` counts collected codes, and a duplicate or `ERR_UNKNOWN` link takes `continue` without growing `codes`, so the break is unreachable on a chain of same-coded links. The BDK path above produces exactly that shape.

Verified locally on f5a7dc1:

```
go test -count=1 -run TestCodeChain ./errors/                                                     ok
SETTINGS_CONTEXT=test go test -count=1 -tags testtxmetacache -run TestRejectedTx ./services/p2p/  ok
SETTINGS_CONTEXT=test go test -count=5 -race -tags testtxmetacache \
  -run TestRejectedTxHandler_DistinctTxIDFloodIsRateLimited ./services/p2p/                       ok
```

On that last one I checked the wall-clock assumption in the flood test, since `LessOrEqual(*published, 6)` only holds if the 500-iteration loop stays under one second at rate 1/s. It runs in 0.01 to 0.02s under `-race` here, about fifty times inside the bound, so it is not a flake risk.

What I did not check: `services/validator` never ran, because of the GoBDK cgo link failure you already noted, so `TestRejectedTxReason` is unverified by me. Nothing here was exercised on a live mesh, so the claim that the new defaults sit above honest rejection traffic is untested against real volume.

services/validator/rejected_tx_reason.go
```diff
@@ -0,0 +1,35 @@
...
+// the rejected-tx Kafka message, which the p2p service re-broadcasts to every
+// mesh peer. The raw error text is unbounded and partly attacker-shaped (it can
+// embed script and transaction details from the rejected input), so the
+// network copy carries only the error codes along the chain (errors.CodeChain:
+// outermost first, deduplicated, at most maxRejectedTxReasonCodes) joined by
+// "/", e.g. "TX_INVALID/PROCESSING" (the enum names print without the ERR_
+// prefix). An error carrying no teranode code at all reports TX_INVALID, the
+// condition that led here.
+func rejectedTxReason(err error) string {
+	codes := errors.CodeChain(err, maxRejectedTxReasonCodes)
```
freemans13: For very nearly every rejection this resolves to the single code `TX_INVALID`.

All 13 `NewTxInvalidError` calls in `services/validator/TxValidator.go` pass a bare message with no wrapped error, so the chain is one link long. `mapBDKValidationError` wraps an inner `ERR_TX_INVALID` inside an outer one at `services/validator/ScriptVerifierGoBDK.go:366` and `:387`, which the dedup in `CodeChain` collapses to one code, as your own `repeated codes deduplicated` case shows. Only the policy branch yields a second code.

The discriminating information for those 13 sites is the `bad-txns-*` literal in the message, which is Bitcoin's standard reject vocabulary and carries nothing attacker-shaped. Full reasoning and the documentation consequence in the review body.

settings/p2p_settings.go
```diff
@@ -57,6 +57,10 @@ type P2PSettings struct {
...
...ish_rate" desc:"Steady-state re-broadcasts per second of internally rejected transactions on the rejected_tx gossip topic" default:"10" category:"P2P" usage:"Caps egress amplification of a junk-transaction flood" type:"int" longdesc:"### Purpose\nBounds how many internally rejected transactions per second the p2p service re-broadcasts to the mesh. Every rejection costs an attacker one invalid transaction submitted to the propagation surface, while each re-broadcast is fanned out by gossipsub to every mesh peer at this node's expense, so without a cap a junk flood is amplified network-wide.\n\n### How It Works\nA token bucket refilled at this rate, with p2p_rejected_tx_publish_burst tokens of headroom. Rejections beyond the budget are dropped, not queued, and counted in teranode_p2p_rejected_tx_publish_suppressed_total{reason=\"rate_limited\"}. Repeated txids are deduplicated before the bucket (reason=\"duplicate\") so they never spend a token.\n\n### Recommendations\n- **10** (default). Peers treat the topic as informational and take no action on it, so the budget only has to cover the honest trickle of rejections.\n- 0 or negative selects the default; there is no unlimited mode."`
```
freemans13: Not these two lines, but line 24 of this same file. The `RejectedTxTopic` longdesc still says "Message contains TX hash, rejection reason (double_spend, invalid_signature, etc.), and Peer ID", which this PR makes false: the reason is now an error-code list, in practice `TX_INVALID`.

`settings/kafka_settings.go:16` has the same problem, listing "Invalid signature / Missing parent / Double-spend / Policy violation / Insufficient fee" as the reasons carried. Both are the longdesc source that generates `docs/references/settings/`, which you regenerated for the two new keys, so the stale text ships to operators alongside the accurate new rows.

services/p2p/rejected_tx_gate.go
```diff
@@ -0,0 +1,116 @@
...
+// When it may not, reason is one of the rejectedTxSuppressed* labels. The
+// dedup check runs first so a repeated txid never spends a rate-limit token;
+// a txid refused by the rate limiter has its dedup grant returned so it is
+// not also treated as already announced.
+func (g *rejectedTxEgressGate) allow(txID, selfID string, now time.Time) (ok bool, reason string) {
+	g.mu.Lock()
+	limiter := g.limiterLocked()
+	g.mu.Unlock()
+
+	if publish, _ := g.seen.Check(txID, selfID, now); !publish {
```
freemans13, line 203: Worth a word on what actually sets the dedup period, because neither value passed to `setLimits` does.

With this node as the sole announcer, a repeat inside the window is refused on the budget. After `seenHashPublishWindow` rolls over, `prevGrantees` still holds us, so the rollover retry grant is refused too. Only at the second rollover is `grantees` empty, `prevGrantees` nil and the txid publishable again. Under a sustained repeat the observed period is therefore about two windows, roughly 30 seconds, governed by an unexported constant. `p2p_seen_hash_ttl` only decides when the entry resets to fresh.

The comment's "at most once per publish window" is a true upper bound, so this is not wrong, but an operator reading the two new settings has no way to work out what the duplicate suppression period really is.

icellan:
Re-reviewed at `1cd869ea2`. All three previous findings are genuinely fixed — checked against the code and by running the tests that pin the exact failure modes, not from the commit message:

- `CodeChain` now bounds by **links visited** (`errors.Walk`, `maxChainWalkDepth = 64`), counting foreign links too, so cyclic and same-coded chains both terminate. The 50k-link and cycle tests cover it.
- `rejectedTxEgress.deleteExpired` is wired into `cleanupPeerMaps` (`server_helpers.go:1020`) and into the stats line.
- The `!sent` path refunds the token via `grant.publishFailed()`, and `publishToNetwork` returns `sent == (err == nil)` on all six exits, so no drop path escapes the refund.

`go vet` clean and `go test -race` green on `errors`, `services/p2p`, `services/validator`, `settings`. The reject-literal allowlist premise holds up: exact equality on `(*Error).Message()` against unformatted literals, and `errors.New` only runs `fmt.Errorf` when params are non-empty, so a formatted message can never match — substring smuggling is explicitly tested. Token accounting has no double-refund and no refund-without-consume. No blockers.

Two things left worth fixing:

**`p2p_seen_hash_ttl` below 15s silently disables the per-txid dedup.** `rejected_tx_gate.go:129` passes `SeenHashTTL` straight into `seen.setLimits`, and `Check`'s "accounting window expired -> fresh announcement, publish=true" branch (`seen_hash_cache.go:277`) fires *before* the 15s publish window is ever consulted. Measured: `ttl = 5s`, one repeatedly-rejected txid, 20 rejections 6s apart -> 20/20 grants; at the 2m default the same shape gives the documented ~30s period. That contradicts the gate's own doc (`rejected_tx_gate.go:53-62`), the new longdesc (`settings/p2p_settings.go:1162`), and `p2p_seen_hash_ttl`'s own text claiming suppression "re-opens on a much shorter fixed publish window (15s)". The token bucket still caps egress, so it's not a hole — but it makes a knob that exists for the block/subtree dedup silently load-bearing for this defence. Clamping the effective TTL to `>= seenHashPublishWindow` in `ttlLocked()` fixes it for all three consumers.

**Every settings-to-gate wiring seam is untested, including the one this commit just repaired.** `Server.go:774` is the only thing making `p2p_rejected_tx_publish_rate` / `_burst` do anything; every gate test calls `setLimits` directly and the settings test only proves the loader fills the struct. Delete line 774 and the whole suite stays green with both settings inert. That's the same failure this package already carries a named regression test for — `TestApplyPeerMapLimits_ConfiguresSeenHashCaches` (`seen_hash_cache_test.go:303`: *"the three p2p_seen_hash_* keys were once declared with struct tags but never read by the loader"*) — and the PR adds two settings of the same kind without extending it. Same class, also unpinned: `server_helpers.go:1020` (delete it, the retention issue returns, suite green — the sibling maps have exactly this test at `capped_peer_map_test.go:833`), `Server.go:2661` (`clear()` on Stop, including the deliberate "bucket is not reset"), and the pre-gate return for external rejections at `Server.go:1195`, so nothing pins that mesh peers can't drain our bucket.

Minor:

- `metrics.go:62` — `duplicate = same txid already re-broadcast within the publish window` is wrong on both halves. The period is two windows (~30s), as `TestRejectedTxEgressGate_RepeatPeriodIsTwoPublishWindows` asserts, and on the first rollover the refusal is because last window's grantee was us, so nothing was re-broadcast "within" that window. A tx rejected once a second for a minute gives 2 publishes and 58 `duplicate` increments, roughly half of them rollover refusals. The longdesc gets this right; the metric help is the one place that doesn't.
- `maxChainWalkDepth = 64` vs `maxIsChainDepth = 1<<20` in `(*Error).Is` / `DeepestPublicCause`. The latter is pure cycle insurance, deliberately above any legitimate depth; 64 is a truncation. On the block-820116 mass-spend shape the new comment cites, `DeepestPublicCause` finds the innermost verdict while `rejectedTxReason` stops at link 64 and emits a bare `TX_INVALID` — yet its doc says the deepest-literal rule "mirrors `errors.DeepestPublicCause`". Fine as a decision, worth saying it's a choice to lose the deep detail. Expressing `Is`/`DeepestPublicCause` in terms of `Walk` would leave the package with one walking primitive instead of three loops and two bounds.
- `rejected_tx_reason.go:37` adds a *message-literal* allowlist while `errors/errors.go:773` already holds a *code* allowlist (`publicCauseCodes`, including `ERR_TX_INVALID`/`ERR_TX_POLICY`) whose comment asserts those messages carry no node-internal state and surfaces them verbatim. Both positions are defensible together — echoing a submitter's own input back to them isn't the same as fanning it out to every mesh peer — but nothing reconciles them; one sentence naming `publicCauseCodes` and why gossip is held to the stricter bar would stop the next reader widening the literal list casually. Related: `gossip_field_validation.go:156-158` still says the reason "is validator error text and may legitimately be long", which is still true for ingress but no longer for what this node sends.
- `rejected_tx_gate_test.go:80` asserts `published <= 6` on wall clock ("the loop runs in well under a second"). The gate takes an injectable `now` but the handler calls `time.Now()`. ~10ms of work so the margin is large, but a CI stall turns it into a rare failure; the exact-cap assertion already exists at gate level, so the handler-level upper bound can go.

services/p2p/Server.go
```diff
@@ -1188,23 +1193,40 @@ func (s *Server) rejectedTxHandler(ctx context.Context) func(msg *kafka.KafkaMes
...
+			rejectedTxPublishSuppressed(why)
+			s.logger.Debugf("[rejectedTxHandler] not re-broadcasting rejected tx %s: %s", txID, why)
+
+			return nil
+		}
+
 		rejectedTxMessage := RejectedTxMessage{
-			TxID:   hash.String(),
+			TxID:   txID,
 			Reason: m.Reason,
```
freemans13: The reason is sanitised at the producer, not at the chokepoint, so half this fix depends on which validator built the message.

This line copies `m.Reason` straight from the Kafka message onto the wire. The only egress bound on it is `sanitizeFields`, which truncates at 1024 bytes and strips markup characters. It never checks the shape.

During a rolling deploy a pre-upgrade validator publishes `err.Error()` on the same in-cluster topic, and this new p2p pod re-broadcasts up to 1KB of attacker-shaped script text to every mesh peer. That is no worse than today, so it is not a regression, and the rate limit still applies. But the PR body's "only the copy that travels network-wide loses the free text" holds only when both pods are new.

Everything else here prefers the chokepoint. `publishToNetwork` re-enforces the ingress size cap outbound precisely because a producer-side check cannot be trusted, and the comment four lines below calls `sanitizeFields` a "self-check against the bounds we enforce on ingress". A shape check in the same spirit, falling back to `TX_INVALID` when the reason is not `CODE(/CODE)*(: <allowlisted literal>)?`, would make the guarantee independent of the validator version. Note that `maxRejectedTxMessageSize` at `Server.go:121` already reasons about mixed-version operation, but only for un-upgraded remote peers, never for our own validator.

services/validator/rejected_tx_reason_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+// Every allowlisted detail must be a fixed literal: no format verbs and no
+// characters the p2p ingress sanitizer would strip, so what the validator
+// sends is exactly what peers log.
+func TestRejectedTxReasonDetails_AreFixedLiterals(t *testing.T) {
+	for detail := range rejectedTxReasonDetails {
+		require.NotContains(t, detail, "%", "detail %q must not be a format string", detail)
+
+		for _, r := range detail {
+			require.True(t, r >= 0x20 && r <= 0x7E, "detail %q must be printable ASCII", detail)
```
freemans13: This asserts a hand-copy of the sanitizer rather than the sanitizer.

`services/p2p/sanitize.go:56-64` is the code that decides what survives, and these two lines restate its two rules (`r < 0x20 || r > 0x7E`, then the `<>&"'` backtick backslash strip set) from a different package. Widen that strip set in p2p, say to stop a colon or a space being embedded in a log line, and this test stays green while every allowlisted detail silently loses characters on the wire. The comment above claims "no characters the p2p ingress sanitizer would strip", which is a claim about code this test never calls.

`services/p2p` is where the real pin belongs, and `TestRejectedTxMessage_CodeReasonPassesIngressBounds` is the right test for it. But its reason at `rejected_tx_gate_test.go:277` is `TX_INVALID/TX_INVALID/TX_INVALID/PROCESSING`, a pure code list. The `: bad-txns-in-belowout` form is the only one carrying a colon, a space and hyphens, and it is the entire reason the allowlist exists. Nothing runs it through `sanitizePeerDisplayString`.

I checked all twelve entries by hand against the current sanitizer and every one survives, so this is a coverage gap, not a live bug. Asserting `sanitizePeerDisplayString(detail, maxGossipReasonLen) == detail` over the set from the p2p side would close it.

services/p2p/rejected_tx_gate.go
```diff
@@ -0,0 +1,219 @@
...
+
+	return &rejectedTxGrant{gate: g, bucket: bucket, txID: txID, selfID: selfID}, ""
+}
+
+// publishFailed returns the grant after a re-broadcast that allow approved did
+// not actually reach the network: the dedup slot, so the next rejection of the
+// same txid can retry rather than being suppressed as a duplicate, and the
+// rate token, since nothing was sent.
+func (gr *rejectedTxGrant) publishFailed() {
+	gr.bucket.refund(time.Now())
```
freemans13: The one bucket operation that reads the wall clock.

`allow`, `take`, `refund` and `deleteExpired` all take an injected `now`, and the handler passes `time.Now()` into `allow`, so production is consistent and nothing is wrong today.

It is a trap for the next test. `advanceLocked` sets `b.last` from whatever it is handed, so a gate driven on a synthetic clock has its bucket timeline rewritten by a single refund. With a base of `time.Unix(0, 0)` the refund credits 56 years of refill, pins `b.last` in the present, and every later injected `take` sees time running backwards and never refills again. No current test hits it, because none of the synthetic-clock tests calls `publishFailed`, and `TestRejectedTxEgressGate_PublishFailedReturnsToken` anchors on `time.Now()`.

Taking `now time.Time` on `publishFailed` and threading it down from the handler removes the seam.

freemans13 (approved):
Approve. Read at `1cd869ea2` (branch tip and PR head agree) and ran the tests rather than trusting the commit message.

```
SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./errors/ ./settings/ ./services/p2p/
ok  github.com/bsv-blockchain/teranode/errors        2.239s
ok  github.com/bsv-blockchain/teranode/settings      3.183s
ok  github.com/bsv-blockchain/teranode/services/p2p 14.220s
```

My three earlier comments are fixed in the code, not just claimed. The `bad-txns-*` vocabulary travels through the `rejectedTxReasonDetails` allowlist with deepest-match. Both stale longdescs are rewritten to describe the real flow. The two-window dedup period is documented on the gate, in the `p2p_rejected_tx_publish_rate` longdesc, and pinned by a test. I replied on those threads.

The two new keys are genuinely wired rather than struct-tag-only: `settings/settings.go:546-547` reads both. Worth stating because this package has shipped tagged-but-dead settings three times.

Three new comments inline, none blocking, in the order I would act on them:

1. The free-text half of this fix is enforced at the validator, not at the p2p egress chokepoint, so it only lands once the validator pod is upgraded too. The rate limit is unaffected by this.
2. The claim that every allowlisted reject literal survives the p2p sanitizer is asserted against a hand-copy of the sanitizer living in a different package, and the detail-carrying reason shape never runs through the real one.
3. `publishFailed` reads the wall clock where every other bucket operation takes an injected `now`.

I checked @icellan's two open items against the code and both hold. No test calls `applyPeerMapLimits` and then asserts anything about `rejectedTxEgress`, so `Server.go:774` is unpinned. And `ttlLocked` passes a sub-window TTL straight through to the expiry branch in `Check`, which fires before the publish window is ever consulted.

Limit worth stating: nobody has run this against real rejection volume. The defaults of 10/s and 100 burst are an estimate of the honest trickle, and the only way to find out is `reason="rate_limited"` on a live mesh.

services/validator/rejected_tx_reason.go
freemans13: Fixed at `1cd869ea2`, verified in the code. `rejectedTxReasonDetails` in `services/validator/rejected_tx_reason.go` is a closed set of twelve fixed literals matched by exact equality on `(*Error).Message()`, deepest match wins, and `TestRejectedTxReason` pins both the substring non-match and the formatted non-match. A low-fee tx, a negative output and a bad signature now read differently on the receiving node.

settings/p2p_settings.go
freemans13: Fixed at `1cd869ea2`, both of them. The `RejectedTxTopic` longdesc now describes an informational topic carrying consensus rejections with an egress cap, and `KAFKA_REJECTEDTX` no longer claims a reason vocabulary of `double_spend` / `invalid_signature` or a filter that suppresses future announcements. The `desc` fields are unchanged, so the `docs/references/settings/` tables needed nothing beyond the two new rows, which are there.

services/p2p/rejected_tx_gate.go
freemans13: Fixed at `1cd869ea2`, and better than I asked for. The two-window period is now spelled out in three places an operator or a maintainer would actually look: the `Dedup period` paragraph on `rejectedTxEgressGate`, a `### Dedup period` section in the `p2p_rejected_tx_publish_rate` longdesc that says plainly the 15s window is a constant and not a setting, and `TestRejectedTxEgressGate_RepeatPeriodIsTwoPublishWindows`, which walks the refusal inside the window, the refusal on first rollover and the grant on the second.

icellan:
Re-reviewed at `4a60d7a22`. Both findings from the last round are fixed, and the new `util/rejectedtx` package turns the reason grammar into an enforced contract at the egress chokepoint rather than a convention — better than what I asked for.

- `ttlLocked()` (`seen_hash_cache.go:184-193`) floors the effective TTL at `seenHashPublishWindow`, covering all three consumers, with a comment naming the exact mechanism (Check's expiry branch runs before the publish window is consulted).
- `TestApplyPeerMapLimits_ConfiguresRejectedTxEgress` (`rejected_tx_gate_test.go:433`) now guards the settings-to-gate path end to end.

Re-checked at this head: the dedup-before-bucket ordering, the `sent bool` refund contract, and `advanceLocked`'s monotonicity guard still hold; the `ERR_name`/`ERR_value` round-trip agrees between `Reason` and `Valid` (enum names carry no `ERR_` prefix); the `Cut(": ")` grammar rejects a pre-upgrade `"TX_INVALID (30): ..."`, so no free text escapes through a mixed-version mesh; and the `errMsg*` literals are bare, so the allowlist stays exact-match-safe. `go test` (and `-race` on p2p) green across `errors`, `settings`, `util/rejectedtx`, `services/p2p`; `go vet` clean.

Two low-severity items left:

**`services/p2p/seen_hash_cache.go:185` — the new clamp is silent.** An operator setting `p2p_seen_hash_ttl=5s` now runs 15s with nothing in the log, and the coercion applies to `blockSeenHashes` and `subtreeSeenHashes` too, not just the new gate. That's the failure mode `announcePeerMapLimits` was written to prevent — its own comment says *"A silently-coerced value looks exactly like a value that was never read"* — and every other coerced bound in `applyPeerMapLimits` gets a startup line. Worth one for `SeenHashTTL < seenHashPublishWindow`.

**`services/p2p/rejected_tx_gate_test.go:459` — the TTL leg of the new wiring test is vacuous.** With `SeenHashMaxSize: 5` the test performs 9 distinct inserts (`distinctTxID(0..7)` in the loop, then `distinctTxID(100)`), so `distinctTxID(0)` was evicted by the size cap long before `clock.advance(time.Minute)`. The final `allow` would be granted with any TTL, including an unwired one. The burst and rate legs do bite. Given this PR's own "delete the call and the test fails" standard for the settings path, assert on a txid that's still resident, or raise the size cap for that leg.

util/rejectedtx/reason.go
```diff
@@ -0,0 +1,184 @@
...
+// Normalize returns reason unchanged when it fits the grammar and Fallback
+// otherwise. It is the chokepoint's guard: a reason built by an un-upgraded
+// validator, or by anything else on the in-cluster topic, never leaves the
+// node as free text.
+func Normalize(reason string) string {
+	if Valid(reason) {
+		return reason
+	}
+
+	return Fallback
```
freemans13, line 183: `Normalize` throws away a valid code list when only the detail fails the allowlist.

The rolling upgrade this commit guards against also runs the other way. Say a later release adds a literal to `details`. Its validator sends `TX_INVALID/TX_POLICY: <new literal>` to a p2p pod still on this release, and the node announces a bare `TX_INVALID`, losing a `TX_POLICY` code that would have passed on its own.

Falling back to the code list when `Valid(codeList)` holds, and to `Fallback` only when the codes fail too, keeps everything that is safe to send. Free text still cannot get through: `TX_INVALID (30): ...` has an invalid code list and still collapses to `Fallback`. Not blocking, since the topic is informational and the loss only lasts for that upgrade window.

freemans13 (approved):
Approve at `4a60d7a22`, which is both the branch tip and the PR head. My approval on 2026-09-10 was on `1cd869ea2`, and this push reworked the reason code substantially, so I re-read the whole diff rather than carrying the old verdict forward.

My three inline comments from that round are fixed. I checked each one by breaking the fix in a scratch worktree and watching the tests fail:

1. The reason grammar is now enforced at the p2p egress point. `rejectedTxHandler` runs the reason through `rejectedtx.Normalize`, so a pre-upgrade validator's error text leaves the node as `TX_INVALID`. Swapping that call for the raw `m.Reason` fails `TestRejectedTxHandler_NormalizesReasonAtChokepoint` and `TestRejectedTxHandler_InternalRejectionReplacesFreeTextReason`.
2. The sanitizer test now runs every allowlisted detail, and the `CODE/CODE: detail` shape, through the real `sanitizePeerDisplayString` in `services/p2p`. The hand-copy is gone.
3. `publishFailed` reads the gate's clock. Pointing it back at `time.Now()` fails `TestRejectedTxEgressGate_PublishFailedUsesGateClock`.

The new TTL clamp in `seenHashCache.ttlLocked` is pinned too: removing it fails `TestRejectedTxEgressGate_SubWindowTTLStillDeduplicates` and `TestSeenHashCache_TTLIsClampedToPublishWindow`. I also checked that all eleven `details` entries are real. Nine are bare, unformatted messages in `services/validator`, and the other two are the constants the GoBDK verifier now builds its errors from. The only consumer of `KafkaRejectedTxTopicMessage` in the tree is `rejectedTxHandler`, so shortening the reason on the Kafka topic costs no other in-cluster reader anything.

@icellan's two open items from 2026-09-14 still stand, and I agree with both. One piece of evidence for the second: changing the TTL argument the gate gets in `applyPeerMapLimits` to `0` leaves `TestApplyPeerMapLimits_ConfiguresRejectedTxEgress` green, so its TTL leg does not test the wiring.

One new comment inline, not blocking: `Normalize` drops a valid code list when only the detail is unknown.

The failing SonarQube gate does not come from this diff. Its two findings are `go:S5332` (an `http://` literal) at `services/p2p/Server.go:2552` and `:2570`, and both lines come from `fdcde79adc`, which is already on main.

Verified locally at `4a60d7a22`:

```
SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./errors/ ./settings/ ./util/rejectedtx/ ./services/p2p/   ok
go vet ./services/validator/ ./util/rejectedtx/                                                                               ok
```

CI on this head is green apart from the Sonar gate.

Not checked: the `services/validator` tests, which do not link locally because of GoBDK cgo (CI's `test` job passed them). The 10/s rate and 100 burst defaults are still untested against real rejection volume.

services/p2p/Server.go
freemans13: Fixed at `4a60d7a22`. `rejectedTxHandler` now runs the reason through `rejectedtx.Normalize` before publishing, so the guarantee no longer depends on which validator built the message. I checked it by swapping the call for the raw `m.Reason` in a scratch worktree: `TestRejectedTxHandler_NormalizesReasonAtChokepoint` (the pre-upgrade error text and unlisted-detail cases) and `TestRejectedTxHandler_InternalRejectionReplacesFreeTextReason` both fail.

services/validator/rejected_tx_reason_test.go
freemans13: Fixed at `4a60d7a22`. The hand-copy is gone. The reason tests moved to `util/rejectedtx`, and `TestRejectedTxMessage_ReasonShapesPassIngressBounds` in `services/p2p` now runs every entry of `rejectedtx.Details()` through the real `sanitizePeerDisplayString`, including the `TX_INVALID/TX_POLICY: <detail>` shape with its colon and space.

services/p2p/rejected_tx_gate.go
freemans13: Fixed at `4a60d7a22`. `publishFailed` now reads the gate's injected clock, the same source `allow` uses. I checked it by pointing the refund back at `time.Now()` in a scratch worktree, and `TestRejectedTxEgressGate_PublishFailedUsesGateClock` fails.

icellan:
Re-reviewed at `9fd4e6b67`. That commit is a pure merge of upstream/main — the PR's own files are byte-identical to `4a60d7a22`: `util/rejectedtx/reason.go`, `services/p2p/rejected_tx_gate.go`, `services/p2p/seen_hash_cache.go` and `services/validator/rejected_tx_reason.go` unchanged, and `errors/error_utils.go` / `services/p2p/Server.go` differ only by upstream hunks (the `IsBlockCorrupt` / `IsMaliciousResponseError` ordering from #4692, nothing in the rejected-tx region). So the two items from my last comment are still open, and the three below are additional findings on the same code rather than anything you've had a chance to respond to.

The design still holds up on another pass: dedup-then-bucket ordering, the refund accounting against `seenHashCache.PublishFailed`, the `ttlLocked` clamp, the `Walk`/`CodeChain` depth bound, the closed detail allowlist (every literal verified verbatim in `TxValidator.go` / `ScriptVerifierGoBDK.go`) and the `Valid` grammar (`ERR_value` keys are prefix-free, so `TX_INVALID` round-trips). `errors`, `util/rejectedtx` and `settings` pass; `go vet` clean on all four packages. `services/p2p` wouldn't link locally — known macOS SDK/cgo breakage on my side, not the diff. No blockers.

**`services/p2p/Server.go:1276` — a persistently failing publish makes the gate a no-op and floods the log.** `if !sent { grant.publishFailed() }` hands back both the token and the dedup slot on every message, so while the mesh is down or the topic isn't joined the gate never engages. Together with the unconditional `s.logger.Errorf("[rejectedTxHandler] publish error: %v", err)` two lines above, a junk-tx flood during a gossip outage produces one Errorf per rejected transaction plus the full per-message work (Normalize, marshal) — the log flood the "counted, not warn-logged" design elsewhere in this PR is built to avoid. Either refund the dedup slot but keep the token when `err != nil` (distinct from a policy drop, where `err == nil`), or rate-limit that log line.

**`util/rejectedtx/reason.go:143` — `Valid` does unbounded string work.** The chokepoint calls `Normalize(m.Reason)` on the raw Kafka field, which during a rolling upgrade carries a pre-upgrade validator's `err.Error()` — up to the broker's message limit, and the known mass-spend chains render very large texts. `strings.Split(codeList, "/")` allocates a string header per `/`, so a ~1MB slash-dense reason costs ~16MB transient per accepted message. Bounded to roughly the gate's rate per second so it isn't a DoS, but a `len(reason) > bound -> false` guard at the top of `Valid` is a one-liner.

**`services/p2p/seen_hash_cache.go:189` — at a clamped TTL the repeat period is one window, not two.** For any configured `p2p_seen_hash_ttl <= 15s` the effective TTL is exactly `seenHashPublishWindow`, so `now.Sub(firstSeen) < ttlLocked()` (`:254`) goes false at the same instant the window rolls: `Check` takes the expiry branch, the entry is treated as fresh, and the rejected-tx repeat period becomes ~15s rather than the ~30s stated in the gate doc comment, the `p2p_rejected_tx_publish_rate` longdesc and the metric help. `TestSeenHashCache_TTLIsClampedToPublishWindow:342` encodes exactly this (`publish == true` at the window boundary). Either clamp to strictly greater than the window, or say the two-window period assumes a TTL above it.

Minor, take or leave: `Valid`'s `len(codes) == 0` branch is unreachable (`strings.Split` never returns an empty slice); `errors.Walk` uses stdlib `errors.Unwrap` while `(*Error).Is` and `SetWrappedErr` use `errors.As` to dig through foreign wrappers, so a foreign link implementing `As` but not `Unwrap` truncates the reason (degradation only, no leak); and `announcePeerMapLimits` still doesn't warn on out-of-range values for the two new keys the way it does for the older ones.

Issue #1701, opened 2026-09-07
[BUG] Pruner filter false positives let mined transactions re-enter block assembly


oskarszoon:
## Describe the bug

The Aerospike pruner can delete a confirmed transaction without recording the `deletedChildren` marker on a surviving parent. A later replay of the same transaction can then pass the idempotent-spend check and recreate its record as **unmined**, with empty `blockIDs` and reset output-spend state. This makes already-mined transactions eligible for block assembly again.

The concrete trigger reproduced here is a **false positive in `PrunedTxSet`**, the cuckoo filter used to skip parent updates. This is a correctness decision based on an approximate membership result.

## Reproduced mechanism

1. `processRecordChunk` calls `prunedSet.CheckAndRemove(parentTxID)`. A hit skips the parent's `deletedChildren` update, without checking whether that parent actually exists.
2. A false positive suppresses the marker on a live parent. The child record is still deleted.
3. `spendMulti` accepts an identical existing spender as an idempotent retry **unless** the child is in the parent's `deletedChildren` map.
4. `SequentialSpendAndCreate` subsequently recreates the absent child. With no mined information supplied, `Create` writes empty mined fields and sets `unminedSince`; outputs start unspent.
5. The validator's successful-create path can send it to assembly. Assembly reload cannot detect its historical confirmation from empty `blockIDs`; checking that the parent is spent by this same transaction does not recover that history.

Pinned source:

- [Parent-update skip](https://github.com/bsv-blockchain/teranode/blob/25748b9c2da85969a1133b39ec2e043b056f3733/stores/utxo/aerospike/pruner/pruner_service.go#L1174-L1180)
- [Same-spender replay guard](https://github.com/bsv-blockchain/teranode/blob/25748b9c2da85969a1133b39ec2e043b056f3733/stores/utxo/aerospike/teranode.lua#L385-L405)
- [Spend followed by create](https://github.com/bsv-blockchain/teranode/blob/25748b9c2da85969a1133b39ec2e043b056f3733/stores/utxo/spend_and_create.go#L61-L102)
- [Empty mined fields and unmined marker on creation](https://github.com/bsv-blockchain/teranode/blob/25748b9c2da85969a1133b39ec2e043b056f3733/stores/utxo/aerospike/create.go#L909-L945)
- [Validator admission to assembly](https://github.com/bsv-blockchain/teranode/blob/25748b9c2da85969a1133b39ec2e043b056f3733/services/validator/Validator.go#L959-L988)

[...]

PR #1706, opened 2026-09-08
fix(utxo): do not skip parent markers via cuckoo filter

icellan:
PrunedTxSet is probabilistic. A false positive suppressed deletedChildren on a live parent, so a replay of a pruned child could recreate it unmined.

Leave prunedSet nil so every parent gets an addDeletedChildren write. Missing parents already return TX_NOT_FOUND.

Refs #1701 — this closes the probabilistic route only. With `pruner_utxoDefensiveEnabled=false` (the default) parent updates and child deletions still go out in one non-atomic `BatchOperate`, so a failed parent update alongside a successful child deletion reaches the same end state, and the `delete_at_height` rescan cannot recover it. Ordering that path like defensive mode is follow-up work, not this hotfix, so the issue stays open.


stores/utxo/aerospike/pruner_replay_protection_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	astore.ResetPrunerServiceForTests()
+	t.Cleanup(astore.ResetPrunerServiceForTests)
+	require.NoError(t, store.CreateIndexIfNotExists(ctx, apruner.IndexName, fields.DeleteAtHeight.String(), aerospike.NUMERIC))
+	require.NoError(t, store.WaitForIndexReady(ctx, apruner.IndexName))
+
+	svc, err := store.GetPrunerService()
+	require.NoError(t, err)
+	n, err := svc.(*apruner.Service).PruneWithPartitions(ctx, 1300, "replay-protection", 1)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
icellan: Fixed in 1de5591. GetPrunerService is now require.NotNil, and the *pruner.Service assertion is checked with require.True.

stores/utxo/aerospike/pruner/pruner_service.go
```diff
@@ -398,14 +397,11 @@ func NewService(settings *settings.Settings, opts Options) (*Service, error) {
...
-	// registers them.
-	if !service.defensiveEnabled {
-		service.prunedSet = NewPrunedTxSet(256, settings.Pruner.UTXOPrunedSetMaxEntries)
-	}
-
+	// Do not construct PrunedTxSet. It is a cuckoo filter used to skip
+	// parent deletedChildren writes; a false positive drops the replay
+	// marker on a live parent. prunedSet stays nil so every parent is
+	// updated. Missing parents already return TX_NOT_FOUND / KEY_NOT_FOUND
+	// and are counted as skipped.
```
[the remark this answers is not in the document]
icellan: Leaving the setting. This PR is only turning the skip off. pruner_utxoPrunedSetMaxEntries is unused while prunedSet stays nil; deleting it is a follow-up, not this hotfix.

stores/utxo/aerospike/pruner_replay_protection_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/bsv-blockchain/teranode/stores/utxo/fields"
+	"github.com/bsv-blockchain/teranode/ulogger"
+	"github.com/bsv-blockchain/teranode/util/test"
+	"github.com/stretchr/testify/require"
+)
+
+// TestPrunerReplayProtection: pruning a fully-spent child whose parent still
+// has another output must leave deletedChildren on that parent, so
+// SpendAndCreate cannot recreate the child as unmined.
+func TestPrunerReplayProtection(t *testing.T) {
```
freemans13, line 22: This test passes on `82771682f`, this branch minus its two commits, with only this file added: 3 runs out of 3, 10.9s.

The parent never enters the filter. `processRecordChunk` Adds only the TXIDs of records it is deleting, and this parent survives pruning because it still has an unspent output, so `CheckAndRemove(parent)` hits only on a real cuckoo false positive. The test confirms the happy path still works; it does not pin the bug.

Two shapes that would fail without the fix are in the review body.

stores/utxo/aerospike/pruner/pruner_service.go
freemans13, line 404: Worth saying here what this does not cover. With defensive mode off, which is the default, parent updates and child deletions still go out in one non-atomic `BatchOperate`, so a failed parent update alongside a successful child deletion reaches the same end state this comment is guarding against, and the `delete_at_height` rescan cannot recover it. Detail in the review body.

freemans13 (changes requested):
The fix itself is right and should land. A cuckoo filter only ever answers "maybe present", so it can never license skipping a write, and the write it was skipping is the one that stops a pruned child being recreated as unmined. Two things block merge, and four are notes.

### 1. The new test passes without the fix

I ran `pruner_replay_protection_test.go` against `82771682f`, this branch minus its two commits, with the test file copied in and nothing else changed. It passed 3 runs out of 3 in 10.9s.

The fixture never gets the parent into the filter. `processRecordChunk` Adds only the TXIDs of the records it is deleting, and this parent survives pruning because it still holds an unspent output, so `CheckAndRemove(parent)` returns true only on a genuine cuckoo false positive, which the comment you deleted put at roughly 3%. The test proves the happy path still works. It cannot tell the fix from its absence, so it will not stop the skip coming back.

Issue #1701 already ships a fixture that does pin it: the "seed filter collision" case seeds a synthetic hash sharing the live parent's fingerprint, index and shard, then drives the real pruner. Lifting that in would be the regression test. If that is too fiddly for a hotfix, a unit assertion in the pruner package that `NewService` leaves `prunedSet` nil with defensive mode off also fails on `82771682f`, which is the minimum bar.

### 2. "Fixes #1701" would close an issue that is half fixed

#1701 documents two routes to a missing marker. This PR removes the probabilistic one. The deterministic one is untouched, and it is the production path, because `pruner_utxoDefensiveEnabled` defaults to false (`settings/settings.go:615`).

With defensive mode off, `flushCleanupBatches` sends parent updates and child deletions in one `BatchOperate` (`pruner_service.go:1590-1596`), and `executeBatchCleanupCombined` tallies per-record outcomes only after that batch has already run (`pruner_service.go:1731-1745`). A parent-update record can fail in the same batch where the child deletion succeeds. The function returns an error, but the retry cannot recover it: phase 2 rediscovers work through a secondary-index range query on `delete_at_height` (`pruner_service.go:575-579`), and the child record it would have re-found is already deleted. The end state is exactly the one this PR is closing, a deleted child with no marker on its surviving parent, and a replay that recreates it unmined.

Merging with the closing keyword auto-closes #1701 while that stays live. Either drop it to a plain reference, or fix the ordering here too.

### 3. The belief that caused the bug is still in the code, holding up the surviving hazard

`pruner_service.go:1578-1581` still says the `deletedChildren` bin "is never read" when defensive mode is off. `teranode.lua:391` reads it on every idempotent re-spend, defensive mode or not. That is the same wrong premise as the comment this PR deleted, and here it is load-bearing: it is the stated justification for merging the two batches into one. Thirty lines further down `executeBatchParentUpdates` says the opposite, "This must complete before Phase 2b (child deletion) to maintain referential integrity" (`pruner_service.go:1685`). The docs pass in `1de5591` caught the skip comments and missed this pair.

### 4. The suite still certifies the behaviour you removed

`TestProcessRecordChunk_SkipsParentsInPrunedSet` (`stores/utxo/aerospike/pruner/pruned_set_skip_test.go:103`) still passes on this branch, I ran it. It asserts that a parent found in the filter gets its update omitted, and it never checks whether that parent still exists. The skip path is now dead but green, so re-enabling it is a one-line change with a passing test to bless it. Either delete the path along with the field, or make that test assert the parent-existence precondition it silently assumes.

### 5. The cost is real and unmeasured

Every input of every pruned transaction now issues a parent-update record, including parents pruned earlier in the same session, which is the case the skip existed for. The benchmark report on this PR covers 139 benchmarks and none of them touch the pruner, so "0 regressions" says nothing about this change. The only pruner benchmarks in the repo are the cuckoo and map microbenchmarks, which measure the thing being switched off. `utxo_pruner_parents_skipped_pruned_total` on a live node is the number that sizes what is being given back. Not a reason to hold a correctness fix, but worth stating rather than assuming free, since the pruner has to keep up to bound disk.

### 6. Minor: the production write path is not in CI

"test aerospike native-ops (fork image)" reports `skipping` on this run. In production the parent marker goes through the native-op wrapper `subOpAddDeletedChildren` (`stores/utxo/aerospike/pruner_provider.go:74`), and this fix makes correctness depend on that write landing for every parent. That variant is not exercised.

Not repeating the `pruner_utxoPrunedSetMaxEntries` point. Copilot raised it and your answer reads fine for a hotfix.

### What I ran

- `stores/utxo/aerospike/pruner` full package on `1de5591c2`: green, 38.6s.
- `TestPrunerReplayProtection` on `1de5591c2`: green, 4.5s.
- `TestPrunerReplayProtection` on `82771682f` with only the test file added: green, 3 of 3, 10.9s. That is the finding in item 1.

I have not run this on a node, so the throughput point in item 5 is reasoning, not measurement.

stores/utxo/aerospike/pruner_replay_protection_test.go
icellan: Right, and confirmed: the behavioural test passes on the unfixed code. 8eabff4 adds stores/utxo/aerospike/pruner/pruned_set_disabled_test.go, a package-pruner unit test asserting NewService leaves prunedSet nil with defensive mode both off and on. Verified it fails on 82771682f (prunedSet must stay nil (defensiveEnabled=false)) and passes on this branch. No Aerospike, no reflect, no exported field. Took the NewService assertion rather than the seeded-collision fixture: it pins the construction site, which is the actual lock, and runs in 0.7s.

Also strengthened the behavioural test — it now re-checks the child is still absent after the replay and pins the error to "invalid spend" instead of the ErrUtxoError class.

stores/utxo/aerospike/pruner/pruner_service.go
icellan: Confirmed and not fixed here. teranode.lua:391 reads deletedChildren on every idempotent re-spend regardless of mode, so the comment at flushCleanupBatches claiming the bin is never read was wrong, and it was the justification for the single batch. 8eabff4 corrects that comment to state the hazard and point at #1701, but leaves the ordering alone: making the non-defensive path two-step doubles batch round-trips on the pruner hot path and is unmeasured, which is not something to land in a hotfix that also goes to a release branch.

PR body changed from Fixes #1701 to Refs #1701, with the surviving route spelled out, so merging will not auto-close it. Leaving this thread open.

stores/utxo/aerospike/pruner_replay_protection_test.go
freemans13: Verified in both directions rather than taking it on trust. `TestNewServiceLeavesPrunedSetNil` fails on `82771682f` on the `defensive=false` leg, with your own message, `prunedSet must stay nil (defensiveEnabled=false)`. The whole pruner package is green on `8eabff4d0` in 42.3s.

Agreed on taking the construction site over the seeded-collision fixture. It is the narrower lock, it needs no container, and the thing worth pinning really is "the set is never built".

I also checked the tightened string assertion does what you said. `ERROR_CODE_INVALID_SPEND` is set in exactly one place in `teranode.lua`, the `deletedChildren` branch at line 397, and `LuaErrorCodeInvalidSpend` is the only case in `createSpendError` that formats "invalid spend for vout" (`stores/utxo/aerospike/spend.go:1083`). So the `Contains` does pin the replay rejection rather than any spend failure.

One residual, not worth changing: the test pins `NewService`, so an assignment to `prunedSet` made anywhere else, or a non-nil set handed straight to `partitionWorkerFn`, would still slip past. Narrow now that the skip path itself is gone.

stores/utxo/aerospike/pruner/pruner_service.go
freemans13: Accepted, and the replacement comment is accurate. Deferring the ordering is the right call for a hotfix that also goes to a release branch: making the non-defensive path two-step doubles the batch round-trips on the prune hot path, and nobody has measured what that costs.

Two things worth carrying into the follow-up. Ordering alone will not fully close it, because the parent update and the child deletion stay two separate operations, so a crash between them lands in the same place; the defensive path has the same gap with a narrower window. And retry cannot rescue either shape, because phase 2 rediscovers work through the `delete_at_height` range query and the deleted child is no longer in that index (`stores/utxo/aerospike/pruner/pruner_service.go:575-579`).

Body change to `Refs #1701` confirmed. Leaving the thread open as you asked.

freemans13 (approved):
Approving. Both blockers are fixed, and I checked each one by running it rather than by reading the diff.

**The test now pins the fix.** `TestNewServiceLeavesPrunedSetNil` fails on `82771682f` on the `defensive=false` leg, with your own failure message. That is the property the previous test could not assert. Covering `defensive=true` as well is a nice touch, even though that leg passed before the fix too. The behavioural test is also stronger: `ERROR_CODE_INVALID_SPEND` is set in exactly one branch of `teranode.lua`, the `deletedChildren` check at line 397, and `LuaErrorCodeInvalidSpend` is the only case in `createSpendError` that produces that string, so the `Contains` really does pin the replay rejection rather than any spend failure.

**The issue will stay open.** `Refs #1701` with the surviving route written out in the body is the right shape. It is also now recorded in three places a maintainer will actually hit: the PR body, the corrected comment at `flushCleanupBatches`, and the deprecation note on the setting.

**The wrong premise is gone.** The old comment claimed the `deletedChildren` bin is never read with defensive mode off. The replacement says the opposite and names the consequence. That was the sentence holding up the single-batch design, so correcting it is worth more than it looks.

**The suite no longer certifies what you removed.** Deleting `TestProcessRecordChunk_SkipsParentsInPrunedSet` and renaming the file to `process_record_chunk_test.go` closes the trapdoor where re-enabling the skip would have come with a green test to bless it.

### Still open, none of it blocking

- Copilot's newest point on `settings/pruner_settings.go:30` is right. `desc` now reads "DEPRECATED (no-op)" but `usage` still describes live behaviour, "0 = use built-in 2B default; on saturation the current generation rotates". One line, and it is the field most likely to end up in generated operator docs on its own.
- Copilot's newest point on the metric help strings I do not agree with, and it is worth saying why rather than leaving it hanging. It asks the help text to state the metric is deprecated while `PrunedTxSet` is disabled. Every one of the four already opens with "Always 0" and cites the issue, and `utxo_pruner_pruned_set_saturated` goes further with "do NOT read 0 here as healthy", which is exactly the trap an operator would fall into. Nothing to change.
- Throughput is still unmeasured, which I would not hold a correctness fix for. The cheap version is not a benchmark: after this deploys, `utxo_pruner_parents_updated_skipped` counts the parent updates that now go out and come back not-found, which is the volume of round-trips being paid for the safety, and prune session duration says whether it matters. If both are flat, the question is closed for free.
- "test aerospike native-ops (fork image)" still reports `skipping`, so the native-op form of the `addDeletedChildren` write is the one production uses and the one CI never exercises. Pre-existing and not yours to fix here, but this change raises what rides on that write.

The dead `PrunedTxSet` implementation, its tests and its benchmarks are still in the tree. Leaving them for the follow-up is fine by me now that nothing constructs the set and nothing asserts the skip.

### What I ran

- `stores/utxo/aerospike/pruner` full package on `8eabff4d0`: green, 42.3s.
- `TestPrunerReplayProtection` on `8eabff4d0`, 2 runs: green, 8.2s.
- `TestNewServiceLeavesPrunedSetNil` on `82771682f` with only that file added: FAIL on `defensiveEnabled=false`, as intended.
- CI on this head: 42 pass, 1 skipping, the native-ops job above.

Still no node time from me, so the throughput point stays reasoning rather than measurement.

oskarszoon (approved):
Reviewed `8eabff4d0` independently. The premise is unarguable and the production change is byte-identical to #1705. Nothing blocks. Notes below plus a merge-order call.

## Agreement, not re-litigation

Your 08:38 findings on #1705 apply here and match mine, reached separately:

- The replay test is not load-bearing. Reverting the fix under overlay against a live Aerospike 8.0 leaves it green; `TestNewServiceLeavesPrunedSetNil` is the only assertion that fails on revert. On this base the fixture's `SpendAndCreate` at least writes a record, so the "must not recreate" assertion is meaningful here — it is vacuous on v0.15, which I've raised on #1705.
- The non-atomic combined batch is **pre-existing**: `flushCleanupBatches` is byte-identical to `82771682f`. Not introduced or aggravated here, so not a gate. #1701 is the right home for your reproduction.

## Merge order

Land this on `main` first, then rebase #1702 on top of it, and take #1705 to `release/v0.15` as-is. #1702 is main-only by construction — new proto enum plus a validator wire flag — and carries an open P0, so it should not gate either of these. Landing #1706 first also means #1702's deletion of `PrunedTxSet` / `util/cuckoo` applies to a tree where the skip is already gone, which is the smaller conflict.

## Minor

- `pruner/pruner_service.go:1152` — the live skip branch has no coverage now that the old test file went. The rewrite covers `processRecordChunk` but not this branch, and `process_record_chunk_test.go:31` leaves dead helpers behind.
- `settings/pruner_settings.go:30` — the `usage` tag for `pruner_utxoPrunedSetMaxEntries` still asserts live behaviour and ships to the operator portal as `usageHint` (`export.go:142`, `settings_handler.go:75`). `desc` was updated, `usage` wasn't.
- `pruner/pruner_service.go:636-655` — catch-rate comment documents the removed path; the real operator-visible shift is `parents_updated` vs `parents_updated_skipped`, undocumented.
- `pruner_provider.go:74` — the native `subOpAddDeletedChildren` write has no CI round-trip; the one end-to-end test sits behind a build tag no job supplies. This fix makes correctness depend on that write landing for every parent, and it is the production transport. Not a blocker at current defaults, worth its own issue.
- No pruner write-path benchmark exists and `benchmark-compare.yaml:52` excludes the package, so the "0 regressions" report doesn't speak to this change. Sizing has to be taken on a base build, since `parents_skipped_pruned` is permanently 0 after merge.

PR #1707, opened 2026-09-09
fix(test): make the disk-txmap throughput floor an order-of-magnitude gate

icellan:
Backport of the assertion change in `25c0f2c43` (#1051 / #1058), which `release/v0.15` never received.

`TestDiskTxMap_ExistenceLayerThroughput` gates on >1M ops/sec under `-race`. The CI runners land below that — 970482 then 910122 ops/sec on two consecutive runs of the same commit — so the `test` job fails on every PR targeting this branch, #1705 included.

The 1M figure becomes an informational log; the hard assertion drops to a 100k floor. That still fails on an order-of-magnitude regression (an accidental lock serialising the existence layer, say) without gating on runner speed.

`services/blockassembly/subtreeprocessor/disk_tx_map_benchmark_test.go` is now byte-identical to `main`.

Local: `go test -race -run TestDiskTxMap_ExistenceLayerThroughput` passes, 6249184 ops/sec.

### DIFF
--- a/services/blockassembly/subtreeprocessor/disk_tx_map_benchmark_test.go
+++ b/services/blockassembly/subtreeprocessor/disk_tx_map_benchmark_test.go
@@ -289,8 +289,18 @@ func TestDiskTxMap_ExistenceLayerThroughput(t *testing.T) {
 	t.Logf("Ops/sec:       %.0f", opsPerSec)
 	t.Logf("Ns/op:         %.0f", float64(elapsed.Nanoseconds())/float64(totalOps))
 
-	// Threshold is 1M ops/sec to pass under -race (race detector adds ~10x overhead).
-	// Without -race, this achieves ~35M ops/sec.
-	require.Greaterf(t, opsPerSec, 1_000_000.0,
-		"Existence layer must achieve >1M ops/sec (with -race), got %.0f ops/sec", opsPerSec)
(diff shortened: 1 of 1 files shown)


oskarszoon (approved):
The trade is right and it's disclosed in the comment rather than buried, so this should land. #1051 checks out — ~953k on a 16-core runner against a 1M floor, and the 32-core runner it currently passes on is meant to be temporary, so the flake is imminent rather than hypothetical.

Measured the remaining sensitivity, since an order-of-magnitude floor can easily end up gating nothing. M3 Max, `-count=1`:

- without `-race`: 23.1M–29.6M ops/sec — 231x–296x above the 100k floor
- with `-race`: 3.49M–6.03M ops/sec — 35x–60x above it

So it catches a collapse (an accidental lock serialising the existence layer) and nothing smaller. That matches what the title and comment claim. No flakes in 8 runs, none close to the floor. Timing mechanics are unaffected by the diff and correct — setup outside the timed region, fixed op count over wall-clock.

One nit, in the justification rather than the code: `disk_tx_map_benchmark_test.go:296-298` says the floor is "10x below target, ~35x below a measured slow CI run". The 10x is right, but 953k/100k is 9.5x, not 35x — 35x is the race-mode local figure (3.49M/100k). Two different measurement conditions read as one. Worth fixing so the reasoning behind the constant stays auditable.

Issue #1708, opened 2026-09-09
Add CLI recovery for replayed transactions and backport to v0.15


oskarszoon:
Previously confirmed transactions affected by #1701 can remain in the UTXO store as unmined and re-enter block assembly after an upgrade or restart. Operators need a supported, repeatable recovery command that discovers each node’s affected records, repairs them safely, and verifies the resulting assembly. Ship on main and backport to `release/v0.15`.

## Why existing reset options are insufficient

The affected state is: a recreated child has empty mined metadata and may have lost its output-spend state; its parent still records the original child as spender but lacks the replay-protection marker.

At release commit `45c603513b1f520f1f92e099172ca00d612b3312`:

- Ordinary reset/restart reloads records from the unmined iterator.
- `--full-reset` repairs inconsistencies involving existing block IDs but [skips empty block IDs](https://github.com/bsv-blockchain/teranode/blob/45c603513b1f520f1f92e099172ca00d612b3312/services/blockassembly/BlockAssembler.go#L2046). Its CLI help suggesting a mempool purge is misleading.
- [`--validate-inputs`](https://github.com/bsv-blockchain/teranode/blob/45c603513b1f520f1f92e099172ca00d612b3312/services/blockassembly/BlockAssembler.go#L2498) can accept the same original spender; it does not recover the child’s historical confirmation.
- Restart parent-chain checks also rely on local metadata. Lowering unmined retention preserves parents rather than deleting the old unmined records.

A merge, reset acknowledgement, empty template, or matching chain tip alone is therefore insufficient recovery evidence.

## Proposed deliverable

Add a dedicated `teranode-cli` recovery command, or an explicit repair mode on `resetblockassembly`. Keep ordinary reset semantics unchanged. The interface should provide:

1. **Read-only discovery by default.** Scan the unmined store and assembly, including affected dependencies; do not hardcode incident TXIDs or inspect only one returned candidate. Produce a machine-readable repair manifest with reasons and evidence, plus a concise operator summary. Bound concurrency, expose progress, and support large stores.
2. **Canonical-chain verification.** Use a configurable trusted source to verify transaction identity, historical confirmation and current output state at an agreed chain tip. Do not treat empty local metadata or a missing UTXO response alone as proof that a transaction may be deleted. Abort/revalidate when the chain changes.
[...]

PR #1715, opened 2026-09-09
Add IDLE-only CLI recovery for persisted replays

oskarszoon:
Replayed confirmed transactions can survive reset/restart because dirty UTXO records remain and surviving parents lack replay markers. Add one IDLE-only `teranode-cli recoverreplayedtransactions --work-dir ...` job to inventory persisted state, prove safe repairs from local canonical archives, journal mutations, and verify the store.

- Audit is read-only by default. `--apply --maintenance` enables repairs; `--resume` deliberately resumes the same job. Persisted IDLE and an unchanged tip are checked throughout. Operators must stop/drain every shared-store writer and restart writers after repair.
- The census covers master/page records, absent-child marker references and affected dependencies beyond assembly membership. A marker on the record owning the spent output is sufficient, matching normal pruning.
- Repairs require authenticated inclusion and ordered spend proofs, plus `max(confirmation height, latest spend height) + retention <= tip`. Retention is the greater of global and effective UTXO retention, sealed into version 3 job/manifest artifacts and rechecked before apply. Unsafe or incomplete components remain untouched and reported.
- Required parent/page markers are conditionally written and read back before child deletion. Spend owners, native types, generations, expiry and retained archives are preserved. SQLite EXTRA commits journal intent before remote mutations; initialization is atomic and an empty interrupted initialization can resume.
- History uses at most two body passes, bounded transitive ancestry expansion, and batched header commits/guard checks. Resume can reuse completed, validated history covering the fresh census. Unfinished scans still restart; set an adequate explicit timeout. Mainnet runtime remains unmeasured.
- Diagnostics stay on stderr; stdout carries the JSON report. Repairable audit findings return exit 2. SIGINT/SIGTERM/SIGHUP cancel through the job context and attempt to persist the report.

No production service, protobuf or dependency changes in the feature diff. The only service file changed is the real assembly recovery test. No external SV Node RPC is required. The operator runbook covers writer isolation, retained evidence, post-repair validation, and the risk of nondurable Aerospike deletes resurrecting records after a cold restart.

Validation at `6e88b943e`:

- Core and CLI recovery suites pass with race detection, including real Aerospike single-job tests. Native recovery tests pass; the real assembly E2E passes three repeated runs for inline and paginated records.
- Regression tests cover retention boundaries/overflow and fresh apply checks, page-only markers, the missing-inclusion classification branch, ancestry pass/call bounds, sealed-history reuse, audit exits, signals, logging and journal initialization. Removing the inclusion guard through a temporary overlay makes its new regression fail.
- CLI build, affected-package vet/staticcheck and recovery/CLI lint pass without disabling linters. Changed native/assembly code passes lint against the existing baseline.
- Standalone gosec reports 12 findings in unchanged CLI/benchmark code, none in the recovery packages. CLI govulncheck reports 25 called vulnerabilities in existing dependencies and the local Go toolchain; dependency upgrades are outside this recovery change.
- The complete Makefile unit package set passed with `-race -tags testtxmetacache`, `SETTINGS_CONTEXT=test`, `-p4` and a 10-minute per-package timeout, without coverage instrumentation: 132 packages passed and 48 had no tests (180 total). At least 15 minutes after push, all 36 current CI checks passed and one was skipped. Automated review found no new blockers.

Limits: Aerospike only. IDLE/tip checks detect changes but are not a distributed maintenance lock. Missing archives, live outputs, finite-expiry parent guards, uncertain ownership or traversal limits can leave an incomplete result. The CLI verifies persisted repairs while IDLE; operators must then restart writers, resume, and validate fresh candidates and the next tip. This repairs the incident's persisted exposure, not arbitrary UTXO corruption.

Refs #1708. Latest upstream/main is merged with a merge commit. The release/v0.15 variant is maintained separately.


cmd/recoverreplayedtransactions/history.go
```diff
@@ -0,0 +1,1023 @@
...
+		}
+		keep := target > 0
+		if !transaction.IsCoinbase() {
+			for _, input := range transaction.Inputs {
+				parent := input.PreviousTxIDStr()
+				if target > 0 {
+					if _, e := tx.ExecContext(ctx, "INSERT OR IGNORE INTO targets VALUES(?)", parent); e != nil {
+						return e
+					}
+				}
```
icellan, line 405: [P1] Stop ancestor expansion at authenticated confirmed transactions

Every confirmed target adds its parents as new targets, causing `Build` to discard its index and rescan the entire history until ancestry is exhausted. When ancestor records are pruned, the retained-reader expansion cannot avoid this: a ten-transaction confirmed chain reproduced eleven complete scans. Long confirmed transaction chains therefore multiply the entire maintenance workload by ancestry depth. Stop recursive expansion at authenticated confirmed transactions rather than tracing their ancestry back to coinbase.

pkg/replayrecovery/manifest.go
```diff
@@ -0,0 +1,406 @@
...
+		if progress != nil {
+			progress(result)
+		}
+	}
+	// Uncertain descendants or affected ancestors block the connected affected
+	// component. Historical parents outside the scan are not treated as affected.
+	_, err = m.db.ExecContext(ctx, `WITH RECURSIVE blocked(id) AS (
+SELECT id FROM entries WHERE classification NOT IN ('fully-spent','confirmed')
+UNION SELECT e.parent FROM edges e JOIN inventory i ON i.txid=e.child AND i.master=1 WHERE e.child NOT IN (SELECT id FROM entries)
+UNION SELECT CASE WHEN e.child=b.id THEN e.parent ELSE e.child END FROM edges e JOIN blocked b ON e.child=b.id OR e.parent=b.id JOIN entries n ON n.id=CASE WHEN e.child=b.id THEN e.parent ELSE e.child END
```
icellan, line 345: [P2] Avoid blocking repairs through unaffected historical parents

If replay R spends one output of normally mined parent P, and normally mined sibling S spends another, neither P nor S needs to appear in `entries`. This UNION nevertheless seeds P as blocked because S is outside `entries`, and the next branch propagates that flag to R. The reproduced result is that a fully proven repair is skipped solely because of an unrelated healthy sibling. Restrict blocking seeds and propagation to affected transactions instead of introducing unaffected historical parents.

cmd/recoverreplayedtransactions/history.go
```diff
@@ -0,0 +1,1023 @@
...
+	}
+	digest, e := historyDigest(context.Background(), db)
+	if e != nil {
+		return nil, e
+	}
+	if historySealDigest(digest, o, c) != seal.Digest {
+		return nil, errors.NewProcessingError("history index integrity mismatch")
+	}
+	o.Guard = options.Guard
+	o.Unconfirmed = options.Unconfirmed
```
icellan, line 652: [P2] Reject resume when consensus-era settings have changed

On resume, the caller supplies the current `GenesisActivationHeight`, but only its guard and transaction reader are copied; classification silently uses the saved activation height. Reopening an index with a different activation height reproduced a bare OP_RETURN transaction remaining `fully-spent` although the current setting makes its output spendable. Store identity and pinned-tip checks do not detect this mismatch, so stale classification can authorize deletion. Compare consensus-sensitive saved options with the current configuration and reject mismatches before applying.

icellan (commented):
The patch suppresses safe repairs, unnecessarily repeats full-history scans, and accepts stale consensus settings on resume. Focused tests, including recovery-engine race tests, passed, but additional regression checks reproduced these defects. Broader verification was limited by sandbox restrictions and an interrupted suite.

icellan (approved):

pkg/replayrecovery/manifest.go
```diff
@@ -0,0 +1,407 @@
...
+			}
+		}
+		if present > 0 {
+			entry.Action = "delete-recreated"
+		} else {
+			entry.Action = "mark-absent"
+		}
+		if entry.Evidence.Classification == FullySpent {
+			tx, parseErr := bt.NewTxFromString(ev.RawTx)
+			if parseErr != nil || tx.TxID() != id || !validHash(ev.BlockHash) || ev.BlockHeight > tip.Height {
```
freemans13, line 303: This is the only height rule on a fully-spent repair, and it allows depth zero. The pruner waits `global_blockHeightRetention` after the last spend before deleting (`teranode.lua:988`) so a reorg can still find the record.

Failure scenario: writers stop before block B's transactions are set mined, so a transaction C in B has empty `blockIDs`. All of C's outputs are spent at or below the tip, or C has no spendable outputs, so it classifies FullySpent. Apply deletes C and writes `deletedChildren[C]` on its parent. B is reorged out and the heavier chain includes C again: re-spending the parent output with the same spending data hits `teranode.lua:391-402` and returns INVALID_SPEND, so the node cannot accept that chain.

A probe test in `pkg/replayrecovery` that sets the evidence height to `source.tip.Height` shows Apply deleting the record. Suggest requiring `max(ev.BlockHeight, every spend height) + retention <= tip.Height` here and in the Apply re-check (`engine.go:398-405`). The incident's records were pruned only after retention, so real repairs are unaffected.

stores/utxo/aerospike/recovery_census.go
```diff
@@ -0,0 +1,254 @@
...
+			raw, ok := value.([]byte)
+			if !ok || len(raw) != 68 {
+				continue
+			}
+			child, _ := chainhash.NewHash(raw[32:64])
+			offset := uint64(item.Page)*uint64(b.batchSize) + uint64(i) // #nosec G115 -- constructor requires positive uint32 batch size; range index is nonnegative.
+			if offset > math.MaxUint32 {
+				return errors.NewProcessingError("spend reference output overflow")
+			}
+			ref := rr.SpendReference{ParentKey: append([]byte(nil), item.Record.Key...), ParentTxID: item.TxID, Vout: uint32(offset), ChildTxID: child.String(), Vin: binary.LittleEndian.Uint32(raw[64:]), Marked: markers[child.String()] == true && masterMarkers[child.String()] == true}
```
freemans13: Requiring the marker on both the page and the master does not match production. The pruner writes `deletedChildren` only on the record that holds the spent output: `CalculateKeySource(parentTxID, input.PreviousTxOutIndex, s.utxoBatchSize)` in `pruner_service.go:1157`. The Lua spend check (`teranode.lua:391-394`) reads only that record, so the page marker alone already rejects a replay.

Failure scenario: parent P has 300 outputs (batch size 128). Child L spends vout 200 and is pruned correctly, so the marker lands on page 1 only. This returns `Marked=false`, and `inventory.go:109` turns L into a `mark-absent` entry. With history available, apply writes a master marker nothing needs. Without it, which is likely for old children, L is Unknown, `Complete` goes false and the run exits 2.

Reproduced on a real container with batch size 2: pruner-deleted child, page marker present, master marker absent, `SpendReferences` reported `marked=false`. Suggest `Marked: markers[child] == true`, the marker on the record holding the output. The test at `recovery_test.go:315` ("master marker is also required") encodes the stricter rule and would need to change with it.

cmd/recoverreplayedtransactions/history.go
```diff
@@ -0,0 +1,1030 @@
...
+		// Authenticate targets before expanding dependencies: confirmed records
+		// already have sufficient ancestry evidence in the validated chain.
+		if err = h.expandUnconfirmed(ctx); err != nil {
+			return err
+		}
+		var after int
+		if err = h.db.QueryRowContext(ctx, "SELECT count(*) FROM targets").Scan(&after); err != nil {
+			return err
+		}
+		if before == after {
```
freemans13, line 290: Each time `expandUnconfirmed` adds targets, the loop re-runs the whole genesis-to-tip body pass. The P1 fix stopped this for confirmed ancestry, but every generation of unconfirmed ancestry still costs a full pass: a probe with a six-deep unconfirmed chain on one confirmed transaction made seven passes. `unconfirmedTransaction` gives up at depth 256 anyway (`:969`), so passes beyond that buy nothing.

Per height, the loop makes roughly six RPCs (`checkTip` calls `Guard.Check`, which does `GetState`, `GetBestBlockHeader`, `GetState`, then `checkTip` adds its own `GetBestBlockHeader`; `historyClient` adds two more) and a `synchronous=FULL` commit. On mainnet that is millions of RPCs per pass against a 30m default timeout, and `run.go:302-311` deletes `history.sqlite` on resume, so a timeout restarts from genesis with writers still stopped. I counted calls; I did not time a run.

Suggest expanding ancestry transitively before a single body pass, and checking the guard every N blocks (the watchdog already runs every 5s).

pkg/replayrecovery/manifest.go
```diff
@@ -0,0 +1,407 @@
...
+		}
+		if inputErr != nil || findings > 0 {
+			entry.Evidence.Classification = Unknown
+			entry.Evidence.Reason = "incomplete local inputs or unsafe inventory"
+		}
+		if e = m.db.QueryRow("SELECT COUNT(*) FROM inventory WHERE txid=? AND (candidate=1 OR reason!='')", id).Scan(&candidates); e != nil {
+			return result, e
+		}
+		if present > 0 && candidates == 0 && findings == 0 && inputErr == nil {
+			tx, parseErr := bt.NewTxFromString(ev.RawTx)
```
freemans13, line 284: No test reaches this branch. Forcing it to always classify Confirmed passes the full suite, including the Aerospike and e2e tests. A Confirmed entry does not seed the blocked set (`:344`), so a wrong Confirmed here could let a fully-spent ancestor in the same component be repaired. `TestApplyBoundaryUnknownDescendantThroughConfirmedEntryBlocksReplay` looks like it covers this, but its unresolved descendant blocks through the "child not in entries" arm of the SQL instead. A variant where the descendant is an entry whose evidence has no `BlockHash` would pin it.

cmd/teranodecli/teranodecli/cli.go
```diff
@@ -158,6 +162,9 @@ func Start(args []string, version, commit string) {
 	tSettings := settings.NewSettings()
 
 	logger := ulogger.InitLogger("teranode-cli", tSettings)
+	if command == "recoverreplayedtransactions" {
+		logger = ulogger.New("teranode-cli", ulogger.WithLevel(tSettings.LogLevel), ulogger.WithWriter(os.Stderr))
```
freemans13, line 176: The stderr writer is ignored when pretty logging is on, which is the default: `prettyZeroLogger` hard-codes `Out: os.Stdout` (`ulogger/zerologger.go:72-78`). Running the built binary put `INFO ... Using default resolver` and a blockchain client connection error on stdout, so `teranode-cli recoverreplayedtransactions ... | jq` breaks. The runbook says the JSON summary goes to stdout. `report.json` still works.

pkg/replayrecovery/journal.go
```diff
@@ -0,0 +1,209 @@
...
+	defer func() {
+		if err != nil {
+			_ = db.Close()
+		}
+	}()
+	if err = db.Ping(); err != nil {
+		return nil, err
+	}
+	if !readonly {
+		if _, err = db.Exec("PRAGMA journal_mode=DELETE; PRAGMA synchronous=FULL; PRAGMA foreign_keys=ON"); err != nil {
```
freemans13: In `journal_mode=DELETE`, a commit becomes durable when the rollback journal file is deleted. `synchronous=FULL` does not sync the directory after that delete; `EXTRA` does. After a power cut right after committing a delete intent, the rollback journal can reappear and SQLite rolls the intent back while the Aerospike delete has already landed. Resume then fails in `inventoryMatches` with "record disappeared without this job's exact delete intent". That fails closed, but the job cannot resume. `PRAGMA synchronous=EXTRA` closes it; `history.go:128` has the same setting.

stores/utxo/aerospike/recovery.go
```diff
@@ -0,0 +1,776 @@
...
+	if err = ctx.Err(); err != nil {
+		return err
+	}
+	key, err := b.key(before.Key)
+	if err != nil {
+		return err
+	}
+	p := recoveryPolicy(ctx, before.Generation)
+	p.GenerationPolicy = as.EXPECT_GEN_EQUAL
+	existed, aerr := b.client.Delete(p, key)
```
freemans13, line 248: Not a durable delete. On a namespace with persistent storage, a cold restart (always the case on Community Edition) can bring back a record deleted without one. Here the returned record would be the recreated unmined child, which is exactly what this tool removes. The pruner has the same gap, so this is not new exposure, but it costs more here. Either set `DurableDelete: true` on this policy or note the risk in the runbook's restart step.

freemans13 (changes requested):
Requesting changes for two correctness issues. Both have small fixes. The rest of the tool is careful work: I found no path that deletes a child record before every parent marker is written with a generation guard and read back, and header, subtree and block Merkle authentication all hold up.

**What needs to change before merge**

1. **Deletes and markers ignore the reorg window.** The only height rule on a repair is `BlockHeight <= tip.Height`. If a node is stopped before a recent block's transactions are marked mined (mid-catchup, say), a transaction confirmed in the tip block can pass as a fully-spent recreated record. The tool then deletes it and writes `deletedChildren` on its parent. If that block is reorged out, the heavier chain carries the same transaction again, and the Lua spend check rejects it as a replay. The fix costs the real incident nothing, because every record the pruner lost was already deeper than `global_blockHeightRetention`. Require `max(confirmation height, every spend height) + retention <= tip`, at Discover and again at Apply. Inline on `manifest.go`.
2. **Correctly pruned children of large parents are reported as unmarked.** The census counts a spend as marked only when both the output's page record and the master record carry the marker. The production pruner writes it only on the page that holds the output, and Lua reads only that record, so the page marker alone already rejects a replay. Every child pruned normally from a parent with more than `utxoBatchSize` (128) outputs becomes a `mark-absent` entry. Where history survives, that writes a needless master marker. Where it does not, which is the usual case for old children, the entry is Unknown and the run ends in exit 2 on a store that has nothing wrong with it. Reproduced on a real Aerospike container: pruner-deleted child, page marker set, master marker absent, `SpendReferences` returns `marked=false`. Inline on `recovery_census.go`.

**Worth fixing, not blocking**

3. **The history scan will not finish on mainnet in the default window, and a timeout throws it away.** `Build` walks every header from tip to genesis, then every block body from genesis to tip. Each height costs roughly six blockchain RPCs, three of them from `Guard.Check`, plus a `synchronous=FULL` SQLite commit. The whole body pass repeats once per generation of unconfirmed ancestry: a six-deep unconfirmed chain made seven full passes in a probe test. The default `--timeout` is 30m, and `--resume` before apply deletes `history.sqlite` and restarts from height 0, all while every writer is stopped. At about 950k heights that is millions of RPCs per pass; I counted calls, I did not time a run. Expanding ancestry transitively in one pass (capped at the existing 256 depth) and checking the guard every N blocks rather than every block would help most.
4. **An untested classification branch.** Removing the "mined record lacks canonical inclusion" guard in `manifest.go` fails no test. Inline.
5. Smaller items inline: logs land on stdout next to the JSON summary; `synchronous=FULL` in DELETE journal mode can roll back the last journal commit on power loss; deletes are not durable deletes.

Also minor, not inlined: a read-only audit that finds repairable records exits 0, which operators will read as "store is clean". A crash between creating `journal.sqlite` and writing its header leaves `--resume` permanently refused. SIGTERM and SIGHUP are not handled, so an ssh drop ends the run without `report.json`, though the journal stays consistent.

**On the earlier review.** icellan's three findings check out as follows. The Genesis-activation resume check is complete. The sibling fix is correct, though the mirror case still blocks: a normally mined entry with any child in the store seeds the blocked set, which fails closed. The ancestor-expansion fix stops at confirmed transactions but not unconfirmed ones (item 3). The Claude bot's summary still asks for sign-off on `BlockAssembler.go` hot-path changes. Those changes are no longer in the net diff: the only `services/` file touched is `replay_recovery_e2e_test.go`.

**What I ran, at 2ea4f8999.** With `-race`: `pkg/replayrecovery`, `cmd/recoverreplayedtransactions` (including `TestRunJobAerospike`), the seven `TestRecovery*` tests in `stores/utxo/aerospike`, and both `TestReplayRecoveryAerospikeAssemblyEndToEnd` variants, all on real Aerospike containers. All pass. I then removed 17 safety guards one at a time: 12 were caught. Besides item 4, the survivors were the `EXPECT_GEN_EQUAL` policies on Mark and Delete (a read-and-compare runs just before each, so only a writer that ignores maintenance inside that gap would get through), the Lookup re-authentication in `history.go:566-571` (the index is already sealed), and the backend readback in `Mark` (the engine re-reads the marker before each delete). The e2e test asserts end state properly, with a positive control first.

**What I could not check.** I did not run the job against a live node or a mainnet-sized store, so the runtime in item 3 is from call counting. The reorg consequence in item 1 is traced to the Lua check; I did not follow `moveBackBlock` end to end, and the probe for it runs on the fake backend. Replay rejection after repair is proven for `SpendAndCreate` through Lua, not through the native Teraspike spend operation.

pkg/replayrecovery/manifest.go
oskarszoon: Fixed in 6e88b943e. Discovery and the fresh pre-apply evidence check now require `max(confirmation height, latest authenticated spend height) + retention <= tip`, using widened arithmetic. The CLI takes the greater of global and effective UTXO retention, rejects zero, and seals the policy into version 3 job/manifest artifacts so resume cannot change it. Recent recreated records and absent-child markers stay Unknown. Tests cover recent confirmation/spends, the exact boundary, overflow, and changed evidence/policy with zero writes.

stores/utxo/aerospike/recovery_census.go
oskarszoon: Fixed in 6e88b943e: SpendReferences uses the marker on the record that owns the output. The extra master lookup is gone. The real Aerospike regression now leaves the page marker present and master marker absent, verifies the census produces no repair entry without needing history, and verifies Lua still rejects the replay.

cmd/recoverreplayedtransactions/history.go
oskarszoon: Addressed in 6e88b943e. After the initial authentication pass, unresolved ancestry expands transitively with depth 256 and shared 10,000-node/input-edge limits; at most one additional body pass authenticates the new targets. Guard checks and header commits are batched at 128 heights, with phase-boundary checks and the existing watchdog retained. The six-generation regression now makes two body passes; a separate test checks guard-call reduction.

Resume reuses a completed, validated history index when it covers the fresh census targets. An interrupted, unsealed scan still restarts: this change does not add a durable partial-scan checkpoint protocol. The runbook states that limitation and requires an adequate explicit timeout; mainnet runtime remains unmeasured. Every accepted inclusion and spend still needs authenticated chain evidence.

pkg/replayrecovery/manifest.go
oskarszoon: Added `TestMinedDescendantEntryWithoutInclusionBlocksRepair` in 6e88b943e. The normally mined descendant is an actual entry with valid raw transaction data but no BlockHash. The test asserts Unknown, a blocked ancestor, and zero writes. I also removed this inclusion guard through a temporary Go overlay: the new test fails with that mutation and passes with the guard restored.

cmd/teranodecli/teranodecli/cli.go
oskarszoon: Fixed in 6e88b943e. Recovery configures its stderr logger before initializing any logger, and disables both pretty and dual JSON logging for this CLI process because both bypass WithWriter. Subprocess tests cover config and environment overrides and assert stdout remains free of diagnostics. No shared logger or service implementation changes.

pkg/replayrecovery/journal.go
oskarszoon: Fixed in 6e88b943e: journal and history use synchronous=EXTRA with DELETE mode. Journal schema and identity header are also initialized atomically. Resume can recover an empty initialization left by a crash, while refusing nonempty headerless state, foreign schemas, and mismatched identities. Regression tests cover the pragma and initialization cases.

stores/utxo/aerospike/recovery.go
oskarszoon: Took the runbook option in 6e88b943e. It now distinguishes restarting application writers from an Aerospike cold restart, explains that nondurable deletes can resurrect old record versions on persistent storage, and requires a fresh audit with writers isolated after such a restart. It also says to retain the journal/backups and that a completed journal does not certify resurrected records. Durable deletes cannot be enabled unconditionally because Community Edition rejects that policy; the native delete policy is unchanged.

oskarszoon:
@freemans13 Pushed `6e88b943e` after merging upstream/main with a merge commit. Replied to and resolved all seven inline threads with fixes and regression evidence. The additional review-body items are covered too: repairable audits return exit 2, journal initialization is atomic/resumable, and SIGTERM/SIGHUP cancel through the job context.

The mirror dependency case remains deliberately conservative and can still block a component. Interrupted unsealed history scans still restart; completed sealed coverage can be reused, and the runbook now states the timeout/mainnet-runtime limitation. The 0.15 variant has the same production fixes and passes its recovery/CLI/native/assembly race tests. At least 15 minutes after the push, all 36 current CI checks passed and one was skipped. Automated review found no new blockers; the complete local unit/race run also passed (132 tested packages, 48 without tests).

One correction to the automated summary: `TestMinedDescendantEntryWithoutInclusionBlocksRepair` is already in this commit in `pkg/replayrecovery/engine_boundary_test.go`, and fails when the inclusion guard is removed through a temporary overlay. The corresponding thread has been replied to and resolved. The fixes are ready for your re-review.

cmd/recoverreplayedtransactions/history.go
```diff
@@ -0,0 +1,1158 @@
...
+		}
+	}()
+	if _, err := txn.ExecContext(ctx, `CREATE TEMP TABLE pending_targets(id TEXT PRIMARY KEY,depth INTEGER,expanded INTEGER DEFAULT 0);
+CREATE INDEX pending_expansion ON pending_targets(expanded,depth,id);
+INSERT INTO pending_targets(id,depth) SELECT id,0 FROM targets WHERE NOT EXISTS(SELECT 1 FROM transactions WHERE txid=targets.id)`); err != nil {
+		return err
+	}
+	defer func() { _, _ = txn.ExecContext(context.Background(), "DROP TABLE pending_targets") }()
+	edges := 0
+	for work := 0; work < maxWork; work++ {
```
freemans13, line 866: This budget is global, and it counts root pops as well as expansions. Every unconfirmed census target not in `transactions` starts in `pending_targets` at depth 0, and each one consumes a `work` unit. Every input of every root consumes one of the shared 10,000 `edges`. Once a store has more than about 10,000 unresolved roots (or about 5,000 two-input ones), the rest are never expanded. Their parents never become targets, so `Check` returns Unknown with "unconfirmed dependency outside proven absence scope", which does not mention the budget.

Probe: 10,050 legitimate two-level unconfirmed chains came out as 10,000 unconfirmed and 50 Unknown at 6e88b943e, and as 10,050 unconfirmed at 789d30a89. It fails closed, and confirmed recreated records are not roots, so repairs outside those components still go through. But a mainnet store routinely holds far more than 10,000 unmined transactions. There, every run ends at exit 2 with a pile of Unknown legitimate transactions, and an operator cannot tell recovery from failure. `coverage.Complete` also stays true when the budget trips.

Suggest giving each root its own budget, as `unconfirmedTransaction` does, or scaling the cap with the root count. Recording a coverage flag and reason when a limit trips would make the report honest.

services/blockassembly/replay_recovery_e2e_test.go
```diff
@@ -0,0 +1,359 @@
...
+	_, err = rr.Inventory(ctx, nativeBackend, manifest, guard)
+	require.NoError(t, err)
+	rawBackend := rr.NewEvidenceBackend(nativeBackend, nil, tip, func(ctx context.Context, id string) (*bt.Tx, error) {
+		h, e := chainhash.NewHashFromStr(id)
+		if e != nil {
+			return nil, e
+		}
+		return store.GetTxFromExternalStore(ctx, *h)
+	})
+	source, err := command.NewHistory(ctx, filepath.Join(recoveryDir, "history.db"), chainStore, archive, command.HistoryOptions{Guard: guard, TargetsPath: manifest, EndHeight: tip.Height, GenesisActivationHeight: settings.ChainCfgParams.GenesisActivationHeight, Unconfirmed: rawBackend.Transaction})
```
freemans13, line 283: `HistoryOptions` has no `BlockHeightRetention` here, so the history runs with retention 0, which `Run` refuses (`run.go:116`). The same holds for `runJob` in `job_test.go`. The result is that no test on real Aerospike exercises a repair through the real history `Check` with the retention rule active. Only the fake source in `pkg/replayrecovery` does. Setting retention to 1 here fails both subtests, because the fixture tip is height 2 and the repair lands inside the window. Raising the fixture tip above retention and running with a positive value would close it.

cmd/recoverreplayedtransactions/run.go
```diff
@@ -0,0 +1,469 @@
...
+		return err
+	}
+	ctx, timeoutCancel := context.WithTimeout(ctx, o.Timeout)
+	defer timeoutCancel()
+	ctx, cancel := context.WithCancelCause(ctx)
+	defer cancel(nil)
+	if s == nil || s.UtxoStore.UtxoStore == nil || s.UtxoStore.UtxoStore.Scheme != "aerospike" {
+		return commandError("replay recovery supports only the Aerospike UTXO backend")
+	}
+	o.blockHeightRetention = max(s.GlobalBlockHeightRetention, s.GetUtxoStoreBlockHeightRetention())
```
freemans13, line 115: Changing this to `o.blockHeightRetention = s.GlobalBlockHeightRetention` passes every test. That would under-protect any node with a positive `utxostore_blockHeightRetentionAdjustment`, the setting `GetUtxoStoreBlockHeightRetention` adds on top. A small test with a positive adjustment would pin the `max`.

cmd/recoverreplayedtransactions/run.go
```diff
@@ -0,0 +1,469 @@
...
+			if reuseErr != nil {
+				return nil, combineErrors(reuseErr, history.Close())
+			}
+			if reuse {
+				return history, nil
+			}
+			if err = history.Close(); err != nil {
+				return nil, err
+			}
+		} else if !errors.Is(err, os.ErrNotExist) && !errors.Is(err, sql.ErrNoRows) {
```
freemans13, line 417: Only `ErrNotExist` and `ErrNoRows` lead to a rebuild. `NewHistory` creates the file and applies the schema in separate steps (`history.go:97-136`). A cancel landing between the end of Inventory and the schema step leaves a 0-byte `history.sqlite` plus `.lock`: a signal, a watchdog cancel or SIGKILL. After that every `--resume` fails with "no such table: coverage". I reproduced it with a context cancelled just before the schema step. Before 6e88b943e resume deleted the file, so this is new. It is safe, because nothing has been applied, but the only way out is a fresh work dir. Treating a missing `coverage` table as unsealed would fix it.

docs/topics/commands/recoverReplayedTransactions.md
```diff
@@ -0,0 +1,183 @@
...
+retained archive blobs. The durable journal records intent before remote writes
+and verified outcomes afterward. Shared parents and dependency order are tracked.
+
+## Interrupted work and results
+
+Keep writers stopped on failure and retain the entire work directory. Explicit
+resume reconciles only this job's exact expected before/after states. Interrupted
+read-only phases can be rebuilt; a partial scan is never treated as complete.
+A completed, sealed history index is reused only after a fresh census confirms
+that all its targets are covered; newly discovered targets require rebuilding.
```
freemans13, line 135: The direction is reversed. `ReuseTargets` accepts when every fresh census target is already a target in the index, so the census must be covered by the index, not the index's targets by the census. Suggest: "reused only when the fresh census finds no target missing from it".

On line 76, "a shared limit of 10,000 processed records/input edges" is two separate counters (`work` and `edges`), each capped at 10,000.

freemans13 (approved):
Approving at 6e88b943e. Both blockers from my earlier review are fixed correctly, and every other item holds up when I break it on purpose. One thing the fix commit introduced deserves a follow-up before anyone runs this on a busy mainnet store: the new shared expansion budget (inline on `history.go:866`). It fails closed, so it is not a safety problem, but it can hold a healthy run at exit 2.

**What I verified, item by item**

- **Reorg window.** `outsideRetention` checks `max(confirmation, last spend) + retention <= tip` in uint64. It runs at Discover for both actions and again per entry in Apply, after `requireTip`. `LastSpendHeight` is the maximum over every spendable output, including same-block spends. The retention used is `max(global, utxo store retention)`, both halves are loaded in `settings/settings.go`, and zero is refused. The policy is sealed into the job, manifest and history. Removing the guard at Discover only, at Apply only, dropping the spend height, overwriting instead of taking the max, and dropping the uint64 widening each fail a test.
- **Page marker.** The census now reads the marker on the record that owns the output, which matches the pruner key (`pruner_service.go:1157`) and the Lua check. Apply still marks master and page, and Verify still requires both, so the two stay consistent.
- **Confirmed branch.** Forcing Confirmed at `manifest.go:288` now fails `TestMinedDescendantEntryWithoutInclusionBlocksRepair`.
- **History cost.** At most two body passes. The tip is still read on every block lookup, and the FSM guard runs every 128 heights, after the last partial batch, after expansion and before the seal. A second pass clears transactions, spends, blocks and gaps first. Resume reuses a sealed index only when the tip matches, retention and Genesis activation match, and every fresh census target is already an index target. Making `ReuseTargets` always accept fails two tests.
- **Journal.** EXTRA on both databases. Header and schema are committed together. Rows without a header, a foreign schema and mismatched identity are refused, and only a provably empty file is initialised.
- **Stdout.** I built the binary and ran it with `jsonLogging=true PRETTY_LOGS=true logLevel=DEBUG` and no reachable node: 0 bytes on stdout, diagnostics on stderr.
- **Signals and exit codes.** SIGTERM and SIGHUP cancel with exit 1 on the binary. A repairable audit returns exit 2 through `auditResult`, and a clean audit returns 0.

**Follow-ups, none blocking**

1. The expansion budget counts roots, so stores with more than 10,000 unresolved unmined roots leave the rest Unknown. Inline.
2. No test on real Aerospike runs a repair with the retention rule active. Inline on the e2e test.
3. The adjustment half of the retention choice is untested: `Run` using only `GlobalBlockHeightRetention` passes CI. Inline on `run.go:115`.
4. A 0-byte `history.sqlite` left by a cancel between Inventory and the schema step makes every `--resume` fail. Inline.
5. Two doc sentences say the opposite of the code, or blur two limits. Inline.
6. Minor, not inlined. A sealed index that has gaps is reused on resume, so restoring missing archives and resuming cannot clear those gaps; a fresh work dir can. Verify does not re-check retention or the sealed policy, which matters only to direct library callers, since the CLI refuses a changed policy on resume. `flag.ExitOnError` exits 2 on a mistyped flag, which collides with the documented meaning of exit 2. That last one predates the fix commit.

**What I ran, at 6e88b943e.** With `-race`: `pkg/replayrecovery`, `cmd/recoverreplayedtransactions`, `cmd/teranodecli/teranodecli`, `stores/utxo/aerospike` and both `TestReplayRecoveryAerospikeAssemblyEndToEnd` variants on real Aerospike containers. All pass. I also ran the mutation checks listed above in a scratch worktree.

**What I could not check.** I have no mainnet-scale runtime and no real census distribution, so how often item 1 bites is unmeasured. SIGKILL and power loss were simulated with file copies and a cancelled context, not reproduced. Replay rejection after repair is still proven only through Lua, not the native Teraspike spend operation.

cmd/recoverreplayedtransactions/history.go
oskarszoon: Deferred as a nonblocking follow-up, with priority before a busy-mainnet recovery run. The current global record and input-edge budgets count initial roots, so they can leave healthy transactions Unknown and produce exit 2. Follow-up scope: per-root or root-count-scaled bounds, explicit expansion-limit reporting separate from archive coverage, and regressions above 10,000 roots and with multiple inputs. Preserve bounded work and the two-pass proof checks. This remains unfixed in the approved revision and the 0.15 variant; resolving this thread records the deferral.

services/blockassembly/replay_recovery_e2e_test.go
oskarszoon: Deferred as a nonblocking test follow-up. Verified that both the assembly and single-job fixtures currently use zero retention. Advance their authenticated history/tip, set positive retention, and exercise both a repair outside the window and a recent protected transaction through real history and Aerospike, including pagination. Carry the fixtures into the 0.15 variant too. This coverage gap remains in the approved revision; resolving as deferred.

cmd/recoverreplayedtransactions/run.go
oskarszoon: Deferred as a nonblocking regression-test follow-up. The current implementation includes the effective UTXO adjustment, but the tests do not pin that configuration path. Add a positive-adjustment case that fails when Run uses global retention alone, and retain a case for the global floor. Apply the test to main and 0.15. No policy change is included here; resolving as deferred.

cmd/recoverreplayedtransactions/run.go
oskarszoon: Deferred as a nonblocking resume fix. The missing coverage table currently reaches the refusal path, so a zero-byte history file left before schema initialization needs a fresh work directory. Follow-up: recognize valid unsealed initialization only before apply, rebuild safely, and add cancellation/zero-byte-file regressions while preserving refusal of corrupt seals and post-mutation rebuilds. Do not delete an existing mutation journal to work around this. The limitation remains in main and the 0.15 variant; resolving as deferred.

docs/topics/commands/recoverReplayedTransactions.md
oskarszoon: Deferred to the operator-documentation follow-up. The precise reuse condition is: every fresh census target must already be in the sealed history index. The current expansion limits are two separate global counters, each capped at 10,000: processed records and input edges. Update that wording together with any budget change. Also document that restored archives need a fresh job to rebuild a sealed index with gaps, and that malformed flags can exit 2 through flag.ExitOnError. The current runbook remains unchanged; resolving as deferred.

PR #1716, opened 2026-09-10
fix(asset): reject subtree-root BUMP proof requests

### LINKED ISSUE #1500: [HARDEN-0805] Subtree-root SPV proofs are still unverifiable: no BUMP leaf node, and a block path over placeholder subtree roots
## What happens

The asset server's `GET /merkle_proof/{hash}` falls back to a *subtree* proof when the
queried hash turns out to be a subtree root rather than a transaction
(`services/asset/httpimpl/GetMerkleProof.go`). That proof is built by
`ConstructSubtreeMerkleProof` and converted to BUMP by `bump.ConvertToBUMP`. The BUMP it
produces cannot be verified by any BRC-74 client, for two independent reasons.

PR #1482 fixed the offsets on this path — they were previously derived from a `-1`
sentinel folded to `0xFFFFFFFF`, so every level carried garbage. The offsets are correct
now. The proof is still unusable, and this issue covers what is left.

### 1. There is no leaf node to start from

`ConstructSubtreeMerkleProof` sets `TxID: chainhash.Hash{}` (`util/merkleproof/merkle_proof.go:459`,
commented "Empty for subtree proofs"). `ConvertToBUMP` only emits the level-0 leaf node —
the BRC-74 flag-`0x02` entry naming the thing being proven — when `proof.TxID` is
non-zero (`util/bump/format.go:159`). So the subtree-proof BUMP has siblings at every
level and nothing to fold them against.

Confirmed end to end against the go-bc reference implementation: build the exact
structure `ConstructSubtreeMerkleProof` returns (zero `TxID`, `TxIndexInSubtree: -1`,
`SubtreeIndex: 3`, two block-proof levels), run it through `ConvertToBUMP` →
`EncodeBinary` → `bc.NewBUMPFromBytes` → `CalculateRootGivenTxid`. The offsets come out
`2` and `0`, which are correct, and go-bc then fails with:

```
the BUMP does not contain the txid: 0100...00aa
```

The obvious fix is to populate the leaf from `proof.SubtreeRoot`, which
`ConstructSubtreeMerkleProof` already sets. That needs a decision first — see the open
question below.

### 2. The block path is built over placeholder subtree roots

`ConstructSubtreeMerkleProof` calls `GenerateBlockMerkleProof(block.Subtrees, subtreeIdx)`
on the raw stored subtree roots (`util/merkleproof/merkle_proof.go:439`). Subtree 0's
stored root is computed with the *coinbase placeholder* in slot 0, not the real coinbase
txid. Reconciling to the block header's merkle root requires substituting the real
coinbase and recomputing subtree 0's root — which `ConstructMerkleProof` does
(`merkle_proof.go:188-193`) and this function does not.

So even once a leaf node exists, the fold will not reach the header's merkle root for any
block whose proof path includes subtree 0's root — which is essentially all of them, since
subtree 0's root is a sibling somewhere on nearly every path. This is the same defect
class already documented in `util/bump/production_repro_test.go`.

## Why it matters

This is a silent-wrong-answer failure, not a crash. The endpoint returns HTTP 200 with a
well-formed BUMP; the failure only surfaces inside whichever third-party SPV client tries
to verify it. Nothing on the node notices.

Severity is bounded by how much this path is used — it is a fallback for subtree-root
queries, not the transaction path, which is correct and covered by
`coinbase_placeholder_crosscheck_test.go`.

## Open question before fixing

BRC-74 defines BUMP leaves as transaction IDs. A subtree root is not a txid, so a "BUMP
proving a subtree root" is arguably not a valid BUMP at all, and a client that accepted
one would be treating an interior node as a leaf. Three options:

1. Populate the leaf from `SubtreeRoot` and fix the placeholder substitution, accepting
   that the result is a BRC-74-shaped proof of a non-txid leaf.
2. Stop serving BUMP for subtree-root queries and return a different representation, or a
   clear error, rather than a proof no client can check.
3. Expand the subtree proof into a real transaction-level proof.

Option 2 is the most defensible if nothing currently consumes this path; option 1 is the
smallest change. Worth confirming who actually calls it before picking.

## Repro

Add to `util/bump/`:

```go
proof := &merkleproof.MerkleProof{
    TxID:             chainhash.Hash{},   // as ConstructSubtreeMerkleProof sets it
    BlockHeight:      800000,
    SubtreeIndex:     3,
    TxIndexInSubtree: -1,                 // subtree-proof sentinel
    SubtreeRoot:      subtreeRoot,
    SubtreeProof:     []chainhash.Hash{},
    BlockProof:       []chainhash.Hash{sib, sib2},
}
b, _ := bump.ConvertToBUMP(proof)
bin, _ := b.EncodeBinary()
parsed, _ := bc.NewBUMPFromBytes(bin)
_, err := parsed.CalculateRootGivenTxid(subtreeRoot.String())
// err: the BUMP does not contain the txid
```

Found while reviewing #1482.

galt-tr:
Subtree-root requests to `/merkle_proof/{hash}` previously returned HTTP 200 with an unverifiable BUMP: the proof omitted its starting leaf and used stored subtree roots containing the coinbase placeholder. BRC-74 leaves are transaction IDs, so the endpoint now returns a clear 404 when no mined transaction exists for the hash. `ConvertToBUMP` also rejects subtree-only proofs.

This implements option 2 from #1500. Transaction-proof offset arithmetic remains unchanged; the earlier 64-bit fix already addressed it. Endpoint documentation and regression tests cover rejection in JSON, hex, and binary modes. The new regressions were observed failing before the fix.

Breaking Go API change:
- `merkleproof.ConstructSubtreeMerkleProof` is removed. It is an exported function in a non-internal package, so downstream Go consumers that call it will not compile after upgrading.
- It is removed rather than deprecated because every value it returned was unverifiable. It hardcoded `TxID` to the zero hash, which makes `VerifyMerkleProof` fold from a zero leaf and fail its own `SubtreeRoot` comparison every time; and it built `BlockProof` from raw stored subtree roots with no coinbase-placeholder substitution while setting `MerkleRoot` from the block header, so the returned struct asserted inclusion under a root its own proof path does not reach. Retaining it behind a deprecation notice would preserve exactly the defect this PR fixes.
- `merkleproof.GenerateBlockMerkleProof` is unchanged and still exported. A consumer that needs the sibling path of a subtree within a block uses it directly and performs the coinbase substitution explicitly.
- The module is pre-1.0 (`v0.16.0-beta-12`), where the public API is not covered by a stability guarantee.

Follow-up, not addressed here: `FindBlocksContainingSubtree` is now unused inside `util/merkleproof` but remains a required method of the exported `MerkleProofConstructor` interface.

Validation:
- Tests and race tests passed for `./util/bump`, `./util/merkleproof`, and `./services/asset/httpimpl` with `testtxmetacache`, including existing go-bc transaction-proof cross-checks.
- `go vet`, `staticcheck`, and changed-code `golangci-lint` and `gosec` checks passed for these packages.
- Independent code review found no issues.
- `govulncheck` reported 12 vulnerabilities in existing dependencies and the Go standard library; dependency versions were not changed.

Closes #1500.


services/asset/httpimpl/GetMerkleProof.go
```diff
@@ -169,35 +167,19 @@ func (h *HTTP) GetMerkleProof(mode ReadMode) func(c echo.Context) error {
 		// Create adapter to use merkleproof helper functions
 		adapter := newMerkleProofAdapter(ctx, h.repository, h.mainChainCache)
 
-		// Try to construct merkle proof - first as transaction, then as subtree
-		var proof *merkleproof.MerkleProof
-
-		// First, try as transaction hash
-		proof, err = merkleproof.ConstructMerkleProof(txHash, adapter)
+		// BRC-74 proofs start at transaction leaves, never at interior subtree roots.
```
freemans13, line 170: The dashboard still calls this endpoint with a subtree root: `subtree-merkle-visualizer/index.svelte:128` calls `api.getMerkleProof(subtreeHash)`. With the fallback gone, that tab now shows this 404 message as an error. The review body has the two ways to clear it.

freemans13 (changes requested):
**Requesting changes for one reason: the dashboard still calls the path this PR removes.**

The backend change is right. BRC-74 leaves are transaction IDs, the old fallback returned an HTTP 200 that no client could verify, and the new guard in `ConvertToBUMP` stops that proof shape coming back. The new test does tell the fix apart from its absence: I put the old `GetMerkleProof.go` from `99b8579b4` back under it and `TestGetMerkleProof_SubtreeHash_ReturnsNotFound` failed in all three modes. `util/bump`, `util/merkleproof` and `services/asset/httpimpl` pass locally with `-tags testtxmetacache` at `ea995eae6`.

### The consumer

Issue 1500 recommends option 2 "if nothing currently consumes this path". One thing does. The subtree page in the dashboard has a Merkle proof tab, and its visualizer calls `api.getMerkleProof(subtreeHash)`, which requests `/merkle_proof/{subtreeHash}/json` (`ui/dashboard/src/internal/components/page/viewer/subtree/subtree-merkle-visualizer/index.svelte:128`, mounted from `viewer/subtree/index.svelte:121`).

In fairness to this PR, that tab does not work on main either. It reads `subtreeIndex`, `blockProof`, `subtreeRoot` and `merkleRoot`, and the BUMP JSON response carries none of them, so today it draws one node with an empty hash. After this PR it shows "Error: transaction not found; BUMP proofs require a transaction ID". Below that is the tab's own hint, "Make sure the subtree hash is valid and exists in the blockchain", which is wrong for a subtree the page has just loaded.

Two ways to clear it:

1. Remove the tab and the visualizer in this PR. Take `merkleproof.ConstructSubtreeMerkleProof` (`util/merkleproof/merkle_proof.go:393`) with them, since the Claude bot already noted it has no production caller left. Also drop the optional subtree fields on `MerkleProofData` in `ui/dashboard/src/internal/api/index.ts`, whose comment says they are "populated for subtree-level merkle proofs".
2. Keep the tab, but rebuild it from `GET /subtree/{hash}` and the block's subtree list so it stops calling an endpoint that now accepts transaction IDs only. That is a bigger job and could be a follow-up issue, provided the tab is hidden until then.

Option 1 fits better, because the endpoint now says plainly that subtree proofs are unsupported.

### CI

`chainintegrity` failed in `TestChainIntegrity3Nodes`: block validation during catchup hit "failed to get expected work required ... context canceled". Nothing in this diff touches that path, and the same job also failed on the unrelated branch `security/quick-body-preflight-v0.15`, so I read it as a flake. A re-run should clear it.

### Not checked

I did not build or open the dashboard. What the tab shows before and after comes from reading the Svelte code, not from a browser.

freemans13 (approved):
**Approving. The dashboard consumer I blocked on is gone, and the new transaction ID guard is safe for the coinbase path.** I read the branch at `0e59a435b`.

### What I asked for

`c81cf8415` removes the subtree page's Merkle proof tab, its visualizer, and the subtree-only fields on `MerkleProofData`. It also deletes `ConstructSubtreeMerkleProof`. An old `?tab=merkleproof` link now falls back to Overview. The new smoke test checks that no `/merkle_proof/` request goes out, and the `smoke` job passed on this head. The tx page's proof tab still calls the endpoint with a real txid, which is the one caller that should.

### The zero transaction ID guard

`ConvertToBUMP` now rejects a zero `TxID` and always emits the level-0 txid node. Its other production route is `ComputeCoinbaseBUMP`, called from `services/blockvalidation/BlockValidation.go:3327` and `services/blockassembly/Server.go:2099`. That route takes the txid from `subtree0.Nodes[0].Hash`. That is either the real coinbase hash or the coinbase placeholder, which is all `0xff`. Neither one is zero, so the guard cannot reject a real coinbase BUMP. Emitting the txid node every time changes nothing for this caller, because its txid was never zero.

### On the Copilot threads

- **Removing `ConstructSubtreeMerkleProof` as a breaking change:** I asked for this deletion. Nothing in the repo calls it now, and `ConvertToBUMP` rejects the only proof shape it builds, so a deprecated copy would just be dead code. I did not check whether any repo outside teranode imports `util/merkleproof`.
- **Swagger still says "Block, transaction, or subtree hash":** that text comes from the shared `hashPathParam` in `swagger_types.go`, which about 40 operations use, `getTransaction` among them, and it was there before this PR. The three merkle proof route summaries already say "for a transaction". A proof-specific parameter would be tidier, but I would not hold the PR for it.

### Worth a line in the description

Two commits are not about issue 1500:

- `0e59a435b` fixes two blockvalidation tests that are broken on main. `c3db82e72` made cached-subtree validation read the conflicting-nodes trailer, and the fixtures were still written with `SerializeNodes()`. I ran both tests on `c3db82e72` and both fail with `unable to read number of leaves: EOF`. They pass on this branch.
- `2f14c0887` replaces `Test_queue`'s sleeps with the package's `fixedClock`.

Both are fine to carry here. The description still covers only the endpoint change, though. Could it also mention the dashboard removal, the deleted constructor and these two test fixes, so a later `git blame` makes sense?

Nit: the i18n key `viewer-subtree.details.tab.merkleproof` (`ui/dashboard/src/internal/i18n/common/en.json:275`) no longer has a user. Keep `DetailTab.merkleproof`, because the tx viewer still uses it.

### What I ran and what I did not

- I ran `util/bump`, `util/merkleproof` and `services/asset/httpimpl` locally with `-tags testtxmetacache`, plus the two blockvalidation tests and `Test_queue` three times. All passed.
- When I posted this, `test`, the `sequential` shards and the `smoketest` shards were still pending on this head. `chainintegrity`, `smoke` and `golangci-lint` had passed.
- I did not build the dashboard or open it in a browser. The evidence for the UI side is the Svelte diff and the smoke test.

services/asset/httpimpl/GetMerkleProof.go
freemans13: Confirmed on `0e59a435b`: the visualizer and the tab are gone, `?tab=merkleproof` falls back to Overview, and the smoke test asserting no `/merkle_proof/` request passed in CI. Approved.

util/bump/format.go
```diff
@@ -81,20 +81,23 @@ func hashToDisplayHex(h chainhash.Hash) string {
...
-// BRC-74 verifier cannot fold it. Two things are missing, and callers should not
-// expect either from this function. The proof leaves TxID as the zero hash, so no
-// level-0 leaf node is emitted and go-bc's CalculateRootGivenTxid reports that the
-// BUMP does not contain the txid; and its block path is built over raw stored subtree
-// roots with no coinbase-placeholder substitution, so it would not reconcile to the
-// header's merkle root even with a leaf present. See issue 1500.
+// Only transaction proofs can be converted: BRC-74 leaves are transaction IDs.
+// Subtree-only proofs are rejected because an interior subtree root cannot be
+// represented as a transaction leaf, and its stored root may contain the coinbase
+// placeholder rather than the real coinbase hash.
```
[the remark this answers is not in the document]
galt-tr: Taking the second half of this (document the removal), not the first (retain it). Reasoning below, because the distinction matters.

`ConstructSubtreeMerkleProof` was not a working API being withdrawn for convenience. Every value it returned was unverifiable, including by this package's own verifier. I ran it on `main` against a block whose stored first subtree carries the coinbase placeholder, which is how subtrees are actually stored:

```
TxID             = 0000...0000  (zero)
TxIndexInSubtree = -1
SubtreeProof len = 0
VerifyMerkleProof  -> ok=false
folded BlockProof  = 18d97c5688eaacd9f6eed97e767ffaf8dfc56af3c1f7be8aaf13fa489f353c2f
proof.MerkleRoot   = f6c8bc20cf2d5e20d80d295f51da65b582db8b4e717cab596ce070d2d098315b
reconciles = false
```

Two structural defects, not incidental ones:

- `VerifyMerkleProof` begins folding from `proof.TxID`, which this constructor hardcodes to the zero hash, then compares the result against `SubtreeRoot`. That can never match, so the package's own verifier returned false for every proof the function produced.
- The function built `BlockProof` from raw `block.Subtrees` with no coinbase-placeholder substitution, while setting `MerkleRoot` from the block header. `ConstructMerkleProof` performs that substitution deliberately. The returned struct therefore asserted inclusion under a merkle root its own proof path does not reach.

Retaining it behind a deprecation notice would preserve a function whose return value states something false, which is the defect #1500 is about. A downstream that compiled against it was already receiving a struct that no verifier in this package, and no BRC-74 verifier via BUMP, accepts. The compile break surfaces a breakage that already existed at runtime.

The primitive underneath it stays exported. `GenerateBlockMerkleProof` is unchanged by this PR, so a consumer that genuinely wants the sibling path of a subtree within a block still has it, and now has to perform the coinbase substitution consciously instead of receiving a struct that silently misreports. That is the honest version of the capability.

On versioning: the module is pre-1.0, currently tagged `v0.16.0-beta-12`, where semver and the Go module documentation both treat the public API as unstable, and the repo carries no CHANGELOG or API-stability policy to version this against. Reaching the removed function also required implementing `MerkleProofConstructor`, a six-method interface over `*model.Block`, `*model.BlockHeader`, `*subtree.Subtree` and internal block IDs, so the realistic external surface is small.

Documented the removal explicitly in the PR description, per your second option.

One genuinely useful thing this comment turns up. With the function gone, `FindBlocksContainingSubtree` is now called zero times inside `util/merkleproof`, yet remains a required method of the exported `MerkleProofConstructor` interface, so any implementer still writes a method nothing calls. I have left it alone to keep this PR scoped to its fix, but it is the real remaining downstream-facing wart and is worth a follow-up.

oskarszoon (approved):
Reviewed at 3628ee4 against merge-base fd94c83 (15 files, +210/-1225) with five reviewer passes — protocol, Go correctness, security, QA and frontend — plus an adversarial verification stage over every finding. No P0 or P1. Approving.

The change is protocol-correct, and not only for the reason the description gives. A subtree root is an interior node of the block merkle tree with no BRC-74 leaf representation, so the old 200 returned something no conformant verifier could fold (go-bc's `CalculateRootGivenTxid` scans `Path[0]` for a 0x02-flagged txid and the old response had no such node). Separately, the stored subtree-0 root is computed over the coinbase placeholder 0xFF×32 while the header root uses the real coinbase txid, so the deleted `ConstructSubtreeMerkleProof`'s block path could not have folded to the header root even with a leaf present. A 404 is strictly better than a 200 that looks like a proof.

Verified rather than taken on trust: reverting GetMerkleProof.go, merkle_proof.go and util/bump/format.go to the merge-base makes the new `TestGetMerkleProof_SubtreeHash_ReturnsNotFound` fail with HTTP 200 on all three response modes, and it passes at HEAD. The offset arithmetic in `ConvertToBUMP` is byte-identical apart from the now-unreachable `txIndex == -1` remap; `globalOffset`, the sibling XOR, the 64-level cap and both segment fit guards are untouched. Every error-return site in `ConstructMerkleProof` maps to the same status as before except the ones the PR intends to change. The `ConstructSubtreeMerkleProof` removal has zero remaining references anywhere in the tree and the module is pre-1.0. Deleting the dashboard visualizer is necessity, not scope creep — its only data source was `api.getMerkleProof(subtreeHash)`, so it would have been permanently broken once the endpoint 404s.

Two follow-ups, neither blocking:

- `util/bump/format.go:97` — the new guard rejects the zero TxID, but Teranode's sentinel that actually occurs is the coinbase placeholder 0xFF×32. `coinbase_bump.go:43` passes `subtree0.Nodes[0].Hash` through verbatim and `services/blockvalidation/BlockValidation.go:3308` substitutes the real coinbase txid only `if block.CoinbaseTx != nil`. On that branch a block is stored with a CoinbaseBUMP whose level-0 "txid" is ffff…ffff, which folds to something that is not the header merkle root — #1500's failure mode surviving on the block-data side. Pre-existing; worth a symmetric one-line guard.
- `util/bump/format.go:168` — with the guard now `len(bump.Path) > 0`, a zero-level proof (single-transaction block) still emits an empty path and no level-0 txid node. It encodes to the 2 bytes `0700`, go-bc refuses to parse it, and `Validate` turns it into a 500. BRC-74's form for that block is treeHeight 1 with just the txid node. Pre-existing and fails closed, but it is the one remaining shape where Teranode hands out an unfoldable BUMP, and nothing covers the zero-level success path in tests.

The `services/blockassembly/subtreeprocessor/queue_test.go` rewrite is a sound flake fix (wall-clock sleeps → the existing `fixedClock` seam, 20× `-race` reruns deterministic) but is unrelated to this PR's purpose and isn't mentioned in the description. Test-only, so bundling it is a judgment call rather than a problem — either split it or name it in the description.

PR #1717, opened 2026-09-11
p2p: carry received counters over GetPeerRegistry

gokutheengineer:
**Closes bitcoin-sv/teranode#4760.**

### Problem
`p2p_api.PeerRegistryInfo` carried none of the interaction-type counters (`blocks_received`, `subtrees_received`, `transactions_received`, `catchup_blocks`), so `peerInfoToP2PProto` and `convertFromAPIPeerInfo` silently dropped them even though `blockchain.PeerInfo` populates them and `p2p.PeerInfo` documents them as part of the public contract. The asset service's `peerTierCache.refresh` classifies `tierMiner` only when `BlocksReceived > 0`, and the gRPC client is the only `ClientI` implementation, so in a real microservice deployment no allowlisted miner was ever promoted and all were capped at the ordinary per-peer rate limit.

All cited locations confirmed still valid on `upstream/main`.

### Fix
- Add fields 30-32 (`blocks_received`, `subtrees_received`, `transactions_received`) to `PeerRegistryInfo`, mirroring the names already used in `blockchain_api.proto`, and regenerate `p2p_api.pb.go`. `catchup_blocks` is deliberately not added: nothing in the repo writes `CatchupBlocks`, so a wire field would be permanently zero.
- Populate them in `peerInfoToP2PProto` (Server.go) and map them back in `convertFromAPIPeerInfo` (Client.go).
- Update the generated proto docs table (`docs/references/protobuf_docs/p2pProto.md`), which was also missing the existing `catchup_*` rows, and correct the `GetPeerRegistry` doc in `Interface.go`, which promised "complete information" while several fields never cross the wire.

No change to the `tierMiner` predicate: the field is the right signal, it was just never transmitted.

Rolling upgrade: a new asset service against an old p2p service still reads `BlocksReceived == 0`, so miners stay at `tierPeer` until p2p is upgraded. Fails closed (stricter limit), same as today.

### Residual
Pre-existing, not introduced here: `convertFromAPIPeerInfo` maps a `0` unix timestamp to `time.Unix(0, 0)`, which is not `IsZero()`. No gRPC consumer relies on `IsZero()` today (`sync_coordinator` uses the in-process registry client). Worth a separate follow-up.

### Tests
- `TestServer_PeerInfoToP2PProto_RoundTripFields`: now asserts the three counters.
- `TestServer_GetPeerRegistry_ReceivedCountersSurviveWire` (new): registry `RecordBlockReceived`/`RecordSubtreeReceived`/`RecordTransactionReceived` -> `Server.GetPeerRegistry` -> `proto.Marshal`/`Unmarshal` -> `convertFromAPIPeerInfo`, asserting the counters survive the wire message.
- `TestServer_PeerRegistryInfo_CarriesEveryPeerInfoField` (new): reflection guard that every exported `p2p.PeerInfo` field has a `PeerRegistryInfo` counterpart unless listed in an explicit, documented not-on-wire allowlist. This turns the class of bug behind this issue into a test failure.
- `TestSimpleClientGetPeerRegistry`: gRPC client maps the three proto fields onto `p2p.PeerInfo`.
- `TestPeerTierCache_Refresh_ClassifiesByBlocksReceived` (new, asset): `refresh` promotes only `BlocksReceived > 0` with reputation at or above the threshold to `tierMiner`.

```
SETTINGS_CONTEXT=test go test -race -tags "testtxmetacache" -count=1 ./services/p2p/...   # ok
go vet -tags testtxmetacache ./services/asset/httpimpl/ ./services/p2p/                    # ok
```

`services/asset/httpimpl` tests could not be executed locally (pre-existing GoBDK cgo link failure on this machine); they type-check and run in CI.

### DIFF
--- a/services/p2p/p2p_api/p2p_api.proto
+++ b/services/p2p/p2p_api/p2p_api.proto
@@ -281,6 +281,14 @@ message AddBanScoreRequest {
     int64 catchup_attempts = 27;
     int64 catchup_successes = 28;
     int64 catchup_failures = 29;
+
+    // Interaction type breakdown, incremented by RecordBlockReceived,
+    // RecordSubtreeReceived and RecordTransactionReceived on the blockchain peer
+    // registry. blocks_received drives the asset service's miner-tier
+    // rate-limit exemption, so it must survive the gRPC hop.
+    int64 blocks_received = 30;
+    int64 subtrees_received = 31;
+    int64 transactions_received = 32;
   }
 
   message GetPeerRegistryResponse {
--- a/docs/references/protobuf_docs/p2pProto.md
+++ b/docs/references/protobuf_docs/p2pProto.md
@@ -374,6 +374,12 @@ Represents comprehensive peer information with all registry metadata.
 | client_name | [string](#string) |  | Human-readable name of the client software |
 | last_catchup_error | [string](#string) |  | Last error message from catchup attempt |
 | last_catchup_error_time | [int64](#int64) |  | Time of last catchup error (Unix timestamp) |
+| catchup_attempts | [int64](#int64) |  | Number of catchup attempts with this peer |
+| catchup_successes | [int64](#int64) |  | Number of successful catchup operations |
+| catchup_failures | [int64](#int64) |  | Number of failed catchup operations |
+| blocks_received | [int64](#int64) |  | Number of blocks received from this peer |
+| subtrees_received | [int64](#int64) |  | Number of subtrees received from this peer |
+| transactions_received | [int64](#int64) |  | Number of transactions received from this peer |
 
 <a name="RecordBytesDownloadedRequest"></a>
 
--- a/services/asset/httpimpl/peer_auth_middleware_test.go
+++ b/services/asset/httpimpl/peer_auth_middleware_test.go
@@ -2,6 +2,7 @@ package httpimpl
 
 import (
 	"bytes"
+	"context"
 	"crypto/rand"
 	"crypto/sha256"
 	"encoding/hex"
@@ -12,6 +13,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/bsv-blockchain/teranode/services/p2p"
 	"github.com/bsv-blockchain/teranode/ulogger"
 	"github.com/bsv-blockchain/teranode/util"
 	"github.com/labstack/echo/v4"
@@ -372,6 +374,54 @@ func TestPeerAuthMiddleware_AllowlistMember_GetsTier(t *testing.T) {
 
 	require.Equal(t, http.StatusOK, rec.Code)
 	require.Equal(t, tierMiner, *captured)
+}
+
+// registryStubP2PClient is a p2p.ClientI that only answers GetPeerRegistry.
+type registryStubP2PClient struct {
+	p2p.ClientI
+	peers []*p2p.PeerInfo
+}
+
+func (c *registryStubP2PClient) GetPeerRegistry(_ context.Context) ([]*p2p.PeerInfo, error) {
+	return c.peers, nil
+}
+
+// TestPeerTierCache_Refresh_ClassifiesByBlocksReceived — refresh's
+// classification predicate: tierMiner requires BlocksReceived > 0 together with
+// ReputationScore >= threshold (inclusive); every other registered peer is
+// tierPeer and an unknown peer stays tierUnverified. The guard that
+// BlocksReceived actually survives the p2p gRPC hop lives in
+// services/p2p/server_handler_test.go (TestServer_GetPeerRegistry_ReceivedCountersSurviveWire).
+func TestPeerTierCache_Refresh_ClassifiesByBlocksReceived(t *testing.T) {
+	newPeerID := func() peer.ID {
+		k, _, err := crypto.GenerateEd25519Key(rand.Reader)
+		require.NoError(t, err)
+		id, err := peer.IDFromPublicKey(k.GetPublic())
+		require.NoError(t, err)
+		return id
+	}
+
+	miner := newPeerID()
+	atThreshold := newPeerID()
+	noBlocks := newPeerID()
+	lowRep := newPeerID()
+	unknown := newPeerID()
+
+	client := &registryStubP2PClient{peers: []*p2p.PeerInfo{
+		{ID: miner, BlocksReceived: 1, ReputationScore: 90},
+		{ID: atThreshold, BlocksReceived: 1, ReputationScore: 50},
+		{ID: noBlocks, BlocksReceived: 0, ReputationScore: 100},
+		{ID: lowRep, BlocksReceived: 50, ReputationScore: 10},
+	}}
+
+	cache := newPeerTierCache(ulogger.TestLogger{}, client, 50)
+	cache.refresh(context.Background())
+
+	require.Equal(t, tierMiner, cache.GetTier(miner))
+	require.Equal(t, tierMiner, cache.GetTier(atThreshold), "reputation exactly at threshold qualifies")
+	require.Equal(t, tierPeer, cache.GetTier(noBlocks), "no blocks received must not be a miner")
+	require.Equal(t, tierPeer, cache.GetTier(lowRep), "reputation below threshold must not be a miner")
+	require.Equal(t, tierUnverified, cache.GetTier(unknown))
 }
 
 // TestParsePeerAuthAllowlist — parsing of the pipe-separated config value.
--- a/services/p2p/Client.go
+++ b/services/p2p/Client.go
@@ -877,6 +877,9 @@ func convertFromAPIPeerInfo(apiPeer interface{}) (*PeerInfo, error) {
 			CatchupAttempts:        p.CatchupAttempts,
 			CatchupSuccesses:       p.CatchupSuccesses,
 			CatchupFailures:        p.CatchupFailures,
+			BlocksReceived:         p.BlocksReceived,
+			SubtreesReceived:       p.SubtreesReceived,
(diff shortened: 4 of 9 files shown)


ctnguyen (approved):
Review at commit [ed85310](https://github.com/gokutheengineer/teranode/commit/ed853107b8d4ebc3cc3a133f1b1c50496277c9e9)

This is a tidy fix for a genuinely dead code path. I confirmed the premise in the tree: `ReportValidBlock` really does write the counter (`services/p2p/handle_catchup_metrics.go:399` calls `RecordBlockReceived`), `peerTierCache.refresh` really does gate `tierMiner` on `BlocksReceived > 0` (`services/asset/httpimpl/peer_auth_middleware.go:116-127`), and before this PR the value had no wire field to travel on — so the exemption was unreachable over gRPC. Adding the field rather than weakening the predicate is the right call. I also checked the regenerated `p2p_api.pb.go` byte-for-byte rather than taking it on trust: the `rawDesc` message length goes `\xe7\b` (1127) to `\xf2\t` (1266), a delta of 139, which is exactly the 41 + 45 + 53 bytes the three new `FieldDescriptorProto` entries occupy; the tag bytes `\x1e`/`\x1f`/`\x20`, the `\x03` int64 type and the `R`-prefixed json names all line up. The descriptor is therefore byte-consistent with what `protoc-gen-go` emits for these three fields, so a reader can take the generated file at face value instead of re-deriving it. The `Interface.go` doc correction is a nice touch — replacing a "complete information" promise that was never true with an explicit list of what does not cross the wire is more useful than the fix itself in the long run. Two non-blocking notes below, both about durability rather than correctness of this change.

## Non-blocking issues

### ChiR1 — The new reflection guard checks that a wire field exists, not that it is mapped

**Problem:** `TestServer_PeerRegistryInfo_CarriesEveryPeerInfoField` (`services/p2p/server_handler_test.go:117-158`) only asserts that a `p2p_api.PeerRegistryInfo` struct field *named* like each exported `p2p.PeerInfo` field exists. It never exercises `peerInfoToP2PProto` or `convertFromAPIPeerInfo`. The bug this PR fixes had two halves — no proto field, and no converter mapping — and the guard closes only the first. Add `Foo int64` to `p2p.PeerInfo`, add `foo = 33` to the proto, forget both converters, and the guard still passes while every gRPC consumer reads zero: the identical failure mode. The converter half is currently covered only by the hand-maintained assertion list in `TestServer_PeerInfoToP2PProto_RoundTripFields`, which asserts 18 of the 30 fields `peerInfoToP2PProto` sets, omitting twelve (`BytesReceived`, `LastBlockTime`, `LastMessageTime`, the three `Interaction*` counters, the three `LastInteraction*` timestamps, `ReputationScore`, `MaliciousCount`, `LastCatchupErrorTime`) and, despite its name, never runs the client-side converter at all. Separately, `wireFields` is built from *all* struct fields including the unexported `state`, `unknownFields` and `sizeCache`, so a future `p2p.PeerInfo.State` would satisfy the guard against the protobuf bookkeeping field rather than against a real wire field.

**Why it matters:** The PR body presents this test as turning "the class of bug behind this issue into a test failure". As written it catches the schema half only, so the regression it is meant to prevent can still land silently. I verified both converters do map all 30 wire fields today (`services/p2p/Server.go:3115-3146`, `services/p2p/Client.go:851-882`), so this is about the guard's future value, not a live defect.

**Fix:** Skip unexported wire fields, and make the guard value-based so it covers the converters end to end:

```go
for i := 0; i < wt.NumField(); i++ {
    if !wt.Field(i).IsExported() {
        continue
    }
    wireFields[strings.ToLower(wt.Field(i).Name)] = true
}

// ... after the existing name check, assert the value actually survives both hops.
src := &blockchain.PeerInfo{ /* every non-allowlisted field set to a distinct non-zero value */ }
got, err := convertFromAPIPeerInfo(peerInfoToP2PProto(src))
require.NoError(t, err)

gv := reflect.ValueOf(*got)
for i := 0; i < dt.NumField(); i++ {
    f := dt.Field(i)
    if !f.IsExported() {
        continue
    }
    if _, skip := notOnWire[f.Name]; skip {
        continue
    }
    require.False(t, gv.Field(i).IsZero(),
        "p2p.PeerInfo.%s is zero after peerInfoToP2PProto/convertFromAPIPeerInfo; add the mapping or list it in notOnWire", f.Name)
}
```

That also retires the hand-maintained assertion list, since the new check subsumes it.

### ChiR2 — Retired field numbers 12 and 13 are still not reserved

**Problem:** `PeerRegistryInfo` jumps from `last_message_time = 11` to `interaction_attempts = 14` (`services/p2p/p2p_api/p2p_api.proto:262-265`) with no `reserved` declaration. `git log -L` on the message shows the gap is not decorative: `bool url_responsive = 12` and `int64 last_url_check = 13` existed and were deleted in d07f1548a ("Remove URLResponsive checks, all handled with reputation score"). Nothing marks them as spent, so a later change can reuse 12 or 13 for an unrelated type and a mixed-version reader will misparse the varint.

**Why it matters:** The repo already treats this as a convention — `services/validator/validator_api/validator_api.proto:72-73`, `services/blockassembly/blockassembly_api/blockassembly_api.proto:158-159` and `services/blockchain/blockchain_api/blockchain_api.proto:660-661` all carry `reserved` plus the retired name, and this very file spells out the reasoning in prose at line 336 for an RPC name that protobuf cannot reserve. This PR is the natural place to close it: it is already doing field-number housekeeping in exactly this message, and appending at 30 is otherwise correct.

**Fix:** Two lines in the message body:

```proto
  // Removed in "Remove URLResponsive checks": url_responsive (bool) and
  // last_url_check (int64). Do not reuse - a mixed-version reader would
  // misinterpret a new field at these numbers.
  reserved 12, 13;
  reserved "url_responsive", "last_url_check";
```

Then `make gen` to refresh `p2p_api.pb.go`.

## Recap

| --- | --- | --- | --- |
| ChiR1 | Guard checks names not mappings |  | 35% |
| ChiR2 | Field numbers 12 and 13 unreserved |  | 25% |

services/p2p/p2p_api/p2p_api.proto
```diff
@@ -281,6 +281,14 @@ message AddBanScoreRequest {
     int64 catchup_attempts = 27;
     int64 catchup_successes = 28;
     int64 catchup_failures = 29;
+
+    // Interaction type breakdown, incremented by RecordBlockReceived,
+    // RecordSubtreeReceived and RecordTransactionReceived on the blockchain peer
+    // registry. blocks_received drives the asset service's miner-tier
+    // rate-limit exemption, so it must survive the gRPC hop.
+    int64 blocks_received = 30;
```
freemans13, line 289: This comment and the PR body say `blocks_received` drives the miner-tier exemption, so carrying it over the wire makes allowlisted miners reach `tierMiner`. On a node that is already synced, that mostly still won't happen. The counter has one production writer: `RecordBlockReceived`, reached only through `ReportValidBlock` from blockvalidation's `reportValidBlockForPeers`. That is called only from `catchup.go:1590`.

A miner at the tip announces a block whose parent we already hold. `processBlockFound` (`services/blockvalidation/Server.go:1542`) sees the parent exists, skips catchup, and goes straight to `ValidateBlockWithOptions`. Nothing on that route credits the peer. So `BlocksReceived` goes above zero only for peers this node once caught up from. A miner the node has followed block by block since startup stays `tierPeer`. `sync_coordinator.go:278` documents the same writer.

Two smaller consequences of that writer:
- `reportValidBlockForPeers` also credits every secondary peer that contributed subtree data. So "miner" here really means "an allowlisted peer we caught up from". The allowlist bounds this, but the setting's longdesc describes it as a mining-peer test.
- `TestServer_GetPeerRegistry_ReceivedCountersSurviveWire` calls `reg.RecordBlockReceived` directly, so no test drives the production writer.

Suggestion: keep this PR as the transport fix, but reword the comment and PR body so they don't claim the exemption now works. Then open a follow-up to credit the delivering peer on the `processBlockFound` success path. That follow-up needs its own look, since `ReportValidBlock` also moves reputation. I don't think it belongs in this diff.

freemans13 (approved):
Approving. Reviewed at `ed853107b`. The wire fix is correct and needed: `blocks_received` now crosses the p2p gRPC hop, and both converters map it. It does not yet make the miner-tier exemption fire on a synced node, though, because the counter's only production writer is catchup. Details are in the inline comment. I'd treat that as a follow-up issue rather than a blocker, since nothing gets worse with this merged.

What I checked:
- The chain end to end. The blockchain service's gRPC already carries the three counters (`services/blockchain/peer_registry_grpc.go:226-228`, `:297-299`). So the p2p server's registry holds real values, and this PR closes the last hop to the asset service.
- Every production reader of the three counters, repo-wide. The only one reached over the p2p gRPC client is `peerTierCache.refresh`. `sync_coordinator` reads the blockchain registry directly, and the RPC `getpeerinfo` handler calls `GetPeerRegistry` but never reads these fields. So turning the values on changes nothing else.
- Every production writer. There is exactly one route: blockvalidation's `reportValidBlockForPeers`, called only from `catchup.go:1590`.
- ctnguyen's two notes, the reflection guard and the unreserved field numbers 12 and 13. I agree with both and have not repeated them.
- The SonarQube gate failure is not this PR's doing. The two `go:S5332` hits are `services/p2p/Server.go:2513` and `:2531`, both unchanged since fdcde79adc in July 2025.

Verification: I ran the five new or changed tests locally with `-race` (four in `services/p2p`, plus `TestPeerTierCache_Refresh_ClassifiesByBlocksReceived` in `services/asset/httpimpl`), and all pass. That covers the asset test the PR body said could not run on the author's machine. I did not mutation-test the converters, and I did not run this on a live multi-service node.

PR #1718, opened 2026-09-11
MvP-1234 Follow-up robustness and cleanup items from the group-completion refactor in #1232

### LINKED ISSUE #1234: Follow-up: non-blocking robustness/cleanup items from #1232 (group-completion refactor)
Follow-up items from the review of #1232 (group-completion protocol for batched UTXO ops). None block that PR — it was approved — but each is a small robustness/cleanup improvement worth doing separately so they don't get lost.

### 1. Propagation client: unbounded wait + blind trust in response length

`services/propagation/Client.go`

`ProcessTransaction` waits with `group.Wait(context.Background(), 0)` — no timeout and no ctx arm — and `handleBatchResponse` completes only the items whose index appears in `response.Errors`:

```go
func (c *Client) handleBatchResponse(batch []*batchItem, response *propagation_api.ProcessTransactionBatchResponse) {
	for i, err := range response.Errors {
		if !err.IsNil() {
			batch[i].complete(err)
		} else {
			batch[i].complete(nil)
		}
	}
}
```

Today this is safe because the server always allocates `Errors: make([]*errors.TError, len(req.Items))` (`Server.go`), so the lengths match and every item completes. But the client relies entirely on that server contract: if a response ever came back with fewer entries than the batch (protocol drift, a proxy, a future server change), the uncovered items would never complete and their submitters would park forever on the unbounded wait — the exact leak class the Aerospike side already guards against with `batcherWaitTimeout`.

**Suggested fix:** bound the wait (mirror `batcherWaitTimeout`) and/or have `handleBatchResponse` complete any batch item not covered by the response, so the client is self-healing rather than contract-dependent.

### 2. Misleading "timed out" wording on shutdown cancel

`stores/utxo/aerospike/spend.go:1282` (`IncrementSpentRecords`) and the sibling `SetDAHForChildRecords`

These now `group.Wait(s.ctx, spendTimeout)`. On graceful shutdown `s.ctx` cancels and the caller gets `ServiceUnavailableError("... batch operation timed out after 30s ...")` even though no timeout actually elapsed (and, per the code's own comment, the enqueued write still drains on Close). Distinguish the ctx-cancel case from a real timeout so the error message isn't misleading and doesn't drive spurious retry/rollback logic.

### 3. Redundant `toEnqueue` copy on the common path

`stores/utxo/aerospike/spend.go:395`

When `spendCircuitBreaker == nil` (the normal case), `toEnqueue` is an element-for-element duplicate of `items` that `PutBatchCtx` then snapshots again — an extra N-pointer alloc + copy per tx (N large on consolidation txs) for no benefit. Pass `items` directly when the breaker is absent, and only build a filtered slice in the rare fast-fail branch.

### 4. Panic-sweep duplication + stale doc example

`stores/utxo/aerospike/batch_completion.go:30` and `services/{blockassembly,propagation,validator}/Client.go`

- The three service clients each hand-roll a "recover → complete every item" panic sweep that duplicates the `signalBatchPanic[T]` generic used by the seven Aerospike dispatch fns; the client copies also omit the stack trace + metric the generic logs. Promote `signalBatchPanic` to shared code and call it from all four.
- The `signalBatchPanic` godoc example still shows the removed `util.SafeSend(it.done, …)` channel pattern that no caller uses anymore — update it to the `it.complete(err)` form so it doesn't mislead the next author.

---
_Filed as follow-up to the review of #1232._

ctnguyen:
This resolve https://github.com/bsv-blockchain/teranode/issues/1234

## What was done

**Bounded the batch hand-off wait** (`services/propagation/Client.go`,
`services/validator/Client.go`). Both waited on the completion group with a background context
and no timeout, so a dispatcher that never signalled parked the submitting goroutine for the
life of the process. The wait is now cancellable by the caller's context and bounded by a
finite backstop. The context arm alone is not sufficient: the callers of this path carry no
guaranteed deadline, so shutdown would otherwise be the only release. The backstop is sized to
outlast the deepest downstream wait these clients front — a guard shorter than the legitimate
budget aborts work the lower layers would still have completed.

An early return is abandonment rather than cancellation, so the result slot is not read on that
path; the dispatcher still owns it and may write it later. The abandonment error is a service
error, so it cannot be mistaken for a queue-full shed, which a caller would unwind by deleting
the record and unspending the inputs. This matches the existing block assembly client.

**A cancelled store context is no longer reported as a timeout**
(`stores/utxo/aerospike/spend.go`). `IncrementSpentRecords` and `SetDAHForChildRecords` wait on
the store's context as well as a timeout but reported either outcome as "timed out after ...",
and the former also bumped the `BatchTimeout` counter. On an orderly shutdown that is false on
both counts, and it puts a `ServiceUnavailable` — which retry layers key on — in front of a
cancellation. Both now classify the error the wait actually returned, and the verdict is never
re-derived from the context afterwards: both arms of the wait can be ready at once, and a
recheck would relabel a real timeout whenever a shutdown happened to overlap.

**Dropped a redundant copy on the spend hot path** (`stores/utxo/aerospike/spend.go`). `Spend`
built a second slice of every item to hand to the batcher, which snapshots the slice itself —
one extra N-pointer allocation and copy per transaction, N being the input count. It now passes
the existing slice and splits off a filtered copy only once the circuit breaker has actually
rejected an input.

**Shared the batcher panic sweep** (`util/batch_panic.go`). Block assembly, propagation and
validator each hand-rolled the "recover, then complete every item" guard, and all three copies
omitted the stack trace and the panic metric the store's generic already logged. The generic
moved to `util`, alongside `DrainBatcher`, and all four callers use it; the store keeps a thin
wrapper for its package-private counter, so its six call sites are unchanged. Both the recovered
value and the stack are now quoted, keeping the record on a single line. The error text is
unchanged.

## Already present in the code

Two items the issue lists were already satisfied, so they carry no diff here:

- **The batch response-length guard.** `handleBatchResponse` in the propagation client and
  `processBatchResponse` in the validator client already check the response length against the
  batch length and fail every item with a clear error instead of stranding the uncovered ones.
- **The panic-sweep godoc example.** The usage example on the helper already showed the current
  `complete(err)` form and matched its live call sites. The doc was rewritten as part of moving
  the helper to `util`, which is a separate change; the example needed no correction. The
  opening roster of call sites was dropped rather than corrected, since that hand-maintained
  list is what goes stale.


ordishs (approved):
Checked the four changes against the code and against go-batcher/v2@v2.0.6. No P0 or P1
findings.

The three risks this change could have introduced do not occur. `Batcher.PutCtx` never
inspects `ctx.Done()` and never drops an item (`batcher.go:323-328`, `404-422`), so a
cancelled caller cannot strand one. `PutBatch`/`PutBatchCtx` snapshot the caller slice
(`batcher.go:369-398`), so aliasing `items` is safe, and nothing writes through the slice
after the loop. The completion group is fully decremented on every path, including
all-rejected, partially rejected, rejected send, and `len(spends) == 0`. The destructive
unwind in `Validator.go:1343-1380` is gated strictly on `ErrThresholdExceeded`, and
`ERR_SERVICE_ERROR` → `codes.Internal` vs `ERR_THRESHOLD_EXCEEDED` → `codes.ResourceExhausted`
keeps the two apart across gRPC.

Mutation testing confirms the tests pin the change: reverting the timeout argument to `0`
fails only the two backstop tests, reverting the ctx argument fails only the two
caller-cancel tests, and injecting `_ = item.result` into either abort branch produces a
`-race` report in 5 of 5 runs.

One nit worth fixing before merge:

- `stores/utxo/aerospike/spend.go:1388` and `stores/utxo/aerospike/spend.go:1161` — both new
  cancellation messages carry two `%s` verbs but supply one parameter. `errors.New`
  (`errors/errors.go:356-371`) extracts the trailing error as the wrapped error and removes
  it from the format parameters, so the last verb starves. On shutdown the message renders
  as:

  ```
  CONTEXT_CANCELED (7): [IncrementSpentRecords][<txid>] store context cancelled while waiting for batch response: %!s(MISSING) -> UNKNOWN (0): context canceled
  ```

  Diagnostic only — no caller keys on the text. Drop the trailing `: %s` and let the wrap
  chain render the cause, or pass `waitErr.Error()` as a plain string, as
  `recoverBatcherShutdown` does at `aerospike.go:156`. The same store documents this trap at
  `aerospike.go:141-144`.

Two smaller test-coverage gaps are in my notes rather than here: the `if filtered { append }`
branch at `spend.go:450-452` executes zero times under the new tests, and the "no
`s.ctx.Err()` recheck" policy is guarded by a function signature rather than by a test
(a stub context whose `Done()` never closes and whose `Err()` returns `context.Canceled`
would pin it deterministically).

services/validator/Client.go
```diff
@@ -495,14 +533,9 @@ func (c *Client) sendBatchToValidator(ctx context.Context, batch []*batchItem) {
 	// (unbuffered handoff, no timeout). complete is CAS-guarded, so
```
freemans13: This comment is now false. With `group.Wait(ctx, batchHandoffTimeout)` a panic here no longer strands submitters with no timeout: they are released by the caller context or after 10 minutes. The same stale text is at `services/propagation/Client.go:601` and `:485`.

services/propagation/Client.go
```diff
@@ -564,14 +601,9 @@ func (c *Client) ProcessTransactionBatch(ctx context.Context, batch []*batchItem
 	// (unbuffered handoff, no timeout). complete is CAS-guarded, so
```
freemans13: Same as the validator client: "no timeout" stopped being true in this PR. `handleBatchResponse` at line 485 also still says single-item `ProcessTransaction` "waits unbounded with no ctx arm".

services/validator/Client.go
```diff
@@ -422,12 +441,31 @@ func (c *Client) ValidateWithOptions(ctx context.Context, tx *bt.Tx, blockHeight
 	}
 	c.batcher.PutCtx(ctx, item)
 
-	// group.Wait(context.Background(), 0): 0 timeout allocates no timer and a
-	// background context never cancels, so this blocks purely on the dispatcher
-	// completing the item — identical to the previous bare <-doneCh receive (no
-	// timeout, no ctx arm). It can only return nil, so the result slot is always
-	// safe to read here.
-	_ = group.Wait(context.Background(), 0)
+	// Bounded by the CALLER's context AND a finite backstop. Every resolved
```
freemans13: The survey misses some batch-mode callers. Subtree validation (`SubtreeValidation.go:432`, errgroup ctx), legacy block pre-validation (`handle_block.go:1434` and `:1590`) and the propagation server Kafka-less branch (`propagation/Server.go:1429`) all reach this wait. None of them changes the conclusion. But legacy `PreValidateTransactions` does not have an "ambiguous hand-off failure" branch: `ServiceError` is not `IsRetryableError`, so it hard-fails and the outer `IsTransientLocalError` check backs the block off. I would drop the enumeration rather than extend it.

util/batch_panic.go
```diff
@@ -0,0 +1,52 @@
...
+// escaping both is what keeps the record on a single line.
+func SignalBatchPanic[T any](recovered any, batch []T, fnName string, logger ulogger.Logger, signal func(item T, err error)) bool {
+	if recovered == nil {
+		return false
+	}
+
+	logger.Errorf("[%s] recovered panic, failing %d batch item(s): recovered=%q stack=%q",
+		fnName, len(batch), fmt.Sprintf("%v", recovered), debug.Stack())
+
+	err := errors.NewProcessingError("panic in %s: %v", fnName, recovered)
```
freemans13: A nil-dereference panic recovers as a `runtime.Error`, which is an `error`, so `errors.New` takes it as the wrapped error and `%v` is orphaned. Measured via a `go test -overlay` case:

```
PROCESSING (4): panic in sendGetBatch: %!v(MISSING) -> UNKNOWN (0): runtime error: invalid memory address or nil pointer dereference
```

This predates the PR (all four old copies had it), but it is one line now:

```go
err := errors.NewProcessingError("panic in %s: %s", fnName, fmt.Sprint(recovered))
```

The existing test panics with a string, so a case that panics with a real `runtime.Error` would pin it.

freemans13 (approved):
Approving. I read all four commits at `274e9334f` and found nothing blocking. The three notes below are about comments that the new bound makes false, plus one error-text defect that predates this PR but now lives in a single place, so it costs one line to fix.

What I checked, beyond what ordishs and the bot already covered:

- **Who receives the new abandonment error.** The validator client's batch path has more production callers than the comment at `services/validator/Client.go:444` lists: subtree validation's `blessMissingTransaction`, legacy block pre-validation in `handle_block.go`, and the propagation server's Kafka-less branch. None of them turns it into an invalid-tx or invalid-block verdict. Subtree validation wraps it as a processing error. On the legacy block path it is not `IsRetryableError`, so `PreValidateTransactions` hard-fails instead of running its own retry loop. The outer handler then classifies it with `IsTransientLocalError`, which walks the wrapped chain and finds the `ServiceError`, so the peer is kept and the block is backed off rather than rejected (`manager.go:1832`). That's the right outcome, but it goes through a different branch than the one the comment describes.
- **Queue time against the 10-minute backstop.** Both subtree validation and legacy pre-validation cap their errgroups (at `SpendBatcherSize*2` and `spendBatcherSize*spendBatcherConcurrency`), so client-side queue depth is bounded. I found no path where a healthy node would sit in the queue anywhere near 10 minutes.
- **The spend alias.** A nil spend returns before anything is enqueued, so `toEnqueue := items` can never carry a nil entry into the batcher.

Notes, all non-blocking:

1. Three existing comments still describe the wait as unbounded, which is no longer true: `services/validator/Client.go:533`, `services/propagation/Client.go:601` (both say "unbuffered handoff, no timeout"), and `services/propagation/Client.go:485` ("single-item ProcessTransaction waits unbounded with no ctx arm"). The lines themselves predate this PR, but the PR is what makes them wrong. The danger is that the next reader believes a panic still strands submitters forever.
2. The caller survey in the validator client comment is incomplete, as described above. The conclusion holds either way: nothing in the call chain guarantees a deadline, so the backstop is needed. I'd drop the enumeration, or at least the word "Every", rather than try to keep a hand-maintained list current. The PR body makes the same argument about the roster you removed from the panic-sweep godoc.
3. `util/batch_panic.go:46` produces a broken item error for the most common real panic. A nil dereference recovers as a `runtime.Error`, which is an `error`, so `errors.New` takes it as the wrapped error and the `%v` has nothing left to format. I ran a nil-pointer panic through `SignalBatchPanic`, using a `go test -overlay` so nothing touched the branch, and the item got:
   ```
   PROCESSING (4): panic in sendGetBatch: %!v(MISSING) -> UNKNOWN (0): runtime error: invalid memory address or nil pointer dereference
   ```
   The log line is fine, since it formats the value itself. All four old copies had the same bug, so "error text unchanged" is accurate. `TestSignalBatchPanic` panics with the string `"boom"`, so it can't catch this. Formatting the value first, `fmt.Sprint(recovered)`, fixes it, the same way `recoverBatcherShutdown` already does at `aerospike.go:150`. A test case that panics with a real `runtime.Error` would pin it.

ordishs's point about the orphaned `%s` in the two new `NewContextCanceledError` messages (`spend.go:1161`, `spend.go:1388`) still applies at this head. I haven't repeated it inline.

**Verification:** CI shows 37 checks passing and 1 skipped. Locally I ran `go test -race` on `./util`, `./services/propagation` and `./services/validator`, filtered to the new wait-bound, abort-path and panic-sweep tests, and all passed. I did not run the aerospike store tests locally (`spend_wait_cancel_test.go`, `spend_enqueue_alias_test.go`), so for those I'm relying on CI.

ctnguyen:
Thanks @ordishs and @freemans13 for the approvals. I'll merge the PR after the CI passes all the tests.

## Fixed

- `spend.go:1161`, `:1388` — the trailing `%s` is dropped from both new cancellation messages;
  `waitErr` stays the last argument, so the cause still renders through the wrap chain.
- `util/batch_panic.go:46` — the verb is filled, but not with the line that was suggested. That form
  renders the recovered value into text, and `errors.New` keeps a recovered `*errors.Error` whole,
  so a classifiable cause would have been flattened out of the chain `IsRetryableError` walks. The
  value is now rendered for the verb *and* passed again as the trailing cause, so the message is
  complete and the chain survives; a non-error value is not passed twice, since it would survive as
  a spare parameter.
- `propagation/Client.go:485`, `:601`, `validator/Client.go:533` — the three comments that still
  described the hand-off wait as unbounded now describe the caller context and the backstop.
- `validator/Client.go:444` — the caller enumeration is gone rather than extended. `:465` — the
  "ambiguous hand-off failure" clause is replaced by the property that is actually guaranteed: no
  caller unwinds a transaction that may still be in flight.
- the no-`s.ctx.Err()`-recheck policy at `spend.go:1381` is now pinned by a test rather than by a
  signature — `TestIncrementSpentRecords_TimeoutIsNotRelabelledByACancelledStoreContext` in
  `spend_wait_cancel_test.go`, using a stub store context whose `Done()` never fires and whose
  `Err()` reports `Canceled`, so the verdict has to come from the error `Wait` returned.

## Deferred

- `spend.go:450-452`, the uncovered `if filtered` branch — it is reachable only via the breaker's
  clock or another goroutine's `RecordSuccess`, so there is no deterministic seam. Covering it means
  making `spendCircuitBreaker` an interface, i.e. a production type change on the per-input spend
  path, or a timing-dependent test.
- `spend.go:503`, `:510`, `:1164`, `:1395` — the same orphaned-verb defect in four pre-existing
  messages. This round fixes what this PR introduced, not what it inherited, which is why the diff
  looks partial around the two edited lines.

PR #1722, opened 2026-09-11
test: fix CI fixtures after transaction re-extension (release/v0.15)

icellan:
Backport of https://github.com/bsv-blockchain/teranode/pull/1721 to `release/v0.15`. The test and documentation changes are identical to main, with no branch-specific adjustments.

Mandatory transaction re-extension exposed three test assumptions: concurrent propagation submissions shared a mutable transaction, the zero-value script fixture depended on the configured minimum fee, and the reassignment smoke test trusted a replacement script that was never stored.

This change gives each concurrent submission its own decoded transaction while retaining the duplicate-submission assertion, explicitly disables the minimum fee only for the unspendable-script control, and checks reassignment maturity with a spend signed for the stored owner. The smoke test also verifies that an unstored replacement script remains rejected after maturity and leaves the output unspent.

**Reassignment limitation:** `ReAssignUTXO` changes the commitment and maturity state but does not persist the replacement locking script. Re-extension therefore retrieves the original script. Supporting a different owner requires an authoritative replacement-script source and corresponding validation support. This limitation is documented in the alert-service guide; production code is unchanged.

Validation on this branch:
- Full propagation and validator packages with `SETTINGS_CONTEXT=test`, the race detector, and `testtxmetacache`, including real Aerospike/PostgreSQL containers.
- Targeted propagation concurrency tests and reassignment smoke test, each repeated three times with the race detector.
- `go vet` and `staticcheck` for the three affected packages.
- Repository pre-commit checks, including Go lint and Markdown lint.

The zero-value fixture's `insufficient-fee` failure was reproduced with a nonzero minimum fee before verifying the fix.

A broader local run also hit an HTTP startup failure on main and a batch-test failure on release. Both full propagation packages and three isolated runs of the release batch test passed on rerun. These tests are unchanged by this patch.

### DIFF
--- a/docs/topics/services/alert.md
+++ b/docs/topics/services/alert.md
@@ -138,6 +138,16 @@ The Alert Service initializes the necessary components and services to start pro
 
 ### 2.4. UTXO Reassignment
 
+**Current limitation:** `ReAssignUTXO` updates the UTXO commitment, clears the freeze,
+and sets a maturity height, but does not persist the replacement locking script.
+Mandatory transaction re-extension therefore still retrieves the original script
+from the UTXO store. A spend signed only for the replacement owner is rejected even
+after maturity; supplying the replacement script in extended transaction bytes
+cannot authorize it. Supporting a different owner requires an authoritative source
+for the replacement script and corresponding validation support. Reassignment to
+the stored owner still enforces the maturity delay and permits a valid spend after
+that delay. The reassignment smoke test covers both behaviors.
+
 ![alert_reassign_utxo.svg](img/plantuml/alert/alert_reassign_utxo.svg)
 
 1. The P2P Alert library initiates the process by calling `AddToConfiscationTransactionWhitelist` with a list of transactions.
--- a/services/propagation/Server_test.go
+++ b/services/propagation/Server_test.go
@@ -924,21 +924,29 @@ func testProcessTransactionInternal(t *testing.T, utxoStoreURL string) {
 			numGoroutines = 10
 		}
 
-		// add the transaction in parallel
+		// Each submission owns its decoded transaction, as real ingress does.
+		// Validation re-extends inputs in place; sharing pointers across requests
+		// races even though the transactions have identical txids.
+		tx2Bytes, tx3Bytes := txs[2].ExtendedBytes(), txs[3].ExtendedBytes()
 		for i := 0; i < numGoroutines; i++ {
+			tx2, err := bt.NewTxFromBytes(tx2Bytes)
+			require.NoError(t, err)
+			tx3, err := bt.NewTxFromBytes(tx3Bytes)
+			require.NoError(t, err)
+
 			g.Go(func() error {
-				if err := ps.processTransactionInternal(t.Context(), txs[2]); err != nil {
+				if err := ps.processTransactionInternal(t.Context(), tx2); err != nil {
 					return err
 				}
 
-				return ps.processTransactionInternal(t.Context(), txs[3])
+				return ps.processTransactionInternal(t.Context(), tx3)
 			})
 		}
 
 		require.NoError(t, g.Wait(), "processTransactionInternal should not return an error for valid transaction")
 
 		// make sure we only added the transaction once to block assembly
-		assert.Len(t, blockAssemblyClient.Mock.Calls, 2, "processTransactionInternal should only call block assembly once for the same transaction")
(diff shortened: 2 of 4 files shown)


oskarszoon (approved):
Tests and docs only, mechanics verified: the extended-bytes round-trip is sound (go-bt
`ReadFrom` detects the `0xEF` marker, so `PreviousTxScript`/`PreviousTxSatoshis`
survive `NewTxFromBytes`), settings are fresh per subtest so the fee override can't
leak, and the per-goroutine decode fixes a real shared-pointer race. Two things before
this lands.

**Test name now says the opposite of what it asserts** — `test/e2e/daemon/ready/reassign_test.go:24`

`TestShouldAllowReassign` asserts reassignment to a different owner is rejected before
maturity, after maturity, and leaves the output unspent. The only surviving success
path reassigns to the stored owner, which changes no ownership. Rename it to what it
pins — `TestReassignEnforcesMaturityAndRejectsUnstoredScript` or similar — otherwise
the next person greps for reassignment coverage and concludes the feature works.

**The regression needs more than a docs paragraph**

Confirmed in both stores: `stores/utxo/sql/alert_system.go:179-186` updates
`utxo_hash`/`frozen`/`spendableIn` only, and
`stores/utxo/aerospike/alert_system.go:245-251` passes just the two hashes and heights
to the Lua UDF. Neither persists the replacement script, so alert-service reassignment
to a new owner is non-functional on release/v0.15 after the security fix.

No tracking issue exists. This is the confiscation path — it needs an issue and a
v0.15.x release-note entry, not only `docs/topics/services/alert.md:141`.

**Smaller**

- `services/validator/utxo_commitment_ambiguity_test.go:184-187` — reaching in to set
  `MinMiningTxFee` and rebuild `txValidator` after construction would be cleaner as an
  option on `newAmbiguityValidator`. Safe as written.
- `reassign_test.go:136` vs `:175` — you assert `SpendingData == nil` after Charles's
  rejection but not after Bob's early rejection, then re-submit that same transaction
  at :178. One more assertion makes the later `require.NoError` unambiguous.

**Unrelated, found while checking the above:** `sql/alert_system.go:174-178` honours
`tSettings.UtxoStore.ReAssignedUtxoSpendableAfterBlocks` but
`aerospike/alert_system.go:249` hardcodes the package constant (1000). Production runs
Aerospike, so that setting is silently a no-op there.

galt-tr (approved):

oskarszoon:
Follow-up on the reviewed-and-approved state here, from reviewing the identical change on `main` in #1721. Not asking to reopen this — the test mechanics are correct and I confirmed that independently. One thing in the docs paragraph needs fixing on this branch too, and it matters more here because `v0.15.9-beta-1` is already tagged.

`docs/topics/services/alert.md` says a spend signed for the replacement owner is rejected. It omits that the **original stored owner is locked out as well**: re-extension supplies the original script so the script check passes, then the store's commitment comparison fails against the new hash. Probed on the PR's own test: `UTXO_MISMATCH (74): [Spend] utxo hash mismatch`. The output is stranded for both parties and there is no inverse operation in the store API.

The risk is an operator reading that paragraph as "the coin doesn't move yet", running a court-ordered confiscation, and permanently burning the output. Recovery — re-freeze and reassign back to the original hash — is undocumented and diverges by backend: SQL's freeze ignores the hash (`sql/alert_system.go:96`), Aerospike's matches it (`aerospike/alert_system.go:74-76`).

Two other sentences in the same paragraph:

- "The reassignment smoke test covers both behaviors" holds for SQL only. The test sets no `UTXOStoreType`, so it runs on sqlite, and the maturity assertion depends on `ReAssignedUtxoSpendableAfterBlocks = 5` — SQL honours it, Aerospike ignores it (`aerospike/alert_system.go` passes the hardcoded constant from `Interface.go`). Production Aerospike has no coverage of either behaviour and could not run this test against a 1000-block gate.
- "Reassignment to the stored owner still enforces the maturity delay" describes a case no caller produces. `services/alert/node.go` always derives a new P2PKH from the confiscation input's pubkey, so `new != old` by construction. The surviving success path in the test is a self-reassignment reachable only via the operator RPC. Every production path is broken; only a no-op still works.

There is no tracking issue for the regression — `gh issue list --search reassign` and `gh search issues ... ReAssignUTXO` / `... confiscation` return nothing. Given this branch ships, the release body wants a `### Notes` entry naming that issue, in the same style `v0.15.9-beta-1` already uses for #1701. Five other docs still advertise the feature working, and `services/alert/node.go` still reports these to the alert network as `Processed` with a comment that is now false.

Full detail on #1721.

PR #1724, opened 2026-09-11
fix(validator): bound HTTP fallback diagnostics

oskarszoon:
The validator HTTP fallback buffers the whole non-200 response before reporting an error. Limit diagnostic reads to a 2 KiB prefix plus one byte to detect truncation, quote control characters, and mark truncated bodies. Preserve validation verdicts from response headers and the existing success behavior.

Responses still close on every successful HTTP exchange. Large error responses close without draining the remainder, which can prevent HTTP/1 connection reuse for those responses.

Validation on the merged upstream content: focused HTTP fallback tests and the full validator package pass under the race detector. Vet, staticcheck, and changed-code golangci-lint pass. Previous local security scans reported existing Go toolchain and unchanged-file findings; those scans were not repeated for this unchanged patch.

Merged upstream/main through `13897b045` without changing the PR's two-file patch. On head `53b86e903`, all three smoke shards, all seven sequential shards, and the four required checks pass. The initial legacy-sync attempt failed during daemon startup with an occupied port; its [targeted rerun](https://github.com/bsv-blockchain/teranode/actions/runs/34842897460/job/103978560782) passes without source changes. All reported checks now pass except the expected skipped native-ops job.


services/validator/Client.go
```diff
@@ -627,6 +627,20 @@ func (c *Client) notifyAllBatchItems(batch []*batchItem, metadata []byte, err er
 	}
 }
 
+// maxHTTPFallbackErrorBodyBytes bounds the raw prefix retained for diagnostics.
+const maxHTTPFallbackErrorBodyBytes = 2 * 1024
```
freemans13, line 661: Non-blocking, but worth a thought before merge. The validator's `/tx` handler writes `"[handleSingleTx] Failed to process transaction: "+err.Error()` (`Server.go:1039`), and `errors.(*Error).Error()` renders the chain outermost first, appending each wrapped error after ` -> `, including any `data` message. The innermost cause, usually the most specific part, is therefore at the tail, and a head-only prefix is the first thing to lose it.

This fallback only runs for transactions too large for gRPC, which is the case most likely to carry a long `data` payload, a list of missing parents for example. In the no-verdict branch the body is the only diagnostic the caller gets.

The verdict header on the same response is allowed 4 KiB (`maxHTTPErrorHeaderBytes`), so 2 KiB for the full diagnostic is tighter than the thing it supplements. Two cheap options that stay bounded: raise the cap to something like 16 KiB, or keep a head and a tail (for example 1 KiB each with `...` between). The tail option costs reading to EOF, which gives up the no-drain property, so raising the cap is probably the simpler trade.

services/validator/client_http_error_body_test.go
```diff
@@ -0,0 +1,177 @@
...
+			require.IsType(t, &errors.Error{}, err)
+
+			actual := err.(*errors.Error)
+
+			diagnostic := fmt.Sprintf("%q", tc.body[:min(len(tc.body), 2048)])
+			if len(tc.body) > 2048 {
+				diagnostic += " (truncated)"
+			}
+
+			require.LessOrEqual(t, len(diagnostic), 4*2048+2+len(" (truncated)"))
```
freemans13, line 154: Nit: `diagnostic` here is the string the test built on the lines above, so this assertion checks the test's own arithmetic and cannot fail because of anything in `Client.go`. The bound is already enforced indirectly by the `require.Equal` against `actual.Message()` and `logger.warning` below. Either drop this line, or point it at the real output, e.g. `require.LessOrEqual(t, len(actual.Message()), ...)` in the service-error branch and `len(logger.warning)` in the verdict branch.

freemans13 (approved):
Approving. The cap is correct and the change is well tested. I have two non-blocking points and one follow-up.

What I checked, on commit 192ef1f2a:

- The bound. `io.LimitReader(body, max+1)` reads at most 2,049 bytes, returns 2,048, and uses the spare byte only to detect truncation. The boundary tests at 2047, 2048, 2049 and 2080 bytes also assert how many bytes were actually read, so a regression to an unbounded read would fail them.
- Log forging. `%q` escapes CR, LF, ESC and NUL in both branches, the returned SERVICE_ERROR and the Warn line. Invalid UTF-8 from a mid-rune cut comes out as `\xNN`, which is harmless.
- Verdict handling. The header is still read before the body, the success path ignores both, and a partial read (`ErrUnexpectedEOF`) still produces the prefix it got.
- Consumers. The only readers of these message strings are the two lines changed here. No test or caller elsewhere matches on the old unquoted `body: ` text.
- Locally, `go test -race -count=1 -run 'TestReadHTTPFallbackErrorBody|TestValidateTransactionViaHTTP' ./services/validator/` passes.

The points worth a look:

1. **2 KiB of head drops the root cause of a long error chain** (inline on `Client.go:631`). Non-blocking.
2. **One test assertion checks the test's own value, not the code's output** (inline on the test file). A nit.

Follow-up, out of scope here: the bot already named `services/propagation/Client.go:373` and `:520`, but the direct twin of this code is `services/propagation/Server.go:1543`. It does the same unbounded `io.ReadAll(resp.Body)` and passes the result to `validatorRejection`, which formats it unquoted into the same two message shapes. After this PR the validator client and the propagation server treat an identical response differently.

CI: 36 checks pass. `smoketest (shard 2)` failed in `TestCatchup` ("Catchup After Pruner", timeout waiting for a block hash at `daemon/test_daemon.go:1648`). That test never sends a transaction large enough to leave gRPC for the HTTP fallback, so I read it as unrelated, but I did not rerun it or check its failure rate on main.

ordishs (approved):
services/propagation/Server.go:1543, services/propagation/Client.go:373, services/propagation/Client.go:520

- The bound lands on one of four identical call sites. These three keep `io.ReadAll(resp.Body)` with no limit and still format the result with raw `%s`.
- services/validator/Client.go:701 states its contract as "Same contract as propagation's fallback", so the relation is explicit in this diff.
- Server.go:1543 feeds validatorRejection, whose own comment calls it "the only path on which a Kafka-wired node answers a submitter with a real verdict".
- Failure: a validator, or a proxy answering for it, returns 400 with a 100 MB body. Server.go:1543 allocates 100 MB per in-flight request, and CR/LF in that body forge extra log lines — the exact class this PR fixes.
- Suggestion: move the helper to a package both services import, or export it, and apply it at all four sites. A follow-up PR is fine; nothing here regresses.

Verified locally: the new tests pass under -race, `go vet ./services/validator/` is clean, and `golangci-lint run ./services/validator/...` reports 0 issues.

Issue #1725, opened 2026-09-11
Reassignment after mandatory re-extension strands UTXOs for both owners


icellan:
Mandatory transaction re-extension exposes a reassignment regression: changing a frozen UTXO to a different owner updates its commitment but does not persist the replacement locking script. Re-extension retrieves the original script, so the new owner's signature fails. The original owner's signature passes that script but the recomputed commitment differs from the stored reassigned commitment. Neither owner can spend the output after maturity.

This affects the ownership-changing alert/confiscation and RPC flows on main after `3e26a5d2fad9e22cf2ea4f40ee4babf362ac5d28` and release/v0.15 after `c7682127778565279fe51369194a88e4925dedd6`. The existing `v0.15.9-beta-1` tag points to `9272b80c9259d00fd51eb8f66ee658a46adb940a`, before that release-branch merge.

Reproduction:
1. Store and mine an output locked to Bob.
2. Freeze it and call `ReAssignUTXO` with a commitment to Charles's script.
3. Advance beyond the reassignment maturity height.
4. Submit a Charles-signed spend with the replacement script in its extended fields: script validation rejects it.
5. Submit a Bob-signed spend: the store rejects it with `UTXO_MISMATCH`.

Operator impact: successful `reassign`/alert processing does not establish spendability. Avoid ownership-changing reassignment until this is resolved. There is no documented, validated recovery procedure; do not assume that re-freezing and restoring the original commitment is portable. SQL freezing locates by outpoint without checking the supplied commitment, while Aerospike checks it.

Required follow-up:
- Provide an authoritative replacement-script source and validation support without trusting submitter-supplied extended fields.
- Restore new-owner spending after maturity and reject the old owner, with SQL and Aerospike end-to-end coverage.
- Define failure/Processed reporting and recovery for already-reassigned outputs.
- Resolve the maturity-setting discrepancy: SQL honors `utxostore_reassignedUtxoSpendableAfterBlocks`, while Aerospike currently uses the fixed 1,000-block constant.
- Add verified operator guidance and release notes when support is restored.

The SQLite smoke test in https://github.com/bsv-blockchain/teranode/pull/1721 (backport https://github.com/bsv-blockchain/teranode/pull/1722) documents and pins the current limitation. Its same-owner reassignment is a control for maturity, not proof that ownership-changing confiscation works. Those PRs remain tests/documentation only.

icellan:
Review follow-ups from PR #1726 for the production reassignment work:

[...]

PR #1726, opened 2026-09-11
fix: reassignment test coverage and legacy shutdown (release/v0.15)

icellan:
Mandatory transaction re-extension leaves an ownership-changing reassignment unspendable by both owners because the replacement script is not persisted. This PR documents that limitation and repairs the release branch’s regression coverage. It also makes every legacy Stop caller cancel and join the service and its internal server, complementing the handler drain now present in `release/v0.15`.

- Reuse the daemon’s local validator, match ingress’s candidate height, require the public validation-error wrapper and exact internal error codes, and check every rejected output remains unspent.
- Verify the configured maturity gate at its exact height, waiting on the store height before querying output status.
- Fix the explicit SQLite test-store URL and verify it opens a usable in-memory database.
- Exercise shutdown with a real SQL batcher query waiting for a database connection, verify concurrent Stop callers wait and late handlers cannot touch the closed store, and close the blockchain fixture’s background resources.
- Consolidate operator guidance and backend differences in the Alert Service guide, correct RPC/store comments and references, and rebuild the reassignment diagrams.

The latest `release/v0.15` is merged. Conflicts were resolved by retaining the release branch’s broader handler draining and shutdown tests, then strengthening the SQL regression. Runtime reassignment guards, script persistence/recovery, and backend parity changes remain tracked in https://github.com/bsv-blockchain/teranode/issues/1725 under the agreed reassignment scope.

Validation: three race-enabled reassignment runs, five shutdown-test runs, SQLite URL regression, vet/staticcheck, repository lint, Markdown checks, and rendered-diagram inspection. Race-enabled model, legacy, block-validation, daemon, container-helper and settings packages also pass on the merged branch. The complete smoke, sequential, legacy-sync, pruner, chain-integrity and unit CI jobs passed on 552336da3. The subsequent direct-Stop fix in 1e08f2bb8 passes five race-enabled regression runs and the complete legacy/daemon race suites, vet, staticcheck and lint; fresh CI is pending for this final commit.

### Release note

Do not use ownership-changing reassignment in releases containing mandatory re-extension: successful RPC or alert processing does not establish spendability, and both owners remain locked out after maturity. See https://github.com/bsv-blockchain/teranode/issues/1725. The existing `v0.15.9-beta-1` tag predates the re-extension security merge and is not marked as affected.


daemon/test_daemon.go
```diff
@@ -110,7 +110,7 @@ type TestOptions struct {
 	StartDaemonDependencies bool
 	FSMState                blockchain.FSMStateType
 	// UTXOStoreType specifies which UTXO store backend to use ("aerospike", "postgres")
-	// If empty, defaults to "aerospike"
+	// If empty, keeps the configured store (SQLite with SystemTestSettings).
```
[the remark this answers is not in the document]
icellan: Added `sqlite` to the supported backend values and kept the clarification that an empty value preserves the configured store. Fixed in a469af0c5eb868171c8912bec0dd049ed3b8584c.

stores/utxo/Interface.go
```diff
@@ -359,8 +359,11 @@ type Store interface {
 	// UnFreezeUTXOs removes the frozen status from UTXOs, allowing them to be spent again.
 	UnFreezeUTXOs(ctx context.Context, spends []*Spend, tSettings *settings.Settings) error
 
-	// ReAssignUTXO reassigns a UTXO to a new transaction output.
-	// The UTXO will become spendable after ReAssignedUtxoSpendableAfterBlocks blocks.
+	// ReAssignUTXO updates a frozen UTXO's commitment and maturity gate.
+	// It does not persist a replacement locking script. Changing the owner
+	// currently strands the output for both owners even after maturity; see
+	// https://github.com/bsv-blockchain/teranode/issues/1725.
+	// SQL honors the configured delay; Aerospike currently uses the fixed constant.
```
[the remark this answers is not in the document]
icellan: Synchronized the checked-in API reference with the Store interface comment and corrected its summary bullet. Both now state that replacement scripts are not stored and changing the owner strands the output after maturity, with the limitation/issue link. Fixed in a469af0c5eb868171c8912bec0dd049ed3b8584c.

icellan:
Pushed 0d4bee483 to fix the smoke-test shutdown panic. Legacy now waits for its internal server to exit, and every sync-manager Stop caller waits for active inventory reads to finish. Late inventory handlers are rejected before they touch the UTXO store. This fixes the lifecycle race exposed by `TestBIP68_TimeBased_Accept`; reassignment storage is unchanged.

The new regression test fails on the previous code and passes ten race-enabled runs with the fix. The legacy/netsync suites, BIP68 smoke run, and complete smoke shard 0 all pass locally with the race detector. Vet, staticcheck, repository lint and commit hooks pass. Local smoke runs use an external port-only Go overlay because OrbStack occupies the default SV Node ports; the PR changes no ports.

CI verification is complete: all three smoke shards pass at https://github.com/bsv-blockchain/teranode/actions/runs/34623788714, and the unit-test job passes at https://github.com/bsv-blockchain/teranode/actions/runs/34623788760.

daemon/test_daemon.go
```diff
@@ -109,8 +109,8 @@ type TestOptions struct {
 	SkipRemoveDataDir       bool
 	StartDaemonDependencies bool
 	FSMState                blockchain.FSMStateType
-	// UTXOStoreType specifies which UTXO store backend to use ("aerospike", "postgres")
-	// If empty, defaults to "aerospike"
+	// UTXOStoreType specifies which UTXO store backend to use ("aerospike", "postgres", "sqlite")
+	// If empty, keeps the configured store (SQLite with SystemTestSettings).
```
galt-tr, line 113: **[code-review 1/15]** The rewritten comment lists `sqlite` as a valid `UTXOStoreType`, but passing it routes through the container manager to `sqlite://memory:`, whose empty URL path panics in `util/sql.go:221` (`storeURL.Path[1:]`).

`TestOptions{UTXOStoreType: "sqlite"}` -> `containers.NewContainerManager` returns `sqlite://memory:` (test/utils/containers/container_manager.go:319-322) -> util/sql.go:212 treats only scheme `sqlitememory` as in-memory, :221 slices `Path[1:]` on an empty path -> runtime panic: slice bounds out of range [1:0]. So neither the empty value (file store `sqlite:///utxo` from `test.SystemTestSettings`) nor `sqlite` behaves as the comment implies; drop `sqlite` from the list (or fix the container URL) rather than advertise a value that crashes.

docs/topics/services/alert.md
```diff
@@ -138,15 +138,31 @@ The Alert Service initializes the necessary components and services to start pro
...
+even after maturity. `ReAssignUTXO` updates the commitment, freeze state and maturity
+height but does not persist the replacement locking script. The new owner's spend
+fails against the original stored script. The original owner's signature passes
+that script, but its spend fails with `UTXO_MISMATCH` against the changed commitment.
+Waiting for more blocks does not restore spendability.
+
+This affects ownership-changing alert/confiscation and RPC reassignment. A successful
+RPC result or alert processing response does not guarantee that the output can be
+spent. There is no documented, validated recovery procedure across backends: SQL
+freezing ignores the supplied commitment, whereas Aerospike checks it, so operators
```
galt-tr, line 152: **[code-review 2/15]** The parity paragraph says only *freezing* ignores the supplied commitment, but SQL `ReAssignUTXO` also never reads `oldutxohash` (SELECT/UPDATE keyed on txid/vout only), so the parameter the new RPC docs call the 'current commitment' is unchecked on SQLite/Postgres and checked on Aerospike.

`reassign <txid> <vout> <wrong-old-hash> <new-hash>` on SQL: stores/utxo/sql/alert_system.go:164 `QueryRow(TxID, Vout)`, :181-188 `UPDATE ... WHERE transaction_id=$3 AND idx=$4 AND frozen=true` (no utxo_hash predicate) -> succeeds and rewrites the commitment; on Aerospike teranode.lua:885 byte-compares -> UTXO_HASH_MISMATCH. The 'recovery is not portable' caveat therefore understates reassignment itself; add it to this list and to rpc.md:1060/rpc_reference.md:1550/minersHowTo:202. Same doc lines also promise 'an error if a hash cannot be parsed' (rpc.md:1066) although `chainhash.NewHashFromStr` zero-pads short hex (go-bt chainhash/hash.go:143-156) and `uint32(c.OldVout)` (handlers.go:2688) wraps -1 to 4294967295, both surfacing as store not-found errors rather than parameter errors.

test/e2e/daemon/ready/reassign_test.go
```diff
@@ -41,9 +44,31 @@ func TestShouldAllowReassign(t *testing.T) {
...
+		t.Helper()
+		// Snapshot before either path runs, and decode separate requests.
+		// Keep tx untouched so retries still supply the original extended fields.
+		submittedBytes := tx.ExtendedBytes()
+		ingress, err := bt.NewTxFromBytes(submittedBytes)
+		require.NoError(t, err)
+		require.Error(t, td.PropagationClient.ProcessTransaction(td.Ctx, ingress))
+		probe, err := bt.NewTxFromBytes(submittedBytes)
+		require.NoError(t, err)
+		_, err = v.ValidateWithOptions(td.Ctx, probe, td.UtxoStore.GetBlockHeight(),
```
galt-tr, line 65: **[code-review 3/15]** The probe validates at the store tip (`td.UtxoStore.GetBlockHeight()`) while the ingress path it mirrors validates at tip+1 (propagation passes 0, which Validator.go:565-568 rewrites to `blockState.Height+1`), and the SQL spend gate compares against the caller's height, so the two legs of `requireRejected` disagree at the maturity boundary.

Store height H, reassigned output with spendableIn == H+1: ingress (H+1 < H+1 is false; sql.go:2151/:2653 use the caller's `item.blockHeight`) ACCEPTS and spends the output, while the probe (H < H+1) and `GetSpend` (sql.go:3446, store height) both report immature -> `require.Error` on the ingress leg fails with the UTXO already spent. Latent in the current sequence (reassign at 4 -> spendableIn 9; immature probe at 4/5; matured at >= 10), but the helper's 'pair every rejection with its exact cause' premise is false. Pass 0, exactly as the ingress path does (Options already match NewDefaultOptions, options.go:182-189).

test/e2e/daemon/ready/reassign_test.go
```diff
@@ -41,9 +44,31 @@ func TestShouldAllowReassign(t *testing.T) {
...
+		nil, nil, nil, td.BlockchainClient)
+	require.NoError(t, err)
+	requireRejected := func(tx *bt.Tx, reason string) {
+		t.Helper()
+		// Snapshot before either path runs, and decode separate requests.
+		// Keep tx untouched so retries still supply the original extended fields.
+		submittedBytes := tx.ExtendedBytes()
+		ingress, err := bt.NewTxFromBytes(submittedBytes)
+		require.NoError(t, err)
+		require.Error(t, td.PropagationClient.ProcessTransaction(td.Ctx, ingress))
```
galt-tr, line 62: **[code-review 4/15]** The ingress leg accepts any error, so an unrelated failure (gRPC `codes.Unavailable` 'server at capacity', sanity/storage errors) paired with an independently matching probe passes without ingress ever rejecting for the asserted reason; and the 'rejected spend leaves the output unspent' check runs after only three of the helper's four calls.

services/propagation/Server.go:993 returns Unavailable under load, :1202/:1206/:1215 return non-validator errors. `PublicError` (errors/errors.go:707-713) drops the wrapped chain but keeps the outermost `[ProcessTransaction][txid] failed to validate transaction` (Server.go:1234-1236) on both the single and batch transports -> assert `require.ErrorContains(err, "failed to validate transaction")` on the ingress leg. Pass the spend into `requireRejected` so the `SpendingData == nil` block (copied at :164-166, :207-209, :218-220 and absent after :188) runs on every rejection. The :49 comment 'redacts some validation errors' understates: no validator reason survives either transport.

test/e2e/daemon/ready/reassign_test.go
```diff
@@ -41,9 +44,31 @@ func TestShouldAllowReassign(t *testing.T) {
...
+		probe, err := bt.NewTxFromBytes(submittedBytes)
+		require.NoError(t, err)
+		_, err = v.ValidateWithOptions(td.Ctx, probe, td.UtxoStore.GetBlockHeight(),
+			&validator.Options{AddTXToBlockAssembly: false})
+		require.ErrorContains(t, err, reason)
```
galt-tr, line 67: **[code-review 5/15]** `"not spendable until"` appears in both backends' messages but the codes differ - SQL raises ERR_TX_LOCKED for an immature reassigned spend, Aerospike ERR_UTXO_FROZEN - a reassignment parity gap the PR's new parity paragraph omits and that this substring would mask if the test ever ran against Aerospike.

sql.go:2151-2152 `NewTxLockedError` vs aerospike/teranode.lua:371-377 FROZEN_UNTIL -> spend.go:896-897 `NewUtxoFrozenError`. Consumers diverge: Validator.go:435 retries only ErrTxLocked (SQL retries 4x with backoff, Aerospike fails fast); propagation HTTP maps ErrFrozen -> 403 and ErrTxLocked -> 409 (Server.go:636-642). Add this to alert.md 2.4's parity list, and pin the store errors with `require.ErrorIs(err, errors.ErrTxLocked)` / `errors.ErrUtxoHashMismatch` (errors.go:154-179 matches by code through the chain; precedent stores/utxo/tests/tests.go:500-502). Keep the substring only for OP_EQUALVERIFY, where ErrTxInvalid would be less specific (the text comes from BDK's C++ error string, ScriptVerifierGoBDK.go:348).

test/e2e/daemon/ready/reassign_test.go
```diff
@@ -158,24 +185,41 @@ func TestShouldAllowReassign(t *testing.T) {
...
 
-	err = td.PropagationClient.ProcessTransaction(td.Ctx, charlesSpendingTx)
-	require.Error(t, err, "an unstored replacement script must not be accepted")
+	requireRejected(charlesSpendingTx, "OP_EQUALVERIFY")
 
-	// Generate blocks to reach reassignment height
-	td.MineAndWait(t, testReassignedUtxoSpendableAfter)
+	// Mine beyond the boundary, then wait for the UTXO store's asynchronous
+	// height update before attributing any rejection to the locking script.
+	td.MineAndWait(t, testReassignedUtxoSpendableAfter+1)
```
galt-tr, line 192: **[code-review 6/15]** Mining `testReassignedUtxoSpendableAfter+1` blocks removes the only assertion that the gate opens at exactly the configured height; the `require.Eventually` that follows already fixes the store-height race the +1 was added for.

Reassign at chain height 4 -> spendableIn = 9 (sql/alert_system.go:177). The old test mined 5 (height 9) and asserted Status_OK, which deterministically caught a `<=` or `+1` regression in sql.go:3446/:2151 (and was racy only toward false failures). The new test mines 6 (height 10), so a gate that opens one block late passes. Mine exactly 5 and rely on the Eventually; stronger still, `Eventually(store height == 4)` before `ReAssignUTXO`, then mine 5.

services/alert/node.go
```diff
@@ -450,7 +450,9 @@ func (n *Node) AddToConfiscationTransactionWhitelist(ctx context.Context, txs []
 				LockingScript: newLockingScript,                                          // new locking script
 			}
 
-			// the new utxo hash allows the original output to be spent by the confiscation transaction
+			// Calculate the replacement commitment. ReAssignUTXO does not persist
```
galt-tr, line 453: **[code-review 7/15]** Eight doc files now say ownership-changing reassignment 'must not be used', yet neither code site guards it: `handleReassign` reassigns unconditionally and returns `null`, and this confiscation path is by construction an ownership change (the new script is derived from the confiscating key) and still calls `ReAssignUTXO` with no warning.

handlers.go:2687-2693 never compares old/new hash; node.go:441 derives `newLockingScript` from the confiscation tx's pubkey, :469 calls `ReAssignUTXO`, and success is reported as absence from `NotProcessed` (:475) -> every processed confiscation alert strands the output while reporting success. Since UTXOHash = SHA256(txid||idx||script||sats) (util/utxo_hash.go:23-26), `!oldUtxoHash.IsEqual(newUtxoHash)` identifies an owner change exactly; refusing (or at minimum `n.logger.Warnf` with both hashes) at both sites leaves the UTXO frozen with its original hash - recoverable - instead of stranded. The node already holds the 'authoritative replacement script' the docs say is missing; persistence is the actual gap.

docs/topics/services/rpc.md
```diff
@@ -73,7 +73,7 @@ The below table summarises the services supported in the current version:
 | getminingcandidate        | Supported  | Returns data needed to construct a block to work on                          |
 | invalidateblock           | Supported  | Permanently marks a block as invalid                                         |
 | isbanned                  | Supported  | Checks if a network address is currently banned                              |
-| reassign                  | Supported  | Reassigns ownership of a specific UTXO to a new Bitcoin address              |
+| reassign                  | Restricted | Updates a frozen UTXO commitment; ownership changes currently strand the output |
```
galt-tr, line 76: **[code-review 8/15]** `Restricted` is a new, undefined status (every other row is Supported/Unimplemented) that reads as an access restriction, but `reassign` is in the `rpcLimited` map so the limited (non-admin) RPC user can still call it; the same section then restates the callout four times and the regression text is pasted into six files with drifting wording.

services/rpc/Server.go:288-339 lists `"reassign"` in `rpcLimited` and :1219-1225 rejects a limited user only for methods absent from that map -> an operator reading 'Restricted' assumes admin-only while the limited credentials can execute an owner-changing reassignment (handlers.go:2648 'Requires admin privileges' is false for the same reason). Use a defined status with the caveat in the description, or add a legend. Also: rpc.md:1094-1097 Important Notes repeat :1047-1049/:1056/:1065/:1080; docs/references/glossary.md:5-6 is the only free paragraph between `**Term**:` entries; the 'Known regression' paragraph appears in 6 files with modality drift (can leave / leaves / strands / unsafe) - GitHub rendering rules out mkdocs snippets on this branch, so keep one canonical paragraph in alert.md 2.4 and one-sentence pointers elsewhere; alert.md:159 hard-codes the 58-char Go test name this PR just renamed.

services/alert/node.go
```diff
@@ -450,7 +450,9 @@ func (n *Node) AddToConfiscationTransactionWhitelist(ctx context.Context, txs []
...
+			// the replacement script, so changing the owner currently strands
+			// the output for both owners after mandatory re-extension (issue 1725).
```
galt-tr, line 455: **[code-review 9/15]** The new inline comment contradicts the godoc of the same function (`AddToConfiscationTransactionWhitelist`, :361-372: 'allows authorized confiscation transactions to spend previously frozen UTXOs'), and the doc sweep missed several other statements that still promise spendability or a destination address.

Untouched and now contradictory: node.go:361-372 godoc; docs/topics/services/validator.md:257 ('must not be able to spend ... until 1,000 blocks', framed as implementation under 2.3); docs/references/services/alert_reference.md:196 'Re-assigns UTXOs to confiscation transactions, allowing them to be spent'; stores/utxo/Interface.go:48-50 constant comment 'blocks that must pass before a reassigned UTXO becomes spendable'; stores/utxo/sql/alert_system.go:147-149 and stores/utxo/aerospike/alert_system.go:215-221 method docs; docs/topics/stores/utxo.md:62 'to another (specified) address'; services/rpc/handlers.go:2654 godoc lists 3 of the 4 `ReassignCmd` fields (omits OldUTXOHash) and :2638-2641 still describes reorg/debug use cases; test/e2e/docker/utxo_test.go:11 still names `TestShouldAllowReassign`.

docs/topics/stores/utxo.md
```diff
@@ -391,12 +391,13 @@ The UTXO Store supports advanced UTXO management features, which can be utilized
...
     - If not frozen, it returns an error
 
-3. **Reassigning UTXOs**: UTXOs can be reassigned to a new transaction output, but only if they are frozen first.
+3. **Reassigning UTXOs**: Updates the commitment of a frozen output.
+    - **Known regression:** changing the owner leaves the output unspendable by both owners after mandatory re-extension, even after maturity. Do not use ownership-changing reassignment. See the [reassignment limitation](../services/alert.md#24-utxo-reassignment) and [issue 1725](https://github.com/bsv-blockchain/teranode/issues/1725).
     - Verifies the UTXO exists and is frozen
     - Updates the UTXO hash to the new value
-    - Sets spendable block height to current + 1,000 blocks (defined by `ReAssignedUtxoSpendableAfterBlocks` constant)
+    - Sets the maturity height to current + 1,000 blocks by default; SQL honors the configured override, while Aerospike currently always uses the constant
     - Logs the reassignment for audit purposes
```
galt-tr, line 399: **[code-review 10/15]** The kept bullet 'Logs the reassignment for audit purposes' is false for the SQL store: `ReAssignUTXO` is a bare UPDATE with no logger call and no audit table; only the Aerospike Lua UDF appends to a `reassignments` bin.

stores/utxo/sql/alert_system.go:150-192 has no logging and sql.go:4722-4733/:4913-4924 define no audit table; aerospike/teranode.lua:921-940 is the only audit trail; the RPC handler logs just '[handleReassign] called' with no identifiers (handlers.go:2664). An operator on SQLite/Postgres relying on this bullet to reconstruct a confiscation history finds nothing. This PR already deleted rpc.md's 'permanent record in the audit log' claim - do the same here or qualify it as Aerospike-only.

settings/utxostore_settings.go
```diff
@@ -45,7 +45,7 @@ type UtxoStoreSettings struct {
...
+	ReAssignedUtxoSpendableAfterBlocks      uint32            `key:"utxostore_reassignedUtxoSpendableAfterBlocks" desc:"SQL reassignment maturity delay in blocks" default:"1000" category:"UtxoStore" usage:"SQL maturity gate; passing it does not restore ownership-changing reassignment" type:"uint32" longdesc:"### Purpose\nDefines the SQL reassignment maturity delay. Aerospike currently ignores this setting and uses the fixed 1,000-block delay.\n\n### How It Works\nSQL holds a reassigned UTXO until this many blocks have passed. Mandatory re-extension currently leaves an ownership-changing reassignment unspendable by both owners even after maturity, because the replacement script is not stored. Do not use ownership-changing reassignment; see https://github.com/bsv-blockchain/teranode/issues/1725.\n\n### Values\n- **Default 1000** - Production value providing safety margin for reassignment operations\n- **Lower values** - Only for testing purposes\n\n### Warning\n**DO NOT reduce in production** - this provides important security guarantees for the alert system."`
```
galt-tr, line 48: **[code-review 11/15]** Relabeling the setting 'SQL reassignment maturity delay' documents an Aerospike divergence in six places that is a four-line fix (`tSettings` is already in scope and the Lua `reassign` UDF already takes `spendableAfter`), the new `desc` is surfaced verbatim in the dashboard settings page, and 'SQL honors the configured delay' is itself false for an explicit 0.

stores/utxo/aerospike/alert_system.go:249 passes the constant `utxo.ReAssignedUtxoSpendableAfterBlocks` while :253 already uses `tSettings`; teranode.lua:864/:942 parameterize `spendableAfter`; sql/alert_system.go:173-176 is the template (git blame: the omission predates the setting; alert_system2_test.go:268 still passes with the fix). settings/export.go:139 -> asset settings_handler.go:92 -> ui/dashboard settings page shows a backend-neutral key as SQL-only. On the same lines and in Interface.go:366/utxo_reference.md:244/utxo.md:398/rpc.md:1096: sql/alert_system.go:172-176 only honors values `> 0` and silently falls back to 1000, so a tester setting 0 for immediate spendability gets a 1000-block gate. Finally, the key is absent from the operator-facing settings reference: docs/references/settings/stores/utxo_settings.md (mkdocs.yml:220) has no row for `utxostore_reassignedUtxoSpendableAfterBlocks` (grep over docs/references/settings/ returns nothing), although alert.md:164 now sends operators to it by name and no page links to that reference; the file is hand-written (no generator - the only longdesc consumer is settings/metadata.go:11 served at runtime), so add a row with the new SQL-only caveat here (18 utxostore_* keys are missing from that page and 6 listed keys no longer exist - pre-existing drift the PR makes load-bearing for this key).

docs/topics/services/rpc.md
```diff
@@ -1041,59 +1041,60 @@ The `isbanned` command checks if a specific network address is currently banned
...
+    - `newutxohash` (string, required): Replacement UTXO commitment, as a hex hash
 
 - **Return Value**:
 
-    - On success: Returns `true` indicating the UTXO was successfully reassigned
-    - On failure: Returns an error if the UTXO cannot be found or reassigned
+    - On success: JSON `null`; this confirms the store operation, not spendability
+    - On failure: An error if a hash cannot be parsed or the store operation fails
 
 #### Process Flow
```
galt-tr, line 1068: **[code-review 12/15]** Removing the `![rpc-reassign.svg]` embed orphans `docs/topics/services/img/plantuml/rpc/rpc-reassign.puml/.svg`, which still render the wrong contract (`ReassignCmd(txid, vout, destination)`, `Return true`, 'Original script replaced with new script') and remain served by gh-deploy.

No remaining reference in docs/, mkdocs.yml, Makefile, scripts or .github; puml :29/:47/:76/:88/:91/:94-95 assert the destination-address flow this PR says does not exist; no strict mkdocs/link check flags orphans, so the stale diagram is published indefinitely. Delete both files or regenerate the diagram for the hash-based contract - 24 of 26 RPC sections keep a diagram, so this section is now also the odd one out.

test/e2e/daemon/ready/reassign_test.go
```diff
@@ -41,9 +44,31 @@ func TestShouldAllowReassign(t *testing.T) {
 	})
 
 	defer td.Stop(t)
+	require.Equal(t, "sqlite", td.Settings.UtxoStore.UtxoStore.Scheme)
+
+	// Public ingress redacts some validation errors. Pair every rejection
+	// there with its exact cause from a real validator over the same stores.
+	// Nil Kafka producers and a disabled assembly handoff keep this probe local.
+	v, err := validator.New(td.Ctx, td.Logger, td.Settings, td.UtxoStore,
```
galt-tr, line 52: **[code-review 13/15]** The test hand-builds a second in-process validator with three nil dependencies although `NewTestDaemon` already constructs the daemon's local validator over the same UTXO store (daemon/test_daemon.go:549 `GetValidatorClient`) and simply does not expose it.

settings.conf:1212 `useLocalValidator=true` -> daemon_stores.go:194-248 builds `validator.New` in-process with an identical error chain (WrapGRPC never applies), so a `td.ValidatorClient` field populated from :549 gives the same exact reasons. The ad-hoc instance allocates a second BDK cgo engine (ScriptVerifierGoBDK.go:151), shares the gocore 'validator' stat, has a nil blockAssembler that Validator.go:1401 dereferences if AddTXToBlockAssembly is ever true, runs with SkipUtxoCreation=false so an unexpectedly passing probe spends/creates in the shared store, and re-runs the immature probe 4x through the TX_LOCKED retry loop (Validator.go:432-452 - if a private validator is kept, build it from a settings copy with `Validator.TxLockedMaxRetries = 0`).

test/e2e/daemon/ready/reassign_test.go
```diff
@@ -41,9 +44,31 @@ func TestShouldAllowReassign(t *testing.T) {
...
+		nil, nil, nil, td.BlockchainClient)
+	require.NoError(t, err)
+	requireRejected := func(tx *bt.Tx, reason string) {
+		t.Helper()
+		// Snapshot before either path runs, and decode separate requests.
+		// Keep tx untouched so retries still supply the original extended fields.
+		submittedBytes := tx.ExtendedBytes()
```
galt-tr, line 59: **[code-review 14/15]** The `ingress` copy is dead work - the propagation client only serializes the tx - and the `probe` copy, the one that must be isolated from the validator's in-place re-extension, re-implements `(*bt.Tx).Clone()`.

Only Validator.go:1046-1047/:539/:1462 mutate the tx (PreviousTxScript/Satoshis, SetTxHash); test_daemon.go:541 wires a gRPC propagation client whose batcher stores the pointer but only calls `SerializeBytes` (Client.go:218-258). go-bt v2.6.5 tx.go:480/:485 `Clone` copies PreviousTxSatoshis/PreviousTxScript so the clone stays extended (precedent services/validator/Server_test.go:166). Six lines become `require.Error(t, td.PropagationClient.ProcessTransaction(td.Ctx, tx)); probe := tx.Clone()`.

test/e2e/daemon/ready/reassign_test.go
```diff
@@ -158,24 +185,41 @@ func TestShouldAllowReassign(t *testing.T) {
...
-	err = td.PropagationClient.ProcessTransaction(td.Ctx, charlesSpendingTx)
-	require.Error(t, err, "an unstored replacement script must not be accepted")
+	requireRejected(charlesSpendingTx, "OP_EQUALVERIFY")
 
-	// Generate blocks to reach reassignment height
-	td.MineAndWait(t, testReassignedUtxoSpendableAfter)
+	// Mine beyond the boundary, then wait for the UTXO store's asynchronous
+	// height update before attributing any rejection to the locking script.
+	td.MineAndWait(t, testReassignedUtxoSpendableAfter+1)
+	require.Eventually(t, func() bool {
```
galt-tr, line 193: **[code-review 15/15]** The maturity wait polls two SQL `GetSpend` rows every 100 ms although both outputs were reassigned back-to-back at the same store height, the status is a pure function of the store's atomic block height, and the closure folds a persistent `GetSpend` error into a generic 30 s timeout.

reassign_test.go:152/:155 call `ReAssignUTXO` with no block between, so both get spendableIn = H+5 (sql/alert_system.go:176) and the later successful bob spend already proves the second; `GetSpend` derives IMMATURE from `s.GetBlockHeight() < *spendableIn` (sql.go:3445-3447, atomic :276) -> poll `td.UtxoStore.GetBlockHeight() >= target` (no I/O) then `GetSpend` once, and fail fast on a non-nil error (e.g. UTXO_MISMATCH from sql.go:3417) instead of reporting 'both outputs must mature'. No helper waits on store height (`MineAndWait` only polls the blockchain client, test_daemon.go:1346-1424); a `td.WaitForUtxoStoreHeight` beside `WaitForBlockhash` (:1602) would serve the Aerospike variant the docs say is missing. The `require.Equal("sqlite", ...)` at :47 restates test/test_settings.go:62 unconditionally.

services/legacy/netsync/manager.go
```diff
@@ -2355,19 +2370,22 @@ func (sm *SyncManager) Start() {
...
+	// The peer handler and the context cancellation hook can both call Stop.
+	sm.stopOnce.Do(func() {
+		sm.inventoryMu.Lock()
+		atomic.StoreInt32(&sm.shutdown, 1)
+		sm.inventoryMu.Unlock()
+
+		sm.logger.Infof("Sync manager shutting down")
+		close(sm.quit)
+		<-sm.handlerDone
+		sm.inventoryWG.Wait()
```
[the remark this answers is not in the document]
icellan: Resolved by the release/v0.15 merge: the release implementation admits and drains transaction, header, inventory and Kafka block handlers through the shared handler group, and joins the sequential block worker. The merge retains that implementation and the PR’s internal-server join. Shutdown replies and pause handling also observe quit.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

services/legacy/netsync/manager_shutdown_test.go
```diff
@@ -0,0 +1,107 @@
...
+	entered chan struct{}
+	release chan struct{}
+	reads   atomic.Int32
+}
+
+func (s *heldInventoryStore) Get(ctx context.Context, hash *chainhash.Hash, fieldNames ...fields.FieldName) (*meta.Data, error) {
+	if s.reads.Add(1) == 1 {
+		close(s.entered)
+		<-s.release
+	}
```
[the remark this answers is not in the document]
icellan: Replaced the pre-Get wrapper with a real SQL boundary: the test reserves the only database connection, waits for database/sql WaitCount to increase from the batcher’s query, then verifies both Stop callers remain blocked until SQL is available. It closes the real store before testing late handlers. The unit test covers the batcher/manager boundary; the BIP68 smoke tests exercise the real service startup and shutdown. The merged release also makes Server.Stop join the sync manager.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

daemon/test_daemon.go
icellan: Fixed the container-manager URL to sqlitememory:// and added a test that initializes the advertised SQLite backend, constructs the real UTXO store, and pings its database. The TestOptions comment now distinguishes explicit in-memory SQLite from an empty option preserving configured file-backed SQLite.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

docs/topics/services/alert.md
icellan: Documented both SQL freezing and SQL reassignment omitting the supplied old-commitment check, with Aerospike checking it. Updated the RPC reference and miner guide parameter descriptions, and documented zero-padding of short hashes and the unchecked uint32 index cast. The canonical backend table is in the Alert Service reassignment section.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

test/e2e/daemon/ready/reassign_test.go
icellan: The local probe now passes height 0, exactly like ingress, so both select the candidate height (tip+1). It uses the daemon’s existing local validator.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

test/e2e/daemon/ready/reassign_test.go
icellan: The ingress assertion now requires the public failed-to-validate-transaction error. The helper takes the spend and checks SpendingData after every rejected attempt, including the previously omitted pre-maturity replacement-owner attempt. Its comment now explains that the validator error chain is removed by public ingress.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

test/e2e/daemon/ready/reassign_test.go
icellan: Pinned the SQL maturity and changed-commitment failures with ErrorIs(ErrTxLocked) and ErrorIs(ErrUtxoHashMismatch). OP_EQUALVERIFY retains its specific script-error assertion. The canonical backend table now documents the SQL ErrTxLocked versus Aerospike ErrUtxoFrozen difference.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

test/e2e/daemon/ready/reassign_test.go
icellan: Pinned the store’s starting height before reassignment, mined exactly the configured five blocks, and waited for the exact maturity height and Status_OK. This retains the asynchronous-height wait while detecting a gate that opens a block late.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

services/alert/node.go
icellan: The warning is operational guidance, not an enforced guard; the docs and function comments now say so explicitly and state that the alert path derives but does not persist the replacement script. Runtime refusal/logging changes remain outside this PR under the author’s explicit tests/documentation-only scope for reassignment. I recorded entry-point guards and recovery in issue #1725, along with the limited-credential RPC exposure.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

docs/topics/services/rpc.md
icellan: Changed the table status to Supported with the limitation in its description, and documented that limited RPC credentials can call the method. Consolidated the long warning and backend differences in alert.md, replaced repeated paragraphs with short pointers, folded the glossary warning into its term entry, and linked the test file instead of hard-coding the function name.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

services/alert/node.go
icellan: Corrected the confiscation godoc, validator and alert references, exported delay constant, concrete store comments, UTXO overview and RPC handler godoc. Removed the obsolete Docker-suite command (that test no longer exists in that suite). The RPC comment now includes all four fields and no longer claims admin-only access or reorg-repair behavior.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

docs/topics/stores/utxo.md
icellan: Qualified the audit claim: Aerospike appends to its reassignments bin; SQL has no reassignment audit record. This distinction also appears in the canonical backend table.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

settings/utxostore_settings.go
icellan: Corrected the setting metadata and references: SQL uses positive values and zero falls back to 1,000; Aerospike currently ignores the setting. The short setting description is backend-neutral. Changing Aerospike’s runtime maturity behavior remains deferred to issue #1725 under the agreed reassignment scope, rather than silently changing consensus-adjacent behavior in this test/documentation follow-up. Added the missing operator settings-reference row and linked it from the canonical backend table in 552336da3.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

docs/topics/services/rpc.md
icellan: Rewrote and regenerated the RPC PlantUML/SVG for the four-argument hash contract, JSON null success, actual parameter handling, and missing script persistence, then restored its embed. Also updated and regenerated the other affected reassignment/state-management diagrams, including the one used by the transaction-lifecycle page, and visually inspected the rendered results.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

test/e2e/daemon/ready/reassign_test.go
icellan: Exposed the validator already created by NewTestDaemon as ValidatorClient and used it for the probe, with local validation selected explicitly in the fixture. Removed the extra validator.New call and its nil dependencies. The probe retains the spend-validation path needed to assert the commitment failure; any unexpected success fails the test.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

test/e2e/daemon/ready/reassign_test.go
icellan: Removed the redundant ingress decode and replaced the probe’s encode/decode copy with tx.Clone(). The ingress client serializes the fixture; only the local validator receives a mutable clone.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

test/e2e/daemon/ready/reassign_test.go
icellan: Added WaitForUtxoStoreHeight to the daemon test helpers. The reassignment test now polls only the atomic height, asserts the exact target, then reads each output once with immediate error/status assertions. The explicit SQLite assertion remains as a scope guard so a future default-backend change cannot silently turn this into an Aerospike test.

Pushed in c1d6bd51d and 552336da3 (release merge: 2d082b403).

icellan:
Merged the latest `release/v0.15` into this branch in 2d082b403 and pushed review fixes in c1d6bd51d and 552336da3. Conflict resolution retains the release branch's broader handler drain and the PR's internal-server join.

All 17 new inline comments have replies. Fifteen threads are resolved. The two production reassignment requests remain open: entry-point guards and Aerospike maturity behavior are tracked in #1725 under the agreed tests/documentation-only reassignment scope. The missing operator settings row, backend differences, RPC contract, and diagrams are now corrected. Review is requested from galt-tr; oskarszoon's review request remains pending.

Local verification passed: three race-enabled reassignment runs; five shutdown regression runs; race-enabled model, legacy, block-validation, daemon, container-helper and settings packages; vet, staticcheck, lint, Markdown and rendered-diagram checks. The SQL regression now waits at the actual database connection pool and verifies both concurrent Stop callers drain the read.

CI passed all three smoke shards, all seven sequential shards, legacy-sync, pruner, chain-integrity, unit tests and lint on `552336da3` ([smoke/integration](https://github.com/bsv-blockchain/teranode/actions/runs/34627481765), [unit/lint](https://github.com/bsv-blockchain/teranode/actions/runs/34627481792)).

Copilot subsequently identified that direct `Server.Stop` callers also need the internal-server join. Fixed and pushed in `1e08f2bb8`, with a regression that fails on the previous implementation and passes five race-enabled runs. The complete legacy and daemon race suites, vet, staticcheck and lint pass. That additional thread is replied to and resolved; only the two production reassignment follow-ups remain open.

Fresh CI for final head `1e08f2bb8` is running: [smoke/integration](https://github.com/bsv-blockchain/teranode/actions/runs/34629190313), [unit/lint](https://github.com/bsv-blockchain/teranode/actions/runs/34629190506). The branch is mergeable, the worktree is clean, and review requests remain pending for oskarszoon and galt-tr.

services/legacy/Server.go
```diff
@@ -641,7 +641,17 @@ func (s *Server) Start(ctx context.Context, readyCh chan<- struct{}) error {
...
+	go func() {
+		defer close(serverDone)
+		s.server.Start()
+	}()
+	defer func() {
+		_ = s.server.Stop()
+		// The daemon closes stores after service Start methods return. Keep
+		// those stores alive until the peer handler and sync manager exit.
+		<-serverDone
+	}()
```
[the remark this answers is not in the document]
icellan: Fixed in 1e08f2bb8. Start now publishes its cancellation function and completion channel under a lifecycle mutex before waiting on the FSM. Every Stop caller cancels and joins that same invocation, including the existing deferred internal-server join. Stop-before-Start prevents later startup.

Added a regression using the real legacy service and a SQLite blockchain store. It holds the internal server's WaitForShutdown open after the sync-manager drain and verifies that two concurrent Stop callers both wait, then checks that Start has exited and restart is rejected. The test fails on the previous implementation and passes five race-enabled runs with the fix. The complete legacy and daemon race suites, vet, staticcheck, and repository lint also pass.

oskarszoon (approved):
Reviewed at `1e08f2bb8`. Approving — everything blocking is closed, and the two things still open in code are both deferred to #1725 by agreement.

Worth saying plainly up front: most of what I'd have raised, galt-tr raised first and you'd already fixed by the time I looked. I verified all fifteen of his items against this HEAD; twelve are genuinely fixed, and nothing he filed was wrong. I checked the two highest-stakes ones by execution rather than by reading:

- **The `sqlite` store-type panic was real and is fixed.** `TestOptions{UTXOStoreType: "sqlite"}` did panic at `util/sql.go:221` on the empty path of `sqlite://memory:`. `container_manager.go:322` is now `sqlitememory://`, and the new `container_manager_test.go` builds the real store and pings it.
- **The maturity gate is mutation-proven.** Flipping the SQL spend gate `<`→`<=` at `sql.go:2151/2653/3446` fails `TestReassignSQLiteEnforcesMaturityAndRejectsUnstoredScript` at `reassign_test.go:203`, with the baseline green. The exact-height assertion that the earlier `+1` had removed is genuinely back.

On the shutdown side, the picture changed under the merge and it's worth recording why. `services/legacy/netsync/manager.go` is now **0 changed lines** against the new merge-base — the drain landed on `release/v0.15` separately, in a better form than the version in this branch's own commit. `beginHandler()` (`manager.go:2507-2515`) takes `handlerMu` and returns false during shutdown, and four call sites pair it with `defer sm.handlers.Done()`: `handleTxMsg` (`:1063`), `handleHeadersMsg` (`:1761`), `handleInvMsg` (`:1958`) and orphan eviction (`:2633`). The block-queue consumer has its own `blockWorker` joined before `handlerDone` closes, and the `blockQueue` send is now guarded by `select` on `sm.quit` (`:2294-2305`).

That matters because the version in `0d4bee483` tracked only `handleInvMsg`, which left `handleTxMsg` → `validationClient.Validate` and the block consumer → `utxoStore.Create`/`SetMinedMulti` able to outlive `Stop`. The superseding mechanism covers all of them. The reworked `TestStop_WaitsForInventoryBeforeClosingStore` was correctly adapted too — removing `sm.handlers.Wait()` fails it.

The godoc rewrite at `handlers.go` is better than what I'd have asked for: it names the actual failure mode, links the issue, and fixes the parameter list to include `OldUTXOHash`.

Non-blocking, in rough priority:

The `1e08f2bb8` lifecycle rework closes the last thing I had. `Stop` now sets `stopping`, cancels the context to unblock an outstanding FSM wait, and defers `<-done` — cancel-then-join, which is the right shape and better than the bare unbounded wait it replaces. Coverage on the lifecycle block at `Server.go:635-655` is now `1` on every statement via the new `server_shutdown_test.go`.

On the latest automated review: both of its points don't hold against this HEAD. `Stop` *does* observe the lifecycle signal — it captures `done` under `lifecycleMu` and defers the join, so it runs before return on every path including the nil-server early return, and the "safe not-started path" it asks for is the `if done != nil` guard. And the startup barrier it says was removed was never there: `d33dfb91d:644` is already `go s.server.Start()`. The readiness concern underneath it may be worth its own look, but it's pre-existing.

Non-blocking, in rough priority:

- **`stores/txmetacache/txmetacache.go:947`** — "reassigns a UTXO from one transaction to another". Wrong independent of the regression; the outpoint never changes.
- **`openapi/rpc_openrpc.json:649`** — still "Moves a frozen UTXO designation to a different output" with no caveat. The params are right, so nothing contradicts, but this is the machine-readable contract operator tooling consumes.
- **`reassign_test.go` has no reference to #1725** despite `:206` and `:217` encoding the lockout as expected behaviour. A real fix will turn this red, and the name reads as intended behaviour — one comment naming the issue and saying "invert when fixed" would save someone an afternoon.
- `alert.md:175` links the test by branch rather than permalink, so it will drift.

galt-tr (approved):
Reviewed the diff and relevant callers at 1e08f2bb8. No actionable defects found. The Start/Stop synchronization cancels and joins the active service and internal-server cleanup for every Stop caller. The reassignment regression checks both owners' rejection, unspent outputs, and the configured maturity boundary; the documentation reflects the existing implementation and clearly identifies the unresolved ownership-changing reassignment limitation tracked in issue 1725.

Local validation passed:
- go test -race -tags testtxmetacache ./services/legacy ./services/legacy/netsync ./test/utils/containers -run 'TestStop_|TestSQLiteURLCreatesUsableStore|TestHeaderProvenance' -count=1
- go test -race -tags testtxmetacache ./test/e2e/daemon/ready -run '^TestReassignSQLiteEnforcesMaturityAndRejectsUnstoredScript$' -count=1
- git diff --check

All reported PR CI checks are green. Full-suite and Aerospike reassignment tests were not rerun locally.

PR #1727, opened 2026-09-11
fix: authenticate catchup bodies and complete legacy shutdown

icellan:
Quick validation could mutate UTXOs before discovering malformed transaction data in a later subtree. It now authenticates the complete checkpoint-proven body before assigning a block ID or starting mutation, and completes legacy peer-server shutdown.

- Recompute subtree roots directly from nodes, verify the cached roots and requested hashes, coinbase placeholder, duplicates, block merkle root and declared transaction count, then parse every transaction body. Merkle validation is the authoritative subtree-length check.
- Honor subtree-read concurrency and mmap configuration. Batches borrow authenticated nodes until all readers finish. Root recomputation avoids cloning mmap-backed nodes. Transaction data is decoded again during processing to keep retained memory bounded.
- Track files created by each catchup attempt. Invalid bodies discard that attempt's pending files after workers join, preserving shared files and promoted subtree data. Transient failures retain download progress.
- Distinguish corrupt cached files from bad downloads made in the current attempt, including `.subtreeToCheck` files retained from earlier attempts. Cached corruption aborts as a local service/storage failure without peer penalties or normal-validation fallback; only the identified corrupt files are evicted after workers join, allowing a later attempt to refetch them. Preflight failures on newly downloaded malformed files retain the invalid-body verdict.
- Bound cleanup using subtree-fetch concurrency (default eight), a five-second per-file timeout and a one-minute overall deadline. Aggregate failures into one warning and retain files whose deletion did not complete.
- Register and join legacy workers before shutdown can wait for them. Ban-event query sends and reply waits now honor shutdown/context cancellation, and buffered replies prevent late responses from blocking the peer handler. The Start reordering also activates UPnP maintenance and the ban-event listener during normal runtime; on the release branch they previously launched only after the synchronous peer handler exited.
- Preserve the catchup-after-pruning smoke fixture's small HTTP burst allowance; production rate limits remain unchanged. Documentation now distinguishes invalid-body aborts, cached-file recovery, incomplete blocks and normal-validation fallback.

Validation for review follow-up `797157b1c`:

- TDD: new shutdown, corrupt-cache/retry and concurrent-cleanup regressions failed on the original implementation before the fixes.
- Focused legacy/blockvalidation regressions passed twice with `-race -tags testtxmetacache`, covering malformed bodies, cached roots, mmap ownership/fallback, retry and shared-file preservation, cleanup timeouts, late replies and listener shutdown.
- Normal legacy package suites passed. The full normal blockvalidation package passed in the repository-wide run (393.458s). An earlier Kafka test's 500 ms processing assertion failed locally and also reproduced on the unmodified PR head under the same configuration.
- Affected-package vet, scoped diff lint, `staticcheck ./...`, formatting and all commit hooks passed.
- A direct 65,536-leaf Merkle comparison produced identical roots and reduced allocation from 7,366,822 to 2,110,379 B/op. This is a local primitive benchmark, not a full mainnet or multi-GB sync benchmark.
- Repository-wide validation is not all green: full vet reports four lock-copy warnings in unchanged `test/utils` code; full lint reports 50 preallocation warnings in unchanged code; gosec reports 746 findings, none on added/modified lines; govulncheck reports 48 reachable Go/dependency vulnerabilities. Dependencies are unchanged.
- Full normal/race repository runs reached failures outside this diff, and remaining broad integration work was stopped: SQL UTXO pruning assertions in the normal run and a `DiskTxMapUint64.writerLoop`/`Stats` race in the race run. The full affected-service race suite also hit that same model race while all legacy suites passed. Both the model regression and the blockvalidation duplicate-transaction regression reproduce the race on the unmodified PR head through a Go source overlay. The focused new regressions remain green; full-suite verification is not green.
- Frontend checks were attempted but lack installed Playwright/Prettier/Vite dependencies and a `typecheck` script; no frontend files changed. GitHub CI for the pushed head is still running.

### DIFF
--- a/services/blockvalidation/catchup.go
+++ b/services/blockvalidation/catchup.go
@@ -844,6 +844,10 @@ func (u *Server) fetchAndValidateBlocks(ctx context.Context, catchupCtx *Catchup
 	}
 	defer u.restoreFSMState(ctx, catchupCtx)
 
+	// Carry ownership through tracing and detached subtree-data downloads.
+	artifacts := &catchupArtifacts{Store: u.subtreeStore, files: make(map[catchupArtifactKey]*catchupArtifact)}
+	ctx = context.WithValue(ctx, catchupArtifactsKey{}, artifacts)
+
 	// Create error group for concurrent operations
 	errorGroup, gCtx := errgroup.WithContext(ctx)
 
@@ -879,6 +883,7 @@ func (u *Server) fetchAndValidateBlocks(ctx context.Context, catchupCtx *Catchup
 	// Wait for both operations to complete
 	err := errorGroup.Wait()
 	if err != nil {
+		artifacts.cleanup(u.logger)
 		catchupCtx.catchupError = err
 	}
 
@@ -1197,6 +1202,8 @@ func (u *Server) tryQuickValidation(ctx context.Context, block *model.Block, cat
 				}
 			}
 		}
+		// Force normal validation to reload metadata after a partial quick attempt.
+		block.SubtreeSlices = nil
 		// Quick validation failed, try normal validation
 		return true, nil
 	}
--- a/services/blockvalidation/catchup.go
+++ b/services/blockvalidation/catchup.go
@@ -883,7 +883,11 @@ func (u *Server) fetchAndValidateBlocks(ctx context.Context, catchupCtx *Catchup
 	// Wait for both operations to complete
 	err := errorGroup.Wait()
 	if err != nil {
-		artifacts.cleanup(u.logger)
+		// Discard rejected bodies, but retain partial downloads after transient
+		// failures (such as peer rate limits) so the next attempt can progress.
+		if errors.Is(err, errors.ErrBlockInvalid) || errors.Is(err, errors.ErrTxInvalid) {
+			artifacts.cleanup(u.logger)
+		}
 		catchupCtx.catchupError = err
 	}
 
--- a/services/blockvalidation/catchup_artifacts.go
+++ b/services/blockvalidation/catchup_artifacts.go
@@ -1,0 +1,94 @@
+package blockvalidation
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/bsv-blockchain/go-bt/v2/chainhash"
+	"github.com/bsv-blockchain/teranode/errors"
+	"github.com/bsv-blockchain/teranode/pkg/fileformat"
+	"github.com/bsv-blockchain/teranode/stores/blob"
+	"github.com/bsv-blockchain/teranode/stores/blob/options"
+	"github.com/bsv-blockchain/teranode/ulogger"
+)
+
+type catchupArtifactsKey struct{}
+
+type catchupArtifactKey struct {
+	hash chainhash.Hash
+	kind fileformat.FileType
+}
+
+type catchupArtifact struct {
+	mu      sync.Mutex
+	created bool
+}
+
+// catchupArtifacts owns only pending files created by one catchup attempt.
+// Validation continues to use the shared store and can promote accepted subtrees.
+// Cleanup must run after all fetchers and async validation writers have joined.
+type catchupArtifacts struct {
+	blob.Store
+	mu    sync.Mutex
+	files map[catchupArtifactKey]*catchupArtifact
+}
+
+func (u *Server) catchupSubtreeStore(ctx context.Context) blob.Store {
+	if store, ok := ctx.Value(catchupArtifactsKey{}).(*catchupArtifacts); ok {
+		return store
+	}
+	return u.subtreeStore
+}
+
+func (s *catchupArtifacts) Set(ctx context.Context, key []byte, kind fileformat.FileType, value []byte, opts ...options.FileOption) error {
+	hash, err := chainhash.NewHash(key)
+	if err != nil {
+		return err
+	}
+	file := catchupArtifactKey{hash: *hash, kind: kind}
+	s.mu.Lock()
+	entry := s.files[file]
+	if entry == nil {
+		entry = &catchupArtifact{}
+		s.files[file] = entry
+	}
+	s.mu.Unlock()
+	// Multiple prefetched blocks can share a subtree. Serialize only writes to
+	// that file, leaving independent subtree downloads and writes concurrent.
+	entry.mu.Lock()
+	defer entry.mu.Unlock()
+	opts = append(opts, options.WithAllowOverwrite(false))
+	if err := s.Store.Set(ctx, key, kind, value, opts...); err != nil {
+		if errors.Is(err, errors.ErrBlobAlreadyExists) {
+			return nil // preserve existing data, and never claim ownership of it
+		}
+		return err
+	}
+	entry.created = true
+	return nil
+}
+
+func (s *catchupArtifacts) cleanup(logger ulogger.Logger) {
+	// The fetch errgroup's context is canceled on failure (and after Wait).
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+	for file, entry := range s.files {
+		if !entry.created {
+			continue
+		}
+		// A previous block in this attempt, or another validator, may already
+		// have accepted this subtree. Keep its shared transaction data intact.
+		validated, err := s.Store.Exists(ctx, file.hash[:], fileformat.FileTypeSubtree)
+		if err != nil {
+			logger.Warnf("[catchup] cannot check artifact %s before cleanup: %v", file.hash.String(), err)
+			continue
+		}
+		if validated {
+			continue
+		}
+		if err := s.Store.Del(ctx, file.hash[:], file.kind); err != nil && !errors.Is(err, errors.ErrNotFound) {
+			logger.Warnf("[catchup] failed to remove pending %s for %s: %v", file.kind, file.hash.String(), err)
+		}
+	}
+}
--- a/services/blockvalidation/get_blocks.go
+++ b/services/blockvalidation/get_blocks.go
@@ -444,12 +444,13 @@ func (u *Server) fetchAndStoreSubtree(ctx context.Context, block *model.Block, s
 	)
 	defer deferFn()
 
+	store := u.catchupSubtreeStore(ctx)
 	dah := block.Height + u.settings.GetSubtreeValidationBlockHeightRetention()
 
 	// Check if we already have the subtree, under either FileTypeSubtreeToCheck
 	// (peer-fetched, pending validation) or FileTypeSubtree (already validated).
 	// See findLocalSubtreeFile for why both must be consulted.
-	localFileType, localExists, err := findLocalSubtreeFile(ctx, u.subtreeStore, *subtreeHash)
+	localFileType, localExists, err := findLocalSubtreeFile(ctx, store, *subtreeHash)
 	if err != nil {
 		return nil, errors.NewStorageError("[catchup:fetchAndStoreSubtree] error checking subtree existence for %s", subtreeHash.String(), err)
 	}
@@ -458,7 +459,7 @@ func (u *Server) fetchAndStoreSubtree(ctx context.Context, block *model.Block, s
 		u.logger.Debugf("[catchup:fetchAndStoreSubtree] Subtree already exists for %s, loading from store", subtreeHash.String())
 
 		// Load existing subtree from store under whichever file type was found
-		subtreeBytes, err := u.subtreeStore.Get(ctx, subtreeHash[:], localFileType)
+		subtreeBytes, err := store.Get(ctx, subtreeHash[:], localFileType)
 		if err != nil {
 			return nil, errors.NewStorageError("[catchup:fetchAndStoreSubtree] Failed to get existing subtree for %s", subtreeHash.String(), err)
 		}
@@ -517,7 +518,7 @@ func (u *Server) fetchAndStoreSubtree(ctx context.Context, block *model.Block, s
 	}
 
 	// Store subtree (for subtreeToCheck) in subtreeStore
-	if err = u.subtreeStore.Set(ctx,
+	if err = store.Set(ctx,
 		subtreeHash[:],
 		fileformat.FileTypeSubtreeToCheck,
 		subtreeBytes,
@@ -541,10 +542,11 @@ func (u *Server) fetchAndStoreSubtreeData(ctx context.Context, block *model.Bloc
 	)
 	defer deferFn()
 
+	store := u.catchupSubtreeStore(ctx)
 	dah := block.Height + u.settings.GetSubtreeValidationBlockHeightRetention()
 
 	// Check if we already have the subtreeData
-	subtreeDataExists, err := u.subtreeStore.Exists(ctx, subtreeHash[:], fileformat.FileTypeSubtreeData)
+	subtreeDataExists, err := store.Exists(ctx, subtreeHash[:], fileformat.FileTypeSubtreeData)
 	if err != nil {
 		return errors.NewProcessingError("[catchup:fetchAndStoreSubtreeData] Error checking subtreeData existence for %s: %v", subtreeHash.String(), err)
 	}
@@ -606,7 +608,7 @@ func (u *Server) fetchAndStoreSubtreeData(ctx context.Context, block *model.Bloc
 	}
 
 	// Store subtreeData (raw data) in subtreeStore
-	if err = u.subtreeStore.Set(ctx,
+	if err = store.Set(ctx,
 		subtreeHash[:],
 		fileformat.FileTypeSubtreeData,
 		subtreeDataBytes,
--- a/services/blockvalidation/quick_validate.go
+++ b/services/blockvalidation/quick_validate.go
@@ -394,7 +394,9 @@ func (u *BlockValidation) processBlockSubtrees(ctx context.Context, block *model
 // This is the fallback when SubtreeBatchPrefetchDepth is 0.
 func (u *BlockValidation) processBlockSubtreesSequential(ctx context.Context, block *model.Block) (uint64, error) {
 	numSubtrees := len(block.Subtrees)
-	block.SubtreeSlices = make([]*subtreepkg.Subtree, numSubtrees)
+	if len(block.SubtreeSlices) != numSubtrees {
+		block.SubtreeSlices = make([]*subtreepkg.Subtree, numSubtrees)
+	}
 	var existingBlockID uint64
 
 	// Get block ID first (check for retry using first tx after reading first batch)
@@ -475,7 +477,9 @@ func (u *BlockValidation) processBlockSubtreesSequential(ctx context.Context, bl
 //   - On retry, Create() returns ErrTxExists, and SetMinedMulti() updates the correct BlockID
 func (u *BlockValidation) processBlockSubtreesPipeline(ctx context.Context, block *model.Block, prefetchDepth int) (uint64, error) {
 	numSubtrees := len(block.Subtrees)
-	block.SubtreeSlices = make([]*subtreepkg.Subtree, numSubtrees)
+	if len(block.SubtreeSlices) != numSubtrees {
+		block.SubtreeSlices = make([]*subtreepkg.Subtree, numSubtrees)
+	}
 	var existingBlockID uint64
 	blockIDSet := false
 
@@ -617,7 +621,9 @@ func (u *BlockValidation) processBlockSubtreesPipeline(ctx context.Context, bloc
 //   - error: If processing fails or context is cancelled
 func (u *BlockValidation) processBlockSubtreesPipelineAsync(ctx context.Context, block *model.Block, prefetchDepth int, writeJobsChan chan<- *SubtreeWriteJob) (uint64, error) {
 	numSubtrees := len(block.Subtrees)
-	block.SubtreeSlices = make([]*subtreepkg.Subtree, numSubtrees)
+	if len(block.SubtreeSlices) != numSubtrees {
+		block.SubtreeSlices = make([]*subtreepkg.Subtree, numSubtrees)
+	}
 	var existingBlockID uint64
 	blockIDSet := false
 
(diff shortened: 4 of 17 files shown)


services/blockvalidation/quick_validate.go
```diff
@@ -1556,5 +1569,24 @@ func (u *BlockValidation) authenticateQuickBlockBody(ctx context.Context, block
...
+	if block.TransactionCount != authenticatedCount {
+		return errors.NewBlockInvalidError("[quickValidateBlock][%s] transaction count %d does not match authenticated body count %d", block.Hash().String(), block.TransactionCount, authenticatedCount)
+	}
+	// Parsing transaction bytes checks each hash against the authenticated nodes
+	// and rejects missing or malformed data even in batches processed much later.
+	for i, hash := range body.Subtrees {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		if result := u.readSubtree(ctx, body, i, hash); result.err != nil {
```
galt-tr: Reviewed: this loop completes transaction parsing for every subtree before either quick-validation entry point reaches block-ID assignment or batch processing. The malformed-later-subtree regressions passed for sequential, pipelined, and async modes, including under the race detector. No change requested.

services/blockvalidation/quick_validate.go
```diff
@@ -1556,5 +1569,24 @@ func (u *BlockValidation) authenticateQuickBlockBody(ctx context.Context, block
 	if err := body.CheckMerkleRoot(ctx); err != nil {
 		return errors.NewBlockInvalidError("[quickValidateBlock][%s] body does not match header merkle root", block.Hash().String(), err)
 	}
+	authenticatedCount := body.TransactionCount
+	if len(body.Subtrees) == 0 {
+		authenticatedCount = 1 // the separately serialized coinbase is the whole body
+	}
+	if block.TransactionCount != authenticatedCount {
```
galt-tr, line 1668: Reviewed: comparing the supplied count against the fresh body's node count avoids overwriting the value being checked. The explicit coinbase-only count also handles GetAndValidateSubtrees returning early for an empty subtree list. The wrong-count cases and persisted-count regression pass. No change requested.

galt-tr (commented):
Reviewed the production changes and regression tests at 4acac5dce. No actionable defects found. Checked transaction-body authentication before mutation, declared transaction counts, retained subtree-node reuse, pipeline ordering, and retry/cleanup behavior.

Validation passed locally:
- go test -tags testtxmetacache ./services/blockvalidation -run '^TestQuickBodyPreflight' -count=1
- go test -race -tags testtxmetacache ./services/blockvalidation -run '^(TestQuickBodyPreflight|TestQuickValidateBlock|TestTryQuickValidation)' -count=1
- go test -race -tags testtxmetacache ./services/legacy/netsync -run '^TestHeaderProvenance_(ResetOrdering|ConcurrentReset)$' -count=1

Review was scoped to this diff and relevant surrounding code; no full-suite or large-block performance run.

oskarszoon (commented):
Reviewed at `4acac5dce`. Commenting rather than blocking since it's still a draft.

The preflight is sound. It's non-vacuous — expected is `block.Header.HashMerkleRoot`, bound to the checkpoint-verified header chain at `catchup.go:1349-1356`, and computed is recomputed from subtree *file contents* (`GetAndValidateSubtrees` never asserts `deserialized.RootHash() == storeKey`), so substituting either the hash list or the files fails. It's the first statement in both entry points, the three `processBlockSubtrees*` variants have exactly two non-test callers and both are downstream of it, and ordering is pinned: a mutant moving it to just before `AddBlock` fails 30 subtests across all three modes. Every component was reverted individually, not just the whole function — preflight → `return nil` fails 27, count check removed fails 18, `readSubtree` loop removed fails 9, node reuse disabled fails 3.

Two things worth fixing before it leaves draft, and they interact.

## A forged body found in the parse loop is `ErrProcessing`, not `ErrBlockInvalid`

`quick_validate.go:1581-1587` · `catchup.go:1175`, `:1191-1201`

The Merkle-root and transaction-count checks correctly return `NewBlockInvalidError`. The per-subtree loop returns `result.err` unwrapped, and `readSubtree` produces **no** `NewBlockInvalidError` at all — eight error constructors, all Storage / NotFound / Processing.

So `catchup.go:1175` doesn't abort, and `:1191-1201` deletes the block's `.subtree` files and returns `true, nil` to re-run normal validation — which is what the comment immediately above that check says must not happen:

> A body-authentication failure must abort without retrying the body through a state-mutating path or deleting another block's shared subtree files.

```
MUTATION late hash mismatch   blockInvalid=false processing=true
MUTATION late missing         blockInvalid=false processing=true
MUTATION late truncated       blockInvalid=false processing=true
MUTATION bad tx count         blockInvalid=true  processing=false
```

Not a validation bypass — the fallback is the stricter path — so this is about the shared `.subtree` deletion, re-running a known-forged body through a state-mutating path, and not charging the peer that served it.

A blanket wrap would be wrong, since an absent `.subtreeData` file is transient rather than a verdict. The split that's needed: deserialize-subtree (`:44`), deserialize-subtree-data (`:62`), invalid coinbase (`:70`) and missing-tx-at-index (`:75`) are verdicts on the body; failed-to-locate (`:7`), not-found-locally (`:10`), failed-to-get-subtree (`:26`) and failed-to-get-subtree-data (`:52`) are not.

## `blockvalidation_subtreeMmapDir` silently stops working on this path

`quick_validate.go:797`

The node reuse takes `block.SubtreeSlices[subtreeIdx]` whenever the slices are populated — which the preflight guarantees — so `readSubtree`'s `if u.mmapDir != ""` branch is now unreachable. Nodes come from `GetAndValidateSubtrees`, which is heap-only. Confirmed by running with the setting configured: `mmapBacked=false` at this HEAD, `true` at the merge-base. The setting's own docs say it saves ~48 GB of heap. No log, no error, no validation — an operator who configured it keeps the directory and silently loses the benefit, on a release branch.

**These two have to move together.** The reuse means `readSubtree` returns the same object as `block.SubtreeSlices[idx]` (pointer identity confirmed), and `batch.Close()` unmaps it on the error paths at `:453-456` and `:726-729` where the slot may not have been replaced yet — with catchup then falling back to normal validation on the same block, whose `GetAndValidateSubtrees` early-returns "already loaded". That can't fire today only because nothing is ever mmapped. Restore mmap without addressing the aliasing and a memory regression becomes a segfault.

## Cost

The transaction-data pass exactly doubles, serially. Counting `blob.Store`, A/B against a preflight-removed build: subtree-data reads 3→6 and bytes 1326→2652 in all three modes, node-file reads halve 6→3. `NewSubtreeDataFromReader` re-parses every transaction and recomputes every double-SHA256 txid, then the decoded `*Data` is discarded at `:1585`. Measured 325 MB/s / 1.2M tx/s per core — about 3 s for a 1 GB block, as a full barrier before any UTXO work. The pass it duplicates runs under `errgroup` with `SafeSetLimit(g, 128)`; the loop at `:1581` is a plain `for`. Worth parallelising or reusing the decoded data rather than discarding it.

## Smaller

- After a post-preflight failure `SubtreeSlices` is left fully non-nil, so `model.Block.GetAndValidateSubtrees:1178-1192` short-circuits on the fallback path and peer-supplied `SizeInBytes` survives unrecomputed. The `.subtree` deletion no longer forces a reload.
- The three `early hash mismatch` subtests are vacuous — they're exactly the three that still pass with the preflight reverted. `require.Error` at `:210` asserts no error class. `require.Empty(t, jobs)` is vacuous in two of three modes. The 100 ms negative window at `:202-206` is timing-dependent in pipeline/async and fails toward a false pass.
- One question rather than a finding: does every producer of a `*model.Block` reaching this path populate `TransactionCount` with the same convention? Both test-file edits here exist because fixtures didn't set it, and a mismatch now hard-aborts catchup with no fallback.

icellan:
Addressed in the follow-up changes:

- Node/data deserialization failures, invalid coinbase entries, and missing transactions now return ErrBlockInvalid. Missing/unreadable files retain storage/not-found classification. Structural node checks precede transaction reads, so a missing data file does not mask a known invalid merkle root or duplicate. The catchup regression checks that malformed bodies abort without deleting shared subtree files.
- Preflight now uses the mmap-aware loader and owns its loaded nodes until all processing/readers join. Batches borrow these nodes; error cleanup clears any remaining aliases before unmapping. Tests cover mapped nodes, batch cleanup, UTXO-write failure and retry, and heap fallback when mmap allocation fails. Normal-validation fallback clears SubtreeSlices before reload.
- Node and transaction reads use block_getAndValidateSubtreesConcurrency, independently of mutation batch size. Decoded transaction objects are released per worker; a test verifies concurrent reads with a configured limit of two. Transaction data is still decoded twice to avoid retaining an entire decoded block.
- Preflight derives SizeInBytes from decoded transaction sizes. The malformed-data cases now assert ErrBlockInvalid. Early malformed cases are coverage of the error class; the later-batch cases are the regressions for mutation ordering, with final UTXO and parent-spend assertions in addition to the scheduling window.
- Production catchup reads TransactionCount explicitly from the wire in model.readBlockFromReader. The block-assembly submission path sums subtree lengths, including the placeholder, or uses one for coinbase-only blocks. The HTTP catchup/retry regression exercises that wire format through persistence. The two earlier fixture changes supply this existing convention.

The root-cache review also prompted a direct check of the subtree dependency: deserialization retains the serialized root. Preflight now recomputes each root from the actual nodes and compares it with the store key and serialized root; altering nodes and matching transaction bytes while preserving the root fails in both heap and mmap tests.

services/blockvalidation/get_blocks.go
```diff
@@ -444,12 +444,13 @@ func (u *Server) fetchAndStoreSubtree(ctx context.Context, block *model.Block, s
 	)
 	defer deferFn()
 
+	store := u.catchupSubtreeStore(ctx)
```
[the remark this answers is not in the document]
icellan: Addressed in b4b3257c6 using the suggested documentation option. `catchupArtifacts.Set` now explicitly states that it preserves existing files regardless of caller overwrite options, and explains why cleanup ownership requires this.

The helpers also have a production caller outside `fetchAndValidateBlocks`: `Server.RevalidateBlock` calls `fetchSubtreeDataForBlock` without installing the ownership wrapper. I retained the explicit overwrite options to preserve that path's behavior. The final patch changes documentation only.

Verification: `go test -race -tags testtxmetacache ./services/blockvalidation -run '^(TestCatchupArtifacts_|TestFetchAndStoreSubtree|TestFetchSubtreeDataForBlock)' -count=1` passed (9.295s), and all commit hooks, including repository lint, passed. The existing shared-file regression explicitly requests an overwrite and verifies that the original data survives the write and cleanup.

services/legacy/peer_server.go
```diff
@@ -2885,31 +2899,32 @@ cleanup:
...
-
 	if s.nat != nil {
-		s.wg.Add(1)
 		go s.upnpUpdateThread()
 	}
-
-	// Start listening for ban events
 	go s.listenForBanEvents(s.ctx)
-
+	s.peerHandler()
```
galt-tr, line 2919: Moving `s.peerHandler()` below the goroutine launches is more than a shutdown-ordering fix, and the PR description doesn't mention it.

On `release/v0.15`, `Start()` called `s.peerHandler()` **synchronously** before `go s.upnpUpdateThread()` and `go s.listenForBanEvents(s.ctx)`. `peerHandler` only returns on `quit`, so neither of those goroutines ever ran while the node was up — they were launched at shutdown, when they immediately exit. This diff makes both live for the first time.

That is almost certainly the right fix: `banChan` is a 100-slot subscriber channel (`internal/banlist/ban_list.go:286`) and `handleBanPeerMsg` sends into it *from the peerHandler goroutine itself* (`peer_server.go:2054`), so with nothing draining it the 101st ban would have wedged the entire peer handler. But it is a live behaviour change on a release branch — UPnP port mapping and ban-driven disconnects both start working — and deserves a line in the description/release notes.

It also leaves one gap of exactly the kind this PR closes elsewhere. Now that `listenForBanEvents` actually runs, `handleBanEvent` does an unguarded `s.query <- disconnectMsg` (line 3778) followed by an unguarded `<-disconnectMsg.reply` (line 3781). Once `peerHandler` has exited, a ban event in flight parks that goroutine forever. Same two-line treatment `OnBlock` just received:

```go
select {
case s.query <- disconnectMsg:
case <-s.quit:
	return
}

select {
case err := <-disconnectMsg.reply:
	if err != nil {
		s.logger.Errorf("Error disconnecting banned peers: %v", err)
	}
case <-s.quit:
}
```

services/blockvalidation/quick_validate.go
```diff
@@ -768,37 +778,89 @@ func (u *BlockValidation) readSubtree(ctx context.Context, block *model.Block, s
...
+				bufferedReader.Reset(subtreeReader)
+				subtree, err = subtreepkg.NewSubtreeFromReader(bufferedReader)
+			}
+		} else {
 			subtree, err = subtreepkg.NewSubtreeFromReader(bufferedReader)
 		}
-	} else {
-		subtree, err = subtreepkg.NewSubtreeFromReader(bufferedReader)
+		if err != nil {
+			return subtreeResult{err: errors.NewBlockInvalidError("[getBlockTransactions][%s] failed to deserialize subtree %s", block.Hash().String(), subtreeHash.String(), err)}
```
galt-tr: Promoting a failed node deserialize to `ErrBlockInvalid` is right when the bytes came from the peer we are catching up from, but `findLocalSubtreeFile` at line 774 deliberately consults **both** file types, so `localFileType` here can be `FileTypeSubtree` — a subtree this node promoted itself, on an earlier attempt or while validating a different block. For those bytes the current peer is being convicted for our own file, and after this PR the consequences are terminal rather than recoverable:

- `tryQuickValidation` (catchup.go:1184) returns early on `ErrBlockInvalid`, so the `.subtree` deletion that used to force a clean reload is skipped;
- `catchupArtifacts.cleanup` won't remove it either — a pre-existing file was never `created` by this attempt;
- `isUnvalidatablePeerError` (Server.go:564, used at Server.go:708) then suppresses alternative-peer retry and calls `reportCatchupMalicious`.

Net effect: one corrupt or truncated local `.subtree` — or a `.subtreeData` retained across attempts by the new transient-failure path — makes catchup unable to self-heal, and burns a fresh honest peer's reputation on every retry.

Suggest scoping the verdict to bytes this attempt actually fetched: keep the storage/processing classification when `localFileType == fileformat.FileTypeSubtree`, and only return `ErrBlockInvalid` for `FileTypeSubtreeToCheck`. The same argument applies to the `.subtreeData` deserialize at line 877.

services/blockvalidation/catchup_artifacts.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		return err
+	}
+	entry.created = true
+	return nil
+}
+
+func (s *catchupArtifacts) cleanup(logger ulogger.Logger) {
+	// The fetch errgroup's context is canceled on failure (and after Wait).
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
galt-tr: `cleanup` is fully serial — one `Exists` plus one `Del` round-trip per tracked file, all sharing a single one-minute budget.

A catchup attempt covers every header between the common ancestor and the target (the buffer comment at catchup.go:831 assumes fetchers can run 2000 blocks ahead of validation), and each block contributes up to two entries per subtree, so `s.files` can easily hold thousands of keys. On an S3/HTTP-backed subtree store at a few ms per round-trip that exhausts the minute part-way through, and the failure mode is unhelpful: every remaining entry hits the `err != nil` branch, emits a `Warnf`, and is skipped — thousands of log lines, and the files the cleanup exists to remove stay on disk. It also runs inline in `fetchAndValidateBlocks`, so it holds up catchup teardown for up to a minute.

Bounded parallelism (`util.SafeSetLimit` at the fetch concurrency) plus a per-file timeout rather than a whole-loop one would make both the latency and the partial-cleanup behaviour predictable.

services/blockvalidation/quick_validate.go
```diff
@@ -1542,19 +1598,102 @@ func (u *BlockValidation) extendBatch(
...
+				return result.err
+			}
+			st := result.subtree
+			body.SubtreeSlices[i] = st
+			if len(st.Nodes) == 0 || (i == 0 && !st.Nodes[0].Hash.Equal(subtreepkg.CoinbasePlaceholderHashValue)) {
+				return errors.NewBlockInvalidError("[quickValidateBlock][%s] first subtree must start with the coinbase placeholder and subtrees must not be empty", block.Hash().String())
+			}
+			// Deserialization reads a cached root from the file. Recompute it
+			// from nodes instead of trusting that cache or the blob-store key.
+			root, err := st.RootHashWithReplaceRootNode(&st.Nodes[0].Hash, st.Nodes[0].Fee, st.Nodes[0].SizeInBytes)
```
galt-tr: Optional, but relevant given this runs for every subtree of every catchup block: `RootHashWithReplaceRootNode` starts with `st.Duplicate()`, which does `make([]Node, len(st.Nodes))` + `copy`. Here the "replacement" is node 0's own hash/fee/size, so the clone buys nothing.

For an mmap-backed subtree that copy pulls every node onto the heap — 48 bytes per node, so ~48 MB for a 1M-leaf subtree, times `concurrency` workers — which is precisely the allocation `blockvalidation_subtreeMmapDir` exists to avoid (and this preflight is what finally makes that setting effective on this path).

`subtreepkg.BuildMerkleTreeStoreFromBytes(st.Nodes)` produces the same root as its last element, with no clone:

```go
store, err := subtreepkg.BuildMerkleTreeStoreFromBytes(st.Nodes)
if err != nil || len(*store) == 0 || !(*store)[len(*store)-1].IsEqual(hash) || !st.RootHash().IsEqual(hash) {
	return errors.NewBlockInvalidError(...)
}
```

services/blockvalidation/quick_validate.go
```diff
@@ -1542,19 +1598,102 @@ func (u *BlockValidation) extendBatch(
...
 	}
 	if err := body.CheckMerkleRoot(ctx); err != nil {
-		return errors.NewBlockInvalidError("[quickValidateBlock][%s] body does not match header merkle root", block.Hash().String(), err)
+		return cleanup, errors.NewBlockInvalidError("[quickValidateBlock][%s] body does not match header merkle root", block.Hash().String(), err)
 	}
-	return nil
+	var authenticatedCount uint64
+	for i, st := range body.SubtreeSlices {
+		if i > 0 && i < len(body.SubtreeSlices)-1 && st.Length() != body.SubtreeSlices[0].Length() {
+			return cleanup, errors.NewBlockInvalidError("[quickValidateBlock][%s] subtree %d has inconsistent length", block.Hash().String(), i)
```
galt-tr: This branch is unreachable: `body.CheckMerkleRoot(ctx)` runs immediately above and already enforces a strictly stronger version of the same rule (`model/Block.go:1427-1447`) — non-final subtrees must all equal `SubtreeSlices[0].Length()`, the final one must not exceed it, and the first subtree's leaf count must be a power of two. Your condition covers `i` in `[1, len-2]`, which is a subset.

The loop itself is still needed to accumulate `authenticatedCount`; it's just the check inside it that reads as a second, weaker copy of a rule enforced one line earlier. Dropping it (or replacing it with a comment pointing at `CheckMerkleRoot`) would save the next reader from having to work out which of the two is authoritative.

services/blockvalidation/catchup.go
```diff
@@ -1197,6 +1206,8 @@ func (u *Server) tryQuickValidation(ctx context.Context, block *model.Block, cat
 				}
 			}
 		}
+		// Force normal validation to reload metadata after a partial quick attempt.
+		block.SubtreeSlices = nil
```
galt-tr, line 1215: Docs drift worth fixing alongside this: `docs/topics/services/blockValidation.md` ("Error Handling", around line 516) still says that if quick validation encounters **any** error it "Removes `.subtree` files to force reprocessing" and "Falls back to normal validation automatically".

After this change a body-authentication failure does neither — it returns `false, err`, deliberately preserves the shared subtree files, aborts catchup and flags the peer. That paragraph is now actively misleading about the security-relevant path, which is the one a reader is most likely to consult. A couple of sentences distinguishing the `ErrBlockInvalid` abort from the fall-back-and-delete path would keep it honest.

galt-tr (approved):
Reviewed at `b4b3257c6` against `release/v0.15` (merge-base `69132e684`), in a separate worktree.

**What I ran**

- `SETTINGS_CONTEXT=test go test -count=1 -tags testtxmetacache ./services/blockvalidation/... ./services/legacy/...` — all green (blockvalidation 271.7s, legacy/netsync 8.6s).
- `go test -race -count=1 -tags testtxmetacache ./services/blockvalidation -run '^(TestQuickBodyPreflight|TestCatchupArtifacts|TestTryQuickValidation|TestQuickValidateBlock|TestServer_catchup)'` — green, no races.
- `go build ./...`, `go vet`, `golangci-lint run --new-from-rev=origin/release/v0.15 ./services/blockvalidation/... ./services/legacy/...` — clean (0 issues).
- Two mutation checks, to confirm the new regressions are load-bearing rather than decorative:
  - making `artifacts.cleanup` unconditional fails `TestCatchupArtifacts_TemporaryFailurePreservesProgress`;
  - replacing the recomputed root with the file-cached `st.RootHash()` fails both `TestQuickBodyPreflightRecomputesStoredRoot` subtests.
- CI at this head: 34/34 checks passing.

**What I checked in the surrounding code**

The preflight itself holds up. `go-subtree`'s `DeserializeFromReader` seeds `st.rootHash` directly from the file header, so recomputing via `RootHashWithReplaceRootNode` really is the check that matters, and the store key alone was never asserted before. The transaction-count convention the preflight now enforces (sum of `subtree.Length()`, with `1` for a subtree-less block) matches what produces those values: `services/blockassembly/Server.go:1448-1473` and the legacy importer's `len(block.Transactions())` with a coinbase placeholder in subtree 0, and `NewBlockFromReader` always populates `TransactionCount` from the wire, so the hard abort has no unset-field footgun. The inconsistent-length rule matches what `GetAndValidateSubtrees` already enforced.

Borrowed-node ownership also looks right: `batch.Close()` skips borrowed slots, `cleanup` nils `block.SubtreeSlices[i]` only while it still aliases the body slice (so a slot replaced by `buildSubtreeAndQueueWrite` survives), the deferred `cancelReaders(); g.Wait()` in both batch readers means no reader can publish into a batch after it is abandoned, and `mmapNodeStore.Close` is `sync.Once`-guarded so the double `cleanup()` in the ownership test cannot double-unmap. The `Server_test.go` edit is correct — `MockStore.StoreBlock` already promotes block 49 under its own lock (`stores/blockchain/mock.go:531`).

Approving. Six comments below, none blocking. The two I would most like to see addressed are the legacy `Start()` reordering (it activates two goroutines that never ran before, which the description doesn't mention) and the now-stale error-handling paragraph in `docs/topics/services/blockValidation.md`.

services/legacy/peer_server.go
icellan: Fixed in 797157b1c. Both the query send and reply wait now select on server shutdown and context cancellation. The reply channel is buffered so an already accepted query can reply after the listener exits without parking the peer handler. The listener itself is registered in the lifecycle wait group and exits on shutdown or channel closure.

Added regressions for send/reply cancellation, late replies, and listener exit/join. They failed before the fix and passed twice under the race detector afterward. The PR description now explicitly calls out the newly active UPnP maintenance and ban-event listener during normal runtime.

services/blockvalidation/quick_validate.go
icellan: Fixed in 797157b1c. Classification uses the attempt's ownership tracker rather than file type alone: `.subtreeToCheck` and `.subtreeData` can both survive earlier attempts. Corrupt cached bytes now produce a local service/storage error without an embedded invalid-block verdict, so catchup skips peer penalties and normal-validation fallback. After all attempt readers/writers join, repair removes only the identified corrupt files; a subsequent attempt can refetch them.

The failing-before/passing-after HTTP regressions cover truncated and forged-root cached nodes under both node file types, and corrupted cached transaction data. They verify no UTXO mutation, preservation of unrelated shared files and successful retry fetching exactly the one bad file. Additional coverage preserves healthy promoted nodes when only transaction data is corrupt, and confirms newly downloaded bad files still return ErrBlockInvalid. These regressions passed twice with the race detector.

services/blockvalidation/catchup_artifacts.go
icellan: Fixed in 797157b1c. Cleanup now uses bounded workers at SubtreeFetchConcurrency (default eight), with a fresh five-second timeout per file. I retained a one-minute overall ceiling to bound teardown even if every storage operation stalls; when it expires, scheduling stops. Failures produce one summary warning with the first error instead of one warning per remaining key.

The concurrency regression failed against the old serial loop. Tests now verify concurrent work with a bounded worker count, progress past an individual timeout, the overall deadline, one summary warning and preservation of shared/promoted files. They passed twice under the race detector.

services/blockvalidation/quick_validate.go
icellan: Fixed in 797157b1c. Preflight recomputation now calls BuildMerkleTreeStoreFromBytes directly and compares its final hash with both the requested hash and serialized cached root. It no longer clones the nodes or invokes replacement of an unchanged node.

The heap/mmap forged-root regressions pass under the race detector. A local 65,536-leaf comparison produced identical roots and measured allocation falling from 7,366,822 to 2,110,379 B/op. The Merkle store still allocates; this removes the redundant clone and duplicate computation in the old helper.

services/blockvalidation/quick_validate.go
icellan: Fixed in 797157b1c. Removed the redundant length branch, retained authenticatedCount accumulation, and added a comment identifying CheckMerkleRoot as the authoritative partitioning/length check. The preflight and declared-count regressions passed twice under the race detector; the full normal blockvalidation package also passed in the repository-wide run.

services/blockvalidation/catchup.go
icellan: Fixed in 797157b1c. The Error Handling section now distinguishes invalid-body aborts, local cached-corruption recovery, incomplete-block aborts and the remaining fallback/delete path. It also documents cleanup ownership and timeout/concurrency limits. The adjacent Merkle paragraph now describes authentication before UTXO mutation rather than a post-processing check. Markdown lint passed in the commit hooks.

PR #1728, opened 2026-09-12
fix(utxo): hold back pruned children until their replay markers land (#1701)

### LINKED ISSUE #1701: [BUG] Pruner filter false positives let mined transactions re-enter block assembly

RobGanon:
## Summary

Fixes #1701 — closes the replay-protection hole where a pruned child could lose its `deletedChildren` marker on a surviving parent and be replayed after re-org. This is the Aerospike-store subset of the upstream #1702 fix.

### The bug

The non-defensive ("combined") cleanup path sent parent marker writes and child record deletions in a **single `BatchOperate`**. Aerospike batches are not atomic across records: a marker write could fail while the child was still deleted, leaving the child in `prunedSet` with **no marker on any surviving parent**. `teranode.lua`'s defensive replay check against that bin then had nothing to see, so an already-mined and pruned transaction could be accepted on a re-org.

Two secondary holes were found while fixing it:

- If a record's input references **could not be resolved** (missing external blob, malformed input bin), its parents were unknowable, yet the old code deleted the record anyway — the child removed with no marker, by construction.
- The plain `MapPutItems` parent-update fallback **fail-opened** on a whole-batch server rejection: the client returns such rejections without per-record errors, so a nil `Err` was read as success and the children were deleted with no marker.

### The fix

- **Two-step flush in all modes** (`flushCleanupBatches`): parent markers are written and tallied first; only the children whose markers actually landed are deleted. Combined-mode batching is gone.
- **Per-parent hold-back**: a failed marker write isolates to that parent's children (via per-record results / a SUCCESS map from the mod-teranode paths). Those children — every master/pagination key and their external blobs — stay in place and are reconsidered next cycle instead of failing the whole chunk (which would wedge `PruneWithPartitions` forever: it never retries non-timeout errors and the worker advances `lastProcessedHeight` regardless).
- **Input-resolution retention**: records whose inputs can't be resolved are retained and counted, not deleted. The missing-blob case is no longer treated as idempotent.
- **Guard keyed on the incoming spender** (spend path): `buildSpendFilterExpression` gains a `deletedChildren` clause, and `teranode.lua` gets a replay guard on the unspent path that checks the marker set the *incoming spender's* markers target. Lua package bumped to `teranode_v62` (a string constant; reconciled with #1702's later bump).
- **Fail-closed batch-write fallback**: `BatchOperate` errors on the plain path fail the chunk rather than classifying nil per-record errors as success.

### Scope

Aerospike store only. Deliberately out of scope here (upstream #1702 follow-ups):

- SQL store `deleted_children` table
- netsync/validator ghost-record cleanup
- `ErrUtxoSpendingTxPruned`
- idempotent-spend / rollback support
- `PrunedTxSet` removal (its plumbing is kept; #1706 merged)

### Testing

- `go build ./...`, `go vet ./stores/utxo/aerospike/...`, `gofmt` — clean
- `go test -tags testtxmetacache ./stores/utxo/aerospike/...` and `.../pruner` — pass
- New unit tests: `parent_update_native_op_test.go`; reworked `prune_policies_test.go`, `parent_update_response_test.go`, `partition_retry_test.go`
- Aerospike TestContainers integration tests were **not run** (no docker access in the dev environment); they skip via `SkipIfContainerUnavailable` — please run them in CI.


stores/utxo/aerospike/pruner/pruner_service.go
```diff
@@ -1154,20 +1194,14 @@ func (s *Service) processRecordChunk(ctx context.Context, blockHeight uint32, ch
...
-					childHashes: []*chainhash.Hash{txHash},
-				}
+			// Failure here means the key source for one of this record's
+			// parents is unreachable - same retention reasoning as above.
+			if err := s.addParentUpdatesForInput(allParentUpdates, parentTxID, input.PreviousTxOutIndex, txHash); err != nil {
+				s.logger.Infof("Retaining %s at height %d: could not enqueue replay marker for parent %s:%d: %v",
+					txHash.String(), blockHeight, parentTxID.String(), input.PreviousTxOutIndex, err)
+				prometheusUtxoInputResolutionErrors.Inc()
+				skippedCount++
+				continue
```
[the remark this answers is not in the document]
RobGanon: > **Major**: this `continue` does not retain the record — it only skips the failed input, so the child is still deleted with no marker... Use a retain/break pattern.

Fixed in adb7e3845: the input loop now sets a `retain` flag and `break`s on `addParentUpdatesForInput` failure, and the record is skipped at the outer-loop level — no longer deleted, no longer double-counted. Also updated `TestExternalFileAlreadyDeleted`, which had asserted the old delete-anyway behavior; the CI `test` job failure was exactly that test, now updated to expect retention when inputs cannot be resolved.

ordishs (changes requested):
**stores/utxo/aerospike/spend.go:831** (with `native_op.go:96-98`, `spend_expressions.go:225-254`) — P1

- The new replay guard reaches only two of the three spend transports. `sendSpendBatchLua` takes the expression path only when `utxoBatchSize == 1` (spend.go:689-695). Every other configuration calls `executeLuaSpendBatch`, which builds the record via `s.teranodeBatchRecord(..., subOpSpendMulti, "spendMulti", ...)`.
- `useNativeForSubOp` fences only `subOpUnspend`. With `aerospike_use_native_teranode_ops` on, `spendMulti` goes to the server-fork native sub-op 2 and `teranode.lua` never runs. The `batchUDFPolicy` at spend.go:822 carries no `FilterExpression`, so the new `deletedChildren` clause does not apply to that record either.
- The pruner in this same PR uses the native `buildAddDeletedChildrenRecord` builder to WRITE the markers. Such a deployment writes replay markers that nothing on the spend path reads.
- Failure: `use_native_teranode_ops=true`, `utxoBatchSize > 1`. Pruner removes child C and writes `deletedChildren[C]` on parent page P. A reorg unspends P's outputs. C is re-delivered. Native sub-op 2 sees the output unspent, has no `deletedChildren` clause, and accepts the spend — the #1701 replay.
- Recommendation: fence `subOpSpendMulti` to the UDF path in `useNativeForSubOp`, the same treatment `subOpUnspend` gets and for the same reason, until the fork implements the check and `probeNativeSpendSemantics` exercises it. The expression clause cannot substitute: one batch record groups many spenders under one key, so a single-spender filter cannot express the rule. If the fork already enforces it, please say so and add the probe assertion.

**stores/utxo/aerospike/pruner/pruner_service.go:1689-1760** — P1

- `flushCleanupBatches` is the gate the whole PR installs, and no test exercises it. The new tests stop at `tallyParentUpdateResults`, which only proves `blocked` is populated.
- Nothing asserts that a blocked child's keys are withheld from `executeBatchDeletions`, that all of a child's pagination keys go together, that its external blob is withheld, or that `heldBack` is returned and subtracted correctly. `isBatchTransportError`, the new `buildSpendFilterExpression` clause, and the reworked `teranode.lua` guard have no tests either. TestContainer tests were not run.
- Failure: latent — no reachable path found. The exposure is regression, not a present defect: a later refactor can reintroduce #1701 with the suite green.
- Recommendation: add a table test over `flushCleanupBatches` with a fake client returning per-record failures for a chosen subset of parents; assert the deletion batch holds every key of the unblocked children and none of the blocked ones. Mirror it for `externalFiles`. Include a `pendingDeletion` with `totalExtraRecs > 0` to pin the grouped-keys guarantee. Run the Aerospike TestContainer tests in CI before merge.

stores/utxo/aerospike/spend_expressions.go
```diff
@@ -221,6 +222,38 @@ func (s *Store) buildSpendFilterExpression(
...
+	// a re-delivered spend would recreate a single live UTXO that no pruner sweep
+	// will ever see again.
+	//
+	// Legitimate deliveries never trip this. txHandler never re-delivers a spend,
+	// and the pruner only removes a child whose parent has been stable for the
+	// retention window, so a spender showing up twice is exactly the replay this
+	// clause exists to catch. When a record is refused here it falls through to
+	// the Lua UDF, which distinguishes a double-spend from an idempotent
+	// re-delivery.
+	if spendingTxID != nil {
```
freemans13, line 240: This clause does not protect native-op deployments, even at `utxoBatchSize == 1`. It extends ordishs's first P1.

When the clause fails, Aerospike returns FILTERED_OUT, and `processSpendBatchResultsExpressions` hands the record to `executeLuaSpendBatch` (line 600). That builds the record with `teranodeBatchRecord(..., subOpSpendMulti, ...)`. With `aerospike_use_native_teranode_ops` on, `useNativeForSubOp(subOpSpendMulti)` is true, so the retry goes to native sub-op 2 and `teranode.lua` never runs. The replay the filter just refused is then accepted.

The comment above (lines 237-238) says a refused record "falls through to the Lua UDF". That is only true when native ops are off. Fencing `subOpSpendMulti` in `useNativeForSubOp` fixes both the batch-size-1 and batch-size-above-1 cases.

stores/utxo/aerospike/teranode.lua
```diff
@@ -382,26 +382,47 @@ function spendMulti(rec, spends, ignoreConflicting, ignoreLocked, currentBlockHe
...
+        -- existingSpendingData: a replayed transaction re-appears with its
+        -- original bytes while the output element may hold anything after
+        -- pruning, so only the incoming bytes still name the offender. The
+        -- expression filter applies the same rule before this UDF is ever
+        -- reached.
+        --
+        -- Idempotent re-deliveries of live spends are unaffected: their txid is
+        -- not in deletedChildren, so they fall through to the same-data branch
+        -- below and are swallowed without re-incrementing spentUtxos.
+        if deletedChildren ~= nil then
```
freemans13, line 401: No test fails if this guard goes back to the old rule. I changed this line to `if deletedChildren ~= nil and existingSpendingData ~= nil then` in a throwaway worktree. That checks only spends of an already-spent output, like before this PR. The store tests matching `Replay|Prune|Pruner|DeletedChildren|Spend|Lua` still passed against real Aerospike (85.5s).

`TestPrunerReplayProtection` replays against an output that is still marked spent, so it never reaches the case this PR adds. A test that would pin it:

1. Build parent P with two outputs and child C spending output 0. Mine both and prune C, as `TestPrunerReplayProtection` already does.
2. `Unspend` P:0, which is what a reorg does to C's spend.
3. `SpendAndCreate` C again, and require `errors.ErrUtxoError` with "invalid spend" in the message, the same assertions `TestPrunerReplayProtection` makes, plus that C was not recreated.
4. Run it once with `EnableSpendFilterExpressions` on and `utxoBatchSize == 1`, so the expression clause is covered too.

Then break this line and confirm the test fails.

stores/utxo/aerospike/pruner/pruner_service.go
```diff
@@ -1499,11 +1555,15 @@ func (s *Service) getTxInputsFromBins(ctx context.Context, blockHeight uint32, b
...
 
-				// Idempotent: External file missing (cleaned by LocalDAH or previous run)
-				// We can still proceed with record deletion - return empty inputs list
-				s.logger.Debugf("external tx %s already deleted from blob store at height %d, proceeding to delete Aerospike record",
+				// Retain the record: its inputs cannot be resolved, so the parents
+				// that must carry its replay markers are unknown. Deleting it here
+				// is exactly the #1701 hole - a child removed with no marker - and a
+				// missing blob is indistinguishable from a concurrent deletion race,
+				// so there is no idempotent reading of it. The record is reconsidered
+				// by the next prune cycle and eventually removed via a normal prune.
```
freemans13, line 1563: This comment says the record is "eventually removed via a normal prune", and I don't see a path that gets there. When the tx blob is really missing (not a race), nothing restores it. The external store is built with `WithDisableDAH(true)`, so the pruner is its only deleter. `getTxInputsFromBins` will return this error on every later cycle too.

The partition query matches `delete_at_height` from 1 to `blockHeight`, so the record comes back every block. Each time it costs a `GetIoReader` plus an `Exists` against the blob store, an Info log at line 1182, and one more on `utxo_pruner_input_resolution_errors_total`. That counter will read as a growing error rate when it is really a fixed set of stuck records.

Retaining is the safe choice, and I would not go back to deleting. But the comment should state that it is permanent, and operators need some way to find these records. A Warn the first time a record is retained, or a gauge of currently-retained records instead of a per-block counter, would both do it.

freemans13 (changes requested):
Requesting changes. I agree with both of ordishs's P1 findings, and I have evidence for each that makes them stronger. There is also one smaller problem in the missing-blob retention path. The two-step flush reads correctly to me: markers go first, and only children whose markers landed get deleted. What worries me is that nothing in the suite would notice if it stopped working.

**The native spend path skips the replay check at every batch size, not only above 1.** ordishs's failure case needs `utxoBatchSize > 1`. It also happens at `utxoBatchSize == 1` with spend filter expressions on. The new `deletedChildren` clause makes the expression refuse the record (Aerospike returns FILTERED_OUT). That record is then re-sent through `executeLuaSpendBatch`. From there `teranodeBatchRecord` sends `spendMulti` to native sub-op 2 whenever `aerospike_use_native_teranode_ops` is on, because `useNativeForSubOp` only keeps `subOpUnspend` on the UDF path. `teranode.lua` never runs, and the refusal turns into an accepted spend. The fix ordishs suggests covers both cases: route `subOpSpendMulti` to the UDF path until the server fork does the check. Code path: `spend_expressions.go:600`, `spend.go:831`, `native_op.go:96-98`, `native_op.go:203`.

**Two mutations survive the tests, so neither half of the fix is pinned.** I ran both against real Aerospike 8 through testcontainers, on adb7e3845 in a throwaway worktree. Nothing was pushed.

1. I deleted the hold-back check in `flushCleanupBatches` (the `if _, held := blocked[*deletion.txHash]` block), so every child gets deleted whether its marker landed or not. `go test -tags testtxmetacache ./stores/utxo/aerospike/pruner/...` still passed (42.6s).
2. I changed the new Lua guard to `if deletedChildren ~= nil and existingSpendingData ~= nil`. That is the old behaviour, where only a spend of an already-spent output was checked. The store tests matching `Replay|Prune|Pruner|DeletedChildren|Spend|Lua` in `./stores/utxo/aerospike/` still passed (85.5s). `TestPrunerReplayProtection` only covers the already-spent case. The reorg case this PR exists for is: parent output unspent, then the pruned child is re-delivered. No test covers it.

So ordishs's second P1 is concrete. Break either half of the fix and CI stays green. I left an inline note on the Lua guard with the test I would add.

**Smaller: a record whose blob is genuinely gone is kept forever.** The code comment says it is "eventually removed via a normal prune", but nothing puts a missing blob back. Details inline.

**What I ran and what I could not check.** Unmodified adb7e3845 passes `./stores/utxo/aerospike/pruner/...` (43.7s) and the filtered store tests above (77.2s), with the containers running, not skipped. I did not run the full `./stores/utxo/aerospike/` package. Upstream CI has not run on adb7e3845 at all: all six workflows for that commit are `action_required`, waiting on a maintainer. I cannot check whether the server fork's native sub-op 2 enforces `deletedChildren`. If it does, please say so and add the probe assertion ordishs asked for.

icellan:
Closing as superseded by #1702. Not a duplicate-close — the reasons, so it's clear
what was compared.

#1728 and #1702 fix the same hole and reach the same shape on the Aerospike parent
marker path: markers written before deletions, transport errors failing the chunk,
server-side rejections classified per record with the affected children held back.
Where they differ, #1702 is ahead on every axis:

- **SQL.** #1728 is Aerospike-only. #1702 carries the SQL half, including the marker
  INSERT joined to the parent output so only the child that actually held the spend
  is marked.
- **Native sub-op fence.** ordishs's P1 here — with `use_native_teranode_ops` on and
  `utxoBatchSize > 1`, `spendMulti` goes to native sub-op 2 and `teranode.lua` never
  runs, so the pruner writes markers nothing reads. #1702 fixes this:
  `isReplayProtectionSubOp` covers `subOpSpendMulti`, `subOpSpend` and
  `subOpAddDeletedChildren`, and `probeNativeReplayProtection` plants a marker
  through sub-op 13 and requires INVALID_SPEND.
- **Fail-closed batch classification.** #1702's `classifyParentUpdateResult` treats
  `ResultCode != OK` with a nil `Err` as a failure, so a record the batch never got
  an answer for is held back. #1728 has no such guard, which is why it has to abort
  the whole chunk on the `MapPutItems` path.
- **Lua guard.** #1702 keys the marker check on the incoming spender and answers it
  ahead of every other record-level answer, so `unspend` can't disarm it. LuaPackage
  v65 against v62.
- **CI.** #1728's workflows have never run — six sit `action_required`, and the only
  two checks on the PR are Socket Security.

Both P1s raised here are carried forward, not dropped:

- ordishs's native-spendMulti fence: fixed in #1702 as above; the probe assertion
  he asked for exists.
- The `flushCleanupBatches` coverage gap, and freemans13's mutation testing behind
  it: I ran the equivalent mutations against #1702 at `f04cf3b0d` on real Aerospike
  8. Both are caught — removing the hold-back fails `TestPrunerReplayProtection`
  (`marker_failure`, `ttl_marker_failure`, `paginated_marker_failure`) and
  `TestPrunedReplayRollbackKeepsSiblingHistoricalSpend`; nesting the Lua guard back
  under `bytes_equal(existingSpendingData, spendingData)` fails `unspent_parent`,
  `paginated_unspent_parent`, `replaced_parent`, `frozen_parent` and the three
  `expressions_*` variants — the reorg case that had no coverage here.

One residual worth keeping: those tests live in `stores/utxo/aerospike`, not in
`stores/utxo/aerospike/pruner`. The pruner package alone stays green with the
hold-back deleted, so anyone verifying that gate by running only the pruner package
gets a false pass.

The parent-marker hold-back is the right mechanism, and #1702 lands the same idea
with the SQL side and the native path covered.

PR #1729, opened 2026-09-13
fix(utxo): persist replacement locking script on ReAssignUTXO (#1725)

### LINKED ISSUE #1725: Reassignment after mandatory re-extension strands UTXOs for both owners

RobGanon:
## Summary

Fixes #1725. `ReAssignUTXO` previously persisted only the new UTXO commitment, never the replacement locking script. When a frozen UTXO was reassigned to a new owner and the maturity gate elapsed, re-extension served the **original** script back to the validator: the new owner's spend failed script validation, and the old owner's spend produced a `UTXO_MISMATCH` on the reassigned hash — both owners stranded.

## Fix

The reassignment now takes the authoritative amended output (`*bt.Output` with the replacement `LockingScript` and `Satoshis`) as the API input. The store derives the commitment from it (`util.UTXOHashFromOutput`) and persists the script alongside the hash:

- **SQL**: UPDATE writes `utxo_hash`, `locking_script`, `satoshis`, sets `frozen=false` and the `spendableIn` gate; fails with `UtxoFrozenError`-class `NewInvalidArgumentError` when the script is missing/empty.
- **Aerospike**: the `reassign` UDF takes the new script and stores it in a new `reassignedScripts` (offset → script) bin; `sendOutpointBatch` reads that bin and applies the script as the authoritative outpoint script during re-extension. Lua package bumped to `teranode_v62`.
- **RPC**: `reassign` gains optional `Newscript` (hex), `Newsatoshis`, `NewUTXOHash`. It refuses the call when script or satoshis are absent, and validates `NewUTXOHash` against the derived commitment when supplied.

E2E coverage in `test/e2e/daemon/ready/reassign_test.go` is inverted to the fixed behaviour:
- new owner held by the maturity gate pre-maturity (`not spendable until`), then spends freely after maturity;
- original owner's spend now fails against the persisted replacement script (`OP_EQUALVERIFY`), leaving the output unspent;
- same-owner (self-reassignment) output still passes once matured.

## Verification (run locally)

- `go build ./...`, `go vet` on affected packages, `gofmt` clean.
- `go test ./stores/utxo/sql ./stores/utxo/logger ./services/rpc ./stores/txmetacache` pass.
- Aerospike CI tests require containers (upstream CI).

## Notes for review

- Two known gaps worth attention:
  1. `reassignedScripts` is keyed page-relative; the decoration override in `sendOutpointBatch` is honoured for base-record (page 0) outputs where offset == vout. An output with `vout >=` the UTXO batch size (pagination record) falls outside the override path.
  2. The Lua history `reassignments` entry keeps its original 4-key shape; the new script lives only in the `reassignedScripts` bin.

- **Design question for the RPC surface**: extending `reassign` with an optional script/satoshis/hash means a caller can still reassign *without* supplying a script (the store refuses, so nothing is persisted), but `chainsvrcmds.go` models the script as optional. An alternative is making the new script a **required** field on the `reassign` command. Feedback welcome on which shape is better for the RPC contract.

### DIFF
--- a/services/alert/node.go
+++ b/services/alert/node.go
@@ -475,23 +475,11 @@ func (n *Node) AddToConfiscationTransactionWhitelist(ctx context.Context, txs []
 				LockingScript: newLockingScript,                                          // new locking script
 			}
 
-			// Calculate the replacement commitment. ReAssignUTXO does not persist
-			// the replacement script, so changing the owner currently strands
-			// the output for both owners after mandatory re-extension (issue 1725).
-			newUtxoHash, err := util.UTXOHashFromOutput(txIn.PreviousTxIDChainHash(), amendedOutputScript, txIn.PreviousTxOutIndex)
-			if err != nil {
-				response.NotProcessed = append(response.NotProcessed, n.getAddToConfiscationTransactionWhitelistResponse(tx.TxIDChainHash().String(), err)...)
-				continue
-			}
-
-			newUtxo := &utxo.Spend{
-				TxID:     txIn.PreviousTxIDChainHash(),
-				Vout:     txIn.PreviousTxOutIndex,
-				UTXOHash: newUtxoHash,
-			}
-
-			// re-assign the utxo
-			if err = n.utxoStore.ReAssignUTXO(ctx, oldUtxo, newUtxo, n.settings); err != nil {
+			// re-assign the utxo. The store derives the replacement commitment
+			// from the amended output and persists the replacement locking
+			// script, so the new owner is spendable after the maturity gate and
+			// the original owner is no longer (issue 1725).
+			if err = n.utxoStore.ReAssignUTXO(ctx, oldUtxo, amendedOutputScript, n.settings); err != nil {
 				response.NotProcessed = append(response.NotProcessed, n.getAddToConfiscationTransactionWhitelistResponse(tx.TxIDChainHash().String(), err)...)
 			}
 		}
--- a/services/blockpersister/Server_test.go
+++ b/services/blockpersister/Server_test.go
@@ -1038,7 +1038,7 @@ func (m *MockUTXOStore) FreezeUTXOs(ctx context.Context, spends []*utxo.Spend, t
 func (m *MockUTXOStore) UnFreezeUTXOs(ctx context.Context, spends []*utxo.Spend, tSettings *settings.Settings) error {
 	return nil
 }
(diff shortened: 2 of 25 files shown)


RobGanon:
Could a maintainer please approve the workflow run for this fork PR? CI and the automated Claude review are both waiting on the approval (they only start once the Terra node PR init workflow succeeds). Thanks!

PR #1730, opened 2026-09-14
Check the merkle root of coinbase-only blocks

icellan:
`model.Block.CheckMerkleRoot` has a branch for blocks with no subtrees — an empty, coinbase-only block's merkle root is just the coinbase txid. Nothing reached it.

`Block.Valid` gates the call behind `subtreeStore != nil && len(Subtrees) > 0`, and quick validation only gets there via `validateSubtrees`, which a zero-subtree block never enters. So the one body shape whose merkle root is cheapest to verify was the one shape that never got verified.

The merkle root is what relates a block's body to its header (`GetHash` covers the header alone), so it should be evaluated for every body shape. A coinbase-only block's outputs become UTXOs like any other, and below a checkpoint the fee arithmetic is skipped, so on those paths nothing read the coinbase at all.

## Changes

- `Block.Valid` runs `CheckMerkleRoot` for zero-subtree blocks. No subtree store needed, so it also covers nil-store callers.
- `checkBlockRewardAndFees` gates the below-checkpoint fee skip on `merkleRootChecked`, matching the sibling `validOrderAndBlessed` skip. That skip's premise is that the checkpoint transitively covers the coinbase, which only holds once the body has been hashed against the header. No effect on real blocks — every production `Valid` caller passes a subtree store.
- `quickValidateBlock` / `quickValidateBlockAsync` check the merkle root on the empty-block path, and require the coinbase to be coinbase-shaped (previously only `len(Inputs) != 0`).

Block assembly already computes an empty block's merkle root as the coinbase txid (`blockassembly/Server.go`), so it agrees with the validator and does not reject its own blocks.

## Tests

Five new tests, written failing first: mismatch rejected and honest empty block accepted, at both the `Block.Valid` and `quickValidateBlock` layers, plus the mismatch case below a checkpoint where the fee check returns early.

Some fallout worth flagging — several fixtures built empty blocks with a placeholder merkle root and an unrelated coinbase, which only validated while this check was missing. `CreateTestBlocks` now binds the two, and there's a `CreateSyntheticBlocksFrom` for chains that get served as block bodies.

One test changed shape: `TestCatchup_SybilAttack` runs its honest chain on real mainnet headers, but the fixture stores headers without coinbases (`coinbase_tx: null`), so those blocks cannot be made to match their headers. It now asserts on *where* each peer fails — Sybil peers turned away during header evaluation, the honest peer reaching block validation — instead of asserting a block whose coinbase cannot satisfy its header is accepted. Getting it back to full strength needs the real coinbases for mainnet blocks 1-20 in the fixture.

## Added after self-review

**Widened the gate.** The second branch was `len(Subtrees) == 0 && len(SubtreeSlices) == 0`, which covered the empty-block shape but left an in-memory block carrying its own slices (`NewBlockFromMsgBlock`'s offline shape) unchecked whenever the store was nil. It now gates on the two lists agreeing, which is what `CheckMerkleRoot` itself requires. No production caller passes a nil store, so no live path changes — it removes a latent gap of the same kind an earlier audit closed for the duplicate-tx check (`TestBlock_Valid_DupTxDetected_NilSubtreeStore`, same gate).

Two nil-store duplicate-tx fixtures depended on that gap and now bind their merkle root, which is the faithful shape anyway: a CVE-2012-2459 mutation produces the *same* root as the honest block, so the duplicate check is what has to catch it.

**`TransactionCount` / `SizeInBytes` for coinbase-only blocks.** Both arrive from the wire and are persisted as given into `blocks.tx_count` / `blocks.size_in_bytes`. `GetAndValidateSubtrees` recomputes them from the real body, but only for blocks that carry subtrees — so this shape kept whatever the sender supplied. Both are known exactly here: one transaction, and a size of header + tx-count varint + coinbase. `RecomputeCoinbaseOnlyCounts` applies that arithmetic (mirroring `GetAndValidateSubtrees` with an empty subtree contribution) and is a no-op for blocks with subtrees.

**Fixture bug found while reviewing.** `CreateSyntheticBlocksFrom` derived everything from its arguments, so the two "competing" chains in `TestCatchup_CompetingEqualWorkChains` came out identical byte for byte — the first-seen assertion was comparing a chain against itself. Verified: `identical=true`. Added a `variant` parameter that offsets timestamps, leaving heights and work unchanged, and documented the 2h future window that bounds chain length.

## Verification

`model`, `model/time`, `services/blockvalidation`, `services/blockvalidation/catchup`, `services/blockassembly` (+ `mining`, `subtreeprocessor`), `services/blockpersister`, `services/legacy/netsync` all green after each commit, including after the merge from `main`. `go vet` clean, `model` green under `-race`.


ordishs (approved):
Verified the production change and re-ran the affected suites on this branch merged with `main`. `model`, `model/time`, `services/blockvalidation`, `services/blockvalidation/catchup`, `services/blockassembly` (+ `mining`, `subtreeprocessor`), `services/blockpersister`, `services/legacy/netsync` all pass, and `go vet` is clean.

Two claims the change rests on both hold:

- Every production `Valid` caller passes a subtree store (`BlockValidation.go:1967, 2066, 2771`, `Server.go:1479`, `blockassembly/Server.go:1974`), so no real block loses the fee skip.
- Block assembly agrees with the validator: `createMerkleTreeFromSubtrees` sets `hashMerkleRoot = coinbaseTxIDHash` when there are no subtrees (`services/blockassembly/Server.go:2076-2077`), and `jobSubtreeHashes` and `job.Subtrees` are both empty in that case.

The new tests are not vacuous. I neutered each guard in turn and re-ran them: the three mismatch tests fail without the fix, and the honest-block counterweights keep passing, so the guards do not over-reject.

No blocking findings. Three test-quality points below.

---

**`services/blockvalidation/catchup_fork_handling_test.go:824-825`**

The two "competing equal-work chains" are the same chain, byte for byte. `CreateSyntheticBlocksFrom` is a pure function of `(prevHeader, length, firstHeight)` — `CreateSimpleCoinbaseTx(height)` is deterministic, the merkle root is that coinbase txid, and the timestamp derives from `prevHeader`. Both calls pass identical arguments.

I built both chains in a probe test against the same genesis header. All 10 headers matched, and both tips were `1fcfa01bf31f6ea68179b1b01481bca7757f66f70bd5c3b3b479a8e2a267ce6f`. So `target2.Hash() == target1.Hash()`, the two `GetBlockExists` mocks collide, and the second `catchup` call replays chain 1 after the test has already marked chain 1 as existing. The one assertion in that half, `assert.Equal(chain1[9].Hash(), firstSeenChain)`, is trivially true; the rest of the second half only calls `t.Logf`.

The old `CreateSyntheticChainFrom` was no better — it seeded timestamps from `time.Now().Unix()`, so two back-to-back calls collided unless a second boundary fell between them. The defect is pre-existing, but this diff makes it deterministic and adds a comment ("Create two different synthetic chains") that asserts the opposite.

Suggestion: give `CreateSyntheticBlocksFrom` a differentiator — a salt or tag folded into the coinbase script, or a per-chain timestamp offset — and pass a distinct value per chain. Then add a `require.NotEqual` on the two tips so the premise cannot silently rot again.

---

**`services/blockvalidation/quick_validate.go:313-317`**

The new guard in `quickValidateBlockAsync` has no test. Both new quick-path tests drive `quickValidateBlock`, which is reached from `Server.go:1637`. Catchup — the initial block download path — calls `quickValidateBlockAsync` (`catchup.go:1637`).

No reachable defect: the two branches are a four-line mirror and both are correct today. The risk is regression, since a future edit to the async pipeline can drop this call and leave the suite green.

Suggestion: add the mismatch case against `quickValidateBlockAsync`, with a `writeJobsChan` the test drains. It can reuse the fixture from `TestQuickValidateBlock_ZeroSubtreeMerkleRootMismatchRejected`.

---

**`services/blockvalidation/catchup_malicious_peer_test.go:495-530`**

Agreed with the reasoning in the description, and the new assertions are tighter than the old `successCount` tally — they do separate the honest peer from the Sybils by where each one fails. The lost property is also covered elsewhere: five other call sites in `catchup_test.go` assert `NoError` on `server.catchup`.

Suggestion: open a follow-up for the real coinbases of mainnet blocks 1-20 in the fixture, and reference it from the comment block so the weakened assertion carries its own exit plan.

---

Two nits, neither blocking:

- `services/blockvalidation/quick_validate.go:774-776` — `checkEmptyBlockMerkleRoot` wraps every `CheckMerkleRoot` error as `BlockInvalidError`, but `CheckMerkleRoot` can return a `StorageError` on a subtree/slice count mismatch (`model/Block.go:2210`). That is an internal inconsistency rather than a consensus violation, and `BlockInvalidError` marks a block permanently bad. Unreachable today — `SubtreeSlices` is always `make(..., len(b.Subtrees))` (`model/Block.go:1881`), so both lengths are zero together on every deserialized path.
- `model/block_zero_subtree_merkle_test.go:156-179` — `bindEmptyBlockMerkleRoot` repeats the `Nonce++` mining loop that `testhelpers.MineHeader` already provides. Cosmetic.

gokutheengineer (commented):
Assuming this gets rebased onto main with #1390, the merkle binding itself already exists there (CheckCoinbaseOnlyBodyBound in Block.Valid and both quick-validate entry points, plus the merkleRootChecked gate on the fee skip). So my comments are on what's left after that.

1. Error class. checkEmptyBlockMerkleRoot (quick_validate.go:771-775) returns BlockInvalid for a merkle mismatch. #1390 made every unbound-body failure BlockCorrupt so a truncated body from a peer doesn't get an honest hash persisted as invalid. This needs to be corrupt too, or just use CheckCoinbaseOnlyBodyBound and drop the helper.
2. RecomputeCoinbaseOnlyCounts. Main rejects TransactionCount > 1 on a no-subtree body; this PR overwrites it instead. We should pick one. Also, on the optimistic path AddBlock runs before Valid (BlockValidation.go:1932 vs :1967), so the recomputed values don't reach the stored row there anyway.
3. TestCatchup_SybilAttack. Main already fixed the fixture by building the honest chain synthetically, so the honest peer actually validates. This PR's version asserts the honest peer fails on merkle root, which is weaker. Keep main's version on rebase.
4. TestCatchup_CompetingEqualWorkChains. Main documents the identical chains as intentional (two peers, same tip, first-seen wins). The variant param changes what the test covers. If we want distinct equal-work chains, better as a new case.

ctnguyen (commented):
Review at commit [c21ca2a](https://github.com/icellan/teranode/commit/c21ca2a50c8cb5c059f37c9fc83dc07351090f3a)

Nice catch on the underlying gap — a coinbase-only block's merkle root, the cheapest one in the system to verify since it's just the coinbase txid, was the one body shape nothing ever checked, and the failing-first tests plus the careful reasoning in the comments are the right instinct. The one thing to keep front of mind: #1390 merged to main at 10:19 today, ~87 minutes *before* this head (c21ca2a, 11:47), and #1390 already lands the merkle binding in `Valid` and both quick-validate entry points. This branch is behind main only because its last merge-from-main (2bffc816, 09:40) predates the #1390 merge — so most of the work below is reconciling with what's now on main rather than the idea, which is sound.

## Blocking issues

### ChiR1 — Both unbound-body failures poison the block instead of triggering re-download
*Same as [@gokutheengineer's point 1 (14-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1730#pullrequestreview-5197472054).*
*Same as [@ordishs's point "wraps every CheckMerkleRoot error as BlockInvalidError" (14-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1730#pullrequestreview-5197025688).*

**Problem:** `checkEmptyBlockMerkleRoot` returns `BlockInvalidError` on *both* pre-binding checks — the coinbase-shape failure (`quick_validate.go:771`) and the merkle mismatch (`:775`) — and on the `Valid` path step 8b returns `CheckMerkleRoot`'s error raw, which is itself `errors.NewBlockInvalidError("... merkle root does not match")` (`model/Block.go:2383`). All three are *body-derived, pre-binding* failures: the received coinbase either isn't coinbase-shaped or doesn't hash to the header's merkle root, but you cannot distinguish a genuinely invalid block from a body a peer truncated or corrupted in transit. #1390 settled exactly this — the body-derived tier (size, coinbase shape, subtree shape, merkle-root comparison, CVE-2012-2459) is "never poison", and on main this check returns `NewBlockCorruptError`, deliberately not matched by `errors.Is(err, ErrBlockInvalid)`. @ordishs' nit found the narrower, currently-unreachable slice of this (a `StorageError` from the length-mismatch guard being wrapped as `BlockInvalidError`); my finding is the reachable pair — the coinbase-shape and consensus-mismatch returns a peer can trigger with any wire block, which the nit's proposed classification would also need to cover.

**Why it matters:** `BlockInvalidError` persists `invalid=true`, cascades to descendants and penalises the peer. On the optimistic-mining path the background `Valid` failure runs `errors.Is(err, errors.ErrBlockInvalid)` → `ReasonInvalidBlock` (`BlockValidation.go:1967-1975`), so one peer serving a corrupt or malformed body for an otherwise-valid hash can get the honest chain tip permanently rejected — the precise incident class #1390/#4692 exist to close (see [bsv-blockchain/teranode#1390](https://github.com/bsv-blockchain/teranode/pull/1390), [bitcoin-sv/teranode#4692](https://github.com/bitcoin-sv/teranode/issues/4692)). svnode reports `bad-txnmrklroot` through `CorruptionOrDoS` for the same reason. This PR's code contradicts that resolution.

**Fix:** Classify *every* unbound-body failure as corruption, not invalidity — ideally by adopting main's `CheckCoinbaseOnlyBodyBound`, which does this and removes the need for `checkEmptyBlockMerkleRoot`. If the helper is kept short-term, both returns must change, not just the merkle one:

```go
if !block.CoinbaseTx.IsCoinbase() {
    return errors.NewBlockCorruptError("[checkEmptyBlockMerkleRoot][%s] block coinbase tx is not a valid coinbase tx", block.Hash().String())
}
if err := block.CheckMerkleRoot(ctx); err != nil {
    return errors.NewBlockCorruptError("[checkEmptyBlockMerkleRoot][%s] merkle root mismatch", block.Hash().String(), err)
}
```

and on the `Valid` path reclassify the zero-subtree mismatch rather than returning `CheckMerkleRoot`'s `BlockInvalidError` verbatim. Tests should assert `ErrBlockCorrupt` for both the not-coinbase and the mismatch cases.

### ChiR2 — Reconcile deliberately with main; the count recompute inside `Valid` becomes a silent no-op after merge
**Problem:** This head predates a rebase onto the #1390 merge, and a real `git merge origin/main` conflicts in 5 of 10 files. Two of the changes here are genuine deltas over main, not duplication, and should be kept — but reconciled by hand rather than by whatever git auto-produces:

- **The count/size recompute is distinct behaviour.** Main's `CheckCoinbaseOnlyBodyBound` *rejects* `TransactionCount > 1` on a no-subtree body but does not normalise a count nor derive `SizeInBytes` at all; `RecomputeCoinbaseOnlyCounts` does both. (Whether normalising is the right call is ChiR3 — but it isn't redundant with main.)
- **The widened nil-store gate** (`else if len(b.Subtrees) == len(b.SubtreeSlices)`, covering an in-memory block that carries its own slices under a nil store) is also a genuine delta.
- **The `merkleRootChecked` conjunct on the below-checkpoint fee skip** is the clearest delta: main threads the flag into `checkBlockRewardAndFees` but its skip at `model/Block.go:1254` still has only three conjuncts and is not gated on it.

The verified hazard is narrower than "the feature disappears": git orders main's `if len(b.Subtrees) == 0 { CheckCoinbaseOnlyBodyBound() }` branch ahead of this PR's `else if len(b.Subtrees) == len(b.SubtreeSlices)` branch, so step 8c's `RecomputeCoinbaseOnlyCounts()` call inside `Valid` becomes reachable only when `len(b.Subtrees) > 0` — exactly where the function early-returns. That one call silently no-ops in the merged tree while still compiling. The sibling call inside `checkEmptyBlockMerkleRoot` auto-merges and still runs, so the recompute is not lost on the quick-validate path.

**Why it matters:** Merging as-is lands a version where the `Valid`-path count recompute is dead code, and risks losing the deltas worth keeping in the same merge. This needs a deliberate rebase, not an auto-resolution.

**Fix:** Rebase onto current main. Fold `checkEmptyBlockMerkleRoot`/the 8b merkle check onto main's `CheckCoinbaseOnlyBodyBound` (per ChiR1), decide the count-recompute question against main's reject-only stance (ChiR3), and preserve the fee-skip conjunct with its dedicated test. Re-verify the reduced diff against main, not against the pre-#1390 base.

## Non-blocking issues

### ChiR3 — `RecomputeCoinbaseOnlyCounts` overwrites counts main treats as a corruption signal, and is inert on the optimistic path
*Same as [@gokutheengineer's point 2 (14-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1730#pullrequestreview-5197472054).*

**Problem:** `TransactionCount` / `SizeInBytes` arrive from the wire and are persisted as given. This PR silently rewrites them to `1` / the derived size for a no-subtree body (`model/Block.go`, `RecomputeCoinbaseOnlyCounts`). Main takes the opposite stance: `CheckCoinbaseOnlyBodyBound` rejects `TransactionCount > 1` as `BlockCorrupt` ("a free consistency check on a shape that can never be honest"). Once the merkle root binds the body, a wire count of 5 on a coinbase-only body is provably a lie — main flags it, this PR discards the signal by overwriting. Separately, on the optimistic-mining path `AddBlock` (`BlockValidation.go:1932`) runs before the background `Valid` (`:1967`), so the recompute inside `Valid` never reaches the persisted row there — the row keeps the sender's values, the shape this change meant to settle.

**Why it matters:** Two behaviours for the same body shape (reject vs. overwrite) shouldn't coexist, and the overwrite doesn't take effect on the path that most needs it. These are informational columns, not consensus, so it isn't a correctness hole — but the divergence from main should be resolved deliberately.

**Fix:** Prefer main's reject-as-corrupt stance and drop the overwrite; if a recompute is still wanted, apply it before `AddBlock` on the optimistic path so the stored row reflects it.

### ChiR4 — `TestCatchup_SybilAttack` is weakened where main strengthened it
*Same as [@gokutheengineer's point 3 (14-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1730#pullrequestreview-5197472054).*
*Same as [@ordishs's point 3 (14-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1730#pullrequestreview-5197025688).*

**Problem:** This PR rewrites the test to assert the honest peer *fails* on merkle root (its synthetic coinbase can't satisfy the mainnet fixture's real headers) and drops the `successCount == 1 / failCount == 5` tally. The two reviewers overlap on the subject but disagree on the remedy, so I arbitrate: @ordishs accepts the weakening and asks only for a follow-up to source real coinbases; @gokutheengineer says keep main's version. The evidence favours @gokutheengineer — main (catchup_malicious_peer_test.go:224-234) builds the honest chain *synthetically* so the honest peer genuinely validates, and keeps `assert.Equal(t, 1, successCount, "Only honest peer should succeed")` (main:589). So @ordishs' follow-up is already unnecessary on main, and this diff would regress a test main has since made stronger.

**Why it matters:** The Sybil test exists to prove the honest peer is *accepted* while attackers are turned away. Asserting the honest peer fails inverts the property it protected, and it will conflict with main on rebase.

**Fix:** On rebase, keep main's synthetic-honest-chain version and its acceptance assertion; discard this PR's weakened form.

### ChiR5 — `variant` on `CreateSyntheticBlocksFrom` repurposes a test main documents as intentionally identical
*Same as [@ordishs's point 1 (14-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1730#pullrequestreview-5197025688).*
*Same as [@gokutheengineer's point 4 (14-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1730#pullrequestreview-5197472054).*

**Problem:** Both reviewers correctly flagged that the two "competing equal-work chains" were byte-identical (a pure function of identical arguments) and that the old comment claimed the opposite. They diverge on the fix. @ordishs asked for a differentiator *and* a `require.NotEqual` on the two tips; this PR implements the differentiator (the `variant` parameter offsets timestamps) but adds no such assertion — there is no `require.NotEqual` in catchup_fork_handling_test.go, so the premise it guards against can still silently rot. More importantly, main (catchup_fork_handling_test.go:820-824) has since *documented* the identical chains as deliberate: two peers offering the same tip over distinct URLs, exercising first-seen acceptance of equal-work, same-hash chains. The `variant` offset turns that into a different scenario (two genuinely distinct tips), which is not what the test on main is for.

**Why it matters:** The premise rot both reviewers found is real, but main resolves it by documenting the intent, not by making the chains differ. Landing `variant` here changes the test's meaning and conflicts with main; and even on its own terms it's a half-fix without the assertion @ordishs asked for.

**Fix:** On rebase, take main's documented same-tip version. If a distinct equal-work fork case is genuinely wanted, add it as a separate test (per @gokutheengineer point 4) with a `require.NotEqual` on the two tips.

### ChiR6 — New `quickValidateBlockAsync` guard has no test
*Same as [@ordishs's point 2 (14-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1730#pullrequestreview-5197025688).*

**Problem:** All three new tests in `quick_validate_zero_subtree_test.go` (`ZeroSubtreeMerkleRootMismatchRejected`, `HonestEmptyBlockAccepted`, `CoinbaseOnlyCountsRecomputed`) drive the synchronous `quickValidateBlock`; the mirror guard added to `quickValidateBlockAsync` (`quick_validate.go:313-317`) — the path catchup actually uses — is untested. No live defect (the two branches are a four-line mirror and both correct), but a future edit to the async pipeline could drop the call and leave the suite green.

**Why it matters:** The async path is the initial-block-download path; an untested guard there is the one most likely to silently regress.

**Fix:** Main already covers both entry points in `quick_validate_coinbase_only_binding_test.go`, so the actionable move is to retain that coverage through the reconciliation rather than write a duplicate here. If this PR lands standalone, add the mismatch case against `quickValidateBlockAsync`, draining a `writeJobsChan`, reusing the fixture from `TestQuickValidateBlock_ZeroSubtreeMerkleRootMismatchRejected`.

### ChiR7 — `CreateSyntheticBlocksFrom` computes `SizeInBytes` one byte short of the arithmetic this PR makes authoritative
**Problem:** The helper sets `SizeInBytes: uint64(80 + len(coinbaseTx.Bytes()))`, but `RecomputeCoinbaseOnlyCounts` — the size formula this same PR promotes to the source of truth — is `80 + util.VarintSize(b.TransactionCount) + uint64(b.CoinbaseTx.Size())`, and `util.VarintSize(1) == 1` (`util/varint.go:6`). The fixture omits the transaction-count varint byte.

**Why it matters:** Minor and self-healing on any path that runs `RecomputeCoinbaseOnlyCounts` after receipt, but the fixture ships a value one byte off from the formula the PR itself declares canonical — a confusing inconsistency to leave in a test helper.

**Fix:** Include the transaction-count varint byte:

```go
SizeInBytes: 80 + util.VarintSize(1) + uint64(coinbaseTx.Size()),
```

## Recap

| --- | --- | --- | --- |
| ChiR1 | Unbound-body failures poison block | YES | 88% |
| ChiR2 | Reconcile with main; 8c recompute no-op | YES | 78% |
| ChiR3 | Count recompute diverges, inert optimistically | | 45% |
| ChiR4 | Sybil test weakened vs main | | 35% |
| ChiR5 | `variant` contradicts main's intent | | 30% |
| ChiR6 | Async guard untested | | 22% |
| ChiR7 | Fixture size off by one byte | | 12% |

services/blockvalidation/quick_validate.go
```diff
@@ -743,6 +755,33 @@ func (u *BlockValidation) processBlockSubtreesPipelineAsync(ctx context.Context,
...
+// transitively: the pinned hash covers the header, the header covers the merkle root,
+// the merkle root covers the body. That last step has to actually be evaluated. For a
+// block with subtrees validateSubtrees does it; a zero-subtree block reaches neither
+// validateSubtrees nor model.Block.Valid, so nothing here read the coinbase at all,
+// even though its outputs become UTXOs once the block commits.
+//
+// For an empty block the check is one hash comparison: the merkle root is the coinbase
+// txid, which model.Block.CheckMerkleRoot computes in its no-subtree branch.
+func checkEmptyBlockMerkleRoot(ctx context.Context, block *model.Block) error {
+	if !block.CoinbaseTx.IsCoinbase() {
```
freemans13, line 770: This branch has no conflict with `main`, so after the merge it runs directly behind `main`'s `block.CheckCoinbaseOnlyBodyBound()`, which sits at the top of both `quickValidateBlock` and `quickValidateBlockAsync`. At that point:

- the `CheckMerkleRoot` branch below is unreachable, because `main` has already rejected any mismatch as `BlockCorrupt`;
- this `IsCoinbase()` check is still reachable and still useful. `main`'s quick path never checks coinbase shape for a coinbase-only body, only `len(Inputs) != 0`.

Suggestion for the reconcile: drop the helper, and keep only the shape check, inline after `CheckCoinbaseOnlyBodyBound`. Once the body is bound, a non-coinbase transaction as the block's only transaction is committed by the header, so `BlockInvalid` is correct there. Here, ahead of the merkle comparison, it is unbound and should be corrupt (ChiR1). Worth one test that pins it: a coinbase-only block whose only transaction is a normal spend, with the header merkle root set to that txid, expecting `ErrBlockInvalid`.

freemans13 (changes requested):
Requesting changes, read at c21ca2a. The idea is right, but #1390 landed the same binding on `main` this morning, so the job now is to reconcile with it. I don't want to pile on what @ctnguyen and @gokutheengineer already wrote, so this covers only what I checked myself and one thing nobody has raised yet.

**Checked independently, and I agree:**

- `main` already binds a coinbase-only body in `Block.Valid` and at both quick-validate entry points, through `model.Block.CheckCoinbaseOnlyBodyBound`, and it classifies a mismatch as `BlockCorrupt`. That means re-download, not a persisted `invalid=true`. This PR returns `BlockInvalid` for the same failure, which contradicts #1390 (ChiR1).
- `git merge-tree` against `upstream/main` (13897b045) reports conflicts in 5 files: `model/Block.go`, `model/Block_test.go`, `catchup_fork_handling_test.go`, `catchup_malicious_peer_test.go` and `testhelpers/block_helpers.go`.
- The `merkleRootChecked` conjunct on the below-checkpoint fee skip is a genuine delta. On `main` that skip still has three conjuncts (ChiR2).

**New: `quick_validate.go` merges cleanly, and that is the trap.** Git reports no conflict there, so resolving the five conflicted files and building will not surface it. In the merged tree both entry points run `main`'s `CheckCoinbaseOnlyBodyBound` first, then this PR's `checkEmptyBlockMerkleRoot` about 40 lines later: merged lines 267 and 304 in `quickValidateBlock`, 376 and 414 in `quickValidateBlockAsync`. The helper's merkle branch becomes unreachable, since `main`'s check has already rejected any mismatch. Its doc comment ("nothing here read the coinbase at all") also becomes false. So the reconciliation has to delete the helper on purpose, because the merge won't.

One part of the helper is worth keeping. On `main`, the quick path never checks that a coinbase-only block's `CoinbaseTx` is coinbase-shaped. It only checks for non-empty inputs, which is `BlockIncomplete`, plus the scriptSig length bound. `IsCoinbase()` adds real coverage. Placed after `CheckCoinbaseOnlyBodyBound`, it runs on a body that is already bound to the header, so `BlockInvalid` is the right class there. In this PR as written it runs *before* the merkle comparison, on an unbound body, and that is why ChiR1 calls it corrupt. The order decides the class. Inline comment below.

**Verified locally at c21ca2a:** the new and changed tests pass (`-run 'ZeroSubtree|CoinbaseOnly|HonestEmpty|SkipsBelowHardcodedCheckpoint|DupTxDetected' ./model/` and `-run 'ZeroSubtree|CoinbaseOnly|HonestEmpty|TestCatchup_SybilAttack|TestCatchup_CompetingEqualWorkChains' ./services/blockvalidation/`). I did not run the full suites; @ordishs did that at 2bffc81. I did not build or test a merged tree, because it doesn't compile until the five conflicts are resolved. Upstream CI shows only the two Socket checks on this head.

icellan:
Closing in favour of #1738.

@ctnguyen @gokutheengineer @freemans13 were right that #1390 landed the same binding on main this morning, ~87 minutes before this head, and that most of this PR duplicates it. Rather than push five conflicts through a history that no longer adds much, #1738 starts from current main with only the reconciled remainder.

What survived, and why:

- **The `merkleRootChecked` conjunct on the below-checkpoint fee skip.** Confirmed as a genuine delta — main threads the flag into `checkBlockRewardAndFees` but its skip still had three conjuncts.
- **The `IsCoinbase()` shape check**, moved after `CheckCoinbaseOnlyBodyBound` exactly as @freemans13 described: once the body is bound the transaction is the miner's own committed body, so `BlockInvalid` is right there, where ahead of the binding it was wrong. Now covered at both entry points, including `quickValidateBlockAsync` (@ordishs's point 2, @ctnguyen's ChiR6).

What was dropped:

- `checkEmptyBlockMerkleRoot` — superseded, and ChiR1 was correct that `BlockInvalid` on an unbound body contradicts #1390. On the optimistic path that could have got an honest tip persisted invalid off one corrupt download.
- `RecomputeCoinbaseOnlyCounts` — main's reject-as-corrupt is the better stance, and the recompute was inert on the optimistic path anyway, where `AddBlock` runs before `Valid` (ChiR3).
- The `variant` parameter and the `TestCatchup_SybilAttack` rewrite — main documents the identical chains as deliberate, and fixed the Sybil fixture properly by building the honest chain synthetically (ChiR4, ChiR5). I had concluded that was impossible; it wasn't.

Also worth recording: ChiR2's merge trap was real. `quick_validate.go` merges cleanly, so the merged tree would have run `CheckCoinbaseOnlyBodyBound` and then a dead `checkEmptyBlockMerkleRoot` 40 lines later, and the step-8c count recompute inside `Valid` would have become reachable only where the function early-returns. Both would have compiled and passed.

One judgement call went against a suggestion in review: the widened nil-store merkle gate is also dropped, not kept. It removes the state `TestBlock_Valid_SkipRefusedWhenMerkleRootNotChecked` is deliberately built on, and no production caller passes a nil store.

Issue #1731, opened 2026-09-14
Testnet sync rejects block 149 in v0.15.9-beta-2; fix main and backport to v0.15


oskarszoon:
## Problem

Fresh testnet sync on `v0.15.9-beta-2` (commit `69132e684`) stops at height **148**. The valid historical block at height **149** is rejected with:

```text
incorrect difficulty bits: got 1d00ffff, expected 1c7fff80
```

Block hash: `00000000291d8e6f5d0d2a59de8f0f206917f3e00ff53edc8f6dcaddd61f3fe9`.

The rejection is persisted as an invalid block. Descendants then fail with `parent block is invalid`, accompanied by repeated waits for the parent's `mined_set` status. These are consequences of the first incorrect difficulty rejection.

This needs work on **main and an explicit backport to `release/v0.15`**. The branches differ in their checkpoint handling, so the backport may need a different implementation.

## Root cause

Two behaviours combine:

1. **Existing historical-difficulty limitation.** `CalcNextWorkRequired` implements the modern 144-block DAA without selecting the historical retarget rules. It returns the minimum difficulty while the parent height is below `144 + 4`; parent height 148 is the first point at which it performs that calculation. Applying it to testnet block 149 produces `1c7fff80`, whereas this historical block correctly uses `1d00ffff`. The calculator is unchanged between beta-1 and beta-2. See [beta-2 Difficulty.go](https://github.com/bsv-blockchain/teranode/blob/69132e684/services/blockchain/Difficulty.go#L66).

2. **New beta-2 checkpoint-proof requirement.** The checkpoint ancestry change in #1723 requires authenticated ancestry before skipping the historical difficulty calculation. Legacy sync successfully downloads and verifies headers through checkpoint **546**, but its `ProcessBlock` RPC does not carry that provenance into block validation. The legacy validation options leave `checkpointProven` false, and checkpoint 546 is not yet present in blockchain storage, so the stored-ancestry fallback cannot establish it either. See [legacy RPC handoff](https://github.com/bsv-blockchain/teranode/blob/69132e684/services/legacy/netsync/handle_block.go#L264), [validation options](https://github.com/bsv-blockchain/teranode/blob/69132e684/services/blockvalidation/Server.go#L1530), and [difficulty-skip gate](https://github.com/bsv-blockchain/teranode/blob/69132e684/services/blockvalidation/BlockValidation.go#L2491).

[...]

PR #1732, opened 2026-09-14
fix: validate historical difficulty before DAA

oskarszoon:
Fresh testnet sync must validate block 149 with its historical `1d00ffff` target. The calculator and native-catchup precheck applied modern DAA at parent height 148 and returned `1c7fff80`, rejecting the block.

Apply the original 2016-block retarget and testnet minimum-difficulty restoration before DAA activation, with mainnet EDA after UAHF. The native header precheck does not calculate or reject expected-difficulty mismatches below DAA activation; its early expected-difficulty rejection applies only at/after activation. Historical headers proceed to full-block validation, which calculates their expected difficulty even while building the checkpoint prefix. STN retains its existing difficulty behavior.

Historical testnet restoration reads at most 32 ancestors per query and checks linkage across batches. One value-owned cache entry reuses a restored target only for matching parent/ancestor hashes and heights within the same retarget interval. Retarget and delayed-candidate rules retain priority. The early ordinary-target/boundary return ensures every cached parent was itself a non-boundary minimum-difficulty block; a comment records this invariant. Failed reads cannot populate the cache, and the existing difficulty-cache setting disables reuse.

Regression coverage:

- Real testnet headers through 547 exercise the calculator and confirm that the native precheck defers these historical headers; separate boundary tests retain incorrect-target rejection at DAA activation. A full legacy `ProcessBlock` replay uses real block/subtree validation, SQLite chain/UTXO stores and temporary blob stores, including the checkpoint-prefix case at block 149. Only that full replay is tagged `longtest`; `make longtest` runs it with race detection.
- Ordinary PR tests now include real mainnet headers 477792–504031. The SQLite-backed calculator matches every mined child target from 478558 through 504031: 25,474 comparisons, 13 periodic retargets and all 43 EDA increases. Header linkage, PoW, pinned hashes and the fixture checksum are checked. [Provenance, reproduction and scope](https://github.com/bsv-blockchain/teranode/blob/465386d08f480ffa18486b8014e4c5f5a3419d0c/services/blockchain/testdata/README.md) are committed alongside the fixture. This is a header/calculator regression, not full mainnet block replay.
- Synthetic cases cover retarget clamps, missing/discontinuous history, activation boundaries, fork/concurrent cache queries, delayed candidates, returned-value ownership and disabled caching. Over 400 sequential minimum-difficulty parents, requested headers fall from 85670 to 21; with 70-parent delayed gaps, from 1136 to 496. Per-block best-tip resets are included. Production retains the calculator; `LocalClient` creates one per call, so the legacy replay does not measure cache reuse.
- Existing catchup setup finishes before subscription readers start. The pruning regression waits for actual parent-record deletion, avoiding stale completion notifications.

The pre-UAHF gate is a canonical-history optimization, absent from SV Node's ungated EDA check. An independent Python scan of linked, PoW-verified mainnet headers 0–510000 compared historical targets for children 1–504031: zero mismatches, 250 retargets and 43 EDA increases. For each applicable pre-UAHF parent, it subtracted the sixth ancestor's 11-header MTP from the parent's 11-header MTP. All 446081 applicable parents at heights 32256–478557 stayed below 43200 seconds; the maximum was 26789 at parent 74648. The fixture README records the source, complete archive checksum and arithmetic. This full historical scan is independent Python evidence; the committed Go replay covers the EDA-era range above.

Validation on `465386d08`: full `-race -tags "testtxmetacache longtest" -count=1` suites passed for blockchain and legacy/netsync, as did the block-validation difficulty/checkpoint regressions. Vet, staticcheck and changed-line lint passed. The new mainnet test takes 5.97 seconds without race instrumentation. Temporary compiler overlays confirmed it fails at 478577 with EDA disabled and at 479808 with the retarget ancestor shifted by one block. The documented fixture download reproduces the pinned checksum.

Deferred: extending assertions into the fixture's earlier pre-UAHF prefix, disabled-cache/per-call-calculator performance, STN parameter/history reconciliation, future-network retarget parameterization, and Sonar's historical-function complexity refactor. Earlier dependency/toolchain scanners reported existing advisories outside this diff.

Refs #1731. Backport #1733 merged as `1d8067bf2`; the memoization and mainnet-fixture follow-ups are confined to this main PR.


services/blockchain/Difficulty.go
```diff
@@ -159,6 +166,83 @@ func (d *Difficulty) CalcNextWorkRequired(ctx context.Context, blockHeader *mode
...
+		}
+		target.Mul(target, big.NewInt(elapsed))
+		target.Div(target, big.NewInt(timespan))
+	} else if params.ReduceMinDifficulty {
+		if blockTime > int64(parent.Timestamp)+2*int64(params.TargetTimePerBlock.Seconds()) {
+			return d.powLimitnBits, nil
+		}
+		// Walk this parent's ancestry, stopping at the last non-special target
+		// or at a retarget boundary. Missing history must not become easy work.
+		for height%interval != 0 && parent.Bits == *d.powLimitnBits {
```
freemans13, line 200: Non-blocking, performance. Every step of this loop is its own `GetBlockHeader` query. `StoreBlock` clears the SQL store's response cache, so during sync none of these lookups hit the cache.

I replayed real testnet history below the DAA fork through this function. The loop made 36,571,899 lookups in total. 18,210 blocks walked 1,000 steps or more, and the worst walked 2,014. I haven't timed it on Postgres. At 0.1 ms per query that comes to about an hour of blockchain-service time over a testnet sync, with up to 2,014 round trips for a single block.

The walk stops at the retarget boundary, so it never needs more than `height%interval + 1` headers. A single `GetBlockHeaders(ctx, parent.Hash(), uint64(height%interval)+1)` returns them newest first from one index scan, and it goes through the chain-walk cache, which `StoreBlock` doesn't clear. Scan that slice for the first header that isn't at the pow limit or sits on the boundary. A length check on the result keeps the missing-history error.

services/blockchain/Difficulty.go
```diff
@@ -159,6 +166,83 @@ func (d *Difficulty) CalcNextWorkRequired(ctx context.Context, blockHeader *mode
...
+			previous, _, err := d.store.GetBlockHeader(ctx, parent.HashPrevBlock)
+			if err != nil || previous == nil {
+				return nil, errors.NewStorageError("[Difficulty] restoring historical testnet target", err)
+			}
+			parent = previous
+			height--
+		}
+		return &parent.Bits, nil
+	} else {
+		if height < params.UahfForkHeight || height < 6 || parent.Bits == *d.powLimitnBits {
```
freemans13, line 210: Nit. SV Node's `GetNextEDAWorkRequired` has no UAHF check: it applies the 12-hour rule to every non-retarget block before DAA activation. So `height < params.UahfForkHeight` exists only in teranode.

I replayed real mainnet with `UahfForkHeight = 0` and it still matched every block up to 504031. No pre-UAHF block ever met the EDA condition, so on the real chain the gate changes nothing, and it saves about 446,000 `GetBlockHeaders` reads. Worth keeping.

Could the comment say it's an optimisation that only matches SV Node because of that history? The earlier bot review on this PR described the gate as mirroring SV Node, and the next reader will probably assume the same.

freemans13 (approved):
Approving. I checked the new calculator against the full real header history, not just the fixtures, and it reproduces every block's nBits.

What I ran, at `d9cd644`:

- **Real-history replay.** I fed every real header from WhatsOnChain's header archive through `CalcNextWorkRequired`: mainnet heights 1 to 504031 and testnet heights 1 to 1,188,697, which is everything below each network's DAA activation. The store was an in-memory fake serving the three lookups the calculator makes. Zero mismatches on either network. That covers what the PR's fixtures can't: 250 mainnet retargets, the whole mainnet EDA window, and 589 testnet retargets. The 0 to 547 fixture contains no retarget and no EDA.
- **The harness can fail.** With target spacing shortened by 10%, it flags the first real difficulty change at mainnet height 32256 and every retarget after it.
- **The legacy test pins the bug.** With the historical branch in `CalcNextWorkRequired` disabled, `TestLegacyHistoricalTestnetSync` fails at block 149 with `got 1d00ffff, expected 1c7fff80`. It passes again once the branch is restored.
- **SV Node comparison.** I read the calculator against SV Node's `GetNextEDAWorkRequired` and `CalculateNextWorkRequired`. The retarget ancestor, clamp, testnet walk and EDA window all match. The one difference is the UAHF gate, noted inline.

Why this reaches past testnet: legacy sync sends every block through `ValidateBlock`, checkpointed ones included, and `skipExpectedDifficulty` now returns false at or below `DaaForkHeight`. So a mainnet sync from genesis runs this calculator on every block up to 504031. CI doesn't cover that, because `legacy-sync` runs on regtest, where difficulty adjustment is off. It may be worth adding a handful of real mainnet vectors to testdata: a retarget, an EDA adjustment, and the 504031 boundary.

Two non-blocking comments inline:

1. The testnet minimum-difficulty walk makes one uncached `GetBlockHeader` query per step, 36.6M over the testnet pre-DAA range. One range read would do the same job.
2. SV Node has no pre-UAHF gate on the EDA. The gate is harmless on the real chain and saves work, but the comment should say why it's safe.

What I didn't check: I didn't time the walk on Postgres, and I didn't run a real testnet or mainnet sync. The replay exercises the calculator alone. Only the 547-block legacy test covers the full block-validation path.

services/blockchain/Difficulty.go
icellan, line 200: This walk issues one `GetBlockHeader` per ancestor — up to 2015 sequential round-trips — for every block validated, not just for rare gaps. `skipExpectedDifficulty` now returns false across the whole ~1.19M-block pre-DAA testnet range, and long runs of consecutive `1d00ffff` blocks since the last retarget boundary are the normal shape of testnet3, so this is the common path during IBD. The SQL response cache softens it but is TTL-bounded.

A single bulk `GetBlockHeaders(parent.Hash(), height%interval+1)` gets the same answer in one query.

services/blockchain/Difficulty.go
icellan, line 210: The `height < params.UahfForkHeight` gate has no counterpart in SV Node: it routes every pre-DAA block through `GetNextEDAWorkRequired`, where `nBits == powLimit` is the only guard. The comment at line 71 says this mirrors SV Node, and the PR description doesn't mention the extra gate.

Failure mode if a qualifying block exists: a pre-UAHF parent with above-minimum bits where `MTP(parent) - MTP(parent-6) >= 12h`. We'd expect `parent.Bits` while the chain carries the 25%-eased target, reject a canonical block, and halt IBD permanently.

I don't think such a block exists on mainnet — the largest known inter-block gap is ~1h40m, so six blocks can't plausibly median out to 12h — but that's an assumption the PR neither states nor tests. Either verify it against real history and say so in a comment, or drop the gate and match the reference.

icellan (commented):
Traced `calcHistoricalWorkRequired` against SV Node's `GetNextWorkRequired` / `CalculateNextWorkRequired` / `GetNextEDAWorkRequired`. Retarget-ancestor off-by-one, branch ordering, MTP window slicing under the short-chain clamp, and pow-limit clamping all check out, and the `<`/`<=` boundaries are consistent across the calculator, `skipExpectedDifficulty` and catchup. Two things worth addressing before merge, inline.

services/blockchain/Difficulty.go
```diff
@@ -67,6 +68,12 @@ func (d *Difficulty) CalcNextWorkRequired(ctx context.Context, blockHeader *mode
 		return &blockHeader.Bits, nil
 	}
 
+	// Activation is measured at the parent, as in SV Node's GetNextWorkRequired.
+	// Historical blocks must not be evaluated with the modern 144-block DAA.
+	if blockHeight < d.settings.ChainCfgParams.DaaForkHeight {
+		return d.calcHistoricalWorkRequired(ctx, blockHeader, blockHeight, currentBlockTime)
```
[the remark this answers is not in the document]
oskarszoon: Fixed in the current branch: the native header precheck defers historical mainnet/testnet blocks, and full-block validation always checks those historical targets. The real testnet header regression and SQLite checkpoint-boundary tests pass. Resolving the remaining open thread; STN retains its existing DAA behavior.

services/blockchain/Difficulty.go
oskarszoon: Fixed in 7344628bd. Historical testnet restoration now uses one bounded `GetBlockHeaders` call, starting at the predecessor because the parent is already available. It stops at the first ordinary target or the retarget boundary, and returns `ErrNotFound` if the required ancestry is absent. SQLite regressions cover a run of minimum-difficulty blocks, the retarget boundary, and missing history. Consumed headers are checked against the requested ancestry so a concurrent reorg cannot substitute a different branch.

oskarszoon: Fixed in 7344628bd. Historical testnet restoration now uses one bounded `GetBlockHeaders` call, starting at the predecessor because the parent is already available. It stops at the first ordinary target or the retarget boundary, and returns `ErrNotFound` if the required ancestry is absent. SQLite regressions cover a run of minimum-difficulty blocks, the retarget boundary, and missing history. Consumed headers are checked against the requested ancestry so a concurrent reorg cannot substitute a different branch.

services/blockchain/Difficulty.go
oskarszoon: Kept the optimization and documented why it agrees with canonical mainnet history in 7344628bd. I independently checked linked, PoW-verified WhatsOnChain headers: the ungated historical calculation matches all 504031 children through DAA activation, including 250 retargets and 43 EDA increases. Across the 446081 applicable pre-UAHF parents, no six-block MTP delta reaches 43200 seconds; the maximum is 26789 seconds at parent 74648. This is an optimization for that recorded history, not a UAHF guard present in SV Node. Source: [WhatsOnChain mainnet header archives](https://api.whatsonchain.com/v1/bsv/main/block/headers/resources).

oskarszoon: Kept the optimization and documented why it agrees with canonical mainnet history in 7344628bd. I independently checked linked, PoW-verified WhatsOnChain headers: the ungated historical calculation matches all 504031 children through DAA activation, including 250 retargets and 43 EDA increases. Across the 446081 applicable pre-UAHF parents, no six-block MTP delta reaches 43200 seconds; the maximum is 26789 seconds at parent 74648. This is an optimization for that recorded history, not a UAHF guard present in SV Node. Source: [WhatsOnChain mainnet header archives](https://api.whatsonchain.com/v1/bsv/main/block/headers/resources).

oskarszoon:
Mainnet-fixture update: `465386d08` supersedes the earlier mainnet-fixture deferral below. The ordinary-suite real-SQLite regression covers every mined target from 478558 through 504031, including all 43 EDA increases and 13 periodic retargets. [Current evidence and dispositions](https://github.com/bsv-blockchain/teranode/pull/1732#issuecomment-5699933134).

Processed the inline feedback in 7344628bd; the same calculator and test fixes are carried to #1733.

- Batched historical testnet restoration; preserved STN's previous difficulty behavior; clarified the pre-UAHF optimization using an independent real-history check.
- Replaced cancellation-based missing-history coverage with actual absent/truncated SQLite history, and copied the legacy harness checkpoint locally.
- Moved catchup validator initialization after the mock chain setup in 03c6ef755, removing a test-only race exposed by the release suite. Ten consecutive targeted race runs pass.
- Deferred adding more committed mainnet fixtures to keep this fix small. The independent Python mainnet history check covers every child through height 504031, with zero mismatches; it does not execute the Go calculator. Compact retarget/EDA/activation vectors remain useful follow-up coverage.

Validation: full affected-package race suites and final historical-calculator race regressions passed; `go vet`, `staticcheck` and lint passed. The historical header regression now takes about 13 seconds under `-race` on this machine. Batch ancestry is checked before using any returned target or MTP window.

Re-requesting reviews from the reviewers who raised these points. Copilot's previous attempts were quota-blocked, so I am not retrying it here.

CI follow-up: chainintegrity passed on retry. Prunertest failed again because its wait consumed a queued height-1 completion before the parent's height-10 deletion. Fixed in 45a23255c: poll the parent metadata record until the store returns `ErrTxNotFound`, with the existing 10-second limit. Other storage errors cannot satisfy the assertion. All eight prunertest cases passed locally under `-race`; vet, staticcheck and lint passed. Independent review found no blocking issues. [Prunertest passed on the new commit](https://github.com/bsv-blockchain/teranode/actions/runs/34958848685/job/104347331663). The same test fix is included in #1733 as eef7651c5.

Latest follow-up in 7b93b6dea: ancestry queries are capped at 32 headers to avoid whole-interval reads for short minimum-difficulty runs. The full legacy replay is behind `longtest` and explicitly included in `make longtest`; ordinary calculator/native-header tests remain enabled. The ancestry cache now follows in main-only commit 1a6879cf1; additional committed real-mainnet vectors remain deferred.

The repeated-scan finding from ordishs is addressed on main in 1a6879cf1. A single value-owned restoration entry is keyed by parent hash and height; linked walks can stop at that verified ancestor across delayed-candidate gaps. Retarget and delayed-block rules retain priority, missing history does not publish a result, and the existing cache setting is honored. Real-SQLite requested-header counts fell from 85670 to 21 over 400 sequential parents and from 1136 to 496 with 70-parent delayed gaps, including per-block best-tip cache resets. Race tests cover forks/concurrent reads, return-value ownership, timestamp ordering, boundaries and errors. Independent consensus and concurrency reviews found no actionable issue.

Previous-head CI record: all eight prunertest cases passed. The first [unit job](https://github.com/bsv-blockchain/teranode/actions/runs/35074448497/job/104723380194) failed in unchanged `TestDaemon_Start_AllServices` with an address-in-use startup error. Three isolated race runs passed, and the [single CI retry](https://github.com/bsv-blockchain/teranode/actions/runs/35074448497/job/104727486546) passed 14977 tests, with 46 skipped. The separate fixture issue remains deferred; a retry is not its fix.

Final check on `1a6879cf1`, more than 15 minutes after publication: 37 checks passed, one skipped, no pending or failed checks. All eight [prunertest cases passed](https://github.com/bsv-blockchain/teranode/actions/runs/35079634260/job/104740250274); the [unit job](https://github.com/bsv-blockchain/teranode/actions/runs/35079634296/job/104740249935) completed 15047 tests with 47 skipped. The benchmark report found no regressions. Inline review threads are resolved, and reviews are requested from ordishs, icellan and freemans13.

Sonar's remaining `go:S3776` finding concerns the cognitive complexity of `calcHistoricalWorkRequired` (77 against a threshold of 50). Its quality gate passed. Splitting the historical algorithms into smaller helpers is deferred to a separate refactor to keep the consensus fix focused; the finding remains open in Sonar. The automated review's short EDA-window observation is unreachable with the configured network activation heights. Its recovery-doc reference is stale: this diff contains a fixture README, not a recovery guide.

Fresh independent consensus, defensive/concurrency, and Go/test reviews of the final commit found no actionable issues. These were static reviews; runtime evidence is the affected-package race suites and current CI recorded above. The approved backport #1733 is already merged as `1d8067bf2`; this cache follow-up remains confined to main.

services/blockchain/Difficulty.go
icellan: Confirmed fixed — one round-trip instead of ~1000, and the continuity check against a main-chain range read racing a reorg is a good addition.

One follow-up on the same line, non-blocking. I measured the repo's own fixture `services/blockchain/testdata/testnet_headers_0_547.bin`: all 548 headers carry `1d00ffff`, 100%, zero non-powLimit blocks. So on early testnet the loop never finds an ancestor to stop at and always runs the full distance to the retarget boundary — the batch is ~`height%2016` rows every time, averaging ~1000. Measured cost in sqlitememory is ~0.55us/header (559us@1000, 1.18ms@2000) plus an uncached double-SHA256 per header for the continuity check. Across most of the 1.19M pre-DAA blocks that is minutes on sqlite and hours on Postgres.

Chunking doesn't help here — with runs this long it just becomes 30-60 sequential fetches. What does: while every block in an interval carries powLimit the answer is invariant across consecutive blocks, and it is simply powLimit. Memoise per retarget interval, or short-circuit the all-powLimit case, instead of re-reading the window per block.

Caveat on my evidence: the fixture spans 0-547, inside the first retarget interval, so it establishes the early range rather than all 1.19M blocks.

services/blockchain/Difficulty.go
icellan: The comment you added states the reasoning, which is what I asked for, so I'm fine with the gate staying.

One thing the wording makes load-bearing: "canonical pre-UAHF mainnet never reaches the EDA threshold" is a claim about the canonical chain. On a PoW-valid non-canonical pre-UAHF fork with a >=12h MTP window, we'd reject a block SV Node accepts.

What would close it: `skipExpectedDifficulty` now forces the expected-nBits check for every pre-DAA block, so a divergence anywhere in mainnet 478559-504031 — the Aug-Nov 2017 window where EDA genuinely fired repeatedly — hard-fails IBD where the checkpoint skip previously absorbed it. That range currently has synthetic coverage only (`TestDifficultyHistoricalEDAActivation` with `UahfForkHeight=20`); the real-block fixture stops at testnet 547, which reaches neither a retarget nor an EDA event. A committed mainnet header fixture over that range would regression-test the out-of-repo scan the PR body relies on, and would settle this thread as a side effect.

services/blockchain/Difficulty.go
oskarszoon: Confirmed: all 548 committed testnet headers carry the minimum target. The 32-header chunks in 7b93b6dea address the short-run over-fetch reported on the release PR, but still read the full remaining interval for this all-minimum fixture. I am not claiming they solve that case or extrapolating this fixture to all pre-DAA history.

Deferring the non-blocking memoization suggestion to a separate optimization. An interval number alone does not identify the ancestry; caching needs an ancestry-aware identity and verified behavior across competing tips, invalidation and concurrent reads. That is a broader change than this historical-rule fix. The PR description now states this limitation explicitly.

services/blockchain/Difficulty.go
oskarszoon: Agreed that committed real-mainnet vectors would strengthen regression coverage. The external check used independent Python arithmetic over linked, PoW-verified headers; it did not run the Go calculator or SQL ancestry reads. I have corrected the PR description and feedback summary to make that distinction explicit.

Deferring the full 478559–504031 fixture from this minimal release fix. A focused follow-up should run real retarget/EDA/activation windows through `CalcNextWorkRequired` with actual mainnet parameters and SQLite, including the prior interval's header required by the first retarget. The current Go EDA tests remain synthetic.

The retained pre-UAHF optimization is scoped to recorded canonical history. A post-UAHF fixture would not settle the separate pre-UAHF fork-policy difference, and I am not claiming that it does.

services/blockchain/Difficulty.go
oskarszoon: The repeated-walk optimization is now implemented on main in 1a6879cf1. One cache entry stores the restored target by value with the parent hash and height. Same-parent/direct-child calls can reuse it; a verified backward walk can also stop at that ancestor after delayed-candidate gaps. Retarget and timestamp rules retain priority, and the cache cannot cross unrelated ancestry or interval boundaries.

The real-SQLite regression dropped from 85670 requested headers to 21 across 400 sequential parents, and from 1136 to 496 with 70-parent delayed gaps. Per-block best-tip cache resets are included. Fork/concurrent-query, aliasing, delay-ordering, boundary and missing-history tests pass under race detection. The earlier cache deferral is therefore addressed for main; the approved release backport merged separately as 1d8067bf2.

icellan:
Holding off on approving until there's a real-header fixture over mainnet 478559-504031.

Not because I found a bug. I've gone through `calcHistoricalWorkRequired` against SV Node's `GetNextWorkRequired` / `GetNextEDAWorkRequired` / `CalculateNextWorkRequired` several times now — retarget gate and the 2015-back ancestor, the /4 and x4 clamps, `headers[6:]` as the parent's 6th ancestor, near-genesis MTP truncation, the three activation gates, the STN exclusion — and it holds up. The fixes on the earlier threads are good; the batch-plus-cache shape is better than what I originally suggested, and the reorg continuity check on the range read is a nice catch.

The reason to wait is the asymmetry in one specific claim. The comment at the EDA gate says canonical pre-UAHF mainnet never reaches the 12h MTP-6 threshold, and that claim is doing real work: with `skipExpectedDifficulty` now forcing the expected-nBits check for every pre-DAA block, a divergence anywhere between height 32256 and 478558 means a syncing node rejects a canonical block and wedges permanently, with no recovery short of a code change. The evidence for it is an out-of-repo scan. Everything else on my list costs a comment or a params lookup; this one costs a production incident on mainnet IBD.

The riskiest slice is 478559-504031 — the Aug-Nov 2017 window where EDA actually fired repeatedly. Current real-block coverage is testnet 0-547, which sits inside the first retarget interval and is 548/548 at `1d00ffff`, so it exercises neither a real 2016-boundary retarget nor a real EDA evaluation. Everything else is synthetic chains built on the implementation's own assumptions, which is the wrong direction of evidence for a consensus claim.

A committed header fixture over that range turns the scan into a regression test and settles it. If a fixture is impractical, putting the scan's method and result in the PR description makes the claim auditable, which I'd also accept.

Two non-blocking things, happy for either to be follow-ups:

- The historical testnet memoisation is sound, but for a reason the code doesn't state. The early return for `height%interval == 0 || parent.Bits != powLimit` sits above every `remember()`, so a cache entry can only exist for a block that was itself powLimit and non-boundary — which is exactly why the parent's-parent and mid-walk-resume hits are safe. Move that early return, or add a `remember()` above it, and both hit paths go wrong silently. Worth a comment on the early return or an assert in `remember`.
- `const interval = 2016` and the hardcoded /4 and x4 clamp, while `timespan` derives from `params.TargetTimePerBlock`. Correct for every chain that reaches this path today, wrong for a future net with non-10-minute spacing and a non-zero `DaaForkHeight`.

oskarszoon:
@icellan Added `465386d08` in response to [this review](https://github.com/bsv-blockchain/teranode/pull/1732#issuecomment-5697091819).

The committed fixture contains mainnet headers 477792–504031, including ancestry for the first retarget in the requested range. An ordinary-suite test runs the production calculator against real SQLite and compares all 25474 mined child targets from 478558 through 504031: 13 real retargets and all 43 EDA increases. It verifies header linkage, PoW, pinned hashes and the fixture checksum. [The fixture README](https://github.com/bsv-blockchain/teranode/blob/465386d08f480ffa18486b8014e4c5f5a3419d0c/services/blockchain/testdata/README.md) includes checked archive hashes, a reproducible download and the independent pre-UAHF scan's method/results. The PR description now states both coverage scopes explicitly.

The test passes; disabling EDA through a temporary compiler overlay makes it fail at 478577, and shifting the retarget ancestor by one block makes it fail at 479808. Full affected-package race suites, focused block-validation regressions, vet, staticcheck and changed-line lint pass. The fixture reproducer was also run against the public archives.

The cache invariant is documented immediately above the early return, before any `remember` call. Future-network parameterization remains a separate follow-up; this change preserves the historical constants used by the supported networks. No runtime algorithm changes were needed.

Follow-up after 15 minutes: 37 checks passed, one skipped, no pending or failed checks. The [unit job](https://github.com/bsv-blockchain/teranode/actions/runs/35114464146/job/104856230391) completed 14872 tests with 46 skipped; [prunertest](https://github.com/bsv-blockchain/teranode/actions/runs/35114464065/job/104856229572) passed all eight cases. Automated review found no new issues and the benchmark report showed no regressions. Fresh independent consensus, defensive and Go/test reviews found no actionable issues; fixture hashes, PoW and ancestry were independently checked. Review requests remain pending. The existing Sonar complexity refactor remains explicitly deferred.

icellan (approved):
Approving. The fixture settles what I was holding on.

`TestDifficultyHistoricalMainnet` is the right shape: real headers pinned by SHA256 and three block hashes, PoW verified per header so it can't be fabricated data, parent linkage checked, and then the calculator's output asserted against the real `child.Bits` for every height 478558-504031 — 13 retargets and 43 EDA firings. Starting the fixture at 477792 so the first retarget has its 2015-back ancestor is a nice detail. That turns the two branches that could silently fork a node into a regression test against the window where EDA actually fired.

The invariant comment at the early return is also exactly right — on the line whose position carries the load, rather than in a doc block that can drift away from it.

Three non-blocking things, none worth another round:

1. The assertion loop starts at 478558, so the pre-UAHF gate itself — the thing the hold was about — is exercised at one point. I checked the fixture's pre-UAHF portion directly: across parent heights 477808-478557, zero windows reach the 12h MTP-6 threshold, max gap 2.15h, no bits changes. So starting the loop at `first+16` instead would sweep those 750 parents under the gate and pass, at no fixture cost. Worth doing whenever this file is next touched — it converts the gate from asserted to tested over that window. 32256-477807 stays unevidenced, which I'm fine with: ~36MB of headers to cover, and EDA is now proven against 43 real firings with the gate sitting on top of it as an optimisation.

2. The walk is cheap only because of `lastTestnetTarget`. With `blockassembly_difficultyCache=false` — a documented setting — each pre-DAA testnet block costs up to 63 round-trips, which `TestDifficultyHistoricalCacheFailureAndDisable` asserts outright. `LocalClient.GetNextWorkRequired` also builds a fresh `Difficulty` per call, so that path never warms. Test and daemon-mode cost rather than a node cost, but the coupling deserves a comment, or the cache moving off the per-call instance.

3. `const interval = 2016` and the hardcoded /4 and x4 clamp, while `timespan` comes from `params.TargetTimePerBlock`. Fine for every chain that reaches this path today; wrong for a future net with non-10-minute spacing and a non-zero `DaaForkHeight`.

One thing for the PR description rather than the code: the catchup precheck no longer covers pre-DAA heights at all. The doc comment on `validateHeaderChainDifficulty` says so now, but the description still reads as though early rejection is intact across the range.

oskarszoon:
@icellan Clarified the description: the native header precheck does not check expected difficulty below DAA activation. Historical headers reach full-block validation, which performs that check even while building the checkpoint prefix. Early header-level expected-difficulty rejection applies only at/after activation.

Recorded the three non-blocking follow-ups: extend assertions into the existing fixture's earlier pre-UAHF prefix, address disabled-cache/per-call-calculator cost, and parameterize the historical retarget constants for future networks. No code changes to the approved commit.

PR #1733, opened 2026-09-14
fix: backport historical difficulty to v0.15

oskarszoon:
Fix fresh testnet sync stopping at height 148 in `v0.15.9-beta-2`. The regression test reproduces block 149's exact `incorrect difficulty bits: got 1d00ffff, expected 1c7fff80` rejection through legacy processing and the real `ProcessBlock` gRPC handler before this fix.

Backport historical difficulty selection: original 2016-block retarget, testnet minimum-difficulty restoration, and post-UAHF EDA before modern DAA activation. Historical testnet restoration reads at most 32 ancestors per query, stopping at the first ordinary target or retarget boundary and checking linkage across batches. This bounds short-run over-fetch; long all-minimum runs still require the interval scan. STN retains its previous difficulty behavior; reconciling its existing parameters and chain history is outside this fix. Production changes are confined to `Difficulty.go`; the release branch's checkpoint ancestry gate, proof-of-work checks, and RPC interfaces stay intact. Correct historical calculation removes the legacy RPC's reliance on carrying checkpoint proof for this check.

The full legacy regression is tagged `longtest` and runs with race detection through `make longtest`; calculator regressions remain in the ordinary test suite. It drives real blocks 1–547 through block/subtree validation, SQLite chain/UTXO stores, and temporary file blob stores, with netsync's verified checkpoint evidence and no stored checkpoint at height 149. The harness performs block assembly's separate coinbase bookkeeping. The existing catchup test now finishes mock-chain setup before starting subscription readers, removing a race exposed during verification. Tests and cleanup use this branch's older service/store APIs; the median-time window uses the same consensus value of 11 because this branch lacks the shared constant.

The pre-UAHF gate is a historical mainnet optimization, not a guard in SV Node. An independent Python check of linked, PoW-verified mainnet headers found zero ungated historical-target mismatches through child 504031, including 250 retargets and 43 EDA increases. No applicable pre-UAHF MTP window reaches 12 hours (maximum 26789 seconds). This checks the historical arithmetic independently; it is not a replay through the Go calculator. Additional committed mainnet vectors remain deferred.

Refs #1731. Companion implementation: https://github.com/bsv-blockchain/teranode/pull/1732 (`main`).

Validation: full `-race -tags testtxmetacache -count=1` tests passed for `services/blockchain`, `services/blockvalidation`, `services/legacy/netsync`; `go vet`, `staticcheck`, and lint passed. The legacy RPC regression reproduced the exact block-149 rejection before the fix and now reaches height 547. Existing synthetic DAA tests explicitly activate DAA at genesis.

Pruner CI follow-up: the matching regression now waits for actual parent-record deletion, avoiding stale completion notifications from earlier cycles. Three consecutive focused `-race` runs passed with `GOMAXPROCS=2`; package vet, staticcheck and changed-line lint passed. This synchronization fix changes one test file only.

Earlier pruner CI follow-up: The updated parent-deletion test passed in release CI. The first [release prunertest run](https://github.com/bsv-blockchain/teranode/actions/runs/34959117088/job/104348208795) failed in a different, unchanged test; the [retry passed](https://github.com/bsv-blockchain/teranode/actions/runs/34959117088/job/104352048289) with all eight tests. That separate failure remains deferred from this minimal fix.

Batch follow-up validation: a real-SQLite read-budget regression failed with 1803 requested headers before the change and passes with the bounded reads. Seventy-block runs cover multiple batches and retarget-boundary stopping. Affected-package race tests, the tagged full replay, vet, staticcheck and changed-line lint passed on both branches. On the same release checkout, two no-race replay runs measured 27.62/25.11 seconds before and 24.45/24.81 seconds after: a modest end-to-end improvement. The existing long-tests workflow is dispatch-only, so the tagged full replay is no longer part of ordinary PR CI.


services/blockchain/Difficulty.go
```diff
@@ -159,6 +166,84 @@ func (d *Difficulty) CalcNextWorkRequired(ctx context.Context, blockHeader *mode
...
+		}
+		target.Mul(target, big.NewInt(elapsed))
+		target.Div(target, big.NewInt(timespan))
+	} else if params.ReduceMinDifficulty {
+		if blockTime > int64(parent.Timestamp)+2*int64(params.TargetTimePerBlock.Seconds()) {
+			return d.powLimitnBits, nil
+		}
+		// Walk this parent's ancestry, stopping at the last non-special target
+		// or at a retarget boundary. Missing history must not become easy work.
+		for height%interval != 0 && parent.Bits == *d.powLimitnBits {
```
freemans13, line 201: Non-blocking, performance. This is the same loop I commented on in #1732. Each step is its own `GetBlockHeader` query, and `StoreBlock` clears the response cache, so none of them hit it during sync. Replaying real testnet history gives 36.6M lookups below the DAA fork, and the worst single block walks 2,014 steps.

On this branch it also shows up in the tests: 145,872 lookups for blocks 1 to 547, which is why `TestDifficultyHistoricalTestnet` takes 107s under `-race`. The walk never needs more than `height%interval + 1` headers, so one `GetBlockHeaders(ctx, parent.Hash(), uint64(height%interval)+1)` call would do. Keep a length check on the result so missing history still errors.

services/blockchain/Difficulty.go
```diff
@@ -159,6 +166,84 @@ func (d *Difficulty) CalcNextWorkRequired(ctx context.Context, blockHeader *mode
...
+			previous, _, err := d.store.GetBlockHeader(ctx, parent.HashPrevBlock)
+			if err != nil || previous == nil {
+				return nil, errors.NewStorageError("[Difficulty] restoring historical testnet target", err)
+			}
+			parent = previous
+			height--
+		}
+		return &parent.Bits, nil
+	} else {
+		if height < params.UahfForkHeight || height < 6 || parent.Bits == *d.powLimitnBits {
```
freemans13, line 211: Nit, same as on #1732. SV Node's `GetNextEDAWorkRequired` has no UAHF check. I replayed mainnet with the gate removed, and it still matched every block to 504031, because no pre-UAHF block ever met the 12-hour condition. So the gate is safe and saves reads. It would help if a comment said it's an optimisation that only agrees with SV Node because of that history.

freemans13 (approved):
Approving. The backported calculator reproduces every real block's nBits below the DAA fork on both networks, and the new tests pass locally.

What I ran, at `60d68c30`:

- **Real-history replay.** I fed every header from WhatsOnChain's archive through this branch's `CalcNextWorkRequired`, using an in-memory store that serves the three lookups the calculator makes. Mainnet children 1 to 504031 and testnet children 1 to 1,188,697 all matched, with zero mismatches. That covers the retargets and the mainnet EDA window, which the 0 to 547 fixture never reaches.
- **The harness can fail.** Moving mainnet's UAHF height up by 2,000 breaks it at child 478577, the first EDA block. Turning off testnet's `ReduceMinDifficulty` breaks it at child 4033.
- **Fixture provenance.** `testnet_headers_0_547.bin` is byte-identical to the first 548 headers of the WhatsOnChain archive.
- **Tests.** The new historical tests in `services/blockchain`, `TestLegacyHistoricalTestnetSync`, and the four DAA and validation tests you re-pointed at `DaaForkHeight = 0` all pass under `-race -tags testtxmetacache`. All 34 CI checks are green.

`calcHistoricalWorkRequired` matches #1732 at `d9cd644` except that the median-time window is the local constant 11 here, not `settings.MedianTimeSpan`. So my two non-blocking comments on #1732 apply here too. I've left them inline as short pointers. If you change anything on #1732, it's worth carrying across.

One point that's new on this branch: the testnet walk's cost is already visible in CI time. `TestDifficultyHistoricalTestnet` takes 107s under `-race` and 3.4s without. `TestLegacyHistoricalTestnetSync` takes 164s under `-race`. For blocks 1 to 547 the walk makes 145,872 `GetBlockHeader` calls, because every early testnet block is at the pow limit and walks back to genesis. The single range read suggested on #1732 would shrink both the tests and a real testnet sync.

What I didn't check: I didn't run a real testnet or mainnet sync on v0.15. I also didn't time the walk on Postgres. The replay exercises the calculator on its own. Only the 547-block legacy test covers the full block-validation path on this branch.

icellan (approved):

icellan:
Reviewed `calcHistoricalWorkRequired` line-by-line against SV Node's `GetNextWorkRequired` / `GetNextEDAWorkRequired` / `CalculateNextWorkRequired`. The port is faithful: activation gate, the `(parentHeight+1) % 2016` boundary, the 2015-ancestor lookup, the /4 and x4 clamps, the testnet min-difficulty walk and its `height % interval` stop, the 17-header MTP window (`headers[:11]` = MTP(parent), `headers[6:]` = MTP(parent-6)) with its short-chain clamp, the `+= (nPow >> 2)` easing, the PowLimit clamp and the compact round-trip all match. Both new tests pass locally (blockchain historical 4.6s, netsync regression 79s).

Three things worth addressing before merge.

**`services/blockchain/Difficulty.go:209` — EDA gate has no upstream counterpart.** The branch is gated on `height < params.UahfForkHeight`; upstream `GetNextEDAWorkRequired` applies the EDA rule to all pre-DAA blocks and relies only on `nBits == nProofOfWorkLimit` to bail early. Any mainnet block below 478558 with non-powLimit parent bits and `MTP(parent) - MTP(parent-6) >= 12h` gets its target eased 25% upstream but returns `parent.Bits` here, so Teranode would reject a valid block with "incorrect difficulty bits". I think the window is empty in practice (difficulty left powLimit at 32256, longest historical inter-block gap is ~99 min, so six blocks spanning 12h is implausible) — but nothing here verifies that and there's no pre-UAHF mainnet test at all. Either drop the gate to match upstream, or add a comment stating it was verified inert across 32256-478557.

**`services/blockchain/Difficulty.go:196` — per-ancestor DB round-trip in the walk.** The testnet min-difficulty restoration walk calls `store.GetBlockHeader` once per ancestor, bounded only by the 2016 retarget boundary. Syncing testnet from genesis — the case this PR exists to fix — every block below `DaaForkHeight` (1188697) whose parent carries `1d00ffff` walks the entire preceding run of min-difficulty blocks, and runs of hundreds are common in that era. That's O(run length) sequential SQL queries per block, ~1.19M times. SV Node does this in-memory over `CBlockIndex`. One `GetBlockHeaders(parent.Hash(), n)` batch collapses it — the EDA branch three lines down already uses exactly that.

**`services/blockchain/Difficulty.go:71` — STN consensus changes silently.** `StnParams` has `DaaForkHeight: 2200` / `UahfForkHeight: 15`, so STN heights 15-2199 now take the EDA path where they previously took the 144-block DAA (and 0-147 previously returned powLimit). An existing STN chain whose blocks 148-2199 were mined under the old rule fails `ValidateBlockHeaderDifficulty` on resync. The description frames this as a testnet fix and doesn't mention STN; no STN test. Confirm STN chains get reset, or gate the backport.

Minor, not blocking:
- `Difficulty.go:180`, `:200`, `:216` — `errors.NewStorageError(msg, err)` on `err != nil || X == nil`: when the nil-result arm fires `err` is nil, so a genuinely absent ancestor is indistinguishable from a DB failure in logs. `:216` is the worst — a legitimately short ancestry surfaces as a storage error.
- `services/legacy/netsync/historical_testnet_test.go:24` — 79s wall in the default `make test` suite with no build tag, plus a 323KB fixture. `test/longtest/` exists for this.
- `historical_testnet_test.go:128` — `&params.Checkpoints[0]` points into the package-global `chaincfg.TestNetParams` backing array (copying `Params` copies the slice header, not the array). Any write through it corrupts global chaincfg for every other test in the process. Copy into a local.
- `difficulty_historical_test.go:167` — `TestDifficultyHistoricalMissingHistory` cancels the context to prove the walk errors, but `SQL.GetBlockHeader` checks `responseCache` first. It passes today only because `StoreBlock` doesn't seed that entry; if it ever does, the assertion stops testing anything. A store stub returning `ErrNotFound` would pin the behaviour.

services/blockchain/Difficulty.go
oskarszoon: Fixed in ffc9899f7. Historical testnet restoration now uses one bounded `GetBlockHeaders` call, starting at the predecessor because the parent is already available. It stops at the first ordinary target or the retarget boundary, and returns `ErrNotFound` if the required ancestry is absent. SQLite regressions cover a run of minimum-difficulty blocks, the retarget boundary, and missing history. Consumed headers are checked against the requested ancestry so a concurrent reorg cannot substitute a different branch.

services/blockchain/Difficulty.go
oskarszoon: Kept the optimization and documented why it agrees with canonical mainnet history in ffc9899f7. I independently checked linked, PoW-verified WhatsOnChain headers: the ungated historical calculation matches all 504031 children through DAA activation, including 250 retargets and 43 EDA increases. Across the 446081 applicable pre-UAHF parents, no six-block MTP delta reaches 43200 seconds; the maximum is 26789 seconds at parent 74648. This is an optimization for that recorded history, not a UAHF guard present in SV Node. Source: [WhatsOnChain mainnet header archives](https://api.whatsonchain.com/v1/bsv/main/block/headers/resources).

oskarszoon:
Processed the [release feedback](https://github.com/bsv-blockchain/teranode/pull/1733#issuecomment-5676993110) in ffc9899f7; matching fixes are in #1732.

1. **EDA gate:** retained and documented as an optimization tied to recorded mainnet history. Independent Python arithmetic over linked, PoW-verified headers found zero historical target mismatches through child 504031; the largest applicable pre-UAHF six-block MTP delta is 26789 seconds, below 43200. SV Node has no UAHF gate here.
2. **Ancestor reads:** queries now request at most 32 ancestors, with linkage checked across batches. The existing parent is excluded. This limits short-run over-fetch; all-minimum intervals still require a full scan. Missing required ancestry still fails closed.
3. **STN:** preserved its previous difficulty behavior, with real SQLite regressions around parents 0/147/148 and 2199/2200. I could not verify a reset coordinated with this release. Existing Teranode STN parameters differ from SV Node, so parameter/history reconciliation is deferred; this PR makes no reset assumption.
4. **Errors:** nil/short results now carry an explicit `ErrNotFound` cause, separately from errors returned by storage.
5. **Checkpoint aliasing:** the harness points to a local checkpoint value.
6. **Missing-history test:** replaced context cancellation with an actually absent predecessor and a truncated EDA window in SQLite. No store stub or cache-state assumption is needed.
7. **Long-test move:** addressed in 6be5a7d32. The full legacy replay remains in its package behind the `longtest` build tag, preserving private-state access; `make longtest` explicitly runs it with race detection. The existing long-tests workflow is dispatch-only. Calculator regressions remain in ordinary PR checks. Additional committed mainnet vectors remain deferred; the external historical check uses independent Python arithmetic and does not exercise the Go calculator.

Validation: full affected-package race suites and final historical-calculator race regressions passed; `go vet`, `staticcheck` and lint passed. The historical header regression now takes about 13 seconds under `-race` on this machine. Batch ancestry is checked before using any returned target or MTP window. The wider runs exposed an existing catchup-test setup race. Validation now starts after the mock best block is initialized; this small test-only fix is included in both branches. The final isolated release suite passed.

Re-requesting reviews from both human reviewers. Copilot's previous attempts were quota-blocked, so I am not retrying it here.

Automated review follow-up: spot-checked Main/TestNet tests outside the three affected service packages on both branches. The only direct difficulty RPC regression found is `test/blockchain_difficulty_testnet_test.go`, using parents 1602681/1602682, above testnet DAA activation. The remaining matches cover PoW limits, checkpoint predicates, RPC formatting, or unrelated store/subtree behavior. No additional low-height modern-DAA assertion requiring an override was identified.

CI follow-up: carried the one-file pruning-test fix from #1732 into eef7651c5. The old wait accepted a queued completion from an earlier height. The test now polls parent metadata until `ErrTxNotFound`, retaining the 10-second limit and rejecting unrelated storage errors. Three consecutive focused race runs passed with `GOMAXPROCS=2`; vet, staticcheck and lint passed. [Main prunertest CI is green](https://github.com/bsv-blockchain/teranode/actions/runs/34958848685/job/104347331663). The updated parent-deletion test passed in release CI. The first [release prunertest run](https://github.com/bsv-blockchain/teranode/actions/runs/34959117088/job/104348208795) failed in a different, unchanged test; the [retry passed](https://github.com/bsv-blockchain/teranode/actions/runs/34959117088/job/104352048289) with all eight tests. That separate failure remains deferred from this minimal fix.

Reviewed the latest non-blocking cache-coverage note: `DaaForkHeight = 0` alone would not restore coverage in those two `server_test.go` fixtures. They use parent heights 1–3; the release base already returned before cache lookup/population below height 148. The two existing tests pass, and focused coverage confirms no cache lookup/population executes. This is a preexisting coverage gap. Deferring a proper fixture reaching the DAA window, with explicit assertions that the cache is populated and then cleared before another difficulty calculation.

The requested [prunertest retry passed](https://github.com/bsv-blockchain/teranode/actions/runs/34959117088/job/104352048289) with all eight tests. The separate deferred failure remains recorded above.

icellan:
Re-reviewed at `eef7651c5`. The STN gate, the split storage errors, the checkpoint copy and the missing-history test are all properly fixed, and `TestDifficultySTNPreservesExistingRules` is the right shape. I traced the new batch walk line-by-line against SV Node's `while (pindex->pprev && pindex->nHeight % nInterval != 0 && pindex->nBits == nProofOfWorkLimit)` — the early return, the per-header `height--` check, the `count` bound, the genesis case and the EDA continuity guard all check out. No correctness bug.

But the batching I asked for is a performance regression, and I was wrong to suggest it in that form.

**`services/blockchain/Difficulty.go:209`**

```go
headers, _, err := d.store.GetBlockHeaders(ctx, parent.HashPrevBlock, uint64(height%interval))
```

The batch is sized by the distance to the retarget boundary, not by the length of the min-difficulty run. Below 2016 that's the entire chain; at real testnet heights it averages ~1008 headers fetched per call regardless of whether the run is 1 block or 500.

The caching makes it worse. The old point-query walk called `GetBlockHeader` per ancestor, which keys the response cache by block hash — so consecutive blocks walking overlapping ancestry reused each other's entries. `GetBlockHeaders` keys its cache by `(startHash, count)`, and `count` is different on every call, so the new path never gets a hit and re-reads the full window for every block.

Measured with `TestLegacyHistoricalTestnetSync` (547 blocks, same machine, `-count=1`, no `-race`):

| commit | run 1 | run 2 |
|---|---|---|
| 60d68c30a (point-query walk) | 54.2s | 34.1s |
| eef7651c5 (batch read) | 103.5s | 75.5s |

~2.2x slower, and the gap grows with height because `height % interval` does. This is the historical testnet sync path the PR exists to fix, so it's worth getting right.

Two options. Read in small chunks — 32 or 64 — and continue only while the whole chunk is min-difficulty, still capped at the boundary; that keeps the query count bounded without the over-fetch. Or just revert to the walk: it measurably wins here, and the 2015-query worst case I was worried about needs 2015 consecutive min-difficulty blocks inside a single retarget interval.

Separately, still open from last round: `services/legacy/netsync/historical_testnet_test.go` has no build tag and is in the default `make test` suite, where it now runs 149s under `-race`.

icellan (approved):

oskarszoon:
@icellan addressed the [short-run over-fetch and slow default test](https://github.com/bsv-blockchain/teranode/pull/1733#issuecomment-5680670561) in 6be5a7d32; the matching main change is 7b93b6dea.

Restoration now requests at most 32 headers and continues only while the target remains minimum, capped at the retarget boundary. Every consumed header is linked to the preceding hash, including across batches. The real-SQLite regression failed with 1803 requested headers before the fix; the bounded version passes. Seventy-block cases exercise multiple batches and boundary stopping.

On this machine, two no-race release replay runs changed from 27.62/25.11 seconds to 24.45/24.81 seconds. That is a modest end-to-end improvement, not evidence of a large speedup. `StoreBlock` resets the response cache, so its point-lookup entries do not survive ordinary consecutive block writes. As your main-PR follow-up notes, chunking still scans the entire interval for an all-minimum run; ancestry-safe memoization remains a separate, non-blocking optimization.

The full legacy replay now has the `longtest` build tag and remains in its package. `make longtest` explicitly runs it with `-race` before the existing long tests. The existing workflow is dispatch-only, so this deliberately removes the heavy replay from ordinary PR CI. Calculator and native-header regressions stay in their normal suites.

Both branches passed affected-package race tests, the tagged full replay, vet, staticcheck and changed-line lint. Fresh independent reviews found no blocker. The PR descriptions and earlier feedback summary now reflect the final batching and test placement. icellan has approved this release commit; the remaining review requests stay open.

Follow-up to the latest automated review: checked the selected `go-chaincfg v1.5.8` source directly. Mainnet uses `UahfForkHeight=478558`, `DaaForkHeight=504031`; testnet uses `1155875` and `1188697`. The same values are present in main's `v1.6.2`. Real committed headers still stop before the first retarget; that path has synthetic coverage, and additional real-mainnet calculator vectors remain deferred. Current restoration uses 32-header chunks; the review's one-batch note describes the earlier implementation.

ordishs (approved):
Compared `calcHistoricalWorkRequired` against SV Node `pow.cpp`. The port is faithful: parent-height activation, ancestor depth 2015, the deliberate full `2016 * 600` timespan, the `[T/4, 4T]` clamp, the PoW-limit ceiling, the testnet restoration walk, and the `headers[0:11]` / `headers[6:17]` MTP windows all match. Fork safety holds: both store reads fall back to a recursive `parent_id` walk, and the new code verifies hash linkage on every returned header. Only mainnet and testnet reach the new path.

One finding.

### P1 — `services/blockchain/Difficulty.go:207-229`

The testnet minimum-difficulty restoration walk costs O(L²) header reads across a run of L consecutive minimum-difficulty blocks. Each block repeats the whole walk from its own parent. The store response cache never hits, because every block starts the walk from a different hash.

Measured against the real `sqlitememory` store, one `CalcNextWorkRequired` call per block in the run:

```
run=100  elapsed=33.5ms   headersRequested=6656
run=400  elapsed=301.9ms  headersRequested=86528
run=800  elapsed=1.1459s  headersRequested=332800
```

The growth is quadratic. A run of 2015 blocks, which the retarget interval permits, extrapolates to about 2.1 million header reads for that one interval. A disk-backed store is much slower than `sqlitememory`.

This path is live in the exact scenario the change fixes. `skipExpectedDifficulty` needs checkpoint proof, and the legacy `ProcessBlock` handoff does not carry it. So a fresh testnet sync runs this walk for every block up to `DaaForkHeight` 1188697. BSV testnet has long low-hashrate stretches where almost every block carries the minimum target.

The new tests do not expose this. `TestLegacyHistoricalTestnetSync` stops at height 547, where the runs are short, and `TestDifficultyHistoricalTestnetRestoresTarget` uses a maximum run of 70.

**Recommendation**: memoize the walk result. The answer is identical for every block in one contiguous minimum-difficulty run inside one retarget interval. Cache the resolved `NBit` against the hash that starts the run, or add a store query that returns the newest ancestor whose bits differ from the PoW limit. A single entry on `Difficulty` is enough, because sync processes the run in order.

### Verification run locally

- `go vet -tags "testtxmetacache longtest" ./services/legacy/netsync/ ./services/blockvalidation/ ./test/e2e/daemon/ready/` — clean
- `go test -race -tags testtxmetacache -run TestDifficulty ./services/blockchain/` — ok, 38.1s
- `go test -race -tags testtxmetacache -run TestServer_catchup ./services/blockvalidation/` — ok, 2.2s
- `go test -race -tags "testtxmetacache longtest" -run '^TestLegacyHistoricalTestnetSync$' ./services/legacy/netsync` — ok, 58.6s

Nits sent separately.

oskarszoon:
@ordishs confirmed the [repeated-walk finding](https://github.com/bsv-blockchain/teranode/pull/1733#pullrequestreview-5220574122) with a real-SQLite read-budget regression. The approved historical-difficulty backport has merged as 1d8067bf2. The performance improvement is being delivered in main PR #1732, commit 1a6879cf1, preserving this release patch's approved scope.

One entry memoizes timestamp-independent restoration by parent hash and height. Same-parent/direct-child calls reuse it, and a linked walk can stop at the cached ancestor across delayed-candidate gaps. Retarget and delayed-block rules retain priority; failures never populate the cache. The entry survives best-tip cache resets because its ancestry hash remains authoritative.

Across 400 sequential parents, requested headers fell from 85670 to 21; with 70-parent delayed gaps they fell from 1136 to 496. These tests retain one calculator as production does and reset the best-tip cache after each block. Fork isolation, concurrent queries, returned-value ownership, delays, retarget boundaries, missing history and cache-disable behavior pass under race detection. The full legacy replay remains correctness coverage: its LocalClient creates a calculator per call.

Memoization remains deferred for the release branch. Additional committed real-mainnet vectors remain a separate coverage follow-up.

PR #1734, opened 2026-09-14
p2p: harden gossip hash parsing, SSRF ranges, registry batcher and coordinator shutdown

gokutheengineer:
**Addresses bitcoin-sv/teranode#4762** (findings 2-7). Finding 1 is a bug in `go-p2p-message-bus` and is fixed by the companion PR bsv-blockchain/go-p2p-message-bus#119; the issue closes fully once that ships and `go.mod` is bumped.

### Problem
Seven low-severity p2p findings from one audit. Re-verified against current `main`:

1. `P2PClient.Close()` closes subscriber channels before stopping the readers that send on them (send-on-closed-channel panic at SIGTERM under gossip load). Still present in the module through v0.1.24 and `main`; not fixable from Teranode, see companion PR.
2. `isUnsafeIP` enumerated four `net.IP` predicates, so literal DataHub URLs in RFC 6598 shared address space (100.64.0.0/10, carrier NAT and Tailscale overlays), 192.0.0.0/24, 198.18.0.0/15, 240.0.0.0/4, multicast/broadcast and IPv6 transition ranges passed SSRF validation. Still valid.
3. `parseHash` and `sanitizeAdvertisedTip` fed gossip hashes straight to `chainhash.NewHashFromStr`, which zero-pads short input: `""` decodes to the all-zero hash with a nil error, so an omitted `best_block_hash` registered a peer with a "known" tip. Still valid (the newer `checkGossipHex` bound deliberately admits empty and short values).
4. Log forgery via raw `%s` of unvalidated hashes: largely already closed on `main` by `validateFields` (printable/hex bounds before any log). What remained: `parseHash` logged the raw value with `%s`, and inner-JSON decode failures were never scored, so a valid envelope with garbage inside was a free log flood (the message bus only counts a malformed outer envelope).
5. `peerRegistryBatcher.flushOnce` returned on flush-budget expiry with the rest of the swapped-out pending map dropped. Still valid.
6. `SyncCoordinator.HandlePeerDisconnected` spawned an untracked, uncancellable sleep-then-`TriggerSync` goroutine that escaped `Stop`'s drain. Still valid.
7. A `forget()` that raced a flush's in-flight `RegisterPeer` could let the registry apply the register after `removePeer`'s `RemovePeer`, resurrecting a banned peer as a live entry nothing reconciles. Still valid.

### Fix
- **SSRF ranges**: `isUnsafeIP` is now default-deny. Anything not global unicast is unsafe, and the global-unicast space is filtered by an explicit RFC 6890 special-purpose table (shared address space, this-network, IETF assignments, TEST-NETs, benchmarking, 6to4 relay anycast, reserved/broadcast, IPv4-compatible and IPv4-mapped forms, NAT64, Teredo, 6to4, IPv6 documentation, discard-only). Each range reports its own reason. `validateDataHubURL` and the settings doc describe the policy, including that an announcing peer is scored.
- **Hash length**: `parseHash` and `sanitizeAdvertisedTip` require exactly 64 hex chars before decoding. Block and subtree handlers score a missing/short hash as a protocol violation (their hash is mandatory and `validateFields` already proved it is bounded hex, so this is the only remaining way to fail). `node_status` stays telemetry: the tip is blanked and not scored, and, improving on the previous early `return`, the peer's storage mode and registry presence are still recorded.
- **Log sinks / inner JSON**: `parseHash` logs the value with `%q` only after its length is known. All four gossip handlers run the banned-peer skip before `json.Unmarshal` (so a banned peer's garbage cannot trigger `AddBanScore` RPCs) and charge a protocol violation for *structurally invalid* JSON only (`!json.Valid`). A field type mismatch is dropped unscored, so another implementation encoding an ignored field (e.g. `Coinbase`, `fee_policy`) differently is not banned. This narrows the issue's "score all decode failures" proposal deliberately.
- **Batcher timeout**: remaining entries are requeued via `requeue` (tombstones and the pending cap still apply, newer observations win on merge); `requeue` now reports whether anything was kept so the log counts are honest. The constant's doc no longer claims drops are intended.
- **Batcher resurrection**: after a peer's RPCs return, the tombstone re-check runs under a single lock hold, and if the peer was forgotten while its `RegisterPeer` was in flight a compensating `RemovePeer` is issued. It is skipped when a fresh observation for the peer arrived during the cycle (tracked in a new flush-scoped `reenqueuedDuringFlush` set), because that peer re-registers next cycle and the removal would wipe writes from paths that bypass the batcher (catchup penalties, coordinator peer info). The compensation always gets at least a 5s budget (detached from the flush ctx when that has less left), so a flush about to expire cannot leave the ghost behind. Chosen over the issue's "route removals through the batcher" alternative as the smaller, ordering-provable change.
- **Coordinator goroutine**: new `goTracked` helper runs the re-select on a wg-tracked goroutine and refuses once `Stop` has begun; the check and `wg.Add` are atomic under a new `lifecycleMu` that `Stop` takes before waiting, so a late libp2p callback cannot race `wg.Add` against `wg.Wait`. The bare `time.Sleep` is a timer select on `stopCh`/`ctx.Done`.

### Residual
- Peers announcing a DataHub URL in the newly blocked ranges are scored, exactly as RFC 1918 announcers already were. A node behind carrier NAT advertising its 100.64/10 address will be banned by peers running `p2p_allow_private_ips=false`. Kept for consistency with the existing private-range policy and documented; softening to drop-without-score for RFC 1918/6598 is a separate policy decision.
- `Start` still calls `wg.Add` directly (the service wires `Start` before `Stop`); the `lifecycleMu` guarantee is documented as scoped to callback-spawned goroutines.
- Finding 1 needs the module release plus a `go.mod` bump in a follow-up.

### Tests
Added:
- `TestIsUnsafeIP` rows for every new range and boundary (100.63/100.128, 192.0.1, 198.17, `::ffff:` and `::a.b.c.d` forms, Teredo vs neighbouring 2001: space); `TestValidateDataHubURL` shared-space/benchmarking/multicast URLs.
- `TestServerHelpers_ParseHash_RequiresFullLength`, empty/short cases in `TestServerHelpers_SanitizeAdvertisedTip_ClampsAndOverflow` (both pin the `NewHashFromStr("")` premise).
- `TestHandleBlockTopic_ShortHashRejectedAndScored`, `TestHandleSubtreeTopic_EmptyHashRejectedAndScored`, `TestHandleNodeStatusTopic_EmptyBestBlockHashDoesNotRegisterTip`, `TestHandleTopics_MalformedInnerJSONScored` (per topic: garbage scored, own message not self-scored, banned peer not re-scored, field type mismatch not scored).
- `TestPeerRegistryBatcher_TimeoutRequeuesRemainder` (expired before start, expired mid-cycle, newer observation wins), `TestPeerRegistryBatcher_RemovalDuringInFlightRegisterIsCompensated`, `TestPeerRegistryBatcher_ReenqueueDuringInFlightRegisterSkipsCompensation`.
- `TestSyncCoordinator_HandlePeerDisconnected_ReselectIsTrackedAndCancellable` (drain returns promptly and no sync decision runs after `Stop`), `TestSyncCoordinator_GoTracked_RefusedAfterStop`.

Run (affected package only):
```
```


ctnguyen (commented):
Review at commit [960e10b](https://github.com/gokutheengineer/teranode/commit/960e10b5afb2cc5889f92d893c775cf7ad3656c9)

Nice batch. Six findings landed in one coherent diff without any of them bleeding into unrelated refactoring, and the design calls are the right ones: flipping `isUnsafeIP` to default-deny with an explicit reason per range is much more durable than chasing predicates one at a time, `goTracked` plus `lifecycleMu` is the minimal correct shape for the coordinator goroutine (I went looking for an `Add`-after-`Wait` hole and could not find one), and moving the banned-peer skip ahead of `json.Unmarshal` in all four handlers is a genuine improvement rather than a shuffle. The test additions are unusually good — `startBlockedFlushPair` actually drives the batcher interleaving instead of asserting around it, and the requeue tests cover expiry-before-start, expiry-mid-cycle and merge precedence separately. Two of the narrowing decisions (scoring only `!json.Valid`, skipping compensation on re-enqueue) are where I ended up spending most of my time; one of them has a gap. Everything below rests on static analysis plus a standalone `encoding/json` probe — I could not run the package test suite here.

## Blocking issues

### ChiR1 — the compensating `RemovePeer` is skipped whenever any enqueue precedes the `forget` in the same cycle

**Problem:** `reenqueuedDuringFlush` is written by `enqueue` (`services/p2p/peer_registry_batcher.go:276-278`) but never cleared by `forget` (`:346-370`), so the compensation guard at `:543` cannot tell "forgotten, then re-observed" from "observed, then forgotten". In the second ordering the guard fires anyway and the ghost entry stays. Concretely: a flush picks up peer P and issues `RegisterPeer(P)`; while that RPC is in flight a topic worker still processing an earlier message from P calls `enqueue(P)`, which sets `reenqueuedDuringFlush[P]`; then the ban event arrives and `removePeer` runs `forget(P)` (`services/p2p/server_helpers.go:903-918`) followed by its own `RemovePeer(P)`. At `:519-520` the cycle reads `removedMidFlight=true` **and** `reenqueued=true`, so `compensateRemove` is not called — but `forget` already did `delete(b.pending, peerID)` at `:351`, so there is no pending registration left to "re-register next cycle".

**Why it matters:** this is the exact resurrection the compensation exists to prevent (issue [bitcoin-sv/teranode#4762](https://github.com/bitcoin-sv/teranode/issues/4762) finding 7), and the window that reaches it is narrow but unconditional: a topic-handler goroutine that passed the banned-peer skip *before* `removePeer` ran `forget` is still in flight, and its `enqueue(P)` lands after the cycle issued `RegisterPeer(P)` and before the tombstone — the gossip worker pool admits several such messages per peer concurrently. `forget` is also reached from the operator `DisconnectPeer` RPC (`services/p2p/Server.go:2798-2804` into `disconnectBannedPeerByID` into `removePeer`), where the peer is never banned and so nothing filters its gossip for the duration of the cycle at all. Nothing reconciles the result — `reconcileConnectionStates` only clears `IsConnected` (`services/p2p/server_helpers.go:1596-1610`), it never removes the entry — which is what the comment at `:513-517` already says. The comment at `:539-542` states an invariant ("it re-registers next cycle anyway") that does not hold in this ordering, and `TestPeerRegistryBatcher_ReenqueueDuringInFlightRegisterSkipsCompensation` (`peer_registry_batcher_test.go:1217-1240`) only exercises forget-then-enqueue, so the gap is invisible to the suite.

**Fix:** clear the marker in `forget`, mirroring the way `enqueue` clears the tombstone at `:275` — one line, and `delete` on a nil map is a no-op so no guard is needed:

  services/p2p/peer_registry_batcher.go, in `forget()` under `b.mu`:

  ```go
    delete(b.lastAsserted, peerID)
    delete(b.pending, peerID)
    // A forget supersedes an earlier re-enqueue: there is no longer a pending
    // registration to justify skipping the compensating RemovePeer.
    delete(b.reenqueuedDuringFlush, peerID)
  ```

  Worth adding the mirrored test case (enqueue, then forget, then finish the blocked flush; assert the peer is gone from the registry).

## Non-blocking issues

### ChiR2 — the unscored half of the new decode rule is the half that echoes attacker bytes into an ERROR line

**Problem:** all four handlers log the raw decode error unconditionally and only then decide whether to score (`services/p2p/server_helpers.go:64-67`, `:260-263`, `:675-678`; `services/p2p/Server.go:1539-1542`). When `json.Valid(m)` is false the error is a short `SyntaxError` — scored, and cheap to log. When it is true but a field type-mismatches, `encoding/json` returns an `UnmarshalTypeError` whose `Value` is `"number " + <raw literal>`, interpolated verbatim by `Error()` — logged in full, and **not** scored. I confirmed this with a standalone probe against a struct carrying the same `Height uint32` shape as `BlockMessage` (`services/p2p/message_types.go:104`): a 9011-byte payload of `{"Height":999…}` gives `jsonValid=true`, `err != nil`, and a 9086-byte error string. `maxBlockMessageSize` is `maxGossipMessageSize` = 10 KB (`services/p2p/Server.go:107`, `:117`), subtree and rejected-tx are 8 KB (`:118`, `:131`), and `ZLoggerWrapper.Errorf` applies no truncation and writes the line twice when the JSON logger is enabled (`ulogger/zerologger.go:291-298`).

**Why it matters:** the PR body's stated goal for this change is that "a valid envelope with garbage inside was a free log flood", and the rule as written exempts exactly the shape that carries ~10 KB of peer-controlled bytes per message. To be precise about scope: the echoed bytes are digits, so this is log **volume**, not the log-line forgery that issue [bitcoin-sv/teranode#4762](https://github.com/bitcoin-sv/teranode/issues/4762) identified as the novel part of finding 4 — `parseHash` closed that properly with `%q` after a length check. It is also not a regression: the pre-PR line logged the same `err`. The point is narrower — the new rule does not cover the sink it was written for, and the "do not ban on a wire-format divergence" reasoning does not require logging the offending literal.

**Fix:** keep the scoring split as designed and bound the log instead, so the two halves line up:

  ```go
    if !isSelf && !json.Valid(m) {
        s.logger.Errorf("[handleBlockTopic] malformed JSON from peer %s: %v", fromID, err)
        s.addProtocolViolation(fromID)
        return
    }
    s.logger.Errorf("[handleBlockTopic] undecodable block message from peer %s", fromID)
    return
  ```

  A shared helper would keep the four handlers from drifting.

### ChiR3 — after widening the static filter, "stopped at connection time instead" only holds for three of the blocked ranges

**Problem:** the `validateDataHubURL` comment gained the bullet "Any other non-global-unicast or IANA special-purpose address" (`services/p2p/server_helpers.go:580`) while keeping the sentence below it that a hostname resolving to an internal address "is stopped at connection time instead" (`:582-586`). `util.DefaultSSRFDialPolicy` blocks loopback, link-local and unspecified only (`util/http.go:172-183`). The settings doc has the same seam: the rewritten paragraph enumerates the widened block list (`docs/references/settings/services/p2p_settings.md:193-201`) and the untouched paragraph immediately below concludes "Accepting such a URL therefore does not make it reachable" (`:203-206`), with a correction that covers private ranges only (`:207-209`).

**Why it matters:** the residual exposure is now larger than the surrounding prose implies — a peer announcing `http://internal.example.com/` that resolves into 100.64.0.0/10, 198.18.0.0/15 or 240.0.0.0/4 passes the static filter *and* is dialled, because the dial policy was not widened alongside it. That is a defensible choice (the dial policy is deliberately narrow, per `util/http.go:167-171`), but a reader of either comment would conclude the opposite, which is the kind of thing that gets relied on in the next review.

**Fix:** name the overlap explicitly in both places — something like "stopped at connection time for loopback, link-local and unspecified addresses only; the remaining special-purpose ranges are blocked by this static filter for literal IPs but are reachable via a hostname that resolves into them" — and extend the doc's `:207` correction sentence to cover the newly added ranges, not just private ones.

### ChiR4 — the cut-short log counts requeued peers against the whole cycle, so it reads as a loss report

**Problem:** `services/p2p/peer_registry_batcher.go:554` logs `requeued` against `len(pending)`, but `requeued` counts only the entries the timeout path put back (`:431-435`) while `len(pending)` is every peer in the cycle, including all the ones already flushed successfully. A cycle that times out on its last peer logs "requeued updates for 1 of 100 peers", which reads as 99 lost when in fact 99 were delivered. The previous line counted down a genuine `unflushed` remainder.

**Why it matters:** this fires only during a registry stall, i.e. precisely when an operator is reading the log to decide whether data was dropped, and the PR body offers these counts as the thing that makes the requeue path honest. Nothing is actually lost except entries `requeue` refused (tombstoned, or over the pending cap), which this line cannot distinguish either.

**Fix:** count the peers that actually entered the timeout branch and report against that, keeping the refused ones visible:

  ```go
    remaining := 0
    // ... in the timeout branch: remaining++; if b.requeue(peerID, u) { requeued++ }
    b.logger.Warnf("[peerRegistryBatcher] flush cut short (%v): %d of %d peers unflushed, %d requeued",
        ctx.Err(), remaining, len(pending), requeued)
  ```

## Recap

| ChiR1 | Compensating removal skipped after reenqueue | YES | 60% |
| ChiR2 | Unscored decode error echoes payload | | 45% |
| ChiR3 | Connection-time reachability claim overstated | | 25% |
| ChiR4 | Misleading requeue counts in log | | 15% |

services/p2p/server_helpers.go
```diff
@@ -370,6 +380,26 @@ func (s *Server) addProtocolViolation(peerID string) {
...
+// that half never writes err to the log.
+func (s *Server) handleGossipDecodeError(handlerName string, m []byte, fromID string, isSelf bool, err error) {
+	if !json.Valid(m) {
+		s.logger.Errorf("[%s] malformed JSON from peer %s: %v", handlerName, fromID, err)
+		if !isSelf {
+			s.addProtocolViolation(fromID)
+		}
+		return
+	}
+	s.logger.Errorf("[%s] undecodable message from peer %s: field type mismatch, dropped unscored", handlerName, fromID)
```
freemans13, line 400: The rule here is "score only when `json.Valid(m)` is false", and the branch it leaves unscored is reachable by payloads that are not a wire-format divergence at all.

I drove the four real handlers with `[]`, `123`, `"x"` and a 4 KB array of integers. Every one is structurally valid JSON, every one fails `json.Unmarshal` into `BlockMessage`, `SubtreeMessage`, `RejectedTxMessage` and `NodeStatusMessage` alike, and every one came back with a ban score of zero plus exactly this line:

```
[handleBlockTopic] undecodable message from peer 12D3Koo...: field type mismatch, dropped unscored
[handleSubtreeTopic] undecodable message from peer 12D3Koo...: field type mismatch, dropped unscored
[handleRejectedTxTopic] undecodable message from peer 12D3Koo...: field type mismatch, dropped unscored
[handleNodeStatusTopic] undecodable message from peer 12D3Koo...: field type mismatch, dropped unscored
```

So the free log flood the PR set out to close is still free, just cheaper per line. That is not a regression against `main`, where the whole decode error was unscored and logged in full, so I am not calling it blocking. It is that the new rule does not cover the sink it was written for.

The tolerance you built this for is right: a peer encoding `Coinbase` or `fee_policy` differently should not be banned. It just does not need to extend to a document whose top level is not an object, which no implementation produces by accident. One extra condition keeps the field-type case unscored and charges the rest:

```go
func (s *Server) handleGossipDecodeError(handlerName string, m []byte, fromID string, isSelf bool, err error) {
	trimmed := bytes.TrimLeft(m, " \t\r\n")
	if !json.Valid(m) || len(trimmed) == 0 || trimmed[0] != '{' {
		s.logger.Errorf("[%s] malformed JSON from peer %s: %v", handlerName, fromID, err)
		if !isSelf {
			s.addProtocolViolation(fromID)
		}
		return
	}
	s.logger.Errorf("[%s] undecodable message from peer %s: field type mismatch, dropped unscored", handlerName, fromID)
}
```

The `%v` on the scored branch stays safe, because a document that is not an object never reaches it with an `UnmarshalTypeError`: the error there is a `SyntaxError` or the short `cannot unmarshal array into Go value of type p2p.BlockMessage`, with no peer literal in it.

Separately, the message text is wrong for these payloads even if you keep them unscored. A bare array is a whole-document type error, not a field type mismatch, so the line misdescribes what the operator is looking at. Worth adding `[]` and a bare scalar as rows to `TestHandleTopics_MalformedInnerJSONScored` either way, so whichever behaviour you choose is pinned.

services/p2p/server_helpers.go
```diff
@@ -473,7 +581,9 @@ func isLocalhostHostname(hostname string) bool {
 // - Invalid schemes (only http/https allowed)
 // - Loopback addresses (127.x.x.x, ::1)
 // - Private network addresses (10.x.x.x, 172.16-31.x.x, 192.168.x.x, fc00::/7)
+// - Shared address space / carrier NAT (100.64.0.0/10)
 // - Link-local addresses (169.254.x.x, fe80::/10)
+// - Any other non-global-unicast or IANA special-purpose address (see isUnsafeIP)
```
freemans13, line 586: This sentence has drifted since the merge, in both directions.

`de9e5292a` came in with your upstream merge and widened `util.DefaultSSRFDialPolicy`, which now refuses RFC1918, `fc00::/7` and `100.64.0.0/10` at dial time as well as loopback, link-local and unspecified. That kills the shared-address-space half of ctnguyen's ChiR3: a hostname resolving into `100.64/10` is no longer dialled with the default `p2p_allow_private_ips=false`.

What is left is the gap this PR opened by widening the static filter past the dial policy. A hostname resolving into `192.0.0.0/24`, `198.18.0.0/15`, `240.0.0.0/4`, the TEST-NETs, multicast, or any of the IPv6 transition ranges passes this filter and is dialled, while the same address written as a literal is rejected. That is a defensible split, since the dial policy is deliberately narrow, but "is stopped at connection time instead" now reads as a guarantee that covers the whole bullet list above it, and the next reviewer will lean on it.

One clause fixes it, something like "stopped at connection time for loopback, link-local, unspecified and private addresses; the remaining special-purpose ranges are blocked here for literal IPs only". The settings doc has the same seam at `docs/references/settings/services/p2p_settings.md:209`, where "Accepting such a URL therefore does not make it reachable" sits under a bullet naming only three of the ranges the paragraph above it now claims to block.

services/p2p/peer_registry_batcher.go
```diff
@@ -481,38 +511,76 @@ func (b *peerRegistryBatcher) flushOnce(ctx context.Context) {
...
 }
 
+// compensateRemove re-issues RemovePeer for a peer whose forget() raced this
+// cycle's RegisterPeer. Ordering is what makes it correct: the RegisterPeer
+// has returned, so the registry has applied it, and this RemovePeer is
+// therefore ordered after it regardless of how removePeer's own RemovePeer
+// interleaved. The RPC gets at least registryCompensateTimeout of budget: the
+// flush ctx is used only while that much remains on it, otherwise a detached
+// budget, so a flush that is about to expire cannot leave the ghost behind.
+func (b *peerRegistryBatcher) compensateRemove(ctx context.Context, peerID string) error {
```
freemans13, line 575: This deliberately escapes the flush budget, and during shutdown the flush budget is the service stop budget. `Server.Stop` passes its `ctx` straight through to `registryBatcher.stop`, which calls `flushOnce(ctx)` synchronously, so every compensation taken with under 5 s left adds up to 5 s to shutdown, one peer after another, with no aggregate cap.

In practice it is narrow: `s.P2PClient.Close()` runs at `services/p2p/Server.go:2616`, well before the batcher stop at `:2648`, so most `forget` calls land before the final cycle starts and those peers are skipped by the tombstone check rather than reaching `sendRegister`. Only a `forget` that lands during the cycle can trigger a compensation. I am flagging it because the escape is unbounded by design rather than because I can construct the pileup: a single wedged registry plus a couple of late disconnects is enough to blow a 10 s stop budget.

A cheap bound would be to cap the total compensating time per cycle, or to skip the detached budget entirely when the flush ctx came from `stop` and just log the ghost, since the next process start re-reconciles anyway.

services/p2p/sync_coordinator.go
```diff
@@ -702,14 +718,49 @@ func (sc *SyncCoordinator) HandlePeerDisconnected(peerID peer.ID) {
...
+		// Trigger selection of a new sync peer after a brief delay so other
+		// peers' status updates can land first. Tracked and cancellable like
+		// every other coordinator goroutine: a bare sleep would outlive Stop's
+		// drain and re-run a full select-and-activate on a stopped coordinator.
+		if !sc.goTracked(func() {
+			timer := time.NewTimer(syncPeerReselectDelay)
+			defer timer.Stop()
+
+			select {
+			case <-timer.C:
```
freemans13, line 730: Three ready arms and no priority, so when the timer and `stopCh` become ready in the same instant Go picks uniformly and `TriggerSync` still runs. The godoc two lines up says the point is that a bare sleep would "re-run a full select-and-activate on a stopped coordinator", and in that tie it does exactly that.

The blast radius is small, because `goTracked` has the goroutine on the WaitGroup and `Stop` waits for it, so the decision runs on a cancelled `sc.ctx` and its RPCs fail fast. It still takes `decisionMu` on the way through. A non-blocking re-check after the timer arm makes the guarantee unconditional:

```go
case <-timer.C:
	select {
	case <-sc.stopCh:
		return
	default:
	}
```

freemans13 (approved):
Approving. The full upstream check suite is green at `d1fc4d2`, and I could not find a correctness hole in the batcher, the SSRF table or the coordinator. One gap in the decode-scoring rule is worth closing, plus three smaller notes, none of them blocking.

`d1fc4d2` is both the PR head and the branch tip on your fork, so the Files-changed view and my read are the same code.

### What I ran

```
SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=1 ./services/p2p/...
946 passed in 3 packages
```

Upstream CI is green on the whole teranode suite at this head, not a fork-gated subset: unit, seven sequential shards, three smoke shards, legacy-sync, prunertest, chainintegrity, golangci-lint, fourteen benchmark jobs and the Sonar gate.

Then I mutation-tested the fixes, because a hardening change is only worth what its tests can tell apart. Seven mutants, seven caught:

| Fix broken | Test that failed |
|---|---|
| `forget` no longer clears the re-enqueue marker | `TestPeerRegistryBatcher_EnqueueThenForgetDuringInFlightRegisterStillCompensates` |
| `parseHash` length guard removed | `TestServerHelpers_ParseHash_RequiresFullLength` |
| block-topic scoring on an unusable advertised tip removed | `TestHandleBlockTopic_ShortHashRejectedAndScored` |
| `198.18.0.0/15` dropped from the SSRF table | `TestValidateDataHubURL/benchmarking_range` |
| `goTracked` no longer refuses once `Stop` has begun | `TestSyncCoordinator_GoTracked_RefusedAfterStop` |
| decode-error scoring removed | `TestHandleTopics_MalformedInnerJSONScored/node_status` |
| flush timeout drops instead of requeuing | `TestPeerRegistryBatcher_TimeoutRequeuesRemainder/budget_spent_mid-cycle` |

Every mutant was restored and the tree is clean. That is a better result than I usually get: the common failure is a test that asserts the call rather than the end state, and these assert the end state.

### Where the previous round stands

ChiR1 is properly closed, and the mutation above shows the suite would catch a regression of it. ChiR3 has partly dissolved under the upstream merge you took: `de9e5292a` widened `util.DefaultSSRFDialPolicy` to refuse RFC1918, `fc00::/7` and `100.64.0.0/10` at dial time, so the shared-address-space half of that finding no longer holds. What is left of it is narrower and still real, and I have put it inline. ChiR2 and ChiR4 are unchanged, and my main finding is a sharper version of ChiR2.

### The one I would actually fix

The unscored half of the decode rule admits payloads that no implementation could have produced by accident, and each one still buys an ERROR line for free. The rule is "score only when `json.Valid(m)` is false". A bare `[]`, a bare `123`, a bare `"x"` or a 4 KB array of integers are all structurally valid JSON, all fail to unmarshal into any of the four message structs, and all land in the unscored branch. I drove the real handlers with those payloads and every one came back with a ban score of zero and one ERROR line:

```
[handleBlockTopic] undecodable message from peer 12D3Koo...: field type mismatch, dropped unscored
```

Same on subtree, rejected_tx and node_status. The stated goal for this change was that "a valid envelope with garbage inside was a free log flood", and it still is, at roughly 130 bytes per line instead of 10 KB. The rate is unbounded and nothing charges for it.

The tolerance you built this for is a peer encoding an ignored field differently, and that tolerance is right. It just does not need to cover a document whose top level is not an object. Checking the first non-space byte for `{` costs nothing and keeps the field-type case unscored. The log text is wrong for these payloads too: a bare array is a whole-document type error, not a field type mismatch.

### Smaller

Three inline, on the residual SSRF reachability claim, the compensating removal escaping the shutdown budget, and the reselect timer having no priority over `stopCh`. The cut-short log from ChiR4 still counts requeued peers against the whole cycle and still reads as a loss report, so I would take Chi's fix there.

### What I did not check

Sonar reports seven new issues behind the quality gate and I have no token for that project, so I have not read them. I did not run the smoke or sequential suites locally, only the p2p package; the upstream run covers them. And I have not soaked this on a live node, so the claim about scoring rates is static analysis plus the handler-level probe above, not observed gossip.

ctnguyen (commented):
Review at commit [d1fc4d2](https://github.com/gokutheengineer/teranode/commit/d1fc4d27dc63453bf60c0c2e4241dfbbe97f5add)

Second pass, and the follow-up commit does what it needed to. The `forget` marker clear is exactly the one-liner ChiR1 asked for, and unlike most one-line fixes it comes with a test that genuinely drives the interleaving it is named after — `startBlockedFlushPair` is a nice piece of test scaffolding and the enqueue-then-forget mirror really does fail without the `delete`. Folding the four duplicated decode-error blocks into `handleGossipDecodeError` is better than the narrow fix I suggested: one place to reason about, scoring semantics preserved exactly, and the type-mismatch half no longer writes `err`. The wider `isUnsafeIP` table and the `goTracked`/`lifecycleMu` shape still read well after the upstream merge, which changed the ground under them. Two things below: the regression test that is supposed to pin the ChiR2 fix never reaches the code it is testing, plus one log-volume note I should have caught the first time round. As before, this is static analysis — I could not run the package suite here.

## Status of previous points

Two caveats that do not fit in a cell. ChiR2's production fix landed and is correct, but the regression test added alongside it does not exercise it — that is ChiR5 below, not a reason to mark the point unfixed. ChiR3 is marked partial for attribution reasons: the settings-doc half is now accurate, but that came from the upstream merge `d1fc4d27d` pulling in `de9e5292a`, which rewrote the connection-time section and widened `util.DefaultSSRFDialPolicy` to cover private ranges and 100.64.0.0/10. This PR's own diff to `p2p_settings.md` touches only the static-check paragraph. The code comment in `validateDataHubURL` is unchanged and still overstates the connection-time overlap, though the residual is now much smaller than when I wrote the point — essentially `198.18.0.0/15` and `192.0.0.0/24`, since the rest is unroutable in practice.

| ChiR1 | Compensating removal skipped after reenqueue | YES | FIXED |
| ChiR2 | Unscored decode error echoes payload | | FIXED |
| ChiR3 | Connection-time reachability claim overstated | | PARTIAL FIXED |
| ChiR4 | Misleading requeue counts in log | | NOT FIXED |

## Blocking issues

### ChiR5 — the regression test for the decode-log bound is 12 KB long, so every handler drops it at the size check before decoding

**Problem:** `TestHandleTopics_MalformedInnerJSONScored/<topic>_field_type_mismatch_not_echoed_to_log` (`services/p2p/gossip_field_validation_test.go:645-656`) builds its payload from three copies of a 4000-character marker, so the fixture is at least 12000 bytes before the peer ID and the JSON punctuation. Every handler's first statement is a size check that returns before `json.Unmarshal`: `maxBlockMessageSize` and `maxNodeStatusMessageSize` are `maxGossipMessageSize` = 10240 (`services/p2p/Server.go:107`, `:117`, `:126`), `maxSubtreeMessageSize` and `maxRejectedTxMessageSize` are 8192 (`:118`, `:131`). So all four subtests return at `server_helpers.go:39`, `:229`, `:666` and `Server.go:1521`, having logged only the `"message size N exceeds max ..."` line. `handleGossipDecodeError` is never called, `require.NotContains(capture.errors(), marker[:64])` passes vacuously, and the test would pass identically against the pre-fix code that interpolated `err`.

**Why it matters:** blocking rather than a coverage nit, for three reasons. First, the defect is in the PR's own claimed evidence: the description asserts that "the unscored type-mismatch half never writes `err` to the log", and this subtest — added in the same commit as the helper — is the only thing offered in support. Merging banks a guarantee that nothing actually checks. Second, the asymmetry it guards is invisible from the helper's signature: `handleGossipDecodeError(handlerName, m, fromID, isSelf, err)` takes `err` and uses it on one branch only, so the two log lines look gratuitously different and unifying them is the natural tidy-up. That regression reintroduces the original ChiR2 flood — up to a topic-cap-sized `UnmarshalTypeError` per message, unscored — with the suite still green. Third, the fix is two lines. The sibling subtest `_field_type_mismatch_not_scored` (`:630-641`) uses a short fixture and does work, so the gap is specifically the log bound, which is the half that was added on review.

**Fix:** size the marker so the message stays under the smallest topic cap. A single mismatching field is enough, and 2000 bytes is still far more than any legitimate error line:

```go
    marker := strings.Repeat("7", 2000)
    mismatch := []byte(`{"PeerID":"` + remotePeerID.String() + `","Hash":` + marker + `,"TxID":` + marker + `,"fee_policy":` + marker + `}`)
    require.Less(t, len(mismatch), maxSubtreeMessageSize, "fixture must reach the decoder on every topic")
    require.True(t, json.Valid(mismatch))
```

Worth also asserting the positive side in the same subtest — that the captured lines do contain `"field type mismatch"` — so the test fails loudly if the payload stops reaching the helper again.

## Non-blocking issues

### ChiR6 — the node_status tip rejection is a per-message WARN on a path the PR deliberately leaves unscored

**Problem:** this one is not in the `5ff374781` delta — the line came in with the PR's first commit `abfba6cff` and my earlier pass missed it, so it is new to the review rather than new to the branch. The length guard in `sanitizeAdvertisedTip` (`services/p2p/server_helpers.go:812-815`) logs `Warnf` on every rejection. From `handleBlockTopic` (`:102-109`) that is bounded, because the caller immediately charges a protocol violation and the peer bans itself out. From `handleNodeStatusTopic` (`services/p2p/Server.go:1597-1608`) it is not: node_status is telemetry and is deliberately never scored for an unverifiable tip, so a peer sending `{"best_height":1,"best_block_hash":""}` gets one WARN line per message at whatever rate gossipsub allows, indefinitely and for free. There is no rate limiting anywhere in `services/p2p`. Before this PR the same message produced no log line at all, because `chainhash.NewHashFromStr("")` returned the zero hash with a nil error — so this is a sink the PR creates, not one it inherits.

**Why it matters:** it is only log volume, and the line is short and carries nothing peer-controlled beyond the bounded peer ID, so this is much milder than the original ChiR2. But it is the same shape of problem the PR just closed elsewhere, on the one gossip path that by design cannot charge for it, and it will also fire steadily against any peer implementation that reports a height without a tip hash — which would bury the interesting WARN lines on a mixed-version network.

**Fix:** drop the node_status case to debug level, keeping the warning where the caller scores. Cheapest version is to let the caller decide:

```go
func (s *Server) sanitizeAdvertisedTip(peerID string, advertisedHeight uint32, advertisedHash string, localHeight uint32, scored bool) (uint32, *chainhash.Hash, bool) {
    if len(advertisedHash) != chainhash.MaxHashStringSize {
        log := s.logger.Debugf
        if scored {
            log = s.logger.Warnf
        }
        log("[sanitizeAdvertisedTip] rejecting advertised tip from peer %s: block hash length %d, expected %d", peerID, len(advertisedHash), chainhash.MaxHashStringSize)
        return 0, nil, false
    }
```

The subtree path is unaffected: its rejection line comes from `parseHash`, which is always followed by a score at the call site.

## Recap

| ChiR5 | Oversized fixture bypasses decode test | YES | 45% |
| ChiR6 | Unscored per-message tip warning | | 15% |

Issue #1002, opened 2026-06-01
go-bt encode-side allocations still 50% of legacy alloc churn after #929 (decode arena)


oskarszoon:
## Summary

`go-bt/v2`'s tx/output **encode** path (`Output.appendTo`, `Tx.toBytesHelper`, `Output.Bytes`) accounts for ~50% of legacy's cumulative allocation churn during mainnet IBD. PR #929 (go-bt arena for decode) eliminated the decode-side allocations but the encode-side hash path is still per-call allocating fresh `[]byte`. Every tx that gets hashed (TxID, UTXO hash) goes through this path; on a busy legacy that's millions of allocations per minute.

## Observed

`allocs` (cumulative alloc-space) profile of legacy on `bsva-ovh-teranode-eu-2`, 2026-06-01, mainnet IBD at height 714k:

| % | GB allocated | Function |
|---|--------------|----------|
| 24.09% | 6,061 | `go-bt/v2.(*Output).appendTo` |
| 19.47% | 4,898 | `go-bt/v2.(*Tx).toBytesHelper` (cum: 11,090 / 44.08%) |
| 7.47% | 1,880 | `go-wire.(*blockArena).Alloc` (already optimized — expected) |
| 6.63% | 1,667 | `bytes.Clone` (called from `WireTxToGoBtTx`, intentional copy) |
| 6.47% | 1,628 | `teranode/util.UTXOHash` |
| 6.43% | 1,617 | `go-bt/v2.(*Output).Bytes` |
| 5.82% | 1,463 | `go-bt/v2.readArenaScript` |
| 2.13% | 537 | `aerospike-client-go/v8.batchCommandOperate.parseRecord` |

Of 25,158 GB cumulative allocation, **~50% (12,463 GB)** comes from the encode-side serialise-for-hashing triad: `appendTo` + `toBytesHelper` + `Output.Bytes`.

The decode-side arena (from #929) is working — `blockArena.Alloc` is only 7.5%. The issue is on the encode side.

## Why this is still happening after #929

PR #929 added `bt.Arena` + `Tx.HashTxIDInto` for the **bulk-stream decode** sites (subtreevalidation, blockpersister, asset/repository). It explicitly excluded single-tx call sites in legacy:

[...]

PR #1735, opened 2026-09-14
MvP-1012 perf(blob): cut the blob write path's CPU and allocations

### LINKED ISSUE #1012: blob/file.SetFromReader is the new top CPU sink (23-44%) on legacy IBD after #1001/#1002
## Summary

After #1001 (aerospike-client `nodeStats.updateOrInsert` fix) and #1002 (go-bt encode arena) shipped, the new top CPU consumer in the legacy mainnet IBD path is **`stores/blob/file.(*File).SetFromReader` at 23–44% of CPU**. Two callers dominate: subtree blob storage (`File.Set`) and `utxopersister/filestorer.NewFileStorer`. Each call performs a per-blob SHA-256, data fsync, parent-directory fsync, and a separate hash-sidecar file write — at the current sync rate (~1 block/s) that's many tens of fsync syscalls per second per node.

## Observed

CPU profiles from `bsva-ovh-teranode-eu-3`, mainnet IBD at height ~742k, two 30s samples on 2026-06-02:

### Snapshot 1 (steady state)

```
Type: cpu, Duration: 30s, Total samples: 37.82s (1.26 cores)
- 23.16%  blob/file.(*File).SetFromReader        ← #1
  - 60.39% from File.Set (subtree blob path)
  - 39.61% from utxopersister/filestorer.NewFileStorer
  - 67.69% in io.Copy (the actual writes)
  - 11.19% in syncAndCloseTempFile (fsync)
  -  7.76% in writeHashFile (sidecar)
```

### Snapshot 2 (~25 min later, mid-burst)

```
Type: cpu, Duration: 30s, Total samples: 33.60s (1.12 cores)
- 44.61%  blob/file.(*File).SetFromReader        ← #1, nearly half
- 37.11%  Syscall6 (mostly downstream of the above + futex churn)
- 23.21%  crypto/sha256.blockSHANI               ← per-blob hashing
- 25.83%  utxopersister/filestorer.NewFileStorer
- 26.67%  aerospike sendStoreBatch (productive)
```

Sync rate now ~0.8s/block average but with **10–14s outliers** (block 742581: 13.85s for only 1,476 txs) — the outliers correlate with bursts of subtree blob writes, where the file-store path becomes the binding constraint.

## What `SetFromReader` does today

`stores/blob/file/file.go:701-790`. Per blob:

1. Acquire a write semaphore permit
2. Construct filename, check overwrite policy
3. Create `.tmp` sibling file
4. Stream data via `io.Copy(file, io.TeeReader(reader, sha256.New()))` — every byte hashed
5. **`file.Sync()`** — fsync of the data file (`fsyncMode != none`)
6. `file.Close()`
7. Atomic rename `.tmp` → final
8. **`dir.Sync()`** — fsync of parent directory (`fsyncMode == full`, the default)
9. `writeHashFile()` — write a `.sha256` sidecar (separate file create + write + rename + fsync)

Default mode is `fsyncModeFull`. So every blob ⇒ **2 fsyncs + 2 file creations + 1 SHA-256 over the full payload**.

## Why this is the binding constraint now

- After #1001 / #1002, legacy is fast enough to drive ~1 block/s
- Each block produces ~10–49 subtree blobs (saw 49 in one block during prior profiling)
- Plus utxopersister files (≥1 per block)
- → **50–100+ blob writes/second**, each with the cost above
- The SHANI 23% is the SHA-256 over those payloads
- The 37% Syscall6 is dominated by `write(2)` + `fsync(2)` + the small-file create/rename churn

The file store's correctness comments are well-considered (parent-dir sync is explicitly load-bearing for crash safety on ext4/xfs/btrfs — see `syncParentDirBestEffort`'s 12-line doc comment). The cost is real and intentional. But the workload pattern (many small blobs per block, many blocks per second) is now exposing it as the dominant CPU sink.

### 1. Batched / coalesced fsync (highest impact)

Today: one `fsync(data) + fsync(dir) + writeHashFile` per blob.

Proposed: group N blobs (per block boundary, or time-window) and fsync once at the end. The atomic visibility guarantee comes from `rename(2)`, not fsync — fsync ensures durability after rename. If we delay fsync until after the block's full subtree set is on disk, crash recovery already needs to handle a partial-block state anyway (the block is not "accepted" until subtreevalidation/blockvalidation completes downstream).

Implementation sketch:
- `SetFromReaderUnsynced` writes data + rename, defers fsync
- `BatchedSyncer` accumulates pending paths, fsyncs all data files + parent dirs once per batch
- Caller (block-processing pipeline) calls `Flush()` at the block commit boundary

Expected: **30–40% legacy CPU reduction**, **~20% throughput win**.

### 2. Drop the per-blob SHA-256 sidecar OR move it off the hot path (high impact, low risk)

`writeHashFile` produces a `.sha256` sidecar for each blob (Goal-comment says "Format mirrors `os.WriteFile`'s pre-PR behaviour"). On a sync path the hash is recomputed every read anyway, so the sidecar is *informational*, not load-bearing.

Options:
- Make it opt-in via a file-store URL parameter (`writeChecksumSidecar=false`)
- Move sidecar writes to a background goroutine that batches them
- Drop entirely and lean on filesystem checksums (`btrfs`/`zfs`) or a single manifest per block

Expected: **~5–8% CPU win**, eliminates the second file create per blob, removes the second fsync.

### 3. SHA-256 SHANI batching (medium impact)

The 23% in `blockSHANI` is per-byte SHA-256 work, hardware-accelerated. Already optimized at the instruction level, but it's done **separately for each blob** via `TeeReader`. If subtree data is already hashed upstream (Merkle root computation), the file-store hash is redundant.

Audit whether the subtree's data hash matches the file-store's SHA-256. If yes, accept the upstream hash and skip the per-blob hashing here.

Expected: **10–20% CPU win** if redundancy is real. Worth a code search first.

### 4. Bigger `io.Copy` buffer (small impact, trivial)

`io.Copy` default chunk is 32 KiB. For multi-MiB subtree blobs that's ~30 syscalls each. Wrap the destination in a `bufio.Writer` with a 1 MiB buffer (or use `io.CopyBuffer` with a pre-allocated 1 MiB slice). Reduces `write(2)` syscall count by 32×.

Expected: **~3–5% CPU win** in the syscall column.

### 5. Move utxopersister storage off the local file store (large impact, large effort)

40% of `SetFromReader` traffic is utxopersister. Aerospike already holds the canonical UTXO set; the file-store copy is for external/diff replay. If the use case allows, route utxopersister writes to a different backend (postgres bulk insert, or an internal append-only log).

Out of scope for an incremental fix; flag for the architecture group.

## Verification

For (1) once implemented:
- [ ] 30s CPU profile pre/post on a busy legacy: `SetFromReader` should drop from ~44% to <15%
- [ ] Block throughput: blocks/second on eu-3 should rise from ~1 to >1.5
- [ ] Crash-recovery test: kill -9 the container during a sync, recover, validate downstream services don't observe corruption (subtreevalidation re-fetches missing data, blockvalidation rejects partial blocks)
- [ ] `gc-pause-ms` stable or lower (less small-object churn from sidecars + temp files)

For (2):
- [ ] Confirm no consumer reads the `.sha256` sidecar in the hot path
- [ ] If sidecar is kept, measure CPU before/after moving it to background goroutine

## Captured profiles (local)

- `probe/eu3-mainnet-sync-2026-06-02/legacy/` — first snapshot (23% in SetFromReader)
- `probe/eu3-mainnet-sync-2026-06-02-snap2/legacy/` — second snapshot (44%)

Available on request.

## Related

- #1001 — aerospike `nodeStats` CPU sink (fixed). This issue surfaced after that fix.
- #1002 — go-bt encode arena (fixed). Same — surfaced after fix.
- #920, #929, #936, #941, #953 — earlier IBD-stability issues in the same investigation series

## Not in scope

- File store **correctness** comments at `syncParentDirBestEffort` are sound — don't drop parent-dir sync naively. The proposed batching preserves the eventual durability guarantee, only changes its granularity from per-blob to per-batch.
- `bytes.Clone` heap pressure in `legacy/netsync.WireTxToGoBtTx` (1.5 GiB inuse) — separate concern, has its own load-bearing doc comment about arena aliasing.
- `go-batcher SetMaxConcurrent.func1` (21–27% CPU + 4,843 goroutines) — separate concurrency-primitives concern.

ctnguyen:
This resolve https://github.com/bsv-blockchain/teranode/issues/1012

`stores/blob/file.SetFromReader` became the top CPU consumer on mainnet legacy IBD once #1001 and
#1002 were fixed (23–45% across two 30s profiles). Every blob write paid a full-payload SHA-256, a
second create/write/fsync/rename/dir-fsync cycle for a `.sha256` sidecar, and a 32 KiB chunked copy
loop. This takes the parts that can be fixed without touching the durability schedule.

## What changed

- **`checksum` store URL parameter implemented** — it was documented but never parsed, so the
  sidecar was unconditional. Defaults `true`, so an unchanged URL behaves exactly as before. With
  `checksum=false` the payload is neither digested nor sidecar'd, and a sidecar from an earlier
  write is removed best-effort on republication.
- **In-memory `Set` writes and hashes once** instead of per 32 KiB chunk.
- **The copy buffer now takes effect.** `io.CopyBuffer` silently discards the caller's buffer when
  the destination implements `io.ReaderFrom`, and `*os.File` does — so the destination is wrapped
  to hide it, and the 1 MiB buffer comes from a `sync.Pool`.
- **`utxoPersister_buffer_size` default 4KB → 256KB.** It writes through a buffered writer over an
  unbuffered pipe, so this decides how many records coalesce per transfer.
- **Two redundant `Del(..., …+".sha256")` calls removed** — `Del` already removes the sidecar with
  its blob, and one of the two escalated a redundant delete into a block-aborting error.
- **`subtreestore` and `temp_store` set `checksum=false`**; `blockstore` keeps checksums. Both
  file-store knobs are now documented in `settings.conf`.

**Blob durability is unchanged** — the fsync/rename/parent-dir-fsync schedule is untouched and
`fsyncMode` keeps its `full` default. Only the sidecar's separate publication goes away.

## Measurements

Three distinct things, not one figure.

**(a) This change, before/after.** Baseline `13897b045` vs branch, identical benchmark names,
checksums enabled both sides, `mode=none` to remove fsync noise:

| | baseline | final |
|---|---|---|
| `Set` 256 B | 361055 ns, 37711 B/op | 161344 ns, 4821 B/op |
| `Set` 4 KiB | 393920 ns, 37658 B/op | 111050 ns, 4822 B/op |
| `Set` 1 MiB | 3021494 ns, 37714 B/op | 2597892 ns, 4877 B/op |
| streaming 256 B | 173291 ns, 37512 B/op | 107458 ns, 4681 B/op |
| streaming 1 MiB | 2763409 ns, 37618 B/op | 2803444 ns, 4819 B/op |

The consistent win is **allocation — ~37.7 KB/op to ~4.8 KB/op, ~8×** — the per-call 32 KiB buffer
replaced by the pooled one. Time improves at small payloads and is **flat at 1 MiB streaming**
(2.76→2.80 ms, within noise); no time win is claimed there.

**(b) What the checksum knob costs** (single revision, `BenchmarkSetFromReader_FsyncModes`):
1 MiB/`mode=none` 3044993 → 202573 ns (~15×); 1 MiB/`mode=full` 180882513 → 83227626 ns (~2.2×).
fsync dominates `full`/`data` on this host (80–180 ms), so `mode=none` is the honest CPU signal.
The axis measures digest and sidecar publication together.

**(c) What the buffer costs** (parameterised, `BenchmarkCompare_FileStorerWrite`, 4 MiB/iteration):
64 B records 26557316 → 17380486 ns (~1.53×); 512 B records 29000990 → 14669616 ns (~1.98×).

## Operator-visible change

**Subtree and temp stores stop writing `.sha256` sidecars; `blockstore` still writes them.** No
read path in the node verifies a sidecar, but an external script globbing `*.sha256` under those
two stores will find nothing after this change.

Revert: set `checksum=true` on the entry in `settings_local.conf` and restart — sidecars resume on
the next write per key, existing ones stay valid, no code change and no migration.

**Pre-upgrade check.** An invalid `checksum` value now fails at store construction where it was
previously ignored. Committed config contains no `checksum=`, but operator URLs and overrides are
not visible from the repo — check `grep -rn 'checksum=' settings*.conf` plus any deployment
overrides; anything not empty/`true`/`false`/`1`/`0` must be fixed first.

`utxoPersister_buffer_size` costs ~252 KiB more per concurrent file storer; set it back to `4KB`
on memory-tight deployments.

## Verification

Build, `go test -race -count=1` over `./stores/blob/...`, `./services/utxopersister/...`,
`./services/blockpersister/...`, `./settings/...` — all pass. `go vet` clean. `golangci-lint` on
the touched packages: 5 `prealloc` findings, all in files this change does not touch and identical
to baseline.

**Issue checkbox closed — no consumer reads the `.sha256` sidecar.** A repo-wide grep returns only
the constant, the writer, the two removal sites, docs, tests, and the two `cmd/bitcointoutxoset`
sites that write a digest beside their own output without going through the blob store. The other
five checkboxes are scoped "For (1)" and belong to the deferred proposal, or need production
measurement on the reporting host. No production CPU number is claimed here.

## Outstanding

- **Proposal 1, batched fsync** — deferred: it adds a `blob.Store` method across 8 implementations
  and changes durability granularity, so it needs its own ticket and a durability sign-off. The
  issue's 44% → <15% target lives there.
- **`blockstore` sidecars** — flipping it, or moving its sidecar write to a background goroutine,
  needs a ruling on external consumers of the published utxo-set artefacts.
- **Aerospike metrics re-enable + fork bump**, the follow-up to the #1001 stop-gap.
- **Four uncommitted docs corrections**: `fsyncMode` absent from `docs/`;
  `blob_settings.md:42` falsely claims sidecars are validated on read; and
  `utxopersister_settings.md` + `utxoPersister.md` still document the old `4KB` default.


ordishs (approved):
Verified locally on the branch: `go build ./...` clean, `go vet` clean, `go test -race -count=1 ./stores/blob/file/... ./settings/... ./services/utxopersister/...` all pass, `golangci-lint run` on those packages reports 0 issues.

I also confirmed the three central safety claims: `File.Del` removes `<blob>.sha256` before the blob (`stores/blob/file/file.go:1413-1417`) and returns nil for a missing blob, so the two removed `Del` calls in `Server.go` are genuinely redundant; no other blob backend writes a sidecar; and every `subtreestore` and `temp_store` context in `settings.conf` carries the new flag, with `blockstore` left alone.

Nothing below blocks the merge.

---

### `stores/blob/file/file.go:114-133` — the pooled copy buffer raises the transient memory ceiling 32x

```go
const copyBufferSize = 1 << 20 // 1 MiB

var copyBufPool = sync.Pool{
	New: func() any { b := make([]byte, copyBufferSize); return &b },
}
```

- `defaultWriteLimit = 256` (`file.go:266`) admits 256 concurrent writes. Each concurrent streaming write holds one buffer for the whole copy, so the ceiling is 256 MiB of live copy buffers.
- The path this replaces — `io.Copy` falling through `os.File.ReadFrom` to `genericReadFrom` — used a 32 KiB buffer per call, so the same ceiling was about 8 MiB.
- The buffers are pooled, so a steady workload retains them across GC cycles rather than releasing them between writes.
- The description documents a ~252 KiB per-storer cost for `utxoPersister_buffer_size` but not this one, which is larger.

The 1 MiB size is well matched to the subtree data path: `bufioReaderPool` in `services/subtreevalidation/check_block_subtrees.go:36` is also 1 MiB, so a pipe read there can fill the buffer. It is not matched to the utxopersister path: `NewFileStorer` wraps the pipe in a 256 KiB `bufio.Writer` (`services/utxopersister/filestorer/FileStorer.go:95`), so a `PipeReader.Read` there never returns more than 256 KiB and the remaining 768 KiB of each buffer is never used.

**Recommendation**: state the ceiling in the description and in the `copyBufPool` comment. If the ceiling is unwelcome, size the buffer to 256 KiB — that costs the subtree path four reads per megabyte instead of one and drops the ceiling to 64 MiB — or bound the pool independently of the write semaphore.

---

### `docs/references/settings/services/utxopersister_settings.md:9,42` and `docs/topics/services/utxoPersister.md:315` — stale default

Both still publish `4KB` as the `utxoPersister_buffer_size` default after this PR changes it to `256KB`. An operator sizing memory from `utxopersister_settings.md:9` budgets 4 KiB per concurrent storer; the node allocates 256 KiB.

The PR lists these as outstanding, but `AGENTS.md` asks for docs to move with the change, and this PR already adds `TestUTXOPersisterBufferSizeDefault` to stop exactly this class of drift between the struct tag and the code.

`utxopersister_settings.md:16` also states "Falls back to 4KB (4096 bytes) if parsing fails". The write-side fallbacks are 128 KiB (`filestorer/FileStorer.go:87`, `blockpersister/streaming_process_subtree.go:275`) and the read-side fallbacks are 4096 (`utxopersister/UTXOSet.go:424,464,614`). That mismatch predates this PR.

---

### `docs/references/settings/stores/blob_settings.md:41-43` — documents an integrity check that does not exist

```
- When `checksum = true`, creates .sha256 files alongside blobs
- Validates checksums during read operations
- Removes checksum files during deletion
```

No read path verifies a sidecar — the description says so, the new `checksum` field comment says so, and a grep for `checksumExtension` finds only the constant, the writer, the two removal sites and tests.

The claim was already false, but this PR turns `checksum` into an operator-facing knob and ships `checksum=false` on two stores. An operator reading this line will believe the PR removes an integrity check that never existed, and may keep checksums on for a reason that is not real. The same document does not describe `checksum=false`, the new startup failure on an invalid value, or `fsyncMode` at all.

**Recommendation**: delete the validation line and document both parameters here, reusing the wording from the new `settings.conf` block.

---

Test quality is high. `TestSetAndSetFromReaderProduceIdenticalBytes` asserts published bytes across payload size, checksum and header axes. `TestSetFromReaderUsesConfiguredCopyBuffer` counts reads, which is the only thing that catches the `io.CopyBuffer` / `io.ReaderFrom` trap. `TestChecksumModeTransitions` pins each of the four documented transitions.

settings/settings.go
```diff
@@ -249,7 +249,7 @@ func NewSettings(alternativeContext ...string) *Settings {
 			GetMissingTransactions:                getInt("blockvalidation_getMissingTransactions", 32, alternativeContext...),
 			QuorumTimeout:                         getDuration("block_quorum_timeout", 10*time.Second, alternativeContext...),
 			ProcessTxMetaUsingStoreBatchSize:      getInt("blockvalidation_processTxMetaUsingStore_BatchSize", 1024, alternativeContext...),
-			UTXOPersisterBufferSize:               getString("utxoPersister_buffer_size", "4KB", alternativeContext...),
+			UTXOPersisterBufferSize:               getString("utxoPersister_buffer_size", "256KB", alternativeContext...),
```
freemans13, line 252: This default is what makes every `NewFileStorer` allocate its full buffer up front (`bufio.NewWriterSize` in `FileStorer.go:95`, not pooled). My benchmark shows a one-blob storer going from ~15 KB/op at 4KB to ~320 KB/op at 256KB, and the legacy path creates at least five storers per block. The per-block figures are in the review body.

I'd keep 256KB and pool the writer. Put a `sync.Pool` of `*bufio.Writer` keyed to the parsed size in `filestorer`, `Reset(pipeWriter)` it in `NewFileStorer`, and in `Close`/`Abort`, once the flush is done, `Reset(nil)` it and return it to the pool. That keeps the coalescing win and the allocation stays close to the old 4KB figure. The same applies to the `bufio.NewReaderSize` readers in `UTXOSet.go:430,470,620` and `streaming_process_subtree.go:278`, which this default also grows.

If pooling is out of scope for this PR, it would be worth keeping the default where it is and stating the per-storer allocation in the description next to the live-memory figure. Either way, `BenchmarkCompare_FileStorerWrite` needs a small-payload row, because at 4 MiB per iteration it cannot show this cost.

settings.conf
```diff
@@ -1346,14 +1346,34 @@ subtree_quorum_path             = ${DATADIR}/subtree_quorum
...
+#
+#   fsyncMode=full|data|none  Default full: fsync the temp file, then fsync the parent
+#                         directory after the rename. "data" skips the parent-directory fsync,
+#                         so a freshly published filename can be lost across a crash while
+#                         already-published content survives - useful where directory fsync is
+#                         expensive or unsupported, notably NFS-backed paths. "none" skips both
+#                         and opts out of crash safety entirely; only appropriate when the
+#                         filesystem provides its own durability or the data is regenerable.
+#
+# Subtree blobs are internal, regenerable and DAH-pruned, and they dominate blob write volume
```
freemans13: "Regenerable and DAH-pruned" is not true for subtrees of persisted blocks. Block persister calls `SetDAH(..., 0)` on both `FileTypeSubtreeData` and `FileTypeSubtree` (`services/blockpersister/streaming_process_subtree.go:83-90`), which makes them permanent. `GetLegacyBlock` and `GetSubtreeData` in the asset service then serve those files to peers.

Turning checksums off is still fine, since no reader verifies a sidecar. The justification here should say that, rather than a retention property the files don't all have. An operator deciding whether to flip `blockstore` would reason from this comment.

stores/blob/file/file.go
```diff
@@ -823,6 +909,21 @@ func (s *File) SetFromReader(ctx context.Context, key []byte, fileType fileforma
...
+		// A blob written while checksums were enabled leaves a sidecar behind; republishing the
+		// blob without checksums would leave that sidecar describing content that no longer
+		// exists. Nothing in this package can prove a sidecar is absent — errorOnOverwrite treats
+		// every Stat error as "missing", renameTempFile has a documented TOCTOU window, and Del
+		// swallows a failed sidecar unlink — so the removal is attempted unconditionally rather
+		// than predicated on any such proof.
+		//
+		// Best-effort: a stale sidecar is inert (nothing in this repository reads one) and must
+		// not turn a successfully published blob into a failed write.
+		if removeErr := s.removeStorePath(filename + checksumExtension); removeErr != nil && !os.IsNotExist(removeErr) {
```
freemans13: Non-blocking. This runs on every write with checksums off, and the sidecar is almost always absent. `removeStorePath` opens an `os.Root`, walks each directory component with `openat`, then `removeat` tries `unlinkat` twice (plain, then `AT_REMOVEDIR`) before it returns ENOENT, and closes the fds. With one hash-prefix directory that is about six syscalls per subtree write.

That still beats publishing a sidecar, so the PR is a net win. The path here comes from `constructFilename`, not from a caller, so the traversal protection `os.Root` gives buys nothing at this call site. A plain `syscall.Unlink(filename + checksumExtension)` does the same best-effort removal in one syscall, which `Del` already does with `os.Remove`.

freemans13 (changes requested):
Requesting changes on one item: the new 256KB default for `utxoPersister_buffer_size` makes every file storer allocate about 21 times more than before. The PR's headline is an allocation cut, and on the legacy sync path this adds back more than the blob store saves. The fix is small, details on `settings/settings.go:252`. Everything else below is non-blocking.

I read branch tip `a7702f7f4`, which matches the PR head. ordishs already covered the 1 MiB copy-buffer ceiling and the stale `4KB` and sidecar-validation docs, so I have not repeated those.

**What I checked.** `go test -race -count=1` passes on `./stores/blob/file/`, `./services/utxopersister/...` and `./settings/`, run locally on darwin/arm64. CI shows 37 passing checks and one skipped fork-image job. I also repeated the sidecar-consumer grep: nothing in the repo reads a `.sha256` file back, so turning checksums off on the subtree and temp stores loses no verification. The removed `Del` calls in `services/utxopersister/Server.go` really are redundant, because `File.Del` unlinks the sidecar before the blob.

**Allocation per storer.** `bufio.NewWriterSize` allocates the whole buffer when `NewFileStorer` runs, and nothing pools it (`services/utxopersister/filestorer/FileStorer.go:95`). I wrote a throwaway benchmark outside the branch: one storer per blob, 50-byte records, `checksum=false`, `fsyncMode=none`, 2000 iterations, three runs each.

| payload | 4KB buffer | 256KB buffer |
|---|---|---|
| 250 B | ~180 µs, ~15 KB/op | ~188 µs, ~324 KB/op |
| 4 KiB | ~176 µs, ~15 KB/op | ~194 µs, ~322 KB/op |
| 64 KiB | ~284 µs, ~14 KB/op | ~227 µs, ~389 KB/op |

The time win is real once a storer carries tens of kilobytes, and your 4 MiB benchmark shows it growing from there. The cost is paid by every storer regardless of size. Per legacy block that is at least five storers: three in `writeSubtree` (`services/legacy/netsync/handle_block.go`, subtree, subtree data and subtree meta) and two in `NewUTXOSet` for additions and deletions. The consolidator adds two 256KB readers on top. That comes to roughly 1.8 MiB of garbage per block, where it used to be about 30 KiB. Early mainnet blocks are a few hundred bytes, so during that stretch of IBD nearly all of it is empty buffer. The benchmark in the PR hides this because a 4 MiB payload dwarfs a 256 KiB allocation, and the description quotes its ns/op but not its B/op.

The limit of this finding: I measured allocations in a micro-benchmark, not GC time on a syncing node. How much it costs depends on block rate and live heap. The allocation count itself is certain from how `bufio` works.

**Smaller points**, inline:

- The `settings.conf` rationale says subtree blobs are "internal, regenerable and DAH-pruned". Block persister makes the subtree and subtree-data files of every persisted block permanent, and the asset service builds legacy blocks from them. The decision still holds, because no reader verifies sidecars, but the stated reason is wrong.
- With checksums off, every write now spends about six syscalls removing a sidecar that almost never exists. It is still a net saving over publishing one. A plain unlink would cost one or two.

ctnguyen:
Thanks for your reviews

---

@ordishs

**Fixed**

- **O1 — the copy buffer's cost.** Recorded on `copyBufPool` itself. The comment states
  checked-out memory as the applied write concurrency times `copyBufferSize`, with 256 MiB at
  the default write limit given as an example of that limit rather than as a ceiling — the
  limit is configurable and validated only into `[MinSemaphoreLimit, MaxSemaphoreLimit]`. It
  separates retained memory from checked-out memory, since a `sync.Pool` is neither a cap nor a
  guaranteed floor, notes that `Set` checks out no buffer at all, and records that the 1 MiB
  size matches the subtree path but is largely unused on the utxo-persister path, where the
  pipe writer's buffer bounds each `PipeReader.Read`.

- **O2 — the stale `4KB` default.** Corrected to `256KB` in `utxopersister_settings.md` and
  `utxoPersister.md`. Both files also recommended tuning sizes that sat below the new default
  and so read as downgrades; those are re-anchored. The parse-failure fallback is documented by
  consumer rather than as one value: the file storer and the block persister's subtree-data
  reader fall back to 128KB, the three utxo-set readers to 4096 bytes. `settings.conf` gained a
  documented entry for the knob, which it did not previously have.

- **O3 — the integrity claim that does not exist.** The "validates checksums during read
  operations" line is removed rather than reworded, since no read path opens a sidecar. Two
  further defects in the same file are fixed while there: it stated the `checksum` default as
  `false` when it is `true`, and it carried no `fsyncMode` entry at all. The new text is taken
  from the `settings.conf` block so the two descriptions cannot drift.

  That `settings.conf` block was itself overstated — it promised that disabling checksums
  "removes any sidecar an earlier write left behind". The removal is attempted after the blob
  is published and a failure is logged rather than failing the write, so a stale sidecar can
  survive. Corrected in the same commit.

**Deferred**

- **O1, second half** — resizing the buffer to 256 KiB, or bounding the pool independently of
  the write semaphore. Resizing changes measured behaviour and independent bounding adds new
  machinery; neither is a correctness defect.

---

@ freemans13

**Fixed**

- **F1 — the per-storer allocation.** The `*bufio.Writer` is pooled. The `256KB` default is
  kept: the file storer writes through a buffered writer over an unbuffered pipe, so the size
  decides how many per-record writes coalesce into one transfer, and reverting it would also
  leave the store's 1 MiB copy buffer even less used on this path than it already is.

- **F1, the readers — all four, none left unpooled.** The three in `UTXOSet.go` (utxo-additions,
  utxo-deletions, previous-utxo-set) and the block persister's subtree-data reader. A consumer
  census decided this: every consumer reads through `io.ReadFull`, and the one `io.CopyN`
  resolves via `dst.(ReaderFrom)` because `*io.LimitedReader` is not an `io.WriterTo`, so no
  consumer needs any of the optional interfaces that embedding provided. The wrapper's
  `*bufio.Reader` is therefore a named field rather than an embedded one — embedding promoted
  `WriteTo`, and `io.Copy` selects a source's `WriteTo` before it ever calls `Read`, which would
  have let a caller stream past the guard and dereference a buffer already handed back. No
  caller does that today; the type now cannot.

  Pooling follows one ownership rule for the writer and the readers alike: every accessor takes
  the owner's mutex, the field is nil'd under that mutex before the buffer is returned and
  accessors are nil-safe, and the cancellation path takes no lock and sits behind no `Once`, so
  an abort is never held up behind another caller's release. `Abort` still signals the pipe
  unconditionally and still waits for the background goroutine.

- **F1, the benchmark.** Added as a sibling function at the payloads that show the cost, with
  `checksum=false`. Medians of six runs:

  | payload | 4KB buffer | 256KB buffer |
  |---|---|---|
  | 256 B | 11.7 KB/op, 118 allocs/op | 12.5 KB/op, 118 allocs/op |
  | 4 KiB | 12.4 KB/op, 118 allocs/op | 12.9 KB/op, 118 allocs/op |

  The ~320 KB/op you measured at `256KB` is gone; the two buffer sizes now differ by well under
  a kilobyte per operation, and the allocation count is identical across all four rows.

- **F2 — the subtree rationale.** You were right that the files are not all regenerable and
  DAH-pruned: the block persister clears the DAH on the subtree and subtree-data files of a
  persisted block, and those are served to peers. The comment now justifies the decision by the
  fact that actually carries it — no read path verifies a sidecar — and says plainly that these
  blobs are not all short-lived.

- **F3 — the stale-sidecar unlink.** Now `os.Remove`, which is what `Del` already does for the
  same path. This gives up `os.Root` confinement against symlinked path components;
  `validatePathWithinBase` does not replace it, being `Abs` + `Clean` + a prefix test with no
  symlink resolution, and `ConstructFilename` rejects `..` in `SubDirectory` without rejecting
  separators there. That residual is accepted rather than disproved: planting such a symlink
  already requires write access to the node's data directory.

**Deferred** — nothing.

---

**Two behaviour changes worth naming.** A write after the storer is closed or aborted now
returns an error rather than `io.ErrClosedPipe`. `Close` called after `Abort` reports the
background reader's error, there being no buffer left to fail a flush.

services/utxopersister/filestorer/pool.go
```diff
@@ -0,0 +1,79 @@
...
+// Why they exist: bufio.NewWriterSize and bufio.NewReaderSize allocate the whole buffer up
+// front, whether or not the payload ever fills it. At the shipped utxoPersister_buffer_size
+// that is a quarter of a megabyte per buffer, and the utxo-persister path builds several
+// writers and readers per consolidated block plus one of each per subtree. Without a pool
+// every one of those is garbage as soon as the blob is published.
+//
+// Neither pool sets New. Get on an empty pool then returns a nil any, which the comma-ok
+// assertions below handle; a New function would have to guess a buffer size, and every
+// caller here needs a specific one.
+var (
```
liam, line 20: `writerPool`/`readerPool` reimplement a bufio-buffer-pool pattern that already exists as a package-level `bufioReaderPool` in four other places: `model/Block.go:41`, `services/blockvalidation/get_blocks.go:722`, `services/blockvalidation/quick_validate.go:29`, `services/subtreevalidation/check_block_subtrees.go:34`. Five near-identical hand-rolled `sync.Pool`-of-`bufio.Reader`/`Writer` implementations now exist in the repo; a correctness fix or tuning change to the pool/reset logic has to be re-applied and re-reviewed in each place separately. Might be worth factoring into one shared, size-parameterized helper (not necessarily in this PR).

services/utxopersister/filestorer/pool.go
```diff
@@ -0,0 +1,79 @@
...
+	writerPool sync.Pool
+	readerPool sync.Pool
+)
+
+// acquireWriter returns a *bufio.Writer of size bytes writing to w, reusing a pooled buffer
+// when one of exactly that size is available. bufio cannot resize a buffer, so a pooled
+// writer of any other size is dropped and a new one allocated; the caller's size is never
+// silently widened or narrowed.
+func acquireWriter(w io.Writer, size int) *bufio.Writer {
+	if bw, ok := writerPool.Get().(*bufio.Writer); ok && bw.Size() == size {
```
liam, line 30: `acquireWriter`/`AcquireReader` silently discard (never re-`Put`) a pooled buffer whose size doesn't exactly match the request, instead of returning it for a future same-sized caller. `utxoPersister_buffer_size` parse failures fall back to different sizes per consumer (128KB for the file storer/subtree reader vs 4096B for the three utxo-set readers), and any buffer-size change between deployments causes mismatched `Get`s to steadily erode pool effectiveness with no visible signal. The doc comment on `AcquireReader` acknowledges this for the small-size-clamp case specifically, but the same silent-discard applies more broadly whenever sizes differ.

stores/blob/file/file.go
```diff
@@ -127,6 +189,25 @@ type File struct {
...
+	//
+	//   - true -> false, key rewritten: the sidecar left by the earlier write is removed
+	//     best-effort as part of the new publication, so it cannot survive describing content
+	//     that is gone.
+	//   - true -> false, key not rewritten: the sidecar persists and still matches the content
+	//     present.
+	//   - false -> true: the sidecar is regenerated on the next write to that key only. Keys
+	//     that are not rewritten have none; there is no backfill.
+	//   - Del removes the sidecar either way, and treats an absent one as success.
+	checksum bool
```
liam, line 210: This comment itself confirms the `.sha256` checksum sidecar is write-only: "Nothing in this repository reads a sidecar back, and no read path verifies one." This PR adds a per-store on/off toggle around it (`parseChecksum`, the `checksum` field, transition-mode semantics, best-effort stale-sidecar removal, and the accompanying tests) instead of removing the dead feature outright. That's meaningful new surface area (a URL parameter, transition-semantics edge cases) to keep an unread artifact optionally skippable in two stores, while blockstore/utxostore keep paying to write sidecars nothing consumes. Worth considering deleting the sidecar feature entirely as a follow-up rather than making it configurable.

liam (approved):
Reviewed the buffer-pooling and checksum-toggle changes end to end (pool lifecycle, Close/Abort/releaseWriter's sync.Once+mutex interplay, the checksum-default flip's interaction with Del's sidecar removal, writeBody/onlyWriter copy-path changes). No correctness, concurrency, or data-loss issue survived verification - the PR already has dedicated regression tests for the double-release, flush-error, and io.Copy/WriteTo bypass hazards a reviewer would normally flag here.

Three non-blocking cleanup/reuse notes below, left as inline comments. None of them block this PR.

freemans13 (approved):
Approving. The blocking item from my last review is fixed. The file storer's writer and all four readers now come from a pool, so the 256KB default no longer allocates a fresh quarter-megabyte for every storer. The two smaller points (the subtree rationale in `settings.conf` and the stale-sidecar unlink) are fixed too.

I read `947722860`, which matches the PR head.

**What I checked.** `go test -race -count=1` passes on `./services/utxopersister/...`, `./stores/blob/file/`, `./settings/` and `./services/blockpersister/`, run locally on darwin/arm64. CI shows 37 passing checks and one skipped. I also followed the pool lifecycle end to end:

- A `Write` blocked in the pipe holds `f.mu`. `Abort` unblocks it with `CloseWithError`, waits for the background goroutine, and only then releases, so a buffer never goes back while a write still owns it.
- `io.PipeWriter.Write` returns only after the reader has finished copying the slice, so bytes from a recycled buffer cannot reach the blob store late.
- On `Close`'s flush-error path the background goroutine can still be running when the buffer is released. That is safe, because the goroutine reads the pipe and never touches the buffer.
- Moving `bufio.Reader` from an embedded field to a named one changes nothing for the decoders. Nothing in `services/utxopersister` type-asserts to `io.ByteReader` or `*bufio.Reader`, and both `io.CopyN` calls go through `io.LimitedReader`, which never exposed `WriteTo`.
- The stale `4KB` defaults and the false "validates checksums during read" line are gone from the docs.

**Two notes, neither blocking.**

The new small-payload benchmark measures a warm pool. One buffer cycles through every iteration and GC rarely runs, so 12.5 KB/op is the best case. `sync.Pool` drops anything left unused across two GC cycles, and a storer that finds the pool empty pays the full 256 KiB, exactly as before this commit. The legacy path opens its three subtree storers concurrently once per block, so how often the pool is warm depends on block rate against GC rate. I have not measured that on a syncing node, and a micro-benchmark cannot show it. It is never worse than the unpooled code, so this is about how much the pool saves, not whether it is safe.

The commit message of `947722860`, and your reply above, say every reader accessor takes the owner's mutex. `pooledBufReader` has no mutex. It relies on a single-goroutine contract, which its own doc comment states correctly, and every current consumer closes on the same frame that reads. The code is right; only the description overstates it.

**What I did not do.** I did not mutation-test the new `FileStorer` tests (for example, delete `releaseOnce` and confirm a test fails), and nothing here ran on a live node.

icellan (approved):
Approving — nothing blocking. I went looking for the usual pooling hazards (buffer recycled while a Write/Read is in flight, double-Put, `io.Copy` selecting a promoted `WriteTo` past the closed guard, defer ordering in the blockpersister frame, whether `Del` still unlinks the sidecar now that the explicit `Del(...+".sha256")` calls are gone) and they're each handled correctly. Build, `go vet`, and `stores/blob/file` + `services/utxopersister{,/filestorer}` tests pass locally.

Four non-blocking notes, all operational/config rather than correctness:

- `stores/blob/file/file.go:110` — copy-buffer footprint is now `writeLimit x 1 MiB`, where the old `file.ReadFrom` fallback used 32 KiB. 256 MiB at the default write limit, and documented as such, but `writeLimit` validates up to 256_000 and is only clamped *down* by the fd budget. An operator raising write concurrency to match an fd ulimit silently authorises multi-GiB of checked-out copy buffers, plus `sync.Pool` retention until the next GC. Worth sizing `copyBufferSize` off the configured buffer setting, or capping total checked-out copy-buffer bytes.
- `stores/blob/file/file.go:95` — `parseChecksum` turns a previously-ignored URL parameter into a startup-fatal error. Nothing read `storeURL.Query().Get("checksum")` before this PR, so any value was inert. Now `checksum=yes`, `checksum=on` or `checksum=enabled` fails `newStore` and the node doesn't boot — and the old doc comment said "when set to `true`", which is exactly the phrasing that invites `yes`. If any helm chart, operator CR or `settings_local.conf` outside this repo carries one of those, it's an upgrade-time boot regression. Accepting the common truthy spellings, or logging and defaulting instead of erroring, removes the hazard.
- `services/utxopersister/filestorer/pool.go:52` — a size-mismatched buffer is dropped rather than returned to the pool, and the per-consumer fallback sizes deliberately differ (128 KiB for the subtree-data reader, 4096 for utxo-additions/deletions/previous-set). That path is only reachable on a parse failure of `utxoPersister_buffer_size`, but when it is: a subtree-data acquire Gets a 4096 reader, discards it and allocates 128 KiB; the next utxo-additions acquire does the mirror image. One typo'd setting turns pooling into pure allocation *and* garbage amplification instead of degrading to the pre-PR behaviour. Cheap fix: put the mismatched buffer back before allocating, or key the pools by size.
- `settings.conf:1375` — shipping `checksum=false` for every `subtreestore.*` context and `temp_store` holds up in-repo: the only `.sha256` readers/writers are `cmd/bitcointoutxoset` and `cmd/bitcointoutxoset/bitcoin/headers.go`, which write their own alongside their own output. Still worth calling out in the PR description, because operator tooling that verifies subtree blobs out-of-band loses its inputs on upgrade, and there's no backfill — flipping back to `true` only regenerates sidecars for keys that get rewritten.

Two nits, neither actionable today: `writeBody`'s `io.WriterTo` branch is only reachable when checksums are off, so a source with non-standard `WriteTo` EOF semantics would surface where the unconditional `TeeReader` used to hide it (nothing in-tree qualifies); and `memReader`'s `writeBody` branch writes `mr.buf` while ignoring the embedded `*bytes.Reader`'s position — correct while only `Set` constructs one, but it would double-write if `SetFromReader` is ever handed a partially-consumed `memReader`.

ctnguyen:
Thanks for the feedbacks

---

@liam

**Size-mismatched pooled buffers were dropped.** Fixed. `acquireWriter` and `AcquireReader` now
put a buffer of the wrong size back in the pool before allocating a new one, instead of
discarding it. One `Get` still inspects one buffer, so a mismatch still allocates — it just no
longer costs the pool the buffer it missed on. icellan raised this independently; it was the one
point both of you found, and it was taken first.

**Factor the two pools into a shared size-parameterised bufio-pool helper.** Deferred, as you
scoped it ("not necessarily in this PR"). The same pattern lives in `model/Block.go`,
`services/blockvalidation/get_blocks.go`, `services/blockvalidation/quick_validate.go` and
`services/subtreevalidation/check_block_subtrees.go` as well as here — five packages across three
services. That is its own ticket, not a rider on this one.

**Delete the write-only `.sha256` sidecar feature outright.** Deferred, as you scoped it. The
sidecar is an operator-facing artefact that `blockstore` still publishes and that out-of-band
tooling may consume — the same external-consumer question icellan raises. Removing it is a
product decision with an operator-visible blast radius and needs its own ticket and a sign-off.

---

@freemans13

**The small-payload benchmark measures a warm pool.** Fixed, as documentation. The benchmark's
doc comment now records that its B/op is a warm-pool best case: one buffer cycles through every
iteration and a benchmark loop rarely collects, a storer that finds the pool empty allocates its
whole buffer exactly as the unpooled code did, and what the pool actually saves on a syncing node
depends on block rate against collection rate, which this benchmark cannot show. No behaviour
changed — as you said, the point is how much the pool saves, not whether it is safe.

**"Every reader accessor takes the owner's mutex" in the commit message.** You are right, and the
claim is wrong. `pooledBufReader` has no mutex; it holds a reader, a closer and a `sync.Once`, and
relies on a single-goroutine contract that its own doc comment states correctly ("Not safe for
concurrent use: a Close must not overlap a Read"). There is no code defect — only the description
overstated it, exactly as you put it. The commit message is not being rewritten: it is already
published, and rewriting history for a prose error is not worth the disruption. Correcting it
here instead.

---

@icellan

**Size-mismatched buffers dropped rather than returned.** Fixed — see the first item under liam,
who raised the same point independently. The cheap arm you proposed (put the buffer back before
allocating) is the one taken. Keying the pools by size is the same restructuring question as
liam's shared-helper point and is deferred with it.

**`checksum=yes`/`on`/`enabled` turned a previously inert value into a node that does not boot.**
Fixed. The truthy spellings `true`/`1`/`yes`/`on`/`enabled` and the falsy `false`/`0`/`no`/`off`/
`disabled` are all accepted now, case-insensitively. Nothing read this parameter before it was
implemented, so any value anywhere was inert, and a store URL outside this repository could not be
ruled out. Your alternative — log and default instead of erroring — was not taken: it would
swallow genuine typos silently and diverge from the `parseFsyncMode` precedent next to it. A
value that is neither truthy nor falsy still fails at startup.

**Operators lose their sidecar inputs on upgrade, with no backfill.** Called out, in both places
an operator actually reads. `settings.conf` and
`docs/references/settings/stores/blob_settings.md` now state it: the subtree and temp stores no
longer write `.sha256` sidecars, out-of-band tooling that verifies subtree blobs loses its inputs
on upgrade, and there is no backfill in either direction — setting `checksum=true` again restores
a sidecar only for keys written after the change. Sidecars already on disk stay valid until their
key is rewritten or deleted.

**Copy-buffer footprint is `writeLimit × 1 MiB`, and `writeLimit` is only clamped down.**
Deferred. Both remedies you propose are design changes on the hot path: `copyBufPool` is a
package-level pool shared by every `File` instance while the buffer setting is per-store, so
sizing the buffer off the setting means either a per-size pool set or a per-store pool — the same
restructuring as liam's point — and a hard cap on checked-out bytes is a second admission-control
mechanism beside `writeSemaphore`. The documentation half is already in: `file.go` states that the
256 MiB figure is the default-limit example rather than a ceiling, and that the limit is
configurable and validated only into `[MinSemaphoreLimit, MaxSemaphoreLimit]`. The rest needs its
own ticket.

**Nit — `writeBody`'s `io.WriterTo` branch is only reachable with checksums off.** Deferred, on
your own scoping ("neither actionable today"). The branch is guarded by `hasher != nil`, and the
in-tree sources are `*bytes.Reader` and `io.Pipe`; there is no change to make without inventing a
source with non-standard `WriteTo` EOF semantics.

**Nit — `memReader`'s branch ignores the embedded `*bytes.Reader`'s position.** Deferred, same
scoping. `memReader` is unexported and `Set` is its only construction site, always from a fresh
`bytes.NewReader`, so a guard would defend against a caller that cannot exist in this package
today.

Issue #1736, opened 2026-09-14
Recovery follow-ups: large censuses, resume and retention tests


oskarszoon:
Stu approved PR #1715 at `6e88b943e` with the following nonblocking follow-ups. These are deferred from that PR, not fixed by resolving its review threads. Carry applicable changes into the `release/v0.15` recovery variant as well.

Prioritize expansion limits before running recovery against a busy mainnet store. The current limit fails closed, but a large healthy unmined population can produce unresolved findings and exit 2. Mainnet runtime and census distribution remain unmeasured.

- [ ] **Scale ancestry expansion and report exhausted limits.** `expandUnconfirmed` has separate global counters of 10,000 processed records and 10,000 input edges. Initial roots consume both budgets, so healthy roots can remain Unknown without an explicit limit reason. Stu's probe classified 10,000 of 10,050 legitimate two-level chains and left 50 Unknown. Use per-root or root-count-scaled bounds while preserving bounded memory/work, the two-body-pass ceiling and fail-closed proof rules. Expose expansion incompleteness separately from archive coverage, with a clear report reason. Add regressions above 10,000 roots and for multi-input roots, cycles and depth limits. [Review thread](https://github.com/bsv-blockchain/teranode/pull/1715#discussion_r4005445842).
- [ ] **Exercise positive retention through real history and Aerospike.** The assembly E2E and single-job fixtures currently use zero retention, although public `Run` refuses it. Advance authenticated fixture history/tip beyond a positive retention window, then prove a repair succeeds through real `LocalHistory.Check` and Aerospike for inline and paginated records; include a recent transaction that stays protected. [Review thread](https://github.com/bsv-blockchain/teranode/pull/1715#discussion_r4005445855).
- [ ] **Pin the configured retention adjustment in a regression.** Exercise the policy selected by `Run` with a positive `utxostore_blockHeightRetentionAdjustment`. The test must fail if selection is changed to global retention alone, and should preserve the global floor when the effective value is lower. Avoid a test that merely repeats the same expression. [Review thread](https://github.com/bsv-blockchain/teranode/pull/1715#discussion_r4005445863).
[...]

PR #1737, opened 2026-09-14
fix: backport IDLE replay recovery to v0.15

oskarszoon:
Backport of #1715 to `release/v0.15`. Confirmed transactions recreated after pruning can contaminate assembly and leave missing replay markers elsewhere in the UTXO store. Add one IDLE-only CLI job to inventory persisted state, authenticate local canonical evidence, repair proven components, and verify journaled results. Audit remains the default; apply requires operator isolation of every shared-store writer.

Includes the merged main implementation and the shared review fixes in #1740: parent spend coverage is required before unconfirmed classification, interrupted empty history initialization can resume before apply, and per-block retention queries use a block-leading index. Schema creation is atomic; foreign/corrupt artifacts remain refused.

Release adaptations remain limited to older SQL close and separate Spend/Create test APIs, registration against the release CLI dispatcher, and the runbook. No production service, protobuf, dependency or settings changes.

The release operator notes cover automatic legacy-sync RUN transitions, FSM persistence failures that can still report success, conflict processing that can erase absent-child evidence before the census, and the expression-spend path that bypasses replay markers. For stores with `utxostore_utxoBatchSize=1`, disable `aerospike_enable_spend_filter_expressions` on every writer before resuming. Keep the existing stored output batch size.

Validation of the latest shared fixes:

- All three regressions reproduced before the fix and pass afterward, including covered-parent, cancelled initialization, foreign-artifact refusal and query-plan checks.
- Recovery core, command and CLI race suites pass on both variants, including real Aerospike single-job coverage.
- Affected vet, CLI builds, recovery lint/staticcheck and recovery gosec pass on both variants.

Earlier backport validation at `512f3979c`: full release unit/race verification passed (121 tested packages, 51 without tests), native Aerospike recovery and assembly E2E passed with race detection, and all 34 CI checks passed. Existing dependency/toolchain vulnerabilities reported by CLI govulncheck remain outside this backport. New CI results are shown in the PR checks.

Deferred follow-ups remain in #1736, including ancestry expansion budgets/reporting and positive-retention integration coverage. Address the expansion limit before a busy-mainnet recovery run. Mainnet runtime is unmeasured; unsealed interrupted scans restart; uncertain components remain untouched. No production recovery has been run.

Refs #1708.

The cancellation regression covers interruption before SQLite initialization and recovery of the resulting empty database. It does not inject cancellation during schema creation or a process kill; transaction-rollback coverage is tracked in #1736.


docs/topics/commands/recoverReplayedTransactions.md
```diff
@@ -0,0 +1,189 @@
...
+**These checks detect changes; they are not an atomic write lock.** Operators
+must isolate writers. IDLE alone does not stop all already-running background
+workers: for example, the pruner's startup gate and catchup pause do not form a
+maintenance barrier. `--maintenance` acknowledges the isolation procedure.
+
+On release 0.15, the blockchain service changes its runtime FSM before persisting
+the new state. A persistence failure is logged and can leave persisted `IDLE`
+behind a different runtime state. Resolve any FSM persistence errors and keep
+external FSM transitions disabled throughout recovery; the CLI's persisted-state
+check cannot establish that writers are inactive.
```
freemans13: This note is accurate, but on v0.15 there is a second way out of IDLE, and "keep external FSM transitions disabled" doesn't cover it.

On v0.15, `Blockchain.Run` goes straight to `SendFSMEvent` without checking for IDLE, and the FSM allows RUN from IDLE. Legacy sync calls `Run` on its own in three places: `startSync`, `handleBlockMsg` and `blockHandler`. So if the legacy service is still running when a peer catches up, the node goes RUNNING. The recovery guard polls persisted state every 5 seconds, so writers can run for up to 5 seconds before it cancels the job. Combine that with the persistence failure this note already describes, and the guard never sees the change at all.

Main closes this path. `sendFSMConvenienceEvent` refuses an automatic RUN or CATCHUPBLOCKS while in IDLE.

Suggested addition: "On 0.15, legacy sync and other services request RUN automatically, and the blockchain service accepts it from IDLE. Stop those services, don't only avoid manual transitions." It may also help to say that `setfsmstate` reports success even when persistence failed.

Code: v0.15 `services/blockchain/Server.go:2952` and `:2858-2875`, `services/legacy/netsync/manager.go:664`, `:1531`, `:2259`. Main `services/blockchain/Server.go:3073-3077`.

stores/utxo/aerospike/recovery_census.go
```diff
@@ -0,0 +1,239 @@
...
+				fail("malformed mined metadata")
+			}
+		}
+	}
+	return nil
+}
+
+// SpendReferences includes mined parents and their pages, even when unsafe flags
+// prevent mutation. Unsupported records are already findings in Inventory.
+func (b *RecoveryBackend) SpendReferences(ctx context.Context, visit func(rr.SpendReference) error) error {
```
freemans13, line 203: This census only finds a missing marker when the parent still holds the child's 68-byte spend. On v0.15, conflict processing can erase that spend first. That v0.15-only code came in with 95527e669 (#1325), and main has none of it.

On v0.15, `ProcessConflicting` treats an absent spender as a "ghost" when no parent lists it in `deletedChildren`. It unspends that slot so the winning transaction can take it. That is exactly the #1708 population: a mined child that was pruned after its parent marker was lost, which can still happen through the combined-batch route of #1701.

Here is how it goes wrong. Mined child C is pruned and P's marker for it is lost. A reorg or conflicting block brings in W, which spends the same outpoint. v0.15 clears P's slot and W spends it. Recovery then sees the edge P to W, finds W present, reports nothing for that outpoint, and can finish Complete. Meanwhile an output that was spent on the canonical chain now belongs to W. On main the stale spend stays in place and recovery adds the marker.

This is a missed repair, not an unsafe write. It needs a lost marker plus a conflicting block, so it's rare. I don't think it blocks. It is still worth a runbook line: on 0.15, run recovery before the node processes any reorg or conflicting block after a suspected replay, because the conflict walk can remove the evidence. Once recovery has written a marker, the heal fails closed instead.

Code: v0.15 `stores/utxo/process_conflicting.go:165-248` and the absent-descendant classification around `:1040-1170`. The dependency here is `snapshotParents` in `recovery.go:765-767`.

stores/utxo/aerospike/recovery.go
```diff
@@ -0,0 +1,776 @@
...
+	}
+	k, err := b.key(before.Key)
+	if err != nil {
+		return rr.Record{}, err
+	}
+	p := recoveryPolicy(ctx, before.Generation)
+	p.RecordExistsAction = as.UPDATE_ONLY
+	p.GenerationPolicy = as.EXPECT_GEN_EQUAL
+	markers[child] = true
+	if err := b.client.Put(p, k, as.BinMap{fields.DeletedChildren.String(): markers}); err != nil {
```
freemans13, line 219: The marker written here blocks a replayed spend on v0.15's default Lua path, but not on its expression path. This is narrow, and noted for completeness.

v0.15 uses filter expressions for spends only when `aerospike_enable_spend_filter_expressions=true` and `utxostore_utxoBatchSize == 1` (`spend.go:487`). That filter checks the creating, conflicting and locked flags, coinbase maturity and list size. It never checks that the slot is still unspent. `ListSetOp` then rewrites the slot and `spentUtxos` goes up by one.

So after recovery marks P and deletes C, a replay of C passes the filter, and Create brings C back. P's `spentUtxos` also drifts, so the next census stops on a count mismatch. Main added a first-seen guard in #1153/#1175 (`spend_expressions.go:194-222`): a spent slot is filtered out and falls through to Lua, and Lua rejects the spend using `deletedChildren`. v0.15 doesn't have that guard.

Neither default takes this path (`false` and `128`), and I found no deployment config that sets batch size 1. A runbook prerequisite would cover it. So would having `Run` refuse that combination on 0.15. Separately, the missing first-seen guard is already a double-spend exposure on v0.15 with that config, regardless of recovery.

freemans13 (approved):
Approving. The backport is faithful, and the recovery tests really ran against Aerospike in CI. I found three places where v0.15's older node code weakens what the runbook promises operators. None is a defect in the backported code, and none needs a code change on this PR. Two are worth a runbook sentence before anyone runs `--apply` on a v0.15 store.

I read the branch at `512f3979c`.

**Parity.** Every production file is byte-identical to main at `6e88b943e`. That covers `pkg/replayrecovery`, the non-test files in `cmd/recoverreplayedtransactions`, and `stores/utxo/aerospike/recovery*.go`. The recovery hunks in `cli.go` also match main. The `--full-reset` help text change comes from main too, and it is accurate for v0.15's `ResetBlockAssemblyFully`.

**Test adaptations.** The `recoverySpendAndCreate` helper spends and then creates. That is the same order as v0.15's validator (`spendUtxos`, then `CreateInUtxoStore`) and main's Aerospike `SpendAndCreate`, so the tests replay the contamination the way production produces it. The assembly E2E checks the end state: the replayed child is present in the candidate before repair and gone after it, and the legitimate unmined transaction survives.

**CI.** The `test` job ran `make test` without `-short` on merge ref `f84db9c5`. It skipped 45 of 11,280 tests, and none of those 45 is a recovery test. So `TestRunJobAerospike`, `TestReplayRecoveryAerospikeAssemblyEndToEnd` and the six `TestRecovery*` tests all ran against real Aerospike containers. Locally I built the CLI, ran `go vet`, and ran `pkg/replayrecovery` and `cmd/teranodecli/teranodecli` under `-race`. All clean. I did not run the container tests locally.

**What differs on v0.15** (details inline):

1. On v0.15, legacy sync can move a node out of IDLE on its own. Main refuses an automatic RUN from IDLE. The runbook's v0.15 note only covers persistence failures.
2. v0.15's conflict processing can clear the parent spend that recovery needs as evidence, before recovery runs. Main has no such code. The result is a missed repair that can still report complete. It is never an unsafe write.
3. With `aerospike_enable_spend_filter_expressions=true` and `utxostore_utxoBatchSize=1`, v0.15's expression spend path ignores the replay marker, so a replay can undo a repair. Neither default does this (they are `false` and `128`), and I found no deployment that sets both. Main is not affected, because #1153/#1175 added a first-seen guard there that v0.15 lacks.

What I did not check: anything already approved on #1715 and unchanged here, and the #1736 follow-ups, which apply to this branch as they do to main.

icellan:
Two adversarial tests repeatedly reproduced incorrect classification and broken initialization resume; the history scan also contains an avoidable repeated full-index scan. Existing focused short tests and focused `go vet` passed, but full-suite verification was not completed.

- **[P2] Require indexed spend coverage for confirmed parents** — `cmd/recoverreplayedtransactions/history.go:1149-1154`

  When the 10,000-item expansion budget is exhausted, a confirmed parent can be retained as another target's spender without itself belonging to `targets`. Since `indexBlock` indexes spends only for target parents, this zero-count result does not prove that the output is unspent. An adversarial test reproduces classification of a canonical double-spend as `Unconfirmed`, which discovery counts as kept rather than unresolved. Require the confirmed parent's spend coverage before accepting this absence result.

- **[P2] Rebuild history after interrupted schema initialization** — `cmd/recoverreplayedtransactions/run.go:417-419`

  An interruption after `NewHistory` creates the database and lock files but before creating `coverage` leaves a valid pre-apply initialization state. On resume, `openHistory` returns `no such table: coverage`, which this branch rejects instead of rebuilding. This reproducibly prevents the documented resume procedure despite no mutations having occurred. Recognize provably uninitialized history databases separately from corrupt sealed indexes and rebuild them during the pre-apply phase.

- **[P2] Avoid scanning all retained transactions for every block** — `cmd/recoverreplayedtransactions/history.go:542-546`

  For retained-history runs with many targets, this query scans every accumulated transaction for each successfully parsed block: the only transaction index starts with `txid`, not `block`. SQLite's query plan confirms a full covering-index scan, making this step scale with blocks times retained transactions. Track whether the current block retained any transactions during insertion, or add a block-leading index, to avoid substantially extending the maintenance window.

docs/topics/commands/recoverReplayedTransactions.md
oskarszoon: Addressed in `fd40b59b8`. Added the release-specific maintenance caveat to the runbook: legacy sync can automatically request RUN from IDLE, so operators must stop automatic transition sources as well as writers. It also states that persistence failure can still return command success, leaving persisted IDLE inconsistent with runtime. The CLI gate is not a service isolation barrier.

stores/utxo/aerospike/recovery_census.go
oskarszoon: Addressed in `fd40b59b8`. Documented that release conflict processing can clear parent references to absent children when their replay markers are missing. Operators must isolate promptly, preserve incident backups, and cannot treat a complete report as certification of evidence erased before the census. No conflict-service changes were added to this backport.

stores/utxo/aerospike/recovery.go
oskarszoon: Addressed in `fd40b59b8`. Added the batch-size-one expression-spend caveat: set aerospike_enable_spend_filter_expressions=false on every writer before resuming, while keeping the existing utxostore_utxoBatchSize. The runbook explicitly states that recovery does not repair that release spend path.

oskarszoon:
@icellan Fixed all three shared findings in `fd40b59b8`; the same two-file code/test change is in main follow-up #1740 (`7ce2f1a44`).

Parent inclusion now requires target spend coverage before an empty spend lookup can support Unconfirmed. Empty, unclaimed history initialization can rebuild through the existing pre-apply resume path, and schema creation is atomic; foreign metadata/schema and corrupt seals remain refused. A block-leading index removes the per-block transaction-table scan.

All three regressions failed before the fixes and pass afterward. Both variants pass recovery/core/CLI race suites, native Aerospike and assembly recovery tests, CLI builds, affected vet, recovery lint/staticcheck and recovery gosec. Stu's three release operator threads are documented and resolved. The expansion-budget follow-up remains in #1736. Please review this update and #1740.

oskarszoon:
CI unit job failed three unchanged blockchain-service startup tests (`TestShouldHandleReorgWithLongerChain`, `TestResetWithBlockchainAhead_Integration`, `TestHandleReorgWithInvalidBlock_Integration`). The first reported IDLE after startup; the other two could not connect to the blockchain gRPC listener. Recovery command/core tests passed in the same CI run.

The failing tests and startup helper have no diff against `upstream/release/v0.15`. All three passed locally in three repetitions with `SETTINGS_CONTEXT=test go test -race -tags testtxmetacache -count=3` (10.335s). Reran only the failed unit job without code changes; the rerun is still running. [Initial job log](https://github.com/bsv-blockchain/teranode/actions/runs/34880103762/job/104096989184).

icellan:
Local head `512f3979c` still has the confirmed performance and resume defects below; origin could not be fetched because of sandbox/network restrictions. Engine tests, engine race tests, command/CLI short tests, and targeted vet passed; broader Aerospike tests were blocked by Docker/socket restrictions.

- **[P2] Avoid scanning all retained transactions for every block** — `cmd/recoverreplayedtransactions/history.go:542-546`

  When many target transactions have been indexed, this query scans all previously retained transaction IDs for every subsequent authenticated block. The only transaction index is `(txid, block)`, which cannot efficiently serve a block-only lookup; `EXPLAIN QUERY PLAN` confirms a full index scan. This adds potentially quadratic work to the genesis-to-tip scan and can substantially extend the required maintenance window. Track whether the current block retained any transactions during insertion, or add an index on `block`.

- **[P2] Recognize an uninitialized history database during resume** — `cmd/recoverreplayedtransactions/run.go:417-419`

  If recovery stops after `NewHistory` creates `history.sqlite` and its lock file but before creating the `coverage` table, every subsequent `--resume` fails here. Opening that legitimate interrupted initialization state produces `no such table: coverage`, not either accepted sentinel; I reproduced this with empty private database and lock files. Since this branch runs before mutations begin, recognize and rebuild a provably uninitialized history database while continuing to reject corrupt sealed indexes.

icellan (approved):

PR #1738, opened 2026-09-14
Close consensus gaps on the quick-validation route

icellan:
Replaces #1730, which predated #1390 and mostly duplicated it. This started as the two leftover conjuncts from that reconciliation; review then found three consensus gaps on the quick-validation route, and fixing them pulled in two more. Description rewritten to match what is actually here.

The quick-validation route runs for blocks at or below a hash-verified checkpoint — the initial-block-download path — and commits blocks with `subtrees_set` and `mined_set` without ever calling `Block.Valid`. It is a hand-maintained mirror of `Valid`, and it had drifted. The structural problem is filed as bitcoin-sv/teranode#4836; this PR fixes the consensus-relevant instances.

## The fee skip was not gated on the binding

`checkBlockRewardAndFees` already took `merkleRootChecked`, but only to classify its verdicts; the skip itself had three conjuncts. Its premise is that the checkpoint transitively commits the coinbase, which holds only once the body has been hashed against the header. The sibling `validOrderAndBlessed` skip already required it.

No live path changes — every production `Valid` caller passes a subtree store, so real blocks are bound and keep the skip.

## CVE-2012-2459 had no counterpart on the quick route

The merkle root cannot detect a duplicated trailing transaction: the duplicate-last-node-when-odd rule makes the mutated body produce the same root, and so the same block hash, as the honest block. A body that binds to a checkpoint-certified header can therefore carry a repeated transaction while reusing the honest block's proof of work, at no mining cost.

`model.CheckSubtreeSlicesForDuplicateTxs` exists for exactly the paths that hold subtree slices and skip `Valid`, and its own doc calls it mandatory before UTXOs are created. **It had no production caller.** `validateSubtrees` now calls it, classified corrupt — the defect is in the body a peer served, and the honest hash it binds to must not be condemned.

The scan skips slot `[0][0]` only when it holds the coinbase placeholder, so "first node is the placeholder" is its unstated precondition. `Valid` enforces that as its step 7; the quick route did not, and a first subtree whose node 0 is a real txid passes the merkle check with the body's true first transaction silently substituted. Now checked.

## The coinbase was not shape-checked, and the predicate was too loose

`block.CoinbaseTx.IsCoinbase()` was never evaluated for the subtree-carrying shape anywhere on this route; the check in `getBlockTransactions` inspects `subtreeData.Txs[0]`, a different object. Added at both entry points for the no-subtree shape and in `validateSubtrees` for the subtree shape, each behind its own binding, so `BlockInvalid` is the right class in both.

Separately, go-bt's `Tx.IsCoinbase` accepts a null prevout hash plus **either** a `0xFFFFFFFF` index **or** a `0xFFFFFFFF` sequence number. Consensus requires the null hash **and** the index (`COutPoint::IsNull`); the sequence number says nothing about coinbase-ness. A transaction with prevout `(0x00…00, 0)` and sequence `0xFFFFFFFF` is accepted by go-bt and rejected by svnode. `model.IsConsensusCoinbase` applies the conjunction.

It is applied at `Block.Valid` step 4a as well as on the quick route, because `Valid` is where the verdict is decided: above the checkpoint it is the only check that runs, and on catch-up a quick-route rejection that is neither corrupt nor incomplete comes back through it. A looser test there overruled the strict one. Teranode's own coinbases are unaffected — `makeCoinbase1` writes a zero hash and index `0xFFFFFFFF`.

`SubmitMiningSolution` accepts a miner-supplied coinbase and checked only the input count and scriptSig length. A malformed submission would now fail `Valid`, which block assembly treats as "the subtreeprocessor created an invalid block" and answers with a reset. The submission path applies the same predicate first, so that is rejected as the submitter's error instead.

## The invalid verdict routed by check order

`validateSubtrees`' invalid verdict was wrapped by its callers in `ErrProcessing`. `errors.Is` walks the cause chain, so the wrap does not hide the class — it makes the error match **both**, and which one a call site sees depends on the order it tests them. The corrupt verdict is returned unwrapped for that reason; the invalid one now is too.

## Tests

Every guard here has a fixture that fails when the guard is reverted, confirmed by mutation rather than assumed. Two cases needed specific work:

- The earlier coinbase tests used an ordinary spend, which go-bt's predicate and the strict one both reject — so the subtree-shape site could be swapped back to `Tx.IsCoinbase()` with the suite green. The loose-but-not-strict shape (null prevout hash, index 0, sequence `0xFFFFFFFF`) now pins it, on the route and through `Valid`.
- The unwrapped verdict needed a subtree body driven end to end, since the no-subtree guard returns before any wrap. Asserting `Is(err, ErrBlockInvalid)` cannot see the wrap, so it asserts `Is(err, ErrProcessing)` is false.

## Verification

`model`, `model/time`, `services/blockvalidation` (+ `catchup`), `services/blockassembly` (+ `mining`, `subtreeprocessor`), `services/legacy/netsync` green under `-race -tags testtxmetacache`, `go vet` clean, after each commit.

## Deliberately not here

Left on bitcoin-sv/teranode#4836: no fee arithmetic on the quick route, no `storeInvalidBlock` (so an invalid verdict does not persist), `legacyUnifiedRoute` gating on height without checkpoint-hash proof, the dead shape check in `getBlockTransactions`, `bad-cb-multiple`, and the fact that the body checks — the new duplicate scan included — run after the UTXO batches rather than before. That last one is pre-existing for every `validateSubtrees` failure and needs the route restructured, not a guard moved.


services/blockvalidation/quick_validate.go
```diff
@@ -268,6 +268,19 @@ func (u *BlockValidation) quickValidateBlock(ctx context.Context, block *model.B
...
+	// The binding says the header commits to this body; it does not say the single
+	// transaction in a no-subtree body is coinbase-shaped. The route's other coinbase
+	// checks are len(Inputs) != 0 and the scriptSig length bound, neither of which
+	// rejects an ordinary spend standing in for the coinbase. Deliberately placed AFTER
+	// the binding: the transaction is then the miner's own committed body, so a
+	// non-coinbase one is genuine invalidity — ahead of the binding the same failure
+	// would be indistinguishable from a corrupted download (bitcoin-sv/teranode#4692).
+	// A no-op when the body carries subtrees; there the coinbase is bound by
+	// validateSubtrees' CheckMerkleRoot and checked as part of the subtree body.
+	if len(block.Subtrees) == 0 && !block.CoinbaseTx.IsCoinbase() {
```
freemans13: This check only runs when `block.Subtrees` is empty, and the comment above says a subtree body's coinbase is "checked as part of the subtree body". Nothing on the quick path checks it. The only shape check on that route is in `readSubtree`, on the subtree data file's first slot, and go-subtree never writes the coinbase there, so that `tx` is always nil. I reproduced a merkle-bound subtree block with an ordinary spend as `block.CoinbaseTx`: both entry points return nil and call `AddBlock`. Details and a tested fix, moving the check into `checkQuickValidationCoinbaseLength` where both shapes are already bound, are in the review body. The same applies to the async twin at line 396.

services/blockvalidation/quick_validate_coinbase_shape_test.go
```diff
@@ -0,0 +1,91 @@
...
+// the single transaction in that body is coinbase-shaped. The quick path's own coinbase
+// checks are len(Inputs) != 0 (BlockIncomplete) and the scriptSig length bound, neither
+// of which rejects an ordinary spend standing in for the coinbase.
+//
+// The check belongs AFTER the binding: once the header commits to this transaction, a
+// non-coinbase one is the miner's own committed body and so genuinely invalid. Ahead of
+// the binding the same failure would be indistinguishable from a corrupted download,
+// which is the classification bitcoin-sv/teranode#4692 settled.
+
+// TestQuickValidateBlock_BoundNonCoinbaseBodyRejected builds a coinbase-only block whose
```
freemans13, line 25: Nit: this doc comment names `TestQuickValidateBlock_BoundNonCoinbaseBodyRejected` but sits on `buildBoundNonCoinbaseBody`, and the block above it doesn't attach to any declaration. godoc will show it as the helper's doc.

freemans13 (changes requested):
Requesting changes for one thing. The new coinbase-shape check only covers blocks with no subtrees, and I reproduced the same hole on blocks that do carry subtrees. A block whose header commits to an ordinary spend in the coinbase slot is committed on both quick-validation entry points. The fee-skip half of the PR is right as it stands.

I read commit 16e653887, which is also the head the PR shows.

**Blocks with subtrees still skip the shape check.** The new comment says that on a subtree body "the coinbase is bound by validateSubtrees' CheckMerkleRoot and checked as part of the subtree body". The first half is true. The second half is not. The only shape check on that route is the `!tx.IsCoinbase()` test in `readSubtree`, and it runs on the subtree data file's first slot. go-subtree never writes the coinbase into that file: `serializeFromReader` starts at index 1 when node 0 is the placeholder, and `Serialize` skips it too. So that `tx` is always nil, and nothing on the quick path ever looks at `block.CoinbaseTx.IsCoinbase()` once subtrees are present.

To check it I wrote a throwaway test in a scratch worktree, not pushed anywhere. It reuses the fixture from the subtrees-present case in `TestQuickValidateBlock_CoinbaseLengthBinding`: one subtree holding the placeholder plus two real txs. `block.CoinbaseTx` is swapped for the same ordinary spend `buildBoundNonCoinbaseBody` uses, and the header root comes from `RootHashWithReplaceRootNode(spend.TxIDChainHash(), 0, 0)`, so the body is really bound. Result at 16e653887:

```
sync:  err=<nil> addBlockCalls=1
async: err=<nil> addBlockCalls=1
```

Your own reasoning covers this shape word for word. Once `validateSubtrees` has passed, the coinbase is the one the miner committed to, so a non-coinbase there is real invalidity. `checkQuickValidationCoinbaseLength` already sits at the point where both shapes are bound, and its doc comment says exactly that. I moved the check there with no `len(block.Subtrees) == 0` condition:

```go
if !block.CoinbaseTx.IsCoinbase() {
    return errors.NewBlockInvalidError("[%s][%s] block coinbase tx is not a valid coinbase tx", caller, block.Hash().String())
}
```

With that change the probe returns `BLOCK_INVALID` with zero `AddBlock` calls on both entry points. Your `TestQuickValidateBlock_BoundNonCoinbaseBodyRejected`, `TestQuickValidateBlock_CoinbaseLengthBinding` and `TestQuickValidate_CoinbaseOnlyBodyBinding` all still pass. There's one trade-off. On the no-subtree shape the check would then run after `AssignBlockID` instead of before it, and on the subtree shape after the UTXO writes. The length check already accepts both. If you'd rather keep the early rejection for coinbase-only bodies, leave your guard where it is and add this as well. Either way the comment at both sites needs fixing, and a subtrees-present case belongs next to the two you added.

This is defence in depth, like the rest of this route. Quick validation only runs at or below a hash-verified checkpoint, where a bound body is historical data. It's still the exact case the PR sets out to close, and a comment claiming coverage that doesn't exist is worse than no check at all.

**The fee skip is correct.** The skip now needs `merkleRootChecked`, which matches the `validOrderAndBlessed` skip. Every production `Valid` caller in blockvalidation passes `u.subtreeStore`. The blockassembly caller passes a nil UTXO store, so it never took the skip before this change either.

**The tests catch what they claim to.** I broke each change on its own and re-ran:
- guard removed from `quickValidateBlock`: the sync subtest fails
- guard removed from `quickValidateBlockAsync`: the async subtest fails
- `merkleRootChecked` dropped from the skip condition: `TestCheckBlockRewardAndFees_SkipsBelowHardcodedCheckpoint` fails

**A pre-existing issue, not this PR's to fix.** go-bt v2.6.9's `IsCoinbase` requires a zero prev-txid and then accepts either index `0xFFFFFFFF` *or* sequence `0xFFFFFFFF`. svnode's `IsCoinBase` wants the null prevout, meaning zero hash *and* index `0xFFFFFFFF`. So a single-input tx spending `000…000:0` with a final sequence passes here and in `Valid`'s step 4a, where svnode would reject the block. Mentioning it only so nobody reads the new check as matching svnode exactly.

**The red `test` job isn't from this PR.** It's a package-level `FAIL services/blockchain` with no failing test named in the log. The PR doesn't touch that package, and `go test -race -tags testtxmetacache ./services/blockchain/` passes locally at 16e653887 in 39s.

What I didn't check: the full `services/blockvalidation` and `model` suites, which I ran only for the tests named above, and any run against a real node.

ctnguyen (commented):
Review at commit [16e6538](https://github.com/icellan/teranode/commit/16e6538870b22161d8ad52998ec36b107935746b)

Two small, independent gaps closed with a 140-line diff — that restraint is welcome, and splitting them cleanly in the description made this quick to reason about. The fee-skip change is the stronger half: I re-derived it and it holds. `merkleRootChecked` is set at `model/Block.go:799` (coinbase-only binding) and `:848` (after `CheckMerkleRoot`), it already gates the sibling `validOrderAndBlessed` skip at `:1020`, and adding it as the fourth conjunct at `:1263` genuinely makes both below-checkpoint skips engage on the same set of blocks — which is what the comments on both already claimed. It is strictly stricter, and the only `Valid` caller without a subtree store (`services/blockassembly/Server.go:1974`) passes a nil UTXO store, so it never reaches `checkBlockRewardAndFees` at all. The fixture tweak in `TestBlock_Valid_SkipRefusedWhenMerkleRootNotChecked` is harmless too: the header is mined over `st.RootHash()`, which does not depend on the coinbase outputs. Covering both quick-validate entry points rather than just the synchronous one was the right instinct. The second change is where the problem is: it ships with a comment asserting coverage the route does not have.

## Blocking issues

### ChiR1 — Coinbase-shape check is skipped on subtree-carrying blocks, and the comment says otherwise

*Same as [@freemans13's point "Blocks with subtrees still skip the shape check." (14-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1738#discussion_r4006659916).*

**Problem:** The new guard at `services/blockvalidation/quick_validate.go:280` and `:396` is gated on `len(block.Subtrees) == 0`, and the comment two lines above each closes with "A no-op when the body carries subtrees; there the coinbase is bound by `validateSubtrees`' `CheckMerkleRoot` and checked as part of the subtree body." I re-derived both halves. *Bound* is true: `CheckMerkleRoot` (`model/Block.go:2405`) injects `b.CoinbaseTx.TxIDChainHash()` at root node 0 via `RootHashWithReplaceRootNode`. *Checked* is false, and it fails for two independent reasons rather than the one. First, `git grep -n "IsCoinbase()"` over tracked non-test code in `model/` and `services/blockvalidation/` returns seven lines — two of them prose mentions inside comments (`model/Block.go:921` and `:1063`) — leaving five call sites: `model/Block.go:922` (`Valid` step 4b, which this route never reaches — `checkQuickValidationCoinbaseLength`'s own doc comment says so), the two new guards at `quick_validate.go:280` and `:396`, `quick_validate.go:1043`, and `services/blockvalidation/skip_unspendable.go:23`. The last of those is not a shape check at all: it passes `tx.IsCoinbase()` as an argument to `utxo.HasNoSpendableOutputs`, consuming coinbase-ness to decide spendability rather than validating it, so it can neither accept nor reject anything on shape grounds. That leaves `quick_validate.go:1043` as the only candidate on the subtree route. Second, that remaining site does not test `block.CoinbaseTx` at all — it tests `subtreeData.Txs[0]`, a different object read off the data file, and it is guarded by `tx != nil`. In go-subtree v1.4.6 (the version in `go.mod`, not the v1.2.0 named elsewhere) both writers `continue` on `idx == 0` when node 0 is the placeholder (`subtree_data.go:204`, `:246`), and `serializeFromReader` only fills `Txs[0]` under `if txIndex == 1 && tx.IsCoinbase()` (`:300`), so on a normally-produced file that entry is nil and the guard short-circuits. Even when it does not, it returns a `ProcessingError`, not a block verdict. So on the subtree shape nothing between the binding and `commitBlock` looks at `block.CoinbaseTx`'s shape: `checkQuickValidationCoinbaseLength` only applies the scriptSig length bound, and `AddBlock` follows with `subtreesSet` and `minedSet`. One thing I would add to the reviewer's report is that the two entry points are not equally exposed. On the native catchup path the header is hash-pinned (`verifyCheckpointsAgainstHeaders`, `catchup.go:1237`), so a substituted coinbase changes the merkle root and `CheckMerkleRoot` rejects it — the gap there is defence-in-depth, the same class as the check the PR adds. The legacy unified route is not hash-pinned: `legacyUnifiedRoute` (`Server.go:1852`) gates on `model.OutpointOnlyEligible`, which is height-only (`model/checkpoint.go:54` → `BelowCheckpoint`), so the header need only be *below* the highest hardcoded checkpoint, and an attacker who mines it chooses the merkle root and the body together. That route is opt-in and `blockvalidation_legacy_unified_below_checkpoint` defaults to `false`, so this is a narrow exposure, not a live one — but it is the shape where the missing conjunct actually costs something.

**Why it matters:** The PR's premise is that the quick path needs its own coinbase-shape check because it never calls `block.Valid`. That premise applies to both body shapes equally; the implementation covers one and the comment states the other is covered. A reader of `quick_validate.go:278-279` has no reason to look further, which is exactly how the automated review on this PR concluded the gating was correct. Leaving a false coverage claim next to a half-closed gap is worse than leaving the gap documented as open, because it removes the prompt for anyone to finish it. Both new comments also cite [bitcoin-sv/teranode#4692](https://github.com/bitcoin-sv/teranode/issues/4692) for the ordering argument, which is sound and settled — the ordering is not the issue, the scope is.

**Fix:** Move the check into `checkQuickValidationCoinbaseLength` (`quick_validate.go:469`) with no `len(block.Subtrees) == 0` condition. That function's own doc comment already establishes the body is merkle-bound on *both* shapes at its call sites and that "there is no unbound shape left on this route for a corrupt verdict to apply to", so `BlockInvalid` is already justified there and the ordering argument survives intact. It also removes the 13-line comment and the guard duplicated verbatim at two sites, which is the same drift argument that put `CheckCoinbaseOnlyBodyBound` in `model` in the first place.

```go
func (u *BlockValidation) checkQuickValidationCoinbaseLength(block *model.Block, caller string) error {
	if !block.CoinbaseTx.IsCoinbase() {
		return errors.NewBlockInvalidError("[%s][%s] block coinbase tx is not a valid coinbase tx", caller, block.Hash().String())
	}

	if !model.CoinbaseScriptSigLengthInBounds(block.CoinbaseTx, u.settings.ChainCfgParams) {
		return errors.NewBlockInvalidError("[%s][%s] bad coinbase length", caller, block.Hash().String())
	}

	return nil
}
```

The trade-off the reviewer names is real — on the no-subtree shape the check would then run after `AssignBlockID` instead of before it, and on the subtree shape after the UTXO writes. If you would rather not give that up, keep the early guard *and* add the unconditional one; either way the last two lines of both comments have to go, since no arrangement makes "checked as part of the subtree body" true. Whichever you pick, `quick_validate_coinbase_shape_test.go` needs a third case built on a subtrees-present fixture with the header root set via `RootHashWithReplaceRootNode(spend.TxIDChainHash(), 0, 0)`, otherwise the shape that is currently uncovered stays uncovered by the suite as well.

## Non-blocking issues

### ChiR2 — `IsCoinbase()` accepts prevouts svnode rejects

*Same as [@freemans13's point "A pre-existing issue, not this PR's to fix." (14-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1738#pullrequestreview-5199505141).*

**Problem:** I confirmed this in the pinned dependency. `go-bt/v2 v2.6.9` `tx.go:283` requires a single input with a zero previous-txid and then returns true if `PreviousTxOutIndex == DefaultSequenceNumber` **or** `SequenceNumber == DefaultSequenceNumber`. svnode's `IsCoinBase` wants the null prevout — zero hash **and** index `0xFFFFFFFF`. So a single-input transaction spending `000…000:0` with a final sequence satisfies the new guard here and `Valid`'s step 4b, where svnode would reject the block. This is pre-existing and reaches further than this diff; raising it only so the new check is not read as svnode parity.

**Why it matters:** The second change's whole value is the strength of `IsCoinbase()`, so it is worth recording where that predicate is looser than the reference implementation rather than discovering it later from a chain split. Not this PR's to fix, and I would not hold the merge on it.

**Fix:** Nothing here. Worth a separate issue against the shared predicate, and if you touch the new comment for ChiR1 anyway, one clause noting the check is "coinbase-shaped per go-bt, which is looser than svnode's null-prevout rule" would stop the next reader assuming parity.

### ChiR3 — The fee-skip comment says "FOUR conditions" but still lists three

**Problem:** `model/Block.go:1221-1224` now reads "only when ALL FOUR conditions hold: the block is at/below the highest HARDCODED checkpoint, the store can actually produce the fee=0 subtrees this skip exists to tolerate, and the block is a confirmed ancestor of the pinned checkpoint." Only the count was updated; the inline enumeration still names three. The numbered list below it does carry the new item 4, so the information is present, just not where the sentence promises it.

**Why it matters:** This comment is the load-bearing explanation for a security-relevant skip, and it is the one a future reader will check the code against. A reader who counts the sentence and gets three has to decide which of the two is stale.

**Fix:** Extend the sentence to match:

```go
	// Skip the coinbase no-inflation check (coinbaseOutput <= subsidy + fees) only when ALL
	// FOUR conditions hold: the block is at/below the highest HARDCODED checkpoint, the store
	// can actually produce the fee=0 subtrees this skip exists to tolerate, the block is a
	// confirmed ancestor of the pinned checkpoint, and the body has been bound to the header.
	// Each condition answers a distinct concern:
```

### ChiR4 — Test doc comment sits on the wrong declaration

*Same as [@freemans13's point "Nit: doc comment on the wrong declaration" (14-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1738#discussion_r4006659924).*

**Problem:** Confirmed in the new file. The block starting "CheckCoinbaseOnlyBodyBound ties a no-subtree body…" (lines 15-23) is followed by a blank line, so it attaches to nothing; the block starting "TestQuickValidateBlock_BoundNonCoinbaseBodyRejected builds a coinbase-only block…" names the test but sits directly on `func buildBoundNonCoinbaseBody`.

**Why it matters:** godoc renders the second block as the helper's documentation, and the first is dropped. Purely cosmetic, but the file's reasoning is the useful part of it and it is currently misfiled.

**Fix:** Keep the rationale as a package-level or file-header comment, and give each function the doc comment that names it:

```go
// buildBoundNonCoinbaseBody builds a coinbase-only block whose only transaction is a
// normal spend, with the header merkle root set to that txid so the body is genuinely bound.
func buildBoundNonCoinbaseBody(t *testing.T) *model.Block {
```

## Recap

| --- | --- | --- | --- |
| ChiR1 | Shape check skips subtree blocks | YES | 70% |
| ChiR2 | IsCoinbase weaker than svnode | | 20% |
| ChiR3 | Comment enumerates three of four | | 12% |
| ChiR4 | Doc comment on wrong declaration | | 8% |

model/Block.go
```diff
@@ -1057,6 +1057,31 @@ func bindClassifiedError(merkleRootChecked bool, format string, args ...interfac
...
+// IsConsensusCoinbase reports whether tx has the coinbase shape consensus requires:
+// exactly one input whose previous outpoint is NULL — a zero hash AND an index of
+// 0xFFFFFFFF. This is svnode's CTransaction::IsCoinBase (vin.size() == 1 &&
+// vin[0].prevout.IsNull(), where COutPoint::IsNull() is hash.IsNull() && n == uint32(-1)).
+//
+// Deliberately NOT go-bt's Tx.IsCoinbase, which accepts a null hash plus EITHER a
+// 0xFFFFFFFF prevout index OR a 0xFFFFFFFF sequence number. The sequence number says
+// nothing about coinbase-ness, so that disjunction admits a transaction svnode rejects:
+// prevout (0x00..00, 0) with sequence 0xFFFFFFFF. Anywhere the answer decides a consensus
+// verdict, the looser predicate is a chain-split shape, so use this one.
```
freemans13, line 1069: `Block.Valid` doesn't follow this yet. Step 4a still calls go-bt's `IsCoinbase()` at line 922, and that's where the verdict gets decided. Catch-up sends every non-corrupt quick-validation failure back to `Valid`, and above the checkpoint `Valid` is the only check that runs. A throwaway test with the fixture from `TestBlockValid_InvalidWhenBoundCoinbaseIsNotACoinbase` and a bound coinbase spending `000…000:0` with a final sequence gets `valid=true` at this commit, and `BLOCK_INVALID` once line 922 calls this function. Numbers and trade-offs are in the review body.

services/blockvalidation/quick_validate_p0_test.go
```diff
@@ -0,0 +1,121 @@
...
+// correct class: the header commits to this transaction.
+func TestValidateSubtrees_NonCoinbaseBodyRejected(t *testing.T) {
+	suite := NewCatchupTestSuite(t)
+	defer suite.Cleanup()
+
+	spend := bt.NewTx()
+	require.NoError(t, spend.From("6a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b", 0, "76a914000000000000000000000000000000000000000088ac", 1000))
+	require.NoError(t, spend.AddP2PKHOutputFromAddress("1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa", 900))
+	spend.Inputs[0].UnlockingScript = bscript.NewFromBytes(make([]byte, 16))
+	require.False(t, spend.IsCoinbase())
```
freemans13, line 75: Both route tests use an ordinary spend, which go-bt's `IsCoinbase()` rejects too. So nothing checks that the quick route calls the strict predicate. When I swapped all three `IsConsensusCoinbase` calls in `quick_validate.go` back to `block.CoinbaseTx.IsCoinbase()`, the package still passed. Building this fixture, or the no-subtree one, from the prevout `000…000:0` / sequence `0xFFFFFFFF` transaction in `TestIsConsensusCoinbase_RejectsNonNullPrevoutIndex` would catch that.

services/blockvalidation/quick_validate_p0_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+// TestValidateSubtrees_DuplicateTransactionRejected — CVE-2012-2459.
+//
+// The merkle root CANNOT detect a duplicated trailing transaction: the
+// duplicate-last-node-when-odd rule makes the mutated body produce the SAME root,
+// and therefore the same block hash, as the honest block. So a body that binds to
+// a checkpoint-certified header can still carry a repeated transaction, reusing the
+// honest block's proof of work. model.CheckSubtreeSlicesForDuplicateTxs exists
+// precisely for slice-holding paths that skip Valid's pooled check, and documents
+// itself as mandatory before UTXOs are created — but nothing called it.
```
freemans13, line 29: Nit: it did have a production caller. netsync's `prepareSubtrees` runs it at `services/legacy/netsync/handle_block.go:519`, before any UTXO write. What was missing was a call on the quick route. The commit message says the same thing, but only this comment can still be corrected.

freemans13 (changes requested):
Still requesting changes, but for a different reason than last time. My earlier blocker is fixed: bodies with subtrees now get the coinbase check in `validateSubtrees`, and I confirmed a test catches its removal. The new strict coinbase test is right, but it only runs on the quick route, and on native catch-up the looser go-bt test overrules it. The fix is one line in `Block.Valid`.

I read commit a5e23c30f, which is also the head the PR shows.

**The strict coinbase test loses to the loose one wherever the verdict is decided.** `IsConsensusCoinbase` requires a zero prevout hash and index `0xFFFFFFFF`, which matches svnode. Its doc comment says to use it "anywhere the answer decides a consensus verdict". `Block.Valid` still uses go-bt's `IsCoinbase()` at step 4a (`model/Block.go:922`). That matters twice.

First, on catch-up. When quick validation returns anything other than corrupt or incomplete, `tryQuickValidation` logs it as a local failure and returns to full validation (`catchup.go:1966-1987`). The new `BlockInvalid` from `validateSubtrees` goes that way, and so does the one at the no-subtree entry point. `Valid` then decides the block using the loose test. So on `quickValidateBlockAsync`, the only entry point catch-up uses, the new predicate changes nothing for the prevout `000…000:0` shape it was written for. It only settles a verdict on the legacy unified route, which is off by default.

Second, above the checkpoint. There the header is not pinned to a known hash, so a miner picks the coinbase. `Valid` is the only check that runs, and it accepts the shape svnode rejects. That's the chain-split case the commit message describes, and it's still open.

To confirm the first part I wrote a throwaway test in a scratch worktree, pushed nowhere. It clones the fixture from `TestBlockValid_InvalidWhenBoundCoinbaseIsNotACoinbase`, with the coinbase swapped for the transaction from your `TestIsConsensusCoinbase_RejectsNonNullPrevoutIndex`: prevout `000…000:0`, sequence `0xFFFFFFFF`, plus one output. The header root is that txid, so the body is bound.

```
at a5e23c30f:                         valid=true  err=<nil>
with IsConsensusCoinbase at step 4a:  valid=false err=BLOCK_INVALID (11): ... block coinbase tx is not a valid coinbase tx
```

The change is `if !IsConsensusCoinbase(b.CoinbaseTx) {` at `model/Block.go:922`. It keeps the one-input guarantee that `CoinbaseScriptSigLengthInBounds` relies on just below it. It shouldn't reject teranode's own blocks either, because `makeCoinbase1` writes a zero hash and index `0xFFFFFFFF` (`model/GetCoinbaseParts.go:197-198`). With that one line applied, `model`, `services/blockvalidation` (+ `catchup`), `services/blockassembly` (+ `mining`, `subtreeprocessor`) and `services/legacy/netsync` all pass under `-tags testtxmetacache`. The one exception is `TestReplayRecoveryAerospikeAssemblyEndToEnd`, which needs Docker for its Aerospike container, and it fails the same way on the untouched head. If you'd rather keep this PR to the quick route, the other honest option is to narrow the doc comment and the PR description, and open an issue for `Valid`. I'd prefer the one-liner, since the predicate and its fixture already exist.

**Nothing drives the strict test through the quick route.** I broke each of the three new checks separately and re-ran the quick-validation tests:
- duplicate scan removed from `validateSubtrees`: `TestValidateSubtrees_DuplicateTransactionRejected` fails
- subtree coinbase check removed: `TestValidateSubtrees_NonCoinbaseBodyRejected` fails
- all three `IsConsensusCoinbase` calls in `quick_validate.go` swapped back to `block.CoinbaseTx.IsCoinbase()`: the package passes

The third survives because both route tests use an ordinary spend, which go-bt's test and the strict one both reject. Only the unit test on the function sees the difference. Details inline.

**The rest holds up.**
- `IsConsensusCoinbase` matches svnode's `COutPoint::IsNull`. go-bt keeps the previous txid as a fixed 32-byte array, so the byte comparison can't be fooled by a short slice (`input.go:156` in v2.6.9).
- The duplicate scan returns corrupt, which both callers already pass through unwrapped.
- Commit 1c86148df's correction is accurate. The order-and-blessed skip needs `OutpointOnlyBelowCheckpoint`, which the fee skip doesn't, so the two skips no longer claim to match (`model/checkpoint.go:55`). The "four conditions" sentence now lists four.

The doc-comment nit on `buildBoundNonCoinbaseBody` from my last review is still there. Not repeating it inline.

What I didn't check: whether a real duplicated body gets through `processBlockSubtrees`' UTXO batches to reach the new scan. The test calls `validateSubtrees` directly. If the batches fail first, catch-up falls back to `Valid`, which runs its own duplicate check, so the outcome is still safe. I also didn't answer the bot's rollback question, and didn't run anything against a real node.

gokutheengineer (approved):
Minor

1. quick_validate.go:990 runs the dedup scan after every createAndSpendUTXOsForBatch batch has already committed, while model/check_duplicate_txs.go:12 says callers "MUST call it before creating or spending UTXOs". The description defers the restructure to bitcoin-sv/teranode#4836, which is fine, but the helper's contract now has a production caller that violates it. Either soften the wording to "before the block is committed" or reference 4836 there so the next reader does not treat it as a bug.
2. The submitMiningSolution guard at Server.go:1848 has no test, despite "every guard here has a fixture that fails when the guard is reverted". A loose-shape coinbase submission asserting the ProcessingError (and no assembly reset) would close that.

ctnguyen (approved):
Review at commit [852902e](https://github.com/icellan/teranode/commit/852902e774e8fe8472be7fde94408438427a0513)

This grew a lot since I last looked at it, and in the right direction. The blocker is genuinely closed rather than relocated: the subtree-carrying shape now gets its own check at `services/blockvalidation/quick_validate.go:1002`, behind `CheckMerkleRoot`, and the comment at `:278-281` finally describes what the code does. `IsConsensusCoinbase` (`model/Block.go:1078`) is the nicer surprise — it is derived correctly from `COutPoint::IsNull` rather than approximated, and applying it at `Block.Valid` step 4a (`:930`) is the right call, because that is where the verdict actually lands on catch-up fallback and above the checkpoint. I went looking for the opposite failure — a strict predicate rejecting something honest — and did not find one: `IsConsensusCoinbase` reuses the same zero-hash test go-bt already applies, so the only restriction it adds over that test is the prevout index, and `makeCoinbase1` (`model/GetCoinbaseParts.go:196-198`) writes `0xFFFFFFFF` there, which keeps Teranode's own coinbases clear of it. The classification work is also subtler than it looks and it is right: the placeholder-substitution case at `:985` really does preserve the honest block hash (because `RootHashWithReplaceRootNode` overwrites node 0 either way), so corrupt rather than invalid is the correct verdict there, and the CVE-2012-2459 scan has the same property. The loose-shape fixtures added in response to the "nothing drives the strict test" point are the right answer to it — they are the tests that actually fail if someone swaps the predicate back. What is left is smaller: one routing change that nothing consumes, one guard on the wrong branch, and some comments that the last few commits outran.

## Status of previous points

ChiR2 is the only one that is not clean. The PR does much more than the previous review asked — it introduces a strict predicate and applies it at every site that decides a verdict (`model/Block.go:930`, `quick_validate.go:286`, `:408`, `:1002`, `services/blockassembly/Server.go:1849`). One consumer of go-bt's loose predicate remains where the answer still has an effect: `services/blockvalidation/skip_unspendable.go:23` passes `tx.IsCoinbase()` into `utxo.HasNoSpendableOutputs` to decide whether a transaction's outputs are written to the UTXO store at all, on the quick-validation catchup path. It is narrow (gated on `SkipUnspendableTxStorageDuringCatchup` plus the skip-lock mode) and it is adjacent to the `bad-cb-multiple` gap already parked on bitcoin-sv/teranode#4836, so I would not hold the merge on it — but it is why the row is not FIXED.

| --- | --- | --- | --- |
| ChiR1 | Shape check skips subtree blocks | YES | FIXED |
| ChiR2 | IsCoinbase weaker than svnode | | PARTIAL FIXED |
| ChiR3 | Comment enumerates three of four | | FIXED |
| ChiR4 | Doc comment on wrong declaration | | FIXED |

## Non-blocking issues

### ChiR5 — The unwrapped invalid verdict has no consumer, and catch-up still calls it a local fault

**Problem:** `quick_validate.go:309` and `:437` now let an `ErrBlockInvalid` from `validateSubtrees` through unwrapped, so a call site can route on the class instead of on check order. I traced both consumers and neither uses it. On catch-up, `tryQuickValidation` (`catchup.go:1867`) branches on exactly two things: `errors.Is(err, errors.ErrBlockIncomplete)` at `:1878` and `errors.IsBlockCorrupt(err)` at `:1900`. Everything else — the new invalid verdict included — falls to `:1967-1992`, which logs "quick validation failed for block %s, removing .subtree files" and whose comment at `:1970-1972` reads "failed for a LOCAL reason (not corrupt, not incomplete — e.g. a UTXO-store hiccup or block-ID assignment failure), the peer-supplied body itself is not implicated". For a bound body whose coinbase is not consensus-shaped, the body is precisely what is implicated. The consequence is concrete rather than cosmetic: that branch deliberately passes only `freshlyWritten` to `removeCatchupSubtreeFiles` (`:1982`) and does not merge `fetchFreshlyWritten`, so the fetched `.subtree`/`.subtreeData` blobs of a body just condemned as invalid are kept for reuse, no peer is struck, and the block is re-run through full normal validation. It does converge — step 4a now applies the same predicate, so `Block.Valid` reaches the same verdict — but at the cost of a whole revalidation and a diagnostic that says the opposite of what happened. On the legacy unified route the unwrap changes nothing at all: `processBlockFound` returns `qErr` verbatim (`services/blockvalidation/Server.go:1805-1808`), and `shouldDisconnectOnBlockErr` (`services/legacy/peer_server.go:1439-1465`) ends at `!errors.IsTransientLocalError(err)`, which covers only the service/storage classes (`errors/error_utils.go:71-79`) — so a `ProcessingError`-wrapped verdict already disconnected the peer and an unwrapped one still does. To be fair to the change: that bucket already received an invalid verdict from `checkQuickValidationCoinbaseLength` (`quick_validate.go:483`), so the mismatch is not newly created here. But this PR adds three more sources of it and specifically reworks the wrapping so the class survives to the caller, which makes this the moment to give it a branch.

**Why it matters:** The stated point of commit 846043878 is that routing should not depend on which predicate a call site tests first. Right now the class survives the trip and then the only call site that could act on it discards the distinction, so the hardening buys nothing yet while the log line actively misdirects an operator debugging a stalled sync. Retaining the fetched blobs for a body that failed a post-binding consensus check is also the reuse pattern that the corrupt branch at `:1919-1943` exists to prevent.

**Fix:** Give the verdict its own branch after the corrupt one, even if it still falls through to normal validation — the point is that the fall-through becomes deliberate and correctly described:

```go
	if errors.Is(err, errors.ErrBlockInvalid) {
		// Body-derived and bound. Normal validation reaches the same verdict through
		// Block.Valid step 4a and persists it, so fall through rather than abort —
		// but this is not a local fault and must not be logged as one.
		u.logger.Warnf("[catchup:tryQuickValidation][%s] block %s from peer %s failed a bound-body consensus check on the quick route, falling through to full validation: %v",
			catchupCtx.blockUpTo.Hash().String(), block.Hash().String(), peerID, err)
	}
```

At minimum, reword `catchup.go:1967-1972` so it no longer claims the body is not implicated.

### ChiR6 — The submission guard sits on one branch of a function that documents the other placement

*Same as [@gokutheengineer's point 2 (15-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1738#pullrequestreview-5210313995).*

**Problem:** The new guard is at `services/blockassembly/Server.go:1849`, inside `if req.CoinbaseTx != nil`. Twenty-five lines below, `validateCoinbaseForSubmission` (`:1705`, called at `:1874`) is the shared convergence point for both branches, and its own comment at `:1719-1722` says the P2SH guard is "placed at the shared submission convergence deliberately, not only on the pool branch" so that `GenerateBlocks` stays in parity. The new shape guard does the thing that comment warns against. I checked whether the `else` branch can currently produce a shape the predicate would reject, and it cannot: `CreateCoinbaseTxCandidate` (`model/mining_candidate.go:16`) reaches `makeCoinbase1`, which writes a zero prev hash and index `0xFFFFFFFF` (`model/GetCoinbaseParts.go:196-198`), and the only input field the branch then rewrites is `SequenceNumber` (`Server.go:1869`), which the predicate ignores. So this is a maintenance hazard rather than a live gap. Extending @gokutheengineer's point on the missing test: I looked for one and there is none — no test under `services/blockassembly/` names `IsConsensusCoinbase` or the guard's message, and the nearest fixtures (`Server_security_test.go:35-88`) cover input count only, so reverting `:1849` leaves the package green.

**Why it matters:** The value of the guard is that a malformed miner submission is charged to the submitter instead of triggering `ba.blockAssembler.Reset(false)` at `:1988`. A guard with no fixture is one refactor away from being dropped silently, and a guard on one branch of a two-branch function is one change to `CreateCoinbaseTxCandidate` away from being bypassed.

**Fix:** Move it into the shared convergence point, where it also picks up the existing nil and input-count guards for free:

```go
	if !model.IsConsensusCoinbase(coinbaseTx) {
		return errors.NewProcessingError("[BlockAssembly][%s] submitted coinbase transaction is not a valid coinbase", jobID)
	}
```

and add a `submitMiningSolution` case that submits the loose shape (null prevout hash, index 0, sequence `0xFFFFFFFF`), asserts the `ProcessingError`, and asserts no assembly reset.

### ChiR7 — The dedup helper now has a production caller that contradicts its contract

*Same as [@gokutheengineer's point 1 (15-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1738#pullrequestreview-5210313995).*

**Problem:** `model/check_duplicate_txs.go:13` says every slice-holding path "MUST call it before creating or spending UTXOs". The new caller at `quick_validate.go:990` runs inside `validateSubtrees`, which is reached only from the tails of `processBlockSubtreesSequential` (`:629`), `processBlockSubtreesPipeline` (`:771`) and `processBlockSubtreesPipelineAsync` (`:921`) — in every case after `createAndSpendUTXOsForBatch` has committed the last batch. The other production caller, netsync's `prepareSubtrees` (`services/legacy/netsync/handle_block.go:519`), does honour the ordering and says so at `:516-518`. I am not re-raising the restructure: bitcoin-sv/teranode#4836 parks it explicitly and the author has recorded there that the block is still rejected before commit, so the verdict is not affected. The narrow problem is that a doc marked MUST now has a caller that does not, with nothing at either end pointing at the other.

**Why it matters:** This helper's doc is unusually load-bearing — it is the argument for why the scan cannot be skipped, and the next person to add a slice-holding path will read it and reasonably assume the two existing callers are the pattern. One of them is not.

**Fix:** Reconcile the wording where it lives, and point at the issue rather than silently weakening the rule:

```go
// model/check_duplicate_txs.go:13 — replace the MUST sentence with:
// Every such path MUST call it before the block is committed, and before creating
// or spending UTXOs wherever the route allows it. The quick-validation route
// (services/blockvalidation/quick_validate.go) currently calls it after its UTXO
// batches because the body checks sit in validateSubtrees; moving them ahead of the
// batches is tracked on bitcoin-sv/teranode#4836.
```

### ChiR8 — The new nil guard sits below the dereference it is meant to precede

**Problem:** `validateSubtrees` opens with the size loop at `quick_validate.go:930-933`, which calls `block.SubtreeSlices[i].Length()` with no nil check — and for a block carrying two or more subtrees that runs with `i == 0`. The PR's new guard for exactly that slot is 45 lines below at `:975-979`, where a nil `block.SubtreeSlices[0]` returns a `ProcessingError`. go-subtree's `Length()` is `st.mu.RLock(); length := len(st.Nodes); st.mu.RUnlock(); return length` at the v1.5.1 tag `go.mod:17` pins, so a nil receiver panics on the `RLock` rather than returning 0. The guard therefore cannot fire for any multi-subtree block — the loop reaches slot 0 first. I am not claiming a live panic, because I looked for a quick-route writer that could nil a slot while `validateSubtrees` runs and found none: `releaseBlockNodes` (`services/blockvalidation/pools.go:141`) is called only from `BlockValidation.go` (`:2070`, `:2152`, `:2180`, `:2220`, `:2305`, `:2332`, `:2349`, `:2401`), all on the `Block.Valid` route, and `tryReleaseBlockNodes` (`pools.go:155`) is reached only through `evictLastValidatedBlock` (`pools.go:181-182`, wired at `BlockValidation.go:435`), while `lastValidatedBlocks` is never touched from `quick_validate.go` — its only `Set` calls are `BlockValidation.go:2170` and `:2389`, both after full validation. So either the guard defends against something this route cannot currently produce, or a release path reaches here later; in the second case the loop faults before the guard gets a say.

**Why it matters:** A guard placed where it cannot do its job reads as coverage that is not there. The only shape it can actually cover is the single-subtree block, where `len(SubtreeSlices)-1 == 0` and the loop never executes — which is the shape that needed it least. Whichever way the reachability question resolves, the ordering is inverted today and costs three lines to put right while the code is open.

**Fix:** Check the slot at its first touch, matching the wording `model` already uses for the same condition (`model/Block.go:1324-1326`):

```go
	for i := 0; i < len(block.SubtreeSlices)-1; i++ {
		st := block.SubtreeSlices[i]
		if st == nil {
			return 0, errors.NewProcessingError("[validateSubtrees][%s] subtree %d of %d was released during validation", block.Hash().String(), i, len(block.SubtreeSlices))
		}

		if i == 0 {
			subtreeSize = st.Length()
		} else if st.Length() != subtreeSize {
			...
		}
	}
```

### ChiR9 — The new test file still says the helper had no caller

*Same as [@freemans13's comment (15-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1738#discussion_r4014190017).*

**Problem:** `services/blockvalidation/quick_validate_p0_test.go:34` still ends "documents itself as mandatory before UTXOs are created — but nothing called it." It was called: `services/legacy/netsync/handle_block.go:519`, in `prepareSubtrees`, and the comment directly above it at `:516-518` describes it as running "before any UTXO or subtree write". The production comment at `quick_validate.go:958-965` was corrected in the same commit run and no longer makes this claim, so this is the one copy left behind. The accurate statement is the narrower one the PR body elsewhere gets right: the quick-validation route had no call, while netsync did.

**Why it matters:** It is the rationale block for the CVE-2012-2459 fixture, so it is the text someone will read when deciding whether the scan is still needed on a fourth route — and it currently overstates the gap by implying the floor was unenforced everywhere.

**Fix:** Narrow the claim to the route that actually lacked the call:

```go
// services/blockvalidation/quick_validate_p0_test.go:34 — replace the last clause with:
// itself as mandatory before UTXOs are created. Legacy netsync calls it
// (services/legacy/netsync/handle_block.go), but the quick-validation route did not.
```

### ChiR10 — CoinbaseScriptSigLengthInBounds still documents IsCoinbase as its guarantee

**Problem:** `model/Block.go:1096` reads "The caller must guarantee coinbaseTx has at least one input before calling (Valid's step 4b above relies on IsCoinbase() for that; the quick-validation path's own nil/empty check gives the same guarantee)". Step 4a no longer calls `IsCoinbase()` — this PR replaced it with `IsConsensusCoinbase` at `:930` and updated the matching note at `:921-922` to say so. The doc on the function that depends on that guarantee was not updated with it.

**Why it matters:** Same class as the point the previous round raised about the fee-skip comment, and same reason: this sentence is the stated precondition for an unconditional `Inputs[0]` index, so the predicate it names should be the one actually providing it. `IsConsensusCoinbase` requires exactly one input, so the guarantee is stronger than before — worth saying, not worth leaving stale.

**Fix:** Name the predicate that now supplies the guarantee, and the step it lives at:

```go
// model/Block.go:1096 — replace the parenthetical with:
// at least one input before calling (Valid's step 4a above relies on IsConsensusCoinbase
// for that, which requires exactly one input; the quick-validation path's own nil/empty
// check gives the same guarantee) — Inputs[0] is indexed unconditionally.
```

## Recap

| --- | --- | --- | --- |
| ChiR5 | Invalid verdict still bucketed local | | 32% |
| ChiR6 | Submission guard one branch untested | | 25% |
| ChiR7 | Dedup contract versus new caller | | 14% |
| ChiR8 | Nil guard below the dereference | | 12% |
| ChiR9 | Stale no-caller claim in test | | 10% |
| ChiR10 | Stale IsCoinbase doc reference | | 6% |

freemans13 (approved):
Approving. My blocker from the last round is fixed. `Block.Valid` now uses `IsConsensusCoinbase` at step 4a, so the strict coinbase test decides the verdict above the checkpoint and on the catch-up fallback. The new route tests show that the quick route uses it too. The red `test` job isn't from this PR: both failing tests fail the same way on `main`.

I read commit 852902e77, which is also the head the PR shows.

**Every fix I asked about now has a test that fails without it.** I undid each change separately in a scratch worktree, pushed nowhere, and re-ran:
- step 4a back to go-bt's `IsCoinbase()`: `TestBlockValid_InvalidWhenCoinbaseIsOnlyLooselyACoinbase` fails
- all three quick-route calls back to go-bt: `TestQuickValidateBlock_LooseCoinbaseRejectedOnRoute`, `TestValidateSubtrees_LooseCoinbaseRejected` and `TestQuickValidateBlock_SubtreeInvalidVerdictNotShadowed` fail
- only the `validateSubtrees` call back to go-bt: the last two of those fail
- placeholder check removed: `TestValidateSubtrees_FirstNodeMustBeCoinbasePlaceholder` fails
- invalid verdict wrapped in `ErrProcessing` again at both entry points: `TestQuickValidateBlock_SubtreeInvalidVerdictNotShadowed` fails

**Two of those changes still pass the suite when undone.** Neither is worth holding the merge for.
- Rewrapping the invalid verdict at `quickValidateBlockAsync` alone leaves the package green, because the not-shadowed test only drives the synchronous entry point. It's the same trap the PR description warns about. Right now it costs nothing, since, as Chi traced in ChiR5, catch-up doesn't branch on the invalid class yet. It will matter once catch-up does.
- Removing the `submitMiningSolution` guard leaves `services/blockassembly` green. That's gokutheengineer's point 2 and Chi's ChiR6. I'm only confirming it by mutation.

**The new placeholder check can't reject a block that full validation accepts.** I went looking for an honest writer that stores a first subtree with the real coinbase txid in slot 0, since reusing such a file would get re-downloaded forever as corrupt. There isn't one. Block assembly swaps the txid into a `Duplicate()`, and the coinbase proof code swaps it into a freshly parsed copy. Neither writes to the stored subtree (`services/blockassembly/Server.go:1897`, `:2161`, `services/blockvalidation/BlockValidation.go:3308-3310`). Netsync adds the placeholder with `AddCoinbaseNode` (`services/legacy/netsync/handle_block.go:462`). `Block.Valid` step 7 already enforces the same rule on the same files.

**The CI failures come from main.** `TestValidateBlock_CachedHeadersShortWindowIsRefused` and `TestValidateBlock_UnusableHeaderRunIsRebuiltByHashWalk` both fail with `cachedSubtreeConflictingNodes ... unable to read number of leaves: EOF`, which comes from subtree validation code this PR doesn't touch. I reproduced both 3 out of 3 on the PR head and 3 out of 3 on `upstream/main` at 4241a96c7. The Sonar gate is red only because it skips when tests fail.

The test comment saying the duplicate scan had no caller is still at `quick_validate_p0_test.go:34`. Chi's ChiR9 has the wording, so I won't repeat it.

What I didn't check: the full `services/blockvalidation` and `model` suites at this head, beyond the tests named above and the two CI failures. I also didn't check which main commit broke those two tests, or run anything against a real node.

ordishs (approved):
Verified the load-bearing claims against the tree:

- go-bt v2.7.1 `Tx.IsCoinbase` (tx.go:283-299) is the disjunction described — `PreviousTxOutIndex == 0xFFFFFFFF || SequenceNumber == 0xFFFFFFFF`. `IsConsensusCoinbase` is strictly stricter, so it can never newly accept a shape.
- `makeCoinbase1` (model/GetCoinbaseParts.go:196-197) writes a zero hash and index `0xFFFFFFFF`. Teranode's own coinbases are unaffected.
- All four production `block.Valid` call sites pass a non-nil subtree store (blockvalidation/BlockValidation.go:2212, :3167, blockvalidation/Server.go:1553, blockassembly/Server.go:1984), so `merkleRootChecked` is true on every live path and the new fee-skip conjunct changes no production behaviour.
- `CheckMerkleRoot` (model/Block.go:2442) substitutes node 0 unconditionally, without testing that it held the placeholder — the new first-node guard is load-bearing, not redundant.
- All three quick-route pipelines (quick_validate.go:629, :771, :921) funnel through `validateSubtrees`, so the new guards cover sequential, pipelined and async modes.
- `CheckSubtreeSlicesForDuplicateTxs` and `checkDuplicateTransactionsInSubtree` skip slot `[0][0]` on the same placeholder condition, so the two dedup implementations stay at parity.

No P0 or P1. Three smaller items, none of which changes what the route accepts:

**services/blockassembly/Server.go:1849** — the new guard has no test. No file under `services/blockassembly/` contains "not a valid coinbase" other than the guard, and the diff adds no blockassembly test. Delete the four lines and the suite stays green — the exact mutation the description says every guard here was proven against. Suggest a `submit_mining_solution_test.go` case for the loose shape (null prevout hash, index 0, sequence `0xFFFFFFFF`), plus the `makeCoinbase1` shape as the no-regression mirror.

**services/blockvalidation/quick_validate.go:976-979** — the new `if first == nil` guard cannot fire on a block with two or more subtrees. The size loop at line 932 already dereferences `block.SubtreeSlices[0].Length()` with no nil check, above it. A concurrent `ReleaseSubtreeNodes` panics at 932 instead of returning the retryable `ProcessingError` the guard exists to produce. `Block.Valid` gets this right (model/Block.go:818-822): it captures the entry once and nil-checks before any use. Suggest hoisting the capture above the loop.

**services/blockvalidation/catchup.go:1967-1991** — `ErrBlockInvalid` lands in the fall-through branch, whose comment reads "the quick validation failed for a LOCAL reason (not corrupt, not incomplete — e.g. a UTXO-store hiccup)". Not a regression: the wrapped `ErrProcessing` took the same branch before. But this PR is what makes `ErrBlockInvalid` reachable here as a distinct class, and at that point the comment is wrong — a bound body with a `(0x00…00, 0)` / sequence `0xFFFFFFFF` coinbase is logged as a local fault and then pays for a full re-validation of a body already proven invalid. Suggest an `errors.Is(err, errors.ErrBlockInvalid)` branch returning `false, err`, or at minimum a corrected comment.

CI note: `test` is red on `TestValidateBlock_CachedHeadersShortWindowIsRefused` and `TestValidateBlock_UnusableHeaderRunIsRebuiltByHashWalk`. Neither file is in this diff, and both fail identically on #1751, which touches none of this code — pre-existing on main, not this PR.

PR #1739, opened 2026-09-14
fix(dashboard): correct subtree numbering

oskarszoon:
Subtree headings use Merkle-tree height where users expect the subtree's position within a block. Carry the existing block-table index through both navigation links and use it in the details and transaction headings.

Preserve the table's zero-based numbering, including index zero. Links without valid block/index context show unnumbered headings. Headings react to same-page navigation and browser history. Merkle-height metadata and backend responses remain unchanged.

Validation: 78 unit tests and 24 browser tests pass, including the new regression coverage. Type-checking reports zero errors and the same 14 existing warnings; the production build, changed-file ESLint, and scoped formatting checks pass. An additional mocked page-two navigation check confirmed that both link types retain the API indices.

Repository-wide lint remains blocked by the same 53 pre-existing source-formatting findings, plus two generated test-report files after browser tests. Running full ESLint separately reports 118 errors in 41 unchanged files; none are in the changed files. No unrelated formatting sweep or configuration change is included.


freemans13 (approved):
Approving. The fix is right, and the tests pin it down: the fixture's Merkle height is now 4, so the old heading would render "#4" where the tests expect #0, #1 and #7. The one test that would have passed under the old code can't. I checked that `index` in `GetBlockSubtrees.go` is the absolute position (`i` runs from `offset`, not from zero on each page), so page two carries the right numbers. I also checked that `valueSet(0)` still renders the index-0 link. CI's dashboard `smoke` job ran on 3a34fae: 78 vitest and 24 Playwright tests passed, `urls.test.ts` among them.

One follow-up, not blocking, and outside the diff, so it isn't an inline comment:

The transaction page already knows the subtree's position and throws it away. `tx-details-card/index.svelte:164` renders the link text as `Subtree #{subtreeIdx}` but builds the URL by hand with only `hash` and `blockHash`. After this PR, a user who clicks "Subtree #3" lands on a page titled "Subtree Details" with no number. `subtreeIdxs` is the same zero-based position within the block (`stores/utxo/meta/data.go:45`), so switching that link to `getDetailsUrl(DetailType.subtree, data.subtreeHashes[i], { blockHash: data.blockHashes[i], index: subtreeIdx })` would close the gap. The block JSON view's `subtrees` array links in `json-tree/index.svelte:235` could pass the array position the same way, though that matters less.

What I did not do: run the suites locally (no `node_modules` in my checkout), or click through a real node's dashboard. The evidence above is CI's run plus reading the code.

icellan (approved):

PR #1740, opened 2026-09-14
fix: tighten recovery history coverage and resume

oskarszoon:
Recovery can classify an unconfirmed transaction without proving spend coverage for its confirmed parent, fail to resume an empty history database after interrupted initialization, and scan the accumulated transaction index for every block. Fix the three shared findings from the review of #1737 following the merge of #1715.

- Require parent target coverage before treating an empty canonical-spend lookup as evidence. Missing coverage leaves the transaction Unknown.
- Create the SQLite schema atomically and let the existing pre-apply resume path rebuild an empty, unclaimed database. Foreign metadata/schema and corrupt sealed artifacts remain refused; post-mutation resume cannot rebuild history.
- Add a block-leading transaction index so per-block retention checks use an indexed lookup.

The change is limited to the history implementation and one regression test file. The same changes are included in the v0.15 backport, #1737. Existing sealed histories remain compatible.

Validation: all three regressions reproduced before the fix and pass afterward. Recovery/core/CLI race suites, affected vet, CLI build, recovery lint/staticcheck and recovery gosec pass on both variants. The tests include a covered-parent control, cancelled initialization, foreign-artifact refusal and a SQLite query-plan assertion.

The global ancestry expansion budget and other deferred work remain tracked in #1736. Mainnet-scale runtime remains unmeasured.

Refs #1708, #1736.

The cancellation regression covers interruption before SQLite initialization and recovery of the resulting empty database. It does not inject cancellation during schema creation or a process kill; transaction-rollback coverage is tracked in #1736.


icellan (approved):

cmd/recoverreplayedtransactions/history_review_test.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func TestHistoryResumeAfterCancelledInitialization(t *testing.T) {
+	tx := reviewChild(t, strings.Repeat("11", 32))
+	original, chain, archive, path, tip := reviewHistoryFixture(t, []*bt.Tx{tx}, tx.TxID(), HistoryOptions{})
+	opts := original.options
+	require.NoError(t, original.Close())
+	path += ".interrupted"
+	ctx, cancel := context.WithCancel(t.Context())
+	cancel()
```
freemans13, line 91: This test cancels before `NewHistory` touches SQLite at all. The PRAGMA exec returns `context.Canceled`, so the file stays zero bytes and the new transaction around the `CREATE` statements never runs.

I replaced the transaction with the old single `db.ExecContext` and all four new tests still pass. So the empty-file resume is pinned, but the guarantee in the comment at `history.go:132` (empty database or full schema, nothing in between) is not. Without the transaction, a cancel landing between two `CREATE` statements leaves a partial schema. `openHistory` then refuses it with "no such table: coverage", which is not `sql.ErrNoRows`, and the operator has to delete the file by hand.

Not blocking, since the transaction is plainly right. Worth either a test that reaches the DDL, or softening the "cancelled initialization" line in the description so it does not read as covering atomicity.

freemans13 (approved):
Approving. All three fixes are correct. I checked each one against the code paths around it, not only the diff.

- **Parent spend coverage.** The target set is fixed for the whole of the final pass: `expandUnconfirmed` only runs between passes, and each pass wipes `spends` before rescanning. So a parent that is in `targets` really does have every in-range spend indexed, and `Check` has already required a complete index from height 0 to the tip before it reaches this path. In a real run, a confirmed parent that is not a target comes from expansion stopping at its 10,000-edge budget. Before this change a child past that budget could be classified Unconfirmed with no spend evidence at all, so Unknown is the right answer.
- **Empty-database resume.** The `sql.ErrNoRows` wrap lands in the rebuild branch of `buildJobHistory`. The post-apply path at `run.go:349` calls `openHistory` directly, so an empty file there still fails closed, as the description says.
- **`transactions_by_block`.** Only a fresh build runs the retention count, and `historyDigest` reads table rows, not index objects, so existing seals still verify.

I ran the four new tests at 7ce2f1a44 with `-race`, and they pass. Then I broke each fix in turn in a scratch worktree. Removing the coverage check, the empty-schema branch or the index each makes its test fail. Removing the transaction around the DDL does not fail anything; details inline. That one is non-blocking.

Not checked: the rest of the package suite locally (CI is green, 37 checks), a process kill during schema creation as opposed to a context cancel, and anything at mainnet scale.

cmd/recoverreplayedtransactions/history_review_test.go
oskarszoon: Captured the missing mid-DDL interruption regression in #1736 for both variants, with the requirement that removing the schema transaction must make it fail. Clarified both PR descriptions: the current regression covers cancellation before SQLite initialization and empty-file recovery, not cancellation during DDL or a process kill. Deferring the additional test as nonblocking per this review; no production code change.

PR #1741, opened 2026-09-15
fix(blockvalidation): bound peer-supplied block and header HTTP reads

liam:
## Summary

Bounds the peer-supplied HTTP reads in block validation's block and header fetches (finding 1 of bitcoin-sv/teranode#4742). Finding 2 (unbounded error-body read in `buildHTTPError`) was already fixed in #1453.

**Streaming block parse.** `fetchSingleBlock` and `fetchBlocksBatch` stream the peer's response through `model.NewBlockFromReader` instead of `io.ReadAll`-ing it. Blocks have no consensus-defined maximum size, so there is no byte cap to put on the response as a whole. Streaming on its own only halves peak memory, though. It does not bound it, because the subtree hash list length is an unvalidated wire varint the parser honours with no ceiling. That is what the per-message cap below is for.

**Per-block-message byte cap.** Each block message is read through an `io.LimitedReader` sized by a new receive-side policy setting, `blockvalidation_max_incoming_block_message_bytes` (default 128 MiB, same shape as `subtreevalidation_max_incoming_subtree_bytes`). Going over the cap fails the fetch with `ErrExternal`, which is kept distinct from a genuine short stream. The cap bounds bytes delivered, which bounds the subtree hash list and so the parsed `*model.Block`. It does **not** bound the coinbase read: go-bt's `readArenaScript` allocates a script's declared length (up to 1 GiB) before reading any of it. That is a pre-existing dependency issue, documented on the setting and in the code, and filed as bsv-blockchain/go-bt#187.

**Batch stops at `n`.** `fetchBlocksBatch` stops at the requested block count instead of reading to EOF, so a peer that streams more well-formed blocks than requested cannot grow the result.

**Explicit fetch deadline.** `DoHTTPRequestBodyReader` falls back to `http_streaming_timeout` (600 s in settings.conf) when the context has no deadline, where the old `DoHTTPRequest` fell back to `http_timeout` (30 s). Two call sites had no deadline of their own: the batch fetch in catchup and the single-block fetch in `processBlockFound`. All four block fetch call sites now use a shared `peerBlockFetchTimeout` (30 s).

**Over-send diagnostic.** After the `n` blocks, a one-byte probe detects a peer that sent more than requested. It logs the event and records it through the diagnostic-only `reportCatchupError` path, and does not charge reputation, since padding on correct data is as likely to be a caching proxy or `?n=` version skew as it is malice. The probe has its own 2 s budget (`overSendProbeTimeout`), so a peer that sends `n` blocks and then holds the response open cannot spend the fetch deadline on it.

**Header cap.** `catchup.FetchHeadersWithRetry` caps its read via `util.DoHTTPRequestBounded`, since header responses have a caller-known exact size. The cap is `(maxBlockHeadersPerRequest+1) * BlockHeaderSize`; the `+1` covers the peer echoing back the common-ancestor header at each iteration boundary.

**Deprecation.** `util.DoHTTPRequest` (the unbounded `io.ReadAll` helper) has no production callers left and is marked `Deprecated` in favour of `DoHTTPRequestBounded` / `DoHTTPRequestBodyReader`.

## Test plan

- [x] `go build`, `go vet`, `golangci-lint` clean on the touched packages
- [x] Full `services/blockvalidation` (including `catchup`), `settings` and `util` suites pass (`-tags testtxmetacache`)
- Regression tests, each proven load-bearing by reverting its fix and confirming it fails:
  - [x] `TestFetchBlocksBatch_CurrentBehavior/Peer_Sends_More_Blocks_Than_Requested`
  - [x] `TestFetchHeadersWithRetry_EnforcesByteCap`
  - [x] `Peer_Sends_A_Block_Message_Exceeding_The_Byte_Cap` (in both `TestFetchBlocksBatch_CurrentBehavior` and `TestFetchSingleBlock_CurrentBehavior`)
  - [x] `Test_Server_processBlockFound_BoundsPeerFetchDeadline` and `TestBatchFetchAndDistribute_BoundsPeerFetchDeadline` (assert the outgoing request's deadline is present and no wider than 30 s)
  - [x] `TestFetchBlocksBatch_OverSendIsDiagnosticOnly` (over-send is recorded without any reputation-charging call; an exact response records nothing; a peer holding the response open after `n` blocks is given up on in about 2 s instead of 30 s)
- [x] Existing `TestCatchup_MultiIterationNoDuplicates/ThreeIterations_25000Headers` caught an off-by-one in the header byte cap during review (the missing `+1`), now passes


ctnguyen (commented):
Review at commit [f644fbf](https://github.com/liam/teranode/commit/f644fbf061abe8dbb74f352b254ec47fa3a65ce7)

This is a tidy, well-argued diff, and the direction is right: streaming into `model.NewBlockFromReader` instead of `io.ReadAll` is the correct answer for an endpoint with no consensus-defined size, bounding the batch loop to `n` closes a genuinely silly hole, and the header cap is the one place where an exact byte bound really is derivable. The counting reader is wired in properly so `RecordBytesDownloaded` keeps working under streaming, the `#1453` substitution check survives the refactor intact, and both new regression tests are load-bearing rather than decorative. Nice touch that after this change `util.DoHTTPRequest` has no production callers left at all. Two things below get in the way of merging as-is: the allocation bound the PR claims to establish is not actually established, and the effective peer-fetch deadline moves from 30 s to 600 s without that being mentioned anywhere.

## Blocking issues

### ChiR1 — Streaming the parse does not bound the block-message allocation

**Problem:** The new comments assert that "parsing off the wire bounds the allocation structurally" and that a hostile peer "can only grow the process by as much as one in-progress block's already-bounded fields" (`services/blockvalidation/get_blocks.go:1064-1068`). That is not what `readBlockFromReader` does. `block.subtreeLength` is an unvalidated wire varint (`model/Block.go:374`), and the hash loop at `model/Block.go:390-406` runs `io.ReadFull` for 32 bytes and `append`s a freshly allocated `*chainhash.Hash` on every iteration, with no ceiling:

  ```go
  // model/Block.go:388 — the 1024 is the *pre-allocation* cap, not a bound on the loop
  block.Subtrees = make([]*chainhash.Hash, 0, min(block.subtreeLength, 1024))
  for i := uint64(0); i < block.subtreeLength; i++ {
      _, err = io.ReadFull(buf, hashBytes[:])
      ...
      block.Subtrees = append(block.Subtrees, subtreeHash)
  }
  ```

  So the heap still grows in proportion to however long the peer keeps streaming — roughly 40 bytes retained (32-byte `Hash` + 8-byte pointer) per 32 bytes consumed. `httpClient`'s transport is a clone of `http.DefaultTransport` with compression left on (`util/http.go:270-280`), so the peer supplies those bytes gzipped; a repeated non-zero 32-byte pattern compresses at roughly the same ratio the issue already measured, and the zero-hash rejection at `model/Block.go:402` does not interfere with that. Nothing upstream of the parse blocks it either: `NewBlockHeaderFromBytes` (`model/BlockHeader.go:75`) does no proof-of-work check, and in `fetchSingleBlock` the `block.Hash().IsEqual(hash)` check only runs after the whole parse has completed.

**Why it matters:** This is the exact OOM in finding 1 of [bitcoin-sv/teranode#4742](https://github.com/bitcoin-sv/teranode/issues/4742), from the same fully peer-controlled `baseURL`, and merging this closes that finding. The change is a real improvement — the old path held the whole `io.ReadAll` buffer *and* the parsed structures, so peak memory drops by roughly half — but halving an unbounded quantity still leaves it unbounded, and the code comments now claim otherwise, which is what will stop the next reader from looking. Worth noting that the issue's own verification notes did not just recommend streaming: they recommended streaming the block through `model.NewBlockFromReader` *off a bounded reader* rather than materialising the whole body. This PR implements the first half of that recommendation and not the second, and the missing half is the half that does the bounding.

**Fix:** Bound what one block *message* may consume. This does not contradict the settled "no byte cap for blocks" resolution: that objection was about `Policy.ExcessiveBlockSize` being a consensus-relevant limit, and a block message carries subtree hashes plus a coinbase, not transactions, so a receive-side policy cap on it is exactly the shape of the existing `SubtreeValidation.MaxIncomingSubtreeBytes` (used a few lines up at `get_blocks.go:954`, documented explicitly as a receive-side DoS cap independent of local assembly policy). Use `io.LimitedReader` so cap-exhaustion is distinguishable from a genuine short stream, otherwise an over-cap block surfaces as `io.ErrUnexpectedEOF` and gets silently swallowed by the existing EOF branch:

  ```go
  maxPerBlock := u.settings.BlockValidation.MaxIncomingBlockMessageBytes // new setting
  for uint32(len(blocks)) < n {
      lr := &io.LimitedReader{R: countingReader, N: maxPerBlock}
      block, err := model.NewBlockFromReader(lr)
      if err != nil {
          if lr.N == 0 {
              return nil, errors.NewExternalError("[catchup:fetchBlocksBatch][%s] block message exceeds %d bytes", hash.String(), maxPerBlock)
          }
          if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
              break
          }
          return nil, errors.NewProcessingError(...)
      }
      blocks = append(blocks, block)
  }
  ```

  Same wrapper in `fetchSingleBlock`. Bounding `subtreeLength` inside `readBlockFromReader` would work too and would cover every caller, but it is a bigger blast radius. Whichever way it goes, the comments at `get_blocks.go:1064-1068` and `1129-1130` need to describe the bound that actually exists — and if the decision is to ship the streaming win now and bound the list separately, then finding 1 of #4742 should stay open rather than be closed by this PR.

### ChiR2 — Peer block fetches silently get a 600 s deadline instead of 30 s

**Problem:** `DoHTTPRequest` applies `http_timeout` when the context carries no deadline (`util/http.go:412-414`); `DoHTTPRequestBodyReader` applies `http_streaming_timeout` instead (`util/http.go:426-428`). In `settings.conf` those are `30000` (line 747) and `600000` (line 752). Two of the four converted call paths have no deadline of their own, so their effective per-fetch budget goes from 30 s to 600 s:

  - `fetchBlocksBatch` — `get_blocks.go:163` in `batchFetchAndDistribute`, whose ctx descends from `fetchAndValidateBlocks` → `errgroup.WithContext` → `catchup()` → `processCatchupChItem` → the `catchupCh` consumer goroutine started on the service-lifetime ctx at `Server.go:821-832`. `git grep -n "WithTimeout\|WithDeadline" -- services/blockvalidation/catchup.go services/blockvalidation/get_blocks.go` finds only the two reputation-RPC timeouts and the detached subtree write, nothing on this chain.
  - `fetchSingleBlock` at `Server.go:1719` in `processBlockFound`, reached from `blockProcessingWorker` (`Server.go:2009` → `2922` → `1719`), whose ctx is the one handed to `Init` at `Server.go:776`.

  The other two `fetchSingleBlock` sites (`Server.go:1047`, `Server.go:2766`) set an explicit 30 s and are unaffected — which is itself the tell that 30 s is the intended budget for this fetch.

**Why it matters:** The threat model this PR is written against is a hostile peer, and the change hands that peer a 20× longer window per request while the description says nothing about it. `batchFetchAndDistribute` loops batches of `FetchLargeBatchSize` (default 100) sequentially and each batch gets a *fresh* 600 s, and catchup is single-flight (callers branch on `errors.ErrCatchupInProgress`), so one dribbling peer now stalls catchup for minutes per batch rather than seconds. The project already treats a widened peer-fetch deadline as a defect in its own right — see [bitcoin-sv/teranode#4743](https://github.com/bitcoin-sv/teranode/issues/4743), where `context.WithoutCancel` stripping a deadline on the `subtree_data` fetch was filed as a bug on exactly this reasoning. The longer streaming timeout exists for 100+ MB subtree data files, not for block messages.

**Fix:** Give both paths an explicit deadline so the transport default never applies, e.g. in `batchFetchAndDistribute`:

  ```go
  fetchCtx, fetchCancel := context.WithTimeout(ctx, blockBatchFetchTimeout)
  blocks, err := u.fetchBlocksBatch(fetchCtx, batchHeaders[len(batchHeaders)-1].Hash(), uint32(len(batchHeaders)), peerID, baseURL)
  fetchCancel()
  ```

  and wrap the `processBlockFound` call the same way the two sibling `fetchSingleBlock` sites already do. If a longer budget for a 100-block batch is deliberate, make it a named `BlockValidation` setting and say so in the PR description rather than inheriting `http_streaming_timeout` by accident.

## Non-blocking issues

### ChiR3 — Stopping at n makes an over-sending peer indistinguishable from an honest one

**Problem:** `for uint32(len(blocks)) < n` (`get_blocks.go:1094`) exits the moment `n` blocks are parsed and ignores whatever else the peer is streaming. Before this PR the response was parsed to EOF, so a peer answering `n=1` with three blocks returned three and tripped the caller's `len(blocks) != len(batchHeaders)` check at `get_blocks.go:168-170`, which aborted the whole catchup cycle and charged the peer a generic catchup failure. The same response now returns cleanly, and the PR's new test asserts that as the desired behaviour (`get_blocks_test.go:845-848`: `require.NoError` plus `require.Len(fetchedBlocks, 1)`). Nothing in any log, metric or peer record now distinguishes that peer from one that answered exactly.

**Why it matters:** To be clear about which half of that is worth restoring: the *behaviour* change is an improvement and should stay, and the charge that went with it should not come back. The old check fired at `get_blocks.go:168`, before `verifyBlockHeaders` at `:174` — so a peer that served `n` perfectly good blocks and then appended one stray byte had those blocks discarded unverified and killed the catchup cycle, and this PR fixes that. A reputation charge would also be the wrong instrument: padding on top of correct data is at least as likely to be a query-string-blind cache in front of a peer's asset service, or version skew in `?n=` semantics, as it is to be malice, and this codebase has been deliberately retiring charges of exactly that shape — see the `ErrBlockPolicyDeclined` exemption at `peer_metrics_helpers.go:166-176`, added so honest peers are not pushed toward the reputation floor that peer selection filters on. What is genuinely lost is *visibility*. An anomaly that used to be impossible to miss is now perfectly silent, and whether it turns out to be a hostile peer or a broken cache, the operator has no way to see it at all.

**Fix:** Detect it and say so; do not charge for it. Look one byte past the last block:

  ```go
  // Diagnostic only: the n blocks above are verified by verifyBlockHeaders and the batch
  // stands. This records that the peer padded, it does not charge reputation for it.
  var probe [1]byte
  if _, err := io.ReadFull(countingReader, probe[:]); err == nil {
      u.logger.Warnf("[fetchBlocksBatch][%s] peer %s streamed more than the %d blocks requested", hash.String(), peerID, n)
      u.reportCatchupError(ctx, peerID, "peer over-sent on /blocks")
  }
  ```

  `reportCatchupError` is the right call precisely because it is diagnostic-only: it stores the string for the peer registry and dashboard (`peer_metrics_helpers.go:258-269`) and does not touch the failure counters that `reportCatchupFailure` drives via `RecordCatchupFailureWithKind` (`:144-146`, `:238-249`). If a bounded drain is preferred over a one-byte probe, the repo already has that shape in `drainErrorBody` (`util/http.go:689-692`).

### ChiR4 — util.DoHTTPRequest now has no production callers

**Problem:** `git grep -n "DoHTTPRequest(" -- "*.go"` over tracked files, excluding `_test.go` and its own definition, returns only `util/http.go:291`. This PR removed the last two production uses, so the unbounded `io.ReadAll` variant survives as a live, exported, un-annotated helper.

**Why it matters:** It is the function the whole of [bitcoin-sv/teranode#4742](https://github.com/bitcoin-sv/teranode/issues/4742) is about, and its name is the obvious one to reach for. Leaving it as the path of least resistance invites the same class of bug back in the next peer-facing fetch that gets written.

**Fix:** Either delete it and fold its tests into `DoHTTPRequestBounded`, or mark it so a reintroduction is a deliberate act:

  ```go
  // Deprecated: DoHTTPRequest reads the whole body with io.ReadAll and applies no cap.
  // Use DoHTTPRequestBounded for a caller-known size, or DoHTTPRequestBodyReader to
  // stream (bitcoin-sv/teranode#4742).
  func DoHTTPRequest(ctx context.Context, url string, requestBody ...[]byte) ([]byte, error) {
  ```

## Recap

| ChiR1 | Subtree list still unbounded | YES | 75% |
| ChiR2 | Peer fetch deadline widened 20x | YES | 55% |
| ChiR3 | Peer over-send is now silent | | 15% |
| ChiR4 | Unbounded DoHTTPRequest left exported | | 12% |

ordishs (approved):
Approving. The core fix is right: streaming the block parse off the wire removes the `io.ReadAll` hole, the `< n` loop bound closes a real unbounded-growth path, and the header cap arithmetic checks out (`maxBlockHeadersPerRequest` 10_000 x `BlockHeaderSize` 80, plus the echoed common-ancestor header = 800_080 bytes). Both new tests are load-bearing.

Two items should land before merge. Both loosen the same peer-facing path this change sets out to tighten.

**`services/blockvalidation/get_blocks.go:1069`, `:1129` — the per-request read budget grows from 60s to 600s**

- `util.DoHTTPRequest` applied `httpRequestTimeout` (`util/http.go:255`, default 60000 ms) when the context carried no deadline. `util.DoHTTPRequestBodyReader` applies `httpStreamingTimeout` instead (`util/http.go:427`), which `settings.conf:752` sets to 600000 ms.
- The catchup context carries no deadline: `catchup.go:177` starts a tracing span only, `catchup.go:1347` derives `gCtx` from an errgroup, and `get_blocks.go:163` passes it straight to `fetchBlocksBatch`.
- A peer that sends one byte per second now pins a catchup fetch goroutine for 10 minutes instead of 1. `Server.go:1719` inherits the same widening for `fetchSingleBlock`.
- The other two `fetchSingleBlock` sites (`Server.go:1047`, `Server.go:2766`) are unaffected: they already wrap a 30-second deadline.
- Suggest an explicit deadline on the fetch context, sized from `n`, rather than the 10-minute streaming default.

**`services/blockvalidation/get_blocks.go:1074-1086`, `:1134-1146` — `RecordBytesDownloaded` now under-reports an over-sending peer**

- `countingReadCloser` counts at `Read`, and the loop stops at `n`, so the surplus is never counted.
- A peer answering `n=1` with one block plus 1 GB of further well-formed blocks is charged for one block. Before this change, `len(blockBytes)` charged the whole body.
- The new `Peer Sends More Blocks Than Requested` test proves the surplus is no longer parsed. It does not assert that the surplus is still counted, which is why this is invisible in CI.
- The reputation signal that would flag this exact abuse is the signal that got weaker.
- Suggest draining the remainder through the counter under a small cap, or recording the surplus separately and reporting the peer.

Smaller items, not blocking:

- `get_blocks.go:1064-1068` — the comment claims the parse "bounds the allocation structurally". `model.readBlockFromReader` calls `bt.Tx.ReadFrom` for the coinbase, which reaches `readArenaScript` (`go-bt/v2@v2.6.9/arena.go:127-134`): that caps a single script at `MaxArenaAlloc` = 1 GiB and then does `make([]byte, l)` before `io.ReadFull`. Pre-existing and not introduced here, but about 9 bytes of varint still buy a 1 GiB transient allocation per in-flight fetch. Worth softening the comment or referencing a follow-up.
- `get_blocks.go:1095-1098` — `io.ErrUnexpectedEOF` breaks the loop, so a truncated response returns a short batch with a nil error. `batchFetchAndDistribute:168` catches the count mismatch, so no bad data reaches validation, but the operator sees "expected 4 blocks, got 3" instead of the network cause. Consider distinguishing EOF at a block boundary from `io.ErrUnexpectedEOF` mid-block.
- `get_blocks.go:1094`, `:1150` — `NewBlockFromReader` now reads from the raw HTTP body, issuing one-byte reads through `wire.ReadVarInt` and one 32-byte `io.ReadFull` per subtree hash (`model/Block.go:357-403`). The transport serves these from its own 4 KB buffer, so this is call overhead rather than syscalls, but a block with 100k subtrees still means 100k+ reader calls on the catchup hot path. `bufio.NewReaderSize(..., 64*1024)` would remove it; `bufio` is already imported in this file.
- `get_blocks.go:1156` — the `block == nil` branch is unreachable. Pre-existing; this change only rewrote its message.

Missing test scenarios:

- `fetchSingleBlock` against a response carrying trailing bytes after the first block. This is new behavior and nothing pins it.
- A truncated mid-stream body in `fetchBlocksBatch`.
- The recorded byte count on the streaming path.

ctnguyen (commented):
Review at commit [83b5817](https://github.com/liam/teranode/commit/83b5817aebe0858075f07a29de9ca7f78b9377f8)

All four points from the previous round landed, and they landed properly rather than cosmetically: the cap is a real receive-side policy setting with the same shape as `SubtreeValidation.MaxIncomingSubtreeBytes`, all four production fetch call sites now carry an explicit deadline (not just the two that were called out), the over-send probe goes through the diagnostic-only `reportCatchupError` path and the test actually asserts that no reputation-charging method fires, and the deprecation notice is in the canonical Go form. The regression tests are the good kind — `Test_Server_processBlockFound_BoundsPeerFetchDeadline` asserts the deadline is *present and narrow* on the outgoing request rather than asserting the fetch merely succeeds, which is the version of that test that can actually fail. What follows is mostly about the second-order claims the fix now makes: the new setting's operator-facing documentation states a memory bound the parser does not establish, and the one-byte probe does a blocking read on a stream the peer still controls.

On `ordishs`'s review, which saw only `f644fbf0`: his first point (the per-request budget widening to the streaming default) is closed by the same change that closed ChiR2 — though the delta's new comments get his 600 s figure wrong, which is ChiR7 below. His second point (`RecordBytesDownloaded` under-reporting an over-sending peer) is half-closed: the anomaly is now visible via the probe, and the byte counter deliberately reports only what was pulled off the wire, which is defensible accounting since the surplus is never actually downloaded — I am not re-raising it. His third point (the parse does not bound the allocation) is *not* closed and is now sharper than when he wrote it — ChiR5. His points on `io.ErrUnexpectedEOF` yielding a short batch with a nil error, the per-hash reader-call overhead, the unreachable `block == nil` branch, and the missing test scenarios are all untouched by the delta.

## Status of previous points

Two notes on what these statuses mean. ChiR1 asked for a specific thing — an `io.LimitedReader` per block message sized by a receive-side policy setting, with comments describing the bound that actually exists — and the delta delivers exactly that, so it is FIXED on its own terms; the vector ChiR1 named (the unvalidated subtree-hash-list varint) is genuinely bounded now. The broader assertion that peak allocation is bounded still does not hold, but that is a different vector and is raised fresh as ChiR5 rather than used to downgrade ChiR1. ChiR3 is likewise FIXED as specified — the probe, the log line and the diagnostic-only report are all present and the negative case is tested — but the probe was implemented verbatim from that point's suggested wording, and that wording did not account for the read blocking; ChiR6 revises the advice.

|---|---|---|---|
| ChiR1 | Streaming parse does not bound allocation | YES | FIXED |
| ChiR2 | Peer fetches get 600 s deadline | YES | FIXED |
| ChiR3 | Over-sending peer indistinguishable from honest | | FIXED |
| ChiR4 | `util.DoHTTPRequest` has no production callers | | FIXED |

## Blocking issues

### ChiR5 — The new byte cap does not bound the coinbase script allocation its documentation claims to bound

*Same as [@ordishs's point *"the comment claims the parse bounds the allocation structurally"* (16-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1741#pullrequestreview-5220671979).*

**Problem:** `io.LimitedReader` caps bytes *delivered*, not bytes *allocated*, and the largest allocation on this path is driven by a declared length rather than by delivered bytes. `model.NewBlockFromReader` ends at `model/Block.go:412-413` with `coinbaseTx.ReadFrom(buf)`. In the pinned `github.com/bsv-blockchain/go-bt/v2 v2.6.9` (`go.mod:14`), `Tx.ReadFrom` is `ReadFromWithArena(r, nil)` (`tx.go:110-111`), whose input path calls `readArenaScript(r, nil, "unlockingScript")` (`input.go:92`). On the nil-arena branch that function does `script = make([]byte, l)` (`arena.go:134`) **before** `io.ReadFull(r, script)` (`arena.go:140`), with `l` validated only against `MaxArenaAlloc = 1 << 30` (`arena.go:19`, `arena.go:127`). So roughly 130 bytes on the wire — an 80-byte header, three small varints with `subtreeLength = 0`, then a coinbase whose first input declares a 1 GiB unlocking script — buys a 1 GiB heap allocation before one script byte has arrived. The 128 MiB cap at `get_blocks.go:1116` and `get_blocks.go:1195` never fires, because it governs the read that never completes; the peer then dribbles bytes to hold the slice for the whole `peerBlockFetchTimeout` window. `ordishs` raised the mechanism against the previous commit as pre-existing and asked only that the comment be softened. This extends his point in the other direction: the delta added a *new* claim rather than softening the old one. `settings/blockvalidation_settings.go:97` now ships operator-facing text saying that streaming the parse "only bounds peak memory to roughly this cap plus one in-flight block, it does not bound it on its own" — on this path that understates the worst case by about 8x, and it is the same defect shape as ChiR1 one layer down, introduced by the fix for ChiR1.

**Why it matters:** this setting is the only knob the PR gives an operator for peer-driven block-message memory, and its `longdesc` is where someone sizing a container memory limit will look. The cap reads as authoritative and is not, for the single largest allocation the parser can be talked into. It also weakens the PR's own claim to close finding 1 of [bitcoin-sv/teranode#4742](https://github.com/bitcoin-sv/teranode/issues/4742): "bound the allocation" now holds for the subtree-hash-list vector and not for the coinbase. The underlying `readArenaScript` behaviour predates this PR and is not a regression — what is new is the documentation asserting it away.

**Fix:** state what the cap actually bounds, and name the residual instead of implying it is covered. In `settings/blockvalidation_settings.go:97`, replace the "only bounds peak memory to roughly this cap plus one in-flight block" clause with something like:

```
  settings/blockvalidation_settings.go:97:  ... this caps the bytes delivered per block
  message, which bounds the subtree hash list and hence the parsed *model.Block. It does
  not bound the coinbase transaction read: go-bt's readArenaScript allocates from the
  declared script length before reading, so a declared length up to MaxArenaAlloc (1 GiB)
  is allocated regardless of this cap. See <follow-up issue>.
```

and apply the same correction to the two in-code comments at `get_blocks.go:1111-1115` and `get_blocks.go:1192-1193`. Then open a follow-up against `go-bt` so `readArenaScript` grows its buffer against bytes actually read rather than `make`-ing the declared length up front, and reference it from both places. A local pre-check is not available here — the allocation happens inside the dependency — so the honest move is to document the residual precisely rather than leave the cap reading as a complete bound.

**Criticality:** 60%

### ChiR6 — The over-send probe blocks on a peer-controlled read, converting a stall into a silent success

**Problem:** the probe at `get_blocks.go:1142-1146` is a blocking `io.ReadFull(countingReader, probe[:])` on the peer's still-open response body. (This is the previous review's ChiR3 suggested fix implemented verbatim — a one-byte probe past the last block; that suggestion did not account for the read blocking, and this revises it.) For an honest peer the body is already at EOF — `Content-Length` reached, or the terminating chunk written — and the probe returns immediately, which is why the two new subtests pass. A peer that streams the requested `n` blocks over chunked encoding and then simply never sends the terminating chunk makes the probe block until `fetchCtx` expires at `peerBlockFetchTimeout` (30 s, `get_blocks.go:174`). At that point `probeErr != nil`, so nothing is logged and nothing is reported, and `fetchBlocksBatch` returns `(blocks, nil)`. `batchFetchAndDistribute` at `get_blocks.go:177-182` sees the full batch and the nil error and moves to the next batch, so the stall is recorded nowhere at all.

**Why it matters:** `batchFetchAndDistribute` fetches batches sequentially with a fresh 30 s context each time, and `FetchLargeBatchSize` defaults to 100 (`settings/settings.go:400`), so a sync peer can pin catchup at roughly 100 blocks per 30 s while every fetch reports success. Compare the pre-PR behaviour: `DoHTTPRequest`'s `io.ReadAll` also stalled on such a peer, but the stall surfaced as `NewNetworkTimeoutError` at `util/http.go:322-324`, so the batch failed and the failure was attributable to the peer. `f644fbf0` removed the stall entirely by stopping at `n`. This delta reinstates the stall for a purely diagnostic purpose and removes the error with it, which leaves a sandbagging peer strictly better off than it was before the PR: it gets the delay for free, with no failure counter touched and no rotation pressure from the sync-peer machinery, since real progress is still being made.

**Fix:** give the probe its own small budget so it cannot spend the fetch deadline. Since the read is governed by the request context, derive a cancellable one inside `fetchBlocksBatch` and arm a timer around the probe:

```go
  reqCtx, reqCancel := context.WithCancel(ctx)
  defer reqCancel()

  bodyReader, err := util.DoHTTPRequestBodyReader(reqCtx, url)
  // ... loop unchanged ...

  // Bound the probe: it is diagnostic only and must never spend the fetch budget.
  probeTimer := time.AfterFunc(probeBudget, reqCancel)

  var probe [1]byte
  _, probeErr := io.ReadFull(countingReader, probe[:])

  probeTimer.Stop()

  if probeErr == nil {
      // ... log + reportCatchupError as now ...
  }
```

with `probeBudget` a small constant (1-2 s) next to `peerBlockFetchTimeout`. A test for this is cheap and worth having: a responder that writes `n` blocks and then blocks on a channel, asserting `fetchBlocksBatch` returns within the probe budget rather than at the fetch deadline.

**Criticality:** 50%

## Non-blocking issues

### ChiR7 — The new comments cite a 5-minute streaming fallback that the shipped configuration never uses

*Same as [@ordishs's point **services/blockvalidation/get_blocks.go:1069**, **:1129** (16-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1741#pullrequestreview-5220671979).*

**Problem:** partial overlap — his heading carries the correct figure (600 s) for the fallback these call sites were widening to; this extends it to the fact that the delta's own comments state that figure wrongly. `util/http.go:260` compiles in `300000` ms, but `settings.conf:752` sets `http_streaming_timeout = 600000`, and `settings.conf` is committed and read on every start, so the fallback the new constant guards against is 600 s. Seven new sites say 5 minutes: `get_blocks.go:28`, `get_blocks.go:172`, `Server.go:1721`, `Server_test.go:493`, `Server_test.go:534`, `get_blocks_test.go:897-898` and `get_blocks_test.go:939`. `Server.go:1721-1722` contradicts itself inside one sentence — "the 5-minute streaming timeout ... a 20x wider window for a hostile peer than the 30s budget"; 20x is right for 600 s, 10x for 300 s. `get_blocks.go:172-173` mixes sources in one sentence the other way: "the 30s http_timeout" is the `settings.conf:747` value, while the compiled-in default at `util/http.go:255` is 60000, so one clause reads its number from the config and the adjacent one from the code.

**Why it matters:** these comments are the entire justification for `peerBlockFetchTimeout` existing, and they are what the next person will check before touching it. Someone who checks finds `600000` in `settings.conf` and has to decide whether the analysis is stale or the number is a typo — which, given the "20x" in the same sentence, is genuinely ambiguous. The pre-existing wording in `util/http.go:396` and `util/http.go:425` says "default 5 minutes", which is accurate as a statement about the compiled-in default; the new sites assert it as the effective budget, which it is not.

**Fix:** say which number is which, once, and use it everywhere. For example at `get_blocks.go:28`:

```
  get_blocks.go:28:  // ... defaults to http_streaming_timeout (600 s as shipped in
  get_blocks.go:29:  // settings.conf; 300 s compiled-in default), sized for large
  get_blocks.go:30:  // subtree_data downloads ...
```

and drop the "5-minute" phrasing from the two `require.LessOrEqual` messages and the three doc comments. `get_blocks.go:172` should either say "the 30 s `http_timeout` from settings.conf" or drop the comparison.

**Criticality:** 25%

### ChiR8 — The PR description still describes only the first commit

**Problem:** the body says "parsing off the wire bounds the allocation structurally instead" — the exact sentence ChiR1 identified as false, and which the delta's own replacement comment at `get_blocks.go:1077-1083` now explicitly contradicts ("Streaming the parse only halves peak memory versus io.ReadAll, though - it does not by itself bound anything"). The description mentions none of `MaxIncomingBlockMessageBytes`, `peerBlockFetchTimeout`, the over-send probe, or the `DoHTTPRequest` deprecation, and its test plan lists only the two tests from `f644fbf0`.

**Why it matters:** a new operator-facing setting with a 128 MiB default is invisible to anyone reading the PR page, and on a squash merge the commit message inherits this body, so the superseded rationale is what lands in the history in place of the delta's own accurate commit message.

**Fix:** add a bullet each for the per-message cap and its setting, the `peerBlockFetchTimeout` constant and the four call sites it covers, the diagnostic probe, and the deprecation; and replace the "bounds the allocation structurally" sentence with the corrected one from the delta's commit message. The test plan should pick up the three new regression tests.

**Criticality:** 10%

## Recap

|---|---|---|---|
| ChiR5 | Cap misses coinbase script allocation | YES | 60% |
| ChiR6 | Over-send probe blocks on peer | YES | 50% |
| ChiR7 | Wrong streaming timeout figure cited | | 25% |
| ChiR8 | PR description not updated | | 10% |

ctnguyen (approved):
Review at commit [2c63de9](https://github.com/liam/teranode/commit/2c63de91f6d87ca210bf7d09b689bac636797736)

This delta does what the last round asked for and does it precisely: the probe now carries its own `overSendProbeTimeout` budget with a `reqCtx`/`AfterFunc` construction that is correct on every path I traced, the setting's `longdesc` gained a `### What It Does Not Bound` section that names the `readArenaScript` residual instead of asserting it away, [bsv-blockchain/go-bt#187](https://github.com/bsv-blockchain/go-bt/issues/187) was actually filed rather than promised, and all seven "5-minute" sites now carry the shipped 600 s figure with the compiled-in 300 s named separately. The stall subtest is the good kind too — `suite.Ctx` really does carry a 30 s deadline (`services/blockvalidation/catchup_test_suite.go:56`), so `require.Less(elapsed, overSendProbeTimeout+3*time.Second)` fails without the timer rather than passing for free. Three rounds in, the peer-facing fetch path is in much better shape than it started. What follows is one bound that the corrected documentation still overstates, and four smaller things — two of which are the other reviewer's still-open points, now sharper than when he wrote them.

## Status of previous points

Two notes on the statuses. ChiR1 stays FIXED as round 2 posted it — the per-message `io.LimitedReader` sized by a receive-side policy setting is present at `get_blocks.go:1133` and `get_blocks.go:1223` and the subtree-hash-list vector it named is genuinely bounded; ChiR9 below is not a re-opening of it but an extension to the batch loop that wraps it, and I say so explicitly there rather than quietly downgrading a posted status. ChiR6 is FIXED on its own terms: the fix landed as the previous round wrote it — the author applied the `probeTimer := time.AfterFunc(probeBudget, reqCancel)` construction verbatim from that point's `Fix` field — and it is regression-tested. One residual from ChiR6's `Why it matters` is unchanged and I am recording it here rather than renumbering it: a peer that holds the response open after `n` blocks is still logged and reported nowhere, and `get_blocks_test.go:1089` now pins that silence (`"a stall is not an over-send and must not be reported as one"`). With the cost down from 30 s to 2 s per batch that is a defensible call, and the ambiguity argument that keeps the over-send diagnostic-only applies at least as strongly to a stall, so I am not asking for it to change.

|---|---|---|---|
| ChiR1 | Streaming parse does not bound allocation | YES | FIXED |
| ChiR2 | Peer fetches get 600 s deadline | YES | FIXED |
| ChiR3 | Over-sending peer indistinguishable from honest | | FIXED |
| ChiR4 | `util.DoHTTPRequest` has no production callers | | FIXED |
| ChiR5 | Cap misses coinbase script allocation | YES | FIXED |
| ChiR6 | Over-send probe blocks on peer | YES | FIXED |
| ChiR7 | Wrong streaming timeout figure cited | | FIXED |
| ChiR8 | PR description not updated | | FIXED |

## Non-blocking issues

### ChiR9 — The cap is minted fresh per block, so the batch bound is `n` times the documented one

**Problem:** the limiter is constructed *inside* the loop, at `get_blocks.go:1133`, so every iteration gets a full, fresh `MaxIncomingBlockMessageBytes` allowance:

```go
  // get_blocks.go:1122
  for uint32(len(blocks)) < n {
      // ... 1133:
      limited := &io.LimitedReader{R: countingReader, N: maxBlockMessageBytes}

      block, err := model.NewBlockFromReader(limited)
      // ...
      blocks = append(blocks, block)   // 1148 — every block is retained
  }
```

`n` is `uint32(len(batchHeaders))` (`get_blocks.go:180`), and `batchHeaders` is sliced by `largeBatchSize` = `BlockValidation.FetchLargeBatchSize`, default 100 (`get_blocks.go:107`, `settings/settings.go:400`). Every parsed block is retained in `blocks` (`get_blocks.go:1116`, `:1148`) until after the whole loop — the count check at `get_blocks.go:186` and `verifyBlockHeaders` at `get_blocks.go:192` both run *after* the memory has been spent. So one `fetchBlocksBatch` call permits `100 x 128 MiB = 12.8 GiB` delivered, not 128 MiB. Retained is worse than delivered on this path: `model/Block.go:390-406` turns each 32 bytes on the wire into a freshly allocated `*chainhash.Hash` plus an 8-byte slice slot, so 128 MiB delivered retains roughly 160 MiB, and the batch retains roughly 16 GiB. To be precise about what is verified and what is bounding in practice: the *code* permits `n x cap`; what a peer can actually push through inside the single `peerBlockFetchTimeout` covering the whole batch (`get_blocks.go:179`) is less, since the parse does ~4.2M `chainhash.NewHash` allocations per maxed-out message. Transparent gzip is on — the transport is a `http.DefaultTransport.Clone()` with `DisableCompression` untouched (`util/http.go:270-279`) — so the wire is not the limiter; CPU and the 30 s deadline are. Several GiB is the honest reading, not 12.8, and that is still one to two orders of magnitude above the figure an operator takes from the setting.

**Why it matters:** the `### What It Does Not Bound` section added by this delta (`settings/blockvalidation_settings.go:97`) exists specifically to enumerate the residuals and ends with "Size memory limits with that residual in mind". It names the one-off 1 GiB `readArenaScript` transient and omits the multiplier that is larger at default settings. Worse, the multiplier lives in a *different* setting, `blockvalidation_fetch_large_batch_size`, whose own `longdesc` says only "more memory" under Drawback — so neither page tells the operator that raising the batch size raises the peer-driven memory ceiling proportionally. This is the same defect shape as ChiR5 one level up, and it is the level where the numbers are biggest. It also caps how much of finding 1 of [bitcoin-sv/teranode#4742](https://github.com/bitcoin-sv/teranode/issues/4742) this closes on the catchup route.

**Fix:** carry one budget across the response instead of re-arming it per block. Three lines, and `limited.N == 0` still attributes exhaustion to the message that tripped it:

```go
  // One budget for the whole response, not a fresh one per message: the loop retains
  // all n parsed blocks, so a per-message cap bounds the batch only at n x the cap.
  responseBudget := maxBlockMessageBytes

  for uint32(len(blocks)) < n {
      limited := &io.LimitedReader{R: countingReader, N: responseBudget}

      block, err := model.NewBlockFromReader(limited)
      if err != nil {
          if limited.N == 0 {
              return nil, errors.NewExternalError("[catchup:fetchBlocksBatch][%s] blocks response from peer %s exceeds %d byte limit", hash.String(), peerID, maxBlockMessageBytes)
          }
          // ... unchanged ...
      }

      responseBudget = limited.N
      blocks = append(blocks, block)
  }
```

The headroom is not close: by the sibling setting's own arithmetic (`settings/subtreevalidation_settings.go:39`, 1M items x 32 bytes per subtree), a 4 GB block references single-digit subtrees, so a 100-block message runs to a few KB of hash list. 128 MiB for the whole response is still four orders of magnitude of slack. If a per-message cap is preferred for error attribution, then the minimum is documentary: say in `### What It Does Not Bound` that the batch path multiplies the cap by `FetchLargeBatchSize`, and add the cross-reference to that setting's `longdesc` too. A test mirroring `Peer_Sends_A_Block_Message_Exceeding_The_Byte_Cap` but with `n=3` and three under-cap messages that sum to over the cap would pin whichever choice is made.

**Criticality:** 45%

### ChiR10 — The cap introduces a second producer of `ErrExternal`, and both consumers document that there is only one

**Problem:** `errors.NewExternalError` is new at `get_blocks.go:1138` and `get_blocks.go:1228`. Neither site records the peer anywhere first — `git grep -n "recordCatchupPeerFailure(" -- "*.go"` over tracked files finds production call sites only at `get_blocks.go:851` and `get_blocks.go:860`, both on the subtree-data path. That matters because both package-level consumers of `ErrExternal` are written against the assumption that it can only have come from the subtree-data path, where per-peer attribution has already happened:

  - `services/blockvalidation/catchup.go:611-619` classifies any `ErrExternal` as `errorType = "peer_data_unavailable"` with `isPeerError = false`, and the comment gives the reason: "The individual failures were already attributed to the peers that produced them (recordCatchupPeerFailure)". For an over-cap block message that premise is false, and the label is wrong on its face — the data was available, it was oversized. `releaseCatchupLock` then logs "Skipping generic peer error report for catchup error type: peer_data_unavailable" (`catchup.go:691`).
  - `services/blockvalidation/Server.go:2193` dispatches on the same code and its comment states, as load-bearing reasoning for why a marker gate was deliberately not re-added, that `fetchAndStoreSubtreeAndSubtreeData` "is the only producer of this ErrExternal in the package" (`Server.go:2222-2225`). After this PR that sentence is untrue.

Two scoping facts, because they decide what the fix may safely do. Only the batch producer actually reaches those consumers: `fetchBlocksBatch` runs solely under `batchFetchAndDistribute` -> `fetchBlocksConcurrently` -> catchup, whereas all three `fetchSingleBlock` call sites are on the announcement and priority-queue routes (`Server.go:1047`, `:1725`, `:2773`, reached from `Server.go:809`, `:1490`, `:1134`, `:2929`, `:2955`), none of which is `processCatchupChItem`. So `get_blocks.go:1228`'s `ErrExternal` is inert with respect to both consumers today — it falsifies the invariant without yet tripping over it. And there is no reputation hole on the batch path either: `Server.go:2193` still reaches `reportCatchupFailureForError(ctx, c.peerID, err)` and `ReportPeerFailure`, and because the new path never sets the already-reported marker, neither is suppressed — the right peer is charged and rotation fires. The damage is telemetry and future reasoning, not accounting.

**Why it matters:** the cap is the security control this PR exists to add, and the operator-visible record of it tripping is a dashboard label that says the peer had no data plus a log line reading "All peers failed for block X" about a single peer that served an oversized message. The underlying error text survives in `PreviousAttempt.ErrorMessage`, so it is recoverable, but only by someone who already suspects the label. The stale invariant in `Server.go:2222-2225` is the more durable cost: that comment is long, careful, and is what the next person will trust when deciding whether the gate can come back.

**Fix:** make the premise true rather than special-casing the classifier, but do it in `fetchBlocksBatch` only:

```go
  if limited.N == 0 {
      capErr := errors.NewExternalError("[catchup:fetchBlocksBatch][%s] block message from peer %s exceeds %d byte limit", hash.String(), peerID, maxBlockMessageBytes)

      // fetchBlocksBatch only - see fetchSingleBlock for why it must not do this.
      u.recordCatchupPeerFailure(peerID, capErr)

      return nil, capErr
  }
```

That routes the charge through the `failedPeers` drain with the peer's own error text, which is what `catchup.go:611`'s `isPeerError = false` assumes has happened, and it is safe here because this function runs only on the catchup route, so the cycle it files into is its own.

Do **not** add the same call to `fetchSingleBlock`. `recordCatchupPeerFailure` writes into `u.activeCatchupCtx` and only no-ops when there is *no* catchup running (`catchup.go:857-868`); when one is running it files into whatever cycle happens to be active, with no interlock. Its own doc comment says so for exactly this shape of caller — "Not every caller is a catchup. RevalidateBlock reaches this path ... so its per-subtree failures land in whatever catchup cycle is active" (`catchup.go:847-856`). A gossip-route over-cap fetch concurrent with an unrelated catchup would therefore be charged into that catchup's peer set, which is a fresh attribution bug in place of a telemetry one. If the single-block route should carry attribution too, it needs an explicit channel rather than the ambient context — pass the catchup context in, or have the caller report — and that is worth its own change rather than being smuggled in here.

Either way, amend `Server.go:2222-2225` so the invariant is stated as the property the dispatch actually needs: "every producer of this `ErrExternal` applies per-peer attribution before returning" rather than naming a single function. That matters most for the single-block producer precisely because it is inert today — the comment is what will be consulted the first time someone routes it somewhere it is not.

**Criticality:** 35%

### ChiR11 — A truncated response still returns a short batch with a nil error, and the obvious split does not work

*Same as [@ordishs's point **services/blockvalidation/get_blocks.go:1095-1098** (16-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1741#pullrequestreview-5220671979).*

**Problem:** partial overlap — he identified the behaviour; this extends it to why the fix he suggested has to be written in a specific order to work at all. `get_blocks.go:1141-1143` breaks the loop on either EOF shape and `get_blocks.go:1175` returns `(blocks, nil)`, so a peer that cuts the connection mid-block yields a short batch with no error. `batchFetchAndDistribute` catches it at `get_blocks.go:186` as "expected %d blocks, got %d", which names no network cause. The shape is pre-existing, but the conditions under which it fires are not: the reader is now a live socket rather than an already-materialised buffer, so mid-stream truncation is a routine peer event on this path instead of an impossible one. His suggestion is right — distinguish clean-boundary EOF from mid-block truncation — but writing it the obvious way silently does nothing. `io.EOF.Error()` is `"EOF"`, which is a substring of `io.ErrUnexpectedEOF.Error()`, `"unexpected EOF"`. When the target is not an `*errors.Error`, teranode's `(*Error).Is` falls back to `strings.Contains(e.Error(), target.Error())` (`errors/errors.go:172-177`), and `(*Error).Error()` renders the wrapped chain inline with `" -> "` separators up to depth 20 (`errors/errors.go:106-115`). So `errors.Is(err, io.EOF)` is already true for a mid-block `io.ErrUnexpectedEOF` chain, which means the `|| errors.Is(err, io.ErrUnexpectedEOF)` clause at `get_blocks.go:1141` is dead today, and any split that tests `io.EOF` first will keep swallowing truncations. The reverse is not symmetric — `errors.Is(err, io.ErrUnexpectedEOF)` is genuinely false for a clean boundary, since `"unexpected EOF"` is not a substring of `"EOF"` — so it is a real discriminator on its own. It is just not a complete one: `io.ReadFull` returns bare `io.EOF`, not `io.ErrUnexpectedEOF`, when it finds *zero* bytes, so a stream that dies exactly on a field boundary inside a block — between two subtree hashes, say, at `model/Block.go:391` — is a genuine mid-block truncation that still renders `"EOF"`. That one case, and only that one, needs a second signal.

**Why it matters:** these are the two cases the operator most needs told apart — a peer that legitimately has fewer blocks than asked for, versus a peer whose stream died — and today they produce the same silent break and the same count-mismatch message. The ordering trap is the kind that passes review and CI: the code reads as if it distinguishes them and does not, and the same trap is documented as load-bearing elsewhere in this package for the same reason (`Server.go:2183-2192`, on `ErrExternal` before `ErrServiceError`).

**Fix:** test the narrower target first, and gate the `limited.N` signal behind an EOF-shaped error rather than using it on its own. Both arms have to stay EOF-shaped: `limited.N < maxBlockMessageBytes` is true for *any* failure that consumed bytes, including the empty-subtree-hash rejection at `model/Block.go:401-403` and every coinbase parse failure, so an ungated `limited.N` test would relabel ordinary bad blocks as truncated responses and drain the `ProcessingError` branch below it entirely. This sits after the existing `limited.N == 0` cap check at `get_blocks.go:1137`, so a cap breach is still reported as a cap breach:

```go
  // ErrUnexpectedEOF FIRST: errors.Is falls back to substring matching for non-*Error
  // targets and "EOF" is a substring of "unexpected EOF", so testing io.EOF first would
  // swallow the partial-read case. The io.EOF arm is gated on bytes having been consumed
  // because io.ReadFull returns bare io.EOF when it finds zero bytes, so a stream dying on
  // a field boundary inside a block lands there rather than on ErrUnexpectedEOF. Both arms
  // stay EOF-shaped on purpose: limited.N alone is true for any error that consumed bytes,
  // including a rejected empty subtree hash or a bad coinbase, which are parse failures and
  // belong in the ProcessingError below.
  if errors.Is(err, io.ErrUnexpectedEOF) || (errors.Is(err, io.EOF) && limited.N < maxBlockMessageBytes) {
      return nil, errors.NewExternalError("[catchup:fetchBlocksBatch][%s] peer %s truncated the response after %d of %d blocks", hash.String(), peerID, len(blocks), n, err)
  }

  if errors.Is(err, io.EOF) {
      break
  }
```

One coupling to watch if ChiR9 is taken: with a shared response budget the limiter is no longer armed at `maxBlockMessageBytes` each iteration, so the "nothing consumed" sentinel becomes the `responseBudget` value the limiter was armed with for this message, not the setting. Two subtests pin the discrimination cheaply: two whole blocks then half of a third (expect a truncation error), and a well-formed block whose subtree list contains a zero hash (expect the existing parse failure, unchanged) — the first is one of the three scenarios `ordishs` listed as missing, the second is the regression guard for this fix.

**Criticality:** 25%

### ChiR12 — The streamed parse reads through three unbuffered wrappers, and the package already has the pool for this

*Same as [@ordishs's point **services/blockvalidation/get_blocks.go:1094**, **:1150** (16-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1741#pullrequestreview-5220671979).*

**Problem:** his point stands and the delta has made the chain one layer deeper than when he wrote it. `model.NewBlockFromReader` issues one-byte reads through `wire.ReadVarInt` and a 32-byte `io.ReadFull` per subtree hash (`model/Block.go:362-391`), and each of those now traverses `io.LimitedReader` -> `countingReadCloser` -> `readCloserWithCancel` -> the transport body. What makes this worth acting on rather than noting is that the fix already exists in the package and is already used on the sibling peer-fetch path: `bufioReaderPool` (`services/blockvalidation/quick_validate.go:29-33`) hands out pooled `*bufio.Reader`s and `get_blocks.go:739-744` wraps the subtree-data body in one, commented "Use pooled buffered reader to reduce GC pressure". The block path is the one peer stream in this file that does not.

**Why it matters:** it is call overhead rather than syscalls, so this is a throughput note and not a defect — which is why it is down here. But it is the catchup hot path, it is 100 messages per batch, and the asymmetry with the subtree-data path a few hundred lines up is the kind that gets copied rather than questioned.

**Fix:** take the buffer once, outside the loop, and keep the wrapping order right — the buffer must sit *under* the per-block limiter, or its readahead is charged against the block's byte budget:

```go
  buffered := bufioReaderPool.Get().(*bufio.Reader)
  buffered.Reset(countingReader)

  defer func() {
      buffered.Reset(nil)
      bufioReaderPool.Put(buffered)
  }()

  // ... then inside the loop:
  limited := &io.LimitedReader{R: buffered, N: responseBudget}
```

One trap worth calling out since it is silent if missed: the over-send probe at `get_blocks.go:1166` must then read from `buffered`, not from `countingReader`. Reading the raw reader would skip whatever the buffer already holds and turn a real over-send into a miss. `bufio` is already imported here, and the pool's buffer is 1 MiB (`quick_validate.go:31`), so a maxed-out message costs 128 buffer fills rather than millions of wrapper hops.

**Criticality:** 15%

### ChiR13 — The new setting has no `settings.conf` entry, unlike the sibling it is modelled on

**Problem:** `git grep -n "max_incoming_block_message_bytes" -- settings.conf` over tracked files returns nothing. The setting the PR explicitly frames itself against does have one, commented out with its rationale, at `settings.conf:1395-1399`. So the two receive-side DoS caps that the PR calls "the same shape" are discoverable by different means: one by reading `settings.conf`, the other only by reading a Go struct tag.

**Why it matters:** `settings.conf` is committed and is where an operator looks to find out which knobs exist and what their defaults are. A 128 MiB cap that silently fails a peer fetch when tripped is exactly the kind of thing someone debugging a stalled catchup needs to be able to find by grepping that file. (`docs/references/settings/services/blockvalidation_settings.md` is also missing it, but so is the sibling, so that one is a pre-existing gap and not this PR's to close.)

**Fix:** mirror the sibling's entry, commented out so the default stays authoritative in code:

```
  settings.conf:  # Maximum bytes accepted per block message when fetching blocks from a peer. This is a
  settings.conf:  # receive-side DoS cap on the /block and /blocks responses, independent of any local
  settings.conf:  # block policy. Note it is applied per block message, so a FetchLargeBatchSize batch
  settings.conf:  # may deliver up to that many times this figure. See bitcoin-sv/teranode#4742.
  settings.conf:  # blockvalidation_max_incoming_block_message_bytes = 134217728
```

The third line drops if ChiR9 is fixed by sharing one budget across the response.

**Criticality:** 10%

## Recap

|---|---|---|---|
| ChiR9 | Cap re-armed per block in batch |  | 45% |
| ChiR10 | New ErrExternal breaks documented invariant | | 35% |
| ChiR11 | Truncated response returns nil error | | 25% |
| ChiR12 | Unbuffered reads on parse hot path | | 15% |
| ChiR13 | Setting missing from settings.conf | | 10% |

PR #1743, opened 2026-09-15
fix(legacy): resolve context suffixes and interpolation for legacy_config_ settings

icellan:
## Problem

`setConfigValuesFromSettings` in `services/legacy/config.go` built the legacy
bsvd-style `config` struct by reflecting over `gocore.Config().GetAll()`.
`GetAll()` returns the raw conf map: it does not resolve context suffixes
(`.dev`, `.docker`, `.operator`, ...) and does not perform `${VAR}`
interpolation. Both of those only happen in the context-aware `Get`/`GetBool`/
etc. accessors.

Concretely, a key like `legacy_config_Upnp.dev` was stripped of its
`legacy_config_` prefix to `Upnp.dev`, which was then passed straight to
`reflect.ValueOf(cfg).Elem().FieldByName("Upnp.dev")`. That never matches any
struct field, so the field lookup silently failed and the value was dropped.
Every context-suffixed `legacy_config_*` key in `settings.conf` was affected
the same way, and any use of `${VAR}` inside a `legacy_config_*` value would
also have reached the struct field unexpanded.

## Fix

`setConfigValuesFromSettings` still uses `GetAll()` to discover which
`legacy_config_*` keys exist, but the segment before the first `.` is used as
the candidate struct field name (context/app suffixes are never part of a Go
field name), and the actual value is now read through the context-aware
`Config.Get` accessor rather than the raw map. That means context suffixes
and `${VAR}` interpolation are both honoured. A key whose base name matches
no struct field now logs a warning instead of being silently discarded.

## What actually changes behaviour

Going through every `legacy_config_*` key currently in `settings.conf`:

- `ConnectPeers`, `DisableCheckpoints`, `Listeners`, `MaxPeers`,
  `MaxPeersPerIP` have no context-suffixed variants and no `${VAR}` usage
  today, so their resolved value is identical before and after this change.
- `DataDir` does use `${DATADIR}`/`${clientName}` interpolation, which now
  resolves correctly, but `cfg.DataDir` is unconditionally overwritten a few
  lines later in `peer_server.go` from `legacy_workingDir`, so this has no
  observable effect.
- `Upnp` is the one field whose resolved value actually changes. With the
  context-suffixed values as previously committed (`.dev = true`,
  `.docker = true`), fixing the loader would have flipped UPnP on in dev and
  docker contexts, a code path that, as far as I can tell, has never actually
  executed. I did not want to enable that silently as a side effect of a
  parsing bug fix, so `legacy_config_Upnp.dev` and `legacy_config_Upnp.docker`
  are set back to `false` in `settings.conf` in the same commit, keeping
  `cfg.Upnp`'s effective value unchanged for now.

## The two UPnP paths

There are two independent settings paths that both end up writing
`cfg.Upnp` in `peer_server.go`'s `newServer`:

- `legacy_config_Upnp` (and its context variants) via the reflection loader
  fixed here.
- `legacy_upnp` -> `tSettings.Legacy.Upnp`, applied a few lines later with
  `if tSettings.Legacy.Upnp { cfg.Upnp = true }`. `settings.conf` does not
  currently set `legacy_upnp`, and a separate open PR is wiring that path up.

Right now neither path can turn UPnP on (this PR keeps `legacy_config_Upnp`
neutralised, and `legacy_upnp` is unset), but having two independent knobs
feeding the same field is confusing and will produce surprising results
whichever PR lands second. I'd rather flag this than merge a fix here that
collides with in-flight work elsewhere: my recommendation is that
`tSettings.Legacy.Upnp` (already migrated to the typed settings struct) is
the one that should win long-term, and `legacy_config_Upnp` should be
removed from `settings.conf` once that's confirmed, rather than reconciling
both here.

## Also worth noting

`legacy_config_Listeners` is quoted in `settings.conf`
(`legacy_config_Listeners = "0.0.0.0"`), and gocore does not strip quotes, so
the field actually receives the 9-character string including the quotes.
This is pre-existing and out of scope here: every reader of `cfg.Listeners`
inside `loadConfig` runs before this loader's `setConfigValuesFromSettings`
call, and the actual listen addresses come from
`s.settings.Legacy.ListenAddresses` via the `listenAddrs` parameter, not from
`cfg.Listeners`. I'm leaving the quoting as-is rather than silently changing
behaviour of a field that appears to be otherwise unused.

## Tests

- `TestSetConfigValuesFromSettings_ContextSuffixResolved` - proves a
  context-suffixed key now reaches its field; fails against the previous
  implementation.
- `TestSetConfigValuesFromSettings_InterpolationResolved` - proves `${VAR}`
  interpolation now happens; fails against the previous implementation.
- `TestSetConfigValuesFromSettings_WarnsOnUnknownField` - proves the new
  warning fires for a key with no matching struct field; fails against the
  previous implementation.
- Existing `Test_setConfigValuesFromSettings` still passes, adapted to the
  new `Config`-based signature via a small fake in the test file.

## Verification

- `go build ./...`
- `go vet ./services/legacy/... ./settings/...`
- `go test ./services/legacy/... ./settings/... -race` (all green)
- `golangci-lint run ./services/legacy/...` - no new findings in the changed
  files (pre-existing `prealloc` hints elsewhere, unrelated)


ctnguyen (approved):
Review at commit [c0f3c13](https://github.com/icellan/teranode/commit/c0f3c13040aeb9b18e76fd184587744f574fa45f)

Nice catch — this is a subtle one and the diagnosis lands. `GetAll()` returns the raw conf map, so context-suffixed `legacy_config_*` keys were being handed to `FieldByName("Upnp.dev")` and quietly dropped. What I like most is the doc comment at `config.go:806-811`: it explains *why* `GetAll()` is the wrong source, not just what changed, which should stop the next person reaching for it. Neutralising the UPnP keys in the same commit is the careful move too — the effective configuration stays provably unchanged instead of flipping as a side effect. Nothing below blocks merge.

## Non-blocking issues

### ChiR1 — `IsValid()` guard stops one character short of `CanSet()`

**Problem:** The new guard at `config.go:840-844` rejects keys that name no struct field, but `reflect.Value.FieldByName` also returns a valid `Value` for the struct's **unexported** fields (`config.go:208-213`: `lookup`, `dial`, `addCheckpoints`, `miningAddrs`, `minRelayTxFee`, `whitelists`). `IsValid()` is `true` for those, so they sail past the guard and hit the type switch at `:854`, where any `Set*` call panics. Four of the six are one-letter-case typos away from real documented keys — `whitelists`/`Whitelists`, `miningAddrs`/`MiningAddrs`, `addCheckpoints`/`AddCheckpoints`, `minRelayTxFee`/`MinRelayTxFee` — and much of this repo's settings vocabulary is lowerCamelCase (`legacy_workingDir`, `legacy_grpcAddress`, `legacy_upnp`), so the typo is an easy one to make. `legacy_config_whitelists = 1.2.3.4/32` takes the `reflect.Slice` branch and panics in `newSlice.Index(i).SetString(val)` at `:914`, because the element type is `*net.IPNet`, not `string`. There is no `recover()` anywhere on the `Server.go:414` → `newServer` → `peer_server.go:3971` path, so this takes down legacy-service startup. To be clear, the panic itself is **pre-existing** — the old code ran the same `FieldByName` on the same key — but this PR does widen the surface, because base-name extraction at `:825-830` means `legacy_config_whitelists.dev` now also reaches the field where previously it was inert.

**Why it matters:** A single-character mistake in an operator-authored settings file crashes the service at boot with a raw `reflect` panic rather than a config warning. This PR adds precisely the guard where the missing check belongs, so it costs one condition to close.

**Fix:** Widen the guard at `config.go:841`:

  ```go
  field := reflect.ValueOf(cfg).Elem().FieldByName(configKey)
  if !field.IsValid() || !field.CanSet() {
      logger.Warnf("legacy_config_%s does not match any settable legacy config field, ignoring", configKey)
      continue
  }
  ```

  `CanSet()` is `false` for every unexported field, so this converts all six cases into the same warn-and-skip path the rest of the loop already uses. A test alongside `TestSetConfigValuesFromSettings_WarnsOnUnknownField` using `legacy_config_whitelists` would pin it.

### ChiR2 — The test double models a weaker gocore than the real one

**Problem:** `config_test.go:19-25` states the fake "reproduces just enough of gocore's behaviour to be meaningful here: context-suffix resolution ... and `${VAR}` interpolation". It diverges from gocore in three ways, and two of them change the answer for keys that are in the committed `settings.conf` today:

- *Context resolution does not strip segments.* `Get` at `config_test.go:58-74` tries `key+"."+context`, then `key`, and stops. gocore's `findValue` strips one trailing segment at a time, right to left, so it can land on an intermediate key. `settings.conf:826-829` defines `legacy_config_Upnp` plus `.dev`, `.docker` and `.docker.m` — nothing else. `.docker.m` is matched exactly, so that one context is not affected; every *other* multi-segment docker context this repo runs has no exact key and reaches `.docker` only through the stripping loop. For example: `docker.ci` (`Makefile:7`) → `.docker.ci` (miss) → `.docker` (hit); `docker.teranode1.test` (`compose/docker-compose-3blasters.yml:152`) → `.docker.teranode1.test` (miss) → `.docker.teranode1` (miss) → `.docker` (hit); `docker.host.teranode1` (`test/docker-compose-host.yml:139`) → `.docker.host.teranode1` (miss) → `.docker.host` (miss) → `.docker` (hit). The `.teranode2`/`.teranode3`, `.coinbase` and `docker.ss.teranode1` variants all resolve the same way. The fake jumps straight from the full-context key to the base key, so in every one of these it returns `legacy_config_Upnp` where gocore returns `legacy_config_Upnp.docker`.
- *Variable lookup is not context-aware.* `interpolate` at `config_test.go:33-42` resolves each `${VAR}` with a direct map read, `f.values[key]`. gocore's `replaceVariables` calls `getInternal`, which runs the whole chain — environment, then `findValue`'s context stripping, then the default — so the variable itself is context-resolved. Live case: `settings.conf:805` is `legacy_config_DataDir.operator = ${DATADIR}/${clientName}/legacy`, and `settings.conf:9-10` sets `DATADIR = ./data` with `DATADIR.operator = /data`. Under the `operator` context gocore yields `/data/teranode/legacy`; the fake yields `./data/teranode/legacy`.
- *Unresolvable variables take the opposite branch.* On a miss, `interpolate` returns `match`, leaving the literal `${VAR}` in the value. gocore substitutes the literal string `{UNKNOWN}`.

  The fake also ignores gocore's `app` segment and does not consult the environment the way `getInternal` does, though neither is load-bearing for these tests.

**Why it matters:** The three new tests pass against a model of gocore that is weaker than gocore, and the gap lands squarely on this PR's subject. Segment stripping is what makes the committed `legacy_config_Upnp.docker` apply under every multi-segment docker context this repo runs bar `docker.m`, and context-aware variable lookup is what makes `legacy_config_DataDir.operator` expand to `/data/...` rather than `./data/...` — both are live behaviours of keys in the committed file, and neither is exercised anywhere in the suite. A regression in either would go green. To be explicit, the double itself is the right call: the gocore singleton fixes its context on first use and there is no exported way to change it from a test, so the real `Config()` cannot reach this.

**Fix:** Mirror `findValue`'s loop, route the variable lookup back through `Get`, and flip the miss branch. Note `services/legacy/config_test.go:3-17` does not currently import `strings`:

  ```go
  // add "strings" to the import block at config_test.go:3-17

  func (f *fakeSettingsConfig) Get(key string, defaultValue ...string) (string, bool) {
      k := key
      if f.context != "" {
          k += "." + f.context
      }

      for {
          if v, ok := f.values[k]; ok {
              return f.interpolate(v), true
          }

          pos := strings.LastIndex(k, ".")
          if pos == -1 {
              break
          }

          k = k[:pos]
      }

      if len(defaultValue) > 0 {
          return defaultValue[0], false
      }

      return "", false
  }

  func (f *fakeSettingsConfig) interpolate(v string) string {
      return interpolationPattern.ReplaceAllStringFunc(v, func(match string) string {
          if val, ok := f.Get(match[2 : len(match)-1]); ok {
              return val
          }

          return "{UNKNOWN}"
      })
  }
  ```

  All four existing tests (`config_test.go:289`, `:321`, `:340`, `:354`) still pass unchanged against this. Then add the two cases the current fake cannot express: context `docker.ci` with only `legacy_config_Upnp` and `legacy_config_Upnp.docker` in the map, asserting the `.docker` value wins; and context `operator` with `DATADIR`/`DATADIR.operator` plus `legacy_config_DataDir.operator = ${DATADIR}/legacy`, asserting `/data/legacy`.

### ChiR3 — The unknown-field warning reports a key the operator cannot find

**Problem:** `config.go:842` formats the warning with `configKey`, the extracted *base* name, not the key actually present in the settings file. An operator who wrote `legacy_config_Foo.dev` is told that `legacy_config_Foo` is unrecognised — a string their `settings.conf` does not contain. With the `processed` dedup at `:832`, only the first enumerated variant warns at all, so there is no second message to disambiguate from.

**Why it matters:** The warning exists to stop a mistyped key being silently dropped, and this is new diagnostic surface added by the PR. Naming a key that greps to nothing sends the operator looking in the wrong place, which partly defeats the point.

**Fix:** Report the raw key and keep the base name as context:

  ```go
  logger.Warnf("%s does not match any known legacy config field (base name %q), ignoring", k, configKey)
  ```

  `TestSetConfigValuesFromSettings_WarnsOnUnknownField` (`config_test.go:354-364`) still passes as written, since `k` equals the full key there.

## Recap

|---|---|---|---|
| ChiR1 | Unsettable field panics loader | | 45% |
| ChiR2 | Test double weaker than gocore | | 35% |
| ChiR3 | Warning names nonexistent key | | 12% |

ordishs (approved):

PR #506, opened 2026-02-16
fix(storage): make block persister the authority for permanent file promotion

freemans13:
## Problem

Production storage is growing faster than expected due to two root causes:

1. **Block validation promotes `.subtree` files to permanent (`DAH=0`) immediately**, but only the **block persister** should do this (when a block is confirmed on the main chain). Orphaned/forked blocks' subtree files never expire.

2. **Pruner never actually deletes blob files** — `getBlobStore()` creates stores without `WithHashPrefix`, so file paths don't match the daemon's layout. `Del()` gets "file not found" → treated as "already deleted" → files stay on disk forever.

### Storage leak sources

| File Type | Creator | DAH Before | DAH After | Problem Fixed |
|-----------|---------|-----------|-----------|---------------|
| `.subtree` | block validation (new write) | **0 (permanent)** | height + retention | validation promoted too early |
| `.subtree` | block validation (existing) | **SetDAH(0)** | no change (keeps assembly DAH) | validation overwrote finite DAH |
| `.subtree` | updateSubtreesDAH | **SetDAH(0)** | skipped (keeps existing DAH) | bulk promotion to permanent |
| `.subtree` | subtreeProcessor (conflicts) | **none (permanent)** | height + retention | no DAH passed |
| `.subtreeMeta` | meta regenerator | **0 (permanent)** | height + retention | regenerator stored permanent |
| `.subtree` | block persister | not set | **0 (permanent)** | persister now promotes |

### 1. Block persister as single authority for permanent storage

Establish the **block persister as the single authority** for promoting files to permanent storage. This aligns with the architecture: the persister is the key differentiator between a full node (persister running, files promoted to permanent) and a pruned node (no persister, files expire after retention window).

**File lifecycle (new behavior):**

1. **Block assembly** creates `.subtree` / `.subtreeMeta` → **temporary** (finite DAH = height + retention)
2. **Block validation** processes `.subtree` → **temporary** (finite DAH), no longer promotes to permanent
3. **Block persister** confirms block on main chain → promotes `.subtree` and `.subtreeData` to **permanent** (`DAH=0`)
4. **Orphaned blocks** (never reach persister) → `.subtree` files expire naturally via finite DAH

### 2. Fix pruner hash prefix mismatch

`getBlobStore()` now reads `hashPrefix` from the URL query param and falls back to per-store-type defaults matching `daemon_stores.go`:

| Store Type | Hash Prefix |
|-----------|-------------|
| TXSTORE | 2 |
[...]

PR #1745, opened 2026-09-15
fix(asset): stop serving subtrees that have not been validated

freemans13:
## What happened

The external audit's finding B-004 is a read-back channel with two halves, and only one of them was still open.

On the write side, block validation's catch-up prefetch used to store a peer's response under the hash the peer asked us to fetch without checking that the bytes hashed to it. That half is already closed on `main`: #1390 made `fetchAndStoreSubtree` verify the calculated subtree root before any write, and reject a body that is not a whole number of 32-byte hashes.

On the read side, Asset deliberately fell back from `FileTypeSubtree` to `FileTypeSubtreeToCheck`. That second file type is the pre-validation copy written while this node fetches a subtree from a peer during catch-up. Serving it published, under a hash a peer chose, bytes this node had not finished checking. This PR closes that half.

The root check makes the stored file content-verified, so it is no longer arbitrary attacker-supplied content. It does not make serving it right. A route that promises a validated subtree must not answer with an unvalidated one whatever the file happens to contain, and an invariant maintained by four writers across three services is not something a public route should rest on.

## Changes

**Five reads now go through one `openValidatedSubtree` helper** in `services/asset/repository/repository.go`, rather than each restating the rule. The fallback existed in five near-identical copies, which is how it drifted in the first place; a sixth reader added later now has an obvious function to reach for. The helper carries the reasoning and the cost in its doc comment.

**The `subtree_data` presence gate lost its pending branch** in `services/asset/repository/GetSubtreeData.go`. That route regenerates transaction data on demand from whichever subtree file it finds, so admitting the pending copy there would have reopened the same hole by the back door. Its regeneration source reads back through `GetSubtreeTxIDsReader`, so a pending subtree cannot be regenerated into transaction data either.

**`GetSubtreeExists` changes least of all.** Its fallback was reached on a store *error*, never on a miss — the old code returned "does not exist" immediately when the validated check simply found nothing. So the only difference a caller sees is that a failing subtree store now surfaces its error instead of being silently re-asked about the pending file.

## Blast radius

Nothing outside Asset reads subtrees through this repository. Block assembly, the block persister, block validation, subtree validation and legacy sync all call the blob store directly, and they still consult the pending copy — correctly, since those are internal readers deciding whether they already hold a subtree, not public serving. Three names look like external callers and are not: block validation and subtree validation each have their own `GetSubtreeExists` on their own type, and two Asset concurrency settings mention the method names in their docstrings.

Two effects, both contained and both intended:

- A node that is itself catching up stops being a subtree source for its peers until it validates. Every path that accepts a block writes `FileTypeSubtree` — quick validation and full validation both do — so a caught-up node serves exactly what it served before.
- The dashboard's subtree viewer calls `/subtree/:hash/json` and `/subtree/:hash/txs/json`, so a subtree fetched but not yet validated renders as not-found rather than rendering.

Live evidence for how narrow that window is: a Hetzner mainnet node's subtree store holds 156 `subtree`, 156 `subtreeData` and 156 `subtreeMeta` files and zero `subtreeToCheck` files. In steady state the fallback was serving nothing.

## Testing

The four tests that asserted the fallback worked are now one test asserting the opposite, across all six readers, against a real in-memory blob store. Each refusal must be a not-found error, which is what the handlers turn into 404, and a final subtest writes the validated file for the same hash and requires the readers to answer. Run against the pre-fix `repository.go` and `GetSubtreeData.go`, five of the six refusal subtests fail. `GetSubtreeExists` passes there too, which is expected: its old fallback ran only on a store error.

- `go build ./...` clean
- `go test ./services/asset/...` — all five packages pass
- `go test ./services/subtreevalidation/...` and `./services/blockvalidation/...` pass, including `blockvalidation/catchup`
- `golangci-lint run services/asset/repository/...` — three `prealloc` suggestions, all pre-existing in files this PR does not touch

### Testing limits

The issue asks for a regression test asserting the pending file type is never returned by any public Asset route. I put it at the repository layer, not at the HTTP handler layer, because the handler tests use a mock repository — a route-level version would only prove the mock returned what the test told it to return. Every one of those routes reaches storage through the six methods changed here, so the repository test is the one that can actually fail.

The two other regression tests the issue lists both belong to the write side and are #1390's, not this PR's.

What this PR does not change: when a `subtreeData` file already exists, the `subtree_data` route serves it without asking whether the subtree has been validated. Catch-up (`fetchAndStoreSubtreeData` in `services/blockvalidation/get_blocks.go`) and `CheckBlockSubtrees` (`processSubtreeDataStream` in `services/subtreevalidation/check_block_subtrees.go`) both write that file before the block is validated. Neither write can carry arbitrary bytes: both parse the stream against the root-checked subtree and reject any transaction whose txid differs from the leaf at its index. So there is no read-back channel there, only transactions this node has not script-validated yet. Gating that branch on `FileTypeSubtree` looked tempting and was left out, because an on-demand `subtreeData` gets a delete-at height counted from the tip at the moment it is requested, while the `subtree` file's was fixed when it was validated (from the block's height under quick validation, from the tip at that moment under full validation), so a data file generated later outlives the subtree file and the gate would start answering 404 for data that is served today.

The mainnet file-type count is one node at one moment, not a fleet-wide measurement. Teranode's testnet and teratestnet containers on that host have no subtree store directory at present, so there is no second data point.

Closes bitcoin-sv/teranode#4842


icellan:
The de-duplication is right — five copy-pasted `FileTypeSubtreeToCheck` fallbacks collapsing into one `openValidatedSubtree` is a clear improvement, and I confirmed all six `FileTypeSubtree` read sites in the package route through it. `go test ./services/asset/...` is green including the new test, and the not-found classification does survive the `errors.NewServiceError` wrap, so the handlers really do return 404.

Two things need addressing before this lands.

## P1

**`services/asset/repository/repository.go:704` — the gate stops at the subtree file and leaves `FileTypeSubtreeData` open.**

`FileTypeSubtreeData` is written pre-validation by the catch-up fetch (`services/blockvalidation/get_blocks.go:773`) for exactly the subtrees that now 404, and it is still served unconditionally by `GetSubtreeData.go:85-104` (`GET /subtree_data/:hash`), `GetLegacyBlock.go:131`, `mining_candidate_legacy_block.go:82` and `repository.go:847`.

So: node is mid-catch-up on block N, a peer asks `/subtree/<hash>` and gets 404, then asks `/subtree_data/<hash>` and gets the full unvalidated transaction body for the same subtree.

Both file types are content-bound to the subtree root — the toCheck write root-checks, and `NewSubtreeDataFromReader` validates each tx against the subtree's nodes — so the two are equally unvalidated and equally verifiable by the requester. Gating one and not the other costs availability and buys no security. Either gate `FileTypeSubtreeData` behind `Exists(FileTypeSubtree)` as well, or state in the comment why subtree-ids are treated as a promise and tx bodies are not.

**`services/asset/repository/repository.go:713` — "a node that has caught up serves exactly what it served before" isn't accurate, and the window is wider than "while it is itself behind".**

`services/subtreevalidation/check_block_subtrees.go:193` writes `FileTypeSubtreeToCheck` for every subtree of an arriving *tip* block the node doesn't already hold, and promotion to `FileTypeSubtree` only happens once `ValidateSubtreeInternal` finishes.

So a fully synced node receiving a large tip block answers 404 on `/subtree/:hash` for the seconds-to-minutes of subtree validation — which is exactly when peers holding the same header are asking it for those subtrees. Every node validating that block is in the same window at the same time, so `fetchAndStoreSubtree`'s alternative-peer failover has nowhere to fail over to.

The comment should state that cost as it actually is. Worth also confirming the catch-up fetch path degrades gracefully here rather than charging the 404ing peer a failure (`recordCatchupPeerFailure`).

## P2

**`services/asset/repository/GetSubtreeData.go:113` — justifying comment is stale as written.** It says regeneration "reads the transaction ids out of whichever subtree file it finds". That was true before this PR; now regeneration goes through `writeTransactionsViaSubtreeStoreStreaming` → `GetSubtreeTxIDsReader` (`GetLegacyBlock.go:295`), which this PR restricted to `FileTypeSubtree` only. The `Exists` check itself is still correct — it avoids committing to 200 OK before discovering regeneration can't run — but the reason given no longer describes the code, which is how someone ends up helpfully re-adding the fallback.

**`services/asset/httpimpl/GetBlockSubtrees.go:174` — pending subtrees now render as zeroes instead of real numbers.** The error is discarded (`subtreeHead, numNodes, _ =`) and `subtreeHead == nil` produces a `SubtreeMeta` with `TxCount`, `Fee` and `Size` at their zero values. `GET /block/:hash/subtrees/json` for a block whose subtrees are still pending previously returned real fees and sizes from the toCheck copy and now returns `{"txCount":0,"fee":0,"size":0}` rows, indistinguishable from a genuinely empty subtree. Not a crash — the nil check is correct — but an API consumer can't tell "not yet validated" from "no transactions". Not in this diff, but this PR is what makes the path reachable.

**`services/asset/repository/GetLegacyBlock.go:295` — legacy block streaming can abort mid-body rather than cleanly 404.** `GetLegacyBlockReader` writes the header and coinbase, then loops subtrees preferring `FileTypeSubtreeData` and falling back to `GetSubtreeTxIDsReader`. If subtree 0 has subtreeData but subtree 5 has only toCheck and no subtreeData yet — the write ordering in `check_block_subtrees.go:264`/`:869` makes that window real, if brief — the header, coinbase and subtrees 0-4 are already in the pipe, so the failure closes the pipe with an error and the SV peer gets an aborted partial block instead of the clean 404 `GetLegacyBlock.go:78` would have produced. Pre-existing shape for genuinely missing subtrees; this PR widens the set of hashes that reach it.

## Checked and clean

The `GetSubtreeExists` comment's claim that the removed fallback fired only on store error, never on a miss, checks out against the old code. Both the `file` and `http` blob stores return `(false, nil)` on a miss, so the only affected callers (`GetTransactions.go:101`, `Search.go:139`) classify the newly-surfaced error correctly.

ordishs (approved):
**services/asset/repository/GetSubtreeData.go:93-101** — P1

- The `subtreeData` branch still serves a stored `FileTypeSubtreeData` file without asking whether the subtree is validated. Catch-up writes that file before block validation (`services/blockvalidation/get_blocks.go:641-760`).
- The PR argues the branch carries no read-back channel because the parser rejects a transaction whose txid differs from the leaf. The attacker chooses the leaves, so it also chooses the transaction bytes, and a transaction can carry an arbitrary payload in `OP_RETURN` or in a script.
- Latent, and the precondition matches B-004's: a peer with a valid-PoW header chain serves a block whose subtree list contains root R over txids it crafted, then answers both fetches. A third party reads those bytes back at `GET /subtree_data/R` until the delete-at height.
- Not a blocker. The stated reason for leaving the gate out is sound: an on-demand `subtreeData` takes its delete-at height from the tip at request time, while the `subtree` file's was fixed at validation, so a naive gate would 404 data served today.
- Recommendation: open a follow-up issue for that branch and reference it from the audit issue, rather than letting this PR record the P0 as fully shut.

Verified against the code: every accept path writes `FileTypeSubtree` (`services/subtreevalidation/SubtreeValidation.go:955`; `CheckBlockSubtrees` runs `ValidateSubtreeInternal` for any subtree lacking it at `check_block_subtrees.go:449-694`; legacy quick-validation at `services/legacy/netsync/handle_block.go:759`). Regeneration reads through `GetSubtreeTxIDsReader` (`GetLegacyBlock.go:295`), so it cannot reach the pending copy. No caller breaks.

icellan:
Re-reviewed at `180b43ed3`. That commit is a pure `Merge upstream/main` — the three files under review are byte-identical to `ca94f9f83`, so neither P1 from the previous comment is addressed yet. `go test ./services/asset/...` is green at this head (5 packages, 0 failures).

Both findings still hold; line numbers shifted with the merge:

- `FileTypeSubtreeData` is still written pre-validation at `services/blockvalidation/get_blocks.go:1050` (was `:773`) and still served ungated at `GetSubtreeData.go:85-94`. Unchanged behaviour, new line numbers.

The merge is worth flagging on the second point, because it changes how serious it is.

## The honest-peer strike is worse than I framed it

I filed this last time as "worth a sentence in the comment". That was too soft, and the code the merge brought in is the reason.

`tryPeerForSubtree` charges any non-`IsLocalError` to the serving peer at `get_blocks.go:1164`. A 404 from the asset server is a `ServiceError`, which is peer-attributable, so the charge lands on a node whose only offence is that it holds the block and is still validating it.

What the merge makes clear is that upstream has just invested a lot of care in not doing exactly this. `get_blocks.go:601` spells the mechanism out:

> It is also not IsLocalError, so tryPeerForSubtree's recordCatchupPeerFailure charge against the serving peer still lands

and the strike-once-only sequencing at `get_blocks.go:623` justifies itself on these grounds:

> excluding a possibly-sole-source honest peer is the self-isolation this work exists to prevent

That work went to considerable lengths — a cache-bypass retry and a deliberate under-strike bias — to avoid penalising an honest peer for a root mismatch it did not cause. This PR opens a fresh path to penalising honest peers, for the completely honest state of "I have this block and am validating it", and it does so fleet-wide and simultaneously: every node validating the same tip block 404s for the same window, so the alternative-peer failover in `fetchAndStoreSubtreeAndSubtreeData` has nowhere honest left to go.

That is the same self-isolation failure mode the merged-in work exists to prevent, reached by a different route. It needs handling in the code, not a caveat in the comment — either the 404 has to be classified so it doesn't draw a strike, or the pending window has to not produce a 404 in the first place.

freemans13:
@icellan, @ordishs: thanks, both. Here is where each point stands at d87abe09b.

**Fixed in d87abe09b (comments only, no behaviour change):**

- `repository.go`, openValidatedSubtree: the comment no longer claims a caught-up node serves exactly what it served before. It now says that `CheckBlockSubtrees` writes `FileTypeSubtreeToCheck` for every subtree of an arriving tip block the node does not hold, so a fully synced node 404s on those subtrees until `ValidateSubtreeInternal` promotes them, and that the 404 covers blocks fetched but not yet accepted.
- `GetSubtreeData.go`: the stale justification is gone. The comment now says regeneration can only read `FileTypeSubtree` (via `writeTransactionsViaSubtreeStoreStreaming` and `GetSubtreeTxIDsReader`), so the `Exists` check is there to avoid committing to 200 OK for a regeneration that is certain to fail, and a pending-file fallback here would have to be re-added on openValidatedSubtree too.

**Still open, waiting on a design call from Stu before any code changes:**

- The honest-peer strike (icellan's second comment): a 404 from a node that holds the block and is still validating it is peer-attributable in `tryPeerForSubtree`. Two routes were suggested: classify the 404 so it draws no strike, or keep the pending window from producing a 404. Not changed yet.
- `FileTypeSubtreeData` is still served ungated on `/subtree_data/:hash` (icellan's first P1, and ordishs's P1 with the follow-up-issue suggestion). Not changed yet.

**Noted, outside this diff:** the zero-valued rows from `GetBlockSubtrees.go` for pending subtrees, and the mid-body abort in legacy block streaming. Both code paths predate this PR. This PR makes them reachable for more hashes, so they will be handled alongside the strike decision above.

icellan:
Re-reviewed at `d87abe09b`.

The two comment fixes in `d87abe09b` are right. `repository.go` now states the tip-block window accurately instead of claiming a caught-up node serves what it served before, and the `GetSubtreeDataReader` justification now gives the real reason for the `Exists` check. I re-verified the call chain it asserts — `writeTransactionsViaSubtreeStoreStreaming` → `GetSubtreeTxIDsReader` → `openValidatedSubtree` — and it holds. Those two are closed.

Three things remain.

## The honest-peer charge has a second path, and I overstated the first one

Correction to my last comment first: I called the `recordCatchupPeerFailure` charge a P1 needing a code fix. That was too strong on its own — the charge is reputation-only, not a ban. The reason it still matters is that there are two independent paths, not one, and the second is worse in kind.

`services/subtreevalidation/SubtreeValidation.go:1036` branches explicitly on the 404 this PR introduces:

```go
if errors.Is(err, errors.ErrNotFound) {
    // Peer cannot provide subtree data - report as invalid subtree
    u.publishInvalidSubtree(spanCtx, subtreeHash.String(), baseURL, peerID, "peer_cannot_provide_subtree")
```

That reaches `p2p.Server.ReportInvalidSubtree` (`services/p2p/Server.go:3048`) and lands `UpdatePeerMetrics(..., recordFailure=true)` at `:3104`. Same shape at `SubtreeValidation.go:1348` for `peer_cannot_provide_subtree_data`.

So the new 404 is not merely a failed fetch charged to a peer — it is filed as an *invalid-subtree report* against a node whose only state is "I hold this block and am validating it". The report reason is literally `peer_cannot_provide_subtree` for a peer that can provide it and will in a few seconds.

Severity stated fairly: reputation only, self-healing on promotion, and `Server.go:3098` deliberately describes the charge as gentle. That is a P2, not a P1. What keeps it worth fixing is that both paths fire together and every node validating the same tip block is in the window at once, so the honest-peer pool shrinks fleet-wide for the duration — the shape `get_blocks.go:623` already goes to some lengths to avoid. Suppressing the invalid-subtree publish for a plain 404 looks like the cheap fix.

## `services/blockvalidation/BlockValidation.go:2245` is now false

```go
// Block assembly, the block persister and the asset service also read SubtreeToCheck directly.
```

That sentence is part of the argument for not deleting the `SubtreeToCheck` blob, and this PR makes the asset-service half untrue. Worth correcting here given the reasoning it supports.

## `FileTypeSubtreeData` is still served ungated

Third pass with no movement on this one, in either direction. `GetSubtreeData.go:85-94` serves the pre-validation `FileTypeSubtreeData` blob with no validated-subtree check, while `/subtree/<hash>` now 404s for the same hash. If the intent is that tx bodies are fine to serve unvalidated and subtree-id lists are not, that is a defensible position but it should be written down, because the current pair of comments does not distinguish them.

Also still open from the first comment: `httpimpl/GetBlockSubtrees.go:174` renders pending subtrees as `{"txCount":0,"fee":0,"size":0}`, and `GetLegacyBlock.go:295` can abort a legacy block mid-body where it would previously have 404'd cleanly.

freemans13:
@icellan, thanks for the re-review and for the correction on the strike severity. Where each point stands at ae2b8300d:

**Fixed in ae2b8300d (comment only):** `services/blockvalidation/BlockValidation.go` no longer lists the asset service as a direct reader of `FileTypeSubtreeToCheck`. It now names block assembly and the block persister, which I checked still read that file type (`SubtreeProcessor.go`, `streaming_process_subtree.go`), and says the asset service serves validated subtrees only. Nothing under `services/asset` reads the pending type any more.

**Still waiting on a design call, no code change yet:**

- The invalid-subtree report on a 404. You are right that there are two paths, and the second is the one that matters: `SubtreeValidation.go` publishes `peer_cannot_provide_subtree` (and `peer_cannot_provide_subtree_data` on the `/subtree_data` fetch) for any 404, which reaches `ReportInvalidSubtree` and records a failure in the peer registry. Suppressing the publish for a plain 404 is cheap, but it also stops charging a peer that announces a block and then withholds its subtrees, which is the case the report exists for. Whether to drop the charge, soften it, or keep the pending window from answering 404 at all is the choice Stu has to make before I change it.
- `FileTypeSubtreeData` served ungated on `/subtree_data/:hash`. Same as last round: either gate it, or write down why tx bodies are acceptable to serve before validation when subtree id lists are not, and track it in a follow-up issue as ordishs suggested. Not decided yet.
- The zero-valued rows in `httpimpl/GetBlockSubtrees.go` and the mid-body abort in `GetLegacyBlock.go`. Both paths predate this PR, but it makes them reachable for more hashes. They go with the 404 decision above.

freemans13:
@icellan thanks, this sent us down a useful hole. Short version: a node never announces a block before it can serve that block's validated subtrees, in either mining mode, so an honest announcer cannot hit the new 404. Details, and what came out of it, below.

**The honest-peer charge on a 404**

Both charges you found (`recordCatchupPeerFailure` in catch-up, and `peer_cannot_provide_subtree` / `peer_cannot_provide_subtree_data` from `SubtreeValidation.go`) are aimed at whoever we fetch from. We fetch a block's subtrees from the peer that announced it, or on failover from peers reporting the top height (`GetPeersAtMaxHeight`). So the question is whether a peer can announce a block, or report its height, while it still holds only `FileTypeSubtreeToCheck`.

It cannot. `ValidateBlock` runs `validateBlockSubtrees` → `CheckBlockSubtrees` → `ValidateSubtreeInternal` → `storeSubtreeFiles`, which writes `FileTypeSubtree`, synchronously and before either branch reaches `AddBlock` (`services/blockvalidation/BlockValidation.go` ~1978, before the optimistic branch at ~2034). Optimistic mining defers only the block-level `block.Valid` checks, not subtree validation. The block announcement comes from the `Block` notification that `AddBlock` sends (`services/blockchain/Server.go` ~1191). Block assembly writes `FileTypeSubtree` for its own subtrees before gossiping them. So the pending window you describe is real for the node doing the validating, but nobody is pointed at that node until it has announced, and by then the validated copy exists. The same holds for tip blocks arriving at a synced node.

The zero-valued rows in `httpimpl/GetBlockSubtrees.go` and the mid-body abort in `GetLegacyBlock.go` follow from the same ordering: both start from a block that is already in the chain, and every subtree of a block in the chain has its `FileTypeSubtree` by the time `AddBlock` runs. Neither path gets new hashes from this PR under that ordering, so I have left both alone.

Limit, stated plainly: this is from reading the code at the current head, not a trace from a running node.

**What the discussion did turn up**

Looking at announcement timing showed that in optimistic mode we relay a block to peers before our own `block.Valid` has finished, while optimistic mining's benefit is purely local. Two draft PRs came out of that, both independent of this one:

- bsv-blockchain/teranode#1833: p2p announces a block only once its `subtrees_set` flag is true, so optimistic blocks go out after background validation. Mined blocks are added with the flag already set, so they announce exactly as fast as today.
- bsv-blockchain/teranode#1829: the periodic `subtrees_set` sweep in block validation was setting the flag on invalid blocks (and firing setMined for them); it now excludes them, and it checks the validated subtree files exist before setting the flag, restoring a guarantee that was lost as a side effect of #506.

**`FileTypeSubtreeData` served before validation**

Written down in the code in c9aa5112d: both pre-validation writers parse the stream against a root-checked subtree and reject any transaction whose txid differs from the leaf at its index, so the file can only hold what the subtree commits to, and gating it on `FileTypeSubtree` would 404 data served today because the on-demand data file takes its delete-at height from the tip at request time. Whether to gate it differently is now bitcoin-sv/teranode#4908, as ordishs suggested.

**`BlockValidation.go` comment** was corrected in ae2b8300d.

icellan (approved):
Reviewed at `c9aa5112d`. `./services/asset/...` and `./services/blockvalidation/...` green locally.

`ae2b8300d` and `c9aa5112d` close both of my substantive objections. Two of my earlier claims were wrong and are worth correcting on the record, because both would mislead anyone reading this thread later.

## The gate I asked for on `FileTypeSubtreeData` would have broken things

I said to gate the subtree_data read behind `Exists(FileTypeSubtree)`, or explain why not. `c9aa5112d` explains why not, and the reason is one I should have found before recommending a fix. `GetSubtreeData.go:231`:

```go
dah := repo.UtxoStore.GetBlockHeight() + repo.settings.GetSubtreeValidationBlockHeightRetention()
```

The on-demand data file takes its delete-at height from the tip at request time, while the subtree file's DAH was fixed when its block was validated. The data file therefore outlives the subtree file as a matter of course, and the gate I proposed would have 404'd subtree_data that is served correctly today for any older subtree whose subtree file has already been pruned. Withdrawn — the documented justification is the right answer here, and the content argument in it (both writers reject any tx whose txid differs from the leaf at its index, so the file can only hold what the subtree commits to) is sound.

## I over-escalated the honest-peer charge

I moved that from "worth a sentence in the comment" to a P1 needing a code fix, and called it a new self-isolation vector. That was wrong. `recordCatchupPeerFailure`'s own doc comment at `catchup.go:855` already says it:

> that does not prove the peer deserves a reputational charge, because a peer that 404s subtree data the pruner removed did nothing wrong

Charging an honest peer for a 404 is pre-existing, already acknowledged in the code, and reputation-only rather than a ban. This PR widens the window in which it fires; it does not introduce the behaviour, and closing it is not this PR's job. The second path I raised (`SubtreeValidation.go:1036` → `publishInvalidSubtree`) is real but sits under the same pre-existing heading. Downgraded to a note, and the comment in `repository.go` now describes the window accurately, which was the legitimate half of the complaint.

## Minor, none blocking

- `GetLegacyBlock.go:131`, `mining_candidate_legacy_block.go:71`, `repository.go:788` and `repository.go:852` also read `FileTypeSubtreeData` ungated. The new justification lives only in `GetSubtreeDataReader`, so the "rule in one place" property this PR establishes for subtree reads does not hold for subtree_data. A cross-reference from those sites would keep the next reader from having to re-derive it.
- `httpimpl/GetBlockSubtrees.go:174` renders a pending subtree as `{"txCount":0,"fee":0,"size":0}`, which a consumer cannot distinguish from an empty subtree.
- `GetLegacyBlock.go:295` can abort a legacy block mid-body where it would previously have 404'd cleanly.
- `repository_additional_test.go:263` — the `ValidatedSubtreeIsServed` positive control re-checks five readers but skips `GetSubtreeDataReader`, the one reader whose refusal has two possible causes. The subtests also share a store and depend on declaration order, so a later `t.Parallel()` would break them.

Approving. The five-way duplication is gone, the rule is enforced in one place for subtree reads, and the comments now state the real cost instead of understating it.

freemans13:
@icellan thanks for the approval and for correcting the record. Your minor notes are in 611f1644f:

- **Cross-references for the other `FileTypeSubtreeData` reads.** `GetLegacyBlock.go` and `mining_candidate_legacy_block.go` now point at the explanation in `GetSubtreeDataReader`. `getSubtreeDataInternal` in `repository.go` needed a different note: its `GetSubtree` call already goes through `openValidatedSubtree`, so an unvalidated subtree is refused before the data file is read. I left `GetSubtreeDataReaderFromBlockPersister` alone, because it reads the block persister's store, which only holds persisted blocks.
- **Positive control skipping `GetSubtreeDataReader`.** It now covers the data reader too. With the validated subtree present and still no data file, the reader must come back and stream data, which proves the earlier refusal came from the missing validated subtree and not the missing data file. Forcing that gate to refuse makes the control fail.
- **Shared store and declaration order.** Each subtest now builds its own repository.

One limit I found while doing it: regeneration takes the process-wide quorum (`assetQuorumOnce` in `GetSubtreeData.go`), which binds to the first repository's subtree store, so the data-reader control fails under `-count` greater than 1. That is the singleton, not the gate, and it is noted in the test. `go test -race ./services/asset/...` passes at the normal count (880 tests), and lint is clean.

`GetBlockSubtrees.go` zero rows and the `GetLegacyBlock.go` mid-body abort: unchanged, for the reason in my previous comment. Both start from a block already in the chain, and every subtree of such a block has its validated file before `AddBlock` runs.

PR #1749, opened 2026-09-15
fix(util): stop peer URLs steering a chosen POST into private services

freemans13:
Closes bitcoin-sv/teranode#4843. Refs bitcoin-sv/teranode#4724.

### What happened

An external audit (finding B-032) showed that a peer can make a node send a POST with a body the peer chose to an internal service, with no request signer and no redirect involved. Two defects combine.

First, every download URL was built by pasting the peer's DataHub base URL in front of a path (`fmt.Sprintf("%s/subtree/%s/txs", baseURL, hash)`). A base ending in `?x=` turns that path into query data, so the request really goes to whatever path the peer named, for example `/v1/debug/bundle` on a Redpanda admin port. The missing-transaction POST carries 32-byte hashes that come from a subtree the same peer served, so the attacker picks the body too.

Second, the connection guard added in PR 1369 resolves each hostname and dials the address it checked, but it judges every connection on its own and deliberately allows private ranges. A peer's DNS can answer with its public server for the subtree GET and a private cluster address for the POST that follows. The auditor delivered an exact 64-byte POST to a pinned Redpanda v25.2.1 admin endpoint this way.

Binding the GET and the POST together, as the issue suggests, is not enough on its own. Subtree validation reads the subtree from the local store when it already has it, and retries with sleeps in between, so the POST can be the very first request to a hostname.

### Changes

**URLs are built from parsed parts.** New `util.JoinPeerURL` parses the base, rejects userinfo, any query (including a bare `?`) and any fragment, and appends path segments with `url.JoinPath`, keeping a real base path such as `/api/v1`. Callers add their own fixed query afterwards. Every peer fetch moved onto it: the subtree GET, the missing-transaction POST and both subtree_data fetches in subtreevalidation; `peerResourceURL`, `/block/`, `/blocks/` and the headers catchup request in blockvalidation; and the subtree meta regenerator. The p2p check on announced DataHub URLs rejects the same shapes, even with `p2p_allow_private_ips=true`, because this is about URL shape rather than address.

**Private-network addresses follow `p2p_allow_private_ips`.** With the setting at its default `false`, the connection guard now refuses a resolved address in RFC1918, IPv6 ULA `fc00::/7` or shared address space `100.64.0.0/10`. So a hostname that resolves to a private address is treated exactly like a private IP literal in an announced URL, which p2p already refuses. The 100.64/10 range is included because EKS custom pod networking and carrier NAT use it internally. The value is a process-wide atomic set by `daemon.Start`, and its zero value refuses, so a process that never sets it fails closed. The availability probe and the fetch path still share one policy.

I checked who this could break. The prod EKS DataHub hostnames resolve to Cloudflare edge addresses (104.21.x, 172.67.x). The dev-ovh clusters and the compose multi-node stacks already set the flag to `true`. The Hetzner nodes announce public hostnames. What I cannot see is the DNS view from inside a private network, so a deployment that peers over private addresses with the flag unset will now fail its fetches and needs the flag set.

**Rebinding guard for nodes that allow private networks.** With the flag `true`, the dialer remembers for ten minutes that a hostname resolved to a public address, and refuses it if it then resolves to a private one. It also refuses an answer that mixes public and private addresses, which would otherwise reach the private one through the dialer's per-address failover. It does not pin exact addresses, because legitimate peers move between public addresses. Memory is capped at 10,000 hostnames, dropping expired entries first. The guard cannot catch a hostname whose first answer is already private; only the default `false` does that.

**Redirects stay on the origin, and POSTs are never redirected.** The redirect check refuses a hop that changes host or port, allowing only an http to https upgrade on the same host. It refuses every redirect of a POST: Go already declines to replay our body on a 307 or 308, but a 301, 302 or 303 turns it into a GET to wherever the peer points. The asset service never issues redirects, so nothing legitimate should depend on this. A CDN in front of a peer that redirects across hosts would now fail; I have not seen one.

**Legacy's own-asset fetch no longer uses the peer client.** The legacy service streams blocks to SV peers by fetching `asset_httpAddress/block_legacy/<hash>` through the peer-guarded client. That address comes from our own settings. Its default is `http://localhost:8090/api/v1`, which the guard already refuses as loopback, so on main a node with default settings probably cannot serve blocks to legacy peers. I have that from the code and an existing unit test, not from a live node. The operator compose file uses `http://asset:8090`, a private bridge address that the new rule would also refuse. The new `util.DoLocalServiceHTTPRequestBodyReader` uses a plain client, and its doc says never to pass it a peer-supplied URL.

### Testing

- `TestRebinding_POSTAfterPublicGETNeverReachesPrivateService` is the regression test the issue asks for, built like the auditor's proof. 127.0.0.1 stands in for the attacker's public address and `::1` for the private victim, on the same port. The attacker serves the GET with `Connection: close`, the same hostname then resolves to the victim, and the POST must fail with the victim's hit counter at zero. It runs with the flag `false` and `true`. With the guard's check disabled, the `true` case fails.
- `TestGetMissingTransactionsBatch_QueryBaseSendsNothing` and `TestGetSubtreeTxHashes_QueryBaseSendsNothing` use the audit's `?x=` base and a catch-all responder that counts every request. Both fail against the old subtreevalidation code and pass now.
- Unit tests cover `JoinPeerURL`, the private-network policy for both flag values (including IPv4-mapped addresses and the 100.64/10 edges), the rebinding guard (window, refresh, expiry, private-first, eviction), origin-preserving redirects, a live 302 answered to a POST, `peerResourceURL`, `validateDataHubURL` with and without the flag, the daemon wiring, and the local-service client reaching localhost while the peer client still refuses it.
- Two existing util tests asserted that private ranges are always allowed; they now set the flag.
- `TestPeerProbeAllowsPrivateAddresses` in p2p pinned the old always-allow rule. It is now `TestPeerProbePrivateAddressesFollowSetting`, asserting that the probe refuses private addresses with the flag off and only fails as a network error with it on.
- `go test -race -tags testtxmetacache` passes on util, util/health, daemon, p2p, blockvalidation (and catchup), subtreevalidation, model, legacy and settings. `golangci-lint --new-from-rev upstream/main` on the same packages reports 0 issues.

Not tested: no deployed node ran this. The legacy block-serving breakage on main is inferred from code, not reproduced.

### Overlap

PR 1734 (gokutheengineer) widens the address ranges the p2p announcement check refuses and edits `services/p2p/server_helpers.go`; whichever merges second takes a small conflict. Issue 4724, assigned to gokutheengineer, is the static-check half of the same root cause: this PR closes its private-range gap at connection time, so it may be closable once this lands.


ordishs (approved):
The root cause is correctly identified and fixed at the right layer. The URL join is the durable property, and it is applied to every peer fetch except one. Two things before merge.

**daemon/daemon.go:277, util/http.go:196-206 — default-deny on private networks is a silent sync halt**

`applyPeerURLPolicy` makes the private-network rule process-wide, from `p2p_allow_private_ips`, whose shipped default is `false`.

- A Kubernetes node whose peers announce `http://asset.<ns>.svc.cluster.local:8090/api/v1` with the flag unset refuses every subtree and block fetch with "private-network address". The probe refuses the same peers, so the selector finds no sync peer and the node stops catching up.
- Nothing warns at start-up; the first symptom is a failed fetch.
- I confirmed `compose/cmd/gennodes/templates/settings.conf.tmpl:59` and `compose/settings_test.conf` already set `true`. The operator charts are not in this repo, so their default is unverified here.

Please confirm every private-network deployment sets `p2p_allow_private_ips=true`, and add a release note. A one-line start-up warning when the flag is `false` would make the cause visible before the first fetch. The per-refusal error already names the setting, which helps.

**services/p2p/peer_selector.go:623 — the availability probe still concatenates the peer base URL**

`health.CheckPeerHTTPServer` appends `/bestblockheader` with `fmt.Sprintf("%s%s", address, healthPath)`, so a base ending in `?x=` would make the probe GET a path the peer chose.

Failure: latent — no reachable path found. Every writer of `DataHubURL` into the registry goes through `validateDataHubURL`, which this PR extends with `util.ValidatePeerBaseURL`: `services/p2p/Server.go:1572`, `services/p2p/server_helpers.go:85` and `services/p2p/server_helpers.go:269`.

Suggest `util.JoinPeerURL(dataHubURL, "bestblockheader")` so the guarantee is local to the call rather than dependent on a check three layers away — the PR's own argument for `JoinPeerURL` applies here unchanged.

**util/http_rebind_test.go:276-277 — test leaks a global**

`SetSSRFProtection(true)` then `defer SetSSRFProtection(false)`. The package default is `true` (util/http.go:668-672), so this leaves SSRF protection off for every later test in package `util`. Use `orig := SSRFProtectionEnabled(); defer SetSSRFProtection(orig)`, as `useTestDialPolicy` already does.

**Verified separately**

- `url.URL.JoinPath` on Go 1.26 gives the intended path for a base with no path, a trailing slash, an `/api/v1` prefix, an uppercase scheme, and `%3F` or `%2e%2e` in the path. `%3F` stays percent-encoded, so a server sees a literal segment. `?` with no value is caught by `ForceQuery`.
- The two failing CI tests (`TestValidateBlock_CachedHeadersShortWindowIsRefused`, `TestValidateBlock_UnusableHeaderRunIsRebuiltByHashWalk`) are not in this diff and PR 1752 exists to repair them. Pre-existing, not caused by this change.

ctnguyen (commented):
Review at commit [f9f985c](https://github.com/freemans13/teranode/commit/f9f985c1da834eda04a4ea019373ec89c65ca955)

This is a good piece of security work. The two defects are attacked at the layer where they
actually live: `util.JoinPeerURL` makes "paste a path onto a peer string" structurally
impossible rather than blacklisting the one shape the auditor used, and the private-network
rule turns a resolved address into the same decision the static check already makes about an
IP literal. I swept every peer fetch independently — `git grep -n "fmt.Sprintf(\"%s"` across
`services`, `model`, `stores`, `util` and `cmd`, plus every `util.DoHTTPRequest*` and raw
`http.NewRequest*` call site — and found exactly one surviving concatenation of a peer base
URL, the same one @ordishs found; there is no second. The regression test is built the way a
regression test for this should be built (same port, two loopback families, victim hit counter
must be zero, and it runs under both settings of the flag), and the "I checked who this could
break / what I cannot see" paragraph in the PR body is the right way to ship a fail-closed
default. Moving the legacy block-serving fetch off the peer-guarded client instead of widening
that client's policy is also the right call. My findings below are about the edges: an
operational signal that is missing, one leftover call site, and two small diagnostics nits.
Note that I could not run tests here — this review is static analysis plus `go vet`.

## Blocking issues

### ChiR1 — Default-deny on private networks stops sync with no operator-visible signal

*Same as [@ordishs's point `daemon/daemon.go:277, util/http.go:196-206` (15-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1749#pullrequestreview-5211783761).*

**Problem:** `applyPeerURLPolicy` (`daemon/daemon.go:279-281`, called at `:290`) makes
`p2p_allow_private_ips` process-wide, and its shipped default is `false`
(`settings/settings.go:565`). With it unset, `privateNetworkDialPolicy`
(`util/http.go:200-206`) refuses every resolved RFC1918 / `fc00::/7` / `100.64.0.0/10`
address, so on a deployment whose peers announce hostnames that resolve into private space
both the fetches and the availability probe fail. @ordishs made this point and asked for a
start-up warning; what I can add is that the failure is quieter than "nothing warns at
start-up". The probe failure that removes the peer from selection is logged at **debug** level
only — `services/p2p/peer_selector.go:524` and `:526` are both `ps.logger.Debugf`. So at the
default log level the observable symptom is not an error at all: the health map returns `false`
for every peer, `isEligible` drops them, and the selector reports no sync peer. Nothing in the
diff emits a single line tying that to the setting.

**Why it matters:** This is the one change here that can take a working node off the chain
rather than block an attacker, and a node that stops catching up while logging nothing above
debug level is the hardest possible failure to attribute. To be explicit about the ask: the
default is not the problem and should not be flipped — `p2p_allow_private_ips=false` *is* the
fix for issue 4843 and reverting it would undo the PR. What must land before merge is the
operator-visible signal, which is a few lines inside this diff's own blast radius, and that
smallness is a large part of why blocking is defensible here. Separating what I checked from
what neither of us can check: I verified the wiring, the default and the log level; @ordishs
verified that `compose/cmd/gennodes/templates/settings.conf.tmpl:59` and
`compose/settings_test.conf` already set `true`, and that the operator charts are not in this
repo. I confirmed the same two compose paths and that `daemon/test_daemon.go:187` disables SSRF
protection outright for Go test daemons. Which real deployments peer over private addresses is
therefore a question for you, not something either review can settle from the tree.

**Fix:** Log once at start-up when the flag is off, so the cause is in the first screen of any
bug report. `applyPeerURLPolicy` already runs before any service starts and a logger is in
scope at the call site:

  ```go
  func applyPeerURLPolicy(logger ulogger.Logger, appSettings *settings.Settings) {
      util.SetSSRFAllowPrivateNetworks(appSettings.P2P.AllowPrivateIPs)

      if !appSettings.P2P.AllowPrivateIPs {
          logger.Warnf("[daemon] p2p_allow_private_ips=false: peer URLs resolving to RFC1918, fc00::/7 or 100.64.0.0/10 are refused, so peers reachable only over a private network will never be selected for sync")
      }
  }
  ```

Raising `services/p2p/peer_selector.go:526` from `Debugf` to `Warnf` for a non-timeout probe
failure would close the other half; either alone makes the breakage diagnosable, both is
better. A release note calling out the flag is worth it too.

## Non-blocking issues

### ChiR2 — The availability probe is the one peer URL still built by concatenation

*Same as [@ordishs's point `services/p2p/peer_selector.go:623` (15-09-2026)](https://github.com/bsv-blockchain/teranode/pull/1749#pullrequestreview-5211783761).*

**Problem:** `checkPeerAvailability` passes the raw DataHub URL to
`health.CheckPeerHTTPServer` (`services/p2p/peer_selector.go:623`), and
`util/health/http_health.go:53` joins it with `fmt.Sprintf("%s%s", address, healthPath)`. A
base ending `?x=` therefore still yields `GET /v1/debug/bundle?x=/bestblockheader`. My sweep
confirms @ordishs's finding and confirms it is the *only* one left. He classified it as latent
because every in-repo gossip writer of `DataHubURL` passes `validateDataHubURL`, which this PR
extends with the shape check at `services/p2p/server_helpers.go:508`; I verified those gossip
call sites (`services/p2p/Server.go:1572`, `services/p2p/server_helpers.go:85` and `:269`) and
they are indeed guarded. What extends his point is that the guarantee is not a property of the
registry the probe reads from, so it is not local to the probe. Two things bypass it. First,
the registry's own service API: `Blockchain.RegisterPeer`
(`services/blockchain/peer_registry_grpc.go:19-25`) calls `protoToPeerInfo`, which copies
`DataHubURL: p.DataHubUrl` verbatim at `:269-276`, and hands it to `Register`, which stores any
non-empty value at `services/blockchain/peer_registry.go:350-352` — there is no URL check
anywhere on that path. This is an internal service API, so I am *not* claiming a remote peer can
reach it; the point is that the invariant holds by convention of the callers rather than by
construction. Second, restart state: `CentralizedPeerRegistry.Load`
(`services/blockchain/peer_registry_persistence.go:244-253`) rebuilds the map from JSON with no
re-validation, and `DataHubURL` is an untagged, persisted field kept for
`p2p_peer_registry_ttl` (default 24h, `settings/p2p_settings.go:62`; wired at
`services/blockchain/Server.go:530,547`), so on a node that sets `blockchain_peerRegistryStore`
— off by default (`settings/blockchain_settings.go:34`) and set in no in-repo config — a base
URL accepted by the *pre-upgrade* validator survives the upgrade into the probe.

**Why it matters:** The PR's own argument for `JoinPeerURL` is that a guarantee enforced at a
distance eventually stops holding. Here it is enforced in a different package from the probe,
by three checks the probe has no way to depend on, past a registry boundary that re-admits an
unchecked URL from two directions. That is exactly the shape @ordishs's suggested fix removes,
and it is a one-line change. On impact I want to stay honest: the dial policy still applies, so
on a default `false` node the target must be a public address the peer controls. The case that
bites is a node with `p2p_allow_private_ips=true`, where this widens the acknowledged
"first DNS answer already private" residual from "GET `<base>/bestblockheader`" to "GET any
path the peer names on an internal service".

**Fix:** Build the probe URL structurally and pass an empty health path, so the property is
local to the call. `checkHTTPServer` already handles that — its second branch requires
`len(healthPath) > 0`, so an empty path leaves the URL untouched:

  ```go
  probeURL, err := util.JoinPeerURL(dataHubURL, "bestblockheader")
  if err != nil {
      return false, err
  }

  checker := health.CheckPeerHTTPServer(ps.httpClient, probeURL, "")
  ```

### ChiR3 — A base-URL rejection is logged as a fetch failure with an empty URL

**Problem:** In both subtree_data paths the join result and the join error are declared
together, then the fetch is guarded on the error being nil, but the shared failure branch
still formats `url` — which is `""` when it was the join that failed. See
`services/subtreevalidation/check_block_subtrees.go:283` and `:302-307`
(`"failed to get subtree data from %s"`), and
`services/subtreevalidation/SubtreeValidation.go:1338` and `:1342-1349`
(`"failed to get subtree data from %s: %v"`). The operator sees `failed to get subtree data
from :` followed by the wrapped cause.

**Why it matters:** These are the messages someone will grep for when a peer suddenly stops
serving subtree data after this change lands, and a truncated line reads like a formatting bug
rather than "that peer announced an unusable base URL". The wrapped error still carries the
real cause, so this is diagnosability rather than correctness.

**Fix:** Report the base URL rather than the un-built target on the join path, e.g. handle the
join error on its own before the fetch:

  ```go
  url, joinErr := util.JoinPeerURL(request.BaseUrl, "subtree_data", subtreeHash.String())
  if joinErr != nil {
      return errors.NewServiceError("[CheckBlockSubtrees][%s] invalid peer base URL", subtreeHash.String(), joinErr)
  }
  ```

which also matches what the sibling subtree fetch a few lines above already does at
`services/subtreevalidation/check_block_subtrees.go:131-134`.

### ChiR4 — `Daemon.Start` lost its doc comment to the new helper

**Problem:** `applyPeerURLPolicy` was inserted between `// Start initializes and starts the
Daemon and its services.` and `func (d *Daemon) Start`, so that sentence is now the first line
of the helper's doc block (`daemon/daemon.go:275-281`) and the exported `Start` at `:283` has
no doc comment at all.

**Why it matters:** `Start` is the daemon's public entry point, so this is a visible godoc
regression, and the helper's comment now opens by describing a different function.

**Fix:** Move the helper above the `Start` comment, or move the `Start` sentence back down onto
`func (d *Daemon) Start`.

## Recap

|---|---|---|---|
| ChiR1 | Private-network default halts sync silently | YES | 60% |
| ChiR2 | Health probe still concatenates base URL | | 50% |
| ChiR3 | Empty URL in join failure logs | | 10% |
| ChiR4 | Start doc comment misattached | | 5% |

freemans13:
@ordishs All three are fixed in 21b5992b3. Thanks for the `JoinPath` checks, I had tested the same cases from the other direction.

**Private-network default, silent sync halt.** You asked for a start-up warning and I have added one, in `applyPeerURLPolicy`, which now takes the logger. It names the setting and the three ranges. @ctnguyen added that the probe failure which drops the peer is `Debugf`, so nothing above debug level said anything at all, and I have raised that half too: a probe refused before any packet leaves is now `Warnf`, an unreachable peer stays at debug. The split is on `errors.Is(err, errors.ErrInvalidArgument)`, which is what the dial policy and the URL shape check return, and `TestPeerProbeRefusalIsInvalidArgument` pins it, including the negative case so a dead peer in the registry does not warn every round.

On the deployment question, I cannot settle it from the tree and I am not going to pretend otherwise. What I have: the production EKS DataHub hostnames resolve to Cloudflare edge addresses, the compose stacks set the flag `true`, and the Hetzner nodes announce public hostnames. What I cannot see is the operator charts, which are not in this repo, or the DNS view from inside anyone's private network. A release note is the right call and I will add one wherever you want it. Point me at the file.

**Availability probe still concatenating.** Fixed. `checkPeerAvailability` now builds the URL with `util.JoinPeerURL(dataHubURL, "bestblockheader")` and passes an empty health path, so `checkHTTPServer` leaves it alone. `TestPeerProbe_QueryBaseSendsNothing` uses the audit's `?x=` base against a catch-all responder that counts every request and asserts the count stays zero. It fails on the old code, on all four shapes.

You called it latent because every gossip writer goes through `validateDataHubURL`. @ctnguyen found two paths into the registry that skip it, the blockchain `RegisterPeer` API and the JSON reload on restart, so the guarantee was not a property of what the probe reads. It is local to the call now.

**Test leaking a global.** Fixed to `orig := SSRFProtectionEnabled()` as you suggested. One correction to the reasoning, and it does not change the fix: the leak is not live today, because `TestMain` in `util/http_test.go` sets protection `false` for the whole package test binary, so `defer SetSSRFProtection(false)` was restoring the package's test default rather than production's `true`. Saving the observed value is still right, since a nested test that enables it would otherwise be undone.

CI: the two `blockvalidation` failures are still red and still pre-existing. Thanks for finding PR 1752, I had only confirmed they fail on `upstream/main` and in main's own CI run.

freemans13:
@ctnguyen All four fixed in 21b5992b3. The independent sweep is worth a lot here: one surviving concatenation found twice, by two people, is much better evidence than one person finding it once.

**ChiR1, silent sync halt.** Both halves are in. `applyPeerURLPolicy` takes the logger and warns once at start-up when the flag is off, naming the setting and the three ranges. The probe-loop half is not a blanket `Debugf` to `Warnf`, because a dead peer in the registry would then warn on every round and bury the signal. It splits on `errors.Is(err, errors.ErrInvalidArgument)`, the class the dial policy and the URL shape check both return: refused before any packet left means a local configuration cause, and that warns; an unreachable peer stays at debug. `TestPeerProbeRefusalIsInvalidArgument` asserts both directions, so the negative case is pinned and the noise cannot come back by accident.

`TestApplyPeerURLPolicy_WarnsWhenPrivateNetworksRefused` asserts on the emitted line through a capturing logger, and I checked it fails when the warning is removed.

You are right that the default should not be flipped, and I am not flipping it.

**ChiR2, probe URL.** Fixed as you both wrote it: `util.JoinPeerURL(dataHubURL, "bestblockheader")` with an empty health path. Your two bypass paths are what made this worth doing rather than deferring, and I verified both. `RegisterPeer` copies `DataHubUrl` verbatim through `protoToPeerInfo` into `Register`, with no URL check on that path. `CentralizedPeerRegistry.Load` rebuilds from JSON with no re-validation. So the announcement check could not carry the guarantee for the probe, and now the probe does not need it to.

`TestPeerProbe_QueryBaseSendsNothing` counts every request a catch-all responder sees and requires zero, for `?x=`, a bare `?`, a fragment and userinfo. All four reach the server on the old code.

**ChiR3, empty URL in the failure log.** Fixed in both paths, with one deviation. In `check_block_subtrees.go` it is the early return you suggested. In `SubtreeValidation.go` the same shape would have wrapped an already deeply nested block in another level, so the failure branch there picks the message instead: an empty `url` means the join failed and it says the base URL is unusable, otherwise the existing "from %s" line stands. Same outcome in the log, smaller diff.

In neither place is the base URL echoed. `JoinPeerURL`'s own errors deliberately never do it, because userinfo is one of the shapes it refuses and it would copy credentials into the log. For the same reason the probe log lines now strip userinfo before logging the peer URL, which the pre-existing `Debugf` did not do and my new `Warnf` would have made louder.

**ChiR4, `Start` doc comment.** Fixed, the helper moved above it.

Testing: `go test -race -tags testtxmetacache` passes on `util`, `daemon`, `services/p2p` and `services/subtreevalidation`. `golangci-lint --new-from-rev upstream/main` reports 0 issues on those packages. The two red `blockvalidation` tests in CI are pre-existing, and @ordishs points to PR 1752 as the repair.

ctnguyen (approved):
Review at commit [4c2cf5e](https://github.com/freemans13/teranode/commit/4c2cf5ec40fab2e2f723af26edfaef2bf17c1975)

All four points from the previous round are addressed, and the two new commits are tight: eight files, +204/-18, no scope creep. The start-up warning is worth calling out specifically — it names the setting, says what is refused and says what to set, which is the shape that actually survives contact with an on-call engineer, and the accompanying test asserts on the emitted line rather than on "a call was made". Raising the probe refusal above debug closes the other half. `peerURLForLog` is a nice unasked-for extra: the probe log used to echo a peer-chosen string verbatim, and stripping userinfo there is the right instinct given `JoinPeerURL` treats credentials as a rejectable shape. The one thing I would still change is in that new warning branch, and it is a wording-and-predicate fix rather than a behaviour change. Note that I could not run tests here — this is static analysis, `go vet`, and one `go run` probe described inline below.

## Status of previous points

All four are fixed. For ChiR3 I checked both sites and swept for a third: `git grep -n "util.JoinPeerURL(" -- '*.go' ':!*_test.go'` gives eleven production call sites, and every one of the other nine (`model/subtree_meta_regenerator.go:335`, `services/blockvalidation/catchup_get_block_headers.go:235`, `services/blockvalidation/get_blocks.go:1069` and `:1123`, `services/blockvalidation/peer_cache_bypass.go:94`, `services/subtreevalidation/SubtreeValidation.go:275` and `:1021`, `services/subtreevalidation/check_block_subtrees.go:131`, `services/p2p/peer_selector.go:639`) already handles the join error on its own before the fetch, so there is no fourth site left logging an empty target. The two that ChiR3 named now diverge in style — `check_block_subtrees.go:284-290` returns early, `SubtreeValidation.go:1350-1358` branches on `url == ""` — but both are correct, and the `url == ""` test is sound because `JoinPeerURL` returns `""` on every error path and never returns `""` on success.

|---|---|---|---|
| ChiR1 | Private-network default halts sync silently | YES | FIXED |
| ChiR2 | Health probe still concatenates base URL | | FIXED |
| ChiR3 | Empty URL in join failure logs | | FIXED |
| ChiR4 | Start doc comment misattached | | FIXED |

## Non-blocking issues

### ChiR5 — The new probe warning matches a wider error class than its message describes

**Problem:** The branch at `services/p2p/peer_selector.go:527-533` warns on the whole `errors.ErrInvalidArgument` class, but that class is not confined to the two causes the comment names. Every rejection in `ssrfCheckRedirect` is an `errors.NewInvalidArgumentError` — `util/http.go:348`, `:359`, `:363`, `:369`, `:375`, `:380` — and `ps.httpClient` installs that function as `CheckRedirect` (built at `services/p2p/peer_selector.go:103`, wired at `util/http.go:441`). So a peer whose DataHub redirects off-origin lands in this branch *after* a successful dial and a complete response. I confirmed the classification with a small `go run` probe against `util.NewSSRFSafeHTTPClient` (permissive dial policy, since loopback is otherwise refused, and a 302 from one httptest origin to a second):

  What that probe printed, which is exactly what `services/p2p/peer_selector.go:527` tests:

  ```text
  err = Get "http://127.0.0.1:46679/elsewhere": INVALID_ARGUMENT (1):
        SSRF redirect check: redirect leaves the origin of the requested URL
  errors.Is(err, errors.ErrInvalidArgument) = true
  ```

  The DNS-rebind guard at `util/http.go:101` is in the same class, and the hop-count limit at `:348` fires even on a node with SSRF protection disabled, because it precedes the `ssrfProtectionEnabled` check at `:351`.

**Why it matters:** The line says "refused before dialling", which is false for every redirect rejection, and the comment tells whoever reads it next that this is "a local configuration cause". That is wrong for the redirect and rebind cases, which are peer behaviour, and it is also wrong for the first cause the comment itself names: an unusable base URL is what the *peer* announced. That case is not hypothetical — the doc comment added right below, at `services/p2p/peer_selector.go:626-630`, is explicit that unchecked base URLs reach this path through `RegisterPeer` and the persisted registry, which is the ChiR2 argument restated. The net effect is an operator-facing warning that points at local configuration for a peer-side problem, repeated at most once per peer URL per ten seconds (the negative result is cached at `:544-548`, `peerHealthCacheTTL` at `:35`). The probe outcome is correct in all of these cases, so this is diagnosability, not behaviour. The branch came out of the previous round's own suggested fix and raising the level was the right call; only the predicate's framing needs narrowing.

**Fix:** Keep the class but drop the two claims it cannot support, so the line stays true for every cause that reaches it:

  ```go
  case errors.Is(err, errors.ErrInvalidArgument):
      // Refused by our own SSRF policy rather than by the network: a base URL the
      // peer announced that we cannot build on, an address this node bars, or an
      // off-origin redirect. Either way the peer silently drops out of selection,
      // so it is not a debug-level event. An unreachable peer stays at debug.
      ps.logger.Warnf("[PeerSelector] Availability probe for %s refused by SSRF policy: %v", peerURLForLog(url), err)
  ```

  Matching the specific causes instead would work too, but it costs more code than the wording fix buys.

PR #1750, opened 2026-09-15
MvP-1139 Blockvalidation: bound catch-up subtree-data prefetch by declared bytes

### LINKED ISSUE #1139: blockvalidation OOM crash-loop during catchup — prefetch bounded by block-count, not bytes (large-script region)
## Summary

During p2p catchup on testnet, blockvalidation enters an OOM crash-loop. It retains a large, growing amount of parsed transaction-script data from prefetched blocks while validation is stalled on one block, hits the container memory cap (~12.7 GB), dies, restarts, retries the same block, and repeats. Chain height pins and the node cannot progress.

Observed on `bsva-ovh-teranode-ttn-eu-4` (testnet, `COMPOSE_PROFILES=p2p`, build `7508c2218`, same class seen in the `bad09b886` lineage), stuck at height ~1,681,787, `RestartCount=9`.

## Symptoms

- blockvalidation memory climbs 75 MiB → ~12.7 GB in ~3 min (bursty near the end: 7→10.7 GB in ~6 s), then dies; ~1–3 restarts per 15 min for hours.
- FSM stuck `CATCHINGBLOCKS`; height pinned (zero `stored block` events for 3h+).
- ~500 goroutines, flat across the memory climb — **not** a goroutine leak. 245 blocked `chan receive, 6 min` (idle downstream worker pools waiting on the stalled pipeline).

## Heap evidence (pprof, captured at 3/6/9/10.5 GB)

- 87% of inuse_space in `go-bt readArenaScript`, but **few, very large allocations** (not in the top allocation-*count* list) — i.e. large scripts, not many small ones.
- Split: input scripts ~940 MB (67%), output scripts ~462 MB (33%) at the 3 GB mark; scales with total to OOM.
- Cum path: `blockvalidation.fetchSubtreeDataForBlock → go-subtree Data.serializeFromReader → go-bt Tx.ReadFrom → readArenaScript`.
- **This is NOT the arena allocator.** `Tx.ReadFrom` passes a nil arena (`go-bt v2.6.7 tx.go:111`); with a nil arena `readArenaScript` uses plain `make([]byte, n)` (`arena.go:107`). The arena is dormant on this path. The retained memory is ordinary GC-reachable parsed-tx script data — don't chase the arena, the function name is misleading.

## Root cause (two coupled problems)

1. **Trigger (not yet pinned):** validation wedged on block ~1,681,787 — height never advances, downstream worker pools idle. Needs separate diagnosis (candidates: the large block itself slow/failing to validate, or a mined-set / quorum / missing-tx gate).
2. **OOM amplifier (the memory bug):** while validation is stalled, catchup keeps prefetching and parsing blocks ahead and holds all their parsed tx-script bytes live. This chain region carries large-script transactions (data-carrier / stress txs), so retained per-block memory is large. The prefetch pipeline bounds **block count** (`blockvalidation_fetch_buffer_size=50`, `blockvalidation_fetch_large_batch_size=100`), not **bytes** — `fetchBlocksBatch` pulls 100 full blocks in one HTTP request and parses every tx into a `[]*model.Block`. In a large-script region a single batch is multiple GB. There is no byte-based back-pressure, so a stalled consumer lets memory climb to OOM.

It synced 0→~1.68M fine; the blow-up only happens when (a) a block stalls validation and (b) the region has large blocks.

## Proposed fix

- Bound catchup prefetch by **bytes in flight**, not block count, so a large-script region or a stalled consumer back-pressures the fetcher.
- Likely also stream/cap `fetchBlocksBatch` so it doesn't materialize 100 fully-parsed blocks at once.
- Separately root-cause and fix the validation stall on the trigger block — without it, the prefetch wouldn't run away.

## Immediate mitigation (config, no rebuild)

```
blockvalidation_fetch_large_batch_size=10   # 100 -> 10 (biggest lever)
blockvalidation_fetch_num_workers=2         # 16 -> 2
blockvalidation_fetch_buffer_size=6         # 50 -> 6
```

Raising `MEM_LIMIT_BLOCKVALIDATION` is not a fix — growth outruns any cap in a large-block region.

## Repro / context

- Node: `bsva-ovh-teranode-ttn-eu-4`, testnet, quickstart, `COMPOSE_PROFILES=p2p`, `listen_only`.
- Build: ghcr `7508c2218` (`bad09b886` lineage).
- Stuck height ~1,681,787 (BSV testnet; region has large-script txs).
- Heap + goroutine dumps on the host: `/tmp/bvheap/` — `heap-{3000,6000,9000,10500}.prof`, `goroutine-*.txt`. Analyze with `go tool pprof -top -inuse_space heap-3000.prof` (the 10500 dump is truncated; use 3000/6000/9000).

## Relevant code

- `services/blockvalidation/get_blocks.go` — `fetchBlocksConcurrently` / `batchFetchAndDistribute` / `fetchBlocksBatch` / `blockWorker` / `orderedDelivery` (the prefetch pipeline + buffer sizing).
- `settings/blockvalidation_settings.go` — `blockvalidation_fetch_large_batch_size` / `_num_workers` / `_buffer_size`.
- Parse site: go-subtree `Data.serializeFromReader` → go-bt `Tx.ReadFrom` (nil arena).

ctnguyen:
This resolve https://github.com/bsv-blockchain/teranode/issues/1139

## What changed

Catch-up ran up to `blockvalidation_fetch_num_workers` × `blockvalidation_subtree_fetch_concurrency`
(16 × 32) concurrent subtree_data fetch-and-parse operations with no bound on the bytes in
flight. In a large-script region that reached 12.7 GB and the service was OOM-killed in a
crash-loop while validation was stalled.

- Each block's subtree-data prewarm reserves against a byte-weighted budget sized by
  `blockvalidation_catchup_prefetch_budget_bytes` (default 256 MiB, matching
  `legacy_blockPrefetchBufferBytes`). The weight is the block's declared size, floored at 64 KiB and
  clamped to the whole budget. `0` disables the budget and the oversized-block subtree-concurrency
  rule; it does **not** restore the previous behaviour exactly, because the streamed store write
  below applies either way.
- A block whose declared size exceeds the budget — or whose declared size does not fit an `int64` —
  is admitted alone and parses its subtrees one at a time.
- A block that declares **no** size is given the same *subtree concurrency* treatment (one at a
  time), because an undeclared size is the value least worth trusting. Its **block-level reservation
  is still the 64 KiB floor**, so several such blocks are still admitted concurrently; what this
  removes is the 32-way subtree fan-out underneath each of them.
- Back-pressure reaches the network: a parked worker stops draining the work queue, which fills
  and blocks the next `/blocks?n=` request.
- Subtree data is streamed to the blob store instead of being built as a second full in-memory copy
  with `Serialize()`. The second copy is avoided on the file store; the memory and batcher stores
  `io.ReadAll` the body and S3 buffers it before upload.
- Corrected the documented defaults, and the descriptions that contradicted them, for
  `blockvalidation_fetch_num_workers` (1 → 16) and `blockvalidation_subtree_fetch_concurrency`
  (8 → 32).

A synthetic benchmark (`get_blocks_bench_test.go`) sweeps three block shapes against three
budgets. On a block larger than the budget the default cuts sampled incremental heap by about
42%, at roughly 7.5× the prewarm time; raising the budget above the largest expected block size
lifts the per-block concurrency limit again. Where the admitted blocks sum under the budget, the
measurements do not show a difference from it being disabled.

The benchmark now also sweeps injected per-response latency on a 64-subtree oversized shape, and
reports `R = t(budget on) / t(budget off)`:

| Injected per-response latency | R — workstation, `GOMAXPROCS=16` | R — `Bench blockvalidation` CI, 4 cores, `07b962b6` |
|---|---|---|
| 0 ms | 13.71 | 3.0 |
| 20 ms | 15.04 | 21.4 to 24.0 |
| 200 ms | 21.37 | 30.5 to 30.8 |

In absolute terms the CI 200 ms cell is 846 ms per block with the budget off against 25.9 s with it
on, for one 64-subtree block. The CI column is reproducible by rerunning that job.

This is a latency-sensitivity experiment on reduced payloads — one block per cell, `httpmock`
responders with injected sleeps, the per-cell budget lowered to force the oversized branch — and
**not** a scale-size catch-up measurement. It confirms the serial-fetch cost rather than answering
it, and exceeds the 7.5x quoted above. The oversized rule's real cost on a scale network is still
unmeasured; the procedure for measuring it is Task 1 of
https://github.com/bitcoin-sv/teranode/issues/4863.

**Those figures came from one developer workstation and do not transfer to CI or production.**
They were taken on a single 16 GiB WSL2 machine at `GOMAXPROCS=16`, against `httpmock` rather
than real peers and a local temp directory rather than the deployed blob backend, and with
fixtures small enough that each block carries fewer subtrees than
`blockvalidation_subtree_fetch_concurrency` — so the 16 × 32 concurrency product this change
bounds is never actually reached. The throughput ratio in particular is a property of that
machine: no network, no contention, no container limit. Treat the numbers as evidence that the
admission control engages and in which direction, not as a forecast of the cost anywhere else.
Re-measure on the target deployment before tuning
`blockvalidation_catchup_prefetch_budget_bytes` away from the default.

## Status against the issue's three proposals

**Resolved** — catch-up prefetch admission is governed by declared-byte reservations rather than
a block count. Actual bytes in flight are not bounded: the weight is the peer's declared block
size (see **Limit**).

**Partly resolved** — the issue also asks to stream/cap `fetchBlocksBatch`. The runaway
prefetch it targets is fixed: a parked worker stops draining the work queue, which blocks the
next `/blocks?n=` request. The per-response byte cap is **not** done — `fetchBlocksBatch` and
`fetchSingleBlock` still `io.ReadAll` a peer body. Deferred because a local configurable cap
returns an attributable error, and since we choose `n` and reject a mismatched block count,
every honest peer asked for the same `n` is charged identically — the self-isolation the
`excessiveblocksize` decline exemption already exists to prevent.

**Not addressed** — the validation stall that triggered the incident is not diagnosed. This
change reduces the memory amplifier that ran alongside it; whether the node survives the stall
is untested.

**Limit** — the reservation weight is the peer's declared block size: a heuristic, not a memory
bound. `subtree_data` may carry extended transactions, which the parser accepts and which add a
`PreviousTxScript` per input beyond that size. One subtree_data payload also remains
irreducible, because it must be materialised to be checked. The budget governs declared
serialized bytes, not process heap. No test here replays a real large-script-region
`subtree_data`.

The issue's config mitigation (`fetch_large_batch_size=10`, `fetch_num_workers=2`,
`fetch_buffer_size=6`) can be **reconsidered** after deployment measurements or a replay, not
reverted on the strength of this PR: the benchmark ran below the shipped concurrency product
and measures a sampled heap increment, not resident memory. Size
`blockvalidation_catchup_prefetch_budget_bytes` against measured headroom on the target
deployment, not against the container memory limit.


services/blockvalidation/get_blocks.go
```diff
@@ -624,6 +647,140 @@ func (u *Server) fetchAndStoreSubtree(ctx context.Context, block *model.Block, s
...
+func (u *Server) boundSubtreeConcurrencyByBudget(configured int, block *model.Block) int {
+	if u.catchupPrefetchBudgetBytes <= 0 || block == nil || block.SizeInBytes == 0 {
+		return configured
+	}
+
+	// A size that does not even fit an int64 cannot fit the budget either, so the
+	// conversion failure and the over-budget case share one verdict.
+	size, err := safeconversion.Uint64ToInt64(block.SizeInBytes)
+	if err != nil || size > u.catchupPrefetchBudgetBytes {
+		return 1
```
freemans13, line 826: Returning 1 here serialises the whole block's subtree work, including the network fetch, not just the parse. Blocks over 256 MiB are common on the scale networks. With `initial_merkle_items_per_subtree = 32768` from settings.conf, a 1 GiB block is roughly 60 to 130 subtrees, now fetched one at a time. The benchmark's oversized fixture has 11 subtrees per block and no latency, so it understates this.

The doc comment is right that an average is unsafe under skew. But there is a per-subtree number that is not an average. Each `.subtree` file carries `SizeInBytes`, and `fetchSubtreeAndDataFromPeer` already has the parsed subtree before it fetches that subtree's data (`get_blocks.go:990`). The peer supplies it, which is the same trust level as `block.SizeInBytes`. `model.Block` even recomputes the block size from these values once the subtrees load (`model/Block.go:2138` and `:2184`).

So here is the alternative. Give each block a `semaphore.Weighted` sized to that block's reservation. Each subtree_data fetch, after the exists check, acquires `min(max(subtree.SizeInBytes, floor), blockWeight)` from it. The worst case is the same as concurrency 1: at most the reservation in flight, or one subtree bigger than the reservation, which is irreducible either way. An evenly split 1 GiB block keeps its 32-way parallelism. A skewed block runs its two huge subtrees alone and the rest in parallel. Blocks that fit the budget do not change.

If you would rather keep this PR as it is, I think it at least needs a catch-up measurement on a scale-size block before 256 MiB ships as the default. As written, raising the budget is the only way to get concurrency back.

services/blockvalidation/get_blocks.go
```diff
@@ -761,21 +918,54 @@ func (u *Server) fetchAndStoreSubtreeData(ctx context.Context, block *model.Bloc
...
+	pr, pw := io.Pipe()
+	done := make(chan struct{})
+
+	go func() {
+		defer close(done)
+
+		// bufio is mandatory, not cosmetic: bt.Tx.SerializeTo delegates to WriteTo, which
+		// emits many 4-byte and varint-sized writes, and every write to an unbuffered
+		// io.Pipe is a synchronous rendezvous with the reader.
+		bw := bufio.NewWriterSize(pw, 1024*1024)
```
freemans13: This allocates a new 1 MiB buffer for every subtree_data, whatever its size. For large payloads that is a clear win over the `Serialize()` copy it replaces. For small subtrees, which covers most of chain history, it allocates more than `Serialize()` did, once per subtree and up to 16 x 32 at a time. A `sync.Pool`, like `bufioReaderPool` in `quick_validate.go`, or a size of `min(1 MiB, subtree.SizeInBytes)` would fix it.

It also means `blockvalidation_catchup_prefetch_budget_bytes = 0` does not restore the previous behaviour exactly, as `Server.go:503` and the PR body say, because the streaming change applies either way. Worth rewording.

One more on the comment above: the second copy is only avoided on the file store. The memory and batcher `SetFromReader` implementations `io.ReadAll` the body, and S3 copies it into a `bytes.Buffer` (`stores/blob/s3/s3.go:262`). That is probably fine, since the subtree store is file-backed in practice, but the comment reads as if it holds for every store.

freemans13 (changes requested):
**Requesting changes for one thing.** Any block over the 256 MiB budget has its subtree fetch concurrency dropped to 1, and on large-block catch-up that costs a lot of throughput. There is a finer weight that keeps the skew safety without the serial fetch. The byte budget itself looks right to me, and the tests are good: on a scratch copy I removed the release call, then separately disabled the oversized rule, and each time the new tests failed (2 and 3 tests respectively). So they guard what they claim.

What matters:

1. **Serial subtree fetch for every block over the budget** (inline on `get_blocks.go:778`). Your 16-core run measured prewarm 7.5x slower. The `Bench blockvalidation` CI job on this PR ran the same benchmark on 4 cores and got about 309 MB/s on the oversized shape with the 256 MiB budget, against 386 to 414 MB/s with it off. So the cost grows with core count. Neither run has network round trips, and a serial fetch pays one per subtree.
2. **A fresh 1 MiB bufio writer per subtree_data** (inline on `get_blocks.go:938`). Small, but it also makes the "0 restores the previous behaviour exactly" claim untrue.

**CI:** the red `test` job is not caused by this PR. `TestValidateBlock_CachedHeadersShortWindowIsRefused` and `TestValidateBlock_UnusableHeaderRunIsRebuiltByHashWalk` fail the same way when I run them locally on `upstream/main` at c3db82e72, the subtree conflict fix that merged today. Because that job failed, Sonar did not run, so nobody has checked coverage of the new code yet.

**Not checked:** I did no real catch-up run and nothing at scale-network block sizes. The throughput figures above come from the two benchmark runs, and I have not measured the per-subtree alternative I suggest inline.

ordishs (approved):
Verified the central equivalence claim against the dependency sources: `Data.Serialize` and `Data.WriteTransactionsToWriter` (go-subtree v1.4.6, `subtree_data.go:78` and `:126`) skip index 0 under the same placeholder condition over the same range, and `Tx.SerializeBytes` / `Tx.SerializeTo` (go-bt v2.6.9, `tx.go:418` and `:756`) branch identically on `IsExtended()`. No partial-write hazard: the file store removes its temp file on a reader error, and the memory and S3 stores buffer the whole body before they commit. The pipe teardown order is the only one that does not deadlock, and the inline release is correctly justified — a `defer` inside a `for` body is function-scoped in Go.

CI note: the two `test` failures (`TestValidateBlock_CachedHeadersShortWindowIsRefused`, `TestValidateBlock_UnusableHeaderRunIsRebuiltByHashWalk`) are pre-existing, not from this change. Neither file is in this diff, both originate in `subtreevalidation.cachedSubtreeConflictingNodes`, and PR 1751 fails on exactly the same two.

Two items below. Neither is a defect in the code as written — both are gaps between what the mitigation bounds and what the issue describes.

---

### `services/blockvalidation/get_blocks.go:684-690, 767-774` — P1: the weight is peer-declared and unvalidated

`model/Block.go:368` reads `SizeInBytes` with `wire.ReadVarInt` straight out of the bytes the peer served. Nothing validates it before it becomes the reservation weight. `boundSubtreeConcurrencyByBudget` also returns `configured` unchanged when `SizeInBytes == 0`, so zero is the single best value for a hostile peer to declare.

**Failure**: a peer serves 16 blocks each declaring `SizeInBytes = 0`, whose `subtree_data` totals several GiB each. `acquireCatchupPrefetch` charges each the 64 KiB floor, so all 16 workers admit against the 256 MiB budget. `boundSubtreeConcurrencyByBudget` returns 32 for every one. That is 16 x 32 = 512 concurrent fully-parsed `subtree_data` payloads — the unbounded state the issue reports, with the mitigation installed and active.

**Recommendation**: add a bound that does not come from the peer. `fetchSubtreeDataFromPeer` already returns a reader exposing `BytesRead()` (used at `get_blocks.go:912`), so a hard cap on bytes read per `subtree_data` response is cheap and stays attributable to the peer. Alternatively charge observed bytes back against the budget as they are read, and treat the declared size as an optimistic pre-charge only. At minimum, stop exempting `SizeInBytes == 0` from the oversized rule — an undeclared size is the case least worth trusting, not the most.

---

### `services/blockvalidation/get_blocks.go:703, 767-778` — P1: the degenerate path is silent

A block larger than the budget is admitted alone AND drops to one subtree at a time: 16 x 32 falls to 1 x 1. The only signal is a `Debugf` in `acquireCatchupPrefetch`. `boundSubtreeConcurrencyByBudget` logs nothing, the diff adds no Prometheus counter, and I found no catch-up stall detector that would surface it.

**Failure**: BSV mainnet has produced blocks well above the 256 MiB default. Catching up across such a region, every block takes the degenerate path. Subtree fetches serialise, so the prewarm becomes latency-bound — 1024 subtrees at 200 ms each is roughly 3.5 minutes per block instead of seconds. At default log level the operator sees catch-up crawl with no indication why, and the knob that fixes it appears in no emitted line.

**Recommendation**: raise the park log to `Warnf`, and add one line in `boundSubtreeConcurrencyByBudget` naming the block, its declared size, the budget and `blockvalidation_catchup_prefetch_budget_bytes`. Add a counter for blocks clamped to the whole budget. Separately, reconsider sizing the default against the largest expected block rather than against `legacy_blockPrefetchBufferBytes` — the two paths can share a number without sharing the failure mode that number has to survive.

---

Approving: the code is correct, the equivalence holds, and the test coverage is unusually thorough for a memory fix, including a goleak check and both pipe teardown orders. The declared-size trust boundary is worth a follow-up issue rather than more scope here.

icellan:
Reviewed the diff. Design is sound, no blockers. Build, vet and the new tests pass locally.

Checked and clear: acquire/release is paired before the `resultQueue` send, so no reservation is held across a blocked channel write; the `ctx.Err()` guard before `TryAcquire` is needed; `io.Pipe` teardown order (`pr.Close()` then `<-done`) is correct against all four blob stores; `WriteTransactionsToWriter(bw, 0, subtree.Length())` emits byte-identical output to `Serialize()`. The `SubtreeFetchConcurrency` 8→32 and `FetchNumWorkers` 1→16 changes are struct-tag/doc corrections only — `settings/settings.go` already used those values.

Three findings, none blocking.

**`services/blockvalidation/get_blocks.go:938` — the 1 MiB write buffer isn't pooled and isn't charged against the new budget**

`bufio.NewWriterSize(pw, 1024*1024)` allocates per in-flight `subtree_data`. With shipped defaults that path can be live `FetchNumWorkers × SubtreeFetchConcurrency` = 16 × 32 = 512 times → up to ~512 MiB of transient buffers, twice the 256 MiB budget this PR adds to bound this exact pipeline. For a typical small subtree the 1 MiB buffer is larger than the `Serialize()` copy it replaced, so it's a net memory regression on the common path. The function already pools its reader (`bufioReaderPool`, get_blocks.go:879) — pool the writer too, and/or drop to 64 KiB. The buffering only exists to absorb `bt.Tx.SerializeTo`'s many tiny writes, which 64 KiB handles just as well.

**`services/blockvalidation/get_blocks.go:967` — producer errors are misfiled as storage errors, disabling peer failover**

`pw.CloseWithError(writeErr)` funnels `ErrTransactionNil` / `ErrTransactionWrite` into `storeErr`, which is wrapped as `NewStorageError`. `IsLocalError` (errors/error_utils.go:309) treats that as local, so `fetchAndStoreSubtreeAndSubtreeData` skips alternative-peer failover and `recordCatchupPeerFailure` doesn't charge the peer. The old code wrapped a `Serialize()` failure as `NewProcessingError`, which is peer-attributable. A peer serving a `subtree_data` body that parses but fails to re-serialize gets kept in rotation and retried while the node blames its own storage. Narrow given the `missing > 0` precheck, but the fix is two lines: capture the producer error separately from `storeErr`.

**`services/blockvalidation/get_blocks.go:702` — no metric for the budget-parked / oversized-block regime**

Parking on the budget is a `Debugf` only. With the 256 MiB default, a block whose declared size exceeds the budget is admitted alone *and* has its subtree concurrency forced to 1, i.e. ~7.5× the per-block time by this PR's own benchmark. That's the intended behaviour, but in the incident this targets it silently turns an OOM crash-loop into an indefinitely slow catch-up with nothing to distinguish it from a stalled peer. `blockassembly` has `teranode_subtreeprocessor_oversize_batch_admitted_total` for exactly this "the ceiling's guarantee is gone" signal; a `catchup_prefetch_budget_parked_total` plus an oversized-block counter would make it alertable.

ctnguyen:
Thanks for your feedbacks

---

@freemans13

fixed: the fresh 1 MiB `bufio` writer per subtree_data is now a pooled 64 KiB buffer, reset before it goes back to the pool.

fixed: the "`0` restores the previous behaviour exactly" claim — `0` disables the byte budget and the subtree-concurrency rule it drives; the streamed store write applies either way. Corrected in the setting's `longdesc`, in `docs/references/settings/services/blockvalidation_settings.md` and in the code comment.

fixed: the "second copy avoided" comment is now scoped to the file store — the memory and batcher stores `io.ReadAll` the body and S3 copies it into a `bytes.Buffer` before upload.

fixed: the benchmark now sweeps injected per-response latency on a 64-subtree oversized shape, and it confirms your objection rather than answering it — R = t(budget on) / t(budget off) is 13.71 at 0 ms, 15.04 at 20 ms and 21.37 at 200 ms injected latency, rising with latency exactly as a serial fetch paying one round trip per subtree would, and above the 7.5x currently quoted in the PR description; scope: one block per cell, reduced payloads, `httpmock` responders with injected sleeps, per-cell budget lowered to force the oversized branch, one workstation.

deferred: per-subtree weighting — `/subtree` serves bare node hashes and the receive path rebuilds every node with `AddNode(hash, 0, 0)` (`get_blocks.go:570`), so a freshly fetched `Subtree.SizeInBytes` is 0 (a locally held, already-validated subtree does carry real sizes; the peer-supplied path does not), and the only real size requires downloading to disk before admitting, which is a new spill-to-disk subsystem with its own capacity and cleanup policy; follow-up issue (parent #1139).

deferred: the scale-size catch-up measurement — the prerequisite remains unmet, it needs a real catch-up run against a network producing blocks above 256 MiB, which this change cannot produce and which the latency sweep above does not substitute for; procedure and observations are specified in the follow-up issue (parent #1139).

deferred: the same `0`-semantics wording correction in the PR description — the replacement text is written but not yet applied, going into the description of this PR.

---

@ordishs

fixed: `SizeInBytes == 0` is no longer exempt from the oversized rule — a block that declares no size parses its subtrees one at a time; its block-level reservation is still the 64 KiB floor.

fixed: the budget-park log raised from `Debugf` to `Warnf`, one line naming the block, the reserved bytes, the budget and the setting key; the subtree-concurrency clamp logs one line per reason on the clamping branches only.

fixed: counters for the degenerate regimes — `catchup_prefetch_budget_parked_total`, `catchup_prefetch_oversized_blocks_total` (including a declared size too large to represent) and `catchup_prefetch_undeclared_size_blocks_total`.

deferred: a hard cap on bytes read per subtree_data response — the value is the whole problem and nothing has measured it, and a cap set too low rejects honest peers during catch-up and self-isolates the node; follow-up issue (parent #1139).

deferred: charging observed bytes back against the budget — several partially parsed payloads can be resident at once, so it needs an explicit admission rule, retained-partial accounting and a deadlock-freedom argument before it is a bound; follow-up issue (parent #1139).

deferred: reconsidering the 256 MiB default — raising it raises the memory exposure #1139 is about and no operating envelope has been measured; blocked on the scale-size measurement in the follow-up issue (parent #1139).

---

@icellan

fixed: the 1 MiB writer is pooled and dropped to 64 KiB, so small subtrees no longer pay for a megabyte per fetch.

fixed: producer errors are peer-attributable again — `io.ErrClosedPipe`, which can only come from our own `pr.Close()` after the store has already failed, is checked first with `errors.Is`, so a storage failure is never charged to the peer and a genuine producer failure no longer suppresses alternative-peer failover.

fixed: the budget-parked and oversized regimes are now visible through the three counters listed above.

services/blockvalidation/metrics.go
```diff
@@ -311,6 +318,33 @@ func _initPrometheusMetrics() {
 		},
 	)
 
+	prometheusCatchupPrefetchBudgetParked = promauto.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: "teranode",
+			Subsystem: "blockvalidation",
+			Name:      "catchup_prefetch_budget_parked_total",
```
freemans13, line 325: These three counters are not in `docs/references/prometheusMetrics.md`, which carries the whole blockvalidation table including every other catch-up counter (`catchup_blocks_fetched_total`, `catchup_errors_total` and the rest at lines 178 to 182). An operator writing an alert on the budget regime reads that table, so the counters are close to invisible without the rows. Three lines with the `teranode_blockvalidation_` prefix is all it needs.

`docs/topics/services/blockValidation.md` and `docs/howto/miners/minersHowToMonitoring.md` also mention blockvalidation metrics, so worth a glance at whether either wants the same rows.

freemans13 (approved):
**Approving.** Everything I raised is fixed or deferred with evidence I checked myself, and the deferral I care about most is right and my suggestion was wrong. One small thing left, inline on `metrics.go`.

**Retracting the per-subtree weighting idea.** You are correct that it does not work on this path. Catch-up asks the peer for `FileTypeSubtreeToCheck`, which is a flat list of 32-byte node hashes with no sizes in it, and the receive path rebuilds every node with `AddNode(hash, 0, 0)`, so a freshly fetched `Subtree.SizeInBytes` is 0 (`get_blocks.go:541-573`). I checked that at the head commit. Deferring it is the right call, and the reason belongs in the follow-up issue rather than in this PR.

**Verified on 07b962b6:**

- The write buffer is pooled at 64 KiB and reset before it goes back, so a pooled writer cannot retain a live pipe.
- The `0` semantics are corrected in all three places, and the "second copy avoided" comment now names the file store.
- I broke each of the three new fixes on a scratch copy and the new tests caught all three: checking the producer sentinel before `io.ErrClosedPipe` fails `TestSubtreeDataWriteFailure_Classification` and `TestFetchAndStoreSubtreeData_StoreAbortMidStreamStaysLocal`; returning a storage error for a genuine producer failure fails the classification test; re-exempting an undeclared size fails four tests.
- The red `test` job is still the same two pre-existing failures in `subtreevalidation.cachedSubtreeConflictingNodes`, unrelated to this diff.

**Your latency sweep reproduces worse on CI hardware than on your workstation.** The `Bench blockvalidation` job ran it on 4 cores at commit 07b962b6 and reported R of 3.0 at 0 ms, 21.4 to 24.0 at 20 ms, and 30.5 to 30.8 at 200 ms. In absolute terms the 200 ms cell is 846 ms per block with the budget off against 25.9 seconds with it on, for one 64-subtree block. Those numbers are reproducible by anyone who reruns the job, so they are worth quoting in the PR description alongside the 13.71, 15.04 and 21.37 from your machine. It does not change my verdict. A slow catch-up an operator can see and tune beats an OOM crash-loop, and the warning plus the three counters are what make it visible.

**The follow-up issue does not exist yet.** Six deferrals across the three reviews point at "follow-up issue (parent #1139)", and nothing in the repo references it. Please open it before this merges, so the deferred work is not carried only in review threads.

One idea to record there. A per-subtree weight the peer declares per response is still available in the `Content-Length` of the `subtree_data` response, which would be a real number for each payload rather than an average. It needs `util.DoHTTPRequestBodyReaderWithRetry` to surface the header, since it returns a bare `io.ReadCloser` today, and it is only safe paired with the hard read cap @ordishs asked for, since the header is as peer-supplied as `SizeInBytes`.

**Not checked:** still no real catch-up run and nothing at scale-network block sizes. The latency sweep uses httpmock responders with injected sleeps, which is not a network. This review is against 07b962b6.

ctnguyen:
Thanks @freemans13 for the review — per-subtree weighting retraction noted.

**Documentation item: done.** The three counters are now in `docs/references/prometheusMetrics.md`,
in the blockvalidation catch-up table, and in `docs/howto/miners/minersHowToMonitoring.md`, in the
blockvalidation watch list. Two clarifications went in with them, because an operator acting on the
counters needs both: the oversized counter keys off the size a block **declares**, not a measured
size; and a sustained parking rate means contention for the budget rather than proof the budget is
too small, since a slow peer or slow storage also prolongs the reservations already held.

**Follow-up issue:** https://github.com/bitcoin-sv/teranode/issues/4863. Your `Content-Length` idea
is recorded there as a candidate under the per-subtree weighting task, with both conditions:
`util.DoHTTPRequestBodyReaderWithRetry` has to surface the header, since it returns a bare
`io.ReadCloser` today, and it is only safe paired with the hard read cap. The issue's claim that no
pre-fetch size exists on that path has been corrected.

**Pending:** the CI figures go into the PR description alongside the workstation ones — R of 3.0 at
0 ms, 21.4 to 24.0 at 20 ms and 30.5 to 30.8 at 200 ms on 4 cores, and the 200 ms cell at 846 ms
with the budget off against 25.9 s with it on for one 64-subtree block. That edit is not made yet.

icellan (approved):
Re-reviewed at `d519d8222`. All three earlier findings are fixed properly: the write buffer is pooled (`bufioWriterPool`, get_blocks.go:1013), `subtreeDataWriteFailure` (get_blocks.go:1079) splits producer errors into `NewProcessingError` so peer failover actually fires, and the three counters landed in metrics.go:321-341 with usable Help text. `go vet` clean, the new budget/stream tests pass under `-race`. The two `TestValidateBlock_*` package failures reproduce on the merge base, so they aren't this PR's.

Approving. The rest is follow-up, none of it blocking.

**`get_blocks.go:259` + `settings/settings.go:404` — the 256 MiB default is smaller than the blocks this PR exists for**

`x/sync/semaphore.TryAcquire` returns false whenever *any* waiter is queued (semaphore.go:122). So an over-budget block clamps to the whole budget and waits for every other prewarm to drain — and while it waits, every other worker's `TryAcquire` fails too, because that waiter is queued. Stack `boundSubtreeConcurrencyByBudget` returning 1 for the same block on top, and prewarm goes from 16x32 to one block parsing one subtree at a time, and stays there for every subsequent oversized block.

The `longdesc` justifies the number as "the same number as legacy_blockPrefetchBufferBytes so the two sync paths share one mental model". That's the part I'd push on: one is a raw-byte read buffer, the other a per-block admission weight. Sharing a number isn't sharing a meaning, and the queued-waiter behaviour above is real regardless of how the default is documented. Worth defaulting higher, or deriving it from `excessiveblocksize`.

**`get_blocks.go:705` and `:841` — WARN in what is the normal regime**

Unthrottled `Warnf` per parked block, plus a second WARN from `warnSubtreeConcurrencyClamped` per oversized/undeclared block. In the steady-state large-block case above that's one to two WARN lines per block for the whole catch-up — routine operation, not an anomaly. The three counters already carry that signal; Debug or throttle these, otherwise the WARN stream masks real catch-up warnings.

**`get_blocks.go:871` — full reservation held across cheap existence checks**

`acquireCatchupPrefetch` reserves the full declared block weight before `fetchSubtreeDataForBlock` discovers, per subtree, that the data is already local (`subtreeDataExists` early return). On a catch-up resumed after a restart with most subtree_data already on disk, blocks that will parse nothing hold their full reservation through the existence checks, so resumption throttles as hard as a cold sync while using almost no memory. Not a correctness bug, but it makes the serialization above bite in a case with nothing to protect.

**`Server.go:1422` — setting doc omits the RevalidateBlock exception**

`RevalidateBlock` calls `fetchSubtreeDataForBlock` with no reservation while catch-up workers may hold the whole budget, so peak prewarm is budget plus one unbounded per-block prewarm. Deliberate, and documented on the `catchupPrefetchBudget` field — but the `longdesc` an operator actually reads doesn't name the exception.

**`get_blocks.go:1063` — comment asserts an invariant two blob stores violate**

"io.ErrClosedPipe can only come from that function's own `pr.Close()`, which runs after SetFromReader has already returned" is false for the memory store (stores/blob/memory/memory.go:181) and the batcher (stores/blob/batcher/batcher.go:572), both of which `defer reader.Close()` inside `SetFromReader`. The classification still lands correctly either way, so it's comment-only drift — but it's an invariant the next reader will build on.

PR #1751, opened 2026-09-15
test(blockassembly): fix head-age drain race test flake under constrained GOMAXPROCS

icellan:
## Problem

`Test_headAgeGaugeConcurrentDrainRace` flakes. It was observed at roughly 10% over 20 runs on an unrelated branch, and reproduces deterministically under a constrained scheduler.

The poller runs 100,000 iterations of atomic-only reads:

```go
for i := 0; i < 100_000; i++ {
    if q.headAgeMillis(nowMillis) != 0 || q.length() == 0 {
```

Atomic loads are not cooperative preemption points, so on a constrained `GOMAXPROCS` — which CI runners commonly are — this loop can hold its P for the entire duration. A producer that has reserved `queueLength` but has not yet completed publish's CAS on `oldestPendingMillis` then gets no turn at all.

`headAgeMillis` is correct here. Reserve-first accounting means `queueLength` is incremented before the batch is linked, so a non-empty queue reading age 0 is a real but **bounded, self-healing** skew, documented at `queue.go:418-426`. The test's tight loop denies it the scheduling opportunity the self-healing depends on, so a transient skew reads as a latched zero and trips the point-5 invariant.

**The defect is in the test, not in the queue.**

## Fix

One `runtime.Gosched()` at the top of the poll loop. `runtime` was already imported.

This does not weaken the assertion — it removes a starvation artifact. Only a genuinely latched value, one that does not clear no matter how many turns other goroutines get, survives the yield.

## Evidence

Reproduced and verified at `GOMAXPROCS=1`, `-race`:

| | Result |
|---|---|
| Without the yield, 40 runs | **FAIL** — `a non-empty queue must never present a zero head age (point-5 invariant)`, `Should be zero, but was 1` |
| With the yield, 40 runs | pass |

Unconstrained, `-race`: 50 runs of the single test pass, and the whole package passes under `-race`. Note the flake does **not** reproduce unconstrained — removing the yield still passes 25/25 on a many-core machine, which is why this presents as an occasional CI failure rather than a local one.

Confirmed the test still catches a real defect **with the yield in place**: injecting a latched zero into `headAgeMillis` (returning 0 for any non-empty queue) fails the same assertion at `queue_test.go:649`. Injection reverted; `queue.go` is untouched by this PR.

## Scope

Test-only, one file, 13 lines added. No production code changed.


ordishs (approved):

ctnguyen (approved):
Review at commit [85e4aef](https://github.com/icellan/teranode/commit/85e4aef161b235802f219280f1fc686ce8523317)

Nothing to flag — the diagnosis and the fix both hold up, and the write-up is unusually easy to check against the code. I traced the change into the test and the queue it exercises, and the two questions that matter come back clean. First, the yield cannot hide a latched gauge: it sits in the inner re-confirmation loop (`queue_test.go:618-636`), which is entered only on a candidate violation and breaks on the first read that fails to reproduce it, so a value that genuinely refuses to clear still runs all 100,000 iterations and still reaches `violations.Add(1)` at line 639. The outer sampling loop (lines 603-612) is untouched, so detection sensitivity is unchanged, and in the ordinary non-violating case the inner loop exits after one iteration, which is the only place the added cost could have shown up. Second, what the test pins is still the documented contract: `headAgeMillis` clamps to 0 when `queueLength` reads 0 (`queue.go:427-429`), and the reserve-before-publish ordering in `enqueueBatch` (`queue.go:206-209`) is what makes that clamp safe, so a non-empty queue momentarily reading age 0 is exactly the bounded producer-side skew described at `queue.go:419-426`. The test's own header already defines the real defect as a zero that "persists across the consumer's passes" (`queue_test.go:536-539`), which is the reading the yield restores rather than weakens — and the two deterministic siblings, `Test_headAgeGaugeRefreshedOnRefusalPath` and `Test_headAgeGaugeConsumerEmptyClearReReadRepair`, pin the refusal-path refresh and the empty-clear re-read on their own, so coverage of those paths does not rest on this racing test. The one thing that looked like it might bite is that the inner loop does not check `stop`: a genuine violation detected late in the 300ms window would burn 100,000 yields while the test body sits in `wg.Wait()` (line 647). It is a non-issue — the producers and the consumer test `stop` on every iteration and exit promptly, after which the remaining yields have no peer to hand the P to, so the delay stays far below any test timeout and `violations.Add(1)` still lands before `wg.Done()`. On method, so the basis is clear: `go test` was not available to me, so this is static analysis plus `go vet`, which is clean on the package; I did not independently reproduce the 40-run before/after result quoted in the description, and the scheduling mechanism itself rests on runtime behaviour I took as given rather than measured.

PR #1752, opened 2026-09-15
test(blockvalidation): fix cached subtree fixtures in header tests

icellan:
## Summary

The [main CI run](https://github.com/bsv-blockchain/teranode/actions/runs/34962547911) failed in `TestValidateBlock_CachedHeadersShortWindowIsRefused` and `TestValidateBlock_UnusableHeaderRunIsRebuiltByHashWalk`: their cached subtree fixtures contained only transaction IDs, so the cached-subtree conflict check returned EOF before reaching the header assertions.

Use `Serialize()` instead of `SerializeNodes()` for both stored fixtures, including the metadata and conflicting-node trailer required by the reader. The original header rejection and recovery assertions remain intact.

## Validation

- Reproduced both failures before the change; both pass after it.
- Full `services/blockvalidation` and `services/subtreevalidation` suites pass with and without `-race`, using `SETTINGS_CONTEXT=test` and `-tags testtxmetacache`.
- `go vet` and `staticcheck` pass for both packages.
- `golangci-lint` passes for changed code in `services/blockvalidation/...`.
- `git diff --check` passes.


ctnguyen (commented):
Review at commit [ef57c23](https://github.com/icellan/teranode/commit/ef57c23c2d15911e952b5546959793296f45e616)

A tight, correctly diagnosed fix. `SerializeNodes()` writes only the 32-byte node hashes, so the cached-subtree reader runs out of bytes before it can even read the leaf count, while `services/subtreevalidation` writes the full `Serialize()` form in production — which makes this a fixture defect rather than a production bug, exactly as the description claims. The change is minimal, leaves the surrounding assertions alone, and both tests keep checks the EOF could never have satisfied (`require.Contains(t, err.Error(), "does not reach genesis")` in one, `require.NoError` plus two `require.Positive` in the other), so neither can now pass vacuously. Adding a short comment at each site is the right instinct too, since four sibling call sites in the same package already store the full form and two of them carry exactly this kind of note. Nothing in the diff is wrong. My concern is entirely outside it: there are two copies of this change in flight, and I think the other one is the better vehicle for it.

## Blocking issues

### ChiR1 — Duplicate change in flight, and this is the copy to close

**Problem:** PR #1756, by the same author and off the same base commit `4241a96c7`, replaces the same single `subtree.SerializeNodes()` call in the same two files with `subtree.Serialize()`. Post-change the two branches' code is identical apart from how the new comment is wrapped: this PR writes one line, 1756 writes the same sentence over two with a leading `Serialize(), not SerializeNodes():`. Neither description mentions the other.

**Why it matters:** Whichever merges second adds nothing but a comment reflow, or conflicts on precisely the lines this PR touches. Per the run linked in the description, `main` is red on these two tests, so one copy needs to land promptly and the other needs to be gone — two open PRs is exactly the setup where a reviewer approves one, considers the job done, and leaves a stale duplicate that resurfaces the same hunks weeks later. I am marking this blocking rather than filing it as housekeeping because the recommendation below is not to merge this PR at all, which is the strongest blocking statement a review can make about one. "Must be resolved before merge" holds trivially when the advice is that there should be no merge here.

**Fix:** Close this PR in favour of #1756. Before closing, copy the link to the failing `main` CI run out of this description and into #1756's — that link is this PR's one genuine advantage over its twin and it transfers in a single sentence. #1756 is the better vehicle on two counts: its comment leads with `Serialize(), not SerializeNodes():`, which is the phrasing this package has already converged on at `header_context_requeue_test.go:214` and `cached_headers_median_window_test.go:354`; and its description quotes the runtime error, names the sibling test that already used `Serialize()`, and records a revert-and-reconfirm step. To be clear, the code here is correct — this is purely about which of two identical copies survives.

## Non-blocking issues

### ChiR2 — Comment gives the reader half of the reason, not the writer half

**Problem:** The added comment explains why the reader needs the full file, but not why the fixture rather than the reader was the thing at fault. That second fact is that `services/subtreevalidation/SubtreeValidation.go:946` serializes with `Serialize()` before storing under `fileformat.FileTypeSubtree` at `:953`, so the full form is production parity. The sibling comment at `cached_headers_median_window_test.go:354-358` already carries the complete version of this note, ending with "services/subtreevalidation writes Serialize() here in production." Since ChiR1 recommends closing this PR, treat this as advice for #1756's comment rather than for this branch.

**Why it matters:** Not as a safety net — swapping these fixtures back to `SerializeNodes()` reddens both tests immediately, so the comment is not load-bearing as a guard and I would not argue it on those grounds. It is a consistency point, and on a diff that is two-thirds comment, comment quality is squarely on topic. One of the two edited files already carries the fuller form of this exact note a couple of hundred lines further down; matching it costs one clause, and naming the writer as well as the reader makes the rule findable by someone grepping from the wrong call rather than only from the right one.

**Fix:** On the surviving branch, extend the comment to name the writer as well:

  ```go
  // header_run_walk_fallback_test.go:165 and its twin at cached_headers_median_window_test.go:138
  // Serialize(), not SerializeNodes(): cached-subtree validation reads the whole file,
  // including its conflicting-node trailer, and services/subtreevalidation writes
  // Serialize() here in production.
  ```

One thing I checked and am deliberately not raising as an item: whether the fix is incomplete. It is not. `git grep -n "SerializeNodes()" -- services/blockvalidation` still reports plenty of matches, but only four of them store the nodes-only payload under `fileformat.FileTypeSubtree`, and none is a live landmine. `BlockValidation_difficulty_test.go:154` and `:291` do reference the subtree from `block.Subtrees`, but both reject on proof-of-work before subtree validation runs — an ordering pinned on purpose by `TestValidateBlock_PoWCheckedBeforeSubtreeValidation` at `:485` — and both assert a specific error substring, so were that ordering ever to change they would fail loudly on the EOF rather than pass quietly. The fixture stored at `:447` is never read back, because that test's block is built with `Subtrees: []*chainhash.Hash{}` on `:395`. Both callers of the `BlockValidation_error_test.go:205` helper are `t.Skip`ped and assert an error regardless. Everything else feeds an httpmock responder for the `/subtree/` endpoint, where the nodes-only payload is the correct wire form. Leaving all of those alone is the right call.

## Recap

| --- | --- | --- | --- |
| ChiR1 | Duplicate change close this PR | YES | 55% |
| ChiR2 | Comment omits production parity | | 10% |

icellan:
Superseded by #1738 (5e8c31462), which landed the same `SerializeNodes()` -> `Serialize()` fix on main, plus the `FileTypeSubtreeData` store this PR was missing. Closing as redundant.

